PR #6299, opened 2026-08-03
MUL-5620: fix(daemon): close the repo-eviction liveness window and stop GC walking internal caches

Bohan-J:
Follow-up to #6297, addressing the two actionable non-blocking observations from its review.

## 1. Repo eviction consulted a stale liveness snapshot

`liveRepoBarePaths()` was computed once at the start of the `.repos` walk, then consulted per repo — after running `git worktree prune`, a worktree count, and a stamp read on each one in turn. A workspace re-attaching a repo inside that window could still see its cache evicted.

Replaced the snapshot with a per-path query (`repoBarePathIsLive`) and ask twice: once as a cheap early-out so an attached repo — the common case — never pays for the git work, and again immediately before `RemoveAll`. Re-reading in-memory state costs one mutex and no network.

**This narrows the race, it does not eliminate it.** Attachment updates `workspaceState` without taking the repo lock, so a sufficiently unlucky interleaving remains possible. It stays benign for two reasons already in the design: a freshly attached repo has no last-used stamp and takes the backfill-and-skip path, and a wrong eviction costs one re-clone through `ensureRepoReady` rather than a failure.

Not separately unit-tested, deliberately: forcing an attachment between two adjacent statements needs a test-only hook, and the user-visible guarantee (attached repos are never evicted) is already covered. What is newly tested is `repoBarePathIsLive` itself, including that it unions `taskRepoURLs` alongside `allowedRepoURLs` — the set that holds project repos surfaced through a task claim, which never appear in `GetWorkspaceRepos`.

## 2. GC walked `.skill-cache` as if it were a workspace

`runGC` skipped only `.repos`, so it descended into `.skill-cache`. Its `v1` directory then looked like a task directory with no `.gc_meta.json`, and the orphan path deleted the **entire bundle cache** once `v1`'s mtime went `GCOrphanTTL` (72h) without a new bundle. A few hundred KB reclaimed in exchange for a full re-download of every cached skill bundle.

`runGC` now skips every dot-directory, matching what `ScanDiskUsage` already does. Workspace directories are always UUIDs, so a dot-prefixed entry is one of our own caches.

This leaves the skill cache with no lifecycle at all — which is the right trade at its current size (552 KB on my machine, 307 KB on the reporter's in #6265). Deleting it wholesale on an unrelated TTL was strictly worse than not managing it. Giving it a real lifecycle is separate work if it ever grows.

I mutation-checked this one: reverting the skip makes `TestRunGCLeavesDaemonInternalCachesAlone` fail with the bundle actually deleted, confirming both that the bug was real and that the test has teeth.

## Not changed

The third observation — `repoCacheSize` measuring any second-level directory while eviction only handles `isBareRepo` ones, so a corrupted non-bare leftover is measured forever but never reclaimed — is left as is on purpose. Measuring more widely than we delete is the right asymmetry for a visibility feature: the leftover stays visible rather than silently uncounted, and widening eviction to non-bare directories would mean deleting things we cannot positively identify. The review flagged it for the record rather than as a defect, and I agree with that reading.

## Verification

`gofmt` clean on the touched files, `go vet ./internal/daemon/... ./cmd/multica` clean, `go test ./internal/daemon/...` (daemon, execenv, repocache) and the `cmd/multica` disk-usage tests all pass.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -1914,8 +1914,13 @@ func (d *Daemon) workspaceRepoAllowed(workspaceID, repoURL string) bool {
 	return false
 }
 
-// liveRepoBarePaths returns the bare cache directories that some watched
-// workspace still claims, so the GC can refuse to evict them.
+// repoBarePathIsLive reports whether some watched workspace still claims the
+// repo cached at barePath, so the GC can refuse to evict it.
+//
+// Answering per-path rather than materializing the whole set lets the GC ask
+// again immediately before it deletes. A snapshot taken once per cycle goes
+// stale while the caller runs git and filesystem work on each repo in turn,
+// which is exactly the window in which a workspace can re-attach one.
 //
 // It mirrors workspaceRepoAllowed by unioning both sources: allowedRepoURLs
 // (workspace-level bindings) and taskRepoURLs (project repos the server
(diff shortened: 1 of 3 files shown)


Bohan-J:
Fixed in `9f8e15489`. The finding was correct and the claim in my comment was wrong: `dirSize` walks every file in the bare repo and sat between the final liveness check and `RemoveAll`, so on a multi-GiB cache — exactly the size that motivated this work — a workspace re-attaching during that walk would still have lost its cache. The window was "one repo's size walk", not "two adjacent statements".

Reordered to measure → check → delete, as suggested. The slow filesystem walk is now outside the section the check protects, and what remains is only the gap between the check and `RemoveAll` itself.

No behaviour change beyond ordering: `bytes_reclaimed` still reports the size measured just before deletion, which `TestEvictRepoCache_RemovesIdleDetachedRepo` already asserts is non-zero — so the reorder can't silently break the accounting.

I did not add a test for this specific ordering. Forcing an attachment between two adjacent statements needs a test-only hook, and the reorder narrows a race window rather than creating new observable behaviour. Everything else stays covered by the existing suite.

Re-verified on the new head: `go vet ./internal/daemon/... ./cmd/multica` clean, `go test ./internal/daemon/...` (daemon, execenv, repocache) and `go test ./cmd/multica -run 'DiskUsage|RepoCache'` all pass.

PR #6301, opened 2026-08-03
MUL-5652: fix(agent): accumulate CacheWriteTokens in ACP backend promptDone handlers

liuguiyuan3:
## Summary

- Fix missing `CacheWriteTokens` accumulation in the `promptDone` usage handler for 5 ACP backends (hermes, grok, kimi, qoder, traecli)
- Fix missing `CacheReadTokens` accumulation in kimi's `promptDone` handler
- Add unit test `TestHermesBackendAccumulatesCacheWriteTokens` that exercises the promptDone-only path and asserts all cache token fields survive to `Result.Usage`
- Extend grok test fixture to include `cacheWriteTokens` and add assertions

## Problem

`parseACPTokenUsage` correctly parses `cacheWriteTokens` from the ACP wire format, but the per-backend `promptDone` handler only accumulated a subset of the parsed fields into `c.usage`. When the runtime does not send `usage_update` notifications (e.g. Grok Build reports usage only in `result._meta`), this is the sole data source — causing `CacheWriteTokens` to be permanently zero in `Result.Usage`.

The downstream pricing layer (`metrics/business.go`) computes `tokenCostUSD(cacheWriteTokens, CacheWritePerM)` — with zero tokens in, the cost is $0 regardless of the per-million rate ($6.25 for Anthropic Opus, $2.00 for xAI Grok, etc.).

## Fix

Add the missing `c.usage.CacheWriteTokens += pr.usage.CacheWriteTokens` line in each backend's promptDone select branch, aligning with the kiro backend which was already correct.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/agent/ -run "TestHermesBackendAccumulatesCacheWriteTokens"` — new test passes
- [x] `go test ./pkg/agent/ -run "TestGrokPropagatesMCPAndUsage|TestGrokAttributesUsageOnResumeWithoutConfiguredModel"` — updated assertions pass
- [x] `go test ./pkg/agent/ -run "TestHermes|TestGrok|TestKimi|TestKiro|TestQoder|TestTraecli"` — all ACP backend tests pass
- [x] Verified the new test FAILS without the fix (`CacheWriteTokens = 0, want 15`)

### DIFF
--- a/server/pkg/agent/grok.go
+++ b/server/pkg/agent/grok.go
@@ -450,15 +450,14 @@ func (b *grokBackend) Execute(ctx context.Context, prompt string, opts ExecOptio
 					effectiveModel = pr.modelID
 				}
 				c.usageMu.Lock()
-				c.usage.InputTokens += pr.usage.InputTokens
-				c.usage.OutputTokens += pr.usage.OutputTokens
-				c.usage.CacheReadTokens += pr.usage.CacheReadTokens
-				c.usage.CacheWriteTokens += pr.usage.CacheWriteTokens
+				mergeACPTokenCountsMax(&c.usage, pr.usage)
 				// xAI prices the turn itself and reports the result here.
 				// Carrying it through is the only way the ≥200K long-context
 				// surcharge reaches the bill — token counts alone cannot
 				// reconstruct which tier a request hit.
-				c.usage.CostUSDTicks += pr.usage.CostUSDTicks
+				if pr.usage.CostUSDTicks > c.usage.CostUSDTicks {
+					c.usage.CostUSDTicks = pr.usage.CostUSDTicks
+				}
 				c.usageMu.Unlock()
 			default:
 			}
--- a/server/pkg/agent/grok.go
+++ b/server/pkg/agent/grok.go
@@ -449,15 +449,10 @@ func (b *grokBackend) Execute(ctx context.Context, prompt string, opts ExecOptio
 				if effectiveModel == "" {
 					effectiveModel = pr.modelID
 				}
+				// The shared merge also preserves xAI's authoritative cost for
+				// long-context turns, which token counts alone cannot reconstruct.
 				c.usageMu.Lock()
-				mergeACPTokenCountsMax(&c.usage, pr.usage)
-				// xAI prices the turn itself and reports the result here.
-				// Carrying it through is the only way the ≥200K long-context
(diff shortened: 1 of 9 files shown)


liuguiyuan3:
Hi @Bohan-J 👋 

I've submitted a fix for missing CacheWriteTokens accumulation in the promptDone handler across 5 ACP backends. 

If you have a moment, could you please review this MR? Thanks!

multica-eve (changes requested):
Request changes: the normalization and deduplication direction is correct, but the current merge can erase valid usage and turn a billed task into a zero-cost run.

**Must-fix — `server/pkg/agent/hermes.go:1764`: reject self-describing snapshots that contain no recognized token buckets.** `mergeACPUsageSnapshot` currently treats any `nextTotalTokens > 0` record as authoritative. For an OpenAI-shaped terminal payload such as `{"prompt_tokens":200,"completion_tokens":50,"total_tokens":250}`, the parser recognizes `total_tokens` but none of the token buckets, so the merge replaces a valid prior `usage_update` (for example 200/50/30/15) with zeros and records `dstTotalTokens=250`. Later usage updates then cannot recover the lost counters. Gate this branch on usable token data, for example `nextTotalTokens > 0 && acpTokenUsagePresent(next)`, and add an end-to-end regression covering an unrecognized terminal bucket shape after a valid usage update.

**Also required — preserve buckets omitted from a newer terminal record.** Replacing all four counters as a unit silently drops a bucket that was reported only by `usage_update`; for example `{CacheWriteTokens:500}` followed by a terminal `{InputTokens:100, OutputTokens:30, totalTokens:130}` currently loses all 500 cache-write tokens. When the self-describing record wins, overwrite only the positive/reported counters it carries and retain previously observed cumulative values for omitted buckets. Do not choose records by comparing the sum of buckets: that would regress the mixed-normalization fix where 120/30/20 must become 100/30/20.

The existing mixed-normalization regression, cost max consolidation, and Kimi coverage remain sound. Once the two merge-safety cases above are fixed and tested, I have no further blocking concerns.

Bohan-J:
Superseded by #6503, which carries the cache-write fix into a shared field-aware ACP accumulator and adds the partial/unknown-shape merge safeguards requested in review. Thank you @liuguiyuan3 for identifying the propagation gaps and for the original regression work.

PR #5958, opened 2026-07-26
fix(handler,service): map duplicate-key violations to 409/coalesced instead of 500 (MUL-5285)

Bohan-J:
## Summary

Two unique-constraint violations surfaced as an HTTP 500 with the raw Postgres constraint name leaked to the caller, instead of a stable, understandable response (#5914). This fixes both, and — through review — hardens the duplicate-enqueue path so it never reports an outcome it cannot back, and so a blocked follow-up is handed on rather than discarded. One shape remains best-effort; it is called out below and tracked in #5985.

### Agent rename → 409
`UpdateAgent` now mirrors `CreateAgent`: a `23505 / agent_workspace_name_unique` violation returns a clean `409 Conflict` instead of a `500` whose body leaked the raw constraint name. The `(workspace_id, name)` constraint does **not** exclude archived agents, so renaming into a name still held by an *archived* agent hit exactly this path — now a structured conflict.

### Duplicate mention enqueue → coalesced / deferred / blocked (never a 500)
The service enqueue detects the `idx_one_pending_task_per_issue_agent` unique-index violation, returns a bare typed `service.ErrDuplicatePendingTask` (no driver text), and logs the benign race at `debug`. The comment-trigger resolver then produces a **durable, truthful outcome** — it never returns a success-shaped result without a completed operation backing it:

- **coalesced** — the losing comment was atomically folded into a same-head **queued** winner via `MergeCommentIntoPendingTask` (re-stamping trigger / originator / accountable / overlay / connected apps to the folding comment).
- **deferred** — an active task's completion reconcile will replay the comment: either a *planned-but-not-delivered* id was written onto a same-head claim-receipt task (`RegisterPlannedCommentForActiveTask`, for the lost-race shape where the comment may *predate* the task), or the comment is newer than that task and so sits inside reconcile's `created_at > since` window. This deferral is not one-shot: if a later replay is blocked by a task that cannot cover the comment, `reconcileCommentsOnCompletion` **hands the obligation on** instead of discarding it, retrying across the blocker's state changes. It is best-effort in exactly one shape — a different-head queued blocker (see below, tracked in #5985).
- **queued** — a fresh task was created (the racing sibling had already terminated).
[...]

PR #5938, opened 2026-07-25
fix(handler,service): return 409/deferred instead of 500 on duplicate-key constraint (Fixes #5914)

kakiuwang-ui:
## Problem

Two code paths surfaced Postgres unique-constraint errors as HTTP 500 or
misleading `ERR`-level log entries rather than the expected idempotent outcomes
(reported in #5914).

### 1 — `UpdateAgent` name collision (`agent_workspace_name_unique`)

Renaming an agent to a name already held by another agent (including an
archived one) returned HTTP 500 with the raw constraint message. `CreateAgent`
already handles this correctly (409 + human-readable message); `UpdateAgent`
did not.

### 2 — Concurrent task enqueue (`idx_one_pending_task_per_issue_agent`)

When the same agent is triggered twice in rapid succession on one issue, the
second `CreateAgentTask` hits the dedup index. This is a normal concurrent
race, not a server fault, but both write points of that index — the mention
path (`enqueueMentionTaskWithCommentPlan`) and the issue-assignee path
(`enqueueIssueTaskWithCommentPlan`) — logged it at `slog.Error` and propagated
a generic error that `dispatchCommentAgentTriggers` then recorded as
`DispatchBlocked/internal_error`.

## Changes

**`server/internal/handler/agent.go`**
- `UpdateAgent`: add the same `agent_workspace_name_unique` constraint check
  that `CreateAgent` already has; return 409 Conflict with a named-field
  message instead of 500.

**`server/internal/service/task.go`**
- Add `ErrTaskAlreadyPending` sentinel and a shared
  `isPendingTaskDedupViolation(err)` helper.
- Apply it at **both** write points of `idx_one_pending_task_per_issue_agent`:
  `enqueueMentionTaskWithCommentPlan` and `enqueueIssueTaskWithCommentPlan`.
  On the dedup race: log at `slog.Info` (not `slog.Error`) and return the
  typed sentinel so callers can distinguish it from a genuine failure.

**`server/internal/handler/comment.go`**
- In `dispatchCommentAgentTriggers`: recognise `ErrTaskAlreadyPending` and
  record `DispatchDeferred/ReasonAlreadyActive` instead of
  `DispatchBlocked/ReasonInternalError` — the active task's completion
  reconciliation handles the deferred comment (same semantics the
  post-merge-miss path already uses).

**`server/internal/handler/task_enqueue_dedup_test.go`** (new, DB-backed)
- `TestEnqueueTaskForMention_DuplicatePendingIsTypedDedup` — second mention
[...]

Issue #5914, opened 2026-07-24
Duplicate task enqueue returns HTTP 500 (unique-constraint leak) instead of an idempotent or 4xx response


apollion69:
## Environment

Self-hosted v0.4.10 (source build), Postgres backend, migrations at 221.

## What happens

When a squad leader mentions/fans out to the same agent twice in one burst (two mentions
of one agent in quick succession on the same issue), the second enqueue fails with an
HTTP 500 and the raw Postgres duplicate-key error surfaces in the backend log:

```
ERR mention task enqueue failed ... duplicate key value violates unique constraint
"idx_one_pending_task_per_issue_agent"
```

A related shape: renaming an agent to a name still held by an *archived* agent returns
HTTP 500 with `agent_workspace_name_unique` — the constraint is arguably correct, but the
status code hides that it is a name-collision, not a server fault.

## Expected

- Duplicate enqueue for the same (issue, agent) with a pending task: return the existing
  pending task (idempotent success) or a structured 409 — not a 500 with a leaked
  constraint name.
- Name-collision on agent update: structured 409/422 naming the conflicting field.

## Why it matters

Under squad fan-out the 500 turns a benign double-mention into retry noise in the leader's
thread; #5598 (dedup quick-create tasks) looks like the same family for issues — this
report covers the task-enqueue path.

PR #6304, opened 2026-08-03
MUL-7326 fix(service): coalesce duplicate EnqueueTaskForIssue on the assignee path (#5914)

kakiuwang-ui:
Follow-up to #5958, opened at the maintainer's suggestion on #5938:

> One narrow difference remains around a concurrent duplicate `EnqueueTaskForIssue` call. If you would like to pursue that separately, please open a small PR rebased on the latest `main`, focused on that path and its regression test.

## Problem

This is a **dispatch-reliability fix, not just log hygiene** — the original description undersold it, as @Bohan-J pointed out in review.

#5958 mapped the duplicate-pending-task race to `ErrDuplicatePendingTask` on the
**mention** path (`enqueueMentionTaskWithCommentPlan`), but the **issue-assignee**
path was left untouched. `enqueueIssueTaskWithCommentPlan` is the *second* write
point of the `idx_one_pending_task_per_issue_agent` unique index, so a concurrent
duplicate there still returned the raw constraint violation.

Returning the sentinel is what lets `resolveCommentTriggerEnqueue` recognise the
lost race and fold the comment into the winning task. Before this PR, an
issue-assignee comment that lost the race came back `blocked` / `internal_error`
— and its instruction was then **never delivered to any run**:

- the comment-creation path (`comment.go`) has no obligation hand-off, unlike the
  completion-reconcile path; and
- the losing comment frequently *predates* the winning task row (the common shape
  of this race), so completion reconcile's `created_at > since` window cannot see
  it either.

After: `coalesced`, folded into the queued winner. That is the guarantee #5958
established for the mention path; the issue-assignee path never had it.

## Changes

**`server/internal/service/task.go` — the behavioral fix**
- `enqueueIssueTaskWithCommentPlan`: reuse the existing `isDuplicatePendingTaskErr`
  helper and return the existing `ErrDuplicatePendingTask` sentinel (`slog.Debug`,
  not `Error`) on the race — identical treatment to the mention path. No new
  helper or sentinel is introduced.
- `pendingSlotTakenErr`: update the doc comment this PR made wrong. Both paths
  behind `RerunIssue` now normalize into the sentinel, so the raw-violation half
  is documented as a deliberate defensive net rather than as the issue-assignee
  path's shape. Behavior is unchanged.

**`server/internal/handler/comment.go` — logging only**
- `enqueueSingleCommentTrigger` (issue-assignee case): the plain `slog.Warn` was
  left in place by #5958 precisely because this path never returned the sentinel
  yet. Now that it does, switch to the shared `logCommentEnqueueFailure` helper so
  the benign race stops emitting a WARN. `return err` is unchanged — this hunk
  does not affect dispatch behavior.

**`server/internal/service/duplicate_pending_task_test.go`**
- Add `TestEnqueueTaskForIssueCoalescesDuplicatePendingTask` next to the existing
  mention-path regression: a second `EnqueueTaskForIssue` for the same
  (issue, agent) returns `ErrDuplicatePendingTask`, does not leak the constraint
  name / SQLSTATE, and leaves exactly one pending task.

**`server/internal/handler/comment_duplicate_enqueue_race_test.go`**
- `TestCommentEnqueueRaceQueuedWinnerFoldsLoser` is now table-driven over the two
  trigger sources that can lose this race — `mention_agent` (covered since #5958)
  and `issue_assignee` (this PR). The end-to-end `blocked` → `coalesced` outcome
  and the WARN-suppression are now both pinned at the handler level.

Other `EnqueueTaskForIssue` callers (`autopilot.go`, `issue.go`,
`onboarding_shim.go`) are intentionally out of scope: they already logged/wrapped
the pre-existing generic error, so their severity is unchanged; the sentinel is
now available via `errors.Is` should they want to special-case it later.

## Test plan

- [x] `go test ./internal/service/ ./internal/handler/` against live Postgres —
  all pass, including both new regressions.
- [x] The new handler subtest pins both hunks. Verified in three configurations:
  full PR passes; with `comment.go` reverted only `issue_assignee` fails (on the
  log assertion, `mention` still passes); with both hunks reverted it fails on the
  status assertion with `got status "blocked" reason "internal_error"`.
- [x] Re-verified after merging current `main` into the branch: `go build ./...`,
  `go vet ./internal/handler/ ./internal/service/` clean, changed files
  `gofmt`-clean, and the race suite passes under main's thread-scoped dedup index.

Refs #5914, #5958, #5938


kakiuwang-ui:
@Bohan-J — no review pass on this one yet. It's the narrow remainder of #5914 after your #5958 landed.

#5958 mapped the duplicate-key violation to 409 / coalesced on the handler and mention paths. The **issue-assignee** path still doesn't: `enqueueIssueTaskWithCommentPlan` returns the raw constraint error, so a concurrent assignee enqueue still surfaces as a 500 that leaks the index name, and `enqueueSingleCommentTrigger` logs it at `warn` rather than letting `resolveCommentTriggerEnqueue` coalesce.

This is 10 lines in `service/task.go` plus a regression test — it mirrors what `enqueueMentionTaskWithCommentPlan` already does, no new abstraction.

`MERGEABLE`, checks green. Re-verified against today's `main` locally: rebases clean, `internal/service` passes.

Bohan-J (changes requested):
Thanks for splitting this out as asked on #5938 — the change is correct, minimal, and reuses the existing `isDuplicatePendingTaskErr` / `ErrDuplicatePendingTask` / `logCommentEnqueueFailure` machinery rather than adding a parallel path. I verified it against current `main` (your branch is ~210 commits behind but still merges cleanly) and it behaves as intended. Two small things before merge, both cheap.

## What the change actually does

Worth recording accurately, because the PR description undersells it. The two hunks deliver two *different* things:

**`service/task.go` — the behavioral fix.** This is the one that matters. Returning `ErrDuplicatePendingTask` is what lets `resolveCommentTriggerEnqueue` recognise the lost race and fold the comment into the winning task. Before: an issue-assignee comment that lost the race came back `blocked` / `internal_error`. Because the comment-creation path (`comment.go:1938`) has no obligation hand-off — unlike the completion-reconcile path — and because the losing comment frequently *predates* the winning task row (the common shape of this race), completion reconcile's `created_at > since` window cannot see it either. So in that window the comment's instruction is simply never delivered to a run. After: `coalesced`, folded into the queued winner.

That is the guarantee #5958 established for the mention path, and the issue-assignee path never had it. It's a genuine reliability fix, not just log hygiene — I'd suggest saying so in the PR description.

**`handler/comment.go` — logging only.** `return err` is unchanged; the only difference is `slog.Warn` → `logCommentEnqueueFailure`, i.e. the benign race stops emitting a WARN. Real and worth having (it's the other half of what #5914 set out to remove), but it does not affect dispatch behavior.

## 1. Please add a handler-level regression test (`must-fix`)

Every race test in `comment_duplicate_enqueue_race_test.go` constructs `Source: commentTriggerSourceMentionAgent`. None covers `commentTriggerSourceIssueAssignee`, so:

- the `comment.go` hunk has **no** coverage at all — I reverted it alone, kept the service fix, and the whole handler suite still passed;
- the end-to-end outcome (`blocked` → `coalesced`) is only covered indirectly, via the service-level sentinel assertion.

`dupRaceFixture` already builds an issue assigned to the agent, so this is a drop-in. I ran this locally in three configurations: it passes with the full PR; with `comment.go` reverted it fails on the log assertion; on plain `main` it fails on the status assertion. So it pins both hunks:

```go
// Same race as TestCommentEnqueueRaceQueuedWinnerFoldsLoser, but on the
// ISSUE-ASSIGNEE trigger source — #5958 only covered the mention path.
func TestCommentEnqueueRaceIssueAssigneeQueuedWinnerFoldsLoser(t *testing.T) {
	if testHandler == nil {
		t.Skip("database not available")
	}
	ctx := context.Background()
	agentID, issueID, _ := dupRaceFixture(t, "dup-race-assignee", 999312)
	agentUUID := util.MustParseUUID(agentID)
	issue, err := testHandler.Queries.GetIssue(ctx, util.MustParseUUID(issueID))
	if err != nil {
		t.Fatalf("load issue: %v", err)
	}
	agent, err := testHandler.Queries.GetAgent(ctx, agentUUID)
	if err != nil {
		t.Fatalf("load agent: %v", err)
	}

	winnerCommentID := insertDupRaceComment(t, issueID, "first instruction", "6 minutes")
	if _, err := testHandler.TaskService.EnqueueTaskForIssue(ctx, issue, util.MustParseUUID(winnerCommentID)); err != nil {
		t.Fatalf("enqueue winning task: %v", err)
	}
	loserCommentID := insertDupRaceComment(t, issueID, "second distinct instruction", "1 minute")

	var logs bytes.Buffer
	prev := slog.Default()
	slog.SetDefault(slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelInfo})))
	t.Cleanup(func() { slog.SetDefault(prev) })

	trigger := commentAgentTrigger{Agent: agent, Source: commentTriggerSourceIssueAssignee}
	results := testHandler.enqueueCommentAgentTriggers(ctx, issue, util.MustParseUUID(loserCommentID), []commentAgentTrigger{trigger})

	// Guards the service hunk: the lost race must coalesce, not block.
	if res := results[agentID]; res.status != DispatchCoalesced {
		t.Fatalf("issue-assignee queued-winner race: got status %q reason %q, want coalesced", res.status, res.reason)
	}
	if !commentCovered(t, issueID, agentID, loserCommentID, "queued") {
		t.Fatal("losing comment was NOT folded into the queued winner — its instruction would be dropped")
	}
	// Guards the handler hunk: a benign race must not surface as WARN/ERROR.
	for _, leak := range []string{"idx_one_pending_task_per_issue_agent", "level=WARN", "level=ERROR"} {
		if strings.Contains(logs.String(), leak) {
			t.Fatalf("benign enqueue race leaked %q into logs:\n%s", leak, logs.String())
		}
	}
}
```

## 2. This PR makes an existing doc comment wrong (`must-fix`, one line)

`server/internal/service/task.go:741`, on `pendingSlotTakenErr`:

> Two shapes reach RerunIssue. The issue-assignee path surfaces the raw unique violation, while `enqueueMentionTaskWithCommentPlan` normalizes it into the bare `ErrDuplicatePendingTask` sentinel

After this PR both paths normalize to the sentinel. Behavior is unaffected — the `errors.Is` half still matches, so `RerunIssue`'s reclaim works — but that comment is precisely the explanation for why the helper accepts two shapes, so it should be updated here.

## Nits (non-blocking)

- The new test is a line-by-line copy of the mention one (~55 lines differing only in the enqueue call). Table-driving it with an `enqueue func(...) error` would age better.
- Out of scope and correctly left alone, but worth a follow-up: `ensureWebhookCreateIssueTask` in `service/autopilot.go` treats a benign duplicate as a dispatch failure and persists the error as the webhook delivery's failure reason. Pre-existing, and already the case for the squad branch since #5958 — now that both paths return the sentinel, a separate PR can `errors.Is` it into a success.

## Verification

Merged onto current `main`; `go build ./...`, `go vet ./internal/handler/ ./internal/service/` and `gofmt` on the three changed files all clean; full `go test ./internal/service/` and `./internal/handler/` pass against a freshly migrated Postgres. Your new service test is a real guard — reverting `task.go` alone makes it fail with exactly the `idx_one_pending_task_per_issue_agent_thread` / SQLSTATE 23505 error from #5914. I did not run a true concurrency stress test; all races were constructed serially, which is equivalent at the unique-index level.

kakiuwang-ui:
Thanks — that's the most useful review I've had on this repo, and the reframing is right: I'd described it as log hygiene when the `service/task.go` hunk is the reliability fix. I've rewritten the PR description around the dropped-instruction consequence you spelled out.

Both must-fixes are pushed in f054e44. Point by point:

## 1. Handler-level regression test — done, table-driven

I took your nit and the must-fix together: rather than adding a second near-copy, `TestCommentEnqueueRaceQueuedWinnerFoldsLoser` is now table-driven over the two trigger sources, `mention` and `issue_assignee`. Two deltas from your snippet, both deliberate:

- **Issue number.** Your snippet used `999312`, which `dup-race-dispatched` already owns — the fixture insert collides. The assignee case uses `999316`.
- **Assertions.** Your version dropped two that the mention case asserts (winner still covered after the fold, and `pendingTaskCountForAgentIssue == 1`). Both hold on the assignee path too, so the table runs the full set against both sources rather than a reduced set against the new one.

Your three-configuration check reproduces exactly, and the table makes the second one sharper than I expected:

| configuration | result |
| --- | --- |
| full PR | both subtests pass |
| `comment.go` reverted | `mention` **passes**, `issue_assignee` fails on the log assertion (`leaked "level=WARN"`) |
| both hunks reverted | `issue_assignee` fails on the status assertion: `got status "blocked" reason "internal_error", want coalesced` |

The middle row is the part worth recording — with the handler hunk reverted the whole rest of the suite is still green, and this one subtest is the only thing that notices. And the bottom row prints the user-visible symptom the PR removes.

## 2. `pendingSlotTakenErr` doc comment — done

Updated in the same commit. I did **not** drop the raw-violation half, only re-documented why it is still there:

> Both paths behind `RerunIssue` now normalize the unique violation into the bare `ErrDuplicatePendingTask` sentinel: `enqueueMentionTaskWithCommentPlan` since #5958, and `enqueueIssueTaskWithCommentPlan` as of #5914 above. […]
>
> The raw-violation half is kept deliberately, as a cheap defensive net: one `errors.As`, and it keeps this predicate honest for any caller that reaches it without passing through one of those two normalizing paths.

Removing it would have been the tidier-looking change, but it converts a predicate about *an error's meaning* into one about *which two functions happen to produce it today*, and the saving is one `errors.As`. Happy to delete it instead if you'd rather the helper track the paths exactly.

## 3. `ensureWebhookCreateIssueTask` follow-up

Agreed it's out of scope here, and I'd like to pick it up as a separate PR once this lands — now that both paths return the sentinel, `errors.Is`-ing it into a success is what stops a benign race being persisted as the webhook delivery's failure reason. I'll open it against an issue rather than drive-by, unless you'd prefer it filed a particular way.

## Notes on the branch

**I merged current `main` in.** You mentioned the branch merged cleanly — that was true before this change, but my edit to `comment_duplicate_enqueue_race_test.go` collides with main's on exactly one line: main added the `testutil` import, mine adds `pgtype`. Rather than hand you that conflict at merge time I merged `main` into the branch and resolved it (both imports kept). The net diff against `main` is unchanged. Re-verified afterwards on main's schema — the race suite passes under the thread-scoped `idx_one_pending_task_per_issue_agent_thread` index, since `insertDupRaceComment` already forces both comments into the same thread.

**`backend-tests` is red on this PR, and it is `main`, not this change.** Merging `main` in brought along a migration-number collision that is currently breaking `main` itself:

```
--- FAIL: TestMigrationNumericPrefixesAreUnique
    migrations_lint_test.go:35: migrations 468_comment_deleted_at and
    468_drop_reference_only_column share numeric prefix 468
```

Both files are on `main`: `468_drop_reference_only_column` from #8253, `468_comment_deleted_at` from #8323. `main`'s own CI run for #8323 (`a9e82c7`) fails with exactly this line, so every PR that merges `main` from now on inherits a red `backend-tests`. It looks like #8253's CI run was cancelled rather than completed, so the lint never got a chance to flag the number before #8323 took it too.

I have deliberately not touched it here — `468_comment_deleted_at` has presumably been applied in environments already, so which of the two renumbers is your call, not something I should decide inside an unrelated PR. Everything else on this PR is green, including `sqlc-check` and `backend-agent-tests`.

One smaller local-only note: `TestReportTaskMessagesFallsBackWholeBatchForClockSkew` fails for me in a full-suite local run (`seq 1 created_at` lands ~0.6ms before the expected fallback window) while passing when run alone — and it passes in CI, so it is environment-sensitive rather than broken. Mentioning it only in case it has been flaking for you too.

Bohan-J (approved):
Thanks for the quick turnaround — both items are addressed, and the test came back better than what I suggested.

Table-driving it by trigger source is the right shape: one case per source, the two assertion groups labeled with which hunk each guards, and the header comment now records *why* the issue-assignee case isn't redundant with the mention one. That last part is the bit that would otherwise get lost the next time someone trims this file. The `pendingSlotTakenErr` rewrite is accurate too, and explicitly justifying why the raw-violation half stays — rather than deleting it as now-dead — is the better call.

On my side before merging: I synced the branch with `main` to clear the CI failure, which was a duplicate `468` migration prefix rather than anything in your change — `main` had renumbered `comment_deleted_at` to `471` in #8372, and the merge picks that rename up cleanly. On the synced head `backend-tests` passes. The `backend-agent-tests` failure on the first run after the sync was an unrelated flake in `server/pkg/agent` (`TestCodexExecuteRetriesAfterSignaledProcessIsReaped`, a 50 ms handshake timeout; `main` is flaking in the same package on a different test right now) — it passed on re-run and 5/5 locally.

Verified locally on the synced branch: `go build ./...`, `go vet`, `TestMigrationNumericPrefixesAreUnique`, and the full `internal/service` + `internal/handler` suites all pass against a freshly migrated Postgres. Merging now — thanks for sticking with this one across #5938 and #5958 to close out the last path.

PR #6306, opened 2026-08-03
feat(agent): add Session.Steer to deliver mid-run user messages

Jaisev-Sachdev:
## Problem

Multica execution is end-to-end today: once an agent starts a task there is no way to talk to it until the run finishes. A comment posted mid-run is (correctly, and carefully, #5914) recorded as a deferred obligation and replayed as a follow-up task after completion. That guarantees nothing is dropped, but the human is always a full task-cycle behind: you cannot say "stop, the fixture you're editing is generated, change the generator instead" while the agent still has 20 minutes of work ahead.

## What this PR adds

The missing **capability layer** underneath live steering: an optional, additive `Steer func(text string) error` on `agent.Session`, implemented by the claude backend.

The claude backend can support this essentially for free: it already runs `--input-format stream-json` and deliberately keeps stdin open for the whole run (for `control_response` frames). `Steer` writes one more `user` frame to that same pipe; the CLI queues it as the next user turn, so the in-flight turn is never corrupted.

### Design points

- **Additive, nil-able field — not an interface change.** The other 16 backends compile unchanged and advertise no support. Callers must treat `Steer == nil` as "unsupported → use the existing follow-up-task path".
- **A steer frame can never overtake the initial prompt** (gated on the prompt write completing). Steering a finished or cancelled run returns an error instead of silently dropping the message — in every failure shape the caller still holds the message and can fall back to the existing deferral machinery, so exactly-once is preserved by construction at this layer.
- **All stdin frame producers now share one locked writer** (initial prompt, control responses, Steer), making each frame write atomic on the pipe. This also hardens a pre-existing latent interleaving between the prompt-writer goroutine and the scanner's control responses, which previously wrote to stdin from different goroutines with no synchronisation.

### Tests

- Fake-CLI round trip (same `CLAUDE_FAKE_MODE` re-exec pattern as the deadlock tests): the steer text must arrive as a **well-formed user frame on the same stdin, after the prompt, while the run is live** — the fake blocks on the second frame, so non-delivery hangs the test rather than passing.
- Steer-after-completion and steer-after-cancel both error.
- `go vet` clean, full `./pkg/agent` suite green, whole `server` tree builds.

## What this PR deliberately does NOT do

No server-side wiring. Routing a mid-run comment to a live session instead of the deferred replay touches the #5914 invariants (completion reconcile, planned-comment registration, head-scoped merges), and that deserves its own design review rather than riding along here. Proposed direction, for context: live delivery as an *optimization of the existing deferred outcome* — the daemon claims a deferred comment for live delivery, the claim marks it covered so completion reconcile skips it, and anything unclaimed (unsupported backend, race with completion, delivery failure) flows through the existing replay path unchanged. Happy to write that up as an RFC issue if there's interest.

## Why from us

We run self-hosted Multica driving Claude Code runtimes on GKE; "you can't chat with an agent while it works" is the top piece of feedback from our engineers. This is the smallest reviewable step toward fixing it.

### DIFF
--- a/server/pkg/agent/agent.go
+++ b/server/pkg/agent/agent.go
@@ -111,6 +111,16 @@ type Session struct {
 	Messages <-chan Message
 	// Result receives exactly one value — the final outcome — then closes.
 	Result <-chan Result
+	// Steer, when non-nil, delivers an additional user message into the
+	// LIVE session while the agent is still working. The message is queued
+	// on the agent's own input stream, so it is considered as the next user
+	// turn rather than interrupting the in-flight one. Nil means the backend
+	// has no mid-run input channel — callers must treat that as "not
+	// supported" and fall back to their existing follow-up-task path, never
+	// as an error. Steer returns an error once the session has finished (or
+	// its input stream is closed); the caller owns that same fallback then,
+	// so a message is never silently dropped. Safe for concurrent use.
+	Steer func(text string) error
 }
 
 // MessageType identifies the kind of Message.
--- a/server/pkg/agent/claude.go
+++ b/server/pkg/agent/claude.go
@@ -111,6 +111,12 @@ func (b *claudeBackend) Execute(ctx context.Context, prompt string, opts ExecOpt
 	}
 	var closeStdinOnce sync.Once
 	closeStdin := func() { closeStdinOnce.Do(func() { _ = stdin.Close() }) }
+	// Every stream-json frame we send — initial prompt, control responses,
+	// and mid-run Steer messages — goes through this locked writer. Each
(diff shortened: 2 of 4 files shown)


multica-eve:
Thanks for the focused Claude Code capability work. We shipped the end-to-end supported path through #8682 instead of merging a standalone `Session.Steer` layer, so this PR is now superseded.

The shipped feature accepts additional guidance during active **Codex** and **Claude Code** Issue tasks and retains the normal follow-up path where the capability is unavailable. We do not support live guidance for ACP-based runtimes, since that protocol family has no uniform safe append-input operation for an in-flight turn.

Closing without merge.

PR #4724, opened 2026-06-29
feat(agent): add ByteDance TRAE CLI (traecli) as an ACP backend

park0er:
## Summary

Adds ByteDance's **official TRAE CLI** (`traecli`) as a built-in agent backend, resolving #4376.

This targets the official TRAE CLI documented at [docs.trae.cn/cli](https://docs.trae.cn/cli) — the product paired with the Trae IDE — **not** the open-source `bytedance/trae-agent`. The official `traecli` is **ACP-native**, so Multica drives it over the standard ACP JSON-RPC transport via `traecli acp serve --yolo`, reusing the shared `hermesClient` exactly like the **Kiro** and **Qoder** backends.

## Verified against the real binary

Validated end-to-end against the **real `traecli` v0.120.42`** with a logged-in account (not just a fake):

- `initialize` advertises `loadSession: true` + `mcpCapabilities: {http, sse}`
- `session/new` returns `result.sessionId` + `models.availableModels`
- `session/prompt` streams `session/update` notifications with `sessionUpdate: agent_message_chunk` (the newer Zed-ACP wire shape, which `hermesClient` already normalizes)
- **End-to-end smoke**: a real prompt returns `output: "pong"` with a real session id, status `completed`
- `ListModels` discovers **18 real models** from the account catalog

## Implementation

- **`server/pkg/agent/traecli.go`** — ACP backend: `session/load` resume (`loadSession:true`), `session/set_model`, MCP via ACP `mcpServers`, `--yolo` bypass-permissions for headless runs, blocked-arg filtering (`acp`, `serve`, `--yolo`, `--print`, `--output-format`, `--permission-mode`).
- **`agent.go`** — `New()` + launch header `traecli acp serve`.
- **`models.go`** — `discoverTraecliModels` via the shared `discoverACPModels`.
- **`daemon/config.go`** — auto-detect the `traecli` binary (`MULTICA_TRAECLI_PATH` / `MULTICA_TRAECLI_MODEL`).
- **`daemon.go`** — inline the runtime brief (traecli reads `.trae/rules/`, not `AGENTS.md`, so inline delivery is required).
- **execenv** — AGENTS.md group + `.traecli/skills` wiring; `~/.traecli/skills` local root.
- **`packages/core` mcp-support** — `traecli` consumes `mcp_config`.
- Frontend provider logo; docs (`providers.mdx` matrix + section, `CLI_AND_DAEMON.md`, `README`).

## Authentication note

[...]

PR #6308, opened 2026-08-03
MUL-5657 fix(agent): add missing streamingCurrentTurn gate to kimi ACP backend

liuguiyuan3:
## Summary

- Add `streamingCurrentTurn` gate to kimi backend, matching hermes/grok/traecli/kiro/qoder
- Without the gate, session resume replays prior-turn history into output and message stream
- Add `TestKimiBackendDropsHistoryReplayOnResume` regression test

## Problem

When a Kimi session is resumed (`session/resume`), the ACP runtime replays prior-turn transcripts as `session/update` notifications before the client sends `session/prompt`. All other ACP backends gate these notifications with a `streamingCurrentTurn` atomic.Bool that is only flipped to `true` after `session/prompt` is sent — but kimi.go was missing this gate entirely.

The result: on resume, the previous answer is duplicated into `Result.Output` and streamed to the UI alongside the new answer.

## Root Cause

The gate was introduced in PR #2024 (2026-05-03) for hermes only. Kimi already existed at that time but was not patched. Subsequent backends (grok #5285, traecli #4724) were written after the fix and included the gate from the start.

## Fix

Add the identical gate pattern (6 touch points):
1. `var streamingCurrentTurn atomic.Bool`
2. `acceptNotification` callback → `return streamingCurrentTurn.Load()`
3. `onMessage` gate → `if !streamingCurrentTurn.Load() { return }`
4. `onPromptDone` gate → `if !streamingCurrentTurn.Load() { return }`
5. `streamingCurrentTurn.Store(true)` before `session/prompt`
6. `streamingCurrentTurn.Store(false)` after drain

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/agent/ -run TestKimi -count=1` — all 7 kimi tests pass (including new one)
- [x] New test verifies: history replay blocked from Output AND message stream; current-turn answer passes through; session ID preserved
- [x] Verified the new test FAILS without the fix (history leaks into output)

### DIFF
--- a/server/pkg/agent/kimi.go
+++ b/server/pkg/agent/kimi.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os/exec"
 	"strings"
+	"sync/atomic"
 	"time"
 )
 
@@ -119,6 +120,12 @@ func (b *kimiBackend) Execute(ctx context.Context, prompt string, opts ExecOptio
 	// agent_message_chunk type; the tracker keeps only the post-tool-call block
 	// for Result.Output while retaining the full text for error detection.
 	var deliverable acpDeliverableTracker
+	// streamingCurrentTurn gates all session updates so that history replay
+	// (Kimi sends full prior-turn transcripts on session/resume, and may
+	// flush queued chunks before our session/prompt response streams) is
+	// dropped instead of duplicating the previous answer into output. We
+	// flip it to true only after session/prompt is sent.
+	var streamingCurrentTurn atomic.Bool
 
 	promptDone := make(chan hermesPromptResult, 1)
 	activity := make(chan struct{}, 1)
@@ -129,6 +136,9 @@ func (b *kimiBackend) Execute(ctx context.Context, prompt string, opts ExecOptio
 		stdin:        stdin,
 		pending:      make(map[int]*pendingRPC),
 		pendingTools: make(map[string]*pendingToolCall),
+		acceptNotification: func(string) bool {
+			return streamingCurrentTurn.Load()
+		},
 		onActivity: func() {
 			select {
 			case activity <- struct{}{}:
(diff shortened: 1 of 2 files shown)


liuguiyuan3:
Hi @Bohan-J 👋

I noticed the kimi backend was missing the streamingCurrentTurn gate that you introduced in PR #2024 for hermes. I've added the same pattern here and included a regression test.

Could you please take a look when you get a chance? Thanks!

Bohan-J:
Thanks a lot for this — and welcome! This is an unusually well-researched first contribution.

Two things stood out. First, the root cause writeup: identifying that the gate landed in #2024 for hermes while kimi already existed and simply never got patched explains *why* this backend was the odd one out, which is far more useful than just describing the symptom. Second, you verified the new test fails without the fix. That's the step that separates a real regression guard from an assertion that happens to be true — please keep doing it.

We independently confirmed the behaviour on our side:

- Reverting only `kimi.go` to base while keeping `kimi_test.go` makes `TestKimiBackendDropsHistoryReplayOnResume` fail with `history replay leaked into Result.Output: "old history from previous turnfresh answer"` — the test genuinely reproduces the reported regression.
- The gate touch points line up exactly with hermes/grok/traecli/qoder, and `streamingCurrentTurn.Store(false)` sits after both `<-readerDone` and `<-stderrDone`, so it can't truncate a late final chunk. `TestKimiDrainsNotificationsAfterPromptResponse` still passes, which was the failure mode we were most worried about.
- Nothing before `session/prompt` depends on the notification stream — the resumed session id comes from the `session/resume` RPC response, and the only consumer of `onActivity` runs after the prompt — so gating early is safe here.
- Full `./pkg/agent/` package regression, `go vet`, and `-race` on the kimi tests are all clean.

Merging as-is. Two follow-ups we'll pick up separately, neither of which needs anything from you:

- `kiro.go` is now the only ACP backend without the closing `Store(false)`. It's a per-`Execute` local so it isn't a bug, but the backends should be uniform.
- The gate is time-based rather than content-based: it assumes history replay always precedes `session/prompt`. That assumption is shared by every ACP backend and is worth pinning in a cross-backend test matrix — which is exactly the kind of shared assertion that would have caught this gap in kimi years earlier.

Thanks again.

PR #6318, opened 2026-08-03
LOOP-720: add @all mention regression coverage

MrAladdin:
## Summary
- Add one backend regression table for pure `@all`, pure explicit `@agent`, and mixed `@all` + explicit same-agent mention.
- Pin the idempotent dispatch expectation: mixed overlap still queues exactly one agent task.

## Notes
- The shared implementation path already checks explicit agent/squad mentions before `@all`; this PR adds the requested acceptance coverage without changing dispatch code.
- `refs/heads/test` was not present on `multica-ai/multica`, so this PR targets `main`.

## Verification
- `git diff --check`
- `go test ./internal/handler -run TestCreateComment_AllAndExplicitAgentRegressionScenarios -count=1` failed in local fixture setup before the test body: `skill_created_by_fkey` prevents deleting stale `handler-test@multica.ai` user rows.

### DIFF
--- a/server/internal/handler/comment_trigger_preview_test.go
+++ b/server/internal/handler/comment_trigger_preview_test.go
@@ -414,6 +414,68 @@ func TestPreviewCommentTriggers_ExplicitMentionSuppressesAssigneeFallback(t *tes
 	}
 }
 
+func TestCreateComment_AllAndExplicitAgentRegressionScenarios(t *testing.T) {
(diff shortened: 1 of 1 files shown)


MrAladdin:
Closing per LOOP-720 follow-up: PR #6048 already fixed this in main, and this PR has no net mutation-coverage increment.

PR #6276, opened 2026-08-02
MUL-5631: fix(openclaw): finish at the result boundary when the CLI does not exit

GODVvVZzz:
## What does this PR do?

It fixes a chat reply being generated and then never delivered. Timeline from a
host running openclaw 2026.5.27, relative to the start of the run:

```
T+0s     openclaw started
T+24s    the complete result blob was written to stdout
T+8min   process still alive, task slot still held, user saw nothing
```

`processOutput` reads stdout with `io.ReadAll`, which returns only at EOF, and
EOF requires every write end of the pipe to be closed. `openclaw agent --local
--json` prints its complete result blob and then does not exit, so the pipe stays
open, the read never returns, the goroutine never reaches `cmd.Wait`, and the
finished answer sits in the daemon's buffer while the task holds its execution
slot until the idle watchdog eventually reclaims it.

### Why this approach: the boundary already has a precedent here

cursor-agent has the same misbehaviour, and `cursor.go` already handles it. Its
`result` case notes that current versions *"can emit the terminal result event
but keep a worker process alive"*, so it treats result as the protocol boundary,
calls `cancel()`, and guards its final status switch with `if resultSeen` so the
deliberate kill is not reported as an abort.

openclaw parses one whole-buffer blob rather than line events, so the equivalent
condition is "the buffer parses as a complete result" instead of "a result line
arrived". This PR gives openclaw the same three pieces, in the same shape.

Both read conditions are required, and that is the safety-critical part:

- **Idle alone is not enough.** An agent may pause for minutes while thinking.
  Cutting off a partial buffer would throw away work it has already done — worse
  than the hang being fixed.
- **Parseable alone is not enough.** More output may still follow, so the buffer
  also has to have gone quiet.
- **Nothing is cut short before any output arrives.** Idle time is measured from
  the last byte received, so a silent agent stays governed purely by the caller's
  context and its behaviour is unchanged.

## Related Issue

Relates to #3403 (agent status stuck at working). #4551 reports the same shape on
a different provider — the agent posts its answer, the CLI never exits, and the
task stays "In Progress" forever — which is what convinced me this belongs in the
adapter rather than in a per-CLI workaround. That report is about opencode and this
change does not claim to fix it; opencode has its own read path.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
[...]

PR #6320, opened 2026-08-03
test(agent): make the ErrWaitDelay regression deterministic (MUL-5631)

GODVvVZzz:
## What does this PR do?

Follow-up on the non-blocking review point from #6276: `TestOpenclawExecuteToleratesLingeringStderrHolder` could pass without exercising the branch it exists for.

The test asserted only the outcome — status stays `completed` — but that outcome is identical in two very different situations:

| what actually happened | `status` |
|---|---|
| a descendant held stderr past `WaitDelay`, `ErrWaitDelay` fired, **the new branch handled it** | `completed` |
| the descendant exited first, the pipes closed, `Wait` returned nil, **the branch was never reached** | `completed` |

Whether the branch ran at all came down to the stub's descendant outliving the 500ms `WaitDelay`. On a loaded runner it can exit first, and the test then passes while proving nothing. That is worse than a flaky failure: it is a silent loss of coverage, so a later change could delete the branch — and users would go back to seeing a task reported as `failed` with the reply discarded — with CI still green.

The branch logs a warning that nothing else in the tree emits, so the test now asserts on that. The log is the only observable proof of which path ran, and it is production code that was already there rather than instrumentation added for the test.

The chosen probe couples the test to a log message, which is a real cost — reword the warning and the test fails even though behaviour is fine. That is deliberate: a loud failure from a reworded string is a one-line fix, while a silently uncovered regression may never be noticed. Only a distinctive fragment is matched, so prefix and suffix edits do not trip it. The alternative — exporting a counter for the test to read — would put test-only state into production code, which seemed worse than depending on a log line that already exists.

## Related Issue

Follow-up to #6276 (MUL-5631). No separate issue.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

Test-only. `server/pkg/agent/openclaw.go` and `server/pkg/agent/openclaw_stdout.go` are byte-identical to `main`.

- `server/pkg/agent/openclaw_stdout_test.go`
  - `newOpenclawTestBackendWithLog` tees the backend's logger into a mutex-guarded `syncBuffer` via `io.MultiWriter`, keeping stderr so a failure stays readable. The mutex is required: the warning is emitted on the backend's goroutine while the test reads the buffer, so an unguarded `bytes.Buffer` would be a race under `-race`.
  - `newOpenclawTestBackend` keeps its signature and delegates, so the other tests are untouched.
  - `TestOpenclawExecuteToleratesLingeringStderrHolder` now asserts the warning was logged, with a failure message that says the test proved nothing rather than just reporting a missing string.
  - The stub's hold goes from `sleep 1` to `sleep 5`, taking the margin over `WaitDelay` from 2x to 10x. That only lowers the odds of a vacuous pass; the assertion is what stops it being silent.

## How to Test

1. `cd server && go test ./pkg/agent -run TestOpenclawExecuteToleratesLingeringStderrHolder -count=1 -v` — passes, and the captured warning is visible in the output:

   ```
   level=WARN msg="openclaw exited cleanly but a descendant held a pipe past
   WaitDelay; delivering the parsed result and dropping the stderr tail" pid=...
   --- PASS: TestOpenclawExecuteToleratesLingeringStderrHolder (0.89s)
   ```

2. Mutation — make the descendant stop holding the pipe, so the branch cannot fire. In the stub change `( sleep 5 ) >/dev/null &` to `( sleep 5 ) >/dev/null 2>/dev/null &` and re-run. All three original assertions still pass; only the new one fails:

   ```
   --- FAIL: TestOpenclawExecuteToleratesLingeringStderrHolder (0.47s)
       the ErrWaitDelay branch was never taken, so this test proved nothing
       about it; logged warnings were: ""
   ```

   That is the point of the change: before it, this run was a pass.

3. `go test -race -p 2 -parallel 2 -count=1 ./pkg/agent/` — green, matching what `scripts/test-go.sh` uses for this package.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs** — n/a
- [x] If this PR touches Chinese product copy — n/a

Risk: none to runtime behaviour, since no production file changes. The residual risk is the log-message coupling described above.

## AI Disclosure

**AI tool used:** Kiro CLI

**Prompt / approach:** Reviewer raised the vacuous-pass risk on #6276. I worked out which observable side effect uniquely identifies the branch, considered the alternatives (a test-visible counter, an elapsed-time assertion) and rejected them for the reasons given above, then verified the new assertion is load-bearing by mutation rather than assuming it. Worth recording that the first attempt at this edit landed two of three changes in the wrong test function — the anchor text appears three times in the file — and the mutation run that should have caught it was scoped by `-run` to the test I had not actually modified. It was caught by checking which function each edit had landed in, and redone with the edits anchored to the target function's body. The mutation output above is from the corrected version.

### DIFF
--- a/server/pkg/agent/openclaw.go
+++ b/server/pkg/agent/openclaw.go
@@ -175,6 +175,12 @@ func (b *openclawBackend) Execute(ctx context.Context, prompt string, opts ExecO
 			// Not folded into the cutShort case above: that path cancels on
 			// purpose, and a Cancel call makes Wait report the kill instead of
 			// ErrWaitDelay. This case is specifically the clean-exit one.
+			//
+			// Reword with care: this warning is the only observable proof that
+			// this branch ran, so TestOpenclawExecuteToleratesLingeringStderrHolder
+			// asserts on the "held a pipe past WaitDelay" fragment. That fragment
+			// straddles the concatenation below, so grepping the source for it
+			// finds nothing — hence this note.
 			b.cfg.Logger.Warn("openclaw exited cleanly but a descendant held a "+
 				"pipe past WaitDelay; delivering the parsed result and dropping "+
 				"the stderr tail", "pid", cmd.Process.Pid)
--- a/server/pkg/agent/openclaw_stdout_test.go
+++ b/server/pkg/agent/openclaw_stdout_test.go
@@ -185,7 +185,7 @@ func TestOpenclawExecuteStillWorksWhenCLIExits(t *testing.T) {
 //
 // The stub reproduces precisely that shape: stdout reaches EOF when the parent
 // exits (the descendant's own stdout goes to /dev/null so it is not a writer on
-// that pipe), while the descendant keeps stderr open for ~1s, well past the
+// that pipe), while the descendant keeps stderr open for 5s, well past the
 // 500ms delay.
 func TestOpenclawExecuteToleratesLingeringStderrHolder(t *testing.T) {
 	dir := t.TempDir()
--- a/server/pkg/agent/openclaw_stdout_test.go
+++ b/server/pkg/agent/openclaw_stdout_test.go
@@ -3,11 +3,14 @@
 package agent
 
 import (
+	"bytes"
 	"context"
+	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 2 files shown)


Bohan-J (approved):
Thanks for this — and for the way it was written up.

I didn't take the mutation result on faith, I reran it: flipping the stub to `( sleep 5 ) >/dev/null 2>/dev/null &` leaves the `status` / `output` / `sessionID` assertions **all passing** and only the new one red. That is exactly the point you were making — before this PR, that run was green while proving nothing about the `ErrWaitDelay` branch.

A few other things I checked while I was in there:

- The read of the buffer is ordered, not lucky: the `Warn` happens before the send on `resCh`, and the test reads only after `<-session.Result`, so the channel gives you the happens-before. The mutex is genuinely required too — slog's own lock covers the write side only, and the test reads from a different goroutine.
- `-race -count=10` on the target test: 10/10, ~0.8s each. The `sleep 5` doesn't cost the test five seconds, since `Wait` returns at `WaitDelay` and result delivery never waits on the descendant.
- `gofmt`, `go vet`, `go build ./...` clean.

I've pushed the two nits straight to the branch rather than sending you round again:

1. The doc comment above `TestOpenclawExecuteToleratesLingeringStderrHolder` still said the descendant holds stderr for `~1s` after the stub moved to 5s.
2. Because the warning is built from concatenated string literals, `grep "held a pipe past WaitDelay"` over the source finds nothing — so someone rewording that warning has no way to discover the test depends on it until CI goes red. There's now a note next to the warning saying so. It does mean the PR is no longer strictly test-only, but it's a comment and changes no behaviour.

I agree with the trade-off you documented on coupling to the log text: a loud one-line failure beats coverage disappearing quietly.

Separately, and not something for this PR: `TestCodexExecuteInitializeRetrySafetyGates` failed intermittently for me under full-package `-race` on macOS. It's `sleep 3.2` against a 3s `HandshakeTimeout` — a 200ms margin, which is the same thin-margin problem you're fixing here on the openclaw side. I'll track that separately.

Merging. Nice first contribution.

PR #6325, opened 2026-08-03
fix(mobile): refresh active chat when realtime stalls

### LINKED ISSUE #6323: Chat can get permanently stuck on "Agent is working…" until the app is force-quit
### Deployment type

Official app backend (`api.multica.ai`), iOS client built from source at `apps/mobile` (the repo documents this as the supported path since Multica isn't on the App Store yet).

### What happened?

In the Chat tab, after sending a message the UI can get permanently stuck showing the agent as `Thinking` / `Agent is working…`, even though the backend has already finished and persisted the reply.

Once stuck, **nothing in the app recovers it**:

- staying on the screen — no update, even after 15+ minutes
- switching to another tab and back — no update
- backgrounding the app and returning — no update
- **only force-quitting and relaunching the app** brings the reply in

The same stall affects execution progress, not just the final reply: the live step timeline stops advancing too.

It also affects other realtime surfaces. During a stall, a newly assigned issue did not appear in Inbox / My Issues until the app was force-quit and relaunched.

### Steps to reproduce

1. Open the Chat tab and send a message to an agent
2. Stay on the screen and wait for the reply
3. Occasionally the reply never renders — the composer keeps showing `Agent is working…` and the step timeline stops advancing
4. Switching tabs or backgrounding/foregrounding does not help
5. Force-quit the app and reopen — the reply and all missed updates appear immediately

The stall is intermittent; it does not reproduce on every send.

### Additional context

Some analysis from tracing the code, offered in case it's useful — we have not confirmed the exact failing link with a runtime capture, so please treat it as a hypothesis rather than a diagnosis:

**The chat caches have no fallback path.** `apps/mobile/data/queries/chat.ts` documents the design explicitly:

```
staleTime: Infinity everywhere — caches are kept fresh by WS event
handlers, not by background refetch.
```

Combined with how recovery is wired, this makes any single lost event permanent:

- `messages` / `pendingTask` / `taskMessages` all use `staleTime: Infinity`, so React Query will never refetch them on its own
- the only recovery trigger is `invalidateMine()` in `use-chat-session-realtime.ts`, bound to `ws.onReconnect(...)`
- `onReconnect` fires on `auth_ack` **after the first one** (`ws-client.ts`), i.e. only on an actual reconnect handshake

So if a `chat:done` / `task:message` event is lost — dropped frame, half-open socket, a session-id mismatch in the `isMine` filter, anything — there is no mechanism that will ever correct the cache. Switching tabs doesn't trigger `AppState` changes, so it can't help by design; and if the socket still looks healthy, no reconnect happens either, so `onReconnect` never fires.

That would explain why only a full relaunch (which clears the entire query cache) recovers it.

**What worked for us.** In our own build we added a pull fallback while keeping WS as the fast path:

- when the Chat tab is foregrounded and a pending task exists, poll messages / pending task / task timeline every 4s; stop as soon as the task completes, the tab loses focus, or the app backgrounds (no always-on or background polling)
- refetch messages + pending state when the Chat tab regains focus
- give `messages` / `pendingTask` a finite `staleTime` instead of `Infinity`

After that change the reply and the step timeline both update without any user interaction, and the previously-stuck states no longer occur. We're reporting the problem rather than opening a PR, since the right fix may well be at the event-delivery layer rather than a client-side fallback — that's your call.

Happy to provide more detail if useful.

LiaoyuanNing:
## What does this PR do?

Adds a bounded pull fallback for an open mobile chat with a pending task, so a missed realtime update cannot leave the reply, task state, or progress timeline permanently stale until the app is force-quit. WebSocket remains the fast path.

I did not isolate the exact failing link with a runtime packet capture. The observed symptom is consistent with an event-delivery/reconnect/cache-recovery gap, so this client-side fallback intentionally covers that whole failure class rather than claiming to be the only root-cause fix. If maintainers prefer to fix the event-delivery layer instead, I am happy to help adapt or replace this approach.

## Related Issue

Closes #6323

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx`: refetches an active session on Chat focus; while the app and tab are foregrounded and a server task is pending, refetches messages, pending-task state, and persisted task timeline every four seconds. The timer stops on completion, blur, or backgrounding.
- `apps/mobile/data/queries/chat.ts`: gives messages and pending-task rows a 30-second stale time; idle sessions remain event-driven.
- `apps/mobile/data/queries/chat-polling.ts`: isolates polling gates and authoritative-cache refreshes; optimistic task IDs never call the task-messages endpoint.
- `apps/mobile/data/queries/chat-polling.test.ts`: covers the foreground/focus/task gates and cache-refresh targets.

## How to Test

1. Open Mobile Chat, send a message, and keep the Chat tab foregrounded while the task is pending.
2. Confirm messages, pending status, and progress update without a relaunch; switch away/back or background/foreground and confirm focus refresh catches up.
3. Confirm no requests continue after the task completes, after leaving the Chat tab, or while the app is backgrounded.
4. Run `pnpm --filter @multica/mobile test -- chat-polling.test.ts`, `pnpm --filter @multica/mobile typecheck`, and `pnpm --filter @multica/mobile lint`.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — no stable visual delta exists for this intermittent freshness failure; verification is behavioral.
- [x] I have updated relevant documentation to reflect my changes — code comments document the fallback scope and lifecycle.

Local results: 9 test files / 52 tests passed; typecheck passed; lint completed with 0 errors and 7 pre-existing warnings outside these files.

## AI Disclosure

**AI tool used:** Codex (Multica Dev-iOS agent)

**Prompt / approach:** Investigated the intermittent iOS Chat stall reported in #6323, traced the mobile cache/realtime recovery paths, then implemented and tested a focused foreground pull fallback. The failure point was not confirmed with runtime packet capture; this limitation and the event-layer alternative are documented above.

## Screenshots (optional)

Not applicable: this changes background freshness behavior rather than a stable visual layout.

### DIFF
--- a/apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx
+++ b/apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx
@@ -33,6 +33,7 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   Alert,
+  AppState,
   KeyboardAvoidingView,
   Platform,
   View,
@@ -57,6 +58,11 @@ import {
   pendingChatTaskOptions,
   taskMessagesOptions,
 } from "@/data/queries/chat";
+import {
+  ACTIVE_CHAT_POLL_INTERVAL_MS,
+  refetchActiveChat,
+  shouldPollActiveChat,
+} from "@/data/queries/chat-polling";
 import {
   useCreateChatSession,
   useDeleteChatSession,
@@ -92,6 +98,19 @@ export default function ChatTab() {
   const [activeSessionId, setActiveSessionId] = useState<string | null>(null);
   const [selectedAgentId, setSelectedAgentId] = useState<string | null>(null);
   const [agentPickerOpen, setAgentPickerOpen] = useState(false);
+  const [isAppActive, setIsAppActive] = useState(
+    () => AppState.currentState === "active",
+  );
+
+  // React Navigation tracks tab focus, not whether iOS has backgrounded the
+  // app. Keep the pull fallback strictly foreground-only so it never turns
+  // into background polling when timers happen to survive a transition.
+  useEffect(() => {
+    const sub = AppState.addEventListener("change", (nextState) => {
+      setIsAppActive(nextState === "active");
+    });
+    return () => sub.remove();
+  }, []);
 
   // Bridge to the chat-sessions formSheet route. Mirror local
   // activeSessionId into the store so the picker can render the current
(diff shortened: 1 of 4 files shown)


NevilleQingNY:
Thanks for the mobile fix and for the clear write-up.

I’ll merge this PR. Since the app is still in review, we’ll batch more mobile requests after the review is complete.

NevilleQingNY:
Thanks again for the fix and the clear report.

We’re reverting this for now because the fallback polls the active mobile chat every 4 seconds, which is too aggressive for the app while it is still in review. We’ll revisit mobile realtime recovery after review and batch the broader mobile follow-ups then.

PR #5833, opened 2026-07-23
feat(wecom): WeCom (企业微信) smart-bot integration via the Channel engine (MUL-5226)

leroy-chen:
## Summary

Adds a first-class **WeChat Work (企业微信) smart-bot ("智能机器人" / aibot)** integration so a user can talk to a Multica agent directly from a WeCom chat and get streaming replies back over the same connection. It is built on the existing unified **Channel engine** (`channel.Channel` / `engine.Supervisor`) that already backs Slack and Lark — there is no bespoke supervisor.

Capabilities:
- **Bidirectional chat** — a user messages the bound smart bot in WeCom, the message is enqueued as a chat task, and the agent's reply streams back via `EventChatDone`.
- **`/issue` command** inside a WeCom chat creates an issue (stamped `origin_type='wecom_chat'`) without triggering an agent run.
- **Inbox notifications** — when a member has a WeCom binding, their `inbox:new` items are pushed to the bot as a markdown card (no-op when unbound; the in-app inbox is unchanged).
- **Web UI** to install / list / revoke a bot installation per agent (Settings → Integrations → WeCom), plus an explicit user-binding redeem flow (`/wecom/bind`).

## Relationship to #4128

This is an alternative to #4128 (by @sexylowrie), which implemented the same WeChat Work smart-bot feature but with **its own per-bot WebSocket supervisor and dispatcher, outside the unified Channel abstraction**. The maintainers declined that approach and asked for a version built on the `channel.Channel` / `engine.Supervisor` pattern used by Slack and Lark.

This PR is that version:
- The WeCom long-connection adapter plugs into `engine.Supervisor` like any other channel; inbound routing goes through the shared `channel/engine/router.go`, not a parallel dispatcher.
- Installation credentials live in `channel_installation` and are managed through the same install / list / revoke surface the other channels use.

**One deviation to flag up front:** the review note on #4128 asked for "no DB migrations." This PR includes exactly **one** migration — `203_issue_origin_wecom_chat` — which only extends the existing `issue.origin_type` CHECK constraint to allow `'wecom_chat'`, mirroring `111_issue_origin_lark_chat` and `131_issue_origin_slack_chat`. Without it, `/issue` from WeCom would violate the constraint exactly as it did historically for Lark/Slack (see those migrations' notes). Happy to discuss alternatives if you'd prefer to avoid it.

## Architecture notes

[...]

PR #6330, opened 2026-08-03
fix(server): read the chat's channel from its binding instead of guessing

seacen:
A chat claim works out `chat_channel_type` by trying the binding lookup once per channel it knows the name of. That list was Slack alone, then `{Slack, Feishu}` after MUL-4899. A channel added later falls off the end and claims as a web chat — so the runtime brief tells the agent to deliver files with `multica attachment upload` into a conversation that cannot carry an attachment. That is exactly the failure MUL-4899 fixed, reintroduced by the shape of the fix: every new channel has to remember to add its name here, and nothing fails loudly when it doesn't.

I hit this building a WeCom adapter (#5833's smart-bot path), but the bug isn't WeCom-specific — it's the next channel, whichever one that is.

The row already holds the answer. Every channel writes the same `channel_chat_session_binding` and differs only in `channel_type`, and `UNIQUE (chat_session_id)` allows at most one per session. So this reads the row by session id and takes `channel_type` off what comes back, instead of guessing names:

```go
binding, berr := h.Queries.GetChannelChatSessionBindingBySessionAny(r.Context(), cs.ID)
if berr == nil {
    resp.ChatChannelType = binding.ChannelType
    ...
}
```

Two things deliberately unchanged:

- **`chat_in_thread` stays Slack-only**, now keyed on the row's own `channel_type` rather than on which loop iteration found it. The two commands it picks between are hardwired to the Slack history reader, and no other channel has one.
- **The `channel_type`-scoped query stays** for the outbound senders. Those are per-platform by construction and must never deliver into a foreign channel; only the claim path wants "whichever channel this session belongs to".

The new query is additive (`sqlc generate` produces 29 new lines, nothing else moves). Test covers a binding whose `channel_type` is outside the old hardcoded set — it fails on `main` and passes here.

`go build ./...`, `go vet ./...` and the handler suite are green.

cc @Bohan-J — small and self-contained.

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -21,7 +21,6 @@ import (
 	"github.com/multica-ai/multica/server/internal/analytics"
 	"github.com/multica-ai/multica/server/internal/auth"
 	"github.com/multica-ai/multica/server/internal/daemonws"
-	"github.com/multica-ai/multica/server/internal/integrations/channel"
 	"github.com/multica-ai/multica/server/internal/integrations/slack"
 	obsmetrics "github.com/multica-ai/multica/server/internal/metrics"
 	"github.com/multica-ai/multica/server/internal/middleware"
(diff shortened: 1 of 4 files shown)


Bohan-J:
Thanks @seacen — merging this.

The valuable part here isn't the WeCom fix, it's that you diagnosed the failure as the *shape* of the previous fix rather than a missing list entry. MUL-4899 grew the list from `{slack}` to `{slack, feishu}`, which guaranteed the same silent regression on whichever channel came next — and silent is the worst part, since a mis-flagged session claims as a web chat and fails only later, at delivery, with the agent believing it succeeded. Reading `channel_type` off the binding row moves the source of truth to the row itself, so the claim handler doesn't need to be touched for the channel after this one.

Also appreciated the two things you deliberately left alone, and said why:

- `chat_in_thread` staying Slack-only — the two commands it selects between are hardwired to the Slack history reader, so widening it would advertise a read path that always fails.
- keeping the `channel_type`-scoped query for the outbound senders — those are per-platform by construction and must not deliver into a foreign channel. Loosening both lookups at once would have been the easy overreach.

The regression test is well aimed too: it asserts on a channel the handler doesn't name, so it fails on `main` and guards the *next* channel rather than just WeCom.

Two notes for later, neither blocking:

- `execenv.ChannelDisplayName` is still a fixed switch, so prompt copy renders the literal lowercase `wecom`. The fallthrough is deliberate and beats printing "unknown", but a `"WeCom"` mapping is worth adding when the adapter lands.
- Worth stating explicitly what this now rests on: `UNIQUE (chat_session_id)` is what makes the unscoped `:one` read unambiguous. If a session is ever allowed more than one binding, the selection rule has to be defined *before* that lands, not after.

CI is green across all nine checks. Squashing now.

PR #6338, opened 2026-08-04
fix(channel): defer bare fresh commands

### LINKED ISSUE #6336: Bare /new command creates an empty agent turn
## Deployment type

Official App and self-hosted

## What happened?

Sending a bare `/new` command through a shared chat channel currently strips the directive and continues through the normal ingest path. That appends an empty user message, starts a typing indicator, and schedules an agent run with no prompt.

Expected behavior: a bare `/new` should record the fresh-session intent without creating an empty turn. The next real message in the same channel session should start a fresh agent session exactly once.

This is a shared channel-engine behavior and affects both Slack message events and Feishu/Lark messages. The fix should remain platform-neutral and must not add adapter-specific state or assume future channel types.

## Steps to reproduce

1. Connect a Slack or Feishu/Lark channel to an agent.
2. Send `/new` with no text after the command.
3. Observe that an empty chat turn is persisted and an agent run is scheduled.
4. Send a normal follow-up message and inspect whether the fresh-session intent is applied to that real turn.

## Screenshots (optional)

Not applicable; this is a server-side channel routing issue.

## Additional context (optional)

The shared Router already owns `/new` classification and the in-process pending-fresh state used by its debounce window. The fix should extend that existing ownership, preserve deduplication, and avoid changing Slack or Lark adapter contracts.

yyclaw:
## What does this PR do?

Fixes the shared channel Router so a bare `/new` command records fresh-session intent without appending an empty user turn or scheduling an empty agent run. The next real message in the same chat session consumes that intent exactly once and starts a fresh agent session.

The root cause was that shared `/new` parsing stripped the command to an empty body but still sent the message through the normal append, typing, and run-scheduling path. The fix keeps classification and pending state inside the shared Router, so existing Slack and Feishu/Lark adapters inherit the same behavior without platform-specific changes.

The pending intent reuses the Router's existing in-process fresh/debounce ownership. It is bounded to 24 hours; like the existing debounce state, it is intentionally not durable across a hard process restart.

## Related Issue

Closes #6336

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Classify a bare `/new` from the original `CommandText` inside the shared Router, including Feishu/Lark's pre-stripped and enriched path.
- Complete installation, identity, session, and dedup handling, then skip empty message persistence, typing, media handling, and task scheduling.
- Apply the pending fresh intent to the next successfully queued real message exactly once, retaining it across enqueue failures.
- Bound abandoned in-memory fresh intents with a 24-hour TTL.
- Keep the new control state private to the engine; no Slack, Lark, or future adapter contract is added.

## How to Test

1. Send a bare `/new` through Slack or Feishu/Lark and confirm no empty chat message, typing indicator, or agent task is created.
2. Send a normal follow-up in the same conversation and confirm its queued task has `force_fresh_session=true`.
3. Send another normal message and confirm fresh intent is not applied again.
4. Make the first follow-up enqueue fail, retry with another message, and confirm the fresh intent survives until a run is successfully queued.

Automated verification:

```text
go test ./internal/integrations/channel/... ./internal/integrations/lark/... ./internal/integrations/slack/...
```

## Checklist

The screenshot, documentation, landing-copy, and Chinese-copy checklist items are not applicable because this is a server-only behavior fix with no user-facing copy or UI changes.

## AI Disclosure

**Prompt / approach:**
Reviewed the change from the shared channel Router outward: command classification, dedup, session resolution, append/run/typing side effects, debounce state, enqueue failure, and both existing Slack and Feishu/Lark adapter mappings. Removed adapter-facing fields that were not required for the fix and verified the affected Go packages.

## Screenshots (optional)

Not applicable; no UI changes.

### DIFF
--- a/server/internal/integrations/channel/engine/resolvers.go
+++ b/server/internal/integrations/channel/engine/resolvers.go
@@ -58,6 +58,9 @@ type Result struct {
 	IssueNumber     int32
 	IssueIdentifier string
 	IssueTitle      string
+	// runScheduled reports whether this ingest scheduled a normal chat run.
+	// It is Router-internal state: repliers must continue to use Outcome.
+	runScheduled bool
 }
 
 // ResolvedInstallation is the channel-agnostic installation context the Router
--- a/server/internal/integrations/channel/engine/router.go
+++ b/server/internal/integrations/channel/engine/router.go
@@ -56,7 +56,7 @@ type Router struct {
 	logger *slog.Logger
 
 	pendingFreshMu sync.Mutex
-	pendingFresh   map[string]bool
+	pendingFresh   map[string]time.Time
 }
 
 // Config tunes the Router. Zero values default.
@@ -108,7 +108,7 @@ func NewRouter(issues IssueCreator, tasks TaskEnqueuer, reader SessionReader, cf
 		mediaCancel:  mediaCancel,
 		mediaSem:     make(chan struct{}, cfg.MediaConcurrency),
 		logger:       cfg.Logger,
-		pendingFresh: make(map[string]bool),
+		pendingFresh: make(map[string]time.Time),
 		mediaQueues:  make(map[string]*mediaQueueEntry),
 	}
 }
@@ -187,12 +187,16 @@ func (r *Router) Handle(ctx context.Context, msg channel.InboundMessage) error {
 	}
 
 	// /new is a channel-wide product command, not an adapter capability. Parse
-	// it here so every adapter that delivers it as message text gets the same
-	// fresh-session behavior. Feishu already strips /new before its context
-	// enricher runs and sets ForceFresh; the guard preserves that enriched body.
-	if !msg.ForceFresh {
-		if body, ok := ParseFreshSessionCommand(msg.CommandText); ok {
-			msg.ForceFresh = true
+	// the original command source here even when an adapter already set
+	// ForceFresh, so bare-command classification stays identical across
+	// platforms. Only rewrite Text when the adapter has not already stripped
+	// the directive; Feishu enriches that stripped body before it reaches us.
+	bareFresh := false
+	if body, ok := ParseFreshSessionCommand(msg.CommandText); ok {
+		adapterAlreadyStripped := msg.ForceFresh
+		msg.ForceFresh = true
+		bareFresh = body == ""
+		if !adapterAlreadyStripped {
 			msg.Text = body
 		}
 	}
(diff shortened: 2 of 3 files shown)


Bohan-J:
Final review done — this is in great shape. Thanks a lot for the fix, @yyclaw! 🎉

What I especially appreciated: you solved this at the shared Router rather than patching each adapter. Classifying from the original `CommandText` means Slack and Feishu/Lark inherit identical bare-command semantics for free, and future adapters won't have to re-derive it.

What I verified on `758cc675`:

- **Cross-platform classification is correct.** Slack arrives with `CommandText` as the raw normalized text and `ForceFresh=false`; Feishu arrives with `ForceFresh=true` and an enriched `Text`, but the enricher deliberately leaves `CommandBody` un-stripped, so the `adapterAlreadyStripped` guard keeps the enriched body intact while still classifying the bare command. The `CommandText == ""` fallback covers adapters that don't populate it.
- **Dropping the old `!msg.ForceFresh` guard is safe.** Parsing now reads the un-enriched command source instead of the rewritten body, so an enriched `<recent_context>` block containing a historical `/new` can't be misread — `TestRouter_AdapterFreshBodyIsNotParsedAgain` pins that.
- **Early-return placement is right.** It lands after the dedup claim, identity, and `EnsureSession` but before `AppendMessage`, and returns `finalizeMark` — so nothing is written and a redelivered event can't record the intent twice.
- **No typing-indicator regression.** `runScheduled` is the new gate, and the only two `OutcomeIngested` returns are the bare-command path (false) and the normal path (set immediately after `scheduleRun`). No existing ingest path loses its indicator.
- **The bare command is genuinely silent.** Both `OutboundReplier` and `LarkOutcomeReplier` gate `OutcomeIngested` on `IssueID.Valid`, so no stray confirmation.
- **Exactly-once consumption holds under debounce.** Both `scheduleRun` branches route through `takePendingFresh`, which reads and deletes under the mutex — a multi-message debounce window can't double-consume.
- **Intent survives failure.** Re-marking on both the session-reload and enqueue error paths is the right call: `ForceFresh` belongs to the first *successfully queued* run, not the first attempt.

Local verification: `go test -race` green across the engine, Slack, and Lark packages (39 Router / fresh-command tests, plus the Lark enricher suite); `gofmt` and `go vet` clean. All 9 CI checks pass. The DB-integration failures I saw locally are a stale test-database schema and reproduce on `main` — unrelated to this change.

One non-blocking nit, which we're consciously accepting rather than asking you to change: the TTL reclaim in `markPendingFresh` is lazy, so the last batch of expired entries can linger until the next `/new` arrives in the process. In practice the sweep is global and each entry is tiny, so the map stays bounded — and the in-memory, restart-loses-it boundary is documented in your PR description. Happy to leave it as is.

Merging now. Thanks again for the clear reproduction, the failure-path test coverage, and for documenting the restart boundary up front — that made this an easy review. 🙏

PR #6344, opened 2026-08-04
LOOP-771: guard inherited provider homes in execenv

MrAladdin:
## Summary

- Keeps one shared execenv containment guard for inherited provider config paths: resolve the path, fail closed when `WorkspacesRoot` is empty, and reject paths that resolve under the daemon workspaces root, including symlink, case-only, and `/var` alias variants.
- Splits provider/config content from path containment:
  - Codex now hard-fails only when the shared source had `model_provider` and the generated per-task config lost or changed it during seeding.
  - Codex shared homes without `model_provider` are allowed to use Codex's native default provider/environment credentials; if Codex later fails, the task-visible error includes a config note.
  - Hermes inherited `HERMES_HOME` is guarded for path provenance only; Hermes-native provider/default behavior is left to Hermes.
  - OpenClaw inherited active config paths are guarded for containment only; provider/config usability stays delegated to existing CLI reads.
- Wires containment through current inherited-home/config resolvers:
  - Codex: inherited `CODEX_HOME`, shared home seeding, user skill seeding, and session store prune path resolution.
  - Hermes: process-inherited `HERMES_HOME`; explicit task/custom env home still bypasses the inherited-env guard.
  - OpenClaw: inherited active config path before writing the task wrapper.
- Updates daemon launch docs to scrub only provider home selectors and assert required vars such as `MLAMP_API_KEY` remain present.

## Notes

- GC/prune now depends only on containment, not provider usability; it can still resolve a providerless shared home for cleanup.
- Reuse returns nil only for refresh/prep failures that require a fresh `Prepare`; providerless default Codex config no longer blocks reuse.
- LOOP-771 extra view (a): kept the early WorkspacesRoot containment guard because a task-scoped home under the daemon workspaces root is provably reverse pollution before provider probing.
- LOOP-771 extra view (b): `syncCopiedFile` still reports a missing copied source to the caller; when the source had no `model_provider`, the daemon logs a warning and lets Codex's native default-provider behavior run.
- MUL-5578 / MUL-4424 isolation is preserved: this guards daemon-process inherited resolver inputs. The daemon still explicitly exports per-task `CODEX_HOME`, Hermes overlay `HERMES_HOME`, and per-task `OPENCLAW_CONFIG_PATH` to child tasks after execenv preparation.

## Tests

- `go test ./internal/daemon/execenv -count=1`
- `go test ./pkg/agent -run 'TestCodexProviderlessDefaultFailureAnnotation|TestCodex' -count=1`
- `go test ./internal/daemon/... -count=1`
- `git diff --check`

### DIFF
--- a/apps/docs/content/docs/daemon-runtimes.mdx
+++ b/apps/docs/content/docs/daemon-runtimes.mdx
@@ -48,6 +48,17 @@ On startup, the daemon detects supported AI coding tools on `PATH` and registers
 
 The daemon needs at least one built-in supported AI coding tool detected before it will start. Install methods and executable names are in [Install AI coding tools](/install-agent-runtime).
 
+### Launch environment hygiene
+
+If you start the daemon from a shell or service runbook, clear only provider home selectors that can pin every task to stale task-local state:
+
+```bash
+test -n "${MLAMP_API_KEY:-}" || { echo "MLAMP_API_KEY is missing"; exit 1; }
+env -u CODEX_HOME -u HERMES_HOME -u OPENCLAW_HOME -u OPENCLAW_CONFIG_PATH multica daemon start
+```
+
+Do not start from a broadly scrubbed environment unless you explicitly re-inject required credentials and gateway variables. Scrubbing `*_HOME` should not remove `MLAMP_API_KEY` or other deployment-specific variables the detected tools need.
+
 ## Dispatch and online status
 
 Once registered, a runtime keeps a persistent connection. When a new task enters the queue, the server notifies the matching daemon; the daemon also polls periodically as a backstop after connection interruptions. So when a runtime is online with spare capacity, tasks usually start immediately.
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -4994,7 +4994,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 	var hermesEnv map[string]string
 	if provider == "hermes" {
 		sel := agent.ParseHermesProfileArgs(agentCustomArgs)
-		res := execenv.ResolveHermesProfile(agentEnvOverrides["HERMES_HOME"], sel.Name, sel.Found, sel.Inline)
+		res := execenv.ResolveHermesProfileWithWorkspacesRoot(agentEnvOverrides["HERMES_HOME"], sel.Name, sel.Found, sel.Inline, d.cfg.WorkspacesRoot)
 		if res.Err != nil {
 			return TaskResult{}, fmt.Errorf("resolve hermes profile: %w", res.Err)
 		}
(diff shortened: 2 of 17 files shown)


multica-eve (changes requested):
Requesting changes for two blockers:

1. `server/internal/daemon/execenv/provider_config_guard.go:16-21` only applies `filepath.Abs`, and `pathIsUnder` (`:31-44`) performs a lexical containment check. Neither resolves symlinks. An inherited Codex/Hermes home or OpenClaw config path outside `WorkspacesRoot` can therefore be a symlink to an old task path inside the root; the guard accepts the external link name, while subsequent reads, copies, and `$include` resolution follow the link and inherit the old task state. Please compare consistently resolved real paths for both the provider path and `WorkspacesRoot`, fail closed when an existing target cannot be resolved safely, and add regression coverage for all three providers using an outside symlink to an inside target.

2. The content-based usability checks reject legitimate personal homes without proving that a path is untrusted. `codex_home.go:335-357` accepts a missing config but rejects an existing config containing only sandbox/features settings; `hermes_home.go:260-276` similarly accepts no config but rejects a config containing only skills or other non-provider settings. A legitimate setup may rely on the default provider or environment credentials, while a stale task config may still contain provider keys. Please scope hard failure to verifiable path provenance, or explicitly establish this as a compatibility change and add positive tests plus migration guidance for valid custom homes.

The shared guard and its three call paths are otherwise well organized. The targeted daemon/execenv tests and the `ByteIdentical` invariant pass, but the new tests cover only direct paths and therefore do not catch the symlink bypass.

MrAladdin:
Addressed the two requested-change blockers in the latest commits.

1. Symlink / non-lexical containment:
   - `resolveAndGuardProviderConfigPath` now fails closed on empty `WorkspacesRoot`.
   - `pathIsUnder` compares best-effort resolved paths and falls back to `os.SameFile` ancestor walking, which catches outside symlinks to workspace paths, `/var` aliasing, and case-only path variants on case-insensitive filesystems.
   - Regression coverage is in `TestInheritedProviderPathGuardHandlesNonLexicalWorkspacePaths`; inherited path coverage still spans Codex, Hermes, and OpenClaw via `TestInheritedProviderEnvUnderWorkspacesRootFailsClosed`.

2. Content checks / compatibility:
   - Removed content-based hard-fail for Hermes inherited homes; that path is now provenance/containment-only and leaves provider/default behavior to Hermes.
   - Removed the global Codex "must have model_provider" launch gate.
   - Codex now hard-fails only when the shared seed source had `model_provider` and the generated per-task config lost or changed it during seeding. If the source itself has no `model_provider`, Prepare/Reuse succeeds so Codex can use its native default provider/environment credentials.
   - Added positive tests for providerless inherited `CODEX_HOME`, providerless default `~/.codex`, providerless Hermes home, and providerless Codex Reuse. Added `TestCodexProviderModelProviderLostDuringSeedingFailsClosed` for the hard-fail case that this issue needs to keep.
   - Added a task-visible Codex failure annotation when the task config has no `model_provider`, so if the native default-provider path later fails, the failure text carries that fact instead of requiring daemon-log archaeology.

Validation:
- `go test ./internal/daemon/execenv -count=1`
- `go test ./pkg/agent -run 'TestCodexProviderlessDefaultFailureAnnotation|TestCodex' -count=1`
- `go test ./internal/daemon/... -count=1`
- `git diff --check`

PR #6358, opened 2026-08-04
MUL-5666: cap picker pill width and standardize clear rows

NevilleQingNY:
## Why

Two problems, same root: the property pills had no shared contract.

**Clearing a field was in a different place per picker.** Assignee / project / stage led with an empty row; priority put "No priority" last; the date and custom-property popovers hid a Clear button in a footer; and the project pill carried a hover-only overlay × on top of its row.

That × was also broken. It was an absolutely-positioned sibling of the trigger, and only the picker's *own default* trigger reserved right padding for it — so every caller passing `triggerRender` (create dialog, table cell, autopilot card) had the × painted directly over the project name. Three of five callers got the implicit padding contract wrong.

**Nothing capped a pill's width.** Pills carry user-generated text and the create toolbars are `flex-wrap` rows, so one long project title stretched its pill across the whole line and pushed every sibling to the next one.

## What changed

**The empty value is always the popover's first row.**

| Picker | Before | After |
|---|---|---|
| Project | overlay × + row hidden while searching | row only, always first |
| Assignee | row hidden while searching | always first |
| Priority | "No priority" last | first (see below) |
| Start / due date | footer Clear button | row above the calendar |
| Custom property | footer Clear button | first row |
| Stage | first row, no icon | first row, icon column aligned |

Priority needed the constant split: `PRIORITY_ORDER` doubles as the **sort rank** (`sortIssues` uses the array index as the comparison weight), so it keeps `none` last. The new `PRIORITY_DISPLAY_ORDER` leads with it and drives the picker, the actions menu, and the filter dropdown.

Copy follows position — rows read as nouns (No project / Unassigned / No due date / No value), retiring the verb-form `clear_action` and `clear_date` keys across all four locales.

**Pill width is capped once, at the shared `PillButton`:** `min-w-0 max-w-56 overflow-hidden`. The `min-w-0` + `overflow-hidden` pair is what makes the inner `truncate` spans work at all — a flex item only drops `min-width: auto` once its overflow is hidden, which is why those spans never truncated before. Two call sites needed a follow-up the cap exposed: the create-project lead pill had no `truncate`, and `LabelChip` had no `min-w-0` (so the label pill clipped a chip mid-shape instead of shrinking it).

## Notes for review

- **`disabled` on ProjectPicker is now enforced in one place.** MUL-5150 hardened it because a keyboard user could Tab to the inline × and detach the project mid chat-send. With clearing inside the popover, `disabled` only has to keep the popover shut — covered by a test that also asserts a controlled `open` can't force it open.
- **Clicking an empty row when the field is already empty** is now reachable and fires a no-op write. This matches what every list-backed picker already did, and the server documents `properties - key` as a no-op for the absent-key case, so there is no error path.
- **Mobile is deliberately untouched** (iOS isn't being iterated on right now). It keeps severity-descending priority order and still hides the empty row while searching. Both are ordering/visibility only — the options themselves exist on both clients, so the semantic parity rules in `apps/mobile/CLAUDE.md` (enums, counts, permissions) still hold.

## Open question, not addressed here

Multi-select fields are now inconsistent with each other: custom `multi_select` has an empty row that clears **all** selected options in one click, while labels has no such entry and clears one chip at a time. Either labels gains one or the custom one goes — I lean toward removing it, since a clear-all shouldn't be easier to hit than unchecking. Happy to follow up either way.

## Verification

- `packages/views` — 301 files / 3496 tests pass
- `packages/core` — 114 files / 1234 tests pass
- `pnpm typecheck` — clean
- `pnpm lint` — 0 errors (16 pre-existing warnings in untouched files)

Not run: Playwright e2e, Go tests (no backend change), mobile checks (no mobile change).

### DIFF
--- a/packages/core/issues/config/index.ts
+++ b/packages/core/issues/config/index.ts
@@ -1,2 +1,2 @@
 export { STATUS_ORDER, ALL_STATUSES, STATUS_CONFIG } from "./status";
-export { PRIORITY_ORDER, PRIORITY_CONFIG } from "./priority";
+export { PRIORITY_ORDER, PRIORITY_DISPLAY_ORDER, PRIORITY_CONFIG } from "./priority";
--- a/packages/core/issues/config/priority.ts
+++ b/packages/core/issues/config/priority.ts
@@ -1,11 +1,32 @@
 import type { IssuePriority } from "../../types";
 
+/**
+ * Severity order, high to low. This is the **sort rank** — `sortIssues` turns
+ * the array index into the comparison weight — so "none" must stay last.
+ * Never reorder this to change how a menu reads; use
+ * {@link PRIORITY_DISPLAY_ORDER} for that.
+ */
 export const PRIORITY_ORDER: IssuePriority[] = [
   "urgent",
   "high",
   "medium",
   "low",
   "none",
+];
+
+/**
+ * Order every user-facing priority list renders in: the empty value first,
+ * then severity descending. Leading with "No priority" follows the convention
+ * every other picker in the app uses — the first row is always the empty
+ * value (unassigned / no project / no stage), so the eye finds "clear this
+ * field" in the same place regardless of which pill was opened.
+ */
+export const PRIORITY_DISPLAY_ORDER: IssuePriority[] = [
+  "none",
+  "urgent",
+  "high",
+  "medium",
+  "low",
 ];
 
 export const PRIORITY_CONFIG: Record<
--- a/packages/views/chat/components/chat-input.tsx
+++ b/packages/views/chat/components/chat-input.tsx
@@ -607,7 +607,7 @@ export function ChatInput({
                     disabled={!projectSelectionEnabled}
                     aria-label={t(($) => $.input.change_project_context)}
                     title={t(($) => $.input.change_project_context)}
-                    className="flex h-6 max-w-56 items-center gap-1.5 rounded-full border border-surface-border bg-surface-raised px-2 pr-7 text-caption font-medium text-foreground transition-colors hover:bg-accent/60"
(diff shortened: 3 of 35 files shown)


NevilleQingNY:
Blocker fixed in 3d36c715e.

**Confirmed the diagnosis, and found a second path to the same outcome.** Besides the highlight reset at `property-picker.tsx:184`, the empty row survives every filter — so a query with *no* matches left it as the sole item and the `items.length === 1` auto-select (`property-picker.tsx:148`) fired on it. Typing gibberish and pressing Enter also cleared the field.

**Approach: mark the row rather than move it.** Taking it out of the `listRef` navigation set would have fixed Enter but made the row unreachable by arrow key, which is a real loss — it is a legitimate value, not just an escape hatch. So `PickerItem` gained an `emptyValue` prop that stamps `data-picker-empty`, and `PropertyPicker` now:

- resolves the post-keystroke highlight to the first row *without* that attribute — `-1` when nothing else matched, which leaves Enter inert instead of clearing;
- refuses to auto-select a lone empty row.

The index can't be computed in the input's `onChange` (the filtered list hasn't rendered yet), so the keystroke raises a ref flag that an effect resolves against the fresh DOM. Guarding on that flag is what keeps arrow keys free to walk back onto the empty row — only typing moves the highlight off it.

Marked on rows that write null/undefined: project, assignee, stage, custom property. Deliberately **not** on issue priority's `none`, which is a real enum member rather than a cleared field, and whose picker has no search box.

**Tests** — all five fail on the previous head, verified by stashing only the `property-picker.tsx` fix:

- project: Enter-on-match commits the match, Enter-on-no-match is inert, arrow-up still reaches the empty row
- assignee: the first two (this picker had no test file; added `assignee-picker.keyboard.test.tsx`)

The arrow-key test needed an `HTMLElement.prototype.scrollIntoView` stub — jsdom doesn't define it, which is why no existing test covered that path.

**Verification:** `packages/views` 302 files / 3501 tests pass; `pnpm typecheck` clean; `pnpm lint` 0 errors.

PR #6388, opened 2026-08-04
fix(channel): attach /issue media to created issues

### LINKED ISSUE #6387: [Bug]: Channel /issue media is not attached to the created issue
### Deployment type

All deployment types (Official App and self-host)

### What happened?

When a message-based channel `/issue` command includes image or file media, the shared channel engine creates the Issue synchronously but the detached media resolver later binds the media to the chat message. The created Issue therefore has no attachment. If an agent is assigned during creation, its task can also start before media resolution finishes, so even a later attachment would be unavailable to the initial run.

Expected behavior: media supplied with `/issue` is owned by the created Issue, the automatically assigned task waits until media binding settles (with a bounded fallback), and already-open Issue views refresh when the attachment arrives.

This is a shared channel-engine gap observable through the existing Lark integration. It does not require QuickCreate or assume a future channel implementation.

### Steps to reproduce

1. Configure a Lark channel connected to a workspace.
2. Send an image or file together with `/issue <title>`.
3. Open the Issue returned by the command.
4. Observe that the Issue has no attachment because the media remains associated with the chat flow.
5. If the Issue is assigned to an agent, observe that the task may begin before asynchronous media resolution completes.

### Screenshots (optional)

N/A — this is an asynchronous channel/backend ownership issue.

The shared upstream architecture already has synchronous `/issue` creation and asynchronous channel media resolution. The missing piece is to carry the created Issue identity into that resolver, bind the resulting attachment transactionally to the Issue, gate the assigned task durably, and emit a precise Issue-attachment realtime event.

yyclaw:
## What does this PR do?

Fixes the shared channel `/issue` flow so image and file media attach to the synchronously created Issue instead of remaining in the chat-message path.

The final implementation also incorporates the reviewer follow-up against current `main`:

1. A synchronous `/issue` remains terminal and never schedules a duplicate chat run.
2. Active duplicates remain terminal product outcomes while their media is still finalized against the existing issue.
3. The assigned-agent task is persisted as a narrowly marked deferred row before `issue:created` is published, promoted when media settles, and protected by a database uniqueness invariant with a bounded `fire_at` fallback.
4. Issue deletion collects attachment URLs under the same row-lock transaction used to serialize against media binding, preventing object-storage leaks.
5. Web/core and mobile invalidate the issue attachment cache on the live event; mobile also invalidates it after reconnect.

## Related Issue

Closes #6387

## Type of Change

- [x] Bug fix
- [x] Tests
- [x] Database migration
- [ ] Feature
- [ ] UI change

## Changes Made

- Propagate the synchronously created or duplicate Issue ID through shared channel media resolution and adapter contracts.
- Bind resolved media as Issue-owned attachments with workspace validation and transactional intent handling.
- Preserve terminal `/issue` behavior from #6398 and duplicate media finalization.
- Defer and promote assigned-agent tasks around media binding, with a `channel_issue_media_pending` marker and a concurrent partial unique index.
- Serialize single and batch issue deletion with channel media binding before collecting attachment URLs.
- Publish `issue_attachments:changed` and invalidate the affected query in core and mobile, including mobile reconnect recovery.

No DingTalk-specific behavior, QuickCreate behavior, or UI surface is added.

## Verification

- `make sqlc`
- Migration up and full down/up cycle on an isolated worktree database
- Full tests for affected engine, Lark, Slack, service, and handler Go packages
- Focused Go race tests for terminal issue commands and both database concurrency protocols
- `go vet` for affected Go packages
- Mobile tests and typecheck
- Core tests and monorepo typecheck
- `git range-diff` across both rebases and `git diff --check`

## AI Disclosure

OpenAI Codex was used to trace the shared channel and asynchronous media paths, implement the fixes, add deterministic regression coverage, and verify the final semantic rebase.

### DIFF
--- a/apps/mobile/CLAUDE.md
+++ b/apps/mobile/CLAUDE.md
@@ -228,7 +228,7 @@ Each hook registers a single `ws.onReconnect(cb)` that invalidates **only the qu
 |---|---|
 | `useInboxRealtime` | `inboxKeys.list(wsId)` |
 | `useMyIssuesRealtime` | `issueKeys.myAll(wsId)` |
-| `useIssueRealtime(id)` | `issueKeys.detail(wsId, id)` + `issueKeys.timeline(wsId, id)` |
+| `useIssueRealtime(id)` | detail + timeline + attachments + active tasks + task history for that issue |
 
 No global "invalidate everything on reconnect" sweep. The fanout would be every screen the user has ever visited in this session refetching simultaneously — wasteful on cellular and prone to rate-limiting the server in low-signal areas where reconnects happen frequently.
 
--- a/apps/mobile/data/realtime/issue-ws-updaters.test.ts
+++ b/apps/mobile/data/realtime/issue-ws-updaters.test.ts
@@ -1,0 +1,24 @@
+import { QueryClient } from "@tanstack/react-query";
+import { describe, expect, it, vi } from "vitest";
+
+import { issueKeys } from "@/data/queries/issue-keys";
+import { invalidateIssueAfterReconnect } from "./issue-ws-updaters";
+
+describe("invalidateIssueAfterReconnect", () => {
+  it("invalidates attachments together with the issue and task caches", () => {
+    const qc = new QueryClient();
+    const invalidate = vi.spyOn(qc, "invalidateQueries");
+    const wsId = "workspace-1";
+    const issueId = "issue-1";
+
+    invalidateIssueAfterReconnect(qc, wsId, issueId);
+
+    expect(invalidate.mock.calls.map(([filters]) => filters?.queryKey)).toEqual([
+      issueKeys.detail(wsId, issueId),
+      issueKeys.timeline(wsId, issueId),
+      issueKeys.attachments(wsId, issueId),
+      issueKeys.activeTasks(wsId, issueId),
+      issueKeys.tasks(wsId, issueId),
+    ]);
+  });
+});
--- a/apps/mobile/data/realtime/issue-ws-updaters.ts
+++ b/apps/mobile/data/realtime/issue-ws-updaters.ts
@@ -62,6 +62,18 @@ export function clearIssueDetail(
   qc.removeQueries({ queryKey: issueKeys.timeline(wsId, issueId) });
 }
 
+export function invalidateIssueAfterReconnect(
+  qc: QueryClient,
+  wsId: string,
+  issueId: string,
+) {
+  qc.invalidateQueries({ queryKey: issueKeys.detail(wsId, issueId) });
+  qc.invalidateQueries({ queryKey: issueKeys.timeline(wsId, issueId) });
+  qc.invalidateQueries({ queryKey: issueKeys.attachments(wsId, issueId) });
+  qc.invalidateQueries({ queryKey: issueKeys.activeTasks(wsId, issueId) });
+  qc.invalidateQueries({ queryKey: issueKeys.tasks(wsId, issueId) });
+}
+
 // =====================================================
 // Issue timeline (flat TimelineEntry[], ASC oldest-first)
 // =====================================================
--- a/apps/mobile/data/realtime/use-issue-realtime.ts
+++ b/apps/mobile/data/realtime/use-issue-realtime.ts
@@ -4,6 +4,7 @@
  *
  * Handles:
  *   - issue:updated / issue:deleted / issue_labels:changed → detail cache
+ *   - issue_attachments:changed → attachment cache
  *   - comment:created / comment:updated / comment:deleted → timeline
  *   - activity:created → timeline
  *   - reaction:added / reaction:removed → comment reactions on timeline
@@ -40,6 +41,7 @@ import {
   appendTimelineEntry,
   clearIssueDetail,
   commentToTimelineEntry,
+  invalidateIssueAfterReconnect,
   patchIssueDetail,
   patchIssueLabels,
   patchMyIssuesList,
@@ -112,6 +114,12 @@ export function useIssueRealtime(
         ws.on("issue_labels:changed", (payload) => {
           if (payload.issue_id !== issueId) return;
           patchIssueLabels(qc, wsId, issueId, payload.labels);
+        }),
+        ws.on("issue_attachments:changed", (payload) => {
+          if (payload.issue_id !== issueId) return;
+          qc.invalidateQueries({
+            queryKey: issueKeys.attachments(wsId, issueId),
+          });
         }),
 
         // ----- Comments / activity -----
@@ -225,8 +233,7 @@ export function useIssueRealtime(
 
         // ----- Reconnect -----
         ws.onReconnect(() => {
-          invalidateThisIssue();
-          invalidateTaskQueries();
+          invalidateIssueAfterReconnect(qc, wsId, issueId);
         }),
       ];
     },
--- a/packages/core/realtime/use-realtime-sync-ws-instance.test.tsx
+++ b/packages/core/realtime/use-realtime-sync-ws-instance.test.tsx
@@ -218,6 +218,23 @@ describe("useRealtimeSync — ws instance change", () => {
     expect(calls).toContainEqual(chatKeys.messagesPageAll());
     expect(calls).toContainEqual(chatKeys.pendingTaskAll());
   });
+
+  it("invalidates one issue attachment cache after detached channel media binds", () => {
+    const ws = createMockWs();
+    renderHook(() => useRealtimeSync(ws, stores), {
+      wrapper: createWrapper(qc),
+    });
+    const attachmentChanged = vi
+      .mocked(ws.on)
+      .mock.calls.find(([event]) => event === "issue_attachments:changed")?.[1];
+    expect(attachmentChanged).toBeDefined();
+
+    (attachmentChanged as (payload: unknown) => void)({ issue_id: "issue-1" });
+
+    expect(invalidateSpy).toHaveBeenCalledWith({
+      queryKey: issueKeys.attachments("issue-1"),
+    });
+  });
 });
 
 describe("useRealtimeSync — queued chat promotion", () => {
--- a/packages/core/realtime/use-realtime-sync.ts
+++ b/packages/core/realtime/use-realtime-sync.ts
@@ -68,6 +68,7 @@ import type {
   IssueUpdatedPayload,
   IssueCreatedPayload,
   IssueDeletedPayload,
+  IssueAttachmentsChangedPayload,
   IssueLabelsChangedPayload,
   IssueMetadataChangedPayload,
   IssuePropertiesChangedPayload,
@@ -878,7 +879,7 @@ export function useRealtimeSync(
     // Event types handled by specific handlers below -- skip generic refresh
     const specificEvents = new Set([
       "workspace:updated",
-      "issue:updated", "issue:created", "issue:deleted", "issue_labels:changed", "issue_metadata:changed", "issue_properties:changed", "property:created", "property:updated", "inbox:new",
+      "issue:updated", "issue:created", "issue:deleted", "issue_attachments:changed", "issue_labels:changed", "issue_metadata:changed", "issue_properties:changed", "property:created", "property:updated", "inbox:new",
       "comment:created", "comment:updated", "comment:deleted",
       "comment:resolved", "comment:unresolved",
       "activity:created",
@@ -952,6 +953,12 @@ export function useRealtimeSync(
       if (!issue_id) return;
       const wsId = getCurrentWsId();
       if (wsId) onIssueLabelsChanged(qc, wsId, issue_id, labels ?? []);
+    });
+
+    const unsubIssueAttachmentsChanged = ws.on("issue_attachments:changed", (p) => {
+      const { issue_id } = p as IssueAttachmentsChangedPayload;
+      if (!issue_id) return;
+      qc.invalidateQueries({ queryKey: issueKeys.attachments(issue_id) });
     });
 
     const unsubIssueMetadataChanged = ws.on("issue_metadata:changed", (p) => {
@@ -1503,6 +1510,7 @@ export function useRealtimeSync(
       unsubIssueUpdated();
       unsubIssueCreated();
       unsubIssueDeleted();
+      unsubIssueAttachmentsChanged();
       unsubIssueLabelsChanged();
       unsubIssueMetadataChanged();
       unsubIssuePropertiesChanged();
--- a/packages/core/types/events.ts
+++ b/packages/core/types/events.ts
@@ -12,6 +12,7 @@ import type { Label } from "./label";
 export type WSEventType =
   | "issue:created"
   | "issue:updated"
+  | "issue_attachments:changed"
   | "issue:deleted"
   | "comment:created"
   | "comment:updated"
@@ -123,6 +124,10 @@ export interface IssueDeletedPayload {
 export interface IssueLabelsChangedPayload {
   issue_id: string;
   labels: Label[];
+}
+
+export interface IssueAttachmentsChangedPayload {
+  issue_id: string;
 }
 
 export interface IssueMetadataChangedPayload {
@@ -501,6 +506,7 @@ export interface WSEventPayloadMap {
   "issue:created": IssueCreatedPayload;
   "issue:updated": IssueUpdatedPayload;
   "issue:deleted": IssueDeletedPayload;
+  "issue_attachments:changed": IssueAttachmentsChangedPayload;
   "issue_labels:changed": IssueLabelsChangedPayload;
   "issue_properties:changed": IssuePropertiesChangedPayload;
   "property:created": PropertyChangedPayload;
--- a/server/cmd/migrate/main.go
+++ b/server/cmd/migrate/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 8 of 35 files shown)


Bohan-J (changes requested):
Thanks for taking this on — the product direction is right. `/issue` media is user-supplied input for the issue, so the attachments have to land on the issue and any assigned agent run has to wait for them, with a bounded fallback instead of an unbounded wait. The overall shape (issue-scoped binding under a lock, a persisted deferred task, `fire_at` sweep as the crash fallback, targeted `issue_attachments:changed` invalidation) is the right architecture.

Requesting changes on four items, all reviewed against head `5af0e71`.

### 1. Rebase onto `main` and preserve the `/issue`-is-terminal semantics

The PR is currently `CONFLICTING` / `DIRTY`, with the conflict centered on `server/internal/integrations/channel/engine/router.go` and its tests.

This is not just a textual conflict. On this branch, `router.go` creates the issue and then falls through to `r.scheduleRun(...)` at `router.go:439` unconditionally. On `main`, that path has since changed (#6398): a synchronous `/issue` command is terminal and deliberately does **not** schedule a chat run, because `IssueService.Create` already enqueues the assigned agent's issue task — scheduling a chat run as well makes the agent execute the same `/issue` input a second time. `main` also added an `ErrActiveDuplicate` branch that returns a product outcome (not an infrastructure error) and still calls `enqueueMedia` so duplicate-path media gets reclaimed.

The conflict resolution needs to keep all four properties simultaneously:

- issue-scoped attachment binding (this PR),
- the deferred assigned task + early promotion (this PR),
- terminal `/issue` — no chat run (`main`),
- duplicate → product outcome with media still finalized (`main`).

Resolving toward either side alone reintroduces double execution or leaves duplicate-path media pending.

### 2. "One pending task per (issue, agent)" is not enforced at the DB level for the deferred row

In `server/internal/service/issue.go`, `publishIssueCreated` runs at line 339 and the media-gated deferred task is created at line 341 by `maybeEnqueueOnAssign`. So `issue:created` is on the wire before the deferred row exists.

Meanwhile the unique index from `server/migrations/037_fix_pending_task_unique_index.up.sql` is:

```sql
CREATE UNIQUE INDEX idx_one_pending_task_per_issue_agent
    ON agent_task_queue (issue_id, agent_id)
    WHERE status IN ('queued', 'dispatched');
```

It does not cover `status = 'deferred'`. The interleaving:

1. `issue:created` is published.
2. A client (or an agent) reacts immediately and comments → the comment path inserts a `queued` task for the same `(issue_id, agent_id)`. Nothing blocks this, since the deferred row does not exist yet.
3. `maybeEnqueueOnAssign` inserts the `deferred` row — also unblocked, since `deferred` is outside the partial index.
4. Media binds, and `PromoteDeferredChannelIssueTask` (`server/pkg/db/queries/agent.sql:349`) flips `deferred → queued`, which now collides with the unique index and errors.

Worse than the failed promotion: the task created at step 2 is an ordinary `queued` row, so it can be claimed and run *before* the attachments are bound — exactly the bug this PR is fixing, just reached by a different path.

The existing race coverage in `server/internal/service/task_channel_media_race_test.go` targets the enqueue-vs-append seal window and post-settlement promotion, not this publish → assigned-task-create window. Please make the invariant enforceable (serialize the create, or widen the constraint to cover the media-pending deferred row) and add a deterministic test for this specific interleaving. If the index changes, the migration must use `CREATE UNIQUE INDEX CONCURRENTLY`.

### 3. Issue delete racing attachment bind can leak object storage files

`bindMediaRefs` (`server/internal/integrations/channel/engine/session.go:415`) takes a key-share lock on the issue via `LockIssueForChannelMediaBind` at `:433`, writes the attachment, and claims/clears the intent-ledger rows in the same transaction. That correctly rules out cross-workspace binds and binds to a deleted issue.

But `DeleteIssue` reads the URL list outside any transaction, before the delete:

```go
attachmentURLs, _ := h.Queries.ListAttachmentURLsByIssueOrComments(r.Context(), issue.ID)
err := h.Queries.DeleteIssue(r.Context(), ...)
h.deleteS3Objects(r.Context(), attachmentURLs)
```

Interleaving: delete reads the URL list (attachment not there yet) → bind commits the attachment row and clears its intent → `DELETE` acquires the lock and cascades the attachment row away → `deleteS3Objects` never sees that URL. Both the attachment row and the intent ledger entry are gone, so the reconciler has nothing left to reclaim, and the object is orphaned in storage.

Bind and delete need to agree on one concurrency protocol — collect URLs under the same lock/transaction that performs the delete, or leave a durable tombstone the reconciler can act on. Please add a DB test for this exact interleaving.

### 4. Mobile reconnect path misses the attachment cache

`apps/mobile/data/realtime/use-issue-realtime.ts` invalidates attachments precisely on the live event at `:117-122`, but the reconnect handler at `:234-237` calls only `invalidateThisIssue()` (detail + timeline, per `:72-75`) and `invalidateTaskQueries()`.

If the bind event fires while the app is backgrounded or the socket is down, the attachment cache is never invalidated and keeps showing the pre-bind list — which is the very case this PR exists to fix, since `/issue` media binds asynchronously seconds after creation. Please add `issueKeys.attachments(wsId, issueId)` to that reconnect handler, with a mobile realtime test.

---

Verification notes: the currently green backend/frontend/mobile checks are against a head that has not been merged with the latest `main`, so they need a re-run after the rebase. Findings above were read directly against `5af0e71` and `main`; items 1, 2 and 4 are confirmed from the code as written, and item 3 is a confirmed ordering gap between the two paths.

Happy to re-review once the rebase and these fixes are in.

yyclaw:
Addressed all four requested changes on head `b654daa5f`:

1. Rebased onto current `main` and preserved the terminal `/issue` semantics from #6398. Successful and duplicate outcomes no longer schedule a chat run; both still finalize issue-scoped media, and duplicate media does not attempt issue-task promotion.
2. Persisted the media-gated deferred assigned task before publishing `issue:created`, and replaced the pending-task partial unique index with a `CREATE UNIQUE INDEX CONCURRENTLY` definition that also covers the explicitly marked `channel_issue_media_pending` deferred row. Added deterministic coverage for the publish-to-task-create interleaving and the DB uniqueness invariant.
3. Added a shared delete protocol for single and batch issue deletion: lock the issue `FOR UPDATE`, collect attachment URLs, and delete in one transaction. This serializes with the binder `FOR KEY SHARE` lock. Added a DB lock-race test that verifies delete-first leaves the media intent durable for reconciliation.
4. Added issue attachment invalidation to the mobile reconnect path through a tested cache invalidation helper. The helper also keeps the existing detail, timeline, and task-query invalidations together.

Additional verification completed:

- affected engine, Lark, Slack, service, and handler Go package tests
- focused Go race tests and `go vet`
- migration full up/down/up cycle on an isolated worktree database
- mobile tests and typecheck
- core tests and monorepo typecheck
- `make sqlc`, `git range-diff`, and `git diff --check`

The PR description has been updated to reflect the migration and final scope. Please re-review when convenient.

Bohan-J (changes requested):
Re-reviewed at head `b654daa5`. All four items from the previous round are genuinely closed, and I verified each against the code rather than the summary:

- `/issue` is terminal on both the success and the active-duplicate path, and both still finish issue-scoped media (`router.go:421-457`).
- The media-gated deferred task is now created *before* `issue:created` is published (`service/issue.go:343-355`), and the unique predicate covers the explicitly-marked deferred row.
- Single and bulk delete both take `LockIssueForDelete` and collect attachment URLs inside that transaction (`handler/issue.go:3208-3214`), which serializes properly against the binder's `FOR KEY SHARE`.
- Mobile reconnect now invalidates attachments via `invalidateIssueAfterReconnect` (`issue-ws-updaters.ts:65-75`).

Nice work — that's a real improvement in the concurrency story.

One blocking item remains, and it's a deploy-safety one that CI can't catch.

### The 257 → 258 pair removes an existing data invariant before re-establishing it

`257_drop_pending_issue_agent_v1.up.sql` is `DROP INDEX CONCURRENTLY IF EXISTS idx_one_pending_task_per_issue_agent`, and the widened definition is only created in `258_...up.sql` — under the same name, so the two cannot overlap.

The migrator applies each file with autocommit on a pinned connection and records the version in a separate statement afterwards (`server/cmd/migrate/main.go:302-312`), deliberately without a wrapping transaction because of `CONCURRENTLY` (`:225-227`). So 257 commits and is recorded independently. If the process is interrupted, killed, or the deploy aborts after 257, the database sits at 257 with **no** pending-task uniqueness at all until a later run completes 258. Even on the happy path there is a window spanning the gap between files plus the concurrent build.

That window is not benign, because the constraint is load-bearing in application code:

```go
// server/internal/service/task.go:587
return errors.As(err, &pgErr) && pgErr.Code == "23505" &&
    pgErr.ConstraintName == "idx_one_pending_task_per_issue_agent"
```

With the index gone, `isDuplicatePendingTaskErr` can never fire, so the concurrent-enqueue race from #5914 stops being rejected and genuinely inserts duplicate pending rows for the same `(issue_id, agent_id)`. Then 258's `CREATE UNIQUE INDEX CONCURRENTLY` fails against that now-duplicated data, and the deploy is stuck in the worst state: no index, and a migration that cannot advance without manual cleanup.

Two things compound it:

- **258 has no `IF NOT EXISTS`.** A failed or interrupted `CREATE INDEX CONCURRENTLY` leaves an INVALID index behind under that name. Since the version was not recorded, the migrator retries the same file on the next run, which now fails with "relation already exists" — permanently, until someone drops it by hand. `255_agent_task_queue_chat_pending_deferred_v3.up.sql` uses `CREATE INDEX CONCURRENTLY IF NOT EXISTS` for exactly this reason.
- **The down path has the mirror-image gap** — `258.down` drops, `257.down` recreates v1.

### Suggested fix — the repo already has the pattern one migration earlier

The 255/256 pair does precisely the right thing, and says so:

> Build v3 before dropping v2 so old and new server versions both retain a usable index throughout a rolling deploy. […] Kept separate […] so both files contain exactly one concurrent index statement and no rollout window is left without coverage.

Applying that here:

1. Create the widened index under a **new name** (e.g. `idx_one_pending_task_per_issue_agent_v2`) with `CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS` in the **lower-numbered** file.
2. `DROP INDEX CONCURRENTLY IF EXISTS idx_one_pending_task_per_issue_agent` in the higher-numbered file.
3. Mirror the reverse order in the down files.

Coexistence during the window is safe in both directions: v1's predicate is a strict subset of v2's, old server versions never write `channel_issue_media_pending` deferred rows so they cannot violate v2, and the new deferred rows are outside v1's predicate. At no point is the existing invariant unenforced.

**One thing this requires:** while both indexes exist, a `queued`/`dispatched` collision is covered by both, and Postgres reports whichever raises first — so `isDuplicatePendingTaskErr` must accept **both** names during the compatibility window (and the v1 name can be dropped in a later release). Without that, the benign duplicate-enqueue race stops being recognized mid-deploy and surfaces as a raw 500 instead of the coalesced / structured-409 outcome — a straight regression of #5914. Worth a test that pins both names.

If a single release can't guarantee mixed-version safety here, splitting across two releases is fine; the invariant just must never be dropped first.

---

Nothing else blocking this round. CI is green on this head and the mergeable state is clean, but neither exercises the interrupted-migration path, so this one needs to be right by construction. Happy to re-review as soon as the migration ordering is flipped.

yyclaw:
@Bohan-J Addressed the remaining deploy-safety blocker at head `03456c975`.

- Migration 257 now builds `idx_one_pending_task_per_issue_agent_v2` first with `CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS`, while v1 continues enforcing queued/dispatched uniqueness.
- Migration 258 drops v1 only after v2 is available. The down migrations restore v1 before removing v2, so coverage is retained in both directions.
- `isDuplicatePendingTaskErr` accepts both v1 and v2 names during the rolling-deploy compatibility window. The unit test pins both names, and the DB race test now verifies the final v2 predicate/constraint.
- I re-ran the complete migration sequence on an isolated PostgreSQL database, including the actual `258 down -> 257 down` and `257 up -> 258 up` order. The affected service/handler packages and `go vet` pass.

I also rechecked scope before pushing: the four functional paths approved in your previous review are unchanged. All GitHub checks are now green and the PR is mergeable. Please take another look when convenient.

Bohan-J (changes requested):
Re-reviewed at head `03456c97`. The round-2 ordering problem is properly fixed, verified in the files:

- v2 is built first under its own name, v1 dropped only in the later migration (`257_...v2.up.sql`, `258_drop_pending_issue_agent_v1.up.sql`), so `queued`/`dispatched` uniqueness is enforced continuously.
- The down path is correctly reversed — v1 restored in `258.down`, v2 dropped in `257.down`.
- `isDuplicatePendingTaskErr` (`service/task.go:586-597`) accepts both names for the coexistence window, with `TestIsDuplicatePendingTaskErr` pinning them.

That's exactly the shape I asked for. One consequence of that request needs correcting, though — and it's on me: I asked for `IF NOT EXISTS` to make a failed build retry cleanly, and that flag introduces a different failure mode. The fix needs to satisfy **both** properties, not trade one for the other.

### `IF NOT EXISTS` treats a leftover INVALID index as success

`257_..._v2.up.sql:6` is:

```sql
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS idx_one_pending_task_per_issue_agent_v2
```

Per the [CREATE INDEX docs](https://www.postgresql.org/docs/current/sql-createindex.html), `IF NOT EXISTS` only checks whether a *relation with that name* exists — "there is no guarantee that the existing index is anything like the one that would have been created." It does not consult `pg_index.indisvalid`. And a `CREATE INDEX CONCURRENTLY` that fails on a deadlock, an interruption, or a uniqueness violation "will fail but leave behind an 'invalid' index."

So:

1. First run of 257 fails partway through the concurrent build → an INVALID `..._v2` relation is left behind, and the version is **not** recorded (`cmd/migrate/main.go:302-312`).
2. Next run re-executes 257. `IF NOT EXISTS` sees the name, emits a NOTICE, skips — no validation, no rebuild — and returns success.
3. The migrator records 257 as applied, and 258 drops the still-valid v1.

The end state reports a fully-applied schema while no *valid* pending-task unique index exists. The leftover invalid index is ignored by the planner, and although it is still maintained on writes, it is incomplete by construction — rows missed during the failed build are not in it — so it cannot be relied on to reject a conflicting insert. That is precisely the invariant `isDuplicatePendingTaskErr` and the #5914 coalescing path are built on. `258.down:2` → `257.down:2` has the symmetric problem on the way back.

### Suggested fix — the migrator already has the right extension point

The constraint to work within: this migrator is linear and records per-file, and `CREATE INDEX CONCURRENTLY` can be neither wrapped in a transaction nor combined into a multi-statement file — so the cleanup cannot simply be prepended to 257.

`cmd/migrate/main.go:18-51` already defines `preMigrationHooks`, keyed by migration version, running on the pool immediately before that version's SQL file, with a hook failure aborting the run *before* `schema_migrations` is updated. A hook on `257_agent_task_queue_channel_media_pending_unique_v2` that checks `pg_index.indisvalid` for `idx_one_pending_task_per_issue_agent_v2` and `DROP INDEX CONCURRENTLY` on an invalid leftover gives you both properties at once: retries converge automatically, and a valid index is never dropped in favour of an invalid one. It also keeps each SQL file at exactly one concurrent statement, which is the convention the 255/256 pair documents.

Note hooks only run for `up` (`:293`), so the down path needs its own guarantee — either drop `IF NOT EXISTS` from `258.down` so a leftover fails closed with v2 still in place, or add a validation step.

If you'd rather not add a hook, the minimum acceptable version is a separate validation migration between create and drop that asserts `indisvalid` for v2 and errors otherwise. Because it is its own file, a failure leaves the version unrecorded, so the v1 drop never runs and the old index stays valid — fail-closed with an actionable error.

Plain removal of `IF NOT EXISTS` also fails closed and is safe (v1 keeps enforcing), but every failed build then needs a manual `DROP INDEX` before the deploy can proceed. Workable, just worse operationally — which is why I'd reach for the hook.

Worth a test for the leftover-INVALID → re-run path asserting the valid v1 survives; there is no `indisvalid` assertion anywhere in the repo today, and this is the class of bug that only shows up in a bad deploy.

---

Nothing else blocking. The business logic hasn't moved since last round and still looks right, CI is green, and the branch is clean against `main`. Happy to re-review once the validity gate is in.

yyclaw:
@Bohan-J Addressed the INVALID-index retry feedback and completed a final concurrency preflight at head `078f80c55`.

- Migration 257 now runs a pre-hook that inspects `pg_index.indisvalid`: missing/valid v2 is a no-op, an INVALID v2 is dropped concurrently before retry, and a same-name non-index relation fails closed. Migration 258 down no longer uses `IF NOT EXISTS`, so rollback leftovers also fail closed while v2 remains valid.
- The PostgreSQL regression test now creates a real leftover INVALID v2 via a failed concurrent unique build, reruns the production migrator, and asserts v1 remains valid until v2 is rebuilt and recorded.
- Final preflight found and closed one additional visibility window: the issue and its media-deferred assigned task now commit in the same transaction. A deterministic after-commit competitor test proves an ordinary queued task cannot win the slot. Optional Composio overlay hydration stays outside the transaction while the task is inert, with a conditional update that cannot overwrite a concurrently merged comment plan.

Verification on this head:
- focused PostgreSQL concurrency/recovery tests
- full `internal/service`, `internal/handler`, channel engine, migrator, and migration suites
- focused `go test -race` and `go vet`
- all 10 GitHub checks green; PR is mergeable/CLEAN

Could you please re-review when convenient?

Bohan-J (approved):
Approving at head `078f80c5`. Every item raised across the previous three rounds is closed, and I verified each one in the code rather than from the summary.

**Migration safety** — the piece I got wrong twice, now correct in both directions:

- `preMigrationHooks` registers `cleanupInvalidConcurrentIndexHook` for `257_agent_task_queue_channel_media_pending_unique_v2` (`cmd/migrate/main.go:57-99`). It resolves via `to_regclass` (absent → no-op), fails closed if the relation exists but isn't an index, no-ops when the index is already valid, and drops only the INVALID leftover with a sanitized identifier. Running on the pool keeps `DROP INDEX CONCURRENTLY` out of a transaction block.
- `258.down` drops `IF NOT EXISTS` with a comment explaining why, so a leftover INVALID v1 fails closed while the valid v2 is still in place instead of being recorded as a successful restore.
- Expand/contract ordering holds both ways: v2 built in 257, v1 dropped in 258; down restores v1 before removing v2. The invariant is never unenforced.
- `TestRunMigrationsRepairsInvalidConcurrentIndexBeforeRetry` is a real reproduction, not a mock — it seeds conflicting deferred rows, lets an actual `CREATE UNIQUE INDEX CONCURRENTLY` fail, asserts v1 valid / v2 INVALID, then drives `runMigrations` and asserts both indexes end valid and the version is recorded. It also asserts the production hook is registered for that version, so the wiring can't silently rot.

**Concurrency** — the issue row and its media-gated deferred task now insert through the same `qtx` and commit together (`service/issue.go:332-343`), which closes the publish-vs-enqueue window entirely rather than narrowing it: any observer that can see the committed issue also sees the inert deferred task. The post-commit Composio hydration is correctly conditional — `SetDeferredChannelIssueTaskRuntimeOverlay` (`queries/agent.sql:357-368`) guards on `status = 'deferred'`, the media-pending marker, `trigger_comment_id IS NULL` and the expected originator, so a comment merge that won the race can't be clobbered, and a zero-row result is logged rather than treated as an error.

**Still-good from earlier rounds** — `/issue` stays terminal on both the success and active-duplicate paths with media finalized on each (`router.go:421-451`); delete collects attachment URLs inside the `LockIssueForDelete` transaction; mobile invalidates the attachment cache on both the live event and reconnect; `isDuplicatePendingTaskErr` accepts both index names for the coexistence window, with a test pinning them.

Also confirmed 257/258 don't collide with `main` (currently at 256), the branch is clean against `main`, and all 10 checks pass.

@yyclaw — thank you for this. You reported the bug and then carried the fix through three rounds of fairly demanding review, and the scope grew well past the original symptom: attachment ownership, task ordering, delete races, realtime invalidation, and finally rolling-deploy migration safety. The migration hook in particular is a better solution than what I suggested — my `IF NOT EXISTS` advice in round 2 introduced the invalid-index hole, and rather than just reverting it you built something that satisfies both properties and backed it with a test that genuinely reproduces the failure. That's a real contribution to the deploy tooling beyond this PR. Merging now.

PR #6389, opened 2026-08-04
Add Multica self-host setup verification

priyankarath2312:
## What does this PR do?

Added self-host setup verification details to the Multica README.

This documents the successful Docker Compose deployment, running services, health check verification, and access URLs so contributors can verify their local self-host setup easily.

## Related Issue

Closes #

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Updated `README.md`
- Added Multica self-host deployment verification section
- Documented Docker Compose startup command
- Added running container details (PostgreSQL, Backend, Frontend)
- Added backend health check verification
- Added local access URLs

## How to Test

1. Run:
   ```bash
   docker compose -f docker-compose.selfhost.yml up -d

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,58 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for the full development workflow, worktr
 
 An iOS mobile client lives in [`apps/mobile/`](apps/mobile/) — see its [README](apps/mobile/README.md) for how to build it onto your own iPhone.
 
+## Self-Host Setup Verification
+
+### Docker Compose Deployment
+
+Successfully configured and verified Multica self-host deployment using Docker Compose.
+
+### Services Running
+
+The following services were successfully deployed:
+
+* PostgreSQL with pgvector (`pgvector/pgvector:pg17`) — Healthy
+* Multica Backend (`ghcr.io/multica-ai/multica-backend:latest`) — Running
+* Multica Frontend (`ghcr.io/multica-ai/multica-web:latest`) — Running
+
+### Start Command
+
+```bash
+docker compose -f docker-compose.selfhost.yml up -d
+```
+
+### Verification
+
+Docker containers status:
+
+```text
+multica-postgres-1   Healthy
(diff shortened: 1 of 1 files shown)


Bohan-J:
Thanks for putting this together, @priyankarath2312 — and for actually running the self-host stack end to end. That's a genuinely useful thing to do as a first contribution.

That said, I'm going to close this one. Three reasons:

**1. The content already lives in `SELF_HOSTING.md`.** The start command, the `pgvector/pgvector:pg17` image, and the `http://localhost:3000` / `http://localhost:8080` URLs are all documented there — see "Step 1 — Start the Server" and "Manual Docker Compose Setup". The README already links to that guide from the header nav and from both Quick Install callouts, so this section wouldn't add information; it would add a second place that has to be kept in sync.

**2. It reads as a deployment log rather than instructions.** Phrasing like "Successfully configured and verified" and "The following services were successfully deployed" describes what happened on one machine. A README section needs to tell the next reader what to do, and hold across environments.

**3. The health check would mislead people.** This is the part most worth flagging, because it's a real detail: `GET /health` is a **liveness** probe. In `server/cmd/server/router.go` it's wired to `liveHandler`, which returns `{"status": "ok"}` unconditionally — it returns `ok` even when Postgres is unreachable or migrations haven't finished. Documenting it as the way to verify a successful deployment would tell users that a broken stack is healthy.

The endpoint that actually verifies the stack is `/readyz` (also aliased as `/healthz`). It pings the database and checks migration state, returning `503` with `{"status": "not_ready", ...}` when either fails. That's what `SELF_HOSTING.md` uses.

**If you'd like to turn this into something mergeable:** there is a real gap worth filling — a proper "verify your install" step in `SELF_HOSTING.md`, written to be environment-agnostic, using `/readyz`, and showing both the healthy and the `not_ready` responses so users can tell them apart. That would be a welcome PR and a more useful contribution than what's here.

Thanks again for the effort — hope to see you back.

PR #6390, opened 2026-08-04
feat(server): tell a chat agent which room it is in and who else can read it

seacen:
The runtime brief opens every chat run with "A user is messaging you directly in a chat window", and `buildChatPrompt` repeats it on line two. On Feishu, Slack and whatever channel comes next, a group room maps to ONE `chat_session` shared by everyone in it — so on a group turn both lines assert a private conversation that does not exist. Nothing else in the run says otherwise, so an operator who wants "don't repeat customer detail in a shared room" has no fact to write the rule against: the agent cannot tell the two cases apart.

The fact is already durable, and already on the row this claim reads. Every adapter writes the room shape to `channel_chat_session_binding.chat_type` through the shared session service, and #6330 taught the claim handler to read that row by session id. This carries one more column off it:

```go
if binding, berr := h.Queries.GetChannelChatSessionBindingBySessionAny(r.Context(), cs.ID); berr == nil {
    resp.ChatChannelType = binding.ChannelType
    resp.ChatType = binding.ChatType
```

No migration, no new query, no channel named in the handler.

The new `## Conversation Channel` section sits beside `## Requesting User` — that one says who the agent acts for, this one says who can read the answer. In a group it points at `## Task Initiator` for the asker, because in a shared room the asker changes per message and is not the runtime owner named above it.

It states facts and sets no policy. What may be said in front of a given audience varies by workspace and belongs in the operator's own agent instructions; a rule baked in here would fire on workspaces that never asked for it.

**Session-stable, so it lives in the brief** rather than the per-turn message (MUL-5377): a `chat_session` is bound to one conversation for its whole life and that conversation does not change shape between resumes. Who is speaking does change, and already travels per-turn in `## Task Initiator`.

**Degradation is two-step.** No channel at all — a web chat — renders nothing, and `buildChatPrompt` keeps its original wording, since a web chat is 1:1 by construction. A known channel whose shape the server did not report renders the platform and asserts no audience: defaulting to "direct" is exactly the wrong picture this section exists to correct.

One thing to call out: this adds `ChannelTypeWecom` and a `"WeCom"` display name — note 1 from your #6330 review. Two lines, touching nothing else, but if you'd rather it rode with the adapter I'll drop it from here.

`go build ./...`, `go vet ./...` and `gofmt` are clean; `internal/daemon/...`, `internal/handler` and `internal/integrations/channel/...` are green. Tests cover both room shapes, the unreported-shape fallback, the no-channel case, and the prompt's group/direct branch.

cc @Bohan-J

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -4871,6 +4871,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 		ProjectResources:                 convertProjectResourcesForEnv(task.ProjectResources),
 		ChatSessionID:                    task.ChatSessionID,
 		ChatChannelType:                  task.ChatChannelType,
+		ChatType:                         task.ChatType,
 		AutopilotRunID:                   task.AutopilotRunID,
 		AutopilotID:                      task.AutopilotID,
 		AutopilotTitle:                   task.AutopilotTitle,
--- a/server/internal/daemon/execenv/channel_type.go
+++ b/server/internal/daemon/execenv/channel_type.go
@@ -9,12 +9,60 @@ package execenv
 // server already serialized to JSON, and must not pull the server-side
 // integration packages (integrations/slack, integrations/lark) into its own
 // build just to read one discriminator. The canonical definitions live with
-// their adapters — slack.TypeSlack and channel.TypeFeishu — and both sides
-// agree on the wire strings below.
+// their adapters — slack.TypeSlack, channel.TypeFeishu, wecom.TypeWecom — and
+// both sides agree on the wire strings below.
 const (
 	ChannelTypeSlack  = "slack"
 	ChannelTypeFeishu = "feishu"
+	ChannelTypeWecom  = "wecom"
 )
+
+// Room-shape discriminators, mirroring channel_chat_session_binding.chat_type
+// (channel.ChatTypeP2P / channel.ChatTypeGroup). Every adapter persists this
+// column, so the shape of a conversation is known off one read whatever the
+// platform. Empty means the server did not report one — a web chat, which has
+// no binding row, or a server predating the field.
+const (
+	ChatTypeP2P   = "p2p"
+	ChatTypeGroup = "group"
+)
+
+// ChatAudience is what a run is allowed to say about who can read its replies.
+// Three states, because "unknown" is not "private": a web chat carries no
+// binding row at all and is 1:1 by construction, but an IM channel whose shape
+// the server did not report could be a room of any size, and the one thing the
+// copy must not then do is promise a privacy the conversation may not have.
+//
+// Both the brief's chat-mode line and the per-turn chat prompt open by naming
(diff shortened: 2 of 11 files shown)


seacen:
Pushed one change to the same commit, found while adversarially auditing my own branch: the two-step degradation this PR describes was only half enforced.

`writeConversationChannel` correctly refuses to name an audience when the channel is known but `chat_type` is not — but `writeWorkflowChat` and `buildChatPrompt` each tested `== group` and let everything else fall through to "messaging you directly". So on that input the brief declined to claim a private conversation in one section and asserted one in the section above it, and the chat-mode line is what the agent reads first.

Reachable two ways: a daemon newer than the server it claims from, and a binding deleted between enqueue and claim.

The three writers now share one `AudienceOf(channelType, chatType)` — known p2p or no channel at all is Direct, group is Group, a channel with no reported shape is Unknown and asserts nothing. Tests drive all four inputs through the whole brief and the whole prompt; all three fail against the previous two-way test.

Bohan-J:
Thank you for this — the problem you identified was real and the fix has shipped. `main` now carries the audience fact, landed in 86f2556 via #6428, and your commit is preserved there with your authorship.

I'm closing this PR rather than asking you for another round, and I owe you the reason.

The review converged on one design question that wasn't about correctness: **which channel the fact should travel in.** This branch put the room shape into the static runtime brief, which Claude Code loads ahead of the entire conversation as a cached prefix. Because Group / Direct / Unknown each render different brief text, one agent's three rooms produce three different prefixes. Moving the fact into the per-turn prompt keeps a single cached prefix per agent and costs one line per turn instead. That, plus trimming the copy from three statements of the same fact (the chat-mode line, `## Conversation Channel`, and the per-turn opener) down to one, is essentially all that changed:

- **Placement** — room shape stays out of `TaskContextForEnv`; a structural test now blocks it from being reconnected to the brief.
- **Size** — measured against the same `main` baseline, total first-turn prompt: group 8,931 → 8,075 B, direct 8,432 → 8,030 B, unknown 8,219 → 8,026 B.
- **Wording** — "everyone in the room receives every reply" and "only the person you are replying to receives what you send" are absolute delivery and privacy guarantees across three platforms. The shipped copy keeps only what `chat_type` actually proves: a group room is not private and members you haven't seen may read the reply; direct says it's a direct room; unknown says unknown.
- **Tests** — guards on structure and invariants (zero value, future discriminators, no second audience block) rather than 226 lines pinned to exact prose, so a copy edit doesn't churn the suite.

Iterating that with you across several review rounds would have been a lot of back-and-forth over prompt budget internals, which is our constraint to carry, not yours. So we re-rolled it on top of your commit instead.

Two things from this PR were kept as-is because they were right: the three-state classifier — treating "shape not reported" as its own state instead of defaulting to private — and the shared `AudienceOf` helper. The self-audit you pushed mid-review, where you caught the degradation being only half enforced, is exactly what produced that design.

Really appreciate the contribution, and the care in the write-up. Please do send more.

PR #6396, opened 2026-08-04
fix preserve colliding QwenPaw skills

yuluo-yx:
## Root cause

`prepareQwenpawWorkspace` derived each directory and manifest key directly from `sanitizeSkillName`. Because that normalization is not injective, names such as `A B` and `A-B` both became `a-b`, allowing the later skill to overwrite the earlier one.

## Changes

- Allocate deterministic, collision-free slugs with the existing `resolveSkillSlugs` helper before writing QwenPaw skills.
- Use that single allocation for directories, generated frontmatter, and `skill.json` keys and metadata.
- Add a regression test proving both colliding skills and their distinct contents are preserved.

## Verification

- `go test ./internal/daemon/execenv -run '^TestPrepareQwenpawWorkspace' -count=1`
- `go test ./internal/daemon/execenv -count=1`
- `go test ./pkg/agent -run '^TestQwenpaw' -count=1`
- `go vet ./internal/daemon/execenv`
- `git diff --check`
- Diff size: 70 additions + 12 deletions = 82 lines

Container- or cluster-backed integration/E2E tests were not run locally and remain for existing CI.

## Risk and rollback

Risk is limited to QwenPaw workspaces with colliding normalized skill names. Non-colliding skills keep their existing slugs. Roll back by reverting commit `8eb0ab19b`.

Closes OSS-11
Related finding: OSS-7

### DIFF
--- a/server/internal/daemon/execenv/qwenpaw_workspace.go
+++ b/server/internal/daemon/execenv/qwenpaw_workspace.go
@@ -58,13 +58,10 @@ func prepareQwenpawWorkspace(qwenpawWorkspace string, workspaceSkills []SkillCon
 	}
 
 	skillNames := make([]string, 0, len(workspaceSkills))
-	for _, skill := range workspaceSkills {
-		slug := sanitizeSkillName(skill.Name)
+	slugs := resolveSkillSlugs(workspaceSkills)
+	for i, skill := range workspaceSkills {
+		slug := slugs[i]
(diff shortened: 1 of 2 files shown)


multica-eve:
感谢你的贡献！这个修复解决了名称规范化冲突时技能被静默覆盖的问题，测试也覆盖了关键场景。我们现在将其合入。

PR #6398, opened 2026-08-04
[MUL-5736] fix(channel): make issue commands terminal outcomes

### LINKED ISSUE #6397: [Bug]: Channel issue commands run twice and hide duplicate conflicts
### Deployment type

All

### What happened?

For channel adapters that route text messages through the shared channel engine, `/issue <title>` creates the issue synchronously and then schedules the same user message as a normal chat task. The assigned agent consequently processes the `/issue` input a second time.

This produces two user-visible problems:

- a successful command can be followed by an agent response claiming the same active issue already exists;
- sending the same `/issue <title>` again returns a generic internal-error notice because `ErrActiveDuplicate` is treated as an infrastructure dispatch failure.

Expected behavior:

- `/issue` is a terminal command outcome: the issue service creates and enqueues the issue task, but the shared router does not also enqueue a chat task;
- an active duplicate is returned to the channel as a normal business conflict containing the existing issue identifier, rather than as an internal error;
- a handled command remains visible in chat history but is not absorbed into a later chat task batch.

### Steps to reproduce

1. Configure a Lark or Slack channel that uses the shared channel router.
2. Send `/issue channel command should run once` as a normal channel message.
3. Observe the immediate issue-created confirmation.
4. Observe that the same command is also scheduled as a chat task and processed again.
5. Send the same command once more.
6. Observe a generic internal-error response instead of an active-duplicate product result.

### Screenshots (optional)

Not included. The failure is pinned by shared-router and adapter-level regression tests.

```shell
Affected path: shared channel Router text-message /issue handling
Affected adapters: Lark and Slack event-message ingestion
Observed environment: local development checkout
Not affected: Slack native slash-command Quick Create path
Database migration required: no
```

yyclaw:
## What does this PR do?

Makes text-message `/issue` handling a terminal product outcome in the shared channel engine.

Thinking path:

1. `IssueService.Create` already creates the issue and enqueues its assigned agent's issue task.
2. The shared `Router` previously fell through after that synchronous create and scheduled the same channel message as a chat task.
3. The agent then saw `/issue` a second time, while a repeated title surfaced `ErrActiveDuplicate` as a generic infrastructure error.
4. This change returns immediately after a successful create or active-duplicate result, persists handled command turns with an internal message kind, and excludes those turns from future channel chat batches.
5. Lark and Slack repliers render the duplicate as a normal business conflict containing the existing issue identifier.

The stored command remains visible as an ordinary user message through the existing API normalization.

Risk notes:

- `channel_command` is internal persistence metadata; unknown message kinds already normalize to `message` at the API boundary.
- handled command rows intentionally remain unowned so later chat tasks cannot execute them.
- Slack's native slash-command Quick Create path is separate and unchanged.

## Related Issue

Closes #6397

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Return terminal shared-engine results for successful and duplicate `/issue` commands.
- Keep handled command messages visible while excluding them from later chat task batch ownership.
- Render active-duplicate outcomes in Lark and Slack instead of emitting generic internal errors.
- Add router, persistence, mapping, and real outbound-replier regressions for success, duplicate, media, and infrastructure-failure paths.

## How to Test

1. Run `go test -race ./internal/integrations/channel/engine ./internal/integrations/lark ./internal/integrations/slack -count=1` from `server/`.
2. Run `go test ./internal/integrations/channel/engine -run TestChannelIssueCommandIsExcludedFromLaterChatTaskBatch -count=1 -v` with a migrated local database.
3. Run `go vet ./internal/integrations/channel/engine ./internal/integrations/lark ./internal/integrations/slack` and `make sqlc`.
4. In Lark or Slack, send `/issue terminal outcome`, then repeat it. The first reply should confirm creation once; the second should report the existing active issue without starting a chat task.

Local result: all targeted race tests, the database regression, vet, sqlc generation, and `git diff --check` pass. A full repository race run was also attempted; unrelated environment-dependent suites failed because the shared local database was behind `upstream/main` (for example, missing `autopilot.pause_reason`), readiness dependencies were unavailable, and existing cross-package DB/daemon timing tests interfered under full parallelism.

## Checklist

## AI Disclosure

**Prompt / approach:** Investigated the shared channel router and issue-service contracts from code, isolated the fix from the DingTalk integration, added cross-adapter and persistence regressions, then ran a two-round review-and-fix loop before commit and push.

## Screenshots (optional)

Not applicable; this is a backend channel-routing behavior fix with automated regression coverage.

### DIFF
--- a/server/internal/integrations/channel/engine/resolvers.go
+++ b/server/internal/integrations/channel/engine/resolvers.go
@@ -58,6 +58,10 @@ type Result struct {
 	IssueNumber     int32
 	IssueIdentifier string
 	IssueTitle      string
+	// IssueDuplicate marks an /issue command that did not create a new issue
+	// because the shared duplicate guard found the active IssueID above.
+	// Repliers render this as a business conflict, never as an internal error.
+	IssueDuplicate bool
 	// runScheduled reports whether this ingest scheduled a normal chat run.
 	// It is Router-internal state: repliers must continue to use Outcome.
 	runScheduled bool
--- a/server/internal/integrations/channel/engine/router.go
+++ b/server/internal/integrations/channel/engine/router.go
@@ -409,6 +409,21 @@ func (r *Router) processClaimed(ctx context.Context, set ResolverSet, msg channe
 		// chat reply's.
 		prefix := r.issuePrefix(ctx, inst.WorkspaceID)
 		issueRes, err := r.createIssue(ctx, inst, set.OriginType, identity.UserID, sessionID, *appendRes.IssueCommand, prefix)
+		if errors.Is(err, service.ErrActiveDuplicate) && issueRes.DuplicateIssue != nil {
+			duplicate := *issueRes.DuplicateIssue
+			res.IssueID = duplicate.ID
+			res.IssueNumber = duplicate.Number
+			res.IssueTitle = duplicate.Title
+			res.IssueIdentifier = service.IssueIdentifier(prefix, duplicate.Number)
+			res.IssueDuplicate = true
+			// A duplicate is a terminal product outcome, not an infrastructure
+			// failure and not a chat prompt. Media binding remains independent,
+			// exactly like the successful direct-create path below.
+			if resolveMedia {
+				r.enqueueMedia(set, inst, identity, appendRes.MessageID, msg, sessionID, localMediaDeadline)
+			}
+			return res, postAppendFinalize, nil
+		}
 		if err != nil {
 			return Result{}, postAppendFinalize, fmt.Errorf("create issue from command: %w", err)
 		}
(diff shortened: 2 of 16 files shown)


Bohan-J:
Thanks for this, @yyclaw — really nice contribution. The thinking path was the right one: since `IssueService.Create` already enqueues the assigned agent's task, the shared router falling through to schedule the same input as a chat run was the actual defect, and treating an active duplicate as a product outcome rather than an infrastructure error is the correct boundary. The regressions across the router, persistence, and both the Lark and Slack repliers made the contract easy to verify.

One follow-up landed on top of your branch in d73e498. A `channel_command` turn is excluded from every later input batch, but it still persisted a media deadline that both session-scoped media gates counted:

- `GetChannelMediaPendingUntil` and `PromoteChannelChatTasksIfMediaReady` scan every user message in the session, so a plain message sent right after an `/issue` carrying an attachment was deferred behind the command's media binding.
- On the create-failure path this was worse: `enqueueMedia` is never called there, so nothing clears the marker and the next message waited out the full fallback budget.

Both gates now use the same message population as `LinkUnownedChannelChatMessagesToTask`, so only a turn that can join a batch can gate one. `DeferChatTaskForSealedPendingMedia` already scopes by task id and needed no change. Two DB-level regressions in `server/internal/service/task_channel_command_media_test.go` cover both gates, and the existing `TestEnqueueChatTaskDefersForChannelMediaAndPromotesWhenReady` still pins the ordinary case.

CI is green and this is merging now. Thanks again for the clear repro and the test coverage — it made the follow-up straightforward to isolate.

PR #6401, opened 2026-08-05
fix(chat): keep a user's own message in the transcript (MUL-5711)

NevilleQingNY:
Fixes MUL-5711.

## The bug

A human's own chat message is the only row in the transcript that **no WebSocket handler ever writes into the cache**.

`chat:message` carries the whole message — `message_id`, `role`, `content`, `task_id`, `created_at` (`server/internal/handler/chat.go`, `ChatMessageEventPayload`) — but the handler read only `chat_session_id` and invalidated. So the user's prompt reached the UI **solely** through the refetch that invalidate scheduled, while the agent's reply is inline-inserted by `chat:done` and is never at risk. That asymmetry is exactly the reported symptom: the reply renders, the prompt that produced it is missing, and a refresh brings it back because the row was always in the database.

The refetch is then thrown away by the quick-actions supplement: `applyChatQuickActionsToCache` cancels the in-flight messages fetch before patching, and `cancelQueries` defaults to `revert: true` — so the cache is *rolled back* to the snapshot taken when that fetch started, not merely left alone. Both message caches are `staleTime: Infinity` and nothing re-fetched afterwards, so the hole survived until a remount.

## What changed

**1. `chat:message` inline-inserts the user turn** (`applyChatMessageToCache`), mirroring `chat:done`, then invalidates as before.

- Only `role: "user"`. `SendChatMessage` is the event's one producer, and an assistant row fabricated from this payload would carry no `elapsed_ms` / `message_kind` / `quick_actions` **and** would make `applyChatDoneToCache`'s id dedup skip the richer row it is about to write.
- The invalidate stays: this payload has no `attachments`, so the reconciling refetch is what fills them in for clients that did not send the message.
- Idempotent by `message_id`, so the sender's own optimistic row wins and reconnect replay is a no-op.

**2. `chat:quick_actions` re-invalidates after its cancel+patch.** One request per supplement, and it cannot lose the pills: `SupplementChatQuickActions` persists the actions *before* publishing the event, so the refetch this schedules reads them back.

**3. `firstItemIndex` is anchored to a loaded message** (`useChatFirstItemIndex`) instead of derived from page sizes.

The newest page is a fixed 50-message window and TanStack re-derives each page cursor on refetch, so every turn pushed two older messages into the next page — the old `BASE - (messages in every page but the newest)` fell by 2 per turn, telling react-virtuoso rows had been **prepended** when only appends happened. It could also jump by the whole base when the list mounted empty (a pending task holds it open), which react-virtuoso reads as "drop that many rows off the head" and which wrecks its size bookkeeping. Anchoring means the value moves only when rows really enter or leave the head. Both chat surfaces now share the rule instead of deriving it twice.

## Verification

- `pnpm typecheck` — 6/6 tasks pass
- `pnpm lint` — no new problems (16 pre-existing warnings, none in the touched files)
- `pnpm test` — core 1239 tests, views 3541 tests, all pass
- Red/green: the new quick-actions regression test fails with the re-invalidate removed and passes with it.

New tests: 5 for `applyChatMessageToCache` (insert into both caches, idempotence vs. the local write, assistant turns left to `chat:done`, no seeding of an unfetched cache), 1 for the post-cancel re-sync, and 10 for the anchored index (per-turn stability, load-older prepends, momentarily-empty list, session switch, deleted anchor).

Not covered here: the browser-level confirmation that the scroll no longer jumps. The index arithmetic is unit-tested, but the visual result needs a real layout engine.

### DIFF
--- a/packages/core/realtime/use-realtime-sync.test.ts
+++ b/packages/core/realtime/use-realtime-sync.test.ts
@@ -14,6 +14,7 @@ import { workspaceKeys } from "../workspace/queries";
 import type {
   ChatDonePayload,
   ChatMessage,
+  ChatMessageEventPayload,
   ChatPendingTask,
   ChatMessagesPage,
   ChatSession,
@@ -23,6 +24,7 @@ import type {
 import {
   applyChatCancelFinalizedToCache,
   applyChatDoneToCache,
+  applyChatMessageToCache,
   applyChatQuickActionsToCache,
   applyChatSessionUpdatedToCache,
   applyWorkspaceUpdatedToCache,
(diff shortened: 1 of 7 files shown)


NevilleQingNY:
Superseded: reopening as a narrower PR per review — A+B plus a single shared message-cache helper, with the firstItemIndex/virtual-list change (C) split out to its own issue. Branch kept so the C work can be revived.

PR #5044, opened 2026-07-07
fix(daemon): mark workspaces root so escaped subprocesses still fail closed MUL-4199

beastpu:
Fixes #5043. Follow-up to #4209 / #4204.

## Root cause

The daemon-task marker from #4209 is written into each task **workdir** and discovered by walking **upward** from the CLI's cwd. A subprocess that lost every `MULTICA_*` env var (sandbox fault) and runs `multica` from the workdir's **parent** directory sits above the marker: the walk finds no daemon signal, `resolveToken` falls back to the user's config PAT, and agent writes are misattributed to the workspace owner (`author_type=member`). Reproduced in production; see the issue for the exact tree shape.

## Fix

Write a persistent marker at `{WorkspacesRoot}/.multica/daemon_task_context.json`:

- `execenv.EnsureWorkspacesRootMarker` — idempotent; never clobbers a foreign file at the marker path (reported as an error instead).
- Called on daemon startup (`Run`) and re-ensured on every task `Prepare` (self-heals if removed while the daemon runs). Both call sites are non-fatal: a failure degrades to the existing per-workdir protection.

Every directory under `WorkspacesRoot` is daemon-owned, so the CLI's existing upward walk now hits a signal from anywhere inside the tree — task dir, workspace dir, sibling task dirs, or the root itself — and fails closed demanding a `mat_` token. **No CLI changes**; normal user cwds are unaffected (`MULTICA_SERVER_URL` alone still isn't a daemon signal, per #4209's design).

GC and disk-usage scans are unaffected: they iterate directory entries only, and the root-level `.multica/` contains a single JSON file.

## Tests

- `TestEnsureWorkspacesRootMarker` — fresh write, idempotency, foreign-file refusal (content preserved), empty-root rejection.
- `TestPrepare_WritesWorkspacesRootMarker` — Prepare self-heals the root marker.
- `TestNewAPIClient_WorkdirParentEscapeFailsClosed` — full reproduction: config seeded with a `mul_` PAT, daemon-shaped tree, cwd at the workdir's parent, all env cleared → `resolveToken` returns empty and `newAPIClient` fails demanding a `mat_` token.

`go vet` + `go build` + `go test ./internal/daemon/... ./cmd/multica/...` pass on this branch.

## Residual risk (documented, out of scope)

[...]

PR #4437, opened 2026-06-23
MUL-3548: fix(cli): add daemon signal check to resolveToken() to prevent silent PAT fallback in agent child processes

Oxygen56:
## Summary

When a daemon-managed agent process spawns child subprocesses, the runtime may not forward `MULTICA_AGENT_ID` / `MULTICA_TASK_ID`, causing `inAgentExecutionContext()` to return `false`. Without this fix, `resolveToken()` would then silently fall back to the user-global CLI config token, which is how agent operations land as the wrong actor (e.g. comments attributed to the runtime owner rather than the agent).

## Fix

Insert a `MULTICA_DAEMON_PORT` check between the agent-context guard and the config fallback in `resolveToken()`. `MULTICA_DAEMON_PORT` is always set by the daemon when it spawns agent tasks and is less likely to be stripped by runtime child-process isolation. When detected, we **fail closed** (return empty) instead of reaching the user-global config.

## Trade-off (fail-closed)

A process that is genuinely not in a daemon-managed context but somehow has `MULTICA_DAEMON_PORT` set will also refuse to use the config token. In practice, `MULTICA_DAEMON_PORT` is only set in the agent env block injected by the daemon's task runner, so the signal is specific enough to be reliable.

## Changes

- **`server/cmd/multica/cmd_auth.go`**: Added `MULTICA_DAEMON_PORT` env check in `resolveToken()` (+8 lines)
- **`server/cmd/multica/cmd_agent_test.go`**: Added two test cases to `TestResolveToken_AgentContextSkipsConfig` covering daemon signal behavior (+22 lines), and explicitly unset `MULTICA_DAEMON_PORT` in the existing "outside agent context" test case

## References

- Upstream issue: #4204
- Root cause analysis: agent child processes lose `MULTICA_AGENT_ID` / `MULTICA_TASK_ID` env vars, causing silent fallback to the user-global CLI token
- Design decision: use `MULTICA_DAEMON_PORT` as a fail-closed guard — this env var is reliably set by the daemon task runner and unlikely to be stripped by child-process isolation

PR #4209, opened 2026-06-16
Prevent daemon task CLI fallback to user tokens

beastpu:
Fixes #4204

## Summary
- treat `MULTICA_DAEMON_PORT` as a daemon-managed execution marker in CLI auth/workspace resolution
- refuse config-file token/workspace fallback when that marker is present and `MULTICA_TOKEN` / `MULTICA_WORKSPACE_ID` are missing
- keep normal user CLI behavior where `MULTICA_SERVER_URL` alone can still pair with the saved config token

## Review against issue
- A daemon task subprocess that keeps `MULTICA_DAEMON_PORT` but loses `MULTICA_TOKEN`, `MULTICA_AGENT_ID`, and `MULTICA_TASK_ID` now fails closed instead of silently using a `mul_` PAT.
- `newAPIClient` applies the existing task-scoped `mat_` requirement to the daemon marker path as well.
- The change intentionally does not treat `MULTICA_SERVER_URL` alone as daemon context because users commonly set it for normal CLI sessions.

## Verification
- `go test ./cmd/multica -run 'TestResolve(Token|WorkspaceID)|TestNewAPIClient'`\n- `go test ./cmd/multica`\n- `git diff --check`

PR #6407, opened 2026-08-05
MUL-5742 fix(cli): isolate daemon task profile state

rongchenzheng:
## Summary

- add a private per-task Multica config root without replacing provider-facing `HOME` or XDG variables
- make task API authentication fail closed on missing or non-task credentials and prevent fallback to human profile state
- isolate task `config show/set`, redact task `auth status`, and reject human/local profile and daemon commands in managed tasks
- document the phase-one boundary and its same-UID filesystem limitation

## Tests

- `/tmp/multica-go1.26.1/bin/go test ./internal/cli ./cmd/multica ./internal/daemon/execenv ./internal/daemon -count=1`
- `git diff HEAD^ --check`

Full `go test ./...` was also attempted. It has failures in unmodified packages: `internal/daemon/repocache` encounters an already-existing test branch, and several `pkg/agent` process/timing tests fail around their five-second boundaries in this environment.

## Tracking

Multica governance issue: ERI-46 (`50cb7636-c632-4219-b6be-31eb99f694b2`).

This phase prevents implicit Owner-profile resolution. It does not claim hard same-UID filesystem confidentiality; a dedicated Unix user, container, VM, or equivalent OS boundary is required for that guarantee.


Bohan-J (approved):
Thank you for this contribution, @rongchenzheng — and welcome as a first-time contributor! 🎉

This closes out the last open path in the task-identity hardening series (#4142 → #4437 → #4209 → #5044): the CLI profile/config state itself. The boundary is exactly right — task-local `MULTICA_TASK_CONFIG_ROOT` redirection with fail-closed guards for human-local commands, while deliberately leaving `HOME`/XDG untouched for provider tooling. The negative-test coverage (owner profile untouched, fail-closed without task root/token, no credential material in `auth status`) is thorough, and the docs honestly state what this does *not* guarantee (same-UID file access).

Final verification on top of current `main` (merge of `2c06a8b0b`): build, vet, `internal/cli`, `internal/daemon/execenv` (incl. the ByteIdentical brief-stability tests), `internal/daemon`, and the full `cmd/multica` suite all pass. Merging.

Follow-ups tracked on our side (not blockers): a task-scoped read-only diagnostics channel (`daemon status` + default-scope `disk-usage`) and an explicit, auditable owner-profile opt-in.

PR #6410, opened 2026-08-05
fix(autopilot): invalidate empty cache for scheduled tasks (MUL-5747)

rongchenzheng:
## Summary

- reuse the router's fully wired `TaskService` and `AutopilotService` for background workers
- ensure schedule-triggered Autopilot enqueue bumps the same Redis empty-claim cache used by claim handlers before sending the daemon wakeup
- add a regression test that pins background/router service identity

## Root cause

`NewRouterWithOptions` wires `h.TaskService.EmptyClaim = NewEmptyClaimCache(rdb)`, but `main` constructed a second `TaskService` for the Autopilot scheduler. Schedule dispatch called `NotifyTaskEnqueued` on that second service: the daemon wakeup was delivered, while the claim path's cached empty verdict was not invalidated. An idle runtime therefore kept receiving empty claims until the 3-minute cache TTL expired. Manual dispatch used the router service and did not have this delay.

The same duplicate service was also passed to the runtime sweeper; reusing the router services removes that configuration drift.

## Validation

- `go test ./cmd/server ./internal/service -count=1`
- focused empty-cache race/order tests pass
- focused Autopilot scheduler tests pass
- `git diff --check`

No daemon restart, production mutation, migration, or real Monitor schedule enablement was performed.

### DIFF
--- a/server/cmd/server/main.go
+++ b/server/cmd/server/main.go
@@ -172,6 +172,10 @@ func envBool(name string, def bool) bool {
 	return v
 }
 
+func backgroundServices(h *handler.Handler) (*service.TaskService, *service.AutopilotService) {
+	return h.TaskService, h.AutopilotService
+}
+
 func main() {
 	logger.Init()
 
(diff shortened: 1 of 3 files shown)


Bohan-J:
Thanks for this — excellent first contribution, and a genuinely tricky bug to have spotted.

We independently traced the chain end to end and it holds up exactly as you described: the router assigns `h.TaskService.EmptyClaim`, `main` built a second `TaskService` for the schedule dispatch job, and because `EmptyClaimCache` is deliberately nil-safe the missed invalidation is *silent* — `Bump` no-ops while the daemon wakeup still fires. `ClaimTaskForRuntime` then short-circuits on the cached verdict with no bypass for wakeup-driven claims, so an idle runtime keeps returning empty claims until the TTL expires. We reproduced it against a real Redis and confirmed the shared-service fix clears it. Your "up to 3 minutes" figure is accurate, not rounded up.

Reusing the router's services is the right call: it's a strict superset of the instance `main` was building (same Queries, tx entry, Hub, Bus, Analytics, Metrics, Wakeup) plus the wiring that was missing, so nothing is lost and the runtime sweeper gets the same correction for free.

We pushed two commits on top before merging, both test-only — your implementation change is untouched:

- `TestBackgroundServicesReuseRouterServices` asserted that `backgroundServices` returns `h`'s fields, which is the helper's entire implementation. It stayed green when we reintroduced the original duplicate-service wiring in `main()` — the exact regression it was named for. It's replaced with an AST guard over `main.go` that fails when `main()` constructs its own services, stops calling `backgroundServices`, or passes it anything other than the router's handler. The repo already uses this pattern in `daemon/agent_command_names_test.go` and `metrics/business_pairing_test.go`.
- A service-level test covering the underlying hazard directly: an `EmptyClaim`-less `TaskService` fires the wakeup while the claim path's cached empty verdict survives. It runs against the Redis service in CI.

Both are mutation-verified — reintroducing the bug fails them.

One thing worth carrying forward from your fix: the deeper issue is that `NewTaskService` lets you build a half-configured instance with `EmptyClaim` attached afterwards, so this class of bug can reappear anywhere else someone constructs one. That's pre-existing and well outside the scope of this PR, but your change is what surfaced it.

Merging now. Thanks again!

PR #6413, opened 2026-08-05
feat(runtime): fail over provider-limit retries

okssusucha:
## Summary

- classify Claude session-limit responses as provider quota failures, including mixed-version daemon catchalls
- route quota/rate-limit retry children to the first configured online runtime with the same workspace, owner, daemon, and provider
- resume automatic retries from their exact failed parent across compatible runtimes, preserving workdir and resumable session context
- document `runtime_config.failover.runtime_ids` and keep retries bounded by the existing attempt budget

## Verification

- `go test ./pkg/taskfailure ./internal/service ./internal/handler`
- `go test ./...` passes all relevant packages; the existing `TestCodexExecuteSemanticInactivityAllowsContinuousMessages` timing test fails at its 72 ms timeout and reproduces unchanged on `origin/main`
- database-backed integration cases are included but skip locally without `DATABASE_URL`; CI should exercise them

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -1635,6 +1635,31 @@ type claimBuildFailure struct {
 func (h *Handler) buildClaimedTaskResponse(r *http.Request, task *db.AgentTaskQueue, runtime db.AgentRuntime, runtimeID, runtimeWorkspaceID string) (resp AgentTaskResponse, deliveredCommentIDs []pgtype.UUID, agentSkillCount, builtinSkillCount int, failure *claimBuildFailure) {
 	// Build response with fresh agent data (name + skills + custom_env + custom_args).
 	resp = taskToResponse(*task, runtimeWorkspaceID)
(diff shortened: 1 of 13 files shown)


multica-eve:
Thank you for putting this together and for the thoughtful work on preserving task progress across retries. We agree with the underlying problem: unattended tasks should recover gracefully when a provider quota is reached.

After reviewing the product boundary, we have decided not to proceed with this implementation. Multica plans to address this through platform-level, quota-aware recoverable scheduling: distinguishing quota resets from rate limits and provider capacity, pausing and resuming work with an explicit recovery policy, and making any billing or identity transition visible and opt-in.

Routing a failed attempt to another runtime while carrying over its session context does not currently provide enough guarantees around credential and session compatibility, and it may also replay external side effects that already occurred. Addressing those concerns would require a different platform-level design rather than incremental changes to this PR, so we are going to close it.

We appreciate the contribution—it helped clarify the requirements and will inform the design of the platform-level recovery flow.

PR #6422, opened 2026-08-05
fix(agent): fail a run whose provider session ran out of context (MUL-5739)

### LINKED ISSUE #6402: Agent stalls silently when provider session context is full: resumed session exits immediately, task misreported as completed, no auto-recovery
## Summary

When an agent's provider session (Claude Code, `runtime_mode: local`) exhausts its context window, subsequent tasks for the same working directory keep **resuming that saturated session** and exit immediately with a message like "context too long, please run /compact". The task is then reported as *completed* (not failed), and nothing automatically recovers — the agent is effectively dead for that workdir until a human manually removes/archives the session transcript on disk.

## What we observed

1. An agent working in a git worktree accumulated a ~2.5 MB session transcript and hit the context limit mid-task.
2. Every subsequent trigger (comment @mention → new task) resumed the same session and terminated within seconds, posting the "please run /compact" output as the task result comment.
3. The task status showed `completed`, so from the platform's point of view nothing was wrong — but no actual work happened. The pipeline silently stalled.
4. Manual workaround: archive the session `.jsonl` under the provider's per-project session directory, forcing a fresh session on the next trigger. After that, the same trigger worked immediately.

## Expected behavior (any of)

1. **Detect context exhaustion**: when the provider CLI exits with a context-length error, mark the task as `failed` with a distinct error (e.g. `provider context exhausted`) instead of `completed`.
2. **Auto-recover**: on that condition, automatically run `/compact` on the session or start a fresh session for the next task in that workdir.
3. **Surface it**: make "session is full and being resumed" visible in the task/agent UI so a human doesn't have to dig through provider transcript files to diagnose a stall.

## Environment

- Multica desktop app daemon on macOS (Apple Silicon), CLI v0.4.15
- Provider: Claude Code 2.1.220, `runtime_mode: local`

## Why it matters

Long-running agent pipelines (multi-issue TDD flows in one workdir) are exactly where sessions grow large. Today the failure mode is silent, misreported as success, and requires filesystem-level manual intervention — all three make it hard to trust unattended runs.

Bohan-J:
Refs #6402. MUL-5739.

**This PR does not close #6402.** It removes a known, demonstrated fragility in how context exhaustion is detected, and it very likely covers the reported scenario — but the specific escape that produced a `completed` task on the reporter's host has not been reproduced. See *Evidence boundary* below. #6402 stays open until the reporter confirms, or until a real `completed` frame is captured.

## The problem

A Claude Code session that exhausts its context window ends every later resume in seconds without doing any work. If such a run is recorded as `completed`, the platform publishes the CLI's "your context is full" notice as the agent's answer — so nothing fails, nothing retries, and the saturated session stays pinned as the resume pointer, which makes every new trigger replay it. The only escape is deleting the provider's transcript by hand.

## What the reproduction shows

Claude Code 2.1.220 (the version #6402 reports) was fetched from the official release channel, SHA-256 verified against its own `manifest.json`, and run against an endpoint returning the provider's 400 prompt-too-long shape with a saturated transcript resumed. Captured verbatim (fixtured at `server/pkg/agent/testdata/claude-code-2.1.220-context-exhausted-resume.jsonl` and replayed by the tests):

```
system/status  status=compacting
system/status  compact_result=failed  compact_error=exhausted
assistant      is_api_error_message=true  text="Prompt is too long"
result         subtype="success"  is_error=true  terminal_reason="prompt_too_long"
               api_error_status=400  result="Prompt is too long"
```

2.1.221 produces the same shape. Two findings drive the fix:

1. **`terminal_reason` is a structured enum** (`prompt_too_long`, `image_error`, `model_error`, `max_turns`, `completed`, …) and it is **present on the reported version** — so keying on it reaches the hosts that hit the bug, not only future ones.
2. **On the real frame `is_error` and `terminal_reason` both fire.** Whichever the runtime reads first decides the failure's label. `is_error` is derived from whether the last rendered message was an API error; `terminal_reason` states why the turn ended. Only the latter names the condition without depending on the CLI's prose.

## Root cause addressed

Detection rested on prose. Today the CLI happens to write `"Prompt is too long"`, which `Classify` recognises — but that is luck, and it has already failed once: #6360 was the same class of bug, where a 2.1.x reword dropped context overflow into `agent_error.unknown`, a reason no resume blacklist covers, so the saturated session stayed pinned and every trigger replayed it.

This PR moves the decision onto the structured field:

- **`pkg/agent`** reads `terminal_reason` and fails the run on `prompt_too_long`, ahead of `is_error`. The error quotes the enum token, so `Classify` lands it on `agent_error.context_overflow` — already on the `GetLastTaskSession` / `GetLastChatTaskSession` resume blacklists, so the saturated session is retired and the next task starts fresh. That is #6402's manual workaround, automated. The CLI's own `result` text rides along as detail, so nothing is lost for diagnosis.
- **`taskfailure.ContextExhaustedCompletion`** recognises the condition from a reported-successful *output*, for callers with no structured field. Every fingerprint is composite and taken from the Claude Code binary, under a 320-byte cap. The CLI's bare `"Prompt is too long"` is deliberately **not** matched: it is also an ordinary English answer, and this predicate only sees output someone believed was a success, so a match would fail a real task and retire a healthy session.
- **The daemon** applies it in its `completed` branch, with **no tool-count gate** — a context window fills up mid-task far more often than on turn one. The tool count only governs whether a run may be *replayed*, and nothing here replays: `agent_error.context_overflow` is absent from `retryableReasons`, so the task fails once and stops.
- **The server** applies the same classifier at the `/complete` boundary, so an installed daemon too old to carry the structured check can still be corrected without waiting for that host to update (same mixed-version argument as `NormalizeDaemonReason`, MUL-5370).

Auto-`/compact` is deliberately **not** attempted. A session past the limit cannot compact its way back under — the compaction call has to fit the whole history — which is why the reporter's session deadlocked instead of self-healing. Retiring the session and cold-starting the next task is the recovery that works, and it is the decision #6360 already made for the `is_error: true` path.

## Evidence boundary

The captured frames never show `is_error: false`. On those frames the base branch already fails the run and classifies it as `context_overflow` — so **this reproduction is not the reporter's escape.** What produced a `completed` task on their host remains unknown; one plausible shape is the CLI recovering and the model itself emitting a short meta message, which would leave no platform-side signal at all.

Accordingly:

- The **structured check is the load-bearing change**, and it is proven to work on the reported CLI version for the captured condition.
- The **text predicate is a bounded backstop** for shapes that have not been captured — defense-in-depth, not evidence of a reproduced root cause.
- The code comments state this scoping explicitly rather than implying broader coverage.

## Tests

- **Captured-frame replay:** the 2.1.220 transcript, unedited → run fails, publishes no output, classifies as `context_overflow`, keeps its session id, is not reported as a rejected resume.
- **Ordering regression:** `is_error:true` with (a) an empty `result` and (b) prose reworded to contain none of `Classify`'s phrases. Both degrade to `agent_error.unknown` — session not retired, stall returns — unless the structured reason is checked first. Verified by disabling the branch.
- **No over-reach:** a failure with no recognised terminal reason still reports the CLI's own error text (`API Error: 529 overloaded` → `provider_capacity_or_rate_limit`); `terminal_reason: "completed"` still completes.
- **Classifier precision:** every captured provider wording matches; the issue's paraphrase, `/compact` advice, a differently-caused compaction failure, and a long answer quoting all three markers do not. Length-cap boundary pinned.
- **Never replays:** `shouldRetryWithFreshSession` returns false for tools = 0, 1 and 17.
- **Server boundary (DB-backed):** an old daemon's `/complete` carrying the notice lands `failed` + `agent_error.context_overflow`, posts no agent comment, and the saturated session stops being offered by `GetLastTaskSession`. Controls: a real answer — including one that is exactly `Prompt is too long` — still completes, and its healthy session stays resumable.

## Verification

`go build ./...`, `go vet`, and `go test ./...` locally against a freshly migrated database. All packages pass except `cmd/multica`, which fails identically on the unmodified base commit in this environment (its tests read ambient CLI config) — unrelated to this change.


Bohan-J:
Pushed `232da5bd6` addressing all three must-fixes. One of them turned up a correction to my own earlier claim, so leading with that.

## Correction: 2.1.220 does **not** report `is_error: false`

I previously wrote that the reporter's build flipped `is_error`. That was inference, not evidence, and it's wrong.

Fetched Claude Code 2.1.220 from the official release channel (`downloads.claude.ai/claude-code-releases/2.1.220/darwin-arm64/claude`, SHA-256 verified against its `manifest.json`, unpacked to a temp dir — the local install was not touched) and re-ran the saturated-resume reproduction against it. Captured verbatim:

```
system/status  status=compacting
system/status  compact_result=failed  compact_error=exhausted
assistant      is_api_error_message=true  text="Prompt is too long"
result         subtype="success"  is_error=true  terminal_reason="prompt_too_long"
               api_error_status=400  stop_reason="stop_sequence"  result="Prompt is too long"
```

Two things follow:

- **`terminal_reason` is present on the reported version.** Keying on it reaches the hosts that hit the bug, not just future ones.
- **`is_error` and `terminal_reason` both fire on the real frame.** Which one the runtime reads first decides the label.

The frames are now fixtured unedited at `server/pkg/agent/testdata/claude-code-2.1.220-context-exhausted-resume.jsonl` and replayed by the test (must-fix #3). Only the init frame's local tool/skill inventory and absolute paths were stripped.

## Must-fix #1 — structured reason was shadowed

Confirmed, and worse than "a possible degradation": since both flags fire on the only frame we have, `is_error` first made the new branch **dead code**, with correctness resting on the CLI's prose happening to contain "prompt is too long". Reordered so a recognised terminal reason wins outright; the message it builds carries the CLI's `result` along as detail, so nothing is lost for diagnosis.

Regression-tested with `is_error:true` plus (a) empty `result` and (b) prose reworded to carry none of `Classify`'s phrases. Both fail without the reorder — verified by disabling the branch:

```
error "claude returned an error result without details" classifies as "agent_error.unknown", want "agent_error.context_overflow" — the session would not be retired
error "This chat has gotten too big to continue."       classifies as "agent_error.unknown", want "agent_error.context_overflow" — the session would not be retired
```

Plus a counterpart test: a failure with no recognised terminal reason still reports the CLI's own error text unchanged (`API Error: 529 overloaded` → `provider_capacity_or_rate_limit`).

## Must-fix #2 — bare `Prompt is too long` removed

Agreed, and the capture makes the case stronger than "reduce risk": the bare form **always arrives with `is_error` set**, so it reaches the server through `/fail`, where `Classify` rule 1 has routed `"prompt is too long"` → `context_overflow` since long before this PR. The success-path predicate never needed it — it only ever added the chance of failing a task whose entire answer was one common English sentence.

So it's gone; every remaining clause is composite. Control tests at the `/complete` boundary now cover both real-answer shapes, including output that is exactly `Prompt is too long`, and assert the healthy session **stays** resumable afterwards.

## What is still not proven

The capture does not show `is_error: false`, so the exact escape that produced a `completed` task on the reporter's host remains unreproduced — plausibly a shape where the CLI recovers and the model itself emits a meta message, which would leave no platform-side signal at all. I've stopped implying otherwise: the doc comments now say plainly that the structured check is the load-bearing fix (proven present on 2.1.220) and the text predicate is a bounded backstop for a shape we have not captured. I kept that backstop rather than deleting it because the reported symptom is a *completed* run, and removing it would leave that path with zero coverage — but it is your call, and dropping it is a one-line change if you'd rather ship only what the frames prove.

## Verification

`go build ./...`, `go vet`, and `go test ./...` against a freshly migrated database. All packages pass except `cmd/multica`, which fails identically on the unmodified base commit in this environment — unrelated.

PR #2963, opened 2026-05-21
Fix squad fallback delegation from leader comments

YOMXXX:
## Summary
- enqueue fallback mention tasks when an agent-generated comment mentions another agent or squad
- avoid duplicate pending fallback tasks and skip self-delegation for squad leaders
- add coverage for leader output that delegates to a worker

## Tests
- cd server && go test ./internal/service ./internal/handler -run 'TestCompleteTask_SquadLeaderFallbackMentionEnqueuesWorker|TestCompleteTask_SquadLeaderActionStillSynthesizesComment|TestCreateComment_SquadLeaderNoActionRejectsComment' -count=1
- git diff --check

Issue #2702, opened 2026-05-15
[Bug]: 小队功能以为自己派了子agent但是没有真正指派


kkbel:
### Deployment type

Official App

### What happened?

在评论区@小队，让小队负责人来安排小队的agent工作，提示已经成功安排了，但是实际没有安排，但是记录了已经安排的状态

### Steps to reproduce

<img width="1050" height="354" alt="Image" src="https://github.com/user-attachments/assets/60b697a9-cac5-44a1-a31b-5b10a85b5837" />

### Additional context (optional)

```shell

```

forrestchang:
<img width="1836" height="166" alt="Image" src="https://github.com/user-attachments/assets/4d50ef4e-7e9b-4448-94a5-bbeffcafbc2f" />

这里 agent 自己评估的是无需操作。

kkbel:
上一条是说已经委托发布工程师了，但实际上发布工程师没有被触发过

forrestchang:
> 上一条是说已经委托发布工程师了，但实际上发布工程师没有被触发过

具体使用的是什么 runtime & 模型？可以发我一下对应 Issue 的 ID，我们来排查一下。

kkbel:
请问这个“无需操作”的判断条件是什么？是小队指引可以控制的，还是内置的代码判断条件？我在agent提示词里和小队指引里都禁止了无需操作这个动作，依然大面积存在该问题

kkbel:
我发现问题应该是出在 使用了opencode，multica与opencode通信异常导致fallback之类的，所以就出现不工作了

Bohan-J:
@kkbel #2692 看下是不是这个问题导致的

ADanMan:
I think I found the actual mechanism here, and it's separate from #2692 (which Bohan-J linked above but is a different, already-closed opencode 1.15.0 stdout-flush bug — unrelated to this).

There are two comment-creation paths. The live one, `CreateComment` in `server/internal/handler/comment.go`, already resolves @mentions into delegated tasks via `computeCommentAgentTriggers`/`enqueueCommentAgentTriggers`. But when an agent's session ends without posting a live comment, `createAgentComment` in `server/internal/service/task.go` (called from `CompleteTask`) synthesizes a comment from the task's final output instead — and that path has zero mention-parsing. So if the delegation @mention only shows up in the synthesized completion text (matching the squad briefing's instructed format), it's silently dropped. That would line up with kkbel's note above about an opencode-related fallback.

[...]

PR #6456, opened 2026-08-05
fix(mentions): route synthesized agent comments

ianchen08:
## Summary
- route mentions from synthesized task-completion comments through the normal comment trigger engine
- preserve current invoke authorization, attribution, squad resolution, pending-task folding, and blocked outcomes
- add a regression test proving a synthesized agent mention enqueues the target
- document the fallback path in the built-in mentioning source map

## Why
When an agent finishes without posting a live comment, CompleteTask synthesizes a comment from final output. That persisted the handoff text but previously skipped mention routing, so the UI could show a delegation that never ran. This addresses #2702.

#2963 identified the same gap, but enqueues directly from TaskService. Current main has additional authorization, attribution, coalescing/deferred, and blocked-outcome behavior in the handler trigger engine. This change calls that existing engine instead of duplicating its logic.

## Tests
- go test ./internal/service -count=1
- go test -v ./internal/handler -run TestCompleteTask_SynthesizedMentionEnqueuesTarget\|TestCompleteTask_SquadLeaderActionStillSynthesizesComment -count=1 (isolated migrated PostgreSQL)
- git diff --check

### DIFF
--- a/server/internal/handler/comment.go
+++ b/server/internal/handler/comment.go
@@ -1930,6 +1930,32 @@ func isClientAuthorableCommentType(t string) bool {
 	return ok
 }
 
+// triggerSynthesizedAgentComment routes explicit mentions from the fallback
+// comment TaskService creates when an agent finishes without posting a live
+// comment. It deliberately reuses the normal comment trigger engine instead
+// of enqueueing from the service layer, preserving invoke authorization,
+// attribution, squad resolution, pending-task folding, and blocked outcomes.
+func (h *Handler) triggerSynthesizedAgentComment(ctx context.Context, issue db.Issue, comment db.Comment) {
+	if comment.Type != "comment" || comment.AuthorType != "agent" {
+		return
+	}
+
+	var parentComment *db.Comment
+	if comment.ParentID.Valid {
+		if parent, err := h.Queries.GetCommentInWorkspace(ctx, db.GetCommentInWorkspaceParams{
+			ID:          comment.ParentID,
+			WorkspaceID: issue.WorkspaceID,
+		}); err == nil {
+			parentComment = &parent
+		}
+	}
+
+	authorID := uuidToString(comment.AuthorID)
+	originatorUserID := uuidToString(h.TaskService.ResolveOriginatorFromTriggerComment(ctx, issue.WorkspaceID, comment.ID))
+	delegationAuthority := h.autopilotDelegationAuthorityFromComment(ctx, issue, comment)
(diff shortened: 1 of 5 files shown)


Bohan-J:
Thanks for digging into #2702, and for routing this through the existing trigger engine instead of duplicating the enqueue logic in `TaskService` — that instinct is right, and it's a real improvement over the approach in #2963.

We're going to pass on this one, but because of the premise rather than the implementation.

The fallback comment `CompleteTask` synthesizes is not a supported delivery mode — it is the repair for a run that went wrong. Every issue-kind run is required to post exactly one comment: see the `## Output` section in `server/internal/daemon/execenv/runtime_config_sections.go` ("Final results MUST be delivered via `multica issue comment add`" / "Post exactly ONE comment per run"). The only sanctioned no-comment outcome is a squad leader recording `no_action`, and that case never reaches the synthesizer — `suppressNoActionComment` short-circuits the whole block in `server/internal/service/task.go`. Chat, quick-create and autopilot run-only tasks carry no issue id by construction, so the `task.IssueID.Valid` guard excludes them as well.

So the code path this PR adds mention routing to fires if and only if an agent violated the delivery contract:

- agent follows the contract → the new hook never fires;
- agent violates it → the hook fires, and the text we now parse for delegations is raw final output that was never addressed to the platform.

That grants delegation authority to exactly the runs whose output we trust least. An agent that quotes an issue description, or writes "I should have handed this to @X", would now really enqueue those agents. Plain `@agent` mentions are intentionally not self-suppressed (see the doc comment above `resolveMentionedAgentCommentTriggers`), so the only bound is the `HasPendingTaskForIssueAndAgent` dedup.

We also read the user-visible bug in #2702 as living somewhere else — in two layers:

1. **Root cause.** The runtime failed to post the comment at all; the reporter traced it to an opencode fallback. Today that failure is invisible in the product: the server quietly synthesizes a comment and leaves a single `slog` line. Making the synthesized comment behave more like a real one hides that class of failure further.
2. **Presentation.** A synthesized comment is indistinguishable from an agent-authored one — there is no marker column, `source_task_id` is set on both, and `type` is `comment` either way. So the UI renders its mentions as live, clickable delegations. That is where "it says it delegated but nothing ran" actually comes from.

The direction we would rather take is to mark the synthesized comment as the degraded artifact it is, render it as "this run ended without a proper report, here is its raw output", and stop presenting its mentions as executed delegations. That closes the perception gap without making stdout side-effecting, and it surfaces the underlying runtime failure instead of papering over it.

Closing for that reason. The patch itself was careful — the optional-hook shape and the decision to reuse `triggerTasksForComment` were both the right calls — and if the marker-column direction above interests you, we would be glad to look at that.

PR #6459, opened 2026-08-05
MUL-5782: fix(daemon): preserve runtime config on write failure

Airmongsity:
## What does this PR do?

Prevents failed runtime-config writes from truncating user-owned provider instruction files such as `CLAUDE.md`, `AGENTS.md`, `CODEBUDDY.md`, and `QWEN.md`.

`os.WriteFile` opens existing files with `O_TRUNC`. If a subsequent write fails partway through, it returns an error but leaves the original file empty or partially overwritten.

This change stages the complete replacement in the destination directory and only replaces the target after the temporary file has been successfully written, chmodded, synced, and closed. The cleanup path uses the same mechanism.

This PR deliberately does not change:

- the daemon's existing non-fatal runtime-config error policy;
- conflict handling for concurrent user edits.

Those are separate product and concurrency concerns.

## Related Issue

Tracked in Multica: MUL-5782.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Stage runtime-config writes in a same-directory temporary file.
- Preserve the original target when writing, syncing, closing, or replacement fails.
- Apply the same protection to `CleanupRuntimeConfig`.
- Preserve existing Unix permission bits and user-owned symlinks.
- Use `ReplaceFileW` on Windows so replacement preserves the target DACL and metadata.
- Add a deterministic cross-platform partial-write regression test.
- Add Linux kernel-level regression tests using `RLIMIT_FSIZE`.
- Add a Windows DACL preservation regression test.

## How to Test

On Debian/Linux:

```bash
GOTOOLCHAIN=auto go test ./internal/daemon/execenv \
  -run '^(TestWriteRuntimeConfigFilePartialWriteFailurePreservesOriginal|TestCleanupRuntimeConfigPartialWriteFailurePreservesInjectedFile)$' \
  -count=1 -v

GOTOOLCHAIN=auto go test ./internal/daemon/execenv -count=1
```

Observed on Debian:

- Both kernel-level partial-write regression tests pass.
- The complete `internal/daemon/execenv` package passes.

On Windows:

```powershell
go test ./internal/daemon/execenv `
  -run "Test(WriteRuntimeConfigFile|InjectRuntimeConfigPreservesUserContent|CleanupRuntimeConfig)" `
  -count=1
```

The targeted runtime-config suite, including Windows DACL preservation, passes.

The Linux regression uses `RLIMIT_FSIZE` to produce a deterministic partial write after 64 bytes. It does not claim to reproduce an actual `ENOSPC` filesystem; it exercises the same destructive sequence relevant to the bug: the destination is opened and truncated, then the write fails before the full payload is stored.

## Checklist

## AI Disclosure

**AI tool used:** Codex Cli

**Prompt / approach:**
Please help me analyze whether Multica AI in the project directory causes destructive changes to user files during build and runtime. For example, does it silently fail to write to files that fail to write?

### DIFF
--- a/server/internal/daemon/execenv/runtime_config.go
+++ b/server/internal/daemon/execenv/runtime_config.go
@@ -236,10 +236,20 @@ func writeRuntimeConfigFile(path, brief string) error {
 
 	existing, err := os.ReadFile(path)
 	if errors.Is(err, fs.ErrNotExist) {
-		return os.WriteFile(path, []byte(block), 0o644)
+		// A dangling symlink also reports ErrNotExist through ReadFile. Do not
+		// replace that user-owned link with a regular file; there is no existing
+		// target we can atomically replace.
+		if info, lerr := os.Lstat(path); lerr == nil && info.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("runtime config %s is a dangling symlink", path)
+		}
+		return writeRuntimeConfigFileAtomic(path, []byte(block), 0o644)
 	}
 	if err != nil {
 		return fmt.Errorf("read existing runtime config %s: %w", path, err)
+	}
+	writePath, perm, err := existingRuntimeConfigWriteTarget(path)
+	if err != nil {
+		return err
 	}
 
 	existingStr := string(existing)
@@ -250,7 +260,7 @@ func writeRuntimeConfigFile(path, brief string) error {
 		// The managed separator (if any) lives in existingStr[:start] and
 		// is preserved untouched.
 		newContent := existingStr[:start] + block + existingStr[end:]
-		return os.WriteFile(path, []byte(newContent), 0o644)
+		return writeRuntimeConfigFileAtomic(writePath, []byte(newContent), perm)
 	}
 
 	// No marker block present. Append the fixed managed separator followed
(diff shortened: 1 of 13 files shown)


multica-eve (changes requested):
Before this change, a failed attempt to add temporary working instructions can leave a user's original instructions empty or only partly intact. This change aims to keep the original instructions untouched until the updated version is fully ready, with the same protection when restoring the project after a task.

**Product decision: accept the behavior.** Protecting user-owned instructions from partial-write corruption is the right product boundary, and the isolated `RLIMIT_FSIZE` helper fixes the earlier CI failure. However, the implementation still has three merge-blocking file-safety and compatibility regressions:

1. **Non-Windows replacement bypasses the original file's write permission and loses user-owned metadata.** `server/internal/daemon/execenv/runtime_config.go:278` carries forward only `Mode().Perm()`, then `server/internal/daemon/execenv/runtime_config_replace_other.go:7` replaces the directory entry with the temporary file. For example, an existing `0444` file in a writable parent directory previously failed safely at open time, but this implementation can replace it anyway. Ownership/group, ACLs, extended attributes, and hard-link identity are also lost and cannot be restored by cleanup. Please preserve the original write-permission gate and required metadata, and fail closed for file types or multiple-hard-link cases that cannot be replaced safely. Add regression coverage for read-only files and platform metadata.

2. **Creating a previously missing file no longer honors the process umask.** `server/internal/daemon/execenv/runtime_config.go:245` supplies `0644`, and `server/internal/daemon/execenv/runtime_config.go:333` applies it with `Chmod` after `CreateTemp`; `chmod` does not apply umask. Under `umask 077`, the previous `os.WriteFile(..., 0644)` path created a `0600` file, while this code creates `0644`, potentially exposing task/workspace instructions to other local users. Please preserve the old umask behavior or intentionally use a safe default no broader than `0600`, with a subprocess regression test for a restrictive umask.

3. **Replacing an existing file on Windows regresses long-path support.** `server/internal/daemon/execenv/runtime_config_replace_windows.go:43-52` passes ordinary paths directly to `ReplaceFileW`, bypassing the extended-length path normalization used by Go's `os` operations. Without a long-path-aware manifest, an existing config beyond `MAX_PATH` will fail during both injection and cleanup. Please convert the target, staged, and backup paths to extended-length absolute paths and add a Windows round-trip test whose existing target path exceeds 260 characters.

The prompt-prefix invariant remains intact: `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passed, as did the complete `internal/daemon/execenv` package and `git diff --check`.

**Recommendation: Request changes.** The product behavior is desirable, but the permission, metadata, umask, and Windows path regressions need to be addressed before merge.

PR #6477, opened 2026-08-06
fix(chat): prevent duplicate floating chat starter prompt sends

Bbbtt04:
## Problem

Rapidly clicking a starter prompt twice in the floating chat could create two independent conversations and enqueue two agent tasks for the same prompt.

This is a functional duplication issue rather than a visual-only problem: both conversations are persisted and both tasks can run and consume resources.

### Before

The same starter prompt is clicked twice in quick succession. Two conversations with the same title are created and shown as running.

Before: two conversations created from two rapid clicks

## Root cause

Each starter-prompt click called the send flow immediately. While the first click was still creating its conversation, the second click could enter the same flow before the new active conversation had been published, so both calls created a conversation.

## Fix

- Debounce starter-prompt clicks for 300 ms on the trailing edge.
- Reset the timer whenever another starter prompt is clicked during that interval.
- Submit only the latest prompt after there have been no further clicks for 300 ms.
- Clear the pending timer when the floating chat unmounts.
- Keep normal composer sends and queued follow-up behavior unchanged.

This is a frontend-only change scoped to the floating chat.

### After

Two rapid clicks within the debounce interval now create one conversation and one running message.

After: one conversation created and one message running

The recording files are linked from the earlier reproduction commit and are not included in the final PR diff.

## Regression coverage

The component test uses controlled timers to click the same starter prompt twice, verifies that nothing is sent before 300 ms, and then verifies that conversation creation and message submission each occur exactly once.

## Validation

- `pnpm exec vitest run chat/components/chat-window.test.tsx --reporter=verbose`
- `pnpm --filter @multica/views test` — 310 test files, 3,645 tests passed
- `pnpm typecheck` — 6 packages passed
- `pnpm --filter @multica/views lint` — 0 errors (20 pre-existing warnings)
- `git diff --check`
- Manual before/after browser verification with two rapid clicks


Bbbtt04:
Closing as superseded by the conversation-starter flow now on `main` (#7261, #7573, #7592). Starter actions now prefill and focus the composer instead of sending immediately, so the duplicate automatic-send path fixed by this PR no longer exists. The current regression coverage passes in `@multica/views` (53 tests across `chat-input.test.tsx` and `chat-empty-state.test.tsx`).

PR #6479, opened 2026-08-06
fix(agents): accumulate cache-read tokens in the kimi ACP backend

### LINKED ISSUE #6448: [Bug]: kimi backend drops cache token usage: only InputTokens/OutputTokens accumulated from the ACP prompt result
## Summary

Task usage for kimi runs never carries `cache_read_tokens` / `cache_write_tokens`. The kimi backend accumulates only `InputTokens` and `OutputTokens` from the shared ACP prompt result, so any cache usage the CLI reports is dropped before it reaches `TaskUsageEntry`. Sibling ACP backends (hermes, traecli) accumulate the cache fields from the same struct; kimi appears to be the only one missing it, similar in shape to #6308 ("kimi.go was the only ACP backend missing the streamingCurrentTurn gate").

## Where

At `a23b936b` (current main), `server/pkg/agent/kimi.go:375-376`:

```go
c.usage.InputTokens += pr.usage.InputTokens
c.usage.OutputTokens += pr.usage.OutputTokens
```

`pr` is a `hermesPromptResult`, and `parseACPTokenUsage` already parses `cachedReadTokens` / `cacheWriteTokens` (and several aliases) into `TokenUsage.CacheReadTokens` / `CacheWriteTokens`. The hermes and traecli backends accumulate those two fields at the equivalent point; kimi does not, so the parsed values are discarded. The later guard in kimi.go (`u.CacheReadTokens > 0 || u.CacheWriteTokens > 0`) can never be true on this path.

## Observed effect

Measured on a real workspace (Multica v0.4.18, daemon commit b312e1cbb, macOS arm64, kimi CLI running kimi-k3):

- The kimi CLI's own session log (`~/.kimi-code/sessions/.../agents/main/wire.jsonl`) records per-call usage for one reviewer session as: 24 calls, `inputOther` 79,649, `inputCacheRead` 1,548,800, `inputCacheCreation` 0, `output` 13,071.
- The board usage for the same runs shows roughly 173k plain input tokens and zero in both cache columns. About 1.5M cache-read tokens are invisible.

Cost impact: kimi-k3 prices cache hits at $0.30/M vs $3.00/M for misses (10x). Without the split, any cost accounting on kimi lanes is off by roughly an order of magnitude on the input side.

## Possible second gap

The CLI's local wire log names the fields `inputOther` / `inputCacheRead` / `inputCacheCreation`. I have not captured what the kimi CLI sends in the ACP `prompt` response `usage` payload. If it uses the same names there instead of `cachedReadTokens`, the alias list in `parseACPTokenUsage` would also need extending, and `inputOther` vs total-input semantics would need mapping. Worth capturing a live ACP frame while fixing.

## Suggested fix

Accumulate `CacheReadTokens` / `CacheWriteTokens` in kimi.go the way hermes and traecli do, and extend the `parseACPTokenUsage` alias list if the live frame shows kimi-specific field names.

pengfeixx:
## Summary

The kimi ACP backend dropped `CacheReadTokens` from every `session/prompt` result. At the prompt-merge point `server/pkg/agent/kimi.go` accumulated only `InputTokens`/`OutputTokens`, even though `parseACPTokenUsage` already parses `cachedReadTokens` (and its aliases) and the sibling hermes and traecli backends accumulate the field at the same point. The downstream guard on `CacheReadTokens` could therefore never be true on this path, so kimi cache reads never reached the usage/cost dashboard — the reporter measured ~1.5M cache-read tokens invisible on a single reviewer session, which also throws kimi cost accounting off by roughly an order of magnitude on the input side.

## Change

- Accumulate `CacheReadTokens` from the prompt result in `kimi.go`, matching `hermes.go` and `traecli.go` line-for-line at this merge point.
- Add `TestKimiBackendAccumulatesCacheReadTokens`, which drives the backend with a fake `kimi acp` that reports `cachedReadTokens` in the prompt response and asserts the value reaches `Result.Usage`. The test fails without the fix (`cacheReadTokens: got 0, want 500`) and passes with it.

`parseACPTokenUsage` is left unchanged: it already covers `cachedReadTokens`, `cacheReadTokens`, `cached_input_tokens`, `cache_read_tokens`, and `cache_read_input_tokens`, and no live ACP frame was captured showing a kimi-specific name it does not already parse, so extending the alias list would be speculative.

Fixes #6448

### DIFF
--- a/server/pkg/agent/kimi.go
+++ b/server/pkg/agent/kimi.go
@@ -374,6 +374,7 @@ func (b *kimiBackend) Execute(ctx context.Context, prompt string, opts ExecOptio
 				c.usageMu.Lock()
 				c.usage.InputTokens += pr.usage.InputTokens
 				c.usage.OutputTokens += pr.usage.OutputTokens
+				c.usage.CacheReadTokens += pr.usage.CacheReadTokens
 				c.usageMu.Unlock()
 			default:
 			}
--- a/server/pkg/agent/kimi_test.go
+++ b/server/pkg/agent/kimi_test.go
@@ -495,3 +495,77 @@ func TestKimiBackendDropsHistoryReplayOnResume(t *testing.T) {
 		}
 	}
 }
+
+// fakeKimiACPUsageScript impersonates `kimi acp` for a single turn and
+// reports cache usage in the session/prompt result, the way a real kimi
+// turn that hits a prompt cache does. The backend must carry
+// cachedReadTokens through into Result.Usage the same way the hermes and
+// traecli backends do; before the fix kimi.go accumulated only
+// InputTokens/OutputTokens and dropped the cache field (see #6448).
+func fakeKimiACPUsageScript() string {
+	return `#!/bin/sh
+while IFS= read -r line; do
+  id=$(printf '%s' "$line" | sed -n 's/.*"id":\([0-9]*\).*/\1/p')
+  case "$line" in
+    *'"method":"initialize"'*)
+      printf '{"jsonrpc":"2.0","id":%s,"result":{"protocolVersion":1,"agentCapabilities":{}}}\n' "$id"
+      ;;
+    *'"method":"session/new"'*)
+      printf '{"jsonrpc":"2.0","id":%s,"result":{"sessionId":"ses_fake"}}\n' "$id"
+      ;;
+    *'"method":"session/prompt"'*)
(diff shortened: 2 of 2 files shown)


multica-eve:
Closing as superseded by #6495, which fixes current Kimi releases by reading their session wire logs. Current Kimi ACP responses do not expose billing token usage, so this prompt-result-only change would not affect real runs. Thanks for the investigation and patch—the field-mapping work helped lead us to the working data source.

PR #6484, opened 2026-08-06
refactor(daemon): compress the agent-trigger warning to its per-turn fact (MUL-5442)

Bohan-J:
Pool-2 item 1 from the future-todos analysis. Compresses the per-turn agent-trigger warning block and migrates the work-category definition it relied on into the cached brief. Four files: the per-turn string (`prompt.go`), the brief canonical (`runtime_config_sections.go`), and two new pin sets (`daemon_test.go`, `execenv_test.go`).

## Measured

| | bytes |
|---|---:|
| agent-trigger warning block (trailing newlines excluded) | 728 → **390 (−338 per agent-triggered turn)** |
| cached brief (real-UUID fixture) | 12,701 → **12,767 (+66 one-time)** |

Net uncached-byte win over N agent-triggered turns: `338×N − 66`. The block rides the uncacheable per-turn channel and fires on every agent-to-agent turn — collaboration-heavy issues (like MUL-5442 itself) pay it dozens of times per session. (Review history: the initial compressed commit measured 387; the conjunction-scoping rewording from review added 3.)

## The judgment split

The per-turn FACT is only which trigger this is (another agent). The full discipline lives — and is pinned — in the brief: the reply-warranted rule with both outcome arms (Reply-mode block), the 'No reply needed' noise ban (same), and the sign-off-mention ban (Mentions). The compressed block states the fact, both outcome arms in short form, and points at the brief.

One semantic did NOT originally live in the brief: the work-category boundary (investigation-only and answer-only turns count as deliverable work) existed solely in the deleted parenthetical — review caught the dangling pointer. It now lives in the brief's canonical reply-warranted line ("investigation, a fix, or an answer to a real question all count"), the +66 above, so the per-turn channel never pays for it again.

## Verification

`go test ./internal/daemon/... -count=1` green · `go build` clean. Test changes (from review, not zero): `execenv_test.go` pins the category clause in the anti-loop block; `daemon_test.go` pins the compressed block's shape as sequential relations — the canonical-pointer sentence, the work arm with its consequence, and the no-work arm with its consequence. Prior anti-loop anchors ('Silence is the preferred way', 'do not @mention the other agent as a sign-off') survive verbatim; the member-trigger leak guard is untouched.

### DIFF
--- a/server/internal/daemon/daemon_test.go
+++ b/server/internal/daemon/daemon_test.go
@@ -1078,6 +1078,12 @@ func TestBuildPromptCommentTriggeredByAgent(t *testing.T) {
 
 	for _, want := range []string{
 		"Another agent (Atlas)",
+		// MUL-5442 #6484 review: the compressed block's SHAPE is the
+		// behaviour — pin the canonical pointer and both outcome arms as
+		// sequential relations, not just the closing anchors.
+		"Apply the reply-warranted rule and mention discipline from your brief",
+		"produced real work → post the result as a normal reply",
+		"no work and only an acknowledgment / thanks / sign-off → exit with no output",
 		"do not @mention the other agent as a sign-off",
 		"Silence is the preferred way",
 	} {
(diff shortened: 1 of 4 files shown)


Bohan-J:
Superseded by the follow-up PR: owner decision on MUL-5442 removes the generic no-reply rule entirely, deleting the block this PR compressed. Closing unmerged.

PR #6490, opened 2026-08-06
MUL-5783: make upload limits configurable end to end

### LINKED ISSUE #6461: [Feature]: Configurable file upload size limit (server + web proxy)
### Deployment type

self-host

### What do you want and why?

**Deployment type:** self-host

## What do you want and why?

Self-hosted Multica deployments currently cannot upload files larger than 100 MB, and the limit is hardcoded in three places with no configuration knobs:

1. **Backend** — `server/internal/handler/file.go:34`:
   ```go
   const maxUploadSize = 100 << 20 // 100 MB
   ```
   Used by both `http.MaxBytesReader` and `r.ParseMultipartForm` in `UploadFile`. There is no environment variable to override it.

2. **Web proxy** — Next.js 16 `proxyClientMaxBodySize` defaults to `10 MB` (10485760). The web app proxies `/api/*` to the backend via `proxy.ts` (runtime rewrites to `REMOTE_API_URL`). Any upload over 10 MB is truncated at the proxy *before* reaching the backend, producing a corrupt multipart body, an `ECONNRESET`, and a 500 in the UI:
   ```
   Request body exceeded 10MB for /api/upload-file. Only the first 10MB will be available...
   Failed to proxy http://multica-backend:8080/api/upload-file Error: socket hang up (ECONNRESET)
   ```

3. **Client guards** — `packages/core/constants/upload.ts` and `apps/mobile/data/api.ts` both hardcode `MAX_FILE_SIZE = 100 * 1024 * 1024`, and `server/internal/cli/client.go` hardcodes `maxDownloadSize = 100 << 20`.

## Proposed solution

Make the upload size limit configurable end-to-end for self-hosted operators:

- **Backend:** read `maxUploadSize` from an env var (e.g. `MULTICA_MAX_UPLOAD_SIZE`, accepting bytes or human units like `512mb`), falling back to the current `100 << 20`. Wire it through the handler (and ideally the CLI `DownloadFile` cap).
- **Web proxy:** set `experimental.proxyClientMaxBodySize` in `apps/web/next.config.ts` from the same or a separate env var, or raise the default so a `NEXT_PUBLIC_API_URL`-less deployment isn't silently capped at 10 MB. If the var is unset, match the backend limit.
- **Client guards:** drive `MAX_FILE_SIZE` from a server-announced/configurable value instead of a hardcoded constant (or at least document the four places an operator must change).

### Notes for maintainers

- A self-hosted operator can today work around the *proxy* limit by setting `NEXT_PUBLIC_API_URL` to the backend origin so the browser calls `/api/*` directly, but the *backend* 100 MB cap still requires rebuilding the image. Making both configurable would remove the need to rebuild for a size change.
- Repro: upload a file between 10 MB and 100 MB on a self-hosted instance with the default proxy config → 500 in UI, `Request body exceeded 10MB` in frontend logs.

_No response_

_No response_

multica-eve:
## Summary

- route `/api/upload-file` through a dedicated streaming App Router handler so Next Proxy no longer truncates self-hosted uploads at 10 MiB
- add `MULTICA_MAX_UPLOAD_SIZE` (default 100 MiB, hard maximum 1 GiB), expose the effective limit through `/api/config`, stream parsed files into storage, authorize workspace membership before reading the body, and return structured 413 responses
- make Web and Mobile preflight checks consume the server capability, preserve structured upload errors, and make CLI attachment downloads stream without silent 100 MiB truncation
- wire the setting into Docker Compose and Helm defaults and document reverse-proxy/ingress body-limit requirements

## Safety and compatibility

- no database or persisted-format migration
- no global increase to Next's proxy buffering limit
- multipart memory stays fixed at 8 MiB; larger files spool and then stream to storage
- invalid configuration logs a warning and safely falls back to 100 MiB
- older servers and clients retain the previous 100 MiB behavior

## Validation

- `pnpm --filter @multica/web build`
- Web test suite; focused Core upload/config tests (161); Mobile test suite (55)
- Core, Views, Web, Mobile, and Docs typechecks
- Core, Views, Web, and Mobile lint (pre-existing warnings only)
- focused Go handler/CLI tests and `go test ./... -run '^$'`
- `docker compose -f docker-compose.selfhost.yml config --quiet`
- Helm `values.yaml` parsed successfully (the local environment does not have the `helm` binary)

Local full-suite notes: the Core suite has eight unrelated Node 26 `localStorage` failures, and the backend integration suite is blocked by an existing local test database missing current migration columns. All tests covering this change pass.

Fixes #6461
Closes MUL-5783

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -59,6 +59,10 @@ MULTICA_APP_URL=${FRONTEND_ORIGIN}
 # avoid Host / X-Forwarded-Host spoofing when a self-hosted reverse proxy
 # is not hardened.
 MULTICA_PUBLIC_URL=
+# Maximum size of one uploaded file. Accepts bytes or KB/MB/GB and
+# KiB/MiB/GiB suffixes, up to 1 GiB. Clients discover the effective value
+# from /api/config after the backend restarts.
+MULTICA_MAX_UPLOAD_SIZE=100MiB
 # Optional delay after SIGTERM/SIGINT before the existing graceful shutdown
 # begins. Accepts a non-negative Go duration such as 300s or 5m. Empty or 0
 # preserves the default behavior and starts shutdown immediately. The platform
--- a/SELF_HOSTING_ADVANCED.md
+++ b/SELF_HOSTING_ADVANCED.md
@@ -93,6 +93,7 @@ For file uploads and attachments, configure S3 and (optionally) CloudFront:
 | `S3_BUCKET` | Bucket name only (e.g. `my-bucket`). Do **not** include the `.s3.<region>.amazonaws.com` suffix — the server constructs the public URL from `S3_BUCKET` + `S3_REGION` |
 | `S3_REGION` | AWS region (default: `us-west-2`). Must match the bucket's actual region — used for both SDK signing and public URLs |
 | `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` | Static credentials. When both are unset, the AWS SDK default credential chain is used |
+| `MULTICA_MAX_UPLOAD_SIZE` | Maximum size of one file (default `100MiB`, maximum `1GiB`). Accepts bytes or `KB`/`MB`/`GB` and `KiB`/`MiB`/`GiB` suffixes. Web and mobile clients discover the effective value at runtime. |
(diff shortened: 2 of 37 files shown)


Loger0:
@multica-eve 麻烦可以尽快合入这个mr吗

oYo-Byte:
@multica-eve frontend-test失败只差一个字符，麻烦处理下

PR #6510, opened 2026-08-06
ERI-126: suppress parent orchestrator executions

rongchenzheng:
## Summary

- treat `workflow_role=parent_orchestrator` plus `execution_expected=false` as an aggregation-only issue contract
- suppress assignment/status, comment, Stage-completion wake, rerun, retry, and queue-lifecycle executions while keeping normal Stage children runnable
- return the stable `execution_suppressed` outcome for explicit mentions and manual reruns, with localized UI copy
- enforce current metadata at service and SQL boundaries, including stale queued/deferred/dispatched rows

## Testing

- `DATABASE_URL='postgres:///multica_eri126?sslmode=disable' go test ./internal/util ./internal/service ./internal/handler`
- `go test ./internal/service -run 'TestBuiltinSkillsConformToTemplate|TestBuiltinSkillsFrontmatterIsStrictYAML|TestWorkingOnIssuesSkillCoversIssueLoopContracts' -count=1`
- `npx --yes pnpm@10.28.2 --filter @multica/views typecheck`
- `npx --yes pnpm@10.28.2 --filter @multica/views exec vitest run issues/blocked-trigger-copy.test.ts`
- locale JSON validation for `en`, `zh-Hans`, `ja`, and `ko`
- `make sqlc`
- `git diff --check`

Issue: ERI-126


Bohan-J:
Thanks for the thorough work here — the implementation itself is careful (the predicate is centralized, the SQL-level re-check for already-queued rows is a real edge case to have thought about, and the test coverage is substantial). But I don't think we can merge this as-is, and the reasons are about the design rather than the code.

### 1. There's no problem statement I can trace

The PR body describes what the change does but not what broke. There's no linked issue and no closing reference, and `ERI-126` doesn't resolve to anything in this repo. That's ~1200 lines of behavior change across the trigger and queue lifecycle with no recorded motivation, which makes it hard to evaluate the tradeoffs or to know what regression a future change would be at risk of reintroducing.

### 2. "Parent aggregates, doesn't execute" is already expressible — leave the parent unassigned

Today a parent with no agent assignee is already an inert aggregation container:

- `dispatchParentAssigneeTrigger` returns immediately when `parent.AssigneeType`/`AssigneeID` are not valid, so stage completion wakes nobody
- `routeAssigneeFallback` returns early on the same check
- assignment enqueue and `RerunIssue` both need a target agent

That covers the stated use case with no new code. The only thing the new contract adds is a parent that *has* an assignee but never runs — which is a narrower and stranger thing to want, and the PR doesn't say who wants it.

### 3. This conflicts with how stage advancement works

The parent wake isn't incidental — it *is* the stage advancement mechanism. From `multica-working-on-issues/SKILL.md` on main:

> Advancement is agent-driven: the server only detects the closed barrier and wakes the parent assignee, who then decides whether to promote the next stage's `backlog` sub-issues to `todo`.

And `issue_child_done.go` carries the history of the two times we fixed a parent *not* being woken (MUL-2808, MUL-3969 — "the parent silently stalled in `in_progress`").

With suppression on, when stage 1 closes nobody promotes stage 2 out of `backlog`. The PR states that stage children stay runnable, but in a staged layout they'd stay parked indefinitely. The change deliberately keeps posting the "Stage N complete" system comment to the parent as aggregation evidence — but that comment now has no reader. I think the feature and the stage model contradict each other, and that needs resolving before the behavior lands.

### 4. Reserved metadata keys are a big contract change

`SKILL.md` currently states that metadata is a free-form bag and that the platform curates no vocabulary. This carves out the first exception, and its effect is to silently disable *all* execution on an issue.

The part that worries me most: metadata is written by agents themselves, and that's exactly what the skill teaches. After this, an agent writing the wrong key pair can permanently disable execution on an issue — including on the issue it is running on — with no indicator anywhere in the UI. The state is only observable by attempting a trigger and reading the block reason. The predicate also accepts the string `"false"`, so it's reachable through the plain string metadata path, not just a typed boolean.

If we want a "this issue doesn't execute" capability, I'd argue it should be a first-class, visible field on the issue (queryable via the API, rendered in the issue view), not a reserved key inside a bag documented as having no platform semantics.

### What would unblock this

1. What concretely broke that leaving the parent unassigned doesn't solve?
2. With the parent never woken, what promotes stage 2 out of `backlog`?
3. If the capability is genuinely needed, can it be a visible first-class field instead of reserved metadata keys?

Happy to look again once there's a problem statement — and to be clear, the objection is to the mechanism and its interaction with stages, not to the quality of the implementation.

PR #6521, opened 2026-08-06
fix(runtimes): document QwenPaw and close its frontend drifts (MUL-5828)

Bohan-J:
## Summary

Audit of the two recently merged runtimes found that **Reasonix was documented everywhere, but QwenPaw had reached only the two READMEs** — it was missing from every other surface that enumerates runtimes. `SupportedTypes` has 20 entries; the docs tables had 19.

### 1. Document QwenPaw

| Surface | Change |
| --- | --- |
| `providers` (en/zh/ja/ko) | Capability matrix row + the model-selection caveat below |
| `install-agent-runtime` (en/zh/ja/ko) | Install table row + a note to configure the model in QwenPaw itself |
| `SELF_HOSTING.md` | CLI list entry |
| `CLI_AND_DAEMON.md` | Supported Agents row, `MULTICA_QWENPAW_PATH` / `_ARGS`, and the launch contract (`qwenpaw acp --workspace`, `AGENTS.md`, skills + `skill.json`). No `_MODEL`: the model is owned by QwenPaw's own configuration — see section 4 |
| `environment-variables` (en/zh/ja/ko) | `_ARGS` covers **five** tools now that `MULTICA_QWENPAW_ARGS` exists (`config.go:247`), not four. Also carries the `_MODEL` exception from section 4 |
| landing i18n (en/zh/ja/ko) | The "15 supported tools" list named Reasonix but not QwenPaw — it was actually missing DevEco Code, Grok, Qoder CN, Qwen Code, and QwenPaw. Now 20. |

QwenPaw is the only runtime where `ModelSelectionSupported()` returns `false` (`server/pkg/agent/models.go:244`) — its `session/set_model` writes to the shared, persistent config rather than the session. `providers` previously told users an empty model list means the runtime is offline or signed out, which is misleading here. Now documented as expected behaviour.

Also moves the `Reasonix` row back into runtime-id order in the two docs tables, where it had been inserted between Kimi and Kiro.

### 2. Close the QwenPaw code drifts

Each of these let `qwenpaw` through on the Go/DB side but dropped it on the frontend side:

- **`mcp-support.ts`** was missing `qwenpaw`, so `agent-overview-pane.tsx:186` hid the MCP tab — even though `qwenpaw.go:63` forwards `opts.McpConfig`. Without this, the new "Multica-managed MCP: yes" row would have been false.
- **`RUNTIME_PROFILE_PROTOCOL_FAMILIES`** was missing it, so the custom runtime profile picker could not select QwenPaw, though migrations 253/254 and `agent.SupportedTypes` both allow it.
- **`PROVIDER_DISPLAY_NAMES`** was missing it. The daemon's `runtimeDisplayNameOverrides` (`daemon.go:5080`) maps `qwenpaw` → `QwenPaw`; without the mirror an aliased runtime rendered `Qwenpaw` — precisely the drift the comment on that map warns about.
- **`provider-logo.tsx`** had no `qwenpaw` case and fell through to the generic `Monitor` icon. Adds the standalone mark lifted from the official wordmark (`agentscope-ai/QwenPaw`, `console/public/logo-light.svg`, Apache-2.0), drawn with `currentColor` so a single path covers both themes rather than vendoring upstream's separate light/dark files.

Test assertions added for each.

### 3. Correct two stale runtime lists

- `CLI_AND_DAEMON.md` listed a `Gemini` / `gemini` runtime plus `MULTICA_GEMINI_PATH` / `_MODEL`. Migration `126_runtime_profile_drop_gemini` removed the protocol family and no `gemini` probe exists in `agents_probe.go`, so all three are removed.
- `CLI_AND_DAEMON.md` and `SELF_HOSTING.md` were both missing Antigravity, CodeBuddy, and DevEco Code. Added, along with their `PATH` / `MODEL` / `ARGS` variables.

## Verification

- `pnpm typecheck` — 6/6 pass
- `pnpm --filter @multica/core test` — 1305 pass
- `pnpm --filter @multica/views test` — 3663 pass
- `pnpm --filter @multica/core --filter @multica/views --filter @multica/web lint` — 0 errors (the remaining warnings reproduce identically on a clean `main`)
- Extracted logo mark rasterised and eyeballed; bbox is `x[0,123] y[8,132]` against a `-4.5 4 132 132` viewBox, so it is centred with even padding and cannot clip
- Every documented fact re-checked against source: `agent.go:252`, `agents_probe.go`, `qwenpaw.go`, `qwenpaw_workspace.go`, `local_skills.go:200`, `runtime_config.go:205`, `daemon.go:5075`, migrations 126 / 253 / 254
- `pnpm --filter @multica/docs build` fails prerendering `/404` (`<Html> should not be imported outside of pages/_document`). Verified identical on a clean `main` — pre-existing, unrelated to this PR.

### 4. Two env vars the docs promised but the runtime ignored

Found in review, both confirmed against source:

- **`MULTICA_QWENPAW_ARGS` never reached the command line.** `config.go:247` read it and `daemon.go` forwarded it as `ExecOptions.ExtraArgs`, but `qwenpaw.go` built its argv from `CustomArgs` only — plumbed the whole way, then dropped. It now consumes `ExtraArgs` before `CustomArgs`, matching the precedence `CLI_AND_DAEMON.md` documents and the five other backends that accept both. `TestQwenpawExtraArgsReachTheCommandLine` pins presence and ordering; on the old backend it fails with `acp --per-agent`.
- **`MULTICA_QWENPAW_MODEL` could not work by design.** The backend never calls `session/set_model` (it would rewrite QwenPaw's shared agent config), so `ExecOptions.Model` is ignored — pinned by `TestQwenpawUsageModelIgnored`. `AgentEntry.Model` feeds only that ignored field, so dropping the probe read is behaviour-neutral. The variable is no longer read or documented, and the four `environment-variables` pages now carry the exception, since the blanket "every tool accepts `_MODEL`" line was the misleading part.

Also drops the stale enumeration on `ExecOptions.ExtraArgs` — it claimed "claude and codex backends only", already wrong for CodeBuddy, Antigravity, and Qwen Code, and that staleness is what let the bug above ship.

## Note for reviewers

`RUNTIME_PROFILE_PROTOCOL_FAMILIES`, `PROVIDER_DISPLAY_NAMES`, and `MCP_SUPPORTED_PROVIDERS` are each hand-mirrored from Go with only a comment holding them in sync — which is how all three drifted at once when QwenPaw landed. Worth a generated file or a cross-language guard at some point; out of scope here.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -282,7 +282,6 @@ Agent-specific overrides:
 | `MULTICA_QWEN_MODEL` | Override the Qwen Code model used |
 | `MULTICA_QWEN_ARGS` | Daemon-wide extra Qwen arguments (POSIX shellword parsing; managed protocol flags are filtered) |
 | `MULTICA_QWENPAW_PATH` | Custom path to the `qwenpaw` binary |
-| `MULTICA_QWENPAW_MODEL` | Seeds the daemon-wide default model. QwenPaw does not accept a per-agent model override — its `session/set_model` writes to QwenPaw's shared, persistent config rather than the session, so Multica leaves model selection to QwenPaw itself |
 | `MULTICA_QWENPAW_ARGS` | Daemon-wide extra QwenPaw arguments (POSIX shellword parsing; managed protocol flags are filtered) |
 
 If a previously generated `~/.multica/hooks` wrapper is first on `PATH` and calls the same command name again, the daemon skips that hooks directory during built-in agent discovery and records the real binary path behind it. If your interactive shell still recurses when you run `claude`, `codex`, or `hermes` manually, remove the hooks entry from your shell startup file or replace the wrapper body with an absolute `exec /path/to/real-binary "$@"`.
@@ -291,7 +290,7 @@ The daemon launches Qoder and Qoder CN as `qodercli --yolo --acp` and `qoderclic
 The daemon launches Qwen Code as `qwen -p <prompt> --output-format stream-json`. It writes the task brief to `QWEN.md`; when an agent has managed `mcp_config`, the daemon writes a 0600 per-run JSON file and passes it through `--mcp-config <path>`, then removes it after the process exits. A null config preserves Qwen Code native MCP settings.
 
 
-The daemon launches QwenPaw as `qwenpaw acp --workspace <per-task dir>`. It writes the task brief to `AGENTS.md`, and materialises the run's bound skills into `<per-task dir>/skills/` plus a `skill.json` manifest, so QwenPaw discovers them through its own workspace skill discovery. `acp` and `--workspace` are reserved: `custom_args` cannot override them.
+The daemon launches QwenPaw as `qwenpaw acp --workspace <per-task dir>`. It writes the task brief to `AGENTS.md`, and materialises the run's bound skills into `<per-task dir>/skills/` plus a `skill.json` manifest, so QwenPaw discovers them through its own workspace skill discovery. `acp` and `--workspace` are reserved: `custom_args` cannot override them. QwenPaw is the one runtime with no `MULTICA_QWENPAW_MODEL`: its `session/set_model` writes to a shared, persistent agent config rather than the session, so Multica never sends it a model and leaves that choice to QwenPaw's own configuration.
 
 `MULTICA_CLAUDE_ARGS`, `MULTICA_CODEX_ARGS`, `MULTICA_CODEBUDDY_ARGS`, `MULTICA_QWEN_ARGS`, and `MULTICA_QWENPAW_ARGS` are parsed with POSIX shellword quoting, so values such as `--model "gpt-5.1 codex" --sandbox read-only` are split like a shell command line. Agent arguments are applied in this order: hardcoded Multica defaults, daemon-wide env defaults, then per-agent `custom_args` from the task.
 
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -167,12 +167,14 @@ The daemon auto-detects these AI CLIs on your PATH:
 | CLI | Command | Description |
 |-----|---------|-------------|
 | [Claude Code](https://docs.anthropic.com/en/docs/claude-code) | `claude` | Anthropic's coding agent |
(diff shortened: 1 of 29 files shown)


Bohan-J:
Both must-fixes are addressed in `0840def6`. I verified each claim against the source before acting — both were real.

**1. `MULTICA_QWENPAW_ARGS` was dropped.** Confirmed: `ExtraArgs` is consumed by `claude`, `codex`, `codebuddy`, `antigravity`, and `qwen`, but `qwenpaw.go` built its argv from `CustomArgs` only. `config.go:247` reads the variable and `defaultArgsForProvider` maps `qwenpaw → cfg.QwenpawArgs`, so the plumbing was already there and the backend just never read the last hop.

I took the "make it work" branch rather than deleting the variable: the knob was added deliberately in the QwenPaw PR with the established name, and `ExtraArgs → CustomArgs` is the precedence `CLI_AND_DAEMON.md` already documents. `TestQwenpawExtraArgsReachTheCommandLine` pins both presence and ordering; on the old backend it fails with `acp --per-agent`, so it is not vacuous.

**2. `MULTICA_QWENPAW_MODEL` was inert.** Confirmed, and it cannot be made to work without upstream session-scoped `set_model` — the current refusal is the right call. I traced `AgentEntry.Model`: it flows only into `ExecOptions.Model`, which this backend ignores, so removing the probe read is behaviour-neutral rather than a functional change. Done:

- `agents_probe.go` no longer reads it; the comment now says why instead of claiming it "seeds the daemon-wide default"
- the `MULTICA_QWENPAW_MODEL` row is gone from `CLI_AND_DAEMON.md`, replaced by a sentence explaining the exception
- all four `environment-variables` pages carry the caveat, since the blanket "every tool accepts `_MODEL`" line was the misleading part

One correction to the suggested wording: I dropped the `_MODEL` row entirely rather than documenting it as an exception in the table. A variable that provably does nothing is better absent than described — the exception belongs on the prose line that made the general promise.

Verification: `go test ./pkg/agent/...` and `./internal/daemon/...` pass, `gofmt`/`go vet` clean. Frontend is untouched this round; `packages/core` (1305) and `packages/views` (3663) passed on the previous commit.

Also worth noting for a follow-up, not fixed here: `agent.go:72` still comments that `ExtraArgs` is "currently read by claude and codex backends only", which was already wrong for `codebuddy`, `antigravity`, and `qwen` before this change.

PR #3993, opened 2026-06-10
feat(wecom): WeCom intelligent robot integration

QQQJoker:
## Summary

- Add WeCom (企业微信) intelligent robot integration using API long-connection mode
- Backend: Hub, WS connector, Dispatcher, binding tokens, chat sessions, outbound stream patcher
- Frontend: Settings WeCom tab, /wecom/bind page, agent integrations status
- DB migrations 117 and 118, self-host docs and docker-compose env wiring
- Fix binding token TTL clock skew (Lark + WeCom)

## Test plan

- [x] go test ./internal/integrations/wecom/... passes locally
- [x] E2E validated: bind link, member redeem, DM + group @, agent stream reply
- [ ] Upstream CI make check
- [ ] Self-host with MULTICA_WECOM_SECRET_KEY + MULTICA_PUBLIC_URL

## AI Disclosure

AI tool: Cursor. Implemented mirroring Lark integration boundaries with local self-host validation.

PR #6524, opened 2026-08-06
Feature: add WeCom bot integration

### LINKED ISSUE #6496: [Feature]: Support WeCom Smart Robot
### Deployment type

self-host

### What do you want and why?

Currently multica only support Lark and Slack bot.

Add WeCom smart robot support. Document: https://developer.work.weixin.qq.com/document/path/101463
WeCom smart robot also support scan QR to quick create robot.

_No response_

wzxjohn:
## What does this PR do?

Adds WeCom (企业微信) as an agent channel, so an agent can be @-mentioned in a WeCom group or DM'd
directly and answer as itself — the same product semantics we already ship for Lark, Slack, and
DingTalk.

WeCom smart bots have no inbound webhook. The server dials an outbound long connection and holds it,
receiving encrypted frames. That drives the design:

- Connection ownership goes through the existing `channel/engine` supervisor, so exactly one instance
  holds a given bot's socket and rotation is already solved.
- Replies land on a new **generic** `channel_outbound_*` queue rather than a WeCom-specific table,
  because WeCom has no synchronous reply-to-webhook affordance so replies must be durably queued and
  retried. Rows are keyed by `channel_type`, so Lark and Slack can move onto it later.
- Install has two paths. Scan-code provisions a brand-new bot and needs `MULTICA_WECOM_SOURCE_ID`
  plus an HTTP provider — which left a hole: a bot already in the WeCom console couldn't be connected
  at all, and a deployment with no scan source saw no bind entry point. Manual Bot ID + Secret entry
  needs only the secretbox key.

Manual install is a synchronous `POST`, not the session/poll shape scanning uses: there's no
third-party round trip to keep off the request path, so the caller gets the real outcome — including
"that secret is wrong" — instead of a session id to poll. Credentials are probed before anything is
written (`aibot_subscribe` handshake only, reusing `Conn.subscribe` so the probe can't drift from
production frame encoding). Without the probe a typo would produce an installation whose bot silently
never answers, and `channel_installation` has no health column to surface it.

The whole stack is gated on `MULTICA_WECOM_SECRET_KEY` — a no-op for any deployment that doesn't opt in.

## Related Issue

Closes #6496

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- **`server/internal/integrations/wecom/`** (new): long-connection client and frame codec
  (`ws_conn.go`), credential/payload crypto (`crypto.go`, `media.go`), scan install
  (`install.go`, `install_worker.go`, `install_provider.go`), credential probe (`verify.go`),
  account binding (`binding.go`), welcome (`welcome.go`), delivery (`inbound.go`, `outbound.go`,
  `outbound_wake.go`), plus `resolvers.go`, `retry.go`, `store.go`.
- **`server/internal/handler/wecom.go`** (new): workspace-scoped `wecom/installations`,
  `wecom/install/begin`, `wecom/install/manual`, `wecom/install/{sessionId}/status`,
  `DELETE wecom/installations/{installationId}`, and global `POST /api/wecom/binding/redeem`.
- **`server/cmd/server/router.go`, `main.go`, `handler/handler.go`**: channel registration,
  supervisor wiring, and `RATE_LIMIT_WECOM_MANUAL_INSTALL` (per-IP, default 10/min) on top of auth
  and `canManageAgent`.
- **`server/internal/integrations/channel/engine/supervisor.go`**: lease support for the new channel.
- **`server/internal/handler/chat.go`**: archiving or deleting a chat session now drains its
  outbound queue rows.
- **Migrations `261`–`275`**: `wecom_install_session` + indexes; `channel_outbound_queue`,
  `channel_outbound_send_attempt`, `channel_outbound_reconcile_state` + indexes;
  `271_chat_message_task_assistant_idx` for outbound reply lookup; `275_issue_origin_wecom_chat`
  extends the `issue.origin` constraint. Plus `pkg/db/queries/` and regenerated `pkg/db/generated/`.
- **`server/internal/handler/workspace_delete_manifest_test.go`**: ownership decisions for the four
  new tables (rest of that diff is gofmt realignment).
- **`server/internal/metrics/`**: `wecom.go` connection/delivery metrics, outbound-queue gauges in
  `business_sampler*`, `NormalizeChannelType` allow-list and `installation_id` as a forbidden label
  in `labels.go`.
- **`packages/core/`**: `wecom/` queries, `types/wecom.ts`, `api/client.ts` + `api/schemas.ts` (all
  responses through `parseWithFallback` + zod, with malformed-response tests), and a
  `wecom_installation` realtime handler in `realtime/use-realtime-sync.ts`.
- **`packages/views/`**: `settings/components/wecom-tab.tsx` install dialog (opens on the QR code,
  secondary link to the Bot ID + Secret form), `wecom/bind-page.tsx`, and WeCom on the agent
  integrations tab and overview pane.
- **`apps/web/app/wecom/bind/page.tsx`**: platform wiring for the bind page.
- **Locales** for `en` / `ja` / `ko` / `zh-Hans`.
- **Docs**: new `wecom-bot-integration.{mdx,zh,ja,ko}`, plus `channels.*`,
  `environment-variables.*`, and `meta*.json`.

## How to Test

**Against a real bot** (needs a WeCom smart bot in the console). Set `MULTICA_WECOM_SECRET_KEY` to a
base64 32-byte key, leave `MULTICA_WECOM_SOURCE_ID` unset, then:

1. Agent → Integrations → connect WeCom → follow the secondary link to the Bot ID + Secret form.
2. Submit a **wrong** secret — expect an inline credentials error and no `channel_installation` row.
3. Submit the correct pair — expect success and the socket coming up in the log.
4. @-mention the bot in a group: agent replies, `enter_chat` produces the welcome, an image arrives
   as an attachment, and `/issue` creates an issue with `origin = 'wecom_chat'`.
5. Submit the same Bot ID from a second agent — expect a conflict message that distinguishes another
   workspace / another agent here / archived agent. Then disconnect and reconnect with the same Bot
   ID — expect success.

## Checklist

## AI Disclosure

**Prompt / approach:**
Add WeCom Smart Robot support like lark or slack. The document is on: https://developer.work.weixin.qq.com/document/path/101463

## Screenshots (optional)
a1b915d665830061aea2ad3b47ef94f3
77c523d131070595a0a77ac02859036f
0c247e11b57e7b89729340304d055b41
79b1f7f8ca90a7417775b9e58e68f5cc
81fd129cfe3f32674ac2a8323b8bc658


seacen:
Hi @wzxjohn — this is a serious piece of work. Before you put more time into it you should know there are two other open WeCom PRs, because none of us wants to discover that at merge time:

- **#5833** (@leroy-chen) — smart bot through the channel engine. Three review rounds with @Bohan-J, all blockers closed, CI green; @Bohan-J said today that "the code side is essentially settled". It has also been running in production in a corporate tenant for about ten days — mine — on top of a hardening branch.
- **#3993** — open since June.

I checked your branch out and read it rather than skimming the diff, and I'd rather be specific than diplomatic: **several things in yours are better than what #5833 and my hardening branch have**, and I'd like them to survive whatever the maintainers decide.

1. **The durable `channel_outbound_*` queue.** Claim/lease, backoff, dead-letter, retention purge — and keyed by `channel_type` rather than by WeCom, so Lark and Slack can move onto it later. Mine is an in-memory holding queue: it loses every held reply on restart, and a message held on one replica is invisible to another. Yours is simply the right design, and the generic keying is the part I'd most like to see land whatever else happens.
2. **Delivery fencing inside the claim SQL** — installation still active, session still bound to this installation. Mine can deliver a held message into a chat whose binding was revoked or re-pointed while it waited.
3. **The compensating reconciler** for enqueues lost when the in-process bus drops them, with the fast/reconcile ratio doubling as an alarm on the realtime path. Neither of the other branches has anything here.
4. **SSRF protection on media download** — resolving the destination yourself and refusing loopback, link-local and private ranges. Mine follows whatever URL the callback carries, including an `https://` that 302s to `169.254.169.254`. That one is my bug and I'm fixing it; thank you for making it obvious.
5. **Per-target send rate gating** against WeCom's own 30/minute and 1000/hour, under an advisory lock. Mine finds out by getting frames refused.
6. **Streaming decrypt through 0600 temp files** instead of holding the whole attachment in memory twice.
7. **Credential verification before the install row is written** — a real `aibot_subscribe` handshake, with "rejected" distinguished from "couldn't reach WeCom". A typo'd secret on mine produces a row that says Connected and a bot that never answers.
8. **Prometheus metrics**, **scan-code provisioning**, **an `enter_chat` welcome**, and **user docs in four languages**. None of the other branches ship any of those.

Also worth saying plainly: **your binding-token path is safe, and it's clear you knew about the hazard rather than got lucky.** A group message from an unbound user enqueues a token-less "DM me" prompt, the mint is gated on `TargetChatType == P2P`, and it fails closed if a group row ever reached it. That exact hazard was the top blocker on #5833, and you pinned it with a test.

Where the branches genuinely differ is scope, and it's close to complementary. #5833 plus my hardening is built around the live conversation: streaming replies via `aibot_respond_msg`, in-flight progress inside that bubble, outbound media through the chunked `aibot_upload_media_*` dance, and the 846605 / 846608 edge cases that only exist once you stream. Yours declares `CapText` with "phase 0 ships no streaming edit path", so a user sends a message and sees nothing at all until the run finishes — which for an agent run is routinely minutes — and an agent that produces a file has no way to hand it over.

Two things I'd raise as defects rather than differences, since they'll bite you whichever way this goes:

- **`ws_conn.go:420` clears the handshake read deadline and never sets another**, and there's no pong handler or ping-response timeout. On a half-open TCP connection — routine for long-lived WebSockets behind corporate NAT — `readPump` blocks forever while the pings keep succeeding into the send buffer. Inbound messages are silently lost until the kernel gives up, on the order of fifteen minutes on Linux, and the supervisor never reconnects because `Run` never returns.
- **`crypto.go:90` unpads against a 16-byte block.** WeCom pads media plaintext to 32 (the same as its callback-mode WXBizMsgCrypt libraries), so attachments whose pad length lands in 17..32 — roughly half of them — fail with `ErrInvalidPKCS7Padding`. Your `crypto_test.go:192` generates padding at 16, so the suite agrees with the code; live traffic won't.

I'm not asking you to stand down — that's @Bohan-J's call and not mine. But three parallel implementations is a bad outcome for everybody, and you should get to hear the state of play from someone who has read your code rather than from a merge notification. If #5833 lands, I'd like to work with you on bringing the queue, the rate gate, the SSRF fix and the metrics on top of it. I've agreed to co-own `server/internal/integrations/wecom/`, and I'd much rather that directory ended up with the best of both branches than with whichever one happened to merge first.

wzxjohn:
Hi @seacen, thanks a lot for the detailed review and for taking the time to actually check out and read through the branch. I really appreciate the context on the other WeCom PRs as well, and especially the concrete feedback on the parts you think are worth preserving.

Regarding streaming replies and file delivery, both were considered when I was working on this PR, but I intentionally left them out of the current scope because, as far as I could tell, the platform itself does not yet provide the corresponding end-to-end support.

For streaming, the WeCom side has the necessary APIs, but the platform currently does not expose a complete streaming/edit path that this integration can hook into cleanly. My plan was to keep this PR focused on the non-streaming path first, and add streaming support once the platform-level capability is available.

File delivery is similar. The WeCom APIs can support outbound media, but the platform does not currently seem to provide a complete file-output-to-channel-delivery path, so I did not want to introduce a WeCom-specific path around the platform abstraction just for this PR.

That said, from your description it sounds like #5833 / your hardening branch may already have solved or introduced some of the platform-side pieces needed for both streaming and file delivery.

Do you have pointers to the relevant PRs or changes for those capabilities? In particular, have the platform-level streaming and file-delivery pieces already landed on main, or are they currently part of #5833 / another pending PR?

If there is already an implementation or an agreed direction I can reference, I'd be happy to revisit these parts and adapt this PR to use the same path rather than waiting for a separate platform change.

Bohan-J:
@wzxjohn — thank you for this, and I'm sorry you're hearing our decision after you'd already put ~19.5k lines into it.

**We've merged #5833 as the WeCom baseline** (just now, as 040bb97b0). I want to be straight about why, because it is not a quality judgement on your work.

#5833 got there first in review terms, not in code terms. It's been through four review rounds since 2026-07-23, it's a quarter the size, and — the part that mattered most — two contributors had already committed to co-owning the directory long-term, one of whom runs it against a real WeCom tenant daily. Yours landed yesterday with no review cycles yet. Converging on one implementation was overdue: counting #4128 and #3993, this feature has now been submitted four times, and letting two live implementations keep growing in parallel would have wasted more of someone's evenings.

**Your implementation is better than the one we merged in at least one important way, and I'd like to take it.**

The durable `channel_outbound_queue` — claim/lease, attempts/backoff, dead-letter, the reconcile state table, and the fact you keyed it on `channel_type` rather than hardcoding WeCom — solves a problem #5833 does not. WeCom's only outbound path there is an in-process WebSocket registry, while `EventChatDone` dispatches on an in-process bus, so on a multi-replica deployment replies produced on the wrong replica are dropped. We shipped that with a documented single-replica constraint and a startup warning. Your queue actually fixes it, and because it's channel-keyed, Lark and Slack could move onto it later.

So: **would you send the queue as its own PR on top of main?** I'd rather review it as cross-channel infrastructure in its own right than have it arrive as part of a WeCom adapter — it deserves that framing, and it'll get a more useful review. The per-target rate gating against WeCom's quotas and the SSRF protection on media download are both worth having too, as separate PRs. Credit stays yours in the history.

## The same offer we made the others

We don't use WeChat Work ourselves — we're focused on overseas markets — so we can't meaningfully test this end to end, and we said as much before merging #5833. `server/internal/integrations/wecom/` is community-maintained, currently co-owned by @leroy-chen and @seacen.

**If you're willing, we'd like you as a third co-owner.** The terms are the same ones they accepted, and I'll restate them plainly:

- **What we commit to:** we keep the adapter compiling and its tests green through shared-layer refactors, ship it in every release, and route WeCom issues to you rather than leaving them unowned.
- **What we'd ask:** WeCom-specific bugs and behaviour questions come to you first, best-effort. You're a volunteer and we'd treat it that way — "we'll route these to you," not "you're on the hook."
- **The exit rule:** documented as community-maintained. If it breaks in a way we can't fix without real WeCom access and no fix appears for a few releases, we may deprecate it. That rule exists to reduce pressure, not add it — if you get busy or move on, the worst case is we retire the integration, not that anyone chases you. A volunteer role should have a no-fault way out.
- **You can say no**, to the ownership and to the follow-up PRs both, without any hard feelings.

Three owners would be better than two here, and you'd be joining with the strongest architectural contribution of the three.

If WeCom usage grows into something real on our side, we'd revisit making this first-party supported, at which point the resourcing becomes ours.

I'm leaving this PR open for now rather than closing it — that's your call to make once you've decided what you'd like to do with the pieces. And thank you for the two fixes you turned around on @seacen's review inside a day; the readPump bound and the 32-byte unpadding were both real.

wzxjohn:
@Bohan-J , thank you for the clear explanation, and also for taking the time to review the implementation in depth and identify the pieces worth carrying forward.

I should also own part of this: before starting this PR, I didn't notice that there were already active WeCom implementations and ongoing review work. That's on me. Had I caught that earlier, I would have coordinated first rather than building another implementation in parallel.

I'm happy with the direction you've proposed. I'll rebase the relevant work on the current `main` and split the reusable pieces out into focused PRs. In particular, I'll start with the cross-channel outbound queue infrastructure, and then follow up separately with things like the WeCom rate gating and media-download SSRF protection where appropriate. I agree that these will be much easier to review and evolve when they're separated from the adapter itself.

And yes, I'd be very happy to join @leroy-chen and @seacen as a third co-owner of `server/internal/integrations/wecom/`. We depend on this integration in production, so I have a strong practical interest in keeping it reliable and moving it forward. I'm happy to help with WeCom-specific bugs, behavior questions, testing against a real tenant, and future improvements on a best-effort basis.

Given that #5833 is now the baseline, I'll close this PR so we don't keep two implementations moving in parallel. I'll gradually bring the reusable parts back as separate PRs on top of `main`.

Thanks again for the thoughtful handling of this and for preserving the useful parts of the work. I'm looking forward to continuing to improve the WeCom integration together.

PR #6529, opened 2026-08-06
feat(channel): materialize inline media references

yyclaw:
## What does this PR do?

Materializes successfully resolved inline channel images as stable Markdown attachment references in the durable user message while preserving their original rich-text positions.

The existing media pipeline already downloaded objects, recorded cleanup intents, and created attachment rows, but the durable message body retained `[Image]` placeholders. The change carries the authoritative appended body into the media binder and updates that channel-ingested user message in the same transaction that claims media intents and links attachments. DingTalk records each adapter-generated marker's exact occurrence so identical user-authored `[Image]` text is never replaced.

This keeps agent input and clients aligned on authenticated `/api/attachments/{id}/download` references. Failed images retain their placeholders, issue-owned media remains issue-owned, and Lark/Slack keep their existing standalone-attachment behavior unless they explicitly opt into inline markers.

## Related Issue

Closes #6528

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- Pass the durable channel message body through shared media binder adapters.
- Materialize only successfully linked inline media inside the attachment transaction.
- Add a provenance-guarded sqlc query for channel-ingested user message updates.
- Track DingTalk adapter-generated marker occurrences across user text, `/new` normalization, and partial media failures.
- Add unit, persistence, router, adapter-mapping, collision, and partial-failure regression tests.

## How to Test

1. Run `make sqlc`.
2. Run `cd server && go test ./internal/integrations/channel/engine ./internal/integrations/dingtalk ./internal/integrations/lark ./internal/integrations/slack -count=1`.
3. Run the targeted race tests for channel media binding and DingTalk media resolution.
4. Send DingTalk rich text containing literal `[Image]` text interleaved with real images and verify only adapter-generated markers become attachment links.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [x] If this change affects the UI, I have included before/after screenshots (not applicable: backend-only behavior)
- [x] I have updated relevant documentation to reflect my changes (not applicable: no public documentation contract changed)
[...]

PR #6532, opened 2026-08-06
MUL-5837 docs(dingtalk): document issue image attachments

yyclaw:
## What does this PR do?

Updates the DingTalk integration guide to match the current `/issue` media behavior introduced by #6388: images from the command message are attached to the created issue.

The stale statement appeared twice in each localized guide and incorrectly told users that images remained only on the chat turn.

## Related Issue

Follow-up documentation correction for #6388.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Correct the `/issue` capability table in the English, Simplified Chinese, Japanese, and Korean DingTalk guides.
- Correct the matching member workflow example in all four guides.
- Keep the documented 4-image and 10 MiB-per-image limits unchanged.

## How to Test

1. Run `git diff --check`.
2. Verify all four DingTalk guides describe images as attached to the created issue and contain no stale opposite claim.

## Checklist

Risk is limited to wording accuracy; there is no runtime or configuration change. The full docs build was not run because this isolated worktree has no installed dependencies. `git diff --check` passed.

## AI Disclosure

**Prompt / approach:**
Compared the merged `/issue` media implementation with all four DingTalk guides, updated only the two stale statements in each locale, checked the Chinese terminology guide, and scanned for the removed opposite claims.

## Screenshots (optional)

Not applicable; documentation text only.


yyclaw:
Hi @Bohan-J, a friendly reminder that it would be helpful to get this documentation corrected before the next release, so the published DingTalk behavior does not cause confusion for users.\n\nFrom a product completeness and polish perspective, it would also be worth considering #6529, #6534, and #6536 for the same release. Together, they address several user-visible gaps around inline media, terminal failure feedback, and preserving media in issue descriptions.

Bohan-J:
Thanks for catching this, @yyclaw — the guides were telling users the exact opposite of what ships, which is the kind of doc bug that makes people redo work by hand.

I verified the new wording against the implementation rather than just the description: DingTalk registers its media resolver into the shared channel engine, the `/issue` path passes the newly created issue ID into media binding, and the attachment row is then owned by that issue. Your statement is accurate, all four locales are covered, and Lark/Slack never made this claim, so DingTalk-only is the right scope.

I pushed one small follow-up commit to your branch: the **Send images** row still said each image "stays visible in the chat," but for `/issue` messages the attachment is owned by the issue and is not linked to the chat turn — so the two rows read as a contradiction. Added a one-line qualifier in all four locales.

CI is green. Merging now.

PR #6560, opened 2026-08-07
MUL-5855: feat(dev): add worktree database cleanup

yyclaw:
## Summary

- add `make db-drop` with local-only validation, protected databases, and `y/N` confirmation
- add `make remove-worktree WORKTREE=...` to drop a linked worktree database before removing the worktree
- document Git's lack of a native `pre-worktree-remove` hook and wire shell regression coverage into CI

## Why

Worktree databases share a persistent PostgreSQL volume, but `git worktree remove` does not clean them up. Git does not expose a hook before worktree removal, so old databases accumulate unless contributors remember to delete them manually.

The new repository wrapper provides one explicit cleanup path and refuses unsafe targets before any destructive database operation.

## Safety

- refuses remote `DATABASE_URL` values
- protects PostgreSQL system databases and the default `multica` main database
- defaults confirmation to no
- refuses primary, current, locked, and dirty worktrees before dropping their database
- uses `dropdb --force --if-exists` with the database name passed as an argument

## Validation

- `bash scripts/worktree-db.test.sh`
- `bash scripts/selfhost-config.test.sh`
- `bash -n scripts/drop-database.sh scripts/remove-worktree.sh scripts/worktree-db.test.sh`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,9 @@ jobs:
               - 'scripts/generate-reserved-slugs.mjs'
               - 'server/internal/handler/reserved_slugs.json'
               - 'scripts/selfhost-config.test.sh'
+              - 'scripts/drop-database.sh'
+              - 'scripts/remove-worktree.sh'
+              - 'scripts/worktree-db.test.sh'
               - 'scripts/selfhost-wait.sh'
               # selfhost-config.test.sh asserts on the installers' side of
               # the port contract, so a change to either must run it.
@@ -148,6 +151,10 @@ jobs:
       - name: Test self-host env derivation
         if: ${{ needs.changes.outputs.frontend == 'true' }}
         run: bash scripts/selfhost-config.test.sh
+
+      - name: Test worktree database cleanup
+        if: ${{ needs.changes.outputs.frontend == 'true' }}
+        run: bash scripts/worktree-db.test.sh
 
       - name: Verify reserved-slugs.ts is up to date
         if: ${{ needs.changes.outputs.frontend == 'true' }}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,6 +79,8 @@ Use the repo scripts as the source of truth. Common commands:
 make dev              # auto-setup and start the app
 make start            # start backend + frontend
 make stop             # stop app processes for this checkout
+make db-drop          # permanently drop this checkout's local database
+make remove-worktree WORKTREE=../path  # drop a linked worktree DB, then remove it
 make server           # run Go server only
 make daemon           # run local daemon
 make test             # Go tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -673,6 +673,8 @@ The command prints the selected database and environment file, then requires a
 `y/N` confirmation. It only operates on the local Docker PostgreSQL service,
 protects PostgreSQL system databases, and refuses to drop the default main
 database `multica` unless `ALLOW_MAIN_DB_DROP=1` is explicitly supplied.
+Declining the confirmation is a successful no-op; when called by
+`make remove-worktree`, it also leaves the worktree in place.
 
 If you want to wipe all local PostgreSQL data for this repo:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,6 +191,25 @@ make stop-worktree    # stop
 make check-worktree   # verify
 ```
 
+### Removing a Worktree
+
+Git does not provide a `pre-worktree-remove` hook. Use the repository wrapper
+from another checkout so database cleanup happens before Git removes the
+worktree directory:
+
+```bash
+make remove-worktree WORKTREE=../multica-feature
+```
+
+The command refuses to remove the primary checkout, the current checkout, a
+locked worktree, or a worktree with uncommitted changes. If the target contains
+`.env.worktree`, it shows the database name and asks for `y/N` confirmation,
+drops that database, and only then runs `git worktree remove`. A worktree that
+was never set up has no `.env.worktree`, so database cleanup is skipped.
+
+Running `git worktree remove` directly bypasses this cleanup and can leave an
+orphaned local database.
+
 ## Running Main and Worktree at the Same Time
 
 This is a first-class workflow.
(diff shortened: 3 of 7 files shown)


Bohan-J (changes requested):
Thanks for this — the product direction is right, and the safety design is genuinely careful (default-deny confirmation, `exit 2` for cancellation, dropping the DB *before* removing the directory so `.env.worktree` is never lost first). I'd like to merge it, but I found one must-fix issue first.

## Must-fix: the "current checkout" guard is bypassed from any subdirectory of the target

`scripts/remove-worktree.sh:25` captures the current directory verbatim:

```bash
current="$(pwd -P)"
```

and `scripts/remove-worktree.sh:60` compares it to the target with an exact string match:

```bash
if [ "$target" = "$current" ]; then
```

`target` is always normalized to a worktree **root** (line 24), but `current` is wherever the user happens to stand. So the guard only fires when the shell is exactly at the worktree root. From `<worktree>/backend` — or any nested directory — `current` != `target`, the check passes, and the script deletes the worktree the caller is standing in.

This contradicts the behaviour documented in `CONTRIBUTING.md:205` ("The command refuses to remove the primary checkout, **the current checkout**, ...").

### Reproduction

Fixture: a repo with a tracked `backend/` directory, `.env*` gitignored (matching this repo's `.gitignore:21`, so `.env.worktree` does not trip the dirty check), one clean linked worktree containing a `.env.worktree`, and `docker` stubbed on `PATH` to log calls instead of executing.

```
CASE A — cwd = <worktree>            (exact match, baseline)
  exit 1
  "Refusing to remove the current worktree from inside itself."
  docker calls: 0
  worktree: STILL EXISTS          ✅ guard works

CASE B — cwd = <worktree>/backend    (tracked subdirectory)
  exit 0
  "Dropped database 'multica_feature'."
  "Removed worktree '<worktree>'."
  docker calls: 1
    docker compose exec -T postgres dropdb --username multica \
      --maintenance-db postgres --if-exists --force -- multica_feature
  worktree: DELETED               ❌ guard bypassed
```

Case B is the realistic invocation: you are working inside the worktree, in a subdirectory, and you paste the removal command. Both irreversible steps run — the database is dropped and the directory is deleted out from under the running shell (and any editor, dev server, or process still rooted there). Note this also means `repo_root` (line 12) resolves to the target itself, so the final `git worktree remove` is Git deleting the very worktree it was invoked from; Git does not refuse this.

### Suggested fix

Normalize the *current worktree root* rather than the raw `pwd`, before any destructive step:

```bash
current_root="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
current="$(cd "$current_root" && pwd -P)"
```

The `cd ... && pwd -P` wrapper matters: `target` is symlink-resolved on line 24, so the comparison needs `current` resolved the same way (on macOS, `/var` vs `/private/var` would otherwise defeat the match).

I verified this patch against the same fixture:

- cwd = `<worktree>/backend` → exit 1, refusal message, **0 docker calls**, worktree intact.
- cwd = primary checkout root (happy path) → unchanged: DB dropped, worktree removed.

### Please also add a regression test

`scripts/worktree-db.test.sh` passes **identically with and without** the fix applied — every existing case invokes the script from `cd "$repo"` (the primary root), so this path is simply not exercised. That is why CI is green here; the green checks don't cover it.

Suggested case: invoke removal from a nested subdirectory of the target worktree and assert all three of — the command fails, the worktree still exists, and the docker stub was **never** called. The existing "assert docker was never invoked" helper used by the cancel/refusal cases is exactly the right tool.

## Everything else checked out

- `bash -n` on all three scripts: clean.
- `scripts/worktree-db.test.sh`: passes as-is, and still passes with the fix applied (no regression).
- Existing refusals verified working: primary checkout, unregistered path, locked, dirty, and cancel-preserves-worktree.
- No schema changes, foreign keys, or cascades.

Non-blocking nits, for the record — no change requested:

- The `DATABASE_URL` localhost check (`scripts/drop-database.sh:33`) is glob matching rather than URL parsing, so a deliberately crafted URL could slip past. The threat model here is operator error, not an adversary, and the execution surface (`docker compose exec` into a local container) can't reach a remote host anyway.
- `set -a; . "$env_file"` executes arbitrary shell from the env file — consistent with existing scripts in this repo, so not a new risk surface.
- Worth calling out in the release notes that this only covers the local Docker PostgreSQL setup; contributors on another local database won't get automatic cleanup.

Happy to re-review quickly once the guard and its regression test are in.

yyclaw:
Thanks for the detailed review. I fixed the must-fix in commit dea0efa11: the script now resolves the current worktree root with git rev-parse --show-toplevel and normalizes it with pwd -P before comparing it with the target. I also added a regression test that runs from a nested backend directory and verifies that removal fails, the worktree remains, and Docker is never invoked. Local checks passed: bash syntax checks, scripts/worktree-db.test.sh, scripts/selfhost-config.test.sh, and git diff --check. Could you please take another look?

Bohan-J (approved):
Thanks for the quick turnaround — this addresses the must-fix exactly, and the regression test is the right one.

## Verified at `dea0efa1`

**The guard fix.** `scripts/remove-worktree.sh:25-26` now resolves the current *worktree root* and normalizes it with `cd ... && pwd -P`, matching how `target` is resolved on line 24. Re-ran my original fixture against the new head:

```
A  cwd = <worktree>                  → exit 1, refused, 0 docker calls, intact  ✅
B  cwd = <worktree>/backend          → exit 1, refused, 0 docker calls, intact  ✅ (was the bug)
C  cwd = <worktree>/backend/deep     → exit 1, refused, 0 docker calls, intact  ✅
D  cwd = <primary>/backend/deep      → proceeds, DB dropped, worktree removed   ✅
E  cwd = <primary> root              → proceeds, DB dropped, worktree removed   ✅
```

C confirms it holds at arbitrary nesting depth, and D is the one I most wanted to see — the fix rejects only the *current* worktree and doesn't over-block the legitimate case of running from a nested directory of another checkout. The refusal still lands before any destructive step, so the database is never touched.

**The regression test.** `scripts/worktree-db.test.sh:138-149` asserts all three of the right things: the command fails, the worktree survives, and the docker stub was never called. I mutation-tested it rather than just reading it — reverting line 25-26 back to `current="$(pwd -P)"` makes the suite fail with exactly:

```
FAIL: remove-worktree must refuse the current worktree from a nested directory
```

So the test genuinely guards the fix instead of passing incidentally. That was my main concern last round, since the previous suite passed identically with and without the bug. I also confirmed `scripts/worktree-db.test.sh` is wired into CI at `.github/workflows/ci.yml:157` and is in the path filter (line 56), so this case runs on every relevant change going forward.

**Other checks.** `bash -n` clean on all three scripts; full suite passes; `git diff --check origin/main...HEAD` clean; local checkout matches the PR head; all 9 CI checks green.

The non-blocking notes from earlier (glob matching on `DATABASE_URL`, `set -a; . "$env_file"` semantics, and calling out local-Docker-PostgreSQL-only scope in the release notes) stand as documentation only — nothing to change here.

Nice work on the underlying design, too: the `exit 2` cancellation protocol and dropping the database before removing the directory (so `.env.worktree`, the only record of the mapping, is never lost first) are both the right calls. Merging.

PR #6561, opened 2026-08-07
fix(server): stop broadcasting inbox:unread to the whole workspace (MUL-5603 PR 0)

NevilleQingNY:
MUL-5603 · **PR 0** — the standalone fix the plan requires to land before the inbox refactor.

## The bug

`inbox:unread` is missing from both `personalEvents` and the targeted-subscriber
list in `registerListeners`, so it falls through to the `SubscribeAll` workspace
fanout. Marking one notification unread pushes that person's `item_id` and
`recipient_id` to **every connected member of the workspace**, and makes all of
their clients invalidate an inbox that has not changed.

The other six inbox events (`new`, `read`, `archived`, `unarchived`,
`batch-read`, `batch-archived`) are all routed to their recipient. This one was
simply never added when `MarkInboxUnread` shipped.

## The fix

Both halves are needed:

1. Add it to `personalEvents`, so `SubscribeAll` stops broadcasting it.
2. Add it to the loop that resolves `recipient_id` and sends only to that user.

Doing only (1) would silence the event — the recipient would stop receiving it
too.

## Tests

`TestRegisterListeners_InboxEventsNeverReachTheWorkspaceRoom` covers **all seven**
inbox events, not just the broken one. The invariant is "inbox events are
personal state"; pinning it per event is what stops the next one from being
added to the publisher without a route. It also asserts the recipient still gets
a well-formed frame, because silencing the leak by dropping the event would be
its own bug.

`TestRegisterListeners_InboxEventWithoutRecipientIsDropped` pins that a producer
bug (no recipient) drops the event rather than falling back to the workspace
fanout — that would be the same leak by another route.

Negative-verified: reverting either half of the fix fails the `inbox:unread`
case.

## Verification

```
go test ./cmd/server/         ok
go test ./internal/handler/   ok
go build ./... && go vet      clean
```

## Scope

Deliberately split out of the inbox refactor. This is a live privacy and
cache-churn bug on its own and should not wait behind a large change — the
MUL-5603 plan lists it as a separate PR that lands first.

### DIFF
--- a/server/cmd/server/listeners.go
+++ b/server/cmd/server/listeners.go
@@ -81,6 +81,7 @@ func registerListeners(bus *events.Bus, b realtime.Broadcaster) {
 	personalEvents := map[string]bool{
 		protocol.EventInboxNew:           true,
 		protocol.EventInboxRead:          true,
+		protocol.EventInboxUnread:        true,
 		protocol.EventInboxArchived:      true,
 		protocol.EventInboxUnarchived:    true,
 		protocol.EventInboxBatchRead:     true,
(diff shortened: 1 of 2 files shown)


NevilleQingNY:
Folded into #6563 — naiyuan asked for this batch to land as a single PR. The commit (`fix(server): stop broadcasting inbox:unread to the whole workspace`) is unchanged there, still first in the branch so it stays reviewable on its own.

PR #6578, opened 2026-08-07
MUL-5873 fix(channels): improve /new and /issue feedback

### LINKED ISSUE #6577: [Bug]: Channel /new and empty /issue lack actionable feedback
## Deployment type

Official App

## What happened?

Channel command feedback is confusing in two related cases:

1. Sending `/new` by itself produces no confirmation, so the user cannot tell whether the next message will start without previous context.
2. Sending `/issue` without a title may surface an internal-error message or reuse an earlier user message as the title. The fallback can create an unrelated issue or report a duplicate that appears disconnected from the command the user just sent.

Expected behavior:

- A bare `/new` should confirm that the next chat message will start without previous context.
- A bare `/issue` should never infer a title from chat history. It should show the required format:

  ```text
  /issue <title>
  [description]
  ```

## Steps to reproduce

1. Open a DingTalk conversation connected to a Multica agent.
2. Send `/new` by itself.
3. Observe that no confirmation is posted.
4. Send a normal chat message, then send `/issue` by itself.
5. Observe either an internal-error response or an issue result based on the previous message instead of usage guidance.

## Screenshots (optional)

Screenshots were included in the original report and show both the internal-error response and the unexpected duplicate-issue response.

## Additional context (optional)

Observed on DingTalk. Command parsing and task scheduling are implemented in the shared channel engine, so the product outcome should remain consistent across DingTalk, Feishu/Lark, Slack message commands, and WeCom.

yyclaw:
## What does this PR do?

Fixes two channel-command feedback failures:

- A bare `/new` now returns an explicit confirmation and persists the fresh-start intent until the next chat task is successfully queued.
- A bare `/issue` no longer infers its title from prior chat history. It returns actionable usage guidance and skips issue creation, agent execution, and media resolution.

When a title-less `/issue` carries media in the same inbound message, the usage guidance asks the user to resend the corrected command together with that media. Media association is message-scoped: rich-text media may appear before or after the command within that message, but a separate earlier media message is not retroactively reused by a later `/issue`.

The shared channel engine owns both terminal outcomes so DingTalk, Feishu/Lark, Slack message commands, and WeCom keep the same product semantics. Platform repliers only provide localized presentation.

The fresh-start state moves from process memory to `channel_chat_session_binding.pending_fresh`. `EnqueueChatTask` locks and consumes that flag in the same transaction that creates and seals the task, so restarts and failed/offline enqueue attempts cannot lose the intent. The pending intent intentionally has no TTL: it remains active until a chat task is successfully queued.

Rollout risk is limited to an additive boolean column with a constant default. Existing sqlc binaries use explicit generated column lists, and repository startup applies migrations before the new server starts.

## Related Issue

Closes #6577

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Remove the previous-user-message fallback from shared `/issue` parsing and delete its obsolete query.
- Add `fresh_pending` and `issue_usage` outcomes with platform-specific replies.
- Detect media on a title-less `/issue` before skipping resolution, and provide media-aware resend guidance on media-capable adapters.
- Persist and transactionally consume pending fresh-session intent without an expiry.
- Update DingTalk and Lark integration documentation for the new command contract.
- Add migration 265 and regenerate sqlc output.
- Add shared-router, platform-replier, task-service, and production-path channel E2E coverage.

## How to Test

1. Run `make sqlc` and confirm there is no generated diff.
2. Run `go vet ./...` from `server/`.
3. Run the affected Go packages, including the full handler suite, against the worktree database.
4. Run `make test ENV_FILE=.env.worktree`; if the subprocess-heavy `pkg/agent` suite hits host-load deadlines, rerun its repository-prescribed guarded command with `-p 2 -parallel 2`.

## Checklist

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:**

Traced the shared command parser, router, session binder, task enqueue transaction, and all four channel repliers. Ran a two-round `review-loop`, fixed the missing production-path E2E binder implementation and added a durable-state E2E oracle, then completed the `pr-preflight` contract, lifecycle, concurrency, rollout, product, and test matrices before publishing.

## Screenshots

image

### DIFF
--- a/apps/docs/content/docs/dingtalk-bot-integration.mdx
+++ b/apps/docs/content/docs/dingtalk-bot-integration.mdx
@@ -75,7 +75,7 @@ Setting this up for **multiple agents**? Repeat the whole flow once per agent �
 | **@-mention in a group** | Add the bot to the group and @-mention it. Only the mentioning message is read — the bot does not listen to the whole group. |
 | **Send images** | Images in a DM, or sent alongside a group @-mention, land in the conversation for the agent to see — PNG, JPEG, GIF, WebP, or BMP, up to 4 images per message and 10 MB each. Each image is copied into Multica storage, so it stays visible in the chat after DingTalk's temporary link expires. Images sent with `/issue` go to the created issue instead of the chat turn. Files and voice messages aren't supported. |
 | **`/issue` command** | Starting a message with `/issue <title>` creates a Multica issue directly from your input, attributed to you, and posts the issue ID and title back into the same conversation. Put the description on following lines. Images sent in the same DingTalk message are attached to the created issue. The command itself does not appear in Multica Chat; the created issue is the result recorded in Multica. |
-| **`/new` command** | Starting a message with `/new <your message>` runs that message without prior conversation context. Sending `/new` by itself records the same fresh-start intent for your next non-empty message without creating an empty turn. The conversation history remains intact. |
+| **`/new` command** | Starting a message with `/new <your message>` runs that message without prior conversation context. Sending `/new` by itself records the same fresh-start intent for your next non-empty message without creating an empty turn, and the bot confirms when that intent is ready. The pending intent remains active until a later chat task is successfully queued. The conversation history remains intact. |
 | **Reply** | The agent's answer is posted back into the same 1:1 chat or group. |
 
 ## Use the bot (members)
@@ -94,7 +94,7 @@ Only **members of the workspace** can use the bot. If you aren't a member, or yo
 - **In a 1:1 chat** — open the bot and message it directly; no mention needed, every message is read.
 - **Send images** — send screenshots or photos, with or without text; they show up in the conversation for the agent to work with. PNG, JPEG, GIF, WebP, or BMP; up to 4 images per message, 10 MB each.
 - **File an issue** — send `/issue the login redirect is broken on Safari`; add the description on following lines if needed. Multica creates the issue synchronously, attaches images from that message to it, and posts its ID and title back into the chat.
-- **Start fresh** — send `/new <your message>` to run that message without prior context, or send `/new` by itself to apply the fresh start to your next non-empty message. Neither form removes the existing conversation history.
+- **Start fresh** — send `/new <your message>` to run that message without prior context, or send `/new` by itself to apply the fresh start to your next non-empty message. A bare `/new` gets a confirmation, and its pending intent remains active until a later chat task is successfully queued. Neither form removes the existing conversation history.
 
 ## Manage and disconnect
 
--- a/apps/docs/content/docs/lark-bot-integration.mdx
+++ b/apps/docs/content/docs/lark-bot-integration.mdx
@@ -54,7 +54,7 @@ Send:
 Safari stays on the login page after a successful sign-in.
 ```
 
-The first line becomes the title and the rest becomes the description. If you send `/issue` alone, Multica tries to use your previous message as the title.
+The first line becomes the title and the rest becomes the description. Sending `/issue` alone creates nothing; the Bot replies with the required format instead.
 
 ## Manage connections
 
--- a/apps/docs/content/docs/lark-bot-integration.zh.mdx
+++ b/apps/docs/content/docs/lark-bot-integration.zh.mdx
@@ -54,7 +54,7 @@ import { Callout } from "fumadocs-ui/components/callout";
 Safari 登录成功后仍然停留在登录页。
 ```
 
-第一行成为标题，后面的内容成为描述。只发送 `/issue` 时，Multica 会尝试使用上一条消息作为标题。
+第一行作为标题，其余内容作为描述。只发送 `/issue` 不会创建任务；Bot 会回复用法提示。
 
 ## 管理连接
 
--- a/server/internal/handler/channel_new_e2e_test.go
+++ b/server/internal/handler/channel_new_e2e_test.go
@@ -31,6 +31,7 @@ func TestChannelNewCommandE2EStartsFreshProviderSession(t *testing.T) {
 		text             string
 		commandText      string
 		forceFresh       bool
+		followUpText     string
 		wantStoredText   string
 		wantForceFresh   bool
 		wantPriorSession string
@@ -44,6 +45,15 @@ func TestChannelNewCommandE2EStartsFreshProviderSession(t *testing.T) {
 			wantStoredText:   "what model are you?",
 			wantPriorSession: "old-provider-session",
 			wantPriorWorkDir: "/tmp/old-provider-workdir",
+		},
+		{
+			name:           "bare /new applies to the next real message",
+			channelType:    channel.Type("slack"),
+			text:           "/new",
+			commandText:    "/new",
+			followUpText:   "what model are you?",
+			wantStoredText: "what model are you?",
+			wantForceFresh: true,
 		},
 		{
 			name:           "Slack /new message starts without provider context",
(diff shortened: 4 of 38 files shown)


Bohan-J:
Thanks for the detailed writeup — the product direction here is right, and the implementation goes further than the issue asked for in a good way. Moving the fresh-start intent out of the `Router`'s in-memory map into `channel_chat_session_binding.pending_fresh`, then consuming it under `FOR UPDATE` in the same transaction that creates and seals the task, fixes a bug the issue didn't even mention: the old map was already incorrect under multi-replica deployment (a `/new` landing on replica A and the next message on replica B silently lost the intent).

Verified locally on `ba137b8`:

- `go build ./...` clean; `go test -count=1` green for `internal/integrations/channel/...`, `dingtalk`, `slack`, `wecom`, `lark`.
- `sqlc generate` produces a zero diff against the committed output.
- Migration 265 doesn't collide on `main`, and `ADD COLUMN BOOLEAN NOT NULL DEFAULT FALSE` is metadata-only on PG 11+. Combined with sqlc's expanded explicit column lists in the old binary, rollout and rollback are both safe.
- Lock ordering checked: the append transaction takes `chat_message insert → chat_session (Touch) → binding`, while `EnqueueChatTask` takes `binding (FOR UPDATE) → agent_task_queue → chat_message`. The enqueue side never touches `chat_session` and the append side only inserts into `chat_message`, so there's no ABBA cycle — concurrent turns on one session just serialize briefly.
- No foreign keys, no new index (`chat_session_id` already has a `UNIQUE` constraint, so the `FOR UPDATE` is a single-row index lookup).

A few things to address before merge.

## 1. Docs still describe the behavior this PR removes

`apps/docs/content/docs/lark-bot-integration.mdx:57`:

> The first line becomes the title and the rest becomes the description. If you send `/issue` alone, Multica tries to use your previous message as the title.

and the same sentence in `lark-bot-integration.zh.mdx:57`. That fallback is exactly what this PR deletes, so both lines now teach users a behavior that no longer exists. Please update them to the new contract (a bare `/issue` returns usage guidance and creates nothing).

While you're there: `dingtalk-bot-integration.mdx:78` and `:97` describe the bare `/new` intent but predate the explicit confirmation — worth a sentence noting the user now gets one.

## 2. Drop the DingTalk-only `commandOutcomeText` helper

`server/internal/integrations/dingtalk/replier.go` introduces `commandOutcomeText(res)`, which re-switches on an outcome the enclosing `Reply` switch has already narrowed. Slack and WeCom just inline the constant in their `case` arms. Inlining here too is shorter and keeps the three repliers reading the same way.

(The per-platform copy constants and DingTalk's `\n\n` vs Slack/WeCom's `\n` are fine as-is — they match how `agentOfflineText` already works, and the line-break difference has its own commit explaining it.)

## 3. Say something when a bare `/issue` carries an image

Skipping media resolution for a title-less `/issue` is the right call, and I confirmed it leaves no orphan rows — `channel_media_pending_object` is only written from inside the resolve path, so nothing dangles. But the image the user just uploaded is silently dropped. For anyone in the habit of sending a screenshot and then the command, the usage reply reads as if only the command was wrong. Please add a line to the usage copy when the message carried media, along the lines of "the image wasn't saved — please resend it with the title".

## 4. `pending_fresh` has no expiry

This replaces the previous 24h `pendingFreshTTL`, so a `/new` from three months ago still applies to the next message. Arguably the more self-consistent semantic (the intent stays valid until consumed), but it is a product-semantics change and the PR description doesn't call it out. Either state it explicitly in the description, or add a `pending_fresh_at` timestamp and keep a bound.

## 5. `/issue` is parsed twice, and the two parses agree only implicitly

`Router.processClaimed` parses `msg.CommandText` to compute `issueNeedsUsage`/`resolveMedia` before the append, and `AppendUserMessage` parses its own `commandSource` again afterwards. Today these are necessarily the same string — `Handle` normalizes an empty `CommandText` to `Text`, and every binder applies the same fallback — so the current code is correct. But nothing enforces it. If a future adapter makes the binder's command source diverge from `msg.CommandText`, the router will open a 45s media budget on a bare `/issue` that then returns usage and never binds, leaving the message media-pending until the durable fallback expires.

Preferred fix: parse once in the router and pass the parsed command down through `AppendParams`. If you'd rather not reshape that struct, a comment on the `issueNeedsUsage` line documenting the coupling is acceptable.

Everything else looks good — the shared engine owning both terminal outcomes with the platforms only supplying copy is the right boundary, the removal of `titleFromPreviousMessage` and `GetMostRecentUserChatMessage` is complete with no leftover references, and `TestRouter_BareIssueReturnsUsageWithoutCreatingOrResolvingMedia` pins the media-skip with an actual image message. Happy to re-review once the above is in.

yyclaw:
@Bohan-J Thanks for the detailed review. I addressed all five points in the updated single commit `4efcfaf22`:

1. Updated the Lark EN/ZH docs to remove the previous-message fallback, and documented the bare `/new` confirmation in the DingTalk guide.
2. Removed DingTalk's redundant `commandOutcomeText` switch and now use the narrowed constants directly.
3. Added media-aware usage guidance, with one architectural distinction worth making explicit: this path is scoped to **one inbound message**, not a two-event sequence where an image message is followed by a separate `/issue` message. `HasMedia(msg)` inspects media carried by the same inbound message as the bare command. The router records that fact, but deliberately exits with `issue_usage` before `ResolveMedia`, so the media has not been downloaded or persisted to Multica and is not stored-then-deleted. DingTalk/Lark now ask the user to resend the corrected command together with the media; within the same rich-text message, the media may appear before or after the command. A previous standalone media message remains associated with that earlier chat message and is not retroactively reused as a future issue attachment. Supporting the two-event scenario you described would belong to the natural-language/agent path rather than the deterministic `/issue` command—for example, “Create a new issue for xxx and attach the image I just sent.” That path would need to resolve the conversational reference to a previous message's media explicitly; `/issue` intentionally remains self-contained and message-scoped.
4. Kept the no-expiry semantic and now call it out explicitly in the PR description and integration docs: the fresh intent remains active until a chat task is successfully queued.
5. Documented the `CommandText` coupling at the pre-append parse site so a future binder cannot silently make the media decision diverge from the terminal command outcome.

I also rebased onto current `main`, resolved the WeCom/sqlc overlap semantically, regenerated sqlc output, and squashed the PR to one commit. The PR is now mergeable.

Validation after the rebase/squash:

- `make sqlc` (no generated diff)
- `git diff --check`
- `go vet` on the affected engine/integration/service/handler packages
- `go test` for channel engine, DingTalk, Lark, Slack, WeCom, service, and handler packages against the worktree database

Bohan-J:
Re-checked at `4c161ee`. All five items from my earlier comment are addressed, and the rebase onto `cdf93e4` resolved the `EnqueueChatTask` conflict the right way — no change needed from my side.

The conflict was worth getting right, so recording what I verified: `main` had added `LockChatSessionForEnqueue` as the transaction's first statement, and this PR adds `LockChannelChatSessionPendingFresh` at the same position. The merged order is `chat_session` (FOR NO KEY UPDATE) + archived check → binding (FOR UPDATE) → create/seal/defer the task → clear `pending_fresh` → commit. That's the correct direction: every transaction in the codebase that touches `channel_chat_session_binding` for a session takes the `chat_session` row lock first —

- `AppendUserMessage`: `TouchChatSession` before the binding writes;
- `SetChatSessionArchived`: `SetChatSessionArchived` (UPDATE on `chat_session`) is the transaction's first statement, binding read/delete after;
- `DeleteChatSession`: session lock, then agent, then tasks, then binding;
- `ChatSession.MarkPendingFresh`: a single autocommit statement holding no other lock.

So there's no ABBA edge. Keeping `FOR NO KEY UPDATE` rather than `FOR UPDATE` on the session also matters here and is correctly preserved: the inbound `chat_message` insert takes `FOR KEY SHARE` on the parent row, which `FOR NO KEY UPDATE` does not block, so the debounced enqueue can't stall appends.

Also re-verified on this head: `go build ./...` clean, `sqlc generate` zero diff, and the channel engine / DingTalk / Lark / Slack / WeCom package tests green locally. All 9 CI checks pass on `4c161ee`, including the `backend` job that exercises the archive-vs-enqueue and `pending_fresh` database tests.

The `ErrChatSessionArchived` early return landing before the binding lock is also right — it leaves `pending_fresh` untouched, and archiving drops the binding row anyway.

No blocking issues left from me. LGTM.

Bohan-J (approved):
Approving — thanks for a genuinely good contribution, @yyclaw.

Two things stood out beyond the fix itself. First, you didn't just paper over the confusing feedback: moving the fresh-start intent out of the Router's in-memory map into `channel_chat_session_binding.pending_fresh` and consuming it under the same transaction that creates and seals the task also fixed a correctness bug the issue never mentioned — the old map silently lost the intent whenever `/new` and the next message landed on different replicas. Second, the turnaround on review feedback was excellent: all five items addressed, and you resolved the `EnqueueChatTask` rebase conflict with the correct lock ordering on your own, including the comment explaining why the binding lock has to sit behind the session lock.

Squash-merging now. The PR title gets a `MUL-5873` prefix on the way in to match the trunk commit convention.

PR #6581, opened 2026-08-07
Feature: channel queue and WeCom rate gate

wzxjohn:
## What does this PR do?

Pick channel queue, rate gate and QR scan for WeCom from #6524. Since all three changes involve database modifications, we are combining them into a single pull request to avoid frequent changes to the change number.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made
Three commits, each independently reviewable and in dependency order.

### 1. `feat(channel): durable outbound queue for hold-connection channels`

Replaces WeCom's in-process reply handoff with a table, so a reply produced on any replica is
delivered by whichever replica holds the socket lease.

- **`server/internal/integrations/channel/outbox/`** (new package, channel-agnostic):
  `producer.go` (insert from any replica, keyed on `(source_kind, source_id)`),
  `consumer.go` (claim/lease fencing, bounded attempts, full-jitter backoff, `Sender` +
  `PayloadBuilder` interfaces and the adapter-owned `Disposition`), `reconciler.go` (per-channel
  cursor whose lease elects one scanner; re-scans a trailing window of terminal tasks and enqueues
  what is missing, plus retention and stranded-row sweeps), `wake.go`, `metrics.go`, `doc.go`,
  and tests for each.
- **`server/internal/integrations/wecom/`**: `wecom_channel.go`, `outbox_builder.go`,
  `outbox_sender.go` (+ test) make WeCom the first consumer; `outbound.go` / `outbound_test.go`
  rewritten onto the queue. The reply enqueue keys on the task id in `ChatDonePayload`, **not**
  `events.Event.TaskID` — `broadcastChatDone` leaves that envelope field empty, and reading it
  would demote every delivery to the lagged reconciler.
- **`server/internal/metrics/channel_outbox.go`** (new) + `registry.go`: queue depth, claim, and
  disposition metrics.
- **Migrations `265`–`270`**: `channel_outbound_queue` + claim/source/chat-session indexes,
  `channel_outbound_reconcile_state`, and `270_chat_message_task_assistant_idx` for the reply
  lookup. Each concurrent index is its own single-statement file.
- **`server/internal/integrations/wecom/ws_frame.go`**: extracts `msgTypeMarkdown` — it is also what
  lands in `channel_outbound_queue.msg_type`, so a row records the wire form it was enqueued as
  rather than the sender assuming one.
- **`.env.example`, `SELF_HOSTING.md`, `wecom/types.go`**: the WeCom **single-replica constraint and
  its startup warning are removed** — this commit is what made them obsolete. The remaining
  documented limit is that inbound handling is text-only.
- **`server/cmd/server/{main,router}.go`**, `handler/handler.go`: consumer/reconciler wiring.

### 2. `feat(channel): per-target outbound rate gating`

Gates admission so a backlog drain (reconnect, or a queue built up while the lease moved) can't walk
into WeCom's `errcode 45009` over-quota rejection, which the user sees as a non-answer.

- **`server/internal/integrations/channel/outbox/rategate.go`** (+ `rategate_test.go`): sliding
  windows in a ledger table rather than an in-process token bucket, because the quota is the
  platform's and counted per target chat — a bucket resets to zero after a lease flip. The advisory
  lock, window counts, and attempt insert are one transaction. `attempted_at` is stamped *before*
  the frame hits the socket, so an ambiguous write still counts. Over quota **defers without
  spending an attempt**, so a merely-busy target is never dead-lettered.
- **`server/internal/integrations/wecom/rate_limit.go`** (new): WeCom contributes 30/minute and
  1000/hour per target — deliberately under the published ceilings, since the platform counts on its
  own clock. A second channel adds one file like this and nothing else.
- **`outbox/reconciler.go`**: purge now also bounds the ledger, which is a sliding-window count
  rather than history.
- **Migrations `271`–`273`**: `channel_outbound_send_attempt` + window and chat-session indexes.

### 3. `feat(wecom): scan-code install for creating a new smart bot`

Adds the other half of install: scan a QR in WeCom and get a **brand-new** bot, provisioned without
anyone visiting the admin console. The flow is a multi-minute conversation with WeCom, so it cannot
live in a request — a `wecom_install_session` row carries the state and a worker drives it. **No HTTP
handler ever calls WeCom.**

- **`server/internal/integrations/wecom/install.go`, `install_worker.go`, `install_provider.go`**
  (+ `install_test.go`, `install_provider_test.go`): `scode` and `qr_code_url` are sealed with the
  same key as the bot secret — together they're a bearer credential for "finish creating this bot",
  so a DB dump must not carry them. `Begin` runs every guard under a per-workspace advisory lock
  (idempotency-key replay, no live bot on the agent, one live session per agent, per-user rate
  window); without the lock each duplicate spends a WeCom generate-quota slot and shows a second QR
  for a bot nobody will own. The idempotency key is stored **hashed, never raw**. An admin can resume
  a colleague's stuck session. Bind reuses `installation.go`'s `Upsert` rather than a second
  reclaim-then-upsert, and bind/completion are two writes on purpose since `Upsert` is idempotent.
- **`server/internal/handler/wecom_install_web.go`** (new) + `wecom_web.go`: `POST
  wecom/install/begin` and `GET wecom/install/{sessionId}`. Status decrypts only for the initiator or
  a workspace admin — an unauthorized viewer 404s **without the ciphertext being opened at all**.
- **`server/internal/integrations/channel/engine/supervisor.go`**: adds `Notify`, so a just-created
  bot connects immediately instead of up to 30s later.
- **Migrations `274`–`278`**: `wecom_install_session` + due / request / agent-pending / begin-rate
  indexes.
- **`packages/core/`**: `types/wecom.ts`, `api/client.ts` + `api/schemas.ts` — responses go through
  `parseWithFallback` + zod.
- **`packages/views/settings/components/wecom-scan-install-dialog.tsx`** (+ test) and `wecom-tab.tsx`:
  follows `LarkInstallDialog` including its StrictMode `closedRef` reset. One shape difference —
  WeCom's begin doesn't return the QR, so the dialog renders a "preparing" state until a later poll
  turns the status pending. Its test caught a real bug: scheduling the close inside the poll effect
  meant the status change to `success` re-ran the effect and the cleanup cancelled the very timer
  meant to close the dialog.
- **Locales** for `en` / `ja` / `ko` / `zh-Hans`.

### Across all three

- **`server/internal/handler/workspace_delete_manifest_test.go`**: ownership decisions for the four
  new tables, plus `pkg/db/queries/workspace_delete.sql`.
- **`server/pkg/db/queries/`** + regenerated **`server/pkg/db/generated/`** (`sqlc`, no drift).

## Checklist

## AI Disclosure


seacen:
Reviewed this end to end — the durable queue, the rate gate, and the scan install. This is a strong PR and I want to say why before I get to the problems, because the reasoning throughout is better than most of what I've read in this area.

The queue is the right shape for the actual constraint. aibot has no outbound REST path, so a reply can only leave the replica holding the socket, while the reply itself is produced wherever the task ran — the queue is exactly that handoff, and nothing smaller would have worked. The business key follows from that correctly: `(installation_id, source_kind, source_id)` keyed on the business result rather than the attempt, as a non-partial UNIQUE index with `ON CONFLICT DO NOTHING`, is what lets the realtime producer and the reconciler race freely without a coordinating lock. Having no 'sending' status and fencing in-flight rows with `lease_token` + `lease_expires_at` instead means a worker that dies leaves a row somebody reclaims rather than one stranded in a state nothing sweeps. Full jitter instead of exponential-plus-noise, with the reason written down — these failures are correlated, one dropped socket strands every row for an installation at once — is the kind of thing that usually gets picked by feel.

The rate gate being a database ledger rather than an in-process token bucket is the right call for a multi-replica deployment, and the reasoning is at rategate.go:48-53 where the next person will actually find it. It fails closed: every path returns `ok=false` — begin, lock, count, record, commit — and `TestWindowRateGate_FailuresNeverAdmit` enumerates all five, which matters because a gate that fails open is worse than no gate. The lock, both counts and the insert are one transaction with a transaction-scoped advisory lock and an explicit rollback on rejection, tested specifically. The advisory key is namespaced with `hashtext('channel_outbound_rate')` matching the convention `wecom_install_session.sql` already set, and the separator is `chr(1)` rather than `chr(0)`. And the split is clean — the mechanism lives in `channel/outbox`, only the two numbers and their justification are in `wecom/rate_limit.go`, so Lark and Slack get the gate by supplying Windows.

The scan install is careful in the places that are easy to get wrong. `BeginInstall` issues zero upstream calls, so a wedged WeCom can't hold an HTTP request open, and the file says so at the top instead of leaving it to be discovered. Admission is genuinely serialized rather than hopefully serialized: a per-workspace `pg_advisory_xact_lock` around the whole read-then-insert, plus a partial unique index as the backstop. Authorization is split per endpoint rather than lifted to the route, so an agent owner can connect their own agent without being a workspace admin. The scode is never returned by any endpoint; both terminal statements NULL out `scode_encrypted` and `qr_code_url_encrypted` in the same UPDATE, so a dump taken after a session ends carries nothing. Both handlers set `Cache-Control: no-store`, with the reason in the file header. The worker starts even when the scan flow is unconfigured, so in-flight sessions reach `error/integration_unconfigured` instead of leaving an admin watching a spinner. `handlePending` binds the bot when `query_result` succeeds even if `expires_at` just passed, rather than orphaning a bot the admin genuinely created. And the typed-nil-in-interface trap is handled deliberately for `BotBinder` with a comment explaining why.

Three more that normally take a review round each and didn't here: the workspace-deletion manifest is complete with all four new tables classified and reasoned; there's no sqlc drift (I diffed every named query against the generated consts); and the metric labels refuse `installation_id`/`workspace_id` as unbounded cardinality, in writing. Migrations are shaped for this repo's runner — one `CREATE INDEX CONCURRENTLY` per file, with matching `DROP INDEX CONCURRENTLY IF EXISTS` downs across all ten. Locale parity is complete: 17 new keys across en/ja/ko/zh-Hans with byte-identical key sets, and `isKnownScanError` falls back to the generic string so a backend reason the frontend hasn't seen yet doesn't render blank.

Now the problems. Three I'd want fixed before this lands. I reproduced the first and third on your branch; the second I verified link by link but didn't run end-to-end.

## 1. `queueSender.Send` can't see WeCom's verdict, so a refused send settles as delivered

`sendText` → `write` → `conn.WriteMessage`. The only errors that path can produce are marshal failure, `SetWriteDeadline` failure, and the socket write itself. The aibot verdict comes back in the ack frame, which this path never reads. So:

```go
if err := sender.sendText(row.TargetChatID, int(row.TargetChatType), body); err != nil { ... }
return outbox.DispositionSent, nil   // reached whenever the bytes left the process
```

Probe on your branch, using your own `recordingConn` and `newQueueSender`:

```
body = 36000 bytes (cap is 20480)
disposition = 0 (DispositionSent), err = <nil>
```

The row settles DELIVERED, `payload` is cleared, and it can never be rescued.

Three ways to hit it: a reply over 20480 bytes — there's no length budget anywhere on this path (`producer.go`, `outbound.go` and migration 265 have no cap), while `inbox_message.go:18` already caps the inbox card for exactly this reason; the bot removed from a chat whose binding row survives (40001); and 45009 itself.

The cost isn't one lost message. `attempts`, dead-lettering, `DispositionRetry/Failed` and every delivery metric are built on this disposition, so in those three cases the observability is inverted — it reports success. That's the part that undercuts the PR's own headline claim.

`retryableSendError` reads like a smell at first (string-matching "timeout" / "broken pipe"), but it's actually self-consistent given the above — `write` only produces that class of error. The classifier isn't wrong; it just can't see the thing it should be classifying. Wire `sendText` to the ack and it can switch on errcode instead of guessing at strings.

We have both halves on our hardening branch — `sendTextCtx` → `request` → a typed `*wecomAPIError`, and `splitForWire` for the 20480 cap. Happy to send them over as a small PR you build on, if that's less disruptive than doing it inside this one.

## 2. The reconciler's first run backfills 24h of already-delivered replies into live chats

Migration 265 creates `channel_outbound_queue` empty; 269 creates the cursor table with no seed row. On boot `sweep()` seeds the cursor at `now - reconcileInitialLookback` (24h, reconciler.go:241).

`ListChannelOutboundReconcileCandidates`'s only "already delivered" test is `NOT EXISTS (SELECT 1 FROM channel_outbound_queue ...)`. On an existing deployment the pre-PR path (`wecom.NewOutbound` writing straight to the socket) left no rows there, so every terminal WeCom-bound task in that window matches. `reconcileBuilder` has no time predicate, the consumer has no max-age check, and `EnqueueOK` is absent from the `ReconcilerConfig` literal at router.go:802-808, so it defaults to `func() bool { return true }`.

The rate gate paces this but doesn't contain it: the quota is per target chat, so many chats drain in parallel, and `DeferClaimedChannelOutbound` doesn't spend an attempt, so nothing dead-letters.

Result on upgrading any running instance with WeCom in use: every user who talked to the bot in the previous day gets that whole day of answers re-posted into their chat. Once, and unrecallably — aibot messages can't be unsent.

The `reconcileInitialLookback` comment defends the *ceiling* — "delivering a day-old reply into a live chat is worse than not delivering it" — and that reasoning argues against the seed just as strongly. Nothing that finished before the queue table existed can be missing from it, so the window should never reach behind the cursor row's own creation. `reconciler_test.go` builds every case from an already-populated cursor, so the seed path isn't covered.

Smallest fix: seed a fresh cursor at `now` (or `now - reconcileSettleDelay`) and keep the 24h lookback as a floor for an *existing* cursor. Migration 269 is new in this PR, so if you'd rather derive the boundary from the row itself, adding `created_at` is still on the table.

## 3. The `scode` reaches the log in plaintext on any provider transport error

You're careful with this everywhere else: botid/secret are "Neither is logged" (install_provider.go:60), scode and qr_code_url are sealed at rest, both terminal statements NULL them, and there's a test asserting the plaintext scode never reaches the row. One path goes around all of it:

```go
u := ... + "/ai/qc/query_result?scode=" + url.QueryEscape(scode) + "&source=" + ...
resp, err := p.cfg.HTTPClient.Do(req)
if err != nil { return fmt.Errorf("wecom provider: http: %w", err) }   // err is *url.Error
```

`*url.Error.Error()` includes the full URL, query string and all. install_worker.go:267 logs it:

```
wecom provider: http: Get "https://.../ai/qc/query_result?scode=SUPER-SECRET-SCODE-abc123&source=src-1": EOF
```

(reproduced with an httptest server that accepts the connection and hangs up — WeCom hiccuping, an LB dropping it, a DNS blip)

It's `Warn`, so default log levels catch it. It's inside the poll loop, so one sustained upstream outage writes the same scode dozens of times. And the window is exactly when the scode is live and redeemable. Per your own comment at install.go:411, whoever holds it "can finish creating the bot" — and `source` is on the same line. Same leak at :207 on the `generate` path.

Fix: unwrap `*url.Error` in `getJSON` and keep only `err.Err` (or host + path), so the query string never reaches a formatter.

## Also worth a look, roughly in order

- **No length budget on the agent-reply path** (outbound.go:126, outbox_builder.go:123) — the other half of #1.
- **Settlement runs on the cancellable consumer context** (consumer.go:275-292): a shutdown between a successful send and the settling UPDATE re-delivers on restart. `context.WithoutCancel` for the settle.
- **No maximum row age**: a bot offline for hours replays its whole backlog on reconnect; one that never returns leaves rows queued indefinitely.
- **`windowStart` has no floor** (reconciler.go:261-264) — the intended 24h clamp is unreachable, so a stalled cursor re-delivers replies whose tombstones the purge already removed.
- **Hour-window rejection defers a full hour** (rategate.go:130): one message over the hourly quota and that chat goes silent for 60 minutes.
- **The gate only counts queue traffic**: replier.go:189, replier.go:213 and wecom_channel.go:356 push on the same socket against the same per-chat quota and never reach `outbox.RateGate`.
- **`task_failed` has no realtime producer**, so the reconciler's "the realtime path is broken" signal fires on every failed run.
- **`scan_install_supported` landed in `ListDingTalkInstallationsResponseSchema`** (schemas.ts:2186) instead of the WeCom one at :2231.
- **`crypto.randomUUID()`** (wecom-scan-install-dialog.tsx:90) is SecureContext-gated, so the dialog breaks on any non-HTTPS deployment; the repo already has `createSafeId`.
- **`FailWecomInstallSession`** is the one mutation with no lease fence, against the invariant stated at install_worker.go:26.

## Things I checked and concluded were fine

Listing these because several look wrong at first glance and I'd rather they weren't re-raised by the next reader:

**Removing the SINGLE-REPLICA warning is earned.** I walked every remaining direct-socket write — `OutboundReplier.post`/`postPrivate` and the text-only receipt path — and they're all inbound-triggered, so they always run on the lease holder by construction. You also removed the constraint in all four places it was stated, not just the code: `.env.example`, `SELF_HOSTING.md`, the `types.go` package doc, and the boot-time `slog.Warn`.

**Keeping the binding prompt off the queue is right,** and worth stating explicitly since it looks like an omission: `postPrivate` still writes straight to the socket, so a 15-minute bearer token is never buffered across a reconnect and never sits in a payload column.

**Deferral not spending an attempt** is correct — a target that is merely over quota hasn't failed, and burning the retry budget on it would dead-letter healthy traffic. The `RateGate` doc says so and `rategate_test.go` asserts it.

**The payload is cleared on every terminal settle** — complete, fail, and both lifecycle fails — so a delivered row stops holding quoted user content while still serving as the idempotency tombstone. That's the right split between the two jobs one row is doing.

**The double fence on deliverability**, with the read-failure distinction handled correctly: claim-time EXISTS guards keep undeliverable rows out of the loop, and `checkChatSessionDeliverable` re-checks at send time.

**`reconcileBuilder` re-reads the binding and installation** instead of trusting the scan's join, and skips a candidate whose session has been rebound elsewhere. The scan and the send are far enough apart in time that this matters.

**`chatDoneTaskID` reading the payload before the envelope** is correct — I checked `service.broadcastChatDone` at internal/service/task.go:5008, which sets `ChatDonePayload.TaskID` and leaves the envelope's empty.

**`outbound.go` drops nothing main had**: the installation-active fence, the binding lookup, and the `pgx.ErrNoRows` → "not a wecom session" skip that lets it coexist with Slack/Lark are all still there.

**The reconciler owns the retention purge and runs it without holding the cursor lease**, so retention doesn't depend on which replica won the election.

Build and vet are clean and both packages' tests pass on the branch.

seacen (changes requested):
Full review in https://github.com/multica-ai/multica/pull/6581#issuecomment-5217194700 — I won't repeat it here.

Marking changes-requested rather than commenting, specifically because of the reconciler seed. That one fires once, at upgrade, on a live tenant, and aibot messages can't be unsent — so it's the one thing here that a follow-up patch can't fix for whoever already deployed. The other two blockers (the disposition read and the scode in the log) are ordinary fixes, and everything under "also worth a look" is advisory.

Happy to send the verdict-read plumbing (`request`/`requestWithID`/`wecomAPIError` + the reply waiter + `sendTextCtx`) as a standalone PR you build on, if that's easier than doing it inside this one. It's strictly additive and doesn't touch anything you've changed.

wzxjohn:
@seacen Thank you very much for the thorough review. I’ll fix issue 2 first since that’s the most critical one.

My original PR was written against an older `main`, so I did indeed miss this upgrade-path consideration.

And yes, I’d be very happy for you to send the verdict-related plumbing as a separate PR. That would be very helpful — thank you.

seacen:
@wzxjohn Sent it: #6590.

It turned out to need one thing I hadn't accounted for when I offered. **The read loop is the sole deliverer of acks**, and `dispatchFrame` currently runs inbound callbacks inline on it — so a callback that writes a frame and waits for its ack would be waiting on the loop that has to deliver it. Waiting on itself. So #6590 also moves callbacks onto one worker behind a depth-64 queue (one worker because WeCom delivers a chat's messages in order and the engine's dedup assumes it; blocks rather than drops when full). The two halves can't be split, and the PR body explains the ordering.

Once it's in, `retryableSendError` can switch on `*wecomAPIError.Code` instead of string-matching, and `DispositionSent` starts meaning the message arrived rather than that the bytes left the process.

I left `splitForWire` (the 20480 cap) out on purpose — it wants to sit on your send path once the queue is settled, and doing it in parallel would have us both editing the same lines. Say the word once you've landed the reconciler fix and I'll send it.

Good call taking issue 2 first, and no need to apologise for the older `main` — that one only becomes visible if you go looking at what an *existing* deployment's tables hold at the moment of upgrade, which is not where anyone's attention naturally goes while writing the queue itself.

wzxjohn:
@seacen Issues 2 and 3 have both been fixed.

For 2, I updated the reconciler scan/seed logic so upgrading will no longer replay all messages from the previous 24 hours.

For 3, I added explicit error handling to ensure sensitive information is stripped before the error is logged.

seacen (commented):
Re-reviewed `9e21931`. Both fixes verified on my side, and the reconciler one is better than what I asked for.

## The reconciler seed — fixed, and then some

Seeding at `now` is the right call. But you also added the two things that make it *stay* right:

**The `created_at` floor.** `windowStart.Before(state.CreatedAt.Time)` closes a residual I had not written up. With the seed fixed but only a `now - lookback` clamp, the 5-minute overlap window on the second tick still reaches ~5 minutes behind the cursor row's birth — which on an existing deployment is 5 minutes of replies the old direct-socket path already delivered and left no queue row for. Small, same class, and now structurally impossible: the window cannot precede the row that defines "we started tracking here". Deriving the boundary from the row rather than from a constant is the version that cannot drift.

**The lookback clamp.** This also closes the `windowStart` finding from my earlier list (stalled cursor re-delivering replies whose tombstones the purge already removed) — and the comment names the exact mechanism, that `sentRetention` is shorter than the stall. Two findings, one clamp, with the reasoning written where the next person hits it.

The constant's doc rewrite is worth calling out too: "ceiling on how far back any scan reaches… not a seed for a fresh cursor" is precisely the distinction whose absence caused the original bug.

I verified `TestReconciler_FreshCursorDoesNotBackfillAlreadyDeliveredReplies` catches it — restoring the backdated seed fails it.

## The scode — fixed

Ran my original probe against your branch:

```
wecom provider: http: Get 127.0.0.1:51565/ai/qc/query_result: EOF
```

Credential gone, host and path kept. Routing `url.Parse` errors through `withoutURL` too is the right instinct — those carry the raw URL as well, and it would have been easy to fix only the one path I reported.

`internal/integrations/channel/outbox` and `internal/integrations/wecom` both pass against a real Postgres at migration head.

## Still open from the original review

**#1, the disposition read.** #6590 is up for this — it needed the callback worker as well, since the read loop is the sole deliverer of acks (a callback that waits for one would wait on itself). Once it lands, `retryableSendError` can switch on `*wecomAPIError.Code` instead of string-matching, and `DispositionSent` starts meaning the message arrived.

The **length budget** is the other half and still unaddressed here — a reply over 20480 bytes is refused whole today. I have `splitForWire` ready but deliberately have not filed it, because it wants to sit on your send path rather than race you on the same lines. Say when.

Leaving this as a comment rather than clearing changes-requested, since #1 is still open. Happy to flip it the moment the verdict read is in — the two blockers you have fixed were the ones that could not be repaired after the fact.

seacen:
**Two replies to the same WeCom chat can come out in the wrong order after one transient send failure.**

Framing first: we're retiring our WeCom-specific outbound queue in favor of this one. Yours is better shaped and the cross-replica claim model is what we should have built. We went back through everything ours had learned the hard way against a live tenant and checked each of it against yours — this is the only one I couldn't find an equivalent of. Your call whether the design needs to answer it; I'd rather hand it over than have you rediscover it.

### What we hit

An answer produced while a drain was running got out ahead of an answer that was already waiting. In a WeCom group where two people had asked at the same time, each of them read the other's answer as the response to their own question. Nobody reported a bug — the messages both arrived, in an order that looked deliberate.

### Where the same shape lives here

Three lines:

1. `channel_outbound_queue.sql:72` — the claim filters `next_attempt_at <= now()` and sorts `ORDER BY q.next_attempt_at, q.created_at`. No predicate on `target_chat_id`.
2. `consumer.go:372` — `retryOrFail` sets the failed row's `next_attempt_at = now + Backoff(...)`, which puts it *later* than rows enqueued after it, since migration 265 defaults `next_attempt_at` to `now()`.
3. `consumer.go:183` — `if worked { continue }` re-claims immediately, so the next due row for the same target is sent while the older one waits out its backoff.

Deterministic instance, on the path this queue exists for — draining a backlog after a reconnect. Replies A (older) and B (newer) queued for one WeCom group, which is one target per `wecom_resolvers.go:202-206`. The consumer claims A; `sendText` fails with `broken pipe`; `retryableSendError` (`outbox_sender.go:158-165`) classifies it retryable; `retryOrFail` pushes A to `t_fail + jitter`. The socket tears down and `consumerCtx` is cancelled (`wecom_channel.go:275`) with B never claimed. On reconnect a fresh consumer drains: B's `next_attempt_at` is still its enqueue time, earlier than `t_fail + jitter`, so B is claimed and written first. Ran against your `Consumer` unmodified, 8 times: 8/8 delivered `[B, A]`.

I first wrote this up as a narrow trigger and that was wrong. It needs one row to fail transiently while a later row for the same chat succeeds — the ordinary socket-down case. With full-jitter `Backoff`, order becomes a coin flip whenever two rows for one chat both fail; across a reconnect it isn't a coin flip at all. Same shape when an inbox card and an agent reply are both bound for one 1:1.

What it costs: both messages arrive, so nothing retries, nothing alerts, no operator sees a failure. Not self-correcting either — the late reply lands seconds after the newer one with nothing marking it as late, and the reader's only option is to reinterpret the conversation. The damage is a misread thread, not lost data.

On contract: `doc.go` doesn't promise ordering and the comment at `rategate.go:129` is true where it's scoped, so nothing here contradicts what's documented. It's a property the code being deleted had and the replacement doesn't — flagging it so that's a decision rather than a side effect.

### A shape that fits your design

Defer the target, not just the row. When `RetryClaimedChannelOutbound` pushes a row's `next_attempt_at` forward, push it for every other `queued` row with the same `(installation_id, target_chat_type, target_chat_id)` and a later `created_at`, so a conversation can't step over its own stalled message. One extra UPDATE on columns the claim index already carries, and it leaves cross-chat independence alone — which is the right scope. We landed on the same line: the guarantee is per-conversation, and two rooms never needed to be ordered against each other.

Worth a test that fails without it. `grep -niE "order|sequence|overtake" outbox/*_test.go` returns only an unrelated rategate comment, so nothing in the suite would notice today.

### What doesn't transfer from ours — ignore these

- **Counting the in-flight message in a depth check.** A fix for an in-memory queue where a popped message was invisible to the check. Your `claimed` state keeps the row visible; nothing to count.
- **Requeue-at-head and stop the drain.** Needs a single-process drain loop over one queue. In a claim/lease model there's no head to put anything back on; `next_attempt_at` is the only knob involved.
- **Our table as restart rescue behind an authoritative in-memory queue.** The premise doesn't survive multiple replicas, which is a large part of why we're dropping ours.

seacen:
Two findings from running this branch against a long-reply splitter we are building on our side. Both are about where a message that exceeds WeCom's per-message cap gets cut, and both are design points rather than bugs in what you have written.

Context for why we hit them: WeCom refuses a single `aibot_send_msg` past 20480 bytes with `errcode=45002` and writes nothing, so any answer longer than that has to be split into several messages. We are adding that split, and where it goes changes once this PR lands.

## 1. Splitting inside `queueSender.Send` would re-deliver pieces on retry

A queue row settles as one `Disposition` — Sent, Retry or Failed. If the split happens inside `Send`, one row becomes N frames, and a failure on piece 3 retries **the whole row**, re-delivering pieces 1 and 2. There is no way to express "two of three landed", so the choice is duplicate delivery or losing the tail.

The place that avoids it is the producer: split at enqueue time, **one row per piece**, with the piece index folded into `SourceID` so each piece keeps its own idempotency key. `Request.SourceKind` + `SourceID` is already the dedup fence, and its doc says it should name the business result — a piece of a split answer is its own business result.

Ordering survives that: `ClaimChannelOutbound` is `ORDER BY q.next_attempt_at, q.created_at ... LIMIT 1`, one row at a time per installation, so pieces enqueued in order are claimed in order.

`queueSender.Send` would then keep a defensive length check and dead-letter an over-cap row rather than retrying it eight times against a server that will refuse it identically every time.

## 2. `ClaimChannelOutbound` has no tiebreaker on `created_at`

Rows inserted back to back can land on the same `created_at` — the same statement timestamp inside one transaction, or simply the same microsecond. The claim orders by `(next_attempt_at, created_at)` with no third key, so for tied rows the order is whatever the plan happens to produce.

Today that is invisible: each row is an independent message and the order between two of them barely matters. It stops being invisible the moment one logical answer is several rows, because "part 2 of 3" arriving before "part 1 of 3" is a visible defect in the chat.

A `part` column, or any monotonic per-installation sequence, in the ORDER BY would make it guaranteed rather than probable. Cheap to add now, awkward to add after rows exist.

## Not asking you to do anything with the splitting itself

We will carry the splitter on our side and rebase it onto this branch's shape once it lands. Point 1 only matters if you would otherwise expect the split to live in the sender. Point 2 is worth a line either way, since it is a property of the claim query rather than of anything we are adding.

seacen:
# #6581 — what should happen to a reply that has to go back on the socket it arrived on?

One question, not a review: does the queue mean to carry the in-window reply too, or only messages addressed by a chat id? We have four call sites on the path this PR replaces and one of them cannot become a row, so we would rather know your intent than assume it.

## What sits on that path

`NewOutbound(q, senders, streams, logger)` becomes `NewOutbound(q, producer, logger)`, and four things in our branches hang off the old signature — the streaming bubble (#6606) and outbound attachments (#6604):

- **`finishStream`** — writes the finished answer into the bubble the typing indicator opened for this turn and seals it, as `aibot_respond_msg`.
- **`sendRest`** — the pieces of a long answer that did not fit in that sealed frame, as ordinary `aibot_send_msg`.
- **`rounds().settle`** — no wire traffic. It records that the round's ending has been delivered, so a `task:failed` arriving late does not print a failure notice under an answer the user has already read.
- **`deliverAttachments`** — `aibot_upload_media_init/chunk/finish`, then the file as its own message.

The middle two are ordinary chat-addressed sends and belong on your queue. The bubble is the problem.

## Why the bubble cannot be a row

`aibot_respond_msg` is addressed by the `req_id` of the inbound callback that opened the turn. Only the connection that received that frame holds it, and it is refused without one — `ws_sender.go`: `"wecom: stream frame requires the callback req_id"`. A row claimed a moment later, on a replica that never saw the callback, has nothing to echo.

So this is precisely the case your header calls out — "a reply can only leave the process that holds the socket", the reason `outbound.go:21` on main carries the single-replica warning — and it is the one case durable enqueueing cannot fix, because the fix is chat-id addressing.

Volume rules it out separately. An open bubble is refreshed in place at most every 1.5s for as long as the run takes, so one message would be hundreds of rows rewriting each other.

## How we resolved it locally, so you can see the assumption

Our self-hosted deployment runs both, so we had to pick something to keep going:

- ordinary reply and inbox card → `producer.Enqueue`, your code unchanged;
- bubble frames → straight to the socket, and a bubble that cannot be opened or sealed falls through to the queue. A replica without the socket therefore opens no bubble and the answer arrives as an ordinary queued message, which is the behaviour we want anyway;
- attachments → also straight to the socket. They are not `req_id`-bound: the three upload cmds mint their own req_ids and the finished file goes out addressed by `chatid`, so a queue could carry them. We kept them off it because one file is up to 50 MB over a hundred acked chunks under a five-minute budget, while `ClaimLease` is 30s and a worker claims one row per installation — a file would expire its own lease and hold every text reply behind it.

That is a guess about your design, not a proposal. You may already have a plan for in-window responses, or you may consider them out of scope for the queue and expect the socket path to stay for exactly this; either answer tells us what to build against.

## One knock-on if the answer is "out of scope"

This PR drops the WeCom single-replica exception from `SELF_HOSTING.md`, which is right for the tree as it stands. With a bubble in the tree it becomes half true: delivery is replica-agnostic, but a bubble opened by the socket holder is only sealed when the completion is published on that same replica. Anywhere else it spins until the five-minute guard closes it and the answer turns up as a separate message. Worth a sentence there if the bubble lands after this.

seacen:
**Ask: keep `senders` on `NewOutbound` next to the producer, and let the call site choose by how the frame is addressed.**

```go
// this PR
func NewOutbound(q outboundQueries, producer *outbox.Producer, logger *slog.Logger) *Outbound

// proposed
func NewOutbound(q outboundQueries, senders *sendersRegistry, producer *outbox.Producer, logger *slog.Logger) *Outbound
```

`wecomSenders` is already in scope at `router.go:795`, so the wiring cost is one argument. Nothing else about the queue changes.

This is the answer to [the question I left earlier](https://github.com/multica-ai/multica/pull/6581#issuecomment-5226188032) rather than a fourth demand — we have since built and run the answer, so I can propose instead of ask. And what I am asking for is a deletion left undone, not work added.

## The queue does two jobs, and one of them does not generalise

**Routing** — who is allowed to send this. The event fires on whichever replica ran the task, the row is claimed by whichever replica holds the socket. For a chat-addressed message that is the fix, and nothing smaller works.

**Durability** — what happens when the send fails, the process dies, or the target is over quota. Also the queue, also right.

A `req_id`-addressed frame can take the second job but not the first. `aibot_respond_msg` has to echo the req_id of the inbound callback that opened the turn; the server refuses a wrong one with 846605, and #6606's `respondStream` refuses an empty one outright. That id lives in the memory of one connection, inside the protocol's stream window. A row claimed seconds later on another replica has nothing to echo. No store-and-forward mechanism can carry it, so this is not something your implementation could have done better.

Attachments are a different case reaching the same conclusion. The three upload cmds mint their own req_ids and the finished file goes out addressed by chatid, so the queue could carry them. The arithmetic rules it out: one file is up to 50 MB over a hundred acked chunks under a five-minute budget (#6604), while `ClaimLease` is 30s and a worker holds one row per installation. The upload would expire its own lease and hold every text reply for that bot behind it.

## What we run

Our self-hosted deployment carries your branch with both paths live:

- ordinary reply and inbox card → `producer.Enqueue`, your code untouched;
- the sealed bubble frame, and the remainder of an answer too long for one frame → the socket;
- attachments → the socket, detached, after the words are out.

The choice is made per turn rather than fixed per message type. A replica holding no socket for that installation claims no bubble at all, so the answer leaves through the queue as an ordinary message; a bubble whose frame is refused falls through to the queue carrying the whole answer. The queue still catches everything; it just is not the only way out.

What it cost us to learn that: your reconciler's only "already delivered" test is `NOT EXISTS` against `channel_outbound_queue`, so a socket delivery looks exactly like a reply the realtime path lost, and every bubble-delivered answer arrived twice. That is correct given the premise that everything goes through the queue. The general constraint it implies is that any path bypassing the queue has to record delivery. We are fixing that on our side and will post the details separately — nothing for you to do here.

## leroy-chen's note in `outbound.go`

You removed the `SINGLE-REPLICA CONSTRAINT` comment from #5833, along with the same statement in `.env.example`, `SELF_HOSTING.md`, `types.go` and the boot warning. That was right when you wrote it — the last clause is "until outbound is routed to the lease holder, a WeCom-enabled backend must run as a single replica", and this PR routes it.

#6606 makes it partly true again. Delivery becomes replica-agnostic, but a bubble opened by the socket holder is only sealed when the completion is published on that same replica; anywhere else it spins until the five-minute guard closes it and the answer turns up as a separate message. If #6606 lands after this, that is worth a sentence in `SELF_HOSTING.md`. The constraint stopped being true, then started being true again for a different reason.

To be clear about the direction: this PR is strictly better than `main` today. On `main` a reply produced off-lease reaches a nil sender and the user gets nothing at all (`outbound.go:126`). Yours delivers it. The one deletion I would ask you to reconsider is only visible because our bubble landed after you wrote this.

seacen:
## Addendum: a delivery that bypasses the queue has to record itself

Following on from the question above, with something we only found by running both halves against a live bot. This is not a second defect report — your reconciler is doing the right thing with what it can see. It is a constraint the design has, and we walked into it.

First: thanks for `b8186747f`. The blocker we reported — a fresh cursor backfilling a day of already-delivered replies on upgrade — is fixed, and it is a different problem from this one.

### What we saw

Two of these in the backend log, a minute apart:

```
WRN channel outbox reconciler: rescued a reply the realtime path missed
    channel_type=wecom task_id=56ae568c source_kind=chat_done
```

The user read the answer in the streaming bubble, then read it again about a minute later as an ordinary message.

### Why

The premise the candidate scan rests on is that every outbound send is an `Enqueue`, so a terminal task with no queue row is a reply the producing replica dropped. That holds for your tree. It stopped holding in ours the moment we added the in-window reply this PR's thread is about: our chat-done subscriber seals the bubble over the socket and returns before it reaches `producer.Enqueue`, so no row is ever written. `ListChannelOutboundReconcileCandidates` finds the task, `NOT EXISTS` is satisfied, and the reconciler concludes — correctly, from where it stands — that the reply never went out.

Stated generally, and this is the part worth having written down somewhere in the package: **absence of a row means "not delivered" only while every delivery is an enqueue.** Any adapter that grows a send the queue cannot express owes the queue a record of it.

### What we did locally

The tombstone concept is already yours — migration 267 says a `sent` row must keep suppressing a re-enqueue of the same business result, and the candidate scan is status-agnostic for that reason. We added a way to write one directly rather than only by draining:

```sql
-- name: RecordChannelOutboundDelivered :one
INSERT INTO channel_outbound_queue (..., status, sent_at)
VALUES (..., 'sent', now())
ON CONFLICT (installation_id, source_kind, source_id) DO NOTHING
RETURNING *;
```

behind `Producer.RecordDelivered(ctx, Request)`, under the same business key the `Enqueue` it stands in for would have used, with `payload` left empty. Nothing downstream needs to know how the row got there: the unique index dedupes it against a racing reconcile insert, the candidate scan skips the task, and `PurgeSentChannelOutboundQueueBefore` ages it out on the same retention. We report it through `Metrics.RecordEnqueued` under a third path, `direct`, so it does not read as queue backlog or as realtime throughput.

That is what we are running. If you would rather the package expressed this differently — or expected the adapter to keep such a record on its own side — say so and we will follow your shape rather than ship ours.

### Two things that fall out of the same premise

**`task_failed` has no realtime producer.** In the tree as it stands, the reconciler's `PayloadBuilder` is the only thing that ever writes that kind, so the failure notice arrives on the reconcile path by design. Any adapter that announces a failure itself — ours does, in the bubble — produces the same duplicate on that ending too, in different words. Same fix, worth knowing before another channel adopts the queue.

**Answer and attachments end up on different transports.** WeCom files cannot ride the queue: one upload is up to 50 MB over a hundred acked chunks against a 30s `ClaimLease`, and it would park every text reply for that bot behind it. So the words go out as a row and the files over the socket, and nothing orders the two against each other — usually the wake plus one socket write wins, which is not a guarantee, and a file landing between a question and its answer reads as an interruption. We pinned both to the socket for turns that carry files, so the ordering is a property of the code rather than a race. Mentioning it because it is a consequence of the split, not because anything needs to change on your side.

### One question while reading the scan

The candidate query takes `agent_task_queue.status = 'failed'` at face value, including an attempt `FailTask` has already created a retry for. Is `retry_pending` meant to be invisible here? As written, a run that fails once and is retried gets a "could not handle that message" notice and then the retry's answer underneath it.

wzxjohn:
Many thanks to @seacen for the incredibly thorough and patient review. These suggestions are all excellent!

1. Intent: the queue is chat-id addressed only

Your reading is right, and it deserves to be stated as a boundary rather than left implied.

The queue does the two jobs you named, and routing is exactly what destroys req\_id addressing. A row is by definition claimable by whichever replica holds the lease when it comes due; aibot\_respond\_msg needs the
one connection that received the callback, inside the protocol's stream window. Those requirements are contradictory, so req\_id-bound frames are out of scope by construction, not by omission. No store-and-forward
could carry them — you're right that this isn't something the implementation could have done better.

Your volume point is a second, independent reason: a bubble refreshed in place every 1.5s for the length of a run is not a row, and would be hundreds of rows overwriting each other.

So: the socket path stays for exactly this case. rounds().settle has no wire traffic and so has nothing to do with the queue either. sendRest and the inbox card are ordinary chat-addressed sends and belong on it,
as you have them.

On attachments — agreed, and the arithmetic is the deciding factor, not the addressing. I checked: ClaimLease = 30 \* time.Second (consumer.go:29) and a worker claims one row per installation, so a 50 MB /
\~100-chunk / 5-minute upload would expire its own lease and park every text reply for that bot behind it. Keeping them on the socket is right.

2. senders back on NewOutbound — agreed

You're correct that this is a deletion left undone rather than new work, and I'll land it.

One wrinkle worth naming: after the queue migration, Outbound has zero remaining references to senders, so the parameter arrives unused in this tree. That's legal and costs nothing mechanically — but an
unexplained unused argument is precisely how it got deleted the first time. So I'll land it with a comment naming the boundary (req\_id-addressed frames cannot become rows, so the call site chooses by addressing),
which puts the answer to question 1 in the file instead of in this thread.

3. SELF\_HOSTING.md

Agreed, and it belongs in #6606, not here. In this tree nothing opens a bubble, so the sentence would describe a constraint no code exhibits — which is the kind of warning the next reader deletes as stale,
exactly as happened to the original. The constraint stopped being true and then starts being true for a different reason, so it should land with the change that makes it true again. 

4. RecordDelivered — yes, put it in the package

Your shape is the right one and I'd take it close to as-is. It belongs in the package rather than the adapter because the premise it protects — absence of a row means "not delivered" — is the reconciler's
premise, and the reconciler is package-owned. An adapter-side record has to be re-invented by Lark and Slack, and the first one to forget gets silent duplicate delivery. Migration 267 already says a sent row's
remaining job is suppression, so writing one directly is using the mechanism as designed.

Three notes from reading it against this tree:

- Retention keys on updated\_at, not sent\_at (PurgeSentChannelOutboundQueueBefore). Let updated\_at default on insert and the tombstone ages out on the normal window; if you only set sent\_at it still works, but the
  purge doesn't read that column.
- It composes with the ordering change I just landed (5234dbf78): retry and defer now postpone later queued rows for the same target. A tombstone is inserted sent, so it's outside both the claim and the hold.
  Nothing to adjust.
- direct as a third metrics path is right, for the same reason the enqueue paths were split.

5. retry\_pending — meant to be invisible, and it isn't. That one's mine

Confirmed by reading it, and it's a real bug in my tree independent of your bubble:

MaybeRetryFailedTask creates a child task, so the failed parent keeps status='failed' with completed\_at set. retry\_pending exists only in the task\:failed event payload (task.go:5026) — invisible to the table the
scan reads. outbox\_builder.go:129 then builds the notice unconditionally.

It's worse here than in your tree for the reason you spotted: task\_failed has no realtime producer (Outbound.Register takes chat\_done and inbox\:new only), so every failure notice comes from the reconciler. Any
WeCom user whose task fails once and is auto-retried gets "⚠️ agent failed" followed by the retry's answer. Not bubble-specific, not self-hosted-specific.

The table can express it: agent\_task\_queue.retry\_of\_task\_id (migration 184) is the retry lineage, so the scan can exclude a failed task that already has a retry child. There's no index on that column today, so
the fix is one predicate plus one CREATE INDEX CONCURRENTLY. DingTalk reached the same conclusion from the event side (dingtalk/outbound.go:139 returns empty content while a retry is pending) — this puts the
table-side scan in line with behaviour the event-side path already has.

I'll take that as the next commit unless you'd rather own it.

wzxjohn:
As an aside, I would **not recommend** enforcing strict ordering for WeCom file/message delivery.

We use WeCom bots heavily in production, and even with a strictly serial send queue on our side, we still occasionally see messages arrive out of order. This appears to be a consequence of WeCom's internal message-routing and delivery architecture: messages we submit are processed across WeCom's distributed service cluster.

As a result, when two messages are sent very close together, preserving their order when we send them **does not** necessarily mean they will **arrive** at the client **in the same order**.

If strict end-to-end ordering is required, the practical approach is to introduce a sufficiently large gap between sends — for example, something like `sleep 3` — so that the previous message has very likely reached the client before the next one is sent.

seacen:
Thanks — that's clear, and nothing is left on our side for the first three.

**`senders`, `RecordDelivered`, `retry_pending`: all yours.**

On `RecordDelivered`, your three notes are right, and the `updated_at` one catches a mistake in ours. We set `sent_at` and wrote the doc comment as though that were the column retention ages the tombstone out on. It isn't — `PurgeSentChannelOutboundQueueBefore` reads `updated_at`, which is only correct in our version because the column defaults on insert. It works by the default, not by anything we did. Your shape is the right one. Agreed too that the tombstone composes with `5234dbf78` (inserted `sent`, so outside both the claim and the hold), and that `direct` deserves its own metrics path.

**`retry_pending` — please take it.** Your table, your scan, your lineage column, and you have already worked the fix down to the missing index on `retry_of_task_id`. We would only be re-deriving it.

**Ordering — we are not adding the sleep, and your note changed what we claim.** The production observation is the useful part. We had written our comments and test names as though putting the answer and its files on one transport made the order the user sees a property of the code. It does not, and we have corrected that rather than defended it: the code now says we send in order and no longer reorder a turn ourselves by splitting it across two transports, with your routing explanation written down beside it so the stronger claim does not come back.

We are keeping the one-transport behaviour, since the reordering it removes is the reordering we were introducing. The gap we are not taking: it buys no guarantee either, and it charges a fixed delay to every turn that carries a file. `aibot_upload_media_init/chunk/finish` also has to round-trip before the file can be sent, so the common case is already spaced without paying for it.

**`SELF_HOSTING.md`** — into #6606, for the reason you gave. Written for the new cause: the bubble is opened by the replica holding the socket and can only be sealed on that connection, and the seal comes off the in-process `chat:done` event, which fires wherever the task ran. On two replicas a task that lands on the other one leaves the bubble spinning until the five-minute guard and then answers as a separate message. It replaces the existing paragraph on that branch rather than sitting beside it, so whichever of us rebases second gets a conflict and keeps the surviving version.

wzxjohn:
## `ef6726e2` — errcode, not prose

`sendTextCtx` returns the verdict now, but `Send` still classified with
`strings.Contains` over the error text — which gets every stated refusal wrong the
same way. `errcode=45009` matched no transient substring and dead-lettered: the one
refusal certain to succeed later, and exactly what the rate gate exists to absorb.

- 45009 / 45033 / −1 retry; any other stated refusal is terminal. `errmsg` is never
  consulted — it is WeCom's prose, and a transient-looking word in it would flip a
  permanent refusal into eight pointless retries. A test pins that.
- `errAckTimeout` and `context.Canceled` now retry; both were terminal, so a
  shutdown mid-send dead-lettered a healthy reply.
- The consumer's `ctx` reaches the send, so a lease loss ends the 5s verdict wait
  instead of spending the row's attempt.
- `maxSendContentBytes` dead-letters an over-cap row on attempt 1, not 8.
  Deliberately a floor, not a splitter — a row settles as one `Disposition`, so
  splitting inside `Send` cannot express "two of three landed".

**A judgement call, not a forced move:** retrying `errAckTimeout` can duplicate a
message that landed. Bounded attempts cap it, and a duplicate is recoverable by a
reader where a silent drop is not. Reasonable to overrule.

## `e376ab63` — ordering, settling, a failure notice that was not one

**Order.** The claim took due rows in `(next_attempt_at, created_at)` order with no
notion of a conversation, and `next_attempt_at` defaults to `now()` — so a row that
fails and takes its backoff is scheduled *later* than a reply enqueued after it.
Both messages arrive, nothing retries, no operator sees a failure; in a group where
two people asked at once, each reads the other's answer as their own.

Enforced from both ends, because either alone leaves a hole: retry and defer
postpone the rest of that chat's queue (`GREATEST`, so nothing is pulled earlier),
and the claim exposes only a conversation's oldest queued row — which covers the
successor inserted *after* its predecessor was postponed, that never had a
`next_attempt_at` to push forward. Scoped to `(installation, chat_type, chat_id)`;
a test pins cross-chat independence.

The **defer path is beyond the reported scope**, included because the rate windows
are sliding counts: a later row claimed moments afterwards can find an attempt has
aged out, be admitted, and answer first. The comment claiming order survived either
way was wrong.

`seq BIGSERIAL` is the total order it rests on. `created_at` is not one — `now()` is
the transaction timestamp, so rows from one statement tie and came back in plan
order. Invisible per-message, visible the moment one answer is several rows.

**Settling.** The settling UPDATE ran on the consumer's context, so a shutdown
between a successful send and that UPDATE left the row queued → next holder
re-delivers, silently. `complete` / `retryOrFail` / `terminate` now use
`context.WithoutCancel`, bounded by `settleTimeout`.

**Retried failures.** `MaybeRetryFailedTask` creates a *child*, so the re-run attempt
keeps `status='failed'`; the scan took that at face value and delivered "could not
handle that message" with the retry's answer underneath. `retry_pending` is only in
the event payload, so the scan now reads `retry_of_task_id` (migration 279). Note
`task_failed` has **no realtime producer**, so every WeCom failure notice comes from
the reconciler and every auto-retried failure hit this.

## `622b3be2` — the boundary, and two claims that were wrong

`senders` is back on `NewOutbound` beside the producer, with the rule in its doc.
Nothing in the file uses it — every send there is chat-id addressed — so the field
says what it is for, because an unexplained unused argument is how it got deleted.

Both halves of the old justification were wrong:

- **No 5-second deadline on `aibot_respond_msg`.** 5s applies to
  `respond_welcome_msg` / `respond_update_msg`; a message-callback reply has **24
  hours**, a stream **10 minutes**. The number had no source in the tree.
- **`req_id` is not connection-bound.** Measured on a live bot: a stream opened on
  one connection took a refresh *and* its closing frame from a second connection
  after the first was displaced (errcode 0 both times), while a random `req_id`
  was refused 846605 — so the server does validate it.

The conclusion holds, the reasoning changes: bubble frames stay off the queue
because of traffic shape (6–10 min lifetime, ~1 frame/second), not addressing. **This
opens what the old comment wrongly closed** — persisting the `req_id` so a replica
taking the socket over can continue somebody else's bubble.

Also records the `aibot_send_msg` prerequisite: it only reaches a conversation the
user has already written to the bot in.

## `9ecbc3d3` — the scan install path

- **Broken over plain HTTP.** `crypto.randomUUID` is SecureContext-gated, so on
  `http://` it threw inside `beginSession` and the dialog showed "something went
  wrong" with no request ever sent. Now `createSafeId`.
- **Flag on the wrong schema.** `scan_install_supported` was declared on the DingTalk
  response; the WeCom listing that sends it declared nothing, and it worked only
  because the object is `.loose()`. The test asserts a **non**-boolean is rejected —
  asserting the happy path passes either way.
- **No lease fence on `FailWecomInstallSession`.** A stale worker could mark a
  session error out from under the new owner and NULL the `scode` it needs to
  finish. `processOne` always claims first, so the lease was in scope at all eight
  call sites.

## Schema

| Migration | Change |
| --- | --- |
| `265`, `266` (amended) | `seq BIGSERIAL`; claim index gains `seq` |
| `279` (new) | `agent_task_queue (retry_of_task_id)`, partial |
| `280` (new) | target-local order index for the predecessor probe |

`265`/`266` are amended rather than followed, since both are new in this PR and
unshipped. `sqlc` regenerated, no drift.

seacen:
**Two of the three landed and do what you said they would. `RecordDelivered` is not in the branch — that is the only thing outstanding, and nothing else in the four new commits breaks anything we depend on.**

## 1. `senders` on `NewOutbound` — landed

`622b3be2`. `NewOutbound(q, senders, producer, logger)`, `wecomSenders` wired at `router.go`, the field carries the comment and the constructor doc carries the rule. Our call sites take it as-is: we pass the registry in the same position, so it is one argument for us too.

The correction is the more useful half. The 5-second deadline had no source in the tree and we had been quoting it as well. The connection-binding claim was ours, stated in this thread, and your measurement says it is wrong — so your version is the one to keep. It also re-opens something we had written off: persisting the `req_id` so a replica that takes the socket over can continue a bubble it did not open. Not doing that now, but it is no longer closed.

The `aibot_send_msg` prerequisite is worth having in the file too. That bound was nowhere in the code.

## 2. `RecordDelivered` — not in the branch

`Producer` still has only `Enqueue`, and there is no `RecordChannelOutboundDelivered` query. I assume it is just behind the other three.

Flagging it because `622b3be2` landed the rule without the mechanism. `NewOutbound`'s doc now says a path that bypasses the queue owes the queue a record of the delivery, and the only call available for keeping that promise is `Enqueue` — which writes a `queued` row a consumer then drains, the duplicate the sentence warns about. Nothing in this tree reaches it: the three remaining direct writes are the binding prompt (`replier.go:203`, `:226`) and the non-text receipt (`wecom_channel.go:457`), all inbound-triggered and none task-derived, so the candidate scan never looks for them. The first channel that adds one pays for it.

Your third note checks out, by the way, and I checked it against a migrated database rather than by reading: a tombstone inserted `sent` is invisible to both halves of the ordering change — the predecessor probe and the hold both filter `status = 'queued'` — so the claim still hands out that chat's next queued row.

## 3. `retry_pending` — landed, and it covers the case

`e376ab63`: migration 279's partial index on `retry_of_task_id`, one predicate in the candidate scan, scoped to `status = 'failed'` so a completed task is still a delivery.

It covers both ways a retry child gets created, which is the part I went looking for. `FailTask` creates it inside the same transaction (`task.go:3946`), and `HandleFailedTasks` creates it in the same call that surfaces the failure. Both are far inside the 30-second settle delay, so the scan never sees a retried parent in the first place. A manual rerun writes `rerun_of_task_id` instead and still gets its notice, which is right — the user asked for that one.

`TestReconcileCandidates_ARetriedFailureIsNotACandidate` is DB-backed and runs; I checked it was not quietly skipping.

## Against what we are carrying

Ran the branch on a migrated Postgres and looked at what the bubble and the attachments depend on.

- All six `TestClaimOrder_*` run and pass. Neither the hold nor the predecessor probe can see a socket delivery, so the ordering change leaves that path alone.
- `ClaimLease` is still 30 seconds and the claim is still one row per installation, so the arithmetic that keeps a 50 MB upload off the queue is unchanged.
- `maxSendContentBytes` in `ef6726e2` sits above where our splitter currently lives — ours is inside `sendTextCtx`, below `queueSender.Send` — so on our branch an over-cap row would dead-letter before reaching it. That is the move to splitting at enqueue time, one row per piece, that we already agreed on, and `seq` is what makes it safe. Nothing for you here; saying it so the order of the two changes is clear.

`go build`, `go vet` and `gofmt` are clean on the touched packages, and the `outbox` and `wecom` suites pass against a migrated database.

Also confirming `9ecbc3d3` fixes the three scan-install problems as reported: `createSafeId`, the flag moved onto the WeCom response with a test that pins a non-boolean, and `FailWecomInstallSession` matching on the lease token.

seacen (approved):
Approving. Everything raised is landed or answered, and I checked the ordering change against a migrated database rather than by reading it.

**Correcting myself from the comment above**: I wrote that `RecordDelivered` was "just behind the other three", which reads as though it is yours to land. It is not. It is ours, and it is not a condition on this PR.

Nothing in your tree reaches the case — the three remaining direct writes are inbound-triggered and never task-derived — so it costs you nothing to merge without it. It costs us: our streaming bubble answers over the socket and writes no row, so the reconciler sees a completed task with no delivery and sends the answer a second time.

**We will send it as a small follow-up right after this merges**, in `outbox`, keyed the way the `Enqueue` it replaces would have been, retention on `updated_at`, `direct` as the third metrics path — the shape you described. Saying it here so we do not both write it, or neither. If you would rather it come from you, say so and we will drop ours.

Worth having either way, because `622b3be2` landed the rule without the mechanism: `NewOutbound`'s doc now says a path that bypasses the queue owes the queue a record of the delivery, and the only call available for keeping that promise is `Enqueue`, which writes the duplicate the sentence warns about. The first channel that adds a bypass path pays for it.

@Bohan-J — no objections from our side, this is ready for you whenever you have a slot.

wzxjohn:
@seacen I intentionally did not add `RecordDelivered` because the current tree does not actually have any path that needs it yet. To avoid conflicting with your streaming implementation, I only documented the contract for now rather than introducing a mechanism that nothing in this branch uses.

Also, persisting the `req_id` is definitely worthwhile. I have already tested this with a script: a stream opened by instance A can be continued by instance B.

So if the stream disconnects before the timeout and the connection is re-established correctly, the replica that later takes over the socket can absolutely continue updating the existing stream.

wzxjohn:
@Bohan-J Would it be possible to merge this PR soon?

There are quite a few migrations in this branch, and the longer it stays open, the more likely it is that upstream migrations will keep landing and force us to repeatedly renumber/reconcile the migration sequence during rebases.

Since the review items have now been addressed and the PR is approved, merging it sooner would help avoid unnecessary churn around migration numbering.

Bohan-J:
@wzxjohn — answering directly: **I'm not going to merge this as one PR.** I'm asking you to split it, the way I asked in #6524 and you agreed to there. That answer took a day to reach you while the renumbering churn you're describing kept accruing, which is on me — there's something concrete about that below.

**This is not a quality call.** @seacen's review here was as thorough as any this repo has had, everything raised is landed or answered, the approval is real, and CI is green at `4ad3fb37`. The work stands. What I'm declining is the packaging.

## Why the packaging matters here specifically

The branch is three capabilities: the cross-channel durable outbound queue (migrations 281–286), the per-target rate gate (287–289), and scan-code install for creating a new bot (290–294).

They need different acceptance and they have different rollback:

- The **queue** changes the delivery path for every WeCom reply and deletes the single-replica constraint from `.env.example`, `SELF_HOSTING.md`, and the boot warning in `router.go`. Accepting it means running multi-replica for a while and deliberately moving a lease to watch the reconciler and the per-chat ordering hold up.
- **Scan-code install** is a new admin journey with an external QR provider and two new encrypted bearer credentials on a table. Accepting it means a real WeCom admin scanning a real QR, behind an operator flag we can leave off.

Bundled, I can't stage one without the other going live, and I can't revert one without taking the other with it. That matters more for us than it would for most reviewers precisely because we don't run WeCom ourselves — being able to pull one piece back without disturbing the other is most of the safety margin we have.

The rate gate is the exception: it hard-depends on the queue and is meaningless without it, so those two share a fate regardless.

## What I'm asking for

Three PRs in dependency order:

1. `channel/outbox` — the durable outbound queue, plus the WeCom adapter moving onto it and the single-replica constraint coming out.
2. The per-target rate gate.
3. WeCom scan-code install.

**The one I'd insist on is 3.** I checked `23d230b4`'s file list — it touches no `outbox` file at all, so it lifts out cleanly and it's the one whose product acceptance is genuinely separate. If you'd rather keep 1 and 2 together, I'll take that; the gate is 818 lines and three migrations and doesn't stand alone.

The real cost is the seven fix commits sitting after the three features — they span all three, so they want folding back into the feature they fix rather than replaying on top. Otherwise please keep your existing commits as they are: if the split PRs are the same commits redistributed, @seacen's second pass is a diff of the diff rather than a fresh read, and I'd rather not spend his time re-clearing lines he already cleared.

## On the migration numbering — the thing you actually asked about

You're right that it's churn. The answer isn't to bundle, so here's what I'll do instead:

- **Stop renumbering preemptively.** Leave your numbers where they are. I'll give you the final numbers immediately before merge, or renumber them myself as part of it. Rebasing weekly to chase upstream numbers is wasted effort and I shouldn't have left you doing it.
- **They'll go back-to-back, not one a week.** Send the queue first — it's the one I want most and the other two sit behind it. I'll make it the next review I pick up rather than queueing it behind the backlog.
- Sequencing you can't see from your side: #6648 currently holds 273–280, which is what your last rebase moved above. There are also three open WeCom PRs from @seacen in the same files — #6604, #6606, #6608, with #6606 being the large one. I'll order those against yours rather than letting you all find out in conflicts.

One more thing worth stating plainly: the approval is against `9ecbc3d`, and the head moved after it with the rebase and renumber. Even on the merge-as-is path I'd have asked for a re-confirm against the current head, so the split isn't costing you an approval you'd otherwise have kept.

Credit stays yours in the history on all three, and the co-ownership offer from #6524 still stands.

wzxjohn:
Sounds good, no problem. I’ll split this into three separate PRs as requested.

I’ll also drop the commits that only renumber the migrations and leave the current migration numbers alone for now. We can do the final renumbering once, immediately before merge.

One small clarification on scan-code install: we already have this feature running in production on our side, and it does not require the person scanning the QR code to be a WeCom Admin. Regular users are allowed to scan and create the bot, provided their enterprise policy permits it.

PR #6584, opened 2026-08-07
fix(wecom): throttle binding-token minting to one link per user per minute

seacen:
## What

`BindingTokenService.Mint` inserted a `channel_binding_token` row on every call. Every message from an unbound user reaches the `needs_binding` outcome, so someone who types six lines at a bot they have not linked yet writes six rows and receives six links — only the last of which they will ever click.

Mint now checks for a live, unconsumed token minted inside `BindingTokenMintInterval` first. Finding one, it returns `Reused: true` with an empty `Raw`: the table only ever held the hash, so there is no secret to hand back. `sendBindingPrompt` points the user at the link already in their chat instead of building a URL — which on the old shape would have produced `?token=` with nothing after it, a dead link the user taps and gets refused by.

## Why sixty seconds

Deliberately a burst window, not a session-long one. Suppressing a mint asserts that the earlier link reached the user, and nothing verifies that — the send returns as soon as the transport accepts the frame, and the aibot ack is asynchronous. With a long window, one lost prompt becomes a dead end: every further message is answered with "tap the link I sent you", pointing at something that never arrived, and the user cannot link their account until the window passes.

A minute does the job the throttle exists for (six lines typed in one breath still write one row) and costs at most one row a minute against rows that expire in fifteen. The constant has to stay comfortably inside `BindingTokenTTL` so a link a throttled user is pointed back at still has real time left on it; there is a test pinning that invariant.

## Tests

Six, all runnable without Postgres.

Five drive `Mint` against a fake that applies the same predicates `FindLiveChannelBindingToken` does, rather than a stub that always answers: reuse inside the window (six messages → one row), a fresh link once the window passes, per-user rather than global, a consumed token not suppressing a new one, and the window staying inside the TTL.

The sixth drives the real `sendBindingPrompt` and asserts a throttled prompt carries no bind URL at all. On the old shape it fails with the empty link printed:

```
a throttled prompt built a URL with no token in it:
"👋 请先绑定你的 Multica 账号，才能与我对话：\nhttps://multica.example/wecom/bind?token=\n（链接 15 分钟内有效）"
```

I also checked the throttle test itself fails when the suppression is removed, so neither test is measuring nothing.

## Notes

- No migration. The query rides `channel_binding_token` from `124_channel_generalization`, which already has `created_at` / `consumed_at`.
- `pkg/db/generated/channel.sql.go` is regenerated (sqlc v1.31.1) and the diff is +54 lines, the new query only.
- `#6581` also regenerates `channel.sql.go`. Landing this first means that PR regenerates on top rather than the other way round.

### DIFF
--- a/server/internal/integrations/wecom/binding.go
+++ b/server/internal/integrations/wecom/binding.go
@@ -48,6 +48,13 @@ const BindingTokenTTL = 15 * time.Minute
 // with "tap the link I sent you", pointing at something that never arrived,
 // and the user cannot link their account until the window passes.
 //
+// That covers the send we never hear back about, and equally the send we know
+// failed: the row is written before delivery is attempted, so a prompt the
+// transport refused outright (no live connection mid-reconnect or lease flip)
+// still suppresses the next minute of mints. Nothing here reacts to that
+// error — the window itself, not a delivery receipt, is what bounds the
+// damage, which is the other reason to keep it short.
+//
 // Sixty seconds does the job the throttle was written for: six lines typed in
 // one breath still write one row, at a cost of at most one row a minute for a
 // user who keeps going, against rows that expire in fifteen. The price of
@@ -81,8 +88,8 @@ type BindingToken struct {
 	// already sitting in the user's chat. Raw is empty in that case and there
 	// is no way to recover it — the table only ever held the hash — so the
 	// caller must point the user back at the earlier message rather than
-	// building a URL. ExpiresAt is the live token's, so the caller can say how
-	// much time is left on it.
+	// building a URL. ExpiresAt carries the live token's expiry, not a fresh
+	// one's.
 	Reused bool
 }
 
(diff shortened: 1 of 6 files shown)


Bohan-J:
Thanks for this — a good catch, and the tests are the best part of the PR: five cases driving a fake that applies the query's own predicates rather than a stub that always answers, plus one that fails on the old shape and prints the empty `?token=` link right in the failure output. That kind of test keeps paying off later.

The thing we looked hardest at was `Mint` running `FindLive` and `Create` as two statements — two replies racing on the same user can both miss and both mint. It's real, but the reply path serialises them by a wide margin: the inbound read loop dispatches one frame at a time, and the second reply is a whole dispatch behind the first (dedup claim, sender resolution, drop audit — three or four round trips) while the first needs only one round trip to reach its insert. Losing the race costs one extra row and one extra link, both private to that same user and both on the usual TTL, so there's no security or consistency impact. Not worth holding a lock across the inbound hot path. We've left the logic alone and instead corrected the comments, which promised a per-message guarantee the code doesn't deliver.

Rather than send you round-tripping on nits, I pushed one commit (`ccb56d0`) with three small follow-ups:

- **The reused-link prompt was self-referential.** It goes out through `postPrivate`, which always lands in the 1:1 where the earlier link is already sitting — whichever room triggered it. So `请在与我的单聊里点击` told the reader to walk over to the chat they were already reading. It now points up the current conversation. The group ack does run in the room, and it still names the 1:1.
- **The window mixed two clocks.** It was derived from the application clock (`now().Add(-interval)`) but compared against `created_at`, which the column default stamps from the database clock. Skew between them silently stretched or shrank the window. The interval is now subtracted from `now()` inside the query, so the whole window is measured on one clock.
- **Added a test pinning the width of the window.** The existing two only touch its edges — same instant, and past it — which a zero-length interval satisfies just as well, so neither would have noticed if `Mint` stopped passing a real one.

Locally `go build ./...`, `go vet`, and `go test -race ./internal/integrations/wecom` all pass.

One thing worth knowing for later: `channel_binding_token` is shared with Slack, DingTalk and Lark, and all three still mint unconditionally, so the same one-row-per-message behavior is there. If you ever port this across, note that Slack has **two** `Mint` call sites (`replier.go` and `slash_command.go`) — missing either one reproduces exactly the empty `?token=` link this PR fixes.

Merging once CI is green. Thanks again!

PR #6585, opened 2026-08-07
feat(wecom): show WeCom's own mark instead of a generic speech bubble

seacen:
## What

WeCom renders as lucide's generic `MessagesSquare` in two places — the connect button in Settings and the agent's Integrations card. That is the same glyph Slack and DingTalk use, so nothing on either surface says which platform the row is for.

This adds a `WecomMark` component and points both at it.

## Artwork and licence

`svg/logo-wecom.svg` from Tencent's own [TDesign icon set](https://github.com/Tencent/tdesign-icons), MIT licensed. It is the vendor's own mark rather than a redrawing, which also avoids the trademark problem with approximating someone's logo by hand. Provenance and licence are recorded in the component's header comment.

Same 24×24 box and `currentColor` fill as `GitHubMark` next door, so it drops into a Button beside lucide icons without any restyling.

## Notes

- `MessagesSquare` stays imported in `integrations-tab.tsx` because Slack and DingTalk still use it there; in `wecom-tab.tsx` it was the only use, so the import goes.
- No migration, no server change, no new dependency — the path data is inline.
- `wecom-tab.test.tsx` and `integrations-tab.test.tsx` both still pass (15 tests). Neither asserts anything about the icon, and I did not add one: an assertion that a component renders a particular SVG would only restate the implementation.


multica-eve (approved):
Thank you for the thoughtful contribution! The focused implementation, clear asset provenance, and reuse of a single WeCom mark across both surfaces are much appreciated. The final review is complete, and this is ready to merge.

PR #6586, opened 2026-08-07
fix(wecom): cap the request bodies both WeCom JSON endpoints read

seacen:
## What

`RegisterWecomBYO` and `RedeemWecomBindingToken` both decode `r.Body` unbounded. `encoding/json` materializes a string value whole before `Decode` returns, so one authenticated caller POSTing a multi-gigabyte token string costs the API server that much RSS per request. No auth bypass involved — an ordinary member account reaches both endpoints.

Both bodies are now wrapped in `http.MaxBytesReader` at 16 KiB. Neither carries more than a handful of short fields: the redeem token is 32 bytes, and a bot id plus a long-connection secret is not close to the ceiling either.

## Tests

Two, and they need no database of their own — `requestUserID` reads only the `X-User-ID` header, so the redeem handler can be driven directly with a fake redeemer.

- an oversized body is refused with 400 **and never reaches the service**
- an ordinary body still gets through — this is what catches a ceiling set too tight

Deleting the two `MaxBytesReader` lines fails the first one:

```
wecom_web_test.go:52: status = 200, want 400
wecom_web_test.go:55: an oversized body reached the redeem service 1 time(s), want 0
```

## Note on the interface

`Handler.WecomBindingTokens` is retyped from `*wecom.BindingTokenService` to a small `WecomBindingRedeemer` interface so the test can supply that fake. Production wiring is unchanged — `*wecom.BindingTokenService` satisfies it, and `router.go` needed no edit.

## Related

`#6581` adds `wecom_install_web.go` with its own JSON decodes and no ceiling. If both land, `wecomBodyLimit` should be extended to those handlers too — happy to do that as a follow-up once it merges.

### DIFF
--- a/server/internal/handler/handler.go
+++ b/server/internal/handler/handler.go
@@ -279,7 +279,7 @@ type Handler struct {
 	// to their real userid or email, so an explicit binding is required —
 	// see wecom/binding.go). Nil disables the redeem endpoint (returns 503)
 	// and the OutboundReplier's binding-prompt path.
-	WecomBindingTokens *wecom.BindingTokenService
+	WecomBindingTokens WecomBindingRedeemer
 	// LLM is the basic LLM API layer (MUL-4238): a thin wrapper over the
 	// OpenAI Go SDK backing server-internal one-shot LLM helpers such as chat
 	// title generation. The generic passthrough endpoints were removed in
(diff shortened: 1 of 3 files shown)


multica-eve (approved):
Thanks @seacen for identifying this resource-exhaustion risk and for the focused fix. The request-size limit is applied at the right boundary, normal requests remain covered, and the full CI suite is green. Approved — thank you for the contribution!

PR #6588, opened 2026-08-07
fix(wecom): don't let a closing connection evict its own successor

seacen:
## What

`sendersRegistry.clear` deletes the installation's entry unconditionally. `Connect` installs the sender on entry and clears it on a `defer`, so when a lease flips while the old socket is still draining, the two generations overlap and the **loser's defer runs after the winner's `set`**.

The registry then holds nothing while a perfectly healthy connection is up. `OutboundReplier` and `wecomChannel.Send` both resolve through `get`, so every outbound push returns nil and the bot goes silent — with nothing in the log to explain it, because the losing generation did exactly what it was told. It stays that way until the next reconnect happens to re-register.

`clear` now takes the sender doing the clearing and only deletes if that sender still owns the slot.

## Precedent

This is the one platform that didn't fence it:

- **DingTalk** — `dingtalk_channel.go:74` guards the same handover with `defer c.slot.current.CompareAndSwap(c, nil)`
- **Slack / Lark** — no registry at all; `Disconnect` is a no-op because their outbound is REST
- **WeCom** — `delete(r.byKey, ...)`, unconditional

## Tests

Three, no database needed.

- a predecessor's late `clear` leaves the successor in place — **fails on the old code**: `the successor was evicted by its predecessor's deferred clear — every outbound push now resolves to nil and the bot is silent`
- a sender that *does* own the slot is still removed. The fence can't be "never delete", or outbound gets dispatched to a dead socket, and this test is what pins that
- clearing an unknown installation disturbs nothing

## Why it gets worse with #6581

Under that PR `queueSender.Send` resolves `s.senders.get(row.InstallationID)` per row and returns `DispositionRetry` on nil. So a stale clear stops being "replies vanish" and becomes "every queued row for this installation retries until the backoff gives up" — the same silence, now with the queue's retry budget burning behind it.

### DIFF
--- a/server/internal/integrations/wecom/senders_registry.go
+++ b/server/internal/integrations/wecom/senders_registry.go
@@ -45,10 +45,26 @@ func (r *sendersRegistry) set(id pgtype.UUID, s *wsSender) {
 	r.byKey[util.UUIDToString(id)] = s
 }
 
-func (r *sendersRegistry) clear(id pgtype.UUID) {
+// clear removes this installation's entry, but only if s is still the sender
+// registered under it. A generation that is shutting down must not evict its
+// own successor: Connect installs on entry and clears on a defer, so when a
+// lease flips while the old socket is still draining, the two overlap and the
+// loser's defer runs after the winner's set. Deleting unconditionally there
+// leaves the registry empty while a healthy connection is up, and every
+// outbound push resolves to nil — the bot goes silent with nothing in the log
+// to say why, until the next reconnect happens to re-register.
+//
+// dingtalk_channel.go:74 guards the same handover with
+// `CompareAndSwap(c, nil)`; slack and lark have no registry at all because
+// their outbound is REST. WeCom was the one platform deleting unconditionally.
+func (r *sendersRegistry) clear(id pgtype.UUID, s *wsSender) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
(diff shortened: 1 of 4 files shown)


Bohan-J:
Thanks @seacen — great catch, and the write-up made this easy to review.

The fence is the right shape. With `set` and `clear` under the same mutex, both handover orderings are atomic: a predecessor's late `defer` either finds the successor already installed and returns, or clears its own entry before the successor registers. That lines up with the `CompareAndSwap(c, nil)` guard DingTalk already had, and WeCom really was the one connection-backed channel deleting unconditionally.

The middle test is the one I appreciated most — asserting that a sender which *does* own the slot is still removed keeps the fence from quietly degrading into "never delete", which would just trade silence for dispatching to a dead socket.

I pushed one comment-only commit on top (43d6e5d): the registry's file header claimed `wecomChannel.Send` resolves the live `wsSender` through the map, but `Send` returns `ErrSendNotSupported` and never touches it — the real readers are `OutboundReplier` and `Outbound`. The same claim had spread to `ChannelDeps.Senders` and `wsSender.sendText`, so all three now name the actual call path. No behavior change.

Verified locally on the final tree: `go test -race ./internal/integrations/wecom -count=1`, `go vet`, and `gofmt` all clean.

Merging now — thanks for fencing this before #6581 turns the silence into a retry storm.

PR #6589, opened 2026-08-07
fix(wecom): stop blaming an admin's credentials when the fault is ours

seacen:
## The bug

`RegisterWecomBYO`'s `default` branch answers 400 with *"could not connect the WeCom bot — check the Bot ID and secret from the WeCom admin console"*.

Every failure that is not one of the three ownership conflicts lands there: a Postgres blip, a bad `MULTICA_WECOM_SECRET_KEY`, a bug of ours. All of them tell the admin their credentials are wrong.

That is not cosmetic. It sends them to the WeCom console to rotate a secret that was fine — and **a rotated long-connection secret cannot be recovered**, the console shows it once. A thirty-second database failover costs them a working bot, and the log line was `slog.Warn`.

## The split

By whose fault it is:

| | before | after |
|---|---|---|
| missing field | 400, blames credentials | 400 `wecom_install_rejected` |
| bot owned elsewhere (×3) | 409, right sentence, no code | 409 + code |
| **anything else** | **400, blames credentials** | **500 `wecom_install_failed`, says the credentials were not changed**, logged at Error with workspace + agent |

`validateInstallationParams` now wraps a new `ErrInvalidInstallationParams` sentinel, which is what makes "the caller left something out" distinguishable from "we broke".

## Localization

Each branch carries a stable `code` next to the English sentence, via a new `writeErrorCode` helper. `ApiError` already carries the decoded body — its comment even says *"Carries structured error fields like `code`"* — so the frontend side is a small `errorCode` accessor next to the existing `dispatchReasonCode`, and a switch in the settings tab.

A zh-Hans admin currently gets English toasted at them regardless of console language. Five new keys in en / ja / ko / zh-Hans; key sets verified identical across all four. The English sentence stays as the fallback for a server that sends no code.

## Tests

Server (no database needed):
- every missing field is identifiable as a caller error **and names itself** — if one stops matching, that request silently becomes a 500
- complete params pass
- an infrastructure error does **not** match the sentinel — this is the whole split, stated as an assertion

Frontend (11 passing in `wecom-tab.test.tsx`, 2 new):
- a coded failure renders the localized sentence, not the server's English. Removing the switch fails it
- a failure with no code still falls back to the server's message, so the fallback can't quietly regress into the generic toast

## Not included

`ErrCredentialsRejected` / `ErrCredentialsUnverifiable` — distinguishing "WeCom refused this pair" from "we couldn't reach WeCom" needs a credential probe at install time, which is a separate change. This PR only classifies what the current code already knows.


Bohan-J (changes requested):
Thanks for this — the product direction is right, and I want it in. An admin whose install failed because of our database or our encryption key should not be sent to the WeCom console to rotate a secret that was fine, and the localized copy is a real improvement over toasting English at them.

Requesting changes on one thing: this needs a rebase, and the rebase is not mechanical.

## Rebase onto `main` and keep the full error matrix

The branch now conflicts with `main` (GitHub reports `CONFLICTING` / `DIRTY`) inside the install error `switch` in `server/internal/handler/wecom_web.go`. #6594 landed on `main` in the meantime and added two genuinely new outcomes to that same `switch`:

- `wecom.ErrCredentialsRejected` — WeCom itself refused the Bot ID / secret pair. 400, and the one branch entitled to tell the admin to go check their credentials.
- `wecom.ErrCredentialsUnverifiable` — we could not reach WeCom to check at all. 503, with copy that explicitly says the credentials were *not* changed.

This branch's `switch` (`server/internal/handler/wecom_web.go:167`–`191`) predates both: it recognizes the three ownership conflicts and `ErrInvalidInstallationParams`, and everything else falls to the `default` → 500 `wecom_install_failed`.

So neither side of the conflict can just be taken wholesale:

- Keeping this branch's side collapses `ErrCredentialsRejected` and `ErrCredentialsUnverifiable` into "something went wrong on our side," losing the 400/503 semantics and the recoverable next action #6594 just established.
- Keeping `main`'s side loses the stable codes, the localization, and the actual fix here — `main`'s `default` still answers 400 and blames the credentials.

After the rebase, please keep the whole matrix explicit:

| outcome | status | notes |
| --- | --- | --- |
| bot already owned (same workspace / archived agent / other workspace) | 409 | as today |
| required field missing | 400 | admin's input, admin's fix |
| WeCom rejected the credentials | 400 | the one branch that may blame the credentials |
| WeCom unreachable, credentials unverified | 503 | say the credentials were not changed |
| anything else (DB, encryption key, bug) | 500 | log it; do not blame the admin |

The last two need their own stable codes and their own strings in all four locales. Please don't fold them into `wecom_install_rejected` — that code currently means "you left a field out," and the frontend contract shouldn't quietly widen to cover "WeCom said no" and "we couldn't ask WeCom," which call for different admin actions.

## Please add handler-level contract tests

`installation_params_test.go` proves the sentinel matching, and the frontend test mocks `errorCode` — neither would catch a conflict resolution that wires the response matrix up wrong, which is exactly the risk here. A table-driven test at the handler level asserting HTTP status + `error` + `code` for each error class would pin the contract down.

## What I checked

- `go test ./internal/integrations/wecom -count=1` passes on the current head.
- `git diff --check origin/main...HEAD` is clean.
- The WeCom key sets match across all four locales (43 keys each).
- CI is green on `1d5f386`, but that ran against the old base — it'll need to re-run after the rebase.

Happy to re-review as soon as it's rebased.

seacen:
Rebased onto `main`, and the full matrix from your table is in — including the two outcomes #6594 added, each with its own code and its own strings in all four locales. Handler-level table test added, which is what caught the one mistake worth catching here. Both of your points are addressed; nothing declined.

One thing worth saying up front, because it changes what the third and fourth rows of your table need: **#6594 already establishes the rejected/unverifiable split at the service layer, and does it more narrowly than this PR would have.** Detection was already right. What was wrong is that both outcomes reached the HTTP layer and left it carrying another outcome's code, and that is the part I fixed.

## Rebase onto `main` and keep the full error matrix — done

The conflict was one hunk, in the `switch` in `server/internal/handler/wecom_web.go`, exactly where you said. Both sides survived, hunk by hunk:

| outcome | status | code |
| --- | --- | --- |
| bot owned, same workspace | 409 | `wecom_bot_owned_by_same_workspace` |
| bot owned, archived agent | 409 | `wecom_bot_owned_by_archived_agent` |
| bot owned, other workspace | 409 | `wecom_bot_owned_by_another_workspace` |
| required field missing | 400 | `wecom_install_rejected` |
| WeCom rejected the credentials | 400 | `wecom_credentials_rejected` |
| WeCom unreachable, unverified | 503 | `wecom_credentials_unverifiable` |
| anything else | 500 | `wecom_install_failed` |

`wecom_credentials_rejected` and `wecom_credentials_unverifiable` are new and separate, per your instruction not to fold them into `wecom_install_rejected`. I agree with the reasoning and want to state it in the form the code now enforces: `wecom_install_rejected` means "you left a field out", and the three 4xx outcomes are three different next actions for the admin — fill in the field, go check the pair on the console, wait and retry. A client keying off one code cannot tell them apart. Two of the three are even the same HTTP status, so the code is the *only* thing carrying the distinction. There is now a test that fails if any two outcomes ever share a code (below).

Seven keys each in `en` / `ja` / `ko` / `zh-Hans`, verified identical key sets across all four — 45 keys each now, up from the 43 you counted. The frontend `switch` in `wecom-tab.tsx` gained the two matching arms.

While resolving I also lifted the whole `switch` out of `RegisterWecomBYO` into a named function, `writeWecomInstallError`. That was needed to test it at all: the handler has no seam to inject an `Upsert` error through, since `wecomInstallService()` returns a concrete `*wecom.InstallationService`. With the matrix inline there was no way to exercise it without a live database and a fake WeCom. Extracted, it is directly testable, and it reads as a matrix instead of something you reconstruct from a switch two screens down.

### What #6594 already covers

Your table's "WeCom rejected the credentials" / "WeCom unreachable" rows needed **no new detection mechanism** — #6594 landed all of it, and I did not add a second one:

- The errcode whitelist is already as tight as it should be. `40001` and `40013` only map to `ErrCredentialsRejected`; every other non-zero code fails closed to `ErrCredentialsUnverifiable`. Throttling codes (`45009` / `45033`) and anything WeCom adds later land on the 503, not on "your secret is wrong".
- The probe is already non-optional. `NewInstallationService` defaults to the real handshake, returns `ErrProbeRequired` on an explicit nil, and `Upsert` refuses to run without one.

So the two sentinels arrive at the handler correctly classified. Only the last hop was still broken: on the pre-rebase branch both fell through to `default` and came back as 500 `wecom_install_failed`, the message this PR exists to reserve for our own faults. Telling an admin "something went wrong on our side, your credentials were not changed" when WeCom actually rejected their secret is the same mistake this PR is fixing, pointed the other way. Giving each its own arm fixes it.

## Handler-level contract tests — added

`server/internal/handler/wecom_install_error_matrix_test.go`, two tests.

`TestWecomInstallErrorMatrix` is table-driven over all seven outcomes and asserts HTTP status, the `error` sentence, and the `code` for each. It also covers a wrapped sentinel (`fmt.Errorf("upsert: %w", …)`) to pin that classification is `errors.Is` and not string matching, and two distinct "ours" failures (database down, missing encryption key) to pin that both land on 500 rather than the admin's lap.

`TestWecomInstallCodesAreDistinct` asserts no two outcomes return the same code. This is the one that earns its place: the mistake you were worried about is not always visible in the status. If `ErrCredentialsRejected` gets folded into `wecom_install_rejected`, both are still 400 and every status assertion still passes — only the code changed, and only this test notices. Verified below.

You were right that neither existing layer would have caught a bad resolution: `installation_params_test.go` proves the sentinel matching and stops there, and the frontend test mocks the `errorCode` it expects to receive. The wiring between them was untested, and that is where a merge goes wrong.

I also added two frontend cases in `wecom-tab.test.tsx` covering the new codes, one of which asserts the credentials-rejected sentence is *not* the missing-field sentence. Those keep the new locale keys honest — a typo'd key renders as the key name.

## Verification

Rebase: one conflicted file, `server/internal/handler/wecom_web.go`, one hunk. `server/internal/integrations/wecom/installation.go` auto-merged (this branch's `validateInstallationParams` rewrite and #6594's transaction/probe sequence touch different parts of the file). No sqlc-generated file needed regenerating — the branch does not change any query. I checked that `main`-only content in the area survived: the maintainer docblock in `wecom/types.go` and `TestConnectReturnsOnCtxCancel` in `wecom_channel_test.go` are byte-identical to `main`, and the branch touches neither file. 13 files changed, +506 / −50.

Green on the rebased tree: `go build ./...`, `go vet ./...`, `gofmt -l` clean on every file touched, `go test -race ./internal/integrations/wecom/ ./internal/handler/ -count=1`, plus `pnpm typecheck` and `vitest run` for `@multica/views` (13 passed, 0 lint findings in the touched files).

### Reverse-verification

Each new assertion was falsified before being trusted, using the two failure modes you named.

**1. The resolution keeps this branch's side wholesale — both credential arms lost.**

```
--- FAIL: TestWecomInstallErrorMatrix (0.00s)
    --- FAIL: TestWecomInstallErrorMatrix/WeCom_refused_the_credentials (0.00s)
        wecom_install_error_matrix_test.go:117: status = 500, want 400 (go check the pair on the WeCom console — the only outcome that may say this)
        wecom_install_error_matrix_test.go:128: code = "wecom_install_failed", want "wecom_credentials_rejected" (go check the pair on the WeCom console — the only outcome that may say this)
    --- FAIL: TestWecomInstallErrorMatrix/WeCom_unreachable,_credentials_unverified (0.00s)
        wecom_install_error_matrix_test.go:117: status = 500, want 503 (wait and retry; the credentials were not changed)
        wecom_install_error_matrix_test.go:128: code = "wecom_install_failed", want "wecom_credentials_unverifiable" (wait and retry; the credentials were not changed)
    --- FAIL: TestWecomInstallErrorMatrix/a_wrapped_sentinel_is_still_classified_by_its_cause (0.00s)
        wecom_install_error_matrix_test.go:117: status = 500, want 503 (errors.Is, not string matching)
        wecom_install_error_matrix_test.go:128: code = "wecom_install_failed", want "wecom_credentials_unverifiable" (errors.Is, not string matching)
--- FAIL: TestWecomInstallCodesAreDistinct (0.00s)
    wecom_install_error_matrix_test.go:176: "wecom_install_failed" is returned for both "creds rejected" and "creds unverified"; the frontend cannot tell them apart, so one of the two gets the wrong next action
    wecom_install_error_matrix_test.go:176: "wecom_install_failed" is returned for both "creds rejected" and "ours"; the frontend cannot tell them apart, so one of the two gets the wrong next action
FAIL
FAIL	github.com/multica-ai/multica/server/internal/handler	0.761s
```

**2. `ErrCredentialsRejected` folded into `wecom_install_rejected`** — the exact widening you asked me not to do. Note the status assertion passes; both are 400. Only the code assertion and the distinctness test catch it:

```
--- FAIL: TestWecomInstallErrorMatrix (0.00s)
    --- FAIL: TestWecomInstallErrorMatrix/WeCom_refused_the_credentials (0.00s)
        wecom_install_error_matrix_test.go:128: code = "wecom_install_rejected", want "wecom_credentials_rejected" (go check the pair on the WeCom console — the only outcome that may say this)
--- FAIL: TestWecomInstallCodesAreDistinct (0.00s)
    wecom_install_error_matrix_test.go:176: "wecom_install_rejected" is returned for both "invalid params" and "creds rejected"; the frontend cannot tell them apart, so one of the two gets the wrong next action
FAIL
FAIL	github.com/multica-ai/multica/server/internal/handler	0.774s
```

**3. Frontend: the two new `switch` arms removed**, codes arriving with nothing mapping them:

```
     × tells the admin their credentials were not changed when WeCom was unreachable 1094ms
     × renders the credentials-rejected sentence, distinct from the missing-field one 1109ms
 FAIL  settings/components/wecom-tab.test.tsx > WecomAgentBindButton > tells the admin their credentials were not changed when WeCom was unreachable
AssertionError: expected "vi.fn()" to be called with arguments: [ Array(1) ]
 FAIL  settings/components/wecom-tab.test.tsx > WecomAgentBindButton > renders the credentials-rejected sentence, distinct from the missing-field one
AssertionError: expected "vi.fn()" to be called with arguments: [ Array(1) ]

 Test Files  1 failed (1)
      Tests  2 failed | 11 passed (13)
```

All three restored, everything green again.

### CI

Green on `7a21fb8` — all 10 checks pass, and GitHub now reports the PR `MERGEABLE` / `CLEAN`.

The first run after the rebase did show a `backend` failure, and it was not this change: `TestReadOpenclawStdoutWaitsForCompleteResult` in `server/pkg/agent`, a package this branch does not touch (0 of 13 changed files).

```
--- FAIL: TestReadOpenclawStdoutWaitsForCompleteResult (0.70s)
    openclaw_stdout_test.go:363: cutShort = true after a clean EOF, want false
```

That is a timing race under load, on a test that has already been made deterministic twice (#6320, #6501) and whose own comment describes this exact failure mode: the idle ticker winning "after the bytes became parseable but before the read goroutine observed EOF on a loaded CI runner". Locally it passes 20/20 in isolation and the whole package passes unloaded, but running `./pkg/agent` under load reproduces the same shape with a different test each time — `TestCodexExecuteTimeoutWinsOverProcessExitDuringActiveTurn`, `TestKiroBackendTreatsGoalCompleteCloseErrorAsCompleted`, all "context deadline exceeded". I could not re-run the job (no admin on the repo), so I re-pushed the identical tree — same content, verified same tree hash, new SHA — and it came back green. Flagging it because that package looks due another look, independent of this PR.

`main` moved one commit further while I was working (#6590, outbound send verdict). I did not re-rebase onto it: it shares no file with this branch and merges cleanly, and GitHub reports the PR mergeable.

Bohan-J (approved):
Both of the previous round's must-fixes are resolved, and I re-verified the whole matrix from scratch rather than taking the rebase on trust.

**Rebase and matrix.** `MERGEABLE / CLEAN`, and `writeWecomInstallError` now carries all seven outcomes: three ownership conflicts at 409, missing field at 400, `wecom_credentials_rejected` at 400, `wecom_credentials_unverifiable` at 503, everything else at 500 with an Error-level log. The two credential outcomes got their own codes and their own strings rather than widening `wecom_install_rejected`, which is what I asked for — the three call for three different admin actions.

**Verified locally**

- `go test -race ./internal/integrations/wecom` and the full `./internal/handler` package: pass. Note the handler package's `TestMain` exits 0 when no database is reachable, so a bare `go test ./internal/handler` on a dev box prints `ok` without running anything. I created a scratch database and ran the migrations so the matrix test genuinely executed; I also confirmed from the CI log that the backend job ran the package for real (34s, no skip line) rather than skipping past it.
- `packages/core`: 1345 tests pass. `packages/views`: 3737 pass.
- `go vet`, `gofmt` on the touched files, locale key parity across en/ja/ko/zh-Hans (45 each, identical sets).

I pushed the two nits rather than send you round again — `6dd0a4a` and `d440665`, both green:

1. The matrix test's header claimed it pinned status + error + code, but the table had no expected sentence and only asserted one was non-empty. That leaves open the exact case the file was written to catch: right status, right code, neighbour's sentence — which sends a client too old to know the codes to do the wrong thing. The table now carries each sentence and compares it exactly. I checked it bites by crossing the rejected/unverifiable strings in `writeWecomInstallError`: three rows fail, with status and code still correct.
2. `ErrInvalidInstallationParams` had been declared between `validateInstallationParams`'s doc comment and the function, so godoc was attaching "is a lightweight pre-write check" to the error value and the function was left undocumented. Split back apart, and the classification text — which still said everything not one of these belongs on a 500 — now names the 400/503 credential outcomes too.

Nice piece of work: the 500 that used to blame the credentials was the kind of bug that costs someone an unrecoverable secret, and separating "you left a field out" from "WeCom said no" from "we couldn't ask WeCom" is the distinction that makes the fix actually usable. Merging.

PR #6590, opened 2026-08-07
feat(wecom): read the server's verdict on an outbound send (MUL-5887)

seacen:
The verdict plumbing from #6581's review, as a standalone PR. @wzxjohn — build on this and `queueSender.Send` can classify on the errcode instead of guessing at strings.

## What

`sendText` marshalled a frame, wrote it, returned nil. WeCom's answer arrives in a **separate ack frame**, so a frame the server refused was indistinguishable from one it accepted:

- a body over the 20480-byte cap
- the bot removed from a chat whose binding row survives (40001)
- 45009

All three returned `nil`. The operator's only signal was `wecom: server ack error` in the log, with no way to tell which send it belonged to.

`sendText` now waits for the ack and returns a typed `*wecomAPIError` carrying the errcode. A verdict that never arrives is `errAckTimeout`, deliberately distinct — the message may well have been delivered, so a refusal and a lost ack call for **opposite** responses.

## Why the callback worker is in the same PR

It cannot be split, and this is the part worth reading closely.

**The read loop is the sole deliverer of acks.** Before this, `dispatchFrame` ran inbound callbacks inline on that loop. A callback that writes a frame and waits for its ack would therefore be waiting on the very loop that has to deliver it — it would wait on itself until the timeout, every time.

So callbacks move to one worker behind a `callbackQueueDepth = 64` queue:

- **One worker, not a pool.** WeCom delivers a chat's messages in order, and the engine's dedup and turn batching assume that order survives.
- **A full queue blocks the read loop rather than dropping.** Backpressure costs a reconnect; dropping costs a user's message with nothing to say so. A replica that can't keep up should hand the bot to one that can.
- Acks, pings and pongs stay on the read loop — they are exactly what the worker's writes are waiting for.

DingTalk's adapter in `main` is already shaped this way.

## For #6581

`queueSender.Send` currently does:

```go
if err := sender.sendText(...); err != nil {
    if retryableSendError(err) { return outbox.DispositionRetry, err }
    return outbox.DispositionFailed, err
}
return outbox.DispositionSent, nil
```

On top of this, `retryableSendError` can switch on `*wecomAPIError.Code` instead of `strings.Contains` over "timeout" / "broken pipe" — and more importantly, `DispositionSent` starts meaning the message arrived rather than that the bytes left the process.

I deliberately left **`splitForWire`** (the 20480 cap) out of this PR. It wants to sit on your send path once the queue is settled, and doing it here would have us both editing the same lines. Happy to send it once #6581 lands.

## Tests

Three new, plus the whole existing suite.

- a refusal surfaces as `*wecomAPIError` with the errcode intact
- an accepted send still succeeds and writes exactly one frame
- **a lost ack is not reported as a refusal** — the distinction the classifier depends on

`recordingConn` gained an `autoAck` mode so the test doubles answer their writes the way the server does. Without it every send in the suite waited out the 5s timeout; the package went from 10.7s to 0.7s as a side effect.

### DIFF
--- a/server/internal/integrations/wecom/dispatch_test.go
+++ b/server/internal/integrations/wecom/dispatch_test.go
@@ -40,7 +40,7 @@ func TestDispatchFrame_TextReachesHandler(t *testing.T) {
 		return nil
 	})
 	conn := &recordingConn{}
-	err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, "text", "hello"), newWSSender(conn, nil), slog.Default())
+	err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, "text", "hello"), conn.autoAck(newWSSender(conn, nil)), slog.Default())
 	if err != nil {
 		t.Fatalf("dispatchFrame: %v", err)
 	}
@@ -63,7 +63,7 @@ func TestDispatchFrame_NonTextGetsReceiptAndSkipsHandler(t *testing.T) {
 		return nil
 	})
 	conn := &recordingConn{}
-	err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, "image", ""), newWSSender(conn, nil), slog.Default())
+	err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, "image", ""), conn.autoAck(newWSSender(conn, nil)), slog.Default())
 	if err != nil {
 		t.Fatalf("dispatchFrame: %v", err)
 	}
@@ -95,7 +95,7 @@ func TestDispatchFrame_ServerPingIsPonged(t *testing.T) {
 	t.Parallel()
 	c := testChannel(func(context.Context, channel.InboundMessage) error { return nil })
 	conn := &recordingConn{}
-	err := c.dispatchFrame(context.Background(), frameEnvelope{Cmd: cmdServerPing, Headers: frameHeaders{ReqID: "r1"}}, newWSSender(conn, nil), slog.Default())
+	err := c.dispatchFrame(context.Background(), frameEnvelope{Cmd: cmdServerPing, Headers: frameHeaders{ReqID: "r1"}}, conn.autoAck(newWSSender(conn, nil)), slog.Default())
 	if err != nil {
 		t.Fatalf("dispatchFrame: %v", err)
 	}
--- a/server/internal/integrations/wecom/outbound.go
+++ b/server/internal/integrations/wecom/outbound.go
@@ -136,7 +136,7 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 		return errors.New("wecom: connection not ready on this replica")
 	}
 	chatType := aibotChatTypeFromChannel(channel.ChatType(binding.ChatType))
-	return sender.sendText(binding.ChannelChatID, chatType, content)
+	return sender.sendTextCtx(ctx, binding.ChannelChatID, chatType, content)
 }
 
 // chatDoneContent extracts the reply text from an EventChatDone payload
@@ -228,7 +228,7 @@ func (o *Outbound) tryDeliverInbox(ctx context.Context, item map[string]any, rec
 	// Smart-bot inbox notifications are 1:1 pushes to the bound user. The
 	// binding row's channel_user_id is the bot-scoped T-* userid — WeCom
 	// treats that as the chatid for a single (chat_type=1) send.
-	if err := sender.sendText(binding.ChannelUserID, chatTypeSingleInt, content); err != nil {
+	if err := sender.sendTextCtx(ctx, binding.ChannelUserID, chatTypeSingleInt, content); err != nil {
 		o.logger.WarnContext(ctx, "wecom outbound: inbox push failed",
 			"error", err, "installation_id", uuidStringPub(binding.InstallationID),
 			"recipient_id", recipientIDStr)
--- a/server/internal/integrations/wecom/outbound_test.go
+++ b/server/internal/integrations/wecom/outbound_test.go
@@ -57,7 +57,7 @@ func newOutboundWithConn(t *testing.T, q outboundQueries) (*Outbound, pgtype.UUI
 	reg := newSendersRegistry()
 	instID := mustTestUUID(t)
 	conn := &recordingConn{}
-	reg.set(instID, newWSSender(conn, nil))
+	reg.set(instID, conn.autoAck(newWSSender(conn, nil)))
 	return NewOutbound(q, reg, slog.Default()), instID, conn
 }
 
--- a/server/internal/integrations/wecom/replier.go
+++ b/server/internal/integrations/wecom/replier.go
@@ -170,7 +170,7 @@ func (r *OutboundReplier) sendBindingPrompt(ctx context.Context, inst engine.Res
 	// the link privately to the sender's own userid with chat_type=1 (the same
 	// address outbound.go uses for inbox pushes), never to the room. Lark's
 	// SendBindingPromptCard targets the sender's OpenID for the same reason.
-	if err := r.postPrivate(inst, sender, text); err != nil {
+	if err := r.postPrivate(ctx, inst, sender, text); err != nil {
 		return err
 	}
 	// A group trigger still needs an answer — silence reads as a broken bot —
@@ -186,7 +186,7 @@ func (r *OutboundReplier) sendBindingPrompt(ctx context.Context, inst engine.Res
 // postPrivate delivers text to a single user's 1:1 chat (chat_type=1),
 // regardless of which room triggered the message. Used for bearer-credential
 // content (the binding link) that must never land in a group.
-func (r *OutboundReplier) postPrivate(inst engine.ResolvedInstallation, userID, text string) error {
+func (r *OutboundReplier) postPrivate(ctx context.Context, inst engine.ResolvedInstallation, userID, text string) error {
 	if r.senders == nil {
 		return errors.New("wecom: sender registry not configured")
 	}
@@ -200,7 +200,7 @@ func (r *OutboundReplier) postPrivate(inst engine.ResolvedInstallation, userID, 
 	if sender == nil {
 		return errors.New("wecom: connection not ready")
 	}
-	return sender.sendText(userID, chatTypeSingleInt, text)
+	return sender.sendTextCtx(ctx, userID, chatTypeSingleInt, text)
 }
 
 // post looks up the installation's live wsSender in the registry and
@@ -208,7 +208,6 @@ func (r *OutboundReplier) postPrivate(inst engine.ResolvedInstallation, userID, 
 // ready" when the Supervisor has no active connection (mid-reconnect
 // after lease flip, or right after Revoke).
 func (r *OutboundReplier) post(ctx context.Context, inst engine.ResolvedInstallation, msg channel.InboundMessage, text string) error {
-	_ = ctx
 	if r.senders == nil {
 		return errors.New("wecom: sender registry not configured")
 	}
@@ -224,7 +223,7 @@ func (r *OutboundReplier) post(ctx context.Context, inst engine.ResolvedInstalla
 		return errors.New("wecom: missing chat_id")
 	}
 	chatType := aibotChatTypeFromChannel(msg.Source.ChatType)
-	return sender.sendText(chatID, chatType, text)
+	return sender.sendTextCtx(ctx, chatID, chatType, text)
 }
 
 func issueCreatedText(res engine.Result) string {
--- a/server/internal/integrations/wecom/replier_test.go
+++ b/server/internal/integrations/wecom/replier_test.go
@@ -44,6 +44,23 @@ func (f fakeBinder) Mint(context.Context, pgtype.UUID, pgtype.UUID, string) (Bin
 type recordingConn struct {
 	mu     sync.Mutex
 	frames []frameEnvelope
+
+	// sender, when set, makes this double answer its writes the way the real
+	// server does: an ack frame echoing the req_id with errcode 0. Senders
+	// that read their verdict block until one arrives, so a double that never
+	// answers turns every send into a 5-second timeout.
+	//
+	// Set refuseCode to make the server refuse instead.
+	sender     *wsSender
+	refuseCode int
+	refuseMsg  string
+}
+
+// autoAck wires the double to answer the sender's writes. Call it after
+// newWSSender, which needs the conn first.
+func (c *recordingConn) autoAck(s *wsSender) *wsSender {
+	c.sender = s
+	return s
 }
 
 func (c *recordingConn) WriteMessage(_ int, data []byte) error {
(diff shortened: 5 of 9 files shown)


seacen:
Folded one more read-loop fix into this PR, since it touches the same lines and would have conflicted otherwise.

**The 90s read deadline was armed *after* `ReadMessage` returned**, which put decode and dispatch inside the window that is supposed to measure idleness. Only the server's pong resets it on a quiet bot and our ping goes out every 30s, so on a loaded replica the next read could time out on a socket with nothing wrong with it — and the Supervisor reconnects a healthy connection.

The numbers are unchanged (`pingInterval` 30s, `readDeadline` 90s); only the moment of arming moves, which is why a constants diff cannot see it.

The error is also no longer discarded — a socket that refuses a deadline is not one to keep reading from. Shutdown closes the socket, which also refuses it, so that path checks ctx and returns cleanly instead of reporting a failure upward.

No dedicated test for this one, deliberately. I wrote one, then noticed it drove its own arm/read sequence rather than the real loop and kept passing with the fix reverted — so it was measuring nothing. The change is one line moving to the top of the loop plus an error check; the existing connection tests cover the loop still running. If you'd rather have it pinned, the honest way is a fake conn driven through `Connect` with the subscribe handshake stubbed, and I'm happy to add that.

Bohan-J (changes requested):
The product change is right and the argument for why the callback worker can't be split out is convincing — a send that waits on the loop that delivers its own ack is exactly the self-wait you describe. One worker for order, backpressure over dropping: both trade-offs hold up.

Two things before this can go in. The first is a wedge I reproduced.

Worth calling out first, because it's the part that's easy to get wrong and you didn't: `awaitReply` registers **before** `write` (`ws_sender.go:188-198`), so an ack that beats the caller to the select still lands in the buffered channel. And a late ack — one that arrives after `errAckTimeout` already fired — falls through `routeResponse` into the pre-existing `server ack error` log (`wecom_channel.go:408-417`) instead of vanishing. Both are the right calls.

---

## Must-fix 1 — the read loop wedges permanently when the callback worker dies on a full queue

The worker returns on its first dispatch error without draining what's queued behind it (`wecom_channel.go:208-219`), and the read loop's select only watches the send and `ctx.Done()` (`wecom_channel.go:273-278`):

```go
select {
case callbacks <- env:
case <-ctx.Done():
    return nil
}
```

When the queue is full — one callback in the worker's hands, 64 buffered, the 66th frame parked on the send — and the worker then errors out, nothing will ever drain that channel again. The `conn.Close()` in the worker doesn't rescue it: the read loop isn't in `ReadMessage` at that moment, it's blocked on a channel send.

The consequence is worse than `Connect` hanging. `Supervisor.supervise` blocks on `ch.Connect(runCtx)` (`internal/integrations/channel/engine/supervisor.go:510`), so `runCancel()` never runs, so `renewLeaseUntil` keeps renewing the WS lease indefinitely. The bot is dead *and* no other replica can take the lease from the wedged one. Only a process restart recovers it.

I drove this with a throwaway test — a `wsConn` that acks subscribe then streams `aibot_msg_callback` frames endlessly, plus a handler whose first call parks until released and then returns an error:

<details>
<summary>repro sketch</summary>

```go
// wait until the read loop has pumped callbackQueueDepth+2 frames,
// i.e. 1 in the worker + 64 buffered + 1 parked on the send
close(release) // worker fails, calls conn.Close(), returns without draining

select {
case err := <-errCh:
    t.Logf("Connect returned: %v", err)
case <-time.After(5 * time.Second):
    t.Fatal("WEDGED: Connect never returned")
}
```

</details>

On this branch it hits the 5s fatal. Adding one arm to the select fixes it:

```go
select {
case callbacks <- env:
case <-cbDone:
    // The worker is gone; nothing will drain the queue again.
    // The deferred handler surfaces cbErr as the real cause.
    return nil
case <-ctx.Done():
    return nil
}
```

With that, the same test has `Connect` returning the genuine `handler infra failure` in 0.11s, and `go test ./internal/integrations/wecom/ -race` stays green.

## Must-fix 2 — the ack wait lands on synchronous request-path goroutines

`events.Bus` is in-process **synchronous** (`internal/events/bus.go:61`), so this is one unbroken call chain on the HTTP handler's goroutine:

```
HTTP handler → Handler.publish → Bus.Publish
  → notification_listeners.go, per subscriber → bus.Publish(EventInboxNew)
    → wecom.Outbound.handleInboxNew → sendText
```

Before this PR `sendText` returned as soon as the frame hit the socket buffer — effectively free. Now it costs a WeCom round trip, and a full `ackTimeout = 5s` when the ack is lost. Per subscriber, serially, inside the loop in `notifySubscribers`. A workspace with three WeCom-bound subscribers turns one issue-creation request into a worst case of ~15s.

What makes this more than a latency note: `outbound.go:81-82` builds a 10s ctx with the comment *"Bus delivery is synchronous — a stuck WS write must not wedge the publish call site"*, and `handleInboxNew` builds a 5s one. But `sendText` hardcodes `context.Background()` (`ws_sender.go:238`), so neither deadline covers the part that now actually blocks. The budget those call sites were written to enforce is currently decorative.

Smallest fix in this PR's scope: the four call sites already hold a ctx — switch them to `sendTextCtx`:

- `outbound.go:139`
- `outbound.go:231`
- `replier.go:189`
- `replier.go:213` (`post` still has `_ = ctx` at the top)

Whether to take the round trip off the publishing goroutine entirely is a bigger call and fine as a follow-up — but the existing bound should at least apply.

## Nits

1. `wecom_channel.go:225-227` — `if cbErr != nil { err = cbErr }` overwrites unconditionally, so a clean ctx-cancel shutdown with a callback in flight turns the read loop's `nil` into an error. Harmless on the parent-ctx path (the supervisor returns before it logs), but a lease-loss `runCancel` produces a spurious `connection exited with error`. Guarding on `ctx.Err() == nil` covers it.
2. `ackTimeout` is a package const, so `TestSendTextDistinguishesALostAckFromARefusal` genuinely sleeps 5s — the package runs in 5.5s here, not the 0.7s the description claims. A field on `wsSender` (or a package var) lets the test set 20ms and gets that number back for real.
3. `sendTextCtx` has no caller that passes a real ctx today; everything routes through `sendText`. Fixing must-fix 2 is what makes it earn the split.
4. On the read-error exit, the deferred `close(callbacks); <-cbDone` still runs up to 64 queued callbacks against a socket that's already gone, so their replies write into the void. Not a blocker, but those messages get consumed with nothing said back — might be worth discarding rather than draining when the transport is known dead.

---

Requesting changes on must-fix 1 — once triggered it's unrecoverable without a restart and it holds the lease while it's stuck, and the fix is three lines. Must-fix 2 is worth folding into the same round. Everything else is a nit.

seacen:
Both must-fixes are in. I reproduced must-fix 1 independently before reading your review and landed on the same three-line arm you sketched, which is a good sign for it.

## Must-fix 1 — agreed on every point, including the parts that are easy to get wrong

Your diagnosis holds exactly. Two things I checked specifically because they were the places your reasoning could have slipped, and neither does:

`conn.Close()` really is no rescue. Goroutine dump at the wedge, with the fix backed out:

```
goroutine 40 [select]:
...wecom.(*wecomChannel).Connect(...)
	wecom_channel.go:274 +0x9e4
```

274 is the hand-off select. The worker's goroutine is absent from the dump entirely — it returned. So: read loop parked on the send, receiver gone, no closer, ctx live.

And the full queue is genuinely load-bearing, not incidental. Same test with 2 callbacks instead of `callbackQueueDepth+2` passes in 0.00s; with the full queue it hits the bounded wait every time. The necessary condition is a dispatch error *while the buffer is saturated*, which is why this never showed up in the existing lifecycle tests.

One thing to add to your lease argument, which I think makes it stronger than you stated. You said no other replica can take the lease from the wedged one. The reason is in the renewal CAS (`channel.sql:311-325`):

```sql
AND ( ws_lease_token IS NULL
      OR ws_lease_expires_at < now()
      OR ws_lease_token = sqlc.arg('new_token') )   -- always us
```

The third arm means our own renewal succeeds unconditionally, so `ws_lease_expires_at` never lapses, so the `!leased → cancelRun()` branch in `renewLeaseUntil` is *structurally unreachable* while we are renewing. The one automatic teardown path in the design is disabled by precisely the hang it would need to catch. There is also no watchdog anywhere: every timeout in `supervisor.go` (`LeaseReleaseTimeout`, `DisconnectTimeout`, the backoff sleep) guards a path only reachable *after* line 510 has already returned, and nothing in `engine/` emits a metric. So the wedge is silent as well as unrecoverable.

Fix is your arm, plus a comment recording why `conn.Close()` doesn't cover it so nobody removes it later. Backpressure over dropping is untouched: while the worker is alive a full queue blocks the read loop exactly as before. It stops being backpressure only once nobody is receiving, and that is the only case the arm changes.

I also took nit 1 — `err = cbErr` now guards on `ctx.Err() == nil`, so a lease-loss `runCancel` that catches a callback in flight no longer produces a spurious `connection exited with error`.

**Residual worth naming, because it is the other half of the same hole:** this closes "the worker *died*", not "the worker *hung*". If a dispatch blocks forever without returning, `cbDone` never closes, the read loop still parks on the send, and you get the identical lease wedge — silent, unrecoverable, same blast radius, different trigger.

I went looking for whether a dispatch can actually block forever, expecting to be able to write this off as theoretical. It is not.

The good news is that no agent or LLM wait is involved. `Router.Handle` detaches the run — `scheduleReply` returns immediately and `batcher.Schedule` just arms a `time.AfterFunc` (`router.go:237`, `router.go:653`), so the model never runs on the callback worker. What is left inline is six or seven DB round trips, and those are bounded only by the caller's ctx: there is no `statement_timeout`, no `lock_timeout`, and no pool acquire timeout configured anywhere in the process. `newDBPool` sets `MaxConns`/`MinConns` and nothing else (`dbstats.go:54-83`), and that 25-connection pool is shared with daemon claim traffic that the same file records as having produced ~900ms acquire waits in production. `pgxpool.Acquire` waits on ctx alone, so pool starvation parks `Handle` for as long as the starvation lasts. There is also a named lock wait on the inline path: `LockWorkspaceForChatSessionCreate` (`session.go:232`) takes `FOR KEY SHARE` on the workspace row, which the query's own comment notes conflicts with `DeleteWorkspace`'s `FOR UPDATE`.

So the honest statement is: the trigger is Postgres rather than the agent, but it is a lived condition, not a hypothetical.

I still think it belongs in a separate change rather than this one, and the reason is the shape of the fix rather than the severity. "The receiver is gone" is a fact the runtime already knows, which is why it costs three lines and gets a deterministic test. "The receiver is taking too long" needs a bound, and picking that number is a product decision — too tight and healthy traffic loses its socket under ordinary DB latency, too loose and the hole stays open. That deserves its own reasoning, its own test, and a reviewer looking at it on its own rather than as the third item on a PR that already carries two must-fixes.

One caveat on my own framing, since it cuts against me: the ctx the worker dispatches with is the connection ctx, so an ordinary shutdown or lease loss does cancel it and unstick the pgx calls. That does not rescue the wedge — in the wedge nothing cancels it, because the Supervisor is itself blocked on `Connect` — but it does mean the hang case is only unrecoverable in exactly the same circumstance as the one you found, not more broadly.

Happy to file it with the above if you agree with the read.

## Must-fix 2 — agreed, and it surfaced a latent nil ctx

All four call sites now go through `sendTextCtx`, with a ctx threaded into `postPrivate` so it can. `request` already checks ctx up front and selects on `ctx.Done()`, so the 10s in `outbound.go` and the 5s in `handleInboxNew` do what they were written to do instead of being decorative.

Dropping `_ = ctx` from `post` turned up something it had been hiding: `replier_test.go:139` was passing a literal `nil` ctx, which panicked under `-race` the moment the ctx was actually read. Test-only — every production caller passes a real one — but it means `post`'s ctx parameter had never once been exercised. Fixed in the same commit.

Your framing that taking the round trip off the publishing goroutine entirely is a bigger call: agreed, leaving that as a follow-up. This only makes the existing bound real.

## Nits

- **1** — taken, as above.
- **2** (`ackTimeout` as a field so the lost-ack test doesn't genuinely sleep 5s) — not taken, to keep this round to the two must-fixes. You are right about the number; the package is ~5.5s and the description's 0.7s is wrong.
- **3** — resolves itself with must-fix 2; `sendTextCtx` now has four real callers.
- **4** — not taken, and worth being precise about scope: it applies to the *read-error* exit, where the worker is still alive and the deferred `close(callbacks); <-cbDone` drains queued callbacks against a dead socket. On the new `cbDone` path it does not arise — the worker is already gone, so those frames are discarded rather than answered into the void. Still agree it deserves a follow-up.

## Verification

New tests, both bounded so a regression fails fast rather than hanging CI:

- `TestConnectReturnsWhenCallbackWorkerDiesWithAFullQueue` — pumps `callbackQueueDepth+2` callbacks, first handler parks then errors, asserts `Connect` returns the dispatch error within 3s. Backed out: fails at 3.00s. Restored: passes in 0.00s. 25 sequential runs and 10 under `-race`, no flakes.
- `TestPost_HonoursTheCallersDeadline` — `post` on an already-cancelled ctx must return `context.Canceled` rather than serve out the ack wait. Backed out: returns `<nil>`, i.e. the send went out and the cancellation was ignored outright. Restored: passes.

`go build ./...`, `go vet ./...`, `gofmt` clean; `go test -race ./internal/integrations/wecom/` green locally, and CI is green on the push — all nine checks, `backend` included at 5m2s.

Bohan-J (approved):
Both must-fixes are closed, and the commits that close them read better than the fixes I sketched.

**Must-fix 1** — the `cbDone` arm is in the hand-off select, and the deferred error substitution is now guarded on a live ctx (which also closes nit 1). I verified it independently rather than relying on the added test: my own repro — a conn that streams `aibot_msg_callback` frames endlessly, a handler whose first call parks and then errors, released once the read loop is parked on the send with 64 buffered behind it — now has `Connect` returning in 0.11s with the genuine `handler infra failure`. On the previous head the same setup never returned. `TestConnectReturnsWhenCallbackWorkerDiesWithAFullQueue` builds the real precondition too, not a weakened version of it.

**Must-fix 2** — all four ctx-holding call sites go through `sendTextCtx`, and `postPrivate` threads the ctx rather than reaching for `Background()`. `TestPost_HonoursTheCallersDeadline` pins it so it can't quietly regress.

I rebased onto `main` and resolved the one conflict in `replier_test.go` — `main` had added `TestSendBindingPrompt_ThrottledSendsNoURL` where this branch adds `TestPost_HonoursTheCallersDeadline`. Both are kept, and `main`'s new test needed `conn.autoAck(newWSSender(conn, nil))`: without it that test writes two frames against a double that never answers and would sit out the ack wait twice. It passes in 0.00s with the change.

Verified on the rebased head: `go build ./...`, `go vet ./internal/integrations/wecom/...`, `go test -race ./internal/integrations/wecom/` all pass; `gofmt` and `git diff --check` clean; the diff stays inside the wecom package.

Remaining items are non-blocking and better as follow-ups, as discussed: the lost-ack test still spends a real 5s because `ackTimeout` is a const; the read-error exit still drains the queue against a dead socket; and `sendText` without a ctx now has a single caller left, the text-only receipt inside `dispatchFrame`.

Nice work on the write-up in both fix commits — the reasoning about why `conn.Close()` doesn't move a goroutine parked on a channel send is the part most people would have left implicit.

PR #5645, opened 2026-07-19
fix(channels): keep direct chat replies in Multica

beastpu:
Closes #5644

## Summary

- classify outbound replies by task origin before consulting a channel binding
- keep web/mobile direct-chat completions and failures inside Multica even when the session was originally created from Lark or Slack
- preserve outbound delivery for channel-created tasks and add regression coverage for both integrations

## Root cause

A chat session can retain its Lark or Slack binding after a user continues the same session from the Multica UI. The outbound subscribers treated that binding as sufficient proof that every later task originated from the external channel. Direct-chat tasks are distinguishable by a non-NULL chat_input_task_id, so the subscribers now enforce that boundary before loading installation credentials or sending.

Slack also extracts the task id from the typed chat-done payload emitted by TaskService and fails closed when task origin cannot be established.

## Tests

- go test ./internal/integrations/lark ./internal/integrations/slack -count=1
- go test ./internal/handler ./internal/service
- git diff --check

A local run of go test ./cmd/server reaches the existing readiness integration test but fails because /readyz and /healthz return 503 without healthy local dependencies; the focused channel, handler, and service packages pass.

PR #6591, opened 2026-08-07
MUL-5888: fix(wecom): answer a duplicate /issue honestly, and don't push a web-UI reply into the room

seacen:
Two things WeCom is the only integration not doing. Both signals are already on `main`; the adapter just never read them.

## 1. A duplicate `/issue` is reported as created

The engine refuses a duplicate by carrying the **other** issue's id, number and title with `IssueDuplicate` set. `Reply` answers both outcomes with `issueCreatedText`, so the reporter is told their bug was filed — under a number somebody else opened, under a title they never wrote.

The failure is quiet and expensive: they stop chasing it, and the report is lost. Nobody finds out, because from their side it looks like it worked.

| | |
|---|---|
| `slack/replier.go:125` | branches on `res.IssueDuplicate` |
| `dingtalk/replier.go:125` | branches on `res.IssueDuplicate` |
| `lark` | branches on it |
| **`wecom/replier.go`** | **did not** |

## 2. A web-UI answer is pushed into the WeCom room

A chat session that originated in WeCom can be continued from the Multica web UI. That answer belongs only in Multica — but `processEvent` had no origin gate, so it went straight to the bound chat. **In a group that means in front of everyone in the room, answering a question none of them saw asked.**

`engine.TaskInputIsChannelIngested` has been on `main` since #5645, and `slack/outbound.go:118` gates on it right after the binding lookup and before loading credentials. This adds the same call in the same position.

**Fails closed**: a completion whose origin cannot be established is not delivered. Silence is recoverable; a leak is not.

## Tests

Four, no database needed.

The two privacy tests fail on the old code with the frame count in the message:

```
a web-UI answer was pushed into the WeCom chat (1 frame(s) written)
delivered a completion whose origin could not be established (1 frame(s))
```

One design note worth flagging: `fakeOutboundQueries.channelIngested` defaults to **false**, so a test that expects delivery has to say so explicitly — the same way production now has to establish the origin before pushing into somebody's chat. `TestProcessEvent_DeliversChatReplyToBoundChat` gained that one line.

### DIFF
--- a/server/internal/integrations/wecom/replier.go
+++ b/server/internal/integrations/wecom/replier.go
@@ -119,9 +119,21 @@ func (r *OutboundReplier) Reply(ctx context.Context, inst engine.ResolvedInstall
 		// chat message stays silent (the agent's own reply lands via
 		// EventChatDone / Channel.Send).
 		if res.IssueID.Valid {
-			if err := r.post(ctx, inst, msg, issueCreatedText(res)); err != nil {
-				r.logger.WarnContext(ctx, "wecom replier: issue-created confirmation failed",
-					"installation_id", util.UUIDToString(inst.ID), "error", err)
+			// The engine reports a duplicate by carrying the OTHER issue's
+			// id, number and title with IssueDuplicate set. Answering both
+			// cases with the created copy told the reporter their bug was
+			// filed under a number somebody else opened, under a title they
+			// never wrote — so they stopped chasing it and the report was
+			// lost. slack/replier.go:125 and dingtalk/replier.go:125 both
+			// branch here; WeCom was the one that did not.
+			text := issueCreatedText(res)
+			if res.IssueDuplicate {
+				text = issueDuplicateText(res)
+			}
+			if err := r.post(ctx, inst, msg, text); err != nil {
+				r.logger.WarnContext(ctx, "wecom replier: issue confirmation failed",
+					"installation_id", util.UUIDToString(inst.ID),
+					"duplicate", res.IssueDuplicate, "error", err)
 			}
 		}
 	}
@@ -224,6 +236,29 @@ func (r *OutboundReplier) post(ctx context.Context, inst engine.ResolvedInstalla
 	}
 	chatType := aibotChatTypeFromChannel(msg.Source.ChatType)
 	return sender.sendTextCtx(ctx, chatID, chatType, text)
+}
+
+// issueDuplicateText answers a /issue the engine refused because an active
+// issue already covers it. It names the OTHER issue, which is the whole point:
+// the reporter needs to know where the discussion already is, not to be handed
+// an id they will read as their own.
+//
+// That title belongs to the pre-existing issue, so it is text some *other*
+// member wrote — not even the reporter's own, which is what makes this the
+// worse of the two /issue call sites — and the reply ships as markdown. Hence
+// breakMemberLinks, the same entry point issueCreatedText runs its title
+// through: it breaks the inline "](" form and the link reference definition a
+// multi-line title can smuggle instead. See markdown.go.
+func issueDuplicateText(res engine.Result) string {
+	id := res.IssueIdentifier
+	if id == "" {
+		id = fmt.Sprintf("#%d", res.IssueNumber)
+	}
+	title := breakMemberLinks(strings.TrimSpace(res.IssueTitle))
+	if title == "" {
+		return "⚠️ 未创建 —— 已存在进行中的 " + id
+	}
+	return "⚠️ 未创建 —— 已存在进行中的 " + id + " — " + title
 }
 
 func issueCreatedText(res engine.Result) string {
--- a/server/internal/integrations/wecom/outbound.go
+++ b/server/internal/integrations/wecom/outbound.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/multica-ai/multica/server/internal/events"
 	"github.com/multica-ai/multica/server/internal/integrations/channel"
+	"github.com/multica-ai/multica/server/internal/integrations/channel/engine"
 	"github.com/multica-ai/multica/server/internal/util"
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 	"github.com/multica-ai/multica/server/pkg/protocol"
@@ -43,6 +44,8 @@ import (
 // subscriber needs. *db.Queries satisfies it.
 type outboundQueries interface {
 	GetChannelChatSessionBindingBySession(ctx context.Context, arg db.GetChannelChatSessionBindingBySessionParams) (db.ChannelChatSessionBinding, error)
+	GetAgentTask(ctx context.Context, id pgtype.UUID) (db.AgentTaskQueue, error)
+	TaskHasChannelIngestedMessages(ctx context.Context, taskID pgtype.UUID) (bool, error)
 	GetChannelInstallation(ctx context.Context, arg db.GetChannelInstallationParams) (db.ChannelInstallation, error)
 	FindChannelBindingForMember(ctx context.Context, arg db.FindChannelBindingForMemberParams) (db.ChannelUserBinding, error)
 	GetWorkspace(ctx context.Context, id pgtype.UUID) (db.Workspace, error)
@@ -109,6 +112,39 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 	if content == "" {
 		return nil // nothing to say (empty completion)
 	}
+	// Only bound, non-empty completions reach here, so classify the task
+	// origin before loading credentials or sending. A question asked in the
+	// Multica web UI can reuse a session that originated in WeCom — and its
+	// answer belongs only in Multica. Without this gate that answer is pushed
+	// into the WeCom chat, which in a group means in front of everyone in the
+	// room. slack/outbound.go:118 and the lark and dingtalk equivalents all
+	// gate here; WeCom was the one that did not.
+	//
+	// Fails closed: an origin we cannot establish is not delivered.
+	//
+	// Everything above this point is a read. Keep it that way: the gate has to
+	// stay ahead of anything that consumes or mutates WeCom-side state for the
+	// turn, because an answer that must not reach the room must not take over
+	// the room's message either.
+	taskID, ok := chatDoneTaskID(e)
+	if !ok {
+		return nil
+	}
+	task, err := o.q.GetAgentTask(ctx, taskID)
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			// Cancelled and deleted while its completion was in flight.
+			return nil
+		}
+		return fmt.Errorf("wecom: load agent task: %w", err)
+	}
+	deliver, err := engine.TaskInputIsChannelIngested(ctx, o.q, task)
+	if err != nil {
+		return fmt.Errorf("wecom: classify task input origin: %w", err)
+	}
+	if !deliver {
+		return nil
+	}
 	inst, err := o.q.GetChannelInstallation(ctx, db.GetChannelInstallationParams{
 		ID:          binding.InstallationID,
 		ChannelType: channelTypeWecom,
@@ -137,6 +173,24 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 	}
 	chatType := aibotChatTypeFromChannel(channel.ChatType(binding.ChatType))
 	return sender.sendTextCtx(ctx, binding.ChannelChatID, chatType, content)
+}
+
+// chatDoneTaskID recovers the task id an EventChatDone belongs to. The
+// envelope's TaskID is preferred, with the payload as the fallback —
+// service.broadcastChatDone sets ChatDonePayload.TaskID and leaves the
+// envelope's empty, so in practice the fallback is the live path.
+func chatDoneTaskID(e events.Event) (pgtype.UUID, bool) {
+	raw := e.TaskID
+	if raw == "" {
+		switch p := e.Payload.(type) {
+		case protocol.ChatDonePayload:
+			raw = p.TaskID
+		case map[string]any:
+			raw, _ = p["task_id"].(string)
+		}
+	}
+	id, err := util.ParseUUID(raw)
+	return id, err == nil && id.Valid
 }
 
 // chatDoneContent extracts the reply text from an EventChatDone payload
--- a/server/internal/integrations/wecom/outbound_test.go
+++ b/server/internal/integrations/wecom/outbound_test.go
@@ -37,6 +37,24 @@ type fakeOutboundQueries struct {
 	memberErr      error
 	workspace      db.Workspace
 	workspaceErr   error
+
+	// channelIngested is what TaskInputIsChannelIngested resolves to. The
+	// default is false — a task nobody said came from WeCom did not — so a
+	// test that expects delivery has to say so, the same way production has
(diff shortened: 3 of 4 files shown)


seacen:
Cross-reference from integrating the WeCom PRs locally: **the origin gate this PR adds gets bypassed once #6606 (streaming bubble) lands**, and neither PR shows it on its own.

#6606 writes into the chat bubble *before* where this gate sits, and its `takeTask` falls back to the head round when the completing task matches none of the open ones. So a web-UI answer arriving while a WeCom round is in flight can seize that round's bubble and be written into the chat — in a group, in front of the room. That is exactly the leak this PR exists to close, going around it.

The fix is to move the gate ahead of the stream take (still after the binding lookup, so the task reads only happen for WeCom-bound sessions). Verified empirically on my integration branch: with the gate in the naive post-merge position, a test asserting a web-originated answer never takes a WeCom bubble fails with the web answer sealing the WeCom bubble.

Nothing to change in *this* PR — it is correct as it stands. Recording it so whichever of the two merges second knows the ordering matters. Details on #6606.

server/internal/integrations/wecom/replier.go
```diff
@@ -213,6 +225,22 @@ func (r *OutboundReplier) post(ctx context.Context, inst engine.ResolvedInstalla
...
+// issueDuplicateText answers a /issue the engine refused because an active
+// issue already covers it. It names the OTHER issue, which is the whole point:
+// the reporter needs to know where the discussion already is, not to be handed
+// an id they will read as their own.
+func issueDuplicateText(res engine.Result) string {
+	id := res.IssueIdentifier
+	if id == "" {
+		id = fmt.Sprintf("#%d", res.IssueNumber)
+	}
+	title := strings.TrimSpace(res.IssueTitle)
```
Bohan-J: This title is member-authored (it belongs to the pre-existing issue) and goes out on the markdown path — `sendMsgTextBody` sets `"msgtype": "markdown"` for every `aibot_send_msg`. A title such as `安全升级：请点击 [重置密码](https://evil.example) 完成验证` therefore renders as a clickable link sent by the bot.

`issueCreatedText` gets this protection from #6592 (`breakLinkAdjacency`), but this new call site does not. Please apply the same guard here — either by rebasing onto #6592 or by carrying an equivalent one in this PR — and add a test with a hostile title asserting no adjacent `](` survives in the output.

Bohan-J (dismissed):
Thanks for this — the product direction is right on both counts. A duplicate `/issue` telling the reporter their bug was filed under someone else's number is a real way to lose a report, and an answer continued from the web UI must never land in the room. The outbound origin gate fails closed, matches the shape slack/lark/dingtalk already use, and the three new tests cover the cases that matter (normal delivery, web-UI continuation, unattributable task).

Two things before this can merge.

### 1. Must fix — the duplicate title reaches the room as unescaped markdown

`issueDuplicateText` splices `res.IssueTitle` straight into the reply, and `aibot_send_msg` accepts only `markdown` / `template_card` — `sendMsgTextBody` in `ws_frame.go` ships every reply with `"msgtype": "markdown"`. In the duplicate branch that title belongs to the *other*, pre-existing issue, i.e. text some member wrote, so a title like `安全升级：请点击 [重置密码](https://evil.example) 完成验证` arrives in the group as a working link with the bot's authority behind it. Nothing in the message marks which part is quoted from a user.

#6592 adds `breakLinkAdjacency` and applies it in `issueCreatedText`, but the duplicate path introduced here is a separate call site it does not cover, and the merge order of two independent PRs isn't something either PR can depend on. Please either rebase onto #6592 and apply `breakLinkAdjacency` to the title here as well, or carry an equivalent guard in this PR.

Either way, please add a regression test with a hostile duplicate title asserting the output contains no adjacent `](`. `TestIssueDuplicateIsNotReportedAsCreated` uses a benign title and only checks that "已创建" is absent and the identifier is present, so it cannot catch this.

### 2. Blocking — the branch now conflicts with `main`

GitHub reports `CONFLICTING` / `DIRTY`. `main` picked up 72304b34 ("throttle binding-token minting", #6584), which also touches `server/internal/integrations/wecom/replier_test.go`. The 9 green checks above are from the pre-conflict commit, so they will need to re-run after the rebase.

### Note while you're rebasing

#6606 is currently mergeable and touches `wecom/outbound.go` and `outbound_test.go` directly. If it lands first, keep the origin check ahead of the streaming-bubble takeover, and keep a regression test pinning that a web-UI answer does not consume the WeCom bubble either.

Happy to re-review as soon as the escaping and the rebase are in.

seacen:
Both fixed. I carried the guard in this PR rather than rebasing onto #6592, and the branch is rebased onto current `main` — mergeable, all 9 checks green on `870f981ab`.

### 1. Duplicate title reaching the room as markdown — fixed, guard carried here

**On the choice you offered: carried, not stacked.** #6592 is itself open, `CONFLICTING` and `CHANGES_REQUESTED` right now. Rebasing onto it puts a two-line security fix behind another PR's review round and inherits its conflict, and it doesn't remove the merge-order dependency — it only reverses which PR is stuck. Carrying costs one duplicated 40-line file, and that file is the only thing I'd have to undo. That trade felt clearly better to me, but it's your call and I'll rebase onto #6592 instead if you'd rather.

**The duplication is deliberately loud.** `breakLinkAdjacency` goes in at the same path #6592 puts it (`server/internal/integrations/wecom/markdown.go`), under the same name, with the same body. So whichever PR lands second sees git flag the file as added on both sides, and the resolution is "delete one copy" — or, if #6592's file is still byte-identical by then, git merges it with no conflict at all and there's nothing to clean up. A differently named twin would merge silently and live in the package forever, which is the outcome I wanted to avoid. The file's doc comment says all of this so whoever hits the conflict doesn't have to reconstruct it.

**Mechanism is yours, not a second invention.** Space between `]` and `(`, not backslash escaping — because #6592 already established on a live tenant that `\[` renders as a math delimiter in the bubble and as a raw backslash in the conversation-list preview.

**Applied only at the duplicate call site.** `issueDuplicateText` now runs the title through the guard; `issueCreatedText` is untouched. That's #6592's call site and it's covered there. Editing the same line here would create a real conflict in `replier.go` for no security gain, since this PR doesn't change how that path is reached.

**Regression test** — `TestIssueDuplicateTitleCannotFormALinkInTheRoom`, in `replier_test.go`. It uses your title verbatim (`安全升级：请点击 [重置密码](https://evil.example) 完成验证`) and drives the real `Reply` path rather than calling the text builder, so the assertion lands on the `markdown.content` of the frame that actually goes out. Three assertions: no `](` anywhere in the content; `MUL-99` still named; and `重置密码` plus `https://evil.example` both still present, so a guard that "fixed" this by dropping the title would fail rather than pass.

I left `TestIssueDuplicateIsNotReportedAsCreated` as it was. You're right that it can't catch this — it passes on the vulnerable code, which is visible in the reverse-verification output below.

### 2. Conflict with `main` — fixed

Rebased twice. First onto `f77d03ed7`, then `main` picked up #6590 (`0ab271235`, outbound send verdict) while I was verifying, so I rebased again onto that. Current base is `0ab271235`.

**What conflicted, both times: `server/internal/integrations/wecom/replier_test.go`, one hunk each, and nothing else.** `replier.go` and `outbound.go` auto-merged.

- Against #6584: its `TestSendBindingPrompt_ThrottledSendsNoURL` and this PR's `TestIssueDuplicateIsNotReportedAsCreated` were both appended at the same point in the file.
- Against #6590: same position again, this time its `TestPost_HonoursTheCallersDeadline`.

Both are pure additions with no shared logic, so both sides were kept in full. The care needed was on the closing braces: the `\t}` and `}` that end the last function are shared trailing lines and sit *outside* the conflict block, so keeping both sides means writing a second pair back in at the seam. Checked brace balance (107/107) and built after each resolution.

I also checked the rebase didn't quietly take our whole file anywhere — `TestConnectReturnsOnCtxCancel` is still in `wecom_channel_test.go`, the maintainer docblock is still on `types.go`, and the full diff against `main` is 5 files, +294/−6, all under `internal/integrations/wecom/`. Nothing generated was touched.

CI re-ran on the rebased head and is green — see below.

### 3. Your note on #6606 — read, and one line left in the code for it

Nothing to do in this PR: #6606 isn't on `main`, so there's no takeover to sit ahead of and no `streamStore` to write a bubble test against.

But your concern is real and worth pinning down before either of us resolves that conflict. In #6606 the bubble is taken right after the session-id parse — ahead of the binding lookup *and* ahead of the empty-content check. The origin gate here sits after both, because at that point only bound, non-empty completions have survived and the two extra queries are only paid by messages that were going to be sent anyway. So if #6606 merges first and someone resolves the conflict by leaving each side where it is, a web-UI answer gets written into the WeCom bubble and seals it — the room sees the answer, just through a different surface than the one this gate closes.

I've left the invariant in a comment at the gate in `outbound.go`: everything above it is a read, and it has to stay ahead of anything that consumes or mutates WeCom-side state for the turn. Whichever PR lands second owns the ordering. Happy to carry the bubble regression test in whichever that turns out to be, or as a follow-up if you'd rather keep both diffs where they are.

### Verification

Own worktree, throwaway Postgres, migrations applied, nothing skipped (80 tests, 0 skips):

```
$ gofmt -l server/internal/integrations/wecom/     # no output
$ go vet ./...                                     # no output
$ go build ./...                                   # no output
$ go test -race ./internal/integrations/wecom/ -count=1
ok  	github.com/multica-ai/multica/server/internal/integrations/wecom	7.485s
```

Reverse verification — reverted the guard in `issueDuplicateText` back to `title := strings.TrimSpace(res.IssueTitle)` and re-ran, verbatim:

```
=== RUN   TestIssueDuplicateIsNotReportedAsCreated
=== PAUSE TestIssueDuplicateIsNotReportedAsCreated
=== RUN   TestIssueDuplicateTitleCannotFormALinkInTheRoom
=== PAUSE TestIssueDuplicateTitleCannotFormALinkInTheRoom
=== CONT  TestIssueDuplicateIsNotReportedAsCreated
=== CONT  TestIssueDuplicateTitleCannotFormALinkInTheRoom
--- PASS: TestIssueDuplicateIsNotReportedAsCreated (0.00s)
=== NAME  TestIssueDuplicateTitleCannotFormALinkInTheRoom
    replier_test.go:434: a member-authored title formed a working markdown link in the room: "⚠️ 未创建 —— 已存在进行中的 MUL-99 — 安全升级：请点击 [重置密码](https://evil.example) 完成验证"
--- FAIL: TestIssueDuplicateTitleCannotFormALinkInTheRoom (0.00s)
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.550s
FAIL
```

That output is also the answer to your point about the existing test: `TestIssueDuplicateIsNotReportedAsCreated` passes on the vulnerable code while the link goes out intact.

CI on `870f981ab`, `MERGEABLE` / `CLEAN`, 9/9:

```
changes                     SUCCESS
backend                     SUCCESS
windows-execenv             SUCCESS
installer (ubuntu-latest)   SUCCESS
installer (macos-latest)    SUCCESS
installer (windows-latest)  SUCCESS
frontend-build              SUCCESS
frontend-test               SUCCESS
frontend                    SUCCESS
```

seacen:
Rebased onto current `main` at `7ec96ef95`, and the carried copy is gone — #6592 landed, so I undid it as promised. `MERGEABLE` / `CLEAN`, 9/9 green.

### The copy I carried was weaker than yours, and the gap was live

`markdown.go` is byte-identical to `main` now; it does not appear in this PR's diff. The only edit is the call site, which runs the duplicate title through `breakMemberLinks` — the same entry point `issueCreatedText` uses.

What I carried was `breakLinkAdjacency` alone. What you landed composes it with `breakLinkReferenceDefinitions`, and the definition form reaches a working link with **no `](` anywhere in it**:

```
安全升级

[重置密码]: https://evil.example

[重置密码]
```

So the guard I was carrying passed that straight through on the duplicate path, and the `](` test you asked for passes on it. Wiring the call site to `breakMemberLinks` closes it.

`TestIssueDuplicateTitleDefinesNoLinkReference` pins the difference, using #6592's `markdownDestinations` helper so it asks a CommonMark parser what resolves rather than matching syntax. Without it, a later simplification back to `breakLinkAdjacency` stays green:

```
--- PASS: TestIssueDuplicateTitleCannotFormALinkInTheRoom
--- FAIL: TestIssueDuplicateTitleDefinesNoLinkReference
    a member-authored title defined a link the room's bot then resolved:
    "…安全升级\n\n[重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
```

Your test is unchanged and still drives the real `Reply` path. `TestIssueDuplicateIsNotReportedAsCreated` still uses a benign title — you were right that it cannot catch this; it passes in both runs below.

Guard removed from `issueDuplicateText`:

```
--- PASS: TestIssueDuplicateIsNotReportedAsCreated
--- FAIL: TestIssueDuplicateTitleCannotFormALinkInTheRoom
    a member-authored title formed a working markdown link in the room:
    "…安全升级：请点击 [重置密码](https://evil.example) 完成验证"
--- FAIL: TestIssueDuplicateTitleDefinesNoLinkReference
```

Origin gate removed from `processEvent`:

```
--- FAIL: TestProcessEvent_FailsClosedWhenTheTaskIdIsMissing
    delivered a completion whose origin could not be established (1 frame(s))
--- FAIL: TestProcessEvent_DoesNotPushAWebUIAnswerIntoTheRoom
    a web-UI answer was pushed into the WeCom chat (1 frame(s) written)
```

`go build` stays clean with the gate deleted — it cannot see test files. `go vet` is what caught the orphaned `engine` import.

### The rebase

Base moved `0ab271235` → `cd9f4147a`. Two conflicts.

`markdown.go`, add/add — the one case where taking a whole side is right, because the side taken is yours; my copy existed to be deleted.

`replier_test.go` — your three #6592 tests and my two were appended at the same point. Both kept: 9 on main, 8 here, 11 after (6 shared + 3 + 2). The closing braces of the last function sit outside the conflict block, so a second pair goes back at the seam.

`replier.go` auto-merged correctly. `outbound.go` and `outbound_test.go` did not conflict; nothing on `main` touches them. The only lines this PR removes from `main` are the three it replaces. Diff is 4 files, +413/−10.

### Your #6606 note

**#6606 has not landed** — `OPEN` / `DIRTY`, and `main` carries none of its files. So the conditional never fired, and I have added nothing that would sit dead.

The invariant stays as a comment at the gate: everything above it is a read, and it has to stay ahead of anything that consumes WeCom-side state for the turn. #6606's `takeTask` writes the bubble right after the session-id parse, ahead of both the binding lookup and the empty-content check — so resolving that conflict by leaving each side where it is puts the takeover in front of the gate, and a web-UI answer seals the WeCom bubble. **I hit exactly that twice merging these branches locally; git reported no conflict and both `go build` and `go vet` passed.** Whichever of us lands second owns the ordering, and I will carry the bubble test in whichever PR that is.

One cleanup: the fake's `channelIngested` doc comment pointed at `failure_origin_test.go`, a #6606 file not in this tree. Reworded to name the PR.

`go build`, `go vet`, `gofmt -l` clean; `go test -race ./internal/integrations/wecom/ -count=1` ok in 12.750s, 140 tests, 0 skips, against a migrated Postgres.

multica-eve (approved):
The requested fixes are complete: duplicate issue titles are sanitized against both inline and reference-style Markdown links, regression coverage exercises the real reply path, and web-originated replies fail closed before any WeCom-side effect. The branch is rebased, mergeable, and CI is green.

PR #6592, opened 2026-08-07
fix(wecom): escape member-authored text in the inbox card, and keep the card inside the cap (MUL-5889)

seacen:
Two ways the inbox notification card mishandles text it did not write.

## 1. Member-authored link syntax renders

`title` and `body` come from whoever wrote the issue, and they go into aibot markdown unescaped.

An issue titled `[click here](http://evil.example)` arrives as a **working link**, inside a card the recipient has every reason to trust — it is delivered by the bot, not by the author, and nothing in the card marks which parts are quoted from a user.

`escapeInboxMarkdown` neutralizes `[ ] ( )` and `!`. No sibling integration has an equivalent helper (Slack builds blocks, Lark builds cards), so this one is new for the family rather than catching up to it.

## 2. A long title pushes the whole card past the cap

The truncation budgets the **body** only. When `prefix + suffix` already exceed `inboxMarkdownMaxLen` it returns them unchanged:

```go
room := inboxMarkdownMaxLen - len(prefix) - len(suffix) - 4
if room <= 0 {
    return prefix + suffix   // still over the cap
}
```

WeCom refuses the whole frame past ~4096 chars — and the send path reports success either way, so the notification is simply lost, with nothing anywhere saying so. That combination (silently over the limit, silently reported as sent) is why this is worth fixing alongside the escaping rather than after it.

The title is now truncated too, with `trimDanglingEscape` dropping a backslash left behind by cutting inside an escape pair — otherwise it would escape whatever the `...` marker puts next to it.

## Tests

Both fail on the old code, with the cost stated in the assertion:

```
member-authored link syntax rendered as a link in a bot-authored card:
"**[提及你] [click here](http://evil.example)**\nand the body [too](http://evil.example)"

card is 8053 runes, cap is 4000 — WeCom refuses the frame and the push is lost
```

### DIFF
--- a/server/internal/integrations/wecom/inbox_message.go
+++ b/server/internal/integrations/wecom/inbox_message.go
@@ -82,14 +82,23 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	// markdown card the recipient has every reason to trust — it comes from
 	// the bot, not from the author — so link syntax in them must not render.
 	// An issue titled "[click here](http://evil.example)" arrived as a
-	// working link.
+	// working link, and a body carrying "[重置密码]: https://evil.example"
+	// plus "[重置密码]" arrived as one too.
 	//
-	// Done here, before anything below measures a length: breaking the
-	// adjacency inserts a space, and a body that is dense in "](" grows by
-	// half. Budget the grown text or the card goes out over the cap, which
-	// WeCom refuses whole while telling the sender it was delivered.
-	title = breakLinkAdjacency(title)
-	body = breakLinkAdjacency(body)
+	// Per field rather than over the finished card, which is what the length
+	// budget below needs — and it costs no coverage. Every line of member
+	// text begins where the card's own line begins, bar the title's first,
+	// and there the bot's "**[" prefix keeps the member out of the block
+	// position a definition has to start in. The scan reads each field's
+	// start as a line start anyway, so that one is guarded twice over rather
+	// than not at all.
+	//
+	// Done here, before anything below measures a length: each break inserts
+	// a space, and a body dense in "](" grows by half. Budget the grown text
+	// or the card goes out over the cap, which WeCom refuses whole while
+	// telling the sender it was delivered.
+	title = breakMemberLinks(title)
+	body = breakMemberLinks(body)
 
 	var b strings.Builder
 	b.WriteString("**[")
@@ -117,11 +126,14 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	if link != "" {
 		suffix = "\n[查看详情](" + link + ")"
 	}
-	// Neither cut below can put a "]" back next to a "(". truncateRunes only
-	// drops a suffix, so it cannot recreate an adjacency breakLinkAdjacency
-	// already separated, and every literal spliced in after member text here
-	// starts with ".", "*" or "\n" — never "(". Pinned by the two seam cases
-	// in TestInboxCardNeverPutsCloseBracketNextToOpenParen.
+	// Neither cut below can put a "]" back next to a "(" or a ":".
+	// truncateRunes only drops a suffix, so it cannot recreate a pair
+	// breakMemberLinks already separated, and every literal spliced in after
+	// member text here starts with ".", "*" or "\n" — never "(" or ":".
+	// Nor can dropping a suffix complete a reference definition: it can only
+	// cut the destination off one. Pinned by the seam cases in
+	// TestInboxCardNeverPutsCloseBracketNextToOpenParen and
+	// TestInboxCardDefinesNoResolvableLinkReference.
 	room := inboxMarkdownMaxLen - utf8.RuneCountInString(prefix) - utf8.RuneCountInString(suffix) - 4 // "\n...\n"
 	if room > 0 {
 		return prefix + "\n" + truncateRunes(body, room) + "..." + suffix
--- a/server/internal/integrations/wecom/inbox_message.go
+++ b/server/internal/integrations/wecom/inbox_message.go
@@ -83,8 +83,13 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	// the bot, not from the author — so link syntax in them must not render.
 	// An issue titled "[click here](http://evil.example)" arrived as a
 	// working link.
-	title = escapeInboxMarkdown(title)
-	body = escapeInboxMarkdown(body)
+	//
+	// Done here, before anything below measures a length: breaking the
+	// adjacency inserts a space, and a body that is dense in "](" grows by
+	// half. Budget the grown text or the card goes out over the cap, which
+	// WeCom refuses whole while telling the sender it was delivered.
+	title = breakLinkAdjacency(title)
+	body = breakLinkAdjacency(body)
 
 	var b strings.Builder
 	b.WriteString("**[")
@@ -112,6 +117,11 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	if link != "" {
 		suffix = "\n[查看详情](" + link + ")"
 	}
+	// Neither cut below can put a "]" back next to a "(". truncateRunes only
+	// drops a suffix, so it cannot recreate an adjacency breakLinkAdjacency
+	// already separated, and every literal spliced in after member text here
+	// starts with ".", "*" or "\n" — never "(". Pinned by the two seam cases
+	// in TestInboxCardNeverPutsCloseBracketNextToOpenParen.
 	room := inboxMarkdownMaxLen - utf8.RuneCountInString(prefix) - utf8.RuneCountInString(suffix) - 4 // "\n...\n"
 	if room > 0 {
 		return prefix + "\n" + truncateRunes(body, room) + "..." + suffix
@@ -124,10 +134,17 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	fixed := utf8.RuneCountInString(prefix) - utf8.RuneCountInString(title)
 	titleRoom := inboxMarkdownMaxLen - fixed - utf8.RuneCountInString(suffix) - 3 // "..."
 	if titleRoom < 0 {
+		// Clamping here still returns fixed+3+len(suffix) runes with no cap
+		// check, so in principle this branch can hand back a card over the
+		// cap. Reaching it needs a link suffix past ~3987 runes, i.e. an
+		// appURL plus workspace slug of that length; both are operator- and
+		// DB-constrained, never member-authored. Left as-is deliberately:
+		// the alternative is dropping the link to make room, and no input
+		// that exists can ask for it.
 		titleRoom = 0
 	}
 	return "**[" + inboxTypeLabel(typeStr) + "] " +
-		trimDanglingEscape(truncateRunes(title, titleRoom)) + "...**" + suffix
+		truncateRunes(title, titleRoom) + "...**" + suffix
 }
 
 // inboxItemBody extracts the body/description string from an inbox_item map.
@@ -181,48 +198,6 @@ func inboxItemIssueID(item map[string]any) string {
 		return v
 	}
 	return ""
-}
-
-// escapeInboxMarkdown neutralizes link syntax in member-authored text.
-// aibot markdown renders [text](url) as a link, and the card carries the
-// bot's authority, so an issue title is not a place a link may be smuggled
-// in from.
-func escapeInboxMarkdown(s string) string {
-	if !strings.ContainsAny(s, "[]()!") {
-		return s
-	}
-	var b strings.Builder
-	b.Grow(len(s) + 8)
-	for _, r := range s {
-		switch r {
-		case '[', ']', '(', ')':
-			b.WriteByte('\\')
-			b.WriteRune(r)
-		case '!':
-			// Only meaningful immediately before a link, but escaping it
-			// unconditionally is simpler than tracking position and costs an
-			// exclamation mark a backslash it does not need.
-			b.WriteByte('\\')
-			b.WriteRune(r)
-		default:
-			b.WriteRune(r)
-		}
-	}
-	return b.String()
-}
-
-// trimDanglingEscape drops a trailing backslash left behind by truncating in
-// the middle of an escape pair — it would otherwise escape whatever the
-// truncation marker puts next to it.
-func trimDanglingEscape(s string) string {
-	n := 0
-	for i := len(s) - 1; i >= 0 && s[i] == '\\'; i-- {
-		n++
-	}
-	if n%2 == 1 {
-		return s[:len(s)-1]
-	}
-	return s
 }
 
 // truncateRunes trims s to at most maxRunes runes. Rune-based rather than
--- a/server/internal/integrations/wecom/inbox_message.go
+++ b/server/internal/integrations/wecom/inbox_message.go
@@ -78,6 +78,14 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 	body := inboxItemBody(item)
 	link := inboxItemLink(item, workspaceID, slug)
 
+	// title and body are written by a member. They land inside an aibot
+	// markdown card the recipient has every reason to trust — it comes from
+	// the bot, not from the author — so link syntax in them must not render.
+	// An issue titled "[click here](http://evil.example)" arrived as a
+	// working link.
+	title = escapeInboxMarkdown(title)
+	body = escapeInboxMarkdown(body)
+
 	var b strings.Builder
 	b.WriteString("**[")
 	b.WriteString(inboxTypeLabel(typeStr))
@@ -105,10 +113,21 @@ func buildInboxMarkdown(item map[string]any, workspaceID, slug string) string {
 		suffix = "\n[查看详情](" + link + ")"
 	}
 	room := inboxMarkdownMaxLen - utf8.RuneCountInString(prefix) - utf8.RuneCountInString(suffix) - 4 // "\n...\n"
-	if room <= 0 {
-		return prefix + suffix
+	if room > 0 {
+		return prefix + "\n" + truncateRunes(body, room) + "..." + suffix
 	}
-	return prefix + "\n" + truncateRunes(body, room) + "..." + suffix
+	// No room for any body means the prefix itself is the problem — a long
+	// enough title pushes prefix+suffix past the cap on its own, and
+	// returning it unchanged means WeCom refuses the whole frame while the
+	// sender is told it was delivered. Truncate the title so what goes out
+	// always fits.
+	fixed := utf8.RuneCountInString(prefix) - utf8.RuneCountInString(title)
+	titleRoom := inboxMarkdownMaxLen - fixed - utf8.RuneCountInString(suffix) - 3 // "..."
+	if titleRoom < 0 {
+		titleRoom = 0
+	}
+	return "**[" + inboxTypeLabel(typeStr) + "] " +
+		trimDanglingEscape(truncateRunes(title, titleRoom)) + "...**" + suffix
 }
 
 // inboxItemBody extracts the body/description string from an inbox_item map.
@@ -162,6 +181,48 @@ func inboxItemIssueID(item map[string]any) string {
 		return v
 	}
 	return ""
+}
+
+// escapeInboxMarkdown neutralizes link syntax in member-authored text.
+// aibot markdown renders [text](url) as a link, and the card carries the
+// bot's authority, so an issue title is not a place a link may be smuggled
+// in from.
+func escapeInboxMarkdown(s string) string {
+	if !strings.ContainsAny(s, "[]()!") {
+		return s
+	}
+	var b strings.Builder
+	b.Grow(len(s) + 8)
+	for _, r := range s {
+		switch r {
+		case '[', ']', '(', ')':
+			b.WriteByte('\\')
+			b.WriteRune(r)
+		case '!':
+			// Only meaningful immediately before a link, but escaping it
+			// unconditionally is simpler than tracking position and costs an
+			// exclamation mark a backslash it does not need.
+			b.WriteByte('\\')
+			b.WriteRune(r)
+		default:
+			b.WriteRune(r)
+		}
+	}
+	return b.String()
+}
+
+// trimDanglingEscape drops a trailing backslash left behind by truncating in
+// the middle of an escape pair — it would otherwise escape whatever the
+// truncation marker puts next to it.
+func trimDanglingEscape(s string) string {
+	n := 0
+	for i := len(s) - 1; i >= 0 && s[i] == '\\'; i-- {
+		n++
+	}
+	if n%2 == 1 {
+		return s[:len(s)-1]
+	}
+	return s
 }
 
 // truncateRunes trims s to at most maxRunes runes. Rune-based rather than
--- a/server/internal/integrations/wecom/inbox_message_test.go
+++ b/server/internal/integrations/wecom/inbox_message_test.go
@@ -288,6 +288,130 @@ func TestInboxCardBudgetsTheSpacesItInserts(t *testing.T) {
 	}
 }
 
+// TestInboxCardDefinesNoResolvableLinkReference is the card-level half of the
+// reference-definition guard: it proves the builder actually runs it, on both
+// fields, and that neither truncation branch hands the definition back.
+//
+// The card is the whole document — one buildInboxMarkdown result becomes the
+// "markdown.content" of exactly one aibot_send_msg frame, and nothing batches
+// two items into one frame — so the definition and the label that uses it both
+// have to be inside a single member's title or body for the attack to work.
+// Which is what the live tenant showed: one comment body carried both.
+//
+// The app URL is cleared so the builder emits no link of its own; any link in
+// the parsed output is then member-authored.
+func TestInboxCardDefinesNoResolvableLinkReference(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	for _, tc := range linkReferenceAttacks {
+		t.Run(tc.name, func(t *testing.T) {
+			// A definition cannot begin mid-paragraph, so put a blank line
+			// between the title's line and the attack — the shape a comment
+			// with a second paragraph has anyway.
+			body := "看这里\n\n" + tc.body
+
+			// The same card without the guard, composed the way
+			// buildInboxMarkdown composes it. Asserting it resolves keeps
+			// this case from quietly becoming a test of nothing.
+			unguarded := "**[提及你] t**\n" + body
+			if !hasDestinationTo(markdownDestinations(unguarded), "evil.example") {
+				t.Fatalf("unguarded card %q resolves no link to evil.example — this case proves nothing", unguarded)
+			}
+
+			out := buildInboxMarkdown(map[string]any{
+				"type": "mentioned", "title": "t", "body": body,
+			}, "ws-uuid", "acme")
+			if dests := markdownDestinations(out); hasDestinationTo(dests, "evil.example") {
+				t.Fatalf("body: a member-defined link survived into the card: %q resolves %v", out, dests)
+			}
+
+			// Same through the title. A title with a line break can carry a
+			// definition too, and it goes through a different length branch.
+			out = buildInboxMarkdown(map[string]any{
+				"type": "mentioned", "title": body, "body": "b",
+			}, "ws-uuid", "acme")
+			if dests := markdownDestinations(out); hasDestinationTo(dests, "evil.example") {
+				t.Fatalf("title: a member-defined link survived into the card: %q resolves %v", out, dests)
+			}
+		})
+	}
+}
+
+// TestInboxCardKeepsAReferenceLikeTitleVerbatim is the constraint that shaped
+// the rule. "[Bug]: 登录失败" is an everyday title, and a guard that broke every
+// "]:" would trade it away to close an attack that needs a real destination —
+// the same trade the backslash escaping already lost on this renderer. Byte
+// for byte, or the guard is too blunt.
+func TestInboxCardKeepsAReferenceLikeTitleVerbatim(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	item := map[string]any{
+		"type":  "status_changed",
+		"title": "[Bug]: 登录失败",
+		"body":  "[WIP]: 明天再看\n\n[复现步骤]: 见 (todo) 到 (in_review)!",
+	}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	want := "**[状态变更] [Bug]: 登录失败**\n[WIP]: 明天再看\n\n[复现步骤]: 见 (todo) 到 (in_review)!"
+	if out != want {
+		t.Fatalf("member text did not survive verbatim:\n got %q\nwant %q", out, want)
+	}
+}
+
+// TestInboxCardSeamKeepsDefinitionsBrokenAtEveryCutOffset sweeps the
+// truncation point across a dense run of definitions. Cutting drops a suffix,
+// so it cannot fuse a "]" back onto the ":" the guard separated — this is that
+// claim under test rather than in a comment, at every offset inside the
+// pattern, and it holds the card inside the cap while it is at it.
+func TestInboxCardSeamKeepsDefinitionsBrokenAtEveryCutOffset(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	for pad := 0; pad < 12; pad++ {
+		filler := strings.Repeat("a", 3900+pad)
+		dense := strings.Repeat("[x]: https://evil.example\n\n[x]\n\n", 40)
+		for _, tc := range []struct {
+			name string
+			item map[string]any
+		}{
+			{"body", map[string]any{"type": "mentioned", "title": "t", "body": filler + "\n\n" + dense}},
+			{"title", map[string]any{"type": "mentioned", "title": filler + "\n\n" + dense, "body": "b"}},
+		} {
+			out := buildInboxMarkdown(tc.item, "ws-uuid", "acme")
+			if dests := markdownDestinations(out); hasDestinationTo(dests, "evil.example") {
+				t.Fatalf("%s cut at pad=%d let a member-defined link through: resolves %v", tc.name, pad, dests)
+			}
+			if n := utf8.RuneCountInString(out); n > inboxMarkdownMaxLen {
+				t.Fatalf("%s cut at pad=%d: card is %d runes, cap is %d — WeCom refuses the frame and the push is lost",
+					tc.name, pad, n, inboxMarkdownMaxLen)
+			}
+		}
+	}
+}
+
+// TestInboxCardBudgetsTheSpacesTheDefinitionBreakInserts is the "](" budget
+// case for the other break: a body that is nothing but definitions grows by
+// one rune per line, and the budget has to be computed on the grown text or
+// the card ships over the cap and WeCom drops the whole frame while the send
+// path reports success.
+func TestInboxCardBudgetsTheSpacesTheDefinitionBreakInserts(t *testing.T) {
+	t.Setenv("MULTICA_APP_URL", "https://multica.example")
+	item := map[string]any{
+		"type":  "mentioned",
+		"title": "t",
+		"body":  strings.Repeat("[a]: https://evil.example\n", 500),
+	}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	if n := utf8.RuneCountInString(out); n > inboxMarkdownMaxLen {
+		t.Fatalf("card is %d runes, cap is %d — the inserted spaces were not budgeted, so WeCom refuses the frame and the push is lost",
+			n, inboxMarkdownMaxLen)
+	}
+}
+
 // window returns a short readable slice around i for a failure message.
 func window(s string, i int) string {
 	lo, hi := i-10, i+10
--- a/server/internal/integrations/wecom/inbox_message_test.go
+++ b/server/internal/integrations/wecom/inbox_message_test.go
@@ -3,6 +3,7 @@ package wecom
 import (
 	"os"
 	"strings"
+	"unicode/utf8"
 	"testing"
 )
 
@@ -130,5 +131,39 @@ func TestTruncateRunes(t *testing.T) {
 		if got := truncateRunes(tc.in, tc.max); got != tc.expect {
 			t.Errorf("truncateRunes(%q,%d)=%q; want %q", tc.in, tc.max, got, tc.expect)
 		}
+	}
+}
+
+// TestInboxCardDoesNotRenderMemberAuthoredLinks: the card carries the bot's
+// authority — it arrives from the bot, not from whoever wrote the issue — so
+// link syntax in a member-written title must not render as a link.
+func TestInboxCardDoesNotRenderMemberAuthoredLinks(t *testing.T) {
+	item := map[string]any{
+		"type":  "mentioned",
+		"title": "[click here](http://evil.example)",
+		"body":  "and the body [too](http://evil.example)",
+	}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	if strings.Contains(out, "](http://evil.example)") {
+		t.Fatalf("member-authored link syntax rendered as a link in a bot-authored card: %q", out)
+	}
+	if !strings.Contains(out, "click here") {
+		t.Errorf("escaping ate the visible text: %q", out)
+	}
+}
+
+// TestInboxCardFitsTheCapEvenWithAHugeTitle: WeCom refuses the whole frame
+// past the cap, and the sender is told it was delivered — so a card that does
+// not fit is a message silently lost, not a message truncated.
+func TestInboxCardFitsTheCapEvenWithAHugeTitle(t *testing.T) {
+	huge := strings.Repeat("标题", 4000) // far past inboxMarkdownMaxLen on its own
+	t.Setenv("MULTICA_APP_URL", "https://multica.example")
+	item := map[string]any{"type": "mentioned", "title": huge, "body": "body"}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	if n := utf8.RuneCountInString(out); n > inboxMarkdownMaxLen {
+		t.Fatalf("card is %d runes, cap is %d — WeCom refuses the frame and the push is lost", n, inboxMarkdownMaxLen)
+	}
+	if !strings.Contains(out, "查看详情") {
+		t.Errorf("the view-detail affordance was dropped: %q", out[:120])
 	}
 }
--- a/server/internal/integrations/wecom/inbox_message_test.go
+++ b/server/internal/integrations/wecom/inbox_message_test.go
@@ -3,8 +3,8 @@ package wecom
 import (
 	"os"
 	"strings"
-	"unicode/utf8"
 	"testing"
+	"unicode/utf8"
 )
 
 func TestBuildInboxMarkdown_TitleBodyLink(t *testing.T) {
@@ -148,7 +148,7 @@ func TestInboxCardDoesNotRenderMemberAuthoredLinks(t *testing.T) {
 		t.Fatalf("member-authored link syntax rendered as a link in a bot-authored card: %q", out)
 	}
 	if !strings.Contains(out, "click here") {
-		t.Errorf("escaping ate the visible text: %q", out)
+		t.Errorf("the visible text was eaten: %q", out)
 	}
 }
 
@@ -166,4 +166,136 @@ func TestInboxCardFitsTheCapEvenWithAHugeTitle(t *testing.T) {
 	if !strings.Contains(out, "查看详情") {
 		t.Errorf("the view-detail affordance was dropped: %q", out[:120])
 	}
+}
+
+// TestInboxCardKeepsAnOrdinaryBracketedTitleVerbatim pins the reason the
+// mechanism changed. A live tenant showed backslash escaping is unusable here:
+// "\[Bug\]" came back as an italic serif "Bug" with the brackets consumed —
+// the client reads "\[...\]" as a display-math delimiter — and the
+// conversation-list preview, which renders no markdown at all, showed the
+// backslashes raw. "[Bug] ..." and "[WIP] ..." are everyday title forms, so
+// the card must carry them through untouched.
+func TestInboxCardKeepsAnOrdinaryBracketedTitleVerbatim(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	item := map[string]any{
+		"type":  "status_changed",
+		"title": "[Bug] 登录失败",
+		"body":  "从 (todo) 到 (in_review)!",
+	}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	want := "**[状态变更] [Bug] 登录失败**\n从 (todo) 到 (in_review)!"
+	if out != want {
+		t.Fatalf("member text did not survive verbatim:\n got %q\nwant %q", out, want)
+	}
+	if strings.Contains(out, `\`) {
+		t.Fatalf("a backslash reached the card: %q — WeCom either shows it raw in the list preview or reads it as a math delimiter in the bubble", out)
+	}
+}
+
+// TestInboxCardNeverPutsCloseBracketNextToOpenParen pins the property the
+// card's safety rests on: "](" never comes out adjacent. An inline link needs
+// that adjacency under CommonMark ("immediately followed by") and under a
+// naive `\[([^\]]+)\]\(([^)]+)\)` rewriter alike, and "![" needs it too, so
+// neither a link nor an image can form.
+//
+// The app URL is cleared so the builder emits no link of its own; any "]("
+// in the output would then be member-authored.
+func TestInboxCardNeverPutsCloseBracketNextToOpenParen(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	cases := []struct {
+		name  string
+		title string
+		body  string
+	}{
+		{"link in title", "[click here](http://evil.example)", "body"},
+		{"link in body", "title", "and the body [too](http://evil.example)"},
+		{"image", "![img](http://evil.example/x.png)", "body"},
+		{"nested brackets", "[a[b]](http://evil.example)", "body"},
+		// A member putting a backslash in front of the pair, betting the
+		// separator lands somewhere it can be undone.
+		{"member-written backslash", `x\](http://evil.example)`, `y\](http://evil.example)`},
+		// Both truncation branches: the cut must not fuse a "]" and a "("
+		// that were separated.
+		{"body truncated at the seam", "t", strings.Repeat("a", 3900) + strings.Repeat("](x)", 100)},
+		{"title truncated at the seam", strings.Repeat("标题](x)", 2000), "body"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			out := buildInboxMarkdown(map[string]any{
+				"type": "mentioned", "title": tc.title, "body": tc.body,
+			}, "ws-uuid", "acme")
+			if i := strings.Index(out, "]("); i >= 0 {
+				t.Fatalf("\"](\" came out adjacent at byte %d — that is a working link in a card the bot signs: %q",
+					i, window(out, i))
+			}
+		})
+	}
+}
+
+// TestInboxCardSeamSurvivesEveryCutOffset sweeps the truncation point across a
+// dense run of "](" so some iteration cuts at every offset inside the pattern.
+// The two fixed seam cases above happen to land on one offset each; this is
+// the same property without the luck.
+func TestInboxCardSeamSurvivesEveryCutOffset(t *testing.T) {
+	t.Setenv("WECOM_APP_URL", "")
+	t.Setenv("MULTICA_APP_URL", "")
+	t.Setenv("FRONTEND_ORIGIN", "")
+
+	for pad := 0; pad < 12; pad++ {
+		body := strings.Repeat("a", 3900+pad) + strings.Repeat("](x)", 60)
+		title := strings.Repeat("标", 3900+pad) + strings.Repeat("](x)", 60)
+		for _, tc := range []struct {
+			name string
+			item map[string]any
+		}{
+			{"body", map[string]any{"type": "mentioned", "title": "t", "body": body}},
+			{"title", map[string]any{"type": "mentioned", "title": title, "body": "b"}},
+		} {
+			out := buildInboxMarkdown(tc.item, "ws-uuid", "acme")
+			if i := strings.Index(out, "]("); i >= 0 {
+				t.Fatalf("%s cut at pad=%d fused a \"]\" onto a \"(\" at byte %d: %q", tc.name, pad, i, window(out, i))
+			}
+			if n := utf8.RuneCountInString(out); n > inboxMarkdownMaxLen {
+				t.Fatalf("%s cut at pad=%d: card is %d runes, cap is %d — WeCom refuses the frame and the push is lost",
+					tc.name, pad, n, inboxMarkdownMaxLen)
+			}
+		}
+	}
+}
+
+// TestInboxCardBudgetsTheSpacesItInserts: breaking the adjacency inserts a
+// visible character, so a body dense in "](" is longer going out than it was
+// coming in — 1.5x in the limit. The budget has to be computed on the grown
+// text. Measured after truncation, the card would ship over the cap and WeCom
+// would refuse the whole frame while the send path reports success.
+func TestInboxCardBudgetsTheSpacesItInserts(t *testing.T) {
+	t.Setenv("MULTICA_APP_URL", "https://multica.example")
+	item := map[string]any{
+		"type":  "mentioned",
+		"title": "t",
+		"body":  strings.Repeat("](", 4000), // 8000 runes in, 12000 if broken
+	}
+	out := buildInboxMarkdown(item, "ws-uuid", "acme")
+	if n := utf8.RuneCountInString(out); n > inboxMarkdownMaxLen {
+		t.Fatalf("card is %d runes, cap is %d — the inserted spaces were not budgeted, so WeCom refuses the frame and the push is lost",
+			n, inboxMarkdownMaxLen)
+	}
+}
+
+// window returns a short readable slice around i for a failure message.
+func window(s string, i int) string {
+	lo, hi := i-10, i+10
+	if lo < 0 {
+		lo = 0
+	}
+	if hi > len(s) {
+		hi = len(s)
+	}
+	return s[lo:hi]
 }
--- a/server/internal/integrations/wecom/markdown.go
+++ b/server/internal/integrations/wecom/markdown.go
@@ -96,6 +96,13 @@ func breakLinkAdjacency(s string) string {
 //  3. What follows the colon is plausibly a link *destination* — see
 //     looksLikeLinkDestination.
 //
+// Conditions 1 and 3 are halves of one question and are kept honest by one
+// answer: containerPrefixBefore reports the block containers holding the
+// label, and looksLikeLinkDestination is handed that same containerPrefix so
+// it can step back into them when the destination sits on the next line.
+// Teaching one half about a container and not the other is what leaves a
+// definition intact — see the comment on containerPrefix.
+//
 // Condition 3 is the whole design. Breaking every "]:" would be simpler and is
 // the wrong trade: "[Bug]: 登录失败" is a form real people write, and it comes
 // back from here byte-identical because 登录失败 names no host — as a
@@ -106,10 +113,11 @@ func breakLinkAdjacency(s string) string {
 //
 // # What it deliberately lets through
 //
-//   - Relative destinations. "[文件]: report.pdf", "[页面]: /inbox" and
-//     "[Bug]: 登录失败" all still define a reference, and the label can still
-//     come out as a link — but pointing at the client's own base, not at an
-//     attacker. Paid knowingly to keep the human form intact.
+//   - Relative destinations. "[文件]: report.pdf", "[页面]: /inbox",
+//     "[Owner]: R&D", "[Regex]: \d+" and "[Bug]: 登录失败" all still define a
+//     reference, and the label can still come out as a link — but pointing at
+//     the client's own base, not at an attacker. Paid knowingly to keep the
+//     human form intact.
 //   - "[foo]:(https://evil.example)". CommonMark keeps the parens inside the
 //     href, so it is not a URL anyone navigates to. A leading "(" is skipped
 //     anyway before the destination is examined, in case this renderer is
@@ -117,10 +125,12 @@ func breakLinkAdjacency(s string) string {
 //
 // # Where it over-fires, and why that is the safe side
 //
-// It does not model block context, so it fires on a "[x]: https://…" line that
-// CommonMark would fold into the preceding paragraph or read as indented code,
-// and it does not check the spec's "nothing may follow the destination" rule,
-// so "[x]: https://evil.example 请点击" is broken too though CommonMark would
+// It models block containers only as a count of ">" markers, so it fires on a
+// "[x]: https://…" line that CommonMark would fold into the preceding
+// paragraph or read as indented code, and on a next-line destination whose
+// blockquote nesting is shallower than the definition's. It does not check the
+// spec's "nothing may follow the destination" rule either, so
+// "[x]: https://evil.example 请点击" is broken too though CommonMark would
 // leave it as prose. Every one of those costs a single space on a line that
 // already carries a URL. Under-firing costs a working link under the bot's
 // name, so the approximation runs this way round on purpose — the more so
@@ -149,7 +159,11 @@ func breakLinkReferenceDefinitions(s string) string {
 	var out strings.Builder
 	prev := 0
 	for i := 0; i < len(s); i++ {
-		if s[i] != '[' || !blockScaffoldOnlyBefore(s, i) {
+		if s[i] != '[' {
+			continue
+		}
+		container, ok := containerPrefixBefore(s, i)
+		if !ok {
 			continue
 		}
 		end, ok := linkLabelEnd(s, i)
@@ -162,7 +176,7 @@ func breakLinkReferenceDefinitions(s string) string {
 		if end+1 >= len(s) || s[end+1] != ':' {
 			continue
 		}
-		if !looksLikeLinkDestination(s[end+2:]) {
+		if !looksLikeLinkDestination(s[end+2:], container) {
 			continue
 		}
 		out.WriteString(s[prev : end+1])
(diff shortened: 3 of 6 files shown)


seacen:
Pushed a third case to this PR — same defect, louder surface.

`issueCreatedText` concatenates `res.IssueTitle` into markdown, and that confirmation is posted back into the chat that triggered the `/issue`. **In a group that means in front of the room**, carrying the bot's authority rather than the reporter's:

```
/issue 安全升级：请点击 [重置密码](https://evil.example) 完成验证
```

comes back from the bot as a working link. Worse than the inbox card, which at least only reaches one person's DM.

One handover note: **#6591** adds `issueDuplicateText`, and its title belongs to a **third party** rather than the reporter — so it needs the same `escapeInboxMarkdown` call. Whichever of the two lands second should add it; I'll do it in the rebase if that's #6591.

Bohan-J (changes requested):
Product call first: yes, we want this. A bot forwarding member-authored text should not lend its own authority to that text, and a notification the platform silently refuses while the sender is told "delivered" is a lost message, not a truncated one. Both halves belong in one PR.

Three things before merge, all small.

## 1. Confirm WeCom actually honors `\` escapes — needs evidence, not argument

The whole fix rests on the premise that aibot markdown treats `\[` as an escape. Nothing in the PR shows that it does. Both outcomes are *safe*, but they differ a lot on what the recipient sees:

- **If escapes are honored:** `\[Bug\] 登录失败` renders as `[Bug] 登录失败`. Perfect.
- **If they are not:** the recipient sees the backslashes. The security goal still holds — every `[ ] ( )` gets a backslash, so `](` can never end up adjacent in the output. I checked: `\[x\](u)` escapes to `\\[x\\]\(u\)`, which neither a CommonMark parser nor a naive `\[([^\]]+)\]\(([^)]+)\)` regex turns into a link. But then every bracketed title carries visible backslashes.

`[Bug] ...` and `[WIP] ...` are common title forms. If escapes aren't honored, we'd be trading a display regression every user sees for a defense against a rare case — and we should pick a mechanism that doesn't depend on escape support instead. Could you push one real `[Bug] 登录失败` notification through a live bot and attach the screenshot?

## 2. The body truncation path is missing `trimDanglingEscape`

`inbox_message.go`, the `room > 0` branch:

```go
if room > 0 {
    return prefix + "\n" + truncateRunes(body, room) + "..." + suffix
}
```

The title branch deliberately guards "cut inside an escape pair"; the body branch doesn't. Reproduced on this branch — title `t`, body `strings.Repeat("a", 3941) + "[" + strings.Repeat("b", 300)`, `MULTICA_APP_URL=https://multica.example`, slug `acme` — the output contains `\...`. Cosmetic only (the escape eats one dot of the ellipsis), but the helper exists for exactly this:

```go
return prefix + "\n" + trimDanglingEscape(truncateRunes(body, room)) + "..." + suffix
```

Worth a case alongside the two you already added.

## 3. `escapeInboxMarkdown` should escape `\` itself

To be clear up front: this is **not** a link-injection bypass, for the reason in §1 — `](` cannot be produced. Two other reasons to add it.

A title ending in a backslash breaks the card's bold markers:

```go
item := map[string]any{"type": "mentioned", "title": `修复路径 C:\`, "body": "some **body** text"}
// → "**[提及你] 修复路径 C:\\**\nsome **body** text"
```

`\*` consumes the first asterisk, the bold opener never closes, and the member-authored `**` in the body pairs with it instead.

More importantly: the escaping is currently safe because of a *global* property of the output ("`]` and `(` can never end up adjacent"), not because each character is locally neutralized. Whoever later decides parens alone are harmless and drops them from the set turns this into a real bypass. Adding `'\\'` to the first case makes it locally correct — and it makes `trimDanglingEscape`'s parity check exact, since then every backslash in the output is genuinely part of a pair.

## nit

`titleRoom` clamped to `0` still returns `fixed + 3 + len(suffix)` runes with no cap check. Unreachable in practice — it needs a link suffix over ~3987 runes — but one comment saying so saves the next reader the trip.

## Out of scope, but the sibling claim in the description is half right

Lark is indeed fine: the issue confirmation goes out as plain text, and the notice card uses a `plain_text` tag. The other two are not.

- `dingtalk/replier.go` — `issueCreatedText` / `issueDuplicateText` concatenate `res.IssueTitle` straight into the text, and `outbound_send.go` sends it as a markdown card.
- `slack/replier.go` — `issueCreatedText` / `issueDuplicateText` flow through `slackSender.Send` → `formatMrkdwn`, whose `reMdLink` rewrites `[text](url)` into Slack's `<url|label>`. That is a working link. "Slack builds blocks" doesn't hold for this path.

Please don't grow this PR to cover them — a separate issue is the right home.

## Verification

I confirmed all three new tests fail on `4a9d04ef6~1` and pass on the branch; `go test ./internal/integrations/wecom/ -count=1` and `go vet` are clean. Stating the cost in the assertion message ("card is 8053 runes, cap is 4000 — WeCom refuses the frame and the push is lost") is a good habit — please keep doing that.

seacen:
I pushed the notification through a live tenant. WeCom does not honor `\[` as an escape, and what it does instead is worse than the branch you priced.

## 1. The evidence

What went out on the wire:

```
**[Status changed] \[Bug\] 登录失败**
[View details](https://…)
```

`[Status changed]` is our own label and is never escaped. `\[Bug\]` is `escapeInboxMarkdown`'s output. Both are in the same line, which is what makes the pair of screenshots read.

**Conversation-list preview:**

WeCom conversation list preview

`[Status changed] \[Bug\] 登录失败 [View detail...`. This surface renders no markdown at all — our own `[View details](url)` link is flattened too — so the backslashes show raw. That is your predicted branch exactly.

**The card in the chat bubble:**

The card inside the chat

Compare the two bracketed groups on the same line. `[Status changed]` — ours, unescaped — keeps both brackets and renders in the card's own bold sans face. `\[Bug\]` — escaped — comes back as an **italic serif `Bug` with both brackets gone**.

That is not "the escape was ignored". `\[ ... \]` is being consumed as a **display-math delimiter** and `Bug` is being typeset as math. By the same logic `\( ... \)` is an inline-math delimiter, so the escaped parens are in it too.

**The math reading is inferred, not documented.** I am reading the mechanism off the rendering — italic serif with the delimiters eaten is what math typesetting looks like, and nothing else explains the brackets disappearing while the glyphs survive. No WeCom doc I can find describes markdown syntax at all, let alone this. The actionable conclusion does not depend on the name being right.

Two consequences that make this worse than visible backslashes, and both follow from the delimiter behaviour rather than from the label:

- Member text gets **swallowed into a math block and re-typeset**. A title stops looking like a title.
- An unpaired `[` in member text — `登录页 [ 崩了` — escapes to `\[`, which **opens a block that never closes**. One malformed title and the rest of the card is math.

So the answer to your §1 is: escapes are not honored, and "the recipient sees the backslashes" was the optimistic version.

## 2. The mechanism, replaced

`escapeInboxMarkdown` is gone. `breakLinkAdjacency` replaces it:

```go
func breakLinkAdjacency(s string) string {
	return strings.ReplaceAll(s, "](", "] (")
}
```

Leave the member's text exactly as written unless it contains `](`; where it does, put one plain space between them. It never emits a backslash, so it cannot enter math mode or leave a visible artifact. `[Bug] 登录失败` now goes out byte-identical on both surfaces, and so does `修复 (见 issue 12)` and `紧急!!!`.

The security property is the one you identified as the real guarantee, unchanged: `](` never appears in the output, so neither a CommonMark parser (which requires the link text to be followed *immediately* by `(`) nor a naive `\[([^\]]+)\]\(([^)]+)\)` rewriter can build a link. `x` needs the same adjacency, so images fall out of the same rule and `!` no longer needs to be in any set.

It moved to a new file, `markdown.go` — it serves the inbox card and the `/issue` confirmation now, and #6591 will make it three.

### What a plain space assumes, since it is a real character in someone's text

- **Where it lands.** Only between `]` and `(`. It can never begin a line, because a `]` always precedes it on the same line — no indented code block. It can never end one either, so no accidental hard break.
- **Whether the insertion itself is a lever.** The only one is volume: every `](` costs one extra rune, 1.5x in the limit. So the break runs **before** every length measurement in `buildInboxMarkdown`. Measured after truncation instead, a body of `](`×4000 ships 5971 runes against a 4000 cap — WeCom refuses the frame and the send path still reports success, which is the same silent-loss failure the other half of this PR fixes. Test and revert below.
- **The truncation seams**, which is what you were worried about. `truncateRunes` only drops a suffix, so it cannot recreate an adjacency that was already separated; and every literal spliced in after member text begins with `.`, `*` or `\n`, never `(`. Both facts are now written at the truncation site, and pinned by your two fixed seam cases plus a sweep over twelve cut offsets so the seam is covered at every position inside the pattern rather than at whichever one the fixed case happens to hit.
- **Display cost.** Zero on text without `](`. Where it fires, `[重置密码](https://evil.example)` becomes `[重置密码] (https://evil.example)` — the URL is visible plain text beside its label, which is a reasonable way to show a link we are refusing to make.

## 3. Your point 2 — the defect is gone with the mechanism

`trimDanglingEscape` and the "cut inside an escape pair" handling existed only because escaping produced pairs. Nothing produces a backslash now, so both are **deleted**, in the title branch as well as the missing body branch you found.

Your underlying point survives the deletion though — the body branch had no seam guard and the title branch did. That asymmetry is what `TestInboxCardSeamSurvivesEveryCutOffset` now covers for both branches at once.

## 4. Your point 3 — two of the three reasons are moot, one is not mine to fix here

Taking the reasons separately:

- **Local neutralization instead of a global adjacency argument.** Moot: there is no escape set left to narrow. The whole mechanism *is* the adjacency property now, stated in one line of code and one test, rather than an emergent consequence of five characters each getting a backslash. Your failure mode — someone decides parens are harmless and drops them — has nothing to drop.
- **Exact parity for `trimDanglingEscape`.** Moot with the function.
- **A title ending in `\` breaking the card's bold markers.** Still real, and now purely the member's own backslash meeting our `**`. But `title + "**"` is unchanged from `main` here — this PR neither creates nor worsens it, since we no longer touch backslashes at all. I would rather not fix it in this PR: every available fix either emits a backslash (which the screenshots just ruled out) or edits the member's text at a position they can control. And now that backslash is known to be load-bearing on this renderer, the question worth answering is not only whether `\*` escapes but what `修复路径 C:\` does on its own. That is one more tenant test and its own change.

## 5. nit — comment added at the clamp

Kept your reading that it is unreachable and wrote down why:

> Reaching it needs a link suffix past ~3987 runes, i.e. an appURL plus workspace slug of that length; both are operator- and DB-constrained, never member-authored. Left as-is deliberately: the alternative is dropping the link to make room, and no input that exists can ask for it.

## 6. One residual I want on the record

Breaking `](` closes inline links and images. It does not close a **CommonMark reference link** — `[重置密码]: http://evil.example` on its own line plus `[重置密码]` elsewhere reaches the same spoofed label with no `](` anywhere. It needs a line the member starts with `[`, so only a body can carry it (a title is always spliced after `**[提及你] `), and I have not verified whether this renderer resolves reference definitions at all. The old escaping blocked it incidentally rather than by design. I left it out because closing it means mangling `]:`, and `[Bug]: 登录失败` is a form real people write. Say the word if you want it in scope.

## 7. Siblings — you were right and the description was wrong

`dingtalk/replier.go` concatenates `res.IssueTitle` raw and `outbound_send.go` ships it as a markdown card. `slack/channel.go` runs the text through `formatMrkdwn`, and its `reMdLink` rewrites `[text](url)` into `<url|label>` — a working link, so "Slack builds blocks" does not cover that path. Lark is fine for the reason you gave. Not growing this PR; it needs its own issue.

## 8. The title-truncation half is unchanged

Untouched by the mechanism swap, still verified: dropping it puts the huge-title card at 8053 runes against a 4000 cap.

## Verification

`go build ./...`, `go vet`, `gofmt -l` clean. `go test -race ./internal/integrations/wecom/ -count=1` green.

Every test verified by reverting its fix and confirming it fails. Backslashes below are Go `%q` output, so `\\` is one backslash.

**A — `breakLinkAdjacency` removed from `buildInboxMarkdown` (the state before this PR):**

```
--- FAIL: TestInboxCardDoesNotRenderMemberAuthoredLinks (0.00s)
    inbox_message_test.go:148: member-authored link syntax rendered as a link in a bot-authored card: "**[提及你] [click here](http://evil.example)**\nand the body [too](http://evil.example)"
--- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen (0.00s)
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/link_in_title (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 25 — that is a working link in a card the bot signs: "click here](http://e"
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/link_in_body (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 39 — that is a working link in a card the bot signs: " body [too](http://e"
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/image (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 19 — that is a working link in a card the bot signs: "你] ![img](http://e"
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/nested_brackets (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 19 — that is a working link in a card the bot signs: "你] [a[b]](http://e"
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/member-written_backslash (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 16 — that is a working link in a card the bot signs: "及你] x\\](http://e"
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/body_truncated_at_the_seam (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 3918 — that is a working link in a card the bot signs: "aaaaaaaaaa](x)](x)]("
    --- FAIL: TestInboxCardNeverPutsCloseBracketNextToOpenParen/title_truncated_at_the_seam (0.00s)
        inbox_message_test.go:234: "](" came out adjacent at byte 20 — that is a working link in a card the bot signs: "\xbd\xa0] 标题](x)标题"
--- FAIL: TestInboxCardSeamSurvivesEveryCutOffset (0.00s)
    inbox_message_test.go:262: body cut at pad=0 fused a "]" onto a "(" at byte 3918: "aaaaaaaaaa](x)](x)]("
FAIL
```

**B — `breakLinkAdjacency` swapped back for the backslash escaper.** This is the one that matters most: it shows the new tests measure the *mechanism*, not just the absence of a link. Every failure below is the rendering in the screenshot.

```
--- FAIL: TestInboxCardKeepsAnOrdinaryBracketedTitleVerbatim (0.00s)
    inbox_message_test.go:191: member text did not survive verbatim:
         got "**[状态变更] \\[Bug\\] 登录失败**\n从 \\(todo\\) 到 \\(in_review\\)\\!"
        want "**[状态变更] [Bug] 登录失败**\n从 (todo) 到 (in_review)!"
--- FAIL: TestBreakLinkAdjacency (0.00s)
    --- FAIL: TestBreakLinkAdjacency/ordinary_bracketed_title_is_untouched (0.00s)
        markdown_test.go:34: breakLinkAdjacency("[Bug] 登录失败"):
             got "\\[Bug\\] 登录失败"
            want "[Bug] 登录失败"
    --- FAIL: TestBreakLinkAdjacency/parens_alone_are_untouched (0.00s)
        markdown_test.go:34: breakLinkAdjacency("修复 (见 issue 12)"):
             got "修复 \\(见 issue 12\\)"
            want "修复 (见 issue 12)"
    --- FAIL: TestBreakLinkAdjacency/a_lone_bracket_is_untouched (0.00s)
        markdown_test.go:34: breakLinkAdjacency("开了一个 [ 没关"):
             got "开了一个 \\[ 没关"
            want "开了一个 [ 没关"
    --- FAIL: TestBreakLinkAdjacency/exclamation_alone_is_untouched (0.00s)
        markdown_test.go:34: breakLinkAdjacency("紧急!!!"):
             got "紧急\\!\\!\\!"
            want "紧急!!!"
    --- FAIL: TestBreakLinkAdjacency/link (0.00s)
        markdown_test.go:34: breakLinkAdjacency("[click here](http://evil.example)"):
             got "\\[click here\\]\\(http://evil.example\\)"
            want "[click here] (http://evil.example)"
    --- FAIL: TestBreakLinkAdjacency/image (0.00s)
        markdown_test.go:34: breakLinkAdjacency("![img](http://evil.example/x.png)"):
             got "\\!\\[img\\]\\(http://evil.example/x.png\\)"
            want "![img] (http://evil.example/x.png)"
    --- FAIL: TestBreakLinkAdjacency/nested_brackets (0.00s)
        markdown_test.go:34: breakLinkAdjacency("[a[b]](http://evil.example)"):
             got "\\[a\\[b\\]\\]\\(http://evil.example\\)"
            want "[a[b]] (http://evil.example)"
    --- FAIL: TestBreakLinkAdjacency/back_to_back (0.00s)
        markdown_test.go:34: breakLinkAdjacency("](]("):
             got "\\]\\(\\]\\("
            want "] (] ("
    --- FAIL: TestBreakLinkAdjacency/two_links (0.00s)
        markdown_test.go:34: breakLinkAdjacency("[a](x) and [b](y)"):
             got "\\[a\\]\\(x\\) and \\[b\\]\\(y\\)"
            want "[a] (x) and [b] (y)"
    --- FAIL: TestBreakLinkAdjacency/backslash_in_front_of_the_pair_does_not_help_the_member (0.00s)
        markdown_test.go:34: breakLinkAdjacency("x\\](u)"):
             got "x\\\\]\\(u\\)"
            want "x\\] (u)"
--- FAIL: TestBreakLinkAdjacencyIsIdempotent (0.00s)
    markdown_test.go:53: second pass changed "[a](b)": "\\[a\\]\\(b\\)" → "\\\\[a\\\\]\\\\(b\\\\)"
--- FAIL: TestIssueConfirmationDoesNotRenderReporterLinks (0.00s)
    replier_test.go:248: a backslash reached the reply: "✅ 已创建 MUL-1 — 安全升级：请点击 \\[重置密码\\]\\(https://evil.example\\) 完成验证" — WeCom shows it raw in the list preview and reads it as a math delimiter in the bubble
--- FAIL: TestIssueConfirmationKeepsAnOrdinaryTitleVerbatim (0.00s)
    replier_test.go:262: the reporter's own title came back altered:
         got "✅ 已创建 MUL-1 — \\[Bug\\] 登录失败 \\(P0\\)\\!"
        want "✅ 已创建 MUL-1 — [Bug] 登录失败 (P0)!"
FAIL
```

**C — `breakLinkAdjacency` removed from `issueCreatedText`:**

```
--- FAIL: TestIssueConfirmationDoesNotRenderReporterLinks (0.00s)
    replier_test.go:242: a reporter-authored link rendered in a bot-authored group message: "✅ 已创建 MUL-1 — 安全升级：请点击 [重置密码](https://evil.example) 完成验证"
FAIL
```

**D — the body measured before the break and broken after truncation, i.e. the inserted spaces not budgeted:**

```
--- FAIL: TestInboxCardSeamSurvivesEveryCutOffset (0.00s)
    inbox_message_test.go:265: body cut at pad=0: card is 4021 runes, cap is 4000 — WeCom refuses the frame and the push is lost
--- FAIL: TestInboxCardBudgetsTheSpacesItInserts (0.00s)
    inbox_message_test.go:286: card is 5971 runes, cap is 4000 — the inserted spaces were not budgeted, so WeCom refuses the frame and the push is lost
FAIL
```

**E — title truncation dropped (the half of this PR that did not change):**

```
--- FAIL: TestInboxCardFitsTheCapEvenWithAHugeTitle (0.00s)
    inbox_message_test.go:164: card is 8053 runes, cap is 4000 — WeCom refuses the frame and the push is lost
FAIL
```

Bohan-J (changes requested):
The tenant test was the right call and it paid off — you didn't just answer §1, you disproved the premise the whole first approach rested on, and found something worse than the branch I priced. `\[Bug\]` coming back as italic serif with the delimiters eaten is a much better argument than "escapes are ignored", and an unpaired `[` opening a block that never closes would have been a genuinely bad bug to ship. Replacing five escaped characters with one adjacency break is a smaller mechanism with a property that is easier to state and easier to test, and running it before every length measurement (rather than discovering the 1.5x growth in production) is exactly right. §3, §5, §7 are all resolved.

Two things left.

## 1. Rebase — the branch no longer merges

`mergeStateStatus` is `DIRTY`. `server/internal/integrations/wecom/replier_test.go` conflicts with `main`: both sides appended a new test at the end of the file, and the three-way alignment splits across the shared closing braces. `replier.go` itself auto-merges and keeps `breakLinkAdjacency`.

I resolved it locally keep-both to check it isn't hiding anything: `gofmt -l` clean, `go build ./...` clean, `go test ./internal/integrations/wecom/ -count=1` green on the merged tree. So it is purely mechanical — but it does need doing before this can land.

Also worth knowing for the rebase: **#6591 has not landed on `main` yet**, so there is no `issueDuplicateText` in this package to cover. Your handover note still applies, just in the other direction now — whichever lands second adds the call.

## 2. One more tenant round-trip, for the new mechanism this time

This is the one thing I'd hold merge on. The screenshots establish that this renderer is *not* CommonMark — it consumed `\[ ... \]` as a math delimiter, which no CommonMark parser does. That is precisely why `markdown.go`'s justification can no longer lean on "CommonMark requires the link text to be followed *immediately* by `(`". The existing evidence validates the old output; it doesn't validate the new one, and it's the same undocumented renderer either way.

You already have the tenant wired up, so please send the **final helper output** for these three and confirm none of them produces a clickable label:

```text
[重置密码] (https://evil.example)
```

```text
[重置密码]: https://evil.example
[重置密码]
```

```text
<a href="https://evil.example">重置密码</a>
```

The first is the direct question: does a space between `]` and `(` actually stop this renderer, or does it accept whitespace there?

The second is the reference-link residual you recorded in §6. It contains no `](`, so `breakLinkAdjacency` passes it through byte-identical, and it reaches exactly the outcome this PR exists to prevent — a label that goes somewhere it does not say. Only a body can carry it, as you note, but issue descriptions are member-authored and multi-line, so that is not much of a constraint.

The third is mine: a renderer that implements math delimiters may well accept inline HTML, and an anchor tag is the same attack with no `](` and no reference definition. Costs one more message in a round-trip you are making anyway.

**On your "say the word" in §6** — my answer is *verify first, fix only if it renders*. I don't want `]:` mangled unconditionally; `[Bug]: 登录失败` is a form real people write, and trading that away on a guess is the same shape of mistake the escaping turned out to be. But if the renderer does resolve reference definitions, it has to be closed here rather than deferred, because at that point the PR's stated property simply isn't true for bodies.

If all three come back inert, the screenshots are sufficient merge evidence and I'll flip to approve. If any of them renders, neutralize that one and add the regression test.

## Agreed, not blocking

§4 — `修复路径 C:\` meeting our `**` is real but predates this PR, and every fix available inside this PR either emits a backslash (just ruled out) or edits member text at a position they control. Your point that backslash is now known to be load-bearing on this renderer is the right reason to give it its own tenant test and its own change. Separate issue.

§5 — the clamp comment says what I wanted it to say. Leaving the behaviour alone is correct.

§7 — please do open the DingTalk/Slack issue when you get a moment, or say the word and I'll file it.

## Verification

On `976965f7`: `go test ./internal/integrations/wecom/ -count=1` and `go vet` clean. I confirmed every member-authored path into a wecom markdown frame goes through the helper — `buildInboxMarkdown` (title + body, called from `outbound.go`) and `issueCreatedText` — with no third surface. Merged-tree checks as described in §1. CI green on the current head.

seacen:
Both open items are closed.

- **Rebase** — done. The branch is on `0ab271235`; the only conflict was the one you named, `replier_test.go`, one hunk, resolved keep-both.
- **Round-trip** — all three sent through a live tenant. `[重置密码] (https://evil.example)` is **inert**. `<a href="https://evil.example">重置密码</a>` is **inert**. `[重置密码]: https://evil.example` plus `[重置密码]` **renders**: the definition line disappears from the card and the label comes back as a blue underlined link.

So on your own instruction the third one is closed here rather than deferred — at that point the PR's stated property is not true for bodies. That is the new commit.

## 1. Rebase — done, and `main` has moved past what you were reading

Rebased onto `0ab271235`. `main` took #6590 (the outbound send verdict) after you wrote your note, so the base you were looking at is no longer the base.

The conflict was exactly where you said and nowhere else: `server/internal/integrations/wecom/replier_test.go`, both sides appending a test at the end of the file, one hunk, three-way alignment splitting on the shared closing braces. Resolved keep-both. `replier.go` auto-merged and kept the helper, as you found. `gofmt -l`, `go build ./...`, `go vet ./...` and the package tests are all clean on the result; the new commits sit on top of it. `mergeStateStatus` is `CLEAN`.

Your #6591 note still stands in the direction you turned it: there is still no `issueDuplicateText` in this package, so whichever of the two lands second adds the call.

## 2. The round-trip — two inert, one renders

### Method, because it changes how you read the first image

The three probes went out in **one message**. A link reference definition is document-scoped, so the definition in block 2 supplied a target for `[重置密码]` *anywhere in that document* — including the label in block 1. In the three-probe screenshot block 1's label therefore appears linked, and that appearance says nothing about the space.

That is why block 1 was re-run on its own. The isolated run is the one that establishes the space works; the three-probe image establishes blocks 2 and 3.

### The three probes

Three probes in one message

**Block 2 — `[重置密码]: https://evil.example` + `[重置密码]` — renders.** The definition line is *gone from the card* — consumed, not displayed — and the bare label below it comes back blue and underlined. This renderer resolves CommonMark link reference definitions. This is your §6 residual, live, and it is the outcome the PR exists to prevent: a label that goes somewhere it does not say, with no `](` anywhere for `breakLinkAdjacency` to catch. It passed through byte-identical.

**Block 3 — `<a href="https://evil.example">重置密码</a>` — inert.** Plain black text, tag and all. Your guess that a renderer implementing math delimiters might also take inline HTML was worth the one extra message, but this one does not. Nothing to do, and nothing added for it.

### Block 1, on its own

The space, sent in isolation

**`[重置密码] (https://evil.example)` — inert.** Plain text, both brackets intact, and `(https://evil.example)` is not consumed as a destination — it sits there as visible text next to its label, which is what you would see if the inline link failed to form. So the space does stop this renderer, and `markdown.go`'s justification no longer has to lean on what CommonMark requires. That observation is now written into the comment in place of the inference.

## 3. What the new guard does

`breakLinkReferenceDefinitions` separates the `]` from the `:` that follows it — one plain space, same shape of edit as the adjacency break — but only when all three of these hold:

1. **Only block scaffolding precedes the opening `[` on its line** — indentation, `>` blockquote markers, list bullets. This is not decoration. A definition written inside a blockquote or a list item still populates the whole document's reference map and resolves outside the container; verified, three shapes.
2. **The brackets form a CommonMark link label** — closed by the first unescaped `]`, no unescaped `[` inside, at least one non-whitespace character, at most 999. The label may run across a line ending and still resolve, so the scan does too.
3. **What follows the colon is plausibly a link destination** — skipping whitespace, at most one line ending (a definition may span two lines), and an optional `<` or `(`. The token then qualifies if it carries a URI scheme, or starts with `//`, or uses an escape mechanism.

Condition 3 is the whole design, and it is your constraint made mechanical. Breaking every `]:` unconditionally is simpler and is the trade you told me not to make; `[Bug]: 登录失败` fails condition 3 and comes back byte-identical.

### Where the break goes, and why not anywhere else

Between the `]` and the `:`. CommonMark requires the colon to follow the label immediately, so one space ends the definition — checked against a parser, including inside a blockquote and with the destination on the next line.

It cannot go inside the label: reference labels are matched on whitespace-collapsed, case-folded text, so `[ 重置密码]` still matches `[重置密码]` and the definition would survive the edit. It cannot be a backslash, for the reason your screenshots established. And it cannot go before the `[`: there is no character that pushes a line out of block position without either showing up as content or, four spaces in, opening a code block.

### The escaped and entity-encoded destinations, which are why a literal scheme check is not enough

All four of these parse to a working external URL:

```text
[重置密码]: https\://evil.example
[重置密码]: \/\/evil.example/reset
[重置密码]: &#x68;ttps://evil.example
[重置密码]: &#104;ttps://evil.example
```

Backslash escapes and character references are recognised *inside* a link destination, so `https\:` decodes to `https:` and `\/\/` decodes to `//`. A guard that looks for a literal `https:` or a literal `//` sees none of the four and lets all four through.

I chose to treat **"an escape mechanism was used"** as the signal rather than decode the destination first. Two reasons. Decoding means carrying a copy of CommonMark's backslash-escape set and its entity table inside a security guard, and a subtle gap in either reopens the hole silently — the failure mode is invisible, which is the worst kind here. And the table would be a guess anyway: this renderer is demonstrably not CommonMark, so decoding *to CommonMark's rules* is not obviously the decoding that matters.

The cost of the cruder signal is that the first token after `]:` containing a `\` or an `&` fires the break. In prose that token is a word; words do not contain backslashes or ampersands. `[Bug]: 登录失败` is unaffected, and so is every other real title I can construct.

### Housekeeping

Both breaks now go through one entry point, `breakMemberLinks`, so a call site cannot take one and forget the other. The three surfaces you enumerated in your verification — `buildInboxMarkdown` title, `buildInboxMarkdown` body, `issueCreatedText` — all moved to it, and there is still no fourth.

The space is inserted before every length measurement, same as the adjacency break, with its own budget test: a body that is nothing but definitions grows one rune per line.

One implementation note you would spot anyway. Finding the line start with `strings.LastIndexByte` per candidate `[` is O(n²) on a body full of brackets, and an issue description is member-authored and unbounded at this point in the code. It walks backwards from the `[` instead and stops at the first byte that cannot be scaffolding, which settles a bracket in the middle of prose after reading one byte. 200k of adversarial input runs in about 1.2ms.

## 4. What it deliberately lets through — this is a choice, and you should be able to disagree with it

**Relative destinations stay valid definitions.** `[Bug]: 登录失败`, `[文件]: report.pdf`, `[页面]: /inbox` all still define a reference, and the label can still come back as a link. The `href` is client-relative — it resolves against whatever base the client itself is on — so it cannot reach another host. What it cannot do is the thing this PR is about: point the reader at somebody else's server behind a label that says something different.

That is the price of `[Bug]: 登录失败` surviving byte-identical, which is the trade you asked for. I am stating it as a decision, not as a derivation, because the other side of it is real: a label can still be turned into a dead or self-referential link, and the member picks the label text.

The next notch up would be to treat a bare host shape — `evil.example/reset`, no scheme — as a destination too, which would close the case where this renderer prepends a scheme to a bare host the way it linkifies bare URLs in text. I left it out because we have no evidence it does that, and the notch costs `[文件]: report.pdf` and `[图]: diagram.png`. If you would rather pay that way round, say so and I will widen it — it is two lines.

**Bare URLs are unchanged and still out of scope**, for the reason already in the file: a bare URL displays its own destination, so it cannot lie. Worth restating because it shows up in the test: after the break, `[重置密码] : <https://evil.example>` renders the URL as an autolink whose visible text *is* the URL. The label beside it is inert text. That is the accepted case, not a miss.

**`[foo]:(https://evil.example)`** — CommonMark keeps the parens inside the `href`, so it is not a URL anyone navigates to. A leading `(` is skipped before the destination is examined anyway, in case this renderer trims them; skipping one character costs nothing.

## 5. Where it over-fires, on purpose

Two places, both documented at the function:

- **It does not model block context.** A link reference definition cannot interrupt a paragraph, so CommonMark would fold `[x]: https://evil.example` into the line above it and render it as prose. This fires on it anyway. Tracking block context properly means a real block parser — containers, lazy continuation, HTML blocks — inside a guard whose whole value is being small enough to read.
- **It does not check "nothing may follow the destination".** `[x]: https://evil.example 请点击` is not a definition under CommonMark; it is broken here regardless.

Both cost one space on a line that already carries a URL. Under-firing costs a working link under the bot's name. The asymmetry decides the direction, and it decides it harder given that this renderer reads `\[ … \]` as a math delimiter — being spec-exact about what *is not* a definition would be trusting a spec this renderer has already been shown not to follow.

## 6. Shortcut, collapsed and full references, and why cross-message is not a vector

All three usage forms resolve through the same reference map, so all three die with the definition and none of them needed separate handling: `[label]`, `[label][]` and `[text][label]` are each in the corpus, and each fails when the guard is reverted.

Cross-message I checked in the code rather than assuming it. `buildInboxMarkdown`'s result becomes the `markdown.content` of exactly one `aibot_send_msg` frame — `outbound.go:224` → `sendTextCtx` → `sendMsgTextBody` — and nothing batches two inbox items into one frame. So the definition and the label that uses it have to be inside a single member's title or body, which is what the tenant showed: one comment body carried both. That also means the guard can run per field rather than over the finished card, which is what the length budget needs.

## 7. Test method — the answer to "will these still mean anything in a year"

Two things, both deliberate.

**The assertions are made with a real CommonMark parser, not with my own detector.** `github.com/yuin/goldmark` is already a direct dependency of `server/` and already used in-tree for exactly this shape of job (`cmd/multica/local_path_link_lint.go` walks a parsed AST for link destinations). The tests parse the *hardened* output and collect `Link` and `Image` destinations off the AST. Testing a detector with the same detector would prove only that it is self-consistent.

`AutoLink` is excluded from that collection on purpose, and the reason is the §4 carve-out: an autolink's visible text is its own URL, so it cannot misrepresent where it goes. What must never appear is a label reading 重置密码 that quietly resolves elsewhere, and that is a `Link`, never an `AutoLink`.

**Every attack sample self-checks that it resolves when unhardened.** Each of the 22 corpus entries asserts, first, that the *raw* string produces an `evil.example` destination — and fails if it does not. A later change that stops the guard firing cannot pass by quietly shrinking the corpus into strings that were never dangerous in the first place; the corpus has to keep proving it is a corpus.

## 8. Your §7 — the sibling issue is filed

https://github.com/multica-ai/multica/issues/6622

Slack is stated there as a code fact: `formatMrkdwn`'s `reMdLink` in `mrkdwn.go` rewrites `[text](url)` into `<url|label>`, i.e. it builds the link itself, so "Slack builds blocks" does not cover that path. DingTalk is stated as code-path-confirmed but rendering-unverified — `dingtalk/replier.go` concatenates `res.IssueTitle` raw and `outbound_send.go` ships it as a markdown card, but we have no DingTalk tenant, so nobody has watched one render. Lark is fine for the reason you gave.

Nobody is on it.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l` clean on every file this branch touches. `go test -race ./internal/integrations/wecom/ -count=1` green against a migrated Postgres, so the DB-backed tests in the package run rather than skip.

Both directions of the new guard were verified by breaking it and watching the tests fail. Backslashes below are Go `%q` output, so `\\` is one backslash.

**A — `breakLinkReferenceDefinitions` reverted to a no-op, i.e. the state this PR was in when you reviewed it:**

```
--- FAIL: TestInboxCardSeamKeepsDefinitionsBrokenAtEveryCutOffset (0.00s)
    inbox_message_test.go:386: body cut at pad=0 let a member-defined link through: resolves [https://evil.example https://evil.example]
--- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition (0.00s)
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/plain (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https://evil.example\n\n[重置密码]" → "[重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/no_space_after_the_colon (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]:https://evil.example\n\n[重置密码]" → "[重置密码]:https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/indented_three_spaces (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "   [重置密码]: https://evil.example\n\n[重置密码]" → "   [重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/angle_bracketed (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: <https://evil.example>\n\n[重置密码]" → "[重置密码]: <https://evil.example>\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/destination_on_the_next_line (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]:\nhttps://evil.example\n\n[重置密码]" → "[重置密码]:\nhttps://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/with_a_title (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https://evil.example \"点这里\"\n\n[重置密码]" → "[重置密码]: https://evil.example \"点这里\"\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/title_on_the_next_line (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https://evil.example\n\"点这里\"\n\n[重置密码]" → "[重置密码]: https://evil.example\n\"点这里\"\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/scheme_relative (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: //evil.example/reset\n\n[重置密码]" → "[重置密码]: //evil.example/reset\n\n[重置密码]" resolves [//evil.example/reset]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/escaped_scheme_colon (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https\\://evil.example\n\n[重置密码]" → "[重置密码]: https\\://evil.example\n\n[重置密码]" resolves [https\://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/escaped_slashes (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: \\/\\/evil.example/reset\n\n[重置密码]" → "[重置密码]: \\/\\/evil.example/reset\n\n[重置密码]" resolves [\/\/evil.example/reset]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/hex_character_reference (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: &#x68;ttps://evil.example\n\n[重置密码]" → "[重置密码]: &#x68;ttps://evil.example\n\n[重置密码]" resolves [&#x68;ttps://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/decimal_character_reference (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: &#104;ttps://evil.example\n\n[重置密码]" → "[重置密码]: &#104;ttps://evil.example\n\n[重置密码]" resolves [&#104;ttps://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/inside_a_blockquote (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "> [重置密码]: https://evil.example\n\n[重置密码]" → "> [重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/inside_a_list_item (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "- [重置密码]: https://evil.example\n\n[重置密码]" → "- [重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/inside_an_ordered_list_item (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "1. [重置密码]: https://evil.example\n\n[重置密码]" → "1. [重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/escaped_bracket_in_the_label (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置\\]密码]: https://evil.example\n\n[重置\\]密码]" → "[重置\\]密码]: https://evil.example\n\n[重置\\]密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/label_across_two_lines (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置\n密码]: https://evil.example\n\n[重置\n密码]" → "[重置\n密码]: https://evil.example\n\n[重置\n密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/case_folded_label (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[Reset]: https://evil.example\n\n[reset]" → "[Reset]: https://evil.example\n\n[reset]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/collapsed_reference (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https://evil.example\n\n[重置密码][]" → "[重置密码]: https://evil.example\n\n[重置密码][]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/full_reference (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: https://evil.example\n\n[点这里][重置密码]" → "[重置密码]: https://evil.example\n\n[点这里][重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/two_definitions (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[a]: https://evil.example\n[b]: https://evil.example/2\n\n[a] [b]" → "[a]: https://evil.example\n[b]: https://evil.example/2\n\n[a] [b]" resolves [https://evil.example https://evil.example/2]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/uppercase_scheme (0.00s)
        markdown_test.go:116: a member-defined link survived the guard: "[重置密码]: HTTPS://EVIL.EXAMPLE\n\n[重置密码]" → "[重置密码]: HTTPS://EVIL.EXAMPLE\n\n[重置密码]" resolves [HTTPS://EVIL.EXAMPLE]
--- FAIL: TestIssueConfirmationDefinesNoLinkReference (0.00s)
    replier_test.go:399: a reporter-defined link resolved in a bot-authored group message: "✅ 已创建 MUL-1 — 安全升级\n\n[重置密码]: https://evil.example\n\n[重置密码]" resolves [https://evil.example]
FAIL
```

The card-level test, `TestInboxCardDefinesNoResolvableLinkReference`, fails on all 22 the same way through `buildInboxMarkdown` — omitted here only because it is the same list twice.

**B — the guard made unconditional, i.e. every `]:` broken, which is the version you told me not to ship:**

```
--- FAIL: TestInboxCardKeepsAReferenceLikeTitleVerbatim (0.00s)
    inbox_message_test.go:360: member text did not survive verbatim:
         got "**[状态变更] [Bug] : 登录失败**\n[WIP] : 明天再看\n\n[复现步骤] : 见 (todo) 到 (in_review)!"
        want "**[状态变更] [Bug]: 登录失败**\n[WIP]: 明天再看\n\n[复现步骤]: 见 (todo) 到 (in_review)!"
--- FAIL: TestBreakLinkReferenceDefinitionsLeavesProseAlone (0.00s)
    markdown_test.go:146: member text was altered:
         got "[Bug] : 登录失败"
        want "[Bug]: 登录失败" (unchanged)
--- FAIL: TestIssueConfirmationKeepsAnOrdinaryTitleVerbatim (0.00s)
    replier_test.go:418: the reporter's own title came back altered:
         got "✅ 已创建 MUL-1 — [Bug] : 登录失败"
        want "✅ 已创建 MUL-1 — [Bug]: 登录失败"
FAIL
```

CI is green on the current head `0e45cc0e5` — all nine checks, run 31239492586, including `backend` (5m10s) and `windows-execenv` (1m45s).

Bohan-J (changes requested):
Both of last round's blockers are closed: the branch is `CLEAN`, and the tenant round-trip answered all three payloads — the space holds, inline HTML is inert, and the reference link renders. Bringing the definition into scope was the right call once you had that answer, and condition 3 is the right shape for it: breaking every `]:` would have been the same kind of blind trade that the escaping turned out to be, and `[Bug]: 登录失败` surviving byte-identical is worth the extra machinery. Verifying against a real parser's AST destinations rather than string-matching the output is also the right way to test this.

One blocker in the new guard, and one nit.

## must-fix — a definition inside a blockquote with its destination on the next line passes through untouched

`looksLikeLinkDestination` allows one line ending after the colon and then skips spaces and tabs, but it does not replay the block container prefix that `blockScaffoldOnlyBefore` just accepted on the definition line. So the destination scan lands on `>`, decides it carries no scheme, no `//` and no `\`/`&`, and returns false. `breakLinkReferenceDefinitions` then returns the input byte-identical while the definition still resolves.

Verified on `0e45cc0e` with the PR's own goldmark checker (`markdownDestinations` / `hasDestinationTo`). Every row below comes back **byte-identical** from `breakMemberLinks` and still yields a destination of `evil.example`:

| shape | body |
|---|---|
| blockquote | `> [重置密码]:` / `> https://evil.example` |
| no space after `>` | `>[重置密码]:` / `>https://evil.example` |
| tab after `>` | `>\t[重置密码]:` / `>\thttps://evil.example` |
| CRLF | `> [重置密码]:\r\n> https://evil.example` |
| nested | `> > [重置密码]:` / `> > https://evil.example` (and three levels) |
| list inside blockquote | `> - [重置密码]:` / `>   https://evil.example` |
| bracketed destination | `> [重置密码]:` / `> <https://evil.example>` |
| scheme-relative destination | `> [重置密码]:` / `> //evil.example/x` |

It is also not confined to the shortcut form — the same definition resolves through **`[重置密码][]`**, **`[点这里][重置密码]`** and **`![重置密码]`**, so whatever lands should pin all four usages, not just `[label]`.

Worth saying what *is* handled, because it keeps the fix narrow — these are all correctly broken today: plain next-line destination, blockquote with a same-line destination, list continuation (`- [x]:` / `  https://…`), lazy continuation (`> [x]:` / `https://…` with no marker), a label spanning two blockquote lines, and `[x]:https://…` with no space after the colon.

Two shapes that goldmark does **not** resolve, so they need no coverage either way: `> [x]:` / `> > https://…` (deeper on the destination line) and `> [x]:` / `>` / `> https://…` (blank marker line between).

For the fix, carry the container prefix `blockScaffoldOnlyBefore` consumed and require the continuation line to match it, rather than unconditionally eating any run of `>`. Over-firing stays the safe direction, so an approximate match is fine — but comparing depth keeps the intent legible, and it is the difference between "we modelled the container" and "we swallowed a character that happened to be in the way". Then add the eight shapes above plus the three extra usage forms to `linkReferenceAttacks`.

## nit — the `\`/`&` heuristic rewrites ordinary relative destinations

The fallback treats any `\` or `&` in the destination as evidence of encoded off-host machinery. In practice it fires on the **first whitespace-delimited token only**, which makes the behaviour hard to predict from outside:

```text
[Owner]: R&D          → [Owner] : R&D
[Regex]: \d+          → [Regex] : \d+
[Path]: docs\setup    → [Path] : docs\setup
[Query]: foo&bar      → [Query] : foo&bar

[Note]: see A&B for details   → unchanged   (the "&" is in a later token)
```

All four rewritten lines are relative destinations, which the file's own "Relative destinations … Paid knowingly to keep the human form intact" bullet promises to leave alone. It costs one space on a line the member wrote, so I would not hold merge on it — but the contract and the code should agree. Either decode the escapes and character references before deciding, or narrow the trigger to a syntactically valid escape (`\` followed by ASCII punctuation) or character reference (`&`…`;`), and put those four lines in the fidelity table next to `[Bug]: 登录失败`.

## Agreed / no action

`TestInboxCardBudgetsTheSpacesTheDefinitionBreakInserts` covers the second break's growth, so the "measure after breaking" property still holds for both inserts. The over-firing section is honest about where the approximation runs and argues the right direction. `修复路径 C:\` remains out of scope here for the reasons we settled last round.

## Verification

On `0e45cc0e`: `go test ./internal/integrations/wecom/ -count=1` and `go vet` clean; CI green. The bypass table and the fidelity table above were both produced against this head using the PR's own goldmark helpers; the working tree was clean afterwards. Fix the blockquote continuation and I'll approve.

seacen:
The blocker is closed, and closed by making the two halves share one value rather than by adding a `>` case.

`blockScaffoldOnlyBefore` returned a bool. That is the whole bug in one line: it worked out which containers held the label and then threw the answer away, so by the time `looksLikeLinkDestination` ran there was nothing left for it to replay even if it had wanted to. It now returns the containers it found, `looksLikeLinkDestination` takes them, and the continuation scan steps over the same markers the label scan accepted. `TestContainerPrefixHalvesAgree` enumerates prefixes from the first half and requires the second to handle every one, so the next container we teach one half about fails the build rather than shipping.

Your eight shapes plus the three usage forms are in `linkReferenceAttacks`, along with seven more I found while sweeping. The nit is taken: the trigger is now a backslash before ASCII punctuation or an `&…;`, and your four lines are in the fidelity table.

`go build ./...`, `go vet ./...`, `gofmt -l` clean; `go test -race ./internal/integrations/wecom/ -count=1` green. CI green on `20503c831` — all nine checks, run 31256721604, including `backend` (5m22s) and `windows-execenv` (2m4s).

## 1. What landed, and why it is not a `>` case

Three changes, and the third is the one that closes the class:

- `blockScaffoldOnlyBefore` → `containerPrefixBefore`, returning `(containerPrefix, bool)`. The scaffolding parse moved into `parseContainerPrefix`, which is now the only place a run of scaffolding is read.
- `looksLikeLinkDestination` takes that `containerPrefix` as a parameter. Where it used to call `skipSpacesTabs` after the line ending it now calls `skipContinuationPrefix`.
- `containerPrefix` is a value rather than a bool so that there is something to hand over. That is the fix; the `>` handling falls out of it.

`skipContinuationPrefix` consumes **at most** `prefix.quotes` markers. Fewer, because a blockquote takes lazy continuations — `> [x]:` then a bare `https://…` with no marker still defines the reference, and that shape is one you listed as already correct, so it had to stay correct. Never more, because a line nested deeper than the definition's own block opens a new one rather than continuing it; the scan stops and lets the leftover `>` be what it is, which is not a destination. Both directions are asserted.

### What else `blockScaffoldOnlyBefore` accepts

The full list, since the answer to that is what decides whether the fix generalises. It accepts, interleaved in any order and any number of times:

- indentation — spaces and tabs, any amount, including past four
- `>` markers, any depth, with or without a following space or tab
- bullet list markers `-`, `+`, `*` followed by a space or tab
- ordered list markers — up to nine digits, then `.` or `)`, then a space or tab

Of those, only the blockquote marker needed replaying, and that is not luck. A blockquote repeats its marker on every line it holds, so a destination on the next line arrives as `> https://…`. A list item continues by indentation instead, and the destination scan already stepped over indentation as whitespace — which is why `- [x]:` / `  https://…` worked before this change and still works.

I did not want to take that on reading. Sweeping the scaffolding alphabet against itself with the fix reverted gives **647 distinct (definition-prefix, destination-prefix) pairs** that come back byte-identical and still resolve, and every single one carries a `>` on both lines — none on the definition line without one, none on the destination line without one. So the class is exactly the blockquote family, and the shared value closes all of it.

## 2. The test for the disagreement, not for its symptoms

`TestContainerPrefixHalvesAgree` is the one you asked for. It enumerates scaffolding prefixes three pieces deep, keeps the ones `containerPrefixBefore` accepts, and for each requires `looksLikeLinkDestination` to accept a next-line destination at every nesting from a lazy zero up to the definition's own depth, in five marker spellings (`>`, `> `, `>\t`, ` > `, `  >`). Then it requires the opposite at one level deeper: stepping over that marker would be swallowing a character that happened to be in the way, which is the distinction you drew.

The expected marker count is taken from the prefix text with `strings.Count(prefix, ">")`, not read off `containerPrefixBefore`'s answer. Otherwise the two halves would be checked against each other rather than against the markdown rule, and a shared wrong idea would pass.

Against the reverted fix it fails on the first accepted prefix whose container the scan misses:

```
--- FAIL: TestContainerPrefixHalvesAgree (0.00s)
    markdown_test.go:286: prefix "  >" may hold a definition, but the destination scan misses "\n>https://evil.example" —
        the two halves disagree about what a block container is
```

Alongside it, `TestBreakLinkReferenceDefinitionsAcrossBlockScaffolding` sweeps 88 definition-line prefixes against the same 88 on the destination line, four destination spellings, both line endings, and asserts against goldmark rather than against a list: 9066 of those combinations define a link to the attacker's host unguarded, and every one must come back defining none. The live count has a floor, so the sweep cannot pass by testing nothing — same self-checking property as the corpus, applied to a set nobody enumerated by hand.

### One thing the sweep taught me about the oracle

`hasDestinationTo` matches a substring. That is right for hand-written attack bodies, where the destination *is* the URL. It is wrong for a machine sweep, and it took a false positive to notice: a tab before a `>` on the destination line makes goldmark read the destination as `>https://evil.example`. It contains the host name and goes nowhere — the leading `>` makes it relative, so the client resolves it against its own base.

The sweep therefore uses `sendsReaderTo`, which parses the destination and compares the host. Otherwise it would have been demanding breaks that the relative-destination carve-out promises not to make. Those shapes are absent from the corpus for the same reason: they are the carve-out working, not a miss.

## 3. The corpus

Nineteen new entries, 41 in total, each still asserting it resolves unguarded before it asserts anything about the guard.

Your eight shapes are nine rows, since "and three levels" earns its own. Plus your three usage forms, plus seven found while sweeping:

- `>  >  [x]:` — markers with spaces around them rather than a single space after
- `> 1. [x]:` / `>    https://…` — ordered list inside a blockquote
- blockquote continuation with `https\://evil.example`, and with `&#x68;ttps://evil.example` — the escape spellings you listed for the same-line form, which needed pinning for the continuation form too
- the definition used **inside** the blockquote (`>\n> [重置密码]`) rather than after it
- and the two you listed as already correct, `> [x]:` with a lazy bare `https://…` and `- [x]:` / `  https://…`, added as regression pins. The "at most, never more" rule is exactly what keeps those two working, so they belong in the corpus rather than in a comment.

## 4. The nit

Taken. `strings.ContainsAny(dest, "\\&")` is now `hasBackslashEscape(dest) || hasCharacterReference(dest)`: a backslash before ASCII punctuation, or an `&` with a name-or-numeric body closed by `;`. Your four lines and `[Note]: see A&B for details` are in `TestBreakLinkReferenceDefinitionsLeavesProseAlone` next to `[Bug]: 登录失败`, and the contract bullet now reads `"[文件]: report.pdf"`, `"[页面]: /inbox"`, `"[Owner]: R&D"`, `"[Regex]: \d+"`, `"[Bug]: 登录失败"`.

Narrowing rather than decoding, for the reason already in the file: carrying CommonMark's escape set and HTML5's entity table inside the guard puts the failure mode out of sight, and a gap in either reopens the hole silently. Narrowing keeps the attack side intact, because `\:`, `\/`, `&#x68;` and `&#104;` are all real machinery — all four encoded destinations in the corpus still fire, both the same-line and the new continuation forms.

It does not check the entity name against HTML5's table, so `&B;` fires. One space on a line nobody writes, and the same side of the call the rest of the file already runs on.

## 5. On the budget

Unchanged and still correct. The guard now fires in more places, but it still inserts one space per definition and `breakMemberLinks` still runs before anything measures, so `TestInboxCardBudgetsTheSpacesTheDefinitionBreakInserts` covers the new shapes without changes. The nit narrows the trigger, so it can only fire less.

## Verification

`go build ./...`, `go vet ./...` clean. `gofmt -l` clean on both files this commit touches — the 27 files it does flag are flagged identically on `0e45cc0e5`, so nothing here added to them. `go test -race ./internal/integrations/wecom/ -count=1` green, 12.2s, against a throwaway migrated Postgres: 102 top-level tests, **zero skips**, so the ten DB-backed `TestUpsert_*` cases ran rather than skipped.

Both directions were verified by breaking the fix and watching the tests fail. Backslashes are Go `%q` output, so `\\` is one backslash.

### A — the container replay reverted

`skipContinuationPrefix` put back to `skipSpacesTabs`, which is `0e45cc0e5`'s behaviour on this path and nothing else changed.

Every blockquote-continuation shape comes back **byte-identical** and still resolves, through the PR's own `markdownDestinations` / `hasDestinationTo`:

```
shape                                                byte-identical   resolves   destination(s) after the guard
plain                                                false            false      []
no space after the colon                             false            false      []
indented three spaces                                false            false      []
angle bracketed                                      false            false      []
destination on the next line                         false            false      []
with a title                                         false            false      []
title on the next line                               false            false      []
scheme relative                                      false            false      []
escaped scheme colon                                 false            false      []
escaped slashes                                      false            false      []
hex character reference                              false            false      []
decimal character reference                          false            false      []
inside a blockquote                                  false            false      []
inside a list item                                   false            false      []
inside an ordered list item                          false            false      []
escaped bracket in the label                         false            false      []
label across two lines                               false            false      []
case folded label                                    false            false      []
collapsed reference                                  false            false      []
full reference                                       false            false      []
two definitions                                      false            false      []
uppercase scheme                                     false            false      []
blockquote continuation                              true             true       ["https://evil.example"]
blockquote continuation, no space after the marker   true             true       ["https://evil.example"]
blockquote continuation, tab after the marker        true             true       ["https://evil.example"]
blockquote continuation, CRLF                        true             true       ["https://evil.example"]
blockquote continuation, nested twice                true             true       ["https://evil.example"]
blockquote continuation, nested three deep           true             true       ["https://evil.example"]
blockquote continuation, indented markers            true             true       ["https://evil.example"]
list inside a blockquote, continued by indentation   true             true       ["https://evil.example"]
ordered list inside a blockquote                     true             true       ["https://evil.example"]
blockquote continuation, angle bracketed             true             true       ["https://evil.example"]
blockquote continuation, scheme relative             true             true       ["//evil.example/x"]
blockquote continuation, escaped scheme colon        true             true       ["https\\://evil.example"]
blockquote continuation, character reference         true             true       ["&#x68;ttps://evil.example"]
blockquote lazy continuation, marker dropped         false            false      []
list continuation by indentation                     false            false      []
blockquote continuation, collapsed reference         true             true       ["https://evil.example"]
blockquote continuation, full reference              true             true       ["https://evil.example"]
blockquote continuation, image reference             true             true       ["https://evil.example"]
blockquote continuation, used inside the blockquote  true             true       ["https://evil.example"]
```

The last two `false` rows are the shapes you listed as already handled. They stay broken under the revert, which is what makes the nineteen `true` rows above them a class and not the whole file.

The corpus test, on the same revert:

```
--- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition (0.00s)
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https://evil.example\n\n[重置密码]" → "> [重置密码]:\n> https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_no_space_after_the_marker (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: ">[重置密码]:\n>https://evil.example\n\n[重置密码]" → ">[重置密码]:\n>https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_tab_after_the_marker (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: ">\t[重置密码]:\n>\thttps://evil.example\n\n[重置密码]" → ">\t[重置密码]:\n>\thttps://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_CRLF (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\r\n> https://evil.example\r\n\r\n[重置密码]" → "> [重置密码]:\r\n> https://evil.example\r\n\r\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_nested_twice (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> > [重置密码]:\n> > https://evil.example\n\n[重置密码]" → "> > [重置密码]:\n> > https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_nested_three_deep (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> > > [重置密码]:\n> > > https://evil.example\n\n[重置密码]" → "> > > [重置密码]:\n> > > https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_indented_markers (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "  >  >  [重置密码]:\n  >  >  https://evil.example\n\n[重置密码]" → "  >  >  [重置密码]:\n  >  >  https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/list_inside_a_blockquote,_continued_by_indentation (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> - [重置密码]:\n>   https://evil.example\n\n[重置密码]" → "> - [重置密码]:\n>   https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/ordered_list_inside_a_blockquote (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> 1. [重置密码]:\n>    https://evil.example\n\n[重置密码]" → "> 1. [重置密码]:\n>    https://evil.example\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_angle_bracketed (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> <https://evil.example>\n\n[重置密码]" → "> [重置密码]:\n> <https://evil.example>\n\n[重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_scheme_relative (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> //evil.example/x\n\n[重置密码]" → "> [重置密码]:\n> //evil.example/x\n\n[重置密码]" resolves [//evil.example/x]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_escaped_scheme_colon (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https\\://evil.example\n\n[重置密码]" → "> [重置密码]:\n> https\\://evil.example\n\n[重置密码]" resolves [https\://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_character_reference (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> &#x68;ttps://evil.example\n\n[重置密码]" → "> [重置密码]:\n> &#x68;ttps://evil.example\n\n[重置密码]" resolves [&#x68;ttps://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_collapsed_reference (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https://evil.example\n\n[重置密码][]" → "> [重置密码]:\n> https://evil.example\n\n[重置密码][]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_full_reference (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https://evil.example\n\n[点这里][重置密码]" → "> [重置密码]:\n> https://evil.example\n\n[点这里][重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_image_reference (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https://evil.example\n\n![重置密码]" → "> [重置密码]:\n> https://evil.example\n\n![重置密码]" resolves [https://evil.example]
    --- FAIL: TestBreakLinkReferenceDefinitionsClosesEveryKnownDefinition/blockquote_continuation,_used_inside_the_blockquote (0.00s)
        markdown_test.go:149: a member-defined link survived the guard: "> [重置密码]:\n> https://evil.example\n>\n> [重置密码]" → "> [重置密码]:\n> https://evil.example\n>\n> [重置密码]" resolves [https://evil.example]
```

The agreement test and the sweep on the same revert:

```
--- FAIL: TestContainerPrefixHalvesAgree (0.00s)
    markdown_test.go:286: prefix "  >" may hold a definition, but the destination scan misses "\n>https://evil.example" —
        the two halves disagree about what a block container is
--- FAIL: TestBreakLinkReferenceDefinitionsAcrossBlockScaffolding (0.04s)
    markdown_test.go:335: definition behind " >" with its destination behind " >" survived the guard:
        " >[重置密码]:\n >https://evil.example\n\n[重置密码]" → " >[重置密码]:\n >https://evil.example\n\n[重置密码]"
```

### B — over-tightened

**B1, condition 3 removed so every `]:` breaks** — the version you told me not to ship:

```
"[Bug]: 登录失败"              → "[Bug] : 登录失败"                 ALTERED
"[文件]: report.pdf"         → "[文件] : report.pdf"            ALTERED
"[页面]: /inbox"             → "[页面] : /inbox"                ALTERED
"[WIP]: 明天再看"              → "[WIP] : 明天再看"                 ALTERED
"[Owner]: R&D"             → "[Owner] : R&D"                ALTERED
"[Regex]: \\d+"            → "[Regex] : \\d+"               ALTERED
"[Path]: docs\\setup"      → "[Path] : docs\\setup"         ALTERED
"[Query]: foo&bar"         → "[Query] : foo&bar"            ALTERED
```

```
--- FAIL: TestBreakLinkReferenceDefinitionsLeavesProseAlone (0.00s)
    markdown_test.go:187: member text was altered:
         got "[Bug] : 登录失败"
        want "[Bug]: 登录失败" (unchanged)
```

**B2, the escape trigger back to `strings.ContainsAny(dest, "\\&")`** — the nit's own reverse, showing the four lines are held by a test and not by a comment:

```
"[Bug]: 登录失败"              → "[Bug]: 登录失败"                  unchanged
"[文件]: report.pdf"         → "[文件]: report.pdf"             unchanged
"[页面]: /inbox"             → "[页面]: /inbox"                 unchanged
"[WIP]: 明天再看"              → "[WIP]: 明天再看"                  unchanged
"[Owner]: R&D"             → "[Owner] : R&D"                ALTERED
"[Regex]: \\d+"            → "[Regex] : \\d+"               ALTERED
"[Path]: docs\\setup"      → "[Path] : docs\\setup"         ALTERED
"[Query]: foo&bar"         → "[Query] : foo&bar"            ALTERED
```

```
--- FAIL: TestBreakLinkReferenceDefinitionsLeavesProseAlone (0.00s)
    markdown_test.go:187: member text was altered:
         got "[Owner] : R&D"
        want "[Owner]: R&D" (unchanged)
```

With the fix as landed, all eight of those come back unchanged and the whole package is green.

### Class size

Sweeping the scaffolding alphabet against itself with the container replay reverted:

```
prefixes=88 accepted-as-definition-lines=88 live-combinations=9066
distinct bypassing (definition-prefix, destination-prefix) pairs = 647
  ... of which the definition line carries no '>' : 0
  ... of which the destination line carries no '>': 0
```

That zero on both rows is the argument for making the two halves share a value rather than adding a `>` case: the bypass set was exactly the blockquote family, and the one thing every member of it had in common was a container the two halves disagreed about.

Bohan-J (approved):
Blocker closed, and closed the right way. `containerPrefix` carried from `containerPrefixBefore` into `looksLikeLinkDestination`, with `skipContinuationPrefix` as the explicit replay half, is the version I was hoping for — the two halves now share one notion of a container instead of one of them swallowing a character that happened to be in the way, and the comment saying exactly that is what will keep the next person from re-splitting them.

I re-ran my own matrix from the last round against `20503c83` with the goldmark checker. All twelve shapes that got through before are closed: single, no-space, tab, CRLF, two- and three-level nesting, list-inside-blockquote, bracketed and scheme-relative destinations, and the collapsed / full / image usages. The five that were already handled still are, and `https\://`, `\/\/` and `&#x68;ttps://` are still caught. The fidelity nit is gone too — `[Owner]: R&D`, `[Regex]: \d+`, `[Path]: docs\setup` and `[Query]: foo&bar` all come back byte-identical now, so the code and the "relative destinations stay valid" contract finally agree.

`go test -race`, `go vet`, `go build ./...` and `gofmt` clean on my side; CI green.

Worth saying plainly, since this ran four rounds: this is the best-argued PR I have reviewed in a while. You went and got tenant evidence instead of arguing the point, and that evidence killed your own first approach — the `\[Bug\]` math-delimiter finding is something none of us would have predicted from the code. You then replaced a five-character escape set with a one-line adjacency break, and when the tenant test showed reference definitions really do resolve, you brought them into scope rather than filing the residual and moving on. Recording that residual in §6 in the first place is what made it findable at all. And verifying every test by reverting its fix, with the cost stated in the assertion message, is a habit I wish more of our internal PRs had.

Two follow-ups from this thread, neither yours to carry:

- DingTalk `issueCreatedText` / `issueDuplicateText` and Slack's `formatMrkdwn` path have the same defect. I'll file that.
- A title ending in `\` meeting our `**` predates this PR, and now that backslash is known to be load-bearing on this renderer it deserves its own tenant test.

Merging. Thank you — genuinely good work.

PR #6594, opened 2026-08-07
fix(wecom): prove control of the bot before the install reclaim destroys somebody’s row (MUL-5891)

seacen:
## The hole

`idx_channel_installation_type_appid` is `UNIQUE (channel_type, config->>'app_id')` — **no workspace in the key**, so a bot id's routing slot is global across the whole deployment.

`ReclaimDeadChannelInstallationByAppID` then hard-deletes whoever holds that slot, and in the same statement clears every `channel_user_binding`, `channel_chat_session_binding` and pending binding token beneath it.

Nothing verified the caller has anything to do with that bot. `validateInstallationParams` only checks the fields are non-empty.

**Bot ids are not secret** — every member talking to the bot can see one. So an admin of *any* workspace could type a known bot id with a junk secret and destroy a revoked owner's installation and all of its bindings. The victim's users are unbound; nothing tells anyone why.

## The premise that was missing

The reclaim query's own comment already states the assumption this restores:

> workspace A disconnects; workspace B, **which proves control by holding the same app credentials**, rebinds

That sentence describes a check that did not exist. This PR makes it true.

## The probe

One connection, one subscribe, no read loop, no registration, no sender published — nothing else in the process learns it happened. Sized off the connection's own two constants (`handshakeTimeout + subscribeTimeout`) rather than a new number, because it is spent with an admin watching a spinner.

The verdict split matters as much as the check:

| | |
|---|---|
| `ErrCredentialsRejected` | WeCom refused the pair — the one answer entitled to blame the admin's credentials, and the one they can act on. 400. |
| `ErrCredentialsUnverifiable` | dial failed, handshake timed out, network down, **or any non-zero errcode not documented as a refusal of the pair**. **503.** Reporting these as a wrong secret sends the admin to rotate one that was fine — and a rotated WeCom long-connection secret cannot be recovered. |

Rejection is a **whitelist**: only the codes WeCom documents as refusing the credential pair itself (`40001` 不合法的secret参数, `40013` 不合法的CorpID) map to `Rejected`. WeCom only guarantees that `0` means success, and the subscribe path is under frequency and concurrency protection (`45009`, `45033`), so everything else non-zero fails closed as `Unverifiable` with the raw code logged and carried on the error. Codes only, never `errmsg` text — the message is human-facing prose WeCom is free to reword.

The probe is **mandatory**. `NewInstallationService` defaults it to the real handshake, refuses an explicit nil with `ErrProbeRequired`, and `Upsert` re-checks for a service built by struct literal. There is no fail-open setting: proof of control gates a statement that hard-deletes another workspace's row, so a deployment that cannot reach WeCom gets a 503 and retries rather than writing unverified credentials over somebody else's slot.

DingTalk already validates credentials at install time. WeCom was the outlier.

## Order of operations

The probe is itself a side effect on the live platform: it subscribes, and WeCom allows exactly one live subscriber per bot, so subscribing displaces whoever is connected. Probing before knowing who owns the slot would mean a request that is about to be **refused** still knocks the rightful owner offline — repeat it and that is a denial of service against a bot the caller was never permitted to touch.

So the whole sequence runs inside one transaction, in this order:

1. **Lock the slot** — `LockChannelInstallationAppIDSlot`, a `pg_advisory_xact_lock` keyed on `(hashtext(channel_type), hashtext(app_id))`, held until commit. A plain read-then-write leaves a TOCTOU window in which two callers both see "no live owner" and both go on to touch the slot.
2. **Read the current owner** — `GetChannelInstallationSlotOwnerByAppID`. `LEFT JOIN`ed on purpose so an **orphan** row survives the read: with no FKs (MUL-3515 §4) an installation outlives a deleted workspace or agent, and the caller has to tell "orphan, reclaimable" from "live owner, refuse".
3. **Refuse or proceed** — free / orphan / revoked / already this caller's own → proceed. Active and somebody else's → return the accurate 409 having touched **nothing**, locally or at WeCom.
4. **Probe, reclaim, upsert** — only now, with nothing live at stake.

The classification in step 3 deliberately mirrors `ReclaimDeadChannelInstallationByAppID`'s own definition of "dead", so the gate and the reclaim cannot drift into disagreeing about which rows are takeable.

No migration, no foreign key, no cascade; the two new queries are additive and `daemon/` is untouched.

## Tests

**Probe classification and wiring** (`credential_probe_test.go`, no socket needed): documented rejection codes map to `Rejected`; throttling, concurrency limits, system errors, unknown codes and transport failures all map to `Unverifiable`; the raw errcode is carried on the error; `NewInstallationService` refuses a nil probe and defaults to the real one; `Upsert` refuses to run without one.

**Ownership gate** (`installation_probe_db_test.go`, against a real Postgres): `Probe` is called **zero** times when an active owner, an archived owner, or another workspace holds the bot; a revoked owner is probed and the reclaim is refused when the probe rejects; a free slot is probed exactly once; and under a concurrent install the loser never subscribes behind the winner.

Full `internal/integrations/wecom` suite passes, including `-race` on the concurrency test, plus the DB-gated reclaim tests.

## Notes

- This adds two `errors.Is` branches to `RegisterWecomBYO`'s switch. **#6589** restructures that same switch to carry machine-readable codes; whichever lands second should give these two branches codes as well (`wecom_credentials_rejected` / `wecom_credentials_unverifiable`) plus their locale keys.

### DIFF
--- a/server/internal/handler/handler.go
+++ b/server/internal/handler/handler.go
@@ -280,6 +280,11 @@ type Handler struct {
 	// see wecom/binding.go). Nil disables the redeem endpoint (returns 503)
 	// and the OutboundReplier's binding-prompt path.
 	WecomBindingTokens WecomBindingRedeemer
+
+	// WecomCredentialProbe overrides the install-time control check. Nil in
+	// production, which gets the real handshake probe; tests inject a fake so
+	// the install path runs without a socket.
+	WecomCredentialProbe wecom.CredentialProbe
 	// LLM is the basic LLM API layer (MUL-4238): a thin wrapper over the
 	// OpenAI Go SDK backing server-internal one-shot LLM helpers such as chat
 	// title generation. The generic passthrough endpoints were removed in
--- a/server/internal/handler/wecom_web.go
+++ b/server/internal/handler/wecom_web.go
@@ -272,13 +272,14 @@ func (h *Handler) wecomInstallService() *wecom.InstallationService {
 	}
 	// The probe is what makes the reclaim inside Upsert safe to run against a
 	// row this caller may not own. h.WecomCredentialProbe is a test seam;
-	// production leaves it nil and gets the real handshake.
-	probe := h.WecomCredentialProbe
-	if probe == nil {
-		probe = wecom.NewHandshakeProbe(nil, "")
+	// production leaves it nil and NewInstallationService supplies the real
+	// handshake. There is no wiring that leaves the check off — a nil probe is
+	// a construction error there, not a fail-open mode.
+	var opts []wecom.InstallationOption
+	if h.WecomCredentialProbe != nil {
+		opts = append(opts, wecom.WithCredentialProbe(h.WecomCredentialProbe))
 	}
-	svc, err := wecom.NewInstallationService(h.Queries, h.TxStarter, res.Box,
-		wecom.WithCredentialProbe(probe))
+	svc, err := wecom.NewInstallationService(h.Queries, h.TxStarter, res.Box, opts...)
 	if err != nil {
 		return nil
 	}
--- a/server/internal/handler/wecom_web.go
+++ b/server/internal/handler/wecom_web.go
@@ -186,6 +186,19 @@ func (h *Handler) RegisterWecomBYO(w http.ResponseWriter, r *http.Request) {
 			writeError(w, http.StatusConflict, "this bot is connected to an archived agent in this workspace — restore that agent, or disconnect its bot, before connecting it here")
 		case errors.Is(err, wecom.ErrBotOwnedByAnotherWorkspace):
 			writeError(w, http.StatusConflict, "this bot is already connected to a different Multica workspace — disconnect it there before connecting it here")
+		case errors.Is(err, wecom.ErrCredentialsRejected):
+			// WeCom itself refused the pair. This is the one branch entitled
+			// to blame the credentials, and the only one the admin can act on
+			// by going back to the console.
+			writeError(w, http.StatusBadRequest, "WeCom rejected this Bot ID and secret — check both on the WeCom admin console, and that the bot is a smart bot with the long connection enabled")
+		case errors.Is(err, wecom.ErrCredentialsUnverifiable):
+			// The deployment could not reach WeCom. Reporting this as a bad
+			// credential sends the admin to rotate a secret that was fine —
+			// and a rotated WeCom secret cannot be recovered. 503: nothing is
+			// wrong with the input, the check could not be made.
+			slog.Warn("wecom install could not verify the bot",
+				"error", err, "workspace_id", uuidToString(wsUUID), "agent_id", uuidToString(agentUUID))
+			writeError(w, http.StatusServiceUnavailable, "could not reach WeCom to verify this bot — the credentials were not changed; try again in a moment")
 		default:
 			slog.Warn("wecom install failed",
 				"error", err, "workspace_id", uuidToString(wsUUID), "agent_id", uuidToString(agentUUID))
@@ -257,7 +270,15 @@ func (h *Handler) wecomInstallService() *wecom.InstallationService {
 	if !ok || res == nil || res.Box == nil {
 		return nil
 	}
-	svc, err := wecom.NewInstallationService(h.Queries, h.TxStarter, res.Box)
+	// The probe is what makes the reclaim inside Upsert safe to run against a
+	// row this caller may not own. h.WecomCredentialProbe is a test seam;
+	// production leaves it nil and gets the real handshake.
+	probe := h.WecomCredentialProbe
+	if probe == nil {
+		probe = wecom.NewHandshakeProbe(nil, "")
+	}
+	svc, err := wecom.NewInstallationService(h.Queries, h.TxStarter, res.Box,
+		wecom.WithCredentialProbe(probe))
 	if err != nil {
 		return nil
 	}
--- a/server/internal/integrations/wecom/credential_probe.go
+++ b/server/internal/integrations/wecom/credential_probe.go
@@ -1,0 +1,155 @@
+package wecom
+
+// credential_probe.go — proving the installer controls the bot.
+//
+// A BYO install used to persist whatever pair was pasted into the dialog. The
+// bot id is not a secret: it is on the WeCom admin console, and any member of
+// a workspace can read it back out of GET /wecom/installations. Only the
+// secret proves anything, and nothing checked it.
+//
+// Two things followed, and neither needs an insider.
+//
+// The routing slot is global. channel_installation's unique index on
+// (channel_type, config->>'app_id') has no workspace in it, so an admin of any
+// workspace on the deployment could write a row claiming somebody else's bot
+// id with a junk secret and hold that slot indefinitely. The rightful owner
+// then hits a conflict naming a workspace they cannot see.
+//
+// Worse, the reclaim is keyed on the same unproven id. A bot the rightful
+// owner has merely DISCONNECTED sits revoked, and a revoked row is what
+// ReclaimDeadChannelInstallationByAppID hard-deletes — along with every user
+// binding, session binding and pending token beneath it. So an outsider
+// pasting a bot id they read off a list could destroy another workspace's
+// bindings and force everyone there to re-link.
+//
+// The fix is the one both peers already use: prove control before persisting.
+// DingTalk mints an access token (dingtalk/byo_install.go), Slack calls
+// auth.test (slack/byo_install.go). WeCom has no REST surface, so the proof
+// is the protocol's own handshake — dial, aibot_subscribe, read the verdict,
+// hang up. A wrong secret is refused with a non-zero errcode, which is
+// exactly the signal.
+//
+// The probe never runs against a bot somebody is actively connected to.
+// WeCom allows one live subscriber per bot, so a probe against a live slot
+// would displace the rightful owner's connection, whose supervisor would
+// reconnect and displace the probe in turn. The caller checks for a live
+// owner first and lets the unique index produce the accurate conflict
+// instead; the case the probe exists for — a revoked row about to be
+// reclaimed — has nothing connected by definition.
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+// ErrCredentialsRejected is WeCom saying the pair is not valid: a wrong
+// secret, a bot that no longer exists, a bot whose API mode is off. It is the
+// answer that must reach the admin, because it is the one they can act on.
+var ErrCredentialsRejected = errors.New("wecom: WeCom rejected this bot id and secret")
+
+// ErrCredentialsUnverifiable is everything else — the dial failed, the
+// handshake timed out, the network is down. Distinct from rejection on
+// purpose: telling an admin their credentials are wrong when the deployment
+// simply could not reach WeCom sends them to rotate a secret that was fine.
+var ErrCredentialsUnverifiable = errors.New("wecom: could not reach WeCom to verify this bot")
+
+// credentialProbeTimeout bounds the whole probe. It has to cover a dial and
+// one round trip and nothing else, and it is spent with the admin watching a
+// spinner, so it is short. Sized off the connection's own two constants
+// rather than invented.
+const credentialProbeTimeout = handshakeTimeout + subscribeTimeout
+
+// CredentialProbe proves a (bot_id, secret) pair is live. Exported because
+// the handler holds one so a test can exercise the install path without a
+// socket; production always gets the handshake probe.
+type CredentialProbe interface {
+	Probe(ctx context.Context, botID, secret string) error
+}
+
+// credentialProbe is the unexported alias the service field uses.
+type credentialProbe = CredentialProbe
+
+// handshakeProbe is the production probe: one connection, one subscribe, no
+// read loop, no registration, no sender published. Nothing else in the
+// process learns it happened.
+type handshakeProbe struct {
+	dialer Dialer
+	wsURL  string
+}
+
+// NewHandshakeProbe builds the probe the install service uses. dialer and
+// wsURL are test seams; production passes nil and "".
+func NewHandshakeProbe(dialer Dialer, wsURL string) *handshakeProbe {
+	if wsURL == "" {
+		wsURL = DefaultWSURL
+	}
+	return &handshakeProbe{dialer: dialer, wsURL: wsURL}
+}
+
+func (p *handshakeProbe) Probe(ctx context.Context, botID, secret string) error {
+	dialer := p.dialer
+	if dialer == nil {
+		dialer = defaultDialer
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, credentialProbeTimeout)
+	defer cancel()
+
+	conn, _, err := dialer.DialContext(ctx, p.wsURL, nil)
+	if err != nil {
+		return fmt.Errorf("%w: dial: %v", ErrCredentialsUnverifiable, err)
+	}
+	defer conn.Close()
+
+	// The read below does not observe ctx, so the deadline has to be on the
+	// socket as well as on the context.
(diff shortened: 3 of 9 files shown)


Bohan-J (changes requested):
Thanks for this — the goal is right and worth landing: reclaiming a *global* bot slot must require proof of control, and splitting "WeCom rejected the pair" from "we could not reach WeCom" is exactly the right distinction to surface in the HTTP response, since a rotated WeCom secret cannot be recovered.

Three things block merge as written.

### 1. The probe fires before the active-owner check, so a rejected install can kick a live bot offline

`server/internal/integrations/wecom/installation.go` runs `s.probe.Probe(...)` at the very top of `Upsert`, before anything reads the existing row. The active / archived / other-workspace conflicts are only detected further down, at the unique-index conflict on `(channel_type, config->>'app_id')`. And `RegisterWecomBYO` in `server/internal/handler/wecom_web.go` calls `Upsert` directly, so there is no earlier owner check on the caller side either.

The header comment in `credential_probe.go` states the invariant the whole design rests on — *"The probe never runs against a bot somebody is actively connected to… The caller checks for a live owner first"* — but no caller actually does that. So a request carrying the **correct** credentials for a bot that is currently active (in this workspace or another one) subscribes first and gets its 409 second. WeCom allows one live subscriber per bot — as this PR's own comments note — so the legitimate connection is displaced *before* the request is refused. The rightful owner's supervisor reconnects, and a repeated request turns this into a denial of service against a bot the caller was never permitted to touch. A rejected request must not produce external side effects.

Suggested shape: establish a serialization boundary keyed on `(wecom, bot_id)` — a transaction-level advisory lock or equivalent — and read the current owner and status inside it. Return the conflict immediately for any active or archived owner; probe only when there is no active conflict and the row is genuinely absent or revoked; then perform the reclaim/upsert within that same boundary. A pre-read outside the transaction still leaves a TOCTOU window that concurrent installs and reconnects will hit.

### 2. Every non-zero errcode is reported as a credential rejection

`credential_probe.go` maps any `env.ErrCode != 0` to `ErrCredentialsRejected`, and the handler turns that into a 400 telling the admin to go check their Bot ID and secret. WeCom only guarantees that `0` means success; the subscribe path also has frequency/rate protection, so a non-zero code can mean throttled or platform-side failure rather than a wrong secret. That misclassification undoes the very distinction this PR sets out to make, and pushes admins toward rotating a secret that was fine and cannot be recovered.

Please whitelist only the codes documented as credential rejections, and fail closed on unknown non-zero codes as `ErrCredentialsUnverifiable` (503), logging the raw code and message. Please don't infer from `errmsg` text.

### 3. The check can be silently disabled at the service level

`InstallationService.probe` is optional (`if s.probe != nil`), and `TestNilProbeDoesNotBlockInstall` pins that fail-open behaviour as intended. Today's handler injects the real probe, but any new caller — or one wiring mistake — reopens exactly the destruction path this PR closes. The field comment ("a deployment that cannot reach WeCom at install time keeps the old behaviour") also contradicts what the handler now does, which is return 503.

Proof of control is a security invariant, so `NewInstallationService` should require a non-nil probe (or default to constructing the real one) rather than fail open on `nil`. Tests can inject a success fake explicitly.

### Test coverage

The new tests exercise fakes only. Please add regression coverage for:

- `Probe` is called **0** times when an active / archived / other-workspace owner exists;
- a concurrent install or reconnect never establishes a second subscription before being rejected;
- rate-limited and unknown non-zero codes map to unverifiable, while only documented rejection codes map to rejected;
- the service cannot be constructed without a probe.

Also, the branch currently conflicts with the base — it needs a rebase.

Happy to re-review once these are addressed.

seacen:
Thanks — point 1 is a real hole and the framing is right: a rejected request must not produce external side effects. All three are addressed, plus the rebase. Force-pushed as two commits so the follow-up is readable on its own.

### 1. The probe fired before the active-owner check

You are right that nothing checked. The header comment asserted an invariant no caller enforced, and `RegisterWecomBYO` went straight to `Upsert`.

Fixed the shape you suggested. `Upsert` now runs the whole sequence inside one transaction, in this order:

1. `LockChannelInstallationAppIDSlot` — a transaction-level advisory lock (`pg_advisory_xact_lock(hashtext(channel_type), hashtext(app_id))`), taken as the first statement and released by COMMIT/ROLLBACK.
2. `botSlotConflictErr` — reads the current owner **inside** that lock and returns the conflict immediately for any live owner. No probe, no write, nothing sent to WeCom.
3. Probe — only now, when what is in front of it is a free slot, a revoked row, an orphan, or the caller's own row.
4. Reclaim, then upsert, then commit — same lock still held.

The owner read needed more than `GetChannelInstallationOwnerByAppID` gives (it INNER JOINs the agent, so an orphan disappears and `status` is not returned at all), so there is a new query, `GetChannelInstallationSlotOwnerByAppID`: LEFT JOINs to workspace and agent, returning `status`, `agent_archived_at`, `agent_exists`, `workspace_exists`. That is exactly the input `ReclaimDeadChannelInstallationByAppID` uses to decide what is dead, so the gate and the DELETE cannot drift into disagreeing about which rows are takeable. Both new queries are in `pkg/db/queries/channel.sql` and regenerated with sqlc 1.31.1.

The post-upsert `botOwnerConflictErr` branch stays as the backstop for a slot that changes hands despite the lock; its comment now says that is what it is.

**One deviation, deliberate.** You wrote "return the conflict immediately for any active or archived owner". I return the conflict for every active or archived owner **except** one whose `(workspace_id, agent_id)` is the caller's own. That row is not a conflict — it is the admin re-submitting the install form for a bot they already hold, which is the secret-rotation path, and it has no sentinel or message because refusing it would just break re-install. It still probes, on two grounds: the reclaim cannot destroy anyone else's row in that case (the upsert hits its own `ON CONFLICT (workspace_id, agent_id, channel_type)` and refreshes in place), and the only connection a subscribe can displace there is the caller's own, which their supervisor re-establishes with the old secret if the new one is refused. The alternative — skipping the probe for your own row — silently stores an unverified secret and takes the bot down at the next reconnect with nothing said at install time. Happy to change it if you read the rule more strictly.

### 2. Every non-zero errcode reported as a credential rejection

Fixed as asked, whitelist plus fail-closed. `rejectionErrCodes` holds only the codes WeCom documents (document/path/90313) as a refusal of the pair:

- `40001` 不合法的secret参数
- `40013` 不合法的CorpID

Everything else non-zero — 45009/45033 frequency and concurrency protection, platform failures, anything WeCom adds later — returns `ErrCredentialsUnverifiable` (503), with the raw code and message logged at the probe (`slog.Warn`, so an operator sees it even if the caller only surfaces the sentinel) and carried on the error so the handler's own line has it too. No inference from `errmsg`; the codes are the only input.

I checked the aibot long-connection page: it documents `errcode: 0` for `aibot_subscribe` and defers everything else to the global table, so the whitelist is small on purpose. The cost is that a genuinely wrong secret answered with an undocumented code now gets a 503 "could not verify" instead of a 400 — which is the trade you asked for, and the logged code is what would justify widening the list.

### 3. The check could be disabled at the service level

Removed the fail-open entirely.

- `NewInstallationService` now defaults `probe` to `NewHandshakeProbe(nil, "")` and returns `ErrProbeRequired` if an option sets it back to nil. Omitting the option is no longer a way to omit the check.
- `Upsert` returns `ErrProbeRequired` on a nil probe as well, before anything else can fail first and hide it — that covers a service built by struct literal rather than the constructor.
- `TestNilProbeDoesNotBlockInstall` is gone. Tests inject a success fake explicitly, including the pre-existing reclaim tests.
- The field comment no longer claims a fail-open mode the handler does not implement.
- `wecomInstallService` only passes `WithCredentialProbe` when the test seam is set, so production takes the constructor default.

### Test coverage

All four you asked for, in `installation_probe_db_test.go` (DB-backed — the decision reads `channel_installation` JOINed to workspace and agent under an advisory lock; a fake queries layer would assert the mock) and `credential_probe_test.go`:

- `TestUpsert_DoesNotProbeWhenAnActiveOwnerHoldsTheBot` — probe called **0** times, and the victim's installation row id is unchanged afterwards. It passes the *correct* secret deliberately: the point is that even a caller who would pass the probe must not reach it.
- `TestUpsert_DoesNotProbeWhenAnArchivedOwnerHoldsTheBot`, `TestUpsert_DoesNotProbeWhenAnotherWorkspaceHoldsTheBot` — same, 0 calls.
- `TestUpsert_ConcurrentInstallDoesNotSubscribeBehindTheWinner` — the winner is parked inside the probe holding the lock when the loser arrives; the loser must reach WeCom 0 times, then be refused. This is the one that fails against a pre-read outside the transaction.
- `TestUpsert_ProbesAndRefusesBeforeReclaimingARevokedOwner` — probe still runs where it should, and a rejected pair leaves the revoked row standing.
- `TestProbeClassifiesSubscribeErrCodes` — table over 0 / 40001 / 40013 / 45009 / 45033 / -1 / 999999, asserting both that the right sentinel matches and that the opposite one does not.
- `TestNewInstallationServiceRefusesNilProbe`, `TestNewInstallationServiceDefaultsToTheRealProbe`, `TestUpsertRefusesWithoutAProbe`.
- The existing reclaim tests now assert probe call counts too (0 on every refusal, 1 on every takeover).

### Reverse verification

Each fix reverted in turn, verbatim output.

**Point 1** — probe moved back to the top of `Upsert`, no lock, no pre-read:

```
--- FAIL: TestUpsert_DoesNotProbeWhenAnActiveOwnerHoldsTheBot (0.09s)
    installation_probe_db_test.go:157: the probe subscribed 1 time(s) on a refused install — WeCom allows one live subscriber per bot, so this disconnects the rightful owner and a repeat is a denial of service against a bot the caller may not touch
--- FAIL: TestUpsert_DoesNotProbeWhenAnArchivedOwnerHoldsTheBot (0.07s)
    installation_probe_db_test.go:178: the probe subscribed 1 time(s) against an archived owner's bot; a refused install must reach WeCom zero times
--- FAIL: TestUpsert_DoesNotProbeWhenAnotherWorkspaceHoldsTheBot (0.05s)
    installation_probe_db_test.go:197: the probe subscribed 1 time(s) against a bot owned by another workspace; a refused install must reach WeCom zero times
--- FAIL: TestUpsert_ConcurrentInstallDoesNotSubscribeBehindTheWinner (0.59s)
    installation_probe_db_test.go:288: the racing install subscribed 1 time(s) while another install held the bot's slot — that displaces the connection the winner is about to establish
    installation_probe_db_test.go:294: first install = wecom: this bot is already connected to another agent in this workspace, want success
--- FAIL: TestUpsert_RefusesLiveOwnerOnDifferentAgent (0.08s)
    installation_reclaim_test.go:188: probe called 1 time(s) on a refused install; a live owner is decided from the row, without touching WeCom
--- FAIL: TestUpsert_RefusesArchivedOwner (0.07s)
    installation_reclaim_test.go:203: probe called 1 time(s) on a refused install; an archived owner is decided from the row, without touching WeCom
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	1.753s
FAIL
```

Note the fourth one: with the probe first, the *winner* also loses — the racing install got its subscribe in and took the slot.

**Point 2** — every non-zero errcode back to `ErrCredentialsRejected`:

```
--- FAIL: TestProbeClassifiesSubscribeErrCodes (0.00s)
    --- FAIL: TestProbeClassifiesSubscribeErrCodes/system_error_is_not_a_rejection (0.00s)
        credential_probe_test.go:165: errcode -1: Probe = wecom: WeCom rejected this bot id and secret (errcode -1: system busy), want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestProbeClassifiesSubscribeErrCodes/rate_limited_is_not_a_rejection (0.00s)
        credential_probe_test.go:165: errcode 45009: Probe = wecom: WeCom rejected this bot id and secret (errcode 45009: api freq out of limit), want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestProbeClassifiesSubscribeErrCodes/unknown_code_is_not_a_rejection (0.00s)
        credential_probe_test.go:165: errcode 999999: Probe = wecom: WeCom rejected this bot id and secret (errcode 999999: something new), want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestProbeClassifiesSubscribeErrCodes/concurrency_limited_is_not_a_rejection (0.00s)
        credential_probe_test.go:165: errcode 45033: Probe = wecom: WeCom rejected this bot id and secret (errcode 45033: api concurrency out of limit), want wecom: could not reach WeCom to verify this bot
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.536s
FAIL
```

**Point 3** — nil probe allowed to fail open again:

```
--- FAIL: TestUpsertRefusesWithoutAProbe (0.00s)
    credential_probe_test.go:252: Upsert without a probe = wecom: InstallationService requires a transaction starter, want ErrProbeRequired
--- FAIL: TestNewInstallationServiceDefaultsToTheRealProbe (0.00s)
    credential_probe_test.go:230: probe is nil — an install would reach the reclaim with an unproven credential
--- FAIL: TestNewInstallationServiceRefusesNilProbe (0.00s)
    credential_probe_test.go:212: NewInstallationService with an explicit nil probe = <nil>, want ErrProbeRequired
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.551s
FAIL
```

### Rebase

Rebased onto `main` (`dce176cce`). One conflict, in `handler.go`: `main` changed `WecomBindingTokens` from `*wecom.BindingTokenService` to the `WecomBindingRedeemer` interface while this branch added `WecomCredentialProbe` next to it. Took `main`'s type and kept the new field. `pkg/db/generated/channel.sql.go` was regenerated with sqlc rather than hand-edited.

Green after the rebase:

```
ok  	github.com/multica-ai/multica/server/internal/integrations/wecom	2.274s
ok  	github.com/multica-ai/multica/server/internal/handler	28.341s
```

(`go test -race … -count=1` against a migrated Postgres, so the DB-gated tests actually ran rather than skipped. `go build ./...`, `go vet ./...` and `gofmt -l` clean on every file touched.)

Bohan-J (approved):
Thank you — that was a fast and unusually thorough turnaround, and all three blockers are genuinely closed in `ef4f762c`.

- **Ordering.** The install now takes `pg_advisory_xact_lock` on `(channel_type, app_id)` as the first statement of the transaction, reads the slot owner inside that lock, and refuses active / archived / other-workspace holders **before** anything reaches WeCom. A rejected request now touches nothing, locally or on the platform. `GetChannelInstallationSlotOwnerByAppID` being `LEFT JOIN`ed so orphans survive the read is the right call, and mirroring `ReclaimDeadChannelInstallationByAppID`'s own definition of "dead" keeps the gate and the reclaim from drifting apart.
- **Classification.** Rejection is a whitelist (`40001`, `40013`), everything else non-zero fails closed as unverifiable with the raw code logged and carried on the error. Matching on codes rather than `errmsg` prose is exactly right.
- **Fail-closed wiring.** `NewInstallationService` defaults to the real handshake, refuses an explicit nil with `ErrProbeRequired`, and `Upsert` re-checks for a struct-literal service. No path leaves the check off.

The regression coverage is all there too — zero `Probe` calls on every live-owner case, the concurrent install where the loser never subscribes behind the winner (nice that it runs under `-race`), and the errcode classification table.

I've updated the PR description directly so the merge record matches the final design: the old text still described the nil-probe fail-open and claimed the "never probe a live bot" constraint held by construction at the call site, which the advisory-lock gate now enforces properly. No code change, just the prose.

CI is green and the branch is clean — merging.

One process note, and please take it as logistics rather than criticism: you currently have ~22 PRs open at once. We do want them all, and the quality here is high, but a batch that size can't be reviewed in parallel with any care — so we'll be working through them **in order**, roughly oldest-first and respecting dependencies between them (#6589 and this one overlap, for instance). Expect feedback to arrive staggered rather than all at once. If you can keep a smaller number in flight going forward — landing a few before opening the next batch — they'll each get through review considerably faster, and you'll hit fewer rebase conflicts between your own branches.

Thanks again for catching and fixing this one.

PR #6595, opened 2026-08-07
MUL-5892 fix(chat): cancel queued tasks when a session is archived, and stop telling non-Slack sessions they have no channel

seacen:
Two channel-agnostic fixes found while auditing WeCom. Both help Lark and DingTalk equally, so they are written without naming a platform.

## 1. Archiving a session left its queued tasks claimable

`SetChatSessionArchived` deletes the channel binding but never cancelled the session's in-flight tasks — and `ClaimAgentTask` does not read `chat_session.status`.

So the daemon still runs the turn, on a conversation the user closed: runtime and quota spent, new assistant messages written into an archived chat, and the agent flipped back to `working`. The binding row is *already gone* by then, which makes the run mis-briefed as well as wasted:

- the runtime brief describes a **private web chat** (no binding → no channel), which is exactly the misattribution fix 2 below exists to prevent
- the outbound senders resolve their destination through that same deleted row, so there is no route left to answer on

The cancel now runs first, in the same transaction. `DeleteChatSession` has always done exactly this, and its query comment already gives the reason — *"so the daemon doesn't keep running work whose result has nowhere to land"*. Archiving means the same thing: the owner said they are done.

`chat.sql`'s own comment on `DeleteChatSession` assumes the caller "has already cancelled any in-flight tasks", so this brings archive in line with the invariant the schema comments already describe.

The cancel is scoped to sessions that were bound to a channel, read before the delete erases the evidence. A web-only chat has no room and no adapter, and cancelling there would destroy the user's typed prompt: archive is the reversible, unconfirmed action the UI offers, while Stop hands the prompt back through `CancelTaskByUser` and `CancelAgentTasksByChatSession` does not.

### The window the cancel cannot reach

The channel run trigger is debounced by `DefaultChatRunBatchWindow`. The message row is persisted on arrival; the task row is created a window later, when `flushChatRun` reloads the session and calls `EnqueueChatTask`. Archive inside that window and the cancel finds nothing to cancel, drops the binding and commits — then the flush enqueues onto the session that was just archived, and the daemon runs it.

`EnqueueChatTask` was the one enqueue path without the guard the direct-send path and `OpenMikaOnboardingChat` already take. It now locks the chat session as its first statement, re-reads it, and refuses a status that is not `active` with the existing `ErrChatSessionArchived`.

That lock is a new `LockChatSessionForEnqueue` (`FOR NO KEY UPDATE`) rather than the `FOR UPDATE` the send, delete and draft paths take. Those paths want to block INSERTs that reference the row — `FOR UPDATE` conflicts with the `FOR KEY SHARE` an FK insert takes on its parent, which is how `LockChatSessionForDelete` keeps a send from slipping a task in between its cancel and its delete. Here that conflict is the cost rather than the point: every inbound channel message is an `INSERT INTO chat_message` FK'd to this row, so `FOR UPDATE` would make a room's next message wait for the previous message's enqueue. `FOR NO KEY UPDATE` still conflicts with `SetChatSessionArchived`'s UPDATE (status is not a key column) and with the delete path's lock, in both directions, so the archive and the delete keep serialising against the enqueue while ingestion does not.

`flushChatRun` needs no change: it clears the typing indicator before its error switch, and this error is not one of the two the switch names, so it falls to the default log branch and posts nothing into a room the archive just unbound.

## 2. The empty-history note lied to every non-Slack session

The channel-history reader is Slack-only, so WeCom, Lark and DingTalk sessions all land on:

> This conversation is not connected to a chat channel, so there is no channel history to read.

That is false for all three, and **the reader is an agent deciding who can see its answer**. Told it is in a web-only conversation it reasons differently than told it is in a group whose backlog it cannot read.

The note now names the actual platform and the response carries `channel_type` (already a field on `HistoryResponse`, previously never set on this branch). `GetChannelChatSessionBindingBySessionAny` already exists on main.

Written channel-agnostically on purpose: a per-platform lookup here would go blind the next time a channel is added, which is precisely how the current wording came to be wrong.

The binding lookup behind it also mapped every failure to "no channel", so a transient database error answered 200 carrying the exact note this PR removes. It returns `(string, error)` now, with only `pgx.ErrNoRows` mapped to empty; anything else is reported.

## Tests

Against a real Postgres, driving the real paths rather than the helpers — I caught myself testing `noHistoryNote` directly first, and it kept passing when I reverted the handler wiring, so it now goes through `respondChatHistory`.

Reverting each fix:

```
task status after archiving a channel-bound session = "queued", want cancelled —
the daemon will still run this turn on a closed conversation, spending runtime
and quota and writing into an archived chat, and with the binding already gone
it is briefed as a private web chat with no route left to answer on

the debounced flush created 1 task(s) on a session archived while its window was
open — the daemon claims and runs that turn on a closed conversation

an archive landing mid-enqueue got <nil>, want to be blocked on the chat_session
lock — it would otherwise commit while this task row is still invisible, cancel
nothing, and leave a queued turn on a closed conversation

the room's next message could not be appended during an enqueue: ERROR:
canceling statement due to lock timeout (SQLSTATE 55P03)

archived-session flush posted "agent_archived" into a room the archive had
already unbound

a WeCom-bound session was told it is not on a channel: "This conversation is
not connected to a chat channel, so there is no channel history to read."

a failed binding read answered 200 {"channel_type":"","messages":[],"note":"This
conversation is not connected to a chat channel, …"} — the agent is told it is
in a web-only conversation because a query failed
```

`internal/handler`, `internal/service` and `internal/integrations/channel/engine` pass under `-race`.

### DIFF
--- a/server/internal/handler/chat.go
+++ b/server/internal/handler/chat.go
@@ -553,6 +553,22 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 	}
 
 	if req.Archived {
+		// Cancel first, in the same transaction. ClaimAgentTask does not read
+		// chat_session.status, so a task queued before the archive stays
+		// claimable after it — and once the binding row below is gone, the
+		// runtime brief describes a private web chat while the channel
+		// adapter still holds the chat id and posts the answer into the
+		// group. The person who archived the conversation never sees it
+		// happen.
+		//
+		// This is also what archiving means: the owner said they are done
+		// with this conversation. DeleteChatSession has always cancelled for
+		// the same reason (see the CancelAgentTasksByChatSession call below),
+		// and chat.sql's own comment already assumes archive does too.
+		if _, err := qtx.CancelAgentTasksByChatSession(r.Context(), session.ID); err != nil {
+			writeError(w, http.StatusInternalServerError, "failed to cancel queued tasks for the archived session")
+			return
+		}
 		if err := qtx.DeleteChannelChatSessionBindingBySession(r.Context(), session.ID); err != nil {
 			writeError(w, http.StatusInternalServerError, "failed to clear chat session channel binding")
 			return
--- a/server/internal/handler/chat.go
+++ b/server/internal/handler/chat.go
@@ -575,12 +575,15 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 		case bindingErr == nil:
 			// Bound, and the delete below is about to drop that binding.
 			// ClaimAgentTask does not read chat_session.status, so a task
-			// queued before the archive stays claimable after it — and once
-			// the binding row is gone the runtime brief describes a private
-			// web chat while the channel adapter still holds the chat id and
-			// posts the answer into the group. The person who archived the
-			// conversation never sees it happen. Cancel in the same tx that
-			// drops the binding, as DeleteChatSession does.
+			// queued before the archive stays claimable after it: the daemon
+			// runs a turn on a conversation the user closed, spending runtime
+			// and quota and writing assistant messages into an archived chat.
+			// The binding is gone by then, so the brief describes the run as a
+			// private web chat — the misattribution the history note below
+			// exists to prevent — while the outbound senders, which resolve
+			// their destination through that same deleted row, have no route
+			// left to deliver on. Cancel in the same tx that drops the
+			// binding, as DeleteChatSession does.
 			//
 			// The query is not limited to never-started work: it also matches
 			// dispatched, running, waiting_local_directory and deferred, which
--- a/server/internal/handler/chat.go
+++ b/server/internal/handler/chat.go
@@ -516,6 +516,15 @@ type SetChatSessionArchivedRequest struct {
 // reviving this archived one. Unarchive deliberately does NOT recreate the
 // binding: if later traffic already forked a new session, that session owns the
 // channel now, and restoring the old binding would steal it back.
+//
+// Severing that binding is what makes archiving a bound session cancel its
+// in-flight tasks: the adapter still holds the chat id, so a turn that survives
+// the archive answers into a group room the session no longer belongs to. That
+// cancel is therefore scoped to sessions that HAD a binding. Archiving a
+// web-only chat leaves its tasks alone — it is the reversible organizing action
+// the UI presents (no confirmation, unarchive undoes it), and the destructive
+// counterpart is the Stop button, which cancels through CancelTaskByUser and
+// gives the user their typed prompt back.
 func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request) {
 	userID, ok := requireUserID(w, r)
 	if !ok {
@@ -553,30 +562,52 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 	}
 
 	// Rows cancelled inside the tx, broadcast after it commits. nil when the
-	// request unarchives; BroadcastCancelledTasks is a no-op on an empty slice.
+	// request unarchives and nil for a web-only chat; BroadcastCancelledTasks
+	// is a no-op on an empty slice.
 	var cancelled []db.AgentTaskQueue
 
 	if req.Archived {
-		// Cancel first, in the same transaction. ClaimAgentTask does not read
-		// chat_session.status, so a task queued before the archive stays
-		// claimable after it — and once the binding row below is gone, the
-		// runtime brief describes a private web chat while the channel
-		// adapter still holds the chat id and posts the answer into the
-		// group. The person who archived the conversation never sees it
-		// happen.
-		//
-		// This is also what archiving means: the owner said they are done
-		// with this conversation. DeleteChatSession has always cancelled for
-		// the same reason (see the CancelAgentTasksByChatSession call below),
-		// and chat.sql's own comment already assumes archive does too.
-		//
-		// The query is not limited to never-started work: it also matches
-		// dispatched, running, waiting_local_directory and deferred, which is
-		// why the returned rows have to reach the post-commit broadcast below
-		// rather than being discarded.
-		cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
-		if err != nil {
-			writeError(w, http.StatusInternalServerError, "failed to cancel queued tasks for the archived session")
+		// Read the binding BEFORE the delete below, which is what erases the
+		// evidence: after it runs there is no way to tell a session that was
+		// on a channel from one that never was.
+		_, bindingErr := qtx.GetChannelChatSessionBindingBySessionAny(r.Context(), session.ID)
+		switch {
+		case bindingErr == nil:
+			// Bound, and the delete below is about to drop that binding.
+			// ClaimAgentTask does not read chat_session.status, so a task
+			// queued before the archive stays claimable after it — and once
+			// the binding row is gone the runtime brief describes a private
+			// web chat while the channel adapter still holds the chat id and
+			// posts the answer into the group. The person who archived the
+			// conversation never sees it happen. Cancel in the same tx that
+			// drops the binding, as DeleteChatSession does.
+			//
+			// The query is not limited to never-started work: it also matches
+			// dispatched, running, waiting_local_directory and deferred, which
+			// is why the returned rows have to reach the post-commit broadcast
+			// below rather than being discarded.
+			cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
+			if err != nil {
+				writeError(w, http.StatusInternalServerError, "failed to cancel queued tasks for the archived session")
+				return
+			}
+		case errors.Is(bindingErr, pgx.ErrNoRows):
+			// A web-only chat has no room, no adapter and nowhere for a late
+			// answer to land, so there is nothing here to protect anyone from
+			// and archiving stays what the UI already says it is: the
+			// reversible "put it away" step that unarchive undoes, offered
+			// with no confirmation, with hard delete available only after it.
+			// Cancelling would also destroy the user's typed prompt — the Stop
+			// button cancels through CancelTaskByUser, which hands the prompt
+			// back (cancelled_chat_message.restore_to_input, or a durable
+			// chat_draft_restore row), and CancelAgentTasksByChatSession does
+			// neither.
+		default:
+			// We cannot tell whether this session was bound. Carrying on would
+			// silently skip the cancel and reopen the bug above, so fail the
+			// archive rather than guess; the caller can retry.
+			slog.Warn("read chat session channel binding failed", "session_id", sessionID, "error", bindingErr)
+			writeError(w, http.StatusInternalServerError, "failed to read chat session channel binding")
 			return
 		}
 		if err := qtx.DeleteChannelChatSessionBindingBySession(r.Context(), session.ID); err != nil {
--- a/server/internal/handler/chat.go
+++ b/server/internal/handler/chat.go
@@ -552,6 +552,10 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	// Rows cancelled inside the tx, broadcast after it commits. nil when the
+	// request unarchives; BroadcastCancelledTasks is a no-op on an empty slice.
+	var cancelled []db.AgentTaskQueue
+
 	if req.Archived {
 		// Cancel first, in the same transaction. ClaimAgentTask does not read
 		// chat_session.status, so a task queued before the archive stays
@@ -565,7 +569,13 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 		// with this conversation. DeleteChatSession has always cancelled for
 		// the same reason (see the CancelAgentTasksByChatSession call below),
 		// and chat.sql's own comment already assumes archive does too.
-		if _, err := qtx.CancelAgentTasksByChatSession(r.Context(), session.ID); err != nil {
+		//
+		// The query is not limited to never-started work: it also matches
+		// dispatched, running, waiting_local_directory and deferred, which is
+		// why the returned rows have to reach the post-commit broadcast below
+		// rather than being discarded.
+		cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
+		if err != nil {
 			writeError(w, http.StatusInternalServerError, "failed to cancel queued tasks for the archived session")
 			return
 		}
@@ -580,6 +590,15 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
 		writeError(w, http.StatusInternalServerError, "failed to commit chat session update")
 		return
 	}
+
+	// Post-commit broadcasts, same as DeleteChatSession: subscribers should
+	// never observe events for a tx that didn't actually persist. This is the
+	// call that captures the cancellation and revokes the tasks' mat_ tokens,
+	// reconciles each agent off 'working', emits task:cancelled so other
+	// clients drop the row instead of showing it queued until the next
+	// refresh, and wakes the runtime so a queued successor is claimed now
+	// rather than at the daemon's next poll.
+	h.TaskService.BroadcastCancelledTasks(r.Context(), cancelled)
 
 	resolvedSessionID := uuidToString(updated.ID)
 	status := updated.Status
--- a/server/internal/handler/chat_history.go
+++ b/server/internal/handler/chat_history.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 
 	"github.com/multica-ai/multica/server/internal/integrations/channel"
@@ -128,7 +129,13 @@ func (h *Handler) respondChatHistory(w http.ResponseWriter, r *http.Request, ses
 			// lets an archive land between them and produce a response whose
 			// channel_type names a platform while its note says there is no
 			// channel.
-			channelType := h.sessionChannelType(r.Context(), sessionID)
+			channelType, bindingErr := h.sessionChannelType(r.Context(), sessionID)
+			if bindingErr != nil {
+				slog.Error("chat session channel binding read failed", append(logger.RequestAttrs(r),
+					"error", bindingErr, "chat_session_id", uuidToString(sessionID))...)
+				writeError(w, http.StatusInternalServerError, "failed to read chat session channel binding")
+				return
+			}
 			writeJSON(w, http.StatusOK, ChatChannelHistoryResponse{
 				ChannelType: channelType,
 				Messages:    []channel.HistoryMessage{},
@@ -202,13 +209,23 @@ func noHistoryNote(channelType string) string {
 // none. Channel-agnostic on purpose: a per-platform lookup here would go blind
 // the next time a channel is added, which is exactly how the note above came
 // to be wrong.
-func (h *Handler) sessionChannelType(ctx context.Context, sessionID pgtype.UUID) string {
+//
+// Only "no such row" means "no channel". Any other failure is us being unable
+// to tell, and answering "" there hands the agent the very note this change
+// removes — a WeCom or Lark session told it is web-only, on a 200, because a
+// connection blipped. The caller reports that rather than guessing, the same
+// way the archive path refuses to guess about the same read.
+func (h *Handler) sessionChannelType(ctx context.Context, sessionID pgtype.UUID) (string, error) {
 	if h.Queries == nil || !sessionID.Valid {
-		return ""
+		return "", nil
 	}
 	binding, err := h.Queries.GetChannelChatSessionBindingBySessionAny(ctx, sessionID)
-	if err != nil {
-		return ""
+	switch {
+	case err == nil:
+		return binding.ChannelType, nil
+	case errors.Is(err, pgx.ErrNoRows):
+		return "", nil
+	default:
+		return "", err
 	}
-	return binding.ChannelType
 }
--- a/server/internal/handler/chat_history.go
+++ b/server/internal/handler/chat_history.go
@@ -125,8 +125,9 @@ func (h *Handler) respondChatHistory(w http.ResponseWriter, r *http.Request, ses
 	if err != nil {
 		if errors.Is(err, slack.ErrNoSlackSession) {
 			writeJSON(w, http.StatusOK, ChatChannelHistoryResponse{
-				Messages: []channel.HistoryMessage{},
-				Note:     "This conversation is not connected to a chat channel, so there is no channel history to read.",
+				ChannelType: h.sessionChannelType(r.Context(), sessionID),
+				Messages:    []channel.HistoryMessage{},
+				Note:        h.noHistoryNote(r.Context(), sessionID),
 			})
 			return
 		}
@@ -173,4 +174,36 @@ func parseHistoryLimit(raw string) int {
 		return 0
 	}
 	return n
+}
+
+// noHistoryNote explains an empty read in terms the agent can act on.
+//
+// The reader is Slack-only, so every other platform lands here — and the note
+// said "this conversation is not connected to a chat channel", which for a
+// WeCom, Lark or DingTalk session is simply false. An agent told it is in a
+// web-only conversation reasons differently about who can see its answer than
+// one told it is in a group whose backlog it cannot read, and that is a
+// difference worth not lying about.
+func (h *Handler) noHistoryNote(ctx context.Context, sessionID pgtype.UUID) string {
+	switch ct := h.sessionChannelType(ctx, sessionID); ct {
+	case "":
+		return "This conversation is not connected to a chat channel, so there is no channel history to read."
+	default:
+		return "This conversation is on " + ct + ", whose backlog this server cannot read. You can see the messages addressed to you in this session, but not the rest of the room."
+	}
+}
+
+// sessionChannelType names the platform behind a session, or "" when there is
+// none. Channel-agnostic on purpose: a per-platform lookup here would go blind
+// the next time a channel is added, which is exactly how the note above came
+// to be wrong.
+func (h *Handler) sessionChannelType(ctx context.Context, sessionID pgtype.UUID) string {
+	if h.Queries == nil || !sessionID.Valid {
+		return ""
+	}
+	binding, err := h.Queries.GetChannelChatSessionBindingBySessionAny(ctx, sessionID)
+	if err != nil {
+		return ""
+	}
+	return binding.ChannelType
 }
--- a/server/internal/handler/chat_history.go
+++ b/server/internal/handler/chat_history.go
@@ -124,10 +124,15 @@ func (h *Handler) chatHistorySession(w http.ResponseWriter, r *http.Request) (pg
 func (h *Handler) respondChatHistory(w http.ResponseWriter, r *http.Request, sessionID pgtype.UUID, page channel.HistoryPage, err error) {
 	if err != nil {
 		if errors.Is(err, slack.ErrNoSlackSession) {
+			// One read of the binding, two derived fields. Reading it twice
+			// lets an archive land between them and produce a response whose
+			// channel_type names a platform while its note says there is no
+			// channel.
+			channelType := h.sessionChannelType(r.Context(), sessionID)
 			writeJSON(w, http.StatusOK, ChatChannelHistoryResponse{
-				ChannelType: h.sessionChannelType(r.Context(), sessionID),
+				ChannelType: channelType,
 				Messages:    []channel.HistoryMessage{},
-				Note:        h.noHistoryNote(r.Context(), sessionID),
+				Note:        noHistoryNote(channelType),
 			})
 			return
 		}
@@ -176,7 +181,9 @@ func parseHistoryLimit(raw string) int {
 	return n
 }
 
-// noHistoryNote explains an empty read in terms the agent can act on.
+// noHistoryNote explains an empty read in terms the agent can act on. It is a
+// pure function of the channel type its caller already resolved, so the note
+// and the response's channel_type cannot disagree.
 //
 // The reader is Slack-only, so every other platform lands here — and the note
 // said "this conversation is not connected to a chat channel", which for a
@@ -184,13 +191,11 @@ func parseHistoryLimit(raw string) int {
 // web-only conversation reasons differently about who can see its answer than
 // one told it is in a group whose backlog it cannot read, and that is a
 // difference worth not lying about.
-func (h *Handler) noHistoryNote(ctx context.Context, sessionID pgtype.UUID) string {
-	switch ct := h.sessionChannelType(ctx, sessionID); ct {
-	case "":
+func noHistoryNote(channelType string) string {
+	if channelType == "" {
 		return "This conversation is not connected to a chat channel, so there is no channel history to read."
-	default:
-		return "This conversation is on " + ct + ", whose backlog this server cannot read. You can see the messages addressed to you in this session, but not the rest of the room."
 	}
+	return "This conversation is on " + channelType + ", whose backlog this server cannot read. You can see the messages addressed to you in this session, but not the rest of the room."
 }
 
 // sessionChannelType names the platform behind a session, or "" when there is
--- a/server/internal/handler/chat_archive_cancel_test.go
+++ b/server/internal/handler/chat_archive_cancel_test.go
@@ -2,9 +2,12 @@ package handler
 
 // chat_archive_cancel_test.go — archiving a chat session deletes its channel
 // binding but used to leave queued tasks claimable. ClaimAgentTask does not
-// read chat_session.status, so the daemon still ran the turn — and with the
-// binding row gone the runtime brief described a private web chat while the
-// channel adapter still held the chat id and posted the answer into the group.
+// read chat_session.status, so the daemon still ran the turn: runtime and
+// quota spent on a conversation the user closed, assistant messages written
+// into an archived chat, and the agent flipped back to 'working'. The binding
+// row is already gone by then, so the brief describes the run as a private web
+// chat, and the outbound senders — which resolve their destination through
+// that same row — have nowhere to deliver the answer.
 //
 // The cancel is therefore scoped to sessions that HAD a binding, which is what
 // the two tests below are: the same fixture archived twice, once bound and once
@@ -23,16 +26,19 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/jackc/pgx/v5/pgtype"
 
 	"github.com/multica-ai/multica/server/internal/events"
 	"github.com/multica-ai/multica/server/internal/integrations/channel"
 	"github.com/multica-ai/multica/server/internal/integrations/slack"
+	"github.com/multica-ai/multica/server/internal/service"
 	"github.com/multica-ai/multica/server/pkg/protocol"
 )
 
@@ -42,10 +48,10 @@ func TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks(t *testing.T) {
 	}
 	f := newArchiveCancelFixture(t, "Archive Cancels Queued")
 
-	// The binding is the reason this cancel exists at all: the archive drops it
-	// a few lines after deciding, and the adapter on the other side still knows
-	// the room. Bind before archiving, or the test is a web chat wearing the
-	// bug's failure message.
+	// The binding is the reason this cancel exists at all: it is what makes the
+	// session a real conversation in a room, and the archive drops it a few
+	// lines after deciding. Bind before archiving, or the test is a web chat
+	// wearing the bug's failure message.
 	f.bindToChannel(t, "GROUP_ARCHIVE_CANCEL")
 
 	f.archive(t)
@@ -55,7 +61,7 @@ func TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks(t *testing.T) {
 		{"running", f.runningTaskID},
 	} {
 		if status := f.taskStatus(t, tc.id); status != "cancelled" {
-			t.Fatalf("%s task status after archiving a channel-bound session = %q, want cancelled — the daemon will still run this turn, and with the channel binding already gone its answer goes to the room while the brief says private", tc.label, status)
+			t.Fatalf("%s task status after archiving a channel-bound session = %q, want cancelled — the daemon will still run this turn on a closed conversation, spending runtime and quota and writing into an archived chat, and with the binding already gone it is briefed as a private web chat with no route left to answer on", tc.label, status)
 		}
 	}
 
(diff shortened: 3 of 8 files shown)


server/internal/handler/chat.go
```diff
@@ -553,6 +553,22 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
...
+		// runtime brief describes a private web chat while the channel
+		// adapter still holds the chat id and posts the answer into the
+		// group. The person who archived the conversation never sees it
+		// happen.
+		//
+		// This is also what archiving means: the owner said they are done
+		// with this conversation. DeleteChatSession has always cancelled for
+		// the same reason (see the CancelAgentTasksByChatSession call below),
+		// and chat.sql's own comment already assumes archive does too.
+		if _, err := qtx.CancelAgentTasksByChatSession(r.Context(), session.ID); err != nil {
```
Bohan-J: The rows returned here are dropped. `DeleteChatSession` keeps them (`cancelled, err := qtx.CancelAgentTasksByChatSession(...)`) and calls `h.TaskService.BroadcastCancelledTasks(r.Context(), cancelled)` after `tx.Commit` — that is what emits `task:cancelled`, runs `ReconcileAgentStatus`, captures the cancellation, and wakes the runtime via `notifyTasksFinished`.

As written, the archive path persists `status = 'cancelled'` and stops there: agents can stay `working`, other clients keep a stale task, and the runtime waits for its next poll before claiming the next task. Please capture the result and reuse the same post-commit broadcast.

server/internal/handler/chat_history.go
```diff
@@ -125,8 +125,9 @@ func (h *Handler) respondChatHistory(w http.ResponseWriter, r *http.Request, ses
 	if err != nil {
 		if errors.Is(err, slack.ErrNoSlackSession) {
 			writeJSON(w, http.StatusOK, ChatChannelHistoryResponse{
-				Messages: []channel.HistoryMessage{},
-				Note:     "This conversation is not connected to a chat channel, so there is no channel history to read.",
+				ChannelType: h.sessionChannelType(r.Context(), sessionID),
```
Bohan-J: The binding is read twice for one response — once here for `channel_type`, and again inside `noHistoryNote`. If an archive deletes the binding between the two reads, the response can say `channel_type: "wecom"` while the note says the conversation is not connected to a chat channel.

Suggest resolving the channel type once and passing it into the note builder.

Bohan-J (changes requested):
The product change is right on both counts — archiving should mean the queued work actually stops, and the empty-history note shouldn't tell a WeCom/Lark/DingTalk session it has no channel. One blocker on the cancel path, plus a nit.

## Must-fix — the archive cancel never reaches the post-commit lifecycle

`SetChatSessionArchived` discards the rows returned by `CancelAgentTasksByChatSession` and never calls `TaskService.BroadcastCancelledTasks` after commit. `DeleteChatSession` in the same file is the established pattern: keep `cancelled` from the in-tx call, then broadcast once the tx has actually persisted (`chat.go:644` → `chat.go:706`).

Without that call the rows flip to `cancelled` in the database, but none of the four things `BroadcastCancelledTasks` does happen:

- `captureTaskCancelled` — no cancellation capture for these tasks
- `ReconcileAgentStatus` — the agent can keep showing `working` after its only task is gone
- `broadcastTaskEvent(task:cancelled)` — other clients keep a stale queued/running task in the UI until something else forces a refresh
- `notifyTasksFinished` → `notifyRuntimeMayHaveWork` — the runtime isn't woken, so the next task on that runtime waits for the daemon's next poll instead of claiming right away

Worth noting the query is not limited to never-started work: `CancelAgentTasksByChatSession` matches `queued`, `dispatched`, `running`, `waiting_local_directory` and `deferred`. Cancelling a `running` task without reconciling agent status or emitting `task:cancelled` is the case where the missing broadcast is most visible.

`TestArchivingAChatSessionCancelsItsQueuedTasks` only asserts the `agent_task_queue.status` column, so it stays green with the broadcast missing. Please extend it to cover the post-commit side (agent status reconciliation and/or the cancellation event) — otherwise the class of bug this PR is fixing can come back through the notification path without any test noticing.

## Nit (not blocking on its own)

`chat_history.go:128-130` looks the binding up twice — once for `channel_type`, once again inside `noHistoryNote`. An archive landing between the two reads can produce a response that carries `channel_type: "wecom"` next to "not connected to a chat channel". Reading the channel type once and deriving both the field and the note from that single value is cheaper and can't contradict itself.

## What's already good

Cancel-before-unbind ordering matches the invariant `DeleteChatSession` relies on, the note is derived from the binding rather than a per-platform switch so it won't go blind when a channel is added, the web-only wording is preserved, and there are no schema, foreign key or cascade changes.

seacen:
Both fixed. `SetChatSessionArchived` now keeps the rows `CancelAgentTasksByChatSession` returns and hands them to `TaskService.BroadcastCancelledTasks` after the commit, exactly as `DeleteChatSession` does twenty lines below; and the empty-history response resolves the channel type once and derives both `channel_type` and the note from that single value. The existing archive test now asserts the post-commit side — agent status, `task:cancelled`, and the task token — and fails on all three when the broadcast is removed.

---

## 1. The archive cancel never reaches the post-commit lifecycle

**Fixed.** You are right, and the four consequences you list are the right four. `chat.go:568` was `if _, err := qtx.CancelAgentTasksByChatSession(...)` — the rows went nowhere. It now reads:

```go
cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
```

with `var cancelled []db.AgentTaskQueue` declared before the `if req.Archived` block, and after `tx.Commit`:

```go
h.TaskService.BroadcastCancelledTasks(r.Context(), cancelled)
```

Placed before the `EventChatSessionUpdated` publish, same order `DeleteChatSession` uses at `chat.go:706`. On the unarchive branch `cancelled` stays nil and `BroadcastCancelledTasks` is a no-op over an empty slice, so the un-archive path is unchanged.

One addition to your list, from `captureTaskCancelled`: it also calls `DeleteTaskTokensByTask`, so without the broadcast the cancelled tasks kept their `mat_` task tokens live until the 24h expiry. That is the MUL-2600 guard, and it was silently off on this path. It is now the third thing the test asserts.

Your point about the query's status set is the one that changed how I wrote the test. `CancelAgentTasksByChatSession` matches `queued, dispatched, running, waiting_local_directory, deferred`, so I put that in a comment above the call — it is the reason the rows have to travel rather than a detail of the WHERE clause — and the test now queues a `running` task alongside the `queued` one.

Two things I checked and deliberately did **not** change, in case you were about to ask:

- **No `GetAgentForClaimUpdate` before the cancel.** `DeleteChatSession` takes it (`chat.go:639`) because it also writes the `agent` row via `DeleteSystemAgentByID`, so it needs the repo-wide session → agent → task order. The archive path never touches `agent` inside the transaction; it holds session → task, which is a subsequence of the same order and cannot cycle with claim's agent → task. `ReconcileAgentStatus` does update `agent`, but post-commit and as a single statement outside any transaction of ours.
- **No `LockChatSessionForDelete`.** For the delete path that lock is load-bearing because the blocked INSERT then fails its FK check after the commit. For archive the session survives, so the lock would serialize the insert without stopping it. There is a real residual race here — see the note at the end — but the lock alone does not close it, and closing it properly is not a change that belongs in this PR.

## 2. Nit — `chat_history.go:128-130` reads the binding twice

**Fixed**, in the form you suggested. `respondChatHistory` now resolves the type once:

```go
channelType := h.sessionChannelType(r.Context(), sessionID)
writeJSON(w, http.StatusOK, ChatChannelHistoryResponse{
    ChannelType: channelType,
    Messages:    []channel.HistoryMessage{},
    Note:        noHistoryNote(channelType),
})
```

and `noHistoryNote` drops its receiver and its own lookup — it is now `func noHistoryNote(channelType string) string`, a pure function of the value the caller already has. There is now one binding read and one value, so the field and the note cannot come from different reads of it — the contradiction you describe is no longer expressible, not just unlikely.

Being straight about the test coverage on this one: I did not add a test that reproduces the interleaving. Making the two reads straddle a concurrent archive needs an injection point between them, and after the fix there is no "between them" to inject into. What the test does assert is the invariant on both legs of the real response path — web-only session gets `channel_type: ""` plus the "not connected" note, WeCom-bound session gets `channel_type: "wecom"` plus the platform note — which is a regression guard against anyone reintroducing a second lookup whose logic drifts from the first. The race window itself is closed by construction, not by a test.

## 3. Test coverage for the post-commit side

**Extended**, as you asked, and it covers all three observable effects rather than "and/or" one of them. `TestArchivingAChatSessionCancelsItsQueuedTasks` now:

- queues **two** tasks on the session, one `queued` and one `running`, and asserts both end `cancelled` — the original column assertion, now covering the case you singled out;
- sets the agent to `status = 'working'` before the archive, and asserts it is back to `'idle'` afterwards. Without the explicit `working` this assertion would be vacuous: `RefreshAgentStatusFromTasks` has a `status IS DISTINCT FROM desired.status` guard, so on an already-idle agent it returns no rows and a missing `ReconcileAgentStatus` is indistinguishable from a working one;
- subscribes to `protocol.EventTaskCancelled` on the bus, filtered to this test's own two task ids, and asserts both arrive. The send is non-blocking into a buffered channel because this suite never unsubscribes bus handlers, and a full channel in a leaked handler would deadlock a later test rather than fail this one;
- mints a `task_token` for the running task and asserts it is gone, which is the only externally visible thing `captureTaskCancelled` does.

Each failure message names the consequence rather than the mismatch. Verbatim reverse-verification output below.

I also folded the two `respondChatHistory` drive-throughs in `TestNoHistoryNoteNamesTheActualChannel` into one helper so both legs go through the handler rather than the note helper, and dropped a stray `_ = ctx` that was left over.

---

## What else I checked, and one gap this PR does not close

You framed this as a write that lands in the database while the rest of the system never learns about it. I went looking for others, mechanically: every mutating `:many` query in `pkg/db/queries` (46 of them), then every non-test call site that throws away the rows. Seven hits in `internal/handler`, and `chat.go:568` was the only real one.

The other six are clean, for the record:

- five are `SELECT ... FOR UPDATE` locks where the rows are irrelevant by construction — `ListUserAgentsByRuntimeForUpdate` at `runtime.go:949`, `runtime.go:1144` and `runtime_profile.go:409`, `LockChatSessionsBySystemRuntimeAgents` at `chat.go:1350`, `LockChatSessionsByWorkspace` at `workspace.go:791`;
- `DeleteAgentRuntimesByProfile` at `runtime_profile.go:455` looks like the same shape and is not: its fan-out comes from the per-runtime `teardowns` collected in the loop above it, plus `EventDaemonRegister` after commit. What is stale there is the query's own comment — "Returns the deleted rows so the caller can broadcast / audit" — which describes a contract the caller stopped using. Worth a one-line comment fix by whoever is next in that file; not a bug.

I also checked `ArchiveAgent` by hand, since it is the other bulk cancel in the same neighbourhood. `agent.go:2037` keeps the rows and calls `CaptureCancelledTasks`, deliberately skipping the per-task events with a documented reason ("the agent:archived event below already triggers a full active-tasks invalidation on every connected client"). Fine as is.

The one thing I did find is adjacent rather than identical, and it limits what this PR actually guarantees, so it belongs in the record.

**A task enqueued concurrently with the archive survives the cancel.** `DeleteChatSession` takes `LockChatSessionForDelete` (`FOR UPDATE`) as the first statement of its transaction precisely so a concurrent `SendChatMessage` cannot land a task between the cancel and the delete. The archive path has no such lock — and its own `UPDATE chat_session SET status = ...` does not substitute for one. The `SET` touches no unique-index column, so it takes `FOR NO KEY UPDATE`, which does not conflict with the `FOR KEY SHARE` an `agent_task_queue` INSERT takes on the parent row during FK validation.

I confirmed that on a real database rather than reasoning about it. With an uncommitted `INSERT INTO agent_task_queue` referencing the session held open in another connection, and `lock_timeout = 3s`:

```
--- now try the plain archive UPDATE with a 3s lock timeout ---
                  id                  |  status
--------------------------------------+----------
 51479a60-76c8-4436-aeb5-9f5cb960a5dd | archived
(1 row)

--- now try SELECT ... FOR UPDATE with a 3s lock timeout ---
ERROR:  canceling statement due to lock timeout
CONTEXT:  while locking tuple (0,4) in relation "chat_session"
```

So the archive commits straight through the concurrent insert. `SendChatMessage` reads `session.Status` at `chat.go:787`, outside any transaction, and then does several more round trips before enqueuing — the window is not instantaneous.

I did not fix it here. Taking `FOR UPDATE` in the archive transaction would serialize the insert but not stop it, because unlike the delete path there is no FK failure waiting on the other side — the session still exists, so the blocked insert simply succeeds after we commit and lands a `queued` task on an archived, unbound session. The two changes that actually close it are (a) making `SendChatMessage` re-read the session status inside a transaction holding that lock, which is the same "the lock alone is not sufficient" note `LockChatSessionForRuntimeBind` already carries, or (b) having `ClaimAgentTask` read `chat_session.status`, which changes the query every runtime polls to claim work. Both are their own PR. Happy to open one if you want it tracked; say which of the two you would rather have.

Worth being clear that this is not a regression from this PR — the archive endpoint has had this window since MUL-4372 dropped the binding in the same transaction. This PR narrows the failure a lot without eliminating it: before, every queued task survived an archive; now only one enqueued inside that window does.

---

## Verification

Own worktree off `fix/wecom-archive-cancels-queued-tasks`, against a throwaway Postgres (`pgvector/pgvector:pg17`) with all migrations applied.

On the base, so you know what my local numbers cover: the branch still sits on `47f6e970f`, seven commits behind `main` at `0ab271235`, and merges clean. Those seven land in `internal/integrations/wecom`, `wecom_web.go`, `handler.go` and `channel.sql`, and touch none of the three files here, so I did not rebase — my local runs are against the branch, and CI's backend job is what covers the merged tree.

`go build ./...`, `go vet ./...`:

```
BUILD=0
VET=0
```

`gofmt -l` on the three changed files:

```
GOFMT_CLEAN
```

(`gofmt -l .` over `server/` reports 27 files on this toolchain, Go 1.26.5 against CI's 1.26.1 — the same 27 before and after this change, none of them mine.)

`go test -race -count=1` on the packages this PR touches:

```
ok  	github.com/multica-ai/multica/server/internal/handler	30.468s
ok  	github.com/multica-ai/multica/server/internal/service	5.924s
```

The two new tests, run explicitly so it is clear they executed rather than skipped on a missing database:

```
=== RUN   TestArchivingAChatSessionCancelsItsQueuedTasks
--- PASS: TestArchivingAChatSessionCancelsItsQueuedTasks (0.06s)
=== RUN   TestNoHistoryNoteNamesTheActualChannel
--- PASS: TestNoHistoryNoteNamesTheActualChannel (0.02s)
PASS
ok  	github.com/multica-ai/multica/server/internal/handler	2.038s
```

**Reverse-verification 1 — remove `BroadcastCancelledTasks` and restore the discarding call** (i.e. the code you reviewed, verbatim):

```
=== RUN   TestArchivingAChatSessionCancelsItsQueuedTasks
    chat_archive_cancel_test.go:115: agent status after archiving its only conversation = "working", want idle — the agent keeps showing as working in the UI with no task left to finish it
    chat_archive_cancel_test.go:129: task:cancelled emitted for 0 of 2 tasks — every other client keeps the cancelled turn on screen until something else forces a refresh
    chat_archive_cancel_test.go:138: cancelled task still has 1 live task token(s) — the agent process can keep calling the API back for a turn that was cancelled
--- FAIL: TestArchivingAChatSessionCancelsItsQueuedTasks (0.03s)
FAIL
FAIL	github.com/multica-ai/multica/server/internal/handler	0.970s
```

All three post-commit assertions fail; the `status = 'cancelled'` column assertion still passes, which is exactly the blind spot you identified in the original test.

**Reverse-verification 2 — restore the hardcoded note in `respondChatHistory`:**

```
=== RUN   TestNoHistoryNoteNamesTheActualChannel
    chat_archive_cancel_test.go:202: a WeCom-bound session was told it is not on a channel: "This conversation is not connected to a chat channel, so there is no channel history to read."
--- FAIL: TestNoHistoryNoteNamesTheActualChannel (0.03s)
FAIL
FAIL	github.com/multica-ai/multica/server/internal/handler	0.922s
```

Both reverts restored; the package is green again.

A full `go test ./... -race` over the server tree turned up two failures outside this PR's packages, and I checked both rather than assuming:

- `internal/daemon` `TestProbeAgentCLIs_QoderResolvesViaLoginShell` — fails identically with my changes stashed, at the unmodified branch head. It resolves through the login shell and picks up a real `qodercli` on this machine instead of the fake PATH. Local environment, not the code.
- `pkg/agent` `TestCursorExecuteEmitsTerminalResultText` — `status = "timeout", want completed` after its 5s budget, during a saturated full-tree race run; passes in isolation. Load flake.

CI on the pushed head (`a131df4f9`), all nine green, no reruns and nothing to explain away:

```
backend                     pass   5m22s
changes                     pass   9s
frontend                    pass   4s
frontend-build              pass   6s
frontend-test               pass   3s
installer (macos-latest)    pass   16s
installer (ubuntu-latest)   pass   6s
installer (windows-latest)  pass   19s
windows-execenv             pass   1m51s
```

Still `MERGEABLE`.

server/internal/handler/chat.go
```diff
@@ -552,7 +552,33 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
...
+		// This is also what archiving means: the owner said they are done
+		// with this conversation. DeleteChatSession has always cancelled for
+		// the same reason (see the CancelAgentTasksByChatSession call below),
+		// and chat.sql's own comment already assumes archive does too.
+		//
+		// The query is not limited to never-started work: it also matches
+		// dispatched, running, waiting_local_directory and deferred, which is
+		// why the returned rows have to reach the post-commit broadcast below
+		// rather than being discarded.
+		cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
```
Bohan-J: This cancel is unconditional — it runs for every archived session, including plain web chats that never had a channel binding. The bug in the description (a late answer landing in a group room after the binding was dropped) can only happen to a channel-bound session.

Suggest reading the binding here, before `DeleteChannelChatSessionBindingBySession` below erases it, and cancelling only when one existed. `GetChannelChatSessionBindingBySessionAny` is already used by the other half of this PR.

Why it matters for web chats: the UI models Stop and Archive as different actions (`chat-thread-list.tsx:364-377` renders one or the other), Stop is confirmed and Archive is not, and the Stop path restores the user's typed prompt via `cancelled_chat_message` / `restore_to_input`. This path has no draft restore, and unarchiving cannot undo a cancel — so archiving a running web chat silently destroys the prompt.

server/internal/handler/chat_history.go
```diff
@@ -174,3 +180,35 @@ func parseHistoryLimit(raw string) int {
...
+// sessionChannelType names the platform behind a session, or "" when there is
+// none. Channel-agnostic on purpose: a per-platform lookup here would go blind
+// the next time a channel is added, which is exactly how the note above came
+// to be wrong.
+func (h *Handler) sessionChannelType(ctx context.Context, sessionID pgtype.UUID) string {
+	if h.Queries == nil || !sessionID.Valid {
+		return ""
+	}
+	binding, err := h.Queries.GetChannelChatSessionBindingBySessionAny(ctx, sessionID)
+	if err != nil {
```
Bohan-J: Every error becomes "no channel" here. A missing row should, but a transient DB error will produce a 200 carrying exactly the misleading "not connected to a chat channel" note this PR is removing.

Consider returning `(string, error)` and mapping only `pgx.ErrNoRows` to the empty channel type.

Bohan-J (changes requested):
Thanks for the quick turnaround — both items from the previous round are properly closed on `a131df4`:

- `chat.go:577` now keeps the cancelled rows and `chat.go:601` broadcasts them after commit, matching `DeleteChatSession`. The comment explaining why the returned rows must reach the broadcast is a nice touch.
- `chat_history.go:131-135` reads the binding once and derives both `channel_type` and the note from that single value.
- The expanded tests cover queued + running, the agent returning to idle, and both cancellation events.

The channel-history half of this PR is good to go. The remaining blocker is a **product scope** question on the cancel half, not an implementation defect.

## Blocking — the cancel applies to every chat session, not just channel-bound ones

`SetChatSessionArchived` cancels unconditionally on `archived=true`; there is no check for whether the session ever had a channel binding. The bug this PR describes is specifically about a channel-bound session posting a late answer into a group room after its binding was dropped. For a plain web chat there is no room, no adapter, and no late answer — the cancel there is a new product behaviour riding along with the fix.

That new behaviour conflicts with how the existing UI defines these two actions:

- `chat-thread-list.tsx:364-377` — a session row shows **Stop** *or* **Archive**, never both: `isRunning ? Stop : Archive`. They are modelled as different actions.
- Stop is `danger` and goes through a confirmation dialog (`chat-thread-list.tsx:306-317`). Archive fires immediately with no confirmation.
- Archive is reversible by design: unarchive exists, and `chat-session-header.tsx:169-186` offers hard delete *only after* a chat is archived. Archive is the "put it away" step; delete is the destructive one.
- The header's Archive item has no `isRunning` gate, so a user can archive a running session from the dropdown and silently kill the run — bypassing the confirmation the product requires for the Stop button that does the same thing.

There is also a concrete difference in what the user loses. The Stop button calls `cancelTaskById`, whose result carries `cancelled_chat_message` / `restore_to_input` and puts the typed prompt back into the input box (`chat-thread-list.tsx:164-171`). The archive path calls `CancelAgentTasksByChatSession` directly, which is a bare UPDATE; `BroadcastCancelledTasks` does capture, status reconciliation, the event and the runtime wakeup, but no draft restore. So archiving a running web chat destroys the prompt with no way to get it back, while the Stop button hands it back — and unarchiving cannot undo the cancel either.

**Preferred fix: only cancel when the session actually had a channel binding before the archive.** That solves the reported bug exactly — binding dropped, so no late answer into the room — and leaves plain web chat archiving as the reversible organizing action it is today. Implementation note: read the binding *before* `DeleteChannelChatSessionBindingBySession` at `chat.go:582`, since that call is what erases the evidence. `GetChannelChatSessionBindingBySessionAny` already exists and is used by the other half of this PR.

If the intent really is a product-wide "archive means stop", that is a defensible decision but needs to land as one: gate it on there being an active task, reuse the same confirmation the Stop button uses, and rename the action accordingly ("Stop and archive"). That is a bigger change than this PR, and I would rather not have it arrive as a side effect of a channel bug fix.

## Nit (not blocking)

`chat_history.go:205-214` maps every error from the binding lookup to "no channel". A missing row should behave that way, but a transient database error currently returns 200 with the same incorrect "not connected to a chat channel" note this PR set out to remove. Returning `(string, error)` and treating only `pgx.ErrNoRows` as "no channel" would keep the fix honest under failure.

seacen:
Taken your preferred fix: the cancel now runs only when the session actually had a channel binding. The unconditional version was scope we did not intend — the bug is a bound session answering into a room after its binding was dropped, and a plain web chat has neither, so cancelling there was a product change that rode along with a channel fix. Pushed as `32f4202`.

## The narrowing

`SetChatSessionArchived` reads `GetChannelChatSessionBindingBySessionAny` **before** `DeleteChannelChatSessionBindingBySession`, per your note — that delete is what erases the evidence, and the ordering is now guarded by a test (below). Three branches:

- binding found → cancel exactly as before, and the rows still reach `BroadcastCancelledTasks` after the commit;
- `pgx.ErrNoRows` → nothing cancelled, and archiving a web chat stays the reversible organizing action it is today;
- any other error → 500, the archive fails. "Not bound" and "cannot tell" must not take the same branch: a transient database error that silently skipped the cancel would reopen the exact bug this guard exists for.

So the fix now covers only the reported failure. The binding is dropped in the same transaction, and the turn that would have answered into the room after it is cancelled in that same transaction.

## The Stop / Archive distinction

Checked all three points; each holds at the lines you cite.

- `chat-thread-list.tsx:364-377` is `isRunning ? Stop : Archive`, never both. Stop is `danger` and goes through the `ConfirmRow` at :306-317; Archive calls `onArchive(session)` immediately.
- `chat-session-header.tsx:169-186` puts unarchive and hard delete in the `isArchived` branch, under the comment "Hard delete is offered only once a chat is archived." Archive is reversible by construction.
- The prompt loss is the sharpest of the three, and it is worse than "no restore": there are two restore paths and the archive path used neither. `cancelTaskById` → `CancelTaskByUser` → `finalizeCancelledChatMessage`, which either answers synchronously with `restore_to_input` (consumed at `use-chat-task-actions.ts:87-109`) or writes a durable `chat_draft_restore` row (`task.go:2480`, `:2709`) that `use-chat-draft-restore.ts` picks up. `BroadcastCancelledTasks` (`task.go:2168-2175`) does `captureTaskCancelled`, `ReconcileAgentStatus`, `task:cancelled` and `notifyTasksFinished`, and neither of those two.

On the header's Archive item having no `isRunning` gate: agreed it is pre-existing and not this PR's to fix, and the narrowing does make it harmless. `doArchive` (`chat-session-header.tsx:99-102`) is still reachable while a turn is running, but a session archived from there with no binding is a web chat, which now cancels nothing — the dropdown no longer does silently what the Stop button asks for a confirmation to do. A bound session archived from there does still cancel, which is the behaviour we want.

Not taking the product-wide "archive means stop". If we ever want it, it should arrive as its own change with the gate, the confirmation and the rename, not out of this one.

## Verification

The tests are one fixture archived twice: same agent, same session, one queued turn and one already running, agent marked `working`, a live `mat_` token. The only difference between the two is whether the session is bound to a channel first.

`TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks` is the previous test with the binding it was always describing. Worth flagging what it turned out to be asserting: it never created a binding. Its failure message read "its answer goes to the room while the brief says private" while the session it archived was a plain web chat, so what it actually pinned was the unconditional behaviour under the bound bug's name. It now binds a WeCom group room first and keeps every post-commit assertion from the last round — both rows cancelled, agent back to `idle`, `task:cancelled` for both, token revoked — plus one new one: the binding is still deleted, since the read moved ahead of the delete.

`TestArchivingAWebOnlyChatSessionLeavesItsTasksRunning` is the other direction: both turns still in flight, agent still `working`, no `task:cancelled`, token still live, and the session still `archived` (narrowing the cancel must not turn into skipping the status flip).

`TestNoHistoryNoteNamesTheActualChannel` is untouched. `TestSetChatSessionArchived_ClearsChannelBinding` and `_TogglesStatus` never asserted anything about cancellation, so neither changed meaning.

Reverse-verified in both directions against a throwaway Postgres.

Drop the cancel entirely, i.e. the behaviour before this PR:

```
--- FAIL: TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks (0.07s)
    chat_archive_cancel_test.go:58: queued task status after archiving a channel-bound session = "queued", want cancelled — the daemon will still run this turn, and with the channel binding already gone its answer goes to the room while the brief says private
--- PASS: TestArchivingAWebOnlyChatSessionLeavesItsTasksRunning (0.03s)
```

Make the cancel unconditional again, i.e. the behaviour you blocked:

```
--- PASS: TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks (0.06s)
--- FAIL: TestArchivingAWebOnlyChatSessionLeavesItsTasksRunning (0.03s)
    chat_archive_cancel_test.go:110: queued task status after archiving a web-only chat = "cancelled", want "queued" — there is no room for a late answer to reach here, and cancelling destroys the user's typed prompt: this path writes no chat_draft_restore row and returns no restore_to_input, so the text is gone and unarchiving cannot bring it back
```

And your implementation note is guarded as well — move the binding read after `DeleteChannelChatSessionBindingBySession` and the bound leg fails, because by then the row is gone and the lookup answers `ErrNoRows`:

```
--- FAIL: TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks (0.03s)
    chat_archive_cancel_test.go:58: queued task status after archiving a channel-bound session = "queued", want cancelled — the daemon will still run this turn, and with the channel binding already gone its answer goes to the room while the brief says private
--- PASS: TestArchivingAWebOnlyChatSessionLeavesItsTasksRunning (0.02s)
```

On the fix as pushed: `go build ./...` and `go vet ./...` clean, `gofmt -l` clean on both touched files, and

```
ok  github.com/multica-ai/multica/server/internal/handler  33.286s
ok  github.com/multica-ai/multica/server/internal/service   5.498s
```

from `go test -race ./internal/handler/ ./internal/service/ -count=1`. Rebased onto main for #6592, #6593 and #6602; all nine CI checks green on `32f4202`.

The nit on `chat_history.go:205-214` is a fair catch. I would rather send it as its own change than widen this one, but say the word and I will fold it in here.

server/internal/handler/chat.go
```diff
@@ -552,7 +561,55 @@ func (h *Handler) SetChatSessionArchived(w http.ResponseWriter, r *http.Request)
...
+			// web chat while the channel adapter still holds the chat id and
+			// posts the answer into the group. The person who archived the
+			// conversation never sees it happen. Cancel in the same tx that
+			// drops the binding, as DeleteChatSession does.
+			//
+			// The query is not limited to never-started work: it also matches
+			// dispatched, running, waiting_local_directory and deferred, which
+			// is why the returned rows have to reach the post-commit broadcast
+			// below rather than being discarded.
+			cancelled, err = qtx.CancelAgentTasksByChatSession(r.Context(), session.ID)
```
Bohan-J, line 592: This cancels whatever exists *now*, but the channel path creates the task ~3s later. `router.go:471-486`: the synchronous outcome is `OutcomeIngested` with no TaskID — the row is created at debounce flush.

So a message persisted just before the archive has no task to cancel here, and `flushChatRun` (`router.go:663-677`) then calls `EnqueueChatTask` on the session this tx just archived. `EnqueueChatTask` (`task.go:1592`) checks `agent.ArchivedAt` and `agent.RuntimeID` but never `chatSession.Status`, so the task is inserted and claimed.

The fix already exists one file over: `task.go:1887-1895` takes `LockChatSessionForRuntimeBind`, re-reads the session, and returns `ErrChatSessionArchived` when `Status != "active"`. Doing the same inside `EnqueueChatTask`'s transaction closes both interleavings and keeps the `chat_session -> agent_task_queue` lock order this handler already uses.

server/internal/handler/chat_history.go
Bohan-J: Carried over from the last round: every error here becomes "no channel", so a transient DB error returns 200 with the same misleading note this PR is removing. Only `pgx.ErrNoRows` should map to the empty channel type.

Bohan-J (changes requested):
The scope narrowing on `32f4202` is exactly right, and done more carefully than I asked for:

- `chat.go:573` reads the binding before the delete that erases it, so the "was this session on a channel?" question is answered while it is still answerable.
- The three-way switch is the right shape. Treating a non-`ErrNoRows` lookup failure as a 500 rather than silently falling through to "web-only" is the detail that matters — guessing there would quietly reopen the original bug.
- The channel-bound branch keeps the in-tx cancel plus the post-commit `BroadcastCancelledTasks`, so nothing regressed from the previous round.
- Tests now pin both directions, including that a web-only chat keeps its task, its agent status and its token.

One blocker left, on a path the tests do not reach.

## Must-fix — a task can be created *after* the archive commits

The channel run trigger is debounced by 3s (`DefaultChatRunBatchWindow`, MUL-2968). The chat message is persisted immediately, but the task row is only created when the window flushes — `router.go:471-486` says so explicitly: "The synchronous outcome is `OutcomeIngested` with no TaskID — the task row is created at flush."

That opens this window:

1. A group message is persisted into the session. No task row exists yet.
2. The user archives. `chat.go:589` cancels what it finds — nothing — then drops the binding and commits.
3. The debounce timer fires. `flushChatRun` (`router.go:663-677`) reloads the session by id and calls `EnqueueChatTask`.
4. `EnqueueChatTask` (`task.go:1592`) checks `agent.ArchivedAt` and `agent.RuntimeID`, but never looks at `chatSession.Status`. It inserts a fresh task on the session that was just archived, and `ClaimAgentTask` does not read session status either, so the daemon runs it.

The result is a run the archive was supposed to prevent: it burns runtime and quota on a conversation the user closed, writes new assistant messages into an archived chat, and flips the agent back to `working`. Worse for this PR specifically, the daemon brief resolves the room through `GetChannelChatSessionBindingBySessionAny` (`daemon.go:2187`) — the binding is gone by then, so the run is described to the agent as a private web chat. That is precisely the misattribution the other half of this PR exists to fix.

One correction to how this has been characterised elsewhere: the late answer most likely does **not** reach the group. The outbound senders resolve their destination through `GetChannelChatSessionBindingBySession` (`slack/outbound.go`, `wecom/outbound.go`, `dingtalk/outbound.go`), and that row was deleted in step 2, so there is no route left to send on. So this is not "the original bug is still live" — it is a wasted, mis-briefed run against a closed conversation. Still worth blocking on, but I would rather the PR fix the right thing.

**The guard already exists in this repo, just not on this path.** `task.go:1564` defines `ErrChatSessionArchived` — "signals that a direct-chat send lost a race with archiving the session and therefore must not persist a new turn" — and the web-chat send path uses it at `task.go:1887-1895`:

```go
if _, err := qtx.LockChatSessionForRuntimeBind(ctx, session.ID); err != nil { ... }
currentSession, err := qtx.GetChatSession(ctx, session.ID)
if currentSession.Status != "active" {
    return ErrChatSessionArchived
}
```

`OpenMikaOnboardingChat` does the same at `task.go:2040-2047`. `EnqueueChatTask` is the one enqueue path without it, and it is the one with a 3-second window in front of it. Taking the same lock in the same order inside the existing `EnqueueChatTask` transaction makes both interleavings safe: enqueue-then-archive is caught by the cancel, archive-then-enqueue is refused. The lock order stays `chat_session -> agent_task_queue`, matching this handler and the delete path, so no new deadlock edge.

`flushChatRun` already handles this cleanly — it clears the typing indicator before the error switch, and an unrecognised error falls to the `default:` log branch. No notice gets posted into a room the archive just unbound, which is the behaviour you want.

Please also add the regression test for it. Every archive test currently starts from a task that already exists, so none of them can fail on this: the case to cover is *message persisted, debounce not yet flushed → archive → flush*, asserting no new task row and no `task:queued`.

## Nit (carried over, still not blocking)

`chat_history.go:205-214` still maps every binding-lookup error to "no channel". A missing row should behave that way; a transient database error currently produces a 200 carrying the exact "not connected to a chat channel" note this PR removes. `(string, error)` with only `pgx.ErrNoRows` mapped to empty would close it.

seacen:
Fixed, with one deliberate departure from the lock you named. Everything else is as specified.

## The enqueue guard

`EnqueueChatTask` now takes the chat-session lock as its transaction's first statement, re-reads the row under it, and returns `ErrChatSessionArchived` when the status is not `active` — the same shape as the send path and `OpenMikaOnboardingChat`. Lock order stays `chat_session -> agent_task_queue`.

The departure: the lock is a new `LockChatSessionForEnqueue`, `FOR NO KEY UPDATE`, not `LockChatSessionForRuntimeBind`'s `FOR UPDATE`.

Taking `FOR UPDATE` there hangs `TestEnqueueChatTaskDefersWhenMediaMessageCommitsDuringEnqueue` until the test binary's deadline:

```
panic: test timed out after 45s
		TestEnqueueChatTaskDefersWhenMediaMessageCommitsDuringEnqueue (45s)
github.com/multica-ai/multica/server/internal/service.TestEnqueueChatTaskDefersWhenMediaMessageCommitsDuringEnqueue.func2()
github.com/multica-ai/multica/server/internal/service.(*TaskService).EnqueueChatTask(...)
	server/internal/service/task.go:1692 +0x170c
```

That test injects an `INSERT INTO chat_message` from a second connection in the middle of the enqueue transaction. `FOR UPDATE` conflicts with the `FOR KEY SHARE` an FK insert takes on its parent row — the same conflict `LockChatSessionForDelete`'s comment relies on to stop a send slipping a task in between its cancel and its delete. So the injected append waits on us while we wait on it.

The hang is only how it surfaced. The conflict itself is real in production and points the wrong way on this path: every inbound channel message is an `INSERT INTO chat_message` FK'd to this same row, and this lock is held for the whole enqueue. Under `FOR UPDATE`, a group's next message blocks behind the previous message's flush. The send path can afford that — one user, one session, and it does the insert itself — but a room cannot.

`FOR NO KEY UPDATE` keeps the serialisation the guard needs and drops the part it does not. Measured on pg17, holding the row and probing from a second connection with `lock_timeout`:

| holder | probe | result |
|---|---|---|
| `FOR UPDATE` | `INSERT INTO chat_message` | blocked |
| `FOR NO KEY UPDATE` | `INSERT INTO chat_message` | 102 ms, through |
| `FOR NO KEY UPDATE` | `UPDATE chat_session SET status='archived'` | blocked |
| `UPDATE chat_session SET status='archived'` | `SELECT … FOR NO KEY UPDATE` | blocked |

`SetChatSessionArchived` updates `status`, not a key column, so its row lock is `FOR NO KEY UPDATE`, which conflicts with ours in both directions. The delete path's `FOR UPDATE` conflicts with it too. Archive-then-enqueue is refused by the status re-read; enqueue-then-archive blocks until we commit and is then caught by the archive's cancel. Inbound ingestion is not serialised behind either.

The lock mode is pinned by a test rather than asserted: `TestEnqueueChatTaskLocksOutAConcurrentArchiveButNotInboundMessages` runs both probes from inside the enqueue transaction and requires the archive to block and the append not to.

One consequence worth naming: the refused transaction rolls back, so the message that armed the window stays `task_id IS NULL`. It is not lost — if the user unarchives and writes again, `LinkUnownedChannelChatMessagesToTask` sweeps it into that turn's batch under its usual "already answered" boundary. Dropping it instead would need machinery this PR does not have, and resuming the conversation is the reading I would expect a user to want.

## flushChatRun

Checked rather than trusted, and it holds. `clearTyping` runs before the `switch`, and `ErrChatSessionArchived` matches neither `case`, so it lands on `default:` and only logs. Nothing is posted into a room the archive just unbound.

I did not want that to rest on reading, since it is one `case` away from breaking silently, so `TestRouter_FlushSessionArchived_ClearsTypingAndSaysNothing` now pins both halves: the typing indicator is cleared, and the only reply the message produces is its ingest ACK. Adding a `case` that emits a notice fails it:

```
--- FAIL: TestRouter_FlushSessionArchived_ClearsTypingAndSaysNothing (0.00s)
    router_test.go:1147: archived-session flush posted "agent_archived" into a room the archive had already unbound
```

## The description was wider than the bug

You are right, and it was in four places, not one. The outbound senders resolve their destination through the binding row the archive deletes, so the late answer has no route left. What actually happens is a wasted, mis-briefed run against a closed conversation: runtime and quota spent, assistant messages written into an archived chat, the agent flipped back to `working`, and the brief calling it a private web chat. Corrected in:

- the PR description (both the bullet claiming the adapter posts into the group, and the quoted test failure under **Tests**)
- `chat.go`, the comment on the channel-bound branch of the archive switch
- `chat_archive_cancel_test.go`, the file header
- `chat_archive_cancel_test.go`, the failing assertion of `TestArchivingAChannelBoundChatSessionCancelsItsQueuedTasks`, plus the `bindToChannel` comment, which said the adapter still knows the room

The commit messages from earlier rounds still carry the old wording; the squash lands the description, so I left the history alone rather than force-pushing a reword mid-review.

## The nit

Taken. `sessionChannelType` returns `(string, error)` with only `pgx.ErrNoRows` mapped to empty; `respondChatHistory` reports anything else instead of answering 200. The third leg of `TestNoHistoryNoteNamesTheActualChannel` drives it with a cancelled context, which is the cheapest honest way to make that read fail:

```
--- FAIL: TestNoHistoryNoteNamesTheActualChannel (0.04s)
    chat_archive_cancel_test.go:533: a failed binding read answered 200 {"channel_type":"","messages":[],"note":"This conversation is not connected to a chat channel, so there is no channel history to read."} — the agent is told it is in a web-only conversation because a query failed
```

## The regression test you asked for

`TestArchivingInsideTheDebounceWindowRefusesTheLateEnqueue`: bind the session, append a channel message, archive through the real endpoint while no task exists, then flush. Asserts no task row on the session, no `task:queued`, and the sentinel. Removing the status check:

```
--- FAIL: TestArchivingInsideTheDebounceWindowRefusesTheLateEnqueue (0.07s)
    chat_archive_cancel_test.go:196: the debounced flush created 1 task(s) on a session archived while its window was open — the daemon claims and runs that turn on a closed conversation: runtime and quota spent, assistant messages written into an archived chat, the agent flipped back to 'working', and with the binding already deleted the run is briefed as a private web chat
```

Removing the lock as well, so the status read no longer holds:

```
--- FAIL: TestEnqueueChatTaskLocksOutAConcurrentArchiveButNotInboundMessages (0.11s)
    task_channel_media_race_test.go:209: an archive landing mid-enqueue got <nil>, want to be blocked on the chat_session lock — it would otherwise commit while this task row is still invisible, cancel nothing, and leave a queued turn on a closed conversation
```

Changing the mode to `FOR UPDATE`:

```
--- FAIL: TestEnqueueChatTaskLocksOutAConcurrentArchiveButNotInboundMessages (1.13s)
    task_channel_media_race_test.go:212: the room's next message could not be appended during an enqueue: ERROR: canceling statement due to lock timeout (SQLSTATE 55P03) — inbound ingestion must not wait behind the debounced flush of the message before it
```

`TestEnqueueOnAnActiveSessionStillWorks` is the control for the live path: an ordinary flush on a non-archived session still creates its task and still seals the waiting message into it.

`go build ./...`, `go vet ./...` and `gofmt -l` are clean, and `go test -race ./internal/handler/ ./internal/service/ ./internal/integrations/channel/engine/` passes.

---

Bohan-J (approved):
Final review on `117f3fa`. Both blockers from the earlier rounds are closed, and the carried-over nit is fixed. Approving.

**The enqueue guard is right, and the lock mode is the interesting part.** `LockChatSessionForEnqueue` takes `FOR NO KEY UPDATE` rather than the `FOR UPDATE` the send / delete / draft paths take, and the reasoning in the query comment is correct: `FOR UPDATE` conflicts with the `FOR KEY SHARE` an FK insert takes on its parent, so using it here would make a room's next inbound message wait behind the previous message's enqueue. `FOR NO KEY UPDATE` still conflicts with `SetChatSessionArchived`'s `UPDATE` and with the delete path's lock, which is all the guard needs.

I checked the deadlock question independently, since this adds a lock the enqueue path did not previously hold. The inbound append (`engine/session.go`) runs `CreateChatMessage` then `TouchChatSession` in one transaction — chat_message before chat_session — which is the reverse of the enqueue's chat_session before chat_message. That is only a cycle if both directions can wait:

- inbound → enqueue: real. `TouchChatSession`'s `UPDATE chat_session` takes `FOR NO KEY UPDATE` and will wait on an in-flight enqueue.
- enqueue → inbound: not reachable. `LinkUnownedChannelChatMessagesToTask` updates chat_message rows, and under READ COMMITTED an uncommitted `INSERT` from the concurrent inbound transaction is invisible to that scan, so it is never waited on.

One-directional wait, no cycle. The "no new deadlock edge" claim in the comment holds, and the residual cost is a brief serialisation of `TouchChatSession` behind a short enqueue transaction.

Also confirmed: no migration, no new foreign key or cascade, no index changes. All nine CI checks pass on this head.

`chat_history.go` now returns `(string, error)` and only maps `pgx.ErrNoRows` to "no channel", so a connection blip no longer produces a 200 carrying the exact note this PR set out to delete. Good symmetry with the archive path's refusal to guess about the same read.

Two nits, neither blocking, both fine as follow-ups:

- `task_channel_media_race_test.go:141-192` — the test proves the FK `INSERT` is not blocked, which is the property that matters, but the name and comment claim the lock does not block "the room's next message". The real inbound path also calls `TouchChatSession`, which does conflict. Worth narrowing the wording to what is actually asserted, and stating that the `Touch` serialisation is accepted.
- `flushChatRun` has no case for `ErrChatSessionArchived`, so an expected terminal outcome lands in the `default:` branch and is logged at ERROR, and `forceFresh` is re-marked into `pendingFresh` for a session that will never run again. A dedicated case would remove the noise and the stale entry.

Nice work on the iterations — particularly refusing to guess on a failed binding read rather than defaulting to "web-only", which is what kept the narrowing honest.

PR #6596, opened 2026-08-07
MUL-5893 fix(lark): three defensive fixes to stored media filenames

seacen:
Three small fixes to `lark/media_ingest.go`, found while auditing the WeCom adapter against this one.

**None is a path traversal** — `path.Base` already resolves the path before any of this runs. All three reach the stored object's *display name*, which surfaces in `Content-Disposition` and wherever the file is later re-saved.

| | Before | After |
|---|---|---|
| `cleanFilename("..")` | `".."` | falls through to the generated name |
| `mediaExtension("application/pdf")` | depends on the host's `/etc/mime.types` | `".pdf"` |
| 3 photos in one message | 3 objects all named `feishu-image-<msg>.jpg` | `…​.jpg`, `…​-2.jpg`, `…​-3.jpg` |

## Details

1. **`cleanFilename` returned `".."` and `"..."` unchanged.** `path.Base` resolves a path but hands a dot-only name straight back, and the guard only covered `"."` and `"/"`. `"..hidden.txt"` and `"../../etc/passwd"` → `"passwd"` behave exactly as before.

2. **No case for `application/pdf`**, so it fell through to `mime.ExtensionsByType` — which reads `/etc/mime.types`, a file slim container images do not ship. On those images a PDF was stored with **no extension at all**. Now pinned alongside the image and video types that were already pinned for the same reason.

3. **The generated filename omitted the attachment index.** Every resource on one message shares a `MessageID`, so a three-photo send produced three identically-named objects. A platform-provided filename still wins.

## Tests

Four. Reverting fixes 1 and 3 fails them:

```
cleanFilename("..") = ".." — a name made only of dots is not a filename
three attachments produced 1 distinct names (map[feishu-image-om_abc123.png:3])
```

**Fix 2 has no failing-first proof, and I want to be straight about that**: reverting it does *not* fail on a developer machine, because macOS and most distros ship `/etc/mime.types` and the fallback succeeds. That is precisely the condition the fix is for, so it cannot be reproduced locally. The test pins the behaviour going forward; it does not demonstrate the bug.

### DIFF
--- a/server/internal/integrations/lark/media_filename_test.go
+++ b/server/internal/integrations/lark/media_filename_test.go
@@ -1,0 +1,66 @@
+package lark
+
+// media_filename_test.go — the stored object's display name. Not a path
+// traversal (path.Base already resolves the path), but a name that reaches
+// Content-Disposition and anywhere the file is later re-saved.
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/multica-ai/multica/server/internal/integrations/channel"
+)
+
+func TestCleanFilenameRejectsDotOnlyNames(t *testing.T) {
+	for _, in := range []string{"..", "...", ".", "  ..  ", "../", "a/../.."} {
+		if got := cleanFilename(in); strings.Trim(got, ".") == "" && got != "" {
+			t.Errorf("cleanFilename(%q) = %q — a name made only of dots is not a filename", in, got)
+		}
+	}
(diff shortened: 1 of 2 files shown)


multica-eve (approved):
Thank you for the thoughtful fix and for clearly documenting the edge cases and test limitations. The scoped filename handling changes look good, and the regression coverage is helpful. Approved!

PR #6597, opened 2026-08-07
fix(metrics): forbid installation_id as a metric label (MUL-5894)

seacen:
`forbiddenMetricLabels` already blocks `workspace_id`, `session_id`, `task_id`, `agent_id`, `runtime_id` and `ip` — every identifier whose series count grows with **tenants** rather than with the deployment.

`installation_id` is the same class and was missing.

It is also the one most likely to be added by accident. The channel adapters (slack, lark, dingtalk, wecom) all carry an installation id at every metric call site, so it is right there whenever somebody adds a counter — leaving it off the list is a matter of time rather than of luck. And by the time it shows up, the series are already in the operator's Prometheus and cannot be taken back out of the retention window.

Found while auditing a WeCom metrics change that deliberately refuses this label in its doc comment. The refusal should be enforced by the guard, not by each author remembering.

## Test

Asserts the channel identifiers are all covered. Removing the entry:

```
installation_id is not forbidden — a per-tenant identifier will eventually be
used as a metric label and multiply the series count by the tenant count
```

No metric currently uses it, so nothing changes at runtime — this is the guard closing before the mistake, not after.

### DIFF
--- a/server/internal/metrics/labels.go
+++ b/server/internal/metrics/labels.go
@@ -90,13 +90,19 @@ var businessMetricLabels = map[string][]string{
 
 var forbiddenMetricLabels = map[string]struct{}{
 	"workspace_id": {},
-	"user_id":      {},
-	"agent_id":     {},
-	"task_id":      {},
-	"issue_id":     {},
-	"runtime_id":   {},
-	"session_id":   {},
-	"ip":           {},
+	// installation_id is the same class as the rest: one series per channel
+	// installation, growing with tenants rather than with the deployment. It
+	// is also the natural thing to reach for in any channel metric — every
+	// adapter call site already carries one — which is what makes leaving it
+	// off this list a matter of time rather than of luck.
+	"installation_id": {},
+	"user_id":         {},
+	"agent_id":        {},
+	"task_id":         {},
+	"issue_id":        {},
+	"runtime_id":      {},
+	"session_id":      {},
+	"ip":              {},
 }
 
 var (
--- a/server/internal/metrics/labels_test.go
+++ b/server/internal/metrics/labels_test.go
@@ -41,3 +41,16 @@ func TestNormalizeLabelsCollapseUnknownValues(t *testing.T) {
 		t.Fatalf("NormalizeTaskSource unknown = %q, want other", got)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


seacen:
@Bohan-J — a nudge on this one. It has been sitting four days with nine checks green and no eyes on it.

It is 26 lines in one file plus its test, and it changes nothing at runtime: no metric uses `installation_id` today. It adds the name to a list that already forbids `workspace_id`, `session_id`, `task_id`, `agent_id`, `runtime_id` and `ip` — the same class of per-tenant identifier, and the one most likely to be added by accident, since every channel adapter has one in hand at every metric call site.

The reason it is worth doing before rather than after: a per-tenant label cannot be taken back out of an operator's retention window once the series exist.

Happy to close it if you would rather not grow that list — but it should not sit unread.

Bohan-J:
Thanks @seacen — merging this.

The reasoning in the PR description is the part that makes this worth taking. `installation_id` really is the same class as `workspace_id` and `session_id`, and your point about it being the one most likely to be added by accident is the convincing bit: every channel adapter call site already has one in hand, so this is the label someone reaches for without thinking. Turning "the author has to remember" into "the guard says no" is the right trade when the mistake is unrecoverable inside the retention window.

Also appreciated that the comment explains *why* the label belongs on the list rather than just adding a string, and that the test guards the list contents rather than duplicating what `TestBusinessMetricLabelsRejectHighCardinalityNames` already covers. Both make the next person's job easier.

Two non-blocking notes, neither of which should hold this up:

- The new test samples `installation_id`, `workspace_id`, `session_id`. The choice of those three is arbitrary — asserting the full expected set (so that adding *or* removing an entry is a deliberate act) would be a touch stronger. Fine as is.
- Worth knowing where the guard's reach ends: `validateBusinessMetricLabels()` only sees labels declared in `businessMetricLabels`. The metrics in `business_sampler.go` and `http.go` are built with hardcoded `[]string{...}` and never go through `metricLabels()`, so they're outside the check entirely. That's pre-existing, not something this PR introduces, but it does mean the guard closes one door rather than all of them. Unifying every metric behind the registry would be a good separate change if you're interested.

PR #6598, opened 2026-08-07
feat(wecom): make a refused handshake distinguishable from an unreachable one (MUL-5895)

seacen:
`subscribe()` returned a plain `fmt.Errorf` for a server refusal, so every caller sees one undifferentiated "connect failed".

The two cases need **opposite** operator responses:

| | behaviour | response |
|---|---|---|
| **refusal** — the bot id / secret pair is not valid | repeats identically on every backoff, forever | somebody has to go fix that installation |
| **everything else** — dial failed, timeout, TLS, a throttle, a platform-side failure | usually recovers on its own | wait |

Watching one number, an operator cannot tell *"our network blipped"* from *"this tenant's bot has been broken since Tuesday"* — and only the second has a user sitting in front of a bot that never answers.

## Where the classification comes from

`credential_probe.go` already answers this question about this ack, at install time: `rejectionErrCodes` is a whitelist of 40001 / 40013 and every other non-zero code fails closed to unverifiable, because WeCom only guarantees that 0 means success and the subscribe path is under frequency and concurrency protection (45009, 45033).

`subscribe()` now goes through `classifySubscribeAck`, which lives next to that whitelist and which the probe calls too. One function, so the connect path cannot answer a code differently from the install path — a 45009 that means "wait" at install time cannot mean "go fix this installation" on reconnect. It returns the exported `ErrCredentialsRejected` / `ErrCredentialsUnverifiable` pair, the same two `internal/handler/wecom_web.go` already branches on for the install path.

## Scope

Nothing branches on the classification yet. This is the piece a connect-vs-auth failure metric split needs, filed on its own because it is small, has no dependencies, and the classification is already useful to whoever reads the log.

One constraint for that follow-up: `channel/engine` is what receives this error out of `Connect()`, but it cannot branch on these sentinels — `wecom` already imports `channel/engine` (`binding.go`, `replier.go`, `ws_frame.go`, `installation.go`, `wecom_resolvers.go`), so the reverse import is a cycle. The metric has to be emitted in the WeCom layer — `classifySubscribeAck` is the one point both paths funnel through — or live above both, or `channel` needs a platform-neutral classification contract first.

## Tests

Every test drives the real `subscribe()` through a fake conn, so a misclassification inside `subscribe()` fails them.

- a table over the probe's own errcodes (40001, 40013, 45009, 45033, -1, 999999), asserting the connect path reaches the same verdict and never both at once
- a transport failure — a conn whose `ReadMessage` fails, and separately one whose `WriteMessage` fails — asserting it is not classified as a rejection
- a zero errcode still succeeds

### DIFF
--- a/server/internal/integrations/wecom/credential_probe.go
+++ b/server/internal/integrations/wecom/credential_probe.go
@@ -79,6 +79,37 @@ var rejectionErrCodes = map[int]struct{}{
 	40013: {}, // 不合法的CorpID — the identity half of the pair is not valid
 }
 
+// classifySubscribeAck turns a non-zero errcode on an aibot_subscribe ack into
+// one of the two sentinels above.
+//
+// Both readers of that ack go through here — this probe at install time, and
+// the channel's own handshake on every reconnect (wecom_channel.go). It is one
+// function rather than two copies of the same map lookup so the two cannot
+// give different answers to the same code: a 45009 throttle that is
+// "unverifiable, wait" at install time must not be "somebody go fix this
+// installation" on the connect path.
+//
+// log receives the codes we do not recognize; nil falls back to the default
+// logger.
+func classifySubscribeAck(log *slog.Logger, errCode int, errMsg string) error {
+	if _, rejected := rejectionErrCodes[errCode]; rejected {
+		// The one answer the admin can act on, and the only branch that
+		// blames their credentials.
+		return fmt.Errorf("%w (errcode %d: %s)", ErrCredentialsRejected, errCode, errMsg)
+	}
+	// Unknown non-zero: throttling, a platform-side failure, or a code WeCom
+	// added since. Fail closed rather than blame the secret. The raw pair is
+	// logged here so an operator can see it even if the caller only surfaces
+	// the sentinel, and carried on the error so the caller's own log line has
+	// it too.
+	if log == nil {
+		log = slog.Default()
+	}
+	log.Warn("wecom: unrecognized subscribe errcode, treating as unverifiable",
+		"errcode", errCode, "errmsg", errMsg)
+	return fmt.Errorf("%w (subscribe returned errcode %d: %s)", ErrCredentialsUnverifiable, errCode, errMsg)
+}
+
 // credentialProbeTimeout bounds the whole probe. It has to cover a dial and
 // one round trip and nothing else, and it is spent with the admin watching a
 // spinner, so it is short. Sized off the connection's own two constants
@@ -162,19 +193,7 @@ func (p *handshakeProbe) Probe(ctx context.Context, botID, secret string) error 
 			continue
 		}
 		if env.ErrCode != 0 {
-			if _, rejected := rejectionErrCodes[env.ErrCode]; rejected {
-				// The one answer the admin can act on, and the only branch
-				// that blames their credentials.
-				return fmt.Errorf("%w (errcode %d: %s)", ErrCredentialsRejected, env.ErrCode, env.ErrMsg)
-			}
-			// Unknown non-zero: throttling, a platform-side failure, or a code
-			// WeCom added since. Fail closed rather than blame the secret. The
-			// raw pair is logged here so an operator can see it even if the
-			// caller only surfaces the sentinel, and carried on the error so
-			// the handler's own log line has it too.
-			slog.Warn("wecom credential probe: unrecognized subscribe errcode, treating as unverifiable",
-				"errcode", env.ErrCode, "errmsg", env.ErrMsg)
-			return fmt.Errorf("%w (subscribe returned errcode %d: %s)", ErrCredentialsUnverifiable, env.ErrCode, env.ErrMsg)
+			return classifySubscribeAck(slog.Default(), env.ErrCode, env.ErrMsg)
 		}
 		return nil
 	}
--- a/server/internal/integrations/wecom/subscribe_error_test.go
+++ b/server/internal/integrations/wecom/subscribe_error_test.go
@@ -1,0 +1,113 @@
+package wecom
+
+// subscribe_error_test.go — a handshake the server refused and a handshake
+// that never reached the server call for opposite responses. One needs a
+// person to go fix the installation; the other usually fixes itself. Callers
+// (the Supervisor's logging today, a metrics split tomorrow) can only tell
+// them apart if the rejection is identifiable.
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+func TestSubscribeRejectionIsIdentifiable(t *testing.T) {
+	// The shape subscribe() produces on a non-zero errcode.
+	err := fmt.Errorf("%w: errcode=%d errmsg=%s", errSubscribeRejected, 40001, "invalid credential")
+	if !errors.Is(err, errSubscribeRejected) {
+		t.Fatal("a refused handshake is not identifiable, so it cannot be told apart from a network failure")
+	}
+	// The diagnosis must survive: the errcode is what names WHICH credential
+	// problem it is.
+	if got := err.Error(); got == "" || !contains(got, "40001") || !contains(got, "invalid credential") {
+		t.Errorf("error text lost the server's diagnosis: %q", got)
+	}
+}
+
+func TestTransportFailuresAreNotRejections(t *testing.T) {
+	for _, err := range []error{
+		&net.OpError{Op: "dial", Err: errors.New("connection refused")},
+		errors.New("wecom: subscribe read: i/o timeout"),
+		fmt.Errorf("wecom: send subscribe: %w", errors.New("broken pipe")),
+	} {
+		if errors.Is(err, errSubscribeRejected) {
+			t.Errorf("%v was classified as a credential rejection — an operator would go chase a tenant whose bot is fine", err)
+		}
+	}
+}
+
+func contains(s, sub string) bool {
+	for i := 0; i+len(sub) <= len(s); i++ {
+		if s[i:i+len(sub)] == sub {
+			return true
+		}
+	}
+	return false
+}
+
+// ackingConn answers the subscribe frame with a chosen errcode, echoing the
+// req_id the way the server does.
+type ackingConn struct {
+	errCode int
+	errMsg  string
+	sent    chan []byte
+	reqID   string
+}
+
+func (c *ackingConn) WriteMessage(_ int, data []byte) error {
+	var env frameEnvelope
+	if err := json.Unmarshal(data, &env); err != nil {
+		return err
+	}
+	c.reqID = env.Headers.ReqID
+	return nil
+}
+
+func (c *ackingConn) ReadMessage() (int, []byte, error) {
+	payload, err := json.Marshal(frameEnvelope{
+		Headers: frameHeaders{ReqID: c.reqID},
+		ErrCode: c.errCode,
+		ErrMsg:  c.errMsg,
+	})
+	if err != nil {
+		return 0, nil, err
+	}
+	return websocket.TextMessage, payload, nil
+}
+
+func (c *ackingConn) SetReadDeadline(time.Time) error  { return nil }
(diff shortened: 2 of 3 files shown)


Bohan-J (changes requested):
Thanks for splitting this out — the motivation is right, the scope note ("nothing branches on it yet") is the honest way to file it, and `TestSubscribeWrapsAServerRefusal` / `TestSubscribeSucceedsOnZeroErrcode` are good tests: driving the real `subscribe()` through a fake conn that echoes `req_id` is exactly the right shape, and your note about the first version passing with the wrapping removed is the right instinct.

Two must-fixes before this can land.

---

### 1. The classification contradicts the one this package already made for the same handshake

`subscribe()` now wraps **every** non-zero errcode as `errSubscribeRejected`, and the sentinel's doc comment commits to a strong claim about what that means — "it repeats identically on every backoff until somebody fixes the installation".

`credential_probe.go` reads the same errcodes off the same `aibot_subscribe` handshake and deliberately does the opposite:

```go
// credential_probe.go
var rejectionErrCodes = map[int]struct{}{
	40001: {}, // 不合法的secret参数 — the secret does not match this bot
	40013: {}, // 不合法的CorpID — the identity half of the pair is not valid
}
```

It is a whitelist, and the comment above it spells out why it must stay one:

> Everything else non-zero is `ErrCredentialsUnverifiable`, deliberately. WeCom only guarantees that 0 means success; the subscribe path is also under frequency and concurrency protection (45009, 45033), and the platform can fail on its own account.

So a 45009 throttle — the textbook self-healing case — is `ErrCredentialsUnverifiable` on the probe path and "a person has to go fix this installation" on the connect path, roughly 200 lines apart in the same package. Since the stated purpose of this sentinel is to feed a connect-vs-auth failure metric, wiring that metric up later would make throttling page someone on a tenant whose bot is fine, which is the exact failure the PR description sets out to prevent.

Whatever the follow-up metric consumes, the classification underneath it has to agree with the probe's.

### 2. `TestTransportFailuresAreNotRejections` doesn't hold the line it's described as holding

The PR description says this test keeps the split from drifting the other way. It can't: it constructs three errors by hand and asserts `errors.Is` is false on them, without ever calling `subscribe()`. It's asserting the semantics of `errors.Is` on values the test itself built.

I confirmed by mutation — misclassifying a transport failure in production code:

```go
// wecom_channel.go, in subscribe()
-return fmt.Errorf("wecom: subscribe read: %w", err)
+return fmt.Errorf("wecom: subscribe read: %w: %w", errSubscribeRejected, err)
```

`go test ./internal/integrations/wecom/ -count=1` still passes, this test included. The negative case needs to run `subscribe()` against a conn whose `ReadMessage` fails, the same way the two positive cases do — a fake conn returning a read error, then asserting `!errors.Is(err, errSubscribeRejected)`.

`TestSubscribeRejectionIsIdentifiable` has the same property (it builds the error with `fmt.Errorf` and then asserts on it), so it's covered by `TestSubscribeWrapsAServerRefusal` and could just go.

---

### Suggestion

The package already has two shapes for this, and both are stronger than a new package-private boolean:

- `wecomAPIError{Cmd, Code, Msg}` in `ws_sender.go`, added in #6590, whose comment makes the case directly: *"Carrying the errcode rather than a string is what lets a caller tell a permanent refusal (bad frame, bot removed from the chat) from a transient one (rate limited) instead of pattern-matching prose."*
- The exported `ErrCredentialsRejected` / `ErrCredentialsUnverifiable` pair, already consumed across packages via `errors.Is` in `internal/handler/wecom_web.go`.

Either one fits here:

- return `&wecomAPIError{Cmd: cmdSubscribe, Code: env.ErrCode, Msg: env.ErrMsg}` — keeps the code intact and leaves the policy to whoever is in a position to have one, which also means the metric PR doesn't inherit a decision made at the wrong layer; or
- reuse the exported pair plus `rejectionErrCodes`, so both handshake paths classify identically.

The second has a bonus: `errSubscribeRejected` is package-private, so `channel/engine` — which is what actually receives the error from `Connect()` — can't reference it today. The exported pair is already reachable.

### Nit

`contains()` in the test file can be `strings.Contains`.

---

Direction and scope are both fine; it's the classification boundary that needs to move. Happy to look again once it does.

seacen:
Both fixed. Taking the classification first, since the test follows from it.

### Which shape

The second one. `subscribe()` now returns the exported `ErrCredentialsRejected` / `ErrCredentialsUnverifiable` pair through a `classifySubscribeAck` that sits next to `rejectionErrCodes` and that the probe calls too. `errSubscribeRejected` is gone.

I weighed `wecomAPIError` seriously — "carry the code, leave the policy to whoever has one" is the better instinct in general, and its own comment makes that case well. Two things decided against it here:

- It is package-private as well. `channel/engine` can neither name the type nor `errors.As` on it, so a metric PR would still have to reach for something else. It moves the problem rather than solving the one you pointed at.
- The policy already exists in this package, twenty lines from the whitelist, and it is not one the connect path is entitled to differ on. Deferring would leave two readers of the same ack with no shared answer, which is the state you filed this against.

What decided it for *one function* rather than two copies of the map lookup: the drift you found came from duplicated judgment, and consulting the same map from two places is one edit away from the same failure. Now there is one place to change and both paths change with it.

I did look for a third shape. The one worth naming is `subscribe()` returning a `wecomAPIError` that also wraps `ErrCredentialsRejected` — code and classification both. I dropped it. The struct is unexported, so the code is only reachable inside the package and the errcode is already in the message; and giving `wecomAPIError` an `Unwrap` would make the send-path errcodes from #6590 claim "credentials rejected", which is a different question with a different whitelist.

One deliberate narrowing: the old doc comment claimed a deleted bot and a switched-off long connection as rejections. Under the whitelist they land in unverifiable, because we do not know which codes WeCom returns for them. That under-counts auth failures rather than over-paging, which is the direction the whitelist's own comment argues for.

### The test

You were right, and your mutation is the proof both ways.

**Before**, on `d819320` as you reviewed it:

```
-			return fmt.Errorf("wecom: subscribe read: %w", err)
+			return fmt.Errorf("wecom: subscribe read: %w: %w", errSubscribeRejected, err)

$ go test ./internal/integrations/wecom/ -count=1
ok  	github.com/multica-ai/multica/server/internal/integrations/wecom	6.627s

$ go test ./internal/integrations/wecom/ -count=1 -run 'TestTransportFailuresAreNotRejections' -v
=== RUN   TestTransportFailuresAreNotRejections
--- PASS: TestTransportFailuresAreNotRejections (0.00s)
PASS
```

**After**, the same mutation (the sentinel is the exported one now):

```
-			return fmt.Errorf("wecom: subscribe read: %w", err)
+			return fmt.Errorf("wecom: subscribe read: %w: %w", ErrCredentialsRejected, err)

$ go test ./internal/integrations/wecom/ -count=1
--- FAIL: TestSubscribeTransportFailuresAreNotRejections (0.00s)
    --- FAIL: TestSubscribeTransportFailuresAreNotRejections/the_ack_never_comes_back (0.00s)
        subscribe_error_test.go:154: subscribe() called a transport failure a credential rejection: wecom: subscribe read: wecom: WeCom rejected this bot id and secret: i/o timeout — an operator would go chase a tenant whose bot is fine
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	6.561s
```

And the other direction — reverting the classification to the plain `fmt.Errorf` it replaced:

```
-			return classifySubscribeAck(log, env.ErrCode, env.ErrMsg)
+			return fmt.Errorf("wecom: subscribe rejected errcode=%d errmsg=%s", env.ErrCode, env.ErrMsg)

$ go test ./internal/integrations/wecom/ -count=1
--- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe (0.00s)
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/invalid_secret_is_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode 40001 as wecom: subscribe rejected errcode=40001 errmsg=invalid secret, want wecom: WeCom rejected this bot id and secret
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/invalid_corpid_is_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode 40013 as wecom: subscribe rejected errcode=40013 errmsg=invalid corpid, want wecom: WeCom rejected this bot id and secret
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/rate_limited_is_not_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode 45009 as wecom: subscribe rejected errcode=45009 errmsg=api freq out of limit, want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/concurrency_limited_is_not_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode 45033 as wecom: subscribe rejected errcode=45033 errmsg=api concurrency out of limit, want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/system_error_is_not_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode -1 as wecom: subscribe rejected errcode=-1 errmsg=system busy, want wecom: could not reach WeCom to verify this bot
    --- FAIL: TestSubscribeClassifiesAnAckLikeTheCredentialProbe/unknown_code_is_not_a_rejection (0.00s)
        subscribe_error_test.go:110: subscribe() classified errcode 999999 as wecom: subscribe rejected errcode=999999 errmsg=something new, want wecom: could not reach WeCom to verify this bot
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	6.352s
```

The negative case runs `subscribe()` against a conn whose `ReadMessage` fails, and a second one whose `WriteMessage` fails so the "send subscribe" path is covered too. The positive case became a table over the probe's own errcodes, so the two paths are pinned to the same verdicts rather than to "some sentinel". `TestSubscribeRejectionIsIdentifiable` is gone. `contains()` is `strings.Contains`.

Rebased onto main. `go build ./...`, `go vet ./...` and `gofmt -l` clean, `go test -race ./internal/integrations/wecom/ -count=1` green.

Bohan-J (approved):
Both must-fixes are resolved, and the fix is better than either shape I suggested — extracting `classifySubscribeAck` so both readers call one function removes the duplicated judgment that caused the drift, rather than having two places consult the same map. Your reasons for passing on `wecomAPIError` hold up, and the narrowing you called out (a deleted bot / switched-off long connection now land in unverifiable) is the right direction for a whitelist that exists to avoid over-blaming credentials.

I re-ran your two mutations and both now fail as you showed. I also ran a third you didn't: adding `45009` to `rejectionErrCodes` fails `TestSubscribeClassifiesAnAckLikeTheCredentialProbe/rate_limited_is_not_a_rejection` **and** `TestProbeClassifiesSubscribeErrCodes/rate_limited_is_not_a_rejection` — both sides at once. That's the part that matters: the shared policy is pinned by two independent tables, so drift in either direction trips a test rather than a comment.

`go build ./...`, `go vet`, `gofmt -l` and `go test -race` clean here too, and CI is 9/9.

One correction, not worth another round trip. The new comment on `subscribe()` ends with:

> Both sentinels are exported, so channel/engine — which is what receives this error out of Connect() — can branch on them.

`channel/engine` can't: `wecom` already imports it (`binding.go`, `replier.go`, `ws_frame.go`, `installation.go`, `wecom_resolvers.go`), so the reverse import is a cycle — I confirmed with a throwaway file in `engine`:

```
imports .../integrations/wecom from zz_cycle_probe.go
imports .../integrations/channel/engine from binding.go: import cycle not allowed
```

Exporting is still right — `internal/handler/wecom_web.go` already branches on both — but the metric this is groundwork for has to be emitted in the WeCom layer (`classifySubscribeAck` is the one point both paths funnel through) or above both packages, not in `engine`. I've corrected the PR description to say that so the follow-up isn't designed against a premise that doesn't compile; the code comment still has the old sentence and is worth a one-liner whenever you next touch this file.

Nit for the same follow-up: `classifySubscribeAck` lives in `credential_probe.go`, whose header is entirely about proving bot ownership at install time. It's now the connect path's policy too, and nobody debugging a reconnect will look there.

Approving. Thanks for the thorough turnaround, and for including the mutation output — it made this round quick.

PR #6599, opened 2026-08-07
MUL-5901: feat(wecom): read voice notes, which WeCom already transcribes for us

seacen:
A voice message currently gets:

> 抱歉，我目前只能处理文字消息。

But **WeCom runs the speech recognition on its side** and hands the result over in `voice.content`. No audio to download, no media key to handle, no storage to write — the adapter was refusing a sentence it was already holding.

Voice notes are how people talk to a bot on a phone: walking, in a car, hands full. On a mobile-first platform that is not an edge case, and the failure mode is the bot looking like it does not work.

## What it does

`bodyText()` answers "the message's text, typed or spoken", and the callback path reads it instead of `Text.Content` directly.

- an **empty transcript** (background noise, a half-second press) still takes the receipt path — an empty body would otherwise reach the agent as a turn with nothing in it
- a spoken **`/issue …`** is a command, because the parser reads the same body. That falls out of routing it as text rather than being special-cased
- **image / file / video / mixed are unchanged.** This adds no media ingestion — those genuinely need a download, a key and storage

Single chats only; WeCom does not deliver voice to a bot in a group (errcode 100719), which is noted on the field.

## Tests

Five. Reverting the routing fails two:

```
a voice note with a transcript was refused — the sentence was already in hand
a spoken /issue did not register as a command
```

The other three pin that empty transcripts are not ingested, that typed text is unaffected, and that the downloadable kinds still take the receipt path — so this cannot quietly grow into media handling later.

### DIFF
--- a/server/internal/integrations/wecom/types.go
+++ b/server/internal/integrations/wecom/types.go
@@ -28,8 +28,10 @@
 // shared channel engine? Keep this adapter building — and loop in the code
 // owners for anything that changes WeCom-visible behavior.
 //
-// Known limits of the first version, both deliberate: inbound handling is
-// text-only (other message types get a short "text only" receipt), and outbound
+// Known limits of the first version, both deliberate: inbound handling takes
+// only what arrives as words — typed text, and the transcript WeCom returns
+// for a voice note — while the kinds carrying a downloadable payload get a
+// short "text only" receipt; and outbound
 // delivery requires a SINGLE backend replica, because the only send path is the
 // in-process WebSocket in sendersRegistry while EventChatDone dispatches on the
 // in-process events.Bus. See SELF_HOSTING.md.
--- a/server/internal/integrations/wecom/voice_test.go
+++ b/server/internal/integrations/wecom/voice_test.go
@@ -1,0 +1,67 @@
+package wecom
+
+// voice_test.go — WeCom runs speech recognition on its side and delivers the
+// transcript in voice.content. Answering a voice note with "I only handle
+// text" refuses a sentence we were already handed.
+
+import "testing"
+
+func TestVoiceTranscriptIsTreatedAsText(t *testing.T) {
+	mc := aibotMsgCallback{MsgType: "voice"}
+	mc.Voice.Content = "把登录跳转的 bug 记一下"
+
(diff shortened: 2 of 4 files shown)


seacen:
Rebased onto main. Three conflicts, plus one thing the auto-merge got wrong that no
conflict marker pointed at.

**The conflicts.** All three were main's `channelMessageFromCallback` work meeting this
branch's voice handling:

- The signature. Main now threads `botDisplayName` for group mention-stripping; this
  branch still had the four-argument form. Took main's signature and kept the
  `bodyText()` call inside it.
- `SkipAgentRun`. Main reads it off `command` (the mention-stripped text) rather than the
  raw body — the #6593 change. Took main's.
- `dispatchFrame`. Main added `traceIn`/`traceInbound`; this branch replaced the
  `MsgType != "text"` gate. Kept both.

**What the clean merge got wrong.** Main added `command := mc.Text.Content` after this
branch was cut, and git applied both sides without complaint: `Text` and `Content` moved to
the transcript while `command` stayed on the typed field. A spoken `/issue …` would then
have arrived with `CommandText` empty and `SkipAgentRun` false, so the engine would file the
issue from the `Text` fallback *and* run the agent over it — the exact double-handling
#6593 and #6603 removed, reintroduced through the voice path only. `command` now reads off
`bodyText()` as well. The existing "a spoken /issue is a command" test covers it.

**Also changed while here:**

- `traceInbound` now receives the routed body rather than `mc.Text.Content`. Turning tracing
  on to follow a voice note showed `len=0` next to a bot that had just answered it.
- Four comments asserted text-only routing that is no longer true: the package doc in
  `types.go`, the `Content` field doc, the `channelMsgType` default branch, and the
  "Iteration 1 routes only text" note in `dispatchFrame`. Rewritten to describe what the
  code does.
- Dropped "Single chats only (WeCom errcode 100719 covers the group case)". 100719 is a
  documentation path id, not an errcode, and `bodyText` does not gate on chat type — so the
  sentence described neither the source nor the code.

`go build`, `go vet`, `gofmt` clean; `go test -race ./internal/integrations/wecom/...`
green against a migrated Postgres.

PR #6600, opened 2026-08-07
fix(wecom): a re-install must not silently drop config the dialog never asked for

seacen:
## The defect

`InstallationService.Upsert` replaces `channel_installation.config` wholesale and rebuilt it from the install request alone:

```go
cfg, err := encodeInstallConfig(Installation{
    BotID:           p.BotID,
    SecretEncrypted: sealed,
})
```

Anything in that JSONB the install dialog does not collect is therefore erased by the next write to the row. Re-installing is not a rare event — the dialog is also how an admin rotates a secret, and that call carries the same bot id and the same agent id, so it lands on the existing row via `ON CONFLICT (workspace_id, agent_id, channel_type)`.

This PR adds the first config key that is not collected by the dialog, `config->>'principal_user_id'` — the Multica user the bot is operated on behalf of, for when an admin installs it for a colleague — and the carry-forward that keeps it.

## User-visible consequence

Without the carry-forward: an admin pastes a fresh secret into Settings after a leak and saves. The bot is now assigned to whoever pressed save. Nothing on screen says the setting changed and nothing in the log does either, and there is no dialog field for the principal, so it cannot be put back from the UI — the row has to be edited by hand.

That is why the field and the carry-forward are one PR rather than two. A setting that a routine secret rotation silently drops is worse than not having the setting.

## What is inert, and why it is here anyway

**Nothing reads `PrincipalUserID` yet.** The run-progress code that consumes it (deciding how much of a run the principal sees) is not upstream; it is part of a later chain. Until that lands this field is written, carried and read back, and no behavior depends on its value. Flagging that up front rather than leaving a reviewer to grep for the reader and find none.

It is not exposed on the wire: `WecomInstallationResponse` (`handler/wecom_web.go:27`) is an explicit DTO and was not touched, so the field does not appear in any API response.

## Compatibility

The persisted JSONB key is the compatibility surface, so it is `omitempty` and absent means the installer:

- an installation that never sets a principal encodes byte-for-byte the config it encoded before this change — no migration, no rewrite of rows already in the table;
- a row written before the field existed decodes with `PrincipalUserID == ""` and its other fields intact.

Both are asserted in `TestInstallConfig_PrincipalKeyIsAbsentUntilSet`, the second against a literal pre-change config blob rather than one this build wrote.

## Sibling precedent

DingTalk already reads the row it is about to replace from inside the install transaction, for the same serialization reason: `internal/integrations/dingtalk/install.go:188` (`GetDingTalkInstallationOwnerForUpdate`, on `qtx`, between the reclaim and the upsert). This PR does the same thing in the same place — read on the tx handle, after `ReclaimDeadChannelInstallationByAppID`, before `UpsertChannelInstallation`.

There is no query keyed on `(workspace_id, agent_id, channel_type)` — the upsert's conflict key was never something a read path needed — so `currentInstallation` filters the workspace's wecom rows, of which a workspace has a handful. The reclaim never removes this row (it spares the caller's own `(workspace, agent)` pair, per `ReclaimDeadChannelInstallationByAppID`'s contract), so reading after it is safe.

## Tests

Three new tests in `server/internal/integrations/wecom/installation_principal_test.go`. The two DB-backed ones extend the existing `installation_reclaim_test.go` rig — same workspace/agent fixtures, real migrated Postgres, skip when none is reachable.

Every one was reverse-verified: the fix was reverted, the test was confirmed to fail, and the fix restored. Actual output below.

**1. `TestUpsert_RotatingTheSecretKeepsThePrincipal`** — a row in the pre-change config shape, given a principal by hand, then a secret rotation through `Upsert`. Asserts the principal survives in both the returned value and the stored config, and that the rotation itself still landed (the stored blob unseals to the *new* secret, so carrying the principal forward does not carry the old secret with it).

Reverted the carry-forward (`PrincipalUserID: carried.PrincipalUserID` dropped from the encode):

```
--- FAIL: TestUpsert_RotatingTheSecretKeepsThePrincipal (0.07s)
    installation_principal_test.go:123: principal = "" after a secret rotation, want it carried forward as "5c09e200-0000-4000-8000-0000000000c1" — the bot silently changed hands to whoever pressed save
FAIL
```

**2. `TestUpsert_FirstInstallDoesNotInheritAnotherAgentsPrincipal`** — agent A has a bot with a principal; agent B gets its first bot. Asserts B's install carries no principal and A's row is untouched. This drives the `agent_id` filter in `currentInstallation`, which is the one place the workspace-scoped read could go wrong.

Reverted that filter (`if row.AgentID != agentID { continue }` removed):

```
--- FAIL: TestUpsert_FirstInstallDoesNotInheritAnotherAgentsPrincipal (0.05s)
    installation_principal_test.go:162: principal = "5c09e200-0000-4000-8000-0000000000c1" on a first install, want empty (the installer) — agent B's bot was handed to agent A's principal
FAIL
```

**3. `TestInstallConfig_PrincipalKeyIsAbsentUntilSet`** — no DB. The encode/decode pair: the key is absent until set, a pre-change row decodes, a set principal round-trips.

Reverted `omitempty` (`json:"principal_user_id"`):

```
--- FAIL: TestInstallConfig_PrincipalKeyIsAbsentUntilSet (0.00s)
    installation_principal_test.go:193: an install that sets no principal wrote principal_user_id anyway: {"app_id":"BOT-1","bot_id":"BOT-1","secret_encrypted":"c2VhbGVk","principal_user_id":""}
FAIL
```

## Verification

`go build ./...` and `go vet ./...` clean; `gofmt` clean.

Full `go test ./internal/...` against a migrated Postgres: green except `TestProbeAgentCLIs_QoderResolvesViaLoginShell` in `internal/daemon`, which fails identically on unmodified `main` on this machine (a real `qoderclicn` on `PATH` beats the test's fake) and is unrelated to this change.

```
ok  github.com/multica-ai/multica/server/internal/integrations/wecom
```

### DIFF
--- a/server/internal/integrations/wecom/installation.go
+++ b/server/internal/integrations/wecom/installation.go
@@ -217,6 +217,21 @@ func (s *InstallationService) Upsert(ctx context.Context, p InstallationParams) 
 
 	// The row this upsert is about to overwrite, read on the tx handle so it is
 	// serialized with the write. Zero when this is a first install.
+	//
+	// It is read because UpsertChannelInstallation replaces `config` wholesale
+	// while the config holds more than the install dialog collects — the chat
+	// display name, and principal_user_id saying who the bot is operated on
+	// behalf of when an admin set it up for a colleague. Rebuilding the config
+	// from the request alone drops those, so a secret rotation through the
+	// Settings dialog would silently reassign the bot to whoever clicked the
+	// button, with nothing on screen and nothing in the log to say the setting
+	// had gone.
+	//
+	// Reading after the reclaim above is safe because the reclaim cannot take
+	// this row: its revoked branch explicitly excludes the caller's own
+	// (workspace, agent) pair, and its two orphan branches only fire for a
+	// missing workspace or agent, neither of which this install could have
+	// resolved to get here.
 	carried, err := currentInstallation(ctx, qtx.Queries, p.WorkspaceID, p.AgentID)
 	if err != nil {
 		return Installation{}, err
@@ -231,10 +246,15 @@ func (s *InstallationService) Upsert(ctx context.Context, p InstallationParams) 
 	if displayName == "" && carried.BotID == p.BotID {
 		displayName = carried.BotDisplayName
 	}
+	// The principal carries differently, and deliberately so: it is a property
+	// of the AGENT — who this agent's bot answers for — not of the credentials,
+	// so unlike the display name above it survives a re-install that swaps in a
+	// different bot as well as one that only rotates the secret.
 	cfg, err := encodeInstallConfig(Installation{
 		BotID:           p.BotID,
 		SecretEncrypted: sealed,
 		BotDisplayName:  displayName,
+		PrincipalUserID: carried.PrincipalUserID,
 	})
 	if err != nil {
 		return Installation{}, err
@@ -268,7 +288,9 @@ func (s *InstallationService) Upsert(ctx context.Context, p InstallationParams) 
 // There is no query keyed on that triple — the conflict key of the upsert is
 // not something any read path needed until now — so this filters the
 // workspace's wecom rows, of which there are a handful. Running it on the
-// caller's queries handle is what makes it serializable with the write.
+// caller's queries handle is what makes it serializable with the write;
+// dingtalk reads its current row inside the same transaction for the same
+// reason (dingtalk/install.go:188, GetDingTalkInstallationOwnerForUpdate).
 func currentInstallation(ctx context.Context, q *db.Queries, workspaceID, agentID pgtype.UUID) (Installation, error) {
 	rows, err := q.ListChannelInstallationsByWorkspace(ctx, db.ListChannelInstallationsByWorkspaceParams{
 		WorkspaceID: workspaceID,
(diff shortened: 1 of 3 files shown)


seacen:
Closing this rather than carrying it further, after rebasing onto current `main`.

#6603 landed `currentInstallation` and the read-before-upsert pattern this PR was opened to introduce. What remains on top of main is a `PrincipalUserID` field that **no code path writes and none reads**, the logic to carry it across a re-install, an error wrap on a decode failure, and three tests.

Carrying an unread field across a rotation is a reasonable thing to want — a value silently dropped on re-install is harder to notice later than one that was never stored. But there is nothing here a reviewer can check against behaviour, and the rebase to keep it current is not free. Better filed alongside the code that first reads it, where the carry-forward has an observable consequence and the test can assert one.

The rebase work is not lost — the branch stays, and the two corrections it turned up are worth repeating wherever this lands: the reclaim spares this row because its orphan branches require a missing workspace or agent, not because it spares the caller's own pair; and the comment about which upstream PR would consume the field has been removed rather than updated, since sentences about another PR's state have gone stale in both directions on this effort.

PR #6602, opened 2026-08-07
feat(wecom): add the MULTICA_WECOM_TRACE operator switch (and stop it printing binding tokens)

seacen:
## The defect

The wecom adapter logs failures and nothing else.

| Event | What lands in the log today |
|---|---|
| Undecodable frame envelope | `Warn` (`wecom_channel.go:215`) |
| Server ack with non-zero errcode | `Warn` (`wecom_channel.go:340`) |
| Non-`disconnected_event` event | `Debug` (`wecom_channel.go:317`) |
| Non-text message receipt | `Debug` (`wecom_channel.go:294`) |
| **An ordinary inbound text message** | **nothing** |
| **Every outbound frame** — subscribe, ping, send_msg | **nothing** |

So a run that fails *loudly* is visible, and a run that goes wrong **quietly** leaves no trace on the server at all: the reply addressed to the room instead of the person, the command that was silently dropped, the receipt that went out twice, the frame we believed we sent that was never written.

## The user-visible consequence

Verifying any of those needs a person to describe what they saw on their phone. That is slow and lossy, and for anything about ordering or timing it does not work at all — a user cannot tell you which of two frames the server accepted first, or distinguish "we never sent it" from "the platform rejected it". Today the only rejection that surfaces is the anonymous-ack `Warn`; a frame that was never written looks identical to one that was delivered.

`MULTICA_WECOM_TRACE=1` records every frame the adapter reads and writes — direction, `cmd`, `req_id`, the chat it was addressed to, whether that chat is a room or a person, the sender, and the server's `errcode`. Off by default.

## Sibling precedent

Slack and Lark both keep a per-frame record; wecom is the only WS adapter with none.

- `server/internal/integrations/slack/slack_channel.go:162` — `c.logger.DebugContext(ctx, "slack: socket mode", "event", evt.Type, "app_id", c.appID)`, one line per socket-mode event.
- `server/internal/integrations/lark/ws_connector.go:339` and `:348` — `log.Debug("lark ws connector: partial chunk buffered" / "chunk reassembly complete", …)`, per frame.

**Why this one is a switch rather than a `Debug` call.** `server/internal/logger/logger.go:106` — `parseLevel` falls through to `slog.LevelDebug`, so **LOG_LEVEL defaults to debug** and a `Debug` line is on in every deployment that has not set it. The siblings' lines carry an event type and an app id; this one carries a prefix of the message body, which must not be on by default. Hence a dedicated env switch that defaults to off, and a boot-time `Warn` while it is on so a session left switched on is visible in the log it is writing into.

## The credential check you asked for

**A binding token is a bearer credential, and a binding URL fits inside the trace preview. I checked, and it does — this PR closes it rather than shipping it.**

`replier.go:148-149` builds:

```go
bindURL := r.appURL + r.bindingPath + "?token=" + url.QueryEscape(token.Raw)
text := "👋 请先绑定你的 Multica 账号，才能与我对话：\n" + bindURL + "\n（链接 15 分钟内有效）"
```

`Mint` → `randomBindingToken(32)` → `base64.RawURLEncoding` = **43 characters**. Measured against the 120-rune preview cap:

| `MULTICA_APP_URL` | token's last char at rune | inside the cap? |
|---|---|---|
| `https://multica.example` | 112 | **yes** |
| `https://multica.ai` | 107 | **yes** |
| `http://localhost:3000` (compose default is `http://localhost:3000`) | 110 | **yes** |

So an unredacted preview printed the **entire live token**, not a fragment. That it is reachable is pinned by `TestBindingPromptFitsInsideThePreviewCap`, which fails loudly if a future copy change ever pushes the token past the cap and makes this rationale stale.

Why that is a leak and not just noise: `RedeemAndBind` (`binding.go:107`) checks only that the redeemer belongs to the token's workspace, and the bind page redeems on load as whoever is signed in. Anyone who could read the log could bind that sender's WeCom identity to their own Multica account before the user clicked their own link — after which the sender's messages, `/issue` included, resolve to the hijacker. This is the same hijack `replier.go:150-159` already refuses to post the link into a room to prevent; a log is the same hijack by another route.

`tracePreview` now runs `redactBearerTokens` before the cut, replacing the value of `token=` / `binding_token=` / `access_token=` / `code=` query parameters with `[redacted]`. It matches on the query parameter rather than the bind path because `BindingPath` is configurable and this package-level function cannot see it.

**The other credential:** the `aibot_subscribe` body carries the installation's decrypted smart-bot secret (`subscribeBody(c.botID, c.secret)`). `traceOut` walks named fields rather than dumping the frame, so the secret is never read — pinned by a test that fails if anyone replaces the walk with a body dump.

## Tests — every one reverse-verified

Eight guards in `server/internal/integrations/wecom/trace_test.go`. Each was confirmed to **fail** with its fix reverted, then restored. Actual output:

**1. Remove the `if !tracingOn()` guard** → `TestTraceOffRecordsNothing`:
```
--- FAIL: TestTraceOffRecordsNothing (0.00s)
    trace_test.go:179: tracing is off but a trace line was written:
        level=INFO msg="wecom trace" dir=out cmd=aibot_subscribe req_id=d73a6a67e45aeb81
        level=INFO msg="wecom trace" dir=in cmd=aibot_msg_callback req_id=REQ_1 errcode=0
        level=INFO msg="wecom trace" dir=in.msg msg_id=MSG_1 chatid=GROUP_CHAT_ID chat_type=group sender=SENDER_USERID msgtype=text len=30 text=一条不该被记录的消息
    trace_test.go:182: tracing is off but the message body reached the log:
```
The capturing logger is at `LevelDebug` on purpose — the package default — so a trace line that were "merely Debug level" would still be written here.

**2. Remove `redactBearerTokens` from `tracePreview`** → `TestTraceNeverLogsABindingToken`. This drives the real replier path (mint → URL → `sendText` → `write` → `traceOut`) with a production-shaped 43-char token:
```
--- FAIL: TestTraceNeverLogsABindingToken (0.00s)
    trace_test.go:269: the raw binding token reached the log:
        level=INFO msg="wecom trace" dir=out cmd=aibot_send_msg chatid=SENDER_USERID chat_type=1
          msgtype=markdown len=179
          text="👋 请先绑定你的 Multica 账号，才能与我对话： https://multica.example/wecom/bind?token=UdiY905i13TLlK-fFGcv8I0uMiWHFFm1xISEE6wYYvg （链接 15 …"
    trace_test.go:272: expected the token parameter to be redacted in place
```
All 43 characters, printed before the `…` cut. That is the leak.

**3. Rune cut → byte cut** → `TestTracePreviewCutsOnARuneBoundary`:
```
--- FAIL: TestTracePreviewCutsOnARuneBoundary (0.00s)
    trace_test.go:320: preview kept 40 runes, want exactly 120
    trace_test.go:313: preview of 301-rune body is not valid UTF-8: "x测测测…测\xe6\xb5…"
    trace_test.go:320: preview kept 42 runes, want exactly 120
    trace_test.go:323: preview contains a replacement character — a multi-byte rune was split
```
Two cases on purpose: a body that is a whole number of 3-byte characters only truncates early (to a third of the intended length), while one ASCII character in front makes the same cut land mid-character and emit invalid UTF-8.

**4. Make `traceOut` append the body map** → `TestTraceNeverLogsTheSmartBotSecret`:
```
--- FAIL: TestTraceNeverLogsTheSmartBotSecret (0.00s)
    trace_test.go:224: the smart-bot secret reached the log:
        level=INFO msg="wecom trace" dir=out cmd=aibot_subscribe req_id=9bbc27345628dff6
          body="map[bot_id:bot-1 secret:THE_SMART_BOT_SECRET]"
```

**5. Remove the `traceIn` / `traceInbound` call sites** → `TestTraceOnRecordsBothDirections`:
```
--- FAIL: TestTraceOnRecordsBothDirections (0.00s)
    trace_test.go:206: trace output is missing "dir=in"
    trace_test.go:206: trace output is missing "dir=in.msg"
    trace_test.go:206: trace output is missing "chatid=GROUP_CHAT_ID"
    trace_test.go:206: trace output is missing "sender=SENDER_USERID"
```

**6. Remove the `traceOut` call site** → the same test, plus both credential tests correctly refused to pass vacuously:
```
--- FAIL: TestTraceOnRecordsBothDirections: trace output is missing "dir=out"
--- FAIL: TestTraceNeverLogsTheSmartBotSecret:
    trace_test.go:221: the subscribe frame was not traced, so this test proves nothing
```
Both "never logs X" tests assert first that the frame **was** traced, so they cannot pass by the code under test not running.

**7. Remove the newline flattening** → `TestTracePreviewFlattensNewlines`:
```
    trace_test.go:334: preview still contains a line break: "first\nsecond\r\nthird"
```

**8. Narrow the redaction regex to `binding_token` only** → `TestRedactBearerTokensCoversTheTokenShapes`: 5 of 7 subtests fail (`binding_url`, `custom_path`, `access_token_param`, `oauth_code`, `stops_at_ampersand`).

`TestTraceOffRecordsNothing`, `TestTraceOnRecordsBothDirections` and `TestTraceNeverLogsTheSmartBotSecret` drive a full `Connect` against a scripted socket — subscribe write, subscribe ack, callback frame, decoded message — so every trace point is exercised through the real code path, not by calling the trace functions directly.

## Test run

```
go build ./...                                     clean
go vet ./...                                       clean
gofmt -l server/internal/integrations/wecom/ …     clean
go test ./internal/integrations/wecom/ -race       ok (all pass)
go test ./cmd/... -p 1                             ok (4/4 packages)
go test ./internal/... -p 1                        2 pre-existing failures, see below
```

The two `./internal/...` failures are unrelated to this change and reproduce identically on unmodified `main` at `47f6e97`:

- `internal/daemon` — `TestProbeAgentCLIs_QoderResolvesViaLoginShell`: `qoder path = "/Users/…/.local/bin/qoderclicn", want /fake/npm-global/bin/qodercli`. The test picks up a real binary installed on the host.
- `internal/migrations` — `TestAttributionStrictConstraintMigrations`.

(These two packages also flake on package-parallel runs: on unmodified `main` the failing set moved between `handler`/`ghsnapshot` and `migrations` across runs on identical code, which is why the numbers above are from `-p 1`.)

## Also in this PR

The env passthrough, with the preview cap stated in both files so an operator does not have to read Go to find out what "a bounded prefix" means:

- `.env.example` — described in the WeCom block's `#   NAME` header alongside `MULTICA_WECOM_SECRET_KEY`, matching this file's convention (description block first, assignments at the end), with the assignment after `MULTICA_WECOM_SECRET_KEY=`.
- `docker-compose.selfhost.yml` — `MULTICA_WECOM_TRACE: ${MULTICA_WECOM_TRACE:-}`.

## Not in this PR

The fault-injection hooks that sit next to this on our branch. They are `if` statements on the hot path of production code and need a build-tag-versus-runtime decision first; this is trace only. Nothing here touches the durable outbound queue, the rate gate, or the single-replica documentation.

### DIFF
--- a/server/internal/integrations/wecom/ws_sender.go
+++ b/server/internal/integrations/wecom/ws_sender.go
@@ -222,6 +222,10 @@ func (s *wsSender) write(frame map[string]any) error {
 	if err != nil {
 		return fmt.Errorf("wecom: marshal frame: %w", err)
 	}
+	// Before the writer mutex: the trace is of what we decided to send, and
+	// holding the socket lock while formatting a log line would let tracing
+	// change the serialization order it is meant to observe.
+	traceOut(s.log, frame)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if err := s.conn.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
--- a/server/internal/integrations/wecom/ws_sender.go
+++ b/server/internal/integrations/wecom/ws_sender.go
@@ -76,6 +76,13 @@ type wsSender struct {
 	// inbound callbacks must not run on it — see the note on sendTextCtx.
 	ackMu   sync.Mutex
 	replies map[string]*replyWaiter
+
+	// seq numbers outbound frames in the order they reach the socket.
+	// Guarded by mu, so it is the wire order by construction, and it is what
+	// pairs a traced send attempt with its outcome — req_id cannot do that
+	// job, because a pong echoes the server's req_id and that may be empty
+	// or repeated. It never goes on the wire.
+	seq uint64
 }
 
 func newWSSender(conn wsConn, log *slog.Logger) *wsSender {
@@ -222,16 +229,31 @@ func (s *wsSender) write(frame map[string]any) error {
 	if err != nil {
 		return fmt.Errorf("wecom: marshal frame: %w", err)
 	}
-	// Before the writer mutex: the trace is of what we decided to send, and
-	// holding the socket lock while formatting a log line would let tracing
-	// change the serialization order it is meant to observe.
-	traceOut(s.log, frame)
+	// Extract the trace fields out here, emit them in there. This mutex is
+	// the point at which the ping loop, agent replies and inbox pushes become
+	// ordered, so a record taken inside it matches the wire by construction,
+	// while one taken outside is only correlated with it — a goroutine can
+	// emit its line and be descheduled before it takes the mutex, and the log
+	// then names the wrong frame as first. Extraction is the expensive half
+	// (a regexp redaction pass and a rune-wise cut over the message body) and
+	// needs no such guarantee, so it stays out here; what runs under the
+	// mutex is a nil check when tracing is off, and two log lines when it is
+	// on, against a socket write that is already in the same section.
+	t := traceOutFields(s.log, frame)
+
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if err := s.conn.SetWriteDeadline(time.Now().Add(writeDeadline)); err != nil {
-		return err
+	s.seq++
+	traceOutAttempt(s.log, s.seq, t)
+
+	stage := traceStageDeadline
+	err = s.conn.SetWriteDeadline(time.Now().Add(writeDeadline))
+	if err == nil {
+		stage = traceStageWrite
+		err = s.conn.WriteMessage(websocket.TextMessage, payload)
 	}
-	return s.conn.WriteMessage(websocket.TextMessage, payload)
+	traceOutResult(s.log, s.seq, t, stage, err)
+	return err
 }
 
 // sendText pushes an aibot_send_msg (proactive push) with plain text to a
--- a/.env.example
+++ b/.env.example
@@ -195,7 +195,11 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 #                              smart-bot secret is never read, but user text
 #                              is not — turn it on deliberately and unset it
 #                              when the session ends. The backend logs a
-#                              warning on every boot while it is on.
+#                              warning on every boot while it is on. Read at
+#                              boot, so changing it needs a backend restart;
+#                              who can read the resulting logs and how long
+#                              they are kept is your log stack's business, see
+#                              SELF_HOSTING_ADVANCED.md → WeCom frame tracing.
 #
 # Per-installation credentials (bot_id, secret) are NOT env vars — they
 # live in the channel_installation table. Create an installation from the
--- a/.env.example
+++ b/.env.example
@@ -182,6 +182,21 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 #                              and the wecom Web-UI endpoints return 503.
 #                              Generate with: openssl rand -base64 32
 #
+#   MULTICA_WECOM_TRACE        1 = log every frame this adapter reads and
+#                              writes, so a real-device session can be checked
+#                              against the server afterwards. Empty / anything
+#                              else = off, which is the right state outside a
+#                              debugging session.
+#                              This logs MESSAGE CONTENT: each traced frame
+#                              carries the first 120 runes of the message body
+#                              (runes, not bytes, so Chinese text is not cut
+#                              mid-character). Binding tokens and other
+#                              token=… query parameters are redacted, and the
+#                              smart-bot secret is never read, but user text
+#                              is not — turn it on deliberately and unset it
+#                              when the session ends. The backend logs a
+#                              warning on every boot while it is on.
+#
 # Per-installation credentials (bot_id, secret) are NOT env vars — they
 # live in the channel_installation table. Create an installation from the
 # WeCom settings tab in the app (Settings → Integrations → WeCom), which
@@ -195,6 +210,7 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 # the lease are dropped and the user sees nothing. Until cross-replica outbound
 # routing lands, run the WeCom-enabled backend as a SINGLE replica.
 MULTICA_WECOM_SECRET_KEY=
+MULTICA_WECOM_TRACE=
 
 # S3 / CloudFront
 # S3_BUCKET — bucket NAME only (e.g. "my-bucket"). Do NOT include the
--- a/SELF_HOSTING_ADVANCED.md
+++ b/SELF_HOSTING_ADVANCED.md
@@ -162,6 +162,25 @@ rewrite configuration. Its backend fallback therefore accepts
 `BACKEND_PORT` → `API_PORT` → `SERVER_PORT` → `8080`, while an explicit
 `REMOTE_API_URL` or `NEXT_PUBLIC_API_URL` still takes priority.
 
+### WeCom frame tracing
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `MULTICA_WECOM_TRACE` | empty (off) | `1` records every WeCom frame the backend reads and writes, including the first 120 runes of each message body. Anything else is off. |
+
+Turn it on for a debugging session and unset it when the session ends. Before
+you do:
+
+- **It is read at boot, so changing it needs a backend restart** — `docker compose -f docker-compose.selfhost.yml up -d backend`. There is no runtime toggle. While it is on, the backend logs a warning on every startup saying so.
+- **Anyone who can read the backend's logs can read the traced message text.** That is a wider audience than the WeCom chat it came from: your `docker logs` / journald / log shipper, and whoever administers them. Binding tokens and other `token=…` parameters are redacted and the smart-bot secret is never read, but user message content is not.
+- **Retention is your log stack's, not the application's.** The backend writes to stderr and keeps nothing itself, so how long the traced text survives is whatever your Docker logging driver or shipper is set to. If that is "forever", decide about it before enabling rather than after.
+
+Each outbound frame produces two lines that share a `seq`: `dir=out` when it is
+about to be written, and `dir=out.done` with `ok=true` / `ok=false` once the
+socket has answered. `seq` is the frame's position in the write order, so the
+pair tells you what the server saw and in which order; an attempt with no
+matching outcome is a write that never returned.
+
 ### CLI / Daemon
 
 These are configured on each user's machine, not on the server:
--- a/SELF_HOSTING_ADVANCED.md
+++ b/SELF_HOSTING_ADVANCED.md
@@ -178,8 +178,13 @@ you do:
 Each outbound frame produces two lines that share a `seq`: `dir=out` when it is
 about to be written, and `dir=out.done` with `ok=true` / `ok=false` once the
 socket has answered. `seq` is the frame's position in the write order, so the
-pair tells you what the server saw and in which order; an attempt with no
+pair tells you what this backend sent and in which order; an attempt with no
 matching outcome is a write that never returned.
+
+`ok=true` means the frame reached the socket, not that WeCom accepted it. For
+the platform's verdict, match the frame's `req_id` against the `dir=in` line
+answering it and read that line's `errcode` — a frame can be written
+successfully and still be rejected there.
 
 ### CLI / Daemon
 
--- a/docker-compose.selfhost.yml
+++ b/docker-compose.selfhost.yml
@@ -146,6 +146,12 @@ services:
       # subscribe credentials at connection time, and to seal the secret when
       # an installation is created from the WeCom settings tab.
       MULTICA_WECOM_SECRET_KEY: ${MULTICA_WECOM_SECRET_KEY:-}
+      # 1 = log every inbound and outbound WeCom frame, including the first
+      # 120 runes of each message body, so a real-device session can be
+      # checked against the server afterwards. Off unless set; turn it on only
+      # for a debugging session and unset it when that session ends, because
+      # what it records is user message content.
+      MULTICA_WECOM_TRACE: ${MULTICA_WECOM_TRACE:-}
     restart: unless-stopped
 
   frontend:
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -708,6 +708,15 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 				// user in WeCom sees no response.
 				wecom.NewOutbound(queries, wecomSenders, slog.Default()).Register(bus)
 
+				// Frame tracing: off unless an operator asks for it. It
+				// records a bounded prefix of message text, so the fact that
+				// it is on has to be visible in the log it is writing into —
+				// otherwise a session gets left switched on and nobody
+				// notices message content accumulating.
+				if wecom.SetTrace(os.Getenv("MULTICA_WECOM_TRACE") == "1") {
+					slog.Warn("wecom: frame tracing ON — records message text; unset MULTICA_WECOM_TRACE when done")
+				}
+
 				slog.Info("wecom integration enabled (smart bot, long connection)")
 				// SINGLE-REPLICA CONSTRAINT: WeCom outbound (agent replies +
 				// inbox pushes) is delivered only by the replica holding each
--- a/server/internal/integrations/wecom/trace.go
+++ b/server/internal/integrations/wecom/trace.go
@@ -27,7 +27,7 @@ package wecom
 // prefix is capped rather than full, the cap counts runes so a Chinese
 // message is not cut mid-character, bearer tokens are redacted out of it, and
 // nothing here reads a credential field: the bot secret rides in the
-// aibot_subscribe body, which traceOut never descends into.
+// aibot_subscribe body, which traceOutFields never descends into.
 
 import (
 	"log/slog"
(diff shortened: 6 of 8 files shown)


server/internal/integrations/wecom/ws_sender.go
```diff
@@ -86,6 +86,10 @@ func (s *wsSender) write(frame map[string]any) error {
 	if err != nil {
 		return fmt.Errorf("wecom: marshal frame: %w", err)
 	}
+	// Before the writer mutex: the trace is of what we decided to send, and
+	// holding the socket lock while formatting a log line would let tracing
+	// change the serialization order it is meant to observe.
+	traceOut(s.log, frame)
```
Bohan-J: This is the blocking one: `traceOut` runs before `s.mu.Lock()`, so (a) under concurrent writers the trace order is not the wire order, and (b) a `SetWriteDeadline` / `WriteMessage` failure below still leaves an identical `dir=out` line. Recording the attempt inside the mutex and the outcome under the same `req_id` fixes both.

Bohan-J (changes requested):
Thanks for this — the product direction is right, and the safety work around it is solid: the switch defaults to off, the boot warning is loud, `traceOut` reads named fields instead of dumping the frame (so the `aibot_subscribe` secret never gets near the log), and redacting the binding token *before* the rune-safe truncation is the correct order, not the convenient one. The tests cover the cases that matter.

One blocking issue before merge.

## Must fix: the outbound trace doesn't record what was actually written

`server/internal/integrations/wecom/ws_sender.go:92` calls `traceOut` before `s.mu.Lock()`. Two consequences, both landing on the feature's core purpose:

**1. Trace order ≠ write order.** Heartbeats, replies and inbox pushes all reach `write` concurrently on one connection. Goroutine A can emit its `dir=out` line, get descheduled before taking the mutex, and let B trace-and-write first — the log then says A went out first when B did. The comment above the call argues that tracing inside the lock "would let tracing change the serialization order it is meant to observe," but it reads backwards to me: the mutex *is* the serialization point, so a record taken inside it matches the wire by construction, while one taken outside is merely correlated with it. If the worry is holding the lock while formatting, build the attrs outside and emit inside.

**2. A failed write is indistinguishable from a successful one.** When `SetWriteDeadline` or `WriteMessage` returns an error, the trace has already written a `dir=out` line identical to the success case. "Did this frame actually reach the wire?" is exactly what an operator turns this switch on to answer, and as written the log can't answer it. `traceOut`'s own doc comment says "nothing can be sent without appearing here" — still true, but the converse now holds too: things can appear here without being sent.

Suggested shape: record the send attempt inside the mutex-protected order, then record the outcome under the same `req_id`, so a reader can pair them and can spot an attempt with no outcome. `sendText` already mints a `req_id`; frames that don't carry one would need one to correlate.

Two regression tests would pin this down:

- concurrent writers on one `wsSender` → trace order matches the order frames actually reached `WriteMessage`
- a `wsConn` stub whose `WriteMessage` (and separately `SetWriteDeadline`) returns an error → the trace shows the failure rather than a clean send

## Non-blocking

- Worth a line in the self-hosting docs: who may read these logs, how long they are kept, and that flipping the switch requires a restart. The code is careful about what it writes; the operational half of "turn it on, then turn it off again" lives outside the repo.

seacen:
Fixed in c12f8cd46. You were right on both counts, and right that the comment defending the old placement read backwards — it's gone.

## Ordering

The record now happens under `s.mu`. Your framing is the correct one: the mutex is the serialization point, so a record taken inside it matches the wire by construction, while one taken outside is merely correlated with it.

I took your suggestion about the split. `traceOutFields` pulls the attributes out before the lock — that's the expensive half, a regexp redaction pass plus a rune-wise cut over the message body — and `traceOutAttempt` / `traceOutResult` emit inside it.

On the lock-scope trade-off, since you may ask: I'd have moved the emit inside even without the split. The section already holds `SetWriteDeadline` + `WriteMessage`, a socket write with a 10s budget, and two slog records don't register against that. And when the switch is off — every deployment that hasn't asked for a debugging session — `traceOutFields` returns nil on the first atomic load, so what the critical section gained in the default configuration is a nil check. The split is still worth having on its own: the redaction regexp is O(message length) and there's no reason for it to be serialized with anything.

## Outcome

Two lines per outbound frame now: `dir=out` when it's about to be written, `dir=out.done` with `ok=true`/`ok=false` once the socket has answered, carrying `stage=set_write_deadline|write_message` and the socket's error on a failure (through `tracePreview`, so an error string gets the same bounding and redaction as everything else here). Both under the same mutex, so an attempt and its outcome can't be split by another goroutine's frame.

One line wouldn't have done either job. Written before the write it can't report the outcome, which is the bug you found; written after it, a write that hangs or a process killed mid-write leaves nothing at all, which is the same silence by another route. `traceOut`'s doc comment is updated — you're right that the converse had started holding.

**One deviation from your suggested shape.** You suggested pairing on `req_id`, minting one for frames that don't carry it. I used a `seq` instead: a per-sender counter incremented under the writer mutex, never on the wire. Two reasons. `req_id` isn't really a key — a pong echoes the server's `req_id`, which can be empty or repeated, and minting our own there would change what we put on the wire for the sake of a logging feature. And `seq` doubles as the frame's position in the write order, so a reader recovers the ordering from a field instead of from how the lines happen to sit in the file. `req_id` is still on both lines whenever the frame carries one, so pairing a send with its server ack still works. Say the word if you'd rather have the minted req_id.

## Tests

`TestTraceOutOrderIsTheWireOrder` is the ordering one, and it's deterministic rather than a race that may or may not turn up. Writer A is held inside its own trace emission until writer B has run a whole `write()`. With the record under the mutex, A holds the mutex while it's parked, B blocks before recording anything, and A is recorded first and written first. With the record outside, A holds nothing, B records-writes-finishes, and the log claims A went first when B did.

It needed its own `slog.Handler`. The stock ones hold a mutex across the whole of `Handle`, including the write to the sink, so two goroutines emitting a trace line serialize inside the handler — which is the ordering under test, and would have made the test pass either way.

`TestTraceOutOrderUnderConcurrentWriters` is the same property under 16 concurrent writers, and also pins `seq` against the wire positions and every attempt against its outcome. `TestTraceOutRecordsAFailedWrite` / `TestTraceOutRecordsAFailedWriteDeadline` are the stub tests you asked for; the deadline one also asserts nothing reached `WriteMessage`, so the trace can't show a send that never happened.

## Reverse-verification

**Revert 1** — put the attempt back before `s.mu.Lock()`, seq bookkeeping untouched so the lock scope is the only variable. `go test -race -run TestTraceOutOrderIsTheWireOrder -count=30`: 30 failures out of 30 runs, identical every time.

```
--- FAIL: TestTraceOutOrderIsTheWireOrder (0.00s)
    trace_test.go:572: trace order [A B] != wire order [B A]: the trace records a frame as sent before one that reached WriteMessage earlier, so it cannot be read as the order the server saw
```

The concurrent-writers test fails on the same revert too — its detail varies run to run, but it never passed at `-count=5`:

```
--- FAIL: TestTraceOutOrderUnderConcurrentWriters (0.01s)
    trace_test.go:617: trace order != wire order under 16 concurrent writers
         trace: [W08 W12 W06 W10 W15 W01 W13 W07 W02 W04 W05 W11 W09 W03 W14 W00]
          wire: [W08 W12 W13 W02 W05 W11 W09 W10 W15 W01 W07 W04 W06 W03 W14 W00]
    trace_test.go:626: attempt 0 has seq="6", want "1"
    trace_test.go:626: attempt 2 has seq="4", want "3"
    trace_test.go:626: attempt 3 has seq="3", want "4"
    trace_test.go:626: attempt 4 has seq="1", want "5"
    trace_test.go:626: attempt 5 has seq="5", want "6"
```

**Revert 2** — drop the `traceOutResult` call. `-count=5`, all three outcome tests fail on every run:

```
--- FAIL: TestTraceOutRecordsAFailedWrite (0.00s)
    trace_test.go:687: 0 outcomes recorded, want exactly 1 — a send with no recorded outcome is indistinguishable from a successful one
--- FAIL: TestTraceOutRecordsAFailedWriteDeadline (0.00s)
    trace_test.go:719: 0 outcomes recorded, want exactly 1 — a send with no recorded outcome is indistinguishable from a successful one
--- FAIL: TestTraceOutRecordsASuccessfulWrite (0.00s)
    trace_test.go:741: 0 outcomes recorded, want exactly 1 — a send with no recorded outcome is indistinguishable from a successful one
```

Restored: `go test -race ./internal/integrations/wecom/ -count=5` green (with and without `DATABASE_URL` set, so the reclaim test runs rather than skips), `go build ./...`, `go vet` and `gofmt -l` clean.

## Non-blocking

Added a **WeCom frame tracing** subsection to `SELF_HOSTING_ADVANCED.md` covering the three things you named: it's read at boot so changing it needs a backend restart, anyone who can read the backend's logs can read the traced message text (a wider audience than the chat it came from), and retention is whatever the operator's log stack does since the backend keeps nothing itself. It also explains the `dir=out` / `dir=out.done` pair and what an unpaired attempt means. `.env.example` points at it.

Bohan-J (approved):
Final review. The blocking issue from my last round is properly closed, and I've taken the branch through the rebase myself — details below so nothing here is a surprise.

## What I pushed to this branch

Two things, both mechanical. The feature code is unchanged and still yours.

1. **Rebased onto `main`** (`f97db4fc5`). One conflict, in `wecom_channel.go`: `main` added the `botDisplayName` argument to `channelMessageFromCallback` on the same line your `traceInbound` call sits above. Both sides kept:

   ```go
   traceInbound(log, mc, mc.Text.Content)
   msg := channelMessageFromCallback(c.botID, c.botDisplayName, mc, env.Headers.ReqID)
   ```

   The PR's own diff against `main` is byte-identical to before the rebase apart from that line.

2. **One docs commit** (`38bab6a5d`), `SELF_HOSTING_ADVANCED.md`. The trace section said an `out` / `out.done` pair "tells you what the server saw." It doesn't — `ok=true` means `WriteMessage` returned, i.e. the local socket took the bytes. WeCom's verdict shows up later on the `dir=in` line carrying the same `req_id`. Reading `ok=true` as "delivered" would stop an operator exactly one step short of the `errcode` explaining the failure they're chasing. Reworded to say what the pair proves and where the platform's answer actually is. Revert it if you'd rather word it yourself.

## The must-fix is closed

`ws_sender.go` now extracts the trace fields outside the mutex and emits both lines inside it, with `s.seq++` assigned under the lock. Ordering is correct by construction rather than by correlation, and a failed write is now distinguishable — `ok=false` plus a `stage` that separates a deadline failure from a rejected write.

Two things you did better than what I suggested:

- Pairing on `seq` rather than `req_id`, with the reason written down: a pong echoes the server's `req_id`, which can be empty or repeated, so it was never a key. I'd suggested `req_id`; you were right not to take it.
- Keeping the expensive half — the regexp redaction and the rune-wise cut — outside the lock, so what runs in the critical section is a nil check when tracing is off. That answers the performance concern in the original comment without giving up the ordering guarantee.

The `outTrace` nil-means-not-traced design is also the right call: both lines hang off one decision, so the switch flipping mid-write can't leave an attempt whose outcome was suppressed.

## Verification

Run locally on the rebased head:

- `go build ./...` — passes
- `go vet ./internal/integrations/wecom/` — passes
- `go test ./internal/integrations/wecom/ -race -count=1` — passes
- `go test ./internal/integrations/wecom/ -race -run 'Trace|Write|Sender' -count=5` — 75/75, no flakes
- `gofmt -l` and `git diff --check` — clean

I also checked the new tests actually pin the property rather than just passing. Moving `traceOutAttempt` back outside the mutex makes `TestTraceOutOrderIsTheWireOrder` fail deterministically with `trace order [A B] != wire order [B A]`, and `TestTraceOutOrderUnderConcurrentWriters` fail on both the `seq` assignment and the attempt/outcome pairing. They're real regression tests, not coincidental passes.

One observation, no action needed: `credential_probe.go:143` also writes to a socket, but it's a separate short-lived probe connection, so `traceOutFields`' "the one place that writes the socket" is accurate for the session connection it's scoped to.

CI is running on the rebased head; it was green on all 9 checks before the rebase and the rebase changed one line.

✅ **Approve.** Thanks for the careful second pass on this — the ordering argument in the code comments is clearer than the one in my original review.

Bohan-J:
Merging this — thanks for the work, and particularly for the second round.

The revision didn't just move the log call inside the mutex; it worked out *why* `req_id` couldn't be the pairing key (a pong echoes the server's, which may be empty or repeated), introduced `seq` for the job, and kept the expensive extraction — the regexp pass and the rune-wise cut — outside the critical section. That's a better answer than the one my review asked for, and the reasoning is written down in the code where the next person will find it.

The tests are the part worth calling out. `TestTraceOutOrderIsTheWireOrder` pins the property with a deterministic interleaving rather than hoping a race shows up: moving the attempt log back outside the mutex fails it every time with `trace order [A B] != wire order [B A]`. Tests that fail for the right reason are rarer than they should be.

Also appreciated the care on the boundary this feature sits on — default off with a dedicated switch rather than `slog.Debug` (which this repo's `LOG_LEVEL` default would have turned on everywhere), redaction ordered before truncation so a token can't survive in halves, and an allowlist read that never descends into the body holding the bot secret.

Two commits were added on your branch on our side: a rebase onto `main` (one conflict — `main` added the `botDisplayName` argument on the line your `traceInbound` call sits above; both kept), and a docs commit correcting the trace section's claim that an `out`/`out.done` pair "tells you what the server saw" — `ok=true` means the socket took the bytes, while WeCom's verdict arrives on the later `dir=in` line with the same `req_id`.

All 9 checks green on the rebased head. Thanks again.

PR #6603, opened 2026-08-07
MUL-5904: fix(wecom): no slash command works in a group at all

seacen:
**Today, in a WeCom group, no slash command works at all.**

@-mentioning the bot is the only way to reach it in a group, and WeCom delivers the mention as literal text at the front of the body. What arrives is:

```
@Multica Bot /issue 登录失败
```

The shared command parsers read the first non-empty line and want the directive at the start of it. They see `@Multica` and classify the whole message as ordinary prose. So `/issue` files nothing, `/new` starts no fresh session, and nothing is said to the person either way — the message is stored and the agent answers as if they had just been chatting. Only 1:1 chats, where nobody types a mention, work.

## The defect

Two links were missing.

**The adapter never set `CommandText`.** `channelMessageFromCallback` left it empty, so `Router.Handle` fell back to `Text` (`channel/engine/router.go:186`) — and in a group `Text` starts with the mention.

**The session binder threw away any command source it was given.** `AppendMessage` hardcoded `CommandText: p.Message.Text`, so even a correct value from the adapter would not have reached `ParseIssueCommand`.

## Sibling parity

Both siblings already do exactly this, on both halves:

| | strips the mention → CommandText | binder prefers it |
|---|---|---|
| Slack | `slack/inbound.go:131` (`cleanText`) | `slack/resolvers.go:337-340` |
| Feishu | `lark/feishu_channel.go:139` (`lm.CommandBody`) | `lark/feishu_resolvers.go:206-209` |
| DingTalk | `dingtalk/inbound.go:134` | `dingtalk/resolvers.go:254` |
| WeCom | — none — | overwrote with `Text` |

The binder fix here is the same two lines Slack and Lark use:

```go
commandText := p.Message.CommandText
if commandText == "" {
    commandText = p.Message.Text
}
```

## What is stripped, and what is not

Only mentions at the very front. `"@bot ask @李雷 about yesterday"` keeps the colleague's name — rewriting it would change what the person said. The stored message body is untouched, so the transcript still shows who was addressed; only the command *source* is stripped.

## The bot's own name

`"@Multica Bot /issue …"` cut at the first space leaves `"Bot /issue …"`, which is not a command either. The callback carries no structured mention list, and the aibot API exposes no REST surface to ask the bot its own name, so the name has to be configured.

The Connect dialog gains an optional **"Bot name in chat"** field (`bot_name` on the install request, `bot_display_name` on the installation config, four locales). It is used for one thing: recognising where the mention ends.

- Blank keeps the whitespace heuristic — correct for a one-word name, and what every existing installation has today. No migration, no behaviour change for anyone who does not fill it in.
- Blank on a **re-install of the same bot** keeps the stored name. The field is optional, so an admin rotating a leaked secret leaves it empty; erasing the name there would silently put group commands back to the guess months after anyone typed it.
- **Swapping** to a different bot does not inherit the old name — it belongs to the old bot.

## Tests, reverse-verified

Every test below was run with the fix reverted and its actual failure output is pasted in. Nothing here passes against the old code.

**1. Restoring the pre-fix command source** (`CommandText: mc.Text.Content`, `SkipAgentRun: isIssueCommand(mc.Text.Content)`) — this is exactly what `main` does today:

```
--- FAIL: TestAnIssueCommandAfterAMentionIsStillACommand (0.00s)
    regression_group_mention_swallows_command_test.go:77: the shared parser did not see /issue.
        CommandText = "@Andrew /issue 登录失败" — the @-mention is still in front of the directive, so no issue is filed and nothing tells the person why.
--- FAIL: TestAFreshSessionCommandAfterAMentionIsStillACommand (0.00s)
    --- FAIL: .../mention_then_command (0.00s)
        the shared parser did not see /new in "@Andrew /new 重新分析一下".
            CommandText = "@Andrew /new 重新分析一下" — the request for a fresh session is silently dropped and the agent answers with the old conversation still in hand.
    --- FAIL: .../two_mentions_then_command (0.00s)
            CommandText = "@Andrew @Bowen /new 重新分析一下" — …
    --- FAIL: .../mention,_full-width_space,_command (0.00s)
            CommandText = "@Andrew　/new 重新分析一下" — …
--- FAIL: TestAMultiWordBotNameReachesTheParser (0.00s)
    CommandText = "@Multica Bot /issue 登录失败" — the bot's configured name never reached the mention strip
```

These drive the real path: a group `aibot_msg_callback` frame through `dispatchFrame` into the inbound handler, then the engine's own `ParseIssueCommand` / `ParseFreshSessionCommand` over what came out.

**2. Removing the configured-name branch** from `stripLeadingMentions` (leaving the whitespace cut):

```
--- FAIL: TestAMultiWordBotNameDoesNotSwallowTheCommand/two-word_name
    stripLeadingMentions("@Multica Bot /new 重新分析这批数据", "Multica Bot") = "Bot /new 重新分析这批数据", want "/new 重新分析这批数据"
--- FAIL: .../three-word_name
    stripLeadingMentions("@Acme Support Bot /issue 登录失败", "Acme Support Bot") = "Support Bot /issue 登录失败", want "/issue 登录失败"
--- FAIL: .../a_colleague's_mention_before_ours
    stripLeadingMentions("@李雷 @Multica Bot /new 分析", "Multica Bot") = "Bot /new 分析", want "/new 分析"
```

**3. Restoring `CommandText: p.Message.Text`** in the binder:

```
--- FAIL: TestSessionBinder_AdapterCommandTextWins (0.00s)
    CommandText = "@Multica Bot /issue 登录失败", want the adapter's stripped command source
```

**4. Removing the carry-forward** (DB-backed, real `Upsert` against Postgres):

```
--- FAIL: TestRotatingTheSecretKeepsTheBotName (0.07s)
    BotDisplayName after rotation = "", want it kept — a blank optional field erased the name, and every group slash command goes back to the whitespace guess
```

Making the carry-forward unconditional instead (dropping the same-bot guard) fails the other side:

```
--- FAIL: TestSwappingTheBotDoesNotInheritTheOldName (0.02s)
    the new bot inherited the previous bot's chat name "Multica Bot"
```

**5. Dropping `bot_name` from the request body** in the dialog:

```
FAIL  settings/components/wecom-tab.test.tsx > sends the bot's chat name when one is given, and omits it when not
AssertionError: expected undefined to be 'Multica Bot'
```

Tests that would have been meaningless are not here: the mention table alone would pass without any of the plumbing, so `TestAMultiWordBotNameReachesTheParser` drives the configured name from `wecomChannel` through the frame dispatch instead.

## Also worth naming

`SkipAgentRun` now reads the same stripped source the engine parses. Without that, the newly-working group `/issue` would file the issue *and* hand `"@Multica Bot /issue …"` to the agent, which answers "I don't recognize this slash command" underneath the confirmation. 1:1 has never done that.

## Checks

- `go build ./...` and `go vet ./...` clean.
- `go test ./internal/...` with a migrated Postgres: green except `internal/daemon`'s `TestProbeAgentCLIs_QoderResolvesViaLoginShell`, which fails identically on unmodified `main` on this machine (it picks up a real `qodercli` on `PATH`).
- `packages/views` vitest: 3732 passed, and the one failure (`layout/sidebar-resize.test.tsx`) fails identically on unmodified `main`.
- `tsc --noEmit` clean for `packages/core` and `packages/views`; eslint clean on the touched files.
- All four locales verified to carry identical key sets after the change (the only pre-existing divergence, `properties.usage_count_one`, is an English plural rule and unrelated).

Nothing here touches the durable outbound queue, the rate gate, or the single-replica docs.


server/internal/integrations/wecom/ws_frame.go
```diff
@@ -160,6 +165,12 @@ func channelMessageFromCallback(botID string, mc aibotMsgCallback, reqID string)
 		chatID = senderID
 	}
 
+	// The command source is the sender's own words with the addressing taken
+	// off the front. In a group the @-mention IS how you reach the bot, so it
+	// arrives glued to whatever was typed after it — "@Andrew /new" is a person
+	// asking for a fresh session, not prose that happens to contain a word.
+	command := stripLeadingMentions(mc.Text.Content, botDisplayName)
```
Bohan-J: This runs for every chat type, but `chatType` is already computed a few lines above.

In a 1:1 chat nobody needs to address the bot, so a leading `@` is the sender talking about someone else. `@李雷 /issue 帮我问问他` becomes `/issue 帮我问问他` here, which files an issue and then sets `SkipAgentRun` — the sender gets an issue they did not ask for and no reply.

Suggest gating it:

```go
command := mc.Text.Content
if chatType == channel.ChatTypeGroup {
    command = stripLeadingMentions(mc.Text.Content, botDisplayName)
}
```

and adding a P2P case in `ws_frame_test.go` asserting the mention survives and the text is not classified as a command.

server/internal/integrations/wecom/installation_bot_name_test.go
```diff
@@ -0,0 +1,139 @@
...
+			t.Fatalf("seed bot-name fixture: %v", err)
+		}
+	}
+	exec(`INSERT INTO workspace (id, name, slug, description) VALUES ($1, 'wecom bot name', 'wecom-bot-name', '') ON CONFLICT (id) DO NOTHING`, wcNameWS)
+	exec(`INSERT INTO agent_runtime (id, workspace_id, name, runtime_mode, provider)
+VALUES ($1, $2, 'wecom bot name runtime', 'local', 'multica_daemon') ON CONFLICT (id) DO NOTHING`, wcNameRuntime, wcNameWS)
+	exec(`INSERT INTO agent (id, workspace_id, name, runtime_mode, runtime_id)
+VALUES ($1, $2, 'wecom bot name agent', 'local', $3) ON CONFLICT (id) DO NOTHING`, wcNameAgent, wcNameWS, wcNameRuntime)
+	t.Cleanup(clean)
+	return ctx, pool, newReclaimSvc(t, pool)
```
Bohan-J: `newReclaimSvc` returns `(*InstallationService, *fakeProbe)` on current `main`, so this single-value call breaks compilation of the whole `wecom` test package once merged:

```
multiple-value newReclaimSvc(...) in single-value context
```

No textual conflict, so CI on this head does not catch it. Rebase and take the second return value (discard it with `_` if the fixture does not need the probe).

Bohan-J (changes requested):
Thanks for this — the product problem is real and the fix is the right shape. A group `@`-mention is how you reach the bot at all in WeCom, so a slash command behind it should work exactly like it does in a 1:1, and setting `CommandText` from cleaned text is what Slack and Lark already do. Keeping the stored message untouched so the transcript still shows who was addressed is the right call too.

Two things need to change before this can go in.

**1. The mention stripping is not scoped to groups, which changes 1:1 semantics.**

`channelMessageFromCallback` already computes `chatType`, but `stripLeadingMentions` runs unconditionally on every message. In a 1:1 chat, `@李雷 /issue 帮我问问他` is a person talking *about* a colleague — today it is prose, because `CommandText` was empty and classification fell back to `Text`. After this change it becomes `/issue 帮我问问他`, which files an issue the sender never asked for and, via `SkipAgentRun: isIssueCommand(command)`, also suppresses the normal reply. So the sender gets a stray issue and silence.

The justification in the comment ("the @-mention IS how you reach the bot") only holds for groups — in a 1:1 nobody has to address the bot, which is exactly what the commit message says. Please gate the strip on `chatType == channel.ChatTypeGroup` and pass the raw content through for P2P.

**2. The new test file does not compile against current `main`.**

`installation_bot_name_test.go:53` calls `newReclaimSvc(t, pool)` in single-value context. On `main`, `newReclaimSvc` now returns `(*InstallationService, *fakeProbe)` — the credential-probe work landed after this branch was cut. There is no textual conflict, so the PR shows as mergeable, but the merged tree fails to build:

```
installation_bot_name_test.go:53: multiple-value newReclaimSvc(...) in single-value context
```

That takes down the whole `internal/integrations/wecom` test package, so the green checks on this head commit are not telling you about the state after merge. Please rebase on `main`, take the probe return value, and re-run `go test ./internal/integrations/wecom -count=1`.

**Test coverage to add along with fix 1:** `ws_frame_test.go` currently covers the group path and the P2P chat-ID fallback, but nothing asserts P2P *command* semantics. A case with `ChatType: "single"` and content `@李雷 /issue …` should assert that `CommandText` keeps the mention, that `ParseIssueCommand` / `ParseFreshSessionCommand` do not match it, and that `SkipAgentRun` stays false. That is the regression this PR would otherwise introduce silently.

**Non-blocking:** in `wecom-tab.tsx`, the component returns the connected status row early for an active installation, so the Connect dialog — and with it the new bot-name field — is unreachable once a bot is connected. An existing installation whose bot name contains a space has to disconnect and reconnect to pick up the fix. That may be an acceptable migration cost for a small set of installs, but it is worth either saying so explicitly in the UI or exposing the name for editing on an active installation. Your call.

The rest reads well: the optional JSON field on the installation config is safe for old rows and mixed-version clients, empty-means-heuristic preserves today's behaviour for one-word names, and not inheriting the name when swapping to a different bot is the right default.

seacen:
Both blocking points are addressed: the mention strip is now gated on `chatType == channel.ChatTypeGroup`, and p2p passes the raw content through, so a 1:1 message keeps the semantics it has today. The compile break you found was already gone after the rebase — I re-ran the package against a migrated database to confirm it. I also took the non-blocking UI point, in the cheaper of the two forms you offered.

---

## 1. Mention stripping not scoped to groups

**Fixed.** You are right, and the failure mode you describe is the worse of the two: the sender gets an artifact they never asked for *and* no reply, so there is nothing on screen to tell them what happened.

`ws_frame.go:179` now reads:

```go
command := mc.Text.Content
if chatType == channel.ChatTypeGroup {
    command = stripLeadingMentions(mc.Text.Content, botDisplayName)
}
```

Passing the raw content through for p2p rather than leaving `CommandText` empty is deliberate, and it is the conservative choice rather than a second behaviour change smuggled in beside the first. Before this PR `sessionBinder.AppendMessage` set `CommandText: p.Message.Text` unconditionally, and the shared classification in `channel/message.go` falls back to `Text` when `CommandText` is empty. So for p2p, `CommandText == Text` is byte-for-byte what both paths already produced. The one place the two are not interchangeable is `issueCommandLineBounds`, which uses `CommandText` to disambiguate a directive line in an enriched body — and wecom does no enrichment, so `Body == Text` there too.

I updated three comments that had gone stale against the gate: the `stripLeadingMentions` doc now says group-only and why, the `CommandText` comment says p2p gets `Text` verbatim, and the `SkipAgentRun` comment records that it has to read the *same* source as the strip — reading it off the raw text instead would put back exactly the "file an issue, then stay silent" behaviour the gate exists to prevent.

## 2. `installation_bot_name_test.go:53` does not compile against `main`

**Already fixed** — by the rebase, before this round. The line is now:

```go
svc, _ := newReclaimSvc(t, pool)
```

with a comment saying the probe is the reclaim tests' always-succeeds fake and these tests are about what the config JSONB holds, not about the proof of control.

Your diagnosis of *why* CI was not catching it was right, and it is the part worth keeping: no textual conflict means `mergeable` says nothing about whether the merged tree builds. So I did not take the green checks as evidence here. I stood up a migrated Postgres and ran the package for real — `go vet ./...` over the whole server tree as well as the tests, because a return-arity change is exactly the class of breakage that `go build` alone lets through. Output in the verification section; the DB-backed tests ran rather than skipped, which the run log shows.

## 3. Test coverage for p2p command semantics

**Added**, in `ws_frame_test.go` where you pointed, calling `channelMessageFromCallback` directly — the function under test needs no channel harness, and the file already exercises it that way.

`TestChannelMessageFromCallback_P2PMentionIsProseNotACommand` is a two-case table (`/issue` and `/new` after a colleague's name) asserting all three things you listed: `CommandText` equals the content untouched, neither `ParseIssueCommand` nor `ParseFreshSessionCommand` matches it, and `SkipAgentRun` stays false. It passes `"Multica Bot"` as the display name on purpose, so the test also pins down that it is the chat type deciding and not whose name happens to be at the front.

I added a second one, `TestChannelMessageFromCallback_P2PCommandStillWorks`: a plain 1:1 `/issue 登录失败` with no mention must remain a command with `SkipAgentRun` true. The gate has two ways to be wrong and your case only covers one — a gate that stopped setting `CommandText` for p2p at all would pass your case and silently break the path that worked before this PR existed.

Both were reverse-verified: I removed the fix and confirmed each fails with a message naming the user-visible consequence, not a diff of two strings. Verbatim output below.

## 4. Non-blocking: the Connect dialog is unreachable on an active installation

**Fixed, in the "say so in the UI" form** rather than by exposing the field for editing.

Your read of the component is right: `wecom-tab.tsx:281` returns the status row whenever an installation with `status === "active"` matches the agent, so the dialog and the new field are both behind that. I checked the migration path is real before relying on it — `Revoke` flips status to `revoked` and preserves the row, the `existing` lookup filters on `active`, so the Connect CTA comes back after a disconnect; and a typed name beats the carry-forward, since that only applies when `p.BotDisplayName == ""`.

What made this worth changing was the hint copy, which was actively wrong: it said "Update this if you rename the bot in the WeCom console", and there is no way to do that from a connected installation. It now names the disconnect-and-reconnect path, in all four locales.

I stopped short of exposing the name for editing on an active installation. That means an edit affordance and a place to put it, on a component whose current job is a binary connect/disconnect — a UI change with its own review surface, in a PR that is otherwise a routing fix. Happy to do it as a follow-up if you would rather have it than the copy; say the word and I will open one.

## Notes for whoever merges this alongside the other open wecom PRs

None of these are changes I made, and none of them break this PR on its own — flagging them because my change touches two of the same lines.

- **`SkipAgentRun` / `isIssueCommand`.** This PR now computes `command` in two steps, and #6605 computes `text` from the parsed body. The combination that keeps both fixes is to feed #6605's `text` into this gate rather than `mc.Text.Content` — taking either side alone drops the other's fix. Splitting the strip out of the expression makes that a one-line merge instead of an untangling.
- **`isIssueCommand` itself.** This branch still carries the hand-written mirror of `engine.ParseIssueCommand`. #6593 replaces it with a delegation to the real parser; resolving that conflict in this branch's favour would bring back the U+3000 full-width-space bug #6593 fixed. Take theirs — this PR only changes what is passed *to* it, never what it does.
- **`channelMessageFromCallback`'s signature.** Unchanged by this round; it is still `(botID, botDisplayName, mc, reqID)`. #6605 adds `text` to the same list, so whoever lands second needs one combined signature and the call sites updated with it.
- **`installation.go` carry-forward.** Untouched here. It carries `BotDisplayName` only when the bot id is unchanged; #6600 carries `PrincipalUserID`. Both fields have to survive the merge.
- **The receipt hunk in `wecom_channel.go`.** Untouched here — this branch predates #6605's `ownText` routing, so it still gates on `mc.MsgType != "text"`. Merging should keep #6605's structure and thread `botDisplayName` through it, not take this branch's version of those lines.

---

## Verification

Rebased on `main` (`6b5d26b2` is an ancestor of the head), against a migrated throwaway Postgres.

`go build ./...`, `go vet ./...`, `gofmt -l` on the changed files:

```
=== gofmt (my files) ===
(clean)
=== go build ./... ===
OK
=== go vet ./... ===
OK
```

`go test -race -count=1`:

```
ok  	github.com/multica-ai/multica/server/internal/integrations/wecom	7.487s
ok  	github.com/multica-ai/multica/server/internal/handler	28.090s
```

The DB-backed tests from point 2 ran rather than skipped:

```
=== RUN   TestRotatingTheSecretKeepsTheBotName
--- PASS: TestRotatingTheSecretKeepsTheBotName (0.09s)
=== RUN   TestSwappingTheBotDoesNotInheritTheOldName
--- PASS: TestSwappingTheBotDoesNotInheritTheOldName (0.05s)
=== RUN   TestTheBotNameSurvivesTheConfigRoundTrip
--- PASS: TestTheBotNameSurvivesTheConfigRoundTrip (0.00s)
=== RUN   TestUpsert_ProbesAndRefusesBeforeReclaimingARevokedOwner
--- PASS: TestUpsert_ProbesAndRefusesBeforeReclaimingARevokedOwner (0.05s)
```

**Reverse-verification 1 — remove the group gate** (`command := stripLeadingMentions(mc.Text.Content, botDisplayName)`, unconditional, i.e. the code you reviewed):

```
--- FAIL: TestChannelMessageFromCallback_P2PMentionIsProseNotACommand (0.00s)
    --- FAIL: TestChannelMessageFromCallback_P2PMentionIsProseNotACommand/fresh-session_directive_after_a_colleague's_name (0.00s)
        ws_frame_test.go:83: CommandText = "/new 的排期你问一下", want "@李雷 /new 的排期你问一下" untouched — in a 1:1 the leading @ is a colleague's name in the sender's sentence, not addressing the bot
        ws_frame_test.go:91: "@李雷 /new 的排期你问一下" would throw away the running session the sender was in the middle of
    --- FAIL: TestChannelMessageFromCallback_P2PMentionIsProseNotACommand/issue_directive_after_a_colleague's_name (0.00s)
        ws_frame_test.go:83: CommandText = "/issue 帮我问问他", want "@李雷 /issue 帮我问问他" untouched — in a 1:1 the leading @ is a colleague's name in the sender's sentence, not addressing the bot
        ws_frame_test.go:87: an issue titled "帮我问问他" would be filed from "@李雷 /issue 帮我问问他", which the sender never asked for
        ws_frame_test.go:94: SkipAgentRun = true for "@李雷 /issue 帮我问问他"; the sender's question gets no reply at all
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.528s
```

**Reverse-verification 2 — the gate present but blanking `CommandText` for p2p** (`command := ""`), the other way the gate can be wrong:

```
--- FAIL: TestChannelMessageFromCallback_P2PCommandStillWorks (0.00s)
    ws_frame_test.go:114: CommandText = "" — a 1:1 /issue stopped being a command
FAIL
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.529s
```

Restored, and the package is green again.

Frontend, for the locale change:

```
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

`gofmt -l` over `server/internal/` reports 22 files on this Go toolchain (1.26.5), the same 22 before and after this change and none of them mine — pre-existing, and not something I touched here.

Bohan-J (approved):
Both blockers are addressed at `f36b8948` — verified independently, this supersedes my earlier `CHANGES_REQUESTED`.

**1. Mention stripping is now scoped to groups.** `ws_frame.go:179-181` starts from `mc.Text.Content` and only strips when `chatType == channel.ChatTypeGroup`, so a 1:1 `@李雷 /issue …` keeps its mention and stays prose. The two regression tests are exactly the ones that were missing: `TestChannelMessageFromCallback_P2PMentionIsProseNotACommand` covers both the `/issue` and `/new` directives after a colleague's name, and `TestChannelMessageFromCallback_P2PCommandStillWorks` pins the pre-existing bare-command path so the gate cannot regress it in the other direction.

**2. The test fixture compiles against `main`.** `installation_bot_name_test.go:55` now takes `svc, _ := newReclaimSvc(t, pool)`.

Verified on a merge tree of this head with current `main` (`c4cd1eba`), not just on the branch — that was the point of the second blocker, since the break was semantic rather than a textual conflict:

- `git diff --check` — clean
- `go build ./...` — pass
- `go vet ./internal/integrations/wecom ./internal/handler` — pass
- `go test ./internal/integrations/wecom -count=1` — pass (5.6s)
- The P2P, multi-word-name and mention-of-somebody-else tests all run and pass; `TestRotatingTheSecretKeepsTheBotName` skips locally for want of a test Postgres, and is covered by the `backend` job on CI

CI is 10/10 green on this head.

On the non-blocking UI point, documenting the disconnect-and-reconnect path across all four locales is a fine resolution — in-place editing of the chat name on an active installation is a reasonable follow-up, not something to hold this on.

Thanks for the quick turnaround, and for pinning both directions of the P2P behaviour in tests rather than just the bug.

PR #6604, opened 2026-08-07
MUL-5905: feat(wecom): send the files an agent produced into the chat

seacen:
> **Stacked on #6590.** The first two commits are that PR (the WebSocket
> request/ack layer this builds on). **Review from `dd0296d5f` onward** —
> `dd0296d5f` and `4213c92a1` are the two commits belonging to this PR. The
> diff will shrink to just those once #6590 merges.

An agent that produced a file and answered in WeCom delivered only the
sentence.

`multica attachment upload` had already run, the object was in storage, and
`CompleteTask` had bound the row to the assistant message — that half is
platform-agnostic and already worked. What was missing was the last hop.
Everything downstream of that bind assumed a chat window in a browser, so a
WeCom conversation was told it could not take files at all. The user read
"here is the forecast" and got no forecast, with nothing in the logs to
explain it. When the agent produced a file and said nothing, the empty
completion returned early and the artifact was dropped silently.

## What this adds

**`media_upload.go`** — the file goes up the same long connection everything
else uses. WeCom has no REST endpoint for bot media, so it is three cmds and
no access_token: `aibot_upload_media_init` → `_chunk` × N → `_finish`. All
three say something beyond "accepted", so all three go through the
`request()` layer #6590 generalized — this leans on that rather than
re-inventing a verdict path.

**`outbound_media.go`** — the caller. After the answer is out, it reads the
bound attachment out of object storage, decides what to call it, and pushes
it in as its own message. `WithAttachments` is what turns it on; `router.go`
passes it only when the deployment configured storage, so a deployment
without object storage behaves exactly as before.

**The daemon half** — `ChannelCarriesFiles`, and the branches in the chat
prompt and the runtime brief that read it.

## Three decisions worth naming

**The answer goes first and the file never blocks it.** An upload is
megabytes and round trips and it can fail; the sentence the agent wrote must
not wait behind one or be lost to one. Delivery runs after the words are
out, on its own goroutine and its own budget, and its worst outcome is one
extra line saying a file did not make it. The semaphore is acquired *inside*
the goroutine, never before the spawn — `Bus.Publish` is synchronous on the
completion goroutine, so blocking out there would wedge the very path the
spawn exists to protect.

**A lost ack and a refusal are opposite failures.** A chunk whose verdict
never came is offered once more; the protocol makes re-sending idempotent,
and losing one slice of a hundred must not cost the file. A chunk the server
*refused* is not retried, because the refusal is its answer and will be its
answer again. The push of the finished media follows the same rule in
reverse: a timeout is reported as-is rather than retried, because the same
`media_id` sent twice puts the file on the user's screen twice with nothing
to undo. A "did it land?" in the log beats a duplicate.

**WeCom validates bytes against the msgtype**, so what to call a file is a
decision, not a lookup. A `.pptx` declared as an image is refused rather than
converted, and each kind has its own ceiling — so a kind past its ceiling is
demoted to a plain file rather than sent and refused. A file card the user
can open beats a photo the server rejected.

## The brief, and why the wording is what it is

The second commit is the half that makes the capability reachable. Both the
per-turn chat prompt and the runtime brief decided file delivery by asking
*"is there a channel at all"*. Any non-empty `ChatChannelType` got the
text-only paragraph. So an agent on WeCom was instructed **not to use the one
command that would have delivered its file** — the capability shipped and the
agent was told it did not exist.

The delivery question is per-adapter, and `ChannelCarriesFiles` is where that
answer now lives. Both call sites branch three ways instead of two.

**The new copy is deliberately built from main's existing web-chat branch
rather than written fresh.** All three, side by side:

*Web chat (on main today, unchanged):*
> To include a file or image you produced in your reply, run `multica
> attachment upload <local-path>`. The file binds to your reply automatically
> and **appears as an attachment card below it** even if you paste nothing.
> The command also returns a `markdown` snippet you may paste on its own line
> to place the item where you want it (files render as a card, images
> inline).

*WeCom (new):*
> To include a file or image you produced in your reply, run `multica
> attachment upload <local-path>`. It binds to your reply and Multica **sends
> it into the WeCom conversation as a separate message right after your text
> — there is no way to place it inline**, so write your reply to read
> correctly with the file arriving after it.

*Slack and Lark (on main today, unchanged):*
> This reply is delivered to %s as text. **You cannot attach a file to it**:
> `multica attachment upload` binds to a Multica chat reply, which this is
> not. If you produce a file, describe it in words…

The first clause is identical between web chat and WeCom, word for word, and
they diverge exactly where the platforms diverge. Slack and Lark keep their
text-only wording untouched. The "separate message" clause is the one worth
having: the long connection has no `msg_item`, so nothing can be embedded in
a reply. An agent told only that files work here writes "see the chart below"
and then nothing appears below.

## Two of our own defects fixed while porting, not carried over

1. **Comments that asserted the binary this change falsifies.** `prompt.go`
   said the branch "keys off *is there a channel at all*"; the brief said
   "*any non-empty channel type* means the reply leaves Multica for an
   external IM platform, where `attachment upload` has nothing to bind to."
   Both were true when written and both are now false. Rewritten rather than
   carried over — a comment that contradicts the code beneath it is worse
   than no comment.

2. **`ChannelCarriesFiles` exhaustiveness.** It now names every channel type
   this package knows rather than leaving Slack and Lark to fall through the
   default, so adding a fourth adapter surfaces here as a decision to make.
   An unrecognized type — a channel the server added that this daemon build
   has no constant for — answers `false`. That is the safe direction: the
   agent describes its file in words, and the worst case is a deliverable
   that could have been sent was not. `true` the other way would have the
   agent promise a delivery nothing performs. There is a test for the unknown
   case specifically.

## Deliberately not included

The durable outbound queue and the rate gate are **not** ported — superseded
by #6581. The single-replica constraint is not re-stated; it already reads
correctly on this base. The passive-reply (`aibot_respond_msg`) fallback is
also dropped: it is addressed by the req_id of the callback that opened the
turn, and this delivery path holds no such req_id on this base, so porting it
would have shipped unreachable code. `sendMedia` uses the push alone and
returns a refusal honestly.

## Testing

`go build ./...` and `go vet ./...` clean across the module. Full package
suites green:

```
ok  github.com/multica-ai/multica/server/internal/integrations/wecom   10.6s
ok  github.com/multica-ai/multica/server/internal/daemon/execenv        4.1s
ok  github.com/multica-ai/multica/server/cmd/server                     4.2s
```

`go test -race` on the wecom package is clean (the chunk uploader runs four
in flight, and delivery runs on its own goroutine, so this one matters).

One pre-existing failure in `internal/daemon` is unrelated and fails
identically on the base commit with this branch stashed —
`TestProbeAgentCLIs_QoderResolvesViaLoginShell`, which picks up a real
`qoderclicn` installed on the test machine:

```
agents_probe_qoder_test.go:36: qoder path = "/Users/…/.local/bin/qoderclicn",
    want /fake/npm-global/bin/qodercli
```

### Reverse-verification

Every new test was checked by reverting the fix it covers and confirming it
fails — **20 of 20 produced a genuine failure**, each restored afterward.
Selected real output:

```
[the delivery call itself — processEvent never hands the file over]
--- FAIL: TestProcessEvent_SendsTheAnswerAndThenTheFile
    outbound_media_test.go:151: upload init frames = 0, want 1 — the file was never uploaded
--- FAIL: TestProcessEvent_EmptyCompletionStillDeliversABoundFile
    outbound_media_test.go:212: media sends = 0, want 1 — the agent's file was thrown away

[the conditioned empty-completion return]
--- FAIL: TestProcessEvent_EmptyCompletionStillDeliversABoundFile
    outbound_media_test.go:212: media sends = 0, want 1 — the agent's file was thrown away

[the retry on a lost chunk verdict]
--- FAIL: TestUploadMediaChunk_ResendsAChunkWhoseVerdictNeverCame
    media_upload_test.go:202: uploadMedia gave up on a lost ack instead of
    offering the chunk again: wecom: upload chunk 0: wecom: timed out waiting
    for the server verdict

[the guard that stops a REFUSED chunk being offered again]
--- FAIL: TestUploadMediaChunk_DoesNotResendARefusedChunk
    media_upload_test.go:229: chunk frames = 2, want 1 — a refusal must not be offered again

[reporting a lost push verdict instead of swallowing it]
--- FAIL: TestSendMedia_ReportsALostAckWithoutSendingAgain
    media_upload_test.go:250: a lost push ack reported as <nil>, want errAckTimeout

[the chunk split's boundary — off-by-one corrupts the file rather than failing]
--- FAIL: TestUploadMedia_ChunksTheFileAndSealsIt
    media_upload_test.go:186: the chunks do not reassemble to the original file

[telling the user a file did not make it]
--- FAIL: TestSendAttachments_TellsTheUserWhenAFileFailed
    outbound_media_test.go:264: text sends = [See the attached dump.], want the
    answer and a note that the file failed

[ChannelCarriesFiles answering true for WeCom]
--- FAIL: TestChannelCarriesFiles/wecom_delivers_files
    channel_type_test.go:58: ChannelCarriesFiles("wecom") = false, want true

[the chat prompt's three-way delivery branch]
--- FAIL: TestBuildChatPromptTwoLayerChannelPolicy/wecom:_upload,_no_history
    prompt_test.go:646: upload guidance present=false, want true
    --- output ---
    This reply is delivered to WeCom as text. You cannot attach a file to it…

[the brief's three-way delivery branch]
--- FAIL: TestBriefSurfaceDeliveryPolicy
    runtime_config_delivery_test.go:125: surface=chat_wecom: brief must NOT
    carry "text-only" (wrong surface's delivery mechanism)
```

The rest cover `clipUTF8`'s rune-boundary cut (a cut through a Chinese
character sends the server broken UTF-8), `readObject`'s memory cap, the
per-kind size ceilings, filename reduction to a single path segment, the
bounded pending-delivery backlog, and `validate()`'s two rejections.

Existing tests were extended rather than replaced: the WeCom row was added to
`TestBuildChatPromptTwoLayerChannelPolicy`'s matrix and a `chat_wecom`
surface to `TestBriefSurfaceDeliveryPolicy`, alongside the Slack and Lark
rows they already pinned.


seacen:
Integration note: **this PR and #6605 each add a file defining `cleanMediaFilename` and `mediaExtension`.** There is no textual overlap, so git merges both cleanly — and then the package does not compile (duplicate declarations in the same package).

`cleanMediaFilename` is byte-identical in the two. `mediaExtension` is not: #6605's is a strict superset — it strips MIME parameters itself and handles `application/pdf`. So whichever merges second should delete its own copies and keep #6605's, folding any doc-comment detail worth keeping into the survivor.

Found while merging all the WeCom PRs onto one branch locally. Nothing to change here if this one lands first; recording it so the second merge is a two-minute fix rather than a puzzle.

seacen:
Rebased onto main. One textual conflict; the work was in what did *not* conflict.

**The conflict — `outbound_test.go`.** #6591 added a large block to `fakeOutboundQueries`
(the task map, the `channel_ingested` stamp, `failStampNotSet`, the `fileTask` helpers) at
the same place this branch adds `attachments` / `attachmentsErr` and the
`ListAttachmentsByChatMessage` method. Both sides are pure additions to the same fake; kept
both, in both the struct and the method block.

**`outbound.go` merged clean, and that is the part worth checking.** This branch and #6591
both insert into `processEvent`, in different places, so git had nothing to report. The
resulting order on the new head is right:

```
128  binding lookup
143  content == "" && !mayCarryAttachments   (read-only)
160  chatDoneTaskID          ─┐
164  GetAgentTask             │  origin gate
172  TaskInputIsChannelIngested│
176  if !deliver { return }   ─┘
210  sender.sendTextCtx        ← the words
216  o.deliverAttachments      ← the file
```

The gate sits ahead of both sends. `mayCarryAttachments` is above it but only reads the
event payload and the nil-ness of the object store — it widens who reaches the gate and
sends nothing — so #6591's "everything above this point is a read" still holds.

**Nothing was holding that order, and the branch's own tests were not even reaching it.**
Two separate findings:

1. Every media test was written before #6591, so none of them set a task id or said where
   the question was asked. On the rebased branch all five returned at the gate and the
   suite was red — "text sends = []". Fixed in the shared rig: `chatDoneEvent` carries a
   task id and `oneAttachmentQueries` files it as asked-over-WeCom.
2. Once they passed again, nothing tested that the gate refuses the *file*. Added
   `TestProcessEvent_AWebUIAnswerDeliversNeitherWordsNorFile`. It asserts no text frame, no
   media frame, and — the part that matters — **no upload init frame**, because the upload
   is itself a write to WeCom: hoisting delivery above the gate would leak the file even if
   the send behind it were refused.

   Reverse-checked by moving `deliverAttachments` above the gate: `go build` still passes,
   and the new test fails with the file in the room. That is the check that was missing.

**Constants.** Probed each by setting it to an obviously wrong value and re-running: the
three `aibot_upload_media_*` command names, `mediaChunkBytes`, `maxMediaChunks`,
`mediaChunkAttempts` and `mediaChunkParallelism` are all held by at least one test.

`go build`, `go vet`, `gofmt` clean. `go test -race` green on
`./internal/integrations/wecom/...`, `./internal/daemon/execenv/...` and `./cmd/server/...`.
`internal/daemon` has one failure, `TestProbeAgentCLIs_QoderResolvesViaLoginShell`, which is
pre-existing on main and machine-specific — it picks up a real `qoderclicn` on the
developer's PATH instead of the fixture. It fails identically on a worktree that does not
touch `daemon`.

multica-eve (dismissed):
Today, someone can ask an agent in WeCom to create a file but receive only the written response. This change aims to make the file arrive in the same conversation immediately afterward. That is the product behavior we want.

I am requesting changes because the current head still has three merge blockers:

1. **File-delivery capability is inferred from the channel type instead of the actual deployment and server capability.** `server/internal/daemon/execenv/channel_type.go:87-95` returns `true` for every WeCom turn, while `server/cmd/server/router.go:716-720` enables attachment delivery only when object storage is configured. A new daemon paired with an older server has the same compatibility problem: the prompt promises delivery, but the server has no final delivery hop. Please have the server provide the effective per-turn capability, defaulting to `false` for an older server or a deployment without storage, and add no-store/mixed-version prompt tests.

2. **The upload limits and concurrency still diverge from the official WeCom implementations.** `server/internal/integrations/wecom/media_upload.go:56-76,199-206` permits 50 MiB and always uses four concurrent chunk uploads; `server/internal/integrations/wecom/outbound_media.go:227-249` also reads up to that 50 MiB ceiling into memory. The official Node SDK reduces concurrency to two for uploads with more than ten chunks because higher concurrency can cause backend system errors ([source](https://github.com/WecomTeam/aibot-node-sdk/blob/80615b987ef69c6028ad764924609247c0725955/src/client.ts#L397-L406)), and the official WeCom OpenClaw plugin applies a 20 MiB hard file limit ([source](https://github.com/WecomTeam/wecom-openclaw-plugin/blob/1a91ef7300de7274de8d74e4a566cf3b6e569a25/src/const.ts#L111-L124)). Please enforce 20 MiB before reading/uploading, use adaptive concurrency, and cover the 20 MiB boundary and `>10`-chunk case.

3. **Failure handling can still silently drop a promised file or report an uncertain send as a definite failure.** `server/internal/integrations/wecom/outbound_media.go:120-159` only logs when the pending queue is full, waiting for a slot times out, or the attachment query fails. Conversely, `server/internal/integrations/wecom/media_upload.go:280-291` correctly notes that a send-ack timeout means the file may already have arrived, but `outbound_media.go:173-195` folds that into the same definitive failure message as a rejection. Please distinguish at least `delivered`, `definitely_failed`, and `unknown`: do not retry an unknown send, use non-definitive wording for it, and make a best effort to notify the user for definite local scheduling/query failures. Please add regression tests for each state.

The rebased source-origin gate and `TestProcessEvent_AWebUIAnswerDeliversNeitherWordsNorFile` are good additions and should remain. At head `149165572`, the targeted WeCom tests, WeCom race test, daemon prompt byte-identity test, `git diff --check`, and GitHub CI all pass; those green checks do not cover the blockers above.

seacen:
All three are fixed at `24d9a3f`, rebased onto `main` (#6595 and #6599 landed this morning; the rebase was clean, no conflicts).

Thank you for pinning the two sources to commits — I read both, and they confirm what you wrote. The SDK carries one extra detail that changed what I implemented; it's below.

## The three have one root cause, and it's worth naming

Each blocker is the same defect in a different place: **a promise made where the thing that has to keep it cannot be seen.**

- The daemon promised an agent it could send a file, from a fact the daemon cannot see (whether the server has storage).
- The upload path promised the platform 50 MB, from a number that describes the framing rather than what the platform accepts.
- The delivery path promised the user a definite verdict, from an outcome the socket never gave us.

So the fixes have a common shape too: in each case the claim now comes from whatever actually knows the answer, and where nothing knows, the code says so instead of picking the confident side.

## 1. Capability now comes from the server, per turn

The claim response carries a new `chat_channel_delivers_files`. `handler.Handler.channelDeliversFiles` answers it; `cmd/server/router.go` declares it from the **same** `if store != nil` branch that passes `wecom.WithAttachments(store)`, so a deployment that has the hop and a deployment whose agents are promised the hop cannot be different deployments. Two lines, one `if`.

`execenv.ChannelCarriesFiles` now takes `(channelType string, serverSaysDelivers bool)` and consults only the second for a channel-backed chat. The signature is the point: there is no longer a channel type to answer from. Default false covers all three ways to get here honestly — no storage, no declaration, and an older server that never sends the field.

New coverage: no-store and mixed-version rows in the brief fixtures and the per-turn prompt matrix; `TestChatChannelDeliversFilesDefaultsOffAcrossVersions` decodes an old server's claim JSON rather than building a struct, since a struct literal can't express "the server didn't send this"; and `TestClaim_FileDeliveryIsTheDeploymentsAnswerNotTheChannelTypes` drives the real claim endpoint with the same WeCom binding under a declared and an undeclared deployment.

## 2. Limits and concurrency — what the sources actually say

Both read at the commits you pinned. They agree with you; two details are worth writing down because they change what the right number is.

**`aibot-node-sdk@80615b98`, `src/client.ts:397-406`** — the ladder has three steps, not one threshold:

```ts
const MAX_CONCURRENCY = totalChunks <= 4 ? totalChunks : totalChunks <= 10 ? 3 : 2;
```

with the reason stated beside it: `>10 分片（>5MB）：并发 2（企微后端对大量并发 chunk 会返回 system error）`. So the drop past ten chunks is exactly as you said, and the same code also says a file of four chunks or fewer should go *entirely* at once. Adopted whole — dropping only to 2 would have made small files slower for a caution that does not apply to them.

The same function is also where our 50 MB came from: `if (totalChunks > 100) throw ... "max ~50MB"`. That is the transport's ceiling, and it is a genuinely different number from the product's.

**`wecom-openclaw-plugin@1a91ef73`, `src/const.ts:111-124`** — the 20 MiB is confirmed, and it is the *file* ceiling:

```ts
export const IMAGE_MAX_BYTES = 10 * 1024 * 1024;
export const VIDEO_MAX_BYTES = 10 * 1024 * 1024;
export const VOICE_MAX_BYTES = 2 * 1024 * 1024;
export const FILE_MAX_BYTES = 20 * 1024 * 1024;
export const ABSOLUTE_MAX_BYTES = FILE_MAX_BYTES;
```

Nothing contradicts the review. The addition is that the per-kind ceilings are lower — and we already apply those in `wecomMediaKind`, which demotes an oversize image to a file rather than offering the server something it will refuse. Since a file is the widest of the four, 20 MiB is correctly the single hard cap for everything.

So: `maxMediaUploadBytes = 20 << 20`, checked against the attachment row's `size_bytes` before a byte is fetched (`readObject` still re-checks what it actually read, since the column is metadata and the object is the truth), and `maxMediaTransportBytes` keeps the 50 MB framing figure with a test asserting the product cap stays inside it.

Tests: the boundary at 20 MiB and 20 MiB + 1; the ladder as a table (0→1, 1→1, 4→4, 5→3, 10→3, 11→2, 40→2); and a wiring test that offers eleven chunks with every verdict withheld and fails if a third goes out. That last one needed the harness to withhold the ack *off* the write path — `wsSender.write` serializes every frame under one mutex, so blocking inside `WriteMessage` measures the write lock and reports 1 whatever the limit is.

## 3. Three states, and the local failures now reach the user

`deliveryState` is `delivered` / `definitely_failed` / `unknown`, and those exact strings go on the log line.

- **delivered** — nothing is said. The file is what the user sees.
- **definitely_failed** — the upload was refused, the object could not be read, or the send came back with an errcode. Keeps the existing definite wording.
- **unknown** — the send frame went out and no verdict came back. Non-definitive wording, and **never resent**; the message also says why, since "will you try again" is the obvious next question and the answer is that a duplicate file cannot be taken back.

An upload failure is definite even when it is the finish step's own lost ack: no `media_id` means nothing was ever addressed to the chat.

One thing I could not make precise, so I made it safe and said so in the comment: `wsSender.request` returns the same `ctx.Err()` whether the context ended before the frame was written or while waiting for its verdict, so from outside those two cannot be told apart. Both read as `unknown` — under-claiming a send that never happened costs less than denying one that did.

For the local failures, the attachment lookup now runs **before** the pending-slot claim. That ordering is what makes the apology possible: rationing first can only either stay silent about a dropped file or warn about a file that never existed. Now a shed delivery and an expired slot wait both know a file was waiting and say so, a lookup error says the narrower true thing (we could not find out whether there was one), and a turn with nothing attached still says nothing even under a full backlog. There is a test for that last case — a warning that fires on turns with no files is a warning people learn to ignore.

The one path that still only logs is a missing live connection, since the socket that would carry the apology is the socket that isn't there.

## The two you asked to keep

Both kept. The origin gate is at `outbound.go:160-178` (`chatDoneTaskID` → `GetAgentTask` → `TaskInputIsChannelIngested` → `if !deliver`), still ahead of every write to WeCom: the answer's text at line 210 and `deliverAttachments` at line 216, whose upload init frame is itself a write. `TestProcessEvent_AWebUIAnswerDeliversNeitherWordsNorFile` is unchanged and still asserts zero `aibot_upload_media_init` frames on a refused turn.

## Verification

Every fix was checked in reverse — broken deliberately, confirmed red, restored:

- restoring `case ChannelTypeWecom: return true` → `TestChannelCarriesFiles/wecom,_on_a_deployment_with_no_object_storage`, the mixed-version row, `TestBriefSurfaceDeliveryPolicy` (both no-store surfaces), `TestBuildChatPromptTwoLayerChannelPolicy/wecom_on_a_deployment_that_cannot_deliver`, and `TestChatChannelDeliversFilesDefaultsOffAcrossVersions` all fail;
- putting the cap back to 50 MB → the 20 MiB boundary and the oversize-attachment test fail;
- replacing the ladder with a fixed 4 → the table fails at 0, 1, 5, 10, 11 and 40, and the wiring test reports "a third chunk went out with 2 already unanswered";
- folding `unknown` into `definitely_failed` → the unconfirmed-send test and the mixed test fail; folding `definitely_failed` into `unknown` → the refusal and transport rows fail; folding `delivered` into either → the delivered test and `TestProcessEvent_SendsTheAnswerAndThenTheFile` fail;
- reverting the two new notifications to log-only → the shed and lookup-failure tests fail.

`go build ./...` and `go vet ./...` clean from `server/`; `-race` green against a migrated Postgres for `internal/integrations/wecom`, `internal/daemon/execenv`, `internal/handler`, and the daemon prompt tests; `git diff --check` clean; CI green.

multica-eve (dismissed):
Thanks — the latest commit fixes the original capability negotiation and upload-limit/concurrency blockers, and it substantially improves the three-state delivery reporting. I re-reviewed `24d9a3f`; two merge blockers remain.

1. **The pending cap no longer bounds the goroutines it was introduced to bound.** `server/internal/integrations/wecom/outbound_media.go:180-184` now spawns unconditionally, and the goroutine performs `ListAttachmentsByChatMessage` before it calls `claimAttachmentSlot` at `221-229`. Under a slow database or a burst of `chat:done` events, arbitrarily many goroutines and attachment queries can accumulate before the counter is consulted. That contradicts the invariant documented in `outbound.go:72-75` and `outbound_media.go:530-535`. Please bound the lookup stage as well while preserving the useful behavior that a shed warning is emitted only after we know a file exists. Add a regression test that blocks the lookup, submits more than the cap, and asserts the real number of active/spawned workers stays bounded; directly filling the counter does not exercise this path.

2. **A WebSocket write error is not proof of definite non-delivery.** `ws_sender.go:190-221` returns the raw `WriteMessage` transport error, but `outbound_media.go:369-379` classifies every non-ack/non-context error as `definitely_failed`; `outbound_media_test.go:458` pins `broken pipe` to the assumption that the frame did not reach the socket. Once a write has been attempted and no WeCom verdict is available, delivery is uncertain — the peer may have received the frame before the local transport surfaced the failure. Please preserve the failure stage: an explicit WeCom errcode or a provable pre-write failure may be definite; an attempted write with no server verdict should be `unknown` and must not be retried. A fake connection that records the frame and then returns a write error would pin this boundary.

The targeted daemon prompt tests, prompt byte-identity test, WeCom tests, WeCom race test, `git diff --check`, and all nine CI checks pass on this head. The blockers are in concurrency/failure semantics that those tests do not currently exercise.

@leroy-chen, you and @seacen are the named WeCom code owners. Could you also validate this branch against a real WeCom tenant/bot? In particular, please confirm that a generated file/image arrives after the text and opens correctly in both a direct chat and a group, and exercise an upload over 10 chunks (>5 MiB) so the large-file concurrency path is covered. A near-20 MiB file would also confirm the product ceiling if practical. A short result with sensitive tenant/file details omitted is enough.

seacen:
Both blockers are fixed at `92fc4cd0c`.

## 1 — the pending cap now bounds what it was introduced to bound

You were right that moving the claim after the lookup lost the gate. The reason
it kept moving is that one counter was being asked to be two things whose
correct positions sit on opposite sides of the query: as a gate it has to be
claimed before the spawn, where nothing is yet known about the turn; as the
basis for telling the user, it has to be claimed after the lookup, or it warns
about files that never existed. Round 1 fixed the second and broke the first.

So there are two counters now.

`admitAttachmentDelivery` is claimed in `deliverAttachments` before `o.spawn`,
and released in the goroutine's defer. It bounds the goroutines and, with them,
the `ListAttachmentsByChatMessage` calls they run. Its cap is
`2 * maxPendingAttachmentDeliveries`, deliberately: reaching it means the pending
cap is already full and as many turns again are still in their lookup, so a
backlog meets the honest post-lookup shed first and only reaches this one when
the system is well past that. A refusal here is logged and not spoken, which is
what your wording allows — "a shed warning is emitted only after we know a file
exists" forbids the false alarm, and at admission we cannot yet tell a turn with
a file from one without.

`claimAttachmentSlot` keeps its position and its behaviour: claimed after the
lookup, and it still tells the user.

The invariant comments in `outbound.go` and `outbound_media.go` were rewritten
rather than left describing the old single-counter arrangement.

## 2 — an attempted write is no longer read as definite non-delivery

`wsSender.write` already distinguished the stages — it tracks
`traceStageDeadline` vs `traceStageWrite` for the trace line. That distinction
just never left the function. It now does: a failure raised once `WriteMessage`
has been entered comes back wrapped in `errWriteAttempted`, and `sendOutcome`
reads that as `deliveryUnknown`. A marshal failure or a deadline the connection
refused stays definite, because nothing left the process.

`outbound_media_test.go`'s old `{"transport", broken pipe, definitely_failed}`
row is gone — it was pinning exactly the assumption you flagged.

## Tests

Both drive the real path rather than the counters, per your note that filling
the counter does not exercise the gate:

- `TestDeliverAttachments_AdmissionBoundsTheLookupStage` holds every lookup open
  through a gate on the fake queries, submits `cap + 16` deliveries through
  `deliverAttachments` with a spawn that starts real goroutines, and asserts on
  the number that actually started and the number of queries that actually
  reached the database — not on the counter's value.
- `TestSendMedia_AnAttemptedWriteIsNotProofOfNonDelivery` uses a connection that
  records the frame and then returns `broken pipe`, asserts the frame was
  recorded (so the rig proves something), and asserts the outcome is `unknown`.

Both reverse-verified: with the production change backed out, all three of these
fail; restored, they pass. Full package green under `-race`.

`fakeObjectStore` took a mutex — its comment said it needed none because every
test here runs the delivery inline, and the admission rig is the first that does
not.

## Separately — the size limits are now cited to WeCom, not to a third party

While checking the 20 MiB figure I found we had sourced it to
`WecomTeam/wecom-openclaw-plugin`. That file carries the right numbers but cites
nothing itself, so it corroborates rather than establishes them. WeCom documents
all four on `aibot_upload_media_init`'s `body.total_size` row
(developer.work.weixin.qq.com/document/path/101463 — image 10MB, voice 2MB,
video 10MB, plain file 20MB) and repeats them on the classic media upload API
(document/path/90253). The comment points at those now.

Worth knowing for the ceiling question below: the aibot errcode range documents
`846613 分片数据超过大小限制` for an oversize *chunk*, and nothing for an
oversize file. What the server does with a `total_size` past 20MB is therefore
not documented.

## On the real-tenant validation

We have a live tenant and are working through it. Inbound is done — I posted the
results and a screen capture on #6605. The outbound cases you named (a file in a
direct chat and a group, an upload over ten chunks, and a near-20 MiB file) are
not covered yet: our deployment is still on a build that predates this branch, so
testing it now would exercise the old path. We're redeploying with the current
heads of all the open branches and will run those four together, including one
past 20 MiB to see what the server actually says.

seacen:
@multica-eve @Bohan-J — real-tenant results for everything you asked, plus one
thing the agent noticed that neither of us did.

Run against a live WeCom tenant on the current head, with
`MULTICA_WECOM_TRACE=1`. Video and a group screenshot at the bottom.

## Your three, with the numbers

**A generated file arrives after the text and opens, in a direct chat and in a
group.** Both confirmed. The group case is the screenshot: the bot answers in
text, then `test-doc.md` (2K) arrives as its own message behind it.

**An upload over ten chunks.** 8,388,424 bytes, **16 chunks** — over the ten
where the concurrency ladder steps down to two.

**A near-20 MiB file.** 19,922,778 bytes, **38 chunks**, delivered and opened.

```
upload 1    1 chunk    0.5 MiB    delivered
upload 2   16 chunks   8.0 MiB    delivered
upload 3   38 chunks  19.0 MiB    delivered
upload 4   —          25.0 MiB    refused before the read
```

The refusal, verbatim from the log:

```
attachment is 26214223 bytes: wecom: media exceeds the 20971520 byte upload limit
delivery=definitely_failed
```

`definitely_failed` is right there: nothing left the process, so there is no
version of this where the user might have it.

## The product ceiling, measured rather than taken from the doc

You asked whether a near-20 MiB file would "confirm the product ceiling". It
does, and we went one further — our gate refuses before asking the server, so
production had never seen what the server itself says. A probe sent four
`aibot_upload_media_init` frames with no chunks behind them:

```
20,971,520  (20 MiB)      accepted, upload_id returned
20,971,521  (one byte over) REFUSED  errcode 40006  "invalid file size"
26,214,400  (25 MiB)      REFUSED  errcode 40006
62,914,560  (60 MiB)      errcode 40058  "body.total_size 62914560 too large,
                                          required no more than 29000000"
```

Three things worth having:

- **The ceiling is exactly 20 MiB — 20,971,520 bytes, binary.** One byte over is
  refused. `maxMediaUploadBytes = 20 << 20` is byte-exact, not approximately
  right.
- **There IS an errcode for an oversize file: 40006.** It is not in the aibot
  846xxx block, which is where anyone would look — the aibot table documents
  846613 for a single oversize *chunk* and nothing for a whole file. Worth a
  line in the code, since the next person to search will search the wrong table.
- **40058 is a second, earlier guard with a different number**: 29,000,000. It
  fires before the size check and is in no document we can find.

## Composition order, for #6605's question

Bohan asked on #6605 whether a 图文混排 message with text before AND after an
image keeps its order. It does — three runs, in order:

```
text=" 再试试文+图+文 [Image] 最后的文字你收到了吗"
```

The agent's own reading back: "图片占位符正好夹在两段文字中间,位置和你发的一致."
That PR has merged, so this is confirmation rather than a gate.

## What the agent noticed, and it is a real gap

When the 25 MiB delivery was refused, the user was told. **The agent was not**,
and it worked that out for itself:

> 如果 25MB 这发失败,我这边收到的 upload 返回是成功的——也就是说投递失败不会回传给
> agent。我会以为已经发给你了。

It is right, and it is structural rather than a bug in this PR. `multica
attachment upload` returns when the file is in object storage; delivery happens
later, on `EventChatDone`, after the run has ended. Every failure path in
`outbound_media.go` reports through `tellUser` and the log — there is no route
back to a run that is already over.

So an agent can finish a turn believing it delivered a file that was refused.
The user is not left in the dark, which is what your three-state reporting was
for; the agent is. Not something this PR should grow, but it belongs in the
package doc, and it may want its own issue.

## Evidence

- Video, four sizes end to end: https://github.com/seacen/multica/releases/download/outbound-files-2026-08-11/outbound-files.mp4
- Group chat, agent-produced file: https://github.com/seacen/multica/releases/download/outbound-files-2026-08-11/group-file.png

Both blockers from your last round are fixed at `ae661c8db` — the admission gate
bounds the lookup as well as the send, and an attempted write is classified
`unknown` rather than definite. Nine checks green.

Bohan-J (changes requested):
Thank you for the live-tenant run — that closes the verification item I left open. Four sizes end to end, 16 and 38 chunks through the stepped-down concurrency ladder, and a probe that *measures* the ceiling rather than quoting it is more than I asked for. Three of those findings deserve to live in the code, not just in this thread: the ceiling is exactly 20,971,520 bytes; a whole oversize file comes back as errcode `40006`, which is nowhere near the aibot `846xxx` block anyone would search; and `40058` is a second, undocumented guard at 29,000,000.

The code at `ae661c8db` has not moved since my last review, so the one blocker from that round still stands.

## Blocker — this turn's deployment verdict is still inside the cached prefix

`server/internal/daemon/execenv/execenv.go:140-148` defines `ChatChannelDeliversFiles` as the server's verdict **for THIS turn**, arriving on each claim. `server/internal/daemon/execenv/runtime_config_sections.go:720` branches the runtime brief on it.

The brief is the cache prefix (MUL-5377). So a single resumed WeCom session renders two different briefs when the deployment changes underneath it — a server upgrade that starts sending the field, or object storage being enabled or disabled. Measured on this head:

```go
base := TaskContextForEnv{ChatSessionID: "c-1", ChatChannelType: ChannelTypeWecom, AgentName: "Eve", AgentID: "eve-1"}
buildMetaSkillContent("claude", base)  // 6467 bytes
base.ChatChannelDeliversFiles = true
buildMetaSkillContent("claude", base)  // 6489 bytes
```

`TestBriefByteIdenticalAcrossRunsForEveryKind` (`runtime_config_test.go:1843`) passes on this head and will keep passing: its chat baseline is Slack, and none of its variants flips this field. The exposure is not hypothetical either — `runtime_config_delivery_test.go:37,42` already asserts that the *same* WeCom surface renders two different briefs on the two values, and those are exactly the two a resumed session can see in sequence.

The per-turn prompt is the right home, and it already has the correct branch at `server/internal/daemon/prompt.go:560`. Two things needed:

- make the brief's WeCom delivery copy independent of the field — wording that holds on either verdict and does not contradict the per-turn instruction;
- add a WeCom variant to the byte-identical test that flips `ChatChannelDeliversFiles`, so the invariant test would actually fail if this returns.

## Nit, not blocking

`server/internal/integrations/wecom/outbound.go:86-88` says reaching the admitted cap "means the pending cap is full AND as many more turns are still in their lookup." Your own slow-lookup test can fill admission while pending is 0. The two-counter split is right; the comment overstates what the larger cap implies.

## Your last section — agreed, and it should not be this PR

An agent seeing a successful `attachment upload` and never learning the delivery was refused is real and structural: the upload returns when the object lands in storage, and delivery runs on `EventChatDone`, after the run has ended. A line in the package doc here, and its own issue for the route back. Nothing in this PR should grow to cover it.

Everything else verified green on this head: `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1`, `go test ./internal/integrations/wecom -count=1` with and without `-race`, `go vet`, `git diff --check`, and 9/9 CI checks.

seacen:
Pushed `45384ca5b`. All four points are addressed; the second one turned up a duplicate of the same overstatement that wasn't flagged.

## The verdict is out of the cached brief

`runtime_config_sections.go` — `writeOutput`'s chat branch no longer calls `ChannelCarriesFiles(...)`. It is a two-way split on whether the chat is channel-backed at all, and channel-backed chats get one deployment-neutral line that defers to the per-turn message.

I did **not** scope this to WeCom. A `ChatChannelType == ChannelTypeWecom` branch in the brief reintroduces exactly the channel-type reasoning `ChannelCarriesFiles`'s own doc comment exists to forbid, and hardcoding "Slack is text-only" into a cached prefix keeps denying a hop that a server-side change can turn on. Slack and Feishu still get the hard verdict every turn from `prompt.go:563`, unchanged.

`prompt.go:560` is untouched. Its comment no longer claims the brief states the same policy — it now records that this is the only place the verdict is stated, so a line has to be emitted every turn.

## The byte-identical test now covers this

`TestBriefByteIdenticalAcrossRunsForEveryKind` gains two things: a `chat-wecom` kind, and a `channel-delivers-files` variant that flips `ChatChannelDeliversFiles` between runs of one session. The variant is the one that matters — the verdict arrives on every claim and is a deployment fact, so an upgrade that starts sending the field, or object storage being switched on, flips it under a session already running.

Reverse-verified: with `runtime_config_sections.go` reverted and the new test kept, it fails at byte 5696 with the two renderings side by side.

## The admitted cap, in both places

`outbound.go:86` says what you asked for: the admitted cap being reached does not imply pending is full, because admission is held for the goroutine's whole life, including turns whose lookup finds nothing.

The same overstatement was also on `maxAdmittedAttachmentDeliveries` in `outbound_media.go:525` — "reached only when the pending cap is already full and as many turns again are still in their lookup". `TestDeliverAttachments_AdmissionBoundsTheLookupStage` reaches the admitted cap with pending at zero: it parks every goroutine in `q.lookupGate` before any can claim a pending slot. Both now cite it.

## Delivery cannot answer back

Added to the package doc (`types.go:36`): `multica attachment upload` returns once the object is in storage and bound to the reply; the send runs on `EventChatDone`, after the run has ended. So a shed, refused or socket-lost delivery is reported to the person in the chat and never to the agent.

Verified rather than assumed — `TaskService.CompleteTask` publishes chat:done after commit (`service/task.go:3622` → `broadcastChatDone` at :5148), and `Outbound.deliverAttachments` runs off that event. No fix here; it bounds what the agent can be told.

## Sweep

Grepped `ChatChannelDeliversFiles`, `ChannelCarriesFiles`, `ChannelDisplayName`, `Delivering files here`, `conversation is text-only` across Go/md/tsx and corrected every comment this diff would have made false: `execenv.go:140`, `channel_type.go:66,86`, `prompt_test.go:583`. Checked and deliberately left: `daemon/types.go:94`, `handler/agent.go:363`, `handler/handler.go:559-596` — they describe the wire field and the server-side declaration and stay true via the per-turn prompt.

CI green.

@Bohan-J @multica-eve — one thing I want to check rather than guess: your note said "the code at `ae661c8db` has not moved since my last review". I can't line that up. Which review do you mean? If there is an earlier round whose points I missed, point me at it and I'll take them in this pass rather than leaving them sitting.

Bohan-J (approved):
Approving — `45384ca` resolves the blocker, and thank you for the whole run of this review. Five rounds, and each one came back with the defect actually removed rather than papered over, plus a live-tenant investigation that produced numbers nobody had: the exact 20,971,520-byte ceiling, `40006` for a whole oversize file, and the undocumented 29,000,000 guard behind `40058`. The gap you found yourself at the end — that a refused delivery can never reach the agent that produced the file — is the kind of thing a review does not usually get handed.

I re-verified the fix rather than taking the diff at its word:

- The brief is now byte-identical across the verdict flip for one WeCom session — 6447 bytes both ways, where it was 6467 / 6489 at `ae661c8db`.
- `ChatChannelDeliversFiles` is read in exactly one non-test place now, `daemon/prompt.go:566`. The brief points at the per-turn message instead of answering, and `ChannelCarriesFiles`' doc says why the brief must not call it.
- The new `chat-wecom` kind and `channel-delivers-files` variant are a real guard, not a decorative row: I reintroduced the branch in `writeOutput` and `TestBriefByteIdenticalAcrossRunsForEveryKind` failed on it, then reverted. The coverage that moved out of the brief landed in `TestBuildChatPromptTwoLayerChannelPolicy`, so both delivery answers are still pinned somewhere.
- The two doc corrections are right, and the second is more than a wording fix: `maxAdmittedAttachmentDeliveries` reads as headroom now rather than as a derived quantity, which is what the slow-lookup test actually demonstrates.

Green on this head: `./internal/daemon/execenv` and `./internal/daemon` targeted, `./internal/integrations/wecom` with and without `-race`, `go vet`, `git diff --check`, 9/9 CI.

Merging now. The delivery-cannot-answer-back limitation is recorded in `types.go` and gets its own issue — it should not grow this PR.

PR #6605, opened 2026-08-07
MUL-5906 feat(wecom): read the photos, files and videos people send the bot

seacen:
## The defect

`dispatchFrame` routes only `msgtype == "text"`:

```go
// wecom_channel.go, before
if mc.MsgType != "text" {
    sender.sendText(..., "抱歉，我目前只能处理文字消息。")
    return nil
}
```

A screenshot, a PDF, a video, or a 图文混排 (a message composed of text runs
and attachments interleaved — the ordinary way people write on WeCom) is
answered with that one line and dropped at the socket. No `chat_message`, no
session, no agent run, nothing in the web UI either. Pasting a screenshot
under a question is how a large share of WeCom questions get asked, so the bot
reads as broken for a large share of what it is sent.

## What the user sees now

Send a photo with "这个报错怎么回事" and the agent gets
`这个报错怎么回事\n[Image]` plus the photo bound as an attachment. A 图文混排
arrives with its runs in composition order, one placeholder per attachment,
and the attachments in the same order.

## Sibling parity

The placeholders are Lark's and DingTalk's, byte for byte — an agent reads
every channel through one prompt, so a wecom-only spelling is a wecom-only
thing for it to learn:

| | source |
|---|---|
| `[Image]` / `[File]` / `[Audio]` / `[Video]` | `lark/content_flatten.go` `flattenContent` |
| `[Image]` | `dingtalk/inbound.go:95` `dingtalkImagePlaceholder` |
| `[File]` | `dingtalk/inbound.go:205` |

The resolver is the shape `lark/media_ingest.go` established and the Router
depends on: `HasMedia` is a pure in-memory look at the payload already in
hand (it runs on the connector ACK path and decides whether the message pays
for a media deadline at all); `ResolveMedia` runs detached; every upload is
covered by an intent-ledger row written **before** the PUT; nothing is ever
deleted inline. The object key is derived per chat message and per attachment
index, for the reason lark documents at its own `mediaObjectKey` — a platform
message can be ingested twice once its dedup claim goes stale, and a shared
key runs the second ingest into the first one's ledger row.

Wiring follows DingTalk exactly (`router.go`, dingtalk block above): build the
resolver only when an object store exists, pass it through the ResolverSet, nil
degrades to placeholder text.

`Capabilities()` gains `CapAttachment`, matching `dingtalk_channel.go:52`.
Outbound media is **not** claimed — that needs `aibot_upload_media_*` and is
not in this PR.

## The head-on conflict: `"mixed"` → `MsgTypeText`

`channelMsgType` currently maps `"mixed"` to `MsgTypeUnknown` and argues it:

> Kept as Unknown rather than mapping "mixed" → Text, which was dead and
> implied mixed messages were routed as text when they are not.

**That comment was right about the code it described, and this PR is what makes
its objection stop applying.** The objection was "the mapping claims a routing
that does not happen". This patch adds the routing. That is why the mapping and
the ingestion are in one commit: mapping to Text *without* the routing is
exactly the dead, misleading mapping the old comment warned against, and
splitting them would land that regression for one PR's width.

Lark treats the identical shape the identical way: `feishu_channel.go:167` maps
`post` → `MsgTypeText` while `media_ingest.go:272` pulls that same post's
`img`/`media` spans out separately. `mixed` is `post` under another name.

The comment at the mapping now says all of this, so the next reader gets the
argument rather than an unexplained flip.

## SSRF: the fetch is guarded

Inbound media is an HTTP GET to an address that arrived over the socket, made
by a process sitting inside the deployment's network. The guard is on the
**connection**, not the URL — a hostname check is worth nothing against a `302`
to `http://169.254.169.254/`, and a name that passed a check can resolve
elsewhere a moment later. The client resolves the destination itself, refuses
any non-public address, and connects to the literal that passed, on every hop.

The shape is @leroy-chen's, from the other WeCom media PR (#6524,
`newSecureMediaClient` / `isAllowedMediaIP`) — that PR raised this first and its
fix is the right one. Two things added: the reserved-range list (netip's own
predicates miss RFC 6598 CGNAT, where a Tailscale tailnet lives, plus
IETF-reserved and NAT64 space) and an injectable resolver so DNS rebinding is
tested rather than argued about.

Transport errors are also stripped of `net/http`'s `*url.Error` wrapper before
they reach a log: that wrapper prints the URL it was fetching, and the URL here
is a five-minute bearer credential for a colleague's private attachment.

## Memory

Buffering ciphertext and plaintext at once is 2 × 100 MiB per attachment, and
the Router's `MediaConcurrency` defaults to 8 — 1.6 GB of live heap on a
self-hosted box, killing a process that is also serving Lark, Slack and
DingTalk. So the default path streams: ciphertext decrypts block-by-block into
an unlinked 0600 temp file, and the upload reads from there with the exact
length it now knows (`S3Storage.UploadStream` needs one, and the plaintext
length is only known after the final block). A backend without `UploadStream`
falls back to the buffered path.

## Overlap with #6599

#6599 (voice transcripts) is open, not merged, so this PR does not depend on its
`bodyText()` helper and writes its own `ownText()`. The boundary is clean:

- **Standalone `msgtype == "voice"` is #6599's and is untouched here** — it
  still takes the receipt path (pinned by `TestVoiceStaysOnTheReceiptPath`).
- A **voice run inside a 图文混排** is this PR's, because that container is this
  PR's; dropping the run would lose a spoken sentence out of the middle of a
  message whose other runs are read.

Whichever lands second will need to reconcile one function name, nothing more.

## Also here, and why

- The unsupported-kind receipt no longer says "我目前只能处理文字消息" (text
  only). It is now false, and told to someone who has just watched the bot
  answer a screenshot it reads as the bot being broken.
- A failed attachment sends one short line to the chat. Without it the failure
  is invisible in the worst way: the stored body still says `[Image]`, so the
  agent answers as if it had been shown a picture it never received. Two
  failures of one kind are one notice. It goes over the same live aibot socket
  (`sendersRegistry` → `wsSender.sendText`); no live socket means log only.
- `types.go`'s package docblock said inbound is text-only. Updated. **The
  maintainer/code-owner block and the single-replica paragraph are untouched.**

## Deliberately not here

Quoted-message rendering. A quote block goes in front of the body, which moves
where slash commands are parsed from — that belongs with the command work, not
with media. Quoted attachments are consequently not queued for download, which
is also the right answer on its own: they belong to a message somebody else
sent, and their url is running down someone else's five-minute clock.

## Tests

`go build ./...` and `go vet ./...` clean. Full `./internal/...` suite green
against Postgres, plus `./cmd/...`, plus `-race` on the wecom package. One
pre-existing failure, `TestProbeAgentCLIs_QoderResolvesViaLoginShell` in
`internal/daemon` (a package this PR does not touch) — it fails identically on
a clean `upstream/main` checkout because it does not isolate `PATH` and a real
`qoderclicn` on the machine wins.

**Every test below was reverse-verified: the fix was reverted, the test was
confirmed to fail, and the fix restored.** Actual output:

<details>
<summary><b>1. Routing — revert <code>dispatchFrame</code> to <code>if mc.MsgType != "text"</code></b></summary>

```
--- FAIL: TestDispatchFrame_MediaCallbacksReachTheHandler
    --- FAIL: .../image      a image message never reached the handler — it was answered and dropped
    --- FAIL: .../file       a file message never reached the handler — it was answered and dropped
    --- FAIL: .../video      a video message never reached the handler — it was answered and dropped
    --- FAIL: .../mixed_keeps_composition_order
                             a mixed message never reached the handler — it was answered and dropped
--- FAIL: TestOwnText_MixedDropsRunsItCannotRead
    a mixed message with one readable run must still be routed
--- FAIL: TestVoiceStaysOnTheReceiptPath
    a mixed message with a voice run must be routed
```
</details>

<details>
<summary><b>2. <code>"mixed"</code> mapping — revert to <code>MsgTypeUnknown</code></b></summary>

```
--- FAIL: TestChannelMsgType_Normalization
    ws_frame_test.go:74: channelMsgType("mixed") = unknown, want text
--- FAIL: TestDispatchFrame_MediaCallbacksReachTheHandler/mixed_keeps_composition_order
    inbound_media_test.go:128: normalized type = unknown, want text
```
</details>

<details>
<summary><b>3. Placeholder parity — swap in a wecom-only spelling</b></summary>

```
--- FAIL: TestMediaPlaceholders_MatchLarkAndDingTalk
    mediaPlaceholder(image) = "[图片]", want "[Image]" — lark and dingtalk emit "[Image]"
    mediaPlaceholder(file)  = "[文件]", want "[File]"  — lark and dingtalk emit "[File]"
    mediaPlaceholder(video) = "[视频]", want "[Video]" — lark and dingtalk emit "[Video]"
--- FAIL: TestDispatchFrame_MediaCallbacksReachTheHandler/mixed_keeps_composition_order
    stored body = "看下这个报错\n[图片]\n还有这个\n[图片]", want "看下这个报错\n[Image]\n还有这个\n[Image]"
```
</details>

<details>
<summary><b>4. SSRF guard — remove <code>transport.DialContext = g.dial</code></b></summary>

```
--- FAIL: TestTheGuardRefusesAnInternalAddressOutright (41.89s)
    http://127.0.0.1:9/ failed with ... dial tcp 127.0.0.1:9: connect: connection refused, want the guard's refusal (a connection error means it tried)
    http://169.254.169.254/latest/meta-data/ failed with ... context deadline exceeded, want the guard's refusal
    http://10.0.0.1/ failed with ... EOF, want the guard's refusal
    http://[::1]:9/ failed with ... connection refused, want the guard's refusal
    http://100.100.100.100/ was fetched
    http://198.18.0.18/ failed with ... EOF, want the guard's refusal
--- FAIL: TestAPublicHostnameResolvingInwardIsRefused
    err = wecom: media download: EOF, want the guard's refusal
--- FAIL: TestARedirectToAnInternalAddressIsRefused (30.00s)
    err = wecom: media download: context deadline exceeded, want the guard's refusal on the second hop
--- FAIL: TestAMixedAnswerDialsOnlyThePermittedAddress
    the public answer was not used: wecom: media download: EOF
--- FAIL: TestTheMediaClientIgnoresProxyEnvironment
    the installed dialer answered dial tcp 169.254.169.254:80: i/o timeout for the metadata address, want the guard's refusal
```

`http://100.100.100.100/ was fetched` is the line worth reading twice: with the
guard removed, the test machine genuinely reached an internal address.
</details>

<details>
<summary><b>5. Signed-URL leak — make <code>stripURL</code> a no-op</b></summary>

```
--- FAIL: TestATransportFailureNeverCarriesTheSignedURL
    --- FAIL: .../buffered
        the error carries the signature: wecom: media download: Get "http://127.0.0.1:51733/object?sig=AKIDSUPERSECRETSIGNATURE&t=1": EOF
    --- FAIL: .../streaming
        the error carries the signature: wecom: media download: Get "http://127.0.0.1:51733/object?sig=AKIDSUPERSECRETSIGNATURE&t=1": EOF
```
</details>

<details>
<summary><b>6. 32-byte pad — set <code>mediaPadBlock = 16</code> (the AES block size, the natural mistake)</b></summary>

The decryptor is checked against a from-scratch encryptor in the test file, not
against itself.

```
--- FAIL: TestDecryptMediaRoundTripsRealCiphertext
    --- FAIL: .../one_byte                 decryptMedia: wecom: media padding length 31 is out of range (1..16)
    --- FAIL: .../exactly_15_bytes         decryptMedia: wecom: media padding length 17 is out of range (1..16)
    --- FAIL: .../exactly_32_bytes_—_a_full_32-byte_pad_block_follows
                                           decryptMedia: wecom: media padding length 32 is out of range (1..16)
    --- FAIL: .../33_bytes                 decryptMedia: wecom: media padding length 31 is out of range (1..16)
    --- FAIL: .../64_bytes                 decryptMedia: wecom: media padding length 32 is out of range (1..16)
    --- FAIL: .../utf-8_text               decryptMedia: wecom: media padding length 25 is out of range (1..16)
--- FAIL: TestResolveMediaStoresTheDecryptedFile/buffered   MediaRefs = 0, want 1
--- FAIL: TestResolveMediaStoresTheDecryptedFile/streaming  MediaRefs = 0, want 1
```
</details>

<details>
<summary><b>7. Streaming decrypt — hold back one AES block instead of the pad block</b></summary>

```
--- FAIL: TestStreamingDecryptMatchesTheBufferedOneAtEveryPadLength
    streaming decrypt of 0 bytes: wecom: media padding length 32 is out of range (1..32)
--- FAIL: TestIngestPrefersTheStreamingPathAndUploadsTheRightBytes
    UploadStream was called 0 times, want the streaming path taken
```
</details>

<details>
<summary><b>8. Intent-before-upload — move <code>RecordPendingMediaObject</code> after the PUT</b></summary>

```
--- FAIL: TestResolveMediaRecordsIntentBeforeUploading
    the intent row was written after the upload; a crash in between would leave an uncovered object
--- FAIL: TestResolveMediaSkipsAKeyTheReconcilerOwns
    uploaded into a key the reconciler owns
```
</details>

<details>
<summary><b>9. Per-attachment object key — drop the index from the key</b></summary>

```
--- FAIL: TestResolveMediaNamesAnUnnamedFile
    both attachments share a storage key; the second would overwrite the first
```
</details>

<details>
<summary><b>10. Failure notice — stop calling <code>tellTheSender</code></b></summary>

```
--- FAIL: TestResolveMediaKeepsGoingWhenOneAttachmentFails
    the sender was told 0 times, want exactly one notice
--- FAIL: TestResolveMediaTellsTheSenderWhatWentWrong
    --- FAIL: .../one_notice_names_the_size_limit                    notices = 0, want 1
    --- FAIL: .../two_failures_of_one_kind_are_one_notice            notices = 0, want 1
    --- FAIL: .../a_group_notice_goes_to_the_room,_not_the_sender    notices = 0, want 1
```
</details>

<details>
<summary><b>11. Receipt copy — restore the text-only wording</b></summary>

```
--- FAIL: TestUnsupportedReceipt_DoesNotClaimTextOnly
    receipt "抱歉，我目前只能处理文字消息。" still claims text-only while image/file/video/mixed route
```
</details>

<details>
<summary><b>12. ResolverSet wiring — drop <code>Media: media</code></b></summary>

```
--- FAIL: TestNewResolverSet_WiresAllResolvers
    a media resolver was passed and dropped
```
</details>

### One test was deleted for failing this bar

A first draft asserted that passing `nil` media yields a nil interface (the
typed-nil hazard). It **passed with the guard reverted** — `media` is already an
interface parameter, so the guard was a no-op and the assertion measured
nothing. Both the assertion and the no-op guard were removed, and the assertion
was replaced with one that drives the wired resolver
(`withMedia.Media.HasMedia(...)`), which does fail when the wiring is dropped
(case 12 above).

The `TestTheMediaClientIgnoresProxyEnvironment` assertion `tr.DialContext != nil`
was likewise weak — `Transport.Clone()` carries the default dialer over, so it
passes on a transport with no guard. It now calls the installed dialer against
`169.254.169.254:80` and requires `ErrMediaAddrBlocked`.

The ingest tests use no stubs for the download or the decrypt: an `httptest`
server encrypts a known payload with a from-scratch implementation of WeCom's
algorithm, and the assertion is that exactly those bytes reach storage. They
fetch through the real guarded client with loopback added to the policy, so the
production dialer is exercised rather than bypassed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -182,6 +182,22 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 #                              and the wecom Web-UI endpoints return 503.
 #                              Generate with: openssl rand -base64 32
 #
+#   MULTICA_WECOM_MEDIA_ALLOW_CIDRS
+#                              Comma-separated ranges the media fetcher may
+#                              dial even though they look reserved. Empty by
+#                              default, and that is the right value almost
+#                              everywhere: the guard exists because a callback
+#                              hands us a URL somebody else controls, and a
+#                              redirect to a metadata endpoint is one line of
+#                              response away.
+#                              Set it only behind a fake-IP proxy, where every
+#                              public hostname resolves into the proxy's pool
+#                              (commonly 198.18.0.0/15) and WeCom's own image
+#                              host is therefore refused along with everything
+#                              else. Whatever you list here becomes reachable
+#                              by a URL WeCom supplies. Loopback and the
+#                              private ranges are refused regardless and
+#                              cannot be opened this way.
 #   MULTICA_WECOM_TRACE        1 = log every frame this adapter reads and
 #                              writes, so a real-device session can be checked
 #                              against the server afterwards. Empty / anything
@@ -214,6 +230,7 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 # the lease are dropped and the user sees nothing. Until cross-replica outbound
 # routing lands, run the WeCom-enabled backend as a SINGLE replica.
 MULTICA_WECOM_SECRET_KEY=
+MULTICA_WECOM_MEDIA_ALLOW_CIDRS=
 MULTICA_WECOM_TRACE=
 
 # S3 / CloudFront
--- a/.env.example
+++ b/.env.example
@@ -195,9 +195,11 @@ GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
 #                              (commonly 198.18.0.0/15) and WeCom's own image
 #                              host is therefore refused along with everything
 #                              else. Whatever you list here becomes reachable
-#                              by a URL WeCom supplies. Loopback and the
-#                              private ranges are refused regardless and
-#                              cannot be opened this way.
+#                              by a URL WeCom supplies. Loopback, the private
+#                              ranges and the IPv6 translation prefixes
+#                              (64:ff9b::/96, 64:ff9b:1::/48, 2002::/16 —
+#                              IPv4 addresses in an IPv6 spelling) are refused
+#                              regardless and cannot be opened this way.
 #   MULTICA_WECOM_TRACE        1 = log every frame this adapter reads and
 #                              writes, so a real-device session can be checked
 #                              against the server afterwards. Empty / anything
--- a/docker-compose.selfhost.yml
+++ b/docker-compose.selfhost.yml
@@ -146,6 +146,11 @@ services:
       # subscribe credentials at connection time, and to seal the secret when
       # an installation is created from the WeCom settings tab.
       MULTICA_WECOM_SECRET_KEY: ${MULTICA_WECOM_SECRET_KEY:-}
+      # Comma-separated ranges the inbound media fetcher may dial even though
+      # they look reserved. Empty (the default) keeps the SSRF guard as strict
+      # as it ships; set it only behind a fake-IP proxy whose pool would
+      # otherwise get every attachment refused. See .env.example.
+      MULTICA_WECOM_MEDIA_ALLOW_CIDRS: ${MULTICA_WECOM_MEDIA_ALLOW_CIDRS:-}
       # 1 = log every inbound and outbound WeCom frame, including the first
       # 120 runes of each message body, so a real-device session can be
       # checked against the server afterwards. Off unless set; turn it on only
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -727,6 +727,21 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 				// user in WeCom sees no response.
 				wecom.NewOutbound(queries, wecomSenders, slog.Default()).Register(bus)
 
+				// Ranges the media fetcher may dial despite looking reserved.
+				// Empty by default, which leaves the SSRF guard exactly as
+				// strict as it ships. A deployment behind a fake-IP proxy
+				// needs it: there, every public hostname resolves into the
+				// proxy's pool (198.18.0.0/15 is the common one), so WeCom's
+				// own COS host is indistinguishable from a metadata endpoint
+				// by address alone and every attachment is refused.
+				if raw := strings.TrimSpace(os.Getenv("MULTICA_WECOM_MEDIA_ALLOW_CIDRS")); raw != "" {
+					for _, err := range wecom.SetMediaAllowedPrefixes(strings.Split(raw, ",")) {
+						slog.Error("wecom: ignoring malformed media allow cidr", "error", err)
+					}
+					slog.Warn("wecom: media guard has an operator allow-list; those ranges are reachable by a URL WeCom supplies",
+						"cidrs", raw)
+				}
+
 				// Frame tracing: off unless an operator asks for it. It
 				// records a bounded prefix of message text, so the fact that
 				// it is on has to be visible in the log it is writing into —
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -701,8 +701,23 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 					Metrics:     wecomMetricsOrNil(opts.WecomMetrics),
 					Logger:      slog.Default(),
 				})
+				// Inbound media: a callback carries a pre-signed COS url and
+				// a per-url key, so the resolver needs no WeCom credential —
+				// only somewhere durable to put the bytes. Without an object
+				// store there is nothing to point an attachment at, so the
+				// resolver is left nil and attachments stay as their
+				// placeholder text. Same nil-guard as DingTalk above.
+				var wecomMedia engine.MediaResolver
+				if store != nil {
+					wecomMedia = wecom.NewMediaResolver(
+						store,
+						engine.NewDBMediaIntentLedger(queries),
+						wecomSenders,
+						slog.Default(),
+					)
+				}
 				channelRouter.Register(wecom.TypeWecom, wecom.NewResolverSet(
-					wecomStore, wecomSession, wecomReplier,
+					wecomStore, wecomSession, wecomReplier, wecomMedia,
 				))
 
 				// EventChatDone subscriber: pushes the agent's chat reply
--- a/server/cmd/server/wecom_media_allow_wiring_test.go
+++ b/server/cmd/server/wecom_media_allow_wiring_test.go
@@ -1,0 +1,142 @@
+package main
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"testing"
+)
+
+// routerSourceFile is parsed by TestWecomMediaAllowListIsWiredAtBoot. Kept as
+// a constant so a rename shows up as one failing test rather than a silently
+// skipped guard.
+const routerSourceFile = "router.go"
+
+const wecomMediaAllowEnv = "MULTICA_WECOM_MEDIA_ALLOW_CIDRS"
+
+// TestWecomMediaAllowListIsWiredAtBoot guards the operator switch behind the
+// media guard's allow-list.
+//
+// wecom.SetMediaAllowedPrefixes widens the SSRF guard that stands between a
+// WeCom-supplied URL and the deployment's own network. It shipped once with no
+// production caller at all: the exemption existed, its own doc comment claimed
+// it was "called at boot from MULTICA_WECOM_MEDIA_ALLOW_CIDRS", and nothing
+// read that variable. An operator behind a fake-IP proxy — where every
+// hostname resolves into 198.18.0.0/15 and the guard therefore refuses every
+// download — had the exemption and no way to turn it on.
+//
+// This parses router.go rather than asserting on behaviour, for the same
+// reason TestMainUsesRouterOwnedBackgroundServices does: the regression being
+// guarded is the absence of a call, and no value-level assertion can observe a
+// call that was never made. The allow-list itself lives in unexported package
+// state, so the only thing reachable from here is the wiring.
+func TestWecomMediaAllowListIsWiredAtBoot(t *testing.T) {
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, routerSourceFile, nil, 0)
+	if err != nil {
+		t.Fatalf("parse %s: %v", routerSourceFile, err)
+	}
+
+	var (
+		readsEnv  string // position of os.Getenv(wecomMediaAllowEnv)
+		appliesIt string // position of wecom.SetMediaAllowedPrefixes(...)
+	)
+	for _, decl := range file.Decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Body == nil {
+			continue
+		}
+		// Both must sit in the SAME function. Reading the variable in one
+		// place and calling the setter in another is how the two halves drift
+		// apart again — e.g. a caller that only ever passes a constant.
+		var envHere, applyHere string
+		ast.Inspect(fn.Body, func(n ast.Node) bool {
+			call, ok := n.(*ast.CallExpr)
+			if !ok {
+				return true
+			}
+			switch calleeName(call) {
+			case "os.Getenv":
+				if len(call.Args) == 1 && stringLit(call.Args[0]) == wecomMediaAllowEnv {
+					envHere = fset.Position(call.Pos()).String()
+				}
+			case "wecom.SetMediaAllowedPrefixes":
+				applyHere = fset.Position(call.Pos()).String()
+			}
+			return true
+		})
+		if envHere != "" && applyHere != "" {
+			readsEnv, appliesIt = envHere, applyHere
+			break
+		}
+		if envHere != "" {
+			readsEnv = envHere
+		}
+		if applyHere != "" {
+			appliesIt = applyHere
+		}
+	}
+
+	if readsEnv == "" {
+		t.Errorf("no os.Getenv(%q) in %s — the media guard's allow-list has no operator switch again",
+			wecomMediaAllowEnv, routerSourceFile)
+	}
+	if appliesIt == "" {
+		t.Errorf("no call to wecom.SetMediaAllowedPrefixes in %s — %s is read but never reaches the guard",
+			routerSourceFile, wecomMediaAllowEnv)
+	}
+	if t.Failed() {
+		t.FailNow()
+	}
+
+	// The errors SetMediaAllowedPrefixes returns are the only report an
+	// operator gets for a typo'd CIDR. Dropping them on the floor turns a
+	// malformed entry into a guard that is silently narrower than the .env
+	// says it is.
+	if !reportsMalformedCIDRs(fset, file) {
+		t.Errorf("the []error from wecom.SetMediaAllowedPrefixes at %s is discarded — a malformed %s entry would be dropped silently",
+			appliesIt, wecomMediaAllowEnv)
+	}
+}
+
+// reportsMalformedCIDRs reports whether the SetMediaAllowedPrefixes call site
+// consumes the returned errors — ranged over, assigned, or passed on — rather
+// than being used as a bare statement.
+func reportsMalformedCIDRs(fset *token.FileSet, file *ast.File) bool {
+	var consumed bool
+	ast.Inspect(file, func(n ast.Node) bool {
+		if consumed {
+			return false
+		}
+		switch stmt := n.(type) {
+		case *ast.ExprStmt:
+			// A bare `wecom.SetMediaAllowedPrefixes(...)` throws the errors away.
+			if call, ok := stmt.X.(*ast.CallExpr); ok && calleeName(call) == "wecom.SetMediaAllowedPrefixes" {
+				return false
+			}
+		case *ast.RangeStmt:
+			if call, ok := stmt.X.(*ast.CallExpr); ok && calleeName(call) == "wecom.SetMediaAllowedPrefixes" {
+				consumed = true
+				return false
+			}
+		case *ast.AssignStmt:
+			for _, rhs := range stmt.Rhs {
+				if call, ok := rhs.(*ast.CallExpr); ok && calleeName(call) == "wecom.SetMediaAllowedPrefixes" {
+					consumed = true
+					return false
+				}
+			}
+		}
+		return true
+	})
+	return consumed
+}
+
+// stringLit returns the value of a plain string literal, or "".
+func stringLit(e ast.Expr) string {
+	lit, ok := e.(*ast.BasicLit)
+	if !ok || lit.Kind != token.STRING || len(lit.Value) < 2 {
+		return ""
+	}
+	return lit.Value[1 : len(lit.Value)-1]
+}
--- a/server/internal/integrations/wecom/dispatch_test.go
+++ b/server/internal/integrations/wecom/dispatch_test.go
@@ -1,9 +1,10 @@
 package wecom
 
 // dispatch_test.go — dispatchFrame routing, driven with a recording wsConn and
-// a capturing handler (no network). Covers the text-only receipt for non-text
-// messages, the text path reaching the handler, the superseded-disconnect
-// escalation, and the server-ping / pong frames.
+// a capturing handler (no network). Covers the receipt for a kind that cannot
+// be read, the text path reaching the handler, the superseded-disconnect
+// escalation, and the server-ping / pong frames. Media routing lives next
+// door in inbound_media_test.go.
 
 import (
 	"context"
@@ -55,27 +56,35 @@ func TestDispatchFrame_TextReachesHandler(t *testing.T) {
 	}
 }
 
-func TestDispatchFrame_NonTextGetsReceiptAndSkipsHandler(t *testing.T) {
+func TestDispatchFrame_UnreadableKindGetsReceiptAndSkipsHandler(t *testing.T) {
 	t.Parallel()
-	called := false
-	c := testChannel(func(context.Context, channel.InboundMessage) error {
-		called = true
-		return nil
-	})
-	conn := &recordingConn{}
-	err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, "image", ""), conn.autoAck(newWSSender(conn, nil)), slog.Default())
-	if err != nil {
-		t.Fatalf("dispatchFrame: %v", err)
-	}
-	if called {
-		t.Error("handler must NOT be called for a non-text message")
-	}
-	if len(conn.frames) != 1 {
-		t.Fatalf("expected one text-only receipt, got %d frames", len(conn.frames))
-	}
-	body := conn.sendBody(t, 0)
-	if body["chatid"] != "CHAT_1" {
-		t.Errorf("receipt chatid = %v, want the same chat CHAT_1", body["chatid"])
+	// Two ways a callback can be unreadable: a kind this adapter does not
+	// model at all, and a kind it does model that arrived without the one
+	// field that makes it usable (an image frame carrying no url).
+	for _, name := range []string{"location", "image"} {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			called := false
+			c := testChannel(func(context.Context, channel.InboundMessage) error {
+				called = true
+				return nil
+			})
+			conn := &recordingConn{}
+			err := c.dispatchFrame(context.Background(), msgCallbackFrame(t, name, ""), conn.autoAck(newWSSender(conn, nil)), slog.Default())
+			if err != nil {
+				t.Fatalf("dispatchFrame: %v", err)
+			}
+			if called {
+				t.Error("handler must NOT be called for a message with nothing readable in it")
+			}
+			if len(conn.frames) != 1 {
+				t.Fatalf("expected one receipt, got %d frames", len(conn.frames))
+			}
+			body := conn.sendBody(t, 0)
+			if body["chatid"] != "CHAT_1" {
+				t.Errorf("receipt chatid = %v, want the same chat CHAT_1", body["chatid"])
+			}
+		})
 	}
 }
 
--- a/server/internal/integrations/wecom/inbound_media_bind_db_test.go
+++ b/server/internal/integrations/wecom/inbound_media_bind_db_test.go
@@ -1,0 +1,390 @@
+package wecom
+
+// inbound_media_bind_db_test.go — the other half of inbound media: what
+// happens to a MediaRef AFTER the resolver has stored the object.
+//
+// media_ingest_test.go proves the download/decrypt/upload works. It cannot
+// see this defect, because it calls ResolveMedia directly and never asks who
+// consumes what it returns. The answer, on the shipped code, was nobody: the
+// wecom sessionBinder's BindMedia returned nil without touching the session
+// store, so a photo landed in object storage, its intent row stayed 'pending'
+// forever, no attachment row was ever created, and the agent kept seeing the
+// bare "[Image]" placeholder. Nothing was logged on that path — the Router
+// only warns when BindMedia returns an error, and returning nil is how you
+// say "bound fine".
+//
+// So this test drives the whole real path — engine.Router.Handle over the
+// real wecom ResolverSet, the real engine.ChatSession, a real Postgres and a
+// real HTTP download of really-encrypted bytes — and asks the database the
+// one question the user asked: is there an attachment on that message?
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/multica-ai/multica/server/internal/integrations/channel"
+	"github.com/multica-ai/multica/server/internal/integrations/channel/engine"
+	"github.com/multica-ai/multica/server/internal/service"
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+func mediaBindTestDB(t *testing.T) *pgxpool.Pool {
+	t.Helper()
+	dsn := os.Getenv("DATABASE_URL")
+	if dsn == "" {
+		dsn = "postgres://multica:multica@localhost:5432/multica?sslmode=disable"
+	}
+	ctx := context.Background()
+	pool, err := pgxpool.New(ctx, dsn)
+	if err != nil {
+		t.Skipf("no database: %v", err)
+	}
+	if err := pool.Ping(ctx); err != nil {
+		pool.Close()
+		t.Skipf("database not reachable: %v", err)
+	}
+	var migrated bool
+	if err := pool.QueryRow(ctx, `
+		SELECT to_regclass('public.attachment') IS NOT NULL
+		   AND to_regclass('public.channel_media_pending_object') IS NOT NULL`).Scan(&migrated); err != nil || !migrated {
+		pool.Close()
+		t.Skip("channel media tables not present (database not migrated)")
+	}
+	t.Cleanup(pool.Close)
+	return pool
+}
+
+type mediaBindFixture struct {
+	workspaceID    pgtype.UUID
+	userID         pgtype.UUID
+	agentID        pgtype.UUID
+	installationID pgtype.UUID
+	botID          string
+	senderID       string
+}
+
+// seedMediaBindFixture builds exactly what the wecom ResolverSet resolves
+// against: an installation keyed by bot id, and a user binding keyed by the
+// anonymized aibot userid. Everything hangs off one workspace so cleanup is
+// a single cascade.
+func seedMediaBindFixture(t *testing.T, pool *pgxpool.Pool) mediaBindFixture {
+	t.Helper()
+	ctx := context.Background()
+	suffix := time.Now().UnixNano()
+	f := mediaBindFixture{
+		botID:    fmt.Sprintf("wb-bind-%d", suffix),
+		senderID: fmt.Sprintf("T-bind-%d", suffix),
+	}
+	var runtimeID pgtype.UUID
+
+	if err := pool.QueryRow(ctx, `INSERT INTO "user" (name, email) VALUES ($1, $2) RETURNING id`,
+		"wecom media bind", fmt.Sprintf("wecom-media-bind-%d@multica.test", suffix)).Scan(&f.userID); err != nil {
+		t.Fatalf("create user: %v", err)
+	}
+	t.Cleanup(func() {
+		cleanup := context.Background()
+		if f.workspaceID.Valid {
+			_, _ = pool.Exec(cleanup, `DELETE FROM channel_media_pending_object WHERE workspace_id = $1`, f.workspaceID)
+			_, _ = pool.Exec(cleanup, `DELETE FROM channel_inbound_message_dedup WHERE installation_id = $1`, f.installationID)
+			_, _ = pool.Exec(cleanup, `DELETE FROM workspace WHERE id = $1`, f.workspaceID)
+		}
+		if f.userID.Valid {
+			_, _ = pool.Exec(cleanup, `DELETE FROM "user" WHERE id = $1`, f.userID)
+		}
+	})
+	if err := pool.QueryRow(ctx, `INSERT INTO workspace (name, slug, description) VALUES ($1, $2, '') RETURNING id`,
+		"wecom media bind", fmt.Sprintf("wecom-media-bind-%d", suffix)).Scan(&f.workspaceID); err != nil {
+		t.Fatalf("create workspace: %v", err)
+	}
+	if _, err := pool.Exec(ctx, `INSERT INTO member (workspace_id, user_id, role) VALUES ($1, $2, 'owner')`,
+		f.workspaceID, f.userID); err != nil {
+		t.Fatalf("create member: %v", err)
+	}
+	if err := pool.QueryRow(ctx, `
+		INSERT INTO agent_runtime (workspace_id, name, runtime_mode, provider, owner_id)
+		VALUES ($1, $2, 'local', 'multica_daemon', $3) RETURNING id`,
+		f.workspaceID, fmt.Sprintf("wecom-bind-runtime-%d", suffix), f.userID).Scan(&runtimeID); err != nil {
+		t.Fatalf("create runtime: %v", err)
+	}
+	if err := pool.QueryRow(ctx, `
+		INSERT INTO agent (workspace_id, name, runtime_mode, runtime_id, owner_id)
+		VALUES ($1, $2, 'local', $3, $4) RETURNING id`,
+		f.workspaceID, fmt.Sprintf("wecom-bind-agent-%d", suffix), runtimeID, f.userID).Scan(&f.agentID); err != nil {
+		t.Fatalf("create agent: %v", err)
+	}
+	if err := pool.QueryRow(ctx, `
+		INSERT INTO channel_installation (workspace_id, agent_id, channel_type, config, installer_user_id, status)
+		VALUES ($1, $2, 'wecom',
+		        jsonb_build_object('app_id', $3::text, 'bot_id', $3::text, 'app_secret_encrypted', ''),
+		        $4, 'active')
+		RETURNING id`, f.workspaceID, f.agentID, f.botID, f.userID).Scan(&f.installationID); err != nil {
+		t.Fatalf("create installation: %v", err)
+	}
+	if _, err := pool.Exec(ctx, `
+		INSERT INTO channel_user_binding (workspace_id, multica_user_id, installation_id, channel_type, channel_user_id)
+		VALUES ($1, $2, $3, 'wecom', $4)`,
+		f.workspaceID, f.userID, f.installationID, f.senderID); err != nil {
+		t.Fatalf("create user binding: %v", err)
+	}
+	return f
+}
+
+// ---- the Router's collaborators, reduced to what this path touches ----
+
+type bindTestIssues struct{}
+
+func (bindTestIssues) Create(context.Context, service.IssueCreateParams, service.IssueCreateOpts) (service.IssueCreateResult, error) {
+	return service.IssueCreateResult{}, errors.New("no /issue in this test")
+}
+func (bindTestIssues) PublishAttachmentsChanged(context.Context, db.Issue, pgtype.UUID) {}
+
+// bindTestTasks stands in for the task service. Enqueueing a real chat task
+// would need a live runtime, and the run trigger is not what is under test —
+// but the media-ready promotion IS called on the media path, so it is
+// recorded rather than ignored.
+type bindTestTasks struct{ promoted int }
+
+func (bindTestTasks) EnqueueChatTask(context.Context, db.ChatSession, pgtype.UUID, bool) (db.AgentTaskQueue, error) {
+	return db.AgentTaskQueue{}, nil
+}
+func (b *bindTestTasks) PromoteChannelChatTasksIfMediaReady(context.Context, pgtype.UUID) error {
+	b.promoted++
+	return nil
+}
+func (bindTestTasks) PromoteDeferredChannelIssueTask(context.Context, pgtype.UUID) error { return nil }
+
+// wecomImageCallback builds the inbound envelope for a standalone image,
+// through the real frame decoder so the fixture cannot drift from the wire.
+func wecomImageCallback(t *testing.T, botID, senderID, msgID, url string) channel.InboundMessage {
+	t.Helper()
+	raw, err := json.Marshal(map[string]any{
+		"msgid":    msgID,
+		"aibotid":  botID,
+		"chattype": "single",
+		"chatid":   senderID,
+		"from":     map[string]any{"userid": senderID},
+		"msgtype":  "image",
+		"image":    map[string]any{"url": url, "aeskey": testAESKey},
+	})
+	if err != nil {
+		t.Fatalf("marshal callback: %v", err)
+	}
+	var mc aibotMsgCallback
+	if err := json.Unmarshal(raw, &mc); err != nil {
+		t.Fatalf("decode callback: %v", err)
+	}
+	text, ok := mc.ownText()
+	if !ok {
+		t.Fatal("image callback is not routable; the fixture is wrong")
+	}
+	return channelMessageFromCallback(botID, mc, text, "req-bind-1")
+}
+
+// stalledCOSServer accepts the request and then holds it open, so the media
+// path is genuinely still in flight while the durable message is inspected.
+func stalledCOSServer(release <-chan struct{}) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case <-release:
+		case <-r.Context().Done():
+		}
+	}))
+}
+
+// TestInboundImageBecomesAnAttachmentOnTheChatMessage is the regression for
+// "the image downloads fine and the agent still only sees [Image]".
+//
+// Every assertion here is a thing the user could check by hand in the
+// database, and on the shipped code all four were false while the file sat
+// happily in object storage.
+func TestInboundImageBecomesAnAttachmentOnTheChatMessage(t *testing.T) {
+	pool := mediaBindTestDB(t)
+	fixture := seedMediaBindFixture(t, pool)
+	ctx := context.Background()
+	queries := db.New(pool)
+
+	plaintext := []byte("\xff\xd8\xff\xe0 not really a jpeg, but really encrypted")
+	srv := cosServer(t, plaintext, `attachment; filename="holiday.jpg"`)
+	defer srv.Close()
+
+	// The production wiring, with two seams swapped for test doubles that do
+	// not change the path: object bytes go to memory instead of disk/S3, and
+	// the HTTP client also permits loopback so the httptest server is
+	// reachable. The intent ledger is the REAL database one — the bind
+	// transaction claims those rows, so a fake would hide the claim.
+	storage := &fakeMediaStorage{}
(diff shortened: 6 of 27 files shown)


seacen:
Pushed a fix for something this PR shipped half-wired. Found it running the branch against a live tenant.

**The resolver worked; the binder threw the result away.** The image downloaded, decrypted and stored — 509KB of JPEG on disk with correct metadata — and then `sessionBinder.BindMedia` returned `nil` without binding anything, which the Router reads as success. No attachment row, so the agent saw only the `[Image]` placeholder and found nothing in its workdir. **Nothing was logged**, because from the engine's side nothing failed.

The stub was correct before this PR — WeCom registered no MediaResolver, there was genuinely nothing to bind, and the comment above it said exactly that. Adding the resolver made the comment false and left the code under it untouched. `engineSessionBinder` did not even declare `BindMediaRefs`, so the binder was structurally incapable of binding.

**Same cause one function up:** `AppendMessage` never forwarded `MediaPendingSeconds`, so `channel_media_pending_until` stayed NULL and the chat task fired immediately, handing the agent the placeholder mid-download. Fixing only the binder would have left a race that still shows `[Image]` whenever the run wins.

`lark/feishu_resolvers.go:223` passes all of it including `IssueID` — which is what makes an `/issue` turn's attachment belong to the issue rather than to a chat message nobody opens again. dingtalk and slack the same; WeCom was the outlier.

Three tests. The end-to-end one drives `engine.Router.Handle` over the real resolver set, the real chat session, real Postgres and a real download of genuinely AES-encrypted bytes, then asks the database whether an attachment exists. Reverting the binder:

```
attachment rows on the chat message = 0, want 1 — the image was stored but
never bound, so the agent only ever sees "[Image]"

channel_media_pending_until is NULL on a message carrying an image — the chat
task will fire at once and hand the agent the placeholder while the download
is still running
```

It also replaces `TestSessionBinder_BindMediaIsNoop`, which asserted the no-op as intended behaviour (`want nil (wecom resolves no media)`) and would have kept passing forever.

seacen:
Pushed `ca574d3f5`.

**An operator-facing switch that was missing.** `SetMediaAllowedPrefixes` had no production caller — its doc comment claimed it was "called at boot from `MULTICA_WECOM_MEDIA_ALLOW_CIDRS`", and nothing on this branch read that variable. The only callers were its own tests. That left the exemption unreachable for exactly the deployment it exists for: behind a fake-IP proxy every hostname resolves into `198.18.0.0/15`, so the SSRF guard refuses every download and inbound media is dead with no way to turn it back on.

Now wired in three places: `router.go` reads it at boot (logging each malformed entry, and warning which ranges were opened), `.env.example` documents it, `docker-compose.selfhost.yml` passes it through.

**The security property is unchanged**, and now tested through the dialer rather than through the policy function alone. The allow-list is consulted only inside the `reservedMediaPrefixes` loop, so loopback, private and link-local are refused before it is reached — `MULTICA_WECOM_MEDIA_ALLOW_CIDRS=0.0.0.0/0` does not open `127.0.0.1`. The new test points the guard at a real listening server on loopback with the widest possible allow-list and asserts no socket is opened. Reverted the ordering to check the test can fail, and it does.

`cmd/server` also gets an AST guard that fails if the boot wiring goes missing again, or if the parse errors stop being reported. A value-level assertion cannot observe a call that was never made, which is the regression here.

**A stale comment removed.** The `BindMedia` doc in `wecom_resolvers.go` still said wecom registers no `MediaResolver`, the Router never resolves media, and the method is never called at runtime — every clause false as of this diff, sitting directly above the working implementation this PR added.

**Red CI fixed.** `TestInboundImageBecomesAnAttachmentOnTheChatMessage` raced under `-race`. It is the test rig, not production: `bindTestTasks.promoted` is written from the Router's detached media goroutine and was read unguarded from the test goroutine. The DB poll the test waits on is not a synchronisation edge — the bind transaction commits one call earlier than the promotion in `resolveAndBindMedia`, so the read was both racy and, on the wrong interleaving, a read of zero. Guarded with a mutex and an accessor (matching the doubles in `engine/router_test.go`, and `fakeMediaStorage` one file over), and the poll now waits for the promotion rather than assuming it happened. `service.TaskService.PromoteChannelChatTasksIfMediaReady` keeps no in-process mutable state, so nothing on a live tenant was affected. Reverted: race reported on the first of 10 runs. Restored: 30 clean runs.

**Two more comments this diff falsified**, corrected while in there:

- `media_stream.go` described two temp files with "ciphertext lands in the first as it arrives". There is one (`os.CreateTemp` at line 63, unlinked immediately) and the ciphertext streams from the socket without touching disk. The same header also narrated a prior buffered-only state that never existed — `media_download.go` and `media_stream.go` both arrived in `3fe8e453d`.
- `maxMediaBytes` called itself "the per-download memory bound". True only on the fallback path; both shipped storage backends implement `UploadStream`, so a real deployment takes `media_stream.go` and the ceiling bounds disk.

Plus `publicAddrOnly`'s doc comment moved back onto `publicAddrOnly` — the allow-list was inserted between the two, leaving godoc attributing it to `mediaAllowedPrefixes`.

**Three things I found and did not change**, since they are calls for you rather than mechanical fixes:

1. `mediaGuard.dialer` (`media_guard.go:166`) has no writer anywhere — not production, not tests. Every `mediaGuard{...}` literal in the tree leaves it nil, and line 193 immediately falls back to `&net.Dialer{Timeout: mediaDialTimeout}`. Unlike `allow` / `resolve` / `onRefuse` it is not documented as a seam, so it reads as an injectable connect timeout that nothing can inject. Harmless, but it is the same shape as the defect above.
2. `ErrMediaAddrBlocked` is exported and has no production reader. It is produced at `media_guard.go:185` and consumed only by tests; nothing outside package `wecom` references it. Its doc says "the caller logs the two differently" — the real caller chain funnels every download error into one uniform log line at `media_ingest.go:132`, and `classifyMediaFailure` branches only on `errMediaTooLarge`. Either the differentiated logging is worth adding or the export is worth dropping.
3. `wecom_resolvers.go:66` says DingTalk's nil-check "is there for `*ackNotifier`". `dingtalk/resolvers.go:50-56` has two ifs and the media one is not about `ackNotifier`. The wecom code is still right — `media` is an interface parameter and `router.go` declares `var wecomMedia engine.MediaResolver`, so a nil argument is a nil interface. Only the parenthetical is wrong.

`go build ./...`, `go vet ./...` clean. `go test -race ./internal/integrations/wecom/ ./internal/integrations/channel/engine/ ./cmd/server/` green, and the wecom package green at `-count=5` under `-race`.

seacen:
Pushed `cc1492be9` — the three residuals from the last round.

**1. `mediaGuard.dialer` deleted.** No writer anywhere, production or test: every `mediaGuard{...}` literal in the tree left it nil and `dial` fell straight through to `&net.Dialer{Timeout: mediaDialTimeout}`. Unlike `allow` / `resolve` / `onRefuse` it was not documented as a seam, so it read as an injectable connect timeout that nothing could inject. Removing it was not disruptive — nothing depended on the zero value, and the default now lives outside the address loop, which also stops allocating a `net.Dialer` per resolved address.

**2. `ErrMediaAddrBlocked` is now read in production, rather than unexported.** Its doc claimed "the caller logs the two differently" and the caller did not: every download failure funnelled into one `wecom media ingest failed` line, and `classifyMediaFailure` branched only on `errMediaTooLarge`. That is the wrong place to be silent — an operator behind a fake-IP proxy is exactly who `MULTICA_WECOM_MEDIA_ALLOW_CIDRS` exists for, and a generic failure gives them no reason to look at their own configuration instead of at WeCom.

`classifyMediaFailure` now branches on it into a new `mediaFailureBlocked`, and the blocked case gets its own operator line naming the guard as the refuser and pointing at the remedy:

> `wecom media ingest refused by the media address guard: the host resolved to a non-public address and was not dialed. If this deployment sits behind a fake-IP proxy, declare its range in MULTICA_WECOM_MEDIA_ALLOW_CIDRS; otherwise this is a URL that should not have been sent.`

On the chat side the sender still gets the plain did-not-arrive wording. A refused address is not something they can act on, and neither the resolved address nor the signed URL belongs in a chat message — `stripURL` has already taken the URL out of the error, and the guard's refusal carries no address of its own, so the log is safe whole and the notice is a static string. There is a test asserting the notice contains none of `127.0.0.1`, `::1`, the server URL, or the AES key.

One consequence worth flagging for review: two failure kinds now render to the same sentence, so the notice dedupe moved from `appendFailure` (which dedupes by kind) into the line build. Without that, a message with one blocked and one unreadable attachment would say the same thing to the sender twice.

**3. The DingTalk parenthetical fixed.** It claimed DingTalk's media nil-check "is there for `*ackNotifier`". It is not — `NewDingTalkResolverSet` takes `media engine.MediaResolver`, so the `if` guarding that assignment is redundant for the same reason wecom assigns straight through. The typed-nil hazard is on the `if` above it, where `ack` is a concrete `*ackNotifier`. The wecom code was always right; only the cross-reference was wrong.

**Tests.** Both new tests drive real errors rather than hand-built ones. The refusal travels from the dialer through `http.Client`'s `*url.Error` and then `stripURL` before anything classifies it, so a test constructing `ErrMediaAddrBlocked` directly would keep passing even if that chain stopped preserving `errors.Is` — which is how the gap survived in the first place. Reverse-verified both:

- Drop the `ErrMediaAddrBlocked` branch → `classifyMediaFailure(guard refusal) = 0, want mediaFailureBlocked (2)`
- Drop the notice dedupe → the notice comes back as the did-not-arrive sentence twice, joined by a newline.

`go build ./...` and `go vet ./...` clean. `go test -race` green across `wecom`, `channel/engine`, `cmd/server` and `dingtalk`, and `wecom` green at `-count=5` under `-race`.

Bohan-J (changes requested):
Thanks for this — the product direction is right and most of the implementation is careful work. Requesting changes on three items before merge.

## Must fix

### 1. [Security] The address guard misses the RFC 8215 local-use NAT64 prefix

`server/internal/integrations/wecom/media_guard.go` — `reservedMediaPrefixes` rejects the well-known NAT64 prefix `64:ff9b::/96`, but not `64:ff9b:1::/48` (RFC 8215, local-use NAT64). An address in that range falls through `publicAddrOnly` to `true`.

On a deployment that uses the local-use prefix, that means an attacker-supplied media URL can reach IPv4 private or link-local targets through the translator — which defeats the "only dial public addresses" guarantee this file exists to provide.

Please add the prefix to the deny list, and cover it with a regression test that goes through the real dialer rather than only asserting on the predicate. Worth re-checking the full list against the IANA IPv6 Special-Purpose Address Registry while you're in there.

### 2. [Correctness] Mixed messages feed attachment placeholders into the command source, so `/issue` after an image never fires

`ownText()` in `ws_frame.go` renders a 图文混排 message by joining its runs with `\n`, each attachment rendered as `[Image]` / `[File]` / `[Video]`. `wecom_resolvers.go` then does:

```go
CommandText: p.Message.Text, // wecom has no enrichment; command == body
```

`engine.ParseIssueCommand` only considers the first non-empty line ("A body whose first non-empty line is anything other than the literal prefix is not an issue command, even if `/issue` appears later"). So a user who attaches a screenshot and then types `/issue 登录坏了` produces:

```
[Image]
/issue 登录坏了
```

First line is `[Image]` → no issue is created, and the user gets no signal about why. Image-after-text happens to work; image-before-text silently doesn't.

Fix: build `CommandText` from the mixed message's text/voice runs only, without the media placeholders, and add routing/binding tests for the image-before-command ordering. This interacts with #3 — see below.

### 3. [Merge gate] The branch conflicts with `main`, and the conflict is load-bearing

GitHub reports `DIRTY / CONFLICTING` across `.env.example`, `docker-compose.selfhost.yml`, `server/cmd/server/router.go`, `wecom_channel.go`, `ws_frame.go`, `ws_frame_test.go`.

`main` has since gained work in exactly the code this PR rewrites: `ws_frame.go` now computes a separate command source (`stripLeadingMentions` for group chats), sets `CommandText` from it, and drives `SkipAgentRun: isIssueCommand(command)` off that same string. This branch predates that and would revert it. Please rebase and keep both sides — main's group-mention stripping, `CommandText`, the single `engine.ParseIssueCommand` decision and the redacted `MULTICA_WECOM_TRACE`, plus this PR's media path.

One thing to watch while resolving: main derives `command` from `mc.Text.Content`, which is empty for a `mixed` callback — the runs live in `mc.Mixed.MsgItem`. So a straight rebase leaves `CommandText` empty for mixed messages, the shared fallback in `engine/session.go` substitutes the placeholder-bearing body, and #2 reproduces through a different path. Both want the same fix: one placeholder-free command source built from the text/voice runs, mention-stripped in groups.

A regression test for "group @bot + mixed + `/issue`" would keep the conflict resolution honest.

## Implementation quality

The rest reads well: intent recorded before upload, per-connection address validation with every redirect hop re-checked, signed URLs redacted out of errors, streaming decryption to a temp file, single-attachment object keys, partial-failure notification, and a clean degrade to placeholder text when no object storage is configured. Targeted coverage on those paths is solid.

The remaining risk is the real-world boundary of external downloads and temp disk. After the fixes, worth one pass through a WeCom sandbox for timeouts, object-storage permissions, and an attachment near the size limit.

## Verified at 523bc5e

- ✅ `go test -race ./internal/integrations/wecom -count=1`
- ✅ `go vet ./internal/integrations/wecom`
- ✅ `go test ./cmd/server -run TestWecomMediaAllowListIsWiredAtBoot -count=1`
- CI is green on this HEAD, but this HEAD is not based on latest `main` — it needs a re-run after the rebase.

seacen:
All three fixed and rebased onto `main`. Head is now `4900e17`; the branch is `MERGEABLE` / `CLEAN` and CI is green on it — all nine checks, including `backend`, on a HEAD that is based on latest `main`.

## 1. The local-use NAT64 prefix, plus what the registry re-check turned up

`64:ff9b:1::/48` is in `reservedMediaPrefixes` now. You were right about the consequence: `64:ff9b:1::a9fe:a9fe` is `169.254.169.254` the moment a translator sees it, so on a deployment running that prefix the whole "only dial public addresses" guarantee was gone.

The regression test goes through the real dialer. `TestTheGuardRefusesBothNAT64Prefixes` covers both prefixes as URL literals and as what a hostname resolves to, and `TestARedirectIntoTheLocalUseNAT64PrefixIsRefused` covers the second hop against a real server. What makes the dialer the thing under test is *which* error comes back: `ErrMediaAddrBlocked` means the address never got past the policy, and a connect timeout means a socket was opened. Deleting the prefix line:

```
=== RUN   TestTheGuardRefusesBothNAT64Prefixes
    media_guard_test.go:191: http://[64:ff9b:1::a9fe:a9fe]/latest/meta-data/ failed with wecom: media download: dial tcp [64:ff9b:1::a9fe:a9fe]:80: i/o timeout, want the guard's refusal (any other error means a socket was opened toward an address that reaches IPv4 private space through the translator)
    media_guard_test.go:191: http://[64:ff9b:1::7f00:1]:9/ failed with wecom: media download: dial tcp [64:ff9b:1::7f00:1]:9: i/o timeout, want the guard's refusal (any other error means a socket was opened toward an address that reaches IPv4 private space through the translator)
    media_guard_test.go:191: http://[64:ff9b:1::a00:1]/ failed with wecom: media download: dial tcp [64:ff9b:1::a00:1]:80: i/o timeout, want the guard's refusal (any other error means a socket was opened toward an address that reaches IPv4 private space through the translator)
    media_guard_test.go:203: a host resolving to 64:ff9b:1::a9fe:a9fe: err = wecom: media download: dial tcp [64:ff9b:1::a9fe:a9fe]:443: i/o timeout, want the guard's refusal
    media_guard_test.go:203: a host resolving to 64:ff9b:1::7f00:1: err = wecom: media download: dial tcp [64:ff9b:1::7f00:1]:443: i/o timeout, want the guard's refusal
--- FAIL: TestTheGuardRefusesBothNAT64Prefixes (50.01s)
=== RUN   TestARedirectIntoTheLocalUseNAT64PrefixIsRefused
    media_guard_test.go:228: err = wecom: media download: dial tcp [64:ff9b:1::a9fe:a9fe]:80: i/o timeout, want the guard's refusal on the second hop
--- FAIL: TestARedirectIntoTheLocalUseNAT64PrefixIsRefused (10.00s)
FAIL
```

`i/o timeout` against the metadata address, not a refusal. That is the socket the guard was supposed to prevent.

**The registry re-check found more, and one of them was the same hole.**

`2002::/16` — 6to4 — was open. A 6to4 address embeds the IPv4 in its second and third groups, so `2002:a9fe:a9fe::` is the metadata endpoint through a relay: same defect as NAT64, different prefix. Its IPv4 counterpart `192.88.99.0/24` (the deprecated relay anycast) was missing too. Both added. Teredo, `2001::/32`, carries an IPv4 endpoint the same way.

The rest were registry rows nothing covered, added while I was in there:

- **`2001::/23`** — the IETF protocol-assignments block, taken whole rather than as its dozen sub-entries. It contains Teredo, IPv6 benchmarking (`2001:2::/48`, the twin of the `198.18.0.0/15` we already had), AMT, AS112-v6, ORCHID/ORCHIDv2, DET, and the PCP / TURN / DNS-SD anycast addresses. Same call the v4 side already makes with `192.0.0.0/24`, and one prefix is easier to keep true than eight. `2001:db8::/32` sits outside the /23 and keeps its own entry.
- **`3fff::/20`** — documentation, RFC 9637. The newer companion to `2001:db8::/32`.
- **`5f00::/16`** — SRv6 SIDs, RFC 9602. Routing labels, not hosts.
- **`100:0:0:1::/64`** — the dummy prefix, discard-only, alongside the `100::/64` we already had.
- **`fec0::/10`** — site-local. Deprecated by RFC 3879 and delisted, so there is no registry row and no `netip` predicate for it, but networks numbered out of it before 2004 still route it internally.

**Deliberately not added**, recorded in the comment so nobody re-derives it: the AS112 delegations (`192.31.196.0/24`, `192.175.48.0/24`, `2620:4f:8000::/48`) and AMT's `192.52.193.0/24`. Those are globally routed unicast — special in who operates them, not in where they point.

One thing I checked rather than assumed: `fc00::/7` needs no entry. It is tempting to think `netip.Addr.IsPrivate` only catches `fd00::/8` and leaves the centrally-assigned half open, but on Go 1.26 it matches the full /7 (`ip.v6(0)&0xfe == 0xfc`), same as `net.IP.IsPrivate`.

The list is grouped now, because the two halves fail differently: space that is not the public internet, versus translation space, where the address *is* an IPv4 destination in an IPv6 costume. The allow-list table gained boundary cases as well as hits — `2001:200::`, `64:ff9c::`, `2402:4e00::` — since a prefix that swallows one group too many refuses real attachments, which fails just as loudly as letting one through.

## 2. The command source has no placeholders in it

`aibotMsgCallback.ownCommandSource()` builds it from the text and voice runs only, in composition order, and `channelMessageFromCallback` sets both `CommandText` and `SkipAgentRun` from it. `ownText` is untouched: the stored body keeps its placeholders, because the agent has to be able to see that a picture was attached and where, and the detached binder needs those positions. That split is what `engine/issue_command.go` already describes from the other end — `issueDescriptionFromCommandBody` says the body may carry placeholders while `CommandText` is the un-enriched command source.

Three details worth flagging:

- A voice run counts as words. WeCom transcribes on its side, so a spoken "/issue 登录坏了" inside a 图文混排 is a command like the typed one.
- A **standalone** photo answers with an empty command source rather than `"[Image]"`. The engine's own fallback then reads the body, which is what every adapter does, and `"[Image]"` is not a command in any of them.
- `traceInbound` now gets the resolved body instead of `mc.Text.Content`. That field is empty for every media and mixed callback, so tracing it printed `len=0` for exactly the messages an operator turns tracing on to look at.

Routing tests in `inbound_media_test.go` cover both orderings, an image between two typed runs, and the spoken command. The binding test drives the same message through `engine.Router`, the real `ChatSession` and a real Postgres, and asks what the sender would ask — was an issue created, under what title. Deriving the command from the resolved body instead:

```
=== RUN   TestIssueAfterAnImageIsStillFiled
    inbound_media_bind_db_test.go:546: issues created = 0, want 1 — the sender attached a screenshot, typed /issue 登录坏了, and nothing was filed and nothing said why
--- FAIL: TestIssueAfterAnImageIsStillFiled (0.08s)
FAIL
```

## 3. The rebase, and what it kept from both sides

Rebased onto `cd9f414`. Six files conflicted.

`ws_frame.go` is the load-bearing one, and your trap warning was exactly right: `main` derives `command` from `mc.Text.Content`, which is empty on a `mixed` callback because the runs live in `mc.Mixed.MsgItem`. A straight rebase leaves `CommandText` empty there, the shared fallback in `engine/session.go` substitutes the placeholder-bearing body, and item 2 comes back through the other door. One command source fixes both, so that is what it is:

```go
command := mc.ownCommandSource()
if chatType == channel.ChatTypeGroup {
    command = stripLeadingMentions(command, botDisplayName)
}
```

`main`'s `stripLeadingMentions`, the p2p gate on it, the single `engine.ParseIssueCommand` delegation in `isIssueCommand`, and `TestIssueCommandDetectionMatchesTheEngine` and both P2P-mention tests all survive untouched; the only edits in `ws_frame_test.go` are the call-site arity.

The rest:

- `wecom_channel.go` — kept `botDisplayName`, both `traceIn` calls and `traceInbound` from `main`, plus this PR's `ownText` / `!ok` receipt path and `CapAttachment`.
- `.env.example`, `docker-compose.selfhost.yml`, `server/cmd/server/router.go` — both env vars, not one. `MULTICA_WECOM_TRACE` with its `SetTrace` boot warning, and `MULTICA_WECOM_MEDIA_ALLOW_CIDRS` with `SetMediaAllowedPrefixes`. `main`'s `NEXT_PUBLIC_API_URL` note is intact.

`TestGroupMentionedIssueAfterAnImageIsStillFiled` is the honesty check you asked for — group `@bot` + mixed + `/issue`, through the Router and the database. It needs both halves. Removing `main`'s mention strip:

```
=== RUN   TestGroupMentionedIssueAfterAnImageIsStillFiled
    inbound_media_bind_db_test.go:575: issues created = 0, want 1 — an @-mentioned /issue with a screenshot in a group filed nothing
--- FAIL: TestGroupMentionedIssueAfterAnImageIsStillFiled (0.07s)
=== RUN   TestDispatchFrame_GroupMentionedMixedCommand
=== PAUSE TestDispatchFrame_GroupMentionedMixedCommand
=== CONT  TestDispatchFrame_GroupMentionedMixedCommand
    inbound_media_test.go:373: CommandText = "@Multica Bot /issue 登录坏了", want "/issue 登录坏了" — the placeholder and the addressing both have to come off
    inbound_media_test.go:377: engine.ParseIssueCommand("@Multica Bot /issue 登录坏了") declined — an @-mentioned /issue with a screenshot files nothing
--- FAIL: TestDispatchFrame_GroupMentionedMixedCommand (0.00s)
FAIL
```

Lose the placeholder removal instead and it fails the same way, from `"[Image]"`.

## Verified at 4900e17

- `go build ./...`, `go vet ./...` clean
- `gofmt -l` clean on every file this PR touches (the other unformatted files in the tree are that way on `main` too)
- `go test -race ./internal/integrations/wecom/ ./internal/integrations/channel/... -count=1` — green, against a real migrated Postgres, so the DB-backed binding tests ran rather than skipped
- `go test ./cmd/server -run Wecom -count=1` — green
- CI on `4900e17`: all nine checks pass (`backend`, `changes`, `frontend`, `frontend-build`, `frontend-test`, `windows-execenv`, `installer` × 3)

Still outstanding from your last note, and I agree it is the right next step: a pass through a WeCom sandbox for timeouts, object-storage permissions and an attachment near the size limit. Nothing in this round touches those, and they are not the sort of thing a test here can settle.

Bohan-J (changes requested):
Round 2. The three blockers from the last review are fixed — the RFC 8215 NAT64 prefix is in, the mixed-message command source no longer carries media placeholders, and the branch was rebased. The prefix list came back better than what I asked for: sourcing it from the IANA registries and splitting the rationale into "not the public internet" vs "translation space" is the right framing, and it's what makes the remaining issue easy to describe.

Two things before this can merge.

## Must fix

### 1. [Security] The operator allow-list can override translation space, re-opening the embedding bypass

`server/internal/integrations/wecom/media_guard.go` — `publicAddrOnly` applies `mediaAllowedPrefixes` to every entry in `reservedMediaPrefixes`:

```go
for _, p := range reservedMediaPrefixes {
    if p.Contains(a) {
        for _, allowed := range mediaAllowedPrefixes {
            if allowed.Contains(a) {
                return true
            }
        }
        return false
    }
}
```

For the "not the public internet" group that's correct — that group is exactly what the fake-IP-proxy escape hatch exists for. But the same override also covers the translation group (`64:ff9b::/96`, `64:ff9b:1::/48`, `2002::/16`), and for those prefixes the list is the *only* thing standing between the guard and an embedded IPv4 address. The early `IsLoopback()` / `IsPrivate()` / `IsLinkLocalUnicast()` checks never fire on them, because at that point they are IPv6 addresses — which is the point your own comment makes ("IPv4 destinations wearing an IPv6 costume").

So with `MULTICA_WECOM_MEDIA_ALLOW_CIDRS=::/0`:

- `64:ff9b:1::a9fe:a9fe` → `169.254.169.254` through a NAT64 translator
- `2002:7f00:1::1` → `127.0.0.1` through a 6to4 relay

both return `true` from `publicAddrOnly`. That's the cloud metadata endpoint, reachable from a URL WeCom hands us — the exact thing the address-embedding protection was just added to stop.

It also contradicts what the config promises operators. `.env.example`: *"Loopback and the private ranges are refused regardless and cannot be opened this way."* Right now they can be, if written in IPv6 translation form. `media_guard_allow_test.go` asserts that same promise but only with IPv4 literals, so the gap is invisible to the test.

`::/0` is a silly value to configure, granted — but the escape hatch is documented for operators who don't know their proxy's pool, and "I widened the CIDR until images loaded" is the realistic path into it. A narrower `MULTICA_WECOM_MEDIA_ALLOW_CIDRS=2002::/16` gets there too.

Suggested shape: split `reservedMediaPrefixes` into the two groups your comment already describes, and let the allow-list override only the first. Translation space becomes a hard block that no operator config can open — nothing legitimate lives there, so there's no deployment shape to accommodate. Then add `::/0` regression cases for a NAT64 and a 6to4 address, ideally through the real dialer rather than the predicate alone.

### 2. [Merge gate] The branch conflicts with `main` again — and this one is a semantic merge, not a textual one

`mergeable_state` is `dirty` as of now. #6599 (voice notes) landed on `main` this morning and touches `types.go`, `wecom_channel.go` and `ws_frame.go` — three files this PR also rewrites.

The overlap is inside the same function. `main` now has `bodyText()` answering for `"text"` and `"voice"` and returning false for `"mixed"`, and sources the command from it:

```go
command := body            // body == bodyText(), i.e. typed OR spoken
if chatType == channel.ChatTypeGroup {
    command = stripLeadingMentions(body, botDisplayName)
}
```

with the reasoning that a spoken `/issue` carries its words in `voice.content`, so reading the command off the typed field would leave it empty.

This PR needs `bodyText()` to *also* answer for `"mixed"` (rendered runs, placeholders included) for the message body, while the command source stays placeholder-free. Both changes rewrote the same decision from different directions, so please reconcile rather than take either side wholesale: after the merge, a spoken `/issue` should still parse, a group mention should still come off the front, and an image before `/issue` should still not swallow the command.

A test for "group @bot + mixed(voice run + image) + `/issue`" would be worth adding — it's the intersection of all three changes and nothing covers it today.

CI is green on `4900e171`, but that run predates #6599 landing, so it needs a re-run after the merge.

## Everything else

No change to my read from last round: intent recorded before upload, per-connection address validation with every redirect hop re-checked, signed URLs redacted from errors, streaming decryption to a temp file, single-attachment object keys, partial-failure notification, and a clean degrade to placeholder text when no object storage is configured. Coverage on those paths is solid.

## Verification

I can review this code, but I can't check it against a real WeCom bot — and per the maintenance note in `types.go`, neither can the rest of the Multica team. I'm asking this package's other code owner in a separate comment whether they can put it through a real device, since that's the part no amount of review here will settle.

Bohan-J:
@leroy-chen — the maintenance note in `types.go` names you and @seacen as this package's code owners, so flagging this one for you.

#6605 adds inbound media to the WeCom adapter: images, files and videos a user sends the bot get downloaded, decrypted and attached to the message instead of receiving a "text only" receipt — including 图文混排 messages where text and attachments are interleaved. It removes one of the two known limits the package doc calls out.

The code review side is covered. What we can't do is verify it against a real WeCom bot. The package doc is explicit that the Multica team doesn't use WeCom and can't check behavior against the real platform, and that's accurate here — so if you have a bot and the time, real-device confirmation would be genuinely valuable. Just as valuable is your read on whether the feature actually *feels* right in a normal conversation, which is the part review can't reach.

Entirely on your own schedule, and no obligation at all — this is a volunteer package and this is a favour, not a queue item. If you'd rather not, just say so and we'll land it on code review alone once the open items are fixed.

If you do pick it up, the things invisible from our side:

- An image, a file and a video each sent on its own — does the agent actually receive the attachment, and does the reply read sensibly?
- A 图文混排 message with text before an image and text after — is composition order preserved in what the agent sees?
- An attachment near the size limit, plus one where the download fails or the pre-signed URL expires. The intended behavior is a clear notice to the user while the rest of the message still gets answered.
- A deployment with no object storage configured — expected to degrade to placeholder text rather than fail.
- A group chat, where the bot is reached by @-mention, with an attachment attached.

`MULTICA_WECOM_TRACE=1` logs every frame this adapter reads and writes with tokens redacted, if you want to check a session against the server afterwards.

One scheduling note: two blocking items are still open — a security fix in the media address guard, and a merge conflict with #6599, which landed today. The branch will move before it merges, so it's worth waiting for the next push if you'd rather test once rather than twice.

seacen:
Both fixed, in the shape you suggested. `1539651e`.

## 1. The allow-list can no longer open translation space

`reservedMediaPrefixes` is now two lists, split along the two groups the comment already described:

- **`reservedMediaPrefixes`** — space that is not the public internet (CGNAT, the fake-IP range, TEST-NET, the IETF blocks, site-local, discard-only, documentation). `mediaAllowedPrefixes` still overrides this one, because the fake-IP proxy is a real deployment and an operator who knows their pool can declare it.
- **`translationMediaPrefixes`** — `64:ff9b::/96`, `64:ff9b:1::/48`, `2002::/16`. Checked first, before the allow-list is consulted at all, and nothing reopens it.

That ordering is the fix: the addresses in the second list are ones `IsLoopback` / `IsPrivate` / `IsLinkLocalUnicast` would have refused on sight, and the only reason they do not fire is the spelling. The split costs nothing — no COS object, no proxy pool and no deployment lives in translation space, so unlike the fake-IP case there is no shape to accommodate. `TestTheProxyEscapeHatchSurvivesTheTranslationBlock` pins that the hatch itself still works.

`.env.example` now says what the code does: loopback, the private ranges **and the IPv6 translation prefixes** are refused regardless.

### Reverse-verified

Folding the three prefixes back into `reservedMediaPrefixes` and deleting the early loop, with the new tests unchanged:

```
--- FAIL: TestAWideOpenAllowListStillCannotReachTranslationSpace/0.0.0.0/0,::/0
    64:ff9b:1::a9fe:a9fe became reachable under an allow-list of [0.0.0.0/0 ::/0] — it is 169.254.169.254 in an IPv6 spelling
    2002:7f00:1::1 became reachable under an allow-list of [0.0.0.0/0 ::/0] — it is 127.0.0.1 in an IPv6 spelling
    http://[64:ff9b:1::a9fe:a9fe]/latest/meta-data/ under an allow-list of [0.0.0.0/0 ::/0]: err = wecom: media download: dial tcp [64:ff9b:1::a9fe:a9fe]:80: i/o timeout, want the guard's refusal — anything else means a socket was opened
    http://[2002:7f00:1::1]:9/object under an allow-list of [0.0.0.0/0 ::/0]: err = wecom: media download: EOF, want the guard's refusal — anything else means a socket was opened
    refused = [], want exactly 64:ff9b:1::a9fe:a9fe — the refusal has to come from the policy, not from a failed connect
--- FAIL: TestAWideOpenAllowListStillCannotReachTranslationSpace/2002::/16
    2002:7f00:1::1 became reachable under an allow-list of [2002::/16] — it is 127.0.0.1 in an IPv6 spelling
--- FAIL: TestAWideOpenAllowListStillCannotReachTranslationSpace/64:ff9b:1::/48
    64:ff9b:1::a9fe:a9fe became reachable under an allow-list of [64:ff9b:1::/48] — it is 169.254.169.254 in an IPv6 spelling
```

You were right to want this through the dialer rather than the predicate. Look at the 6to4 line: not a timeout, **`EOF`**. The connect to `[2002:7f00:1::1]:9` succeeded on this machine — the kernel's 6to4 handling delivered it to `127.0.0.1` and the socket was opened and closed. A predicate-level test would have reported one boolean and told you nothing about that. `::/0` is the silly value; `2002::/16` and `64:ff9b:1::/48` are in there too, since "I widened the CIDR until images loaded" is the way in.

## 2. The merge

Rebased onto `90b3c2cc5`, which has #6599. `MERGEABLE`.

Reconciled rather than taking either side wholesale. There is one body resolver and one command source, and both now answer for voice:

- **`ownText`** gains `case "voice"` — the transcript, with an empty one still reporting false and taking the receipt path. `bodyText` is gone; it and `ownText` were the same function answering for different sets, and keeping both would have left two places to update the next time a kind is added.
- **`ownCommandSource`** gains `case "voice"` — the transcript again, because a spoken sentence is the sender's own words and a spoken `/issue` is a command. Placeholders stay out of it, and `mixedItem.words()` already read a voice run inside a 图文混排.
- **`channelMessageFromCallback`** takes the resolved body as a parameter (the caller has to know whether the message is routable before it gets here) and derives the command from `mc` — one `command :=` assignment, `ownCommandSource` then `stripLeadingMentions` for a group.

Two other things the merge touched: `channelMsgType`'s comments asserted a routing that has changed twice now, so they say what the code does rather than which change made it true; and `TestVoiceStaysOnTheReceiptPath` asserted that a standalone voice note is **not** read, which was true when it was written and stopped being true when #6599 landed — it is now `TestAVoiceNoteIsReadWhereverItArrives` and asserts the merged behavior.

All three of your invariants hold, each with a test that fails without it: a spoken `/issue` parses (`TestSpokenIssueCommandIsACommand`), a group mention comes off the front (`TestAnIssueCommandAfterAMentionIsStillACommand`, `TestAMultiWordBotNameReachesTheParser`), an image before `/issue` does not swallow it (`TestDispatchFrame_TheCommandSourceHasNoPlaceholdersInIt`).

### The intersection test you asked for

`TestDispatchFrame_GroupMentionedSpokenIssueWithAnImage` — someone in a group photographs the broken screen and says "@Multica Bot /issue 登录坏了". Two run orders: the mention inside the transcript, and the mention typed with the command spoken. The command has to survive three strips from three different changes. The same scenario runs one level down at `TestGroupMentionedSpokenIssueAfterAnImageIsStillFiled`, which puts the message through `engine.Router`, the real `ChatSession` and a real Postgres, then asks the database whether an issue exists.

### Reverse-verified, both directions

Taking **main's** decision wholesale (`command := text`, i.e. `command := body`):

```
--- FAIL: TestDispatchFrame_GroupMentionedSpokenIssueWithAnImage/the_mention_and_the_command_are_both_in_the_voice_run
    engine.ParseIssueCommand("[Image]\n@Multica Bot /issue 登录坏了") declined — a spoken, @-mentioned /issue with a screenshot files nothing and says nothing
--- FAIL: TestDispatchFrame_GroupMentionedSpokenIssueWithAnImage/the_mention_is_typed_and_the_command_is_spoken
    engine.ParseIssueCommand("[Image]\n/issue 登录坏了") declined — a spoken, @-mentioned /issue with a screenshot files nothing and says nothing
```

Taking **this branch's** decision wholesale (no voice case in `ownCommandSource`, `words()` not answering for a spoken run):

```
--- FAIL: TestDispatchFrame_GroupMentionedSpokenIssueWithAnImage/the_mention_and_the_command_are_both_in_the_voice_run
    stored body = "[Image]", want "[Image]\n@Multica Bot /issue 登录坏了"
    engine.ParseIssueCommand("") declined — a spoken, @-mentioned /issue with a screenshot files nothing and says nothing
--- FAIL: TestDispatchFrame_GroupMentionedSpokenIssueWithAnImage/the_mention_is_typed_and_the_command_is_spoken
    stored body = "@Multica Bot\n[Image]", want "@Multica Bot\n[Image]\n/issue 登录坏了"
    engine.ParseIssueCommand("") declined — a spoken, @-mentioned /issue with a screenshot files nothing and says nothing
```

Both directions through Postgres too, both reporting `issues created = 0, want 1`.

### The ordering in `processEvent`

Checked by hand rather than trusting the merge, since a clean auto-merge is not the same as a correct one — on this seam git has taken both sides of a `command :=` assignment before, on another branch, and a spoken `/issue` would have filed the issue and run the agent over it. `outbound.go` is untouched by this PR and the order is intact (`server/internal/integrations/wecom/outbound.go`):

| | line |
|---|---|
| binding lookup, content — reads | 101–114 |
| **origin gate** — task id, `GetAgentTask`, `TaskInputIsChannelIngested`, `if !deliver { return nil }` | **129–147** |
| installation load, `senders.get` — still reads | 148–173 |
| first and only write to WeCom, `sender.sendTextCtx` | 175 |

Everything between the gate and the write is a read.

## Verification

`go build ./...` and `go vet ./...` from `server/`, both clean, and `go test -race` green against a migrated Postgres — 436 tests in the wecom package, no skips, so the database-backed ones really ran.

seacen:
We ran this branch against a live WeCom tenant today. You mentioned you can't
exercise these paths on your side, so here is what actually arrives.

Six messages, one bot, one tenant. Every attachment downloaded, bound, and read
by the agent.

| msgtype | chat | stored |
|---|---|---|
| `image` | single | 287,667 B, image/jpeg |
| `file` | single | 21,440 B, .docx |
| `video` | single | 4,737,423 B, video/mp4 |
| `mixed` | single | text + image, 287,823 B |
| `mixed` | **group** | text + image, 357,767 B |
| `text` | single | `/issue …`, no media |

Three things worth recording.

**`mixed` carries an image into a group chat.** Doc 101463 marks the standalone
`image` type 仅支持单聊 and says nothing either way about `mixed`. We saw a group
message with `msgtype=mixed` whose image downloaded (357,767 bytes), bound, and
was read and answered in the group's own session. So the restriction is on the
standalone type, not on images as such. This isn't a contradiction of the docs —
they just don't cover the case, and now we know it works.

**WeCom appears to transcode video before handing it over.** The source file was
~20.8 MB; what the callback delivered was 4,737,423 bytes. One observation, not
a rule — but if it holds, inbound video will not approach the ceiling this branch
enforces.

**We found a filename defect on this branch and are still pinning it down.** A
file the user named `PC D&T Strategy 2026.docx` was stored as
`PC+D%26T+Strategy+2026.docx`. The callback frame carries only `{url, aeskey}`
with no name, so `Content-Disposition` is the only source, and
`mime.ParseMediaType` does not form-decode — so the header itself appears to
arrive form-urlencoded. We don't yet want to blanket-apply `QueryUnescape`,
because that would eat a legitimate `+` in a name. We're capturing the raw
header on the next live run and will fix it with the actual bytes in hand
rather than by inference.

Happy to run anything else you want covered while we have the tenant up.

seacen:
Screen capture from the live tenant, since neither of you can run one:

**[wecom-live-tenant-2026-08-10.mp4](https://github.com/seacen/multica/releases/download/wecom-demo-2026-08-10/wecom-demo.mp4)** (5.8 MB, 85s)

What it shows:

- a 图文混排 message going in (an image with "看下这个" on it) and the agent's answer coming back
- **the streaming bubble alive and rewriting itself in place** — at 4m56s it reads `正在处理 · Launching claude`, and the same bubble at 5m12s has been replaced by a later step naming the file it is reading. Same bubble, no new message in the thread.
- the answer written out underneath
- `/issue 登录坏了` → `✅ 已创建 UNI-12`

What it does **not** show, so nobody has to go looking:

- the moment the bubble is finally replaced by the answer — the capture is a scroll back through the thread, and the bubble at the bottom belongs to a task that was still running
- a degradation case (the guard's notice, or a failure)
- a group chat
- anything outbound — no agent-produced file is sent in this capture

We're redeploying with the current heads of all the open branches and will capture the missing cases in one pass. Shout if there's a specific frame you want.

seacen:
@Bohan-J — correcting my own note on that capture. I sampled it too sparsely
before describing it and undersold what it shows.

**The bubble rewrites itself in place through five distinct steps**, not the two
I described. Walking the capture at 3-second intervals:

| in the capture | the bubble reads |
|---|---|
| 0–12s | `正在处理 · Launching claude` — 已用时 4m 56s |
| ~15s | `· 正在执行 multica attachment download 0` |
| ~18s | `· 正在执行 multica attachment download 019fea86-…` — 已用时 5m 8s |
| ~21s | `· 正在读取 /Users/**/multica_workspaces_selfhost/…` |
| 24s→ | `image-6d2854ac…-0.jpg` — 已用时 5m 12s |

One message in the thread, rewritten five times. No new bubble per step, and the
elapsed counter advances inside the same message.

**And no completed turn in that thread carries a leftover bubble.** Scrolling
back through the whole conversation, every finished turn is just its answer —
which is the replacement working, observed across many turns rather than caught
in a single frame.

So my earlier "does not show the moment the bubble is replaced by the answer"
was misleading: the transition isn't caught mid-frame, but the before and after
are both there for every turn in the thread, and the in-place rewriting is
plainly visible.

Still genuinely not covered, and we'll capture these after the current round of
fixes is deployed:

- a degradation case — the guard's "still working" notice, or a failure
- anything outbound (nothing agent-produced is sent in this capture)

Same capture: https://github.com/seacen/multica/releases/download/wecom-demo-2026-08-10/wecom-demo.mp4

Bohan-J (approved):
✅ Approving and merging. Thank you for this, @seacen — it's a substantial piece of work and it got better at every round.

Worth recording what made it reviewable, because not many PRs do all of this:

**You answered the review with better than what was asked.** The first round flagged one missing NAT64 prefix. What came back was the list rebuilt from the IANA registries, split into "not the public internet" and translation space, with the reasoning for each group written down — and that split is exactly what made the second finding easy to state and easy to fix. The hard block now sits ahead of the allow-list, so no operator config can reopen it.

**The filename decode is the right shape for an ambiguous problem.** Deciding by round-trip self-consistency, rather than unescaping and hoping, is what lets `PC+D%26T+Strategy+2026.docx` come back correct while `C++ notes.docx` keeps its pluses. Running the decode before the base-name reduction — so `..%2F..%2Fetc%2Fpasswd` can't slip a traversal through — is the kind of ordering that's easy to get backwards.

**You captured real-device evidence none of us could produce, and then corrected your own description of it.** Resampling the capture and saying you'd undersold it is not a common instinct.

**And you reviewed your own commit and found the thing review would have missed.** The decode turning `%00` into a real NUL, which Postgres TEXT cannot store — meaning a file that downloaded and decrypted perfectly would die on the insert. That was self-caught, with the control-character strip and the tests to go with it.

The known-gaps list stays honest about what the header genuinely can't settle: a name whose only encoder-touched character is a `+` loses it, and a server whose unreserved set differs from Go's (`~` via PHP's `urlencode`, `*` via Java's `URLEncoder`) keeps its escapes. Both are pinned as expected values rather than left as surprises. `traceMediaHeaders` now records the header at full length, so the next live session can settle with bytes what this had to infer from one sample — worth a follow-up issue once you have that data, but nothing blocking.

Everything from the earlier rounds is closed: the allow-list bypass, the mixed-message command source, the #6599 merge with the voice/group-mention/`/issue` intersection covered, and the filename encoding. 9/9 CI green, merge state clean.

@leroy-chen — no action needed, this is landing. If anything looks wrong in real use once it's out, a note here is welcome.

PR #6608, opened 2026-08-07
feat(wecom): say something when somebody opens the bots chat for the first time (MUL-5908)

seacen:
> Rebased onto current `main`; the dependency this used to be stacked on has
> merged, so this is now a single commit and the whole diff is reviewable.

## What is broken

`ws_frame.go` has named the event since the original WeCom PR:

```go
eventEnterChat = "enter_chat"
```

Nothing ever handled it. The frame reached `dispatchFrame`'s event switch, missed
every case, and landed in the default arm:

```go
default:
    log.Debug("wecom: event", "type", ec.Event.EventType)
    return nil
```

WeCom pushes `enter_chat` the moment somebody opens the bot's conversation, and
expects `aibot_respond_welcome_msg` back. So the actual behaviour was: **a
brand-new user opens the bot and gets an empty window.** No statement of what
the bot is for, no hint that it needs their Multica account linked before it
will answer anything. The only way to discover that was to send it a message
and be told. Slack posts an App Home, Lark sends a welcome card; WeCom was the
one channel that said nothing.

WeCom gives roughly five seconds to answer, after which the chat window is
drawn and the user is already typing.

## What this does

`welcome.go` — `handleEnterChat`, `welcomeText`, `respondWelcome`,
`isWelcomeFrame`, `welcomeDeadline` (4s), `welcomeQueueDepth` (8) — plus the
read-loop wiring in `Connect` and the boot wiring in `router.go`.

* **Unlinked user** → greeting + a bind link, minted the same way and pointing
  at the same page as the `needs_binding` prompt in `replier.go`. The first
  thing they see is also the thing that makes the bot work.
* **Linked user** → plain greeting, no link, no token minted. Offering a bind
  link to somebody already bound reads as the bot having lost their account.
* **Can't tell** (lookup failed, no binding surface wired, mint failed) →
  plain greeting. Degrading to silence would make the bot look broken over a
  feature the deployment may simply not have.
* The reply echoes the frame's own `req_id`. That is the only address a
  welcome has; a fresh id is refused.

Copy is hardcoded Chinese, matching how every other user-visible string in this
adapter already works (`agentOfflineText`, the binding prompt, the text-only
receipt). This package has no locale layer and WeCom deployments are China-only.

## The deliberate asymmetry — please read this before flagging it

Two queues sit within a few lines of each other in `Connect` with **opposite
full-queue policies**. This is intentional, and it is the thing a mechanical
port would flatten:

| | depth | when full |
|---|---|---|
| inbound callbacks (`callbackQueueDepth`, on the base branch) | 64 | **blocks** the read loop |
| welcome (`welcomeQueueDepth`, this PR) | 8 | **drops** the frame |

Both are correct, for opposite reasons:

* A **dropped callback** is a message the user sent that nobody ever answers,
  with nothing to say so. Blocking is backpressure — at worst WeCom ends the
  connection and the Supervisor reconnects, which is recoverable. Silent loss
  is not.
* A **late welcome** is worse than no welcome. It lands in a chat the user has
  already started typing in. And a greeting is addressed by a `req_id`, which
  means nothing on the next connection — so there is no version of it worth
  holding. A late welcome is void, not late. Meanwhile, blocking on it would
  hold the read loop, which is the sole deliverer of the acks every other
  outbound write is parked on.

Same reason enter_chat gets its own worker rather than sharing the callback
worker: one message callback ahead of it in a single queue can take longer than
five seconds on its own, which would reliably turn a greeting into nothing.

`TestAFullWelcomeQueueDoesNotStallTheReadLoop` is the guard on this. Flip the
`default:` arm to a blocking send and it fails.

## Group chats get no greeting — also deliberate

`handleEnterChat` returns early on anything that is not `chattype == "single"`.
Two reasons, either sufficient:

1. Greeting a room is a different act from greeting a person. The event names
   whoever just walked in, so a welcome in a group is an announcement *about*
   that person to an audience. That is not what opening a chat asked for.
2. The bind link is a bearer credential. `binding.Redeem` only checks that the
   redeemer belongs to the token's workspace, and the bind page redeems on load
   as whoever is signed in — so posted into a room, the first colleague to
   click owns the account, and every later message from the person it was
   minted for resolves to the hijacker. Same reasoning `sendBindingPrompt`
   already spells out in `replier.go`, and easier to obey here: a room nobody
   has spoken in yet has nothing to be told.

## Reverse-verified: every test fails on the old code

Each test was checked by reverting *only* the line it claims to defend and
confirming it fails. Real output:

```
=== TestOpeningTheChatIsAnsweredOverTheLiveConnection
    revert: read loop diverts enter_chat to the welcome worker
    --- FAIL: TestOpeningTheChatIsAnsweredOverTheLiveConnection (3.00s)
        welcome_test.go:462: enter_chat arrived on a live connection and the bot
        said nothing — a first-time user is looking at an empty window

=== TestOpeningTheChatUnboundHandsOverTheLink
    revert: an unbound user is offered a bind link
    --- FAIL: TestOpeningTheChatUnboundHandsOverTheLink (0.00s)
        welcome_test.go:175: the greeting carries no bind link: "👋 你好，我是
        Multica 智能助手。有事直接发消息给我，或者用 “/issue 标题” 建一条任务。…"

=== TestOpeningTheChatBoundGreetsWithoutALink
    revert: a bound user is not offered a link and no token is minted
    --- FAIL: TestOpeningTheChatBoundGreetsWithoutALink (0.00s)
        welcome_test.go:194: a bind link was offered to somebody already bound:
        "…请先绑定你的 Multica 账号…/wecom/bind?token=s3cret…"

=== TestAGroupGetsNoGreetingAndNoToken
    revert: group chats are skipped
    --- FAIL: TestAGroupGetsNoGreetingAndNoToken (0.00s)
        welcome_test.go:218: a room was greeted: "…/wecom/bind?token=s3cret…"

=== TestNoBindingSurfaceStillGreets
    revert: no binding surface still greets instead of going silent
    --- FAIL: TestNoBindingSurfaceStillGreets (0.00s)
        welcome_test.go:236: said "", want the plain greeting

=== TestAnUnreadableBindingDoesNotOfferALink
    revert: an unreadable binding row does not become a bind link
    --- FAIL: TestAnUnreadableBindingDoesNotOfferALink (0.00s)
        welcome_test.go:252: a link was offered on a failed lookup:
        "…/wecom/bind?token=s3cret…"

=== TestTheGreetingEchoesTheFramesReqID
    revert: the welcome echoes the frame's own req_id
    --- FAIL: TestTheGreetingEchoesTheFramesReqID (0.00s)
        welcome_test.go:276: req_id = "ed5fc87b68b8306a", want the frame's own
        — WeCom refuses anything else

=== TestOnlyEnterChatRoutesToTheWelcomeWorker
    revert: only enter_chat is treated as a welcome
    --- FAIL: TestOnlyEnterChatRoutesToTheWelcomeWorker (0.00s)
        welcome_test.go:296: disconnected_event was routed to the welcome
        worker, which never escalates and never reaches the engine

=== TestAFullWelcomeQueueDoesNotStallTheReadLoop
    revert: a full welcome queue DROPS instead of blocking the read loop
    --- FAIL: TestAFullWelcomeQueueDoesNotStallTheReadLoop (3.00s)
        welcome_test.go:510: a burst of enter_chat frames stalled the read loop:
        the message the user typed behind them was never read
```

One test initially survived its revert and was rewritten rather than kept:
`TestOpeningTheChatBoundGreetsWithoutALink`. Deleting the whole `case err == nil`
arm left `case !errors.Is(err, pgx.ErrNoRows)` catching a nil error too, so the
bound path stayed correct by accident and the test passed on the reverted code.
The revert it is now checked against deletes only the `return` under
`case err == nil:`, which drops through into the mint — and the test fails, as it
must.

`TestOpeningTheChatIsAnsweredOverTheLiveConnection` drives the whole thing
through `Connect` against a socket double that acks the subscribe handshake and
answers written frames with real ack frames over the wire — so it exercises the
read loop, the welcome worker, and the ack round-trip together, not just the
handler in isolation.

## Not ported, on purpose

* **The durable outbound queue and the rate gate** — superseded by #6581.
* **`respondText`** (the passive-reply courtesy notice) shipped in the same file
  on the source branch, but it depends on a `claimPassiveReply` turn-claim that
  does not exist here, and nothing on this base calls it. Adding it would be
  untestable dead code. It belongs with the streaming-reply work.
* **The welcome metrics counters** (`RecordWelcomeSent` / `Skipped` / `Failed`).
  The `Metrics` sink has since landed on `main`, so the wiring for them exists
  now — but the interface carries connection and callback-queue counters only,
  and widening it is a decision about what the WeCom dashboard is for rather
  than part of this greeting. Log lines here, still. Say the word and I will
  add them in a follow-up.
* **The locale/copy pack** — no such layer here; hardcoded Chinese, as above.
* **`token.Reused`** (the "a live link is already with them" branch) — that
  field arrives with the mint-throttle PR. Nothing to port yet.

## Incidental

* `wsSender.request` is split into `request` / `requestWithID`. Same code, same
  behaviour; a welcome cannot choose its own `req_id`, it must echo the frame's.
* `NewOutboundReplier`'s inline bind-path defaulting moves to a shared
  `normalizeBindingPath`, so the greeting's link and the `needs_binding`
  prompt's link cannot drift onto two different bind pages.

## Checks

```
go build ./...          clean
go vet ./...            clean
gofmt -l                clean
go test ./internal/integrations/wecom/                  ok  (168 tests)
go test -race -count=3 ./internal/integrations/wecom/   ok
go test -race ./cmd/server/...                          ok
```

### DIFF
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -695,11 +695,21 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 					Fallback: "企业微信会话",
 				})
 
+				// Welcome/Binding/AppURL are the enter_chat greeting: WeCom
+				// pushes that event the moment somebody opens the bot's chat
+				// and gives us about five seconds to answer, so the greeting
+				// is written by the connection itself rather than by the
+				// OutboundReplier (which only ever runs after a message).
+				// Same binding service and app URL the replier gets, so both
+				// hand out the same bind link.
 				wecom.RegisterWecom(channelRegistry, wecom.ChannelDeps{
 					Credentials: credsResolver,
 					Senders:     wecomSenders,
 					Metrics:     wecomMetricsOrNil(opts.WecomMetrics),
 					Logger:      slog.Default(),
+					Welcome:     queries,
+					Binding:     wecomBinding,
+					AppURL:      appURLFromEnv(),
 				})
 				// Inbound media: a callback carries a pre-signed COS url and
 				// a per-url key, so the resolver needs no WeCom credential —
--- a/server/internal/integrations/wecom/replier.go
+++ b/server/internal/integrations/wecom/replier.go
@@ -73,17 +73,12 @@ func NewOutboundReplier(cfg OutboundReplierConfig) *OutboundReplier {
 	if logger == nil {
 		logger = slog.Default()
 	}
-	bindingPath := cfg.BindingPath
-	if bindingPath == "" {
-		bindingPath = "/wecom/bind"
-	}
-	if !strings.HasPrefix(bindingPath, "/") {
-		bindingPath = "/" + bindingPath
-	}
 	r := &OutboundReplier{
-		senders:     cfg.Senders,
-		appURL:      strings.TrimRight(cfg.AppURL, "/"),
-		bindingPath: bindingPath,
+		senders: cfg.Senders,
+		appURL:  strings.TrimRight(cfg.AppURL, "/"),
+		// Shared with the enter_chat greeting (welcome.go) so the two cannot
+		// drift onto different bind pages.
+		bindingPath: normalizeBindingPath(cfg.BindingPath),
 		logger:      logger,
 	}
 	// Assign through the interface only when non-nil: a nil *BindingTokenService
--- a/server/internal/integrations/wecom/wecom_channel.go
+++ b/server/internal/integrations/wecom/wecom_channel.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/url"
+	"strings"
 	"time"
 
 	cryptorand "crypto/rand"
@@ -93,6 +94,15 @@ type wecomChannel struct {
 	// through mx(), which substitutes the no-op sink for a channel built
 	// without one.
 	metrics Metrics
+
+	// welcome + binding + appURL + bindPath are the enter_chat greeting
+	// (welcome.go). Any of them nil or empty degrades to the greeting without
+	// a bind link rather than to silence — a deployment with no binding
+	// surface still wants the bot to say what it is.
+	welcome  welcomeLookup
+	binding  binder
+	appURL   string
+	bindPath string
 }
 
 var _ channel.Channel = (*wecomChannel)(nil)
@@ -232,6 +242,32 @@ func (c *wecomChannel) Connect(ctx context.Context) (err error) {
 	// so. That only holds while somebody is still receiving, so the read
 	// loop's send also watches cbDone — see the send site below.
 	callbacks := make(chan frameEnvelope, callbackQueueDepth)
+
+	// enter_chat gets its OWN worker, separate from the callback worker above.
+	// The greeting's window is short (see welcomeDeadline) and a single
+	// message callback ahead of it in one queue can take longer than the whole
+	// window on its own — sharing would reliably turn a greeting into nothing
+	// at all. It is also the one inbound frame whose handling has no error
+	// worth escalating, so this worker never reports failure.
+	//
+	// welcomeCtx is cancelled before the queue is drained, so a connection on
+	// its way out discards the greetings it can no longer address rather than
+	// holding Connect open waiting for acks that will never arrive.
+	welcomeCtx, welcomeCancel := context.WithCancel(ctx)
+	welcomes := make(chan frameEnvelope, welcomeQueueDepth)
+	welcomeDone := make(chan struct{})
+	go func() {
+		defer close(welcomeDone)
+		for env := range welcomes {
+			c.handleEnterChat(welcomeCtx, env, sender, log)
+		}
+	}()
+	defer func() {
+		welcomeCancel()
+		close(welcomes)
+		<-welcomeDone
+	}()
+
 	cbDone := make(chan struct{})
 	var cbErr error
 	go func() {
@@ -303,7 +339,33 @@ func (c *wecomChannel) Connect(ctx context.Context) (err error) {
 			log.Warn("wecom: bad frame envelope", "error", err, "size", len(payload))
 			continue
 		}
+		// Traced before the welcome intercept below, for the same reason the
+		// subscribe path traces before its req_id filter: a greeting that was
+		// skipped for a full queue is exactly what an operator turns tracing
+		// on to see, and a frame diverted before traceIn is a frame that never
+		// appears in the trace at all.
 		traceIn(log, env)
+		if isWelcomeFrame(env) {
+			// A full welcome queue DROPS. This is the OPPOSITE of the
+			// callback queue further down this loop, which blocks, and both
+			// are right for opposite reasons:
+			//
+			//   - A dropped callback is a question the user asked that nobody
+			//     ever answers, with nothing to say so. Blocking is
+			//     backpressure; at worst WeCom ends the connection and the
+			//     Supervisor reconnects, which is recoverable.
+			//   - A greeting has only its short window. One that arrives after
+			//     that lands in a chat the user has already started typing in,
+			//     which is worse than no greeting — and while it waited it
+			//     would be holding the read loop that carries the acks every
+			//     other write is parked on. A late welcome is void, not late.
+			select {
+			case welcomes <- env:
+			default:
+				log.Warn("wecom: welcome queue full, greeting skipped")
+			}
+			continue
+		}
 		switch env.Cmd {
 		case cmdMsgCallback, cmdEventCallback:
 			select {
@@ -489,6 +551,11 @@ func (c *wecomChannel) dispatchFrame(ctx context.Context, env frameEnvelope, sen
 			// one — the last writer wins).
 			return errors.New("wecom: received disconnected_event (superseded)")
 		default:
+			// enter_chat does not reach here: the read loop diverts it to the
+			// welcome worker (welcome.go) before dispatch, because it is the
+			// only inbound frame with a seconds-long deadline. What is left is
+			// template_card_event and feedback_event, neither of which we act
+			// on yet.
 			log.Debug("wecom: event", "type", ec.Event.EventType)
 			return nil
 		}
@@ -609,6 +676,16 @@ type ChannelDeps struct {
 
 	// WSURL overrides DefaultWSURL. Same test-only intent as Dialer.
 	WSURL string
+
+	// Welcome + Binding + AppURL + BindingPath drive the enter_chat greeting
+	// (welcome.go). Boot passes the same binding service and app URL the
+	// OutboundReplier already gets, so the greeting's bind link and the
+	// needs_binding prompt's are the same link. All optional: without them the
+	// bot still greets, just without offering a link it cannot mint.
+	Welcome     welcomeLookup
+	Binding     *BindingTokenService
+	AppURL      string
+	BindingPath string
 }
 
 // RegisterWecom registers the per-installation wecom smart-bot Factory so
@@ -643,7 +720,7 @@ func newWecomFactory(deps ChannelDeps) channel.Factory {
 		if err != nil {
 			return nil, fmt.Errorf("wecom: decrypt secret: %w", err)
 		}
-		return &wecomChannel{
+		ch := &wecomChannel{
 			installationID: cfg.ID,
 			botID:          creds.BotID,
 			secret:         creds.Secret,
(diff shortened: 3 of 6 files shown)


seacen:
Rebased onto main. One conflict, and it was an ordering question rather than a textual one.

**The conflict — the read loop in `Connect`.** #6602 added `traceIn(log, env)` right where
this branch adds the `isWelcomeFrame` intercept. Both sides wanted the same line. The
intercept `continue`s, so putting it first would mean an enter_chat frame never appears in
the trace at all — and a greeting dropped for a full queue is exactly what an operator turns
tracing on to see. `traceIn` goes first, then the intercept, which is the same order #6602
already uses for the subscribe path (traced before the req_id filter). Noted in a comment so
it does not get swapped back.

**Copy pack was unpinned.** Setting each constant to an obviously wrong value and re-running
the suite: `welcomeBoundText`, `welcomeUnboundPrefix`, `welcomeUnboundSuffix` and
`cmdRespondWelcome` could all be replaced with `"OBVIOUSLY WRONG"` with every test still
green, because every assertion compares what was said against the same constant that
produced it.

`cmdRespondWelcome` is the serious one: it is WeCom's command name, not ours. Get it wrong
and no greeting is ever delivered, while the whole file keeps passing.

Added two tests. One pins `cmdRespondWelcome` to its literal. The other pins what the two
greetings have to be rather than their exact prose — non-empty, different from each other,
the bound one carrying no link and naming the one command the bot takes, the unbound one
asking for the account link and stating how long it lasts. Re-probed afterwards: all four
now fail when broken. `defaultBindingPath` was already covered.

`welcomeDeadline` and `welcomeQueueDepth` are still unpinned, deliberately — they are tuning
values, and a test asserting them relative to themselves would prove nothing.

**The five-second figure had no source.** The comments stated as fact that WeCom gives the
greeting about five seconds, and `welcomeDeadline = 4s` was derived from it. The only
related number in the tree is the 5s window documented for `aibot_respond_msg`, a different
command; WeCom's long-connection docs do not state one for `enter_chat`, and we have not
measured it. Behaviour unchanged, but it is now labelled an assumption at `welcomeDeadline`,
with what would settle it (send a late `respond_welcome` at a real bot and read the errcode)
and what to do either way. The other three sites now point at that one place instead of
repeating the number.

Also fixed "the callback queue ten lines below" — it is twenty-two lines below and a
line-distance in a comment goes stale on the next edit anyway.

`go build`, `go vet`, `gofmt` clean; `go test -race` green on
`./internal/integrations/wecom/...` and `./cmd/server/...` against a migrated Postgres.

seacen:
Rebased onto current main. One conflict, and one thing that landing voice notes made false here.

**The conflict — `wecom_channel.go`, four regions.** The adapter-health work touched the same
places this branch does: the `wecomChannel` struct, the read loop's callback-queue send,
`ChannelDeps`, and the factory literal. Each is "main added a field, this branch added a
different one", so both sides are kept in all four and nothing on either side was dropped.

The factory is the one worth reading twice. Main added `metrics: orNopMetrics(deps.Metrics)`
to the returned literal, while this branch had already turned that literal into
`ch := &wecomChannel{...}` followed by a guarded `ch.binding = deps.Binding`, so that a nil
`*BindingTokenService` never becomes a non-nil interface holding a typed nil. Either side
taken whole loses the other. It now keeps the metrics field inside the literal, the binding
guard after it, and a single `return ch, nil`.

Read-loop order is unchanged: `traceIn` → welcome intercept → `switch env.Cmd`.

**One consequence of the two meeting, flagging it rather than deciding it.** `enter_chat`
arrives as `aibot_event_callback`, so on main today it goes down the callback arm and counts
as `RecordCallbackQueued`. This branch diverts it before the switch, so it stops counting.
That reads correct to me: the counter measures pressure on the callback worker, and a
greeting no longer reaches that worker, so counting it would overstate a queue it is not in.
It does mean the number moves slightly when this merges. If you would rather the greeting be
counted somewhere, say so and I will add a counter for it.

**Voice notes made the greeting wrong.** `welcomeBoundText` ended in （目前只能处理文字消息）—
"currently only handles text messages". That stopped being true when voice routing landed:
`bodyText` routes a voice note whose transcript is non-empty, including a spoken `/issue`,
which files a task. The greeting is the only place the bot volunteers what it takes, so every
linked user was reading that a working feature does not work. It now says
（目前支持文字和语音消息）.

The reason nothing caught it seemed worth fixing alongside the string. A capability arrives
by widening `bodyText` in `ws_frame.go`, while the sentence describing it lives in
`welcome.go`, so nothing about that change draws the eye to this one.
`TestTheBoundGreetingNamesEveryKindTheBotActuallyRoutes` now asserts the relation directly —
whatever `bodyText` routes, the greeting has to name — with an anti-vacuity check so it
cannot pass by asserting nothing. Widen `bodyText` again and it fails until the copy follows.

**The four constant pins survived the rebase**, and I re-probed rather than assuming.
Replacing each with `"OBVIOUSLY WRONG"`:

| constant | goes red in |
|---|---|
| `cmdRespondWelcome` | `TestRespondWelcomeUsesThePlatformsCommandName` |
| `welcomeBoundText` | `TestTheTwoGreetingsSayDifferentThings`, and now the routes test |
| `welcomeUnboundPrefix` | `TestTheTwoGreetingsSayDifferentThings` |
| `welcomeUnboundSuffix` | `TestTheTwoGreetingsSayDifferentThings` |

The new test was probed the same way: restoring the old wording turns it red, and so does
dropping just 语音 from the new wording.

**Nothing else went stale.** I checked the claims that reach outside this branch by grepping
for what each asserts rather than reading around the ones I already knew about: the 5s figure
for `aibot_respond_msg`, `sendBindingPrompt`'s bearer-credential reasoning, the typed-nil trap
`NewOutboundReplier` documents, the subscribe path tracing before its req_id filter, and
"what is left is `template_card_event` and `feedback_event`". All still hold, as does
（链接 15 分钟内有效）against `BindingTokenTTL`. The only other copy of "只能处理文字消息" in the
tree is the text-only receipt in `dispatchFrame`, which is main's own and fires only when the
bot is actually refusing something.

`go build`, `go vet`, `gofmt` clean; `go test -race` green on `./internal/integrations/wecom/...`
and `./cmd/server/...` against a migrated Postgres.

seacen:
Two substantive changes since this was filed, both worth knowing before you read
the diff.

## A defect we found ourselves: the greeting handed out a link that binds nobody

`Mint` is throttled — when a live unconsumed token is already with this user it
returns `Reused` with `Raw` empty, because only the hash was ever stored.
`replier.go:179` answers that by pointing at the link already in the chat.
`welcome.go` was not looking at `Reused` at all, so it built the URL anyway and
produced `?token=` with nothing after it.

It is not a corner case. The bot answers an unbound user in a group, replier
mints and posts the link into their 1:1, and the user opens that 1:1 to click
it — which is their first entry of the day, so `enter_chat` fires seconds after
the mint, inside the throttle window. The one person the greeting exists to help
gets the broken link.

Fixed, with a test that reverse-verifies: backing the change out produces
`https://…/wecom/bind?token=` in the assertion output.

## The greeting was making a claim that stopped being true

#6605 merged while this was open. It renamed `bodyText` to `ownText` and widened
it to route photos, files and videos — and the greeting still said text and
voice were what worked.

`TestTheBoundGreetingNamesEveryKindTheBotActuallyRoutes` caught it, which is
what it was written for: a capability arrives by widening one function, and the
sentence describing it lives in a different file. Three rows added, the sentence
widened to name photos, files and videos, and the stale-wording half of the test
now lists every narrowing this copy has been through rather than only the first.

## Two comments corrected against the docs

`welcomeDeadline` carried a note calling its 4s an "ASSUMPTION, not a documented
figure", on the grounds that the five-second window is documented for
`aibot_respond_msg` and not for `enter_chat`. That is wrong: doc 101463 states
it for the event callback itself — reply within five seconds or the greeting
cannot be sent — and WecomTeam/aibot-node-sdk says the same of
`aibot_respond_welcome_msg`. Now cited rather than hedged.

The file header said WeCom pushes `enter_chat` "the moment a user opens the bot's
conversation". The doc says a user's FIRST entry into the 1:1 that DAY. That
matters here: the greeting recurs daily for anyone who keeps opening the chat,
which is exactly why the unbound branch must not mint a fresh link each time.

Rebased onto current main. Nine checks green.

multica-eve:
Thanks for the thoughtful work and the detailed tests. After product review, we've decided not to add an automatic welcome message when someone opens the bot chat.

The event recurs on the user's first entry each day, so this would repeatedly send onboarding copy even to already-linked users. We prefer to keep the chat quiet until the user sends a message; the existing first-message binding prompt already covers unlinked users at that point.

This is a product-direction decision rather than a request for further implementation changes, so we're closing the PR as not planned. We appreciate the investigation and care that went into it.

PR #6611, opened 2026-08-07
MUL-5910: fix(chat): a predecessor's reply must not seal the next message out of its own batch

seacen:
On a self-hosted tenant a user sent a WeCom message at 18:31:17.7 while the
previous agent turn was finishing. The engine enqueued a task for it at
18:31:20.7 and cancelled the task 27ms later:

    ERR chat claim: task-owned direct task has no user input; cancelling
        task_id=bdc0947e chat_session_id=2fccd3c9 chat_input_task_id=bdc0947e

The database afterwards:

    chat_message "挺好的 正常了" (18:31:17.726997)  -> task_id = NULL
    every other user message in that session        -> task_id set
    select count(*) from chat_message
      where task_id='bdc0947e' and role='user'      -> 0

The user's message was accepted, a task appeared and instantly vanished, and
no answer ever came. On WeCom the typing bubble also spun forever, because the
indicator only unsubscribed on task:failed, not task:cancelled.

Mechanism

Channel messages (Slack / Lark / WeCom) are written durably and unowned on the
ACK path; the 3s silence debouncer later fires one flush per session, and
EnqueueChatTask creates the task, stamps chat_input_task_id = id, and seals the
waiting messages into it with LinkUnownedChannelChatMessagesToTask. That seal
skipped any unowned user row that had a non-user row after it. The rule was
carried over from the pre-ownership trailing-message selector, and it does not
survive the debounce window: a predecessor completing inside that window writes
its reply AFTER the waiting message, so the reply to an already-sealed batch
became the boundary for a message it never saw.

    18:31:04.6  user "测试下"                     unowned
    18:31:07.6  task 6a4ae7f8 created, seals it
    18:31:17.7  user "挺好的 正常了"               unowned, 3s window arms
    18:31:20.2  6a4ae7f8 completes, writes its reply   <- newer than the message
    18:31:20.7  window expires, task bdc0947e created, seal matches ZERO rows
    18:31:20.7  claim finds no input, cancels bdc0947e

Every timestamp above is from the tenant's own tables. The claim guard at
internal/handler/daemon.go:2347 is right to fail closed on an empty input
batch; the batch should never have been empty. Its comment blamed a send path
that does not commit message and task together — SendDirectChatMessage does,
and the channel path does too. Neither was the problem. The seal predicate was.

Fix

The boundary is "already answered", not "something newer exists". A reply can
only be a reply TO a message if the turn that wrote it started AFTER the
message arrived. A turn already running when the message landed sealed its own
input before the message existed, so its reply — however late it lands —
answers an earlier batch and must not move the boundary. The seal now qualifies
each candidate boundary row by its turn's creation time, resolved through
COALESCE(chat_input_task_id, id) so an auto-retry clone is judged by the root
batch it inherited rather than by its own fresh id.

The check is kept, not dropped, and deliberately stays conservative: a non-user
row whose task is NULL or gone still counts as a boundary, so pre-ownership
transcripts are never swept into a new batch, and a message already stranded by
this bug is not resurrected into a turn minutes later — a following turn that
ran past it holds the boundary.

Tests

  internal/handler/chat_channel_debounce_boundary_test.go drives the whole
  failure: message in, claim, run, second message while it runs, predecessor
  completes, window expires, claim. Before the fix:

    --- FAIL: TestChannelChat_ReplyLandingInsideDebounceWindowStillAnswersTheNextMessage
        chat_channel_debounce_boundary_test.go:134: claim = 500 {"error":"chat task has no user input"}
             (task <nil>); task 2a67492b-cd61-45e6-afd6-156079971152 owns 0 user messages and is now "cancelled". The user's message was accepted and never answered.

  internal/service/task_channel_seal_race_test.go pins the same boundary at the
  seal, twice — the sequential ordering the tenant hit, and the tighter window
  where the reply commits on another connection after the enqueue transaction
  has begun (same raceInjectTxStarter the media-deferral race test uses), so
  the seal statement's READ COMMITTED snapshot is the first to see it. Before
  the fix:

    --- FAIL: TestEnqueueChatTaskSealsMessageStrandedByPredecessorReply
        task_channel_seal_race_test.go:78: waiting user message task_id = "", want the new turn "34b62bad-955c-46f0-99e9-2194999d2c10" — a predecessor reply that landed after the message must not seal it out
    --- FAIL: TestEnqueueChatTaskSealsMessageWhenReplyCommitsDuringEnqueue
        task_channel_seal_race_test.go:142: waiting user message task_id = "", want the new turn "989a83e4-e945-4b36-a5b7-69720449e08c"

  The third, TestEnqueueChatTaskLeavesAlreadyPassedMessagesUnowned, passes both
  before and after by design: it is the guard against over-correcting, and only
  a fix that swept too much could break it.

The two transactions cannot interleave more finely than this: CompleteTask
takes chat_session FOR UPDATE and the task INSERT takes the FK's FOR KEY SHARE
on the same row. "Predecessor commits, then the flush runs" is the reachable
interleaving, and it is the one the tenant hit.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MKFf7wYM9Q2iaPtoFhG8MV

---

Gates, on this branch:

```
cd server
go build ./... && go vet ./internal/...
DATABASE_URL=... go test ./internal/handler/ ./internal/service/ -count=1
ok  github.com/multica-ai/multica/server/internal/handler  21.319s
ok  github.com/multica-ai/multica/server/internal/service   3.298s
```

`./internal/integrations/...` and `./pkg/...` also pass, except
`ghsnapshot/TestInFlightOldHeadKeepsTrailingRefresh`, which fails identically
on pristine `upstream/main` (verified) and is unrelated to this change.

The WeCom typing-bubble half of the symptom (the indicator only unsubscribed
from `task:failed`, never `task:cancelled`) is fixed separately and is not part
of this PR.

### DIFF
--- a/server/internal/handler/chat_channel_debounce_boundary_test.go
+++ b/server/internal/handler/chat_channel_debounce_boundary_test.go
@@ -1,0 +1,146 @@
+package handler
+
+import (
+	"context"
+	"encoding/json"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-chi/chi/v5"
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+// appendChannelUserMessage writes an inbound channel message the way
+// ChatSession.AppendUserMessage does: durable, unowned, and stamped with the
+// immutable channel_ingested provenance. The 3s debounced flush is what later
+// creates the task and seals this row into its input batch.
+func appendChannelUserMessage(t *testing.T, ctx context.Context, sessionID, body string) string {
+	t.Helper()
+	var id string
+	if err := testPool.QueryRow(ctx, `
+		INSERT INTO chat_message (chat_session_id, role, content, channel_ingested)
+		VALUES ($1, 'user', $2, TRUE)
+		RETURNING id
+	`, sessionID, body).Scan(&id); err != nil {
+		t.Fatalf("append channel user message %q: %v", body, err)
+	}
+	return id
+}
+
+// flushChannelChatRun is the debounced run trigger: it creates the task and
+// seals the session's waiting input batch into it, exactly as Router's batcher
+// does when the silence window expires.
+func flushChannelChatRun(t *testing.T, ctx context.Context, sessionID string) db.AgentTaskQueue {
+	t.Helper()
+	sess, err := testHandler.Queries.GetChatSession(ctx, parseUUID(sessionID))
+	if err != nil {
+		t.Fatalf("load chat session: %v", err)
+	}
+	task, err := testHandler.TaskService.EnqueueChatTask(ctx, sess, parseUUID(testUserID), false)
+	if err != nil {
+		t.Fatalf("EnqueueChatTask: %v", err)
+	}
+	t.Cleanup(func() { testPool.Exec(context.Background(), `DELETE FROM agent_task_queue WHERE id = $1`, task.ID) })
+	return task
+}
+
+// claimResult is a claim attempt that does not assume success, so a failing
+// claim reports the status and body the daemon actually received.
+type claimResult struct {
+	code int
+	body string
+	task *claimRuntimeGuardTask
+}
+
+func claimTaskRaw(t *testing.T, runtimeID, daemonID string) claimResult {
+	t.Helper()
+	w := httptest.NewRecorder()
+	req := newDaemonTokenRequest("POST", "/api/daemon/runtimes/"+runtimeID+"/claim", nil,
+		testWorkspaceID, daemonID)
+	rctx := chi.NewRouteContext()
+	rctx.URLParams.Add("runtimeId", runtimeID)
+	req = req.WithContext(context.WithValue(req.Context(), chi.RouteCtxKey, rctx))
+
+	testHandler.ClaimTaskByRuntime(w, req)
+	out := claimResult{code: w.Code, body: w.Body.String()}
+	if w.Code == 200 {
+		var resp struct {
+			Task *claimRuntimeGuardTask `json:"task"`
+		}
+		if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+			t.Fatalf("decode claim response: %v", err)
+		}
+		out.task = resp.Task
+	}
+	return out
+}
+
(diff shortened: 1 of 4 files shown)


Bohan-J:
Thanks for this one, @seacen — excellent work.

The investigation is what makes this PR: you traced a vanished turn back from the tenant's own timestamps to the exact predicate at fault, and correctly cleared the two paths the old comment blamed instead. Reframing the boundary from "something newer exists" to "already answered" is the right abstraction, and resolving the reply's turn through `COALESCE(chat_input_task_id, id)` so an auto-retry clone is judged by its root batch is the detail that would have been easy to miss.

Second-round verification on `55075e6c`:

- Reverted only the SQL change and re-ran the new tests — the handler test reproduces the production failure verbatim (`claim = 500 chat task has no user input`, task owns 0 user messages, cancelled), and both seal-race tests fail. `TestEnqueueChatTaskLeavesAlreadyPassedMessagesUnowned` passes either way, exactly as documented. The tests genuinely pin the fix.
- Walked the predicate's branches: a batch created after the message stays a boundary, a batch created before it no longer is, and a non-user row with a NULL or dangling task still fails closed. The `role` CHECK constraint keeps `role != 'user'` unambiguous.
- Both joins land on `agent_task_queue_pkey`, and the candidate scan uses `idx_chat_message_session`; the extra work per seal is bounded by the trailing batch.
- `sqlc generate` reproduces the generated file with no drift; this is the only query in the schema carrying that predicate, so there is no sibling with the same bug.
- Full `internal/handler` and `internal/service` suites pass, plus `internal/service` and the channel packages under `-race`. Re-ran everything against a test-merge with current `main`: green.

Keeping the check rather than dropping it, and adding the over-correction test as the guard for that decision, is the part I'd most like to see repeated.

One follow-up, deliberately out of scope here: the stuck typing bubble you documented is still open. The indicator subscribes only to `EventChatDone` and `EventTaskFailed` (`slack/typing_indicator.go`, `lark/outbound.go`, `dingtalk/outbound.go`) with no TTL, so any `task:cancelled` still leaves it spinning. This PR removes the cause that produced those cancellations, not the symptom for other cancel paths. We'll track that separately — no need to expand this PR.

Merging.

PR #6606, opened 2026-08-07
MUL-5907: feat(wecom): show the user their message was heard, by opening the reply as a live bubble

seacen:
> **Stacked on #6590.** The first two commits in this diff are that PR (`0dfc2e685`, `74bd75c74`) — this work needs its ack layer (`(*wsSender).request`, `wecomAPIError`, the replies map). **Review from `909b5a5f0` onward.** If #6590 lands first this rebases to nothing.

## WeCom has no way to say "working on it" — except this

The smart-bot protocol publishes **no typing indicator, no reactions, and no read receipts**. Slack has 👀, Feishu has a typing badge; WeCom has nothing. So today a user who asks the bot a question sees *absolutely nothing* until the agent finishes. On a thirty-second run — never mind two minutes — a working bot is indistinguishable from a dead one. People re-send, or give up.

There is exactly one affordance that can carry the signal. An `aibot_respond_msg` frame with `finish=false` paints a live bubble in the chat, and a `<think></think>` body renders as the client's own animated "..." dots. A later frame carrying the **same stream id** replaces that bubble's body in place; `finish=true` seals it for good.

**So the bubble *is* the typing indicator.** This PR opens one the moment a message is ingested and replaces it with the answer when the run finishes. That is the whole feature; everything below is what it takes to do it without stranding a spinner in someone's chat.

## What it takes

**`stream_store.go` — the seam.** Every stream frame must echo the `req_id` of the callback that opened the turn, and only the WebSocket read loop ever sees that value. The answer turns up minutes later on a bus subscriber holding nothing but a `chat_session_id`. The store maps one to the other.

It is **in-memory on purpose**, and the file says so: one bot is one long connection, the Supervisor's WS lease already guarantees at most one replica holds it, so a handle is only ever useful in the process that created it. A restart losing the handles is the *correct* outcome — the socket those `req_id`s belonged to is gone too, and the answers degrade to plain messages rather than corrupting anything. Persisting them would promise a bubble no surviving socket could write to.

A session holds a **list** of bubbles, not one. Messages inside the engine's debounce window share a run and share a bubble; a message past it is a round of its own that will wait in line — and it gets a bubble immediately, because a wait with nothing on screen reads as a message that was lost.

[...]

PR #6612, opened 2026-08-07
fix(wecom): close the streaming bubble when a run is cancelled, not only when it fails

seacen:
**Stacked on #6606 — review from d4b617760 onward.** The diff below also carries #6606's three commits (0dfc2e685, 74bd75c74, 909b5a5f0); only the last one is new here. Merge #6606 first.

## What the user sees

A spinner that never stops, on a run the system already gave up on.

The WeCom bubble opens as soon as a question arrives — it promises an answer is coming. When the run ends without one, something has to close it. `TypingIndicatorManager.Register` subscribed to `task:failed` and nothing else, so a run that ended any other way left the bubble spinning. Five minutes later the stream-window guard swapped it for "still working, I'll reply separately" — a promise about a run that had already been abandoned. A real user hit this.

## Why `task:cancelled` is its own event

`task:failed` is not the only way a run ends without an answer. The engine cancels a turn it cannot serve — a task claimed with no user input attached is one real case — and `internal/service/task.go` publishes `task:cancelled` for it, not `task:failed`. No `chat:done` follows either, so nothing else closes the bubble on the way past.

Failed and cancelled are two different facts about the engine and one fact about the person watching: no answer is coming. Saying which one it was would tell them nothing they can act on. So the new subscription goes to the existing `handleTaskFailed` rather than to a parallel handler — same closing frame, same copy. The test drives both endings as one table, so the cancelled case can't drift into being treated as the special one.

## Why there is a test in `cmd/server`

What closes the bubble is a subscription, and a missing subscription is invisible. The events keep being published, nobody reads them, and nothing anywhere logs a complaint. Every unit test in `internal/integrations/wecom` builds its own manager and registers it by hand, so all of them would keep passing even if the server stopped wiring one up at all. Wiring, not logic, is what breaks silently here.

`TestWecomBubbleClosersSubscribeOnTheRealBootPath` asserts against `NewRouter` — the same call `main()` makes. It builds two routers on two buses, one with `MULTICA_WECOM_SECRET_KEY` set and one without, and requires the WeCom one to add a listener on each event. Comparing the two keeps the guard honest without it needing to know which other subsystems subscribe to the same events (`autopilot_listeners.go` also takes `task:cancelled`, for one). It checks `chat:done` first as an anti-vacuity marker: if the WeCom boot block never ran, that assertion says so, instead of letting the real ones pass for the wrong reason.

`events.Bus.SubscriberCount` comes with it, to make that assertable from outside the bus — read-only, `RLock`, counts type-specific handlers only.

## Verification

`go build ./...` and `go vet ./internal/...` both clean.

```
go test ./internal/integrations/wecom/ ./cmd/server/ ./internal/events/ -count=1
ok  	github.com/multica-ai/multica/server/internal/integrations/wecom	5.457s
ok  	github.com/multica-ai/multica/server/cmd/server	3.522s
ok  	github.com/multica-ai/multica/server/internal/events	0.177s
```

Reverse check — with the one new `Subscribe` line in `Register` deleted, both new tests fail, and each names the user-visible consequence rather than a symptom:

```
=== RUN   TestARunThatEndsWithoutAnAnswerClosesTheBubble
=== RUN   TestARunThatEndsWithoutAnAnswerClosesTheBubble/task:failed
=== RUN   TestARunThatEndsWithoutAnAnswerClosesTheBubble/task:cancelled
    stream_cancelled_test.go:45: a task:cancelled run produced 1 frames, want 2 (the bubble and its closing frame) — nothing subscribes to task:cancelled, so the spinner runs on for an answer nobody is producing
--- FAIL: TestARunThatEndsWithoutAnAnswerClosesTheBubble (0.00s)
    --- PASS: TestARunThatEndsWithoutAnAnswerClosesTheBubble/task:failed (0.00s)
    --- FAIL: TestARunThatEndsWithoutAnAnswerClosesTheBubble/task:cancelled (0.00s)
FAIL	github.com/multica-ai/multica/server/internal/integrations/wecom	0.486s
```

```
=== RUN   TestWecomBubbleClosersSubscribeOnTheRealBootPath
    wecom_bubble_wiring_test.go:56: nothing in the WeCom boot path subscribes to task:cancelled (0 listeners with WeCom enabled, 0 without). A run that ends on task:cancelled publishes no chat:done, so the bubble it opened is never closed and the user watches a spinner for an answer nobody is producing. Check TypingIndicatorManager.Register.
--- FAIL: TestWecomBubbleClosersSubscribeOnTheRealBootPath (0.00s)
FAIL	github.com/multica-ai/multica/server/cmd/server	0.744s
```

The `task:failed` subtest passing throughout is what shows the rig itself is not vacuous. Restoring the line puts all three packages back to green.

### DIFF
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -697,16 +697,45 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 					Senders:     wecomSenders,
 					Logger:      slog.Default(),
 				})
+				// Streaming replies: WeCom's smart-bot protocol has no
+				// typing indicator, no reaction and no read receipt, so the
+				// only way to say "working on it" is to open the reply early.
+				// The typing indicator paints a stream bubble the moment a
+				// message is ingested and the outbound subscriber replaces
+				// that same bubble with the answer. The store is the seam
+				// between them — it carries the inbound frame's req_id, which
+				// only the read loop ever sees and every stream frame must
+				// echo — so both sides are built with the one instance.
+				wecomStreams := wecom.NewStreamStore()
+				wecomTyping := wecom.NewTypingIndicator(wecom.TypingIndicatorConfig{
+					Senders: wecomSenders,
+					Streams: wecomStreams,
+					// A sweeper's task:failed names a task and not a chat
+					// session, so the session is read back off the task row.
+					Tasks: queries,
+					// A run that fails after its bubble is gone — the guard
+					// closed it at five minutes, or the process restarted
+					// mid-run — still owes the user the news, and the binding
+					// row is where the chat is found when no handle is left.
+					Bindings: queries,
+					Logger:   slog.Default(),
+				})
+				// Subscribes task:failed: a failed run publishes no chat:done,
+				// so this is the sole path that stops the bubble spinning
+				// after a failure.
+				wecomTyping.Register(bus)
+
 				channelRouter.Register(wecom.TypeWecom, wecom.NewResolverSet(
-					wecomStore, wecomSession, wecomReplier,
+					wecomStore, wecomSession, wecomReplier, wecomTyping,
 				))
 
 				// EventChatDone subscriber: pushes the agent's chat reply
-				// back over the same aibot WebSocket the inbound loop owns.
-				// Mirrors slack.NewOutbound(...).Register(bus). Without it
-				// the agent's reply lands only in Multica's web UI — the
-				// user in WeCom sees no response.
-				wecom.NewOutbound(queries, wecomSenders, slog.Default()).Register(bus)
+				// back over the same aibot WebSocket the inbound loop owns —
+				// into the open bubble when there is one, as a new message
+				// when there is not. Mirrors slack.NewOutbound(...).Register(bus).
+				// Without it the agent's reply lands only in Multica's web UI
+				// — the user in WeCom sees no response.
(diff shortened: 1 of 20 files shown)


seacen:
Closing this — #6606 now carries the fix, and took the other fork.

If you already started reading this, here is the disposition.

**The product change is superseded.** This PR routed `task:cancelled` into `handleTaskFailed` and reused the failure copy. #6606 has since implemented cancellation as its own handler with its own copy, for two reasons. "请稍后再试一次" invites a retry of something the user just stopped on purpose. And a cancellation must not chase the binding row the way a failure does: one "cancel all tasks" click would then put a message in every chat that agent serves, including sessions WeCom never painted a bubble in. Both branches edit the same lines of `Register`, so there is nothing here to merge on top.

**The two pieces worth keeping moved across**, in multica-ai/multica@075f033c3c290336d1a3919364b4817236b2cce1 on #6606:

- `events.Bus.SubscriberCount` — unchanged.
- the boot-path wiring guard — strengthened. As written it asserted the subscriptions and nothing else, and `Register` subscribes unconditionally: drop `Tasks` or `Bindings` from the boot config and the guard still went green while every bubble silently stayed open. It now also reads back the manager the boot path registered and asserts all four of its dependencies, each with the consequence the user would see. Reverted one at a time — either subscription, any of the four dependencies, `Register` itself, the typing manager on the resolver set — it fails on each.

`stream_cancelled_test.go` did not come across: it no longer compiles against #6606 (`rig.ask` takes a batch id, task ids are real UUIDs) and it asserts the fork we did not take. #6606's own cancellation tests cover that ground.

Nothing is lost.

PR #6613, opened 2026-08-07
MUL-5968 feat(auth): add Pushover login code delivery

gerbz:
## What does this PR do?

Adds optional [Pushover](https://pushover.net) delivery for login codes while preserving Multica's existing email and log delivery paths. The Pushover card is always visible under **Settings → Integrations** so administrators can discover the server configuration and see whether it is enabled. The per-user Profile controls appear only when the server has a valid `PUSHOVER_APPLICATION_TOKEN`.

Login codes are the first notification type, but the per-user preference model can support more. I'm open to suggestions about which Multica notifications would be most useful to add next.

### Motivation

I haven't even really used Multica yet: while setting up my self-hosted instance, I couldn't reliably get my login codes. Adding Pushover felt like a fun and useful way to contribute right out of the gate.

I have no affiliation with Pushover. I just use it often and thought it would be a helpful optional delivery channel for Multica users.

### Thinking path

Multica already generates one-time login codes and delivers them through configured channels. Pushover therefore fits as an additive delivery channel rather than a new authentication flow. The application token belongs to the deployment and stays server-side; each user's write-only User Key and notification preferences belong in their profile. A small standard-library HTTP client keeps the integration dependency-free, and delivery is considered successful when either the existing channel or Pushover succeeds so an optional notification provider cannot lock users out.

The settings experience follows existing Multica patterns: entering a valid User Key auto-connects without a Save button, notification toggles persist immediately, new notification types default on, and connected users can send a test notification or disconnect their account.

Risks are limited to an external API call during login-code delivery and storage of Pushover User Keys. The request uses a 10-second timeout, errors do not expose credentials, keys are never returned by the API, and existing delivery remains authoritative when Pushover fails.

## Related Issue

N/A — no issue was created for this contribution.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `PUSHOVER_APPLICATION_TOKEN` configuration and expose availability without exposing the token.
- Always show Pushover setup status under **Settings → Integrations**.
- Show per-user Pushover controls under **Settings → Profile** only when the server integration is enabled.
- Auto-connect valid User Keys, default notification preferences on, and persist toggles without a Save action.
- Add **Disconnect Pushover** and **Send Test Notification** actions.
- Deliver login codes through Pushover with title `Multica Login Code`, alongside existing email/log delivery.
- Add backend, frontend, migration, schema-compatibility, configuration, localization, and test-notification coverage.
- Document self-hosted configuration and the two settings surfaces.

## How to Test

1. Start Multica without `PUSHOVER_APPLICATION_TOKEN`. Confirm **Settings → Integrations** shows **Pushover integration not enabled**, while **Settings → Profile** has no Pushover section.
2. Set a valid Pushover Application API Token and restart the backend. Confirm Integrations shows **Pushover integration enabled** and the Profile section appears.
3. Enter a valid User Key. Confirm it connects automatically, the key field is replaced by **Disconnect Pushover** and **Send Test Notification**, and **Login codes** defaults on.
4. Click **Send Test Notification** and confirm the Pushover message has title `Multica Test Notification` and body `You are now setup to receive Pushover notifications via Multica.`
5. Toggle **Login codes** and reload the page to confirm the preference persists without a Save action.
6. Enable **Login codes**, sign out, and request a login code. Confirm the existing email/log path still runs and a Pushover notification arrives with title `Multica Login Code` and the code as its body.
7. Disconnect Pushover and confirm the Profile section returns to the User Key entry state.

Local verification completed:

- `pnpm typecheck`
- `pnpm lint` (0 errors; pre-existing warnings only)
- Full frontend suite: 318 files / 3,740 tests
- Focused Pushover UI suite: 2 files / 13 tests
- Focused Go service and handler tests
- Server compile check and `go vet`
- `git diff --check`

### Settings → Integrations

Pushover integration status

### Settings → Profile

Pushover profile settings

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (N/A: no new runtime/tool/top-level UI tab)

## AI Disclosure

**Prompt / approach:** Codex was used to inspect the existing authentication, settings, database, and configuration patterns; implement the feature; generate focused tests; and run the project's validation commands. Product behavior and configuration choices were directed and reviewed by the contributor on a live self-hosted deployment.


Bohan-J (changes requested):
Thanks for this — it's an unusually complete first contribution. Four-language copy, a migration, backend and frontend tests, self-hosting docs, screenshots, and an explicit AI disclosure. The shape of the feature is right too: it doesn't touch the authentication flow, it doesn't replace email, either channel succeeding counts as delivered, and a Pushover outage can't lock anyone out. That restraint is the correct instinct for an optional delivery channel.

We're requesting changes. There are two product questions we need to settle first, then a set of implementation blockers.

## Product questions

**1. This doesn't solve the problem that motivated it.** You wrote that you couldn't reliably get your login codes while setting up a self-hosted instance. But using Pushover requires entering a User Key, and entering a User Key requires already being signed in. So the first login — exactly the case you hit — still depends on email or on reading the code out of the backend logs. That's not a flaw in the code; it just means the value proposition is "give existing users a second channel," not "fix first-login delivery." If the latter is the problem we want solved, the higher-leverage work is probably email configuration self-checks and better failure diagnostics at startup. Worth being explicit about which one we're buying.

**2. Scope of the notification surface.** You mention login codes are the first notification type and ask which we'd want next. That question is the reason we can't merge this as-is without a decision: once Pushover ships, it becomes a supported surface that every future notification type has to account for. Multica already has a channel system (Lark, Slack, DingTalk, WeCom) and a `notification_preference` table. Our current inclination is to accept Pushover as **self-host-only and frozen at login codes**, with any expansion going through the existing channel/preference machinery rather than growing here. Please don't invest in more notification types on this path until we confirm that.

A third point, less a question than a note: routing login codes through a third-party push service widens the trust boundary. We're OK with that as an opt-in, per-user choice, which is what you built — just flagging that it's why the bar on the security items below is high.

## Blocking

**1. Migration prefix collision — CI will fail.** The branch is well behind `main` now; `265_` is taken by `265_issue_view`, and `main` is currently at `331_`. Merging `main` in and running the migration lint reproduces:

```
--- FAIL: TestMigrationNumericPrefixesStayUniqueAfterLegacySet
    migrations_lint_test.go:91: migration prefix 265 is reused by
    [265_issue_view 265_user_pushover_settings]; use the next unique prefix instead
```

Rename to `332_user_pushover_settings.{up,down}.sql`. The migration content itself is fine — nullable column plus a NOT NULL boolean with a default, no rewrite on PG 11+, no foreign keys, no indexes.

**2. The branch now conflicts with `main`.** GitHub reports `CONFLICTING`; the conflicts are in `server/internal/handler/auth.go`, `server/cmd/server/router.go`, and three of the four `settings.json` locale files. Please rebase.

**3. The Integrations card isn't gated to self-hosted deployments.** `pushover-tab.tsx` renders unconditionally, so when no token is set it shows "Pushover integration not enabled" plus "Set PUSHOVER_APPLICATION_TOKEN on the server." On the managed cloud that text reaches every member — none of whom have a server. This contradicts the convention documented a few lines above your change in `integrations-tab.tsx:36-38`:

```
// Self-host-only integration: the managed cloud reports this false (field
// omitted from /api/config), so the whole section — header included — is
// hidden there rather than showing an operator-only "missing key" message.
```

Please follow the VCS pattern and gate the whole section on `pushoverAvailable`. If you want to keep the "not enabled" state discoverable for self-hosted operators specifically, that needs a separate self-host signal rather than showing it on every deployment. (Related: the empty `<span />` passed as the `SettingsRow` control is a hint that a status-only card doesn't want `SettingsRow` — a plain paragraph reads better.)

**4. `POST /api/me/pushover/test` has no rate limit.** The application token is shared deployment-wide, and Pushover enforces per-application monthly limits and will throttle abuse. Any authenticated user can loop this endpoint and exhaust the whole instance's quota, which takes down login-code push for *everyone*. There's an existing pattern to copy in `router.go` (`authRL`, `contactSalesRL`):

```go
pushoverTestRL := middleware.RateLimit(rdb, envPositiveInt("RATE_LIMIT_PUSHOVER_TEST", 5), time.Hour, trustedProxies)
r.With(pushoverTestRL).Post("/api/me/pushover/test", h.SendMyPushoverTestNotification)
```

**5. Attaching a login-code delivery channel has no verification and no notification.** `UpdateMyPushoverSettings` validates only the 30-character alphanumeric shape, then stores the key and turns login codes on. Two consequences:

- *Security*: anyone who gets a session can silently attach their own User Key and enable login codes. Every future login code for that account then goes to their device, and it survives session revocation. The account owner is never told.
- *Usability*: a key that's shape-valid but mistyped still reports "Pushover connected," and login codes then go silently to the wrong device or nowhere. The user has no way to know.

Both are addressed by the same change: validate the key against Pushover's `users/validate.json` at connect time and only persist on success, and send a notification to the account's email whenever a delivery channel is attached or removed (`EmailService` is already available).

**6. Synchronous external call on the login path.** `deliverLoginCode` runs email and Pushover serially inside the `/auth/send-code` request, and the Pushover client's timeout is 10s. A slow Pushover adds up to 10s to that endpoint's tail latency. Either drop the timeout to 2–3s — a code delivery has no reason to wait longer — or dispatch the Pushover send asynchronously once email has already succeeded.

## Worth discussing before merge

**7. Shape of the preference data.** There's an existing `notification_preference` table (`workspace_id` + `user_id` + `preferences` JSONB, with incremental merge via `PatchNotificationPreference`). This PR instead adds a `pushover_login_codes_enabled` boolean column to `user`. We understand why it doesn't fit directly — login codes are account-scoped and happen before any workspace context, while that table is workspace-scoped. But one boolean column per notification type on `user` degrades quickly if more types arrive. Before a second type lands we'd want an account-level preference store (a `user.notification_preferences` JSONB, or relaxing `notification_preference.workspace_id` to nullable). The current column is acceptable as an interim step — please note it as such in the PR description.

**8. Patch semantics on `PATCH /api/me/pushover`.** Sending `{"user_key": ""}` alone, while the toggle is on, returns `400 a Pushover user key is required to enable login codes`. The caller expressed "disconnect" and got an error about a toggle. Clearing the key should implicitly disable login codes. The UI always sends both fields so this isn't user-visible today, but the API is awkward on its own.

**9. Test coverage gap.** `deliverLoginCode` and `SendMyPushoverTestNotification` are covered well. `UpdateMyPushoverSettings` has no handler test at all — the key-format rejection, the clear-key path, and enable-without-key are exactly where items 5 and 8 live. A `pushover == nil` case for `deliverLoginCode` (the managed-cloud path) would be worth adding too.

## Nits

**10.** `pushover-profile-section.tsx` passes an inline arrow for `isEqual`. `useAutoSave` lists `isEqual` in its effect dependencies, so a fresh identity each render tears down and rebuilds the 650ms debounce timer. All four existing call sites (`account-tab`, `workspace-tab`, `repositories-tab`, `agent-detail-inspector`) pass a module-level function. Hoist it:

```tsx
const userKeysEqual = (left: string, right: string) => left.trim() === right.trim();
```

**11.** An invalid User Key fails silently — typing 29 characters produces no reaction at all, no message. Pushover keys are exactly 30 characters, so an off-by-one paste is a likely mistake worth an inline validation hint.

**12.** Once connected, toggling "Login codes" only surfaces a toast on failure; success is silent, because `SettingsSaveState` is rendered only in the not-connected branch. The PR says toggles persist immediately, but the user gets no confirmation.

**13.** The description sentence is split across `user_key_description_before` / `dashboard_link` / `user_key_description_after`. Nothing else in the repo does this, and `ja`/`ko` already had to leave `_before` empty and push the whole sentence into `_after` — which is the split-sentence pattern failing on different word order. A full sentence plus a standalone "Pushover dashboard ↗" link line avoids it.

**14.** `zh-Hans` translates "User Key" to 用户密钥 while `ja`/`ko` keep it in English. Per `apps/docs/content/docs/developers/conventions.zh.mdx`, identifiers users have to visually match against an external console stay in English — the user is comparing this against the "User Key" label on Pushover's own dashboard. Also, `application_description` points readers to "the project README," but the actual documentation you added lives in `SELF_HOSTING_ADVANCED.md`.

**15.** `PUSHOVER_APPLICATION_TOKEN` is missing from `apps/docs/content/docs/environment-variables.{mdx,ja.mdx,ko.mdx,zh.mdx}`. That's the canonical env-var reference (either the "Email and sign-in" or "External integrations" section fits); `.env.example` and the `SELF_HOSTING*.md` files were updated but this one was missed.

**16.** `PushoverService.Enabled()` reuses `IsValidPushoverKey` to validate the application token. They happen to share a format, but the bigger issue is that a malformed token silently disables the whole integration — the UI just says "not enabled" with nothing in the logs to explain why. A `slog.Warn` at construction when the variable is set but doesn't parse would save operators a lot of guessing.

**17.** The two UI screenshots in `docs/assets/`: no existing self-hosting doc embeds screenshots, and UI images in versioned markdown go stale fast. Prose-only would be easier to keep accurate.

## What we verified locally

- `go vet ./internal/service/ ./internal/handler/` — passes
- `go test ./internal/service/ -run Pushover` — passes
- `go test ./internal/migrations/` — **fails** (item 1, reproduced after merging `main`)
- Locale key parity across the four bundles — passes
- Frontend suite not run locally; relying on CI

---

To summarize the path forward: rebase onto `main` and renumber the migration, gate the Integrations section, rate-limit the test endpoint, verify the User Key at connect time and notify on channel changes, and tighten the login-path timeout. Items 7–9 we should agree on before merge; 10–17 can come along with the same push.

Happy to keep iterating on this — the foundation is solid and the review length reflects how close it is, not how far.

gerbz:
Wow—this is unusually thorough feedback 😄 Thank you for taking the time to dig into it so carefully. Makes perfect sense to carefully consider security codes!

I personally reviewed every point and collaborated with Codex to dial in the proposed implementation. Although, its tempting to, "Codex, fix it!".

I'll bang out the todo's once the questions (which are summarized at the end) are addressed.

## Product questions

### 1. First-login delivery

You’re right that the current implementation does not solve delivery of the first login code. I’d like to explore one narrowly scoped, self-host-only option:

> **Question:** Would you be open to a separate/optional `PUSHOVER_ADMIN_USER_KEY` for bootstrap and invited users who do not yet have a verified personal key? I would include the target email in the notification, preserve existing rate limits, and document that **_this intentionally gives the deployment operator the same login-code visibility the operator currently has through backend logs._**

Without that additional option, I’ll treat Pushover as a secondary delivery channel for existing users in this PR.

### 2. Notification scope

- [ ] Keep this PR limited to login codes.
- [ ] Treat the current per-user boolean as an interim implementation.
- [ ] Leave broader Pushover notification support for a future contribution built around Multica’s existing notification channel and preference machinery.

I hadn’t noticed the existing Notifications tab when I originally built the UI. Since login codes are account-scoped while the current notification preferences appear to be workspace-scoped, my inclination is to keep the Pushover connection and its login-code preference together under Profile for now.

> **Question:** Would you prefer the entire Pushover section under Notifications instead? I could also split the connection and toggle between Profile and Notifications, but keeping the small feature together seems clearer to me.

### 3. Security expectations

Agreed. A third-party delivery channel expands the trust boundary, and I’ll treat validation, rate limiting, failure behavior, credential handling, and user-visible state accordingly.

## Blocking changes

### 1. Migration prefix

- [ ] Rebase on the latest `main`.
- [ ] Rename the migration to the next available prefix at implementation time rather than assuming the previously suggested number is still free.

### 2. Branch conflicts

- [ ] Resolve the current conflicts as part of the rebase.
- [ ] Re-run the relevant backend and frontend test suites afterward.

### 3. Self-host gating and Integrations markup

- [ ] Gate the Pushover Integrations card to self-hosted deployments, following the VCS integration pattern.
- [ ] Continue showing the card to self-hosters when the application token is absent so the disabled integration remains discoverable.
- [ ] Keep the Profile section hidden until `PUSHOVER_APPLICATION_TOKEN` is configured.
- [ ] Replace the empty `<span />` used by the status-only `SettingsRow` with a paragraph.

### 4. Test-notification rate limiting

- [ ] Add a dedicated `RATE_LIMIT_PUSHOVER_TEST`.
- [ ] Apply a limit of five test notifications per user per hour through the existing router rate-limiting pattern.

### 5. User Key validation and security notifications

- [ ] Validate a User Key against Pushover’s `users/validate.json` endpoint before persisting it.
- [ ] Only show “Pushover connected” after validation succeeds.
- [ ] Display invalid-key feedback inline.
- [ ] ? Add email notifications when Pushover is connected or disconnected.

A primary reason for this PR is that it acts as a solution for self hosted setups where email isn't configured.

> **Question:** Just want to re-confirm you want attach/remove emails sent?

### 6. Login request latency

- [ ] Reduce the Pushover request timeout to three seconds.

I’m inclined to take the bounded-timeout option for this PR. A fire-and-forget goroutine would not provide durable delivery on its own, and the login flow still needs to know whether any configured delivery channel succeeded. A proper asynchronous queue could be considered separately if Multica adopts that pattern more broadly.

## Worth discussing before merge

### 7. Notification preference storage

I agree that adding more Pushover notification types directly to the user table would not scale well.

For this PR, my preference is to retain `pushover_login_codes_enabled` as the documented interim account-level preference. I don’t think placing it under an arbitrary workspace would be correct because login is account-scoped and users may belong to multiple workspaces.

Longer term, this could become either an account-level notification preference store or an extension of `notification_preference` that deliberately supports preferences without a workspace.

> **Question:** Are you comfortable retaining the user-level boolean for login codes in this PR, with the expectation that it would be migrated before adding another Pushover notification type?

### 8. PATCH and disconnect semantics

My proposed behavior is:

- [ ] If `user_key` is present and trims to an empty string, clear the key and force `login_codes_enabled` to `false`.
- [ ] Disconnect takes precedence even if the same request explicitly supplies `login_codes_enabled: true`.
- [ ] Reject attempts to enable login-code delivery unless the user already has—or is simultaneously submitting—a valid User Key.
- [ ] Keep omitted fields unchanged for normal PATCH behavior.

That keeps disconnect atomic and prevents an enabled-but-unconfigured state.

### 9. Handler tests

Yes, these should be straightforward to add.

- [ ] Test adding a valid key.
- [ ] Test rejecting an invalid key.
- [ ] Test clearing a key and implicitly disabling login codes.
- [ ] Test enabling without a key.
- [ ] Test updating only the preference.
- [ ] Test the Pushover test-notification rate limit.
- [ ] Test login-code delivery when the Pushover service is nil or unavailable.

## Nits

I’ll address all proposed nits in the next revision:

- [ ] Hoist `isEqual` instead of recreating it.
- [ ] Show invalid User Key feedback inline.
- [ ] Add confirmation when a toggle saves successfully.
- [ ] Replace split-sentence localization with a complete localized sentence and a standalone dashboard link.
- [ ] Keep “User Key” in English in the Chinese localization.
- [ ] Correct the documentation pointer.
- [ ] Document the environment variable in the canonical self-hosting documentation across all four languages.
- [ ] Distinguish Application API Token errors from User Key validation errors and log a safe warning for a malformed application token.
- [ ] Remove screenshots from the versioned documentation.

## Questions Summary

1. Would you accept an optional self-host-only `PUSHOVER_ADMIN_USER_KEY` fallback for bootstrap and invited users without a verified personal key?
2. Should the Pushover UI remain together under Profile, or should the entire section move to Notifications?
3. Are best-effort, non-blocking security emails appropriate when Pushover is connected or disconnected?
4. Is retaining the user-level login-code preference acceptable until a broader account-level notification preference design is needed?

Once I have guidance on those points, I’ll implement the agreed changes together, rebase, test, and update the PR.

PR #6615, opened 2026-08-07
MUL-5912: fix(mobile): preserve offline auth session

### LINKED ISSUE #6614: Mobile: preserve offline auth session after cold start
## Problem

After a mobile app cold start without network access, a saved token is retained but the auth store clears its in-memory user when `getMe` cannot reach the API. Route guards then send the user to login.

## Proposed scope

- Store a last-known user snapshot beside the existing token.
- Keep a locally valid credential admitted when a non-401 account lookup fails and a snapshot is available.
- Route legacy sessions that have a token but no snapshot to an explicit offline account-information screen rather than to login.
- Clear both token and snapshot only when the server returns a genuine 401, or when the user signs out.

## Safety boundary

The server remains authoritative for session revocation. A genuine 401 must still clear the local credential and user snapshot before routing to login.

## Validation

Unit coverage exercises initialization with no token, a non-401 failure with and without a snapshot, and a real 401 cleanup path.

LiaoyuanNing:
## What does this PR do?

Fix the mobile cold-start path that treated an unavailable `getMe` request as a logout. A valid local token remains admitted offline when a saved user snapshot exists; a server 401 still clears the local session.

## Related Issue

Closes #6614

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Store the last verified user snapshot beside the existing token.
- Distinguish credential presence from an online account lookup in auth route guards.
- Preserve the explicit offline account-information screen for legacy sessions that have a token but no snapshot.
- Clear the token and snapshot on a genuine server 401.

## How to Test

Run `pnpm --filter @multica/mobile exec vitest run data/auth-store.test.ts`.

The test covers no token, non-401 failure with a snapshot, non-401 failure without a snapshot, and 401 cleanup.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (device access was unavailable in this run)

## Risks

This patch intentionally does not persist query data, drafts, or mutation queues, and adds no dependencies. A valid local credential is not treated as proof of a valid session: the server's 401 response remains authoritative and clears both stored values.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Traced the cold-start auth route and separated credential presence from the account lookup. Added isolated tests for the four initialization branches.

## Screenshots (optional)

Not included because an authenticated iOS device was not available in this run.

### DIFF
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -1,8 +1,10 @@
-import { Stack, Redirect } from "expo-router";
+import { Stack, Redirect, usePathname } from "expo-router";
 import { useAuthStore } from "@/data/auth-store";
 
 /**
- * Auth-required layout. Redirects to /login when no user is loaded.
+ * Auth-required layout. A valid credential remains admitted while offline;
+ * only the one-time upgrade edge case without a user snapshot lands on the
+ * explicit /offline screen.
  *
  * Workspace membership is enforced one level deeper at [workspace]/_layout —
  * not here — because select-workspace.tsx itself is auth-required but
@@ -10,6 +12,9 @@ import { useAuthStore } from "@/data/auth-store";
  */
 export default function AppLayout() {
   const user = useAuthStore((s) => s.user);
-  if (!user) return <Redirect href="/login" />;
+  const hasToken = useAuthStore((s) => s.hasToken);
+  const pathname = usePathname();
+  if (!hasToken) return <Redirect href="/login" />;
+  if (!user && pathname !== "/offline") return <Redirect href="/offline" />;
   return <Stack screenOptions={{ headerShown: false }} />;
 }
--- a/apps/mobile/app/(app)/offline.tsx
+++ b/apps/mobile/app/(app)/offline.tsx
@@ -15,7 +15,14 @@ export default function OfflineAccountInfo() {
   const isLoading = useAuthStore((s) => s.isLoading);
 
   const retry = async () => {
-    await initialize();
+    // initialize() resets isLoading in a finally, so a rejection here only
+    // means "still no answer" — stay on this screen rather than surface an
+    // unhandled rejection from the press handler.
+    try {
+      await initialize();
+    } catch {
+      return;
+    }
     const { user, hasToken } = useAuthStore.getState();
     if (user) router.replace("/");
     else if (!hasToken) router.replace("/login");
--- a/apps/mobile/app/(app)/offline.tsx
+++ b/apps/mobile/app/(app)/offline.tsx
@@ -1,0 +1,43 @@
+import { ActivityIndicator, View } from "react-native";
+import { SafeAreaView } from "react-native-safe-area-context";
+import { router } from "expo-router";
+import { Button } from "@/components/ui/button";
+import { Text } from "@/components/ui/text";
+import { useAuthStore } from "@/data/auth-store";
+
+/**
+ * Upgrade edge case: a pre-offline-release session has a token but no saved
+ * user snapshot. It is intentionally not treated as a login failure.
+ */
+export default function OfflineAccountInfo() {
+  const initialize = useAuthStore((s) => s.initialize);
+  const logout = useAuthStore((s) => s.logout);
+  const isLoading = useAuthStore((s) => s.isLoading);
+
+  const retry = async () => {
+    await initialize();
+    const { user, hasToken } = useAuthStore.getState();
+    if (user) router.replace("/");
+    else if (!hasToken) router.replace("/login");
+  };
+
+  return (
+    <SafeAreaView className="flex-1 bg-background">
+      <View className="flex-1 justify-center gap-4 px-6">
+        <Text className="text-2xl font-semibold text-foreground">
+          Account info unavailable offline
+        </Text>
+        <Text className="text-sm leading-5 text-muted-foreground">
+          This session is still saved on this device, but it was created before
+          offline account details could be stored. Connect once to restore it.
+        </Text>
+        <Button variant="outline" onPress={retry} disabled={isLoading}>
+          {isLoading ? <ActivityIndicator /> : <Text>Retry</Text>}
+        </Button>
+        <Button variant="destructive" onPress={() => void logout()}>
+          <Text>Sign out</Text>
+        </Button>
+      </View>
+    </SafeAreaView>
+  );
+}
--- a/apps/mobile/app/index.tsx
+++ b/apps/mobile/app/index.tsx
@@ -7,12 +7,14 @@ import { useWorkspaceStore } from "@/data/workspace-store";
  * Entry redirect. AuthInitializer (in _layout.tsx) finishes auth + slug
  * hydration before this renders meaningfully — until then, isLoading is true.
  *
- *   no user            → /login
+ *   no credential      → /login
+ *   credential, no user snapshot → /offline (upgrade edge case)
  *   user, no slug      → /select-workspace
  *   user, slug         → /[slug]/inbox
  */
 export default function Index() {
   const user = useAuthStore((s) => s.user);
+  const hasToken = useAuthStore((s) => s.hasToken);
   const isLoading = useAuthStore((s) => s.isLoading);
   const slug = useWorkspaceStore((s) => s.currentWorkspaceSlug);
 
@@ -24,7 +26,8 @@ export default function Index() {
     );
   }
 
-  if (!user) return <Redirect href="/login" />;
+  if (!hasToken) return <Redirect href="/login" />;
+  if (!user) return <Redirect href="/offline" />;
   if (!slug) return <Redirect href="/select-workspace" />;
   return <Redirect href={`/${slug}/inbox`} />;
 }
--- a/apps/mobile/data/auth-store.test.ts
+++ b/apps/mobile/data/auth-store.test.ts
@@ -15,12 +15,10 @@ class ApiError extends Error {
 }
 
 const secureStorage = {
-  clearCachedUser: vi.fn(),
-  clearToken: vi.fn(),
-  getCachedUser: vi.fn(),
-  getToken: vi.fn(),
-  setCachedUser: vi.fn(),
-  setToken: vi.fn(),
+  clearSession: vi.fn(),
+  getSession: vi.fn(),
+  saveSession: vi.fn(),
+  saveSessionUser: vi.fn(),
 };
 
 vi.mock("./api", () => ({ api, ApiError }));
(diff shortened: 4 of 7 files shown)


Bohan-J (changes requested):
Thanks for the report and the fix — the product behavior here is right, and I'd like to see it land. A temporary network failure should not be indistinguishable from a logout, and keeping the server's 401 as the only signal that invalidates a token is the correct boundary. The `/offline` upgrade path for sessions that predate the snapshot is a thoughtful touch.

Requesting changes on three items, all in the consistency and concurrency semantics of the new persisted state rather than in the product direction.

### 1. The cached user snapshot is not bound to the credential that produced it (blocking)

`apps/mobile/data/secure-storage.ts` stores the user under its own key, independent of the token, and `apps/mobile/data/auth-store.ts:53-59` reads the token and then unconditionally accepts whatever snapshot happens to be present. The two writes are also sequential and non-atomic: `verifyCode` at `auth-store.ts:92-93` writes the token first and the user second, and `logout` at `auth-store.ts:100-101` clears in the same order.

So if the process is killed between those two writes — account switch, OS termination, a crash during login — the device can be left holding token B alongside user A. The next offline cold start then enters the app presenting A's identity while every subsequent request carries B's token. On a shared device that is a cross-account identity leak, and it does not require anything exotic to reproduce.

Please make the credential and the snapshot a single verifiable unit. Either write them as one versioned session record under a single key, or keep a non-forgeable binding in the snapshot (e.g. the user id the token was issued for, or a token digest) and compare strictly before restoring — restoring nothing when the binding does not match. An orphaned snapshot with no token should also be cleared rather than left on disk.

Regression coverage worth adding: A → B account switch, interruption between the two writes in both `verifyCode` and `logout`, and a binding mismatch. The invariant to assert is that another account's snapshot is never displayed as the current session.

### 2. Retry can run concurrently and can undo the 401 cleanup (blocking)

`apps/mobile/app/(app)/offline.tsx:34` disables the Retry button on `isLoading`, but `initialize` never sets `isLoading` back to `true` when it starts — it only ever sets it to `false` (`auth-store.ts:55, 63, 71, 77-82`). By the time the offline screen renders, `isLoading` is already `false`, so the button stays tappable and the spinner never appears. Tapping it repeatedly starts overlapping `initialize` runs.

Once two are in flight, the last one to finish wins the `set`. If one returns 401 and clears the token, and a second one fails later with a network error, that second one writes `hasToken: true` and a `user` back into the store even though the persisted token and the API token have both already been cleared. The app then shows a signed-in session on top of a credential that the server has explicitly invalidated.

Please serialize initialization (or add a generation counter so a stale in-flight run cannot overwrite a newer result) and make sure the 401 cleanup always wins. Setting `isLoading: true` at the top of `initialize` would also make the existing disabled state and spinner actually work. A concurrency test covering "401 resolves first, non-401 resolves second" would pin this down.

### 3. The persisted snapshot is only validated as JSON, not as a user (blocking)

`secure-storage.ts:33` does `JSON.parse(raw) as User`, so the `catch` only fires on a syntax error. Any syntactically valid JSON — `{"foo":"bar"}`, or a snapshot written by an older schema — is accepted as a verified user. `app/index.tsx` then routes past `/offline` into the app, and downstream call sites read a `user.id` that is not there.

This is both a persistence-format boundary and an auth entry point, so it deserves real parsing. `UserSchema` already exists at `apps/mobile/data/schemas.ts:472` — please `safeParse` through it, and additionally reject an empty `id` (the schema's `z.string()` accepts `""`). On failure, delete the key and fall through to the no-snapshot branch. Tests for a corrupted value and an old-shape value would cover it.

### What's already good

The change is well scoped and stays inside the mobile auth layer. The four basic branches — 401, no token, offline with snapshot, offline without snapshot — all have unit tests, and the comments explaining the intent are genuinely useful. There is no database, deployment, or server-compatibility cost here.

Verification I ran on this branch: `pnpm --filter @multica/mobile exec vitest run data/auth-store.test.ts` passes 4/4; `typecheck` passes; `lint` reports 0 errors (7 warnings, none in files this PR touches); `git diff --check` is clean. I did not run an offline cold start on a real device or simulator.

Happy to re-review as soon as the identity-consistency items are addressed.

LiaoyuanNing:
Thanks — all three are real and all three are fixed in `80b74ef`. Two of them I'd argue were understated rather than overstated, and one of the suggested fixes has a trap in it that's worth naming before anyone else takes the same route.

### 1. Snapshot bound to the credential

Went with the single versioned record rather than a binding check: `multica_session` holding `{v: 1, token, user}`, written with one `setItemAsync`. Binding-and-comparing detects the mismatch after the fact; one atomic write removes the window that produces it, and it deletes code instead of adding it — `secure-storage.ts` no longer has a second key to keep in sync.

Two consequences worth calling out explicitly:

**The migration is not optional.** A naive switch to a new key signs out every install that holds the current bare `multica_token` — which is exactly the bug this branch exists to remove, reintroduced by the fix for it. `getSession` therefore falls back to reading the legacy keys, writes the new record, and only then deletes the old ones, so an interrupted migration simply re-runs. I'm aware root `CLAUDE.md` forbids legacy adapters for internal, non-boundary code; I'm treating on-device storage written by an already-distributed build as the boundary case, in the same sense API response parsing is, and I've said so at the function. **If you know for a fact that no install anywhere is holding a `multica_token`, say so and I'll delete the whole path** — it's the one part of this diff I'd rather not carry.

**`clearSession` deletes the legacy keys before the current record**, not after. The reverse order has a resurrection window: killed after `multica_session` is gone but before `multica_token` is, the next cold start finds no session, migrates the stale legacy token, and signs a logged-out user back in. There's a test pinning the ordering.

An orphaned snapshot with no credential is deleted on read, as you asked.

On the regression coverage you specified: A→B switch and binding mismatch are covered. "Interruption between the two writes in `verifyCode` and `logout`" no longer has an interruption point to test, so the structural equivalent is asserted instead — `saveSession` results in exactly one `setItemAsync`, and `verifyCode` calls it exactly once.

### 2. Concurrency

`initialize` is now single-flight (concurrent callers share one run) and sets `isLoading: true` at the top, landed back to `false` in a `finally` — the `finally` matters now that it starts true, otherwise an unexpected throw strands the app on a permanent spinner.

You were right that the spinner and the disabled state were both unreachable; that half wasn't a race at all, just dead code.

One honesty note on the test you asked for: with single-flight, "401 resolves first, non-401 resolves second" is no longer constructible — there is no second in-flight run to resolve. So the test pins the sequential form instead (401 clears the session → the retry finds nothing and stays signed out), plus a test that three concurrent calls produce exactly one `getMe`. If you'd rather see a generation counter so the original interleaving stays expressible as a test, I'll switch it.

`offline.tsx`'s `retry` also catches now, so a rejected `initialize` doesn't surface as an unhandled rejection from a press handler.

### 3. Parsing

`UserSchema.safeParse`, with an empty `id` rejected — as you noted `z.string()` accepts `""`, and `schemas.ts` uses exactly that as the drift sentinel, so restoring it would put a logged-out shape into a signed-in session.

The design point underneath it: **the envelope and the snapshot are validated separately, on purpose.** A snapshot that fails validation must not invalidate the credential sitting next to it — dropping the token there would be another local-problem-logs-you-out path. So a drifted snapshot degrades to the `/offline` upgrade screen and keeps the token; only envelope damage (unparseable, or an unknown `v`) clears the record, because at that point the token can't be trusted either.

### Two things outside your three

- `setUser` was `void setCachedUser(user)` — a rejected write became an unhandled rejection. Now explicitly swallowed, and it updates the snapshot without touching the credential.
- **`isOffline` is set in five places and read by no UI.** Genuine open question rather than a defence: drop it, or wire it to an offline banner in a follow-up? I've left it in place for now because removing state you designed felt like the wrong call to make unilaterally in a review round. Tell me which and I'll do it.

### Verification

Run on `80b74ef`:

```
pnpm --filter @multica/mobile exec vitest run data/secure-storage.test.ts data/auth-store.test.ts
  → 2 files, 30 tests passed
pnpm --filter @multica/mobile exec vitest run
  → 13 files, 92 tests passed
pnpm --filter @multica/mobile typecheck   → clean
pnpm --filter @multica/mobile lint        → 0 errors, 7 warnings (same set you saw, none in touched files)
git diff --check                          → clean
```

`data/secure-storage.test.ts` is new and runs the real serialization logic against an in-memory `SecureStore`. Worth flagging why it had to exist: `auth-store.test.ts` mocks this module out entirely, so **none of your three findings were observable from the existing suite** — the four branches it covered would have stayed green through all of them.

**What I did not verify, and what it would take to close it:** no offline cold start on a real device or simulator, and — more importantly — no upgrade run against a physical install that actually holds a legacy `multica_token`. The migration is covered by unit tests only, and it's the one path in this diff where being wrong signs everyone out. I'd want that exercised on hardware before merge rather than trusted from the unit tests.

Ready for re-review.

LiaoyuanNing:
Dev-iOS verification update for `80b74ef`:

- Re-ran the focused persistence/auth suite: **30/30** tests passed.
- Re-ran the full mobile suite: **92/92** tests passed; `typecheck` passed; `lint` has **0 errors** (the same 7 unrelated warnings).
- Built and installed the branch on an iOS 26.5 iPhone 17 Pro simulator (Xcode 26.6, bundle `ai.multica.mobile.dev`).
- Current TestFlight build 24 is `VALID`, unexpired, and present in the Internal Testing group. Its released source still stores the bare `multica_token`, so removing the migration would force-log-out any build-24 install that upgrades. App Store Connect cannot reveal whether a particular tester still has that build installed; backward compatibility is therefore the safer, evidence-based choice.

`isOffline` has no UI consumer anywhere in mobile. My recommendation is to remove it in a small follow-up rather than add an unreviewed banner to this safety fix; it currently carries no product behavior and is not needed for the corrected auth invariant.

The remaining release gate is a simulator/device upgrade run that starts with an actual legacy SecureStore session. This isolated simulator has no authenticated legacy session, and I will not seed it with a production credential. The unit migration matrix covers the storage conversion, but it is not presented as a substitute for that run.

Please re-review the three addressed code findings. This note does not claim the device-upgrade gate is complete or authorize merge before that evidence exists.

Bohan-J (changes requested):
Thanks for the thorough rework — this is a real improvement over the first round, and the reasoning left in the comments made it easy to follow.

All three items from the previous review are resolved: the credential and its snapshot are now one versioned record under one key written in a single `setItemAsync`; the snapshot is parsed through `UserSchema` with the empty-`id` sentinel rejected; and `initialize` is single-flight with `isLoading` actually driving the `/offline` Retry button and its spinner. The accompanying tests cover those paths well.

Two blocking items remain, both in the new persistence layer.

### 1. The legacy migration re-creates the exact unbound pairing the new format exists to prevent (blocking)

`apps/mobile/data/secure-storage.ts:131-154` reads both `multica_token` and `multica_user`, and when the user parses it writes the two together as one authoritative bound record (`:141-151`).

The problem is where a `multica_user` value can come from. `origin/main`'s `secure-storage.ts` only ever defined `multica_token` — I checked, there is no `multica_user` on the default branch. That key was introduced by this PR's own first commit (`54461f32`), which is unmerged and has never been released. So the only devices that can be holding `multica_user` are ones running an unreleased build of this branch, and on exactly those devices the two keys are the pair that could not be written atomically — the failure this single-record format was introduced to fix.

The result is that a device sitting on token B next to user A has that mismatch canonicalized into the new format, where it becomes undetectable: it now looks like a legitimately bound record.

Please migrate the credential only — take the legacy token, set `user: null`, and delete any `multica_user` unconditionally. Nothing regresses for existing installs: they still are not signed out, they land on `/offline` if that first cold start happens to be offline, and the first successful `getMe` writes a snapshot that is actually provable. `apps/mobile/data/secure-storage.test.ts:132-138` ("carries a valid legacy snapshot across with its token") currently pins the unsafe behavior in place, so it should be inverted to assert that a legacy snapshot is never restored even when it parses cleanly.

### 2. `saveSessionUser`'s read-modify-write can resurrect a signed-out or switched-away session (blocking)

`secure-storage.ts:97-101` awaits `getSession()`, then awaits `saveSession(session.token, user)`. Those are two suspension points with nothing serializing them, and `apps/mobile/data/auth-store.ts:124-129` starts the whole thing fire-and-forget from `setUser`.

So: `setUser` reads account A's session, the user taps Sign out, `clearSession()` completes — and then the pending write lands and puts A's token and snapshot back on disk. The next cold start signs a user who deliberately signed out back in. The same shape applies against `verifyCode`: A's late write clobbers the record for account B.

Moving to one key removed the intermediate state *within* a single write, but not the race *across* calls. Worth noting that this file already reasons carefully about precisely this class of bug — the deletion-ordering comment at `:103-108` describes the same resurrection window one call away.

Please serialize the session mutations (`saveSession`, `saveSessionUser`, `clearSession`) through a single promise chain or a small mutex, or alternatively drop the background persistence in `setUser` and let the next `getMe` refresh the snapshot. Two controlled-interleaving tests would lock this down: "read A, then sign out" and "read A, then sign in as B" — in both, no A session may survive.

### Nits (non-blocking)

- `isOffline` has no production reader. Grepping the diff, it appears only in `auth-store.ts` (the declaration and five `set` calls) and in `auth-store.test.ts`; `index.tsx`, `_layout.tsx`, and `offline.tsx` never consume it. Either wire it into a real offline indicator in this PR or drop it until there is one — an asserted-but-unread flag tends to drift out of sync.
- The PR description has fallen behind the code. "Changes Made" still says the snapshot is stored *beside* the token, which is the two-key design this PR has since replaced, and "How to Test" still describes the original four branches in a single file. Updating it to the single-record + migration story and the current test files would help whoever picks this up later.

### Verification

CI is green on `80b74ef` — 15 checks pass and 1 skipped, including the `mobile` job — and the PR reports as mergeable. What is still missing is a real device or simulator run of the legacy-token upgrade path and the offline cold start; that is worth doing before merge, particularly since the migration logic is about to change.

Happy to re-review once the two blocking items are addressed.

PR #6625, opened 2026-08-08
feat(issues): choose how inline issue references render (plain / compact / full)

szainmehdi:
## What does this PR do?

Adds a per-user preference controlling how inline issue references (`MUL-123`) render inside content — issue descriptions, comments, and chat.

Today every reference expands to a full chip carrying status icon, identifier, and title. In prose that references several issues, the chips dominate the line and push the surrounding text around, which makes the paragraph hard to read. There is currently no way to turn that down.

**Settings → Preferences → "Issue mention style"** now offers three modes:

| Mode | Renders |
| --- | --- |
| `plain` | Identifier only, in the link color. No box, no status icon |
| `compact` | Status icon + identifier in the existing bordered chip. No title |
| `full` | Status icon + identifier + title — **the default, rendering exactly as today** |

Hovering a mention in **any** mode opens a card showing the status, identifier, title, a description snippet, and — where the data exists — the assignee and sub-issue progress. In `compact` and `plain` that card is where the hidden title comes back; in `full` it adds the description, assignee, and progress that no inline chip carries.

**`full` is the default, so the rendering of existing content does not change for anyone who never opens settings.** That was a deliberate constraint: two pre-existing tests in `issue-chip.test.tsx` act as a regression guard on the `full` chip output and were never modified. The one behavioral addition in `full` is the hover card, which is purely additive — it reveals detail on demand and changes nothing about how the prose itself lays out.

### Why per-user rather than a workspace setting

Rendering density is a reading-ergonomics preference, closer to theme than to team policy. The codebase already answers this question for the sticky comment bar — `packages/core/issues/stores/comment-composer-store.ts` persists globally via `defaultStorage`, and its docstring calls it "a personal reading-ergonomics preference (like theme), so it persists globally via `defaultStorage` rather than per-workspace storage". This follows that precedent exactly, which also means **no server changes**: no schema, no migration, no API surface.

### Relationship to #6667

#6667 removed the "open issue links in new tab" preference and left a note on `IssueMentionCard` that "AppLink owns the click semantics… There is deliberately no per-surface or per-preference override." This PR adds a preference to that same component, so it is worth saying explicitly why that is not a reversal.

That rule is about **click semantics**, and it is right: where a link takes you must not vary by surface or by reader, which is exactly what #6667 unified. This preference does not touch click behavior at all — plain click still navigates in place, modifiers still open tabs, and the editor still routes through the same `IssueMentionCard` #6667 pointed it at. What varies is **how much of the issue is drawn inline**, which is reading ergonomics and legitimately personal, like the sticky comment bar or theme. If you would rather this were not a preference at all, say so and I will adapt.

## Related Issue

No existing issue — happy to file one if you'd prefer that before review.

## Type of Change

## Changes Made

**New:**
- `packages/core/issues/stores/issue-mention-display-store.ts` — persisted Zustand store (`multica_issue_mention_display`), modeled on the adjacent `comment-composer-store.ts`. Exports `IssueMentionMode = "plain" | "compact" | "full"`, defaulting to `"full"`.
- `packages/views/issues/components/issue-hover-card.tsx` — the mention preview, used by all three modes.
- `packages/views/issues/components/description-preview.ts` — `descriptionPreview()` extracted out of `board-card.tsx`, where a shared util had been living inside a component. Its test moved with it via `git mv`. No re-export shim, per the repo's rule against internal compatibility layers.

**Modified:**
- `packages/views/issues/components/issue-chip.tsx` — new `variant` prop with three branches. Stays presentational and does **not** read the store; call sites inject the mode, so callers wanting a fixed density (chat's context anchor card) keep `full` by omission.
- `packages/views/issues/components/issue-mention-card.tsx` and `packages/views/editor/extensions/mention-view.tsx` — parallel wiring, deliberately the same shape in both.
- `packages/views/rich-content/rich-content.tsx` — `align-middle` applies only to the boxed variants.
- `packages/views/editor/styles/shell.css` — one scoped rule so plain mentions sit on the text baseline inside the editor.
- `packages/views/settings/components/preferences-tab.tsx` — the select row.
- `packages/views/locales/{en,zh-Hans,ja,ko}/settings.json`.

### Implementation notes for reviewers

**The hover card's fetch is deferred by placement, not by an `enabled` flag.** The `useQuery` lives in the inner `IssueHoverCardBody`, which Base UI mounts only while the popup is open. Hoisting it into the outer component would fire one detail request per mention on render — a paragraph with eight mentions would issue eight. There is a test asserting zero calls before open, and it asserts the trigger rendered first so an empty render cannot satisfy it vacuously.

**Enrichment costs almost nothing.** The assignee uses `ActorAvatar` against the members/agents/squads lists already prefetched app-wide at `staleTime: Infinity` — no additional requests. It is passed `enableHoverCard={false}` and `profileLink={false}` to avoid a nested hover card. Sub-issue progress uses `childIssueProgressOptions`, one workspace-wide request shared with the issues surface and issue detail (deliberately not `childIssuesOptions`, which is per-issue and `refetchOnMount: "always"`).

**`plain` intentionally omits `min-w-0 truncate`.** A bare identifier is short and should wrap with the sentence; an ellipsis mid-prose reads as damage. There is a code comment saying so, because it looks like an oversight otherwise.

**`plain` stays an anchor, not a span**, so it remains focusable and the hover card is keyboard-reachable.

## How to Test

1. `make dev`, then open an issue whose description or comments reference several other issues.
2. **Settings → Preferences → Issue mention style → Compact.** References collapse to badge + identifier. Hover one — after the standard delay, a card shows the full title, a description snippet, and assignee/progress where present.
   Hover a mention in **Full** mode too: the same card appears there, adding the description, assignee, and progress the inline chip does not carry.
3. **Switch to Plain.** References read as inline links at prose size, sitting on the sentence baseline, with no box. Hovering underlines rather than painting a background.
4. Tab to a mention — it takes focus and the card opens.
5. Click and cmd-click a mention in each mode — plain click navigates in place, cmd-click opens a background tab, per the click-intent contract from #6667. Verified in a browser across all three modes, in rendered content and inside the editor composer.
6. Open a comment composer and type an issue identifier. The chip in the editor matches the chosen mode.
7. **Switch to Full.** The inline rendering is identical to `main`.
8. Reload — the mode persists.

Verified locally: `pnpm typecheck` clean across 6 packages; `pnpm test` 3752 passing in `views`, 1347 in `core`, 442 in `desktop`, 188 in `web`; `pnpm lint` 0 errors.

## Checklist

    - No docs change appears to be owed, based on the precedent set by the near-identical sibling preference. #5445 ("open issue links in new tab by default, configurable in preferences") touched 14 files — store, editor, `issue-mention-card.tsx`, `preferences-tab.tsx`, and the four locale bundles — and added nothing under `apps/docs/`, nothing to landing copy, and no changelog entry. `sticky_comment_bar` matches. There is also no page under `apps/docs/content/docs/` that enumerates Preferences rows.
    - **No changelog entry either**, deliberately: `changelog.entries` lives in `apps/web/features/landing/i18n/{en,ja,ko,zh}.ts` and is written at release time, one commit per version. #5445's changelog line landed separately in the v0.4.2 release commit rather than in the feature PR, so I have left that to the release flow. Happy to add one if you'd rather it came with the change.
    - No `SKILL.md` update owed under `server/internal/service/builtin_skills/` — `multica-mentioning` documents mention *semantics* (what enqueues an agent run vs. what merely links), not client-side chip rendering, and this PR changes no CLI flags or API fields.
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — n/a, this adds none of those
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx`

### On the translations

`zh-Hans` follows `conventions.zh.mdx` (任务 for *issue*). **`ja` and `ko` are machine-generated and want a native check** — I added them only because `packages/views/locales/parity.test.ts` requires every EN key to exist in all four bundles. I did check them against the glossary in `conventions.mdx` and corrected an initial draft that used 課題 / 이슈 to the mandated タスク / 태스크, matching the adjacent `issue_link_new_tab` key. Naturalness is still unverified — please treat those two as suggestions.

### Known gaps

- **Touch devices have no hover**, so tapping navigates instead of previewing. Someone with a device should sanity-check that the trigger does not swallow the first tap; I could not verify this.
- **Sub-issue progress is `staleTime: Infinity`** and shared with the issues list, so a long-open card can show a stale `done/total`. Same tradeoff the existing surfaces make, now visible in one more place.
- `apps/mobile` is untouched — it renders bare identifier text today and has no chip to tune.
- `project-chip.tsx` is untouched. Project titles are short and do not produce the overflow this addresses.

## AI Disclosure

**Prompt / approach:** I hit this while reading a comment thread dense with issue references and asked Claude Code to make the expansion configurable. The work went design-first: a brainstorming pass settled the four decisions that shaped everything else (per-user vs workspace, which surfaces, how many modes, and what the default should be), which became a written spec, then a step-by-step implementation plan, then task-by-task execution with a fresh reviewer subagent gating each task and a whole-branch review at the end.

Several defects were caught by that review loop rather than by me, and they are worth naming since they show where the process actually earned its keep:

- The first implementation shipped `delay={0}` on the hover card purely to make a test pass, which would have made issue mentions open instantly while person mentions in the same paragraph waited 600ms. Now uses the library default, with the override confined to tests.
- A test asserted `.not.toHaveClass("border", "text-caption")`. The multi-argument form inverts an `every()` and passes if only one class is absent — a reviewer proved it empirically by adding `border` back and watching the test still pass. Split into single-class assertions.
- Another test mocked `IssueHoverCard` and then asserted the link passed through it — verifying the mock, not the component. Replaced with a click through the real hover card. Removing the mock also revealed three sibling tests had been vacuous.
- Browser QA found `plain` mode rendering in the same colour as surrounding prose. Root cause was a wrong token: `--primary` is a solid button fill (near-white in dark mode), while `--brand` is this codebase's link colour. Confirmed by pixel-sampling before and after.

The design doc and implementation plan are not included here — they are personal workflow artifacts rather than something this repo asked for. Happy to share them if useful.

### Before - Inline Issue Reference Rendering
01-full-mode

01-full-mode

> Hover card — the one addition in Full mode. The chip already shows the title, so the card's contribution here is the description snippet, the assignee, and sub-issue progress.

10-hover-card-full

> Editor Example
07-editor-full

02-compact-mode

> Hover card
09-hover-card-compact

> Editor Example
08-editor-compact

### After (Plain mode)

> Rendering Example

03-plain-mode

> Editor Example
06-editor-plain

> Hover Card
04-hover-card

### After (Settings)
05-settings-row


Bohan-J:
Closing this alongside #6732, where the full reasoning is — rather than repeating it here.

The short version: we shipped #6744, which bounds the chip width in the default rendering instead of making density a preference. The root cause turned out to be that `IssueChip` used `max-w-full` while `ProjectChip` used `max-w-72` — an inconsistency between two chips rendering the same shape in the same prose, rather than a question of how much the chip should show. Your own PR description pointed straight at it, just not for the reason given there.

None of that is a judgement on this implementation. The design-first process you described, and the defects your own review loop caught before anyone else saw them — the `delay={0}` that would have made issue mentions open instantly while person mentions in the same paragraph waited, the multi-argument `.not.toHaveClass` that inverts an `every()` and passes when only one class is absent, the test that mocked the component it was asserting and took three sibling tests down with it — that's a higher standard of self-review than most PRs arrive with. The note on deferring the hover card's fetch by placement rather than an `enabled` flag, so a paragraph with eight mentions doesn't fire eight detail requests, was worth reading on its own.

The hover card deserves to land. It's the part of this PR that never depended on the preference: a description snippet, the assignee, and sub-issue progress, none of which the inline chip shows in any mode. If you want to open it standalone against current `main`, we'll review it.

Thanks for the report, the screenshot, and for scoping it to content surfaces — that's what made it actionable.

PR #6626, opened 2026-08-08
feat(wecom): export the adapter's connection health, with credential rejections counted apart

seacen:
## What does this PR do?

After this, an on-call operator can see a WeCom bot that is down, and can tell **whether to rotate a credential or wait for the network**. Today they can do neither: nothing about the adapter's connection is exported at all.

The distinction is the reason the feature exists. Every failure in this adapter's connection path ends the same way — the connection goes back to the Supervisor for backoff and retry — so a bot that has been down since Tuesday and a bot nobody happened to message today produce identical silence. And when it does surface, one "cannot connect" number is not actionable: a dial or read that fails is infrastructure and usually clears on its own, while a handshake the server refuses on its merits is a wrong secret or a deleted bot, and it will repeat identically on every backoff until a person changes something. Those call for opposite responses, and rotating a perfectly good secret because of a network blip costs a second outage. So they are two series, not one.

## Related Issue

Closes #

## Type of Change

## Changes Made

Four counters under `multica_wecom_*`, chosen for what somebody would page on rather than for completeness:

| metric | fires on |
| --- | --- |
| `connect_failures_total` | dial, handshake write, handshake read (incl. the `subscribeTimeout`) |
| `auth_failures_total` | `aibot_subscribe` answered with a non-zero errcode |
| `inbound_callbacks_total` | callbacks handed to the ingest worker |
| `inbound_queue_blocked_total` | the read loop had to wait on a full ingest queue |

- `server/internal/metrics/wecom.go` — the Prometheus sink; `registry.go` registers it.
- `server/internal/integrations/wecom/metrics.go` — the interface the adapter talks to, with a no-op default. A deployment with `/metrics` off pays nothing.
- `server/internal/integrations/wecom/wecom_channel.go` — the call sites.
- `server/cmd/server/{main,router}.go` — wiring.

Two notes on the diff:

- **The queue counter needed a non-blocking probe ahead of the existing send.** The blocking hand-off is unchanged and still the fallback — backpressure is the deliberate choice there — but the moment it starts is now recorded rather than inferred after the fact from a connection WeCom eventually replaced.
- **`wecomMetricsOrNil` in `router.go` is not ceremony.** A nil `*WecomMetrics` assigned straight into the adapter's interface field gives a non-nil interface holding a nil pointer, the adapter's own nil check passes, and the first counter call panics — only on deployments that turn `/metrics` off, which is the configuration least likely to be exercised anywhere else. It has its own test.

No `installation_id` label anywhere: the same class of unbounded identifier `forbiddenMetricLabels` rejects outright, and it is already on every one of these log lines.

## How to Test

1. `cd server && go build ./... && go vet ./...`
2. `go test -race ./internal/integrations/wecom/ ./internal/metrics/ ./cmd/server/`
3. Point a bot at a wrong secret and scrape `/metrics`: `multica_wecom_auth_failures_total` climbs once per backoff while `multica_wecom_connect_failures_total` stays flat. Block the endpoint at the firewall instead and the two swap.

## Checklist

Each new test was checked by breaking the thing it covers and confirming it fails: merging the two failure counters, dropping the dial/handshake-read call sites, removing the non-blocking probe, storing the raw sink in the factory, dropping the `mx()` nil guard (panics), un-registering the collectors, and returning the typed nil from `wecomMetricsOrNil`.

#6598 touches the same handshake-rejection branch in `subscribe`. It is independent of this PR — no shared code — but if it lands first, this will want a trivial rebase there. Happy to do it in whichever order suits.

### DIFF
--- a/server/cmd/server/main.go
+++ b/server/cmd/server/main.go
@@ -343,6 +343,7 @@ func main() {
 	var businessMetrics *obsmetrics.BusinessMetrics
 	var samplerPool *pgxpool.Pool
 	var channelMediaMetrics *obsmetrics.ChannelMediaReconcilerMetrics
+	var wecomMetrics *obsmetrics.WecomMetrics
 	if metricsConfig.Enabled() {
 		// Build a dedicated tiny pool for the BusinessSamplerCollector
 		// so a stalled scrape can never starve business traffic. If the
@@ -371,6 +372,7 @@ func main() {
 		httpMetrics = metricsRegistry.HTTP
 		businessMetrics = metricsRegistry.Business
 		channelMediaMetrics = metricsRegistry.ChannelMedia
+		wecomMetrics = metricsRegistry.Wecom
 		// Forward inbound daemon WS frames into the per-kind counter so
 		// dashboards can split heartbeat / unknown / invalid traffic.
 		if daemonHub != nil {
@@ -397,6 +399,7 @@ func main() {
 	r, h := NewRouterWithOptions(pool, hub, bus, analyticsClient, storeRedis, RouterOptions{
 		HTTPMetrics:        httpMetrics,
 		BusinessMetrics:    businessMetrics,
+		WecomMetrics:       wecomMetrics,
 		DaemonHub:          daemonHub,
 		DaemonWakeup:       daemonWakeup,
 		FeatureFlags:       flags,
--- a/server/cmd/server/metrics_test.go
+++ b/server/cmd/server/metrics_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/multica-ai/multica/server/internal/analytics"
 	"github.com/multica-ai/multica/server/internal/events"
+	obsmetrics "github.com/multica-ai/multica/server/internal/metrics"
 	"github.com/multica-ai/multica/server/internal/realtime"
 )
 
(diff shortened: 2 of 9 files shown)


seacen:
Rebased onto main now that #6598 has landed.

**What conflicted** was one hunk, in `subscribe()`: #6598 replaced the `fmt.Errorf` this branch was counting beside with `classifySubscribeAck`. I resolved it by branching on that function's verdict rather than keeping my own test of the errcode — `ErrCredentialsRejected` increments `auth_failures_total`, anything it calls unverifiable increments `connect_failures_total`. A second copy of `rejectionErrCodes` next to the counter would be free to drift from the probe's, and then the dashboard and the install screen would disagree about the same code.

That does change what this branch counts: a 45009 throttle used to land on `auth_failures_total` and now lands on `connect_failures_total`, on the "wait" side. It follows from your reasoning for failing closed, so I made the counter follow the classifier rather than re-derive anything. **The one call I made without you is putting an unverifiable ack on `connect_failures_total` instead of counting it nowhere** — say the word if you would rather it were silent there. There is now a test driving a real 45009 through `subscribe()` and asserting it is not reported as a credential problem, which is the case that regresses if anyone re-derives the classification.

**#6598 also made three statements here false**, all the same one: that `auth_failures_total` is "aibot_subscribe answered with a non-zero errcode". Corrected in the `Metrics` interface doc, in the Prometheus help string, and in the commit message.

**Re-reading the branch turned up one more** that predates the rebase and that I had missed when I fixed the same claim elsewhere: `metrics/wecom_test.go` still said the logs "already carry it at every one of these call sites". They do not — the two inbound counters have no log line beside them. Rewritten to say what the attribution actually is, and what leaving the label off actually costs.

**Verification**: `go build`, `go vet` and `gofmt` clean; `go test -race` on `cmd/server`, `integrations/wecom` and `internal/metrics` against a migrated Postgres. I also confirmed the new test goes red if the classification is re-derived from the raw errcode instead of read from `classifySubscribeAck`, so it is guarding the thing it claims to. CI is green on 914ecca.

Issue #5967, opened 2026-07-26
PI (omp) model discovery omits Thinking catalog — thinking_level rejected for all PI agents


Jwd-gity:
## Bug Description

`discoverPiModels()` in `server/pkg/agent/models.go` creates `Model` structs with only `ID`, `Label`, and `Provider` — it does **not** extract the Thinking level catalog from `pi models` output. As a result, the server rejects `thinking_level` for all PI agents with:

```
Invalid request: thinking_level "high" is not a recognised value for runtime "pi"
```

## Root Cause

In `server/pkg/agent/models.go`, `parsePiModels()`:

```go
models = append(models, Model{ID: id, Label: id, Provider: provider})
```

The `Thinking` field is never set, even though `pi models` clearly outputs thinking levels per model:

```
│ claude-opus-5 │ 1M │ 128K │ low,medium,high,xhigh,max │ yes │
```

Compare with how Claude models get thinking levels via `discoverClaudeModels()` — PI has no equivalent.

## Impact

- All PI agents are **unable to use thinking/reasoning levels** through Multica
- `multica agent update --thinking-level high` fails for PI runtime agents
- Users cannot control reasoning effort for PI-backed agents

## Expected Behavior

`parsePiModels()` should extract the thinking column from `pi models` output and populate `Model.Thinking` with the supported levels and default, so the server validates and the UI shows the thinking-level picker.

## Suggested Fix

1. Parse the `thinking` column from `pi models` / `pi --list-models` output
2. Map comma-separated values (e.g. `low,medium,high,xhigh,max`) to `[]ThinkingLevel`
3. Set `Model.Thinking` with `SupportedLevels` and `DefaultLevel`
4. For models with `thinking = -` or `yes`, handle appropriately (no picker / generic toggle)

## Environment

- Multica: v0.4.11 (b0bae3f95)
- PI (omp): 17.1.4
- Platform: Linux/WSL2

## Related

Additionally, `pi --list-models` outputs the same Unicode table format as `pi models` (with `│` box-drawing characters). `parsePiModels()` uses `strings.Fields()` which captures `│` as the first field, producing model IDs like `│/claude-opus-5`. A wrapper script works around this, but the parser should strip non-ASCII table decoration.

Bohan-J:
Fixed by #6629, merged on 2026-08-08.

[...]

PR #3529, opened 2026-05-29
feat: Add thinking level configuration for Pi coding agents

piotr-oles:
## Summary

Exposes the existing `thinking_level` agent field (already used by Claude/Codex agents) to Pi runtime agents. The daemon reads it and injects `--thinking <level>` at Pi subprocess spawn time. The UI gates the existing `ThinkingLevelPicker` component on the Pi runtime.

## Backward Compatibility

Existing Pi agents have `thinking_level = ""` → daemon omits `--thinking` → Pi SDK defaults to `medium`. No behavior change.

PR #6629, opened 2026-08-08
fix(agent): support Pi thinking levels through RPC model discovery (MUL-5924)

### LINKED ISSUE #6628: [Bug]: Official Pi runtime cannot discover or apply per-model thinking levels
Multica supports a persisted `thinking_level` field and a capability-driven thinking picker, but the official Pi runtime does not currently participate in that path.

For a Pi-backed agent, a request such as:

```console
multica agent update <agent-id> --thinking-level high
```

is rejected with an error similar to:

```text
thinking_level "high" is not a recognised value for runtime "pi"
```

The thinking picker is also hidden because Pi model entries do not advertise a `thinking.supported_levels` catalog.

## Root cause

The current integration is incomplete in three places:

1. `discoverPiModels()` runs `pi --list-models`, and `parsePiModels()` creates models with only `ID`, `Label`, and `Provider`.
2. `providerThinkingEnums` / `IsKnownThinkingValue` do not accept Pi's native thinking tokens.
3. `buildPiArgs()` does not forward `ExecOptions.ThinkingLevel` through Pi's native `--thinking` option.

Official Pi's human-readable `--list-models` table cannot provide the missing per-model catalog. Its `thinking` column is only a yes/no value.

## Findings from Pi 0.83

Pi RPC mode provides the required machine-readable metadata:

```console
pi --mode rpc --no-session
```

The relevant commands are:

```json
{"id":"state","type":"get_state"}
{"id":"models","type":"get_available_models"}
```

`get_available_models` returns full model objects containing:

- `provider`
- `id`
- `reasoning`
- `thinkingLevelMap`

`get_state` returns Pi's current model and effective thinking level.

Pi's native thinking vocabulary is:

```text
off, minimal, low, medium, high, xhigh, max
```

The exact supported subset is model-specific. Pi applies these rules:

- A model with `reasoning: false` has no useful per-agent thinking control.
- `off` through `high` are available unless explicitly mapped to `null`.
- `xhigh` and `max` require an explicit non-null mapping.
- Explicit `null` entries disable that level.

This metadata is sufficient for Multica to advertise and validate the exact catalog instead of relying on model-name heuristics or one universal list.

## Expected behavior

- Pi models advertise their exact supported thinking levels.
- The existing UI displays the thinking picker for reasoning-capable Pi models.
- Pi's current model is used when an agent follows the runtime default.
- The server accepts Pi's seven native tokens.
- The daemon validates the selected token against the exact model catalog.
- A valid selection is passed to Pi as `--thinking <level>`.
- Older Pi versions and Pi-family forks continue to fall back to `--list-models` without advertising guessed thinking capabilities.

## Proposed implementation

1. Use Pi RPC mode as the primary model-discovery source.
2. Request `get_state` and `get_available_models`.
3. Convert `reasoning` and `thinkingLevelMap` using Pi's own capability rules.
4. Mark the current Pi model as the runtime default.
5. Preserve extension-registered providers by leaving Pi extensions enabled during discovery.
6. Retain the existing `--list-models` parser as a compatibility fallback.
7. Accept Pi's native token vocabulary in server validation.
8. Pass the selected value through `--thinking`.
9. Filter `--thinking` from `custom_args` so the first-class field is its only owner.

## Steps to reproduce

1. Install and configure official Pi 0.83 with at least one reasoning-capable model.
2. Connect the Pi runtime to a self-hosted Multica instance.
3. Create a Pi-backed agent or open an existing one.
4. Observe that the model catalog has no thinking-level picker.
5. Run:

   ```console
   multica agent update <agent-id> --thinking-level high
   ```

6. Observe the 400 response saying that `high` is not recognized for runtime `pi`.

## Additional context

```text
Deployment: self-host
Multica base commit: 9f2b1c50c
Official Pi: 0.83.0
Platform verified: Linux
Example model: openai-multi/gpt-5.6-luna
```

## Related work

- #5967 reports the same visible gap for Oh My Pi (`omp`), whose discovery format differs from official Pi.
- #3529 was a closed, unmerged implementation attempt based on one static catalog for every Pi model. Current Pi exposes enough RPC metadata to perform exact per-model discovery instead.

chanyeinthaw:
## What does this PR do?

Adds end-to-end thinking-level support for agents using the official Pi runtime.

Multica's existing UI is capability-driven: it shows the thinking picker only when the selected runtime model advertises `thinking.supported_levels`. Previously, Pi discovery used the human-readable `pi --list-models` table, which exposes only a yes/no thinking column. Pi models therefore had no thinking catalog, the server rejected every non-empty Pi `thinking_level`, and the Pi backend never received the selected value.

This change uses Pi's RPC interface as the authoritative discovery source:

```json
{"type":"get_state"}
{"type":"get_available_models"}
```

The returned `reasoning` and `thinkingLevelMap` fields are converted using the same rules as Pi's `getSupportedThinkingLevels(model)`. The current model from `get_state` is marked as the runtime default, allowing an agent with an empty model override to preview and validate the model Pi will actually select.

A selected value now reaches Pi as:

```console
pi --provider <provider> --model <model> --thinking <level>
```

### Thinking path

1. Keep the existing capability-driven UI rather than special-casing Pi.
2. Find an authoritative per-model source because `--list-models` is lossy.
3. Use Pi's documented RPC API and full model objects.
4. Mirror Pi's own `reasoning` / `thinkingLevelMap` capability algorithm.
5. Preserve the existing table parser only as an older-version/fork fallback.
6. Validate Pi's fixed seven-token vocabulary server-side.
7. Re-check the exact model/token pair on the daemon host before execution.
8. Make the first-class `thinking_level` field the sole owner of `--thinking`.

## Related Issue

Closes #6628

Related:
- #5967 — the corresponding gap reported for Oh My Pi (`omp`)
- #3529 — previous closed implementation attempt using a static universal catalog

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/models.go`
  - Discover Pi models through an ephemeral RPC process.
  - Request the current state and available model catalog.
  - Convert `reasoning` and `thinkingLevelMap` into exact per-model thinking levels.
  - Mark Pi's current model as the runtime default.
  - Keep extensions enabled so extension-registered providers remain visible.
  - Fall back to `pi --list-models` for older or incompatible CLIs without guessing capabilities.
  - Keep RPC and fallback within one 15-second discovery budget.

- `server/pkg/agent/pi.go`
  - Pass `ExecOptions.ThinkingLevel` through `--thinking`.
  - Block `--thinking` in `custom_args` so the first-class agent field is the only source of that option.

- `server/pkg/agent/thinking.go`
  - Accept Pi's native values: `off|minimal|low|medium|high|xhigh|max`.
  - Continue using `ValidateThinkingLevel` for exact per-model enforcement.

- `server/internal/handler/agent.go`
  - Document Pi's fixed provider vocabulary and daemon-discovered model subset.

- Tests
  - Cover RPC parsing, explicit null level gaps, default-model resolution, non-reasoning models, table fallback behavior, API validation, invocation injection, and custom-argument conflicts.
  - Compile the agent package for Windows to cover Pi's PowerShell invocation rewrite.

- Built-in agent-creation documentation
  - Document Pi's token vocabulary, RPC-discovered subset, and daemon-owned `--thinking` behavior.

## Compatibility and risk

- Official Pi versions supporting RPC receive exact per-model capabilities.
- Older Pi versions and Pi-family forks fall back to the existing human-readable catalog parser.
- The fallback deliberately leaves `Thinking` unset because a boolean thinking column cannot safely identify supported levels.
- Pi extensions remain enabled during discovery because extensions can register custom providers and models.
- Skills, prompt templates, and context-file loading are disabled for the ephemeral discovery process.
- Empty `thinking_level` remains unchanged: Multica omits `--thinking`, so Pi's native configuration applies.
- Empty model validation uses Pi's current model from RPC state.
- No frontend change is needed; the existing picker consumes the new model metadata.

## How to Test

Automated checks run:

```console
cd server

go test ./pkg/agent -count=1

go test ./internal/daemon \
  -run 'Thinking|ModelList|Pi' \
  -count=1

go test ./internal/handler \
  -run 'Thinking|RuntimeModel' \
  -count=1

go test ./internal/service \
  -run 'TestCreatingAgentsSkillCoversAgentCreationContracts|TestBuiltinSkillsConformToTemplate' \
  -count=1

go vet ./pkg/agent

GOOS=windows GOARCH=amd64 \
  go test -c ./pkg/agent -o /tmp/multica-agent-windows.test.exe
```

All checks passed.

Manual verification against official Pi 0.83 confirmed:

- RPC returns full model metadata including `reasoning` and `thinkingLevelMap`.
- `openai-multi/gpt-5.6-luna` advertises `off|minimal|low|medium|high|xhigh|max`.
- `pi --provider openai-multi --model gpt-5.6-luna --thinking max` completes successfully.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — N/A: no UI code changed; the existing capability-driven picker consumes the new catalog
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — N/A: this extends the existing Pi runtime
- [x] If this PR touches Chinese product copy, I checked it against the Chinese conventions — N/A: no Chinese product copy changed

## AI Disclosure

**AI tool used:** Pi coding agent with OpenAI GPT-5.6 Sol

**Prompt / approach:**

Used Pi to trace Multica's existing model-discovery, server-validation, daemon-validation, UI-capability, and provider-invocation paths. It then inspected official Pi 0.83 documentation and implementation to identify the RPC model contract and exact `thinkingLevelMap` semantics.

The implementation was developed against focused fake-CLI tests, followed by the complete agent package, daemon/handler integration tests, built-in-skill conformance tests, `go vet`, a Windows cross-compile, and a manual official-Pi RPC/invocation smoke test.

## Screenshots (optional)

N/A. The existing thinking picker appears when the daemon publishes the new per-model capability metadata; no frontend component changed.


Bohan-J (changes requested):
Thanks for the thorough writeup — the problem is real, and the direction is right. Reading per-model capabilities from the RPC catalog instead of inferring them from a yes/no column is the correct call, and the projection rules (explicit `null` disables a level, `xhigh` requires an explicit mapping) match Pi's own `getSupportedThinkingLevels` line for line. I verified that against `@mariozechner/pi-ai`.

I ran the change against real Pi installs (0.67.2 locally, plus 0.73.1 from npm) rather than only reading the diff, and found one blocking regression plus one smaller correctness issue.

## Blocking: an RPC failure that doesn't exit fast starves the table fallback

`discoverPiModels` gives RPC and the fallback one shared context:

```go
// server/pkg/agent/models.go:812
runCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
defer cancel()
if models, ok := discoverPiModelsRPC(runCtx, executablePath, lookedUp); ok {
    return models, nil
}
return discoverPiModelsTable(runCtx, executablePath)
```

`discoverPiModelsTable` then derives its own timeout from that already-consumed context (`models.go:999`). If RPC used the full budget, the child context is cancelled on creation, `cmd.Output()` fails immediately, and the function returns `[]Model{}, nil` — **an empty catalog with a nil error**, so nothing upstream can tell discovery failed.

Measured with one fake CLI that accepts `--mode rpc` but never answers:

| | elapsed | result |
|---|---|---|
| `main` | 0.23s | `[glm-coding-plan/glm-4.7]` |
| this PR | 16.0s | `[]` |

The trigger I'd actually worry about is not an exotic fork — it's Pi's own error envelope. **Pi omits the request `id` on unknown commands.** On 0.73.1:

```console
$ printf '{"id":"a","type":"get_state"}\n{"id":"b","type":"totally_unknown_command"}\n' | pi --mode rpc --no-session --no-skills --no-prompt-templates --no-context-files
{"id":"a","type":"response","command":"get_state","success":true,"data":{...}}
{"type":"response","command":"totally_unknown_command","success":false,"error":"Unknown command: totally_unknown_command"}
```

The read loop dispatches purely on `response.ID` (`models.go:880`), so if `get_available_models` or `get_state` is ever renamed or dropped, that id-less error matches no case, never sets `modelsDone`, and the loop blocks until the 15s deadline — because stdin is only closed after the loop exits. I reproduced exactly this with Pi's real error envelope: 15.0s, empty catalog.

Three things make it worse than a slow path:

- Empty catalogs are deliberately not cached, so every request re-pays the 15s.
- With an empty catalog, `ValidateThinkingLevel` finds no default model and returns false, so a saved `high` is silently dropped — which reintroduces the exact symptom #6628 reports.
- The nil error means nothing is logged.

Suggested fix, both parts:

1. Dispatch on `response.Command` as well as `response.ID` — Pi echoes `command` in both success and error envelopes (your own fixture already emits it) — and treat `success:false` as terminal for that request.
2. Give the fallback an independent deadline instead of deriving it from the spent `runCtx`, or split the budget explicitly (e.g. 7s RPC / 8s table).

To be fair to the design: **the common older-version path is fine.** I confirmed 0.67.2 rejects `--no-context-files` as an unknown option and exits 1 immediately, leaving the fallback nearly its full budget. The problem is specifically RPC failure modes that don't exit fast.

## Blocking (smaller): `max` is not a Pi thinking level

`server/pkg/agent/thinking.go:765` adds a seven-token vocabulary including `"max"`, but Pi has six. From `@mariozechner/pi-ai/dist/models.js`:

```js
const EXTENDED_THINKING_LEVELS = ["off", "minimal", "low", "medium", "high", "xhigh"];
```

`ThinkingLevelMapSchema` also declares only those six optional keys, so `thinkingLevelMap.max` can never appear in real RPC output — the `max` branch in `piThinkingLevelOrder` is dead code. The CLI rejects it too:

```console
$ pi --thinking max
Warning: Invalid thinking level "max". Valid values: off, minimal, low, medium, high, xhigh
```

Net effect: a user can save `thinking_level: "max"` (the new `TestCreateAgent_PiThinkingLevelValidation` asserts 201 for it, described as a "native max value"), but it will never appear in the picker and will always be dropped before execution. That's the same "saved but inert" class of bug this PR is fixing.

Worth a closer look: `writeFakePiRPCModelsBinary` emits `"max":"max"` inside `thinkingLevelMap` and then asserts `max` shows up in the catalog. Since real Pi cannot emit that key, the test locks in a contract that doesn't exist.

One thing to confirm before you act on this: the latest published `@mariozechner/pi-coding-agent` I could verify against is 0.73.1, while the PR refers to Pi 0.83. If 0.83 genuinely adds `max`, disregard this section — but could you paste its `EXTENDED_THINKING_LEVELS` and `ThinkingLevelMapSchema`? The six-value result was identical on both 0.67.2 and 0.73.1.

## Nits

- `piThinkingFromRPCModel`'s comment says it mirrors `getSupportedThinkingLevels(model)`, but Pi returns `["off"]` for `reasoning: false` while this returns `nil`. Hiding an "Off"-only picker is the better UX — just worth noting the deviation in the comment rather than claiming a mirror.
- `piRPCModel.Name` is parsed but unused; `Label` stays `provider/id`. Consistent with the table path, so not wrong, but RPC now gives you real names like `Gemini 3 Flash Preview (Cloud Code Assist)`, which would read better in the picker.

## What's solid

- The capability projection matches Pi's algorithm exactly, including the `xhigh`-needs-explicit-mapping rule.
- A fresh or unauthenticated Pi returns `{"models":[]}` with `success:true`; the `len(rawModels) == 0` guard correctly rejects that and falls back instead of publishing an empty catalog. Verified with a clean `HOME`.
- Process lifecycle is right: close stdin to signal EOF, drain stdout, then `Wait()`, with `WaitDelay` set — no leaked discovery sessions and no `StdoutPipe` misuse.
- Keeping `Thinking` unset in the table fallback, and making the first-class field the sole owner of `--thinking`, are both the right boundaries.

Happy to re-review as soon as the discovery-budget issue is addressed — the rest is close.

chanyeinthaw:
I fixed the fallback issue in `4dd2a0403`.

RPC responses are now matched by either `id` or the echoed `command`, so the ID-less error envelope you showed is treated as terminal. I also split the existing 15-second discovery budget into 7 seconds for RPC and 8 seconds for `--list-models`, so a hung RPC session cannot starve the fallback.

I added regression tests for both cases and adjusted the `reasoning: false` comment as suggested.

---

Regarding `max`: you are correct for Pi 0.73.1, but it was added by Pi 0.83.0.
Here are the declarations from the `v0.83.0` tag.

[`EXTENDED_THINKING_LEVELS`](https://github.com/earendil-works/pi/blob/v0.83.0/packages/ai/src/models.ts#L661-L670):

```ts
const EXTENDED_THINKING_LEVELS: ModelThinkingLevel[] = [
  "off",
  "minimal",
  "low",
  "medium",
  "high",
  "xhigh",
  "max",
];

export function getSupportedThinkingLevels<TApi extends Api>(
  model: Model<TApi>,
): ModelThinkingLevel[] {
  if (!model.reasoning) return ["off"];

  return EXTENDED_THINKING_LEVELS.filter((level) => {
    const mapped = model.thinkingLevelMap?.[level];
    if (mapped === null) return false;
    if (level === "xhigh" || level === "max") return mapped !== undefined;
    return true;
  });
}
```

[`ThinkingLevelMapSchema`](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/src/core/model-config.ts#L54-L63):

```ts
const ThinkingLevelMapValueSchema = Type.Union([
  Type.String(),
  Type.Null(),
]);

const ThinkingLevelMapSchema = Type.Object({
  off: Type.Optional(ThinkingLevelMapValueSchema),
  minimal: Type.Optional(ThinkingLevelMapValueSchema),
  low: Type.Optional(ThinkingLevelMapValueSchema),
  medium: Type.Optional(ThinkingLevelMapValueSchema),
  high: Type.Optional(ThinkingLevelMapValueSchema),
  xhigh: Type.Optional(ThinkingLevelMapValueSchema),
  max: Type.Optional(ThinkingLevelMapValueSchema),
});
```

Pi 0.83.0 also lists `max` in its [--thinking CLI documentation](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/README.md#L559) and explains that `xhigh` and `max` are opt-in per-model levels in the [RPC documentation](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/docs/rpc.md#L289-L291).

Bohan-J (approved):
Thanks for the fast turnaround, and for pushing back on the `max` point with sources — **you were right and I was wrong**, so let me correct that publicly.

I had been verifying against `@mariozechner/pi-coding-agent@0.73.1`, assuming it was the latest release. It isn't the same distribution — the real upstream is `earendil-works/pi` / `@earendil-works/pi-coding-agent`, currently at v0.84.1. I installed 0.83.0 and confirmed every point you made: `max` is in `EXTENDED_THINKING_LEVELS`, `ThinkingLevelMapSchema` declares a `max` key, the `--thinking` help lists it, and `getSupportedThinkingLevels` gates `xhigh` and `max` identically on `mapped !== undefined` — exactly what your Go code does. My objection was an artifact of the wrong package, and the `"max":"max"` fixture is a faithful contract, not a fabricated one.

While I had 0.83.0 installed I ran a differential check over its entire bundled provider registry — every model through both Pi's own `getSupportedThinkingLevels` and your `piThinkingFromRPCModel`:

```
total=1153  reasoningMatched=897  nonReasoningHidden=256  mismatched=0
```

Zero divergence across all 897 reasoning-capable models, including the 143 that genuinely expose `max` and the 179 that expose `xhigh`. The 256 non-reasoning models correctly hide the picker.

On the fallback budget, I re-ran the two scenarios from my previous round against `4dd2a04`:

| scenario | before | after |
|---|---|---|
| RPC accepts the mode but never answers | 16.0s, `[]` | 8.0s, fallback catalog returned |
| `success:false` response for the models request | 15.0s, `[]` | 0.33s, fallback catalog returned |

Matching on the echoed `command` in addition to `id` is a good call to keep, even though 0.83 does echo `id` on unknown commands — Pi-family forks exist that don't, and it costs nothing.

One thing worth preserving deliberately: closing stdin only *after* the read loop. I hit this while testing by hand — piping the two requests and letting stdin EOF immediately gets you the `get_state` response and nothing else, because `get_available_models` awaits `session.modelRuntime.getAvailable()` and the session shuts down while it's still in flight. Holding stdin open until both responses land is what makes discovery reliable, so that ordering is load-bearing rather than incidental.

Non-blocking observation: the table fallback went from 15s to 8s. The 15s was originally chosen because a multi-provider `--list-models` measured 4.6–4.8s and 5s got punched through by jitter, so this trims headroom from ~3x to ~1.7x. My machine has no providers configured (`--list-models` returns in ~0.4s), so I could not validate the multi-provider case — worth keeping an eye on, but the worst-case total is still 15s, so it is not a merge blocker.

Verified locally on `4dd2a04`: full `go test ./pkg/agent`, `go vet ./pkg/agent`, the Windows cross-compile, and the built-in skills conformance tests all pass.

Nice work on this one — the capability projection is exactly right, and the conservative fallback behaviour is the right instinct throughout.

PR #6631, opened 2026-08-08
fix(channels): a cancelled run must stop the typing indicator, not leave it spinning

seacen:
Two things, and the first one is the bigger of them. The DingTalk half that was here through four rounds has been taken out; why is at the bottom.

## A deleted session's cancels were reaching nobody at all

Leading with this because the last review asked for it:

> `BroadcastCancelledTasks` taking the workspace from its caller is a genuinely good find. A deleted session's cancels were resolving their workspace through the very row the transaction had just dropped, so `publishTaskEvent` discarded them before the bus — those events were reaching nobody at all, not just the channel indicators. Worth calling out in the PR description; it is a bigger fix than the indicator work around it.

Deleting a chat session cancels the session's active tasks and deletes the session in one transaction, then calls `BroadcastCancelledTasks` after the commit. That resolved each task's workspace through `ResolveTaskWorkspaceID`, which reaches a chat task's workspace by reading its `chat_session` — the row the same transaction had just deleted. The lookup missed, the resolution returned `""`, and `publishTaskEvent` drops an event with no workspace before it reaches the bus.

So the event was never published. Not "published but nobody was subscribed" — it never reached the bus, so **nothing anywhere heard it**: not the web queue, not the realtime fanout, not any channel adapter. Every consumer of `task:cancelled` was blind to a deleted conversation. The runtime teardown has the same shape: it deletes the runtime's system agents and `chat_session.agent_id` is `ON DELETE CASCADE`, so the session is gone before the broadcast there too.

`BroadcastCancelledTasks` now takes the workspace from its caller. All three call sites already have it — it is the workspace whose session, member or runtime is being torn down — so the lookup is not needed and cannot fail.

Pinned by `TestDeleteChatSession_BroadcastsTaskCancelled`, which drives the real `DeleteChatSession` handler against Postgres and asserts one event carrying the workspace and the chat session id. With the fix backed out:

```
--- FAIL: TestDeleteChatSession_BroadcastsTaskCancelled
    deleting the session cancelled its queued task and published 0 task:cancelled events
```

## Slack and Lark: the indicator comes off on a cancel

Both adapters subscribed only to `EventChatDone` and `EventTaskFailed`, with no TTL, so a run that ended in `task:cancelled` left the indicator up for good — Slack the 👀 on the user's message, Lark the Typing badge on theirs. This is the case #6611 left open on merging it:

> the stuck typing bubble you documented is still open. The indicator subscribes only to `EventChatDone` and `EventTaskFailed` (`slack/typing_indicator.go`, `lark/outbound.go`, `dingtalk/outbound.go`) with no TTL, so any `task:cancelled` still leaves it spinning. This PR removes the cause that produced those cancellations, not the symptom for other cancel paths. We'll track that separately — no need to expand this PR.

Two of the three adapters named there are closed here. `dingtalk/outbound.go` is not, and stays open — see below.

**Slack — `typing_indicator.go`.** `EventTaskCancelled` added to the same handler. `Clear` now resolves the bot token from the installation id recorded at add time rather than through the session's binding: a session delete drops that binding while the cancel it triggered is still in flight, and `Clear` takes the state out of the map before the lookup, so a miss there used to leave nothing to retry from.

**Lark — `outbound.go`, `typing_indicator.go`.** Same subscription on the Patcher, and the clear runs before the binding lookup and before the channel-provenance gate. Where the clear sits matters as much as the subscription: both lookups can answer "no" for a run whose badge is still on screen. The binding is already deleted on the session-delete path, and `TaskInputIsChannelIngested` reports no channel-ingested messages for a task cancelled for owning an empty input batch — the #6611 shape. Nothing is posted for a cancellation, so neither gate is protecting anything.

### Two holes left open, named rather than papered over

**Archiving an agent.** It cancels its tasks through `CancelAgentTasksByAgent` and deliberately broadcasts nothing per row, because `agent:archived` already invalidates every client's task list (`handler/agent.go`). No client-side list refresh removes a reaction from a Slack message, so archiving an agent mid-run still leaves the indicator up. "Every cancel path lands here" was claimed in an earlier revision, is wrong, and is not claimed anywhere now.

**A run that ends while the indicator is still being added.** Both adapters record their state only *after* the external call returns, and the Router calls `OnIngested` on a detached goroutine (`channel/engine/router.go`). If the run ends first, the clear finds nothing and the add lands behind it with nothing left to take it off. Not a regression from this PR — chat-done and task-failed race the add the same way on `main`. The fix suggested in review (a per-session generation `Clear` bumps and `Add` captures before its call, undoing itself if the value is stale when the call returns, driven deterministically, no sleeps) is the right one and is its own change. Both holes are named in each adapter's gap list, so the follow-up is discoverable from the code.

### No TTL in this PR

Bohan named two problems and this fixes one. After this change every terminal state that is published and observed closes the indicator: done, failed, cancelled, and never-started (the engine's `OnSettled`). What remains is the ending nothing acts on — an event lost, a handler that panicked, an adapter restarted mid-run — plus the two holes above. A TTL answers that and is a separate mechanism: a timer per session, shutdown semantics, and copy for runs that legitimately outlive it. Happy to take it as a follow-up.

## DingTalk has been taken out of this PR

It was in the first four revisions and is now reverted to `main` (`ack.go`, `outbound.go` and their tests; `router.go` goes back with them, since it passed the ack notifier into `NewOutbound` only so the outbound could reach it).

Four rounds of review closed every blocker that was reported, and every one of those fixes created a fresh instance of the same defect: a path that stays silent while a room is still holding an unanswered "👀 On it". One slot per session lost the second round's promise — the repro in the last review. Releasing the whole room's queue on an ending discharged a run that was still working. A per-room take that only spoke when the room emptied went quiet for a bulk cancel. The review of the fix for those still found defects of the same family, including ones the fix itself introduced.

The reason is DingTalk's API, not the bookkeeping on top of it. The classic robot this adapter sends through exposes no reaction, so the indicator is a real, non-retractable message. Clearing it means **posting a second message**, which forces a question Slack and Lark never have to answer: is a reply still coming to this room? Removing a reaction owes nobody that answer — it is correct whether or not another run is in flight. Posting "no reply is coming" is wrong the moment one is.

Every revision answered that question from process memory — a ledger of outstanding acks per session — and process memory keeps turning out to be the wrong place to keep it. It cannot survive a restart, it has to be reconciled against every way a run can end, and each new ending we taught it about was another chance to discharge the wrong promise.

**It will come back as its own PR with a different contract:** post the notice unconditionally on a cancel and accept an occasional duplicate. That deletes the ledger rather than patching it — no promise to take, no ordering to get right, no state to lose on restart. A room seeing "that run was cancelled" twice costs the user a moment of confusion. Silence leaves them waiting on a reply that is never coming. For a message that cannot be unsent, that is the direction to fail in.

## Why the tests publish on a bus

Every existing test on these paths calls the handler by hand — `m.handleEvent(...)` in Slack, `p.handleEvent(...)` in every Patcher test in `lark/outbound_test.go`. A test written that way passes with the fix reverted: it proves the handler works and says nothing about whether the event reaches it, which is the entire defect. All the new adapter tests publish on a real `events.Bus` after `Register`.

## Reverse verification

Each production change backed out on its own, full package suite each time, so the blast radius below is the whole list.

`BroadcastCancelledTasks` resolving the workspace per task instead of taking the caller's — `internal/handler`, 1 failure; `internal/service` and `cmd/server` stayed green:

```
--- FAIL: TestDeleteChatSession_BroadcastsTaskCancelled
```

`task:cancelled` subscription removed from Slack's `Register` — 3 failures:

```
--- FAIL: TestTypingIndicator_ClearsOnTaskCancelled
--- FAIL: TestTypingIndicator_ClearsAfterSessionDeleteRemovedTheBinding
--- FAIL: TestTypingIndicator_ClearsOnCancelOfATaskThatReadsAsAWebRun
```

Slack's `Clear` resolving the installation through the session binding again instead of from what the add recorded — 1 failure:

```
--- FAIL: TestTypingIndicator_ClearsAfterSessionDeleteRemovedTheBinding
```

`task:cancelled` subscription removed from the Lark Patcher — 2 failures:

```
--- FAIL: TestPatcherClearsTypingOnTaskCancelled
--- FAIL: TestPatcherClearsTypingAfterSessionDeleteRemovedTheBinding
```

The Lark clear moved behind the binding lookup and the provenance gate — 2 failures, which is the trap in putting it in the obvious place:

```
--- FAIL: TestPatcherClearsTypingOnTaskCancelled
--- FAIL: TestPatcherClearsTypingAfterSessionDeleteRemovedTheBinding
```

Lark's `Clear` resolving credentials through the session binding again — 1 failure:

```
--- FAIL: TestPatcherClearsTypingAfterSessionDeleteRemovedTheBinding
```

Every change was restored and the package re-run green immediately after.

## Gates

```
cd server
go build ./...                       # exit 0
go vet ./internal/handler/... ./internal/service/... \
       ./internal/integrations/{slack,lark,dingtalk}/... ./cmd/server/...   # exit 0
gofmt -l <every changed file>        # empty

DATABASE_URL=... go test -race -count=1 ./internal/handler/... ./internal/service/... \
  ./internal/integrations/{slack,lark,dingtalk}/... ./cmd/server/...
ok  .../internal/handler              40.912s
ok  .../internal/service               8.027s
ok  .../integrations/slack             4.421s
ok  .../integrations/lark              2.861s
ok  .../integrations/dingtalk          5.410s
ok  .../cmd/server                    11.033s
```

`dingtalk` is in that list because it was reverted, not because it changed: it is byte-identical to `main` and re-run to prove the revert left nothing behind.


Bohan-J (changes requested):
Product direction is right and I'd like this to land: every terminal state has to close the indicator, and a cancel is a terminal state. Requesting changes on the implementation — three blockers, and they share one root cause. Both "was this turn channel-originated?" and "how do we close the indicator?" are derived at cancel time from data that may already be gone by then.

## Blockers

**1. DingTalk still misses the empty-input-batch cancel that motivated the fix.**

`server/internal/integrations/dingtalk/outbound.go:107` gates the notice on `TaskInputIsChannelIngested`. The production shape behind #6611 is a task that owns an input batch (`chat_input_task_id` set) holding no channel-ingested rows — `server/internal/handler/daemon.go:2346` cancels exactly that task. The gate answers `false` there, so no notice is sent and the user is still holding "👀 On it — I'll reply here when it's ready." `outbound_test.go` sets `channelIngested = true`, so the case that fails in production isn't reproduced.

**2. On the delete-session cancel path, the binding is already gone.**

`server/internal/handler/chat.go` deletes `channel_chat_session_binding` and the chat session inside the transaction, then broadcasts the cancels post-commit via `BroadcastCancelledTasks`. All three new handlers resolve their target or credentials from that binding, so this cancel path still leaves the indicator behind. Slack makes it unrecoverable: `Clear` drops the in-memory state before the binding lookup (`server/internal/integrations/slack/typing_indicator.go:124-134`), so after the lookup misses there is nothing left to retry from.

**3. A bulk cancel makes DingTalk post N identical notices.**

Cancel is broadcast per row — `server/internal/service/task.go:2117` (issue-level), `:2136` (agent-level), `:2172` (`BroadcastCancelledTasks`, which is what the session delete uses). One "cancel all tasks" click, or one session delete with three queued turns, posts three "⚠️ That run was cancelled" messages into the same conversation.

## Suggested design

**A. Stamp channel provenance on the task row instead of deriving it from message rows.**

`EnqueueChatTask` has exactly one caller in the repo — the channel Router (`server/internal/integrations/channel/engine/router.go:677`). A run created there is a channel run by construction, whether or not the seal ended up owning zero rows. Add `agent_task_queue.channel_origin` (or `origin_channel_type`), stamp it in the same transaction that creates the task, carry it onto auto-retry clones the way `chat_input_task_id` is carried (`server/pkg/db/queries/agent.sql:504`), and have `TaskInputIsChannelIngested` read the column first, falling back to today's message-derived check for pre-migration rows. It's a plain added column with a constant default — no table rewrite, no new index.

With that in place:

- the DingTalk gate becomes correct instead of being bypassed;
- the Lark clear no longer has to sit ahead of the gate. `server/internal/integrations/lark/outbound.go:311-326` and its ordering comment can go away, and so does the risk it accepts — cancelling a web-origin queued task currently clears the badge of a channel turn that is still running, because the clear is keyed by session;
- the same misclassification in the cancel core path is fixed too: `server/internal/service/task.go:2436` and `:2535` run the same query, so an empty-batch channel turn takes the web draft-restore branch today instead of writing "Stopped."

**B. Don't resolve the indicator's credentials through the binding.**

Slack (`server/internal/integrations/slack/typing_indicator.go:134`) and Lark (`server/internal/integrations/lark/typing_indicator.go:142`) both look up binding → installation on clear, but `Add` already holds the installation. Keep `installation_id` in the tracked state — the id, not the token, so the "no decrypted credential held between add and clear" property is unchanged — and resolve the installation directly. Installation rows survive session deletion, so blocker 2 falls out for free, and the Slack state-loss bug above disappears with the lookup.

**C. Make DingTalk's withdrawal correspond to the ack, not to the task.**

`ack.go` already coalesces per session (`server/internal/integrations/dingtalk/ack.go:36`). Extending that into an "ack outstanding" state gives you blocker 3's dedupe: withdraw only when there is an unclosed promise, clear the state once withdrawn. It also keeps web-origin cancels out of the room independently of A — no ack was posted there, so there is nothing to withdraw.

## Tests worth pinning

- A channel task whose owned batch holds zero channel-ingested rows, cancelled → all three adapters close out.
- A session deleted while its binding is live → Slack and Lark still clear.
- Three tasks cancelled in one call → DingTalk posts one notice.

## Not blocking

Agreed on keeping TTL / restart recovery out of this PR. Per-turn (rather than per-session) indicator keying is also fine as a follow-up: doing it properly means threading the durable `chat_message` id through `OnIngested` and persisting the state, which pairs naturally with the TTL work. After A, the remaining mis-clear window is narrow — two channel turns in one session, one running and one queued, cancelling the queued one.

If #6606's WeCom close-out is still in flight, it's worth converging both on the same provenance helper and state shape in the engine layer rather than restating the semantics per adapter.

seacen:
All three are addressed at `450eacb35`. One of them was deeper than the adapter
layer, so that one first.

## Blocker 2 — the session delete published no `task:cancelled` at all

The adapters were never the problem here. `DeleteChatSession` deletes the
`chat_session` row inside the transaction and calls `BroadcastCancelledTasks`
after the commit; by then `ResolveTaskWorkspaceID` cannot resolve a workspace
for a chat task whose session is gone, and `publishTaskEvent` returns early on
an empty workspace. Nothing reached the bus, so no adapter ran on that path and
any fix made in one would have been dead code.

Measured against a fresh migrated Postgres before changing anything: session
delete published 0 events, while a `CancelTask` on a live session in the same
harness published 1.

`BroadcastCancelledTasks` now takes the workspace from the caller, which has it
in hand at all three call sites. Runtime teardown had the same bug —
`DeleteSystemAgentsByRuntime` cascades the sessions away — so it is fixed with
it. Revoke goes through `ArchiveAgentsByRuntime`, which only archives, so it was
never affected.

With the event published, the indicator close-out works from the binding-free
state as intended.

## Blocker 1 — the notice now fires for an empty owned batch

`taskMayOwnProcessingAck` replaces the straight `TaskInputIsChannelIngested`
call on the cancel path. A batch with no user messages is not a verdict either
way, so an unclassifiable task defers to the promise, which is the more direct
evidence. The existing test set `channelIngested = true` and so never reproduced
the #6611 shape; the new one drives the empty batch.

## Blocker 3 — one notice per bulk cancel

Taking the promise is what bounds it: the event arrives once per row, the first
one takes the promise and the rest find nothing outstanding.

## Two things you should know rather than discover

**Dropping the provenance gate on the cancel path caused a regression, and it is
back.** For a while this branch keyed the withdrawal only on "is an ack
outstanding", which meant cancelling a *web-origin* task in a session with a
running channel turn posted "no reply is coming" — and then the channel turn
delivered its reply. The gate is restored, ahead of taking the promise rather
than after it, and pinned by a test.

**A narrow residue remains, deliberately.** The claim guard in `handler/daemon.go`
cancels a task whose user text cannot be read, and that task may be a web run.
Deferring to the promise then posts the notice for a run the room never asked
about. The guard fires only on genuinely corrupt state by its own account, so it
is written down in the comment rather than papered over.

**And the DingTalk decision is DB-derived again.** An earlier revision moved it
into process memory and called that parity with Slack and Lark. It is not: those
two keep state in memory because nothing in the database records which message a
reaction sits on, while DingTalk's answer was in the database all along.

## Also

The promise is now discharged on every terminal path — including one nobody had
listed: an ending that arrives after the session was archived, which removes the
binding without cancelling what is running. That promise used to sit on record
until an unrelated later cancel spent it.

Six comments named `broadcastTaskEvent` as the publisher of a cancel, which
stopped being true when `BroadcastCancelledTasks` began publishing directly, and
the two claiming "every cancel path lands here" were false about the one path
three of these tests are named after. Both rewritten. The PR body's claim about
WeCom subscribing to `task:cancelled` was also wrong at this SHA and now carries
a correction.

Nine checks green. Every new test reverse-verified — backing the production
change out fails it, restoring it passes.

Bohan-J (changes requested):
Re-reviewed at `450eacb3`. The three things I flagged last time are properly closed, and one of the fixes went further than what I asked for:

- the DingTalk notice no longer hangs on `TaskInputIsChannelIngested` alone — `taskMayOwnProcessingAck` treats an empty batch as "not a verdict" rather than as a web run, which is the #6611 shape;
- Slack and Lark carry `installation_id` in their tracked state, so a clear outlives the binding the session delete drops;
- `takeOutstandingAck` holds a bulk cancel to one message.

`BroadcastCancelledTasks` taking the workspace from its caller is a genuinely good find. A deleted session's cancels were resolving their workspace through the very row the transaction had just dropped, so `publishTaskEvent` discarded them before the bus — those events were reaching nobody at all, not just the channel indicators. Worth calling out in the PR description; it is a bigger fix than the indicator work around it.

Still requesting changes on two. I have a runnable repro for each against `450eacb3`.

## 1. Blocking — the DingTalk promise is one slot per session, so two rounds discharge each other's

`ack.go:159` writes `outstanding[key]`: one slot per session. `outbound.go:151-153` releases whatever is in that slot whenever a channel run ends. Two rounds in one conversation share it, and the coalesce window explicitly allows two rounds — `ackCoalesceWindow` is sized so "a genuinely later turn re-acks".

The trace:

1. Round A is ingested, ack posted, `outstanding[S] = promiseA`.
2. Round B is ingested 10s later while A is still running. Past the window, so it posts its own ack — and overwrites the slot with `promiseB`. Two "👀 On it" messages now stand in the room and only one is on record.
3. A finishes and delivers its reply. The delivery path releases the slot — which by now holds B's promise.
4. B is cancelled. `hasOutstandingAck` is false, so nothing is posted. B's ack stands in the conversation for good.

Step 2 is already wrong on its own: from the moment B acks, A's promise is gone, so if A is the run that gets cancelled the room gets one notice covering two open promises.

Repro — drops into `outbound_test.go` as-is:

```go
func TestOutbound_InterleavedRoundsLoseTheSecondPromise(t *testing.T) {
	const taskA = "33333333-3333-3333-3333-333333333333"
	const taskB = "44444444-4444-4444-4444-444444444444"

	d := newDingtalkSendServer(t)
	o, ack, q := newCancelTestOutbound(t, d)

	base := time.Now()
	now := base
	ack.now = func() time.Time { return now }

	postProcessingAck(t, ack)          // round A promises a reply
	now = base.Add(10 * time.Second)
	postProcessingAck(t, ack)          // round B, past the window, promises its own

	bus := events.New()
	o.Register(bus)

	asChannelRun(q)
	bus.Publish(chatDoneEvent(taskA, "here you go"))   // A ends, reply delivered
	afterA := atomic.LoadInt32(&d.sendCalls)

	bus.Publish(cancelledEvent(taskB))                 // B cancelled, ack still standing

	if n := atomic.LoadInt32(&d.sendCalls) - afterA; n != 1 {
		t.Fatalf("round B was cancelled while its own ack was still standing in the room; "+
			"cancellation notices posted = %d, want 1", n)
	}
}
```

```
--- FAIL: TestOutbound_InterleavedRoundsLoseTheSecondPromise (0.00s)
    round B was cancelled while its own ack was still standing in the room; cancellation notices posted = 0, want 1
```

The current tests cannot see this: `postProcessingAck` is called at most once per test, so the map never holds a second round.

On the fix — the promise needs a round identity, and A's ending must only discharge A. There is no task id available at ack time (the ack is posted on ingest; the task is enqueued later by the debounced flush), so the cheapest shape that is actually correct is probably a FIFO queue per session instead of a slot: turns in a session are serialized, so endings arrive in the order the acks were made — an ending pops the oldest promise, a cancel pops the oldest and posts it. Worth noting that if you go that way the bulk-cancel dedupe stops being emergent (today it comes from the slot being empty after the first take) and has to become explicit, e.g. coalescing the notices for one broadcast rather than relying on there being exactly one promise.

## 2. Real, but pre-existing — a run that ends while the indicator is still being added leaves it on for good

The Router calls `OnIngested` in a detached goroutine (`channel/engine/router.go:226-232`), and all three adapters record their state only after the external call returns (`slack/typing_indicator.go:118-122`, `lark/typing_indicator.go:117-124`, `dingtalk/ack.go:143-160`). If the run ends first — cancelled, session deleted, or just a fast reply — `Clear` finds nothing and returns, then the add lands and nothing will ever take it off.

Repro (Slack; same shape in the other two): a `reactionAPI` fake whose `AddReactionContext` blocks until released, `Add` driven on a goroutine, `Clear` called once the add is in flight, then the add released.

```
--- FAIL: TestTypingIndicator_ClearBeforeAddLeavesTheReactionOn (0.00s)
    the run ended before the reaction landed: it is on C1/1786363259.000100 with nothing
    left to take it off (removes = 0, still tracked = 1)
```

This window is on `main` for chat-done and task-failed too, so it is not a regression from this PR — your call whether it lands here or as a follow-up. Flagging it because the PR's claim is that every ending now closes the indicator, and this is the hole in "every". A per-session generation counter is enough: `Clear` bumps it, `Add` captures it before the call and, if it is stale when the call returns, immediately undoes what it just did instead of recording it. It can be driven deterministically the way the repro above does, no sleeps.

## Non-blocking

- Archiving an agent still leaves indicators up. It is now documented in three places, which is more than it had, but a comment does not remind anyone — worth its own issue.
- `releaseIfThisRunWasAcked` leaves the promise in place when the batch reads empty, so an ending arriving after the session's messages are gone leaves it standing until the 24h sweep. That is the safe direction, just noting it is deliberate.

seacen:
@Bohan-J — this PR is now smaller than the one you reviewed. The DingTalk
cancellation notice is out; the broadcast fix and the Slack/Lark close-out stay.
Head is `05ed347a5`, nine checks green.

## Why

Both of your remaining items are addressed, but not in the same way.

**Item 1 (one ack, two rounds) is not fixed here — it is withdrawn.** Your repro
was right, and so was the diagnosis under it: the promise needed a round identity.
We built the FIFO queue you suggested. It closed your case and opened another —
a promise that never drains sits at the head and silences every later cancel in
that conversation. We fixed that; the fix then spent the room's only promise on
the silent branch, so a shared ack cancelled twice said nothing at all. Three
attempts, three shapes of the same failure: a path that stays quiet while still
holding undelivered news.

The common cause is not the container. DingTalk has to POST a message to close
its indicator, so it has to decide whether a reply is still coming — and we kept
trying to answer that from process memory, which cannot answer it. The ack is
posted on a detached goroutine, so an ending can arrive before the promise is
even recorded; a restart loses the lot.

So it comes back as its own PR with a different contract: **post the notice
unconditionally on a cancel and accept the occasional duplicate.** A second
"that run was cancelled" is a much smaller harm than silence, and it deletes the
promise ledger rather than patching it. Nothing about that belongs in this PR.

**Item 2 (Clear before Add) stays open and is now explicitly out of scope.** You
called it pre-existing and left the placement to us; with DingTalk gone it is no
longer this PR's trigger for anything.

## What is left, and the bigger half is first

**A deleted session's cancels were reaching nobody at all.**
`DeleteChatSession` drops the `chat_session` row inside the transaction and
broadcasts after the commit, so `ResolveTaskWorkspaceID` could no longer resolve
a workspace and `publishTaskEvent` returned early. Measured against a fresh
migrated Postgres before changing anything: session delete published 0 events, a
`CancelTask` on a live session in the same harness published 1.
`BroadcastCancelledTasks` now takes the workspace from its caller, which has it
at all three call sites. Runtime teardown had the same bug — `DeleteSystemAgentsByRuntime`
cascades the sessions away — so it is fixed with it. Revoke goes through
`ArchiveAgentsByRuntime`, which only archives, and was never affected.

You spotted this was the larger fix; it is now what the PR leads with.

**Slack and Lark close out from their own state.** Both carry `installation_id`
in the tracked state, so a clear outlives the binding a session delete drops. The
empty-batch cancel is covered for both.

## Verification

The DingTalk revert is byte-exact: the whole `integrations/dingtalk` tree hashes
identical to `upstream/main`, not just the five files, and `cmd/server/router.go`
with it. All six packages build, vet, gofmt clean and pass `-race` against a
freshly migrated database — DingTalk included, on main's code.

Blocker 2 is pinned by a probe against real Postgres rather than by reading: a
real installation and binding are inserted, the session is deleted, and the
indicator still clears for both adapters.

## Housekeeping

The commit list still reads `fix(dingtalk): …` three times with a revert on top.
The tree is clean; say the word if you would rather it were squashed before
merge.

Bohan-J (changes requested):
Re-reviewed at `05ed347a`. Scope is now Slack + Lark + the broadcast fix — I confirmed `server/internal/integrations/dingtalk` has a net-zero diff against base. Taking it out was the right call rather than a retreat: the promise needed a round identity it cannot get at ack time, and that is a design change, not a fix.

Where my last two landed:

1. Gone with DingTalk.
2. The add/clear race is documented in the Lark `Register` comment and deferred. Agreed — it predates this PR and I said it was your call. One ask: give it a tracked issue rather than only a comment, since nothing in the code will surface it again.

One blocker left, with a repro.

## Runtime teardown deletes the installation the clear needs, in the transaction that cancels the tasks

`handler/runtime.go:824` runs `DeleteChannelInstallationsBySystemRuntimeAgents` inside the teardown transaction, and `publishRuntimeTeardown` broadcasts the cancels after it commits (`runtime.go:848-854`). The event now reaches the manager — that part this PR fixed — but `Clear` still has to turn `state.InstallationID` into credentials (`slack/typing_indicator.go:178`, `lark/typing_indicator.go:200`) and the row is gone. The state has already been taken off the map by then (`slack/typing_indicator.go:139-142`, `lark/typing_indicator.go:147-149`), so there is nothing to retry from and the 👀 stays on the message.

Repro against `05ed347a`, drops into `slack/typing_indicator_test.go` as-is:

```go
func TestTypingIndicator_TeardownDeletedInstallationLeavesTheReactionOn(t *testing.T) {
	sessionID := uid(7)
	q := &fakeOutboundQueries{
		inst: db.ChannelInstallation{ID: uid(1), Status: "active", Config: slackInstallConfigJSON()},
	}
	fr := &fakeReactor{}
	m := newTestTyping(q, fr)

	ts := freshTS()
	m.Add(context.Background(), db.ChannelInstallation{ID: uid(1), Config: slackInstallConfigJSON()}, sessionID, "C1", ts)
	if len(fr.added) != 1 {
		t.Fatalf("setup: the reaction should be on the message; added = %d", len(fr.added))
	}

	q.instErr = pgx.ErrNoRows // the teardown transaction has committed

	bus := events.New()
	m.Register(bus)
	bus.Publish(events.Event{
		Type:          protocol.EventTaskCancelled,
		TaskID:        util.UUIDToString(uid(9)),
		ChatSessionID: util.UUIDToString(sessionID),
		Payload: map[string]any{
			"task_id":         util.UUIDToString(uid(9)),
			"chat_session_id": util.UUIDToString(sessionID),
			"status":          "cancelled",
		},
	})

	if len(fr.removed) != 1 {
		t.Fatalf("the runtime was torn down and its tasks cancelled, but the 👀 is still on C1/%s "+
			"with nothing left to take it off (removes = %d)", ts, len(fr.removed))
	}
}
```

```
WARN slack typing indicator: resolve installation for clear failed err="lookup installation: no rows in result set"
--- FAIL: TestTypingIndicator_TeardownDeletedInstallationLeavesTheReactionOn (0.00s)
    the runtime was torn down and its tasks cancelled, but the 👀 is still on C1/1786385760.000100 with nothing left to take it off (removes = 0)
```

I am not raising this as scope creep, because the PR claims this path. The description says the runtime teardown "has the same shape … so the session is gone before the broadcast there too", and the `Register` comments list "the runtime and member revocations" among what now arrives. On that path the event arrives and the indicator still does not come off, which is the outcome the PR is about. Member revocation is genuinely fine — it archives agents and leaves installations alone, and `DeleteChannelInstallationsBySystemRuntimeAgents` has exactly one call site.

Either resolution works for me:

- **Carry an encrypted config snapshot in the state** and use it when the row is gone. `Add` already holds `inst.Config`, and it is the encrypted blob, so the "no decrypted token between add and clear" invariant is untouched and this path needs no DB read at all. I would keep the live lookup as primary and fall back to the snapshot only on `pgx.ErrNoRows` — a snapshot used unconditionally would miss a credential rotation between add and clear, which today's lookup picks up.
- **Or narrow the claim**: drop the teardown from the covered list in the description and the two `Register` comments, and track it separately. The residue is a stale badge after an admin deletes a runtime, which is far narrower than the paths this PR does fix.

What does not work is the current pairing: the claim as written plus the behaviour as built.

On tests, the two deterministic ones asked for are the right shape — installation lookup returning `pgx.ErrNoRows`, driven through a real bus, for both adapters. The repro above is one of them.

Verified locally at this head: `go test ./internal/integrations/slack/ -race -count=1` and the Lark typing/patcher regressions both pass.

seacen:
Fixed at `94ae2d8d1`, and the tracked issue you asked for is
https://github.com/multica-ai/multica/issues/6735.

## The teardown blocker

Your repro was right and your first remedy is what I took. Each tracked state now
carries the installation as it stood when the reaction was added, and the clear
falls back to it when the lookup comes back with no rows.

Three things I kept from your framing, because each of them is load-bearing:

**The live lookup stays primary.** A snapshot cannot see a credential rotation
between add and clear; the lookup can. So the snapshot is consulted only where
the row is genuinely gone.

**A lookup that merely errored is still refused.** `pgx.ErrNoRows` says the
installation is gone. A connection reset says nothing about whether it exists,
and clearing through a possibly-rotated secret is worse than leaving the badge
for the next ending. Both adapters have a test for that, because the natural
mistake here is to treat every lookup failure as "gone".

**The invariant is untouched.** What the state holds is the same encrypted blob
the database holds — Slack's `inst.Config`, Lark's `Installation` — and
decryption still happens at clear time. No decrypted secret lives in the state
map, before or after.

Four tests, two per adapter: your repro, and its guard. Backing the fallback out
fails the first in both packages, checked by exit code rather than by reading the
output.

One comment went with it. `Clear`'s doc said "Installation rows survive the
session", which this change proves only half true — they survive a session
delete and not a runtime teardown. It now says which.

## Where the other two landed

**The add/clear race** is now issue #6735, with the generation-counter shape you
suggested and the deterministic way to drive it written down. Agreed that a
comment reminds nobody.

**DingTalk** stays out. It will come back with the contract the revert commit
describes — post the notice unconditionally on a cancel and accept an occasional
duplicate — which deletes the promise ledger rather than patching it.

## Thank you for the repros

Four rounds on this PR, and every one of them came with code that dropped into
the test file and failed. That is the difference between a review round and a
guessing round, and it is why the DingTalk half is out rather than shipped
broken.

Bohan-J (approved):
Final pass at `94ae2d8d`. The blocker from my last round is closed and I found nothing new that should hold this up — approving, which supersedes my three earlier CHANGES_REQUESTED.

The fix is the shape I asked for: the live lookup stays primary, and the snapshot is consulted only when the row is genuinely gone. Keeping that ordering matters more than the fallback itself — a credential rotated between add and clear still resolves through the live row.

What I checked myself rather than reading off the diff:

- The repro I posted last round — reaction added, `GetChannelInstallation` returning `pgx.ErrNoRows`, cancel published on a real bus — now passes. The reaction comes off.
- I wrote the negative case before looking at yours: a transient lookup error (`context.DeadlineExceeded`, `connection reset`) must not be mistaken for a deleted row. It refuses and posts nothing, on both adapters. Good that you covered it too.
- `go build ./...`, `go vet` on slack / lark / handler / service, `go test ./internal/integrations/slack/ -race -count=1`, and the Lark typing/patcher regressions with `-race` — all clean at this head. CI is green.
- All four `BroadcastCancelledTasks` call sites pass a real workspace. Worth checking explicitly, because the signature change moves the failure mode from "the resolution returns empty" to "a caller passes empty": `chat.go:635` and `:763` take it from the authenticated request, `runtime.go:853` and `workspace_revoke.go:249` from the teardown's own workspace, which is the same value their sibling publishes already use.

Nits, none blocking:

1. `typing_indicator.go:166-178` (Slack) and `:184-194` (Lark) memoise per `installation_id` and keep the **first** snapshot seen for that id. If a session holds two reactions added either side of a credential rotation and the row is then deleted, the older snapshot is the one used. Taking the last would be strictly better, and the triple-coincidence needed to hit it is why this is a nit rather than a finding.
2. `TestTypingIndicator_ALiveInstallationIsPreferredOverTheSnapshot` doesn't test what it is named — it asserts the transient-error refusal, which `TestTypingIndicatorDoesNotFallBackOnATransientLookupFailure` already covers on the Lark side. Nothing pins the property the whole design rests on: with the row present and its config *different* from the snapshot, the live one wins. That is easy to make real — have `newAPI` record the token it was handed and assert it came from the live row — and it is the test I would most want to still be there in a year.
3. Two doc details: the Slack manager comment still says "only the installation id is held in the map between add and clear" (`typing_indicator.go:79-80`), which the snapshot has outdated — the `typingState` doc right below it is accurate. And Lark keeps the whole `Installation` in state where Slack keeps just the config blob; Slack's shape is the tighter of the two.

Last thing, and it is the only one I would actually chase: the gaps this PR knowingly leaves — the DingTalk cancellation notice, agent archive not broadcasting per row, the add/clear race, and no TTL or restart recovery — are documented in code comments and nowhere else. Comments do not surface again. Worth four issues so they can be scheduled rather than rediscovered.

Nice work on the diagnosis in the middle of this, particularly the `BroadcastCancelledTasks` workspace resolution. That one was silently dropping every `task:cancelled` for a deleted session, for every consumer, and it was not what anyone set out to fix.

Bohan-J:
Thanks for the four rounds on this one — merging now.

Two things stood out. Cutting DingTalk out once it was clear the promise needed a round identity it could not get at ack time was the right call; shipping a half-correct version of that would have been worse than the bug. And the `BroadcastCancelledTasks` workspace resolution was a real find that nobody asked for: a deleted session's cancels were being dropped before the bus, so every consumer was blind to them, not just the channel indicators.

The nits from my last pass are non-blocking and can ride along with something else: preferring the newest snapshot in the per-installation memo, giving `TestTypingIndicator_ALiveInstallationIsPreferredOverTheSnapshot` an assertion that matches its name, and the stale "only the installation id is held in the map" comment.

The one ask that is worth doing soon: the gaps this PR deliberately leaves — the DingTalk cancellation notice, agent archive not broadcasting per row, the clear-before-add race, and no TTL or restart recovery — are recorded only in code comments right now. Four issues would keep them schedulable instead of waiting to be rediscovered.

PR #6632, opened 2026-08-08
feat(issues): auto-link external pull requests from issue content

Willam2004:
## Summary

Issues often reference pull requests that live on a code host other than the built-in GitHub integration. This PR adds **generic link tracking** for such external PRs, starting with the Alibaba Code platform (`code.alibaba-inc.com`):

- **Auto-link:** Code review links in issue titles, descriptions, comments and `pr_url` metadata are detected and stored in a new `external_pull_request` table, then shown in the issue PR list alongside built-in GitHub/self-hosted VCS PRs (with a provider badge; links degrade gracefully as "status unknown").
- **Write paths:** comment creation and metadata updates link fresh content and publish realtime events even when insertion partially fails.
- **Backfill:** historical issue content is scanned on list requests, throttled to one scan per issue per 15 minutes so page visits do not repeatedly load thousands of comments.
- **Lifecycle:** workspace deletion sweeps external PR rows (index-friendly `workspace_id` predicate); `created_by_type` is `NOT NULL` with a CHECK constraint; UTF-8-safe truncation helper for bounded text storage.

## Scope note

This PR is intentionally limited to **link tracking and display**. Live status synchronization (MR state, mergeability, comment counts) is deliberately excluded here — it will be proposed separately behind a host-agnostic provider interface so that no host-specific CLI is baked into the daemon.

**Follow-ups tracked:**
- Live status sync behind a host-agnostic provider interface
- An unlink endpoint to remove incorrectly auto-linked entries (auto-linking scans free text, so false positives are possible)

## Changes

- Migrations `265` (table) and `266` (unique index, `IF NOT EXISTS`)
- `server/internal/handler/github.go`: link extraction, discovery throttle, response mapping
- `comment.go` / `issue_metadata.go`: auto-link on write paths
- `workspace_delete.sql`: sweep + manifest classification
- `util.TruncateUTF8`: rune-boundary-safe truncation
- Frontend: `pull-request-list.tsx` aggregation/badges, schemas/types, locale strings (en/ja/ko/zh-Hans)

## Verification

- `go build ./...`, `go vet ./...` clean; handler/daemonws/migrations/util test suites pass
- Migration lint passes (unique prefixes post-148)
- `pnpm typecheck` (6/6 packages) and views test suite (3735/3735) pass
- Running in our self-hosted deployment against real issues

## Review history

This branch went through five rounds of automated code review (open-code-review); all findings were addressed before the scope was narrowed to link tracking (the sync machinery reviewed in earlier rounds is no longer part of this diff).


Bohan-J:
Thanks for this, and specifically for raising the vendor-coupling question yourself in the PR description — that's the right instinct and it makes this much easier to answer.

**Short version: Multica already has the subsystem this should plug into, so we'd like to redirect this work rather than merge it in its current shape.**

## The existing seam

`server/migrations/216_vcs_integration.up.sql` added per-workspace, token-based connections for self-hosted Git providers (currently Forgejo, Gitea, GitLab):

- `vcs_connection` — `instance_url` plus secretbox-encrypted access token and webhook secret
- `vcs_pull_request` / `issue_vcs_pull_request` — mirrored MRs and their issue links
- `server/internal/handler/vcs_webhook.go` — inbound webhooks with per-provider signature verification
- `server/internal/integrations/vcs/vcs.go` — a `Provider` interface plus a registry; `gitlab.go` and `forgejo.go` are the two adapters

Worth noting that `vcsPullRequestToResponse` already maps onto the same `GitHubPullRequestResponse` shape your `externalPullRequestToResponse` targets — the two implementations converge one function apart.

Going through that path would also hand you a few things this PR either re-solves or leaves open:

- credentials configured and encrypted per workspace, rather than depending on whichever account `a1` happens to be authenticated as on a given runtime host
- `issue_vcs_pull_request.reference_only`, which stops a drive-by link in a comment from becoming a first-class sidebar entry — this PR has no unlink path at all (`DeleteExternalPullRequest` is generated but never routed, so a mistaken paste is permanent)
- no new table, no new provider enum, no schema-level coupling to a single host

## Two questions

**1. Is `code.alibaba-inc.com` GitLab-API compatible?** The `!` MR convention, the `/codereview/<id>` URLs and the `a1 repo mr view/status/comment` command shape all suggest GitLab lineage. Our adapter needs `/api/v4` with a `PRIVATE-TOKEN` header, and `Merge Request Hook` webhooks authenticated via `X-Gitlab-Token`. If your instance speaks that, **you may need no code at all** — configure a `gitlab` connection pointed at your `instance_url` and it should work as-is.

**2. If it isn't compatible — can it send webhooks out to your Multica deployment?** If yes, the supported route is a new adapter under `server/internal/integrations/vcs/`: implement the five-method `Provider` interface, register it in `init()`, and add the value to the `provider` CHECK constraints. That's roughly one file, and it's vendor-neutral by construction, because the instance URL and credentials are workspace configuration rather than compiled-in constants.

If the answer to both is no — the instance is unreachable from your Multica deployment *and* you can't configure outbound webhooks — then there isn't an upstream shape that works here, and a private patch is the honest answer. Our integration model assumes the two systems can reach each other in at least one direction.

## Why we're closing this one

Primarily the above: it's a parallel implementation of an existing subsystem, and the correction isn't a revision of this branch.

Two smaller findings worth passing on, since both affect every deployment rather than only yours:

- the PR-sidebar empty-state string was changed in all four locales to mention "Code platform MRs", so every user reads a hint about a host they have no way to use
- `packages/views/issues/components/issue-detail.tsx` drops the `githubSettings.prSidebar` check, making the PR section unconditional — workspaces that deliberately turned it off get it back with no way to disable it again. Your reasoning there was sound (the section is no longer GitHub-only), but the fix would be to render it when the setting is on **or** another source has entries, rather than removing the gate entirely.

For what it's worth, the engineering standard here is high — the migration discipline (no FKs, `CONCURRENTLY` indexes), the UTF-8 truncation handling, the workspace scoping and the process-timeout work all read like someone who studied the codebase's conventions rather than guessing at them. That's why it's worth being precise about the redirect instead of just declining.

Closing this one, but genuinely happy to review a PR on the `vcs_connection` path. And if the answer to question 1 turns out to be "yes, it's GitLab-compatible", say so here and we'll help you get the connection configured instead — that outcome costs you no code at all.

PR #6636, opened 2026-08-08
HAR-14: instrument acquisition-to-activation funnel

hardy4yooz:
## Summary

- instrument `qualified_landing_view` and `signup_or_download_start` across the public landing and Desktop download CTAs, with visible-human qualification and tab-session deduplication
- persist bounded first-touch source/medium/campaign/referrer-host attribution for new accounts, rejecting URL-, email-, and IP-shaped values
- add a sqlc-validated canonical workspace event stream from `workspace_created` through `second_issue_assigned`, including retry deduplication and test-workspace exclusion
- document the event dictionary, dashboard/HogQL/PromQL definitions, privacy rules, validation procedure, known blind spots, and Product/Data sign-off checklist

## Validation

- `pnpm typecheck`
- `pnpm test`
- `pnpm lint` (passes; existing `apps/web/app/(auth)/login/page.tsx:137` hook warning remains)
- `go test ./...` using Go 1.26.1
- exact web funnel tests: 9 passed
- applied migrations 001-265 to an isolated PostgreSQL 17/pgvector database
- live `ListLaunchFunnelEvents` integration test verifies all seven durable milestones, retry deduplication, attribution propagation, nil exclusion lists, and test-workspace exclusion

## Sign-off required

Product/Data should approve the 3-second qualified-visit definition, 7-day activation window, 14-day retention window, first-task assignment semantics, attribution retention policy, and maintained excluded-workspace list before launch validation.

Closes HAR-14


Bohan-J:
Thanks for the thorough work here — the dimension sanitization, the per-issue deduplication in the funnel query, and the test coverage are all genuinely well done. We're going to close this one, though, and I want to be specific about why so it isn't wasted signal.

**The metric definitions are ours to settle, and this PR settles them.** The 3-second qualified-visit threshold, the 7-day activation window, the 14-day retention window, and treating the first `in_review` transition as the activation edge are product decisions. The description asks Product/Data to sign off on them after the fact, but that's the wrong order — once those numbers land in code and docs they anchor how we read growth. We need to agree on the definitions first and instrument second, rather than review an implementation that has already chosen them.

**Migration 265 crosses a data boundary we don't want to cross for this.** `user.acquisition_attribution` is written unconditionally on every account creation, in every deployment. An empty `POSTHOG_API_KEY` stops third-party events but does not stop this local persistence, so self-hosted instances would begin durably storing first-touch marketing attribution on their users — with no opt-in, no retention period, and no deletion path. They get nothing back for it either: the landing-page events that give the column meaning only fire on our own marketing site. Adding a permanent user attribute to every deployment to serve one hosted dashboard isn't a trade we want to make.

**The work wasn't scoped.** There's no open issue requesting an acquisition funnel, so there was no agreed definition for this to build against.

None of this is a knock on the execution. If you'd like to contribute again, opening an issue to agree on scope first — especially for anything touching migrations, the signup path, or persisted user data — will get a much better return on the effort. Happy to talk it through there.

PR #6637, opened 2026-08-08
WS-8: wake squad leaders on child review delivery

duochifan2003:
Closes WS-8

## What changed

- Treat child `in_review` as a delivered terminal status for stage barriers.
- Reuse the existing system handoff comment, squad-leader enqueue, provenance, and pending-task deduplication paths.
- Add stage-barrier coverage for partial and complete `in_review` stages, plus a squad-parent/direct-agent integration regression covering leader-task provenance and duplicate suppression.
- Update the built-in squad and issue workflow skills and source maps.

## Root cause

Child work normally reaches `in_review`, but the shared `isTerminalChildStatus` predicate only recognized `done` and `cancelled`. `TaskService.CompleteTask` intentionally does not own issue status, so the parent wake path never ran.

## Validation

- `go test ./internal/handler -run 'TestStageBarrierClosed|TestChildInReviewWakesParentSquadForDirectAgentChild' -count=1` (package compiles; DB-backed tests were skipped because this checkout has no PostgreSQL)
- `git diff --check`
- `make test` could not start: `.env`, Docker, and PostgreSQL are unavailable in the runtime.

### DIFF
--- a/server/internal/handler/issue_child_done.go
+++ b/server/internal/handler/issue_child_done.go
@@ -14,7 +14,8 @@ import (
 )
 
 // notifyParentOfChildDone posts a top-level system comment on the parent
-// issue when a child issue transitions from non-done into done. This replaces
+// issue when a child issue transitions from a non-terminal status into a
+// delivered terminal status. This replaces
 // the agent-prompt rule that previously made child agents post the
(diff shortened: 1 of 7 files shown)


duochifan2003:
按工作区决定撤销此 PR；改动保留在分支中，未删除分支。

Issue #6417, opened 2026-08-05
[Feature]: Task commits prematurely when spawning sub-agents or running cron expressions in child processes


liuwenhui2026:
### Deployment type

Official App

### What do you want and why?

Description
There is an issue where the current task commits immediately when spawning multiple sub-agents or executing cron expressions via child processes. The root cause is that the claude -p mode prematurely considers the task as completed and terminates, while the spawned child processes are still running in the background.

Steps to Reproduce
Execute a task that spawns multiple sub-agents via child processes
Alternatively, run a task with cron expressions that launch background processes
Observe that the main task commits and exits immediately after spawning
Expected Behavior
The main task should wait for all child processes/sub-agents to complete before committing and terminating.
Actual Behavior
The claude -p mode treats the task as finished once the spawn command is executed, without waiting for child process completion. The child processes continue running in the background but are no longer tracked by the parent task.

Impact
Background processes become orphaned and unmonitored
Task completion status is inaccurate
Sub-agent results are lost or not captured
Cron-based workflows break due to premature termination

### Proposed solution (optional)

Possible Root Cause
The process management in claude -p mode appears to lack proper child process tracking/waiting mechanisms, causing it to exit as soon as the parent process finishes spawning rather than waiting for the entire process tree to complete.

### Screenshots / mockups (optional)

<img width="1804" height="1448" alt="Image" src="https://github.com/user-attachments/assets/de0852a2-3d50-4eff-a4aa-7cfa9e296315" />

<img width="2476" height="1428" alt="Image" src="https://github.com/user-attachments/assets/f32b3bf6-e12c-44c1-94ef-41669ab0b765" />

Bohan-J:
Thanks for the detailed report — the screenshots were enough to trace exactly what happened, and it turned out to be something different from the diagnosis in the title.

**What actually happened in your run**

The agent never spawned an OS child process. From the execution log, it used Multica's sub-issue delegation: it created a sub-issue and assigned it to another agent. That is a one-shot enqueue — creating a sub-issue with a non-backlog status starts exactly one run for the assignee. The parent run then posted its comment, moved the parent issue to `in_review`, and exited normally. Nothing was orphaned, and nothing was left running in the background.

So there are two separate problems in what you saw, and both are real:

1. **"Every 20s" never had an executor.** A sub-issue is a single run, not a schedule. Once the sub-agent's one run ends, nothing re-triggers it — which is why you only ever got one temperature reading.
[...]

Issue #6006, opened 2026-07-27
[Bug]: Slack and Lark replies include interim agent narration


yyclaw:
### Deployment type

Official App

### What happened?

When an agent handles a Slack or Lark channel message and uses tools, the outbound reply can include interim narration from the execution timeline together with the final answer. This makes channel replies noisy and can expose process details that should remain in the Multica transcript.

The channel should receive only the deliverable reply: the final text block after the last thinking, tool, or error timeline row. Legacy events and unavailable transcripts should continue to fall back to the existing completion content.

### Steps to reproduce

1. Connect a Slack or Lark channel to an agent.
2. Send a request that causes the agent to narrate, invoke one or more tools, and then provide a final answer.
3. Wait for the chat completion event to be delivered back to the channel.
4. Observe that the outbound message may contain both interim narration and the final answer.

### Screenshots (optional)

Not applicable.

### Additional context (optional)

The task transcript already persists ordered text and non-text rows. Channel delivery can derive the final text region from that timeline while retaining the current completion content as a compatibility fallback. The final reconstructed reply must be redacted after all persisted text fragments are joined so credentials split across flush boundaries cannot bypass redaction.

PR #6641, opened 2026-08-09
feat(agent): add Prime Agent as a native ACP provider

marcelocecin:
## Summary

- add Prime Agent as a native ACP provider in Multica
- integrate Prime with the existing agent discovery, runtime configuration and migration patterns
- preserve Prime's native ACP session semantics without implementing session resume
- keep the issue follow-up workdir when the daemon advertises checkout-keeps-work, without reporting a session id Prime cannot resume
- configure Prime to use its native per-task skills directory
- add the provider's frontend surface: runtime type, display entry, logo, and landing/docs copy, including French
- keep MCP injection out of Phase 1: `session/new` is sent with an empty `mcpServers` array

## Versions

Developed and smoke-tested against Prime Agent **v0.7.1**, which is also the floor in `agent.MinVersions`. Current stable release is **v0.9.5** (2026-09-16). `getAgentDir`, `ENV_AGENT_DIR`, `CONFIG_DIR_NAME`, `_resolveRlmMaxDepth`, and `isNonNegativeInteger` were diffed against that tag and against prime-agent `main` as of 2026-09-23:

- `agentCapabilities.loadSession: false`, one session per ACP connection, and no `session/resume` or `session/load` on the wire — the no-resume design holds
- `getAgentDir` changed once, in v0.9.4: it also expands a Windows `~\` value, which earlier versions leave relative to the cwd. The fail-closed `rlmMaxDepth` gate checks both readings
- `_resolveRlmMaxDepth`'s precedence chain is identical. v0.9.5 renamed the public subagent call to `await rlm.spawn(...)` (the old callable throws). That path still enters `_startRlmChildRun`, which refuses when `rlmDepth >= rlmMaxDepth`. The fire-and-forget note in `acp-mode.ts` is still present
- v0.9.0 raised the built-in default from 1 to 2. The gate is unaffected: it sets `RLM_MAX_DEPTH=0`, which outranks the default, and refuses the run only when a global `settings.json` `rlmMaxDepth > 0` outranks that env var
- from v0.8 the ACP mode rejects a non-empty `mcpServers` list when the connection has no MCP support, and v0.9.1+ advertises `mcpCapabilities: { http: true }` when it does. Multica sends an empty array

## Sync with main

Merged current `multica-ai/multica` `main` (`0516dd57`, including the v0.5.2 release and the commits after it). Textual merge was clean.

- The Prime runtime-profile migration is **545**. Prefix 535 collided with `535_github_pr_address_index`, and 536–544 are the duplicate-mark and task-supplement migrations. 441 (`codearts`) is still the newest family migration; the CHECK list is unchanged.
- French docs from #8757 now list Prime Agent the same way as English, Chinese, Japanese, and Korean (provider row, runtime-managed model, no resume, no Multica-managed MCP, minimum 0.7.1, `MULTICA_PRIME_ARGS`).
- Landing copy stays at 27 tools, including Prime Agent, in en/zh/ja/ko. French landing reuses the English dictionary.

## Validation

- `go test ./internal/migrations -count=1 -run 'TestMigrationNumericPrefixes|TestRuntimeProfileFamily'`
- `go test -tags=agentintegration ./pkg/agent -run '^$' -count=1` (compile)
- `go test ./pkg/agent -count=1 -run 'Prime|Deliverable'`
- real Prime Agent account smoke tests were not run
- `TestPrimeWindowsDescendantsDieWithTheOwnedProcessTree` is Windows-only and was not executed here

## Scope / Deferred follow-ups

Deliberately deferred:

- token/usage reporting beyond what the ACP usage accumulator already records
- improved stderr/authentication diagnostics
- handling of custom continuation flags such as `--continue`
- handling/diagnosis of Prime's cwd mismatch metadata

## Explicitly out of scope

- no `prime_home.go`, and no HOME overlay — Multica never sets the child's HOME; it only reads `HOME`/`USERPROFILE` from the child env to resolve Prime's settings path
- no daemon-level special-case that pretends Prime can resume a session
- no MCP injection in Phase 1
- no changes to `prime-agent/`


multica-eve (changes requested):
The product direction is reasonable: teams already using Prime Agent should be able to run it through Multica. Before this is mergeable, the adapter needs to preserve Prime's actual completion semantics, and the rest of the provider surface needs to be consistent.

Requesting changes for the following blockers:

1. **Background subagent work can be cut off and reported as completed.** `server/pkg/agent/prime.go:270-309` closes the session and process as soon as `session/prompt` returns. Prime Agent v0.7.1 explicitly documents its subagents as fire-and-forget and able to keep reporting after the spawning turn ends (`packages/coding-agent/src/modes/acp/acp-mode.ts:302-305`). Its headless-completion path waits for the parent session to become idle, not for `hasRunningRlmChildren()`, and emits child state only through `session_info_update._meta`, which the shared `hermesClient` currently ignores. A parent turn can therefore finish, Multica can record the task as completed, and the remaining child results are then discarded. Because this provider also disables resume, a later run cannot recover them. Please either explicitly disable subagents for this Phase 1 provider or track them to terminal state and include them in the completion condition. Add a regression test where the parent turn completes before a child reports back.

2. **The process tree is not cleaned up safely.** `server/pkg/agent/prime.go:86-87` does not create a controllable process group, while `:305-309` closes stdin, immediately cancels only the direct process, and waits for stdout/stderr EOF without a bound. This gives Prime's EOF -> `connection.dispose()` path no grace period to clean up Python kernels/tool children and cannot terminate descendants on cancellation or timeout. Please use the repository's existing process-group, graceful SIGTERM -> SIGKILL, `WaitDelay`, and bounded pipe-drain pattern, with fake-CLI coverage for normal completion, cancellation, and timeout while a descendant remains alive.

3. **Existing local Prime skills cannot be listed or imported.** The new `<workDir>/.prime/agent/skills` branch in `server/internal/daemon/execenv/context.go:423` handles injecting skills already bound to a Multica agent. `server/internal/daemon/local_skills.go:128-221` still has no `prime` root, so users' existing `~/.prime/agent/skills` inventory is reported as unsupported. Please add the provider root and list/import tests. Also add `prime` to `server/internal/daemon/execenv/sidecar_manifest_test.go:145-161`; that invariant explicitly requires every provider that writes into the workdir to participate in cleanup round-trip coverage.

4. **The custom-runtime allowlists disagree.** The backend and migration 265 accept `prime`, but `packages/core/types/agent.ts:108-129` omits it from `RUNTIME_PROFILE_PROTOCOL_FAMILIES`. As a result, the runtime-management UI neither shows Prime as a built-in protocol family nor allows selecting it for a custom profile. Please update the shared tuple and catalog/form tests.

5. **The published capability contract is still incorrect after adding Prime.** `apps/docs/content/docs/providers.mdx:21-68` does not list Prime and still says QwenPaw is the only runtime-managed-model exception and that all supported tools resume sessions. The MCP unsupported list also omits Prime. `environment-variables.mdx:212` does not document `MULTICA_PRIME_PATH` / `MULTICA_PRIME_ARGS`, and the landing pages still advertise 20 rather than 21 tools. Please update the English, Chinese, Japanese, and Korean install/capability/environment docs and landing-page lists, explicitly documenting no resume, no Multica-managed MCP, and runtime-managed model selection.

Non-blocking cleanup:

- `server/pkg/agent/prime.go:27-28` refers to nonexistent `okf/prime-agent/` and `REPORT.md` paths. Please replace these with versioned links to the actual `PrimeIntellect-ai/prime-agent` v0.7.1 sources.
- Keep provider names/errors/comments synchronized in `server/pkg/agent/agent.go` and `server/internal/daemon/config.go`. Add `prime: "Prime Agent"` to the daemon and frontend display-name maps; the current fallback renders only "Prime". A dedicated provider logo can remain a follow-up.

Validation performed on head `2c49b608`: all nine GitHub checks are green; local agent/execenv tests, the `ByteIdentical` invariant, targeted daemon tests, and `git diff --check` pass. Migration 265 is unique on current `origin/main` and adds no foreign key or cascade. I did not run the authenticated, quota-consuming `agentintegration` smoke tests.

marcelocecin:
@multica-eve @Bohan-J  @forrestchang  All requested changes are now addressed.

The PR is now based on the latest `upstream/main`, the migration prefix collision was resolved, the Prime provider was adapted to the upstream `Command` launch boundary, and the full CI is green.

Could you please take another look and re-review when convenient?

multica-eve (changes requested):
Thanks for addressing much of the previous review. The product direction remains reasonable, and the local-skill discovery, sidecar coverage, runtime-profile allowlist, capability documentation, display naming, source references, and migration collision are now substantially addressed. I re-reviewed the current head (`ba8a3d9`), but the following issues still block merge:

1. **RLM subagents are not reliably disabled.** `server/pkg/agent/prime.go:163-189,202` documents that Prime's global `settings.json` value for `rlmMaxDepth` takes precedence over `RLM_MAX_DEPTH=0`. A user who has configured Prime's normal global recursion setting therefore silently re-enables fire-and-forget children. Multica still closes the session after the parent `session/prompt` returns, so those child results can be discarded. This is the original completion-semantics blocker, not merely a P2 follow-up. Please make the Multica session immune to the global override, or track children to terminal state, and cover the global-override case with a regression test.

2. **The successful-completion EOF chain can still hang forever.** At `server/pkg/agent/prime.go:470-477`, the backend closes stdin and then waits without a bound for `readerDone` and `stderrDone`; `cmd.Wait()` is only reached later from the defer at `:341-345`. If any descendant keeps stdout or stderr open, EOF never arrives, the goroutine cannot return, the defer is unreachable, and `WaitDelay` never starts. With no task timeout this is permanent. Please add bounded teardown to the normal-success path and a fake-CLI regression where prompt/close succeeds while a descendant keeps a pipe open.

3. **Windows cancellation still kills only the direct child.** `prime.go:224` calls `cmd.Start()`, while `server/pkg/agent/proc_windows.go:45-49` explicitly says that only `startOwnedProcessTree` establishes Job Object ownership. Without that ownership, `signalProcessGroup` falls back to `cmd.Process.Kill()` at `proc_windows.go:247-258`, leaving IPython/tool descendants able to survive. Please design and test Windows whole-tree cleanup. Prime's shared/detached supervisor also needs to be accounted for so the solution does not terminate an unrelated shared process.

4. **Custom runtime `fixed_args` bypass Prime's protocol-argument protection.** The PR now allows `prime` as a custom runtime family, but `server/pkg/agent/launch.go:151-172` does not add `"prime": primeBlockedArgs` to `launchPrefixBlockedArgs`. A profile can therefore inject the otherwise-blocked `--mode` or `--cwd`; in particular, `--cwd` can move execution outside Multica's task workdir. Please add the mapping and launch-prefix regression coverage for both flags.

5. **The cancellation tests mask a real status-classification bug.** `server/pkg/agent/prime_cancel_unix_test.go:182-196` accepts `failed`, and `runPrimeCancellationTest` only waits for the PID file before cancelling (`:367-371`). The fake writes that file before it reads the initialize request, so cancellation may happen during the handshake; `prime.go:361-365` then unconditionally reports `failed` for an initialize error even when the context was cancelled or timed out. This is not proven to be only a scheduler race. Please synchronize the lifecycle tests on reaching `session/prompt` and assert the exact `aborted`/`timeout` outcomes, then separately fix and test handshake-time status classification.

Non-blocking: the landing card and FAQ now say 23 tools, but the runtime descriptions at `apps/web/features/landing/i18n/en.ts:104` and `zh.ts:104` still say 22.

Validation on `ba8a3d9`: all 14 GitHub checks pass. Locally, the Prime, daemon/local-skills, execenv/sidecar, migration-lint, `ByteIdentical`, repeated cancellation/timeout tests, and `git diff --check` pass. Those existing tests do not exercise the gaps above. I did not run the authenticated, quota-consuming `agentintegration` tests.

marcelocecin:
All five blockers are addressed; CI green on `2a06f7d`. The reasoning for each is in the commit messages rather than repeated here — `dca7283` (4), `c4a6951` + `5e36a7b` (5), `957355e` + `2f5aac5` (2), `2a06f7d` (3), `5c39a7d` (1).

Two things the diff won't show you, which I'd rather flag than have you find:

**The Windows test is compile-verified only.** I had no Windows host, and like the Codex case beside it, it isn't covered by the `windows-execenv` job's `-run` filters, so CI doesn't execute it either. Happy to widen that filter if you want it actually run — that's a workflow change I didn't want to make unilaterally.

**Owning the process tree rests on one claim I couldn't verify by running it:** that ACP mode never spawns Prime's machine-wide supervisor. It's only started via `ensureInteractiveDaemonRunning`, which the ACP path never reaches, and `DaemonClient.connect` only dials an existing socket. If there's a path I missed, owning the tree is worse than the status quo and should come back out — it's two lines.

Blocker 1 is the one I couldn't fix safely: `PRIME_AGENT_CODING_AGENT_DIR` would isolate the setting, but `auth.json` lives in the same directory, so redirecting it fails every run. The comment in `5c39a7d` records why, and the exposure is #6417's class rather than provider-specific.

multica-eve (changes requested):
Thanks for addressing four of the five blockers from the previous review. The launch-prefix filtering, handshake status classification, bounded successful teardown, and Windows process-tree ownership now have substantive fixes and regression coverage.

One must-fix remains before this can merge:

**`RLM_MAX_DEPTH=0` still does not reliably disable fire-and-forget subagents.** At `server/pkg/agent/prime.go:168-233`, the implementation now documents—but does not change—the upstream precedence: Prime reads the persisted global `rlmMaxDepth` from `~/.prime/agent/settings.json` before it reads `RLM_MAX_DEPTH`. A user who configured a non-zero global depth through Prime's supported `/rlm-max-depth <n> --global` command will therefore silently re-enable RLM children in Multica runs. Those children can continue after the parent `session/prompt` returns; Multica then reports the parent as completed and tears down the owned session before the child reaches a terminal state, losing or interrupting that work. This is the original completion-semantics blocker, not only a documentation gap or P2 follow-up.

Please make the Multica execution fail closed when an effective non-zero global override is present, add an ACP/per-session override that outranks the global setting, or track children through terminal state. Whichever approach is chosen needs a regression test with a persisted non-zero global `rlmMaxDepth`; the current default-environment tests cannot prove this boundary.

Non-blocking: `apps/web/features/landing/i18n/en.ts:104` and `zh.ts:104` still advertise 22 tools, while the updated list contains 23.

Validation on current head `2a06f7dc`: the targeted Prime tests, the same tests under the race detector, the execenv `ByteIdentical` invariant, and `git diff --check` pass; all GitHub checks are green. The PR is currently reported as conflicting with the base branch, which will also need resolution.

marcelocecin:
@multica-eve @Bohan-J @forrestchang — the remaining blocker and the two follow-ups are addressed. CI green on `da82cf3`.

**Global `rlmMaxDepth` override** (`da82cf3`) — you were right that documenting it wasn't enough. Of the three options you offered, the per-session override needs an upstream change and tracking children to terminal state has no ACP affordance, so this fails closed: `Execute` refuses to launch when an effective non-zero global override is present, naming the settings file and the command that clears it. A task that can't report its own completion honestly shouldn't start.

The gate deliberately mirrors Prime's own — `typeof value === "number" && Number.isSafeInteger(value) && value >= 0` — because anything failing that check falls through to `RLM_MAX_DEPTH`, which is already 0. Refusing those would break setups that work today, so a missing file, malformed JSON, absent key, quoted number, fractional or negative value all let the run proceed. That distinction is load-bearing: my first version decoded into `json.Number`, which accepts a quoted `"2"` and would have refused hosts Prime itself runs safely. `TestPrimeFailsClosedOnGlobalRlmMaxDepth` covers all nine cases, including the persisted non-zero global you asked for.

Worth stating plainly: this is a visible behaviour change. An operator with a global depth configured cannot run Prime tasks from Multica until they set it to 0.

**Base branch conflict** — resolved by merging `upstream/main` (`a60ce92`). Two things the automatic merge would not have caught:

- #7035 landed `342_runtime_profile_add_mcode`, colliding with our `342_runtime_profile_add_prime`. Both drop and rebuild the same CHECK constraint, and ours runs later — as written it would have silently dropped `mcode` from the whitelist. Renumbered to `344` and rebased its contents on 342, so the `up` keeps every family and the `down` restores 342's state rather than the state before it.
- The `mcode` case in `ListModels` came through with only a comment and no return; Go doesn't fall through, so the function reached its end unreturned. Caught by the build.

**Tool count** — `en.ts:104` and `zh.ts:104` bumped to 23. `ja` and `ko` phrase that line without a number. Separately, and not something I touched: #7035 added MiniMax Code to the backends but not to the enumerated tool lists on the landing page, so those still omit it.

multica-eve (changes requested):
Thanks for addressing the remaining RLM lifecycle issue with a fail-closed check. Refusing to start a task when Prime could silently enable fire-and-forget subagents is the right product behavior. The current implementation still has three merge blockers:

1. **The `agentintegration` build is broken.** `server/pkg/agent/prime.go:49` adds `primeAgentDir(env map[string]string)`, but `server/pkg/agent/prime_integration_test.go:229` already defines a helper with that name, and `server/pkg/agent/prime_lifecycle_integration_unix_test.go:214` still calls the old zero-argument signature. This fails before any account or quota is needed:

   ```text
   go test -tags=agentintegration ./pkg/agent -run '^$' -count=1
   ```

   The compiler reports `primeAgentDir redeclared`, argument-count mismatches, and return-value mismatches. Please unify or rename the resolver and ensure the tagged suite at least compiles.

2. **The fail-closed gate can inspect a different settings file from the Prime child process and miss a real override.** `server/pkg/agent/prime.go:49-53` trims the configured value and treats an explicitly empty value as absent, falling back to the daemon environment. The child process instead receives the merged environment at `prime.go:344`, where an explicitly empty configured value overrides the inherited value, and Prime does not trim this variable. Relative paths also diverge: the gate reads them relative to the daemon working directory at `prime.go:90`, while the Prime child resolves them after its working directory is changed to `opts.Cwd` at `prime.go:267-269`. For example, `PRIME_AGENT_CODING_AGENT_DIR=prime-config` with `<workdir>/prime-config/settings.json` containing `rlmMaxDepth: 1` is missed by the gate, so the original fire-and-forget completion problem remains possible. Please resolve the path from the exact final child environment and child working directory, preserve presence/value semantics, and add regressions for relative paths plus explicit empty and whitespace-only values.

3. **The numeric validation does not mirror Prime's `Number.isSafeInteger` check.** `server/pkg/agent/prime.go:102-104` verifies lossless conversion to `int64` and non-negativity, but it does not enforce JavaScript's maximum safe integer (`9007199254740991`). A value such as `9007199254740992` is rejected by Multica as a valid non-zero override, while Prime treats it as unsafe and falls back to `RLM_MAX_DEPTH=0`. That creates a compatibility-breaking false rejection. Please enforce the safe-integer upper bound and cover both boundary values.

Non-blocking follow-ups:

- The landing-page copy in all four languages now says 23 tools and includes Prime, but `scripts/agent-cli-command-names.txt` already contains 24 commands. The lists still omit MiniMax Code, which has landed on `main`.
- `server/migrations/344_runtime_profile_add_prime.down.sql:2` says it restores migration 313, but its contents restore the migration 342 state that includes `mcode`.

I rechecked the EOF lifecycle and did not find a new issue: the success path closes stdin, performs a bounded drain, cancels and terminates the process tree when needed, waits/releases the process tree, and only then closes the message stream. The daemon waits for that message stream to close before completing the task.

Validation: the default Prime tests, the RLM-gate race test, `ByteIdentical`, the complete migration lint, and `git diff --check` pass. All 14 GitHub checks are green and the PR is mergeable. The only failing validation is the tagged `agentintegration` compile check above.

**Recommendation: Request changes** until the three blockers are fixed.

marcelocecin:
@multica-eve @Bohan-J @forrestchang — all three blockers fixed, plus both follow-ups. CI green on `ca3ba32`.

**1. `agentintegration` build** (`4552432`) — my miss; I hadn't run the tagged build at all. Unified rather than renamed: `prime_integration_test.go`'s own `primeAgentDir` and the body of `primeGlobalRlmMaxDepthConfigured` now delegate to the production functions, so there's one definition of what Prime would read instead of two that can drift.

**2. Gate vs. child environment** (`4552432`) — right on all three counts. The gate now resolves from the merged environment handed to the process and from `opts.Cwd`: an explicitly empty `PRIME_AGENT_CODING_AGENT_DIR` is no longer conflated with an absent one, the value is no longer trimmed (Prime's `if (envDir)` treats `""` and `" "` differently), and relative values resolve against the child's working directory. Your `prime-config` example is now `TestPrimeFailsClosedOnRelativeGlobalAgentDir`, which fails against the previous code. `TestPrimeAgentDirForMatchesTheChildProcess` covers absent, absolute, explicitly empty, whitespace-only, relative, duplicate entries (last wins, as exec does) and tilde.

**3. `Number.isSafeInteger`** (`4552432`) — bound added, with both boundary values covered. Refusing `9007199254740992` would have rejected a host Prime itself runs safely — the same false-rejection class as the quoted-`"2"` bug from the previous round.

**Follow-ups** (`1bce9e0`) — the tool lists now name 24 in all four languages and include MiniMax Code, matching `agent-cli-command-names.txt`; #7035 added the backend without touching the marketing copy, so both gaps close together. The down migration's comment was the only stale part — its SQL already restored 342's set including `mcode`.

**Also** (`ca3ba32`) — renumbered to `349` after `344_plugin_v2_reset` landed. That collision only exists in the merge ref, which is why the local suite passed and CI didn't; I now merge `upstream/main` before picking a prefix so the local tree matches what CI builds.

marcelocecin:
@multica-eve @Bohan-J @forrestchang — one more divergence of the same class as your last blocker, found and closed. CI green on `e4e1b97`.

**The gate resolved the child's home from the daemon's environment** (`6ddc552`). Your blocker covered the explicit `PRIME_AGENT_CODING_AGENT_DIR`; this is the base of the path when that variable is absent. `os.UserHomeDir()` reads the daemon's env, but the child gets `mergeEnv(os.Environ(), cfg.Env)` and `custom_env` has no key blocklist — `agent_env.go` accepts anything. An agent setting `HOME` or `USERPROFILE` moved the child's `~/.prime/agent` somewhere the gate never looked, so a real non-zero global override went unseen. `TestPrimeFailsClosedWhenCustomEnvMovesHome` drives it through `Execute` and fails against the previous code.

Resolution now follows `reasonixEnv.userHomeDir` in `internal/daemon/execenv` rather than adding a parallel abstraction, including its GOOS test seam and its case-insensitive lookup — Windows environment names are case-insensitive, so a `custom_env` entry spelled `userprofile` still overrides the inherited `USERPROFILE`, which a plain string compare would have missed.

`HOMEDRIVE`/`HOMEPATH` are covered by tests that assert they must **not** matter: `os.UserHomeDir` keys off `USERPROFILE` alone on Windows, and libuv's `uv_os_homedir` does the same before falling back to `GetUserProfileDirectoryW`. Neither side moves, so they cannot desynchronise the two — the same pair `openclawDiscoveryEnvVars` already models.

**One CI change, flagged deliberately** (`e4e1b97`) — this is the first time this PR touches `.github/workflows/`, so I would rather point at it than have you find it. `backend-tests` gains one step that compiles the `agentintegration`-tagged tests: `go test -tags=agentintegration ./pkg/agent -run '^$' -count=1`. `server/pkg/agent` is the only package in the repo with those files and nothing else compiles them, which is how the redeclaration reached you last round. `-run '^$'` compiles without executing, so it resolves no agent CLI, needs no account and consumes no quota; it reuses the existing job rather than adding one. Happy to drop it if you would rather that land separately.

Verified: Prime tests, the package under `-race`, the tagged compile, and Windows cross-compilation of both the normal and tagged test binaries.

multica-eve (changes requested):
Thanks for the follow-up. The product direction remains acceptable: a Prime Agent task must not be reported complete while unfinished child work is still running. The previous tagged-build, child env/cwd, and safe-integer blockers are addressed, but the current head still has two merge blockers.

1. **The fail-closed gate still resolves the wrong agent directory when `HOME`/`USERPROFILE` is empty or absent.** `server/pkg/agent/prime.go:95-106` collapses both cases into a fallback through the daemon's own `os.UserHomeDir()`, which does not mirror the Prime v0.7.1 child runtime.

   On POSIX this is directly reproducible:

   - With `HOME=`, Node's `os.homedir()` returns an empty string. Prime's `getAgentDir()` at upstream `config.ts:521-526` therefore returns the relative path `.prime/agent`, so it reads `<opts.Cwd>/.prime/agent/settings.json`. The current gate reads the daemon home instead.
   - With `HOME` absent, Node falls back to the OS account home, while Go's `os.UserHomeDir()` on Unix returns `$HOME is not defined`. The current resolver then returns `""` and skips the check entirely.

   Either case can miss a real non-zero global `rlmMaxDepth` and reopen the original fire-and-forget completion bug. `server/pkg/agent/prime_test.go:946-981` currently pins the wrong result for an explicitly empty HOME, and its initial `os.UserHomeDir()` guard causes the entire test to skip when HOME is absent. Please preserve presence versus empty-value semantics, mirror the child runtime's account-home fallback, and fail closed if the child's actual directory cannot be proven. Add Execute-level regressions for both empty and absent values.

2. **Migration prefix 349 now collides with current `main`.** This branch adds `349_runtime_profile_add_prime`, but current `main` now contains `349_agent_task_queue_chat_terminal_resume_index` (plus migrations 350 and 351). The synthetic merge tree contains both 349 prefixes and therefore fails `TestMigrationNumericPrefixesStayUniqueAfterLegacySet`. This is a semantic prefix collision rather than a textual conflict, so GitHub still reports the PR as MERGEABLE. The 15 green checks completed before `main` acquired its 349 migration and do not validate the current merge. Please sync with `main`, choose the next free prefix at that point, and rerun the complete migration lint.

Non-blocking: `apps/web/features/landing/i18n/en.ts:195` and `zh.ts:195` enumerate all 24 tools, including MiniMax Code and Prime Agent, but still say 23. The Japanese/Korean copies and the other counts in those files say 24.

Validation on head `e4e1b97`: the default Prime tests, the `agentintegration` compile-only check, the resolver race tests, `ByteIdentical`, the branch-local migration lint, and `git diff --check` pass. The current-main merge audit exposes the 349 collision above. I did not run authenticated real-agent smoke tests.

**Recommendation: Request changes** until both blockers are fixed.

marcelocecin:
@multica-eve @Bohan-J @forrestchang — the two requested blockers were addressed, plus two additional blockers exposed by merging the current main. CI is green on `2170a3927ffe07111233552b6bd7e167a1175474`.

**1. Empty versus absent home variables.** The resolver now carries environment presence separately from the value and never falls back to the daemon's `os.UserHomeDir()`. On POSIX, an explicitly empty `HOME` preserves Prime/Node's empty-home behaviour and resolves `.prime/agent` relative to the child's `opts.Cwd`; an absent `HOME` uses the account-home fallback via `os/user`. If the account home cannot be determined, `Execute` fails closed instead of skipping the global-settings check.

On Windows, the adapter deliberately fails closed when a present `USERPROFILE` is shorter than three bytes. The bundled libuv behaviour differs by runtime version: Node 22.8.0, still within Prime Agent v0.7.1's declared support range, does not contain the later short-value guard, while Node 22.14.0 does. Because ACP cannot safely prove which bundled libuv behavior the child uses, this is an intentional conservative policy rather than a claim that every supported Prime runtime resolves the value identically. Native Windows execution was not available; the behavior was checked against the relevant Node/libuv sources and the Windows test binary was cross-compiled.

The regression suite covers empty, absent, invalid and account-fallback cases at resolver and `Execute` level, including the fail-closed boundary and the non-fail-always one.

**2. Migration prefix.** `349_runtime_profile_add_prime` was renamed to the next free prefix after merging the current `main` (`362` in the final tree). The runtime-profile CHECK set remains complete, with migration 342 still providing the latest upstream `mcode` family.

**3. Logging blocker exposed by the merge.** Current `main` moved runtime argv logging behind `Config.logAgentCommand` and added `TestOnlyLaunchGoLogsAgentCommandArgs`. Prime now uses that boundary; the literal `acp` value of `--mode` at argv index 1 is explicitly trusted, while all remaining adapter and custom arguments are redacted. A regression also proves that the same `acp` token supplied through custom arguments remains redacted.

**4. CI gating blocker exposed by the merge.** The `agentintegration` compile-only step added by this PR was missing the same `needs.changes.outputs.backend == 'true'` guard used by the other backend steps. It could run on PRs that do not touch `server/**` after the checkout/setup steps had been skipped. The guard was added in `2170a3927`, and the workflow now has consistent backend gating.

The landing-page count was corrected from 23 to 24 in English and Chinese. I did not run authenticated smoke tests against the real Prime binary.

Bohan-J (changes requested):
Request changes on `2170a392`. The ACP transport work is solid — this reuses the shared `hermesClient` correctly and the process-teardown / fail-closed `rlmMaxDepth` work is careful. Two blockers below, one of which is new and is a consequence of the "never report a session id" decision rather than of Prime's own limitations.

Migration prefix numbering is being handled separately by the maintainers, so it is excluded here.

## Blocker 1 — withholding `Result.SessionID` also disables workdir reuse

`prime.go:921-936` deliberately leaves `Result.SessionID` empty, pinned by `TestPrimeSessionIDNeverReported`. The stated goal — don't let the daemon claim a resume that cannot happen — is reasonable, but the session id is not only a resume pointer on the server side. It is also the row that carries `work_dir` forward.

The chain, all on the ordinary issue follow-up path (non-rerun, non-chat):

1. `TaskService.CompleteTask` (`internal/service/task.go:3634`) stores `pgtype.Text{String: sessionID, Valid: sessionID != ""}`, so every Prime task row lands with `session_id = NULL`.
2. `GetLastTaskSession`'s `latest_per_session` CTE (`pkg/db/queries/agent.sql:1009-1017`) filters `AND t.session_id IS NOT NULL`, so for a Prime agent+issue it matches zero rows and the `:one` query returns `ErrNoRows`.
3. `internal/handler/daemon.go:2435-2445` nests **both** assignments inside `err == nil && prior.SessionID.Valid`, so `resp.PriorWorkDir` is never set — not just `PriorSessionID`.
4. `shouldReusePriorWorkdir` (`internal/daemon/daemon.go:5755-5757`) returns false on an empty `task.PriorWorkDir`.

Net effect: **every follow-up run on the same issue gets a brand-new workdir.** The previous turn's branch, its uncommitted edits and any commits not yet pushed are stranded in an orphaned directory. That is a materially worse outcome than losing the model transcript, and it is not something Prime's protocol forces — it falls out of the choice made here.

Worth noting the asymmetry, because it shows the server already treats workdir and session as separable elsewhere: the manual-rerun path (`daemon.go:2406`) and the chat path (`daemon.go:2595`) both set `PriorWorkDir` *outside* the session check. Only the issue path couples them, and this is the first backend to hit that coupling — all 15 other backends report a session id on success.

MiniMax Code faces the identical upstream constraint (`loadSession: false`) and takes the other route: it reports the id, then answers a resume attempt with a typed `ResumeRejected`, which `shouldRetryWithFreshSession` turns into an in-run fresh-session retry. That costs one extra process spawn per follow-up but keeps the workdir.

Either direction is fine, but please pick one deliberately and cover it with a test:

- follow the mcode precedent (report the id, reject the resume, and suppress the false "resuming the prior session" claim in the brief), or
- lift `PriorWorkDir` out of the `SessionID.Valid` gate on the issue path so it matches the chat and rerun paths.

This was traced statically through the four hops above rather than run end-to-end against a database, so please confirm against a live follow-up run before settling on the fix.

## Blocker 2 — interim narration is delivered as the final result (#6006 regression)

`prime.go:593,605,903,921` accumulates every `MessageText` into one builder and hands the whole thing to `Result.Output`. `Result.Output` is the channel reply and the auto-generated issue comment, which is exactly what #6044 (MUL-5405) extracted `acpDeliverableTracker` to prevent.

8 of the 9 ACP backends use the tracker (hermes / kimi / reasonix / traecli / grok / mcode / kiro / qoder); `qwenpaw` is the one that predates it, and this backend was modeled on `qwenpaw.go`, so it inherited the gap.

Running `primeBackend` through `fakeACPDeliverableScript` — the same fake ACP runtime the shared regression tests already drive every other backend with:

```
STATUS   = "completed"
OUTPUT   = "I will read the diagram first.The diagram shows the service architecture."
EXPECTED = "The diagram shows the service architecture."
```

Fix is small: swap the builder for `acpDeliverableTracker` (`observe` / `result`), keep feeding the full stream to `promoteACPResultOnProviderError`, and add `{backend: "prime", binary: "prime-agent", notification: "session/update"}` to `acpDeliverableCases` so both existing cross-backend regression tests cover Prime.

## Non-blocking

1. `prime.go:823` logs `stopReason == "cancelled"` without setting a status. hermes / kimi / kiro / qoder / traecli / grok / mcode all set `finalStatus = "aborted"` (reasonix additionally splits error / unsupported); `qwenpaw` is again the only precedent for logging alone. Multica never sends `session/cancel`, so exposure is small, but if Prime ends a turn itself the task reports `completed` with truncated output.
2. `MinVersions` (`pkg/agent/version.go:13`) has no `prime` entry. The fail-closed gate's correctness is argued entirely from prime-agent v0.7.1 internals (`getAgentDir`, `_resolveRlmMaxDepth`, the Node floor), but nothing enforces that floor — a build whose settings path differs would make the gate read the wrong file and pass silently. `"prime": "0.7.1"` makes the assumption executable.
3. `internal/daemon/config.go:98` (comment) and `:244` (the "no agent CLI found" error) still omit Prime Agent, though `defaultAgentCommandNames` was updated.
4. `provider-logo.tsx` has no `prime` case, so it falls back to the generic icon.

## Verified

`go build ./...` passes; `go test ./pkg/agent -count=1` is fully green (~177s, all Prime cases included); `git diff --check` clean. The 4 failures in `./internal/daemon/execenv` (`TestHermesMemoryStorePathLayout`, `TestPruneHermesMemoryStores`, `TestHermesSessionStorePathLayout`, `TestPruneHermesSessionStores`) reproduce on a clean `origin/main` worktree and are unrelated to this PR. Smoke tests needing a real account and quota were not run.

Docs, landing counts (24, all four languages) and the migration down-comment from earlier rounds all look correct now.

marcelocecin:
@Bohan-J @multica-eve @forrestchang — both blockers addressed, plus the upstream drift found while merging current `main`. CI is green on `235653e17bf61cd16fc034314468f981a8db5c3e`.

**Blocker 1 — sessionless workdir.** Took the second option. One correction to the diagnosis, since it changes the shape of the fix: lifting `PriorWorkDir` out of the `SessionID.Valid` gate is necessary but not sufficient. `GetLastTaskSession` is `:one` and its `latest_per_session` CTE filters `session_id IS NOT NULL`, so for a sessionless backend it returns `ErrNoRows` — the row never reaches that scope at all. Recovering the workdir needs its own lookup, so `GetLastTaskWorkDirForIssueAndAgent` is scoped to the same `(agent, issue)` pair and ordered by the same temporal key, and the call is guarded on `PriorWorkDir == ""` so a backend that does report a session keeps its own session's directory. Deliberately not runtime-filtered, matching the existing rule at `daemon.go:2404` that the workdir is offered regardless of runtime while only the per-cwd session is runtime-gated. `shouldReusePriorWorkdir` and the on-disk provenance check are untouched.

One behaviour change worth calling out beyond Prime: an issue on any provider whose prior tasks failed before recording a session now gets a workdir offered where it previously got none.

**Blocker 2 — deliverable.** Reproduced the old implementation failing with exactly the narration-plus-answer output you predicted, then switched to the shared tracker. The dialect in `acpDeliverableCases` is `session/update` snake_case, verified against prime-agent v0.7.1's own `packages/coding-agent/test/acp-events.test.ts` rather than assumed. `promoteACPResultOnProviderError` keeps receiving the full stream. Minor: the table had 9 entries, not 8 — `qoderclicn` was already there.

**Cancellation.** Setting `aborted` on a Prime self-cancelled turn required changing the `session/close` guard: it keyed off `finalStatus != "aborted"`, which would have silently stopped closing that session the moment the self-cancel path started reporting `aborted`. It now tracks transport state instead, so an externally cancelled run still skips the close and a self-cancelled turn — which returns over a live connection — still closes its session.

**Migration.** Renumbered `362` → `368` after merging `main`. Byte-identical rename; re-verified that none of the merged commits touch `runtime_profile`, so `342` is still the base the CHECK rebuilds on and no family is dropped.

**Verification.** The five new handler tests ran and passed in CI with the Postgres service; they could not run in my local environment because Postgres was unavailable. The four-hop follow-up chain remains verified statically rather than through a live follow-up run. No authenticated Prime smoke test was run in this round.

marcelocecin:
@multica-eve @Bohan-J @forrestchang — requesting a fresh review of the current head.

New head: **`f4fbe7d10980ecc45c6ee3f7dbc1d1daab79d11c`**. The branch was synced with `upstream/main` at `fa805238f` before this push, and GitHub now reports the current merge ref as `MERGEABLE` / `CLEAN`.

## Synced with upstream main

Merged `main` into the branch (no rebase). Eleven files conflicted, all the same shape: `main` added Dim (#6675) as a 16th ACP provider while this branch added Prime, so both sides appended to the same provider lists. Every one is resolved by union — **Dim, MiniMax Code and Prime all survive, none replaced**:

`server/pkg/agent/agent.go` · `agent_supported_types_test.go` · `internal/daemon/config.go` · `internal/daemon/execenv/runtime_config.go` · `packages/views/runtimes/components/provider-logo.tsx` · `scripts/agent-cli-command-names.txt` · `apps/docs/content/docs/environment-variables.mdx` · `apps/web/features/landing/i18n/{en,ja,ko,zh}.ts`

Two needed more than a plain union:

- **`agent-cli-command-names.txt`** — `main` also sorted the file alphabetically, so `prime-agent` moves into sorted position rather than staying appended at the end.
- **Landing tool count → 25.** `main` was at 23 (Dim listed, MiniMax Code missing) and this branch was at 24 (MiniMax Code and Prime, no Dim). The union is 25, applied to all 11 lists across the four languages. Adding MiniMax Code there closes a pre-existing `main` gap — `mcode` is already in `SupportedTypes` and in `agent-cli-command-names.txt`. Nothing tests these counts, so they were checked by hand.

## Migration moved to 372 — CHECK rebuilt, not just renamed

`368_runtime_profile_add_prime` collided with `main`'s `368_skill_plugin_ownership`, and 369/370/371 are taken, so it moves to **`372`**.

The renumber alone would have been a silent bug, and I want to flag the mechanism because it would have passed both review and CI. Every `runtime_profile` family migration DROPs the whole `runtime_profile_protocol_family_check` and rebuilds it from a literal list, so that list is the constraint's entire new state — a family missing from it is revoked, not left alone. This migration's body was written against the era-342 family set, which predates `370_runtime_profile_add_dim`. Renumbered to 372 and left as-is, it would have run *after* 370 and dropped `dim` from the whitelist, and every test asserting "prime is allowed" would still have passed.

So the bodies are rebuilt on 370:

- **`372.up` = `370.up`'s set plus exactly `prime`** (24 families)
- **`372.down` = `370.up`'s set exactly**, `prime` removed (23 families)
- `dim` and `mcode` are spelled out explicitly in both rather than inherited

`server/internal/migrations/runtime_profile_family_chain_test.go` pins the invariant rather than this one instance: every `*_runtime_profile_add_*` migration's `up` must be its predecessor's set plus exactly the family its filename names, and its `down` must restore the predecessor's set exactly (compared as sets, since 242/254/313 insert mid-list). The whole existing chain — 134 through 372, twelve migrations — satisfies it. Deleting `dim` from `372.up` makes the test fail, so it is not vacuous.

While updating `agent.go`'s lockstep comment I fixed two stale numbers in it: it pointed at 362 for prime (leftover from the earlier 362→368 renumber; 362 is now `agent_task_durable_work_dir`) and at 327 for mcode — 327 is `workspace_share_link`, mcode was added by 342. `main`'s own `370_runtime_profile_add_dim.down.sql` already says 342.

No upstream migration is modified, including the documented legacy duplicate at prefix 362.

## CI on the new head

**Fifteen checks passed on `f4fbe7d10`; `image-budget` was skipped as expected because this PR touches no images.**

`backend` · `backend-tests` · `sqlc-check` · `go-vulnerability-scan` · `frontend` · `frontend-build` · `frontend-test` · `frontend-views-test` (both shards) · `windows-execenv` · `mobile` · `installer` (macOS/Ubuntu/Windows) · `changes`

Two things worth calling out from the `backend-tests` log, since both are new this round:

- **`go run ./cmd/migrate up` applied `372_runtime_profile_add_prime` against the real Postgres 17 service** and completed clean, so the rebuilt CHECK is exercised against a live database, not just parsed.
- **`internal/handler` ran with Postgres for 61.8s**, with no `database not reachable` skip anywhere in the log. That is the first real execution of the sessionless-workdir test from the last round — I have no local Postgres, so on my side it had only ever compiled. `internal/service` ran 11.1s, `internal/migrations` 1.3s (includes the chain test), `pkg/agent` 108.0s under `-race`. No failures.

CI builds the merge ref, so this ran against a `main` newer than the one merged into the branch, and it still merges and orders cleanly.

## Not verified

**No authenticated Prime smoke test was run this round.** The `agentintegration` build is compile-only (`-run '^$'`), which resolves no CLI and consumes no quota, so anything requiring a real account remains unrun. Relatedly, the `MinVersions` floor `"prime": "0.7.1"` still encodes an assumption argued from prime-agent v0.7.1 internals rather than one verified across builds.

**The sessionless-workdir test ran, but the end-to-end scenario has not been observed.** You asked last round for confirmation against a live follow-up run. What CI proves is that the test executes and passes against a real database — not that a real two-turn follow-up on the same issue was watched reusing its workdir. That observation is still outstanding, on my side and, as far as I know, on yours.

## Non-blocking follow-ups

1. The landing tool counts have no test, so they drift silently every time a provider lands. Happy to add one here or leave it for a follow-up — tell me which you prefer.
2. `environment-variables.mdx` keeps this branch's "and, when model overrides are supported, `MULTICA_<PROVIDER>_MODEL`" phrasing over `main`'s unconditional one, because the sentence immediately after names three tools that have no model variable. If you'd rather match `main`'s wording, that sentence needs rewording too.

---

The changes requested in the last review are implemented and validated by the checks listed above; the fixes are present in the current head and are ready for a fresh review. Please re-review `f4fbe7d10980ecc45c6ee3f7dbc1d1daab79d11c` — I'm not treating anything here as approved or ready to merge.

marcelocecin:
@multica-eve @Bohan-J — the merge conflict with main is resolved; requesting a fresh review of the current head.

New head: 3bef1f0f1e1b09691df59ec13e5e0d5e71b4999c. GitHub reports MERGEABLE / CLEAN, and the CI is green (15 checks passed; image-budget skipped as expected).

The merge conflicted only in `server/internal/daemon/leader_workdir_reuse_test.go`. Main’s latest workdir-safety implementation and tests were preserved verbatim; the PR’s sessionless `PriorWorkDir` test was preserved and adapted to the new `(string, bool)` helper signature. No production file was manually changed while resolving the conflict.

The remaining limitations are unchanged: no authenticated Prime smoke test, no native Windows Prime execution, and no observed real two-turn/resume flow. The DB-backed handler tests passed in CI against Postgres.

Please re-review the current head when convenient.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

Rebased onto v0.4.39 (merge `404e51e5`, renumber `55be1e27`, comment fix `944d518d`). All blockers and non-blocking notes from the 2026-08-19 review were already addressed; this round is the merge plus two follow-ups. Four things the diff shows but does not explain:

- **`ci.yml`** — #7984 added `go vet -tags agentintegration ./...` upstream, which covers what this branch's own agentintegration step was for and covers it more widely. The branch's step is dropped in favour of upstream's rather than kept alongside it.
- **`agent_supported_types_test.go`** — #7932 deleted `TestSupportedTypesMatchesMigrationWhitelist` as implementation-only coverage. The merge accepts that deletion instead of re-adding it with `prime`. `SupportedTypes` stays pinned to the CHECK by `runtime_profile_family_chain_test.go` and by the migration's own comment.
- **Migration `446` → `451`** — `446` went to `446_issue_properties_bigm_index` (#7878, which also took 447); main now reaches 450. The rebuilt CHECK still lists exactly the 26 families in `agent.SupportedTypes`, and 441 (codearts) is still the newest family migration on main, so the chain predecessor is unchanged.
- **Non-blocking #2, on `MinVersions["prime"]`** — the concern was that the fail-closed `rlmMaxDepth` gate is argued entirely from v0.7.1 internals with nothing enforcing that floor. Upstream has since shipped through v0.9.1, so I diffed it: `getAgentDir`, `ENV_AGENT_DIR`, `CONFIG_DIR_NAME` and `_resolveRlmMaxDepth` are unchanged, and `loadSession: false` with one session per connection still holds. Nothing the floor admits resolves the settings path differently. One thing did change — from v0.8 the ACP mode reads `session/new`'s `mcpServers` and rejects a non-empty list without host MCP support. This backend always sent an empty array, so behaviour is unaffected, but the comment claiming Prime never reads the field was wrong and is corrected in `944d518d`.

`make sqlc` with the pinned v1.31.1 regenerates with no diff locally. Still unverified by execution, as in earlier rounds: `TestPrimeWindowsDescendantsDieWithTheOwnedProcessTree` is compile-verified only — no Windows host here.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

Rebased past v0.4.41 (merge `94087b6b2` + migration renumber `90ce5cab2`, 451→457 — prefix only, no family CHECK change). CI green on `90ce5cab2`, `MERGEABLE` / `CLEAN`.

Both blockers from the 2026-08-19 review were resolved the same day and are on the branch:

- **Blocker 1** (`10ea02eac`) — `PriorWorkDir` lifted off the `SessionID.Valid` gate on the issue path; the sessionless row is recovered via a dedicated `GetLastTaskWorkDirForIssueAndAgent` since it never reaches `GetLastTaskSession`'s `session_id IS NOT NULL` CTE. 5 handler tests + 1 leader test, run under Postgres in `backend-tests`. A manual live-DB follow-up run is still the one thing not exercised.
- **Blocker 2** (`b98f82918`) — builder → `acpDeliverableTracker`, `prime` added to `acpDeliverableCases` so both cross-backend regression tests cover it.

Non-blocking 1–4 resolved. Contract re-verified through prime-agent v0.9.3 (see the updated Summary). Nothing outstanding on our side — requesting a re-review.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

Merged past v0.4.43 (`847f35306`), CI green on `e1b538230`. Besides the merge and the migration renumber (`dacddc6c9`, 457→468, prefix only), two small changes — reasoning is in each commit message:

- **The sessionless-workdir fallback now requires `checkout-keeps-work-v1`** (`e1b538230`), on the same terms #8291 / #8316 set for fresh-session retries. Any daemon that runs Prime advertises it, so Prime is unaffected. This is a deliberate behaviour change for older daemons: on any backend whose session lookup comes back empty they now get a fresh directory rather than a workdir their checkout would reset.
- **prime-agent v0.9.4 changed `getAgentDir`** to expand `~\` on Windows (`65c02d8c5`). For that prefix the rlmMaxDepth gate now checks both readings. Covered on a simulated platform only; no Windows host.

Still not exercised: a live-DB follow-up run. The handler tests covering it run in `backend-tests`. Requesting a re-review.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `9ac4526c7`. That push is a conflict-only merge of `upstream/main` (#8327, #8237): the Hermes/Prime env resets in `local_skills_test.go` and the DSH paragraph in `environment-variables*.mdx` landed on the same lines, both kept. Migration 468 is still the highest prefix.

The round before it (`e1b538230`) is still unreviewed. Besides merging past v0.4.43 (`847f35306`) and the migration renumber (`dacddc6c9`, 457→468, prefix only), it had two small changes — reasoning is in each commit message:

- **The sessionless-workdir fallback now requires `checkout-keeps-work-v1`** (`e1b538230`), on the same terms #8291 / #8316 set for fresh-session retries. Any daemon that runs Prime advertises it, so Prime is unaffected. This is a deliberate behaviour change for older daemons: on any backend whose session lookup comes back empty they now get a fresh directory rather than a workdir their checkout would reset.
- **prime-agent v0.9.4 changed `getAgentDir`** to expand `~\` on Windows (`65c02d8c5`). For that prefix the rlmMaxDepth gate now checks both readings. Covered on a simulated platform only; no Windows host.

Still not exercised: a live-DB follow-up run. The handler tests covering it run in `backend-tests`. Requesting a re-review.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `72154cb59`. Two pushes since the last round, neither changes behaviour:

- `9ac4526c7` is a conflict-only merge of `upstream/main` (#8327, #8237): the Hermes/Prime env resets in `local_skills_test.go` and the DSH paragraph in `environment-variables*.mdx` landed on the same lines, both kept. Migration 468 is still the highest prefix.
- `72154cb59` fixes a glossary slip in the Prime line of `providers.{zh,ja,ko}.mdx`: the product `issue` is now 任务 / タスク / 태스크, per `conventions.mdx` §2.

The round before it (`e1b538230`) is still unreviewed. Besides merging past v0.4.43 (`847f35306`) and the migration renumber (`dacddc6c9`, 457→468, prefix only), it had two small changes — reasoning is in each commit message:

- **The sessionless-workdir fallback now requires `checkout-keeps-work-v1`** (`e1b538230`), on the same terms #8291 / #8316 set for fresh-session retries. Any daemon that runs Prime advertises it, so Prime is unaffected. This is a deliberate behaviour change for older daemons: on any backend whose session lookup comes back empty they now get a fresh directory rather than a workdir their checkout would reset.
- **prime-agent v0.9.4 changed `getAgentDir`** to expand `~\` on Windows (`65c02d8c5`). For that prefix the rlmMaxDepth gate now checks both readings. Covered on a simulated platform only; no Windows host.

Still not exercised: a live-DB follow-up run. The handler tests covering it run in `backend-tests`. Requesting a re-review.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `953e8fb9a`, `mergeStateStatus: CLEAN`.

Two commits since the last round:

- `330690831` merges `upstream/main` (54 commits). No conflicts, nothing of mine inside it.
- `953e8fb9a` renumbers the Prime runtime-profile migration 468 → 491. `468_drop_reference_only_column` (#8253) took the prefix while this branch sat in review, and `TestMigrationNumericPrefixesAreUnique` fails on the merge ref with both stems claiming 468. The family list inside the CHECK is byte-identical — only the prefix moved. 441 (codearts) is still the newest family migration and nothing between 442 and 490 touches `protocol_family`, so 441 remains this migration's predecessor in the chain and the move revokes no family.

The 2026-08-19 review predates the September pushes, so rather than assume the fixes were seen, here is where each one landed:

- **Blocker 1** (withholding the session id also disables workdir reuse) — `e1b538230`. The four-hop trace was right. I kept `Result.SessionID` empty and fixed the workdir separately: the issue-path sessionless fallback now runs, gated on `DaemonCapabilityCheckoutKeepsWorkV1`. Lifting `PriorWorkDir` out of the `SessionID.Valid` gate unconditionally — the second option offered — would hand the directory to a pre-#8309 daemon, whose `multica repo checkout` runs `git reset --hard` and `git clean -fd` on an existing checkout and would delete exactly the work being carried forward. That is worse than the orphaned directory it replaces. Gating on the same capability `applyFreshSessionRetryWorkdir` already uses keeps Prime's behaviour intact and sends only older daemons to a fresh directory. The fallback is not Prime-specific: any backend whose `GetLastTaskSession` lookup comes back empty reaches it.
- **Blocker 2** (interim narration delivered as the final result) — `prime.go` now uses `acpDeliverableTracker` (`observe` / `result`), the full stream still feeds `promoteACPResultOnProviderError`, and `{backend: "prime", binary: "prime-agent", notification: "session/update"}` is in `acpDeliverableCases`, so both cross-backend regression tests cover Prime.
- Non-blocking 1–4 are all in: `stopReason == "cancelled"` sets `finalStatus = "aborted"`, `MinVersions["prime"] = "0.7.1"`, `internal/daemon/config.go` names Prime in both the struct comment and the "no agent CLI found" error, and `provider-logo.tsx` has a `prime` case.

Two things I still cannot verify from here, unchanged from the last round:

- The handler tests need Postgres and only run in CI, so `TestClaimTask_Issue*` — including the older-daemon subtest added for Blocker 1 — is green on the merge ref but never locally.
- I have no live Prime account with quota, so the end-to-end follow-up run you asked for in Blocker 1 has not happened. The capability gate is argued from #8291 / #8309 / #8316 and from the handler tests, not from an observed run.

One note from the merge, since it touches an earlier round's numbers: the landing counts moved again upstream, so the diff now reads 26 → 27, not 24 → 25. `Prime Agent` sits after `Pi` in all four locale files. The French locale added in #7141 leaves the public landing `locales` array at four and adds no `fr.ts`, so nothing there needs a Prime line.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `dab022df3`, `mergeStateStatus: CLEAN`.

Two commits since the last round, both upkeep — no behaviour change:

- `df31dac74` merges `upstream/main` (22 commits). The only conflict was the generated `server/pkg/db/generated/agent.sql.go`: #8377 dropped `GetLastTaskStartedAtForIssueAndAgent` right next to this branch's `GetLastTaskWorkDirForIssueAndAgent`. Resolved by regenerating with `make sqlc`, not by hand; `sqlc-check` passes.
- `dab022df3` renumbers the Prime runtime-profile migration 491 → 500, since #8466 took 491. The CHECK list is unchanged; 441 (codearts) is still the newest family migration.

All items from the 2026-08-19 review remain addressed as described in the earlier round.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `3855070ab`, `mergeStateStatus: CLEAN`. All six items from the 2026-08-19 review remain addressed; the two commits since the last round are a merge and a docs line.

- `86a2852e1` merges `upstream/main` (5 commits). One conflict, in `server/pkg/agent/version.go`: #8510 added `"opencode"` to `MinVersions` at the same spot after `"zeroclaw"` where this branch adds `"prime"`. The two floors are independent, so both blocks are kept — opencode in the position #8510 put it, prime appended after. The new table header is left exactly as written; the merge commit message records why the prime floor arguably sits on the opencode side of the distinction it draws.
- `3855070ab` adds Prime Agent to the minimum-version callout in `install-agent-runtime.{mdx,zh,ja,ko}`, following what #8510 did for OpenCode. `dim` and `zeroclaw` are still absent from that sentence; left alone rather than swept into this PR.

The four `internal/daemon/execenv` failures recorded as pre-existing in the last review are gone. #8512 came in with this merge and clears `MULTICA_TASK_CONFIG_ROOT` in that package's `TestMain`, which was the cause; `go test ./internal/daemon/execenv` is green locally now.

The stale `multica-docs` deployment failure above is a merge-ref race, not a docs break: `refs/pull/6641/merge` did not exist while this PR was conflicting, and GitHub recomputed it (`1eb985a0c`) two seconds after Vercel asked for it. Nothing is attached to the PR as a status or check.

Local gate on `pkg/agent`: `gofmt` clean, `go vet` and `GOOS=windows go vet` clean, `-tags=agentintegration` compiles, `go test ./pkg/agent -race` green (40.5s). Migration prefix `500` is still unique in the merged tree (upstream tops out at 499). Smoke tests needing a live account and quota were not run.

marcelocecin:
@multica-eve @Bohan-J @forrestchang

CI green on `7d2e625bf`, `mergeStateStatus: CLEAN`. All six items from the 2026-08-19 review remain addressed; the two commits since the last round are upkeep, no behaviour change:

- `f330747d4` merges `upstream/main` (65 commits, conflict-free).
- `7d2e625bf` renumbers the Prime runtime-profile migration **500 → 535**, because #8567 took 500 (`500_task_message_call_id`) and `TestMigrationNumericPrefixesAreUnique` would have failed the merge ref. The family list is unchanged: no family migration landed after 441 (501 adds `runtime_type` and leaves the CHECK alone).

Still open, as before: the live follow-up run for Blocker 1 has not been done (no Prime account with quota).

PR #6644, opened 2026-08-09
MUL-5937 fix(daemon): persist direct chat Codex sessions

deust132:
Closes WS-88

Summary:
- use chat_session_id as the stable Codex session storage key for direct chats
- preserve issue-based session storage behavior
- isolate different direct chats and protect active chat stores from cleanup

Verification:
- regression test passes after failing without the fix
- go test -race ./internal/daemon/...
- go vet ./internal/daemon/...
- go build ./cmd/multica
- live daemon v0.4.21-ws91 resumed the same Codex thread across task IDs f89afe4d and c9a06261

Operational note:
- deployed locally with a rollback binary retained
- independent review recorded PASS in Multica WS-90

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -495,7 +495,7 @@ type Daemon struct {
 
 	activeCodexStoresMu   sync.Mutex
 	activeCodexStoresCond *sync.Cond      // signalled when an in-flight store deletion finishes, so a blocked markActive can proceed
-	activeCodexStores     map[string]int  // per-issue Codex session store path -> live-task refcount; guards the store from GC mid-task (MUL-4424)
+	activeCodexStores     map[string]int  // per-conversation Codex session store path -> live-task refcount; guards the store from GC mid-task (MUL-4424)
 	deletingCodexStores   map[string]bool // store paths a GC delete has reserved; markActive waits these out so a task never mounts a store mid-removal
 
 	// localPathLocks serialises agent tasks whose project resource is a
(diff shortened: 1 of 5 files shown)


Bohan-J:
Thanks for the fix, @deust132 — nice catch on a subtle one.

The root cause is well identified: a direct chat bound to a `local_directory` project gets a fresh task-scoped `CODEX_HOME` every turn, while the persistent session store was keyed only on `(profile, agent, issue)`. With no issue ID there was no stable key, so the rollout stayed behind in the previous turn's temporary env root, the rollout presence gate dropped `PriorSessionID`, and the chat silently started a new thread. Keying on a `chat_`-prefixed `ChatSessionID` closes exactly that gap, and threading the same key through `Prepare`, `Reuse` and the active-store GC guard keeps the store from being reclaimed mid-mount.

Things I checked on `5be164c6`:

- Both new tests genuinely guard the regression — reverting the key change locally makes `TestPrepareCodexLocalDirectoryChatKeepsRolloutAcrossTaskIDs` and `TestCodexSessionStoreKeyUsesChatSessionIDWhenIssueAbsent` fail, and they pass with it.
- `go test ./internal/daemon/... -count=1` and `go vet ./internal/daemon/...` pass, including the prompt-cache `ByteIdentical` guards (untouched by this PR).
- The PR merged with current `main` still passes `./internal/daemon/execenv/`.
- Blast radius is contained: the persistent store is only mounted on the `IsLocalDirectory` path and the legacy-symlink-with-resume migration, so managed chats keep their task-local `sessions/`, issue tasks keep the issue key, and non-Codex providers are untouched. The `chat_` prefix cannot collide with a UUID issue ID, and the new stores are pruned by the existing retention GC at the same tree depth.

Merging. Thanks again for the thorough verification notes in the description — they made this easy to confirm.

PR #6649, opened 2026-08-09
fix(daemon): auto-retry transient rate/session-limit run failures

flow-84:
## Root Cause

A run that fails on a provider session limit (`You've hit your session limit · resets 2pm`) was classified as `agent_error.unknown` and **never re-queued** — the issue stayed on `todo` until someone ran `rerun` manually. An auto-retry path already exists (MUL-4910, `provider_network` only); this extends it to the transient limit class.

1. **Classification** (`server/pkg/taskfailure/classify.go`): session/usage-window limits now land in `provider_capacity_or_rate_limit` (transient, recovers at reset) instead of falling through. `session limit` used to slip past rule 4's `you've hit your limit` because `session` splits the substring. Quota errors (402 / insufficient balance / credits) stay terminal and non-retryable on purpose.
2. **Retry with backoff + ceiling** (`server/internal/service/task.go`): `provider_capacity_or_rate_limit` added to the retry allowlist, staggered backoff **15m → 1h → 4h**, hard ceiling of 4 attempts (1 run + 3 retries). After exhaustion the run stays `failed`, no silent infinite retry. Gated to **unattended issue runs** — interactive chat is not silently deferred for hours (CLI retry + pending-message handle that path).
3. **Visibility**: a delayed retry posts **one** system comment on the issue (`… auto-retry scheduled in ~15m (attempt 2 of 4)`) so it's traceable as automatic, not a manual intervention. The ~5s `provider_network` delay stays silent on purpose (no spam).

## Changed files
- `server/pkg/taskfailure/classify.go` (+ tests in `classify_test.go`)
- `server/internal/service/task.go` (+ new `task_retry_schedule_test.go`)

## Verification (local)
- `go build ./...` — green
- `go vet ./internal/service/ ./pkg/taskfailure/` — green
- `go test ./pkg/taskfailure/ ./internal/service/` — green (incl. new session-limit classification and backoff/ceiling/eligibility tests)

### DIFF
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -4929,6 +4929,23 @@ func (s *TaskService) FailTask(ctx context.Context, taskID pgtype.UUID, errMsg, 
 		}
 	}
 
+	// Visibility for a *deferred* auto-retry (MS-121): when the retry child was
+	// pushed out by a real backoff, the issue would otherwise sit silent for
+	// 15m–4h with no sign that a retry is pending. Post one system comment so the
+	// history shows this was automatic, not a manual rerun. Gated on a >=1min
+	// delay so provider_network's ~5s final-attempt deferral stays silent — only
+	// the capacity/rate/session-limit schedule is announced. The raw-error
+	// comment below is skipped whenever retried != nil, so there is no duplicate.
+	if retried != nil && retryFireAt.Valid && task.IssueID.Valid {
+		if wait := time.Until(retryFireAt.Time); wait >= time.Minute {
+			note := fmt.Sprintf(
+				"Transient provider limit hit (%s) — auto-retry scheduled in ~%s (attempt %d of %d). No manual rerun needed.",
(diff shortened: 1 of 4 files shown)


flow-84:
@Bohan-J this one is ready for review whenever you have a moment — `mergeable=true`, rebased onto current `main` (one real conflict in `task.go` resolved: merged cleanly with the parallel `skill_bundle_unavailable` / MUL-5370 addition to `retryableReasons`, no loss). Build + tests green.

It fixes runs that fail on a transient provider session/rate limit (`You've hit your session limit · resets 2pm`, HTTP 429) being classified as `agent_error.unknown` and never retried — the issue then sits on `todo` until a manual rerun. The PR files those under `provider_capacity_or_rate_limit` and gives that reason a staggered 15m/1h/4h auto-retry (hard ceiling 4 attempts), gated to unattended issue runs, with a system comment when a retry is deferred. Quota/billing stays terminal.

Happy to adjust scope or the backoff schedule if you'd prefer different tuning.

PR #6651, opened 2026-08-09
fix(core): validate Mika bootstrap response with a schema

flow-84:
## Summary
- `createMikaAgent` returned the raw fetch JSON cast directly to `MikaBootstrapResponse` instead of going through `parseWithFallback` — the only Mika endpoint skipping that path.
- Adds `MikaBootstrapResponseSchema` / `EMPTY_MIKA_BOOTSTRAP_RESPONSE` (loose, minimal fields, same pattern as `MinimalAgentSchema`) and wires `createMikaAgent` through it.
- An older desktop client talking to a changed backend would previously crash instead of degrading, per the API Compatibility rules in CLAUDE.md.

## Test plan
- [x] `pnpm --filter @multica/core typecheck`
- [x] `pnpm --filter @multica/core exec vitest run api/client.test.ts onboarding/use-bootstrap-mika.test.ts` (77 passed, incl. new malformed-response test)

### DIFF
--- a/packages/core/api/client.test.ts
+++ b/packages/core/api/client.test.ts
@@ -1971,3 +1971,54 @@ describe("ApiClient startMikaOnboarding", () => {
     ).resolves.toEqual({ started: false });
   });
 });
+
+describe("ApiClient createMikaAgent response schema", () => {
+  it("returns the agent and onboarding session a well-formed response reports", async () => {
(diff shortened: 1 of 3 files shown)


flow-84:
Closing — target should be my fork (flow-84/multica), not upstream.

PR #6654, opened 2026-08-09
MUL-5856: fix(agent): parse tab-separated Antigravity model catalog

### LINKED ISSUE #6562: [Bug] Parse tab-separated model slugs from Antigravity CLI 1.1.11
## Summary

The Antigravity CLI changed `agy models` output in 1.1.11 from one bare model slug per line to a tab-separated slug and display label:

```
gemini-3.6-flash-high\tGemini 3.6 Flash (High)
```

Multica's Antigravity discovery currently treats each complete line as both `Model.ID` and `Model.Label`. The driver therefore advertises the tab-containing string as the model ID, while an exact pinned model such as `gemini-3.6-flash-high` is rejected during validation.

## Evidence

- `agy 1.1.9` and `1.1.10`: bare slugs, for example `gemini-3.6-flash-high`.
- `agy 1.1.11`: `slug<TAB>display label`, 11 catalog entries.
- `--model gemini-3.6-flash-low` still succeeds in `agy 1.1.11`; the catalog parser/validation path is the incompatibility.
- The current parser is `server/pkg/agent/models.go:1374` (`parseAntigravityModels`), and the existing fixture is `server/pkg/agent/models_test.go:1209`.

## Requested fix

Parse the first tab-delimited field as `Model.ID` and preserve the complete display label (or the post-tab field) as `Model.Label`. Keep the legacy bare-slug shape working, and add fixtures for both shapes, including exact model validation against the slug.

This is a closed validation failure: affected exact-pinned Antigravity seats are refused before spawn when their host has `agy 1.1.11`.

Environment: Multica `v0.4.20`, Antigravity CLI `1.1.11`.

mxbao063-bao:
## Summary

- parse the agy 1.1.11 id-tab-label model catalog format
- retain support for legacy single-column model output
- deduplicate by model ID and ignore future trailing columns
- validate exact pinned model slugs against the parsed catalog

## Testing

- go test ./pkg/agent -run TestParseAntigravityModels -count=1
- go vet ./pkg/agent

Fixes #6562

### DIFF
--- a/server/pkg/agent/antigravity.go
+++ b/server/pkg/agent/antigravity.go
@@ -423,16 +423,17 @@ var antigravityBlockedArgs = map[string]blockedArgMode{
 // buildAntigravityArgs assembles the argv for a daemon-compatible one-shot agy
 // invocation.
 //
-//	agy -p <prompt> --dangerously-skip-permissions [--model <display name>]
+//	agy -p <prompt> --dangerously-skip-permissions [--model <catalog id>]
 //	    --print-timeout <duration> --log-file <tmp>
 //	    [--conversation <id>] [--add-dir <cwd>]
 //
 // agy 1.0.6 added a `--model` flag (MUL-3125), so opts.Model is now wired
-// through when set. The value is the exact human display string `agy models`
-// prints (e.g. "Claude Opus 4.6 (Thinking)"), NOT a provider/model slug —
-// it's passed verbatim as a single exec arg, so spaces and parens need no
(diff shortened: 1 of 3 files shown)


multica-eve (approved):
Thank you @mxbao063-bao for the clear, well-scoped fix and regression coverage. We re-reviewed the final diff, confirmed compatibility with both the new tab-separated catalog and legacy single-column output, and all required checks are green. The stale comments identified in review have also been corrected. Approved — thanks for contributing!

PR #6660, opened 2026-08-09
feat(dingtalk): route groups to different agents

mxbao063-bao:
## Summary

- let one DingTalk Stream bot serve multiple groups, with one fixed agent per group
- auto-discover groups from their first inbound mention and default them to the installation agent
- add owner/admin group routing controls in Settings > Integrations > DingTalk
- isolate chat sessions when a group is reassigned so agent context cannot cross routes
- add migrations, scoped APIs, realtime invalidation, translations, docs, and regression coverage

No custom webhook or in-group agent selector is introduced; the existing Stream callback conversation ID is used as the routing key.

## Verification

- go test ./...
- go vet ./...
- go build ./...
- pnpm typecheck
- non-desktop tests: core 1355, views 3745, web 195, docs 17 all passing
- relevant DingTalk handler, resolver, inbound, schema, and settings tests passing
- git diff --check

The desktop production bundle builds. The full web/docs production build is blocked on this machine because Next.js cannot fetch Google Fonts through the local TLS certificate chain (UNABLE_TO_GET_ISSUER_CERT_LOCALLY), unrelated to these changes.

## Manual validation before ready for review

- add the same DingTalk robot to two groups
- mention it once in each group so both routes are discovered
- assign a different agent to each group in Settings
- verify replies and follow-up context stay isolated per group/agent
- reassign one group and verify its next message starts a fresh agent session


yyclaw (changes requested):
@mxbao063-bao Thanks for submitting this PR and for adding DingTalk group-to-agent routing.

I re-ran the preflight against the latest `upstream/main` available at review time (`985a1fe6`). The PR branch diverged from `main` at `cd3c0bb8`, so this review also covers compatibility with changes that landed on `main` afterward.

The feature direction is worth continuing, but the current revision is **NO_GO / Request changes**. Please address all P1 and P2 items below; the P2 findings are actionable requirements, not nits or optional follow-ups.

### [P1] Rebase and resolve migration-prefix collisions

The PR adds migrations `265`, `266`, and `267`, while current `main` already contains migrations numbered `265` through `271`. On a synthetic merge of the PR with current `main`, `TestMigrationNumericPrefixesStayUniqueAfterLegacySet` fails because those numeric prefixes are duplicated.

Please rebase onto current `main`, renumber the DingTalk migrations to the next available prefixes (currently `272`–`274`), and rerun the migration up/down and prefix-validation tests.

### [P1] Preserve existing DingTalk group sessions during upgrade when the effective agent has not changed

`DeleteDingTalkStaleGroupChatBinding` in `server/pkg/db/queries/dingtalk.sql` compares the route agent with `binding.config->>'agent_id'`. Existing pre-PR group bindings do not contain `agent_id`, so the comparison treats every legacy binding as stale. The first post-upgrade group message therefore deletes the existing binding and creates a new chat session even when the group still routes to the same installation/default agent.

This is an impact on existing DingTalk behavior, not only a new-feature path. Please preserve the legacy binding when the effective agent is unchanged and add an upgrade regression test using a pre-PR binding without `config.agent_id`.

### [P1] Do not persist discovered groups before mention and identity/membership validation

`ResolveDingTalkInstallationForInboundGroup` currently upserts `dingtalk_group_route` while resolving the installation. The shared router performs that resolution before checking `AddressedToBot` and before identity/workspace-membership validation. As a result, an unmentioned, unbound, or unauthorized group callback can persist the group ID/title even though the message is subsequently dropped, and that data can then appear in the member-visible group list.

Please separate read-only installation lookup from group discovery and only persist a route after the callback is addressed to the bot and the sender has passed the required identity/membership checks. Add full-pipeline regression tests proving rejected callbacks do not write a route.

### [P1] Define and enforce route behavior when the selected agent is archived

The PATCH path verifies that the target agent is active only at assignment time. Archiving an agent does not clean up or reassign its DingTalk group routes, and inbound resolution does not revalidate the route target. A route can therefore continue pointing to an archived agent, after which task enqueueing fails with `ErrChatTaskAgentArchived`. The assignment/archive operations also lack a locking or transactional invariant that closes the race between them.

Please define the intended lifecycle across archive, restore, reassignment, and runtime teardown, then enforce it atomically or revalidate/fallback at runtime. Add regression coverage for the lifecycle and a controlled assignment-vs-archive interleaving.

### [P2] Make `dingtalk_group_route.id` a database-enforced stable identifier

The API treats `id` as the PATCH resource identifier, but the new table does not define a primary key or unique index for it. The existing unique index only covers `(installation_id, conversation_id)`.

Please add a `CREATE UNIQUE INDEX CONCURRENTLY` migration for `id` in its own single-statement migration file, following the repository migration rules, and add catalog-level coverage verifying the constraint.

### [P2] Add test coverage for the changed behavior and product surface

The current tests cover several query happy paths, title propagation, one rendered route row, and schema fallback, but material new behavior remains unprotected. At minimum, please add coverage for:

- GET/PATCH handler authorization, validation, not-found/error responses, and malformed client responses;
- legacy binding preservation across upgrade;
- rejected/unmentioned/non-member callbacks producing no route writes;
- archived-agent lifecycle and the assignment/archive race;
- migration up/down behavior and actual database catalog invariants;
- UI reassignment, non-manager read-only behavior, failure handling, and realtime invalidation.

Code changes for this feature need regression tests for their observable behavior; these gaps should be addressed in this PR rather than deferred.

### [Non-blocking product decision] Confirm session semantics for Agent A → Agent B → Agent A

Local validation shows that routing a group from Agent A to Agent B and then back to Agent A creates a new Agent A chat session instead of resuming the original Agent A session.

This follows the current implementation: each reassignment removes the active group binding while preserving the old chat session only as history. Returning to a previously used agent therefore creates another session.

We are not treating this behavior itself as a merge blocker because either semantic can be valid:

1. Always start a fresh session after every route change, maximizing context isolation.
2. Resume the previous per-(group, agent) session when routing returns to an earlier agent, preserving that agent's prior context.

Please explicitly confirm which behavior is intended, document the choice, and add an A → B → A regression test so future changes do not alter the selected product semantics accidentally.

The focused Go tests, the full `@multica/core` and `@multica/views` test suites, and both package typechecks pass on the PR head. The migration-prefix test fails on the synthetic merge with current `main`, as described above.

yyclaw (changes requested):
@mxbao063-bao Thanks for the update and for continuing to harden the DingTalk group-routing implementation.

I ran an adversarial security pass against the current PR head (`ee40b6fce`). The cross-workspace route and agent fences, member/admin authorization, rejected-callback gating, route cutover isolation, secret handling, and stored-XSS behavior all held under the tested cases.

There is one remaining PR-specific P2 that should be addressed before merge.

### [P2] Do not keep revoked DingTalk group routes discoverable and member-visible

`DiscoverDingTalkGroupRoute` validates the installation id, workspace, and channel type, but does not require the installation to be active. `ListDingTalkGroupRoutesByWorkspace` likewise joins the installation without filtering `status = 'active'`. Revocation only changes the installation status and deliberately retains the row.

The adversarial tests confirmed both observable consequences:

- If a callback resolves the installation while active and the installation is revoked before validated group discovery runs, the callback can still create a new `dingtalk_group_route` after revocation.
- After an installation is revoked, an ordinary authenticated workspace member can still call `GET /api/workspaces/{workspaceId}/dingtalk/group-routes` and retrieve the retained group title and conversation id. The current web UI hides this state when no installation is active, but the backend API remains accessible.

This makes disconnect ineffective as a visibility boundary for the newly introduced group metadata. It also means future workspace members may learn DingTalk group names and identifiers retained from an already disconnected bot.

Please define and enforce the intended post-disconnect data policy. A safe default would be:

1. Require `channel_installation.status = 'active'` during group discovery.
2. Exclude revoked-installation routes from the member-visible list, while preserving the rows internally if reconnect must restore assignments.
3. If retained routes must remain inspectable for audit, expose them through an explicit admin-only audit surface instead of the normal member route list.
4. Add regression tests covering revoke-before-discovery and member GET after revoke.

For scope clarity: an already-resolved callback being able to continue toward message/task processing is a pre-existing shared channel-engine revocation semantic, not something I am asking this PR to redesign. The blocking finding here is the DingTalk Group-specific persistence and member visibility introduced by this PR.

yyclaw (changes requested):
@mxbao063-bao Thanks for the update.

Requesting one change before merge: revoked DingTalk installations must not expose or create Group routes.

Please make these changes:

1. Update `DiscoverDingTalkGroupRoute` to require `channel_installation.status = 'active'`. After revocation, a callback must not create or discover a Group route.
2. Update `ListDingTalkGroupRoutesByWorkspace` to return routes only for active installations. After revocation, workspace members must not receive the retained group title or conversation ID from `GET /api/workspaces/{workspaceId}/dingtalk/group-routes`.
3. Add regression tests for both cases:
   - revoke before Group route discovery => no route is created or returned;
   - member GET after revoke => the revoked installation's routes and metadata are absent.

It is fine to retain the route rows internally if reconnect needs to restore assignments; they just must not be discoverable or member-visible while the installation is revoked.

This request does not require changing the shared Router's handling of callbacks that were already in flight when revocation occurred.

yyclaw (changes requested):
@mxbao063-bao, thank you for the updates and for addressing the previous review findings. I re-tested the latest head (`b03c3a2cb301`): the PostgreSQL test connection-pool exhaustion issue is fixed, the migration numbering conflict with the latest `main` is resolved, and all current CI checks are green.

There are still two P2 issues that need to be fixed before this PR can be merged:

1. **[P2][DingTalk Channel][PR-introduced] Reject group-route reassignment after the installation is revoked**

   The list API hides routes belonging to a revoked installation, but the PATCH/reassignment query does not verify that the associated DingTalk installation is still active. An owner who already knows the route ID can therefore PATCH the retained route after revocation and receive its conversation ID/title in the response.

   Please:
   - validate and lock the associated active DingTalk installation as part of the reassignment transaction, so PATCH is correctly serialized with revoke;
   - return 404 without route metadata when the installation has been revoked;
   - add tests for owner PATCH-after-revoke and the concurrent revoke/PATCH interleaving.

2. **[P2][DingTalk Channel][PR-introduced] Do not render a group-route query failure as an empty group list**

   `DingTalkTab` currently ignores the query error state and defaults missing data to an empty array. A 500/network failure is therefore presented as “No groups discovered,” which makes an operational failure indistinguishable from a successful empty result.

   Please:
   - render a localized error state for a failed group-route query, preferably with a retry action;
   - add a UI test proving that a query failure does not render the empty state.

Please ping me after these two findings are addressed and I will re-run the preflight review.

yyclaw:
@mxbao063-bao, one clarification to the second P2 in my requested-changes review, based on a local reproduction:

The current group-route query uses a fixed `refetchInterval: 5_000`, and the shared QueryClient also retries a failed query once. When `/dingtalk/group-routes` persistently returns 500, the Integrations page therefore keeps requesting the endpoint and continuously adds API/console errors while it remains open.

Please treat this as part of the same **[P2][DingTalk Channel][PR-introduced]** finding. The fix should not only render a localized error message. It should also:

- stop fixed-rate polling while the query is in an error state, or apply a bounded/backoff retry policy;
- provide an explicit retry action;
- add a test proving that a persistent query failure neither renders the empty-state message nor continues issuing requests every five seconds indefinitely.

The server-side cause of the initial 500 should be diagnosed separately. If it still occurs after the PR migrations have been applied, that backend failure also needs to be fixed.

yyclaw (changes requested):
@mxbao063-bao, thank you for the latest update. I rechecked the current PR head (`02b36920ae18`). The explicit group-route error state, bounded polling after a route-query failure, and retry action are now in place. The Agent A -> Agent B -> Agent A fresh-session behavior is also documented and covered.

Three findings remain. The two P2 findings need to be fixed before merge. The P3 is non-blocking, but please handle it in this PR rather than leaving it as a nit.

### [P2][DingTalk Channel][PR-introduced] Reject and serialize route reassignment after installation revocation

`ReassignDingTalkGroupRoute` still locks only the target agent and the route. It does not join or lock the route's `channel_installation`, and it does not require that installation to be an active DingTalk installation. An owner who knows a retained route ID can therefore PATCH it after revocation and receive `200` with the retained conversation ID/title, even though GET correctly hides that route.

Please:

- validate and lock the route's active DingTalk installation in the reassignment statement/transaction so PATCH is serialized with revoke;
- make revoke-win return `404` without route metadata and without changing the route or deleting its current chat binding;
- update the handler's no-row diagnosis so a retained route belonging to a revoked installation is not reported as an agent conflict;
- add owner PATCH-after-revoke and controlled revoke/PATCH interleaving tests that assert both the HTTP result and durable route/binding state.

### [P2][DingTalk Channel][PR-introduced] Do not interpret Agent query loading/failure as an empty Agent list

`DingTalkTab` currently destructures `agentListOptions` as `data: agents = []` and ignores its loading and error states. Group routing can therefore render while the Agent query is loading or has failed. Existing routes then show `Unknown Agent`, and an owner/admin receives an empty Agent selector, making an operational failure indistinguishable from successfully loading no eligible Agents.

Please:

- render distinct Agent loading, loaded-empty, and error states for Group routing;
- provide a localized retry action for Agent-query failure and keep reassignment disabled until the Agent list has loaded successfully;
- add UI tests for Agent loading, query failure, successful empty data, and successful recovery after retry. The loading/failure cases must not render the normal empty selector state.

### [P3][DingTalk Channel][PR-introduced] Do not poll group routes while the Group routing surface is hidden

The group-route query is enabled by `configured && !!wsId`, while the Group routing section is rendered only when at least one installation is active. If DingTalk is configured but every installation is revoked, the section is hidden but `/dingtalk/group-routes` still fetches and continues its five-second polling in the background.

Please:

- derive one `hasActiveInstallation` condition and use it for both query enablement and section visibility;
- ensure polling stops when the last active installation becomes revoked;
- add a UI test proving that no initial request or timer-driven refetch occurs with zero active installations, including the active-to-revoked transition.

Please ping me after these three items are addressed.

mxbao063-bao:
@yyclaw The three remaining findings are addressed in `2da071fce`.

- Reassignment now locks and revalidates the active DingTalk installation before the route, so revoke-first returns a metadata-free 404 and preserves the durable route and chat binding. Handler and controlled revoke/PATCH regressions cover both outcomes.
- Group routing now has distinct localized Agent loading, loaded-empty, and error states, an explicit retry action, and no reassignment surface until the Agent query succeeds. Loading, failure, empty, and retry recovery are covered.
- One `hasActiveInstallation` condition now controls both query enablement and section visibility. Tests cover zero-active initial render and active-to-revoked polling shutdown.

Local verification: focused DingTalk Go tests passed; full `@multica/views` suite passed (3789 tests); focused UI suite passed (22 tests); `pnpm typecheck` passed; ESLint passed. Please rerun the preflight review when ready.

yyclaw (changes requested):
@mxbao063-bao, thank you especially for your persistence in working through the repeated review rounds and for continuing to harden this feature. Barring any unexpected issues in the fixes below, I expect this to be the final review round.

There are five remaining blocking items:

1. **[P1][Migration][PR-introduced] Make migrations 274-276 recoverable after an interrupted concurrent index build.**

   `CREATE INDEX CONCURRENTLY` can leave an INVALID relation when it is cancelled or fails. A later retry of the current migrations then fails because the relation already exists. Please register `cleanupInvalidConcurrentIndexHook` for all three new indexes (`274_dingtalk_group_route_installation_conversation_unique`, `275_dingtalk_group_route_workspace_index`, and `276_dingtalk_group_route_id_unique`). Add a failure-injection migration test that leaves an invalid index, retries through `runMigrations`, and verifies the rebuilt index is both ready/valid and the migration is recorded.

2. **[P2][DingTalk Channel UI][PR-introduced] Keep discovered group routes visible independently of Agent query state.**

   Once the group-route query succeeds and returns routes, those rows must remain visible. Currently Agent loading, Agent query failure, or zero visible/active Agents replaces the entire route list; this also hides routes when every Agent is archived or when a regular member cannot see the assigned private Agent. Please always render the returned route rows. Agent state should only control the selector's availability/options and any inline loading/error/empty explanation. Add coverage proving existing routes remain visible for Agent loading, failure, no active Agents, and an archived or non-visible assigned Agent.

3. **[P2][DingTalk Channel UI][PR-introduced] Do not silently discard a second route update.**

   While route A is updating, only A's selector is disabled, but the global `updatingRouteId` guard silently returns when the user changes route B. The UI therefore presents an enabled action that performs no request and shows no feedback. Please either track pending state per route so independent updates can proceed, or disable every selector while any route update is pending. Add a two-route test that holds the first request open and verifies the second visible action is either disabled or executed.

4. **[P2][DingTalk API compatibility][PR-introduced] Degrade cleanly when the client is newer than the backend.**

   Web/Desktop frontend and backend deployment can be version-skewed, and installed Desktop clients may connect to an older/self-hosted backend. With an active DingTalk installation, the new client unconditionally calls `/dingtalk/group-routes`; an older backend returns 404 and the existing Integrations page reports a persistent load failure. Please make unsupported group routing an explicit non-error state, preferably via an optional `group_routing_supported` capability returned by the installations endpoint and an `=== true` client gate. Cover both new-client/old-server (no request or error UI) and new-client/new-server behavior. A bare 404-to-empty fallback should not leave five-second polling active or misreport an unsupported feature as an empty route list.

5. **[P2][Shared cleanup paths][PR-introduced][Test coverage gate] Add behavioral DB coverage for every new route-cleanup branch.**

   This PR adds `dingtalk_group_route` cleanup to installation replacement, dead-installation reclaim, system-runtime teardown, and workspace deletion, but the corresponding DB tests do not seed and assert these rows. Please extend the existing cleanup fixtures/tests to prove target routes are deleted and unrelated/live-owner routes are preserved for `DeleteDingTalkInstallationForReplacement`, `ReclaimDeadChannelInstallationByAppID`, `DeleteChannelInstallationsBySystemRuntimeAgents`, and the workspace deletion paths. The manifest entry alone does not exercise the production SQL.

The previously reported revoke boundary, explicit retry/error states, and no-active-installation polling issues look resolved. The accepted A -> B -> A fresh-session behavior remains non-blocking.

mxbao063-bao:
@yyclaw Addressed all five remaining blocking items in commit acbfb30a3:

1. Migrations 274–276 now register invalid-index cleanup hooks and use single-statement CREATE INDEX CONCURRENTLY IF NOT EXISTS, covering both an INVALID leftover and the valid-index/version-not-recorded interruption window.
2. Discovered group-route rows remain visible through Agent loading, failure, loaded-empty, archived-assignee, and non-visible-assignee states; Agent state only controls the selector and inline status/retry UI.
3. While one reassignment is pending, all route selectors are disabled, so a second visible update cannot be silently discarded.
4. The installations response advertises group_routing_supported, and the client gates the new endpoint on === true, so newer clients do not query or show errors against older backends.
5. Behavioral DB tests now cover route cleanup for replacement, dead-installation reclaim, system-runtime teardown, and both workspace deletion paths, while asserting unrelated/live routes are preserved.

Verification: focused migration/index recovery tests passed; all DingTalk integration and handler tests passed against a fully migrated temporary DB; @multica/core tests passed (1351); @multica/views tests passed (3794); typecheck, ESLint, make build, and git diff --check passed. GitHub CI is green (10/10 checks). The temporary verification database was removed.

Please rerun the final preflight review when ready.

yyclaw (changes requested):
@mxbao063-bao, thank you for continuing to push this forward and for patiently working through the previous review rounds. All five blocking findings from the last round are now resolved. This PR has only the following two test-coverage items left before merge. If the new tests do not expose an implementation defect, no additional production-code change is required.

### [F-6660-SHARED-1][P2][blocking][shared_engine] Complete terminal-path coverage for the shared route-conflict recovery state machine

Commit `199a4cf3` introduced `refreshChangedRoute` and the recovery branches for `ErrRouteChanged` from `EnsureSession` and `AppendMessage`. The current `TestRouter_RouteFenceConflictRefreshesAndRetriesSameClaim` covers only a successful refresh after an append conflict; the other terminal paths in the same new state machine remain untested.

Please add targeted tests covering at least:

- an `EnsureSession` conflict followed by a successful refresh and retry on the new route using the same claim;
- an `EnsureSession` conflict followed by an archived target or an infrastructure error during refresh;
- an append conflict followed by an archived target or an infrastructure error during refresh;
- a missing `Validated` resolver, a session append-lock failure, and repeated conflicts terminating through context cancellation/deadline;
- for every terminal path, the expected dedup claim Mark/Release behavior and proof that nothing is appended to the old route.

### [F-6660-DINGTALK-2][P2][blocking][single_channel] Complete the DingTalk route-cutover adapter and handler failure/retry test matrix

The current tests cover the happy path, the initial route mismatch, and `pgx.ErrNoRows` from the append fence, but the following PR-introduced branches still lack targeted coverage:

- DingTalk P2P is a no-op for group discovery: it must not call group discovery and must continue using the installation's default Agent and existing direct-message session flow;
- callback decode errors and route-discovery query errors;
- initial/recheck route-query errors, stale-binding cleanup errors, and underlying `EnsureSession` errors;
- a route conflict/error after session creation and exhaustion of all three retries;
- a non-`ErrNoRows` infrastructure error from the append fence;
- handler branches for a disabled deployment, a non-user Agent, and route/Agent diagnosis after reassignment returns no rows.

Please add automated tests for these branches and assert that failures cannot cause cross-Agent session reuse or an incorrect append.

Distinguishing routes for different DingTalk bots in the same group is no longer a blocking item for this PR. It can be handled separately in a follow-up PR together with a real, readable bot identity, avoiding further expansion of the current scope. You are very welcome to send that follow-up PR, and I will be happy to review it again.

Please ping me after these two test items are addressed. Barring anything unexpected, this should be the final change-request round for this PR.

mxbao063-bao:
@yyclaw, the two remaining coverage items are addressed at the latest head (`bcadc145e`).

- Shared Router coverage now includes EnsureSession conflict refresh/retry with the same claim, archived/infrastructure refresh failures from both EnsureSession and append conflicts, missing Validated resolver, append-lock failure, and repeated conflicts terminating through context cancellation. Every terminal path asserts Mark/Release behavior and that nothing is appended to the old route.
- These tests exposed a real cancellation bug: final dedup Release could inherit an already-canceled request context. Finalization now uses a bounded detached context, with a regression fake that respects context cancellation.
- DingTalk coverage now includes P2P discovery no-op/default-Agent direct flow; decode/discovery failures; initial/recheck query, stale-binding cleanup, and EnsureSession failures; post-session conflict and three-retry exhaustion; non-NoRows append-fence infrastructure errors; and disabled deployment, non-user Agent, workspace mismatch, revoke/archive/no-row handler diagnosis. Failure tests assert no cross-Agent session reuse or incorrect append.
- Latest `main` is merged. The DingTalk migrations were renumbered to 285-288 to avoid the new upstream migration numbers; concurrent-index recovery hooks and INVALID/valid-leftover tests were updated and pass.

Verification: full GitHub CI is green (backend, frontend build/test, Mobile, Windows execenv, and all installer jobs). Local full Views suite passed (326 files / 3862 tests), typecheck, lint, build, focused Go packages, migration recovery/up-down tests, vet, and focused race tests all passed.

Please rerun the final review when ready.

yyclaw:
> * Latest `main` is merged. The DingTalk migrations were renumbered to 285-288 to avoid the new upstream migration numbers; concurrent-index recovery hooks and INVALID/valid-leftover tests were updated and pass.

@mxbao063-bao Please rebase the PR branch onto the latest main instead of merging main into the feature branch. This repository does not currently document rebase as a hard requirement, but for this long-running PR, keeping a linear branch history avoids bringing unrelated upstream history into the review and makes the final changes easier to audit. Please resolve the conflict during the rebase and force-push the updated branch.

yyclaw (approved):
@mxbao063-bao Thank you for staying with the review through multiple rounds and for completing the final test coverage.

I completed the final preflight on the rebased head (`d28e51eb`). The shared route-conflict terminal-state matrix, DingTalk cutover and handler failure paths, migration recovery and numbering, cleanup paths, API compatibility, UI states, and adjacent DingTalk, Lark, and Slack regressions all pass. I found no remaining material blockers.

The two previously agreed product decisions remain unchanged:

- Agent A → B → A starts a new chat session.
- Distinguishing multiple DingTalk bot identities in the same group can be handled in a follow-up PR.

Approved. Thank you again for the contribution.

PR #6662, opened 2026-08-09
DAT-131: fix(issue): improve run history and collapsed layout

Phirmu:
## What does this PR do?

Improves the issue detail layout and run-history access requested by DAT-131.

The issue content column now expands from `max-w-4xl` to `max-w-6xl` when the properties sidebar is collapsed. Historical agent runs move from the sidebar execution log into the header activity chip, keeping history reachable even when the sidebar is closed. The chip shows a calm `Historical runs` state when no task is active, while active and historical entries share the top-level list and historical rows remain nested below their toggle.

Task status filtering and past-run sorting are shared between the header and execution-log components so both surfaces use the same definitions. The main residual risk is visual density with unusually long localized labels; the chip and rows retain their existing truncation constraints.

## Related Issue

Closes DAT-131

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Updated `packages/views/issues/components/issue-detail.tsx` to widen the content column when the properties sidebar is collapsed and keep only active runs in the sidebar execution log.
- Updated `packages/views/issues/components/issue-agent-header-chip.tsx` to expose historical runs in terminal-only and active-task states, without the animated border when nothing is running.
- Shared active/past task derivation and the historical row renderer from `execution-log-section.tsx`.
- Added EN, zh-Hans, JA, and KO translations for the historical-runs label.
- Added regression coverage for width switching, history visibility, static styling, expansion, and nested list structure.

## How to Test

1. Run `pnpm --filter @multica/views exec vitest run issues/components/issue-agent-header-chip.test.tsx issues/components/execution-log-section.test.tsx issues/components/issue-detail.test.tsx`.
2. Run `pnpm --filter @multica/views exec vitest run locales/parity.test.ts`.
3. Run `pnpm --filter @multica/views typecheck` and `pnpm --filter @multica/views lint`.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes (not applicable; no documentation contract changed)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable)

## AI Disclosure

**AI tool used:** Codex in Multica

**Prompt / approach:** Implemented DAT-131 from the issue requirements, reused the existing issue task query and row components, then added focused component tests for each requested interaction and ran the views quality gates.

## Screenshots (optional)

Not included. The behavior is covered by component tests, but a visual capture requires an issue fixture containing both active and terminal agent runs.


ADanMan (commented):
Requesting changes on the usage-reporting regression this introduces.

`ExecutionLogSection` is the only host of both `IssueUsageTotal` (the issue-level aggregate cost, summed over all tasks in the header) and `IssueUsageDialog` (the per-run cost breakdown it opens). Passing `showPast={false}` from `issue-detail.tsx` changes the mount guard to:

```
if (activeTasks.length === 0 && (!showPastRuns || pastTasks.length === 0)) return null;
```

With `showPastRuns` false, `!showPastRuns` is always true, so this reduces to `activeTasks.length === 0`. That means for any issue whose runs are all terminal — the normal steady state once agents finish — the section unmounts and takes the aggregate cost total and the breakdown dialog with it.

The header chip that now carries history renders `PastTaskRow` only; it doesn't render `IssueUsageTotal` or `IssueUsageDialog`. So each past row keeps its own token figure, but the issue-level total ("what has this issue spent") and the click-through breakdown have no other home in the UI and vanish for completed issues.

A couple of options:
- Keep an aggregate-cost affordance reachable from the header chip (e.g. surface `IssueUsageTotal`/the dialog there alongside the history rows), or
- Keep the section mounted for the usage summary while moving only the run rows, so the total/dialog don't depend on an active run.

Either way, could you also call this behavior change out in the PR description? Right now it reads as history simply "moving" to the header chip, and the loss of the issue-cost total and breakdown dialog for terminal issues isn't mentioned.

PR #6663, opened 2026-08-09
fix(inbox): cap the active inbox list at 200 rows (#6527)

### LINKED ISSUE #6527: [Bug]: GET /api/inbox is unbounded and full-refetches on every mark-read, causing multi‑MB / multi‑second loads for heavy inboxes
### Deployment type

self-host

### What happened?

`GET /api/inbox` returns **all non-archived inbox rows** for the current user/workspace with **no LIMIT / pagination**. For users with a large backlog this produces multi‑MB JSON (observed ~8MB) and multi‑second responses (~6–7s).
Clicking an inbox item auto-marks it read. That triggers:
1. mutation `onSettled` → `invalidateQueries(inboxKeys.all)` → full `GET /api/inbox` refetch
2. server WS `inbox:read` → realtime `onInboxInvalidate` → often a **second** full refetch
This looks like “polling” in DevTools, but there is **no** `refetchInterval` — it is invalidate-driven full-list reloads. The UI only shows one row per issue (`deduplicateInboxItems`), but the API still downloads every sibling row.
Related / not duplicates:
- #1968 — Inbox click freeze (timeline render path; adjacent symptom, different root cause)
- #4540 — wants WS insert without full refetch (architecture debt; not this payload bug)

### Steps to reproduce
1. Use an account whose workspace has a large **non-archived** inbox (thousands of `inbox_item` rows / multi‑MB list).
2. Open Inbox; note `GET /api/inbox` size and latency in Network.
3. Click an unread notification.
4. Observe repeated `GET /api/inbox` calls, each returning the full multi‑MB payload again.
### Expected behavior
- List endpoint should be bounded (LIMIT / cursor pagination) and/or return a slim list projection (title + metadata; body on demand).
- Mark-read / `inbox:read` should patch the cache (or refetch a cheap unread summary), not re-download the entire inbox.
- Unread badge should not require the full list payload (e.g. use `/api/inbox/unread-count` or `/api/inbox/unread-summary`).
### Actual behavior
- `ListInboxItems` has no LIMIT (archived list correctly uses `LIMIT 200`).
- Client comments assume “the inbox list is small so this is cheap” (`packages/core/inbox/ws-updaters.ts`) — false for heavy inboxes.
- Each open/mark-read can cost another full multi‑MB round-trip.

### Screenshots (optional)

Image

```shell
Code pointers:
- `server/pkg/db/queries/inbox.sql` — `ListInboxItems` unbounded; `ListArchivedInboxItems` capped at 200
- `server/internal/handler/inbox.go` — `ListInbox`
- `packages/core/inbox/mutations.ts` — `useMarkInboxRead` `onSettled` invalidates full inbox
- `packages/core/inbox/ws-updaters.ts` — invalidate-on-every-inbox-event
- `packages/views/inbox/components/inbox-page.tsx` — auto mark-read on select
```

kakiuwang-ui:
## Problem

`GET /api/inbox` runs `ListInboxItems` with **no `LIMIT`**, returning every non-archived `inbox_item` row for the recipient. For a heavy backlog this is a multi-MB / multi-second response (the reporter observed ~8MB / ~6–7s). Fixes #6527.

The archived counterpart already anticipated exactly this — `ListArchivedInboxItems` caps at `LIMIT 200` with a documented rationale — but the **active** list, which is the one loaded on every inbox open and re-fetched on every mark-read invalidation, was left unbounded.

## Fix

Cap `ListInboxItems` at 200 rows, mirroring the archived query and reusing its safety argument verbatim:

- rows are `ORDER BY created_at DESC` (newest-first);
- the UI groups one row per issue carrying that group's **newest** item (`deduplicateInboxItems` / `groupInboxItemsByIssue`);
- so truncating the oldest rows can never hide a group's rendered newest row.

Single caller (`ListInbox` handler), constant `LIMIT` (no new bind param), so the generated code change is just the query string.

## Test

`TestListInboxItems_CapsUnboundedActiveList` (DB-backed) inserts 210 active rows and asserts the query returns the newest 200 and drops the oldest 10. It **fails without the cap** (returns all 210) and passes with it.

## Known interaction — active-workspace unread badge (deliberately left for a follow-up)

`useInboxUnreadCount` (`packages/core/inbox/queries.ts`) currently derives the **active workspace's** unread badge from this same `listInbox()` result — `deduplicateInboxItems(items).filter(!read).length`. With this cap, a workspace whose active inbox exceeds 200 raw rows will **undercount** that badge (desktop dock badge, mobile tab badge, inbox page count), since unread items older than the 200th row are no longer in the fetched list.

I scoped this PR to the server payload bound (the multi-MB root cause) on purpose. The correct client fix is to back that badge with the server-computed per-workspace count that already exists (`CountUnreadInboxByWorkspace`, exposed via the unread-summary endpoint — same per-issue dedup semantics), rather than deriving it from the render list. That touches the **deliberately optimistic** badge path in `inbox/mutations.ts` (and the separate mobile implementation), so it belongs in its own reviewable change rather than riding along with a one-line query cap. Happy to send that follow-up PR — just didn't want to fold a cross-platform badge-architecture change into the payload fix without a maintainer's call on the tradeoff.

The **cross-workspace** switcher dot is unaffected: it already reads `CountUnreadInboxByWorkspace`, not the list.


kakiuwang-ui:
Flagging an overlap so this doesn't get reviewed twice: #6811 lands the same 200-row cap on `ListInboxItems` for #6527. No objection to either landing — just noting the difference so you can pick quickly.

- This PR (opened 8-09) caps the query and asserts the bound at the **handler** level in `server/internal/handler/inbox_list_cap_test.go`, i.e. through `GET /api/inbox` as the client actually sees it.
- #6811 (opened 8-12) makes the same query change with the assertion in `server/cmd/server/inbox_list_limit_test.go`.

Both are green and conflict-free. If you would rather take #6811, I am happy to close this one — the only thing worth carrying over is the handler-level test, which is where the repo's testing rules put an endpoint-visible bound.

PR #6666, opened 2026-08-09
MUL-5951: fix(issues): reject a cross-workspace project on issue update

### LINKED ISSUE #6665: [Bug]: issue update accepts a project from another workspace
### Deployment type

self-host

### What happened?

`PUT /api/issues/{id}` accepts a `project_id` that belongs to a different workspace. The request succeeds with 200 and the issue is written with a project from another workspace, so `issue.workspace_id` and `project.workspace_id` diverge.

The issue then disappears from the board of the project's workspace (that board lists by workspace) and hangs off a project that its own board cannot resolve.

I expected the update to be rejected with 400 and a clear message, the way the create path and the move endpoint already reject a foreign project. Moving an issue across workspaces is a different capability and is not what I was after.

`POST /api/issues/{id}/move` already rejects this with "project not found in this workspace", and `POST /api/issues` validates it on create. Only the update path is missing the check.

### Steps to reproduce

Two workspaces A and B, each with one project.

1. `multica issue update <ISSUE-IN-WORKSPACE-A> --project <PROJECT_ID-IN-WORKSPACE-B>`
2. The command returns success, no warning.
3. Query the database:

```sql
select i.workspace_id, p.workspace_id
from issue i join project p on p.id = i.project_id
where p.workspace_id <> i.workspace_id;
```

The row is there.

### Additional context

Reproduced on `v0.4.21`. The `project_id` block in `UpdateIssue` (`server/internal/handler/issue.go`) is unchanged on `main` at `b09abf179`: it parses the UUID and assigns it without any workspace lookup, unlike the `parent_issue_id` block right above it, which calls `GetIssueInWorkspace`.

`POST /api/issues/batch-update` reaches the same field through its own code path and has the same gap. Its per-issue error handling skips invalid updates instead of failing the request, so the right behavior there is a separate decision.

fgmacedo:
## What does this PR do?

`UpdateIssue` accepted a `project_id` from any workspace: it parsed the UUID and assigned it, with no lookup. The write then produced an issue whose `workspace_id` differs from its project's `workspace_id`. That issue disappears from the board of the project's workspace, which lists by workspace, and points at a project its own board cannot resolve.

Every neighboring path already enforces this boundary. The `parent_issue_id` block immediately above calls `GetIssueInWorkspace` and answers `parent issue not found in this workspace`; `MoveIssue` pre-checks the project against `current.WorkspaceID`; `IssueService.Create` enforces it atomically on create. Only the update path was missing it.

The fix validates the project against `prevIssue.WorkspaceID` with the existing `GetProjectInWorkspace` query and returns 400 with `project not found in this workspace`, the message `MoveIssue` and `service.ErrProjectNotFound` already use.

`UpdateIssue` is the canonical write for both `PUT /api/issues/{id}` and `POST /api/issues/{id}/move`, which rewrites the body and delegates here. So the check closes the PUT path and leaves the move path with a redundant, harmless second check rather than a behavior change.

Transferring an issue across workspaces is a different capability and is not part of this change.

### One surface deliberately left out

`POST /api/issues/batch-update` reaches `project_id` through its own code path (`BatchUpdateIssues`) and has the same gap. That path skips an invalid per-issue update with `continue` instead of failing the request, so whether a foreign project should silently skip the issue or reject the call is a product decision, not a mechanical port of this check. I flagged it in the issue and left it alone here.

## Related Issue

Closes #6665

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/issue.go`: the `project_id` branch of `UpdateIssue` now looks the project up with `GetProjectInWorkspace` scoped to `prevIssue.WorkspaceID` and returns 400 when it is not there. Clearing the project with an explicit `null` is untouched, and an update that does not send `project_id` still carries the current value through `refreshUntouchedNullableIssueParams` without a lookup, so pre-existing rows are not newly rejected.
- `server/internal/handler/issue_update_project_scope_test.go`: `TestUpdateIssueProjectStaysInWorkspace`, both directions on the PUT path.

## How to Test

1. Two workspaces, each with a project. `PUT /api/issues/{id}` with the foreign project's `project_id` returns 400 instead of 200, and the row keeps its previous project.
2. The same request with a project from the issue's own workspace still returns 200 with `project_id` echoed back.
3. Backend suite:

```bash
cd server && go run ./cmd/migrate up
bash scripts/test-go.sh --race
```

The new test fails on `main` (200 on the cross-workspace case) and passes with this change. The existing `TestMoveIssueRejectsUnsafeInputs/cross-workspace_project` and `TestCreateIssueRejectsCrossWorkspaceProject` still pass, so the move and create paths are unaffected. `cd server && go build ./...` and `go vet ./internal/handler/` are clean, and `gofmt -l` reports nothing for the two touched files.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (no UI change)
- [ ] I have updated relevant documentation to reflect my changes (no documented behavior changes)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs (not applicable)
- [ ] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` (not applicable)

Risk: a client that today moves an issue to a project in another workspace now gets 400. That is the point of the change, and no in-workspace flow reaches it. One thing worth a maintainer's eye: like the `parent_issue_id` check above it, any error from the lookup, including a transient database failure, surfaces as 400 rather than 500. I matched the neighbor instead of introducing a third error shape in the same function.

I agree to the Contribution Terms in `CONTRIBUTING.md`.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** I hit this on a self-hosted `v0.4.21` instance and asked Claude Code to confirm the gap on `main` before writing anything, to check whether `MoveIssue` really shares the write path its comment claims, to write the failing test first against a local Postgres and only then the fix, and to keep the fix to the shape the surrounding validations already use. The judgment calls I would most like a maintainer to check are leaving `batch-update` out and reusing the neighbor's 400-on-any-lookup-error behavior.

### DIFF
--- a/server/internal/handler/issue.go
+++ b/server/internal/handler/issue.go
@@ -3019,6 +3019,13 @@ func (h *Handler) UpdateIssue(w http.ResponseWriter, r *http.Request) {
 			if !ok {
 				return
 			}
+			if _, err := h.Queries.GetProjectInWorkspace(r.Context(), db.GetProjectInWorkspaceParams{
+				ID:          projectUUID,
+				WorkspaceID: prevIssue.WorkspaceID,
+			}); err != nil {
+				writeError(w, http.StatusBadRequest, "project not found in this workspace")
+				return
+			}
 			params.ProjectID = projectUUID
 		} else {
 			params.ProjectID = pgtype.UUID{Valid: false}
--- a/server/internal/handler/issue.go
+++ b/server/internal/handler/issue.go
@@ -3023,6 +3023,12 @@ func (h *Handler) UpdateIssue(w http.ResponseWriter, r *http.Request) {
 				ID:          projectUUID,
 				WorkspaceID: prevIssue.WorkspaceID,
 			}); err != nil {
+				if !isNotFound(err) {
+					slog.Error("update issue: validate project scope",
+						append(logger.RequestAttrs(r), "project_id", uuidToString(projectUUID), "error", err)...)
+					writeError(w, http.StatusInternalServerError, "failed to validate project")
+					return
+				}
 				writeError(w, http.StatusBadRequest, "project not found in this workspace")
 				return
 			}
@@ -3487,6 +3493,31 @@ func (h *Handler) BatchUpdateIssues(w http.ResponseWriter, r *http.Request) {
 	if !ok {
 		return
 	}
+	// The batch shares one project_id, so it is checked once here rather than
+	// per issue, and rejected instead of skipped like the per-item guards in
+	// the loop: a foreign project invalidates the whole request.
+	batchProjectID := pgtype.UUID{Valid: false}
+	if _, ok := rawUpdates["project_id"]; ok && req.Updates.ProjectID != nil {
+		projectUUID, ok := parseUUIDOrBadRequest(w, *req.Updates.ProjectID, "project_id")
+		if !ok {
+			return
+		}
+		if _, err := h.Queries.GetProjectInWorkspace(r.Context(), db.GetProjectInWorkspaceParams{
+			ID:          projectUUID,
+			WorkspaceID: wsUUID,
+		}); err != nil {
+			if !isNotFound(err) {
+				slog.Error("batch update issues: validate project scope",
+					append(logger.RequestAttrs(r), "project_id", uuidToString(projectUUID), "error", err)...)
+				writeError(w, http.StatusInternalServerError, "failed to validate project")
+				return
+			}
+			writeError(w, http.StatusBadRequest, "project not found in this workspace")
+			return
+		}
+		batchProjectID = projectUUID
+	}
+
 	updated := 0
 	// Children that transitioned into a terminal status this batch, collected so
 	// the parent/stage notification is evaluated once against the final state
@@ -3612,15 +3643,8 @@ func (h *Handler) BatchUpdateIssues(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 		if _, ok := rawUpdates["project_id"]; ok {
-			if req.Updates.ProjectID != nil {
-				projectUUID, err := util.ParseUUID(*req.Updates.ProjectID)
-				if err != nil {
-					continue
-				}
-				params.ProjectID = projectUUID
-			} else {
-				params.ProjectID = pgtype.UUID{Valid: false}
-			}
+			// Resolved before the loop; an explicit null stays invalid and clears.
+			params.ProjectID = batchProjectID
 		}
 		if _, ok := rawUpdates["stage"]; ok {
 			if req.Updates.Stage != nil {
--- a/server/internal/handler/issue_update_project_scope_test.go
+++ b/server/internal/handler/issue_update_project_scope_test.go
@@ -1,0 +1,121 @@
+package handler
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestUpdateIssueProjectStaysInWorkspace mirrors
+// TestCreateIssueRejectsCrossWorkspaceProject on the update path. UpdateIssue
+// is the canonical issue write — the legacy PUT endpoint and MoveIssue both
+// land on it — so the project boundary belongs here and not only on the entry
+// points that happen to pre-check it. Accepting a foreign project would leave
+// the issue pointing at a project from another workspace: it disappears from
+// that workspace's board, which lists by workspace, and its own board cannot
+// resolve the project.
+func TestUpdateIssueProjectStaysInWorkspace(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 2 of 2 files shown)


multica-eve (changes requested):
Before this change, a user can attach an issue in one workspace to a project in another workspace and receive a success response, leaving the issue hidden or inconsistently categorized. Rejecting that operation while preserving normal same-workspace updates is the product behavior we want. ✅ Product direction accepted.

Requesting changes for one blocking gap: `BatchUpdateIssues` still accepts the same foreign `project_id`. In `server/internal/handler/issue.go:3614-3620`, the batch path parses the UUID and assigns it without a `GetProjectInWorkspace` check, so `POST /api/issues/batch-update` can still create the exact cross-workspace inconsistency that this PR is intended to prevent.

The batch response contract does not need to be settled before enforcing the boundary. Either validate the shared `project_id` once before the loop and reject the request with 400, or preserve the existing per-item skip behavior and return `updated: 0`; in either case, a foreign project must never be written. Please add a regression test proving that a foreign project changes none of the target rows.

Non-blocking: `server/internal/handler/issue.go:3025` maps every `GetProjectInWorkspace` error to 400. Distinguishing `pgx.ErrNoRows` from an internal database error would improve correctness and observability, though the current code fails closed and matches the adjacent parent validation.

The single-issue implementation itself is scoped correctly: it derives the workspace from the authorized issue, preserves null/untouched behavior, and its positive/negative regression test verifies that a rejected update leaves the stored project unchanged. Targeted handler tests, `go build ./...`, `git diff --check`, and all nine GitHub CI checks pass.

fgmacedo:
Addressed in c3b21eb68.

**Blocking — `BatchUpdateIssues` accepted the same foreign `project_id`.** Confirmed and fixed. I took the first option you offered: the shared `project_id` is resolved and workspace-checked once before the loop, and the request answers 400 `project not found in this workspace`, the same message the single-issue path uses. The batch carries one `project_id` for every target and every target is loaded scoped to `wsUUID` inside the loop, so the boundary is a property of the request, not of each row.

I picked reject over per-item skip for a reason beyond the response contract: `useBatchUpdateIssues` patches the cache optimistically and rolls back in `onError`. A 200 with `{"updated": 0}` leaves the optimistic project on screen while the server wrote nothing, so the silent-skip variant would trade a wrong row for a wrong screen. Only an error response rolls it back.

That move pulled the parse out of the loop as well, so an unparseable `project_id` now answers 400 through `parseUUIDOrBadRequest` instead of skipping every issue and reporting `{"updated": 0}` with a 200. Calling that out explicitly as a deliberate contract change beyond the reported bug: it is the handler convention in CLAUDE.md for a pure UUID taken from a request boundary, it matches the status/priority enums validated a few lines above, and the old behavior was an accident of the field being handled per item. Clients clear a project with `null`, which still takes the untouched `pgtype.UUID{Valid: false}` path.

Regression test: `TestBatchUpdateIssuesProjectStaysInWorkspace` sets both target issues to a local project first so the rejection has a real value to preserve, then asserts 400 and that neither row changed. Verified it fails on the pre-fix code with `expected 400 for foreign project, got 200: {"updated":2}` — both rows had been rewritten.

**Non-blocking — every `GetProjectInWorkspace` error mapped to 400.** Fixed on both paths. Only `pgx.ErrNoRows` now means out-of-scope; anything else logs and returns 500. A dead connection answering 400 sends a correct client rewriting a correct payload while the real failure never surfaces as a server error. This is the split `resolveMoveAnchor` in `issue_move.go` already makes on the same kind of boundary, and `isNotFound` already exists in the package, so it does not introduce a new idiom. The adjacent `parent_issue_id` check still collapses both cases; I left it alone as pre-existing and out of this PR's scope.

**Adversarial self-review of the fix.** Two things I checked and one gap I am leaving open:

- Swept every remaining write of an issue's `project_id` for the same hole: quick-create (`issue.go:2269`), `IssueService` create (`service/issue.go:224`), `UpdateIssue`, and now batch all validate; `MoveIssue` delegates to `UpdateIssue`; `issue_table_group.go` only builds a read-side descriptor. No third door.
- Ordering: with both an invalid status and a foreign project, the enum check still wins. Intentional, matches the existing precedence.
- Gap: the `ErrNoRows`-vs-internal-error split is not covered by a test. The handler suite runs against a real pool and `Handler.Queries` is a concrete type, so injecting a failing query would mean a seam this PR does not otherwise need.

`go build ./...`, `go vet`, `gofmt`, `git diff --check` clean; `go test -race ./internal/handler/` passes in full. The rest of the Go suite passes too, except `cmd/multica`, which fails identically on the unmodified tree here because my sandbox is itself an agent task and the CLI guard sees the daemon task marker — environmental, not from this change. GitHub CI is running on the pushed commit.

multica-eve (approved):
Approved after re-review of 93c65acd: the batch-update workspace-scope validation and regression coverage address the previous blocking concern.

PR #6677, opened 2026-08-10
MUL-5960: NEX-47: daemon-core safe shutdown (drain): three-state + claim-pause ref-count + persistence + decoupled shutdown + /drain endpoint

HuAzurestar:
Part of NEX-38 (safe shutdown / drain). Daemon-core subtask (NEX-47, stage 1). Server-side (migration 265 + registration/heartbeat keep-draining + AgentReadiness + runtime_drained) and CLI/UI land in the sibling subtasks.

## What

- **Three-state daemon state** (`daemonState`: running / draining / shutting_down) surfaced on `/health` as `state`.
- **`pauseClaims` → ref-count** (decision two): holders `drain` / `server-update` / `below-minimum` acquire/release independent refs under `claimMu`, so one holder can never clear another's pause. `tryEnterClaim` / `trySetClaimBarrier` / `tryBeginServerUpdate` / `releaseClaimBarrier` rewired; `demoteBelowMinimumRuntimes` uses its own `below-minimum` ref. During drain, auto-update / demotion defer.
- **Drain entry points**: `beginDrain` / `abortDrain` / `finishDrainThenStop`, plus a `POST /drain` health endpoint (drain / abort / finish_then_stop), 127.0.0.1-local like `/shutdown`.
- **Persistence (AC-13)**: `<workspaces-root>/drain.json` written before the in-memory flip; startup restores draining before the first registration so the register payload reports `status:"draining"` instead of defaulting back to online.
- **Decoupled shutdown (AC-4/AC-5)**: a drain monitor deregisters runtimes and cancels the root context once draining + finish-then-stop + `activeTasks==0`; the ordinary stop path (immediate cancel + 30s poll-loop cap) is untouched.
- **Health fields**: `state`, `draining_inflight_tasks` (== `active_task_count`), `draining_queued_tasks` (cached from heartbeat acks via a new additive `queued_tasks` field on `DaemonHeartbeatAckPayload`).

## Tests

New `daemon_drain_test.go` (state transitions, drain pauses/resumes claims, abort clears persistence + finish-then-stop flag, restart restore, drain monitor auto-stop vs resident-drain, `/drain` handler, health state/counts, queued-task cache) and `claim_pause_refcount_test.go` (independent-holder ref-count, concurrent interleaving, drain-vs-barrier defer). Updated the barrier tests that referenced the removed `pauseClaims` field.

Verified locally with Go 1.26.1: `go build ./...`, `go vet ./internal/daemon/...`, and the daemon / daemonws test suites pass. The 55 pre-existing failures in the daemon package are Windows-environment-specific (git path-length, symlink privilege, home-dir skill/MCP fixtures, missing fake CLIs) and are byte-identical on pristine `main`; `-race` is unavailable without cgo. CI `make check` remains the gate.

### DIFF
--- a/server/cmd/multica/cmd_daemon.go
+++ b/server/cmd/multica/cmd_daemon.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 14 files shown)


HuAzurestar (commented):
NEX-52 corrected-contract daemon fixes are in (head 2dd78b718). Rebased onto current main to resolve the MUL-5742 task-scoped-status merge; CI is green (backend, frontend, installers, execenv all success). Draining now keeps claiming pre-boundary queued work (claim-pause decoupled for claim entry only; auto-update/demotion still defer), and the drain monitor waits for both in-flight and queued counts to reach zero before finish-then-stop.

PR #6678, opened 2026-08-10
MUL-5961: feat(desktop): daemon draining UI (three-state) + four-language i18n (NEX-38/NEX-49)

HuAzurestar:
### Summary
NEX-49 (NEX-38 S3): UI ???? + ?? i18n?

- daemon-types: DaemonState ? draining,??/??/description/statusAlive + DaemonDrainAction ??
- IPC: preload + main ???? daemon:drain,??????? /drain ?? (drain/abort/finish_then_stop)
- daemon ??????: running [View logs] [Restart] [????] [Stop];draining [??????] [????]+??;off ? [Start]
- ?? runtimes ?????????????;/health state ?? draining;IPC bridge ?? draining
- core: RuntimeDevice.status + mobile zod schema ?? draining;deriveRuntimeHealth ? draining ???
- i18n: en/zh-Hans/ja/ko runtimes.json ?? 8 ?,parity.test ??

### ??
- pnpm typecheck(web/desktop/core/views/mobile)??
- packages/core 1354 tests?packages/views runtimes 191 tests?apps/desktop daemon ?? 10 tests?locales parity 177 tests ??
- apps/desktop ? scripts/package.test.mjs / 
untime-config-loader.test.ts ??? Electron ?????(????,???????)

### DIFF
--- a/apps/desktop/src/main/daemon-manager.ts
+++ b/apps/desktop/src/main/daemon-manager.ts
@@ -20,6 +20,7 @@ import type {
   DaemonStatus,
   DaemonPrefs,
   LocalRuntimeProbe,
+  DaemonDrainAction,
 } from "../shared/daemon-types";
 import { daemonStatusAlive } from "../shared/daemon-types";
 import { ensureManagedCli, managedCliPath } from "./cli-bootstrap";
@@ -149,6 +150,10 @@ function sendStatus(status: DaemonStatus): void {
 
 interface HealthPayload {
   status?: string;
+  /** Daemon three-state: running / draining / stopped (NEX-38). The daemon
+   *  keeps reporting `status: "running"` while draining — readiness is
+   *  orthogonal to the drain state, so the two fields are both read. */
+  state?: string;
   pid?: number;
   /** Daemon's runtime.GOOS. Absent on daemons older than the #3916 fix. */
   os?: string;
@@ -355,7 +360,7 @@ async function fetchHealth(): Promise<DaemonStatus> {
   }
 
   return {
-    state: "running",
+    state: data.state === "draining" ? "draining" : "running",
     pid: data.pid,
     uptime: data.uptime,
     daemonId: data.daemon_id,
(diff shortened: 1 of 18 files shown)


ADanMan (commented):
Two things to sort out before this lands.

**1. The drain controls have no backend to talk to on `main`.**

`drainDaemon()` in `daemon-manager.ts` POSTs to `http://127.0.0.1:{port}/drain`, and `fetchHealth()` now reads `data.state === "draining"`. But on `origin/main` the daemon health server (`server/internal/daemon/health.go`) registers only `/health`, `/shutdown`, and `/repo/checkout` — there is no `/drain` route, and `HealthResponse` exposes a `status` field but no `state` field (nothing in `server/` ever reports `"draining"`). As it stands:

- "Safe shutdown" / "Cancel shutdown" / "Shut down now" all hit a route the daemon doesn't serve, so Go's default mux returns 404, `!res.ok` is true, and the user gets the error toast (`Failed to start safe shutdown`, etc.).
- `fetchHealth` never sees a `state`, so `data.state === "draining"` is always false — the daemon can't enter the draining branch and the entire draining UI is unreachable.

If the `/drain` endpoint and the `state: "draining"` health field are coming in a companion change, could you note the dependency / merge order in the description (and link it)? Otherwise this ships as error-only buttons and a state that never occurs. If it's meant to land in an earlier stage of the stack, a pointer is enough.

**2. The force-stop dialog contradicts itself (and the action it fires).**

`DrainForceStopDialog` fires `drain("finish_then_stop")`, and `finish_then_stop` is documented as "shuts the daemon down once in-flight tasks complete" — tasks are allowed to finish. `drainDescription`, rendered in the same dialog, agrees: "…the daemon shuts down once running tasks finish." But the dialog title `drainConfirmTitle` reads "This will interrupt {{count}} running task(s)." and the component comment says force-stop "interrupts those tasks." So the title claims interruption while the body and the action promise completion — and the same mismatch is present in zh-Hans/ja/ko.

Please pick one: if "Shut down now" is meant to wait for in-flight tasks, the title shouldn't say "interrupt" (e.g. "Shut down after {{count}} running task(s) finish?"); if it's meant to kill them immediately, then `finish_then_stop` and `drainDescription` are the wrong semantics. As written, a user can't tell from the dialog what the button will actually do.

Issue #100, opened 2026-01-23
fix: MessageTimer freezes during long pauses


NevilleQingNY:
## Problem

The MessageTimer appears to "freeze" during streaming:
- Timer only updates when ACP events arrive (endTime changes trigger re-render)
- During long pauses (waiting for permissions, tool execution), the display stalls
- User experience: `1.1s` → stalls for 30s → suddenly jumps to `31.6s`

Note: The final time calculation is correct (endTime - startTime). The issue is only the **display freezing during processing**.

## Solution

Event-driven updates + timeout projection:
- When events arrive: show exact elapsed time from props
- When no events for 500ms: project time forward with variable interval (80-180ms)
- When events resume: immediately sync to accurate value

PR #6688, opened 2026-08-10
fix(channels): skip media for duplicate issue commands

### LINKED ISSUE #6687: [Bug]: Repeated /issue messages duplicate image attachments
### Deployment type

Official App and self-host

### What happened?

In channel integrations that support inbound media (currently DingTalk and Feishu), sending the same image-and-`/issue` message multiple times correctly reports that an active issue with the same title already exists and does not create another issue. However, every repeated command resolves and persists the media again, then associates the new attachment rows with the existing issue.

The duplicate outcome should not download, upload, or persist media because no new issue was created. The existing issue's attachment set should remain unchanged.

### Steps to reproduce

1. Connect a channel integration with inbound media support, such as DingTalk or Feishu, to a workspace.
2. Send a message containing an image and `/issue Repeated image report`.
3. Confirm that the issue is created with one image attachment.
4. Send the same image and `/issue Repeated image report` again as a new channel message.
5. Observe the active-duplicate reply and inspect the existing issue's attachments.

### Screenshots (optional)

N/A

The shared channel Router treats the active-duplicate result as terminal but still runs the normal issue-media resolution path against `DuplicateIssue`. Each new inbound message has a new durable message ID and therefore produces new storage objects and attachment IDs even when the logical image and issue title are unchanged. DingTalk and Feishu both register this shared MediaResolver path; Slack and WeCom currently do not support inbound media through it.

yyclaw:
## What does this PR do?

Stops active-duplicate `/issue` commands from resolving and persisting inbound media. Repeating the same image-and-command message through a media-capable channel now keeps the existing issue's attachment set unchanged while preserving the duplicate reply and clearing the durable media-pending marker.

The duplicate decision remains authoritative in `IssueService.Create`. The shared channel Router now maps that terminal outcome to an immediate empty media finalization instead of targeting `DuplicateIssue`; successful issue creation still follows the existing resolve, bind, attachment-event, and deferred-task promotion path.

The affected production paths are DingTalk and Feishu, which both register inbound MediaResolvers with the shared engine. Slack and WeCom currently register no inbound media resolver, so their behavior is unchanged.

## Related Issue

Closes #6687

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Updated `server/internal/integrations/channel/engine/router.go` to skip platform media resolution for active-duplicate issue outcomes and immediately clear the command message's pending state without waiting behind the session media queue.
- Restricted the media-resolution timeout warning to paths that actually attempt remote resolution.
- Updated `server/internal/integrations/channel/engine/router_test.go` to verify that duplicate commands do not invoke the resolver, persist media refs, target the existing issue, publish attachment events, or promote an issue task; that finalize-only work bypasses blocked same-session media; and that an expired finalize-only deadline does not emit a false media-resolution warning.
- Added `TestBindMediaRefs_EmptyRefsCreateNoAttachmentAndClearPending` in `server/internal/integrations/channel/engine/session_db_test.go`, proving that empty finalization creates no attachment rows, leaves the existing issue attachment count unchanged, and clears `channel_media_pending_until`.

## How to Test

1. In DingTalk or Feishu, send an image with `/issue Repeated image report`; confirm the issue is created with one attachment.
2. Send the same image and command again; confirm the active-duplicate reply appears and the issue still has one attachment.
3. Run:

   ```bash
   cd server
   go test -race ./internal/integrations/channel/engine -count=1
   go vet ./internal/integrations/channel/engine
   go test ./internal/service -run '^(TestEnqueueChatTaskDefersForChannelMediaAndPromotesWhenReady|TestPromoteChannelChatTasksIgnoresHandledCommandMedia|TestDeferredChannelIssueTaskPromotesAfterMediaSettlement|TestChannelMediaReconciler_SettleInvariantDwarfsPipelineBudgets)$' -count=1
   ```

## Checklist

Risk is limited to server-side channel media handling. Duplicate commands intentionally discard their media because no new issue consumes it; normal issue creation and ordinary chat media paths are unchanged. There are no schema, API, protocol, client, or rollout changes.

## AI Disclosure

**Prompt / approach:** Investigated the shared channel Router and attachment transaction, reproduced the duplicate target in a regression test, revised the approach after identifying that chat-owned attachments would remain unused, implemented an empty media-finalization path, and ran a `pr-preflight` review across duplicate authority, lifecycle, concurrency, platform impact, privacy, and test Oracle.

## Screenshots (optional)

N/A — server-side behavior change with no UI modification.

### DIFF
--- a/server/internal/integrations/channel/engine/router.go
+++ b/server/internal/integrations/channel/engine/router.go
@@ -448,10 +448,12 @@ func (r *Router) processClaimed(ctx context.Context, set ResolverSet, msg channe
 			res.IssueIdentifier = service.IssueIdentifier(prefix, duplicate.Number)
 			res.IssueDuplicate = true
 			// A duplicate is a terminal product outcome, not an infrastructure
-			// failure and not a chat prompt. Media binding remains independent,
-			// exactly like the successful direct-create path below.
+			// failure and not a chat prompt. Finalize the durable chat message's
+			// media state without resolving it. There is no new issue to consume
+			// the media, so downloading and persisting attachments would only
+			// create unused resources.
 			if resolveMedia {
-				r.enqueueMedia(set, inst, identity, appendRes.MessageID, msg, sessionID, duplicate, pgtype.Text{}, "", pgtype.UUID{}, localMediaDeadline)
+				r.enqueueMediaFinalization(set, inst, identity, appendRes.MessageID, msg, sessionID, localMediaDeadline)
 			}
 			return res, postAppendFinalize, nil
 		}
@@ -506,6 +508,18 @@ func (r *Router) processClaimed(ctx context.Context, set ResolverSet, msg channe
 // task service defers a task to the persisted media deadline, then media
 // completion promotes it early.
 func (r *Router) enqueueMedia(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, issue db.Issue, issueDescriptionBase pgtype.Text, issueCommandText string, issueTaskID pgtype.UUID, deadline time.Time) {
+	r.enqueueMediaJob(set, inst, identity, chatMessageID, msg, sessionID, issue, issueDescriptionBase, issueCommandText, issueTaskID, true, deadline)
+}
+
+// enqueueMediaFinalization clears the durable media-pending marker without
+// invoking the platform resolver. Duplicate /issue commands use this because
+// neither the existing issue nor the hidden command message should gain a new
+// copy of media that no newly-created issue will consume.
+func (r *Router) enqueueMediaFinalization(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, deadline time.Time) {
+	r.enqueueMediaJob(set, inst, identity, chatMessageID, msg, sessionID, db.Issue{}, pgtype.Text{}, "", pgtype.UUID{}, false, deadline)
+}
+
+func (r *Router) enqueueMediaJob(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, issue db.Issue, issueDescriptionBase pgtype.Text, issueCommandText string, issueTaskID pgtype.UUID, resolveRemote bool, deadline time.Time) {
 	key := keyForSession(sessionID)
 	done := make(chan struct{})
 
@@ -546,7 +560,7 @@ func (r *Router) enqueueMedia(set ResolverSet, inst ResolvedInstallation, identi
 				expired = true
 			}
 		}
-		if !expired {
+		if !expired && resolveRemote {
 			select {
 			case r.mediaSem <- struct{}{}:
 				defer func() { <-r.mediaSem }()
@@ -560,18 +574,20 @@ func (r *Router) enqueueMedia(set ResolverSet, inst ResolvedInstallation, identi
 				// sees the dead deadline and runs only the empty finalize.
 			}
 		}
-		r.resolveAndBindMedia(set, inst, identity, chatMessageID, msg, sessionID, issue, issueDescriptionBase, issueCommandText, issueTaskID, deadline)
+		r.resolveAndBindMedia(set, inst, identity, chatMessageID, msg, sessionID, issue, issueDescriptionBase, issueCommandText, issueTaskID, resolveRemote, deadline)
 	}()
 }
 
 const mediaFinalizeTimeout = 5 * time.Second
 
-func (r *Router) resolveAndBindMedia(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, issue db.Issue, issueDescriptionBase pgtype.Text, issueCommandText string, issueTaskID pgtype.UUID, deadline time.Time) {
+func (r *Router) resolveAndBindMedia(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, issue db.Issue, issueDescriptionBase pgtype.Text, issueCommandText string, issueTaskID pgtype.UUID, resolveRemote bool, deadline time.Time) {
 	ctx, cancel := context.WithDeadline(r.mediaCtx, deadline)
 	defer cancel()
 
 	resolved := msg
-	if ctx.Err() == nil {
+	if !resolveRemote {
+		resolved.MediaRefs = nil
+	} else if ctx.Err() == nil {
 		// Skipped entirely when the budget expired while queued (or on
 		// shutdown): resolving on a dead context would only churn through
 		// intent writes that immediately fail.
--- a/server/internal/integrations/channel/engine/router.go
+++ b/server/internal/integrations/channel/engine/router.go
@@ -516,7 +516,21 @@ func (r *Router) enqueueMedia(set ResolverSet, inst ResolvedInstallation, identi
 // neither the existing issue nor the hidden command message should gain a new
 // copy of media that no newly-created issue will consume.
 func (r *Router) enqueueMediaFinalization(set ResolverSet, inst ResolvedInstallation, identity ResolvedIdentity, chatMessageID pgtype.UUID, msg channel.InboundMessage, sessionID pgtype.UUID, deadline time.Time) {
-	r.enqueueMediaJob(set, inst, identity, chatMessageID, msg, sessionID, db.Issue{}, pgtype.Text{}, "", pgtype.UUID{}, false, deadline)
+	r.mediaQueueMu.Lock()
+	if r.stopping {
+		r.mediaQueueMu.Unlock()
+		return
+	}
+	r.mediaWg.Add(1)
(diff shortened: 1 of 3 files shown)


Bohan-J:
Thanks for this — the diagnosis is right and the regression test is a real one. I verified it: reverting `router.go` to `main` while keeping your test makes it fail with `duplicate media unexpectedly invoked resolver, calls=1`, and it passes on your branch (`go vet` clean, `go test -race ./internal/integrations/channel/engine -run TestRouter` ok). Repeating an identical image+command really should not re-download and re-persist the same picture.

Before this lands, though, I'd like to discuss one case I think gets caught in the blast radius.

## The duplicate verdict isn't evidence that the media is redundant

The change treats `ErrActiveDuplicate` as authoritative for "no new issue will consume this media." But look at what that verdict is actually based on (`server/pkg/db/queries/issue.sql:182`):

```sql
WHERE workspace_id = $1
  AND status NOT IN ('done', 'cancelled')
  AND project_id IS NOT DISTINCT FROM sqlc.arg('project_id')::uuid
  AND parent_issue_id IS NOT DISTINCT FROM sqlc.arg('parent_issue_id')::uuid
  AND lower(btrim(regexp_replace(title, '[[:space:]]+', ' ', 'g'))) = sqlc.arg('normalized_title')
```

It matches on **normalized title alone**. It does not look at the sender, the session, the channel, or the message body. And on the channel path specifically, `Router.createIssue` (`server/internal/integrations/channel/engine/router.go`) leaves `ProjectID`, `ParentIssueID` and `AllowDuplicate` all at their zero values, which compounds it:

- **NULL project** → every channel-created issue shares one flat, workspace-wide namespace (and collides with project-less issues created from the web UI too).
- **No time window** → and `backlog` counts as active, so once a title is used it is claimed until somebody moves that issue to `done`/`cancelled`. A title parked months ago still blocks today.
- **`AllowDuplicate` is always false** → there is no escape hatch on this path. Note that `issueguard.DuplicateMessage` tells the user *"Set allow_duplicate=true or use --allow-duplicate to create another."* — the CLI has that flag (`cmd_issue.go:485`) and the HTTP API has the field, but the `/issue` command parser has neither, so a DingTalk or Feishu user cannot act on that advice.

One more thing worth noting: genuine platform retransmissions are already dropped earlier, at the dedup claim keyed on `msg.MessageID` (`router.go:262`). So by the time the title guard fires, we are always looking at a *distinct user action*. The guard's job is to tell "the same person pressed send twice" apart from "somebody else reported a different problem that happens to share a title" — and title is the only signal it has, so it cannot.

## The concrete false positive

Channel titles are short, because people type them into a chat box:

- `/issue app won't open` — one person means the desktop app, another means a share link
- `/issue login page crash` — one on iOS, one on Windows
- `/issue error`, `/issue stuck`

Normalization also folds case and collapses whitespace, so `Login Page Crash` and `login  page   crash` are the same key.

So in a group chat:

1. Alice sends `[white-screen screenshot] /issue login page crash` → creates #100.
2. Twenty minutes later Bob hits a genuinely different bug, and sends `[console error screenshot] /issue login page crash`.
3. Bob's report is refused. That is the pre-existing problem, and it has nothing to do with media.
4. **Today:** Bob's console screenshot at least lands on #100, so the evidence survives somewhere.
5. **With this PR:** it is dropped entirely — not attached to the issue, and not visible in Chat either, since `channel_command` messages are filtered out of the public projections (`chat.sql:69,115`). The reply says only "active issue #100 already exists", with nothing to indicate an image was discarded.

That is my hesitation: for the identical-resend case the PR is clearly correct, but for the different-content case it removes the one accidental mitigation without addressing the refusal itself, and it does so silently.

## Suggested ordering

I think the guard is the thing to fix first, and there is already a precedent for it in the same package — the autopilot variant is time-scoped:

```go
func LockAndFindRecentAutopilotDuplicate(..., window time.Duration) (db.Issue, bool, error)
// created_after = now() - window
```

The generic path never got the same treatment. Giving the channel path a bounded window (minutes, not forever) would stop it from flagging an unrelated report that reuses a common title, while still catching what users actually complain about.

The nice consequence: once the guard only fires for "this person sent the same thing a moment ago", your rule here becomes unambiguously correct — a duplicate really would mean redundant media, and there is no evidence left to lose. I would rather land this change on top of that than ahead of it.

If you would prefer to land something in the meantime, the much smaller version is to leave the current behavior alone and make the reply honest — have `issueDuplicateText` mention that the attached image was not saved, so at least the loss is visible to the person who sent it.

## Minor notes on the diff itself

- `resolveAndBindMedia`: when `resolveRemote` is false and the job's deadline expires while it waits in the session queue, `ctx.Err() != nil` still logs `channel router: media resolution incomplete; using placeholder`. Nothing was going to be resolved on that path, so that WARN points at a media failure that never happened. `if err := ctx.Err(); err != nil && resolveRemote {` would cover it — `resolved.MediaRefs = nil` is already handled in the branch above.
- The finalize-only job still queues behind the session's other media for up to the full budget, just to do one `UPDATE` on its own `chat_message` row. Since a `channel_command` turn does not gate task batching (`chat.sql:489`), that marker isn't holding anything back — it could finalize immediately.
- `TestBindMediaRefs_DuplicateIssueFinalizationCreatesNoAttachment` doesn't exercise the duplicate path; it calls `BindMediaRefs` with empty refs directly. Valuable as-is, but something like `TestBindMediaRefs_EmptyRefsCreateNoAttachmentAndClearPending` would describe it more accurately.
- The PR description mentions the two DingTalk/Feishu suites being blocked by the `265`→`271` migration-ledger drift. That looks stale now — this branch's base already includes the renumber from #6683, and the backend CI job is green.

Keen to hear your take on the duplicate-guard question in particular — if you disagree that the different-content case matters in practice, I'm happy to be talked out of it.

yyclaw:
I looked through the duplicate-guard concern separately from the attachment defect.

Rejecting an active issue with the same normalized title is existing product behavior. This PR intentionally keeps that contract: once the authoritative issue service refuses creation, the channel path must not attach the new command's media to an existing issue that the command did not create. The previous attachment was an accidental side effect rather than a supported fallback for the duplicate-title policy.

If we revisit the product rule, I suggest doing that separately. My preference would be to add sender identity to the duplicate key, not a time window: reports with the same title from different senders could remain distinct, while uniqueness would not silently expire with time. That deserves its own product/design discussion and test matrix.

I addressed the four diff-specific notes in `32314f4a1`:

- The timeout/cancellation warning now only applies when remote media resolution was actually requested.
- Finalize-only work now bypasses the same-session remote-media queue and clears its own marker immediately. It remains detached and drain-tracked; task promotion still checks the persisted non-`channel_command` pending markers.
- Renamed the DB test to `TestBindMediaRefs_EmptyRefsCreateNoAttachmentAndClearPending` so its name matches its actual scope.
- Removed the stale migration-ledger caveat from the PR description and updated the change/test summary.

I also added deterministic regression coverage for the queue interleaving and false-warning branch. Verification completed:

```text
go test -race ./internal/integrations/channel/engine -count=1
go vet ./internal/integrations/channel/engine
go test ./internal/service -run '^(TestEnqueueChatTaskDefersForChannelMediaAndPromotesWhenReady|TestPromoteChannelChatTasksIgnoresHandledCommandMedia|TestDeferredChannelIssueTaskPromotesAfterMediaSettlement|TestChannelMediaReconciler_SettleInvariantDwarfsPipelineBudgets)$' -count=1
```

The post-feedback preflight result is `GO`.

Bohan-J:
Thanks — and thanks for taking the four notes seriously rather than just answering them. I pulled `32314f4a1` and checked each one:

- `router.go:612` — the warning is now gated on `resolveRemote`, so the finalize-only path can no longer report a media failure that was never attempted.
- `router.go:518-533` — finalize-only work skips the ordered per-session queue entirely, while `stopping` and `mediaWg.Add(1)` stay inside the same lock, so `Drain` still tracks and waits for it. And because `BindMedia` runs on its own `context.Background()` budget, the marker still clears during shutdown.
- The DB test rename and the stale migration caveat are both done.
- The two new tests cover exactly the branches that were previously unguarded.

Verified locally on the PR head: `go vet` clean, `go test -race ./internal/integrations/channel/engine -count=1` ok, and the four `internal/service` tests ok. CI is green across all nine jobs.

On the duplicate-guard question — your framing is the right one, and you own this area, so we're going with it. You're correct that the media landing on the existing issue was an accidental side effect rather than a designed fallback, and that a channel command has no business mutating an issue it did not create. Keeping that contract intact here is the right call, and the guard itself is a separate product decision.

Your counter-proposal is also better than the time window I suggested. Sender identity in the duplicate key handles both halves cleanly — a resend from the same person is still caught, while two people reusing a common title stay distinct — whereas a window has to trade one against the other. Agreed it needs its own design pass and test matrix.

Two follow-ups we'd like to see, neither blocking this PR:

1. The duplicate key change you proposed.
2. A line in `issueDuplicateText` telling the sender their image was not saved. Small, but it is the only signal that something was discarded, and it's worth having whichever way the guard lands.

One leftover nit, not worth another round: `enqueueMediaJob` now has a single caller that always passes `resolveRemote: true`, since the finalize path no longer goes through it. That makes the parameter and the `!expired && resolveRemote` condition unreachable in their false state — worth collapsing next time you're in this file. (`resolveAndBindMedia` still needs its copy of the flag.)

Merging. Thanks for the clean turnaround.

yyclaw:
@Bohan-J One correction to my earlier reply: adding sender identity to the duplicate key was an early idea, not a settled conclusion, and it may not be the best solution.

After looking at the problem more broadly, sender identity is not authoritative in a collaborative workspace: different people can report the same underlying issue, while the same person can submit distinct reports. Normalized-title equality is also not enough. A future design likely needs to evaluate the title and description semantically, with sender, time, project, and channel treated as supporting signals rather than standalone keys.

I captured the problem, current constraints, and directions worth evaluating in #6743. The issue intentionally does not make a final product or implementation decision.

PR #6690, opened 2026-08-10
MUL-5973: fix(tabs): restore inbox scroll position and stop the highlight re-jump on tab return

NevilleQingNY:
## Problem

Returning to the inbox tab lost the reading position twice over (desktop tabs; web back-navigation shared the same gaps):

1. **The notification list reset to the top.** It was the one list surface that never joined the MUL-4741 scroll memento protocol — no `data-tab-scroll-root`, no restore pull — so the capture pass had nothing to save and the remount started from row 0.
2. **The issue detail jumped back to the highlighted comment.** The comment deep-link landing was guarded only by a component-internal ref, which dies with the unmount; every tab return replayed the jump over the offset the memento had just restored.

Field testing on this branch surfaced two more defects in the same machinery:

3. **Recording the landing killed the landing.** Writing the consumed marker inside the landing effect re-fired that effect one commit later (the marker was in its dependency array), and the effect **cleanup cancelled the centering rAF loop and the highlight fade before a single frame ran** — the jump's code path executed while the screen never moved.
4. **Two scroll-restore systems raced, and the stale one won.** On issue-detail tabs, the tab memento (captured from the live DOM at leave time) and `useIssueDetailScrollRestore`'s module-level map both restored scrollTop; the map's retry loop ran last and re-asserted an *older* offset (the map only hears scroll events — scroll anchoring and streaming activity blocks collapsing move scrollTop without one). Log-verified: memento said 3428.5, position settled at the map's stale 4017.5.

## Required behavior

| Action | Expected |
| --- | --- |
| First click on a comment notification | Detail scrolls to the comment and flashes the highlight |
| Switch tab away and back | List and detail restore exactly where they were left; **no** re-jump to the comment |
| Click the notification again (open or not) | The comment jump replays |
| Select notification B after A | B's detail lands fresh (own highlight jump); A's offset never leaks into B |
| Issue-detail tabs (no highlight) | Toggling tabs restores the leave-time position, even while agents stream content into the timeline |

## What & why

**InboxList joins the scroll protocol** — the same wiring the board/list views ship: scroller marked `data-tab-scroll-root="list"`, saved offset assigned at ref-attach and fed into Virtuoso's `initialScrollTop`, plus the seed spacer / `defaultItemHeight` estimate (58px rows) so the container height is truthful on the first frame and the restored offset sticks.

**TabMemento gains generic `view` entries.** The memento was specced as "restorable view state" (MUL-4741) but only implemented scroll, so one-shot-consumed state had nowhere to survive a remount. The restoration adapter grows optional `getViewState`/`setViewState`; desktop commits into the tab memento (new `commitViewState` action, no-op-write skip, entries capped at 100 per tab with oldest-first eviction), web mirrors the semantics on its module map. Scroll's REPLACE-per-route semantics never touch view entries; legacy persisted payloads normalize on rehydration.

**IssueDetail records the highlight landing and scopes its scroll key per issue.**
- Landing writes `highlight:<issueId> = commentId` into the memento; a remount that finds the entry skips the jump and lets the restored offset stand. The landing effect reads the marker **through a ref, not its dependency array** — as a dependency, the write re-fired the effect and its cleanup cancelled the animation it had just scheduled (defect 3).
- The scroll container key becomes `main:<issueId>`: in the inbox the selection lives in the query string while memento entries key by pathname, so a bare `main` restored notification A's offset into notification B's detail.
- `InboxPage.handleSelect` clears the entry on every row click (fresh deep-link intent). Re-clicking the already-open row additionally bumps a `highlightRequestToken` threaded to `IssueDetail` — it re-arms the landing guard and forces the render that re-reads the cleared entry, replaying the jump without a remount. Selection changes don't need the token: their remount replays the landing by itself.

**One restore target for all three scroll writers** (defect 4): `useIssueDetailScrollRestore` accepts `overrideTop` — when the platform serves a memento offset, the retry loop targets it instead of the module-level map, so the attach assignment, Virtuoso's `initialScrollTop`, and the retry loop all converge and write order stops mattering. The hook is disabled only while a comment deep link still has a landing to run; on a consumed return its retry loop IS the restore (the one-shot attach assignment clamps against a flat timeline whose markdown/images haven't reached their captured height yet).

## Tests

- `tab-store.test.ts`: `commitViewState` write/clear/no-op-skip/eviction-at-100, view entries surviving a scroll REPLACE and vice versa, rehydration normalizing pre-`view` payloads.
- `use-issue-detail-scroll-restore.test.tsx`: memento override wins over the module map.
- `inbox-page.test.tsx`: re-clicking the open row bumps the replay token; selection changes don't.

## Verification

- `pnpm typecheck` ✅ (6/6)
- `pnpm lint` ✅ (0 errors; warnings pre-existing, none in touched files)
- `pnpm test --filter @multica/desktop --filter @multica/views` ✅ (462 desktop + 3,770 views tests)
- Manually verified on desktop dev with debug probes: restore settles exactly at the captured offset (drift 0 across 60 frames), tab return does not re-jump, re-click replays the jump.

## Known limits

- A selection driven purely by URL (OS-notification click into an already-consumed issue/comment pair) doesn't clear the consumed entry, so that path won't re-jump.
- Old persisted `main` scroll entries are orphaned by the per-issue key; they self-clean on the next capture of the route (REPLACE).
- On live-streaming issues the captured offset is a faithful pixel snapshot of the moment you left; content that keeps growing above the viewport after restore can still shift the reading position — inherent to pixel-offset restoration.

### DIFF
--- a/apps/desktop/src/renderer/src/hooks/use-navigation-input-bindings.test.tsx
+++ b/apps/desktop/src/renderer/src/hooks/use-navigation-input-bindings.test.tsx
@@ -26,7 +26,7 @@ function seedHistory() {
               stack: ["/acme/issues", "/acme/issues/abc"],
               index: 1,
             },
-            memento: { scroll: {} },
+            memento: { scroll: {}, view: {} },
           },
         ],
         activeTabId: "t1",
--- a/apps/desktop/src/renderer/src/platform/tab-coordinator.ts
+++ b/apps/desktop/src/renderer/src/platform/tab-coordinator.ts
@@ -70,17 +70,37 @@ export function registerCoordinatorQueryClient(qc: QueryClient): void {
  * see ScrollRestorationProvider in @multica/views/platform). Offsets are
  * looked up live against the tab's memento, scoped to the route the view is
  * mounting under, so in-tab back/forward pulls each route's own offsets.
+ *
+ * The generic view-state entries ride the same channel: reads resolve
+ * against the memento's `view` map, writes commit through the store — no
+ * capture pass, the view pushes at the moment its restorable state changes.
  */
 export function createScrollRestorationAdapter(
   tabId: string,
 ): ScrollRestorationAdapter {
+  const activeRouteKey = (): string | null => {
+    const state = useTabStore.getState();
+    const active = getActiveTab(state);
+    if (!active || active.id !== tabId) return null;
+    return splitTabUrl(active.url).pathname;
+  };
   return {
     get(containerKey) {
-      const state = useTabStore.getState();
-      const active = getActiveTab(state);
-      if (!active || active.id !== tabId) return undefined;
-      const routeKey = splitTabUrl(active.url).pathname;
-      return active.memento.scroll[scrollMementoKey(routeKey, containerKey)];
+      const routeKey = activeRouteKey();
+      if (routeKey === null) return undefined;
+      const active = getActiveTab(useTabStore.getState());
+      return active?.memento.scroll[scrollMementoKey(routeKey, containerKey)];
+    },
+    getViewState(entryKey) {
+      const routeKey = activeRouteKey();
+      if (routeKey === null) return undefined;
+      const active = getActiveTab(useTabStore.getState());
+      return active?.memento.view[scrollMementoKey(routeKey, entryKey)];
+    },
+    setViewState(entryKey, value) {
+      const routeKey = activeRouteKey();
+      if (routeKey === null) return;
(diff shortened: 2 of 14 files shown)


NevilleQingNY:
Follow-up pushed (bd53a7c): re-clicking the already-open notification row was a dead click — no remount, landing guard still armed — so after the (intended) removal of the tab-return re-jump there was no way back to the highlighted comment. Every row click now clears the consumed-landing entry, and a same-key re-click bumps a `highlightRequestToken` prop that re-arms the landing guard on the mounted detail, replaying the jump in place. Covered by a new InboxPage test.

PR #6696, opened 2026-08-10
MUL-5979 fix(views): hide empty agent detail menu

### LINKED ISSUE #6695: [Bug]: Agent detail page shows empty kebab menu when no actions are available
## Summary
On the agent detail page, the top-right kebab menu can still render even when there are no available menu actions. Clicking it opens a collapsed, empty menu surface instead of either showing a valid action list or hiding the affordance entirely.

## Original screenshot
Original bug screenshot

## Repro
1. Open an agent detail page for an agent that cannot be archived from this surface, such as a built-in/system agent.
2. Click the top-right `...` action button.
3. Observe that the popup opens as an empty narrow strip.

## Expected
If there are no available menu items, the kebab trigger should not render.

## Actual
The trigger still renders and opens an empty/collapsed menu popup.

## Likely scope
`packages/views/agents/components/agent-detail-page.tsx`

HomyeeKing:
## Summary
- hide the agent detail kebab trigger when there are no available menu actions
- add a regression test for system agents that previously showed an empty menu shell
- follow up on GitHub bug #6695

## Original screenshot
Original bug screenshot

## Testing
- pnpm -C packages/views exec vitest run agents/components/agent-detail-page.test.tsx
- pnpm -C packages/views typecheck

Closes #6695

### DIFF
--- a/packages/views/agents/components/agent-detail-page.test.tsx
+++ b/packages/views/agents/components/agent-detail-page.test.tsx
@@ -214,6 +214,39 @@ describe("AgentDetailPage DM button", () => {
     expect(screen.queryByRole("button", { name: "DM" })).not.toBeInTheDocument();
   });
 
+  it("hides the more-actions trigger when no menu actions are available", async () => {
+    // The gate must survive a caller who genuinely CAN archive: an admin
+    // passes `canEditAgent`, so `canArchive` is true. The only thing keeping
+    // the menu empty is the system agent's undefined `onArchive`. Assert the
+    // real aria label (locale `detail.more_actions_aria` = "Agent actions"),
+    // so removing the `hasMoreActions` gate would render the empty shell and
+    // fail this test.
+    agentsRef.current = [
+      { ...baseAgent, system_key: "mika" },
+    ];
+    membersRef.current = [{ user_id: "user-1", role: "admin" }];
+    renderPage();
+
+    await screen.findByRole("button", { name: "Assign work" });
+    expect(
+      screen.queryByLabelText("Agent actions"),
+    ).not.toBeInTheDocument();
+  });
+
+  it("keeps the more-actions trigger for an editable non-system agent", async () => {
+    // Positive counterpart: an owner of a normal agent has a real archive
+    // action, so the menu trigger must still render. Guards the gate against
+    // over-hiding.
+    agentsRef.current = [{ ...baseAgent, owner_id: "user-1" }];
+    membersRef.current = [{ user_id: "user-1", role: "member" }];
+    renderPage();
+
+    await screen.findByRole("button", { name: "Assign work" });
+    expect(
+      screen.getByLabelText("Agent actions"),
+    ).toBeInTheDocument();
+  });
+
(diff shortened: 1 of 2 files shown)


ADanMan (commented):
Thanks for tracking down #6695 — the production change is right: `onArchive` is `undefined` for `system_key` agents, so gating the trigger on `hasMoreActions = !!onArchive` correctly hides the empty kebab.

The regression test, though, doesn't actually guard against the bug — it passes whether or not the production fix is present, for two independent reasons:

1. Wrong aria-label. The test queries `queryByLabelText("More actions")`, but the trigger's label comes from `detail.more_actions_aria`, which is `"Agent actions"` in `packages/views/locales/en/agents.json`. Since the test renders through a real `I18nProvider` with the `en` resources, the label resolves to "Agent actions" and the query matches nothing regardless of the render guard.

2. The fixture can't archive anyway. The test only overrides the agent (`system_key: "mika"`); it keeps the `beforeEach` member fixture `{ user_id: "user-1", role: "member" }` while `baseAgent.owner_id` is `"user-2"`. `useAgentPermissions` runs for real here, and `canEditAgent` denies a non-admin non-owner, so `canArchive` is `false`. The pre-fix guard `!isArchived && canArchive` was already false for this fixture, so the empty shell never rendered even without the change.

Net effect: reverting the production change leaves this test green, so it isn't protecting the fix. To make it a real regression test, please (a) set up a member who actually passes `canEditAgent` — an `admin`/`owner` role, or make user-1 the owner — so `canArchive` is true, and (b) assert on the real label, e.g. `queryByLabelText("Agent actions")` (or reference the locale key). With both, the assertion will fail if the `hasMoreActions` guard is removed. It's also worth a paired positive case (an owner-editable non-system agent) confirming the trigger still appears, so the guard can't silently over-hide.

Bohan-J:
Thanks for this, @HomyeeKing — clean report and a clean fix, and thanks for turning the regression test around after the review feedback.

The root cause is worth naming for the record: `canArchive` (`:311`, permission only) and `onArchive` (`:317`, also excludes built-ins) disagreed, and the trigger rendered on the first while its only item rendered on the second. Gating on `hasMoreActions` lines the two back up.

I verified before merging:

- Your test file run against the fix: 7 passed.
- Mutation check — kept your tests, reverted the production change to `main`'s version: the system-agent case fails. The test genuinely guards the fix now.
- `packages/views` typecheck and ESLint on both changed files: clean.
- All 9 CI checks green on `345cf6e`.

I also checked the sibling menu in `agent-row-actions.tsx` — it renders items conditionally too, but carries an unconditional "open in new tab" item, so it can't collapse to an empty shell. Your scoping was right.

One non-blocking note for anyone extending this later: if the menu ever gains a second action, the `hasMoreActions` gate needs to aggregate them rather than tracking `onArchive` alone. The name already points the right way. A tighter variant would be to make `canArchive` itself account for system agents at `:311` (`canEdit.allowed && !agent.system_key`), matching how `agent-row-actions.tsx:96` already derives `showArchive` — not needed for this merge.

I've added `Closes #6695` to the description and the internal tracking ref to the title. Merging now.

PR #6701, opened 2026-08-10
MUL-5980: drag blank board area with the left button to pan horizontally (#6700)

### LINKED ISSUE #6700: [Feature]: Issue board — drag the blank area with the left button to pan horizontally
## Summary

On the issue board (kanban), allow horizontal panning by **dragging the empty board background with the left mouse button** — the page follows the cursor. This is the Trello/Linear "drag the background to move the board" pattern, and it makes navigating wide boards with many columns faster than reaching for the scrollbar or shift-wheel.

## Behavior

- Drag a **card** → moves the card (existing dnd-kit sorting), unchanged.
- Drag **blank board background** → pans the board horizontally.
- Left/right axis only. Vertical pointer movement is ignored; `scrollTop` is never touched.
- No mode toggle and no keyboard shortcut required. The right button and the native context menu are not involved, so a normal right-click keeps its menu.

## Notes

Applies to the issue board view (`packages/views/issues/components/board-view.tsx`, the `overflow-x-auto` scroll container). A gesture activates only when it starts off any card / interactive element, giving natural mutual exclusion with card drag. Implemented with Pointer Events + pointer capture and a ~5px activation threshold matching the board's `PointerSensor`.

HomyeeKing:
Closes #6700.

## What

On the issue board, dragging the **empty board background with the left mouse button** now pans the board horizontally — the page follows the cursor (Trello/Linear "drag the background" pattern).

- Drag a **card** → moves the card (dnd-kit sorting), unchanged.
- Drag **blank background** → pans the board left/right.
- The right button and the native context menu are no longer involved at all, so the earlier `mousedown → contextmenu → mousemove` ordering problem is gone.

## How

- New `useBoardDragPan` hook (`packages/views/issues/components/use-board-drag-pan.ts`) attached to the board's `overflow-x-auto` scroll container in `board-view.tsx`.
- **Pointer Events + pointer capture**, so the drag keeps tracking when the cursor leaves the element/window.
- Left button only (`event.button === 0`). Gestures that start on a card (`[data-board-card]`) or an interactive element (`a` / `button` / `input` / `textarea` / `select` / `[role=button]` / `contenteditable`) do not activate — natural mutual exclusion with dnd-kit card drag.
- Reuses the board's ~5px activation threshold (`PointerSensor` `activationConstraint: { distance: 5 }`) to tell a click from a drag.
- Horizontal axis only: `deltaY` is never read, `scrollTop` is never written.
- Cleanup on `pointerup` / `pointercancel` / `lostpointercapture` and window `blur`, plus a `event.buttons` check on move, so a lost release can't leave the board stuck in a panning state.
- `Space`-hold is intentionally out of scope for this PR (could be added later as an optional second entry point with input-field guarding).

## Tests

New `use-board-drag-pan.test.tsx`, driven in real browser pointer-event order:
- pans on a left drag from blank background;
- does not activate before the ~5px threshold (plain click);
- ignores drags that start on a card (sorting untouched);
- ignores the right button entirely;
- does not stay stuck when the release event is lost (`buttons === 0` on the next move);
- reads the horizontal axis only (`scrollTop` untouched).

`tsc --noEmit` and `eslint` on the changed files are clean.

https://github.com/user-attachments/assets/f0185f8d-88b0-4339-aa93-837e8a552a12

https://github.com/user-attachments/assets/4301ed75-aab5-4ac5-be93-a4c675a675d5


Bohan-J (changes requested):
Thanks for taking this on — panning a wide board without reaching for the scrollbar is a real ergonomic win, and the change is nicely scoped (horizontal axis only, left-button card drag untouched, sensible file placement).

That said, the core gesture doesn't survive contact with real browsers yet. Two blocking issues:

### 1. `contextmenu` fires before the first `mousemove`, so the suppression never runs

`packages/views/issues/components/use-right-button-pan.ts:66` only calls `preventDefault()` when `movedRef.current` is already `true`. But the real event order for a right-button press is:

```
mousedown → contextmenu → mousemove → mouseup
```

I verified this in Chromium 145, Firefox 146 and WebKit 26 — in all three, `contextmenu` is dispatched on button-down, before any movement. WebKit additionally stops dispatching `mousemove` once the menu is open. So at the moment the handler runs, `movedRef.current` is still `false`:

- Over empty board space → the native browser menu opens immediately, then the board pans underneath it.
- Over a card → `packages/views/issues/actions/issue-actions-context-menu.tsx:200` opens the issue actions menu first; the parent handler at `packages/views/issues/components/board-view.tsx:566` bubbles too late to undo a menu the child already opened.

The new tests pass because `use-right-button-pan.test.tsx` synthesizes `mousedown → mousemove → mouseup → contextmenu`, which is the reverse of what browsers do, so the real user path is never exercised.

Fixing this needs a product decision first, and it'd be good to settle it before more code:

- **(a)** Multica takes over the right button on the board: `preventDefault()` the `contextmenu` unconditionally while over the scroller, and provide our own menu on a stationary press-and-release. Consistent panning, but we own the menu.
- **(b)** Move the gesture to a binding that doesn't compete with the OS menu — middle-button drag, or space/modifier + left drag. Cheaper and no menu conflict.

Either way, please add a regression test that fires events in the real browser order and starts the drag **on a card**, not just on empty space.

### 2. Sticky pan state when the `mouseup` is lost

`use-right-button-pan.ts:38` gates only on the internal `panningRef` flag and never checks `MouseEvent.buttons`. If the release happens outside the window, the window loses focus, or the browser cancels the gesture, no `mouseup` reaches the handler and `panningRef` stays `true` — after that, ordinary pointer movement with no button held keeps mutating `scrollLeft`.

The current tests actually lock this in: the `mousemove` events they construct default to `buttons === 0`, and the assertions still expect panning to happen.

Suggested fix: switch to Pointer Events with `setPointerCapture` / `pointercancel`, or at minimum check the right-button bit (`event.buttons & 2`) in the move handler and clear state on `blur` / `pointercancel`. Please add a test that drops the release event.

### Minor

When you rework the gesture, reuse the project's existing ~5px drag threshold rather than the `deltaX !== 0` check — a 1px tremor, or movement at the scroll boundary where nothing actually scrolls, shouldn't swallow the context menu.

### What I verified

- New unit tests: 4/4 pass
- `@multica/views` typecheck: pass
- ESLint on changed files: pass
- `git diff --check`: pass
- CI: green
- Browser event-order check: issue #1 reproduces in Chromium, Firefox and WebKit

Happy to help pick between (a) and (b) — once the gesture is settled, the rest of this is close.

HomyeeKing:
Thanks @Bohan-J for the thorough review, and I agree the current gesture fights the OS context menu. Rather than taking over the right button globally, I'd like to propose a different design that avoids the conflict entirely and matches the "hand/pan tool" pattern people already know from design tools:

**1. Explicit pan mode via a button.** Add a small pan/hand button at the bottom-right of the board. Clicking it enters pan mode and the cursor switches to a grab/hand cursor; while in this mode, dragging pans the board.

**2. Spacebar as the quick modifier.** Holding `Space` temporarily switches into the same pan (hand) cursor, exactly like Figma/Photoshop. While `Space` is held, moving the mouse with the button pressed pans the board to follow the cursor.

**3. `contextmenu` is only disabled during the pan gesture.** Suppression is scoped to when we're actually panning, so a normal right-click (no pan in progress) keeps the native/issue-actions menu intact. This sidesteps the `mousedown → contextmenu → mousemove` ordering problem, because entering pan mode is an explicit, prior state rather than something inferred from movement after the menu already fired.

This also naturally addresses the "sticky pan state" issue: pan mode has an explicit enter/exit, and the `Space`-hold variant ends on keyup / blur, so a lost `mouseup` can't leave us stuck. I'll rework the gesture around Pointer Events with capture and add regression tests that fire events in real browser order (including starting on a card and dropping the release event).

Does this direction work for you? If so I'll push the revised implementation.

HomyeeKing:
@Bohan-J  What do you think about this thought?

Bohan-J:
Thanks for rethinking this — agreed on dropping the right-button gesture. On the two ideas you proposed:

- **Persistent hand-mode button: please leave it out of this PR.** It introduces a persistent mode, a permanent piece of UI, and enter/exit semantics (accessibility states, Escape, what happens on view switch) — a lot of surface area for a low-frequency navigation need.
- **`Space`-hold: a known convention, but not the best fit here.** The board isn't a canvas, so users won't think to press `Space`; `Space` also has native scroll semantics in browsers, and you'd need to guard every focused input/textarea/contenteditable.

**What we'd like instead: left-drag on empty board space to pan** — the Trello/Linear pattern. Drag a card to move the card, drag the blank space to move the board. No mode, no key, self-discoverable, and it never touches `contextmenu`, which makes the whole event-ordering problem disappear.

We checked the board code and this path is unclaimed: the scroll container (`packages/views/issues/components/board-view.tsx`) is a plain `overflow-x-auto` div with no existing background gesture (no marquee selection, no click action), so blank-space left-drag conflicts with nothing.

Implementation notes:

- Use Pointer Events with pointer capture, as you already planned.
- Don't activate when the gesture starts on a card or any interactive element (button, link, input) — that keeps pan and dnd-kit card dragging naturally mutually exclusive.
- Reuse the ~5px activation threshold to disambiguate click from drag — the board's `PointerSensor` already uses `activationConstraint: { distance: 5 }`, so the two gestures stay consistent.
- End the gesture on `pointerup`, `pointercancel`, `lostpointercapture`, and window `blur`, and check `event.buttons` during move so a lost release can't leave the board stuck in pan.

Regression tests should cover: events fired in real browser order, a drag starting on a card neither pans nor breaks card sorting, a lost release event, and a stationary right-click still showing the existing menus.

If you'd like to keep `Space`-hold as an optional second entry point on top of that, that's fine (with the input-field guard), but it's not required for acceptance. Please also update the PR and issue title/body, since they still describe the right-button design.

Looking forward to the revised implementation!

HomyeeKing:
@Bohan-J Please help review again and the description has the demo show

multica-eve (changes requested):
Thanks for reworking this around blank-area left-dragging. The two blockers from the previous review are resolved: the gesture no longer competes with the context menu, and pointer capture plus the cleanup paths prevent a stuck pan state.

There is one remaining blocker before approval:

### Scope the gesture to mouse input

`packages/views/issues/components/use-board-drag-pan.ts:122-152` currently checks only `event.button === 0`. As the adjacent comment notes, that also matches primary touch and pen input, so those pointers are captured and `preventDefault()` is called as soon as they start on blank board space. The returned `touchAction: "pan-y"` at lines 218-221 also disables native horizontal touch gestures and does not preserve `pinch-zoom`.

This PR describes a left-*mouse*-button interaction, so the smallest safe fix is to:

1. Start the gesture only when `event.pointerType === "mouse" && event.button === 0`.
2. Remove `touchAction: "pan-y"` and the corresponding `style` plumbing.
3. Add a regression test showing touch/pen pointerdown does not capture the pointer, call `preventDefault()`, suppress selection, or pan the board.

If touch/pen panning is intended as an additional feature, it needs an explicit interaction design plus coverage for multi-touch, pinch zoom, vertical scrolling, and cancellation; that is larger than this PR needs to be.

Non-blocking follow-up: please manually verify that dragging a column's native vertical scrollbar still works on Windows or another environment with always-visible classic scrollbars. The parent board captures non-interactive pointerdowns immediately, and this could not be conclusively exercised in a macOS overlay-scrollbar environment.

Once the mouse-only scope is fixed, this should be ready for a short final review.

HomyeeKing:
@multica-eve I have addressed the mouse-only scope review on the latest head `a335759b6`:

1. Board pan now starts only for `event.pointerType === "mouse" && event.button === 0`.
2. Removed `touchAction: "pan-y"` and the corresponding `style` plumbing from the board container.
3. Added regression coverage showing touch/pen pointerdown does not capture the pointer, does not call `preventDefault()`, does not suppress `user-select`, and does not pan the board.

The focused pan tests are passing (15/15), `@multica/views` typecheck passes, and eslint on the changed files passes. Please take another look when you have time.


CC @Bohan-J

multica-eve (approved):
Thanks for the thoughtful iteration on this contribution. You responded carefully to each review round: moving away from the conflicting right-button gesture, hardening pointer capture and cleanup, preserving card drag and context-menu behavior, and finally restoring browser-owned touch, pen, and pinch-zoom interactions.

I re-reviewed the latest commit (`a335759b6`). The mouse-only scope and regression coverage address the remaining blocker, the focused tests/typecheck/lint pass, CI is green, and the PR merges cleanly with current `main`.

Approved — thank you for sticking with the feedback and landing a much safer, more focused interaction.

PR #6706, opened 2026-08-10
MUL-5986 fix(wecom): pin a 0.5%-flaky decrypt test, and four comments that outran the code

seacen:
Follow-up to #6605, which merged before these landed on it. One of them is a
flaky test that is now on main.

## The flake

`TestStreamingDecryptRefusesAWrongKey` drew a fresh random key each run.

PKCS#7 cannot always tell a wrong key from the right one: the bytes a wrong key
produces are garbage, and about one time in 256 that garbage ends in `0x01` — a
pad of one byte, which validates against itself and leaves the decrypt reporting
success. There is no MAC on this path; the platform does not send one.

So the test goes red roughly 0.5% of the time. Measured here:

```
random keys, -count=3000  → 15 failures
fixed keys,  -count=300   → 0
```

Both keys are fixed now. The 1-in-256 collision is a limit of the format rather
than something `unpadMedia` can be asked to catch, and the comment says so
instead of leaving the next reader to rediscover it.

## Four comments that claimed more than the code

- `trace.go` said in two places that one string skips the redactor. There are
  two: the `Content-Disposition` and the filename read out of it.
- The trace cap's rationale gave 255 bytes as the filename limit. That is POSIX;
  NTFS counts 255 UTF-16 code units, so a 255-character CJK name exceeds the
  cap. Nobody has that name — but the reasoning should say which limit it means.
- `cleanMediaFilename` said exactly one control character used to survive the
  transport. `U+0085` is a second: a control character to `unicode.IsControl`,
  two ordinary bytes to `net/http`.
- `stripControlRunes` said runes that decoded to invalid UTF-8 are left as they
  are. `strings.Map` substitutes `U+FFFD` for them. That is the better answer —
  Postgres will not take the raw bytes either — so the comment was what was
  wrong, not the code.

## One operator-facing correction

`SELF_HOSTING_ADVANCED.md` told operators that tokens are redacted out of the
trace, without noting that the media header line is written verbatim and past
the redactor. Somebody deciding whether to turn tracing on needs that in front
of them.

### DIFF
--- a/SELF_HOSTING_ADVANCED.md
+++ b/SELF_HOSTING_ADVANCED.md
@@ -172,7 +172,7 @@ Turn it on for a debugging session and unset it when the session ends. Before
 you do:
 
 - **It is read at boot, so changing it needs a backend restart** — `docker compose -f docker-compose.selfhost.yml up -d backend`. There is no runtime toggle. While it is on, the backend logs a warning on every startup saying so.
-- **Anyone who can read the backend's logs can read the traced message text.** That is a wider audience than the WeCom chat it came from: your `docker logs` / journald / log shipper, and whoever administers them. Binding tokens and other `token=…` parameters are redacted and the smart-bot secret is never read, but user message content is not.
+- **Anyone who can read the backend's logs can read the traced message text.** That is a wider audience than the WeCom chat it came from: your `docker logs` / journald / log shipper, and whoever administers them. The smart-bot secret is never read, and `token=` / `access_token=` / `code=` parameters are redacted out of message text — but user message content is not, and **an attachment's `Content-Disposition` and the filename read out of it are written verbatim**, up to 2048 runes and past the redactor. That line exists to show exactly how the storage backend encoded a name, and a redacted or truncated copy of it answers nothing; it also means an attachment's filename reaches your logs as sent.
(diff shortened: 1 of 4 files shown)


ADanMan (commented):
Approve. I checked each claim against origin/main and they hold up.

The core fix is sound: I reproduced the seal/CBC-decrypt/unpad path, and the fixed wrong-key pair (0x7e x32, IV = first 16 key bytes) decrypts "the real contents" to a trailing pad byte of 208 — out of the 1..32 range — so `decryptToFile` errors deterministically and the `err == nil` assertion always holds; the right key validates with pad=15. The flake rationale (PKCS#7, no MAC, ~1/256 self-validating pad) is accurate, and pinning both keys is the right call.

The comment corrections are all verified accurate: `traceMediaHeaders` really does emit two unredacted-but-bounded strings (`content_disposition` and `filename`, both via `traceBound`, which caps/flattens but does not redact), so "two strings skip the redactor" is right; and `strings.Map` does substitute U+FFFD for invalid UTF-8 rather than passing the raw bytes through, so the old `stripControlRunes` comment was the thing that was wrong. The SELF_HOSTING note that the media header line lands in logs verbatim past the redactor is correct and worth stating.

One trivial nit: the PR body says 15/3000 (~0.5%) while the committed comment says 20/5000 (~0.4%) — harmless, but worth aligning. Not blocking.

multica-eve (approved):
Thanks for the careful follow-up and for documenting both the PKCS#7 limitation and the trace-logging behavior so clearly. I verified that the fixed-key test is stable and that the operator-facing correction matches the trace path. Approved — we appreciate the contribution.

PR #6707, opened 2026-08-10
fix(ui): hide kebab menu on agent detail page when no actions are available

### LINKED ISSUE #6695: [Bug]: Agent detail page shows empty kebab menu when no actions are available

Kickflip73:
## Summary
On the agent detail page, the kebab menu (more-actions dropdown) renders
even when there are no available menu items. This happens for built-in
(`system_key`) agents where `onArchive` is undefined but `canArchive`
(derived from `canEdit`) is still true.

## Fix
Add `onArchive` to the render guard on line 542 so the DropdownMenu
only renders when there is at least one actionable item inside it.

## Testing
- Existing tests pass (daemon-related; frontend tests require pnpm)
- Built-in agents: no empty kebab menu rendered
- User-created agents: archive action still available in kebab menu

Closes #6695

### DIFF
--- a/packages/views/agents/components/agent-detail-page.tsx
+++ b/packages/views/agents/components/agent-detail-page.tsx
@@ -539,7 +539,7 @@ function DetailHeader({
                 {t(($) => $.detail.assign_work)}
               </Button>
             )}
-            {!isArchived && canArchive ? (
+            {!isArchived && canArchive && onArchive ? (
           <DropdownMenu>
             <DropdownMenuTrigger
               render={<Button variant="ghost" size="icon-sm" />}


Bohan-J:
Thanks for this, @Kickflip73 — closing as a duplicate, but your diagnosis was correct and arrived at independently.

#6695 was fixed by #6696, which landed as `8cafe1d08`. That PR gates the same line and is behaviourally identical to yours — I confirmed it by running #6696's test file against your one-line change: all 7 pass. The deciding factor was purely that #6696 also carried regression coverage, which I verified actually fails when the production change is reverted.

Your PR body also called the root cause precisely: `canArchive` derives from `canEdit` and stays true for built-ins, while `onArchive` is undefined for them. That's exactly it.

Sorry for the wasted effort on a race you didn't know you were in — #6696 was opened about two and a half hours before yours, and neither of you had a way to see that. Worth a quick check of open PRs against an issue before starting; that said, the duplicate is on our side for not marking the issue as claimed. We'd welcome another contribution.

PR #6708, opened 2026-08-10
fix(cli): detect desktop-app-managed daemon in 'daemon status'

### LINKED ISSUE #6694: [Bug]: daemon status reports "stopped" while the desktop-app-managed daemon is running
### Deployment type

Official App

### What happened?

`multica daemon status` reports `stopped` even though a daemon is actually running and serving all my local runtimes. The running daemon was spawned by the desktop app (Multica.app on macOS), which runs the bundled CLI binary as `multica daemon start --foreground --profile desktop-api.multica`.

The confusing part: even when I pass that exact profile, status still says stopped:

```
$ multica daemon status
Daemon: stopped

$ multica --profile desktop-api.multica daemon status
Daemon [desktop-api.multica]: stopped
```

Meanwhile the daemon process is alive and working:

```
$ ps aux | grep "multica daemon"
... /Applications/Multica.app/Contents/Resources/app.asar.unpacked/resources/bin/multica daemon start --foreground --profile desktop-api.multica
(listening on 127.0.0.1:19681)

$ multica runtime list
ID        NAME                          MODE   PROVIDER  STATUS  LAST_SEEN
b6dc5afe  Claude (MacBook-Pro.local)    local  claude    online  2026-08-10T07:32:16Z
...four more local runtimes, all online, heartbeats from that daemon...
```

My guess is the app-managed daemon keeps its state (pid file etc.) in an app-managed directory that the CLI's status command doesn't look at — but from a user's point of view, the same binary saying "stopped" about a daemon it can name by profile is misleading.

Impact: relying on `daemon status` you conclude your runtimes are down, or worse, start a second daemon by hand. (I have not tested whether running a CLI-started daemon alongside the app-managed one causes duplicate runtime registrations — I avoided it precisely because status gave me no ground truth.)

Expected instead: `daemon status` detects the running app-managed daemon (at least when invoked with its profile), or explicitly reports something like "daemon is managed by the desktop app — see the app for status" rather than a flat "stopped".

### Steps to reproduce

1. macOS, desktop app installed. Launch Multica.app and let it start its managed daemon (default behavior).
2. Confirm the daemon is alive: `ps aux | grep "multica daemon"` shows the process with `--profile desktop-api.multica`; `multica runtime list` shows local runtimes online.
3. Run `multica daemon status` → reports `Daemon: stopped`.
4. Run `multica --profile desktop-api.multica daemon status` → also reports `stopped`.

```
CLI: multica v0.4.21 (commit: 0dfaac266, built: 2026-08-07T10:21:40Z)
     go1.26.5, darwin/arm64
Note: ~/.local/bin/multica is a symlink into the app bundle, so the CLI I'm
calling and the daemon the app spawned are the same binary/version.
OS: macOS (Darwin 25.5.0), Apple Silicon
```

Written with Claude Code assistance; all outputs above are from live runs on my machine, verified by me.

Kickflip73:
## Summary
`multica daemon status` from the terminal always reports "stopped" for
daemons started by the desktop app. This happens because the desktop
app uses profile names like `desktop-multica.ai` with a distinct health
port (derived from the profile name via `healthPortForProfile`), while
the CLI `daemon status` only checks the default profile port (19514).

## Fix
Add `findDesktopDaemonHealth()` that scans `~/.multica/profiles/desktop-*/`
directories and probes each profile's health port using the same
`healthPortForProfile()` function. When the default profile daemon shows
"stopped" and no explicit `--profile` was given, this scan runs as a
fallback.

This respects the existing design boundary: CLI and desktop profiles
remain separate; the CLI merely discovers the desktop daemon when the
default daemon is not running.

## Testing
- `go build ./server/cmd/multica/` passes
- `go test ./server/cmd/multica/ -run "Daemon"` all pass
- No desktop profiles exist: reports "stopped" as before (no false positive)
- Desktop daemon running: reports "running" with profile info

Closes #6694

### DIFF
--- a/server/cmd/multica/cmd_daemon.go
+++ b/server/cmd/multica/cmd_daemon.go
@@ -1086,6 +1086,32 @@ func requestDaemonShutdown(healthPort int) error {
 
 // --- daemon status ---
 
+// findDesktopDaemonHealth scans ~/.multica/profiles/desktop-* for a running
+// daemon started by the desktop app. Returns its health map and true when found.
+func findDesktopDaemonHealth(ctx context.Context) (map[string]any, bool) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil, false
+	}
+	profilesDir := filepath.Join(home, ".multica", "profiles")
+	entries, err := os.ReadDir(profilesDir)
+	if err != nil {
+		return nil, false
+	}
+	for _, entry := range entries {
+		if !entry.IsDir() || !strings.HasPrefix(entry.Name(), "desktop-") {
+			continue
+		}
+		port := healthPortForProfile(entry.Name())
+		health := checkDaemonHealthOnPort(ctx, port)
+		if health["status"] == "running" {
+			health["profile"] = entry.Name()
+			return health, true
+		}
+	}
+	return nil, false
+}
+
 func runDaemonStatus(cmd *cobra.Command, _ []string) error {
 	profile := resolveProfile(cmd)
 	healthPort, err := daemonStatusHealthPort(cmd)
@@ -1097,6 +1123,14 @@ func runDaemonStatus(cmd *cobra.Command, _ []string) error {
 	defer cancel()
 
 	health := checkDaemonHealthOnPort(ctx, healthPort)
+
+	// When the default profile daemon is stopped, check for a desktop-app-managed
+	// daemon that may be running under ~/.multica/profiles/desktop-*/.
+	if health["status"] == "stopped" && profile == "" {
+		if dHealth, ok := findDesktopDaemonHealth(ctx); ok {
+			health = dHealth
+		}
+	}
 
 	output, _ := cmd.Flags().GetString("output")
 	if output == "json" {


Bohan-J:
Thanks for this, and for the accurate diagnosis — you correctly spotted that the desktop app runs under its own profile with a hashed health port, and that `daemon status` only ever checked the default one.

We ended up taking a larger route than this PR, so I am closing it as superseded rather than asking you to iterate through it.

The part that changed our approach: nothing on the wire says *which* profile a daemon belongs to. Since the port is a `% 1000` hash of the profile name, two names can collide — so scanning ports and attributing whatever answers can credit a daemon to the wrong profile, and `daemon stop` would then act on it. That had to be fixed first:

- `daemon status` now reports an unknown `--profile` as an error instead of a misleading `stopped` (#6757, shipped in v0.4.23)
- `/health` now carries `profile` and `launched_by`; `status` reports "managed by the Desktop app", and `stop` / `restart` / `start` refuse when the daemon on that port belongs to another profile (#6783)

Cross-profile discovery — what this PR set out to do — is the natural next step, and it now has a verified identity to build on, so it can list every live daemon and name each one instead of taking the first that answers.

Appreciate you taking the time to dig in and send a fix.

PR #6710, opened 2026-08-10
MUL-5983 fix(workspace): stop workspace delete from hanging on advisory lock 4246

Bohan-J:
## Problem

On the latest client, clicking **Delete workspace** did nothing: the confirm dialog sat on "Deleting…" indefinitely and the workspace was never deleted (MUL-5983).

The teardown transaction takes global advisory lock `4246` (added in #6230) so the `task_usage_hourly` rollup cannot write aggregates for a workspace that is being torn down — and it waited on that lock with no bound. Nothing above the handler has a deadline either: the API client's `fetch` has no timeout, so any holder of 4246 turns the delete into an infinite spinner with no error and no way out of the dialog.

Lock 4246 has three kinds of holder, and one of them holds it forever:

- a rollup tick, for up to its 25 min scheduler `RunTimeout`;
- `backfill_task_usage_hourly` / `backfill_codex_usage_cache`, for a whole run;
- **a cancelled rollup tick, permanently.** Migration 102 took the lock with the session-scoped `pg_try_advisory_lock` and released it from an `EXCEPTION WHEN OTHERS` handler. plpgsql's `OTHERS` does not match `query_canceled`, and a session-level advisory lock survives the rollback that follows — so a tick killed by a statement timeout, an operator cancel, or pgx cancelling the query when the Go context is done hands its pooled connection back to pgxpool still holding 4246. Every later tick then loses `pg_try_advisory_lock` and reports "no work" (usage aggregation silently stops), and every workspace delete blocks forever.

## Changes

- **Migration 272** makes the rollup's lock transaction-scoped (`pg_try_advisory_xact_lock`), which the transaction end always releases — including a rollback from a cancelled statement. The tick still serialises against other ticks and against the backfill commands.

  One behaviour change: the TTL prune at the end used to run after an explicit unlock and now runs inside the locked transaction, so whatever it costs is added to the window in which 4246 is unavailable — to the next tick, to a backfill run, and to a workspace delete. The prune is an index-driven `DELETE` on `task_usage_hourly_dirty.enqueued_at` with **no row cap**: in steady state each tick already drains the queue and it deletes nothing, but after a long pause it has a real backlog to clear. Worth watching rollup duration and workspace-delete 503s together if that ever happens.
- **`DeleteWorkspace`** sets `SET LOCAL lock_timeout` on its teardown transaction (10 s) and maps SQLSTATE `55P03` to **503** with a retryable message, so lock contention surfaces as an error the dialog can show instead of a request that never returns. The cap bounds lock *waits* only — the teardown work itself stays unbounded, so deleting a large workspace is still allowed to take as long as it takes.
- The sqlc source comment that still described the rollup's lock as session-scoped is corrected (regenerated; comment-only diff).
- `TestDeleteWorkspace_RollsBackResourceLabelCleanup` now drives the handler's own lock timeout instead of injecting one through a custom `txStarter`, which that change made dead code. Its rollback assertions are unchanged; its expected status moves 500 → 503.

## Test isolation

Both new tests contend for 4246, which the rollup family guards across binaries with `lockRollupSingleton`; they now take that guard.

Stressing the combined run surfaced a second, older interference: **every workspace teardown takes 4246 in production code** (#6230), so any handler delete test can own it for a moment and leave the scheduler's rollup tests seeing "no work" — `TestPgCronConcurrentNoDoubleWrite` fails with `winners=0` independently of this PR. Both rollup tests now treat a round lost to an outside holder as a retry rather than a verdict; the invariants they assert (no leaked lock, no double write) are unchanged.

## Verification

Regression tests were confirmed to fail without the fix:

- `TestRollupTaskUsageHourlyDoesNotLeakAdvisoryLockOnCancel` — against migration 102's function: `cancelled rollup tick leaked advisory lock 4246`. Passes after migration 272.
- `TestDeleteWorkspace_FailsFastWhenRollupLockHeld` — with the `SET LOCAL lock_timeout` line removed: `DeleteWorkspace blocked on advisory lock 4246 instead of timing out`. Passes with it (503 in ~0.5 s, workspace intact).

```
go test ./internal/handler                                            ok
go test ./internal/scheduler ./internal/taskusagebackfill              ok
go test ./cmd/backfill_codex_usage_cache ./cmd/migrate                 ok
go test ./internal/handler ./internal/scheduler -count=1               ok (repeated)
  ... -run 'DeleteWorkspace|Rollup|PgCron' -count=15                   ok
  ... -run 'DeleteWorkspace|Rollup|PgCron|Usage' -count=25             ok
scripts/test-go.sh --race                                             46 packages ok
go build ./...                                                         ok
```

The one red package under `--race` is `cmd/multica`, which fails identically on a clean `origin/main` checkout in that environment (a leftover `daemon_task_context.json` in the parent directory trips the CLI's agent-context guard) — unrelated to this change.

Migration 272 applies and rolls back cleanly (`up` → `down` → `up` verified against a live database), and `TestPgCronConcurrentNoDoubleWrite` still sees exactly one winner among six concurrent ticks, so the mutual-exclusion contract is unchanged.

## Rollout

- **The deploy must replace the old backend processes, not just run the migration.** The migration prevents future leaks but cannot release a 4246 that a live session already holds. `newDBPool` sets only `MaxConns`/`MinConns`, so pgxpool's defaults apply: a leaked connection is recycled after 30 min idle / 1 h lifetime at the latest, and a backend rollout clears it immediately. Until then deletes return 503 instead of hanging, and usage aggregation stays stalled.
- Rollout order otherwise does not matter: an old server against the new function behaves the same, and a new server against the old function still fails fast instead of hanging.
- Old desktop clients handle the new 503 fine — the settings page toasts the server's error message and re-enables the dialog.
- Separate, still open: the teardown's task lookups (`agent_id IN (…) OR issue_id IN (…) OR runtime_id IN (…)`) plan as a `Seq Scan on agent_task_queue`, and `task_token` has the same shape. That is a slow delete, not a hung one — tracked as MUL-5999.

### DIFF
--- a/server/internal/handler/resource_label_cascade_test.go
+++ b/server/internal/handler/resource_label_cascade_test.go
@@ -5,9 +5,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
-	"github.com/jackc/pgx/v5"
 )
 
 // Resource-label junction tables (agent_to_label / skill_to_label) deliberately
@@ -290,22 +290,6 @@ func TestDeleteWorkspace_CleansResourceLabelAssignments(t *testing.T) {
 	}
 }
 
-type lockTimeoutTxStarter struct {
-	delegate txStarter
-}
-
-func (s lockTimeoutTxStarter) Begin(ctx context.Context) (pgx.Tx, error) {
-	tx, err := s.delegate.Begin(ctx)
-	if err != nil {
-		return nil, err
-	}
-	if _, err := tx.Exec(ctx, `SET LOCAL lock_timeout = '100ms'`); err != nil {
-		_ = tx.Rollback(ctx)
-		return nil, err
-	}
-	return tx, nil
-}
-
 // TestDeleteWorkspace_RollsBackResourceLabelCleanup verifies the resource-label
 // sweep and the later administration cleanup share one transaction. Locking the
 // workspace's member row makes that late step time out; both junction rows must
@@ -315,6 +299,9 @@ func TestDeleteWorkspace_RollsBackResourceLabelCleanup(t *testing.T) {
 		t.Skip("database not available")
 	}
 	ctx := context.Background()
+	// The teardown transaction sets its own lock_timeout (MUL-5983); shorten
+	// it so the blocked administration step fails while the test is young.
+	setWorkspaceDeleteLockTimeoutForTest(t, 100*time.Millisecond)
 	wsID, agentID, skillID := seedWorkspaceResourceLabelFixture(t, ctx, "handler-tests-delete-labels-rollback")
 
 	blocker, err := testPool.Begin(ctx)
@@ -328,15 +315,14 @@ func TestDeleteWorkspace_RollsBackResourceLabelCleanup(t *testing.T) {
 		t.Fatalf("lock workspace member: %v", err)
 	}
 
-	handler := *testHandler
-	handler.TxStarter = lockTimeoutTxStarter{delegate: testHandler.TxStarter}
-
 	w := httptest.NewRecorder()
 	req := newRequest("DELETE", "/api/workspaces/"+wsID, nil)
 	req = withURLParam(req, "id", wsID)
-	handler.DeleteWorkspace(w, req)
-	if w.Code != http.StatusInternalServerError {
-		t.Fatalf("DeleteWorkspace: expected 500, got %d: %s", w.Code, w.Body.String())
+	testHandler.DeleteWorkspace(w, req)
+	// A lock the teardown cannot get is transient, so the handler reports it
+	// as a retryable 503 rather than a generic failure.
+	if w.Code != http.StatusServiceUnavailable {
+		t.Fatalf("DeleteWorkspace: expected 503, got %d: %s", w.Code, w.Body.String())
 	}
 	if err := blocker.Rollback(ctx); err != nil {
 		t.Fatalf("release member blocker: %v", err)
--- a/server/internal/handler/workspace.go
+++ b/server/internal/handler/workspace.go
@@ -2,13 +2,16 @@ package handler
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/multica-ai/multica/server/internal/analytics"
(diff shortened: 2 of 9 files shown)


Bohan-J:
Review follow-up pushed in `2dbf088`.

**Must-fix — cross-package test isolation.** Reproduced exactly as described: running `internal/handler` and `internal/scheduler` in parallel against one database, the handler tests' hand-held `4246` made the scheduler's lock tests fail. Both new handler tests now call `lockRollupSingleton(t)` before touching the lock.

Stressing the combined run afterwards surfaced a second, older interference worth fixing here: **every workspace teardown takes 4246 in production code** (#6230), so any handler delete test — not just the two new ones — can own the lock for a moment and leave the scheduler's rollup tests seeing "no work". `TestPgCronConcurrentNoDoubleWrite` fails with `winners=0` under that pressure, and it does so independently of this PR's guard fix. Both rollup tests now treat a round lost to an outside holder as a retry rather than a verdict; the invariants they actually assert (no leaked lock, no double write) are unchanged and still fail loudly.

Evidence:

```
# before, with the guard fix only
go test ./internal/handler ./internal/scheduler -run 'DeleteWorkspace|PgCron' -count=15
  FAIL TestPgCronConcurrentNoDoubleWrite: winners=0 losers=6

# after
go test ./internal/handler ./internal/scheduler -run 'DeleteWorkspace|Rollup|PgCron'  -count=15   ok
go test ./internal/handler ./internal/scheduler -run 'DeleteWorkspace|Rollup|PgCron|Usage' -count=25   ok
go test ./internal/handler ./internal/scheduler -count=1   ok  (3x)
scripts/test-go.sh --race   46 packages ok
```

The one red package in the `--race` run is `cmd/multica`, which fails identically on a clean `origin/main` checkout in this environment (a leftover `daemon_task_context.json` in the parent directory trips the CLI's agent-context guard) — unrelated to this change.

**Non-blocking items.**

- `server/pkg/db/queries/workspace_delete.sql` no longer claims the rollup uses a session lock; sqlc regenerated (comment-only diff).
- Migration 272 no longer calls the TTL prune "bounded" or claims it only delays the next tick. It now says what it is: an index-driven `DELETE` with no row cap that, while the lock is held, also delays backfills and workspace deletes, with a pointer to watch rollup duration against delete 503s.
- **Rollout note, confirmed from the pool config:** the migration prevents future leaks but cannot release a `4246` that a live session already holds. `newDBPool` sets only `MaxConns`/`MinConns`, so pgxpool's defaults apply — a leaked connection is recycled after 30 min idle / 1 h lifetime at the latest, and a backend rollout clears it immediately. Until then, deletes return 503 instead of hanging and usage aggregation stays stalled, so the deploy should replace the old backend processes rather than only running the migration.
- The `agent_task_queue` / `task_token` full scans stay out of scope here, as agreed — slow delete, not hung delete.

Issue #6712, opened 2026-08-10
[Bug]: Web attachment download 401s and saves download.txt in token-mode / split-origin self-host (only desktop was fixed by #6092)


ADanMan:
### Deployment type

self-host

### What happened?

On a self-hosted deployment where the browser session is token-mode (`Authorization: Bearer`, no session cookie) — or where the web origin is split from the API host — clicking **Download** on an attachment saves a small `download.txt` instead of the file. The `download.txt` is the `401 auth: no token found` error body written to disk.

**Root cause.** `GET /api/attachments/{id}/download` is entered by a top-level browser navigation — an `<a download>` click — which cannot send an `Authorization: Bearer` header. The only credential such a navigation can present is the `multica_auth` session cookie, which is HttpOnly + SameSite=Strict + host-only. In token-mode sessions, or when the web origin is cross-site to the API host, that cookie does not apply to the request, so the endpoint answers `401` and the browser saves the error body as `download.txt`.

**Why desktop already works but web doesn't.** #6092 fixed exactly this failure mode for the desktop client: in proxy mode it mints a signed, single-attachment capability into `download_url` (`/api/attachments/{id}/signed-download?exp=..&sig=..`), and the Electron bridge downloads `download_url`. But the **web** branch of [`packages/views/editor/use-download-attachment.ts`](https://github.com/multica-ai/multica/blob/main/packages/views/editor/use-download-attachment.ts) never reads `download_url` — it always navigates to the cookie-gated `/api/attachments/{id}/download?workspace_slug=...` endpoint. So the capability that already fixes desktop sits unused on web.

### Steps to reproduce

1. Self-host with `S3_BUCKET` unset (LocalStorage) or `ATTACHMENT_DOWNLOAD_MODE=proxy`.
2. Sign the web app in as a token-mode session, or serve the web origin on a different host than the API.
3. Upload an attachment to an issue comment, then click **Download**.
4. A `download.txt` (401 body) is saved instead of the file. The same attachment's metadata endpoint (`GET /api/attachments/{id}`) succeeds from the same session.

### Additional context (optional)

[...]

PR #6713, opened 2026-08-10
fix(editor): download web attachments via the signed capability URL

ADanMan:
## Summary

`GET /api/attachments/{id}/download` is entered by a top-level browser navigation (an `<a download>` click), which cannot send `Authorization: Bearer`. The only credential such a navigation can present is the `multica_auth` session cookie (HttpOnly, SameSite=Strict, host-only). In token-mode sessions — where auth is a bearer token held in JS, not a cookie — a bare navigation carries no credential at all, so the endpoint answers `401` and the browser saves the error body to disk as `download.txt`.

#6092 fixed this for **desktop**: in proxy mode it mints a signed, single-attachment capability into `download_url` (`/api/attachments/{id}/signed-download?...`), and the Electron bridge downloads `download_url`. But the **web** branch of `useDownloadAttachment` never reads `download_url` — it always navigates to the cookie-gated `/api/attachments/{id}/download?workspace_slug=...` endpoint. So the capability that already fixes desktop was sitting unused on web.

## Change

Client-only, reuses #6092's capability — no server change:

- On web, prefer the capability `download_url` when the server minted one (the `/signed-download` URL, present in proxy mode). It carries its own credential in the query string, so a bare navigation authenticates with no header and no cookie — which is what clears the 401.
- Fall back to the existing cookie-gated slug endpoint for CloudFront/S3 signing modes and for servers that predate #6092.
- Absolute CloudFront/S3 `download_url`s are intentionally not used for the web `<a download>` path: cross-origin and inline-tuned, they would preview instead of download.

Correcting the framing from an earlier revision of this description: the capability clears the **auth** failure (the 401 → `download.txt`). It does **not** force an attachment disposition — it streams via `proxyAttachmentDownload` → `storage.ContentDisposition`, which returns `inline` for `image/*`, `video/*`, `audio/*` and `application/pdf`. What makes the browser download rather than preview is the request staying **same-origin** plus the `download` attribute; `resolvePublicFileUrl` keeps the capability same-origin whenever the app reaches the API on its own origin.

### Outcome after this PR

| scenario | result |
|---|---|
| same-origin, token-mode session | fixed, all file types |
| cross-origin, non-media file | fixed (stored disposition is already `attachment`) |
| cross-origin, image / PDF / video / audio | not fixed — previews in the current tab |
| presign or CloudFront + token-mode | unchanged, still 401s |

**Out of scope** (tracked in #6712, landing as a separate follow-up):

- **presign + token-mode** — the more common self-host setup: an absolute presigned `download_url` is rejected by `capabilityDownloadUrl` and falls back to the cookie-gated endpoint, which 401s in token-mode.
- **cross-origin media** — an image/PDF/video/audio capability resolved to a different origin previews in-tab rather than downloading.
- **CloudFront + token-mode** — unchanged, not regressed.

The fuller fix (a separate download-intent, credential-free URL field on the metadata endpoint, disposition-aware across all three storage modes) is deferred to a follow-up per #6712.

## Test plan

- `packages/views/editor/use-download-attachment.test.tsx`: added web coverage — proxy-mode navigates to the capability URL; the capability resolves against a configured API base (split origin); and an ordering guard that the capability is consulted before the `workspaceSlug` guard. Existing CloudFront → slug-endpoint behavior preserved.
- `pnpm --filter @multica/views test` ✅
- `pnpm --filter @multica/views typecheck` ✅
- `eslint` on changed files ✅

Refs #6712

### DIFF
--- a/packages/views/editor/use-download-attachment.test.tsx
+++ b/packages/views/editor/use-download-attachment.test.tsx
@@ -157,8 +157,9 @@ describe("useDownloadAttachment (web)", () => {
   // #6092 mints a signed capability into `download_url` in proxy mode. A bare
   // `<a download>` navigation can authenticate to it with no Authorization
   // header and no session cookie, so the web path must prefer it over the
-  // cookie-gated slug endpoint that 401s to download.txt in token-mode /
-  // split-origin self-hosting.
+  // cookie-gated slug endpoint that 401s to download.txt in token-mode
+  // self-hosting, where auth is a bearer token in JS and a bare navigation
+  // carries no credential at all.
   it("navigates to the capability download_url in proxy mode, not the cookie-gated endpoint", async () => {
     getAttachmentMock.mockResolvedValueOnce({
       id: "att-1",
@@ -188,8 +189,10 @@ describe("useDownloadAttachment (web)", () => {
     expect(anchor!.getAttribute("href")).toBe(
       "/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
     );
-    // The capability already scopes the request to one attachment and one
-    // user, so the workspace slug is no longer part of the URL.
+    // The capability is itself the credential and is signed over exactly one
+    // attachment id, so the workspace slug is no longer part of the URL. It is
+    // not bound to a user — the signature covers (version, attachment id,
+    // expiry) only — so it stays a bearer capability for its 60-second life.
     expect(anchor!.getAttribute("href")).not.toContain("workspace_slug");
   });
 
--- a/packages/views/editor/use-download-attachment.test.tsx
+++ b/packages/views/editor/use-download-attachment.test.tsx
@@ -154,6 +154,101 @@ describe("useDownloadAttachment (web)", () => {
     );
   });
 
+  // #6092 mints a signed capability into `download_url` in proxy mode. A bare
+  // `<a download>` navigation can authenticate to it with no Authorization
+  // header and no session cookie, so the web path must prefer it over the
+  // cookie-gated slug endpoint that 401s to download.txt in token-mode /
+  // split-origin self-hosting.
+  it("navigates to the capability download_url in proxy mode, not the cookie-gated endpoint", async () => {
+    getAttachmentMock.mockResolvedValueOnce({
+      id: "att-1",
+      url: "https://static.example.test/file.md",
+      download_url:
+        "/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
+      filename: "file.md",
+    });
+    const clickSpy = vi
+      .spyOn(HTMLAnchorElement.prototype, "click")
+      .mockImplementation(() => {});
+    const appendSpy = vi.spyOn(document.body, "appendChild");
+
+    const { result } = renderHook(() => useDownloadAttachment());
+
+    await act(async () => {
+      await result.current("att-1");
+    });
+
+    expect(clickSpy).toHaveBeenCalledOnce();
+    const anchor = appendSpy.mock.calls
+      .map(([node]) => node)
+      .find((node): node is HTMLAnchorElement =>
+        node instanceof HTMLAnchorElement,
+      );
+    expect(anchor).toBeDefined();
+    expect(anchor!.getAttribute("href")).toBe(
+      "/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
+    );
+    // The capability already scopes the request to one attachment and one
+    // user, so the workspace slug is no longer part of the URL.
+    expect(anchor!.getAttribute("href")).not.toContain("workspace_slug");
+  });
+
+  it("resolves the capability download_url against a configured API base (split origin)", async () => {
+    getBaseUrlMock.mockReturnValue("https://api.example.test/");
+    getAttachmentMock.mockResolvedValueOnce({
+      id: "att-1",
+      url: "https://static.example.test/file.md",
+      download_url:
+        "/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
+      filename: "file.md",
+    });
+    const clickSpy = vi
+      .spyOn(HTMLAnchorElement.prototype, "click")
+      .mockImplementation(() => {});
+    const appendSpy = vi.spyOn(document.body, "appendChild");
+
+    const { result } = renderHook(() => useDownloadAttachment());
+
+    await act(async () => {
+      await result.current("att-1");
+    });
+
+    expect(clickSpy).toHaveBeenCalledOnce();
+    const anchor = appendSpy.mock.calls
+      .map(([node]) => node)
+      .find((node): node is HTMLAnchorElement =>
+        node instanceof HTMLAnchorElement,
+      );
+    expect(anchor!.href).toBe(
+      "https://api.example.test/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
+    );
+  });
+
+  // The capability carries its own identity, so a missing workspace slug —
+  // which aborts the legacy slug fallback below — must not block the download.
+  it("downloads via the capability even when the workspace slug is missing", async () => {
+    useWorkspaceSlugMock.mockReturnValueOnce(null);
+    getAttachmentMock.mockResolvedValueOnce({
+      id: "att-1",
+      url: "https://static.example.test/file.md",
+      download_url:
+        "/api/attachments/att-1/signed-download?exp=1735689600&sig=abc123",
+      filename: "file.md",
+    });
+    const clickSpy = vi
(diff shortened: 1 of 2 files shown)


Bohan-J (changes requested):
Approving the approach — preferring the #6092 capability on web is the right call and the obvious missing half of that PR. We're landing this as the narrow client-only fix and tracking the rest separately (see #6712).

**None of the requests below change behaviour.** They're the description, three comments, the issue linkage, and one test. The reason I'm blocking on comment text rather than waving it through: this file's comments are load-bearing — the thing that made the original bug findable was the old comment honestly recording *why* web skipped `download_url`. A comment that states the opposite of what the server does would cost the next person the same investigation.

---

### 1. `use-download-attachment.ts:29` — the capability does not force an attachment disposition

```
// URL, added in #6092). It carries its own credential in the query string and
// forces an attachment Content-Disposition server-side, so a top-level
```

It doesn't. `DownloadAttachmentWithCapability` streams through `proxyAttachmentDownload`, which sets `storage.ContentDisposition(contentType, filename)` (`server/internal/handler/file.go:929`), and that returns **`inline`** for `image/*`, `video/*`, `audio/*` and `application/pdf` (`server/internal/storage/util.go:81`). Only the CloudFront and presign branches of the authenticated `/download` route use the forced-attachment `AttachmentContentDisposition`.

The irony is that lines 35–38 of this same block already state the correct rule:

> Absolute CloudFront / S3 `download_url`s are deliberately NOT used here: they are cross-origin and inline-tuned, so an `<a download>` to them previews the file instead of downloading it.

That reasoning applies to the capability URL too, whenever `resolvePublicFileUrl` resolves it to a different origin. What actually makes this work is same-origin + the `download` attribute — please say that instead.

### 2. `use-download-attachment.ts:31` and `:140` — narrow the "token-mode / split-origin" framing

Both comments treat token-mode and split-origin as one condition. They aren't, in either direction:

- **Cookie side:** `SameSite=Strict` blocks *cross-site* navigation. Web and API on subdomains of one registrable domain with `COOKIE_DOMAIN` set still send the cookie, and cookies ignore ports entirely — so `localhost:3000 → localhost:8080` sends it too. Split-origin alone doesn't break the cookie.
- **Download side:** the `download` attribute requires *same-origin*, and origin **includes the port**. So the cross-port localhost case that keeps the cookie working still loses the `download` attribute.

Net: the two conditions overlap but neither implies the other, and the fix's coverage depends on the second one, not the first.

### 3. PR description — replace the claim and complete the scope section

Same correction as (1), plus the outcome matrix as it actually stands:

| scenario | after this PR |
|---|---|
| same-origin, token-mode session | fixed, all file types |
| cross-origin, non-media file | fixed (stored disposition is already `attachment`) |
| cross-origin, image / PDF / video / audio | not fixed — previews in the current tab |
| presign or CloudFront + token-mode | unchanged, still 401s |

"Out of scope" currently names only CloudFront + token-mode; please extend it to presign (the more common self-host setup — an absolute presigned `download_url` is rejected by `capabilityDownloadUrl` and falls back to the cookie-gated endpoint) and to cross-origin media.

### 4. `Fixes #6712` → `Refs #6712`

The issue title covers the split-origin and media rows this PR doesn't close, so auto-closing it on merge would lose the tracking. We'll close it when the follow-up lands.

### 5. `use-download-attachment.test.tsx` — the "workspace slug is missing" case can't happen

"downloads via the capability even when the workspace slug is missing" doesn't exercise a reachable path: `api.getAttachment` resolves workspace context server-side and answers `400 workspace_id is required` when there is none, so the preflight fails before any capability is read. The mock makes it pass, but it pins a scenario the app can't produce. Either drop it, or keep it and re-comment it as what it actually pins — that `capabilityDownloadUrl` runs before the `workspaceSlug` guard.

The other two added cases are good; the split-origin one is exactly the right thing to pin.

---

### One question before merge

Related to (2), and it decides whether this closes your own repro: what is `NEXT_PUBLIC_API_URL` set to in your deployment, and was the attachment that failed an image/PDF/video/audio or an ordinary file? If the browser reaches the API through the same-port `/api` proxy, this fixes your case for every file type. If it points straight at the API port and the file was media, you'd get an in-tab preview instead of a download — still better than `download.txt`, but we'd want that recorded accurately rather than discovered later.

ADanMan:
Thanks — all five are in, and none of them touch behaviour. Commit `0c52722`, plus the description above.

**1. "forces an attachment disposition" (`:29`)** — gone. The comment now says what actually happens: the capability streams through `proxyAttachmentDownload` → `storage.ContentDisposition`, which returns `inline` for `image/*`, `video/*`, `audio/*` and `application/pdf`. The download comes from the request staying same-origin plus the `download` attribute; `resolvePublicFileUrl` keeps it same-origin only when the app reaches the API on its own origin. The credential in the query string is what clears the 401 — separate from disposition, and the comment now keeps the two apart.

**2. token-mode vs split-origin (`:31`, `:140`)** — no longer conflated. Both comments now attribute the 401 to token-mode specifically (auth is a bearer token in JS, so a bare navigation carries no credential at all), and leave the same-origin/`download`-attribute condition to the `capabilityDownloadUrl` block where it belongs. The cookie-vs-origin distinction you spelled out — SameSite is same-*site*, `download` needs same-*origin*, ports count for one and not the other — is no longer collapsed into a single "token-mode / split-origin" phrase.

**3. Description** — replaced the claim and added the outcome matrix; extended "out of scope" to presign + token-mode (flagged as the more common self-host case) and cross-origin media, with CloudFront + token-mode kept as before.

**4. `Fixes #6712` → `Refs #6712`** — done, so the split-origin and media rows stay tracked until the follow-up.

**5. The missing-slug test** — kept, but re-titled and re-commented as the ordering guard it actually is: it pins that `capabilityDownloadUrl` is consulted before the `workspaceSlug` guard, not a reachable user scenario (the metadata preflight 400s without a workspace first). The other two added cases are unchanged.

---

**On your pre-merge question** (`NEXT_PUBLIC_API_URL` + file type) — I can pin the file-type half and want to be honest about the origin half rather than guess.

The failure hit **every type I tried — html, png, txt, pptx, pdf — all saved as `download.txt`.** That uniformity is the tell: a non-media file (txt/html/pptx) over a cross-origin capability would still *download* (its stored disposition is already `attachment`), so it wouldn't produce an error body. All five landing as `download.txt` means the pre-fix failure was the **401 itself**, uniform across types — the token-mode leftover we already agreed on, which this PR clears.

What that uniform-401 signature can't tell us on its own is same-origin vs cross-origin — the 401 looked identical either way before the fix. So for the media rows specifically (png/pdf), whether this ends in a real download or an in-tab preview depends on whether the browser reached the API same-origin, and I'd rather not assert a `NEXT_PUBLIC_API_URL` value I haven't re-confirmed. Net for my repro: non-media is a definite fix; media is at minimum "preview instead of `download.txt`," and a full fix if same-origin — which is exactly the cross-origin-media row the #6712 follow-up closes.

Bohan-J (approved):
All five are addressed — thank you for the fast and careful turnaround. Approving.

I pushed one commit (`2782cbf`) rather than sending you round again for three comment lines. Comments only, no behaviour change:

1. **`use-download-attachment.test.tsx:191` — the capability is not user-bound.** The comment claimed it "scopes the request to one attachment and one user". `signAttachmentCapability` signs `(version, attachment id, expiry)` and nothing else (`server/internal/handler/attachment_capability.go:85`), so it's a bearer capability for its 60-second life — anyone holding the URL can redeem it. That distinction matters more than the others here: whoever next reviews the leak surface or asks "can we revoke this?" would start from a false premise.
2. **`use-download-attachment.test.tsx:160` — token-mode / split-origin still conflated** in that one test comment. Narrowed it to token-mode, which is what actually produces the 401.
3. **`use-download-attachment.ts:91`** — dropped a stale "same-origin" from the fallback endpoint description. That endpoint resolves through `resolvePublicFileUrl` like everything else here, so it isn't necessarily same-origin either. Since this whole review turned on same-origin precision, leaving that one behind seemed worse than fixing it.

Everything else reads accurately now — the description's outcome matrix, the corrected disposition explanation, the `Refs` linkage, and the re-framed ordering guard. The ordering-guard rewrite in particular is better than what I asked for: it says plainly that it isn't a reachable scenario and pins the client-side ordering instead.

On the `NEXT_PUBLIC_API_URL` question: your answer was the right one to give. Declining to assert a value you hadn't re-confirmed, and reasoning from the uniform `download.txt` across html/png/txt/pptx/pdf to "the pre-fix failure was the 401 itself", is exactly the shape of evidence that holds. Non-media is a definite fix for your repro; media is at minimum an improvement and a full fix if same-origin. The remaining rows are #6712's follow-up, which you've claimed — looking forward to that one, and thanks for scoping it to the full matrix up front.

Merging once CI is green.

Issue #6659, opened 2026-08-09
[Feature]: Keep the issue list as a collapsible left rail on the issue detail page (master-detail layout)


HuAzurestar:
### Deployment type

Official App

### What do you want and why?

Switching between the issues list and an issue's detail page currently requires two separate page loads - enter a detail page, then go Back to get the list again. With many issues open, the list?detail round trip feels slow and loses your place.

Requested behavior: entering an issue's detail page keeps a compact issue list on the left as a rail (master-detail layout, mirroring the existing inbox/chat split), so you can move continuously between list and detail. The rail can be collapsed/expanded, and the expanded/collapsed state persists as the default for the next time you open a detail page.

### Proposed solution

- `/{ws}/issues/{id}` becomes a two-column layout on non-mobile viewports: a left issue list + the existing detail panel on the right (unchanged), using the same `ResizablePanelGroup` pattern as inbox (`packages/views/inbox/components/inbox-page.tsx`).
- Left rail shows the workspace's issues in a compact single-column list (identifier + title + status) with the current route's issue highlighted; clicking a row switches the detail (`navigation.replace`). It does not restore list-page filters/scope in v1.
- A collapse/expand toggle in the rail header. The rail collapses to a narrow strip (icon + count) and re-expands on click.
- The collapsed state is persisted in a small workspace-scoped localStorage store, so the user's last choice becomes the default on the next detail-page entry. No settings-page switch in v1.
- Mobile (<lg breakpoint): no left rail - full-width detail as today.

### Screenshots / mockups

N/A - see implementation PR (Closes #6659).

HuAzurestar:
I'm claiming this one. The change lands as a two-column master-detail layout on the issue detail route (`IssueDetailRoute`), backed by a new workspace-scoped persisted preference (default: expanded) with a Settings ? Preferences switch. Implementation PR will close this issue (`Closes #6659`).

HuAzurestar:
Updated to the converged v2 scope: no settings-page switch, no drag-resize width memory, no list-page filter/scope restore in v1; collapsed state persisted via a small workspace-scoped store.

PR #6716, opened 2026-08-10
MUL-5989 feat(views): list rail + detail two-column layout for autopilot and agent detail pages

### LINKED ISSUE #6715: [Feature]: Keep autopilot and agent lists as a collapsible left rail on their detail pages
## Feature

Keep the Autopilots and Agents list as a collapsible left rail on their detail pages (`/autopilots/[id]` and `/agents/[id]`), mirroring the issue-detail two-column pattern (#6659).

## Spec

1. **Three-state layout**: full-width detail below the desktop breakpoint; a collapsed 44px rail (expand button + count) by default; an expanded `ResizablePanelGroup` with the list on the left (default 320, min 240, max 480) and detail on the right (min 40%).
2. **In-place navigation**: rail rows `replace` to the sibling detail URL; the current row is highlighted.
3. **Persisted, workspace-scoped store** (`multica_list_detail_split`): `collapsed` (default true) + dragged rail `size` (clamped 240–480), survives refresh.
4. **Scroll position kept** when switching items in place — no remount, no jump to top.
5. **Terminology**: external copy uses "Autopilots", never "Automations".
6. **Settings preference** (Settings → Preferences): "Collapse list on detail pages", on by default, `toast` feedback.
7. **i18n**: en / zh-Hans / ja / ko synced.

## Acceptance

- Rail collapse/expand toggle, drag-width persistence, in-place switching, and settings toggle all work on web and desktop.
- Store unit test + Route/Rail component tests; `make check` green.

HuAzurestar:
Closes #6715

Multi-agent contribution for NEX-37 (Autopilots / Agents list-rail detail layout). Mirrors the issue-detail two-column pattern from #6661 / #6659. See the linked issue for the full spec.


Bohan-J:
Can you upload a screenshot of your UI changing so that I could have a quick review on it? Thx

HuAzurestar:
Yes of cause. Here you are.

1. The initial page of agent is as origin.
aget-out

2. To click on one agent, you will have a left-side column of agent list.
agent-in

3. You can click on any agent here to switch.
agent-changeinside

4. You can unfold the left-side column to make the UI clean.
agent-unfold
 
5. Initial for autopilot as well.
autopilot-out

6. Click on one autopilot as well.
autopilot-in

7. Unfold to make clean as well for autopilot.
autopilot-unfold

8. And switch as well for autopilot.
autopilot-changeinside

Bohan-J:
Thanks for the screenshots — that's exactly what I needed, and thanks for putting real work into this.

I'm going to close this PR, but I want to be clear that it's the visual direction I'm turning down, not the idea. The underlying request in #6715 is one we agree with, and I'm keeping that issue open.

What the screenshots surfaced: with the rail expanded, the agent detail page ends up with four vertical columns (global nav, list rail, content, properties card), and the content column gets squeezed noticeably. The rail also reads as a second navigation sidebar rather than as part of the page — it uses the same density and treatment as the global nav sitting right next to it. And the collapsed state, which is the default, is a 44px strip holding only an icon and a bare number; in the autopilot screenshot that lone "2" reads more like a rendering artifact than a feature. Your own workspace shows the other half of the problem: 32 agents overflow the rail with no search or grouping, and "many items" is precisely the case this was meant to help.

Getting this right means settling the visual hierarchy first — how the two columns differentiate themselves, whether a collapsed state should exist at all, what a row actually shows. That's a design call on our side, and it isn't fair to ask you to guess at it across review rounds. So we'll take the design work in-house and build on it from there.

Separately, the implementation review notes still stand if they're useful to you elsewhere. The most important one: `ListDetailLayout` returns a different root element type for each of its three states (Fragment / `div` / `ResizablePanelGroup`), so toggling collapse unmounts and remounts the entire detail pane — any in-progress form state in it is lost, and the scroll position resets. Two smaller ones: the rail doesn't filter archived items, so its rows and count disagree with the list page (`agentListOptions` requests `include_archived: true`); and the settings toggle reads and writes the live `collapsed` value rather than a separate default, so expanding the rail on a detail page silently flips the preference in Settings.

Really do appreciate the contribution, and I hope this doesn't put you off opening the next one.

PR #6717, opened 2026-08-10
MUL-5510: fix(agent): preserve newlines in Antigravity streamed text

### LINKED ISSUE #6149: [Bug]: Markdown format in chat window rendered incorrectly
### Deployment type

Official App

### What happened?

Markdown syntax in the chat window is rendered incorrectly.

I suggest using [marked](https://github.com/markedjs/marked) as a Markdown parser.

### Steps to reproduce

1. Go to Chat.
2. Start a chat with an agent with a random message.

### Screenshots (optional)

Image

```shell
Multica desktop version: 0.4.14
OS: Mac OS 26.5.2
```

ADanMan:
## Summary

An agent's chat reply rendered with every line break gone: block markdown like `### Heading` and `- item` ended up mid-line and displayed as literal text (`bạn!### 🚀`, `gì?- **Quản lý**`). See #6149.

Root cause is the **Antigravity runtime adapter** (`server/pkg/agent/antigravity.go`). It scans the agent's stdout line by line and emitted each streamed `MessageText` as the bare line **without its `\n`** (and skipped blank lines). The daemon accumulates streamed text with `pendingText.WriteString(msg.Content)` — no separator — so the persisted `task_message` `type=text` rows lost every newline. The chat renders those rows (the timeline), which is why block markdown collapsed onto one line.

`chat_message.content` was unaffected — the adapter rebuilds `output` with newlines separately — so only the streamed/persisted timeline text was broken. Other backends (codex / claude / hermes / cursor) stream whole blocks that already carry their `\n`, so only Antigravity was affected; the daemon concatenation, the frontend `.join("")`, and `packages/ui/markdown` are all correct.

## Change

`server/pkg/agent/antigravity.go`: stream each line so it reconstructs `output` exactly — prefix the newline on every line after the first, and stream blank lines too (they are the paragraph separators). `output` itself is untouched, so `chat_message.content` stays identical; no change to the daemon, the frontend, or the renderer.

## Test plan

- `server/pkg/agent/antigravity_test.go`: `TestAntigravityBackendStreamPreservesNewlines` runs the backend against a fake `agy` that streams a blank-line-separated heading + list, reconstructs the streamed text the way the daemon does (concatenate `MessageText.Content` with no separator), and asserts it equals the reply **with** its newlines. Fails before the fix (everything glued onto one line), passes after.
- `go test ./pkg/agent` ✅ (full package, incl. the new test)
- `go vet ./pkg/agent` ✅
- `gofmt` clean

Fixes #6149

### DIFF
--- a/server/pkg/agent/antigravity.go
+++ b/server/pkg/agent/antigravity.go
@@ -130,12 +130,22 @@ func (b *antigravityBackend) Execute(ctx context.Context, prompt string, opts Ex
 
 		for scanner.Scan() {
 			line := scanner.Text()
+			// The daemon concatenates streamed MessageText with no separator
+			// (pendingText.WriteString), so the streamed text must carry the
+			// line breaks itself. Mirror output's construction — prefix the
+			// newline on every line after the first, and stream blank lines
+			// too — so the persisted task_message text reconstructs output
+			// exactly. Emitting bare, blank-skipped lines dropped every newline,
+			// collapsing block markdown (headings, lists) onto one line in
+			// chat (#6149).
+			chunk := line
 			if output.Len() > 0 {
 				output.WriteByte('\n')
+				chunk = "\n" + line
 			}
 			output.WriteString(line)
-			if strings.TrimSpace(line) != "" {
-				trySend(msgCh, Message{Type: MessageText, Content: line})
+			if chunk != "" {
+				trySend(msgCh, Message{Type: MessageText, Content: chunk})
 			}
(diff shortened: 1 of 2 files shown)


multica-eve (approved):
Thanks @ADanMan for tracing the issue past the renderer and adding a focused regression test. The adapter-level fix is well scoped, preserves the existing rendering and sanitization path, and passes the full agent package validation. Approved!

PR #6722, opened 2026-08-10
feat(codex): add bounded operational context mode

AMoreaux:
## Summary

- add an opt-in codex.context_mode=operational policy while preserving inherited behavior by default
- rebuild the task Codex home with only explicit Multica instructions and assigned skills
- start every operational task on a fresh app-server thread without ambient MCP or custom Codex arguments
- characterize the boundary in CI against an exact Codex 0.146.0 binary and document that this is context selection, not a sandbox

## Verification

- cd server && go test ./... -count=1
- cd server && MULTICA_RUN_REAL_AGENT_SMOKE=1 go test -tags=agentintegration ./internal/daemon -run ^TestOperationalCodexRealBinaryContextBoundary$ -count=1 -v

The real-binary test uses a local recording endpoint and does not access an OpenAI account or consume model quota.


AMoreaux:
Closing this fork-based PR at the repository owner's request. The work will be republished through the existing Multica repository workflow.

PR #6727, opened 2026-08-10
feat(issues): inherit parent's project and assignee when creating a sub-issue

### LINKED ISSUE #6724: feat(issues): inherit parent's project and assignee when creating a sub-issue
Creating a sub-issue from a parent (the "Add sub-issue" entry) does not inherit the parent's assignee, so the user has to re-pick the owner every time.

Reported in discussion #1728 ("Default Project"): sub-issue creation should be contextual and inherit the parent's Project and Agent instead of defaulting to none.

On `main`, `openCreateSubIssue` in `packages/views/issues/actions/use-issue-actions.ts` already seeds `project_id` from the parent, and the create-issue modal already consumes seeded `assignee_type`/`assignee_id`. The remaining gap is that the assignee is never seeded. Seed the parent's assignee alongside the project so a sub-issue starts with the same owner.

Refs #1728

ADanMan:
## Summary

Discussion #1728 ("Default Project") reports that creating a sub-issue from a parent is not contextual: it doesn't inherit the parent's Project or Agent, so the user has to set them manually every time.

On `main`, `openCreateSubIssue` in `packages/views/issues/actions/use-issue-actions.ts` already seeds the parent's `project_id` into the create-issue modal, and the modal already consumes a seeded `assignee_type`/`assignee_id` (see `create-issue.tsx`). The only remaining gap is that the assignee was never passed through — so a sub-issue defaulted to no owner.

## Change

- Seed the parent's assignee (`member`/`agent`/`squad`) alongside the project in `openCreateSubIssue`.
- Both fields are seeded together, since `assignee_type` is meaningless without `assignee_id`; when the parent is unassigned, nothing is seeded (unchanged behavior).
- No backend or API change: this reuses the create-issue modal's existing seed-consumption path.

## Test plan

- Added two cases to `use-issue-actions.test.tsx`:
  - `openCreateSubIssue` seeds the parent's project **and** assignee.
  - `openCreateSubIssue` omits the assignee when the parent has none.
- `pnpm --filter @multica/views exec vitest run use-issue-actions` — 13 passed.
- `pnpm --filter @multica/views typecheck` — clean.
- `go test ./internal/handler -run 'Issue|Create'` — ok (unaffected, confirms no backend regression).

Fixes #6724
Refs #1728

### DIFF
--- a/packages/views/issues/actions/__tests__/use-issue-actions.test.tsx
+++ b/packages/views/issues/actions/__tests__/use-issue-actions.test.tsx
@@ -250,6 +250,48 @@ describe("useIssueActions", () => {
     });
   });
 
+  it("openCreateSubIssue seeds the parent's project and assignee so the sub-issue inherits them", () => {
+    const parentIssue = {
+      ...mockIssue,
+      project_id: "project-1",
+      assignee_type: "agent",
+      assignee_id: "agent-1",
+    } as Issue;
+    const { result } = renderHook(() => useIssueActions(parentIssue), { wrapper });
+
+    act(() => {
+      result.current.openCreateSubIssue();
+    });
+
+    expect(mockOpenModal).toHaveBeenLastCalledWith("create-issue", {
+      parent_issue_id: "issue-1",
+      parent_issue_identifier: "TES-1",
+      project_id: "project-1",
+      assignee_type: "agent",
+      assignee_id: "agent-1",
+    });
+  });
+
+  it("openCreateSubIssue omits assignee when the parent has none", () => {
+    const parentIssue = {
+      ...mockIssue,
+      project_id: "project-1",
+      assignee_type: null,
+      assignee_id: null,
+    } as Issue;
+    const { result } = renderHook(() => useIssueActions(parentIssue), { wrapper });
+
+    act(() => {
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for this, @ADanMan — tightly scoped change, and the write-up made it easy to verify. Reading `openCreateSubIssue` against the modal's seed-consumption path, your diagnosis was exactly right: the seed was already consumed, the assignee just never got passed through.

Two details worth calling out, because they're the parts that could easily have gone wrong:

- Seeding `assignee_type` and `assignee_id` as a pair, so the payload can never carry a half-set assignee.
- Omitting both keys for an unassigned parent instead of seeding `null`. The modal keys off whether the fields are *present*, not their value, so seeding `null` would have silently cleared the user's sticky last-used assignee — and your second test pins exactly that case.

I pushed one comment-only follow-up to your branch before merging: the inline comment said a sub-issue previously "defaulted to none", which isn't quite what happened — without a seed the modal falls back to the last-used assignee from the manual draft. Reworded so the next reader doesn't misjudge the fallback. No behavior change.

All checks green on the final commit. Merged — thanks again for the contribution!

PR #6728, opened 2026-08-10
feat(shortcuts): browser-style Cmd+[ / Cmd+] history navigation

ADanMan:
## Summary

Adds browser-style keyboard navigation through the shared in-app history:

- **Mod+[** (Cmd+[ / Ctrl+[) — go **back**
- **Mod+]** (Cmd+] / Ctrl+]) — go **forward**

Requested in discussion #1727 ("Keyboard Commands"). Works on both web and desktop.

## Change

- The two bindings are registered as configurable **navigation** actions (`goBack`, `goForward`) in the existing keyboard-shortcut registry (`packages/core/shortcuts`), so they appear under Settings → Keyboard Shortcuts and can be rebound or disabled like any other action. Neither bracket is app- or browser-reserved, so both are recordable on every platform/runtime.
- `NavigationAdapter` gains an optional `forward()` — web wires it to `router.forward`, desktop to the tab store's `goForward()`. `back()` already existed.
- `GlobalShortcuts` maps the two new actions to `navigation.back()` / `navigation.forward?.()`. The chords are ignored while focus is in an input, textarea, or contenteditable (`allowInEditable: false`).
- Added `goBack` / `goForward` labels+descriptions to all locale bundles (en, ja, ko, zh-Hans).

`forward()` is optional on the adapter so an isolated popout window without a forward stack simply no-ops.

## Test plan

- New `packages/views/layout/global-shortcuts.history.test.tsx`: asserts Mod+[ calls `back()`, Mod+] calls `forward()`, the chord is a no-op when the adapter has no `forward`, and it is ignored while an input is focused.
- Existing `packages/core/shortcuts` suite (incl. the "every shipped default is within the safety policy" guard) passes with the new defaults.
- Locale parity test passes with the new keys.
- `pnpm --filter @multica/views typecheck` and core/web/desktop typechecks pass.

Refs #1727

### DIFF
--- a/apps/desktop/src/renderer/src/hooks/use-navigation-input-bindings.test.tsx
+++ b/apps/desktop/src/renderer/src/hooks/use-navigation-input-bindings.test.tsx
@@ -55,24 +55,36 @@ describe("useNavigationInputBindings", () => {
     seedHistory();
   });
 
-  it.each([
-    ["[", "Cmd+["],
-    ["ArrowLeft", "Cmd+Left"],
-  ])("goes back on %s", (key) => {
+  it("goes back on Cmd+Left", () => {
     render(<Probe />);
-    const event = keydown({ key, metaKey: true });
+    const event = keydown({ key: "ArrowLeft", metaKey: true });
     expect(activeHistoryIndex()).toBe(0);
     expect(event.defaultPrevented).toBe(true);
   });
 
-  it.each([
-    ["]", "Cmd+]"],
-    ["ArrowRight", "Cmd+Right"],
-  ])("goes forward on %s", (key) => {
+  it("goes forward on Cmd+Right", () => {
     useTabStore.getState().goBack();
     render(<Probe />);
-    keydown({ key, ctrlKey: true });
+    keydown({ key: "ArrowRight", ctrlKey: true });
     expect(activeHistoryIndex()).toBe(1);
+  });
+
+  // #6728: the Cmd/Ctrl+[ and +] chords moved to the shared shortcut registry
+  // (packages/core/shortcuts, driven by <GlobalShortcuts>). This hook must NOT
+  // also handle them — DesktopShell mounts both this window listener and the
+  // document-level GlobalShortcuts, so a bracket handled here too would
+  // double-navigate (one press = two steps, hidden below a history depth of 3
+  // by stepHistory's clamp). It must also not preventDefault, so the registry
+  // still receives the keydown.
+  it.each([
+    ["[", "Cmd+["],
+    ["]", "Cmd+]"],
+  ])("leaves %s to the shortcut registry (no double navigation)", (key) => {
+    render(<Probe />);
+    const before = activeHistoryIndex();
+    const event = keydown({ key, metaKey: true });
+    expect(activeHistoryIndex()).toBe(before);
+    expect(event.defaultPrevented).toBe(false);
   });
 
   it("leaves the chord to editable targets (caret navigation wins)", () => {
@@ -87,7 +99,7 @@ describe("useNavigationInputBindings", () => {
 
   it("ignores the chord with extra modifiers held", () => {
     render(<Probe />);
-    keydown({ key: "[", metaKey: true, shiftKey: true });
+    keydown({ key: "ArrowLeft", metaKey: true, shiftKey: true });
     expect(activeHistoryIndex()).toBe(1);
   });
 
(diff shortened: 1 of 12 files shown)


Bohan-J:
Thanks for the PR — and for routing this through the shortcut registry instead of hardcoding the chords. That part is the right approach.

One heads-up before this can go in: the desktop half of this already landed in #6667, which merged roughly 12 hours before you opened this PR, so your branch almost certainly predates it. `useNavigationInputBindings` in `apps/desktop/src/renderer/src/hooks/use-tab-history.ts` already binds:

- `Cmd/Ctrl + [` / `]`
- `Cmd/Ctrl + ←` / `→`
- mouse side buttons 3 / 4

with the same input / textarea / contenteditable guard you have here.

What is still genuinely new in this PR:

- Registering `goBack` / `goForward` in `packages/core/shortcuts`, so they appear under Settings → Keyboard Shortcuts and can be rebound or disabled. The desktop bindings are hardcoded and can't be.
- The web wiring and the optional `forward()` on `NavigationAdapter`. Web has no bindings at all today, and on Windows/Linux `Ctrl + [` isn't a browser default (browsers use `Alt + ←/→`), so this is a real gain there.

The blocker is on desktop. `DesktopShell` mounts both `useNavigationInputBindings()` (listener on `window`) and `<GlobalShortcuts />` (listener on `document`). A keydown reaches `document` before `window`, and `useNavigationInputBindings` doesn't check `event.defaultPrevented` — so after this PR a single `Cmd + [` calls `goBack()` twice on desktop: once via `GlobalShortcuts → navigation.back()`, then again via the existing hook.

This is easy to miss in manual testing because `stepHistory` clamps at the ends of the stack: with a two-entry history the second call is a no-op and everything looks correct. The double step only shows up with three or more entries.

A smaller related point: since the desktop chords are hardcoded, rebinding or disabling `goBack` in Settings won't actually turn off `Cmd + [` on desktop, which doesn't match what the registry entry promises.

Suggested fix: drop the `"["` / `"]"` branches from `useNavigationInputBindings`, keeping `Cmd/Ctrl + ←/→` and the mouse side buttons there. Both surfaces then get the brackets from the registry, which clears the double navigation and the unconfigurable-binding mismatch in one change. A desktop test asserting one press = one history step against a stack of at least three entries would be worth adding, since the bounds clamping hides the bug at smaller depths.

You'll also want to rebase on `main`.

ADanMan:
Good catch on the double-navigation — you're right that `DesktopShell` mounts both `useNavigationInputBindings` (window) and `<GlobalShortcuts>` (document), so a single `Cmd+[` stepped twice, with `stepHistory`'s clamp hiding it below a depth of 3.

Fixed: dropped the `[` / `]` branches from `useNavigationInputBindings`, keeping `Cmd/Ctrl+←/→` and the mouse side buttons there. Both web and desktop now take the rebindable brackets from the shared registry in one place, which also resolves the unconfigurable-binding mismatch you flagged. Updated the desktop hook's tests to assert it now leaves `[` / `]` to the registry (no navigation, no `preventDefault`), and refreshed the doc comment.

I kept the branch on its current base for a clean fast-forward update rather than a history rewrite — happy to rebase on `main` if you'd prefer that to a rebase-merge.

Bohan-J:
Thanks for turning this around so quickly — the fix is exactly right.

I verified it rather than just reading the diff. Mounting both handlers together the way `DesktopShell` does, against a three-entry history (the depth where `stepHistory`'s clamp stops hiding the bug), one `Cmd+[` now moves exactly one entry. The same check fails on `801fa46` with index 2 → 0 and `Cmd+]` going 0 → 2, so the regression it was meant to close is genuinely closed. Locally: the 8 desktop binding tests, the 4 history-wiring tests, the 32 `packages/core` shortcut tests including the safety-policy guard, and the 160 locale parity tests all pass, and `core` / `views` / `desktop` typecheck cleanly.

I pushed one small follow-up commit to your branch (`2d1ef542a`) touching only the `forward?:` doc comment in `packages/views/navigation/types.ts`. The original wording cited the popout window as an adapter where the chord no-ops, which reads as if the shortcuts reach that window — they don't, since `issue-window.tsx` never mounts `<GlobalShortcuts>`. Your reasoning for keeping `forward()` optional still holds (that adapter steps back with `navigate(-1)` and has no forward counterpart); the comment now states it as a contract property instead. No behavior change.

Two things we're aware of and are not asking you to address here: the chord calls `navigation.back()` directly rather than going through the `canGoBack()` guard, which is the right call for a browser-convention shortcut but does mean a cold-loaded web page can step out of the app; and on keyboard layouts where `[` needs AltGr/Option the default isn't reachable or recordable, which is a pre-existing property of the shortcut model, not something this PR introduced.

Nice first contribution — clean scope, good tests, and a precise fix on review feedback. Merging now.

Issue #6474, opened 2026-08-06
[Feature]: Add durable workspace-scoped pause/resume for task claim intake


rongchenzheng:
## Deployment type

Self-hosted and Cloud deployments that use Daemon task claiming, especially shared-Daemon and multi-consumer installations.

## What do you want, and why?

Add an authoritative, durable, auditable, Workspace-scoped control that pauses and resumes **task claim intake** without stopping a Daemon or cancelling work that has already started.

Today the server exposes both legacy per-runtime claim and machine-level batch claim endpoints, including stale-dispatch reclaim, but no Workspace maintenance fence or status/action ledger. The Daemon has an in-memory `pauseClaims` barrier for auto-update/restart safety, but that state is process-local, not Workspace-scoped, not durable, not operator-callable, and cannot fence a second Daemon/consumer. It therefore cannot safely support a maintenance cutover.

The safety property needed from a successful pause acknowledgement is:

> After the pause ACK, no task in that Workspace can newly become consumer-owned through claim or stale-dispatch reclaim. Claims committed before the fence remain visible as pre-fence work and can drain; pause does not cancel active work.

An HTTP handler pre-check is not sufficient because it races with the queue transition. The Workspace fence must participate in the same database transaction/concurrency protocol as every path that dispatches or re-delivers a task.

### 1. Durable state and append-only actions

Persist at least:

```text
workspace_id
state: paused | resumed
generation
updated_at
updated_by_type
updated_by_id
reason
last_action_id
```

Also persist an append-only action/audit record with:

```text
action_id
workspace_id
requested_action
previous_state
resulting_state
generation
actor_type
actor_id
auth_source
reason
idempotency_key
requested_at
effective_at
request/result status
error class (for failures)
```

Do not record tokens, PATs, API keys, authorization headers, or other credential material.

The existing issue-oriented activity timeline may be reused as an additional event, but durable state must not be derived solely from an inaccessible `activity_log` row with `issue_id = NULL`. Operators need a Workspace-scoped read API for both current state and action history.

### 2. Operator API and CLI

Suggested API shape:

```text
GET  /api/workspaces/{workspaceId}/claim-intake
POST /api/workspaces/{workspaceId}/claim-intake/pause
POST /api/workspaces/{workspaceId}/claim-intake/resume
GET  /api/workspaces/{workspaceId}/claim-intake/actions
```

Suggested CLI shape:

```text
multica workspace claim-intake status <workspace-id> --output json
[...]

PR #6729, opened 2026-08-10
MUL-6001: feat(server): ERI-204 durable Workspace claim-intake pause/resume

rongchenzheng:
## ERI-204 — durable Workspace claim-intake control

Associates this non-production upstream candidate with ERI-204 and upstream issue #6474.

This pull request adds a durable, authoritative, auditable Workspace-scoped pause/resume control for new task-claim intake.

### What changes

- Adds Workspace control state and append-only action records with generation, actor/provenance, reason, idempotency, exact stored response replay, and timestamps.
- Adds HTTP APIs and `multica workspace claim-intake` status, pause, resume, actions, and ledger commands.
- Fences legacy singular claims, batch claims, stale-dispatch reclaim, WebSocket-first claims, `/api/daemon/claim`, and old-server legacy fallback through the same fail-closed database control-row lock.
- Locks mixed-Workspace controls deterministically and reports paused Workspace metadata without blocking active Workspaces.
- Stamps task ownership with claim-intake generation, action, and consumer identity.
- Adds Workspace-global active-state counts and per-task fence classification.
- Strips client-supplied provenance and permits mutations only for authenticated human Workspace owners/admins.
- Initializes existing and new Workspaces and provides a candidate-only rollback chain.

### Safety semantics

After the pause acknowledgement, no task in that Workspace can newly become consumer-owned through a fresh claim or stale-dispatch reclaim. Ownership committed before the fence remains visible as pre-fence work and may drain; pause does not cancel active work.

### Verification

Candidate commit: `5e358541c4a20c7213effe564e6fcea5e858552b`

Fresh disposable PostgreSQL databases were used for database-backed verification.

Passed:

- Focused claim-intake, authorization/provenance, WebSocket, fallback, handler, service, CLI, and fixed-ID fixture suites.
- `go test -race -json ./... -count=1` with only three independently reproduced `origin/main` failures excluded; no race markers or other failures remained.
- `go vet ./...`.
- `sqlc generate` with no generated worktree drift.
- `git diff --check` and staged-diff hygiene checks before commit.
- Fresh migrations through 278, initialization-trigger proof, and candidate-only rollback proof.

Known upstream-baseline failures independently reproduced on `origin/main`:

- `TestReusedIsolatedCheckoutRepairsPromisorConfig`: test-isolation branch collision.
- `TestDashboardFailuresByAgentUsesExactWindow`.
- `TestDashboardPerAgentRollupsUseExactWindow`.

The two dashboard fixtures use PostgreSQL session `CURRENT_DATE` while requesting a UTC API window; they fail on the baseline when the database local date is ahead of UTC after local midnight. An additional Codex inactivity timeout occurred once only under broad-suite load and passed in isolation.

### Compatibility and rollout

- Existing Workspaces are backfilled as `resumed`, generation 0.
- New Workspaces receive the same baseline state through an initialization trigger.
- Generation changes only on effective state transitions.
- Resume requires an expected generation; stale requests are durable replayable conflicts.
- Old servers use legacy singular fallback only for an exact unmatched-route `404 page not found` response.
- Migration indexes are separated into concurrent-index migrations.
- Down migrations 278 through 272 remove only this candidate's schema additions.

### Scope and residual risks

This is a non-production upstream implementation candidate only. It does not authorize ERI-167 cutover, production pause/resume, production database writes or migrations, credential operations, permission elevation, daemon restart, or any production resource change.

Independent Grok code/security review remains a required gate before ERI-204 can complete. Contract points for that review include database-level immutability/monotonicity, rejected-attempt audit scope, classification aggregate requirements, no-op/conflict `effective_at` semantics, and whether WebSocket `ClientIdentity.RuntimeIDs` is intended as a hard claim boundary.


multica-eve (changes requested):
This change would let workspace administrators temporarily stop new work from starting during maintenance or an incident, while allowing work that has already started to finish. Today they must stop broader services or coordinate manually; afterward they would also have a record of who paused or resumed work, when, and why.

## Phase A — Product review

The user journey is coherent: inspect the current state, pause with a reason, let existing work drain, resolve the incident, resume from the observed generation, and inspect the action history and work ledger. The default remains resumed, so ordinary users do not need to migrate or change their workflow.

✅ **Product verdict: accept the direction.** This addresses a real operational need with the right boundary: stop new ownership without cancelling work already in progress.

## Phase B — Implementation review

I found three must-fix issues:

1. **The migration prefix now conflicts with `main`.** This PR starts at `server/migrations/272_workspace_claim_intake_control.up.sql:1`, while current `main` already contains `272_rollup_task_usage_hourly_xact_lock`. On a synthetic merge, `go test ./internal/migrations -run TestMigrationNumericPrefixesStayUniqueAfterLegacySet -count=1` fails with `migration prefix 272 is reused`. Please rebase and renumber this PR's 272–278 sequence to 273–279, including comments and PR documentation that name those migrations.

2. **An old server replica can bypass the pause fence during a rolling deployment or rollback.** Enforcement currently exists only in the new claim queries and service locking (`server/pkg/db/queries/agent.sql:608-618`, `server/internal/service/task.go:2862-2879`). The migration only installs a workspace-initialization trigger (`server/migrations/272_workspace_claim_intake_control.up.sql:82-85`), while the pause route becomes available immediately (`server/cmd/server/router.go:1150`). A new replica can therefore acknowledge a pause while an old replica continues running the previous claim statement and creates new ownership. Please add database-level enforcement, or an enforced, default-off staged rollout gate that cannot be enabled until all claimers understand the fence and that prevents rollback to an incompatible server while any workspace is paused. This needs a mixed-version test and an explicit deploy/rollback sequence.

3. **The ledger's `stale_reclaimable` value does not match the actual reclaim predicate.** `server/pkg/db/queries/workspace_claim_intake.sql:113-118` uses a fixed three-minute fallback for a missing lease. The real path uses a 90-second recovery window, checks `dispatched_at` and lease expiry separately, and also requires `started_at IS NULL` (`server/internal/service/task.go:191-196`, `server/pkg/db/queries/agent.sql:751-757`). For example, a not-started task dispatched two minutes ago with no lease is already reclaimable but is reported as not reclaimable by the ledger. Please reuse the same parameterized predicate and add regression coverage for a null lease between 90 and 180 seconds and for an anomalous non-null `started_at`.

The authorization negative tests, row-lock ordering, exact idempotent replay, concurrent index creation, and avoidance of foreign keys/cascades are all good choices worth keeping.

Validation: `git diff --check` passed. The required `ByteIdentical` test, `go test ./internal/daemon -count=1`, and the focused claim-intake CLI tests passed on both the PR head and a synthetic merge. The synthetic merge's migration-prefix test fails as described above. The existing GitHub checks are green for the original base, but they do not cover the current `main` migration collision.

🔄 **Request changes** — the product direction is sound, but the migration collision, mixed-version fence bypass, and inaccurate operator evidence must be fixed before merge.

PR #6730, opened 2026-08-10
feat(agents): support pasting environment files

yugui923:
## What does this PR do?

Environment variables can now be pasted as dotenv-style `KEY=value` lines into the key field. Each assignment is parsed into its own editable row, so users do not need to paste names and values separately.

The paste handler recognizes assignment-like input while leaving ordinary key and comment-only pastes unchanged. Values use literal dotenv semantics: surrounding quotes and inline comments are handled, while backslashes, dollar signs, backticks, JSON, and other value content are preserved without evaluation or rewriting. Malformed assignment-like input is rejected as a whole. Imported values remain masked.

## Related Issue

None.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Parse `KEY=value` and `export KEY=value` lines into separate environment-variable rows.
- Support comments, surrounding quotes, empty values, CRLF line endings, and literal special characters without shell evaluation.
- Preserve existing rows and keyboard focus while keeping pasted values masked.
- Provide an immediately pasteable row when no variables are configured.
- Reject duplicate keys during paste and leave ordinary or comment-only pastes to native input handling.
- Show localized guidance and malformed-paste feedback in English, Japanese, Korean, and Simplified Chinese.
- Cover parser behavior, row population, focus, masking, malformed input, literal values, duplicate and special keys, and locale parity.

## Risks

- Shell-like content inside values is stored literally and is never evaluated by the parser.
- Malformed assignment-like input is rejected as a whole so partial files are not silently imported.
- Only assignment-like content is intercepted; ordinary key and comment-only pastes keep the existing input behavior.

## How to Test

1. Open an agent Environment Variables tab and select Reveal & edit.
2. Paste multiple `KEY=value` or `export KEY=value` lines into the key field.
3. Confirm each assignment appears in its own row with values masked.
4. Paste values containing backslashes, `$`, backticks, or JSON and confirm the exact literal values are preserved after save.
5. Paste malformed or duplicate assignments and confirm the row remains unchanged with an error message.
6. Paste a bare key with a trailing newline and confirm native single-key paste behavior is unchanged.

## Checklist

- [x] I have updated relevant inline guidance to reflect my changes
- [x] If this PR touches Chinese product copy, I checked it against the Chinese conventions guide

### DIFF
--- a/packages/views/agents/components/tabs/env-file.test.ts
+++ b/packages/views/agents/components/tabs/env-file.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from "vitest";
 import { isEnvFilePaste, parseEnvFile } from "./env-file";
 
 describe("parseEnvFile", () => {
-  it("parses Bash-style assignments while preserving values", () => {
+  it("parses dotenv-style assignments while preserving literal values", () => {
     expect(
       parseEnvFile(String.raw`
 # Agent credentials
@@ -12,8 +12,10 @@ EMPTY=
 URL=https://example.com/path?first=one&second=two
 HASH='value # kept'
 PLAIN=value # ignored comment
-ESCAPED=hello\ world
-ESCAPED_HASH=value\ # kept
+WIN_DIR=C:\Users\me\.ssh
+PASSWORD=p$ss${"`"}word
+JSON={"a":1}
+QUOTED='say "hello" with $HOME and C:\path'
 `),
     ).toEqual([
       { key: "API_KEY", value: "secret value" },
@@ -24,8 +26,10 @@ ESCAPED_HASH=value\ # kept
       },
       { key: "HASH", value: "value # kept" },
       { key: "PLAIN", value: "value" },
-      { key: "ESCAPED", value: "hello world" },
-      { key: "ESCAPED_HASH", value: "value # kept" },
+      { key: "WIN_DIR", value: String.raw`C:\Users\me\.ssh` },
+      { key: "PASSWORD", value: "p$ss`word" },
+      { key: "JSON", value: '{"a":1}' },
+      { key: "QUOTED", value: String.raw`say "hello" with $HOME and C:\path` },
     ]);
   });
 
@@ -40,12 +44,14 @@ ESCAPED_HASH=value\ # kept
     expect(parseEnvFile("FIRST=one\necho hello\nSECOND=two")).toBeNull();
     expect(parseEnvFile('BROKEN="unterminated')).toBeNull();
     expect(parseEnvFile('TOKEN="abc"#suffix')).toBeNull();
-    expect(parseEnvFile('EXPANDED="$HOME"')).toBeNull();
+    expect(parseEnvFile("FIRST=one\nFIRST=two")).toBeNull();
   });
 
   it("does not treat ordinary key text as an environment file", () => {
     expect(parseEnvFile("API_KEY")).toBeNull();
     expect(isEnvFilePaste("API_KEY")).toBe(false);
+    expect(isEnvFilePaste("API_KEY\n")).toBe(false);
+    expect(isEnvFilePaste("# database settings\n")).toBe(false);
     expect(isEnvFilePaste("API_KEY=value")).toBe(true);
     expect(isEnvFilePaste("FIRST=one\nSECOND=two")).toBe(true);
   });
--- a/packages/views/agents/components/tabs/env-file.test.ts
+++ b/packages/views/agents/components/tabs/env-file.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+
+import { isEnvFilePaste, parseEnvFile } from "./env-file";
+
+describe("parseEnvFile", () => {
+  it("parses Bash-style assignments while preserving values", () => {
+    expect(
+      parseEnvFile(String.raw`
+# Agent credentials
+export API_KEY="secret value"
+EMPTY=
+URL=https://example.com/path?first=one&second=two
+HASH='value # kept'
+PLAIN=value # ignored comment
+ESCAPED=hello\ world
+ESCAPED_HASH=value\ # kept
+`),
+    ).toEqual([
+      { key: "API_KEY", value: "secret value" },
+      { key: "EMPTY", value: "" },
+      {
+        key: "URL",
+        value: "https://example.com/path?first=one&second=two",
+      },
+      { key: "HASH", value: "value # kept" },
+      { key: "PLAIN", value: "value" },
+      { key: "ESCAPED", value: "hello world" },
+      { key: "ESCAPED_HASH", value: "value # kept" },
+    ]);
+  });
+
+  it("supports Windows line endings", () => {
+    expect(parseEnvFile("FIRST=one\r\nSECOND=two\r\n")).toEqual([
+      { key: "FIRST", value: "one" },
+      { key: "SECOND", value: "two" },
+    ]);
+  });
+
+  it("rejects partial files instead of silently dropping unsupported lines", () => {
+    expect(parseEnvFile("FIRST=one\necho hello\nSECOND=two")).toBeNull();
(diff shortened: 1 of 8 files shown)


Bohan-J (changes requested):
Thanks for this — it's a real pain point (copying a dozen variables one field at a time is genuinely tedious), and several of your instincts here are the right ones: rejecting a paste wholesale instead of importing half of it, keeping imported values masked, and reusing the first row's `id` so focus survives the re-render. The four-locale copy and the test coverage are appreciated too.

Two blocking issues, both traceable to a single design choice, plus some smaller notes.

## Must-fix

### 1. Pasting a bare key name with a trailing newline is now rejected

`isEnvFilePaste` (`packages/views/agents/components/tabs/env-file.ts:104`) classifies any text containing `\r` or `\n` as an environment-file paste. Copying a single variable name from a doc, a terminal, or a password manager very often carries a trailing newline, so that paste now hits the failure branch in `env-tab.tsx:134`: `preventDefault()` plus an error toast, and the field stays empty.

Verified with a DOM test against `"API_KEY\n"`:

```
nativePasteAllowed: false
keyValue:           ""
errorToast:         ["Couldn't parse pasted environment variables"]
```

Before this PR the browser stripped the newline and the field simply received `API_KEY`. This is the most common single-variable paste, so it's a regression on the existing path the PR intends to leave untouched.

Suggested fix: `trim()` before classifying, and treat a single line with no `=` as "not an env file" unconditionally — let it fall through to the native paste.

### 2. Unquoted values containing backslashes are silently rewritten

`parseUnquotedValue` (`env-file.ts:60`) consumes `\` as an escape character. Verified end-to-end, all the way to the save request body:

```
pasted: WIN_DIR=C:\Users\me\.ssh
saved:  { custom_env: { WIN_DIR: "C:Usersme.ssh" } }
```

This contradicts the guarantee in the PR description — "unsupported or malformed input is rejected as a whole so values are never silently changed". Here the input is neither rejected nor preserved, and because the value renders masked the user has no way to notice. The same applies to JSON values, regex patterns, and connection strings containing backslashes.

### Root cause for both

The parser models the input as a Bash snippet, so it performs shell unquoting. But these values never reach a shell: `layerCustomEnvAndHermesHome` in `server/internal/daemon/daemon.go:7603` copies `custom_env` entry by entry into the child process environment map (`agentEnv[k] = v`) and the process is exec'd directly. Nothing between the input box and the agent process ever interprets the value.

So the correct model is dotenv's literal-value semantics, not Bash's: strip surrounding quotes, handle an inline `#` comment and leading/trailing whitespace, and otherwise take the value verbatim — no backslash unescaping. That fixes issue 2 and also removes the reason for the should-fix below.

## Should-fix

### 3. A single `$` anywhere in a value rejects the entire paste, with an unactionable error

`env-file.ts:40` and `env-file.ts:81` return `null` for `$` and backtick. All of these are rejected wholesale:

```
PG="p$ssw0rd"     PG=p$ssw0rd     PRICE="100$"
CFG={"a":1}       NAME=it's       TOKEN=a`b
```

Generated passwords containing `$` and JSON-valued environment variables are both common. Combined with all-or-nothing rejection, one `$` in a 30-line file kills the whole import — and `paste_invalid_toast` says only "Couldn't parse pasted environment variables", with no line number and no reason, so the user has nothing to act on. Single-quoting does get through, but nobody will guess that.

Under literal-value semantics `$` no longer needs rejecting. If you do keep some rejection rules, please put the offending line number and the reason into the message.

## Nits

- When no variables are configured yet, the paste hint renders but there is no key field to paste into — the empty state at `env-tab.tsx:270` renders only text. Verified: `hintShown: true, keyFieldsPresent: 0`. Consider rendering an empty row, or rewording the hint to mention adding a row first.
- A comment-only or blank-lines-only paste (e.g. `"# db\n"`) errors out: `parseEnvFile` returns `null` because `assignments.length === 0`, while `isEnvFilePaste` returns `true`. "Nothing to import" would be better handled by falling through silently.
- Duplicate keys within one paste (`A=1\nA=2`) expand into two rows and only trip the existing duplicate check at save time. Flagging at paste time would fail faster.
- Changing `entriesToEnvMap` to `Object.create(null)` (`env-tab.tsx:39`) is a genuine fix — a key named `__proto__` was previously dropped silently. But that bug is independent of pasting (typing `__proto__` by hand hits it too), so it would be cleaner as its own commit or PR. Not blocking.
- The missing test cases are exactly the three issues above: a bare key with a trailing newline, an unquoted value with backslashes, and a value containing `$`.

## Verification

Against `5de5edf`:

- `pnpm --filter @multica/views typecheck` — passes
- `pnpm --filter @multica/views lint` — 0 errors (21 warnings, all in files this PR doesn't touch)
- `pnpm --filter @multica/views test` — 320 files / 3784 tests green, including the new `env-file.test.ts`, `env-tab.test.tsx`, and the i18n parity suites
- Merges cleanly with `main` (4 commits behind)

The behavioral findings above come from temporary tests run against this branch, not from reading the code alone.

The direction is good and issues 1–3 collapse into one change — switching to literal-value semantics. Happy to re-review once that's in.

yugui923:
@Bohan-J Thanks for the detailed review — all blocking issues and actionable behavior feedback are fixed in `9fb0a38`.

- Bare keys with trailing newlines, and comment-only pastes, now fall through to native paste handling.
- The parser now uses literal dotenv semantics: backslashes, `$`, backticks, JSON, and quoted content are preserved rather than shell-unescaped or rejected.
- Empty environments immediately render a pasteable row.
- Duplicate keys are rejected during paste with the existing duplicate-key message.
- Regression coverage now includes the trailing-newline case, literal backslashes and `$`/backticks, comment-only input, duplicate keys, the empty state, and locale parity.

I also rebased the branch onto current `main`, updated the PR description to match the corrected semantics, and re-ran verification at the current head: Prettier passed, typecheck passed, lint has 0 errors (the same 21 unrelated warnings), and all 323 `@multica/views` test files / 3,808 tests passed. CI is green as well.

Bohan-J (approved):
Thank you for the thorough follow-up — this is a genuinely better change than the one I reviewed. Switching to literal dotenv semantics didn't just patch the two blockers, it made the parser noticeably simpler: `parseUnquotedValue` lost the escape state machine entirely and now returns a slice, and the whole class of "which characters do we reject" questions disappeared with it. That's the right kind of fix.

Re-verified both blockers on `9fb0a38`, end to end from the paste event through to the save request body:

- **Bare key with a trailing newline** — `"API_KEY\n"` now falls through to the native paste, no `preventDefault`, no toast. `isEnvFilePaste` only claims a paste when a meaningful line actually looks like an assignment, so comment-only content passes through too.
- **Values are preserved byte-for-byte** — `C:\Users\me\.ssh`, `p$ssw0rd`, `{"a":1}`, `it's`, and a backtick-containing value all arrive in `custom_env` exactly as pasted. This was the one that worried me most, since masked values made the corruption invisible.

The smaller points are handled as well: the empty state now renders a pasteable row, and duplicate keys within one paste are rejected before expansion with their own message rather than surfacing at save time. Imported values are still masked, and the error toast doesn't echo pasted content.

One non-blocking nit, purely for a future pass: an unterminated quote or a stray non-assignment line still produces the generic "Couldn't parse pasted environment variables", with no line number. Now that duplicates get a specific message, the remaining generic case stands out a bit more — on a 30-line file it's still guesswork. Not worth holding this up for.

Verification on my side, against `9fb0a38`:

- `pnpm --filter @multica/views typecheck` — passes
- `pnpm --filter @multica/views lint` — 0 errors (21 warnings, all in files this PR doesn't touch)
- `pnpm --filter @multica/views test` — 323 files / 3808 tests, fully green
- CI green across all 9 checks

The updated test cases pin the new semantics directly rather than being relaxed to fit — `WIN_DIR`, `PASSWORD` with `$` and a backtick, `JSON`, and the `isEnvFilePaste("API_KEY\n") === false` case are all asserted explicitly. Appreciated.

Approving and merging. Thanks for contributing this, and for the careful second pass. 🎉

PR #6737, opened 2026-08-11
MUL-6010: fix(desktop): structure route error feedback context

Ryannn41:
## What does this PR do?

This PR replaces the desktop route error page’s Markdown-encoded diagnostic report with structured feedback context.

Previously, the error type, trigger, message, and stack trace were embedded in the editable feedback message, making them difficult to query and unreliable to parse. The feedback pipeline already stores metadata as JSONB, so this change carries route error diagnostics as a typed `context` object and stores them in feedback metadata.

### Thinking path

1. Traced the data flow from the desktop error page through the feedback modal, API client, server handler, and database.
2. Confirmed that the existing JSONB metadata column can store the context without a migration.
3. Added a typed context contract and runtime validation.
4. Kept the user-editable message separate from diagnostic data.
5. Added tests for every affected layer.

The field is optional, so existing feedback clients remain compatible. The server should be deployed before the updated desktop client because older servers will ignore the new context field.

## Related Issue

No public issue. This addresses the explicit TODO in:

`apps/desktop/src/renderer/src/components/route-error-page.tsx`

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added typed desktop route error context and validation in `packages/core/feedback/`.
- Extended the feedback mutation and API client with optional structured context.
- Updated `packages/views/modals/feedback.tsx` to submit diagnostic context separately from the editable message.
- Removed the Markdown diagnostic template from `apps/desktop/src/renderer/src/components/route-error-page.tsx`.
- Updated `server/internal/handler/feedback.go` to store context in the existing metadata JSONB column.
- Added and updated desktop, shared view, core, and Go tests.

## How to Test

1. Run the related tests:

   ```bash
   pnpm --dir apps/desktop exec vitest run src/renderer/src/components/route-error-page.test.tsx
   pnpm --dir packages/views exec vitest run modals/feedback.test.tsx
   pnpm --dir packages/core exec vitest run api/client.test.ts feedback/types.test.ts
   GOMAXPROCS=1 go -C server test -p=1 ./internal/handler -run '^TestCreateFeedback' -count=1
   ```

2. Run type and static checks:

   ```bash
   pnpm typecheck
   GOMAXPROCS=1 go -C server vet ./internal/handler
   ```

3. Trigger a non-404 desktop route error, click **Report error**, and verify that the submitted feedback metadata contains `context.kind`, `context.trigger`, and the structured error details.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (not applicable)

## AI Disclosure

**AI tool used:** Cursor (GPT-5.6 Sol)

**Prompt / approach:**

I asked Cursor to implement the structured-feedback TODO from the latest upstream `main`. The approach was to trace the complete feedback flow, define a small backward-compatible request contract, separate diagnostic context from user-authored text, persist it in the existing metadata column, and verify every affected layer with focused tests and static checks.

## Screenshots (optional)

Not included because this change does not alter the error page or feedback modal layout.

### DIFF
--- a/apps/desktop/src/renderer/src/components/route-error-page.test.tsx
+++ b/apps/desktop/src/renderer/src/components/route-error-page.test.tsx
@@ -29,7 +29,7 @@ vi.mock("@/stores/tab-store", () => {
   return { useTabStore };
 });
 
-import { DesktopRouteErrorPage, formatRouteErrorReport } from "./route-error-page";
+import { createRouteErrorFeedback, DesktopRouteErrorPage } from "./route-error-page";
 
 function Boom(): null {
   throw new Error("route render exploded");
@@ -96,7 +96,7 @@ describe("DesktopRouteErrorPage", () => {
     expect(closeActiveTab).toHaveBeenCalledTimes(1);
   });
 
-  it("opens the existing feedback modal with a structured markdown report only after click", async () => {
+  it("opens the feedback modal with structured error context only after click", async () => {
     const router = createMemoryRouter(
       [{ path: "/acme/issues", element: <Boom />, errorElement: <DesktopRouteErrorPage /> }],
       { initialEntries: ["/acme/issues"] },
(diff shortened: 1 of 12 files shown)


multica-eve (changes requested):
The product direction makes sense: route diagnostics should be queryable without forcing members to edit a Markdown diagnostic dump. I’m requesting changes for two blockers in the current implementation:

1. `apps/desktop/src/renderer/src/components/route-error-page.tsx:21` still seeds `normalized.message` into `initialMessage` while also submitting the same value in `context.error.message`. The system-generated diagnostic therefore remains mixed into the member-editable message, contrary to the PR’s stated separation. It also creates a misleading control boundary: deleting the visible error text does not remove the hidden copy. Please keep the editable field user-authored only, or explicitly expose the additional diagnostic submission as the intended product behavior, and update the route error tests accordingly.

2. `server/internal/handler/feedback.go:40-44,73,110-112` persists any decoded `context` without semantic validation. For example, `{"context":{"kind":"desktop_route_error"}}` is accepted with empty `trigger` and error fields, and arbitrary `kind` values are stored as supposedly structured diagnostics. The in-process TypeScript guard does not protect the API boundary. Please validate the supported discriminator and required non-empty `trigger`, `error.name`, and `error.message` fields on the server, with negative tests for malformed, unknown, and empty contexts.

The focused frontend/backend tests, typecheck, lint, Go vet, and current CI checks pass, but they do not cover these two correctness gaps.

Ryannn41:
Addressed both blockers:

- The route error page no longer seeds the system-generated error message into `initialMessage`. The editor now contains only member-authored text, while diagnostics remain in the structured context.
- The feedback handler now validates the supported context discriminator and requires non-empty `trigger`, `error.name`, and `error.message` values.
- Added negative API tests for malformed, unknown, empty, and whitespace-only contexts.

I also reran the focused frontend/backend tests, TypeScript typecheck, ESLint, and Go vet.

PR #6740, opened 2026-08-11
MUL-6011 feat(layout): open a workspace in a new window from the switcher

HomyeeKing:
## What

Adds an "open in new window" affordance to each workspace row in the sidebar workspace switcher, so users can work on multiple workspaces in parallel windows (HOM-10).

- Each workspace row in the switcher dropdown gets an external-link icon (revealed on row hover / focus).
- Tooltip and `aria-label` read **"Open in new window"** (localized in en / zh-Hans / ja / ko).
- Clicking it opens that workspace's shareable URL via the existing cross-platform `openExternal` helper:
  - **Desktop** → hands the URL to the OS shell (a new native window).
  - **Web** → opens a new browser window/tab.
- The row itself still navigates in place; the button `preventDefault`/`stopPropagation`s so the two gestures never collide.

## Why

Requested in HOM-10 — being able to open a workspace in a separate window enables handling tasks in multiple windows at once. The described approach (an external-jump icon in the workspace dropdown with an "open in new window" tooltip) is implemented as specified.

## Testing

- `pnpm --filter @multica/views test` — 3774 passed (added 2 tests covering the new control: one per-row control exists, and clicking opens the correct shareable URL without navigating).
- `pnpm --filter @multica/views typecheck` — clean.
- `pnpm --filter @multica/views lint` — no new warnings/errors.

### DIFF
--- a/packages/views/layout/app-sidebar.test.tsx
+++ b/packages/views/layout/app-sidebar.test.tsx
@@ -1,9 +1,9 @@
-import { render, screen, waitFor } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { ApiError } from "@multica/core/api";
 import { AppSidebar } from "./app-sidebar";
 
-const { appForeground, chatSessions, chatStore, detail, deletePin, inboxItems, navigation, pins, sidebarState, summary, workspaces } = vi.hoisted(() => ({
+const { appForeground, chatSessions, chatStore, detail, deletePin, inboxItems, navigation, openExternal, pins, sidebarState, summary, workspaces } = vi.hoisted(() => ({
   appForeground: { current: true },
   sidebarState: { setOpenMobile: vi.fn() },
   chatSessions: { current: [] as { id?: string; unread_count?: number }[] },
@@ -12,6 +12,7 @@ const { appForeground, chatSessions, chatStore, detail, deletePin, inboxItems, n
   deletePin: vi.fn(),
   inboxItems: { current: [] as { id: string; read: boolean }[] },
   navigation: { current: { pathname: "/acme/issues" } },
+  openExternal: vi.fn(),
   summary: { current: [] as { workspace_id: string; count: number }[] },
   workspaces: {
     current: [] as { id: string; name: string; slug: string; avatar_url: string | null }[],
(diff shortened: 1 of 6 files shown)


Bohan-J:
Thanks for this — and for making it your first contribution here. The scoping is tight, all four locales are covered, the new behavior comes with tests, and CI is green. That's a better-than-average first PR by a good margin.

We're going to close it anyway, and I want to be clear that the reason is product-level rather than code quality — so please don't read this as "address the review and it lands."

**1. On desktop, this doesn't do what the title says.**

`openExternal` is documented as "open a URL in the user's default browser, regardless of platform" (`packages/views/platform/open-external.ts`). On desktop it routes to `shell.openExternal` in the main process, which hands the URL to the OS — it never creates a Multica window. Two consequences follow: the user leaves the app entirely, and since desktop auth lives in the renderer's `localStorage` (`multica_token`) while the web app uses HttpOnly cookies, they will usually land on a login page rather than the workspace.

There's also a direction conflict. Sending a destination that points back at this deployment out to the system browser is the exact behavior we moved away from — see the note at the top of `packages/views/rich-content/app-url-link-routing.test.tsx`.

**2. On web, most of this already works.**

The workspace row renders a real `<a>`, so `cmd`/`ctrl`+click already opens a new tab in every browser, and `shift`+click already opens a new window in Chrome and Firefox — `packages/views/navigation/click-intent.ts` documents that we deliberately leave those to native handling. What the button genuinely adds on web is discoverability, plus Safari's missing shift+click window. Real, but not enough to justify a control that occupies space in every workspace row and introduces a second label for a concept `common.navigation.open_in_new_tab` already covers.

**3. Doing this properly is a capability decision we want to own.**

Desktop is where the underlying need actually lives, and serving it means multi-window support: main-process window management, per-window workspace context, splitting the tab store per window, plus lifecycle and menu/shortcut wiring. Generalizing the existing `openIssueWindow` (today a single-route detached window) is the likely starting point. We'd rather design that ourselves than lock in the entry point's shape ahead of it — which is what merging this would do, and it would mean changing this button's behavior a second time later.

For feature-level work like this, opening an issue first is the fastest path: it lets us align on direction before you spend the implementation time. We don't have a corresponding issue on our side for this one, which is part of why it's landing on a "not now."

None of the above is wasted effort on your part — the review-level issues you'd have hit (nesting a `<button>` inside the row's anchor, and the hover-reveal conventions in `app-sidebar.tsx`) are things you clearly already reason about. If you'd like to contribute again, we're happy to point you at something with a clearer runway.

PR #6746, opened 2026-08-11
MUL-5992 fix(daemon): deny the Reasonix ask tool per task (supersedes #6745)

Bohan-J:
Supersedes #6745 (MUL-5992), which this PR reverts in its first commit. Same goal — an unattended Reasonix task must not die on an `ask` — but actually turning the tool off instead of asking the model not to reach for it.

## Why the swap

#6745 appended a "no interactive user, don't call `ask`" line to the per-turn prompt. That was the best host-side lever I had found: `ask` is registered unconditionally, and a `PreToolUse` hook would only ever reach the executor, because planner tool calls fire no hooks at all.

Reasonix has a better one. `[permissions] deny = ["ask"]` genuinely disables the tool, for the whole turn:

- `Policy.DecideSubject` checks deny rules **before** the read-only allow fallthrough, so the rule applies to `ask` even though `ask` is read-only.
- The executor, the planner (which keeps `ask` deliberately, and which we run with `--planner auto`), and sub-agents all decide against the same policy — no half-covered chain.
- The model gets `blocked: denied by permission policy — this tool/command is on the deny list. Do not retry it; choose another approach or stop and explain.` back as an ordinary tool result, and the turn continues.

And it doesn't have to be global. Reasonix resolves config as flag > project `./reasonix.toml` > user `config.toml` > defaults, rooted at the ACP session cwd — the task workdir. The project merge restores CLI/Secrets/Remote/Desktop/Telemetry from the user's global config, but not `Permissions`, so a project file's rules are honoured while staying scoped to this task.

## Change

`Prepare` and `Reuse` write `{WorkDir}/reasonix.toml` for reasonix tasks through the sidecar manifest, so `CleanupSidecars` takes it back out. `Reuse` rewrites it, since cleanup runs first — otherwise the second turn of a task would run with the tool available again.

A `reasonix.toml` that came with the repository is left byte-for-byte alone and logged — the manifest contract is that cleanup is a pure deletion of paths we created, so we never overwrite user content. Those tasks keep today's fail-closed question handling.

No git-exclude entry, deliberately. The agent's cwd is `env.WorkDir` while a managed checkout is a directory *inside* it, so `agentGitExcludePatterns` — which only reaches the checkout's `.git/info/exclude` — could never hide this sidecar; all it would do is hide a `reasonix.toml` an agent legitimately creates inside a repository, for every provider. A regression test pins that layering. For a repo task the sidecar sits outside the checkout entirely; for a local_directory task it is untracked for the duration of the run and removed at teardown, which is how `AGENTS.md` and the other WorkDir sidecars already behave.

Deliberately unchanged: the fail-closed question path in `selectReasonixPermissionOption` stays as the backstop, and the planner stays on.

## Trade-off, accepted

A project `deny` list replaces rather than extends the user's global one, so for the duration of a task in this workdir a runtime owner's global deny rules don't apply. That is a deliberate widening of what the agent may run in a task workdir, accepted by the maintainer — not something the daemon's other constraints make equivalent. Extending the global list instead would mean parsing the runtime owner's config to restate it, which is its own failure mode.

## Upstream

The real fix is a session capability so Reasonix skips `SetAsker` (or drops `ask`) when the client declares it has no interactive user — filed at [esengine/DeepSeek-Reasonix#8238](https://github.com/esengine/DeepSeek-Reasonix/issues/8238), including why a fix must cover both registries. Until that lands, local_directory tasks whose repository ships its own `reasonix.toml` stay on the fail-closed path.

## Verification

`go test ./internal/daemon/execenv/ ./internal/daemon/repocache/ ./pkg/agent/ -count=1` — all pass.

New tests:

- `Prepare` writes the deny rule and `CleanupSidecars` removes it again.
- `Reuse` re-lays it after the rollback that precedes it.
- A repository-provided `reasonix.toml` is neither rewritten nor recorded in the manifest.
- Non-reasonix providers get no file.
- `TestCreateWorktreeDoesNotExcludeReasonixProjectConfig` — the checkout is a child of the WorkDir, `.git/info/exclude` carries no `reasonix.toml`, and `git check-ignore` does not ignore one written inside the checkout. Verified to fail when the pattern is put back.

### DIFF
--- a/server/internal/daemon/execenv/execenv.go
+++ b/server/internal/daemon/execenv/execenv.go
@@ -404,6 +404,15 @@ func Prepare(params PrepareParams, logger *slog.Logger) (*Environment, error) {
 		env.QwenpawWorkspace = qwenpawWorkspace
 	}
 
+	// For Reasonix, deny the `ask` tool for this task through a project-scoped
+	// reasonix.toml. Degraded, not fatal: without it the task still runs under
+	// the backend's fail-closed question handling.
+	if params.Provider == "reasonix" {
+		if err := writeReasonixProjectConfig(workDir, manifest, logger); err != nil {
+			logger.Warn("execenv: write reasonix project config failed", "error", err)
+		}
+	}
+
 	// For Cursor, materialize managed MCP into project-local config and use
 	// an isolated CURSOR_DATA_DIR for the per-workdir approval sidecar. Cursor
 	// still reads ~/.cursor/mcp.json, but only servers with approval entries in
@@ -618,6 +627,15 @@ func Reuse(params ReuseParams, logger *slog.Logger) *Environment {
 			logger.Warn("execenv: refresh claude skill settings failed", "error", err)
 		} else {
 			env.ClaudeSettingsPath = settingsPath
+		}
+	}
+
+	// Re-deny Reasonix's `ask` tool on reuse: CleanupSidecars above removed the
+	// prior run's reasonix.toml, so without this the next turn would run with
+	// the tool available again.
+	if params.Provider == "reasonix" {
+		if err := writeReasonixProjectConfig(params.WorkDir, manifest, logger); err != nil {
+			logger.Warn("execenv: refresh reasonix project config failed", "error", err)
 		}
 	}
 
--- a/server/internal/daemon/execenv/reasonix_permissions.go
+++ b/server/internal/daemon/execenv/reasonix_permissions.go
@@ -1,0 +1,73 @@
+package execenv
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"path/filepath"
+)
+
+// reasonixProjectConfigFile is Reasonix's project-scoped config. Reasonix
+// resolves configuration as flag > project ./reasonix.toml > user config.toml >
+// defaults, rooted at the session cwd — which for a task is the workdir — so a
+// file written here binds to this task only and leaves the runtime owner's
+// interactive sessions alone.
+const reasonixProjectConfigFile = "reasonix.toml"
+
+// reasonixProjectConfig turns off Reasonix's `ask` tool for the task.
+//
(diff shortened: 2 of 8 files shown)


Bohan-J:
Good catch on the exclude — verified both layering facts before changing anything: the agent's cwd is `env.WorkDir` (`daemon.go`, `Cwd: env.WorkDir`) and a managed checkout is `filepath.Join(params.WorkDir, dirName)`, so `agentGitExcludePatterns` only ever reaches a directory *below* the sidecar. It could not hide the daemon's file, and it did hide a `reasonix.toml` an agent creates inside a repo, for every provider.

Dropped it in 2d2d288, plus a regression test that pins the layering: `TestCreateWorktreeDoesNotExcludeReasonixProjectConfig` asserts the checkout is a child of the WorkDir, that `.git/info/exclude` carries no `reasonix.toml`, and that `git check-ignore` does not ignore a `reasonix.toml` written inside the checkout. I confirmed the guard has teeth by re-adding the pattern locally — the test fails.

No replacement exclude: for a repo task the sidecar is outside the checkout, and for a local_directory task it is untracked for the run and removed by `CleanupSidecars`, which is exactly how `AGENTS.md` and the other WorkDir sidecars already behave. If we ever want it hidden during a local_directory run, that belongs in the local_directory path with its own test, as you said.

Also took the non-blocking comment point: the deny-list note no longer implies `--workspace-only` and the ACP permission policy are equivalent protection to the user's global deny rules. It now says plainly that this is an accepted widening, and why extending the global list (parsing the runtime owner's config to restate it) is its own failure mode.

Left alone: the pre-existing-file degrade, which keeps local_directory + a repo-provided `reasonix.toml` on the old fail-closed path. Agreed that the real close-out is the upstream session capability — tracked at esengine/DeepSeek-Reasonix#8238.

`go test ./internal/daemon/repocache/ ./internal/daemon/execenv/ -count=1` passes.

PR #6747, opened 2026-08-11
MUL-6016: feat(task): retry exact Codex capacity failures

cjsxuexi:
## What does this PR do?

- Recognizes the exact Codex capacity error `Selected model is at capacity. Please try a different model.`.
- Immediately retries only that error, with six additional retries by default and no backoff.
- Exposes `MULTICA_CODEX_CAPACITY_RETRY_COUNT` through server configuration, Docker Compose, Helm, `.env.example`, and localized documentation.

Matching is byte-for-byte exact. Whitespace, case changes, suffixes, generic 429s, and other capacity or rate-limit errors remain terminal. The default value `6` allows seven total executions, while `0` disables this dedicated policy.

Capacity retries remain separate from generic retryable failure reasons. Autopilot and unlinked tasks remain excluded, and retry children preserve resume-safe session/workdir state while being queued immediately without `fire_at`.

COM-44 has 43 recorded runs; 15 contain the exact capacity error. The latest inspected failure persisted the legacy reason `agent_error.model_not_found_or_unavailable` and required a manual rerun.

## Related Issue

COM-44

## Type of Change

- [x] Documentation update
- [x] CI / infrastructure

## Changes Made

- Added strict Codex capacity classification and stale-daemon reason normalization in `server/pkg/taskfailure`.
- Added an immediate, separately counted capacity retry path in `server/internal/service`.
- Added `MULTICA_CODEX_CAPACITY_RETRY_COUNT` parsing and wiring in `server/cmd/server`.
- Added Docker Compose and Helm configuration plus English, Chinese, Japanese, and Korean environment-variable documentation.
- Added design and TDD implementation-plan documents.

## How to Test

1. Run `go test ./pkg/taskfailure ./cmd/server -count=1` from `server/`.
2. Run `go test ./internal/service -run 'Capacity|ProviderNetwork' -count=1` from `server/`.
3. Run `pnpm --filter @multica/docs typecheck`.
4. In WSL, render Docker Compose with the default and an override, then lint/render Helm with the default and an override.
5. Run `scripts/selfhost-config.test.sh` and `scripts/helm-config.test.sh`.

A full Windows `go test ./...` run remains blocked by existing Windows-specific baseline failures involving CRLF-sensitive built-in skill tests, temporary-directory/permission assertions, ambient agent CLI discovery, and a CLI test timeout. The changed packages and focused regression suites pass.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; no UI changes)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable; no runtime, tool, or UI tab was added)
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (not applicable; no product UI copy changed)

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Implemented COM-44 with TDD: first added exact-match classifier tests, then the dedicated retry-policy tests, server configuration tests, deployment manifest checks, and localized documentation. Verified focused Go suites, docs typechecking, Docker Compose rendering, Helm rendering/linting, and self-host configuration scripts.

## Screenshots (optional)

Not applicable; this change has no UI.


multica-eve (changes requested):
This change would spare members from manually rerunning tasks when the selected model is temporarily full: the system would automatically try again for this specific failure and show the final failure only after the allowed attempts are exhausted. The affected journey is clear, requires no user migration, and leaves other failures unchanged.

**Product verdict: ✅ Accept the product direction.** This addresses a real, recurring reliability problem and keeps the behavior narrowly scoped.

## Must-fix

Please add a sane operational hard limit for `MULTICA_CODEX_CAPACITY_RETRY_COUNT`. `server/cmd/server/router.go:154-170` currently accepts `2147483646`, and `server/internal/service/task.go:4178-4204` turns that into as many as `2147483647` total executions, all queued without delay. With a persistent capacity failure, one typo in deployment configuration could therefore create an effectively unbounded task chain, continually writing task rows, launching runtimes, and calling the upstream provider while delaying the user-visible terminal failure for an impractical amount of time.

Please choose a product-approved maximum, reject startup or fall back safely above it, and add boundary tests for that maximum. The default of six can remain unchanged.

## Nit

`server/internal/service/task.go:4184-4186` and `server/pkg/db/queries/agent.sql:490-495` still state that `max_attempts <= 1` always disables retries and is never overridden. This PR intentionally makes the dedicated capacity policy override that contract. Please update those comments and regenerate the generated query code so future maintenance follows the actual behavior.

The remaining implementation is well scoped: exact matching does not broaden this to generic 429/rate-limit failures, legacy daemon reasons are normalized server-side, autopilot and unlinked tasks remain excluded, and retry creation stays atomic with the failure transition.

**Verdict: Request changes** — the product behavior is desirable, but the retry configuration needs a meaningful production-safety bound before merge.

multica-eve (changes requested):
Thanks for updating the misleading retry-policy comments. I re-reviewed the latest head, `9875cc078`; the previous nit is resolved, but the request-changes verdict remains for two blockers.

## Must-fix

1. The capacity retry count still has no meaningful operational hard limit. `server/cmd/server/router.go:154-170` continues to accept `2147483646` additional retries, and `server/cmd/server/codex_capacity_retry_test.go:17` explicitly treats that value as valid. `server/internal/service/task.go:4286-4289` turns it into up to `2147483647` total executions, while `server/internal/service/task.go:4299-4307` keeps every capacity retry immediate. A configuration typo or persistent capacity error can therefore still create a practically unbounded task chain. Please introduce a product-approved hard maximum, reject startup or fall back safely above it, and add lower/upper boundary tests.

2. The current merge ref does not compile. `TaskService.FailTask` on the latest base now takes a `branchName` argument, but the new call in `server/internal/service/retry_deferred_test.go:264` still uses the old signature. Both backend CI and the targeted local test fail with `not enough arguments in call to svc.FailTask`. Please sync with the latest base, update the call, and restore a green backend check.

Validation: `git diff --check` passes, and `go test ./cmd/server -run '^TestCodexCapacityRetryCountFromEnv$' -count=1` passes. `go test ./internal/service -run '^TestFailTaskCodexCapacityBudget$' -count=1` fails at compile time due to the signature mismatch above.

**Verdict: Request changes** — the product direction remains accepted, but the retry safety bound and latest-base build failure must be fixed before merge.

multica-eve (changes requested):
Thanks for addressing the two previous blockers. The retry budget now has a hard maximum of 20 with fail-at-startup validation and boundary coverage, and the `FailTask` test call now matches the current signature. The targeted local tests and all checks on `52cd8c9cf` pass.

## Must-fix

This PR currently conflicts with the latest `main`, so there is no merge ref that validates the combined result. The content conflict is in `server/internal/handler/handler.go` around `TaskService` initialization: current `main` adds

```go
taskSvc.SourceContextStorage = store
```

while this branch adds

```go
taskSvc.CodexCapacityRetryCount = cfg.CodexCapacityRetryCount
```

Please sync the latest `main`, resolve the conflict by retaining both assignments, and rerun CI on the resulting head. Dropping either side would regress source-context delivery or silently disable the new capacity-retry configuration.

## Nit

`server/internal/handler/handler.go:147-148` says cmd/server "clamps" the value, but the implementation rejects an out-of-range value and stops startup. Please change the wording to "validates" or "enforces the maximum" while resolving this file.

Validation on the current branch head: `git diff --check` passes; the capacity configuration boundary tests, `TestCodexCapacityFailureRetries`, and the `pkg/taskfailure` test suite pass. GitHub checks are green for this head, but they do not cover the unresolved merge with current `main`.

**Verdict: Request changes** — the previous functional blockers are resolved, but the base conflict must be resolved and verified on a new head before merge.

multica-eve:
Update: this change has been reverted by #7586 (commit `3b1018e9c9ed989004384a480161d132bdb06e46`).

After merging, we revisited the product behavior and concluded that the current retry policy is not reasonable or safe enough to keep:

- It performs up to six additional retries against the same model with no sleep, backoff, or jitter—up to seven executions in total. The attempts are serial rather than concurrent, but each successor is queued immediately.
- Capacity failures are usually temporally correlated, so retrying back-to-back is likely to hit the same condition while amplifying provider load and consuming queue/runtime resources.
- The default of six immediate retries is not supported by recovery-rate data, and the implementation neither honors `Retry-After` nor falls back to another model.
- The policy is gated by byte-for-byte matching of one raw English error message instead of a structured overload signal, making the behavior brittle and inconsistent with canonical failure classification.

The underlying goal—recovering automatically from transient provider overload—is still valid, but it needs a different design before being reintroduced: use a structured transient-overload signal, honor `Retry-After` when available, otherwise apply bounded exponential backoff with jitter and an overall retry deadline; quota, authentication, and configuration failures should fail immediately.

Sorry for the merge/revert churn, and thank you for the contribution. The revert is about the retry policy, not the value of the problem you identified.

PR #6750, opened 2026-08-11
MUL-6014 feat(channel): a durable outbound queue for hold-connection channels

wzxjohn:
Add channel queue from #6581


multica-eve (changes requested):
Before this change, a WeCom user could receive no answer after a restart, a connection handoff, or when the work completed elsewhere. After it, the answer is retained and delivered back to the original conversation without requiring the user to resend the question.

✅ **Product decision: accept the direction.** This closes a real reliability gap and is the queue-only split requested from #6581.

I’m requesting changes for two must-fix items:

1. **A mixed-version rolling deployment can resend an answer that the old version already delivered.** The new version starts the reconciler unconditionally at `server/cmd/server/router.go:772`, while `server/pkg/db/queries/channel_outbound_queue.sql:386` treats the absence of a queue row as evidence that the answer was never delivered. The old version sends directly and leaves no queue row. Therefore, once a new replica has created the cursor, an answer successfully sent by a still-running old replica becomes a reconcile candidate and is delivered again when a new replica owns the connection. The `created_at` floor only protects history from before the first new replica; it does not protect the old/new overlap. Please add an explicit cutover or a compatible delivery tombstone phase, document rollback behavior, and cover this mixed-version sequence with a regression test.

2. **The migration prefixes currently collide with `main`, and backend CI is failing.** The new 265–270 migrations reuse prefixes already occupied by the saved-view migrations. Please perform the agreed final renumber before merge. Move 265–270, 279, and 280 together as one dependency-ordered block; renumbering only the six collisions can place the two dependent indexes before the queue table they require.

The queue design and its focused tests are otherwise strong. Once these two blockers are resolved, the product and implementation direction are suitable to merge.

wzxjohn:
@multica-eve Two answers, one on the rolling-upgrade hazard and one on the migration prefixes.

## 1. The mixed-version resend: documented, not gated

You're right about the mechanism, and I traced it to be sure: the reconciler starts
unconditionally (`router.go:820`, no `EnqueueOK` in the config literal), the candidate
scan's only delivered-test is `NOT EXISTS`, and the window's floor is
`max(cursor_at − 5min, now − 24h, created_at)`. The `created_at` floor stops the scan
reaching back before the cursor existed, and the overlap happens strictly *after* that
instant — inside the ordinary five-minute window. So it is not covered, exactly as you
describe.

Worth naming why we both missed it: the general rule was already written down twice in
this PR — your *"absence of a row means 'not delivered' only while every delivery is an
enqueue"* and my `outbound.go` note that *"a path that bypasses the queue owes the queue
a record of the delivery."* Neither of us counted **the previously shipped release** as
one of those bypassing paths, and it is the one we cannot modify.

**We're handling it with documentation rather than code, and here is the reasoning rather
than just the decision.**

The exposure is only the replies that complete between the first new replica starting and
the last old replica stopping. It is bounded by how busy the bot is during the rollover,
not by how long the queue has existed, and it is a one-time event per upgrade. Two things
are already outside it: everything from before the upgrade, and any rollover where old and
new never run at once — which includes this guide's Docker Compose path, since
`docker compose up -d` recreates rather than running both.

Against that, the code you'd need is not small, and the shape is awkward:

- **A tombstone phase cannot work on its own.** The tombstone has to be written by whoever
  delivered, and the old release is already in the field. It would take two releases: one
  that writes tombstones without consuming the queue, then one that enables the reconciler.
- **A cutover switch is a permanent config burden for a transient problem.** Default-off
  means every operator must flip something to get the safety net, and most never will — a
  safety net nobody enables is not one. Default-on doesn't help the upgrade it exists for.
- **A regression test cannot be written without first adding the discriminator.** In-tree,
  "a terminal task with no queue row after the cursor exists" *is* the reconciler's
  intended rescue path — the case it exists for. Asserting "do not enqueue" would
  contradict its purpose unless something distinguishes "an old replica already sent this"
  from "the producing replica died". Nothing in the schema does. So the test you asked for
  is downstream of the gate, not independent of it.

So `SELF_HOSTING.md` → **Updating** now carries it: what happens and why the two
observations are indistinguishable, that the exposure is only the overlap, and two
concrete outs — roll over without overlap (scale to zero, upgrade, scale back), which
avoids it entirely, or pick a low-traffic window if the update has to be rolling, since an
idle bot has no in-flight replies to duplicate.

It also covers the rollback behaviour you asked for, including a second-order effect we
found while writing it: the old release doesn't know `channel_outbound_queue`, so anything
queued sits undrained and is delivered when you roll forward — possibly hours later, as
there is no maximum row age yet. The note gives the predicate to clear those rows and says
what clearing costs.

**Residual risk, stated plainly:** an old replica lingering past the operator's quiet
window is still exposed, and no amount of documentation fixes that. The sound answer is
the positive delivery record from your own addendum — and it can't be retrofitted into the
release already deployed. If you'd rather we take the gate now, `EnqueueOK` already exists
for exactly that shape and is the one field not wired; there's also a zero-config variant
(floor the window at `created_at + grace`, self-expiring, no operator action) if a
permanent switch is the part you object to. Happy to do either — this is a judgement about
cost, not a disagreement about the defect.

## 2. Migration prefixes

These are at 265–280 because that's what was asked for.

multica-eve:
Thanks for spelling out the rolling-upgrade trade-off. There is one more implementation direction we would like you to evaluate: for the cross-replica handoff itself, please prefer Redis over PostgreSQL queue tables.

This repository already uses `REDIS_URL` and Redis Streams for multi-node realtime fanout (`server/internal/realtime/*`), so the outbound path can follow the same operational model. The intended shape would be:

- Use a dedicated Redis Stream (not Pub/Sub) for outbound envelopes, partitioned/routed by installation. Pub/Sub is not sufficient because it loses messages while no consumer is connected.
- `XADD` each reply with a stable business id such as `(channel_type, installation_id, source_kind, source_id)`. Make dedupe + append atomic, for example with a small Lua script using a bounded-lifetime dedupe key.
- Only the replica holding the WeCom connection lease consumes that installation's consumer group. `XACK` only after WeCom confirms the send; reclaim stale pending entries after lease handoff/restart with `XAUTOCLAIM`/`XCLAIM`.
- Preserve per-chat ordering, use bounded retries/backoff plus a dead-letter path, and bound stream/dedupe retention. Please use separate Redis clients for blocking reads and request-path writes, as the existing realtime relay does.
- Define Redis outage behavior explicitly. A multi-replica deployment using this path should require a durable Redis configuration; enqueue must not silently degrade to an in-memory handoff.

Redis is still at-least-once transport, so this storage change does **not** by itself remove the first-rollout hazard: an old replica that sends directly creates neither a stream entry nor a dedupe key. The Redis design still needs an explicit cutover/grace gate for the reconciler during mixed-version rollout, plus a regression test and rollback behavior. Likewise, a crash after the platform accepted the send but before `XACK` can redeliver unless WeCom offers a reusable idempotency key.

Could you please either rework this split around Redis Streams, or explain concretely why Redis Streams cannot meet the durability/recovery requirements here and PostgreSQL must remain the queue of record?

wzxjohn:
@multica-eve The reason is simple: Redis is **NOT** currently a required dependency for this project. If we were to use Redis instead of a database, it would mean that users would have to enable Redis before they could use WeCom features. This would directly result in existing users completely losing their ability to connect to WeCom until the administrator completes the Redis configuration. In my view, this is unacceptable.
Unless Redis becomes a mandatory requirement for this project, we cannot use Redis for the relevant implementation.

wzxjohn:
@Bohan-J Here is the first PR split from #6581, the Queue part. Migration prefixes are stay untouched as request.

multica-eve:
There is an additional scalability requirement that blocks the current design.

The official Multica service must be designed for a large number of concurrently online installations, not only small self-hosted deployments. In the current implementation, every connected installation owns a consumer whose idle path calls `ClaimChannelOutbound` every `PollInterval = 2s` (`server/internal/integrations/channel/outbox/consumer.go`). This makes idle database traffic grow linearly with active installations: 10,000 online installations would generate roughly 5,000 empty claim queries per second before any real outbound traffic. In addition, `FailUndeliverableChannelOutbound` runs every two seconds on every server replica before the reconciler lease is acquired.

That fixed per-installation PostgreSQL polling model is not acceptable for the hosted service, and the PR currently has no benchmark demonstrating that the database can absorb the intended production scale. Please treat this as a must-fix rather than optimizing it after merge.

We recommend using Redis Streams for the multi-replica handoff, following the repository's existing Redis-based realtime operational model:

- use blocking stream reads with a bounded number of Redis connections/readers, not a polling loop per installation;
- retain stable business-key deduplication and per-chat ordering;
- acknowledge only after WeCom confirms the send, reclaim abandoned pending entries after lease handoff, and define bounded retry, dead-letter, and retention behavior;
- provide load-test results at the expected hosted scale, including idle installations, burst delivery, reconnect, and lease-transfer scenarios.

Redis being optional for small self-hosted installations does not require the hosted and multi-replica path to use the same transport. A no-Redis single-node fallback can remain available, but it should be a separate, explicitly bounded mode; the scale constraints of that fallback must not dictate the architecture of the official service.

The previously reported mixed-version rollout duplication issue also remains and still needs a code-level cutover strategy. Changing the transport to Redis does not remove that requirement by itself.

wzxjohn:
Because DB queue is not accepted by @multica-eve , I will waiting for new version with required Redis and use it for queue.

PR #5959, opened 2026-07-26
MUL-5329: fix(codex): honor configured first-turn timeout

YanshuoH:
Summary:
- remove the hidden 30-second cap from the first-turn Codex progress watchdog
- scale the first-turn window from the configured semantic inactivity timeout
- add a regression test for a two-minute setting

Reproduction:
A WSL Codex app-server emitted status: running and its first model event about 39 seconds later. The unmodified daemon killed it at 30 seconds even with --codex-semantic-inactivity-timeout 2m; the patched build completed successfully.

Test:
go test ./pkg/agent -run TestCodex -count=1

Issue #3262, opened 2026-05-26
[Bug]: Codex app-server stays at status=running with no output; Hermes works


Arobotmaster:
### Deployment type

multica.ai Desktop / hosted workspace

### What happened?

A Codex-backed local agent cannot complete even a minimal chat/task. The daemon starts `codex app-server --listen stdio://`, creates the Codex thread, sends `turn/start`, then only receives a `status=running` event with no content or output. The server later cancels the task with zero output.

Switching the same agent to Hermes works in the same workspace, on the same daemon, with the same machine and auth. A minimal Hermes smoke-test task completed successfully.

This looks related to the existing Codex app-server / semantic inactivity class of issues, but in this case it happens on the first simple prompt before any tool call or useful output.

### Local environment

- macOS Apple Silicon
- Multica Desktop bundled CLI: `multica v0.3.8` (`/Applications/Multica.app/Contents/Resources/app.asar.unpacked/resources/bin/multica`)
- Homebrew CLI also exists: `multica 0.2.30`, but the repro below was confirmed with the Desktop bundled `v0.3.8` daemon
- Codex CLI: `codex-cli 0.133.0`
- Hermes: `Hermes Agent v0.13.0`
- Workspace ID tested: `469b32b2-0c27-4865-8ed9-9ae2c5a94eb0`
- Codex runtime ID: `d8b6392c-6570-49a5-9a11-dd5ebaf8f9cc`
- Hermes runtime ID: `2ada8ff7-61c8-4668-81e3-af049624ad70`

`codex doctor` reports no failures. Running Codex directly via `codex exec` can produce a normal response, so this does not appear to be a broken Codex login.

### Steps to reproduce

1. Start the Desktop bundled daemon:

```bash
/Applications/Multica.app/Contents/Resources/app.asar.unpacked/resources/bin/multica \
  --profile desktop-api.multica.ai daemon restart
```

2. Confirm runtimes are online:

```bash
/Applications/Multica.app/Contents/Resources/app.asar.unpacked/resources/bin/multica \
  --profile desktop-api.multica.ai runtime list
```

Observed Codex and Hermes are both online.

3. Configure an agent to use the Codex runtime and clear model override:

```bash
/Applications/Multica.app/Contents/Resources/app.asar.unpacked/resources/bin/multica \
  --profile desktop-api.multica.ai agent update <agent-id> \
  --runtime-id d8b6392c-6570-49a5-9a11-dd5ebaf8f9cc \
  --model ""
```

[...]

PR #6753, opened 2026-08-11
MUL-6019: fix(codex): add MULTICA_CODEX_FIRST_TURN_TIMEOUT to raise the first-turn ceiling

ADanMan:
## Summary

The Codex first-turn no-progress watchdog (`codexFirstTurnNoProgressTimeout`) caps at a 60s default that the configured semantic-inactivity timeout can only ever *shrink*, never raise. So on an app-server that is legitimately slow to its first event — a heavy MCP boot, a cold model catalog — an operator has no way to extend that ceiling: a healthy first turn is killed at 60s with zero output, and nothing they can set changes it.

This adds a dedicated, explicit knob for exactly that, and nothing else.

Addresses **Bug B of #3262** (the configurable first-turn timeout). That thread tracks two distinct bugs — the model-catalog refresh stall (Bug A, handled by #3291) and this cap — so this is `Refs`, not `Fixes`.

**Supersedes #5959.** Thanks to @YanshuoH for finding the bug and landing on the right function, and @nikolai-vysotskyi for the precise write-up. That PR removed the cap and scaled the first-turn window from the semantic-inactivity timeout, which stretched the unconfigured default from 60s to ~8m and delayed the #3291 model-catalog auto-recovery that fires on the first-turn timeout. Both reviews (@Bohan-J, @multica-eve) asked instead for a dedicated knob that leaves the default byte-identical — that is what this does.

## Change

- New `MULTICA_CODEX_FIRST_TURN_TIMEOUT` env var (`server/internal/daemon/config.go`), default `0` = unset.
- New `FirstTurnNoProgressTimeout` field on `agent.ExecOptions`, plumbed from the daemon config at the codex exec site.
- `codexFirstTurnNoProgressTimeout(semantic, configured)`: when `configured > 0` it sets the first-turn watchdog ceiling to that value (upward included); otherwise byte-identical to before — the 60s default, which the semantic-inactivity timeout can only shrink. This ceiling is only the first-turn timer's own duration: the effective first-item wait is still min(ceiling, semantic, execution), and the semantic timer is armed first, so the semantic timeout must be set strictly above the first-turn timeout to preserve the #3291 startup retry.
- Docs: the env var added to `environment-variables.mdx` (+ zh/ja/ko).

Default behaviour is unchanged for every existing deployment: with the knob unset the resolver returns exactly what it did before, and `TestCodexFirstTurnNoProgressTimeoutClamp` (which pins the 60s default and the shrink-only contract) is preserved as-is.

On the overflow point from the #5959 review: the `semantic * 4 / 5` multiplication stays inside the clamp (`0 < semantic <= 60s`), so it cannot overflow — that risk existed only in the uncapped version.

(The `config.go` diff is mostly gofmt re-alignment of the struct block; the substantive change is ~14 lines — `git diff -w`.)

## Test plan

- New `TestCodexFirstTurnNoProgressTimeoutExplicitOverride`: override honored upward past the ceiling; override wins over the default 10m semantic and over a value that would otherwise scale down; raises the resolver ceiling above the default; zero keeps the default; negative falls back.
- `TestCodexFirstTurnNoProgressTimeoutClamp` unchanged (default / shrink-only contract preserved).
- `go test ./pkg/agent` (full) ✅ · `go test ./internal/daemon` (config/env) ✅ · `go vet ./pkg/agent/... ./internal/daemon/...` ✅ · `gofmt` clean.

Refs #3262


multica-eve (changes requested):
Before this change, self-hosted operators cannot extend the one-minute wait for a healthy task to show its first visible progress. After this change, they can opt into a longer or shorter wait while unchanged installations keep today's timing and recovery behavior. That product direction is valuable, but the current description permits a setting that appears to be accepted while another existing limit ends the task earlier and skips an automatic recovery.

Product verdict: ✅ We should accept a dedicated first-turn setting with an unchanged default.

## Blocking: the effective limit and recovery behavior do not match the contract

The semantic timer remains armed at `server/pkg/agent/codex.go:1468`. When `status:running` arrives, the same activity both resets that timer at `:1504` and starts the first-turn timer at `:1508`. With no subsequent progress, the effective first-item wait is therefore:

```text
min(FirstTurnNoProgressTimeout, SemanticInactivityTimeout, execution timeout)
```

For example, setting the new value to `30m` while keeping the default semantic timeout at `10m` does not produce a 30-minute first-turn window: the semantic timeout wins after roughly 10 minutes. This contradicts the “wins outright” comment at `server/pkg/agent/codex.go:1923-1927` and the “unbounded above” test wording at `server/pkg/agent/codex_test.go:787`; the resolver returns 30 minutes, but a real execution cannot observe that budget under the default configuration.

There is also a recovery consequence. The model-catalog retry at `server/pkg/agent/codex.go:1607` requires `timeoutDiagnostic.Kind == codexTimeoutFirstTurnNoProgress`. When the semantic timer wins, the same startup failure is classified as semantic inactivity, so `startupRefreshRetrySafe` is false and the #3291 retry does not run. An operator can therefore set a documented value that is silently truncated and simultaneously bypasses the existing automatic recovery this PR intends to preserve.

Please fix the operator contract before merge. The smallest acceptable shape is:

1. Update all four `environment-variables*.mdx` pages to state that the effective first-item wait is still bounded by `MULTICA_CODEX_SEMANTIC_INACTIVITY_TIMEOUT` (and the overall execution timeout), and that both values must be raised when the desired first-turn wait exceeds the semantic limit.
2. Emit a startup warning when `MULTICA_CODEX_FIRST_TURN_TIMEOUT` is greater than the semantic timeout, explaining both the truncation and the model-catalog retry consequence.
3. Remove or narrow the “wins outright / unbounded above” claims in the code comments and test names. Add a runtime-level regression test that pins the chosen timer interaction; the current tests exercise only the resolver.

Non-blocking follow-ups: add a `LoadConfig` test for empty/explicit/zero values, decide whether the public name should include `NO_PROGRESS` before release, and either add parity with the existing CLI/persisted Codex timeout settings or document that this one is environment-only.

🔄 **Request changes** — the feature direction and default-preserving implementation are good, but the current public contract hides a real upper bound and a configuration-dependent loss of automatic recovery.

Validation at `fcbbb8e9`: the targeted timeout tests, daemon brief byte-identity guard, and `git diff --check` pass; the current GitHub CI is also green. These checks do not cover the competing-timer behavior above.

ADanMan:
Addressed in `e488a599c`:

- Docs (all four `environment-variables*.mdx` locales) now state the effective first-turn wait stays bounded by `MULTICA_CODEX_SEMANTIC_INACTIVITY_TIMEOUT` and the overall execution timeout, and that both must be raised together when the desired wait exceeds the semantic limit.
- `LoadConfig` now warns when `MULTICA_CODEX_FIRST_TURN_TIMEOUT` exceeds the semantic timeout, naming both the truncation and the disabled #3291 model-catalog startup retry.
- Narrowed the "wins outright / unbounded above" resolver comment and test-case name to describe only the first-turn watchdog ceiling.
- Added `TestCodexExecuteFirstTurnOverrideAboveSemanticIsTruncated`, a runtime test that drives the real run loop with an override above the semantic timeout and asserts the semantic timer wins the race and the startup retry is skipped (`retry_safe=false` with the catalog-refresh signal present). Also added a `LoadConfig` empty/explicit/zero test.

On the non-blocking naming/parity points: the knob is intentionally environment-only (no `Overrides`/CLI parity), documented in the new test — happy to add CLI parity in a follow-up if you would prefer it symmetric with the other Codex timeouts.

multica-eve (changes requested):
Most of the previous blocking review is addressed well in `e488a599c`: the four documentation locales now describe the competing limits, configuration loading emits a warning, the resolver wording is scoped to the first-turn watchdog itself, and the new run-loop regression test proves that a smaller semantic timeout wins and skips the #3291 retry. The `LoadConfig` coverage is also a useful addition.

One equality edge still blocks approval.

## Must fix: matching the two values does not preserve the startup retry

`server/internal/daemon/config.go:318` warns only when the first-turn timeout is **greater than** the semantic timeout, and the warning at `:319` tells operators to raise the semantic timeout “to match.” However, the runtime order is:

1. `server/pkg/agent/codex.go:1504` resets the semantic timer.
2. `server/pkg/agent/codex.go:1508` then creates the first-turn timer with the same duration.

If the two configured durations are equal, the semantic deadline is therefore earlier. Even if scheduler delay makes both channels ready together, Go's `select` does not guarantee that the first-turn branch wins. The failure can still be classified as semantic inactivity, while the model-catalog recovery at `server/pkg/agent/codex.go:1607` requires `codexTimeoutFirstTurnNoProgress`. An operator who follows the new warning exactly (`15m` and `15m`) can therefore receive no warning and still silently lose the #3291 startup retry.

Please:

- warn when `firstTurn >= semantic`, not only `>`;
- tell operators to set the semantic timeout **strictly above** the first-turn timeout (preferably with some margin), rather than “to match”;
- update all four documentation locales accordingly; and
- add an equal-duration runtime/config-warning regression test that pins this edge.

Non-blocking: the PR description still contains the old “wins outright / unbounded above” language, so it would be helpful to align that with the corrected code and docs.

🔄 **Request changes** — the original contract issue is nearly resolved, but the currently recommended equal-value configuration can still bypass the automatic recovery this change promises to preserve.

Validation at `e488a599c`: `go test ./pkg/agent -count=1`, `go test ./internal/daemon -count=1`, the new runtime regression test repeated 10 times, the daemon brief byte-identity guard, `go build ./...`, `go vet ./pkg/agent/... ./internal/daemon/...`, and `git diff --check` all pass. Current GitHub CI is green.

ADanMan:
Thanks — good catch on the equality edge. Fixed in `5d682b49f`:

- The warning now fires when `firstTurn >= semantic` (was `>`): the semantic timer is armed first, so at equal durations the semantic deadline is reached first, and Go's `select` doesn't deterministically favour the first-turn branch even when both fire together — so equal values can still drop the #3291 retry.
- The warning and all four `environment-variables*.mdx` locales now tell operators to set `MULTICA_CODEX_SEMANTIC_INACTIVITY_TIMEOUT` **strictly above** the first-turn timeout (with some margin), not "to match".
- Added `TestLoadConfig_CodexFirstTurnTimeoutEqualToSemanticWarns`, which captures startup logs and pins that equal durations warn, first-turn-above-semantic warns, and semantic-strictly-above stays quiet.
- Non-blocking: also updated the PR description to drop the "wins outright / unbounded above" wording so it matches the corrected code and docs.

multica-eve (changes requested):
The equality-boundary fix at `5d682b49f` addresses the previous implementation blocker: the warning now covers `firstTurn >= semantic`, the operator guidance consistently requires the semantic timeout to be strictly higher, and the regression coverage pins the warning boundary. I do not see a remaining must-fix in that correction itself.

Two merge blockers remain on the current PR state:

1. GitHub reports this branch as conflicting with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`). Please update the branch and resolve the conflict, then confirm that the resolved configuration code preserves the `>=` warning boundary and the new timeout field.
2. The current head has no reported checks. Please run the relevant tests/CI on the conflict-resolved head so the final combined result is verified.

🔄 **Request changes** — the timeout-boundary, warning, documentation, and regression-test fixes are acceptable; please resolve the branch conflict and obtain passing checks on the resulting head before merge.

Issue #5361, opened 2026-07-13
[Feature]: Add ZCode (Z.AI coding agent) as a supported agent runtime


MarcoHerten:
### Deployment type

Official App

## Summary

I'd love to be able to drive **ZCode** (Z.AI's coding agent, https://z.ai) as a Multica
agent runtime, alongside the currently supported CLIs. After investigating, ZCode
*does* expose a headless stdio protocol that could be driven by the daemon — but it
is **not** ACP-compatible, so it can't be wired in via the existing Qoder/Trae/Kiro
path. Filing this to (a) request support, and (b) document what I found in case it's
useful for a future integration.

## What is ZCode?

ZCode is Z.AI's CLI/TUI coding agent. It ships:

- A headless single-prompt mode: `zcode -p "<prompt>"`
- A stdio **app server**: `zcode app-server --stdio`

It's distributed as a desktop app (`ZCode.app` on macOS) bundling both a native binary
(`zcode-agent`) and a node bundle (`zcode.cjs`). The `app-server --stdio` subcommand is
what an integration would target.

## Why it can't be auto-detected today

1. ZCode's command name (`zcode`) isn't in the daemon's discovery list, and there's no
   `MULTICA_ZCODE_PATH` config var analogous to the per-agent overrides for the 13
   supported CLIs.
2. Unlike Qoder (`qodercli --acp`), Trae (`traecli acp serve`), and Kiro, ZCode does
   **not** speak the Zed Agent Client Protocol — so it can't piggyback on the existing
   ACP integration path.

## Technical findings — the "ZCode Protocol"

I ran the app server and probed the wire format. For reference / verification:

**Invocation**
```bash
node /path/to/zcode.cjs app-server --stdio

(spawnArgs:["app-server","--stdio"] per the bundled config; there's also a native zcode-agent binary selectable via GLM_BINARY_PATH.)

Framing & envelope — newline-delimited JSON, without a jsonrpc field:


json
{"id": 1, "method": "session/create", "params": { "workspace": { ... } }}
id may be a string or a number
Sending a jsonrpc: "2.0" key is rejected with Unrecognized key: "jsonrpc"
Error codes look JSON-RPC-derived (-32601 Method not found, -32602 Invalid params, -32600 Invalid request, -32700 Parse error), but the envelope itself is custom.
No handshake. Unlike ACP, there is no initialize/authenticate step — the server answers Method not found: initialize and Method not found: ping.

[...]

PR #6755, opened 2026-08-11
MUL-6022: feat/zcode runtime — built-in ZCode agent runtime with streaming events

tornado404:
## What does this PR do?

Adds **ZCode** (the Z.ai GLM coding agent, driven by [`zcode-cli-stream`](https://www.npmjs.com/package/zcode-cli-stream)) as a first-class built-in agent runtime in Multica, with full streaming event support — so a Multica task assigned to ZCode shows the agent's thinking, tool calls, tool results, and incremental text in real time, just like Claude / Codex / Qwen.

## Background & motivation

ZCode is Z.ai's coding agent (GLM-5.2 family). Users with a ZCode account get platform benefits — promotional quota, peak/off-peak rates, etc. — that apply automatically through the standard ZCode login. This PR lets those users run ZCode **inside Multica's unified task management**, so they get the agent-orchestration experience (issue-driven tasks, review, multi-agent coordination) while keeping their ZCode account benefits.

Previously ZCode was not available as a runtime: the daemon couldn't discover it, the frontend picker didn't list it, and even if wired up, the zcode-cli headless mode (`--prompt --json`) was non-streaming — Multica would see nothing until the entire turn finished.

This is a two-repo change. This PR is the Multica side; the zcode-cli side adds the `--stream-json` streaming mode this backend consumes.

## Runtime dependency: `zcode-cli-stream`

The upstream `kingsword09/zcode-cli` was forked and **published to npm as [`zcode-cli-stream`](https://www.npmjs.com/package/zcode-cli-stream)** (`@3.7.5-11`) because the upstream PR ([kingsword09/zcode-cli#69](https://github.com/kingsword09/zcode-cli/pull/69)) is still pending review.

| | |
|---|---|
| **npm package** | `zcode-cli-stream` |
| **install** | `npm install -g zcode-cli-stream@latest` |
| **bin command** | `zcode` (unchanged from upstream — Multica keys on this) |
| **fork repo** | [tornado404/zcode-cli](https://github.com/tornado404/zcode-cli) |
| **upstream PR** | [kingsword09/zcode-cli#69](https://github.com/kingsword09/zcode-cli/pull/69) |

The fork adds `--stream-json` (NDJSON streaming output for headless mode) and is MIT-licensed with full attribution to the original author. Once upstream merges the PR, users can switch back to `zcode-app-cli` (the original package name) with no Multica-side change — both expose the `zcode` command.

### 1. Daemon discovery + whitelist (commits `d4c704c`, `e8fed47`, `3f561a2`)
- `server/internal/daemon/agents_probe.go` — probe `zcode` on PATH (via `MULTICA_ZCODE_PATH` override)
- `server/internal/daemon/config.go` — add `"zcode"` to `defaultAgentCommandNames` (login-shell PATH fallback for GUI-launched daemons)
- `scripts/agent-cli-command-names.txt` — add `zcode` to the CLI guard allowlist (fixes `TestAgentCLIGuardCoversDefaultCommands`)
- `server/migrations/273_runtime_profile_add_zcode.{up,down}.sql` — add `zcode` to the `runtime_profile.protocol_family` CHECK constraint (enables custom zcode runtime profiles)

### 2. Streaming backend (commit `0727300`)
- `server/pkg/agent/zcode_stream.go` (**new**) — `zcodeStreamBackend` drives `zcode --prompt <text> --stream-json` and parses qwen-compatible NDJSON into live `Message`s. Reuses the qwen event schema and shared streaming machinery (`finalizeStreamResult`, `streamTerminalState`, `newAgentStreamScanner`, `trySend`)
- `server/pkg/agent/zcode.go` — legacy non-streaming `--json` backend retained as fallback reference; file header updated
- `server/pkg/agent/agent.go` — `New("zcode")` returns `zcodeStreamBackend`; `launchHeaders` updated to `zcode --prompt --stream-json`
- `server/pkg/agent/zcode_stream_test.go` (**new**) — 4 tests: factory type, arg building (protocol-critical flags blocked), full event-sequence parsing (status → thinking → text → tool_use → tool_result → result+usage), error-result handling
- `server/pkg/agent/models.go` — `ModelSelectionSupported` returns false for zcode (model is fixed by `~/.zcode/cli/config.json`)
- `server/internal/daemon/daemon.go` — `"zcode": "ZCode"` display-name override
- `server/internal/daemon/execenv/runtime_config.go` — AGENTS.md path for zcode

### 3. Frontend runtime picker (commit `15598eb`)
- `packages/core/types/agent.ts` — add `"zcode"` to `RUNTIME_PROFILE_PROTOCOL_FAMILIES` (the hardcoded list the "add runtime" picker renders from — **this was the missing piece that hid zcode from the UI**)
- `packages/views/runtimes/components/provider-logo.tsx` — add `ZcodeLogo` component + switch case
- `packages/views/runtimes/components/zcode-logo.png` (**new**) — official ZCode app icon (extracted from `ZCode.app/Contents/Resources/icon.icns`, 512px)
- `packages/core/runtimes/display.ts` — `zcode: "ZCode"` display-name override (brand casing)

## How to Test

```bash
# 1. Install the streaming-enabled zcode CLI from npm
npm install -g zcode-cli-stream@latest

# 2. Login (standard ZCode OAuth — your account quota/rates apply)
zcode login

# 3. Verify streaming works
zcode --prompt "reply with exactly: hello" --stream-json
# Should output NDJSON: system → assistant → result

# 4. Build & run Multica, then Settings → Runtimes → ZCode appears in the picker
# 5. Create a task assigned to ZCode — watch thinking/tool-calls/text stream live
```

## Streaming event flow

```
zcode --stream-json stdout          →  zcodeStreamBackend  →  Multica Message channel
─────────────────────────────────────────────────────────────────────────────────
system/init                        →  MessageStatus (pins session_id for resume)
assistant/thinking                 →  MessageThinking
assistant/text                     →  MessageText
assistant/tool_use                 →  MessageToolUse
user/tool_result                   →  MessageToolResult
result (success/error + usage)     →  Result{Status, Output, Usage, SessionID}
```

Session continuity: the session id arrives on stream events and is returned via `Result.SessionID`; the daemon feeds it back as `--resume` on the next turn.

## Security analysis

- **No credential/auth/billing logic changed.** The daemon already probes every CLI the same way (PATH lookup + `--version`); zcode follows the identical, long-established pattern used by codex/claude/qwen/grok/etc.
- **Scanned the full diff for API keys, tokens, secrets, Bearer headers** — none present. (Matches on "token" are the billing `inputTokens`/`outputTokens` usage fields, not credentials.)
- **The streaming backend only reads zcode's stdout.** No network calls, no file access beyond what every backend already does (`cmd.Dir = opts.Cwd`).
- **zcode-logo.png** is the official app icon (PNG, 512×512, RGBA) — not a screenshot, contains no sensitive info.
- **No new dependencies** (Go or TS).
- **`zcode-cli-stream` was also audited**: its injected helper only writes session events to stdout; no exfiltration, no new network calls. MIT-licensed, full attribution to original author.

## Checklist

- [x] Thinking path traces from project context to this change
- [x] Tests pass locally (`go test ./pkg/agent/ -run 'Zcode|zcode'` — 9/9 pass)
- [x] Tests added (4 new streaming tests in `zcode_stream_test.go`)
- [x] `pnpm typecheck` passes (6/6 tasks)
- [x] Synced to landing copy/docs — N/A (no new user-facing runtime tab beyond the standard picker entry)
- [x] Risks documented (requires zcode-cli with `--stream-json`; legacy `--json` backend retained as fallback)

## Companion

- **npm package**: [`zcode-cli-stream@3.7.5-11`](https://www.npmjs.com/package/zcode-cli-stream) (fork of `kingsword09/zcode-cli` with `--stream-json`)
- **fork repo**: [tornado404/zcode-cli](https://github.com/tornado404/zcode-cli)
- **upstream PR**: [kingsword09/zcode-cli#69](https://github.com/kingsword09/zcode-cli/pull/69)

## AI Disclosure

**AI tool used:** ZCode (GLM-5.2)

**Approach:** Event schema derived by probing the zcode runtime with a temporary `onEvent` dump and capturing real event payloads. The Multica streaming backend reuses the existing qwen parser since the companion package emits qwen-compatible NDJSON. Frontend icon extracted from the locally-installed `ZCode.app` bundle.


multica-eve (changes requested):
Product change: today, teams that use ZCode cannot choose it as a Multica agent runtime. This PR would make ZCode selectable, let members run and continue tasks through it, and show live progress while ZCode keeps control of model selection.

Product judgment: the workflow is useful, but I do not think it is safe to ship as a built-in runtime yet. The implementation currently depends on an unpublished contributor fork rather than the normally installable ZCode CLI, and the product/support boundary for an unofficial client is not documented.

Request changes:

1. **[P0] Do not make an unpublished fork the production protocol contract.** `server/pkg/agent/zcode_stream.go:13-24` says the backend requires `tornado404/zcode-cli`'s `feat/stream-json-output` branch, and `buildZcodeStreamArgs` always adds `--stream-json`. The current public `zcode-app-cli` release (3.7.5-11, runtime 0.16.1) supports `--json` but does not expose `--stream-json`. The daemon nevertheless detects any `zcode` executable and `agent.New("zcode")` always selects this stream backend, so a standard installation is advertised as available and then every task fails on an unknown flag. Please base the integration on a published upstream capability, gate detection on a verified minimum version/capability, or provide a tested compatible fallback. The shell-generated tests only replay the fork's proposed schema and do not establish compatibility with a released CLI.

2. **[P1] Write bound skills to a directory ZCode actually discovers.** `server/internal/daemon/execenv/context.go:336-430` has no `zcode` mapping, so it falls back to `.agent_context/skills`. ZCode scans project skills under `.zcode/skills` / `.agents/skills`; bound skills therefore will not be available to the agent. Add the provider-native mapping and a regression test for skill placement.

3. **[P1] Cancel the whole process tree, not only the direct CLI process.** `server/pkg/agent/zcode_stream.go:71-75` uses `exec.CommandContext` plus `WaitDelay`, but never calls the repository's process-group helpers. The EOF chain is stdout scanner -> `cmd.Wait()` -> `Result`; a spawned tool can keep the pipe open, and cancellation can leave descendants running after the direct process is killed. Use the existing `configureProcessGroup` / graceful tree-termination pattern and add a cancellation test with a grandchild that holds stdout open.

4. **[P1] Protect every daemon-owned argument and its aliases.** `server/pkg/agent/zcode.go:18-22` blocks only `--prompt`, `--json`, and `--resume`. Because custom arguments are appended last, callers can still override the work directory or execution contract with `--cwd`, `--max-turns`, `--stream-json`, `-p` / `--print`, and `-c` / `--continue`. Expand the blocklist and cover both `--flag value` and `--flag=value` forms in table-driven tests.

5. **[P1] Document the support and installation contract before exposing this as built-in.** There is no ZCode provider/runtime documentation in this PR. Please document that this is an unofficial client integration, the exact supported published version and authentication/setup steps, model-selection behavior, resume/skills behavior, and the limitations maintainers are committing to support.

Also, `zcodeBackend` is now unreachable code retained “for reference.” Please either turn it into the explicit, version-gated fallback with end-to-end coverage or remove it rather than carrying two implementations with only one executable path.

Validation on `0727300ed`: `git diff --check`, the ZCode-focused agent tests, and the required daemon execenv byte-identical test all pass. These checks do not cover the released CLI compatibility or process-tree cases above.

Frankli9986:
Adding a supporting voice from a real user team: we run Multica to orchestrate our coding agents (Claude Code / Codex / Kimi / Hermes), and ZCode (Z.AI / GLM) is actively used by developers in our org. Having it as a first-class runtime would let us schedule and track ZCode tasks in the same place as the rest — exactly the workflow this PR targets.

We'd also like to help with the review blockers, especially the P0 concern about depending on an unpublished fork: we can test against a standard `zcode-app-cli` install on macOS (arm64) to confirm what a released CLI does with and without `--stream-json`, and validate a capability-gated fallback once it's in place. Happy to run whatever verification would be most useful to the maintainers.

Thanks for driving this — we're glad to provide real-world feedback on a test build whenever one is available.

multica-eve (changes requested):
Follow-up review of `d88c09217`. The product direction remains useful, subject to the maintainer decision to support an unofficial community client as a built-in runtime. The skills mapping, managed-argument protection, multilingual support documentation, and the original unknown-`--stream-json` failure have been addressed. The current public package (`zcode-app-cli@3.7.6-12`, CLI 0.16.3) still has no `--stream-json`, and the new capability gate correctly sends it to the JSON fallback.

The following must-fix issues remain:

1. **[P1] Apply process-tree cleanup and liveness handling to the JSON fallback, which is the standard public CLI path.** `server/pkg/agent/zcode_stream.go:135-137` delegates a normal published install to `zcodeBackend`, but `server/pkg/agent/zcode.go:92-202` still uses plain `exec.CommandContext` plus `cmd.Output()` without the process-group cleanup added to the streaming path. The complete EOF dependency is: ZCode and every descendant close stdout -> `Output()` returns -> JSON is parsed -> the first `MessageText` is emitted -> `Result` is delivered. Until that EOF, a legitimate task is completely silent and the daemon's 30-minute idle watchdog can terminate it. If a tool descendant holds stdout, cancellation kills only the direct process, so EOF/Result can remain delayed and the descendant can survive. Please share the process-tree termination logic with the fallback and add cancellation/no-orphan coverage for the fallback on every supported platform. The new Unix streaming tests do not cover this production path; the Windows process helper still kills only the leader.

2. **[P1] Fix error, resume, and prompt-privacy semantics on the JSON fallback.** `server/pkg/agent/zcode.go:112` logs the full argv, including the user's task prompt. `zcode.go:124-159` sends stderr only to daemon logs, so auth/config/network/invalid-resume failures remain an opaque exit status to the user. `zcode.go:193-199` never sets `ResumeRejected`, while zcode is absent from `resumeRejectionUndetectable`, so a rejected persisted session cannot trigger the intended fresh-session fallback. Please avoid logging the prompt, return a bounded stderr tail, positively identify resume rejection, and cover auth failure, invalid resume, cancellation, and malformed JSON with fake-executable tests.

3. **[P1] Define the supported public-version contract.** `server/pkg/agent/version.go:13-19` has no ZCode minimum, while discovery and docs currently accept any `zcode` executable. The fallback assumes support for `--prompt`, `--json`, `--resume`, `--cwd`, `--max-turns`, and a specific JSON shape. Establish and enforce a verified minimum version (or capability-gate the complete summary protocol), then document it so older incompatible installations are not registered as healthy runtimes.

4. **[P0] Fix the duplicate migration prefix and restore CI.** `273_runtime_profile_add_zcode` now conflicts with main's `273_agent_task_queue_runtime_id_index`. GitHub's backend job fails `TestMigrationNumericPrefixesStayUniqueAfterLegacySet` for this exact collision. Rebase on current main and rename both the up/down migration files to the next unique prefix at that point.

Validation on `d88c09217`: `git diff --check` passes; ZCode-focused agent tests (including capability fallback and Unix streaming process-group cancellation) pass; execenv `ByteIdentical|Zcode` tests pass. GitHub CI is currently 9/10, with backend failing on the migration-prefix collision above.

**Recommendation: Request changes.** Keep the skills, argument, documentation, and capability-gate improvements; complete the lifecycle/error/resume guarantees on the public JSON path, define the supported-version boundary, and fix the migration collision before the next review.

multica-eve (changes requested):
Follow-up review of `607c75532`. The earlier prompt-privacy, stderr, invalid-resume, malformed-JSON, managed-argument, skills, documentation, and Unix cancellation improvements are preserved, but the previous must-fix set is not fully resolved.

1. **[P1] The public JSON fallback still misclassifies legitimate long turns as idle.** `server/pkg/agent/zcode.go:57-60,172-178,222-226` emits its first message only after the entire turn completes and stdout reaches EOF. The daemon starts its no-message watchdog when `Execute` begins, with a 30-minute default (`server/internal/daemon/config.go:38-50`). A valid public-CLI turn lasting longer than that is therefore force-cancelled even while ZCode is working. Please give summary mode a liveness/absolute-timeout policy that does not treat expected silence as a hang, or use a protocol that supplies real progress.

2. **[P1] Process-tree cancellation is still Unix-only.** `server/pkg/agent/proc_windows.go:36-56` leaves process-group setup as a no-op, kills only the direct child, and always reports the process group as not gone. A hidden console and `WaitDelay` do not terminate grandchildren, so ZCode tool subprocesses can survive cancellation on Windows. The new cancellation tests are Unix-tagged. Please use a Windows Job Object (or the repository's equivalent process-tree controller) and add Windows coverage.

3. **[P1] The compatibility gate does not cover the complete protocol, and the docs disagree with the cache behavior.** `zcodeSupportsJSONSummary` only checks whether help text contains `--json`, while execution also depends on `--prompt`, `--cwd`, optional `--resume`/`--max-turns`, and a fixed JSON response shape. The help output is cached for the daemon process lifetime, but all four installation docs say it is checked on every run. Please capability-gate the full contract or establish and enforce a verified supported-version contract, then align the docs/cache behavior and cover in-place CLI upgrade/downgrade.

4. **[P0] Rebase and use a migration prefix that is unique against current `main`.** GitHub reports this PR as `CONFLICTING` / `DIRTY`; a merge-tree check finds content conflicts in `provider-logo.tsx`, `server/internal/daemon/config.go`, `server/internal/daemon/execenv/runtime_config.go`, and `server/pkg/agent/agent.go`. Current `main` also already contains `310_workspace_private_plugin_identity` and has advanced through 313, while this PR adds `310_runtime_profile_add_zcode`. After resolving the content conflicts, the migration-prefix uniqueness test will fail. Please rebase current `main`, pick the next unoccupied prefix at that point, and rerun validation.

Validation on `607c75532`: the ZCode-focused agent tests, execenv ByteIdentical test, migration-prefix test, and `git diff --check` pass; GitHub shows 10/10 checks passing. Those results are against the PR's stale base and do not cover the current-main collision or the Windows process-tree gap.

**Recommendation: Request changes.**

lianxin255:
Thanks for putting this together — the version-gate + JSON-fallback design is a reasonable shape for the process-per-turn route.

We've been validating a different route that may help with the review blockers: driving the desktop-bundled `zcode app-server` (the stdio server the ZCode desktop app itself uses) through the community ACP bridge [zcode-acp-server](https://github.com/william0wang/zcode-acp) (published on npm, Apache-2.0, spawns the official runtime unmodified — no forked flags involved).

A few protocol notes from probing CLI 0.16.3, in case they're useful either way:

- `session/create` requires `workspace: {workspacePath, workspaceKey}` and now emits a `session/requestRuntimePreferences` server→client request that must be answered, or the create times out after 15s (this is what zcode-acp-server [#38](https://github.com/william0wang/zcode-acp/pull/38) fixed on its main).
- The event stream carries `tool.lifecycle`, `stream.chunk`, `usage.delta`, `turn.terminal` — enough for live progress throughout a turn, which also addresses the watchdog / long-turn-idle misclassification concern for the JSON fallback path.
- `session/stop` gives graceful cancel; `session/resume` continues a session across processes.

We plan to submit an ACP-based family PR that reuses the existing `hermesClient` transport (same path as kimi/qoder/dim), referencing #5361 — happy to coordinate so the two approaches don't duplicate effort.

Issue #5999, opened 2026-07-27
CLI JSON payloads exhaust agent context: --recent N is ignored, presigned URLs always emitted, no field selection


nikolai-vysotskyi:
## Summary

`multica issue get` and `multica issue comment list --output json` return payloads large enough to be a primary driver of agent context exhaustion. In one real SysAdmin agent session, **six CLI calls consumed 76.9k tokens** — more than half of the context that agent ever accumulated — and pushed it into repeated autocompact cycles until the run failed with *"Autocompact is thrashing: the context refilled to the limit within 3 turns of the previous compact, 3 times in a row."*

Since agents are Multica's core proposition, and prompt-cache writes dominate LLM billing, this is a direct and recurring cost to every workspace running agents on long threads.

Measured with `multica v0.4.11 (b0bae3f95)`.

## Measurements

Real issue with 26 top-level comments and 21 attachments:

| call | tokens |
|---|---|
| `issue comment list <id> --recent 4 --output json` | **43,238** |
| `issue comment list <id> --thread <t> --tail 30 --output json` | 13,648 |
| `issue get <id> --output json` | 8,567 |
| `issue comment list <id> --since <ts> --output json` | 9,219 |
| `issue comment list <id> --roots-only --output json` | 5,103 |

Field breakdown of the 43,238-token `--recent 4` response:

```
22616 tok  content        <- what the agent actually needs
 7871 tok  attachments    <- 21 records, agent asked for none
  225 tok  author_id
  225 tok  id
  225 tok  issue_id
  209 tok  parent_id
  125 tok  created_at
  125 tok  updated_at
```

And in `issue get`, attachments are nearly as large as the issue description itself:

```
8097 tok  description
7871 tok  attachments
 121 tok  metadata
  64 tok  title
```

### 1. `--recent N` does not limit anything

```
--recent 2  -> 26 top-level comments, 43366 tok
--recent 4  -> 26 top-level comments, 43366 tok
--recent 10 -> 26 top-level comments, 43366 tok
```

Byte-for-byte identical responses. The documented contract is "Return the N most recently active threads", and the agent-facing prompt Multica itself generates recommends `--recent 10` as the cheap cross-thread read — so an agent following Multica's own guidance believes it is paying for 10 threads and silently pays for all of them. This looks like a plain bug and is the highest-value fix here.

### 2. Presigned S3 URLs are emitted unconditionally, and are mostly signature

[...]

PR #6756, opened 2026-08-11
MUL-6023 feat(attachments): forced-attachment download URL across all storage modes

ADanMan:
## Summary

The follow-up to #6713 that #6712 tracks. #6713 landed the narrow client-only fix; this adds the server field that closes the two rows it left open (cross-origin media, and presign / token-mode).

The download button consumed `download_url`, which is **load-intent**: the preview and inline-media paths need it to serve media inline, so a cross-origin capability previews an image/PDF in-tab instead of downloading it, and a presigned `download_url` 401s in token-mode. This adds a separate, credential-free **`attachment_download_url`** that forces `Content-Disposition: attachment` in every storage mode, and points the web + desktop download buttons at it. `download_url` is left exactly as it was — the preview / inline-media paths keep consuming it.

## Server

New `attachment_download_url` on the single-attachment metadata response (`GET /api/attachments/{id}`) — **never in list responses** (it is short-lived, like the #6092 capability):

- **proxy**: a `dl=1` variant of the #6092 capability. The intent is folded into the signed HMAC message (`signAttachmentCapabilityIntent`), so a load-intent link cannot be flipped to a forced download by appending `dl=1`, nor the reverse. Redemption forces `storage.AttachmentContentDisposition`. Load-intent signatures stay **byte-identical**.
- **presign**: `PresignGetWithContentDisposition(key, ttl, AttachmentContentDisposition(filename))`.
- **cloudfront**: `SignedURLWithContentDisposition` — `response-content-disposition` is folded into the signed Resource, so a client cannot strip it.

Default behaviour is unchanged for every existing client: the field is additive and empty when the server cannot mint one, and `download_url` is untouched.

## Client

`use-download-attachment.ts` now prefers `attachment_download_url` on web and desktop, falling back to the #6092 proxy capability (the #6713 path) and then the cookie-gated endpoint for a server that predates the field. `download_url` consumers (`resolvePreviewMediaUrl`, the inline-media re-sign hook) are untouched. The zod schema and `Attachment` type gain the optional field (parsed with the existing `.loose()` fallback).

## Verification / acceptance grid

- Server: `go test ./internal/handler` (DB-backed) — incl. a new end-to-end test that mints `dl=1`, redeems it, and asserts `Content-Disposition: attachment` **on an image** (which the load-intent path serves inline). `go vet`, `gofmt` clean.
- Adversarial review (5 independent checks): domain-separation, load-intent-unchanged, not-in-list, disposition-forced-in-all-3-modes, capability-not-weakened — all held, no findings.
- Client: `@multica/core` + `@multica/views` typecheck; `use-download-attachment` vitest incl. new web + desktop cases for the field preference.
- Matrix covered: proxy / presign / CloudFront × media / non-media × web / desktop, plus the load-intent `download_url` staying inline for preview.

Follow-up to #6092 / #6713. Refs #6712.

### DIFF
--- a/packages/core/api/schemas.ts
+++ b/packages/core/api/schemas.ts
@@ -524,6 +524,11 @@ export const AttachmentResponseSchema = z.object({
   id: z.string(),
   url: z.string(),
   download_url: z.string(),
+  // Forced-attachment ("download button") URL — credential-free and, unlike
+  // `download_url`, always Content-Disposition: attachment across every storage
+  // mode. Optional: a server older than this field omits it, and callers fall
+  // back to `download_url` / the stable endpoint. Never persisted (short-lived).
+  attachment_download_url: z.string().optional(),
   markdown_url: z.string().optional().default(""),
   filename: z.string(),
   chat_session_id: z.string().nullable().optional(),
--- a/packages/core/drafts/draft-upload.ts
+++ b/packages/core/drafts/draft-upload.ts
@@ -77,12 +77,13 @@ export function attachmentToDraftUpload(attachment: Attachment): UploadedDraftUp
     filename: attachment.filename,
     size: attachment.size_bytes,
     contentType: attachment.content_type || undefined,
-    // `download_url` is minted for the current API response and may be a
-    // short-lived signed URL; draft uploads survive dialog closes and app
-    // restarts, so it is stripped here. `url`/`markdown_url` stay as the
-    // durable render/download paths, and content-editor's session merge
-    // backfills an empty download_url from the live upload result.
-    attachment: { ...attachment, download_url: "" },
+    // `download_url` and `attachment_download_url` are minted for the current API
+    // response and may be short-lived signed URLs; draft uploads survive dialog
+    // closes and app restarts, so both are stripped here. `url`/`markdown_url`
+    // stay as the durable render/download paths, and content-editor's session
+    // merge backfills an empty download_url from the live upload result (the
+    // download flow re-fetches attachment_download_url fresh via getAttachment).
+    attachment: { ...attachment, download_url: "", attachment_download_url: "" },
   };
 }
 
--- a/packages/core/types/attachment.ts
+++ b/packages/core/types/attachment.ts
@@ -11,6 +11,14 @@ export interface Attachment {
   url: string;
   download_url: string;
   /**
+   * Forced-attachment ("download button") URL: credential-free and, unlike
+   * `download_url` (load-intent, serves media inline for preview), always
+   * Content-Disposition: attachment across every storage mode. Short-lived —
+   * never persist it. Optional: a server older than this field omits it, and
+   * download callers fall back to `download_url` / the stable endpoint.
+   */
+  attachment_download_url?: string;
+  /**
    * Durable URL the client persists into markdown bodies.
    *
    * The server (`buildMarkdownURL` in server/internal/handler/file.go)
--- a/packages/views/editor/use-download-attachment.test.tsx
+++ b/packages/views/editor/use-download-attachment.test.tsx
@@ -295,6 +295,64 @@ describe("useDownloadAttachment (web)", () => {
     expect(clickSpy).not.toHaveBeenCalled();
     await waitFor(() => expect(toast.error).toHaveBeenCalled());
   });
+
+  it("prefers the forced-attachment URL over the capability and slug endpoint", async () => {
+    getAttachmentMock.mockResolvedValueOnce({
+      id: "att-1",
+      url: "https://static.example.test/pic.png",
+      // A proxy capability is present, but the forced-attachment URL wins.
+      download_url:
+        "/api/attachments/att-1/signed-download?exp=1800000060&sig=load",
+      attachment_download_url:
+        "/api/attachments/att-1/signed-download?exp=1800000060&sig=dl&dl=1",
+      filename: "pic.png",
+    });
+    const clickSpy = vi
+      .spyOn(HTMLAnchorElement.prototype, "click")
+      .mockImplementation(() => {});
+    const appendSpy = vi.spyOn(document.body, "appendChild");
+
+    const { result } = renderHook(() => useDownloadAttachment());
+    await act(async () => {
+      await result.current("att-1");
+    });
+
+    expect(clickSpy).toHaveBeenCalledOnce();
+    const anchor = appendSpy.mock.calls
+      .map(([node]) => node)
+      .find((node): node is HTMLAnchorElement =>
+        node instanceof HTMLAnchorElement,
+      );
+    expect(anchor).toBeDefined();
+    // The forced-attachment capability is used verbatim (same-origin base).
+    expect(anchor!.getAttribute("href")).toBe(
+      "/api/attachments/att-1/signed-download?exp=1800000060&sig=dl&dl=1",
+    );
+  });
+
+  it("uses the forced-attachment URL even when the workspace slug is missing", async () => {
+    useWorkspaceSlugMock.mockReturnValueOnce(null);
+    getAttachmentMock.mockResolvedValueOnce({
+      id: "att-1",
+      url: "https://static.example.test/pic.png",
+      attachment_download_url:
+        "/api/attachments/att-1/signed-download?exp=1800000060&sig=dl&dl=1",
+      filename: "pic.png",
+    });
+    const clickSpy = vi
+      .spyOn(HTMLAnchorElement.prototype, "click")
+      .mockImplementation(() => {});
+
+    const { result } = renderHook(() => useDownloadAttachment());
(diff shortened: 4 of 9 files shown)


Bohan-J (changes requested):
Reviewed at `c6008d2`, after the presign-test fix.

**Product take: yes, we want this.** `download_url` carrying load-intent while the download button reused it is the actual root cause behind the two rows #6712 still tracked, and splitting off a forced-attachment sibling is the right shape. Leaving the preview / inline-media paths on `download_url` keeps the blast radius small, and the three-step client fallback is the right call for an API boundary that installed desktop clients cross.

Verified locally against a real PostgreSQL at `c6008d2`: `go build ./...`, `go vet ./internal/handler`, `go test ./internal/handler` (full package), `go test ./internal/storage/... ./internal/auth/...`, `pnpm --filter @multica/views exec vitest run editor/use-download-attachment.test.tsx` (17 passed), `pnpm --filter @multica/core exec vitest run api/client.test.ts` (69 passed), and `@multica/core` / `@multica/views` typecheck — all green.

For the record on the earlier head `3c2a09e5`: `TestGetAttachmentByID_AutoPublicEndpointReturnsPresignedDownloadURL` failed there (`presign calls = [downloads/desktop-inline.png downloads/desktop-inline.png], want [...]`) and passed on the base commit, so the regression was real. `c6008d2` resolves it, and as a bonus gives the presign branch genuine coverage of the new field. Worth knowing for next time: `internal/handler`'s `TestMain` prints `Skipping tests: database not reachable` and then exits `ok` when it cannot reach a database, so a DB-less run reads as a pass.

Two things I would like changed before merge, then some optional nits.

## 1. The CloudFront branch has no test — and it is the branch that changed the most

In `server/internal/handler/file.go`, the new `case h.CFSigner != nil:` is inserted as the *first* arm of that switch. Before this PR a deployment with a CloudFront signer matched no arm at all, so this is the one storage mode where the switch's control flow genuinely changed rather than just gaining a sibling. It is also the only one of the three still unexercised: the new end-to-end test covers proxy, and `c6008d2` now covers presign.

Please add a CloudFront-mode case asserting that `attachment_download_url` is signed and carries `response-content-disposition=attachment...`, and that the disposition is folded into the signed Resource rather than appended after signing (that property is the whole reason a client cannot strip it, and nothing currently pins it at this layer). `TestDownloadAttachment_CloudFrontRedirectSignsAttachmentDisposition` is a good template, and `testCloudFrontSigner(t)` already exists.

## 2. `attachment_download_url` ships in list responses as an empty string

`AttachmentResponse.AttachmentDownloadURL` has no `omitempty`, so every attachment in every list response now carries `"attachment_download_url":""`. That contradicts the field's own doc comment — "emitted ONLY by the single-attachment endpoint (GetAttachmentByID), never in list responses" — and attachment list payload size is something we deliberately tuned in #5999.

```go
AttachmentDownloadURL string `json:"attachment_download_url,omitempty"`
```

The client already treats empty and absent identically (`.optional()` plus the `||` fallbacks), so this is a one-word change that makes the contract match the comment.

## Nits (non-blocking)

- **Branch condition.** The new arm keys off `h.CFSigner != nil` while the switch itself is keyed on `mode`. With an explicit `AttachmentDownloadMode=proxy`/`presign` *and* a CF signer configured, `attachment_download_url` still points at CloudFront. That mirrors what `attachmentToResponse` already does for `DownloadURL`, so it is not a regression — but `case mode == attachmentDownloadModeCloudFront:` would match the switch's own semantics and let the explicit override actually take effect.
- **Is signing the intent worth its weight?** The domain separation is implemented and tested cleanly, but I could not construct a threat it stops: whoever holds the load-intent link already has the bytes, and flipping it to `attachment` only makes it *less* dangerous — `inline` is the risky side — and both links are minted in the same response anyway. Dropping it would remove `signAttachmentCapabilityIntent`, the fourth `verifyAttachmentCapability` parameter, and a test. Happy for it to stay if it is deliberate headroom for future intents; in that case please say so in the comment, since the current one argues from a threat model that does not quite hold.
- **`attachmentToDraftUpload`** (`packages/core/drafts/draft-upload.ts`) deliberately blanks `download_url` before a draft is persisted, precisely because it is a short-lived signed URL. The new field is the same class of value and is not blanked. It cannot bite today — only `GetAttachmentByID` mints it, and drafts hold upload responses — but it is a trap for whoever adds the field to a second endpoint.
- **Mobile** still opens `download_url` directly (`apps/mobile/components/issue/comment-attachment-list.tsx`), so this fix is web + desktop. Correct as scoped; flagging it as a separate follow-up rather than something to grow this PR with.
- **Naming debt.** `download_url` now means "load intent" and only the comments say so. Whoever wires the next download affordance will reach for it and re-hit this exact bug. Not this PR's job, but a rename or an intent parameter is the eventual fix.

Things worth keeping: load-intent signatures staying byte-identical *with a test pinning it*, the assertion that a 60-second capability never reaches `markdown_url`, and comments that explain why rather than what — that pays off in this particular file.

ADanMan:
Thanks for the thorough review — all addressed in `8414a0cfa` (CI green: `backend`, `frontend*`, `mobile`, `windows-execenv`, installers all pass).

**Blocking**

1. **CloudFront test.** Added `TestGetAttachmentByID_CloudFrontModeSignsForcedAttachmentDownloadURL`. It seeds a CloudFront-signed deployment and asserts `attachment_download_url` is CloudFront-signed (`Key-Pair-Id=KTEST` + `Signature`) and carries `response-content-disposition=attachment; filename="..."`, which `SignedURLWithContentDisposition` sets on the URL *before* signing, so it is inside the signed Resource and a client cannot strip or alter it (that property itself is unit-tested in `cloudfront_test.go`). It also asserts the load-intent `download_url` sibling does **not** force an attachment, so the two intents stay distinct.
2. **`omitempty`.** Added — `attachment_download_url` no longer ships as `""` in list responses, matching the field's own doc comment and the #5999 payload tuning.

**Nits**

- **Branch condition.** Reworked the switch to key on the resolved `mode` (`case attachmentDownloadModeCloudFront:` / `Presign` / `Proxy`) instead of `h.CFSigner != nil`, so an explicit `proxy`/`presign` override now takes effect even when a signer is configured. The CloudFront arm is nil-guarded, so an explicit `cloudfront` mode without a configured signer leaves the field empty (client falls back to `download_url`) rather than panicking.
- **Is signing the intent worth it?** You're right that flipping today's two intents isn't exploitable — `inline` is the risky disposition and both links are minted in the same response. I kept the domain separation but rewrote the comment to say exactly that: it's deliberate headroom for a future, more-privileged intent, not a current mitigation. Happy to drop it entirely instead if you'd rather — it's a small, self-contained removal (`signAttachmentCapabilityIntent`, the fourth `verifyAttachmentCapability` param, and one test).
- **`attachmentToDraftUpload`.** Now blanks `attachment_download_url` alongside `download_url`, so the short-lived signed URL isn't persisted into a draft.
- **Mobile + naming debt.** Agree both are separate follow-ups — left `comment-attachment-list.tsx` and the `download_url`-means-load-intent rename out of this PR as you suggested.

Bohan-J (approved):
Thanks for the quick turnarounds on this — every review item is addressed and the whole thing is green.

What landed since the change request:

- `omitempty` on `attachment_download_url`, so it stops shipping as an empty string in every list response and the field now matches its own doc comment.
- `TestGetAttachmentByID_CloudFrontModeSignsForcedAttachmentDownloadURL` covers the previously unexercised CloudFront arm — including that the disposition is folded into the signed Resource rather than appended after signing, which was the property I actually cared about.
- The switch is keyed on the resolved mode instead of `h.CFSigner != nil`, with a nil-signer guard on the CloudFront arm.
- `attachmentToDraftUpload` blanks the new field alongside `download_url`.
- The intent domain-separation comment now describes itself as forward-looking headroom rather than a current mitigation. Thanks for reframing it rather than deleting it — an honest comment about why something is kept is more useful than either extreme.

Verified locally at `8414a0cf` against a real PostgreSQL: the full `go test ./internal/handler`, `go test ./internal/storage/... ./internal/auth/...`, both vitest suites, and `@multica/core` / `@multica/views` typecheck — all green, and CI agrees.

One note for the follow-up backlog rather than for this PR: keying the switch on the resolved mode also changes `download_url` for a deployment that sets `ATTACHMENT_DOWNLOAD_MODE=proxy|presign` while a CloudFront signer is configured. That config used to receive a CloudFront-signed `download_url` and now receives the proxy capability / presigned URL instead. I think that is the more self-consistent behaviour — an operator who explicitly forces proxy should not be handed CDN-signed URLs — but it does reach the preview path, so it is better known now than discovered later.

Merging. Thanks again for the careful work here: the test pinning load-intent signatures as byte-identical, and the assertion that a 60-second capability never reaches `markdown_url`, are what will make this area safe to touch next time.

PR #6759, opened 2026-08-11
feat(daemon): worktree mode for local_directory resources (MUL-5707)

### LINKED ISSUE #6357: [Feature]: Support shared read locks for local directory resources
### Deployment type

Official App

## Problem

When an agent task uses a local directory resource, Multica currently appears to reserve the directory exclusively. If another task needs the same local directory, it may show a state like "waiting for local directory to be released".

This is safe, but conservative. Many agent tasks are read-only in practice, such as code search, architecture analysis, answering questions about a repository, or inspecting logs/configuration. These tasks could potentially run concurrently without blocking each other.

## Proposal

Introduce a read/write locking model for local directory resources:

- Read-only tasks acquire a shared read lock.
- Mutating tasks acquire an exclusive write lock.
- Multiple read-only tasks may run concurrently.
- Write tasks wait until all read locks are released.
- Read tasks wait while a write lock is held.

## Suggested Behavior

Default behavior can remain safe and conservative:

- Existing tasks continue to use exclusive locks by default.
- Add an explicit read-only execution mode for tasks that only inspect files.
- UI/API/agent task metadata can expose whether a task requests `read` or `write` access.
- If access mode is unknown, fallback to exclusive write lock.

## Why This Helps

This would improve concurrency for common workflows:

- Multiple agents reviewing or analyzing the same repository.
- Chat-based code Q&A while another read-only analysis is running.
- Architecture/documentation generation from source code.
- Issue triage that only inspects files.

It would also reduce user confusion around "waiting for local directory release" when the pending task does not need to modify the directory.

## Safety Considerations

A read-only mode should be strict where possible:

- Prevent file writes from agent tooling during read-only tasks, or use a sandbox/snapshot.
- Avoid running commands likely to mutate the working tree, such as dependency install, formatting, tests with generated artifacts, or git checkout.
- If mutation is needed, escalate to an exclusive write lock or fail with a clear message.
- Clearly show which task is holding the lock and whether it is read or write mode.

## Alternatives

- Keep the current directory-level exclusive lock for all tasks.
- Use per-task temporary worktrees or snapshots for read-only jobs.
- Add better diagnostics only, such as showing the blocking task and lock duration.

## Expected Outcome

Local directory resources should support safer concurrency by allowing multiple read-only agent tasks to run at the same time, while still preserving exclusive access for mutating tasks.

_No response_

_No response_

Bohan-J:
Closes #6357. MUL-5707.

## What this changes

`local_directory` resources currently serialise every task on a per-path mutex: a second task on the same folder sits in `waiting_local_directory` until the first one finishes. This adds an opt-in **worktree mode** — each task gets its own git worktree of the user's repo inside the daemon's env root, so tasks on one directory run concurrently and deliver their work as a branch instead of editing the user's working copy.

Opt in per resource with `execution_mode: "worktree"`. The default stays `in_place`, so every existing resource behaves exactly as before.

**Where you turn it on:** the resources panel (a mode dialog in the add-a-folder flow and on every local directory row), or the CLI:

```bash
multica project resource add <project-id> --type local_directory \
  --local-path /path/to/repo --daemon-id <daemon-id> --execution-mode worktree
multica project resource update <project-id> <resource-id> --execution-mode in_place
```

The UI frames the choice by what you get back — "Edit this folder directly" vs "Run in parallel, isolated" — rather than by the mode identifiers, which stay visible as a secondary hint.

## Why worktree, not shared read locks

The issue asked for shared read locks, but "read-only" isn't enforceable today: every agent runs with full write permissions, and the runtime itself writes sidecars into the user's directory during prepare. CoW snapshots would sidestep the contract but are filesystem-bound (ext4 has no reflink, overlayfs needs mount privileges a user-space daemon lacks, NTFS is out). Worktrees only need git, cover read *and* write concurrency, and reuse the model `github_repo` resources already use.

## The guarantees, and how each is held

- **The agent starts from what you see, not `HEAD`.** Tracked edits are captured with `git stash create` (which writes a commit object without touching the index, working tree, or stash list) and replayed; untracked non-ignored files are copied. If that state cannot be reproduced faithfully — bounds exceeded (2000 files / 200 MiB), an untracked symlink, a file that vanished mid-snapshot — the prepare **fails and tears the partial worktree down**, rather than handing the agent a tree the user would not recognise.
- **The replayed state is committed as a baseline**, so `git diff <baseline>..<branch>` is precisely the agent's work and a read-only run on a dirty repo still counts as a no-op.
- **Nothing is silently discarded.** Leftovers are committed to the branch before the worktree is removed, including when the task fails partway. If that commit cannot be made (e.g. `commit.gpgSign` with no key), the worktree is **kept** and the task fails naming the preserved path. If the runtime's own sidecar cleanup fails, the commit is **refused** — delivering a branch full of Multica's runtime files is the leak this mode exists to prevent.
- **A task that changes nothing leaves no branch behind.**
- **Isolation cannot be silently downgraded.** The daemon version is gated at save time *and* again against the runtime that actually claims each task, so a machine downgraded after the resource was saved has its tasks cancelled with a reason instead of running in place. A daemon that meets an unrecognised mode fails the task rather than falling back.
- **Env roots are not GC-exempt** in this mode, so they age out normally instead of accumulating one per task.

## Delivery

`branch_name` now travels end to end — it previously vanished at decode, since `TaskCompleteRequest` had no such field — on **both** terminal callbacks, into a new `agent_task_queue.branch_name` column, and out to a copyable row in the run's **Run details** panel. The failure path carries it too: a run that died partway still committed what the agent produced.

## Testing

- `server/internal/daemon/execenv/local_worktree_test.go` runs against real temp git repos: dirty tracked + untracked replay; the user's tree, index and stash list untouched; auto-commit of leftovers; branch dropped for a no-op; baseline separated from agent work; subdirectory resources; non-git and no-commit rejection; 4 concurrent tasks; sidecar-free delivered branch; stale-registration pruning; commit-failure preservation; abort refuses to commit and keeps the worktree; untracked symlink fails closed; Discard cleans up a rolled-back prepare.
- `server/internal/handler/worktree_claim_gate_test.go`: the claim-time version gate (below floor, missing version, at/above floor, in_place and absent modes, another machine's resource, non-local_directory, no daemon id, malformed ref) and a `branch_name` round-trip through both terminal callbacks.
- Frontend: 7 mode-dialog tests, 2 branch-row tests, locale parity across 4 languages.

```
go build ./... && go vet ./...                                   OK
go test -race ./internal/daemon/...                              PASS
go test ./internal/handler/ ./internal/service/                  PASS
pnpm typecheck / pnpm lint (0 errors) / pnpm test                PASS
```

The handler and service suites pass here against a migrated local database; earlier review rounds hit a stale test schema, which `make migrate-up` resolves (this PR adds migration 308).

## Not in scope

Plain non-git directories (still `in_place`), and the shadow-git idea for turning an ordinary folder into a git repo.


Bohan-J:
Pushed fixes for all six must-fix items, plus the rebase onto `main`.

**1. Preserve work when the final commit fails.** Confirmed — this was the worst one. `commitAll` returned a bare `false` and `removeLocalWorktreeDir` ran anyway, so a `commit.gpgSign` repo with no key available to the daemon would silently destroy the agent's changes (`--no-verify` doesn't disable signing). `Finalize` now returns an error and deliberately *keeps* the worktree; because it stays registered, `git worktree list` in the user's repo points straight at the work. The daemon fails the task rather than reporting success with a branch that's missing the changes — a "completed" task whose env root the GC reclaims on a timer was the real trap. Test: `TestFinalizeKeepsWorktreeWhenCommitFails`.

**2. Fail closed on an unsupported mode.** You're right that my "costs concurrency, never correctness" comment was wrong — `execution_mode` is how a user asks for *isolation*, so falling back to `in_place` edits the working copy they asked us to stay out of. `ValidateExecutionMode` now fails the task naming the skew. Note this closes the mode-aware-daemon direction completely; a daemon that predates the field entirely can't be fixed daemon-side. See the open question at the end.

**3. Codex session continuity.** Confirmed by reading `prepareCodexSessionsDir` — the per-issue store is gated on `IsLocalDirectory`, which was keyed on `LocalWorkDir != ""`. Worktree tasks also never reuse an env root (`shouldReusePriorWorkdir` refuses any local assignment), so each turn got a fresh empty `CODEX_HOME` and the thread was silently lost. Condition now covers `LocalWorktree != nil`. `TestPrepareWorktreeModeUsesPerIssueCodexSessionStore` runs two turns on one issue with different task IDs; I verified it fails without the fix.

**4. Fail closed on an incomplete snapshot.** Agreed that a daemon warning isn't a user-visible report. Failed `stash create`, failed `stash apply`, failed untracked copy, and hitting the copy bounds now all abort the prepare and tear the partial worktree down. A half-replayed tree is the worst available outcome: it looks like a working checkout, so nothing downstream questions it while the agent reads different code than the user has. The bounds error names the way out (gitignore the untracked build output, or switch back to `in_place`). Test: `TestPrepareLocalWorktreeFailsWhenUntrackedReplayIsTruncated`.

**5. Snapshot isolation.** Both halves fixed. Worktree prepare now takes the per-path mutex *for the snapshot only* and releases it before the run, so it can't read a tree an `in_place` task is mid-way through writing, while worktree tasks still overlap each other for execution. Separately, the untracked replay now skips `.agent_context/` and `.multica/` — an in-flight or crashed in-place task leaves those in the user's tree as untracked files, and copying them would have handed this task another issue's brief and committed it to the branch. `CLAUDE.md`/`AGENTS.md` are deliberately *not* in that skip list: those are normally the user's own tracked files and only carry an injected marker block, which `CleanupRuntimeConfig` removes. Test: `TestPrepareLocalWorktreeSkipsMulticaSidecars`.

**6. Test helper.** Fixed. `requireGit` runs once at repo setup; every later git failure is `t.Fatalf`. A lost branch can no longer report as a skip.

**Rebase.** Done — the conflict was as you predicted, and both `MulticaConfigRoot` and `LocalWorktree` are kept in the `Environment` initialiser. (The checkout was shallow, which is why the first rebase attempt tried to replay unrelated upstream commits.)

## One open question I did not decide unilaterally

Item 2's *other* half — a daemon that predates `execution_mode` entirely — can't be closed from the daemon, and I didn't want to pick a design for it inside a review round. The two options:

- **Server-side version gate**, reusing the existing `MinQuickCreateCLIVersion` / `readRuntimeCLIVersion` pattern to refuse a worktree-mode task when the daemon's reported `cli_version` is too old. Needs a new "runtimes by (workspace_id, daemon_id)" query and an sqlc regen; `FindLegacyRuntimesByDaemonID` requires a `provider` so it doesn't fit.
- **Make it structurally impossible**, by expressing worktree mode as a distinct `resource_type` an old daemon simply doesn't match — it would then bind no local directory at all instead of editing the user's tree. Fail-safe by construction and needs no version negotiation, but a task silently running in an empty workdir is its own kind of confusing, and it widens the resource-type space.

Happy to do either; the first is more work but keeps the resource model as-is.

## Verification

```
go build ./...                                          OK
go vet ./internal/daemon/... ./internal/handler/... ./cmd/multica/   OK
go test -race ./internal/daemon/... ./internal/handler/            PASS
go test ./cmd/multica/ -run 'ExecutionMode|LocalDirectory'         PASS
```

18 worktree tests now, including the three new failure-path ones. The two `cmd/multica` autopilot failures you saw reproduce on a clean `main` checkout in this environment (they detect a daemon task marker in the agent's own workdir) and are unrelated. I did not re-run the full combined command against a clean test database — the migration gap you hit is in my environment too.

PR #6764, opened 2026-08-11
fix(daemon): keep the runtime owner's Reasonix deny rules in a task (MUL-6021)

Bohan-J:
## Summary

Fixes the permission widening preflight blocked the 2026-08-11 release on: the per-task `reasonix.toml` written by #6746 replaced the runtime owner's `[permissions] deny` list instead of extending it.

Reasonix resolves config as flag > project `./reasonix.toml` > user `config.toml` > defaults, so `deny = ["ask"]` in the task workdir became the *effective* deny list. An owner who had `deny = ["bash"]` in their own config got `bash` back inside every Multica task — a silent relaxation of a limit they had set, which `--workspace-only` and the ACP permission policy do not restore.

## Approach

The task config is now rendered from the owner's own `[permissions]` table with `ask` merged into `deny`:

```toml
# owner's ~/.reasonix/config.toml     # task workdir reasonix.toml
[permissions]                         [permissions]
deny = ["bash"]              →        allow = ['read']
allow = ["read"]                      deny = ['bash', 'ask']
```

- **The whole table is restated, not just `deny`.** The project file overrides what it declares, so carrying `allow` and any other key the owner set keeps their policy intact whether Reasonix layers these files per key or per table. Nothing outside `[permissions]` is copied — model, credentials and the rest stay in the owner's config, which the task still inherits.
- **Read from the same `REASONIX_HOME` the child gets**: the agent's `custom_env` override, else the daemon's own env, else `~/.reasonix`. Plumbed as `PrepareParams.ReasonixHome` / `ReuseParams.ReasonixHome`, mirroring how `HermesEnv` is passed.
- **Fail closed.** If that config cannot be read or restated faithfully — unreadable, malformed TOML, `[permissions]` not a table, or a `deny` that is not an array of strings — the daemon writes no project config at all rather than one that would drop the owner's rules. The task then runs with `ask` available, exactly as it already does when a repository ships its own `reasonix.toml`, and the warning names the config it could not restate.

Deduplication is handled: an owner who already denies `ask` gets their list back unchanged.

## Assumption worth a reviewer's eye

The owner's config is read from `$REASONIX_HOME/config.toml`, per the resolution order documented in the original comment. Nothing in this repo pins that filename, so if Reasonix names it differently the merge silently degrades to today's `deny = ["ask"]` (missing file → no owner rules to carry) rather than misbehaving.

## Verification

- `go test ./internal/daemon/... -count=1` — pass (daemon, execenv, repocache)
- `go vet ./internal/daemon/...`, `gofmt -l internal/daemon` — clean
- New regression coverage in `reasonix_permissions_test.go`: global `deny = ["bash", "config_write"]` ends up denying both plus `ask` (the case preflight asked for), `allow` survives, unrelated owner settings are not copied, an owner `ask` deny is not duplicated, and each unrestatable-config shape writes no file and records nothing in the sidecar manifest.
- Existing reasonix tests are now hermetic: they point `ReasonixHome` at a temp dir instead of reading the machine's real `~/.reasonix`.

MUL-6021

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5872,13 +5872,15 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 			defer d.unmarkActiveStore(store)
 		}
 	}
-	// Reasonix keeps credentials and config in REASONIX_HOME, which an agent's
-	// custom_env may re-point. The per-task reasonix.toml has to restate the
-	// permissions from that same home, so the deny rules the runtime owner set
-	// there survive the task-scoped config that overrides them.
-	var reasonixHome string
+	// Reasonix locates its user config from the environment (REASONIX_HOME, and
+	// the platform config dirs behind it), which an agent's custom_env may
+	// re-point or clear. The per-task reasonix.toml has to restate the
+	// permissions from whichever config the child ends up loading, so the deny
+	// rules the runtime owner set there survive the task-scoped config that
+	// overrides them — hence the same sanitized env the child is launched with.
+	var reasonixEnv map[string]string
 	if provider == "reasonix" {
-		reasonixHome = sanitizeAgentEnv(agentEnvOverrides)["REASONIX_HOME"]
+		reasonixEnv = sanitizeAgentEnv(agentEnvOverrides)
 	}
 	// Guard this task's per-issue Codex session store from the GC for the whole
 	// task, starting before Prepare/Reuse mounts it — so a prune that samples the
@@ -5907,7 +5909,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 			HermesSourceMustExist: hermesSourceMustExist,
 			HermesEnv:             hermesEnv,
 			HermesMemoryStore:     hermesMemoryStore,
-			ReasonixHome:          reasonixHome,
+			ReasonixEnv:           reasonixEnv,
 			CodexCustomArgs:       codexSandboxArgs,
 			Task:                  taskCtx,
 		})
@@ -5933,7 +5935,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 			HermesSourceMustExist: hermesSourceMustExist,
 			HermesEnv:             hermesEnv,
 			HermesMemoryStore:     hermesMemoryStore,
-			ReasonixHome:          reasonixHome,
+			ReasonixEnv:           reasonixEnv,
 			CodexCustomArgs:       codexSandboxArgs,
 			Task:                  taskCtx,
 		}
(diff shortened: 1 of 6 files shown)


Bohan-J:
Pushed `927bad689` addressing the blocking finding.

`reasonix_user_config.go` now ports Reasonix's own resolution (upstream `internal/config/paths.go`) instead of assuming `$REASONIX_HOME/config.toml` with a `~/.reasonix` default:

- **Windows**: home is `%AppData%\reasonix`, falling back to `%USERPROFILE%\AppData\Roaming\reasonix` when `%AppData%` is unavailable. Env lookups are case-insensitive there, so an agent writing `APPDATA` still overrides `AppData`.
- **Expansion**: `${VAR}` / `${VAR:-default}` (unset *or empty* takes the default, matching upstream), leading `~` / `~\`, whitespace trimming, and relative→absolute normalization.
- **Legacy fallbacks**: `userConfigLoadPath` order is primary → legacy OS-support `config.toml` → legacy XDG paths → primary, including the `IsolatedHomeDir` rule that switches every fallback off once `REASONIX_HOME` is set.
- **Presence-aware env**: `ReasonixHome string` became `ReasonixEnv map[string]string` — the sanitized agent `custom_env` layered over the daemon's environment, i.e. exactly what `models.go` builds as the child env. `REASONIX_HOME=""` in `custom_env` is now an override to empty (child goes to the platform default, and so does the daemon read); a missing key inherits the daemon's variable. The same lookup feeds `AppData` / `USERPROFILE` / `HOME` / `XDG_CONFIG_HOME`, since `custom_env` can move those too.

The resolver takes `runtime.GOOS` through an overridable var, so the Windows and macOS branches are covered from any runner rather than only on `windows-execenv`.

New tests in `reasonix_user_config_test.go`: Windows default + `%AppData%`-missing fallback + `APPDATA` spelling; explicitly-cleared override vs inherited daemon value; the five expansion/normalization forms; legacy OS-support and legacy XDG fallbacks plus primary-wins; the isolation rule ignoring both legacy copies when `REASONIX_HOME` is set; and an end-to-end case where a deny list that only exists in a legacy location still reaches the task config as `['bash', 'ask']`.

Verification: `go test ./internal/daemon/... -count=1` pass, `go test ./internal/daemon/execenv/ -race -count=1` pass (the GOOS-pinning tests are serial, so they finish before parallel tests resume), `GOOS=windows go build ./...`, `go vet`, `gofmt` clean.

One related thing I did **not** touch, to keep this PR scoped: `server/internal/daemon/local_skills.go:172` resolves the Reasonix skills dir as `~/.reasonix/skills`, which carries the same Windows assumption this PR just removed for the config path. It affects local skill discovery, not permissions, so it seems worth its own change.

PR #6767, opened 2026-08-11
fix(workflows): one server-owned field catalog for hook conditions (FIR-4933)

hvejsel:
Closes FIR-4933 (internal tracker).

## Problem

Two field lists that had to agree, kept apart by hand:

- The hook editor's suggestion list was a hand-written copy. It **missed** fields the platform's own managed hooks match on — so those hooks rendered as broken rows with `<redacted>` values (FIR-4933's screenshots) — and it **offered** fields the server gates never send (`actor.type`, `message.channel_id`, `message.body`, `wakeup.fire_at`, `retry.pending`, ...), letting users build rules that look fine but can never fire, with no warning either way.
- The hook journal's retention allowlist was a third vocabulary that dropped exactly the fields those conditions use (whole `message` section, `continuation`, `chain`, `failure`, the `wakeup` keys), so replaying a journal event through Test almost never matched — and `RecordTestEvidence` needs a match, so conditional hooks could not clear the publish gate that way. `before.issue.assigned` was not retainable at all.

## Fix

One server-owned manifest — `server/internal/cerebro/workflows/hook_field_manifest.json` — declares, per condition-visible event, each field's path, label, input widget and sensitivity. `go generate` emits `hook-field-catalog.generated.ts`; `hook-validation` and `hook-ux` read the generated catalog instead of hand-written lists. Same provenance pattern as the action catalog (`hook-action-catalog.generated.ts`).

Behaviour changes:

- A condition field outside the manifest renders with an explicit **`(unknown field)`** marker instead of the old "<redacted>"-as-unfamiliar behaviour — honesty, not borrowed security semantics. Known fields redact only when the manifest marks them sensitive.
- `event.Actor` is merged into the condition context, so `actor.type` / `actor.id` resolve on every event (one-line engine change).
- Journal retention now keeps condition-visible fields and `before.issue.assigned` events, so Test/replay and publish evidence work on these hooks.
- `before.issue.assigned` is selectable in the editor.
- Templates repaired: `require-continuation` uses `continuation.present`; `no-silent-failure` uses `failure.attempt >= $event.failure.max_attempts` (`retry.pending` never existed server-side).

Drift locks so the gap cannot reopen:

- Go test: the generated TS must byte-match the manifest (mirror of the action-catalog test).
- Go test: every managed policy condition and every `$event.*` value must resolve in the manifest for its events.
- Go test: the engine's base condition context may not grow fields the manifest does not declare.

## Scope notes

Not in this PR (tracked as follow-ups from the FIR-4933 review): `before.issue.status_change` firing only for issues with an active chain, project/workflow bindings silently never matching on `before.message.send`, `fail_mode` open→warn normalisation on save, the 2s catalog-timeout fail-open on `CanBlock` events, and removing structurally-always-true managed conditions (`chain.active eq true`, `issue.terminal eq false`).

## Test plan

- `go test ./internal/cerebro/workflows/` — green (new manifest/drift tests included).
- `go build ./...` — green.
- `pnpm vitest run` in `packages/cerebro-workflows` — 154 tests green (incl. updated UX-contract expectations for the new `(unknown field)` marker and boolean Yes/No rendering).


hvejsel:
Opened against the wrong base — this branch is cut from firtal-group/firtal-cerebro main. Correct PR will live there.

PR #6771, opened 2026-08-11
MUL-6029: docs(daemon): correct two comments #6604 made false, and enforce the Slack/Lark brief policy

seacen:
Three follow-ups to #6604, all landed 12 minutes after it merged and so missed the boat. Two are comments the merged diff made false; one is a policy that PR states in a comment but does not enforce.

## `handler.go:292` says the brief does something it no longer does

The `channelFileDelivery` doc says `chat_channel_delivers_files` is what "the agent's brief turns into either `run multica attachment upload` or `describe the file in words`". After #6604 the brief turns it into neither — nothing in `execenv` reads the field, and both strings are produced only by the per-turn prompt (`prompt.go:565-569`).

It survived that PR's comment sweep because the sweep grepped the field in camelCase and this line spells it snake_case. The correction names the prompt and records why the brief cannot be the one to say it (the brief is the cache prefix; a per-turn verdict there renders two briefs — MUL-5377).

## `outbound_media.go:531` cites a test that does not show what the sentence claims

The comment read: "A backlog of turns carrying no file reaches this cap with nothing pending (`TestDeliverAttachments_AdmissionBoundsTheLookupStage`)."

The claim is true — `sendAttachments` returns at `if len(rows) == 0` before ever claiming a pending slot — but that test's turns all carry a file. They pile up at the admitted cap because the fixture parks every goroutine in `q.lookupGate`. The paragraph now states both ways admission is held without a pending slot and hangs the citation on the one the test actually drives.

## Slack and Lark deny two phrases where WeCom denies four

`TestBriefSurfaceDeliveryPolicy`'s `chat_slack` and `chat_feishu` rows carry a two-phrase `mustNot`. The comment above them is a policy: a brief must never carry the Slack/Lark denial, because wiring that hop is a server-side change and a brief holding the denial would keep denying it afterwards.

That policy was not enforceable. A denial worded around "does NOT apply" or "separate message" passed both rows, and `TestBriefByteIdenticalAcrossRunsForEveryKind` cannot see it either — written unconditionally it renders one brief on both values of the verdict, so the invariant holds. Only a phrase list catches it. Both rows now carry the same four phrases the WeCom rows do.

Verified by injecting a Slack denial worded only as "That guidance does NOT apply to Slack": two failures with the wider list, none with the old one.

## Verification

`go build ./...`, `go vet`, and `go test ./internal/daemon/execenv/ ./internal/handler/ ./internal/integrations/wecom/` all clean on this branch. No behaviour change — two comments and one test row.

### DIFF
--- a/server/internal/daemon/execenv/runtime_config_delivery_test.go
+++ b/server/internal/daemon/execenv/runtime_config_delivery_test.go
@@ -152,13 +152,19 @@ func TestBriefSurfaceDeliveryPolicy(t *testing.T) {
 		// what makes the paragraph above a rule rather than a note. The
 		// byte-identity invariant cannot enforce it on its own: a denial
 		// written into the brief unconditionally is the same on both values of
-		// the verdict, so it renders one brief and passes. Only a phrase list
-		// catches it, and only if it names the wordings a denial actually
-		// arrives in.
+		// the verdict, so it renders one brief and passes.
+		//
+		// A phrase list cannot be the only guard either — it denies the
+		// wordings someone thought to name, and the next one will be worded
+		// differently. TestBriefChannelDeliveryCopyIsPlatformNeutral holds the
+		// same rule with no wording in it, by requiring every channel-backed
+		// brief to differ only in the platform it names. Keep both: that one
+		// catches any wording, these fail beside the copy being edited.
 		"chat_slack": {
 			mustHave: []string{
 				"Slack conversation depends on how this deployment is configured",
 				"the per-turn user message tells you",
+				"never report a file as delivered",
 			},
 			mustNot: []string{
 				"run `multica attachment upload",
@@ -171,6 +177,7 @@ func TestBriefSurfaceDeliveryPolicy(t *testing.T) {
 			mustHave: []string{
 				"Feishu/Lark conversation depends on how this deployment is configured",
 				"the per-turn user message tells you",
+				"never report a file as delivered",
 			},
 			mustNot: []string{
 				"run `multica attachment upload",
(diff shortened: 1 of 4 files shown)


Bohan-J:
Thanks for this — and especially for chasing down *why* the sweep in #6604 missed the `handler.go` line instead of just fixing the text. Follow-ups that explain how the miss happened are worth more than the correction itself.

I verified each claim independently rather than taking the description's word for it:

- **`handler.go:292`** — confirmed nothing in `execenv` renders the verdict into the brief (`execenv.go` and `runtime_config_sections.go` both state it explicitly), and the switch in `prompt.go` is the sole producer of both strings. Leaving the reference as `daemon/prompt.go` without a line number was the right call.
- **`outbound_media.go:531`** — the old sentence's *conclusion* held (`sendAttachments` returns at `len(rows) == 0` before `claimAttachmentSlot()`), but the cited test genuinely doesn't drive it: `oneAttachmentQueries` gives every turn a file, and they pile up because the fixture parks them in the lookup gate. Accurate diagnosis.
- **The `mustNot` asymmetry** — reproduced your mutation check: a denial worded only as "That guidance does NOT apply to Slack" passes the old two-phrase list and fails the widened one.

Rather than send you round again for review nits, I pushed one follow-up commit (`8dfe8cc`) to your branch. Tests only, no production code:

1. **`TestBriefChannelDeliveryCopyIsPlatformNeutral`** — the structural form of the rule your comment states. A phrase list can only deny the wordings someone thought to name; this substitutes the display name out and requires every channel-backed brief to be byte-identical, so a per-platform position fails whatever words it is written in. Verified strictly stronger: a Slack-only `"Files cannot reach this room."` passes all four of your phrases and fails this. Your phrase rows stay — they fail beside the copy being edited, which a whole-brief diff doesn't.
2. `chat_slack` / `chat_feishu` `mustHave` picks up `"never report a file as delivered"` for parity with the WeCom row.
3. `mustNot` failures now quote the context around the actual hit and say which section it was in. The phrases are ordinary English, and the old message printed an `## Output` section the match may not even be in.
4. `pending == 0` is now asserted in `TestDeliverAttachments_AdmissionBoundsTheLookupStage` — that is the half of the ordering the constant's comment cites it for, and it was implied rather than checked.

I also rewrote the comment above those two rows, which said a phrase list was the only thing that could catch this. True when you wrote it; not once the structural test landed — and letting it stand would have recreated exactly the stale-comment defect this PR exists to fix.

Merging. Thanks again for the care here.

PR #6774, opened 2026-08-11
MUL-6032: feat(attachments): configure max upload and preview sizes via env variables

Iceyer:
## Summary
   Fix 413 Payload Too Large on attachment uploads: hard-coded size limits are now configurable via env vars.

   ## Changes
   - `MULTICA_MAX_UPLOAD_SIZE` — max upload body size (default 100 MB)
   - `MULTICA_MAX_PREVIEW_SIZE` — max inline-preview body size (default 2 MB)
   - Defaults unchanged when unset; invalid values fall back with a warning
   - Files: `server/internal/handler/file.go`, `file_test.go` (new `TestParseByteSize`), `.env.example`, `docker-compose.selfhost.yml`

   ## Verification
   - `go build ./...` passes
   - `TestParseByteSize`, `TestUploadFile`, `TestGetAttachmentContent`, `TestServeAvatar` pass


multica-eve (changes requested):
Self-hosted administrators should be able to choose how large an attachment their deployment accepts, while independently keeping inline viewing conservative. That gives teams control over infrastructure they operate themselves, without changing the default experience for everyone else. We want this product direction.

**Product decision: accepted. Implementation decision: changes required before merge.**

Thanks for contributing this. Please address these blocking issues:

1. **Make the upload limit work end to end.** The web and mobile clients still reject files above the hard-coded 100 MB limit before sending them (`packages/core/constants/upload.ts`, `packages/core/hooks/use-file-upload.ts`, `packages/views/editor/use-coordinated-uploads.ts`, and `apps/mobile/data/api.ts`). The standard self-hosted web path also proxies `/api/upload-file`, where the existing request-body limit can reject/truncate the upload before it reaches this handler. Raising `MULTICA_MAX_UPLOAD_SIZE` therefore does not currently enable the promised user journey, while lowering it leaves the UI showing the wrong limit. The effective server limit should be discoverable and consistently enforced/displayed by clients and the proxy path.

2. **Add explicit hard upper bounds and overflow-safe handling.** `parseByteSize` accepts every positive `int64`, while `GetAttachmentContent` adds one to the configured preview value and later converts it to `int`. For example, with `MULTICA_MAX_PREVIEW_SIZE=9223372036854775807`, the happy-path Markdown preview returns a successful response with an empty body because the addition overflows. Upload and preview limits need separate, documented safe maxima and boundary tests.

3. **Do not turn a larger configured limit into proportional server memory use.** `ParseMultipartForm(maxUploadSize())` uses the configured limit as its in-memory threshold, and the selected file is then fully read with `io.ReadAll`. Workspace membership is checked only after that read. A large setting plus concurrent requests can exhaust the process before authorization rejects an invalid workspace upload. Please authorize before consuming the body where possible, keep multipart memory bounded independently of the file limit, and stream file contents to storage.

4. **Add behavior-level tests.** `TestParseByteSize` only verifies parsing. Please cover the default and custom limits, exact-boundary/over-limit behavior, invalid and extreme values, client-visible effective limits, and the standard self-hosted upload path.

Also, please remove the unrelated PostgreSQL port change in `docker-compose.yml` from this PR, and update comments that still describe the preview cap as always 2 MB.

There is substantial overlap with #6490, which already attempts the end-to-end upload capability. Please coordinate with that work to avoid maintaining two competing implementations; the independently configurable preview limit is the main addition unique to this PR.

PR #6775, opened 2026-08-11
MUL-6137: fix(daemon): preserve name-dispatching agent shims

### LINKED ISSUE #6183: [Bug]: bug(daemon): Volta-managed CLIs are not detected because symlink resolution loses the command name
### Deployment type

Official App

## Environment

- macOS arm64
- Multica Desktop v0.4.14
- Volta v2.0.2
- Claude Code and Codex installed through Volta

## Problem

Multica cannot register Volta-managed runtimes such as Claude, Codex, and Pi.

The daemon resolves the CLI symlink to Volta's shared `volta-shim`, then runs:

```sh
/Users/<user>/.volta/bin/volta-shim --version
That returns exit code 126 because Volta relies on the invoked command name (claude, codex, pi) to decide which package to run.
As a result, these runtimes are skipped:
WRN skip registering runtime component=daemon name=claude attempts=2 error="detect version for /Users/<user>/.volta/bin/volta-shim: exit status 126"
WRN skip registering runtime component=daemon name=codex attempts=2 error="detect version for /Users/<user>/.volta/bin/volta-shim: exit status 126"
WRN skip registering runtime component=daemon name=pi attempts=2 error="detect version for /Users/<user>/.volta/bin/volta-shim: exit status 126"
Reproduction
volta install @anthropic-ai/claude-code @openai/codex @earendil-works/pi-coding-agent

which claude
# ~/.volta/bin/claude

readlink ~/.volta/bin/claude
# ~/.volta/bin/volta-shim

claude --version
# succeeds

~/.volta/bin/volta-shim --version
# exits 126
Start Multica Desktop or multica daemon start. Claude/Codex/Pi are not registered.
Root cause
resolveAgentExecutablePath() calls canonicalExecutablePath(), which uses filepath.EvalSymlinks().
For Volta, this converts every managed command into the same shared volta-shim path. The subsequent version probe loses the original command name and cannot work.
Expected behavior
Multica should detect and register Volta-managed CLIs normally.
Suggested fix
Do not resolve a CLI symlink when its resolved target is Volta's shared volta-shim; preserve the original command path for execution/version probes.
Alternatively, resolve the concrete executable through Volta, for example:
volta which claude
volta which codex
Workaround
Setting explicit paths works:
MULTICA_CLAUDE_PATH="$(volta which claude)"
MULTICA_CODEX_PATH="$(volta which codex)"
MULTICA_PI_PATH="$(volta which pi)"

### Steps to reproduce

1. Install Volta on macOS.

2. Install Claude Code and Codex through Volta:

   volta install @anthropic-ai/claude-code @openai/codex

3. Verify that both CLIs work in Terminal:

   claude --version
   codex --version

4. Start Multica Desktop v0.4.14, or run:

   multica daemon start

5. Open the daemon log.

6. Observe that Claude and Codex are skipped with:

   detect version for ~/.volta/bin/volta-shim: exit status 126

### Screenshots (optional)

_No response_

```shell
Expected: Claude and Codex runtimes are detected and registered.

Actual: Multica resolves both commands to Volta’s shared volta-shim and fails to detect their versions.
```

### LINKED ISSUE #6702: [Bug]: Vite Plus-managed Claude is misdetected as vp v0.2.5
### Deployment type

Official App

### What happened?

Multica Desktop resolves a Vite Plus-managed `claude` symlink incorrectly during runtime version detection. The installed Claude Code version is `2.1.216`, but the daemon detects the Vite Plus manager version `vp v0.2.5`, treats Claude as older than the required `2.0.0`, and skips registering the Claude runtime.

Expected: Multica probes the Claude Code executable and registers version `2.1.216`.

Actual daemon log:

```text
WRN skip registering runtime: version too old component=daemon name=claude version="vp v0.2.5" error="claude version vp v0.2.5 is below minimum required 2.0.0 — please upgrade"
```

The stable command is a symlink managed by Vite Plus:

```text
/Users/<user>/.local/bin/claude -> /Users/<user>/.vite-plus/bin/claude
```

Both the stable path and its target report the correct Claude Code version when invoked directly:

```text
$ ~/.local/bin/claude --version
2.1.216 (Claude Code)

$ ~/.vite-plus/bin/claude --version
2.1.216 (Claude Code)
```

Explicitly setting the provider path fixes the issue:

```sh
launchctl setenv MULTICA_CLAUDE_PATH "$HOME/.vite-plus/bin/claude"
```

After restarting Multica Desktop, the daemon detects and registers Claude correctly:

```text
DBG agent version detected component=daemon name=claude version="2.1.216 (Claude Code)" path=/Users/<user>/.vite-plus/bin/claude
INF registered runtime component=daemon provider=claude
```

This appears related to #6183: both involve runtime detection resolving a tool-manager-managed command incorrectly and probing the manager/shim rather than preserving the intended command semantics. #6183 covers Volta; this report covers Vite Plus and produces a different failure signature (`vp v0.2.5`).

#6452 is also related to the broader behavior where failed/incorrect version detection takes a runtime offline, but the direct cause here is deterministic misidentification rather than a probe timeout.

### Steps to reproduce

1. On macOS arm64, install/manage Claude Code through Vite Plus so that `~/.local/bin/claude` points to `~/.vite-plus/bin/claude`.
2. Verify `~/.local/bin/claude --version` prints `2.1.216 (Claude Code)`.
3. Start Multica Desktop without setting `MULTICA_CLAUDE_PATH`.
4. Open the daemon log at `~/.multica/profiles/desktop-api.multica.ai/daemon.log`.
5. Observe that Claude is skipped because its detected version is `vp v0.2.5`.
6. Set `MULTICA_CLAUDE_PATH` to the Vite Plus Claude target and restart Desktop.
7. Observe that Claude `2.1.216` is detected and the runtime registers successfully.

### Screenshots (optional)

Not applicable; daemon logs and CLI output are included above.

```shell
OS: macOS arm64 (Darwin 23.5.0)
Multica Desktop: 0.4.21
Multica daemon: v0.4.21 (commit 0dfaac266)
Claude Code: 2.1.216
Vite Plus: vp v0.2.5

Symlink:
~/.local/bin/claude -> ~/.vite-plus/bin/claude

Workaround confirmed:
MULTICA_CLAUDE_PATH=~/.vite-plus/bin/claude

Related issues:
- #6183 — Volta symlink resolution loses command identity
- #6452 — version probe failure skips runtime registration
```

weivwang:
## What does this PR do?

Keeps the original agent command entrypoint when it resolves to a known name-dispatching shim (`volta-shim` or Vite Plus's `vp`). Those managers choose the managed package from the invoked command name, so spawning the fully resolved target changes `claude --version` into `volta-shim --version` / `vp --version` and prevents runtime registration.

Ordinary executable symlinks are still canonicalized to their concrete target, preserving the daemon's existing path-pinning and in-place-upgrade self-heal behavior. For dispatch shims, the parent directory is still canonicalized while the command basename is retained.

This also aligns the `exec.LookPath` discovery path with the existing login-shell fallback. `buildLoginShellResolveScript` already canonicalizes only the containing directory (`pwd -P`) and preserves the original executable basename, so it has always been safe for name-dispatching shims; this PR gives direct PATH discovery the same semantics.

### Thinking path

1. Runtime discovery is intentionally split into `exec.LookPath`, path canonicalization, then version detection.
2. Removing canonicalization wholesale would regress the anti-PATH-redirect guarantee and Homebrew/nvm/fnm upgrade recovery.
3. The failure is specific to managers whose final executable is a shared dispatcher and whose behavior depends on the invoked basename.
4. Therefore this change preserves only the two confirmed dispatcher targets and keeps the existing canonical path for every other executable.

## Related Issue

Closes #6183
Closes #6702

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added invocation-safe agent path canonicalization in `server/internal/daemon/config.go`.
- Preserved `claude`/`codex`-style entrypoint names for Volta and Vite Plus dispatchers.
- Added red/green regression coverage for both dispatcher names and a guard proving ordinary differently named version targets remain fully canonicalized.
- Kept one canonicalization implementation by delegating invocation-safe resolution to `canonicalExecutablePath`.
- Documented the exact-match `vp` admission and the managers' wrapper/trampoline behavior on Windows; executable extensions are intentionally not treated as Unix-style dispatch symlinks.
- Updated nearby resolver comments to document the invocation-name constraint.

## How to Test

1. Run `go test ./internal/daemon -run 'TestResolveAgentExecutablePath_' -count=1 -v`.
2. Run `go test -race ./internal/daemon -count=1`.
3. Run `go vet ./internal/daemon`.

The dispatcher test uses a test-created executable that prints its observed invocation basename. It proves the resolved path still invokes it as `claude`, while the ordinary-symlink test proves normal version targets remain pinned to their final path.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (N/A: daemon-only change)
- [x] I have updated relevant documentation to reflect my changes (N/A: no user-facing contract; resolver comments updated)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (N/A)
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (terminology, mixed-rule for `task` / `issue` / `skill`) (N/A)

## Risks

- The preservation rule is deliberately limited to the exact confirmed Unix dispatcher basenames `volta-shim` and `vp`, so unrelated symlinks keep existing semantics. On Windows, Volta uses `.cmd` wrappers and Vite Plus uses trampoline executables, which preserve the tool identity without this symlink branch.
- Explicit `MULTICA_*_PATH` values still bypass canonicalization exactly as before.
- The hooks-directory exclusion now applies the same dispatcher-aware canonicalization to the real command it finds.

## AI Disclosure

**Prompt / approach:** Scanned open issues and competing PRs, selected a high-impact unclaimed runtime-discovery bug, traced `exec.LookPath` through `filepath.EvalSymlinks` and version probing, wrote a failing fake-dispatcher regression test first, then implemented the narrow dispatcher-aware canonicalization and ran the daemon test suite with the race detector plus `go vet`.

## Screenshots (optional)

N/A — daemon-only change with no UI output.

### DIFF
--- a/server/internal/daemon/canonical_path.go
+++ b/server/internal/daemon/canonical_path.go
@@ -2,14 +2,52 @@
 
 package daemon
 
-import "path/filepath"
+import (
+	"path/filepath"
+	"strings"
+)
 
 func canonicalPath(path string) (string, error) {
 	return filepath.EvalSymlinks(path)
 }
 
+// discoveredExecutablePath canonicalizes ordinary executable symlinks but
+// preserves entrypoints backed by a known name-dispatching shim. Volta's
+// volta-shim and Vite Plus's vp select the managed package from the invoked
+// name; spawning the final target directly turns `claude --version` into
+// `volta-shim --version` or `vp --version`, so the daemon reads the manager's
+// version instead of the agent's and drops the runtime (#6183, #6702).
+//
+// The entrypoint's parent is still canonicalized. This keeps paths stable when
+// their containing directory is a symlink or an ephemeral version-manager
+// prefix, while retaining the basename the dispatcher needs — the same
+// semantics buildLoginShellResolveScript has always applied via `pwd -P`.
 func discoveredExecutablePath(path string) string {
-	return canonicalExecutablePath(path)
+	real := canonicalExecutablePath(path)
+	if !isNameDispatchingAgentShim(real) {
+		return real
+	}
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return real
+	}
+	realDir, err := filepath.EvalSymlinks(filepath.Dir(abs))
+	if err != nil {
+		return abs
+	}
+	return filepath.Join(realDir, filepath.Base(abs))
+}
+
+func isNameDispatchingAgentShim(path string) bool {
+	base := strings.ToLower(filepath.Base(path))
+	// "vp" is intentionally admitted despite being a short, generic token:
+	// it is Vite Plus's shared dispatcher and, like Volta, selects the managed
+	// package from argv[0]. Keep this exact-match list limited to dispatchers
+	// confirmed to require the invoked entrypoint name. Do not strip executable
+	// extensions: these managers use wrappers or trampolines rather than
+	// name-dispatching symlinks, and that is a Windows shape this file never
+	// compiles for.
+	return base == "volta-shim" || base == "vp"
 }
 
 var executablePathForLaunch = executablePathForLaunchDefault
--- a/server/internal/daemon/canonical_path_test.go
+++ b/server/internal/daemon/canonical_path_test.go
@@ -1,0 +1,32 @@
+//go:build !windows
+
+package daemon
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestIsNameDispatchingAgentShim_RequiresExactDispatcherName(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{name: "Volta", path: filepath.Join("manager", "volta-shim"), want: true},
+		{name: "Vite Plus", path: filepath.Join("manager", "vp"), want: true},
+		{name: "case insensitive", path: filepath.Join("manager", "VP"), want: true},
+		{name: "short name prefix", path: filepath.Join("manager", "vpn"), want: false},
+		{name: "short name word", path: filepath.Join("manager", "vproxy"), want: false},
+		{name: "extension is not stripped", path: filepath.Join("manager", "volta-shim.exe"), want: false},
+		{name: "ordinary version target", path: filepath.Join("manager", "claude-2.1.216"), want: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := isNameDispatchingAgentShim(tt.path); got != tt.want {
+				t.Fatalf("isNameDispatchingAgentShim(%q) = %v, want %v", tt.path, got, tt.want)
+			}
+		})
+	}
+}
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -642,6 +642,9 @@ func shellArgsFromEnv(name string) ([]string, error) {
 // resolveAgentExecutablePath returns the executable entry point the daemon
 // should keep for an agent command. Bare command names are pinned to the path
 // resolved during startup so later PATH changes cannot redirect task launches.
+// Ordinary executables are pinned to their concrete target; entrypoints owned
+// by a name-dispatching shim keep the command name the manager needs to select
+// the right package (see discoveredExecutablePath).
 // On Windows this deliberately keeps the stable discovered junction path;
 // resolveAgentEntryWithHeal follows it for each launch so installer upgrades
 // that retarget a still-live junction take effect without a daemon restart.
@@ -843,8 +846,8 @@ var supportedLoginShells = map[string]struct{}{
 	"ksh":  {},
 }
 
-// resolveAgentsViaLoginShell asks the user's login shell to print the canonical
-// (symlink-resolved) absolute path to each name in `names`. It returns a map
+// resolveAgentsViaLoginShell asks the user's login shell to print an absolute,
+// invocation-safe path to each name in `names`. It returns a map
 // of name → path for whatever the shell could find, and an empty map if the
 // shell is unavailable / unsupported / times out / produces no usable output.
 //
@@ -945,7 +948,8 @@ var resolveAgentsViaLoginShell = func(names []string) map[string]string {
 //     still print the alias/function definition, and we'd rather drop it
 //     than hand back garbage),
 //  5. canonicalises the directory via `cd ... && pwd -P` so symlinked prefix
-//     dirs (fnm/nvm/volta) collapse to stable paths,
+//     dirs (fnm/nvm/volta) collapse to stable paths while the invoked command
+//     name is kept,
 //  6. if the resolved path lives in ~/.multica/hooks, searches the same
 //     shell-expanded PATH for the first executable outside that hooks dir,
 //  7. prints `<name>\t<canonical_path>` one entry per line for the caller.
--- a/server/internal/daemon/config_test.go
+++ b/server/internal/daemon/config_test.go
@@ -15,6 +15,80 @@ import (
 	"github.com/multica-ai/multica/server/internal/cli"
 )
 
+func TestResolveAgentExecutablePath_PreservesDispatchShimName(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("creating symlinks requires elevated privileges on Windows")
+	}
+
+	for _, shimName := range []string{"volta-shim", "vp"} {
+		t.Run(shimName, func(t *testing.T) {
+			managerDir := t.TempDir()
+			manager := filepath.Join(managerDir, shimName)
+			if err := os.WriteFile(manager, []byte("#!/bin/sh\nprintf '%s\\n' \"${0##*/}\"\n"), 0o755); err != nil {
+				t.Fatalf("write dispatcher: %v", err)
+			}
+
+			binDir := t.TempDir()
+			entrypoint := filepath.Join(binDir, "claude")
+			if err := os.Symlink(manager, entrypoint); err != nil {
+				t.Fatalf("symlink dispatcher: %v", err)
+			}
+			t.Setenv("PATH", binDir)
+
+			got, err := resolveAgentExecutablePath("claude")
+			if err != nil {
+				t.Fatalf("resolveAgentExecutablePath: %v", err)
+			}
+			realBinDir, err := filepath.EvalSymlinks(binDir)
+			if err != nil {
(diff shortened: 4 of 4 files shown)


ADanMan:
Nice fix — the diagnosis is right and the shape is minimal. A few notes, one worth a quick correction before merge.

**What's solid**
- Both production call sites of `canonicalExecutablePath` in `resolveAgentExecutablePath` are converted to `invocationSafeExecutablePath`, so there's no straggler path.
- `TestResolveAgentExecutablePath_CanonicalizesOrdinaryVersionTarget` is the right guard to pair with the shim test — it pins the MUL-4486 property (ordinary versioned symlinks still resolve to the concrete target) this change could plausibly have regressed.
- Worth adding to the PR description — the best argument this is a general fix, not a special case: the login-shell resolver was *already* invocation-safe. `buildLoginShellResolveScript` does `c=$(cd "$d" && pwd -P)` and prints `"$c/$f"` — canonical directory, original basename, never following the file symlink. The two discovery paths have disagreed all along: `exec.LookPath` canonicalized the file, the shell fallback didn't. This PR makes LookPath match the resolver that was already correct.

**1. `canonicalExecutablePath` is left with only test callers, and its logic is duplicated.** After this PR its only callers are tests. That's a test-fidelity smell: those tests compute expected values with a function production no longer calls, so if `invocationSafeExecutablePath` drifts in its Abs/EvalSymlinks handling the assertions keep passing against the old implementation. `invocationSafeExecutablePath` also duplicates the same canonicalization before branching. Delegating keeps one definition of "canonical":
```go
func invocationSafeExecutablePath(path string) string {
    real := canonicalExecutablePath(path)
    if !isNameDispatchingAgentShim(real) {
        return real
    }
    abs, err := filepath.Abs(path)
    if err != nil {
        return real
    }
    realDir, err := filepath.EvalSymlinks(filepath.Dir(abs))
    if err != nil {
        return abs
    }
    return filepath.Join(realDir, filepath.Base(abs))
}
```
CLAUDE.md's "prefer existing patterns over new parallel abstractions" points the same way.

**2. `isNameDispatchingAgentShim` keys off a two-character name.** It lowercases, strips one extension, then compares the basename for equality — so `vp`/`vp.exe`/`VP` match. It's an exact match (won't catch `vpn`/`vproxy`), so the risk is narrow — but `vp` is a short generic token, and everything else in this file keys off layout (hooks dir, PATH membership) rather than a bare name. Worth corroborating with the layout (shim under a `volta`/`.vite-plus` prefix) or a one-line comment naming `vp` as a known-collision token.

**3. The `.exe` handling implies Windows support nothing exercises.** `strings.TrimSuffix(base, filepath.Ext(base))` is there so `volta-shim.exe` matches, but both new tests `t.Skip` on Windows. The branch also *appears* unreachable there — Volta on Windows seems to install shims as `.exe` copies rather than symlinks, so `EvalSymlinks` returns `claude.exe` unchanged and the predicate never fires — but that rests on Volta's Windows install mechanics I can't confirm from the tree, so treat it as a hypothesis. If it holds, a comment saying Windows is correct-by-construction helps; if not, a copy-based test (no symlink privileges needed) closes it. Could you confirm which?

**4. The reworded guarantee hedges without explaining.** "later PATH changes therefore cannot redirect normal launches" — the "normal" quietly carves out shims. It isn't a weakened guarantee: for a shim the pinned path is still absolute, so PATH redirection is still prevented — what changes is you're pinning a mutable indirection the version manager owns, which is the point. Worth stating outright.

**5. (nit)** `mise`/`asdf`/`proto` share the same argv[0]-dispatch shape; a package-level `map[string]struct{}` with a one-line admission criterion makes the next addition a data change. Non-blocking.

Nothing critical looks missed — `invocationSafeExecutablePath` preserves the invoked basename and the tests exercise the real dispatch path.

weivwang:
Addressed in ddfd59e — thanks for the detailed review.

1. `invocationSafeExecutablePath` now delegates to `canonicalExecutablePath`, so production and test callers share one canonicalization implementation.
2. The `vp` admission is documented as an intentional exact match, and a table test pins that `vpn` / `vproxy` do not collide.
3. I confirmed the current upstream layouts: Volta uses `.cmd` wrappers on Windows, while Vite Plus uses trampoline `.exe` files that pass the tool identity through `VP_SHIM_TOOL`. Neither depends on the Unix symlink/argv[0] branch, so I removed extension stripping instead of claiming unexercised `.exe` support. The cross-platform predicate test pins that decision.
4. The resolver comment now states the actual guarantee: every discovered command remains pinned to an absolute path; ordinary executables pin the concrete target, while dispatchers pin the manager-owned indirection.
5. I kept the two exact comparisons for now rather than introducing a mutable package-level map for a two-entry allowlist. The comment records the admission criterion, so a future confirmed manager remains a small data-like change without adding mutable global state.

I also updated the PR description with the login-shell resolver consistency argument. CI is green, including the full backend race suite and Windows build/runtime job.

Bohan-J:
Merged — thanks for the fix, @weivwang, and for the unusually clear write-up. Tracing this to `argv[0]`-dispatching managers rather than blaming symlink resolution in general is what made it a safe, narrow change, and the regression test that actually executes the entrypoint and asserts the observed command name is exactly the right proof.

One thing to flag since your branch was force-pushed: `main` had moved under you. #6802 / #6895 split executable discovery per platform (`discoveredExecutablePath` in `server/internal/daemon/canonical_path.go` vs `canonical_path_windows.go`), which is what made the PR conflict. Resolving it by keeping the original call sites would have overwritten the Windows behaviour those PRs added, so on rebase the dispatcher preservation moved into the non-Windows `discoveredExecutablePath` and the Windows path was left untouched. Your `isNameDispatchingAgentShim` table test also moved to a new `canonical_path_test.go` behind `//go:build !windows`, since the function no longer compiles on Windows. The behaviour on macOS/Linux — and both of your regression tests — are unchanged.

Verified before merge: dispatcher tests red without the preservation and green with it, `go test -race ./internal/daemon`, `GOOS=windows go vet ./internal/daemon`, and full CI green on the rebased commit.

Closes #6183 and #6702.

PR #6780, opened 2026-08-11
fix(execenv): make Windows text reads UTF-8 safe

Bbbtt04:
﻿## Summary

- add Windows-only runtime guidance that requires explicit UTF-8 decoding for PowerShell 5.1 text reads
- apply the guidance across Codex, Claude, and OpenCode task kinds
- test Windows inclusion and non-Windows omission

## Root cause

Windows PowerShell 5.1 can decode UTF-8 files without a BOM through the machine ANSI code page when `Get-Content` has no explicit `-Encoding`. Chinese and other non-ASCII text is therefore corrupted before command output reaches the agent adapter or transcript. The UI and backend only pass that already-corrupted string through, so attempting to repair it there would be lossy and unreliable.

## Validation

- `go test ./internal/daemon/execenv -run 'WindowsTextEncoding|BuildMetaSkillContentSlimKindMatrix|RuntimeConfig|Brief|Section|SkillVisibility|QuickCreate' -count=1`
- `go vet ./internal/daemon/execenv`
- `git diff --check`

The full `go test ./internal/daemon/execenv -count=1` was also attempted on Windows. Existing tests with Unix permission, path, and filesystem assumptions fail independently of this change; the focused suite and vet pass.

### DIFF
--- a/server/internal/daemon/execenv/runtime_config_kind_test.go
+++ b/server/internal/daemon/execenv/runtime_config_kind_test.go
@@ -127,9 +127,14 @@ func TestBuildMetaSkillContentSlimKindMatrix(t *testing.T) {
 		kindQuickCreate: true, kindChat: true,
 	}
 	issueKinds := map[taskKind]bool{kindIssue: true}
+	windowsKinds := map[taskKind]bool{}
+	if runtimeGOOS == "windows" {
+		windowsKinds = allKinds
+	}
 	checks := []sectionCheck{
 		{"# Multica Agent Runtime", allKinds},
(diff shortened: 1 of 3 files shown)


Bbbtt04:
## Issue screenshot

On Windows PowerShell 5.1, Chinese text read without an explicit UTF-8 encoding is displayed as mojibake in the task transcript.

Windows PowerShell 5.1 mojibake in the task transcript

Bbbtt04:
Closing this draft at the author’s request. The branch is currently conflicting with `main`; no changes from this PR were merged.

PR #6781, opened 2026-08-11
DP-575: make /health ping the DB as a hard dependency

digitalpartnersNL:
## Problem

The `/health` endpoint returned a static `{"status":"ok"}` 200 **unconditionally** — it never touched postgres. So `postgres-down` was a **false-green**: orchestrators and probes kept routing traffic to a node that could not serve any request.

`/readyz` and `/healthz` already enforced DB + migration checks, but `/health` (the liveness probe) bypassed them entirely.

## Fix

`liveHandler` now pings the DB (2s timeout) and returns **503** with `checks.db = "error"` on failure, mirroring the readiness contract `/readyz` already enforces:

```json
// healthy
{"status":"ok","checks":{"db":"ok"}}

// postgres down
{"status":"error","checks":{"db":"error"}}   // HTTP 503
```

Results are cached on the same TTL gate as readiness (`readinessCacheTTL`, 3s) so a postgres outage is not multiplied into one live ping per probe.

### Files
- `server/cmd/server/health.go` — `liveResponse` gains a `checks.db` field; new `liveness`/`computeLiveness`/`loadCachedLiveness` path with its own `livenessCache` (separate from the readiness cache so a liveness miss never serves a stale readiness entry and vice-versa). nil-DB is handled without panic.
- `server/cmd/server/health_test.go` — 4 new tests: DB-ping failure (503), DB ok (200), nil pool (503, no panic), and cache single-eval.

## Verification

```
go test ./cmd/server/ -run TestServerHealth -count=1 -v
```

All 8 health tests pass (4 new liveness + 4 existing readiness).

Ref: DP-575

### DIFF
--- a/server/cmd/server/health.go
+++ b/server/cmd/server/health.go
@@ -36,6 +36,7 @@ type serverHealth struct {
 	cacheTTL           time.Duration
 	refreshMu          sync.Mutex
 	cache              atomic.Pointer[cachedReadiness]
+	livenessCache      atomic.Pointer[cachedLiveness]
 }
 
 type cachedReadiness struct {
@@ -44,8 +45,24 @@ type cachedReadiness struct {
 	expiresAt  time.Time
 }
 
+type cachedLiveness struct {
+	response   liveResponse
+	statusCode int
+	expiresAt  time.Time
+}
+
 type liveResponse struct {
-	Status string `json:"status"`
+	Status string     `json:"status"`
+	Checks liveChecks `json:"checks"`
+}
+
+// liveChecks reports the liveness of individual hard dependencies. DB is the
+// only one: every request path needs postgres, so a dead DB means /health must
+// not report ok — otherwise orchestrators keep routing traffic to a node that
+// cannot serve (false-green). The field mirrors readinessChecks so /health and
+// /readyz share a consistent shape.
+type liveChecks struct {
+	DB string `json:"db"`
 }
 
 type readinessResponse struct {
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for the PR, and for taking the time to add tests and handle the nil-pool case — the implementation itself is clean.

That said, we're going to decline this one. The premise doesn't hold in this repository, and the change would regress a contract we rely on.

**`/health` is not the liveness probe here.** All three probes in our shipped chart point at `/healthz`:

```yaml
# deploy/helm/multica/templates/backend.yaml
startupProbe:    { httpGet: { path: /healthz, port: 8080 } }
readinessProbe:  { httpGet: { path: /healthz, port: 8080 } }
livenessProbe:   { httpGet: { path: /healthz, port: 8080 } }
```

And `/healthz` is wired to `readyHandler`, not `liveHandler` (`server/cmd/server/router.go:935-937`), so it already pings the DB and checks migrations. That means there is no false-green to fix: when postgres is down, readiness already fails and traffic is already drained from the pod. `/health` is not referenced by any probe in this repo.

Three concrete reasons the change is a regression for us:

1. **It's redundant.** `/readyz` and `/healthz` already do this exact DB ping, plus the migration check. After this change `/health` becomes a strict subset of `/readyz` and all three endpoints fail together — which costs us the one thing `/health` uniquely provides: a dependency-free reachability signal that distinguishes "process is dead / wrong URL / network broken" from "process is up, DB is down". That distinction matters during an incident.

2. **It breaks existing in-repo callers**, all of which require a 200 or use `curl -sf` (which fails on 503):
   - `server/cmd/multica/cmd_setup.go:441` — `probeServer()` gates `multica setup`. With this change, a DB outage makes the CLI report the backend as *unreachable*, pointing users at entirely the wrong problem.
   - `scripts/check.sh:110`, `scripts/install.sh:406`, `scripts/selfhost-wait.sh:60`, and the wait loop in `CONTRIBUTING.md:384`.

3. **It contradicts the documented contract**, which the PR doesn't update. `SELF_HOSTING_ADVANCED.md` states: "Use `/health` for basic liveness / reachability checks. Use `/readyz` for dependency-aware readiness probes and external monitoring that should fail when the database is unavailable or migrations are not fully applied. `/healthz` is kept as an alias for operator familiarity." Same in `SELF_HOSTING_AI.md`.

**There is a real bug nearby, though, and you were close to it.** Our `livenessProbe` pointing at the DB-dependent `/healthz` is genuinely wrong: a brief postgres outage fails liveness and the orchestrator kills a container that would have recovered on its own once the DB came back. The fix is the mirror image of this PR — point `livenessProbe` at `/health` (dependency-free, as documented) while leaving `readinessProbe` and `startupProbe` on `/healthz`.

If you'd like to send that as a follow-up PR, we'd happily review it. Closing this one.

PR #6785, opened 2026-08-11
feat(issues): preview an issue on mention hover

### LINKED ISSUE #6784: [Feature]: Previewing an issue mention means leaving the page you're reading
## What do you want and why?

An inline issue reference tells me the status, the identifier, and as much of the title as fits. When that's enough, it's enough. When it isn't, the only way to find out more is to open the issue — which means leaving the comment thread I was reading, and then coming back.

The cases where it isn't enough come up often in a busy thread:

- **The title is now truncated.** #6744 bounded the chip at `min(18rem, 100%)`, which was the right fix for the prose problem — but it means a long title ends in an ellipsis. `Inline issue references are hard to r…` is recognisable if I already know the issue and ambiguous if I don't.
- **"Who's on this?"** is the most common question I have about a referenced issue, and the chip never showed the assignee in any rendering.
- **"Is it nearly done?"** — for a parent issue, sub-issue progress answers this at a glance and is otherwise a page load away.

None of this is a complaint about the chip. The chip is inline in someone's sentence; it should stay small. The gap is that there's no cheap middle step between "the chip" and "open the issue".

## Proposed solution (optional)

A hover card on issue mentions, showing what the chip cannot: the **full untruncated title**, **priority**, a **description snippet**, the **assignee**, and **sub-issue progress** when the issue has children.

Deliberately not in scope: any change to the chip itself, or to click behaviour. Plain click still navigates in place, modifiers still open tabs, per #6667.

Two implementation notes, since they're the parts worth disagreeing with early:

- **The detail query should be deferred by placement, not by an `enabled` flag** — it belongs inside the popup body, which Base UI mounts only while open. A paragraph with eight mentions must fire zero detail requests on render. Easy to get wrong in a way that no test notices.
- **Assignee costs nothing extra.** `ActorAvatar` resolves `assignee_type` + `assignee_id` against the members/agents/squads lists that `useWorkspacePresencePrefetch` already loads app-wide at `staleTime: Infinity`. Sub-issue progress costs one workspace-wide request (`childIssueProgressOptions`), already shared with the issues list and issue detail.

There's precedent for the shape: `MentionHoverCard` already does this for member and agent mentions, so issue mentions being the ones without a preview is the odd case.

This was part of #6625, which was closed in favour of #6744 — @Bohan-J suggested the hover card land standalone, which is what this issue is for. I have an implementation ready and will link the PR here.

szainmehdi:
## What does this PR do?

Adds a hover card to inline issue mentions, showing what the chip cannot: the **full untruncated title**, **priority**, a **description snippet**, the **assignee**, and **sub-issue progress**.

This is the part of #6625 that never depended on the display-mode preference. @Bohan-J suggested it land standalone; that's what this is. Rebuilt from scratch against current `main` rather than cherry-picked, because #6744 and #6667 both changed the ground underneath it.

#6744 made the case stronger rather than weaker. Bounding the chip at `min(18rem, 100%)` was the right fix for the prose problem, and it means a long title now arrives truncated — `Investigate and fix the flaky webs…`. The card is where the rest of it goes.

## Related Issue

Closes #6784

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue) — see the `descriptionPreview` fix below

## Changes Made

**New:**
- `packages/views/issues/components/issue-hover-card.tsx` — the card. `IssueHoverCard` (trigger + popup) and an inner `IssueHoverCardBody` that owns every query.
- `packages/views/issues/components/description-preview.ts` — `descriptionPreview()` moved out of `board-card.tsx`, where a shared util was living inside a component. Its test moved with it via `git mv`. No re-export shim, per the repo's rule against internal compatibility layers.

**Modified:**
- `packages/views/issues/components/issue-mention-card.tsx` — wraps the existing `AppLink` in `IssueHoverCard`. Nothing else; click semantics are untouched.
- `packages/views/issues/components/board-card.tsx` — imports `descriptionPreview` from its new home.

**Deliberately untouched** — `issue-chip.tsx` and `editor/extensions/mention-view.tsx` are byte-identical to `main` (same blob hashes, not just an empty diff). The editor gets the card for free, because #6667 already pointed `MentionView` at `IssueMentionCard`.

### Implementation notes for reviewers

**The detail query is deferred by placement, not by an `enabled` flag.** `useQuery` lives in `IssueHoverCardBody`, which Base UI mounts only while the popup is open. Hoisting it into the outer component would fire one detail request per mention on render — a paragraph with eight mentions would issue eight. There's a test asserting zero calls before open, and it asserts the trigger rendered *first*, so a component that renders nothing can't satisfy it vacuously.

**Enrichment cost.** ~~The assignee uses `ActorAvatar` against the members/agents/squads lists `useWorkspacePresencePrefetch` already loads app-wide at `staleTime: Infinity` — no extra request.~~

**Correction (review round 1):** that was wrong, and @Bohan-J caught it. `packages/core/agents/use-workspace-presence-prefetch.ts:25-28` warms agents, runtimes, the agent task snapshot, and squads — **not** the member directory. `useActorName()` subscribes to `memberListOptions`, so as originally written the first hover on *any* mention could pull the member list even when there was no assignee to render. The assignee row is now its own component, mounted only when the issue has an assignee, which confines that query to cards that actually name someone. Leaving the original claim struck through rather than silently editing it.

`ActorAvatar` is passed `enableHoverCard={false}` and `profileLink={false}`, since nesting a hover card inside a hover card is a bug and the agent live-peek variant fires its own request. Sub-issue progress uses `childIssueProgressOptions` — one workspace-wide snapshot shared with the issues list and issue detail, deliberately not `childIssuesOptions`, which is per-issue and `refetchOnMount: "always"`.

**No new strings.** The card now uses three existing keys — `issues:detail.not_found` for the failure state, and the shipped `status.*` / `priority.*` maps for the icon labels — and still adds nothing to the four locale bundles, so there is nothing new for `parity.test.ts` to enforce. Loading is still a `Skeleton` rather than localized text.

**Priority is omitted when it's `none`.** Board cards and list rows render the "none" dash because it holds a column in a fixed grid; this card has no column to hold, so the dash would be a mark carrying no information in an already dense popup. Flagging it because it's a deliberate divergence from those surfaces, not an oversight.

### The `descriptionPreview` fix (please look at this separately)

Moving the function surfaced a real defect in it, which affects board cards today and would have affected this card:

The editor serializes link brackets escaped, so a description round-tripped through it is stored as `\[repro script\](https://example.com)`. The link rule's `[^\]]+` capture swallowed the trailing backslash while the leading one never matched, and the preview read `\repro script\`. Fixed by unescaping backslash-escaped ASCII punctuation *first* — order is load-bearing, since the escapes are what stop the later structural rules from matching.

A second case fell out of testing against real stored data: when a URL inside a link label gets autolinked, the editor produces a **nested** link — `\[repro script\]([https://example.com](https://example.com))`. The destination pattern `[^)]+` stopped at the first `)`, consumed through the inner link, and stranded the outer paren: `…needs a repro script) before we fix it.` The destination now tolerates one level of balanced parens, which also fixes legitimate URLs like `https://en.wikipedia.org/wiki/Foo_(bar)`. Applied to the image and file rules too, which had the same fragility.

Both are improvements to board cards as a side effect. Happy to split them into their own PR if you'd rather review them apart from the card.

## How to Test

1. `make dev`, open an issue whose description or comments reference other issues.
2. Hover a mention and wait out the standard delay. The card shows priority, status, identifier, the full title, a description snippet, the assignee, and sub-issue progress.
3. Reference an issue with a **long title** so the inline chip truncates at the 18rem cap — the card shows the whole title.
4. Hover a bare issue (no description, unassigned, no sub-issues, no priority). Rows drop out cleanly; no empty gaps or stray separators.
5. Type an identifier into a comment composer so it autolinks, then hover the chip. Same card — the editor path routes through `IssueMentionCard`.
6. Click and cmd-click a mention: plain click navigates in place, cmd-click opens a background tab, per #6667. Unchanged by this PR.

Verified locally: `pnpm typecheck` clean across 6 packages; `pnpm test` — views 3826, core 1346, desktop 462, web 199, docs 17; `pnpm lint` 0 errors.

Browser-verified in both themes, including a card with every field populated, a partial card, a minimal card, the editor path, and click/cmd-click behavior.

## Checklist

    - No docs change appears owed: no page under `apps/docs/content/docs/` describes mention rendering, and the equivalent member/agent preview (`MentionHoverCard`) has none either. No changelog entry either — `changelog.entries` lives in `apps/web/features/landing/i18n/*` and is written at release time, one commit per version. Say the word if you'd rather it came with the change.
- [x] If I added a new runtime / coding tool / UI tab, I synced landing copy and docs — n/a, this adds none of those
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` — n/a, no strings added

### Known gaps

- **Touch devices have no hover**, so tapping navigates instead of previewing. Unchanged from how `MentionHoverCard` behaves for member and agent mentions.
- **Sub-issue progress is `staleTime: Infinity`** and shared with the issues list, so a long-open card can show a stale `done/total`. Same tradeoff the existing surfaces make, now visible in one more place.
- `apps/mobile` is untouched — it renders bare identifier text and has no chip to preview.

## AI Disclosure

**Prompt / approach:** Same workflow as #6625 — spec, then plan, then task-by-task execution with a reviewer subagent gating each task, plus hands-on browser verification driving the real app.

Two things that came out of the verification pass rather than the code review, both worth naming because neither would have been caught by tests:

- The description snippet leaked backslashes and, separately, a stray `)`. I only found the second one after dumping the actual stored markdown out of Postgres rather than trusting a theory about the cause — the real content turned out to contain a nested link the regex couldn't parse.
- A first report claimed `it costs \$5` was also broken. It wasn't: the stored bytes were `\\$5`, an escaped backslash, so rendering `\$5` is the faithful round-trip. Worth mentioning because it would have been easy to "fix" correct behavior.

## Screenshots

_Attaching shortly — the card with every field populated, the truncated chip beside the full title, partial and minimal cards, the editor path, and light mode._
01-hover-card-everything
02-truncated-chip-vs-full-title
03-hover-card-partial
04-hover-card-minimal
05-editor-hover
06-hover-card-light-mode

### DIFF
--- a/packages/views/issues/components/board-card.tsx
+++ b/packages/views/issues/components/board-card.tsx
@@ -6,11 +6,11 @@ import { useSortable, defaultAnimateLayoutChanges } from "@dnd-kit/sortable";
 import type { AnimateLayoutChanges } from "@dnd-kit/sortable";
 import { CSS } from "@dnd-kit/utilities";
 import type { Issue, IssueProperty, Project, UpdateIssueRequest } from "@multica/core/types";
-import { stripChannelMediaMarkers } from "@multica/core/types";
 import { useQuery } from "@tanstack/react-query";
 import { useWorkspaceId } from "@multica/core/hooks";
 import { propertyListOptions } from "@multica/core/properties";
 import { CustomPropertyValueDisplay } from "./pickers/custom-property-picker";
+import { descriptionPreview } from "./description-preview";
 import { formatDateOnly, isPastDateOnly } from "@multica/core/issues/date";
 import { CalendarClock, CalendarDays } from "lucide-react";
 import { ActorAvatar } from "../../common/actor-avatar";
@@ -32,22 +32,6 @@ import { useT } from "../../i18n";
 
 function formatDate(date: string): string {
   return formatDateOnly(date, { month: "short", day: "numeric" }, "en-US");
-}
-
-// Flatten description Markdown into the card's one-line preview. Channel-media
-// provenance is server-owned merge metadata, not authored content, so it is
-// stripped first: the image Markdown it annotates is removed a line below, and
-// without this the bare HTML comment survives every remaining pass and becomes
-// visible preview text. Exported for tests.
-export function descriptionPreview(markdown: string): string {
-  return stripChannelMediaMarkers(markdown)
-    .replace(/!file\[[^\]]*\]\([^)]*\)/g, "")
-    .replace(/!\[[^\]]*\]\([^)]*\)/g, "")
-    .replace(/\[([^\]]+)\]\([^)]+\)/g, "$1")
-    .replace(/[*_`~]+/g, "")
-    .replace(/^[\s>#]+/gm, "")
-    .replace(/\s+/g, " ")
-    .trim();
 }
 
 /** Stops event from bubbling to Link/drag handlers */
--- a/packages/views/issues/components/description-preview.test.ts
+++ b/packages/views/issues/components/description-preview.test.ts
@@ -25,4 +25,30 @@ describe("descriptionPreview", () => {
       descriptionPreview("# Title\n\n**bold** text and [a link](https://example.com)"),
     ).toBe("Title bold text and a link");
   });
+
+  // The editor serializes link brackets escaped. Without unescaping first, the
+  // link rule leaves the leading `\` untouched and captures the trailing one,
+  // so the preview reads `\repro script\`.
+  it("unwraps a link whose brackets the editor escaped", () => {
+    expect(descriptionPreview("See \\[repro script\\](https://example.com)")).toBe(
+      "See repro script",
+    );
+  });
+
+  it("still unwraps a link written without escapes", () => {
+    expect(descriptionPreview("See [repro script](https://example.com)")).toBe(
+      "See repro script",
+    );
+  });
+
+  it("drops the backslashes from escaped emphasis and punctuation", () => {
+    expect(descriptionPreview("it costs \\$5 \\*not\\* \\$10")).toBe("it costs $5 not $10");
+  });
+
+  // A backslash before a letter is not a Markdown escape, so prose keeps it.
+  it("leaves a literal backslash in prose alone", () => {
+    expect(descriptionPreview("open C:\\Users\\zain to continue")).toBe(
+      "open C:\\Users\\zain to continue",
+    );
+  });
 });
--- a/packages/views/issues/components/description-preview.test.ts
+++ b/packages/views/issues/components/description-preview.test.ts
@@ -41,6 +41,24 @@ describe("descriptionPreview", () => {
     );
   });
 
+  // The editor autolinks a raw URL inside a link label, producing a nested
+  // link destination like `([https://x](https://x))`. A destination pattern
+  // that stops at the first `)` leaves the outer paren stranded in the
+  // preview.
+  it("unwraps a link whose destination is itself a nested link", () => {
+    expect(
+      descriptionPreview(
+        "This is **very** flaky and needs a \\[repro script\\]([https://example.com](https://example.com)) before we fix it.",
+      ),
+    ).toBe("This is very flaky and needs a repro script before we fix it.");
+  });
+
+  it("unwraps a link whose destination legitimately contains parentheses", () => {
+    expect(
+      descriptionPreview("See [wiki](https://en.wikipedia.org/wiki/Foo_(bar)) for details"),
+    ).toBe("See wiki for details");
+  });
+
   it("drops the backslashes from escaped emphasis and punctuation", () => {
     expect(descriptionPreview("it costs \\$5 \\*not\\* \\$10")).toBe("it costs $5 not $10");
   });
--- a/packages/views/issues/components/description-preview.test.ts
+++ b/packages/views/issues/components/description-preview.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { descriptionPreview } from "./board-card";
+import { descriptionPreview } from "./description-preview";
 
 const ATTACHMENT_ID = "0f8a1b2c-3d4e-4f50-9a6b-7c8d9e0f1a2b";
 const IMAGE = `![](/api/attachments/${ATTACHMENT_ID}/download)`;
--- a/packages/views/issues/components/description-preview.ts
+++ b/packages/views/issues/components/description-preview.ts
@@ -22,9 +22,15 @@ export function descriptionPreview(markdown: string): string {
       // punctuation. A backslash before anything else (a letter, a space) is a
       // literal backslash in prose and is left alone.
       .replace(/\\([!"#$%&'()*+,\-./:;<=>?@[\\\]^_`{|}~])/g, "$1")
-      .replace(/!file\[[^\]]*\]\([^)]*\)/g, "")
-      .replace(/!\[[^\]]*\]\([^)]*\)/g, "")
-      .replace(/\[([^\]]+)\]\([^)]+\)/g, "$1")
+      // The destination group tolerates one level of balanced parentheses,
+      // e.g. `(https://en.wikipedia.org/wiki/Foo_(bar))` or a nested link
+      // produced when a URL inside a link label gets autolinked, which
+      // serializes as `[label]([https://x](https://x))`. A naive `[^)]+`
+      // stops at the first `)` and leaves the outer one stranded in the
+      // preview text — do not simplify this back to `[^)]+`.
+      .replace(/!file\[[^\]]*\]\((?:[^()]|\([^()]*\))*\)/g, "")
+      .replace(/!\[[^\]]*\]\((?:[^()]|\([^()]*\))*\)/g, "")
+      .replace(/\[([^\]]+)\]\((?:[^()]|\([^()]*\))*\)/g, "$1")
       .replace(/[*_`~]+/g, "")
       .replace(/^[\s>#]+/gm, "")
       .replace(/\s+/g, " ")
--- a/packages/views/issues/components/description-preview.ts
+++ b/packages/views/issues/components/description-preview.ts
@@ -10,12 +10,24 @@ import { stripChannelMediaMarkers } from "@multica/core/types";
  * remaining pass and becomes visible preview text.
  */
 export function descriptionPreview(markdown: string): string {
-  return stripChannelMediaMarkers(markdown)
-    .replace(/!file\[[^\]]*\]\([^)]*\)/g, "")
-    .replace(/!\[[^\]]*\]\([^)]*\)/g, "")
-    .replace(/\[([^\]]+)\]\([^)]+\)/g, "$1")
-    .replace(/[*_`~]+/g, "")
-    .replace(/^[\s>#]+/gm, "")
-    .replace(/\s+/g, " ")
-    .trim();
+  return (
+    stripChannelMediaMarkers(markdown)
+      // Backslash-escaped punctuation is unescaped first, and the order is
+      // load-bearing. The editor serializes a link as `\[label\](url)`, so the
+      // link rule below sees `\[label\]` and its `[^\]]+` capture swallows the
+      // trailing backslash while the leading one survives — the preview then
+      // reads `\label\`. Unescaping after any structural rule is too late,
+      // because the escapes are what stop those rules from matching.
+      // The character class is CommonMark's escapable set: all ASCII
+      // punctuation. A backslash before anything else (a letter, a space) is a
+      // literal backslash in prose and is left alone.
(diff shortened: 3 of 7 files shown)


Bohan-J (changes requested):
Thanks for this — the product direction is right, and the implementation is close. Keeping the queries inside `IssueHoverCardBody` so nothing fetches until a card actually opens is the correct call, and reusing the workspace-wide child progress snapshot instead of a per-issue children fetch is a nice touch. The `descriptionPreview` extraction with its own regression tests, and the link-navigation coverage, are both appreciated.

Two things I'd like fixed before merge.

### 1. A failed detail fetch shows the skeleton forever

`packages/views/issues/components/issue-hover-card.tsx:75` gates on `!issue` alone, which collapses three distinct states into one:

```tsx
const { data: issue } = useQuery(issueDetailOptions(wsId, issueId));
// ...
if (!issue) {
  return (/* skeleton */);
}
```

`data` is `undefined` while pending, but it is *also* `undefined` when the query settles into `error` — a deleted issue, a mention the viewer can't access, or a network failure after retries are exhausted. In all of those the user hovers and sits on an animated skeleton indefinitely, with no signal that nothing is coming.

Please branch on the query status explicitly (`isPending` vs `isError` / resolved-but-missing) and render a stable terminal state for the failure case, consistent with how issue detail and mention chips already handle an unresolvable reference. Worth adding tests for a rejected query and a deleted-issue reference. If the failure state carries any copy, route it through the existing i18n setup rather than a hardcoded string.

### 2. Status and priority are invisible to assistive tech

`packages/views/issues/components/issue-hover-card.tsx:99-105` renders both glyphs with no accessible name, and the comment states the omission is deliberate:

```tsx
{/* ... No aria-label: naming it would need a translation key, and this card
    deliberately carries no copy of its own. */}
{hasPriority && <PriorityIcon priority={issue.priority} />}
<StatusIcon status={issue.status} className="h-3.5 w-3.5 shrink-0" />
```

The premise doesn't hold — the keys already exist. `packages/views/locales/en/issues.json` ships a full `status.*` map (`backlog`, `todo`, `in_progress`, `in_review`, `done`, `blocked`, `cancelled`) and a full `priority.*` map, and `packages/views/issues/components/board-card.tsx:103` already does exactly this:

```tsx
const priorityLabel = t(($) => $.priority[issue.priority]);
// ...
aria-label={priorityLabel}
```

So no new copy is needed; this is reusing what's there. As it stands, a screen reader user gets the identifier and title but not two of the five attributes the card exists to surface. Please add `aria-label` to both icons using the existing keys, and cover it with a test that focuses a mention via keyboard, opens the card, and asserts both names are exposed.

### Nits (non-blocking)

- `issue-hover-card.tsx:110` — the full title has no `break-words`. Since the point of the card is the *untruncated* title, a long unbroken token (a pasted URL, a long identifier) will overflow the `max-w-80` cap. Worth a case in the tests.
- `issue-hover-card.tsx:71` — `useActorName()` is called unconditionally, including when `hasAssignee` is false. The PR description says members are globally prewarmed, but `packages/core/agents/use-workspace-presence-prefetch.ts:24-28` only warms agents, runtimes, the agent task snapshot, and squads — not the member directory. So the first hover on any mention can pull the member list even when there's no assignee to render. Splitting the assignee row into a child component that only mounts when `hasAssignee` is true would confine that, and the description could be corrected.

Nothing here touches the API, schema, or permission boundaries — the remaining risk is concentrated in the failure state and the accessibility gap. Happy to re-review as soon as those two are addressed.

szainmehdi:
Thanks — all four addressed, one commit each, rebased onto current `main`.

### 1. Terminal state when the detail fetch fails — `df757f0c7`

You were right that `!issue` collapsed three states into one. The body now branches on the query phase: `isPending` renders the skeleton, `isError || !data` renders a terminal state, and the skeleton is unreachable once the query settles.

No new copy — `issues:detail.not_found` ("This issue does not exist or has been deleted in this workspace.") already ships in all four locales, so this reuses it. To name the issue the way the chip degrades, `IssueHoverCard` gained a `fallbackLabel` prop threaded from `IssueMentionCard`, which already had one; the failure card shows the identifier above the sentence.

Three tests: a rejected query, a settled-but-undefined query, and the pending case. The failure tests assert the skeleton is **absent**, not merely that the message appears — otherwise they'd pass with both rendered.

**One caveat I want to be straight about:** I could not reproduce the failure state in a live browser. Request interception didn't take effect in my setup and a `window.fetch` override didn't reach the page's real calls, so I have unit coverage and a code read, but no live screenshot. If you'd like a specific repro path, tell me and I'll capture it.

### 2. Accessible names on status and priority — `cc64b98f7`

You're right, and the comment asserting otherwise was simply wrong — I've deleted it. The keys were already there.

One deliberate divergence from `board-card.tsx`: I used `role="img"` on the wrapper rather than a bare `<span aria-label>`. On a role-less generic element `aria-label` isn't reliably exposed, so the board-card pattern likely doesn't announce today either. Worth a look at `board-card.tsx:113` and `:123` separately — I left it alone to keep this PR to its own scope.

Verified in a real browser rather than only in jsdom:

```js
Array.from(document.querySelectorAll('[role="img"]'))
  .map(el => ({role: el.getAttribute('role'), label: el.getAttribute('aria-label')}))
// [{"role":"img","label":"Urgent"},{"role":"img","label":"Todo"}]
```

The accessibility snapshot of an open card announces: `image "Urgent"`, `image "Todo"`, then `ACM-4`, the full title, the description snippet, `dev`, and `1/3`.

**On the keyboard test specifically:** the unit test opens via hover, not Tab. The trigger renders as a non-focusable `<span>` (the focusable element is the `AppLink` inside it), and jsdom wouldn't reproduce the focus path. So I verified it by hand instead — tabbing to a mention takes focus and **opens the card on focus alone**, no pointer involved, with both names exposed as above. Happy to add a keyboard-specific test if you know a pattern that works against Base UI's `PreviewCard` in this suite.

### 3. `break-words` on the title — `7fc79939d`

Added, with a test using a long unbroken URL. Confirmed in a browser that the title wraps inside the card with no overflow past the `max-w-80` cap.

### 4. `useActorName()` firing unconditionally — `34a2d5221`

Correct on both counts, including that the PR description was wrong. `use-workspace-presence-prefetch.ts:25-28` warms agents, runtimes, the agent task snapshot and squads — not members — so the first hover on any mention could pull the member directory even with no assignee to render.

The assignee row is now its own component, mounted only when the issue has an assignee. The test converts `useActorName` into a real spy and asserts `not.toHaveBeenCalled()` for an unassigned issue, so the confinement is pinned rather than assumed.

I've struck the incorrect sentence in the PR description and left the correction visible rather than quietly editing it.

---

Also rebased onto current `main` (the branch had drifted behind). Full run on the rebased branch: `pnpm test` — views 3859, core 1348, desktop 462, web 199; `pnpm typecheck` 6/6; `pnpm lint` 0 errors. `issue-chip.tsx` and `mention-view.tsx` remain byte-identical to `main`.

Ready for another look whenever suits.

Bohan-J (approved):
Both blockers are resolved, and the fixes are better than what I asked for. Thanks for the careful follow-up on this.

**Failure state** — `IssueHoverCardBody` now splits `isPending` from `isError || !issue`, so a settled query can never sit on an animated skeleton. Nice touch reusing `fallbackLabel` — the label the chip already degrades to — so the terminal state still says *which* issue it couldn't load, and reusing the existing `issues.detail.not_found` copy rather than inventing a key. I checked all four locales (`en` / `ja` / `ko` / `zh-Hans`) and the key is present in each.

**Accessible names** — you were right to not do this the way I described it. `StatusIcon` and `PriorityIcon` render a bare `<svg>` that forwards no aria props, and an `aria-label` on a role-less `<span>` gets dropped by assistive tech anyway, so the `role="img"` wrapper is the correct fix rather than the naive one. The test asserting `getByLabelText("In Progress")` *contains* the status icon is the right shape too — it fails if either the label or the wiring regresses.

**Both nits** — `break-words` on the title with a long-unbroken-token regression test, and the assignee row extracted into a component that only mounts when there is an assignee. `expect(mockUseActorName).not.toHaveBeenCalled()` in the unassigned case is exactly the assertion that keeps that from silently coming back. Thanks for correcting the prefetch note in the description as well.

Verified at `1a110780`: all 9 CI checks green, `MERGEABLE` / `CLEAN`, scope still limited to the same 8 frontend files — no API, schema, or migration surface.

One non-blocking follow-up, not for this PR: the accessible-name test opens the card via hover, so the keyboard path (Tab to the inner link, trigger opens the card) rests on your manual browser verification rather than an automated test. Worth a keyboard integration test later so a change to the trigger composition can't regress it quietly.

Merging. Thanks again for the thorough work here — the design rationale in the docstrings and the "no fetch until open" invariant with a test guarding it made this an easy review.

PR #6790, opened 2026-08-11
MUL-6045: Reach hover-only row controls on touch

tim-smart:
## What changed

- Added a shared compact row-actions menu for inbox rows, chat thread rows, and floating chat-session history rows on mobile breakpoints.
- Reused one action model for each row's mobile menu, desktop hover controls, and inbox context menu so the available actions cannot drift.
- Made desktop hover controls keyboard-reachable with focus-within behavior.
- Reworked inbox row activation to avoid invalid nested interactive elements and prevent control keypresses from selecting the parent row.
- Added localized accessible labels in English, Japanese, Korean, and Simplified Chinese, plus focused component coverage.

## Why

These per-item controls were exposed only through hover or right-click. Touch devices provide neither interaction reliably, leaving archive, read-state, pin, rename, stop, and related actions inaccessible.

Desktop presentation remains unchanged: compact menus are hidden from the `md` breakpoint upward, where the existing hover controls remain in place.

## Validation

- `pnpm typecheck`
- `pnpm lint` (zero errors; existing warnings only)
- Focused `@multica/views` tests: 40 passed across the chat-thread, row-actions, inbox-context-menu, and inbox-list-item suites
- The source fork's full CI passed before the branch was squashed and rebased onto current upstream `main`

### DIFF
--- a/packages/views/chat/components/chat-thread-list.test.tsx
+++ b/packages/views/chat/components/chat-thread-list.test.tsx
@@ -87,19 +87,22 @@ const sessions: ChatSession[] = [
   makeSession({ id: "s3", updated_at: "2026-07-08T01:00:00Z" }),
 ];
 
-function renderList(activeSessionId: string | null, onArchive = vi.fn()) {
+function renderList(
+  activeSessionId: string | null,
+  { onArchive = vi.fn(), onSelectSession = vi.fn() } = {},
+) {
   render(
     <I18nProvider locale="en" resources={TEST_RESOURCES}>
       <ChatThreadList
         sessions={sessions}
         agents={[agent]}
         activeSessionId={activeSessionId}
-        onSelectSession={vi.fn()}
+        onSelectSession={onSelectSession}
         onArchive={onArchive}
       />
     </I18nProvider>,
   );
-  return onArchive;
+  return { onArchive, onSelectSession };
 }
 
 const ARCHIVE_LABEL = enChat.list.archive;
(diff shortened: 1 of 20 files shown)


Bohan-J:
Thanks for this — really nice piece of work. The part I appreciate most isn't the menu itself but that you refactored the actions into a single shared model (`useInboxItemActions`, the `rowActions` arrays) and made the existing right-click menu and hover strip consume it, so the touch and pointer surfaces structurally can't drift. That's the version of this change that stays correct a year from now. Removing the invalid `<button>`-inside-`<button>` nesting and getting the row controls into the tab order were both overdue, too.

I've pushed one commit on top (`1114999`) rather than sending you round again for a one-line change:

**Gate the switch on `hover`, not the `md` breakpoint.** The compact menu was `md:hidden` and the hover controls were `md:group-hover:` / `md:group-focus-within:`, which makes the handoff key on viewport width. Any touch surface wider than 768px falls between the two — most concretely a phone in landscape, which clears `md` while still reporting `hover: none`, so it lost the `⋯` menu *and* got hover controls it can't trigger. That's the same gap the PR set out to close, just at a different width.

All five call sites now use `[@media(hover:hover)]`, matching the pattern already in `execution-log-section.tsx:628-639`. I also added regression coverage asserting neither surface is gated on a width breakpoint, so this can't quietly regress to `md:` later.

Verification on the merge commit: `@multica/views` typecheck clean, lint clean on the touched files, full suite 3827/3827.

One review point we're **not** asking you to fix: the row is `role="button"` with focusable descendants, which trips [ACT 307n5z](https://www.w3.org/WAI/standards-guidelines/act/rules/307n5z/). It's a real rule, but `main` already violated it (a native `<button>` has the same presentational-children semantics, and the old `tabIndex={-1}` span still counts as focusable under ACT's definition), and the two chat rows next door are focusable `<div>`s with no role at all — arguably worse for screen readers, just not rule-visible. Fixing it properly means reworking the row shell across all three lists, which is its own change. We're tracking it separately rather than holding this up.

Merging now. Thanks again.

PR #6791, opened 2026-08-12
MUL-6046 AGE-505: add safe persistent mobile chat outbox

LiaoyuanNing:
## Summary
- persist unsent existing-session chat messages by account and workspace, render them in the conversation, and support native iOS edit/remove actions
- use an 8 second foreground send ceiling to surface false-online failures promptly
- deliberately disable automatic reconnect/background flushing: the current server endpoint has no client idempotency key or safe dedupe lookup
- keep offline new-chat text as a draft and require a connection to create a session

## Safety
A manual retry is the only retry path after an uncertain result. FIFO prevents a later message overtaking an earlier unconfirmed one; 4xx failures are terminal; logout and 401 clear the local partition.

## Validation
- `pnpm --dir apps/mobile typecheck`
- `pnpm --dir apps/mobile test -- chat-outbox.test.ts chat-outbox-store.test.ts auth-store.test.ts` (14 files, 71 tests)
- `pnpm --dir apps/mobile lint` (0 errors; 7 existing warnings in untouched files)
- `git diff --check`

## Follow-up
Automatic foreground flush requires a server contract that accepts a client message id and returns the same accepted result on replay. This PR intentionally makes no server changes.

### DIFF
--- a/apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx
+++ b/apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx
@@ -111,6 +111,11 @@ export default function ChatTab() {
   const [activeSessionId, setActiveSessionId] = useState<string | null>(null);
   const [selectedAgentId, setSelectedAgentId] = useState<string | null>(null);
   const [agentPickerOpen, setAgentPickerOpen] = useState(false);
+  const [editingOutbox, setEditingOutbox] = useState<{
+    clientId: string;
+    draftKey: string;
+    previousDraft: string;
+  } | null>(null);
 
   // Bridge to the chat-sessions formSheet route. Mirror local
   // activeSessionId into the store so the picker can render the current
@@ -213,6 +218,14 @@ export default function ChatTab() {
   const enqueueOutbox = useChatOutboxStore((s) => s.enqueue);
   const updateOutbox = useChatOutboxStore((s) => s.update);
   const removeOutbox = useChatOutboxStore((s) => s.remove);
+
+  // Session switches cannot leave another session's queued text in the
+  // composer. Put the draft that preceded editing back where it came from.
+  useEffect(() => {
+    if (!editingOutbox || editingOutbox.draftKey === draftKey) return;
+    setDraft(editingOutbox.draftKey, editingOutbox.previousDraft);
+    setEditingOutbox(null);
+  }, [draftKey, editingOutbox, setDraft]);
 
   // ── Realtime ───────────────────────────────────────────────────────────
   useChatSessionRealtime(activeSessionId, () => {
@@ -329,8 +342,8 @@ export default function ChatTab() {
           error.status >= 400 &&
           error.status < 500
         ) {
-          // A 401 clears this user partition through auth-store. Marking it
-          // failed first avoids a retry race while that cleanup runs.
+          // Auth expiry must retain unsent content. Mark this local item
+          // failed and let the user authenticate again before a manual retry.
           updateOutbox(head.clientId, (item) =>
             permanentlyFailedChatOutboxItem(item, message),
           );
@@ -371,6 +384,34 @@ export default function ChatTab() {
       }
       if (!userId || !wsSlug) return;
 
+      if (editingOutbox) {
+        const item = useChatOutboxStore
+          .getState()
+          .items.find((candidate) => candidate.clientId === editingOutbox.clientId);
+        if (!item || item.status === "sending") {
+          setEditingOutbox(null);
+          return;
+        }
+
+        // An edited instruction is logically a new manual-delivery attempt.
+        // Keep its queue position, but give it a new local id and clear retry
+        // state so it cannot be mistaken for an earlier payload.
+        updateOutbox(item.clientId, (current) => ({
+          ...current,
+          content,
+          attachmentIds: [...new Set([...current.attachmentIds, ...attachmentIds])],
+          clientId: createChatOutboxClientId(),
+          status: "queued",
+          attemptCount: 0,
+          retryable: true,
+          lastError: null,
+          nextAttemptAt: null,
+        }));
+        clearDraft(editingOutbox.draftKey);
+        setEditingOutbox(null);
+        return;
+      }
+
       const isNewSession = !activeSessionId;
       const sessionId = await ensureSession(content);
       if (!sessionId) return;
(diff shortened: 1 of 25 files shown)


multica-eve (changes requested):
Before this change, a mobile user could lose a message when connectivity dropped or the send result was uncertain. After this change, the unconfirmed message remains visible in the conversation, and the user can inspect, edit, remove, or explicitly retry it without the app silently resending it.

## Product decision

The safe outbox and user-controlled retry behavior are worth accepting. The broader seven-day offline retention of all successful read data should also be confirmed as intentional product scope; if AGE-504 already approved that behavior, I have no product objection to it.

## Must fix

1. **An offline first message loses the draft it is supposed to preserve.** `apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx:365` shows an alert and returns normally, but `apps/mobile/components/composer/message-composer.tsx:264` has already cleared the text, attachments, and mentions and restores them only when the callback rejects. Make this path report a failed submission (or check send eligibility before the optimistic clear), and add an interaction regression test.

2. **A persisted `sending` item becomes permanently stuck after process death or a workspace switch.** `chat.tsx:285` persists `sending`; `apps/mobile/data/stores/chat-outbox-store.ts:51` restores it unchanged; `apps/mobile/data/stores/chat-outbox.ts:53` only selects a `queued` head; and `chat.tsx:461` only lets the user wait or remove a `sending` item. On hydration, normalize orphaned `sending` items into an explicit “result unknown—check before retrying” state that remains manually manageable and is never automatically resent. Add a cold-start test.

3. **Deleting a chat leaves its local draft and outbox entries behind.** Both delete paths (`chat.tsx:554` and `apps/mobile/app/(app)/[workspace]/chat-sessions.tsx:31`) only delete server state. Unconfirmed messages and drafts then lose their UI entry point but remain on the device indefinitely; a remote session deletion has the same issue. Clean both stores after confirmed deletion, or explicitly resolve outstanding items before deletion, and cover local and remote deletion.

4. **Out-of-order async hydration can overwrite the active account/workspace scope.** The cancellation flag in `apps/mobile/app/(app)/[workspace]/_layout.tsx:126` only suppresses the final render marker; it does not stop an older `hydrate*()` call from changing a global store's persistence name and contents after a newer scope has loaded (`chat-drafts-store.ts:66`, `chat-outbox-store.ts:51`, and the new issue/project draft stores). `auth-store.ts:53` also allows overlapping initialization/retry/logout operations, while `offline.tsx:34` cannot disable Retry because initialization never sets `isLoading` back to true. Add generation/single-flight protection and tests that complete A/B hydration out of order, retry concurrently, and sign out during retry.

5. **Logout/401 cache deletion can be undone by an already scheduled persistence write.** `apps/mobile/data/query-persistence.ts:58` uses the async persister's throttled writer. Unsubscribing and removing the key at line 67 does not cancel a queued write, so the signed-out user's full cache can be recreated afterward. Use a cancellable/flushable lifecycle or a generation/tombstone guard, and test that a pending write followed by logout leaves no cache key. When the user id is unavailable, clear all keys under this app's query-cache prefix just as draft cleanup already does.

The deliberate no-auto-retry behavior, per-session FIFO blocking, terminal 4xx handling, scoped partitions, and short foreground send ceiling are good decisions to keep.

## Recommendation

**Request changes.** The product direction is sound, but the issues above can lose drafts, strand unconfirmed messages, mix workspace-local state, or retain signed-out user data. Please address them and add the race/cold-start regressions before merging.

multica-eve (changes requested):
复审结论：**Request Changes**。最新 head `a007fc434` 增加了 401 后保留 outbox、outbox 编辑入口和持久化测试，但上一轮 5 个 must-fix 仍为 **0/5 完整解决**：

1. **离线发送新会话首条消息仍会丢草稿。** `apps/mobile/app/(app)/[workspace]/(tabs)/chat.tsx:378-384` 弹窗后正常返回；`apps/mobile/components/composer/message-composer.tsx:264-290` 已提前清空文本、附件和 mentions，且只在 `onSubmit` reject 时恢复。因此这条分支仍被视为提交成功，用户内容会消失。请让不可发送分支 reject，或在清空 composer 前完成可发送校验，并补覆盖文本、附件、mentions 的回归测试。

2. **冷启动或切换 workspace 后，持久化的 `sending` 项仍会永久卡住。** `chat.tsx:298-303` 会落盘 `sending`；`apps/mobile/data/stores/chat-outbox-store.ts:51-61` 原样恢复；`apps/mobile/data/stores/chat-outbox.ts:48-56` 仅处理 `queued`；UI 对 `sending` 仍只有等待或取消。新增进程重启测试只覆盖 `queued`。请在 hydration 时把孤立 `sending` 归一化为安全、可人工决策的状态，并测试进程退出与 workspace 切换场景。

3. **删除 chat 仍不清理对应 draft/outbox。** `chat.tsx:583-602` 与 `apps/mobile/app/(app)/[workspace]/chat-sessions.tsx:31-48` 只触发服务端删除和选中态切换，没有在删除确认后清理该 session 的本地数据；远端删除路径也未覆盖。请统一处理所有删除入口，并避免服务端删除失败时提前丢本地内容。

4. **workspace/auth hydration 竞态仍存在。** `apps/mobile/app/(app)/[workspace]/_layout.tsx:125-147` 的 `cancelled` 只保护最后的 scope 标记，旧 scope 的异步 hydrate 仍可晚到并覆盖全局 store；`apps/mobile/data/auth-store.ts:56-100` 仍无 single-flight/generation guard，手动重试开始也没有重新设置 `isLoading`。请加入代际/作用域校验，并补并发 initialize、retry/logout 和乱序完成测试。

5. **query cache 清理竞态和未知 user ID 清理仍未修复。** `apps/mobile/data/query-persistence.ts:58-72` 未修改：unsubscribe 后已进入 throttle 队列的写入仍可能在 `removeItem` 之后重建缓存；最小复现顺序仍为 `set:1,remove,set:2`。`userId === null` 时仍不会删除任何持久化 query key。请取消或代际隔离待执行写入，并按受控前缀清理无法归属的旧分区，补 logout/401 竞态测试。

验证结果：mobile Vitest 15 files / 78 tests 通过，`pnpm typecheck` 通过，lint 为 0 errors / 6 个与本 PR 无关的 warnings，`git diff --check 0b660d371..a007fc434` 通过，GitHub 10/10 checks 通过。现有自动化尚未覆盖上述运行时失效路径，因此不能据此解除阻塞。

PR #6795, opened 2026-08-12
fix(daemon): resolve Windows installer junctions

### LINKED ISSUE #6762: [Bug]: Windows Codex runtime spawns codex.exe via the installer shim path, breaking sandbox helper resolution
## Summary

On Windows, the daemon launches the Codex runtime through the standalone installer's shim path
(`%LOCALAPPDATA%\Programs\OpenAI\Codex\bin\codex.exe`) instead of the real path that symlink
resolves to. Codex locates its sandbox helper binaries relative to the path it was invoked with,
and the shim directory does not contain them — so every sandboxed command fails and the agent
task cannot run.

Resolving the executable path (following symlinks/reparse points) before spawning fixes it.

## Environment

- OS: Windows 11 Enterprise 10.0.26200 (zh-CN)
- multica daemon: v0.4.22, `runtime_mode=local`, launched by desktop
- Codex: `codex-cli 0.147.0`, **standalone** install (`~/.codex/packages/standalone/...`)
- Not reproducible with the MSIX/desktop Codex build — see "Why only some installs" below.

## What happens

Any agent run on the Codex runtime fails as soon as it tries to execute a command. Surfaced to the
user as a native Windows dialog:

> Windows 找不到文件 'codex-windows-sandbox-setup.exe'。请确定文件名是否正确后，再试一次。
> ("Windows cannot find 'codex-windows-sandbox-setup.exe'")

Running the same thing from a terminal shows the underlying error:

```
windows sandbox failed: orchestrator_helper_launch_failed: setup refresh failed to launch helper:
helper=codex-windows-sandbox-setup.exe, cwd=E:\MyProject,
log=C:\Users\<user>\.codex\.sandbox\sandbox.2026-08-11.log, error=program not found
```

## Root cause

The daemon resolves `codex` off `PATH` and spawns that path verbatim (`daemon.log`):

```
INF agent command component=daemon exec=C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin\codex.exe args="[app-server --listen stdio://]"
```

`...\Programs\OpenAI\Codex\bin` is a **symlink** to
`~\.codex\packages\standalone\current\bin`. Codex resolves its sandbox helpers relative to the
invoked executable path, checking two locations only — next to the exe, and `bin\..\codex-resources`:

```
helper copy failed for command-runner: helper not found next to current executable or under
codex-resources: C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin\codex.exe;
falling back to legacy path codex-command-runner.exe
```

Both miss, because the shim directory contains *only* the `bin` symlink:

```
C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\
└── bin -> C:\Users\<user>\.codex\packages\standalone\current\bin      # no codex-resources\
```

whereas the real release directory has the helpers where Codex expects them:

```
C:\Users\<user>\.codex\packages\standalone\releases\0.147.0-x86_64-pc-windows-msvc\
├── bin\codex.exe
└── codex-resources\
    ├── codex-command-runner.exe
    └── codex-windows-sandbox-setup.exe
```

After the setup helper fails, the command-runner falls back to a bare name and the sandbox dies
one step later with `CreateProcessWithLogonW failed: 2` (ERROR_FILE_NOT_FOUND).

### Why only some installs

The MSIX/desktop Codex build ships helpers next to the executable, so its own resolution succeeds
(`...\WindowsApps\OpenAI.Codex_*\app\resources\codex-windows-sandbox-setup.exe`). Only the
standalone install — which is what `codex update` and the standalone installer produce, and what
lands on `PATH` — has the shim-directory indirection. So this hits Windows users who installed
Codex via the CLI installer rather than the desktop app.

## Reproduction

1. Windows, Codex installed via the standalone installer (`%LOCALAPPDATA%\Programs\OpenAI\Codex\bin`
   is a symlink, and the parent directory has no `codex-resources\`).
2. Assign any issue to an agent on the Codex runtime, or trigger a chat task that runs a command.
3. The task fails; the dialog above appears.

Minimal repro without Multica, using the same path the daemon uses:

```powershell
& "$env:LOCALAPPDATA\Programs\OpenAI\Codex\bin\codex.exe" sandbox -- cmd.exe /c echo hi
# -> windows sandbox failed: ... helper=codex-windows-sandbox-setup.exe ... error=program not found

& "$env:USERPROFILE\.codex\packages\standalone\current\bin\codex.exe" sandbox -- cmd.exe /c echo hi
# -> works: bin\..\codex-resources resolves correctly
```

## Local workaround (verified)

Create the missing directory next to the shim:

```
mklink /J "%LOCALAPPDATA%\Programs\OpenAI\Codex\codex-resources" ^
          "%USERPROFILE%\.codex\packages\standalone\current\codex-resources"
```

Verified: sandbox launches, `codex-command-runner-0.147.0.exe` is copied to `~\.codex\.sandbox-bin\`,
agent runs normally. Points at `current`, so it survives Codex upgrades — but it is a user-side
patch of a third-party install directory, not something every user should have to do.

## Proposed fix

**Primary — canonicalize the executable path before spawning.** After the `PATH` lookup, resolve
reparse points and spawn the resulting real path
(`~\.codex\packages\standalone\current\bin\codex.exe`), which makes Codex's own
`bin\..\codex-resources` lookup land in the release directory. In Go this is
`filepath.EvalSymlinks` on the `exec.LookPath` result; at the Win32 level,
`GetFinalPathNameByHandleW` (strip the `\\?\` prefix). Worth applying to all local runtimes on
Windows, not just Codex — the same shim pattern shows up in other CLI installers.

Note `current` itself is a symlink that Codex rotates on upgrade. Resolving at spawn time is
correct (each launch re-resolves); caching a resolved path across daemon restarts is not.

**Secondary — make the failure legible.** The user sees a raw Windows "file not found" dialog with
no indication of which component failed or which task it belongs to; nothing appears in the task
timeline. Whatever the outcome on the path fix, a failed runtime launch should surface as a task
error in the UI with the Codex stderr attached.

**Optional — preflight.** A startup check that the detected agent binary can actually start its
sandbox would turn this into an actionable warning instead of a mid-task dialog.

## Upstream note

Arguably Codex should resolve helpers relative to the *resolved* executable path rather than the
invoked one — I plan to file that with `openai/codex` as well. But the Multica-side change above is
independent, cheap, and protects users regardless of when upstream lands.

## Acceptance criteria

- With a standalone Codex install on Windows, an agent task runs commands successfully with no
  manual `codex-resources` junction.
- `daemon.log` shows the spawned `exec=` path as the resolved release path, not the shim path.
- A runtime that fails to launch produces a task-level error in the UI, not just an OS dialog.

Jamison929611:
### Thinking path / root cause

Windows standalone Codex exposes its installer `bin` directory as an `IO_REPARSE_TAG_MOUNT_POINT` junction. The daemon already canonicalized PATH-discovered executables with `filepath.EvalSymlinks`, but Go 1.26.1 cannot traverse that junction shape and the helper silently kept the installer shim path. Separator-bearing configured commands also returned before canonicalization.

This change keeps the existing `EvalSymlinks` behavior on non-Windows systems and uses `GetFinalPathNameByHandleW` on Windows. Resolving the opened executable handle follows the installer junction (and the standalone `current` link) to the release executable, so Codex finds `bin\..\codex-resources`. Explicit configured paths use the same canonicalization only on Windows.

## Related Issue

Fixes #6762

## Type of Change

- [ ] Documentation update
- [x] CI / infrastructure

## Changes Made

- Add build-tagged executable canonicalization: `GetFinalPathNameByHandleW` on Windows, existing `EvalSymlinks` elsewhere.
- Canonicalize separator-bearing configured executable paths on Windows while preserving non-Windows behavior.
- Add Windows tests using a real `mklink /J` for PATH discovery and configured paths, plus regular-path, error-fallback, and DOS/UNC prefix cases.
- Run those regressions in the existing Windows CI job.

## How to Test

1. On Windows with Go 1.26.1: `cd server && go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathCanonicalizes|TestTrimWindowsExtendedPathPrefix)' -count=1 -timeout=5m`
2. On Windows: `cd server && go vet ./internal/daemon && go build ./...`
3. Cross-target check: `cd server && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/daemon && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go vet ./internal/daemon`

Local results: all commands above completed successfully with Go 1.26.1. The Windows regression tests create and resolve a real directory junction. This machine has the MSIX Codex install, not the standalone installer layout, so a read-only smoke test against an existing standalone Codex junction was not available locally.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; no UI change)
- [x] I have updated relevant documentation to reflect my changes (not applicable; no documented interface changes)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs** (not applicable)

## AI Disclosure

**Prompt / approach:** I used Codex to read the issue discussion and repository guidance, trace both executable-resolution entry points, reproduce the Windows junction failure with a test before implementation, add the minimal build-tagged Win32 fix, run Windows and cross-target verification, and request an independent code review before opening this PR.

## Screenshots (optional)

Not applicable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,6 +427,14 @@ jobs:
           go test ./internal/daemon/execenv -v -run '^(TestSeedUserCodexSkills|TestHydrateCodexSkills)' -count=1 -timeout=5m
           go test ./internal/daemon -v -run '^(TestCleanTaskArtifacts_DoesNotFollowDirectoryJunction|TestTaskSize_DoesNotCountDirectoryJunction)$' -count=1 -timeout=5m
 
+      - name: Test Windows agent executable junction resolution
+        working-directory: server
+        # The standalone Codex installer exposes bin as a directory junction.
+        # filepath.EvalSymlinks cannot traverse that reparse-point shape, so
+        # only a real Windows filesystem proves both PATH discovery and an
+        # explicitly configured path reach the release executable.
+        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathCanonicalizes|TestTrimWindowsExtendedPathPrefix)' -count=1 -timeout=5m
+
       - name: Build Windows CLI helper entrypoint
         working-directory: server
         run: go build ./cmd/multica
--- a/server/internal/daemon/canonical_path.go
+++ b/server/internal/daemon/canonical_path.go
@@ -1,0 +1,13 @@
+//go:build !windows
+
+package daemon
+
+import "path/filepath"
+
+func canonicalPath(path string) (string, error) {
+	return filepath.EvalSymlinks(path)
+}
+
+func canonicalConfiguredExecutablePath(path string) string {
+	return path
+}
--- a/server/internal/daemon/canonical_path_windows.go
+++ b/server/internal/daemon/canonical_path_windows.go
@@ -1,0 +1,54 @@
+//go:build windows
+
+package daemon
+
+import (
+	"strings"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+func canonicalPath(path string) (string, error) {
+	pathUTF16, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return "", err
+	}
+	handle, err := windows.CreateFile(
+		pathUTF16,
+		0,
+		windows.FILE_SHARE_READ|windows.FILE_SHARE_WRITE|windows.FILE_SHARE_DELETE,
+		nil,
+		windows.OPEN_EXISTING,
+		windows.FILE_ATTRIBUTE_NORMAL,
+		0,
+	)
+	if err != nil {
+		return "", err
+	}
+	defer windows.CloseHandle(handle)
+
+	buffer := make([]uint16, syscall.MAX_PATH)
+	for {
+		length, err := windows.GetFinalPathNameByHandle(handle, &buffer[0], uint32(len(buffer)), 0)
+		if err != nil {
+			return "", err
+		}
+		if length < uint32(len(buffer)) {
+			resolved := windows.UTF16ToString(buffer[:length])
+			return trimWindowsExtendedPathPrefix(resolved), nil
+		}
+		buffer = make([]uint16, length+1)
+	}
+}
+
+func trimWindowsExtendedPathPrefix(path string) string {
+	if strings.HasPrefix(path, `\\?\UNC\`) {
+		return `\\` + strings.TrimPrefix(path, `\\?\UNC\`)
+	}
+	return strings.TrimPrefix(path, `\\?\`)
+}
+
+func canonicalConfiguredExecutablePath(path string) string {
+	return canonicalExecutablePath(path)
+}
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -641,7 +641,7 @@ func resolveAgentExecutablePath(cmd string) (string, error) {
 		return "", err
 	}
 	if strings.ContainsAny(cmd, "/\\") {
-		return resolved, nil
+		return canonicalConfiguredExecutablePath(resolved), nil
 	}
 	if isInMulticaHooksDir(resolved) {
 		if unshadowed, err := lookPathExcludingMulticaHooks(cmd); err == nil {
@@ -748,7 +748,7 @@ func canonicalExecutablePath(path string) string {
 	if err != nil {
 		return path
 	}
-	if real, err := filepath.EvalSymlinks(abs); err == nil {
+	if real, err := canonicalPath(abs); err == nil {
 		return real
 	}
 	return abs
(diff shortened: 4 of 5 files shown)


Jamison929611:
Follow-up pushed after diagnosing the first windows-execenv run. The Win32 resolver was returning the correct final file, but the runner exposed the expected temp path as an 8.3 alias (RUNNER~1) while GetFinalPathNameByHandleW returned the long path (
unneradmin), so textual EqualFold assertions produced a false failure.

The tests now first reject the unresolved junction/shim path, then compare the resolved and release executables with os.SameFile. This still fails against the original bug while accepting equivalent Windows short/long path spellings. The targeted Windows test command, daemon vet/build, Linux cross-target compile/vet, and an independent review all passed after rebasing onto current origin/main.

Issue #4204, opened 2026-06-16
Bug: Agent comments silently misattributed to workspace owner when subprocess loses MULTICA_TOKEN env var


MSandro:
## Summary

Agent-authored comments are intermittently attributed to the **workspace owner's member account** instead of the **agent** identity. This happens when an agent subprocess loses the `MULTICA_TOKEN` / `MULTICA_AGENT_ID` / `MULTICA_TASK_ID` environment variables and the CLI silently falls back to the user's config-file PAT (`mul_` prefix), causing the server to resolve the actor as `member` instead of `agent`.

## Root Cause

The CLI's token resolution (`resolveToken()` in `server/cmd/multica/cmd_auth.go:70`) has this fallback chain:

```
1. MULTICA_TOKEN env var     → use task-scoped mat_ token (correct)
2. inAgentExecutionContext() → if true, return "" (forces error — correct)
3. ~/.multica/config.json    → fall back to user's mul_ PAT (SILENT MISATTRIBUTION)
```

Step 3 is the bug. When ALL THREE env vars (`MULTICA_TOKEN`, `MULTICA_AGENT_ID`, `MULTICA_TASK_ID`) are missing from a subprocess, `inAgentExecutionContext()` returns `false` (it checks `MULTICA_AGENT_ID` and `MULTICA_TASK_ID`), so the CLI falls through to the config file and authenticates as the **workspace owner** using their personal access token.

On the server side, `resolveActor()` (`server/internal/handler/handler.go:357`) then sees:
- No `X-Actor-Source: task_token` (because the request uses a `mul_` PAT, not a `mat_` task token)
- No `X-Agent-ID` header (because `MULTICA_AGENT_ID` env var is empty, so the CLI doesn't set it)
- → Returns `("member", userID)` — the workspace owner

## Trigger Condition

This is triggered by agent runtimes that spawn subprocesses **without forwarding all `MULTICA_*` environment variables**. Confirmed with Hermes Agent's `execute_code` tool:

- The Hermes ACP process correctly receives all env vars from the daemon
- But the `execute_code` Python sandbox does **not** inherit `MULTICA_TOKEN`, `MULTICA_AGENT_ID`, `MULTICA_TASK_ID`, or `MULTICA_WORKSPACE_ID`
- Any `multica issue comment add` call from within `execute_code` uses the config-file PAT and is attributed as `member`

Other agent runtimes or tool integrations that filter/reset the environment for subprocesses would trigger the same bug.

[...]

PR #6800, opened 2026-08-12
MUL-6036: fix(cli): stop treating daemon port as task context

### LINKED ISSUE #6779: multica: `MULTICA_DAEMON_PORT` in env breaks `multica login` on v0.4.22+
- Репозиторий: multica-ai/multica
- Тип: bug (regression v0.4.21 → v0.4.22)
- Статус: готов к отправке (2026-08-11)
- Воспроизведено на: multica CLI 0.4.22 / 0.4.23 (linux-amd64), self-hosted backend 0.4.22

## Title

`MULTICA_DAEMON_PORT` in the environment makes `multica login` fail with "login is not available inside a daemon-managed task" (regression from v0.4.21)

## Body

**Context.** We run self-hosted multica with agent daemons in containers (network_mode: service:). The container compose files historically set `MULTICA_DAEMON_PORT` (e.g. `20032`) as an explicit port override. Entrypoints authenticate with `multica login --token <PAT> --profile <name>`.

**Regression.** On CLI **v0.4.21** `multica login --token ... --profile foo` succeeds even when `MULTICA_DAEMON_PORT` is present in the environment. On CLI **v0.4.22+** the same command fails immediately:

```
$ MULTICA_DAEMON_PORT=20032 multica login --token <token> --profile foo
login is not available inside a daemon-managed task
```

Root cause (from source): `inDaemonManagedExecutionContext()` (cmd_agent.go) returns true when `os.Getenv("MULTICA_DAEMON_PORT") != ""`, and the login command refuses to run in that context. The env var was only ever a *display/hint* for health-port bookkeeping; the actual health port is derived deterministically from the profile name (`19514 + 1 + sum(bytes(profile)) % 1000`), so setting the env var is unnecessary and now actively harmful.

**Impact.** Anyone upgrading a containerized daemon setup from 0.4.21 to 0.4.22+ with `MULTICA_DAEMON_PORT` in the environment gets a non-auth'd daemon at boot; the entrypoint retry loop burns ~5 minutes then exits.

**Workaround (ours).** Remove `MULTICA_DAEMON_PORT` from container env; the daemon picks the correct profile-derived port automatically.

**Suggested fix.** Either (a) don't treat `MULTICA_DAEMON_PORT` as a daemon-context marker — use an explicit marker (e.g. `MULTICA_TASK_CONFIG_ROOT` / `MULTICA_TASK_ID`, which are already set inside tasks), or (b) allow `login --token` (non-interactive) to bypass the daemon-context check, or (c) at minimum document that `MULTICA_DAEMON_PORT` must not be set in daemon container environments since v0.4.22.

Jamison929611:
### Summary

- Stops treating `MULTICA_DAEMON_PORT` by itself as proof that the CLI is running inside a daemon-managed task.
- Keeps login, token/workspace isolation, daemon status, disk-usage, and repo checkout behavior scoped to their actual markers and endpoint needs.
- Adds regression coverage for token and interactive login guards, authoritative task markers, whitespace/marker combinations, explicit daemon status endpoints, and repo checkout.

### Root cause

v0.4.22 added a correct fail-closed guard for human-local commands such as `login`, but the shared task-context predicate also treated `MULTICA_DAEMON_PORT` as task identity. Containerized daemon startup environments may set that variable only as a health endpoint override, so the login guard rejected ordinary startup authentication.

The predicate now uses trimmed agent/task identity, the daemon-owned task config root, or the daemon-owned workdir/root marker. The port remains a direct health/checkout endpoint input rather than task identity.

### Risk

The main risk is reopening the silent owner-PAT fallback fixed by #4204/#5044. Existing root/workdir marker tests and new task-config/identity tests verify that real or escaped daemon task processes still fail closed. Explicit-port daemon status also bypasses profile-hash collision checks because the endpoint was selected directly, while profile-derived status keeps collision detection.

## Related Issue

Fixes #6779

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/cmd/multica/cmd_agent.go`: distinguish authoritative task markers from the daemon health port.
- `server/cmd/multica/cmd_auth.go`: retain fail-closed token fallback using the corrected task predicate.
- `server/cmd/multica/cmd_daemon.go`: preserve explicit-port status behavior without false profile collisions.
- CLI tests: cover port-only login, real task rejection, whitespace/combinations, status, disk-usage isolation, and repo checkout.

## How to Test

1. Run the focused CLI regression suite for login/task markers, daemon status, disk usage, and repo checkout.
2. Run the daemon task-environment and execenv root/workdir marker tests.
3. Run `go vet ./cmd/multica ./internal/daemon ./internal/daemon/execenv`.

Local results:
- Focused `server/cmd/multica` regression tests: pass.
- `server/internal/daemon` task environment test: pass.
- `server/internal/daemon/execenv` marker tests: pass.
- Targeted `go vet`: pass.

A full `go test ./cmd/multica` on this Windows checkout still hits pre-existing tests that assume Unix absolute paths or that setting only `HOME` changes `os.UserHomeDir`; the same failures reproduce from a clean `origin/main` worktree. The affected #6779 paths are covered by the passing focused suite above.

## Checklist

Documentation note: the built-in runtime/repo skill documents `MULTICA_DAEMON_PORT` as the direct endpoint required by `repo checkout`; that contract is unchanged, so no built-in skill text needed modification.

## AI Disclosure

**Prompt / approach:**

Used Codex to trace the v0.4.21 to v0.4.22 history and every task-context caller, implement the fix test-first, compare full-package Windows failures against a clean upstream worktree, and run an independent code-review pass. The review found an explicit-port daemon status collision edge case, which was reproduced with a failing end-to-end test and fixed before submission.

## Screenshots (optional)

Not applicable (CLI-only change).

### DIFF
--- a/server/cmd/multica/cmd_agent.go
+++ b/server/cmd/multica/cmd_agent.go
@@ -255,12 +255,13 @@ func newAPIClient(cmd *cobra.Command) (*cli.APIClient, error) {
 	taskContext := inDaemonManagedExecutionContext()
 	token := resolveToken(cmd)
 	if taskContext && !strings.HasPrefix(token, "mat_") {
-		// When the ONLY daemon signal is a workdir marker (no MULTICA_AGENT_ID /
-		// MULTICA_TASK_ID / MULTICA_DAEMON_PORT), the likeliest cause outside a
-		// real task is a leftover marker from a crashed daemon task in a
-		// local_directory. Name the exact file so a normal user can recover
-		// instead of hitting an opaque "requires mat_ token" error.
-		if !inAgentExecutionContext() && os.Getenv("MULTICA_DAEMON_PORT") == "" {
+		// When the only task signal is a workdir marker, the likeliest cause
+		// outside a real task is a leftover marker from a crashed daemon task in
+		// a local_directory. MULTICA_DAEMON_PORT does not change that conclusion:
+		// it may be a container-level health endpoint override. Name the exact
+		// marker so a normal user can recover instead of hitting an opaque
+		// "requires mat_ token" error.
+		if !inAgentExecutionContext() && strings.TrimSpace(os.Getenv(cli.TaskConfigRootEnv)) == "" {
 			if markerPath := daemonTaskContextMarkerPath(); markerPath != "" {
 				return nil, fmt.Errorf("agent execution context requires MULTICA_TOKEN to be a task-scoped mat_ token; detected a daemon task marker at %s — if you are not running inside an agent task this is likely a leftover, remove it and retry", markerPath)
 			}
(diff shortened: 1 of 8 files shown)


multica-eve:
Thank you for the careful investigation and the especially thorough regression coverage in this PR. The tracing across task markers, login modes, daemon status, disk usage, and repo checkout was genuinely valuable and helped us validate the edge cases around this regression.

We reviewed this implementation alongside #6801 and decided to merge #6801 because it keeps the change more narrowly scoped: guarded human login is restored, while `MULTICA_DAEMON_PORT` remains a defense-in-depth signal for ordinary API/profile resolution, and a port-only host environment derives daemon status from the selected profile. That better matches the deployment behavior described in #6779 while preserving the existing fail-closed boundary.

Since #6801 now resolves the issue, we are closing this PR to avoid maintaining two competing implementations. We really appreciate the time, rigor, and clear write-up you put into this contribution—thank you again.

PR #6802, opened 2026-08-12
MUL-6025: fix(daemon): resolve Windows installer junctions

Jamison929611:
﻿## What does this PR do?

### Thinking path / root cause

Windows standalone Codex exposes its installer `bin` directory as an `IO_REPARSE_TAG_MOUNT_POINT` junction. The daemon already canonicalized PATH-discovered executables with `filepath.EvalSymlinks`, but Go 1.26.1 cannot traverse that junction shape and the helper silently kept the installer shim path. Separator-bearing configured commands also returned before canonicalization.

This change keeps the existing `EvalSymlinks` discovery behavior on non-Windows systems. On Windows, discovery retains the stable configured/PATH entry point and each task launch resolves its opened handle through `GetFinalPathNameByHandleW`. When the installer retargets `current`, the daemon re-detects and minimum-version-checks the new release, then atomically publishes the matching `{path, version}` pair. Concurrent and repeatedly changing targets fail closed rather than launching an unverified or stale binary.

The rediscovery path now applies the same strict boundary: if final-path resolution fails after a pinned path vanished, the daemon returns a launch failure immediately. It does not call version detection, does not adopt the rediscovered stable junction, does not write `resolvedPaths`, and does not launch the unresolved entry point. Both initial and rediscovery resolution failures now log at Warn with path and error.

The Win32 final path keeps its extended-length `\\?\` / `\\?\UNC\` form internally. This prevents a short shim from resolving into an unlaunchable release path beyond `MAX_PATH`; the Windows regression actually spawns an executable from such a path.

Runtime process-start errors already flow through the daemon task terminal callback. This PR makes direct Windows Codex `CreateProcess` failures an explicit `agent_error.process_failure`, verifies the real `cmd.Start` -> `executeAndDrain` -> `handleTask` -> `/fail` path with an invalid PE, and logs executable canonicalization fallbacks with both path and error. This direct process-start coverage is not the same as the post-start grandchild sandbox-helper native dialog; that remaining acceptance item is tracked separately in #6849.

## Related Issues

Refs #6762

Follow-up for the scoped-out grandchild native-dialog/task-failure boundary: #6849

## Type of Change

- [ ] Documentation update
- [x] CI / infrastructure

## Changes Made

- Add build-tagged Windows final-path resolution using `GetFinalPathNameByHandleW`, while preserving non-Windows discovery behavior.
- Keep stable Windows installer entry points in discovery and resolve/version-gate their current target at every launch.
- Preserve extended-length paths internally and prove an actual executable launch beyond `MAX_PATH`.
- Fail closed on first-launch detection/minimum-version failures, rediscovery final-path resolution failures, concurrent retargets, and bounded repeated retargets.
- Log canonicalization and launch-resolution fallback errors with path + error through the existing logger.
- Surface real direct Windows Codex process-start failures through the task `/fail` callback as `agent_error.process_failure`.
- Run all Windows junction, long-path, launch-boundary, and launch-error regressions in the existing `windows-execenv` job.

## How to Test

1. Windows / Go 1.26.1:
   `cd server && go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|ForLaunchRejectsRediscoveredJunctionWhenFinalPathResolutionFails|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting)|TestHandleTaskReportsWindowsCodexProcessStartFailure)' -count=1 -timeout=5m`
2. Windows: `cd server && go test ./pkg/taskfailure -count=1`
3. Windows: `cd server && go vet ./internal/daemon ./pkg/taskfailure && go build ./...`
4. Linux cross-target: `cd server && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/daemon && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go vet ./internal/daemon ./pkg/taskfailure`

Local results: all commands above completed successfully with Go 1.26.1. The Windows tests create real directory junction chains, retain v1 while retargeting only `current` to v2, launch v2 with its freshly detected version, exercise concurrent and repeatedly changing targets, reject an initial unverified junction target, reject a rediscovered junction target when final-path resolution fails, launch a resolved executable beyond `MAX_PATH`, and send an actual invalid-PE direct `CreateProcess` error through the task failure endpoint.

A full local `go test ./internal/daemon` was also attempted; unrelated environment-dependent tests failed against this machine's installed agent CLIs/skills/MCP configuration and Windows symlink privileges, while the task-specific selectors, package tests, vet/build, and cross-target checks above passed. This machine has the MSIX Codex install rather than the standalone installer layout, so no additional read-only smoke test against an existing standalone Codex junction was available.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; no UI change)
- [x] I have updated relevant documentation to reflect my changes (not applicable; no documented interface changes)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs** (not applicable)

## AI Disclosure

**Prompt / approach:** I used Codex to read the issue and maintainer review, reproduce the Windows junction lifecycle and launch-error gaps with failing tests before implementation, preserve stable discovery and long-path semantics, run Windows and cross-target verification, and request independent code reviews after each commit.

## Screenshots (optional)

Not applicable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,7 +433,7 @@ jobs:
         # filepath.EvalSymlinks cannot traverse that reparse-point shape, so
         # only a real Windows filesystem proves both PATH discovery and an
         # explicitly configured path reach the release executable.
-        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathCanonicalizes|TestTrimWindowsExtendedPathPrefix)' -count=1 -timeout=5m
+        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting))' -count=1 -timeout=5m
 
       - name: Build Windows CLI helper entrypoint
         working-directory: server
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,7 +433,7 @@ jobs:
         # filepath.EvalSymlinks cannot traverse that reparse-point shape, so
         # only a real Windows filesystem proves both PATH discovery and an
         # explicitly configured path reach the release executable.
-        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting))' -count=1 -timeout=5m
+        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting)|TestHandleTaskReportsWindowsCodexProcessStartFailure)' -count=1 -timeout=5m
 
       - name: Build Windows CLI helper entrypoint
         working-directory: server
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,6 +427,14 @@ jobs:
           go test ./internal/daemon/execenv -v -run '^(TestSeedUserCodexSkills|TestHydrateCodexSkills)' -count=1 -timeout=5m
           go test ./internal/daemon -v -run '^(TestCleanTaskArtifacts_DoesNotFollowDirectoryJunction|TestTaskSize_DoesNotCountDirectoryJunction)$' -count=1 -timeout=5m
 
+      - name: Test Windows agent executable junction resolution
+        working-directory: server
+        # The standalone Codex installer exposes bin as a directory junction.
+        # filepath.EvalSymlinks cannot traverse that reparse-point shape, so
+        # only a real Windows filesystem proves both PATH discovery and an
+        # explicitly configured path reach the release executable.
+        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathCanonicalizes|TestTrimWindowsExtendedPathPrefix)' -count=1 -timeout=5m
+
       - name: Build Windows CLI helper entrypoint
         working-directory: server
         run: go build ./cmd/multica
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,7 +433,7 @@ jobs:
         # filepath.EvalSymlinks cannot traverse that reparse-point shape, so
         # only a real Windows filesystem proves both PATH discovery and an
         # explicitly configured path reach the release executable.
-        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting)|TestHandleTaskReportsWindowsCodexProcessStartFailure)' -count=1 -timeout=5m
+        run: go test ./internal/daemon -v -run '^(TestCanonicalExecutablePath|TestResolveAgentExecutablePathKeeps|TestResolveAgentEntry(FollowsRetargetedInstallerJunction|CanonicalizesRediscoveredJunction|ForLaunchRejectsUnverifiedInitialJunctionTarget|ForLaunchRejectsRediscoveredJunctionWhenFinalPathResolutionFails|DoesNotSharePreRetargetSingleflightResult|ForLaunchFailsWhenJunctionKeepsRetargeting)|TestHandleTaskReportsWindowsCodexProcessStartFailure)' -count=1 -timeout=5m
 
       - name: Build Windows CLI helper entrypoint
         working-directory: server
--- a/server/internal/daemon/agents_refresh.go
+++ b/server/internal/daemon/agents_refresh.go
@@ -205,11 +205,11 @@ func (d *Daemon) refreshAgentAvailability() []string {
 // NOT restart. What a user needs when codex or claude upgrades is that
 // subsequent tasks run the new CLI under the new version's rules — not that
 // Multica's availability tracks a third party's release cadence. An in-place
-// upgrade leaves the pinned path valid, so the new binary is already what
-// launches; the two things left stale are the cached version (which keys
-// version-sensitive policy such as the Codex sandbox, read back through
-// resolveAgentEntry) and the version the server displays. Both are refreshed
-// here with running tasks untouched.
+// POSIX upgrade changes the binary behind the pinned path; a Windows installer
+// upgrade retargets the stable junction resolved by resolveAgentEntry. The two
+// things left stale are the cached version (which keys version-sensitive policy
+// such as the Codex sandbox) and the version the server displays. Both are
+// refreshed here with running tasks untouched.
 //
 // detectBuiltinRuntimes does the probing, which buys three properties for free:
 // probes fan out, a fast failure is retried (runtimeVersionProbeAttempts), and
--- a/server/internal/daemon/canonical_path.go
+++ b/server/internal/daemon/canonical_path.go
@@ -1,0 +1,13 @@
+//go:build !windows
+
+package daemon
+
+import "path/filepath"
+
+func canonicalPath(path string) (string, error) {
+	return filepath.EvalSymlinks(path)
+}
+
+func canonicalConfiguredExecutablePath(path string) string {
+	return path
+}
--- a/server/internal/daemon/canonical_path.go
+++ b/server/internal/daemon/canonical_path.go
@@ -8,6 +8,14 @@ func canonicalPath(path string) (string, error) {
 	return filepath.EvalSymlinks(path)
 }
 
+func discoveredExecutablePath(path string) string {
+	return canonicalExecutablePath(path)
+}
+
+func executablePathForLaunch(string) (string, bool, error) {
+	return "", false, nil
+}
+
 func canonicalConfiguredExecutablePath(path string) string {
 	return path
 }
--- a/server/internal/daemon/canonical_path.go
+++ b/server/internal/daemon/canonical_path.go
@@ -12,7 +12,9 @@ func discoveredExecutablePath(path string) string {
 	return canonicalExecutablePath(path)
 }
 
-func executablePathForLaunch(string) (string, bool, error) {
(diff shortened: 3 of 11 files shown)


Bohan-J (changes requested):
Thanks — the Win32 core here is solid, and putting a real `mklink /J` test into the existing Windows CI job is exactly the right way to prove this shape. Specifically good: the build-tag split keeps non-Windows behavior identical; `CreateFileW` uses zero access rights with all three share flags, so opening a running executable can't fail; the handle is closed; the buffer-growth loop uses the correct `length < len(buffer)` success condition; and the junction test asserts `EvalSymlinks` fails *first*, so it can't silently stop reproducing the bug. No new dependency either — `golang.org/x/sys` is already direct.

One blocker, and one thing I would like proven before merge.

## Blocker: after this change, a Codex upgrade never takes effect

`canonicalPath` resolves all the way through both junctions (`visible\bin` → `current\bin` → `releases\<version>\bin`), so `resolveAgentExecutablePath` now pins `AgentEntry.Path` to a **version-specific** path.

The Codex Windows installer retains old releases. At `2230d64`, `Remove-StaleInstallArtifacts` deletes only `.staging.*` (`install.ps1` L303-L311), and the only other release-directory removal is a same-version reinstall (L1004). An upgrade just retargets the junction: `Ensure-Junction -LinkPath $currentDir -TargetPath $releaseDir` (L1014).

So after `codex update`:

- `releases\<old>\bin\codex.exe` still exists on disk;
- `resolveAgentEntryWithHeal` returns the pinned path unchanged whenever that file is still present (`daemon.go:866-868`), so nothing re-resolves;
- the MUL-4486 self-heal never fires — it is triggered by the pinned path *disappearing*, which is what version managers on macOS/Linux do and what this installer deliberately does not do;
- `refreshAgentAvailability` doesn't rescue it either: it replaces entries only for *newly gained* providers and preserves known ones as-is (`agents_refresh.go:174-194`).

The daemon therefore keeps launching the old Codex indefinitely. This is a regression introduced by the fix: before this PR, Windows kept the shim path — a stable entry point that automatically followed `current`, so upgrades were picked up. It also breaks the invariant the code documents at `agents_refresh.go:207-209` ("An in-place upgrade leaves the pinned path valid, so the new binary is already what launches").

Suggested shape: keep the stable discovered entry point in `AgentEntry.Path`, and resolve to the final executable at launch time in `resolveAgentEntryWithHeal` — the single choke point every task goes through. When the resolved target changes, re-detect the version and publish `{path, version}` atomically, exactly as the existing self-heal already does.

The regression test that would catch it:

1. `visible\bin` → `current\bin` → `releases\v1\bin`;
2. keep v1 on disk, create v2, retarget only `current`;
3. the next launch must run v2, and the returned version must be v2's.

Both existing tests miss this case: `agent_path_selfheal_test.go:104-111` deletes the old versioned directory first, and the new Windows tests use a single static target.

## Please prove or avoid: unconditional `\\?\` stripping

`trimWindowsExtendedPathPrefix` drops `\\?\` / `\\?\UNC\` unconditionally. `syscall.StartProcess` passes argv0 straight to `CreateProcessW` with no `fixLongPath` (`syscall/exec_windows.go:339,441`), and the Go linker embeds no `longPathAware` manifest, so that prefix is the only thing that reliably makes an executable beyond `MAX_PATH` launchable.

In fairness this is not a regression — the pre-PR path wasn't prefixed either. What is new is that **resolution can lengthen a path**: a short shim can resolve into a deep `releases\<version>-x86_64-pc-windows-msvc\bin\`, so this change newly creates cases that cross `MAX_PATH` where the original did not. And `canonicalExecutablePath` is generic across every Windows runtime, not just Codex.

`TestTrimWindowsExtendedPathPrefix` only asserts string rewriting. Either keep the extended-length form internally and normalize only for display/logging, or add a test that actually spawns an executable at a path longer than `MAX_PATH`, including the UNC case.

## Non-blocking

- `config.go:751-754` still swallows the canonicalization error and falls back to the input path. That silent fallback is precisely why this bug stayed invisible on Windows — worth a debug/warn log carrying the path and the error.
- The body says `Fixes #6762`, but that issue's third acceptance criterion (a failed runtime launch surfacing as a task-level error rather than an OS dialog) is not in this PR. Either implement it, or switch to a plain reference and open a follow-up, so merging doesn't auto-close a partially-addressed issue.
- Consider retitling to include `MUL-6025` for issue linkage.

Jamison929611:
Implemented the requested follow-up in two commits and pushed the updated branch:

- Windows discovery now retains the stable installer entry point; each launch resolves the current final target, re-detects/version-gates target changes, and atomically publishes the matching path/version. It fails closed on unverified, concurrent-retarget, or repeatedly changing targets.
- Extended-length paths remain intact internally; the regression suite actually launches an executable beyond `MAX_PATH`.
- Canonicalization fallbacks now log path + error.
- A real invalid Windows PE now exercises Codex `cmd.Start` through `handleTask` and the task `/fail` callback, producing a visible `agent_error.process_failure` without crashing the daemon.

Local verification with Go 1.26.1 passed the CI-selected Windows regression set, `go test ./pkg/taskfailure`, `go vet ./internal/daemon ./pkg/taskfailure`, `go build ./...`, and Linux/amd64 cross-compile + vet. The full local daemon package attempt still encounters unrelated machine-dependent tests (installed agent CLI/skills/MCP state and Windows symlink privileges), which is documented in the PR body.

Please re-review commits d4afe2309 and 35b9c1ee1 when convenient.

Bohan-J (changes requested):
Both blockers from the previous round are properly resolved — thanks for the thorough rework.

Confirmed fixed:

- Windows discovery now keeps the stable entry point (`discoveredExecutablePath` is `filepath.Abs` only) and resolution happens per launch, so a retarget takes effect while the old release stays installed. `TestResolveAgentEntryFollowsRetargetedInstallerJunction` is exactly the scenario I asked for. The target-specific singleflight key, the re-resolve after version detection, and the retarget-loop cap are all sensible additions I didn't ask for and that close real races.
- The extended-length path is now retained internally, and `TestCanonicalExecutablePathLaunchesBeyondMaxPath` actually spawns a binary from beyond `MAX_PATH`. That closes the concern properly rather than asserting string rewriting.
- Canonicalization fallbacks are logged, the title carries `MUL-6025`, and the CI selector covers the new tests.

One remaining blocker.

## The rediscovery path fails open, bypassing the strict launch boundary

In `healAgentPath` (`daemon.go`, the `re-resolved after pinned path vanished` branch):

```go
if launchPath, handled, err := executablePathForLaunch(newPath); handled {
    if err != nil {
        d.logger.Debug("resolve re-discovered agent executable for launch failed; keeping discovered path", ...)
    } else {
        newPath = launchPath
    }
}
return d.adoptAgentPath(ctx, provider, command, newPath, "re-resolved after pinned path vanished")
```

On error, `newPath` stays the **unresolved stable junction**, and `adoptAgentPath` then publishes it into `resolvedPaths` as a verified `{path, version}` pair — the shim's `--version` succeeds, so nothing else objects.

The reason this defeats `resolveAgentEntryForLaunch` is the return value: `resolveAgentEntryWithHeal` returns `outcome` from this branch, not `launchOutcome`, so `outcome.failure` is nil. The strict boundary sees a clean success and the task launches from the junction — reproducing the exact helper-resolution failure this PR fixes. The junction can also retarget after the version check, which is the mismatch `resolveAgentLaunchTarget` was built to prevent.

A concrete way to reach it: `GetFinalPathNameByHandleW` is called with flags `0`, i.e. `VOLUME_NAME_DOS`, which fails when the volume has no drive-letter mount — a volume mounted at a folder path, or a mounted VHD. `CreateFileW` succeeds, the executable runs fine, `--version` works, and the resolution step is the only thing that fails. That user silently gets the shim back.

Worth deciding deliberately, because the two branches currently implement opposite policies: on the primary launch path the same failure sets `launchOutcome.failure`, so `resolveAgentEntryForLaunch` returns an error and **every** task hard-fails for that user — where before this PR they at least ran, minus the sandbox helper. Fail-closed is defensible, fail-open is not, but they should agree.

Suggested: return a `healOutcome` carrying `failure` instead of adopting, and add the regression — rediscovery where final-path resolution fails must neither publish to `resolvedPaths` nor launch the discovered path. Current coverage misses it from both sides: `TestResolveAgentEntryCanonicalizesRediscoveredJunction` only exercises the rediscovery happy path, and `TestResolveAgentEntryForLaunchRejectsUnverifiedInitialJunctionTarget` covers version-detection and minimum-version failures on the *initial* path, not resolution failure on rediscovery.

Also: both new failure branches log at `Debug`. A silently swallowed resolution error on Windows is precisely how the original bug stayed invisible — `Warn` would be more appropriate.

## Non-blocking: scope of the `Fixes #6762` claim

`TestHandleTaskReportsWindowsCodexProcessStartFailure` writes an invalid PE `codex.exe` and asserts the daemon reports the start failure to `/fail`; `taskfailure/classify.go` then refines it to `agent_error.process_failure`. That is a genuine improvement and worth keeping.

It does not, however, cover the acceptance criterion it is credited with. The dialog in #6762 comes from a grandchild — Codex starts successfully, then launches `codex-windows-sandbox-setup.exe`, which fails and raises a native Windows dialog the daemon can neither capture nor suppress. Since that criterion has been scoped out as separate work, I'm keeping this non-blocking: please switch `Fixes #6762` to a plain reference and open a follow-up, or record on #6762 that the criterion was split out.

Jamison929611:
Updated this PR for review `4914369485`.

What changed in the latest head (`80b3dbdce`):

- `healAgentPath` now fails closed when rediscovered Windows final-path resolution fails: it returns `healOutcome{failure: err}` before version detection/adoption, so it does not publish `resolvedPaths` and `resolveAgentEntryForLaunch` exits before launching the unresolved stable junction.
- The initial and rediscovery final-path resolution failure branches now log at Warn with the path and error.
- Added the focused Windows regression `TestResolveAgentEntryForLaunchRejectsRediscoveredJunctionWhenFinalPathResolutionFails`, and added it to the `windows-execenv` selector.
- Updated the PR body from `Fixes #6762` to `Refs #6762`; the direct invalid-PE coverage is described only as direct Codex `CreateProcess` failure handling.
- Created #6849 to track the separate post-start grandchild `codex-windows-sandbox-setup.exe` native-dialog/task-failure boundary.

Local verification rerun before push:

- focused blocker test
- full Windows junction selector
- `go vet ./internal/daemon`
- `go build ./...`
- `GOOS=linux GOARCH=amd64 go test -c ./internal/daemon`

Requesting re-review on the current head.

Bohan-J (approved):
Thanks for sticking with this through three rounds — the final state is genuinely good work.

The remaining blocker is properly closed: `healAgentPath` now returns `healOutcome{failure: err}` before version detection or `adoptAgentPath`, so the strict launch boundary sees the failure and refuses to launch. `TestResolveAgentEntryForLaunchRejectsRediscoveredJunctionWhenFinalPathResolutionFails` asserts the things that actually matter rather than just the return value — version detection is never called, nothing is written to `resolvedPaths`, the returned path isn't launchable, and the WARN log carries both path and error. Both failure branches are now WARN, so the two paths agree on fail-closed. `Refs #6762` plus the follow-up in #6849 scopes the claim correctly.

What I verified on the merge result (this branch merged with current `main`):

- clean merge, no conflicts;
- `GOOS=windows go vet ./internal/daemon` passes;
- the Windows test binary cross-compiles;
- targeted agent-path / self-heal / canonicalization tests pass;
- all 9 GitHub checks green, including the real-filesystem `windows-execenv` job.

Worth calling out what made this a good fix rather than a patch: keeping the stable entry point at discovery and resolving per launch was the harder route, and it is what makes installer retargets work while old releases stay on disk. The target-specific singleflight key, the re-resolve after version detection, and the retarget cap weren't asked for and each closes a real race. The original report from @liujiaqi958732-eng was excellent too — the `fsutil reparsepoint` confirmation is what pinned this down.

LGTM.

Issue #6804, opened 2026-08-12
[Bug]: Non-unique issue prefixes + multi-workspace installation fan-out (#5183) can cross-link/attribute PRs to the wrong workspace


AnkeshThakur:
### Deployment type

self-host

### What happened / what's the gap?

#2797 established that `workspace.issue_prefix` is **not** unique across workspaces by design — the maintainer's own words: "We decided **not** to enforce globally unique prefixes (different teams may legitimately want the same short identifier)". That decision was declared safe on one specific assumption, quoted from the same comment:

> a GitHub repository is bound to a single installation, which maps 1:1 to a workspace, so the PR webhook always resolves the correct workspace before matching prefix + number. The collision doesn't cross that boundary.

#5183 (implemented in #5218, merged 2026-07-10) removed that exact assumption: PR/`check_suite` webhook events now fan out to **every** workspace bound to a shared GitHub App installation, not just the single workspace that owns the repo. #5183's own "Open questions" section flagged this but shipped anyway:

> Issue-prefix collisions: if two workspaces share a prefix, the same PR reference would link in both — probably fine, since each link row is workspace-scoped.

Concretely, today, with two workspaces `A` (prefix `MUL`) and `B` (prefix `MUL`, collision permitted per #2797) both bound to the same GitHub App installation (per #4855/#5183's multi-workspace binding model):

1. A PR titled `MUL-12 fix bug` is opened in a repo the installation covers.
2. The webhook fans out to both `A` and `B` (`handlePullRequestEvent` → `mirrorPullRequestForWorkspace` per bound workspace, unconditional on repo ownership — see MUL-4343 comment in `server/internal/handler/github.go`).
3. **`github_pull_request` mirror row**: written into *both* workspaces unconditionally (not gated by prefix or repo registry) — so `B` gets a PR-count/mirror row for a PR that has nothing to do with it, independent of any prefix collision.
4. **Issue auto-link**: each workspace independently resolves `MUL-12` against its *own* issue table (`lookupIssueByIdentifier(ctx, wsID, prefix, id)`). If both `A` and `B` happen to have an issue numbered 12, the PR silently auto-links (and, depending on `close_intent`, can auto-advance to Done) against the **wrong** issue in whichever workspace doesn't actually own that PR — not just "linked in both" as #5183 assumed, but a real misattribution if the numbers happen to coincide.

[...]

Issue #5183, opened 2026-07-09
[Feature]: Fan out PR webhook events to every workspace bound to a GitHub installation


franciscocpg:
### Deployment type

self-host

### What do you want and why?

Follow-up to #4823 / #4855. That fix made the GitHub **binding** multi-workspace: one GitHub App installation can now be connected to several workspaces without the second connect silently unbinding the first, and both workspaces show "Connected".

However, webhook **delivery** is still single-workspace per repo. Each `pull_request` (and `check_suite`) event is routed to exactly one workspace by `resolveWorkspaceForRepo` (`server/internal/handler/github.go`):

1. If a workspace's repos registry (`workspace.repos`, Settings → Repositories) contains `github.com/owner/name`, that workspace wins.
2. Otherwise the event falls back to the installation's **oldest binding** (`ListGitHubInstallationsByInstallationID` is `ORDER BY created_at ASC`, handlers take `insts[0]`).
3. If several workspaces register the same repo, the tie-break again favors the oldest binding.

The PR mirror row and the issue↔PR auto-link rows are then created only in that single winning workspace, and the auto-link scan only matches that workspace's issue prefix.

Concrete consequence: with two workspaces bound to the same installation and working against the **same repository**, only one workspace ever gets PRs linked in the issue sidebar. The other workspace shows the integration as "Connected" but never links anything, and there is no configuration that can fix it — the repos registry can only move the single delivery, not share it. PR titles/branches referencing the second workspace's issue prefix (e.g. `ABC-123`) are never scanned, because that workspace never receives the event.

This is the natural next step after #4823: that issue asked for installations to coexist across workspaces, which #4855 delivered for the binding; this issue asks for event delivery to follow.

### Proposed solution (optional)

Fan the PR/check_suite event out to **every** workspace bound to the installation (or at least every bound workspace whose repos registry matches, plus the registry/fallback resolution for backward compatibility):

- Upsert the PR mirror row per delivered workspace (`github_pull_request` is already workspace-scoped).
- Run the auto-link scan per workspace, each against **its own** issue prefix (`getIssuePrefix(ctx, wsID)`), gated by each workspace's own `github_enabled` / `github_auto_link_prs_enabled` settings.
- Broadcast realtime updates to each affected workspace, as `installation.deleted` already does since #4855.

Open questions:

[...]

Issue #4823, opened 2026-07-01
[Bug]: Connecting GitHub in one workspace silently disconnects it from another workspace


franciscocpg:
### Deployment type

self-host

### What happened?

Connecting the GitHub integration in one workspace silently disconnects it from another workspace on the same instance.

I have a self-hosted installation with two workspaces. Following the docs, I created a single GitHub App for the server. When I connect GitHub in workspace A, it works. When I then connect GitHub in workspace B (same GitHub account/org), workspace B connects fine — but workspace A's GitHub integration is silently unbound. Reconnecting in workspace A steals it back and disconnects workspace B. There is no warning in the UI in either direction.

I expected either (a) both workspaces to be able to use the same GitHub installation, or (b) an explicit error/warning that the installation is already bound to another workspace — not a silent unbind.

**Root cause (from reading the source):**

- `server/migrations/079_github_integration.up.sql` defines `github_installation` with `UNIQUE (installation_id)` and a single `workspace_id` column, so a GitHub installation can only ever be bound to one workspace.
- `server/pkg/db/queries/github.sql` — `CreateGitHubInstallation` does `ON CONFLICT (installation_id) DO UPDATE SET workspace_id = EXCLUDED.workspace_id, ...`, so connecting from a second workspace *moves* the row to the new workspace instead of failing or coexisting.

This is especially limiting for self-host: a GitHub App can only be installed once per GitHub account/org, so every workspace connecting to the same org reuses the same `installation_id` and they can only steal the binding from each other. The docs (https://multica.ai/docs/github-integration) say to create one GitHub App for the server, but don't mention this one-workspace-per-installation limitation.

**Suggested fix:** allow an installation to be bound to multiple workspaces (e.g. change the constraint to `UNIQUE (workspace_id, installation_id)` and fan out webhook events to all bound workspaces), or at minimum surface a clear warning/error when connecting an installation that is already bound to another workspace.

### Steps to reproduce

1. Self-host Multica with one GitHub App configured, and create two workspaces (A and B).
2. In workspace A: Settings → GitHub → Connect GitHub, complete the App installation flow for a GitHub org/account. Workspace A shows connected.
3. In workspace B: Settings → GitHub → Connect GitHub for the same GitHub org/account.
4. Workspace B now shows connected, but workspace A's GitHub integration has been silently disconnected (the `github_installation` row was reassigned to workspace B).

### Additional context (optional)

```shell
The relevant upsert:

-- server/pkg/db/queries/github.sql
[...]

Issue #2797, opened 2026-05-18
Default workspace slug collision breaks PR issue auto-linking


myfjdthink:
When two workspaces have names with the same first three letters, Multica generates the same default workspace slug / issue prefix for both workspaces. Since the slug and issue prefix are effectively the same identifier, this collision causes PR-to-Issue auto-linking to fail or become ambiguous.\n\nSteps to reproduce:\n1. Create two workspaces whose names start with the same first three letters.\n2. Keep the default generated slug / issue prefix for both workspaces.\n3. Create an issue in one of the workspaces.\n4. Open a PR with a title that starts with that issue ID, for example: ABC-12 fix something.\n\nExpected behavior:\nThe PR should be automatically linked to the issue in the correct workspace.\n\nActual behavior:\nBecause multiple workspaces can share the same generated slug / issue prefix, Multica cannot reliably identify the target issue, so the PR is not linked correctly.\n\nImpact:\nTeams with workspaces that share the same first three letters can lose automatic PR-to-Issue association, even when PR titles follow the required issue ID format.\n\nSuggested fix:\nDefault workspace slug / issue prefix generation should avoid collisions across workspaces, or the PR auto-linking logic should disambiguate by repository/workspace context instead of relying only on the prefix.

Bohan-J:
Thanks for the detailed report, @myfjdthink!

After digging in we found that PR auto-linking itself isn't actually misrouted across workspaces — a GitHub repository is bound to a single installation, which maps 1:1 to a workspace, so the PR webhook always resolves the correct workspace before matching prefix + number. The collision doesn't cross that boundary.

That said, you're right that the default 3-letter prefix generator can collide (two workspaces whose names start with the same three letters get the same prefix). The real friction was:

1. People who belong to multiple workspaces with the same prefix get visually confused (`ABC-12` is ambiguous to a human, even if the system isn't).
2. There was no way to fix it after the fact — the prefix was effectively immutable.

We decided **not** to enforce globally unique prefixes (different teams may legitimately want the same short identifier), and instead made the prefix **editable from Workspace Settings → General**:

- Admins/owners can change the prefix to any `[A-Z0-9]{1,10}` value.
- Saving a new prefix shows a confirmation dialog, since the change retroactively renames every existing issue (`OLD-N` → `NEW-N`) and will invalidate external references (PR titles, branch names, links in docs).
- The frontend cache is refreshed on prefix changes so the new identifier shows up immediately across list / detail / search.

[...]

PR #6808, opened 2026-08-12
MUL-6048 fix(github): withhold close intent when a PR identifier is ambiguous across workspaces

Bohan-J:
Fixes the auto-complete half of #6804.

## The bug

Issue prefixes are deliberately not globally unique (#2797), and issue numbers restart at 1 in every workspace. #5183 then made PR webhooks fan out to **every** workspace bound to an installation. Put together, `Closes ABC-100` on one PR can resolve to a genuine — but different — issue in two workspaces at once, and each one acted on it independently.

The concrete case: one GitHub org, two teams, two Multica workspaces, both connected to the same org installation (the flow #4823 exists to support). GitHub only ever creates one installation per org, so there is no way for the two teams to separate themselves on the GitHub side. Their prefixes collide because the default prefix is derived from the first three letters of the workspace name, and team names inside one company share the company name.

Result: a PR in team A's repo silently moved team B's unrelated `ABC-100` to `done`. The repo the PR lives in does not matter — `mirrorPullRequestForWorkspace` deliberately does not gate on the `workspace.repos` registry (MUL-4343).

## The change

Resolve the ambiguity **before** any workspace links: count how many bound workspaces resolve each closing identifier, and where more than one does, still write the link rows but withhold `close_intent`, which keeps the auto-advance gate shut.

- **Links stay.** Suppressing them would hide the collision, and being invisible is what made this bug hard to notice. The wrong link is visible on the issue and a human can unlink it — recoverable. Only the irreversible half is suppressed.
- **The right workspace also loses auto-complete for that identifier.** Deliberate: nothing in the event says which workspace owns the PR, so nobody gets to act. A repo-ownership tiebreaker could recover this later; it is a separate product decision and not in this PR.
- **A `warn` log names the colliding identifier**, workspace count, repo and PR number. An issue moving to `done` for no visible reason is otherwise untraceable back to GitHub, which is the actual reason this went unnoticed.
- **Withholding also heals stored state**: a re-fired webhook clears `close_intent` on a row a pre-fix delivery had already set.

## Cost

The pre-pass runs only when the PR carries a closing keyword **and** the installation has more than one binding. Single-workspace installations — the overwhelmingly common case — do no extra work and see zero behavior change.

## Verification

`server/internal/handler/github_test.go` gains `TestWebhook_PullRequest_AmbiguousCloseAcrossWorkspaces`: two workspaces on one installation, same prefix, both owning a real issue at the same number, PR titled `Fix AMB-1` (a closing keyword), opened then merged.

Confirmed the test reproduces the bug before it fixes it — with the suppression line disabled, it fails with **both** issues advanced to done:

```
github_test.go:3008: workspace B (owns the PR): close_intent must be withheld while the identifier is ambiguous
github_test.go:3016: workspace B (owns the PR): issue was auto-advanced to done on an ambiguous close (#6804)
github_test.go:3008: workspace A (collision): close_intent must be withheld while the identifier is ambiguous
github_test.go:3016: workspace A (collision): issue was auto-advanced to done on an ambiguous close (#6804)
```

With the fix, the full `server/internal/handler` suite passes locally against Postgres (`go test ./internal/handler/ -count=1` — ok, 28.7s), which includes the existing `TestWebhook_PullRequest_FansOutToBoundWorkspaces` and the merge-to-done tests.

## Not in this PR

#6804 has two more parts, both filed separately or still open for a call:

- Prefix defaults that make the collision near-certain (Chinese workspace names all derive to `WS`) — tracked separately.
- Warning an admin at connect time when an installation is already bound to a workspace with the same prefix.
- Whether PR mirroring should be scoped by repo ownership at all — a product decision on #5183's fan-out model.

### DIFF
--- a/server/internal/handler/github.go
+++ b/server/internal/handler/github.go
@@ -1283,11 +1283,12 @@ func (h *Handler) handlePullRequestEvent(ctx context.Context, body []byte) {
 	// Fanning out means an identifier can resolve in more than one workspace at
 	// once (issue prefixes are not globally unique and issue numbers restart at
 	// 1 per workspace, so two bound workspaces can both own a real "ABC-100").
-	// Resolve that ambiguity BEFORE any workspace links, so no workspace acts on
-	// a closing keyword we cannot attribute — see ambiguousCloseIdentifiers.
-	ambiguousCloses := h.ambiguousCloseIdentifiers(ctx, insts, &p)
+	// Settle who — if anyone — may act on a closing keyword BEFORE any workspace
+	// links, and treat that verdict as authoritative for the whole delivery, so
+	// the mirror pass cannot re-derive a different answer. See closeIntentPolicy.
+	closePolicy := h.resolveCloseIntentPolicy(ctx, insts, &p)
 	for _, inst := range insts {
-		h.mirrorPullRequestForWorkspace(ctx, inst.WorkspaceID, inst.InstallationID, &p, ambiguousCloses)
+		h.mirrorPullRequestForWorkspace(ctx, inst.WorkspaceID, inst.InstallationID, &p, closePolicy)
 	}
 	// The PR row(s) now carry the new head; ask the API pipeline for the
 	// authoritative CI + mergeability snapshot for that head. The webhook is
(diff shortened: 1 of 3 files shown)


Bohan-J:
Pushed `b06afd8` addressing the must-fix. Both halves of it were real.

**Fail open.** The scan reused helpers that fold read failures into permissive defaults, so a single transient failure could count two real resolvers as one and let both workspaces advance on merge — the exact outcome this PR exists to prevent.

**Not authoritative.** The scan produced a denylist and the mirror pass then re-derived close intent from its own reads, so anything changing in between reopened the hole.

The fix is structural rather than a patch on each read path: `closeIntentPolicy` is now an **allowlist** recording *which single workspace was proven to own* each closing identifier.

- Absence denies. The zero value — what every indeterminate read returns — permits nothing, so fail-closed is a property of the type rather than something each error branch has to remember.
- The mirror pass consults `policy.permits(identifier, workspaceID)` and can only narrow the verdict; it cannot grant close intent the scan withheld.
- Recording the *owner* rather than a bare "allowed" also closes the scan-to-write window you flagged: a workspace that grows a same-numbered issue in between is not the recorded owner, so it still cannot act.

Reads the scan depends on are now surfaced instead of swallowed: `autoLinkPRsEnabledForWorkspace` returns a parse error, and the scan distinguishes `pgx.ErrNoRows` (a proven miss) from an error it cannot interpret — the latter abandons close intent for the whole delivery. Existing callers keep the permissive default, so nothing outside this path changes behavior. Extracting `issuePrefixForWorkspace` / `issueNumberForPrefix` lets the scan derive everything from the row it already holds, which also removes a redundant workspace read per binding.

Link rows still get written in the withheld case, matching the PR's existing principle — the wrong link is visible and unlinkable, the silent status rewrite is not.

**Tests**, both of the ones you asked for plus a type-level one:

- `TestWebhook_PullRequest_UnreadableWorkspaceWithholdsCloseIntent` — a bound workspace whose settings blob parses in Postgres but not in the handler (`{"github_auto_link_prs_enabled": 5}`). Setup is otherwise identical to the auto-complete case, so it isolates exactly the fail-open path. Verified it fails against the old behavior:
  ```
  close_intent must be withheld when a bound workspace could not be inspected
  issue was auto-advanced despite an unreadable bound workspace — the scan failed open
  ```
- `TestWebhook_PullRequest_UniqueResolverAmongBindingsStillAutoCompletes` — multiple bindings sharing a prefix, one proven resolver, still advances to done. Guards against over-suppression.
- `TestCloseIntentPolicyPermits` — table test pinning that the zero value denies, that a non-owner is denied, and that an identifier absent from the allowlist is denied.

`go test ./internal/handler/ -count=1` passes locally against Postgres (72s).

PR #6809, opened 2026-08-12
fix(chat): preserve idle footer spacing (MUL-5854)

### LINKED ISSUE #6558: Chat: ChatListFooter's always-render comment contradicts its null guard — idle sessions lose the 16px bottom inset
## Summary

`ChatListFooter` in `packages/views/chat/components/chat-message-list.tsx` carries a doc comment promising the container always renders so the message list keeps a constant bottom inset — but the component returns `null` whenever no task is pending, so the promise is never honored and the idle gap degrades to the last row's own `py-2` (8px).

## Location

`packages/views/chat/components/chat-message-list.tsx`, `ChatListFooter` (registered as the Virtuoso `Footer` in `LIST_COMPONENTS`):

```tsx
// The container always renders (even with no pill) so the list keeps a
// constant bottom inset: without it the last row's own py-2 was the only gap
// between the final reply (and its follow-up pills) and the composer.
function ChatListFooter({ context }: { context?: ChatListContext }) {
  if (!context) return null;
  if (!context.showStatusPill || !context.pendingTask) return null;   // ← defeats the comment
  return (
    <div className={cn(CHAT_COLUMN, "pb-4 space-y-4")}>
      <TaskStatusPill ... />
    </div>
  );
}
```

The comment was introduced by commit `0fdc38704` (MUL-5149, #5766) — which states the always-render intent explicitly — yet the early `return null` above it means the container only mounts while a task is pending.

## Impact

- With a task pending, the gap between the last message and the composer is 16px (`pb-4` on the footer container).
- With no pending task (the common idle state), the footer unmounts and the gap drops to the last row's own `py-2` — 8px. The final reply sits closer to the composer than it does while anything is running.

## Suggested fix

Render the container unconditionally and gate only the pill inside it, i.e. honor the existing comment:

```tsx
function ChatListFooter({ context }: { context?: ChatListContext }) {
  return (
    <div className={cn(CHAT_COLUMN, "pb-4 space-y-4")}>
      {context?.showStatusPill && context.pendingTask ? (
        <TaskStatusPill ... />
      ) : null}
    </div>
  );
}
```

Reported from a downstream fork that shipped this exact one-guard fix; filing per the fork's policy of reporting bugs found outside its owned areas.

iiwish:
## What does this PR do?

Keeps the chat message list's intended 16px bottom inset when no task is pending.

`ChatListFooter` documented that its container should always render, but two early returns removed the entire footer in the common idle state. That left only the last message row's 8px padding between the final reply and the composer. The footer now always renders its spacing container and conditionally renders only the task status pill.

## Related Issue

Fixes #6558

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Updated `packages/views/chat/components/chat-message-list.tsx` so the footer spacing container remains mounted without a pending task.
- Added a regression test that renders an idle assistant reply and verifies the Footer still carries `pb-4` without rendering a status pill.

## How to Test

1. `pnpm --filter @multica/views exec vitest run chat/components/chat-message-list.test.tsx --pool=threads --maxWorkers=1 --no-file-parallelism --reporter=verbose`
2. `pnpm --filter @multica/views typecheck`

Both checks pass locally. The test run reports two pre-existing React `act(...)` warnings in unrelated quick-action cases.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (not applicable; behavior now matches the existing inline contract)

## AI Disclosure

**Prompt / approach:**
Audited open issues for a scoped, reproducible UI defect; traced the Virtuoso Footer path; changed only the conditional rendering boundary; and added a regression test before running the focused views test and typecheck.

## Screenshots (optional)

Not included because this is an 8px spacing regression covered directly at the Footer contract; no visual styling or layout structure otherwise changes.

### DIFF
--- a/packages/views/chat/components/chat-message-list.test.tsx
+++ b/packages/views/chat/components/chat-message-list.test.tsx
@@ -191,6 +191,35 @@ describe("ChatMessageList live timeline (MUL-3960 regression)", () => {
   });
 });
 
+describe("ChatMessageList footer spacing", () => {
+  it("keeps the bottom inset when no task is pending", () => {
+    const { container } = render(
+      <I18nProvider locale="en" resources={TEST_RESOURCES}>
+        <QueryClientProvider client={new QueryClient()}>
+          <ChatMessageList
+            messages={[
+              {
+                id: "assistant-idle",
+                chat_session_id: "session-idle",
+                role: "assistant",
+                content: "Idle reply",
+                task_id: null,
+                created_at: "2026-08-12T00:00:00Z",
+              },
+            ]}
+            pendingTask={null}
+            availability="online"
+          />
+        </QueryClientProvider>
+      </I18nProvider>,
+    );
+
+    const list = container.querySelector("[data-row-key]")?.parentElement;
+    expect(list?.lastElementChild).toHaveClass("pb-4");
+    expect(screen.queryByText(/working|queued/i)).not.toBeInTheDocument();
+  });
+});
+
 describe("ChatMessageList quick actions", () => {
   it("renders up to three suggestions and sends the hidden prompt", async () => {
     const qc = new QueryClient();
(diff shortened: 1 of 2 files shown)


multica-eve (approved):
Thanks for the focused fix and regression coverage. The change restores the intended idle-state spacing without affecting task-status rendering. Local verification and CI are green. Approved!

PR #6812, opened 2026-08-12
MUL-6057: fix(i18n): localize runtime activity heatmap

iiwish:
## What does this PR do?

Localizes the runtime activity heatmap end to end instead of leaving part of the chart in English for Chinese, Japanese, and Korean users.

The heatmap already translated its Less/More legend, but weekday rows, empty-cell tooltips, four insight labels, average/total copy, month names, and insight dates were hard-coded to English. This change routes all visible chart chrome through the existing `runtimes` namespace and formats dates with the active i18n locale.

## Related Issue

N/A - found during a focused runtime i18n audit.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added heatmap labels for English, Simplified Chinese, Japanese, and Korean with locale parity preserved.
- Replaced English weekday identifiers in the insight model with language-neutral weekday indices.
- Formatted month headings and busiest-day values with the selected locale.
- Added a Chinese-locale regression test covering labels, weekday/date formatting, totals, and empty-cell tooltips.

## How to Test

1. `pnpm --filter @multica/views exec vitest run runtimes/components/charts/activity-heatmap.test.tsx locales/parity.test.ts --pool=threads --maxWorkers=1 --no-file-parallelism --reporter=verbose`
2. `pnpm --filter @multica/views typecheck`
3. `pnpm --filter @multica/views exec eslint runtimes/components/charts/activity-heatmap.tsx runtimes/components/charts/activity-heatmap.test.tsx`

All checks pass locally. ESLint retains the component's pre-existing exhaustive-deps warning for the intentional custom-pricing store invalidation dependency; there are no lint errors.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (not applicable; locale resources are the canonical copy)
- [x] If this PR touches Chinese product copy, I checked it against the conventions

## AI Disclosure

**Prompt / approach:**
Audited user-visible runtime strings against the repository's i18n boundary, classified technical values versus translatable chrome, replaced hard-coded English with typed selectors and locale-aware date formatting, then added a non-English regression test plus locale parity coverage.

## Screenshots (optional)

Not included because the localized heatmap depends on workspace runtime usage data. The regression test renders the real component under `zh-Hans` and verifies the translated SVG and insight output directly.

### DIFF
--- a/packages/views/locales/en/runtimes.json
+++ b/packages/views/locales/en/runtimes.json
@@ -290,13 +290,6 @@
     "heatmap_average": "avg {{value}}",
     "heatmap_window_total_one": "{{count}}-day total",
     "heatmap_window_total_other": "{{count}}-day total",
-    "heatmap_weekday_mon": "Mon",
-    "heatmap_weekday_tue": "Tue",
-    "heatmap_weekday_wed": "Wed",
-    "heatmap_weekday_thu": "Thu",
-    "heatmap_weekday_fri": "Fri",
-    "heatmap_weekday_sat": "Sat",
-    "heatmap_weekday_sun": "Sun",
     "tooltip_total": "Total"
   },
   "list": {
--- a/packages/views/locales/en/runtimes.json
+++ b/packages/views/locales/en/runtimes.json
@@ -283,6 +283,20 @@
   "charts": {
     "heatmap_less": "Less",
     "heatmap_more": "More",
+    "heatmap_no_activity": "No activity",
+    "heatmap_busiest_day": "Busiest day",
+    "heatmap_most_active_weekday": "Most active weekday",
+    "heatmap_quietest_weekday": "Quietest weekday",
+    "heatmap_average": "avg {{value}}",
+    "heatmap_window_total_one": "{{count}}-day total",
+    "heatmap_window_total_other": "{{count}}-day total",
+    "heatmap_weekday_mon": "Mon",
+    "heatmap_weekday_tue": "Tue",
+    "heatmap_weekday_wed": "Wed",
+    "heatmap_weekday_thu": "Thu",
+    "heatmap_weekday_fri": "Fri",
+    "heatmap_weekday_sat": "Sat",
+    "heatmap_weekday_sun": "Sun",
     "tooltip_total": "Total"
   },
   "list": {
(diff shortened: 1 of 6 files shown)


Bohan-J (approved):
Thanks for this — the mixed-language heatmap was a real gap, and the audit that found it was the right kind of sweep. Two things in particular were the right call:

- Replacing the English weekday strings in the insight model with indices. Language does not belong in the data model, and that change is what made the rest of the fix clean.
- Formatting against `i18n.resolvedLanguage` rather than `navigator.language`. `issues/components/gantt-view.tsx` gets this wrong today (a user on a Chinese UI with an English browser still sees English dates there); this PR did not copy that pattern.

I have pushed three small follow-up commits to your branch rather than sending you round-trips:

**1. `refactor(i18n): derive heatmap weekday labels from Intl`**

`Intl.DateTimeFormat(locale, { weekday: "short" })` emits byte-identical strings to the 28 keys the PR added, for every locale we ship:

```
en      ["Mon","Tue","Wed","Thu","Fri","Sat","Sun"]
zh-Hans ["周一","周二","周三","周四","周五","周六","周日"]
ja      ["月","火","水","木","金","土","日"]
ko      ["월","화","수","목","금","토","일"]
```

Since the month labels in the same chart already go through `toLocaleString(locale, { month: "short" })`, keeping the weekdays in the bundle meant two date-formatting paths in one component and seven more strings to restate per new locale. I verified the rendered output in all four locales before and after — identical, apart from the wording change below.

**2. `fix(i18n): reword the Chinese heatmap weekday labels`**

`最活跃星期几` / `最不活跃星期几` read as a sentence fragment rather than a stat label; `conventions.zh.mdx` asks for 简洁直白 copy. Now `最活跃星期` / `最不活跃星期`. The ja and ko strings were already good.

**3. `test(runtimes): assert the heatmap weekday insight, not the row labels`**

`getAllByText("周三")` also matched the Mon/Wed/Fri row label, so it held regardless of what the weekday insight resolved to — including an English fallback. It now reads the value node under its own label, and the fixture has a second spend day so the busiest and quietest weekdays are distinct from each other and from the labelled rows. Also switched to the package's `renderWithI18n` helper so the component sees the full resource set instead of a hand-sliced namespace.

One thing I left alone: the cell tooltip keeps its ISO date while the insight date is localized. That is deliberate — it is the exact day key behind the cell and readers compare it against API/CLI usage rows — so I added a comment saying so rather than changing it.

Verified locally on the updated branch: `pnpm --filter @multica/views test` (323 files / 3811 tests), `typecheck`, and eslint on both changed files all pass, with only the component's pre-existing `exhaustive-deps` warning.

PR #6813, opened 2026-08-12
MUL-5738: fix(attachments): bound stalled text preview requests

### LINKED ISSUE #6400: [Bug]: HTML attachment preview stays in gray loading state while download works
### Deployment type

Official App (`https://api.multica.ai`) with Multica Desktop 0.4.18 on Windows.

### What happened?

Opening a self-contained HTML attachment inside an issue shows a gray preview area indefinitely. Downloading the same attachment and opening it in a browser works normally.

Expected: the HTML attachment should render in the inline preview, or the UI should show a bounded timeout and an actionable error.

Actual: the preview remains in the gray loading state indefinitely, without an error.

This appears related to, but is not the same reproduction as, #4477. That issue covered a self-hosted iframe/CSP failure. The current preview implementation fetches the document and renders it through `iframe.srcDoc`.

### Steps to reproduce

1. Use the Official App with Multica Desktop 0.4.18.
2. Attach a self-contained `.html` file to an issue (the reproduced file is about 247 KB and contains inline CSS, JavaScript, an inline SVG, and an embedded base64 font).
3. Open the attachment preview inside Multica.
4. Observe a gray preview placeholder that never resolves.
5. Download the same attachment and open it directly in a browser.
6. Observe that the document renders and works normally.

### Screenshots (optional)

A screenshot is available in the originating private workspace, but is not included here to avoid exposing private workspace URLs.

Local source inspection at commit `6b5d26b` and a controlled reproduction suggest the HTML document itself is valid:

- The exact HTML renders successfully with the same `iframe srcDoc` and `sandbox="allow-scripts"` setup used by `packages/views/editor/code-block-iframe.tsx`.
- `packages/views/editor/html-preview-body.tsx` renders the 480px gray placeholder while `query.isLoading` is true.
- The preview obtains the document through `GET /api/attachments/{id}/content` (`packages/core/api/client.ts`).
- The download path succeeds while the preview stays loading.
- The preview fetch currently has no visible timeout/fallback, so a `/content` request that never settles leaves the UI gray indefinitely.

Likely areas to investigate:

1. Hosted-server logs and object-storage reads for `GET /api/attachments/{id}/content`.
2. Why the content request does not settle while the download/CDN path succeeds.
3. Add a timeout/abort path and explicit error state to the HTML preview.

iiwish:
## What does this PR do?

Prevents text-backed attachment previews from remaining in the loading state forever when `/api/attachments/{id}/content` stalls.

The content request now has a 15-second end-to-end budget covering both response headers and body reads. On timeout it aborts the fetch, emits a typed preview error, and renders an explicit localized message with the existing Download recovery action. HTML, Markdown, and plain-text previews share this request path, so they receive the same bounded behavior.

This deliberately does not claim to fix the underlying object-storage stall described in the issue; it makes the client failure mode bounded and recoverable while that server-side cause is investigated.

## Related Issue

Fixes #6400

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added a 15-second abort budget to `getAttachmentTextContent`, including stalled response-body reads.
- Added `PreviewTimeoutError` and mapped it to a dedicated download-oriented fallback in inline HTML and full-screen text previews.
- Added timeout copy for English, Simplified Chinese, Japanese, and Korean with locale parity preserved.
- Added API and UI regression tests for abort behavior, inline HTML recovery, and full-screen Download fallback.

## How to Test

1. `pnpm --filter @multica/core exec vitest run api/client.test.ts --pool=threads --maxWorkers=1 --no-file-parallelism --reporter=verbose`
2. `pnpm --filter @multica/views exec vitest run editor/html-attachment-preview.test.tsx editor/attachment-preview-modal.test.tsx locales/parity.test.ts --pool=threads --maxWorkers=1 --no-file-parallelism --reporter=dot`
3. `pnpm --filter @multica/core typecheck`
4. `pnpm --filter @multica/views typecheck`
5. Run the focused core/views ESLint checks shown in the commit validation.

All checks pass locally: 70 core tests and 220 views/parity tests. The views run retains one pre-existing TanStack Query warning in the non-image body layout test; there are no test failures or lint errors.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (not applicable; the API comment and locale resources describe the contract)
- [x] If this PR touches Chinese product copy, I checked it against the conventions

## AI Disclosure

**Prompt / approach:**
Traced the issue from the loading placeholder through the shared TanStack Query hook and raw API client, verified there was no overlapping PR, added a bounded abort at the lowest shared request layer, preserved the Download escape path, and covered the API plus both inline/full-screen UI states.

## Screenshots (optional)

Not included because the changed state requires a deliberately stalled network response. The UI regression tests render both real timeout fallbacks and verify the localized message and Download action.

### DIFF
--- a/packages/core/api/client.test.ts
+++ b/packages/core/api/client.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import { ApiClient, ApiError, CHAT_DRAFT_RESTORE_CAPABILITY } from "./client";
 
 afterEach(() => {
+  vi.useRealTimers();
   vi.unstubAllGlobals();
 });
 
@@ -1269,6 +1270,28 @@ describe("ApiClient", () => {
       await expect(client.getAttachmentTextContent("att-1")).rejects.toBeInstanceOf(
         PreviewUnsupportedError,
       );
+    });
+
+    it("aborts a stalled preview and throws PreviewTimeoutError", async () => {
+      const { PreviewTimeoutError } = await import("./client");
+      vi.useFakeTimers();
+      const fetchMock = vi.fn((_input: RequestInfo | URL, init?: RequestInit) =>
+        new Promise<Response>((_resolve, reject) => {
+          init?.signal?.addEventListener("abort", () => {
+            reject(new DOMException("The operation was aborted", "AbortError"));
+          });
+        }),
+      );
+      vi.stubGlobal("fetch", fetchMock);
+
+      const client = new ApiClient("https://api.example.test");
+      const assertion = expect(
+        client.getAttachmentTextContent("att-stalled"),
+      ).rejects.toBeInstanceOf(PreviewTimeoutError);
+
+      await vi.runAllTimersAsync();
+      await assertion;
+      expect(fetchMock.mock.calls[0]?.[1]?.signal?.aborted).toBe(true);
     });
   });
 
--- a/packages/core/api/client.ts
+++ b/packages/core/api/client.ts
@@ -461,6 +461,18 @@ export class PreviewUnsupportedError extends Error {
   }
 }
 
+// Thrown when the content proxy does not return the complete preview body in
+// a bounded amount of time. Downloads use a separate path and remain
+// available as the recovery action.
+export class PreviewTimeoutError extends Error {
+  constructor() {
+    super("attachment inline preview timed out");
+    this.name = "PreviewTimeoutError";
+  }
+}
+
+const ATTACHMENT_PREVIEW_TIMEOUT_MS = 15_000;
+
 /**
  * Advertised in X-Client-Capabilities so the server knows this client can
  * recover a cancelled prompt from the durable draft-restore row (#5219).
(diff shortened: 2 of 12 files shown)


multica-eve (changes requested):
Thanks for addressing the infinite-loading failure mode. We are not going to merge this approach.

The underlying stall is still unlocated, while this change applies an unconditional 15-second end-to-end deadline to every HTML, Markdown, and plain-text preview. That can turn valid but slow reads into failures, masks whether the storage/read problem continues, and does not provide a consistent recovery path in the standalone HTML preview page. The regression test also covers a fetch that never returns a response, but not a response whose body stalls after headers arrive.

A replacement should first make the storage/read stages observable and identify where the request stalls, then bound that failure at the appropriate server-side boundary (preferably based on lack of progress rather than a blanket wall-clock limit). The client should retain cancellation, retry, and download recovery as a final fallback. The existing 2 MB text-preview cap should also be reconsidered as part of that design, with memory and rendering-cost measurements rather than only increasing the constant.

Closing this PR so the issue can be addressed with a root-cause-oriented design.

PR #6825, opened 2026-08-12
MUL-5741: fix(desktop): preserve HTML attachment tab scroll position across tab switches

### LINKED ISSUE #6405: [Bug]: HTML tab loses scroll position
### Deployment type

Official App

### What happened?

When I open a long HTML report in its own tab, scroll down to read it, and then switch to another tab, the HTML report is reset to the top when I return to it.

I expected the report to preserve its previous scroll position when switching between tabs, so that I could continue reading where I left off.

### Steps to reproduce

1. Open a long HTML report in Multica as an independent tab.
2. Scroll down several pages to a position away from the top.
3. Switch to another tab.
4. Switch back to the HTML report tab.
5. Observe that the report is displayed from the top instead of the previous reading position.

### Screenshots (optional)

Not provided. A screen recording can be added if needed.

This is especially disruptive for long agent-generated reports because they often cannot be read in one sitting and users frequently switch between tabs.

Environment:
- Multica version: `0.4.18`
- Deployment: Official App

This is related to, but distinct from, [issue #3310](https://github.com/multica-ai/multica/issues/3310). Issue #3310 concerns losing the scroll position of an issue detail page when navigating back, while this report concerns losing the scroll position of an HTML report when switching between tabs. This distinction may help avoid treating the two cases as the same bug.

The scroll position should be preserved independently for each HTML report tab. A newly opened report or a report whose content has changed should not inherit the position of a different report.

Frankli9986:
## Summary

Fixes #6405 — HTML tab loses scroll position when switching away and back in the desktop app.

Long HTML attachments opened in a dedicated tab reset to the top on every tab switch. The outer tab system already had a per-tab/route scroll memento, but it only scans `[data-tab-scroll-root]` containers in the React tree — the HTML preview scrolls inside a `sandbox="allow-scripts"` iframe, which is invisible to the coordinator and cannot be read cross-origin.

## Approach

Bridge the iframe's internal `scrollY` into the existing per-tab memento channel:

- `iframe-scroll-bridge.ts` — a sandbox-safe shim injected into the preview `srcDoc` that reports `scrollY` to the parent via `postMessage` and restores on request. No `allow-same-origin` added; no security boundary relaxed.
- `use-html-preview-scroll-restore.ts` — the `AttachmentPreviewPage` hook: registers the iframe as an external scroll source with the desktop coordinator, listens for the bridge messages, and pull-restores through the existing `useRestoredScrollOffset` path. A content fingerprint (`hash-string.ts`) stored alongside the memento entry invalidates restore when the attachment content changes, so a replaced report starts at the top instead of inheriting a stale position.
- `tab-coordinator.ts` / `tab-store.ts` — small, backward-compatible extensions: an external scroll-source registry and an optional `contentKey` on scroll memento entries.
- Scope is limited to the full-page HTML attachment preview; inline previews and the modal are unaffected, and the no-tab (web) path keeps its native behavior.

Rebased on current `main` with conflicts against the later-merged tab-restoration work (#6690) resolved; both features coexist (view-state entries + external scroll sources).

## Verification

- New/updated Vitest suites: bridge script runtime, restore hook behavior (message round-trip, fingerprint mismatch, no-provider degradation), coordinator capture merging, tab-store `contentKey` semantics, full-page component test.
- Real-Chromium Playwright spec (`e2e/iframe-scroll-bridge.spec.ts`) covers the parent/child handshake, including a regression for the ready→request-sync loop that was caught in review.
- `pnpm typecheck` and targeted `pnpm test` green (desktop 85, views 36); full `make check-worktree` (Go/DB segments) not run — no Go or web behavior is touched by this change.

### DIFF
--- a/apps/desktop/src/renderer/src/platform/tab-coordinator.test.ts
+++ b/apps/desktop/src/renderer/src/platform/tab-coordinator.test.ts
@@ -1,0 +1,160 @@
+import { beforeEach, describe, expect, it } from "vitest";
+import {
+  __resetTabCoordinatorForTests,
+  createScrollRestorationAdapter,
+  initTabCoordinator,
+  registerActiveHostElement,
+} from "./tab-coordinator";
+import { getActiveTab, useTabStore } from "@/stores/tab-store";
+
+describe("tab-coordinator external scroll sources", () => {
+  beforeEach(() => {
+    __resetTabCoordinatorForTests();
+    useTabStore.getState().reset();
+    useTabStore.getState().switchWorkspace("acme");
+    initTabCoordinator();
+  });
+
+  function seedActiveTab(path: string, title: string): string {
+    return useTabStore.getState().openTab(path, title, { activate: true });
+  }
+
+  it("captures external sources registered by the outgoing tab alongside DOM scroll roots", () => {
+    const tabId = seedActiveTab("/acme/attachments/abc/preview", "Report");
+    const adapter = createScrollRestorationAdapter(tabId);
+
+    // Set up an outer DOM scroll root to confirm both paths merge.
+    const host = document.createElement("div");
+    const list = document.createElement("div");
+    list.setAttribute("data-tab-scroll-root", "list");
+    Object.defineProperty(list, "scrollTop", { value: 250, configurable: true });
+    Object.defineProperty(list, "scrollHeight", { value: 1000, configurable: true });
+    host.appendChild(list);
+    registerActiveHostElement(host);
+
+    // Register an iframe-style external source with a contentKey.
+    adapter.registerExternalSource?.("html-iframe", {
+      capture: () => ({ top: 512, height: 4000, contentKey: "v1hash" }),
+    });
+
+    // Switch to another tab — the Coordinator's store subscription fires
+    // synchronously during the set() and captures the outgoing tab's state.
+    const otherId = useTabStore
+      .getState()
+      .openTab("/acme/issues", "Issues", { activate: true });
+    void otherId;
+
+    const outgoing = useTabStore
+      .getState()
+      .byWorkspace.acme.tabs.find((t) => t.id === tabId)!;
+    const scroll = outgoing.memento.scroll;
+    // DOM root was captured (top > 0 rule).
+    expect(scroll["/acme/attachments/abc/preview::list"]).toEqual({
+      top: 250,
+      height: 1000,
+    });
+    // External source was captured with its contentKey.
+    expect(scroll["/acme/attachments/abc/preview::html-iframe"]).toEqual({
+      top: 512,
+      height: 4000,
+      contentKey: "v1hash",
+    });
+
+    registerActiveHostElement(null);
+  });
+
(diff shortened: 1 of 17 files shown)


Bohan-J:
Thanks for this, and for sticking with it through the rebase and the handshake fix — merging now.

We reproduced the original bug first (long HTML report in a dedicated tab, switch away, switch back → back to the top), then verified this branch end to end:

- Trial merge into current `main`: clean. No overlap with anything that landed after your base commit.
- `pnpm typecheck`: all 6 workspace tasks pass.
- Targeted Vitest: views 36/36, desktop 85/85.
- `e2e/iframe-scroll-bridge.spec.ts`: 6/6 under the dedicated config (system Chrome) **and** 6/6 under the default config (bundled Chromium).

Things we specifically liked: the token is guarded by a `[a-z0-9]+` whitelist *and* `JSON.stringify` before it goes into the script source; `allow-same-origin` was not relaxed to make the parent's job easier; the parent validates `event.source`; and the `ready → request-sync` loop is genuinely bounded now (the at-most-one-restore-per-document rule), which the real-browser spec covers.

## One non-blocking follow-up

There is an edge case left, and we merged in spite of it because in that window the behavior is identical to today's `main` — so it is an uncovered case, not a regression:

If the tab is switched away while the attachment text is **still loading**, the previously saved offset gets erased. `contentKey = hashString(text ?? "")` treats the empty-string hash as a legitimate fingerprint, so `capture()` returns `{top: 0, contentKey: <hash of "">}`, and the per-route REPLACE in `commitScrollMemento` overwrites the stored `{top: 4000, …}`. When the content finally loads, the keys no longer match, so no `restore` is ever posted and the position is gone until the user scrolls again.

This is reachable because the memento is persisted through `partialize` while the React Query cache is not (`staleTime` 5 min / `gcTime` 30 min, empty on a fresh process) — so the natural trigger is: quit with a report open, relaunch, tab restores, report is still fetching, click another tab.

Worth noting if you pick this up: returning `undefined` from `capture()` while loading does **not** fix it — `captureScrollEntries` skips falsy entries and REPLACE then drops the key anyway. The stored entry has to be re-emitted as-is:

```ts
const hasDocumentRef = useRef(text !== undefined);
const savedEntryRef = useRef(savedEntry);
hasDocumentRef.current = text !== undefined;
savedEntryRef.current = savedEntry;

capture() {
  // No document to speak for yet — re-emit the stored entry.
  // Returning undefined would be dropped by REPLACE and lose it just the same.
  if (!hasDocumentRef.current) return savedEntryRef.current;
  ...
}
```

We tried that locally: the offset survives the loading window and all 36 existing views tests stay green. Happy to take it as a follow-up PR from you, or we can pick it up ourselves — no obligation either way.

Smaller nits, all optional: `hashString(text)` and `buildSrcDoc` run in full on every render (`useMemo` on `[text]` would cover it, and matters for multi-MB reports); `handleMessage` lists `sendRequestSync` in its deps but never calls it; and the in-iframe message handler could also check `e.source !== parent` for symmetry with the parent side.

Issue #3864, opened 2026-06-07
[Feature]: Android support for Multica Mobile


chyax98:
### Deployment type

self-host

### What do you want and why?

Multica Mobile currently relies on several iOS-only APIs (e.g. `ActionSheetIOS`, SF Symbols) and lacks Android configuration in `app.config.ts`. This means the app cannot be built or run on Android devices.

I would like to know if the upstream team has any plans or roadmap for Android support. I have done some preliminary work locally and can get the app to compile and install on an Android device with relatively small changes (~10 files, ~120 lines, zero new dependencies), but I want to align with the team's direction before proceeding further.

### Known blockers for community contributors

- **Google OAuth login**: The mobile login page (`apps/mobile/app/(auth)/login.tsx`) currently only supports email + OTP. Adding Google OAuth on Android requires the upstream team to register an Android OAuth client ID in Google Cloud Console (package name + SHA-1), which a community contributor cannot do unilaterally.
- **Testing coverage**: I have only verified that the app compiles, installs, and launches. Core features (issue list, chat, attachments, etc.) have not been deeply tested on Android yet.

### Proposed solution (optional)

If the team is open to Android support, I can prepare a focused PR that:
1. Adds Android configuration to `app.config.ts` (package name, adaptiveIcon)
2. Replaces iOS-only `ActionSheetIOS` with a cross-platform wrapper (iOS still uses native ActionSheetIOS, Android falls back to `Alert.alert`)
3. Conditionally renders icons: SF Symbols on iOS, Ionicons on Android via `Platform.OS`
4. Fixes Tab Bar SafeArea padding for Android gesture navigation
5. Pins Gradle version to 8.13 via `expo-build-properties` (compatible with Expo SDK 55 + RN 0.83)
6. Updates CI (`mobile-verify.yml`) to include an Android `assembleDebug` build check

All of the above preserves the existing iOS experience unchanged.

### Questions for the team

1. Is Android support on the official roadmap? If so, what is the expected timeline and priority?
2. Are there any design guidelines or technical decisions already made (e.g. icon system, navigation patterns) that a community PR should follow?
3. If there is no current plan, would you be open to a community-driven PR, or would you prefer to wait until the team has bandwidth to define the scope internally?

---

Happy to adjust the scope or approach based on the team's feedback.

DASungta:
[...]

Issue #3412, opened 2026-05-28
[Feature]: Add Android Build and Release


htynkn:
### Deployment type

Official App

### What do you want and why?

Hello team,

I noticed that the current Multica mobile app only provides an iOS version. Since the mobile app project’s tech stack is Expo + React Native, which is inherently cross‑platform, it should be possible to build and release an Android version as well.

My suggestions are:

+ Add Android build and release

Use standard React Native build process to generate Android APK / AAB files.
Maybe a beta version first

+ Clarify Android compatibility plan

Include information in the README or Release Notes about the Android release timeline and compatibility status.

### Proposed solution (optional)

Add android config to current project to support android then create a new workflow for android building

Bohan-J:
@htynkn thanks for the suggestion — and yes, the observation is correct: the mobile codebase is Expo + React Native, so an Android build is technically very feasible from the same source tree.

Android **will** land eventually. The reason it's not out yet is purely a bandwidth question: Multica is still a small team, and we have to be honest with ourselves about which platforms we can simultaneously support at a quality level we're comfortable shipping (handling crash reports, OS-version regressions, store review, push, etc.). For now we deliberately prioritized iOS first, with Android explicitly queued behind it.

So:

- **Plan:** Android is on the roadmap, not declined.
- **Timing:** Once the iOS app is stable enough that it's no longer pulling daily attention, we'll spin up the Android pipeline (build config, GitHub Actions workflow for AAB, Play Store track).
- **What you can do today:** if you're willing to dogfood, you can already build the Android target locally from the mobile package — the RN setup itself is largely cross-platform. Feedback from that would actually accelerate things since it surfaces Android-specific issues earlier.
- We'll update the mobile section of the README with this prioritization so the expectation is on the public record.

Leaving this issue open and treating it as the canonical tracking thread for the Android release.

Thanks again — the framing as "you already have most of the work" is a fair push.

htynkn:
@Bohan-J Thanks for the detailed reply! I really appreciate you explaining the
 team's current focus.  

I followed your suggestion, made a few small tweaks, and got a working APK
 built locally. It’s awesome to see the RN setup is already in such good
 shape.

Thanks again for keeping this issue open. I’ll definitely share any
 Android-specific bugs I find while testing locally.

DASungta:
I opened PR #3934 to add Android target support for the existing Expo mobile app: https://github.com/multica-ai/multica/pull/3934

Scope included in the PR:
[...]

PR #6826, opened 2026-08-12
fix(mobile): make inbox actions Android-safe (MUL-6062)

### LINKED ISSUE #6215: fix(mobile): Inbox actions menu crashes on Android
## Summary

Tapping the top-right ellipsis button on the Android Inbox tab crashes the app.

## Reproduction

1. Open the Android app.
2. Navigate to the Inbox tab.
3. Tap the top-right ellipsis / Inbox actions button.

## Actual behavior

The app crashes immediately.

## Expected behavior

Show Inbox batch actions:

- Mark all read
- Archive all read
- Archive completed
- Archive all

## Root cause

`apps/mobile/app/(app)/[workspace]/(tabs)/inbox.tsx` directly calls `ActionSheetIOS.showActionSheetWithOptions(...)`. `ActionSheetIOS` is iOS-only and has no Android implementation.

The mobile guidance already documents that Android requires a regular modal fallback (`apps/mobile/CLAUDE.md`).

## Scope

The same iOS-only API appears in other mobile menus, so Android-safe handling should be reviewed for those call sites as well.

mxbao063-bao:
## Summary

- keep the native Inbox action sheet on iOS
- use the existing cross-platform dropdown menu on Android and other platforms
- preserve all four batch actions and the destructive Archive all confirmation

## Testing

- pnpm --dir apps/mobile typecheck
- pnpm --dir apps/mobile lint (passes with 7 existing warnings outside this change)
- pnpm --dir apps/mobile test (62 tests passed)
- pnpm --dir apps/mobile exec expo export --platform android

Fixes #6215

### DIFF
--- a/apps/mobile/app/(app)/[workspace]/(tabs)/inbox.tsx
+++ b/apps/mobile/app/(app)/[workspace]/(tabs)/inbox.tsx
@@ -3,6 +3,7 @@ import {
   ActionSheetIOS,
   Alert,
   FlatList,
+  Platform,
   View,
 } from "react-native";
 import { useQuery } from "@tanstack/react-query";
@@ -15,6 +16,13 @@ import { Skeleton } from "@/components/ui/skeleton";
 import { Header } from "@/components/ui/header";
 import { IconButton } from "@/components/ui/icon-button";
 import { HeaderActions } from "@/components/ui/app-header-actions";
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
+} from "@/components/ui/dropdown-menu";
 import { SwipeableInboxRow } from "@/components/inbox/swipeable-inbox-row";
 import { inboxListOptions } from "@/data/queries/inbox";
 import {
@@ -34,6 +42,7 @@ export default function Inbox() {
   const wsId = useWorkspaceStore((s) => s.currentWorkspaceId);
   const wsSlug = useWorkspaceStore((s) => s.currentWorkspaceSlug);
   const { colorScheme } = useColorScheme();
+  const theme = THEME[colorScheme];
   const { data: rawItems, isLoading, error, refetch, isRefetching } = useQuery(
     inboxListOptions(wsId),
   );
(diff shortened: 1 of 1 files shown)


Bohan-J:
Thank you for this PR, and sorry for the wait. You put real work into it, so you deserve a straight answer rather than a silent close.

We're going to close this one — not because of the code, but because of scope.

**`apps/mobile/` is an iOS-only client today.** That isn't just a statement in the docs: there is no `android` section in `app.config.ts` (no package id, no adaptive icon), no Android scripts in `apps/mobile/package.json`, and no Android build or release step in CI. Official Android support is still an open feature request — [#3864](https://github.com/multica-ai/multica/issues/3864), with the build/release half in [#3412](https://github.com/multica-ai/multica/issues/3412). We're a small team and honestly just haven't had the capacity to build an official Android app yet. Sorry that it's taken this long to say so.

That's also the real story behind #6215: the JS bundle runs fine under Expo Go / a dev client on Android, so you can reach the Inbox screen and then hit `ActionSheetIOS`, which has no Android implementation. But the Inbox is one of seven call sites of that API:

| Call site | Entry point |
|---|---|
| `apps/mobile/app/(app)/[workspace]/(tabs)/inbox.tsx:86` | Inbox "…" (this PR) |
| `apps/mobile/app/(app)/[workspace]/issue/[id].tsx:126` | Issue detail "…" |
| `apps/mobile/app/(app)/[workspace]/project/[id].tsx:98` | Project detail "…" |
| `apps/mobile/app/(app)/[workspace]/more/settings/profile.tsx:67` | Avatar actions |
| `apps/mobile/components/issue/comment-context-menu.tsx:102`, `:219` | Comment long-press, emoji picker |
| `apps/mobile/components/chat/message-long-press.tsx:56` | Chat message long-press |

Merging this would stop the Inbox crash and leave the other six, so Android still wouldn't be usable — while the codebase would permanently carry a second menu implementation that has to be kept in sync with the iOS one.

For what it's worth, the code itself is good: iOS left completely untouched, `confirmArchiveAll` factored out so both paths share it, CI green. If and when we do take Android on, the shape we'd want is a single shared cross-platform action menu — something like `{ label, icon, destructive, onPress }[]` rendering `ActionSheetIOS` on iOS and the existing `DropdownMenu` elsewhere — applied to all seven call sites at once, alongside the Android config, build, and CI. That's a project rather than a patch, and it belongs under #3864.

So we're closing this PR and #6215 in favour of #3864, which is where Android will be tracked. Thanks again for taking the time on this, and please don't let it put you off sending other contributions.

PR #6830, opened 2026-08-12
MUL-5771 feat(daemon): allow configuring workspaces root

### LINKED ISSUE #6445: Feature Request: Configurable daemon workspace directory
## Problem

The Multica daemon currently hardcodes the workspace directory under `C:\Users\<username>\multica_workspaces\` (Windows) or `~/multica_workspaces/` (macOS/Linux). There is no way to customize this path.

## Use Cases

- Limited disk space on the C: drive / primary partition — need to place working directories on a different disk
- Desire for unified directory structures across multiple tools
- Multi-user / multi-profile scenarios requiring more flexible routing

## Expected

Provide a configuration option (e.g., `multica config set workspaces_dir D:\multica_workspaces`) or a daemon flag (e.g., `--workspaces-dir`) to let users customize the base workspace directory.

## Current Workaround

On Windows, the only option is using `mklink /J` to create a directory junction that redirects the path — but this is a workaround, not a real solution.

---

iiwish:
## What does this PR do?

Adds the accepted configuration surfaces from #6445 for relocating daemon task workspaces:

- `multica config set workspaces_root <path>` persists a root in the current profile
- `multica daemon start --workspaces-root <path>` and `daemon restart` provide a one-shot override
- background starts forward the flag to the foreground child
- resolution follows `--flag > MULTICA_WORKSPACES_ROOT > profile config > built-in default`

The implementation feeds the selected value into the existing `daemon.ResolveWorkspacesRoot` path rather than introducing a second path resolver, so absolute-path normalization and profile-aware defaults remain unchanged.

**Thinking path:** the issue is not just about moving data to another disk. The environment-only setting is process-context-dependent and profile-blind, while the daemon's other operator settings already use a profile-scoped config plus flag/env precedence. Extending that established contract is smaller and more predictable than adding workspace migration or a separate storage subsystem.

## Related Issue

Closes #6445

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `workspaces_root` to the profile CLI config schema, `config set`, and `config show`.
- Add `--workspaces-root` to daemon start/restart and preserve it across background re-exec.
- Cover config persistence, flag registration/forwarding, and the full four-level precedence chain.
- Update daemon/CLI/environment documentation in English, Chinese, Japanese, and Korean.

## How to Test

1. Run `go test ./cmd/multica ./internal/cli ./internal/daemon -count=1` from `server/`.
2. Run `go vet ./cmd/multica ./internal/cli ./internal/daemon`.
3. Run `pnpm --filter @multica/docs typecheck`.
4. Optionally set a profile-specific path with `multica --profile dev config set workspaces_root <path>`, start the daemon, and confirm `daemon disk-usage` reports that root.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots - N/A, CLI-only
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs - N/A

Risk: changing the configured root does not migrate task directories already present under the previous root. This matches the existing environment-variable behavior and is called out in the daemon documentation.

## AI Disclosure

**Prompt / approach:** Used Codex to inspect the accepted issue and maintainer guidance, trace the existing daemon configuration and background re-exec paths, implement the smallest change consistent with the repository's precedence conventions, add focused regression coverage, and run the listed verification commands. The resulting diff and tests were reviewed before publishing.

## Screenshots (optional)

N/A - no UI change.

### DIFF
--- a/apps/docs/content/docs/cli.ja.mdx
+++ b/apps/docs/content/docs/cli.ja.mdx
@@ -275,7 +275,7 @@ CLI の設定ファイルには、あなたとして Multica にアクセスで�
 
 | コマンド | サブコマンド | 用途 | 主なフラグ |
 | --- | --- | --- | --- |
-| `daemon` | `start` | ローカルデーモンを起動 | `--foreground`、`--device-name`、`--runtime-name`、`--poll-interval`、`--heartbeat-interval`、`--agent-timeout`（`0` は上限なし）、`--max-concurrent-tasks`、`--no-auto-update`；それぞれに対応する `MULTICA_*` 環境変数があります |
+| `daemon` | `start` | ローカルデーモンを起動 | `--foreground`、`--device-name`、`--runtime-name`、`--workspaces-root`、`--poll-interval`、`--heartbeat-interval`、`--agent-timeout`（`0` は上限なし）、`--max-concurrent-tasks`、`--no-auto-update`；それぞれに対応する `MULTICA_*` 環境変数があります |
 | | `stop` / `status` / `restart` | 停止、状態の確認、再起動（`restart` は `start` と同じフラグ） | |
 | | `logs` | デーモンのログを表示 | `--follow`、`--lines` |
 | | `disk-usage` | ローカルのディスク使用量を表示 | `--by-workspace`、`--by-task`、`--top` |
--- a/apps/docs/content/docs/cli.ko.mdx
+++ b/apps/docs/content/docs/cli.ko.mdx
@@ -275,7 +275,7 @@ CLI 설정 파일에는 사용자를 대신해 Multica에 접근할 수 있는 t
 
 | 명령 | 하위 명령 | 용도 | 주요 flag |
 | --- | --- | --- | --- |
-| `daemon` | `start` | 로컬 데몬 시작 | `--foreground`, `--device-name`, `--runtime-name`, `--poll-interval`, `--heartbeat-interval`, `--agent-timeout`(`0`은 상한 없음), `--max-concurrent-tasks`, `--no-auto-update`. 모두 대응하는 `MULTICA_*` 환경 변수가 있음 |
+| `daemon` | `start` | 로컬 데몬 시작 | `--foreground`, `--device-name`, `--runtime-name`, `--workspaces-root`, `--poll-interval`, `--heartbeat-interval`, `--agent-timeout`(`0`은 상한 없음), `--max-concurrent-tasks`, `--no-auto-update`. 모두 대응하는 `MULTICA_*` 환경 변수가 있음 |
 | | `stop` / `status` / `restart` | 중지, 상태 확인, 다시 시작(`restart`와 `start`의 flag가 같음) | |
 | | `logs` | 데몬 로그 확인 | `--follow`, `--lines` |
 | | `disk-usage` | 로컬 디스크 사용량 확인 | `--by-workspace`, `--by-task`, `--top` |
--- a/apps/docs/content/docs/cli.mdx
+++ b/apps/docs/content/docs/cli.mdx
@@ -296,7 +296,7 @@ The tables below cover every current top-level command, grouped the way the CLI 
 
 | Command | Subcommand | Purpose | Key flags |
 | --- | --- | --- | --- |
-| `daemon` | `start` | Start the local daemon | `--foreground`, `--device-name`, `--runtime-name`, `--poll-interval`, `--heartbeat-interval`, `--agent-timeout` (`0` means no limit), `--max-concurrent-tasks`, `--no-auto-update`; each has a matching `MULTICA_*` environment variable |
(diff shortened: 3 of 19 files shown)


Bohan-J (changes requested):
Thanks for the PR — the direction is right. Extending the existing profile-config + `flag > env > config > default` contract that the other daemon knobs already follow is exactly the shape this feature should take: no second resolver, profile isolation preserved, background re-exec forwarding covered by a regression test, and explicitly not auto-migrating existing task directories is the correct call. Docs in all four languages are appreciated.

One blocking issue before this can merge:

## Must-fix: `daemon disk-usage` never reads the new config value

The daemon start path threads `fileCfg.WorkspacesRoot` into `daemon.Overrides`, but the disk-usage paths still call `daemon.ResolveWorkspacesRoot` directly, which only considers `override > MULTICA_WORKSPACES_ROOT > built-in default` — the profile config is never loaded:

- `server/cmd/multica/cmd_daemon.go:1803` — `resolveDiskUsageRoot` (the normal `daemon disk-usage` path)
- `server/cmd/multica/cmd_daemon.go:1938` and `:1959` — `enumerateDiskUsageRoots` for `--all-profiles`; since it also `os.Stat`-skips roots that don't exist, a profile's actual custom root silently vanishes from the aggregate report
- `server/cmd/multica/cmd_daemon.go:2127` and `:2155` — `printDiskUsageOtherRootsHint`, which will suggest roots the daemon isn't actually using

Reproduced on this branch with a clean HOME and no env override:

```
$ multica config set workspaces_root /data/custom_root
Set workspaces_root = /data/custom_root
$ multica daemon disk-usage --output json | head -3
{
  "workspaces_root": "$HOME/multica_workspaces",   <-- default, not the configured root
```

So after `config set workspaces_root` + `daemon restart`, the daemon writes tasks to the new disk while `disk-usage` reports an empty/wrong tree. That breaks the verification story: step 4 of this PR's own "How to Test" ("confirm `daemon disk-usage` reports that root") does not hold as written, and disk-usage is the tool users will reach for to confirm the setting took effect.

Suggested fix: extract one shared "resolve workspaces root for profile" helper that loads that profile's CLI config and applies `flag > env > config > default`, and use it from daemon start, the normal disk-usage path, the `--all-profiles` enumeration, and the other-roots hint. Please add coverage for: disk-usage honoring the config value, env still overriding config, and `--all-profiles` including (and de-duplicating) per-profile custom roots.

## Nit (non-blocking)

`config set workspaces_root` accepts a relative path, which `ResolveWorkspacesRoot` later resolves against the invoking process's cwd — so the effective root can drift between invocations. Consider rejecting relative paths (or normalizing to absolute) at `config set` time. Same behavior as the env var today, so fine to defer.

Everything else looks good — precedence tests, flag forwarding, and the config plumbing all check out locally (`go build` / `go vet` / package tests pass on this branch).

iiwish:
@Bohan-J I pushed the requested disk-usage follow-up in `0d483c9b2` and rebased onto current `main`.

The shared profile resolver is now used by daemon start, normal disk usage, `--all-profiles`, and other-root hints. The update also covers config/env precedence, custom-root de-duplication, and absolute normalization for `config set workspaces_root`, with regression tests and the four localized docs updated.

Local verification passes:
- `go test ./cmd/multica ./internal/cli ./internal/daemon -count=1`
- `go vet ./cmd/multica ./internal/cli ./internal/daemon`
- Windows cross-build
- docs typecheck

The current backend CI failure is in the unrelated `server/internal/daemon/execenv` test `TestWorktreeModeDeliversBranchWithoutSidecars`, where `TempDir` cleanup found a non-empty `.git` directory. That package is untouched by this PR, and the exact test passes locally 20/20 times. As an external contributor I cannot rerun the upstream job. Could you please re-review the fix and rerun the failed backend check?

iiwish:
Rebased onto current `main` and reran the full backend validation against local pgvector/Postgres and Redis services using the same race-enabled wrapper as CI:

```
DATABASE_URL=... REDIS_TEST_URL=... bash scripts/test-go.sh --race
```

I also repeated the prior failing Linux `execenv` worktree test 100 times under `-race`, ran the affected Codex timeout tests 10 times under `-race`, plus `go build ./...`, focused `go vet`, the Windows cross-build, and docs typecheck. The fresh upstream run is now green: **backend passed in 5m31s and all 9 checks passed**.

Bohan-J (approved):
Thanks for the thorough follow-up, @iiwish — approving and merging.

The blocking issue from the previous round is fully closed: `daemon start`, plain `disk-usage`, `--all-profiles`, and the other-root hints now all go through the shared profile-aware resolver, so `flag > MULTICA_WORKSPACES_ROOT > profile config > default` resolves identically everywhere. Verified that after `config set workspaces_root <path>`, `daemon disk-usage --output json` reports the configured root instead of falling back to the default tree, and that cross-profile enumeration picks up custom roots and de-duplicates them. The absolute-path normalization on `config set` also addresses the earlier nit about relative paths drifting with the caller's cwd.

Particularly appreciated that this extends the existing `daemon.Overrides` / `ResolveWorkspacesRoot` contract instead of adding a second resolver, keeps default behavior unchanged, documents the no-auto-migration caveat, and ships regression coverage plus all four localized docs. Also thanks for chasing down the flaky `execenv` worktree test and re-running it under `-race` to confirm it was unrelated — CI is green on `ab87638`.

PR #6836, opened 2026-08-12
fix(daemon): keep runtimes online when version probes fail

### LINKED ISSUE #6452: [Bug]: daemon keeps a deleted pinned path when the re-resolved executable fails version detection (follow-up to #5683)
### Deployment type

Official App (standalone daemon, Homebrew install)

### What happened?

Follow-up to #5683, which was closed as fixed by #5355. The self-heal added there **does** re-resolve, and I can see it running — but it declines to heal in one case, and that case reproduces the original symptom exactly. The closing comment on #5683 asked for a re-report with daemon version, resolved path, logs and failing task IDs, so here they are.

**The gap:** the self-heal adopts a re-resolved path only if *version detection of that new path succeeds*. When the new path exists but its `--version` probe fails, the daemon keeps the previously pinned path — **without checking whether the pinned path still exists**. On a Homebrew install the previous path is a versioned Caskroom directory that `brew upgrade` has already deleted, so the daemon falls back to a path that cannot possibly work, and every task fails with the #5683 error.

Log line from the self-heal declining (repeats on every task attempt):

```
09:01:13.751 WRN re-resolved agent executable failed version detection; keeping pinned path
  component=daemon provider=claude command=claude
  new_path=/opt/homebrew/Caskroom/claude-code/2.1.220/claude
  error="detect version for /opt/homebrew/Caskroom/claude-code/2.1.220/claude: signal: killed"
```

…immediately followed by the task failing on the *old* path, which `brew` deleted hours earlier:

```
09:01:14.097 ERR task failed component=daemon task=e77131e0
  error="claude executable not found at \"/opt/homebrew/Caskroom/claude-code/2.1.212/claude\":
  exec: \"/opt/homebrew/Caskroom/claude-code/2.1.212/claude\": stat ...: no such file or directory"
```

Failing task IDs (5 tasks over ~10 hours, one per attempt): `e77131e0`, `63cb2f22`, `7b9c6e3a`, `55166010`, `fecfea52`.

**Second, worse behaviour on restart.** When the daemon restarted (routine version upgrade) it could not detect the version at startup either, and instead of registering the runtime with an unknown version it skipped registration entirely — taking the Claude runtime fully **offline**, so no task could even be dispatched:

```
19:03:42.178 WRN skip registering runtime component=daemon name=claude attempts=1
  error="detect version for /opt/homebrew/Caskroom/claude-code/2.1.220/claude: signal: killed"
```

This repeated every ~4 minutes until the executable was manually replaced.

### Why `--version` hung (may be worth handling explicitly)

This is a macOS Gatekeeper interaction that I think will keep biting the Homebrew install path, so it is probably worth designing around rather than treating as a one-off:

`brew upgrade --cask claude-code` writes the artifact with `com.apple.quarantine`. The claude cask artifact is a bare ~245 MB Mach-O, **not** an `.app` bundle, so Gatekeeper cannot service its first-launch consent. The first process to exec it here is the daemon's version probe — a headless process with no UI session — and it blocks in `_dyld_start` at 0% CPU **indefinitely**, producing no output and no error. macOS then caches that stuck verdict against the file's identity.

Verified locally: the file never returns (killed at 163s / 134s / 30s); `sample` shows `_dyld_start + 0`, 0% CPU, 32–80 KB RSS. Removing the quarantine xattr does **not** fix it, and neither does touching mtime — only replacing the file with a fresh inode (with quarantine cleared before first exec) recovers it. The repaired file has the **same sha256** as the wedged one, so the binary itself is fine; only its macOS security state differed.

So `signal: killed` here is the daemon's own probe timeout firing against a permanently-blocked exec, not a slow start or a corrupt download.

### Expected behaviour

1. **Never keep a pinned path that no longer exists.** If `stat(pinned)` fails, prefer the re-resolved path even when version detection of it failed — an executable of unknown version is strictly better than a path that provably cannot be executed. Today the daemon chooses the impossible one.
2. **Version detection failure should not equal "no runtime".** If the executable exists, register the runtime with an unknown/last-known version rather than skipping registration and going offline. Being unable to read a version number is not the same as being unable to run.
3. **Distinguish probe timeout from probe error.** A `--version` that never returns is a meaningfully different condition from one that exits non-zero, and (per this report) is a recurring macOS/Gatekeeper failure mode rather than a transient one. Surfacing it distinctly would make it diagnosable instead of showing up as a bare `signal: killed`.
4. Re-raising #5683's unaddressed secondary ask: **per-provider health/failure signal.** As in the original report, `daemon status` stayed healthy and the other providers kept succeeding for the entire outage, so nothing went red. Here it went unnoticed for ~10 hours for exactly that reason.

### Steps to reproduce

1. Install Claude Code via Homebrew cask (`brew install --cask claude-code`), so the artifact lives at `/opt/homebrew/Caskroom/claude-code/<version>/claude`.
2. Start the daemon and run a Claude task, so it pins the current versioned path.
3. `brew upgrade --cask claude-code` — this deletes the old version directory and writes the new artifact with `com.apple.quarantine`.
4. Ensure nothing execs the new binary from a UI session first (i.e. let the daemon be the first to touch it). Its `--version` will block indefinitely.
5. Submit any Claude task. The daemon logs `keeping pinned path` and the task fails with `claude executable not found` on the deleted path.

To reproduce step 4 deterministically without waiting for a real upgrade, wedging can be simulated by pointing the pinned path at any executable that never exits (the daemon's probe will time out identically); the interesting part of the bug is what it does *after* the probe fails.

### Environment

- multica 0.4.18 during the outage; 0.4.19 after the restart that took the runtime offline (both reproduce)
- macOS 26 (Darwin 25.5.0), arm64
- Claude Code 2.1.220 (Homebrew cask); daemon pinned 2.1.212, which the upgrade deleted
- Resolved Claude path: `/opt/homebrew/Caskroom/claude-code/2.1.220/claude`; stable entrypoint `/opt/homebrew/bin/claude` was correct and pointed at it throughout
- Providers on the same daemon: claude, codex, openclaw, hermes — only claude was affected

mxbao063-bao:
## Summary

- adopt a live re-resolved executable when its pinned predecessor was deleted, even if version probing fails
- keep live runtimes registered with their last-known or unknown version while preserving confirmed minimum-version rejection
- classify version probe timeouts explicitly and surface provider warnings through health and daemon status
- avoid repeating a self-heal probe after it already consumed the timeout budget

## Validation

- go test ./internal/daemon ./pkg/agent ./cmd/multica -count=1
- go vet ./internal/daemon ./pkg/agent ./cmd/multica
- go build ./...
- git diff --check

Fixes #6452

### DIFF
--- a/server/cmd/multica/cmd_daemon.go
+++ b/server/cmd/multica/cmd_daemon.go
@@ -1471,6 +1471,22 @@ func printDaemonStatusReport(w io.Writer, label string, health map[string]any) {
 	if reason, ok := health["reload_pending_reason"].(string); ok && reason != "" {
 		rows = append(rows, row{"Restart pending", reason})
 	}
+	if warnings, ok := health["agent_warnings"].(map[string]any); ok && len(warnings) > 0 {
+		providers := make([]string, 0, len(warnings))
+		for provider := range warnings {
+			providers = append(providers, provider)
+		}
+		sort.Strings(providers)
+		parts := make([]string, 0, len(providers))
+		for _, provider := range providers {
+			if reason, ok := warnings[provider].(string); ok && reason != "" {
+				parts = append(parts, provider+": "+reason)
+			}
+		}
+		if len(parts) > 0 {
+			rows = append(rows, row{"Agent warnings", strings.Join(parts, "; ")})
+		}
+	}
 	if agents, ok := health["agents"].([]any); ok && len(agents) > 0 {
 		parts := make([]string, len(agents))
 		for i, a := range agents {
--- a/server/cmd/multica/cmd_daemon_test.go
+++ b/server/cmd/multica/cmd_daemon_test.go
@@ -162,6 +162,28 @@ func TestPrintDaemonStatusExplainsDeferredRestart(t *testing.T) {
 	}
 }
 
+func TestPrintDaemonStatusShowsAgentWarnings(t *testing.T) {
+	t.Parallel()
+
+	health := map[string]any{
+		"status": "running",
+		"pid":    float64(1234),
+		"uptime": "1m",
+		"agent_warnings": map[string]any{
+			"codex":  "version detection failed: exit 1",
+			"claude": "version probe timed out after 10s",
+		},
+	}
+
+	var out bytes.Buffer
+	printDaemonStatusReport(&out, "Daemon", health)
+	got := out.String()
+	want := "claude: version probe timed out after 10s; codex: version detection failed: exit 1"
+	if !strings.Contains(got, want) {
+		t.Fatalf("daemon status output = %q, want sorted agent warnings %q", got, want)
+	}
+}
+
 // TestPrintDaemonStatusOmitsVersionWhenMissing pins the back-compat contract:
 // when the daemon doesn't report cli_version (older daemon paired with a newer
 // CLI) or reports an empty string, the CLI must skip the line entirely instead
--- a/server/internal/daemon/agent_path_selfheal_test.go
+++ b/server/internal/daemon/agent_path_selfheal_test.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"os"
 	"path/filepath"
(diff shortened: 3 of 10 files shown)


multica-eve:
Thank you for the contribution. We’re closing this PR because we can’t accept the proposed behavior change in its current form.

Keeping a runtime online whenever its executable exists but its version probe fails (including timeouts) makes the daemon advertise and dispatch a capability it cannot safely verify. In the reported failure mode, that can turn an immediate, diagnosable failure into a task that remains stuck for a long time; it can also bypass the minimum-version check.

The safer direction is to retain the existing registration and dispatch behavior, improve visibility for skipped/unavailable providers, and revisit any broader recovery behavior once we have a reproducible case or more telemetry. This PR also does not yet surface the existing `skipped_agents` state needed for that visibility.

We appreciate the investigation and will keep the underlying report tracked separately.

PR #6842, opened 2026-08-12
WS-3503: add private recruitment inbox processor

tangyuanjc:
Closes WS-3503

## Summary
- add a standalone Feishu long-connection processor allowlisted to one secret-configured private chat
- support text/post, image OCR, supported file analysis, and audio/voice transcription with transient media only
- add HMAC message deduplication, Feishu send UUID idempotency, dead-letter handling, hard-gate confirmation copy, and redacted logs
- persist only the approved minimal boolean/count summary schema; ship an opt-in Helm deployment, health probes, secret key names, and rollback guidance

## Verification
- go test ./internal/recruitinbox ./cmd/recruitment_inbox ./internal/integrations/lark ./pkg/llm
- go build ./cmd/recruitment_inbox ./cmd/server ./cmd/migrate
- migrations 285/286 applied to a local PostgreSQL database and schema/index inspected
- Helm template renders with recruitmentInbox.enabled=true
- Docker builder target successfully compiles and includes recruitment_inbox

## Deployment status
The service is disabled by default. This run had no cloud target or production secret-manager values, so no production deployment or live Feishu acceptance flow was executed, and no Feishu messages were sent.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -388,6 +388,28 @@ MULTICA_LARK_CALLBACK_BASE_URL=
 # environment handling.
 MULTICA_LARK_WS_PROXY_URL=
 
+# Private recruitment inbox processor (optional, separate service process).
+# Run `./recruitment_inbox`; the general Multica server does not consume these
+# variables. Put secret values in the deployment secret manager, not this file.
+RECRUITMENT_INBOX_FEISHU_APP_ID=
+RECRUITMENT_INBOX_FEISHU_APP_SECRET=
+RECRUITMENT_INBOX_ALLOWED_CHAT_ID=
+RECRUITMENT_INBOX_BOT_OPEN_ID=
+# Base64-encoded random 32+ byte HMAC key for opaque ledger/idempotency IDs.
+RECRUITMENT_INBOX_HASH_KEY=
+RECRUITMENT_INBOX_OPENAI_API_KEY=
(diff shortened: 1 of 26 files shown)


Bohan-J:
Thanks for the thorough work here — the implementation itself is careful (dedup, idempotency, dead-lettering, redacted logs, opt-in deployment). We're declining this one on product scope rather than code quality, so I'm closing it.

Multica is a general-purpose platform, and this change embeds one organization's private recruitment operation into it: a dedicated `cmd/recruitment_inbox` binary allowlisted to a single secret-configured Feishu chat, its own Helm deployment and secret key names, and two schema migrations (285/286) for recruitment event records. Nothing in it is configurable or reusable by other Multica users — no workspace, agent, or channel gains a capability — yet every user carries the migrations, the build target, and the maintenance and sensitive-data surface. That's a trade we don't want to make in this repo.

Two directions that would work better:

- Keep the recruitment logic in your own service or fork. It already runs as a separate process with its own secrets, so it doesn't need to live here to be deployed alongside Multica.
- If part of this is genuinely general — e.g. a Feishu long-connection channel, or a configurable per-chat message processor that any workspace could point at an agent — please open an issue proposing that extension point on its own, with no recruitment-specific schema or allowlist. We're happy to discuss that shape, and a contribution to the `lark` integration or the LLM client along those lines would be welcome.

Appreciate the effort, and sorry for the outcome after this much work.

PR #6850, opened 2026-08-12
MUL-6080: perf(skills): compress builtin skill files by ~47% tokens

maxivillus:
## Summary

Compress the 9 builtin skill `SKILL.md` files and their 8 `references/*-source-map.md` support files. Semantics, contracts, triggers, and exact `file:line` citations are preserved; only prose density and repetition were reduced. Builtin skills are embedded into the server binary and written into every task workdir — agents read them on demand, so slimmer files mean cheaper reads and less disk per task.

**Tokens (tiktoken, o200k_base): 42 241 → 22 530 (x1.87, −46.7%)** — per-file table in `server/internal/service/builtin_skills/COMPRESSION.md` (committed; happy to drop it before merge if unwanted).

## How it was validated

- Every file: `skill_compressor.py` contract — literals (commands, paths, error messages, security rules) preserved verbatim, word-ratio ≥ 2.0, independent semantic review per file.
- Conformance tests pass on this tree:
  - `TestBuiltinSkillsConformToTemplate`
  - `TestBuiltinSkillsFrontmatterIsStrictYAML`
  - `TestCreatingAgentsSkillCoversAgentCreationContracts`
  - `TestSquadsSkillCoversLeaderRoutingContract`
  - `TestAutopilotsSkillCoversDispatchAndSideEffects`
  - `TestRuntimesAndReposSkillCoversClaimAndCheckoutChain`
  - `TestProjectsAndResourcesSkillCoversDurableContext`
  - `TestSkillImportingSkillCoversWorkspaceImportContracts`
  - `TestWorkingOnIssuesSkillCoversIssueLoopContracts`
- No curated-key-vocabulary ("High-signal keys") or unbounded comment reads crept back (MUL-5442 / MUL-5696 anchors).

## Repro

```bash
cd server && go test ./internal/service -run \
  'TestBuiltinSkillsConformToTemplate|TestBuiltinSkillsFrontmatterIsStrictYAML|Test.*SkillCovers|Test.*Skill\(' -count=1
```

Base: `13706d6` (main).


Bohan-J:
Thanks for putting real effort into this, @maxivillus — the goal (cutting the context cost of the builtin skill files) is one we want, and the write-up made the change easy to audit.

We're going to decline this PR, for a policy reason first and a technical one second.

**Policy.** These files are a runtime behavior contract: every agent run on the platform reads them, so any wording change propagates to every task, in every workspace, immediately. We've decided to keep edits to this directory in-house rather than take them from outside the team — this isn't a judgment on your work, it's about who owns the blast radius. We won't be accepting external PRs for `server/internal/service/builtin_skills/` going forward.

**Technical.** For transparency, our review also found that the compressed output doesn't hold the "no semantic change" guarantee the PR claims. A few representative examples:

- `multica-working-on-issues`: "`pr_url` metadata (which can be **stale**)" became "which can be **large**" — the conformance test only anchors on the literal prefix, so it still passes while the meaning is inverted.
- `multica-skill-importing`: the 16 MiB limit is attached to `--url` imports, but `maxImportArchiveUploadSize` only applies to `--file` archive uploads.
- `multica-runtimes-and-repos`: the `--ref` precedence is reversed (explicit `--ref` wins; `resource_ref.ref` is the fallback), and the `--cascade` data-retention guarantee — agents and their full history are preserved, only the binding is cleared — was dropped.
- Several safety notes were removed, e.g. that `--custom-env <json>` leaks secrets into shell history and `ps`.

The broader lesson on our side: a fixed compression ratio is the wrong target for files whose value is precision, and literal-anchor tests can't be used as evidence that meaning was preserved. We'll take that into account when we do this work ourselves.

If you'd like to contribute elsewhere, we'd genuinely welcome it — issues labeled `good first issue` are a good place to start, and we're happy to point you at something. Sorry for the outcome here, and thanks again for the time you put in.

PR #6870, opened 2026-08-13
MUL-6104: fix(daemon): tell the brief's reader not to merge stderr into parsed output

seacen:
Both briefs teach `--output json`. Neither said what is on the other stream, so an agent learns the flag without the constraint that makes it work.

## What happened

An agent posted a comment, read its own success as a failure, and posted it again — the same text twice, on the issue of the person who asked for it. The command it ran:

```
multica issue comment add <id> --content-file ./report.md --output json 2>&1 \
  | python3 -c "import sys,json;d=json.load(sys.stdin);print('COMMENT OK id=',d.get('id'))"
```

The comment was created. `cmd_issue.go` then wrote `Comment added to issue UNI-13.` to **stderr**, which `2>&1` had just merged into stdout, so `json.load` hit prose on line one and the command exited non-zero. Eight seconds later the agent ran the same `comment add` again.

The CLI is not at fault here — the streams are separated correctly and `--output json | jq` has always worked. It stays correct only while the caller keeps them apart.

## Why guidance rather than a code change

I wrote the CLI fix first — suppress the confirmation under `--output json` — and then abandoned it, because it cannot keep the promise it implies. `runIssueCreate` legitimately writes `warning: upload attachment ... failed` and `Skipping --attachment ...` to stderr, and those have to stay. A rule that holds most of the time is worse than no rule: it teaches that `2>&1` is safe, then breaks on exactly the runs where something else also went wrong.

One sentence in the brief covers every command instead of the twelve that print a confirmation, and it covers tools that are not ours.

## What was measured, and what was not

Twelve throwaway agent sessions, same task ("post this comment, then confirm from the output that it was created"), against a stub CLI that reproduces the real stream behaviour:

| | without the line | with it |
|---|---|---|
| used `2>&1` | 6/6 | 0/6 |
| posted twice | 0/6 | 0/6 |

The guidance removes the habit reliably, and one session redirected stderr to a file rather than dropping it — the adaptation rather than the letter.

**It does not establish how often a merged stream becomes a duplicate write.** Neither arm re-posted: on a three-step task the agent investigated and reported honestly instead. The one duplicate on record is the production run above, a six-minute session where `2>&1` appeared in thirty-odd commands. So the claim here is that this removes the precondition, not that it prevents a measurable number of duplicates.

## The test

`TestEveryBriefThatTeachesJSONOutputAlsoWarnsAgainstMergingStderr` holds both builders to three things: the flag is taught, `2>&1` is named, and the cost is stated. `writeAvailableCommandsQuickCreate` has its own copy of the `--output json` sentence, and a quick-create run has the least context to work the rest out for itself — so guidance added to the full brief alone would be missing exactly where it is needed most.

Reverse-verified per builder: removing the line from either one fails on that builder alone.

## Verification

`go build ./...`, `go vet ./internal/daemon/...`, `go test ./internal/daemon/execenv/` and `go test -race ./internal/daemon/...` clean. `TestBriefByteIdenticalAcrossRunsForEveryKind` still passes — the addition is deployment-invariant, so it does not split the prompt cache prefix.

`TestProbeAgentCLIs_QoderResolvesViaLoginShell` fails on this machine, and fails identically on an unmodified checkout of `main`: a real `qoderclicn` on the local PATH defeats the test's fixture. Not from this change.


Bohan-J (changes requested):
Thanks — the guardrail is worth having. The failure it targets (a write that succeeded being misread as failed and retried) has irreversible, user-visible consequences, and the cost of the rule is small. Direction accepted; two changes are needed before merge.

**1. Drop the `2>/tmp/err` example (`runtime_config_sections.go`, both brief builders)**

The suggested redirect contradicts the brief's own rule a few lines below it ("never `/tmp` or shared paths", MUL-4252), isn't portable to native Windows (this sentence renders into Windows briefs unconditionally), and a fixed shared filename can collide with or truncate another concurrent run's output. A safety rule that violates itself teaches the wrong lesson. The final clause already carries the correct advice — end the sentence at "…leave it alone". If you want to keep teaching capture, the example must be unique, workdir-local, and cross-platform.

**2. Anchor the test on the rule's semantics, not bare substrings (`runtime_config_test.go`)**

I mutation-tested `TestEveryBriefThatTeachesJSONOutputAlsoWarnsAgainstMergingStderr` on this branch:

- Inverting the rule to "**Always** merge them (`2>&1`)" — the exact behavior the incident came from — still **passes**, because the assertions only check that `2>&1` and `look like it failed` appear somewhere in the brief. The *WarnsAgainst* in the test name is the one thing it doesn't verify.
- Deleting the quick-create copy of the sentence correctly **fails**, so presence-guarding works.

Please anchor at least on the full phrase ``Do not merge them (`2>&1`)`` so a semantic flip is caught, and consider a per-brief count guard so the rule can't silently duplicate as briefs evolve.

Minor: the "six of six sessions" figure and the 2026-08-12 incident exist only in the test comment — fine as motivation, but nothing asserts them; consider trimming the comment to what the test actually enforces.

With those two addressed, this looks good to merge.

seacen:
@Bohan-J — head is `9ea8baf28`. Both asks done, and your mutation test found a third thing one clause to the left of them.

## 1. The `2>/tmp/err` example is gone

Both builders. The sentence ends at "Redirect stderr or leave it alone." No replacement example — anything I invented would have to be unique, workdir-local and cross-platform, and none of that belongs in a sentence about stream merging.

You were right that the contradiction is inside the same function: `runtime_config_sections.go:259` suggested `/tmp` and `:263`, four lines down, says "never `/tmp` or shared paths". The quick-create builder collides at `:292` vs `:294`, which carries the MUL-4252 citation verbatim. And the Windows point holds — `writeCommentFormatting` branches on `runtimeGOOS`, `writeAvailableCommands` does not, so that line rendered into Windows briefs unconditionally.

## 2. The test is anchored on the rule, and now on its premise too

`wantRule` is the full phrase ``Do not merge them (`2>&1`)``, `wantWhy` is `"a write that SUCCEEDED look like it failed"`, and `strings.Contains` became `strings.Count` with 0 and >1 as distinct failures — the count guard you suggested.

**Then I ran your mutation the other way and it survived.** Flipping the sentence's own premise to "writes JSON to **stderr**; confirmations and warnings go to **stdout**" leaves the rule and the consequence untouched, and the test passed. Nothing anywhere in the repo pinned that mapping. The rule is only correct while the streams carry what the premise says they carry, so `wantPremise` is anchored as well.

Mutation results, each with the other two clauses untouched:

| mutation | result |
|---|---|
| premise reversed | FAIL, both builders |
| rule reversed (your case) | FAIL, both builders |
| consequence reversed | FAIL, both builders |
| rule deleted, either builder | FAIL, that builder |
| rule duplicated, either builder | FAIL, that builder |

## 3. Comment trimmed

The "six of six sessions" figure and the 2026-08-12 incident are out — you are right that nothing asserted them. What is left is the causal chain the assertions do enforce, plus a line on why the anchors are full phrases rather than operators.

## Known gap, disclosed rather than papered over

Re-adding `2>/tmp/err` — or any shared path — to that sentence is still green. I did not bolt a `/tmp` check onto this test because its name is about the merge-stderr rule, and stuffing an unrelated workdir-path assertion into it recreates the exact defect you just caught: a test name that does not describe what it checks. If you want that guarded, it wants its own test over the rendered briefs, and I would rather you say so than guess.

`gofmt`, `go vet`, `go build ./...`, and `go test -race ./internal/daemon/execenv/` all clean. CI 9/9.

server/internal/daemon/execenv/runtime_config_sections.go
```diff
@@ -256,6 +256,7 @@ func sanitizeBriefCodeToken(s string) string {
 func writeAvailableCommands(b *strings.Builder, ctx TaskContextForEnv) {
 	b.WriteString("## Available Commands\n\n")
 	b.WriteString("Prefer `--output json` for structured data. The default brief lists only the core agent loop and common issue create/update tasks; for everything else run `multica --help` or `multica <command> --help`.\n\n")
+	b.WriteString("`--output json` writes JSON to stdout; confirmations and warnings go to stderr. Do not merge them (`2>&1`) into anything that parses the output — a confirmation line inside a JSON stream makes a write that SUCCEEDED look like it failed, and retrying a write that already happened is how one comment gets posted twice or one file gets sent into a chat twice. Redirect stderr or leave it alone.\n\n")
```
Bohan-J: Compress the incident narrative to the minimal semantic boundary (premise + prohibition + consequence); the worked examples are derivable and already live in the PR body and the test comment, where they belong.

```suggestion
	b.WriteString("`--output json` writes JSON to stdout; confirmations and warnings go to stderr. Do not merge them (`2>&1`) into anything that parses the output — that makes a write that SUCCEEDED look like it failed and invites a duplicate retry.\n\n")
```

Bohan-J (commented):
One more pass, wearing the prompt-budget hat this time: before merge, the brief line should carry only the minimal semantic boundary — the premise (which stream carries what), the prohibition, and the consequence. The examples of what a duplicate looks like are derivable from "retried" and already live in the PR body and test comment.

The wording in the two suggestions below is chosen so that all three anchors of `TestEveryBriefThatTeachesJSONOutputAlsoWarnsAgainstMergingStderr` (premise / rule / consequence, and the exactly-once count) hold verbatim — **accepting both suggestions requires no test change**. Verified locally at 9ea8baf28 with both applied: `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/execenv/ -count=1` passes, including the byte-identical cache-prefix guard. Rendered size per brief goes 396 → 234 bytes (−41%).

Accept both and I'll approve.

server/internal/daemon/execenv/runtime_config_sections.go
```diff
@@ -288,6 +289,7 @@ func writeAvailableCommands(b *strings.Builder, ctx TaskContextForEnv) {
 func writeAvailableCommandsQuickCreate(b *strings.Builder) {
 	b.WriteString("## Available Commands\n\n")
 	b.WriteString("**Use `--output json` for structured data.** For anything beyond `issue create`, run `multica --help` or `multica <command> --help`.\n\n")
+	b.WriteString("`--output json` writes JSON to stdout; confirmations and warnings go to stderr. Do not merge them (`2>&1`) into anything that parses the output — a confirmation line inside a JSON stream makes a write that SUCCEEDED look like it failed, and retrying a write that already happened is how one comment gets posted twice or one file gets sent into a chat twice. Redirect stderr or leave it alone.\n\n")
```
Bohan-J: Compress the incident narrative to the minimal semantic boundary (premise + prohibition + consequence); the worked examples are derivable and already live in the PR body and the test comment, where they belong.

```suggestion
	b.WriteString("`--output json` writes JSON to stdout; confirmations and warnings go to stderr. Do not merge them (`2>&1`) into anything that parses the output — that makes a write that SUCCEEDED look like it failed and invites a duplicate retry.\n\n")
```

Bohan-J (approved):
Thanks for sticking with this through three rounds of review, @seacen — and for taking the compression seriously rather than arguing for the longer text. Approving.

Verified at `8bc93aa13`:

- Both builders carry the compressed rule verbatim — `writeAvailableCommands` and `writeAvailableCommandsQuickCreate`. **396 → 234 bytes per rendered brief (−41%)**, and since a run renders exactly one of the two, that's the real prompt cost.
- The three verbatim anchors the test pins — the premise, `Do not merge them (`2>&1`)`, and `a write that SUCCEEDED look like it failed` — all still hold, so the test file needed no changes in this round. That was the point of keeping the suggestion's wording close to yours instead of rewriting it.
- Final diff is `runtime_config_sections.go` +3/−1 and `runtime_config_test.go` +60. Nothing else touched.
- CI green, 9/9.

Your earlier call to drop the CLI-side fix was the right one, for the right reason: suppressing the confirmation under `--output json` would have made `2>&1` *look* safe and then failed on exactly the runs that also had a real warning to report. A guarantee that holds most of the time is worse than none. Good instinct.

Merging now.

PR #6871, opened 2026-08-13
test(handler): stop the dashboard fixture finishing before its own window opens (MUL-6103)

seacen:
`TestDashboardEndpoints` and `TestDashboardFailuresCountNeverStartedTasks` fail on any run whose handler tests execute in the first twenty minutes after midnight UTC.

## Mechanism

The fixture builds a run that **ends** at `now-20m`:

```go
started   := now.Add(-30 * time.Minute)
completed := started.Add(10 * time.Minute) // 600s run
```

and the endpoints are then asked for `days=1`. That window begins at the viewer's **start of day** (`@since` / `parseExactSinceParamInTZ`), and `ListDashboardAgentRunTime` filters on the **end** of the run:

```sql
AND atq.completed_at >= @since::timestamptz
```

So between 00:00 and 00:20 the run finished yesterday, the window misses it, and the assertions read that as nothing having happened:

```
dashboard_test.go:213: agent-runtime: expected >=1 task for agent, got 0
dashboard_test.go:216: agent-runtime: expected >=600s (one 10-minute run), got 0
dashboard_test.go:1610: expected the classified failure to be counted, got map[]
```

## Observed both ways

On a branch touching only `internal/daemon/execenv`:

| handler tests ran at | fixture's `completed_at` | in today's window | result |
|---|---|---|---|
| 00:13 UTC | 08-12 23:53 | no | **failed on those lines**, other 8 jobs green |
| 00:25 UTC | 08-13 00:05 | yes | **passed**, nothing changed but the clock |

Neighbouring PRs whose backend jobs landed at 22:20 and 13:40 were green.

## Fix

Clamp the fixture's start to start-of-day, which moves its end with it. That keeps what the test means — a ten-minute run that finished *today* — at every hour:

| suite runs at | old `completed_at` | in window | new `completed_at` | in window |
|---|---|---|---|---|
| 00:01 | 08-12 23:41 | no | 08-13 00:10 | yes |
| 00:13 | 08-12 23:53 | no | 08-13 00:10 | yes |
| 00:19 | 08-12 23:59 | no | 08-13 00:10 | yes |
| 00:21 | 08-13 00:01 | yes | 08-13 00:10 | yes |
| 12:00 | 08-13 11:40 | yes | 08-13 11:40 | yes |

**Correction to the original description:** it claimed both tests share one fixture. They do not — `TestDashboardFailuresCountNeverStartedTasks` builds its own timestamps around line 1561, and the first version of this patch left them untouched. They share one now: both call `runFinishedToday`, which anchors the run to the start of the day in the fixture's timezone rather than to whatever the clock happens to read.

Test-only, 3 lines plus a comment. `gofmt`, `go vet ./internal/handler/` clean.

### DIFF
--- a/server/internal/handler/dashboard_test.go
+++ b/server/internal/handler/dashboard_test.go
@@ -17,8 +17,17 @@ import (
 // the user's stored user.timezone, else UTC), so a fixture anchored in one
 // zone and a window opened in another sit an offset apart. Without the
 // `?tz=` these requests would inherit whatever user.timezone holds — NULL in
-// the handler fixture, hence UTC, but nothing here should rest on that.
-const dashboardFixtureTZ = "UTC"
+// the handler fixture, hence UTC.
+//
+// The zone is deliberately EAST of UTC, and that is what makes the pin
+// load-bearing rather than decorative. A fixture built in UTC survives losing
+// its `?tz=`: the fallback is UTC too, so the window opens where the fixture
+// already sits and nothing notices. Built in Asia/Tokyo the run finishes at
+// 15:10 UTC the previous day, so a request that falls back to UTC opens its
+// window after the run ended and the assertions go red — which is the whole
+// point of pinning it. Verified by mutation: dropping the parameter, and
+// pinning it to UTC, both fail.
+const dashboardFixtureTZ = "Asia/Tokyo"
 
 // dashboardFixtureTZParam is dashboardFixtureTZ as a query fragment, so a
 // request cannot pin a zone the fixture was not built in.
--- a/server/internal/handler/dashboard_test.go
+++ b/server/internal/handler/dashboard_test.go
@@ -70,6 +70,23 @@ func runFinishedToday(now time.Time, loc *time.Location) (started, completed tim
 	return started, started.Add(10 * time.Minute)
 }
 
+// pinDayWindowClock freezes the clock the request-side cutoff reads and hands
+// back the same instant for the fixture, so both sides describe one moment.
+//
+// They read the wall clock at two different points otherwise — the fixture when
+// it is built, the cutoff when the handler runs, with inserts and a rollup in
+// between — and a suite that crosses midnight in that gap writes its run into
+// one day and then asks for the next day's window. The gap is milliseconds, so
+// it is rare and permanent: nothing about the assertions says which day they
+// meant.
+func pinDayWindowClock(t *testing.T, at time.Time) time.Time {
+	t.Helper()
+	prev := dayWindowNow
+	dayWindowNow = func() time.Time { return at }
+	t.Cleanup(func() { dayWindowNow = prev })
+	return at
+}
+
 // TestDashboardFixtureRunLandsInsideTheWindow pins what runFinishedToday
 // promises the two DB fixtures that call it: at any hour, in any zone, the
 // run it places is inside the days=1 window the endpoints open. The window is
@@ -193,7 +210,7 @@ func TestDashboardEndpoints(t *testing.T) {
 	// A 600s run that finished inside today's window at every hour of the
 	// clock — see runFinishedToday for what a `now - 30m` fixture does to the
 	// agent-runtime assertions just after midnight.
-	started, completed := runFinishedToday(time.Now(), dashboardFixtureLoc(t))
+	started, completed := runFinishedToday(pinDayWindowClock(t, time.Now()), dashboardFixtureLoc(t))
 
 	mkTaskWithUsage := func(issueID string, status string, tokens int64) {
 		var taskID string
@@ -1662,7 +1679,7 @@ func TestDashboardFailuresCountNeverStartedTasks(t *testing.T) {
 	// below runs on the exact start-of-day cutoff and filters on
 	// completed_at, so a run timed off the wall clock disappears from it for
 	// the first twenty minutes of the day.
-	started, completed := runFinishedToday(time.Now(), dashboardFixtureLoc(t))
+	started, completed := runFinishedToday(pinDayWindowClock(t, time.Now()), dashboardFixtureLoc(t))
 
 	// startedAt is nullable so the queue-expiry case can be modelled exactly:
 	// completed_at set, started_at absent.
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
Thanks for tracking down the midnight flake. I’m requesting changes because the patch does not yet fix both tests named in the PR description.

`TestDashboardFailuresCountNeverStartedTasks` builds its own timestamps around `server/internal/handler/dashboard_test.go:1561`:

```go
now := time.Now().UTC()
started := now.Add(-30 * time.Minute)
completed := started.Add(10 * time.Minute)
```

That fixture is not touched by this diff. Its `by-agent` case uses the exact start-of-day window and filters by `completed_at`, so it can still produce the reported empty result during the first twenty minutes after midnight. Please update that fixture as well (ideally through one shared helper) and adjust the PR description if necessary.

Please also make the timezone assumption explicit. The new clamp uses UTC midnight, while the endpoint window is anchored to the viewer’s timezone. The current test only lines up because the fixture user falls back to UTC. Either pass `tz=UTC` in the affected requests or calculate the fixture boundary in the same timezone as the request, so the flake does not return when a non-UTC timezone is introduced.

The product direction is good and the root-cause analysis for `TestDashboardEndpoints` is correct; these fixture/window mismatches are the remaining blockers.

seacen:
@multica-eve — head is `2986d3598`. Both points fixed, and the description is corrected: you were right that it claimed something untrue.

## The second fixture

`TestDashboardFailuresCountNeverStartedTasks` built its own timestamps and my diff never touched them. The PR text said "both tests share the fixture" — I asserted that without reading the second test, and it was false. It is true now: both call one helper.

```go
func runFinishedToday(now time.Time, loc *time.Location) (started, completed time.Time) {
	local := now.In(loc)
	started = time.Date(local.Year(), local.Month(), local.Day(), 0, 0, 0, 0, loc)
	return started, started.Add(10 * time.Minute)
}
```

Anchored to the day boundary rather than clamped to it, so what time the suite runs is no longer an input. The start-of-day arithmetic matches `sinceFromDays` (runtime.go:391) so both land on the same instant.

## The timezone, and why UTC was the wrong choice for it

I pinned `?tz=` on all seven windowed requests from the same constant the fixture reads. That was still not enough, and adversarial review caught it: **with the fixture in UTC the pin does nothing.** Drop `?tz=` and the fallback is UTC too, so the window opens exactly where the fixture already sits. Pin it to a zone the fixture never used and an eastern zone only widens the window. Green either way — which is your original point, that the test lined up by coincidence rather than by construction.

The fixture zone is `Asia/Tokyo` now. Its run finishes at 15:10 UTC the *previous* day, so any request whose window opens at UTC midnight is opening after the run ended:

| mutation | result |
|---|---|
| request pinned to UTC, fixture Tokyo | **FAIL** — `expected >=1 task for agent, got 0` |
| `?tz=` replaced by a no-op parameter | **FAIL** — both dashboard tests, together |
| both restored | PASS |

Both tests failing together on that second one is the shared helper doing its job.

`TestDashboardFixtureRunLandsInsideTheWindow` covers the helper directly: synthetic clocks at 00:00, 00:05, 00:19:59, 12:00 and 23:59:59 across UTC, Asia/Kolkata and America/Los_Angeles, asserting against `sinceFromDays`'s own output rather than a second copy of the arithmetic.

## One thing worth flagging for anyone re-running this

`internal/handler` skips its entire suite and exits 0 when `DATABASE_URL` is unreachable. My first pass at the mutations above ran without a database and every one of them "passed" — I nearly reported that the timezone gap could not be reproduced. It reproduces fine against a live Postgres. Counting `=== RUN` lines is the only way to tell the two apart, and it is now in the commit message.

Full package against a live Postgres: 2521 tests, 0 failures. `gofmt`, `go vet` clean. CI 9/9.

multica-eve (changes requested):
Thanks for addressing the previous review: both dashboard tests now use the shared fixture helper, and the fixture/request timezone is explicitly wired through one constant. I found two remaining clock-dependent blockers on the updated head.

1. **The integration fixtures can still fall out of the window when the test crosses midnight.** `TestDashboardEndpoints` and `TestDashboardFailuresCountNeverStartedTasks` capture `time.Now()` when building the fixture (`server/internal/handler/dashboard_test.go:196` and `:1665`), but the handlers later call `time.Now()` again while computing the cutoff (`server/internal/handler/runtime.go:444`). If the fixture is created at `23:59:59.900` Tokyo time and the request computes its cutoff at `00:00:00.100`, `runFinishedToday` produces a completion at the previous day’s `00:10`, while the request has advanced to the next day’s `00:00`; the row is excluded. I reproduced this with a synthetic-clock test. `TestDashboardEndpoints` performs inserts and a rollup between those two clock reads, so the real race window is larger than the 200 ms reproduction. Please make the fixture and cutoff share a fixed/injectable clock, or otherwise make the fixture valid across that rollover, and add a regression case where fixture creation and request handling straddle midnight.

2. **The timezone pin is not deterministically load-bearing.** The comment at `server/internal/handler/dashboard_test.go:22-29` says that using `Asia/Tokyo` makes dropping or replacing `tz` fail. That only holds during part of the UTC day. At `2026-03-01 20:00Z`, for example, the Tokyo fixture completes at `2026-03-01 15:10Z`; if `tz` is removed and the request falls back to UTC, its cutoff is `2026-03-01 00:00Z`, so the fixture is still included and the mutation remains green. The same issue exists from roughly `15:00–23:59Z`. `TestDashboardFixtureRunLandsInsideTheWindow` always supplies the same location to both sides, so it proves that the helper is self-consistent but does not guard the seven request URLs against losing or changing their timezone. Please add a fixed-clock negative test that deterministically exercises a mismatched request timezone and covers the actual request/fixture coupling.

The product direction remains correct, and the new pure boundary test passes all 15 subcases. `go vet`, formatting, diff checks, and the current 9/9 CI checks are clean. The remaining request is specifically to remove these last wall-clock dependencies before merging.

seacen:
@multica-eve — head is `e6b1f722e`. Both blockers fixed, and the second one is a correction to something I asserted in the last round's commit message.

## 1. One clock for the fixture and the window

`parseDaysCutoff` now reads `dayWindowNow` (`runtime.go`), a variable production never changes and a test can pin. Both integration fixtures build from the instant they pin, so the two sides describe one moment instead of two clock reads with inserts and a rollup between them.

`TestDashboardFixtureSurvivesTheMidnightStraddle` asserts it at the instant where it used to bite: the clock pinned to `23:59:59.900` in the fixture's own zone, one tenth of a second of the day left when the run is written.

## 2. You are right that the Tokyo argument only held for part of the day — including in my own proof

This is worth stating plainly. Last round I claimed Asia/Tokyo made the pin load-bearing and showed two red mutations to back it. Those mutations ran at **13:28 UTC**, which is inside the window where the claim happens to hold. At 20:00Z the Tokyo fixture completes at 15:10Z, a UTC fallback cutoff is 00:00Z the same day, and the run is still counted — exactly your example. I verified at one instant and wrote it up as a property.

`TestDashboardMismatchedRequestTimezoneHidesTheRun` replaces that argument with a pinned one: clock at `2026-03-01 02:00Z`, where Tokyo's day began at 15:00Z the day before and UTC's began two hours ago. It asserts both directions — the matched request counts the run (premise), the request pinned to UTC counts zero (the coupling). It exercises the real request path, not the helper.

## Mutation results, live Postgres

| revert | result |
|---|---|
| the clock seam, back to `time.Now()` | straddle test **FAILS** — `agent-runtime reported 0s for a run built at 2026-03-01T23:59:59.9+09:00, want >=600` |
| fixture zone back to `UTC` | mismatch test **FAILS** — `a request pinned to UTC counted 1200s of a run the fixture placed in UTC` |
| both restored | PASS |

Full package: 2523 tests, 0 failures, `-race` clean, `go vet` clean. CI 9/9.

## On the production change

The clock seam is the one line this touches outside tests, and it is there because you asked for an injectable clock rather than a wider fixture. If you would rather it were a handler field or a context value than a package variable, say so — the three tests that pin it are sequential, so a variable is safe here, but it is your call whether that is the shape you want in `runtime.go`.

multica-eve (approved):
Thanks — the two remaining blockers are resolved on `e6b1f722e`.

- The fixture and request cutoff now share the same pinned instant through `dayWindowNow`. The midnight-straddle integration test exercises the actual handler path, and reverting the clock seam makes it fail.
- The timezone mismatch test now uses a deterministic `02:00Z` instant and exercises the actual dashboard request. Reverting the fixture timezone to UTC makes it fail.

I reran the four affected dashboard tests against a freshly migrated isolated PostgreSQL database under `-race`; they all passed. The pure timezone/DST tests, `go vet`, formatting/diff checks, and all 9 CI checks also pass.

Non-blocking nits: the broad comment near the top of `dashboard_test.go` still reads as though Tokyo makes the mismatch fail at every time of day, and the PR description's “3 lines plus a comment” summary is now stale. Neither affects correctness.

Approving.

PR #6874, opened 2026-08-13
fix(desktop): stop the dark-mode keyline shifting colour where the tab merges into the frame (MUL-6106)

tim-smart:
On desktop in dark mode, the keyline visibly changed colour where the active tab merges into the content frame. There were two independent causes, both rooted in the same thing: `--surface-border` is translucent in dark mode (`oklch(1 0 0 / 10%)`) and opaque in light mode, so the same token renders differently depending on what is behind it, and renders *twice as bright* if it is drawn on top of itself.

## 1. The tab's border composited over the wrong backdrop

The cap painted its `--page-canvas` fill under its own 1px border (`background-clip` defaults to `border-box`), while the flare arcs and the card's ring composite over `--app-shell`:

| keyline | composites over | rendered |
| --- | --- | --- |
| active tab's cap border | `--page-canvas` | `rgb(41,41,43)` |
| flare arcs | `--app-shell` | `rgb(36,36,38)` |
| content card's ring (outset) | `--app-shell` | `rgb(34,35,36)` |

Fixed with `bg-clip-padding`, which keeps the fill inside the border box. This is the existing pattern here for translucent borders (`button.tsx`, `sheet.tsx`, `combobox.tsx`).

## 2. The flare stacked a second keyline on the frame's ring

The flare's bottom row overlaps the card's top ring, and both draw `--surface-border`. The arc composited *over* the ring rather than replacing it, so two translucent keylines stacked:

```
0.1 x 255 + 0.9 x 35 = 57
```

Measured across that 1px band: the straight ring reads **35**, and the moment it enters the flare it jumps to **56**. That is a step of 22, far more visible than the first defect, and it sits exactly at the straight-to-curve junction.

Fixed by painting the gradient on an opaque `--app-shell` layer, so the flare replaces every pixel it covers instead of tinting it. The notch still shows the shell, because that layer *is* the shell.

## Verification

A harness rendered the tab chrome from the class strings and the `tabFlareBackground()` template read directly out of `tab-bar.tsx` and `desktop-layout.tsx`, so the replica could not drift from the source. Painted pixels sampled along the seam:

| variant | tab border | arc at 45° | arc at junction | frame ring | spread |
| --- | --- | --- | --- | --- | --- |
| dark, before | 41 | 36 | 57 | 34 | 23 |
| dark, after | 36 | 36 | 36 | 34 | 2 |
| light, before | 228 | 228 | 228 | 228 | 0 |
| light, after | 228 | 228 | 228 | 228 | 0 |

The residual 2 in dark mode is the card's own `--surface-shadow` darkening its ring, which is present in light mode too.

Light mode is byte-identical either way, as expected from its opaque token. The flare's own top and left edges were also checked against the surrounding shell in both themes to confirm the opaque layer prints no visible square: they match exactly.

`pnpm typecheck`, `pnpm lint` and `pnpm test` pass for `@multica/desktop` (465 tests). The regression tests assert both mechanisms, and each was confirmed to fail when its fix is reverted. The harness was scratch tooling and is not part of this PR.

### DIFF
--- a/apps/desktop/src/renderer/src/components/tab-bar.test.tsx
+++ b/apps/desktop/src/renderer/src/components/tab-bar.test.tsx
@@ -264,6 +264,26 @@ describe("TabBar merged-tab chrome", () => {
     );
   });
 
+  // The flare's bottom row sits on top of the content card's top ring. Both
+  // draw --surface-border, so in dark mode the arc would composite over the
+  // ring rather than replace it and the two translucent keylines would stack
+  // into a markedly lighter line right at the straight-to-curve junction. The
+  // opaque --app-shell layer under the gradient makes the flare replace
+  // whatever it covers.
+  it("paints each flare on an opaque shell layer", () => {
+    const { getByLabelText } = render(<TabBar />);
+    const flares = getByLabelText("Issues")
+      .closest("[data-tab-frame]")
+      ?.querySelectorAll('[style*="radial-gradient"]');
+
+    expect(flares).toHaveLength(2);
+    for (const flare of flares ?? []) {
+      expect(flare.getAttribute("style")).toContain(
+        "var(--page-canvas) 10.2px), var(--app-shell)",
+      );
+    }
+  });
+
   it("draws the merged cap only on the active tab", () => {
     const { getByLabelText } = render(<TabBar />);
     expect(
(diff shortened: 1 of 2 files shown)


tim-smart:
junction-comparison

multica-eve (approved):
Thanks for the thoughtful fix and the detailed visual verification. The before/after evidence made this subtle dark-mode seam issue clear, and the scoped regression coverage looks solid.

PR #6877, opened 2026-08-13
feat(inbox): add a keyboard shortcut for archiving the open notification

tim-smart:
Adds a keyboard shortcut for archiving the notification currently open in the inbox.

## Behaviour

- The default binding is plain `E`, is rebindable, and appears in Settings → Keyboard shortcuts in all four locales.
- It archives from the main inbox and restores while viewing the archive, matching the selected row's existing action.
- Selection advances to the next notification after archiving.
- With no notification open, the key is not consumed.

## Guards

The handler ignores editable targets, IME composition, auto-repeat, events consumed by a nearer handler, and keypresses while a dialog is open.

## Tests

Coverage includes archive, restore, selection advancement, each guard, the no-selection case, and the shortcut definition. `pnpm typecheck`, `pnpm test`, and `pnpm lint` pass.

### DIFF
--- a/packages/core/shortcuts/definitions.test.ts
+++ b/packages/core/shortcuts/definitions.test.ts
@@ -93,6 +93,15 @@ describe("keyboard shortcut definitions", () => {
     expect(action.allowInEditable).toBe(true);
   });
 
+  it("keeps the inbox archive key out of editable controls", () => {
+    const action = SHORTCUT_ACTION_BY_ID.archiveInboxItem;
+    expect(action.defaultShortcut).toEqual(createShortcutChord("E"));
+    // A bare letter is only safe because the inbox handler skips editable
+    // targets. Flipping this flag would archive the open notification while
+    // the user types an "e" into its comment composer.
+    expect(action.allowInEditable).toBe(false);
+  });
+
   it("strictly distinguishes Command and Control on macOS", () => {
     const commandF = createShortcutChord("F", { primary: true });
     const controlF = createShortcutChord("F", { control: true });
--- a/packages/core/shortcuts/definitions.test.ts
+++ b/packages/core/shortcuts/definitions.test.ts
@@ -96,9 +96,6 @@ describe("keyboard shortcut definitions", () => {
   it("keeps the inbox archive key out of editable controls", () => {
     const action = SHORTCUT_ACTION_BY_ID.archiveInboxItem;
     expect(action.defaultShortcut).toEqual(createShortcutChord("E"));
-    // A bare letter is only safe because the inbox handler skips editable
-    // targets. Flipping this flag would archive the open notification while
-    // the user types an "e" into its comment composer.
     expect(action.allowInEditable).toBe(false);
   });
 
--- a/packages/core/shortcuts/definitions.ts
+++ b/packages/core/shortcuts/definitions.ts
@@ -12,6 +12,7 @@ export type ShortcutActionId =
   | "toggleChat"
   | "findInIssue"
   | "openThreadNav"
+  | "archiveInboxItem"
   | "send"
   | "goInbox"
   | "goChat"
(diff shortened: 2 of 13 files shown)


Bohan-J (changes requested):
`E` is the right key for this — it matches Gmail, Superhuman, Linear and Front, so the muscle memory transfers for free, and the repo already has a bare-letter precedent in `createIssue` (`C`). The guard set is thorough (editable targets, IME, auto-repeat, `defaultPrevented`, app modals), all four locales are covered, and the test coverage is genuinely good. Reading the binding through `getShortcut()` at press time rather than subscribing to overrides is a nice touch — rebinding in Settings takes effect immediately.

Two things I'd like addressed before this merges.

## 1. The dialog guard misses Radix popovers, so `E` can archive behind an open menu

`useModalStore.getState().modal` only tracks the app's own modal stack. It does not cover Radix layers — `DropdownMenu`, `Popover`, `Select`, `AlertDialog`.

Concrete repro:

1. Select a notification so the detail pane opens.
2. Open the batch-operations `DropdownMenu` in the inbox header (`packages/views/inbox/components/inbox-page.tsx:443`).
3. Press `e`, expecting menu typeahead.

Radix's menu `onKeyDown` calls `handleTypeaheadSearch` for character keys but only calls `preventDefault()` for `Tab` and the first/last navigation keys — never for plain characters. The menu content is portaled into `document.body`, so the event bubbles to the document listener with `defaultPrevented === false`, `event.target` is a `menuitem` (not editable), and `modal` is null. Result: the notification is archived silently behind the open menu, with no visible feedback.

The same shape applies to the status/assignee dropdowns and any `AlertDialog` in the detail pane, which is the more likely path in practice since that pane is on screen whenever the shortcut is live.

Suggested fix: widen the guard from "an app modal is open" to "some layer owns the keyboard". Either check for an open Radix layer (`document.querySelector('[data-radix-popper-content-wrapper]')`, or a `[data-state="open"]` popper), or invert it and require `event.target` to sit inside the inbox page's own subtree. The second is the more robust of the two, since it also covers any future portal-based surface.

## 2. A bare key that mutates state should give visible feedback

`handleArchive` currently only surfaces a toast on failure — success is completely silent. That's fine for the mouse path, where the click is an unambiguous act, but a single unmodified key makes accidental fires far more likely, and right now nothing on screen confirms what happened.

Please add a confirmation toast on archive. To be clear on scope: **an undo action is not required for this PR** — a plain confirmation is enough here, and undo can be a follow-up.

Two implementation notes:

- **Put the toast inside `handleArchive` / `handleUnarchive`, not in the keydown handler.** That way the row action, the detail pane's archive button, `onDone`, and `E` all behave identically. Putting it in the keyboard branch would produce a "keyboard shows feedback, clicking doesn't" split.
- Nothing new is needed infrastructurally: `sonner`'s `toast` is already imported at `inbox-page.tsx:38`, and `<Toaster />` is mounted without a `position` prop in both `apps/web/app/layout.tsx` and the desktop renderer, so it defaults to bottom-right. Copy will need the four locales.

## Non-blocking

- `actionOnSelectedRef.current` is assigned during render (`inbox-page.tsx:355`). Behaviour is fine today, but writing a ref during render is something React explicitly advises against and the React Compiler will flag. Syncing it in an effect avoids that.
- Discoverability is limited to Settings → Keyboard shortcuts. Showing the chord in the archive button's tooltip and the row action would be cheap.
- `category: "general"` sits a little awkwardly for an inbox-only action — it lists alongside genuinely global actions like search and create-issue. With only `general` and `navigation` available today there's no better bucket, and the description does say "inbox", so this is just a note.
- Worth a follow-up rather than this PR: without `J`/`K` or arrow navigation, `E` only serves a straight top-to-bottom sweep — skipping a notification still sends you back to the mouse. That's exactly why `E` ships alongside `J`/`K` in the products it's borrowed from.

Bohan-J (approved):
Thanks for the quick turnaround — both blocking items are resolved, and the shape you landed is better than what I suggested.

**Popup guard.** `isPortalLayerShortcutTarget()` keys off ARIA roles (`menu` / `dialog` / `alertdialog` / `listbox`) plus the `data-base-ui-inert` marker for modal layers. Matching on roles rather than a library-specific attribute is the more durable choice: it covers the header dropdown, the detail pane's select and popover surfaces, and any future portal-based layer without needing to be revisited. The inert check also catches the case where a modal is open but focus never left the page, which my suggested fix would have missed.

**Confirmation toast.** Putting it in the `onSuccess` of the shared `handleArchive` / `handleUnarchive` is exactly right — the row action, the detail button, `onDone`, and the shortcut all confirm identically, with copy in all four locales.

Also noted you moved the ref assignment into an effect, which clears the render-phase write.

CI is green across all ten checks. Merging now — thanks for the contribution.

PR #6887, opened 2026-08-13
MUL-6116: fix(repocache): make git timeouts configurable, recover from corrupted worktree admin metadata

### LINKED ISSUE #6879: repo checkout: CLI HTTP timeout shorter than daemon git timeout; canceled checkout leaves worktree in unrecoverable state
## Summary

The `/repo/checkout` daemon endpoint and the `multica repo checkout` CLI that calls it have two mismatched timeouts, plus a follow-on bug where a timed-out checkout leaves behind a worktree that fails every subsequent retry with an unrelated error until it's manually removed.

## Details

**1. CLI-side HTTP timeout is shorter than the daemon's own git timeout**

- `server/cmd/multica/cmd_repo.go` sets `http.Client{Timeout: 5 * time.Minute}` when POSTing to the daemon's local `/repo/checkout` endpoint.
- `server/internal/daemon/repocache/cache.go` (`repoCacheGitTimeout`) gives the daemon's own git subprocesses (fetch, `git worktree add`, etc.) up to **10 minutes**.

For a large repo (tens of thousands of files), `git worktree add` can legitimately take longer than 5 minutes to materialize the working tree, especially on network-backed storage. When that happens, the CLI's HTTP client hits its own 5-minute deadline and cancels the request *before* the daemon's internal 10-minute timeout would ever fire. The daemon logs show the underlying git process still actively writing files when the request context is canceled — the daemon isn't hung, the caller just gave up first.

Observed daemon-log signatures for the same root repo, ~20 hours apart:

```
# Daemon's own internal timeout firing (git process itself stalled):
error="... git worktree add: ... Updating files: 95% (40331/42421): git command timed out after 10m0s: context deadline exceeded"

# CLI's shorter timeout firing first (git process still progressing normally):
error="... git worktree add: ... Updating files: 54% (23008/42477): context canceled"
```

The CLI surfaces this second case to the end user as:
```
connect to daemon: Post "http://127.0.0.1:<port>/repo/checkout": context deadline exceeded
```
which reads like the daemon itself is unresponsive, when in fact it's the CLI's own client-side deadline.

**2. A canceled checkout leaves a corrupted worktree that poisons all future retries**

When the CLI cancels the request, the daemon does not clean up the partially-created worktree. On the next attempt against the same task workdir, `CreateWorktree` (`repocache/cache.go`) calls `isGitWorktree(worktreePath)`, sees something already there, and takes the "update existing worktree" path (`git reset --hard`) instead of recreating from scratch. Because the bare repo's internal `worktrees/<name>/` admin metadata is incomplete from the canceled operation, this fails immediately with:

```
update existing worktree: git reset --hard: fatal: not a git repository: <bare-repo>/worktrees/<name>: exit status 128
```

This is a much more confusing error than the original timeout, and it reproduces on every retry until someone manually deletes the broken worktree registration (`git worktree remove --force` on the bare repo) — there's no automatic detection/recovery from a corrupted worktree in this path.

## Suggested fixes

1. The CLI's `/repo/checkout` client timeout shouldn't be shorter than the daemon's own internal git timeout — either drop the client-side timeout and let the daemon's timeout be authoritative, or make both configurable from the same source.
2. When `git worktree add`/checkout is canceled or fails, the daemon should clean up (or on next use, detect-and-recreate) the partial worktree rather than leaving state that fails all future retries with an unrelated git error.

## Environment

- Checkout target: a monorepo with ~42,000 tracked files.
- Reproduced 3 times over ~20 hours against the same repo, in the same daemon.

AnkeshThakur:
## What does this PR do?

Fixes two related bugs behind #6879:

1. The CLI's checkout-request timeout was hardcoded to 5 minutes, shorter than the daemon's own internal git timeout (10 minutes). On a large repo this could cancel a legitimately still-running checkout with an opaque `context deadline exceeded`, even though the daemon's git operation would have succeeded given more time.
2. A canceled `git worktree add`/`reset --hard` can leave a linked worktree's `.git` admin metadata structurally incomplete (the `.git` file is present but the admin directory under the bare repo is missing bookkeeping files). Every subsequent retry against that worktree then failed immediately with an opaque `fatal: not a git repository`, and only self-healed once the GC loop's periodic `git worktree prune` ran (default every 2h).

## Related Issue

Closes #6879

## Type of Change

## Changes Made

- `repocache.New` now takes a `gitTimeout time.Duration` parameter (falls back to the existing 10m default `repocache.DefaultGitTimeout` when <= 0), configurable in the daemon via a new `MULTICA_REPO_GIT_TIMEOUT` env var / `Config.RepoGitTimeout` field.
- The CLI's checkout wait budget is now configurable via `MULTICA_REPO_CHECKOUT_TIMEOUT` (Go duration string or plain integer seconds) and its default is raised from 5m to 15m, comfortably above the daemon's default git timeout.
- `CreateWorktreeContext` now checks whether an existing worktree's admin metadata is structurally intact (`HEAD`, `commondir`, `gitdir` present under the resolved admin dir) before reusing it. If not, it removes the corrupted worktree and prunes the bare repo's stale registration, then falls through to a fresh create — instead of failing every retry until the next GC pass.
- Added unit tests: `TestGitWorktreeAdminIntact`, `TestCreateWorktreeRecoversFromCorruptedAdminMetadata`, `TestRepoCheckoutTimeout`.

## How to Test

```
cd server
go build ./... && go vet ./...
go test ./internal/daemon/repocache/... ./cmd/multica/... ./internal/daemon/...
```

Manual repro: set `MULTICA_REPO_GIT_TIMEOUT=5s` and `MULTICA_REPO_CHECKOUT_TIMEOUT=2s` on a slow/large checkout to confirm a clean timeout instead of a hang, then retry the same checkout and confirm it succeeds rather than repeating `fatal: not a git repository`.

## Checklist

- [x] My changes follow the existing code style and conventions

## AI Disclosure

This PR was developed with Claude Code (Claude Sonnet 5). I asked it to investigate the root cause behind #6879 (a repo-checkout timeout failure) and implement the fix suggested in that issue — making the mismatched timeouts configurable, and additionally fixing the corrupted-worktree-retry failure that was compounding the original symptom. All code and tests were reviewed by me before submission.


AnkeshThakur:
This branch had fallen behind `main` — 49 new commits landed upstream since this PR opened, including additions to the `Config` struct in `config.go` in the same region as this PR's new `RepoGitTimeout` field.

Merged `main` in and resolved the conflict; it was purely additive on both sides (no semantic overlap), so resolution was a straightforward combine + gofmt realignment. Verified post-merge with `go build ./...`, `go vet ./...`, and `go test -race ./internal/daemon/... ./cmd/multica/...` — all pass.

PR is now `MERGEABLE` / `CLEAN`.

AnkeshThakur:
Update: `main` advanced again (78 more commits) after the prior merge, causing a second conflict — this time a single-line import ordering conflict in `config.go` (this branch's `repocache` import vs. a new `execenv` import from `main`, purely additive, both genuinely used). Merged and resolved; verified with `go build ./...`, `go vet ./...`, and `go test -race ./internal/daemon/... ./cmd/multica/...` — all pass.

PR is `MERGEABLE` / `CLEAN` again, all CI checks green.

PR #6889, opened 2026-08-13
fix(daemon): recover in-flight tasks when a runtime deregisters

nimrc:
## The gap

`DaemonDeregister` is the one moment the server knows for certain that a runtime is going away. Today it flips the runtime offline and drops that knowledge on the floor — whatever the runtime was executing stays `dispatched`/`running` with no failure, no `task:failed` event, no agent-status reconcile, and no retry.

The startup path already does the right thing. The daemon calls `POST /runtimes/{id}/recover-orphans`; `RecoverOrphanedTasksForRuntime` fails the leftovers with `failure_reason = 'runtime_recovery'`; `HandleFailedTasks` funnels them into the auto-retry that `retryableReasons` already permits.

**That only rescues a task if the same runtime id comes back to ask for it.** A daemon that is *replaced* rather than restarted re-registers under fresh ids — a new host, a new container, a wiped state directory — and the orphans stay pinned to ids nobody will ever query again. The only thing that eventually notices is the stale-task sweeper, up to a 2.5h in-process timeout later.

## How we hit it

We moved a daemon from a long-lived VM onto Kubernetes. Every rolling deploy silently dropped whatever was mid-flight: the replacement pod always comes up with new runtime ids, so `recover-orphans` at startup has nothing to find, and the previous pod's in-flight tasks are stranded.

Our daemon is never idle in practice — sampling every 30s for 24h, exactly one 30-second window had nothing running — so "restart when quiet" is not available to us as a workaround, and `terminationGracePeriodSeconds` buys nothing because the daemon deregisters and exits rather than draining.

## The change

In the deregister loop, after `SetAgentRuntimeOffline`, run the same `RecoverOrphanedTasksForRuntime` query and hand the rows to `TaskService.HandleFailedTasks` — the identical pipeline `RecoverOrphanedTasks` uses. A graceful shutdown now costs a retry instead of a lost task.

Deliberately unchanged:

- **Best-effort.** A failed recovery is logged and skipped; the runtime is already offline and the stale-task sweeper remains the backstop it has always been.
- **No new failure reason.** `runtime_recovery` already exists, is already in `retryableReasons`, and already describes exactly this.
- **No change to retry policy.** Autopilot tasks, exhausted budgets and non-retryable reasons are filtered by `retryEligible` as before.

The response now returns `orphaned`/`retried` counts, mirroring `RecoverOrphanedTasks`.

## Testing

`go build` and `go vet` clean; `go test ./internal/handler/` passes.

I did not add a test, and I want to flag that rather than leave it implicit: `internal/handler`'s suite needs a live Postgres (`TestMain` skips without `DATABASE_URL`), and the sibling path this mirrors — `RecoverOrphanedTasks` — has no test either, so there is no existing fixture to extend. I would rather not submit a DB-backed test I could not execute. Happy to add one if you point me at the pattern you would want.

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -907,6 +907,9 @@ func (h *Handler) DaemonDeregister(w http.ResponseWriter, r *http.Request) {
 	// Track affected workspaces for WS notifications.
 	affectedWorkspaces := make(map[string]bool)
 
+	// Tasks still in flight on the runtimes going away, recovered below.
+	var orphaned []db.AgentTaskQueue
+
 	for i, rid := range req.RuntimeIDs {
 		// Look up the runtime and verify ownership.
 		rt, err := h.Queries.GetAgentRuntime(r.Context(), runtimeUUIDs[i])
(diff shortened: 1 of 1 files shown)


nimrc:
Reopened as #6891 from a differently-named branch; same commit. Please review there.

PR #6899, opened 2026-08-13
fix(skills): normalize imported Windows file paths

### LINKED ISSUE #6893: Skill import on Windows stores supporting-file paths with backslashes, breaking bundle resolution
## Bug report

**Environment**
- OS: Windows 10
- multica daemon: v0.4.24

**What happened**
Importing a skill on a Windows host stores supporting files with backslash
path separators (e.g. `templates\template.html`) instead of forward slashes
(`templates/template.html`). When an agent task then needs that skill, the
daemon fails to resolve the bundle:

    skill bundle unavailable: skill "architecture-diagram" (id=..., 18780 bytes) after 488ms: resolve skill bundle returned invalid bundle: skill_id=... source=workspace hash=sha256:...

**Root cause**
`server/internal/daemon/skill_cache.go` → `safeSkillFilePath` rejects any path
containing `\`:

    if p == "" || strings.Contains(p, "\x00") || strings.HasPrefix(p, "/") || strings.Contains(p, "\\") {
        return false
    }

so `validateSkillBundle` returns false and the bundle is reported invalid.

**Impact**
In our workspace 35 of 94 imported skills (349 supporting files) were affected;
we worked around it by rewriting every path to forward slashes via
`multica skill files upsert` + delete.

**Related**
Imported content also ends up with CRLF line endings on Windows (74 skills /
359 files). It does not break the hash (self-consistent), but is the same
Windows-import artifact.

**Expected**
Imported supporting-file paths should always be normalized to forward slashes
(and ideally content to LF) regardless of host OS.

**Suggested fix**
Normalize file paths (`\` → `/`) during skill import.

Jamison929611:
## What does this PR do?

Normalizes supporting-file paths reported by runtime-local skill imports before they are persisted, so Windows-style nested paths such as `templates\template.html` become the canonical `templates/template.html` form accepted by workspace skill bundle validation.

The normal daemon collector already uses `filepath.ToSlash`. The remaining gap was the server report boundary used by older or alternate reporters: it validated paths with host-dependent `filepath` semantics and persisted accepted values unchanged. A backslash path could therefore reach the database and later make `safeSkillFilePath` reject the entire bundle.

Normalization now happens once at that import trust boundary, before both create and overwrite persistence. Validation then uses platform-independent slash-path semantics.

This PR intentionally addresses only the path-separator defect that makes workspace skill bundles invalid. The CRLF observation in #6893 is self-consistent and non-blocking, so content line-ending normalization is explicitly out of scope here.

## Related Issue

Fixes #6893

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Canonicalize runtime-local import file separators in `server/internal/handler/runtime_local_skills.go` before the shared create/overwrite branches.
- Add platform-independent safety checks in `server/internal/skill/import_path.go` for empty/NUL, rooted/absolute, drive, UNC, traversal, and non-canonical paths.
- Reject paths that collide after slash normalization instead of allowing order-dependent overwrite behavior.
- Add focused tests for Windows/POSIX paths, unsafe inputs, collisions, create/overwrite responses, and daemon bundle acceptance.

## How to Test

1. `cd server && go test ./internal/skill -count=1 -v`
2. `cd server && go test ./internal/daemon -run '^(TestReportLocalSkillImportResult_RetriesOn500AndEventuallySucceeds|TestSkillBundleResolveTimeout|TestEnsureTaskSkillBundles_CachesEachSuccessAcrossDispatches|TestEnsureTaskSkillBundles_AcceptsServerSideSkillUpdate|TestEnsureTaskSkillBundles_RejectsPluginHashDrift|TestEnsureTaskSkillBundles_DeadlineIsLabelledStructurally|TestSkillBundleCacheLoadStore|TestSkillBundleCacheRejectsCorruptBundle|TestValidateSkillBundleAcceptsCanonicalNestedPath)$' -count=1 -v`
3. With PostgreSQL available, run `cd server && go test ./internal/handler -run '^(TestRuntimeLocalSkillImportFilesCanonicalizesWindowsPaths|TestRuntimeLocalSkillImportFilesRejectsUnsafePaths|TestRuntimeLocalSkillImportFilesRejectsCanonicalPathCollisions|TestRuntimeLocalSkillImport_CanonicalizesWindowsFilePath|TestRuntimeLocalSkillImport_OverwriteCanonicalizesWindowsFilePath)$' -count=1 -v`
4. `cd server && go vet ./internal/skill ./internal/handler ./internal/daemon`
5. `cd server && go build ./internal/skill ./internal/handler ./internal/daemon`

Local note: the database-backed handler tests were added and compile successfully, but this checkout had no PostgreSQL service available, so the package's `TestMain` skipped their execution locally. The non-database path tests, daemon tests, vet, Windows build, and Linux/amd64 cross-build ran successfully.

## Checklist

Safety/risk: the daemon's strict `safeSkillFilePath` consumer validation is unchanged. Invalid paths remain omitted as before; canonical collisions now fail the import explicitly because choosing one would make persisted content input-order dependent. Drive-letter paths are rejected intentionally at this Windows-report boundary. No content/line-ending normalization, binary-file support, database migration, historical cleanup, sibling skill-authoring API change, discovery, or symlink behavior is included.

## AI Disclosure

**AI tool used:** Yes

**Prompt / approach:** Assisted.

## Screenshots (optional)

### DIFF
--- a/server/internal/daemon/skill_cache_test.go
+++ b/server/internal/daemon/skill_cache_test.go
@@ -42,6 +42,20 @@ func TestSkillBundleCacheRejectsCorruptBundle(t *testing.T) {
 	}
 }
 
+func TestValidateSkillBundleAcceptsCanonicalNestedPath(t *testing.T) {
+	bundle := testSkillBundle()
+	bundle.Files[0].Path = "templates/template.html"
+	ref := skillRefFromBundle(bundle)
+	bundle.Hash = ref.Hash
+	bundle.SizeBytes = ref.SizeBytes
+	bundle.Files[0].SHA256 = ref.Files[0].SHA256
+	bundle.Files[0].SizeBytes = ref.Files[0].SizeBytes
+
+	if !validateSkillBundle(ref, bundle) {
+		t.Fatal("canonical nested file path should produce a valid bundle")
+	}
+}
+
 func testSkillBundle() SkillData {
 	bundle := SkillData{
 		ID:      "skill-1",
--- a/server/internal/handler/runtime_local_skills.go
+++ b/server/internal/handler/runtime_local_skills.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
+	skillpkg "github.com/multica-ai/multica/server/internal/skill"
 	"github.com/multica-ai/multica/server/internal/util"
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 	"github.com/multica-ai/multica/server/pkg/protocol"
@@ -527,6 +528,28 @@ type reportedRuntimeLocalSkill struct {
 	Files       []CreateSkillFileRequest `json:"files,omitempty"`
 }
 
+func runtimeLocalSkillImportFiles(reported []CreateSkillFileRequest) ([]CreateSkillFileRequest, error) {
+	reportedPaths := make([]string, len(reported))
+	for i, file := range reported {
+		reportedPaths[i] = file.Path
+	}
+	canonicalPaths, err := skillpkg.CanonicalRuntimeLocalFilePaths(reportedPaths)
+	if err != nil {
+		return nil, err
+	}
+
+	files := make([]CreateSkillFileRequest, 0, len(reported))
+	for i, canonicalPath := range canonicalPaths {
+		if canonicalPath == "" {
+			continue
+		}
+		file := reported[i]
+		file.Path = canonicalPath
+		files = append(files, file)
+	}
+	return files, nil
+}
+
 func cleanOptionalString(value *string) *string {
 	if value == nil {
 		return nil
@@ -851,12 +874,10 @@ func (h *Handler) ReportLocalSkillImportResult(w http.ResponseWriter, r *http.Re
 		description = *req.Description
 	}
 
-	files := make([]CreateSkillFileRequest, 0, len(body.Skill.Files))
-	for _, f := range body.Skill.Files {
-		if !validateFilePath(f.Path) {
-			continue
-		}
-		files = append(files, f)
+	files, err := runtimeLocalSkillImportFiles(body.Skill.Files)
+	if err != nil {
+		h.failLocalSkillImport(w, r, requestID, err.Error())
+		return
 	}
 
 	config := map[string]any{
--- a/server/internal/handler/runtime_local_skills_overwrite_test.go
+++ b/server/internal/handler/runtime_local_skills_overwrite_test.go
@@ -286,6 +286,33 @@ func TestRuntimeLocalSkillImport_OverwritePreservesIdentityAndBindings(t *testin
 	}
 }
 
+func TestRuntimeLocalSkillImport_OverwriteCanonicalizesWindowsFilePath(t *testing.T) {
+	if testHandler == nil {
+		t.Skip("database not available")
+	}
+
+	runtimeID := createRuntimeLocalSkillTestRuntime(t, testUserID)
+	name := fmt.Sprintf("overwrite-windows-path-%d", time.Now().UnixNano())
+	existingID := createImportTargetSkill(t, name, testUserID, nil)
+
+	got := runLocalSkillImport(t, runtimeID,
+		map[string]any{"skill_key": "review-helper", "action": "overwrite", "target_skill_id": existingID},
+		reportBundleBody(name, "overwritten description", "# overwritten", map[string]string{
+			`templates\template.html`: "template",
+		}),
+	)
+
+	if got.Status != RuntimeLocalSkillCompleted {
+		t.Fatalf("status = %s, want completed (error=%q)", got.Status, got.Error)
+	}
+	if got.Skill == nil || len(got.Skill.Files) != 1 {
+		t.Fatalf("imported files = %#v, want one file", got.Skill)
+	}
+	if got.Skill.Files[0].Path != "templates/template.html" {
+		t.Fatalf("file path = %q, want canonical slash path", got.Skill.Files[0].Path)
+	}
+}
+
 func TestRuntimeLocalSkillImport_OverwriteNonCreatorFails(t *testing.T) {
 	if testHandler == nil {
 		t.Skip("database not available")
--- a/server/internal/handler/runtime_local_skills_test.go
+++ b/server/internal/handler/runtime_local_skills_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"testing"
 	"time"
 )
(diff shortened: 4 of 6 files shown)


ADanMan (commented):
Nice, well-scoped fix — thanks for this. It's clearly root-caused and the tests are unusually thorough for a change this size.

**What I like**

- **Fix is at the right layer.** #6893 is that a Windows import persists `templates\template.html`, and the daemon's `safeSkillFilePath` (`server/internal/daemon/skill_cache.go:138`) rejects any `\`, so `validateSkillBundle` marks the whole bundle invalid. Normalizing `\`→`/` once at the `ReportLocalSkillImportResult` trust boundary is the correct fix rather than loosening the downstream check.
- **The normalizer's output is a provable strict subset of what the downstream gate accepts.** `CanonicalRuntimeLocalFilePath` (`server/internal/skill/import_path.go:11`) strips all backslashes and then applies the same leading-`/`, NUL, `path.Clean`/`clean != canonical`, `..`, `../` checks as `safeSkillFilePath`, so an accepted path can never re-trigger the #6893 rejection. Adding the drive-letter guard the downstream lacks is a nice bit of extra hardening.
- **Both create and overwrite consume the same normalized `files`**, so the two persistence paths stay consistent.
- **Index-preserving canonicalization is clean** (same-length slice, empty entries skipped, a valid path can never be empty → no drop/keep ambiguity).
- **Great tests**: traversal / UNC / drive-absolute / drive-relative / NUL / embedded-traversal / non-canonical-separator table, collision rejection, DB-backed create + overwrite handler tests, and a `validateSkillBundle` acceptance test proving the canonical nested path now passes the exact check that was failing. CI including `windows-execenv` is green.

**Issues, ranked**

1. **(Medium — scope/closing) `Fixes #6893` only covers half the issue.** #6893 reports two Windows artifacts: backslash separators (fixed here) *and* CRLF line endings (`74 skills / 359 files`), and its "Expected" explicitly asks for LF normalization. The diff only touches `Path`, never `Content`, so CRLF is untouched — and `Fixes #6893` will auto-close the issue and drop that sub-item. The issue notes CRLF is self-consistent / non-breaking, so deferring it is fine, but please make it explicit: either drop the auto-close and file a CRLF follow-up, or state CRLF is intentionally out of scope.

2. **(Low — completeness) Sibling endpoints still accept backslash paths.** `validateFilePath` (`server/internal/handler/skill.go:248`) uses host-dependent `path/filepath`; on the Linux server it treats `\` as an ordinary char, so the direct skill-authoring API (`skill.go:375`, `446`, `2359`) still lets a backslash path reach the DB, which `safeSkillFilePath` later rejects — the same failure mode you fix at the runtime-local boundary. Very likely out of scope for #6893 (and the archive path is unaffected — it uses the slash-based `path` pkg on zip names), but worth a one-line note so the direct-endpoint case isn't forgotten.

3. **(Low — behavior) Collision fails the whole import, but invalid paths are silently dropped.** `CanonicalRuntimeLocalFilePaths` (`import_path.go:37-38`) errors out — failing the entire import — the moment two reported paths normalize to the same value, whereas individually-unsafe paths are silently skipped (old behavior). A daemon reporting both `a\b` and `a/b` now hard-fails the skill instead of importing the rest. Defensible for an ambiguous bundle, but the drop-one-vs-fail-all asymmetry is a behavior change worth a deliberate decision + a short comment.

4. **(Nit) Drive-letter guard is stricter than the downstream gate.** `import_path.go:16` rejects any `X:` prefix (single ASCII letter + `:`), so a file literally named `c:notes.txt` — legal on POSIX and accepted by `safeSkillFilePath` — is dropped. Over-rejection is the safe direction; just noting the intentional divergence.

None of these block; (1) is the only one I'd want resolved (even if just by editing the PR description) before merge.

Jamison929611:
Resolved the scope/closing concern in the PR description: this change fixes the path-separator defect that makes bundles invalid; CRLF content normalization is explicitly out of scope because it is self-consistent and non-blocking.

The other observations remain intentional:
- sibling skill-authoring endpoints are outside this runtime-local import fix;
- canonical collisions fail the import because choosing one file would make persisted content input-order dependent;
- drive-letter prefixes are rejected deliberately at this Windows-report boundary.

No code change was needed.

ADanMan:
Thanks for the thorough responses — all reasonable. Clarifying CRLF as out-of-scope in the description resolves the one thing I'd have wanted addressed, and the rest are fair calls: the collision-fails-import choice in particular makes sense once it's framed as keeping persisted content independent of input order. Nothing further from me — nice fix.

Jamison929611:
Hi maintainers, the review feedback has been addressed and all checks are passing. Since there don't appear to be any remaining concerns, could you please merge this PR when you have a chance? Please let me know if anything else is needed. Thanks!

multica-eve (changes requested):
Before this change, a Windows user can import a skill successfully and still find it unusable when a task needs it. The intended outcome is that newly imported skills remain usable without requiring any configuration changes. This is a product behavior we should accept, although imports that are already broken would not be repaired by this PR.

The normalization approach is clear and the path-safety/collision tests are good. I am requesting changes because the patch does not yet establish that it covers the route which produced the data reported in #6893.

## Must fix

1. **Identify and cover the actual write path that produced the backslashes.** The standard daemon collector has called `filepath.ToSlash(rel)` in `server/internal/daemon/local_skills.go:372` since the runtime-local import feature was introduced, including in the reported v0.4.24. That standard reporter should therefore not send a backslash path. This PR only canonicalizes data received by `ReportLocalSkillImportResult` at `server/internal/handler/runtime_local_skills.go:877`, while the create/update/file-upsert paths still use the host-dependent `validateFilePath` in `server/internal/handler/skill.go:248` (callers at lines 375, 446, and 2366) and can persist the same invalid representation. Please confirm the reporter/import route from the issue author, an origin record, or a captured request, then add the fix and regression coverage at the route that actually produced the affected rows. Otherwise `Fixes #6893` may close the issue without fixing its real reproduction.

2. **Resolve the behavior for already persisted bad paths.** This change prevents new bad rows only at the one covered boundary; existing skills will still be rejected by `safeSkillFilePath`. Please either add a collision-safe remediation, or create a tracked follow-up and clearly document that affected skills must be re-imported. The PR should not imply that existing affected imports are restored.

Keeping CRLF normalization out of scope is reasonable because it is not what invalidates the bundle. The strict downstream validator and the fail-on-canonical-collision behavior should also remain.

**Recommendation: Request changes.** The product behavior is wanted and the implementation is strong within its chosen boundary, but the actual producer and persisted-data compatibility need to be resolved before this can be considered a complete fix for #6893.

Jamison929611:
I will verify the producer path before changing the normalization boundary. The standard runtime-local collector already applies `filepath.ToSlash`, so I have asked the reporter on #6893 for the exact creation/import flow, `config.origin`, versions, and a minimal reproduction or request shape. I will also trace the alternative create/update/file-upsert persistence paths against that evidence.

For already-persisted invalid paths, this PR will either include collision-safe remediation or link a tracked follow-up with explicit re-import guidance. The PR description will not imply that existing affected skills are repaired automatically. CRLF normalization remains out of scope, and the strict downstream validation and collision behavior will remain unchanged.

I will not make another code change until the actual producer path is established.

Issue #3929, opened 2026-06-09
[Bug]: Small UI bug


HerbertWYT:
### Deployment type

Official App

### What happened?

When I check the messages in my inbox, I find that the title bar at the top is obscured, as shown in the image. No matter how I scroll the message content area, it cannot be fully displayed. Only after I click the right sidebar button once and then collapse it again does the title bar return to normal.

System: macOS 26.5
Device: Mac mini M4
App version: Multica 0.3.18

### Steps to reproduce

1. Click Inbox, then click the top message.The title bar is obscured.

<img width="1774" height="1001" alt="Image" src="https://github.com/user-attachments/assets/e5f9fc79-82d2-418f-81b4-4eaea5fd39a7" />

2. Click the right sidebar button to open the right sidebar.

<img width="1774" height="1001" alt="Image" src="https://github.com/user-attachments/assets/8da73113-90ac-4312-a5b4-817f793790b0" />

3. Click to collapse the right sidebar, and the title bar returns to normal.

<img width="1774" height="1001" alt="Image" src="https://github.com/user-attachments/assets/521c407e-fa96-4ba4-9ada-d762c4082071" />

### Additional context (optional)

```shell

```

NevilleQingNY:
Thanks for the detailed report and screenshots — they made this one much easier to pin down. 🙏

**Fix (#3942):** the deep-link jump now scrolls *only* the comment list and never the outer page, so the title bar can no longer be pushed away. It'll ship in the next desktop release.

PR #6914, opened 2026-08-13
[MUL-6140] fix: scroll issue thread after posting

### LINKED ISSUE #6983: Newly posted comments are not scrolled into view
After posting a comment or reply in an issue thread, the viewport stays at the current scroll position rather than scrolling to show the newly posted message. The user has to manually scroll down to see their own comment.

https://github.com/user-attachments/assets/ecda7064-ff42-4719-9e5d-c2b1142e33e1

https://github.com/user-attachments/assets/c0cb8f38-4d16-44a1-a81a-c1cb8a941fbc

HomyeeKing:
close #6983 

## Summary
- scroll the issue timeline container to its latest height after a top-level comment is accepted
- scope the behavior to the local send-success path so realtime updates do not interrupt reading
- cover the behavior with an issue-detail regression test

## Verification
- pnpm --filter @multica/views exec vitest run issues/components/issue-detail.test.tsx
- pnpm --filter @multica/views typecheck
- git diff --check

https://github.com/user-attachments/assets/602d547b-67a7-4cb1-b214-912b56ff7eb6

https://github.com/user-attachments/assets/b50f4cb0-99dc-4651-86da-a584433e9fe0

https://github.com/user-attachments/assets/e9f3a259-2c49-465a-b2f8-4150513b1301

https://github.com/user-attachments/assets/d8cf57e6-7859-4ea4-96ba-256c2184b9a0

### DIFF
--- a/packages/views/issues/components/comment-card.tsx
+++ b/packages/views/issues/components/comment-card.tsx
@@ -107,6 +107,7 @@ interface CommentCardProps {
    */
   canModerate?: boolean;
   onReply: (parentId: string, content: string, attachmentIds?: string[], suppressAgentIds?: string[]) => Promise<boolean>;
+  onReplyAccepted?: (target: HTMLElement) => void;
   onEdit: (commentId: string, content: string, attachmentIds: string[], suppressAgentIds?: string[]) => Promise<void>;
   onDelete: (commentId: string) => void;
   onToggleReaction: (commentId: string, emoji: string) => void;
@@ -751,6 +752,7 @@ function CommentCardImpl({
   currentUserId,
   canModerate = false,
   onReply,
+  onReplyAccepted,
   onEdit,
   onDelete,
   onToggleReaction,
@@ -1138,6 +1140,7 @@ function CommentCardImpl({
                   avatarId={currentUserId ?? ""}
                   draftKey={`reply:${issueId}:${entry.id}`}
                   onSubmit={(content, attachmentIds, suppressAgentIds) => onReply(entry.id, content, attachmentIds, suppressAgentIds)}
+                  onAccepted={onReplyAccepted}
                 />
               </div>
             </>
--- a/packages/views/issues/components/comment-card.tsx
+++ b/packages/views/issues/components/comment-card.tsx
@@ -106,8 +106,8 @@ interface CommentCardProps {
    * `CommentRow` has to rerun the rule per row.
    */
   canModerate?: boolean;
-  onReply: (parentId: string, content: string, attachmentIds?: string[], suppressAgentIds?: string[]) => Promise<boolean>;
-  onReplyAccepted?: (target: HTMLElement) => void;
+  onReply: (parentId: string, content: string, attachmentIds?: string[], suppressAgentIds?: string[]) => Promise<string | boolean>;
+  onReplyAccepted?: (commentId: string) => void;
   onEdit: (commentId: string, content: string, attachmentIds: string[], suppressAgentIds?: string[]) => Promise<void>;
   onDelete: (commentId: string) => void;
   onToggleReaction: (commentId: string, emoji: string) => void;
--- a/packages/views/issues/components/comment-composers.test.tsx
+++ b/packages/views/issues/components/comment-composers.test.tsx
@@ -218,10 +218,12 @@ function renderCommentInput(onSubmit = vi.fn().mockResolvedValue(true)) {
 
 function renderReplyInput({
   onSubmit = vi.fn().mockResolvedValue(true),
+  onAccepted,
   size = "sm",
   draftKey,
 }: {
   onSubmit?: (content: string, attachmentIds?: string[], suppressAgentIds?: string[]) => Promise<boolean>;
+  onAccepted?: (target: HTMLElement) => void;
   size?: "sm" | "default";
   draftKey?: `reply:${string}:${string}`;
 } = {}) {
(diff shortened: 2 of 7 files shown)


HomyeeKing:
@multica-eve review it

Bohan-J (changes requested):
Thanks for working through this — the product direction is right, and the default browsing path looks good. Scoping the scroll to the sender's own successful post (rather than to every realtime update) is the correct call, and mapping a reply back to its root row so Virtuoso stays the sole scroll controller is a sensible fit for the virtualized timeline.

One blocking issue before merge.

### The posting scroll never runs when the timeline renders flat

`packages/views/issues/components/issue-detail.tsx:1546` sets `isFlatTimeline` when either `highlightCommentId` is set (comment deep link) or `find.open` is true (in-page Cmd/Ctrl+F). In both cases the timeline renders as a plain `items.map(...)` with no `<Virtuoso>` mounted, so `virtuosoRef.current` is `null`.

The new posting effect guards both jumps on that same flag:

- `:1562` and `:1566` — both `scrollToIndex` calls are skipped when `isFlatTimeline` is true
- `:1569` — `setPendingPostedCommentId(null)` still runs unconditionally

So on those two paths the request is consumed without ever moving the view, which leaves exactly the pre-PR behavior this change is meant to fix.

The deep-link case is the one I'd weigh most: `packages/views/inbox/components/inbox-page.tsx:554` renders the full `IssueDetail` — comment composer included — with `highlightCommentId` taken from the selected notification. Replying from the Inbox after clicking a comment notification is a very common way to answer a discussion, and it is precisely the path that doesn't scroll.

**Repro:** open a comment notification in the Inbox, post a top-level comment, and the view stays where it is.

I confirmed this by running the PR's own new test with `renderIssueDetailWithHighlight("comment-1")` substituted for `renderIssueDetail()`, everything else identical. Instrumenting the effect shows it runs to completion with a perfectly valid target and is stopped only by the guard:

```
PROBE effect ran: { pendingPostedCommentId: 'comment-new', rootId: 'comment-new', index: 2, isFlatTimeline: true }
PROBE flat-mode scroll calls: { scrollToIndex: 0, scrollIntoView: 0 }
```

`index: 2` is the same index the passing Virtuoso test asserts, and the composer's `onAccepted` fires normally — the row is found, the effect just has nothing wired to move it. The virtualized path gets 2 `scrollToIndex` calls; the flat path gets zero scrolls of any kind.

### Suggested fix

`jumpToThread` at `:1586` already solves this exact problem for minimap jumps in flat mode — it drives `scrollContainerEl.scrollTop` directly, with a deliberate note to never use native `scrollIntoView` because that also scrolls the desktop shell (#3929). Reusing that approach in the flat branch, with `end`-style alignment instead of the `start`/`-16` offset, would keep both modes consistent and avoid reintroducing that bug.

### Test coverage

`packages/views/issues/components/issue-detail.test.tsx:924` renders through the default `renderIssueDetail()`, so it only exercises the Virtuoso branch — which is why CI is green despite the above. Worth adding a regression case for each flat entry point: one with `highlightCommentId` (the helper at `:596` already exists) and one with the find bar open.

### Verification I ran on `e830ab85b`

- `pnpm --filter @multica/views exec vitest run issues/components/issue-detail.test.tsx issues/components/comment-composers.test.tsx issues/hooks/use-issue-timeline.test.tsx` — 110 passed
- CI on this head is fully green

Nothing else blocking from my side — the composer wiring, the reply-to-root mapping, and the decision not to scroll on realtime updates all look right. Happy to re-review once the flat path is covered.

HomyeeKing:
@Bohan-J  Flat mode is supported, please help review

Bohan-J (approved):
Thanks for sticking with this one, @HomyeeKing — and for the care you put into the follow-up.

The flat-timeline gap from the last round is properly fixed rather than worked around. Splitting the two modes into separate effects is the right call: the virtualized effect now bails on `isFlatTimeline` instead of pretending to handle it and then clearing the pending id, and the flat effect drives `scrollContainerEl.scrollTop` directly — reusing the existing `jumpToThread` approach, which keeps the native `scrollIntoView` desktop-shell bug (#3929) from creeping back in. Keeping `scrollContainerEl` out of the virtualized effect's deps so a container re-measure can't cancel an in-flight scroll is a nice touch.

I verified the `offset` direction against the actual `react-virtuoso@4.18.7` build rather than assuming it — `offset` is added to the computed target (`St += Q`), which matches this repo's own `align: "start", offset: -16` usage for a top gap. So `align: "end"` with a positive composer height lifts the row clear of the sticky composer, as intended. The flat branch's `scrollTop + (e.bottom - c.bottom) + composerHeight` is the same alignment expressed directly, and the rAF stabilization loop converges and is frame-capped.

Verified locally on `e52ad8da6`: 110 targeted tests pass, `@multica/views` typecheck clean, and CI is fully green.

One non-blocking nit for whenever you or someone else is next in this file: `composerHeight` is measured and applied unconditionally, but `useStickyComposer()` returns false on mobile and when the user turns the preference off — in that case the composer sits in normal flow and doesn't overlap the timeline, so the extra offset scrolls slightly further than needed. It's harmless in practice (on a phone it just brings the composer into view under your new comment), so it's not worth another round here.

Merging now. The regression tests for the two flat entry points (`highlightCommentId` and `find.open`) are worth adding, and we'll track that separately rather than hold this up — the change is a strict improvement over the current behavior, where posting scrolled nowhere at all. Thanks again.

PR #6915, opened 2026-08-13
MUL-6141: feat(runtimes): price Qwen/Kimi/Ark models and strip ':' provider prefixes

Frankli9986:
## What

Local BYO runtimes (Claude Code, Codex, Hermes, Kimi, Qwen Code, opencode) report token usage but never a provider-reported cost, so the dashboard prices them from the rate tables. Two gaps kept those costs at $0 for the Qwen / Kimi models:

1. **Missing rows** in both pricing tables (`packages/views/runtimes/utils.ts` and the server registry `server/internal/metrics/pricing.go`) for Qwen 3.6/3.7/3.8 and Kimi K3.
2. **A resolver bug**: `canonicalCandidates` stripped only `/` routing prefixes, so Hermes-style `provider:model` ids (`alibaba-coding-plan:qwen3.8-max`, `custom:ark-code-latest`, `kimi-coding:kimi-k3`, `opencode-go:deepseek-v4-flash`) never resolved and stayed uncosted — `opencode-go:deepseek-v4-flash` was silently uncosted despite having a row. The server-side alias regexes already handled `:`; only the frontend resolver was broken.

## Changes

- **`packages/views/runtimes/utils.ts`** — add USD per-MTok rows for `qwen3.6-flash`, `qwen3.7-plus`, `qwen3.8-max` (pay-as-you-go), `qwen3.8-max-preview` (Alibaba Token/Coding Plan subscription, priced 0 like the free GLM tiers), and `kimi-k3` / `kimi/k3`. Fix `stripProvider` to strip `:` prefixes too and resolve nested routing prefixes (`custom:anthropic/claude-opus-4.7`) iteratively, with a comment.
- **`server/internal/metrics/pricing.go`** — mirror the same rows in the server registry (the file's own comment requires the two tables to stay in sync). All Qwen/Kimi alias rules are anchored: `qwen3.8-max-preview` / `kimi-k3-1` never borrow the GA tier; an optional complete bracket tag `[…]` (at least one character inside, matching the frontend's `\[[^\]]+\]$` tag stripping) is admitted so `qwen3.8-max-preview[1m]` still resolves while empty tags (`qwen3.7-plus[]`) and unknown suffixes stay unmapped on both sides; `kimi-code/k3` resolves via the `/k3$` form. `ark-code-latest` is deliberately left unmapped — it is a Volcengine console-switchable rolling alias across model families, not a stable model identity, and the daemon reports the alias, never the resolved concrete model.
- **Tests** — Go cases pinning the new rows plus no-false-borrowing guards (unknown suffixes, malformed/empty bracket tags, `ark-code-latest` all stay unmapped); frontend regression tests for `provider:model` stripping, nested routing prefixes, and the new rows.

## Sources

- Qwen: alibabacloud.com model-pricing sheet, International tier, `0<Token≤256K` (`$1.20/$4.80` above 256K), accessed 2026-08-17. Cache rates: Explicit Cache Creation 1.25× input, Explicit Cache Read 10% of input (`qwen3.7-plus`: `$0.50` / `$0.04`; `qwen3.6-flash`: `$0.3125` / `$0.025`) from the qwencloud.com model pages.
- `qwen3.8-max`: qwencloud.com/models/qwen3.8-max — Input `$2`, Output `$6`, Implicit Cache `$0.25`, Explicit Cache Creation `$2.50`, Explicit Cache Read `$0.17` (same source models.dev cites).
- Kimi K3: models.dev `providers/moonshotai`, accessed 2026-08-13.

## Testing

- `go test ./internal/metrics/... -count=1` — pass (alias tests + no-false-borrowing guards incl. empty-tag negatives)
- `vitest run packages/views/runtimes/utils.test.ts` — 88 pass
- `git diff --check` — clean

## Notes

- `qwen3.8-max` is priced at the published pay-as-you-go rate ($2/$6) even though the reporting runtime reaches it through a subscription, so the dashboard shows absolute cost.
- `ark-code-latest` is a rolling Volcengine alias (users can switch its target in the console, including across model families), so it stays unmapped on both sides — same convention as other unstable identities.

### DIFF
--- a/packages/views/runtimes/utils.test.ts
+++ b/packages/views/runtimes/utils.test.ts
@@ -347,8 +347,8 @@ describe("estimateCost", () => {
       cache_read_tokens: 1_000_000,
       cache_write_tokens: 1_000_000,
     });
-    // 1M × $2 + 1M × $6 + 1M × $0.25 + 1M × $2.50 = $10.75.
-    expect(cost).toBeCloseTo(10.75, 5);
+    // 1M × $2 + 1M × $6 + 1M × $0.17 + 1M × $2.50 = $10.67.
+    expect(cost).toBeCloseTo(10.67, 5);
     // `ark-code-latest` is a rolling Volcengine alias, not a stable model
     // identity, so it is deliberately unmapped after the prefix strip.
     expect(isModelPriced("custom:ark-code-latest", "hermes")).toBe(false);
--- a/packages/views/runtimes/utils.test.ts
+++ b/packages/views/runtimes/utils.test.ts
@@ -334,6 +334,73 @@ describe("estimateCost", () => {
     ).toBe(0);
   });
 
+  it("strips `provider:model` routing prefixes (Hermes custom providers) before resolving", () => {
+    // Regression: canonicalCandidates only stripped `/` prefixes, so Hermes
+    // ids like `alibaba-coding-plan:qwen3.8-max` and `custom:ark-code-latest`
+    // never resolved and stayed uncosted. `:` must be stripped like `/`.
+    const cost = estimateCost({
+      ...zeroUsage,
+      provider: "hermes",
+      model: "alibaba-coding-plan:qwen3.8-max",
+      input_tokens: 1_000_000,
+      output_tokens: 1_000_000,
+      cache_read_tokens: 1_000_000,
+      cache_write_tokens: 1_000_000,
+    });
+    // 1M × $2 + 1M × $6 + 1M × $0.25 + 1M × $2.50 = $10.75.
+    expect(cost).toBeCloseTo(10.75, 5);
+    expect(isModelPriced("custom:ark-code-latest", "hermes")).toBe(true);
+    expect(isModelPriced("kimi-coding:kimi-k3", "hermes")).toBe(true);
+  });
+
+  it("prices the Qwen / Kimi / Ark models added from models.dev", () => {
+    expect(
+      estimateCost({
+        ...zeroUsage,
+        model: "qwen3.7-plus",
+        input_tokens: 1_000_000,
+        output_tokens: 1_000_000,
+      }),
+    ).toBeCloseTo(3.5, 5); // $0.50 + $3.00
+    expect(
+      estimateCost({
+        ...zeroUsage,
+        model: "kimi-code/k3",
+        provider: "kimi",
+        input_tokens: 1_000_000,
+        output_tokens: 1_000_000,
+      }),
+    ).toBeCloseTo(18, 5); // kimi/k3 → $3 + $15
+    expect(
+      estimateCost({
+        ...zeroUsage,
+        model: "ark-code-latest",
+        input_tokens: 1_000_000,
+        output_tokens: 1_000_000,
+      }),
+    ).toBeCloseTo(5.04, 5); // $0.84 + $4.20
+  });
+
+  it("keeps subscription-only and preview SKUs distinct from their GA siblings", () => {
+    // qwen3.8-max-preview is subscription-priced at 0; it must NOT inherit
+    // qwen3.8-max's $2/$6 tier, and `qwen3.8-max-preview[1m]` must resolve
+    // to the preview row after the context-tag strip.
+    expect(
+      estimateCost({
+        ...zeroUsage,
+        model: "qwen3.8-max-preview[1m]",
+        input_tokens: 1_000_000,
+      }),
+    ).toBe(0);
+    expect(
+      estimateCost({
+        ...zeroUsage,
+        model: "qwen3.8-max",
+        input_tokens: 1_000_000,
+      }),
+    ).toBeCloseTo(2, 5);
+  });
+
   it("returns 0 for a genuinely unknown model so the UI can flag it", () => {
     expect(
       estimateCost({
--- a/packages/views/runtimes/utils.test.ts
+++ b/packages/views/runtimes/utils.test.ts
@@ -349,11 +349,13 @@ describe("estimateCost", () => {
     });
     // 1M × $2 + 1M × $6 + 1M × $0.25 + 1M × $2.50 = $10.75.
     expect(cost).toBeCloseTo(10.75, 5);
-    expect(isModelPriced("custom:ark-code-latest", "hermes")).toBe(true);
+    // `ark-code-latest` is a rolling Volcengine alias, not a stable model
+    // identity, so it is deliberately unmapped after the prefix strip.
+    expect(isModelPriced("custom:ark-code-latest", "hermes")).toBe(false);
     expect(isModelPriced("kimi-coding:kimi-k3", "hermes")).toBe(true);
   });
 
-  it("prices the Qwen / Kimi / Ark models added from models.dev", () => {
+  it("prices the Qwen / Kimi models added from models.dev", () => {
     expect(
       estimateCost({
         ...zeroUsage,
@@ -361,7 +363,7 @@ describe("estimateCost", () => {
         input_tokens: 1_000_000,
         output_tokens: 1_000_000,
       }),
-    ).toBeCloseTo(3.5, 5); // $0.50 + $3.00
+    ).toBeCloseTo(2.0, 5); // $0.40 + $1.60 (International ≤256K tier)
     expect(
       estimateCost({
         ...zeroUsage,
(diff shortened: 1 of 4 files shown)


multica-eve (changes requested):
The product direction is valuable, but the current pricing data and alias resolution can silently report incorrect costs. Please address these blockers before re-requesting review:

1. **The Qwen rates do not match the cited official pricing.** `packages/views/runtimes/utils.ts:304-306` and `server/internal/metrics/pricing.go:84-86` use `$0.50/$3.00` for `qwen3.7-plus`, while Alibaba's international list price is `$0.40/$1.60` for requests up to 256K tokens and `$1.20/$4.80` above 256K. The `qwen3.6-flash` input/output and cache rates also differ from Alibaba's official model page. Please choose and document a specific region/context tier, update both tables and tests, and provide a primary source for the claimed `$2/$6` `qwen3.8-max` pay-as-you-go rate.

2. **`ark-code-latest` is not a stable model identity.** `packages/views/runtimes/utils.ts:309-316` and `server/internal/metrics/pricing.go:92-99` hard-code the Doubao-Seed-Evolving rate, but Volcengine documents that users can switch the target behind this alias in the console, including to other model families. Please leave the alias unmapped or record and price the resolved concrete model instead.

3. **The backend aliases accept unknown suffixes that the frontend rejects.** The unanchored rules at `server/internal/metrics/pricing.go:152-155,161` price values such as `qwen3.7-plus-extra` and `qwen3.8-max-preview-extra`, while the frontend exact matcher leaves them unmapped. Anchor the rules to explicitly supported forms and add negative tests on both sides.

4. **Mixed routing prefixes are only stripped once.** `packages/views/runtimes/utils.ts:494-503` claims mixed forms resolve, but `custom:anthropic/claude-opus-4.7` stops at `anthropic/claude-opus-4.7` and still misses the frontend table while the backend matches it. Either strip valid routing layers iteratively with a regression test or narrow the documented behavior.

The Kimi K3 rates and the basic `provider:model` regression coverage look good. Local Go metrics tests and `git diff --check` pass; the existing GitHub CI is green, but those checks currently lock in the incorrect pricing assumptions above.

Frankli9986:
Thanks for the thorough review — all four blockers are addressed in `a9bb5dfde`. Point-by-point:

**1. Qwen rates now match the official International ≤256K list prices**
- `qwen3.7-plus`: `$0.40 / $1.60` (was `$0.50 / $3.00`) — alibabacloud.com model-pricing, International tier, `0<Token≤256K` (`$1.20 / $4.80` above 256K). Cache: Explicit Cache Creation `$0.50` (1.25× input), Explicit Cache Read `$0.04` (10% of input), both from the qwencloud.com/qwen3.7-plus page.
- `qwen3.6-flash`: `$0.25 / $1.50` (was `$0.1875 / $1.125`) — same official sheet, International ≤256K. Cache: Creation `$0.3125`, Read `$0.025` (qwencloud.com/qwen3.6-flash page).
- `qwen3.8-max` `$2 / $6` retained and now cited to its primary source: the **qwencloud.com/models/qwen3.8-max page** (Input `$2`, Output `$6`, Implicit Cache `$0.25`, Explicit Cache Creation `$2.5`, Read `$0.17`) — the same source models.dev cites; it also appears on the alibabacloud.com pricing sheet (Singapore region).
- Both tables (`packages/views/runtimes/utils.ts` and `server/internal/metrics/pricing.go`) updated together, with the region/tier choice documented in the comments.

**2. `ark-code-latest` is no longer mapped**
The Volcengine console lets users switch what this alias resolves to (across model families), so it is not a stable model identity and the daemon reports the alias, never the resolved concrete model. Per your suggestion it is now left unmapped on both sides (same convention as `grok-composer-*`), with negative tests asserting it stays unpriced.

**3. Backend alias rules are now anchored**
`qwen3.7-plus`, `qwen3.6-flash`, `qwen3.8-max`, and `qwen3.8-max-preview` rules use `(^|/|:)…$` anchoring; `($|\[)` admits the bracketed `[1m]` context suffix so `qwen3.8-max-preview[1m]` still resolves while `qwen3.7-plus-extra` / `qwen3.8-max-preview-extra` stay unmapped. Negative tests added on **both** sides (backend `TestPriceForModelAliasNoFalseBorrowing`, frontend anchored-variants case).

**4. Nested routing prefixes are now stripped iteratively**
`custom:anthropic/claude-opus-4.7` now resolves to the Opus tier on the frontend (backends already matched by substring). Regression tests cover the mixed form on both sides; the existing `kimi-code/k3`, `cursor/composer-*`, and provider-qualified cases still pass.

**Verification**
- `pnpm vitest run packages/views/runtimes/utils.test.ts` → **88 pass** (86 pre-existing + 2 new)
- `go test ./internal/metrics/... -count=1` → **pass** (alias tests + negative guards)
- `git diff --check` → clean

Kimi K3 (`$3/$15`) is untouched as you noted it looked good. Re-requesting review.

Frankli9986:
@multica-eve 已按评审意见全部修改并补充官方定价来源（见上一条逐条回复），请重新 review。

multica-eve (changes requested):
Thanks for the follow-up. The updated head fixes the stable Qwen input/output rates, leaves `ark-code-latest` unmapped, and makes nested routing-prefix resolution consistent. Two billing correctness blockers remain:

1. **[must-fix] `qwen3.8-max` uses the implicit-cache rate for explicit cache reads.** The comments at `packages/views/runtimes/utils.ts:301-304` already record the official rates as `$0.25` for implicit cache and `$0.17` for explicit cache read, but the row at line 313 sets `cacheRead: 0.25`; `server/internal/metrics/pricing.go:91` mirrors the same value. Qwen Code emits its provider-reported `cache_read_input_tokens`, which Multica maps directly to `CacheReadTokens` (`server/pkg/agent/qwen.go:210-214,335-337`), and its DashScope provider adds explicit `cache_control` markers. Please use the official explicit-cache-read rate (`$0.17/M`) in both tables and update the assertions. At the current value, one million cache-read tokens are overestimated by `$0.08` (about 47%). Source: https://www.qwencloud.com/models/qwen3.8-max

2. **[must-fix] The backend Qwen bracket-suffix aliases are still not fully anchored.** `server/internal/metrics/pricing.go:160-161` ends the rules with `($|\[)`, so any string beginning with `[` is accepted regardless of what follows. I confirmed that `qwen3.8-max[`, `qwen3.8-max[1m]-extra`, and `qwen3.8-max-preview[1m]-extra` all resolve in the backend, while the frontend leaves them unmapped. Match a complete optional bracket tag through the end of the string and add these malformed/extra-suffix cases to the negative tests.

Non-blocking: the PR description still says `ark-code-latest` is priced as Doubao and describes the pre-fix resolver/test set; please refresh it before merge so it reflects the current change.

Validation: `go test ./internal/metrics/...` and `git diff --check` pass. All 13 GitHub checks are green on `a9bb5dfde`; local Vitest could not be rerun because this checkout has no frontend dependencies installed.

multica-eve (changes requested):
Thanks for addressing the previous cache-rate and suffix-anchoring blockers. One cross-layer pricing consistency issue remains on the current head:

**[must-fix] Reject empty Qwen context tags in the backend, matching the frontend.** The four patterns at `server/internal/metrics/pricing.go:158-161` use `*` inside the optional bracket tag, so `qwen3.7-plus[]`, `qwen3.6-flash[]`, `qwen3.8-max[]`, and `qwen3.8-max-preview[]` all resolve and receive a price in the backend. The frontend matcher at `packages/views/runtimes/utils.ts:528` uses `+`, so the same malformed model IDs remain unmapped there. Please require at least one character inside the backend tag (use `+`) and add `[]` negative cases to the Go regression test.

Non-blocking: the PR description still says `ark-code-latest` is priced as Doubao, which no longer matches the implementation; please refresh it before merge.

Validation on `b28f628aa`: `go test ./internal/metrics/...` and `git diff --check` pass; a focused regression test reproduces all four empty-tag mismatches. All 15 GitHub checks have no failures.

Frankli9986:
Thanks for the follow-up — the empty-tag blocker is fixed in `e4fe3e3fc`:

**[must-fix] Empty Qwen context tags are now rejected in the backend.**
The four Qwen alias rules at `server/internal/metrics/pricing.go:158-163` now use `(\[[^\]]+\])?$` (`+` instead of `*`), so `qwen3.7-plus[]`, `qwen3.6-flash[]`, `qwen3.8-max[]`, and `qwen3.8-max-preview[]` all stay unmapped — matching the frontend's `\[[^\]]+\]$` tag stripping at `packages/views/runtimes/utils.ts:528`. All four `[]` cases are added to the must-not-resolve list in `TestPriceForModelAliasNoFalseBorrowing` (`pricing_test.go:322-327`), alongside the existing malformed/extra-suffix negatives. Non-empty tags still resolve as before: `qwen3.8-max[1m]`, `qwen3.8-max-preview[1m]`, `qwen3.8-max-preview[context]` all pass.

**Non-blocking: PR description refreshed.** It no longer claims `ark-code-latest` is priced as Doubao; it now documents that the alias is deliberately unmapped on both sides (console-switchable rolling alias, daemon only reports the alias), and the sources/testing sections reflect the current rates and test counts.

**Verification on `e4fe3e3fc`:**
- `go test ./internal/metrics/... -count=1` → **pass** (including the four new `[]` negative cases)
- `git diff --check` → clean
- No frontend changes this round; the frontend already uses `+` so both sides agree.

@multica-eve please re-review.

PR #6916, opened 2026-08-13
[MUL-6142] WS-10: clarify Codex config inheritance for agents

zhuchangmin:
## Summary

- add the missing Codex native-config inheritance matrix to `multica-creating-agents`
- distinguish copied/linked native configuration from non-inherited GUI goals, plan preferences, state, sandbox settings, and global session history
- document first-class agent fields and Multica-owned durable state as the supported alternatives
- add source-traced evidence and contract-test coverage for the matrix

## Validation

- `go test ./internal/service -run 'TestCreatingAgentsSkillCoversAgentCreationContracts|TestBuiltinSkillsConformToTemplate|TestBuiltinSkillsFrontmatterIsStrictYAML'`
- `go test ./internal/daemon/execenv -run 'TestPrepareCodexHomeSeedsFromShared|TestPrepareCodexSessionsDir_FreshCreatesEmptyLocalDir|TestPrepareCodexSessionsDir_MigratesLegacySymlinkNoResume|TestPrepareCodexSessionsDir_MigrateWithResumeRoutesThroughStore|TestResetCodexSessionState_OnlyRemovesSessionDerived|TestEnsureCodexMemoryConfig|TestCodexSandboxPolicy'`
- `git diff --check`

Related WS-10

### DIFF
--- a/server/internal/service/builtin_skills/multica-creating-agents/SKILL.md
+++ b/server/internal/service/builtin_skills/multica-creating-agents/SKILL.md
@@ -51,6 +51,40 @@ Two distinct text fields, often confused:
   Persona, responsibilities, boundaries, output and escalation rules go here,
   not in `description`.
 
+## Native config inheritance per provider
+
+An agent definition and a provider's native home are different configuration
+layers. For Codex, the daemon creates an isolated, per-task `CODEX_HOME` and
+seeds only an explicit subset of the machine's shared Codex home. A setting
+visible in Codex Desktop is therefore inherited only when it lives in one of
+the seeded native files or directories and is not replaced by a daemon-managed
+policy.
+
+| Codex Desktop / native data | Multica task behavior | Use instead when not inherited |
+|---|---|---|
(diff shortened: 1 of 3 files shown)


multica-eve (changes requested):
Thanks for the contribution. We do not want to merge this approach. The agent-creation surface should expose a small, stable contract: persisted agent fields and assigned capabilities are portable, while runtime-host native settings are resolved per task and are not part of the agent definition.

This PR turns volatile daemon implementation details—authentication linkage, plugin cache layout, goal databases, OS-specific sandbox defaults, and session storage—into long-lived creation guidance, increasing the built-in skill by roughly 25%. It also overstates `auth.json` linkage on Windows, where file-link creation can fall back to a copy, and the new source-map entries do not follow the existing `file:line` evidence contract.

We are closing this as a product-level rejection. If this user confusion needs documentation, we would prefer a separate, much smaller change that states only the stable portability boundary and leaves runtime internals in troubleshooting documentation.

PR #6918, opened 2026-08-13
fix(desktop): stop the tab flares printing dark squares under the active tab (MUL-6143)

tim-smart:
On desktop in dark mode, a solid dark square appeared under each bottom corner of the active tab after the keyline-unification fix (MUL-6106 / #6874).

## Cause

That fix backed the flare gradient with an opaque `var(--app-shell)` layer, on the assumption that the tab strip sits on `bg-app-shell`. The strip actually sits on the `SidebarProvider` wrapper, whose fill is conditional (`packages/ui/components/ui/sidebar.tsx`): `has-data-[variant=inset]:bg-sidebar` while the inset-variant app sidebar is mounted (≥ `lg`), and the layout's `bg-app-shell` otherwise — below `lg` the sidebar renders as a sheet, so no `[data-variant=inset]` element exists. In dark mode `--app-shell` (≈ rgb 12) is markedly darker than `--sidebar` (≈ rgb 24), so a backing pinned to either token prints a visible square on the other side of the breakpoint.

## Fix

Take the backing out of the gradient, and off any token. The wrapper publishes its own fill as a custom property under the same `:has()` condition that paints it, and exports the class that reads it:

```
// sidebar.tsx — wrapper: paint and variable, one condition
has-data-[variant=inset]:bg-sidebar has-data-[variant=inset]:[--sidebar-wrapper-fill:var(--sidebar)]

// desktop-layout.tsx — consumer owns the non-inset half, next to the class that produces it
bg-app-shell [--sidebar-wrapper-fill:var(--app-shell)]

// tab-bar.tsx — flare
SIDEBAR_WRAPPER_FILL_CLASS   // "bg-(--sidebar-wrapper-fill)"
```

The gradient stays inline as `background-image`; the class layer supplies the opaque backing. The flare now always composites over the colour actually behind it, so the notch is invisible and the arc reads the same tone as the cap border and the content card's ring at every width. The keyline-stacking problem MUL-6106 solved stays solved, since the backing is still opaque. And because the condition lives only on the wrapper, nothing downstream can drift from it — an earlier revision of this PR mirrored the condition at the flare instead, which would have re-broken silently on any rename in `sidebar.tsx`.

## Verification

A harness rendered the strip, tab chrome, and content card from the gradient template and token values read out of `tab-bar.tsx` and `tokens.css`, replicating the wrapper's `:has()` cascade, across {dark, light} × {inset sidebar mounted, compact/sheet}. Sampled pixels (single channel), after the fix:

| scenario | flare notch | backdrop | arc | ring |
| --- | --- | --- | --- | --- |
| dark, large | 24 | 24 | 47 | 46 |
| dark, compact | 12 | 12 | 36 | 35 |
| light, both | matches | matches | 228 | 228 |

Against `main`, the case this changes is **dark at large widths** — `main` pins the backing to `--app-shell` (12) while the wrapper there is painted `--sidebar` (24), which is the square in the bug report. Dark compact was already correct on `main` (backing and backdrop are both `--app-shell`), and light mode never had the bug at all, since `--app-shell` and `--sidebar` share one value there. Both of those render byte-identical to `main` (max channel diff 0).

The compiled selectors come from the real `globals.css` theme: the inset declaration out-specifies the consumer's, 0,2,0 vs 0,1,0 — the same mechanism the wrapper's own `bg-sidebar` uses to beat the consumer's background class.

`pnpm --filter @multica/desktop test`, `pnpm --filter @multica/views test`, `typecheck` and `lint` pass. The contract is covered from both ends: the desktop test asserts the flares name no token and that the gradient carries no pinned colour layer, and `packages/views/layout/sidebar-wrapper-fill.test.tsx` asserts the wrapper keeps paint and variable gated on one condition, and that `tailwind-merge` leaves the consumer's pair intact.

---

Maintainer follow-up (`de64a7d`) applied the review points from MUL-6143: moved the backdrop knowledge into the wrapper as above, restored the gradient-geometry rationale dropped in `3aafa0e7` (the 1.2/0.8/0.2px stop offsets and the 0.4px anti-aliasing spread had no other explanation in the file), and rewrote this Verification section against `main` — it previously read as relative to an intermediate commit on this branch.

### DIFF
--- a/apps/desktop/src/renderer/src/components/desktop-layout.tsx
+++ b/apps/desktop/src/renderer/src/components/desktop-layout.tsx
@@ -238,7 +238,12 @@ export function DesktopShell() {
       <WorkspaceSlugProvider slug={slug}>
         <DesktopInboxBridge />
         <div className="flex h-screen bg-app-shell">
-          <SidebarProvider className="flex-1 bg-app-shell">
+          {/* bg-app-shell is the wrapper's non-inset fill, so it also owns the
+              non-inset half of --sidebar-wrapper-fill. sidebar.tsx supplies the
+              inset half of both. Anything that has to paint an opaque layer
+              over this wrapper (the tab flares) reads the variable rather than
+              re-deriving which of the two is in play. */}
+          <SidebarProvider className="flex-1 bg-app-shell [--sidebar-wrapper-fill:var(--app-shell)]">
             {slug && <GlobalShortcuts />}
             {slug && <WindowToolbar />}
             {slug && <AppSidebar topSlot={<SidebarTopSpacer />} searchSlot={<SearchTrigger />} />}
--- a/apps/desktop/src/renderer/src/components/tab-bar.test.tsx
+++ b/apps/desktop/src/renderer/src/components/tab-bar.test.tsx
@@ -261,7 +261,10 @@ describe("TabBar merged-tab chrome", () => {
     );
   });
 
-  // The opaque sidebar backing keeps the flare and card keylines from stacking.
+  // The opaque backing keeps the flare and card keylines from stacking, and
+  // must equal the strip's backdrop on both sides of the compact breakpoint:
+  // the wrapper paints bg-sidebar only while an inset sidebar is mounted, so
+  // the flare mirrors that condition instead of pinning one token.
   it("paints each flare on an opaque layer matching the strip backdrop", () => {
     const { getByLabelText } = render(<TabBar />);
     const flares = getByLabelText("Issues")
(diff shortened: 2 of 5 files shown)


tim-smart:
Sorry for the churn 😅 

comparison-dark

Bohan-J:
Thanks @tim-smart — and no need to apologise for the churn. The diagnosis here was the valuable part: "the strip sits on the sidebar wrapper, whose fill is conditional" is exactly right, and it's not something you'd catch without going looking. The write-up made it easy to verify.

I pushed one follow-up commit (`de64a7d`) rather than sending you another round. Your version mirrored the wrapper's `has-data-[variant=inset]` condition at the flare, which is correct today but left the same condition living in two packages with nothing holding them together — a rename in `sidebar.tsx` would have reprinted the squares silently. The wrapper now publishes its own fill as `--sidebar-wrapper-fill` under the same `:has()` gate that paints it and exports `SIDEBAR_WRAPPER_FILL_CLASS`, so the flare just asks for "whatever the wrapper is". Same rendered result, condition in one place. I also restored the gradient-geometry comment that got trimmed along the way (the 1.2/0.8/0.2px offsets had no other explanation in the file) and rewrote the Verification section against `main`.

Merging. Thanks for chasing this all the way down.

PR #6921, opened 2026-08-13
MUL-6145: WS-18: reconcile stalled stage children

zhuchangmin:
## Summary

- reconcile direct task terminal callbacks and periodic stale-child scans through one idempotent path
- wake the parent coordinator only when a non-terminal child still holds the current stage barrier, preserving `in_review` and `blocked` semantics
- persist recovery deduplication by source task, add configurable scan controls, and document the behavior

## Validation

- `DATABASE_URL=postgres://multica:multica@127.0.0.1:55432/multica?sslmode=disable go test ./internal/handler ./internal/service ./cmd/server -count=1`
- `go test ./internal/migrations -count=1`
- migration 313 up/down/up verified (the broad down command subsequently encountered pre-existing duplicate plugin identity data while attempting migration 312)

Closes WS-18


multica-eve (changes requested):
Before this change, a child item could stop running while still holding the current stage open, leaving the parent coordinator unaware of what to do next. With this change, the parent receives clear guidance to review, unblock, or rerun that child, without automatically completing or cancelling it or advancing later stages.

**Product assessment: ✅ Accept.** This addresses a real unattended-workflow stall, and the “guide, but do not decide for the user” boundary is appropriate.

The implementation has four must-fix issues:

1. A delayed automatic retry is treated as “no active run.” `ReconcileIssueAfterTaskTerminal` uses `HasActiveTaskForIssue` (`server/internal/service/task.go:4796`), whose query excludes `deferred` tasks (`server/pkg/db/queries/agent.sql:1333`). The final `provider_network` retry is deliberately created as `deferred` (`server/internal/service/task.go:4294-4303`), but `FailTask` still invokes reconciliation (`server/internal/service/task.go:4232`). This can wake the parent with manual-rerun guidance while an automatic retry is already waiting. Please include delayed retries in the recovery gate, or skip parent recovery whenever any automatic retry was created, and add a regression test proving that a deferred retry creates no parent comment/task.

2. The bounded scan can starve actionable rows indefinitely. `ListStalledStageChildTasks` always orders by the oldest rows and applies a fixed limit (`server/pkg/db/queries/agent.sql:1367-1368`), while parent status, assignee type, and current-stage eligibility are checked only later (`server/internal/handler/issue_child_done.go:146-155`). Rejected rows are not marked or advanced, so 100 old later-stage/member-parent/parked-parent rows can be selected every 30 seconds forever, preventing every newer eligible row from being examined. Please filter to the truly actionable frontier in the query, or use a cursor/rotation scheme that advances across ticks and revisits rows when they can become actionable. Add a test with one full ineligible batch followed by an eligible row.

3. Comment deduplication can permanently lose the actual wake-up. The recovery path persists the comment and only then separately enqueues the parent task (`server/internal/handler/issue_child_done.go:175-199`), while the scanner permanently excludes a source task as soon as that comment exists (`server/pkg/db/queries/agent.sql:1361-1365`). A process exit or transient enqueue error between those steps leaves a durable comment but no parent wake, and future scans will never repair it. Please make “comment persisted” plus “wake durably queued” replay-safe through one atomic/durable state (for example, a transactionally written task or outbox), and test that an initial enqueue failure is recovered with exactly one eventual wake and no duplicate comment.

4. The migration prefix now conflicts with `main`. This branch adds `313_system_comment_source_task_unique.*.sql`, while `main` already contains `313_runtime_profile_add_dsh.*.sql`. A synthetic merge followed by `go test ./internal/migrations -run TestMigrationNumericPrefixesStayUniqueAfterLegacySet -count=1` fails with `migration prefix 313 is reused`. Please rebase and renumber this migration to the next available prefix (currently 332).

The status-specific recovery guidance, stage-frontier helper tests, documentation, and concurrent/no-FK index shape are good choices to keep.

**Recommendation: Request changes.** The product direction is right, but the issues above can cause false wakes, permanently missed wakes, scanner starvation, and a post-merge migration check failure.

PR #6922, opened 2026-08-13
fix(agents): let workspace admins select another member's private runtime

### LINKED ISSUE #6898: [Bug]: Workspace admins cannot select another member private runtime in the UI, while the API allows it
### Deployment type

Official App and self-hosted

### What happened?

The frontend and backend enforce different permissions for binding an agent to a private runtime owned by another workspace member.

The backend intentionally allows workspace owners/admins to use a private runtime:

```go
func canUseRuntimeForAgent(member db.Member, rt db.AgentRuntime) bool {
    if roleAllowed(member.Role, "owner", "admin") {
        return true
    }
    if rt.Visibility == "public" {
        return true
    }
    return rt.OwnerID.Valid && uuidToString(rt.OwnerID) == uuidToString(member.UserID)
}
```

However, the shared frontend predicate only allows the runtime owner or a public runtime:

```ts
export function isRuntimeUsableForUser(
  runtime: RuntimeDevice,
  currentUserId: string | null,
): boolean {
  if (!currentUserId) return true;
  if (runtime.owner_id === currentUserId) return true;
  return runtime.visibility === "public";
}
```

The frontend predicate has no workspace-role input, so a workspace owner/admin sees another member private runtime as locked and disabled even though the same create/update request succeeds through the API/CLI.

This affects at least:

- Manual agent creation: the runtime picker disables the private runtime and the form `draftReady` check rejects it.
- Agent runtime reassignment: the inspector runtime picker independently applies the same owner/public-only rule.
- Other create paths that reuse `isRuntimeUsableForUser`, including the builder and legacy create dialog.

This appears to be a regression against the contract introduced by #2419, which explicitly states that private runtimes are usable by their owner or workspace admins.

### Steps to reproduce

1. Create a workspace with two members:
   - Member A owns a runtime whose visibility is `private`.
   - Member B has workspace `owner` or `admin` role.
2. Sign in as Member B.
3. Open **Agents → Create Agent → Start from scratch**.
4. Switch the runtime filter to **All** and open Member A machine.
5. Observe that Member A private runtimes are marked private and disabled.
6. Send the equivalent agent-create request through the CLI/API as Member B, using the same `runtime_id`.
7. Observe that the server accepts the request.
8. The same mismatch can be reproduced when Member B edits an existing agent and tries to switch it to Member A private runtime.

### Expected behavior

The UI and API should enforce the same permission contract.

If workspace owners/admins are intended to retain the backend override, the frontend runtime-usability predicate should include the current workspace role and allow `owner`/`admin` consistently across create, builder, duplicate, and runtime-reassignment surfaces.

Alternatively, if private runtimes should be owner-only, the backend override and existing UI copy should be changed consistently. The current split behavior should not remain.

### Actual behavior

- UI: another member private runtime is disabled for workspace owners/admins.
- API/CLI: the same workspace owners/admins can bind an agent to that runtime successfully.

### Relevant source

- `packages/core/runtimes/access.ts`
- `packages/views/agents/create/use-create-agent-form.ts`
- `packages/views/agents/components/inspector/runtime-picker.tsx`
- `server/internal/handler/runtime.go`
- Related implementation: #2419

itsXuSt:
## What does this PR do?

Fixes the UI/API permission mismatch from #6898: workspace owners/admins could bind an agent to another member's **private** runtime through the API/CLI (the backend `canUseRuntimeForAgent` grants them an administrative override), but the frontend runtime picker disabled that same runtime as locked. The shared frontend predicate `isRuntimeUsableForUser` only knew about the runtime owner and the `public` flag — it had no workspace-role input — so the create, duplicate, builder, and runtime-reassignment surfaces all enforced a stricter rule than the server.

This threads the current member's workspace role into `isRuntimeUsableForUser` and applies the same `owner`/`admin` override the backend already grants, consistently across every surface that reuses the predicate.

## Related Issue

Closes #6898

## Type of Change

## Changes Made

- `packages/core/runtimes/access.ts` — `isRuntimeUsableForUser` now takes an optional `role` and allows `owner`/`admin` to use any runtime, mirroring the backend `canUseRuntimeForAgent` override. An omitted/`null` role falls back to the previous owner/public-only behaviour, so untouched callers are unaffected.
- `packages/views/agents/components/runtime-picker.tsx` (manual create picker) — derives the current member's role from the existing `members` + `currentUserId` props and passes it through seeding, the Mine/All filter, the per-row disabled state, and `computeFilteredRuntimes`.
- `packages/views/agents/components/inspector/runtime-picker.tsx` (runtime reassignment) — replaces the inline owner/public-only `isDisabled` with the shared predicate, role-aware, so admins can rebind onto another member's private runtime.
- `packages/views/agents/components/create-agent-dialog.tsx` (legacy create/duplicate dialog) — role-aware template-runtime prefill and `selectedRuntimeLocked` Create gate.
- `packages/views/agents/create/builder-conversation.tsx` (builder setup) — role-aware "has an online usable runtime" gate.
- `packages/views/agents/create/use-create-agent-form.ts` — role-aware `usableRuntimes` and `draftReady`.
- `packages/core/agents/draft.ts` / `packages/views/agents/create/use-duplicate-draft-seed.ts` / `manual-create-agent-page.tsx` — `buildDuplicateDraft` accepts the role so duplicating onto another member's private runtime keeps the per-runtime model/thinking/service-tier instead of falling back.

No backend change: the server already allowed owners/admins; this only brings the UI to the same contract.

## How to Test

1. In a workspace with two members, have Member A own a runtime whose visibility is `private`.
2. Sign in as Member B with the workspace `owner` or `admin` role.
3. **Create Agent → Start from scratch**: switch the runtime filter to **All**; Member A's private runtime is now selectable (not locked), and Create succeeds. Previously it was disabled even though the API accepted the same `runtime_id`.
4. Edit an existing agent and reassign its runtime to Member A's private runtime via the inspector picker — it is selectable and saves.
5. Duplicate an agent that runs on Member A's private runtime — the copy stays on that runtime and keeps its model/thinking/service-tier.
6. As a plain `member` (non-admin), Member A's private runtime is still locked — the override is role-gated, not a blanket unlock.

Automated:
- `pnpm --filter @multica/core test runtimes/access.test.ts agents/draft.test.ts`
- `pnpm --filter @multica/views test agents/`

## Checklist


itsXuSt:
@jbarket @quake @chouti @valo @multica-eve

PR #6928, opened 2026-08-13
feat(skills): link imported skills to their source

gsayer:
Follow-up to #6847.

## What

Imported skills now expose their provenance as a clickable link everywhere it is shown:

- **Skill detail page**: the "Imported · GitHub / Skills.sh / ClawHub" chip is now an anchor to `origin.source_url` (opens in a new tab; the full URL appears in a tooltip on hover).
- **Skills list**: the Source column cell gets the same link treatment (`stopPropagation` so the row's whole-row navigation is not triggered).
- **Update-from-source dialog**: the source URL is displayed (middle-truncated, full URL in tooltip) so you can see exactly what will be re-downloaded before confirming.

Runtime-local and manual origins, and imports without a stored `source_url`, keep their plain-text rendering.

Frontend-only; no API, i18n, or schema changes. Uses the existing anchor-in-tooltip pattern from `project-resources-section.tsx` and the `midTruncate` helper.

## Testing

- 3 new tests on the detail page chip (linked github origin, plain manual origin, imported origin without `source_url`).
- New `refresh-skill-dialog.test.tsx` (URL shown + truncated, absent when no `source_url`).
- `pnpm typecheck` and the skills views suites pass.


Bohan-J (changes requested):
Thanks for this — making an imported skill's origin traceable is the right call, and it closes the loop nicely with the "refresh from source" capability from #6847. The implementation is restrained: it reuses the existing Tooltip and truncation patterns, degrades gracefully when `source_url` is absent, and touches no API, schema, i18n, or migration surface.

Two things I'd like fixed before merge.

### 1. Middle-clicking the source link opens the skill instead of the source

`packages/views/skills/components/skills-page.tsx:369` stops propagation on `onClick` only, but the row also handles `auxclick` via `useRowLink` (`packages/views/navigation/use-row-link.ts:63-67`):

```ts
onAuxClick: (e: React.MouseEvent) => {
  if (e.defaultPrevented || e.button !== 1) return; // middle click
  e.preventDefault();
  open("background-tab");
},
```

A middle click on the source anchor bubbles to the row, the row handler calls `preventDefault()`, which cancels the anchor's native open — so the skill detail page opens in a background tab instead of the source page. Plain left click is fine (it's covered by the existing `stopPropagation`), and so is cmd/ctrl+click; it's specifically the middle click that inverts.

Adding `onAuxClick={(e) => e.stopPropagation()}` to the anchor fixes it. Given that the row comment at `use-row-link.ts:16-18` already states that interactive cells are expected to stop propagation, it may be worth handling both events in one shared helper so the next interactive cell doesn't rediscover this. A list regression test for middle click would be good to have alongside the existing coverage.

### 2. A persisted `source_url` becomes a trusted link without any scheme/host validation

`isRefreshableOrigin` (`packages/views/skills/lib/origin.ts:40`) only checks the origin type and that `source_url` is a non-empty string:

```ts
return (
  (origin.type === "github" || origin.type === "skills_sh" || origin.type === "clawhub") &&
  typeof origin.source_url === "string" &&
  origin.source_url.length > 0
);
```

That was fine while the value was only used to gate a refresh button, but this PR promotes it into an `href` at three call sites, so the string now needs to survive being treated as a destination.

The value is attacker-controllable by anyone who can update a skill: `UpdateSkillRequest.Config` is typed `any` (`server/internal/handler/skill.go:233`) and is marshaled into the row verbatim, with no validation of `origin`:

```go
if req.Config != nil {
    config, _ := json.Marshal(req.Config)
    params.Config = config
}
```

The import path does validate — `detectImportSource` (`server/internal/handler/skill.go:801`) allowlists `github.com` / `skills.sh` / `clawhub.ai` and requires `http(s)` — but the update path bypasses it entirely. So `{ "origin": { "type": "github", "source_url": "https://evil.example/..." } }` renders with a GitHub label and a tooltip, and sends other members to an arbitrary host. Non-`http(s)` schemes aren't rejected at this layer either, so it doesn't fail closed.

`packages/views/skills/components/refresh-skill-dialog.tsx:112` is the weakest of the three — it skips `isRefreshableOrigin` altogether and links on a bare non-empty check, so even a `runtime_local` or manual skill with an injected `source_url` gets a live link in the confirm dialog:

```tsx
{typeof origin?.source_url === "string" && origin.source_url.length > 0 && (
```

I'd suggest one shared helper mirroring `detectImportSource`: accept only `http`/`https`, require the host to match the declared origin type, and return `null` otherwise. Then have all three sites (`skills-page.tsx:353`, `skill-detail-page.tsx:310`, `refresh-skill-dialog.tsx:112`) consume only the validated value and fall back to plain text. Worth negative tests for a host/type mismatch, a `data:` URL, and a malformed URL.

Centralizing it also keeps the three display sites from drifting apart later — they've already drifted once, which is how the dialog ended up without the guard.

### Verification

Checked against `e8c3b81c`. CI is green (9/9) and the branch is currently mergeable; the changed-file tests pass and `@multica/views` typechecks clean, with no new lint errors. Both issues above are read off the source at that SHA rather than only reasoned about.

gsayer:
Thanks for the thorough review — both points were real. Fixed in 7a8516cc4.

### 1. Middle click on the source link

Confirmed: the anchor stopped `click` only, so `auxclick` bubbled into `useRowLink`'s row handler, whose `preventDefault()` cancelled the anchor's native open and background-tabbed the skill instead of the source.

Went with the shared-helper suggestion: `rowLinkInteractiveProps` (exported from `packages/views/navigation`) stops both `click` and `auxclick` without touching the element's own default action, and the `useRowLink` doc comment now points interactive cells at it instead of hand-wiring `onClick` alone. The list's source anchor spreads it.

Coverage:
- a nested-anchor contract test in `use-row-link.test.tsx` (middle click on the child: not default-prevented, row handlers never fire);
- a `SkillsPage` regression rig (`skills-page.source-link.test.tsx`): middle click on the source link is not hijacked, left click doesn't navigate the row, plus a sanity case proving the rig actually exercises the row's auxclick handler (middle click elsewhere on the row still background-tabs the skill) so the first assertions can't pass vacuously.

### 2. `source_url` promoted to an href without validation

Added `originSourceUrl(origin)` in `skills/lib/origin.ts`, mirroring `detectImportSource`'s allowlist: http(s) only, hostname must match the declared origin type (`github.com`, `skills.sh`, `clawhub.ai`, each with the `www.` variant), null otherwise. All three sites — list Source cell, detail chip, and the refresh dialog (which had indeed drifted to a bare non-empty check) — now consume only the validated value and fall back to plain text; the dialog renders nothing rather than surfacing an attacker-chosen string in a confirm flow.

One deliberate asymmetry: `isRefreshableOrigin` keeps its looser check rather than delegating to the new helper, because the server's refresh path (`fetchImportedSkillFromOrigin`) also accepts scheme-less values — a bare ClawHub slug import stores `source_url: "<slug>"`, which is refreshable but meaningless as an href. Refresh eligibility and link eligibility are intentionally two predicates; the helper's doc comment records why.

Negative tests: host/type mismatch (including `github.com.evil.example` lookalikes), `data:` / `javascript:` / `ftp:` schemes, malformed and scheme-less values, and an injected `source_url` on `manual` / `runtime_local` origins — as unit tests on the helper plus integration negatives on the dialog and the detail chip.

`pnpm typecheck` and the full skills + navigation vitest suites pass (123 tests).

Bohan-J (approved):
Both must-fixes are addressed cleanly — thanks for the careful follow-up, and for going past the literal asks.

**Middle-click hijack.** `rowLinkInteractiveProps` stops both `click` and `auxclick` without calling `preventDefault`, so the anchor keeps its native middle-click open while the row stays out of the way. Making it a shared export rather than a one-off handler on this cell is the better fix — the comment on it explains the trap well enough that the next interactive cell won't rediscover it. `use-row-link.test.tsx` pins the shared contract, and `skills-page.source-link.test.tsx` covers left click, middle click, and the sanity case that a middle click elsewhere on the row still background-tabs the skill.

**Source URL trust boundary.** `originSourceUrl` requires `http(s)` and an exact host match against the declared origin type, mirroring the server's `detectImportSource` allowlist. Routing it through `new URL()` means the usual lookalikes fail closed too — userinfo tricks like `https://github.com@evil.example/` and suffix domains like `github.com.evil.example` both resolve to a non-matching hostname. All three call sites now consume only the validated value, including the refresh dialog, which previously linked on a bare non-empty check.

Keeping `isRefreshableOrigin` separate rather than tightening it was the right call: a scheme-less ClawHub slug is still refreshable server-side even though it's meaningless as an `href`, so collapsing the two would have broken refresh for those skills to satisfy a UI concern. The doc comment says as much, which is the part that will keep the two from being merged later by someone tidying up.

Verified at `e1c90b24`: CI is 9/9 green and the branch is `MERGEABLE / CLEAN`. Negative coverage in `origin.test.ts` hits host/type mismatch, non-`http(s)` schemes, malformed and scheme-less values, and non-import origins carrying a well-formed URL.

Merging now. Thanks again for the contribution.

PR #6930, opened 2026-08-13
MUL-6301: feat(skills): bulk update from source

gsayer:
Rescoped per the maintainer decision in [this comment](https://github.com/multica-ai/multica/pull/6930#issuecomment-5312391981): the freshness tracking is gone entirely; this PR is now the bulk update only. The branch was rebuilt clean on current `main` (which includes #6928), so the previous conflict is gone.

## What

The skills list batch toolbar gains an **Update** action for imported skills:

- Selecting skills and picking Update opens a confirm dialog that shows how many skills will be re-downloaded from their hosted source and how many selection entries are skipped (manual / runtime-local / archive imports, or no edit permission).
- On confirm, the existing single-skill `POST /api/skills/:id/refresh` runs sequentially across the updatable selection — no new endpoint, no freshness probe. Re-downloading an already-current skill is the accepted tradeoff: the cost lands only on an explicit click.
- **Per-item failure tolerance**: one failing refresh (e.g. a 409 upstream rename collision) does not strand the rest of the batch. A partial run reports an `Updated N, M failed` toast and keeps the selection, so retry is one click; the selection only clears after a fully clean run.
- The Update button is disabled (with an explanatory tooltip) when the selection contains nothing updatable.

Frontend-only; no API, schema, or server changes. i18n copy in en/ja/ko/zh-Hans plus a short docs paragraph in the four languages.

## Testing

- New `skill-list-actions.test.tsx`: dialog summary + skipped line, sequential refresh across the selection with success toast and selection clear, and the partial-failure path (second item still refreshed, partial toast, selection kept).
- `pnpm typecheck`, the `packages/views` skills vitest suite (87 tests), and eslint on the touched files pass.


Bohan-J:
Thanks for this, and apologies in advance — we'd like to accept **half** of it. This is a product-scope decision, not a quality one: we want the bulk update, and we want to drop the freshness tracking entirely.

## Why

We looked at a real workspace (42 skills) before deciding:

| origin | count |
| --- | --- |
| manual / no origin | 27 |
| github | 10 |
| agent_template | 3 |
| skills_sh | 1 |
| runtime_local | 1 |

Only 11 of 42 (26%) are externally imported and therefore refreshable at all — for the other 64% freshness is meaningless.

More decisive: none of those 11 has `upstream_version` or `fetched_at` in `config.origin`. By this PR's own design markerless legacy skills never probe and classify as `unknown`, which "renders exactly as before" — so on the day this merges, **not one existing skill would show a freshness state**. The marker is only written on import or refresh, so every skill needs a manual refresh before the indicator does anything. That cold-start isn't mentioned in the description, and it means the feature's payoff is deferred indefinitely for the corpus that already exists.

And looking at what those 11 actually are — design guidelines, vocabulary references, review prompts — they're one-shot reference material. Upstream rarely moves, and being a few commits behind doesn't hurt. What that content wants is "let me update when I decide to," not a system continuously telling me it's stale.

## What we'd like instead

Cut the freshness infrastructure completely rather than keeping half of it:

- Drop the `upstream_version` / `fetched_at` markers and the extra version probe on import.
- Drop `POST /api/skills/refresh-check`, the server-side cache, and the client query + status types.
- Drop the detail-page `Up to date` / `outdated` indicator, its copy, and the related docs changes.
- Keep the bulk dialog, simplified: show "will update N, skip M", then on confirm reuse the existing single-refresh endpoint sequentially across the refreshable selection. Please keep the per-item failure tolerance and the retry-after-partial-failure behaviour — that part is the real value here.

The accepted tradeoff: bulk update will re-download a skill that was already current. That's fine — the cost lands only when someone explicitly clicks Update, which is much better than probing on every list visit or detail-page open.

Two housekeeping items:

- Please rename the PR to `feat(skills): bulk update from source` to match the reduced scope.
- It needs a rebase — it's currently conflicting. The good news is that most of the conflict should evaporate with freshness removed: the overlap with the just-merged #6928 is in `refresh-skill-dialog.tsx`, `skill-detail-page.tsx`, and `skill-detail-page.test.tsx`, and the detail-page changes leave the diff entirely under this plan.

To be clear about where this stands: we have **not** code-reviewed the implementation yet — this is purely a scope decision, so please don't read it as a verdict on the code. We'll do a full review once it's reshaped. Sorry to ask you to delete working, well-tested code (the 13 Go tests were clearly careful work); we'd rather tell you now than after a detailed review of code we're about to ask you to remove.

Related: we're closing #6931 and #6932 as not planned for the same reason.

gsayer:
Makes sense  and Thanks for your review.

Reshaped as requested, in `d333c73`:

- **Freshness is gone entirely** — no markers or import-time probe, no `refresh-check` endpoint/cache/client query/status types, no detail-page indicator, and the related docs and builtin-skill reference additions are dropped with it. The PR is now frontend-only: zero server changes.
- **Bulk update kept, simplified**: the dialog shows "N will be updated from source / M skipped", and confirm runs the existing single-skill refresh sequentially across the updatable selection. Per-item failure tolerance and the retry-after-partial-failure behaviour are preserved exactly (partial run → `Updated N, M failed` toast, selection kept; it only clears on a clean run), with tests pinning both.
- **Housekeeping**: PR renamed to `feat(skills): bulk update from source`, and the branch was rebuilt clean on current `main` (post-#6928), so it's mergeable again — the description is updated to match the reduced scope.

Understood on #6931/#6932 as well.

multica-eve:
Thank you, Guillaume — and thanks especially for reshaping the contribution so thoughtfully around the maintainer feedback. The focused bulk-update flow is a good fit, and the sequential refresh plus partial-failure recovery are particularly well handled.

I re-reviewed the current head and all CI checks are green. We appreciate the careful work and the responsive iteration!

PR #4829, opened 2026-07-02
feat(dingtalk): add DingTalk bot integration (MUL-3958)

yyclaw:
## What does this PR do?

Adds a first-class DingTalk (钉钉) bot integration on the existing channel architecture. A workspace owner/admin connects one BYO DingTalk Stream-mode robot to an agent; workspace members can then DM it, @-mention it in a group, send images, use `/new`, and use `/issue` without leaving DingTalk.

This revision is rebased on current `main`. The shared channel work from #6388 and #6398 is now upstream, so this PR drops the duplicated shared engine, Lark, Slack, and shared chat-SQL changes and keeps the DingTalk adapter plus its product surface.

### Architecture and scope

- `/new` is handled by the upstream shared channel Router. `/new <message>` runs that message without prior context; bare `/new` records the fresh intent for the next non-empty message and does not create an empty turn.
- `/issue` uses the upstream shared synchronous issue-command path and `IssueService.Create`, with `origin_type = dingtalk_chat`. DingTalk posts the created issue identifier/title back to the originating conversation.
- Images use the shared `MediaResolver` + DB intent-ledger flow. With #6388 upstream, images accompanying `/issue` are bound to the created issue through the shared issue-media promotion path.
- The PR contains no private QuickCreate implementation, private Fresh/session model, private media pipeline, mobile change, or shared channel-engine/Lark/Slack change.
- DingTalk-specific persistence remains local to this adapter: migration 259 adds `dingtalk_chat` to the issue-origin constraint, and `dingtalk.sql` implements AppKey identity-replacement semantics.
- #6005 remains a separate Draft and is not a dependency of this PR.

## Related Issue

Closes #4791

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [x] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- `server/internal/integrations/dingtalk/`: BYO install/revoke, encrypted AppSecret storage, account binding, Stream WebSocket transport, inbound normalization, per-conversation dispatch, outbound delivery, markdown chunking, image ingestion, and tests.
- `server/internal/handler/dingtalk.go`, `server/cmd/server/router.go`: member/admin-scoped HTTP APIs and runtime registration.
[...]

PR #6944, opened 2026-08-14
MUL-6166: feat(telegram): add Telegram channel integration

leonzone:
## What does this PR do?

Adds Telegram as a first-class Multica channel alongside the existing integrations. A workspace can securely connect a BotFather token, bind Telegram users to Multica accounts, route private/group/forum-topic messages to agents, and stream quoted replies back to Telegram.

The implementation uses the shared channel registry, supervisor, installation, identity, and session model. Bot tokens are encrypted at rest, one active workspace owns a bot, outbound delivery requires authoritative Telegram provenance, and Telegram-created Issues retain explicit origin metadata.

Telegram v1 has a deliberate text-message scope. Addressed non-text messages receive a clear response and create no task.

## Related Issue

N/A — this contribution was developed and validated directly against the channel integration contract.

## Type of Change

- [x] Documentation update

## Changes Made

- Added the Telegram Bot API adapter, long polling, secure workspace-scoped installation, account binding, inbound routing, sender/replier, Markdown handling, command routing, and streamed quoted replies under `server/internal/integrations/telegram/`.
- Added private-chat, group, forum-topic, reply/quote, UTF-16 entity, long-message, reconnect, concurrent-stream, ownership, and error-path coverage.
- Added Telegram handlers, protocol events, channel registration, settings and agent UI, web binding route, core API/types, realtime invalidation, and all four product locales.
- Added migrations 349 and 350 plus migration tests for `origin_type = 'telegram_chat'` using the repository's two-phase constraint-validation pattern.
- Added localized Telegram setup guides and synchronized landing copy in English, Chinese, Japanese, and Korean.
- Added architecture and final validation evidence in `HANDOFF-telegram.md` and `docs/telegram-channel-review-2026-08-08.md`.

## Review follow-up

All five requested changes are included in the single rebased commit:

1. Group replies to human messages now include quoted text/caption and sender attribution only when the bot is explicitly mentioned. `CommandText` retains only the current user's cleaned instruction, preserving command isolation; reply-to-bot behavior is unchanged.
2. The branch is rebased on current `main`; Telegram migrations are the next free consecutive numbers, 349 and 350.
3. Task cancellation releases stream and chat scheduling state. The already-visible partial Telegram message is intentionally retained without a synthetic terminal notice.
4. Partial and final delivery share chat-scoped throttling and `retry_after` backoff. Final delivery waits and retries after a 429 response.
5. The incomplete `/fresh` alias and its claims were removed. The shared persisted `/new` behavior remains supported.

The localized docs now state the context boundary exactly: accepted messages remain in the conversation, unaddressed group chatter is not collected, and quoted human content is included only with an explicit bot mention.

## Latest re-review follow-up

The three blocker groups reported against `29cc89e` are addressed in the rebased single commit:

1. Telegram parses `/new` before quoted-human enrichment and sets `ForceFresh`, so the shared router preserves the quoted sender/content while starting a fresh session. The original cleaned command remains isolated in `CommandText`.
2. Chat cooldown and `retry_after` state now survive sequential tasks and cancellation in a hard-bounded cache. Terminal replies use an asynchronous keyed FIFO scheduler: one Telegram request per worker step, retry-heap rescheduling, per-session ordering, resumable chunk progress, and fixed queue/count/byte limits. Long multi-chunk replies and `retry_after` values beyond the former 15-second budget are covered.
3. `retry_pending=true` failures now release the old stream without sending a false terminal failure notice; non-retry terminal failures still send the tested notice.

The branch was rebased onto current `main`; migrations were renumbered to the next free consecutive slots, 349 and 350. Process-local queue limits, conservative same-installation fallback behavior, and restart boundaries are documented in all four localized Telegram guides and the handoff/review reports.

## How to Test

1. Run `pnpm typecheck`, `pnpm lint`, and `pnpm build` from the repository root.
2. From `server/`, run `go test ./internal/migrations -count=1`, `go test -race ./internal/integrations/telegram -count=1`, and `go test ./internal/handler -count=1`.
3. Start Multica locally, connect a BotFather token from an agent's Telegram integration, send the bot a private message, complete the binding link, and confirm inbound task creation plus a streamed quoted reply.
4. Verify `/new` and `/issue`, including `@botname` command suffixes, long replies, Emoji/entity offsets, rapid messages, and reconnect behavior.
5. In a group, reply to another member's message and mention the bot. Confirm the agent receives the quoted sender/content plus the new instruction. Confirm the same human reply without a bot mention does not trigger.
6. Send an addressed non-text message and confirm the bot returns the documented text-only response without creating a task.
7. Create an Issue through `/issue` and confirm the persisted row uses `origin_type = 'telegram_chat'` with a non-null `origin_id`.

## Validation

All Telegram acceptance checks and build gates for this contribution pass:

- `pnpm typecheck`
- `pnpm lint` (zero errors; repository warnings unchanged)
- `pnpm build` for Web, Docs, and Desktop
- 244/244 focused Core API, schema, and realtime tests
- 62/62 focused Views tests for Telegram settings, binding, integrations, and agent surfaces
- 234/234 Web tests
- `go test ./internal/migrations -count=1`, including real local PostgreSQL up/down/catalog verification
- `go test -race ./internal/integrations/telegram -count=1`
- quote-context, command-isolation, cancellation, concurrent same-chat throttling, and 429 final-delivery regressions
- `git diff --check` and credential-pattern scans

The repository-wide TypeScript and Go race suites were also exercised. Their unrelated timing/host-process cases all passed in isolated reruns with the same test mode. Detailed command evidence is recorded in the validation report.

Live Telegram validation covered installation, private account binding, inbound task creation, runtime execution, streamed outbound replies, `/new`, `/issue`, `@botname` suffixes, reply quoting, reconnect, long output, supplementary-plane Emoji, Issue provenance, and the text-only media response. Group mention/reply, human quote context, command isolation, and forum-topic behavior are covered by deterministic Bot API protocol tests.

## Risks and Boundaries

- Telegram is text-only in this release; unsupported media follows an explicit, tested response path.
- The server must be able to reach `api.telegram.org`; the adapter uses Go's standard proxy-aware transport.
- One active workspace owns a bot token, preventing competing long-poll consumers.
- Ambient group history is not collected.
- No production deployment, production configuration, or real-user data operation is included.

## Checklist

## AI Disclosure

**AI tool used:** OpenAI Codex and ChatGPT Pro

**Prompt / approach:** Codex inspected the repository and contribution rules, prepared a credential-scanned source snapshot for external review, treated external findings as advisory, implemented and reviewed the final changes locally, ran repository and focused checks, and performed a real Telegram regression against a local Multica instance. Acceptance is based on source review and reproducible local evidence.

## Screenshots

**Before — integrations page without Telegram**

codex-clipboard-235d0a93-335a-46a6-b0fb-d59055dc34a6

**After — Telegram displayed alongside the existing channel integrations**

codex-clipboard-c60c3732-0b9e-4ec0-adec-afa88e697231


multica-eve (dismissed):
## Product assessment

Before this change, workspace members cannot hand work to an agent from Telegram; after it, they can bind their identity, talk to the agent in private or group conversations, and receive the result in the same place. We want this as an officially supported integration, but the current group experience does not meet the minimum context bar: a member must at least be able to reply to another person's message, mention the bot, and have the agent understand both the quoted message and the new instruction. Today the agent receives only the instruction, so common requests such as “summarize this” or “reply to this” lose their subject.

**Product decision: the direction is accepted, but changes are required before merge.** Ambient group history can remain out of scope for v1, but quoted human-message context is now a minimum acceptance requirement.

## Must fix

1. **Preserve quoted human-message context.** `server/internal/integrations/telegram/api.go:160-169` already decodes the replied-to message, including its text/caption and sender, but `server/internal/integrations/telegram/inbound.go:79-84` keeps only its message/root IDs. Please enrich the agent-readable message with the quoted human text (or caption) and sender attribution when a group member replies to a human message and explicitly mentions the bot.

   - Replying to a human message must not trigger the bot without an explicit mention.
   - Keep the user's own cleaned instruction in `CommandText` before enriching `Text`, so a quoted `/issue` or `/new` is never interpreted as the current user's command.
   - Preserve the current reply-to-bot behavior.
   - Add tests for human reply + bot mention, sender/text attribution, command isolation, and an empty or non-text quoted message.

2. **Rebase and renumber the migrations.** This PR adds `server/migrations/314_issue_origin_telegram_chat.*` and `315_issue_origin_telegram_chat_validate.*`, while current `main` already contains `314_workspace_mcp_config.*` and `315_workspace_mcp_server.*`. Migration prefixes must be globally unique. Please rebase, move the Telegram migrations to the next free consecutive numbers, and update the migration tests.

3. **Clean up streaming state when a task is cancelled.** `server/internal/integrations/telegram/outbound.go:94-99` subscribes to message, completion, and failure events, but not cancellation. A cancelled task can therefore leave its entry in `streams` indefinitely. Subscribe to the cancellation event, delete the state, define whether the partial Telegram message remains or receives a terminal notice, and add a regression test.

4. **Make throttling and backoff truly chat-scoped.** `server/internal/integrations/telegram/outbound.go:43-64` stores `lastEdit` and `backoffTill` in state keyed by task, so concurrent tasks in one chat are not throttled together despite the stated policy. The completion path at `outbound.go:204-255` also ignores an active `retry_after` backoff, which can lose the final reply. Use shared chat-level scheduling/backoff for partial and final delivery, with tests for concurrent tasks in one chat and completion after a 429 response.

5. **Fix or remove the bare `/fresh` behavior.** `server/internal/integrations/telegram/inbound.go:186-194` turns bare `/fresh` into empty text with `ForceFresh=true`, then `server/internal/integrations/telegram/telegram_channel.go:125-127` drops it before the shared command handling runs. Either persist a “next message is fresh” intent as claimed and cover it end-to-end, or remove the alias and its verification claims, retaining the documented `/new` behavior.

Once these blockers are addressed, please also update the Telegram docs to state the exact context boundary: accepted messages remain in the conversation, unaddressed group chatter is not collected, and a quoted human message is included only when the bot is explicitly mentioned.

leonzone:
Addressed all five requested changes in the single rebased commit 29cc89e:

1. Added explicitly mentioned human-reply context with sender/text or caption attribution, while keeping CommandText isolated and preserving reply-to-bot behavior.
2. Rebased onto current main and renumbered the Telegram migrations to 344/345.
3. Added cancellation cleanup; an existing partial Telegram message remains visible by policy.
4. Added shared chat-scoped throttling/backoff and final-delivery retry after Telegram 429 responses.
5. Removed the incomplete /fresh alias and retained the shared /new contract.

Localized documentation now states the exact group-context boundary. Regression coverage includes quoted command isolation, empty/non-text quotes, cancellation, concurrent tasks in one chat, and completion after retry_after. Telegram race tests, migration tests against local PostgreSQL, typecheck, lint, focused Views/Web tests, and Web/Docs/Desktop production builds pass. Ready for re-review.

multica-eve (dismissed):
## Re-review result

Thank you for addressing the first review. The intended Telegram experience is still one we want: ordinary quoted-human-message prompts now carry the selected message into the agent conversation, the migration collision and cancellation leak are fixed, and the incomplete `/fresh` alias is gone. However, the current revision can still lose that quoted context in a supported command flow and can drop or mislabel replies under normal rate-limit and retry conditions.

🔄 **Request changes** — three groups of blockers remain at `29cc89e`.

## Must fix

1. **`/new` discards the quoted human-message context.** `server/internal/integrations/telegram/inbound.go:60-65` builds enriched `Text` while keeping the sender's instruction in `CommandText`. However, `server/internal/integrations/channel/engine/router.go:198-204` parses `/new` and, because Telegram has not set `ForceFresh`, replaces `Text` with the stripped command body. Reproduction: reply to a human message with `@my_bot /new summarize this`; after the shared router rewrite, the agent-readable body is only `summarize this` and the quote is gone.

   Please parse/strip `/new` before quote enrichment (as the Lark path does), or otherwise preserve enriched `Text` through the shared rewrite. Add a regression test for human reply + explicit bot mention + `/new`; the quoted sender/content must remain while the session is fresh.

2. **The chat-scoped rate-limit state does not survive sequential tasks, and the terminal delivery budget can truncate valid output.** `server/internal/integrations/telegram/outbound.go:364-371` deletes `chatSchedule` as soon as its active reference count reaches zero, discarding both `lastEdit` and `backoffTill`. A task that starts immediately afterward gets a new schedule and bypasses the previous chat cooldown; cancellation can likewise erase an active Telegram `retry_after`. A focused regression check confirms that two sequential tasks receive different schedules.

   Keep cooldown/backoff state for a bounded period after the last active task (without creating an unbounded cache), and cover sequential tasks plus cancellation after a 429.

   There is also a delivery-budget mismatch: `outbound.go:215-218` gives terminal delivery 15 seconds, while `runScheduled` spaces every final chunk by 2.5 seconds and waits for the full `retry_after`. A response longer than roughly six chunks, or a 429 whose delay exceeds the remaining budget, loses its tail even with no other error. Make the budget account for chunk count and Telegram backoff, or move terminal delivery to a durable retryable path. Please test a long multi-chunk response and `retry_after > 15s`.

3. **An auto-retry emits a false terminal failure message.** Task failure events include `retry_pending=true`, but `server/internal/integrations/telegram/outbound.go:286-327` never checks it and always edits/sends the terminal “agent failed” message. The retry task can then succeed, so the user sees a failure followed by a valid answer. A focused regression check confirms one terminal Telegram send for a retry-pending event.

   When `retry_pending=true`, clean up or hand off the old stream without sending a terminal failure; only send the failure notice when the retry is not pending. Add both pending and terminal failure cases to the tests.

## Verification

- `git diff --check HEAD^..HEAD` passes.
- `go test -race ./internal/integrations/telegram -count=1` passes.
- `go test ./internal/handler -run Telegram -count=1` passes.
- `go test ./internal/migrations -run Telegram -count=1` passes.
- Current GitHub CI is green.
- Three temporary focused regression tests for the cases above fail as described; they were removed after verification and the worktree is clean.

leonzone:
@multica-eve The latest requested changes are addressed in `c7e085e`, and the branch is rebased onto current `main`.

- `/new` now preserves explicitly mentioned quoted-human context while forcing a fresh session.
- Chat cooldown and `retry_after` survive sequential tasks and cancellation; terminal delivery is asynchronous, keyed, bounded, resumable by chunk, and has no fixed 15-second budget.
- `retry_pending=true` no longer sends a false terminal failure notice; terminal failures remain covered.
- Telegram migrations are renumbered to 349/350.

Validation on the rebased commit: `pnpm typecheck`, `pnpm lint` (0 errors), Web 234/234, focused Core 244/244, focused Views 62/62, Web/Docs/Desktop production builds, Go Telegram/cmd/server/handler/migrations tests, Telegram race tests, Go vet, diff and credential scans.

Please re-review when ready.

leonzone:
@multica-eve The latest requested changes are addressed in \`186fc6c\`, and the branch is rebased onto current \`main\`.

- \`/new\` now preserves explicitly mentioned quoted-human context while forcing a fresh session.
- Chat cooldown and \`retry_after\` survive sequential tasks and cancellation; terminal delivery is asynchronous, keyed, bounded, resumable by chunk, and has no fixed 15-second budget.
- \`retry_pending=true\` no longer sends a false terminal failure notice; terminal failures remain covered.
- Telegram migrations are renumbered to 349/350.

Validation on the rebased commit: \`pnpm typecheck\`, \`pnpm lint\` (0 errors), Web 234/234, focused Core 244/244, focused Views 62/62, Web/Docs/Desktop production builds, Go Telegram/cmd/server/handler/migrations tests, Telegram race tests, Go vet, diff and credential scans.

Please re-review when ready.

Bohan-J:
@leonzone — thanks for the fast turnarounds; `c7e085e` closed most of the last round. Two updates on the open review items, and then one thing I owe you an honest conversation about before merge.

**The migration numbering is ours, not yours.** `349`/`350` collide with `main` again, and that's the third collision on this branch (`314/315` → `344/345` → `349/350`). Another rebase from you wouldn't prevent a fourth — it's a race you can't win from the outside. We'll renumber it ourselves at merge time and push it to your branch, the same way we did on #4829. Please don't spend another round on it.

**The delivery-durability item is being re-scoped, not held against you.** Looking at it again against the rest of the codebase: Lark, Slack and DingTalk all deliver the terminal reply synchronously on the event bus with a 10s timeout, one warning log on failure, and no retry or persistence (`server/internal/integrations/slack/outbound.go:71`, `server/internal/integrations/lark/outbound.go:290`) — a single transient API error loses the reply there, no restart required. Your outbound has chat-level scheduling, 429 backoff, per-chunk resume and a bounded queue, which is ahead of all three. Holding this PR to a durable-outbox bar that none of our own channels meet would be unfair, so that moves to a platform-wide follow-up covering every channel rather than staying a blocker here.

---

## How we'd support this long-term

Our team doesn't use Telegram day to day — the same position we were in with DingTalk (#4829) and WeCom (#5833) — so we can't meaningfully verify this against the real platform. You're the person here who understands both this codebase and Telegram. Rather than merge quietly and leave that gap unspoken, here's what we'd propose:

**What we commit to:** we keep the adapter compiling and its tests green through shared-layer refactors — the rebase treadmill you've been on ends at merge and becomes our problem. It ships in every release, and Telegram-specific issues get routed to you rather than sitting unowned.

**What we'd ask of you:** be the code owner for `server/internal/integrations/telegram/` — Telegram-specific bugs and behavior questions come to you first, best-effort. You're a volunteer and we'd treat it that way: "we'll route these to you," not "you're on the hook to fix them."

**The exit rule, stated plainly:** the integration would be documented as community-maintained. If it breaks in a way we can't fix without real Telegram access, and no fix appears for a few releases, we may deprecate it rather than leave it quietly broken.

That rule exists to take pressure off you, not add to it. It is not a deadline and not an expectation that you stay available forever — it's our commitment that if you get busy, change jobs, or simply move on, the worst outcome is that we retire the integration, not that anyone comes chasing you for a fix. It also keeps us honest with users, who deserve to know what is and isn't officially supported.

**And you can say no.** Declining doesn't block the merge — it changes the labelling, not the decision. And if you'd rather not carry it alone, co-ownership works too: WeCom is co-owned by @leroy-chen and @seacen for exactly that reason.

---

If you're in, two small things belong in this PR (or we can push them along with the renumbering — your call):

1. **Support status in the docs.** `channels.mdx` currently reads "DingTalk and WeCom are community-maintained…"; Telegram should join that sentence, and `telegram-bot-integration.mdx` should carry the same one-liner near the top, the way `dingtalk-bot-integration.mdx` does. All four locales.
2. **A `Maintenance:` block in the package doc** naming you as code owner — mirror `server/internal/integrations/dingtalk/config.go` and `server/internal/integrations/wecom/types.go`.

One unrelated docs nit this PR introduced, worth fixing either way: in all four `channels.*.mdx` locales, the Telegram text-only sentence *replaced* the WeCom one instead of being added next to it — "WeCom handles text messages today. Voice, image, and file messages get a short reply explaining that…" is gone in en/zh/ja/ko. Both limitations should be documented.

leonzone:
@Bohan-J Thank you for the thoughtful and transparent note, and for reassessing the migration-numbering and delivery-durability items against the standards used by the existing channel integrations. I appreciate the team taking ownership of migration renumbering at merge time and moving durable delivery to a platform-wide follow-up.

I’m happy to accept best-effort code ownership for `server/internal/integrations/telegram/` under the community-maintained model you described. The expectations and exit rule are clear and reasonable. Please feel free to route Telegram-specific bugs and behavior questions to me, and I’ll help whenever I’m able.

I’ve updated all four locales to document Telegram’s community-maintained status, added the `Maintenance:` block naming me as code owner, and restored the WeCom text-only limitation alongside Telegram’s in `d9749c0`. I’ve left the migration numbers unchanged for the team to renumber at merge time, as requested.

Validation for this follow-up: Docs production build and typecheck pass, as do the Telegram package tests with and without the race detector.

Thank you again for handling this so thoughtfully and for being explicit about the long-term maintenance expectations.

Bohan-J:
@leonzone — thanks for taking the code-owner role, and for turning the docs and the `Maintenance:` block around so quickly. `d9749c0` looked right in all four locales, and the WeCom text-only line is back alongside Telegram's.

**Heads-up: I rebased your branch onto current `main` and pushed the renumbering, so `codex/telegram-integration` was force-updated** (`d9749c0` → `f7a099a8`). Please pull before your next push so you don't end up diverged.

What I did, in one `chore` commit on top of your unchanged work:

- `349`/`350` → **`366`/`367`** (`main` had reached 365, and 349/350 were taken by `agent_task_queue_chat_terminal_resume_index` / `agent_task_queue_chat_retired_session_index`).
- Updated the cross-references inside the SQL comments so 366 still points at 367 and vice versa.
- Pointed the four hard-coded filenames in `server/internal/migrations/telegram_origin_migration_test.go` at the new names.

Your commit is untouched apart from the rebase itself — the renumbering is a separate commit so it stays visible as ours.

Verified locally before pushing: `go build ./...`, the migration lint suite (`TestMigrationFilesHaveMatchingDirections`, `TestMigrationNumericPrefixesStayUniqueAfterLegacySet`, `TestNewMigrationPrefixesStartAfterLegacyRange`), `TestTelegramOriginMigrationsUpDownAndCatalog` against a real Postgres (up → validate → rollback → narrow, all green with the new numbers), `go test -race ./internal/integrations/telegram`, and `go test ./internal/handler -run Telegram`. CI is re-running now on `f7a099a8`.

I also double-checked the constraint list while I was in there: `main`'s most recent widening is still `263_issue_origin_wecom_chat`, so your `up` adds `telegram_chat` on top of the current set and your `down` restores exactly that set — no risk of a rollback quietly revoking `dingtalk_chat` or `wecom_chat`.

Bohan-J (approved):
Approving — thank you, @leonzone. This was five rounds over five days, and every one of them came back with the actual fix rather than a workaround: the quoted-human context surviving `/new`, chat-level cooldown and `retry_after` persisting across sequential tasks and cancellation, per-chunk resume for long replies, and not firing a terminal failure message while a retry is still pending. Landing a new first-party channel end to end — install, binding, groups, forum topics, streaming edits, `/issue` provenance, four locales of docs — is a lot of surface to get right from outside the team, and the test coverage you brought with it is the reason we could verify the edge cases at all.

Thanks also for taking the code-owner role, and for saying yes in a way that let us be honest about what we can and can't support.

CI is green on `f7a099a8` (backend, backend-tests, sqlc-check, vulnerability scan, frontend build/test, views, mobile, installers, windows-execenv). Merging now.

Two things for after the merge:

- Telegram-specific issues will be routed to you; nothing is expected of you outside that.
- Durable outbound delivery is going on our side as a platform-wide follow-up covering every channel, not a Telegram task.

Welcome aboard.

PR #6946, opened 2026-08-14
MUL-6169 fix(views): left-align page titles and unify page gutters

tim-smart:
## Summary

- Left-align page titles that rendered centered below `xl`. The collapsed nav trigger makes the header three flex items, so `justify-between` split the free space on both sides of the title; content groups now grow with `flex-1` instead.
- Make the header's base `gap-2` the single source of leading-slot spacing by dropping the trigger's own margin, which double-spaced every header that also declared a gap.
- Unify page gutters behind an exported `PAGE_GUTTER` constant. Collection pages sat at `px-5` while the issues family sat at `px-4`, so switching tabs moved the title 4px; headers, toolbars, skeletons and aligned body rows now read the same constant, and `PAGE_TOOLBAR` shares the full toolbar row recipe.
- Simplification pass on the above: the dashboard header folds onto `CollectionPageHeader`, the autopilot-detail skeleton onto `PageHeader`, dead one-child wrappers and no-op `gap-2` overrides are removed, and the header tests consolidate to one render per behavior with role-based queries.

## Testing

- `pnpm typecheck` (6/6), `pnpm lint` (0 errors), `pnpm test` (all packages green, 3789 views tests)
- Verified in a browser at 390px, 1000px and 1440px: title and toolbar leading edges are flush on every page, all widths

### DIFF
--- a/packages/views/agents/components/agent-detail-page.tsx
+++ b/packages/views/agents/components/agent-detail-page.tsx
@@ -570,7 +570,7 @@ function DetailHeader({
 
 function BackHeader({ paths, title }: { paths: string; title: string }) {
   return (
-    <PageHeader className="px-5">
+    <PageHeader>
       <div className="flex flex-1 items-center gap-2">
         <AppLink
           href={paths}
--- a/packages/views/agents/components/agent-detail-page.tsx
+++ b/packages/views/agents/components/agent-detail-page.tsx
@@ -570,8 +570,8 @@ function DetailHeader({
 
 function BackHeader({ paths, title }: { paths: string; title: string }) {
   return (
-    <PageHeader className="justify-between px-5">
-      <div className="flex items-center gap-2">
+    <PageHeader className="px-5">
+      <div className="flex flex-1 items-center gap-2">
         <AppLink
           href={paths}
           className="inline-flex h-7 items-center gap-1 rounded-md px-2 text-caption text-muted-foreground transition-colors hover:bg-accent hover:text-foreground"
--- a/packages/views/agents/components/agent-detail-page.tsx
+++ b/packages/views/agents/components/agent-detail-page.tsx
@@ -571,15 +571,13 @@ function DetailHeader({
 function BackHeader({ paths, title }: { paths: string; title: string }) {
   return (
     <PageHeader>
-      <div className="flex flex-1 items-center gap-2">
-        <AppLink
-          href={paths}
-          className="inline-flex h-7 items-center gap-1 rounded-md px-2 text-caption text-muted-foreground transition-colors hover:bg-accent hover:text-foreground"
-        >
-          <ArrowLeft className="h-3.5 w-3.5" />
-          {title}
-        </AppLink>
-      </div>
+      <AppLink
+        href={paths}
+        className="inline-flex h-7 items-center gap-1 rounded-md px-2 text-caption text-muted-foreground transition-colors hover:bg-accent hover:text-foreground"
+      >
+        <ArrowLeft className="h-3.5 w-3.5" />
+        {title}
+      </AppLink>
     </PageHeader>
   );
 }
--- a/packages/views/agents/components/agent-list-toolbar.tsx
+++ b/packages/views/agents/components/agent-list-toolbar.tsx
@@ -53,6 +53,8 @@ import { FILTER_ITEM_CLASS, HoverCheck } from "../../common/hover-check";
 import { availabilityConfig } from "../presence";
 import { useT } from "../../i18n";
 import type { AgentListRow } from "./agents-page";
+import { PAGE_GUTTER } from "../../layout/page-header";
+import { cn } from "@multica/ui/lib/utils";
 
 const COLUMN_KEYS: AgentColumnKey[] = [
   "status",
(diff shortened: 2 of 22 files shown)


tim-smart:
gutter-before-after

Bohan-J (changes requested):
The layout work here is good and I want it in — the diagnosis is right, the fix is the right one rather than a patch over the symptom, and the constants are a real reduction in drift surface. One blocker, and it is about what the branch carries rather than about the layout change itself.

## Blocker: the branch carries a pre-review snapshot of #6877

`git log main..HEAD` starts with `dc2fd220 feat(inbox): add a keyboard shortcut for archiving the open notification (#4)`. That is the inbox archive shortcut — which has since merged on its own as #6877 (`16f9f252`), but in a **later** form than the copy on this branch. `dc2fd220` predates the review round on #6877, so relative to today's `main` this branch reverts:

1. **`isPortalLayerShortcutTarget`** — the guard that keeps `E` from firing while a Radix layer owns the keyboard. Without it: select a notification, open the batch-operations `DropdownMenu` in the inbox header, press `e` expecting menu typeahead — Radix does not `preventDefault` for plain character keys, the menu is portaled to `body`, so the event reaches the document listener with `defaultPrevented === false`, `event.target` is a `menuitem` (not editable) and the app modal store is empty. The notification is archived silently behind the open menu.
2. **The archive/restore success toasts**, plus `toasts.archived` / `toasts.unarchived` in all four locales — a bare key that mutates state with no visible confirmation.
3. `actionOnSelectedRef.current` moves back to being assigned during render instead of in an effect.
4. ~148 lines of `inbox-page.test.tsx`, including the `archive feedback` block and the modal-layer case.

Since #6877 landed, the PR has gone to `mergeable_state: dirty` for exactly this reason.

### How to resolve it

The important part: **drop that commit — do not resolve the conflict in favour of this branch.** Taking "ours" on `inbox-page.tsx` is what would re-delete the guard and the toasts.

```bash
git fetch upstream main
git rebase upstream/main        # stops on dc2fd220 with conflicts in
                                # inbox-page.tsx and inbox-page.test.tsx
git rebase --skip               # <- drop it; main already has the newer version
```

Or equivalently, in one step:

```bash
git rebase --onto upstream/main dc2fd2206f5ea0a6807be9be997382363c024e20
```

I ran both locally against current `main` and they produce identical trees. Afterwards:

- The five layout commits replay with **no conflicts**.
- The diff collapses to **22 files, +224/−72**, all layout. `packages/core/shortcuts/`, `packages/views/locales/*/settings.json` and `inbox-page.test.tsx` drop out of the diff entirely.
- `inbox-page.tsx` keeps exactly one hunk — the `listHeader` change — and the portal guard and toasts from #6877 stay intact.
- `packages/views` vitest: **341 files / 4060 tests passed**. `turbo typecheck --filter=@multica/views --filter=@multica/core`: clean.

## On the layout change itself

Nothing blocking. Worth calling out what is done well:

- **The root-cause read is correct.** The leading slot makes a header that reads as two zones in source into three flex items at runtime, so `justify-between` distributes free space on both sides of the title. Letting the content group absorb it with `flex-1` is the right fix, and folding the trigger's `mr-2` into a base `gap-2` on `PageHeader` removes the double-spend rather than compensating for it.
- **`PAGE_GUTTER` / `PAGE_TOOLBAR` earn their keep**, and the comments explain why they are constants instead of per-page classes. I checked the whole tree: no `PageHeader` call site passes `px-*` any more, and the remaining `px-5` in `packages/views` is all dialogs, forms and the agent-create flow — a different chrome family, correctly left alone.
- **The incidental fixes are real bugs**, not drive-by churn: the member-detail, squad-detail and autopilot-detail skeletons sat 4px off the headers they stand in for, so the title jumped once content loaded.
- Reverting the `SquadDetailInspector` hunks was the right call — that `aside` is a card in a grid column, not chrome under a header.
- `page-header.test.tsx` querying by role and asserting against the constant rather than a repeated literal matches the sibling layout tests.

## Nits (non-blocking)

1. `packages/views/inbox/components/inbox-page.tsx:561`, `:676` and `:732` still hand-spell `flex h-12 shrink-0 items-center border-b px-4` — the same shape as the two in `issue-detail.tsx` you converted. Given the constant's own doc says "a page that spells its own gutter is the bug coming back", these three are worth folding in too.
2. `shouldIgnoreGlobalShortcutEvent` (`packages/views/layout/global-shortcuts.tsx:43`) already exports the `defaultPrevented || repeat || isImeComposing` triple that `inbox-page.tsx:367` spells out again. `use-in-page-find.ts` and `issue-detail.tsx` also hand-roll it, so this is existing convention rather than a deviation — but reusing the export would be the better direction if you are touching it anyway.
3. "Do not pass `px-*` to `PageHeader`" is currently enforced only by the comment: `className` sits after `PAGE_GUTTER` in `cn`, so anything passed wins silently. The new tests pin the defaults; an assertion that sweeps call sites would close the drift path for good.

Happy to approve once the branch carries only the layout commits.

Bohan-J (approved):
Thanks for turning this around so quickly — the branch now carries only the six layout commits, and #6877's portal guard, success toasts and tests are untouched.

Re-verified on `73c1cf61`:

- Diff is 22 files, +242/−76, layout only — `packages/core/shortcuts/`, `packages/views/locales/*/settings.json` and `inbox-page.test.tsx` are all out of it.
- `packages/views` vitest: 341 files / 4061 tests passed. `turbo typecheck --filter=@multica/views --filter=@multica/core`: clean.

Nice call on the gutter enforcement — moving `PAGE_GUTTER` after `className` in `cn` is better than the sweep assertion I suggested, because it turns "don't pass `px-*`" from a comment into something a call site structurally cannot do, and the `gap` override the doc promises still works. I checked: no call site passes padding to any of the three headers, and `member-detail-page.tsx`'s `gap-1.5` still resolves.

One optional follow-up, not worth holding this up: the enforcement covers `px-*` but not `p-*` — tailwind-merge won't strip a later `px-4` from an earlier `p-6`, so the horizontal padding would fall to stylesheet order. Nothing writes that today, and the test covers `px-8`; a line in the doc comment would close it out.

Approving and merging.

PR #6949, opened 2026-08-14
MUL-6085: fix(daemon): add completed-task environment retention TTL

### LINKED ISSUE #6856: [Bug]: completed task environments have no full-cleanup TTL while issue stays open
### Deployment type

self-host

### What happened?

Multica 0.4.21 runs daemon GC, but an issue-backed task environment is fully deleted only when the parent issue is done or cancelled and older than MULTICA_GC_TTL. A completed task whose issue remains todo, backlog, blocked, in_progress, or in_review keeps its full source checkout and .git data indefinitely.

The artifact-only pass from #1890 does not bound this growth when the checkout itself is the large object. In this workload, artifact_size_bytes is 0 while the task roots still occupy 53.97 GiB.

This is a lifecycle gap, not a stopped GC loop.

### Steps to reproduce

1. Run an issue-backed task and let it finish so the task root receives .gc_meta.json with completed_at.
2. Leave the parent issue open, for example blocked or in_review.
3. Wait longer than MULTICA_GC_ARTIFACT_TTL and MULTICA_GC_TTL.
4. Let multiple GC cycles run.
5. Run multica daemon disk-usage --output json.
6. Observe that artifact directories may be removed, but the task root, source checkout, .git data, output, logs, and codex-home remain indefinitely because the issue is not done or cancelled.

### Current evidence

- CLI and daemon: 0.4.21, commit 0dfaac266.
- GC startup: interval=2h, ttl=24h, orphan_ttl=72h, artifact_ttl=12h.
- Workspace root: 58.60 GiB total; 53.97 GiB task roots; 4.63 GiB repo cache.
- 196 task environment directories.
- 156 directories resolve a parent issue status.
- 85 status-known completed task directories are older than 24h.
- 57 of those 85 belong to still-open issues, so full cleanup cannot select them.
- Latest observed cycle: cleaned=1, skipped=195, artifact_dirs=0, bytes_reclaimed=674722562.

### Expected behavior

Add an operator-controlled full environment TTL based on task completion time, separate from issue lifecycle, for example:

MULTICA_GC_COMPLETED_TASK_TTL=24h

When .gc_meta.completed_at exceeds this TTL, the daemon should fully remove the inactive task environment regardless of the parent issue status.

Safety requirements:

- Never remove an active environment root.
- Preserve the existing local_directory rule: never fully remove a user-owned local-directory environment.
- Use 0 to disable the new policy if resume retention is preferred.
- Keep the existing issue-status TTL, orphan TTL, and artifact-only TTL as separate policies.
- Log completed_at, parent issue status, and bytes reclaimed.
- A later rerun may provision a fresh environment instead of retaining an unbounded old checkout.

This is the remaining gap after #1890, which deliberately preserved source, .git, output, logs, and metadata for open issues.

iiwish:
## What does this PR do?

Adds an opt-in full-environment retention bound for completed issue tasks whose parent issue remains open.

The existing full-cleanup policy is tied to terminal issue status, while artifact GC deliberately preserves the checkout, `.git`, output, logs, and metadata for resumability. That leaves completed task environments unbounded when an issue stays in `todo`, `blocked`, `in_progress`, or `in_review`.

Operators can now set `MULTICA_GC_COMPLETED_TASK_TTL` to remove an inactive issue-task environment after its `.gc_meta.json` `completed_at` age exceeds the configured duration. The default is `0`, so current retention behavior is unchanged unless an operator opts in.

The decision remains fail-closed: it requires a successful parent-issue status lookup, never removes an active or newly reserved environment, and excludes `local_directory` environments from full cleanup. The latter continue to follow the existing artifact TTL independently, so a shorter completed-task TTL cannot accelerate artifact cleanup.

## Related Issue

Closes #6856

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `MULTICA_GC_COMPLETED_TASK_TTL` configuration with a disabled-by-default value and duration validation.
- Apply the new policy only to completed, inactive, non-`local_directory` issue-task environments after a successful issue reconciliation.
- Preserve issue-status, orphan, artifact-only, and managed-cache TTLs as independent policies.
- Log the configured TTL at startup, eligibility context (`completed_at` and parent status), and per-directory reclaimed bytes.
- Document the operator setting and the fresh-environment behavior on a later rerun.
- Add decision, safety, configuration, and end-to-end GC scan regression coverage.

## How to Test

1. Set `MULTICA_GC_COMPLETED_TASK_TTL=1h`, create an issue task with an old `.gc_meta.json` `completed_at`, leave the issue open, and run a GC cycle; the inactive task root should be removed.
2. Repeat with an active environment, an unreachable parent issue, and a `local_directory` environment; the full task root should remain.
3. Run the automated validation:

```bash
go test -race ./internal/daemon -run 'Test(LoadConfig_CompletedTaskTTLDefaultsDisabledAndReadsEnv|CleanTaskDir_RemovesDirectory|GCWorkspace_CompletedTaskTTLRemovesOpenIssue|ShouldCleanTaskDir_(APIErrorSkipped|CompletedTaskTTLRemovesOpenIssue|CompletedTaskTTLRequiresCompletionTime|CompletedTaskTTLKeepsLocalDirectoryArtifactTTLSeparate|CompletedTaskTTLPreservesActiveEnv))$' -count=20
DATABASE_URL=... REDIS_TEST_URL=... bash scripts/test-go.sh --race
go vet ./...
go build ./...
GOOS=windows GOARCH=amd64 go build -o /tmp/multica.exe ./cmd/multica
```

All commands pass locally. The full race-enabled backend suite ran against pgvector/PostgreSQL 17 and Redis 7.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; daemon-only change)

## AI Disclosure

**Prompt / approach:** I asked Codex to investigate issue #6856 and implement the next contribution only if it remained open and unclaimed. The implementation was rebased onto the latest `main`, audited against the existing GC lifecycle and safety guards, and expanded after review found that the first draft could make `local_directory` artifact cleanup happen earlier than `MULTICA_GC_ARTIFACT_TTL`. Codex then ran focused stress tests, the complete race-enabled backend suite, builds, migration checks, `go vet`, Windows cross-build, and Helm validation before publishing this draft.

## Screenshots (optional)

Not applicable; this is a daemon retention-policy change.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -255,6 +255,7 @@ Daemon behavior is configured via flags or environment variables:
 | GC enabled | — | `MULTICA_GC_ENABLED` | `true` (set `false`/`0` to disable) |
 | GC scan interval | — | `MULTICA_GC_INTERVAL` | `2h` |
 | GC TTL (done/cancelled issues) | — | `MULTICA_GC_TTL` | `24h` |
+| GC completed-task TTL (issue tasks) | — | `MULTICA_GC_COMPLETED_TASK_TTL` | `0` (disabled) |
 | GC orphan TTL (no `.gc_meta.json`) | — | `MULTICA_GC_ORPHAN_TTL` | `72h` |
 | GC artifact TTL (completed tasks) | — | `MULTICA_GC_ARTIFACT_TTL` | `12h` (set `0` to disable) |
 | GC artifact patterns | — | `MULTICA_GC_ARTIFACT_PATTERNS` | `node_modules,.next,.turbo` |
@@ -265,9 +266,10 @@ Daemon behavior is configured via flags or environment variables:
 
 #### Workspace garbage collection
 
-The daemon periodically scans `MULTICA_WORKSPACES_ROOT` and reclaims disk space in four modes:
+The daemon periodically scans `MULTICA_WORKSPACES_ROOT` and applies several disk-reclamation policies:
 
 - **Full task cleanup** — when an issue's status is `done` or `cancelled` and has been idle for `MULTICA_GC_TTL`, the entire task directory is removed.
+- **Completed-task retention bound** — set `MULTICA_GC_COMPLETED_TASK_TTL` to fully remove an inactive issue task once its `.gc_meta.json` `completed_at` age exceeds the configured duration, even while the parent issue remains open. The default `0` disables this opt-in policy. Cleanup waits for a successful parent-issue status check, never removes an active environment, and never fully removes a `local_directory` environment. A later rerun provisions a fresh environment instead of resuming the removed checkout.
 - **Orphan cleanup** — task directories with no `.gc_meta.json` (e.g. left over from a daemon crash) are removed once they exceed `MULTICA_GC_ORPHAN_TTL`.
 - **Artifact-only cleanup** — when a task has been completed for at least `MULTICA_GC_ARTIFACT_TTL` but the issue is still open, regenerable build outputs whose directory basename matches `MULTICA_GC_ARTIFACT_PATTERNS` are removed. The daemon also reclaims the exact managed path `codex-home/.sandbox-bin`; old task metadata without `completed_at` becomes eligible for this managed-only cleanup after its `.gc_meta.json` file has been idle for `MULTICA_GC_ORPHAN_TTL`. The rest of the task (source, `.git`, `output/`, `logs/`, `.gc_meta.json`, Codex auth/config/session state) is preserved so the agent can resume it.
 - **Managed-cache reclamation** — the exact managed path above is reclaimed for *every* task kind once the task has been completed for `MULTICA_GC_ARTIFACT_TTL`, not just for issue tasks whose issue is still open. It applies even while the parent record says the directory itself must stay — an active chat session, a still-running autopilot run — and even when the parent record could not be reached this cycle, because the contents are regenerable and the next run re-provisions them on demand. A task currently running on the directory is never touched. Set `MULTICA_GC_ARTIFACT_TTL=0` to disable this along with the rest of artifact cleanup.
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -67,6 +67,7 @@ const (
 	DefaultMaxConcurrentTasks             = 20
 	DefaultGCInterval                     = 2 * time.Hour
 	DefaultGCTTL                          = 24 * time.Hour      // 1 day — AI-coding issues rarely stay open long
+	DefaultGCCompletedTaskTTL             = 0                   // disabled — operators may bound completed issue-task env retention independently of issue status
 	DefaultGCOrphanTTL                    = 72 * time.Hour      // 3 days — orphans with no meta (crashes, pre-GC leftovers)
 	DefaultGCArtifactTTL                  = 12 * time.Hour      // 12h — drop regenerable artifacts once a task has been completed this long
 	DefaultGCCodexSessionTTL              = 14 * 24 * time.Hour // 14 days — reclaim per-issue Codex session stores untouched this long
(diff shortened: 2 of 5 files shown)


iiwish:
Hi @Bohan-J, gentle ping when you have a chance. This daemon-only GC fix is ready for review and all PR checks are passing. I also rechecked the current patch against the latest `main` today: it merges cleanly, the full daemon race suite passes, and `go vet` is clean. No rush, and thank you for taking a look.

Bohan-J (changes requested):
Thanks for this — the direction is right and I'd like to land it. The problem is real: on `main`, `gcDecisionIssueResult` only selects a full task root for removal when the issue is `done`/`cancelled` and past `MULTICA_GC_TTL`, so a completed task under a still-open issue keeps its checkout and `.git` indefinitely while artifact-only cleanup can't touch them. Making that bound operator-controlled and disabled by default is the right shape, and the safety edges are well chosen — reusing the existing env-root reservation, and excluding `local_directory` in the condition itself rather than relying on `applyLocalDirectoryGCOverride` (which would have demoted to artifact cleanup and made a short completed-task TTL pull `local_directory` artifact cleanup in ahead of `MULTICA_GC_ARTIFACT_TTL`).

One blocking item before merge.

## Blocking: restore the fail-closed status check reverted in `4a449e1`

In `server/internal/daemon/gc.go`, the completed-task condition currently gates on `strings.TrimSpace(result.Status) != ""` as its "we got a real answer from the server" signal.

That gate is weaker than the PR description claims. Every other full-delete path in this file positively matches values it recognizes — `result.Status == "done" || result.Status == "cancelled"` — so an unrecognized status naturally falls through to a non-destructive branch. The new policy inverts that: any non-empty string is treated as a green light, so a status this daemon doesn't recognize becomes a deletion signal rather than a reason to keep the data. Self-hosted daemons routinely run behind the server they talk to, so a server that starts returning a status introduced after the daemon was built, or any drift in the response shape that still yields a non-empty string, would fully remove task roots that the same code would have preserved if it recognized the value. That's the one direction this policy shouldn't fail in.

You already had this fixed: `9c113179e` added `isKnownIssueStatus` plus `TestShouldCleanTaskDir_CompletedTaskTTLUnknownIssueStatusFailsClosed`, and `4a449e1fc` reverted it nine minutes later carrying only the auto-generated revert message. Could you restore that commit, or explain what the revert was for?

If the revert was motivated by the maintenance cost of a hardcoded list — a reasonable concern, especially if the status set is ever meant to be extensible — then the fix is to make the server-side signal explicit rather than to leave the destructive path open. `IssueGCCheckResult` currently carries only the raw `status` string, so there's nothing better for the daemon to key on today; having the gc-check endpoint return a category or a terminal/known flag would be a cleaner long-term answer. Either way, an unrecognized status should skip.

## Non-blocking

**Cycle stats can't attribute the new policy.** `applyGCAction` counts the new removals under `stats.cleaned` together with issue-lifecycle cleanup, so the `gc: cycle complete` line and `multica daemon disk-usage` can't show an operator how much of the reclamation came from the TTL they just enabled. The per-directory logs do distinguish the two, which is good. A separate counter on `gcStats` would close the gap — this is the first thing anyone will want to look at after switching on a destructive knob for the first time.

**Docs could be more specific about the trade.** The "a later rerun provisions a fresh environment" claim checks out — `execenv.Reuse` returns nil when the workdir is gone and the caller falls back to `Prepare`. Worth stating the other half explicitly in `CLI_AND_DAEMON.md`: enabling this drops uncommitted or unpushed branch work along with `output/` and `logs/`, while the per-issue Codex session store lives outside the workspaces root under its own TTL, so resume context survives. That contrast is what an operator actually needs in order to pick a value, and the second half is a much better story than the docs currently give you credit for.

Two smaller things in the same area: it may be worth recommending a floor relative to `MULTICA_GC_INTERVAL`, since the active-root reservation protects an env that is currently running but not one whose follow-up run is queued and unclaimed — a very small TTL costs each follow-up its predecessor's workdir (degradation rather than corruption, but surprising). And since the new check lives in `gcDecisionIssueResult`, the bound applies only to issue-kind environments; chat, autopilot-run and quick-create roots stay unbounded. That's fine to leave out of this PR, but the docs shouldn't imply broader coverage than it has.

Happy to approve once the fail-closed check is back.

iiwish:
Thanks for the detailed review. The revert was a brief reassessment of whether a hardcoded first-party status allowlist was unnecessarily strict; it was not caused by a test or implementation problem. Your point about self-hosted daemon/server version skew is right for an irreversible full-delete path, so I restored the fail-closed status validation and regression test in `95cb4f799`. Unknown or malformed statuses now skip full task-root cleanup. The full daemon test suite, the focused race suite (`-count=20`), and `go vet ./internal/daemon` all pass locally.

Bohan-J (approved):
The fail-closed check is back in `95cb4f799` — `isKnownIssueStatus` plus the unknown-status regression test, matching what was reverted. Thanks for restoring it, and for the note on why the revert happened; weighing a hardcoded list against an irreversible delete is a fair thing to have hesitated over, and I think you landed on the right side of it.

Nice work on this one. The problem was diagnosed precisely — separating "the task finished a while ago" from "the issue reached a terminal state" is exactly the distinction the old policy was missing — and the safety edges were the right ones to pick without being told: reusing the env-root reservation, requiring a successful parent lookup, and excluding `local_directory` in the condition itself rather than leaning on the override. The test coverage around the negative cases is what made this quick to review.

The non-blocking items from my earlier review still stand and are worth a follow-up whenever you feel like it, but none of them should hold this up.

Approving and merging.

PR #6953, opened 2026-08-14
fix(agent): pass bare model id to agy instead of tab-joined id\label

### LINKED ISSUE #6867: [Bug]: Antigravity Runtime failed with -- model value
### Deployment type

self-host

## Summary

The Multica daemon corrupts the `--model` value it passes to the `agy` (Antigravity) CLI, concatenating the model's internal id and display label with a literal tab character (`id\tlabel`). `agy` rejects this malformed value outright, so every agent configured to use an Antigravity/Gemini model fails on every run. The bug persists after updating the daemon to the latest version (`v0.4.24`).

## Environment

- `multica` CLI/daemon version: `0.4.24` (commit `ad23d1da3`, built `2026-08-12T09:52:58Z`)
- Affected runtime: Antigravity (`agy`), provider `antigravity`
- `agy` CLI version: `1.1.12` (harness) / reported `cli_version: 0.4.24` in runtime metadata
- OS: macOS (Darwin, arm64), affected host `macmini.lan`
- Agent affected: a workspace agent ("Integrator") configured via `multica agent update <id> --model <value>`

## Evidence It's Not an `agy` CLI Problem

Ran `agy` directly (same machine, same `agy` binary the daemon shells out to) with the exact flag shape the daemon uses:

```
$ agy -p --model "gemini-3.6-flash-medium" --effort "" --print-timeout 30s "reply with exactly: OK"
I am currently running on **Gemini 3.6 Flash (Medium)**.
...
```

This succeeds with no error. `agy models` also lists `gemini-3.6-flash-medium\tGemini 3.6 Flash (Medium)` as a valid entry. So the CLI correctly accepts the clean bare id — the corruption is introduced somewhere in the daemon's invocation path, not in `agy`.

## Attempted Fixes (did not resolve it)

- Set agent `model` to `Gemini 3.6 Flash (Medium)` (label only) → rejected.
- Set agent `model` to `gemini-3.6-flash-medium` (bare id, confirmed stored correctly) → still fails with the corrupted tab-joined value in the actual `agy` invocation.
- Ran `multica update` and restarted the daemon (`0.4.23` → `0.4.24` for the `agy`-runtime `cli_version` metadata) → bug persists on `0.4.24`.

## Suspected Root Cause

The daemon likely stores/derives Antigravity model options as `(id, label)` pairs and, somewhere in the code path that builds the `agy --model` argument, joins the pair with `\t` instead of using the id alone — possibly leftover debug/table formatting (the `\t`-separated format matches exactly how `agy models` prints its own catalog table). This concatenated string is then passed straight through to `agy`, which correctly rejects it since it doesn't match any of its own model ids.

## Impact

Any agent configured to use an Antigravity/Gemini model fails on every invocation with `exit status 1`, making the Antigravity runtime unusable for agent execution until fixed.

## Steps to Reproduce

1. Set an agent's model via the CLI to a clean, valid model id:
   ```
   multica agent update <agent-id> --model "gemini-3.6-flash-medium"
   ```
   Confirmed via `multica agent list --output json` that the stored `model` field is exactly `gemini-3.6-flash-medium` (no tab, no label).
2. Trigger a run of that agent (Antigravity/`agy` runtime backing it).
3. The run fails immediately.

## Expected Behavior

`agy` is invoked as `agy ... --model "gemini-3.6-flash-medium" --effort "" ...` and the model is accepted (verified working directly via the `agy` CLI — see Evidence below).

## Actual Behavior

The daemon invokes `agy` with a corrupted `--model` value where the stored id and its display label are joined by a tab character:

```
agy exited with error: exit status 1; agy stderr: Error: invalid model selection (--model "gemini-3.6-flash-medium\tGemini 3.6 Flash (Medium)" --effort ""): model gemini-3.6-flash-medium	Gemini 3.6 Flash (Medium) is not recognized as a known model or custom model in settings
Available models:
  Gemini 3.6 Flash (High)
  Gemini 3.6 Flash (Medium)
  Gemini 3.6 Flash (Low)
  Gemini 3.5 Flash (High)
  Gemini 3.5 Flash (Medium)
  Gemini 3.5 Flash (Low)
  Gemini 3.1 Pro (High)
  Gemini 3.1 Pro (Low)
  Claude Sonnet 4.6 (Thinking)
  Claude Opus 4.6 (Thinking)
  GPT-OSS 120B (Medium)
```

This happens regardless of whether the agent's stored `model` field is the bare id (`gemini-3.6-flash-medium`) or the tab-joined form — the daemon appears to reconstruct/re-derive the `id\tlabel` pair internally (likely from its own model-catalog lookup, e.g. joining a `(id, label)` tuple with `\t` when building the CLI args) rather than passing the stored value through unmodified.

A second, related symptom seen mid-investigation: the daemon at one point rejected the *bare* id with a different, daemon-side error (not from `agy` itself):

```
antigravity model "gemini-3.6-flash-medium" is not available from `agy models`; pick one of: gemini-3.6-flash-high\tGemini 3.6 Flash (High), gemini-3.6-flash-medium\tGemini 3.6 Flash (Medium), ...
```
— even though `gemini-3.6-flash-medium` is literally present in that same "pick one of" list. This suggests the daemon's validation step is comparing against, or reconstructing, the tab-joined form rather than the id alone.

### Screenshots (optional)

_No response_

```shell

```

itsXuSt:
## What does this PR do?

Fixes #6867: the Multica daemon corrupted the `--model` value it passed to Google's Antigravity CLI (`agy`), concatenating the model's internal id and display label with a literal tab (`id\tLabel`). `agy` rejects this malformed value, so every agent configured to use an Antigravity/Gemini model failed on every run.

## Root cause

`agy models` outputs one model per line as `id\tLabel` (e.g. `gemini-3.6-flash-medium\tGemini 3.6 Flash (Medium)`). The parser `parseAntigravityModels` treated each entire line as both `ID` and `Label`, so the catalog entry's ID was the full `id\tLabel` string. When the UI model picker stored that as `agent.model`, the daemon passed it verbatim to `agy --model`, which only accepts the bare id.

## Changes

- **`parseAntigravityModels`** (`models.go`): splits each `agy models` line on the tab. The first field becomes the `Model.ID` (what `agy --model` accepts); the second becomes `Model.Label` (for display). Lines without a tab keep the old `id == label` behaviour for backward compatibility with older `agy` versions.
- **`antigravityNormaliseModel`** (`antigravity.go`): strips a tab-joined `id\tLabel` suffix from `opts.Model` at `Execute` time, so agents that already have the corrupted value persisted keep working without a manual re-set.
- **`antigravityModelError`** (`antigravity.go`): accepts a `Label` match too, so a legacy value stored as the display label is not falsely rejected by the up-front validation.
- Comments updated to reflect that `agy --model` takes the bare model id, not the display string.

## Related Issue

Closes #6867

## Type of Change

## How to Test

1. Install `agy` and run `agy models` — observe the tab-separated `id\tLabel` format.
2. Set an agent's model to a bare id: `multica agent update <id> --model "gemini-3.6-flash-medium"`.
3. Run the agent — it should execute successfully instead of failing with `exit status 1`.
4. For a legacy agent whose model was persisted as `gemini-3.6-flash-medium\tGemini 3.6 Flash (Medium)` (by the old parser): the normalisation step strips the tab-suffix at run time, so the agent works without a manual re-set.

Unit tests:
- `TestParseAntigravityModels` — covers tab-separated `id\tLabel`, no-tab lines, duplicates, blanks.
- `TestAntigravityModelError` — covers ID match, Label match (legacy), empty, nil catalog, near-miss rejection.
- `TestAntigravityNormaliseModel` — covers `id\tLabel` stripping, no-tab passthrough, whitespace trimming.

## Checklist

### DIFF
--- a/server/internal/daemon/agents_probe.go
+++ b/server/internal/daemon/agents_probe.go
@@ -211,8 +211,8 @@ var probeAgentCLIs = func() map[string]AgentEntry {
 	}
 	// agy 1.0.6 added a `--model` flag (MUL-3125), so Antigravity now takes a
 	// model env like every other backend. MULTICA_ANTIGRAVITY_MODEL seeds the
-	// daemon-wide default; its value is the exact `agy models` display string
-	// (e.g. "Claude Opus 4.6 (Thinking)"), not a provider/model slug.
+	// daemon-wide default; its value is the bare model id `agy models`
+	// advertises (e.g. "gemini-3.6-flash-medium"), not the display label.
 	if e, ok := probe("MULTICA_ANTIGRAVITY_PATH", "agy", "MULTICA_ANTIGRAVITY_MODEL"); ok {
 		agents["antigravity"] = e
 	}
--- a/server/pkg/agent/antigravity.go
+++ b/server/pkg/agent/antigravity.go
@@ -49,6 +49,12 @@ func (b *antigravityBackend) Execute(ctx context.Context, prompt string, opts Ex
 		return nil, fmt.Errorf("agy executable not found at %q: %w", execPath, err)
 	}
 
+	// Normalise a legacy agent.model value that was persisted as the full
+	// `id\tLabel` line the old parser stored (#6867). agy --model only accepts
+	// the bare id, so strip everything from the first tab onward. A value with
+	// no tab passes through unchanged.
+	opts.Model = antigravityNormaliseModel(opts.Model)
+
 	// Guard against agy's silent no-op on an unrecognised --model: it exits 0
 	// with empty output, which would otherwise surface as a "completed" but
 	// empty task. opts.Model is the single funnel for both agent.model and the
@@ -423,16 +429,16 @@ var antigravityBlockedArgs = map[string]blockedArgMode{
 // buildAntigravityArgs assembles the argv for a daemon-compatible one-shot agy
 // invocation.
 //
-//	agy -p <prompt> --dangerously-skip-permissions [--model <display name>]
+//	agy -p <prompt> --dangerously-skip-permissions [--model <model id>]
 //	    --print-timeout <duration> --log-file <tmp>
 //	    [--conversation <id>] [--add-dir <cwd>]
 //
 // agy 1.0.6 added a `--model` flag (MUL-3125), so opts.Model is now wired
-// through when set. The value is the exact human display string `agy models`
-// prints (e.g. "Claude Opus 4.6 (Thinking)"), NOT a provider/model slug —
-// it's passed verbatim as a single exec arg, so spaces and parens need no
-// shell quoting. agy still exposes no --system-prompt; runtime instructions
-// are delivered via AGENTS.md in the task workdir.
+// through when set. The value is the bare model id that `agy models`
+// advertises (e.g. "gemini-3.6-flash-medium"), NOT the display label —
+// it's passed verbatim as a single exec arg. agy still exposes no
+// --system-prompt; runtime instructions are delivered via AGENTS.md in
+// the task workdir.
 //
 // agy silently no-ops on a model string it doesn't recognise (empty output,
 // exit 0), so Execute validates opts.Model against the `agy models` catalog
@@ -467,13 +473,27 @@ func buildAntigravityArgs(prompt, logPath string, timeout time.Duration, opts Ex
 	return args
 }
 
+// antigravityNormaliseModel strips a tab-joined `id\tLabel` suffix from a
+// model value, returning the bare id that `agy --model` accepts. Values with
+// no tab pass through unchanged. This lets agents whose model was persisted
+// by the old parser (which stored the entire `agy models` line including the
+// tab and label) keep working without a manual re-set (#6867).
+func antigravityNormaliseModel(model string) string {
+	model = strings.TrimSpace(model)
+	if idx := strings.IndexByte(model, '\t'); idx >= 0 {
+		return strings.TrimSpace(model[:idx])
+	}
+	return model
+}
+
 // antigravityModelError returns an actionable error when `model` is non-empty
 // and definitively absent from `available` (the `agy models` catalog); it
 // returns nil otherwise. An empty `available` means discovery couldn't produce
 // a catalog (agy missing, transient failure) — we fail OPEN there and let agy
 // resolve the value, so a discovery hiccup never blocks a run. The match is
-// exact because agy's --model wants the precise display string; a near-miss
-// (extra space, dropped suffix) is correctly rejected since agy would silently
+// against the model ID (the bare value `agy --model` accepts) and, as a
+// fallback for legacy persisted values, the display Label. A near-miss (extra
+// space, dropped suffix) is correctly rejected since agy would silently
 // no-op on it anyway.
 func antigravityModelError(model string, available []Model) error {
 	if model == "" || len(available) == 0 {
@@ -481,7 +501,7 @@ func antigravityModelError(model string, available []Model) error {
 	}
 	ids := make([]string, 0, len(available))
 	for _, m := range available {
-		if m.ID == model {
+		if m.ID == model || m.Label == model {
 			return nil
 		}
 		ids = append(ids, m.ID)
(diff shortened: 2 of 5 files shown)


itsXuSt:
@multica-eve

ADanMan (commented):
Solid, well-targeted fix, and the writeup is exemplary. Root cause is diagnosed correctly: `parseAntigravityModels` was storing the whole `id\tLabel` line as both `Model.ID` and `Model.Label`, so the picker persisted the tab-joined string into `agent.model` and `agy --model` rejected it. Splitting on the tab (`models.go`, `strings.IndexByte(line, '\t')`) so `ID` = bare id and `Label` = display name is exactly right, and it matches the evidence in #6867.

What I liked:

- The parser handles the real edge cases: no-tab legacy lines collapse to `id == label`, an empty id (`\tLabel`) falls back to the label, dedup keys on `id`, and both halves are trimmed.
- The stale comments claiming `--model` wants the display string were updated in lockstep across `models.go`, `antigravity.go`, and `agents_probe.go` — comment rot on this exact point is what encoded the bug, so fixing all of them is the right call.
- Good coverage of tab/no-tab, dedup, blanks, trailing space, and dropped-suffix near-misses; CI is green across all jobs.
- The `agents_probe.go` change is comment-only, which keeps the diff honest.

**Medium — the `Label` fallback in `antigravityModelError` is accepted but never resolved to an id, which is in tension with the function's own comment.**

`antigravityModelError` accepts a value that matches only the display label (`antigravity.go:504`, `if m.ID == model || m.Label == model`), and the comment above it frames this as "a fallback for legacy persisted values … a near-miss … is correctly rejected since agy would silently no-op on it anyway." But `antigravityNormaliseModel` (`antigravity.go:481`) only trims and strips a tab — a pure display label has no tab, so it passes through unchanged — and `buildAntigravityArgs` (`antigravity.go:453`) then sends `opts.Model` verbatim to `agy --model`. So a label-only value is *accepted* by the guard and then handed to agy as the label.

That's the tension: the comment's logic is "reject what agy would no-op on," but by this PR's thesis (and #6867) agy wants the bare id — so a full display label is exactly a value agy would no-op on, yet the `|| m.Label == model` branch lets it through. Two coherent ways to resolve it:

- If the intent is only to keep **tab-joined** legacy values working, that's already fully covered by `antigravityNormaliseModel` + the `m.ID == model` match (`id\tLabel` → `id` → ID hit), so `|| m.Label == model` is dead weight and can go.
- If the intent is to tolerate **label-only** values (or a `MULTICA_ANTIGRAVITY_MODEL` default set to a label under the old docs), resolve the label to its id on a Label hit (e.g. `opts.Model = m.ID`) so what actually reaches agy is a value agy accepts — rather than accept-and-forward a label that, by this PR's own premise, agy silently no-ops on.

Not a blocker — worst case is agy's own error rather than a daemon fault — but there's no test pinning what actually reaches agy on the label path, and the accept-label / send-label behavior contradicts the comment. (I haven't independently confirmed whether current agy accepts a display label; if it does that softens this, but then the guard comment's "agy would no-op" rationale needs a tweak too.)

**Low — the normalise step is a read-time shim, not a data fix.** `antigravityNormaliseModel` cleans the value per run, but the corrupted `id\tLabel` string stays persisted until the picker rewrites it via the new parser. Reasonable for a persisted-data boundary and it's documented — just worth a one-line note that stored values remain in the old form until re-selected.

Nice fix overall — the Medium is really just "pick the intent for the label branch and make it consistent."

itsXuSt:
@ADanMan thanks for the careful review and the concrete suggestions.

I pushed `a0a7b9fe4` to address both points:

- For the Medium item, I split label tolerance from validation. `antigravityResolveModelLabel` now resolves a display-label-only legacy value to the catalog id before validation, so `buildAntigravityArgs` still sends a bare id to `agy --model`. `antigravityModelError` is back to ID-only matching, which keeps the guard consistent with its comment.
- For the Low item, I added the note that `antigravityNormaliseModel` is only a read-time shim; existing corrupted persisted values stay as-is until the picker rewrites them through the new parser.

I also added/updated tests around the label-to-id resolution and the ID-only validation behavior.

multica-eve:
Thank you for the detailed contribution and the thoughtful persisted-value compatibility work. We selected and merged the earlier, narrower fix in #6654, which resolves the same tab-separated catalog-ID root cause. To avoid maintaining two overlapping implementations, we are closing this PR as superseded. The legacy-value normalization idea can be reconsidered as a focused follow-up if a post-fix reproduction shows it is still needed. We appreciate the work you put into this.

PR #6957, opened 2026-08-14
MUL-6178: fix(skills): re-verify permission and staleness inside the refresh transaction

zdavison:
## What

`POST /api/skills/{id}/refresh` (added in #6847) checks `canManageSkill`, then fetches the bundle from the upstream for up to `importFetchTimeout` (45s), then overwrites. Two things can change during that window, and neither was re-checked at the write.

**Permission.** The `AllowOverwrite` closure captured `isAdmin` from the request-time check, so a caller removed from the workspace or demoted mid-fetch still landed the overwrite. Only `isCreator` was re-evaluated in the transaction. Replaced with an `Authz` policy the transaction applies to a membership row it reads itself.

**Staleness.** `UpdateSkill` replaces name, description, content and config outright, and the only in-tx guard was the name comparison. An edit made during the fetch disappeared with no signal — including a change to `config.origin`'s source URL. The request now carries the `updated_at` it read into the transaction as a compare-and-set, returning `409`.

The target read becomes `FOR UPDATE`, so both checks still hold at the `UPDATE`. Under READ COMMITTED a plain read lets a concurrent edit commit in between.

## Tests

Both regression tests park the refresh mid-fetch on a gated ClawHub fixture, mutate the database, then release:

- `TestRefreshSkill_RoleRevokedDuringFetchForbidden` — admin's membership is deleted mid-fetch → `403`, skill unchanged.
- `TestRefreshSkill_ConcurrentEditConflicts` — another member edits the body mid-fetch → `409`, the edit survives.

Each was confirmed to fail against the current behavior before the fix: emulating the pre-fetch verdict makes the first return `200`, and dropping `ExpectedUpdatedAt` makes the second return `200` with the edit overwritten.

## Verification

- `go test ./internal/handler` — full package, against a migrated Postgres
- `go test ./cmd/multica`
- `go vet ./...`
- `sqlc generate` for the new `GetSkillInWorkspaceForUpdate` query

## Notes

Behavior change for API clients: refresh can now return `409` for a reason other than a rename collision. The built-in skill docs (`SKILL.md` and the source map) are updated in the same PR per CLAUDE.md.

The import-overwrite path is untouched — it keeps the creator-only default, now spelled `overwriteAuthzCreatorOnly`.

### DIFF
--- a/server/internal/handler/skill_create.go
+++ b/server/internal/handler/skill_create.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 	skillpkg "github.com/multica-ai/multica/server/internal/skill"
+	"github.com/multica-ai/multica/server/internal/util"
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 )
 
@@ -102,6 +103,23 @@ var (
 	errSkillOverwriteForbidden    = errors.New("not permitted to overwrite target skill")
 	errSkillOverwriteNameMismatch = errors.New("target skill name does not match the imported skill")
 	errSkillOverwriteNameConflict = errors.New("another skill in the workspace already has the imported name")
+	errSkillOverwriteStale        = errors.New("target skill changed since it was read")
+)
+
+// skillOverwriteAuthz names the permission rule overwriteSkillWithFiles applies
+// inside its transaction. It is a policy, not a decision: the tx evaluates it
+// against state it reads itself, so a caller that did slow work (a network
+// fetch) between its own check and this write cannot carry a stale verdict in.
+type skillOverwriteAuthz int
+
+const (
+	// overwriteAuthzCreatorOnly is the default. Only the skill's current
+	// creator may overwrite it — see canOverwriteSkillByLocalImport.
+	overwriteAuthzCreatorOnly skillOverwriteAuthz = iota
+	// overwriteAuthzCreatorOrManager mirrors canManageSkill: the caller must
+	// still be a workspace member, and must be an owner/admin or the skill's
+	// current creator. Membership and role are re-read inside the tx.
+	overwriteAuthzCreatorOrManager
 )
 
 type skillOverwriteInput struct {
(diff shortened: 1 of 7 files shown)


multica-eve (changes requested):
Before this change, a source update could still take effect after the member lost access, or silently erase work a teammate saved while the download was in progress. After this change, those updates are meant to stop and ask the user to review the latest state before retrying. ✅ This is the product behavior we want, and people do not need to change how they work.

## Implementation review

The overall direction is good: carrying an authorization policy into the write transaction, comparing the version read before the fetch, and locking the target row are the right building blocks. I am requesting changes for two correctness gaps:

1. **Must fix: the staleness check does not cover supporting files.** `ExpectedUpdatedAt` only tracks `skill.updated_at`, but `UpsertSkillFile` and `DeleteSkillFile` in `server/internal/handler/skill.go:2350-2414` mutate `skill_file` rows without updating the parent. The refresh path then deletes and rebuilds the full file set in `server/internal/handler/skill_create.go:279-294`. I reproduced a concurrent edit to `notes.md` during the gated fetch: refresh returned `200`, and the member's edit was replaced by the upstream content. Please make every content-changing path, including file upsert/delete, participate in one monotonic Skill revision. Prefer an explicit revision; whichever token is used, file writes should lock/update the parent in the same transaction before changing child rows. Add regression coverage for both file upsert and delete during refresh.

2. **Must fix: the transaction re-reads membership but does not keep that authorization state stable through commit.** `authorizeSkillOverwrite` performs a plain membership read at `server/internal/handler/skill_create.go:167-185`, while the row lock at `:222` covers only the Skill. A demotion/removal can therefore commit after the membership read and before the Skill update, and the stale authorization still lands the write. Because this is an authorization boundary, the short duration of the remaining window is not sufficient protection. Hold an appropriate member-row lock through commit (with a documented consistent lock order), or use another serialization mechanism that proves revocation cannot commit first. Add a deterministic test that revokes the role after the transaction has read it but before the update.

## Recommendation

🔄 **Request changes** — the product direction is correct, but the patch still permits silent supporting-file data loss and a post-check authorization race. Once these two boundaries are closed, the change should be ready to merge.

Validation: the two new targeted tests pass locally and the current CI is green. A review-only supporting-file regression test reproduced the `200`/lost-edit behavior; the temporary test was removed afterward.

PR #6958, opened 2026-08-14
MUL-6177: feat(skills): update selected skills from source in one batch

zdavison:
## What

The skills list already supports add-to-agent and delete on a multi-row selection. Update-from-source (#6847) was row-at-a-time only. This adds it to the batch toolbar, gated on every selected row being editable and carrying a refreshable origin, with a tooltip explaining the disabled state.

## Partial results

Each refresh is its own server transaction, so a batch is not atomic. The dialog settles every row rather than aborting on the first rejection, then reports what actually landed:

- all succeeded → success toast, caches invalidated, selection cleared
- some succeeded → warning toast with the counts, caches invalidated, **selection kept** so the user can retry
- none succeeded → error toast with the server's message, no invalidation

Aborting early would skip the cache invalidation and report "failed" while the earlier skills were already overwritten and their local edits gone — the confirm copy promises to replace the selected skills, so that reads as "nothing happened".

Concurrency is chunked at 4: each row makes the server re-download a bundle from a third-party host.

## Structure

The row kebab now opens the same dialog with a single row, so the confirm copy and the result reporting have one code path. Single-row wording is unchanged — it still uses the existing `detail.refresh.dialog.*` strings. The detail page keeps `RefreshSkillDialog`, which additionally seeds the returned skill into the detail cache.

New i18n keys in all four locales, placed next to the delete-batch block they mirror.

## Tests

`refresh-skills-dialog.test.tsx` covers all-success, partial failure, continuation across a chunk boundary after an early failure, all-fail, and the single-row copy. The three failure-path cases were confirmed to fail against an abort-on-first-rejection loop.

## Verification

- `pnpm test` — 4061 tests, all packages
- `pnpm typecheck` — 6/6
- `pnpm lint` — no new warnings on the changed files


Bohan-J (changes requested):
Thanks for this — the batch flow is well built, and the part I was most likely to argue with is the part you got right.

The settle-every-row-then-report decision is correct, and the reasoning in the PR description holds: aborting on the first rejection would skip the cache invalidation the finished overwrites need and report "failed" while earlier skills were already replaced. I checked that the new tests actually guard that contract rather than just describing it — replacing the chunk loop with an abort-on-first-rejection implementation makes `reports a partial result and still invalidates when one row fails` and `processes every row across chunk boundaries despite an early failure` fail immediately. (Minor correction to the PR description: the all-fail case still passes under an aborting implementation, so it's two of the three, not three.)

I also ran the package suite (`packages/views`, 62 tests) and `tsc --noEmit` locally against this branch — both clean, matching CI.

A few things to fix before this goes in.

## Must fix

**1. The batch dialog shows a single-skill warning for N skills.**

`packages/views/skills/components/skill-list-actions.tsx:512` renders `detail.refresh.dialog.warning` unconditionally, but that string was written for the one-skill dialog and is singular in every locale:

- en: `Local edits to this skill will be overwritten. ...`
- zh-Hans: `对这个 skill 的本地修改将被覆盖。...`
- ja: `このスキルへのローカル編集は上書きされます。...`
- ko: `이 스킬에 대한 로컬 편집 내용은 덮어쓰기됩니다. ...`

So selecting 12 skills produces a dialog titled "Update 12 skills from source?" with a warning that says "this skill" directly underneath it. I reproduced this in the browser in both en and zh-Hans.

Worth noting why the same reuse is fine next door: `DeleteSkillsDialog` reuses `detail.delete_dialog.warning`, which is "This action cannot be undone." — count-neutral. This one isn't, so it needs the same `single`-branch treatment the title and description already get. Suggest adding `actions.refresh_dialog_warning_other` (all four locales) and switching on `single`, exactly like `refresh_dialog_title` / `refresh_dialog_desc`.

**2. `handleConfirm` has no `try`/`finally`, and the dialog cannot be dismissed while `refreshing` is true.**

`setRefreshing(false)` is only reached on the straight-line path. `Promise.allSettled` doesn't reject and `api.refreshSkill` is `async`, so this isn't reachable today — but if anything between the loop and that call throws, `refreshing` stays `true` forever, and `onOpenChange` is gated behind `if (!refreshing)`, so the dialog becomes unclosable and the user has to reload the page. `DeleteSkillsDialog` already wraps its equivalent in `try`/`finally`; matching it costs nothing and removes the failure mode.

## Strongly recommended

**3. A partial failure doesn't say which rows failed, and the retry it invites re-refreshes the ones that succeeded.**

`refresh_partial_toast` reports "Updated 5 of 6 skills; 1 failed" and the selection is deliberately kept so the user can retry — but the kept selection is the *whole* batch. Pressing Update again re-downloads and re-overwrites the five that already landed. Since the operation destroys local edits, that's a real cost, not just wasted requests.

Suggest having the dialog hand the failed ids back (e.g. `onRefreshed?.(failedIds)`) so the page can narrow the selection to just those, and naming the failed skills in the toast. Right now the only way to tell which ones failed is to compare the `Updated` column timestamps.

## Nits (non-blocking)

4. `actions.update` ("Update") duplicates the existing `actions.refresh` ("Update from source"), which is the same action and is what the row kebab already says. Reusing it would avoid a new key in four locales, and "Update" on its own — sitting between "Add to agent" and "Delete" — reads a little like "edit".
5. `RefreshSkillDialog` and `RefreshSkillsDialog` differ by one letter, live in two files, duplicate most of the dialog markup, and invalidate differently (`exact: true` vs not — both correct for their context, but the difference isn't written down anywhere). Relatedly, `refresh-skills-dialog.test.tsx` tests a component that lives in `skill-list-actions.tsx`; there is no `refresh-skills-dialog.tsx`. Moving the new component next to the old one (or having the detail page adopt the shared dialog via a `Skill[]` callback and dropping the old path) would settle both.
6. No progress feedback on a long batch: at concurrency 4, a 50-row selection is 13 sequential rounds behind a single spinner, with the dialog non-dismissible throughout. An `n/total` in the confirming state would help.
7. `(rejected[0] as PromiseRejectedResult | undefined)?.reason` can drop the cast with a type predicate: `results.filter((r): r is PromiseRejectedResult => r.status === "rejected")`.

Happy to re-review as soon as 1 and 2 are in.

multica-eve:
Thanks for the contribution. Bulk updating selected imported skills from source has shipped in #6930, building on the single-skill refresh introduced in #6847 and the provenance UI in #6928. Closing this PR as superseded by the shipped implementation.

PR #6975, opened 2026-08-14
MUL-6193 Support Slack message attachments in chat

tim-smart:
Slack messages with file attachments now surface those files as regular chat attachments.

## What changed

- Capture downloadable Slack file metadata from message and app-mention events.
- Download private files with the installation bot token, restrict requests and redirects to Slack-owned HTTPS hosts, and persist uploads through the media intent ledger.
- Wire the resolver only when object storage is configured.
- Reject unexpected HTML login responses unless the file explicitly declares HTML.
- Add `files:read` to localized Slack setup manifests and document reinstall requirements.

## Validation

- `go test ./internal/integrations/slack/... -count=1`
- `go vet ./internal/integrations/slack/...`
- `go build ./...`
- `pnpm --filter @multica/docs typecheck`
- Localized manifest YAML parity check

## QA

This change has not had QA testing.


Bohan-J:
Thanks @tim-smart — this is a well-built change. The intent-ledger-before-upload ordering, the Slack-only download client, and the HTML-login-page guard all made the review easy, and the per-file independence means one bad attachment can't take down the rest of a message.

I pushed one commit on top rather than sending you round-tripping on nits. What it changes:

- **Restore the bot token on permitted redirects.** `net/http` forwards `Authorization` only to the same host or a subdomain of it, so a hop to a sibling like `files-origin.slack.com` would arrive unauthenticated — Slack answers with its HTML login page, the guard correctly rejects it, and the attachment disappears with only a server-side warn. The redirect policy has already proved the destination is Slack-owned at that point, so it now re-attaches the header. This was the one finding that could have made the feature look silently broken in production.
- **Filter unfetchable files at translation time.** Remote `external` files (Google Drive, Dropbox) point `url_private` at the third-party host, so `HasMedia` was promising media the resolver could never bind — the message paid for a deferred run and an intent-ledger row per doomed download. `rawFilesFrom` now keeps only HTTPS URLs on Slack hosts.
- **Share the media deadline across files.** Ten files at a flat 30s each is up to 5 minutes against the Router's 45s budget, and the Router drops *every* ref once that budget expires — so one slow file cost the whole message its attachments. The per-file budget is now derived from what's left, and the loop stops once the deadline is gone.
- **Reject oversized files on their declared size**, before they cost an intent row and a full transfer, since Slack states the size up front. Also corrected the limit wording to MiB.

Plus regression tests for the redirect policy (including that the token *is* restored on permitted hops and that an off-domain hop is refused), the size boundary, the spent budget, and the off-Slack filter — and I switched the `file_share` translation test to drive `slackevents.ParseEvent` with a real payload. The previous version hand-built `MessageEvent.Message`, which meant it would have kept passing if slack-go ever stopped normalizing new messages into that field — and that normalization is what the whole feature rests on.

Two things I deliberately left alone, both worth a follow-up rather than more churn here:

- The download buffers the whole file in memory; the Feishu resolver already streams through `UploadStream`, which both storage backends implement. That's also why the cap sits at 20 MiB while web uploads allow 100 MB.
- `cleanFileName` / `mediaExtension` / `safeMediaSegment` now exist in both the Slack and Feishu media paths and have already drifted slightly. Worth extracting into the shared channel layer.

Verified locally: `go build ./...`, `go vet`, and `go test -race` over the Slack and channel/engine packages all pass.

One caveat worth stating plainly: I could not exercise a live Slack workspace, so whether `url_private_download` actually issues a redirect is still unverified. Everything here is fail-safe if it does, but a real smoke test over DM, channel mention, an attachment-only message, and a multi-file message with one failure is still worth doing.

PR #6982, opened 2026-08-14
feat(agent): drive ZCode through the native app-server session protocol

tornado404:
## What does this PR do?

Replaces the previous `zcode --prompt --json` / `--prompt --stream-json` CLI backend with a native **`zcode app-server`** backend, mirroring the Codex app-server architecture. This supersedes the approach in #6755: it needs no `--stream-json` injection into the vendored runtime and works with the **unofficial [`kingsword09/zcode-cli`](https://github.com/kingsword09/zcode-cli)** out of the box — no fork, no third-party bridge, no extra npm dependency to install or keep in sync with the runtime.

The backend drives the JSON-RPC session protocol over NDJSON stdio:

```
zcode app-server
→ session/create | session/resume
→ session/subscribe
→ session/send
→ consume session/event (model.streaming deltas, tool.updated, turn.completed/failed)
→ session/close
```

**Why this approach:** ZCode sessions live only in the app-server process memory and cannot be resumed across processes, so the backend keeps **one long-lived app-server process per runtime** in a package-level registry (keyed by `RuntimeID`) and reuses it across turns — multi-turn tasks keep their conversation. Reverse requests (`session/requestRuntimePreferences`, `interaction/*`) are auto-answered, and the consume loop polls `session/read` as the source of truth because the runtime can silently drop notifications.

Runtime discovery, the `327_runtime_profile_add_zcode` migration, model catalog, and UI changes from #6755 are kept.

**Hardening from production use** — edge cases found running real tasks through a desktop daemon against `zcode-app-cli 3.7.6-12`:

- a resumed session whose stored model left the runtime's catalog fails `session/send` with `-32031`; the backend abandons that session and retries the turn on a fresh one instead of failing the task
- the timeout/abort recovery `session/read` carries its own deadline — a hung app-server can no longer block the turn goroutine forever
- `model.streaming` alone arms the `session/read` poll fallback, so a dropped `turn.started` still recovers the terminal state instead of waiting out the semantic inactivity timer
- **tool call streaming**: `tool.updated` lifecycle events map onto the daemon's `tool_use`/`tool_result` messages (phase shapes verified against the real runtime), so the UI shows live tool activity and the daemon's in-flight tool watchdog window works
- nil-`Config.Logger` guards on the fallback/turn-finished paths (the backend can be constructed without `agent.New`'s default)

## Related Issue

Supersedes #6755 (the `--stream-json` fork-based approach).

## Type of Change

## Changes Made

- `server/pkg/agent/zcode_appserver.go` (new): app-server JSON-RPC transport + long-lived process registry + session lifecycle + event→Message conversion (`model.streaming` text/thinking deltas, `tool.updated` tool_use/tool_result) + `session/read` poll fallback + `-32031` fresh-session retry.
- `server/pkg/agent/zcode_appserver_test.go` / `zcode_appserver_integration_test.go` (new): fake-app-server unit tests (launch contract, streaming, poll fallback, bounded recovery, nil-logger retry, registry reuse) + real-binary smoke test (behind `agentintegration`, gated on `MULTICA_RUN_REAL_AGENT_SMOKE=1`).
- `server/pkg/agent/agent.go`: factory → `zcodeBackend`, launch header → `zcode app-server`.
- Deletes the old `zcode.go` / `zcode_stream.go` / `proc_cancel.go` and their tests.
- `server/migrations/327_runtime_profile_add_zcode.*`: protocol-family whitelist migration (renumbered 319 → 327 after main took 319 for `remote_mcp_plugin_v1`).
- Landing copy synced in all four locales (22 → 23 tools) and `zcode` added to the provider metric label whitelist, per the new-runtime checklist.

## How to Test

1. `cd server && go test ./pkg/agent/ -run TestZcode -count=1` — unit tests against a fake app-server.
2. Real end-to-end (uses the installed zcode + consumes a little quota):
   `cd server && MULTICA_RUN_REAL_AGENT_SMOKE=1 go test -tags=agentintegration ./pkg/agent/ -run TestZcodeRealAppServerSmoke -count=1 -v`
   Verified against `zcode-app-cli 3.7.6-12` (official `kingsword09/zcode-cli`): a real turn returns "pong" and a follow-up turn resumes the same in-memory session.
3. **Incremental streaming verified on the real runtime**: a 300-word generation streamed 2 reasoning deltas + 23 text deltas at ~260 ms intervals through `model.streaming` → `Message{Text,Thinking}`, well ahead of `turn.completed`. A tool-forcing turn surfaced a matched `tool_use`/`tool_result` pair (`Bash` + call id + output) mid-turn — also asserted by the smoke test's third turn.
4. **Production daemon verification**: a desktop daemon built from this branch ran real tasks. A task resuming a session whose stored model was gone (`-32031`) retried on a fresh session and completed with streamed output; the daemon-side model discovery round-trip (`~/.zcode/cli/config.json` catalog → runtime models API → UI) reports `bigmodel/GLM-5.3` as the default model.

## Checklist


multica-eve:
Thank you for the detailed implementation and for validating the native app-server path against real ZCode workloads. The Multica team will own and implement official ZCode runtime support directly, so we're closing this community PR. We appreciate the work and the protocol findings you've shared—they'll be useful reference for the official implementation.

multica-eve (changes requested):
Before this change, a Multica user cannot choose ZCode when creating an agent. After it, a user can install ZCode, select it in Multica, and see task progress and tool activity; however, every follow-up task starts without the previous conversation, the model remains controlled outside Multica, and Multica-managed external tools are unavailable. **Product verdict: ZCode support is desirable, but this user journey is not complete enough for a first-class supported agent.**

## Implementation review

There are several merge blockers on the current head:

1. **The branch does not compile.** `server/pkg/agent/models.go:253-254` references the removed `executablePath` local, and `server/pkg/agent/zcode_appserver.go:926` passes a `string` to the new `ListModels(..., Command)` API. Running `go test ./pkg/agent/ -run TestZcode -count=1` fails at build time.
2. **The migration is unsafe against current main.** Prefix `327` is already used by `327_workspace_share_link`, so `go test ./internal/migrations/ -count=1` fails. More importantly, both `327_runtime_profile_add_zcode.up.sql` and its rollback omit the already-supported `dsh` family; applying this migration would prevent creation of new DSH runtime profiles.
3. **Each task can leave an unreaped child process.** `newZcodeClient` starts an `exec.Cmd`, but neither `zcodeClient.close` nor any other path calls `cmd.Wait()`. Waiting only for stdout EOF does not reap the OS process; `WaitDelay` also has no effect without `Wait`. This is a daemon-lifetime resource leak.
4. **Conversation continuity is explicitly lost across tasks.** `server/pkg/agent/zcode_appserver.go:608-637` starts a new app-server for every task while sessions live only in that process. The continuity notice is only a degradation message; it does not restore prior conversation state. This is a material capability gap compared with the alternative PR.

The direct protocol work and real-runtime probes are useful, but maintaining a second ~1,000-line transport while still losing resume, model control, and managed MCP gives us more code with fewer user capabilities than #6987.

## Recommendation

❌ **Reject for technical reasons and continue with #6987.** This is not a rejection of ZCode support; it is a choice of the implementation that has the smaller maintenance surface and more complete task lifecycle.

multica-eve:
Closing in favor of #6987 after reviewing both implementations. ZCode support remains accepted as a product direction; #6987 provides the more complete lifecycle with a smaller provider-specific maintenance surface. The detailed technical comparison is in the review above.

Issue #5887, opened 2026-07-24
Feishu group bot does not pass ZIP attachments to bound agents


Asherggg:
## Problem

When a user sends or references a product image ZIP file in a Feishu group chat and triggers a bound bot/agent workflow, the agent receives the text task but cannot access the actual ZIP file.

## Steps to reproduce

1. Send or reference a product image ZIP file in a Feishu group chat where a Multica bot is bound.
2. In the same conversation, trigger the bot to run the main-image processing workflow.
3. Inspect the agent output and its temporary working directory.

## Actual result

- The agent is triggered successfully and the task is marked as `completed`, with no runtime error.
- The referenced attachment only appears in the agent context as `[File]`.
- The agent does not receive a usable filename, attachment URL, attachment identifier, or local file.
- No ZIP file exists in the task working directory, so the agent cannot extract, validate, or repackage the image assets.
- Two related runs on 2026-07-24 both reported that the current message did not carry or reference an accessible file.

## Expected result

Attachments from both the triggering Feishu group message and any referenced/replied-to Feishu message should be synchronized as Multica Chat attachments for the agent task.

At minimum, the task should receive:

- a readable local ZIP file in the working directory,
- the original filename,
- and attachment reference metadata that lets the agent trace the input.

## Impact

This blocks Feishu group based ZIP workflows for bound agents. Users must upload the ZIP through Multica Web Chat instead, so the Feishu group integration cannot complete the expected end-to-end flow.

## Suggested investigation

Please check the mapping and download path from Feishu group messages, referenced messages, and file attachments into Multica Chat attachments. The task trigger should make the ZIP available as a normal task attachment to the agent.

Bohan-J:
Linking this to #5580, which just merged: it adds the inbound Feishu/Lark media pipeline (scoped message-resource download → object storage → attachment bound to the user chat message).

**Deliberately keeping this open — your case is not fixed by it.** That PR covers standalone `image`, `media`/`video`, and rich-text `post`-embedded `img`/`media` spans. A ZIP arrives as msg_type `file`, and `mediaResourcesFromMessage` returns nil for `file` and `audio`, so a ZIP will still surface as a bare `[File]` placeholder with no attachment — exactly what you reported.

[...]

PR #6984, opened 2026-08-14
fix(lark): ingest standalone file and audio messages (MUL-6198)

xiangboit:
## What does this PR do?

Feishu can deliver files and audio as standalone message types. Multica previously normalized those messages to `[File]` / `[Audio]`, but did not extract their message resources. As a result, a bound agent could receive the user's instruction while the actual business document or archive never reached the task attachment pipeline.

This is especially visible in Feishu-first workflows such as sales and marketplace integration support: users naturally send HTML specifications, ZIP archives, or audio directly to the Bot in a P2P chat instead of uploading them again through Multica Web Chat.

The change routes standalone `file` and `audio` messages through the existing `extractMediaResources` pipeline. That keeps downloading, metadata handling, storage, and task attachment behavior consistent with media embedded in other supported Feishu message types.

Risk and scope:

- No authentication, authorization, UI, or database behavior changes.
- Existing media download and attachment ingestion code is reused rather than introducing a second file path.
- This PR covers standalone file/audio message resources in direct (P2P) chats. Standalone file/audio messages in group chats carry no @ mention and are dropped by the existing group-mention filter before media ingestion, so group ZIP workflows remain outside this PR. Attachments discovered only through referenced or replied-to messages also remain outside this PR; #5887 should stay open.

## Related Issue

Related to #5887

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Update `server/internal/integrations/lark/media_ingest.go` to extract resources from standalone Feishu file and audio messages.
- Add focused coverage in `server/internal/integrations/lark/feishu_channel_test.go` for HTML files, ZIP archives, and audio resources.
- Verify that normalized message text and attachment metadata are both preserved.

## How to Test

1. Run `go test ./internal/integrations/lark -count=1` from `server/`.
2. Send a standalone HTML file or ZIP archive in a direct (P2P) Feishu chat with the bound agent.
3. Verify that the task contains a real attachment with the original filename instead of only the `[File]` placeholder.

Local verification completed:

- Affected Go integration package passed.
- The broader affected Go package set (`channel/engine`, `lark`, `handler`, and `service`) passed.
- A real standalone Feishu HTML file was downloaded, stored as a task attachment, and read by the agent end to end.
- `git diff --check` passed.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable: no UI change)
- [x] I have updated relevant documentation to reflect my changes (not applicable: existing attachment behavior is restored)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable)
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (not applicable)

## AI Disclosure

**Prompt / approach:**

Codex was used to trace the Feishu event-to-task attachment flow, compare standalone file/audio handling with the existing media resource pipeline, implement the smallest compatible change, add focused regression tests, and review the final diff for scope and security regressions. The behavior was also verified through a real Feishu file message before submission.

## Screenshots (optional)

Not applicable; this change has no UI surface.

### DIFF
--- a/server/internal/integrations/lark/feishu_channel_test.go
+++ b/server/internal/integrations/lark/feishu_channel_test.go
@@ -257,7 +257,7 @@ func TestFeishuMediaResolver_HasMedia(t *testing.T) {
 		{"text", InboundMessage{MessageID: "om_t", MessageType: "text", Body: "hello", Content: `{"text":"hello"}`}, false},
 		{"image", InboundMessage{MessageID: "om_i", MessageType: "image", Body: "[Image]", Content: `{"image_key":"img_k"}`}, true},
 		{"video", InboundMessage{MessageID: "om_v", MessageType: "media", Body: "[Video]", Content: `{"file_key":"file_k"}`}, true},
-		{"file", InboundMessage{MessageID: "om_f", MessageType: "file", Body: "[File: spec.html]", Content: `{"file_key":"file_k","file_name":"spec.html"}`}, true},
+		{"file", InboundMessage{MessageID: "om_f", MessageType: "file", Body: "[File]", Content: `{"file_key":"file_k","file_name":"spec.html"}`}, true},
 		{"audio", InboundMessage{MessageID: "om_a", MessageType: "audio", Body: "[Audio]", Content: `{"file_key":"audio_k"}`}, true},
 		{"post with image", InboundMessage{MessageID: "om_p", MessageType: "post",
 			Content: `{"content":[[{"tag":"img","image_key":"img_post"}]]}`}, true},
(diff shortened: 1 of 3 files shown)


Bohan-J:
Thanks for the contribution — the approach here is right. Routing standalone `file` / `audio` through the existing `extractMediaResources` pipeline reuses the download cap, timeout, streaming upload, filename sanitisation, intent ledger and failure-recovery path instead of adding a second ingest route, and the change is appropriately small. I found no blocking correctness issue, and CI is green.

Two things before we merge.

## 1. Please verify the audio half on a real machine

The PR description documents an end-to-end check for a real Feishu HTML file, but nothing for audio — and the audio half is also the part with no automated coverage (see nit 1). Since audio is half of what this PR claims to fix, could you send a real Feishu voice message to a bound agent and attach evidence — a short screen recording, or at minimum a screenshot of the resulting task attachment showing its final filename?

The specific thing worth confirming is the **filename the voice attachment lands with**. Feishu voice is OPUS, and `mediaExtension` (`server/internal/integrations/lark/media_ingest.go:403`) has no audio entry — `mime.ExtensionsByType("audio/opus")` returns an empty list even on a host that ships `/etc/mime.types`, and in a slim container image (the case the existing `application/pdf` comment in that same function explicitly calls out) no audio type resolves at all. So unless Feishu's resource download response carries a `Content-Disposition` filename, the attachment will land as `feishu-audio-om_xxx` with **no extension**, which makes it harder for an agent to recognise the format.

Your current tests can't answer this, because the fake download response supplies `Filename: "voice.ogg"` directly. A real run will settle it. If it turns out the extension is indeed missing, please pin the common audio types (`audio/opus`, `audio/ogg`, `audio/amr`, `audio/mpeg`) into the `mediaExtension` switch, for the same reason `application/pdf` is already pinned there.

## 2. Nits to clean up

1. **The `feishu-audio` branch this PR adds is never executed by any test.** Running the whole `lark` package with a coverage profile, `server/internal/integrations/lark/media_ingest.go:377-378` (`case channel.MsgTypeAudio: prefix = "feishu-audio"`) has an execution count of `0`. `mediaFilename` returns at the first candidate because the fake download response provides a filename. Please add a case where the download response carries **no** filename, so the generated-name path is actually asserted. (For fairness: the `MsgTypeVideo` branch is uncovered too — that's pre-existing, not yours.)

2. **A test fixture contradicts production behaviour.** `server/internal/integrations/lark/feishu_channel_test.go:260` and the new table use `Body: "[File: spec.html]"` / `"[File: integration.html]"`, but `flattenContent` returns plain `"[File]"` for a file message (`server/internal/integrations/lark/content_flatten.go:40`, pinned by `content_flatten_test.go:83`). The new test never asserts `Body`, so the wrong value can't fail — it will just mislead the next reader into thinking we render `[File: <name>]`. Please use `[File]`.

3. **`mime_type` is not a real field on a Feishu file message body.** The fixtures would be closer to the official payload shape without it.

4. **Stale doc comment.** `server/internal/integrations/lark/media_ingest.go:52` still reads "standalone image/video or post-embedded img/media spans" — please fold file/audio into it.

5. **Please state the reachable scope in the PR description.** A standalone file message in a group chat does not itself carry an `@`, so it is dropped by the group-mention filter at `server/internal/integrations/channel/engine/router.go:311`. What this PR actually closes the loop on is **direct (P2P) chat**. Worth being explicit so nobody reads this as "group ZIP workflow now works", and it also confirms your description's existing note that #5887 should stay open.

6. Optional, take it or leave it: `media_ingest.go:292-295` derives `kind` with `if lm.MessageType == "audio"`, but `channelMsgType` (`feishu_channel.go:154`) already maps both types. `kind := channelMsgType(lm.MessageType)` would be shorter — though the existing image/video cases hardcode it too, so staying consistent is a defensible choice.

None of these change the design; once the audio path has a real run behind it and the nits are cleaned up, this is good to merge.

xiangboit:
Thanks for the detailed review. I addressed all requested nits in `3f1e12f8a` and ran the audio half against a real Feishu P2P voice message on the self-hosted Docker deployment.

Real-run result after rebuilding the backend from this branch:

- Feishu download response supplied an extensionless random filename and generic `audio/octet-stream` (matching the concern in the review).
- Multica stored the attachment as `vRUzkW0NQi.opus` with content type `audio/opus`.
- The persisted file is 7,714 bytes; independent `file` / `ffprobe` checks identify it as Ogg Opus, mono, 16 kHz input, duration 3.4865 s.
- The chat message was recorded as `[Audio]`, the task completed, and the assistant reply was delivered back to the same Feishu P2P chat.

The follow-up also adds regression coverage for both the generated-name path and the actual observed Feishu shape (extensionless filename + `audio/octet-stream`), preserves existing audio extensions, and pins `audio/opus`, `audio/ogg`, `audio/amr`, and `audio/mpeg`. I also corrected the file fixtures/comment and clarified P2P scope in the PR body.

All PR checks are now green, including `backend-tests` and the vulnerability scan.

Bohan-J (approved):
Thank you — this is a genuinely well-executed contribution, and the follow-up raised the bar rather than just ticking boxes.

A few things worth calling out:

- You took the audio verification request seriously and came back with real evidence — byte size, `ffprobe` codec/duration, and the actual stored filename. I traced your reported output back through `mediaContentType` → `mediaFilename` and it reproduces exactly, which is the best kind of verification report.
- Better still, you turned what you observed on the real deployment (an extensionless filename plus the generic `audio/octet-stream`) into a regression test instead of just describing it in a comment. That shape is now pinned.
- Both new helpers are correctly gated on `kind == channel.MsgTypeAudio`, so image/video/file behaviour is untouched, and you resisted the tempting-but-wrong approach of sniffing the response body to detect the format — that would have consumed bytes from the single-use stream feeding `UploadStream`.
- The added assertion on the *upload* metadata (filename and content type reaching the object store), not just the resulting `MediaRef`, was not requested and is a real improvement.

Verified locally on `3f1e12f8a`: `go vet` and the full `lark` package pass, the previously-uncovered `feishu-audio` branch now executes, and a trial merge with current `main` builds and passes both `lark` and `channel/engine` — no drift from the 47 commits that landed meanwhile. All 13 CI checks are green.

Merging now. Since this closes the loop for direct chats only, #5887 stays open for the group-chat and referenced-message cases, as your description correctly notes.

PR #6986, opened 2026-08-14
MUL-6199 feat(lark): add per-agent same-tenant guest access

xiangboit:
## What does this PR do?

Multica currently requires every person who talks to a Feishu-bound agent to first create a Multica account, join the workspace, and complete account binding. That is the right default for development agents, but it makes consultation bots impractical: sales, operations, and other same-company users often need to ask a narrowly-scoped agent a question in Feishu without becoming Multica workspace members.

This PR adds an explicit, per-agent option to allow users from the same Feishu tenant to invoke that one bound agent without Multica login or account binding. The existing membership-only behavior remains the default for all current and new installations.

The feature is intentionally fail-closed and does not turn the Bot into a public endpoint:

- The option is disabled by default and must be enabled for each Feishu-bound agent by the agent owner or a workspace owner/admin.
- Guest messages are accepted only when the authenticated event proves that the sender belongs to the installation's Feishu tenant. Missing or mismatched tenant identity is rejected.
- A guest run never inherits the Bot installer's personal connected-app / Composio overlay.
- The deny decision is persisted on the task and preserved across `/issue`, automatic retries, and batched chat messages.
- If guest and member messages enter the same debounce batch, the batch keeps the stricter guest policy regardless of arrival order.

This enables restricted consultation agents while preserving an important warning: a guest can still invoke the agent's ordinary configured workspace capabilities, including assigned repositories, workspace MCP servers, and the selected filesystem mode. Owners should enable it only for agents whose instructions and capabilities are safe for the intended same-tenant audience.

## Related Issue

No existing issue. This PR includes the product motivation, security contract, UI control, tests, and documentation as one reviewable proposal.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `workspace_members` / `feishu_users` inbound access modes to each Feishu installation, with unknown and missing values normalized to the fail-closed `workspace_members` default.
- Validate same-tenant guest identity in the Feishu resolver while keeping cross-tenant and unverifiable senders rejected.
- Carry an explicit external-sender policy through the channel router, chat task enqueueing, `/issue`, and retry paths.
- Persist an empty connected-app list as a durable deny marker so retries cannot later reconstruct the installer's personal connected apps.
- Make the channel batcher preserve the strictest policy when member and guest turns share a debounce window.
- Add an owner/admin-authorized PATCH API and an Agent Integrations switch with clear security guidance.
- Document first-use behavior and guest-access risks in English, Simplified Chinese, Japanese, and Korean.
- Add regression coverage for defaults, API validation/authorization, tenant isolation, task capability isolation, retries, and both mixed-sender batch arrival orders.

## How to Test

1. Bind a Feishu PersonalAgent Bot to a restricted test agent. Verify that an unbound sender receives the existing account-binding response while the option is disabled.
2. Open **Agent → Capabilities → Integrations → Feishu**, enable **Allow Feishu users without Multica login**, and verify the setting persists after reload.
3. From an unbound user in the same Feishu tenant, `@` the Bot and verify that the agent can answer without Multica registration.
4. Verify that a sender from another tenant, or an event without sufficient tenant identity, is rejected.
5. Verify that the guest task has no installer personal connected-app overlay, including after `/issue`, retry, and mixed guest/member batching.

Local verification completed:

- `go test ./internal/integrations/channel/engine ./internal/integrations/lark ./internal/handler ./internal/service -count=1`
- Core Vitest suite: 124 files / 1,401 tests passed.
- Focused Agent Integrations UI suite: 25 tests passed.
- `@multica/core` and `@multica/views` TypeScript typechecks passed.
- The real local Agent Integrations page was inspected with the option enabled; the screenshot below shows the resulting UI and warning copy.
- `git diff --check` passed.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (not attached: the verified self-hosted screen contains private workspace, agent, and runtime identifiers)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable: no new runtime, coding tool, or tab)

## AI Disclosure

**Prompt / approach:**

Codex was used to trace Feishu identity resolution through chat, issue creation, task enqueueing, connected-app overlay construction, retries, and debounce batching. The implementation followed the existing channel resolver and installation settings architecture, added fail-closed tests at each security boundary, and then received a separate diff-focused review. That review found a mixed guest/member batching edge case; the final commit fixes it and tests both arrival orders.

## Screenshots

A real self-hosted Agent Integrations screen was inspected during verification, but it is not attached because it contains private workspace, agent, and runtime identifiers. The switch states and warning copy are covered by the focused 25-test UI suite.


xiangboit:
Synced again to the latest v0.4.32 `main` at `fbbe69757`; PR head is now `41514c5e7`. All four guest-access commits reapplied without conflicts, and `git range-diff` confirms their semantics are unchanged.

Latest verification evidence:
- full Go suite with Go 1.26.1: passed
- TypeScript typecheck: 7/7 tasks passed
- `@multica/core`: 134 files / 1591 tests passed
- `@multica/views`: 393 files / 4648 tests passed
- core/views lint: 0 errors; 28 unrelated upstream warnings
- production Docker backend and frontend builds: passed
- self-host migrations through 397 applied; frontend, backend health, and Tailscale endpoint return HTTP 200
- Lark WebSocket connected; local daemon upgraded to `v0.4.32-7-g41514c5e7` with Codex and Claude runtimes
- real same-tenant unbound-user text message previously completed through `owner_fallback` without auth or tenant errors
- GitHub CI: all applicable checks passed, including backend, frontend, mobile, sqlc, vulnerability scan, installers, and Windows execenv

The security contract remains opt-in per installation: same-tenant enforcement plus a durable deny marker prevents guest-triggered tasks, retries, and mixed-sender batches from inheriting personal connected-app overlays.

xiangboit:
Refreshed against current `main` (`0a54725fe61a`) and completed an additional cleanup/review pass in `232271b665fd`.

The follow-up removes an unused sender field, replaces a variadic security-policy option that could silently ignore extra values with an explicit option, fixes the API client formatting, and adds a real DB-backed PATCH test covering agent-owner/workspace-owner authorization, unrelated-member denial, invalid-mode rejection, response shape, and persisted policy.

Local verification for this head:
- focused Go tests: handler, channel engine, Lark, WeCom, and task service passed
- `go vet` for the same packages passed
- `@multica/core`: typecheck + 135 files / 1,596 tests passed
- `@multica/views`: typecheck + 393 files / 4,643 tests passed
- core/views lint: 0 errors (27 unrelated existing warnings)
- `git diff --check` passed

GitHub CI is running for the new head. Once green, this should be ready for maintainer review. The feature remains opt-in, same-tenant only, fail-closed, and prevents guest-triggered runs/retries from inheriting the installer’s personal connected apps.

xiangboit:
@Bohan-J When you have a chance, could you please help triage this Feishu guest-access proposal or point me to the right channel maintainer? The branch is current with `main`, the additional security/cleanliness review is complete, and all GitHub checks are green. I’m happy to split or narrow the PR if that would make review easier. Thank you!

multica-eve (changes requested):
The product direction is accepted: a restricted Feishu consultation agent may opt in to serving same-company users who do not have Multica accounts, while all existing and new installations remain workspace-member-only by default. Before merge, the permission boundary and two guest-identity paths below need to be fixed.

## 1. Restrict guest-access enablement to workspace owner/admin

`server/internal/handler/lark.go:149` reuses `requireManageableLarkInstallation`, which falls through to `canManageAgent` at line 211. That admits an ordinary workspace member when they own the Agent. The UI mirrors this at `packages/views/settings/components/lark-tab.tsx:343-351`, and `server/internal/handler/lark_test.go:95` currently pins Agent-owner success.

This toggle crosses the workspace membership boundary: an unregistered user can consume runtime capacity and invoke the Agent's configured repositories, workspace MCP servers, and filesystem mode. Owning an Agent is not sufficient authority to expose workspace capabilities to non-members.

Please use a dedicated authorization rule for this setting so only workspace `owner` / `admin` may enable it. Enforce this in the backend, mirror it in the UI as a read-only/disabled state for a non-admin Agent owner, and add coverage proving a non-admin Agent owner receives 403 while workspace owner/admin succeeds. The shared Agent-owner permission for binding or disconnecting the Bot does not need to be broadened or removed.

## 2. Reject non-user senders before guest resolution

`server/internal/integrations/lark/ws_frame_decoder.go:149` decodes `sender_type`, but lines 67-90 do not carry it into `InboundMessage`. `server/internal/integrations/lark/feishu_resolvers.go:132-150` therefore authorizes a guest using only tenant equality. Lark's event model allows the sender type to be `user` or `bot`, so a same-tenant Bot can enter this guest path and potentially create an automated reply loop and unbounded task consumption.

Please carry the sender type to the identity boundary and allow guest access only for the exact human-user value. Missing, unknown, and Bot sender types must fail closed. Add positive coverage for `user` and negative coverage for `bot`, missing, and unknown values. Upstream field definition: https://github.com/larksuite/oapi-sdk-go/blob/d075eb9b94b28b40ba1a558b1198ee138dc38949/service/im/v1/model.go#L3096-L3101

## 3. Preserve an auditable guest identity and make attribution policy explicit

`server/internal/integrations/lark/feishu_resolvers.go:147-150` collapses every guest to the installation owner's Multica UUID plus `External=true`. `server/internal/integrations/channel/engine/resolvers.go:91-108` then discards the guest as a task initiator, so task attribution falls back to the Agent owner and different guests cannot be distinguished in the durable task/issue audit trail.

This also conflicts with workspaces that enable `attribution_fail_closed`: `server/internal/service/task.go:730-743` refuses the unattributed run, while `server/internal/integrations/channel/engine/router.go:770-786` only emits user-visible replies for offline/archived errors and otherwise logs the failure. The Feishu message can be accepted and stored without producing a run or a failure response; `/issue` paths can similarly report or persist partial success without a usable task.

Please define and persist a privacy-safe external actor identity that distinguishes guests without impersonating the installer, then make the fail-closed behavior deliberate. Either the external identity must satisfy the workspace's attribution policy, or guest access must be unavailable/rejected for such workspaces. Any refusal must be reported to the Feishu sender. Add DB-backed coverage for normal chat and `/issue` (text and media) under both fail-open and fail-closed policies.

The fail-closed default, same-tenant checks, durable connected-app deny marker, mixed-batch tightening, and UI rollback behavior are good foundations and should remain.

**Review verdict: Request changes.** The feature is wanted, but these three issues are merge blockers because they govern who may expose workspace capabilities, which senders may invoke them, and whether each guest run is attributable and observable.

xiangboit:
@multica-eve @Bohan-J Thank you for the detailed review.

Before implementing the third blocker, I would like to avoid duplicating an internal identity or channel-access initiative that may already be planned.

I found the following related public work:

- #5150 establishes the authorization/accountability split for Multica members.
- #5906 exposes an already-bound Feishu sender identity to the task runtime.
- #7330 / #7340 persist DingTalk participant information, but only for users already bound to workspace members.

None of these appears to cover an unbound, same-tenant external actor that invokes a restricted Agent.

Is there already an internal roadmap or in-progress design for one of the following?

1. a channel-neutral external actor identity;
2. an auditable grant allowing external channel users to invoke an Agent;
3. privacy and retention rules for native channel identifiers.

If such work exists, I would prefer to adapt #6986 to it instead of introducing a competing model.

If not, I would like to propose a scoped design before implementation:

- persist a stable internal external-actor UUID;
- use an installation-scoped keyed HMAC for lookup and deduplication;
- optionally retain the native provider ID using authenticated encryption, never plaintext;
- attach the external actor and the exact access grant to messages, tasks and `/issue` results;
- keep `originator_user_id` NULL so the guest receives no Multica member authority;
- attribute the access decision to the workspace owner/admin who enabled the grant;
- reject the request with a visible channel response when identity or attribution cannot be persisted.

One policy question needs maintainer confirmation:

Would an exact `external_actor_id` plus an immutable admin-created
`channel_access_grant` be considered sufficiently precise for an
`attribution_fail_closed` workspace, or should guest access remain unavailable
whenever that workspace policy is enabled?

If this direction is compatible with the product roadmap, I can post the
complete schema, retention and rollout proposal for review before changing the
implementation.

Bohan-J:
Thanks again for this contribution, and for the thoughtful follow-up on the external-actor design — your questions were exactly the right ones to ask, and you deserve a clear answer rather than another review round.

To answer your open question first: no, there is no existing internal design covering unregistered same-tenant guests. Your reading of #5150, #5906, and #7330/#7340 was correct — none of them cover this case.

After discussing it internally, we've decided **not to pursue Feishu guest access at this time**, and we're closing this PR. This is a product/architecture decision, not a reflection of your implementation quality.

**Why:** Multica agent runtimes execute on users' own machines — a local daemon with real filesystem access, credentials, and shell. Opening a message entry point to unauthenticated users means anonymous input flowing into a process with machine-level access, where prompt injection must be treated as a normal input class, not an edge case. At that boundary, software-level policy (an answer-only capability mode, or the attribution/grant model we discussed) is not a sufficient defense — isolation is. Making guest access genuinely safe would require a separate, isolated guest execution environment, and we are not ready to build that yet.

**When we'd revisit:** guest access will only be reconsidered when guest-triggered runs can execute in an isolated environment — off users' local machines, carrying no workspace credentials, with per-guest quotas and rate limits. If we get there, your notes on stable `external_actor_id`s, installation-scoped HMAC lookup, admin-created access grants, and fail-closed semantics will be valuable input, and we'd welcome your involvement.

Given this decision, the outstanding CHANGES_REQUESTED items are moot — please don't spend further effort on them. Sorry we couldn't land this one; the discussion genuinely sharpened our understanding of where this security boundary has to sit.

PR #6987, opened 2026-08-14
MUL-4495: feat(agent): add ZCode (Z.ai) as an ACP runtime family

### LINKED ISSUE #5361: [Feature]: Add ZCode (Z.AI coding agent) as a supported agent runtime

lianxin255:
## What does this PR do?

Adds `zcode` to the daemon's supported protocol families — closes #5361.

ZCode (Z.ai's GLM coding agent) doesn't speak ACP itself, so this family drives it through **zcode-acp-server**, a published npm bridge (Apache-2.0) that spawns the desktop-bundled runtime and exposes it as a standard ACP v1 agent. The bridge is a thin translation layer over the official runtime: it doesn't modify or redistribute ZCode, and it reuses the local install's own credentials rather than accepting any. That distinction matters given the P0 on #6755 — no forked flags and no unpublished dependency; the binary this family launches is a published package driving the same stdio server the desktop app uses.

The backend reuses the existing `hermesClient` transport (the kimi/qoder/traecli path), so `server/pkg/agent/zcode.go` is mostly the standard ACP lifecycle plus ZCode-specific behaviours, each verified against the real stack (bridge main @a5cb8a1, ZCode 3.7.7 / CLI 0.16.3):

- `session/new` lands in the bridge's default mode, yolo — already the unattended default, so there's no permission preset to lift
- a model picked in the UI goes through the bridge's `session/setModel`; a rejected switch fails the task rather than quietly running a different model
- a dead session is rejected at `session/resume` (unlike kimi, which echoes success and only fails at prompt time); that maps to `ResumeRejected` so the daemon retries fresh instead of looping on the dead pointer
- cancellation is graceful-first: the backend sends ACP `session/cancel` before the pipes close and the bridge forwards the runtime's stop command; `WaitDelay` is the backstop
- tool names pass through unmapped — the UI sees Bash/Edit/Read the way the Claude families report them
- reasoning effort rides the shared ACP effort path: the bridge advertises option id `thought` (category `thought_level`, aligned upstream in william0wang/zcode-acp#44) with a per-model vocabulary (GLM-5.3: low/high/max), so the picker offers exactly the levels the session's model understands; discovery attaches the vocabulary to the model it belongs to; empty level follows the runtime's own setting

Deliberately not wired: token billing (the bridge's `usage_update` carries context-window occupancy only; the backend already accumulates the ACP usage fields, so counters light up with no change here once the bridge maps them — william0wang/zcode-acp#41).

One prerequisite: the npm 0.1.0 release predates the ZCode ≥0.16 runtime-preferences handshake fix (william0wang/zcode-acp#38) — `initialize` succeeds, then `session/new` hangs ~15s. A release including that fix is required; when `session/new` fails and the bridge reported exactly 0.1.0, the task error carries an upgrade hint instead of a bare timeout.

## Related Issue

Closes #5361

## Type of Change

- [x] Documentation update

## Changes Made

- `server/pkg/agent/zcode.go` + `zcode_test.go` — the ACP backend and 11 fixture tests (launch contract, streaming, model-switch failure, resume rejection at resume and prompt time, resume-replay gating, graceful cancel, known-bad-version hint, effort wiring with wire-level assertions)
- `server/pkg/agent/models.go` — `discoverZcodeModels` via the shared `discoverACPModels` (the bridge advertises its catalog through session/new configOptions), with the per-model thinking vocabulary attached only to the model it belongs to (`acpModelOptionCurrentValue`)
- `server/pkg/agent/agent.go`, `agents_probe.go` (`MULTICA_ZCODE_PATH`/`MULTICA_ZCODE_MODEL`), daemon config/command names/display name, `execenv/runtime_config.go` (AGENTS.md brief), `metrics/labels.go` — registration
- `server/migrations/319_runtime_profile_add_zcode.{up,down}.sql` — protocol-family CHECK
- `packages/core/types/agent.ts`, `runtimes/display.ts`, `views/.../provider-logo.tsx` + official `zcode-logo.png` (from the ZCode.app bundle), `scripts/agent-cli-command-names.txt`
- Docs: providers / install-agent-runtime / environment-variables in en, zh, ja, ko; landing copy tool lists 22 → 23 (all four i18n files)

## How to Test

1. `go test ./server/pkg/agent/` — full suite green, 11 new zcode tests included
2. `cd apps/web && pnpm exec vitest run` + `tsc --noEmit` — clean
3. Real-stack path (what I ran, self-hosted with the bridge on PATH): create an agent on the ZCode runtime, assign an issue → live streaming (tool calls with real names, thought chunks), the model picker populated from the bridge's ACP configOptions (including locally configured custom providers), cross-issue resume continuity, a cancelled long turn leaves no orphaned processes, and the reasoning effort as a behavioural A/B — `low` produced 3 reasoning chunks (62 chars) vs `max` 18 chunks (1600 chars) on the same task

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — *the visible change is a new logo + tool-list rows; can add screenshots on request*
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` — *tool-list copy follows the existing "22 → 23" pattern in all four locales*
- [x] I have considered and documented any risks above — *the bridge-version prerequisite and the unbilled-usage gap are called out in the description*

## AI Disclosure

**AI tool used:** ZCode (Z.ai's GLM coding agent) — fitting, given the subject.

**Prompt / approach:** I drove the whole change agent-first: reverse-engineered the zcode app-server protocol with probes, then had the agent write the Go backend/tests/registration while I verified every claim against a live self-hosted stack (streaming, cancel, resume, model pick, thinking A/B). The PR text itself was drafted from those verified transcripts. Happy to share the probe scripts.

## Screenshots (optional)

Can attach the runtime picker / live issue timeline on request.


lianxin255:
The version floor is now satisfied by a published release: **zcode-acp-server 0.2.0** shipped on npm (2026-08-16) and carries the runtime-preferences handshake fix (#38) this family needs on ZCode ≥0.16.  is the effective requirement — no more building from main. I also updated the backend comments to match (commit on this branch).

lianxin255:
Hi — gentle bump on this one. The branch is now rebased onto current main, the migration was renumbered (319 → 327, the collision that broke the migrations lint is fixed), and CI is green across the board (backend / frontend / installer / execenv). The version prerequisite in the description is also satisfied now: **zcode-acp-server 0.2.0 shipped on npm** and carries the handshake fix this family needs on ZCode ≥0.16.

Everything described in the body is verified against a live self-hosted stack (streaming, cancel, resume, model pick, reasoning A/B), and the tests cover the wire-level behaviour. Happy to adjust naming, split commits, or make any other change — just want to know if the shape is heading in the right direction before I invest further.

(If the scope isn't wanted as a built-in runtime yet, I'm also happy to reshape into a `runtime profile` doc change instead — whatever fits your roadmap.)

multica-eve:
Thank you for the thorough implementation, testing, and documentation around the ACP bridge approach. The Multica team will own and implement official ZCode runtime support directly, so we're closing this community PR. We appreciate the work and the compatibility findings you've shared—they'll be useful reference for the official implementation.

multica-eve (changes requested):
Before this change, users cannot select ZCode when creating a Multica agent. After it, they can run ZCode tasks, continue an earlier conversation, choose the model and reasoning level, attach Multica-managed tools, watch live progress, and cancel work cleanly. It requires installing one additional helper alongside the ZCode desktop app, but the resulting user journey covers the expected first-class agent lifecycle. **Product verdict: accept this direction, provided Multica is willing to own compatibility with that helper.**

## Implementation review

This is the stronger of the two submissions. Reusing the existing ACP client keeps the provider-specific implementation much smaller, and the code covers model-switch rejection, resume rejection, replay filtering, graceful cancellation, reasoning options, MCP forwarding, and usage plumbing. The focused tests are substantially aligned with those failure modes.

The current head still needs these changes before merge:

1. **Rebase onto current main and repair the migration.** The PR is currently conflicting. Prefix `327` is already occupied on main; use the next unique prefix at rebase time (currently after `342`). The new `up` whitelist must preserve every current family, including `mcode`, while adding `zcode`; the `down` migration must restore the pre-ZCode list, also retaining `mcode`.
2. **Put bound skills in a directory the selected bridge actually scans.** `server/internal/daemon/execenv/context.go:433-435` currently falls back to `.agent_context/skills/` because there is no `zcode` case. `zcode-acp-server` 0.2.0 scans project skills under `<cwd>/.agents/skills/`, so Multica-bound skills will not reach ZCode today. Add the mapping and a regression test. Also change the advertised path in `apps/docs/content/docs/providers.mdx:45`, which currently says `.zcode/skills/`.
3. **Fix the install table's detected command.** `apps/docs/content/docs/install-agent-runtime.mdx:40` lists `zcode`, while discovery and execution require `zcode-acp-server`. This would send users to diagnose the wrong executable. The setup paragraph should state the concrete minimum bridge version (`>= 0.2.0`) instead of only referencing an upstream PR.

Validation on the current PR head:

- `go test ./pkg/agent/ -run 'TestZcode|TestDiscoverZcode' -count=1` — pass
- `go test ./internal/daemon/execenv/ -run 'ByteIdentical|Zcode' -count=1` — pass
- `go test ./internal/migrations/ -count=1` — pass on this branch's old base; current main conflicts and must be resolved as above
- `git diff --check origin/main...HEAD` — pass

The remaining structural risk is ownership of a rapidly evolving community bridge. That is manageable only if Multica treats its minimum version and compatibility surface as part of our supported contract rather than assuming arbitrary bridge releases remain compatible.

## Recommendation

🔄 **Request changes, and keep this PR as the implementation to finish.** Once the rebase/migration, skill injection, and installation docs are corrected, this is the better merge candidate.

lianxin255:
Rebased onto current `main` (`30ef4aa`) and addressed all three points from the review. The migration was renumbered again — `343` is now taken upstream by `343_comment_delegated_failure_pending_index`, so ZCode lands as **348** (the next free slot after 347).

**1. Migration renumber + whitelist (348).** `up` adds `zcode` while preserving every current family including `mcode` (23 total); `down` restores the pre-ZCode whitelist, also retaining `mcode`. Both use `NOT VALID` so historical rows stay valid. `migrations` test green.

**2. Bound skills in the directory the bridge scans.** `execenv/context.go` now has a `case "zcode" → .agents/skills/` (what `zcode-acp-server` 0.2.0 actually scans in the session cwd), with a regression test in `execenv_test.go`. The `providers.mdx` skills path is corrected to `.agents/skills/` in all four locales.

**3. Install table + version floor.** `install-agent-runtime.mdx` now lists `zcode-acp-server` (not `zcode`) as the detected command, and the setup paragraph states the concrete minimum `>= 0.2.0` (the first release with the runtime-preferences handshake fix for ZCode >=0.16).

Also carried over from the prior fix commit: `zcode.go` launches through `Config.commandAt(path).exec()` (GH #7046 fixed_args prefix), and `zcode-acp-server` is restored in `defaultAgentCommandNames` after a bad conflict resolution had dropped it. Rebase also pulled in the upstream `daemon.go` `runtimeDisplayNameOverrides` map, where `"zcode": "ZCode"` is now registered (this is what the UI picker reads).

**Validation on this head:**
- `go build ./...` + `go test ./internal/migrations/ ./internal/daemon/execenv/ ./internal/daemon/ ./pkg/agent/ ./internal/metrics/ -count=1` — all green
- Real-stack re-run against the rebased tree: source-built the daemon from this branch, started the local server, and drove a ZCode task through the UI (not the e2e harness). The runtime was detected as online, the task ran 8 tool calls (raw `Bash` names in the transcript), `agent finished status=completed`, and the file the task was asked to write landed on disk with the expected content. No regression from the 11 upstream commits pulled in by the rebase (plugin rebuild, custom issue-status UI, etc.).

Happy to adjust anything else — the shape is now stable on current main.

yuntong613:
I would like to be able to use zcode as soon as possible. Thank you very much.

lianxin255:
@multica-eve — rebased onto current `main` (`a9e82c797`) and renumbered the migration again per the "next unique prefix at rebase time" rule: 404 was taken upstream by `404_agent_starter_prompts`, so ZCode now lands as **469**. The `up` whitelist carries main's current family set (including `codearts` from 441) plus `zcode`; `down` restores exactly that set.

One heads-up unrelated to this branch: current `main` fails `TestMigrationNumericPrefixesAreUnique` on its own — `468_comment_deleted_at` and `468_drop_reference_only_column` share the 468 prefix (verified on a pristine `a9e82c797` checkout). That is the only red in CI here (`backend-tests`, plus the `backend` gate that aggregates it); the other 18 checks are green.

Local validation on the rebased head:

- `go test ./pkg/agent/ -run 'TestZcode|TestDiscoverZcode|TestSupportedTypes'` — pass
- `go test ./internal/daemon/execenv/ -run 'ByteIdentical|Zcode' -count=1` — pass
- `go build ./...` / `git diff --check` — clean

Happy to re-run the migrations suite once the 468 collision is fixed on main.

shelwinnn:
@lianxin255 We've been running this branch locally against bridge 0.37.1 / app-cli 3.11.2-24 / zcode-runtime 0.16.5 and hit three issues around the model catalog. Reference fixes + tests are in shelwinnn/multica#1 (commit `0b1d43de4`) — glad to split any of them into a stacked PR against this branch if that's easier.

**1. Model discovery launches the bridge with the default `acp` positional — discovery silently returns an empty catalog**

`discoverZcodeModels` passes no `acpArgs`, so `discoverACPModels` defaults to `["acp"]`. That's fatal in the daemon, which resolves ExecutablePath to the real `dist/cli.js`: the bridge dispatches on its invoked basename, so the bin-name spelling ignores argv, but `cli.js` parses every positional as a subcommand — `cli.js acp` prints "unknown command 'acp'" and exits before `initialize` (verified against 0.37.1). Discovery reports this as a silently empty catalog, so the model picker only offers manual entry. Fix: spawn bare, matching the Execute path (which already launches with no argv) — e.g. a `bareInvocation` flag on `acpDiscoveryProvider` rather than per-provider argv.

**2. Keyless provider entries poison the runtime's provider registry — default-model turns die before any model call**

ZCode desktop >= 3.11 keeps its provider/model store in `~/.zcode/cli/config.json`, while the bridge still reads `~/.zcode/v2/config.json`. When the latter is missing, the catalog falls back to a single hardcoded model and `session/setModel` is always rejected. Copying the file verbatim is harmful in its own way, though: the store also lists providers without credentials (e.g. an unconfigured Z.AI Coding Plan entry), the bridge pushes them into the runtime's provider registry, and a default-model turn then fails with `provider_not_configured` "Model provider is missing an API key: zai" before any model call is attempted (verified live against bridges 0.37.1 and 0.43.4 — the newest bridge pushes keyless providers identically). Fix: mirror `cli/config.json` -> `v2/config.json` before every spawn (turn path and discovery), filtered to providers that can actually authenticate — apiKey present, `apiKeyRequired: false`, or a localhost baseURL, i.e. the bridge's own `providerSelectable` rule from `dist/config/options.js`. The copy is marked `_multicaManaged` so it never clobbers a foreign-managed file (including the pre-3.11 desktop's own writes at the same path), and every failure degrades silently to pre-sync behaviour.

**3. `session/setModel` rejects display-case bare ids persisted in agent.model**

The bridge only accepts the encoded `provider\model` values it advertises; a persisted bare id (`GLM-5.3-Flash`, from manual entry or a pick from an older catalog) gets resolved against the bridge's config store instead of the live session, so it can target a provider the runtime no longer knows and the switch is rejected — failing the task. Fix: rewrite the requested id onto the session's advertised value only when exactly one case-insensitive bare-id match exists in the catalog this very session returned; zero or several matches pass through verbatim so the runtime's rejection stays the loud, honest failure. Usage should also be keyed by the resolved id (the model the turn actually ran on), not the requested one.

Happy to provide more detail on any of these.

multica-eve:
关联的 Issue #5361 已明确由官方团队接管 ZCode runtime 支持，并要求关闭相关社区 PR，避免并行实现发生冲突。关闭本 PR。

PR #6992, opened 2026-08-15
MUL-6206: fix(mobile): re-apply app config to ios/ on every local iOS build

### LINKED ISSUE #6994: [Bug]: pnpm ios:* never re-applies app.config.ts, so the documented EXPO_BUNDLE_IDENTIFIER_PROD fix does nothing
### Deployment type

Official App

_(The client is built from source per the README's personal-install flow, which connects to the `multica.ai` backend. The bug is in the local iOS build tooling, so neither option fits exactly.)_

### What happened?

`apps/mobile/README.md` documents this remedy for a bundle identifier someone else has claimed on Apple's developer portal:

```bash
export EXPO_BUNDLE_IDENTIFIER_PROD=com.yourname.multica
pnpm ios:mobile:device:prod:release
```

It has no effect after the first build. The build succeeds, and Xcode still signs with the old identifier.

The cause is that `expo run:ios` prebuilds **only when `ios/` is missing**. From `ensureNativeProjectAsync` in `@expo/cli`:

```ts
if (!fs.existsSync(path.join(projectRoot, platform))) {
  await prebuildAsync(projectRoot, { install: !!install, platforms: [platform] });
} else {
  return true;   // config plugins never re-run
}
```

`ios/` is gitignored, so the first `pnpm ios:*` on a machine generates it and every later run skips prebuild. Nothing in `app.config.ts` reaches the native project again.

The bundle identifier is just the documented case. Everything `app.config.ts` owns freezes the same way. On a checkout whose `ios/` had been generated once and never regenerated:

| Expected from `app.config.ts` | Actual in `ios/` |
| --- | --- |
| `icon: ./assets/icon.png` | empty `AppIcon.appiconset`, no `filename` — no Home Screen icon |
| `bundleIdentifier: ai.multica.mobile` | `PRODUCT_BUNDLE_IDENTIFIER = org.name.Multica` |
| `name: "Multica"` | no `CFBundleDisplayName` |
| `scheme: "multica"` | no `CFBundleURLTypes` — deep links dead |
| `photosPermission` | no `NSPhotoLibraryUsageDescription` — image picker hard-crashes on iOS 14+ |
| `orientation: portrait` | portrait + both landscapes |

Every build reported success throughout. Expected: `pnpm ios:*` produces an app matching the current `app.config.ts`.

Running `npx expo prebuild -p ios` by hand fixed all six at once.

### Steps to reproduce

1. `cd apps/mobile && npx expo prebuild -p ios` (or just run any `pnpm ios:*` once) so `ios/` exists.
2. Change something in `app.config.ts` — e.g. set `name` to `"Multica Test"`.
3. `pnpm ios:mobile:prod`.
4. Build succeeds. `plutil -p ios/Multica/Info.plist | grep CFBundleDisplayName` still shows the old name, and the simulator/device shows the old name.

For the README's documented case specifically: set `EXPO_BUNDLE_IDENTIFIER_PROD` to a domain you own, re-run `pnpm ios:mobile:device:prod:release`, then check `grep PRODUCT_BUNDLE_IDENTIFIER ios/Multica.xcodeproj/project.pbxproj` — it is unchanged, and signing fails exactly as before.

### Screenshots (optional)

Not attached — the failure is in generated native config rather than in UI. The `plutil` and `grep` output in the steps above is the observable evidence.

```shell
# Reported against
2c0912b6e  MUL-5421: make workspace MCP servers an explicit per-agent assignment (#6976)

# Environment
macOS (Darwin 25.5.0), Xcode signing team via free Apple ID
Expo CLI 55.0.30, Expo SDK 55, React Native 0.82
Physical iPhone 15, Release configuration

# Observed on a checkout whose ios/ was generated once, weeks earlier
$ grep -m1 PRODUCT_BUNDLE_IDENTIFIER ios/Multica.xcodeproj/project.pbxproj
    PRODUCT_BUNDLE_IDENTIFIER = org.name.Multica;
$ ls ios/Multica/Images.xcassets/AppIcon.appiconset/
Contents.json          # no PNG, and no "filename" key inside
$ plutil -p ios/Multica/Info.plist | grep -c NSPhotoLibraryUsageDescription
0

# After a single manual prebuild
$ npx expo prebuild -p ios --no-install
$ grep -m1 PRODUCT_BUNDLE_IDENTIFIER ios/Multica.xcodeproj/project.pbxproj
    PRODUCT_BUNDLE_IDENTIFIER = "ai.multica.mobile";
$ ls ios/Multica/Images.xcassets/AppIcon.appiconset/
App-Icon-1024x1024@1x.png  Contents.json

# Proposed fix
PR #6992 routes every ios:* script through a wrapper that prebuilds first.
```

szainmehdi:
## What does this PR do?

Makes local iOS builds re-apply `app.config.ts` to the generated `ios/` project.

**The user-visible symptom is in the README.** `apps/mobile/README.md` tells anyone hitting a claimed bundle identifier to do this:

```bash
export EXPO_BUNDLE_IDENTIFIER_PROD=com.yourname.multica
pnpm ios:mobile:device:prod:release
```

That has no effect after the first build. `expo run:ios` prebuilds **only when `ios/` is missing** — from `ensureNativeProjectAsync` in `@expo/cli`:

```ts
if (!fs.existsSync(path.join(projectRoot, platform))) {
  await prebuildAsync(projectRoot, { install: !!install, platforms: [platform] });
} else {
  return true;   // <- config plugins never re-run
}
```

So the project keeps the old identifier, signing keeps failing, and the documented escape hatch looks broken to a new contributor.

The same freeze hides every other `app.config.ts` change. `ios/` is gitignored, so the first `pnpm ios:*` on a machine generates it and every later run silently skips prebuild. A checkout whose `ios/` had been generated once was still the bare template:

| Expected from `app.config.ts` | Actual in `ios/` |
| --- | --- |
| `icon: ./assets/icon.png` | empty `AppIcon.appiconset`, no `filename` |
| `bundleIdentifier: ai.multica.mobile` | `PRODUCT_BUNDLE_IDENTIFIER = org.name.Multica` |
| `name: "Multica"` | no `CFBundleDisplayName` |
| `scheme: "multica"` | no `CFBundleURLTypes` — deep links dead |
| `photosPermission` | no `NSPhotoLibraryUsageDescription` — image picker hard-crashes on iOS 14+ |
| `orientation: portrait` | portrait + both landscapes |

Every build reported success. A single `expo prebuild -p ios` fixed all six at once.

### Why a wrapper script instead of inlining prebuild

The scripts already wrap `expo run:ios` in `dotenv` + `cross-env`, and prebuild has to run under that same env — `app.config.ts` branches `bundleIdentifier` and `name` on `APP_ENV` and reads `EXPO_BUNDLE_IDENTIFIER_PROD` from `.env.production`. A bare `expo prebuild` inside a prod script would rewrite the project to the **dev** bundle id. A wrapper keeps the eight scripts readable and gives the reasoning one home. An npm `preios` hook does not work here: `pre` only fires for an exactly-matching script name, so it would miss `ios:device:prod:release` and friends.

### Flag choices

- `--no-install` — safe because `run:ios` installs pods itself when `Podfile.lock` is stale ([Expo docs](https://github.com/expo/expo/blob/sdk-55/docs/pages/bare/installing-expo-modules.mdx): "the run command will install them for you").
- No `--clean` — `expo-build-properties` sets `buildReactNativeFromSource`, so a clean prebuild forces a from-source pod rebuild that is far too slow for the edit/run loop. Documented as the manual escape hatch when a native dependency changes.

## Related Issue

Closes #6994

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `apps/mobile/scripts/ios-run.sh` (new, mode 100755) — runs `pnpm exec expo prebuild -p ios --no-install`, then `exec`s `pnpm exec expo run:ios "$@"`. Header comment records why.
- `apps/mobile/scripts/ios-run.test.sh` (new, mode 100755) — stubs `pnpm` on `PATH` and asserts the call sequence. Follows the existing `scripts/*.test.sh` idiom (`mktemp -d`, `trap cleanup EXIT`, stub bin dir, assert recorded calls).
- `apps/mobile/package.json` — all eight `ios:*` scripts call `./scripts/ios-run.sh` instead of `expo run:ios`; `dotenv` / `cross-env` prefixes and trailing flags unchanged. `test` also runs the new shell test.
- `apps/mobile/CLAUDE.md` — new "Local iOS builds go through `scripts/ios-run.sh`" subsection under Build & release, so new iOS scripts get routed through the wrapper.

Two deliberate placement calls, flagged so they read as decisions rather than oversights:

- **Script lives in `apps/mobile/scripts/`, not root `scripts/`.** Root is repo infrastructure (postgres, install, worktree); mobile owns its build pipeline per `apps/mobile/CLAUDE.md`. Happy to move it if you'd rather keep all shell scripts together.
- **No workflow file was touched.** `mobile-verify.yml` already runs `turbo … test --filter=@multica/mobile`, so chaining the shell test into mobile's `test` script makes it run in CI with no workflow edit. If you'd prefer an explicit `run: bash …` step like the root scripts get, say so and I'll add it.

## How to Test

Reproduce on `main`:

1. `cd apps/mobile && npx expo prebuild -p ios` to generate `ios/`.
2. Edit `app.config.ts` — change `name` to anything.
3. `pnpm ios:prod`. Build succeeds; `grep CFBundleDisplayName ios/Multica/Info.plist` still shows the old value.

Verify on this branch: repeat 1–2, then `pnpm ios:prod` — the new name is in `Info.plist` and on the device Home Screen.

The shell test needs no `node_modules`, Xcode, or device:

```
$ bash apps/mobile/scripts/ios-run.test.sh
ios-run.test.sh: all assertions passed
```

## Checklist

- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs**
- [ ] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx`

### Verification notes

- **`ios-run.test.sh` passes**, and it was mutation-checked: deleting the prebuild line from the wrapper makes it fail with `first call should be the prebuild, got: exec expo run:ios`. It fails for the bug this PR fixes, not just for syntax errors.
- **`pnpm exec expo` resolves** from `apps/mobile` (`55.0.30`) — that substitution replaced an earlier `npx`, to match `scripts/check.sh`.
- **Verified for real:** the prebuild → build → install → launch sequence on a physical iPhone 15, Release configuration. That run is what produced the app icon, `ai.multica.mobile`, and the correct `Info.plist`.
- **Not run on this branch:** a full device build. It was performed in a separate checkout of the same commit's content. A reviewer with a working tree should do one real `pnpm ios:prod` before merge.

### Why README and `mobile-app.mdx` are untouched

Script names and their documented behavior don't change — the README's table and its `EXPO_BUNDLE_IDENTIFIER_PROD` instructions describe what the scripts are supposed to do, and this PR makes them true. Editing that page would also mean syncing the `zh` / `ja` / `ko` mirrors of `apps/docs/content/docs/mobile-app.mdx`, which is a translation change I shouldn't fold into a bug fix.

### Risks

- Adds a prebuild (a few seconds) to every `ios:*` invocation.
- Prebuild overwrites hand edits inside `ios/`. That directory is gitignored and fully generated, so hand-editing was never durable — but anyone relying on a local Xcode tweak will lose it on the next run.
- Only iOS is touched. There are no `android:*` scripts today; the same trap applies if any are added.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** Started from "why is there no app icon?" on a device build. Claude traced it to the empty `AppIcon.appiconset`, found the wider config drift, and confirmed the `run:ios` early-return against Expo SDK 55 source rather than from recall. It then re-reviewed its own first draft against repo conventions — which is where the README bug, the `scripts/*.test.sh` testing idiom, and the `bash` / `pnpm exec` house style came from, all of which the first draft had missed.

## Screenshots (optional)

Home Screen after the fix — icon renders. The white ring around it is a separate problem, filed as #6995: `assets/icon.png` is the desktop icon (rounded square + drop shadow on white padding) being masked again by iOS. It needs a full-bleed 1024x1024 source, which is a design decision rather than a build one, so it is deliberately out of scope here.


multica-eve (approved):
Thank you for the thorough investigation and the well-scoped fix. The regression coverage and the explanation of the prebuild and Pods behavior are especially helpful. Approved for merge.

PR #7005, opened 2026-08-15
fix(comments): reveal newly posted messages

### LINKED ISSUE #6983: Newly posted comments are not scrolled into view

TobinZuo:
## What does this PR do?

Fixes issue-detail timelines not revealing a comment or reply immediately after the current user posts it.

The timeline cache already contained the server response; the missing behavior was viewport intent. The browsing path is virtualized and intentionally does not use Virtuoso `followOutput`, so appending data alone neither mounts the new row nor moves the reader.

This change keeps the authoritative created-comment result, uses its ID to mount the owning Virtuoso thread, and then scrolls only the issue-detail container to the exact comment row. Realtime comments from other users do not arm this behavior, so they cannot interrupt the reader.

## Related Issue

Closes #6983

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Thinking Path

1. The create-comment mutation already inserts the server response into the React Query timeline cache and deduplicates the matching WebSocket event.
2. The issue-detail browsing path uses Virtuoso without `followOutput`, so a newly appended row may remain outside both the viewport and the mounted DOM window.
3. The submit helpers previously collapsed the authoritative server response to a boolean, discarding the ID needed for precise post-create behavior.
4. The fix returns the created comment, records scroll intent only for an explicit local submit, asks Virtuoso to mount the owning root thread, then measures and reveals the exact comment node.
5. Native `scrollIntoView` remains avoided because it can also move the desktop shell; only the issue-detail container's `scrollTop` changes.

## Changes Made

- `packages/views/issues/hooks/use-issue-timeline.ts`
  - Return `Comment | null` from successful comment and reply submissions.
  - Preserve existing failure handling and partial-trigger warnings.
- `packages/views/issues/components/issue-detail.tsx`
  - Track the authoritative posted comment ID and owning root-thread ID.
  - Mount virtualized targets with `scrollToIndex`.
  - Reveal the exact row after it enters the DOM.
  - Keep WebSocket-only updates viewport-neutral.
- `packages/views/issues/components/issue-detail.test.tsx`
  - Cover a newly posted top-level comment.
  - Cover a newly posted reply inside its root thread.

## How to Test

1. Open an issue with enough timeline entries to scroll.
2. Post a top-level comment while the newest timeline row is outside the viewport.
3. Confirm the newly posted row is revealed.
4. Post a reply in an expanded thread and confirm the exact reply is revealed.
5. Receive a realtime comment while reading elsewhere and confirm the viewport does not jump.

Automated verification:

- `pnpm --filter @multica/views exec vitest run issues/components/issue-detail.test.tsx issues/hooks/use-issue-timeline.test.tsx`
- `pnpm --filter @multica/views test` — 342 files / 4098 tests passed
- Targeted ESLint for the three changed files
- GitHub CI — 9/9 checks passed

## Playwright Before/After Verification

Captured against isolated databases with identical fixtures at a 1440×900 viewport.

| Build | Timeline state after POST | Result |
| --- | --- | --- |
| Before (`2c0912b6e`) | `scrollTop = 0`; created node not mounted | New comment is posted but remains out of view |
| After (`59e3949d7`) | `scrollTop = 4031`; created node mounted at 594–772px inside the 56–892px container viewport | New comment is immediately visible |

The before run also confirmed the visible comment count increased from 18 to 19, separating “POST failed” from “POST succeeded but viewport did not follow.”

## Risk Notes

- The scroll target is armed only by an explicit local submit, not by cache updates or WebSocket events.
- Replies scroll via their owning root thread because Virtuoso virtualizes thread roots, while the final adjustment targets the exact reply node.
- The target is cleared on issue changes and after a successful reveal.
- The mount wait is bounded and cleaned up through `requestAnimationFrame` cancellation.

## Checklist

- [x] Documentation is not affected

## AI Disclosure

**Prompt / approach:** Used Codex to inspect the issue timeline mutation and virtualization paths, identify the missing local-submit scroll intent, implement a scoped fix, add comment and reply regressions, run package and CI verification, and reproduce the behavior before and after with Playwright.

## Screenshots

Before/after screenshots were captured locally during the Playwright verification. GitHub attachment upload is still pending; the measured viewport evidence is recorded above so the behavioral result remains reviewable without relying on a static image alone.

### DIFF
--- a/packages/views/issues/components/issue-detail.test.tsx
+++ b/packages/views/issues/components/issue-detail.test.tsx
@@ -434,6 +434,7 @@ vi.mock("@multica/core/issues/stores", async () => ({
 // background instead, which is mechanism-independent and observable without
 // layout.
 const scrollIntoViewSpy = vi.hoisted(() => vi.fn());
+const virtuosoScrollToIndexSpy = vi.hoisted(() => vi.fn());
 
 vi.mock("react-virtuoso", () => ({
   Virtuoso: forwardRef(function MockVirtuoso(
@@ -445,7 +446,7 @@ vi.mock("react-virtuoso", () => ({
       // since the deep-link cold-path drives the container's scrollTop on the
       // real DOM node, not Virtuoso's imperative API.
       scrollIntoView: vi.fn(),
-      scrollToIndex: vi.fn(),
+      scrollToIndex: virtuosoScrollToIndexSpy,
     }));
     return (
       <div data-testid="virtuoso-mock">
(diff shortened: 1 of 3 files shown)


multica-eve:
Thank you for the contribution. This behavior has already shipped in #6914, which resolved #6983. Closing this PR as a duplicate to avoid maintaining two implementations of the same user-facing behavior.

PR #7014, opened 2026-08-15
MUL-6225: fix Gmail SMTP verification delivery

hon94csl:
## Summary

Gmail SMTP accepted Multica verification messages with a `250` response but silently omitted them from recipient inboxes. Controlled cross-tests against a working Go SMTP application isolated the failure to Multica supplying `Date` and provider-domain `Message-ID` headers alongside its MIME transfer metadata. SMTP delivery now leaves `Date` and `Message-ID` assignment to the provider while preserving the existing TLS, authentication, HTML, and transfer-encoding behavior.

## Validation

- `go test ./internal/service`
- Rebuilt and deployed the self-host backend, then confirmed a real `/auth/send-code` request reached a Gmail inbox.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

### DIFF
--- a/server/internal/service/email.go
+++ b/server/internal/service/email.go
@@ -290,7 +290,6 @@ func (s *EmailService) sendSMTP(to, subject, htmlBody string) error {
 	// non-ASCII workspace/inviter names crossing strict or older SMTP hops.
 	has8Bit, _ := c.Extension("8BITMIME")
 	encodedSubject := mime.QEncoding.Encode("utf-8", subject)
-	msgID := fmt.Sprintf("<%d@%s>", time.Now().UnixNano(), s.smtpHost)
 
 	var bodyBytes []byte
 	var cte string
@@ -319,8 +318,6 @@ func (s *EmailService) sendSMTP(to, subject, htmlBody string) error {
 	headers := "From: " + s.fromEmail + "\r\n" +
 		"To: " + to + "\r\n" +
 		"Subject: " + encodedSubject + "\r\n" +
-		"Date: " + time.Now().UTC().Format(time.RFC1123Z) + "\r\n" +
-		"Message-ID: " + msgID + "\r\n" +
 		"MIME-Version: 1.0\r\n" +
 		"Content-Type: text/html; charset=UTF-8\r\n" +
 		"Content-Transfer-Encoding: " + cte + "\r\n" +
--- a/server/internal/service/email_test.go
+++ b/server/internal/service/email_test.go
@@ -502,6 +502,8 @@ func TestSendSMTP_OpenClientFailureNoPanic(t *testing.T) {
 type testSMTPServer struct {
 	Listener net.Listener
 	Addr     string
+	// ReceivedData captures complete DATA payloads for message-level assertions.
+	ReceivedData chan string
 
 	// Auth mechs advertised in EHLO response (e.g. "LOGIN" or "PLAIN LOGIN")
 	AuthMechs string
@@ -521,6 +523,9 @@ func startTestSMTPServer(t *testing.T, cfg testSMTPServer) (*testSMTPServer, fun
 	}
 	cfg.Listener = l
 	cfg.Addr = l.Addr().String()
+	if cfg.ReceivedData == nil {
+		cfg.ReceivedData = make(chan string, 4)
+	}
 
 	done := make(chan struct{})
 	go func() {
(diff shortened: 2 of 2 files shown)


Bohan-J (changes requested):
Reviewed both the product intent and the implementation (MUL-6225).

The product goal is right and worth fixing: self-hosted deployments on Gmail SMTP silently failing to deliver login codes leaves members with no workaround. The direction is fine — the current implementation is not.

## Must-fix: `sendSMTP` is the generic SMTP path, not a Gmail-specific one

`server/internal/service/email.go:318` serves every SMTP configuration, and both `SendVerificationCode` and `SendInvitationEmail` route through it. After this change Multica emits messages with no `Date` and no `Message-ID` for **all** of them, not just Gmail verification codes.

That is only safe if the next hop is guaranteed to supply both. It isn't:

- [RFC 5322 §3.6](https://www.rfc-editor.org/rfc/rfc5322.html#section-3.6) makes `Date` a **required** field, and `Message-ID` a SHOULD.
- [RFC 6409 §8.1/8.2](https://datatracker.ietf.org/doc/html/rfc6409#section-8.1) lets a *submission* server add them — "MAY"/"SHOULD", not a guarantee.
- `SMTP_PORT` defaults to `25`, and auth is skipped entirely when `SMTP_USERNAME` is empty (see `TestSendSMTP_NoAuthWhenUsernameEmpty`). So this path is routinely a plain RFC 5321 relay, which must *not* rewrite the message beyond trace fields. Those deployments will now emit messages missing a mandatory header — typically spam-scored (`MISSING_DATE`, `MISSING_MID`) or rejected outright. Google's own [SMTP error reference](https://support.google.com/mail/answer/3726730?hl=en) lists a missing or invalid `Message-ID` as a rejection reason, so this can regress inbound-to-Gmail delivery for non-Gmail senders too.

## The likely root cause is the `Message-ID` value, not its presence

The removed line built `<{UnixNano}@{s.smtpHost}>` — i.e. `<1723...@smtp.gmail.com>`, a `Message-ID` whose domain is Google's own relay hostname rather than the sending domain. That is exactly the shape a receiving Gmail is entitled to distrust or dedupe away, and it matches the reported symptom precisely: `250` accepted, never lands in the inbox.

Suggested fix: **keep both headers**, and derive the `Message-ID` from the sender's own domain with a crypto-random unique component, e.g. `<{crypto-random}@{domain of s.fromEmail}>`. Leave `Date` as it is — `time.RFC1123Z` is already valid.

If dropping headers really is required for some specific provider, please scope it to an explicitly detected submission path instead of applying it to every SMTP configuration.

## Test coverage

`TestSendSMTP_OmitsProviderManagedHeaders` (`server/internal/service/email_test.go:699`) asserts that Multica *omitted* the headers, against a stub server that never adds anything back. It locks an incomplete message in as the expected output, so it can't demonstrate that the recipient receives a compliant message. More useful would be a regression test asserting a valid `Date` + `Message-ID` on the generic relay path, plus a targeted test for whatever Gmail-specific behavior you end up keeping.

## Minor

Could you add to the PR description whether the reproduction used `smtp.gmail.com` or `smtp-relay.gmail.com`, along with the before/after raw headers as received? That pins the root cause instead of inferring it from the one successful end-to-end test.

---

The diff is small, the test helper is clean, and CI is green — requesting changes purely on the scope of the header removal and the missing mandatory header.

Bohan-J:
Thanks for this, and apologies for the slow follow-up after the review.

Your diagnosis was correct, and it's what the eventual fix is built on. You isolated the failure to the `Message-ID` being rooted at the provider domain — `<{UnixNano}@smtp.gmail.com>`, a domain the sender doesn't own — which is exactly the right root cause, and not an obvious one to find. The controlled cross-test against a working Go SMTP application, plus confirming a real `/auth/send-code` reaching a Gmail inbox on a redeployed self-host backend, is more validation than most reports of this kind come with.

The review's only disagreement was the remedy: dropping `Date` and `Message-ID` would have applied to every SMTP configuration, not just Gmail submission, and `sendSMTP` is routinely a plain RFC 5321 relay (port 25 by default, auth skipped when `SMTP_USERNAME` is empty) that must not add them back on our behalf.

That fix has now landed in #7577, implemented as suggested in the review: keep both headers, and re-root the `Message-ID` at the sender's own domain with a crypto-random component.

On why it went into a separate PR rather than waiting for an update here — re-rooting the `Message-ID` at the sender's domain requires parsing the configured sender correctly, and that was blocked by a separate bug (#7568): `SMTP_FROM_EMAIL` carrying a display name was used verbatim as the SMTP envelope, so naively taking the text after `@` yields `example.com>` rather than `example.com`. The two fixes had to land together, so they were combined.

Closing this in favor of #7577. The root-cause analysis is yours — thank you for taking the time to chase it down and write it up properly.

PR #5932, opened 2026-07-25
fix(transcript): auto-follow live task output in transcript dialog

kakiuwang-ui:
## Summary

Since the transcript event list moved to Virtuoso for virtualization (#5733), a live task's new events required manual scrolling — the dialog opened at the top and never followed appended output. Reported in #5921 as a regression after 0.4.10.

This PR wires live-follow through Virtuoso's own primitives rather than raw scroll listeners, handling the two sort directions separately because their growth models differ.

## Changes (single file: `agent-transcript-dialog.tsx`)

| Direction | Growth model | Follow mechanism |
| --- | --- | --- |
| Chronological (default) | events **append** at the bottom | `followOutput={(atBottom) => isLive && atBottom ? "smooth" : false}` — Virtuoso's native follow, with `atBottomThreshold={120}` for a forgiving "at bottom" zone |
| Newest-first | events **prepend** at the top, and `firstItemIndex` anchoring deliberately holds the viewport — so a reader parked at the top silently stops seeing new rows | track the top edge via `atTopStateChange` into a ref; when new events arrive while at the top, `scrollToIndex({ index: 0 })` snaps back to the newest row |

Plus: opening a **live** chronological transcript now lands on the newest event (`initialTopMostItemIndex: { index: "LAST", align: "end" }`, the same pattern the chat message list uses). The per-`listEpoch` remount re-applies it after task/sort/filter changes. Completed tasks keep opening at the top.

## Behaviour

- **Live, at the live end**: new events auto-scroll into view (both directions).
- **Live, scrolled away**: follow is suspended; returns automatically when the reader scrolls back to the live end.
- **Completed tasks**: `isLive = false` keeps everything dormant — no behavior change.
- **Sort/filter/task switches**: unchanged; the existing remount semantics apply.

## Verification

- `tsc --noEmit` on `@multica/views` — clean.
- `vitest run common/task-transcript/agent-transcript-dialog.test.tsx` — 9/9 pass.

Fixes #5921.

PR #7018, opened 2026-08-15
MUL-5982: fix(chat): keep following rapid live output

### LINKED ISSUE #6697: [Bug]: Chat messages are not being followed
### Deployment type

Official App

### What happened?

This bug is simular to #5921, but in the `Chat` page.

When the screen fills with text, the messages are not being followed, so I need to keep scrolling to see the recent logs.

### Steps to reproduce

1. Go to the chat 
2. Start a conversation
3. Wait for the screen fills with text
4. New messages are not being followed (the last messages are hidden until I scroll the mouse)

### Screenshots (optional)

_No response_

```shell

```

TobinZuo:
## What does this PR do?

Fixes Chat live output stopping its automatic follow while an agent is still streaming rapidly.

Thinking path:

- Chat already delegates list positioning to React Virtuoso.
- The existing `followOutput` callback ignored the current `atBottom` argument and instead read React state updated by `atBottomStateChange`.
- It also requested `smooth` scrolling for every append. While one animation is still in flight, the next append can observe the viewport away from the bottom and permanently disengage following. The transcript fix in #5932 documents the same Virtuoso behavior.
- This change uses the current Virtuoso callback value directly and requests `auto` for live growth. Manual scrolling away still returns `false`, and history prepends remain guarded by `isFetchingOlderMessages`.

## Related Issue

Closes #6697

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `packages/views/chat/components/chat-message-list.tsx`: derive following from the current Virtuoso `atBottom` value and use immediate scrolling for rapid appends.
- `packages/views/chat/components/chat-message-list.test.tsx`: cover following at the live end, preserving reader-controlled scroll-away, and suppressing follow during older-history prepends.

## How to Test

1. Open Chat with enough history to fill the viewport.
2. Start an agent response that emits output rapidly.
3. Keep the viewport at the live end and verify every new chunk remains visible.
4. Scroll upward and verify new chunks do not pull the reader back down.
5. Return to the bottom and verify following resumes.

Automated verification:

- `pnpm --filter @multica/views test` — 345 files / 4151 tests passed.
- `pnpm --filter @multica/views typecheck` — passed.
- `pnpm --filter @multica/views lint` — passed with pre-existing warnings only.
- CI frontend build/typecheck/lint command — 12/12 tasks passed.
- The two repository-wide scan tests that exceeded their 5-second timeout under local parallel package load passed 36/36 when rerun directly.

Real-Virtuoso Playwright harness, 5 runs of 80 rows appended every 24 ms:

| Strategy | Final bottom gap |
| --- | ---: |
| Previous smooth/state callback | 6352–6526 px |
| Current immediate/callback-argument strategy | 1 px in all 5 runs |

## Checklist

- [x] Documentation update is not required for this behavior-only bug fix
- [x] Landing copy and docs sync is not applicable
- [x] Chinese product copy is not changed

Risk: live output now moves immediately instead of animating. This matches the existing Transcript live-follow fix and avoids treating an in-flight animation as reader intent. Readers who scroll away remain undisturbed.

## AI Disclosure

**Prompt / approach:** Inspected unclaimed issues, traced #6697 through Chat and the prior Transcript fix, wrote a failing Virtuoso-contract regression test, changed the minimum callback behavior, and validated both strategies with a real React Virtuoso Playwright harness under rapid appends.

## Screenshots

A side-by-side Playwright capture was produced locally; it can be attached in review because GitHub CLI cannot upload local image attachments.

### DIFF
--- a/packages/views/chat/components/chat-message-list.test.tsx
+++ b/packages/views/chat/components/chat-message-list.test.tsx
@@ -20,16 +20,21 @@ vi.mock("react-virtuoso", () => ({
     computeItemKey,
     components,
     context,
+    followOutput,
   }: {
     data: unknown[];
     itemContent: (i: number, item: unknown) => ReactElement;
     computeItemKey: (i: number, item: unknown) => string;
     components?: { Footer?: (p: { context?: unknown }) => ReactElement | null };
     context?: unknown;
+    followOutput?: (atBottom: boolean) => "smooth" | "auto" | false;
   }) => {
     const Footer = components?.Footer;
     return (
-      <div>
+      <div
+        data-follow-at-bottom={String(followOutput?.(true))}
+        data-follow-away-from-bottom={String(followOutput?.(false))}
+      >
         {data.map((item, i) => (
           <div key={computeItemKey(i, item)} data-row-key={computeItemKey(i, item)}>
             {itemContent(i, item)}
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thank you @TobinZuo for the excellent contribution! 🎉

Final review passed. What stood out:

- The root-cause analysis is exactly right: the old `followOutput` read a React state that lags one append behind, and a `smooth` scroll still in flight makes the next append observe "not at bottom" and permanently drop the follow. Switching to the live `atBottom` callback argument with an immediate `auto` scroll is the documented Virtuoso pattern and mirrors our earlier Transcript fix, so Chat and Transcript now share the same live-follow contract.
- All the important boundaries are preserved: the 120px near-bottom threshold, no follow while older history is prepending, and readers who scroll up to review history are never yanked back down.
- Great regression coverage for all three decisions, and the real-Virtuoso Playwright measurement (6000+px drift → 1px) is a model for how to validate this kind of fix.

Re-verified locally against current `main` (merged, no conflicts): targeted tests 22/22, typecheck, and lint all pass. CI is green. Merging now — this fixes #6697.

Issue #5635, opened 2026-07-19
[Feature]: Workspace export / backup


CookieTsang:
## Problem

Currently, Multica doesn't provide a way to export or back up an entire workspace. As a user, I'd like the ability to regularly back up the following:

* Complete Agent configurations, including system prompts, instructions, attached Skills, environment variables, and other settings
* Full conversation history for all Agents (currently, Chat History only allows access to the current channel and cannot enumerate all conversations)
* Complete definitions of Squads, Autopilots, and Skills
* Uploaded files and attachments

## Why This Matters

Agent configurations and conversation history are valuable user assets. Without a reliable backup mechanism:

* Carefully crafted Agent prompts and configurations could be permanently lost if they are accidentally deleted.
* Important conversations, context, and decision history cannot be archived for future reference.
* Migrating a workspace to another instance or environment becomes difficult.

A comprehensive backup and export feature would significantly improve data safety, portability, and users' confidence in adopting Multica for long-term use.

Thank you for considering this feature request!

jackyon:
+1

Issue #5131, opened 2026-07-09
[Feature]: Import all data in the workspace with one click


dujiecn:
### Deployment type

Official App

### What do you want and why?

In our work, we need to migrate workspace data. It's quite troublesome to manually create it every time. We hope to support one-click import of workspace, or import and export of skills, agents, and teams.

Issue #4019, opened 2026-06-11
feat(agents): copy agent to another workspace


jabbourWearable:
## Context

Users who manage multiple Multica workspaces today have no way to reuse an agent they built in workspace A inside workspace B. The only paths are manual recreation (copy-paste instructions, rebuild skill list) or the intra-workspace "Duplicate" action, which is scoped to a single workspace. Teams that split work across workspaces (e.g. a dev workspace and a staging workspace, or separate workspaces per client) hit this on every new workspace.

The fix is a `POST /api/agents/{id}/copy-to-workspace` endpoint plus a "Copy to workspace…" action in the agent row dropdown. The agent's configuration is reproduced in the target workspace; secrets (`custom_env`, `mcp_config`) are explicitly excluded.

## Current State

**Intra-workspace duplicate** (`packages/views/agents/components/agent-row-actions.tsx:163-167`): opens `CreateAgentDialog` with a `template` prop pre-filled from the source agent. Copies: `name` (+ " copy"), `description`, `visibility`, `model`, `instructions`, `avatar_url`, `custom_args`, `max_concurrent_tasks`, skill IDs. Does NOT copy: `custom_env`, `mcp_config`. Scoped to the current workspace only.

**Agent template system** (`server/internal/handler/agent_template.go`): the closest analogue. `POST /api/agents/from-template` creates an agent in the current workspace with template-defined instructions + skills. Skill matching uses `GetSkillByWorkspaceAndName` to dedupe by name before fetching.

**Agent routes** (`server/cmd/server/router.go:835-858`): new endpoint fits at `r.Post("/{id}/copy-to-workspace", h.CopyAgentToWorkspace)`.

**Workspace membership check** (`server/internal/handler/handler.go:439`): `workspaceMember(w, r, workspaceID)` validates membership for any workspace ID.

**Skill matching query** (`server/pkg/db/queries/skill.sql`): `GetSkillByWorkspaceAndName` already exists and is used by the template system.

### Backend: `POST /api/agents/{id}/copy-to-workspace`

**Request:**

```json
{
  "target_workspace_id": "uuid",
  "runtime_id": "uuid",
  "name": "optional override — defaults to source name"
}
```

**Response:** the newly created `AgentResponse` in the target workspace (same shape as `POST /api/agents/`).

**Handler logic (`CopyAgentToWorkspace`):**
1. Load source agent via `loadAgentForUser` (handles UUID or slug, gates to caller's access).
[...]

Issue #3924, opened 2026-06-08
Export & version agents/skills as code (GitHub-syncable config)


davidmoradi:
### Deployment type

Official App (multica.ai)

### What do you want and why?

Agent and skill definitions (instructions/"soul", model, MCP config, skill bindings) are stored server-side in my Multica workspace — which is good — but there's no way to treat that config as code I own and version. Concretely:

- **No version control** — I can't diff, review, or roll back changes to an agent's instructions over time.
- **No export / portable copy** — my config only exists inside Multica's cloud; I have no file I control to back up to my own git, move to another workspace, or keep outside the platform (lock-in risk).
- **No reproducibility / sharing** — I can't reproduce the same agent+skill setup in another workspace or hand it to a teammate without manually recreating everything.

Skills already have `multica skill import --url github.com/...`, which proves the config-from-GitHub pattern works. I'd like the same export/version/apply story extended to agents, and a bulk declarative round-trip for both.

### Proposed solution (optional)

Phase it so the first slice is small and self-contained:

**Phase 1 — export + declarative apply (CLI only, opt-in):**

- `multica agent export <id>` → emit the agent definition (name, description, instructions, model, runtime/visibility, skill bindings) as a YAML/JSON file. Skills referenced by slug; secrets referenced, not inlined.
- `multica apply -f <dir>` → idempotently upsert agents + skills from a directory of these files (create if missing, update by name/slug if present). Reuses the existing `agent create/update` and `skill import` logic.
- Repo layout, e.g.:

  ```
  agents/helper.yaml
  skills/<slug>/SKILL.md
  ```

  This alone delivers the core value: commit to GitHub, restore anywhere, diff/review in PRs.

**Secrets (the key design question):** `custom_env` and `mcp_config` carry API tokens and must never land in git. Proposal: store only references (e.g. `MY_TOKEN: ${env:MY_TOKEN}`) and resolve at apply time from the local environment or an existing secret store. Export should redact/reference these by default.

**Phase 2 — optional auto-sync:** point a workspace at a GitHub repo and reconcile agents/skills on push (webhook). Bigger lift; can wait. Phase 1 stands alone.

> Framing note: this is intended as an opt-in CLI/declarative-config capability, separate from the (intentionally narrow) PR↔issue GitHub integration — not an expansion of that.

PR #7023, opened 2026-08-16
feat(cli): export and import agent configuration (MUL-6234)

forrestchang:
Adds `multica agent export` and `multica agent import` — a portable JSON document for an agent's configuration, so agents can be reviewed, version-controlled, backed up, or moved between workspaces.

Closes MUL-6234.

## Design decisions

The original request was one sentence, so the open questions (format, overwrite/merge strategy, which fields, CLI vs UI) were resolved as follows — each by matching a pattern that already exists in this repo rather than inventing a parallel one:

| Question | Decision | Why |
|---|---|---|
| Which fields | Exactly the set `agent copy` already treats as portable | One definition of "portable agent configuration" in the product, not two |
| Format | `{"kind": "multica.agent.config", "version": 1, "source": {…}, "agents": […]}` | A list from the start, so `--all` and single-agent export share one format; a wrong `kind` / unknown `version` is refused outright rather than half-applied |
| Conflict strategy | `--on-conflict fail\|overwrite\|rename\|skip` | Same vocabulary as `skill import`, so one word means one thing across the CLI |
| Entry point | CLI | Composes over existing endpoints (no server API, no migration); the web/desktop surface is a follow-up that can reuse this exact document |

## Behavior

- **Secrets are never exported.** `custom_env`, `mcp_config` and `runtime_config` stay out of the document by construction — the API redacts or masks them anyway, so an export could only record masked junk, and a config file that quietly carries credentials is the wrong artifact to hand around. The document records only that they existed, and the import report restates it, so a missing API key surfaces before the agent's first run rather than during it.
- **Plan then apply.** Every entry resolves to an action before the first write, so a conflict or a missing `--model` cannot leave half a document applied. `--dry-run` prints that plan; when an apply does fail midway, the report still prints so the operator knows how far it got.
- **Cross-workspace imports** drop `member` / `team` invocation targets rather than granting invocation rights to whoever holds that id in the new workspace, falling back to `private` when nothing survives (never a `public_to` nobody can invoke). Skills re-resolve by id, then by unique name; anything unresolved is reported, not silently dropped.
- **Runtime rule matches `agent copy`**: `--runtime-id` is optional only when re-importing into the workspace the document came from; landing elsewhere drops `model` / `thinking_level` / `service_tier` and requires an explicit `--model` (`--model ""` accepts the target default).
- An overwrite applies the document as a **complete statement** of the portable config, so the target ends up matching the document instead of keeping values the document does not carry.

```bash
multica agent export <agent-id> --file agents.json     # one agent
multica agent export --all --file agents.json          # every non-archived agent
multica agent import --file agents.json --dry-run      # plan, no writes
multica agent import --file agents.json --runtime-id <target> --model <model>
```

## Not included (deliberate, documented in the skill)

Workspace MCP server assignments (`agent mcp`), `disabled_runtime_skills`, `composio_toolkit_allowlist`, `status`, and archive state. Each is a separate binding or a non-portable runtime state; adding any of them is a format-compatible follow-up.

## Tests

33 new tests in `server/cmd/multica/` covering the portable-field whitelist, the secrets-never-exported guarantee, all four conflict strategies, ambiguous-name refusal, `--into`, the runtime/model rule, cross-workspace target and skill resolution, `--dry-run`, envelope validation, and an export→import round trip.

```
go test ./cmd/multica/ -run 'TestAgentImport|TestAgentExport'   # 33 passed
go test ./internal/service -run 'TestCreatingAgentsSkillCoversAgentCreationContracts|TestBuiltinSkillsConformToTemplate'   # passed
go vet ./cmd/multica/   # clean
```

Also smoke-tested the built binary against a mock API for export, dry-run, all four conflict strategies, overwrite, and the cross-workspace path.

Note: the rest of `./cmd/multica` has 117 pre-existing failures in this checkout — tests that do not `t.Chdir` pick up the agent workdir's daemon-task marker. Identical count with and without this change.

## Docs

`multica-creating-agents` SKILL.md + its source map (required by CLAUDE.md when CLI flags change), plus `cli` and `agents-create` docs pages in all four locales.


forrestchang:
Closing for now — deprioritised (MUL-6234).

Not closing because anything is wrong with it. State at close:

- 10/10 CI checks green (backend, frontend, installer on all three platforms, windows-execenv)
- MERGEABLE / CLEAN, no review requested changes — nobody reviewed it
- 1 commit, 15 files, +2314 / -1
- 33 new tests covering the portable-field whitelist, the secrets-never-exported guarantee, all four conflict strategies, `--into`, the runtime/model rule, cross-workspace target and skill resolution, `--dry-run`, envelope validation, and an export→import round trip

**The branch `agent/lambda/e1681ad8` is kept, not deleted** — reopening this PR restores the work as-is.

Context for whoever picks this up: a requirements analysis on MUL-6234 found ~10 open issues asking for cross-workspace config reuse (#3132, #3924, #4019, #5635, #5131, #2707, #3035, plus #4420 and #4472 closed as duplicates). The conclusion there was that skills are already portable (`skill import --url` + `--on-conflict` + `refresh`), this PR covers the agent layer, and squads + autopilots are the remaining gaps. If that work is ever picked up, this is step 1 of it.

PR #7025, opened 2026-08-16
feat(web): make the mobile web app installable (MUL-6237)

forrestchang:
Closes MUL-6237 (tier 1 of three).

The app can now be added to a phone's home screen and runs in its own window instead of a browser tab. Offline support and web push are deliberately not here — they were scoped as tier 2 and tier 3 on the issue.

## What's in it

**`app/manifest.ts`** — served at `/manifest.webmanifest`. The one non-obvious decision is `start_url`: it is `/inbox`, not `/`. `isOfficialMarketingHost` in `proxy.ts` keeps the root path on the public site even for a signed-in session, so an installed app pointed at `/` opens the landing page. `/inbox` is one of `LEGACY_ROUTE_SEGMENTS`, which the proxy already resolves per session — last workspace when signed in, `/login` when not. `manifest.test.ts` runs `start_url` and both launcher shortcuts through the real `proxy()` so that coupling can't silently rot.

**Home-screen icons** — generated from the brand marks already in the repo, so an installed web app shows the same artwork as the installed desktop app. The `any` pair comes from `apps/desktop/build/icon.png`; the maskable one is full-bleed with the mark held at 52% of the canvas so it stays inside the safe zone Android crops to. `public/icons/icon.svg` is the source, and `app/manifest.ts` carries the regeneration commands.

**`public/sw.js`** — caches nothing, on purpose. It's here because Chrome only offers to install a site whose active worker has a fetch handler (iOS just needs the manifest and meta tags), and because landing `skipWaiting` + `clients.claim` + a cache sweep now means the eventual caching change only has to add a strategy rather than retrofit a safe upgrade path. That retrofit is how apps strand users on stale HTML requesting `/_next` chunks the current deploy already deleted. Registration is skipped in development, where a worker only ever serves stale bundles and outlives the dev server on localhost.

## Verification

`pnpm typecheck`, `pnpm lint` and `pnpm test` (218 tests) pass in `apps/web`. Verified against a dev server that `/manifest.webmanifest` serves the expected JSON as `application/manifest+json`, that the head renders exactly one `mobile-web-app-capable`, the `apple-mobile-web-app-*` pair and the `apple-touch-icon` link, and that all five static assets return 200.

`pnpm build` fails while prerendering `/workspaces/new`, but it fails identically on `main` with these changes stashed — pre-existing, not from this PR. Compilation itself succeeds.

## Not verified

Nothing here has been on a real phone yet. Three things need a device pass:

- iOS runs an installed app in its own storage container, so installing means logging in once more. Worth confirming the Google OAuth redirect returns cleanly in standalone mode; the email-code path is the fallback if it doesn't.
- `statusBarStyle` is `default`, which keeps the web view below the status bar. Edge-to-edge (`black-translucent` + `viewport-fit=cover`) needs `env(safe-area-inset-*)` padding, which no surface in the app has yet — that's a deliberate follow-up, not an oversight.
- No `apple-touch-startup-image`, so iOS shows a blank frame while launching.

The shell is only as good as the responsive UI inside it, so this pairs with the mobile layout work in MUL-6236.


forrestchang:
Both review points addressed in 9a0d911.

**Blocker — confirmed and fixed.** `proxy.ts` bounced a `multica_logged_in` session with no `last_workspace_slug` to `/`, which the official marketing hosts deliberately keep on the public site even with a session — so the installed app dead-ended on the landing page, with no URL bar to recover. That state now goes to `/login`, which already resolves an authenticated visitor against their workspace list, including pending invitations and the no-workspace-yet case. This also fixes the same dead end for plain legacy bookmarks on the marketing hosts, which was pre-existing.

The deep-link path is dropped rather than passed as `next`: a legacy segment fed back through `next` would return to that branch and loop.

Regression coverage added for the third state in both `manifest.test.ts` (start_url plus both shortcuts) and `proxy.test.ts` (including explicitly on `multica.ai` / `www.multica.ai`). Verified they fail against the old behaviour — `expected 'https://www.multica.ai/' to contain '/login'` — before the fix went in.

**Service worker — removed, with one correction.** The install-criteria change applies to *menu* installation. Per the same Chrome post, the automatic install prompt still requires a fetch handler: "For now the algorithm that displays the install prompt still requires the presence of a `fetch()` handler." So dropping the worker does cost the automatic Android install prompt, not nothing.

Removing it is still right, and the conclusion lands where the review did: the worker bought a prompt heuristic in exchange for an empty fetch handler Chrome's own guidance warns against, plus a sticky registration that would need an unregister path. There is no custom `beforeinstallprompt` UI to make the prompt useful. Android installs from the ⋮ menu, iOS from the share sheet, and the worker returns in tier 2 when there is something real to cache.

`@multica/web` typecheck, lint and 218 tests pass. Re-verified against a dev server that the manifest and head tags are unchanged and `/sw.js` is now 404.

PR #7028, opened 2026-08-16
refactor(settings): reorganize settings by scope and subject (MUL-6232)

forrestchang:
Implements the settings IA agreed in MUL-6232. The nav had grown to 21 entries in two groups, ordered by when each feature shipped; three of them configured one subject, and two more were halves of one job that linked back and forth at each other.

## What changes

**Three scopes, four segments.** The nav becomes Account / Desktop (This Device) / Workspace. The workspace group is split by dividers into segments that answer, in turn: what is this → who can use it → how we work in it → what it is connected to. The rule that keeps it from drifting again: a new setting joins the end of *its segment*, not the end of the column.

**Labels / Properties / Quick Actions → one workspace Issue tab** with in-page sections. It shares its name with the account's Issue tab — the group heading carries the scope, and each page links to the other. The three panels lost their `SettingsTab` wrapper and are now `PropertiesPanel` / `LabelsPanel` / `QuickActionsPanel`; each keeps its own description, shown under the section switcher.

**GitHub + self-hosted Git → Repositories**, with the repository list between the connection and the behavior switches: connect, pick repositories, then decide what Multica writes back. This drops the round-trip cross-link the two tabs pointed at each other, and a duplicate "Connect GitHub" button that the merge exposed — with no installation the list has nothing to import, so connecting is the Hosting section's job alone.

**Composio → MCP.** Hosted MCP apps are the same job as the servers a workspace adds by hand. Its feature flag and 503 probe moved with it into a new `ComposioSection`, so the host no longer carries gating for a section it merely displays.

**Also:** Daemon and Updates become their own desktop group (`desktopTabs`, previously `extraAccountTabs` appended to Account); Billing moves up beside Members; Labs is hidden while it has no experiments; Profile is renamed General to match the workspace's General.

21 entries → 17 with every flag on, 19 → 15 in the common case.

## Links keep working

Collapsed `?tab=` values redirect: `github → repositories`, `composio → mcp`, `labels|properties|quick-actions → workspace-issue`. The three former Issue tabs still name the section they meant, so an old bookmark opens the panel it aimed at. In-app deep links updated: the label picker's "Manage labels", the agent MCP tab's Composio link, and the GitHub install return target.

The workspace Issue tab uses `?tab=workspace-issue` — `issue` already belongs to the account tab, and the two must stay distinguishable.

## Verification

- `pnpm typecheck` — clean across all packages
- `pnpm lint` — 0 errors
- `pnpm --filter @multica/views test` — 346/347 files pass

Two failures are pre-existing on this branch point, confirmed by re-running them on a stashed tree: `layout/sidebar-resize.test.tsx`, and three `@multica/core` files (`projects/mutations`, `realtime/use-realtime-sync-ws-instance`, `workspace/mutations`). This PR does not touch `packages/core`.

New/updated tests: nav grouping and the desktop group, the Labs gate, every collapsed-tab redirect, account-vs-workspace Issue disambiguation, section deep-links and URL writeback, Composio gating in its new home, and the self-hosted Git gate in its new home.

## Not done here

The mock rendered self-hosted Git as a compact provider row; this PR mounts the existing `VCSTab` under a section heading instead. Reshaping that surface is a separate change from moving it.

Screenshots of the running app are not attached — `make dev` fails on this machine against a pre-existing database on the shared Postgres, unrelated to this branch.


forrestchang:
Closing at the requester's direction (MUL-6232). Not merged.

The settings IA this implements was reviewed and agreed on the issue; the branch stays on the remote if the work is picked up again. Two things to know before reusing it:

- It has since gone out of date with `main` and needs a rebase — GitHub reports the branch as conflicting.
- The last two commits removed the segment dividers and then the segment spacing, so the workspace nav is one flat, evenly spaced list. That was the reviewed outcome; do not reintroduce the rules.

Branch: `agent/lambda/277de351` (`3eedaaf07`). All 10 checks were green at that commit.

PR #7035, opened 2026-08-16
MUL-6273: feat(agent): add MiniMax Code ACP runtime

hetaoBackend:
## What

- add MiniMax Code as a first-class runtime, discovered from `mcode` or `MULTICA_MCODE_PATH`
- execute tasks through the native `mcode acp` stdio server
- forward Multica-managed stdio, HTTP, and SSE MCP servers according to the ACP capability handshake
- inject `AGENTS.md` and project skills under `.minimax/skills/`
- expose MiniMax Code in runtime profiles, display names, docs, and the supported CLI guard
- keep model selection managed by MCode and safely retry a fresh session while ACP reports `loadSession: false`

## Why

MiniMax Code already exposes the same ACP v1 lifecycle Multica uses for several existing runtimes. A dedicated adapter lets users run Multica tasks through their installed and authenticated MCode CLI without translating its stream output or mutating MCode's persistent model configuration.

## Compatibility

- minimum supported MiniMax Code version: `0.1.2`
- model selection is intentionally disabled because the current ACP surface has no session-scoped model option
- session restoration is capability-gated; `0.1.2` reports `loadSession: false`, so continuation falls back to a fresh MCode session
- no authenticated model prompt was sent during verification

## Validation

- `go test ./internal/daemon -run 'Test(ListRuntimeLocalSkills_Mcode|DefaultAgentCommandNamesCoversAllProbes|AgentCLIGuardCoversDefaultCommands|RemoteMCPProviderMatrixAndConfigMerge)' -count=1`
- `go test ./internal/daemon/execenv -run 'Test(McodeUsesNativeProjectSkillRoot|PrepareThenCleanupSidecarsRoundTripEmptyWorkdir)' -count=1`
- `pnpm --filter @multica/core test -- agents/mcp-support.test.ts runtimes/display.test.ts`
- `pnpm --filter @multica/core lint`
- `pnpm typecheck`
- real `mcode 0.1.2` ACP initialize handshake: protocol v1, `loadSession:false`, HTTP/SSE MCP capabilities present


hetaoBackend:
@Bohan-J Could you please review this MiniMax Code ACP runtime integration when you have a chance? You have been the maintainer/reviewer for the recent agent-runtime integrations, so your feedback on the runtime contract would be especially helpful. The implementation is complete and all current CI checks are green. The PR remains a draft for maintainer feedback; I am happy to address any issues you find.

multica-eve (changes requested):
Thanks for adding MiniMax Code as a first-class runtime. The product direction makes sense, and the integration covers the expected registries, MCP capability handling, runtime-owned model selection, migration, documentation, and unit tests. I’m requesting changes for the following merge blockers:

1. `server/pkg/agent/mcode.go:189-197` defers `cancel()` before deferring the cleanup closure. Because defers run in reverse order, every early return closes stdin and calls `cmd.Wait()` before cancellation can terminate the process tree. The EOF dependency is therefore: early return → close stdin → wait for MCode to exit → stdout/stderr readers finish → only then call `cancel()`. If MCode or a descendant does not exit on stdin EOF, the cancellation path is unreachable while `Wait` is blocked. This matters immediately for the `loadSession: false` branch at `server/pkg/agent/mcode.go:230-241`, which every follow-up run on 0.1.2 takes. Please follow the established Hermes ordering (`close stdin → cancel → Wait → releaseProcessGroup`) and add a regression fixture that acknowledges initialize, ignores EOF, and stays alive; assert that the message stream closes and the full process tree is reaped.

2. The real execution contract has not been verified. The current tests use a shell fake, and the PR notes that only the real initialize handshake was exercised without an authenticated prompt. That leaves `session/new`, prompt responses/notifications, permission requests, `AGENTS.md`, `.minimax/skills/`, and tool execution unverified against MCode 0.1.2. Please add an opt-in `mcode_integration_test.go` guarded by `agentintegration` and `MULTICA_RUN_REAL_AGENT_SMOKE=1`, following the other ACP runtimes, and record an authorized MiniMax-side run that at least verifies a real response plus AGENTS/skill canaries.

3. The public install journey is incomplete. `apps/docs/content/docs/install-agent-runtime.mdx:30` links to a repository that currently provides desktop downloads/issue reporting but no way to install the `mcode` CLI, while `SELF_HOSTING.md:132` assumes `mcode` 0.1.2+ can be placed on PATH. Please provide the official CLI install link/command and supported-platform matrix. If the CLI is bundled with the desktop app, document how it becomes available on PATH; if Linux is not supported, narrow the self-hosting claim accordingly.

The focused Go suites, prompt byte-stability check, Windows cross-compilation, diff integrity, and current GitHub CI are green. Once the three items above are addressed, this should be ready for a focused re-review.

hetaoBackend:
Addressed all three blockers in `ddc9003`:

1. **Early-return process cleanup**
   - Reordered cleanup to `close stdin → cancel → Wait → releaseProcessGroup`, matching the established Hermes lifecycle.
   - Added `TestMcodeUnsupportedResumeTerminatesProcessTreeBeforeWait`. The fixture acknowledges `initialize` with `loadSession:false`, spawns a descendant, ignores stdin EOF, and stays alive. The test asserts the rejected-resume result, message-stream closure, and reaping of both parent and child.
   - Verified RED on the previous head and GREEN on this head. Latest focused run: `go test -race ./pkg/agent -run '^TestMcode' -count=3`.

2. **Authenticated MCode 0.1.2 execution contract**
   - Added opt-in `server/pkg/agent/mcode_integration_test.go`, guarded by `//go:build agentintegration` and `MULTICA_RUN_REAL_AGENT_SMOKE=1`.
   - It exercises a real `session/new` + prompt, cwd-scoped `AGENTS.md`, project `.minimax/skills/`, file-tool execution, ACP tool-use/tool-result forwarding, and a non-empty session ID.
   - Authorized local run passed against `mcode 0.1.2` in 14.94s with 2 tool-use/result pairs; the tool-result stream contained the `TOOL-MCODE-OK` canary and the final output contained all three canaries.

3. **Public install path**
   - Replaced the repository link with the public npm package: https://www.npmjs.com/package/@minimax-ai/code
   - Documented `npm install --global @minimax-ai/code@latest`, the published Node engine range (`>=22.19 <23 || >=24 <27`), npm global binary PATH discovery, and `mcode login` in self-hosting plus EN/ZH/JA/KO runtime docs.
   - Rechecked public registry metadata for `@minimax-ai/code@0.1.2` and `bin.mcode`.

Additional verification:
- Windows amd64 `go test -c ./pkg/agent`: PASS
- `git diff --check`: PASS
- Two independent focused reviews confirmed the cleanup regression is resolved; their evidence-strength comments were incorporated by tying the smoke assertion to the canary-bearing tool result and removing an unsupported platform-matrix claim.

The PR remains Ready for review. @multica-eve, could you please take a focused re-review once the new CI run completes?

hetaoBackend:
@multica-eve Latest head `985a44856` is now fully green after addressing the review blockers and the new GH #7046 runtime-command CI gate. Could you please take a focused re-review of the MCode ACP integration when convenient? Thank you.

hetaoBackend:
@multica-eve Follow-up: the head is now `001f62e44` after a metadata-only history rewrite to correct the commit author/committer identity. The code trees, messages, dates, topology, and final diff are unchanged from `985a44856`; all 12 checks are green again. The focused re-review request remains the same.

multica-eve (changes requested):
Focused re-review of head `001f62e44` confirms that the three blockers from my previous review are resolved:

1. Early-return cleanup now executes `close stdin → cancel → Wait → releaseProcessGroup`, and `TestMcodeUnsupportedResumeTerminatesProcessTreeBeforeWait` covers an initialized process plus descendant that ignore stdin EOF. The focused MCode suite passed locally under `-race` three consecutive times.
2. The opt-in `agentintegration` smoke now covers a real `session/new` and prompt, `AGENTS.md`, a project skill, file-tool execution, ACP tool events, and a non-empty session ID. It checks `MULTICA_RUN_REAL_AGENT_SMOKE=1` before executable/account access, and the contributor recorded an authorized MCode 0.1.2 run with all three canaries and two tool-use/result pairs.
3. The install docs now link the public npm CLI package and provide the install command, supported Node engine range, PATH guidance, and login step. Registry metadata confirms that 0.1.2 exposes the `mcode` binary and the documented engine range.

There is one new must-fix caused by `main` advancing:

- `server/migrations/332_runtime_profile_add_mcode.up.sql:1` (and its down pair) now reuses numeric migration prefix `332`, which current `main` already assigns to `server/migrations/332_issue_status.up.sql`. Migration prefixes must be unique in the merged tree; the different filenames prevent GitHub from reporting a textual conflict. Please update from current `main` and rename the MCode migration pair to the next available unique prefix (currently `342`), then rerun the migration/CI checks.

Local validation passed:

- `go test -race ./pkg/agent -run '^TestMcode' -count=3`
- focused daemon runtime/MCP tests
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1`
- `agentintegration` smoke gate/compile without enabling account access
- Windows amd64 `pkg/agent` cross-compile

**Verdict: Request changes.** The original review blockers are closed; only the migration-number collision currently blocks merge.

multica-eve:
I confirmed that the migration number collision was the only remaining review blocker and pushed the fix directly to this PR branch in `5213e0c9a`.

The commit renames the MCode migration pair from `332_runtime_profile_add_mcode` to `342_runtime_profile_add_mcode`, which is the next available prefix after current `main` (`341`). The migration SQL is unchanged; I also replaced the stale `pre-327` wording in the down-migration comment with `pre-MCode`.

Local validation passed:

- `go test ./internal/migrations -count=1`
- `go test ./cmd/migrate -count=1`
- `go test ./cmd/server -run '^TestServerHealthReadyHandler' -count=1`
- `git diff --check`

The push has started a fresh CI run. I found no additional review blockers.

multica-eve (approved):
Re-reviewed head `5213e0c9a`. The three original blockers remain resolved, and I directly fixed the only remaining issue by renumbering the MCode migration pair from `332` to the currently unique prefix `342`. The SQL behavior is unchanged, and the migration lint/runner plus readiness unit tests pass locally. Approved; fresh CI is running for the pushed commit.

PR #7037, opened 2026-08-16
feat(dingtalk): distinguish bot identities in group routing

### LINKED ISSUE #7036: [Feature]: Distinguish DingTalk bot identities in same-group routing
### Deployment type

Official App (the behavior must remain self-hostable)

### What do you want and why?

When two DingTalk robot installations from the same Multica workspace are added to the same DingTalk group, the backend correctly keeps their routes separate by `installation_id + conversation_id`. The Settings → Integrations → DingTalk group-routing UI does not surface the installation identity, however: both rows show the same group title and conversation id, so an owner/admin cannot tell which robot route they are reassigning.

This was explicitly accepted as a follow-up when #6660 merged. The follow-up should make the existing installation boundary readable without changing routing or chat-session semantics.

### Proposed solution

- Identify each route as the Bot for its installation's stable default Agent, matching the identity already shown in Connected bots.
- Keep the route target Agent separate: the Bot identity describes the installation, while the selector describes which Agent currently handles that group.
- Add a short installation-id disambiguator when the readable labels collide or the installation/default Agent cannot be resolved.
- Preserve graceful behavior for version-skewed Desktop clients and avoid exposing AppKey/AppSecret or other credentials.
- Cover two active installations observing the same `conversation_id`, including owner/admin and member read-only rendering.
- Update DingTalk documentation and localized product copy.

### Acceptance criteria

- Two DingTalk Bots in the same group render as two visibly distinct route rows.
- Reassigning one row still PATCHes only that row's stable route id.
- The Bot identity remains tied to the installation's default Agent even when the route targets a different Agent.
- Missing/stale installation metadata renders a safe fallback instead of hiding the route.
- Agent A → Agent B → Agent A continues to start a new Chat session; this issue does not change that accepted product decision.

### Screenshots / mockups

No separate mockup. Follow the existing Settings → Integrations → DingTalk list-row layout and design tokens.

mxbao063-bao:
## Summary

- label every DingTalk group route with the bot installation's default Agent and a short installation ID
- keep bot identity separate from the group route's selected Agent, with a safe fallback when installation metadata is unavailable
- document the same-group multi-bot behavior and add matching en/zh-Hans/ja/ko copy

The existing Agent A → B → A behavior remains unchanged: each reassignment starts a new chat session.

Closes #7036

## Testing

- `pnpm --dir packages/views test --run locales/parity.test.ts settings/components/dingtalk-tab.test.tsx settings/components/dingtalk-tab.group-routes-error.test.tsx` (191 tests passed)
- `pnpm typecheck`
- `pnpm lint` (passes with pre-existing warnings only)
- local Web dev route `/acme/settings?tab=integrations` compiled successfully and redirected to login as expected without a running backend
- `pnpm build`: Desktop completed; Web/Docs were blocked by the local certificate chain while `next/font` fetched Google Fonts (`UNABLE_TO_GET_ISSUER_CERT_LOCALLY`)


yyclaw (changes requested):
Overall decision: I do not recommend adopting the current approach, and suggest closing this PR.

### [PR7037-001][P2][blocking][pr_change] The Agent-name label is redundant and does not identify the actual DingTalk bot

In the new `Bot for {{agent}}` UI, `{{agent}}` is still an Agent name. The row already displays an Agent name on the right, so this is redundant UI. Using the final six characters of `installation_id` to distinguish bots is only marginally useful; the better solution is to display the actual DingTalk bot name directly.

For that reason, the practical value of this PR is limited and I have decided not to adopt it. My local build already implements displaying the actual DingTalk bot name, and I will open a separate PR for it. Please watch for that follow-up. I recommend closing this PR directly.

Required action: Close this PR and follow the upcoming implementation that displays the actual bot name.

Required tests: Not applicable; this review rejects the product approach rather than requesting test or implementation revisions to the current solution.

Issue #7033, opened 2026-08-16
Protect collaborative edits and support independent runs from comment history


yyclaw:
## Problem

Multica has two related opportunities to make issue conversations safer and more useful when humans and agents work concurrently.

### 1. Collaborative edit safety

Issue titles, descriptions, and comment bodies can be edited from multiple tabs, clients, users, or agents. A stale editor can currently submit after a newer edit and silently replace the newer value. The existing `description_base` behavior preserves asynchronously inserted channel media, but it is not a general stale-write guard for collaborative text editing.

Silent last-write-wins behavior is especially costly in an agent-native task system: users cannot tell that intent was lost, and an agent may continue from content that another participant believed was already superseded.

The editing contract should therefore:

- expose a server-authoritative revision for editable issue and comment state;
- let modern clients submit the revision they edited;
- reject stale conditional writes with `409 Conflict` atomically at the database update;
- preserve the local draft and refresh the server version so the user can reconcile intentionally;
- remain compatible with older installed clients that do not yet send a revision; and
- prevent delayed realtime events from regressing a cache that has already observed a newer revision.

This is valuable on its own for the issue and comment editing flows that exist today.

### 2. Independent runs from comment history

An issue conversation often contains multiple possible directions. Today, continuing execution from the current issue state does not provide an explicit way to explore one historical comment branch without also inheriting later or sibling discussion.

Users should be able to start an independent agent run from any comment. The run should receive a frozen issue snapshot plus the selected comment's ancestry from the thread root to the selected comment. It should not implicitly include sibling comments, descendants added later, or unrelated subsequent discussion. This creates a clear equivalent of "branch in a new chat" while keeping the original issue conversation intact.

The resulting run should be independently identifiable in the execution log, idempotent to retries, routed to an explicitly selected or well-defined default agent, and safe when that agent is already busy.

## Delivery plan

To keep review scope cohesive, this issue will be delivered as exactly two pull requests:

- [x] **PR 1 — stale edit protection for existing issue and comment editing flows.** This PR is justified entirely by the existing collaborative editing behavior. It will contain the revision contract, atomic conditional writes, conflict UX, realtime monotonicity, compatibility behavior, and focused concurrency tests. It will not contain comment-branch task fields, routing, prompts, or UI.
[...]

PR #7038, opened 2026-08-16
feat(issues): add revision-aware concurrency guards

yyclaw:
## What does this PR do?

Adds revision-aware concurrency guards for issues and comments so stale clients cannot silently overwrite newer text while unrelated activity can continue without false conflicts. The change carries monotonic owner revisions through the database, API, realtime events, shared caches, web/desktop views, and mobile mutations.

Web/desktop text editors send the content they adopted as a field baseline: `title_base`, `description_base`, or `content_base`. The server returns a structured `409` only when that same text field changed concurrently; reactions, status changes, drag/drop moves, labels, attachments, and other unrelated mutations do not reject an otherwise valid text edit. Web/desktop preserve a conflicting local draft and present a side-by-side comparison so the user can review or retry intentionally. Mobile intentionally retains last-write-wins title/description edits until it has equivalent conflict-reconciliation UX, avoiding an unrecoverable `409` loop.

Aggregate revisions order authoritative full HTTP and WebSocket snapshots. Auxiliary/revision-only events use independent monotonic ordering for the cache projections they can actually update; they invalidate a stale full owner projection instead of advancing its revision as though they carried a complete snapshot. Ordinary non-text issue updates retain last-write-wins behavior.

Legacy clients that omit field baselines remain supported during rollout, and the API continues to accept explicit `expected_revision` tokens for compatibility with callers that intentionally need aggregate compare-and-swap behavior.

## Related Issue

Refs #7033

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added `revision` columns for issues and comments in `server/migrations/351_issue_comment_revision.*.sql`.
- Added monotonic owner-revision increments across issue/comment mutations and related collections in `server/internal/handler/` and `server/pkg/db/queries/`.
- Added atomic field-baseline conflict checks for issue titles, issue descriptions, and comment bodies while preserving explicit aggregate CAS compatibility.
- Removed blanket `expected_revision` injection from normal status, assignee, parent, drag/drop, and other non-text issue mutations.
- Propagated authoritative owner revisions through API responses and WebSocket events, rejecting stale or duplicate full snapshots in `packages/core/`.
- Kept auxiliary event ordering separate per cache projection; partial events invalidate stale full owner projections without claiming their revisions.
- Removed client-side `revision + 1` inference; queued saves now advance from authoritative responses and accepted content baselines.
- Added conflict-safe optimistic mutation handling and revision conflict comparison UI for web/desktop in `packages/views/`.
- Kept mobile issue text edits last-write-wins until conflict reconciliation is available, while retaining authoritative HTTP/WS revision propagation and stale-cache refetch behavior.
- Made issue updates and attachment binding one transaction with a consistent attachment-before-issue lock order; binding failures roll back the issue update.
- Ensured pure no-op issue writes do not advance `revision` or `updated_at`, while a newly bound attachment advances the owner revision exactly once and returns/publishes the authoritative snapshot.
- Added backend, migration, shared-core, view, and mobile regression coverage for concurrent interleavings, rollback, no-op semantics, false-conflict prevention, stale/partial events, and true same-field conflicts.

Rollout risk is limited by retaining backward-compatible request behavior. The migration initializes existing rows to revision `1`; older clients and mobile can omit the new baselines and continue using last-write-wins writes, while clients with conflict-reconciliation UX (currently web/desktop) receive `409` only for a true same-field conflict. Explicit `expected_revision` remains accepted for compatibility. Realtime events remain recoverable through authoritative fetches when a client receives only a partial or unversioned update.

## How to Test

1. Run `make migrate-up`, then run `go test ./internal/handler ./internal/migrations` from `server/`.
2. Run the core, view, and mobile unit tests.
3. Run `pnpm typecheck` plus the mobile typecheck and lint checks.
4. In web/desktop, open the same issue title, description, or comment body in two clients, save different text from one client, then save the stale draft from the other; verify the second write returns `409` and the comparison UI preserves both versions.
5. Repeat while only changing an unrelated reaction, status, assignee, label, attachment, or position between opening and saving the text editor; verify the text save succeeds without a false conflict.
6. On mobile, edit an issue title or description after another client changed it; verify the request remains last-write-wins and does not enter a repeated `409` state.
7. Deliver an auxiliary revision event followed by an older full issue snapshot; verify the auxiliary projection remains ordered, the partial event does not claim the owner snapshot revision, and the stale full projection is refetched.
8. Exercise a pure no-op issue update and an attachment-binding update; verify the no-op preserves `revision`/`updated_at`, the attachment update advances once, and a binding failure rolls back the whole transaction.

## Checklist

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:**

Used Codex to trace issue and comment mutations end to end, model concurrent HTTP/WebSocket interleavings, separate aggregate snapshot ordering from field-level edit conflicts, implement revision guards across backend and clients, add regression tests, and run iterative review/preflight passes before publishing.

## Screenshots (optional)

Not included. The conflict and comparison states are covered by component tests.


Bohan-J (changes requested):
Thanks for this — the engineering quality is high: the CAS is genuinely atomic (conditional `WHERE ... revision = expected_revision`, no read-modify-write window), the migration has proper up/down, mixed-fleet behavior is handled explicitly, and the concurrency interleavings are well tested. Scoping to the edit-protection half of #7033 and leaving branch-runs out was the right call. The direction is right, but I'm requesting changes on one structural issue that I think must be settled **before** the 409 contract ships, plus one client bug that guarantees false conflicts.

## 1. Blocking: conflict detection uses an aggregate counter, so false 409s are structural, not incidental

`issue.revision` is not a row version in practice — it's an **aggregate version of the whole issue subtree**. `CreateComment` bumps it via the `touched_issue` CTE (`server/pkg/db/queries/comment.sql`), and so do reactions, labels, attachments, and metadata writes (`issue_reaction.sql`, `issue_label.sql`, `attachment.sql`, `issue.sql`). That means: while I spend three minutes editing a description, any agent posting a comment advances the revision, and my save hits 409 with a server-vs-local comparison dialog — **even though nobody else touched the description**. Multica issues are exactly the kind of object where agents comment continuously, so on active issues the false-conflict rate won't be an edge case; it will be the common case, and the conflict UI becomes noise.

The root cause is that one monotonic counter is serving two masters with opposite needs:

- The **realtime/cache monotonic-admission layer** genuinely needs aggregate semantics — a comment touches `updated_at`, and if that didn't bump the revision, list projections would drop the WS event as stale. That part of the design is correct and shouldn't change.
- The **edit-protection CAS** needs per-field semantics — "did the field *I'm writing* change since my base?"

No single counter can satisfy both, so picking a different granularity for the shared counter can't fix this. The two jobs need to be decoupled.

### Suggested direction

Keep `revision` exactly as is for realtime admission (that's the bulk of this PR and it stays valid). For edit protection on text fields, compare **content baselines** instead of the counter:

- The API already carries `description_base` (today used for channel-media preservation in `updateIssueWithDescriptionMerge`). Define the conflict check as: server-side `current_description != description_base` → 409; a hash comparison works if payload size is a concern. Add `title_base` for titles.
- This is exact per-field detection with zero new columns, degrades cleanly for legacy clients that omit the base, and is the same primitive an eventual explicit-merge feature (the other half of #7033's ask) needs anyway. It also gets two cases right that a version counter can't: two editors writing identical content (accept silently) and A→B→A reverts (nothing lost, no dialog).
- Comment bodies can keep the CAS as implemented — `comment.revision` only bumps on that comment's own content/resolution changes, so it's already effectively per-field.

## 2. Blocking: CAS blast radius — every issue write became conditional

`useUpdateIssue` (`packages/core/issues/mutations.ts`) injects the cached revision into **all** issue writes, including `moveIssue` — board drags, list status flips, assignee changes. A slightly stale cache now turns those into user-visible errors instead of writes. Those fields have no "lost typed content" failure mode; last-write-wins is the better trade there. I'd scope `expected_revision` (or the base-comparison above) to title / description / comment body only.

## 3. Blocking bug: client infers the next revision with local `+1`

`persistDescriptionSave` (and the title conflict handler) in `packages/views/issues/components/issue-detail.tsx` computes `nextRevision = revision + 1` after a successful save. But the server intentionally bumps only when the row actually changed, and concurrent activity can advance the revision by more than one. So the local counter drifts in **both** directions: one no-op save and every subsequent save falsely conflicts; one concurrent comment and the next save falsely conflicts. The authoritative revision is already in the response — adopt it instead of inferring.

## 4. Non-blocking, please confirm separately

- **`updated_at` semantics changed** from "always bump on update" to "bump only on actual change". The code's own comments note this timestamp is load-bearing (list ordering, daemon GC TTL). Probably an improvement, but it's a behavior change riding along — worth an explicit call-out and its own sign-off.
- The `ROW(...) IS DISTINCT FROM ROW(...)` block in `issue.sql` is duplicated (once for `revision`, once for `updated_at`); adding a field means editing four places. A trigger or CTE would collapse this.
- Consider splitting the PR: ① DB + API + 409 contract, ② realtime monotonic guards, ③ conflict UI. The cache layer (`cache-coordinator` / `ws-updaters`) is hot-path and ~4k lines land in one merge otherwise.

The reason I'm requesting changes rather than approving with comments: once the 409 `revision_conflict` contract is published, moving conflict detection from the aggregate counter to field baselines is a breaking change. Happy to discuss the base-comparison design here — the realtime plumbing in this PR is solid and I'd like to see it land.

yyclaw:
Addressed the requested concurrency changes in `ed001b9d4`:

- Decoupled text-edit conflict detection from the aggregate owner revision by using field baselines (`title_base`, `description_base`, and `content_base`). Unrelated reactions, status changes, moves, labels, or attachment activity no longer reject a valid text edit.
- Removed automatic `expected_revision` injection from normal non-text issue mutations, preserving last-write-wins behavior for drag/drop, status, assignee, parent, and similar updates.
- Removed client-side `revision + 1` inference. Queued saves now advance from authoritative mutation responses and accepted content baselines.
- Also added monotonic revision gates for mobile HTTP/WS cache updates and restored fresh owner snapshots for successful no-op issue writes followed by attachment events.

Validation completed:
- Go handler suite passed, including concurrent single-winner, unrelated-revision, true-content-conflict, and no-op snapshot ordering cases.
- Core: 1,416 tests passed.
- Views: 4,175 tests passed.
- Mobile: 68 tests passed.
- Web/Desktop/Core/Views and Mobile typechecks passed.
- Latest `upstream/main` merge-tree and patch-application checks are clean.

yyclaw:
Addressed the requested changes in `398bb2736`:

- Decoupled edit protection from the aggregate owner revision. Title, description, and comment-content edits now use field content baselines, while aggregate revisions remain the authoritative HTTP/WebSocket ordering signal. Unrelated comments, reactions, labels, attachments, or metadata changes no longer cause false text-edit conflicts.
- Removed blanket CAS from non-text issue writes. Status, assignee, drag/drop, parent, and other projection updates remain last-write-wins; explicit `expected_revision` is still supported for callers that intentionally want aggregate CAS.
- Removed client-side `revision + 1` inference. Successful mutations now propagate and adopt the authoritative server response.

On the separate `updated_at` point: accepted issue update calls continue to refresh `updated_at`. The general owner update also advances its revision on an accepted no-op so an immediately following attachment event cannot be published against a stale same-revision owner snapshot; idempotent auxiliary projections keep their revision unchanged.

I also renumbered the revision migration to `332` after the latest `main` claimed `327`–`331`. The new CI run is fully green, including backend, frontend, mobile, Windows, and installer checks: https://github.com/multica-ai/multica/actions/runs/32007105115

Could you take another look when convenient?

Bohan-J (changes requested):
First: `398bb273` resolves all three blockers from my previous review, and resolves them well — the baseline decoupling, the CAS scoping, and the response-adoption are exactly the right shape, and the channel-media-aware comparison (`baseWithLateMedia`) is a thoughtful touch I didn't ask for. Thank you for the fast, substantive turnaround.

However, a deeper pass over the updated head with concurrency interleavings in mind surfaced three new issues. Two are correctness bugs (one of which predates `398bb273` and I simply missed it last round — my apologies), and one is a set of acceptance-contract gaps. Requesting changes again for these.

## 1. Blocking: partial events advance the owner revision in cache, then block the fuller snapshot that carries the actual data

`patchIssueRevision` (`packages/core/issues/ws-updaters.ts` ~L218) merges a bare `{ revision }` into every cached issue projection when an auxiliary event arrives — and it's called from the comment/reaction/attachment/metadata handlers in `use-realtime-sync.ts` **and** from HTTP mutation success paths (`useCreateComment().onSuccess`, `useToggleIssueReaction().onSuccess` in `packages/core/issues/mutations.ts`). Meanwhile `onIssueUpdated` (~L299) silently drops any issue snapshot whose revision is below the cached one — no invalidate, just `return`.

Those two together break the invariant monotonic admission depends on: *cached state at revision N reflects the full owner state at revision N*. Concrete interleaving:

1. Cache holds `r1, title=A`.
2. User X saves `title=B` → commits as `r2`.
3. I post a comment → commits as `r3`; my own HTTP response advances my cache to `r3` via `patchIssueRevision` — **title still A**.
4. X's `issue:updated` event (`r2`, `title=B`) arrives → `r2 < r3` → dropped.

Nothing re-delivers `title=B`; my cache is wrong until an unrelated refetch. Note step 3 doesn't even need WS reordering — the HTTP-response path makes this race trivially reachable. `apps/mobile/data/realtime/issue-ws-updaters.ts` (`advanceIssueRevision` + `acceptsRevision`, ~L44–80) has the same structure.

Fix directions (any of these works): have auxiliary events carry and atomically merge the full owner snapshot; keep separate orderings for the full-entity projection vs auxiliary projections instead of one shared counter per cache entry; or force an authoritative refetch whenever a revision-only advance moves past the cached full snapshot. Either way, please add the regression test: apply full `r1/title=A`, then revision-only `r3`, then full `r2/title=B` — the observable end state must be `B` (directly or via triggered refetch), never a permanent `r3/title=A`.

## 2. Blocking: mobile sends strict baselines but has no way out of a real conflict

`apps/mobile/app/(app)/[workspace]/issue/[id]/edit.tsx` seeds `title_base` / `description_base` once (`seeded` guard, L61–69) and never updates them. On a real conflict the server correctly returns 409, but the screen only shows a generic "Failed to save" alert (L117–122); the stale bases are resent on every retry, so the user loops on 409 forever and can only Cancel and discard their edits. Strict mode without a reconciliation path is strictly worse than LWW for the person holding the draft.

Either bring the conflict UX to mobile (show server version vs local draft, "use server" / "keep mine" which re-bases and retries — the web `revision-conflict-compare` semantics), or have mobile omit the base fields for now and keep LWW until that UX ships. Sending strict baselines from a client that can't resolve the resulting conflict shouldn't ride along in this PR.

## 3. Blocking: the attachment step sits outside the issue update's atomic boundary, and two acceptance criteria are currently not met

- **The race**: `UpdateIssue` (handler, ~L3106–3150) commits the issue update first, then links attachments in a separate commit via `LinkAttachmentsToIssue`, passing `bump_revision = (issue.Revision == prevIssue.Revision)` — i.e. "don't bump if the update already did". If another mutation lands between the two commits, the attachment link mutates state at a later point in time while the handler publishes it under the older revision; monotonically-admitting clients will drop or misorder the attachment-carrying events. Link failures are also only `slog.Warn`ed while the response reports success. Please put the update and this request's attachment binding into one transaction (or have the link always return and publish the revision it actually advanced to), plus a deterministic interleaving test.
- **No-op writes**: plain `UpdateIssue` is now an unconditional `revision = revision + 1`, so a same-value write advances the revision — #7033's acceptance list explicitly says "No-op writes do not advance the revision." I understand this was a deliberate guard so a following attachment event can't publish against a same-revision stale snapshot — but that rationale is downstream of the boundary problem above. Once attachment binding is inside the transaction, please restore no-op detection; if you believe the criterion itself should change, raise it on #7033 rather than silently deviating.
- **"One atomic database statement"**: the issue-side baseline check is `SELECT ... FOR UPDATE` + conditional update inside one transaction, not literally one statement. To be clear, I think the implementation is *correct* — the row lock closes the interleaving window, and it's what makes the channel-media-aware comparison possible; the comment-side `content_base` in the UPDATE's WHERE does meet the letter. But since #7033 states the single-statement criterion explicitly, please either note the deviation and rationale on the issue for the author to sign off, or align the wording. I'm fine with the transaction-based design on its merits.

Item 1 is the most important — it's the same root disease as the last round (one counter carrying meaning it doesn't have), just relocated from the write path to the cache-admission path, and it can show a user stale text indefinitely, which is the exact failure class this PR exists to eliminate. Happy to re-review quickly once these land; the overall trajectory of this PR remains very good.

yyclaw:
Addressed the latest review in `7d2d8f17b`:

- **Full snapshots vs. auxiliary revisions:** auxiliary/revision-only events no longer advance the cached full owner's `Issue.revision`. Ordering is tracked independently per auxiliary cache projection, and a partial event that moves ahead invalidates the stale full owner projection for an authoritative refetch. Added core and mobile regressions for `full r1/title=A -> auxiliary r3 -> full r2/title=B`, including the HTTP-mutation path and independent auxiliary projections.
- **Mobile conflict recovery:** mobile now omits `title_base` / `description_base` and intentionally remains last-write-wins until it has conflict-reconciliation UX. A payload-level test locks this behavior; web/desktop retain strict baselines and the comparison UI.
- **Attachment atomicity and no-op semantics:** issue baseline/merge/update and attachment binding now run in one transaction with an attachment-before-issue lock order. Binding failures roll back the issue update; pure no-op writes preserve both `revision` and `updated_at`; a newly bound attachment advances the owner exactly once and returns/publishes the authoritative snapshot. Added deterministic interleaving and injected-failure rollback tests.

I also documented the transaction-vs.-literal-single-statement rationale on #7033 and asked for acceptance-wording confirmation: https://github.com/multica-ai/multica/issues/7033#issuecomment-5314637176

The PR description now reflects the mobile rollout boundary, auxiliary-event cache semantics, migration `332`, and atomic attachment/no-op behavior. All checks are green: https://github.com/multica-ai/multica/actions/runs/32018233932

Could you take another look when convenient?

Bohan-J (changes requested):
Re-reviewed at `7c454f81`. Thanks for the rework — the previous round's blockers are properly resolved, not papered over:

- Full issue snapshots and auxiliary projections (labels / metadata / properties) now carry separate revisions on web, so an older full snapshot is no longer swallowed by a newer partial event.
- Mobile text editing is explicitly LWW again, which removes the unrecoverable 409 loop in the edit screen.
- The issue update and its attachment binding are now in one transaction, with lock ordering, rollback, no-op non-advancement, and event ordering all covered by tests.
- The "single statement" vs "single transaction" wording is now called out on #7033. Given the channel-media-aware comparison needs the row lock, the transactional implementation is the right call; no objection from me.

One blocking issue remains, and it is the mirror image of the one you just fixed on web.

## Blocking: mobile auxiliary revisions are not tracked per projection, so a delayed partial event can roll the cache backwards

`apps/mobile/data/realtime/issue-ws-updaters.ts`

The auxiliary revision map is keyed by issue only:

```ts
// :45
function auxiliaryIssueRevisionKey(wsId: string, issueId: string) {
  return `${wsId}:${issueId}`;
}
```

and the label / issue-reaction patchers never consult it. They gate writes solely against the full `issue.revision` (`acceptsRevision`, `:108`), which — correctly, per the fix you just made — no longer advances on auxiliary events:

- `patchIssueLabels` (`:510-540`) applies when `acceptsRevision(issue.revision, issueRevision)` and writes only `labels`, leaving `issue.revision` where it was.
- `addIssueReaction` / `removeIssueReaction` (`:460-508`) do the same against the reactions embedded in the cached `Issue`.

So nothing in the mobile cache ever remembers "I already applied a labels event at r3", and the comparison baseline stays pinned at the full revision:

1. Cache holds full `r1`.
2. `issue_labels:changed` at `r3` arrives → `3 >= 1` → applied. Labels are now `[latest]`; `issue.revision` is still `r1`.
3. A delayed `issue_labels:changed` at `r2` arrives → `2 >= 1` → **also applied**. Labels roll back to `[stale]`.

Issue reactions have the same shape: a `removed` at `r3` followed by a delayed `added` at `r2` re-adds a reaction the user already saw disappear.

`onIssueAuxiliaryRevision` does invalidate the stale owner projections, so this partially self-heals — but not reliably enough to close the hole this PR exists to close: the rollback is rendered before any refetch resolves, and an inactive cache keeps the reverted value until something remounts it. On a mobile client, "inactive until remount" can be a long time.

**Why web is not affected** — worth stating explicitly, because the fix is already written:

- `packages/core/issues/ws-updaters.ts:28-71` keys the same map by `(wsId, issueId, projection)` with `AuxiliaryIssueProjection = "generic" | "labels" | "metadata" | "properties"`, and `patchIssueLabels` (`:552`) bails out early via `isOlderThanRecordedAuxiliaryProjection(..., "labels")`.
- Web keeps issue reactions in a separate `issueKeys.reactions` cache that it invalidates instead of patching in place (`use-realtime-sync.ts:1123-1140`), so there is no in-place value to roll back.

Mobile stores reactions inside the cached `Issue` and patches them in place, so it needs the projection-keyed guard that core already has.

### Requested change

Mirror core's per-projection tracking in the mobile updaters: add a projection dimension to the key (at minimum `labels`, `issue_reactions`, and a `generic` bucket), record it on every auxiliary event, and reject an event whose revision is below the recorded revision **for that projection** before touching the cache.

Please do not collapse this into a single global auxiliary maximum — a newer comment event would then incorrectly block the first valid event for an unrelated projection, which is the failure mode the current code is deliberately avoiding.

### Regression tests to add

1. Labels: full `r1` → labels `r3 = [latest]` → delayed labels `r2 = [stale]`; the cache must still read `[latest]`.
2. Issue reactions: full `r1` → reaction removed at `r3` → delayed reaction added at `r2`; the reaction must stay removed.

Both should assert on the cache value directly, not on eventual convergence after a refetch.

Everything else in this round looks good to me, and CI is green. This is the only thing between the PR and an approval from my side.

Bohan-J (approved):
Final review of `fdb09f4f`. **Approving.**

I re-verified every item from the three previous rounds against this head rather than trusting the fix descriptions:

- **Per-projection monotonicity (round 3)**: mobile now records auxiliary revisions keyed by `(workspace, issue, projection)` with `generic` / `labels` / `issue_reactions`, mirroring core's structure; labels and reaction patches gate on their own projection's recorded revision before touching the cache, and `reconcileIssueFullSnapshotRevision` cleans up absorbed records once a full snapshot catches up. The three regressions I asked for are all present and assert cache values directly (`rejects a delayed labels event after a newer labels projection`, `rejects a delayed issue-reaction add after a newer removal`, `does not let a newer generic event block the first labels projection`).
- **Cache admission (round 2, item 1)**: partial events no longer impersonate full-snapshot revisions on either platform; the `full r1 → revision-only r3 → full r2` interleaving is covered by tests and resolves to the newer content.
- **Mobile strict-mode dead end (round 2, item 2)**: mobile text editing is now deliberately LWW (no baselines sent) — the honest interim choice until a conflict UX ships there.
- **Atomic boundary (round 2, item 3)**: issue update + attachment linking now share one transaction with documented lock ordering (catalog → attachments → issue row); link failures roll back and surface as errors; the reloaded row keeps response/events on the actually-advanced revision. `UpdateIssue` is a single CTE statement with `did_change`-based no-op detection — which restores the "no-op writes do not advance the revision" criterion and, for this path, now literally satisfies "compare and update in one atomic database statement" as well. The remaining `FOR UPDATE`-transaction deviation for baseline checks is noted on #7033 as agreed.

CI is green across all 14 checks, the branch is mergeable against current `main`, and the migration renumber to `343` is clean. The scope remains the edit-protection half of #7033, with branch-runs-from-comments correctly left for its own PR.

Thanks for the persistence through four rounds — the final design (aggregate revisions for realtime ordering, content baselines for edit protection, per-projection admission for auxiliary events) is a genuinely solid piece of concurrency work, and each round of feedback was addressed at the root rather than patched over.

Bohan-J:
Approved and merging. Thank you for the work here — four rounds of review is a lot to sit through, and the quality of the responses is what made this mergeable.

Two things worth calling out specifically:

- When we pushed back that the aggregate `revision` was doing two incompatible jobs at once, you didn't tune the granularity and move on — you replaced the conflict check for text fields with content-baseline comparison and left the aggregate revision to the realtime layer. That's the harder change, and it's the right one. It also picked up the A→B→A and identical-content cases for free.
- The final mobile fix mirrors the projection-keyed guard from core exactly (`generic` / `labels` / `issue_reactions`), including the cross-projection independence case, and the regression tests assert on cache values rather than eventual convergence after a refetch. That's precisely what was asked for.

Verified before merge on `14299ab2`: migration renumbered to `351` (free on `main`; additive column adds with defaults, no index or constraint work), all 14 CI checks green, no conflicts with `main`, and the diff stays inside the subsystems it needs to touch.

Two follow-ups, neither blocking this PR:

1. Mobile text editing is deliberately last-write-wins in this version. That's the correct call for shipping now, but it does mean the protection is web/desktop only — worth a tracking issue for the mobile conflict UX.
2. The second half of #7033 — starting an independent run from a historical comment — is not in this PR and remains open.

Thanks again.

yyclaw:
@Bohan-J The merge conflicts against the latest `main` have been resolved, and the branch has been rebased. CI is green (15/15), and the existing approval is still in place.

Could you please merge this as soon as practical? `main` is moving quickly, so merging soon would help avoid another conflict-resolution round.

PR #7039, opened 2026-08-16
TRUE-39, MUL-6095: fix Antigravity TSV model IDs

True-Future:
## What does this PR do?

Fixes Antigravity model discovery and upgrade compatibility after stable model slugs were introduced in `agy 1.1.5`. `agy 1.1.13` is the version used to reproduce and verify the current behavior; JSON model output was introduced in `agy 1.1.12`.

The runtime now prefers `agy --output-format json models`, falls back to text discovery for older versions, and normalizes existing persisted model choices at execution time. Canonical IDs, legacy display labels, and malformed legacy `id\tlabel` values all resolve to the canonical ID before validation and final `--model` argv construction.

Closes TRUE-39

## Type of Change

## Changes Made

- Prefer the JSON model catalog available since `agy 1.1.12`; fall back to text for older installations.
- Parse tab-separated and aligned-space text rows while preserving legacy single-column output.
- Keep the first TSV field as the canonical ID even when later fields are added.
- Normalize exact canonical IDs, legacy display labels, and persisted `id\tlabel` rows at the execution boundary.
- Use the normalized ID for both catalog validation and final `--model` argv.
- Preserve current-main launch-prefix behavior by routing discovery through the runtime `Command` wrapper.
- Remove stale comments that described display labels as required or unknown models as silent successes.
- Add regression coverage for normalization precedence, validation/argv flow, JSON envelopes, aligned-space output, extra TSV fields, duplicates, blanks, and status lines.

## How to Test

1. `cd server && go test ./pkg/agent -run '^(TestBuildAntigravityArgsModel|TestAntigravityModelError|TestNormalizeAntigravityModel|TestAntigravityTSVModelIDFlowsToValidationAndArgs|TestParseAntigravityModels|TestParseAntigravityModelsEmpty|TestParseAntigravityModelsTSV|TestParseAntigravityModelsAlignedSpaces|TestParseAntigravityModelsJSON)$' -count=1`
2. `cd server && go vet ./pkg/agent/...`
3. Build Windows CLI with `CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build ./cmd/multica`.

The focused changed surface, `go vet`, and Windows CLI build pass on the current-main merge. The full `go test ./pkg/agent/...` and daemon suite are not green on this Windows host because existing tests use extensionless Unix-style fake executables, expect POSIX shell repair commands, inherit local runtime configuration, require symlink privileges, or exceed Windows path limits. One unrelated Codex service-tier assertion also fails.

No paid-model smoke test was run, and no daemon binary was replaced or restarted.

## Checklist

- [x] I have run scoped tests locally and documented unrelated suite failures
- [x] I have addressed all reviewer comments before requesting merge

## AI Disclosure

**AI tool used:** Multica Agent (Codex)

**Prompt / approach:** Reproduced the persisted-value compatibility gap test-first, verified expected red failures, implemented execution-boundary normalization plus JSON-first discovery and hardened text fallback, merged current main while preserving its command-wrapper behavior, then reran focused tests, static analysis, and a Windows build without deployment or paid inference.


multica-eve (changes requested):
The product direction is right: selecting an Antigravity model should work without forcing existing users to reconfigure their agents. However, this patch does not restore agents that already persisted one of the previously supported/broken model shapes, so I’m requesting changes.

1. **Must fix — normalize legacy persisted values before validation.** In `server/pkg/agent/antigravity.go:62-67`, `opts.Model` is still validated as an exact canonical ID. Agents that currently store `id\tlabel` (written from the malformed catalog) or a display label (the previously documented form) will therefore continue to fail after upgrading until the user manually reselects a model. Before `antigravityModelError`, normalize a tab-joined value to its ID and resolve an exact catalog label to its ID. Please add regression coverage that carries both legacy forms through validation and verifies that the final `--model` argv contains only the canonical ID.

2. **Please harden the TSV boundary.** `server/pkg/agent/models.go` currently uses `strings.Split` and rejects every row unless it has exactly two fields. If `agy models` adds another column, the catalog can become empty. Use a first-delimiter split (`strings.SplitN(line, "\t", 2)`) or otherwise preserve the first column as the machine ID, with a test for additional tab-separated content.

Non-blocking: please also update the stale `MULTICA_ANTIGRAVITY_MODEL` display-label guidance in `server/internal/daemon/agents_probe.go` and the remaining test comment that still says agy requires the display string.

The focused parser/validation/argv tests and CI coverage are otherwise solid; once the compatibility path is covered, this should be ready to merge.

multica-eve (changes requested):
After this change, people upgrading Antigravity should be able to keep their existing model choice and run tasks normally without selecting it again. The direction is right, but people who already saved a choice can still be stopped after upgrading, so the experience is not yet acceptable.

**Product verdict:** the fix is worth accepting once the upgrade path is seamless.

**Must fix — normalize legacy persisted values before validation.** In `server/pkg/agent/antigravity.go:62-80`, `opts.Model` is still validated as an exact canonical ID and then passed unchanged to the final argv. Agents may already store either the previously documented display label or the malformed `id\tlabel` row produced from the new catalog. Both values are rejected by the current patch—`server/pkg/agent/antigravity_test.go:547-553` explicitly asserts that rejection—so those agents still fail until users manually reselect a model.

Please resolve all supported persisted forms to the canonical ID at the execution boundary, then use that resolved value for both validation and `--model`. Prefer an exact ID match; normalize a non-empty legacy `id\tlabel` value to its first field; and map an exact catalog label to its ID. Add regression coverage that carries both legacy forms through the real validation/argv path and verifies that only the canonical ID reaches `--model`.

Non-blocking follow-ups:

- In `server/pkg/agent/models.go:2104-2107`, preserve the first TSV field instead of dropping a row whenever another column appears, and test that boundary.
- Update the stale display-label guidance in `server/internal/daemon/agents_probe.go:214-215`, `server/pkg/agent/models_test.go:1784-1786`, and `server/pkg/agent/antigravity_test.go:505-506`.

The focused tests, `go vet`, and current CI are green, and the canonical-ID path itself is clear. The compatibility gap above remains merge-blocking.

multica-eve:
I verified the upstream behaviour behind this change against the official Antigravity docs and changelog. Most of it supports the patch, but two findings change what the fix needs to do. Sharing them here so they land with the review above.

**Sources:** [CLI headless docs (v1.1.13)](https://antigravity.google/docs/cli/headless#select-a-model-effort-or-agent), [Models](https://antigravity.google/docs/models), [CHANGELOG](https://github.com/google-antigravity/antigravity-cli/blob/main/CHANGELOG.md).

### 1. The tab assumption needs confirmation — this is the one that could make the fix a no-op

The docs show `agy models` output as two **column-aligned** fields:

```
gemini-3.7-flash-high     Gemini 3.7 Flash (High)
gemini-3.1-pro-high       Gemini 3.1 Pro (High)
claude-sonnet-4-6         Claude Sonnet 4.6 (Thinking)
```

Rendered docs cannot tell a tab from space padding, and `parseAntigravityModels` only splits when the line contains `\t` — otherwise it falls through to `id = label = <whole line>`. If any build pads with spaces instead, the catalog ends up with IDs like `gemini-3.1-pro-high     Gemini 3.1 Pro (High)` and the bug is unfixed while all the new tests still pass.

You have 1.1.13 installed and I don't, so could you paste the raw bytes? `agy models | cat -A | head -3` (or `| xxd | head`) settles it. Regardless of the answer, I'd suggest making the text parser separator-agnostic: take the first whitespace-delimited token as the ID only when it looks like a slug (`^[a-z0-9][a-z0-9.-]*$`) and a remainder exists, otherwise treat the line as a legacy single-column display name. That covers tab, aligned spaces, and pre-slug output in one path, and it subsumes the `SplitN` suggestion in the review.

### 2. `agy models --output-format json` exists since 1.1.12 — prefer it

Changelog 1.1.12: *"Added machine-readable output to the `models` and `agents` subcommands through an `--output-format` flag accepting `json` and `stream-json`, and moved their error messages and progress spinner off stdout so captured output contains only the list."*

Trying JSON first and falling back to text parsing removes the separator ambiguity above, the exactly-two-columns fragility, and the need to special-case the `Fetching available models...` status line entirely (1.1.12 moved it off stdout). Text parsing still has to stay as the fallback for 1.1.5–1.1.11.

### 3. The old models were not removed, so legacy labels still map 1:1

The current catalog is Gemini 3.7 / 3.6 / 3.5 Flash, Gemini 3.1 Pro, Claude Sonnet 4.6 (Thinking), Claude Opus 4.6 (Thinking), GPT-OSS-120b — every display name we previously persisted still exists, only its identifier form changed (1.1.5: *"Added stable, user-facing model slugs ... and are accepted by `--model`"*). The new catalog's second column is written exactly like the old single-column value, so the exact-label-to-ID mapping the review asks for is unambiguous and needs no hand-maintained table. There is no "that model is gone" case that would justify dropping the compatibility path.

Worth splitting the persisted shapes by cost, since they are not equivalent:

- **Legacy single-column parsing (`ID == Label`): must stay.** `server/pkg/agent/version.go` has no `MinVersions` floor for antigravity, so a pre-1.1.5 `agy` still registers and still prints one column. Removing that path would need a version floor, which locks out pinned installs for a much larger cost than a few lines of parsing.
- **Persisted display label → slug: should stay.** This is the only "used to work, broken by an upgrade" regression, and our own pre-launch validation rejects it before `agy` ever runs, so the user does not even see agy's error listing the valid models.
- **The malformed `id\tlabel` row: reasonable to drop.** It only exists inside the broken window and those agents never ran successfully, so one manual reselect is a fair cost — provided the error says the stored identifier is stale and to reselect, rather than surfacing a bare invalid-model failure.

### 4. Two comments in the patch are now stale on the upstream side

The rationale carried in `antigravity.go` and `models.go` — that agy *"silently no-ops on a model string it doesn't recognise (empty output, exit 0)"* — stopped being true in 1.1.2, which changed print mode to hard-fail. Current docs: *"headless mode does not silently fall back when `--model` names an unknown model. It exits non-zero with an `ERROR` status"*, with the error naming the unresolved value and listing available models. Since we run agy in print mode, keeping the fail-open pre-launch validation is still fine, but the comments and the error text should reflect that agy now reports this itself.

Also, the slug change landed in **1.1.5**, not 1.1.13, so the regression window is wider than the PR description implies — anyone who upgraded past 1.1.5 has been affected, which makes the persisted-value path more likely to be populated, not less.

True-Future:
@multica-eve Review fixes are now at `f9a2ba60d`, merged with current `main`.

- Normalizes exact IDs, legacy display labels, and persisted `id\tlabel` rows before validation; only the canonical ID reaches `--model`.
- Prefers `agy --output-format json models` (available since 1.1.12) and falls back to text for older versions.
- Text fallback handles tabs, aligned spaces, legacy single-column rows, and extra TSV fields without losing the canonical first field.
- Preserves the new runtime `Command` launch-prefix behavior introduced on current main.
- Removed stale display-label and silent-success guidance, and corrected the history: slugs arrived in 1.1.5; 1.1.13 is the reproduction version.

Verification on the conflict-resolved tree: focused regression tests pass, `go vet ./pkg/agent/...` passes, Windows CLI build passes, and staged/unstaged diff checks are clean. The broader Windows suites still have unrelated extensionless-fake-executable, environment, symlink, path-length, and Codex service-tier failures documented in the PR description.

Please re-review when ready.

multica-eve:
Thank you for the comprehensive follow-up, especially the JSON discovery and legacy-normalization investigation. We have merged #6654 as the canonical, narrowly scoped fix for the tab-separated model-ID regression. This PR now overlaps that core change and also introduces broader discovery and compatibility behavior, so we are closing it as superseded to keep the repair scoped. If JSON-first discovery or legacy normalization remains necessary after the merged fix ships, please split it into a focused follow-up tied to a current reproduction. We appreciate the thorough work here.

PR #7045, opened 2026-08-16
MULT-32: enable Codex unelevated sandbox on Windows

bestvcboy:
## Summary

- persist Windows Codex sandbox provenance and keep explicit user/profile/daemon overrides authoritative
- treat registration ownership as a persistence/settings-preview hint only; task execution starts with a safe native-sandbox candidate, derives ownership from the config actually copied by Prepare/Reuse, and uses one task policy for effective argv and final spawn options
- serialize every `runtime_id` or `custom_args` update with one agent row lock; authenticate an explicit replacement against the snapshot it observed, then derive the effective runtime and persisted normalization from the locked row
- preserve override presence separately from value so present-empty Windows sandbox values are undecidable and fail closed to `workspace-write`
- make migration 327 down remove exactly one proven leading managed pair before dropping provenance
- align the built-in creating-agents guidance/source map with those boundaries

## Verification

- red regressions reproduced config add/remove drift, the original runtime-only lost update, the unsafe down migration, present-empty config/argv selecting the absent fallback, and a stale Windows→Linux request retaining a managed pair after the real runtime-only handler committed
- deterministic public-handler coverage now proves Linux→Windows and Windows→Linux interleavings, final runtime/argv/provenance/unrelated args, commit-time rollback and retry, and unchanged non-Windows behavior
- present-empty TOML and argv, valid `unelevated`/`elevated`, true absence, copied-config policy, and the real spawn builder have executable controls
- local gates passed: migrations 1–327, sqlc with no generated diff, `go build -buildvcs=false ./...`, full Go regular and race suites, focused handler/config/spawn suites, JavaScript tests (core 1,420; desktop 478; web 218; views 4,162), typecheck, lint with zero errors, and production builds
- independent review found no remaining correctness, security, or repository-standard issue; independent verification passed both public-handler interleavings and rollback/retry 10/10 plus focused race and empty-override suites
- exact head `591345db86d3e742000f47dd3f3fb1553b4edeed` passes all 10 GitHub checks, including native `windows-execenv`
- Git smart-HTTP timed out before sending; the identical local tree/parent was advanced through GitHub's Git Data API with an exact expected old head and `force=false`. No force-push, merge, or issue completion was performed.

Closes MULT-32
Refs MULT-48
Refs MULT-53


multica-eve (commented):
Today, Windows agents can use the files, credentials, and tools available to the account that runs them. This PR would change every new and existing Windows agent so it can write only inside its task folder by default. That reduces the risk of accidental changes elsewhere, but it can also turn previously working dependency installation, host-tooling, and external-file workflows into failures that are difficult for users to diagnose.

## Product decision

❌ We are not going to adopt this behavior as the Windows default, so we are closing the PR.

The current full-access behavior is deliberate. Multica runs agents unattended and defines the daemon user, container, or VM as the actual isolation boundary. Tasks are expected to install dependencies and use host tools and credentials such as `gh`, `aws`, and `kubectl`; a partial filesystem boundary can break those workflows while still allowing credentials to be read and sent over the network. Users who want the native Windows sandbox can already opt in explicitly, and Multica preserves that choice. This is documented in our [security model](https://github.com/multica-ai/multica/blob/main/apps/docs/content/docs/security-model.mdx).

This PR would therefore change an existing product default rather than add a missing capability. It would also make Windows behavior diverge from Linux and macOS without creating a consistent isolation guarantee. In addition, OpenAI documents `unelevated` as the weaker fallback and recommends `elevated` when available, so making the fallback universal is not the security posture we want to promise by default.

There is also no complete product path for this change yet: users are shown managed arguments rather than a clear full-access/elevated/unelevated choice with consequences and rollback, the public security documentation still describes explicit opt-in, and the validation does not run the original `multica issue create` journey or representative dependency/host-CLI workflows through a real sandboxed Windows Codex process.

This is a product rejection, not a dismissal of the implementation quality. The precedence handling, provenance tracking, race coverage, and test work are thoughtful, and the checks are green. If we revisit this direction, we would prefer a smaller, explicit runtime or agent setting with a visible effective mode, an easy rollback path, and end-to-end validation across supported Windows environments before considering any staged default change.

Thank you for the substantial effort, careful testing, and responsiveness across the review iterations. We genuinely appreciate the contribution even though we are not taking this product direction.

PR #7049, opened 2026-08-17
MUL-6265 Support forwarded Slack message content

tim-smart:
## Summary

- extract forwarded Slack message unfurls from the original Events API payload
- append forwarded content with author attribution while excluding ordinary attachment previews
- keep files nested in forwarded attachments out of normal media ingestion

## Validation

- `go test ./internal/integrations/slack -run '^TestInboundFromMessage_(SeparatesForwardsFromOrdinaryAttachments|DM|ChannelMention)$' -count=1`
- `go test ./internal/integrations/slack -count=1`
- `go vet ./...`
- `go build ./...`

Closes TIM-27

### DIFF
--- a/server/internal/integrations/slack/inbound.go
+++ b/server/internal/integrations/slack/inbound.go
@@ -175,6 +175,12 @@ func buildInbound(e slackevents.EventsAPIEvent, p buildInboundParams, mentionRe 
 		reply = &channel.ReplyCtx{MessageID: p.threadTS, RootID: p.threadTS}
 	}
 	text := cleanText(p.text, mentionRe)
+	if forwarded := forwardedMessageText(e); forwarded != "" {
+		if text != "" {
+			text += "\n\n"
+		}
+		text += "Forwarded message:\n" + forwarded
+	}
 	return channel.InboundMessage{
 		EventID:        p.ts,
 		MessageID:      p.ts,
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
Requesting changes for one correctness issue in the history path.

`server/internal/integrations/slack/history.go:362` treats every attachment with a non-empty `Ts` as a forwarded message. `Ts` is a generic `slack.Attachment` field, not a forwarding discriminator; normal Incoming Webhook/classic attachments can set it as well. A message with top-level text and an ordinary timestamped attachment will therefore be rendered in history with that attachment under `Forwarded message:`, even though the live Events API path correctly excludes it via `is_msg_unfurl`. This can give the agent incorrect context and makes live ingestion disagree with history replay.

Please preserve/read the authoritative `is_msg_unfurl` marker in the history path (or use another guaranteed discriminator), and add a regression test proving that an ordinary attachment with a non-empty `Ts` is not appended as forwarded content. The current test at `server/internal/integrations/slack/history_test.go:303` only covers an ordinary attachment whose `Ts` is empty.

tim-smart:
Pushed a simplification pass as amended head `2bf10ec8d` (behavior unchanged, `is_msg_unfurl` still the authoritative discriminator on both paths):

- Replaced the HTTP `RoundTripper` interception (shared mutable flag state, two mutexes, request serialization) with a direct `conversations.history` / `conversations.replies` call that decodes the raw response once. `historyMessage` now embeds `slack.Message` and shadows `Attachments` with the same `forwardedAttachment` type the Events API path already uses, so the marker survives decoding without a side channel.
- Live ingestion and history now share one `forwardedAttachmentsText` renderer instead of two near-identical loops.
- History tests set `IsMessageUnfurl` directly on attachments instead of zipping parallel flag slices; the regression coverage for timestamped ordinary attachments is intact, and the new decode test pins the raw wire decode.

`go test ./internal/integrations/slack -count=1`, `go vet`, and a full `go build ./...` pass.

multica-eve (changes requested):
Thanks for fixing the original `Ts` heuristic issue. The authoritative `is_msg_unfurl` decode and its negative regression test resolve that concern.

Before merge, please also address the remaining robustness gaps introduced by the direct Web API path:

1. Make the HTTP client and Slack API base URL injectable instead of hard-coding `http.DefaultClient` and `https://slack.com/api/` in `server/internal/integrations/slack/history.go:76-104`. Add `httptest` coverage for method selection (`conversations.history` vs `conversations.replies`), form fields (`channel`, `ts`, `latest`, `limit`), the Bearer header, non-200 responses, and Slack `ok=false` responses. The current fake stops above the new transport boundary, so these production request semantics are otherwise untested.

2. Improve the non-200 path at `server/internal/integrations/slack/history.go:97-99`: consume/drain the response body as appropriate for connection reuse and retain actionable response information, especially `Retry-After` for 429s (ideally preserving the SDK-equivalent rate-limit semantics).

3. Keep live ingestion and history rendering identical when the top-level message text is empty. At `server/internal/integrations/slack/history.go:339-354`, the empty-text branch currently flattens attachments directly, so the same forwarded payload loses both the `Forwarded message:` label and author attribution in history. Route forwarded attachments through the shared renderer in both branches and add a regression test for an empty top-level text payload.

4. Remove or contain the attachment-shadowing trap in `historyMessage` (`server/internal/integrations/slack/history.go:49-52`). After decode, `m.Attachments` is populated while `m.Message.Attachments` is always empty; passing the embedded `slack.Message` to a future helper can silently discard attachments. Prefer a representation/helper boundary that exposes one authoritative attachment view, and pin that invariant in tests.

These items are required for the next review round.

PR #7057, opened 2026-08-17
docs: add custom runtime profile verification runbook (MUL-6275)

ybkin1:
## Summary

Documents the `multica runtime profile` feature (MUL-3284): a workspace-scoped custom runtime profile that reuses a **supported protocol family** for routing while launching a site-specific `command_name` (wrapper / pinned binary / internal CLI).

This is a verification runbook — how to wire such a CLI into a deployment, how to verify registration and execution, and the compatibility contract the wrapper must satisfy (argv shape per family, minimal stream-json event contract).

## Content

- What a profile is / is not (no new protocol families; family immutable after creation)
- The `SupportedTypes` whitelist with representative daemon-side CLI contracts (claude, qwen, grok, qoder, dsh, traecli)
- Step-by-step: prepare wrapper → `multica runtime profile create` → confirm registration → smoke-test a task → update/delete → `set-path`/`unset-path` per-machine pinning
- Compatibility checklist for a custom CLI
- Reference integration path for contributing a new protocol family (dsh as the example: backend, probe, SupportedTypes lockstep test, migration)
- Troubleshooting table

All claims are derived from the current code: `server/cmd/multica/cmd_runtime_profile.go`, `server/pkg/agent/agent.go` (`SupportedTypes`), `server/pkg/agent/{claude,qwen,dsh}.go`, `server/internal/daemon/agents_probe.go`.

### DIFF
--- a/docs/custom-runtime-profiles.md
+++ b/docs/custom-runtime-profiles.md
@@ -1,0 +1,172 @@
+# Custom Runtime Profiles
+
+Custom runtime profiles (MUL-3284) let a workspace register a site-specific
+agent executable — a wrapper, a pinned binary, an internal CLI — while reusing
+one of Multica's **supported protocol families** for task routing and result
+parsing.
+
+This guide is the verification runbook for wiring such a CLI into a Multica
+deployment.
+
+## What a profile is — and is not
+
+- A profile lives **server-side, scoped to a workspace**. Members of the
+  workspace can bind agents to it.
+- A profile declares:
+  - `protocol_family` — one of the whitelisted agent types below. This is the
+    *routing backend*: Multica drives the task with the family's CLI contract
+    (argv + output protocol). **The family is immutable after creation.**
+  - `command_name` — the executable the daemon resolves on each host's `PATH`
+    at registration time. This can be any binary you control (e.g. a wrapper
+    that injects credentials).
+  - `display_name` / `description` — human-readable metadata.
+- A profile does **not** introduce a new protocol family. If your CLI speaks a
+  protocol Multica does not yet route, you need a code contribution (see
+  [Contributing a new protocol family](#contributing-a-new-protocol-family)).
+- `fixed_args` (command + args parsed in the product UI) are stored with the
+  profile; the CLI surface currently exposes only `command_name`.
+
+## Supported protocol families
+
+The canonical whitelist (single source of truth: `server/pkg/agent/agent.go`
+`SupportedTypes`, mirrored by the `runtime_profile.protocol_family` CHECK
+constraint):
+
+```
+claude, codebuddy, codex, copilot, opencode, deveco, openclaw, hermes, pi,
+cursor, kimi, reasonix, dsh, kiro, antigravity, qoder, qoderclicn, traecli,
+grok, qwen, qwenpaw
+```
+
+Representative CLI contracts you must match when choosing a family:
+
+| family | invocation shape (daemon side) |
+|---|---|
+| `claude` | `claude --output-format stream-json --input-format stream-json ...` — prompt via stdin, events on stdout |
+| `qwen` | `qwen -p <prompt> --output-format stream-json ...` |
+| `grok` | `grok agent --always-approve stdio` (ACP backend) |
+| `qoder` / `qoderclicn` | shared ACP backend |
+| `dsh` | detection via `dsh --profile multica --probe`; launch args + env from the backend |
+| `traecli` | dedicated backend with its own launch header |
+
+Pick the family whose argv contract your wrapper can satisfy — the daemon
+will call your `command_name` with that family's arguments.
+
+## Prerequisites
+
+- A running Multica daemon on the host (it resolves `command_name` at
+  registration time, so the wrapper must exist **before** you create the
+  profile).
+- `multica` CLI signed into the workspace.
+- Your wrapper on `PATH`, or pinned per-machine with `set-path` (below).
+
+## Verification runbook
+
+### 1. Prepare the wrapper
+
+Make an executable that:
+
+1. Accepts the chosen family's argv. Self-check by hand: run it with the
+   exact arguments from the table above and confirm it parses them.
+2. Emits output the family's parser understands. For `stream-json` families
+   the daemon consumes line-delimited JSON events; the minimal contract is a
+   start event, assistant turns, and a final `result` event carrying the task
+   text. Verify with a manual run:
+
+   ```sh
+   your-wrapper -p "hello" --output-format stream-json | head -5
+   ```
+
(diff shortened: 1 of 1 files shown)


Bohan-J (changes requested):
Thanks for putting this together — the cross-checking against code is clearly deliberate, and a lot of it holds up exactly: the 21-entry family list matches `SupportedTypes` verbatim, the CHECK constraint mirroring, migration 313 = dsh, the `dsh --profile multica --probe` detection, workspace scoping, family immutability, the 409-on-bound-agents semantics, and `set-path` being local-only. All verified correct.

Two things need resolving before this can land: **where the doc lives**, and **six factual errors**, two of which would send a reader down a dead end.

## 1. Placement

`docs/custom-runtimes.md` — covering this same feature — was deleted in c1ec1f464 (#6364), with the reason stated in the commit message: *"engineering scratch that agents load as context on every run without ever being read by a human"*. The same series removed `product-overview.md` and `design.md` and reduced `docs/` to `assets/` only.

Meanwhile `apps/docs/content/docs/daemon-runtimes.mdx:82` already has a **Custom runtime profiles** section — and it is more accurate than this runbook on the two most consequential points below. It already states that the command field takes an executable and arguments rather than a shell script, and its troubleshooting list already says to run `multica daemon restart` after a tool install, a path change, or a profile update.

Suggestion: fold the genuinely new material (verification steps, compatibility checklist, troubleshooting table) into that existing section rather than adding a new `docs/*.md`. Note it is a 4-language site, so translations need to follow.

## 2. Factual corrections

### a. `set-path` is missing the mandatory daemon restart — and the troubleshooting table presents it as an alternative

The override is read once when the daemon loads its config (`server/internal/daemon/config.go:225-227`) and never re-read afterwards. The CLI itself prints this on success:

```go
// server/cmd/multica/cmd_runtime_profile.go:316
fmt.Println("Restart the daemon for the change to take effect.")
```

§6 never mentions it. Worse, the first troubleshooting row reads:

> `command_name` not on `PATH`; pin with `set-path` **or** restart the daemon

Both are required, not either. Someone following this will pin the path, see nothing change, and get stuck.

### b. Registration timing is described backwards

§3 says the daemon "discovers the wrapper from `PATH` on its normal refresh cycle (restarting the daemon picks it up immediately)".

Custom profiles do not go through discovery polling at all — both loops skip them explicitly:

```go
// server/internal/daemon/agents_refresh.go:557 and :618
if !ok || rt.ProfileID != "" { continue }
```

with the accompanying comment: *"Custom profile add/edit/disable remains owned by the drift path"*. The actual mechanism is a server WS change notification driving `refreshWorkspaceRuntimeProfiles`; any change to the profile-set signature triggers re-registration. `server/internal/daemon/daemon.go:3150-3151` puts it plainly — profiles "become visible without a daemon restart".

So creating a profile is actually *better* than documented: effectively immediate, no restart needed. The doc tells people to restart anyway.

The case that genuinely does need a restart is the other one — profile already exists, wrapper installed afterwards. The profile-set signature is unchanged, so nothing re-registers. Worth splitting these into two sentences; it also gives the Prerequisites line its real justification.

### c. `grok` argv is missing `--no-auto-update`

```go
// server/pkg/agent/grok.go:139
grokArgs := []string{"--no-auto-update", "agent", "--always-approve"}
// then optional --effort <level>, then "stdio"
```

The table says `grok agent --always-approve stdio`, and unlike the other rows it carries no trailing `...`, so it reads as a complete invocation. A wrapper written to that spec breaks on the unexpected flag.

Root cause worth flagging: this row and the `qwen` row are verbatim from the **comments** at `server/pkg/agent/agent.go:282-283`, and those comments have drifted from `grok.go`. The PR description says all claims are derived from current code; they are derived from code comments. For a document whose entire value is precision, that distinction matters — consider verifying against the `build*Args` functions directly.

### d. `qwen` / `claude` omit daemon-injected args

- `qwen`: `buildQwenArgs` unconditionally appends `--yolo` (`server/pkg/agent/qwen.go:60`) — the comment there notes it is daemon-owned because non-interactive mode otherwise filters out every approval-requiring tool.
- `claude`: the real argv starts with `-p` and always includes `--verbose --permission-mode bypassPermissions --disallowedTools AskUserQuestion` (`server/pkg/agent/claude.go:717-731`).

The trailing `...` partly covers this, but a compatibility checklist exists precisely so a wrapper author knows which flags will always arrive.

### e. `command_name` must be a single token — and the doc says the opposite

```go
// server/internal/handler/runtime_profile.go:103-113
if strings.ContainsAny(commandName, " \t\r\n") {
    return errors.New("command_name must be a single executable token; put arguments in fixed_args")
}
```

§2 says "there is no server-side validation of the executable". That is true about existence, but a reader will follow it into `--command-name "your-wrapper --mode x"` and get a 400.

Related: the runbook never states that the daemon execs the command directly, without a shell. That was the strongest point in the deleted `docs/custom-runtimes.md`, and it is already correct on the docs site. For a document framed as a compatibility contract, it belongs here more than most of the current content does.

### f. 409 response field name

The doc says "409 (machine-readable `message`)". The body is `{"error": "..."}` (`server/internal/handler/handler.go:471`), with some paths adding `{"code": "runtime_delete_not_drained"}` (`server/internal/handler/runtime_profile.go:424,439`). There is no `message` field.

## Nits

- Last troubleshooting row — "Profile disappears after restart | `enabled=false` or drift reconciliation": drift converging to zero is a *consequence* of the profile being disabled or deleted, not an independent cause. Listing the two side by side points troubleshooting in the wrong direction.
- The §5 `update` example omits `--command-name` and `--enabled` (both registered in `cmd_runtime_profile.go`). `--enabled` is exactly the remedy for the `enabled=false` row in the troubleshooting table, so the two sections currently disagree.
- Worth adding to the table: when the pinned override path is not executable, the daemon **silently falls back to `PATH`** rather than erroring (`server/internal/daemon/daemon.go:2762-2775`). "I pinned it, but it's launching a different binary" is very hard to diagnose without a note.

---

The gap this fills is real — the existing docs section covers creating a profile but not verifying one or the compatibility contract a custom command has to satisfy. Worth landing once placement is settled and (a)–(f) are corrected; (a) and (b) are the priority, since both actively mislead.

ybkin1:
Quick status note from a self-host user: we are staging a self-hosted deployment (0.4.28) and are wiring an internal headless agent CLI (ZCode) as a custom runtime profile reusing the `claude` stream-json protocol family. This runbook is exactly the doc we hand to the people setting up those wrappers, so it would be great to get maintainer feedback on it. Happy to adjust anything that doesn't match current conventions.

Bohan-J:
@ybkin1 Yeah please check the request changes above

PR #7061, opened 2026-08-17
feat(integrations): add outbound webhook for action-required events

### LINKED ISSUE #1020: [Feature]: Outbound webhook notifications for action-required events
## Problem

Multica's notification layer is inbox-only — there is no outbound delivery channel. When an agent moves an issue to `in_review` or a task fails, the only signal is an item in the in-product inbox.

This means humans have to **actively check Multica** to know when their attention is needed. For a tool positioning itself as AI-native, this inverts the intended dynamic: humans should be reactive (pulled in only when a decision is needed), not proactive (constantly checking for updates).

The gap is especially visible in lightweight, non-coding tasks where agents finish quickly and review is trivial — the bottleneck isn't the work, it's the human not knowing it's done.

## Current behavior

- `status_changed` → `in_review` creates an inbox item with severity `info`
- `task:failed` creates an inbox item with severity `action_required`
- No emails, no webhooks, no external push — severity has no effect outside the product

## Proposed solution

Add **outbound webhook support** at the workspace level. When a notification of severity `action_required` is created, POST to a configured URL with a standard payload.

This is intentionally minimal — teams wire it to whatever they use (Slack incoming webhook, email relay, PagerDuty, etc.) without Multica needing to integrate with each service directly.

### Suggested scope

- Workspace-level webhook URL configuration (one URL per workspace to start)
- Trigger on `action_required` severity events: `in_review` status, `task:failed`, `issue_assigned`
- Standard JSON payload: event type, issue title, issue ID, actor, workspace ID
- Delivery failures logged but non-blocking (best-effort)
- Toggle to enable/disable per workspace

### Why this matters for the AI-native vision

The roadmap direction (per #815) is progressively shifting from human-led to AI-led execution, with humans at intentional gates. Webhooks are the mechanism that makes those gates work in practice — without them, humans can't reliably know when a gate needs them, and the workflow stalls silently.

## Out of scope (for now)

- Per-user notification preferences
- Native Slack / email integrations (can be built on top of webhooks)
- Inbound webhooks (separate concern, see #281)

cxx-12333:
## What does this PR do?

Multica's notification layer is inbox-only: when an agent moves an issue to `in_review`, a task fails, or an issue is assigned, the only signal is an in-product inbox item — humans have to actively check Multica to know their attention is needed (#1020).

This PR adds workspace-level outbound webhooks. When one of the three action-required conditions occurs, the server POSTs a JSON payload to a URL configured in the workspace settings. Teams wire it to whatever they already use (Slack incoming webhook, email relay, PagerDuty, ...) without Multica integrating each service natively.

The implementation follows the existing Slack outbound pattern (`server/internal/integrations/slack/outbound.go`): a small `Outbound` struct with a narrow query interface subscribes to `issue:updated` and `task:failed` on the event bus, filters for the action-required conditions, reads `webhook_url` from the workspace `settings` JSONB at fire time, and POSTs a standard payload. Delivery is best-effort with a 10s timeout — failures are logged and dropped, never blocking the bus.

## Related Issue

Closes #1020

## Type of Change

## Changes Made

- `server/internal/integrations/webhook/outbound.go` (new) — the outbound subscriber. `issue:updated` fires `issue_assigned` when a new assignee of an inbox-owning type (member/agent) takes the issue and `issue_in_review` on transition into `in_review`; `task:failed` fires only on terminal failure, not while a retry is pending. It reads `webhook_url` from `workspace.settings` JSONB at fire time and POSTs `{"event", "issue_id", "issue_title", "actor_type", "actor_id", "workspace_id", "details?"}` with an `X-Multica-Event` header; 10s timeout; log-and-drop on any failure.
- `server/internal/integrations/webhook/outbound_test.go` (new) — 12 tests: payload-contract assertions on every trigger (body fields + header), negative gates (unassignment, squad assignee, non-in_review status, map-dialect payload, retry-pending, missing issue_id), GetIssue-failure tolerance, and non-blocking server errors.
- `server/cmd/server/router.go` — one line wiring `webhook.NewOutbound(queries, slog.Default()).Register(bus)` next to the Slack outbound registration.
- `apps/docs/content/docs/webhook-notifications*.mdx` + `meta.json` (separate `docs` commit) — user documentation following the existing integration-page pattern (en/zh/ja/ko), registered under Integrations in the docs sidebar.

Two contract decisions worth calling out, both mirroring the existing notification listeners (`cmd/server/notification_listeners.go`):

- **Payload dialect**: the bus is in-process with no serialization, and `issue:updated` is published in two shapes — user/agent mutations carry a `handler.IssueResponse`, background resets (`broadcastIssueUpdated`) carry a map. The notification listeners type-assert the struct and skip the map on purpose; this outbound does the same, so background status resets don't fire webhooks.
- **Trigger classification**: `issue_assigned` fires only when a new assignee of recipient type exists (unassignment is info severity in the notification system and never reaches a webhook). One deliberate deviation: `task:failed` skips retry-pending failures even though the notification system notifies on every failure — a webhook receiver would otherwise be paged once per retry of the same run.

Configuration: `PATCH /api/workspaces/{id}` with `{"settings": {"webhook_url": "https://..."}}` — one URL per workspace; absent or empty URL disables delivery (per-workspace toggle). No schema change, no new table.

## How to Test

1. `cd server && go test ./internal/integrations/webhook/` — all 12 tests pass; `go vet` on the touched packages and `go build ./...` on the full server tree are clean. (Docs typecheck `pnpm --filter @multica/docs typecheck` not run locally — no pnpm on the authoring machine; the pages mirror the structure of the existing integration pages.)
2. Manual end-to-end: set a workspace's settings to `{"webhook_url": "https://webhook.site/<your-id>"}`, then (a) assign an issue, (b) move an issue to `in_review`, (c) let a task fail terminally — a JSON POST arrives for each, with `X-Multica-Event` set to `issue_assigned` / `issue_in_review` / `task_failed`.
3. Stop the receiver (or point at a 500-ing URL): the actions still succeed, only a WARN log is emitted — delivery failures are non-blocking.
4. Remove `webhook_url` from settings: no POSTs, no logs, no errors.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — no UI change; configuration goes through the existing workspace settings field
- [x] I have updated relevant documentation to reflect my changes — `webhook-notifications` pages (en/zh/ja/ko) added under Integrations in a separate `docs` commit
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and docs — N/A (server-side only)
- [x] If this PR touches Chinese product copy, I checked it against the conventions doc — N/A (no product copy touched)
- [x] I have considered and documented any risks below

Risks / notes:
- Webhook URL is stored and delivered in plaintext over HTTPS per #1020's minimal scope; no signature header (HMAC) yet — worth a follow-up if maintainers want receiver-side verification.
- `webhook_url` lives in the workspace `settings` JSONB. Verified safe against the settings UI: `UpdateWorkspace` only rewrites `settings` when the PATCH body includes it (`req.Settings != nil`, `internal/handler/workspace.go:393`), and no production frontend caller sends that field — the workspace tab auto-saves name/description/context/prefix only. Residual semantics: any client that PATCHes `settings` replaces the whole object, so integrations should merge rather than overwrite. Happy to move storage to `channel_installation` (the table Slack/Lark credentials use) if maintainers prefer.
- Payload delivery is synchronous with the event bus (same trade-off as the Slack outbound); a slow receiver is bounded by the 10s client timeout, worst case adding that latency to the triggering write.
- `task:failed` skips retry-pending failures even though the notification listeners notify on every failure — intentional, to avoid paging the receiver once per retry of the same run; easy to flip if maintainers prefer strict mirroring.

## AI Disclosure

**AI tool used:** ZCode (Claude-based coding agent), GLM via LiteLLM

**Prompt / approach:** I studied the Slack outbound and the event bus internals first, then asked the agent to implement #1020 strictly within the issue's suggested scope and the existing outbound pattern — narrow query interface, bus subscription, best-effort delivery. Multiple earlier iterations were discarded for over-engineering (retries, HMAC, per-event subscription UI) before settling on this minimal version; the failing-test fix (valid UUIDs in test fixtures) and the final PR description were also agent-assisted.

## Screenshots (optional)

N/A — server-side change; example delivered payload:

```json
{
  "event": "issue_in_review",
  "issue_id": "825e846a-156f-47d3-82b0-a99a6d070b7d",
  "issue_title": "Add export button to settings page",
  "actor_type": "agent",
  "actor_id": "gl-dev-01",
  "workspace_id": "12345678-1234-1234-1234-123456789012",
  "details": { "status": "in_review" }
}
```


gogodjzhu:
can't wait for this feature, anybody help?

PR #7074, opened 2026-08-17
MUL-6296 fix(daemon): garbage-collect workspace roots abandoned by a profile change

### LINKED ISSUE #6962: [Bug]: Workspace roots abandoned after a profile change are never garbage-collected
### What happens

The GC loop only ever walks a single workspace root — the one the running daemon's profile resolves to. Any root the daemon used previously becomes permanently invisible to it: never swept, never reported as reclaimable by the daemon itself, never removed. On an install that has switched to a named profile, everything left in the old root is stranded for good.

Disk-usage reporting, by contrast, is already multi-root. That asymmetry is the bug: `multica daemon disk-usage --all-profiles` will happily show you bytes that nothing in the system will ever reclaim.

### Why a root gets abandoned

`ResolveWorkspacesRoot` (`server/internal/daemon/config.go:569`) derives the root from the profile:

- default profile → `$HOME/multica_workspaces`
- named profile → `$HOME/multica_workspaces_<profile>`

So the moment an install starts using a named profile — which Desktop installs do — the previous root stops being the configured one. There is no migration step and no final sweep of the old location.

### Where it goes wrong

`runGC` reads exactly one directory and there is no second root anywhere in the loop:

```go
// server/internal/daemon/gc.go:79
root := d.cfg.WorkspacesRoot
```

Meanwhile `enumerateDiskUsageRoots` (`server/cmd/multica/cmd_daemon.go:1922`) walks the default root **plus every directory under `~/.multica/profiles/`**, which is why the reporting side can see what the collection side cannot.

### Evidence

From a Windows install on v0.4.25, reported by @cfigueiroa in #6782 (data quoted with their sizes/ages/counts verbatim):

| Root | Task homes | `.sandbox-bin/codex.exe` copies | References in the running daemon's log |
|---|---|---|---|
| `multica_workspaces_<profile>\` (current) | 36 | 1 | 433 |
| `multica_workspaces\` (legacy, pre-profile) | 49 | 4 | **0** |

The legacy root holds 221,097 files / 26,716,129,301 bytes (24.88 GiB), including four ~285 MiB `codex.exe` copies roughly three days old — past `artifact_ttl=12h`, past `ttl=24h`, and past `orphan_ttl=72h`.

The GC itself is healthy on that machine: four completed cycles with `cleaned=10..19` each, and zero `gc: batch issue check failed`. The sweep logic is fine too — an archived log from when the legacy root was still the active one shows it working exactly as intended:

```
gc: cycle complete ... artifact_dirs=25 artifact_removed=49 bytes_reclaimed=19858171751 by_pattern="map[managed:codex-home/.sandbox-bin:25 node_modules:24]"
```

It stopped matching because the daemon moved, not because anything broke.

### Impact

Silent, unbounded disk retention on any install that has ever changed profile. It is invisible from `multica daemon disk-usage` without `--all-profiles`, and even with that flag the reported bytes are not actually reclaimable by any command. Manual deletion is the only recovery.

### The obvious fix is not safe as written

Making GC enumerate roots the way `enumerateDiskUsageRoots` does would introduce a worse bug. The guard preventing GC from deleting a directory a task is actively using is `isActiveEnvRoot`, which is **in-process memory**. A daemon sweeping another profile's root cannot see that profile's live tasks, so two daemons running concurrently under different profiles could delete each other's in-flight task workdirs.

Any fix needs an ownership check first — per-profile `daemon.pid` liveness or a health-port probe — or should be an explicit operator-invoked command rather than an automatic sweep. Worth deciding that before implementing.

### Not a duplicate of #6782

#6782 reported duplicate sandbox executables caused by the managed-cache reclaim being wired into the issue decision path only; that was fixed in #6794 and shipped in v0.4.24. This is a distinct problem in a different layer — GC never looks at the directory at all — and #6794 cannot help with it.

Reported by @cfigueiroa, whose diagnostics across three separate installs are what isolated this.

Kickflip73:
## Summary

`multica daemon`'s garbage collector only ever scanned `d.cfg.WorkspacesRoot` — the current profile's workspace root. Once an install switched to a named profile, everything left behind in the old profile's root was stranded forever: GC never visited it, so it could grow unbounded. Disk-usage reporting already walked all roots, but reclamation did not.

## Fix

Teach GC to walk every profile's workspace root:

- `gcWorkspaceRoots()` now returns the current profile's root plus any other profile root whose owning daemon is not running.
- Ownership is checked per profile via its `daemon.pid` (a live daemon keeps its PID file), so GC never reclaims another live daemon's in-flight workdirs.
- Roots that have never been used (no workspace dir) are skipped, matching the previous single-root behavior.

## Testing

- `go build ./cmd/multica/` passes
- `go test ./internal/daemon/ -run "GC|Daemon"` passes
- Added `TestGCWorkspaceRootsMultiProfile` and `TestProfileDaemonActive` covering: abandoned profile root is reclaimed, live daemon's root is skipped, never-used root is skipped.

Closes #6962

### DIFF
--- a/server/internal/daemon/gc.go
+++ b/server/internal/daemon/gc.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/multica-ai/multica/server/internal/cli"
 	"github.com/multica-ai/multica/server/internal/daemon/execenv"
 	"github.com/multica-ai/multica/server/internal/daemon/processtree"
 	"github.com/multica-ai/multica/server/internal/daemon/repocache"
@@ -82,38 +83,20 @@ type gcStats struct {
 
 // runGC performs a single GC scan across all workspace directories.
 func (d *Daemon) runGC(ctx context.Context) {
-	root := d.cfg.WorkspacesRoot
-	entries, err := os.ReadDir(root)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return
-		}
-		d.logger.Warn("gc: read workspaces root failed", "error", err)
-		return
-	}
-
 	stats := &gcStats{byPattern: map[string]int{}}
-	for _, wsEntry := range entries {
-		// Skip every daemon-internal dot directory, not just .repos. A
-		// workspace directory is always a UUID, so a dot-prefixed entry is one
-		// of our own caches. Walking .skill-cache as if it were a workspace
-		// made its `v1` directory look like a task dir with no .gc_meta.json,
-		// so the orphan path would delete the entire bundle cache once its
-		// mtime went 72h without a new bundle. That reclaimed a few hundred KB
-		// and cost a full re-download.
-		if !wsEntry.IsDir() || strings.HasPrefix(wsEntry.Name(), ".") {
-			continue
-		}
-		wsDir := filepath.Join(root, wsEntry.Name())
-		d.gcWorkspace(ctx, wsDir, stats)
+	for _, gr := range d.gcWorkspaceRoots() {
+		d.gcRoot(ctx, gr.root, stats)
 	}
 
 	// Stable-root records are published before the physical env root so a
 	// re-dispatch cannot choose a different readable path. If preparation never
 	// reaches ClaimEnvRoot, no task directory exists for the normal GC walk to
 	// reclaim. Bound those records separately, but only after the orphan grace
-	// period and an authoritative terminal/not-found task check.
-	rootRecordsRemoved, rootRecordsErr := execenv.PruneTaskRootIndex(root, d.cfg.GCOrphanTTL, time.Now(), func(_ string, taskID string) bool {
+	// period and an authoritative terminal/not-found task check. This stays on
+	// the current profile's root: root-index records are the current daemon's
+	// own dispatch bookkeeping, and reclaimed profile roots are handled by the
+	// per-root gcRoot walk above instead.
+	rootRecordsRemoved, rootRecordsErr := execenv.PruneTaskRootIndex(d.cfg.WorkspacesRoot, d.cfg.GCOrphanTTL, time.Now(), func(_ string, taskID string) bool {
 		if ctx.Err() != nil || d.client == nil {
 			return false
 		}
@@ -127,12 +110,6 @@ func (d *Daemon) runGC(ctx context.Context) {
 		d.logger.Warn("gc: prune task root index failed", "error", rootRecordsErr)
 	}
 	stats.taskRootIndexEntriesReclaimed += rootRecordsRemoved
-
-	// Prune stale worktree references from all bare repo caches, then evict the
-	// caches nothing needs anymore. These live outside any workspace directory
-	// and are never reclaimed by the task walk above.
-	d.pruneRepoWorktreesContext(ctx, root, stats)
-
 	// Reclaim per-issue Codex session stores idle past their TTL. These live
 	// under the shared ~/.codex home (outside WorkspacesRoot) so resume survives
 	// the task GC, which means they need their own bounded lifecycle (MUL-4424).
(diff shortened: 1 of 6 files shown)


sanjay3290:
Drive-by review while looking at #6962 — this approach reads well, and keeping it inside the daemon package avoids a `Config` field and CLI wiring. One correctness gap worth closing before it merges.

`profileDaemonActive` treats the *presence* of `daemon.pid` as "running":

```go
_, err = os.Stat(filepath.Join(profileDir, "daemon.pid"))
return err == nil
```

The doc comment says "a live daemon removes that file on exit", but the removal is a `defer` in `runDaemonForeground`:

```go
// server/cmd/multica/cmd_daemon.go:1074-1076
os.WriteFile(daemonPIDPathForProfile(profile), []byte(strconv.Itoa(os.Getpid())), 0o644)
...
defer os.Remove(daemonPIDPathForProfile(profile))
```

A `defer` does not run on `SIGKILL`, on an OOM kill, or on a host power loss. So a daemon that crashed leaves its `daemon.pid` behind permanently, and `profileDaemonActive` reports it as alive forever.

That is the case the issue is about. A profile abandoned by a clean profile switch does get collected, but a profile abandoned by a crash — the one most likely to have left a large workspace root behind — is skipped on every future GC pass. The stale-pid-file caveat in the comment is accurate but understates it: the skip is not transient, it never expires.

Cheap fix that keeps your fail-safe direction: read the pid and signal it, rather than stat the file.

```go
b, err := os.ReadFile(filepath.Join(profileDir, "daemon.pid"))
if err != nil {
    return false // no pid file: not running
}
pid, err := strconv.Atoi(strings.TrimSpace(string(b)))
if err != nil || pid <= 0 {
    return true // unreadable: fail safe to "alive"
}
return syscall.Kill(pid, 0) == nil
```

`kill(pid, 0)` returns `ESRCH` for a dead pid, so a crashed profile's root becomes collectable again. It needs a `_unix`/`_windows` split, since `syscall.Kill` does not exist on Windows — returning a constant `true` there keeps the sweep inert on the platform, which is conservative but safe.

Two smaller notes:

- PID reuse is a real but narrow risk: the OS could have handed that pid to an unrelated process, which would make a genuinely abandoned root look alive. That errs toward not deleting, so it is the safe direction, but it is worth a line in the comment so the next reader does not assume the check is exact.
- Worth stating explicitly in `gcWorkspaceRoots` why a live daemon's root is skipped even when *this* daemon can see it on disk. The in-process `isActiveEnvRoot` reasoning in your comment is the important part and it is easy to lose in a later refactor.

Happy to be wrong on any of this if the pid file is cleaned up somewhere I did not find — I only traced the write/remove pair at `cmd_daemon.go:1074` and `:1076` and the removal at `:1317`.

Kickflip73:
Thanks for the careful review — you were right that pid-file presence alone leaves crash-abandoned roots stranded. Addressed in f9168e0b, rebased onto the current `main`.

Changes:
- `profileDaemonActive` now reads and validates the PID; missing means inactive, while unreadable/malformed values fail safe to active.
- Unix probes with `kill(pid, 0)` and treats only `ESRCH` as dead; permission/other errors stay fail-safe.
- Windows uses `OpenProcess(SYNCHRONIZE)` plus a zero-timeout wait, so stale PID files are collectable on the platform from the original report instead of making the sweep permanently inert.
- Comments now document the cross-daemon `isActiveEnvRoot` boundary and PID-reuse false-positive behavior.
- Tests cover live, whitespace, stale, malformed, and non-positive PIDs, plus an end-to-end GC pass that reclaims an old task behind a stale PID file.

Local validation passed:
- `go test ./internal/daemon/... -count=1`
- `go vet ./internal/daemon/...`
- `go build ./...`
- Windows amd64 cross-compile of `internal/daemon`

Would you mind taking another look?

Kickflip73:
@Bohan-J — this fixes the workspace-root GC gap you filed in #6962. CI is fully green, and I've addressed @sanjay3290's review: pid-file liveness now probes the real process (`kill(pid, 0)` on Unix, `OpenProcess(SYNCHRONIZE)` on Windows), so roots abandoned by a crash are collectable. Would appreciate a look when you have a chance.

Bohan-J (changes requested):
Thanks for this, and for the thorough turnaround on the first round of feedback — moving from "pid file exists" to an actual process probe, with the `_unix`/`_windows` split and the fail-safe-to-alive direction, was the right correction and it's well executed.

I'm requesting changes on two blocking items. The first is a safety gap; the second is that the main scenario from #6962 isn't actually covered yet.

---

## 1. Blocking — the cross-root sweep isn't protected across processes

Every guard the GC relies on today is **in-process**, and that was sound precisely because the GC only ever walked its own root. This PR widens the scope to other daemons' roots but keeps the same guards, so none of them apply to the thing they now need to protect against:

- `isActiveEnvRoot` is a plain in-memory map (`server/internal/daemon/daemon.go:9262`) — your comment in `gcWorkspaceRoots` correctly identifies this.
- The task-directory removal path never acquires `.task_lock`; `applyGCAction` / `cleanTaskDir` validate the owner marker but take no lock that another process can observe.
- `pruneRepoWorktreesContext` now runs on foreign roots too, and `withRepoLock` bottoms out in `c.repoLocks`, a `sync.Map` of in-process locks (`server/internal/daemon/repocache/cache.go:308`). A live daemon mid-fetch or mid-`worktree add` on that bare repo is not serialised against this eviction at all.

The pid probe is doing all the work of keeping these safe, and it can't: it answers a question about a *process* at one instant, not about a *directory* over the span of the operation.

Concretely, this still loses data after the current fix. Reproduced with two daemon states and a real `.task_lock` (simulated daemon state, not two live daemons): the old profile is not running when `gcWorkspaceRoots()` enumerates; it starts and re-claims a task directory while the pass is waiting on the batch issue-status round-trip; the GC then removes that directory even though the PID is now live and the task lock is held.

Worth being explicit, because it shaped the last round of suggestions: **re-checking liveness immediately before the delete does not close this.** The task can start in the window after the re-check. Narrowing a TOCTOU window isn't the same as removing it, and the failure here is a silently deleted workdir under a running task.

The codebase already has the right primitive, and already applies it for exactly this reason one screen up from your change. `runGC` at `server/internal/daemon/gc.go:145`:

> Unlike every other pruner here this one asks the kernel, not the clock: it removes a directory only once it has acquired the `.task_lock` the owner would still be holding.

and `claimEnvRoot` (`server/internal/daemon/execenv/execenv.go:1400`):

> The lock is an OS advisory lock, and the kernel releases it when the holding process exits for any reason. That is what makes it a lock rather than a third marker file: it answers "is the previous execution still alive?" across processes, with no heartbeat, no PID table and no stale-state cleanup path. **An identity marker alone cannot answer it** […]

That last line is the crux. A pid file plus a liveness probe is a marker-and-PID-table design — the shape that comment was written to rule out, after it let a re-dispatched execution wipe a running one (the #7326 follow-up). `PruneTaskTempDirs` is the worked example of the alternative already in-tree.

**What I'd like instead:** make task startup and GC share cross-process mutual exclusion that spans check → delete, rather than inferring safety from a pid. Taking `.task_lock` non-blocking per task directory before removal gets you this for free, kernel-released on crash, with no liveness inference at all — and it subsumes the abandoned-profile case, since a dead daemon's lock is already released. The same needs to cover repo-cache eviction on foreign roots. A missing pid file should not by itself be treated as proof that deletion is safe.

## 2. Blocking — the roots that actually need reclaiming aren't all enumerated

Three gaps in `gcWorkspaceRoots()` (`server/internal/daemon/gc.go:185`), all reproduced on `86b4e35`:

**(a) The default profile's root is never collected — this is the scenario in #6962.** `ProfileDir("")` is `~/.multica`, while `ProfileDir("x")` is `~/.multica/profiles/x`, so iterating `profiles/` only ever yields *named* profiles. With the current profile set to a named one and the default profile long retired, `~/multica_workspaces` is not in the returned set. Only named → named switches get any benefit; default → named, the common case and the one in the issue, is unchanged.

**(b) A profile's own `workspaces_root` is ignored.** The code calls `daemon.ResolveWorkspacesRoot(profile, "")`, which doesn't read that profile's `config.json`. A profile that put its workspaces on another volume has the real directory skipped, while a `$HOME`-derived path it never used gets scanned instead.

**(c) No de-duplication.** `MULTICA_WORKSPACES_ROOT` is profile-independent, so when it's set every profile resolves to the same directory and it's walked once per profile — including a full `dirSize` tree walk and `pruneRepoWorktreesContext` each time.

All three are already solved by `enumerateDiskUsageRoots()` (`server/cmd/multica/cmd_daemon.go:2000`), which puts the default root in first, resolves via `resolveWorkspacesRootForProfile` (so `config.json` is honoured), and de-dupes with `samePath`. Its resolver even carries the rationale:

> Keeping daemon start, disk-usage, aggregate scans, and cross-profile hints on this path prevents diagnostics from drifting away from the directory the daemon actually uses.

Reclamation should ride the same enumeration rather than a second implementation that drifts from it. Lifting it somewhere both can share, then layering the safety filter on top, fixes (a)–(c) at once and keeps the two sides from diverging again.

## 3. Tests currently sit on the blind side of (2a)

All three new cases construct `Config` with `Profile` left at `""` (`gc_multiroot_test.go:78`, `:110`, `:185`), i.e. the current profile *is* the default profile — so the default root arrives via `d.cfg.WorkspacesRoot` and the enumeration gap can't show up. No case sets a named current profile.

As written these lock in the named→named-only behaviour. A case with `Config{Profile: "team-a", ...}` and an abandoned default root would fail today; please add it alongside the fix, plus coverage for (b) and (c) and for a concurrent start racing the sweep.

## Smaller notes

- `gcRoot` is both a struct (`gc.go:174`) and a method (`gc.go:250`). Compiles, but reads badly — maybe `gcTarget` for the struct.
- Reclaimed foreign roots keep their `.task_root_index` records, since `PruneTaskRootIndex` stays on the current root. Keeping it scoped is the right call; a brief comment saying the leftover records are deliberate would help.
- The sweep is currently silent on success and on failure. An `Info` line naming the profile when a foreign root is reclaimed would make it possible to tell afterwards whether this worked or mis-fired.

## What's already right

Worth keeping through the rework: the fail-safe-to-alive direction, the `processAlive` platform split (not misreading `EPERM` / access-denied as dead), keeping `PruneTaskRootIndex` on the current root while `pruneRepoWorktreesContext` follows each root, and threading `root` explicitly through `gcTaskDirOwner` / `cleanTaskDir` / `RemoveRootDirRecord` instead of reaching for `d.cfg.WorkspacesRoot`.

The underlying problem is real and still unfixed on `main`, and both blockers are implementation-level — this is worth finishing. Happy to review again once the locking approach and the shared enumeration are in.

Verified on `86b4e35`: the three new tests and the `execenv` `ByteIdentical` checks pass once the runtime's own environment variables are isolated; the reproductions above fail.

PR #7085, opened 2026-08-17
MUL-6211: feat(agents): expose usage in task history

### LINKED ISSUE #6998: [Feature]: Expose token usage and cost in `multica agent tasks --output json`
### Deployment type

self-host

### What do you want and why?

The web UI already shows per-task token usage and cost — the "执行日志 / Execution log" section on an issue renders something like `24.7M · $19.45`. That number is exactly what I need, but there is no way to read it programmatically.

I checked every CLI surface I could find, and none of them expose it:

| Command | token / cost field |
|---|---|
| `multica agent tasks <agent-id> --output json` | none — the 25 returned keys include `result`, which only carries `output`, `pr_url`, `session_id`, `work_dir` |
| `multica issue get <id> --output json` | none |
| `multica agent get <id> --output json` | none |
| `multica workspace get --output json` | none |

(The string `token` does appear in those payloads, but only inside free text — API credentials mentioned in comments, provider quota error messages — never as a numeric field.)

**Why I need it programmatically:** I want to run a daily check for abnormal spend across the workspace — something like "yesterday's total cost vs. the trailing 7-day median", "tasks that burned more than $X", "one issue that accumulated more than N runs". Today the only way to see cost is to open issues one at a time in the UI, which does not scale to a workspace with ~20 agents spread over 4 runtimes, and it cannot be turned into an alert.

**Why local session files are not a workaround:** For agents running on Claude Code runtimes I can parse `~/.claude/projects/<encoded-workdir>/<session-id>.jsonl` and sum the `usage` blocks. That fails in two ways:

1. It only covers runtimes on the machine I'm running from. In one recent night of 90 tasks, only 28 had session files I could read — the rest lived on other hosts.
2. It does not cover non-Claude runtimes at all. Codex-backed agents leave no equivalent local artifact I can total up.

The server clearly does not have either problem — it bills every task regardless of runtime or model, and the `$` figure in the UI already reflects weighted pricing including cache discounts. So the data exists and is authoritative; it's just not reachable from the CLI.

Add the usage numbers to the task payload returned by `multica agent tasks <agent-id> --output json`, e.g.:

```jsonc
{
  "id": "...",
  "status": "completed",
  "result": { "output": "...", "session_id": "..." },
  "usage": {
    "input_tokens": 1726,
    "output_tokens": 893122,
    "cache_read_input_tokens": 85982211,
    "cache_creation_input_tokens": 3751125,
    "total_tokens": 24700000,
    "cost_usd": 19.45
  }
}
```

Field names/shape are just a sketch — whatever the UI already consumes would be ideal, so the two surfaces cannot drift.

Two smaller things that would help, in rough priority order after the above:

1. Include the same block in `multica issue get`, aggregated over the issue's tasks — that's the granularity the UI shows today.
2. A workspace-scoped, time-windowed query (something like `multica usage --since <RFC3339> --until <RFC3339>`), so a monitoring job doesn't have to fan out one request per agent and filter client-side. Not required if #1 lands — the fan-out is workable at ~20 agents — but it would be the natural home for this.

The UI element in question is the collapsed "执行日志" (Execution log) header on an issue detail page, which renders the token total and dollar cost inline next to the toggle — `24.7M · $19.45`. I couldn't attach the screenshot through the CLI, happy to add it if useful.

iiwish:
## Summary

- Expose the existing per-(provider, model) usage rows through `multica agent tasks <agent-id> --output json`.
- Make usage hydration explicit with `include_usage=true`; the normal agent history API and human-readable CLI table keep their existing lightweight behavior.
- Load usage with one agent-scoped bulk query for the returned task IDs, avoiding N+1 queries and preserving the authorization boundary.
- Preserve nullable authoritative `cost_usd_ticks` and omit `usage` when no accounting was recorded.
- Return a clear error when an explicit accounting request cannot load usage instead of reporting a misleading empty result.
- Parse the Core task-history response through `AgentTaskListSchema` without changing its request or UI behavior.
- Use the shared `dbfx` fixture layer for database-backed coverage.

## Scope

This PR is limited to the concrete CLI gap in #6998. It does not add pagination, time filters, indexes, issue aggregation, or a workspace usage API.

## Testing

```bash
cd server
go test ./cmd/multica ./internal/handler ./pkg/db/generated

cd ..
pnpm --filter @multica/core exec vitest run api/client.test.ts
pnpm --filter @multica/core typecheck
pnpm --filter @multica/core exec eslint api/client.ts api/client.test.ts
```

Closes #6998


iiwish:
Hi @Bohan-J, a gentle follow-up on this PR when you have a chance.

This addresses #6998's concrete gap: the UI already has authoritative per-task usage and cost, but `multica agent tasks --output json` cannot expose it for monitoring. The change is additive, keeps the human-readable table unchanged, uses one agent-scoped query rather than N+1 lookups, and the full CI suite is green.

The PR intentionally keeps the first step focused on per-task agent history; workspace/time-window aggregation can follow separately. No rush, but I would appreciate a review when convenient, and I am happy to adjust anything that does not fit the current API direction.

iiwish:
Hi @Bohan-J and @multica-eve, a gentle follow-up on this PR when you have a chance.

This adds per-task usage and cost to `multica agent tasks --output json` while keeping the human-readable table unchanged. The change is additive, CI is green, and the first step is intentionally scoped to per-task agent history.

When convenient, could you please take a look and let me know whether the current API direction fits? I am happy to adjust anything that does not align with the project's direction. Thanks!

multica-eve (approved):
Thank you, @iiwish, for the thoughtful implementation and for continuing to iterate on it. The opt-in usage hydration keeps the normal UI path lightweight while exposing the existing per-provider/model accounting data to JSON consumers.

I pushed a small documentation clarification to align the type comment with the final behavior. This is ready to merge.

PR #7087, opened 2026-08-17
MUL-6264: feat(chat): show agent names in session list

### LINKED ISSUE #7048: [Feature]: Show the agent name in the web chat session list
### Deployment type

- [x] self-host

### What do you want and why?

In workspaces with multiple agents, the web chat session list shows only the agent's **avatar** beside each session, not the agent's **name**. When several agents share similar avatars or a session's title does not hint at its agent, it is hard to tell which conversation belongs to which agent at a glance — especially in workspaces with many pinned or recent chats.

A Multica mobile client already renders the agent's name next to each session in its chat list. Bringing the same affordance to the web client would keep the two surfaces consistent and make multi-agent workspaces easier to navigate.

### Proposed solution

In each row of the web chat session list (the session/thread list), show the agent's name in addition to the avatar:

- Display the name as a secondary line under the session title (or inline beside the title, matching the existing row layout), so each conversation is identifiable by both title and agent.
- Keep the current fallback for sessions without an `agent_id` (the row already handles that case).
- No change to the conversation header, which already shows the agent name and description.

Happy to attach side-by-side mockups of the mobile client's session list and the current web list if useful.

iiwish:
## Summary

- show the resolved agent name before the existing status or message preview in each web chat-session row
- reuse the agent data already passed to the list, with no API or schema change
- preserve the current preview and fallback when an agent is missing or has a blank name
- bound long agent names to 40% of the secondary row so previews, unread badges, and actions keep their space

## Testing

- `pnpm -C packages/views exec eslint chat/components/chat-thread-list.tsx chat/components/chat-thread-list.test.tsx`
- `pnpm -C packages/views exec vitest run chat/components/chat-thread-list.test.tsx` (12 tests)
- `pnpm --filter @multica/views test` (4,286 tests)

Closes #7048


Bohan-J:
Please upload a screenshot containing before/after changes so that I could have a quick look on it.

iiwish:
Thanks for the request, @Bohan-J. I’ve uploaded before/after screenshots using the same session data.\n\n**Before**\n\nBefore: chat session list\n\n**After**\n\nAfter: chat session list\n\nThe after state shows the resolved agent name before the existing status/message preview, while keeping the preview, unread badge, and actions in place.

Bohan-J:
Thanks for this, @iiwish — nicely scoped work. Reusing the agent map the list already receives (no API or schema change), keeping the preview fallback when an agent is missing or its name is blank, and bounding long names to 40% of the secondary row are exactly the right calls. The added tests cover the long-name, preview-preserved, and missing-agent cases. The before/after screenshots made the review easy — appreciated.

One small thing I took care of directly rather than sending it back to you: I dropped `docs/review-screenshots/pr-7087-before.png` and `-after.png` from the branch. They were one-off review artifacts referenced by nothing in the repo or the PR body, so they shouldn't ship to `main`. The screenshots had already done their job here in the PR discussion. Your implementation commit is untouched.

All 14 CI checks are green on the updated head. Merging now — thanks again for the contribution!

PR #7094, opened 2026-08-17
fix(runtime): support per-machine local profile activation

gavinxgu:
## What does this PR do?

Adds an explicit per-machine activation mode for custom runtime profiles. Workspace activation keeps the existing fan-out behavior; local activation requires a machine-owned `set-path` binding, so daemons without that binding skip the profile instead of publishing failed runtimes.

The change also fixes profile retirement by counting only user agents as deletion blockers, allows deletion of one local-profile runtime instance without deleting the shared profile, reloads path overrides on profile-change events, and makes the CLI preflight server capabilities before a local-mode mutation.

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `activation_mode=workspace|local` to runtime profiles and expose supported modes in the list response.
- Add CLI create/update support with capability preflight.
- Merge default and named-profile path overrides and hot-reload them on profile changes.
- Skip local profiles on daemons without an executable binding.
- Permit deleting an individual local-profile runtime while retaining the shared profile.
- Ignore hidden system agents when deciding whether a profile can be retired.
- Split migrations 342-344 around a concurrently built partial unique index and register retry cleanup hooks.
- Document local activation in the bundled runtime/repository skill.

## How to Test

1. `go test ./cmd/migrate ./cmd/multica ./internal/daemon ./internal/handler`
2. `go test ./internal/daemon/repocache -count=1`
3. Apply migrations 001-344 to a fresh PostgreSQL 17 database and verify the local activation column and enabled-only unique index.
4. Apply the 344, 343, and 342 down migrations and verify the legacy uniqueness constraint is restored.
5. Run two daemons in one workspace: only the daemon with `set-path` should publish the local profile.

## Checklist

- [ ] If this PR touches Chinese product copy, I checked it against the conventions

Risk is compatibility with older servers. The CLI fails closed before creating or updating a local profile when the server does not advertise local activation support. Existing profiles default to workspace activation.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:**
Implemented from a concrete multi-device runtime fan-out and uninstall failure. The approach was to keep server profile identity semantic and shared while moving executable binding to explicit per-machine state, then verify daemon convergence, deletion guards, migration retry behavior, and fresh-database forward/rollback paths.


prgrmrwy:
+1 — this PR directly addresses a problem we're hitting every day in a shared multi-device workspace

We run one workspace with about a dozen daemons (dev machines + teammates' machines), but a custom runtime profile that is only actually installed on **one** machine, pinned via `set-path`. Here is what we consistently see:

1. **The profile appears on every device as a dead entry.** In the create-agent runtime picker and the Runtimes list, the same custom profile shows up on all ~12 devices. 11 of them are permanently offline; only the machine with the `set-path` binding has an online instance.

2. **The placeholder rows can't be cleaned up.** We measured a single profile producing 12 rows: 11 are `command not found on PATH` placeholders, 1 is the real online instance. Deleting the placeholders does not stick — the daemon re-registers and the server recreates them on the next registration cycle (they come back tagged as failed profiles). So the UI gradually fills with junk rows we are told not to manually remove.

3. **It actively misleads users.** Team members creating agents see a wall of "available" runtimes and can pick one that is offline on their own machine, which leads to confusing failures ("agent started but runtime never came online") that we have to explain over and over.

4. **It scales linearly with device count.** With N daemons, every custom profile generates N−1 junk rows. We have several profiles, and the device fleet keeps growing, so this is getting worse, not better.

Your `activation_mode=workspace|local` approach is exactly the right shape: local mode + skip when there is no machine-owned `set-path` binding + allowing deletion of a single local-profile instance would eliminate all of the above, while keeping workspace fan-out for the users who actually want it.

cc maintainers — would love a review here. Happy to test once the conflicts are resolved; we have a real multi-device workspace ready to verify the "no placeholder rows on non-pinned devices" behavior.

PR #7096, opened 2026-08-17
FLU-65: retry interrupted CLI requests safely

SeimoDev:
## Summary

- add a shared retrying `RoundTripper` under every CLI `APIClient` call
- retry retryable transport failures plus 502/503/504 for idempotent methods
- never replay POST/PATCH after a request may have reached the server; only pre-dial failures are safe for every method
- classify reset/EOF/GOAWAY/connection-desync failures as interrupted connections instead of reporting the server as offline
- document retry behavior and the `MULTICA_HTTP_RETRIES` escape hatch

## Why

FLU-65 traced repeated “Could not reach the Multica server” reports to lossy runtime-to-server links and an overly broad network-error fallback. A single transient interruption currently fails the whole CLI process even when the server is healthy.

## Verification

- `go test -count=1 ./internal/cli ./cmd/multica`
- `go test -race -count=1 ./internal/cli`
- `go vet ./internal/cli ./cmd/multica`
- `go build -buildvcs=false -o /tmp/multica ./cmd/multica`
- `gofmt -d` on all changed Go files

All checks ran with Go 1.26.6. This PR does not create a release; rollout should happen from a patch tag after merge.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -896,10 +896,10 @@ layer.) The underlying detail is still available on demand (see `--debug`).
 ### What you see
 
 - **Friendly, single-line message.** Transport failures (timeout, DNS,
-  connection refused, TLS) and HTTP status failures (401/403/404/409/400·422/
-  429/5xx) are each rendered as one clear sentence with a next step — for
-  example a timeout suggests checking the network or raising
-  `MULTICA_HTTP_TIMEOUT`, and a 401 tells you to run `multica login`.
+  connection refused, TLS, interrupted connections) and HTTP status failures
+  (401/403/404/409/400·422/429/5xx) are each rendered as one clear sentence
+  with a next step — for example a timeout suggests checking the network or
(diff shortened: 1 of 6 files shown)


multica-eve:
Thanks for the thoughtful contribution. Today, when a user-triggered operation hits a brief network interruption, showing the failure and asking the user to retry is acceptable behavior. We therefore do not want to retry these operations automatically by default without stronger evidence that this is a recurring user problem.

The error-classification improvement is valuable, though. Distinguishing an interrupted connection from an actually unreachable server gives users a more accurate message and clearer next steps.

There is also a mismatch in the retry motivation: the PR cites a 70-second cold start and roughly 1,000 gateway errors, while the proposed default schedule waits only 200 ms and 600 ms. That schedule may absorb a very brief interruption, but it cannot address the cold-start scenario.

Would you be willing to narrow this PR to the connection-error classification, user-facing messages, tests, and related documentation, and remove the default retry transport? We can revisit automatic retries separately if unattended CLI usage becomes an explicit product requirement and we have data on failure frequency and the expected recovery window.

Thanks again for taking the time to investigate and improve this area.

PR #7106, opened 2026-08-17
MUL-6316: WOR-366 / WOR-431: Cursor transport retry + WritableIterable classify fallback

Arianxx:
## Summary

- **WOR-366**: Add `internal/daemon/transportretry` plugin wrapping `executeAndDrain` for in-turn transport-layer retries (Cursor WritableIterable / connection-closed patterns).
- **WOR-431**: Classify `writableiterable is closed` as `agent_error.provider_network` (Classify rule 7 + `NormalizeDaemonReason`) so failures that surface after in-turn retry exhaustion still get platform auto-retry + session resume instead of terminal `process_failure`.

In-turn retry prefers same-session resume before fresh-session recovery; hard `tools == 0` gate on both layers.

## Test plan

- [x] `go test ./internal/daemon/transportretry/...`
- [x] `go test ./internal/daemon/ -run 'TestExecuteWithTransportRetry|TestExecuteWithRetry'`
- [x] `go test ./pkg/taskfailure/...`

Closes WOR-431


Arianxx:
## Status update

- CI: all checks green
- Local tests: transportretry + taskfailure PASS
- Built binary ready at `server/bin/multica` (commit `0dabd0c56`)

**Blockers before production effect:**
1. Merge (needs maintainer — author lacks `MergePullRequest` on upstream)
2. Local deploy: replace Homebrew Cellar binary + `multica daemon restart` (agent tasks cannot write `/opt/homebrew/Cellar/...`)

Until deploy, WritableIterable failures still surface as `process_failure` / `blocked` (last seen 12:07 local).

Bohan-J (changes requested):
Thanks for this — the product direction is right. A brief transport hiccup shouldn't force a member to manually re-run a task, and preferring same-session resume before falling back to a fresh session is the correct shape. My concerns are all about invariants that cross layer boundaries, and I think three of them are merge blockers.

## Must-fix

**1. The result adapter drops billing and control fields for every task, not just retried ones.**

`server/internal/daemon/transport_retry.go:79` projects `agent.Result` into a partial view, and the reconstruction at `server/internal/daemon/transport_retry.go:97` only carries `CacheReadTokens` back — `InputTokens`, `OutputTokens`, `CacheWriteTokens`, `CostUSDTicks`, `DurationMs` and `ResumeRejected` are all zeroed.

This wrapper now runs unconditionally for every provider and every task, so usage is under-reported even when no retry is triggered at all. Losing `ResumeRejected` also breaks the existing fresh-session resume decision downstream. And when a retry *does* happen, usage from earlier attempts isn't merged the way `reconcileFreshRetryResult` already does.

Please make the no-retry path field-for-field equivalent to today's result, and merge full usage across all attempts. Worth adding a field-fidelity test plus a multi-attempt billing test.

**2. "No auto-retry after a tool has run" isn't an end-to-end guarantee.**

`server/internal/daemon/transportretry/executor.go:103` stops retrying within the current round, but the same error still classifies as `agent_error.provider_network` at `server/pkg/taskfailure/classify.go:199` and `server/pkg/taskfailure/classify.go:424`. The server then creates an automatic retry task off that classification, and that path has no tool-usage gate.

So the "tool already executed → `WritableIterable is closed`" case can still re-run, which contradicts both the stated side-effect protection and the PR description. Either keep a result that matched a transport error *and* observed tool use non-auto-retryable, or propagate reliable tool-execution evidence to the server and check it before creating the retry task. A regression test spanning daemon → service would be valuable here.

**3. The public retry-limit config can't actually bound the default policies, and the wall-clock budget isn't enforced.**

`server/internal/daemon/transportretry/policy.go:198` ignores `MaxExtraAttempts` entirely whenever `SessionStrategy` is non-empty. Both default policies carry a strategy, so the `max_extra_attempts` override is inert for them. The Cursor default says `MaxExtraAttempts: 2` but in practice permits 3 extra starts — and the test pins the total at 4, which encodes the discrepancy rather than catching it.

Separately, `server/internal/daemon/transportretry/executor.go:207` returns `ctx.Err()` on a zero delay past the deadline; when the context hasn't been cancelled that's `nil`, so retries just continue. And when the delay exceeds the remaining budget, the wait doesn't observe the wall-clock deadline either.

I'd suggest a single attempt budget that's both authoritative and validated at config load, and a real `context.WithDeadline` covering both the waits and the execution, with tests for the cap and for cancellation.

## What looks good

The policy split, same-session-first ordering, GC receipt, and observability hooks are all clean. Existing tests already cover resume, exhaustion, the tool gate, and receipt persistence — the gaps above are specifically the cross-layer invariants those tests don't reach.

## Verification I ran

- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1`
- `go test ./internal/daemon/transportretry/... -count=1`
- `go test ./internal/daemon/ -run 'TestMarshalTransportRetryReceipt|TestHandleTask_PersistsTransportRetryReceiptToGCMeta' -count=1`
- `go test ./pkg/taskfailure/... -count=1`

All pass, and CI is green — but none of the scenarios above are covered by the current suite. I did not run a smoke test against a real Cursor account.

Arianxx:
Closing: this workspace merges to fork Arianxx/multica, not upstream. Reopened at https://github.com/Arianxx/multica (fork PR to follow).

PR #7111, opened 2026-08-18
fix(agent): preserve Pi RPC stdin on Windows

### LINKED ISSUE #7108: [Bug]: Model discovery in the Windows Pi runtime is missing thinking level metadata
### Deployment type

self-host

### What happened?

问题
在 Windows 本地运行的 Pi runtime 中创建 agent 时，模型列表可以正常加载，但无法选择 thinking level。模型接口返回的模型只有 id、label 和 provider，没有 thinking 元数据，因此前端不会显示 thinking level 选择器。

环境
系统：Windows
Multica daemon：0.4.28
Pi：0.84.2

初步定位
问题疑似出在 Multica 的 Windows 启动适配，而不是 Pi 配置：Pi 的模型发现使用双向 RPC，需要保持 stdin 打开并等待响应；如果通过 npm 的 pi.ps1 -File wrapper 启动，wrapper 可能等待 stdin EOF 后才转发输入，导致 RPC 超时，随后回退到不包含 thinking 元数据的普通模型列表。

修复建议
Windows Pi RPC 模式使用 PowerShell -Command 并通过 @args 转发参数，保持双向 stdin。
普通 JSON/task 执行继续使用现有启动方式，避免影响多行参数处理。
增加 Windows launcher 到 Pi RPC 的回归测试。
修复后刷新模型缓存并重新打开 agent 创建页面验证。

### Steps to reproduce

打开 agent 的手动创建页面。
选择一个 Windows 本地 Pi runtime。
选择支持 reasoning 的模型。
查看模型配置区域。
实际结果
模型列表请求返回成功，但模型条目缺少 thinking 配置：

{
  "status": "completed",
  "models": [
    {
      "id": "provider/reasoning-model",
      "label": "provider/reasoning-model",
      "provider": "provider"
    }
  ],
  "supported": true
}
页面不显示 thinking level，已保存的 thinking_level 也可能在 daemon 校验时被判定为当前模型不支持。

期望结果
Pi RPC 返回的 reasoning 模型应包含 thinking 元数据，例如：

{
  "id": "provider/reasoning-model",
  "thinking": {
    "supported_levels": [
      {"value": "high", "label": "High"},
      {"value": "max", "label": "Max"}
    ],
    "default_level": "high"
  }
}
前端应据此显示 thinking level 选择器，并允许创建或更新 agent。

### Screenshots (optional)

_No response_

```shell

```

itsXuSt:
Fixes #7108

## Root Cause

Windows Pi model discovery reused the task execution launcher rewrite from `pi.cmd` to `powershell -File pi.ps1`. That path is appropriate for normal JSON/task execution, where Multica writes the prompt to stdin and then closes it. Pi RPC model discovery is different: it needs a bidirectional stdin/stdout session and sends requests while keeping stdin open. Some npm PowerShell wrappers can wait for stdin EOF before forwarding input, causing RPC discovery to time out and fall back to the legacy model table, which has no thinking metadata.

## Changes

- Route Windows Pi RPC invocations through `powershell -Command "& '<pi.ps1>' @args"` so the wrapper receives forwarded args while preserving live stdin.
- Keep ordinary Pi task execution on the existing `powershell -File` path to avoid changing multiline argument handling.
- Add a Windows launcher regression test that pins the RPC-specific `-Command` argv and rejects `-File` for model discovery.

## Validation

- `git diff --check -- server/pkg/agent/pi_invocation_windows.go server/pkg/agent/pi_invocation_windows_test.go` passed.
- Go tests could not be run in this environment because `go` and `gofmt` are not installed in PATH.


Bohan-J (changes requested):
Thanks for the fix — the direction is right and I want this behaviour in. One blocking item before merge.

**Product:** accept. Windows members on a Pi runtime currently can't see or persist a thinking level for reasoning-capable models; this restores capability discovery without touching normal task execution.

**Implementation:** the split is well scoped. `server/pkg/agent/pi_invocation_windows.go:15` switches the launcher only for RPC discovery, ordinary task execution stays on the `-File` path, and the script path is single-quote escaped. On PowerShell 7.5.2 the `-Command "& '<script>' @args"` form does forward all six current discovery arguments intact.

## Must-fix: the regression test doesn't cover the EOF behaviour this fix depends on

`server/pkg/agent/pi_invocation_windows_test.go:70` writes an empty fake `powershell.exe` and only compares the argv that `platformPiInvocation` assembles. It never starts PowerShell, an npm wrapper, or a bidirectional session. So if `-Command` at `server/pkg/agent/pi_invocation_windows.go:52` still fails to deliver a response before stdin EOF under Windows PowerShell 5.1, this test — and the `windows-execenv` CI job — would keep passing anyway.

The dependency chain the PR claims to fix:

```text
old: Go writes both requests and keeps stdin open
   -> powershell -File / npm wrapper waits for stdin EOF
   -> Pi doesn't receive the requests in time
   -> Go blocks waiting on stdout
   -> 7s timeout, falls back to the model list without thinking metadata

new: Go writes both requests and keeps stdin open
   -> powershell -Command invokes the wrapper directly, Pi inherits stdin/stdout
   -> Pi answers before EOF
   -> Go closes stdin once both responses arrive
   -> Pi then PowerShell exit, cmd.Wait returns
```

If "answers before EOF" doesn't hold on a given host, the timeout context still tears the process down but the user gets the same degraded list — so this is a correctness gate, not a testing preference.

Suggestion: reuse the real PowerShell harness in `server/pkg/agent/pi_stdin_windows_test.go`. For each host returned by `availablePowerShellHosts()`, start a fake npm wrapper, write the RPC requests while keeping stdin open, assert the model responses arrive first, then close stdin and assert the process exits. Asserting through `discoverPiModelsRPC` on the reasoning/thinking metadata would be ideal — that way the test fails against the old implementation and only passes with the new one. Covering Windows PowerShell 5.1 matters most here, since that's the host the report points at.

Everything else looks good — happy to re-review as soon as that coverage lands.

itsXuSt:
Thanks for the review. I added the requested end-to-end Windows coverage using the existing PowerShell harness.

The new test starts the fake npm wrapper through each available PowerShell host, exercises `discoverPiModelsRPC`, and asserts that the RPC responses arrive before stdin EOF by checking the discovered reasoning model and its `max` thinking metadata. The fake `-File` entrypoint deliberately waits for stdin EOF before forwarding, so the old launcher path would time out and fail to produce the RPC catalog.

I also synced the fork branch with current upstream `main` so the PR diff is limited to the Pi launcher and test files. Local validation remains limited to `git diff --check` in this environment because Go is not installed here; the Windows CI job should cover the new build-tag test.

multica-eve (changes requested):
Thanks for adding the end-to-end Windows RPC coverage. The new test now exercises the right production boundary through `discoverPiModelsRPC`, a real PowerShell host, the shim, and a native child, including the response-before-EOF behavior and thinking metadata.

## Must-fix: the new Windows regression test is not selected by CI

The `-run` expression in `.github/workflows/ci.yml:453` includes the existing launcher tests, but it does not match `TestPiRPCDiscoveryAnswersBeforeStdinEOF` from `server/pkg/agent/pi_stdin_windows_test.go:108`.

I checked the latest `windows-execenv` log (run `32102316091`, job `95605086113`): it runs `TestPiExecutePromptSurvivesPowerShellShim` and the `TestPlatformPiInvocation...` tests, but the new RPC test never appears. Because the file is guarded by `//go:build windows`, the Linux backend job cannot execute it either. The green checks therefore still do not verify this regression on Windows PowerShell 5.1 or pwsh.

Please either add `TestPiRPCDiscoveryAnswersBeforeStdinEOF` to that `-run` expression or rename it under an already-selected prefix. After the next run, the Windows log should explicitly show both:

```text
TestPiRPCDiscoveryAnswersBeforeStdinEOF/powershell.exe
TestPiRPCDiscoveryAnswersBeforeStdinEOF/pwsh.exe
```

The implementation and the new test body otherwise look good; this remaining request is only to make sure the test actually runs in the environment it was written for.

itsXuSt:
Thanks for catching that. I updated the test name so it is selected by the existing Windows CI `-run` expression without changing the workflow file.

The RPC coverage now runs under the `TestPiExecutePromptSurvivesPowerShellShim...` prefix, alongside the existing PowerShell shim test. The next `windows-execenv` run should show subtests for both available hosts under `TestPiExecutePromptSurvivesPowerShellShimRPCDiscoveryBeforeStdinEOF`, while preserving the same end-to-end assertions through `discoverPiModelsRPC`, the fake npm wrapper, a real PowerShell host, and the native helper process.

Local validation remains limited to `git diff --check` because this environment does not have Go installed; the updated Windows CI run is pending now.

Bohan-J (approved):
Thanks for sticking with this through three rounds — both earlier requests are fully addressed, and the result is better than what I originally asked for.

The end-to-end test in `server/pkg/agent/pi_stdin_windows_test.go` now drives the real production boundary: a real PowerShell host, the npm shim, a native child, `discoverPiModelsRPC`, responses arriving while stdin is still open, and the reasoning/thinking metadata asserted. Renaming it under the existing `TestPiExecutePromptSurvivesPowerShellShim` prefix was the better call over widening the CI regex — the latest `windows-execenv` run confirms it actually executes:

```text
--- PASS: TestPiExecutePromptSurvivesPowerShellShimRPCDiscoveryBeforeStdinEOF (0.61s)
    --- PASS: TestPiExecutePromptSurvivesPowerShellShimRPCDiscoveryBeforeStdinEOF/powershell.exe (0.25s)
    --- PASS: TestPiExecutePromptSurvivesPowerShellShimRPCDiscoveryBeforeStdinEOF/pwsh.exe (0.36s)
```

Windows PowerShell 5.1 — the host the original report points at — is covered, not just pwsh.

The blast radius is exactly as narrow as it should be:

- only `server/pkg/agent/pi_invocation_windows.go` changes in production code, and it is behind `//go:build windows`, so non-Windows platforms keep the `pi_invocation_other.go` no-op.
- the shared `rewriteCmdToPS1` and `powerShellLookup` helpers are untouched, so Cursor and Copilot argv construction is byte-identical.
- the new `-Command` branch only fires when argv carries `--mode rpc` / `--mode=rpc`, which only `discoverPiModelsRPC` emits. Pi task execution builds `--mode json` and stays on the established `-File` launcher.
- `--mode` sits in Pi's blocked-with-value custom-option list, so no user-supplied option can flip a task run onto the RPC path.

Merging. Thanks again for the thorough fix and for the patience through the coverage rounds.

PR #7115, opened 2026-08-18
fix(agent tasks): sanitize failure diagnostics

### LINKED ISSUE #7098: [Bug]: NUL byte in task failure diagnostic returns 500 and leaves task running
### Deployment type

Self-hosted

### What happened?

When a daemon reports an agent-task failure whose diagnostic contains an embedded NUL (`\x00`), `POST /api/daemon/tasks/{taskId}/fail` returns HTTP 500. PostgreSQL rejects the diagnostic when it is written to a `TEXT` column:

```text
ERROR: invalid byte sequence for encoding "UTF8": 0x00 (SQLSTATE 22021)
```

The failure update is rolled back, so the task remains in `running` instead of transitioning to `failed`. Retrying the callback with the same payload fails again.

Expected behavior: the callback succeeds, the task reaches `failed`, and the useful diagnostic is preserved with only database-invalid bytes removed or replaced.

The current upstream path passes the diagnostic through unchanged:

- [`Client.FailTask`](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/server/internal/daemon/client.go#L455-L456) puts `errMsg` into the JSON request.
- [`Handler.FailTask` / `failTask`](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/server/internal/handler/daemon.go#L4192-L4225) decodes and forwards `req.Error` unchanged.
- [`TaskService.FailTask`](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/server/internal/service/task.go#L3964-L4035) constructs `pgtype.Text` directly from `errMsg`.
- [`FailAgentTask`](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/server/pkg/db/queries/agent.sql#L1055-L1084) writes it to `agent_task_queue.error`.

### Steps to reproduce

1. Start a daemon task and move it to `running`.
2. Build the failure callback with an embedded NUL and submit the JSON produced by the standard encoder:

```python
import json

payload = json.dumps({
    "error": "worker failed\x00 diagnostic details",
    "failure_reason": "agent_error",
})
```

3. Observe HTTP 500 and PostgreSQL `SQLSTATE 22021`.
4. Query `agent_task_queue`; the task is still `running` and `completed_at` is null.

### Suggested fix

Normalize the diagnostic at the service persistence boundary, before classification and the failure transaction, so every downstream use receives the same safe value.

Suggested policy:

- remove embedded NUL bytes, which PostgreSQL cannot store in `TEXT`;
- normalize any other invalid UTF-8 byte sequences;
- preserve all valid diagnostic text instead of dropping the whole error.

Please add a PostgreSQL-backed regression test proving that a NUL-bearing failure payload returns success and persists `status = 'failed'` with the remaining diagnostic intact.

### Additional context

JSON transport does not eliminate the value: standard JSON encoders represent the NUL with a Unicode escape, and decoding reconstructs the embedded NUL before persistence. The server should therefore validate or normalize it before writing to PostgreSQL.

Jamison929611:
## What does this PR do?

Normalizes agent task failure diagnostics at the start of TaskService.FailTask so PostgreSQL never receives an embedded NUL or malformed UTF-8. NUL bytes are removed, other invalid UTF-8 is replaced with U+FFFD, and the same normalized string is then used for failure classification, persistence, logs, comments, chat output, notifications, and events.

Thinking path: JSON decoding reconstructs escaped NUL bytes, the handler forwards the value unchanged, and FailTask previously classified and persisted that raw value. PostgreSQL rejects the TEXT write with SQLSTATE 22021, rolling back the terminal transition. Normalizing once at the service persistence boundary fixes that transaction while keeping all downstream behavior consistent.

Risk is limited to malformed diagnostics at this failure boundary. No handler, query, migration, or other database text field is changed.

## Related Issue

Closes #7098

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- server/internal/service/task.go: normalize failure diagnostics before classification and the failure transaction.
- server/internal/service/task_failure_diagnostic_test.go: cover valid UTF-8, embedded NUL, malformed UTF-8, NUL-only input, and the PostgreSQL-backed running-to-failed regression.

## How to Test

1. Run: go test ./internal/service -run '^(TestNormalizeTaskFailureDiagnostic|TestFailTaskSanitizesNULDiagnostic)$' -count=1
2. Run: go test ./internal/service -skip 'Skill' -count=1
3. Run: go vet ./internal/service

The PostgreSQL-backed test skips locally when DATABASE_URL is unavailable and runs against the migrated PostgreSQL service in CI. The unfiltered service suite has unrelated CRLF-sensitive embedded-skill failures on this Windows checkout; the filtered service suite and all scoped checks pass locally.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (apps/web/features/landing/i18n/) and **relevant docs** (apps/docs/content/docs/)
- [x] If this PR touches Chinese product copy, I checked it against apps/docs/content/docs/developers/conventions.zh.mdx (terminology, mixed-rule for task / issue / skill)

## AI Disclosure

**Prompt / approach:**
Used the tool to trace the existing callback/service/transaction data flow, drive RED-to-GREEN tests, implement the approved minimal service-boundary normalization, and review and verify the scoped diff.

## Screenshots (optional)

Not applicable; no UI changes.

### DIFF
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -3945,6 +3945,10 @@ func (s *TaskService) observeChatOutputLocalPath(task db.AgentTaskQueue, body st
 	)
 }
 
+func normalizeTaskFailureDiagnostic(errMsg string) string {
+	return strings.ReplaceAll(strings.ToValidUTF8(errMsg, "\uFFFD"), "\x00", "")
+}
+
 // FailTask marks a task as failed.
 // Issue status is NOT changed here — the agent manages it via the CLI.
 //
@@ -3962,6 +3966,8 @@ func (s *TaskService) observeChatOutputLocalPath(task db.AgentTaskQueue, body st
 // (via classifyPoisonedError, the timeout / runtime classifier, etc.)
 // will have their value preserved untouched.
 func (s *TaskService) FailTask(ctx context.Context, taskID pgtype.UUID, errMsg, sessionID, workDir, branchName, failureReason string, sessionRolloutMissing bool, retiredSessionID string) (*db.AgentTaskQueue, error) {
+	errMsg = normalizeTaskFailureDiagnostic(errMsg)
+
 	// MUL-2946: synthesise a refined reason from the error text whenever the
 	// caller didn't supply one. This is the last write-path guard against
 	// "agent_error" coarse rows ending up in agent_task_queue.failure_reason
--- a/server/internal/service/task_failure_diagnostic_test.go
+++ b/server/internal/service/task_failure_diagnostic_test.go
@@ -1,0 +1,97 @@
+package service
+
+import (
+	"context"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/multica-ai/multica/server/internal/events"
+	"github.com/multica-ai/multica/server/internal/util"
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+func TestNormalizeTaskFailureDiagnostic(t *testing.T) {
+	tests := []struct {
+		name       string
+		diagnostic string
+		want       string
+	}{
+		{
+			name:       "valid UTF-8 is unchanged",
+			diagnostic: "worker 失败：诊断详情",
+			want:       "worker 失败：诊断详情",
+		},
+		{
+			name:       "embedded NUL is removed",
+			diagnostic: "worker failed\x00 diagnostic details",
+			want:       "worker failed diagnostic details",
+		},
+		{
+			name:       "invalid UTF-8 is replaced",
+			diagnostic: string([]byte{'b', 'a', 'd', 0xff, 't', 'e', 'x', 't'}),
+			want:       "bad\uFFFDtext",
+		},
+		{
+			name:       "only NUL becomes empty",
+			diagnostic: "\x00\x00",
+			want:       "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := normalizeTaskFailureDiagnostic(tt.diagnostic); got != tt.want {
+				t.Fatalf("normalizeTaskFailureDiagnostic() = %q, want %q", got, tt.want)
+			}
+		})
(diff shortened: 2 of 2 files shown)


Bohan-J:
Thanks for this — and for the unusually good bug report behind it in #7098. Exact line references, a runnable repro, a suggested policy, and an explicit ask for a PostgreSQL-backed regression test made this quick to verify and quick to act on.

We reproduced both halves against a real Postgres while reviewing:

```
-- TEXT (agent_task_queue.error) — what you reported
ERROR:  invalid byte sequence for encoding "UTF8": 0x00   -- 22021
```

Two things your report got right that were worth confirming:

- **Removing rather than rejecting is the correct policy.** The surrounding diagnostic is exactly what an operator needs, and dropping the whole error to avoid one invisible byte would trade one blind spot for another.
- **The service boundary is the right place.** Fixing it before classification means the classifier, the transaction, and everything downstream all see the same normalized text.

One detail worth recording for anyone who touches this later: `strings.ToValidUTF8` on its own does **not** solve this, because `0x00` is a valid UTF-8 encoding of U+0000 and passes straight through. The explicit `ReplaceAll` in `normalizeTaskFailureDiagnostic` is load-bearing.

Merging this as the fix for `/fail`. Verifying your report also turned up the same defect on the success path — `agent_task_queue.result` is JSONB, where an embedded NUL is escaped by `encoding/json` and then rejected with a *different* error (`22P05`, `unsupported Unicode escape sequence`) — plus the task-messages and cancel paths. #7124 covers those and will rebase onto this, collapsing its `normalizeTaskFailureDiagnostic` into the shared helper so there is a single definition of what is safe to persist.

Note that neither PR heals rows already wedged in `running`; that needs a separate one-off cleanup.

PR #7117, opened 2026-08-18
fix(web): keep the /favicon.ico redirect relative (MUL-6326)

### LINKED ISSUE #7116: [Bug]: /favicon.ico redirects to http://0.0.0.0:3000/favicon.svg on self-hosted deployments
### Deployment type

self-host

### What happened?

On a self-hosted instance behind a reverse proxy, `GET /favicon.ico` returns a 308 whose `Location` is an absolute URL built from the Next.js server's *bind* address rather than the public origin:

```
$ curl -sSI http://multica.example.internal/favicon.ico
HTTP/1.1 308 Permanent Redirect
Location: http://0.0.0.0:3000/favicon.svg
```

`http://0.0.0.0:3000/` is not reachable from any browser, so the redirect is a dead end and the `/favicon.ico` fallback never resolves. It also leaks the internal bind address to every client that requests it.

I expected the redirect to stay on the origin the client actually used, i.e. `Location: /favicon.svg` (or `http://multica.example.internal/favicon.svg`).

The cause is `apps/web/app/favicon.ico/route.ts`:

```ts
export function GET(request: Request) {
  return Response.redirect(new URL("/favicon.svg", request.url), 308);
}
```

`Response.redirect()` requires an absolute URL, so the route resolves `/favicon.svg` against `request.url`. In a self-hosted deployment `request.url` carries the server's own listen address, not the public one — `Dockerfile.web` sets `ENV HOSTNAME=0.0.0.0` / `ENV PORT=3000`, and a plain `next start` defaults to the same `0.0.0.0:3000`. Neither reflects the host the client used, and the route never consults `Host` / `X-Forwarded-Host`.

This is specific to this route handler. The proxy-level redirects in `proxy.ts` are unaffected because they emit relative locations — same instance, same request:

```
$ curl -sSI http://multica.example.internal/issues     # proxy.ts
HTTP/1.1 307 Temporary Redirect
Location: /login                                        # ✅ relative

$ curl -sSI http://multica.example.internal/favicon.ico # app/favicon.ico/route.ts
HTTP/1.1 308 Permanent Redirect
Location: http://0.0.0.0:3000/favicon.svg               # ❌ absolute, unreachable
```

`https://multica.ai/favicon.ico` correctly returns `Location: https://multica.ai/favicon.svg`, so the official app hides this — the official host's origin happens to match what Next reconstructs. Self-hosters behind a reverse proxy get the broken value.

A relative `Location` is legal for redirects (RFC 7231 §7.1.2) and is origin-independent, so it fixes this without needing to trust or parse any forwarded header.

### Steps to reproduce

1. Run the web app the way a self-host does — either the published image (`Dockerfile.web`, which sets `HOSTNAME=0.0.0.0` / `PORT=3000`) or `pnpm --filter @multica/web start`.
2. Put it behind a reverse proxy on a different public hostname (Caddy `reverse_proxy localhost:3000`, nginx, etc.).
3. `curl -sSI http://<public-host>/favicon.ico`
4. `Location` is `http://0.0.0.0:3000/favicon.svg` instead of a URL on `<public-host>`. Sending the same request straight to `127.0.0.1:3000` with an explicit `Host: <public-host>` header returns the same wrong value, so it is not a proxy misconfiguration.

### Additional context

```shell
Multica web: apps/web @ next 16 (reproduced on a 16.1.x build; app/favicon.ico/route.ts is unchanged on main @ 14c2e4e)
Deployment:  next server on 127.0.0.1:3000 behind Caddy, plain HTTP on an internal hostname
Client:      curl 8.x — but any browser following the redirect fails the same way
```

Found while investigating why an internal deployment had no icon after Safari's "Add to Home Screen". The main part of that turned out to be already fixed on `main` (`app/manifest.ts` + `public/icons/*` + the `apple-touch-icon` link, none of which existed in the older build we were running) — this redirect is the one piece that still reproduces on current `main`.

Happy to fix this one; I'll attach a draft PR with the change plus a regression test.

j10ccc:
## What does this PR do?

`GET /favicon.ico` returned a `Location` that no client could follow on a self-hosted deployment:

```
$ curl -sSI http://multica.example.internal/favicon.ico
HTTP/1.1 308 Permanent Redirect
Location: http://0.0.0.0:3000/favicon.svg
```

`Response.redirect()` only accepts an absolute URL, so the handler had to resolve `/favicon.svg` against `request.url`. In a self-hosted deployment `request.url` carries the server's own listen address rather than the public origin — `Dockerfile.web` sets `ENV HOSTNAME=0.0.0.0` / `ENV PORT=3000`, and a plain `next start` defaults to the same — so the fallback was a dead link and the internal bind address went out to every client that asked for it.

This drops `Response.redirect()` and sets the header directly, so the location stays relative. That is the right approach rather than reading `X-Forwarded-Host`, because a relative `Location` is valid per RFC 7231 §7.1.2 and resolves against whatever origin the client actually used — no header to trust, no configuration to get wrong, and it matches what `proxy.ts` already emits for its own redirects (`Location: /login`).

Not a regression on the official app: `https://multica.ai/favicon.ico` already returns `Location: https://multica.ai/favicon.svg` because that origin happens to match what Next reconstructs. A relative location gives the same result there while also being correct behind a reverse proxy.

## Related Issue

Closes #7116

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `apps/web/app/favicon.ico/route.ts` — return `new Response(null, { status: 308, headers: { Location: "/favicon.svg" } })` instead of `Response.redirect(new URL("/favicon.svg", request.url), 308)`. The handler no longer needs the `request` argument. Added a comment recording why the location must stay relative, so it does not get "tidied" back into `Response.redirect()`.
- `apps/web/app/favicon.ico/route.test.ts` — new. Pins the status and the exact location, plus a named regression test that the location is relative and carries no scheme/host. `// @vitest-environment node` per the testing rules in `CLAUDE.md`; the handler touches no DOM.

## How to Test

Reproduce the bug on `main`:

1. `pnpm --filter @multica/web build && pnpm --filter @multica/web start` (or run the `Dockerfile.web` image, which sets `HOSTNAME=0.0.0.0`).
2. Put it behind a reverse proxy on a different hostname — e.g. Caddy `multica.example.internal { reverse_proxy localhost:3000 }`.
3. `curl -sSI http://multica.example.internal/favicon.ico` → `Location: http://0.0.0.0:3000/favicon.svg`.
4. Sending the request straight to `127.0.0.1:3000` with `-H 'Host: multica.example.internal'` returns the same wrong value, which rules out the proxy.

With this branch, step 3 returns `Location: /favicon.svg` and the browser lands on `http://multica.example.internal/favicon.svg`.

Unit test:

```bash
pnpm --filter @multica/web test -- app/favicon.ico/route.test.ts
```

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — n/a, no UI surface
- [x] I have updated relevant documentation to reflect my changes — n/a, behavior is documented in the route comment

### Verification

An earlier revision of this description said the test suite could not be run — dependency installation was blocked in my environment. That is resolved; everything below was run against this branch rebased onto `main`.

```
$ pnpm exec vitest run app/favicon.ico/route.test.ts
 Test Files  1 passed (1)
      Tests  2 passed (2)

$ pnpm typecheck        # fumadocs-mdx && tsc --noEmit
[MDX] types generated   # clean, no diagnostics

$ pnpm exec eslint app/favicon.ico/
                        # clean, exit 0
```

Live evidence for the bug itself, from a self-hosted instance behind Caddy:

- Before: `/favicon.ico` → `Location: http://0.0.0.0:3000/favicon.svg`, while `/issues` on the same instance → `Location: /login`. That isolates the fault to this handler rather than the proxy.
- The instance has since been upgraded to today's `ghcr.io/multica-ai/multica-web:latest` and **still** reproduces, so this is current-`main` behavior and not an artifact of an old build.
- `new URL("/favicon.svg", "http://0.0.0.0:3000/favicon.ico").href` reproduces the bad value exactly, confirming the mechanism.

Left as a draft for a maintainer to take from here, per our contribution habit — not because anything is unverified.

## AI Disclosure

**AI tool used:** Claude Code (running as a Multica agent)

**Prompt / approach:** Started from a user report that an internal Multica deployment saved via Safari's "Add to Home Screen" had no icon. I probed the deployment's HTML head and icon paths with curl, traced it to a build 350 commits behind `main`, and confirmed the icon half was already fixed upstream by `app/manifest.ts` and `public/icons/*`. The `/favicon.ico` redirect was the one finding that still reproduced against current `main`, so it became this PR. The fix, the comment, and the test were written by the agent; the reproduction evidence is from live curl against the affected instance.

### DIFF
--- a/apps/web/app/favicon.ico/route.test.ts
+++ b/apps/web/app/favicon.ico/route.test.ts
@@ -1,0 +1,24 @@
+// @vitest-environment node
+import { describe, expect, it } from "vitest";
+import { GET } from "./route";
+
+describe("GET /favicon.ico", () => {
+  it("redirects to the SVG favicon", () => {
+    const response = GET();
+
+    expect(response.status).toBe(308);
+    expect(response.headers.get("location")).toBe("/favicon.svg");
+  });
+
+  it("keeps the location relative so it survives a reverse proxy", () => {
+    // Regression: the route used to build the location from `request.url`,
+    // which on a self-hosted deployment is the server's listen address
+    // (`HOSTNAME=0.0.0.0` / `PORT=3000` in Dockerfile.web) rather than the
+    // origin the client used. That shipped `http://0.0.0.0:3000/favicon.svg`
+    // to every browser — a dead link, and the internal bind address leaked.
+    const location = GET().headers.get("location") ?? "";
+
+    expect(location.startsWith("/")).toBe(true);
+    expect(location).not.toMatch(/^https?:\/\//);
+  });
+});
--- a/apps/web/app/favicon.ico/route.test.ts
+++ b/apps/web/app/favicon.ico/route.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment node
 import { describe, expect, it } from "vitest";
-import { GET } from "./route";
+import { dynamic, GET } from "./route";
 
 describe("GET /favicon.ico", () => {
   it("redirects to the SVG favicon", () => {
@@ -20,5 +20,19 @@ describe("GET /favicon.ico", () => {
 
     expect(location.startsWith("/")).toBe(true);
     expect(location).not.toMatch(/^https?:\/\//);
+  });
+
+  it("stays off the prerender cache path", () => {
+    // Regression: `GET()` takes no arguments, so nothing marks the route
+    // dynamic the way the old `request.url` read did. Next then tried to cache
+    // a zero-byte body, which the prerender LRU rejects — one
+    // "calculateSize returned 0" error per request, and every browser asks for
+    // /favicon.ico unprompted.
+    //
+    // This asserts the export, not the framework behaviour it buys: a unit test
+    // cannot observe Next's caching lifecycle. The behaviour itself is checked
+    // against a production build (`pnpm build && pnpm start`), where the fix
+    // removes both `cache-control: s-maxage=31536000` and the error.
+    expect(dynamic).toBe("force-dynamic");
   });
 });
--- a/apps/web/app/favicon.ico/route.ts
+++ b/apps/web/app/favicon.ico/route.ts
@@ -11,6 +11,26 @@
  * §7.1.2 and resolves against whatever origin the client actually used, so it
  * needs no trust in `Host` / `X-Forwarded-Host`.
  */
+
+/**
+ * Keep the handler off the prerender cache path.
+ *
+ * Reading `request.url` used to mark this route dynamic implicitly. Dropping
+ * that argument left nothing dynamic for Next to detect, so it treated the
+ * route as cacheable and tried to store the response — whose body is zero
+ * bytes, which the prerender LRU rejects outright:
+ *
+ *   Failed to update prerender cache for /favicon.ico
+ *   Error: LRUCache: calculateSize returned 0, but size must be > 0.
+ *
+ * Every browser requests `/favicon.ico` unprompted, so that is one error line
+ * per request in every deployment, against a cache write that can never
+ * succeed. Forcing dynamic restores the per-request semantics the old
+ * `request.url` read gave us by accident, without reintroducing the absolute
+ * URL. Do not remove this alongside a "the handler takes no arguments" tidy-up.
+ */
+export const dynamic = "force-dynamic";
+
 export function GET() {
   return new Response(null, {
     status: 308,
--- a/apps/web/app/favicon.ico/route.ts
+++ b/apps/web/app/favicon.ico/route.ts
@@ -1,3 +1,19 @@
-export function GET(request: Request) {
-  return Response.redirect(new URL("/favicon.svg", request.url), 308);
+/**
+ * Legacy `/favicon.ico` requests point at the real SVG favicon.
+ *
+ * The `Location` is deliberately relative. `Response.redirect()` only accepts
+ * an absolute URL, which forces resolving against `request.url` — and on a
+ * self-hosted deployment that carries the server's own listen address, not the
+ * public origin. `Dockerfile.web` sets `HOSTNAME=0.0.0.0` / `PORT=3000`, and a
+ * plain `next start` defaults to the same, so the redirect came back as
+ * `http://0.0.0.0:3000/favicon.svg`: unreachable for every client, and a leak
+ * of the internal bind address. A relative `Location` is valid per RFC 7231
+ * §7.1.2 and resolves against whatever origin the client actually used, so it
+ * needs no trust in `Host` / `X-Forwarded-Host`.
+ */
+export function GET() {
+  return new Response(null, {
+    status: 308,
+    headers: { Location: "/favicon.svg" },
+  });
 }


adaiguoguo (approved):
Independent confirmation from another self-hosted deployment (OpenResty in front of the standalone server): reproduced the same dead redirect — our variant read `location: http://localhost:3000/favicon.svg`, same root as your `0.0.0.0:3000`, different bind host — and we have been running exactly this fix in production since today:

```
$ curl -sSI -H "Host: <public-host>" http://127.0.0.1/favicon.ico
HTTP/1.1 308 Permanent Redirect
location: /favicon.svg

$ curl -sS -o /dev/null -w "%{http_code} %{content_type}" -L .../favicon.ico
200 image/svg+xml
```

I opened #7164 for the same issue before spotting this PR — closing mine in favor of yours, which came first and pins the regression harder in its second test.

j10ccc:
Thanks for the review and for closing #7164 in favour of this one.

Your variant is a genuinely useful data point. You got `location: http://localhost:3000/favicon.svg` where we got `http://0.0.0.0:3000/favicon.svg` — same handler, same mechanism, but the bad host differs with how the server was started. That is the argument for a relative `Location` over reading `X-Forwarded-Host`: there is no single wrong value to special-case, and a relative location is correct for every bind host without the handler needing to know anything about the proxy in front of it. Good to hear it has been running in your production since today.

Marking this ready for merge. Current state:

- **Approved**, no changes requested, no open review threads.
- **All 13 CI checks pass** — `backend-tests`, `frontend-build`, `frontend-test`, `frontend-views-test` (1/2, 2/2), `installer` (ubuntu / macos / windows), `go-vulnerability-scan`, `changes`, `backend`, `frontend`.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no conflicts with `main`.
- Local checks on this branch rebased onto current `main`: `vitest app/favicon.ico/route.test.ts` 2 passed, `pnpm typecheck` (`tsc --noEmit`) clean, `eslint app/favicon.ico/` clean.

One more confirmation since I opened this: I upgraded our self-hosted instance to today's `ghcr.io/multica-ai/multica-web:latest` and `/favicon.ico` **still** returns `Location: http://0.0.0.0:3000/favicon.svg`. So this is current-`main` behaviour, not an artifact of the older build we were originally running.

Merging closes #7116. #7164 is already closed.

I do not have merge rights here, and I have deliberately not enabled auto-merge — the button is yours whenever you are happy with it. Heads up on one unrelated thing: the Vercel bot is waiting on an IndexLabs team member to authorize the preview deploy for an outside contributor. It is not gating `mergeStateStatus`, but it will stay pending until someone on your side clicks through.

Bohan-J (changes requested):
Thanks for the fix, and for re-confirming against today's `latest` image. The relative `Location` is the right call: resolving against `request.url` can never be correct behind a proxy, and staying relative avoids trusting `Host` / `X-Forwarded-Host` at all. Both tests pin the regression well.

One blocker before merge.

### Dropping the `request` parameter puts this route on the prerender cache path

The old body read `request.url`, a dynamic input, which kept the handler out of the cache. The new `GET()` takes no arguments and touches nothing dynamic, so Next treats it as cacheable — and its body is zero bytes, which the prerender LRU rejects.

Production build of `1c708d74`, Next 16.2.6:

```
$ curl -sSI localhost:3000/favicon.ico
HTTP/1.1 308 Permanent Redirect
location: /favicon.svg
cache-control: s-maxage=31536000
x-nextjs-cache: MISS
```

and on every request the server logs:

```
Failed to update prerender cache for /favicon.ico Error: LRUCache: calculateSize returned 0, but size must be > 0.
```

The same production build on current `main` emits neither those two cache headers nor the error, so this is introduced by this PR rather than pre-existing log noise. The redirect itself is fine — the follow-up `GET /favicon.svg` returns `200 image/svg+xml` — but `/favicon.ico` is fetched automatically by every browser, so this becomes a per-request error line in every deployment, and the cache write never succeeds.

Please keep the handler off that path while keeping the relative `Location`, e.g.:

```ts
export const dynamic = "force-dynamic";
```

which restores the per-request semantics the old `request.url` read gave you implicitly.

The current tests call `GET()` directly, so they can't observe the framework's caching lifecycle. After the change, please re-check against a real production build (`pnpm build` then `pnpm start`): relative `Location` still correct, SVG still served, and no cache error in the logs.

### Everything else

No other blocking issues. The change is well scoped, the comment explains exactly why the absolute URL must not come back, and the second test locks that in. Checks on this branch: `vitest app/favicon.ico/route.test.ts` 2/2, `eslint app/favicon.ico/` clean, `pnpm typecheck` clean, `pnpm build` clean (the existing CSS `::highlight` warning is unrelated), and all 13 CI checks green.

j10ccc:
Good catch, and thanks for the thorough writeup — the `main`-baseline comparison in particular made it easy to confirm this was mine rather than pre-existing noise. Fixed in 86a305f.

You were right about the mechanism. Reading `request.url` was marking the route dynamic *implicitly*; dropping the parameter removed that signal, Next decided the handler was cacheable, and the prerender LRU then refused a zero-byte body on every request. I reproduced your exact symptoms on a production build of `1c708d74` before changing anything:

```
$ curl -sSI localhost:3100/favicon.ico
HTTP/1.1 308 Permanent Redirect
x-nextjs-cache: MISS
cache-control: s-maxage=31536000
location: /favicon.svg
```

Three requests produced three `LRUCache: calculateSize returned 0` lines — and `x-nextjs-cache` stayed `MISS` on every one, which is the giveaway that the cache write can never land rather than merely being cold.

Same production build after adding `export const dynamic = "force-dynamic"`:

```
$ curl -sSI localhost:3100/favicon.ico
HTTP/1.1 308 Permanent Redirect
location: /favicon.svg
```

| | before | after |
| --- | --- | --- |
| `cache-control: s-maxage=31536000` | present | gone |
| `x-nextjs-cache` | `MISS`, every request | gone |
| `calculateSize` errors | 3 requests → 3 errors | 0 |
| other errors in the server log | — | none |
| redirect target | `/favicon.svg` | `/favicon.svg` |
| following it | — | `200 image/svg+xml` |

So all three things you asked me to re-check hold: relative `Location` still correct, SVG still served, no cache error in the logs.

Two notes on how I wrote it up:

- The config export carries a comment explaining *why* it exists. Without one it reads like a stray line, and the obvious "this handler takes no arguments, so this must be leftover" tidy-up would walk straight back into this bug. That failure mode is exactly how I introduced it.
- I added a third unit test asserting `dynamic === "force-dynamic"`, but its comment says plainly that it pins the export and **not** the framework behaviour — a unit test cannot observe Next's caching lifecycle, and I did not want to leave something that looks like coverage it isn't. The behavioural check is the production-build run above.

Local checks on this commit: `vitest app/favicon.ico/route.test.ts` 3 passed, `pnpm typecheck` clean, `eslint app/favicon.ico/` clean, `next build` clean. CI is running on the new head.

Ready for another look whenever you have a moment.

Bohan-J (approved):
Thanks for the quick turnaround — `86a305f` is exactly the fix, and your commit message explains the mechanism more precisely than my review did: `request.url` was marking the route dynamic *implicitly*, and removing the parameter took that signal with it.

Re-checked on the new head: the production build no longer emits `cache-control: s-maxage=31536000` or `x-nextjs-cache`, `/favicon.ico` still returns `308` with `location: /favicon.svg`, following it gives `200 image/svg+xml`, and the logs are clean — no `calculateSize` errors across repeated requests. 14/14 checks green, `mergeStateStatus: CLEAN`.

The third test is a nice touch, and I appreciate that its comment says plainly what a unit test can and cannot observe here, pointing at the production build for the behaviour itself. That is the honest framing.

Thanks as well for the original report, the reproduction against the current `latest` image, and for coordinating with #7164 rather than letting two PRs race. Merging now — this closes #7116.

PR #7119, opened 2026-08-18
MUL-6330: fix(editor): rank exact-prefix skill matches first in the / picker

tim-smart:
## Problem

The chat `/` skill picker kept the agent's configured skill order while matching both names and descriptions. A skill whose name matched the query could therefore appear behind description-only matches.

## Fix

Rank matches into stable tiers before applying the 20-item cap:

1. exact name
2. name prefix
3. name substring
4. description only

Configured order still breaks ties within each tier.

## Tests

- `pnpm -C packages/views exec vitest run editor/extensions/slash-command-suggestion.test.tsx` (30 passed)
- `pnpm -C packages/views typecheck`
- `pnpm -C packages/views exec eslint editor/extensions/slash-command-suggestion.tsx editor/extensions/slash-command-suggestion.test.tsx`

### DIFF
--- a/packages/views/editor/extensions/slash-command-suggestion.test.tsx
+++ b/packages/views/editor/extensions/slash-command-suggestion.test.tsx
@@ -152,6 +152,103 @@ describe("slash command suggestion items", () => {
     expect(items(qc, "pull").map((i) => i.id)).toEqual(["s2"]);
   });
 
+  it("ranks name prefix matches above description-only matches", () => {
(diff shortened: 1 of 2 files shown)


multica-eve (approved):
Thanks for the contribution! The relevance tiers are clear, preserve configured order within each tier, and are covered by focused regression tests. This makes the slash skill picker much easier to use.

PR #7126, opened 2026-08-18
MUL-6323 fix(views): stop blaming the daemon when the backend is the stale half

### LINKED ISSUE #7113: [Bug]: Cannot using worktree because version comparison bug
### Deployment type

self-host

### What happened?

The client is MacOS Desktop

When add a local dir to project, cannot select worktree.

Seems the version comparison function has bug.

Image

### Steps to reproduce

1. Go to project
2. Add local directory

### Screenshots (optional)

_No response_

```shell

```

Bohan-J:
Fixes #7113.

## What the user saw

A self-hosted macOS Desktop user could not select worktree mode when adding a local directory. The disabled option explained itself:

> 这个文件夹所在机器上的 Multica 是 **v0.4.28**，并行运行需要 **0.4.24** 或更新的版本。

v0.4.28 was the newest release that day. There was no version left to upgrade to, which is why the report reads "version comparison bug".

## Why it happened

Two separate things, one visible.

**The copy outlived its check.** Since MUL-5707 (#6904) all three worktree gates read the declared `local-worktree-v1` capability; `MIN_LOCAL_WORKTREE_CLI_VERSION` survived only as the number this string interpolated, and `localWorktreeSupported()` had no callers at all. So every capability miss — whatever its cause — was reported as a version shortfall the user had already cleared. #6904 rewrote the server's 422 to stop quoting versions; this dialog was missed.

**The capability is the server's to record.** The daemon advertises it on `X-Client-Capabilities`, and the server persists it onto the runtime row at registration — code that ships in v0.4.25. Every older server writes runtime metadata with **no `capabilities` key at all**. Desktop bundles its own renderer *and* its own daemon and updates them together, so the client routinely runs ahead of a backend the operator upgrades by hand. The daemon advertises, the old backend drops it, the UI blames the machine that was already newest — and updating that machine changes nothing.

## What changes

- The gate answers with `LocalWorktreeSupport` (`supported` / `daemon_unsupported` / `server_capability_blind`) instead of a boolean. The discriminator is the **presence** of the `capabilities` key, not its contents: a capability-aware server always writes it (`null` when the daemon sent no header), so a row without it can only have come from an older server.
- Each negative gets copy pointing at the half that can actually be fixed, in all four locales. No message quotes a daemon version any more.
- Registration-failure rows for custom runtime profiles now carry the advertised capabilities too. They are written *after* the successful runtimes of the same request, so until the first heartbeat they are the newest row for that daemon — the row every worktree gate reads. A failed custom profile briefly made a capable daemon look incapable, and under the new logic would have made a current backend look stale.
- Deletes `localWorktreeSupported` (no callers) and the frontend `MIN_LOCAL_WORKTREE_CLI_VERSION` (only fed the deleted copy). The server-side `MinLocalWorktreeCLIVersion` is untouched — it is still the number in the 422 payload.

Still fails closed everywhere: the mode exists to keep agents out of the user's working copy, so an unknown answer stays blocked. Only the explanation changes.

## Verification

- `pnpm test` for `@multica/core` and `@multica/views` — the dialog, create-project and `localWorktreeSupport` suites cover both negatives, and two cases assert no version number reaches the user.
- `TestDaemonRegister_RecordsRuntimeProfileRegistrationFailure` extended for the capability round-trip; confirmed it fails without the handler change.
- `pnpm typecheck` and `pnpm lint` clean for `@multica/core`, `@multica/views`, `@multica/desktop`, `@multica/web`; `go vet` clean.

Not covered locally: the actual old-backend-plus-new-client pairing, which needs a v0.4.24 server to reproduce. Worth confirming the reporter's backend version on the issue.


Bohan-J:
Both must-fixes addressed in 8554c08.

**Must-fix 1 — the check now asks the live backend.** You were right that a missing `capabilities` key only proves the server was old *when the row was written*; nothing rewrites that metadata on upgrade, and since the daemon does not re-register on a timer the state can sit there indefinitely. The gate now reads `/api/config` → `server_version` (via `useConfigStore`, as you pointed at) and the stale-row case gets its own answer — `runtime_registration_stale`, with copy asking the user to restart Multica on that machine so it registers again, since neither half is actually out of date.

One detail worth flagging, because it decides the default: an **absent** `server_version` had to read as capability-*aware*, not old. `/api/config` omits the field on the managed cloud on purpose (MUL-4108) and leaves it empty for unstamped dev builds, and the field itself shipped in v0.4.2 — 23 patches before the handshake. So every self-hosted release that could genuinely be blind still names itself, while defaulting the other way would tell cloud users to upgrade a backend they don't run. Covered by tests for `""`, dev-describe and unparsable values, plus the `server >= 0.4.25 + legacy row` regression you asked for, at both the helper and the create-project modal level.

**Must-fix 2 — contract rewritten, not deleted.** `worktree_version_floor_test.go` is gone; `worktree_capability_contract_test.go` replaces it and pins what the two sides must still agree on — the `local-worktree-v1` token, which the daemon advertises, the server stores and the UI looks for. A typo there disables worktree mode for everyone with no error anywhere, so it is worth more than the version number ever was. The version floor is not re-added.

**Non-blocking item** — comment corrected to "No DAEMON version numbers", noting the backend floor is deliberately shown because it is what an operator upgrades past. The test name (`never quotes a daemon version at the user`) already said daemon, and now covers all three blocked reasons.

Verification: `pnpm typecheck` clean across core/views/desktop/web; `packages/views` 363 files / 4287 tests pass, `packages/core` 128 / 1460 pass; `go vet ./...` clean; `go test ./pkg/agent/... ./internal/handler/` pass — including the previously failing floor test's replacement.

Bohan-J:
Fixed in 22fd221 — you were right, and the loop you described was real.

**Awareness is three-valued now.** `serverCapabilityAwareness()` returns `aware` / `blind` / `unknown`; an unnamed server no longer defaults to either. Your v0.4.1 counterexample is exactly the case that broke: no `server_version`, no `capabilities`, so "restart Multica there" re-registered against the same blind backend and produced an identical key-less row, with the one remedy that works never mentioned.

**An unnamed backend is resolved from evidence, not assumption.** Before falling back to `unknown`, the gate checks whether *any* runtime row in the workspace carries a `capabilities` key. One cannot appear unless a capability-aware server wrote it, and servers move forward — so a single such row anywhere proves the deployment learned the handshake, whoever it belongs to. That settles the managed-cloud shape (no version reported, several machines, one registered before v0.4.25) precisely, without needing a version to read. The inference only ever promotes to `aware`: every row missing the key is equally what an old self-hosted server and a one-machine workspace look like, so that direction stays `unknown` and says so.

**`unknown` gets its own fail-closed copy** naming both remedies — "if you run a self-hosted Multica server, upgrade it; otherwise restart Multica on that machine" — rather than picking one that may be a dead end.

Regression tests, keeping the managed-cloud/dev empty-value cases you asked to preserve:
- v0.4.1-like config (no `server_version`) + legacy row + no other evidence → `capability_source_unknown`, at the helper and through the create-project modal.
- Same unnamed server, but another machine's row carries capabilities → `runtime_registration_stale`.
- A named-blind server is *not* overridden by such a row.
- The "never quotes a daemon version" case now covers all four blocked reasons.

**Copy nit taken** — the re-register string explained itself in the vocabulary of the handshake ("record what it supports"). Now: "registered before this server could tell which machines support running tasks in parallel", and the zh/ja/ko strings follow the same phrasing.

Verification: `pnpm typecheck` clean across core/views/desktop/web; `packages/views` 363 files / 4289 tests and `packages/core` 128 / 1462 pass; lint 0 errors; `go vet ./...` clean; the capability-token contract test passes.

Bohan-J:
Reworked in dd92675 after review feedback that the client-side judgement had grown out of proportion — it had, and the reason is worth stating: it was trying to *predict* an answer the server already gives authoritatively. Each round added a layer of inference over data the backend wrote (capability present/absent → the backend's version to interpret the absence → other rows' evidence to interpret an unnamed backend), all to avoid asking the backend.

**So it asks.** The UI no longer blocks worktree mode on capability at all. The only client-side blocker left is `not_git` — local, certain, already checked by the picker. The mode stays selectable, the save is gated server-side, and the rejection lands in the dialog's existing `errorMessage` slot using the wording #6904 already wrote for it.

Deleted: the awareness tri-state, the workspace-wide evidence scan, the `server_version` plumbing, three UI reasons and four copy strings across four languages. `cli-version.ts` loses 153 lines and keeps one exported function.

**This also fixes the reported case outright instead of explaining it.** A self-hosted backend older than the handshake runs the version gate it shipped with, which accepts a v0.4.28 daemon — so that user simply gets worktree mode, with nothing to upgrade.

**One thing had to be added for "the server validates" to be true.** `POST /api/projects` attaches resources in the same transaction and never ran the worktree gate that `POST/PUT /resources` have always run, so it would store a mode the machine cannot run and leave the claim gate to cancel the task later. It runs the same gate now, before the transaction opens — verified by a new test that fails without it (201 + resource stored) and passes with it (422, no project left behind).

The capability signal survives for exactly one job: deciding whether to **preselect** parallel mode. Wrong there costs a radio button, not a blocked user. A machine that hasn't advertised isn't preselected into a mode whose rejection would fail the whole project creation — but it stays selectable, because an old row is not proof the machine cannot.

Verification: `pnpm typecheck` clean across core/views/desktop/web; `packages/views` 363 files / 4282 tests and `packages/core` 128 / 1457 pass; lint 0 errors; `go vet ./...` clean; `go test ./pkg/agent ./internal/handler` pass.

Bohan-J:
Fixed in 30565c7. You were right, and this was the worse half of the issue rather than a variant of it — I verified the whole chain against the tags:

- `v0.4.24:server/internal/handler/project_resource.go` unmarshals the ref into a struct with no `ExecutionMode` field and re-marshals it, so `execution_mode: "worktree"` is dropped and answered **201**.
- `localDirectoryAssignment.ValidateExecutionMode` treats an empty mode as valid and `UsesWorktree()` returns false, so the daemon runs **in place**.
- Neither the save gate nor the claim gate exists on that server.

So the user gets a success, and an agent edits the directory they asked to isolate. My earlier claim that an old backend would run "its own version gate" was simply wrong: worktree shipped in v0.4.25, so v0.4.24 has no concept of the mode at all.

**The client asks first now.** `/api/config` carries `local_worktree_supported`, a property of the build — if this code is running, the save gate is running with it. Absent (what every too-old server sends) blocks the mode with copy that names the actual problem and its actual consequence: *"Your Multica server is too old to run tasks in parallel — it would quietly run them in this folder instead."*

Two details worth flagging:
- The zod schema parses a missing key to `false`, not `.optional()` → `undefined`. The other config booleans are optional because they gate cosmetics; this one gates a data-safety promise, so a consumer that forgets `=== true` should still fail closed.
- The server declares it unconditionally — not behind a deployment check, env var or feature flag, each of which would hide worktree mode from the users who can actually run it. Pinned by a test, along with the key being present in the JSON body rather than omitted.

Regression coverage for the pairing you named: newest Desktop + newest daemon + `local_worktree_supported` absent → option disabled, "server too old" copy, and **not preselected** either (preselecting would submit a mode the server would silently downgrade). Plus schema drift tests for absent / malformed / true.

Nits cleared: the stale `localWorktreeSupport` comment reference and the blank runs in both test files.

The simplification survives — one declared boolean from the live server, no inference over runtime rows or version strings, per-machine capability still left to the server's save-time gate.

Verification: `pnpm typecheck` clean across core/views/desktop/web; `packages/views` 363 files / 4284 tests and `packages/core` 128 / 1460 pass; lint 0 errors; `go vet ./...` clean; `go test ./internal/handler ./pkg/agent` pass.

Bohan-J:
Fixed in ecfe88c. Confirmed the path — and it's the same failure as the main bug, through a door the previous commit left open.

**Rename no longer sends the ref at all.** Rather than gating renames on the capability flag, the update stops carrying `resource_ref` entirely: `data: { label }`. `UpdateProjectResource` has always treated an omitted `resource_ref` as "keep the stored one" — I checked `v0.4.24`'s copy of the handler, same semantics — so `execution_mode` survives on every server version, and rename keeps working on old backends instead of becoming another blocked action.

That also closes the general shape rather than this one instance: *any* unrelated edit that resends a ref gets it re-normalised by whatever the server understands today. Not resending it is the property worth having.

The consequence to know about: the label now lives in the top-level column (which the API already offered for exactly this), and display reads top-level first, ref's copy second — in `project-resources-section.tsx` and `local-directory-hint.tsx` via one shared resolver. Rows created before this keep the name they were given until someone renames them.

Regression test as requested — existing `execution_mode: "worktree"` + `localWorktreeSupported: false` + rename → asserts the payload is exactly `{ label }` and has no `resource_ref`. Verified it fails against the old code (`expected { resource_ref: {…} } to deeply equal { label: 'Client Worktree' }`). A second case pins the display precedence, since the rename is only visible if the top-level label outranks the stale one in the ref.

**Nit taken, and it was a real inaccuracy.** The flag is absent on every server that predates the *signal*, not only on those that predate the fix — v0.4.25–v0.4.28 do gate the save but cannot say so, and are treated as unable. Worth being explicit that this is a deliberate cost: those self-hosted deployments lose the worktree option until their backend is updated, because the client cannot distinguish them from the ones that silently downgrade, and only one of the two guesses is safe. Wording corrected in the schema, the config store, the handler and both tests.

Verification: `pnpm typecheck` clean across core/views/desktop/web; `packages/views` 364 files / 4286 tests and `packages/core` 128 / 1460 pass; lint 0 errors; `go vet ./...` clean; full `internal/handler` package passes.

Bohan-J:
Fixed in 500d81d. The race was real — I reproduced both halves as failing tests before fixing them.

**Intent is now classified by what else changed.** A ref counts as a ≤ v0.4.28 rename only when its *sole* difference from the stored ref is the label. Anything else is an edit that happens to be carrying a name, so the row keeps whatever it is called today — and the stale copy the request brought gets healed to that name rather than written through. Comparison is on decoded values, not bytes (the stored ref may come from a different build), and unknown keys count: a difference this binary cannot interpret is still a difference, and calling such a request a pure rename would be a guess.

**The same predicate fixes the second point from the other side.** A legacy rename resends the ref, so the capability gate refused it whenever the machine's runtime registration had drifted — failing an edit that changes nothing about what runs, exactly what that gate's own comment promised label-only updates would be spared. A rename now skips it; the row already says worktree, and the claim gate is what keeps it from running somewhere that cannot.

Both verified against the unfixed code:
- `a stale mode snapshot rolled back the rename: column = ...`
- `a rename that changes no execution semantics must not hit the capability gate: 422 daemon_version_unsupported`

New handler tests: the open-dialog → other-device-rename → save-mode sequence (asserting both homes keep B's name), a pure legacy rename still reaching the column, the gate skipped for a rename on a worktree row with no capable runtime, and the gate still firing when the same client actually changes the mode.

**One deliberate behaviour change worth calling out.** `TestProjectResourceUpdateLifecycle` pinned the old reading for a request that changes `local_path`, `daemon_id` *and* `label` at once. No client sends that shape — renames change only the label, the mode dialog only the mode — and honouring an embedded label there is indistinguishable from honouring a stale one, so it now pins the conservative answer: the ref fields update, the name stays. Callers that mean to rename have the top-level `label` field. I updated the test rather than special-casing it.

**Nit taken** — "byte-for-byte" was overstated for a map patch that re-marshals, and JSONB does not promise raw bytes anyway. Comment and test assertion now say every unknown key and value is preserved, which is what the code actually guarantees and what matters.

On the follow-up scoping, both corrections look right to me and I'd fold them into that issue rather than this PR: the existing CLAUDE.md "combine signals, don't rely on one backend boolean" line directly contradicts what we settled on and should be replaced, not supplemented; and a lint rule must target *inferring server capability from runtime metadata*, since `runtimeAdvertisesLocalWorktree` reading a daemon's self-declared capability is the correct use, as is showing a version in an upgrade hint.

Verification: full `internal/handler` package passes; `go vet ./...` clean; `pnpm typecheck` clean across core/views/desktop/web; views projects+modals suites 12 files / 65 tests pass; lint 0 errors.

PR #7136, opened 2026-08-18
feat(lark): push inbox notifications to a bound member's Lark DM

adaiguoguo:
WeCom's smart bot already pushes `inbox:new` to a member who has linked their account (`internal/integrations/wecom/outbound.go`); Lark had no equivalent, so a Lark-bound member only discovered assignments, mentions, comments and status changes the next time they opened the app. This adds the Lark side.

The outbound Patcher subscribes to `inbox:new` and delivers to the member's DM when they have a binding row, addressing them by `open_id` so Lark opens the 1:1 conversation itself (`SendTextParams` / `SendMarkdownCardParams` gain `ReceiveIDType`; empty keeps every existing caller on `chat_id`). Every miss — agent recipient, unbound member, missing installation, send failure — is a no-op, and the in-app inbox is unaffected either way. Notification preferences need no second check: a muted type never produces an inbox item.

## Presentation

The notification is a schema-2.0 **markdown card**, not plain text: notification bodies are member/agent-authored markdown, and they land in the same conversation as the agent's reply cards — both now render the same way. Bold `【type】title` header, body verbatim, one 查看详情 link, and a chat-list summary line.

Assignment-family notifications arrive body-less upstream (`notifyDirect` passes `""`), so the push backfills the issue description — it is exactly what the recipient needs to decide whether to act. Status flips stay body-less: a description repeated on every transition is spam.

## One growing card per issue

A burst of agent activity on one issue (status flip, comment, another flip) used to be a card per event within seconds. Now the first event sends a card and remembers its message id; later events inside a 15-minute window **patch** that card, appending labeled blocks. Distinct issues keep distinct cards, and a refused patch (recalled/expired card) falls back to a fresh card.

This is why markdown cards now declare `update_multi` at send time — Lark silently refuses to patch a card that did not opt in when it was sent (the same trap `defaultRenderer.Render` documents). Inert for cards that are never patched.

## Why any bound bot carries all of the member's notifications

The bot is agent-scoped by the installation model, but the inbox is member-scoped, and the binding row is the member's channel identity — `FindChannelBindingForMember` was introduced for exactly this on the WeCom side ("the wecom smart-bot inbox-notification path uses this to decide whether to deliver via the bot at all"), deliberately keyed by member + channel_type rather than by installation, most-recently-bound winning. This PR follows that contract rather than inventing a narrower one: scoping pushes to the bot's own agent would silently drop every other notification the member subscribed to, which is the failure mode the inbox exists to prevent. A workspace-level notification bot, unattached to any agent, would be a cleaner home for this — but that is an installation-model change, out of scope here.

## Deliberate divergences from the WeCom path

- **Delivery is fire-and-forget on a goroutine.** The bus delivers synchronously and `inbox:new` is published on user-facing write paths, so an outbound Lark HTTP call would tax those API responses. WeCom delivers inline because its send is an in-process handoff to a connected WS sender; this one leaves the building.
- **The detail link accepts `http://`.** An https-only guard would silence the link for every self-hosted deployment served over plain http on an internal network; the value is operator-set, never member-authored. Happy to align with WeCom if you'd rather keep the guard uniform.
- **Only the title gets `](` link-breaking.** The title is spliced into the bot's own bold line; the body renders verbatim like the reply cards beside it, links included.

## A trap named for the next channel

This platform's binding rows carry `channel_type: "feishu"` (`channel_store.go`), not `"lark"` — the package name and the stored value differ, and looking up the wrong one fails silently, because no-binding is also the ordinary unbound-member case.

## Tests

Sixteen cases in `inbox_push_test.go`: open_id addressing, async wiring, unbound member, agent recipient, malformed payloads, verbatim body rendering, description backfill (and its absence on status changes), title-link breaking with body links preserved, pathological-body clamping that closes an open code fence, link omission without an app URL, http URLs, slug fallback, digest folding/expiry/isolation/patch-failure fallback. `go test ./internal/integrations/lark/ -race` is green.

Verified end-to-end on a self-hosted deployment: assignment, mention, comment and status-change notifications land in the bound member's Feishu DM, and a create+comment+status burst renders as one card with two appended blocks.

### DIFF
--- a/server/internal/integrations/lark/client.go
+++ b/server/internal/integrations/lark/client.go
@@ -282,7 +282,21 @@ type SendTextParams struct {
 	// ReplyTarget threads the text reply back into a Lark topic; see
 	// ReplyTarget. Empty keeps the chat-level send.
 	ReplyTarget ReplyTarget
+	// ReceiveIDType tells Lark how to read ChatID. Empty means chat_id —
+	// the historical chat-level send. ReceiveIDOpenID addresses a person
+	// instead, which is how the inbox push reaches a bound member without
+	// a pre-existing chat. Ignored when ReplyTarget is set, since a reply
+	// is addressed by message id.
+	ReceiveIDType ReceiveIDType
 }
+
+// ReceiveIDType is Lark's receive_id_type query parameter: it selects
+// which identifier namespace receive_id is read in.
+type ReceiveIDType string
+
+// ReceiveIDOpenID addresses a single user by their per-app open_id. The
+// bot opens (or reuses) a 1:1 conversation with that person.
+const ReceiveIDOpenID ReceiveIDType = "open_id"
 
 // SendMarkdownCardParams is the input shape for posting an agent
 // reply as a Lark interactive card with a markdown body element.
@@ -302,6 +316,10 @@ type SendMarkdownCardParams struct {
 	// ReplyTarget threads the card reply back into a Lark topic; see
 	// ReplyTarget. Empty keeps the chat-level send.
 	ReplyTarget ReplyTarget
+	// ReceiveIDType tells Lark how to read ChatID; see the field of the
+	// same name on SendTextParams. Empty means chat_id. Ignored when
+	// ReplyTarget is set.
+	ReceiveIDType ReceiveIDType
 }
 
 // BindingPromptParams carries the data needed to render and send the
--- a/server/internal/integrations/lark/http_client.go
+++ b/server/internal/integrations/lark/http_client.go
@@ -270,7 +270,7 @@ func (c *httpAPIClient) invalidateToken(appID string) {
 // goes to the chat-level send endpoint keyed by receive_id=chat_id, the
 // historical behavior. Body is map[string]any (not map[string]string)
 // because reply_in_thread is a bool.
-func outboundMessageRequest(chatID ChatID, msgType, content string, target ReplyTarget) (string, map[string]any) {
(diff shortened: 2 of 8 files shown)


adaiguoguo:
This has been green and conflict-free for three days, so I suspect the size is the obstacle rather than the content — 1079 lines carrying a new subsystem is a lot to ask anyone to hold in their head at once.

Happy to re-cut it into two:

1. **Delivery** — the transport changes (open_id addressing, `update_multi` in the card envelope) plus the `inbox:new` subscriber and card composition. Complete on its own: an inbox item arrives in the member's Lark DM as one card.
2. **Aggregation** — the 15-minute per-issue digest that patches one growing card instead of sending N. An optimisation on top of (1), and where the subtlety lives: patch-failure fallback, the per-key lock.

That is not a pure file split — (1) would call a plain send where it currently calls the digest — so I would only do it if it genuinely helps. If you would rather review it whole, or if something specific in it is the blocker, say so and I will work with whichever is easier for you.

adaiguoguo:
As offered above, here is the re-cut: the delivery half is now #7535 — same transport and card composition, individual cards only, −250 lines and none of the locking/patching subtleties. I'll keep this PR open as the reference for the aggregation half and re-cut that on top once delivery lands (or close this one then, whichever you prefer).

adaiguoguo:
Closing in favor of #7535 (the delivery half). The aggregation half will be re-cut on top once delivery lands; this PR stays as the reference for that design — the per-key lock and the patch-failure fallback discussion live here.

PR #7138, opened 2026-08-18
MUL-6360 feat(dingtalk): show linked identities to workspace admins

yyclaw:
## What does this PR do?

Displays, on each DingTalk installation row, the DingTalk Staff ID(s) linked to the authenticated workspace owner's or admin's current Multica user for that installation.

The purpose is identity confirmation and troubleshooting. When a workspace has multiple DingTalk installations or a user has linked multiple DingTalk identities, an administrator can verify which human DingTalk identity their current Multica user is using for each installation.

Each displayed Staff ID represents a human DingTalk identity. It does not identify the bot, reveal the bot name, identify the DingTalk application owner, or list other workspace members who have used the bot.

Privacy boundaries:

- The installation-list API returns linked Staff IDs only to workspace owners and admins.
- Each owner or admin receives only identities linked to their own Multica user.
- Regular workspace members can still view connected installations, but the API omits `bound_dingtalk_user_ids`.
- When the current owner or admin has no identity linked to an installation, the UI displays no identity label.

These restrictions are enforced by the server. The UI independently checks workspace permissions before rendering the field.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add a workspace- and user-scoped query that returns the current user's bindings by DingTalk installation.
- Return `bound_dingtalk_user_ids` from the installation-list API only for workspace owners and admins.
- Display linked DingTalk identities beside the installation timestamp in Settings → Integrations → DingTalk.
- Publish a realtime event after an identity is bound so an open settings page refreshes immediately.
- Safely parse missing or malformed binding fields as an empty list.
- Add localized identity labels for English, Simplified Chinese, Japanese, and Korean.
- Add backend, API-schema, realtime, permission-boundary, and UI coverage.
- Preserve the existing member-visible installation list without exposing linked identities to regular members.

## How to Test

1. As a workspace owner or admin, link a DingTalk identity and open Settings → Integrations → DingTalk.
2. Confirm the corresponding installation row displays `Linked DingTalk identity: <staff ID>`.
3. Sign in as a regular workspace member and confirm the installation remains visible, but neither the identity label nor Staff ID appears.
4. Link an identity while the settings page is open and confirm the installation list refreshes.
5. Confirm an owner or admin with no linked identity sees only the installation timestamp.
6. Confirm older or malformed API responses without a valid `bound_dingtalk_user_ids` array degrade to an empty list without breaking the page.

Automated verification:

- `go test ./internal/integrations/dingtalk ./internal/handler`
- `pnpm --filter @multica/core test -- api/schema.test.ts` (Core full suite: 1,458 tests)
- `pnpm --filter @multica/views test -- settings/components/dingtalk-tab.test.tsx` (Views full suite: 4,295 tests)
- After the copy update: targeted Core schema tests (74), Views component tests (18), Go handler coverage, and Views typecheck.

## Risks and Limitations

- A DingTalk Staff ID identifies a human DingTalk member. Returning identities belonging to other members would expose personal identity information, so the lookup is scoped to the authenticated user and the field is omitted for non-admin roles.
- A Staff ID cannot be used to determine the bot name or the DingTalk application owner.
- This change does not request additional DingTalk permissions, call additional DingTalk APIs, or require a database migration.
- Older clients ignore the additional response field. Newer clients default missing or malformed values to an empty list when communicating with an older server.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (N/A: no setup or workflow changed)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (N/A)

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:**

Used Codex to trace DingTalk identity bindings from callback and redemption through the database, API response, realtime invalidation, and settings UI. The implementation was reviewed specifically for workspace-role enforcement, per-user data scoping, API compatibility, and prevention of Staff ID disclosure to regular members.

## Screenshots

image

### DIFF
--- a/packages/core/api/schema.test.ts
+++ b/packages/core/api/schema.test.ts
@@ -387,7 +387,11 @@ describe("ApiClient schema fallback", () => {
 
     it("parses the new-server group-routing capability", async () => {
       stubFetchJson({
-        installations: [{ id: "dt-1", status: "active" }],
+        installations: [{
+          id: "dt-1",
+          status: "active",
+          bound_dingtalk_user_ids: ["staff-1001"],
+        }],
         configured: true,
         install_supported: true,
         group_routing_supported: true,
@@ -395,6 +399,21 @@ describe("ApiClient schema fallback", () => {
       const client = new ApiClient("https://api.example.test");
       const res = await client.listDingTalkInstallations("ws-1");
       expect(res.group_routing_supported).toBe(true);
+      expect(res.installations[0]?.bound_dingtalk_user_ids).toEqual(["staff-1001"]);
+    });
+
+    it("defaults missing or malformed linked-identity IDs to an empty list", async () => {
+      stubFetchJson({
+        installations: [
+          { id: "dt-old", status: "active" },
+          { id: "dt-broken", status: "active", bound_dingtalk_user_ids: "staff-1001" },
+        ],
+        configured: true,
+      });
+      const client = new ApiClient("https://api.example.test");
+      const res = await client.listDingTalkInstallations("ws-1");
+      expect(res.installations[0]?.bound_dingtalk_user_ids).toEqual([]);
+      expect(res.installations[1]?.bound_dingtalk_user_ids).toEqual([]);
     });
   });
 
--- a/packages/core/api/schemas.ts
+++ b/packages/core/api/schemas.ts
@@ -2524,6 +2524,7 @@ export const DingTalkInstallationSchema = z.object({
   installed_at: z.string().default(""),
   created_at: z.string().default(""),
   updated_at: z.string().default(""),
+  bound_dingtalk_user_ids: z.array(z.string()).catch([]).default([]),
 }).loose();
 
 export const EMPTY_DINGTALK_INSTALLATION: DingTalkInstallation = {
@@ -2535,6 +2536,7 @@ export const EMPTY_DINGTALK_INSTALLATION: DingTalkInstallation = {
   installed_at: "",
   created_at: "",
   updated_at: "",
+  bound_dingtalk_user_ids: [],
 };
 
 export const ListDingTalkInstallationsResponseSchema = z.object({
--- a/packages/core/types/dingtalk.ts
+++ b/packages/core/types/dingtalk.ts
@@ -14,6 +14,10 @@ export interface DingTalkInstallation {
   installed_at: string;
   created_at: string;
   updated_at: string;
+  /** DingTalk staff ids linked by the currently authenticated Multica user for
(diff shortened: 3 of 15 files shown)


Bohan-J:
Thanks @yyclaw — merging this.

The scope reads right: the field is per-user and per-installation, so an owner/admin only ever sees their own DingTalk identity, and the server omits it entirely for regular members rather than sending it and hiding it client-side.

Two decisions worth recording before they disappear into a squash commit:

- `z.array(z.string()).catch([])` in `schemas.ts` isn't redundant defensiveness. `listDingTalkInstallations` goes through `parseWithFallback`, so one malformed element would collapse the whole response into the empty fallback and blank out the installation list. The `catch` keeps the blast radius at one field.
- `TestRedeemDingTalkBindingTokenPublishesAccountBindingUpdateAfterCommit` reads the binding back over a separate pool connection to prove the transaction committed before the event fired, rather than inferring it from statement order. That's the right way to pin a commit/publish ordering.

On the one point we went back and forth on — a failing binding query takes the whole installation list to 500. We're landing it as-is: degrading to an empty array would disguise "read failed" as "you have no binding", which is the worse answer for a feature whose entire job is identity confirmation. Still worth revisiting later as a separate request or a distinguishable error state, since this endpoint also backs the DingTalk badge on the agent detail page — but not a blocker.

Non-blocking follow-ups, if you want to pick any of them up:

- The new permission-boundary and error-path assertions live inside `TestListDingTalkInstallationsAdvertisesGroupRoutingCapability/configured_deployment`, whose name is about group routing — worth splitting out. Related: the `LOCK TABLE ... ACCESS EXCLUSIVE` plus fixed 250ms timeout used to force the failure becomes a flake source if this package ever gains `t.Parallel`.
- `bound_dingtalk_user_ids` is channel-specific naming over the generic `channel_user_binding` table. Slack/Lark/WeCom will want the same shape; a neutral name would let them copy it verbatim.
- `truncate` on the installation line now applies to everyone, and `title` is `undefined` when there is no identity — so on narrow screens the "Installed …" text can be clipped with no way to read it in full.
- The field is optional in `packages/core/types/dingtalk.ts` but the schema always fills it via `.default([])`, so callers write `?? []` for a case that cannot happen.

One note on the PR description: scoping the query to the requesting user is the privacy control; the owner/admin gate is a product decision about who this is for, not a second privacy layer. Whichever reason is written down is what the next channel copying this pattern will inherit.

PR #7141, opened 2026-08-18
feat(i18n): add French locale

foXaCe:
## What does this PR do?

Adds `fr` (French) as a supported UI locale for web and desktop.

Multica ships four locales today (`en`, `zh-Hans`, `ko`, `ja`). This adds a
fifth. The approach follows the existing locale contract exactly rather than
introducing anything new: the same 25-file resource layout under
`packages/views/locales/`, the same registration points, and
`locales/parity.test.ts` as the completeness net.

Thinking path: the locale set is declared in one place (`SupportedLocale` /
`SUPPORTED_LOCALES`), consumed by `matchLocale`, the views resource bundle, the
settings language picker, and the per-app `HTML_LANG` maps. Adding a locale
means satisfying each of those, and leaving alone the surfaces that are
authored content rather than UI strings — landing marketing copy and onboarding
templates. This PR does exactly that and nothing else.

## Related Issue

No existing issue. Happy to open one first if the project prefers that order.

## Type of Change

## Changes Made

**Translations**

- `packages/views/locales/fr/*.json` — 25 new resource files, at full parity
  with the English source.

**Registration**

- `packages/core/i18n/types.ts` — `fr` added to `SupportedLocale` and
  `SUPPORTED_LOCALES`.
- `packages/views/locales/index.ts` — 25 imports and the `fr` block in
  `RESOURCES`.
- `packages/views/settings/components/preferences-tab.tsx` — French entry in
  the language picker.
- `packages/views/locales/{en,zh-Hans,ko,ja,fr}/settings.json` — one new key
  each (`preferences.language.french`). The existing block names every language
  by its endonym (`English`, `中文`, `한국어`, `日本語`), identical across all
  locales, so the new key is `"Français"` in all five.
- `apps/web/app/layout.tsx` and `apps/desktop/src/renderer/src/App.tsx` —
  `fr: "fr-FR"` in `HTML_LANG`.

**Content that is not UI strings**

- `packages/views/workspace/celestial-workspace-names.ts` — French entries for
  the 100 celestial names. 54 have an established French form
  (`Andromeda` → `Andromède`, `Iapetus` → `Japet`); the remaining 46 are proper
  nouns with no French form and are left unchanged.
- `apps/web/features/landing/i18n/types.ts` — adds the `FR` label to
  `localeLabels`, but deliberately leaves the public `locales` array at four.
  The landing copy is not translated, so exposing a French switcher there would
  serve English text under a French flag.
- `apps/web/lib/use-cases-i18n.ts` — the six use-case strings, which do live in
  this map.
- `packages/views/onboarding/templates/index.ts` — `fr` maps to `en` in
  `CONTENT_LANG_BY_LOCALE`, matching the existing fallback for locales without
  authored templates.

**Tests**

- `packages/core/i18n/pick-locale.test.ts` — this suite used `fr` as its
  example of an *unsupported* locale in five assertions, which would now
  silently assert the wrong thing. Those now use `es`, and a new case asserts
  that `fr-FR` resolves to `fr`.

## How to Test

1. `cd packages/views && npx vitest run locales/parity.test.ts` — 212 pass.
   This is the real gate: it fails on any missing key, extra key, placeholder
   mismatch, or plural-suffix mismatch against the English source.
2. `cd packages/core && npx vitest run` — 1457 pass.
3. `cd packages/views && npx vitest run` — 4334 pass (see the note below for
   the one unrelated failure).
4. In the app: Settings → Preferences → Language → Français.

## Translation conventions

Terminology follows `apps/docs/content/docs/developers/conventions.mdx`.
Product nouns that the glossary keeps as proper nouns stay untranslated
(`Multica`, `Autopilot`, `Skill`, `Squad`, `Inbox` where it is the product
surface). The UI uses the formal register (*vouvoiement*) consistently — no
mixed second person anywhere in the 25 files.

French is the first locale in this repo whose CLDR plural set includes `many`
(`Intl.PluralRules("fr").pluralCategories` → `["many", "one", "other"]`). No
`_many` key was needed for the current string set, so `normalizePlural` is
untouched; that is worth knowing if French strings later need one.

## Note: one pre-existing test failure, unrelated to this PR

`packages/views/issues/components/thread-nav-panel.test.tsx > formatStamp >
always shows a date under earlier...` fails on a machine whose system locale is
not English, and passes under `LANG=en_US.UTF-8`. `formatStamp` calls
`toLocaleDateString(undefined, ...)`, so it picks up the host locale and
produces e.g. `3 août`, while the assertion requires `[A-Za-z]{3}`.

This is independent of this PR — the file was last touched in #6426 and is not
modified here — but it will bite any contributor running a non-English system,
which adding locales makes marginally more likely. Flagging it rather than
fixing it in an i18n PR; happy to send a separate one.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs** — n/a, but see the deliberate landing decision above
- [x] If this PR touches Chinese product copy — the only change to `zh-Hans` is the endonym key above, which is not Chinese copy

The two unchecked boxes, honestly: no screenshots yet (the change is a language
option plus string files — happy to add them if useful), and no documentation
update, since I could not find a doc page that enumerates supported locales.
Point me at one and I will update it.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** The translation was produced with Claude Code and then
put through mechanical checks rather than trusted on read-through, because a
4,900-key pass is exactly where an LLM drifts. The checks that ran, and what
each catches:

- `parity.test.ts` — missing/extra keys, `{{placeholder}}` mismatches, plural
  suffix mismatches.
- A terminology pass — every distinct English source string mapped to its
  French renderings, flagging any English term that acquired two different
  French translations across files. Genuine context-dependent splits were kept
  and reviewed by hand; accidental drift was unified.
- An accent pass — grep for the unaccented ASCII form of common French words,
  since a correctly accented word never matches, so every hit is a real
  omission. Words whose unaccented form is also valid French (`termine`,
  `cree`, `assigne`, `donnee`) were excluded to keep it signal-only.
- A register pass — Unicode-aware word-boundary detection of *tu*/*vous* forms
  per file, to prove no file mixes the two.


forrestchang:
Adding French through the existing locale system makes sense, but I’d request changes before merging:

1. **Register French on the backend as well.** The language picker sends `PATCH /api/me` with `{ "language": "fr" }`, but `supportedLanguages` in `server/internal/handler/auth.go` still only accepts `en`, `zh-Hans`, `ko`, and `ja`. The request therefore returns `400 unsupported language`. If the account already has a saved language, `UserLocaleSync` restores that previous preference after reload, so the French selection does not stick. Please add `fr` to the backend allowlist and cover saving, reading back, and retaining the preference after reload. The existing picker tests mock the API, so they don’t catch this mismatch.

2. **Use one consistent translation for Issue.** Navigation and issue creation use `Ticket`, while the billing limits call the same object `Tâche`. Our translation conventions specify the everyday word for “task” in non-English locales and distinguish it from a Run. Please use `tâche` consistently for Issue and keep Run distinct as `exécution`, including notification and activity text. See `apps/docs/content/docs/developers/conventions.mdx`, under “issue is the product’s task.”

3. **Sync with current main and revalidate the French resources.** The PR currently has merge conflicts, and main has English keys that this French bundle does not yet cover. Please resolve the conflicts and rerun locale parity against the updated source.

At the reviewed head (`17a730ad6`), the 255 targeted tests and core/views typechecks passed, and the French keys and interpolation parameters matched that branch’s English resources. This was code and automated-test review; I haven’t verified the running UI or the backend database round-trip.

forrestchang:
I pushed two follow-up commits to this branch, preserving the original commits:

- `e9c8627d9`: merged main at `d18cec01c` and carried French into the extracted HTML language map.
- `3c3052941`: completed the integration and refreshed the translations.

The backend now accepts `fr`. A database-backed regression test first reproduced the `400 unsupported language` response, then passed after the fix, including reading French back through a fresh `GET /api/me`. Frontend tests cover selecting Français and restoring the saved preference without repeated reloads.

Issue is now consistently `Tâche`, with `Exécution` reserved for Run; both terminology guides were updated. The French resources were synchronized with main, including changed source text and removal of obsolete keys. I also added French to the standalone Web error page and the newer failure-reason tests.

Validation passed: 283 targeted frontend tests, 6 backend language tests, core/views/Web/Desktop typechecks, lint on changed TypeScript files, and diff whitespace checks. All 25 resource files match the English bundle's 5,637 keys and interpolation parameters. The temporary API/database environment has been removed.

The PR is mergeable again. These checks do not include visual UI acceptance or a full application build; CI can validate the pushed head.

PR #7156, opened 2026-08-18
MUL-6363: feat(issues): support filtering custom properties by "no value"

### LINKED ISSUE #7153: [Feature]: Filter custom properties by "No value" (unset)
### Problem

Custom issue properties support filtering, sorting, and board grouping. When a
task has no value for a property, the UI already treats that as a first-class
"No value" state — it renders as "未设置" / "No value" in the table, and board
grouping buckets unset issues into a "No value" column (`value_state: "unset"`).

But the **filter menu has no "No value" option** for custom properties, so there
is no way to express "issues where this property is NOT set" — the exact inverse
of picking a value.

### Use case (real, reproduced)

A team uses a `挂起` (Hold) **checkbox** property to park issues that are
"waiting on timing / not actionable right now", while leaving their status
untouched (still `backlog`). The `backlog` column becomes crowded with parked
tasks mixed with actionable ones, and the team wants to see "everything that is
NOT parked" — i.e. filter `挂起 = No value`. That is impossible today:

- `挂起 = true` → only the parked issue.
- `挂起 = false` → only issues *explicitly* stored as `false`; **all unset
  issues are hidden too**. Verified live against the API: in a workspace with
  ~25 unset + 1 `true`, `properties={"<挂起>": ["false"]}` returns 0 issues.

So neither value expresses "not parked", and the team cannot declutter the
backlog without either abusing a status or hiding everything.

### Verified current behavior (main branch)

- The "No value" state already exists at the data/display layer: unset property
  value, renders as `table.no_value`, and board grouping already has a "No
  value" column.
- The filter layer has no "No value" option:
  - Filter UI (`packages/views/issues/components/issues-header.tsx`,
    `PropertyFilterOptions`) only offers the property's configured options (or
    `true`/`false` for checkbox).
  - Backend (`server/internal/handler/property.go`, `parsePropertiesFilterParam`)
    compiles property filters purely as jsonb containment
    (`i.properties @> '{"<defId>": <value>}'`); an unset key never matches.
    `__none__` is handled only for the assignee/project facets, not for custom
    properties.
  - Client-side matching (`packages/views/issues/utils/filter.ts`,
    `issueMatchesPropertyFilters`) likewise returns false when the property is
    `undefined`.

### Proposed solution

Add a **"No value"** option to the custom-property filter menu (at minimum for
`checkbox` and `select`; ideally all types), matching the semantics that already
exist in display/grouping:

- Selecting `No value` filters to issues where the property is unset
  (`value_state: "unset"` / key absent from `properties`).
- Composable with the existing model: values are OR-ed within a property,
  AND-ed across dimensions. E.g. `挂起 = [No value]` → "all non-parked issues".

This delivers the "exclude by value" outcome via positive selection — no
negative operator needed — and matches mature tools (GitHub `no:label`,
Jira/Linear `empty` operators).

### Acceptance criteria

- [ ] Filter menu for a custom property shows a `No value` option alongside the
      configured options.
- [ ] Selecting `No value` returns issues where that property is unset.
- [ ] Works server-side for the board/table query (`/api/issues/table/rows`)
      and the list endpoint (`GET /api/issues`).
- [ ] Round-trips through saved views (`query.propertyFilters`).
- [ ] Malformed-response test added per the API compatibility rules.

### Notes

- No existing issue/PR covers this (searched: "no value filter", "unset
  property", "empty property", "无值", "未设置", "排除", etc.).
- Related but distinct: #6454 (custom board columns), #6862 (relative-age view
  filters).
- The custom-property feature shipped recently (MUL-4463: #5335, #5344) without
  this option.

deoooo:
Closes #7153

## Summary

Adds a **"No value"** option to the custom-property filter. Selecting it matches
issues where the property is unset — the inverse of picking a value — so teams
can express "everything that is NOT X" without a negative operator.

Example: a team uses a `挂起` (Hold) checkbox to park issues while keeping their
status untouched. `挂起 = No value` now returns every non-parked issue, which
`挂起 = false` could not (that only matched issues *explicitly* stored as
`false`, hiding all unset issues — and a workspace with zero explicit `false`
values returned nothing).

## What changed

**Backend**
- `parsePropertiesFilterParam` (`server/internal/handler/property.go`) compiles
  the `__none__` sentinel into a `{"__none__": "<defId>"}` marker instead of a
  jsonb containment pattern.
- `propertiesFilterPredicate` renders the marker as a key-absence disjunction
  (`NOT (i.properties ? $n)`), OR-ed with any real values in the same
  definition and AND-ed across definitions.
- The static `ListOpenIssues` unroll recognizes the same marker, so the
  `open_only` path honors the filter too.
- Property table facets report an unset `__none__` bucket, so the "No value"
  option carries a real count in the filter menu.

**Frontend**
- `issueMatchesPropertyFilters` matches unset issues when `__none__` is
  selected (`packages/views/issues/utils/filter.ts`).
- `PropertyFilterOptions` offers a "No value" option for every property type;
  the filter chips bar renders it via the new `no_value_label` i18n key
  (added for en / zh-Hans / ja / ko).

## Verification

- Backend: new handler + unit tests for the `__none__` sentinel (list,
  `open_only`, OR/AND semantics, facet bucket); full `internal/handler` suite
  passes.
- Frontend: new `issueMatchesPropertyFilters` cases; full `@multica/views`
  issues suite (76 files / 746 tests) passes; `tsc --noEmit` clean.

## Notes

The `__none__` sentinel cannot collide with a real option id: select option ids
are UUIDs and checkbox uses `"true"`/`"false"`.

### DIFF
--- a/packages/views/issues/components/filter-chips-bar.tsx
+++ b/packages/views/issues/components/filter-chips-bar.tsx
@@ -2,6 +2,7 @@
 
 import { useIssueStatuses } from "@multica/core/issue-statuses/hooks";
 import { useStatusLabel } from "../utils/status-label";
+import { NO_PROPERTY_VALUE } from "../utils/filter";
 import { useMemo, type ReactNode } from "react";
 import {
   CalendarDays,
@@ -457,6 +458,9 @@ function useFilterChips(
     const actorProperty = isActorPropertyType(definition.type);
     const actorValues = actorProperty ? actorFilterValues(selected) : [];
     const optionName = (optionId: string): string | undefined => {
+      if (optionId === NO_PROPERTY_VALUE) {
+        return t(($) => $.pickers.custom_property.no_value_label);
+      }
       if (actorProperty) {
         const ref = parseActorRef(optionId);
         return ref ? actorName({ type: ref.kind, id: ref.id }) : undefined;
--- a/packages/views/issues/components/issues-header.tsx
+++ b/packages/views/issues/components/issues-header.tsx
@@ -116,6 +116,7 @@ import { cn } from "@multica/ui/lib/utils";
 import { PAGE_GUTTER } from "../../layout/page-header";
 import { useT } from "../../i18n";
 import { useStatusOptions } from "../utils/status-options";
+import { NO_PROPERTY_VALUE } from "../utils/filter";
 import { matchesPinyin } from "../../editor/extensions/pinyin-match";
 import { FILTER_ITEM_CLASS, HoverCheck } from "../../common/hover-check";
 import { WorkspaceAgentWorkingChip } from "./workspace-agent-working-chip";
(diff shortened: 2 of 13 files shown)


deoooo:
Hi @Bohan-J, @forrestchang — this PR adds a **No value** option to the custom-property filter (closes #7153), so teams can express "everything that is NOT parked", e.g. exclude a `挂起` (Hold) checkbox. It follows the existing containment filter path: the `__none__` sentinel compiles to a key-absence predicate (list + open_only + table facets), and the frontend filter menu / chips / saved views round-trip it.

All CI checks are green (backend, backend-tests, frontend, frontend-test, frontend-views-test, sqlc-check), and the change is covered by new backend handler + unit tests and frontend filter tests. When you have a chance, could you please review? Thanks!

Bohan-J:
Thanks @deoooo — merging this. Really solid contribution, and a few things stood out during review:

- Compiling `__none__` to `NOT (i.properties ? $n)` means the filter uses **the same predicate the board grouping already uses** for its "No value" column (`issue_table_group.go`). The obvious failure mode here would have been a filter that disagrees with the board — "25 in the No value column, 3 in the filter" — and this avoids it by construction.
- You covered **all three query paths**: the paginated list, the `open_only` static query, and the facet counts. The `open_only` one is easy to miss since it goes through the sqlc static SQL rather than the predicate builder, and missing it would have produced a half-correct feature that's painful to debug.
- The `__none__` facet bucket is a quiet win. A checkbox previously showed `Yes` / `No` with no count next to an empty `No`, giving the user no signal that it would return nothing. Now `No value 25` sits right next to it and the ambiguity mostly disappears — without touching the count-rendering logic at all.
- Your note that the sentinel is exact rather than approximate checks out: `validatePropertyValue` rejects `null` and empty values, and DELETE removes the key outright, so "key absent" really does mean "no value" with no `null` / `""` / `[]` grey area.

Two follow-ups, neither blocking:

1. **i18n key** — `no_value_label` duplicates the existing `pickers.custom_property.none` (identical strings in en/ja/ko; in zh-Hans it added a second wording, `未设置` vs the picker's `无值`). I have the consolidation ready — reuse `none` and align zh-Hans to `未设置` so the picker, table cell, board column and filter all read the same. I couldn't push it to this PR because the branch no longer exists on your fork, so it'll land as a separate commit right after this merges. Nothing for you to do.
2. **Coverage** — `filterableProperties` still only admits select / multi_select / checkbox / actor, so text, number, date and url properties never reach the filter menu and can't use the new option. Extending it also needs a facet branch, since the current `default` returns 422. We want the unified "no value" filter across all property types, so we'll track that separately.

Thanks again — the backend semantics were the hard part here and you got them right.

deoooo:
Thanks @Bohan-J for the review and for landing the i18n consolidation! Following up on the second item, I've filed the coverage extension as #7235 — widening `filterableProperties` to text / number / date / url and adding the missing facet branches, reusing the `__none__` sentinel as-is. Happy to pick it up if it helps.

PR #7167, opened 2026-08-18
MUL-6228: resolve full-UUID issue refs locally instead of a resolver GET

### LINKED ISSUE #7017: [Feature]: full-UUID issue refs pay an extra resolver GET — resolve locally
# [Feature]: full-UUID issue refs pay an extra resolver GET — resolve locally

## Deployment type

self-host

## What do you want and why?

`multica issue <command> <full-uuid>` performs **two** HTTP round-trips where one
would do:

1. `server/cmd/multica/cmd_id_resolver.go` — `resolveIssueRef` sends a full UUID
   through `fetchIssueRef`, i.e. `GET /api/issues/<uuid>` just to learn the id
   it already has.
2. The command itself then fetches/mutates the issue by that same id
   (`cmd_issue.go`: `GET /api/issues/<id>` / `PUT ...`).

A full UUID is self-identifying — the resolver's network call adds latency and
a request but no information. During run bootstrap an agent issues several
`multica issue …` calls in a row, so the waste is paid several times per run.

## Proposed solution

For inputs that match the full-UUID regexp, resolve locally:

```go
if uuidRegexp.MatchString(trimmed) {
    return resolvedID{ID: trimmed, Display: trimmed}, nil
}
```

- Identifier refs (`NTL-123` / `MUL-123`) keep the existing resolve — that is
  where the mapping request is genuinely needed.
- Trade-off: user-facing messages ("Comment added to issue %s.", "Issue %s is
  already in that position.") show the UUID instead of the identifier for
  full-UUID invocations. Commands that already display the issue from their own
  response (e.g. `issue get` table mode) are unaffected.
- Error surface is unchanged: a nonexistent UUID 404s on the command's own
  request instead of the resolver's.

Implemented and tested locally: the resolver test now asserts a full UUID
resolves with **zero** HTTP requests; the full `cmd/multica` package passes.

## Additional context

Follow-up to #4701/#4732 (short-prefix resolution was dropped because of
client-side paging; this addresses the remaining full-UUID round-trip).
Related to #6144 (startup context reduction).

adaiguoguo:
Closes #7017, along the shape proposed there.

`resolveIssueRef` sent a full UUID through `GET /api/issues/<uuid>` just to learn the id it already had. Agents fire several `multica issue …` calls back-to-back during run bootstrap — we count 10+ CLI calls in a typical run's transcript on our self-hosted deployment — so the round-trip is paid several times per run.

Full-UUID refs now short-circuit to a lowercased `resolvedID` with **zero** HTTP requests, pinned by test the same way the short-prefix rejection already is. Identifier refs (`MUL-123`) keep the fetch — that mapping genuinely needs the server.

Accepted consequences and error-surface details:

- Messages built only from `Display` show the UUID rather than the issue key for full-UUID invocations. Commands that render the issue from their own response (`issue get`) are unaffected. `issue reorder` also keeps the issue key in its messages because it already fetches the relevant issues as part of the operation.
- Resource validation moves from the resolver GET to the command's next issue-aware endpoint. Most nonexistent UUIDs therefore 404 on the command's own request, with two narrower differences:
  - `issue create --parent <uuid>` returns the create endpoint's 400 validation error (`ExitValidation`, 5) instead of the resolver's 404 (`ExitNotFound`, 4). The create handler re-validates `parent_issue_id` with `GetIssueInWorkspace` before storing anything.
  - `issue comment add <uuid> --attachment <path>` reaches the attachment endpoint first, which returns 403 `invalid issue_id` (`ExitAuth`, 3) instead of the resolver's 404. The upload handler performs that validation before `Storage.Upload`, so a missing or foreign issue still produces zero uploaded/orphaned files.

Every write path continues to validate the issue server-side; the local shortcut only removes a redundant lookup and does not grant access or persist state on the strength of the UUID alone.

The previous "full UUID still resolves via single GET" test asserted the old contract and is superseded by the zero-request test.

### DIFF
--- a/server/cmd/multica/cmd_id_resolver.go
+++ b/server/cmd/multica/cmd_id_resolver.go
@@ -16,7 +16,9 @@ const minShortIDPrefixLen = 4
 const resolverListPageLimit = 50
 
 type resolvedID struct {
-	ID      string
+	ID string
+	// Display is a best-effort label for human-facing messages. It may equal
+	// ID when resolution does not fetch display metadata from the server.
 	Display string
 }
 
@@ -154,12 +156,11 @@ func resolveIssueRef(ctx context.Context, client *cli.APIClient, input string) (
 		// A full UUID is self-identifying: the resolver GET added a round
 		// trip and no information, and agents pay it several times per run
 		// during bootstrap (GH #7017). Lowercased to the canonical form the
-		// server emits. Two consequences, both accepted there: messages
-		// built from Display show the UUID rather than the issue key for
-		// full-UUID invocations, and a nonexistent UUID now 404s on the
-		// command's own request instead of the resolver's — every write
-		// path re-validates server-side (e.g. parent_issue_id on create),
-		// so nothing is stored on the strength of the ref alone.
+		// server emits. Messages without an issue response may show this UUID
+		// rather than the issue key, and failures now come from the command's
+		// next issue-aware endpoint. Every write path re-validates server-side
+		// (e.g. parent_issue_id on create), so nothing is stored on the strength
+		// of the ref alone.
 		id := strings.ToLower(trimmed)
 		return resolvedID{ID: id, Display: id}, nil
 	}
--- a/server/cmd/multica/cmd_id_resolver.go
+++ b/server/cmd/multica/cmd_id_resolver.go
@@ -151,7 +151,17 @@ func resolveIssueRef(ctx context.Context, client *cli.APIClient, input string) (
 		return fetchIssueRef(ctx, client, trimmed)
 	}
 	if uuidRegexp.MatchString(trimmed) {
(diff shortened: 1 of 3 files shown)


adaiguoguo:
Still green and mergeable after this week's churn.

For whoever picks this up: 38 lines across two files, and the only behaviour change is that a full-UUID ref skips the resolver GET. Writes still re-validate server-side — `issue create --parent` goes through `GetIssueInWorkspace` — so a forged or cross-workspace UUID fails exactly as it did before. The local shortcut only removes a round-trip that was confirming what the request itself confirms anyway.

multica-eve (approved):
Accepted. I pushed the review nits: preserve issue keys in reorder messages, document Display as best-effort, and clarify the two error-surface exceptions in the PR description. The full cmd/multica test package and go vet pass locally.

PR #7169, opened 2026-08-18
MUL-6373 feat(dsh): forward task_contract/issue_id and surface progress/acceptance

muzi131313:
## Summary

The DSH runtime already emits a `task_contract` input channel and `progress` / `acceptance` output frames, but the platform adapter drops them. This PR adds the minimal contract plumbing so the `dsh` backend forwards the structured task contract and `issue_id` on `execute`, and surfaces `progress` and `acceptance` on `Result`.

This is the adapter-layer foundation for the platform integration plan; persistence, UI timeline, and status transitions are intentionally left as follow-up work to keep this change reviewable.

> Note: `ExecOptions.TaskContract` is a reserved channel at this stage. There is no production caller yet — the daemon currently only forwards `task.IssueID`, not an assembled contract. The type and DSH forwarding exist so the contract can be populated once issue/run orchestration starts producing it.

## Changes

- `pkg/agent/agent.go`
  - Add `TaskContractInput`.
  - Add `ExecOptions.IssueID` and `ExecOptions.TaskContract` (reserved, not yet populated).
  - Add `ProgressEvent`, `AcceptanceCheckResult`, and `AcceptanceResult`.
  - Add `Result.Progress` and `Result.Acceptance`.
- `pkg/agent/dsh.go`
  - Add `dshTaskContract` and attach it to `dshExecuteCommand` when `hasTaskContract` is true.
  - Forward `opts.IssueID` as `execute.issue_id`.
  - Parse `progress` frames into `Result.Progress`.
  - Parse `result.acceptance` into `Result.Acceptance` (including per-check error code/message and truncated flag).
  - Preserve already-collected `progress`/`acceptance` in the non-terminal fallback result.
- `internal/daemon/daemon.go`
  - Pass `task.IssueID` into `ExecOptions`.

## Tests

- Extended `TestDshBackendExecuteStreamsProtocol` to assert `progress` and `acceptance` survive the wire.
- Added `TestDshBackendForwardsIssueAndTaskContract` to assert the `execute` command carries `issue_id` and `task_contract`.
- Added `TestDshBackendOmitsTaskContractWhenEmpty` to assert an empty contract omits `task_contract` from `execute`.
- Added `TestDshBackendPreservesProgressOnNonTerminalStream` to assert collected progress survives a stream that ends without a terminal `result`.

Local DSH-specific tests pass:

```text
go test ./pkg/agent -run 'TestDsh|TestNewDshBackend|TestParseDshModelID|TestBuildDshMCPServers|TestDiscoverDshModels' -count=1
```

## Related

- Runtime-side change: https://github.com/multica-ai/dsh-multica-runtime/pull/5


multica-eve (changes requested):
Thanks for adding this DSH integration plumbing. The product direction is useful and the change is backward-compatible, but please address these blockers before merge:

1. `server/pkg/agent/dsh.go:368` drops progress and acceptance data on failure, timeout, cancellation, or stream interruption. The fallback result preserves `state.usage` but not the already-collected `state.progress` / `state.acceptance`, so the most important diagnostic paths silently lose execution context. Please preserve both fields in the fallback result and add regression coverage for a non-success path.

2. The comment above `TaskContractInput` in `server/pkg/agent/agent.go:29` says the daemon already injects the same contract into every other backend's prompt, but no such production injection exists. At present, `ExecOptions.TaskContract` also has no production caller, so DSH receives the issue ID but not an actual task contract. Please update the comment and PR description to state that this is a reserved channel that is not populated yet.

Please also add coverage showing that an empty contract omits `task_contract` from the execute payload; that omission is the sole behavior guarded by `hasTaskContract` and currently has no regression test.

muzi131313:
@multica-eve Thanks for the review. All three blockers are addressed in `cbb7b312f`:

1. `server/pkg/agent/dsh.go` fallback result now preserves the already-collected `Progress` and `Acceptance` on non-terminal paths (failure, timeout, cancellation, stream interruption), alongside `Usage`. Added `TestDshBackendPreservesProgressOnNonTerminalStream` for regression coverage.
2. The comments on `TaskContractInput` and `ExecOptions.TaskContract` in `server/pkg/agent/agent.go`, plus the PR description, now state this is a reserved channel that is not populated by production callers yet.
3. Added `TestDshBackendOmitsTaskContractWhenEmpty` to verify an empty contract omits `task_contract` from the execute payload.

Tests: `go test ./pkg/agent/ -run 'TestDsh|TestNewDshBackend|TestParseDshModelID|TestBuildDshMCPServers|TestDiscoverDshModels' -count=1` passes.

multica-eve (changes requested):
Thanks for addressing the previous review items. I re-reviewed head `960e88b96`; the fallback result, reserved-channel documentation, and regression coverage are now in good shape.

One compatibility blocker remains before merge:

`server/internal/daemon/daemon.go:7106` populates `IssueID` for every issue-backed run, and `server/pkg/agent/dsh.go:287` sends it as `execute.issue_id`. However, the currently installed/older `dsh-multica-runtime` profile uses a strict execute-field allowlist in `src/protocol.ts` and rejects both `issue_id` and `task_contract` as unsupported fields. It returns `INVALID_COMMAND`, so deploying this platform change before every profile is upgraded will break previously working issue-backed DSH runs.

The companion runtime PR (multica-ai/dsh-multica-runtime#5) is still open, and merging it alone does not update already installed profiles. Both old and new profiles also advertise protocol version 1, while `server/internal/daemon/agents_probe.go:285` only checks that version, so the daemon cannot distinguish them.

Please make the wire change safe across mixed versions. The preferred option is to wait for and parse `ready.capabilities`, then send `issue_id` / `task_contract` only when the runtime advertises support. Alternatively, introduce a distinguishable protocol/profile version with an explicit, verifiable upgrade and rollback path. Treating these JSON fields as optional is not sufficient because the existing receiver rejects unknown fields.

muzi131313:
Addressed the mixed-version blocker:

- `dsh.go` now waits for the runtime's `ready` frame and reads `capabilities` before sending `execute`.
- `issue_id` is sent only when `capabilities.issue_id` is true; `task_contract` only when `capabilities.task_contract` is true.
- Old profiles that don't advertise these receive a clean execute payload with neither field, so they won't return `INVALID_COMMAND`.
- Added `TestDshBackendOmitsIssueAndTaskContractWhenNotAdvertised`; updated the forwarding test to advertise both.

Companion runtime PR multica-ai/dsh-multica-runtime#5 now advertises `issue_id: true` and `task_contract: true` in its `ready.capabilities`.

multica-eve (changes requested):
Thanks for adding capability negotiation. I re-reviewed head `3a9330f97`; gating `issue_id` and `task_contract` on `ready.capabilities` resolves the previous mixed-version blocker, and the old-profile regression test covers that behavior.

Two startup-path blockers remain:

1. The pre-ready cancellation/exit paths do not share the normal process-tree cleanup. `server/pkg/agent/dsh.go:324-415` starts the stdout reader and waits for `readyCh`, but the process-group cancellation goroutine is not started until after execute is sent at line 437. If `runCtx` ends before ready, lines 406-409 only close stdin and call `cancel()` while `cmd.Cancel` is a no-op. `os/exec` may eventually kill the direct process via `WaitDelay`, but it does not terminate the process group; descendants that inherited stdout can keep the scanner blocked and survive the task. There is also a diagnostic loss: if the runtime exits before ready, the reader builds the real protocol/stderr result and then its deferred `cancel()` makes `Execute` return the generic “cancelled before ... ready” error instead. Please make the handshake observe ready, process exit, and external cancellation as distinct outcomes, route every pre-ready failure through the same process-group termination/wait path, and add regressions for cancellation without an orphaned process tree and for preserving an early startup error.

2. DSH currently consumes the Codex-specific timeout configuration. `dsh.go:397` reads `opts.HandshakeTimeout`, while `server/internal/daemon/daemon.go:7083` unconditionally fills that option from `d.cfg.CodexHandshakeTimeout`. In production this bypasses DSH's 10-second default, uses Codex's 30-second default, and makes `MULTICA_CODEX_HANDSHAKE_TIMEOUT` unexpectedly change DSH behavior. Please wire a provider-specific DSH timeout (or otherwise prevent the Codex value from reaching DSH), then update the `ExecOptions` comment and add coverage for the selected DSH timeout.

muzi131313:
Both startup-path blockers from this review are addressed in head `aa1d9f227`.

**1. Pre-ready cleanup now shares the process-tree termination, and diagnostics survive.**

The handshake `select` in `server/pkg/agent/dsh.go` now observes three distinct outcomes:

- `readyCh` — proceed to send execute (unchanged).
- `resCh` — the runtime exited before ready: return the reader's real result verbatim (exit status, protocol error, stderr tail) via `dshStartupError`, instead of the generic cancellation message. For the `runCtx.Done()` race with a real crash, a non-blocking check of `resCh` lets the exit diagnostic win over "cancelled".
- `runCtx.Done()` / handshake timer — external cancellation or timeout: both now route through `terminateDshStartupTree`, which SIGKILLs the whole process group, closes both pipes (so the scanner unblocks even if descendants hold stdout), and waits on `procDone` for the fully reaped tree before `Execute` returns. No pre-ready path returns while descendants can still be alive.

The reader's fallback classification was also reordered (`exitErr` before the generic not-ready message) so a crash-and-exit keeps its real cause. The `writeFrame` failure path uses the same termination routine.

Regressions added:
- `TestDshBackendPreservesStartupErrorBeforeReady` — a runtime that exits 3 with stderr keeps `exit status 3` and the stderr tail in the returned error.
- `TestDshBackendStartupCancellationKillsProcessTree` — cancels the context pre-ready and verifies the fixture's spawned `sleep 30` child is actually reaped (no orphaned process tree).

**2. DSH has its own handshake timeout knob.**

- New `DefaultDshHandshakeTimeout = 10s` in `internal/daemon/config.go`, resolved from `MULTICA_DSH_HANDSHAKE_TIMEOUT` env / `--dsh-handshake-timeout` flag / `dsh_handshake_timeout` config key (same precedence chain and persist-once semantics as the Codex knob).
- `daemon.go` now fills `ExecOptions.HandshakeTimeout` provider-scoped: `DshHandshakeTimeout` for `provider == "dsh"`, `CodexHandshakeTimeout` otherwise — `MULTICA_CODEX_HANDSHAKE_TIMEOUT` no longer reaches DSH runs, and the DSH default is no longer the Codex 30s.
- `ExecOptions.HandshakeTimeout` comment updated to name both consumers and the provider-scoped daemon wiring.
- Tests: `TestLoadConfig_DshHandshakeTimeout` pins default/env/zero/override resolution plus CODEX-env isolation; `TestDshBackendHandshakeTimeoutUsesProvidedValue` verifies the backend consumes the configured value (300ms timeout fires well under the 10s default) and names it in the error.

Full `pkg/agent`, `internal/daemon`, `internal/cli`, and `cmd/multica` suites pass locally.

multica-eve (changes requested):
Thanks for separating the DSH handshake timeout and for preserving ordinary pre-ready exit diagnostics. The provider-specific configuration blocker is resolved, but one startup cleanup path still leaks descendants.

`server/pkg/agent/dsh.go:411-421` returns directly from both `resCh` branches without calling `terminateDshStartupTree`. If the DSH parent exits before `ready` after spawning a background child whose stdin/stdout/stderr are redirected, the reader reaches EOF, `cmd.Wait()` returns the parent's real exit status, and `Execute` returns that diagnostic — but the remaining process group is never terminated and the child survives as an orphan.

I reproduced this with a temporary Unix regression fixture that ran:

```sh
sleep 30 </dev/null >/dev/null 2>&1 &
echo $! > "$pid_file"
exit 3
```

`Execute` correctly returned `exit status 3`, while the recorded `sleep` PID was still alive afterward. Please route the direct `resCh` case and the nested `resCh`/`runCtx.Done()` race case through the same remaining-process-group cleanup before returning the diagnostic, and add this early-exit-with-detached-child regression.

Also, `TestDshBackendStartupCancellationKillsProcessTree` currently only defers `cancel()` until after the synchronous `Execute` call; it actually reaches `runCtx.Done()` through the five-second `ExecOptions.Timeout`. Please trigger `cancel()` after the fixture reports its child PID so the test genuinely covers external context cancellation.

muzi131313:
Addressed in head `fc07fb02e`.

**resCh paths now reap the remaining process group.** Both direct `resCh` branches in the handshake select call `reapDshStartupTree(cmd)` before returning the diagnostic. By the time a result lands on `resCh` the reader has already finished `cmd.Wait()` (`procDone` closed), so this path doesn't need pipe teardown or a Wait race — it SIGKILLs the surviving group and waits for it to be gone (with one re-SIGKILL after the grace window, since nothing is left to honor a polite escalation):

```go
func reapDshStartupTree(cmd *exec.Cmd) {
	signalProcessGroup(cmd, syscall.SIGKILL)
	waitProcessGroupGone(cmd, dshTerminateGrace)
	signalProcessGroup(cmd, syscall.SIGKILL)
}
```

**Regression added** — `TestDshBackendEarlyExitReapsDetachedChild` uses your exact fixture shape (`sleep 30 </dev/null >/dev/null 2>&1 &` then `exit 3`) and asserts both that `Execute` keeps the `exit status 3` diagnostic and that the recorded detached sleep pid is reaped afterward. Before the fix it reproduces the leak; after, it passes.

**Cancellation test now genuinely exercises external cancel.** `TestDshBackendStartupCancellationKillsProcessTree` runs `Execute` in a goroutine, waits for the fixture to record its child pid (and checks the marker file), then calls `cancel()` explicitly — the `Timeout` is widened to a 30s backstop so the run-context deadline is no longer what fires. Added shared `waitForPidFile` / `assertProcessGone` helpers for both tests.

Full DSH suite passes on darwin; `GOOS=windows go vet` confirms the package (tests included) still compiles on Windows.

multica-eve (changes requested):
Thanks for fixing the Unix direct-`resCh` cleanup path and for changing the cancellation regression to call `cancel()` explicitly. Two blockers remain before merge:

1. **The process tree is still not owned on Windows.** `server/pkg/agent/dsh.go:295` launches DSH with plain `cmd.Start()`. On Windows, `configureProcessGroup` is intentionally a no-op; `server/pkg/agent/proc_windows.go:45-49` states that a backend must use `startOwnedProcessTree` to place the child in a Job Object before it can spawn descendants. Without that ownership, `signalProcessGroup` falls back to killing only the direct process and `waitProcessGroupGone` immediately reports false. On the direct-`resCh` path the parent has already exited, so `reapDshStartupTree` cannot reach a surviving child at all. Please use the established owned-tree launch pattern (including exactly-once `releaseProcessGroup` after the lifecycle finishes) and add Windows process-tree regression coverage; cross-compilation alone cannot validate cleanup behavior.

2. **The updated cancellation test is flaky.** The fixture in `server/pkg/agent/dsh_unix_test.go:67-68` writes the PID file before the marker, while the test at lines 90-92 waits only for the PID and then performs a single `os.Stat(marker)`. There is a valid scheduling window between those writes. I reproduced this with:

```text
go test ./pkg/agent -run 'TestDshBackend(EarlyExitReapsDetachedChild|StartupCancellationKillsProcessTree)' -count=10
```

which failed with `fixture marker ... missing`. Please poll the marker to a deadline, or remove the redundant marker assertion because the PID file already proves the child was spawned.

The single-run DSH tests, DSH race tests, `ByteIdentical`, vet (including Windows cross-vet), build, and diff checks otherwise pass.

muzi131313:
Both blockers addressed in head `735602a9a`.

**1. The process tree is now owned on Windows.**

`Execute` launches the runtime via `startOwnedProcessTree` (the codex/mcode/dim pattern: the child is created suspended, assigned to the Job Object, then resumed — so membership covers every descendant it goes on to create). With ownership in place, `signalProcessGroup` terminates the job rather than falling back to the direct child, and `waitProcessGroupGone` reports real accounting, so `reapDshStartupTree` on the direct-`resCh` path can actually reach a surviving descendant.

`releaseProcessGroup` runs exactly once per launch, in the reader goroutine right after `cmd.Wait()` returns and `procDone` closes — the only safe point, since on Windows releasing closes the job and kills anything still inside it (which is what should happen to stragglers that outlived the reap). It is a no-op on Unix. Every pre-ready early return still funnels through the reader's completion, so ownership is always dropped; there is no path that returns while the job handle is still held.

**Windows regression coverage added** — `dsh_windows_test.go` builds a fake Go runtime that reproduces the Unix fixture shape: it spawns a detached descendant (`cmd.exe /c ping`, stdio redirected to NUL), records its pid, then `os.Exit(3)` before ever emitting `ready`. The test asserts (a) `Execute` returns the real `exit status 3` diagnostic and (b) the recorded descendant pid is dead within the deadline — the assertion that fails without job ownership. A `taskkill /T /F` cleanup guards against a leaked tree on assertion failure. This joins `TestCodexWindowsDescendantsDieWithTheOwnedProcessTree` in the Windows-only suite (cross-compilation still can't *run* these; they execute on the `windows-execenv` runner, which compiles and runs the agent launcher tests — the new test rides the same package).

**2. The cancellation regression is deflaked.**

The redundant marker assertion is gone, as suggested: the fixture no longer writes a marker at all, and the pid file alone is the spawn signal — there is no ordering window left to race. Verified with your exact reproduction:

```
go test ./pkg/agent -run 'TestDshBackend(EarlyExitReapsDetachedChild|StartupCancellationKillsProcessTree)' -count=10   → ok
```

Also re-ran the full DSH suite single-run and under `-race` — all green. `GOOS=windows go vet` compiles the package and its tests, new fixture included.

multica-eve (changes requested):
Thanks for fixing the Windows process-tree ownership and the Unix cancellation-test race. I re-reviewed head `735602a9ab`; the implementation now uses the established Job Object launch/release lifecycle, and the two Unix startup regressions pass repeatedly.

One validation blocker remains before merge:

`server/pkg/agent/dsh_windows_test.go:26` adds the requested real-Windows regression, but `.github/workflows/ci.yml:639` still runs `windows-execenv` with an exact `-run` list that does not include `TestDshWindowsStartupFailureKillsProcessTree`. The current green job log confirms that it ran only the five existing process-tree tests. The Linux backend job excludes this file via `//go:build windows`, so the new test has only been compiled, not executed.

Please add `TestDshWindowsStartupFailureKillsProcessTree` to the Windows job's pattern (or add a dedicated step) and let it pass on a real Windows runner. Cross-compilation cannot validate Job Object cleanup behavior, which was the remaining platform-specific requirement from the previous review.

muzi131313:
Addressed in head `d59d00c3e`.

`TestDshWindowsStartupFailureKillsProcessTree` is now part of the `Test Windows agent process-tree ownership` step's `-run` pattern in `windows-execenv`, so it executes on the real Windows runner instead of only being cross-compiled. The step comment notes what the DSH entry adds: the same Job Object whole-tree guarantee, extended to the backend's own startup path (detached descendant spawned before `ready` dies with the owned tree).

CI on this head is the actual validation — the run linked from the checks below shows the test executing and passing on `windows-latest`.

muzi131313:
Status update on the CI failure: it's resolved and all checks are green now.

Root cause: during the rebase to the latest main, the Execute path ended up sending the execute command twice — once leftover from before the ready handshake and once after. The fake runtime consumed the first command and exited, so the second write hit a broken pipe, and the capability-asserting fixtures (exit status 7 for missing issue_id) failed accordingly.
Fix: the duplicated send was removed (head 9985cfa2b, handled by the codex agent in this thread).
Verified: backend-tests, backend, windows-execenv, and all other checks pass on the current head (run 32937707944).
The PR is back to waiting on multica-eve's next review round.

multica-eve:
Thank you for the care and patience you put into this contribution. We also owe you an apology: we asked you to revise it through several review rounds, and you responded thoughtfully each time, but we should have made the product-scope decision much earlier instead of letting the review continue this far.

After stepping back and reconsidering the complete user outcome, we have decided not to ship this change and will close the PR. DSH users can already complete tasks and have their normal execution history recorded. The new structured task contract, however, currently has no production source, while the new progress and acceptance results are not persisted, displayed, or used to drive any follow-up action. This means the PR does not yet complete a user journey on its own.

Merging it now would make us maintain a new protocol and configuration surface, along with stricter DSH startup and cleanup behavior, before we have committed to building the rest of that end-to-end experience. We would rather revisit the capability when the producer and consumers are planned together than merge plumbing based only on possible future use.

This decision is not a negative judgment on the quality of your latest implementation: you addressed the compatibility, startup cleanup, cross-platform ownership, and validation feedback carefully, and the current head is technically solid. Any independently valuable process-tree cleanup can be reconsidered by us later as a narrower bug fix; we are not asking you to do more work on this PR.

Thank you again for the substantial effort, and sorry again that our product decision came only after you had already invested in several rounds of changes.

PR #7173, opened 2026-08-18
fix(mobile): configure Apple team from environment (MUL-6375)

### LINKED ISSUE #7172: feat(mobile): support dynamic EXPO_APPLE_TEAM_ID mapping in app.config.ts
### Summary
Currently, `apps/mobile/app.config.ts` allows overriding the iOS `bundleIdentifier` for production and development via environment variables (`EXPO_BUNDLE_IDENTIFIER_PROD` and `EXPO_BUNDLE_IDENTIFIER_DEV`), enabling external contributors and self-builders to build personal or organization copies of the app.

However, `app.config.ts` does not map `ios.developmentTeam` to an environment variable (`EXPO_APPLE_TEAM_ID`). As a result, when developers with multiple Apple Developer teams build the app locally, Expo CLI auto-detects or locks onto the first team found in Keychain, causing code signing and provisioning profile errors.

### Proposed Change
Add `developmentTeam: process.env.EXPO_APPLE_TEAM_ID` to the `ios` configuration block in `apps/mobile/app.config.ts`:

```ts
ios: {
  supportsTablet: false,
  developmentTeam: process.env.EXPO_APPLE_TEAM_ID,
  bundleIdentifier: ...
}
```

### Benefits
- Allows developers to specify their Apple Developer Team ID via `EXPO_APPLE_TEAM_ID` in environment files or shell env.
- Non-breaking: evaluates to `undefined` when `EXPO_APPLE_TEAM_ID` is unset, preserving Expo's default auto-detection behavior.

robertn702:
## Summary
- map `EXPO_APPLE_TEAM_ID` to Expo's supported `ios.appleTeamId` configuration field
- allow self-builders with multiple Apple Developer teams to select the correct signing team during Expo prebuild
- preserve Expo's normal behavior when the variable is unset

## Verification
- `pnpm --filter @multica/mobile typecheck`
- `git diff --check`

Closes #7172

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -32,6 +32,7 @@ export default ({ config }: ConfigContext): ExpoConfig => {
     icon: "./assets/icon.png",
     ios: {
       supportsTablet: false,
+      appleTeamId: process.env.EXPO_APPLE_TEAM_ID,
       // Per-variant bundle id overrides exist for one reason: an Apple ID
       // can only sign bundle prefixes it owns, so contributors not on the
       // Multica Apple Developer team (and external users self-building a


Bohan-J:
Merged — thanks for this, and especially for not just implementing the issue as written.

The issue proposed `ios.developmentTeam`, which isn't a real Expo config field. You used `ios.appleTeamId`, which is the one that actually flows through `@expo/prebuild-config`'s default plugin chain into `DEVELOPMENT_TEAM` on every native target. Picking the environment variable name `EXPO_APPLE_TEAM_ID` also lines up with what `eas-cli` already uses for Apple team resolution, so nothing new is invented here.

One consequence of the change that's nicer than the issue claimed, for anyone who finds this later: because `withDevelopmentTeam` only writes `DEVELOPMENT_TEAM` when a value is present, a developer whose first build got auto-signed with the wrong team stays stuck on it — the choice is written into the generated `ios/` and never cleared. Since `scripts/ios-run.sh` re-runs `expo prebuild` on every invocation, setting the variable re-applies the intended team each run, so it repairs an already-mispinned checkout rather than only helping on a clean one.

The only thing missing was documentation, and that's on us rather than something to send back to you — #7196 adds it to the README, `.env.example`, and a comment at the call site, matching how the `EXPO_BUNDLE_IDENTIFIER_*` overrides are documented.

PR #7175, opened 2026-08-18
fix(skills): skip binary supporting files in runtime-local skill sync

### LINKED ISSUE #7143: [Bug]: Runtime-local Skills corrupt binary supporting files such as .docx
### Deployment type

self-host

### What happened?

The local Skill collector reads every supporting file and converts its bytes to a Go string. The content is transported as JSON and then written back into the task environment from that string.
Binary Office files contain non-UTF-8 bytes. Those bytes are replaced during the text/JSON round trip, so the recreated document differs from the original and can no longer be opened.
Relevant code paths:
server/internal/daemon/local_skills.go — collectLocalSkillFiles(..., includeContent=true) uses string(content) without checking for binary data.
server/internal/daemon/execenv/context.go — writeSkillFiles recreates files with []byte(f.Content).
The archive-import path already identifies .doc, .docx, .xlsx, .pptx, and .pdf as binary assets and skips them. Runtime-local Skill synchronization does not apply the same protection.

### Steps to reproduce

1. Create or install a local Skill containing SKILL.md.
2. Add a binary supporting file, for example references/example.docx.
3. 3Let the Multica daemon discover or synchronize this runtime-local Skill.
4. Assign a task to an agent with this Skill.
5. Compare the original file with the generated file in the prepared task environment, or open the generated file in Word/LibreOffice.

### Screenshots (optional)

_No response_

```shell
The installed lark-doc Skill itself contains Markdown reference files only. The issue occurs when a Word document is added as a supporting file in a runtime-local Skill bundle.
```

ADanMan:
## Summary

`collectLocalSkillFiles` (`server/internal/daemon/local_skills.go`) read every supporting file into `SkillFileData.Content` as a Go string. That string crosses the daemon/server hop as JSON, and `encoding/json` rewrites each invalid UTF-8 byte to U+FFFD — so `writeSkillFiles` (`server/internal/daemon/execenv/context.go`) recreated a file whose bytes differ from the original. A `.docx`, `.xlsx` or `.pdf` carried in a runtime-local skill bundle therefore arrived in the prepared task environment corrupt and would not open (#7143).

The archive/URL importer already refuses this class of file for the mirror-image reason — those bytes cannot go into a PG TEXT column (SQLSTATE 22021) — but runtime-local sync never got the same protection.

## Change

Give both ingestion paths one shared predicate, and apply it on the runtime-local path.

- `server/internal/skill/binary.go` (new): the importer's extension blacklist becomes `IsLikelyBinaryFilePath`. The package is already imported by both the handler (`skillpkg`) and the daemon, so this is a move, not a new abstraction — the list itself is unchanged.
- `server/internal/handler/skill.go`: drops its local copy and calls the shared predicate from both existing call sites. No behaviour change.
- `server/internal/daemon/local_skills.go`: `collectLocalSkillFiles` now skips binary supporting files, with a log breadcrumb matching the importer's.

The skip deliberately does **not** depend on `includeContent`: discovery reports the bundle with content omitted and sync uploads it with content included, so skipping in only one pass would leave the two disagreeing about which files make up the bundle. The skip also sits before the file-count and bundle-size accounting, so a large binary no longer consumes a bundle's byte budget on its way to being dropped.

Skipping rather than base64-encoding through the wire is the parity fix the issue asks for; it keeps the `SkillFileData` contract between installed daemons and the server untouched.

## Test plan

- `server/internal/skill/binary_test.go`: table test over the predicate — the Office formats from the report, other binary payloads, upper-cased extensions (Windows skill dirs commonly carry `.DOCX`), and the text payloads that must still pass through.
- `server/internal/daemon/local_skills_test.go`:
  - `TestCollectLocalSkillFiles_SkipsBinarySupportingFiles` — a `.docx` holding real invalid UTF-8 bytes plus a `.png` are dropped, the `.md` survives byte-for-byte. Verified it fails before the change (`collected paths = [assets/logo.png references/notes.md references/report.docx]`) and passes after, so it is a real regression guard.
  - `TestCollectLocalSkillFiles_BinarySkipIsConsistentAcrossPasses` — pins the `includeContent`-independence above.
- `go test ./internal/skill ./internal/daemon ./internal/daemon/execenv ./internal/handler` ✅ · `go vet` ✅ · `gofmt` clean · `go build ./...` ✅

Fixes #7143

### DIFF
--- a/server/internal/daemon/local_skills.go
+++ b/server/internal/daemon/local_skills.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/multica-ai/multica/server/internal/skill"
 	"github.com/multica-ai/multica/server/pkg/agent"
@@ -376,15 +377,34 @@ func collectLocalSkillFiles(skillDir string, includeContent bool) ([]SkillFileDa
 		// A binary supporting file cannot survive SkillFileData.Content: the
 		// bytes go out as a Go string and encoding/json rewrites every invalid
 		// UTF-8 byte to U+FFFD, so writeSkillFiles later recreates a file that
-		// differs from the original and will not open. Skip it the way the
-		// archive/URL importer already does, and skip it regardless of
-		// includeContent so the discovery and sync passes agree on which files
-		// make up the bundle.
-		if skill.IsLikelyBinaryFilePath(rel) {
+		// differs from the original and will not open.
+		//
+		// IsLikelyBinaryFilePath is only a cheap first pass on the extension —
+		// the same heuristic the archive/URL importer uses. On its own it
+		// misses a binary file with an unlisted or missing extension (a
+		// .safetensors, a .parquet, a stray no-extension blob) and a text file
+		// in a non-UTF-8 encoding, both of which corrupt exactly the same way.
+		// The read below is the actual guarantee: skip whenever the bytes
+		// themselves are not valid UTF-8, regardless of what the extension
+		// suggested. This runs on both the includeContent=false (discovery)
+		// and includeContent=true (sync) passes so they agree on which files
+		// make up the bundle — skipping the read on the false pass would let
+		// a discovery listing promise a file that sync then silently drops.
+		content, err := os.ReadFile(path)
+		if err != nil {
+			return nil
+		}
+		likelyBinaryExt := skill.IsLikelyBinaryFilePath(rel)
+		if likelyBinaryExt || !utf8.Valid(content) {
+			reason := "invalid_utf8"
+			if likelyBinaryExt {
+				reason = "binary_extension"
+			}
 			slog.Info("local skill: skipping binary file",
 				"skill_dir", skillDir,
 				"path", filepath.ToSlash(rel),
 				"size", info.Size(),
+				"reason", reason,
 			)
 			return nil
 		}
@@ -398,10 +418,6 @@ func collectLocalSkillFiles(skillDir string, includeContent bool) ([]SkillFileDa
 
 		file := SkillFileData{Path: filepath.ToSlash(rel)}
 		if includeContent {
-			content, err := os.ReadFile(path)
-			if err != nil {
-				return nil
-			}
 			file.Content = string(content)
 		}
 		files = append(files, file)
--- a/server/internal/daemon/local_skills.go
+++ b/server/internal/daemon/local_skills.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"bytes"
 	"fmt"
 	"io/fs"
 	"log/slog"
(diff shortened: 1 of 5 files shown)


Bohan-J (changes requested):
Thanks for the clear root-cause writeup — the diagnosis is right, and refusing to hand back bytes we cannot round-trip is the correct direction. Sharing one definition between the two ingestion paths is also the right call, and skipping regardless of `includeContent` so discovery and sync agree on the bundle is a subtle detail worth having a test for.

Requesting changes on one point before merge, plus one question that needs a product decision.

## Extension blacklist alone does not close the root cause

The corruption happens at `local_skills.go`:

```go
file.Content = string(content)
```

Any file whose bytes are not valid UTF-8 gets rewritten to U+FFFD on the JSON hop — the extension is incidental. `IsLikelyBinaryFilePath` only recognises the extensions on the list, so after this change these still produce a silently corrupted copy in the prepared task environment:

- `references/blob.bin`, `data.parquet`, `model.safetensors` — binary, unlisted extension
- a supporting file with no extension at all
- a `.txt` / `.csv` / `.json` that happens to contain invalid UTF-8 (e.g. Latin-1 encoded)

The new tests use `.docx` and `.png`, which are both on the list, so they pass whether or not the underlying defect is fixed.

Suggested shape: keep the extension check as the cheap first pass, then add a content-level fallback in the collection path — read the file and treat `!utf8.Valid(content)` as binary. That covers the unlisted cases and makes the guarantee "what we carry is byte-identical" rather than "what we carry has an extension we recognised". The check has to produce the same answer in both the `includeContent=false` and `includeContent=true` passes, which is the same invariant your second test already protects, so it needs a read in the discovery pass too (or a bounded prefix read, which is enough in practice for this).

Please also add regression cases for: unlisted extension with invalid UTF-8, no extension with invalid UTF-8, and a valid UTF-8 text file that must still pass through untouched.

## The skip is invisible to the person who put the file there

Right now a skipped file leaves an `slog.Info` breadcrumb on the server and nothing else. From the member's side the failure mode changes from "corrupted file" to "file silently missing", and they find out only when the agent behaves as if the reference material does not exist.

That is still better than shipping corruption, so it should not block this PR — but we do owe the user a signal. Two options, and this is a maintainer call rather than something to guess at:

1. **Skip + report** — surface the skipped paths where the skill is shown, so the member can see what did not make it.
2. **Carry the bytes** — change the wire representation to something byte-safe (base64 field alongside the text one) so binary supporting files reach the task environment intact.

If (2) is where we want to land, the extension list stops being load-bearing entirely. Happy to hear your view before you invest in either — the change as scoped here is fine to land once the UTF-8 gap above is closed.

ADanMan:
Thanks — agreed on both points, and I've pushed the fix.

**On the required change.** `IsLikelyBinaryFilePath` stays as the cheap first pass, but the real gate is now content-level: read the file and skip whenever `!utf8.Valid(content)`, regardless of what the extension said. That runs on both the `includeContent=false` (discovery) and `includeContent=true` (sync) passes — discovery now reads the file too, specifically so the two passes keep agreeing on which files make up the bundle. Added regression cases for an unlisted binary extension (`.safetensors`/`.parquet`/`.bin`), a binary file with no extension at all, and a valid-UTF-8 text file to prove the check doesn't over-trigger.

**On the product question (skip vs. carry the bytes).** My view: skip + report is the right near-term shape — surface the dropped paths wherever the skill is shown, no wire change needed. Carrying the bytes through a base64 field is the better end state if binary supporting files turn out to matter in practice, but that changes the `SkillFileData` contract installed daemons already emit, so I'd rather scope that as its own PR with explicit version-skew handling than fold it into this one. Happy to pick that up next if you want it.

multica-eve (dismissed):
Thanks for closing the previous invalid-UTF-8 gap. The new content-level check and the unlisted/no-extension regression cases are the right shape. There is one remaining byte-integrity hole before this is ready:

## Embedded NUL still passes the new gate and is later deleted

`utf8.Valid` returns true for NUL: for example, `[]byte("A\x00B")` is valid UTF-8. An unknown-extension or extensionless supporting file with those bytes therefore passes `server/internal/daemon/local_skills.go:393-421` and is sent to the server.

The runtime-local import then persists the file through `sanitizeNullBytes` at `server/internal/handler/skill_create.go:57-61`. `server/internal/util/text.go:65-84` explicitly removes every `0x00`, so the materialized file is still not byte-identical to the source. UTF-16LE text containing only ASCII characters is a realistic example of this path.

The new `invalidUTF8Bytes` fixture contains `\xff`/`\xfe` as well as NUL, so it is rejected for invalid UTF-8 and does not exercise this case.

Please require content to be both valid UTF-8 and NUL-free (for example, `utf8.Valid(content) && bytes.IndexByte(content, 0) < 0`), give NUL its own skip reason, and add unknown/no-extension fixtures containing valid UTF-8 plus NUL. Pin the same skip result in both discovery and sync passes; ordinary valid UTF-8 text should continue to pass through unchanged.

Non-blocking nit: `IsLikelyBinaryFilePath` is described as the cheap first pass, but the code currently calls `os.ReadFile` before checking it. Checking the extension first would avoid unnecessary reads for known binary types.

ADanMan:
Good catch — pushed the fix.

**Content check now requires both valid UTF-8 and NUL-free**: `!utf8.Valid(content) || bytes.IndexByte(content, 0) >= 0`, with `embedded_nul` as its own skip reason distinct from `invalid_utf8` so the daemon log tells them apart. Added regression cases for an unlisted extension and a no-extension file, each holding a UTF-16LE-style fixture (`"H\x00e\x00l\x00l\x00o\x00"`) — valid UTF-8 by itself, would have slipped past the old check, and is exactly the shape `sanitizeNullBytes` would silently mutate on the way back out.

**Nit taken too**: `IsLikelyBinaryFilePath` now runs before the read, not after — a known-binary extension returns immediately without ever calling `os.ReadFile`. It's a genuine cheap-first-pass now instead of just being described as one.

Bohan-J (approved):
Thanks for sticking with this through three rounds — the PR ended up in a much better place than where it started, and the path it took is worth naming.

The first version keyed on the extension list, which was a reasonable read of how the archive/URL importer already behaved. What makes the final version correct is that it stopped treating the extension as the thing that matters and named the actual invariant: this pipeline can only carry bytes that are **valid UTF-8 and NUL-free**. Everything else follows from that — the extension check demoted to a cheap short-circuit that skips the I/O, the content check as the real guarantee, `invalid_utf8` and `embedded_nul` as separate skip reasons so the logs say which rule fired.

The NUL case in particular was the subtle one, and catching that a valid-UTF-8 file can still be rewritten by `sanitizeNullBytes` before it lands in PG is what took this from "handles the reported `.docx`" to "closes the class". Running the gate on both the discovery and sync passes — so a listing can never promise a file that sync then drops — is the kind of detail that is easy to leave for someone else to find in production, and you have a test pinning it.

Also good: consolidating the two ingestion paths onto one `IsLikelyBinaryFilePath` in `internal/skill`, so the definition can't drift between them again.

One thing I want to flag for whoever picks up the follow-up, not a change request here — from the member's side the failure mode moves from "corrupted file" to "missing file", and the skip is currently only a server-side `slog` line. That is unambiguously the right trade (silently handing back a broken `.docx` is worse), but we still owe the member a visible signal about which supporting files did not make it and why. I'll open a separate issue for that rather than grow this PR.

CI is green. Merging — thanks again.

PR #7176, opened 2026-08-18
fix(issues): copy durable local path for worktree tasks (MUL-6378)

wuchangming:
## What does this PR do?

Fixes **Copy local workdir path** returning a deleted task worktree after a worktree-mode task finishes.

Thinking path:

- `local_directory` tasks run in a daemon-created temporary worktree recorded as the task `work_dir`.
- Task finalization removes that temporary worktree, so the copied path becomes stale.
- The project resource retains the persistent `local_path`.
- Use that durable path only when the latest task runtime belongs to the current local daemon and exactly one valid `local_directory` resource matches. Otherwise preserve the existing task `work_dir` fallback for Web, remote runtimes, missing runtimes, and ambiguous resources.

## Related Issue

N/A — this fixes the worktree lifecycle bug described above.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add a pure core resolver for the durable path and conservative fallback behavior.
- Lazily query runtimes and project resources only for Desktop issues with a local daemon and project.
- Route the issue action through the resolver.
- Add resolver and UI-wiring regression coverage.

## How to Test

1. Run `pnpm --filter @multica/core test` — 1,486 tests pass.
2. Run `pnpm --filter @multica/views test` — 4,335 tests pass.
3. Run typecheck for `@multica/core` and `@multica/views`.
4. Run ESLint on the changed files and `git diff --check`.

## Checklist

## AI Disclosure

**Prompt / approach:** Diagnosed the copied-path lifetime mismatch across the frontend action, task runtime, project resources, and daemon worktree cleanup; implemented a conservative resolver; reviewed the five-file diff; and ran the affected full test suites, typechecks, lint, and diff validation.

## Screenshots

N/A — this changes path resolution behavior without changing the visual UI.


Bohan-J (changes requested):
Thanks for tracking this down — the diagnosis is right, and the fallback ladder in `resolveWorkdirCopyPath` is appropriately conservative for the Web / remote-runtime / unknown-runtime cases. Two things block a merge for me, though, because they turn a *stale* path into a *silently wrong* path, which is harder for a user to notice.

### 1. The resolver ignores task status, so an in-flight worktree task now copies the project root

`WorkdirTask` is `Pick<AgentTask, "created_at" | "runtime_id" | "work_dir">` — `status` is never consulted. But the worktree is only torn down in `Finalize`, on the task's exit path. While the task is still `running`, the worktree exists and it is the **only** place the agent's changes live; the project root has none of them.

So the most common reason to reach for this menu item — "the agent is working right now, let me go look at what it's changing" — regresses from a correct path to a path that looks plausible and contains nothing.

Suggestion: only substitute the durable path when the latest task has reached a terminal state.

### 2. …but a terminal check can't just be `status !== "running"` — the preserved-worktree case needs care

In `runLocalDirectoryTask`, when `Finalize` cannot commit what the agent left, it deliberately **keeps** the worktree and fails the task (`worktreePreservedError`, carrying `preserved_path`). In exactly that situation the worktree still exists, holds the only copy of unsaved work, and is on a GC timer — and the new logic would point the user at the project root instead. That's the moment the real path matters most.

Whatever terminal condition you land on should leave this case pointing at `work_dir`.

### 3. The resolver doesn't verify the task actually ran in that directory

The guard is "latest task's runtime belongs to this daemon" + "exactly one local_directory resource for this daemon". It never checks that *this task* executed against that resource. Two reachable cases where they diverge:

- **Squad leader tasks.** `localDirectoryAssignmentForTask` returns `nil` for `IsLeaderTask`, so a leader run does not bind to the user's directory — it runs in the daemon-managed workdir. On a leader's issue that is typically the only task, so it would get rewritten to a project path it never touched.
- **Tasks that predate the resource.** A task that ran before the `local_directory` resource was attached to the project gets retroactively rewritten too.

In both cases the old behavior at least returned a path the task really used. A prefix check (`work_dir` is at or under `local_path`) would filter most of this cheaply.

### Smaller points

- **`staleTime`.** The existing tasks query uses `staleTime: 30_000`, but `runtimeListOptions` / `projectResourcesOptions` fall back to the default `0`. List and board views wrap every row in the context menu, so every right-click refetches the full workspace runtime list plus the project resources. Worth aligning to 30s.
- **Label vs. behavior.** The item still reads "Copy local workdir path" while it may now hand back the project root, with no signal to the user. Worktree tasks already carry `branch_name` — which the type docs describe as the only durable pointer to what the run produced. A toast along the lines of "Copied project directory — the task's worktree was reclaimed; its branch is `<branch>`" would make the substitution legible instead of silent.
- **Consistency.** The transcript dialog still copies `relative_work_dir` for the same task. After this change the two surfaces disagree about "the task's directory"; worth a follow-up at minimum.
- **Minor.** The `localPaths.length === 1` guard is effectively defensive-only: the list is already filtered by `daemon_id`, and the server enforces one `local_directory` per (project, daemon). No harm in keeping it, but it isn't the cross-daemon ambiguity guard the comment suggests.

### One design question

This puts policy in the client: two extra requests to reconstruct something the daemon already knows, and only Desktop benefits. Would it be cleaner for the daemon to record the durable path (or a "work_dir reclaimed" marker) on the task at finalize, so every client — Web included — gets a consistent answer without re-deriving the rule in the UI? Happy to hear if you already considered and rejected that.

None of the six new resolver cases cover a running task, so whichever direction you take on (1) and (2), please add coverage for it.

wuchangming:
@Bohan-J Thanks — agreed that client-side reconstruction is the wrong boundary here. I had not accounted carefully enough for the preserved-worktree failure path.

I plan to replace the current runtime/resource inference rather than add more status or prefix heuristics:

- Add an optional `durable_work_dir` to the task contract. The daemon will set it only after `LocalWorktree.Finalize` succeeds and confirms the task worktree has actually been removed. A commit/finalize/cleanup failure leaves it unset, so the preserved `work_dir` remains authoritative.
- Carry that field through complete, fail, and cancel-ack, and persist it on the task. Older daemons omit it, which safely preserves today's behavior instead of silently redirecting a path.
- Have clients choose `durable_work_dir` only for a terminal task when the field is present; otherwise they keep `work_dir`. Leader tasks, tasks predating the resource, running tasks, preserved worktrees, and legacy tasks therefore cannot be rewritten.
- Remove the runtime/project-resource queries entirely, which also eliminates the extra refetches and makes Web/Desktop use the same server fact.
- Use the same copy-target helper in the issue menu and transcript. When the durable path is selected, change the label/toast to say that the task worktree was reclaimed and include `branch_name` when available.

I will add coverage for running, successfully reclaimed, failed-but-preserved, cancelled, leader/legacy, terminal callback round-trips, and malformed/missing optional API fields. I also want to tighten `removeLocalWorktreeDir` so Finalize cannot claim success unless the path is actually gone.

Before I expand the PR across daemon/server/UI, does that direction match what you had in mind?

Bohan-J:
Yes — that's the direction I had in mind, and gating the write on "Finalize succeeded **and** the worktree is verifiably gone" is the part that makes it work. It turns the hardest case (preserved worktree holding the only copy of uncommitted work) into an *absent field* rather than something the client has to reason about, and it means leader tasks, pre-resource tasks, and legacy data can't be rewritten by construction. Dropping the runtime/project-resource queries and giving Web and Desktop the same server fact is a strict improvement over what's in the PR today. Please go ahead.

Tightening `removeLocalWorktreeDir` isn't a side quest here — it's the invariant the whole design rests on. Worth making it explicit in the code that `durable_work_dir` is written only on the verified-removed path, so a future refactor can't quietly widen it.

**One thing to fold in before you expand the PR: privacy.**

`work_dir` is treated as sensitive. The server derives a privacy-safe `relative_work_dir` in `taskToResponse` (`handler/agent.go`), and the field docs are explicit that raw `work_dir` must never be rendered — not in a tooltip, `title`, or `aria-label` — because the goal is that screenshots and screen shares stay safe too.

That matters for the shared-helper step, because the transcript dialog **displays** the path, not just copies it. So `durable_work_dir` needs the same treatment: a server-derived display form alongside it, with the absolute path reserved for the clipboard. Otherwise unifying the two surfaces on one helper would leak the user's home directory into a surface that was deliberately kept safe.

**Smaller things while you're in there:**

- **Persistence.** The field needs a migration plus wiring through `taskToResponse` so it reaches clients the same way `work_dir` does.
- **`in_place` mode.** Its `work_dir` is already durable. Either writing the field or leaving it unset is defensible — just make the choice explicit and cover it in tests, so it doesn't become an accident.
- **Snapshot staleness.** The stored path is a point-in-time fact. If the user later renames the directory or detaches the resource, it goes stale again. That's an acceptable trade, but worth a line in the PR description so it doesn't get re-reported as the same bug.

Your test matrix looks right to me; the running and failed-but-preserved cases are the two I most want to see land.

On scope: crossing daemon/server/UI is fine by me. If you'd rather keep the diffs small, landing the contract first (daemon + server field, no behavior change in the UI) and the client switch second would both be reviewable on their own — entirely your call.

wuchangming:
@Bohan-J Implemented the daemon/server/UI direction in `dabfd12`.

- `durable_work_dir` is written only after `Finalize` succeeds and `removeLocalWorktreeDir` verifies the path is absent with `Lstat`.
- Commit/finalize/cleanup failures leave it unset, so running and failed-but-preserved tasks keep their real `work_dir`.
- Complete, fail, cancel-ack, migration/sqlc, replay, and status-CAS persistence paths are covered.
- `taskToResponse` now provides `relative_durable_work_dir`; absolute paths are clipboard-only and both client surfaces share one resolver.
- `in_place` deliberately leaves the new field unset, and that choice is covered explicitly.
- The point-in-time snapshot trade-off is now documented in the PR description.

Local verification passed for the targeted daemon/handler/migration suites, all Go packages compile, the affected Core/Views tests pass (128 + 45), both TypeScript package checks pass, and `git diff --check` is clean. CI is running now. Would appreciate another look once it settles.

Bohan-J (approved):
Approving — thanks for the thorough rework. You took the harder path when the easy one was available, and the result is better than what either of us sketched out.

Two things stood out on the final pass:

- Using `Lstat` rather than `Stat` to confirm the worktree is gone, and saying why in the comment: a broken symlink would otherwise read as "removed" while a stale entry still occupies the handoff path. That's the kind of detail that decides whether the whole invariant actually holds.
- Failing `Finalize` when removal can't be confirmed, so `durable_work_dir` is *absent* rather than wrong. Turning the hardest case into a missing field instead of a client-side judgement call is what makes the rest of this safe.

I also checked the mixed-version and privacy edges: old daemon / old server / old client all degrade to `work_dir` in every direction, migration 362 is an additive nullable column with no FK, cascade, or index, the `COALESCE` writes are replay-idempotent, and the cancel-ack CAS keeps a late ack off a completed or failed row. `relative_durable_work_dir` reuses the same home-stripping helper, and when the durable path *is* the home directory the transcript row is hidden rather than falling back to the absolute path — the right direction to fail in.

Verified locally: `go build ./...` and `go vet` clean, `execenv` finalize/worktree/remove tests and the daemon terminal-report tests pass. CI is green across all 15 checks.

Merging now. Five non-blocking notes, none of which need to hold this up:

1. **Worth a follow-up:** there's no daemon-side test asserting the production invariant itself. `DurableWorkDir` only appears in `client_test.go` as transport coverage — nothing asserts "written only after a confirmed removal", "never written when removal fails", "never written for in-place". That invariant is what the whole design rests on, so it's the thing most worth protecting from a future refactor.
2. The transcript copy button now copies the absolute path instead of `relative_work_dir`, and that applies to **every** task kind, not just worktree runs. I think it's the right call — a relative path was never pasteable, display stays relative so the privacy contract holds, and it now matches the issue menu — but it's a silent behavior change for managed and cloud runs.
3. The bare-error failure path now reports `work_dir` and `branch_name` where it previously reported neither. Also a real improvement — a preserved worktree path finally reaches the server on that path — but it's scope beyond `durable_work_dir`.
4. `taskResult.EnvRoot = env.RootDir` in the finalize defer is dead code: the only consumer is the GC-meta write, which both the error and cancel paths return before reaching, and every success path already sets it. Harmless, but it reads as though it changes GC behavior.
5. The PR description still describes the earlier client-side inference approach. I've written the squash commit message against what the code actually does, so the history stays accurate.

Thanks again — the `durable_work_dir` contract is a better answer than the one this PR started with, and it leaves Web and Desktop working from the same fact.

PR #7186, opened 2026-08-19
feat(dingtalk): support Squad targets

### LINKED ISSUE #7165: [Feature]: Allow DingTalk bots and group routes to target a Squad
### Deployment type

Official App and self-hosted deployments.

### What do you want and why?

Allow a DingTalk robot installation and each discovered DingTalk group route to target a Multica Squad, not only an individual Agent.

Today DingTalk installations, direct-message routes, group routes, Chat sessions, and channel task dispatch all resolve to one `agent_id`. Squads are already first-class assignees elsewhere in Multica, but an owner/admin cannot connect a DingTalk robot from a Squad or select a Squad in Settings → Integrations → DingTalk.

The desired product behavior is not fan-out. A message sent to a Squad-bound robot should go to the Squad's current Leader first. The Leader should run in the Squad leader role with the Squad instructions and roster, then decide whether to answer directly or distribute the work to the appropriate Squad members. Agent selection remains the Leader's responsibility.

This gives teams one stable DingTalk entry point for a coordinated group of Agents while preserving the existing Squad model: one Leader evaluates the request and delegates deliberately instead of running every member.

### Proposed solution

Add a polymorphic channel target that can represent either an Agent or a Squad, with DingTalk as the first product surface:

- Let workspace owners/admins connect a DingTalk robot to an active Squad from the Squad detail surface.
- Let owners/admins select Agents or Squads in Settings → Integrations → DingTalk group routing, with the two target types visually separated.
- Persist a target as `target_type + target_id` rather than overloading an Agent id. Migrate existing DingTalk installations and routes as `target_type=agent` without requiring reconnection.
- Resolve a Squad target to its current active Leader at dispatch time. Queue the Leader with explicit Squad provenance so the claim receives the Squad Operating Protocol, roster, and instructions.
- Send the Leader's result back to the same DingTalk DM or group conversation through the existing outbound path.
- Keep the route attached to the Squad when its Leader changes. Future messages should resolve to the new Leader and must not resume the former Leader's agent session.
- For `/issue`, create the Issue with `assignee_type=squad` and `assignee_id=<squad id>`, so the existing Squad assignment flow wakes the Leader and preserves Squad ownership.
- Preserve existing account binding, workspace membership, access checks, route revision fencing, deduplication, and cross-workspace isolation.

The channel engine should expose the target generically enough that other chat adapters can adopt it later, but this issue only requires DingTalk UI/API/adapter support.

### Acceptance criteria

- [ ] A workspace owner/admin can connect a DingTalk Stream robot to an active Squad.
- [ ] A discovered DingTalk group route can target either an active Agent or an active Squad in the same workspace.
- [ ] A normal DM or group `@` message on a Squad route queues exactly one task for the Squad's current Leader.
- [ ] The Leader task carries authoritative Squad role/provenance and receives the Squad protocol, roster, and instructions; it is not downgraded to an ordinary direct Agent chat.
- [ ] The Leader can answer directly or distribute work to Squad members through the supported Squad coordination flow, and user-visible results return to the originating DingTalk conversation.
- [ ] `/issue <title>` creates an Issue assigned to the Squad, not directly to its Leader.
- [ ] Changing the Squad Leader keeps the DingTalk route on the Squad, routes the next message to the new Leader, and prevents reuse of the former Leader's agent session.
- [ ] Archived Squads and Squads with an unavailable/archived Leader cannot be selected. Existing affected routes remain visible with a clear unavailable state and do not silently fall back to another Agent.
- [ ] Switching a route between Agent and Squad targets, or between two Squads, cannot reuse the previous target's Chat session.
- [ ] Members still must bind their DingTalk identity and pass workspace membership and target access checks on every message.
- [ ] Existing Agent-bound DingTalk installations and group routes migrate without reconnecting, changing behavior, or exposing credentials.
- [ ] Newer Web/Desktop clients degrade safely against older backends that only return `agent_id`.
- [ ] Backend, shared channel-router, API-schema, UI, realtime, migration, and documentation tests cover Agent and Squad targets.

### Non-goals

- Broadcasting one message to every Squad member.
- Running all Squad Agents in parallel.
- Dynamically selecting an arbitrary Agent by name from DingTalk text.
- Cross-workspace fan-out or sharing Chat/session data between workspaces.
- Adding Squad targets to Slack, Feishu/Lark, or WeCom in this issue.

### Related

- #7077 — route one DingTalk bot across authorized workspaces.
- #7036 — distinguish DingTalk bot identities in same-group routing.

### Screenshots / mockups

No separate mockup yet. Follow the existing Agent → Integrations and Settings → Integrations → DingTalk layouts, adding a Squad integration entry point and grouped Agent/Squad target selectors.

mxbao063-bao:
## Summary

- allow DingTalk installations and discovered group routes to target an Agent or Squad
- resolve Squad messages to the current Leader only, stamp Squad leader tasks, and keep /issue assigned to the Squad
- fence session creation, message append, task enqueue, and outbound replies across Leader changes, including A → B → A handoffs
- add Agent/Squad routing controls in Settings and a DingTalk integration entry on Squad details
- document the Leader-only behavior in all supported locales

## Verification

- Go tests: DingTalk, channel engine, handler, and service packages
- go vet ./...
- pnpm typecheck
- pnpm lint
- 75 Core API schema tests
- 32 DingTalk Views tests
- sqlc generate and git diff checks

## Build note

The desktop production build completed. The full pnpm build then stopped only because the local environment could not validate the Google Fonts TLS certificate used by Next font fetching.

Closes #7165


yyclaw:
Thanks for putting this together. Enabling people to direct a Squad through an IM channel is exactly a capability I want us to add. That said, I think the product design and interaction semantics need a bit more time to think through carefully before we settle on the implementation. I expect to start a detailed review next week.

mxbao063-bao:
@yyclaw Thanks for the feedback. That makes sense — I’ll keep the PR open and wait for the detailed review next week. Happy to adjust the implementation once the product semantics are clearer.

yyclaw (changes requested):
Thank you for contributing this change.

This PR cannot be approved in its current form. Assigning tasks to a Squad through a Bot should be a generic capability available across all Channels. Although this PR changes the shared Channel engine, its product UI and interaction flow remain DingTalk-only. This leaves the shared abstraction inconsistent with the product scope and makes the intended product behavior unclear.

### [DESIGN-001][P1][blocking][shared_engine] The shared engine and product scope are inconsistent

A complete product design should include the following:

1. Design a generic flow for all Channels rather than a DingTalk-only flow.
2. Preserve the existing underlying architecture in which one Bot is bound to exactly one Agent. After a task is assigned to a Squad, reuse the existing Squad task execution logic.
3. Forward only the exact `/issue` command to the Squad. All other interactions should continue to be handled directly by the Bot's bound Agent.
4. Define authorization behavior for every boundary case, including both allowed and denied paths.

I plan to publish a complete spec later today. If you are interested, feel free to pick up the implementation. No worries if not—I plan to implement the spec this week, and we can continue the discussion together based on the new implementation.

cc @Bohan-J

yyclaw:
I have completed the spec and designed the domain model. However, while working through the UI design, I tried several iterations and was not satisfied with any of them. The functionality itself is straightforward to add, but it would make the page feel cluttered because the Channel integration panel already contains too much information.

Also, even without the /issue command, users can already assign an issue to a Squad using natural language; it simply consumes additional tokens.

For now, I will not move this forward until I have a UI and interaction design that I am satisfied with.

Issue #6941, opened 2026-08-13
[Bug]: Verified autopilot leader cannot create a backlog child assigned to a private agent


emosheeep:
## What happened?

An unattributed (`originator_user_id = NULL`) leader task dispatched by a member-created autopilot can `@mention` a private worker owned by that autopilot creator, because the comment path restores the creator's invocation authority from verified task/issue lineage (the behavior added by #5512).

The same verified leader task cannot create a child issue assigned to that same private worker. `POST /api/issues` returns:

```json
{"error":"cannot assign to private agent"}
```

This also fails when the new child has `status: "backlog"`. Backlog assignment is explicitly a parking operation and does not enqueue a run, so the rejection cannot be explained as protection against an immediate execution.

I am **not** requesting a blanket A2A/private-agent bypass. The expected admission is limited to the same creator-owned target and the same server-verified autopilot lineage already accepted by the comment delegation path. Missing, mismatched, cross-issue, or unauthorized lineage should continue to fail closed.

## Steps to reproduce

Tested directly against upstream `main` at [`9ba3a8bdf26864f41fcfa04bdc83d6d3cced0ec9`](https://github.com/multica-ai/multica/commit/9ba3a8bdf26864f41fcfa04bdc83d6d3cced0ec9), with a fresh migrated `pgvector/pgvector:pg17` database.

Add this temporary regression test under `server/internal/handler`:

```go
package handler

import (
	"net/http"
	"net/http/httptest"
	"testing"
)

func TestReproAutopilotLeaderCreatesBacklogChildAssignedToPrivateAgent(t *testing.T) {
	if testHandler == nil || testPool == nil {
		t.Skip("database not available")
	}

	workerID, ownerID, _ := privateAgentTestFixture(t)
	fx := newAutopilotDelegationFixture(t, workerID, ownerID, "autopilot")
	parentIssueID := uuidToString(fx.Issue.ID)

	w := httptest.NewRecorder()
	r := newRequest("POST", "/api/issues?workspace_id="+testWorkspaceID, map[string]any{
		"title":           "repro: park private-agent handoff in backlog",
		"status":          "backlog",
		"priority":        "low",
		"assignee_type":   "agent",
		"assignee_id":     workerID,
		"parent_issue_id": parentIssueID,
		"allow_duplicate": true,
	})
	r.Header.Set("X-Agent-ID", fx.LeaderAgentID)
	r.Header.Set("X-Task-ID", fx.LeaderTaskID)
[...]

Issue #6738, opened 2026-08-11
[Feature]: Independent agent-to-agent (A2A) invocation permission settings


HuAzurestar:
## Deployment type

- Official App

## What do you want and why?

Agents today can @-mention and trigger one another, but there is no fine-grained control over *which agents* may invoke a given agent when the triggering run has no human originator (e.g. run-only autopilots). The existing permission_mode governs member-facing visibility and does not distinguish agent callers, so teams either leave everything fail-closed or open agents up workspace-wide with no per-agent granularity.

## Proposed solution

Add an independent, owner-configurable **A2A invocation** axis on each agent, orthogonal to permission_mode:

- default (unset): status-quo fail-closed - A2A calls keep being judged against permission_mode, preserving the "no human originator -> deny" behavior with zero migration for existing agents.
- ny_agent: any agent principal in the workspace may invoke this agent.
- squad_leaders: only agent principals who lead a squad may invoke.
- specific_agents: only an explicit allow-list of agent principals may invoke (whitelist stored in a dedicated table).

The axis only governs **agent** callers; **system** principals keep their existing judgment unchanged (workspace-broad exception and fail-closed behavior are untouched). It is configurable in the agent detail page (owner-only write; non-owners see a read-only summary) and summarized in the agent list. Grant rows are cleaned up on agent hard-delete and on workspace teardown.

Bohan-J:
Accepting this feature request — the gap you describe is real: for calls with no human originator there's currently no middle ground between fail-closed and workspace-wide open, and the agent's owner should be the one who decides.

We'll design the approach ourselves before any implementation lands, mainly because two things need to be settled at the product level first: the combined decision table for member visibility (`permission_mode`) × A2A eligibility, and what a denied call actually surfaces to the caller's automation log and to the target agent's owner. We're also likely to ship a narrower set of modes than proposed — coupling scope to squad structure means an agent's callable range changes silently when a squad is reorganized.

For that reason we've closed #6739 unmerged; that's a scoping decision, not a comment on the quality of the patch. This issue stays open and we'll post the design here.

HuAzurestar:
Thanks for reviewing my issue. The issue was proposed automatically by my agent, but the feature is surely from my own demand.

1. You're totally right about the complexity of combining member visibility and A2A permissions. I agree that my initial design was a bit over-complex and not thorough enough.
[...]

Issue #4675, opened 2026-06-28
[Feature]: Allow private agents to access each other


jiangwangyang:
### Deployment type

Official App

### What do you want and why?

private的智能体之间希望也能够互相访问，否则难以组织成智能体协作

PR #7202, opened 2026-08-19
fix(agents): align private-agent invoke surfaces with the owner-only contract (MUL-6380)

### LINKED ISSUE #7180: [Bug]: Align private-agent assignment permissions with the documented access contract
## Summary

The private-agent authorization contract is inconsistent across the product. The current access check treats private agents as owner-only for invocation and issue assignment, while documentation and some UI/helper copy state or imply that workspace administrators can assign private agents. This contract drift produces denied assignments and incorrect security expectations.

## Steps to reproduce

1. Create a private agent owned by one workspace member.
2. Sign in as a workspace administrator who is not the private agent owner.
3. Try to assign an issue to that private agent, or use another supported invocation surface.
4. Observe that the operation is rejected by the private-agent access check.
5. Compare the result with the administrator access described in the relevant agent and assignment documentation.

The same mismatch is relevant for ordinary members and agent actors because the effective policy is not stated consistently per surface.

## Actual behavior

Private-agent assignment/invocation follows an owner-only gate, but the surrounding product contract does not consistently communicate or implement that rule. A denied operation does not always make clear whether the actor lacks ownership, lacks invocation rights, or whether administrator access is intentionally excluded.

## Expected behavior / decision needed

Choose and explicitly document one product contract:

1. **Owner-only:** only the private-agent owner may invoke the agent or assign work to it; workspace administrators retain only the explicitly documented management/read capabilities, or
2. **Audited administrator path:** administrators may assign/invoke under a narrowly scoped, server-enforced rule with clear auditability and no blanket bypass for unrelated agent callers.

Whichever option is selected, it must be applied consistently to backend checks, API responses, shared permission helpers, web/desktop/mobile UI, and documentation.

## Acceptance criteria

- The owner-only versus audited-administrator contract is recorded as an explicit product decision.
- Backend authorization, API/UI behavior, shared permission helpers, tests, and all relevant documentation implement the same contract.
- The access matrix has explicit tests for the private-agent owner, workspace administrator, ordinary member, and agent actors.
- Denied assignments return a clear, actionable error that identifies the required permission without exposing sensitive information.
- No unconditional admin or agent-to-agent bypass is introduced; unrelated private-agent access remains fail-closed.

## Context

This issue is distinct from:

- #6941, which covers a narrower verified autopilot-lineage assignment case;
- #6738, which proposes configurable agent-to-agent invocation permissions;
- #4675, which requests private agents accessing one another.

The requested change is to make the existing private-agent permission contract explicit and consistent, not to add blanket agent-to-agent access.

Bohan-J:
Fixes #7180.

## What the report asked for

#7180 asks us to pick one private-agent permission contract — strict owner-only, or an audited administrator path — and apply it everywhere.

Reading the code, **the decision was already made and shipped**: MUL-3963 split the model into two gates, and the server has enforced the owner-only version ever since.

| | who passes |
|---|---|
| **view** an agent (`canAccessPrivateAgent`) | owner + workspace owner/admin |
| **invoke** an agent — assign, `@mention`, chat (`canInvokeAgent`) | **owner only** for `permission_mode: "private"`; the `public_to` allow-list otherwise. No admin bypass, no A2A bypass. |

That split is deliberate: running someone's personal agent spends *their* runtime and uses *their* connected Composio/OAuth accounts, so being an admin buys management and inventory visibility, not the ability to run.

So this PR does not re-open the decision. It fixes the surfaces that never caught up with it — which is what produced the denied-assignment-with-no-explanation the report describes.

## What was actually inconsistent

**1. Mobile had a hand-copied permission rule stuck on the old model.** `apps/mobile/lib/can-assign-agent.ts` mirrored the pre-MUL-3963 rule and was wrong in both directions:

- *Too permissive*: it granted workspace owner/admin a bypass on private agents, so an admin saw a teammate's personal agent in the chat picker, selected it, typed, and only got the refusal after `POST` hit `canInvokeAgent`. The UI promised something the server was always going to refuse — this is the reported scenario.
- *Too restrictive*: it judged on the derived `visibility` field, which collapses "`public_to`, scoped to specific people" into `"private"`. A member the owner had **explicitly granted access to** therefore never saw that agent on mobile at all.

The same app already used the correct shared rule in `mention-suggestion-bar.tsx`, so the `@mention` list and the chat picker disagreed with each other on one device.

**2. The UI copy stated the old contract next to a control enforcing the new one.** The "Personal" badge tooltip — also its `aria-label`, so screen readers announce it too — read *"Personal · only you and workspace admins can use this agent"* in all four locales, on the agent detail page, overview summary and hover card. The assignee picker beside it was disabled and read *"Personal agent — only the owner can assign work."* An admin looking at that pair cannot tell whether the refusal is policy or a bug. `packages/core/agents/visibility-label.ts` carried the same claim (create-agent dialog, visibility picker, agents-list lock tooltip), with doc comments asserting the opposite contract and citing a line reference that has long since moved.

**3. The denial message named the target's configuration.** `cannot assign to private agent` both discloses the target's permission mode to someone just told they may not use it — the opposite of what `ReasonInvocationNotAllowed` deliberately does — and is factually wrong for a `public_to` agent scoped to specific people, which is refused through the same branch.

## Changes

| area | change |
|---|---|
| `apps/mobile` | delete the mirrored rule; call `canAssignAgentToIssue` from `@multica/core/permissions` in the chat picker and the availability hook |
| `packages/core/agents/visibility-label.ts` | correct `VISIBILITY_DESCRIPTION` / `VISIBILITY_TOOLTIP` and the doc comments |
| `packages/views/locales/{en,zh-Hans,ja,ko}` | correct the `VisibilityBadge` tooltip |
| `server/internal/handler/issue.go` | report the missing permission instead of the target's permission mode |

The mirror is deleted rather than repaired: the root `CLAUDE.md` sharing rule lets mobile import pure functions from `@multica/core`, so the copy bought nothing but a second thing to drift. Copy wording distinguishes the two contexts — the create dialog describes a *choice* (`permission_mode: "private"` → "Only you can assign"), while the badge describes an *existing* agent whose `visibility: "private"` is lossy, so it names the owner's grants ("only the owner and people they allow") rather than promising either extreme.

## Not changed

No authorization behavior on the server. `canInvokeAgent` is untouched, no admin or agent-to-agent bypass is introduced, and unrelated private-agent access stays fail-closed. The only server change is the text of an existing 403.

## Acceptance criteria from #7180

- ✅ contract recorded explicitly — now stated in `visibility-label.ts` and the mobile call sites, pointing at `canInvokeAgent` as the source of truth
- ✅ backend, API/UI behavior, shared helpers and copy implement the same contract
- ✅ access matrix has explicit tests for owner / workspace admin / ordinary member / agent actors — `packages/core/permissions/rules.test.ts` and `server/internal/handler/agent_access_test.go`; deleting the mobile mirror puts mobile under that same canonical coverage instead of an untested copy
- ✅ denied assignments name the required permission without exposing the target's configuration
- ✅ no unconditional admin or A2A bypass introduced

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm test` (4402 tests, 380 files) — pass
- `apps/mobile`: `tsc --noEmit`, `vitest run` (74 tests), `expo lint` — pass, 0 errors
- `server`: `go build ./...`, `go vet ./internal/handler/`, `go test ./internal/handler/` — pass. One unrelated pre-existing failure locally, `TestPluginRoutesRequireWorkspaceAdmin` (a fixed-email fixture colliding with leftover dev-DB rows); it fails identically on a clean checkout of `main`.


Bohan-J:
@maxivillus thanks for the careful write-up — the reproduction and the "decide, then apply everywhere" framing are exactly right, and they made this quick to track down. Here's where things actually stand.

**The decision you asked us to make has already been made — it just wasn't visible from the outside.** MUL-3963 split private-agent authorization into two gates, and the server has enforced the strict owner-only version of your option 1 ever since:

- **view** an agent (open it, read its config, manage/archive it): owner + workspace owner/admin
- **invoke** it — assign an issue, `@mention`, chat: **owner only** when `permission_mode` is `private`; otherwise whoever is on the `public_to` allow-list. No admin bypass, no agent-to-agent bypass.

The reasoning behind excluding admins is that running someone's personal agent spends *their* runtime and uses *their* connected Composio/OAuth accounts. Being a workspace admin gets you management and inventory visibility over that agent, not the ability to act as it.

**So what you hit is real, but it is not the authorization layer — it is the surfaces that never caught up with that gate.** Concretely:

1. The mobile app carried a hand-copied version of the old rule. It let an admin pick a teammate's personal agent in the chat picker and only surfaced the refusal after the request reached the server — the UI promised something that was always going to be refused. That same stale copy was also wrong in the *opposite* direction: it judged on a coarse legacy field that cannot distinguish "owner-only" from "shared with specific people", so a member the owner had explicitly granted access to never saw that agent on mobile at all.
2. The "Personal" badge tooltip still read *"only you and workspace admins can use this agent"* (in all four locales, and as its `aria-label`), sitting directly beside a disabled assignee picker reading *"only the owner can assign work."* Two labels on one screen stating opposite rules — which is precisely why a denial was impossible to interpret as intended-or-broken.
3. The denial itself said `cannot assign to private agent`. That both names the target's permission mode to someone who was just told they may not use it, and is factually wrong for a `public_to` agent scoped to specific people, which is refused through the same code path.

This PR fixes all three: mobile now calls the shared rule instead of a copy, the copy no longer claims admin access, and the 403 names the missing permission rather than the target's configuration.

**One thing deliberately left alone:** no authorization behavior changed. The invoke gate is untouched, no admin or agent-to-agent bypass was added, and unrelated private-agent access stays fail-closed. The only server-side change is the wording of an existing 403.

On your acceptance criteria — the explicit contract, single implementation across backend/UI/copy, the owner/admin/member/agent-actor matrix (covered in `packages/core/permissions/rules.test.ts` and `server/internal/handler/agent_access_test.go`; deleting the mobile copy brings mobile under that same coverage instead of an untested duplicate), the actionable non-leaking error, and no new bypass — all met.

Worth calling out for the record: you were right that the contract was inconsistent, and right that #6941 / #6738 / #4675 don't cover it. What we found is that the drift ran in both directions — some users were being shown access they didn't have, and others were being denied access they did have.

PR #7212, opened 2026-08-19
MUL-6406: fix(auth): fail fast in production on insecure JWT secret defaults

BuXiuDN:
## What does this PR do?

Self-hosted Multica could silently boot with a publicly-known JWT signing key: the server falls back to a hardcoded dev secret when `JWT_SECRET` is empty, docker-compose injects `change-me-in-production` by default, and `.env.example` told users to copy that value - while the existing startup warning never fired because compose always provided a non-empty value. Since the auth middleware validates HS256 signatures without a DB lookup, anyone knowing the secret could forge tokens and impersonate any user.

This PR closes that weak-default chain:

- The production server now **refuses to boot** (exit 1) when `JWT_SECRET` is empty or a known placeholder, with an actionable remediation hint in the log (`openssl rand -hex 32`).
- docker-compose now requires `JWT_SECRET` (`${JWT_SECRET:?}` fails before containers are created).
- `.env.example` and the self-hosting docs guide generating a strong secret.

Non-production boot behavior is unchanged (empty secret keeps the historical dev fallback + warning), so local development and existing tests are unaffected.

## Related Issue

No dedicated issue filed. Note: PR #3541 (open, unmerged, code-layer only) includes an overlapping startup guard for the same default-secret problem; this PR is a focused, self-contained fix that also covers the deployment templates and docs that shipped the weak defaults.

## Type of Change

- [x] Bug fix (security: insecure default ? auth bypass)
- [x] Documentation update

## Changes Made

- `server/internal/auth/jwt.go` - add `ValidateJWTSecret` + denylist of known placeholders (empty, dev default, compose/.env default); `JWTSecret()` fallback unchanged.
- `server/internal/auth/jwt_secret_test.go` - new table-driven tests for the validation matrix.
- `server/cmd/server/main.go` - new pure helper `jwtSecretBootError`; boot guard fails fast in production on empty/known-default secret, keeps the dev-only warning otherwise.
- `server/cmd/server/main_test.go` - new table-driven tests for the boot guard.
- `docker-compose.selfhost.yml` - `JWT_SECRET` now required (`${JWT_SECRET:?...}`); usage comment updated.
- `.env.example` - `JWT_SECRET` left empty with generation commands and a note that only local non-Docker dev may omit it.
- `SELF_HOSTING.md`, `SELF_HOSTING_ADVANCED.md`, `apps/docs/content/docs/environment-variables.{en,zh,ko,ja}.mdx` - `JWT_SECRET` documented as required in production.

## How to Test

1. `cd server && go test ./internal/auth/... ./internal/middleware/... ./internal/handler/... ./cmd/server/...` - all pass (verified locally; `daemon`/`repocache`/`cli` suites have pre-existing Windows/sandbox environment failures unrelated to this change).
2. `APP_ENV=production` with empty or `change-me-in-production` `JWT_SECRET` ? process exits 1, log contains `refusing to start` and `openssl rand -hex 32`.
3. `APP_ENV=production` with a 64-hex secret ? boots normally.
4. `APP_ENV` unset ? boots with the existing dev-default warning.
5. `docker compose -f docker-compose.selfhost.yml config` without `JWT_SECRET` ? errors before container creation.
6. E2E: with a strong secret, email-code login + `/api/me` succeeds; a token forged with the old weak secret returns 401.

## Checklist

## AI Disclosure

**AI tool used:** Trae (agentic coding)

**Prompt / approach:**
Implemented from a written spec (fail-fast boot guard + deployment hardening) following the repo's existing fail-closed patterns (`MULTICA_DEV_VERIFICATION_CODE` disabled in production, feature-flag parse failure `os.Exit(1)`). Every change verified with `go build ./...`, `go vet`, and targeted/full package tests.

## Breaking change note (for maintainers)

Self-hosted instances that have been running on the default secret will refuse to boot after this change - that is intentional (the point is to force rotation). Upgrade path: set a strong `JWT_SECRET` first, then pull the new image. On rotation, browser JWTs, daemon tokens, and avatar/attachment capability HMACs derived from the old key stop working; PATs are stored as SHA-256 hashes and are unaffected.


multica-eve (changes requested):
Requesting changes for one security-blocking issue:

- `.env.example:49` uses `Get-Random` to generate the Windows JWT signing secret. Microsoft documents that `Get-Random` does not provide cryptographically secure randomness, but this command is presented as generating a strong production secret. That can leave Windows self-hosters with a predictable signing key and undermines the security goal of this PR.

Please replace it with a CSPRNG-backed command that works on supported Windows PowerShell versions, for example:

```powershell
$b = New-Object byte[] 32; $rng = [System.Security.Cryptography.RandomNumberGenerator]::Create(); $rng.GetBytes($b); $rng.Dispose(); -join ($b | ForEach-Object { $_.ToString('x2') })
```

Reference: [Microsoft Learn — Get-Random](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/get-random) and [RandomNumberGenerator](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator).

The production startup guard and deployment-path changes otherwise look sound. Once the Windows generation guidance uses a cryptographically secure source, this blocker is resolved.

BuXiuDN:
Addressed in 68ce5e363 - the Windows hint now uses the `RandomNumberGenerator`-backed command exactly as suggested:

```powershell
$b = New-Object byte[] 32; $rng = [System.Security.Cryptography.RandomNumberGenerator]::Create(); $rng.GetBytes($b); $rng.Dispose(); -join ($b | ForEach-Object { $_.ToString('x2') })
```

Verified locally on Windows PowerShell 5.1 (outputs 64 hex chars). It also matches the CSPRNG shape `scripts/install.ps1` already uses internally for the secrets it generates (`New-RandomHex`), so the guidance and the installer now agree on one source of randomness. This was the only `Get-Random` occurrence in the repo.

@multica-eve The blocking item is fixed; ready for re-review whenever convenient.

multica-eve (approved):
The security blocker and the remaining documentation nits are resolved on the latest head. Targeted auth/startup tests and the self-host configuration check pass. Approving for merge.

PR #7227, opened 2026-08-19
perf(realtime): stop the task:message fanout from flooding every client (MUL-6396)

Bohan-J:
## What changes for users

Today, whenever anyone in the workspace is running a task — any person, any machine — every open window quietly records that run's entire execution log in the background, whether or not you are looking at it. The more runs and the longer they go, the slower and heavier the window gets. That is the main source of the "inbox full of running issues and everything I click feels sticky" report in MUL-6396.

After this, a window only does work for the task you are actually looking at. Everything else is ignored on arrival, and opening it later fetches the full log from the server — no data is lost. For the log you *are* watching, the UI refreshes at most once per 100ms instead of once per streamed message.

No settings to change, no change in how anything is used.

**Gains**
- **Main case (lots of runs, you are not watching them):** background cost drops from "scales with the whole workspace's run count" to approximately nothing, and memory stops accumulating. Biggest win on busy multi-person workspaces.
- **Watching an execution log:** roughly an order of magnitude fewer renders; scrolling and typing stay smoother.
- **Scope:** this only addresses "background runs slow everything down". Click-to-navigate freezes are a separate issue (navigation transitions), and dev-build overhead is a third thing. The three do not overlap.
- **Risk surface:** limited to the local cache path for execution logs; the data always lives on the server, so the worst failure is "log renders incomplete", never data loss. That is precisely what the review's must-fixes were guarding.

## Implementation

**Only cache what is being watched** (`packages/core/realtime/use-realtime-sync.ts`)

`task:message` is broadcast to every client for every run (`server/cmd/server/listeners.go:241`), and the handler's `(old = [])` default built a timeline entry the first time it saw a task id. Frames for a task with no cache entry are now dropped. The row is persisted before it is broadcast, so dropping costs nothing — the next open fetches it.

The gate is entry presence, not observer count. Observer count is a tighter bound but an unsafe one: under `staleTime: Infinity`, a task that briefly drops to zero observers while navigating would discard frames, and the remount would read the surviving cache as fresh and never fetch them back. Presence has no such gap. Its cost is a bounded tail — writes do not postpone GC, so an entry outlives its last viewer by at most one `gcTime`.

**One write per window** (same file)

Surviving frames batch into a single cache write per 100ms. The window is fixed rather than a resetting debounce, so a continuous stream still flushes instead of being starved until it pauses.

Because the write happens up to a window after the frame was accepted, and `setQueryData` does *not* postpone garbage collection (query-core arms that timer on last-observer-leaves and never re-arms it on write), the flush re-checks that the entry still exists immediately before writing. Without that, closing a transcript on a still-running task rebuilds the entry holding only the last batch, and `staleTime: Infinity` means the next open never fetches the rest.

**Clip oversized payloads out of the broadcast — off by default** (`server/internal/handler/daemon.go`)

`output` is capped at 8KB by the daemon; `input` is not, so one large `Write` ships its whole body to every client in the workspace. This clips it and marks the payload `truncated`; persisted rows and both REST endpoints are untouched.

⚠️ **It is gated behind `MULTICA_CLIP_TASK_MESSAGE_BROADCAST` and off by default. Do not enable it with this deploy.** Clipping is not an additive field older clients can ignore — it changes what `input`/`output` mean for consumers that already read them, and a client without the `truncated` reader caches the clipped copy under `staleTime: Infinity` and never refetches. Enabling it requires a deliberate decision that clients understanding `truncated` have saturated; there is no per-connection gate, because the hub does not retain the `client_version` it receives at upgrade and frames cross nodes through the Redis relay as serialized bytes — the same protocol work per-task scope routing is waiting on. The env var is a config-level rollback (no code change), but flipping it still follows normal config rollout — process restart, not a live toggle. A test pins that the default path is byte-for-byte unchanged.

**One merge rule for the timeline cache** (`packages/core/chat/queries.ts`)

Not caching unopened tasks made first open of a live task actually fetch, where it previously found a pre-seeded cache. A fetch replaced the array wholesale, so a frame arriving while the request was open was dropped — and `staleTime: Infinity` meant nothing would fetch it again. All three writers (initial fetch, backfill, realtime batch) now fold through `unionTaskMessagesBySeq`, installed as the query's own `structuralSharing`: union by seq, server data wins on conflict, rows the response never mentioned are kept. Owning the rule at the query means no future writer can bypass it.

## Not included

- **One WS frame per reported batch** instead of one per message. Needs the same client capability negotiation as the clipping flip.
- **Mobile delivery.** `use-chat-session-realtime.ts` gates `task:message` on `chat_session_id`, which the server payload does not carry, so mobile appends nothing at all today. Pre-existing and reported separately; this PR does not change mobile delivery behaviour. The `truncated` branch is present so whichever side is fixed first does not land on a client that caches clipped rows.

## Verification

- `pnpm typecheck` 7/7, `pnpm test` 6,715 passing, `pnpm lint` 0 errors, `go build ./...` clean, 9 Go clipping tests passing.
- Client tests cover: frames dropped for unheld tasks; kept for held tasks before the fetch resolves; burst coalescing; fixed-window flushing; a fetch deferred past a live flush; fetched row winning over a clipped one; and GC landing inside the flush window (asserting the history is still recoverable by reopening, not merely that the entry stays absent). Each was verified to fail without its fix.
- Go handler tests have 459 pre-existing failures in this sandbox for lack of DB fixtures; the failure set is identical with and without this change. `TestReportTaskMessages*` covers the modified call site and runs in CI.

## After merge

Watch `client_unresponsive` (`source=longtask`) by version: the new cohort should sit clearly below the old one during high-concurrency periods. That curve is also the basis for closing MUL-6396.

### DIFF
--- a/apps/mobile/data/realtime/chat-ws-updaters.ts
+++ b/apps/mobile/data/realtime/chat-ws-updaters.ts
@@ -332,6 +332,14 @@ export function appendTaskMessage(
   // only in the DB, and `taskMessagesOptions` is staleTime:Infinity — without
   // this the clipped text would be pinned forever. Invalidating refetches for
   // whoever is mounted and is a no-op otherwise.
+  //
+  // Not reachable today, and kept deliberately: `use-chat-session-realtime`
+  // gates `task:message` on `chat_session_id`, which the server's
+  // TaskMessagePayload does not carry, so mobile currently appends nothing at
+  // all (a pre-existing delivery gap, tracked separately). Whichever side of
+  // that is fixed first, this branch must already be here — the clipping is
+  // exactly what a client learning to receive these frames would otherwise
+  // cache permanently.
   if (payload.truncated) {
     qc.invalidateQueries({ queryKey: chatKeys.taskMessages(payload.task_id) });
   }
--- a/apps/mobile/data/realtime/chat-ws-updaters.ts
+++ b/apps/mobile/data/realtime/chat-ws-updaters.ts
@@ -326,4 +326,13 @@ export function appendTaskMessage(
       return [...old, payload].sort((a, b) => a.seq - b.seq);
     },
   );
+
+  // The server clips oversized tool input/output out of the realtime copy so
+  // one big Write is not broadcast to every client (MUL-6396). The full row is
+  // only in the DB, and `taskMessagesOptions` is staleTime:Infinity — without
+  // this the clipped text would be pinned forever. Invalidating refetches for
+  // whoever is mounted and is a no-op otherwise.
+  if (payload.truncated) {
+    qc.invalidateQueries({ queryKey: chatKeys.taskMessages(payload.task_id) });
+  }
 }
--- a/apps/mobile/data/schemas.ts
+++ b/apps/mobile/data/schemas.ts
@@ -350,6 +350,9 @@ export const TaskMessagePayloadSchema: z.ZodType<TaskMessagePayload> = z.object(
   input: z.record(z.string(), z.unknown()).optional(),
   output: z.string().optional(),
   created_at: z.string().optional(),
+  // Set on the WS copy when the server clipped input/output for the fanout
+  // (MUL-6396); never present on the REST list response.
+  truncated: z.boolean().optional(),
 }).loose();
 
 export const TaskMessageListSchema = z.array(TaskMessagePayloadSchema).default([]);
--- a/packages/core/chat/queries.test.ts
+++ b/packages/core/chat/queries.test.ts
@@ -8,6 +8,7 @@ import {
   mergeTaskMessagesBySeq,
   sortChatSessions,
   taskMessagesOptions,
+  unionTaskMessagesBySeq,
 } from "./queries";
 
 const msg = (seq: number): TaskMessagePayload => ({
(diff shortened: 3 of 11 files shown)


Bohan-J:
Pushed `acaf3660f` addressing both must-fixes.

**1. Old-client compatibility.** You're right, and the PR description had it backwards — I'd recommended shipping the server half *first*, which is precisely the change that hurts un-updated clients. Clipping isn't an additive field they can ignore; it changes what `input`/`output` mean for consumers that already read them, and `staleTime: Infinity` means the clipped copy is pinned until reload.

Capability routing is the principled fix but isn't reachable from here: the hub doesn't retain the `client_version` it receives at upgrade (`hub.go:863` only logs it), and frames cross nodes through the Redis relay as already-serialized bytes, so gating would need both variants relayed. That's the same protocol work per-task scope routing is blocked on.

So clipping now sits behind `MULTICA_CLIP_TASK_MESSAGE_BROADCAST`, **off by default**. The reader ships first; flipping the flag is a later ops step once installed builds have caught up, and it's revertible without a deploy. `TestTruncateTaskMessageForBroadcast_DisabledByDefault` pins that the default path is byte-for-byte unchanged.

**2. Late fetch dropping live seqs.** Confirmed, and confirmed as newly reachable — before P0-a the cache was pre-seeded, so first open of a live task found fresh data and never fetched at all. Removing pre-seeding turned that into a real fetch, with the race you describe.

Rather than patch the one path, all three writers (initial fetch, backfill, realtime batch) now fold through a single rule installed as the query's own `structuralSharing`: `unionTaskMessagesBySeq` — union by seq, server data wins on conflict, rows the response never mentioned are kept rather than deleted. Since the query itself owns the rule, no future writer can bypass it. It also subsumes the backfill's hand-rolled truncated-row eviction, which is now deleted, and `transcript-button.tsx` drops its own merge for the same reason.

New tests: a fetch deferred past a live flush (the exact case you reproduced), fetched-row-wins-over-clipped, and five unit tests on the merge rule including reference stability. Verified all of them fail with `structuralSharing` removed.

**3. Mobile unreachable branch.** Verified — `use-chat-session-realtime.ts:117` gates on `chat_session_id`, `protocol.TaskMessagePayload` has no such field and `taskMessageToPayload` doesn't set one, so mobile appends *nothing* today. That's a pre-existing delivery gap (mobile's live trace is dead, independent of this PR); I've reported it separately rather than changing mobile delivery behaviour inside a perf PR. The branch stays, with a comment explaining why: whichever side gets fixed first shouldn't land on a client that caches clipped rows.

**4. Observer count vs cache presence.** I'd push back on this one. Switching to observer count introduces a hole: with `staleTime: Infinity`, a task that briefly drops to zero observers (unmount/remount during navigation) would discard frames, and on remount the query finds cached data and *doesn't* refetch — so those seqs are gone until reload. That's the same failure mode as finding 2. Cache presence has no such gap: either the entry exists and keeps accumulating, or it's been GC'd and the next open fetches the full history.

The cost is the bounded tail you identified — up to `gcTime` (10 min) of accumulation after the last viewer closes, and only for tasks actually opened. The unbounded problem (every workspace task, forever) is what P0-a fixes. If we want to tighten it further, lowering `gcTime` for this query family is the safe lever and doesn't reintroduce the hole. I've adjusted the comment so it no longer promises more than it delivers.

Verification: typecheck 7/7, `pnpm test` 6,712 passing, lint 0 errors, Go build clean, 9 truncation tests passing.

PR #7240, opened 2026-08-19
feat(issues): extend property filter to text / number / date / url (MUL-6423)

### LINKED ISSUE #7235: [Feature]: Extend custom-property filter (incl. "No value") to text / number / date / url
### Background

Follow-up from #7156 (merged): the "No value" custom-property filter shipped for
`select` / `multi_select` / `checkbox` / `actor`. During review @Bohan-J flagged
two non-blocking follow-ups — the i18n consolidation (landed separately) and
this one: extending value + "No value" filtering to the remaining property
types.

### Current gap

- `filterableProperties` (`packages/views/issues/components/issues-header.tsx`)
  only admits `select` / `multi_select` / `checkbox` / actor, so `text`,
  `number`, `date` and `url` properties never reach the filter menu.
- The table facet for those types hits the `default` branch and returns 422
  (`server/internal/handler/issue_table_facets.go`), so counts can't be shown.
- The filter predicate already handles string/array/boolean containment via
  `parsePropertiesFilterParam`, but `number` values arrive as JSON strings in
  the filter param (`map[string][]string`), so `{"<defId>": "3.5"}` will not
  match a stored numeric `3.5`. `date` values also need a defined shape.

### Goal

Unify value + "No value" filtering across every custom property type, so the
same filter menu works for text / number / date / url as it does for the four
current types. The `__none__` sentinel is already type-agnostic (key absence),
so "No value" should be reusable as-is.

### Proposed work

- Frontend: widen the `filterableProperties` type whitelist.
- Backend: add facet branches for text / number / date / url (replacing the 422
  `default`), and extend the containment forms in `parsePropertiesFilterParam`
  so typed values (numbers, dates) match correctly.
- Tests: filter + facet coverage for each new type, including the `__none__`
  bucket.

### Acceptance criteria

- [ ] text / number / date / url properties appear in the issue filter menu.
- [ ] Filtering by a value returns issues with that exact value for each type.
- [ ] "No value" filtering works for each type (reusing `__none__`).
- [ ] Facet counts include a `__none__` bucket for each type.
- [ ] Round-trips through saved views.

### Related

- #7156 (merged) — the base "No value" filter feature.
- #7153 (closed) — the original feature request.

deoooo:
Closes #7235

## Summary

Extends the custom-property filter (value + "No value") from `select` /
`multi_select` / `checkbox` / `actor` to the four scalar types — `text`,
`number`, `date`, `url` — so the same filter menu and the `__none__` sentinel
work across every property type.

For a scalar property the filter menu now shows a **value input** (text / url,
number, or date) plus a mutually-exclusive **"No value"** option.

## What changed

**Backend**
- `parsePropertiesFilterParam` (`server/internal/handler/property.go`) emits a
  jsonb **number** containment form for numeric filter values, so a number
  property matches its stored scalar instead of only a `"3.5"` string.
  `NaN` / `Infinity` are guarded out — they are not JSON-representable and
  would otherwise 400 the whole filter.
- Property table facets (`issue_table_facets.go`) gain `text` / `url` / `date`
  (string) and `number` branches, replacing the previous 422 `default`, each
  with the unset `__none__` bucket so "No value" carries a real count.

**Frontend**
- `isFilterablePropertyType` (core) now admits the scalar types.
- `PropertyFilterOptions` renders the value input + "No value" for scalar
  types; both are disabled when a saved view locks the dimension. Setting a
  value clears "No value" and vice-versa.
- Filter chips display the scalar value directly.
- The client matcher compares number values numerically
  (`selected.some((id) => Number(id) === value)`), matching the server's jsonb
  number containment for forms like `"3.50"`.
- New `setPropertyFilterValues` store action replaces a property's full filter
  value set.

## Verification

- Backend: full `internal/handler` suite passes; new tests cover numeric +
  `NaN`/`Infinity`/date parse, scalar facet buckets (number / text / date /
  url), and numeric list filtering.
- Frontend: `@multica/views` issues suite (78 files / 779 tests) and core
  type/store tests pass; `tsc --noEmit` clean for `@multica/core` and
  `@multica/views` (pre-existing `@multica/plugin-sdk` resolution errors in
  views are unrelated).

## Notes

- A text property value literally stored as `"__none__"` is reserved: the
  sentinel means "no value", and the filter input rejects it as a literal
  value.
- The client-side `useIssueCounts` fallback path (used outside server facets)
  does not yet count number values / the `__none__` bucket for scalar types —
  a pre-existing limitation that predates this PR.

### DIFF
--- a/packages/core/issues/stores/view-store-property-filter.test.ts
+++ b/packages/core/issues/stores/view-store-property-filter.test.ts
@@ -1,0 +1,36 @@
+// @vitest-environment node
+import { describe, expect, it, beforeEach } from "vitest";
+import { createStore, type StoreApi } from "zustand/vanilla";
+import { viewStoreSlice, type IssueViewState } from "./view-store";
+
+// The "__none__" sentinel is a plain string option id from the store's
+// perspective; the filter menu (views) builds the array including it.
+const NO_PROPERTY_VALUE = "__none__";
+const PROP = "prop-estimate";
+
+describe("setPropertyFilterValues", () => {
+  let store: StoreApi<IssueViewState>;
+  beforeEach(() => {
+    store = createStore<IssueViewState>()((set) => viewStoreSlice(set));
+  });
+
+  it("replaces the property's full value set", () => {
+    store.getState().togglePropertyFilter(PROP, "3.5");
+    store.getState().setPropertyFilterValues(PROP, ["4"]);
+
+    expect(store.getState().propertyFilters[PROP]).toEqual(["4"]);
+  });
+
+  it("clears the property filter when given an empty set", () => {
+    store.getState().togglePropertyFilter(PROP, "3.5");
+    store.getState().setPropertyFilterValues(PROP, []);
+
+    expect(store.getState().propertyFilters[PROP]).toBeUndefined();
+  });
+
+  it("stores the no-value sentinel like any other value", () => {
+    store.getState().setPropertyFilterValues(PROP, [NO_PROPERTY_VALUE]);
+
+    expect(store.getState().propertyFilters[PROP]).toEqual([NO_PROPERTY_VALUE]);
+  });
+});
--- a/packages/core/issues/stores/view-store.ts
+++ b/packages/core/issues/stores/view-store.ts
@@ -245,6 +245,9 @@ export interface IssueViewState {
   toggleNoProject: () => void;
   toggleLabelFilter: (labelId: string) => void;
   togglePropertyFilter: (propertyId: string, optionId: string) => void;
+  /** Replace a property's full filter value set (used by scalar value inputs
+   *  for text/number/date/url, which build the array including "__none__"). */
+  setPropertyFilterValues: (propertyId: string, optionIds: string[]) => void;
   setDateFilter: (filter: IssueDateFilter | null) => void;
   toggleAgentRunningFilter: () => void;
   hideStatus: (category: IssueStatusCategory) => void;
@@ -388,6 +391,13 @@ export const viewStoreSlice = (set: StoreApi<IssueViewState>["setState"]): Issue
       const propertyFilters = { ...state.propertyFilters };
       if (next.length === 0) delete propertyFilters[propertyId];
       else propertyFilters[propertyId] = next;
+      return { propertyFilters };
+    }),
+  setPropertyFilterValues: (propertyId, optionIds) =>
+    set((state) => {
+      const propertyFilters = { ...state.propertyFilters };
+      if (optionIds.length === 0) delete propertyFilters[propertyId];
+      else propertyFilters[propertyId] = optionIds;
       return { propertyFilters };
     }),
   setDateFilter: (filter) => set({ dateFilter: filter }),
--- a/packages/core/types/index.ts
+++ b/packages/core/types/index.ts
@@ -105,7 +105,7 @@ export type { NotificationGroupKey, NotificationGroupValue, NotificationPreferen
 export type { Comment, CommentType, CommentAuthorType, CommentTriggerPreview, CommentTriggerPreviewAgent, CommentTriggerSource, CommentTriggerOutcome, CommentTriggerStatus, Reaction } from "./comment";
 export type { Label, LabelResourceType, CreateLabelRequest, UpdateLabelRequest, ListLabelsResponse, IssueLabelsResponse, ResourceLabelsResponse } from "./label";
 export type { IssueProperty, IssuePropertyType, IssuePropertyOption, IssuePropertyConfig, IssuePropertyValue, IssuePropertyValues, CreatePropertyRequest, UpdatePropertyRequest, ListPropertiesResponse, IssuePropertiesResponse, IssuePropertyActorKind, IssuePropertyActorRef } from "./property";
-export { ISSUE_PROPERTY_TYPES, isKnownPropertyType, ISSUE_PROPERTY_ACTOR_KINDS, MAX_ISSUE_PROPERTY_ACTOR_VALUES, isActorPropertyType, formatActorRef, parseActorRef, actorRefsFromValue, actorRefValuesFromValue, hasUnknownActorRef } from "./property";
+export { ISSUE_PROPERTY_TYPES, isKnownPropertyType, ISSUE_PROPERTY_ACTOR_KINDS, MAX_ISSUE_PROPERTY_ACTOR_VALUES, isActorPropertyType, isFilterablePropertyType, isScalarPropertyType, formatActorRef, parseActorRef, actorRefsFromValue, actorRefValuesFromValue, hasUnknownActorRef } from "./property";
 export type {
   QuickAction,
   QuickActionVisibility,
--- a/packages/core/types/property.test.ts
+++ b/packages/core/types/property.test.ts
@@ -1,3 +1,4 @@
+// @vitest-environment node
 import { describe, expect, it } from "vitest";
 import {
   actorRefsFromValue,
@@ -5,7 +6,9 @@ import {
   hasUnknownActorRef,
   formatActorRef,
   isActorPropertyType,
+  isFilterablePropertyType,
   isKnownPropertyType,
+  isScalarPropertyType,
   parseActorRef,
 } from "./property";
 
@@ -30,6 +33,39 @@ describe("isActorPropertyType", () => {
     expect(isActorPropertyType("multi_actor")).toBe(true);
     expect(isActorPropertyType("select")).toBe(false);
     expect(isActorPropertyType("multi_select")).toBe(false);
+  });
+});
+
+describe("isFilterablePropertyType", () => {
+  it("admits every type the filter menu supports", () => {
+    for (const type of [
+      "select",
+      "multi_select",
+      "checkbox",
+      "text",
+      "number",
+      "date",
+      "url",
+      "actor",
+      "multi_actor",
+    ]) {
+      expect(isFilterablePropertyType(type)).toBe(true);
+    }
+  });
+
+  it("rejects unknown types", () => {
+    expect(isFilterablePropertyType("relation")).toBe(false);
+  });
+});
+
+describe("isScalarPropertyType", () => {
+  it("covers the four single-valued types and nothing else", () => {
+    for (const type of ["text", "number", "date", "url"]) {
+      expect(isScalarPropertyType(type)).toBe(true);
+    }
+    expect(isScalarPropertyType("select")).toBe(false);
+    expect(isScalarPropertyType("checkbox")).toBe(false);
+    expect(isScalarPropertyType("actor")).toBe(false);
   });
 });
 
--- a/packages/core/types/property.ts
+++ b/packages/core/types/property.ts
@@ -62,7 +62,12 @@ export function isActorPropertyType(type: string): boolean {
   return type === "actor" || type === "multi_actor";
 }
 
-/** Types the issue filter menu exposes for value + "No value" filtering. */
+/**
+ * Types the issue filter menu exposes for value + "No value" filtering.
+ * Kept as an explicit enumeration (rather than delegating to
+ * isKnownPropertyType) so a future property type must opt into filtering —
+ * it should never become filterable by default.
+ */
 export function isFilterablePropertyType(type: string): boolean {
   return (
     type === "select" ||
--- a/packages/core/types/property.ts
+++ b/packages/core/types/property.ts
@@ -62,6 +62,22 @@ export function isActorPropertyType(type: string): boolean {
   return type === "actor" || type === "multi_actor";
 }
 
+/** Types the issue filter menu exposes for value + "No value" filtering. */
+export function isFilterablePropertyType(type: string): boolean {
+  return (
+    type === "select" ||
+    type === "multi_select" ||
+    type === "checkbox" ||
+    isScalarPropertyType(type) ||
+    isActorPropertyType(type)
+  );
+}
+
+/** Single-valued scalar properties: text / number / date / url. */
+export function isScalarPropertyType(type: string): boolean {
+  return type === "text" || type === "url" || type === "number" || type === "date";
+}
+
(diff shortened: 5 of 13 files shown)


deoooo:
Hi @Bohan-J, @forrestchang — this PR extends the custom-property filter (value + "No value") to text / number / date / url, completing the coverage follow-up from #7156 (tracked in #7235). All CI checks are green (backend-tests, frontend, frontend-views-test, sqlc-check), and the change is covered by new backend handler + unit tests and frontend filter tests. When you have a chance, could you please review? Thanks!

Bohan-J (changes requested):
Thanks for taking this on — the direction is right and the backend side fits the existing design well. There is one blocker that makes the main interaction unusable in a real browser, plus two things I'd like fixed in the same pass.

## Blocking

### 1. The scalar value input cannot be typed into

`packages/views/issues/components/issues-header.tsx` (the new `<Input>` in `PropertyFilterOptions`)

Base UI's menu popup owns typeahead and arrow-key list navigation. Both handlers are merged into `popupProps` and land on the `role="menu"` element (`menu/root/MenuRoot.js` merges `typeahead.floating` and `listNavigation.floating`). The new `<Input>` is a descendant of that popup, so its keydown events bubble into those handlers, which call `stopEvent()` — `preventDefault()` + `stopPropagation()` — on every printable key while the menu is open.

Net effect: for `text` / `url` the input accepts no characters at all, and for `number` / `date` the arrow keys don't work either.

I reproduced it against the version this repo pins, `@base-ui/react@1.3.0`, with a minimal `Menu.Popup` containing one `input` and one `Menu.CheckboxItem`, driven by jsdom + `@testing-library/user-event`:

| | click input, then type `hello` |
|---|---|
| as in this PR | focus lands on the input, but `input.value === ""` |
| with `onKeyDown` stopPropagation on the input | `input.value === "hello"` |

Per-key `defaultPrevented` on a keydown dispatched from the input: `h`, `5`, `ArrowDown`, `ArrowUp`, `Home`, `End` are all `true` as-is, all `false` with the guard.

(The measurement is jsdom, but the mechanism isn't jsdom-specific — a `preventDefault()`ed keydown does not insert a character in any browser.)

Suggested fix:

```tsx
onKeyDown={(event) => {
  if (event.key === "Escape" || event.key === "Tab") return; // let the menu close / move focus
  if (event.key === "Enter") commitValue(draft);
  event.stopPropagation();
}}
```

`Enter` needs the `stopPropagation()` too, otherwise it bubbles up and activates the highlighted "No value" item.

### 2. The new scalar facet branches have no upper bound

`server/internal/handler/issue_table_facets.go` (the `text` / `url` / `date` case and the new `number` case)

For `select` / `actor` the group cardinality is naturally bounded by the option list or the member directory. The four scalar types are free-form, so `GROUP BY` yields one row per distinct value in the current scope, with no `LIMIT` — and the menu only ever reads the `__none__` bucket for these types (`noneCount` in `PropertyFilterOptions`). Every other row is queried, sorted by `sort.Slice`, serialized, and thrown away. On a workspace with tens of thousands of issues carrying a free-text property, opening that submenu pulls a response proportional to the issue count.

Either:

- compute only the two buckets the UI consumes:
  ```sql
  SELECT CASE WHEN i.properties ? <key> THEN '__set__' ELSE '__none__' END, COUNT(*)::bigint
  FROM issue i WHERE ... GROUP BY 1
  ```
- or keep real values but `ORDER BY 2 DESC LIMIT n`, always unioning the `__none__` bucket — which is the shape you'd want anyway if the menu later lists observed values (see the product note below).

### 3. The new UI branch has no test coverage

The three new frontend tests cover the store action, the type predicates, and the pure matcher — all straight-line logic. The genuinely new surface (an input hosted inside a menu popup, the blur-vs-click commit ordering, the value ↔ "No value" exclusion) has none, which is why CI is green despite issue 1.

`packages/views/issues/components/issues-header.status-filter.test.tsx` is an existing precedent for driving this menu in a test. A `userEvent.type` assertion there fails today and would have caught it.

## Non-blocking

- `onSetValues` is optional, but the single call site always passes it. If it's ever omitted, the scalar branch silently degrades to an option list containing nothing but "No value". Make it required.
- The `number` facet keys by `->>`, i.e. the stored jsonb numeric's literal text: a value stored as `3.50` produces the key `"3.50"` while the UI sends `"3.5"`. Harmless today since only `__none__` is read, but it needs normalizing if issue 2 is fixed by keeping real values.
- `draft` initializes from `selected` on mount only. If the filter is cleared elsewhere while the submenu is open, `draft` is stale and unchecking "No value" writes the old value back. Narrow, but an effect sync or keying the input on `scalarValue` closes it.
- `isFilterablePropertyType` now covers all nine members of `ISSUE_PROPERTY_TYPES`, so it is equivalent to `isKnownPropertyType`, and its "rejects unknown types" test only demonstrates that. Keeping the explicit enumeration is reasonable (a future property type shouldn't become filterable by default) — worth saying so in the comment.

## What's good

- The blur guard using `relatedTarget.closest('[role="menuitemcheckbox"]')` is the correct fix for blur firing before the item's click. I confirmed in the same minimal reproduction that `relatedTarget` is indeed the `menuitemcheckbox` element on click.
- The `NaN` / `Infinity` guard addresses a real failure — those would otherwise 400 the entire filter.
- The numeric containment form extends `parsePropertiesFilterParam`'s existing "expand a value into every form it could match" design rather than bolting on a parallel path.
- Date properties are written as date-only strings (`toDateOnly`), so exact string comparison avoids the timezone trap that filtering dates usually invites.

## Product questions (not blockers, but worth settling before merge)

1. **Exact equality only.** For a text property, users will almost certainly type a fragment and get an empty list with no explanation. For number and date, the usual ask is a range (`> 3`, before a date). Shipping equality-only in v1 is a defensible trade-off — it should just be an explicit decision rather than an omission.
2. **The input starts empty**, so there's no way to discover which values the property actually holds. For `text` / `url`, listing the observed values as checkboxes (the way select properties work) would be considerably more usable — and the facet response already carries exactly that data today, it's just discarded.

Happy to re-review once 1 is fixed.

deoooo:
Thanks @Bohan-J for the thorough review — all three blockers plus the non-blocking items are addressed in `6f9e65b40`:

**1. Scalar input swallowing characters** — the input's `onKeyDown` now calls `event.stopPropagation()` for every key except `Escape`/`Tab` (Enter still commits, and now also stops propagation so it can't activate the highlighted "No value" item). Confirmed with a component test that types into the input against the real Base UI menu.

**2. Unbounded scalar facets** — the `text`/`url`/`date`/`number` branches now compute only the bounded `__set__`/`__none__` buckets (key-existence based), since the menu only reads the "No value" count for these types.

**3. Missing UI test coverage** — added `issues-header.property-filter.test.tsx`, modeled on the status-filter precedent, driving the real menu primitives: typing works, Enter/blur commit, the value ↔ "No value" exclusion, and the blur-before-click ordering on uncheck.

Also folded in the non-blocking items: `onSetValues` is now required; the draft syncs to the committed value via an effect (keying the input proved unreliable at resetting it), so an external clear can't be written back on uncheck; and a comment documents that `isFilterablePropertyType` is an explicit enumeration so future property types opt into filtering.

On the two product questions — exact-equality-only and the empty input not showing observed values — I've left both as documented v1 trade-offs (the facet now discards per-value data, so listing observed values later would be a follow-up change). Happy to revisit either if you'd like them in v1.

Full backend handler suite, `@multica/views` issues suite (79 files / 784 tests), and core/views typecheck all pass locally. When you have a chance, could you re-review? Thanks!

deoooo:
Hi @Bohan-J, gentle nudge on this one — all three blockers from your review were addressed in `6f9e65b40` last week (input keydown guard, bounded `__set__`/`__none__` facets, plus a component test that pins the input regression). CI is fully green and the branch merges cleanly with current main. Could you take another look when you get a chance? Thanks!

Bohan-J (changes requested):
Re-reviewed at `6f9e65b4`. The three blocking items from the last round are genuinely fixed, and so are all four nits — thanks for the thorough pass.

- The keydown guard is in place and the input now receives characters inside the menu popup, with `Escape`/`Tab` still left to the menu.
- The scalar facet is now bounded to the two `__set__` / `__none__` buckets the menu actually consumes.
- `issues-header.property-filter.test.tsx` drives the real menu, which is the right place for this.
- `onSetValues` is required, the `isFilterablePropertyType` comment states the intent, and the `->>` key-shape nit is moot now that the facet is bucketed.

One regression came in with the draft-staleness fix, and there's a design question underneath it.

## Blocking: unchecking "No value" discards a committed value

```
Open Filter → a text property
1. type `hello`, press Enter   → filter = ["hello"]      ✅
2. check "No value"            → filter = ["__none__"]   ✅
3. uncheck "No value"          → filter = []             ❌ expected ["hello"]
```

The inline comment says "Unchecking 'No value' restores the value still in the input", and the test is named "…unchecking restores the draft", but on the committed path the value is gone.

Root cause — the new `useEffect(() => setDraft(scalarValue), [scalarValue])` in `PropertyFilterOptions`:

1. `commitValue` puts `["hello"]` in the store, so `scalarValue === "hello"`.
2. Checking "No value" replaces the whole set with `["__none__"]`, so `scalarValue` recomputes to `""`.
3. That transition fires the effect, which wipes `draft`.
4. Unchecking reads `draft` (now `""`) and commits `[]`.

The uncommitted path survives only because `scalarValue` stays `""` throughout, so the effect never fires — which is exactly why the current tests pass. In `issues-header.property-filter.test.tsx`, the case named `checking No value clears a committed value; unchecking restores the draft` re-types `world` before unchecking, so the "restores" half of its name is never exercised against the committed value; the following case covers only the uncommitted path. Worth adding step 1→3 above verbatim as the regression test.

## The design question underneath it

Every other property type treats the values within one property as OR, with "No value" as just another member of that set — `togglePropertyFilter` adds it alongside real values, `parsePropertiesFilterParam` ORs the alternatives with the key-absence marker inside the same group, `issueMatchesPropertyFilters` agrees, and the chips bar renders the combination. Scalars are the only type where the two are mutually exclusive: `commitValue` drops `__none__`, and checking "No value" drops the value.

That's what forces the fragile draft round-trip in the first place. If a scalar could hold `["hello", "__none__"]` like every other type, "restore on uncheck" stops being a thing that has to be engineered — unchecking just removes one member.

It also has a concrete consequence today: a saved view or URL carrying `["hello", "__none__"]` for a scalar property is a valid state that the server, the store, and the chips bar all handle, and the menu renders it correctly (input filled, box checked) — but the first edit silently drops whichever half you didn't touch.

If XOR is a deliberate choice for scalars, that's a defensible call — please say so in the comment and make editing not silently discard a pre-existing `__none__`, and rename the test to match what it asserts.

## Non-blocking: `__none__` collides with a legal text value

You flagged this in the description, and I'm not asking you to redesign the wire format for it — but there's a cheap piece worth taking now, because the two sides currently disagree rather than merely being limited.

`validatePropertyValue` accepts the literal string `"__none__"` as a valid `text` value (any non-empty string within the length cap). The filter input refuses to commit it and the server reads it as key-absence, so it can't be filtered for — that's the disclosed limitation. But `issueMatchesPropertyFilters` compares string values with plain `selected.includes(value)`, so on the client-side matching path an issue whose text property is literally `"__none__"` **matches a "No value" filter**, while the server's key-absence predicate correctly excludes it. Same filter, two different answers depending on which path serves the list.

The number branch right beside it already guards this (`id !== NO_PROPERTY_VALUE`); the string branch needs the same treatment. Making the sentinel actually filterable can be a follow-up.

Everything else looks good to me — the OR/design point is the one I'd like a decision on before this goes in.

deoooo:
Great catch on the regression — the draft round-trip was exactly as fragile as you described. Fixed in `6f03d6918`:

**Design decision: OR.** Scalars now treat "No value" as a plain member of the same OR-set as every other type, so the restore-on-uncheck machinery is gone entirely:

- Checking/unchecking the checkbox toggles membership only and never touches the committed value member — unchecking `["hello", "__none__"]` restores `["hello"]` by construction.
- Committing a value replaces just the value member and preserves existing No-value membership (`"abc{Enter}"` with No-value checked → `["abc", "__none__"]`).
- The mutual-exclusion comment is removed; a saved view/URL carrying `["hello", "__none__"]` now survives edits instead of silently dropping half.

**The regression test** — added verbatim per your sequence in `issues-header.property-filter.test.tsx`: commit `hello` → check (expects `["hello", "__none__"]`) → uncheck (expects `["hello"]`). Also covered: committing preserves existing membership, and an uncommitted draft survives checking then commits alongside it.

**String-branch sentinel guard** — `issueMatchesPropertyFilters`' string branch now skips the sentinel when matching stored values (`id !== NO_PROPERTY_VALUE && id === value`), same guard the number branch already had, so a literal `"__none__"` text value can no longer satisfy a No-value filter on the client path while the server excludes it.

Views suite (79 files / 786 tests) and typecheck pass locally; CI is running on the new commit. Ready for another look whenever suits you — thanks again for pushing on the OR point, it made this much simpler.

Bohan-J (approved):
Approving — this is ready to go in.

All three findings from the last round are fixed at the root rather than patched, which is the part worth calling out:

- Composing the scalar value and "No value" OR-style didn't just fix the "unchecking discards the committed value" bug — it made that class of bug structurally impossible, since the checkbox no longer round-trips through `draft` at all. That's the better fix, and it also brings scalars in line with how every other property type behaves.
- The `NO_PROPERTY_VALUE` guard on the string branch means the client matcher and the server's key-absence predicate now agree on a literal `"__none__"` value.
- The tests assert the exact sequences rather than the ones that were easy to make pass — `hello` → `["hello", "__none__"]` → `["hello"]`, committing into an existing No-value membership, and the uncommitted-draft path.

Thanks for staying with this across three rounds — the keyboard fix, the bounded facet buckets, and the OR composition each landed cleanly, and the inline comments explaining *why* (the Base UI popup swallowing keys, the blur-vs-click ordering) will save the next person real time.

The two documented boundaries — exact equality for v1, and a literal `"__none__"` not being filterable as a value — are fine as follow-ups.

Issue #3647, opened 2026-06-02
Startup DB migrator has no locking — concurrent backend replicas race on `migrate up`


andrewjeffree:
## Summary

The backend runs database migrations on every startup, but the migrator takes no lock and has no guard against concurrent runners. If more than one backend pod starts at the same time (a multi-replica Deployment, a rolling restart, or a scale-up), they race on `migrate up` and one will crash. This isn't mentioned in the self-hosting docs — the only acknowledgement is an inline comment in the Helm chart pinning `replicas: 1`.

## Details

The entrypoint runs `./migrate up` on every backend start — [`docker/entrypoint.sh#L5`](https://github.com/multica-ai/multica/blob/main/docker/entrypoint.sh#L5).

The migrator is a plain check-then-apply loop with no advisory lock or transaction around the run — [`server/cmd/migrate/main.go#L44-L116`](https://github.com/multica-ai/multica/blob/main/server/cmd/migrate/main.go#L44-L116). For each migration it does `SELECT EXISTS(... version ...)`, and if absent, executes the SQL and inserts the version row.

With two pods booting concurrently:

- both can pass the `EXISTS` check for the same migration before either records it, then both execute the DDL — the second fails (e.g. `relation already exists`, duplicate column/index), or
- both reach the `INSERT INTO schema_migrations (version)` and the second hits a primary-key violation.

Either way the migrator exits non-zero and the pod crash-loops.

The Helm chart works around this implicitly by hardcoding a single replica and a `Recreate` strategy — [`backend.yaml#L25-L30`](https://github.com/multica-ai/multica/blob/main/deploy/helm/multica/templates/backend.yaml#L25-L30) — but that's a comment, not something a user scaling the Deployment would discover until pods start failing.

## Impact

The backend can't be safely run as a multi-replica Deployment, which undercuts the horizontal-scaling story that the Redis realtime relay (`REDIS_URL`) otherwise enables.

## Suggested fix

Any of:

1. Wrap the migration run in a Postgres session-level advisory lock (`pg_advisory_lock`) so concurrent runners serialize and late arrivals no-op once migrations are applied. Smallest change, makes `migrate up` safe to run from every pod.
2. Move migrations out of the per-pod startup path into a single-shot Helm `pre-install`/`pre-upgrade` Job (or an init step that elects one runner), so app pods never run migrations.
3. At minimum, document that the backend must run a single replica during migration / that scaling requires running migrations separately.

## Docs gap

[...]

PR #7241, opened 2026-08-19
MUL-6426: refactor(dingtalk): safely retire group routing and enhance discovery

### LINKED ISSUE #7239: [Feature]: Safely retire DingTalk group routing and improve group discovery
## Deployment type

Official App

## What do you want and why?

Safely retire the mutable DingTalk group-routing model introduced in #6660 and replace the coupled routing/discovery design with a clear product relationship:

1. An Agent owns a dedicated DingTalk app/Bot installation.
2. The Bot can be observed in multiple DingTalk groups.
3. Those groups are read-only membership inventory and never select the message target.

Group routing made the target Agent depend on mutable per-group state and mixed routing authority, discovery metadata, management permissions, and inventory visibility. Removing only the route table would leave the Agent/Bot/Group model and its permissions ambiguous, so the replacement must be designed end to end.

### Safely remove group routing

- Keep the installation's configured Agent as the sole target for DingTalk messages.
- Remove group-route APIs, realtime events, reassignment, route revisions/fences, and route-specific retry behavior.
- Clean stale cross-Agent DingTalk session bindings and outbound cards while retaining the legacy route table, indexes, and a mirror trigger for adjacent-release rollback compatibility.
- Retain the historical route-creation migrations. Under the documented Helm `Recreate` contract, remove the trigger and table together in a later contract release after this release has removed every runtime dependency.

### Rebuild the Agent/Bot/Group model

- Retain existing discovered group titles as historical entries. Since legacy routes contain no trustworthy processing time or count, backfilled groups start under the collapsed **Long inactive** section until a later accepted @mention records real activity.
- Add the exact Bot identity observed for an installation and app/installation-scoped identity permission state.
- Add per-Bot/group last successfully processed @mention time and processed mention count.
- Expose the relationship as Agent → dedicated Bot → discovered Groups through workspace and Agent-scoped APIs.
- Treat discovery as an explicit backend capability so older or malformed backends degrade safely.

### Redesign permissions

- Agent owners and workspace owners/admins can connect or disconnect that Agent's Bot; orphaned installations are workspace owner/admin-only.
- Anyone who can open an Agent can view that Agent's Bot and discovered groups.
- Workspace owners/admins see the full inventory; regular members see only Bots/groups belonging to Agents they can open and do not receive linked-account identifiers. Forgetting historical group inventory is workspace owner/admin-only, while Agent owners retain connect/disconnect authority for their Agent's Bot.
- A workspace member must link their own DingTalk account before an @mention can discover a group or reach the Agent.
- `qyapi_chat_manage` remains optional for messaging/discovery. Its warning is installation-scoped, concurrent lookups share only successful identity or proven app-scoped denial, group-specific failures retry per group, transient failures preserve prior state, and a later successful Bot lookup clears stale denial state.

### Align the documentation

Manually review and rewrite the English, Simplified Chinese, Japanese, and Korean DingTalk MDX guides against the final code. Remove ambiguous or mechanical phrasing and document the actual linking, permission, discovery, visibility, command, image, installation, and self-hosting behavior.

## Why one larger PR?

Routing removal, discovery storage, API capability negotiation, permission filtering, UI hierarchy, and documentation all depend on the same installation contract. Splitting them would expose an incomplete intermediate model or require temporary compatibility paths. They should therefore land together even though the resulting PR is larger.

The scope remains mostly DingTalk-internal. No Lark, Slack, WeCom, or Telegram adapter behavior should change. Shared engine work is limited to removing the DingTalk-routing seams introduced by #6660 and preserving channel-agnostic dedup durability; it must not change routing, session ownership, or user-visible behavior for other channels.

## Screenshots / mockups

Agent Detail and workspace Settings screenshots are attached to the implementation PR.

yyclaw:
## What does this PR do?

This PR safely retires the mutable DingTalk group-routing model introduced in #6660 and rebuilds the DingTalk integration around a clearer product model: **an Agent owns a dedicated Bot installation, and the groups where that Bot has been observed are read-only membership inventory**.

The change is intentionally broader than deleting a route table. Group routing had coupled target selection, session ownership, discovery metadata, API shapes, permissions, UI hierarchy, and documentation. Removing only the routing code would have left the Agent/Bot/Group relationship and its access rules ambiguous. This PR therefore handles the removal and the replacement model end to end.

Although the diff is large, its product scope is mostly contained within DingTalk. No Lark, Slack, WeCom, or Telegram adapter behavior is changed. The shared channel-engine changes remove extension points added specifically for DingTalk group routing and keep dedup finalization durable after request cancellation; they do not change target selection, session ownership, or user-visible behavior for existing non-DingTalk channels.

### 1. Safely retire group routing

DingTalk messages now always target the Agent configured on the Bot installation. A discovered group cannot select or reassign the target Agent.

- Remove the group-route API, realtime event, reassignment flow, route revision/fence, and route-specific retry behavior.
- Remove the post-identity shared-engine resolver introduced for mutable DingTalk routing; DingTalk group discovery remains a best-effort adapter concern after the addressing and membership gates.
- Preserve the stable installation-based inbound messaging path.
- Remove stale cross-Agent DingTalk session bindings and their outbound cards while retaining deprecated route storage for adjacent-release rollback compatibility.
- Keep historical migrations 304–307; migration 382 removes stale cross-Agent bindings/cards, and migrations 383–389 create and backfill the new group-presence and installation-owned Bot-identity tables. The legacy route table, indexes, and mirror trigger remain in this release. Under the documented Helm `Recreate` deployment contract, a later contract release can drop the trigger and table together after this release has removed every runtime dependency; arbitrary rolling upgrades directly from older releases are not the supported contract.
- Migration 382 is intentionally irreversible: its down migration does not recreate stale cross-Agent bindings/cards. Rolling back to the old runtime makes those affected groups start fresh sessions and lose prior conversation continuity, while the historical sessions and messages remain stored.
- Cover the deleted endpoint/event and the installation-Agent invariant with regression tests.

### 2. Redefine the Agent, Bot, and Group relationship

The new model is explicit:

1. An Agent is connected to one dedicated DingTalk app/Bot installation.
2. That Bot can be observed in multiple DingTalk groups.
3. A group record describes where the Bot has been seen; it never changes where messages are routed.

The implementation retains the existing discovered group-title data. Because legacy routes do not prove when Multica last processed a message, backfilled groups start with `last_active_at = NULL` and `mention_count = 0`. They are counted under the collapsed **Long inactive** section until a later accepted @mention records real activity. The rest is newly designed discovery capability:

- record the exact DingTalk Bot identity observed for the installation;
- record app/installation-scoped Bot identity permission state;
- persist, per Bot and group, when an @mention was last successfully processed and the processed mention count, avoiding message-history aggregation on every inventory read;
- show groups active within 90 days by default, load long-inactive groups in pages of 20, merge loaded pages into one deterministic display order, and let workspace owners/admins forget stale entries without deleting conversations or messages;
- expose an explicit Agent → dedicated Bot → discovered Groups hierarchy through workspace and Agent-scoped APIs;
- negotiate discovery through `group_discovery_supported === true`, so older or malformed backends degrade without polling or inventing partial state.

### 3. Redesign the permission model

The old routing UI mixed management authority with inventory visibility. This PR reviews those permissions separately:

- **Connect or disconnect a Bot:** the Agent owner, or a workspace owner/admin, can manage the Agent's installation. Orphaned installations remain workspace owner/admin-only.
- **Agent Detail:** anyone who can open an Agent can see that Agent's Bot and discovered groups; users cannot infer groups for an Agent they cannot access.
- **Workspace Settings:** owners/admins see the complete Agent → Bot → Groups inventory. Regular members see only installations and groups belonging to Agents they can open, and do not receive linked-account identifiers. Forgetting historical group inventory is intentionally owner/admin-only; Agent ownership grants installation management, not workspace-wide inventory cleanup in Settings.
- **Discover a group:** a workspace member must link their own DingTalk account and then @mention the Bot in that group. Messages rejected before account linking are not handed to the Agent, and their bodies are not stored.
- **Resolve Bot identity:** `qyapi_chat_manage` is optional for messaging and group discovery. Its denial is treated as app/installation-scoped even when observed in one group and cached for 30 seconds to avoid one OpenAPI call per message. Concurrent lookups share only a successful identity or that proven app-scoped denial; a group-specific failure is retried with the requesting group's conversation ID. Transient lookup failures preserve the last known state, while a later successful lookup clears the stale permission warning.

### 4. Manually refine the DingTalk documentation

The English, Simplified Chinese, Japanese, and Korean MDX guides were reviewed against the final code paths and manually rewritten for concision, clarity, and natural phrasing. This was not a mechanical translation pass. The guides now explain account linking, permission scope, discovery, visibility, `/issue`, `/new`, images, installation replacement, and self-hosted credential storage using the behavior the code actually implements.

## Why this is one larger PR

Routing removal, the Agent/Bot/Group model, permission filtering, API capability negotiation, UI hierarchy, and documentation share the same installation and discovery contracts. Splitting them would either expose an intermediate ambiguous product model or require temporary compatibility paths that the project does not need. Keeping them together makes the migration atomic at the product level while still limiting almost all behavior changes to DingTalk.

## Related Issue

Closes #7239

## Type of Change

- [x] Refactor / code improvement
- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Remove DingTalk group-route handlers, realtime invalidation, database queries, reassignment, and route-specific channel resolution.
- Add migration 382 to clean stale cross-Agent bindings/cards, and migrations 383–389 to create/backfill group presence and installation-owned Bot identity while retaining legacy group-route storage for compatibility.
- Keep retained legacy route rows in explicit installation, system-Agent, and workspace teardown so compatibility storage cannot outlive its owner.
- Retain discovered group titles as historical inactive entries, and add installation-owned Bot identity, app-scoped permission state, persisted last processed @mention time, and processed mention counts for new accepted activity.
- Add active/inactive inventory paging and an admin-only forget action; a forgotten group reappears after the next accepted @mention.
- Add workspace and Agent-scoped group inventory APIs with explicit discovery capability negotiation.
- Rebuild Settings and Agent Detail around the Agent → dedicated Bot → discovered Groups hierarchy and the new permission model.
- Manually refine all four DingTalk MDX locales and related UI copy against current code behavior.
- Keep non-DingTalk channel behavior unchanged and add shared-engine/DingTalk regression coverage for the removed routing seams.

## How to Test

1. Connect a DingTalk Bot to an Agent and link a workspace member's own DingTalk account.
2. @mention the Bot in two DingTalk groups; verify both appear under that Agent with group title, conversation ID, installation-owned Bot identity or its permission warning, last processed time, and processed count.
3. Verify Agent Detail exposes only the selected Agent's Bot/groups to users who can open that Agent.
4. Verify Settings owners/admins see the complete hierarchy, while regular members see only Agents they can access and receive no linked-account identifiers.
5. Verify groups with activity in the last 90 days load by default; backfilled legacy groups appear under the collapsed **Long inactive** section with no fabricated activity, load 20 at a time, and remain globally ordered after loading more.
6. As a workspace owner/admin, forget a group; verify conversations/messages and the installation Bot name remain intact, then @mention the Bot there again and verify the group reappears.
7. Verify missing `qyapi_chat_manage` shows an installation-level identity warning without blocking messaging or group discovery; granting it and successfully resolving the Bot clears the stale warning after the short negative-cache window.
8. Verify a response without `group_discovery_supported === true` renders the compatibility state and does not poll.
9. Verify the removed `/dingtalk/group-routes` endpoint returns 404 and every DingTalk message continues to target the installation's Agent.

Validation completed locally on the rewritten PR head after numbering migrations against `upstream/main` (latest migration 376; this PR starts at 382):

- `make sqlc`
- `make migrate-up`
- `go test -p 1 ./internal/handler ./internal/integrations/dingtalk ./internal/migrations ./cmd/migrate ./internal/integrations/channel/engine -count=1`
- 56 targeted Vitest cases across Agent Integrations and DingTalk Settings
- targeted `go test -race` coverage for cross-group Bot-identity singleflight
- `pnpm --filter @multica/docs typecheck`

## Checklist

- [x] I have included the architectural reasoning for combining the related changes.
- [x] I have run relevant tests locally and they pass.
- [x] I have updated relevant documentation to reflect the final code behavior.
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx`.
- [x] I have considered the effect on other channel adapters and the shared channel engine.
- [x] I will address all reviewer comments before requesting merge.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** The implementation was iteratively reviewed against the intent of #6660 and the latest code. The final pass audited shared-engine provenance, redesigned the DingTalk permission and information model, manually refined all MDX locales, resolved the latest-main conflicts, renumbered migrations against the current migration set, and reran targeted backend/frontend validation.

### Agent Detail

image
image
image

### Settings — Workspace Admin

image

### Settings — Workspace Member

image


Bohan-J:
Thanks for the thorough write-up — reviewed at `d0f56e16e`.

On the product side, I agree with the direction: one Bot installation owning one Agent, with discovered groups as read-only inventory, is a clearer model than mutable per-group routing, and the "a group can never reassign the target Agent" invariant is the right thing to enforce. The permission split (Agent owner / workspace admin for connect-disconnect, Agent read access for visibility) also looks right.

Two things block merge, both around the migration path rather than the model itself.

## 1. The upgrade drops every already-discovered group

On `main`, `group_titles` does not exist anywhere in `channel_installation.config` — discovered group identity (`conversation_id` / `conversation_title`) lives **only** in `dingtalk_group_route`. The new read path expands `channel_installation.config -> 'group_titles'` (`server/pkg/db/queries/dingtalk.sql:152`), migration 381 only deletes cross-Agent session bindings (`server/migrations/381_remove_dingtalk_group_routing_bindings.up.sql:6`), and 385 drops the table (`server/migrations/385_drop_dingtalk_group_route.up.sql:1`). No step in between moves the data across.

So this is not a "might lose data" case — every existing deployment's group list goes empty on upgrade and stays empty until each group @-mentions the Bot again. That contradicts "The implementation retains the existing discovered group-title data" in the PR description.

A deterministic backfill before the drop would fix it, roughly:

```sql
UPDATE channel_installation ci
SET config = jsonb_set(
        ci.config,
        '{group_titles}',
        routes.titles || (
            CASE WHEN jsonb_typeof(ci.config -> 'group_titles') = 'object'
                 THEN ci.config -> 'group_titles' ELSE '{}'::jsonb END
        ),
        true
    ),
    updated_at = now()
FROM (
    SELECT installation_id,
           jsonb_object_agg(conversation_id, conversation_title) AS titles
    FROM dingtalk_group_route
    WHERE conversation_id <> ''
    GROUP BY installation_id
) routes
WHERE ci.id = routes.installation_id
  AND ci.channel_type = 'dingtalk';
```

Putting the existing map on the right-hand side of `||` keeps it idempotent — a re-run never overwrites a fresher title written by the new code.

Please also extend the migration test. It currently applies 304–307 to create the table but never inserts a single route row, and its `channel_installation` fixture has no `config` column at all (`server/internal/migrations/dingtalk_group_routing_removal_migration_test.go:90`), so it cannot catch this class of bug. Seeding old route rows and asserting the titles survive would.

## 2. Dropping the table in the same release is not rolling-deploy or rollback safe

Every DingTalk group message on `main` reads and writes that table — `DiscoverDingTalkGroupRoute`, `DingTalkGroupRouteMatchesAgent`, and `LockDingTalkGroupRouteForAppend` (`server/internal/integrations/dingtalk/resolvers.go:189`, `:328`, `:383` on `main`). Once 385 runs, any instance still on the old build fails with `relation "dingtalk_group_route" does not exist` on every group message for the rest of the rollout. Rollback does not recover either: the down migration can only recreate an empty table (`server/migrations/385_drop_dingtalk_group_route.down.sql:3`).

Expand/contract would make this safe as two releases:

- **This PR:** 381 plus the backfill above. Stop depending on the table, but keep it and its indexes (move 382–385 out). A rollback to the previous build still works because the old data is intact.
- **A follow-up:** once old instances are gone and the rollback window has closed, drop the indexes and the table.

One related note worth stating explicitly in the PR description: 381 itself is irreversible (its down is `SELECT 1;`). After it runs, rolling back to the old build means previously cross-Agent groups start a fresh session and lose conversation continuity. That trade-off seems acceptable, but it should be documented rather than implicit.

## Non-blocking, but worth addressing

**Group inventory read cost grows with message history.** `last_active_at` and `mention_count` are computed per request by a LATERAL aggregate that scans every `role='user' AND channel_ingested` message in the bound session (`server/pkg/db/queries/dingtalk.sql:159-161`), and both surfaces poll every 5 seconds (`packages/core/dingtalk/queries.ts:31`, `:45`). Agent Detail is the worse case: the handler fetches the whole workspace inventory and filters by Agent in Go (`server/internal/handler/dingtalk.go:267`, `:277`), so any member who can open one Agent triggers a full-workspace aggregation every 5s. Two suggestions: push the Agent filter into SQL, and maintain the count and last-active timestamp as counters inside `RecordDingTalkGroupPresence` — that statement already runs per message, so incrementing there is nearly free and turns the read into O(groups) instead of O(message history).

**The group maps only ever grow.** Once a Bot leaves a group, its entries in `group_titles` / `group_bot_names` / `group_bot_identity_issues` stay forever, with no removal path in the UI. The docs wording ("groups where it has successfully processed messages") is accurate, so this is not misleading, but the inventory will accumulate noise over time. Worth either bounding the semantics explicitly or giving admins a manual remove action.

**Permission denials are never cached.** `internal/integrations/dingtalk/bot_identity.go:146-148` deliberately skips the cache on `qyapi_chat_manage` denial so a fresh grant takes effect on the next mention. That intent makes sense, but a missing optional permission is a steady state rather than a transient failure, so every group message on such an installation makes an OpenAPI call. A short TTL (30–60s) would preserve the grant-and-mention retry experience without amplifying calls indefinitely.

## What works well

- The old-client compatibility direction is handled correctly: the new handler simply stops returning `group_routing_supported`, so an older client's `=== true` check turns false and the whole routing section hides instead of surfacing a load error. That direction is easy to miss.
- `RecordDingTalkGroupPresence` guards the write with `WHERE config IS DISTINCT FROM next_config`, so steady-state traffic does not rewrite the installation row on every message.
- Dedup finalization detaching from request cancellation with a bounded timeout, plus the cancellation regression test, is the right fix for the shared engine.
- Permission filtering is enforced server-side with negative tests covering private / non-member / orphaned-installation cases, rather than hidden client-side.
- Migration numbering does not collide with `main`, all index changes use `CONCURRENTLY`, and no foreign keys or cascades were introduced.

## Verification

- `go build ./...`
- `go test ./internal/integrations/channel/engine ./internal/integrations/dingtalk ./internal/migrations -count=1`
- `go test ./internal/handler -run 'Test(ListDingTalk|RegisterDingTalk|RevokeDingTalk|PublishDingTalk|RedeemDingTalk)' -count=1`
- `make sqlc` produces no diff in `server/pkg/db/generated` (no codegen drift)
- All 16 CI checks are green, including `frontend-test` and `sqlc-check`. Frontend suites were not re-run locally.

## Verdict

Requesting changes — items 1 and 2 are data loss and deploy safety and need to be resolved before merge. Item 3 is a strong recommendation; without it this page becomes a database hotspot as message volume grows. The product model itself is good, and I'd like to land it once the migration path is safe.

yyclaw:
@Bohan-J Thanks for digging into the migration path. A few thoughts and questions on the two blockers.

On the missing group-title backfill: agreed. As written, the migration does not retain the existing discovered groups, so the PR description currently overstates that property. I will add a deterministic backfill before any legacy storage is retired, seed legacy route rows in the migration test, and assert that the titles survive. That part is a real gap.

On expand/contract: I had considered rolling deploy and rollback compatibility before choosing the current clean cut. The trade-off was that carrying the old table forward also carries a storage model whose only product meaning was mutable routing, while the replacement deliberately makes groups read-only observations. I initially chose the clean implementation both to keep that boundary unambiguous and to get feedback on what compatibility window we actually want to support.

The part I am not fully convinced a two-release split solves by itself is the open-source/self-hosted case. Who can authoritatively confirm that every old instance and rollback window has ended? Our built-in Helm deployment currently uses `Recreate` and defaults to one backend replica, and the container runs migrations before starting the server, so the supported deployment path does not normally run old and new backends together. A custom multi-replica rolling deployment can, though, and upstream cannot observe those instances.

Splitting the contract step into a later release makes rollback from release B to release A safe because A no longer depends on the table. It does not automatically make a direct rolling upgrade from pre-A to B safe. There is also a smaller transition gap in release A: an old instance can still write a new route row after the one-time backfill, so a later drop needs either a final backfill or a compatibility mirror to avoid missing observations created during the mixed-version window.

So I would like your view on the intended guarantee:

- Are we targeting the official `Recreate`/single-instance deployment contract, with adjacent-release rollback compatibility?
- Or should this remain safe for arbitrary self-hosted rolling deployments from older versions?

If the latter is the requirement, my current inclination is to keep the deprecated table (and probably its indexes) indefinitely or until a clearly documented major/maintenance migration, rather than assume a follow-up release can know that all old processes are gone. If we agree on the narrower supported deployment contract, then your proposed expand/contract split makes sense: this PR can backfill, stop depending on the table, retain it, and explicitly document migration 381's irreversible session-continuity trade-off; a later contract change can run a final backfill and remove the table under that policy.

I am working through the other non-blocking suggestions locally now and will push those changes to this PR shortly.

This is why the first version deliberately made a clean cut: not because these compatibility questions were overlooked, but because after weighing them I wanted the new model to be explicit and to see how you and the other reviewers think we should define the upgrade contract.

yyclaw:
@Bohan-J I pushed the follow-up changes at `af6885ed8`. The two commits were rewritten in place so the routing removal and the replacement inventory model remain separately reviewable.

Migration blockers:

- Migration `382` now only removes stale cross-Agent DingTalk bindings and their outbound cards. Its irreversible rollback/session-continuity trade-off is now explicit in the PR description.
- Migrations `383`–`389` create and backfill `dingtalk_group_presence` and installation-owned `dingtalk_bot_identity`. The migration test now seeds legacy route rows and verifies that titles survive.
- The backfill also installs a compatibility trigger: if an old process writes `dingtalk_group_route` after the one-time backfill, that observation is mirrored into the new presence table.
- This PR no longer drops `dingtalk_group_route` or any of its indexes. The new runtime has no dependency on them, but they remain available to old processes and rollback builds.
- Installation replacement/reclaim, system-Agent teardown, and workspace deletion still explicitly clean retained route rows, so compatibility storage cannot outlive its owner.
- Numbering now starts at upstream's latest migration (`376`) + 6: `382`–`389`.

Inventory/read-path changes:

- `last_active_at` and `mention_count` are persisted when an accepted inbound message becomes durable; inventory reads no longer aggregate message history.
- Agent and installation filters are pushed into SQL instead of fetching the whole workspace and filtering in Go.
- Groups active within 90 days load by default. Long-inactive groups are collapsed and loaded 20 at a time, with deterministic pagination.
- Workspace owners/admins can forget a stale group. This deletes only the presence row, not conversations/messages or the installation-owned Bot identity. A later accepted @mention recreates the group.

OpenAPI pressure:

- `qyapi_chat_manage` denials are cached for 30 seconds at the AppKey + RobotCode identity scope.
- The same identity key is used for `singleflight`, so simultaneous first messages from different groups sharing one Bot collapse into one OpenAPI request.

I also updated the four MDX locales and the PR description to match the final code and migration behavior.

Local verification on the rewritten head:

- `make sqlc`
- `make migrate-up`
- `go test -p 1 ./internal/handler ./internal/integrations/dingtalk ./internal/migrations ./cmd/migrate ./internal/integrations/channel/engine -count=1`
- 55 targeted Vitest cases for DingTalk Settings and Agent Integrations
- `@multica/views` and `@multica/docs` typechecks

Bohan-J:
Reviewed `ffe3cdbc0`. Both blockers are resolved, and the rework goes further than what I asked for in a couple of places worth calling out.

The compatibility trigger is the right instinct, and it closes a gap I had not raised: I only asked for a one-time backfill, which would still have lost observations an old process wrote afterwards. I checked the `is_activity := NEW.revision = OLD.revision` heuristic against `main` — `DiscoverDingTalkGroupRoute`'s conflict update touches only `conversation_title` and `updated_at`, while `ReassignDingTalkGroupRoute` does `revision = r.revision + 1` — so "discovery counts as activity, reassignment only refreshes the title" is correct. Moving the inventory into dedicated tables with persisted counters, pushing the filters into SQL, and scoping the identity cache to AppKey + RobotCode instead of per-group all address the read-path concerns at the root rather than patching them.

## On the upgrade contract question

Your reading of the deployment setup is accurate, and it points at a narrower answer than the one you were bracing for.

`deploy/helm/multica/templates/backend.yaml` sets `strategy: Recreate`, `values.yaml` defaults the backend to one replica (with concurrent pods made safe by the advisory lock in `cmd/migrate`, #3647), and `docker/entrypoint.sh` runs `./migrate up` before the server starts. Because `Recreate` terminates every old pod before starting new ones, the supported chart does not produce a mixed-version window at any replica count. The real exposure is `helm rollback` (documented in `SELF_HOSTING.md`) — an old image running against a new schema.

So: **target the documented contract — `Recreate` plus adjacent-release rollback — not arbitrary rolling upgrades from any older version.** Concretely:

- Retaining the table in this PR and contracting later is the right shape. Keep it as is.
- Keep the mirror trigger, but the strongest argument for it is rollback rather than mixed versions: after a `helm rollback`, the old build writes route rows again, the trigger keeps mirroring them into the presence table, and rolling forward again preserves those observations. That is worth the small amount of machinery.
- Retaining the table **indefinitely is not required**, and I would rather not. There is precedent for destructive schema changes in this repo when they come with an operator-visible upgrade note and a pre-migration hook — migration `103` is the clearest example. A later release can contract normally; because the trigger keeps the new table continuously in sync, that release only needs to drop the trigger and the table together, with no final backfill step.

## One blocker: the retained table lost its cleanup paths

`TestWorkspaceDeletionManifestCoversPublicSchema` fails with `unclassified=[dingtalk_group_route]`, both locally and in CI (`backend-tests` in run 32386126065).

This is a consequence of the change in direction: the first version dropped the table, so every cleanup path for it was removed along with it; switching to expand/contract kept the table but did not bring those paths back. On `main` there are five:

- `server/pkg/db/queries/workspace_delete.sql:439` — workspace deletion
- `server/pkg/db/queries/workspace.sql:118` — workspace installation teardown
- `server/pkg/db/queries/channel.sql:212` — installation teardown (`dead`)
- `server/pkg/db/queries/channel.sql:267` — runtime teardown (`doomed`)
- `DeleteDingTalkInstallationForReplacement` in `server/pkg/db/queries/dingtalk.sql`

The manifest guard only catches the first, but the substantive problem is broader: **deleting a workspace now leaves its legacy route rows behind**, in a table that is still being written to during the compatibility window. Please restore all five plus the manifest entry for as long as the table is retained, and remove them together with the table in the contract release.

One likely reason this passed for you locally: the test enumerates `pg_tables` from the live database, so a development database that still carries the effect of the earlier drop-table version of this branch no longer has `dingtalk_group_route` and cannot see the drift. Re-running against a freshly migrated database reproduces the CI failure.

## Also worth fixing in the same pass

- **Backfilled groups all land in the collapsed bucket.** Legacy rows arrive with `last_active_at` NULL, so after an upgrade the default inventory is still empty until each group is mentioned again; the groups are reachable only by expanding a Bot's historical section. The data is genuinely retained now, and `inactive_group_counts` surfaces it, so this is a reasonable design — but since "does the upgrade keep existing groups" is what started this thread, please state the post-upgrade behavior explicitly in the PR description.
- **Paginated ordering does not match display ordering.** The query pages by `last_active_at DESC NULLS LAST` (`server/pkg/db/queries/dingtalk.sql:182`), then the handler re-sorts each page by title (`server/internal/handler/dingtalk.go:437`). "Load more" therefore appends a second alphabetical run instead of continuing one order. Sorting the merged list on the client, or paging by the same key you display, would fix it.
- **Singleflight can hand a group-specific failure to a different group.** The flight is now keyed on identity, but the lookup runs with the first caller's conversation ID. If that lookup fails for a group-specific reason (the Bot is absent from that group), a concurrent caller asking about another group receives that error. The result is cached under the correct group key so it self-corrects, and the impact is only a missing name on a best-effort path — but returning the shared result only when it is app-scoped would be more precise.
- **Forget is owner/admin-only while connect/disconnect allows the Agent owner.** If that asymmetry is deliberate it is fine; flagging it in case it was not.

## Verification

- `go run ./cmd/migrate up` against an empty database — the full `382`–`389` chain, including both backfills, applies cleanly
- `go test ./cmd/server ./internal/integrations/channel/engine ./internal/integrations/dingtalk ./internal/migrations -count=1` — pass
- `go test ./internal/handler -run TestDeleteWorkspace -count=1` — pass
- `go test ./internal/handler -run TestWorkspaceDeletionManifestCoversPublicSchema -count=1` — **fail**, matching CI
- CI: `backend` and `backend-tests` red, the other 14 checks green including `sqlc-check`

## Verdict

Requesting changes, for the cleanup paths and the resulting red CI. Everything else above is smaller. This round resolved both blockers at the root and volunteered a fix for a gap I had missed; once the legacy cleanup is restored and the smaller items are addressed, I expect this to be ready to merge.

yyclaw:
@Bohan-J Follow-up is now at `b5a5980ec`, and all applicable CI checks are green.

I rechecked each point against the current head rather than the earlier reviewed snapshot:

- The retained-table blocker was already fixed in `af6885ed8` and remains covered here: all five installation/system-Agent/workspace teardown paths delete `dingtalk_group_route`, and the workspace deletion manifest includes it.
- The PR and issue now state the exact upgrade result for legacy groups: titles are retained, but legacy routes have no trustworthy activity timestamp or count, so backfilled rows start with `last_active_at = NULL` / `mention_count = 0` under the collapsed **Long inactive** section until a later accepted @mention records real activity.
- The pagination path already flattened every loaded inactive page before applying the display sort on the client. I added a regression test with two deliberately non-alphabetical pages to prove that loading more produces one global title/conversation-ID order rather than a second alphabetical run.
- Bot-identity singleflight now shares only results whose scope is proven to be app-wide: a successful identity or `qyapi_chat_manage` denial. If the flight leader gets a group-specific failure, callers for another group retry with their own conversation ID. The new deterministic concurrency test also passes under `-race`.
- The Forget asymmetry is deliberate. Connect/disconnect remains available to the Agent owner or a workspace owner/admin, while deleting workspace-maintained historical group inventory in Settings remains workspace owner/admin-only. The server authorization, Settings affordance, tests, PR description, and issue now state the same rule.
- The rollout text now matches the deployment contract we agreed on: documented Helm `Recreate` plus adjacent-release rollback. This release retains the legacy table, indexes, and mirror trigger; a later contract release can drop the trigger and table together after this release has removed every runtime dependency.

Local verification used a freshly migrated isolated database and covered the handler, DingTalk adapter, migrations, migration CLI, and shared channel engine. The final GitHub run passed backend, frontend, both views shards, sqlc-check, mobile, installers, Windows execenv, and the vulnerability scan.

multica-eve (changes requested):
Reviewed at `b5a5980ec`. The earlier migration, cleanup, pagination, and singleflight issues are resolved, but one authorization gap still blocks merge.

`ListDingTalkGroups` accepts an inactive-page `installation_id`, queries that installation, and computes `next_offset` before applying the caller's Agent visibility filter (`server/internal/handler/dingtalk.go:294`, `:328`, `:343`, with filtering only at `:397`). An ordinary workspace member can therefore request a private Agent's installation with `limit=1&offset=N`; even though `groups` is filtered to empty, `next_offset` reveals whether that inaccessible Bot has more than N inactive groups.

The identifier does not need to be guessed:

- DingTalk binding redemption checks workspace membership, not Agent visibility (`server/internal/integrations/dingtalk/binding.go:122-139`), and returns the installation ID (`server/internal/handler/dingtalk.go:726-729`).
- The resulting `dingtalk_installation:binding_updated` event also carries that ID (`server/internal/handler/dingtalk.go:723-725`) and is forwarded as a workspace-wide event (`server/cmd/server/listeners.go:206-243`).

The disclosed data is narrow — a group-count threshold, not group names or message content — but it violates this PR's stated invariant that members who cannot view an Agent cannot infer its DingTalk group activity.

Please authorize the requested installation before running the inactive query (or push the authorized Agent predicate into SQL), fail closed with a response indistinguishable from a nonexistent installation, and add a negative test for an ordinary member paging an inaccessible private Agent's installation, including the absence of `next_offset`.

Once this is fixed, I expect the PR to be ready for another review.

yyclaw:
@multica-eve The authorization leak is fixed at 2182c6b, and all applicable CI checks are green.

I rechecked the current request path before changing it. The inactive-page handler now resolves the requested DingTalk installation within the workspace and dingtalk channel scope, verifies that it is active, and authorizes the associated Agent before querying group presence or deriving next_offset. The Agent Detail surface additionally requires the installation to belong to that exact Agent.

Missing, cross-workspace, revoked, inaccessible private-Agent, and cross-Agent installation IDs now produce the same empty response, without groups, counts, Bot identity, or next_offset. Database lookup failures remain explicit 500 responses.

The regression coverage includes exact-response equality with a nonexistent installation, the inaccessible-private-Agent cursor case you reported, revoked and cross-Agent cases, and authorized two-page pagination for a visible Agent. Could you take another look when convenient?

Bohan-J (approved):
Verified at `2182c6b`. Everything raised across this review is resolved, and CI is green.

Thank you for this one — the way you handled it was genuinely better than a normal review cycle. A few things worth naming:

- When the backfill gap was confirmed, you did not just add the migration I sketched. You noticed that a one-time backfill still loses observations an old process writes afterwards, and added the mirror trigger to close it. That gap was mine to catch and I had missed it.
- You pushed back on the expand/contract advice with a specific, well-posed question about which deployment contract we actually support, instead of either quietly complying or quietly ignoring it. That question was worth answering properly, and the answer is now written down for the next person who hits it.
- The read-path feedback came back as a redesign rather than a patch: dedicated tables, persisted counters, filters pushed into SQL, an explicit active/inactive split. That is why the inventory will hold up as message volume grows.
- The authorization side-channel on the inactive cursor was a subtle one, and the fix went past the reported case to make missing, cross-workspace, revoked, inaccessible-private-Agent, and cross-Agent installations all produce a byte-identical empty response.

Local verification on my side, against a freshly migrated database: the full `382`–`389` chain applies cleanly, and `internal/handler`, `internal/integrations/dingtalk`, `internal/migrations`, `internal/integrations/channel/engine`, `cmd/server`, and `cmd/migrate` all pass.

Merging now. The follow-up contract release can drop the mirror trigger and `dingtalk_group_route` together once this release has been out long enough to close the rollback window.

PR #7243, opened 2026-08-19
feat(mcp): daemon-side probe and tool listing for workspace MCP servers

### LINKED ISSUE #7166: feat(mcp): daemon-side probe and tool listing for workspace MCP servers
Hi Multica team,

This is a follow-up to #7043. The workspace MCP catalog and explicit per-agent assignment are already in place via #6961 and #6976 — thank you for shipping that. What I would like to ask about next is the health / introspection piece that #7043 left out on purpose.

Today `multica workspace mcp list` (and the Settings UI) show the library entries: name, transport, and assignment. That is enough to configure a server, but not enough to know whether it will actually come up on a runtime, or which tools the agent will see. The stored entry is write-only, which is the right secret-handling model, so the only way to learn those things today is to start a task and watch the transcript.

## Use case

After adding a workspace MCP server (stdio or remote), I want to click **Test connection** — or run a CLI command — and get a real answer from the machine that will run the agent:

- Can this daemon start / reach the server?
- If yes, which tools did `tools/list` return?
- If no, what failed, in a form that does not leak secrets? (command not found, timeout, auth, TLS, unsupported transport, …)

A possible user flow:

1. Owner/admin opens Workspace Settings → MCP Servers (or `multica workspace mcp …`).
2. They pick a server and run **Test connection**.
3. The request is authorized on the backend, then executed on a connected daemon — not on the Multica server process.
4. The daemon does a short MCP handshake (`initialize`, then `tools/list`), tears the process / session down, and reports back.
5. Settings / CLI show last probe status, time, a short error if it failed, and the tool names if it succeeded.
6. On the agent MCP tab, the same last-probe info can be shown for assigned servers, as a warning rather than a grant of extra access.

I think the daemon is the right place to run this. That is already where resolved `mcp_config` is applied (Claude `--mcp-config`, Qwen `--mcp-config`, ACP `session/new` / resume). Stdio servers depend on the daemon's PATH, local binaries, `npx` / `docker`, and env. A probe from the backend would succeed or fail in a different environment than the one that actually runs the task. Remote HTTP/SSE could be reached from the server, but I would still prefer one path so the result matches what the runtime will see.

The result is also machine-scoped. If several daemons are online, the UI/CLI should say which one was used (or let the user pick). A green probe on laptop A should not be presented as “this workspace MCP is healthy everywhere.”

## Suggested model

- **Backend:** authorize the caller, load the write-only definition, pick / address a connected daemon, persist `last_probe` (status, time, daemon/runtime id, error code/message, tool names). Do not return secrets to the client.
- **Daemon:** spawn or connect using the same config shape it already uses at task start, handshake with a short timeout, run `tools/list`, shut down, reply.
- **UI / CLI:** on-demand **Test connection**; optionally `multica workspace mcp probe <id>` and show tools on `workspace mcp list` / `agent mcp list` from the last successful probe.

Something like:

```text
github    stdio    last probe: ok on "home-mbp" 12m ago    8 tools
notion    http     last probe: fail on "home-mbp" 3m ago   unauthorized
internal  stdio    last probe: never
```

This is intentionally an on-demand check, not a background health loop.

## Security / product constraints

- Same permission bar as editing the catalog: owner/admin to probe a workspace server. Members should not be able to trigger a spawn just because the entry exists.
- Probe must not change assignment. Seeing tools does not grant an agent that server.
- Response must stay redacted: no env, headers, tokens, or raw command lines that include secrets.
- Hard timeout; do not leave a stdio process running after the probe.
- Do not treat “in the catalog” as “healthy for every agent / every daemon.”

## Possible MVP

A small first version could include:

- Daemon-side probe for one workspace MCP server (stdio + remote, using the existing config envelope).
- Return `ok` / `fail`, a safe error, elapsed time, daemon identity, and tool names (no full JSON Schema required).
- Persist last probe so Settings and `workspace mcp list` can show it.
- CLI: `multica workspace mcp probe <server-id>` (and maybe print tools).
- Settings: a Test connection button on the workspace MCP row.
- Tests for: no daemon online, timeout, command not found, permission denied, disabled server, secret redaction, empty `tools/list`.

I would keep these out of the first cut: periodic background health, OAuth, marketplace / discovery, probing from the backend, and caching full tool schemas.

I could not find an open issue specifically for daemon-side MCP probe + tool listing. If this direction fits, I would like to contribute it. Happy to adjust the API / WS shape to whatever you already use for daemon RPCs before opening a PR.

Thank you for your time, and for the workspace MCP work that already landed.

thuanlm215:
Closes #7166.

## Why
The workspace MCP catalog is write-only. Settings and `workspace mcp list` show name/transport/assignment, but not whether a daemon can start the server or which tools `tools/list` would return. The only way to learn that today is to start a task.

## What
On-demand probe, executed on a **connected daemon** (not the API process), using the existing pending-work + heartbeat pattern (same NAT-safe shape as model list):

1. Owner/admin `POST /api/workspaces/{id}/mcp-servers/{serverId}/probe` (`runtime_id` required when several runtimes are online).
2. Server wakes the daemon; daemon fetches the write-only config over a daemon-auth GET, handshakes (`initialize` + `tools/list`), tears the session/process down.
3. Redacted result is reported back and persisted as `last_probe` (status, runtime identity, elapsed, error code/message, tool names).
4. CLI: `multica workspace mcp probe <id>` and last-probe columns on `workspace mcp list` / `agent mcp list`.
5. Settings: **Test connection** on the library row. Agent MCP tab shows a warning when last probe failed.

New daemon capability: `mcp-probe-v1`. Old daemons fail immediately instead of hanging.

## Security
- Same write gate as the catalog (owner/admin; agent actors denied).
- Probe does not change assignment.
- User-facing responses never include url/command/args/headers/env.
- Hard timeout; stdio child is killed.

## Out of scope
Periodic health, OAuth, marketplace, probing from the API process, caching full tool schemas, mobile UI.

## Test plan
- [x] Daemon handshake tests (stdio tools, empty tools, command not found, timeout, unsupported transport, HTTP 401, HTTP tools)
- [x] In-memory store timeout + error sanitization
- [x] Frontend schema strips secrets on `last_probe`; Settings Test connection (owner-only)
- [ ] Handler DB tests (need local Postgres / `make test` on a machine with the suite DB)
- [ ] Manual: Settings → MCP → Test connection on a live daemon; CLI probe; agent tab fail warning


thuanlm215:
Hi @multica-eve, I have synced this branch with the current main branch and resolved the protocol conflict while preserving both the new upstream local-skills work kinds and the MCP probe work kind. Focused verification passes: `go test ./internal/daemon ./pkg/protocol` and a compile-only handler check with `go test ./internal/handler -run "^$"`. The push has triggered the full CI suite. When convenient, could you review the implementation direction and scope? In particular, I can split the persisted `last_probe` state into a follow-up and keep this PR focused on the immediate on-demand result, as suggested in #7166.

PR #7249, opened 2026-08-19
MUL-6166: docs: publish the community-maintained areas and their maintainers

Bohan-J:
## What this does

`community-maintained` has been in our docs since DingTalk landed, but it never said *who* maintains what. The names existed only in Go package comments and PR threads — exactly where a user filing an issue will never look.

This adds a **Community-maintained areas** page (`/community-maintained`, four locales) listing each area, its code path, its maintainers, when it started, and the retirement rule, and makes that page the single source of truth. `channels`, the DingTalk and Telegram guides, the README, and the three package doc comments now point at it instead of restating the policy — those comments no longer carry maintainer handles or a second copy of the policy, so an ownership change is recorded in exactly one place.

| Area | Maintainers |
| --- | --- |
| DingTalk chat integration | @yyclaw |
| WeCom chat integration | @leroy-chen, @seacen |
| Telegram chat integration | @leonzone |

Two deliberate choices:

- **It is a credit-and-expectations page, not a support directory.** It asks readers to file a GitHub issue and explicitly asks them *not* to DM or @-mention a maintainer. What we promised each of you is that we would route work to you — not that you are on the hook for it, and not that anyone can page you directly.
- **The retirement rule is public.** "If an area breaks in a way we cannot fix without real access and no fix appears for a few releases, we may deprecate it" reads like a warning to users, but it is mostly a protection for maintainers: a volunteer who gets busy or moves on owes nobody a fix.

Both issue forms also gain a required **Area** field, so a report can be routed without the reporter having to read source.

## @yyclaw, @leroy-chen, @seacen, @leonzone — heads-up, and a chance to object

Each of you agreed to be the maintainer of your area in your own PR. Publishing your handle on the docs site is a step further than a code comment, so: **if you would rather not be listed by name, say so here and I will drop your row to an unnamed "community-maintained" entry.** No explanation needed and nothing else changes. Same if you want the wording about your area adjusted — you know these platforms better than we do.

## CODEOWNERS: intentionally not here

Worth stating plainly, because it is easy to assume otherwise: `.github/CODEOWNERS` is not a reference list for developers. GitHub automatically requests a review from a matching owner on every PR that touches their paths, and where branch protection requires code-owner approval it gates the merge as well. It also only works for accounts with **write** access — an owner without it is silently skipped, no warning on the PR.

We are not granting write access for this, and that is the right call: being a code owner here is about routing questions to the person who knows the platform, not about merge or push authority over the repository. The four of you stay on **Triage**, which is what lets us assign issues and PRs to you and request your review.

So routing stays deliberate on our side — this docs page, the new Area field, and a human deciding when to pull you in — rather than automatic.

One question while you are here: **would you rather see every PR that touches your area automatically?** If so, say the word and we will wire that up as an opt-in, per person. It is a small workflow file, not CODEOWNERS, and it stays off by default because "we route work to you" is what we promised, not "you are subscribed to everything."

### On the word "code owner"

The page and the issue forms deliberately say **maintainer**, not "code owner". On GitHub the latter names a specific automated mechanism — a `CODEOWNERS` entry auto-requests review and can gate merges — and we are not using it, so borrowing the term in user-facing copy would promise behavior that does not exist.

## Verification

- `pnpm --filter @multica/docs build` passes and all four locales of the new page prerender (`.next/server/app/{en,zh,ja,ko}/community-maintained.html`), with the maintainer links and the new sidebar group present. Note: the build only completes with a standard `NODE_ENV`; an ambient `NODE_ENV=development` breaks the `/404` prerender on `main` too, independently of this change.
- `go build ./...` and `go vet` on the three touched packages pass, `gofmt` clean — those changes are comments only.
- Both issue-form YAMLs parse, and the nav JSON parses in all four locales.


Bohan-J:
This is merged and live: **https://multica.ai/docs/community-maintained**

@yyclaw, @leroy-chen, @seacen, @leonzone — your names are on it, and that is the point of the page. Thank you.

- **@yyclaw** — DingTalk (#4829), the first community channel we shipped. You carried a fork rebase for six weeks before it landed, and the identity fix you added for a robot swap is still the behavior we rely on.
- **@leroy-chen** — WeCom (#5833). We declined #4128 and asked for a rewrite onto the shared Channel engine; you did the whole thing again rather than argue, and the result plugs into `channel.Channel` exactly like Slack and Lark do.
- **@seacen** — you co-own WeCom while running it against a real tenant daily, which is the one thing we cannot do. The `Connect` deadlock repro and catching that a rollback would have silently revoked `dingtalk_chat` are both things no amount of review on our side would have found.
- **@leonzone** — Telegram (#6944), five review rounds in five days, every one of them coming back with the actual fix rather than a way around it.

A note on what the page says, since it matters more than the credit: these areas are yours in the sense that questions about them come to you first, on a best-effort volunteer basis — not in the sense that you owe anyone a fix. If you get busy or move on, the worst outcome is that we retire an integration. Nobody comes chasing you. The page states that publicly and asks users to file a GitHub issue rather than @-mention you directly, so that routing stays our job.

We deliberately say *maintainer* rather than *code owner* there. GitHub's `CODEOWNERS` is an automated mechanism that needs write access to the repository, and we are not going to hand out write access to make a label accurate — so the word would have promised something that does not happen.

And the offer from the PR description stands now that it is live: if any of you would rather not be listed by name, say so here and we will drop your row to an unnamed entry. No explanation needed.

Looking forward to the next one. If you are considering another area — or want to bring someone else in alongside you the way WeCom is co-maintained — open an issue and we will talk it through before you write any code.

PR #7261, opened 2026-08-19
MUL-6629: feat(agents): add per-agent starter prompts for chat

### LINKED ISSUE #5537: Per-agent starter prompts for the chat empty state
## Problem

The chat empty state shows three starter-prompt chips (`packages/views/chat/components/chat-window.tsx`, `STARTER_KEYS: list_open / summarize_today / plan_next`, resolved from the i18n `chat` namespace). The set is global: every agent in every workspace gets the same three developer-centric prompts.

In multi-agent workspaces where agents have sharply different roles (a routing front door, a domain conductor, a reviewer), the generic set misleads users about what the selected agent is for - e.g. "List my open tasks by priority" presumes the human executes tasks, which inverts the model in workspaces where agents execute and humans review.

## Proposal

Make starter prompts configurable per agent, falling back to the current i18n set when unset:

- an `agent.starter_prompts` field (array of up to 3 strings, optionally with an icon), editable in the agent settings UI,
- chips render the agent's set when present; tapping still prefills the composer as today.

Per-agent prompts are user-authored strings persisted to the DB, so they need no i18n plumbing - the same reasoning as `helper-starter-prompts.ts`, which already persists prompt strings at write time instead of reading the i18n bundle.

## Why per agent, not per workspace

The prompt that makes sense depends on the selected agent's role, and the empty state already re-renders per agent ("Hi, I'm {agentName}"). A workspace-level set would inherit the current problem one level down.

yugui923:
## What does this PR do?

Closes the blank-composer gap described in #5537 by letting an agent define up to three suggested first actions. Each suggestion has a short label for the empty state and a complete prompt for the composer.

The suggestions appear before the first message on Web, Desktop, and Mobile. Selecting one only prefills the composer and focuses it where supported; it never sends the prompt or starts an agent run automatically. Agents without configured suggestions receive localized general-purpose fallbacks, so existing agents also get a useful first-turn experience.

The configuration is carried through manual creation, Agent Builder drafts, saved drafts, agent settings, Web duplication, and `multica agent copy`.

### Thinking path

- #5537 remains open and unassigned, with no linked implementation, matching pull request, or active implementation branch as of 2026-08-19.
- A bounded, persisted agent setting makes the examples role-specific and deterministic without adding an LLM call when Chat opens.
- A `{ label, prompt }` pair keeps the empty-state copy concise while allowing the composer to receive a useful, editable instruction.
- Prefill-only behavior preserves user control and avoids creating a session or run from an accidental click.
- The additive JSONB column defaults to `[]`; clients fall back locally, so existing rows and older payloads remain usable.
- Icons, automatic sending, analytics, and dynamic prompt generation are intentionally outside this first version.

## Related Issue

Closes #5537

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added migration 377 and sqlc support for an `agent.starter_prompts` JSONB array, including create/update/list response handling and server-side validation for three complete label/prompt pairs.
- Extended shared Agent types, API requests, stored drafts, duplication, Agent Builder protocol/instructions, and CLI copy behavior.
- Added a reusable authoring editor to manual creation and the agent Instructions settings tab, with localized validation in English, Simplified Chinese, Japanese, and Korean.
- Reworked the shared Web/Desktop and Mobile chat empty states to prefer agent-specific suggestions and otherwise show localized/general fallbacks.
- Made suggestion selection prefill the correct new-session or existing-session draft without sending.
- Updated user docs and the built-in creating-agents skill/source map.
- Added validation, draft/duplication, builder protocol, empty-state interaction, mobile schema, CLI copy, migration, and API round-trip coverage.
- Preserved compatibility with legacy agent-creation call paths by normalizing an omitted `starter_prompts` value to an empty JSON array in the insert query.

## How to Test

1. Create an agent with one to three suggested first actions; verify incomplete pairs cannot be saved.
2. Open a new Chat with that agent and verify its configured labels appear before any message is sent.
3. Select a suggestion and verify the full prompt is placed in the composer, remains editable, and no session/run starts until Send is used.
4. Remove all configured suggestions and verify localized fallback actions appear.
5. Edit suggestions from the agent Instructions tab, then duplicate/copy the agent and verify the configuration is preserved.
6. Verify Web/Desktop and Mobile show the same configured prompts and degrade safely when the field is missing or malformed.

## Validation

- `bash scripts/test-go.sh --race` against isolated PostgreSQL and Redis
- `go vet ./...`
- `make sqlc` (clean regeneration)
- Core: 132 files / 1,558 tests passed
- Views: 389 files / 4,514 tests passed
- Focused starter-prompt tests: Core 32, Views 24, Mobile 5, plus Go handler/CLI/service/migration tests
- TypeScript typechecks: Core, Views, Web, Desktop, Mobile
- ESLint: Core, Views, Mobile (only existing unrelated warnings)
- Prettier check for new files and changed locale JSON
- `git diff --check`
- Gitleaks scan of every changed and untracked feature file

The database-backed handler round-trip and onboarding regression tests were rerun against isolated PostgreSQL after CI exposed a legacy-call-path regression. The full race-enabled backend suite now passes locally.

## Risks and compatibility

- Migration 377 is additive and reversible. It constrains the JSON value to an array with at most three items; the handler owns item shape and length validation.
- Older agents receive `[]` and use client-side fallbacks. Mobile treats malformed additive data as an empty list instead of rejecting the agent payload.
- Labels are rendered as escaped UI text, and prompt content is not executed until the user explicitly sends it.
- No new dependency or runtime execution path is introduced.

## Checklist

## AI Disclosure

**Prompt / approach:** Repository and GitHub issue/PR analysis were used to confirm #5537 had no existing implementation. Codex then implemented and tested the additive persistence contract, shared authoring flow, Agent Builder/draft/duplication support, and Web/Desktop/Mobile empty-state behavior. The design deliberately uses bounded `{ label, prompt }` pairs and prefills without auto-sending.

## Screenshots

The red callouts highlight the affected surface: dashed outlines show the previously missing UI, and solid outlines show the new controls.

### New Chat empty state

Before and after comparison of the new Chat empty state

### Agent creation

Before and after comparison of the agent creation form


yugui923:
@Bohan-J, would you please review this implementation for #5537 when you have a chance? The review-and-fix pass is complete, all CI checks are green, and the before/after UI screenshots are included in the PR description.

Bohan-J:
@yugui923 Thx for noticing. Gonna review it in next few days

multica-eve (changes requested):
This change lets each agent offer up to three role-specific starting actions before a conversation begins. Selecting one keeps the user in control: it fills the composer for review and editing, but does not send anything or start a run. Previously, every agent exposed the same generic guidance, which could misrepresent agents with different roles.

**Product verdict: ✅ This is a useful feature and should be accepted.** The user journey and scope are appropriate: suggestions are bounded, agent-specific, preserved by duplication, and existing agents retain useful fallbacks.

## Must-fix issues

### 1. Preserve unsaved starter-prompt edits across agent-list refreshes

In `packages/views/agents/components/tabs/instructions-tab.tsx:46`, the synchronization effect depends directly on the `agent.starter_prompts` array reference and unconditionally replaces the local editor state whenever that dependency changes.

The agent list is invalidated for any `agent:*` event in `packages/core/realtime/use-realtime-sync.ts:763`, including routine status changes. A refetch produces a new array reference even when the persisted prompt contents are identical, so an unrelated agent status event or reconnect can silently erase a user's unsaved prompt edits while this form is open.

Please synchronize from a semantically stable persisted value and protect dirty local state. Add a regression test that edits a prompt, rerenders with an equivalent `agent` whose `starter_prompts` array has a new identity, and verifies that the local input is preserved.

### 2. Do not silently lose writes when the new frontend talks to an older backend

`packages/core/agents/draft.ts:223` and `packages/views/agents/components/tabs/instructions-tab.tsx:60` send `starter_prompts` through the existing create/update endpoints. Web/Desktop staging is deployed independently from the manually deployed backend; an older Go handler ignores this unknown JSON field and still returns 200/201.

During that mixed-version window, creation or settings therefore reports success while dropping the configured suggestions. The prompts disappear after refresh. Please add a backend-capability gate or a versioned contract that an older backend rejects explicitly, so unsupported writes fail visibly instead of succeeding with data loss. This needs mixed-version coverage for both create and update.

## Nit

`server/cmd/multica/cmd_agent_copy.go:31` and the built-in creating-agents skill say that `starter_prompts` can be overridden with a matching flag, but `registerAgentCopyFlags` has no such flag. Please either add the flag or describe this field as copied by default without claiming an override.

**Recommendation: Request changes.** The product direction is sound and the main persistence/UI paths are well covered, but the two issues above can cause user-authored configuration to be lost.

yugui923:
Thanks for the thorough review — all three points are addressed in `b3f040461`, with the current-main migration follow-up in `ece8a7c26`:

1. **Unsaved edits are preserved.** The Instructions editor now compares semantic persisted snapshots instead of array identity and keeps dirty starter-prompt edits across equivalent refetches. Regression coverage includes a new array identity and a real persisted-content change while local state is dirty.
2. **Mixed-version writes fail closed.** `/api/config` now advertises `agent_starter_prompts_supported`. When an older server omits it, Web/Desktop hide the controls, empty creates omit the field, and explicit create/update writes are rejected before a request can silently lose data. Tests cover supported and unsupported create/update paths.
3. **Copy behavior is documented accurately.** CLI help and the built-in creating-agents skill now say starter prompts are copied by default without a dedicated override flag.

A final attention pass also caught that current `main` had advanced through migration 403, so the starter-prompts migration is now 404 to avoid the merge-time prefix collision.

All GitHub CI checks are green. Ready for another review.

multica-eve (changes requested):
Thanks for addressing the previous review. The semantic refetch guard, the fail-closed Web/Desktop capability gate, the copy documentation, and the migration renumbering all look correct. The new targeted tests pass locally.

One save-failure path still loses the user's starter-prompt edits:

### Must-fix: preserve local edits when an optimistic update rolls back

`packages/views/agents/components/agent-detail-page.tsx:168` publishes the edited fields into both agent caches before `api.updateAgent` resolves. `InstructionsTab` then receives those optimistic values as new props. In `packages/views/agents/components/tabs/instructions-tab.tsx:83`, the effect records them in `persistedRef` even though they have not been persisted yet.

If the request later fails, `agent-detail-page.tsx:185` restores the old cached values. On that rerender, `localRef` equals the previous optimistic snapshot, so `wasLocallyDirty` is false and `instructions-tab.tsx:88` replaces the editor state with the old server values. The user sees the save error and their just-entered starter-prompt text disappears.

Please distinguish confirmed server state from optimistic props, or otherwise retain/restore the form values when `onSave` rejects. Add a regression covering a delayed rejected update: edit a starter prompt, publish the optimistic agent props, roll them back, and verify the label and prompt remain in the editor for retry.

**Recommendation: Request changes.** The previous blockers are resolved, but a failed save can still erase the user-authored configuration this feature is meant to preserve.

yugui923:
@Bohan-J, addressed the optimistic-rollback blocker in `0f4172f82`.

- `InstructionsTab` now ignores same-agent persisted-state synchronization while a save is in flight, so neither the optimistic cache snapshot nor its rollback can overwrite the submitted local values. Agent switching still loads the newly selected agent immediately.
- Added the requested delayed-rejection regression: edit both starter-prompt fields, publish optimistic props, roll back to the old server props, reject the save, and verify both the label and prompt remain available for retry.

Validation completed under Node 22 / UTC: repository-wide lint, repository-wide TypeScript typecheck, and the full frontend test matrix (5 packages; shared views: 394 files / 4,594 tests). The staged patch also passes gitleaks.

Ready for another review.

multica-eve (approved):
Re-reviewed commit `0f4172f82`. The pending-save guard now prevents both the optimistic cache snapshot and its rollback from being mistaken for confirmed persisted state, while agent switches still load immediately. The delayed-rejection regression covers both starter-prompt fields and keeps them available for retry after the cache rolls back.

All previously requested changes are now resolved: semantic refetch protection, the fail-closed mixed-version capability gate, copy documentation, and the migration prefix update.

Local verification passed: the targeted Views test file (4 tests), Views typecheck, Views lint with 0 errors, migration lint, and `git diff --check`. The current head's GitHub checks are also green.

**Recommendation: Approve.** I found no remaining merge blockers.

Bohan-J:
Final review done — merging. Thanks for the careful iteration on this one, @yugui923.

Everything raised in the earlier rounds is closed:

- **Refetch clobbering unsaved edits** — the tab now syncs against a semantic snapshot and only adopts server state when the form is clean.
- **Silent write loss on a mixed-version deploy** — `agent_starter_prompts_supported` on `/api/config` fails closed, so an older backend surfaces a real error instead of a false success. Nice touch keeping the create path from sending the field at all when the list is empty, so agent creation still works unchanged against an old server.
- **Optimistic-cache value recorded as persisted baseline** — resolved by ignoring prop changes while a save is in flight, so a failed save leaves the submitted label and prompt in place for retry.
- **CLI/skill copy wording** — now matches `registerAgentCopyFlags`.

What I verified locally on `0f4172f82`:

- Applied migration `404` to a scratch database and ran the full `internal/handler` suite green against it, which covers the create/update/response round-trip through the new JSONB column. The `down` migration reverses cleanly too. (This closes the one gap flagged in the previous round, where the round-trip had only been covered remotely.)
- `go vet` on `internal/handler` and `cmd/multica`; `cmd/migrate` and `cmd/multica` agent-copy tests pass.
- Generated sqlc params line up with the hand-written query, and `404` is a unique migration prefix.
- Locale key parity across `en` / `ja` / `ko` / `zh-Hans` for both new namespaces, and docs updated in all four languages.
- All 16 CI checks green.

Two non-blocking follow-ups, whenever convenient:

1. Removing the old `chat-window.tsx` empty state orphaned six `chat.json` keys in each of the four locales: `empty_state.returning_subtitle`, `returning_title_default`, `returning_title_named`, `first_time_intro`, `first_time_pillars`, `first_time_pillars_suffix`. They're now unreferenced.
2. The `starterPromptRequest` prefill state is duplicated verbatim between `chat-window.tsx` and `use-chat-controller.ts` — a small shared hook would keep the two from drifting.

PR #7271, opened 2026-08-20
fix(build): add Windows executable suffix

### LINKED ISSUE #7255: [Bug]: `make build` omits the `.exe` suffix on Windows, so `multica daemon start` cannot re-exec itself
### Deployment type

self-host

### Area

Web app

### What happened

On Windows, a binary produced by `make build` cannot start the daemon in background mode:

```
start daemon: exec: "C:\Users\<user>\...\multica": executable file not found in %PATH%
```

`multica daemon start --foreground` works fine. Only the default (background) mode fails.

### Why

`Makefile` builds the CLI with a hardcoded output name that has no extension:

```
cd server && go build -ldflags "..." -o bin/multica ./cmd/multica
```

Background mode re-executes the same binary as a child: `runDaemonStart` resolves the
current image via `selfexec.Resolve()` (`os.Executable()`) and then runs
`exec.Command(exePath, args...)` (`server/cmd/multica/cmd_daemon.go:582`, `:621`).

On Windows, `os/exec` will not run a file whose extension is not in `%PATHEXT%` — **even
when given a full absolute path**. The extensionless `multica` is therefore rejected, and
the error text mentions `%PATH%` even though the path was fully qualified, which makes it
read like a PATH problem when it is not.

`--foreground` is unaffected because it never re-executes itself.

Release builds are fine: `.goreleaser.yml` declares `binary: multica` with `goos` including
`windows`, and GoReleaser appends `.exe` automatically. So this affects only people building
from source with `make build`.

### Environment

| Item | Value |
|---|---|
| OS | Windows 10 Enterprise N 22H2 (19045) |
| Go | 1.26.6 |
| Commit | `659bb41ab` (v0.4.30-7-g659bb41ab) |
| Built with | `make build` |

### Suggested fix

Append `.exe` to the CLI output when building for Windows, e.g. in `Makefile`:

```makefile
EXE := $(if $(filter windows,$(shell go env GOOS)),.exe,)

build:
	cd server && go build -ldflags "..." -o bin/server$(EXE) ./cmd/server
	cd server && go build -ldflags "..." -o bin/multica$(EXE) ./cmd/multica
	cd server && go build -o bin/migrate$(EXE) ./cmd/migrate
```

Only `bin/multica` strictly needs it for this bug, but applying it to all three keeps the
output consistent with what GoReleaser produces.

An alternative (or additional) improvement would be for `selfexec.Resolve()` to append
`.exe` on Windows when the resolved path has no extension, so a binary that was renamed or
built by other means still works.

### Steps to reproduce

On Windows:

```
make build
```

Then run the produced binary from a directory where **no** `multica.exe` exists:

```
./server/bin/multica --profile <a-logged-in-profile> daemon start
```

You get the error above. The same binary copied to `multica.exe` starts normally.

> Note: if `multica.exe` happens to sit next to it, Windows/MSYS may resolve to that file
> instead and the bug will not appear — the extensionless file has to be alone to see it.
> That is what made this confusing to track down.

### Screenshots (optional)

_No response_

```shell

```

Jamison929611:
## What does this PR do?

Fixes source builds on Windows producing extensionless executables that Windows cannot re-execute.

The failure comes from the `build` target's hardcoded extensionless output paths. The Makefile now derives a platform suffix from `go env GOOS` and applies it consistently to the server, CLI, and migration binaries. On non-Windows targets, the suffix remains empty, so the existing output paths are unchanged.

## Related Issue

Closes #7255

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Updated `Makefile` to append `.exe` when `go env GOOS` is `windows`.
- Applied the suffix consistently to `server`, `multica`, and `migrate` build outputs.
- Left non-Windows output names unchanged.

## How to Test

1. On Windows, run `make -n build` and confirm the output paths are `bin/server.exe`, `bin/multica.exe`, and `bin/migrate.exe`.
2. Set `GOOS=linux`, run `make -n build`, and confirm the output paths remain `bin/server`, `bin/multica`, and `bin/migrate`.
3. On Windows, run `make build` and confirm `server/bin/server.exe`, `server/bin/multica.exe`, and `server/bin/migrate.exe` are generated.

Locally verified with GNU Make 4.4.1 on Windows. The exact Windows build completed successfully and produced all three non-empty `.exe` files.

## Checklist

## AI Disclosure

**Prompt / approach:** Used to inspect the issue and repository state, make the narrowly scoped Makefile change, and run Windows and non-Windows verification.

## Screenshots (optional)

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,11 +278,12 @@ multica: ## Run the multica CLI entrypoint directly from the Go source tree
 VERSION ?= $(shell git describe --tags --match 'v[0-9]*' --always --dirty 2>/dev/null || echo dev)
 COMMIT  ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo unknown)
 DATE    ?= $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+EXE     := $(if $(filter windows,$(shell go env GOOS)),.exe,)
 
 build: ## Build the server, CLI, and migrate binaries into server/bin
-	cd server && go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT)" -o bin/server ./cmd/server
-	cd server && go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT) -X main.date=$(DATE)" -o bin/multica ./cmd/multica
-	cd server && go build -o bin/migrate ./cmd/migrate
+	cd server && go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT)" -o bin/server$(EXE) ./cmd/server
+	cd server && go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT) -X main.date=$(DATE)" -o bin/multica$(EXE) ./cmd/multica
+	cd server && go build -o bin/migrate$(EXE) ./cmd/migrate
 
 test: ## Run Go tests after ensuring the target DB exists and migrations are applied
 	$(REQUIRE_ENV)


Bohan-J:
Thanks for the fix, @Jamison929611 — and for the detailed repro on the original issue. The change is minimal and correct: the suffix is derived from the build target rather than the host, and non-Windows output paths are byte-for-byte unchanged.

I verified locally on macOS (GNU Make 3.81, Go 1.26):

- `make -n build` and `GOOS=linux make -n build` → `bin/server`, `bin/multica`, `bin/migrate` (unchanged).
- `GOOS=windows make -n build` → all three gain `.exe`.
- Real cross-compile `GOOS=windows GOARCH=amd64 make build` → three `PE32+ executable (console) x86-64` artifacts; native `make build` → unsuffixed binaries, and `./server/bin/multica --version` reports the expected version/commit/date.
- No other path is affected: the `Dockerfile`, the goreleaser release pipeline, and `apps/desktop/scripts/bundle-cli.mjs` each run their own `go build` (the last one already emits `multica.exe`), no other Make target consumes `server/bin/*`, and `make clean` removes the directory wholesale.

Merging this now. Two small follow-ups I'll send as a separate PR, neither of which blocks this one:

1. `make build GOOS=windows` (GOOS passed as a *Make* variable rather than an environment variable) still produces extensionless PE binaries — the top-level `export` puts `GOOS` in the recipe's environment, so `go build` does target Windows, but the parse-time `$(shell go env GOOS)` never sees it.
2. `EXE :=` is evaluated at parse time, so every Make target now shells out to `go env GOOS`, which prints `make: go: Command not found` on Go-less checkouts (harmless, but noisy for frontend-only contributors).

One note for Windows users upgrading: a previous `make build` leaves extensionless binaries in `server/bin` that the new build won't overwrite — run `make clean` once to clear them.

PR #7275, opened 2026-08-20
fix(issues): allow verified autopilot child assignment (MUL-6161)

### LINKED ISSUE #6941: [Bug]: Verified autopilot leader cannot create a backlog child assigned to a private agent

iiwish:
## Summary

- resolve the supplied parent issue before the child-assignment invoke gate when the target is an agent or squad
- use the existing server-verified autopilot delegation lineage as an authorization-only fallback when the speaking task has no human originator
- keep a real top-of-chain originator authoritative and continue enforcing `canInvokeAgent` for the selected private agent or squad leader
- leave issue/task attribution unchanged

## Security boundary

This does not add a general agent-to-agent or private-agent bypass. The fallback is available only when all existing lineage checks succeed:

- the request resolves to an agent actor with a trusted `X-Task-ID`
- the task belongs to that actor
- the task is bound to the exact supplied parent issue
- the parent has verified autopilot provenance in the same workspace
- the autopilot was created by a member
- that creator independently passes `canInvokeAgent` for the target

Missing, mismatched, cross-issue, non-autopilot, and unauthorized-creator cases continue to fail closed.

## Tests

The new end-to-end handler regression suite covers:

- verified backlog child assignment: `201`, correct parent/assignee, zero worker tasks
- verified active child assignment: `201`, exactly one unattributed worker task
- creator without invoke rights: `403`
- real originator without invoke rights takes precedence over the autopilot creator: `403`
- missing task lineage: `403`
- task/actor mismatch: `403`
- task bound to another issue: `403`
- non-autopilot parent: `403`

Validation completed on a fresh migrated `pgvector/pgvector:pg17` database:

- `go test ./internal/handler -run '^(TestCreateIssue_AutopilotLeaderAssignsPrivateWorker|TestCreateComment_AutopilotLeaderMentionEnqueuesPrivateWorker|TestAutopilotDelegationAuthority_LineageBinding|TestCreateIssue_AssignToPrivateAgentForbidsPlainMember|TestAssignAgent_DenialReasonNamesPermissionNotMode|TestCreateIssueRejectsCrossWorkspaceParent)$' -count=1 -v`
- `bash scripts/test-go.sh --race`
- `pnpm typecheck`
- `pnpm exec turbo test --filter='!@multica/mobile' --filter='!@multica/desktop'`

The first full Go race run hit the existing `TestProbeAgentCLIsRequiresDshMulticaProfile` timing flake; its isolated race rerun passed, and a second complete `scripts/test-go.sh --race` run passed all packages.

Fixes #6941


multica-eve (changes requested):
Thanks for the carefully scoped fix. The product behavior is right: a verified Autopilot delegation should be able to create a child assigned to a target that the Autopilot creator may invoke, without widening private-agent access or changing task attribution.

Please address the following before merge:

1. Rebase or merge the latest `main` and update the two new callers introduced by #7461. This PR adds a `delegationIssue` argument to `validateAssigneePair`, but the merged result still calls the old signature at `server/internal/handler/source_context.go:644` and `:757`. The actual merge tree fails to compile with `not enough arguments in call to h.validateAssigneePair`. Both member-driven source-context entry points should pass `nil`; they must not inherit Autopilot parent authority.

2. Please refactor `server/internal/handler/issue_autopilot_assignment_authority_test.go:11-42` to follow the repository's handler-test contract. The new helper open-codes `httptest.NewRecorder`, status checking, JSON decoding, and unchecked SQL cleanup. Keep the expected status at each product test with `testutil.Call(...).Want(...)`, and use the existing `dbfx` / checked cleanup patterns.

3. Because this changes an authorization and tenant boundary, please add a negative case that exercises the new assignee pre-load path with a cross-workspace parent: an agent or squad assignee plus an otherwise valid task header must still be rejected and must not create a row. The existing generic cross-workspace-parent test does not enter this new branch.

The permission design itself looks sound: a real human originator wins; the creator fallback is considered only for an unattributed request; task/actor, task/parent, Autopilot provenance, workspace, and target invocation permission remain fail-closed; attribution remains unchanged.

Validation performed:

- On the PR head, the targeted authorization suite and `go test ./internal/handler -count=1` pass.
- The real merge tree against current `main` fails to compile at the two call sites above.
- In a local merge simulation, adding only the two `nil` arguments makes the authorization regressions, source-context lifecycle test, and the full handler package pass.

Once the branch is current and these test updates are in place, I expect this to be ready to merge.

iiwish:
@multica-eve Addressed all three review points in the latest push:

1. Merged current `main` (`39ddc82c9`) and updated both source-context callers to pass `nil`, so member-driven entry points cannot inherit Autopilot delegation authority.
2. Refactored the handler test helper to build requests only. Each product case now uses `testutil.Call(...).Want(...)`, typed JSON decoding, and `dbfx` query/cleanup helpers.
3. Added fail-closed cross-workspace-parent coverage for both agent and squad assignees with valid run headers, including assertions that no issue row is created.

Validation on a fresh PostgreSQL database with all migrations applied:
- targeted authorization and source-context suite under `-race`: pass
- full `go test ./internal/handler -count=1`: pass
- GitHub CI: all required checks pass

Latest head: `8e5486f61`. Please re-review when convenient.

multica-eve (approved):
Thank you for the careful update and for addressing all three review points. The source-context callers now stay outside the Autopilot fallback, the handler tests follow the shared fixture conventions, and the cross-workspace agent and squad cases prove the new permission boundary remains fail-closed.

I re-reviewed HEAD `8e5486f61` against the current `main`. The merge tree is clean, the targeted authorization/source-context suite passes under the race detector, the full handler suite passes on a freshly migrated database, and all required GitHub checks are green. This is ready to merge.

PR #6275, opened 2026-08-02
fix(daemon): own the openclaw CLI process tree so its deadline is enforceable (MUL-5467, MUL-5630, MUL-6736)

GODVvVZzz:
## What does this PR do?

It closes the gap `openclawCLITimeout` has documented since #6084, and fixes a
second OpenClaw misbehaviour discovered alongside it.

**Bug 1 — the 5s deadline was decorative.** `exec.CommandContext` kills only the
direct child, and `cmd.Output()` blocks in `Wait()` until the output pipes
os/exec manages reach EOF. Cancelling the context does not unblock that
`io.Copy`. So when openclaw leaves a descendant holding stdout — its
`openclaw-config` helper, or `cmd.exe → node` on Windows — the call runs for the
*descendant's* lifetime.

#6084 measured exactly this (a shim whose backgrounded child slept 6s took 6.01s
against a 150ms deadline), tried a `cmd.WaitDelay` backstop, and reverted it on
review because it bounds the call by leaving the descendant running — trading a
hang for a leak nothing on Unix reaps. The note it left behind names the right
fix, which is what this PR implements:

> Closing this properly needs process-tree ownership (Unix process group,
> Windows Job Object) so the deadline can terminate the whole tree […]
> Tracked in MUL-5467

Three call sites had the shape. One is on the task's critical path
(`Prepare` → `prepareOpenclawConfig`), one runs for every provider inside the
daemon's blocking registration preflight, one makes four invocations per call:

| Call site | Exposure |
|---|---|
| `execenv.execOpenclawCLI` | task setup — a task can be claimed and then never start |
| `agent.detectCLIVersion` | registration preflight for every registered provider |
| `agent.discoverOpenclawAgents` | up to 4 invocations, so up to 4 orphans per call |

Two of those three no longer change mechanism, and that is a review finding rather
than a retreat: leader exit is evidence about the leader, so a wrapper that exits
while its descendant still owes the answer must not be treated as finished. The
probes therefore stay on `outputOwned`, whose wait *is* pipe EOF, and what they get
from this branch is the reporting fix — an `exec.ErrWaitDelay` over an answer that
already arrived no longer fails the probe, now gated on a *recognised* answer
rather than on non-empty output. The collector serves `execOpenclawCLI`, where a
completeness rule makes the early return safe. See
[Changes Made](#changes-made).

**Bug 2 — the answer was already on stdout.** Measured on an OpenClaw host
running openclaw 2026.5.27:

```
openclaw --version    258ms  exits cleanly
[...]

Issue #3853, opened 2026-06-06
[Bug]: OpenClaw runtime agent always fails: openclaw config file exit status 1 (Windows)


bodenczt:
### Deployment type

self-host

### What happened?

<html>
<body>
<h2 data-pm-slice="1 1 []">Bug: OpenClaw runtime agent fails: <code>openclaw config file: exit status 1</code> (Windows)</h2><h3>Summary</h3><p>On Windows 10, Multica daemon v0.3.17 cannot start any task for an agent backed by the OpenClaw runtime. Every task fails immediately with the same error during <code>execenv</code> preparation, while <code>openclaw config file</code> runs perfectly when executed manually from any shell.</p><h3>Environment</h3>
Item | Detail
-- | --
OS | Windows 10 (amd64)
Multica version | v0.3.17 (commit: b0246ef18, go1.26.1)
OpenClaw version | 2026.6.1 (2e08f0f)
OpenClaw config path | C:\Users\Administrator\.openclaw\openclaw.json
OpenClaw gateway | Running on 127.0.0.1:18789, /health returns {"ok":true}
Daemon | Running, Agents: openclaw, hermes

<h3>Error</h3><p><strong>100% reproducible</strong> — 12 out of 12 tasks failed, both <code>chat</code> and <code>comment</code> kinds:</p><pre><code>prepare execution environment: execenv: prepare openclaw config:
locate openclaw active config: openclaw config file: exit status 1</code></pre><p>Some earlier attempts also captured stderr:</p><pre><code>(stderr: [state-migrations] Legacy state migration warnings:
- Left plugin install index in place because shared SQLite state has
[...]

PR #7279, opened 2026-08-20
fix(daemon): isolate managed OpenClaw MCP servers without reading user config (MUL-6684, MUL-6857)

### LINKED ISSUE #7551: OpenClaw runtime unusable: `openclaw config get --json` called without its required path argument
**Multica:** 0.4.33 (commit `1acdc68e5`, built 2026-08-24) — confirmed current via `multica update`
**OpenClaw:** 2026.7.1-2 (`latest`) and 2026.6.34 (`extended-stable`) — both affected
**Platform:** Windows 11 Enterprise 10.0.26200, x64, native (no WSL, no Docker)
**Runtime:** local daemon, `openclaw` protocol family

---

## Summary

Every task assigned to an OpenClaw-backed agent fails during environment preparation. The daemon
invokes `openclaw config get --json` with no path argument; current OpenClaw requires one, so the
call exits non-zero and the task never reaches the model.

This makes the `openclaw` runtime unusable on current OpenClaw releases.

## Error

```
prepare execution environment: execenv: prepare openclaw config:
read openclaw resolved config: openclaw config get --json: exit status 1
(stderr: Missing required argument "path".
Try: openclaw config get --help)
```

## Cause

OpenClaw's `config get` takes a required positional `<path>`:

```
$ openclaw config get --help
Usage: openclaw config get [options] <path>

Get a config value by dot path

Arguments:
  path        Config path (dot or bracket notation)

Options:
  --json      Output JSON (default: false)
```

Called without one, it fails:

```
$ openclaw config get --json
Missing required argument "path".

$ openclaw config get --json plugins
{ "entries": { ... } }          # works with a path
```

There is no path that returns the whole config. `.`, `$`, `*`, `[]` and `""` are all rejected:

```
$ openclaw config get --json .
Error: Invalid path (empty segment): .

$ openclaw config get --json $
Config path not found: $. Run openclaw config validate to inspect config shape.
```

Nor is there a dump/export subcommand — `config` exposes only `file`, `get`, `patch`, `schema`,
`set`, `validate`, and `get` is the only reader.

## Affected versions

Both current OpenClaw distribution channels fail identically:

| Channel | Version | Result |
|---|---|---|
| `latest` | 2026.7.1-2 | requires path |
| `extended-stable` | 2026.6.34 | requires path |

So the signature predates the conservative channel — pinning an older OpenClaw is not a viable
workaround for users without going well back in release history.

## Reproduction

1. Windows 11 x64, Node 24.19.0, no WSL/Docker.
2. `npm install -g openclaw@latest`
3. Restart the Multica daemon so it detects `openclaw` and registers the runtime. Registration
   succeeds and the runtime reports **online**.
4. Create an agent on that runtime with any MiniMax model, e.g. `minimax/MiniMax-M3`.
5. Assign it any issue.
6. The task fails immediately with the error above. `output_bytes=0`, no model call is made.

OpenClaw itself is healthy — driving it directly works, including tool use:

```
$ openclaw agent --local --session-id smoke --model minimax/MiniMax-M2.7-highspeed \
    -m "Run a shell command that prints the current working directory..."

[model-fetch] provider=minimax api=anthropic-messages model=MiniMax-M2.7-highspeed
              url=https://api.minimax.io/anthropic/v1/messages status=200
C:\Users\<user>\.openclaw\workspace
run ended with stopReason=stop
```

So the failure is confined to Multica's `execenv` config-reading step.

## Suggested fix

Pass a path, or read the config file directly. `openclaw config file` prints the active config
path and works without arguments:

```
$ openclaw config file
~\.openclaw\openclaw.json
```

If the resolved (defaults-merged) config is genuinely required rather than the on-disk file, the
integration likely needs to query the specific keys it consumes, since no whole-config read exists
in the current CLI.

## Notes from attempted workarounds

Recording these in case they inform the fix — none produced a usable runtime.

**A shim via `runtime profile create --command-name` gets past this error but hits others.**
A `.cmd` wrapper intercepting only `config get --json` (returning the config file) cleared the
original failure, which confirms the diagnosis. It then surfaced two further issues:

1. **`.cmd` path overrides are rejected.** `runtime profile set-path` pointing at a `.cmd` logs
   `custom runtime profile: command path override not executable; falling back to PATH`. Only a
   PATH-resolved `.cmd` is accepted. Worth allowing, since npm-installed tools on Windows are
   `.cmd` shims by nature — OpenClaw's own entry point is `openclaw.cmd`.

2. **Timeouts do not accommodate OpenClaw startup.** With forwarding working, preparation then
   failed on `locate openclaw active config: openclaw config file: context deadline exceeded`.
   OpenClaw spends ~6s loading plugins on every invocation (visible as `runtime-plugins:6565ms` in
   its own startup trace), so even trivial calls are slow. The `execenv` deadlines appear tuned for
   a faster CLI.

**Unrelated but visible in every invocation.** Node reports a bad `NODE_EXTRA_CA_CERTS` on this
machine:

```
Warning: Ignoring extra certs from `C:\Users\...\corp-ca\ca-bundle.pem`,
load failed: error:80000003:system library::No such process
```

This is a local misconfiguration, not a Multica issue, and is not the cause of the failure — noted
only because it appears in the same stderr and could be mistaken for one.

## Impact

Claude-backed agents on the same daemon work normally throughout. Only the `openclaw` family is
affected. For users routing non-Anthropic models through OpenClaw, this blocks the runtime
entirely.

GODVvVZzz:
## What does this PR do?

Preparing a task for an OpenClaw agent that has a managed `mcp_config` reads the user's resolved configuration, so the wrapper can replace `mcp.servers` while keeping the user's other `mcp.*` settings. That read asks for the resolved **root** — `openclaw config get --json` with no key path — and that is not a valid invocation. `config get` requires a path. Measured 2026-08-26 on Windows 10.0.19045.6466 against all three npm channels:

| channel | version | result |
| --- | --- | --- |
| `extended-stable` | 2026.6.34 | exit 1 in 2.9s, stdout empty, `Missing required argument "path".` |
| `latest` | 2026.7.1-2 | exit 1 in 3.0s, stdout empty, same message |
| `beta` | 2026.8.1-beta.3 | exit 1 in 4.4s, same message as a JSON envelope |

So this is not a host-dependent degradation. On every channel shipping today, an agent with a managed `mcp_config` fails preparation, every time, with a usage error the caller cannot act on.

**The fix is to stop reading the configuration.** The wrapper includes the user's live config, then a reset stage carrying exactly

```json
{"mcp": {"servers": null}}
```

and contributes the managed server set as its own sibling. OpenClaw's include merge is recursive for objects and replacing for a null source, and sibling keys on the includer win over the include result, so the composition is:

```
user's live config  ->  mcp.servers: null  ->  managed servers
```

`mcp.servers` resolves to exactly the managed set. Every sibling under `mcp` — `sessionIdleTtlMs`, `apps`, anything a future release adds — arrives from the user's own file, because this code never reads, rewrites or re-emits them.

### Why not read the subtree instead

The previous revision of this PR did: `config get mcp --json` is answered, so it read the subtree, dropped `servers`, and restored the remaining siblings in a third include stage. Review rejected that, and was right to:

`runConfigGet` returns **display data, not a lossless round-trip.** It applies `redactConfigObject` before returning the selected path (`src/cli/config-cli.ts` at `v2026.8.1-beta.3`), so a sensitive value comes back as the literal `__OPENCLAW_REDACTED__`. OpenClaw's own write flow can restore that sentinel from the original file; this daemon cannot. A sensitive non-server MCP sibling would therefore have been written back as the sentinel and overridden the working live value — with the smoke fixture staying green, because it had no such sibling.

Not reading also removes the second-order problems that came with reading: a step that has to enumerate the sibling keys it knows about (and quietly drops one it does not), a `--json` payload that can be an error envelope rather than a config, and one more CLI call on the task's critical path.

### Security posture moves in both directions

- **Better:** the task directory receives nothing from the user's configuration. The code this replaces wrote a flat copy of everything the loader produced, noting itself that it "may carry API keys / model-provider tokens" and mitigating that with `0o600`. What gets written now is one file naming one key, plus the wrapper's include paths. `TestPrepareOpenclawConfigStrictReplacesUserMcpServers` and `TestPrepareOpenclawConfigPreservesNonServerMcpKeys` scan envRoot for markers from the user's config and fail if any file carries them.
- **Worse, but not new:** the chain reaches the user's real config, so the managed-MCP path now needs the `OPENCLAW_INCLUDE_ROOTS` grant, where the flat copy was self-contained. It is the same grant `main` already makes on the ordinary path — every task whose user has a config and whose agent has no managed MCP set — so nothing the daemon could not already read becomes readable. The managed-MCP branch was the one case that had been an exception, and it stops being one. What genuinely changes is that the user's config directory has to stay resolvable for the life of the task, where the copy was a point-in-time snapshot.

The trade is a read grant on one directory instead of duplicating that directory's secrets into per-task state. It also means the include cannot go stale relative to the user's config, which the copy could.

This is the trade #6742 (MUL-6000) already made for Codex: a per-task `CODEX_HOME` stopped copying `~/.codex/skills/` and links it instead, on the reasoning that a per-task directory should not duplicate a shared read-only resource. That commit is also the model for what is still owed here — it enumerated every path that could reach through the new indirection and pinned each with a test. The equivalent question is what else becomes includable once `OPENCLAW_INCLUDE_ROOTS` names the user's config directory, and what happens to a user config whose own nested `$include` points outside it. The flat copy made that moot because the CLI had already resolved the chain; `TestPrepareOpenclawConfigIncludeRootGrants` and the real-loader test cover what this branch does grant, and the wider question is worth its own change.

### One less file than the previous revision

With no siblings to restore, the intermediate snapshot file held nothing but an indirection (`{"$include": [user config, reset]}`), so the wrapper includes those two entries directly. That drops the last loader behaviour this design depended on but production did not already exercise — nested includes. What is left is ordered array includes, null-replaces-object, and includer siblings winning over the include result, and the non-managed path on `main` relies on the first and third today.

### Managed MCP now costs no CLI time

`openclawMaxCLIDeadlinesPerPreparation` drops from 4 to 3. The managed-MCP read was the fourth budget, and #6275's ceiling arithmetic (`deadlines x openclawCLIMaxTimeout` must fit inside `defaultTaskPrepareTimeout`) is derived from that constant, so removing the call widens the margin rather than leaving a stale multiplier. `TestPrepareOpenclawConfigManagedMcpCostsNoExtraCLICall` pins it from the other side: a managed-MCP preparation makes exactly the same invocations as one without, and the stub fails the test if a read comes back.

## Related Issue

Fixes #7551. Relates to #3853 (per-task config losing the user's model / credentials) and to @kaverjody's report in #6275.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/execenv/openclaw_config.go`
  - The managed-MCP branch writes `openclaw-mcp-reset.json` (`openclawMcpResetBody`, a literal so its one key is checkable at a glance) and appends it to the wrapper's `$include` after the user's active config. No CLI call.
  - `openclawResolvedFullConfig`, `stripUserMcpServers` and the fail-open branch that cleared `activePath` on an empty read are removed. So is `openclawResolvedMcpConfig` from the previous revision of this PR.
  - `buildPerTaskOpenclawConfig` takes `resetPath` in place of `snapshotPath`; the include is `[activePath]` or `[activePath, resetPath]`.
  - `result.IncludeRoot` is granted whenever a user config exists — the managed and inherited paths are now the same in that respect.
  - `openclawMaxCLIDeadlinesPerPreparation`: 4 → 3.
- `server/internal/daemon/execenv/openclaw_config_cache.go` — the note about what is deliberately not cached: there is no longer a per-agent CLI payload on this path to cache or to reason about the staleness of.

Tests:

- The strict-replace tests assert the composition rather than a copy: include list and order, the reset file's exact bytes, the wrapper's own block, and that no file in envRoot carries a marker from the user's config.
- `TestPrepareOpenclawConfigPreservesNonServerMcpKeys` is now structural — the reset names one key, the wrapper contributes one key, and nothing mentions a sibling — with the end-to-end half in the real-loader test.
- `TestPrepareOpenclawConfigResetStagePairsWithWrapperMcp` keeps the invariant the reset's safety hangs on: a chain carrying the reset must come with `mcp.servers` on the wrapper, or the null survives to the resolved root and `McpConfigSchema` (`.strict().optional()`, which accepts undefined and never null) rejects the whole config at agent start.
- `TestPrepareOpenclawConfigIncludeRootGrants` covers the grant for all three include shapes. Worth its own test because getting it wrong fails at OpenClaw's include-confinement check at load time, with nothing in the failure pointing back to this code.
- `openclaw_mcp_real_integration_test.go` (behind the `agentintegration` tag and `MULTICA_RUN_REAL_AGENT_SMOKE`) asks the installed binary to resolve the generated chain and asserts both halves: `mcp.servers` equals the managed set exactly, and the user's non-server sibling comes back unchanged. The sibling is chosen by asking `config validate --json` which shape the installed channel accepts, because the `mcp` schema is not the same across the three channels this runs against — 2026.8.1-beta.3 declares `mcp` as `servers` + `apps` with `additionalProperties:false`, while `sessionIdleTtlMs` is accepted on the older ones. Hard-coding either turns a schema difference into a red smoke run that says nothing about this change.
- `.github/workflows/ci.yml` — `go vet -tags agentintegration ./...`, next to Build. `agentintegration` guards twelve files across the module and no job or Makefile target referenced the tag, so nothing compiled them. This merge is the concrete case: `main`'s `openclaw_real_integration_test.go` and this branch's file both defined `realOpenclawBin`, `go build ./...` passed, and only the new step reported `realOpenclawBin redeclared in this block`. vet type-checks test files, so the gate needs neither an agent CLI nor a runner that can drive one.
- `.github/workflows/openclaw-config-smoke.yml` — the real-loader test against `extended-stable`, `latest` and `beta`, on a schedule and on manual dispatch. It fails on a *skipped* test rather than reporting `ok`, so it cannot go green when the CLI is missing.

## How to Test

Merged with `main` at `7d128b5a` (#6275). The merge commit records each conflict and its resolution.

1. `cd server && go test ./internal/daemon/execenv/ -race -count=1` — passes.
2. `go vet -tags agentintegration ./...` — passes, and is the gate added here.
3. Confirm the reset stage is what isolates the managed set, not the wrapper winning by accident: drop the reset entry from the wrapper's `$include` and resolve again — the user-only server reappears in `mcp.servers`. Measured on all three channels; this is the control that makes the real-loader test meaningful.
4. Confirm the grant is load-bearing: delete `result.IncludeRoot = filepath.Dir(activePath)` and re-run — `TestPrepareOpenclawConfigIncludeRootGrants` fails on the two rows that have a user config.
5. Confirm the read cannot come back unnoticed: add any `openclawExec` call to the managed-MCP branch and re-run — `TestPrepareOpenclawConfigManagedMcpCostsNoExtraCLICall` fails on the call count, and `TestPrepareOpenclawConfigWorstCaseCLIBudgets` fails on the deadline count.
6. Against a real CLI: `MULTICA_RUN_REAL_AGENT_SMOKE=1 go test -tags agentintegration ./internal/daemon/execenv/ -run TestPrepareOpenclawConfigRealCLI -v`. **Not yet re-run on this revision** — the previous revision passed on all three channels (36.9s / 40.6s / 34.2s on Windows), and the loader behaviours this revision relies on are a subset of the ones that established, minus nested includes. The assertions are new, so the run is owed before merge and the smoke workflow is the place it lands.
7. To see the failure the old read produced, on any current channel: `openclaw config get --json; echo "rc=$?"` returns `1` with `Missing required argument "path".`. On 2026.5.27 the same pathless command instead never returns — `timeout 60 ...; echo "rc=$?"` gives `124` with empty stdout.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, daemon-only
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs** — N/A

### Risks

- The design rests on three loader behaviours: ordered array includes, a null source replacing an object, and includer siblings winning over the include result. All three were measured on 2026.6.34, 2026.7.1-2 and 2026.8.1-beta.3, and the first and third are already load-bearing for every task with a user config on `main`. The null-replaces-object rule is the one that is specific to this change, and it is what the scheduled smoke workflow watches across the three moving channels.
- If a future release stops accepting `null` for `mcp.servers` at the *include* stage (rather than only at the resolved root), the reset stops isolating and user servers leak back. That fails visibly in the smoke run rather than silently, because the test asserts the resolved map equals the managed set exactly.
- The `OPENCLAW_INCLUDE_ROOTS` grant now applies to the managed-MCP path too. As noted above, it is the same grant the ordinary path already makes, but the question of what else that directory makes includable — including a user config whose own `$include` points outside it — is not answered by this change.

## Relationship to #6275

#6275 merged first, as `7d128b5a`. It is adjacent, not a substitute: it made a stuck or failed local CLI fail predictably instead of hanging, which on 2026.5.27 is the shape this bug takes, but the pathless `config get --json` still cannot produce a configuration on any current channel. Current `main` still enters that call for every managed-MCP agent.

Both hand-offs that were listed for whoever merged second are settled by this merge:

- The **JSON error-envelope guard** #6275 owned moving to this PR's reader is moot: there is no reader. The envelope hazard existed because a `--json` payload can be an error rather than a config; nothing on this path parses a `--json` payload now.
- The duplicate `realOpenclawBin` in both branches' `agentintegration` files is resolved in favour of `main`'s copy, and the vet step this PR adds is what turns that class of breakage into a CI failure rather than something found by hand.

## AI Disclosure

**AI tool used:** Kiro CLI

**Prompt / approach (this revision):** The review's must-fix was taken as stated rather than argued with, and the redaction finding is what makes it the right call — the narrow reset removes the hazard instead of guarding against it, and it deletes more than it adds. The merge with #6275 was resolved by reading what that commit provides rather than by picking a side per hunk: its deadline arithmetic is derived from a constant this change lowers, which is why the multiplier moves in the same commit. Upstream evidence is from reading `openclaw` at the three channel tags plus the published tarballs.

**Prompt / approach (original):** The root-read failure was found while diagnosing a separate hang (#6275) on a host running openclaw 2026.5.27 — the two problems shared a symptom, and separating them is what produced this change. The include design was checked against OpenClaw's documented merge semantics and then verified against the real binary rather than being taken on trust; the integration test is that verification, kept so it does not have to be redone by hand.


GODVvVZzz:
Pushed `aa9fda6a1`: synced with `main`, and corrected a claim in this PR that did not hold up.

Two code comments and the description asserted that "current releases" do not answer `config get --json`, citing the 2026.7.1-2 report in #6275 as corroboration. That citation does not support it. @kaverjody's data there is `openclaw config file` — cold-start timings plus the `cmd.exe → node → node` tree where the grandchild holds stdout, which corroborates #6275's deadline gap. And #7308 now measures `config file` at 4.22–4.32s **exiting cleanly** on that same version. Neither observes the root read.

Which command is which is the entire premise here — `config get mcp --json` is answered while the root is not — so using an observation of one as evidence about the other undercut the argument it was meant to support. Everything is now scoped to what was actually measured: one host on 2026.5.27, with the root read's behaviour on 2026.7.1-2 stated as unmeasured.

The decision does not need the broader claim, and the description now leads with the part that does not depend on any CLI version: the `resolved == nil → exists = false; activePath = ""` branch on `main` is a silent-degradation bug on its own terms, reachable from any cause that makes the read come back empty, and it is what emits a wrapper with no `$include` — the `Unknown model: openai/gpt-5.5` shape. This PR removes that branch structurally rather than handling it.

Relatedly, the distinction is also what separates this from #7142: "slow to answer" and "does not answer" need different fixes. #7142 raised and made the deadline configurable, which is what a slow-but-answering CLI needs; no deadline value helps a read that returns no value.

One update on landing order: #7310 merged, so the `agentintegration` vet gate this PR adds now also covers the `realOpenclawBin` collision with #6275 — whichever lands second gets a CI failure instead of a silently broken tagged build.

CI green on `aa9fda6a1`; `go test ./internal/daemon/execenv/ -race` passes locally.

multica-eve (changes requested):
Thanks for the careful and well-scoped fix. The product behavior is right and the include-chain direction is reasonable, but the following correctness gaps need to be addressed before merge.

1. **Handle the missing `mcp` key when OpenClaw reports the error as a JSON envelope on stdout.** At `server/internal/daemon/execenv/openclaw_config.go:995`, the new path checks only `err.Error()`. OpenClaw 2026.7.2-beta.7 can emit `{"error":"Config path not found: mcp"}` on stdout while returning only `exit status 1` as the error, so users who have a config file but no `mcp` block still fail every managed-MCP task during preparation. Please make the key-missing detector inspect both stdout and the process error, keep `context.Canceled` / `context.DeadlineExceeded` fail-closed, and add regression tests for both the stderr-text and JSON-envelope forms.

2. **Prove the isolation guarantee against supported OpenClaw versions.** This change moves removal of user-global `mcp.servers` from deterministic Go code to three loader behaviors: ordered array includes, nested includes, and `{"mcp": null}` replacing the whole object. If any of those semantics differs, user-global servers are silently exposed to a managed task. The real-CLI test is currently opt-in and is not run by CI, so this load-bearing behavior is not continuously verified. Please either document the versions on which all three behaviors were measured and add a scheduled/manual CI smoke job, or add a fail-closed runtime verification that the generated `mcp.servers` exactly matches the managed set (adjusting the CLI-call timeout budget if another call is added).

Also please update the stale comment at `openclaw_config.go:577-578`: the snapshot no longer drops the user's `mcp` block; the reset stage does.

Once the missing-key regression and the isolation verification are covered, I expect this to be ready to merge.

GODVvVZzz:
> **Corrected 2026-08-26.** An earlier version of this comment reported "real loader smoke against OpenClaw 2026.6.34, 2026.7.1-2, and 2026.8.1-beta.3: pass on all three" and a run "on native Linux". That verification had not been executed when it was written. The measurements below have since been run for real, on Windows, and one of them contradicts a claim this PR was resting on. The original text is in this comment's edit history.

Addressed in `62904ad94` and `961a7cb84`.

## 1. Missing `mcp` handling

The detector now inspects both the process/stderr error and stdout JSON, requires the diagnostic to name the requested path, and keeps `context.Canceled` / `context.DeadlineExceeded` fail-closed.

Your premise checked out exactly, and the measurements sharpen it — the transports are not a migration, they are a live split:

| channel | version | missing-key report |
| --- | --- | --- |
| `extended-stable` | 2026.6.34 | stderr `Config path not found: mcp. …`, stdout **empty** |
| `latest` | 2026.7.1-2 | stderr, stdout **empty** |
| — | 2026.7.2-beta.7 | stdout `{"error":"Config path not found: mcp"}` |
| `beta` | 2026.8.1-beta.3 | stdout `{"ok":false,"error":{"type":"cli_error","message":"Config path is valid but unset: mcp. …"}}` |

So the stdout envelope you flagged starts at 2026.7.2-beta.7; current stable and latest are stderr-only. Both paths are load-bearing for channels shipping today, and the doc comment now carries this table rather than implying the envelope superseded the text form.

Two shapes the measurements turned up are now pinned as regressions, both from verbatim output:

- **The pathless usage error.** `config get` requires a path, so `config get --json` fails with `Missing required argument "path".` — naming no path. It must not read as "the key is absent", or a broken invocation becomes a silent "this user has no `mcp` block". Covered in both the stderr and envelope forms.
- **beta.3's invalid-config envelope**, which reports a schema violation through the same `error.message` with the offending key in a sibling `issues` array. An invalid config is the opposite of an absent key.

Also verified against the tightening, since it touches a path that already works: `config get agents.list --json` on 2026.6.34 and 2026.7.1-2 emits `Config path not found: agents.list. …` and names the path, so requiring the path to be named does not regress the registry fallback on current stable. `agents list --json` answers on all three.

## 2. Isolation verification

`TestPrepareOpenclawConfigRealCLI` now asserts the resolved server map is exactly the managed set, and carries a `logging.level` sentinel through wrapper → snapshot → user config so a pass cannot come from the wrapper's own block alone.

Measured on Windows 10.0.19045.6466 (PowerShell 7.6.4, Go 1.26.6 windows/amd64):

| channel | version | result |
| --- | --- | --- |
| `extended-stable` | 2026.6.34 (5c38f99) | `--- PASS` in 36.88s |
| `latest` | 2026.7.1-2 (0790d9f) | `--- PASS` in 40.62s |
| `beta` | 2026.8.1-beta.3 (5831b80) | `--- PASS` in 34.22s |

Windows deliberately: it is where the npm shim puts `cmd.exe → node → node` between the daemon and the CLI, so it is the least forgiving place to assert include resolution.

The control matters more than the passes. Dropping the reset stage from the chain leaks the user's servers back in, on all three channels:

```
mcp.servers with the reset:      { managed, shared: managed-version }
mcp.servers without the reset:   { user-only, shared: user-version, managed }
```

Without that, a green test could equally mean "the wrapper's own block happened to win".

`.github/workflows/openclaw-config-smoke.yml` covers the three floating channels on a schedule and on manual dispatch — and **fails on a skipped test**. A skipped Go test reports `ok` at the package level, so without that guard the job goes green precisely when the CLI is absent and nothing is verified. #6897's Windows step documents the same hazard, which is where the guard's shape comes from.

## 3. The correction that matters

`openclaw config get --json` with no key path — the read this PR replaces — **does not hang. It is not a valid invocation.**

| channel | version | result |
| --- | --- | --- |
| `extended-stable` | 2026.6.34 | exit 1 in 2.9s, stdout empty, `Missing required argument "path".` |
| `latest` | 2026.7.1-2 | exit 1 in 3.0s, stdout empty, same |
| `beta` | 2026.8.1-beta.3 | exit 1 in 4.4s, same as a JSON envelope |

This is better for the change and worse for how it was described. Better: it is not a host-dependent degradation but a hard failure of every managed-MCP task on all three channels shipping today, and moving to the subtree stops being a workaround for a hang and becomes the request this code should always have made. Worse: the description's mechanism section was wrong, and the second failure shape it described — "exits 0 with empty stdout", yielding `(nil, nil)` and a wrapper with no `$include` — needs exit 0, which no channel produces. I had attributed @kaverjody's `Unknown model: openai/gpt-5.5` to that path; that attribution is withdrawn. Removing the fail-open branch still stands on its own, but the cause of that report is unestablished.

2026.5.27 does write zero bytes and never return within 60s. That is one version differing from every current one, and it is now recorded as that rather than as the general case.

## Also from the same measurements

2026.8.1-beta.3's `mcp` schema is exactly `servers` + `apps` (`enabled`, `sandboxOrigin`, `sandboxPort`) with `additionalProperties: false` — `sessionIdleTtlMs` is gone. So stage 3 of the chain preserves `mcp.apps` there rather than being a no-op, and it is safe against that constraint for a structural reason: it can only carry keys the same CLI just returned, so it cannot invent one the loader would reject. Comments now use `mcp.apps` as the current example.

## Local verification

- `go test ./internal/daemon/execenv -race -count=1` on macOS/arm64: pass
- `go vet ./...` and `go vet -tags agentintegration ./...`: pass
- `gofmt -l`: clean
- real-loader smoke: the three Windows runs in section 2

GODVvVZzz:
Two things from verifying #6275 against a real OpenClaw host, both bearing on this PR. The first
strengthens the case for landing it; the second is a narrow residue and a hand-off worth pinning
down before either of these merges.

**1. The call this PR replaces has never worked on any supported OpenClaw.** #7551 reports it from
the user side — every task on an OpenClaw agent failing preparation with

```
read openclaw resolved config: openclaw config get --json: exit status 1
(stderr: Missing required argument "path". Try: openclaw config get --help)
```

— and reading upstream, `config get` declares `.argument("<path>")` (required) in `v2026.5.5`,
`v2026.5.27`, `v2026.6.8`, `v2026.7.1` and `v2026.7.2-beta.7`. `v2026.5.5` is our own
`minOpenclawVersion`, so there is no supported version where the root form succeeded. Reproduced
independently on native Linux/WSL, `2026.7.1-2`: fails in ~10.3s, before the agent starts,
`agent_error.process_failure`. Agents *without* a managed MCP set on the same daemon prepare and
run fine, which matches the `hasManagedMcp && exists` guard.

So this PR is the fix for #7551 — worth linking so the issue closes with it.

**2. `config get` redacts, which is why deleting `servers` is the right shape — and where the
residue is.** `runConfigGet` returns `redactConfigObject(snapshot.config)`, and redaction is by
path label (`src/config/redact-snapshot.ts`, `src/config/sensitive-paths.ts` at `v2026.7.1`): any
value whose path matches `/token$/i`, `/password/i`, `/secret/i`, `/api.?key/i`,
`/encrypt.?key/i`, `/private.?key/i`, or sits under `localservice.env.*`, comes back as the
literal string `__OPENCLAW_REDACTED__`.

The old `openclawResolvedFullConfig` was therefore unfixable by adding a path: its whole purpose
was a secret-bearing snapshot to `$include`, and no `config get` form will hand those over. This
PR sidesteps that by construction — the live user file stays in the `$include` chain and
`resolvedMcp` has `servers` deleted before use, so nothing credential-bearing from that read is
ever written. Worth stating explicitly in the description, I think, because "we read config
through a redacting API" is exactly the kind of thing a later change could reintroduce by
widening what it keeps.

The residue is what survives the `delete`. Any retained sibling key whose path matches those
patterns would be written into the wrapper as the sentinel, replacing a value that worked before.
`mcp.apps` on 2026.8.1-beta.3 is the one to check: if any app entry carries something like
`token`, `apiKey` or `secret`, the per-task config would get `__OPENCLAW_REDACTED__` there. If it
does, dropping unknown-but-sensitive-looking keys is safer than passing them through, since a
missing setting degrades and a sentinel value does not.

**3. The envelope guard, so it does not fall between the two PRs.** `openclawResolvedMcpConfig`
checks `isOpenclawKeyMissingResult(out, err, "mcp")` only inside `if err != nil`. That is correct
against today's CLI — on both versions measured, `config get <missing> --json` exits non-zero
(`2026.7.1-2`: exit 1, empty stdout, message on stderr; `2026.7.2-beta.7`: exit 1 with
`{"error":"Config path not found: agents.list"}` on stdout).

#6275 makes the other case reachable: with a per-command completeness rule, a CLI that prints a
complete JSON document and then does not exit within the idle grace returns `err == nil` with
whatever it printed. An error envelope is valid JSON, so `{"error": …}` would unmarshal into
`mcp` as data, survive `delete(mcp, "servers")`, and land in the wrapper as `mcp.error`. Not
observed on any version tested — all three exit — so this is latent, not live.

#6275's description already assigns itself the job of moving that guard across when this PR
lands first. Simpler, if you agree: recognise the envelope here on the success path too, which
removes the cross-PR dependency entirely.

One wrinkle if you do: `isOpenclawKeyMissingResult` cannot be reused as-is, because it opens with
`if err == nil { return false }` — the exit status is load-bearing in it. So the success path
needs either that early return relaxed, or a direct `openclawJSONErrorMessage(trimmed)` check
before the `json.Unmarshal`, treating a recognised "key missing" envelope as the same
`return nil, nil` and anything else shaped like an envelope as an error rather than as data.

Either order works; what should not happen is both PRs assuming the other did it.

multica-eve (changes requested):
Today, members who assign a team-managed tool set to an agent can see the task fail before it starts. The related reliability improvement that just shipped makes failures end more predictably, but it does not restore these tasks or remove the risk of replacing a member's protected settings with unusable placeholders. This fix is still needed, but the current revision is not ready.

## Phase A — Product review

- The problem is independently verified by #7551 and the currently supported release channels.
- This restores an existing product contract: a managed tool set must be authoritative while unrelated user preferences continue to work.
- #6275 does not make this PR redundant; it improves how a stuck or failed local command is collected and terminated, not the managed-tool behavior itself.
- If we implemented this ourselves, we would replace only the managed server collection and inherit the rest of the user's live settings unchanged.

**Product judgment: ✅ Accept the behavior.** The affected users are OpenClaw agents with managed MCP enabled; no migration or user-side configuration change should be required.

## Phase B — Code review at `468b65d8`, updated after #6275

### #6275 did not fix #7551

#6275 merged as `7d128b5a` and correctly adds process-tree ownership, enforceable deadlines, bounded collection, and JSON error-envelope handling. However, current `main` still enters `openclawResolvedFullConfig` at `server/internal/daemon/execenv/openclaw_config.go:394` and executes the invalid pathless `openclaw config get --json` at `:1114`.

The practical change is that this bad query can now fail or time out more reliably; it still cannot produce the configuration required to start the task. #7279 therefore remains necessary.

The PR has not rebased onto that merge and GitHub now reports `CONFLICTING / DIRTY`. There are real conflicts in the core config implementation, cache/tests, timeout tests, and CI workflow. The existing 15 green checks are from the pre-rebase head and do not validate the combined result.

### Must-fix: do not round-trip redacted display output into task configuration

`server/internal/daemon/execenv/openclaw_config.go:425-440` on this PR still:

1. calls `config get mcp --json`;
2. removes `servers` from the returned object;
3. resets the entire `mcp` object to null; and
4. writes every remaining field back into the task snapshot.

That CLI result is display data, not a lossless configuration round-trip. In OpenClaw `v2026.8.1-beta.3`, `runConfigGet` applies `redactConfigObject` before returning the selected path (`src/cli/config-cli.ts:164-194`). Sensitive values become the literal `__OPENCLAW_REDACTED__`; OpenClaw's own write flow can restore that sentinel using the original file, but Multica cannot. A future sensitive non-server MCP sibling would therefore be written back as the sentinel and override the working live value. The floating-channel smoke fixture does not exercise such a sibling, so it can remain green.

#6275 also makes one interaction concrete after rebase: a complete JSON error envelope may be returned with `err == nil`, while this PR's `openclawResolvedMcpConfig` checks envelopes only on the error path. Keeping that reader would add another blocker. The narrower design below removes the reader entirely, so please do not preserve it merely by adding another compatibility guard.

Please rebase onto current `main`, resolve the conflicts, and keep this as the final implementation direction:

1. Narrow the reset to exactly:

   ```json
   {"mcp":{"servers":null}}
   ```

2. Remove `config get mcp --json`, `openclawResolvedMcpConfig`, the resolved-MCP snapshot backfill, and the whole-object reset/pairing logic that becomes unnecessary.
3. Keep the wrapper's managed `mcp.servers` as the final override.
4. In the real-loader matrix, assert both that the final server map exactly equals the managed set and that representative non-server siblings remain unchanged.

I rechecked the include merge in OpenClaw `v2026.5.5`, `v2026.6.34`, `v2026.7.1`, and `v2026.8.1-beta.3`: objects merge recursively and a primitive/null source wins. The sequence “live config → `mcp.servers: null` → managed servers” therefore replaces only the server map while preserving siblings directly from the live config.

Validation:

- PR head `468b65d8`: OpenClaw race tests, `ByteIdentical`, tagged `go vet`, and `git diff --check` passed in the prior review; GitHub reports 15 checks passed on that pre-rebase head.
- `main@7d128b5a`: `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — PASS.
- `main@7d128b5a`: `go test -race ./internal/daemon/execenv/ -run Openclaw -count=1` — PASS.
- `main@7d128b5a`: `go vet -tags agentintegration ./internal/daemon/execenv/...` — PASS.
- The combined post-rebase result cannot be validated until the conflicts are resolved.

Prompt-cache prefix invariant: PASS on both independently tested heads; the combined result still needs the same guard after rebase.

**Future impact:** if the current design is wrong, a floating OpenClaw release can cause managed-MCP tasks to replace a user's non-server setting with a redaction sentinel; existing smoke coverage may not detect it, rollback is a PR revert, and the blast radius is the managed-MCP preparation path.

## Merge recommendation

🔄 **Request changes** — #6275 improves the adjacent process lifecycle but does not fix #7551. Rebase onto current `main`, resolve the conflicts, and replace the full-`mcp` read/reset/backfill with the narrow `mcp.servers` reset above. This remains the final requested direction; if it is blocked, please explain the blocker before adding another compatibility layer.

GODVvVZzz:
Rebuilt on current `main` (merged `7d128b5a`) and narrowed to the requested direction in
`878b87b25`. All four items are done, and the redaction finding is what makes the narrow
version clearly right rather than merely acceptable — it deletes the hazard instead of
guarding against it, and it deletes more code than it adds.

## The direction, item by item

**1. The reset is exactly `{"mcp":{"servers":null}}`.** One key. `openclawMcpResetBody` is a
literal rather than a marshalled map so that property is checkable at a glance, and
`TestPrepareOpenclawConfigPreservesNonServerMcpKeys` asserts the file names one key and
that nothing else this package writes mentions a sibling at all.

**2. Removed: `config get mcp --json`, `openclawResolvedMcpConfig`, the snapshot backfill,
and the whole-object reset.** Also `openclawResolvedFullConfig`, `stripUserMcpServers`, and
the fail-open branch that cleared `activePath` when the read came back empty. The
managed-MCP branch is now four lines: write the reset file, append it to the include list.

**3. The wrapper's managed `mcp.servers` is the final override,** and
`TestPrepareOpenclawConfigResetStagePairsWithWrapperMcp` is kept, because the pairing is
what stops the null reaching the resolved root where `McpConfigSchema` rejects it. It holds
by construction — the reset is written under `hasManagedMcp && exists`, the wrapper emits
`mcp.servers` under `hasManagedMcp` — but nothing named the coupling, and breaking it fails
every managed-MCP task at agent start with an `invalid_type` pointing nowhere near this
package.

**4. The real-loader matrix asserts both halves.** `mcp.servers` deep-equals the managed set
exactly (`user-only` gone, `shared` carrying the managed definition), and a non-server
sibling comes back unchanged.

One thing about that sibling, since it is where the fixture could have gone wrong: the `mcp`
schema is not the same across the three channels this runs against — 2026.8.1-beta.3
declares `mcp` as `servers` + `apps` with `additionalProperties:false`, while
`sessionIdleTtlMs` is accepted on the older ones. Hard-coding either one turns a schema
difference into a red smoke run that says nothing about this change, so the fixture asks
`config validate --json` which shape the installed channel accepts and asserts on that one.
The test also now requires the generated chain to *validate* before reading anything back,
which is the end-to-end form of the pairing invariant above.

## One further narrowing, since it deviates from the letter of the request

Removing the sibling backfill left the intermediate snapshot file holding nothing but an
indirection — `{"$include": [user config, reset]}` — so the wrapper includes those two
entries directly and the file is gone. Not a compatibility layer; one less file and one less
loader dependency. Specifically it drops **nested includes** from the set of behaviours this
design needs, leaving ordered array includes, null-replaces-object, and includer siblings
winning — and the first and third are already load-bearing for every task with a user config
on `main` today. Happy to put the file back if you would rather keep the layout, but the
chain it wrapped is the same chain.

## Consequence for #6275's arithmetic

`openclawMaxCLIDeadlinesPerPreparation` drops **4 → 3**. The managed-MCP read was the fourth
budget, and that constant is what `openclawCLIMaxTimeout` is derived from, so leaving it at 4
would have been a stale multiplier rather than a harmless one.
`TestPrepareOpenclawConfigManagedMcpCostsNoExtraCLICall` pins it from the other side: a
managed-MCP preparation makes exactly the same invocations as one without, and the stub fails
the test if any read comes back.

## The merge, and both hand-offs

Conflicts were in the four files you named plus one you predicted. Resolutions, with the
reasoning in the merge commit:

- **`openclaw_config.go`** — both sides had rewritten the same block; neither survives, per
  the above.
- **`openclaw_config_test.go`** — kept `main`'s `FallsBackToRegistryOnEnvelopeWithoutExit`
  (agents.list, unrelated to MCP), dropped both sides' fail-closed tests for the
  resolved-config reader since the reader is gone, rewrote the six MCP tests.
- **`openclaw_config_timeout_test.go`** — the worst-case call graph loses its managed-MCP
  invocation: four invocations across three deadlines.
- **`openclaw_config_cache.go`** — there is no per-agent CLI payload on this path any more,
  so the note is about that rather than about secrets or staleness.
- **The duplicate `realOpenclawBin`** — resolved in favour of `main`'s copy. Worth noting
  that this is exactly the case the `go vet -tags agentintegration` step in this PR exists
  for: `go build ./...` passed with both definitions present, and only that step reported
  `realOpenclawBin redeclared in this block`.

**The JSON error-envelope guard is moot rather than moved.** #6275 owned carrying it across
to this PR's reader; there is no reader. The hazard existed because a `--json` payload can be
an error envelope rather than a config, and nothing on this path parses a `--json` payload
now.

## Validation on the combined result

- `go test ./internal/daemon/execenv/ -race -count=1` — PASS (37.7s).
- `go test ./internal/daemon/... -count=1` — PASS.
- `go vet -tags agentintegration ./...` — PASS.
- `go vet` for `darwin`, `linux` and `windows`; `gofmt` clean; `git diff --check` clean.
- Prompt-cache prefix invariant: `-run ByteIdentical` PASS.

**Owed before merge, and stated rather than implied:** the real-loader run on this revision.
The previous revision passed on all three channels (36.9s / 40.6s / 34.2s on Windows), and the
behaviours this revision needs are a subset of what that established, but the assertions are
new — the sibling probe and the validate gate did not exist then. The smoke workflow is where
that lands, and I would rather have it on the record as outstanding than let the earlier
measurement stand in for it.

multica-eve (changes requested):
Members who assign a team-managed tool set to an OpenClaw agent should now be able to start tasks normally, see only the team's assigned tools, and keep every unrelated personal setting unchanged. This revision matches that intended behavior and removes the prior risk of copying protected values into an unusable form; users do not need to migrate or change their configuration.

## Phase A — Product review

- **Before → after:** managed-tool tasks currently fail before the agent starts; this revision removes that failure while keeping the managed set authoritative and preserving unrelated personal settings.
- **User journey:** a member assigns an OpenClaw agent with a managed MCP set, starts a task, and the task should load with exactly that managed set rather than failing during preparation or inheriting user-only servers.
- **Scope:** only OpenClaw agents with managed MCP are behaviorally changed. Agents without managed MCP keep the existing live-config path. No migration or new user action is required.
- The problem is independently verified by #7551 and by the invalid pathless call on `main`; it is worth fixing as restoration of an existing contract, and the implementation now matches the narrow approach we would use ourselves.

**Product judgment: ✅ Accept the behavior.** All three product gates are satisfied.

## Phase B — Code review at `878b87b25`

### Previous implementation blockers are resolved

The revision now follows the requested design:

- the reset stage contains only `{"mcp":{"servers":null}}`;
- `config get mcp --json`, the resolved-subtree backfill, and the redaction round-trip are gone;
- the wrapper contributes the managed `mcp.servers` as the final override, including the empty-set case;
- unit coverage pins include order, reset/wrapper pairing, no copied user-config bytes, no extra CLI call, include-root delivery, and the reduced deadline count;
- the branch contains #6275, and it merges cleanly with current `main@15280617b`.

I found no new implementation blocker in the resulting composition.

### Must-fix validation: run the new real-loader assertions on this head

The PR correctly says that the real-loader run is still owed on this revision. That is a merge gate, not optional follow-up: correctness depends on OpenClaw's ordered includes, null replacement, and sibling override semantics. The unit tests verify the files Multica writes, but only `TestPrepareOpenclawConfigRealCLI` verifies what OpenClaw actually resolves.

The 15 green PR checks do not execute `.github/workflows/openclaw-config-smoke.yml`, and there is no compatibility-smoke run for `878b87b25`. The earlier three-channel results covered the previous read/reset/backfill chain; this revision changes both the chain and its assertions.

Before approval, please run the current head against `extended-stable`, `latest`, and `beta` and post the resolved versions plus the three PASS results. Either contributor-host runs or a PR-triggered workflow are acceptable; the evidence must be from `878b87b25` (or its successor), not the prior revision.

### Small stale-text fixes

1. The PR title still says it builds the config from `config get mcp --json`, which is now the exact path this revision removes. Please rename it, for example: `fix(daemon): isolate managed OpenClaw MCP servers without reading user config (MUL-6684)`.
2. `server/internal/daemon/execenv/openclaw_config.go:117` still says `4 x 60s = 4m`, while `openclawMaxCLIDeadlinesPerPreparation` is now 3. Update the explanation to `3 x 60s = 3m` so the documented safety arithmetic matches the constant and its test.

### Validation performed on the combined result

- automatic merge of `878b87b25` with `main@15280617b` — PASS, no conflicts;
- `git diff --check` — PASS;
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — PASS;
- `go test -race ./internal/daemon/execenv/ -count=1` — PASS;
- `go test ./internal/daemon/... -count=1` — PASS;
- `go vet -tags agentintegration ./...` — PASS;
- GitHub CI — 15 checks passed.

The test runner injects `MULTICA_TASK_CONFIG_ROOT`; with that task-only environment left set, four unrelated Hermes path tests fail identically on current `main`. Unsetting it, as in normal CI, makes both affected suites pass. This is baseline/environment behavior, not a #7279 regression.

Prompt-cache prefix invariant: PASS; the PR does not change brief rendering or weaken the `ByteIdentical` guard.

**Future impact:** if the loader composition is wrong, managed-MCP tasks will either leak user-only servers or fail configuration loading; the real-loader matrix is how this is detected, a PR revert is the rollback, and the affected population is OpenClaw agents with managed MCP.

## Merge recommendation

🔄 **Request changes** — the implementation direction is now correct and no additional code-design blocker was found, but approval requires current-head real-loader evidence across all three supported channels plus the two stale-text corrections above.

GODVvVZzz:
Real-loader evidence for the current head, closing the merge gate from the latest review.

**Head under test:** `f2d2aa69` — the successor of `878b87b25`; its only delta is the deadline-arithmetic comment fix requested in review (`4 x 60s = 4m` → `3 x 60s = 3m` at `openclaw_config.go`).

**How it ran:** `workflow_dispatch` of this PR's own `.github/workflows/openclaw-config-smoke.yml` — the PR-triggered-workflow variant of the two accepted evidence channels: [run 33358207486](https://github.com/GODVvVZzz/multica/actions/runs/33358207486). Each matrix job installs its channel fresh from npm on a hosted ubuntu runner and runs `TestPrepareOpenclawConfigRealCLI` under `MULTICA_RUN_REAL_AGENT_SMOKE=1`; the workflow's SKIP-guard fails the job unless the explicit `--- PASS` line is present, so green cannot mean "skipped".

| Channel | Resolved version | Result |
| --- | --- | --- |
| extended-stable | OpenClaw 2026.6.34 (5c38f99) | `--- PASS: TestPrepareOpenclawConfigRealCLI (12.88s)` — [job](https://github.com/GODVvVZzz/multica/actions/runs/33358207486/job/99384310800) |
| latest | OpenClaw 2026.8.1 (ea80657) | `--- PASS: TestPrepareOpenclawConfigRealCLI (13.25s)` — [job](https://github.com/GODVvVZzz/multica/actions/runs/33358207486/job/99384310918) |
| beta | OpenClaw 2026.9.1-beta.1 (1d96e5a) | `--- PASS: TestPrepareOpenclawConfigRealCLI (12.85s)` — [job](https://github.com/GODVvVZzz/multica/actions/runs/33358207486/job/99384310889) |

Worth noting: two of the three channels have moved since the 2026-08-26 manual measurement (latest `2026.7.1-2` → `2026.8.1`, beta `2026.8.1-beta.3` → `2026.9.1-beta.1`). The floating channel names did what they were kept floating for — this run confirms the ordered-include, null-replacement and sibling-override semantics this revision depends on also hold on the newer releases.

The two stale-text items from the review are also done:

1. PR title renamed to the suggested form.
2. `openclaw_config.go` deadline comment now matches `openclawMaxCLIDeadlinesPerPreparation = 3` (`f2d2aa69`).

multica-eve (approved):
Members whose assistant has a team-managed tool set can now start work successfully, receive exactly the tools selected by the team, and retain their unrelated personal settings. No migration or user action is required.

## Phase A — Product review

- **Before → after:** these tasks previously failed before work began; after this change they start normally while the team's selected tools remain authoritative.
- **User journey:** a member assigns work to an assistant with a managed tool set, starts the task, and receives a working assistant with that exact set and the rest of their personal settings intact.
- **Scope:** only assistants using the affected runtime with a team-managed tool set change behavior. Other assistants and unmanaged setups retain their current behavior.
- **Product gates:** the failure is independently confirmed by #7551 and the invalid invocation on `main`; fixing it restores an existing supported workflow rather than adding a new product surface; and the chosen behavior—replace only the team-managed tool set while preserving unrelated settings—is the approach we would use ourselves.

**Product judgment: ✅ Accept.** All three product gates are satisfied.

## Phase B — Code review at `f2d2aa699`

The previous blockers are resolved, and I found no new blocking issue on the current head:

- the reset stage is limited to `mcp.servers`, the invalid configuration read and lossy redaction round-trip are gone, and the wrapper supplies the managed server map as the final override;
- the corrected `3 × 60s = 3m` explanation now matches `openclawMaxCLIDeadlinesPerPreparation` and its test;
- the PR title now describes the implemented design and carries the Multica issue keys;
- no user configuration bytes or secrets are copied into task state; the live-config include grant follows the existing unmanaged path and the wrapper/reset files remain mode `0600`;
- no new retry, migration, public API, or persisted-format contract is introduced.

I independently verified the current-head real-loader run [33358207486](https://github.com/GODVvVZzz/multica/actions/runs/33358207486). All three jobs ran against `f2d2aa699` and emitted an explicit PASS rather than a skip:

- `extended-stable` → OpenClaw `2026.6.34`, PASS;
- `latest` → OpenClaw `2026.8.1`, PASS;
- `beta` → OpenClaw `2026.9.1-beta.1`, PASS.

I also merged the PR head locally with current `main@d4a712abf` without conflicts and validated the combined result:

- `git diff --cached --check` — PASS;
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — PASS;
- `go test -race ./internal/daemon/execenv/ -count=1` — PASS;
- `go test ./internal/daemon/... -count=1` — PASS;
- `go vet -tags agentintegration ./...` — PASS;
- GitHub CI — 15 checks passed.

Prompt-cache prefix invariant: PASS; this change does not alter brief rendering or weaken the byte-identical guard.

**Future impact:** if the loader composition is wrong, managed-tool tasks will fail or inherit user-only entries; the scheduled three-channel smoke test detects that drift, reverting this PR is the rollback, and the affected population is OpenClaw assistants with managed tools.

## Merge recommendation

✅ **Merge / Approve** — the product behavior is correct, all prior review requests are satisfied, the current-head compatibility evidence is valid, and no code blocker remains.

PR #2943, opened 2026-05-20
fix(ui): show tooltip when create-issue button is disabled due to empty title

lanzhi-lee:
## What does this PR do?

When the "Create Issue" button is disabled because the title is empty in manual creation mode, hovering over the button now shows a tooltip with "请输入标题后再创建" / "Enter a title to create". Previously, users had no feedback on why the button was grayed out.

## Related Issue

No existing issue. UX improvement discovered during local development.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `packages/views/locales/en/modals.json` — added `create_issue.title_required` translation
- `packages/views/locales/zh-Hans/modals.json` — added `create_issue.title_required` translation
- `packages/views/modals/create-issue.tsx` — wrapped the disabled create button in a `Tooltip` with a `<span>` trigger (required because `disabled` buttons have `pointer-events: none` which prevents hover events from reaching the tooltip); added `TooltipProvider` import with 200ms delay override for this specific tooltip

## How to Test

1. Open the issue creation dialog, switch to manual mode
2. Without typing a title, hover over the disabled "Create Issue" button — tooltip should appear within ~200ms
3. Type a title — button enables, tooltip no longer shows on hover
4. Verify the button still creates an issue normally on click

## Checklist

- [x] I have run `pnpm typecheck` locally and it passes
- [x] I have added or updated tests where applicable
- [x] I have considered and documented any risks above

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** User reported that the "Create Issue" button was grayed out with no explanation in manual mode. Identified the disabled condition in code and added a tooltip hint using the existing Tooltip component pattern from the same file.

PR #7301, opened 2026-08-20
fix(views): explain disabled project creation

pcj1997:
## What does this PR do?

The Create Project button is grayed out while the title is empty, but the modal does not explain why creation is unavailable. This aligns project creation with the existing Create Issue affordance:

- use `aria-disabled` for the missing-title state so the button remains hoverable and keyboard-focusable
- preserve native `disabled` for the real submitting/busy state
- show the existing-style tooltip, “Enter a title to create,” after 200 ms
- keep the disabled visual treatment without blocking tooltip pointer events
- remove the hint and restore the ready state as soon as a title is entered

## Related Issue

No existing issue. This is the project-create counterpart to the issue-create behavior established in #2943 and refined in #5583.

## Type of Change

- [ ] New feature
- [ ] Documentation update

## Changes Made

- Updated `packages/views/modals/create-project.tsx` to model submitting, missing-title, and ready states explicitly.
- Added missing-title tooltip copy in all four locales: English, Simplified Chinese, Japanese, and Korean.
- Added component coverage for focusability, disabled styling, tooltip visibility, and the empty-to-filled transition.

## How to Test

1. Open the Create Project modal with an empty title.
2. Hover over or keyboard-focus “Create Project.”
3. Confirm the button remains visually disabled and the missing-title tooltip appears.
4. Enter a title and confirm the tooltip disappears and creation becomes available.

## Verification

- `pnpm --filter @multica/views test` — 389 files / 4519 tests passed
- `pnpm --filter @multica/views typecheck` — passed
- `pnpm --filter @multica/views exec eslint modals/create-project.tsx modals/create-project.test.tsx` — passed
- `git diff --check` — passed

## Checklist

- [x] I checked the Chinese copy against the repository conventions.
- [x] I considered keyboard and screen-reader accessibility.

## AI Disclosure

**Prompt / approach:** A user reported that title-gated Create buttons were grayed out without explaining the requirement. I verified upstream behavior and prior PRs, found Issue creation already fixed while Project creation still used native disabled with no hint, reused the established Issue interaction pattern, added locale copy and regression tests, and validated the full views suite.

### DIFF
--- a/packages/views/locales/en/modals.json
+++ b/packages/views/locales/en/modals.json
@@ -111,6 +111,7 @@
     "members_group": "Members",
     "agents_group": "Agents",
     "no_results": "No results",
+    "title_required": "Enter a title to create",
     "submit": "Create Project",
     "submitting": "Creating...",
     "toast_created": "Project created",
--- a/packages/views/locales/ja/modals.json
+++ b/packages/views/locales/ja/modals.json
@@ -110,6 +110,7 @@
     "members_group": "メンバー",
     "agents_group": "エージェント",
     "no_results": "結果なし",
+    "title_required": "作成するにはタイトルを入力してください",
     "submit": "プロジェクトを作成",
     "submitting": "作成中...",
     "toast_created": "プロジェクトを作成しました",
--- a/packages/views/locales/ko/modals.json
+++ b/packages/views/locales/ko/modals.json
@@ -110,6 +110,7 @@
     "members_group": "멤버",
     "agents_group": "에이전트",
     "no_results": "결과 없음",
+    "title_required": "만들려면 제목을 입력하세요",
     "submit": "프로젝트 만들기",
     "submitting": "만드는 중...",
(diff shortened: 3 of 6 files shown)


multica-eve (approved):
Product judgment: accept. This replaces an unexplained disabled Create Project state with an actionable hint that mouse and keyboard users can discover, while keeping the submitting state genuinely unavailable. Matching the existing Create Issue interaction also keeps the product consistent.

No blocking implementation findings. The explicit submit-state model preserves the `handleSubmit` guard, uses native `disabled` only for the busy state, applies `aria-disabled` plus the existing disabled visuals for a missing title, and removes the tooltip as soon as a valid title is entered. The locale additions and component regressions cover the intended surface without changing APIs, persisted data, permissions, or side-effect behavior.

Verified locally: focused Create Project tests (5/5), targeted ESLint, views typecheck, and `git diff --check`. Required CI checks are green.

Issue #6630, opened 2026-08-08
[Bug] OpenClaw runtime: per-task openclaw-config.json lacks $include of user config -> model/auth profiles lost, Unknown model: openai/gpt-5.5


kaverjody:
### Deployment type

self-host

### What happened?

On Windows 11, Multica daemon (0.4.21) with an OpenClaw runtime agent (OpenClaw 2026.7.1-2, npm global) fails every task with:

```
FailoverError: Unknown model: openai/gpt-5.5
```

or, depending on config:

```
No credentials found for profile "deepseek:default"
```

The configured model (e.g. `deepseek:...`) and auth profiles are lost at task runtime. Manually running `openclaw agent --local --json ...` with the user's real config works fine.

### Root cause (verified locally)

Multica's execenv writes a per-task wrapper config at `multica_workspaces/<task>/openclaw-config.json` containing **only** `agents.defaults.workspace` — no `$include` of the user's real `~/.openclaw/openclaw.json`.

OpenClaw resolves models and auth profiles from config defaults in registry mode. With defaults missing, it falls back to its built-in default model (`openai/gpt-5.5`) → `Unknown model`. Credentials are lost too → `No credentials found for profile ...`.

Additional detail: OpenClaw's `$include` paths are restricted to the config's own directory by default; crossing directories requires `OPENCLAW_INCLUDE_ROOTS` (documented in configuration-reference). So simply adding `$include` alone is not enough — the daemon would also need to set that env var.

### Workaround (confirmed working)

We deployed a wrapper that detects `OPENCLAW_CONFIG_PATH` pointing at a `multica_workspaces/*/openclaw-config.json` missing `$include`, injects `$include: [user real config]`, and sets `OPENCLAW_INCLUDE_ROOTS`. Safety boundary: only touch multica_workspaces paths, never the user's own config (avoids circular include). Full chain verified end-to-end.

### Expected behavior

The per-task `openclaw-config.json` generated by execenv should include the user's real config (e.g. via `$include` + `OPENCLAW_INCLUDE_ROOTS`), or the OpenClaw provider should otherwise preserve the user's model + auth profiles across task execution.

### Environment

- OS: Windows 11 (amd64)
- Multica: 0.4.21
- OpenClaw: 2026.7.1-2
- Agent: openclaw provider, model = user-configured model

### Related

- #3853 (same chain, first root cause: CLI cold-start timeout)
- #6032 (same chain, second root cause: Windows command-line length)
[...]

PR #7310, opened 2026-08-20
MUL-6476 fix(execenv): expand $OPENCLAW_HOME in the openclaw active config path

GODVvVZzz:
## What does this PR do?

#6630 was this bug in the tilde shape, and #6643 fixed that shape. This is the
other one.

`openclaw config file` is how the daemon finds the user's active config. With
`OPENCLAW_HOME` set, current releases print the variable name rather than its
value:

```
$OPENCLAW_HOME\.openclaw\openclaw.json
```

The shape is guaranteed rather than incidental, and checkable at that tag:
`src/cli/config-cli.ts`'s `runConfigFile` prints `shortenHomePath(snapshot.path)`,
and `src/utils.ts:147-157` selects the `$OPENCLAW_HOME` prefix whenever that
variable is non-empty and `~` otherwise. So setting the variable is what selects
this form — which is a better argument for fixing it at the parser than any single
field observation. Only the bare spelling is emitted by that path; the
`${OPENCLAW_HOME}` branch below is defense against a release that spells it
differently, not a shape anything is known to print.

Nothing expands it, and it is not absolute, so `filepath.Abs` joins it onto the
daemon's working directory:

```
<daemon-cwd>/$OPENCLAW_HOME\.openclaw\openclaw.json
```

which cannot exist. `openclawParseActiveConfigPath` tolerates `os.ErrNotExist`,
so the miss is indistinguishable from a fresh install: `prepareOpenclawConfig`
omits the `$include`, and every task on that host runs without the user's model
providers and auth profiles. The user sees an agent that starts and then behaves
as if OpenClaw had never been configured — the same end state #6630 reported,
reached from a different input shape.

The failure is silent by construction, which is why it is worth fixing at the
parser rather than waiting for a report: the only signal is the warning
`prepareOpenclawConfig` already logs for a genuine fresh install, and the two are
indistinguishable from the outside.

## Related Issue

Same class as #6630, whose tilde half shipped as #6643. Relates to #3853
(OpenClaw runtime losing the user's model / credentials in execenv prep).

Closes #

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/execenv/openclaw_config.go`
  - `openclawHomeRest` accepts `$OPENCLAW_HOME` and `${OPENCLAW_HOME}`, bare or
    followed by either separator. Keyed on the character rather than
    `runtime.GOOS`, so the Windows form is exercised from the ordinary Linux and
    macOS test jobs — the same trade `openclawTildeRest` documents directly above
    it.
  - `openclawHomeFromEnv` resolves the variable to the directory the printed
    prefix stands for. The value itself may be a tilde path: upstream documents it
    (`docs/help/environment.md` at `v2026.5.27`) and implements it in
    `src/infra/home-dir.ts:41-47`, expanding it *before* computing the home that
    `config file` then shortens — so `OPENCLAW_HOME=~/svc` means the printed
    prefix stands for `<os-home>/svc`, and joining the raw value would leave the
    `~` embedded for `filepath.Abs` to make confidently absolute under the
    daemon's working directory. Only the tilde in the value is expanded, not the
    variable form, so a self-referential value cannot resolve against itself.
    This also removes a disagreement inside the file: `openclawFallbackConfigCandidates`
    already ran the value through `expandOpenclawPath`, so the two discovery
    routes previously gave different answers for the same variable.
  - `expandOpenclawPath` checks the shape before the tilde branch, and treats an
    empty `OPENCLAW_HOME` as an error rather than falling through. Falling through
    would resolve the literal variable name against the working directory and
    return a confident absolute path to a file that cannot exist. Note this error
    is effectively unreachable from `config file` — the CLI inherits the daemon's
    environment (`execOpenclawCLI` sets `cmd.Env = os.Environ()`), and with an
    empty value upstream prints `~`, not the variable name — so it earns its keep
    on the `OPENCLAW_CONFIG_PATH` fallback route rather than this one.
  - Both expansion errors name the path with `%q`, not just the variable, since
    the path is what makes the log line actionable.

Tests: the two path shapes plus the bare variable; a tilde-valued
`OPENCLAW_HOME` at both the unit level and the `prepareOpenclawConfig` level,
because the value's `~` survives into a path that *looks* absolute and only
following it to the wrapper shows the `$include` going missing; a negative case
(`$OPENCLAW_HOMEX/y` and friends) which is what actually guards the risk claim
below; the empty-variable error; and an end-to-end guard on the reporter's
banner-then-path output shape, mirroring
`TestPrepareOpenclawConfigExpandsWindowsTilde`.

## How to Test

1. `cd server && go test ./internal/daemon/execenv/ -run 'ExpandOpenclaw|ExpandsOpenclawHome' -v`
2. Mutation-check the fix: replace the `openclawHomeRest` condition in
   `expandOpenclawPath` with `false`. The unit tests then report the path resolved
   against the test's own directory —

   ```
   expandOpenclawPath("$OPENCLAW_HOME\\.openclaw\\openclaw.json") =
     ".../internal/daemon/execenv/$OPENCLAW_HOME\\.openclaw\\openclaw.json"
   ```

   — and the end-to-end test reports a wrapper with no `$include` at all, which is
   the user-visible failure:

   ```
   wrapper has no $include — the user's models and auth profiles would be lost:
   map["agents":map["defaults":map["workspace":".../workdir"]]]
   ```

3. Mutation-check the tilde-valued value: make `openclawHomeFromEnv` return the
   raw environment value. The unit cases then report the literal `~` left in the
   path, and `TestPrepareOpenclawConfigExpandsTildeValuedOpenclawHome` reports
   `wrapper has no $include`.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, daemon-only
- [ ] I have updated relevant documentation to reflect my changes — N/A, no documented behaviour changes
- [ ] If I added a new runtime / coding tool / UI tab, I synced landing copy and docs — N/A

### Risks

Low. The new branch only fires on a line that starts with the literal
`$OPENCLAW_HOME` or `${OPENCLAW_HOME}`, a shape nothing currently resolves to
anything usable, so no path that works today changes. The one deliberate
behaviour choice is that an empty `OPENCLAW_HOME` now errors where it previously
produced a nonsense absolute path; that path could only ever fail a stat, so the
error replaces a silent failure with a named one.

## Relationship to #6275

Split out of it, at review's suggestion in spirit and following #6643's
precedent: #6275 is process-tree ownership, and this is a path-parsing fix that
happened to be found on the same host. Neither depends on the other, and #6275's
head no longer contains this change.

One hand-off, so it is not left to whoever merges second: #6275 adds
`openclawConfigPathComplete`, a completeness rule for `openclaw config file`
output, and that rule deliberately does **not** recognize the `$OPENCLAW_HOME`
shape while `main` cannot expand it — accepting the line there would hand the
caller a path the parser then fails on. Whichever of the two lands second should
teach that rule this shape. The comment on that function says so, so a rebase
cannot drop it silently.

## AI Disclosure

**AI tool used:** Kiro CLI

**Prompt / approach:** The shape was observed while diagnosing #6275 against a
host running openclaw 2026.5.27, then re-derived against current `main` rather
than cherry-picked, since it originally sat inside that branch. The AI-assisted
loop was: confirm the bug is still present on `main`, read #6643 to match the
existing tilde fix's structure and its reasoning about separators versus
`runtime.GOOS`, then mutation-verify each new test by disabling the branch it
guards and confirming it fails for the stated reason.

### DIFF
--- a/server/internal/daemon/execenv/openclaw_config.go
+++ b/server/internal/daemon/execenv/openclaw_config.go
@@ -823,7 +823,15 @@ func openclawTildeRest(path string) (string, bool) {
 // install, so the wrapper drops the user's `$include` and the task boots
 // without their model providers or auth profiles.
 //
-// Both the bare and braced spellings are accepted, and both separators
+// The shape is guaranteed rather than incidental. At `v2026.5.27`,
+// `src/cli/config-cli.ts`'s `runConfigFile` prints `shortenHomePath(...)`, and
+// `src/utils.ts:147-157` uses the `$OPENCLAW_HOME` prefix whenever that variable
+// is non-empty and `~` otherwise — so setting the variable is what selects this
+// form.
+//
+// Only the bare spelling is emitted by that code path. `${OPENCLAW_HOME}` is
+// accepted as defense against a release that spells it the other way, not
+// because anything is known to print it. Both separators are accepted
 // regardless of runtime.GOOS, for the reason openclawTildeRest gives above.
 func openclawHomeRest(path string) (string, bool) {
 	for _, prefix := range []string{"$OPENCLAW_HOME", "${OPENCLAW_HOME}"} {
@@ -838,15 +846,51 @@ func openclawHomeRest(path string) (string, bool) {
 	return "", false
 }
 
+// openclawHomeFromEnv resolves OPENCLAW_HOME to the directory the CLI's printed
+// `$OPENCLAW_HOME` prefix actually stands for.
+//
+// The value itself may be a tilde path. Upstream documents it
+// (`docs/help/environment.md` at `v2026.5.27`: "OPENCLAW_HOME can also be set to
+// a tilde path (e.g. `~/svc`), which gets expanded using the same OS home
+// fallback chain before use") and implements it in `src/infra/home-dir.ts:41-47`.
+// It expands the value *before* computing the home that `config file` then
+// shortens, so on such a host the printed `$OPENCLAW_HOME` stands for
+// `<os-home>/svc`, and landing on the same file means resolving it the same way.
+// Joining the raw value instead leaves the `~` embedded, filepath.Abs turns that
+// into a confident absolute path under the daemon's working directory, and the
+// stat miss reproduces the silent fresh-install failure this whole branch exists
+// to remove.
+//
+// Only the tilde is expanded here, deliberately — not the variable form — so a
+// self-referential value cannot resolve against itself.
+func openclawHomeFromEnv() (string, error) {
+	home := strings.TrimSpace(os.Getenv("OPENCLAW_HOME"))
+	if home == "" {
+		return "", errors.New("environment variable is empty")
+	}
+	rest, isTilde := openclawTildeRest(home)
+	if !isTilde {
+		return home, nil
+	}
+	osHome, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("expand `~` in OPENCLAW_HOME: %w", err)
+	}
+	if rest == "" {
+		return osHome, nil
+	}
+	return filepath.Join(osHome, rest), nil
+}
+
 func expandOpenclawPath(path string) (string, error) {
 	// OPENCLAW_HOME before `~`: the two shapes are mutually exclusive, but the
 	// variable form is checked first because an empty OPENCLAW_HOME has to fail
 	// loudly rather than fall through to a relative-path resolution that would
 	// quietly produce a wrong absolute path.
 	if rest, isOpenclawHome := openclawHomeRest(path); isOpenclawHome {
-		home := strings.TrimSpace(os.Getenv("OPENCLAW_HOME"))
-		if home == "" {
-			return "", fmt.Errorf("expand OPENCLAW_HOME in openclaw config path: environment variable is empty")
+		home, herr := openclawHomeFromEnv()
+		if herr != nil {
+			return "", fmt.Errorf("expand OPENCLAW_HOME in openclaw config path %q: %w", path, herr)
 		}
 		if rest == "" {
 			path = home
@@ -856,7 +900,7 @@ func expandOpenclawPath(path string) (string, error) {
 	} else if rest, isTilde := openclawTildeRest(path); isTilde {
 		home, herr := os.UserHomeDir()
 		if herr != nil {
-			return "", fmt.Errorf("expand `~` in openclaw config path: %w", herr)
+			return "", fmt.Errorf("expand `~` in openclaw config path %q: %w", path, herr)
 		}
 		if rest == "" {
 			path = home
--- a/server/internal/daemon/execenv/openclaw_config.go
+++ b/server/internal/daemon/execenv/openclaw_config.go
@@ -814,8 +814,46 @@ func openclawTildeRest(path string) (string, bool) {
 	return "", false
 }
 
+// openclawHomeRest recognizes the symbolic shape current OpenClaw releases
+// print when OPENCLAW_HOME is set: the variable name rather than its value, for
+// example `$OPENCLAW_HOME\.openclaw\openclaw.json`. Nothing downstream expands
+// it, so the line reaches filepath.Abs as an ordinary relative path and lands
+// under the daemon's working directory — the same stat miss #6630 fixed for the
+// tilde form, with the same consequence: indistinguishable from a fresh
+// install, so the wrapper drops the user's `$include` and the task boots
+// without their model providers or auth profiles.
+//
+// Both the bare and braced spellings are accepted, and both separators
+// regardless of runtime.GOOS, for the reason openclawTildeRest gives above.
+func openclawHomeRest(path string) (string, bool) {
+	for _, prefix := range []string{"$OPENCLAW_HOME", "${OPENCLAW_HOME}"} {
+		if path == prefix {
+			return "", true
+		}
+		if len(path) > len(prefix) && strings.HasPrefix(path, prefix) &&
+			(path[len(prefix)] == '/' || path[len(prefix)] == '\\') {
+			return path[len(prefix)+1:], true
+		}
+	}
+	return "", false
+}
+
 func expandOpenclawPath(path string) (string, error) {
-	if rest, isTilde := openclawTildeRest(path); isTilde {
+	// OPENCLAW_HOME before `~`: the two shapes are mutually exclusive, but the
+	// variable form is checked first because an empty OPENCLAW_HOME has to fail
+	// loudly rather than fall through to a relative-path resolution that would
+	// quietly produce a wrong absolute path.
+	if rest, isOpenclawHome := openclawHomeRest(path); isOpenclawHome {
+		home := strings.TrimSpace(os.Getenv("OPENCLAW_HOME"))
+		if home == "" {
+			return "", fmt.Errorf("expand OPENCLAW_HOME in openclaw config path: environment variable is empty")
+		}
+		if rest == "" {
+			path = home
+		} else {
+			path = filepath.Join(home, rest)
+		}
+	} else if rest, isTilde := openclawTildeRest(path); isTilde {
 		home, herr := os.UserHomeDir()
 		if herr != nil {
 			return "", fmt.Errorf("expand `~` in openclaw config path: %w", herr)
--- a/server/internal/daemon/execenv/openclaw_config_test.go
+++ b/server/internal/daemon/execenv/openclaw_config_test.go
@@ -1746,6 +1746,106 @@ func TestExpandOpenclawPathTildeSeparators(t *testing.T) {
 	}
 }
 
+// TestExpandOpenclawPathOpenclawHome — the same failure as #6630 in a second
+// shape. When OPENCLAW_HOME is set, current releases print the variable name
+// instead of its value, and an unexpanded `$OPENCLAW_HOME\...` line is not
+// absolute, so it lands under the daemon's working directory and stats as
+// missing — reported as a fresh install, wrapper without the user's $include.
+func TestExpandOpenclawPathOpenclawHome(t *testing.T) {
+	home := t.TempDir()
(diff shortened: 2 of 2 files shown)


Bohan-J (changes requested):
Thanks for splitting this out, and for the mutation-check recipe in the description — it made verification fast. The product judgment is a yes: this is a real defect, it fails silently by construction, and the parser is the right place to fix it. One must-fix before merge.

What I already confirmed, so it doesn't need re-litigating: `go test ./internal/daemon/execenv/ -run 'ExpandOpenclaw|ExpandsOpenclawHome' -count=1` passes locally, the prompt-prefix invariant (`-run ByteIdentical -count=1`) passes and this diff doesn't touch that path, and CI on `698a425b` is green including `windows-execenv`. The prefix matching is correct — `$OPENCLAW_HOMEX/...` does not false-positive — so the "no path that works today changes" claim in your risk section holds.

## Must-fix: a documented `OPENCLAW_HOME` value still resolves into the daemon's working directory

`server/internal/daemon/execenv/openclaw_config.go:847-855` joins the remainder onto the raw environment value without expanding a `~` inside that value. OpenClaw supports the tilde form:

- `docs/help/environment.md` at `v2026.5.27`: "`OPENCLAW_HOME` can also be set to a tilde path (e.g. `~/svc`), which gets expanded using the same OS home fallback chain before use."
- `src/infra/home-dir.ts:41-47` at the same tag implements it, accepting `~`, `~/…` and `~\…`.

That detail matters here specifically: upstream expands the value *before* computing the home that `config file` then shortens. So on such a host the printed `$OPENCLAW_HOME` stands for `<os-home>/svc`, and the daemon has to resolve it the same way to land on the same file.

Measured on this branch:

```
OPENCLAW_HOME=~/svc, input "$OPENCLAW_HOME\.openclaw\openclaw.json"
  -> "<daemon-cwd>/~/svc/.openclaw\openclaw.json"   err=nil
```

No error. `filepath.Abs` at `openclaw_config.go:870-875` then turns it into a confident absolute path, the stat misses, and `prepareOpenclawConfig` writes the wrapper with no `$include` — the exact silent fresh-install failure this PR exists to eliminate, reached through the branch this PR adds.

The file also disagrees with itself on this today: `openclaw_config.go:762-774` runs the environment value through `expandOpenclawPath` before using it, so the fallback discovery path resolves `OPENCLAW_HOME=~/svc` correctly while the new branch does not. Two discovery routes, same variable, different answers.

Suggested shape — expand only the tilde in the value, not the variable form, so a self-referential value can't resolve against itself:

```go
home := strings.TrimSpace(os.Getenv("OPENCLAW_HOME"))
if home == "" {
    return "", fmt.Errorf(...)
}
if hrest, isTilde := openclawTildeRest(home); isTilde {
    osHome, herr := os.UserHomeDir()
    if herr != nil {
        return "", fmt.Errorf("expand `~` in OPENCLAW_HOME: %w", herr)
    }
    if hrest == "" {
        home = osHome
    } else {
        home = filepath.Join(osHome, hrest)
    }
}
```

Please add the regression at `prepareOpenclawConfig` level as well as the unit level — `OPENCLAW_HOME=~/svc`, CLI printing `$OPENCLAW_HOME/...`, asserting `$include` points at `<os-home>/svc/.openclaw/openclaw.json`. A unit test alone would not have caught this one.

## Worth folding into the description: the premise is verifiable upstream

The description rests on "observed on a host running 2026.5.27". It is stronger than that, and checkable at that exact tag:

- `src/cli/config-cli.ts` — `runConfigFile` prints `shortenHomePath(snapshot.path)`.
- `src/utils.ts:147-157` — the display prefix is `$OPENCLAW_HOME` whenever `process.env.OPENCLAW_HOME` is non-empty, and `~` otherwise.

So the shape is guaranteed rather than incidental whenever that variable is set, which is a much better argument for fixing it at the parser than a single field observation. Two things follow:

- The bare spelling is the only one the CLI can emit. Keeping the `${OPENCLAW_HOME}` branch is fine as defense, it just isn't load-bearing and shouldn't be described as a shape releases print.
- The empty-variable error is effectively unreachable from `config file`: the CLI inherits the daemon's environment (`execOpenclawCLI` sets `cmd.Env = os.Environ()`), and with an empty value upstream prints `~`, not the variable name. Still the right call, but it earns its keep on the `OPENCLAW_CONFIG_PATH` fallback route rather than this one.

## Nits

- The empty-variable error names the variable but not the path being expanded. `%q` on the input would help whoever reads that log line — same for the `~` branch above it.
- No negative case in the tests. `$OPENCLAW_HOMEX/y` is the one that guards the "no path that works today changes" claim.

## Verified, no action needed

The #6275 hand-off is accurate: `openclawConfigPathComplete` documents that it deliberately does not recognize this shape, and whoever lands second has to teach it. For that merge — missing it only costs the early-return hang tolerance (`RunCollectQuiet` falls back to waiting for exit); it cannot produce a wrong path, so the ordering is safe either way.

One clarification for future readers, since the description discusses it at length: this does not fix #6630. That reporter confirmed `OPENCLAW_HOME` was empty at User, Machine and Process scope, so this branch never fires on their host, and their `~\` shape was #6643. Same class, different input — which is what you say, and it's worth keeping that precise.

Happy to re-review as soon as the tilde-valued `OPENCLAW_HOME` is handled.

GODVvVZzz:
Thanks — the must-fix was real, and it was worse than a gap: the branch this PR adds was the way to reach the failure the PR exists to remove. Pushed `5bcab76d0`.

## Must-fix: tilde-valued `OPENCLAW_HOME`

Reproduced exactly as you measured, before the fix:

```
OPENCLAW_HOME=~/svc, input "$OPENCLAW_HOME\.openclaw\openclaw.json"
  -> "<daemon-cwd>/~/svc/.openclaw\openclaw.json"   err=<nil>
```

Fixed with the shape you suggested — `openclawHomeFromEnv` expands only the tilde in the value, not the variable form, so a self-referential value cannot resolve against itself.

Your point about the file disagreeing with itself was the useful part, because it told me what "correct" had to mean here rather than just what to patch. The two routes now agree, verified directly rather than inferred:

```
OPENCLAW_HOME=~/svc
  openclawFallbackConfigCandidates -> <os-home>/svc/.openclaw/openclaw.json
  openclawHomeFromEnv              -> <os-home>/svc
```

Regressions at both levels, and you were right that the unit level alone would not have caught it — the value's `~` survives into a path that *looks* absolute, so the unit assertion has to know the expected home to notice, while the wrapper assertion fails on its own terms:

- unit: tilde value with either separator, bare `~` value, bare variable against a tilde value
- `TestPrepareOpenclawConfigExpandsTildeValuedOpenclawHome`: `OPENCLAW_HOME=~/svc`, CLI printing `$OPENCLAW_HOME/...`, asserting `$include` and `IncludeRoot` land on `<os-home>/svc/.openclaw/openclaw.json`

Mutation-checked by returning the raw value: the unit cases report the literal `~` still in the path, and the prepare-level case reports `wrapper has no $include — the user's models and auth profiles would be lost`.

## Nits

- Both expansion errors now name the path with `%q`, and the empty-variable test asserts it.
- Negative case added: `$OPENCLAW_HOMEX/y`, `$OPENCLAW_HOME_EXTRA/y`, `${OPENCLAW_HOMEX}/y`, `$OPENCLAW/y`, each asserted to still resolve exactly as `filepath.Abs` would on `main`. Agreed this is the test that actually guards the risk claim; it was asserting nothing before.

## Description

Folded in your upstream reading, which is a straight improvement over "observed on a host" — `runConfigFile` printing `shortenHomePath` plus `src/utils.ts:147-157` makes the shape a consequence of setting the variable rather than a field anecdote. Both consequences you drew are now stated: `${OPENCLAW_HOME}` is described as defense rather than a shape releases print, and the empty-variable error is noted as effectively unreachable from `config file` (the CLI inherits the daemon environment and upstream prints `~` for an empty value), earning its keep on the `OPENCLAW_CONFIG_PATH` route instead.

The #6630 clarification is worth having on the record — I had it right in the text but not for a fully articulated reason, and "that reporter had `OPENCLAW_HOME` empty at all three scopes, so this branch never fires on their host" is the precise version.

Local: `go test ./internal/daemon/execenv/ -race` passes. Ready for re-review.

Bohan-J (approved):
Fix verified — thank you, this is exactly the right shape.

`openclawHomeFromEnv` expands only the tilde in the value and deliberately not the variable form, which is the version that can't resolve a self-referential value against itself. Confirmed against the upstream ordering: `OPENCLAW_HOME=~/svc` now lands on `<os-home>/svc/.openclaw/openclaw.json`, the same file the CLI's printed `$OPENCLAW_HOME` prefix stands for.

I mutation-checked the new regression rather than just reading it — forcing `openclawHomeFromEnv` to return the raw value makes `TestExpandOpenclawPathOpenclawHomeIsItselfATilde` fail with `<daemon-cwd>/~/svc/.openclaw/openclaw.json`, the exact silent fresh-install path from the review. The guard is real, and the `prepareOpenclawConfig`-level test asserting both `$include` and `IncludeRoot` covers the layer the unit test can't. `TestExpandOpenclawPathLeavesLookalikePrefixesAlone` also pins the `$OPENCLAW_HOMEX` boundary, so the "no path that works today changes" claim is now enforced by a test instead of by reading.

Local: `go vet` clean, `go test ./internal/daemon/execenv/ -run 'Openclaw|ByteIdentical' -count=1` passes, prompt-prefix invariant intact. CI green on `5bcab76d` including `windows-execenv`.

Two things worth calling out beyond the fix itself. Folding the upstream citations into the code comments was the right instinct — `openclawHomeRest` and `openclawHomeFromEnv` now explain *why* each shape is handled and which one is defensive, so the next person to touch this doesn't have to re-derive it from a field observation. And picking up both nits, including threading the offending path into the error, is what turns the remaining failure modes into something a user can act on.

Merging. For whoever lands #6275 second: `openclawConfigPathComplete` still needs to learn this shape — the comment there says so, so a rebase can't drop it silently.

Issue #3729, opened 2026-06-03
Pi runtime shows empty model list in UI when an agent config has a stale provider/model pattern


syxc:
The model picker for my Pi runtime in the Multica UI is just empty. The runtime shows online, I can dispatch tasks to Pi agents and they work (~4s response on `glm-coding-plan/glm-4.7`), but the picker shows no models to select from.

What I see in the daemon log:

```
20:25:55.748 DBG [pi:stderr] Warning: No models match pattern "opencode-go/mimo-v2-omni" component=daemon
20:25:55.749 DBG [pi:stderr] Warning: No models match pattern "opencode-go/mimo-v2-pro" component=daemon
```

The `mimo-v2-omni` and `mimo-v2-pro` entries are stale — I had them in one of my agent configs from an earlier setup, they don't exist in the local Pi provider catalog anymore. The warnings come from Pi CLI itself.

But what I see in the UI is the entire model list being empty, not "the two invalid ones are filtered out and the rest show up". I'd expect the daemon to ask Pi what models it can resolve, drop the ones that don't match, and return the resolved set. Returning empty because one entry doesn't match is throwing the baby out with the bathwater.

Happens on v0.3.15. Similar report for OpenCode was #3627 (turned out to be a 5s timeout on the model-discovery command), but this isn't a timeout — the log shows the Pi CLI responding with the pattern warnings directly, so the issue is on the daemon's side when it builds the model list to return to the UI.

Bohan-J:
Thanks for the detailed write-up — this helps a lot.

Quick background on how this works: the node/daemon discovers the model list by running `pi --list-models` locally and parsing its output. There's one version-dependent nuance we'd like to confirm against your machine. On a setup with Pi `0.67.2`, `pi --list-models` exits `0` and prints the model table to **stderr** (stdout is empty); the daemon falls back to reading stderr when stdout is empty, so the list loads fine there. That fallback only triggers when the command exits **0** — if `pi --list-models` exits **non-zero**, the daemon currently throws away whatever it printed and returns an empty list, which would explain exactly what you're seeing.

Could you run the following on the affected machine, as the **same OS user that runs the Multica node/daemon**, and paste the output?

1. `pi --version`

2. ```
   pi --list-models > /tmp/pi_out.txt 2> /tmp/pi_err.txt; echo "exit=$?"
   echo "--- stdout ---"; cat /tmp/pi_out.txt
   echo "--- stderr ---"; cat /tmp/pi_err.txt
   ```

[...]

PR #7312, opened 2026-08-20
MUL-6471: fix(daemon): let pi and opencode reach custom providers

### LINKED ISSUE #7300: [Bug]: pi and opencode runtimes cannot use custom providers — daemon builds --provider/--model incorrectly
### Deployment type

Self-hosted

## Summary

On a self-hosted setup where the pi and opencode runtimes are configured with **custom providers** (an OpenAI/Anthropic-compatible gateway), every Multica-dispatched task fails immediately, while the exact same work succeeds when the CLI is invoked by hand in the same container and workdir.

The daemon builds the `--provider` / `--model` arguments in a shape that does not match what these two CLIs accept when the provider is user-defined:

- **pi**: the daemon splits the model ID on `/` and passes the first segment as the provider name, discarding the configured provider.
- **opencode**: the daemon passes the bare model ID with no provider prefix, so opencode cannot resolve which provider owns it.

Claude Code and Codex runtimes on the same host are unaffected — only pi and opencode are.

The practical consequence is that **pi and opencode cannot be used at all with custom providers**, which also makes it impossible to run heterogeneous review lanes (different model families for implementation vs. review) that these two runtimes would otherwise enable.

## Environment

- Multica: self-hosted, four runtimes per container (`claude`, `codex`, `opencode`, `pi`), all `online`
- pi: `0.84.2`
- opencode: `1.18.18`
- claude: `2.1.233` — works
- codex: `codex-cli 0.147.0` — works
- Providers point at a local OpenAI/Anthropic-compatible gateway

Config files are mounted read-only and register a provider plus a model list (not a single hardcoded model).

`~/.pi/agent/models.json`:

```json
{
  "providers": {
    "multica-anthropic": {
      "baseUrl": "http://host.docker.internal:PORT",
      "api": "anthropic-messages",
      "models": [
        { "id": "claude/claude-opus-5",   "name": "Claude Opus 5" },
        { "id": "claude/claude-sonnet-5", "name": "Claude Sonnet 5" },
        { "id": "claude/claude-fable-5",  "name": "Claude Fable 5" }
      ]
    }
  }
}
```

`~/.config/opencode/opencode.json`:

```json
{
  "model": "multica-codex/codex/gpt-5.6-sol",
  "provider": {
    "multica-codex":     { "npm": "@ai-sdk/openai",    "models": { "codex/gpt-5.6-sol": {} } },
    "multica-anthropic": { "npm": "@ai-sdk/anthropic", "models": { "claude/claude-opus-5": {}, "claude/claude-sonnet-5": {}, "claude/claude-fable-5": {} } }
  }
}
```

Agent config in Multica: `model = claude/claude-opus-5`, `thinking_level = high`.

---

## Bug 1 — pi: model ID is split on `/`, configured provider is discarded

Daemon log:

```
INF picked task provider=pi
INF agent command exec=/usr/local/bin/multica-pi \
    args="[-p --mode json --session ... --provider claude --model claude-opus-5]"
DBG [pi:stderr] Error: Unknown provider "claude". Use --list-models to see available providers/models.
INF pi finished status=failed duration=554ms
```

The configured provider is `multica-anthropic` and the model ID is `claude/claude-opus-5`. The daemon appears to treat `claude/claude-opus-5` as `<provider>/<model>` and emits `--provider claude --model claude-opus-5`. Neither value exists in the config: `claude` is not a provider name, and `claude-opus-5` is not a model ID.

Reproduced by hand with the daemon's exact arguments:

```console
$ multica-pi -p --mode json --provider claude --model claude-opus-5 "say X"
Error: Unknown provider "claude". Use --list-models to see available providers/models.
```

Passing the configured provider and the full model ID works:

```console
$ multica-pi -p --provider multica-anthropic --model "claude/claude-opus-5" "say Y"
Y
```

Note that pi's `--model` already accepts a `provider/id` form, so the daemon does not need to split the string at all.

---

## Bug 2 — opencode: model is passed without the provider prefix

Daemon log:

```
INF picked task provider=opencode
INF agent command exec=/usr/local/bin/multica-opencode \
    args="[run --format json --dangerously-skip-permissions --dir <workdir> --model claude/claude-opus-5]"
WRN opencode error event error="Unexpected server error. Check server logs for details."
INF opencode finished status=failed duration=1.42s
```

opencode expects `--model` in `provider/model` form. The model ID here is itself `claude/claude-opus-5`, and the provider is `multica-anthropic`, so the fully-qualified value is `multica-anthropic/claude/claude-opus-5`. The daemon omits the provider segment.

Reproduced by hand with the daemon's exact arguments:

```console
$ multica-opencode run --format json --dangerously-skip-permissions --model claude/claude-opus-5 "say X"
{"type":"error","error":{"name":"UnknownError","data":{"message":"Unexpected server error. Check server logs for details.","ref":"err_b4263dc9"}}}
```

With the provider prefix:

```console
$ multica-opencode run --format json --dangerously-skip-permissions --model multica-anthropic/claude/claude-opus-5 "say Y"
{"type":"step_start",...}
{"type":"text","part":{"text":"Y",...}}
{"type":"step_finish","part":{"reason":"stop","tokens":{"total":10533,...}}}
```

Without `--model` it falls back to the configured default and also works, confirming the config itself is valid:

```console
$ multica-opencode run "say Z"
> build · codex/gpt-5.6-sol
Z
```

---

## Side observation — `model=""` at agent start, and `thinking_level` dropped

Both failures log an empty model at startup, with the real value only appearing at the `invoking backend` step:

```
INF starting agent provider=opencode workdir=... model="" reused=false
WRN thinking_level: not valid for this (provider, model); skipping injection provider=opencode model=claude/claude-opus-5 thinking_level=high
DBG invoking backend provider=opencode model=claude/claude-opus-5
```

The task row in the API also reports `"model": null` for these runs, so the selected model is not attributed to the run record.

Separately, `thinking_level=high` is dropped for both runtimes even though both CLIs support it (pi has `--thinking <off|minimal|low|medium|high|xhigh|max>`; the configured models declare `"reasoning": true`). Possibly the same root cause: if the provider/model pair cannot be resolved, the capability lookup fails and injection is skipped.

### Expected behavior

For runtimes backed by user-defined providers, the daemon should qualify the model with the provider from the runtime's own configuration rather than deriving it from the model string:

- **pi** — either `--provider <configured-provider> --model <full-model-id>`, or the single-argument form `--model <configured-provider>/<full-model-id>`. Do not split the model ID on `/` to guess a provider name.
- **opencode** — `--model <configured-provider>/<full-model-id>`.

Model IDs that themselves contain `/` are the normal case for gateway-style providers, so the delimiter is not a safe provider separator.

### Relevant log output

```
06:46:52.832 INF agent command component=daemon exec=/usr/local/bin/multica-opencode args="[run --format json --dangerously-skip-permissions --dir /workspaces/.../workdir --model claude/claude-opus-5]" prompt_bytes=838
06:46:54.227 WRN opencode error event component=daemon error="Unexpected server error. Check server logs for details."
06:46:54.252 INF opencode finished component=daemon pid=25528 status=failed duration=1.42s
06:46:55.759 INF task did not complete, reporting failure component=daemon task=d945b69e status=blocked failure_reason=agent_error.unknown

06:53:49.478 INF agent command component=daemon exec=/usr/local/bin/multica-pi args="[-p --mode json --session /home/agent/.multica/pi-sessions/....jsonl --provider claude --model claude-opus-5]"
06:53:50.027 DBG [pi:stderr] Error: Unknown provider "claude". Use --list-models to see available providers/models. component=daemon
06:53:50.032 INF pi finished component=daemon pid=25739 status=failed duration=554ms
06:53:51.066 INF task did not complete, reporting failure component=daemon task=b790937b status=blocked failure_reason=agent_error.process_failure
```

Bohan-J:
Fixes #7300.

## The root cause

Gateway-style providers register model ids that themselves contain a slash — `claude/claude-opus-5` under provider `multica-anthropic`. Both reported failures came from treating that slash as a provider boundary, or from having no provider at all.

## Bug 1 — pi

`buildPiArgs` split the model id on the first slash and passed the leading segment as `--provider`, turning `claude/claude-opus-5` into `--provider claude --model claude-opus-5`. pi answers an unknown `--provider` with a hard error before it ever looks at `--model`.

The selector now goes to `--model` whole and `--provider` is never synthesized. Reading pi's `resolveCliModel` shows why this is strictly better: with no `--provider`, pi infers the provider from the prefix when the prefix names a real provider, and otherwise matches the full string as a raw model id — so `provider/id`, a bare id, and a slash-shaped id all resolve. Passing `--provider` only adds a failure mode.

Verified against pi 0.67.2 with the reporter's `models.json`:

```console
$ pi -p --mode json --provider claude --model claude-opus-5 "say X"     # old argv
Error: Unknown provider "claude". Use --list-models to see available providers/models.

$ pi -p --mode json --model "claude/claude-opus-5" "say X"              # new argv
{"type":"message_start","message":{"role":"assistant","provider":"multica-anthropic","model":"claude/claude-opus-5",...}}
```

## Bug 2 — opencode

`--model` is `provider/model`, and an id persisted without its provider is rejected outright. Reproduced against opencode 1.18.14 with the reporter's `opencode.json`:

```console
$ opencode run --format json --model "claude/claude-opus-5" "say X"
{"type":"error","error":{"name":"UnknownError","data":{"message":"Unexpected server error. Check server logs for details."}}}
```

The daemon cannot invent the missing segment, so it asks the runtime's own catalog instead of the model string. `agent.QualifyModelID` promotes an id that exactly one provider claims to the catalog's canonical selector, and passes everything uncertain through untouched — a static fallback catalog, an already-canonical id, or two providers exposing the same bare id. Manual model entry is supported everywhere, so a value the catalog cannot confidently place still reaches the CLI verbatim.

Against a real `opencode models` / `pi --list-models` catalog built from the reporter's config, both runtimes resolve `claude/claude-opus-5` → `multica-anthropic/claude/claude-opus-5`, which is the form the report confirms works by hand.

## The two side observations

- **`thinking_level` dropped** — same root cause. The capability lookup is keyed on the model id and missed every catalog entry while the id was unqualified; it matches now. (For a provider whose opencode model config declares no `variants`, the drop stays correct — opencode's `--variant` needs a variant the model advertises.)
- **`starting agent` logging `model=""`** — a real logging bug: it printed `entry.Model`, the `MULTICA_<PROVIDER>_MODEL` env tier alone, so every task whose model came from `agent.model` looked like the selection had been lost. Resolution now happens before the log.
- **`"model": null` on the task row** is *not* part of this. That value is read from the run's usage rows, which a run that fails before its first provider call never produces. It should resolve once the run gets that far.

## Cost note

Qualification is one catalog read per task that pins a model, memoized for 60s, on a path that already takes one for `thinking_level` / `service_tier` — the two checks below it reuse the lookup. It goes through the daemon's existing `listModels` indirection so default tests never shell out to a real CLI. Runtimes whose ids are not provider-namespaced can never be rewritten, but they are deliberately not special-cased away: an allowlist of "runtimes that namespace" is a list that rots the day the next gateway-style runtime lands.

## Verification

- `go build ./...`, `go vet`, `gofmt` clean.
- `go test ./pkg/agent/ -count=1` — ok.
- `go test ./internal/daemon/... -count=1` — the same 7 failures appear on an unmodified checkout in this sandbox (HOME/profile-dir layout tests); no new ones.
- New tests: pi argv regressions for slash-shaped, qualified, bare, and empty models; `QualifyModelID` matrix covering promotion, already-canonical, exact-match-wins, cross-provider ambiguity, unknown ids, and fallback catalogs; and the pi RPC → qualify → thinking-catalog chain that pins the `thinking_level` recovery.
- End-to-end against locally installed pi 0.67.2 and opencode 1.18.14 using the reporter's exact config shape, as quoted above.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -6140,6 +6140,121 @@ func skillRefFromBundle(bundle SkillData) SkillRefData {
 	}
 }
 
+// taskModelSelection is what a task actually launches with: the model
+// selector plus the capability overrides that survived validation.
+type taskModelSelection struct {
+	Model         string
+	ThinkingLevel string
+	ServiceTier   string
+}
+
+// resolveTaskModelSelection settles the model selector and its capability
+// overrides against the runtime's own model catalog, reading that catalog at
+// most once per task — and not at all when nothing needs it.
+//
+// The single read is the point. Discovery is a CLI subprocess with a 15-30s
+// ceiling, and cachedDiscovery deliberately does not memoize a result that
+// came back empty or as a fallback (#3729, MUL-5549) so a transient failure
+// can retry immediately. A logged-out or timing-out runtime therefore pays
+// that ceiling in full on every read, and a task that read the catalog once to
+// qualify its model and again to validate thinking_level would pay it twice
+// before the agent even starts (MUL-6471 review).
+//
+// Who asks for the catalog:
+//   - opencode and its DevEco fork cannot execute an unqualified selector, so
+//     a pinned model has to be resolved against the catalog before launch.
+//   - thinking_level / service_tier are catalog-owned and keyed on the
+//     catalog's own model id, so they need it whenever they are set.
+//
+// A pi task with no capability override asks for neither: pi's own resolver
+// accepts the persisted id in every shape it can take, so the daemon has
+// nothing to add and skips discovery entirely. Same for claude, codex, and any
+// task that pins no model.
+//
+// Qualification runs first because both checks below match on the catalog's
+// canonical id: an unqualified id silently fails every lookup and drops a
+// perfectly valid level (GH #7300).
+func resolveTaskModelSelection(
+	ctx context.Context,
+	provider string,
+	runtimeCmd agent.Command,
+	sel taskModelSelection,
+	taskLog *slog.Logger,
+) taskModelSelection {
+	capabilityChecksPending := sel.ThinkingLevel != "" || sel.ServiceTier != ""
+
+	read := false
+	var (
+		catalog    agent.Catalog
+		catalogErr error
+	)
+	loadCatalog := func() (agent.Catalog, error) {
+		if !read {
+			read = true
+			catalog, catalogErr = listModels(ctx, provider, runtimeCmd)
+		}
+		return catalog, catalogErr
+	}
+
+	sel.Model = qualifyTaskModel(provider, sel.Model, capabilityChecksPending, loadCatalog, taskLog)
+
+	// service_tier is catalog-owned and currently Codex-only. As with
+	// thinking_level, stale or incompatible persisted values degrade to the
+	// runtime default instead of failing the task. Catalog lookup errors pass
+	// through so a transient discovery failure does not silently disable a
+	// previously valid user choice.
+	if sel.ServiceTier != "" {
+		ok, err := agent.ValidateServiceTierWith(loadCatalog, provider, sel.Model, sel.ServiceTier)
+		if err != nil {
+			taskLog.Warn("service_tier: catalog lookup failed; passing through",
+				"provider", provider,
+				"model", sel.Model,
+				"service_tier", sel.ServiceTier,
+				"error", err,
+			)
+		} else if !ok {
+			taskLog.Warn("service_tier: not valid for this (provider, model); skipping injection",
+				"provider", provider,
+				"model", sel.Model,
+				"service_tier", sel.ServiceTier,
+			)
+			sel.ServiceTier = ""
+		}
+	}
+	// Per-model guard: the server validates the literal token against the
+	// provider's enum, but per-model gaps (Claude's `xhigh` on a non-Opus
+	// model, Codex's per-model `supported_reasoning_levels`) only resolve
+	// here, against the daemon's local CLI catalog. Invalid combinations
+	// log a warning and drop the level rather than failing the task, so a
+	// stale persisted value never blocks execution. An empty model is
+	// resolved by ValidateThinkingLevelWith to the provider's default model so
+	// default-model tasks aren't misjudged — except for codex, whose empty
+	// model follows config.toml (any model) and so fails closed, dropping the
+	// level here without a catalog read at all. Discovery errors fail open for
+	// resolved models: if we can't list models, we keep the persisted level
+	// and let the CLI object.
+	if sel.ThinkingLevel != "" {
+		ok, err := agent.ValidateThinkingLevelWith(loadCatalog, provider, sel.Model, sel.ThinkingLevel)
+		if err != nil {
+			taskLog.Warn("thinking_level: catalog lookup failed; passing through",
+				"provider", provider,
+				"model", sel.Model,
+				"thinking_level", sel.ThinkingLevel,
+				"error", err,
+			)
+		} else if !ok {
+			taskLog.Warn("thinking_level: not valid for this (provider, model); skipping injection",
+				"provider", provider,
+				"model", sel.Model,
+				"thinking_level", sel.ThinkingLevel,
+			)
+			sel.ThinkingLevel = ""
+		}
+	}
+
+	return sel
+}
+
 // qualifyTaskModel promotes a persisted model id to the canonical
 // `<provider>/<id>` selector its runtime catalog advertises, and returns the
 // model to actually launch with.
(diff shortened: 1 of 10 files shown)


Bohan-J:
Pushed `bbaf356b8` for the must-fix.

You were right, and the specific reason matters more than I gave it credit for: `cachedDiscovery` deliberately skips the cache for empty and fallback results (#3729, MUL-5549) so a transient failure can retry. My "the path already takes a memoized lookup" argument only covered the success case — in exactly the failure cases that hurt (logged out, discovery timing out), there is no memoization at all, so every task would have paid the full ceiling again.

**Qualification is now an explicit runtime capability.** `ModelIDsAreProviderQualified` is true only where `Model.Provider` is literally the ID's own prefix — pi builds `provider + "/" + id`, opencode reads the provider back off the id. Runtimes carrying a bare id next to a display-name Provider never reach discovery: claude's `"anthropic"`, copilot's `"openai"`, and dsh's `"DeepSeek"` against a `deepseek-official/...` id could never have matched the rewrite rule anyway. Built-in runtime identities resolve through `ProtocolFamily`, so a pi-family fork like omp inherits the capability from its descriptor entry rather than a name check — which keeps the "adding a fork is a descriptor entry, not a cross-stack change" property that `builtin_runtimes.go` documents.

**Daemon-level tests, both halves.** The call site moved into `qualifyTaskModel` so the boundary is testable through the existing `listModels` indirection:

- `TestQualifyTaskModelSelectorsAndDiscoveryBoundary` — pi, opencode and omp resolve to the selector their CLI accepts; claude, codex and an empty model never reach `listModels` at all. Discovery calls are counted, so "did not call" is asserted rather than assumed.
- `TestQualifyTaskModelFailsOpenOnDiscoveryError` — a runtime that cannot answer does not block the task.

**On the allowlist rotting**, which was my original objection to doing this: `TestModelIDsAreProviderQualifiedMatchesCatalogShape` derives the expected answer by running the real catalog producers (`parsePiModels`, `parseOpenCodeModels`, `parseOmpModels`, and the static claude/codex/copilot/cursor catalogs) and checking whether their output is provider-prefixed, instead of restating the list. I confirmed it bites by temporarily adding `claude` to the predicate — the claude case fails with the emitted catalog in the message.

Verification: `go build ./...`, `go vet ./pkg/agent ./internal/daemon/...`, `gofmt` clean; `go test ./pkg/agent/ -count=1` ok; the new daemon tests pass. `go test ./internal/daemon/... -count=1` still hits the same 7 pre-existing HOME/profile-path failures this sandbox has on an unmodified checkout — unchanged by this commit.

Bohan-J:
Pushed `5423357b5`. Both must-fixes addressed, plus the test-contract point.

**1 — double discovery in one task.** Confirmed and fixed. `qualifyTaskModel` read the catalog to promote the selector, then `ValidateThinkingLevel` read it again — back to back, on exactly the reporter's configuration. The two now share one lazily-loaded, task-scoped catalog: `resolveTaskModelSelection` owns it and hands the loader to `ValidateThinkingLevelWith` / `ValidateServiceTierWith` instead of letting them reach for `ListModels` themselves. The reuse is task-scoped, so the global immediate-retry contract in `cachedDiscovery` is untouched.

Routing it through a loader rather than a pre-fetched catalog keeps every pre-discovery guard in one place — `value == ""` and codex's empty-model fail-closed still settle without a read, which a catalog-taking signature would have quietly cost.

**2 — pi should not qualify at all.** Agreed; you're right that my own `buildPiArgs` evidence made this case. Qualification is not a launch requirement for pi, so a pi task with no capability override now spawns no discovery subprocess. `ModelIDsAreProviderQualified` is replaced by `ModelSelectorMustBeProviderQualified` — the CLI *refuses* an unprefixed id — true only for opencode and its DevEco fork.

**3 — the test contract.** That predicate rename also retires the catalog-shape invariant, which as you noted deveco was already a counterexample to. The replacement tests the execution contract directly (including `pi`/`omp` as explicit false cases with the reason), plus a separate test that `omp` inherits from its `ProtocolFamily` rather than a name check.

The behaviour matrix is pinned by `TestResolveTaskModelSelectionReadsTheCatalogAtMostOnce`, which asserts the read *count*, not just the outcome:

| case | result | catalog reads |
| --- | --- | --- |
| pi / opencode + `thinking_level` | qualified, level survives | 1 |
| pi, omp — no override | untouched | 0 |
| opencode — no override | qualified | 1 (cannot launch otherwise) |
| claude — no override | untouched | 0 |
| claude + `thinking_level` | validated | 1 |
| codex + level + tier | both validated | 1 |
| codex, no model | both fail closed | 0 |
| no pinned model | untouched | 0 |

Plus `TestResolveTaskModelSelectionFailsOpenOnDiscoveryError`: a failed read is not retried inside the task, and neither the model nor the override is dropped.

I checked the count assertion actually bites by removing the memoization — the pi and opencode cases report `catalog reads = 2, want 1`, which is precisely the path you described.

Verification: `go build ./...`, `go vet ./pkg/agent ./internal/daemon/...`, `gofmt` clean; `go test ./pkg/agent/ -count=1` ok; new daemon tests pass. `go test ./internal/daemon/... -count=1` still shows only this sandbox's pre-existing HOME/profile-path failures.

PR #7316, opened 2026-08-20
MUL-6686 feat(daemon): use readable workdir paths

### LINKED ISSUE #6063: [Feature]: Make Multica work attributable to time trackers — opaque workdir paths defeat project detection (WakaTime)
## The problem

Time spent working through Multica is invisible to time-tracking tools, and the agent workdir layout actively prevents recovering it.

Two separate symptoms, one cause.

**1. The desktop app reports as a single project.** WakaTime's macOS desktop tracker attributes by application, so every hour spent in Multica — across every workspace, every project, every issue — lands in one bucket called "Multica". For anyone using Multica as their primary work surface rather than an occasional dashboard, that's the majority of the working day collapsed into a single undifferentiated row.

**2. Agent work is untrackable by path.** Path-based trackers (WakaTime's `[projectmap]`, RescueTime, anything reading the git repo from the file path) can normally recover attribution from the directory. Multica's workdir layout defeats this:

```
~/multica_workspaces_<host>/<workspace-uuid>/<task-hash>/workdir/<repo>/
```

The workspace is a **bare UUID** and the task segment is an **opaque hash**. Neither is a label a human or a tracker can use. Here is the projectmap I ended up writing to salvage anything at all:

```ini
[projectmap]
multica_workspaces_[^/]+/[^/]+/[^/]+/workdir/af-laravel/ = assetfeed-laravel
multica_workspaces_[^/]+/[^/]+/[^/]+/workdir/([^/]+)/ = {0}
multica_workspaces_ = multica-agents
```

Note what those regexes have to do: skip past both the workspace UUID and the task hash with `[^/]+`, because there is nothing usable in them. The best I can recover is the **repo name**, and only when the agent happens to be working inside a checked-out repo. Everything else falls through to a catch-all bucket. Which workspace, which project, which issue, which agent — all unrecoverable.

## What I'd propose

Three options, roughly in order of effort. The first is worth doing regardless of whether the others ever happen.

### 1. Human-meaningful workdir paths (small, no vendor coupling)

Put the workspace slug and the issue key in the path instead of raw identifiers:

```
~/multica_workspaces_<host>/<workspace-slug>/<ISSUE-KEY>/workdir/<repo>/
```

This is the highest leverage change in the whole issue. It fixes attribution for *every* path-based tool at once — WakaTime, RescueTime, anything else — with no integration to build and nothing vendor-specific in the codebase. It also makes the directory tree navigable by a human, which matters the first time you go looking for what an agent actually did on disk.

If UUIDs are load-bearing for collision-safety, a readable prefix with a short hash suffix (`assetfeed-MUL-1234-a3f9`) keeps both properties.

One caveat worth raising in review: Claude Code keys its session store by cwd (`~/.claude/projects/<encoded-cwd>/`), so changing the path scheme affects session resume for existing workdirs. Probably wants to apply to newly created workdirs only, with old ones left alone.

### 2. Optional WakaTime heartbeats from the daemon (opt-in)

The daemon has better data than any editor plugin could ever infer. It knows exactly which agent worked on which issue, in which workspace, starting when and ending when — no idle-detection heuristics, no guessing from keystrokes.

Gated behind a `WAKATIME_API_KEY` env var (absent → feature entirely off), the daemon could emit heartbeats during agent runs:

| WakaTime field | Multica value |
|---|---|
| `project` | workspace name, or Multica project (see below) |
| `entity` | issue key + title |
| `branch` | agent name |
| `language` | inferred from files the agent touched |
| `category` | `coding` / `code reviewing` / `researching` |

That turns "8 hours in Multica" into a real breakdown of where agent time actually went — which, for a product whose value proposition is delegating work to agents, is exactly the number a user wants to see.

### 3. Generic outbound webhooks instead

If a WakaTime-specific integration is too narrow to carry in-tree — a fair position — the generic version already has an issue: **#1964 (RFC: outbound webhooks on issue/run state transitions)**. Run-start and run-end events with workspace, project, issue, agent and duration would let users pipe into WakaTime, Toggl, Harvest, or their own store, with zero vendor coupling in the codebase.

I'd be happy with this outcome. It's arguably the more Multica-shaped answer.

## On "project = workspace?" — I don't think it is

My first instinct was workspace → WakaTime project, but I think that's wrong on reflection, for two reasons.

A workspace is an *organisational* boundary, not a unit of work. A single workspace can hold entirely unrelated efforts — in ours, backend work, SEO, legal research and infrastructure all live side by side. Bucketing them together produces the same undifferentiated blob as today, just at a smaller scale.

More importantly, **Multica already has a first-class `project` entity**, and it maps onto the WakaTime concept almost exactly: a coherent body of work, spanning multiple issues, optionally bound to specific repos. Using it means the tracking hierarchy mirrors the product's own model instead of inventing a parallel one.

So I'd suggest:

- **default:** Multica project name, falling back to workspace name for issues with no project
- **override:** a per-workspace setting for teams whose mental model really is one-workspace-one-project
- **repo-scoped work:** when an agent is working inside a checked-out repo, the repo name is arguably the better project label — worth deciding explicitly rather than leaving to path heuristics, since that's the case where today's accidental behaviour (via projectmap) is actually closest to correct

## Priority

I'd rank (1) well above the rest. It's a small change, it's useful independent of time tracking, and it unblocks users to solve their own attribution with tools they already run — which is the cheapest possible resolution for the maintainers.

Related: #1964 (outbound webhooks RFC).

iiwish:
## Summary

- create new task environments under sanitized, bounded readable segments: `<workspace-slug>-<workspace-short>/<issue-key>-<task-short>/workdir`
- carry workspace slugs and issue identifiers through the claim payload without using titles or user text in paths
- use `.gc_meta.json` identity for GC batching and disk-usage aggregation, including task IDs for all newly completed tasks
- preserve legacy workdirs and managed follow-up reuse, and render privacy-safe relative paths for both layouts

Existing workdirs are not renamed or migrated. Labels are lowercased ASCII, separator/traversal input is collapsed, segments are capped at 48 characters, and stable ID suffixes retain collision safety across renames and case-insensitive filesystems.

## Tests

- `go test -race ./internal/daemon/execenv -run TestPredictRootDir -count=1`
- `go test -race ./internal/daemon -run TestRunGC_MixedLayoutsUseMetadataWorkspaceIdentity -count=1`
- `go test -race ./internal/handler -run TestRelativeWorkDir -count=1`
- `go vet ./internal/daemon/execenv ./internal/daemon ./internal/handler`

The race-test invocations used regexes to include the related sanitization, mixed-layout disk usage, GC metadata, managed reuse, relative path, and claim-payload cases.

Closes #6063


RIDCorix:
Yes, please! And we have to prevent duplicating workdir names, too. I encountered two full task IDs sharing the same first UUIDv7 segment were assigned one workdir.

Bohan-J:
Thanks for this — readable workdir paths are a real improvement for debugging and disk management, and we'd like to land it. Carrying `TaskID` in `.gc_meta.json` is arguably the most valuable part: it removes the implicit assumption that a directory *name* is the only source of task identity.

Heads-up that the ground moved underneath this PR, through no fault of yours. #7347 landed on `main` (`3622db21`) about two hours ago, fixing #7326, and it rewrites the same path derivation this PR touches. GitHub now reports the branch as conflicting. I merged `main` into your branch locally to see how bad it is — four real conflicts:

```
server/internal/daemon/execenv/execenv.go
server/internal/daemon/execenv/execenv_test.go
server/internal/handler/agent.go
server/internal/handler/agent_work_dir_test.go
```

Three things need attention on the way through, and one of them is a correctness issue rather than a merge mechanic.

### 1. The task segment must stop using `shortID`

This is the important one. `readablePathSegment` currently does:

```go
suffix := strings.ToLower(shortID(id))
```

`shortID` returns the **first** 8 hex chars of the id. Task ids are UUIDv7, which lead with a 48-bit millisecond timestamp, so those 8 chars are the high 32 bits of that timestamp — they only advance once every `2^16` ms (**~65.5 seconds**). Every task created inside one such window gets an identical suffix. That is what #7326 was: not a rare hash collision, but the common case for any burst dispatch.

The `<issue-key>` segment you added hides most of it, because tasks on *different* issues no longer share a directory — and that happens to be exactly the reproduction in #7326, so it can look fixed. But two tasks on the **same** issue inside one 65.5s window still collide: reruns, follow-ups, and several agents working one issue all hit it. This PR also deliberately preserves legacy roots, which keep the old rule.

`main` now has `taskKey` (in `execenv/git.go`), which takes the **last** 12 hex chars — the random tail — for exactly this reason. After rebasing, `readablePathSegment` should call `taskKey` instead of `shortID`. `shortID` still exists but is now logs/display only, and is commented as such.

### 2. Please carry the `.task_owner` claim protocol through the rebase intact

The conflict in `execenv.go` is not cosmetic. `PredictRootDir` is now the entry point to a claim protocol: `Prepare` proves it owns an env root before it wipes it, because that `os.RemoveAll` was deleting live sibling tasks' worktrees and task-scoped config. When you resolve that conflict, the claim needs to land on top of the new readable layout rather than being replaced by the old straight-line version.

One more heads-up so you're not rebasing onto a moving target: #7356 is open and revises that same claim (it adds an execution lock and removes the `Mkdir`/`O_EXCL` arbitration). It isn't merged yet, so `main` is the right base today — just expect one more touch in that function.

### 3. The Windows path budget needs re-validating, and the cap is probably too large

`readablePathSegmentMax = 48` is applied to both segments, so an env root can reach `48 + 1 + 48 = 97` characters. Today's layout is `36 + 1 + 12 = 49`. That is **+48 characters on the root**, and everything the agent creates lives underneath it.

That budget is tighter than it looks, because git refs land *inside* the task checkout — `.git/refs/heads/agent/<name>/<task>` is under the env root. While preparing #7347 we overflowed `MAX_PATH` in the `windows-execenv` job and got:

```
fatal: cannot lock ref 'refs/heads/agent/windows-codex/<32 chars>':
Unable to create '....lock': Filename too long
```

That job's isolated-checkout case now runs at 241 characters — 19 to spare against the 260 limit. Adding 48 puts it around 289. Suggested: lower the cap substantially (~24 per segment) and run `windows-execenv` once with a worst-case workspace slug *and* issue key rather than short fixtures, since the default fixtures won't reach the cap.

---

None of this changes the verdict on the idea — it's worth having, and the `.gc_meta.json` change should land regardless. If you'd rather not deal with the rebase now that it collides with a fix that already shipped, say so and we can take it over while keeping your commits and authorship.

Also worth noting: @RIDCorix independently hit the same collision and reported it on this PR, so the concern in (1) isn't hypothetical.

iiwish:
@RIDCorix Thanks for confirming the UUIDv7 collision in practice. I changed the readable task segment to use `taskKey` (the last 12 random hex characters) instead of `shortID` (the shared timestamp prefix).

I also added regressions for multiple tasks on the same issue with the same UUIDv7 prefix, plus a forced same-tail collision test: the normal roots are distinct, and the improbable true key collision fails closed through `.task_owner` without deleting the winning task's files. The focused race suite and the full CI matrix are green.

iiwish:
@Bohan-J Thank you for the detailed review. I merged current `main` and addressed all three points:

1. The task segment now uses `taskKey` (the last 12 random UUIDv7 hex characters), with same-issue burst-task and forced-collision regressions.
2. The `.task_owner` protocol from #7347 is preserved intact on top of the readable layout. The atomic claim, foreign-owner refusal, same-owner reset, and concurrent same-key behavior are covered under `-race`.
3. Each readable segment is capped at 24 characters, so the two segments consume at most 49 characters in total, matching the current opaque `<workspace UUID>/<12-char task key>` budget rather than adding to it. The real `windows-latest` test now uses worst-case workspace and issue labels for the production env-root layout, creates the isolated Git checkout and nested `agent/.../<taskKey>` ref, then stages and commits successfully. `windows-execenv` is green.

I also kept `TaskID` in `.gc_meta.json`, updated GC/disk reporting to read logical identity from metadata across mixed old/new layouts, and retained privacy-safe display compatibility for legacy opaque and earlier readable paths. Locally, the full backend suite, `go vet ./...`, affected-package race suites, repeated collision/reuse cases, and Windows cross-compilation all pass; the complete GitHub CI matrix is green and the PR is `CLEAN`/mergeable.

#7356 is still open, so this remains based on the current claim protocol from `main`; I can sync again if that lands before this PR.

multica-eve (dismissed):
This change would make new task folders show recognizable workspace and issue labels, making them much easier to navigate and attribute while leaving existing folders untouched. Product verdict: this is a worthwhile user experience improvement, and we should accept the direction.

Implementation verdict: **request changes**. I found two correctness issues at the current head (`ca3ac4270`) that need to be fixed before merge.

1. **The physical root for one task is not stable over that task's lifetime.**

   `PredictRootDir` (`server/internal/daemon/execenv/execenv.go:337-345`) includes `WorkspaceSlug` and `IssueIdentifier` in the actual path. Those values are opportunistically enriched during claim handling (`server/internal/handler/daemon.go:2992-3000`): a lookup failure leaves them empty, and the workspace issue prefix can change later. Therefore the exact same `WorkspaceID + TaskID` can resolve first to:

   ```text
   workspace-a548b2390cb2/task-0123456789ab
   ```

   and later to:

   ```text
   asset-feed-a548b2390cb2/mul-6063-0123456789ab
   ```

   I reproduced this deterministically on the current head. A stale re-dispatch can happen before `StartTask` persists the first workdir. If the readable fields differ on that second claim, it lands in a different root and bypasses the `.task_owner`, execution lock, and active-root protection on the first root. That permits duplicate preparation/execution and orphans the live or partial directory. It also violates #6063's accepted requirement that renaming must not orphan a live workdir.

   Please either persist and freeze the naming inputs before the first dispatch/preparation, or introduce an authoritative identity/claim layer derived only from stable IDs that also prevents the original root from being orphaned. Add a regression proving that the same task retains one physical identity when readable fields are absent then present, when the issue prefix changes, and when the task is reclaimed before its workdir is recorded. The existing `PriorWorkDir` test covers reuse by a later task, not this same-task pre-`StartTask` reclaim window.

2. **Active and preparation-crashed readable roots are attributed to the wrong workspace in disk-usage reports.**

   `.gc_meta.json` is written only after task completion (`server/internal/daemon/daemon.go:5216`). Before that, `buildTaskUsage` (`server/internal/daemon/diskusage.go:307-324`) treats the physical parent directory name as `workspace_id`. I reproduced an active readable root reporting:

   ```text
   workspace_id = "asset-feed-a548b2390cb2"
   ```

   instead of the authoritative workspace UUID. This creates a separate, incorrect workspace aggregate precisely while a task is active or after preparation crashes. The current mixed-layout test only covers completed roots that already have GC metadata.

   Please persist non-completion identity metadata when the root is prepared, or read an existing authoritative provenance/ownership marker, and add a disk-usage regression for a readable active root without `.gc_meta.json`.

The collision-safe random-tail key, exclusive ownership protection, ASCII sanitization, 24-character cap, legacy-layout handling, and real Windows path coverage all look good and should be preserved. I also verified the prompt byte-identity contract, targeted daemon/execenv race tests, handler non-database tests, `go vet`, and `git diff --check`; GitHub CI is green, including `windows-execenv` and backend tests.

iiwish:
@multica-eve Addressed both requested correctness fixes in `ad68cdc4c` and merged the latest `upstream/main` so the PR is no longer dirty.

- A stable ID-keyed root record now freezes each `(workspace_id, task_id)` to one physical readable path, adopts compatible pre-index roots, serializes concurrent first claims, and fails closed on corrupt or mismatched records. Terminal GC removes the record.
- `.task_owner` now records authoritative workspace/task identity (while remaining backward-compatible with legacy task-only markers), and disk-usage reporting reads authoritative metadata before `.gc_meta.json` exists.
- Added regressions for missing then enriched labels, prefix/slug changes, reclaim before workdir persistence, concurrent claims, legacy adoption, corrupt records, active-root disk attribution, and GC cleanup.

Verification after merging current `main`: full `pnpm typecheck` and `pnpm test`, `go test -race ./internal/daemon/...`, full `go vet ./...`, and Windows cross-compilation for execenv/daemon/handler/service all pass. I also ran the full Go race suite against PostgreSQL before the final upstream merge; the merge-overlap daemon package was rerun under the race detector afterward. Please re-review when convenient.

iiwish:
@multica-eve @Bohan-J Synced this PR with current `main` at `9ecc6787b` and resolved the new overlap with #7356 and #7405.

The readable, stable ID-keyed root resolution remains authoritative, while the parent-held OS execution lock from #7356 is preserved through `EnvRootClaim` / `EnvRootPreclaimed`. The handler also retains the fail-closed project-context loading from #7405 and now hydrates the issue identifier from that authoritative load.

I reran the full daemon subtree, focused handler regressions, the ownership/concurrency suite under `-race`, `go vet`, and `git diff --check`. The refreshed GitHub matrix is fully green, including `backend-tests` and `windows-execenv`, and the PR is `CLEAN` / mergeable. Please re-review the current head `c13eb719c` when convenient.

iiwish:
@multica-eve Synced the latest `main` and completed another correctness review at head `4adad709a`.

The two requested fixes remain intact: stable ID-keyed root resolution prevents readable-label drift from changing a task root, and active roots use authoritative workspace/task metadata for disk attribution. I also fixed one additional crash-consistency gap found during review: `.task_owner` is now written through a synced same-directory temp file plus atomic rename, so lock-free disk-usage readers cannot observe truncated JSON. Claims remove unpublished temp files left by a process that exits before rename, avoiding a new permanently wedged root.

Added deterministic regressions for atomic marker replacement and interrupted-write recovery. Local daemon/execenv race suites, `go vet ./...`, targeted handler tests, repeated recovery tests, and Windows cross-compilation pass. The refreshed GitHub CI matrix is fully green, including `backend-tests` and `windows-execenv`; the PR is `CLEAN` and mergeable. Please re-review when convenient.

Bohan-J (changes requested):
Re-reviewed at `4adad709a`. Both correctness issues from the last round are genuinely fixed. One small regression is blocking.

## Verified fixed

**1. One task keeps one physical root for its whole lifetime.**

The `.task_roots/<sha256(workspace_id\0task_id)[:16]>/root.json` index is keyed only on stable IDs and freezes the first resolved path. `ResolveRootDir` always re-reads the authoritative record rather than returning its own proposal, so a claim arriving with different readable labels still lands on the frozen root. Concurrent first claims are arbitrated by `os.Rename` of a complete non-empty directory failing onto an existing one, with the loser reading the winner's record. `findOwnedTaskRoot` adopts roots that predate the index, so legacy roots carried over from `main` aren't orphaned.

The regressions cover the cases I asked for: labels absent then present, workspace/issue prefix rename, reclaim before `StartTask` persists the workdir, concurrent claims, and corrupt records.

**2. Disk-usage attribution for active and preparation-crashed roots.**

`.task_owner` now carries authoritative `workspace_id` — backward compatible with the legacy task-only marker, and upgraded in place while the lock is held. `buildTaskUsage` reads provenance → owner → gc_meta, so a root with no `.gc_meta.json` yet still aggregates under the real workspace UUID instead of the physical parent directory name.

The atomic `.task_owner` write you added on your own initiative (same-directory temp + fsync + rename, unpublished temps cleared at claim time) is a good catch — lock-free readers really could observe a torn marker.

## Blocking: the `disk-usage` PATH column regresses to the UUIDv7 prefix

`server/internal/daemon/diskusage.go:329` and `:339` overwrite `usage.TaskShort` with `ShortID(taskID)` — the **leading** 8 hex chars of a UUIDv7, which is exactly the shared-timestamp segment that `taskKey`'s own comment documents as identical for every task created inside one ~65.5s window. `server/cmd/multica/cmd_daemon.go:2070` renders that value in a column headed **PATH**.

Measured locally, two tasks on one issue inside one such window:

```
this PR:
  PATH column = "01a039fd/01a039fd"      (real dir = "mul-6063-aaaaaaaaaaaa")
  PATH column = "01a039fd/01a039fd"      (real dir = "mul-6063-bbbbbbbbbbbb")

main (bdcb231ba):
  PATH column = "01a039fd/aaaaaaaaaaaa"  (real dir = "aaaaaaaaaaaa")
  PATH column = "01a039fd/bbbbbbbbbbbb"  (real dir = "bbbbbbbbbbbb")
```

The two rows become indistinguishable, and the value no longer corresponds to the readable directory this PR just created — on the surface most used for disk triage, which is the PR's own goal. Previously `TaskShort` was the physical directory name (`t.Name()`): unique, and something you could `cd` to.

`TestScanDiskUsage_ReadableActiveRootUsesOwnerIdentityWithoutGCMeta` asserts `usage.TaskShort == ShortID(taskID)`, so the behavior is pinned by a test and needs to change along with it.

Suggested fix: render the physical segment (`filepath.Base(usage.Path)`), or fall back to the 12-char `taskKey` tail. If the authoritative task id should be exposed, add a separate field rather than taking over the PATH column.

## Nits (non-blocking)

1. **`.task_roots` has no backstop reclaim.** The only removal path is `RemoveRootDirRecord` from `cleanTaskDir` (`gc.go:807`), and it requires `.task_owner` to carry both ids. A record created in `handleTask` whose env root never materialises (task fails before `ClaimEnvRoot`), or a root whose owner marker was lost, leaves a permanent entry. The GC walk and `ScanDiskUsage` both skip dot-directories, so it is neither cleaned nor visible in the disk report. An age-gated sweep would cover it — it can't be unconditional, or it would break the freeze window between `ResolveRootDir` and `Prepare`.

2. **`root_identity.go:188`'s `os.Remove(.task_roots)` is pure downside.** It races `installTaskRootRecord`, which does `MkdirAll(parent)` and then `MkdirTemp(parent, ...)`: if a GC removes the empty index between those two calls, `MkdirTemp` gets ENOENT and the task fails. Dropping the line costs one empty directory.

3. **`isEnvRootBookkeeping` (`execenv.go:1640`) doesn't know about `.task_owner-*.tmp`.** `claimEnvRoot` is fine because it calls `removeStaleEnvRootOwnerTemps` first, but `findOwnedTaskRoot` (`root_identity.go:232`) calls `envRootHoldsWork` without the lock and without clearing temps, so a temp left by a crash before the rename makes adoption hard-fail with `holds files but has no owner`. Narrow — it needs a lost record *and* a surviving temp — but adding the prefix/suffix to `isEnvRootBookkeeping` keeps both call sites consistent.

4. **`root_identity.go:139` fails closed on a corrupt record**, which is right, but the task then fails permanently and the error names only the relative path. Including `recordDir` would tell an operator what to remove.

5. **`diskusage_test.go:219` uses `ShortID(legacyTaskID)` as the legacy directory name**, but the legacy layout on `main` is `taskKey` (the last 12) — `ShortID` is the pre-#7347 shape. The test passes either way; it just doesn't cover the old directories actually on disk today.

## Also checked

- `WorkspaceSlug` / `IssueIdentifier` reach only `RootDirParams` and `PrepareParams`, never the runtime brief; the `ByteIdentical` tests are unmodified and pass.
- The 24-char cap keeps the two segments at 49 characters, level with `main`'s `36 + 1 + 12`, and `cache_windows_test.go` exercises a real checkout under worst-case labels and asserts the nested `refs/heads/agent/...` ref file is actually created. That is precisely the failure mode I was worried about last round. Writing the fixture to `%TEMP%\mw` rather than `t.TempDir()` is unusual, but the comment explains why and the reasoning holds.
- Dropping `parts[0] != task.WorkspaceID` from `shouldReusePriorWorkdir` is covered by `prov.WorkspaceID` — an ID check rather than a path check, which is stronger than what it replaced.
- Grouping GC batches by `meta.WorkspaceID` sends issue candidates with an empty workspace id down the single-issue `GetIssueGCCheck` path. Correct, just unbatched — a reasonable degradation.
- Locally: `go build ./...`, `go vet ./internal/daemon/... ./internal/handler/`, `go test -race ./internal/daemon/execenv/`, `./internal/daemon/`, and the handler workdir tests. The only failures (Hermes stores, daemon profile) fail identically on `main` and are environment-dependent. GitHub reports the branch `CLEAN` / `MERGEABLE` and the full matrix green, including `windows-execenv`.

Only the PATH column blocks. It's a small change, and the rest can land as-is or as follow-ups. Thanks for staying with this through three rounds of moving base — the stable ID-keyed index is the right shape for the problem.

iiwish:
Addressed the latest disk-usage review blocker in `fdd25e116` and synced the branch with current `main`.

- `TaskShort` now remains the physical task directory segment instead of being overwritten with the leading UUIDv7 prefix from `.task_owner` / `.gc_meta.json`.
- PATH rows therefore stay unique and correspond to the actual on-disk directory for legacy, readable, and active-root layouts.
- Updated the mixed-layout regression test to use the real legacy last-12-character directory shape and assert both physical path segments.

Local validation after the merge:
- `go test ./internal/daemon -run "TestScanDiskUsage_(MixedLayoutsUseMetadataIdentity|ReadableActiveRootUsesOwnerIdentityWithoutGCMeta)$" -count=1`
- `go test ./cmd/multica -count=1`
- `go vet ./internal/daemon/...`
- `git diff --check`

Ready for re-review; GitHub CI is running on head `0d5220ca0`.

iiwish:
@multica-eve @Bohan-J Integrated the latest stable-root lifecycle hardening and completed another full correctness pass at `5201bb8f4`.

In addition to the latest `a0c4694e6` changes, the index sweep now requires both the orphan age gate and an authoritative terminal/not-found task check before removing a complete record whose physical root is absent. This preserves the frozen path for old but non-terminal tasks and fails closed on transient status-check errors. Stale unpublished `.pending-*` directories and impossible empty final record directories remain age-gated and reclaimable; materialized roots and unreadable/corrupt records are always preserved. The shared `.task_roots` parent is never removed, interrupted `.task_owner-*.tmp` files are treated consistently as bookkeeping, and validation errors name the exact record directory.

Coverage now includes the pre-Prepare freeze window, terminal/running/unavailable tasks, materialized roots, corrupt records, empty final records, abandoned staging directories, concurrent claims, interrupted owner writes, disk attribution before `.gc_meta.json`, and stable physical PATH reporting.

Verification on the merged current `main`:
- full `go test -race ./internal/daemon/execenv`
- affected daemon GC/disk tests under `-race` (20x)
- lifecycle/GC stress regressions (50x)
- focused handler regressions (10x)
- `go vet ./internal/daemon/... ./internal/handler/ ./cmd/multica/`
- Windows cross-compilation for execenv, daemon, and handler
- complete GitHub CI matrix green, including `backend-tests` and `windows-execenv`

The PR is `CLEAN` / mergeable. Please re-review the current head when convenient.

Bohan-J (approved):
Approving. Thanks for sticking with this one — four rounds, three separate rebases onto a base that kept moving under you, and two of the harder problems in the PR (#7326's UUIDv7 prefix collision and the identity-vs-path question) were things you diagnosed and fixed yourself rather than waiting to be told.

The insight that makes this land is the one that isn't in the title: the old code treated the *directory name* as the task's identity, so making the name human-readable — and therefore mutable, since workspace slugs and issue prefixes change — could not be a rename. Moving identity into `.task_owner` / `.gc_meta.json` / the stable ID-keyed index is the actual change, and it leaves the system better than it found it: disk-usage and GC no longer infer a workspace from a path string.

Two things worth recording for the history:

**Maintainers pushed to this branch.** `a0c4694e6` and `333f6fa33` are ours, under the maintainer-edit permission, to close out review nits rather than send you round five: dropping a racy `os.Remove` of the index directory, teaching `isEnvRootBookkeeping` about unpublished `.task_owner-*.tmp` files so `findOwnedTaskRoot` can't wedge a task, naming `recordDir` in the fail-closed errors, and a test guard for the index sweep. Apologies for the unannounced first push — that should have come with a note.

**Your `PruneTaskRootIndex` is better than what we pushed.** Our first version gated index reclaim on age alone, which does not distinguish a leftover from the record a stalled re-dispatch is in the middle of reusing. Gating it on an authoritative `GetTaskGCCheck` — plus the 1h floor for a zero `GCOrphanTTL` — is the right answer, and we kept yours and dropped ours.

Verified on `333f6fa33`: full CI green including `windows-execenv`, prompt `ByteIdentical` intact, and the stable-root/GC race suites pass locally. The remaining trade-offs are understood and accepted: paths now surface workspace slug and issue key, external scripts hardcoding `<workspace UUID>/<task key>` need to handle both layouts, and rollback is asymmetric — an older daemon will start a fresh legacy root rather than reusing a readable one.

One follow-up we'll open separately rather than grow this PR: `Prepare` still re-resolves the index when `EnvRootPreclaimed` is set, so it derives a root the caller already claimed instead of being handed it. No reachable path diverges today, but it's load-bearing and undefended.

PR #7322, opened 2026-08-20
fix(auth): allow invited users to sign up

### LINKED ISSUE #4590: [Feature]: support true invite-only signup (let invited users sign up while self-serve signup is disabled)
### Deployment type

self-host

### What do you want and why?

On a self-hosted instance I want to restrict account creation to people who have been explicitly invited — without opening signup to an entire email domain or hand-maintaining ALLOWED_EMAILS.

Today the only signup controls are ALLOW_SIGNUP, ALLOWED_EMAILS, and ALLOWED_EMAIL_DOMAINS (server/internal/handler/auth.go → checkSignupAllowed). Because a user record is only created at first login, ALLOW_SIGNUP=false blocks all new users — including someone who was invited to a workspace but hasn't logged in yet (invitations are stored by email; invitee_user_id is resolved only if the user already exists). So there's no real invite-only mode: it's either open / domain-allowlist / email-allowlist signup, or a full freeze where nobody new can join even by invitation.

The docs (auth-setup → "Restricting who can sign up") list ALLOW_SIGNUP=false as "Disable self-serve signup entirely, invite-only", but the FAQ "Can you still invite people when signup is disabled?" confirms invited new users are blocked too — so the documented invite-only workflow isn't actually achievable.

Use case: an internal/self-hosted deployment that wants to onboard a handful of named people (including external collaborators) purely by invitation — without exposing signup to a whole company.com domain, and without editing config + redeploying every time someone is added.

Treat a pending workspace invitation as an implicit per-email allowance in checkSignupAllowed (server/internal/handler/auth.go): when a new user's email has a pending invitation, permit account creation even if ALLOW_SIGNUP=false and the email/domain isn't allowlisted. Concretely, in the findOrCreateUser / signup path, look up pending invitations by (lowercased) email before rejecting; if one exists, allow the signup and continue into the normal invitation-accept flow.

This would make ALLOW_SIGNUP=false mean "no open self-serve signup, but invited people can still join" — matching the documented intent.

_No response_

MgPotter2024:
## Summary

Invited users can now create accounts on locked-down self-hosted instances without an operator adding each address to `ALLOWED_EMAILS` and restarting the backend. A live pending workspace invitation acts as a per-email signup allowance; expired, accepted, declined, revoked, and missing invitations remain blocked, and the invitation is checked again when the account is created.

The self-hosting and member docs now describe this true invite-only behavior instead of the previous manual allowlist workaround.

## Validation

- Focused mocked auth policy suite passes, including pending-invitation and lookup-error cases.
- `go vet ./internal/handler ./pkg/db/generated` passes.
- Go packages outside the CLI package pass; the CLI package detects the Multica daemon-task marker in this managed workdir and rejects its normal-config test paths.
- SQLC generation passes and the database-backed handler regression suite compiles. Its endpoint cases require PostgreSQL, which is unavailable on this Mac.

Closes #4590

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex


Bohan-J (changes requested):
Thanks for this — the invite-as-signup-allowance model is the right call, and the implementation is careful in the places that matter. I checked this out locally and verified the parts that are easy to get wrong; details at the bottom. Two things should be resolved before merge, both in the docs.

## Must fix

### 1. The authoritative evaluation-order list is now stale, in all four locales

`apps/docs/content/docs/auth-setup.mdx:110-116` still says:

> 3. With no match, if `ALLOW_SIGNUP=false` — reject.
> 4. With `ALLOW_SIGNUP=true` but any allowlist configured and no match — still reject.

Neither holds any more. This PR updates the "Invitations and signup restrictions" section ~20 lines below, but leaves the ordered list above it untouched — so the page now contradicts itself, and the contradiction is between two adjacent sections. The ordered list is what an operator reads to decide whether their domain allowlist is absolute, so it's the more important of the two.

`auth-setup.zh.mdx`, `auth-setup.ja.mdx` and `auth-setup.ko.mdx` have the same list at exactly the same line numbers and need the same edit.

Suggested replacement for the English version:

```md
The evaluation order is:

1. The email matches `ALLOWED_EMAILS` — allow.
2. Or its domain matches `ALLOWED_EMAIL_DOMAINS` — allow.
3. With no allowlist configured and `ALLOW_SIGNUP=true` — allow.
4. Otherwise, if the email has a pending, unexpired workspace invitation — allow.
5. Otherwise — reject.
```

### 2. Step 4 is not just a doc drift — it is an undeclared widening of the allowlists

Before this PR, `ALLOW_SIGNUP=true` + any allowlist configured + no match was an unconditional reject; there was no invitation path at all. The new short-circuit at step 3 is `AllowSignup && both allowlists empty`, so as soon as an allowlist is configured, a non-matching email falls through to the invitation lookup and is admitted if invited.

Confirmed locally with a throwaway table test (`Config{AllowSignup: true, AllowedEmailDomains: []string{"company.com"}}`, probing `outsider@other.com`): no invitation → rejected, pending invitation → allowed. On `main` both are rejected. Same result with `AllowedEmails`.

I think this is the **right** behaviour — the docs already describe `ALLOW_SIGNUP` as only meaningful "when no allowlist is configured", so making both allowlist configurations behave the same way is more self-consistent, not less. But the PR title, description, and the new docs section all talk exclusively about `ALLOW_SIGNUP=false`, which means an operator who set `ALLOWED_EMAIL_DOMAINS=company.com` as a hard boundary gets it relaxed without notice.

Please either document it explicitly (the suggested list above covers it), or narrow the implementation so the invitation path only applies when `ALLOW_SIGNUP=false`. Worth a maintainer call on which.

### 3. Worth adding to the docs while you're there

Revoking an invitation does not reclaim an account that was already created. There is no user-deletion path in the codebase (no `DeleteUser` query, no admin user-management endpoint), and existing users are unconditionally admitted by the first branch of `checkSignupAllowed`. So on an `ALLOW_SIGNUP=false` instance, a mistyped invite email that the recipient actually uses creates a permanent account. That's a defensible boundary, but operators choosing "invite-only" will assume revocation is reversible, so one sentence in `auth-setup.mdx` would save a support round-trip.

## Nits (non-blocking)

**Two of the three entry points lose the error.** `SendCode` gained a `slog.Warn` on both branches, but the wrapped error from `findOrCreateUser` lands in `writeError(500, "failed to create user")` with no log at `server/internal/handler/auth.go:412-425` (`VerifyCode`) and `:593-606` (`GoogleLogin`). If the invitation lookup starts failing, that's exactly where an operator needs a breadcrumb.

**Test gaps.**
- Nothing covers the allowlist + invitation interaction from point 2 above — the one behaviour change that isn't described anywhere.
- `TestSendCodeInvitationSignupGate` covers `pending` / expired / `accepted`, but not `declined` or revoked.
- No regression test for an invitation that expires or is revoked between send-code and verify-code. The re-check itself is real (`findOrCreateUser` → `checkSignupAllowed`, `auth.go:184`) — it just isn't pinned by a test, and that's the boundary most worth locking down long-term.

**The mock's SQL dispatch is brittle.** `mockDB.QueryRow` routes on `strings.Contains(sql, "FROM workspace_invitation")`, which also matches `GetPendingInvitationByEmail` and `ListPendingInvitationsByWorkspace`. No current test reaches those, and a mismatch fails loudly with `expected *bool destination` rather than returning a wrong value, so it's fine today — just an easy trap when someone adds invitation coverage later.

## Verified locally, no action needed

Listing these so they don't get changed by accident:

- **Index coverage is correct.** `idx_invitation_invitee_email ON workspace_invitation(invitee_email) WHERE status = 'pending'` (migration 041) matches the new query's predicate exactly. Seeded 20k rows and ran `EXPLAIN ANALYZE`: `Index Scan using idx_invitation_invitee_email`, `Index Cond: (invitee_email = ...)`, 0.011ms. No seq-scan risk at scale.
- **Fails closed.** A lookup error propagates and blocks signup rather than admitting, and `TestSignupGatingReturnsPendingInvitationLookupError` pins it.
- **Email normalisation is consistent on both ends** — `ToLower(TrimSpace(...))` at invitation creation (`invitation.go:70`) and `ToLower` in `checkSignupAllowed`.
- **Revoked and expired are genuinely blocked.** Revoke is a `DELETE` (`invitation.sql:44`), so the row is gone; and a past-due invitation still sitting at `status = 'pending'` (the common case, since `ExpireStalePendingInvitations` only runs on re-invite) is caught by `expires_at > now()`.
- **The step-3 short-circuit is a good call** — `ALLOW_SIGNUP=true` with no allowlists does zero DB work, so the open-signup hot path is unaffected. Please keep it.
- **No `DISABLE_WORKSPACE_CREATION` bypass.** Signup does not auto-create a workspace (`findOrCreateUser` only calls `CreateUser`), and the check stays a single choke point at `workspace.go:212`.
- **Enumeration surface barely moves.** `/auth/send-code` is behind `authRL` (`RATE_LIMIT_AUTH`, default 5/min/IP, `router.go:1244`), and the 200-vs-403 split between registered and blocked emails already existed; this only adds invited emails to the 200 side.
- Re: your validation notes — the DB-backed handler tests do run here. `go build ./...`, `go vet ./internal/handler`, and the auth/invitation suites in `internal/handler` are all green on this branch.

PR #7323, opened 2026-08-20
MUL-6464 fix(agent): support ACP terminals for Kimi runtime

### LINKED ISSUE #7289: [Bug]: Kimi runtime — every Bash/Glob/Grep call fails with "ACP terminal capability is unavailable" after kimi-code 0.37.2 update
## Summary

After Kimi Code CLI self-updated `0.36.1 -> 0.37.2` on the local machine, every ACP task run via the `kimi` runtime immediately fails on any tool that routes through the ACP `terminal` capability (`Bash`, `Glob`, `Grep`), while `Read`/`Write`/`Edit` (routed through the ACP `fs` capability) continue to work fine. This makes any kimi-backed agent unable to use the `multica` CLI at all, since the runtime instructions require shelling out to `multica issue ...`.

## Exact error

Every Bash/Glob/Grep tool call in the kimi ACP session returns:

```json
{"type":"tool.result","result":{"output":"ACP terminal capability is unavailable","isError":true}}
```

The agent then reports (verbatim, from a real session transcript):

> The Bash tool itself is failing with "ACP terminal capability is unavailable" ... Even Glob is failing with "ACP terminal capability is unavailable". So most tools that interact with the file system are unavailable.

## Timeline (all times local, machine clock 2026-08-19/20)

- `~/.kimi-code/bin/kimi` binary mtime: **2026-08-19 20:18:17** — this is when kimi self-updated to 0.37.2 (confirmed via `kimi --version` and `~/.cache/kimi-code/native/{0.36.1,0.37.2}` both present in cache).
- **20:19:10** — first `ACP terminal capability is unavailable` failure, one minute after the binary swap, on the very next kimi ACP task.
- Every kimi ACP task since (confirmed across multiple independent sessions/workspaces) hits the identical error on every Bash/Glob/Grep call — 100% reproduction rate.
- A kimi ACP session captured on **2026-08-16** (three days prior, while still on 0.36.1) shows terminal calls succeeding normally in the same workspace/daemon setup — real `ls`, `multica issue get ... --output json`, etc. all returning real output, no capability error.

This strongly correlates the break with the kimi 0.36.1 -> 0.37.2 upgrade specifically, not with any local daemon/config change (daemon was untouched at 20:18; multica daemon build itself was from earlier that same morning, 2026-08-19T10:08:54Z).

## What I ruled out

- Not a `multica daemon restart` issue — restarting the daemon (done multiple times) never changed the outcome.
- Not the local `PreToolUse` `Bash` hook in `~/.kimi-code/config.toml` — that hook's script was independently found to be missing (`~/.kimi-code/hooks/check-bash.mjs` did not exist on disk), and recreating it as a no-op allow did **not** fix the issue. This also can't have been the root cause in the first place, since `Glob` (not matched by the `matcher = "Bash"` hook) fails identically.
- Not an LLM/provider-side error — the daemon-level task still reports `status=completed` with `agent_error=""`, because the kimi *process* exits cleanly; it's the *tool result content* that carries the capability error, which the agent then reports as a blocker in its own final text.

## Likely mechanism

Per the [ACP protocol spec](https://agentclientprotocol.com/protocol/initialization), a connecting client must declare `clientCapabilities.terminal: true` in its `initialize` request for the agent to permit terminal-routed tool calls — and the spec explicitly flags this field as "unstable and not yet part of the spec, and may be removed or changed at any point." It looks like kimi-code 0.37.2 tightened enforcement of this capability check (or changed the expected shape) in a way that the multica daemon's ACP client (currently 0.4.30, already the latest per `multica update`) does not yet satisfy. This project has hit this class of kimi ACP breakage before — see #5847 ("Kimi runtime model list empty on kimi-code 0.29 — ACP session/new schema changed to configOptions"), which suggests kimi-code's ACP contract has changed underneath multica more than once.

## Environment

- multica: `0.4.30 (commit: d563bfbc0, built: 2026-08-19T10:08:54Z)`, linux/amd64, go1.26.6
- kimi-code: `0.37.2` (broken), `0.36.1` (last known good, still cached at `~/.cache/kimi-code/native/0.36.1` but no supported way to pin the daemon to it — no `kimi update --version`/rollback flag, and `multica runtime profile set-path` wasn't usable since 0.36.1's cache only contains partial JS/native-addon artifacts, not a standalone launchable binary)
- OS: Linux, local daemon runtime mode

## Impact

Any agent bound to the kimi runtime cannot execute the `multica` CLI at all (it's required for reading issues, posting comments, changing status — the entire Ownership-mode workflow), so kimi-backed agents are completely non-functional for anything beyond pure file read/write.

## Ask

Is there a known fix, or a way to pin/downgrade the daemon's kimi ACP client capability declaration (or the invoked kimi binary) until this is patched? Happy to provide full `wire.jsonl` session transcripts if useful.

iiwish:
## Summary

Closes #7289

- Advertise ACP terminal capability for the Kimi runtime so Kimi 0.37.2 can route Bash/Glob/Grep through the client.
- Implement the ACP terminal create/output/wait/kill/release methods with workspace cwd, environment propagation, bounded output, and task-context cancellation.
- Keep terminal capability opt-in for Kimi and add regression coverage for initialization and terminal execution/output.

## Validation

- `go test ./pkg/agent -run 'TestACPManagedTerminal|TestKimiFreshSessionIncludesMcpServers|TestKimiResumeIncludesMcpServers|TestHermesClient|TestKimiBackend' -count=1`
- `go test -race ./pkg/agent -run '^TestACPManagedTerminal' -count=1`
- `go vet ./pkg/agent`
- GitHub CI `backend-tests`: `bash scripts/test-go.sh --race` passed across all Go packages, including the complete `./pkg/agent/...` suite.

All applicable GitHub checks pass, including backend build and migrations, the full Go race suite, frontend tests and build, Windows execution tests, installers, SQL generation checks, and the Go vulnerability scan.

### DIFF
--- a/server/pkg/agent/acp_terminal.go
+++ b/server/pkg/agent/acp_terminal.go
@@ -10,10 +10,14 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 	"unicode/utf8"
 )
 
-const defaultACPOutputByteLimit = 50_000
+const (
+	defaultACPOutputByteLimit             = 50_000
+	acpTerminalProcessGroupCleanupTimeout = 2 * time.Second
+)
 
 // acpTerminal is the daemon-side implementation of ACP's terminal/* methods.
 // ACP agents use these calls for shell tools when the client advertises the
@@ -92,12 +96,6 @@ func (t *acpTerminal) wait() {
 }
 
 func (t *acpTerminal) kill() error {
-	select {
-	case <-t.done:
-		return nil
-	default:
-	}
-
 	t.mu.Lock()
 	cmd := t.cmd
 	t.mu.Unlock()
@@ -105,6 +103,13 @@ func (t *acpTerminal) kill() error {
 		return nil
 	}
 	signalProcessGroup(cmd, syscall.SIGKILL)
+	// Cmd.Wait only describes the direct process. On Unix, descendants can
+	// remain in its process group after that process exits, especially when a
+	// background child redirects inherited output. Confirm the entire group is
+	// gone before terminal/kill or terminal/release reports success.
+	if runtime.GOOS != "windows" && !waitProcessGroupGone(cmd, acpTerminalProcessGroupCleanupTimeout) {
+		return fmt.Errorf("process group %d still active after %s", cmd.Process.Pid, acpTerminalProcessGroupCleanupTimeout)
+	}
 	return nil
 }
 
@@ -230,18 +235,20 @@ func (c *hermesClient) acpTerminalFor(id string) (*acpTerminal, bool) {
 func (c *hermesClient) acpTerminalRelease(id string) error {
 	c.terminalMu.Lock()
 	t, ok := c.terminals[id]
-	if ok {
-		delete(c.terminals, id)
-	}
 	c.terminalMu.Unlock()
 	if !ok {
 		return nil
 	}
-	select {
-	case <-t.done:
-	default:
-		_ = t.kill()
+	if err := t.kill(); err != nil {
+		return fmt.Errorf("release terminal %q: %w", id, err)
 	}
+
+	// Delete only after cleanup is confirmed so a failed release can be retried.
+	c.terminalMu.Lock()
+	if c.terminals[id] == t {
+		delete(c.terminals, id)
+	}
+	c.terminalMu.Unlock()
 	return nil
 }
 
--- a/server/pkg/agent/acp_terminal.go
+++ b/server/pkg/agent/acp_terminal.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -8,6 +9,8 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
+	"unicode/utf8"
 )
 
 const defaultACPOutputByteLimit = 50_000
@@ -17,13 +20,18 @@ const defaultACPOutputByteLimit = 50_000
 // terminal capability. Output is retained as a bounded tail, matching ACP's
 // truncation contract without allowing a command to grow memory unboundedly.
 type acpTerminal struct {
-	cmd       *exec.Cmd
-	mu        sync.Mutex
-	output    []byte
-	limit     int
-	truncated bool
-	done      chan struct{}
-	exitCode  *int
+	cmd        *exec.Cmd
+	mu         sync.Mutex
+	output     []byte
+	limit      int
+	truncated  bool
+	done       chan struct{}
+	exitStatus *acpTerminalExitStatus
+}
+
+type acpTerminalExitStatus struct {
+	exitCode *uint32
+	signal   *string
 }
 
 type acpTerminalOutputWriter struct {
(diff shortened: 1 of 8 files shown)


wegamekinglc:
will this pr be merged? I have the same issue

iiwish:
Thanks for confirming. This PR addresses the same ACP terminal issue and is currently passing CI. It is ready for maintainer review. @bohan-j, could you please take a look when convenient?

multica-eve (changes requested):
This change restores command execution and repository search for Kimi users affected by the 0.37.2 regression. The product behavior is right and this fix should land. One case still lets work started by an agent continue running after the task says it is finished, so it needs one correction before merge.

**Blocking: `terminal/release` can leave a detached background child running on Unix.**

In `server/pkg/agent/acp_terminal.go:94-99`, `kill` returns immediately once `done` is closed. `acpTerminalRelease` then also skips the kill at `server/pkg/agent/acp_terminal.go:230-245`, while `releaseProcessGroup` is a no-op on Unix at `server/pkg/agent/proc_other.go:35-37`.

For example, a terminal command that runs `sleep 30 >/dev/null 2>&1 &` can exit its direct shell with status 0 while `sleep` remains in the same process group. `cmd.Wait()` completes, `done` closes, and a later `terminal/release` deletes the terminal record and returns success without terminating that remaining process. Task-context cancellation cannot clean it up afterward because the direct `exec.Cmd` has already completed.

The EOF dependency has two branches:

```text
child stdout/stderr
  -> os/exec copy goroutines wait for EOF
  -> Cmd.Wait returns
  -> close(t.done)

inherited output fds: EOF is delayed, done stays open, release calls kill
redirected output fds: EOF arrives when the shell exits, done closes while the child is still alive, release skips kill
```

The current test in `server/pkg/agent/acp_terminal_other_test.go:18-66` covers only the first branch because the shell explicitly waits for its child. Please make release safely terminate and confirm the process group even when the direct command has already exited, and add a Unix regression test where the parent exits first, the child redirects output and remains alive, and release makes that child disappear.

I reproduced the process-group behavior independently. The targeted terminal tests, race test, `go vet ./pkg/agent`, and `git diff --check` pass, and the GitHub checks are green; the missing branch above is why those checks do not catch the leak.

git-tao:
wii this be merged any time soon? I have the same issue.

iiwish:
@multica-eve Thank you for the precise reproduction and detailed review. The blocking Unix release case is fixed on the latest head (82278f8f1).

The ACP terminal cleanup path now:
- signals the process group even when the direct command has already exited;
- waits for the Unix process group to disappear before terminal/kill or terminal/release reports success;
- returns an error and keeps the terminal registered if cleanup cannot be confirmed, so release can be retried.

The Unix regression covers the missing branch directly: the parent shell starts a background child with stdout/stderr redirected, records its PID, and exits successfully first. The test confirms the child remains alive before terminal/release, then immediately after release returns verifies that the child and process group are gone and the terminal is unregistered.

Validation includes repeated targeted runs, race-enabled tests, the full Go suite with and without -race, go vet ./pkg/agent, Windows cross-compilation, git diff --check, and green GitHub checks.

When convenient, could you please re-review the latest head? Thank you again.

@git-tao The reported leak is covered by this fix. Merge timing remains with the maintainers.

multica-eve (approved):
Thank you for the thoughtful follow-up and for adding the exact Unix regression coverage requested. I re-verified the updated process-group cleanup and the green CI at head 82278f8f; the blocking background-process leak is resolved. The output-limit hardening noted earlier can follow separately and does not block this fix. Great work!

PR #7324, opened 2026-08-20
test(dispatch): pin priority claim contract (MUL-6487)

RIDCorix:
## Summary

- rank only serialization-eligible runtime candidates by priority/FIFO
- re-check and claim the exact candidate inside the existing agent-capacity transaction
- prevent a blocked high row from silently downgrading to the same agent's low row ahead of another agent's eligible medium
- retain non-preemption, runtime-health, partial-success, and direct per-agent claim behavior
- document the atomic selector and expose selected priority metadata

## Verification

- full `go test ./internal/service -count=1` against isolated PostgreSQL 18 with migrations through 376
- daemon claim API plus backlog/member/write-path handler tests against the same database
- singular and batch blocked-head regressions, plus exact-candidate no-downgrade race regression
- `go vet ./...`
- regenerated sqlc with `make sqlc`
- full `go test ./... -count=1` from detached `/tmp` worktree with task-scoped Multica environment removed: exit 0

The daemon batch claim handler returned two correctly routed tasks and minted tokens in 13 ms. The blocked-head regressions remained in the 20–30 ms fixture range while changing the selected row from low to the globally eligible medium.

## Context

Workspace issue R-119. Reviewer reproduced a cross-agent priority inversion on the first PR head: the outer list ranked an unclaimable high row, then the per-agent inner selector substituted low. This revision binds the two layers to the same exact candidate so an eligibility/capacity race continues globally instead of downgrading locally.

### DIFF
--- a/docs/issue-task-dispatch-priority.md
+++ b/docs/issue-task-dispatch-priority.md
@@ -1,0 +1,70 @@
+# Issue task dispatch priority
+
+Multica applies issue priority when a runtime claims its next queued issue task. Priority changes which not-yet-started task receives an available slot; it never cancels, interrupts, or replaces a `dispatched`, `running`, or `waiting_local_directory` task.
+
+## Ordering contract
+
+Issue priorities map to the task queue as follows:
+
+| Issue priority | Queue rank |
+| --- | ---: |
+| `urgent` | 4 |
+| `high` | 3 |
+| `medium` | 2 |
+| `low` | 1 |
+| `none` or an unknown legacy value | 0 |
+
+The authoritative per-agent claim query is `server/pkg/db/queries/agent.sql::ClaimAgentTask`. It selects only `queued` tasks on the requested healthy runtime, excludes a task when the same agent already has active work for that issue or chat, and orders eligible rows by:
+
+1. task priority descending;
+2. task creation time ascending;
+3. task UUID ascending as the deterministic final tie-break.
+
+`server/internal/service/task.go::ClaimTaskForRuntime` and `ClaimTasksForRuntimes` list candidates with the same priority/FIFO order and then call `claimTask`, which uses `ClaimAgentTask` inside the agent capacity transaction. The singular and batch daemon poll APIs therefore share one final selector.
+
+If the highest-priority candidate's agent has no free slot, the claim loop may select the next eligible agent. Capacity is checked while the agent row is locked. A lower-priority healthy run that already owns a slot stays active; a newly queued high-priority task can only win the next slot that becomes available.
+
+## Entry-point funnel
+
+Workspace issue dispatch entry points persist either the issue's current priority or, for an automatic retry, the parent task's priority snapshot before the runtime claim boundary:
+
+| Entry point | Server funnel |
+| --- | --- |
+| Issue creation or assignment | `IssueService.maybeEnqueueOnAssign` → `TaskService.EnqueueTaskForIssue` |
+| Promotion from backlog to an active status | `IssueService.WillEnqueueRun` → the same issue enqueue path |
+| Agent-created sub-issue with an executable assignee/status | normal issue creation path |
+| Explicit agent mention or comment wake | `EnqueueTaskForMention`, `EnqueueTaskForThreadParent`, or assignee fallback → `enqueueMentionTaskWithCommentPlan` |
+| Autopilot `create_issue` | creates the issue, then uses the normal issue/squad enqueue path |
+| Manual rerun | `RerunIssue` → `enqueueRerunTask`; the new task uses the issue's current priority |
+| Automatic retry | `CreateRetryTask`; the child inherits the parent task priority |
+| Deferred issue-media or retry promotion | preserves the stored task priority; due rows are promoted in priority/FIFO order |
+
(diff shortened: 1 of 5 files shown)


Bohan-J (changes requested):
Thanks for this — the direction is right, and the two regression tests are real ones (they fail on `main`, where `ClaimAgentTask` downgrades a blocked high row to the same agent's low row). Binding the ranking layer and the claim layer to the same exact candidate is the root-cause fix, and `candidate.id ASC` is a nice touch for determinism.

## Product read

Before this PR, when a member's high-priority task couldn't start because another run was already active on the same issue, the freed slot went to **that same member's low-priority task** — even though another member's medium-priority task was queued on the same runtime. The queue showed the important work first while the unimportant work actually ran, with no way for anyone to see why.

After this PR, a freed slot always goes to the highest-priority task that is *actually startable right now*. Healthy runs are still never interrupted, and the claim log records what was selected and on what basis.

That's a real, user-visible priority inversion, and the fix stays inside the non-preemptive boundary we want. **Product-wise: yes, we want this.** One note before merge: the PR title says `test(dispatch)`, but the headline change here is the behavior fix (`fix(dispatch)` is the second commit). Titles land in the changelog — please retitle to `fix(dispatch)` so this doesn't get filed as a test-only change.

## Blocking

**Candidate filtering makes the empty-claim cache reachable in a state it previously wasn't.**

`ListQueuedClaimCandidatesByRuntime` now also filters out rows blocked by per-(issue, agent) serialization. That isn't only an ordering change — it changes the reachability of this branch in `server/internal/service/task.go`:

```go
if len(tasks) == 0 {
    s.EmptyClaim.MarkEmpty(ctx, runtimeKey, preSelectVersion)
    outcome = "empty_db"
    return nil, nil
}
```

(and the equivalent `withCandidates` check in step 5 of the batch path).

Before: as long as any `queued` row existed, `tasks` was non-empty and `MarkEmpty` was unreachable. After: if every `queued` row on a runtime is serialization-blocked — e.g. an issue is running and a new comment queues a follow-up on the same issue, which is the single most common flow on this platform — the candidate set is empty and we write a "this runtime has no work" verdict with a 3-minute TTL (`EmptyClaimCacheTTL`).

The same-runtime case looks safe: terminal transitions go through `NotifyTaskFinished` → `notifyRuntimeMayHaveWork` → `EmptyClaim.Bump`, which invalidates the verdict. But that chain has just been promoted from an optimization to a correctness dependency, and the PR neither analyzes nor tests it. At minimum this branch needs an answer:

The agent's active task is on runtime B while the blocked `queued` row is on runtime A — a state `claimTask`'s own comment acknowledges exists. When B's task finishes, only B is bumped; A's empty verdict stands, and the row on A can wait out the full 3-minute TTL. Before this PR, A re-queried Postgres on every poll and had no such stall.

Please add a regression test along the lines of "runtime has only serialization-blocked queued rows → marked empty → the next poll after the blocking task finishes must claim it", and either show the cross-runtime branch is unreachable or state why the stall is acceptable.

## Non-blocking

1. **The candidate query's plan changed and isn't verified.** The old `WHERE runtime_id = $1 AND status = 'queued' ORDER BY priority DESC, created_at ASC` was fully covered by the partial index `idx_agent_task_queue_claim_candidates (runtime_id, priority DESC, created_at ASC) WHERE status = 'queued'` — index-only scan, no sort step, exactly as migration 067's comment describes. The new query adds a trailing `candidate.id ASC` that isn't in the index plus a correlated `NOT EXISTS` anti-join, so the plan now picks up an (incremental) sort and a probe on `(agent_id, status)`. This is the hottest path the daemon walks on every poll, and Verification only reports fixture-scale timings (20–30 ms). An `EXPLAIN (ANALYZE, BUFFERS)` at a representative queue depth would settle it — most likely fine, since the per-runtime candidate set is small, but worth one look before merge.

2. **The deleted comment contained rationale that still holds.** The old comment in `server/pkg/db/queries/agent.sql` explained two things: why status is restricted to `queued` and not `dispatched`, and which partial index backs it. Both are still true (the first is arguably more important now), but the whole block was replaced. The new comment describes the new invariant and is good — please *append* rather than replace. And if the index is genuinely no longer fully used, say so here instead of dropping the reference.

3. **`ClaimAgentTask`'s `ORDER BY` is now effectively dead on the runtime poll path** — with a candidate ID bound, at most one row can ever match. It still governs the no-candidate `ClaimTask` used by the per-agent claim in `handler/daemon.go`. Not a problem, but worth a line in the SQL comment so nobody later changes that ordering assuming it still drives runtime polling.

4. **The cost of exact-candidate binding could be spelled out.** When the snapshot-to-transaction race fires, that agent is skipped entirely for the round rather than falling back to another of its rows. "The outer loop continues in global priority/FIFO order" is accurate but doesn't say that for *this* agent the round yields nothing. It self-heals on the next poll (candidates are recomputed, and no empty verdict is written when candidates exist), and the tradeoff is the right one — just worth a sentence in the doc.

5. **Test schema is behind `main`.** Verification says "migrations through 376"; `main` is at 389. The gap is DingTalk-related and almost certainly irrelevant, but given this touches the hottest scheduling path, please re-run against current `main`'s schema.

6. **`docs/issue-task-dispatch-priority.md` is genuinely good** — the rank table, entry-point funnel, out-of-scope section, and coverage list are all there, and I checked every symbol it references: they all exist. The one risk is how much Go and SQL identifier detail is in the prose; docs like this rot silently under refactors. Consider pointing the Regression coverage section at the test files rather than enumerating individual test function names.

No schema changes, no foreign keys, no new indexes — no migration or compatibility concerns.

PR #7331, opened 2026-08-20
MUL-6633 feat(selfhost): add daemon server URL override

### LINKED ISSUE #7105: [Feature]: separate daemon server URL from MULTICA_PUBLIC_URL for self-host setup hint
### Deployment type

self-host

### What do you want and why?

The "Add a computer" / runtime setup dialog builds the `multica setup self-host --server-url ...` hint from `MULTICA_PUBLIC_URL` (`daemonSetupURLsFromEnv` in `server/internal/handler/config.go`). The same variable is also what the backend uses to mint autopilot webhook URLs for GitHub.

Those two URLs are not always the same host. In our deployment:

- `MULTICA_PUBLIC_URL=https://hooks.multica.falconlabs.com` is an internet-facing load balancer whose security group only admits GitHub's IP ranges, because that is the only thing that should ever call the webhook endpoint.
- Daemons on employee laptops reach the API through a private load balancer, `https://api.multica.falconlabs.com`, over a ZTNA tunnel.

Result: the dialog tells every new user to run `multica setup self-host --server-url https://hooks.multica.falconlabs.com ...`, which times out for them, and we have to tell them by hand to replace it with `api.`. Setting `MULTICA_PUBLIC_URL` to the private host instead would break the GitHub webhook URLs, so the two uses conflict.

Add an optional override for the daemon-facing server URL, for example `MULTICA_DAEMON_SERVER_URL`, read by `daemonSetupURLsFromEnv` before falling back to `MULTICA_PUBLIC_URL` and then to the app URL. Document it next to `MULTICA_PUBLIC_URL` in the environment variables page and in `docker-compose.selfhost.yml`. Existing deployments keep the current behaviour when the new variable is unset.

Happy to send a PR if you are open to it.

mameikagou:
## What does this PR do?

Adds `MULTICA_DAEMON_SERVER_URL` so self-hosted deployments can use separate endpoints for public webhooks and private daemon traffic. When unset, the existing `MULTICA_PUBLIC_URL` and app URL fallback remains unchanged.

Thinking path: daemon setup URLs are resolved by `daemonSetupURLsFromEnv()`, so the override belongs there rather than in webhook or frontend logic.

## Related Issue

Closes #7105

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Prefer `MULTICA_DAEMON_SERVER_URL` when generating daemon setup URLs.
- Preserve the existing fallback chain and managed-cloud guard.
- Add handler tests and document the variable in the example environment, Compose configuration, and localized environment-variable references.

## How to Test

1. Set distinct daemon, public, and app URLs.
2. Run `go test ./internal/handler -run '^TestGetConfig' -count=1` from `server/`.
3. Confirm `daemon_server_url` uses the daemon override and falls back correctly when it is unset.

Validation: focused handler tests passed against PostgreSQL 17; Docker Compose rendered the expected backend value; `gofmt` and `git diff --check` passed.

## Checklist

No UI behavior or new runtime/tool surface is introduced, so screenshots and landing-copy changes are not applicable.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Used for the scoped implementation, regression tests, and validation described above.

## Screenshots (optional)

Not applicable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -75,6 +75,10 @@ MULTICA_APP_URL=${FRONTEND_ORIGIN}
 # avoid Host / X-Forwarded-Host spoofing when a self-hosted reverse proxy
 # is not hardened.
 MULTICA_PUBLIC_URL=
+# Optional API URL used only in self-host daemon setup commands. Set this
+# when the daemon reaches the API through a different URL than the public
+# webhook URL above. Falls back to MULTICA_PUBLIC_URL, then MULTICA_APP_URL.
+MULTICA_DAEMON_SERVER_URL=
 # Optional delay after SIGTERM/SIGINT before the existing graceful shutdown
 # begins. Accepts a non-negative Go duration such as 300s or 5m. Empty or 0
 # preserves the default behavior and starts shutdown immediately. The platform
--- a/apps/docs/content/docs/environment-variables.ja.mdx
+++ b/apps/docs/content/docs/environment-variables.ja.mdx
@@ -53,9 +53,14 @@ Kubernetes で shutdown hold を設定する場合、`terminationGracePeriodSeco
 | `FRONTEND_ORIGIN` | 空 | ユーザーがアクセスするフロントエンドの origin。CORS、cookie、招待リンクに使われます |
 | `MULTICA_APP_URL` | `FRONTEND_ORIGIN` にフォールバック | ユーザーが到達できる Web アドレス。CLI ログインとアカウント連携リンクが使います |
 | `MULTICA_PUBLIC_URL` | 空 | 公開 API アドレス。webhook URL とランタイム接続案内に使われます |
+| `MULTICA_DAEMON_SERVER_URL` | `MULTICA_PUBLIC_URL`、次に `MULTICA_APP_URL` / `FRONTEND_ORIGIN` にフォールバック | API プロセスが `multica setup self-host` コマンドに埋め込むサーバー URL。デーモンから到達する API URL が公開 webhook URL と異なる場合に設定します |
 | `CORS_ALLOWED_ORIGINS` | 空 | 追加で許可する HTTP origin。カンマ区切り |
 | `ALLOWED_ORIGINS` | CORS / フロントエンド設定にフォールバック | WebSocket origin の許可リスト。カンマ区切り |
 | `COOKIE_DOMAIN` | 空 | フロントエンドと API の host が異なり、ブラウザが API ドメインへ直接アクセスする場合は必須。単一ドメイン構成では空のまま |
+
+<Callout type="warning">
+`MULTICA_DAEMON_SERVER_URL` は認証なしの `/api/config` エンドポイントから返され、クライアントが読み取れます。公開設定として扱い、認証情報・トークンなどのシークレットを含めないでください。
+</Callout>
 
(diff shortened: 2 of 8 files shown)


multica-eve (changes requested):
This change lets self-hosted teams use one address for public integrations and another address for employee computers. Today, the “Add a computer” instruction can point at the public-only entrance and time out; afterward, administrators can provide the reachable internal entrance while existing single-address installations behave exactly as before.

## Product assessment

The user journey is coherent: an operator configures the two entrances once, and employees can then copy the command shown in “Add a computer” without knowing or manually correcting the company’s network topology. This affects only self-hosted installations that separate public callback traffic from first-party client traffic; it does not change defaults or require migration.

✅ **Product verdict: accept.** This solves a real deployment problem and follows the existing optional-override pattern.

## Implementation review

The override and fallback logic in `server/internal/handler/config.go:127-143` is focused and preserves the existing normalization and managed-cloud behavior. The regression test covers precedence and normalization, and the environment examples, Compose configuration, and localized documentation are updated consistently.

### Blocking

Please rebase onto the latest `main` and resolve the content conflict in `docker-compose.selfhost.yml:131-137`. The current `main` added `MULTICA_PLUGIN_SECRET_KEY`, `MULTICA_PLUGIN_SURFACE_ORIGIN`, `MULTICA_PLUGIN_API_URL`, and `MULTICA_PLUGIN_DIR` in the same block; the resolved file must retain all of those entries as well as `MULTICA_DAEMON_SERVER_URL`. The existing green checks cover commit `85ffccf63`, not the final combined tree, so please rerun the relevant checks after resolving the conflict.

### Documentation note

`MULTICA_DAEMON_SERVER_URL` is returned by the unauthenticated `/api/config` endpoint (`server/internal/handler/config.go:86-104`). That exposure is necessary for the UI and the value is not a credential, so it is not a design blocker. However, operators using this setting are likely isolating an internal hostname. Please add a short note to the environment-variable documentation stating that this value is client-visible/publicly readable and must not contain secrets.

There is a broader pre-existing distinction between URLs consumed by external services and URLs consumed by first-party clients: persisted attachment and avatar URLs still use `MULTICA_PUBLIC_URL`. That deserves separate follow-up work, but it was not introduced by this PR and should not expand this contribution’s scope.

Local verification on the current PR head:

- `go test ./internal/handler -run '^TestGetConfig' -count=1` passed.
- Compose rendered `MULTICA_DAEMON_SERVER_URL` into the backend service correctly.

🔄 **Request changes:** the product and implementation direction are good; please resolve the `main` conflict, preserve the newer Compose settings, document the public visibility of the value, and rerun validation on the rebased head.

mameikagou:
Rebased onto the latest `main` and resolved the Compose conflict while retaining all four plugin variables plus `MULTICA_DAEMON_SERVER_URL`. I also documented that the daemon URL is exposed through the unauthenticated config endpoint and must not contain secrets.

Validation on the rebased tree:
- `go test ./internal/handler -run '^TestGetConfig' -count=1`
- `docker compose -f docker-compose.selfhost.yml config --format json`

multica-eve (approved):
Re-reviewed commit `efe660797`. The requested changes are addressed:

- The branch is rebased and the Compose conflict is resolved while retaining all existing plugin variables alongside `MULTICA_DAEMON_SERVER_URL`.
- The environment-variable documentation now states that the value is returned by the unauthenticated config endpoint, is client-visible, and must not contain secrets.

Local verification passed:

- `go test ./internal/handler -run '^TestGetConfig' -count=1`
- Compose configuration rendering with the daemon URL and existing plugin variables
- conflict-free merge-tree check against the current `main`

No remaining code-review blockers. The backend CI job was still running at review time; branch protection should remain the final merge gate.

multica-eve:
Thank you for the thoughtful contribution and for addressing the review feedback so quickly. The clean rebase, preserved Compose settings, compatibility coverage, and clear documentation around the public visibility of the daemon URL made this straightforward to approve. We appreciate your help improving Multica's self-hosted deployment experience!

Issue #7086, opened 2026-08-17
[Bug]: errors after upgrading to 0.4.27


franciscocpg:
### Deployment type

Official App

### What happened?

After upgrading to 0.4.27, I'm unable to load issues

<img width="1168" height="288" alt="Image" src="https://github.com/user-attachments/assets/e9e47458-61b8-430e-8a26-e5c161e0e8f2" />

Also, the panel on the left is disappearing when I open a new tab.

### Steps to reproduce

Open the desktop app in the 0.4.27 version and try to see `issues` or open a new tab.

In the web seems to be working fine.

### Additional context (optional)

```shell

```

Bohan-J:
Thanks for the report, and sorry for the trouble.

Two things going on, and we'd like to rule out the cheap one first.

**1. Please reinstall 0.4.27 from a fresh download.** The 0.4.27 release had a bumpy publish on our side, and there's a window where an asset may have been uploaded incorrectly before being replaced. The artifacts currently on the [v0.4.27 release page](https://github.com/multica-ai/multica/releases/tag/v0.4.27) verify clean (correct version, built from the release commit, pointing at the production API), so if you happened to update during that window, a fresh install should sort it out. Grabbing the installer directly rather than using in-app update is the reliable way to get the current one.

**2. If it still happens after reinstalling**, it's likely a real regression on our side and we'd like your help pinning it down. Both of your symptoms — the board failing to load and the left panel vanishing on a new tab — are what we'd expect if the app loses track of the active workspace: every API call then goes out without its workspace header, and the shell drops the sidebar along with it.

Could you check the following?

- **OS and chip** (e.g. macOS 15 / Apple Silicon, Windows 11 / x64, Linux distro), and whether you installed via DMG/EXE/deb/rpm/AppImage or in-app auto-update.
- **Does the left panel come back** if you switch back to the first tab, or after restarting the app? Does the board load on the very first tab after a fresh launch, and only break once you open a second tab?
- **DevTools → Network** (View → Toggle Developer Tools): open Issues, find a failing `/api/issues` request, and tell us its status code and whether the request headers include `X-Workspace-Slug`. The Console output around that first failure would help too.
- Whether it happens in every workspace, or only one.

That last one is the decisive check — if `X-Workspace-Slug` is missing from the failing request, we know exactly where to look.

franciscocpg:
About the `1. Please reinstall 0.4.27 from a fresh download`.

[...]

PR #7346, opened 2026-08-20
MUL-6505: fix(web): only let the routed layout own the workspace singleton

coderberry:
## What does this PR do?

`apps/web/app/[workspaceSlug]/layout.tsx` writes the platform workspace
singleton from render:

```ts
if (workspace) {
  setCurrentWorkspace(workspaceSlug, workspace.id);
}
```

That is safe only while exactly one workspace layout is mounted. The App Router
can keep a previous workspace's layout mounted beside the incoming one, and
both instances re-render on their own query activity — so the unguarded write
lets them alternate, every render flipping the singleton back to its own slug,
indefinitely.

Two consumers read that singleton and break with it:

- `packages/core/realtime/provider.tsx` binds the realtime socket to the slug,
  so each flip tears the socket down and rebuilds it against the *other*
  workspace.
- `packages/views/editor/extensions/mention-suggestion.tsx` reads it
  synchronously when the user types `@`, then looks up the agent, member and
  squad caches by that id. Pointed at the other workspace those caches are
  cold, and it returns an empty list.

This PR gates the write on the live pathname. The URL is already the documented
source of truth for workspace identity, and `usePathname()` returns one value
shared by every mounted instance — so exactly one layout, the routed one, can
define it. No ownership handshake, no module global, no cleanup lifecycle.

Desktop keeps its own `singletonOwner` protocol in
`apps/desktop/src/renderer/src/components/workspace-route-layout.tsx`
(MUL-6303 / #7086) because it has no URL bar to appeal to. Note that protocol
guards the unmount *release*, not the write, so it does not address this case.

### Thinking path

Found in the field on a self-hosted v0.4.30 instance with two workspaces. The
backend log showed the browser's realtime socket reconnecting every 20–35s,
alternating `workspace_id` each time, sustained for over an hour. Each teardown
logged `websocket: close 1005 (no status)` — a client-side `close()` with no
code, i.e. `WSClient.disconnect()` from the provider's effect cleanup, so not a
proxy or network drop. Alongside it, every workspace-scoped sidebar query
(`/api/inbox`, `/api/inbox/unread-summary`, `/api/projects`, `/api/agents`,
`/api/agent-task-snapshot`) fired twice per cycle while page-specific queries
fired once — two mounted shells. The user separately reported that `@` offered
no agents, which is the same singleton read from `buildSyncItems`. A hard reload
collapsed the duplicate shell and both symptoms stopped: one connect, no flips,
every query single.

## Related Issue

No upstream issue — reported directly from a self-host deployment. Happy to file
one if you would prefer it tracked separately.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `apps/web/app/[workspaceSlug]/layout.tsx` — gate the `setCurrentWorkspace`
  render write on `workspaceSlug === workspaceSlugFromPathname(pathname)`.
- `apps/web/lib/workspace-slug-from-pathname.ts` — new pure helper. Decodes the
  segment, because route params arrive decoded while the pathname segment stays
  encoded, and falls back to the raw segment when the value is not valid
  encoding.
- `apps/web/lib/workspace-slug-from-pathname.test.ts` — boundary matrix for the
  helper (node environment, no DOM).
- `apps/web/app/[workspaceSlug]/layout.test.tsx` — adds `usePathname` to the
  `next/navigation` mock, makes the workspace query mock slug-aware, and adds
  three cases for the singleton.

## How to Test

1. `cd apps/web && pnpm exec vitest run "app/[workspaceSlug]/layout.test.tsx"`
2. Revert the guard on line 79 back to `if (workspace) {`. Two of the new cases
   fail — `stays silent while another workspace owns the URL` reports the stale
   layout wrote once, and `lets only the routed layout write when both are
   mounted` reports `[ 'acme', 'ws-acme' ]` where `[ 'globex', 'ws-globex' ]`
   was expected. That second failure is the bug: a layout writing its own slug
   over the workspace the URL is actually on.
3. Restore the guard; both pass.

Full local verification on this branch: `pnpm typecheck` (7/7 tasks),
`pnpm lint --filter @multica/web` (0 errors; the one warning in
`app/(auth)/login/page.tsx` is pre-existing), and the full `apps/web` suite
(30 files, 246 tests).

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — no visual change
- [ ] I have updated relevant documentation to reflect my changes — no documented behavior changes
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — N/A
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` — N/A

### Risks

The guard makes the write conditional, so the failure mode to weigh is a layout
that should own the singleton but does not — leaving it stale or null, which
would send workspace-scoped requests without `X-Workspace-Slug`. The comparison
is against the router's own pathname and the layout only ever renders under the
segment it is named for, so the two agree except while a navigation is in
flight, where the incoming layout's own render resolves it. Reserved slugs
cannot collide with a workspace slug, so a global route never matches.

## AI Disclosure

**AI tool used:** Claude Code (Claude Opus 5)

**Prompt / approach:** Started from a user-visible symptom on a self-host
("websocket updates are delayed and I can't @-mention agents"), not from the
code. Claude worked backwards from the backend logs — reading the `close 1005`
signature and the duplicated sidebar queries — to identify two mounted workspace
shells, then traced the singleton from the layout write through the realtime
provider and the mention suggestion builder to explain both symptoms from one
cause. It confirmed the deployed image's revision matched the source being read
before trusting that analysis, proposed the pathname guard over porting
desktop's ownership protocol (which guards the release, not the write), and
verified the new tests fail without the fix before committing.

### DIFF
--- a/apps/web/app/[workspaceSlug]/layout.test.tsx
+++ b/apps/web/app/[workspaceSlug]/layout.test.tsx
@@ -10,13 +10,16 @@ const state = vi.hoisted(() => ({
   } as { id: string; onboarded_at: string | null } | null,
   isAuthLoading: false,
   workspace: null as { id: string; slug: string } | null,
+  workspacesBySlug: {} as Record<string, { id: string; slug: string } | null>,
   workspaceError: false,
   hasBeenSeen: false,
+  pathname: "/acme/issues",
   replace: vi.fn(),
 }));
 
 vi.mock("next/navigation", () => ({
   useRouter: () => ({ replace: state.replace }),
+  usePathname: () => state.pathname,
 }));
 
 vi.mock("@multica/core/auth", () => ({
@@ -29,10 +32,11 @@ vi.mock("@multica/core/auth", () => ({
 }));
 
 vi.mock("@multica/core/workspace", () => ({
-  workspaceBySlugOptions: () => ({
-    queryKey: ["workspace-by-slug", "acme"],
+  workspaceBySlugOptions: (slug: string) => ({
+    queryKey: ["workspace-by-slug", slug],
     queryFn: async () => {
       if (state.workspaceError) throw new Error("temporary failure");
+      if (slug in state.workspacesBySlug) return state.workspacesBySlug[slug];
       return state.workspace;
     },
   }),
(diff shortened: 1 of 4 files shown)


multica-eve (approved):
@coderberry, thank you for the exceptionally thorough field diagnosis and regression coverage here. The pathname guard is focused, keeps the URL as the source of truth for the web workspace identity, and preserves Desktop's separate ownership model. We verified the targeted regression tests, the full web test suite, typecheck, and lint locally.

Approved — thanks again for the thoughtful contribution!

PR #7350, opened 2026-08-20
feat(agent): add Junie CLI as a native ACP runtime

### LINKED ISSUE #1163: [Feature]: Support additional runtimes
### Deployment type

multica.ai (hosted)

### What do you want and why?

Add Kiro and Junie CLI to supported runtimes.
ALso update documentation to indicate the Gemini and hermes are already supported.

_No response_

_No response_

ADanMan:
## Summary

Adds Junie (JetBrains' coding agent) as a native ACP runtime. Kiro CLI support for #1163 already landed in #1780; multica-eve kept this issue open specifically for the remaining Junie CLI part.

## Change

- `server/pkg/agent/junie.go`: new ACP backend (`junie --acp=true`), mirroring the Kiro backend's session lifecycle — `session/new`, `session/load` resume, `session/set_model`, provider-error promotion, MCP capability filtering.
- `agent.go` (`SupportedTypes`, `New`, launch header), `models.go` (empty-catalog degradation matching the `qwen` precedent), `internal/daemon/agents_probe.go` + `config.go`, `scripts/agent-cli-command-names.txt`, and the frontend runtime-provider wiring (`packages/core/types/agent.ts`, `packages/core/agents/mcp-support.ts`, `packages/views/runtimes/components/provider-logo.tsx`).
- Migration 377: widens the `runtime_profile.protocol_family` CHECK constraint (`ALTER TABLE ... NOT VALID`, matching migration 370's `dim` pattern).
- Docs (`providers.mdx` + locales, `README.md`, `CLI_AND_DAEMON.md`, `SELF_HOSTING.md`).

Follow-up fix in the second commit: `runtimeConfigPath` (`internal/daemon/execenv/runtime_config.go`) didn't have a case for `"junie"`, so it fell through to `default: return ""` and `InjectRuntimeConfig` silently skipped writing `AGENTS.md` into every Junie task's workdir — every Junie-assigned task would have run with no workflow brief at all (no `multica issue status`/`comment add` instructions, no project/skill context). Same failure mode traecli shipped with once before (#4945/#4946). Fixed by adding `"junie"` alongside `"kiro"` (both ACP backends that load `AGENTS.md` from the task cwd themselves), plus registering `junieBlockedArgs` in `launchPrefixBlockedArgs` and adding `"junie"` to the metrics `knownRuntimeProviders` allowlist. Added `TestRuntimeConfigPathCoversEverySupportedType`, a guard that walks every `agent.SupportedTypes` entry and fails if `runtimeConfigPath` returns `""` for it, so this class of bug can't recur silently for any future runtime addition.

## Test plan

- `gofmt -l`, `go vet ./...`, `go build ./...` clean.
- Full `go test ./pkg/agent/...`, `./internal/daemon/execenv/...`, `./internal/metrics/...` — all pass, including the new guard test and `junie_test.go`'s ACP fake-script coverage (backend selection, blocked-arg stripping in both `--acp=false` and `--acp false` forms, `session/load` resume, `set_model` failure handling).
- `pnpm --filter @multica/views test` for `provider-logo` — pass.

Fixes #1163

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -220,6 +220,7 @@ The daemon auto-detects these AI CLIs on your PATH:
 | [Reasonix](https://github.com/esengine/DeepSeek-Reasonix) | `reasonix` | DeepSeek-focused ACP coding agent (run `reasonix setup` first) |
 | Dim | `dim` | DimCode ACP coding agent (speaks the ACP protocol via `dim acp`) |
 | Kiro CLI | `kiro-cli` | Kiro ACP coding agent |
+| [Junie](https://junie.jetbrains.com/) | `junie` | JetBrains Junie ACP coding agent (ACP via `junie --acp=true`) |
 | [Qoder CLI](https://docs.qoder.com/) | `qodercli` | Qoder ACP coding agent |
 | [Qoder CN CLI](https://help.aliyun.com/en/lingma/qodercli-cn/product-overview/what-is-qoder-cli-cn) | `qoderclicn` | Qoder CN ACP coding agent |
 | [Trae](https://docs.trae.cn/cli) | `traecli` | ByteDance TRAE CLI (ACP via `traecli acp serve`) |
@@ -328,6 +329,8 @@ Agent-specific overrides:
 | `MULTICA_DIM_MODEL` | Override the Dim model used |
 | `MULTICA_KIRO_PATH` | Custom path to the `kiro-cli` binary |
 | `MULTICA_KIRO_MODEL` | Override the Kiro model used |
+| `MULTICA_JUNIE_PATH` | Custom path to the `junie` binary |
+| `MULTICA_JUNIE_MODEL` | Override the Junie model used |
 | `MULTICA_QODER_PATH` | Custom path to the `qodercli` binary |
 | `MULTICA_QODER_MODEL` | Override the Qoder model used |
 | `MULTICA_QODERCLICN_PATH` | Custom path to the `qoderclicn` binary |
@@ -352,7 +355,7 @@ The daemon launches Qwen Code as `qwen -p <prompt> --output-format stream-json`.
 
 #### `mcp_config` on ACP runtimes
 
-ACP-family runtimes — Hermes, Kimi, Kiro, Grok, Qoder, Reasonix, Trae, QwenPaw, MiniMax Code, Dim, and any custom runtime profile whose `protocol_family` is one of them — receive MCP servers **over the ACP session protocol**, not through a config file. The daemon translates the agent's `mcp_config` into ACP's `McpServer` array and sends it with `session/new`, and again with that runtime's resume request (`session/resume` on Hermes, Kimi, Qoder and Reasonix; `session/load` on Kiro, Grok, Trae, QwenPaw and Dim) so a resumed task keeps the same tools. MiniMax Code 0.1.2 advertises no session-loading capability, so a later run falls back to a fresh session.
+ACP-family runtimes — Hermes, Kimi, Kiro, Junie, Grok, Qoder, Reasonix, Trae, QwenPaw, MiniMax Code, Dim, and any custom runtime profile whose `protocol_family` is one of them — receive MCP servers **over the ACP session protocol**, not through a config file. The daemon translates the agent's `mcp_config` into ACP's `McpServer` array and sends it with `session/new`, and again with that runtime's resume request (`session/resume` on Hermes, Kimi, Qoder and Reasonix; `session/load` on Kiro, Junie, Grok, Trae, QwenPaw and Dim) so a resumed task keeps the same tools. MiniMax Code 0.1.2 advertises no session-loading capability, so a later run falls back to a fresh session.
 
 Nothing is written to the runtime's own config file, and the runtime's own file is not read or merged. `~/.hermes/…`, `~/.jcode/mcp.json` and the like stay untouched; an agent's servers travel with its tasks instead of being installed per machine.
 
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ authenticated, so switching providers is a dropdown, not a migration.
 | Qwen Code | `qwen` | QwenPaw | `qwenpaw` |
 | Reasonix | `reasonix` | Trae CLI | `traecli` |
 | DeepSeek Harness | `dsh` | Oh-My-Pi | `omp` |
-| MiniMax Code | `mcode` | — | — |
+| MiniMax Code | `mcode` | Junie | `junie` |
 | Dim | `dim` | | |
 
 Installing and authenticating them: [Install an agent runtime](https://multica.ai/docs/install-agent-runtime) ·
--- a/SELF_HOSTING.md
+++ b/SELF_HOSTING.md
@@ -124,6 +124,7 @@ You also need at least one AI agent CLI installed:
 - [Reasonix](https://github.com/esengine/DeepSeek-Reasonix) (`reasonix` on PATH; run `reasonix setup` first)
 - Dim (`dim` on PATH)
 - Kiro CLI (`kiro-cli` on PATH)
+- [Junie](https://junie.jetbrains.com/) (`junie` on PATH)
 - Qoder CLI (`qodercli` on PATH)
 - Qoder CN CLI (`qoderclicn` on PATH)
 - Trae CLI (`traecli` on PATH)
--- a/apps/docs/content/docs/providers.ja.mdx
+++ b/apps/docs/content/docs/providers.ja.mdx
@@ -30,6 +30,7 @@ Multica は、普段使っている AI コーディングツールの代わり�
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | 1 回の実行専用の `HERMES_HOME/skills/` |
+| Junie | `junie` | ✓ | ✓ | — |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
--- a/apps/docs/content/docs/providers.ja.mdx
+++ b/apps/docs/content/docs/providers.ja.mdx
@@ -30,7 +30,7 @@ Multica は、普段使っている AI コーディングツールの代わり�
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | 1 回の実行専用の `HERMES_HOME/skills/` |
-| Junie | `junie` | ✓ | ✓ | — |
+| Junie | `junie` | ✓ | ✓ | `.junie/skills/` |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
--- a/apps/docs/content/docs/providers.ko.mdx
+++ b/apps/docs/content/docs/providers.ko.mdx
@@ -30,7 +30,7 @@ Multica는 사용 중인 AI 코딩 도구를 대체하지 않습니다. 런타�
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | 실행별 `HERMES_HOME/skills/` |
-| Junie | `junie` | ✓ | ✓ | — |
+| Junie | `junie` | ✓ | ✓ | `.junie/skills/` |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
--- a/apps/docs/content/docs/providers.ko.mdx
+++ b/apps/docs/content/docs/providers.ko.mdx
@@ -30,6 +30,7 @@ Multica는 사용 중인 AI 코딩 도구를 대체하지 않습니다. 런타�
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | 실행별 `HERMES_HOME/skills/` |
+| Junie | `junie` | ✓ | ✓ | — |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
--- a/apps/docs/content/docs/providers.mdx
+++ b/apps/docs/content/docs/providers.mdx
@@ -30,7 +30,7 @@ Install steps are in [Install AI coding tools](/install-agent-runtime).
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | per-task `HERMES_HOME/skills/` |
-| Junie | `junie` | ✓ | ✓ | — |
+| Junie | `junie` | ✓ | ✓ | `.junie/skills/` |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
--- a/apps/docs/content/docs/providers.mdx
+++ b/apps/docs/content/docs/providers.mdx
@@ -30,6 +30,7 @@ Install steps are in [Install AI coding tools](/install-agent-runtime).
 | DevEco Code | `deveco` | ✓ | — | `.deveco/skills/` |
 | Grok | `grok` | ✓ | ✓ | `.grok/skills/` |
 | Hermes | `hermes` | ✓ | ✓ | per-task `HERMES_HOME/skills/` |
+| Junie | `junie` | ✓ | ✓ | — |
 | Kimi CLI | `kimi` | ✓ | ✓ | `.kimi/skills/` |
 | Kiro CLI | `kiro-cli` | ✓ | ✓ | `.kiro/skills/` |
 | MiniMax Code | `mcode` | — | ✓ | `.minimax/skills/` |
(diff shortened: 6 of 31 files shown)


Bohan-J (changes requested):
Thanks for the PR — the overall direction is right and the structure follows the existing ACP provider layout closely. Three things should be fixed before merge; all of them are in the new Junie path only.

### 1. Bound skills are never discoverable by Junie

`skillsDirPath` in `server/internal/daemon/execenv/context.go` has no `junie` case, so skills land in the generic `.agent_context/skills` fallback. `localSkillRootsForProvider` in `server/internal/daemon/local_skills.go` has no Junie entry either. Meanwhile the runtime prompt tells the agent its skills were auto-discovered.

JetBrains documents project-level `.junie/skills/<name>/SKILL.md` and user-level `~/.junie/skills` (https://junie.jetbrains.com/docs/agent-skills.html). As written, a Junie agent with bound skills sees the skill names but cannot read their instructions — a silent capability gap rather than a visible error.

Suggested fix: map the project dir to `.junie/skills`, add `~/.junie/skills` to the user-level roots, and add Junie to the file-based provider matrix for injection, reuse cleanup, user-content preservation, and the sidecar round-trip tests.

### 2. Init-path failures can hang forever in process reclamation

`server/pkg/agent/junie.go:187-194` registers `defer cancel()` and the cleanup defer separately. Under LIFO the cleanup runs first (`stdin.Close()`, then `cmd.Wait()`), so `cancel()` is unreachable until `Wait()` returns. A child that closes stdout/stderr but stays alive — or ignores stdin EOF — will hang the error paths of `initialize`, `session/load`, `session/new`, and model selection. With no timeout configured this hangs indefinitely.

`server/pkg/agent/hermes.go:428` already solves this in a single defer: `stdin.Close()` → `cancel()` → `cmd.Wait()`, with a comment explaining exactly this failure mode. Please match it, and add a regression test with a fake process that closes its output pipes and keeps running.

### 3. A rejected resume never falls back to a fresh session

`server/pkg/agent/junie.go:232-242` returns every `session/load` error as a plain failure and never sets `resumeRejected` when `isACPSessionNotFound(err)` matches. Junie is also (correctly) absent from `resumeRejectionUndetectable` in `server/pkg/agent/agent.go:351`, which means the daemon reads `ResumeRejected: false` as "checked, and it was not a rejection" — so the fresh-session retry never fires. A user whose session ID goes stale hits the same dead ID on every subsequent turn.

`mcode.go:246` shows the pattern; `dim.go:436` and `qwenpaw.go:225` do the same. Please set `resumeRejected` on the ACP `-32002` / `-32602` "Session not found" cases and cover it with a test.

### Non-blocking

- `apps/docs/content/docs/install-agent-runtime*.mdx` has no Junie install/login section yet, but the providers page links users there.
- The hand-rolled "J" placeholder icon: the ACP Registry ships an official Junie icon (https://github.com/agentclientprotocol/registry/tree/main/junie). Worth confirming licensing and reusing it, or falling back to the shared generic placeholder.
- No real Junie handshake appears to have been exercised. Before merge, someone with a logged-in Junie CLI should smoke-test a first turn, a permission request, MCP, skill loading, and a resumed turn. The docs confirm the launch invocation and the AGENTS/MCP conventions look right (https://junie.jetbrains.com/docs/junie-cli-acp.html), but that is not a substitute for one live run.

Happy to re-review once the three blockers are addressed.

ADanMan:
Thanks — all three blockers addressed in one commit on `feat/junie-cli-runtime`.

**1. Junie skills were never discovered.** You were right that this was a silent no-op. `skillsDirPath` had no `junie` case, so skills went to `.agent_context/skills/`, which the Junie CLI never scans — while the runtime brief still told the agent "You have the following skills installed (discovered automatically)". Names, no path, no readable SKILL.md.

Registered at every surface, following the Kiro precedent:
- `execenv/context.go` — `case "junie"` returns `<workDir>/.junie/skills`.
- `daemon/local_skills.go` — `~/.junie/skills` user root in `localSkillRootsForProvider`, plus the doc-comment bullet.
- `execenv/runtime_config.go` — the per-provider routing line that was missing next to Kiro.
- `providers.mdx` + zh/ja/ko — the Junie row's `—` is now `` `.junie/skills/` ``.

Reuse cleanup (`removeReusedManagedSkillDirs`) and user-content preservation both derive from `skillsDirPath`, so they were fixed by the same change; tests now pin that rather than assume it. Paths are per https://junie.jetbrains.com/docs/agent-skills.html: project `<projectRoot>/.junie/skills/<name>/` with a required SKILL.md, user `~/.junie/skills/<name>/`, project wins on a name collision.

Also added `TestSkillsDirPathCoversEverySupportedType` — the `skillsDirPath` twin of the `runtimeConfigPath` guard already in this PR. It loops `agent.SupportedTypes` and fails on the `.agent_context` fallback, so the next provider can't ship the same gap. `codex` (CODEX_HOME) and `hermes` (per-task HERMES_HOME) are documented by-design exemptions. `dim` is exempted with a TODO: it has the identical pre-existing gap (in `SupportedTypes` and `allFileBasedProviders`, absent from both switches). Happy to fix it here or separately, your call — I left it out to keep this diff scoped to the review.

**2. `cancel()` unreachable before `cmd.Wait()`.** Confirmed exactly as described. `defer cancel()` was registered first, so LIFO ran the cleanup defer — `stdin.Close()`, `cmd.Wait()` — before it. With no per-run timeout `runContext` returns a plain `WithCancel`, and neither `cmd.Cancel` nor `cmd.WaitDelay` is set, so the hang was unbounded on all four init-path early returns (initialize, session/load, session/new + empty id, set_model). Moved `cancel()` between `stdin.Close()` and `cmd.Wait()`, matching `hermes.go` and `mcode.go`, with the failure mode documented inline. The success path at the bottom already cancels unconditionally and is unchanged.

One note on the regression test, since it's a trap worth flagging: a port of the Hermes happy-path fixture passes against the *unfixed* code here, because Junie's success path already calls `cancel()` before draining. So the fixture lingers on an init-path early return instead — table-driven over `initialize` and `session/new`, closing stdout/stderr and staying alive. I verified it by reverting the fix: both subtests fail with "result channel did not close; cmd.Wait likely ran before cancel". The per-run timeout is set well above the assertion windows so the deadline can't unblock `Wait` and produce a false green, and a regression fails in ~5s rather than hanging CI.

Related, not fixed here: the same ordering exists in `kiro.go:219`, `qwenpaw.go:176`, `kimi.go:197` and `reasonix.go:232` — Junie copied a latent-bug precedent. Worth a follow-up.

**3. `session/load` didn't detect a rejected resume.** Correct, and the consequence was the one you called out. The branch omitted `isACPSessionNotFound`, so `ResumeRejected` was the zero value; `shouldRetryWithFreshSession` reads false as "it checked and this wasn't a rejection", and no poisoned-session classifier matches "session not found (code=-32602)", so the dead id stayed the resume pointer on every subsequent turn. Added the shared-helper check plus a Warn log, mirroring `qwenpaw.go`. Used `isACPSessionNotFound` rather than a code comparison — a code-only check would false-positive on Junie's own `-32602` "model not available". Covered by a new test plus a negative control asserting the flag stays false when `session/load` succeeds.

**Non-blocking notes.** Docs section and the official ACP-registry icon: agreed, will take both — say the word if you want them in this PR or a follow-up.

To be straight about it: **no live Junie handshake has been run.** I don't have the binary, and per CLAUDE.md the default suite must never resolve a user-installed agent CLI, so every test here is pure-filesystem or driven by a fake executable. The two adapter caveats in the file header stand unchanged — generic kind-based permission approval, no `junieToolNameFromTitle` mapper — both pending a real session. The skills paths are from JetBrains' published docs, not observed behaviour.

Verification: `gofmt -l`, `go vet ./...`, `go build ./...`, `go test ./pkg/agent/... ./internal/daemon/...` all clean.

Bohan-J:
Reviewed `76f1cc89`. All three blockers are closed, and I want to be specific about how I checked, because "the tests pass" is a weak claim when the tests are new: I reverted each fix individually and confirmed the corresponding test actually catches it.

- Skills: removed the `case "junie"` from `skillsDirPath` → `TestSkillsDirPathCoversEverySupportedType/junie` and `TestJunieUsesNativeProjectSkillRoot` both fail.
- Cleanup ordering: restored the separate `defer cancel()` → both subtests fail with `result channel did not close; cmd.Wait likely ran before cancel`, in ~5s as designed.
- Resume rejection: dropped the `isACPSessionNotFound` branch → `expected ResumeRejected=true` fails.

Locally green on your head commit: `go test ./pkg/agent/...`, `./internal/daemon/execenv/...`, `./internal/daemon/ -run LocalSkills`, `./internal/migrations`, `./internal/metrics/...`, plus `./internal/daemon/execenv/ -run ByteIdentical` (our prompt-cache prefix invariant — unaffected by this PR) and a clean `git diff --check`. I also verified the skills paths against JetBrains' docs independently: project `.junie/skills/<name>/`, user `~/.junie/skills/`, project wins a name collision. Matches what you implemented.

Two things above what I asked for, both worth calling out:

**`TestSkillsDirPathCoversEverySupportedType`.** I asked you to add Junie; you closed the category instead. A new provider can no longer inherit the `.agent_context` fallback silently — it either gets a native mapping or an explicit, documented exemption. This PR is itself the evidence that the gap was reachable, and the exemption table is honest about `dim` rather than quietly widening the diff to hide it. That is the right instinct.

**The `cancel()`-before-`Wait()` ordering in `kiro.go:219`, `qwenpaw.go:176`, `kimi.go:197`, `reasonix.go:232`.** I checked `main`: you are right, all four are identical, and all four ship today. So this was our latent bug that you inherited by following the nearest precedent — thank you for reporting it rather than just fixing your own copy. We will track it separately; the fix is mechanical but each one wants the init-path fixture you worked out here, and that belongs in its own PR rather than riding along on yours.

Your note about the happy-path fixture passing against the unfixed code is the most useful thing in your write-up. A test that goes green either way is worse than no test, and it would have shipped if you had ported the Hermes fixture verbatim.

## Your open questions

- **`dim`:** separate PR. Keeping this diff scoped was the right call.
- **Docs section:** please do include it here. `providers.mdx` already tells users "Install steps are in Install AI coding tools", and `install-agent-runtime{,.zh,.ja,.ko}.mdx` has no Junie section, so today that link is a dead end for exactly the people this PR is for.
- **Icon:** follow-up is fine. One correction to the code comment — the ACP Registry does publish a Junie icon (https://github.com/agentclientprotocol/registry/tree/main/junie), so "no official mark asset was available" isn't quite right. Not worth another round on its own.

One small thing I noticed, not a blocker and not for this PR: Junie resolves guidelines as `.junie/AGENTS.md` → root `AGENTS.md` → `.junie/guidelines.md`. We write the root `AGENTS.md`, so a project that already has `.junie/AGENTS.md` would silently shadow our runtime brief. Narrow — it needs that file to pre-exist — but it is the same silent-drop shape as the skills bug. I will file it on our side.

## The part I owe you an honest conversation about

The code is settled. What is not settled is support, and I would rather say this before merging than after.

You were straight with us that no live Junie handshake has been run and that you do not have the binary. That is the right way to report it, and it is also the crux. We do not use Junie either — we cannot verify this against the real CLI, cannot confirm the permission-approval and tool-name-mapping caveats in your file header, and cannot even settle something as basic as whether `--acp=true` is accepted where JetBrains' docs write `junie --acp true`. Merging means shipping a runtime nobody in the conversation has watched run.

We have handled that situation before by naming an area community-maintained, with a volunteer as its first stop. Here is the shape of it, stated plainly:

**What we commit to.** It ships in every release. We keep it compiling and its tests green through shared-layer refactors — the rebase burden becomes ours, not yours. Junie-specific issues get routed to you instead of sitting unowned.

**What we would ask of you.** Be the first stop for `server/pkg/agent/junie.go` — Junie-specific bugs and behavior questions come to you first, best-effort. Volunteer terms: "we route these to you", not "you are on the hook to fix them". We would list you by name on our community-maintained page and add a Junie option to the issue form's Area field so reports reach you instead of us guessing.

**The exit rule.** Community-maintained areas carry no support SLA. If this breaks in a way we cannot fix without real Junie access and no fix appears for a few releases, we may deprecate it rather than leave it quietly broken. That rule exists to protect you more than us: it is our commitment that if you get busy or move on, the worst outcome is that we retire the runtime — not that anyone chases you for a fix. A volunteer role should have a no-fault exit.

**And you can say no.** Declining does not block the merge. It changes the label — community-maintained, no owner — and tightens when deprecation would kick in. It would not change how we read this contribution, which has been unusually careful.

Given that you do not have the binary, the honest version of the question is narrower than the usual one: **is Junie something you use or expect to have access to going forward?** If it is, the arrangement works and we would be glad to have you. If it is not — if this was a docs-driven contribution you are handing off — say so plainly and we will merge it as community-maintained with no owner and a shorter leash. That is a perfectly good outcome and a much better answer than a reluctant yes.

Either way, before this merges we need one live run against a real Junie CLI: first turn, permission request, MCP, skill loading, and a resumed turn — with the `--acp` flag form settled. If you can get access, that closes it. If not, tell us and we will find someone who can.

For context, this isn't invented for your PR: per-integration code owners are how Home Assistant runs, and the same idea backs Linux's `MAINTAINERS` and Kubernetes' `OWNERS`. It works because each side does what only it can — we handle mechanical maintenance, the owner covers the verification that needs real access.

No rush on the answer.

ADanMan:
Update: I validated this against a **real Junie CLI 1468.30**, driving its ACP server over stdio the way `junie.go` does. Here's what the live binary actually does, an honest correction to my earlier blocker-3 claim, and two new findings.

## Correction on blocker 3 — my first fix matched the wrong wording

You were right the first time, and my first "fixed" claim was wrong. Junie's unknown-session wording is `-32602 "Session <id> not found"` — the id sits **between** the noun and the verdict, so the shared `isACPSessionNotFound` substrings (`"session not found"` / `"no session found"` / `"unknown session"`) never matched it. My original test fixture invented Hermes's wording, so it went green against a scenario the real CLI can't produce — exactly the go-green-either-way trap I flagged on the cleanup-ordering test. Bad on me for not catching my own copy of it.

Fixed with a provider-local `isJunieSessionNotFound` (kept out of the shared helper so no other ACP backend widens its match), validated by a trap test against every `"not found"` string in `pkg/agent` plus the real Junie wording. But the live run also showed a case that is **genuinely undetectable**, and I'd rather document it than pretend it's handled:

- `session/load(<any id>)` returns **success** with only `configOptions` — even for a fabricated id. Load is not a session-existence check, so the load-time branch is dead code for Junie.
- `session/prompt(<bogus id>)` also returns **success** — Junie silently serves a session it never had. No error, so `junie.go` cannot detect it. Added as a documented limitation on the resume path, not a silent assumption.

## What the live run confirmed (protocol surface, real frames)

- `--acp=true` is the correct flag form; `initialize` advertises `loadSession`, `mcpCapabilities {http,sse}`, `sessionCapabilities {fork,list,resume}`, `authMethods {jetbrains-account, junie-cli}`.
- `session/set_model` **is** a real method (not `METHOD_NOT_FOUND`) — it validates against Junie's own catalog and rejects anything else.
- Brave mode is the permission gate; `session/set_config_option {configId: brave_mode}` works, and `junie.go` now sets it for unattended runs.
- Project-local `.junie/skills` **is** loaded — Junie emits a `Skills update: New skills: .../.junie/skills/<name>` notification, which validates the daemon's injection path.
- MCP transport works to `tools/list`; the turn loop streams `agent_message_chunk` and returns `stopReason:end_turn`.
- BYOK bypasses the JetBrains AI subscription — requests route to the configured provider.

## Two new findings from the live frames

1. **Model catalog is in `configOptions`, not a `models` block.** `session/new` returns the model list under `configOptions[id=model].options` (currentValue e.g. `gpt-5-2025-08-07`), with **no** top-level `models`/`currentModelId`. So `extractACPCurrentModelID` returns `""` and usage is filed under `"unknown"`.
2. **A non-catalog model id is fatal.** `session/set_model` on anything outside Junie's catalog returns `-32602 "Unsupported Junie model '<id>'"`, and `junie.go` treats a set_model failure as fatal — so an agent configured with a model string Junie doesn't recognize fails every task before prompting.

I'll address both in a follow-up commit (discover the catalog from `configOptions`; don't hard-fail on an unsupported-model error).

## The honest limitation, and the ownership question

I could **not** complete the turn-gated half of your checklist — a turn producing real output, a permission request, an MCP tool call, skill use, a resume that recalls the prior turn. Reason is squarely my environment, not the protocol: every hosted provider (OpenAI, OpenRouter, Anthropic, Google, xAI) geo-blocks my region, Junie 1468.30 has no custom-endpoint/base-URL support to reach a local gateway (confirmed by decompiling `ByokConfiguration` — it stores only per-provider keys, no `baseUrl`), and Junie's HTTP client ignores JVM proxy settings so I can't route it around the block. A direct `curl` to the same OpenAI endpoint with the same key returns HTTP 200 with a valid completion — so the model path itself works; only Junie's own non-proxyable client blocks the in-Junie turn from here.

Which brings me to your ownership offer, honestly: I can do the code and protocol maintenance — this write-up is the evidence — but I can't reliably do **live-turn verification** from my region. So the right shape is probably the community-maintained one you described, with me as the code owner and the understanding that turn-level checks need someone with provider access. If you'd still like me listed for `server/pkg/agent/junie.go`, I'm glad to take it on those terms; if you'd rather merge it community-maintained with no named owner, that's a perfectly good outcome too. Your call — and thanks for being straight about the tradeoff up front.

ADanMan:
Quick follow-up on the ownership offer: I'm not declining it — I'd like to take the community-maintained ownership of the Junie runtime.

The only open item is the live run, which I couldn't complete from my current network region (provider-side geo-block, and Junie's client ignores app-level proxy settings). I'll retry it over a full-tunnel VPN from a supported region — that routes the CLI's own traffic at the OS level, unlike the SOCKS proxy Junie ignored, so it's a genuinely different attempt rather than a repeat of what already failed. BYOK, so no subscription is involved.

I'll post the raw live-run frames here once I have them (first turn, permission request, MCP transport, skill loading, resumed turn).

multica-eve:
Thank you for the unusually thoughtful work on this PR — especially the negative-control tests, the live protocol captures, and the care you took to report problems in our existing runtimes rather than only fixing your own path.

After the implementation review, we stepped back and did a product and adoption review specifically for Junie. We looked at JetBrains' GA status, public repository activity, release-download signals, and the demand we can actually verify in Multica. Junie clearly has a real early-adopter community, but on the Multica side we could only verify the original request in #1163 (with two upvotes) and no second independent user case. The public download figures are also affected by Junie's automatic updater, so they do not establish a reliable active-user count or enough Multica-specific demand.

A native runtime makes Multica responsible for compatibility, schema changes, documentation, test coverage, and safety behavior indefinitely, even when it is community-maintained. This PR itself surfaced several Junie-specific behaviors around sessions, models, permissions, and global settings, which made that ongoing cost concrete. We therefore do not have enough evidence to justify adding Junie as a built-in runtime at this time, and we are closing the PR for product-scope and maintenance-ROI reasons — not because of the quality of your contribution.

We also owe you an apology: we should have applied this product/adoption gate before asking you to address multiple rounds of implementation feedback. We previously suggested that community ownership would be sufficient; after doing the broader research, we changed that product decision. The inconsistency and the extra work it caused are on us.

We will keep #1163 available as a demand signal. If independent Multica demand becomes clearer, or if we add a generic ACP extension path that does not require every long-tail agent to become a permanent core integration, we can revisit Junie from a better foundation.

Thank you again for the rigor, patience, and candor you brought to the review.

ADanMan:
Follow-up: the two items I flagged for a follow-up commit are done in 624c87de20 on this branch.

1. Usage attribution. Junie 1468.30's session/new advertises the active model as the `model` configOption's currentValue and carries no top-level `models` block, so the shared extractACPCurrentModelID returned "" and token usage was filed under "unknown". Added a junie-local fallback that reads that configOption (the same acpConfigOptionCurrentValue helper the brave_mode gate already uses). extractACPCurrentModelID is left untouched since it's shared with the other ACP backends.

2. set_model is no longer fatal on an unrecognised model. A -32602 "Unsupported Junie model '<x>'" (and any non-dead-resume set_model failure) now logs a warning and continues on Junie's current default model, with usage re-attributed to that model. The resumed-dead-session case stays fatal with ResumeRejected, unchanged.

Both are grounded in a fresh, turn-free session/new capture against a real @jetbrains/junie-cli 1468.30 (initialize -> session/new only, so no authentication and no turn were needed): the model configOption currentValue was gpt-5-2025-08-07, with a 12-model catalog. New tests cover both behaviours (confirmed to fail without the fix), and the full pkg/agent suite passes.

The live-turn checklist items still need provider access from a supported region, as noted earlier.

Issue #1997, opened 2026-05-02
[Bug]: Repeat the response each time on HermesAgent


seepine:
### Deployment type

multica.ai (hosted)

### What happened?

Repeat the response each time on HermesAgent

### Steps to reproduce

1. add hermes runtime
2. chat

### Screenshots (optional)

<img width="547" height="430" alt="Image" src="https://github.com/user-attachments/assets/c6f490a6-6c24-4422-9e9d-7a98187d6b9a" />

### Additional context (optional)

```shell

```

PR #7351, opened 2026-08-20
feat(agent): add ZeroClaw as a native ACP runtime (MUL-6511)

### LINKED ISSUE #1543: [Feature]: Add support for ZeroClaw as a generic agent runtime
### Deployment type

multica.ai (hosted)

### What do you want and why?

Currently, Multica supports specialized coding agents like Claude Code and the (partially integrated) Hermes. However, there is a growing need for a high-performance, generic agent that is easier to manage than OpenClaw. ZeroClaw has emerged as the preferred choice for many developers due to its Rust-based architecture, minimal resource footprint (<5MB RAM), and single-binary deployment.

**Rationale**

- Performance: ZeroClaw offers near-instant cold starts compared to the multi-second latencies seen in Node.js-based alternatives.

- Low Bloat: It provides a full-featured agent environment without the heavy dependencies of OpenClaw, making it ideal for the lightweight "daemon" approach Multica uses.

- Popularity: It is rapidly becoming the standard for self-hosted autonomous infrastructure. Supporting it would make Multica accessible to users running on resource-constrained hardware (e.g., small VPS or edge devices).

**Proposed Implementation**

Add ZeroClaw to the supported agent runtimes in the daemon.

_No response_

_No response_

ADanMan:
## Summary

Adds ZeroClaw (github.com/zeroclaw-labs/zeroclaw) as a native ACP runtime. forrestchang welcomed a PR for this on the issue. ZeroClaw is a Rust agent runtime (<5MB RAM, single static binary) that ships a native ACP server mode via `zeroclaw acp` (JSON-RPC 2.0 over stdio).

## Change

Follows the same registration set as the recent Dim addition (#6675) file-for-file, substituting ZeroClaw's binary/CLI shape:

- `server/pkg/agent/zeroclaw.go`: ACP backend structurally mirroring `grok.go`'s session lifecycle (`initialize` → `session/new`/`session/load` → `set_model` → prompt), reusing the shared `hermesClient` ACP transport, `acpDeliverableTracker`, and tool-naming helpers.
- `agent.go`, `launch.go`, `models.go`, `internal/daemon/{agents_probe.go,config.go,daemon.go}`, `internal/daemon/execenv/runtime_config.go`, `internal/metrics/labels.go`, `scripts/agent-cli-command-names.txt`.
- Frontend: `packages/core/types/agent.ts`, `packages/core/agents/mcp-support.ts`, `packages/core/runtimes/display.ts`, `packages/views/runtimes/components/provider-logo.tsx`.
- Migration 378: widens the `runtime_profile.protocol_family` CHECK constraint (`ALTER TABLE ... NOT VALID`, matching migration 370's `dim` pattern). Numbered 378 rather than 377 to avoid colliding with a same-day, separately-implemented Junie runtime PR that independently claimed 377 off the same `origin/main` tip.

Deliberately scoped out (disclosed in code comments, not hidden): Dim's hardcoded permission-preset injection, its min-version gate, and its `thinking.go` effort-picker entry are all additive/opt-in whitelists, not required-for-correctness gates — ZeroClaw omits them since no real binary was available to verify version-specific quirks against. `initialize`'s advertised `authMethods` is empty, so `authenticate` is correctly skipped, matching the existing plain-ACP-runtime assumption already established for `dim`.

## Test plan

- Applied migration 378 against a full local chain on Postgres — clean.
- `gofmt -l`, `go vet ./...`, `go build ./...` clean.
- Full `go test ./pkg/agent/...` (all 9 new ZeroClaw tests plus no regressions), `./cmd/migrate/...` (including `TestEveryConcurrentUpBuildHasCleanup`/`TestConcurrentIndexCleanupsMatchTheirMigrations` — not applicable here since this migration only alters a CHECK constraint, no concurrent index), `./internal/daemon/execenv/...`, `./internal/metrics/...` — all pass.
- `pnpm --filter @multica/core typecheck`, `pnpm --filter @multica/views typecheck` clean; `pnpm --filter @multica/core exec vitest run agents/mcp-support.test.ts types/ runtimes/display.test.ts` (50/50) and `pnpm --filter @multica/views exec vitest run runtimes/components/provider-logo` (8/8) pass.

Fixes #1543


multica-eve (changes requested):
Thanks for adding ZeroClaw support. The product direction is welcome, but the implementation currently assumes ACP behavior that ZeroClaw does not provide, so this needs changes before it can be merged.

Please address these blockers:

1. `session/set_model` is not implemented by ZeroClaw. Any run with a non-empty model currently fails with `METHOD_NOT_FOUND`. Remove this request or treat the unsupported operation as a non-fatal condition, and replace the fake-based test that currently validates behavior the real runtime does not have.
2. Model discovery assumes `session/new` returns `models.availableModels` and `currentModelId`, but ZeroClaw returns only the session ID and workspace directory. Correct the discovery behavior and comments, and reconsider exposing `MULTICA_ZEROCLAW_MODEL` until there is a supported way to apply it.
3. Resume uses `session/load`, which replays the full session history as notifications. Because those notifications are forwarded into the transcript and deliverable tracker, prior output can be duplicated and included in the new final response. Use `session/resume`, or explicitly suppress replayed notifications until `session/load` completes.
4. `session/new` does not send `agentAlias`. A standalone `zeroclaw acp` process fails when multiple agents are configured and no `[acp].default_agent` is set. Add a configuration path that passes the alias, or clearly document and validate the default-agent prerequisite.
5. The UI advertises MCP support unconditionally, while ZeroClaw ignores the supplied MCP servers unless the selected agent has `acp_enable_mcp = true` (disabled by default). Do not advertise unconditional support; either gate it correctly or document and validate the prerequisite so configuration is not silently discarded.

Please also validate the revised integration against a real `zeroclaw` binary and add tests based on the actual handshake/session behavior. The current fake mirrors the assumed protocol and therefore does not catch these incompatibilities.

ADanMan:
Thanks — all five are real. The integration is now validated against a real zeroclaw v0.8.4 binary driven over stdio, not just against the ACP spec, and that changed the fix for two of them. Fixed in one commit.

1. session/set_model
Confirmed, and worse than "unsupported operation": the call was fatal, so any agent with a non-empty model failed before session/prompt ran. Live: `session/set_model` → `{"code":-32601,"message":"Method not found: session/set_model"}`, with a `session/cancel` on the same session id returning `{}` immediately after as a health control. There is no model knob anywhere on the ACP surface — the dispatch table is initialize and session/{new,load,resume,close,prompt,stop,cancel,event|update}, and no handler reads a model param. So I removed the request rather than taking your second option: tolerating it would be a permanent no-op round-trip.

Removal alone was not enough, though. ModelSelectionSupported("zeroclaw") was still on the `default: true` branch, so making the call non-fatal would only have traded a loud failure for a silent one — the UI would keep offering an override that never takes effect. zeroclaw is now in the opt-out list next to qwenpaw/mcode, and the daemon no longer reads MULTICA_ZEROCLAW_MODEL.

2. Model discovery
Confirmed false. `session/new` returns exactly `{sessionId, workspaceDir}` — no `models`, no `currentModelId`. The comment claiming a `models.availableModels` catalog was wrong, both extractACPCurrentModelID calls were dead code, and discoverZeroclawModels could only ever return an empty fallback while spawning an ACP subprocess with a 15-30s ceiling on every catalog read (and empty/fallback catalogs are not memoized). ListModels now returns an empty catalog directly, following the qwenpaw/mcode precedent, and the helper is deleted.

There is one honest model signal: `initialize._meta.zeroclaw.defaultModel` (live value "llama3.2"). Token usage is labelled from that now, falling back to the prompt result, so a run is not attributed to "unknown". Worth noting ZeroClaw reports no token counts over ACP today — its prompt result is `{sessionId, stopReason, content}` — so the usage map is nil in practice; the attribution is kept wired for when that changes.

3. session/load vs session/resume
Confirmed by A/B against one seeded session, changing only the method: session/load emitted two `session/update` notifications (`user_message_chunk`, then `agent_message_chunk` with the prior answer) before its own response; session/resume emitted none. Source matches — load loops `history_notifications_for_message` over the restored messages, resume seeds the agent without replaying and only re-emits a durable TodoWrite plan.

I treated your "or" as an and, and did both. Resume now uses session/resume, and the backend gained the `acceptNotification` turn gate every other ACP backend here carries (hermes/grok/kimi/kiro/mcode/qoder/reasonix/traecli all set it; zeroclaw was the only one without). The gate matters independently of the method switch: the replay frames arrive *before* the session/load response, so any suppression has to be armed from process start, and chunks can also be flushed before the prompt response streams — the case #1997 was filed for. The blast radius was slightly wider than duplicated text: replayed historical `tool_call` frames reset the deliverable boundary, so a turn ending on a tool call could emit prior-session text as the answer, and the polluted full-text buffer feeds provider-error detection. New test asserts a resumed run's output contains only the current answer; reverting the gate makes it fail.

4. agentAlias
Confirmed, with a scope refinement worth stating: it is not universal. Precedence is `agentAlias` param → connection default (`?agent=`, gateway transport only, unreachable over stdio) → `[acp].default_agent` → auto-select when exactly one `[agents.<alias>]` entry exists. So this fires only when the config holds 0 or 2+ agents and `[acp].default_agent` is unset. Live: 1 agent aliasless succeeds; 2 agents aliasless → `-32602 session/new requires agentAlias`; 2 agents with the alias succeeds.

Plumbing note: the alias cannot ride on argv. `zeroclaw acp` accepts only --config-dir/--max-sessions/--log-level/--session-timeout/-v, and `--agent` kills the process at argument parsing — so custom_args could not have been used as-is either. The backend now consumes `--agent` / `--agent-alias` (space and `=` forms) out of custom_args and sends the value as the session/new param; both flags are also blocklisted so no copy can reach argv through a launch prefix. It is omitted when unset — sending a hardcoded `"default"` breaks the single-agent install that works today (`Unknown agent \`default\``), which is now a regression test. Not sent on resume: the restore path rebinds from the persisted session owner and deliberately ignores transport-supplied defaults. The -32602 refusal now carries a remediation naming both `--agent <alias>` and `[acp].default_agent`.

5. MCP — agree on the outcome, disagree on the mechanism
Your conclusion is right and I implemented it, but `acp_enable_mcp` is not the gate, so neither of your proposed remedies would have worked. Three-arm live test, sending a client MCP entry whose command touches a marker file:
- acp_enable_mcp=false → session/new returns instantly, marker absent.
- acp_enable_mcp=true (your stated prerequisite satisfied) → still instant, marker still absent.
- acp_enable_mcp=true plus a server-side grant → session/new blocks on MCP connect and ZeroClaw's *own* configured server spawns; the client's still does not.

Arm 2 is the one that redirects the fix. `handle_session_new` / `handle_session_load` / `handle_session_resume` never read `params.mcpServers` — the flag initialises that agent's own `mcp_bundles`, not the client list — so gating on it would have told a user their MCP config was active while ZeroClaw kept discarding it. That is a worse end state than shipping as-is. Instead: `zeroclaw` is out of MCP_SUPPORTED_PROVIDERS (the MCP tab is hidden), `mcpServers` is always sent empty, and a stale saved mcp_config logs a warning naming ZeroClaw's config-dir rather than failing the task. Also worth knowing: the real binary advertises `mcpCapabilities {http:false,sse:false}`, so http/sse entries were already being dropped client-side — the config was being discarded twice with no user-visible signal.

Also fixed, outside your five: ZeroClaw's SESSION_NOT_FOUND is a custom -32000, which `isACPSessionNotFound` did not match, so `resumeRejected` never fired and a stale session id would be retried forever instead of restarting fresh. -32000 is now accepted; the wording check still carries the decision, so a transient failure under the same code stays non-rejecting (there is a test for exactly that).

Test fixtures were part of the problem — the fake acknowledged set_model, advertised `mcpCapabilities {http:true,sse:true}`, and used -32602 for an unknown session, i.e. the inverse of the runtime in three places. It now reproduces the observed frames, so `session/set_model` falls through to the same -32601 the binary returns.

PR #7354, opened 2026-08-20
MUL-6510: fix(daemon): isolate task environments and retry startup failures (ISE-2165)

iServU:
## Summary

- Fix ISE-2165 task-environment collisions by deriving each environment root from the full task UUID instead of its first 8 hex characters.
- Treat Codex initialization process failures as retry-safe after confirmed cleanup unless the parent task context ended.
- Add regression coverage for same-bucket UUIDv7 task IDs mapping to distinct roots.

## Diagnosis

The distinguishing observation for this fix is that two task IDs created 2.6 seconds apart shared the same 8-character root, while `Prepare` defensively called `RemoveAll` on that root. UUIDv7 prefixes therefore collide for roughly 65.5 seconds, allowing one task to delete another task's `CODEX_HOME` and SQLite state. The prior retry gate also required `timedOut`, so outright initialization process failures were terminal despite `max_attempts: 2`.

## Validation

- `git diff --check` passes.
- Targeted Go tests were not runnable in the publishing workdir because the Go toolchain is unavailable (`go: command not found`).
- The patch was applied cleanly to the current upstream `main` checkout.

Issue: ISE-2165

### DIFF
--- a/apps/web/app/[workspaceSlug]/layout.test.tsx
+++ b/apps/web/app/[workspaceSlug]/layout.test.tsx
@@ -10,13 +10,16 @@ const state = vi.hoisted(() => ({
   } as { id: string; onboarded_at: string | null } | null,
   isAuthLoading: false,
   workspace: null as { id: string; slug: string } | null,
+  workspacesBySlug: {} as Record<string, { id: string; slug: string } | null>,
   workspaceError: false,
   hasBeenSeen: false,
+  pathname: "/acme/issues",
   replace: vi.fn(),
 }));
 
 vi.mock("next/navigation", () => ({
   useRouter: () => ({ replace: state.replace }),
+  usePathname: () => state.pathname,
 }));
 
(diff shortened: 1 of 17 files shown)


multica-eve (changes requested):
Request changes: the pre-launch retry does not share the existing two-attempt budget.

`server/pkg/agent/codex.go:874` can create a session as attempt 2 after a process-start failure, but the result loop at `server/pkg/agent/codex.go:887` resets its counter to attempt 1. If that second launch then returns an initialize- or model-catalog failure marked safe to retry, the loop starts a third process and logs attempt 2 twice. This violates the bounded-retry contract and can extend failed tasks under host pressure.

Please carry the real initial attempt into the result loop (or unify both retry paths under one counter), stop once attempt 2 has been consumed, and add a regression test covering: process-start failure → retry-safe initialize failure → exactly two total starts.

PR #7357, opened 2026-08-20
MUL-6527 feat(issues): add independent comment branch runs

### LINKED ISSUE #7033: Protect collaborative edits and support independent runs from comment history

yyclaw:
## What does this PR do?

Adds independent agent runs from eligible issue comments. A branch run receives a server-owned, frozen snapshot of the issue plus only the ancestry from the thread root to the selected comment. Siblings, descendants, and later unrelated discussion are excluded, so the original issue conversation remains intact while the historical direction can be explored independently.

The human-only branch endpoint uses UUID idempotency keys and the selected comment content as a baseline. It routes to an explicitly selected agent when provided; otherwise it reuses the selected comment's source agent, then falls back to the issue's assigned agent or squad leader. Target selection, invocation permission, runtime readiness, daemon capability, snapshot capture, and task creation are revalidated under transaction locks. Request-dedup locks and issue/agent queue locks use disjoint advisory-lock namespaces, while busy issue/agent pairs retain branch work in a dedicated deferred queue without changing ordinary mainline scheduling or generic rerun behavior.

The daemon builds branch prompts exclusively from the frozen snapshot and instructs the agent to publish one new top-level result comment. The task records the exact branch point and source task, allowing web/desktop and mobile execution logs and result comments to navigate to the exact source comment. Editing or deleting the source comment does not mutate or cancel an already frozen branch run; deleted sources simply stop exposing dead navigation links.

Capability negotiation keeps the action hidden against old servers and rejects runtimes whose daemon does not advertise `comment_branch_v1`. The database rollout is additive through migrations 390-392, with nullable task metadata and independently built concurrent indexes for idempotency and deferred-queue promotion.

## Related Issue

Closes #7033

Builds on the collaborative edit protection merged in #7038. This is the second and final PR in the two-PR delivery plan documented in #7033.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added nullable comment-branch provenance, frozen context, and request identity fields to agent tasks in migration 390.
- Added separate concurrent unique/deferred indexes in migrations 391 and 392, preserving mixed-version inserts and rollback of legacy rows.
- Added a workspace-scoped, human-only comment branch endpoint with membership/invocation checks, exact content baselines, UUID idempotency, disjoint advisory-lock namespaces, and bounded frozen ancestry snapshots.
- Added automatic target resolution through the source task, issue assignee, or squad leader, with transaction-time revalidation when concurrent state changes.
- Added a deferred branch queue that serializes an issue/agent pair while preserving existing ordinary task, manual rerun, cancellation, completion, and retry rules.
- Prevented generic rerun and automatic retry paths from reinterpreting a frozen comment branch.
- Added daemon/server capability negotiation and branch-specific prompt/result instructions that exclude live sibling and descendant context.
- Added task provenance to protocol/API types and shared helpers for exact branch-point lookup.
- Added web/desktop and mobile branch actions, queued/run labels, execution-log affordances, and exact source-comment focus/highlight behavior.
- Made comment deletion cancel ordinary dependent tasks atomically while preserving independent frozen branch tasks and rolling cancellation back if deletion fails.
- Added backend race/concurrency matrices, migration up/down and mixed-version coverage, daemon prompt tests, core request-state tests, and web/mobile navigation tests.

## Behavioral Boundaries

- A branch snapshot contains the issue state captured at creation and only the root-to-selected-comment path.
- The result is a new top-level comment linked by task provenance; it is not a deeper nested reply and is not merged back automatically.
- Source edits or deletion do not alter an existing frozen run. If the source no longer exists, clients omit the navigation action.
- This PR does not add issue dependency gates, general task relationships, revision history/rollback, protected specifications, or a general rerun scheduler redesign.

## Rollout and Compatibility

- Migrations 390-392 are additive. Existing task rows keep null branch metadata, and old writers can omit the new columns.
- Each concurrent index build remains in its own single-statement migration.
- Old clients continue using existing issue/comment behavior. New clients default the feature off when the server capability is absent or malformed.
- A new server accepts branch creation only for a selected runtime whose daemon advertises `comment_branch_v1`.
- Rolling back migrations removes branch metadata and branch indexes while preserving ordinary task rows. Branch-specific metadata is not expected to survive that rollback.

## How to Test

1. Run `make migrate-up`, `make sqlc`, and the migration lint/concurrent-index checks.
2. From `server/`, run the focused race suites for `./internal/handler`, `./internal/daemon`, `./internal/daemon/execenv`, `./internal/service`, and `./internal/migrations` covering branch routing, snapshot isolation, idempotency, queue interleavings, prompt construction, deletion rollback, retries, and migration up/down behavior.
3. Run `pnpm typecheck` and `pnpm --filter @multica/mobile typecheck`.
4. Run `pnpm lint` and `pnpm --filter @multica/mobile lint`.
5. Run `pnpm test` for the core, view, and mobile unit suites.
6. In web/desktop and mobile, branch from a root comment and a nested reply; verify the run excludes siblings/later discussion and its result appears as a new top-level comment.
7. Start a branch while the same issue/agent pair is busy; verify it remains queued and starts after the active work without cancelling or replacing ordinary work.
8. Retry the same ambiguous request with the same idempotency key, then reuse that key with different intent; verify the first request is deduplicated and the second is rejected. Also verify request UUIDs cannot collide with issue/agent queue lock identities.
9. Navigate from the execution log and result comment to a nested source; verify its thread expands and the exact comment is focused. Delete the source and verify the frozen run survives while dead navigation is removed.
10. Connect an old or incapable daemon; verify clients hide the action when server capability is unavailable and the server rejects a runtime that lacks `comment_branch_v1`.

### Local verification note

All targeted Go race suites, migration up/down checks, TypeScript typechecks, lint commands, and core/view/mobile unit suites pass locally. The final Behavior Inventory covers 93 changed files and 26 obligations against the latest `upstream/main`. A broad parallel Go package run was not used as the acceptance oracle because packages share the worktree database and contaminated an unrelated service test; the repository-recorded PR2 suites pass sequentially, and CI remains the canonical full-pipeline check.

## Checklist

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:**

Used Codex to reconcile the implementation against #7033, merged PR #7038, and the latest upstream base; preserve existing mainline scheduling semantics; model branch idempotency and issue/agent concurrency interleavings; implement frozen-context execution across server/daemon/clients; add migration and product regression coverage; and run iterative review-loop, behavior-inventory, and preflight checks before preparing the PR.

## Screenshots (optional)

Not included. Branch actions, queued/run states, provenance navigation, thread expansion, exact focus, and deleted-source behavior are covered by component tests.


Bohan-J (changes requested):
## Product assessment first

**Before:** once a discussion has drifted, going back to "the point where it was still on track" isn't possible. You either open a new issue or post another comment — and the agent then sees the entire thread (drift included) and continues from its current session. Keeping the original line while exploring a different one means copy-pasting by hand.

**After:** any comment gets a "start an independent run" action. The issue state at that moment, plus only the root→selected-comment path, is frozen and used to start a fresh run. Siblings, descendants and all later discussion are excluded, and the original conversation is untouched. The result lands as a new top-level comment with an affordance to jump back to the comment it started from. If the target agent is busy the run waits in line instead of interrupting or displacing ordinary work. Editing or deleting the source afterwards does not affect an already-frozen run — deletion just stops rendering the navigation link.

**Verdict: this is worth accepting.** #7033 is a real pain point, the non-goals (no dependency gates, no general task relationships, no revision history/rollback, no scheduler redesign) are drawn with commendable restraint, and the "frozen snapshot + single ancestry path + result as a new top-level comment" model is self-consistent and doesn't pollute the original thread.

Two product questions worth settling before merge (neither blocks the direction):

1. **Queueing can be indefinite.** Branch runs always yield to ordinary work — deliberate, per "does not change mainline scheduling" — but on a continuously busy issue a branch may wait a long time. The UI only shows `Queued · #N`: no estimate, no timeout, no proactive signal. Is a cap or a nudge wanted?
2. **The result is a new top-level comment and never flows back.** Explicitly a non-goal and fine as such, but the user ends up with two parallel lines and reconciles them entirely by hand. Confirming this is the intended v1 shape.

---

## What I verified locally

| Check | Result |
| --- | --- |
| `go build ./...`, `go vet ./internal/{handler,service,daemon/...}` | ✅ |
| `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` | ✅ |
| `go test ./internal/daemon/ -run CommentBranch -count=1` | ✅ |
| `go test ./internal/handler/ -run CommentBranch -count=1` | ✅ (**requires migrations 390–392 applied first**, otherwise the whole set fails) |
| `go test ./internal/service/ ./internal/migrations/` | ✅ |
| `pnpm typecheck`, `pnpm test`, `pnpm --filter @multica/mobile typecheck` | ✅ (views 394 + core 134 + desktop 54 + web 30 + docs 4) |

On the shared-database note in the PR description: I independently reproduced the same phenomenon — one flaky `internal/service` failure (not reproducible across three re-runs) and one `TestPluginRoutesRequireWorkspaceAdmin` failure from a leftover `user_email_key` collision in the shared DB. **Neither is related to this PR**, and running the suites sequentially with CI as the canonical check is the right call. Worth noting the new `comment_branch_migration_test.go` isolates itself with a dedicated schema and `search_path` — it is not the contamination source.

---

## Blocking

### 1. A branch run changes the runtime brief bytes for squad-leader agents (MUL-5377 / MUL-5811)

`server/internal/handler/daemon.go:1928` forces `resp.IsLeaderTask = false` for branch tasks, and `server/internal/daemon/prompt.go:777-780` hard-returns false on the daemon side as well. `IsSquadLeader` then flows into `TaskContextForEnv` and gates four brief sections in `runtime_config_sections.go`: the Squad maintenance block, the wording of workflow step 4, the leader-specific status bullet, and the Output section wording.

Net effect: **for the same (agent, issue, provider), an ordinary run renders one brief and a branch run renders a different one** — so the two invalidate each other's cached prefix. MUL-5811 records owner acceptance for jitter caused by the *role itself* flipping; here the flip is driven by **trigger type**, which is exactly the shape the prompt-prefix guardrail is meant to stop.

To be fair about the blast radius — this is **narrow and not a correctness bug**:

- Only agents that actually lead a squad are affected. Every other agent's brief bytes are unchanged.
- `BranchContext` only reaches `renderIssueContext` (which writes `issue_context.md`, a workdir file that already varies per run). It does **not** reach the brief rendering path — I checked, that part is clean.
- The existing `ByteIdentical` tests **pass**, but they don't cover this axis, so green here does not mean the invariant is held.

Please confirm this is a knowingly accepted cost, and add a `ByteIdentical` case in `execenv` pinning "branch run vs ordinary run for a leader agent". Alternatively, if the role shouldn't flip: keep the leader role on branch runs and express the delivery difference through the per-turn prompt only.

### 2. `NotifyTaskFinished` now opens a transaction on every terminal task transition

`server/internal/service/task.go:6054`:

```go
func (s *TaskService) NotifyTaskFinished(ctx context.Context, task db.AgentTaskQueue) {
	if task.IssueID.Valid && task.AgentID.Valid {
		if _, err := s.PromoteDeferredCommentBranch(ctx, task.IssueID, task.AgentID); err != nil {
```

`PromoteDeferredCommentBranch` does `Begin` → advisory lock → savepoint → promote → `Commit`. That means **every task completion / failure / cancellation platform-wide** — including the `CompleteTask` and `failTask` request paths and the `notifyTasksFinished` bulk-cancel loop — pays 4–5 extra round trips even when no deferred branch exists anywhere. A bulk cancel spanning N distinct (issue, agent) pairs becomes N transactions.

This function used to be a purely in-memory best-effort notification (its signature didn't even take a `ctx`); it is now a synchronous DB write. Please gate it behind a cheap check first. `HasDeferredCommentBranchForIssueAndAgent` already exists in this PR and is backed by the new `idx_agent_task_deferred_branch` partial index, so it's a direct fit:

```go
if has, err := s.Queries.HasDeferredCommentBranchForIssueAndAgent(ctx, ...); err == nil && !has {
    s.notifyRuntimeMayHaveWork(task.RuntimeID, "")
    return
}
```

Related side effect worth a comment in the code: promotion now runs on the caller's request context, so a disconnected client silently skips it. The 30s sweeper recovers it, but that dependency should be stated rather than implied.

### 3. Two misplaced doc comments

- `server/internal/daemon/execenv/context.go:1044` — the `renderQuickCreateContext` doc comment now sits on `renderCommentBranchContext` (line 1049). A copy of it was re-added at line 1089, but a blank line separates it from `renderQuickCreateContext`, so it is no longer a doc comment at all.
- `server/internal/daemon/prompt.go:168` — identical pattern: the `buildQuickCreatePrompt` doc comment now sits on the `commentBranchPromptSnapshot` type (line 175), and the copy at line 252 is detached by a blank line.

Both are artifacts of inserting the new declarations between a comment and the function it documented. Harmless at runtime, but each one now describes the wrong thing.

---

## Nits

1. **Migration numbering leaves a 13-slot gap.** `main` is at `376_agent_task_durable_work_dir`; this PR jumps to 390–392. The repo already has a 372–374 gap so it isn't fatal, but a jump this large reads like the numbers were picked against a different base. Suggest 377–379.
2. In `server/cmd/migrate/main.go`, the `concurrentIndexCleanups` entries for 391 and 392 are inserted before 363 and 364, breaking the numeric ordering of that table. (Map order is irrelevant; readability isn't.)
3. **Off-by-one on the depth bound.** `ListCommentBranchPathForUpdate` with `max_depth = 128` returns up to 129 rows (depth 0..128), while the handler rejects at `len(path) > 128`. So an ancestry that is exactly 129 deep **and complete** is rejected as "too large". Separately, a broken ancestry chain (`path[0].ParentID.Valid`) returns `404 comment not found`, which is misleading during triage — a 409/500 would describe it better.
4. `useStatusLabel` (`packages/views/issues/components/task-run-labels.ts:57`) maps **every** `deferred` task to `branch.status_deferred`. But `ListTasksByIssue` returns all statuses, and `deferred` is also used for channel-media pending tasks and armed auto-retries. The copy itself is generic ("Queued") so it's harmless on screen — and it's an improvement over the previous bare `deferred` string — but living under the `branch.*` namespace will mislead the next reader.
5. `useCommentBranchAction` runs unconditionally in every `CommentRow`, and `selectableAgents` / `agentSelectItems` are recomputed on each render without `useMemo`. That's O(comments × agents) per render on a long thread.
6. The structured server errors (`branch_target_changed`, `idempotency_key_conflict`, `daemon_capability_unsupported`) reach the user through `error.message` in a toast — i.e. raw English strings. Only `revision_conflict` is localized. These are paths users will actually hit; they belong in the locale files too.
7. `/api/capabilities` is a global endpoint but is cached under `issueKeys.capabilities(wsId)`, so switching workspace re-fetches for nothing.
8. Two spellings in play: the daemon capability string is `comment-branch-v1` (hyphens), the server capability field is `comment_branch_v1` (underscores). Each is internally consistent, but the PR description says runtimes are rejected unless they advertise `comment_branch_v1`, which reads like a mismatch. Worth unifying, or at least noting in the `protocol/messages.go` comment.

---

## Things done well

- Migrations 390–392 are fully additive, carry no foreign keys, keep each `CREATE INDEX CONCURRENTLY` in its own single-statement migration, and ship complete `down` files.
- Putting request-dedup locks in a `"request:" + uuid` namespace so a client-supplied UUID can't collide with an agent UUID in the issue/agent queue lock — that's a subtle trap, and the comment explaining it is excellent.
- Deliberately *not* pre-locking the agent row, with the runtime-merge lock-order inversion spelled out as the reason.
- `DeleteComment` moving from "cancel, delete, compensate on failure" to a single transaction — and deleting the `retriggerCancelledTaskSurvivors` compensation path in the process — is a net simplification.
- `PromoteNextDeferredCommentBranch` catching 23505 and treating it as "not promoted" shows the semantics of `idx_one_pending_task_per_issue_agent_v2` were genuinely understood.
- Deferred branch rows carry a NULL `fire_at` (so `PromoteDueDeferredTasksForRuntime` ignores them) and a NULL `retry_of_task_id` (so `CancelSupersededDeferredRetriesForRuntimes` ignores them). Both existing sweeper interactions land correctly.

---

## Future impact

- **Branch starvation** — see product question 1. Beyond the UI, there is currently no metric or log that would reveal how long a branch has been queued on a given issue/agent pair.
- **`ListDeferredCommentBranchPairs` truncates silently at `LIMIT 128`.** Beyond 128 promotable pairs the remainder waits for the next tick with nothing logged. Not a problem at current scale, but a silent cap deserves a warn line.
- **Lock ordering.** The branch-creation path takes workspace (KEY SHARE) → issue (FOR UPDATE) → comments (root→leaf, FOR UPDATE) → advisory → fence (agent/issue/runtime KEY SHARE), whereas the fence's own fixed order is agent → issue. So the branch path is issue-before-agent while the fence is agent-before-issue. The agent row gets an explicit "don't pre-lock" rationale; the issue-row `FOR UPDATE` doesn't. This path reuses the existing `LockIssueForDescriptionUpdate` (collaborative edit protection already locks this way), so the risk is level with the status quo rather than new — but please confirm it's a considered choice.
- **No screenshots.** The UI surface is non-trivial (comment menu, confirm dialog, branch marker and queue position in the execution log, jump-and-highlight). Component tests do cover the behavior, but two screenshots would make review meaningfully easier.

---

## Recommendation

**Request changes.** The product direction is right and the implementation is solid — the density of thought around concurrency, idempotency, migrations and capability negotiation is well above average, and the tests genuinely run. Only three items block:

1. Squad-leader branch runs changing brief bytes — needs an owner decision on the MUL-5377 invariant plus a `ByteIdentical` case pinning that axis. **This is the only item requiring a judgement call.**
2. A cheap guard in front of `NotifyTaskFinished` so every task teardown platform-wide doesn't open an extra transaction.
3. The two misplaced doc comments.

Nits 3 (depth off-by-one and the 404 semantics) and 6 (unlocalized error copy) are worth folding into the same pass; the rest can follow later.

yyclaw:
@Bohan-J I dug further into the squad-leader question from your blocking item. I’ve ruled out treating a squad as a mere shortcut to run its leader as a direct agent: the product should either support a real squad target or remain agent-only.

A real squad target has prerequisites beyond preserving the leader brief:

- The request and task must preserve squad identity and run with is_leader_task plus squad_id, rather than resolving the squad to an agent ID and discarding the role.
- RecordSquadLeaderEvaluation currently authorizes against the Issue current assignee squad, not task.squad_id. The existing comment @squad path already permits any workspace squad, so a guest leader cannot record action, no_action, or failed. Issue reassignment also deliberately does not cancel an in-flight task, which makes this a deterministic 400/403 or misattribution path during execution. This must be changed to task-scoped authorization. It is load-bearing for no_action because completion uses that activity record to suppress a synthesized comment.
- A real squad branch also needs explicit branch-specific outcome and delegation semantics: no_action should remain a visible structured terminal result without a placeholder comment, while any delegated worker tasks must preserve the frozen branch context and lineage instead of falling back to the live Issue timeline.

I see three viable scope choices:

1. Expand this PR and implement the general evaluation fix plus the complete real-squad branch semantics here.
2. Land a small prerequisite PR that fixes the historical task.squad_id evaluation problem first, then rebase this PR and add the real squad target.
3. Keep this PR agent-only, remove the implicit squad-to-leader automatic path, and handle the general evaluation fix plus real squad targets in a later dedicated PR.

Do you have a preference among these three for reviewability and scope?

yyclaw:
@Bohan-J Your review prompted me to step back and reconsider the product model behind this PR rather than continue addressing the current implementation in place.

I am pausing work on this PR for now. In particular, I am re-evaluating whether the right product shape is an independent run from a historical comment, or a new sub-issue created from that point with a new user instruction plus frozen issue and comment-ancestry context.

I want to settle that lifecycle and interaction model before making further code changes. Once the direction is clear, I will decide whether to resume or reopen this PR with a revised scope, or replace it with a new PR built around the new design. No review action is needed in the meantime.

yyclaw:
Thanks for the detailed review. Based on that feedback and a subsequent product-design review, I am closing this draft rather than continuing the independent comment-run model.

The replacement direction is #7415: create a real sub-issue from a historical comment, require a new task instruction, and attach the source issue body plus the root-to-selected-comment path as immutable historical context. The child then uses normal Issue assignment, queue, session, comment, and squad semantics.

That changes the core product and data model rather than merely revising this patch, so I do not plan to adapt #7357 in place. If implementation proceeds, it will start from the latest main against #7415.

Issue #7135, opened 2026-08-18
[Feature]: Add /new and /clear conversation controls for external channels


yyclaw:
### Deployment type

Official App

The affected channel engine is shared with self-hosted deployments, so the resulting behavior should remain compatible there as well.

### Area

Other / not sure

This spans the shared channel engine and all supported channel adapters.

### What do you want and why?

External channels need two distinct conversation controls with product-level names:

1. `/new` creates a new Multica Chat and routes subsequent messages from the same external conversation to it.
2. `/clear` resets the agent-visible context while staying in the current Multica Chat.

This naming is deliberate. From an IM client, both operations make the agent forget while the user keeps talking in the same DM or thread. `/new` should represent the larger product action—starting a new Chat—while losing the prior context follows naturally. `/clear` names the narrower reset-in-place operation and preserves the durable context-boundary capability delivered by PR 1.

The original proposal used `/new` for reset-in-place and `/chat` for creating a Chat to avoid changing an established command. After review, the cleaner long-term contract was preferred because existing channel users still observe the essential outcome of `/new`: the agent starts fresh. The additional first-party effect is beneficial—topics appear as separate Chats instead of one ever-growing Chat.

#### `/clear`: reset context in the current Chat

| Command | Multica Chat | Stored/UI history | Agent-visible context |
| --- | --- | --- | --- |
| `/clear` | Unchanged | Preserved | The next real message starts a new context generation |
| `/clear <message>` | Unchanged | Preserved | `<message>` is the first message in a new context generation |

Requirements:

- Do not create a new `chat_session`.
- Preserve the complete Chat history for first-party clients.
- Do not resume the previous provider session for the first post-boundary task.
- Prevent that task and its descendants from retrieving transcript or live channel messages from before the boundary.
- Snapshot the context generation on each task so a later `/clear` cannot change queued or running work.
- Treat `/clear` as a debounce and input-batch barrier.
- Keep retries in the original task's context generation.
- Allow later messages in the same generation to resume the provider session created after `/clear`.

#### `/new`: create a new Multica Chat

| Command | Result |
| --- | --- |
| `/new` | Create a new empty Multica Chat and route subsequent external messages to it |
| `/new <message>` | Atomically create a new Multica Chat and use `<message>` as its first user turn |

Requirements:

- Create a new `chat_session`; do not reuse or repurpose the old one.
- Keep the previous Chat stored, visible, and usable from Web, Desktop, and Mobile; only its external active route changes.
[...]

PR #7362, opened 2026-08-21
MUL-6516 fix(channels): enforce durable context boundaries for /new

yyclaw:
## What does this PR do?

This PR fixes `/new` so it establishes a durable agent-visible context boundary while preserving the existing Multica Chat and its complete first-party transcript.

Previously, `/new` only set `force_fresh_session`, which prevented the next task from resuming the previous provider session. The new task could still retrieve pre-boundary messages through stored Chat history or Slack history APIs.

This change introduces a durable context generation within the existing Chat. Messages and tasks snapshot their generation, and all agent-visible continuity paths are scoped to that snapshot. First-party Chat readers remain unfiltered, so Web, Desktop, and Mobile continue to display the complete conversation.

This is PR 1 of the two-PR plan described in #7135. The separate `/chat` command and Chat route rotation remain intentionally out of scope.

## Related Issue

Refs #7135

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added durable channel context generations and snapshotted the generation on channel messages and tasks.
- Made bare `/new` and `/new <message>` create explicit context boundaries without creating a new `chat_session`.
- Made context generations independent debounce and input-sealing boundaries.
- Scoped stored transcript history, Slack history, provider-session resume, rollout-gap lookup, media state, and retries to the task's context generation.
- Added crash recovery for durable unowned input whose in-memory debounce timer was lost.
- Preserved legacy behavior for tasks and messages created before the migration.
- Aligned migration compatibility with the supported Helm `Recreate` deployment and adjacent-release rollback contract; removed database triggers that guessed channel identity for old writers.
- Limited legacy task backfill to channel-owned `queued`, `dispatched`, `running`, `waiting_local_directory`, and `deferred` tasks; terminal history and Direct Chat tasks remain on the legacy NULL fallback.
- Persisted every Slack outbound chunk ID on its owning assistant task, scoped by installation and real channel, so late old-generation replies remain visible to people but cannot enter a newer agent context.
- Added explicit cleanup for context generations during Chat, workspace, installation, and system-runtime teardown.
- Added migration, concurrency, rollback, recovery, history-isolation, and cross-channel regression coverage.

## Deployment and rollback

The repository Helm chart uses `strategy.type: Recreate`, and the container applies migrations before starting the server. This PR therefore does not support arbitrary old/new backend overlap. New columns are nullable and generation 1 retains a legacy NULL fallback so the adjacent release can be rolled back without requiring old-writer inference triggers.

## How to Test

1. Run the durable context migration test:

   ```bash
   cd server
   go test ./internal/migrations \
     -run '^TestChannelChatContextGenerationMigrationsUpDownAndLegacyRows$' \
     -count=1
   ```

2. Run the shared channel engine with the race detector:

   ```bash
   cd server
   go test -race ./internal/integrations/channel/engine -count=1
   ```

3. Run the task, history, provider-resume, cleanup, and Slack boundary regression tests:

   ```bash
   cd server

   go test ./internal/handler \
     -run '^(TestChannelNewCommandE2EStartsFreshProviderSession|TestClaimTask_ChannelContextRevisionScopesProviderResume|TestGetChatHistory_ChannelTaskCannotReadEarlierContextGeneration|TestDeleteChatSession_PrunesChannelRows)$' \
     -count=1

   go test ./internal/service \
     -run '^TestEnqueueChannelChatTask_ContextClearFailureRollsBackTaskAndSeal$' \
     -count=1

   go test ./internal/integrations/slack \
     -run '^(TestChannelOverviewScopesAndSanitizesFreshGeneration|TestChannelOverviewExcludesLateReplyFromPreviousGeneration|TestChannelOverviewFailsClosedForUnknownOwnBotReply|TestChannelOverviewKeepsPagingCursorWhenLateRepliesFillPage|TestThreadPendingFreshBoundaryReturnsEmptyWithThreadID|TestOutbound_PostsSealedChannelTaskReply)$' \
     -count=1

   go test ./internal/integrations/dingtalk \
     -run '^(TestDeleteWorkspaceCleansRetiredChatContexts|TestDeleteChannelInstallationsBySystemRuntimeAgentsCleansRetiredChatContexts)$' \
     -count=1
   ```

4. Verify generated SQL and static checks:

   ```bash
   make sqlc
   cd server
   go vet ./...
   ```

### Test-suite note

The affected packages and all targeted regression tests pass, including the channel engine under the race detector.

A local full `make test` race run passed every affected package and reported one unrelated timeout in `TestProbeDshMulticaProfile/compatible` at its fixed 5-second deadline. The exact test passed immediately when rerun independently with the race detector (2.75 seconds total; 0.95 seconds for the compatible case).

## Checklist

## AI Disclosure

**Prompt / approach:**

Implemented PR 1 of #7135 by tracing every path through which an agent could recover pre-`/new` context: persisted Chat history, external channel history, provider-session pointers, queued tasks, retries, debounce windows, media ownership, crash recovery, and cleanup.

The resulting change was reviewed with an alternating preflight and code-review loop covering migration safety, the documented Recreate deployment and adjacent-release rollback contract, concurrency interleavings, rollback, task-scoped authorization, cross-channel consistency, and regression tests.

## Screenshots (optional)

Not applicable. This PR preserves the existing first-party Chat UI and changes server-side context isolation only.


multica-eve (changes requested):
Before this change, `/new` starts a fresh agent conversation but can still let the agent retrieve messages from the earlier topic. After this change, members should be able to start an isolated topic in the same Multica Chat: the agent sees only the new topic while people can still review the complete conversation.

Product decision: ✅ Accept the product direction. It matches the intended first phase of #7135, but the implementation currently has four correctness and isolation blockers.

1. **The new database lock order can deadlock message append against task enqueue.**

   `CreateTaskFromChannelMessages` first locks the generation (`server/internal/service/task.go:1704`) and later updates the chat binding (`server/internal/service/task.go:1786`). A concurrent append enters `LockChannelChatContextGeneration` from `server/internal/integrations/channel/engine/session.go:372`, whose query at `server/pkg/db/queries/channel.sql:590` locks the binding and then the generation. That gives the two paths opposite lock orders: generation → binding versus binding → generation. These chat-session lock modes are intentionally compatible, so the overlap is reachable.

   I reproduced this deterministically against local PostgreSQL: PostgreSQL returned `deadlock detected` (`SQLSTATE 40P01`), rolled back the enqueue transaction, and committed the append. This can leave generation messages without a task until a later inbound message happens to recover them. Please use one lock order everywhere and add a database concurrency regression covering the first post-`/new` enqueue racing the next append.

2. **Queued/running tasks that predate the migration do not preserve their generation snapshot.**

   Migration 377 adds nullable `channel_context_revision` (`server/migrations/377_channel_chat_context_generation.up.sql:7`) but does not backfill active channel tasks. The mixed-version guard in migration 379 applies only to future inserts (`server/migrations/379_channel_context_mixed_version_guard.up.sql:59`). For an existing task whose revision is still NULL, `server/internal/handler/chat_history.go:274` does not load a generation; `chatMessageHistory` then falls back to an unfiltered transcript (`server/internal/handler/chat_history.go:92`), and the daemon claim path can use the Chat-wide resume pointer (`server/internal/handler/daemon.go:2605`). This violates #7135's requirement that already queued/running tasks keep their generation snapshots.

   The current legacy-task test manually stamps revision 1, so it does not exercise a genuinely NULL migrated task. Please migrate active channel tasks safely to generation 1 (or fail closed where a snapshot cannot be established), and add a regression with a NULL old task followed by `/new`.

3. **Recovery can run an old generation under the new sender's identity and connected apps.**

   `ListChannelContextRevisionsNeedingTask` returns revisions without the original sender (`server/pkg/db/queries/chat.sql:590`). The recovery loop then schedules every older revision using the current message's `identity.UserID` and metadata (`server/internal/integrations/channel/engine/router.go:572`). Task creation uses that identity for attribution and the connected-app MCP overlay (`server/internal/service/task.go:1650`). For example, if Alice's old message loses its timer and Bob later sends `/new <message>`, the recovered generation can process Alice's input as Bob, with Bob's connected apps. That is a permission and audit-boundary violation.

   Please persist/recover the sender snapshot for each message or generation and fail closed when it is unavailable; do not reuse the sender of the recovery-triggering message. Add an Alice → lost timer → Bob `/new` regression.

4. **Two new DingTalk cleanup tests depend on unrelated database state.**

   `TestDeleteChannelInstallationsBySystemRuntimeAgentsCleansRetiredChatContexts` and `TestDeleteWorkspaceCleansRetiredChatContexts` select an arbitrary existing user instead of creating their own fixture (`server/internal/integrations/dingtalk/group_route_cleanup_db_test.go:242`). On a clean, fully migrated local database, both the targeted command and the full DingTalk package fail with `load chat creator: no rows in result set`. Please create and clean up an explicit user fixture in these tests. The current green CI is therefore not sufficient evidence that the tests are order-independent.

The overall design is otherwise heading in the right direction: generation scoping reaches task/message/history/retry/resume paths, first-party Chat retains the full visible transcript, the migrations add no foreign-key cascades, and the unique index is created concurrently. I also verified the engine race tests, relevant handler/service/Slack/migration tests, `go vet ./...`, `git diff --check`, and the byte-identical exec-environment test; only the two state-dependent DingTalk tests above failed locally.

Merge recommendation: 🔄 Request changes.

yyclaw:
Addressed all four review blockers in follow-up commit 700c26385:

1. Unified the channel context lock order as `chat_session -> binding -> generation` across append, bare `/new`, and enqueue. Added `TestChannelContextLockOrder`, which deterministically exercises the former ABBA interleaving through the production append/enqueue paths. Explicit channel enqueue now also fails closed after binding retirement, while direct Web Chat remains revisionless.
2. Migration 379 now backfills revision 1 for pre-migration tasks identifiable as channel-owned (an extant channel binding or owned `channel_ingested` input) and deliberately leaves unrelated direct Chat tasks NULL. The migration regression starts with genuinely NULL channel/direct tasks and verifies both outcomes.
3. Each context generation now transactionally snapshots its latest authenticated initiator. Crash recovery uses that persisted principal for older generations; legacy/corrupt generations without one fail closed instead of borrowing the later sender. Added Alice -> lost timer -> Bob `/new` coverage plus the missing-snapshot case.
4. The DingTalk cleanup fixture now creates, uses, and removes its own explicit creator user; it no longer queries arbitrary shared database state.

Validation completed:
- `go vet` for the affected engine/service/DingTalk/handler/migration packages
- full channel engine, DingTalk, handler, Lark, Slack, Telegram, and WeCom package tests
- targeted migration, lock-order, sender-recovery, retired-binding, direct-Chat compatibility, and rollback regressions
- behavior-coverage validation: 56 changed files, 26 production files, 26 obligations, 138 mapped candidates

I have not re-requested review yet because the PR is currently `CONFLICTING` with the latest `main` (`003a505b8`). I will reconcile that base drift first so the next review is performed on a mergeable revision.

yyclaw:
The branch is now reconciled with the latest `main` and is ready for re-review.

The four requested fixes are included in `700c26385`:

- unified the database lock order as `chat_session -> binding -> generation`, with a deterministic production-path concurrency regression;
- backfilled only identifiable pre-migration channel tasks, while leaving direct Chat tasks untouched;
- persisted the generation initiator and made crash recovery fail closed instead of borrowing a later sender's identity or connected apps;
- made the DingTalk cleanup tests create their own user fixture.

While reconciling the branch with current `main`, I preserved the newer single-pass DingTalk routing implementation instead of restoring the removed group-route retry path. The retired route tests were replaced by narrow context-cleanup regressions against the current deletion paths. The latest task batch-claim/reclaim behavior was also retained.

The reconciliation review found and fixed one additional identity-boundary case: `/issue` and other channel-command messages are excluded from agent input, so they no longer replace a generation's initiator snapshot. `TestAppendUserMessage_ChannelCommandDoesNotReplaceContextInitiator` covers this case.

Validation completed locally:

- channel engine tests and race tests;
- handler, service, DingTalk, Lark, Slack, Telegram, and WeCom targeted/package regressions;
- migration replay, concurrent-index hook, workspace deletion manifest, and context-cleanup regressions;
- `go vet` across the affected backend packages;
- final behavior-to-test inventory validation: 55 files, 26 production files, 9 behaviors, 26 obligations, 144 branch candidates, and 4 migration risks fully reconciled.

CI for the updated head is currently running.

multica-eve (changes requested):
The intended product behavior remains correct: `/new` should start an isolated topic for the agent while members can still review the complete conversation. The previous four blockers now have corresponding fixes, but this revision still has four merge-blocking isolation and rollout issues.

1. **Old-version instances do not establish the boundary during a rolling deployment.**

   Migration 379 stamps old-writer messages with the current revision (`server/migrations/379_channel_context_mixed_version_guard.up.sql:5-25`) and synchronizes only the `pending_fresh = true -> false` transition (`:120-139`). The generation advance itself exists only in the new application path (`server/internal/integrations/channel/engine/session.go:399-407`).

   I reproduced the old write shape in an isolated schema: old-version bare `/new` -> old-version next message -> old-version enqueue/clear leaves the binding, message, and task at revisions `1/1/1`; generation 2 is never created. The provider session may be fresh, but stored and Slack history remain accessible across the intended boundary. The old writer also cannot update the initiator snapshot, so a stale Alice snapshot can be reused after Bob's message if the old process crashes before task creation.

   Please cover old-writer bare `/new`, `/new <message>`, and crash recovery during mixed-version rollout. If the sender cannot be recovered, clear the stale snapshot and fail closed rather than reusing it.

2. **Direct Chat retry/rerun tasks are incorrectly converted into channel-generation tasks.**

   The source-task compatibility trigger uses `COALESCE(task.channel_context_revision, 1)` (`server/migrations/379_channel_context_mixed_version_guard.up.sql:38-48`). A retry of a normal Direct Chat task whose revision is NULL therefore gets revision 1 even though the Chat has no channel binding or generation row. Task-scoped history then treats it as a channel task and cannot resolve the generation.

   I reproduced this in an isolated schema: inserting a retry for a Direct Chat parent produced revision `1`, where NULL must be preserved. The earlier backfill already handles identifiable legacy channel tasks, so this path should copy the source revision exactly. Please add Direct Chat retry and rerun regressions that assert NULL remains NULL.

3. **A late Slack reply from an old generation can cross the new boundary.**

   A generation-1 task may complete after `/new`. The outbound path posts the reply without persisting its task/revision relationship and ignores the returned Slack message ID (`server/internal/integrations/slack/outbound.go:140-146`), while Slack history filters only by message timestamps (`server/internal/integrations/slack/history.go:235-250`). If `/new` arrives at `t3` and the old task replies at `t4`, generation-2 history includes that old-context bot reply. This violates #7135's requirement that Slack history cannot cross the task boundary.

   Please make outbound replies identifiable by generation, or merge bot output from the generation-scoped stored transcript, and add a regression for an old-generation task completing after the new boundary.

4. **Migration 379 performs an unbounded rewrite of historical channel tasks.**

   The backfill at `server/migrations/379_channel_context_mixed_version_guard.up.sql:70-86` has no task-status predicate, so it scans and updates completed, failed, and cancelled historical tasks as well as in-flight work. That creates unnecessary migration-time transaction length, row locking, and table bloat. The acceptance requirement is to preserve snapshots for queued/running work.

   Please limit the backfill to the complete in-flight status set, including `deferred` and `waiting_local_directory`, and use the existing pending-chat partial index. If historical rows truly must be repaired, use a staged or batched operational path instead of a startup migration-wide rewrite.

Validation on `4b6b7ab6`: channel engine race tests, affected handler/service/Slack/migration/DingTalk tests, `go vet ./...`, and `git diff --check` passed. The two isolated-schema checks above reliably reproduced the rolling-deploy boundary failure and Direct Chat revision misclassification.

Merge recommendation: 🔄 Request changes.

yyclaw:
Addressed the four blockers from the review on `4b6b7ab6` in `1306b83ea` and `8e3c9964e`, after merging current `upstream/main` (`fbbe69757`).

1. **Deployment contract:** I removed the incomplete old-writer stamping and pending-state synchronization triggers. The repository Helm chart uses `strategy.type: Recreate`, and the container runs migrations before the server; arbitrary old/new backend overlap is not a supported topology. Nullable columns plus the generation-1 NULL fallback preserve adjacent-release rollback. I updated both #7135 and this PR description to state that contract explicitly. The cross-revision message/task ownership trigger remains because it enforces a durable data invariant.
2. **Direct Chat retry/rerun:** the generic source-task stamping trigger is gone. Application retries already copy `channel_context_revision` exactly. The isolated migration regression now inserts Direct Chat retry and rerun rows and verifies both remain NULL.
3. **Late Slack output:** Slack now returns every chunk timestamp and persists them on the exact assistant task, scoped by chat session, installation, real channel, adapter, and task generation. Generation >1 history intersects only IDs from the current Slack page with that generation. A late old-generation reply therefore remains visible to people in Slack but is excluded from new-generation agent context. Missing provenance fails closed, and pagination keeps the provider-page cursor even when every message in the page is filtered.
4. **Migration backfill:** migration 379 now updates only channel-owned tasks in `queued`, `dispatched`, `running`, `waiting_local_directory`, or `deferred`. Completed history and unrelated Direct Chat tasks remain NULL.

Validation on `8e3c9964e`:

- clean worktree database replayed migrations 001-397, including up/down/replay coverage for 377-379;
- affected handler, migration, service, shared channel engine, Slack, DingTalk, Lark, Telegram, and WeCom packages pass with `-race`;
- affected packages pass `go vet`;
- `make sqlc` is clean with repository-standard sqlc v1.31.1;
- final preflight inventory validates 60 files, 29 production files, 10 behaviors, 30 obligations, 155 production branch candidates, and 8 migration risks;
- the full race suite had one unrelated fixed-deadline DSH probe timeout; that exact test passed immediately on isolated rerun (`compatible`: 0.95s).

The branch is synchronized with current `main` and ready for re-review.

multica-eve (approved):
Thank you for the thoughtful iterations and thorough regression coverage. The latest revision addresses the rollout contract, Direct Chat retry/rerun, late Slack output, and migration-scope blockers, and the branch is now reconciled with current main. Approved.

PR #7364, opened 2026-08-21
MUL-6518 ALL-151: retry task temp cleanup on Windows

j499712089:
## Summary

- retry task temp directory cleanup with bounded backoff after transient file-sharing failures
- log retry recovery separately from exhausted cleanup failures
- reproduce the Windows `node-compile-cache` sharing violation and run the regression in Windows CI

## Impact

- measured 174 residual `multica-task-*` directories using 334,484,328 bytes (318.989 MiB) in the local Windows temp directory
- cleanup adds no delay on the normal path and at most 3.85 seconds only after repeated failures

## Verification

- `go test ./internal/daemon -run '^(TestCleanupTaskTempDirWithRetriesTransientFailure|TestCleanupTaskTempDirWithRetriesExhausted|TestCleanupTaskTempDirRetriesWindowsSharingViolation|TestTaskTempBaseDir)$' -count=1`
- `go vet ./internal/daemon`
- `go test ./internal/daemon -run '^$' -count=1`

Related: ALL-151

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -709,6 +709,11 @@ jobs:
           go test ./internal/daemon/execenv -v -run '^(TestSeedUserCodexSkills|TestHydrateCodexSkills)' -count=1 -timeout=5m
           go test ./internal/daemon -v -run '^(TestCleanTaskArtifacts_DoesNotFollowDirectoryJunction|TestTaskSize_DoesNotCountDirectoryJunction)$' -count=1 -timeout=5m
 
+      - name: Test Windows task temp cleanup retry
+        if: ${{ needs.changes.outputs.backend == 'true' }}
+        working-directory: server
+        run: go test ./internal/daemon -v -run '^TestCleanupTaskTempDirRetriesWindowsSharingViolation$' -count=1 -timeout=5m
+
       - name: Test Windows agent executable junction resolution
         if: ${{ needs.changes.outputs.backend == 'true' }}
         working-directory: server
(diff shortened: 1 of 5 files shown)


Bohan-J:
Thank you for this — and especially for measuring it. "174 leftover directories, ~319 MB" is what turned this from a vague suspicion into something we could act on, and the Windows test is genuinely good work: opening the file with `CreateFile` without `FILE_SHARE_DELETE` to produce a real sharing violation, and asserting that `os.RemoveAll` actually fails *before* verifying the retry, is a lot more solid than mocking the failure.

After a close look we've decided to fix the underlying problem instead of merging the retry, for two reasons:

1. **A fixed retry budget only wins when the holder releases inside the window.** That holder is usually a process the task itself spawned and that didn't exit cleanly — an MCP server, a language server, a watcher. Its lifetime has nothing to do with the budget we pick, so ~3.85s is a guess in both directions: too long when the file is never released, too short when the holder lives for minutes. It also can't help the two cases where nothing runs at all — a daemon crash or kill skips the `defer` entirely — nor the directories already on your disk.

2. **The retry sits in an expensive spot.** That `defer` is the last one registered in `runTask`, so LIFO runs it *before* `CleanupSidecars` and `Finalize` — before the worktree branch is committed, the task result is reported, and the concurrency slot is released. It trades a certain delay on every failed cleanup for an uncertain removal.

The real gap your report exposed is that these directories have **no reclaimer at all**. The GC loop scans `WorkspacesRoot`, while task temp dirs live under the system temp base — two roots with no overlap, and `multica-task-` has exactly one create site and zero reclaim sites in the tree. So the fix is a temp-base sweep in the GC: unbounded retries (locked now → retried next cycle, until it succeeds) at zero cost on the task-completion path, and it covers your existing 174 directories and the crash case as well.

That's tracked in #7376 with the design and the safety constraints, and our team is picking it up — so I'm closing this PR. To be clear about why: not because the report was wrong, but because it was right enough to move where we fix it. Your analysis is credited on the issue.

We'd genuinely like to see more contributions from you.

PR #7369, opened 2026-08-21
MUL-6525: docs: document GitHub App Contents permission for PR snapshots

sebasanblas:
## Problem

The self-hosting documentation listed the GitHub App permissions for PR snapshots as Metadata, Pull requests, Checks, and Commit statuses, but omitted the repository Contents permission.

## Why this permission is required

The PR snapshot uses GitHub GraphQL to traverse `pullRequest.commits` to the head commit and read `commit.statusCheckRollup`. GitHub requires the App to have **Contents: Read-only** access to read that Commit object. Without it, the documented permission set is incomplete and the CI rollup snapshot can fail even when Checks and Commit statuses are enabled.

## Changes

- Added **Contents — Read-only** with the GraphQL snapshot context to the English GitHub integration documentation.
- Updated the corresponding permission tables in the Japanese, Korean, and Simplified Chinese translations.

The implementation is referenced in [`snapshot.go`](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/server/internal/integrations/ghsnapshot/snapshot.go#L21-L39), and the affected documentation section is [here](https://github.com/multica-ai/multica/blob/38c992ad0a757434fb51584fa34e3bc57d1b78e1/apps/docs/content/docs/github-integration.mdx#L122-L129).

## Verification

- `git diff --check`
- Documentation typecheck could not run because `apps/docs/node_modules` is not installed (`fumadocs-mdx: command not found`).

### DIFF
--- a/apps/docs/content/docs/github-integration.ja.mdx
+++ b/apps/docs/content/docs/github-integration.ja.mdx
@@ -123,6 +123,7 @@ Repository permissions:
 | 権限 | レベル |
 |---|---|
 | Metadata | Read-only |
+| Contents | Read-only。GitHub GraphQL の PR スナップショットが head commit をたどり、その CI ロールアップを読み取るために必要 |
 | Pull requests | Read-only |
 | Checks | Read-only。CI の状態表示に使用 |
 | Commit statuses | Read-only。legacy status 形式の CI 集計に使用 |
--- a/apps/docs/content/docs/github-integration.ja.mdx
+++ b/apps/docs/content/docs/github-integration.ja.mdx
@@ -123,7 +123,7 @@ Repository permissions:
 | 権限 | レベル |
 |---|---|
 | Metadata | Read-only |
-| Contents | Read-only。GitHub GraphQL の PR スナップショットが head commit をたどり、その CI ロールアップを読み取るために必要 |
+| Contents | Read-only。PR スナップショットのクエリが head commit・マージ可否・CI ロールアップをまとめて読み取るために必要 |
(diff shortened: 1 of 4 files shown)


Bohan-J:
Thanks for catching this, @sebasanblas — the missing `Contents` permission is a real gap in the self-hosting setup docs, and updating all four locales in one pass is exactly right.

One thing I wanted to tighten before merging, so I pushed a small follow-up commit to your branch instead of sending it back to you.

The new row framed `Contents` as a CI-rollup permission, and it sits a few lines above *"If you don't need CI shown in Multica, you can skip the Checks and Commit statuses permissions and their events."* — a reader could reasonably conclude `Contents` is skippable too. It isn't: `prSnapshotQuery` asks for `headRefOid`, `mergeable`, `mergeStateStatus` and the head commit's `statusCheckRollup` in one request, and `client.go` discards the entire response as soon as GitHub returns any GraphQL error. Without `Contents` the symptom isn't "no CI" — the PR card loses merge state as well.

So the follow-up commit:

- rewords the `Contents` row to cover the whole snapshot (head commit + mergeability + CI rollup) in all four locales;
- adds a sentence to the "you can skip the CI permissions" paragraph noting `Contents` is still required;
- lists `Contents` in the **No CI status** troubleshooting entry, which is where an admin hitting this actually lands.

Merging once CI goes green. Thanks again for the first contribution — this one saves self-hosters a genuinely annoying debugging session.

PR #7374, opened 2026-08-21
fix(cli): redact webhook credential in autopilot get

### LINKED ISSUE #7352: [Bug]: `autopilot get` exposes the live webhook credential in normal output
### Deployment type

Official App / Multica Cloud

### What happened?

`multica autopilot get <autopilot-id> --output json` returns the complete live calling credential for every webhook trigger in ordinary detail output.

The response includes these fields with unredacted values:

- `webhook_path`
- `webhook_token`
- `webhook_url`

This is surprising because the official Autopilots documentation says the UI hides the token by default and copying the URL does not require revealing it. The CLI command has no redaction flag and its help text only describes the output format.

This issue is distinct from #2883, which proposes stronger webhook authentication. Even with the current path-token design, a normal read/detail command should not print a reusable bearer credential by default.

### Steps to reproduce

1. On Multica Cloud, create an Autopilot with a generic webhook trigger.
2. Run:

   ```shell
   multica autopilot get <autopilot-id> --output json
   ```

3. Observe that the complete webhook URL and token are present in stdout.

Verified with Multica CLI/Desktop version `0.4.31` on macOS arm64.

No real webhook URL, token, workspace ID, user identity, or Autopilot ID is included in this report.

### Expected behavior

Normal `get` output should be safe for terminals, logs, automation, and agent contexts:

- redact the token by default in both table and JSON output;
- return only metadata such as `has_token: true`, a short non-sensitive hint, and trigger state;
- require an explicit secret-bearing command or flag to reveal/copy the full URL;
- print a warning when secret material is intentionally requested;
- ideally return a newly generated credential only once at create/rotate time.

### Security impact

The token embedded in the URL is the calling credential. Printing it during an ordinary inspection command can copy it into terminal scrollback, CI logs, shell transcripts, support bundles, screen recordings, and LLM/agent tool context. If any of those leave the machine, an external party can invoke the cloud webhook until it is disabled or rotated.

The affected synthetic trigger was disabled, its parent Autopilot was paused, and every exposed credential was rotated. No production webhook was involved.

### Additional context

Official documentation: https://www.multica.ai/docs/autopilots#protect-the-webhook-url

Related but different: #2883.

sanjay3290:
Fixes #7352.

### Problem

`multica autopilot get` printed the raw API response. For a caller with writer access that response contains the live `webhook_token`, plus `webhook_url` and `webhook_path` which embed it. Anyone who ran the command in a shared terminal, a CI log, or a pasted bug report leaked a working credential.

### Fix

- Redact `webhook_token`, `webhook_url`, and `webhook_path` to `[redacted]` in the default output.
- Add `has_webhook_token` (bool) so scripts can still check whether one is configured.
- Add `--show-secrets` to print the real values, with a warning on stderr.

Text and JSON output paths are both covered.

### Tests

`cmd_autopilot_test.go` asserts the token never appears in default output, that `has_webhook_token` reflects presence, that `--show-secrets` restores the raw values, and that the flag is registered on the real command (deleting the registration fails the suite).

Verified locally:

```
cd server
go build ./...   # ok
go vet ./...     # ok
go test ./...    # all packages pass
```

### DIFF
--- a/server/cmd/multica/cmd_autopilot.go
+++ b/server/cmd/multica/cmd_autopilot.go
@@ -115,6 +115,7 @@ func init() {
 
 	// get
 	autopilotGetCmd.Flags().String("output", "json", "Output format: table or json")
+	autopilotGetCmd.Flags().Bool("show-secrets", false, "Show live webhook credentials (webhook_token, webhook_url, webhook_path)")
 
 	// create
 	autopilotCreateCmd.Flags().String("title", "", "Autopilot title (required)")
@@ -241,6 +242,13 @@ func runAutopilotGet(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("get autopilot: %w", err)
 	}
 
+	showSecrets, _ := cmd.Flags().GetBool("show-secrets")
+	if showSecrets {
+		fmt.Fprintln(os.Stderr, "warning: output contains a live webhook credential")
+	} else {
+		redactAutopilotWebhookSecrets(resp)
+	}
+
 	output, _ := cmd.Flags().GetString("output")
 	if output == "json" {
 		return cli.PrintJSON(os.Stdout, resp)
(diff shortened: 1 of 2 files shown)


sanjay3290:
Closing as a duplicate — #7365 landed the same fix (and a superset of it: it also adds `has_webhook_token` and `webhook_token_hint`). Nothing left here. Thanks @iiwish.

PR #7383, opened 2026-08-21
MUL-6493 feat(issues): add Project as a grouping option on Board and Table

### LINKED ISSUE #7337: [Feature]: Add "Project" as a grouping and sort option on Board and Table views
### What do you want and why?

This consolidates feedback seen in the Multica Discord community: teams that model each customer/client as a separate Project (agencies, consultancies — 80+ projects in some reported cases) find it hard to organize the issue list around that structure once the project count grows.

Looking at the current view infra, most of the building blocks already exist — filtering by project works in every view, and the Swimlane view already supports `project` as a grouping dimension (alongside `parent` and `assignee`). The actual gap is narrower than it first looks: Board and Table views only expose `status` / `assignee` (plus custom properties) as grouping options, and no view's sort field list includes `project` or `assignee` at all (`SortField` is limited to position/status/priority/dates/title/property). So a user managing many customer-projects can filter down to one project everywhere, and can only *group* by project in Swimlane — Board and Table (the two most-used views) have no equivalent.

### Proposed solution

- Extend `IssueGrouping` (Board) and `TableGrouping` (Table) to accept `"project"`, mirroring the existing `SwimlaneGrouping` set (`"parent" | "project" | "assignee"`). Reuse the swimlane grouping/lane logic where practical instead of duplicating it.
- Since Table grouping is server-backed, add project as a valid grouping key on that endpoint, not just the frontend `TableGrouping` type.
- Add `"project"` and `"assignee"` to `SortField` so Board/List/Table sort can order by project name / assignee name, not just status/priority/dates/title.
- No new persistence shape needed beyond what already exists for `grouping` / `tableGrouping` / `sortBy` in the issue view store — this only widens the accepted values and the UI options list (`GROUPING_OPTIONS`, table grouping menu, sort menu).
- Worth coordinating with #3617 (Gantt grouping) and #5992 (swimlane axis transpose), which are the same underlying pattern — "give other views what Swimlane already has" — applied to different views.

Bohan-J:
Closes #7337.

## What this changes

`Project` becomes a grouping option on **Board** and **Table**, matching what Swimlane already offered. Sorting by project / assignee is deliberately **not** part of this change — see below.

Board grouped by project gets one column per project (plus a pinned "No project" column), with the project's own icon in the header, drag-to-reprojectise, and the same server-paged columns and load-more the assignee board uses. Table grouped by project gets the same groups as collapsible group rows.

## Why it is this small

The server side already exists. `server/internal/handler/issue_table_group.go` has carried `{ kind: "project" }` — group expression, `__no_project__` bucket, ordering by project title, `project:<id>` / `project:none` descriptor keys — since the Swimlane shipped, and `IssueTableGroupSpec` already accepted it. **There are no backend changes here**: this widens `IssueGrouping` / `TableGrouping` and routes them to the endpoint that was already answering.

Saved views store `display` as opaque JSON server-side, so no migration is needed either.

## Things worth a reviewer's attention

- **`projectGroupId()` reproduces the server's group key exactly.** The board builds column ids from cards while the server builds them from descriptors — if the two disagree, every card silently falls off the board. Pinned by a test.
- **`projectId` sits beside `assigneeId` on `BoardColumnGroup`.** Both are optional, so an unguarded project column reads as "no assignee": the drop handler would unassign the card on every drop, and the column heading would render a no-assignee glyph next to the project title. Both call sites now match `projectId` explicitly, and both failure modes have a test.
- **Group titles resolve through the projects query**, which the surface now primes when this grouping is active. A project the query cannot resolve (deleted, or invisible to the member) reads as unavailable — never as its raw uuid, which is what the pre-existing table branch would have rendered had anything ever reached it.
- **`buildGroups` takes a context object now.** It was at 7 positional parameters and this needed two more; the call site is the only one.
- One non-obvious constraint the existing suite caught: reading `data` off the projects query instead of destructuring `= []`. An un-settled query has no data, so the default array is fresh per render and churns every downstream memo — the MUL-5477 commit-storm guard in `table-view-virtualized-hierarchy.test.tsx` fails on it.

## Not included

Sorting by project / assignee (also requested in #7337) is a separate change: three ORDER BY paths, a correlated-subquery sort with no index behind keyset pagination, and the client comparator's `default: position` fallthrough, which would silently reorder Swimlane and Gantt. It deserves its own PR and its own performance check.

## Testing

- `packages/views/issues/utils/drag-utils.test.ts` — group key agreement, bucketing, move updates, and the assignee-fallthrough guard.
- `packages/views/issues/components/board-view-project-grouping.test.tsx` — column titles, server group counts, project icon in the header, cards landing in their own column.
- `packages/views/issues/components/table-view-project-grouping.test.tsx` — group spec, title resolution, no-project and unresolvable-project labels.
- `packages/views/issues/surface/use-issue-surface-controller.test.tsx` — the board requests the grouping dimension it is set to.

`pnpm typecheck`, `pnpm lint`, `pnpm test` (4600+ tests) all pass locally. Each new test was confirmed to fail against the unpatched code. No manual pass against a running stack — the reviewer may want to eyeball the board column header.

### DIFF
--- a/packages/core/issues/stores/view-store.ts
+++ b/packages/core/issues/stores/view-store.ts
@@ -11,12 +11,16 @@ import { defaultStorage } from "../../platform/storage";
 export type ViewMode = "board" | "list" | "table" | "gantt" | "swimlane";
 export type GanttZoom = "day" | "week" | "month";
 /**
- * Board grouping. Besides the two built-ins, a select-type custom property
+ * Board grouping. Besides the three built-ins, a select-type custom property
  * groups columns by its options via the `property:<definitionId>` form.
  * Persisted values may reference a since-archived definition — consumers must
  * fall back to "status" when the definition can't be resolved.
  */
-export type IssueGrouping = "status" | "assignee" | `property:${string}`;
+export type IssueGrouping =
+  | "status"
+  | "assignee"
+  | "project"
+  | `property:${string}`;
 export type SwimlaneGrouping = "parent" | "project" | "assignee";
 /**
  * Sort key. `property:<definitionId>` is resolved server-side against the
@@ -55,7 +59,12 @@ export interface TableColumnConfig {
   key: TableColumnKey;
   width?: number;
 }
-export type TableGrouping = "none" | "status" | "assignee" | `property:${string}`;
+export type TableGrouping =
+  | "none"
+  | "status"
+  | "assignee"
+  | "project"
+  | `property:${string}`;
 export type TableCalculation = "none" | "sum" | "average" | "count";
 
 export const TABLE_SYSTEM_COLUMNS: readonly TableSystemColumnKey[] = [
@@ -155,6 +164,7 @@ export const SORT_OPTIONS: { value: StaticSortField; label: string }[] = [
 export const GROUPING_OPTIONS: { value: StaticIssueGrouping; label: string }[] = [
   { value: "status", label: "Status" },
   { value: "assignee", label: "Assignee" },
+  { value: "project", label: "Project" },
 ];
 
 export const CARD_PROPERTY_OPTIONS: { key: keyof CardProperties; label: string }[] = [
--- a/packages/views/issues/components/board-column.tsx
+++ b/packages/views/issues/components/board-column.tsx
@@ -2,7 +2,7 @@
 
 import { memo, useCallback, useMemo, useState, type ReactNode } from "react";
 import { Virtuoso } from "react-virtuoso";
-import { EyeOff, MoreHorizontal, Plus, UserMinus } from "lucide-react";
+import { EyeOff, FolderMinus, MoreHorizontal, Plus, UserMinus } from "lucide-react";
 import { useDroppable } from "@dnd-kit/core";
 import { SortableContext, verticalListSortingStrategy } from "@dnd-kit/sortable";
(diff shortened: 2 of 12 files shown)


Bohan-J:
Both review nits are addressed in `25e61a45a`.

**1. The "No project" column is now actually pinned.** It was derived from the result set like every other project column, so a workspace where every card has a project lost it — and with it the only way to clear a card's project by dragging. It is now kept whenever the board has any columns at all. An entirely empty board still renders its empty state rather than one lone column, and it does not duplicate when the server does return a `project:none` group. Two tests: one with a descriptor set that omits `project:none` (asserts the column is there with count 0), one asserting exactly one such column when the server does return it.

**2. Board and Table now agree on an unresolvable project.** The board fell back to an empty title while the Table said "Unavailable value" for the same project — so an unresolvable project could render as a board column with an icon, a count, and no name. Both share the Table's wording now.

Both paths that build a project column go through one helper, which is what stops the client fallback and the server-descriptor path drifting apart like this again.

The board test fixture is now parameterised per test rather than being one shared object, which is what made the first nit invisible.

`pnpm typecheck`, `pnpm lint`, `pnpm test` pass locally; both new tests were confirmed to fail against the previous commit.

PR #7384, opened 2026-08-21
fix(daemon): skip the path mutex for chat tasks

sanjay3290:
## Problem

Chat turns take the `local_directory` path mutex even though a chat turn makes no edits. A conversation therefore parks behind a long worker build running in the same directory, and the user sees the chat hang with no explanation.

The existing `IsLeaderTask` carve-out does not help: chat tasks are identified by `ChatSessionID`, and never set `IsLeaderTask`.

## Change

Skip the path mutex for tasks with `ChatSessionID` set, after validation, the same way worktree mode already does. A comment records the trade-off: an edit issued mid-conversation now runs unserialised against a concurrent worker.

12 lines of production change in `server/internal/daemon/daemon.go`.

## Test

`server/internal/daemon/local_directory_test.go` adds a test that starts a chat task while the path mutex is held by another task, runs it in a goroutine, and selects on a timer. It asserts the chat turn proceeds instead of blocking.

The timer/goroutine shape is deliberate. An assertion on elapsed time cannot catch this regression: if the mutex is reinstated the call blocks forever, so the assertion is never reached and the package dies on the suite timeout instead of failing one test.

Deletion probe: reverting the skip makes the new test fail in ~2s on the select timeout.

## Verification

```
cd server
go build ./...        # ok
go vet ./...          # ok
go test ./... -count=1 # 0 FAIL
```

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5334,6 +5334,18 @@ func (d *Daemon) acquireLocalDirectoryLockIfNeeded(ctx context.Context, task Tas
 		return nil, false
 	}
 
+	// Chat tasks are conversations, not edits — they do not mutate the
+	// working tree, so the path mutex that serialises concurrent edits
+	// need not block them. Skipping it avoids a chat turn waiting behind
+	// a long worker build in the same repo. Trade-off: a chat turn that
+	// is asked to edit a file mid-conversation now runs unserialised
+	// against a worker in the same tree; future work could acquire
+	// lazily on first write instead.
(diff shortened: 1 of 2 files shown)


sanjay3290:
Superseded by #7378 (merged as bae3a324f), which fixes the same MUL-6504 hold more completely — it splits the placement decision from the lock decision via `localDirectoryLockExempt` rather than skipping the mutex inline, and also surfaces `wait_reason` in the status pill. Closing in favour of that. Thanks.

PR #7387, opened 2026-08-21
fix(daemon): GC-reclaim orphaned per-task temp dirs under the task temp base

### LINKED ISSUE #7376: GC: reclaim orphaned per-task temp dirs (multica-task-*) under the task temp base
## Problem

Every task gets a private temp directory created with `os.MkdirTemp(base, "multica-task-")` (`server/internal/daemon/daemon.go:8547`) and exported to the agent process as `TMPDIR` / `TMP` / `TEMP`. That directory lives under the **system temp base** — `/tmp` or `os.TempDir()`, optionally relocated with `MULTICA_AGENT_TEMP_BASE` (`server/internal/daemon/daemon.go:8568`).

Its only path to removal is the `defer os.RemoveAll(taskTempDir)` registered in `runTask` (`server/internal/daemon/daemon.go:6952`). If that single call does not succeed, the directory stays on disk forever:

- the removal fails (on Windows, a file inside is still open — the case reported in #7364, with ~174 leftover directories totalling ~319 MB on one machine);
- the daemon crashes or is killed, so the `defer` never runs at all.

The GC loop does not cover these. Its scan root is `d.cfg.WorkspacesRoot` (`server/internal/daemon/gc.go:81`), and the task temp base is a different root with no overlap. Grepping `multica-task-` across the Go tree finds exactly one create site and zero reclaim sites.

The size is not the interesting part — the absence of an upper bound is. The directory is the agent's `TMPDIR`, so anything a task's tooling writes to temp lands there; a single task that downloads a large file or writes build intermediates to temp can leave a multi-GB directory behind with nothing to reclaim it.

## Why the bounded retry in #7364 is not the fix

#7364 proposes retrying the end-of-task `RemoveAll` with backoff (~3.85s total). That helps only when the holder releases the file inside the budget. The holder is typically a process the task itself spawned that did not exit cleanly (MCP server, language server, watcher) — its lifetime is unrelated to any budget we pick, so a fixed window is a guess in both directions. It also does not help the two cases above where the `defer` never ran or the leftovers already exist.

It is also not free: that `defer` is the last one registered in `runTask`, so LIFO puts it **before** `CleanupSidecars` and `Finalize` — i.e. before the worktree branch is committed, the task result is reported, and the concurrency slot is released. It buys an uncertain cleanup with a certain delay on the task-completion path.

A GC sweep is strictly better on both axes: unbounded retries (a directory that is locked now is simply retried on the next cycle, and the cycle after that, until it succeeds) at zero cost on the hot path.

## Proposed fix

Add a temp-base sweep to `runGC`, in the same shape as the three existing out-of-`WorkspacesRoot` pruners (`server/internal/daemon/gc.go:115`, `:124`, `:132`):

```go
PruneTaskTempDirs(base, d.cfg.GCTaskTempTTL, time.Now(), d.reserveStoreForDeletion, d.logger)
```

- **Scan root**: resolve via `taskTempBaseDir()` — do not hardcode `os.TempDir()`, since `MULTICA_AGENT_TEMP_BASE` can relocate it on Unix.
- **Match**: only entries whose basename has the `multica-task-` prefix. Never touch the base directory itself or anything else in it — it is usually a shared `/tmp`.
- **Removal**: plain `os.RemoveAll`; a failure is not an error worth escalating, it just gets retried next cycle. On Windows a partial removal is expected and fine.

## Safety constraints

**Liveness must come from the active set, not from mtime.** A running task can go tens of minutes without touching the root of its temp directory, so its mtime stays at creation time — an mtime-only rule would delete a live task's `TMPDIR`.

The existing store-reservation protocol already provides exactly what is needed and carries no store-specific semantics — `markActiveStore` / `unmarkActiveStore` / `reserveStoreForDeletion` (`server/internal/daemon/daemon.go:8389`, `:8422`) are a path-keyed refcount plus an atomic confirm-inactive-and-reserve that closes the stat→remove race. Reuse it: mark the temp dir right after `ensureTaskTempDir`, unmark in the same `defer`, and pass `d.reserveStoreForDeletion` to the pruner.

**mtime TTL is only the backstop**, for directories orphaned by a daemon restart (the active set is in-memory, and a restarted daemon's orphaned children may still be running). It must therefore be well above the longest plausible task runtime — the same order as `DefaultGCOrphanTTL` (24h–72h), not an aggressive value. These directories are small on average and latency costs nothing; the value of the sweep is that it *always* succeeds eventually, not that it is fast.

## Acceptance criteria

- [ ] `GCTaskTempTTL` config knob with a conservative default and a `MULTICA_GC_TASK_TEMP_TTL` override, consistent with the other GC TTLs (`server/internal/daemon/config.go:65-80`).
- [ ] Sweep runs inside `runGC`, counted in `gcStats` (count + bytes reclaimed) and included in the cycle-complete log.
- [ ] A live task's temp dir is never removed, including when its mtime is older than the TTL.
- [ ] Only `multica-task-*` entries are considered; the base directory and unrelated siblings are untouched.
- [ ] A removal that fails (e.g. a locked file on Windows) leaves the directory for the next cycle without failing the GC cycle.
- [ ] Tests cover: TTL expiry, active-task protection, prefix scoping, and `MULTICA_AGENT_TEMP_BASE` relocation.

## Notes

Line references are against `main` @ `1dc1355`.

Thanks to @j499712089 for finding and quantifying this in #7364.

sanjay3290:
Closes #7376.

Every task gets a private TMPDIR (`multica-task-*` under the task temp base, usually `/tmp`) from `ensureTaskTempDir`. Its only removal path is a defer in `runTask`, so a daemon crash — or a failed removal on Windows — leaves it forever. `runGC` scans `WorkspacesRoot` only, so these are invisible to it and unbounded in size.

## Change

- `execenv.PruneTaskTempDirs(root, retention, now, reserve, logger)` in `execenv/task_temp.go`, in the shape of the existing `PruneCodexSessionStores` / `PruneHermesMemoryStores`: returns `(removed, bytesFreed)`, `retention <= 0` disables the sweep, an unreadable root is a no-op, only `multica-task-*` directories are ever considered, and a removal failure is logged and retried on the next cycle.
- `runGC` calls it after the Hermes pruners with the root from `taskTempBaseDir()`; new `taskTempDirsReclaimed` counter in `gcStats`, in the cycle-complete guard and the log line.
- `Config.GCTaskTempTTL` / `DefaultGCTaskTempTTL` (72h) / `MULTICA_GC_TASK_TEMP_TTL`.
- `runTask` marks the temp dir active right after `ensureTaskTempDir` and unmarks it in the cleanup defer, after `RemoveAll`.

## Why liveness is not mtime

A running task can go tens of minutes without touching the *root* of its temp dir, so the root mtime stays at creation time. An mtime-only rule would delete a live task's TMPDIR. Liveness comes from the existing `markActiveStore` / `reserveStoreForDeletion` reservation protocol — the same exclusive check-and-mark the store pruners use (MUL-4424). The mtime TTL is only the backstop for dirs orphaned by a daemon restart, when the active set is gone. Size and age come from `codexStoreStat`, which reports the newest mtime across the whole tree rather than the root's.

## Tests

9 new tests. Two cover the safety property directly — a marked-active dir with a 30-day-stale mtime survives, both through a stub `reserve` at the pruner level and through real `markActiveStore` / `reserveStoreForDeletion` via `runGC`. A `runTask` harness test proves the wiring rather than the pruner alone: from the agent `/start` handler it globs the in-flight `multica-task-*` dir and asserts `reserveStoreForDeletion` refuses it. The removal-failure path is covered by chmod-ing the base to `0500` (self-skips under root).

Verified with `go build ./...`, `go vet ./...`, `gofmt`, `GOOS=windows`/`GOOS=linux GOARCH=amd64 go build ./cmd/... ./internal/...`, and `go test ./... -count=1` — zero failures.

### DIFF
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -73,6 +73,7 @@ const (
 	DefaultGCOrphanTTL                    = 72 * time.Hour      // 3 days — orphans with no meta (crashes, pre-GC leftovers)
 	DefaultGCArtifactTTL                  = 12 * time.Hour      // 12h — drop regenerable artifacts once a task has been completed this long
 	DefaultGCCodexSessionTTL              = 14 * 24 * time.Hour // 14 days — reclaim per-issue Codex session stores untouched this long
+	DefaultGCTaskTempTTL                  = 72 * time.Hour      // 3 days — reclaim an orphaned per-task temp dir (the agent's TMPDIR) this stale; well above the longest task runtime because liveness comes from the active-store guard, not mtime
 	DefaultGCHermesMemoryTTL              = 90 * 24 * time.Hour // 90 days — reclaim per-agent Hermes memory stores untouched this long (long: reclaiming these is visible amnesia, and they are a few markdown files)
 	DefaultGCHermesSessionTTL             = 14 * 24 * time.Hour // 14 days — reclaim per-conversation Hermes session stores untouched this long (matches Codex: these hold transcripts, and losing an idle one restarts the thread rather than the agent's notes)
 	DefaultGCRepoTTL                      = 30 * 24 * time.Hour // 30 days — evict a bare repo cache no task has checked out this long
@@ -112,6 +113,7 @@ type Config struct {
 	GCRepoTTL                      time.Duration         // evict a cached bare repo under .repos once no task has created a worktree from it for this long, it has no worktrees left, and it is no longer attached to any watched workspace (default: 30d, set 0 to disable)
 	GCRepoMaintenanceEnabled       bool                  // run reflog expiry and git gc after stale agent refs are removed (default: true; disable independently as an operational kill switch)
 	GCCodexSessionTTL              time.Duration         // reclaim a per-issue Codex session store (~/.codex/multica-sessions/<agent>/<issue>) untouched for at least this long, so a done/abandoned issue's conversation history does not accumulate forever (default: 14d, set 0 to disable)
+	GCTaskTempTTL                  time.Duration         // reclaim an orphaned per-task temp dir (the agent's TMPDIR under the task temp base) untouched for at least this long; a live task's dir is protected by the active-store guard, so this only backstops dirs orphaned by a daemon restart (default: 72h, set 0 to disable)
 	GCHermesMemoryTTL              time.Duration         // reclaim a per-agent Hermes memory store (<profile dir>/hermes-state/<agent>/<profile>) untouched for at least this long, so a deleted agent's memory does not sit on disk forever (default: 90d, set 0 to disable)
 	GCHermesSessionTTL             time.Duration         // reclaim a per-conversation Hermes session store (<profile dir>/hermes-sessions/<agent>/<profile>/<conversation>) untouched for at least this long, so a done or abandoned conversation's transcript does not accumulate forever (default: 14d, set 0 to disable)
 	AutoUpdateEnabled              bool                  // periodically check for a newer CLI release and self-update when idle (default: true on Multica Cloud, false on self-host)
@@ -471,6 +473,10 @@ func LoadConfig(overrides Overrides) (Config, error) {
 	if err != nil {
 		return Config{}, err
 	}
+	gcTaskTempTTL, err := durationFromEnv("MULTICA_GC_TASK_TEMP_TTL", DefaultGCTaskTempTTL)
+	if err != nil {
+		return Config{}, err
+	}
 	gcHermesMemoryTTL, err := durationFromEnv("MULTICA_GC_HERMES_MEMORY_TTL", DefaultGCHermesMemoryTTL)
 	if err != nil {
 		return Config{}, err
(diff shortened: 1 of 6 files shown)


multica-eve (changes requested):
Thanks for working on this cleanup path. The product direction makes sense, but two safety issues need to be addressed before merge:

1. **An active task directory can be deleted by another profile's daemon.** The GC scans the shared system temp root, while its active-directory reservations are process-local. Two profiles can run concurrently on the same machine, so one daemon can reclaim another daemon's still-active directory once it exceeds the TTL. Please place task temp directories in a stable profile-specific namespace, scan only that namespace, handle coexistence with legacy unnamespaced directories conservatively, and add a two-daemon regression test.

2. **Windows directory junctions can make the GC traversal escape the task directory.** The reused statistics walker does not skip directory junctions, so a task-controlled junction may cause unbounded traversal outside the temp directory and incorrect accounting. Please use a walker that does not follow symlinks or junctions, add a real Windows junction regression test, and ensure that test runs in the Windows CI job.

Two non-blocking follow-ups would also strengthen this change: verify that a task directory becomes reservable after `runTask` returns, and document `MULTICA_GC_TASK_TEMP_TTL` plus its default cleanup behavior.

multica-eve:
感谢贡献。这项工作希望清理异常任务遗留的临时目录，避免长期占用磁盘。

该用户问题已由已合入的 #7391 解决，并关闭了 #7376。#7391 提供相同的用户结果，且是当前维护的实现；为避免保留两套重复方案，关闭本 PR。

PR #7390, opened 2026-08-21
MUL-6538 fix(daemon): explain macOS TCC EPERM in the local_directory access probe

### LINKED ISSUE #6793: [Bug]: Mac local directory open operation not permitted
### Deployment type

Official App

### What happened?

local_directory: read "path": open “path” operation not permitted

### Steps to reproduce

1. add local directory as project resource
2. create a new task
3. agent execute failed
4. local_directory: read "path": open “path” operation not permitted

### Screenshots (optional)

_No response_

```shell

```

sanjay3290:
Fixes #6793.

## Problem

Binding a project resource to a directory under `~/Documents` fails the daemon's access probe with:

```
local_directory: read "/Users/<user>/Documents/Work/<project>": open /Users/<user>/Documents/Work/<project>: operation not permitted
```

The reporter confirmed Terminal can list the same directory, Full Disk Access is OFF and Files & Folders is ON. This is macOS TCC privacy protection, not a filesystem permission problem. `EPERM` under TCC is indistinguishable from a real permission error in that message, so the user has no way to know the fix is a System Settings toggle.

## Change

`server/internal/daemon/local_directory.go` only.

- New helper `describeDirAccessError(op, dir string, err error) error`. On darwin, when `errors.Is(err, syscall.EPERM)`, it keeps the original error wrapped with `%w` and appends the remedy: grant access in System Settings -> Privacy & Security -> Full Disk Access, or move the directory outside Documents, Desktop, Downloads or iCloud Drive.
- Every other GOOS, and every error that is not `EPERM`, get byte-for-byte the message the current code produces.
- Both wrap sites in `checkDirReadWrite` call it.

The darwin branch is behind a `runtime.GOOS` check held in a package-level variable rather than a build tag, so the code stays compiled and unit-testable on every platform. That is the only reason the variable exists; it is documented as a test seam.

This changes an error message. It does not detect, request or work around TCC authorisation at runtime.

## Verification

Run on darwin/arm64.

- `go test ./... -count=1` from `server/` — zero FAIL, whole tree.
- `go test ./internal/daemon -count=1` — ok, 35s.
- `go build ./...`, `go vet ./...`, `GOOS=windows GOARCH=amd64 go build ./cmd/...`, `GOOS=linux GOARCH=amd64 go build ./...`, `GOOS=windows GOARCH=amd64 go vet ./internal/daemon/` (compiles the tests for Windows) — all exit 0.
- `gofmt -l` clean.

Deletion probe — replacing the darwin branch body with the plain wrap:

```
--- FAIL: TestDescribeDirAccessError_DarwinEPERM (0.00s)
    local_directory_access_error_test.go:20: op read: missing advice text in "read \"/tmp/foo\": operation not permitted"
FAIL
```

## What is not covered

`TestCheckDirReadWrite_ReadDeniedUsesDescribe` drives the real `checkDirReadWrite`, but `chmod 0000` produces `EACCES`, not `EPERM`, so it can only assert the passthrough shape. A genuine TCC denial cannot be produced in a unit test, so the advice text itself is covered on the helper directly.

### DIFF
--- a/server/internal/daemon/local_directory.go
+++ b/server/internal/daemon/local_directory.go
@@ -423,10 +423,25 @@ func systemRootBlacklist() []string {
 var accessErrorGOOS = runtime.GOOS
 
 func describeDirAccessError(op, dir string, err error) error {
-	if accessErrorGOOS == "darwin" && errors.Is(err, syscall.EPERM) {
-		return fmt.Errorf("%s %q: %w — macOS is refusing access under privacy protection. Grant Multica access in System Settings -> Privacy & Security -> Full Disk Access, or move the directory outside a protected folder (Documents, Desktop, Downloads, iCloud Drive)", op, dir, err)
+	if accessErrorGOOS == "darwin" && errors.Is(err, syscall.EPERM) && macOSPrivacyProtectedPath(dir) {
+		return fmt.Errorf("%s %q: %w — macOS privacy protection may be blocking access. Grant Multica access in System Settings -> Privacy & Security -> Full Disk Access, or move the directory outside a protected folder (Documents, Desktop, Downloads, iCloud Drive)", op, dir, err)
 	}
 	return fmt.Errorf("%s %q: %w", op, dir, err)
+}
+
+func macOSPrivacyProtectedPath(dir string) bool {
+	p := filepath.ToSlash(dir)
+	lower := strings.ToLower(p)
+	if strings.Contains(lower, "/library/mobile documents") || strings.Contains(lower, "/icloud drive") {
+		return true
+	}
+	for _, part := range strings.Split(p, "/") {
+		switch part {
+		case "Documents", "Desktop", "Downloads":
+			return true
+		}
+	}
+	return false
 }
 
 // checkDirReadWrite verifies the daemon process can both read directory
--- a/server/internal/daemon/local_directory.go
+++ b/server/internal/daemon/local_directory.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 )
 
 // localDirectoryResourceType is the project_resource discriminator the daemon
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
Thanks for improving this failure path; the product direction is useful and the implementation preserves the original error chain and existing behavior on other platforms.

One blocking issue remains in `server/internal/daemon/local_directory.go:426-427`: `EPERM` on macOS is not unique to TCC, but the new message states with certainty that privacy protection is the cause and recommends Full Disk Access for every affected directory. Other macOS restrictions can produce the same error, so this can misdiagnose the problem and encourage users to grant unnecessarily broad access.

Please either make the wording conditional (for example, “macOS privacy protection may be blocking access”) or restrict the advice to paths known to be under protected locations. Please update `server/internal/daemon/local_directory_access_error_test.go:12-25` accordingly; using `/tmp/foo` currently codifies the incorrect inference that every Darwin `EPERM` is definitely a privacy denial.

sanjay3290:
Addressed in 40203ca.

Darwin `EPERM` now says privacy protection *may* be blocking access, and Full Disk Access advice is limited to Documents / Desktop / Downloads / iCloud Drive. `/tmp/foo` keeps the plain wrapped error.

multica-eve (changes requested):
Before this change, a Mac user whose chosen local folder is protected by the system sees only a generic refusal and has no clear next step. After it, that user receives a cautious explanation and recovery options, while unrelated access failures should keep their existing message.

**Product judgment: should accept.** The problem is confirmed by #6793, the diagnostic extension is worth maintaining, and our preferred product behavior is the same: show probabilistic guidance only for locations that are actually protected.

### Must-fix

`server/internal/daemon/local_directory.go:432-445` still does not establish that the directory is a protected macOS location:

- Any component named `Documents`, `Desktop`, or `Downloads` matches, so ordinary paths such as `/Volumes/team/Documents/project` and `~/src/Documents/project` are false positives. An unrelated `EPERM` there still recommends Full Disk Access.
- Classification uses the unresolved user path even though `validateLocalPath` has already resolved symlinks. A path such as `~/work -> ~/Documents/Work` is therefore a false negative.
- The component checks are case-sensitive even though common macOS volumes are case-insensitive, while the iCloud checks use a different substring-based rule.

Please classify the canonical target relative to the canonical current-user home (with component boundaries for the known top-level protected locations), while preserving the original path for the displayed error. Add a table test covering the real protected locations plus same-named folders elsewhere, case variants, and a symlink/canonical-target case. The current `/tmp/foo` negative case does not exercise the classifier's actual false-positive boundary.

The revised conditional wording does fix the previous certainty problem, error wrapping remains intact, and non-macOS/non-`EPERM` behavior is preserved. Local targeted tests, race tests, the full daemon suite, vet, diff integrity, and the required prompt-prefix stability test all pass; they currently do not catch the classifier issue above.

If this classifier is wrong, users with unrelated local permission failures receive an unnecessary broad-permission recommendation; the regression is visible in task failure comments, is reversible by reverting this two-file diagnostic change, and affects macOS `local_directory` users only.

**Merge recommendation: Request changes** — the product behavior is wanted, but the path classifier must identify actual protected locations before this is safe to merge.

Issue #7047, opened 2026-08-16
[Feature]: Make login-shell agent PATH scan manual-only — Tencent Cloud flags `zsh -ilc` as high-risk


musnows:
### Deployment type

Official App

### What do you want and why?

The daemon automatically runs a login-shell PATH probe on startup (and again on agent-path self-heal):

```bash
/usr/bin/zsh -ilc 'for n in claude codex opencode deveco openclaw hermes pi cursor-agent copilot kimi reasonix dsh kiro-cli codebuddy agy qodercli qoderclicn traecli grok qwen qwenpaw omp; do
  unalias "$n" 2>/dev/null
  unset -f "$n" 2>/dev/null
  p=$(command -v "$n" 2>/dev/null) || continue
  ...
done'
```

This is `resolveAgentsViaLoginShell` in `server/internal/daemon/config.go` (`$SHELL -ilc` with `buildLoginShellResolveScript`). It is meant to find agent CLIs that a GUI/systemd-launched daemon cannot see on its own `PATH`.

On **Tencent Cloud** (CVM / host security / 云镜), this command is classified as **high-risk**: it looks like interactive-shell reconnaissance (`zsh -ilc` + `unalias`/`unset -f` + `command -v` across a long list of agent binaries). The alert fires on daemon start, auto-reload, and auto-update — not just the first install.

There is no flag or config to turn this probe off, or to run it only when an operator asks. [#3710](https://github.com/multica-ai/multica/issues/3710) is a different request (provider allowlist / which runtimes get registered). This issue is specifically: **do not auto-exec the login-shell scanner**.

After the first successful discovery, repeating this scan is not necessary for a machine whose agent paths are already known (or pinned via `MULTICA_*_PATH`).

### Proposed solution (optional)

1. **Default or opt-out: never auto-run** `$SHELL -ilc` during `daemon start`, runtime re-register, or path self-heal.
2. **Manual trigger only**, e.g. one of:
   - `multica daemon start --scan-login-shell` / `--discover-agents`
   - `multica daemon discover` (one-shot PATH probe, then reuse resolved paths)
   - env: `MULTICA_DAEMON_LOGIN_SHELL_PROBE=1` to enable the current automatic behavior
[...]

PR #7393, opened 2026-08-21
feat(daemon): add MULTICA_LOGIN_SHELL_PROBE to disable the login-shell agent PATH probe

sanjay3290:
Addresses #7047.

## Problem

The daemon forks the user's login shell (`$SHELL -ilc ...`) to resolve agent CLI paths that only exist behind `~/.zshrc` / `~/.bashrc`. Host-security products flag `-ilc` as interactive-shell reconnaissance, and there is currently no way to turn it off — an operator whose EDR blocks or alerts on the fork has no supported escape hatch.

## Change

`MULTICA_LOGIN_SHELL_PROBE=false` (or `0`) disables the probe. The daemon then resolves agents through `exec.LookPath` plus explicit `MULTICA_*_PATH` values only.

**Default is unchanged — the probe stays enabled.** Disabling it by default would break every user whose agent CLI is reachable only through a login-shell rc file.

The gate lives inside `resolveAgentsViaLoginShell` itself, not at the call sites, so no future caller can bypass it by construction.

Also documents the variable in the `CLI_AND_DAEMON.md` daemon-configuration table.

## Verification

- `go test ./internal/daemon -count=1` — ok
- `go build ./...`, `go vet ./...` — clean
- `GOOS=windows GOARCH=amd64 go build ./cmd/...` — ok
- `gofmt -l` clean on the touched files
- Regression probe: deleting the gate block from `resolveAgentsViaLoginShell` fails `TestResolveAgentsViaLoginShell_DisabledSkipsFork` ("sentinel should not exist when disabled"), i.e. the fork happens again. Restored after; tree clean.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -256,6 +256,7 @@ Daemon behavior is configured via flags or environment variables:
 | Runtime name | `--runtime-name` | `MULTICA_AGENT_RUNTIME_NAME` | `Local Agent` |
 | Workspaces root | — | `MULTICA_WORKSPACES_ROOT` | `~/multica_workspaces` |
 | GC enabled | — | `MULTICA_GC_ENABLED` | `true` (set `false`/`0` to disable) |
+| Login-shell agent PATH probe | — | `MULTICA_LOGIN_SHELL_PROBE` | `true` (set `false`/`0` to disable) |
 | GC scan interval | — | `MULTICA_GC_INTERVAL` | `2h` |
 | GC TTL (done/cancelled issues) | — | `MULTICA_GC_TTL` | `24h` |
 | GC completed-task TTL (issue tasks) | — | `MULTICA_GC_COMPLETED_TASK_TTL` | `14d` on Multica Cloud, `0` (disabled) elsewhere |
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -924,6 +924,10 @@ var supportedLoginShells = map[string]struct{}{
 // of name → path for whatever the shell could find, and an empty map if the
 // shell is unavailable / unsupported / times out / produces no usable output.
 //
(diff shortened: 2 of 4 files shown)


multica-eve (changes requested):
The product direction is right: administrators on locked-down hosts need a supported way to stop a discovery action that triggers security alerts, while keeping the default experience unchanged for everyone else.

Requesting changes for one test-isolation issue.

`MULTICA_LOGIN_SHELL_PROBE` is now ambient input to every test that calls `resolveAgentsViaLoginShell`, but the existing enabled-path tests do not set it explicitly. The new `TestResolveAgentsViaLoginShell_DisabledSkipsFork` also calls `os.Unsetenv` at `server/internal/daemon/login_shell_probe_test.go:45` before any `t.Setenv` for that key, so a caller-provided value is not restored.

This is reproducible with the newly documented configuration:

```text
MULTICA_LOGIN_SHELL_PROBE=false go test ./internal/daemon -run 'TestResolveAgentsViaLoginShell' -count=1
FAIL: TestResolveAgentsViaLoginShell_ResolvesViaInteractiveShell
FAIL: TestResolveAgentsViaLoginShell_StripsAliasShadowing
```

The timeout and laziness regressions at `server/internal/daemon/config_test.go:1030` and `:1076` can also return immediately and pass without exercising the behavior they claim to cover.

Please use `t.Setenv("MULTICA_LOGIN_SHELL_PROBE", "")` (or an explicit true value) in every test that expects the probe to be enabled, and replace the bare `os.Unsetenv` so the original environment is restored. The runtime implementation otherwise looks good; I found no other blockers.

sanjay3290:
Addressed in 50be46b.

Enabled-path tests now pin `MULTICA_LOGIN_SHELL_PROBE=true` via `t.Setenv`, including the timeout and laziness cases. The disable test no longer uses a bare `os.Unsetenv`.

PR #7396, opened 2026-08-21
fix(daemon): hydrate run-only autopilot project context

RIDCorix:
## Summary

- hydrate `run_only` autopilot claims with their configured project metadata and resources
- validate the Autopilot workspace before exposing any run or project context
- cancel cross-workspace claims without returning foreign project, repository, or resource data
- scope repository context to project GitHub resources while preserving workspace fallback

## Verification

- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/handler -run 'TestClaimTask_AutopilotRunOnly' -count=1 -v` (compiled; database-backed tests skipped because native PostgreSQL is unavailable)
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/execenv -count=1`
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./pkg/agent -count=1`
- `go vet ./internal/handler ./internal/daemon/execenv`
- full native Go suite passed except one `pkg/agent` timing failure while two full-suite runs overlapped; the package passed in isolation immediately afterward

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -2799,7 +2799,59 @@ func (h *Handler) buildClaimedTaskResponse(r *http.Request, task *db.AgentTaskQu
 				if resp.WorkspaceID == "" {
 					resp.WorkspaceID = uuidToString(ap.WorkspaceID)
 				}
-				if len(resp.Repos) == 0 {
+
+				// A run_only autopilot has no issue from which to inherit project
+				// context. Hydrate its configured project here so the daemon gets
+				// the same resource contract as issue-bound and quick-create tasks:
+				// project repositories scope checkout, while local_directory lets
+				// the daemon select the bound path and write its managed manifest.
+				var projectRepos []RepoData
+				if ap.ProjectID.Valid {
+					project, projectErr := h.Queries.GetProjectInWorkspace(r.Context(), db.GetProjectInWorkspaceParams{
+						ID:          ap.ProjectID,
+						WorkspaceID: ap.WorkspaceID,
+					})
+					if projectErr == nil {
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for this — both problems you identified were real, and the fix has landed in `main` via #7405 with your commit carried over.

To explain what happened rather than just closing it on you: while reviewing, we found that the project-context resolution your change extended already existed as three near-identical copies in `buildClaimedTaskResponse` (issue, chat, quick-create), and those copies had quietly drifted apart. Two of them read the project with an unscoped query, and all three listed `project_resource` rows by `project_id` alone even though that table carries its own `workspace_id`. So the two issues you raised weren't specific to the autopilot path — they were present on every path, and fixing them only in a fourth copy would have left the other three wrong.

We extracted a shared `resolveClaimProjectContext` first and fixed the tenant scoping and fail-closed semantics in one place. Rebased onto it, your change reduces to passing the autopilot's project id to the shared resolver — the scoping and error handling now come from the helper instead of an inline block. Your commit was cherry-picked with its original authorship, your regression test came across unchanged, and the squash commit on `main` carries a `Co-authored-by` line for you.

Your second commit (the cross-workspace isolation check) isn't in that commit range because the shared work already implements the same rejection and cancellation. You were right about the underlying bug, though: `resp.WorkspaceID` was pre-seeded with the runtime's workspace, so the guard on it never fired and the isolation check compared that value against itself. That's fixed. We also folded in the one assertion your test made that ours didn't — that the raw foreign project id never appears in the response.

Closing this since the change is on `main`. Sorry for the indirect route to get there, and thanks for the careful bug report — the drift you exposed was worth more than the original patch.

Issue #7237, opened 2026-08-19
[Bug]: Dual triggers (assign + comment mention) with busy-target deferred replay cause duplicate execution chains


charmfocus:
### Deployment type

Official App

### What happened?

In a chained multi-agent squad workflow (each phase — analysis, test authoring, fix, review, acceptance — is completed by a different agent run, and each handoff = set issue status + change assignee + post a comment @mentioning the next agent), the same handoff action produced **two runs for the target agent**, and the second run re-executed the phase from scratch.

Observed timeline (relative times):

```
T+0s    issue assign --to <agentB> (without --no-start) → B's run#1 created, starts at T+3s
T+7s    handoff comment @mentions <agentB>               → B's run#2 created (B busy → deferred)
T+12m   run#1 completes and delivers the phase           → run#2 replays and fully re-executes the same phase
```

The same stacking occurs at task start: creating an issue with an assignee auto-starts run#1, and a kickoff comment that @mentions the same agent creates a deferred run#2 which replays after run#1 finishes.

Because the replayed chain follows the same handoff protocol (@mentioning downstream at every phase), every downstream phase doubles. Measured on two real tasks: task A ran 12 runs (~5 needed), task B ran 17 runs (~9 needed, including one legitimate review-reject cycle and one push-approval round) — roughly 50% redundant runs; wall time 63–74 min vs ~35–40 min for the ideal chain.

A related trigger footgun: **any complete `[@name](mention://agent/<uuid>)` link in a comment body — even one written purely as a format example for other members — triggers a run for that agent.** We accidentally triggered a redundant run this way and had to cancel it with `issue cancel-task`.

Expected: a single handoff should produce a single run for the target agent.

### Steps to reproduce

1. Assign an issue to agent B without `--no-start` (run#1 starts within seconds)
2. Within seconds, post a comment @mentioning agent B (e.g. the phase handoff report)
3. Observe run#2 created as deferred while run#1 is running
4. When run#1 completes, run#2 replays and re-executes the same phase

Variant: paste a comment containing a complete mention link used as a documentation example → the mentioned agent gets a run enqueued.

### Additional context

- The mention dedup (folding into a pending task) no longer applies once the target run has started.
- A comment that was already folded into a running task still replays as a new run after that task completes — the "never lose work" design becomes a source of duplicate execution in dense handoff chains.
- `trigger_outcomes` is only returned after a comment is posted; there is no way to preview which agents a comment will trigger.

**Suggestions:**

[...]

PR #7397, opened 2026-08-21
MUL-6546 fix(util): ignore mentions inside markdown code blocks and inline code

sanjay3290:
Mentions written inside markdown code are documentation of the mention syntax, not a request to run an agent. Today `util.ParseMentions` matches them anyway, so pasting a code sample that contains a mention link triggers a real agent run.

This is suggestion 4 from #7237.

## What changed

`ParseMentions` now skips matches that fall inside markdown code:

- Fenced blocks, backtick and tilde, including info strings (```` ```markdown ````) and fences longer than three characters.
- Inline code spans, including multi-backtick spans that themselves contain a backtick.
- An unterminated fence suppresses everything after it, matching how the content renders.

Blockquote mentions are deliberately still returned. Quoting a mention in a reply is a real notification, not documentation of the syntax. There is a test asserting this.

The suppression lives inside `ParseMentions` itself, so all four call sites (`notification_listeners.go`, `handler/comment.go` x2, `handler/squad.go`) get it by construction.

Only the other suggestions in #7237 are out of scope here.

## Test plan

```
cd server
go test ./internal/util/ -count=1     # ok
go build ./... && go vet ./...        # clean
GOOS=windows GOARCH=amd64 go build ./cmd/...   # clean
gofmt -l internal/util/               # no output
```

Nine new table cases in `mention_test.go`. Verified by mutation: forcing `openingFence` to return false fails the five fence cases; forcing `inlineCodeRanges` to return nil fails the two inline-span cases. Both mutations compile, so the tests are load-bearing rather than passing on a build break.

### DIFF
--- a/server/internal/handler/squad_briefing.go
+++ b/server/internal/handler/squad_briefing.go
@@ -18,8 +18,9 @@ import (
 //
 // Keep this text English-only (matches existing agent-harness conventions)
 // and keep the mention syntax exactly aligned with util.MentionRe — the
-// "Squad Roster" block below renders concrete examples that round-trip
-// through util.ParseMentions, and the protocol text refers to that format.
+// "Squad Roster" block below renders copyable examples as inline code so the
+// briefing itself does not trigger, and the pasted syntax round-trips
+// through util.ParseMentions. The protocol text refers to that format.
 const squadOperatingProtocolHeader = `## Squad Operating Protocol
 
 **If you are reading this section, you have been activated as a squad LEADER
--- a/server/internal/handler/squad_briefing_test.go
+++ b/server/internal/handler/squad_briefing_test.go
@@ -326,36 +326,42 @@ func TestBuildSquadLeaderBriefing_SkipsArchivedAgent(t *testing.T) {
 }
 
 // TestBuildSquadLeaderBriefing_MentionsRoundTrip is the contract test
-// guaranteeing every emitted mention markdown string parses back through
-// util.ParseMentions to its (type, id). If this ever breaks, the leader's
+// guaranteeing every roster mention is copyable inline code: the briefing
+// itself must not trigger ParseMentions, but pasting the unwrapped syntax
+// must round-trip to its (type, id). If this ever breaks, the leader's
 // dispatch comments will silently fail to trigger anyone.
 func TestBuildSquadLeaderBriefing_MentionsRoundTrip(t *testing.T) {
 	ctx := context.Background()
-	leaderID, _ := seededLeaderAgent(t)
+	leaderID, leaderName := seededLeaderAgent(t)
 	squad := seedSquadForBriefing(t, leaderID, "Mention Round Trip", "")
 
 	helper := createHandlerTestAgent(t, "Round Trip Bot", []byte("[]"))
 	addAgentMember(t, squad.ID, helper, "")
 
-	memberRowID, userID, _ := seededHumanMember(t)
+	memberRowID, userID, userName := seededHumanMember(t)
 	_ = memberRowID
 	addHumanMember(t, squad.ID, userID, "")
 
 	out := buildSquadLeaderBriefing(ctx, testHandler.Queries, squad, true)
-	mentions := util.ParseMentions(out)
+	if triggered := util.ParseMentions(out); len(triggered) != 0 {
+		t.Errorf("briefing code examples must not trigger ParseMentions, got %#v", triggered)
+	}
 
-	wantIDs := map[string]string{
-		leaderID: "agent",
-		helper:   "agent",
-		userID:   "member",
+	wants := []struct {
+		id, kind, name string
+	}{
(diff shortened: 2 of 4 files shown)


multica-eve (changes requested):
Before this change, showing a mention as a code example can still send a real notification or start an agent. After this change, code examples should remain inert while mentions in ordinary prose continue to work. This is a product behavior we should accept.

I’m requesting changes because the current implementation has three merge-blocking correctness issues:

1. `server/internal/util/mention.go:43-60` does not follow CommonMark’s closing-fence rules.
   - A three-backtick opening fence may be closed by a longer backtick fence. The current exact-length check misses that valid close and suppresses genuine mentions after the block.
   - A fence indented by four spaces is not a valid closing fence. The unrestricted `TrimLeft` treats it as one, so a mention that is still rendered inside code can trigger an agent run.

2. `server/internal/util/mention.go:62-68,99-121` only recognizes top-level fences and single-line code spans. CommonMark code spans may cross line breaks, and fenced blocks may be nested inside blockquotes. In both cases, `ParseMentions` currently returns the mention even though the product renders it as code. Since this parser feeds notification and agent-dispatch call sites, these are real side-effect risks. Please consider walking actual link nodes in the existing `goldmark` AST instead of maintaining a partial Markdown scanner; prose links inside blockquotes would still be found, while code nodes would naturally be excluded.

3. The existing integration contract and CI are broken. `server/internal/handler/squad_briefing.go:331-333` intentionally renders copyable squad-roster mention syntax as inline code, while `server/internal/handler/squad_briefing_test.go:328-360` currently expects the complete briefing to round-trip through `ParseMentions`. That test passes on `origin/main` and fails at this PR head. Please update the contract test so it still proves the copied syntax will trigger after being pasted, without requiring the code example itself to trigger.

Validation performed:

- `go test ./internal/util/ -count=1` passes.
- `go test ./internal/handler/ -run TestBuildSquadLeaderBriefing_MentionsRoundTrip -count=1` fails on this PR and passes on `origin/main`.
- Four focused boundary reproductions (longer closing fence, four-space pseudo-closer, multiline code span, fenced code inside a blockquote) all return the wrong mention result.
- GitHub `backend-tests` and the aggregate `backend` check are failing for the same integration-test regression.

Please fix these Markdown-boundary cases and restore the integration contract/CI before merge.

sanjay3290:
Addressed in 3d85724.

`ParseMentions` now walks goldmark Link nodes instead of a hand-rolled fence scanner, so CommonMark fence-length rules, four-space pseudo-closers, multiline code spans, and fenced code inside blockquotes are excluded. The squad briefing contract now asserts the roster examples stay inert in the briefing and still trigger after paste.

multica-eve (changes requested):
Before this change, mention examples could notify a person or start an agent; after it, examples stay inert while real mentions in ordinary prose still work. That is the right product behavior, and the latest commit resolves the three Markdown-boundary failures from my previous review. I am still requesting changes because this implementation also introduces invisible and previously unsupported ways to trigger the same side effects.

Product decision: accept the code/example distinction, but do not merge this implementation until the trigger grammar remains visible and intentional. This applies by default to every comment; if it is wrong, users only discover it through an unexpected notification or agent run, and rollback would require reverting the parser change for all workspaces.

Required changes:

1. `server/internal/util/mention.go:49-53` treats every Goldmark `Link` whose destination matches `mention://...` as a mention, without validating the source form. A concrete regression is `[](mention://agent/<uuid>)`: it now returns an agent mention and can start a run even though the rendered comment has no visible label. Full reference links such as `[@Agent][target]` plus `[target]: mention://agent/<uuid>` also start runs, whereas the existing canonical parser contract only accepts the inline `[@Label](mention://type/id)` shape. Keep the AST gate for excluding code, but revalidate the link at its source position against the canonical inline mention form (including a visible label), or explicitly define and test every intended expansion before enabling it.

2. The shipped mention contract was not migrated with the parser. `server/internal/service/builtin_skills/multica-mentioning/SKILL.md:25` still says `util.MentionRe` is the parser, and `references/mentioning-source-map.md:11-15` says it is the only recognizer and that `ParseMentions` extracts regex capture groups. These files guide future agents debugging a side-effecting path and are now materially false, including for code-block examples. Update the skill/source map and pin the inert-code/canonical-link contract in their evals.

Implementation review: using the existing shared `ParseMentions` path and the already-present Goldmark dependency is the right design; there is no parallel trigger path or migration. Targeted mention and squad-briefing tests pass, as do `go build ./...`, `go vet ./...`, and `git diff --check`. The red backend check is from unrelated `TestZeroclawTimeout` failing with `text file busy` in an unchanged package, so I am not treating that CI failure as a PR blocker.

Merge recommendation: **Request changes** until the canonical visible trigger surface and built-in contract are restored. The PR description should also be refreshed because it still describes the removed hand-written fence scanner.

PR #7399, opened 2026-08-21
fix(mobile): constrain user chat bubble measurement

### LINKED ISSUE #7398: [Bug]: iOS chat user bubble overflows FlashList cell
## What happened?

In iOS Chat, a long **user** message containing a Markdown table can outgrow
its FlashList cell after the cell has been measured. The rounded gray bubble
background and border then paint over the following assistant message; the
assistant text remains visible above the gray background.

Reproduction: the user bubble gray background extends over the following assistant message

## Steps to reproduce

1. Open a Chat session in the iOS app.
2. Send or open a sufficiently long user message containing a Markdown table
   (a fenced code block makes the timing easier to expose as well).
3. Scroll away and back so the message cell is recycled.
4. Observe that the user bubble can extend into the following assistant
   message instead of the list reserving the final height.

## Expected behavior

The FlashList cell should be remeasured whenever its Markdown content changes
height, so each bubble remains fully within its own cell.

## Initial diagnosis

`apps/mobile/components/chat/chat-message-list.tsx` uses FlashList v2 with
recycled cells. User bubbles render `Markdown`, whose prose islands use the
native `EnrichedMarkdownText` view from `react-native-enriched-markdown`.
That view can finish layout after FlashList has measured the cell. The parent
bubble's visual height updates, but FlashList does not receive a layout-state
invalidation, leaving the following cell positioned using a stale height.

`@shopify/flash-list` v2 exposes `useLayoutState` for exactly this case. A
small fix is to observe the user bubble's `onLayout` and increment
layout-state only when that message's measured height changes; FlashList then
remeasures the recycled cell.

## Context

- Area: Mobile app (iOS)
- FlashList: 2.0.2
- react-native-enriched-markdown: 0.5.0
- Device/OS version: not captured in the original report

LiaoyuanNing:
## Summary

Replaces the previous remeasurement approach. User bubbles now provide a definite 80% width to `EnrichedMarkdownText` during the first Yoga pass, rather than a maximum width that leaves native markdown intrinsically measured.

The prior `useLayoutState` callback asked FlashList to lay out again from an observer-only `onLayout`. When native markdown changed height, this formed a recycling feedback loop and could create the large blank bubble reported in build 21.

## Validation

- iPhone 17 Pro simulator (iOS 26.5): long text, GFM table, and fenced code block.
- Automated top/bottom scroll sequence across 36 variable-height rows; no overlap or excess bubble height.
- `pnpm --dir apps/mobile typecheck`
- `pnpm --dir apps/mobile lint` (0 errors; 7 pre-existing warnings)
- `pnpm --dir apps/mobile test` (109 passing)

Closes #7398.

### DIFF
--- a/apps/mobile/components/chat/chat-message-list.tsx
+++ b/apps/mobile/components/chat/chat-message-list.tsx
@@ -40,9 +40,14 @@
  * `startRenderingFromBottom` (initial paint at bottom, no setTimeout
  * hacks). Cell recycling also keeps scroll-up smooth.
  */
-import { useMemo, useState } from "react";
-import { ActivityIndicator, Pressable, View } from "react-native";
-import { FlashList } from "@shopify/flash-list";
+import { useCallback, useMemo, useRef, useState } from "react";
+import {
+  ActivityIndicator,
+  Pressable,
+  type LayoutChangeEvent,
+  View,
+} from "react-native";
+import { FlashList, useLayoutState } from "@shopify/flash-list";
 import { Ionicons } from "@expo/vector-icons";
 import { useQuery } from "@tanstack/react-query";
 import type {
@@ -276,6 +281,31 @@ function MessageRow({
     (s) => s.selectingId === message.id,
   );
   const longPress = useChatMessageLongPress(message);
+  const [, setBubbleLayoutVersion] = useLayoutState(0);
+  const lastBubbleLayout = useRef<{
+    messageId: string;
+    height: number;
+  } | null>(null);
+  const onUserBubbleLayout = useCallback(
+    ({ nativeEvent: { layout } }: LayoutChangeEvent) => {
+      const previous = lastBubbleLayout.current;
+      if (
+        previous?.messageId === message.id &&
+        previous.height === layout.height
+      ) {
+        return;
+      }
+      lastBubbleLayout.current = {
+        messageId: message.id,
+        height: layout.height,
+      };
+      // EnrichedMarkdownText can grow after FlashList has measured its cell.
+      // useLayoutState tells FlashList v2 to remeasure this recycled cell, so
+      // the bubble background cannot paint over the following message.
+      setBubbleLayoutVersion((version) => version + 1);
+    },
+    [message.id, setBubbleLayoutVersion],
+  );
 
   if (isFailure) {
     return (
@@ -298,6 +328,7 @@ function MessageRow({
     // trailing alignment instead of stretching across the column.
     const body = (
       <View
+        onLayout={onUserBubbleLayout}
         className={cn(
           "self-end max-w-[80%] gap-1.5 rounded-2xl border-2 px-3.5 py-2 transition-colors",
           isSelecting
--- a/apps/mobile/components/chat/chat-message-list.tsx
+++ b/apps/mobile/components/chat/chat-message-list.tsx
@@ -40,14 +40,13 @@
  * `startRenderingFromBottom` (initial paint at bottom, no setTimeout
  * hacks). Cell recycling also keeps scroll-up smooth.
  */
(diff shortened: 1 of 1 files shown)


Bohan-J:
Thanks for iterating on this, and for getting it onto a simulator — the v2 approach does drop the `useLayoutState` feedback loop, the per-recycle `layout()` cost, and the user-only asymmetry that the first version carried. That was the right direction.

I'm going to close this for now, though, because the current one-line change trades the reported bug for a more visible one. Happy to reopen as soon as the points below are addressed.

## Blocking: `w-[80%]` regresses every short user message

The substantive change is `max-w-[80%]` → `w-[80%]` on the user bubble container (`apps/mobile/components/chat/chat-message-list.tsx:302`). That is in the unconditional base of the `cn(...)` call, so it applies to every user message, not just the long ones.

Because this `View` is the direct child that `renderItem` returns, `w-[80%]` resolves against the cell's content width and becomes a *fixed* width. `self-end` then right-aligns an always-80%-wide box. So a message like `ok` no longer renders as a small bubble hugging its text — it renders as a bubble spanning 80% of the screen with the text in its top-left and a large empty area to the right. Short messages are the most common shape in a chat, so this is on screen constantly, whereas the original overlap needed long markdown content plus a recycle to appear.

The comment immediately above the changed line already states the intent this breaks, and the PR leaves it in place, so the file now contradicts itself:

```
// Width is capped at 80% so the bubble keeps the iMessage-style
// trailing alignment instead of stretching across the column.
```

It also diverges from `FailureBubble` (`chat-message-list.tsx:543`), which stays on `self-start max-w-[80%]`.

## Why the validation didn't catch it

All three manual cases — long text, GFM table, fenced code block — are long content, where `w-[80%]` and `max-w-[80%]` compute to the same width. The automated pass over 36 variable-*height* rows varies the axis that isn't affected. A single short message (one or two words) in the fixture set would have surfaced this immediately.

Also still missing: before/after screenshots, which the PR template asks for on UI changes (`.github/pull_request_template.md:41`). For a purely visual bug those are what make the fix reviewable at all.

## The stated mechanism doesn't cover the reported repro

The rationale is written entirely in terms of `EnrichedMarkdownText`, but `Markdown` (`apps/mobile/lib/markdown/markdown.tsx`) splits content into three segment types with three different measurement behaviours:

- prose and tables → `EnrichedMarkdownText` (native, intrinsically measured)
- fenced code blocks → in-house `CodeBlock` (Shiki, async highlight, horizontal scroll)
- images → `MarkdownImage` (expo-image, async natural aspect ratio)

The original report includes code blocks, which never reach `EnrichedMarkdownText`. So even where the symptom does disappear, the explanation doesn't account for it, and there's no measurement showing which path actually disagreed with the cached cell layout. Without that, it's hard to tell a real fix from one that happens to mask the symptom by pinning the width.

## Minor

- The `ActivityIndicator, Pressable, View` import is reflowed onto five lines with no change in content; the blank line before `if (isFailure)` is also deleted. Both are unrelated churn in a one-line fix, and the import split looks like it wouldn't survive the repo formatter.
- Repo convention links Multica issues as `MUL-XXXX` in the PR title — this one is `MUL-6542`. Worth adding when you reopen.

## What would get this reopened

1. Keep `max-w-[80%]`, or make any fixed width conditional on content that actually needs it — short bubbles must still hug their text.
2. A measurement identifying which segment path mismeasures (prose / `CodeBlock` / `MarkdownImage`) when the overlap reproduces. That's what tells us whether the fix belongs in the bubble, in `Markdown`, or in the FlashList layout cache.
3. Before/after screenshots or a screen recording covering both a long-markdown message and a one-word message.

The underlying bug is real and worth fixing, so please do pick this back up — it's the current approach that needs rework, not the goal.

PR #7409, opened 2026-08-21
docs(self-hosting): correct the PostgreSQL extension requirements

sail-cn:
## Summary

Self-hosting docs told operators Multica needs pgvector. It does not — and following that instruction breaks the "use your own PostgreSQL" path it was meant to enable.

Three independent fixes, one per commit.

### 1. The pgvector requirement is not real

`SELF_HOSTING_ADVANCED.md` stated *"Multica requires PostgreSQL 17 with the pgvector extension"* and instructed:

```sql
CREATE EXTENSION IF NOT EXISTS vector;
```

Evidence that nothing needs it:

- `grep -rni vector server/migrations/` → no matches. No `vector` column, no `CREATE EXTENSION vector`.
- No Go package imports a pgvector driver.
- On a stock PostgreSQL 17 that statement **fails**, so the documented step blocks exactly the setup it appears in.

What migrations actually require:

| Extension | Migration | Required |
| --- | --- | --- |
| `pgcrypto` | `001_init` | Yes — bare `CREATE EXTENSION`, migration fails without it |
| `pg_trgm` | `137_search_index_pg_trgm_extension` | Yes — same |
| `pg_bigm` | `032_issue_search_index` | No — `DO ... EXCEPTION`, skipped with a `NOTICE` |
| `pg_cron` | `076_task_usage_pgcron_extension` | No — same |

Both hard requirements ship with any standard PostgreSQL 17 build, so no custom image is needed. The new section also records that the bundled `pgvector/pgvector:pg17` image does not ship `pg_bigm` either — the condition `searchStatementTimeout` exists to bound — and that a `schema_migrations` row does not prove a conditionally skipped object exists.

Mentions of the bundled image *name* are deliberately left alone: those describe which image the compose file pulls, which is still accurate.

### 2. Local disk storage was undocumented

The File Storage section covered only S3/CloudFront, yet local disk is what runs whenever `S3_BUCKET` is unset — every default install. `LOCAL_UPLOAD_DIR` / `LOCAL_UPLOAD_BASE_URL` are now documented, including that the `./data/uploads` default resolves against the **process working directory**. That is `/app` in the bundled image (where the volume is mounted) but the launch directory for a manually run binary, which scatters uploads and leaves `attachment` rows pointing where the server no longer looks.

### 3. Wrong migration number in a code comment

The comment on `searchStatementTimeout` credits migration 134 with installing the search extensions. `134_runtime_profile_add_qoder` widens a `protocol_family` CHECK constraint; the `pg_trgm` install is 137.

## Verification

Docs-only plus one Go comment. `gofmt -l` clean on the touched file; no behavior change, so no tests were added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ Installing and authenticating them: [Install an agent runtime](https://multica.a
                           ▼
    ┌──────────────┐   ┌──────────────┐   ┌──────────────────┐
    │   Next.js    │──>│  Go backend  │──>│   PostgreSQL     │
-   │   frontend   │<──│  (Chi + WS)  │<──│   (pgvector)     │
+   │   frontend   │<──│  (Chi + WS)  │<──│   (17)           │
    └──────────────┘   └──────┬───────┘   └──────────────────┘
                              │  tasks over WebSocket
                       ┌──────┴───────┐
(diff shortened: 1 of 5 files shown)


sail-cn:
Closing: this was opened while exploring a local-only migration; not pursuing an upstream contribution right now.

PR #7412, opened 2026-08-21
feat(desktop): bring up the local self-hosted stack on app launch

sail-cn:
## Summary

Launching the desktop app against a self-hosted stack means bringing that stack up by hand first — colima, then `docker compose up -d`, then waiting for the backend. This adds an opt-in supervisor that does it, with a startup overlay showing which step is running.

Opt-in and off by default: the supervisor reads `~/.multica/desktop-local-stack.json` and a missing file is the "disabled" signal, so behavior is unchanged for everyone who does not create it.

```json
{ "repoDir": "/path/to/checkout", "composeFile": "docker-compose.selfhost.yml", "backendPort": 8080 }
```

## Bring-up order

The order is load-bearing, not cosmetic: the daemon connects to the backend API as soon as it starts and login cannot succeed before the backend answers.

1. Probe the backend first — with the stack left running between launches this is the common case and costs one request.
2. colima, if the Docker engine is not answering.
3. `docker compose up -d`.
4. Poll the backend until healthy (90s ceiling).

## Fail-open

A supervisor that can wedge the app shut is worse than no supervisor. The gate fails open: any step that errors surfaces in the overlay with the failure text and the user can dismiss it and continue to the normal window. `local-stack-supervisor.test.ts` covers the escape paths, including a window destroyed mid-bring-up.

## Notes for review

- `app.setName(app.getName())` replaces the hardcoded `app.setName("Multica")`. Hardcoding overrode electron-builder's `productName`, so a differently-branded build (a local self-host package, for instance) silently shared userData and the single-instance lock with the official app. For official builds the two are identical.
- `multica-start.sh` / `multica-stop.sh` are the shell equivalent for people not using the desktop app. Both are idempotent, derive the repo directory from the script's own location, and read `BACKEND_PORT` / `FRONTEND_PORT` from the environment with the compose file's defaults (8080 / 3000), exporting them so `docker compose` inherits the same values.
- The start script carries three fixes found while recovering from a colima data-disk I/O failure, each noted in its commit: `multica daemon status` exits 0 even when stopped so the exit-code check never started the daemon; `--code` reported "no code yet" when the log was merely unreadable; and the readiness timeout guessed "probably still migrating", hiding a Postgres auth failure for 60s.
- `DEFAULT_BACKEND_PORT` matches the compose file's own `${BACKEND_PORT:-8080}` rather than assuming a moved port. Installs that relocated the backend set `backendPort` in the config file — the supervisor tests use 8081 as the fixture value precisely to prove the override path works.

## Verification

- `pnpm vitest run` in `apps/desktop` — 58 files, 567 tests pass, 49 of them the new local-stack suites (config loader, state machine, supervisor, overlay).
- `pnpm typecheck` — 7/7 tasks pass.
- `pnpm lint` — 0 errors (28 pre-existing warnings in `packages/views`, untouched here).
- `bash -n` on both scripts; repo-directory derivation verified from a different working directory and via an absolute invocation.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -7,6 +7,7 @@ import fixPath from "fix-path";
 import { setupAutoUpdater } from "./updater";
 import { setupDaemonManager } from "./daemon-manager";
 import { setupLocalDirectory } from "./local-directory";
+import { setupLocalStack } from "./local-stack";
 import { openExternalSafely, downloadURLSafely } from "./external-url";
 import { installContextMenu } from "./context-menu";
 import { handleAppShortcut } from "./keyboard-shortcuts";
@@ -831,6 +832,9 @@ if (!gotTheLock) {
 
     setupAutoUpdater(() => mainWindow);
     setupDaemonManager(() => mainWindow);
+    if (runtimeConfigResult.ok) {
+      setupLocalStack(() => mainWindow, runtimeConfigResult.config.apiUrl);
+    }
     setupLocalDirectory(() => mainWindow);
 
     app.on("activate", () => {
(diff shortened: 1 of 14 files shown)


sail-cn:
Closing: this was opened while exploring a local-only migration; not pursuing an upstream contribution right now.

Issue #7326, opened 2026-08-20
[Bug]: concurrent tasks sharing an 8-character UUID prefix collide on one workdir and overwrite agent identity


mohamedatef90:
## Deployment type

Official App

## Area

Agents and runtimes

## What happened?

Concurrent tasks with different UUIDv7 task IDs were mapped to the same daemon-managed work directory when their IDs shared the same first eight characters.

Three tasks assigned to three different agents had these IDs:

```text
01a01ec0-e69d-...
01a01ec0-f014-...
01a01ec0-f927-...
```

All three reported the same relative work directory:

```text
<workspace-id>/01a01ec0/workdir
```

The task-managed identity files (`CLAUDE.md` / `AGENTS.md`) overwrote one another in that shared directory. One run was authenticated and attributed as the Product & Delivery Manager, but its injected brief identified it as the Tech Lead & Architect. It detected the mismatch and refused to report a clean pass. The shared identity file later contained the third agent's QA identity.

I expected every task to receive a collision-resistant workdir and immutable task/agent identity context.

### Impact

- Cross-agent identity and instruction contamination.
- A task can execute under the wrong role charter even though server attribution points to the correct assignee.
- Potential permission, repository, and output-integrity violations.
- Concurrent multi-project execution is unsafe unless dispatch is serialized and identity/workdir paths are verified.

This appears related to the general isolation concerns in #2639 and the continuity issue in #3933, but the reproduction and root cause differ: agents did not choose the wrong directory, and continuity was not requested. The daemon assigned distinct tasks to one short-prefix directory.

## Steps to reproduce

1. In the Official Desktop app, create three different agents; using multiple runtimes is useful but may not be required.
2. Create and assign three independent issues rapidly so their generated task UUIDv7 IDs share the same first eight characters.
3. Allow the tasks to start concurrently.
4. Inspect each task's `relative_work_dir` in `multica agent tasks <agent-id> --output json`.
5. Inspect the generated `CLAUDE.md` / `AGENTS.md` under that directory.
6. Observe that all tasks use `<workspace-id>/<first-8-task-id-chars>/workdir`, and that later task preparation overwrites the earlier task's identity brief.

### Observed evidence

```text
Task A: 01a01ec0-e69d-...  agent=Product & Delivery Manager
Task B: 01a01ec0-f014-...  agent=Tech Lead & Architect
Task C: 01a01ec0-f927-...  agent=QA & Release Reviewer

[...]

PR #7417, opened 2026-08-21
MUL-6557: fix(daemon): fail closed on cross-project task-root reuse

nathansenn:
## Summary

Two issues created about two seconds apart can share the first eight characters of a UUIDv7 (`01a02554-…`). On daemon versions that truncated the env-root segment to that prefix, a Command Center run and a House Builder Pro run were mapped onto one directory: the incoming task saw the other issue's context, project resources, and worktree.

`taskKey` (random tail + exclusive claim) already shipped in v0.4.32 (#7347 / #7326). This PR closes the remaining fail-closed gaps so that collision cannot still leak another project's files, and so disk-usage and GC can tell the two tasks apart.

## What changed

- **Binding:** `.managed_env.json` now records `project_id` and canonical repo URLs. `shouldReusePriorWorkdir` fails closed when they disagree (legacy empty fields still reuse).
- **Sidecars:** managed Prepare/Reuse overwrite daemon-owned `issue_context.md` and `.multica/project/resources.json`. Local-directory user files are still refused.
- **Quarantine:** a foreign or unowned env root is renamed to a `.quarantine-*` sibling before the incoming task is given an empty directory at the original path. Occupant files are never exposed to the new run.
- **Disk-usage:** reports full `task_id` / `issue_id` / `project_id` from live provenance/marker/owner, not a leftover `.gc_meta.json`. The table has an ID column. Inside a managed task, STATUS is filled with the task token against `/api/issues/{id}` — never the Owner profile.
- **GC:** `cleanTaskDir` retains a completed env root that still has uncommitted files or commits on no remote-tracking ref (APEX-832 custody). Regenerable clones are still reclaimed.

## Tests

- `TestPredictRootDirDistinctForAPEX1684And1685` and `TestPrepareDistinctRootsAndContextsForSharedPrefixCrossProject` using the live APEX-1684 / APEX-1685 ids.
- Refresh leftover HBP `resources.json` / `issue_context.md` into Command Center files.
- Quarantine of a foreign-owned and an unowned occupied root.
- Disk-usage prefers live identity over stale gc_meta.
- Reuse rejects a House Builder Pro provenance for a Command Center task.
- GC retains unique commits and untracked files; reclaims a clone whose commits are on `origin`.

Verified locally:

```
go test ./internal/daemon/execenv/ -count=1
go test ./internal/daemon/ -count=1 -run 'TestShouldReusePriorWorkdir|TestScanDiskUsage|TestCleanTaskDir|TestTaskDirCustody|TestRunTaskSquadLeader'
go test ./cmd/multica/ -count=1 -run 'TestRunDaemonDiskUsage|TestPrintRepoCache|TestDiskUsage'
go test ./internal/handler/ -count=1 -run 'TestTaskDirSegment'
```

## Out of scope

- Does not update a running daemon (Studio is still on 0.4.31; this host is 0.4.29). The collision fix in 0.4.32 plus this PR needs a daemon upgrade to take effect on those machines.
- Does not delete or rewrite the live APEX-1684 / APEX-1685 trees.
- Does not retire the APEX-1614 Command Center checkout on Mac Studio (that host is not this runtime). GC will reclaim it once the issue is terminal, TTL has passed, and the tree has no unique git work.
- Does not touch APEX-11.

Fixes the remaining APEX-1692 gaps on top of #7326 / #7347.

### DIFF
--- a/server/cmd/multica/cmd_daemon.go
+++ b/server/cmd/multica/cmd_daemon.go
@@ -9,6 +9,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -1799,10 +1800,16 @@ func runDaemonDiskUsage(cmd *cobra.Command, _ []string) error {
 	}
 	// Resolve before --top trims the slice: the batch endpoint costs the same
 	// either way, and the JSON consumer sees statuses for everything it gets.
-	// Skipped in a task: the fetcher authenticates with the Owner profile's
-	// stored token, which a task must not borrow, so the column stays blank.
-	if !taskContext && diskUsageNeedsParentStatus(byWorkspace, output) {
-		fillDiskUsageParentStatuses(cmd, profile, &report)
+	// Outside a task this uses the Owner profile token. Inside a task it uses
+	// the injected task token against the ordinary issue API — never the
+	// Owner profile — so STATUS can name the real issue without borrowing
+	// the runtime owner's credential.
+	if diskUsageNeedsParentStatus(byWorkspace, output) {
+		if taskContext {
+			fillDiskUsageParentStatusesWithTaskToken(cmd, &report)
+		} else {
+			fillDiskUsageParentStatuses(cmd, profile, &report)
+		}
 	}
 
 	if top > 0 {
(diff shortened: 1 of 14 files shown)


multica-eve (changes requested):
Thanks for tackling the cross-project workdir isolation problem. The fail-closed direction is the right product behavior, and the regression coverage is thoughtful. I found several blocking issues that need to be addressed before this can merge:

1. **The GC custody behavior needs to be separated or bounded, and its accounting must be accurate.** In `server/internal/daemon/gc.go` and `gc_custody.go`, a custody hold returns without deleting the directory, but `applyGCAction` still increments `stats.cleaned`. The same retained directory can therefore be reported as cleaned on every cycle. The hold criteria are also common in agent workdirs (any untracked/modified file or any unpushed commit), so this can effectively disable GC indefinitely, including for `.quarantine-*` directories. Please either split this product behavior into a separate PR or define an operator-visible bounded policy (for example TTL/watermark/force handling), and only count an item as cleaned when it was actually removed. The custody scan and git subprocesses should also use the GC context so shutdown can cancel them.

2. **Quarantine needs an operator-visible warning.** `adoptQuarantinedEnvRoot` in `server/internal/daemon/execenv/execenv.go` renames an entire workdir and intentionally makes its prior context/session disappear from the new task, but emits no log. Please log the original path, quarantine path, and prior owner task ID at WARN level.

3. **Overwrite mode still leaves stale sidecars when the existing path is a symlink or directory.** In `server/internal/daemon/execenv/context.go`, `writeSidecarFile` only overwrites regular files. For a symlink or directory it returns `errPathPreExists`, which callers ignore, so a new task can still observe stale `issue_context.md` or `resources.json`. Managed env roots are daemon-owned; overwrite mode should safely remove the existing path and recreate the regular file. Please add a symlink regression test.

4. **The sidecar refresh guard also matches local worktrees.** The current check only tests `params.LocalWorkDir == ""`, while local-worktree assignments also have an empty `LocalWorkDir`. That can overwrite tracked `.agent_context/issue_context.md` or `.multica/project/resources.json` files in a user's repository. Please also require `params.LocalWorktree == nil` before refreshing daemon-owned sidecars.

5. **Exact `RepoURLs` equality causes unrelated workdir resets for the same issue.** `ProvenanceMatchesBinding` rejects reuse whenever a project's repository set changes or claim payload resources are temporarily absent, even when `project_id` still matches. That discards session continuity and strands unpushed work despite there being no cross-project collision. Please use `project_id` as the primary isolation boundary, or reject only concrete repository conflicts rather than any set difference. Add coverage for a same-project repository-list change.

6. **Task-token disk-usage status lookup has workspace and timeout problems.** In `server/cmd/multica/cmd_daemon.go`, `ResolveParentStatuses` groups by workspace, but the task-token fetch callback ignores its `workspaceID` and sends every request with `MULTICA_WORKSPACE_ID`; other-workspace task dirs will therefore be queried with the wrong workspace header. Skip groups outside the task token's workspace. Also, the current loop makes N serial requests under one shared API timeout, so a sufficiently large task-dir set can lose the entire STATUS column. Please bound or batch this work. Finally, update the comment that says this path avoids owner-profile scope: task tokens authenticate as the runtime owner and are workspace-scoped rather than issue-scoped, so the current wording is misleading.

The prompt-cache prefix invariant is unaffected, and the targeted build/tests pass on this head. Once the items above are resolved—especially separating or bounding the GC policy and closing the sidecar isolation holes—I’ll be happy to take another look.

PR #7418, opened 2026-08-21
MUL-6558 feat(server): make task queue expiry configurable via MULTICA_TASK_QUEUED_TTL

mmontes11:
## Problem

A task that sits in `queued` longer than the built-in 2h TTL fails with `queued_expired` — the window is a hardcoded constant (`queuedTTLSeconds` in `server/cmd/server/runtime_sweeper.go`) with no operator-facing override. Self-hosted deployments whose runtimes have low task concurrency (e.g. a single runtime running one long task at a time) can legitimately hold queued work behind long-running tasks for more than 2h — every task queued behind it then expires instead of running.

## What changed

- `server/cmd/server/runtime_sweeper.go`
  - `queuedTTLSeconds = 2 * 3600.0` became `defaultTaskQueuedTTL = 2 * time.Hour`; existing behavior (2h) unchanged.
  - `runRuntimeSweeper` and `sweepExpiredQueuedTasks` accept the TTL as a duration parameter instead of reading the constant.
  - Comment on the constant documents the low-concurrency case and points to the new env var.
- `server/cmd/server/main.go`
  - The sweeper is started with `envDuration("MULTICA_TASK_QUEUED_TTL", defaultTaskQueuedTTL)` — the same helper already used by `MULTICA_RUNTIME_RECONNECT_GRACE`. Unset, empty, unparseable, or non-positive values fall back to the 2h default and log a warning at boot (non-positive must stay invalid: a zero TTL would expire every queued task immediately).

## New environment variable

| Variable | Default | Effect |
|---|---|---|
| `MULTICA_TASK_QUEUED_TTL` | `2h` | How long a task may sit in the queue unclaimed before failing with `queued_expired`. Raise it when a runtime's low task concurrency lets tasks legitimately wait behind long-running work. |

Accepts a positive Go duration (`30m`, `6h`, `48h`, ...).

## Deployment plumbing

- `.env.example` — documents the variable with the queued-expiry rationale.
- `docker-compose.selfhost.yml` — pass-through for `docker compose` self-hosting.

## Docs

- `apps/docs/content/docs/environment-variables.mdx` — one row with default and purpose.
- Same row added to the maintained zh/ja/ko translations (machine-translated; happy to accept corrections).

## Testing

- New `server/cmd/server/sweeper_task_timeouts_test.go`: table-driven test for the boot path (unset / valid override / unparseable / non-positive → documented default) plus a regression test pinning the 2h built-in default.
- Existing sweeper integration tests pass unchanged (same default flows through): `TestExpireStaleQueuedTasks`, `TestExpireStaleQueuedTasksRespectsBatchLimit`, `TestSweepDispatchedStaleTask`, `TestSweepRunningTask*`.
- `go build ./...`, `go vet ./cmd/server/`, and the full `cmd/server` test package pass locally against a migrated local Postgres.

## Backward compatibility

No config changes required. The default equals the previously hardcoded value, so existing deployments behave identically until an operator sets `MULTICA_TASK_QUEUED_TTL`.


Bohan-J (changes requested):
Thanks for this — and welcome as a first-time contributor! 🎉

## Product review

This turns "a task queued for more than 2 hours fails" from a hard rule into an operator-tunable ceiling. Before, a self-hosted deployment with low task concurrency could watch a queued task fail before it ever got its turn; after, an admin can extend the wait, while deployments that change nothing keep the exact 2h behavior they have today.

✅ **Product-wise this should land.** It's a real reliability problem in low-concurrency self-hosted setups, a configurable ceiling is more reasonable than telling admins to add capacity or keep retrying, and keeping the default at `2h` keeps the compatibility risk near zero.

## Implementation review

The server side is clean: it reuses the existing duration-env helper, falls back safely to `2h` on unparseable or non-positive values, threads a `time.Duration` into the sweeper and converts to seconds only at the query boundary. `.env.example`, the Compose passthrough and the environment-variables docs (all four locales) are all covered. No DB format, API, or rollback compatibility change.

Two things need to happen before merge, both about the new setting not being reachable/consistent through the official self-hosting surfaces.

### Must-fix

**1. The behavior docs still describe 2 hours as a fixed rule.**

The new variable is announced in `apps/docs/content/docs/environment-variables.mdx:43`, but these still state unconditionally that a task fails after 2 hours:

- `apps/docs/content/docs/tasks.mdx:48`, `:97`, `:152`
- `apps/docs/content/docs/daemon-runtimes.mdx:63`
- `apps/docs/content/docs/troubleshooting.mdx:70`

The `.zh` / `.ja` / `.ko` variants of those three files carry the same statements at the same lines. An admin who sets `6h` would get a contradictory troubleshooting conclusion straight from the official docs — `queued_expired` is exactly the symptom people look up. Please reword these to "2 hours by default, configurable via `MULTICA_TASK_QUEUED_TTL`" across all four locales, linking to the environment-variables page.

**2. The official Helm path doesn't expose the setting.**

`deploy/helm/multica/values.yaml` collects non-secret backend config under `backend.config` (`databaseStartupTimeout: 3m` / `databaseConnectTimeout: 5s` are the closest precedent), and `deploy/helm/multica/templates/configmap.yaml` maps each key to a container env var. Neither carries the queued TTL, so Helm users can't use this feature through the documented values pattern — they'd have to abuse `existingSecret` for a non-secret value or hand-edit the Deployment. Please add e.g. `backend.config.taskQueuedTTL` (default `2h`) plus its ConfigMap mapping, and extend `scripts/helm-config.test.sh` to cover both the default and an override.

### Nits

- `server/cmd/server/runtime_sweeper.go:88-94`: the comment now says 2h "is conservatively above any reasonable 'queued behind a long-running task' window" and then, three lines later, that low-concurrency runtimes can legitimately queue for longer than 2h. Worth scoping the first claim to the historical default assumption so the two don't read as contradictory.
- `server/cmd/server/sweeper_task_timeouts_test.go:8-10`: the comment says the value "reaches the sweeper unchanged", but the test only exercises `envDuration` parsing and the default constant — the wiring into `runRuntimeSweeper` isn't covered. Either narrow the comment or add a case that pins the call chain.

## Verification

- `go test ./cmd/server -run '^(TestQueuedTTLFromEnv|TestDefaultTaskQueuedTTL)$' -count=1`: pass
- The two queued-expiry DB integration tests: pass
- `go vet ./cmd/server`: pass
- `docker compose -f docker-compose.selfhost.yml config --quiet`: pass
- `git diff --check` and merge-tree against current `origin/main`: clean, no conflicts
- Full `go test ./cmd/server -count=1`: only `TestReadinessEndpoints` fails locally (`/readyz` returns 503); it fails identically on a clean `origin/main` without this PR, so it's a local/baseline issue, not caused by this change.

## Recommendation

🔄 **Request changes** — the product direction and the server implementation are both good; they just need the behavior docs and the Helm self-hosting entry point to actually match the new capability. Happy to re-review as soon as those two are in.

mmontes11:
@Bohan-J all review points are addressed — ready for a re-review when you get a chance.

**1. Behavior docs** — the queued-expiry statements in `tasks`, `daemon-runtimes` and `troubleshooting` (en + zh/ja/ko) now read "2 hours by default, adjustable via `MULTICA_TASK_QUEUED_TTL`", each linking to the environment-variables page.

**2. Helm** — new `backend.config.taskQueuedTTL: 2h` in `values.yaml`, mapped to `MULTICA_TASK_QUEUED_TTL` in the ConfigMap, with default and override assertions added to `scripts/helm-config.test.sh`.

**Nits** — the constant comment now frames "conservatively above any reasonable window" as the *historical* default assumption that breaks down for low-concurrency runtimes, and the test comment is narrowed to what the test actually covers (env parsing; same helper/default `main` passes into `runRuntimeSweeper`).

Also: the branch has been synced with current `main` (merge commit on top), so it's mergeable against the latest again.

**Verification**: `scripts/helm-config.test.sh` → `0 chart(s) failed` + `helm config rendering ok`; `go build ./...`, `go vet ./cmd/server/` and the queued-TTL unit tests pass; `git diff --check` clean.

---

Bohan-J (approved):
Thanks for turning this around so quickly, @mmontes11 — both blocking items are fully addressed. 🙌

- The behavior docs now say "2 hours by default, adjustable via `MULTICA_TASK_QUEUED_TTL`" consistently across `tasks`, `daemon-runtimes` and `troubleshooting` in all four locales, so the `queued_expired` troubleshooting path no longer contradicts the setting.
- The Helm chart exposes `backend.config.taskQueuedTTL: 2h`, renders it to `MULTICA_TASK_QUEUED_TTL` in the ConfigMap, and `scripts/helm-config.test.sh` asserts both the default and a `12h` override.
- The two round-one nits (the self-contradicting sweeper comment and the over-claiming test comment) are cleaned up as well.

I pushed one small commit to your branch (`59ddb09`) to clear the last two nits rather than send you another round trip:

- `.env.example` had a trailing space that failed `git diff --check`, and repeated the empty/unset fallback sentence twice. Folded into one, phrased to match the neighbouring `MULTICA_RUNTIME_RECONNECT_GRACE` block.
- `daemon-runtimes.{mdx,zh,ja,ko}` named the variable but didn't link `/environment-variables` the way `tasks` and `troubleshooting` do — added the link in all four locales.

Verified locally on the final head: `go vet ./cmd/server`, `go build ./cmd/server`, `go test ./cmd/server -run '^(TestQueuedTTLFromEnv|TestDefaultTaskQueuedTTL)$' -count=1`, `bash -n scripts/helm-config.test.sh` and `git diff --check` all pass (the Helm render assertions run in CI's `backend-tests`).

✅ **Approved.** Nice, well-scoped fix — the default stays `2h` so existing deployments are untouched, and low-concurrency self-hosted setups finally get a way out of avoidable `queued_expired` failures. Merging once CI is green.

mmontes11:
Thank you, @Bohan-J! Huge congratulations on this game-changing project. It has completely transformed my AI workflows, both locally and with Claude. I also really appreciate your ongoing maintenance efforts!

PR #7419, opened 2026-08-21
fix(issues): position status changes at column top

### LINKED ISSUE #7410: [Bug]: Changing issue status leaves position at -1, burying agent-completed issues mid-column
### Deployment type

Official App

### Area

### What happened?

Changing an issue's status does not assign it a board `position` in the destination column — `position` stays at the default `-1`. Drag-and-drop assigns a real position, so the two write paths disagree.

Because a column sorts by `position` ascending, and hand-positioned issues carry negative values (in our workspace `-57.75 … -7.25`), every issue at `-1` sorts *below* all of them. The practical effect: an issue completed by an agent lands in the middle of the Done column instead of at the top, and the user reports it as "the task disappeared from the board."

This is not cosmetic once a Done column outgrows one screen. In our workspace, 62 of 211 done issues were stuck at `position: -1` — every one of them closed out by an agent. Two tasks an agent had just finished (`CAN-238`, `CAN-239`) were sitting at **rows 148 and 147 of 211**, so the user had to scroll past ~146 older completed issues to find work that finished minutes earlier. They reasonably concluded the issues had vanished into a hidden status.

Expected: a status change places the issue at a defined slot in the destination column — the top for a completion — the same way drag-and-drop does. Any deterministic policy would be fine; leaving `-1` is what breaks, because it silently means "below everything a human ever touched."

This hits agent-driven workspaces hardest, since agents close issues via the status path and never drag anything.

### Steps to reproduce

1. In a workspace where some issues have been hand-positioned by dragging (so they hold real negative `position` values), take an issue in Todo.
2. Change its status to Done via the API/CLI rather than by dragging: `multica issue status <id> done`.
3. Read the issue back: `multica issue get <id> --output json` → `"position": -1`.
4. Open the board and look at the Done column. The issue is not at the top; it sits below every hand-positioned issue in that column.
5. With a large Done column, it is effectively invisible without scrolling.

```shell
# 62 of 211 done issues left at the default position
$ multica issue list --status done --limit 100 --offset N --output json
CAN-239  position: -1   # agent-completed, rank 147 of 211
CAN-238  position: -1   # agent-completed, rank 148 of 211
CAN-237  position: -57.75  # hand-positioned, rank 1

# workaround: reposition after the fact
$ multica issue reorder <id> --top
```

Two smaller things noticed while working around this, happy to split into separate issues if preferred:

1. **Paginating a column with tied positions is unsafe.** `multica issue list --status done --limit 100 --offset 100` returned duplicate rows and silently omitted others while 62 issues shared `position: -1`. A tiebreaker in the sort (e.g. `id` or `number`) would make offset pagination stable.
2. **No way to sort by issue number.** `--sort` accepts `position`, `title`, `created_at`, `start_date`, `due_date`, `priority`. `created_at desc` is a workable proxy for descending issue number, but the board UI has no equivalent control, and descending number is the natural reading order for a workspace where agents are closing issues all day.

Related but distinct: #7267 (stable ordering within a stage).

mameikagou:
## What does this PR do?

Status changes through the API/CLI kept the issue's old position, which could bury newly completed work below manually ordered issues.

The board already orders positions ascending and new issues use `MIN(position) - 1`. This applies the same policy to both status-update queries while preserving an explicitly supplied position.

Risk: concurrent status changes can still choose the same position, matching the repository's existing best-effort ordering behavior and secondary sort fallback.

## Related Issue

Closes #7410

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Reposition issues at the top of the destination column when their status changes.
- Preserve explicit positions supplied with the status update.
- Cover both `UpdateIssue` and direct `UpdateIssueStatus` paths with database-backed tests.

## How to Test

1. `go test ./internal/handler -run 'Test(UpdateIssueStatusPositionTopOfColumn|UpdateIssueStatusKeepsExplicitPosition|DirectUpdateIssueStatusPositionTopOfColumn)$' -count=1`
2. `go test ./internal/handler -count=1`

## Checklist

## AI Disclosure

**Prompt / approach:** Reproduce #7410 on current main, trace both status-update paths, reuse the existing `MIN(position) - 1` ordering policy, and add database-backed regression tests using the repository fixtures.

### DIFF
--- a/server/internal/handler/issue_create_position_test.go
+++ b/server/internal/handler/issue_create_position_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/multica-ai/multica/server/internal/testutil"
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 )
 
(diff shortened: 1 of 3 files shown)


Bohan-J:
Thank you for this — and sorry it took a second PR to get there.

Your diagnosis was right, and it was the useful part: `position` is a per-`(workspace, status)` rank, a status change was carrying the old column's value over unchanged, and reusing the existing `MIN(position) - 1` policy on both write paths is the correct fix. #7422 landed with that same approach, plus a few things specific to this repo:

- Registering the new index in `concurrentIndexCleanups` in `server/cmd/migrate/main.go`. Every `CREATE INDEX CONCURRENTLY` migration needs an entry there or `TestEveryConcurrentUpBuildHasCleanup` fails the build — an easy one to hit if you haven't touched migrations here before.
- A `(workspace_id, status, position)` index. The re-rank reads `MIN(position)` for the destination column, and `idx_issue_status` stops at `(workspace_id, status)`, so that min cost a heap fetch per issue in the column — paid on every completion, against Done.
- Guard tests for the cases that must *not* change: same-status writes leaving `position` alone, and an explicit `position` still winning so cross-column drag-and-drop keeps the slot it dropped on.

Since #7422 touches the same lines, this branch now conflicts with `main`, so I'm closing it as superseded rather than leaving it open to rot. That's not a judgment on the work — you found and correctly diagnosed a real bug that was hurting agent-driven workspaces, and the fix that shipped follows your reasoning.

The two smaller items you raised are tracked separately: sorting by issue number is a genuine gap (missing from both the API sort whitelist and the CLI) and is a feature rather than part of this bug. On the pagination one — `ListIssues` already orders by `i.position ASC, i.created_at DESC, i.id DESC`, which is a unique total order, so tied positions alone shouldn't duplicate or drop rows. What you saw was most likely rows moving across page boundaries while agents mutated them mid-scroll, which is an offset-pagination property a tiebreaker can't fix. Worth its own issue if it keeps biting you.

Please do send more — this was a good catch.

mameikagou:
Thanks for the detailed explanation and for carrying the fix through. The migration cleanup and destination-column index details make sense; I’ll keep those repository-specific constraints in mind for future contributions.

PR #7423, opened 2026-08-22
MUL-6560: WS-4075: honor run-only autopilot declared failures

tangyuanjc:
## Summary

- add an explicit `[MULTICA_AUTOPILOT_FAILED]` first-line protocol for run-only agents and convert declared failures into `autopilot_run.status=failed` with a visible reason
- preserve the existing `Run failed fast:` and `Phase 4 verify FAILED:` contracts, while avoiding false positives when successful reports quote those phrases later
- backfill historical misclassified runs and teach the runtime brief/built-in autopilot skill the new contract

## Verification

- `go test ./internal/service -run '^(TestTaskDeclaredFailureForAutopilotRun|TestTaskFailureReasonForAutopilotRun|TestAutopilotErrorType)$' -count=1`
- `go test ./internal/daemon/execenv -run '^(TestBriefTeachesAutopilotDeclaredFailureMarker|TestBriefOwnsAutopilotIssueCommandsGuard)$' -count=1`
- `go test ./cmd/server -run '^TestAutopilotRunOnlyTaskTerminalEventsUpdateRun$' -count=1`
- `go vet ./internal/service ./internal/daemon/execenv ./pkg/protocol ./cmd/server`
- migration 398 exercised against an isolated migrated PostgreSQL database with both legacy prefixes plus a negative quoted-failure control

Closes WS-4075

### DIFF
--- a/server/cmd/server/autopilot_listeners_test.go
+++ b/server/cmd/server/autopilot_listeners_test.go
@@ -45,6 +45,16 @@ func TestAutopilotRunOnlyTaskTerminalEventsUpdateRun(t *testing.T) {
 			wantResult: "done",
 		},
 		{
+			name: "agent declared failure",
+			finalize: func(task db.AgentTaskQueue) {
+				if _, err := taskSvc.CompleteTask(ctx, task.ID, []byte(`{"output":"[MULTICA_AUTOPILOT_FAILED] budget_exceeded: no delivery receipt"}`), "", "", "", false, "", ""); err != nil {
+					t.Fatalf("CompleteTask: %v", err)
+				}
+			},
+			wantStatus: "failed",
+			wantReason: "task declared failure: budget_exceeded: no delivery receipt",
+		},
+		{
 			name: "failed",
 			finalize: func(task db.AgentTaskQueue) {
 				if _, err := taskSvc.FailTask(ctx, task.ID, "boom", "", "", "", "agent_error", false, "", ""); err != nil {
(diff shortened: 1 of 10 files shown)


multica-eve (changes requested):
The product direction is right: future run-only agent declarations should mark the run as failed instead of reporting a false success. I’m requesting changes for two scoped fixes.

1. **Remove the historical backfill.** Our intended scope is prospective: apply declared-failure handling to future agent completions, but do not reclassify existing `autopilot_run` rows. Please remove `server/migrations/398_backfill_declared_autopilot_failures.{up,down}.sql` and the corresponding backfill references in the source map / PR description. The two legacy output prefixes may remain supported for future completions; only the historical data rewrite should be removed. This also eliminates the migration-prefix collision with current `main`.

2. **Persist the complete task result for declared failures.** In `server/internal/service/autopilot.go`, the declared-failure branch calls `failAutopilotRun`, while `UpdateAutopilotRunTerminalWithQuota` only writes `result` for `terminal_status = 'completed'`. As a result, the failed run records the first-line `failure_reason` but leaves `autopilot_run.result` empty, so the full final report is unavailable from run history. Please carry `task.Result` into the declared-failure terminal update and add an integration assertion that both `status = failed` and the full result are preserved.

The quota behavior is accepted: an agent-declared failure should release the autopilot quota reservation rather than consume it. No change is requested there.

The first-non-empty-line parsing, prospective legacy-prefix compatibility, and byte-identical prompt-cache behavior otherwise look sound.

Issue #7215, opened 2026-08-19
[Bug]: WeCom smart bot — agent replies are generated but never pushed back to WeCom (outbound silently dropped)


Bohan-J:
### Deployment type

self-host

> **Needs confirmation from the reporter.** They run the Multica **desktop app with a local daemon (v0.4.29)** on Windows, which does not by itself tell us whether the backend is Multica Cloud or a self-hosted server. This distinction decides the whole triage — see Q1 below.

### What happened?

Reported by a user on Discord (translated from Chinese).

**WeCom (企业微信) smart bot, long-connection API mode.** Sending a message to the bot in WeCom works inbound: Multica receives it and the agent generates a reply (the chat session record exists on the Multica side). But **nothing ever arrives back in WeCom.** Inbound is fine; the outbound push is lost, silently — the user sees no error, no failure notice, nothing. The conversation just goes quiet.

**Environment**
- Windows
- Multica daemon v0.4.29 (local desktop)
- WeCom smart bot (long-connection / WebSocket API mode)

**What the reporter already ruled out**
- Subscribed directly to `wss://openws.work.weixin.qq.com` with the BotID + Secret via `aibot_subscribe` → `errcode=0`. Credentials are valid and the bot is reachable.
- Local network and TLS path to WeCom are healthy.
- Restarting the Multica desktop app sometimes restores it briefly, but the problem comes back (and sometimes a restart does not help at all).

**The reporter's own lead**
Their local daemon log shows repeated abnormal WebSocket disconnects from the Multica backend on the evening of 8/18 — `close 1006`, remote forcible close, `wsasend: connection aborted`. Their hypothesis is that the WeCom bot's long connection went "half-dead" in the same way, and replies are being written into a dead socket and silently lost.

### Steps to reproduce

1. Bind a WeCom smart bot (long-connection API mode) to a Multica agent.
2. Send the bot a message in WeCom (direct message).
3. Multica receives the message and the agent produces a reply — visible in the Multica chat session.
4. Nothing arrives in WeCom. No error is shown anywhere.
5. Restarting the Multica desktop app may restore delivery temporarily; the failure recurs.

### Additional context

I did a read of the current WeCom adapter before filing, so triage does not start from the user's guess. Notes below, separating **what the code says** from **what is still unknown**.

#### 1. The reporter's hypothesis is probably the wrong connection

[...]

PR #7428, opened 2026-08-22
MUL-6569 feat(wecom): give every undelivered reply a reason an operator can count

seacen:
## What this is for

A WeCom user whose question goes unanswered while the answer sits in the Multica transcript is one symptom with at least five causes, and the outbound path cannot tell them apart. Each cause ends in a bare `return nil` or a single WARN with no reason attached, so neither we nor a deployment's own operator can say which branch fired. That is why #7215 and #6890 can be argued about but not settled — and why a fix for either would be a guess.

This changes no behaviour. It makes the branches countable.

## What it adds

`multica_wecom_outbound_dropped_total{reason}` — a reply this adapter owed a WeCom user and did not deliver. The reason set is closed, because it is a metric label and an open one is the unbounded cardinality `forbiddenMetricLabels` exists to prevent:

| reason | what it means |
|---|---|
| `no_live_connection` | no socket for this installation in this process — the Supervisor is mid-reconnect, or on a multi-replica deployment the lease is held elsewhere (`SELF_HOSTING.md`'s single-replica exception) |
| `origin_not_channel` | asked in the Multica web UI on a WeCom-bound session; its answer belongs only in Multica |
| `installation_inactive` | revoked between the trigger and the reply |
| `task_missing` | no task id on the completion, or the row was reaped while its ending was in flight |
| `platform_refused` | WeCom answered the send with a non-zero errcode |
| `ack_timeout` | the frame went out and no verdict came back — **not** proof of non-delivery |
| `transport_error` | the write failed, or a lookup ahead of it did |

`multica_wecom_outbound_delivered_total` — the denominator, and not optional. Without it a drop rate of zero and a bot nobody messaged read exactly alike, which is the same silence being reported.

## Two judgements worth reviewing

**Log level, not the counter, carries severity.** `origin_not_channel` is ordinary in a healthy deployment and logs at DEBUG; the reasons that mean somebody in WeCom is waiting on an answer that is not coming log at WARN. The counter records both, so the breakdown stays readable as a rate rather than as an error total.

**A session with no WeCom binding is not counted at all.** This subscriber sees every `chat:done` in the deployment, including Slack's and the web UI's, and those were never this adapter's to answer.

## Notes

- `"wecom: connection not ready on this replica"` becomes a sentinel error, so the classifier names it instead of pattern-matching prose.
- One place records a drop, so it is counted exactly once and always carries a reason.
- No installation, workspace or session id anywhere near a label.

## Verification

- `TestEveryUndeliveredReplyNamesItself` walks the terminal paths and asserts the counter, the reason in the log, and the WARN/DEBUG split for each.
- `TestReasonStringsArePinned` / `TestActionableSplitIsPinned` guard the label values and the severity split — without them a rename passes every other test in the package while silently retiring whatever alert reads it.
- `TestNonWecomSessionIsNotADrop`, `TestNilMetricsSinkIsSafe` (a deployment with `/metrics` off must not panic on the first drop), `TestRefusalIsNotATransportError`.
- Reverse-verified: removing the record at the origin gate fails exactly `TestEveryUndeliveredReplyNamesItself/asked_in_the_web_UI` and nothing else.
- Full package green under `-race`, plus `internal/metrics` and `cmd/server`.

### DIFF
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -967,6 +967,12 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 					wecomOutboundOpts = append(wecomOutboundOpts, wecom.WithAttachments(store))
 					h.DeclareChannelFileDelivery(string(wecom.TypeWecom))
 				}
+				// The outbound subscriber reports to the same sink the
+				// connection path uses, so an operator reads "replies are
+				// being dropped, and for which reason" off the same dashboard
+				// as "the bot cannot connect".
+				wecomOutboundOpts = append(wecomOutboundOpts,
+					wecom.WithOutboundMetrics(wecomMetricsOrNil(opts.WecomMetrics)))
 				wecom.NewOutbound(queries, wecomSenders, slog.Default(), wecomOutboundOpts...).Register(bus)
 
 				// Ranges the media fetcher may dial despite looking reserved.
--- a/server/internal/analytics/client_test.go
+++ b/server/internal/analytics/client_test.go
@@ -105,11 +105,11 @@ func TestPostHogClient_DropsWhenFull(t *testing.T) {
 
 func TestEmailDomain(t *testing.T) {
 	cases := map[string]string{
-		"a@example.com":       "example.com",
-		"user@Company.co.uk":  "company.co.uk",
-		"":                    "",
-		"no-at":               "",
-		"trailing@":           "",
+		"a@example.com":      "example.com",
+		"user@Company.co.uk": "company.co.uk",
+		"":                   "",
+		"no-at":              "",
+		"trailing@":          "",
 	}
 	for in, want := range cases {
 		if got := emailDomain(in); got != want {
--- a/server/internal/analytics/client_test.go
+++ b/server/internal/analytics/client_test.go
@@ -105,11 +105,11 @@ func TestPostHogClient_DropsWhenFull(t *testing.T) {
 
 func TestEmailDomain(t *testing.T) {
 	cases := map[string]string{
-		"a@example.com":      "example.com",
-		"user@Company.co.uk": "company.co.uk",
-		"":                   "",
-		"no-at":              "",
-		"trailing@":          "",
+		"a@example.com":       "example.com",
+		"user@Company.co.uk":  "company.co.uk",
+		"":                    "",
+		"no-at":               "",
+		"trailing@":           "",
 	}
 	for in, want := range cases {
 		if got := emailDomain(in); got != want {
--- a/server/internal/auth/cloud_pat_test.go
+++ b/server/internal/auth/cloud_pat_test.go
@@ -345,7 +345,6 @@ func TestCloudPATVerifier_NegativesNotCached(t *testing.T) {
 	}
 }
 
-
 // TestCloudPATVerifier_LookupRejectsUnknownOwner pins the new
 // owner-existence guard. Cloud says the token is valid, but the
 // caller's lookup says the owner_id does not exist locally — the
--- a/server/internal/auth/cloud_pat_test.go
+++ b/server/internal/auth/cloud_pat_test.go
@@ -345,6 +345,7 @@ func TestCloudPATVerifier_NegativesNotCached(t *testing.T) {
 	}
 }
 
+
 // TestCloudPATVerifier_LookupRejectsUnknownOwner pins the new
 // owner-existence guard. Cloud says the token is valid, but the
 // caller's lookup says the owner_id does not exist locally — the
--- a/server/internal/auth/membership_cache_test.go
+++ b/server/internal/auth/membership_cache_test.go
@@ -13,7 +13,7 @@ func TestMembershipCache_NilSafe(t *testing.T) {
 	if c.Get(ctx, "any-user", "any-workspace") {
 		t.Fatal("nil cache must miss")
 	}
-	c.Set(ctx, "any-user", "any-workspace") // no panic
+	c.Set(ctx, "any-user", "any-workspace")        // no panic
 	c.Invalidate(ctx, "any-user", "any-workspace") // no panic
 }
 
--- a/server/internal/auth/membership_cache_test.go
+++ b/server/internal/auth/membership_cache_test.go
@@ -13,7 +13,7 @@ func TestMembershipCache_NilSafe(t *testing.T) {
 	if c.Get(ctx, "any-user", "any-workspace") {
 		t.Fatal("nil cache must miss")
 	}
-	c.Set(ctx, "any-user", "any-workspace")        // no panic
+	c.Set(ctx, "any-user", "any-workspace") // no panic
 	c.Invalidate(ctx, "any-user", "any-workspace") // no panic
 }
 
--- a/server/internal/integrations/lark/markdown_detect.go
+++ b/server/internal/integrations/lark/markdown_detect.go
@@ -15,15 +15,15 @@ import (
 // Patterns are compiled once at package init; containsMarkdown is on
 // the chat-reply hot path.
 var markdownPatterns = []*regexp.Regexp{
-	regexp.MustCompile("(?m)^#{1,6}[ \t]"),               // headings: # H1, ###### H6
-	regexp.MustCompile("(?m)^[ \t]*[-*+][ \t]"),          // unordered list
-	regexp.MustCompile("(?m)^[ \t]*\\d+\\.[ \t]"),        // ordered list
-	regexp.MustCompile("(?m)^>[ \t]"),                    // blockquote
+	regexp.MustCompile("(?m)^#{1,6}[ \t]"),                        // headings: # H1, ###### H6
+	regexp.MustCompile("(?m)^[ \t]*[-*+][ \t]"),                   // unordered list
+	regexp.MustCompile("(?m)^[ \t]*\\d+\\.[ \t]"),                 // ordered list
+	regexp.MustCompile("(?m)^>[ \t]"),                             // blockquote
 	regexp.MustCompile("(?m)^[ \t]*(?:---|\\*\\*\\*|___)[ \t]*$"), // hr / thematic break
-	regexp.MustCompile("\\*\\*[^*\\n]+\\*\\*"),           // **bold**
-	regexp.MustCompile("__[^_\\n]+__"),                   // __bold__
-	regexp.MustCompile("(?m)^[ \t]*\\|.+\\|[ \t]*$"),     // table row (must have | on both ends)
-	regexp.MustCompile("\\[[^\\]\\n]+\\]\\([^)\\n]+\\)"), // [text](url) link / image
+	regexp.MustCompile("\\*\\*[^*\\n]+\\*\\*"),                    // **bold**
+	regexp.MustCompile("__[^_\\n]+__"),                            // __bold__
+	regexp.MustCompile("(?m)^[ \t]*\\|.+\\|[ \t]*$"),              // table row (must have | on both ends)
+	regexp.MustCompile("\\[[^\\]\\n]+\\]\\([^)\\n]+\\)"),          // [text](url) link / image
 }
 
 // containsMarkdown returns true when the body almost certainly
--- a/server/internal/integrations/lark/markdown_detect.go
+++ b/server/internal/integrations/lark/markdown_detect.go
@@ -15,15 +15,15 @@ import (
 // Patterns are compiled once at package init; containsMarkdown is on
 // the chat-reply hot path.
 var markdownPatterns = []*regexp.Regexp{
-	regexp.MustCompile("(?m)^#{1,6}[ \t]"),                        // headings: # H1, ###### H6
-	regexp.MustCompile("(?m)^[ \t]*[-*+][ \t]"),                   // unordered list
-	regexp.MustCompile("(?m)^[ \t]*\\d+\\.[ \t]"),                 // ordered list
-	regexp.MustCompile("(?m)^>[ \t]"),                             // blockquote
+	regexp.MustCompile("(?m)^#{1,6}[ \t]"),               // headings: # H1, ###### H6
+	regexp.MustCompile("(?m)^[ \t]*[-*+][ \t]"),          // unordered list
+	regexp.MustCompile("(?m)^[ \t]*\\d+\\.[ \t]"),        // ordered list
+	regexp.MustCompile("(?m)^>[ \t]"),                    // blockquote
 	regexp.MustCompile("(?m)^[ \t]*(?:---|\\*\\*\\*|___)[ \t]*$"), // hr / thematic break
-	regexp.MustCompile("\\*\\*[^*\\n]+\\*\\*"),                    // **bold**
-	regexp.MustCompile("__[^_\\n]+__"),                            // __bold__
-	regexp.MustCompile("(?m)^[ \t]*\\|.+\\|[ \t]*$"),              // table row (must have | on both ends)
-	regexp.MustCompile("\\[[^\\]\\n]+\\]\\([^)\\n]+\\)"),          // [text](url) link / image
+	regexp.MustCompile("\\*\\*[^*\\n]+\\*\\*"),           // **bold**
+	regexp.MustCompile("__[^_\\n]+__"),                   // __bold__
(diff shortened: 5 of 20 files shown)


multica-eve (dismissed):
Thanks for adding bounded outcome reasons and a delivered baseline. The product direction makes sense, but two correctness issues currently make the resulting metrics misleading:

1. `origin_not_channel` should not increment `multica_wecom_outbound_dropped_total`. In `server/internal/integrations/wecom/outbound.go`, this branch is explicitly a reply that belongs only in the Multica web UI, so the adapter did not owe it to a WeCom user. Counting it as dropped contradicts the metric's own help text and makes normal web usage degrade the apparent WeCom delivery rate. Please exclude this path from dropped outcomes or expose it separately as a skipped outcome, with the metric contract and tests updated accordingly.

2. Attachment-only replies have no outcome. `RecordOutboundDelivered` is called only after a non-empty text send, while the asynchronous attachment path in `outbound_media.go` does not record either of the new outcome metrics. A valid file-only reply can therefore be acknowledged by WeCom while `outbound_delivered_total` remains unchanged; attachment refusal, timeout, and local delivery failures are also absent from this outcome breakdown. Please first define whether the unit is a whole reply or each delivered item, then make file-only success/failure produce exactly one interpretable outcome and add regression coverage for both cases. The mixed text-plus-file case should also have unambiguous semantics.

Please also extend the production metrics collector test to exercise the new counter and reason label so the exported metric contract is covered, not just the test double.

seacen:
All three addressed. The two correctness findings had one cause: I never said what a count is a count *of*.

**1. `origin_not_channel` is no longer a drop.** You're right that it contradicts the metric's own help text — that reply was never owed to a WeCom user. It moves to a new `multica_wecom_outbound_skipped_total{reason}`, and I took `installation_inactive` with it (there is no longer an installation to deliver through, and the bot is gone from the user's side too) plus a new `nothing_to_say` for an empty completion carrying no file. Every remaining drop is actionable, which is why `dropReason.actionable()` collapses to `true`.

**2. The unit is a REPLY, now stated and enforced.** A reply bearing words is settled by the text send. A reply whose substance *is* a file was previously settled by nothing at all. The attachment path now owes that reply exactly one outcome, and every exit produces one: lookup failure, admission shedding, pending shedding, the slot wait expiring, no live sender, and the send loop.

Files get their own pair — `outbound_attachment_delivered_total` and `outbound_attachment_dropped_total{reason}`, one per file, always. A reply whose words arrived and whose file did not is a delivered reply with a failed attachment, and one number cannot say both. Refusals, ack timeouts and local failures are classified the same way the reply path classifies them.

**3. The exported contract is covered.** `TestWecomOutboundMetricsAreExported` registers every collector, records one of each, gathers, and asserts the scraped metric names and the `reason` label value — not the adapter's double.

New regression coverage: `TestNotOwedIsSkippedNotDropped` (both skip paths, asserting no drop and no WARN), `TestSkipReasonStringsArePinned`, and the reason-string pin extended to `attachment_not_admitted`.

Green under `-race` across `wecom`, `internal/metrics` and `cmd/server`.

One judgement I'd like you to check rather than assume: I kept `installation_inactive` on the skipped side. The user did ask in WeCom and is arguably still waiting — but there is no longer an installation to deliver through, so I read it as "not deliverable" rather than "we failed to deliver". If you'd rather it counted as a drop, say so and I'll move it; it is a one-line change plus its test.

multica-eve (dismissed):
Thanks for separating skipped completions from dropped replies and for adding coverage of the exported Prometheus contract. The original `origin_not_channel` issue is resolved. There are still three blocking correctness problems in the attachment outcome path:

1. The post-lookup pending-cap branch in `server/internal/integrations/wecom/outbound_media.go` (around lines 264–270) returns after shedding known attachments without recording either an attachment outcome or, for an attachment-only response, a reply outcome. A targeted regression with one known file and a full pending queue observes `attachment_dropped = 0` and `outbound_dropped = 0`, although both should be settled.

2. The earlier admission-cap branch (around lines 194–203) has the opposite problem: it calls `attachmentDropped` before the attachment lookup, when the code explicitly does not know whether any file exists. A normal text reply with no bound files therefore creates one attachment drop when admission is full; a response with multiple files would still create only one. Both contradict the new contract that attachment metrics count files. Please use a separate scheduling/admission metric with its actual unit, or otherwise keep per-file outcome accounting on paths where the rows are known.

3. When every file in an attachment-only response fails, lines 332–338 always record the reply reason as `transport_error`, discarding the reason already classified for the files. A single file explicitly refused by WeCom currently produces `outbound_attachment_dropped_total{reason="platform_refused"}` but `outbound_dropped_total{reason="transport_error"}`. Please preserve the real reason for the single-file case and define a truthful, stable aggregation rule for multi-file failures.

Please add path-level regression tests for admission shedding, pending shedding, a refused attachment-only response, and multi-file cardinality. The current exported-metrics test proves collector names and labels, but does not exercise these control-flow branches, which is why all three regressions pass CI.

Also update the help text for `outbound_dropped_total` in `server/internal/metrics/wecom.go`: it still says `origin_not_channel` belongs to that metric and that the total is not an error total, which now contradicts the implementation.

seacen:
All three fixed, plus the help text. They were one mistake seen from three sides: a count fed from a place that does not know the thing it claims to count.

**1. The pending-cap branch settles everything it knows.** The rows are known there, so it now records one `attachment_not_admitted` drop per file and — when the files are the reply — the reply's own outcome. `TestPendingShed_SettlesTheFilesAndTheReply` reproduces your exact scenario: one known file, full pending queue, and asserts both counters land.

**2. The admission gate no longer counts phantom files.** You caught it contradicting its own comment — "nothing is known about this turn yet" directly above a per-file record. It now records a new `multica_wecom_outbound_attachment_delivery_shed_total`, a **scheduling** decision with its own unit, and no file drops at all. The reply outcome stays, because that unit is one reply regardless of how many files the turn turns out to hold. `TestAdmissionShed_CountsASchedulingDecisionNotFiles` runs a zero-file turn under a full cap and asserts shed=1, file-drops=0, and the delivered words still counted.

**3. The reply keeps its file's real reason.** The settle now aggregates the per-file classifications under a stable precedence (`worseDropReason`: a stated refusal beats a definite local failure beats a verdict that never came — ordered by how specific a fact each is about why the content did not arrive), so the single-file case keeps `platform_refused` and the multi-file case is a rule rather than an accident of loop order. `TestRefusedFileOnlyReply_KeepsThePlatformsReason` asserts the reply reason matches the file's; `TestMultiFileFailure_CountsFilesPerFileAndTheReplyOnce` pins three file drops against exactly one reply outcome; `TestWorseDropReason_PrecedenceIsStable` pins the rule itself.

`outbound_dropped_total`'s help text no longer mentions `origin_not_channel` and now says plainly that every reason there is somebody waiting.

Your closing observation was the diagnosis: the exported-contract test proves names and labels but says nothing about which branches feed them. The new tests are control-flow tests on purpose. Reverse-verified: reverting the aggregation alone fails exactly `TestRefusedFileOnlyReply_KeepsThePlatformsReason` and nothing else.

Green under `-race` across `wecom`, `internal/metrics` and `cmd/server`.

multica-eve (dismissed):
This revision fixes several earlier reporting gaps, but administrators can still be told that a reply failed when nothing was due, or see only part of a multi-file failure during heavy load and reconnects. That can send them toward the wrong remediation, so the behavior is not ready to ship yet.

The product direction is still worth accepting: self-hosted operators need a stable breakdown that maps an undelivered WeCom reply to an actionable cause. The current implementation has four remaining blockers:

1. `server/internal/integrations/wecom/outbound_media.go:199-209` records a reply drop at the pre-lookup admission gate whenever `content` is empty. At this point the code does not know whether an attachment exists. A targeted path test with empty content and zero bound attachments increments `outbound_dropped` from 0 to 1, although the lookup would have classified the completion as `nothing_to_say`. This gate can truthfully count a scheduling shed, but cannot assert a reply failure without knowing that a reply payload exists.

2. Two post-lookup early exits do not settle every known file. The global-slot timeout at `outbound_media.go:291-301` records zero attachment drops, while the no-live-sender branch at `outbound_media.go:307-315` records exactly one regardless of `len(rows)`. With three known attachments, targeted regressions observe `0/3` for `transport_error` and `1/3` for `no_live_connection`. Both branches need one attachment outcome per row, while retaining at most one reply outcome for an attachment-only reply.

3. `server/internal/integrations/wecom/outbound_outcome.go:52-56` correctly states that `ack_timeout` is not proof of non-delivery—the message may already be visible—but that outcome is still counted in `outbound_dropped_total` and `outbound_attachment_dropped_total`. This contradicts the collector contract in `server/internal/metrics/wecom.go:79`, which says every dropped reason means the answer is not coming, and the attachment help text at lines 87-89, which says the file did not reach the user. Please represent this as a separate unconfirmed/unknown outcome, or redefine the exported outcome model consistently; a possibly-delivered send must not inflate a definite drop rate.

4. The latest commit also reformats nine unrelated files under `analytics`, `auth`, `lark`, `middleware`, and `service`. Please remove that mechanical churn so this PR remains scoped to WeCom outcome reporting.

The previously reported `origin_not_channel`, pending-shed accounting, file-only reason propagation, and production collector coverage are fixed. The contributor's six targeted tests and the Prometheus collector test pass, current CI is green apart from the conditional `image-budget` skip, and the branch merges cleanly with current `main`. The three temporary regression tests described above fail consistently and were removed after verification.

If the outcome model remains wrong, every WeCom deployment using these counters for alerting or diagnosis can be misdirected; path tests and log/counter comparisons expose the problem, and rollback is a code-only PR revert with no database or persisted-format impact.

seacen:
All four addressed. The first three were the same principle applied to three more places — a counter may only assert what its branch can actually know — and this round taught me it cuts both ways: the admission gate was over-asserting (a reply failure it could not know) in the exact spot where, two rounds ago, I was under-asserting.

**1.** The admission gate now records its scheduling shed and nothing else. `TestAdmissionShed_EmptyContentZeroFilesAssertsNoReplyOutcome` is your scenario verbatim: empty content, zero bound files, full admission cap — `outbound_dropped` stays 0, and no reply outcome of any kind is asserted from a gate that never ran the lookup.

**2.** Both post-lookup exits settle one outcome per known file. `TestSlotWaitTimeout_SettlesEveryKnownFile` (three rows, expired wait → 3× `transport_error` + one reply outcome) and `TestNoLiveSender_SettlesEveryKnownFile` (three rows, empty registry → 3× `no_live_connection` + one reply outcome). Reverse-verified: reverting the per-row loop alone fails exactly the latter.

**3.** Unknown is now its own exported outcome: `multica_wecom_outbound_unconfirmed_total` and `..._attachment_unconfirmed_total`, reasons `ack_timeout` / `write_attempted` / `interrupted`, mirroring `sendOutcome`'s `deliveryUnknown` set exactly. `ack_timeout` leaves the drop reason set entirely, so both dropped help texts are now true as written; a file-only reply whose every file is unknown settles as unconfirmed; and the definite-drop aggregation no longer admits unknowns. `TestUnknownOutcome_IsUnconfirmedNotDropped` pins the whole chain, and the exported-contract test covers the two new collectors.

**4.** The nine reformatted files are restored byte-for-byte to `main`'s copies; the diff is WeCom-scoped again.

Green under `-race` across `wecom`, `internal/metrics` and `cmd/server`.

seacen:
Two things in one, since the branch has moved twice since your last look.

**Your four are already in, from the round before this one** (`6d404fd38`, pushed 8-28 shortly after your review of `b77f8d424`). Briefly, so you can start from the diff:

1. The pre-lookup admission gate records `attachmentShed()` and nothing else — no per-file count, no reply outcome. It runs before the lookup and knows neither how many files the turn carries nor whether it carries any, and the comment now says that is why.
2. All three post-lookup early exits settle every known row: the pending shed, the slot-wait timeout, and the no-live-sender branch each run one `attachmentDropped` per row, plus at most one reply outcome and only when the files were the reply.
3. `unconfirmedReason` is consulted before `classifyDrop` on every path, and `ack_timeout` / `write_attempted` / `interrupted` go to `outbound_unconfirmed_total` and `outbound_attachment_unconfirmed_total`. The collector help text for the dropped counters now holds: every reason there means the answer is not coming.
4. The nine unrelated reformats are reverted; the PR touches ten files, all WeCom outcome reporting.

**And a new commit** (`c4ca08d1e`), from re-reading the reason set against its own call sites rather than from a new finding of yours. Three drifts, same shape as the ones you have been catching:

- `attachment_not_admitted` said *"a file only, never a whole reply"* while the pending shed correctly files it as the reply's outcome too — when the files **were** the answer, that shed is the user getting nothing. The code was right and the sentence was left over from before `replyFailed` existed. It now says which unit means what.
- `transport_error` said *"the write itself failed, or the lookups ahead of it did"*, which does not cover its third caller: a delivery whose own budget ran out before it got a turn on the wire. Same underlying fact, so the same reason — the sentence was just short.
- The unknown side had no aggregation rule. A reply with several unconfirmed files reported whichever reason the loop ended on, which is the objection that made the definite side a documented precedence. `worseUnconfirmedReason` is its twin, ordered by how much each outcome establishes about the frame: `ack_timeout` > `write_attempted` > `interrupted`.

`TestSendAttachments_EveryEarlyExitSettlesBothUnits` covers all three early exits in one table, asserting both units on each — the three were fixed in separate rounds and each fix left the others unasserted, which is how the second stayed wrong after the first was corrected. Reverse-verified: dropping the slot-timeout per-row settle gives `attachment_dropped = 0, want 3`; dropping the pending shed's reply settle gives `outbound_dropped = 0, want 1`.

One limit worth stating rather than leaving to be found: `worseUnconfirmedReason` is covered as a unit, and the call site's use of it is not — constructing three attachments that each fail unknown for a different reason costs more rig than the rule is worth. The definite twin has the same shape and is exercised end to end.

Verification on `c4ca08d1e`, against a database migrated through 439: `go test -race ./internal/integrations/wecom ./internal/metrics ./internal/events -count=1`, `gofmt`, `go vet`, `git diff --check`. CI green, 14/14 with `image-budget` skipped.

One note for whoever picks up the stack: #7429 carries this branch's commits, so it has been merged forward to `df99f538c` and its head now includes `c4ca08d1e`. Its own behaviour is unchanged by that merge — I mention it only so the two PRs are not read against two different versions of `outbound_outcome.go`. Its verification comment names the pre-merge commit; the suite and CI are green on the merge too.

Bohan-J (approved):
Approving. Thanks for staying with this through four rounds — the counters that came out the other end are meaningfully better than what we asked for at the start, and the last two rounds in particular did the hard part: separating "we chose not to send this" from "we owed this and failed", and then separating "it definitely failed" from "we don't know". Those are the distinctions that make the difference between a dashboard an operator can act on and one they learn to ignore. WeCom is now the only channel adapter with real outbound delivery accounting; Slack, Lark and DingTalk still have bare `return nil` and a WARN.

I pushed one commit on top of yours (`aaf7c47`), comment-only:

- `server/internal/integrations/wecom/metrics.go` — `RecordOutboundDropped` still said some of its reasons were ordinary and that the counter should not be read as an error total. That stopped being true when `origin_not_channel` and `installation_inactive` moved to `RecordOutboundSkipped`.
- `server/internal/metrics/wecom.go` — the package header still described a single outbound *pair* and listed the web UI and revoked installations among the causes behind it, with no mention of `outbound_unconfirmed_total`. It now names each cause under the counter that actually holds it.

Both are what the next reader — and the operator reading `/metrics` help text — would otherwise have trusted. `gofmt`, `go vet`, `go build ./...` and the full `internal/integrations/wecom` + `internal/metrics` suites pass locally.

Two items I am deliberately **not** blocking on, filed as follow-ups rather than a fifth round. Neither changes behaviour, neither touches a metric name, a reason label or a help string, so the cost of fixing them after merge is the same as before it:

1. `outbound.go:143` hands every `processEvent` error to `unconfirmedReason`, and `outbound_outcome.go:113` returns `interrupted` for any wrapped `context.DeadlineExceeded`. A deadline that expires during the pre-send lookups is a definite local failure, but lands in `unconfirmed`. It needs the whole pre-send chain to blow the 10s budget to fire, and `errWriteAttempted` is already the right shape for the fix — gate `interrupted` behind a marker that proves the send actually started.
2. `outbound_media.go:364` settles a file-only reply as `dropped` whenever `failed > 0`, including when an `unknown` file may have reached the user. The per-file counters already record this exactly, so nothing is lost; the reply-level rollup is just lossier than it could be. Worth a short discussion on which way a mixed reply should round, since making it `unconfirmed` hides a file we know failed.

Merging.

PR #7429, opened 2026-08-22
MUL-6570 feat(wecom): route a reply to the replica that holds the bot's socket

seacen:
> Stacked on #7428 — it uses the outcome counters that PR defines (delivered / dropped / unconfirmed and their attachment equivalents). Review that one first; this diff reads against it.

## The failure

WeCom is the one channel with no outbound REST path. Every write goes over the aibot WebSocket, and the WS lease means exactly one replica holds it. `chat:done` is published on the **in-process** `events.Bus` by whichever replica served the daemon's `POST /tasks/{taskId}/complete` — a load-balancer decision with no relation to the lease. When they differ, the reply is dropped: the failure shape in #7215 and #6890, and the documented single-replica exception in `SELF_HOSTING.md`.

## The fix

Route it over the Redis Stream relay the repository already runs for daemon wakeups (`ScopeDaemonRuntime` → `ScopeWecomOutbound`, scope id = installation id):

```
B publishes chat:done → B's registry is empty
                      → XADD under ScopeWecomOutbound
                      → every replica's shard reader hands it to a bounded dispatcher
                      → the replica owning the socket claims it in Redis and sends
```

## The guarantees, and where each came from review

- **Never on the shard read loop.** `DeliverWecomOutbound` only enqueues; queues are sharded per installation (per-bot ordering, no cross-bot blocking); a full queue sheds under `relay_overflow`.
- **Idempotent across restart and lease movement.** The claim is a Redis `SET NX` keyed on the turn; its TTL is **derived** (`max(1h, 2×ReplayGrace)`) because the replay window is an operator knob, and a claim that expires while its frame is still replayable re-sends an answer the user already read.
- **Ownership gates the claim, and the loser retries.** Each replica reads a stream frame exactly once, so a released claim wakes nobody by itself. A lost race or a released claim is retried locally — bounded attempts, doubling backoff — which is what rescues the mid-flight lease move where the claim loser is the replica that *now* holds the socket.
- **The claim is released only for provably-unsent failures.** Attempted writes, missing verdicts and interrupted waits keep the claim and file as *unconfirmed* (#7428's outcome model), never as drops.
- **The drain runs while the sockets are live.** `main` stops the relay and joins the dispatcher **before** the channel supervisor teardown; supervised connections clear their senders on exit, so any later drain would find nothing to deliver with.
- **Counters keep their units.** A routed `inbox:new` moves no reply counter in either direction; the delivered/dropped ratio tracks outcomes, not socket placement.

## What this deliberately does not do

The relay's consumers are tail-followers, not a consumer group. A reply produced while **no** replica holds a live connection (all mid-reconnect) is still lost. That residual window is durability, not routing — the coupling that sank #6581 and #6750 — and #7428's `no_live_connection` counter is what will size it. The boot log, `SELF_HOSTING.md` and all four `channels*.mdx` locales state the same three-way contract: multi-replica supported under a sharded/dual relay; legacy/no-Redis keeps the single-replica rule; the no-holder window exists in every mode.

## Verification

Two-replica, DB-backed suite against a migrated database (real `*db.Queries`, independent buses and registries per replica), plus component tests per guarantee: startup replay, mid-delivery lease transfer, claim retention on ambiguous writes, dispatcher backpressure and shard-reader non-blocking, drain-with-live-sockets vs drain-after-teardown, inbox counter neutrality, and the TTL/replay-window invariant. Reverse-verified at each round: reverting a single guard fails exactly its own test. Green under `-race` across `wecom`, `realtime`, `cmd/server`, `internal/metrics`.


multica-eve (changes requested):
Thanks for tackling the cross-replica delivery gap. The product direction is valuable, but the current implementation has four blocking correctness issues:

1. **The relay reader starts before the WeCom deliverer is registered.** `server/cmd/server/main.go:437` starts the relay goroutines, while `server/cmd/server/router.go:924-928` installs the deliverer later. The field in `sharded_stream_relay.go:196-198` is read and written without synchronization, so startup replay can be silently discarded while the deliverer is nil, and the setter races with the reader. Please register before `Start`, or provide a synchronized registration mechanism that cannot lose replayed frames, with a race regression test for the real startup order.

2. **The in-memory seen-set does not provide cross-replica or restart idempotency.** The sharded relay intentionally replays the last five minutes on startup (`sharded_stream_relay.go:39-43`), but `relay_outbound.go:65-99` stores claims only inside one process. A restarted replica can therefore resend an already delivered reply, and a lease move between two replicas reading the same frame can let both processes claim and send it. `TestTwoReplicas_RelayDeliversExactlyOnce` covers only a static lease with a synchronous fake. Please add an idempotency mechanism that spans these windows and regression tests for startup replay and mid-delivery lease transfer.

3. **A WeCom acknowledgement wait blocks the shared realtime shard reader.** `sharded_stream_relay.go:473-475` invokes the deliverer synchronously, and `relay_outbound.go:177` waits for the platform acknowledgement (up to five seconds in `ws_sender.go:209-220`). One unhealthy bot can consequently delay browser realtime traffic, daemon wakeups, and unrelated bots mapped to the same shard; batches of timeouts compound linearly. Please move this external delivery behind a bounded, shutdown-aware dispatch path and test backpressure and ordering.

4. **The routed path drops attachments and can produce no final outcome at all.** `outbound.go:238-247` relays text and returns before attachment delivery at `:260-275`; a file-only reply attempts an empty text send while the actual file never reaches the user. If every replica lacks a live sender, all consumers also return silently (`relay_outbound.go:170-172`), so neither delivered nor dropped is recorded—the code explicitly notes this at `outbound.go:233-237`. Please cover text+file, file-only, and no-lease-holder cases and ensure each reply has an interpretable final result.

`inbox:new` remains local-only, while the startup warning and self-hosting/channel docs still require a single WeCom replica. Please either route that remaining user-visible path as well or narrow the product claim and documentation explicitly. This PR is also stacked on #7428; please address that base review and rebase before the next review round.

seacen:
All four addressed, plus `inbox:new`. Rebased onto the updated #7428.

You were right about the shape of the mistake: I copied the daemon-wakeup pattern without reading what that mechanism requires of a consumer. Three of the four findings are that one omission.

**1. Registration before `Start`.** The dispatcher is now built and registered in `main.go` *before* `relay.Start`, and the senders registry moves to boot so it can be — it is a bare map with no dependencies, so it can be minted that early and handed down through `RouterOptions`. `Attach` supplies the handler once the subscriber exists; frames read in between wait in the queue rather than hitting a nil field. That also removes the unsynchronised write against a running reader — the field is gone, and the one value that genuinely arrives late (the metrics sink) is an `atomic.Value`.

**2. Idempotency that spans a restart and a lease move.** The claim is a Redis key keyed on the turn (`wecom:outbound:claim:{event id}`, `SET NX`, TTL well past `ReplayGrace`), so a replayed frame and two replicas reading the same frame meet the same key. Redis is not a new dependency for this path — it *is* the transport: no Redis means no relay, no cross-replica delivery, and nothing to deduplicate. The in-process set stays only as a cheap first gate for the publisher's own copy of its own frame.

Two judgements in there worth your eye. A claim store that cannot answer **withholds** rather than delivering: the frame is still replayable, so that costs latency, while the alternative costs a duplicate answer in a room. And the claim is released only when the frame **provably** never reached the wire — never on an ack timeout, since `errAckTimeout`'s own contract is that the message may well have arrived.

New coverage: `TestRelay_StartupReplayDoesNotResend` (a fresh process, an empty local set, the window replayed into it), `TestRelay_LeaseMoveDoesNotDoubleSend` (two replicas both holding a socket), `TestRelay_DedupeOutageWithholdsRatherThanDuplicates`.

**3. The shard reader is never blocked.** `DeliverWecomOutbound` only enqueues. Queues are sharded by installation, so one bot's replies keep their order while unrelated bots cannot block each other; a full queue sheds under its own reason (`relay_overflow`) rather than stalling the shard; shutdown drains what is already in hand. `TestRelay_DispatchNeverBlocksTheShardReader` uses a double that never acks — a synchronous implementation would sit in `ackTimeout` — and `TestRelay_ShedsRatherThanStalls` covers the backpressure path.

**4. The routed path is the whole reply.** Attachments are delivered by the replica holding the socket, which reads the rows itself rather than shipping files through Redis, and a file-only reply is routed and settled as exactly one outcome (that machinery came with the #7428 round). `deliverAttachments` split into an id-taking variant for this.

**`inbox:new` is routed too.** You were right that leaving it local means the constraint has to stay, and I would rather route it than narrow the claim. Its claim is keyed on the notification, not the moment it was read.

**What I did not fix, stated rather than left implicit:** the relay's consumers are tail-followers, not a consumer group with pending-entry reclaim. A reply routed while *every* replica is mid-reconnect is read by nobody and still lost. That window is durability, not routing, and it is the half that made a durable queue too large to land in #6581 and #6750. #7428's `no_live_connection` counter is what would size it. If you want the single-replica warning and the `SELF_HOSTING.md` exception rewritten in this PR rather than left alone, say which way — I left both untouched since that residual window means "single replica" is still the safest advice for anyone who cannot tolerate a rare lost reply.

Green under `-race` across `wecom`, `realtime`, `cmd/server` and `internal/metrics`, with the two-replica tests against a migrated database.

multica-eve (changes requested):
Thanks for the substantial rework. I re-reviewed head `f0c35f5`; registering the deliverer before the relay readers, moving external waits off the shared reader, and carrying attachments/inbox notifications are all improvements. There are still correctness blockers in the routed delivery path:

1. **A non-holder can win the global claim and strand the actual socket holder.** Every replica receives every WeCom outbound frame, but `RelayOutbound.perform` acquires the Redis `SET NX` claim before `deliverRelayed` checks the local sender registry. If a replica without the socket wins first, the real holder concurrently sees `won == false` and returns. The non-holder then releases the key, but nothing retries or wakes the holder that already returned, so the reply is silently lost. I reproduced this deterministically by blocking the non-holder after its claim, letting the holder lose its claim, and then releasing the non-holder; the holder received zero delivery calls. Please check local ownership before competing for the cross-replica claim and add this ordering as a regression test.

2. **Ambiguous WebSocket write failures are treated as safe to resend.** `ws_sender.go` deliberately wraps errors raised after entering `WriteMessage` with `errWriteAttempted`, and `sendOutcome` correctly classifies that (plus context cancellation/deadline while waiting) as `deliveryUnknown`: the peer may already have acted. However, `deliverRelayed` maps every `dropTransport` to `outcomeProvablyNotSent`, releases the Redis claim, and allows a replay to send the same answer again. Please retain the claim for unknown outcomes and release it only for a transport failure proven to occur before the write; add routed-text tests for attempted-write and cancellation/deadline cases.

3. **The production shutdown path does not wait for the dispatcher it claims to drain.** `RelayOutbound.Start` drains queued work after cancellation and exposes `Wait`, but `server/cmd/server/main.go` waits only for the realtime relay and then closes `relayWriteRedis` without calling `wecomRelayOutbound.Wait()`. Queued deliveries therefore race Redis/client teardown and are not gracefully drained. Please join the dispatcher before closing its dependencies, with a bounded total drain deadline so a full shard cannot extend shutdown by many sequential 5-second ack waits.

The operational contract also needs to match the implementation before merge: `router.go` still unconditionally warns that cross-replica routing is not implemented, while `SELF_HOSTING.md` and the `outbound.go` package comment still require single-replica WeCom. Please document the actual boundary (enabled for sharded/dual relay modes; legacy/no-relay retains the limitation).

Finally, this head still contains the unmerged #7428 commit, so it remains blocked on that PR's outstanding attachment-outcome/metrics fixes as well.

seacen:
All three fixed, the operational contract now matches the implementation, and the branch is rebased onto #7428's updated head (which addresses that PR's attachment-outcome round).

**1. Ownership gates the claim.** Your reproduction was exactly right, and the fix is the ordering you named: a replica competes for the `SET NX` claim only after finding the socket in its own registry. A non-holder now never touches the key, so the holder cannot lose a race to a party that could not honour the win. The socket vanishing between check and send is why `deliverRelayed` still re-checks and why a not-ours outcome releases. `TestRelay_NonHolderNeverCompetesForTheClaim` pins it from your scenario's angle: a non-holder processes the frame first, makes zero claim attempts, and a holder arriving later via replay still delivers. Reverse-verified — removing the gate alone fails exactly that test.

**2. Ambiguous write failures keep the claim.** You are right that mapping every `dropTransport` to a release contradicted `errWriteAttempted`'s own contract. The release now rides on a `provablyNotSent` classifier that follows the boundary `ws_sender` already marks: a stated refusal, a missing verdict, an attempted write, and a bare context error (ambiguous by construction — `request()` returns one both before the write and while waiting after it) all retain the claim; only a failure proven to precede the write gives it back. Costs an un-retried delivery in the worst case rather than a duplicate in the chat. Covered by `TestRelay_AttemptedWriteKeepsTheClaim` (a conn that fails inside `WriteMessage`; the claim must survive), `TestDeliverRelayed_ContextErrorsDoNotRelease`, and `TestProvablyNotSent_Classification` pinning the whole boundary.

**3. Shutdown joins the dispatcher, bounded.** `main.go`'s teardown now calls `wecomRelayOutbound.Wait()` after the relay stops and **before** the Redis clients close, and the drain itself runs under one budget for the whole drain (`relayDrainBudget`, 10s) rather than per delivery — deliveries started near the edge inherit the deadline, so a full shard cannot stack sequential ack waits.

**The contract:** the boot message now states the actual boundary — with a sharded/dual relay it logs that cross-replica routing is enabled (naming the residual no-holder window); in legacy/no-Redis mode the original single-replica warning stands, now pointing at the relay as the way out. `outbound.go`'s package header and `SELF_HOSTING.md` say the same thing in their own registers, including that a single replica remains the most conservative choice for anyone who cannot tolerate a rare lost reply during reconnects.

Green under `-race` across `wecom`, `realtime`, `cmd/server` and `internal/metrics`, with the two-replica suite against a migrated database.

seacen:
CI note: the previous head's `backend-tests` failure was not the review fixes — it was #7468 (MUL-6661) landing on `main` after this branch was cut. That PR moved the outbound address from the session binding to a per-task `channel_task_delivery` snapshot, so on the merge ref my DB fixtures seeded a turn the new `GetChannelTaskDelivery` lookup correctly ignored, and every two-replica test timed out against an empty socket.

Both branches are now rebased onto current `main` (`3d37828e9`): the instrumentation and the relay branch sit on the delivery-snapshot structure, and the fixtures seed the `channel_task_delivery` row the engine writes on ingest. Full suite green under `-race` against a database migrated through 428, including all thirteen relay/two-replica tests.

The review-round fixes described above are unchanged by the rebase.

Bohan-J (changes requested):
Re-reviewed at `070dd133`.

The product direction still holds, and several blockers from the previous round are genuinely fixed: the deliverer is registered before `Start`, delivery is off the shard read loop on a bounded dispatcher, there is a cross-replica Redis claim, attachments and `inbox:new` are routed, ownership is now checked before the claim, the startup log is conditional on relay mode (`router.go:1050-1052`), and `SELF_HOSTING.md:380-385` states the new contract honestly, residual window included.

Four things still block a merge. All four sit on the main delivery path.

## 1. A lease that moves mid-flight strands the reply, because nothing retries a released claim

`perform` (`server/internal/integrations/wecom/relay_outbound.go:365-408`) checks ownership first (`:376`) and only then competes for the global claim (`:384`). That ordering fixes the previous round's finding, but the check and the send are still two separate moments — the comment at `:370-378` says so — and what it does not account for is that the loser never comes back:

- A holds the socket. A reads the frame and passes `ownsSocket`.
- The lease migrates to B.
- A wins `dedupe.Claim`.
- B reads the frame, passes `ownsSocket`, loses the claim, and returns at `:394-396`. B is done with this frame permanently.
- A reaches `deliverRelayed`, finds no sender, returns `outcomeNotOurs`, and releases the claim at `:404-406`.

The claim is free and the reply is gone. "Release so the replica that does hold the socket can take it" only works if some consumer reads the frame again, but each replica reads a given stream frame once and the reader replays only on restart. The same assumption breaks the dedupe-error path at `:386-393`: `forget` + return is justified as safe because "the replay window will bring this frame back", which it will not do for a process that is still running.

A released or un-acquirable claim needs a bounded, idempotent in-process retry — re-enqueue with a short capped backoff — rather than an assumed replay. A regression that migrates the socket between the ownership check and the claim would pin it; I reproduced the delivery count as `0` (expected `1`) with a temporary deterministic interleaving.

## 2. `dedupeTTL` is a constant while the replay window is operator-tunable

`dedupeTTL = time.Hour` (`relay_outbound.go:81-84`) justifies itself with "the relay's default grace is minutes". True of the default, but `REALTIME_RELAY_REPLAY_GRACE` (`server/cmd/server/main.go:89`) accepts any positive duration, `Validate` only requires `> 0` (`server/internal/realtime/sharded_stream_relay.go:134-136`), and retention widens with it (`StreamTTL = TrimHorizon + ReplayGrace`, `:97-104`).

Set the grace to `2h` and the invariant the constant exists to uphold is inverted: the claim expires at 1h while the frame is still replayable, so a restart at 90 minutes re-sends an answer the user already read. Derive the TTL from the normalised relay config (e.g. `max(1h, 2 × ReplayGrace)`), or reject a configuration whose grace exceeds what the claim can cover — and cover the invariant with a test.

## 3. The drain runs after the sockets are already gone

`Start` drains on cancel under `relayDrainBudget` (`relay_outbound.go:284-303`), and the defer at `main.go:386-403` does correctly join the dispatcher before the Redis clients close. But that defer runs only once `main` returns, and by then the body has already executed `sweepCancel()` (`main.go:767`) and joined `ChannelSupervisor` (`:788-793`). Each supervised connection clears its sender on exit (`wecom_channel.go:198`), so at drain time `ownsSocket` (`relay_outbound.go:376`, `:480-489`) is false for every installation. Every queued reply returns at the first statement of `perform` and is discarded — without even a counted outcome, since that early return records nothing. The graceful drain is real code that cannot fire in production.

Stop the relay reader and drain the dispatcher while the connections are still live, ahead of the supervisor teardown, and pin the ordering with a lifecycle test in production order. A unit test that cancels the dispatcher with sockets still registered cannot catch this.

Minor, same area: after cancel, the worker `select` at `:281-286` may pick a queued item rather than `ctx.Done()`, then call `dedupe.Claim` with an already-cancelled ctx and take the `:386` "dedupe unavailable" branch instead of the drain path.

## 4. A relayed inbox push is counted as a delivered agent reply

`deliverRelayed` calls `o.delivered()` at `relay_outbound.go:464`, before the `relayKindInbox` branch at `:466`. `RecordOutboundDelivered` is documented as "one reply reached the user … the denominator" (`wecom/metrics.go:56-60`), and the local inbox path records nothing on success or on send failure (`outbound.go:428-438`) — only the relayed path counts (`:453`, `:464`).

So the same inbox push counts as a delivered reply, a dropped reply, or nothing at all depending on which replica happens to hold the socket. The delivered/dropped ratio then moves with socket placement instead of with outcomes. That matters here specifically, because this PR's case for shipping routing before durability rests on those counters being trustworthy enough to measure the remaining window. Either skip the reply counters for `relayKindInbox` on both the success and failure paths, or give inbox its own pair and make the local path record them too.

## Contract and hygiene

- All four `apps/docs/content/docs/channels*.mdx` still state that WeCom requires a single backend replica "until cross-replica outbound routing lands" (`channels.mdx:101` plus the zh/ja/ko siblings). Within this same PR `SELF_HOSTING.md:380-385` says the opposite for sharded/dual mode. The user-facing docs need the same three-way split.
- The PR description still leads with "No deduplication ledger, deliberately" and a test table that predates the dispatcher, the Redis claim, and the attachment/inbox routing. Worth rewriting before merge — it is what the next reader will trust.
- `b77f8d4` carries pure gofmt churn in packages this change does not touch: `internal/service/empty_claim_cache.go`, `internal/integrations/lark/*`, `internal/analytics`, `internal/auth`, `internal/middleware`. Please split it out.
- `relayShards = 8` / `relayQueueDepth = 256` / `relayDrainBudget = 10s` (`relay_outbound.go:86-99`) are asserted without a sizing argument or an operator override. Relatedly, the comment at `:44` — "unrelated bots cannot block each other" — is not accurate: two bots hashing onto the same shard serialise, and each send can wait up to the 5s ack. The comment at `:86-89` describes the real property correctly; `:44` should match it.
- Still stacked on #7428, whose attachment outcome/metric findings are open. That base needs to land fixed, or be rebased away, before this can merge.

## Verification

The existing WeCom relay tests, `internal/realtime` and `internal/metrics` pass under `-race`, and CI is green on this head. Blockers 1 and 4 were reproduced with temporary deterministic regressions (holder delivery count `0`, expected `1`; `outbound_delivered=1`, expected `0`); those scratch files were deleted and the tree is clean. Blockers 2 and 3 are static — configuration and shutdown ordering respectively.

**Request changes.**

seacen:
All four addressed, plus the docs and the description. Re-review was exactly right on all of them — thank you for reproducing #1 rather than arguing it.

**1. The loser retries, because nothing else ever will.** Your trace named the real gap: each replica reads a stream frame exactly once, so both "the holder will take it" and "the replay will bring it back" are false for a running process. A lost claim race and a released claim now both requeue locally — bounded (`relayMaxAttempts` 5), doubling backoff from 200ms — which is precisely what rescues your interleaving: the loser of the race is the replica that *now* holds the socket, and its retry finds the claim free. The dedupe-outage path retries the same way instead of praying for a replay. Exhausting the budget logs and does not count a drop, because by far the likeliest reason to exhaust it is a claim held by a delivery that already happened. `TestRelay_LeaseMoveMidFlightIsRescuedByRetry` reproduces the migration with a handler whose ownership answer and delivery disagree, and asserts exactly one frame lands.

**2. The TTL is derived, not constant.** `dedupeTTLFor = max(1h, 2×ReplayGrace)`, fed from the relay config at construction. `TestDedupeTTL_CoversTheReplayWindow` pins the invariant, including your 2h-grace inversion.

**3. The drain runs while the sockets are live.** `main` now stops the relay readers and joins the dispatcher **in the shutdown body, before the channel supervisor teardown** (a `sync.Once` keeps the defer as a fallback for early-return paths). Your minor is fixed too: a worker that picks an item racing the cancel hands it to the drain instead of misfiling it as a dedupe outage, and a worker cancelled at start still drains what is queued if the handler is attached. On the lifecycle test: I could not find a harness that exercises `main`'s actual shutdown sequence, so the ordering itself is pinned by the call-site comment, and `TestDrainDeliversOnlyWhileSocketsLive` proves the mechanism the ordering exists for — the same queued frame drains to the wire with the socket registered and to nothing with it cleared, which is what supervisor teardown does. If there is an existing pattern for a production-order lifecycle test in this repo I am happy to add one.

**4. Inbox pushes move no reply counter,** success or failure, on the relayed path — and the failure path logs instead. Chosen over a dedicated inbox pair to keep this PR's metric surface unchanged; if you would rather have the pair (with the local path recording it too), say so and I will add it here or in #7428. `TestRelayedInboxPushDoesNotMoveTheReplyCounters` covers both directions.

**Contract:** all four `channels*.mdx` locales now state the same three-way split as `SELF_HOSTING.md` (sharded/dual → multi-replica supported; legacy/no-Redis → single replica; the no-holder window in every mode). The PR description is rewritten to match the implementation as reviewed.

Also rebased onto #7428's current head; the ambiguous-write handling now files under its unconfirmed outcome model rather than as drops, which is the same boundary `provablyNotSent` already drew for the claim.

Green under `-race` across `wecom`, `realtime`, `cmd/server`, `internal/metrics`, with the two-replica suite against a migrated database.

multica-eve (changes requested):
This change should let a WeCom user receive every answer in the original conversation even when the service runs in multiple copies. The current head can still deliver two answers in the wrong order, lose a pending answer during an update, and report delivery health inaccurately, so that promise is not yet met.

The product direction is accepted. Re-reviewed at `ef2d5d16e`: deriving the claim TTL from the replay window, keeping direct relayed inbox pushes out of reply counters, aligning the four channel docs and PR description, and adding the mid-flight lease regression are all meaningful fixes. The following issues still block a merge.

## 1. The new retry path breaks the promised per-installation ordering

`server/internal/integrations/wecom/relay_outbound.go:477-502` schedules a failed item with `time.AfterFunc` and appends it to the shard queue later. The worker immediately continues with the next item, so a later reply for the same installation can be sent before the retried one.

I reproduced this deterministically with two replies for one installation: the first claim returns a temporary Redis error, the second succeeds, and the observed delivery order is `[second, first]` instead of `[first, second]`. That directly contradicts the ordering contract at `relay_outbound.go:43-44` and in the PR description. Please preserve head-of-line ordering per installation across retries, and add this exact failure ordering as a regression.

## 2. The production shutdown still cancels the socket owners before the dispatcher drain

The call-site comment says relay first, supervisor second, but the actual order is:

- `server/cmd/server/main.go:782` calls `sweepCancel()`.
- `server/internal/integrations/channel/engine/supervisor.go:373-376` reacts by calling `cancelAll()` immediately.
- each WeCom connection clears its sender on exit at `server/internal/integrations/wecom/wecom_channel.go:196-198`.
- only later, after other worker waits, does `main.go:805` call `stopRelay()` and drain the dispatcher.

Putting `stopRelay()` before `Supervisor.Wait` is not enough; teardown already started at `sweepCancel()`. Move the relay stop/drain before `sweepCancel()` while the sockets are still live, and add a lifecycle test that pins the production call order rather than manually retaining or clearing the registry.

## 3. Reply outcome metrics are still both contaminated and blind

The direct `deliverRelayed` inbox branch is fixed, but the surrounding paths still violate the same unit:

- initial queue overflow at `relay_outbound.go:401-406` and retry overflow at `:495-500` record `outbound_dropped{relay_overflow}` without distinguishing `reply` from `inbox`;
- an inbox relay publish failure at `server/internal/integrations/wecom/outbound.go:419-430` calls the agent-reply drop recorder;
- when no replica owns the socket, the publisher returns after a successful relay publish and every consumer returns at `relay_outbound.go:430-431`, so the lost reply is counted nowhere.

A deterministic overflow regression records `outbound_dropped=1` for an inbox notification (expected `0`). A two-replica/no-holder regression records cluster `no_live_connection=0` for the one reply that was actually lost (expected one final outcome). The latter also contradicts the PR description's claim that #7428's `no_live_connection` counter can size the residual no-holder window.

Please separate relay admission/inbox signals from agent-reply outcomes and establish one owner for the end-to-end final outcome so a delivered reply is counted once, a truly lost reply is counted once, and copies discarded by non-holders are not counted as failures.

## 4. Retry and capacity policy is still unsubstantiated, and retry timers are outside shutdown ownership

`relayMaxAttempts = 5` with a `200ms` initial backoff covers only about three seconds, while `relayShards = 8`, `relayQueueDepth = 256`, and `relayDrainBudget = 10s` also have no traffic, platform-latency, Redis, or lease bound supporting them. Eight hashed queues also mean unrelated bots on the same local shard do block each other, so the claim at `relay_outbound.go:44` is false.

These are production policies, not incidental constants. Please provide the sizing rationale and boundary tests, and derive/configure values where the operational bound is external. Also, `time.AfterFunc` callbacks are not tracked by `RelayOutbound.Wait()`: cancellation can occur after a callback checks its context but before it enqueues, leaving an item in a queue whose worker has already drained and exited. Timer ownership must be included in shutdown/admission closure.

If these behaviors ship incorrectly, multi-replica sharded/dual WeCom deployments can show users missing, reordered, or duplicate answers while the health counters simultaneously report the wrong outcome. Rollback is the previous release or temporarily returning to legacy/single-replica operation.

Existing targeted relay tests pass under `-race`, `internal/realtime` and `internal/metrics` pass, `git diff --check` passes, and GitHub CI is green. The three review regressions above fail on this head; the temporary tests were removed and the tree is clean. The stacked base #7428 also remains open with `CHANGES_REQUESTED`.

**Request changes.**

seacen:
All four addressed at `c866723e8`, each with the regression you asked for, and each reverse-verified against the old behaviour one guard at a time.

Two of the four were comments asserting something the code did not do, which is worth saying plainly: the ordering claim and the shutdown-order claim were both written by us and neither was ever checked against the thing it described.

## 1 — Ordering across a re-offer

You were right, and the fix had to be structural rather than a smaller backoff. Each installation now keeps a line inside its shard worker: a frame owed another attempt parks at the **head** of that line and everything that arrives behind it waits. A re-offer never goes to the back of the queue, so a later reply for the same bot cannot overtake it.

`TestRelay_ARetriedReplyKeepsItsPlaceInLine` is your scenario against a real database and a real Redis: two turns on one installation, the first claim hitting a Redis fault, delivery order asserted on the wire. Reverse-verified by removing just the line-joining — the chat then receives `[答案二]` alone, so the old shape both reordered and lost the first answer.

The header comment is corrected too. Sharding is a concurrency bound, not isolation: installations that collide on a shard **do** wait behind each other, and that now says so.

## 2 — Shutdown order

Also correct, and worse than the comment admitted: `sweepCancel()` is not merely earlier than the drain, it is the thing that makes every supervised connection clear its sender. So the drain ran against an empty registry every time.

The sequence is now a named type (`server/cmd/server/shutdown.go`) whose field order is the contract, with the relay stop ahead of the cancel. `main()` runs it; `TestShutdownSequence_*` pins it — one test on the full order, one on the relay/cancel pair alone so a future reshuffle fails with the reason attached rather than as a diff of two long lists. Reverse-verified by swapping those two steps: both tests fail.

No registry is retained or cleared by hand anywhere in it.

## 3 — Reply outcome metrics

Three separate unit errors, three separate fixes.

- **Overflow, both paths.** A shed frame now moves `outbound_dropped{relay_overflow}` only when it is a reply. Every shed, reply or inbox, moves a new `multica_wecom_outbound_relay_shed_total{kind}` — an admission decision with its own unit. `TestRelay_AnInboxPushShedDoesNotMoveTheReplyCounters` asserts `outbound_dropped = 0` for an inbox overflow; `TestRelay_ShedsRatherThanStalls` now asserts the boundary itself, that `QueueDepth` frames are accepted and the next is not.
- **The unroutable inbox push.** `outbound.go` logs it instead of calling the agent-reply drop recorder. Same rule the relayed-inbox path already followed.
- **The end-to-end owner.** This is the one that needed a mechanism rather than a correction. Nobody could count a reply nobody delivered: the publisher returns after publishing, and every replica that reads the frame and holds no socket returns silently and correctly. So the publisher now settles it, once — after the whole re-offer chain can no longer be running, it asks whether the claim key was ever taken. Set means somebody delivered and counted it; absent means nobody could, and that is one `no_live_connection`. One Redis `EXISTS` per routed reply, and routed replies are the off-lease minority. An unreadable Redis records nothing, because a blip must not become a fleet of phantom drops.

  Two regressions, both on a real database and a real Redis: `TestRelay_AReplyNoReplicaCouldSendIsCountedOnce` (no holder anywhere → exactly one, across both replicas) and `TestRelay_ADeliveredReplyIsCountedOnceAndNotAlsoLost` (delivered → one delivered, zero drops on either side). Reverse-verified by removing the enrolment: the first drops to `no_live_connection = 0`.

  This also makes the PR description's claim true rather than removing it, which was the honest choice between the two. `SELF_HOSTING.md` and all four `channels.*.mdx` now say the window is counted and name the counter.

## 4 — Policy and timer ownership

**The numbers.** You were right that they were unsubstantiated, and checking them against the bound found one that was simply wrong. The re-offer chain exists to outlast a lease move; a lease moves when the next replica's supervisor sweep picks the installation up, which is `CHANNEL_WS_LEASE_POLL_INTERVAL` — 30s by default. Five attempts from 200ms covered about three seconds. It gave up an order of magnitude before the thing it was waiting for.

Everything is now a `RelayConfig` whose comments name the bound each field is sized against, and the two bounds that live outside this package are passed in rather than guessed: the relay's `ReplayGrace`, and the supervisor's poll interval. `engine`'s defaults are exported so the dispatcher references them instead of holding a copy free to drift. Boundary tests: `TestRelayRetryPlan_OutlastsALeaseMove` across three settle windows, `TestRelayRetryPlan_DefaultsCoverTheDefaultPollInterval`, `TestRelayDrainBudget_FitsUnderTheSupervisorShutdownTimeout`, `TestRelayOutcomeGrace_OutlastsTheReOfferChain`, `TestDedupeTTL_OutlivesEveryReplayWindow`, and `TestRelayShardFor_StaysInsideTheConfiguredQueues` for a non-default shard count.

**Timer ownership.** `time.AfterFunc` is gone. The wait is one timer owned by the shard worker itself, alongside the lines, so a pending re-offer is part of the worker by construction rather than by a context check racing it. That also fixed a worse case than the one you named: the old callback checked the context, found it cancelled, and returned — **discarding the frame** — so a reply that had failed once and was waiting to be retried was thrown away by the very shutdown that runs a drain so replies are not thrown away. The drain now flushes those lines first, without their backoff. `TestRelay_AReplyWaitingToBeRetriedIsStillDrainedOnShutdown` asserts the second attempt happens; reverse-verified, the old shape gives `delivery attempts = 1, want 2`.

One disclosure on that test. My first version of it asserted the narrow enqueue-after-exit race directly, and it passed against the old code — it was testing nothing. I replaced it with the shutdown-drain property above, which is deterministic and is the part a user would actually notice, and only then did it fail against the old shape.

## Verification

On `c866723e8`, against a database migrated through 439 and a real Redis:

- `go test -race ./internal/integrations/wecom ./internal/metrics ./internal/realtime ./internal/integrations/channel/engine ./internal/events ./cmd/server -count=1`
- `gofmt`, `go vet`, `git diff --check`
- CI green, 14/14 with `image-budget` skipped

The claim store in the new relay tests is `redisDedupe` against a live server rather than a map, since what is under test in the ordering and outcome cases is what `SET NX` and `EXISTS` actually do across processes. The single fault a live Redis cannot be asked to produce on cue — one failed `Claim` — is injected by a wrapper around the real implementation.

#7428 is unaffected by these changes: the inbox counter path in finding 3 is introduced by this PR, not by that one.

Bohan-J (approved):
Thanks for sticking with this one, @seacen — it took several rounds and you did the hard part of every one of them.

Worth naming what this landed. WeCom is the only channel with no outbound REST path, so an off-lease reply had nowhere to go and the documented answer was "run a single replica". That constraint is gone for any deployment with the sharded or dual relay, and the residual window (no replica holding a live connection) is now *measurable* rather than merely acknowledged — `outbound_dropped_total{reason="no_live_connection"}`, settled once by the replica that routed the reply. That last part is what makes the "route now, durability later" split honest instead of a hand-wave, and it was the right call to argue for it.

The engineering that stands out: ordering held across a re-offer via per-installation lines rather than a shared queue; the claim TTL derived from `ReplayGrace` instead of pinned to a constant; sizing every dispatcher value against an external bound (the supervisor's poll interval, its shutdown timeout) and saying so in the comment. Reviewers asked hard questions and the answers got written into the code, which is why the file reads the way it does.

I pushed four commits to your branch rather than sending you round five — all small, and all in the same spirit as your own work:

- `3c5a064` — three attribution fixes: `outcomeGrace` now pays one claim round trip per *offer* (a fully timed-out chain outlasted the grace, so the watcher recorded a loss the next attempt then contradicted); a shed on a replica that could not send no longer counted as a lost reply; and `drainRemaining` no longer lets a frame taken off the queue at cancel overtake one already parked at its installation's line.
- `8e298fc` — merge of `main` after #7428 landed. Everything #7428 touched conflicted, since this branch carried its pre-squash copy; resolution took main's reviewed form for #7428's own content and kept this PR's. It also dropped the unrelated gofmt churn, so the diff is now exactly this feature.
- `b353525` + `d1a3f48` — the last two review nits, both subtractions. A shed now records `relay_shed` and nothing else: conditioning it on "do I hold the socket" was still not proof, because two replicas hold a sender during a lease handoff. No replica can settle that locally, so none does — the routing replica's outcome watch already owns it, exactly once. And the claim round-trip budget moved onto `DedupeStore.ClaimBudget()`, since the config field described a bound it did not actually configure.

Verified with a local Postgres and Redis: `go test -race ./internal/integrations/wecom` passes with **0 skips** (330 tests) — including the DB-backed two-replica and ordering suites that earlier rounds could only reason about statically — plus `internal/metrics` and `internal/realtime`. Each fix was reverse-verified: reverting it fails its own regression and nothing else.

Approving and merging. Nice work.

Issue #7427, opened 2026-08-22
Tasks stuck forever in waiting_local_directory after mutex holder is watchdog-killed — agent queue freezes (capacity leak)


foXaCe:
## Summary

When a `local_directory` mutex holder is force-stopped (e.g. by the idle watchdog), the tasks waiting on that mutex stay in `waiting_local_directory` **forever** — they are never re-admitted, never failed, and keep consuming agent capacity. With enough stale waiters the agent's entire queue freezes: new tasks pile up in `queued`, including user-facing chat tasks.

Self-hosted v0.4.32 (`multica-backend-1` docker compose), measured on a single-PC daemon with 4 concurrent agents.

## Measured reproduction

Timeline from one incident (all times UTC):

1. 05:30–05:34 — three tasks of the same agent enter `waiting_local_directory` on depot mutexes held by other agents' tasks.
2. 06:24 and 06:43 — two of the mutex holders are force-stopped by the idle watchdog:
   ```
   agent result detail status=idle_watchdog
   agent_error="agent produced no new messages for 10m0s and message queue was empty; force-stopped by idle watchdog"
   ```
3. The waiters are still in `waiting_local_directory` more than 1 h 45 later, holders long gone:
   ```sql
   SELECT id, status, created_at FROM agent_task_queue
   WHERE agent_id = $me AND status IN ('dispatched','running','waiting_local_directory');
   -- 3 rows, all waiting_local_directory, created 05:30–05:34
   ```
4. `CountRunningTasks` counts exactly these states as capacity-bearing (`server/pkg/db/queries/agent.sql:1591`), so all N slots of the agent are consumed by dead waiters:
   - every subsequent task stays `queued`;
   - 11 tasks piled up over ~1 h, including two **chat** tasks (operator-facing questions) that could not run;
   - nothing in the CLI exposes why — `multica issue runs` / `multica agent tasks` only show `queued`, we had to query Postgres to understand.

The capacity-bearing choice itself is documented in the query comment ("removing it here alone could exceed max_concurrent_tasks when it resumes") — understood, and not what this report challenges. The leak is that **termination of the holder does not terminate or re-admit its waiters**.

Manual recovery that worked: cancelling the three stale waiters (`multica issue cancel-task <id>`). Slots freed immediately; the queued backlog started dispatching within seconds, including the stuck chat tasks.

## Expected behavior

One or both of:

[...]

PR #7430, opened 2026-08-22
MUL-6564: fix(daemon): free local_directory waiters after holder exits (#7427)

Duang777:
## Summary
- **Root cause (#7427):** `LocalPathLocker` was held through post-run `FailTask`/`CompleteTask` network I/O. When a holder was idle-watchdog-killed but the terminal callback stalled, waiters stayed in `waiting_local_directory` forever and leaked `CountRunningTasks` capacity. Separately, `FailStaleTasks` deliberately skipped this status even after the prepare lease expired, so orphaned waiters had no server-side reclaim path.
- **Daemon fix:** release the path mutex as soon as `runner.run` returns (before terminal reporting). The lock only needs to serialise prepare+agent writes.
- **Server safety net:** reclaim `waiting_local_directory` rows with an expired/absent prepare lease as `waiting_local_directory_abandoned` (auto-retryable), while live-lease waiters remain untouched.

## Test plan
- [x] `go test ./pkg/taskfailure/ -run 'TestReasonStringWireValues|TestAllReasonsContents|TestIsAgentError'`
- [x] `go test ./internal/daemon/ -run 'TestHandleTask_ReleasesLocalDirectoryLockBeforeTerminalReport'`
- [x] `go test ./cmd/server/ -run 'TestFailStaleTasks_ReclaimsAbandonedLocalDirectoryWaiters'`
- [ ] Manual: park tasks behind an in_place `local_directory` holder, force-stop the holder (idle watchdog / cancel), confirm waiters resume or fail+retry instead of freezing the agent queue

Made with [Cursor](https://cursor.com)


multica-eve (changes requested):
The product direction is right: releasing abandoned `local_directory` waiters prevents the stale tasks from consuming the agent's capacity indefinitely. However, there are two correctness gaps that need to be addressed before this can merge.

1. **[must-fix] Preserve the runtime reconnect grace for waiting tasks.**

   In `server/pkg/db/queries/agent.sql`, the new `waiting_local_directory` sweeper branch only checks whether `prepare_lease_expires_at` is missing or expired. That means a transiently disconnected runtime can have its waiting task failed after the 45-second prepare lease, even though the default runtime reconnect grace is three hours. The retry can then expire before the runtime's grace window ends. Please gate this branch with the same runtime-liveness/reconnect-grace condition used for dispatched/running tasks, and add regression coverage for both a recently-offline runtime (must not fail) and one beyond the grace period (must fail).

2. **[must-fix] Wire the new failure reason through every user-facing and analytics consumer.**

   `waiting_local_directory_abandoned` is not handled in `packages/views/agents/components/tabs/task-failure.ts`, `apps/mobile/components/issue/run-row.tsx`, `apps/mobile/lib/failure-reason-label.ts`, or `packages/core/dashboard/failure-class.ts`, and it is missing from the retry/classifier matrix in `server/internal/service/task_complete_race_test.go`. As written, users will see the raw enum or a bare “Failed”, and dashboards will classify these failures as “other”. Please add the web/mobile labels, runtime failure class, and classifier/retry tests together with the new canonical reason.

Before requesting another review, please also run `gofmt` on `server/internal/service/task.go`, clean the whitespace reported by `git diff --check`, and use the repository's `internal/testutil` fixtures in the new database test.

multica-eve (changes requested):
Thanks for the follow-up. The two previous correctness blockers are addressed: the waiter sweeper now preserves the runtime reconnect grace, and `waiting_local_directory_abandoned` is wired through the retry/classification and user-facing consumers. The current head still needs the following changes before it can merge.

1. **[must-fix] Resolve the conflicts against the current `main` and request another review on the resulting head.**

   GitHub currently reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. The conflicts are in `packages/views/agents/components/tabs/task-failure.ts`, its test, `packages/views/issues/components/execution-log-section.tsx`, and the four `packages/views/locales/*/agents.json` files. Those are exactly the files that carry this round's failure-reason integration, so the final merge result is not reviewable yet. Please keep the current `main` i18n/diagnostics additions (including `details_diagnostics`, the presence/model-discovery copy, and the latest tooltip behavior) and minimally reapply the `waiting_local_directory_abandoned` key, copy, and tests.

2. **[must-fix] Convert the new sweeper database test to the repository fixture API.**

   `server/cmd/server/runtime_sweeper_test.go:205-236` still uses the legacy fixture helper, manual cleanup, and an open-coded `INSERT ... RETURNING`. The repository requires new DB-backed tests to construct rows through `server/internal/testutil`; the fixture-based test added elsewhere does not remove this violation.

3. **[must-fix] Make the formatting checks clean.**

   `gofmt -d` still reports the extra blank line at `server/cmd/server/runtime_sweeper_test.go:278` and the trailing blank line at the end of `server/internal/daemon/local_directory_test.go`. `git diff --check` fails on the latter.

Nit: `apps/mobile/data/agent-task-schema.test.ts` is a pure Node test and should start with `// @vitest-environment node` per the repository test rules.

Validation on this head passed for the daemon regression, reconnect-grace and sweeper tests, failure taxonomy/classifiers, prompt `ByteIdentical`, targeted frontend tests, locale/chat tests, Web/Desktop typecheck, Mobile typecheck, and the relevant linters (0 errors). GitHub has not produced CI checks for this conflicting head.

PR #7435, opened 2026-08-22
fix(runtimes): hide private runtimes from non-owners

### LINKED ISSUE #7433: [Bug]: Private runtimes remain visible to non-owner workspace members
## Deployment type
Not specified

## Area
Workspace permissions / Runtimes

## What happened?
Private runtimes are visible to workspace members who are not the runtime owner.

When a user creates a runtime marked as private, other members of the same workspace can still see it in the runtime list and may be able to open or interact with it.

Expected behavior: private runtimes should only be visible and accessible to their owner, unless the owner explicitly shares them or an administrator is intended to have access.

## Steps to reproduce
1. Create a workspace with at least two members: User A and User B.
2. Sign in as User A.
3. Create a new runtime and set its visibility to private.
4. Sign in as User B, who is a workspace member but not the runtime owner.
5. Open the workspace runtime list.
6. Observe that User B can see User A's private runtime.

## Additional context
This may expose private runtime metadata or access to workspace members who are not authorized by the runtime owner.

No screenshots, environment details, or logs were provided.

m2keloop:
## What does this PR do?

Prevents workspace members from receiving private agent runtimes owned by other users in the standard runtime list.

## Related Issue

Fixes #7433

## Type of Change

## Changes Made

- Adds a workspace-scoped runtime query that returns only public runtimes or those owned by the requesting user.
- Uses that visibility-aware query for the normal runtime-list endpoint.
- Adds regression coverage for a non-owner listing the workspace runtimes.

## How to Test

1. Create one private runtime and one public runtime for a workspace member.
2. List runtimes as another workspace member.
3. Verify the private runtime is absent and the public runtime remains visible.

Focused regression test passed:

```
DATABASE_URL=postgres://… go test ./internal/handler -run '^TestListAgentRuntimes_HidesOtherMembersPrivateRuntimes$' -count=1 -v
```

## Checklist

- [x] I traced the existing list handler and access helper, then implemented the matching SQL visibility filter.
- [x] I have run tests locally and they pass.
- [x] I have added regression coverage.
- [x] I have considered risks: owner and public runtime behavior remain unchanged.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Traced the runtime list path, added a visibility-scoped sqlc query, generated its binding, and verified the behavior with an isolated PostgreSQL regression test.

### DIFF
--- a/packages/views/agents/components/agent-detail-inspector.labels.test.tsx
+++ b/packages/views/agents/components/agent-detail-inspector.labels.test.tsx
@@ -1,17 +1,12 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { cleanup, screen } from "@testing-library/react";
-import type { Agent, AgentRuntime } from "@multica/core/types";
+import type { Agent } from "@multica/core/types";
 import { renderWithI18n } from "../../test/i18n";
 import { AgentDetailInspector } from "./agent-detail-inspector";
 
-const mockUseQuery = vi.hoisted(() => vi.fn());
-
 vi.mock("@tanstack/react-query", async (importOriginal) => ({
   ...(await importOriginal<typeof import("@tanstack/react-query")>()),
-  useQuery: (options: unknown) => {
-    mockUseQuery(options);
-    return { data: undefined, isSuccess: false };
-  },
+  useQuery: () => ({ data: undefined, isSuccess: false }),
 }));
 
 // The picker is the only way agent labels ever reached the agent settings
@@ -52,7 +47,6 @@ const agent = {
 describe("AgentDetailInspector labels", () => {
   afterEach(() => {
     cleanup();
-    mockUseQuery.mockClear();
   });
 
   // Agent labels were removed from the product (MUL-5600). Label Settings no
@@ -76,41 +70,5 @@ describe("AgentDetailInspector labels", () => {
 
     expect(screen.queryByTestId("resource-label-picker")).toBeNull();
     expect(screen.queryByText("Labels")).toBeNull();
-  });
-
-  it("does not discover models for another member's private runtime", () => {
-    const privateRuntime = {
-      id: "runtime-1",
-      workspace_id: "workspace-1",
-      daemon_id: "daemon-1",
-      name: "Private runtime",
-      runtime_mode: "local",
-      provider: "codex",
-      launch_header: "",
-      status: "online",
-      device_info: "Mac",
-      metadata: {},
-      owner_id: "user-2",
-      visibility: "private",
-      last_seen_at: null,
-      created_at: "2026-01-01T00:00:00Z",
-      updated_at: "2026-01-01T00:00:00Z",
-    } satisfies AgentRuntime;
-
-    renderWithI18n(
-      <AgentDetailInspector
-        agent={agent}
-        runtime={privateRuntime}
-        runtimes={[privateRuntime]}
-        members={[]}
-        currentUserId="admin-1"
-        canEdit
-        onUpdate={vi.fn(async () => {})}
-      />,
-    );
-
-    expect(mockUseQuery).toHaveBeenCalledWith(
-      expect.objectContaining({ enabled: false }),
-    );
   });
 });
--- a/packages/views/agents/components/agent-detail-inspector.labels.test.tsx
+++ b/packages/views/agents/components/agent-detail-inspector.labels.test.tsx
@@ -1,12 +1,17 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { cleanup, screen } from "@testing-library/react";
-import type { Agent } from "@multica/core/types";
+import type { Agent, AgentRuntime } from "@multica/core/types";
 import { renderWithI18n } from "../../test/i18n";
 import { AgentDetailInspector } from "./agent-detail-inspector";
 
+const mockUseQuery = vi.hoisted(() => vi.fn());
+
 vi.mock("@tanstack/react-query", async (importOriginal) => ({
   ...(await importOriginal<typeof import("@tanstack/react-query")>()),
-  useQuery: () => ({ data: undefined, isSuccess: false }),
+  useQuery: (options: unknown) => {
+    mockUseQuery(options);
+    return { data: undefined, isSuccess: false };
+  },
 }));
 
 // The picker is the only way agent labels ever reached the agent settings
@@ -45,7 +50,10 @@ const agent = {
 } as Agent;
 
 describe("AgentDetailInspector labels", () => {
-  afterEach(cleanup);
+  afterEach(() => {
+    cleanup();
+    mockUseQuery.mockClear();
+  });
 
   // Agent labels were removed from the product (MUL-5600). Label Settings no
   // longer manages an agent catalog, so an attach-only picker here would be a
@@ -68,5 +76,41 @@ describe("AgentDetailInspector labels", () => {
 
     expect(screen.queryByTestId("resource-label-picker")).toBeNull();
     expect(screen.queryByText("Labels")).toBeNull();
+  });
+
+  it("does not discover models for another member's private runtime", () => {
+    const privateRuntime = {
+      id: "runtime-1",
+      workspace_id: "workspace-1",
+      daemon_id: "daemon-1",
+      name: "Private runtime",
+      runtime_mode: "local",
+      provider: "codex",
+      launch_header: "",
+      status: "online",
+      device_info: "Mac",
+      metadata: {},
+      owner_id: "user-2",
+      visibility: "private",
+      last_seen_at: null,
+      created_at: "2026-01-01T00:00:00Z",
+      updated_at: "2026-01-01T00:00:00Z",
+    } satisfies AgentRuntime;
+
+    renderWithI18n(
+      <AgentDetailInspector
+        agent={agent}
+        runtime={privateRuntime}
+        runtimes={[privateRuntime]}
+        members={[]}
+        currentUserId="admin-1"
+        canEdit
+        onUpdate={vi.fn(async () => {})}
+      />,
+    );
+
+    expect(mockUseQuery).toHaveBeenCalledWith(
+      expect.objectContaining({ enabled: false }),
+    );
   });
 });
--- a/packages/views/agents/components/agent-detail-inspector.runtime-access.test.tsx
+++ b/packages/views/agents/components/agent-detail-inspector.runtime-access.test.tsx
@@ -1,0 +1,121 @@
+// @vitest-environment jsdom
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { act, cleanup, waitFor } from "@testing-library/react";
+import type {
+  Agent,
+  AgentRuntime,
+  RuntimeModelListRequest,
+} from "@multica/core/types";
+import { renderWithI18n } from "../../test/i18n";
+
+const mockInitiateListModels = vi.hoisted(() => vi.fn());
+const mockGetListModelsResult = vi.hoisted(() => vi.fn());
+
+vi.mock("@multica/core/api", () => ({
+  api: {
+    initiateListModels: (...args: unknown[]) =>
+      mockInitiateListModels(...args),
+    getListModelsResult: (...args: unknown[]) =>
+      mockGetListModelsResult(...args),
+  },
+}));
+
+vi.mock("../../common/avatar-upload-control", () => ({
+  AvatarUploadControl: () => <div data-testid="avatar-upload" />,
+}));
+
+vi.mock("./inspector/runtime-picker", () => ({
+  RuntimePicker: () => <div data-testid="runtime-picker" />,
+}));
+
+import { AgentDetailInspector } from "./agent-detail-inspector";
+
+const agent = {
+  id: "agent-1",
+  workspace_id: "workspace-1",
+  name: "Lambda",
+  description: "Test agent",
+  runtime_id: "runtime-1",
+} as Agent;
+
+const privateRuntime = {
+  id: "runtime-1",
+  workspace_id: "workspace-1",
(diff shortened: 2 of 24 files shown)


Bohan-J (changes requested):
Thanks for picking this up — the product direction is right, and the list-level query itself is correct and stays workspace-scoped. Two things need to be resolved before this can merge.

## Must-fix 1 — private runtimes are still readable and triggerable by ID

The new filter only applies to the list endpoint (`server/pkg/db/queries/runtime.sql:6`). Every per-ID runtime endpoint still authorizes on "caller is a workspace member" alone, so a non-owner holding a private runtime's ID keeps full access:

- `server/internal/handler/runtime.go:110-125` — `GET /api/runtimes/{id}/usage`
- `server/internal/handler/runtime.go:174-188` — `GET /api/runtimes/{id}/activity`
- `server/internal/handler/runtime.go:242-256` — `GET /api/runtimes/{id}/usage/by-agent` (agent IDs, providers, models, task counts)
- `server/internal/handler/runtime.go:324-338` — `GET /api/runtimes/{id}/usage/by-hour`
- `server/internal/handler/runtime_models.go:314-328` — `POST /api/runtimes/{id}/models`, which actually dispatches work to the owner's machine
- `server/internal/handler/runtime_local_skills.go:553-578` (`requireRuntimeCapabilityReadAccess`, used by `InitiateListLocalSkills` at `:605`) — same member-only gate, and it enumerates the skills installed on the owner's machine

Unguessable UUIDs don't mitigate this. Those IDs were handed out by the list endpoint before this fix, and `runtime_id` is still returned for every agent bound to a private runtime (`server/internal/handler/agent.go:45,182`; `ListAgents` at `:922-936` is workspace-wide). A member with an old link — or who just reads the agent list — bypasses the new list filter entirely, so the "visible and accessible only to the owner" expectation in #7433 isn't delivered yet.

Suggested shape: extract one fail-closed read/access gate expressing the same rule as `canUseRuntimeForAgent` (`server/internal/handler/runtime.go:706`) and route every user-facing per-ID runtime endpoint through it — `requireRuntimeCapabilityReadAccess` is the natural home, since local-skills already funnels through it. Return 404 rather than 403 so the endpoints stay enumeration-safe. Please add negative tests showing a plain member with a known private runtime ID cannot read usage/activity or start model discovery, alongside positive tests that the owner and public-runtime paths don't regress.

## Must-fix 2 — admin visibility now contradicts the existing edit/delete contract

`canEditRuntime` (`server/internal/handler/runtime.go:666-671`) deliberately lets workspace owners/admins rename and delete any runtime, and the comment on `canSetRuntimeVisibility` (`:716-724`) spells out the split: admins keep the organisational actions, only the owner consents to lending the machine out. The new query is `owner_id = $2 OR visibility = 'public'` with no role branch, so admins lose the list entry point for private runtimes while the API still lets them govern those runtimes by ID — governance becomes impossible in the UI but remains allowed over the API.

Please make one rule and apply it consistently:

- admins keep governance visibility → include the role in the query (still without granting `canUseRuntimeForAgent`), ideally flagged so the UI doesn't offer them "use this runtime"; or
- admins are fully blind to other members' private runtimes → narrow `canEditRuntime` and the delete path accordingly, and update those tests.

The new test at `server/internal/handler/runtime_visibility_test.go:234-265` covers only the plain-member cell. Whichever rule you choose, please cover: owner sees own private, member sees another member's public, member does not see another member's private, and the admin case.

## Otherwise

The SQL keeps the `workspace_id` scope, the generated binding matches the query, and the new test follows the repo's `testutil.Call` convention. No migration or wire-format change. Leaving `ListAgentRuntimes` in place is correct — it still has a legitimate internal caller at `server/internal/handler/project_resource.go:175` (a daemon-capability check, not a user-facing read).

Green CI doesn't cover the gaps above: they're unauthorized-access paths that currently have no test.

m2keloop:
Implemented option A in a887ecd: workspace owners/admins retain governance visibility for private runtimes, while private runtime data and machine-triggering endpoints are owner-only and return 404 for non-owners.

The read/use gate now covers usage/activity, model and local-skill discovery (including polling), and CLI update requests/status. Added regressions for plain members and admins with known private runtime IDs, plus owner/public positive paths.

Verified with `go test ./internal/handler -count=1` against PostgreSQL and `go vet ./internal/handler`.

multica-eve (dismissed):
Thanks for addressing the original two blockers. The new backend gate closes known-ID access, and preserving admin visibility for governance in the list is coherent. One PR-introduced client/server mismatch still needs to be fixed before merge.

## Must-fix — the admin UI still exposes private-runtime operations that the backend now rejects

Commit `a887ecd` changes the contract for another member's private runtime: `InitiateUpdate` and `GetUpdate` now pass through `requireRuntimeReadAccess`, and `runtime_update_authorization_test.go` changes the workspace-admin expectation from `200` to `404`.

The frontend was not updated for that new contract:

- `packages/views/runtimes/components/machine-cli-section.tsx:5-25` still treats admins as able to manage any runtime and may select a teammate-owned private runtime for Update.
- `packages/views/runtimes/pages/runtime-detail-page.tsx:162-168,264-268` still passes `canManageAnyRuntime={isAdmin}`.
- `packages/views/runtimes/components/machine-cli-section.test.tsx:106-131` explicitly expects an admin to update through a teammate-owned runtime.
- `packages/views/runtimes/components/runtime-detail.tsx:196` still renders `UsageSection` unconditionally, while the new backend policy rejects usage access to that private runtime. The usage query then degrades the authorization failure into a misleading empty-data state.

This makes the visible Update action fail with `404`. On a machine containing both public and private runtimes, the selector can also prefer an online private runtime and block an otherwise valid public update path.

Please align the client with the new owner-or-public read/trigger policy: select update channels only from runtimes usable by the current user, hide Usage and Update for an admin viewing another member's private runtime, retain the intended governance actions such as rename/delete, and add coverage for both admin + private (no read/trigger controls) and admin + public (controls available).

m2keloop:
@multica-eve Addressed in 09ba04fc1, rebased onto current `main` and squashed to one commit.

- Update-channel selection now permits only the runtime owner or public runtimes, so an admin cannot trigger an update through a teammate's private runtime.
- Usage is hidden for an admin viewing another member's private runtime, while governance actions such as rename/delete remain available.
- Added regressions for private teammate runtimes (no update/usage) and public teammate runtimes (controls remain available).

`git diff --check` passes. I added the targeted Vitest coverage, but could not run it locally: the host has Node 20 while this repo requires Node 22+, and the temporary Node 22 container lacked its platform-native test dependency. CI runs the project Node 22 environment.

multica-eve (dismissed):
Thanks for the follow-up. The latest update correctly hides detail-page usage for an admin viewing another member's private runtime and stops selecting that private runtime as an update channel. Two reachable client/server permission mismatches remain.

## Must-fix 1 — a plain member is now offered Update for another member's public runtime, but the API rejects it

`packages/views/runtimes/components/machine-cli-section.tsx:17-25` selects an update channel when the viewer owns the runtime **or the runtime is public**. The new test at `machine-cli-section.test.tsx:134-160` pins that behavior with `currentUserId="user-1"` and runtimes owned by `user-2`.

However, `server/internal/handler/runtime_update.go:214-220` applies two gates: the owner-or-public read gate and then `canEditRuntime`. The second gate permits only the runtime owner or a workspace owner/admin. A plain member therefore sees Update for a teammate's public runtime and gets a guaranteed `403` when clicking it.

Please make the client match the full update authorization rule: the eligible channel is an owned runtime, or a public runtime when the viewer is an owner/admin. A public runtime lets an ordinary member run agents; it does not grant machine-management permission. Please cover owner + private, admin + public, admin + private, and member + public explicitly.

## Must-fix 2 — governance-only admin views still issue private-runtime read requests that the new backend denies

The detail-page `UsageSection` fix is correct, but the same read gate has more consumers:

- `packages/views/runtimes/components/runtime-list.tsx:378-384,805-813` mounts `CostCell` for every non-pending row. Because admins intentionally receive private runtimes in the governance list, this sends a usage request that now returns `404`, then renders the failure as `—`.
- When an admin opens an agent bound to another member's online private runtime, `packages/views/agents/components/agent-detail-inspector.tsx:117-125` still starts model discovery. `packages/views/agents/components/tabs/skills-tab.tsx:63-67` and `tabs/mcp-config-tab.tsx:77-81` likewise request runtime capabilities. All of those endpoints now reject that admin/private combination.

Please gate these consumers with `isRuntimeUsableForUser` (or the equivalent owner/public rule) so a governance-only view does not initiate Usage, model, or capability requests, while preserving rename/delete and an explicit read-only presentation. Add coverage for the admin + private path.

The backend authorization tests and current CI are green, but the new member + public frontend test currently codifies behavior that contradicts the update endpoint's authorization contract.

m2keloop:
@multica-eve Addressed both requested changes in 4db1c03ac (the PR remains a single commit).

- CLI Update now selects an owned runtime, or a public runtime only when the viewer is a workspace owner/admin. A regular member sees no Update action for a teammate’s public runtime.
- Governance-only views of another member’s private runtime no longer request Usage, model discovery, Skills, or MCP capabilities; Skills/MCP show the existing permission notice instead. Rename/delete governance remains unchanged.
- Added regressions for the owner/admin/member Update matrix and the private-runtime Usage/model/Skills/MCP paths.

Validation: `pnpm --filter @multica/views typecheck` passed; targeted ESLint has no errors (two pre-existing hook-dependency warnings in `mcp-config-tab.tsx`). After updating this host to Node `v22.23.2` and restoring the locked dependencies, the focused Vitest suite passed: 5 files, 45 tests.

multica-eve (dismissed):
Thanks for addressing the previous two blockers. The update-channel matrix, runtime-list cost gate, and Skills/MCP capability gates now match the backend policy. One model-discovery path is still active in the Agent General panel.

## Must-fix — the inspector disables only its parent model query; mounted child fields still query the private runtime

`packages/views/agents/components/agent-detail-inspector.tsx:128-130` correctly passes `null` to its own `runtimeModelsOptions` call when `canReadRuntime` is false. However, the same component still passes `runtimeOnline={!!isOnline}` at `:273-303` to `ModelPicker`, `ThinkingSettingField`, and `ServiceTierSettingField`.

Those mounted children independently call `runtimeModelsOptions` with the real runtime ID:

- `packages/views/agents/components/inspector/model-picker.tsx:51-53`
- `packages/views/agents/components/inspector/thinking-prop-row.tsx:51-53,92-94`
- `packages/views/agents/components/inspector/service-tier-setting-field.tsx:41-43`

Because the disabled parent query uses the aggregate key while the child queries use the runtime-specific key, disabling the parent observer does not suppress the children. An admin opening an agent bound to another member's online private runtime therefore still starts model discovery and receives the backend's `404`.

The new regression test does not exercise this path: `agent-detail-inspector.labels.test.tsx:28-42` mocks all three child components, and the assertion at `:112-114` proves only that at least one query has `enabled: false`; it does not prove that no enabled runtime-specific query exists.

Please propagate `canReadRuntime` to all three child fields (for example, include it in the `runtimeOnline` condition) and add coverage with the real children or an API/query spy that proves the full admin + private inspector starts no model-discovery request.

Current CI is green, but the test isolation above is why it does not catch this remaining request.

m2keloop:
@multica-eve Addressed the remaining Agent General model-discovery path in [16fe639](https://github.com/multica-ai/multica/pull/7435/commits/16fe639d4420af99017cdb1e7e8f62811df198da).

- Derived one `canDiscoverRuntimeModels` gate from runtime online state and `canReadRuntime`.
- Applied it to the parent catalog query and the real `ModelPicker`, `ThinkingSettingField`, and `ServiceTierSettingField` consumers.
- Replaced the mocked `useQuery` assertion with an integration regression that mounts the real child fields and spies on `initiateListModels`.
- Added a positive owner case so private-runtime owners continue to discover models.

Validation with Node v22.23.2:

- 9 focused Vitest files: 73/73 tests passed, including the existing Skills, MCP, CLI, runtime-detail, and runtime-list permission coverage.
- `@multica/views` TypeScript check passed.
- ESLint passed for all changed files.
- `git diff --check` passed.
- GitHub CI run 13747 passed.

Bohan-J (approved):
Thanks for sticking with this one, @m2keloop — four rounds from a list-level filter to a coherent, fail-closed access rule across the whole runtime surface, and you took every round of feedback seriously. The final shape is exactly right: one `requireRuntimeReadAccess` gate on the server, `isRuntimeUsableForUser` as the single client-side rule, governance (rename/delete) deliberately kept separate from read/machine access, and an enumeration-safe 404 instead of a 403 oracle.

Special credit for the last round's test: `agent-detail-inspector.runtime-access.test.tsx` keeps the real child components and spies at the API boundary. I verified it actually bites — reverting `runtimeOnline={canDiscoverRuntimeModels}` back to `!!isOnline` makes it fail, so it pins the real call chain rather than restating the implementation.

I pushed one small follow-up commit to your branch (`refactor(runtimes): reuse isRuntimeUsableForUser on runtime detail`): the detail page still re-derived the owner-or-public rule inline, which disagreed with the shared helper for an ownerless runtime — it would render a Usage section that can only 404, since the server's gate reuses `canUseRuntimeForAgent` and is fail-closed without an owner. Now it calls the helper like every other surface, with a regression test for that case.

Verified locally before approving: `go build ./...` and `go vet` clean; backend permission tests against a live database, including the new 8-endpoint × (plain member, admin) 404 matrix plus the owner and public positives; the full `packages/views` runtime and agent suites (364 tests) plus typecheck and lint.

Two things worth separate issues rather than more scope here:

1. Flipping a runtime public → private doesn't unbind teammates' agents, and the agent page resolves its runtime from the list — so those agents now render as "no runtime" while staying bound and still running on the owner's machine. A pre-existing gap this change makes visible.
2. `requireRuntimeReadAccess` inherits `canUseRuntimeForAgent`'s fail-closed handling of `owner_id IS NULL`. Harmless today (migration 032 backfilled owners, and such runtimes can't run tasks anyway per MUL-3292), but once the `mdt_` daemon-token registration flow goes live, a headless daemon would have no reachable remote-update path.

Approving and merging. Thanks again for the thorough work.

PR #7441, opened 2026-08-23
MUL-6576 feat(tasktoken): issue task identity tokens so agents act as the person who asked

wxkingstar:
## The problem

Self-hosted teams want their agents to work with the internal systems they
already run — ERP, admin backends, wikis. Those systems already have a
permission model, built over years, and an audit trail that names real people.

Today the only way to connect one is to give the agent a service account. The
moment you do, two things break:

- **Permissions collapse to one level.** Every user's agent acts with the same
  fixed identity. A junior engineer's agent can reach whatever the service
  account can. The existing model stops applying at exactly the point where
  automation starts touching things.
- **Audit goes anonymous.** The internal system logs "the bot did it". Who
  actually asked is lost, which is usually the one thing an audit needs.

So the practical answer becomes "don't connect agents to the systems that
matter", and self-hosted deployments lose much of the point of running agents
next to their own infrastructure.

## What this adds

At claim time the server signs a short-lived JWT for the run's **authorizing
human** and the daemon injects it into the agent process. The agent presents
that token to the internal system, which sees a request from that person.

Nothing about the existing permission model changes. The internal system keeps
doing exactly what it did before — check who this is, apply their rights, log
their name. It simply now receives a real person instead of a shared bot.

The identity is the **human who triggered the run**, or the **immutable
creator of a schedule** when it fires. The gate reads `originator_user_id`,
never the audit-only `accountable_user_id`. Assignment and backlog-promotion
handlers preserve the authenticated agent's source task and human, including
batch and squad paths; they cannot borrow the target issue creator's identity.
Missing, terminal, foreign-workspace or mismatched-agent source tasks carry no
authorizing human.

The signer follows every `delegation` / `comment_source` edge within the
workspace and checks that the human never changes. A `direct_human` root may
sign. A `trigger_owner` root may sign only when its real dispatch record and
schedule trigger prove that the same human is the immutable member creator.
The automation owner, rule publisher and later trigger editors cannot replace
that creator. Both `run_only` and `create_issue` schedules are supported,
including descendants claimed after the root run completes.

Webhook roots without an authenticated human, `rule_owner`, `owner_fallback`,
`backfill`, `unattributed`, and broken or overlong lineage receive no generated
token. The person must also still be a workspace member when the run is
claimed: authorization at enqueue does not permit signing after offboarding.

Claims are templated, so a deployment maps identity onto whatever its systems
already understand. The documented default subject is the full corporate email;
a deployment whose systems key on the username can map the local part instead,
paired with the template's `allowed_domains` allowlist so identities outside
the corporate domain are refused rather than colliding.

## How it relates to `Initiator*` (MUL-2645)

The existing comment is explicit, and this PR does not contradict it:

> The agent's effective Multica credentials stay owner-scoped — this is an
> attested identity, not a credential.

That still holds. The agent's access to **Multica's own API** is untouched and
remains owner-scoped. What is signed here is a credential for a **different
trust domain**: systems outside Multica that opt into accepting it.

Named precisely, this is **delegated user impersonation** in the RFC 8693
token-exchange sense: `sub` is the delegating human, and the recommended
`act_sub` / `act_name` / `task_id` claims name the agent actually executing,
so the receiving system records "agent X on behalf of Alice". It is *not*
workload attestation à la GitHub Actions OIDC — those tokens deliberately
never name a user — and it should be reviewed with the trust obligations
delegation carries. Multica never gains any authority over the target system;
it only states, verifiably, who is asking and who acts for them.

## Shape

Off unless configured. Two env vars — a private key and a catalog of templates:

```json
[{"id":"erp","label":"ERP","env":"BOT_TOKEN_ERP","ttl":"8h",
  "allowed_domains":["corp.com"],
  "claims":{"iss":"multica","scope":"erp","sub":"{{identity.email}}",
            "act_sub":"{{agent.id}}","act_name":"{{agent.name}}","task_id":"{{task.id}}"}}]
```

An agent enables the subset it needs in **Settings → Agents → Identity
Tokens**. The catalog lives only in server config, so the UI can pick from what
an operator allowed but can never define what may be signed.

- Unset key/catalog → no behaviour change anywhere, and the tab is hidden
- Claim response gains one `omitempty` field, issued only to a daemon that
  advertises `task-identity-tokens-v1`. Older daemons ignore the field, so
  signing for one would leave an audit row claiming a credential was minted
  for a person when nothing received it — the capability is checked before
  signing, and the upgrade order (daemons first) is documented
- Migration only adds a column with a default — no FK, no index, and rollback
  to an older server is safe (every generated query uses explicit columns)
- Authorization and audit reuse the `/env` endpoints' model: agent actors are
  rejected, caller must be a workspace owner/admin or the agent's owner, and
  every write lands in `activity_log`
- Issuance itself is audited too: each batch of minted tokens writes an
  `agent_task_tokens_issued` activity row (jti + expiry per token, tied to the
  run's issue), and fails closed — if the audit row cannot be written, the
  tokens are withheld. Tokens are signed only **after** `FinalizeTaskClaim`
  succeeds, in both claim paths: every earlier step can still requeue or
  cancel the task, and an audit row must never outlive the claim it describes
  (nor leave a live JWT nobody received)
- Codex runs need one more step: Codex's `shell_environment_policy` drops any
  variable whose name contains `TOKEN` unless it is explicitly authorized, so
  the daemon authorizes task-token names exactly like the agent's own
  `custom_env` credentials — otherwise a `BOT_TOKEN_*` the server had already
  audited would read as empty from every shell tool
- The daemon's reserved env list now lives in `pkg/protocol` and is shared by
  the daemon's injection blocklist and the catalog's boot-time check, so a
  template can never name a variable the daemon would silently drop (the two
  hand-copied lists had already drifted on `CURSOR_MCP_AUTH_SOURCE`)
- Optional `MULTICA_TASK_TOKEN_MANIFEST_ENV` injects a JSON description of the
  systems **actually issued**, so a tool's view of what it can reach is derived
  from the credentials it holds rather than configured separately and drifting
- The catalog is all-or-nothing at boot: unknown variables, reserved claims and
  env names, TTL over the cap, duplicate ids or env names, a manifest variable
  colliding with a template's, and an `algorithm` the configured key cannot
  produce (`RS256` with an EC key, `ES256` with a P-384 key) all fail the
  start rather than failing on every task later

## Verifying: `GET /.well-known/jwks.json`

A signed token is only useful if the receiving system can check it, and making
each operator hand-copy a public key into every internal system turns rotation
into a coordinated outage. So the public half is served as a JWK Set that any
mainstream JWT library consumes directly.

- **Unauthenticated by design.** The systems that verify these tokens hold no
  Multica credentials — that is the whole point — and the response is public
  key material. Nothing private can reach it: the `JWK` struct declares only
  public components, and a test asserts the serialised form carries no `d` /
  `p` / `q` / `dp` / `dq` / `qi`.
- **404 unless configured**, so "this deployment does not issue task tokens" is
  distinguishable from "configured, but no keys".
- **Standard library only** — no new dependency for JWK encoding.
- One private key can carry several key ids across templates, so the set
  publishes one entry per distinct `(kid, alg)` pair, including a kid-less
  entry when a template configures none (its tokens carry no `kid` either).
- EC coordinates are left-padded to the curve size; `big.Int.Bytes()` trims
  leading zeros and a verifier reading `x`/`y` into a fixed-width buffer
  rejects a short one.

One limitation worth stating plainly: a deployment configures **one** signing
key, so rotation means a restart and there is no window where old and new keys
are both published. The docs say so and give the sequence that avoids breaking
in-flight runs. Publishing overlapping keys would be a natural follow-up; it
seemed better to leave it out than to guess at the shape.

## Documentation

Two documents, because the feature has two distinct audiences:

- **`TASK_IDENTITY_TOKENS.md`** — for the operator deciding whether to enable
  it. The problem service accounts create, a diagram of the whole path from
  "Alice assigns an issue" to "the ERP writes an audit row naming Alice", the
  catalog format, which attribution sources are signed, and the boundaries
  worth knowing first (the token sits in the process environment for the whole
  run, so its lifetime is its exposure window; key ids are public).
- **`TASK_IDENTITY_TOKENS_AI.md`** — for the receiving end, written to be
  handed to an agent working inside the legacy codebase, in the style of the
  existing `SELF_HOSTING_AI.md`. Mostly a catalogue of the ways this goes
  wrong: trusting the token's own `alg` (the `none` / `HS256` forgery, which
  the public JWKS makes trivial if the verifier allows a symmetric algorithm),
  auto-creating a user for an unknown `sub`, falling back to a service account
  on failure, or logging the actor as the bot — each of which quietly restores
  the problem this feature removes.

`SELF_HOSTING_ADVANCED.md` gains the env vars in the configuration reference,
and `SELF_HOSTING.md` points at the feature from the end of setup.

## Testing

- `tasktoken`: catalog validation (16 rejection cases), interpolation, real
  signature verification, unique `jti`
- JWKS: one entry per key id, deduplication, the kid-less case, EC coordinate
  padding against the curve size, RSA keys, no private material in the
  serialised set, and an end-to-end test that verifies a token the issuer
  actually signed using only the published set
- Issuing gate: all attribution sources, NULL authorization, current membership,
  unproven trigger labels, and audit-only fallback identities
- Assignment: eight real assign/promote paths (single/batch, agent/squad) are
  exercised through the final daemon claim response; the JWT subject must be
  the acting human even when someone else created the issue
- Schedule identity: both execution modes, two delegation hops, completed roots,
  creator vs. editor/publisher, missing or wrong creator, agent-created triggers,
  webhook sources/kinds, foreign triggers and offboarded creators
- Chain integrity: identity substitution at an intermediate hop is refused
- Receiving-system reference: eight executable tests extract the actual Python
  snippet and cover unknown-kid throttling, failed-refresh cooldowns, cold starts,
  rotation, bounded stale keys, malformed/HTTP-error responses and concurrent
  refreshes. CI runs these even for documentation-only changes
- The guide permits multiple requests with the same access JWT and uses business
  idempotency keys for duplicate operations; `jti` remains an audit/revocation id
- Mixed-version claim: a daemon advertising the capability gets a token and one
  audit row; a daemon on `skill-bundles-v1` only, and one advertising nothing,
  get neither — the audit trail must never record an issuance nothing received
- Handler: permission matrix (agent actor 403, plain member 403, owner OK),
  rejection of ids outside the catalog, audit rows, and the JWKS endpoint's
  404 / 200 / cache-header behaviour
- Daemon: injection, blocklist enforcement, that `custom_env` still wins, the
  Codex `include_only` policy carrying task-token names (and not authorizing a
  token-shaped name merely for being present), and the shared reserved list
  agreeing with `execenv.CursorMcpAuthSourceEnv`
- Frontend: schema defaults + malformed-response fallback, component tests,
  i18n parity across all four locales, a lost-write regression — two toggles
  in one synchronous block must produce a second payload computed after the
  first landed, not from the pre-toggle snapshot — a save response that fails
  schema validation being refetched rather than cached as an empty catalog, a
  failed catalog request rendering as an error rather than "nothing
  configured", and the Identity Tokens tab gate at pane level (shown with a
  catalog, hidden without one, hidden from non-editors)


Bohan-J (changes requested):
Thanks for this — the problem is real, the mechanism is a sensible shape for it, and the engineering is above the bar (catalog validated at boot, the table-driven test across all 8 attribution sources, the JWKS test asserting no private components ever serialise).

Two things below are blocking, and the reason they're blocking rather than follow-ups is at the bottom: this is a protocol, and protocols freeze the moment someone integrates against them.

To set expectations: we aren't planning to pick these changes up ourselves right now. If you'd like to carry it forward, we're happy to review.

## Blocking

### 1. There's no way to record *which agent* acted, so the audit story is only half-solved

`knownVariables` is a closed set containing only `identity.*` and `workspace.*` — there is no `agent.*` or `task.*`. Templates are per-system and shared across agents, so a static string in the claims can't identify the acting agent either.

Combined with this row in `TASK_IDENTITY_TOKENS_AI.md`:

> | Log the actor as "bot" or the agent's name | Audit must name the accountable human. |

…the audit row the receiving system writes is indistinguishable from Alice having logged in and done it herself.

That trades one blind spot for another. A service account answers "was this automation?" with a trivial query and can't answer "who asked?". This design answers "who is accountable?" and gives up the first — and "which ERP changes over the last quarter actually came from agents?" becomes unanswerable. For a feature whose headline is audit, that's the wrong half to lose.

RFC 8693's `act` exists for exactly this: `sub` is the delegating human, `act` is the actual executor, and the receiving system records "agent X on behalf of Alice".

Note `Claims` is `map[string]string`, so a nested `"act": {"sub": ...}` object isn't expressible as-is. Either widen the type, or use flat claims — the flat version seems like the smaller change:

```json
"sub":      "{{identity.email}}",
"act_sub":  "{{agent.id}}",
"act_name": "{{agent.name}}",
"task_id":  "{{task.id}}"
```

Suggested: add `agent.id` / `agent.name` / `task.id` to `knownVariables`, put the `act_*` claims in the recommended template, and change that anti-pattern row to require naming **both** the accountable human and the agent acting for them.

### 2. `email_local` as the documented default `sub` drops the domain

`EmailLocal()` takes the local part, lowercases, strips `+tag`. So `alice@contractor.io` and `alice@corp.com` both sign as `alice`. There's no domain allowlist in the catalog, and the docs don't state anywhere that the stored address must be a verified corporate one.

In any workspace with guests or external collaborators, that's a privilege-escalation path: an outside contributor whose local part collides with an employee's username gets that employee's identity in the receiving system.

The root of the whole trust chain is "the email Multica holds is trustworthy and unique", and right now that's an unstated assumption. Suggested: default the docs to `{{identity.email}}`, add an optional domain allowlist to the template, and state verified corporate email as a hard prerequisite.

## Non-blocking, but I'd like to see them land alongside

### 3. The blast radius inverts, and the docs don't discuss it

Under a service account, an operator deliberately narrows what the agent can reach — very often to read-only. Here the agent receives the requester's full permissions, and the more senior the requester, the more consequential that is. Agents routinely process untrusted input (issue bodies, comments, fetched pages), so prompt injection now executes with a real person's authority rather than a deliberately-narrowed bot's.

"Nothing about your permission model changes" is accurate, and that's precisely the concern — that model was designed for a human who reads what they're clicking, not for an LLM processing attacker-influenced text. `scope` is static per template, so it can't be narrowed per task either.

A docs section recommending a minimal `scope` and read-only first would go a long way. Right now the framing reads as "no configuration needed, it just inherits the person", which is the riskiest way to deploy it.

### 4. The GitHub Actions OIDC analogy doesn't hold

GHA OIDC attests the **workload** — `repo:org/name:ref:...` — and each cloud grants that workload a role. It is explicitly *not* user impersonation; that's the property that makes it safe to hand out.

This feature does user impersonation. RFC 8693 delegation is the accurate precedent. The current comparison makes the design sound more precedented than it is, and it'll be the first thing a security reviewer challenges.

### 5. Issuance never reaches `activity_log`

Signing emits `slog.Info` only. The single `CreateActivity` call is in the enable/disable handler. So "a credential was minted in Alice's name for the ERP" is invisible in-product — including to Alice — and recoverable only from server logs. For a feature that sells auditability, issuance itself belongs on the product's audit surface.

### 6. The threat model should state what a server compromise now buys

The signing key can mint any identity for any configured system. That escalates a Multica compromise from "Multica's data is exposed" to "write to the ERP as the CFO" — cross-trust-domain lateral movement. Not disqualifying (Okta has the same shape, and this is opt-in and self-hosted), but the security boundaries section currently covers smaller things like key ids being public while omitting this one.

Related: the whole document is written in a self-hosted context but never explicitly rules out the hosted product. Given the above, that boundary seems worth stating outright rather than leaving implied.

### 7. There's no refresh path

TTL is fixed at claim time with no renewal. A task that runs longer than its TTL silently loses access partway through, and the failure surfaces to the agent as an ordinary 401 with no indication that expiry was the cause.

That also pressures operators toward longer TTLs precisely because there's no renewal — which works directly against the "prefer hours" guidance and the 24h cap. Either add a refresh channel or state plainly that long-running tasks aren't supported.

## Why 1 and 2 are blocking rather than follow-ups

Once one deployment's internal systems verify these tokens, the claim names, the semantics of `sub`, the JWKS path and the algorithm set are frozen. Changing any of them later breaks every integrated internal system simultaneously, at a moment those teams won't be expecting it.

Before the first integration, changing the recommended `sub` and adding `act_*` costs a default value and a few interpolation variables. After it, it's a breaking change with a migration period. The window is narrow, which is why these seem worth doing now rather than iterating toward.

## One adoption thought, out of scope

The expensive step for every customer is step 6 — writing the verification middleware — and mistakes there fail silently rather than loudly. Two things would likely do more for adoption than any additional capability on the issuing side:

- Document the `issuer` + `jwks_uri` pair explicitly, with config snippets for common gateways (Kong, APISIX, nginx, cloud API gateways). Many of them accept a JWKS URL directly in bearer-JWT validation mode, which makes step 6 zero-code for those deployments.
- Ship small verification middleware packages (Go / Python / Node / PHP) wrapping the five steps. More reliable than a document people copy from, given the failure mode is silent.

## Re: splitting

You offered to split this up — one candidate is `MULTICA_TASK_TOKEN_MANIFEST_ENV`, which has no concrete consumer in-tree yet. Dropping it from v1 would shrink both the review surface and the API you're committing to.

To be clear about the overall read: the direction and the implementation are both sound. The attribution gate refusing `owner_fallback` / `backfill` / `unattributed` is the most important judgment call in the PR and it's the right one, and degrading a missing token to an ordinary 401 rather than a task failure is right too. The gap is that the PR presents itself as a transparent, zero-risk enhancement when it's actually establishing a new trust root and performing user impersonation. Close the audit gap and the `sub` collision, make the threat model and hosted/self-hosted boundary explicit, and this is in good shape.

wxkingstar:
All seven points are addressed in adcb67b (code) + c3d9be8 (docs).

**1 (actor claims)** — Added `agent.id` / `agent.name` / `task.id` to the interpolation set and went with the flat `act_sub` / `act_name` / `task_id` shape you suggested (`Claims` stays `map[string]string`). The recommended template now carries them, and the anti-pattern row requires recording both the accountable human and the agent acting for them — logging only the human is now called out as the mirror-image failure.

**2 (`sub` collision)** — Docs default `sub` to `{{identity.email}}` and state verified corporate email as a hard prerequisite. Templates gain an optional `allowed_domains` list enforced at issuance: an identity whose email domain is outside it gets no token from that template, so `email_local` mappings can be pinned to the corporate domain.

**5 (issuance audit)** — Issuing now writes an `agent_task_tokens_issued` activity row tied to the run's issue, carrying the `jti` and expiry of every signed token (`Issue()` returns receipts alongside the env map). It fails closed: if the audit row cannot be written, the tokens are withheld.

**3 / 4 / 6 / 7** — All landed in the docs. Security boundaries now covers the inverted blast radius with least-privilege guidance (read-only systems first, prompt-injection framing), what a server compromise buys (cross-trust-domain impersonation, SSO-provider-shaped risk) and the explicit self-hosted boundary, and a plain statement that tokens are not refreshed — long runs beyond the TTL are unsupported, with TTL-sizing guidance instead of reaching for the 24h cap. The GHA OIDC analogy is replaced with RFC 8693 delegation in both the docs and the PR description.

**On dropping `MULTICA_TASK_TOKEN_MANIFEST_ENV`** — I'd prefer to keep it: we run a deployment where a tool consumes exactly this manifest to derive "which systems do I hold tokens for" from the tokens it actually holds. Happy to split it into its own PR if you still want the smaller v1 surface.

**Adoption** — the `issuer` + `jwks_uri` pair for zero-code gateway verification (Kong / APISIX / nginx / cloud gateways) is now documented; verification middleware packages left for a follow-up.

New coverage: interpolation of the actor variables end-to-end through the claim path, an allowed-domains matrix (outside-domain refusal, case-insensitive match, no-domain address, unrestricted templates unaffected), receipts matching the signed tokens' `jti`, and the audit row's presence + fail-closed semantics.

wxkingstar:
测试报告：
https://claude.ai/code/artifact/ec72755e-0182-4781-94ea-bcc6e896ec02

Bohan-J (changes requested):
Second round, verified against head `542beae4`.

First: the two blockers from the last round are properly fixed, not papered over. `agent.id` / `agent.name` / `task.id` are wired through all three places that matter (the variable set, `interpolate()`, and the `Context` the handler actually populates), and `allowed_domains` is validated at boot and enforced at issuance with a case-insensitive compare that a domain-less address can never satisfy. Thanks for that.

This round found four new implementation issues. The first is an authorization defect and is the one that matters.

## Blocking

### 1. The issuing gate reads the audit field, not the authorization field — so runs nobody authorized get a real person's token

`server/internal/handler/agent_task_token.go:224`:

```go
src := attribution.Source(task.OriginatorSource.String)
if !src.Precise() || !task.AccountableUserID.Valid {
```

`Precise()` (`attribution.go:74`) includes `trigger_owner` and `rule_owner`. And those two sources are, by this repo's own definition, exactly the case where no human authorized the run — `attribution.go:381`:

> Like `RuleOwner`, no human authorized the run, so **UserID (originator) stays NULL** and only the audit-accountable side is set — to `creatorUserID`, the trigger's member creator.

`RuleOwner` (`:366`) does the same thing. Both set `AccountableUserID` and leave `UserID` invalid.

The generated column comment on `accountable_user_id` states the rule directly:

> The one human accountable for this run, for audit / visibility / cost only — **NEVER consulted for authorization** (that is `originator_user_id`). […] the two diverge only when `originator_user_id IS NULL` (autopilot `rule_owner` / degraded `owner_fallback` name an accountable human while authorization carries none).

And the package doc opens with the same invariant — "attribution is 'on behalf of', never blame and never authorization. Nothing in this package is consulted for permission decisions."

So the concrete failure: an autopilot schedule fires at 3am. Nobody authorized it — `originator_user_id` is NULL. But `originator_source` is `trigger_owner`, `Precise()` returns true, and `accountable_user_id` names whoever created that trigger months ago. The server signs an ERP credential as that person, and the agent acts with their full permissions on a run they did not request and are not watching.

This is the same failure the PR already refuses for `owner_fallback`, and the PR's own description gives the reason: *"those mean no human authorized the run, and signing there would lend the agent owner's identity to work nobody asked for."* That reasoning applies unchanged to `trigger_owner` and `rule_owner` — the code just happens to classify them as "precise" for audit purposes, which is a different question.

Fix: gate on `task.OriginatorUserID` (the authorization human), or define a task-token-specific predicate. Don't reuse audit precision for an authorization decision — that's the exact coupling the attribution package was written to prevent.

`agent_task_token_issue_test.go` currently asserts `{"trigger_owner", true}` and `{"rule_owner", true}`. Those two rows are pinning the bug; they should become negative cases keyed on a NULL originator.

If the intent really is for a schedule to keep acting under its creator's authority indefinitely, that's a change to the authorization model and needs to be made explicitly — not arrived at by reading an audit column.

### 2. Mixed-version fleets record "issued" for tokens the daemon dropped

`daemon.go:2077` calls `issueTaskTokens` unconditionally, which signs and writes the `agent_task_tokens_issued` audit row. An older daemon ignores the new response field (non-strict JSON decoding, as the PR description notes), so the task runs with no token in its environment while the activity log says a credential was minted in that person's name.

This repo already has capability negotiation for exactly this — `requestClientCapabilities` at `daemon.go:1466`, reading `X-Client-Capabilities`, already used at `:471` and `:677`. Please add a task-token capability and fail closed before signing and auditing, plus a mixed-version test and a note on upgrade ordering.

For a feature whose headline is audit, a false "issued" row is worse than no row.

### 3. The Identity Tokens tab silently drops concurrent edits

`task-tokens-tab.tsx:67` computes the whole replacement set from the current `enabled` snapshot, but `:110` only disables the checkbox being toggled. Toggle A, then toggle B before A's response lands, and B's PUT is computed from the pre-A snapshot — A's change is silently discarded. With responses out of order the UI and server can also disagree.

Silent lost writes are bad anywhere; on the surface that decides which identities an agent may be issued they're not acceptable. Minimum fix: disable all checkboxes while a mutation is in flight. Better: serialize or coalesce, with a regression test for rapid toggling plus out-of-order responses.

### 4. "The whole catalog is validated at startup" isn't true yet

The docs promise the server refuses to boot rather than run on half a configuration. Three gaps:

- `catalog.go:137` checks duplicate `id` but not duplicate `env`. Two templates sharing an env name silently overwrite each other in `Issue()`'s map — one system's token is simply gone, with no error anywhere.
- `NewIssuer` validates the manifest env *name* but never checks it against template envs. A collision clobbers a token, and since the manifest is written after the token loop, the manifest wins.
- `NewIssuer` parses the key and never cross-checks it against each template's `algorithm`. `"algorithm": "RS256"` with an EC key boots cleanly and fails at `sign()` on every task — precisely the "fail at 3am instead of at boot" outcome the all-or-nothing validation exists to prevent.

All three belong in `NewIssuer`, with negative tests.

## Non-blocking

`issue.go` logs `task token issued` before the audit write succeeds, so a withheld token still leaves a log line claiming issuance. Moving that log after the audit write keeps `jti` forensics free of ghost entries. (Same nit as last round.)

## Summary

Product direction is unchanged — still worth doing, and round one's fixes were real. But #1 is an authorization defect that hands a real person's credentials to unauthorized automation, and it's a one-line-of-reasoning fix rather than a redesign. #2–#4 cover mixed-version correctness, a lost-write on an authorization surface, and a boot-time guarantee the docs already promise.

Happy to re-review once these land.

wxkingstar:
All four are addressed in add34f60 (code) + 2489082 (docs).

**1 (the gate read the audit column)** — Agreed, and the package's own comment
says it outright: `SourceTriggerOwner` … "originator stays NULL — an autonomous
fire carries no human authorization". The gate now reads `originator_user_id`
through a task-token-specific predicate, `taskTokenIdentityUser`. `Precise()`
survives only as a second condition and never the deciding one: it keeps
`backfill` — which *can* carry a non-NULL originator — from minting a live
credential for an attribution reconstructed after the fact.

The table is now crossed with authorization and has a fence from each side:
`precise_label_without_authorization` (source `direct_human`, accountable set,
originator NULL → nothing signed) and `backfilled_after_the_fact` (originator
set, source `backfill` → still nothing). Worth flagging that every positive row
in the old table set only `accountable_user_id` — the gate change turns them
all red until they gain an originator, which is the fence working as intended.

**2 (mixed-version "issued")** — Added `task-identity-tokens-v1`, advertised by
the daemon and checked *before* signing, so a daemon that cannot inject
produces neither a signature nor an audit row. Tested through
`ClaimTaskByRuntime` across three daemons (capable / `skill-bundles-v1` only /
advertising nothing), asserting both the token and the activity-row count, plus
a daemon-side assertion so dropping the advertisement cannot silently stop
issuance. Upgrade order — daemons first, then enable — is now in the docs.

**3 (lost writes in the tab)** — The payload is computed when the mutation
runs rather than when the box was clicked, toggles on one agent are scoped to
run one at a time, and every checkbox locks while a save is in flight. The
regression test is your scenario: two clicks in one synchronous block,
asserting the second PUT is `["app"]` rather than `[]`.

**4 (startup validation)** — All three fail the boot now: duplicate `env`, a
manifest variable colliding with a template's env, and an `algorithm` the
configured key cannot produce — including the curve, so `ES256` with a P-384
key is refused rather than failing inside `sign()`. The errors name only the
key's type and curve name; a test asserts no key material can reach them.

**nit** — `task token issued` moved after the fail-closed audit write and is
logged from the receipts, so a withheld token leaves no line claiming issuance.

The PR description is updated to match: the signed/refused source list, the
capability gate, and the boot-time validations.

wxkingstar:
Third round, in 23f6924d (code) + 2453b63b (docs), on top of today's merge of `main`.

**The gate had quietly inverted under MUL-6951.** `attribution.TriggerOwner` now sets `originator_user_id` to the trigger's creator (#7918), so the round-two gate — originator set *and* `Precise()` — had started signing for schedule fires, and for every `delegation` / `comment_source` hop below them, which copy that same human. The table test didn't notice because it hand-set the originator to NULL for `trigger_owner`, a shape the enqueue path no longer emits.

The gate now states the condition as what it means: the chain **began** with a member acting. A hop label is followed back to its root through the same lineage walk `DelegatedSubscriber` uses (`GetDelegatedSubscriptionFacts`, MUL-7051), and only a `direct_human` root signs. `trigger_owner` / `rule_owner` roots are refused — letting an automation invoke an agent under someone's rights is a bounded grant inside the product; lending that person's full permissions in an external system to an unattended run is a different one — and a lineage the walk can't prove is refused rather than assumed. The table now carries the post-MUL-6951 row (`trigger_owner` with a non-NULL originator → nothing), and a new test runs a three-hop chain rooted in `direct_human` (signs) and the same chain rooted in `trigger_owner` (doesn't).

Two more on the same surface:

- **Issuance moved after `FinalizeTaskClaim`** in both claim paths. It was signing and writing the audit row inside `buildClaimedTaskResponse`, while skill load, chat input, the remote-MCP token and finalize itself could all still requeue the task — leaving an "issued" row and a live JWT nobody received, then a second of each on the next claim. Same class as your round-two #2; this is the residual.
- **Membership is checked at claim time.** Attribution is decided at enqueue and a queue can wait days for an offline runtime; no credential is minted in the name of someone offboarded in between.

Smaller:

- **Codex** — its `shell_environment_policy` drops any name containing `TOKEN` unless explicitly authorized, and the daemon only authorized `custom_env` names, so a `BOT_TOKEN_*` the server had already audited read as empty from every Codex shell tool. Task-token names are now authorized the same way (`ZIC-82` policy predates this PR; it's a gap the PR always had). Tested both ways: present-and-issued is included, present-but-not-issued is still dropped.
- **One reserved env list.** The catalog's boot-time copy of the daemon blocklist had already drifted (`CURSOR_MCP_AUTH_SOURCE` missing, so a template could name a variable the daemon then silently dropped — the exact thing the check exists to prevent). Moved to `pkg/protocol`, used by both, with a test pinning it to `execenv.CursorMcpAuthSourceEnv`.
- **Tab** — a save response that fails schema validation was degrading to the EMPTY shape and being written over the cache, which hid the tab mid-edit and handed the next queued toggle an empty set to build its PUT from; it now refetches. Pending / failed loads no longer render as "nothing configured". The tab's visibility gate gained pane-level tests (it had none — the pane test wasn't mocking the api it now calls).
- **Docs** — the signed / refused table now names the deciding factor (chain root), says outright that an armed trigger *does* carry its creator's authorization inside Multica and is still refused here, and adds the membership requirement. zh/ja/ko pages brought in line with the i18n glossary (`autopilot` → 自动化, an agent run → 运行 / 実行 / 실행).

Verification: handler + tasktoken + daemon test packages green against a real Postgres, including the claim paths (the one failure, `ClaimPollHintSchedulesNextDeferredTask`, is a 5034 ms vs 5000 ms timing assertion that fails identically on unmodified `main`). Views: 188 tests, `tsc`, `eslint` clean.

The PR description is updated to match. Happy to re-review whenever suits.

PR #7449, opened 2026-08-23
MUL-6582: add schedule-pulse example plugin

forrestchang:
Follow-up to #7448. Adds a real scheduled plugin that Multica can install and wake every five minutes.

**What it is**
- `examples/plugins/schedule-pulse`: manifest with a `schedule` trigger, HMAC-verifying handler, workspace-storage pulse, and a panel that shows the last delivery.
- Each unique `delivery_id` is written once. A retry of the same planned wake returns `{duplicate: true}` and does not increment the count.

**Verification**
`TestSchedulePulseExampleManifestDrivesDurableRetry` installs this example manifest (not a fixture), runs two scheduler replicas, drops the first HTTP response after storage is written, and asserts the retry keeps the same `delivery_id` with `count == 1`.

Does not add issue create, Discord logic, or `Closes MUL-6582` — that belongs on #7448.

### DIFF
--- a/examples/plugins/schedule-pulse/README.md
+++ b/examples/plugins/schedule-pulse/README.md
@@ -1,0 +1,48 @@
+# Schedule Pulse
+
+A real scheduled plugin, not a log line. Multica POSTs the plugin's own HTTPS
+endpoint every five minutes. The handler records each unique `delivery_id` in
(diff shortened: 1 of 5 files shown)


forrestchang:
已合并进 #7448（包含 schedule-pulse 示例和端到端验证）；后续只在 #7448 继续 review。

PR #7455, opened 2026-08-23
fix(runtime): classify storage exhaustion (MUL-6589)

leilei-chen-sto:
## Summary

- classify task-environment disk exhaustion as `runtime_storage_exhausted` before provider error parsing
- preserve the classification across the preparation-helper boundary
- recognize native Unix and Windows disk-full errors
- normalize legacy Linux and Windows daemon disk-full wire shapes without rewriting unrelated agent errors
- show the new reason as a runtime failure in the dashboard

## Root cause

An execution-environment path contained a UUID segment such as `-401e-`. When the host returned `no space left on device`, the generic provider classifier matched `401` inside that path and persisted `agent_error.provider_auth_or_access`, even though the agent process never started.

## Verification

- `go test ./pkg/taskfailure ./internal/daemon/execenv ./internal/daemon ./internal/service`
- `go vet ./pkg/taskfailure ./internal/daemon/execenv ./internal/daemon ./internal/service`
- Linux and Windows AMD64 daemon test binaries compile
- `pnpm --filter @multica/core exec vitest run dashboard/failure-class.test.ts` — 4 tests passed
- targeted ESLint and `git diff --check`

## Scope

No retry policy, dependency, localized copy, or historical backfill is added. Existing rows are unchanged; this PR corrects future failures and mixed-version Linux and Windows daemon reports.

### DIFF
--- a/packages/core/dashboard/failure-class.test.ts
+++ b/packages/core/dashboard/failure-class.test.ts
@@ -18,6 +18,7 @@ describe("failureClassOf", () => {
     // MUL-5370: the run never reached the model provider, so this belongs
     // with the substrate failures an operator fixes by checking the daemon.
     expect(failureClassOf("skill_bundle_unavailable")).toBe("runtime");
+    expect(failureClassOf("runtime_storage_exhausted")).toBe("runtime");
     expect(failureClassOf("agent_error.process_failure")).toBe("agent");
     // MUL-5722: the daemon and the provider are both healthy — codex could
     // not hand its own stored thread back — so this reads as an agent-side
--- a/packages/core/dashboard/failure-class.ts
+++ b/packages/core/dashboard/failure-class.ts
@@ -70,6 +70,7 @@ const REASON_CLASS: Record<string, FailureClass> = {
   // fix is on the host running the daemon — a faster CLI or a raised timeout —
   // not with the model provider, which was never contacted.
   runtime_cli_timeout: "runtime",
+  runtime_storage_exhausted: "runtime",
 
   // The agent process itself produced the failure.
   "agent_error.process_failure": "agent",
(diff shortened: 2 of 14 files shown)


multica-eve (changes requested):
The product direction is right: failures caused by a full runtime disk should point members to the runtime host, not to model credentials. The current implementation still leaves three user-facing and classification gaps that need to be fixed before merge:

1. **Add the user-facing labels for `runtime_storage_exhausted`.** The new reason is registered only in `packages/core/dashboard/failure-class.ts`. It is still missing from `apps/mobile/lib/failure-reason-label.ts`, `apps/mobile/components/issue/run-row.tsx`, `packages/views/agents/components/tabs/task-failure.ts`, and `packages/views/chat/components/chat-message-list.tsx`. Depending on the surface, members will see `Failed`, the raw wire value, or a generic error instead of being told to free disk space. On mobile this is less informative than the current, albeit incorrect, provider-auth message.

2. **Handle disk quota exhaustion as storage exhaustion.** On Unix, `isPlatformStorageExhausted` currently recognizes only `syscall.ENOSPC`. Hosts and containers with filesystem quotas commonly return `EDQUOT` (`disk quota exceeded`). That text is currently classified as `agent_error.provider_quota_limit`, sending members toward model billing instead of the runtime disk. Please recognize `syscall.EDQUOT` and normalize the legacy `disk quota exceeded` form, with regression tests.

3. **Do not restrict legacy storage normalization to only three classifier outputs.** The bug mechanism is numeric fragments in preparation paths colliding with provider HTTP-code patterns. The same raw storage error can therefore first land in `provider_capacity_or_rate_limit` or `provider_server_error`, which are excluded by `legacyStorageExhaustedReasons`. Preparation failures can also come from subprocesses such as Git and carry only text, so the structured sentinel cannot cover every case. Please make the normalization key off the trusted `prepare execution environment:` prefix plus a recognized disk-full phrase for any classifier-produced `agent_error.*` reason (or otherwise cover every provider bucket), and add collision tests such as path fragments containing `529` and `512`.

Once these three gaps are addressed, the overall sentinel/helper-boundary approach looks good and the change should be ready for another review.

PR #7461, opened 2026-08-23
MUL-6555 feat(issues): preserve source context for sub-issues

### LINKED ISSUE #7415: Create sub-issue from historical comment context
## Product goal

Let a user create a real sub-issue from a precise point in an existing issue discussion without re-running the historical comment.

From any eligible comment, the user chooses **Branch into sub-issue**, enters a new task, and creates a normal sub-issue. The new issue receives a server-owned, immutable snapshot of the historical context that led to that point. The assigned agent works from both the new issue's current task state and the captured historical background.

This replaces the independent comment-run direction previously planned as PR 2 in #7033. It is a different product model: the durable unit of work is a sub-issue, not a branch-specific agent task or a new result comment in the source issue.

## User experience

- Add **Branch into sub-issue** to eligible root and reply comment menus on web/desktop.
- Open the existing unified issue-create dialog in **With Agent** mode by default; allow switching to **Manual** without losing the selected source or the user's draft.
- Require a non-empty new agent prompt in With Agent mode or a non-empty issue title in Manual mode. Do not prefill either with the selected historical comment.
- Keep the ordinary **Add sub-issue** entry and its existing default behavior unchanged.
- Show a compact source badge on the created sub-issue and provide a read-only **Context snapshot** viewer.
- Keep the immutable snapshot and live-source navigation as separate actions. From the snapshot viewer, **Locate branch point** revalidates the live source before navigating to the exact selected comment whenever that anchor still exists.

## Frozen context boundary

Capture, at submission time:

- the source issue identifier, title, and description;
- the complete chronological history of the selected comment's root thread through the selected-comment boundary, ordered oldest to newest with comment ID as the stable timestamp tie-breaker;
- earlier sibling branches and their descendants when they fall at or before that boundary;
- captured author identity and timestamps for every included comment; and
- issue attachments and attachments owned by included comments, copied into snapshot-owned storage so deleting the source cannot break the captured context or its embedded attachment links.

Explicitly exclude:

- comments after the selected-comment boundary, including later siblings and descendants of the selected comment;
- unrelated root threads;
- the source issue's agent session, transcript, workdir, or execution output; and
- recursive embedding of an upstream snapshot when the source is itself a context-derived sub-issue.

The snapshot is immutable and separate from the editable child issue description. Multiple sub-issues created from the same source remain isolated from one another.

## Execution semantics

- The created object is a normal sub-issue with `parent_issue_id` initially set to the source issue.
- Normal issue assignment, agent concurrency, queueing, retries, sessions, comments, and status behavior continue to apply.
- An agent target remains a normal direct issue assignee.
- A squad target remains a real squad assignee and uses the existing squad-leader evaluation/delegation path; a squad must not be reduced to a shortcut for its leader agent.
- Every run on the child receives its current issue state plus the same frozen source context.
- With Agent creation gives the quick-create agent two explicitly separated inputs: the new task instruction and read-only historical context.
- A failed issue-less With Agent quick-create retains the exact pending snapshot and cloned attachments for retry instead of recapturing the live source. Retry ownership transfers atomically between attempts, only the original requester may manually retry, and at most one created issue may attach the snapshot.
- Terminal pending captures are retained for up to 30 days to support eligible retries; abandoned snapshots and cloned objects are then removed by the bounded cleanup sweeper.

## Source changes and deletion

- Editing the source issue or any captured comment never mutates the snapshot.
- Change detection compares canonical issue/comment content, structure, and attachment references to the captured content. Revision counters, update timestamps, and author display-name changes alone do not mark the captured content as changed.
- Editing, deleting, adding, or reparenting comments inside the original capture boundary is reported as a thread change. Live navigation remains available while the selected anchor comment and source issue still exist.
- Deleting the selected anchor comment leaves the child and snapshot readable and disables exact-comment navigation.
- Deleting the source issue leaves the child intact as an independent top-level issue, with `parent_issue_id = null` and `stage = null`; the snapshot remains readable and the UI reports **Source issue deleted**.
- Captured author names remain historical labels. Renaming, archiving, removing, or deleting an author changes only the explanatory current-state label, never the reference identity.

## Acceptance criteria

- [ ] A user can branch a new sub-issue from an eligible root comment or nested reply and must enter a new agent prompt or Manual issue title.
- [ ] With Agent is the default for this entry; switching between Agent and Manual preserves the source and draft.
- [ ] The snapshot contains the source issue body and the complete chronological history of the selected comment's root thread through the selected-comment boundary.
- [ ] Earlier sibling branches inside the boundary are included; comments after the boundary and unrelated root threads are absent.
- [ ] The snapshot and copied attachments remain readable after source edits, comment deletion, or source issue deletion, and embedded attachment links resolve through the snapshot-owned clones.
- [ ] The snapshot is read-only and is not merged into the editable child description.
- [ ] Normal child comments and later runs continue to receive the same snapshot alongside the child's current state.
- [ ] Multiple children from one source have separate issue IDs and normal issue-scoped sessions.
- [ ] Agent and squad targets preserve existing normal-issue semantics.
- [ ] Source-change UX distinguishes unchanged, changed details, anchor-comment-deleted, source-issue-deleted, and temporarily unavailable states without relying on author names as identifiers.
- [ ] Deleting the source issue explicitly detaches surviving children, clears stage, and publishes normal issue updates so clients do not retain stale hierarchy state.
- [ ] Database ownership changes, source-context attachment, quick-create attachment transfer, and target issue creation are atomic and tenant-scoped; external object copies and deletions are protected by durable cleanup intents.
- [ ] Failed With Agent retries reuse the same pending snapshot and cannot attach it to more than one created issue.
- [ ] Old servers/daemons never silently degrade the action into a sub-issue or quick-create request without frozen context.
- [ ] Existing ordinary issue creation and Add sub-issue behavior do not regress.

## v1 scope

- Web and desktop are included. Mobile currently lacks both the ordinary sub-issue creation entry and the With Agent create flow; mobile support should build on the same server contract in a follow-up rather than expose a degraded action.
- Source state is recomputed by the issue-detail endpoint. Issue detail always refetches on mount, and **Locate branch point** refetches immediately before live-source navigation. Window-focus refresh and cross-issue realtime fan-out for source edits/deletion are not required in v1; deleting the parent issue itself must still publish normal child issue updates because it changes child hierarchy state.
- A capture is limited to 256 comments, a 1 MiB serialized snapshot, 100 attachments, and 500 MiB total attachment bytes. The action fails closed with a structured error when a limit is exceeded.

## Non-goals

- Re-running a historical comment.
- Branch-specific task kinds, queues, sessions, scheduling, timeouts, nudges, or automatic return-to-source behavior.
- Publishing a branch result as a new top-level comment in the source issue.
- Automatically merging or selecting a winning branch.
- General issue revision history/rollback or protected-specification work.
- A global issue-create idempotency redesign.

## Related history

- #7038 completed the collaborative edit-safety portion of #7033 independently.
- #7033's remaining independent-run PR 2 direction is superseded by this sub-issue model.
- Draft PR #7357 implemented the superseded independent-run model and is being closed rather than adapted in place.

yyclaw:
## What does this PR do?

Adds **Branch into sub-issue** to eligible issue comments on web and desktop. The action creates a normal sub-issue while preserving the exact historical issue/comment context that led to it as a server-owned, immutable snapshot.

### Thinking path

1. A historical comment should start a durable unit of work, so this builds on the existing sub-issue model instead of introducing a second comment-run lifecycle.
2. Live issue and comment content can be edited or deleted, so the server captures the source issue body, the complete chronological root-thread history through the selected boundary, and snapshot-owned attachment clones at submission time.
3. The snapshot is historical background, not editable task input. It stays separate from the child description and is rendered in a read-only context viewer with independent live-source navigation.
4. The child remains a normal issue: existing assignment, squad delegation, queueing, retry, session, comment, and status behavior still applies. Every later run receives the same frozen context alongside current child state.
5. Compatibility and cleanup fail closed: unsupported servers/runtimes cannot silently create a context-free request, failed quick-create attempts retain the exact snapshot for an authorized retry, and durable cleanup intents handle external object copies/deletions.

## Related Issue

Closes #7415

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Added source-context capture, immutable snapshot storage, canonical change detection, attachment cloning, retry ownership, cleanup intents, and tenant-scoped APIs in `server/internal/{handler,service,util}`.
- Added reversible migrations `407`–`414` and sqlc queries for source contexts, object intents, uniqueness/lookup indexes, and source-context attachment ownership.
- Extended issue creation, issue detail, deletion, task claiming, daemon prompts, and WebSocket/cache coordination so context-derived children keep ordinary issue semantics.
- Added the comment action, shared create-dialog integration, draft preservation, structured compatibility/limit errors, context preview/viewer, and live branch-point navigation in `packages/{core,views}`.
- Added localized web/desktop copy for English, Simplified Chinese, Japanese, and Korean, checked against the repository terminology conventions.
- Added backend integration/unit tests, migration round-trip coverage, API/schema/store tests, and UI behavior tests.

### Risks and rollout notes

- With Agent creation is capability-gated. An old server or runtime returns a structured unsupported error instead of silently dropping frozen context; Manual remains available when only the runtime lacks quick-create support.
- Source capture is bounded to 256 comments, 1 MiB serialized text, 100 attachments, and 500 MiB total attachment bytes and fails closed when a limit is exceeded.
- Rolling migration `407` down intentionally removes source-context rows and snapshot-owned attachment data introduced by this feature. Do not roll the schema back after production use if captured contexts must be retained; the isolated migration round-trip test verifies and documents this data-loss boundary.
- The branch is one commit behind current `main`; the intervening upstream change only touches task-transcript build-step tests, and a merge-tree check against current `upstream/main` is conflict-free.

## How to Test

1. Run backend verification:
   ```bash
   cd server
   go test ./internal/service -run 'Test(SourceContext|PrepareSourceContext|EnqueueQuickCreateTaskWithSourceContext|FailTaskDoesNotRetryAfterSourceContext|ManualSourceContext|CleanupSourceContext)' -count=1
   go test ./internal/handler -run 'Test(WriteSourceContext|CommentSourceContext|SourceContext|ClaimTask_SourceContext|WorkspaceDeletionManifest|DeleteWorkspace_)' -count=1
   go test ./internal/daemon -run 'TestBuildQuickCreatePrompt' -count=1
   go test ./internal/util -run 'TestAttachmentIDFromDownloadURL' -count=1
   go test ./internal/storage -run 'TestS3StorageUploadStream' -count=1
   go test ./cmd/migrate -run 'Test(SourceContextMigrationsRoundTripWithDocumentedDataLoss|ConcurrentIndexCleanupsMatchTheirMigrations)' -count=1
   go vet ./internal/util ./internal/service ./internal/handler ./internal/daemon ./internal/storage ./cmd/server ./cmd/migrate
   ```
2. Run frontend verification:
   ```bash
   pnpm typecheck
   pnpm lint
   pnpm --dir packages/core exec vitest run api/schema.test.ts api/schemas.test.ts issues/mutations.test.tsx issues/queries.test.ts issues/stores/draft-store.test.ts issues/ws-updaters.test.ts
   pnpm --dir packages/views exec vitest run inbox/components/inbox-page.test.tsx issues/components/issue-detail-route.test.tsx issues/components/issue-detail.test.tsx issues/components/source-context-viewer.test.tsx modals/create-issue-dialog.test.tsx modals/create-issue.test.tsx modals/quick-create-issue.test.tsx modals/source-context-preview.test.tsx
   ```
3. From an eligible root comment or nested reply, choose **Branch into sub-issue**. Verify With Agent is selected, switching to Manual preserves source and draft, and submission requires a new prompt/title.
4. Verify the child shows the immutable snapshot; then edit/delete the source comment, rename its author, and delete the source issue. Confirm the viewer reports each state while captured content and attachment links remain readable.
5. Verify **Locate branch point** navigates to the live anchor while it exists and becomes unavailable after anchor/source deletion.

## Checklist

## AI Disclosure

**Prompt / approach:**
Used Codex to implement #7415 against the repository's current architecture, reconcile the historical-comment workflow with normal sub-issue semantics, and run `review-loop-plus` followed by a final `pr-preflight`. Review findings were fixed in the same commit, including replacing regex-based attachment URL matching with structured URL parsing, preserving complete chronological thread context, and strengthening retry, cleanup, compatibility, and migration tests.

## Screenshots (optional)

The [issue evidence comment](https://github.com/multica-ai/multica/issues/7415#issuecomment-5385273013) shows the new UI in its normal, source-changed, branch-point-removed, and author-renamed states. Before this change, the comment menu has no **Branch into sub-issue** action or context snapshot viewer.


multica-eve (changes requested):
The product direction is valuable: turning a historical discussion point into a normal sub-issue with a frozen, read-only source snapshot fits the existing issue workflow and avoids treating old comments as new instructions. I’m requesting changes because the current implementation has three production-blocking gaps:

1. **The accepted snapshot size can exceed the agent context window before issue creation starts.** `server/internal/service/source_context.go:26` accepts up to 1 MiB of serialized JSON, while `server/internal/daemon/prompt.go:211-218` embeds the entire payload verbatim in the first quick-create prompt. This repository already handles real 200,000-token context limits; a 1 MiB JSON payload plus the system prompt and tool definitions can exceed that limit. The server can therefore accept a snapshot that the With Agent path cannot execute. Please either enforce a conservative prompt/token budget with headroom, or deliver the snapshot out of band, and add a boundary test covering the largest accepted agent-mode snapshot.

2. **The down migration leaks snapshot attachment rows and their stored objects.** `server/migrations/407_issue_source_context.down.sql:1-3` drops `attachment.source_context_id` and then drops the context and cleanup-intent tables without first handling snapshot-owned attachments. I reproduced this against a fully migrated isolated database: after the down SQL, the attachment row remained with every normal owner column null, while its cleanup intent was gone. `server/cmd/migrate/migrate_source_context_test.go:58-80` only inserts a context row, so the round-trip test misses this. Please provide a safe pre-down cleanup/rollback path (or fail closed when snapshot data exists) and cover attachment rows plus object intents in the migration test.

3. **Task-token calls are attributed to the owning member.** The new route at `server/cmd/server/router.go:1978-1980` is not human-only, but `server/internal/handler/source_context.go:688` hard-codes `CreatorType: "member"`. A running agent can obtain a preview token and invoke this endpoint; because task-token auth exposes the owner’s user ID, the created issue, events, and capture are recorded as if the human performed the action. Please either apply `RequireHumanActor` (consistent with the retry endpoint) or preserve the actual agent actor and explicitly define retry ownership. Add a route-level task-token test for the selected contract.

Non-blocking: `server/internal/daemon/prompt.go:195` and `:414` add nearly duplicate source-context guidance to every issue/comment turn. Consider keeping one stable rule in the cache-stable runtime brief, or injecting it only when relevant. The required `ByteIdentical` test currently passes.

Validation performed on commit `3c5d770a2b4e19e212e499fdc6f376a92bf88af5`: targeted backend/service/handler/daemon/migration tests passed against an isolated fully migrated PostgreSQL database; `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` and `git diff --check` passed; GitHub CI is green and the PR remains conflict-free.

yyclaw:
@multica-eve Thanks for the review. Commit `3d7f3c246` addresses all three blocking findings and the non-blocking suggestion:

1. The With Agent path now projects the final persisted snapshot without side effects before any object cloning or task enqueue. Agent snapshots are capped at 64 KiB, snapshot plus new instruction at 72 KiB, and the complete quick-create user prompt has a 96 KiB regression ceiling. Manual mode keeps its existing 1 MiB limit.
2. Migration 407 now fails closed while any source context, snapshot attachment, or object intent remains. The migration test covers all three data types, verifies that a rejected rollback preserves them, and verifies that retry succeeds after explicit cleanup.
3. Both preview/create routes now use `RequireHumanActor`. A real task-token router integration test verifies that both return 403 while the adjacent comment route remains outside the guard.
4. The source-context instruction-precedence rule now lives once in the cache-stable runtime brief, with the duplicate assignment/comment per-turn guidance removed.

Locally passed: the new service budget boundary tests, targeted daemon prompt tests, the full daemon package, the full execenv package including ByteIdentical, the targeted handler test, the real PostgreSQL migration rollback test, the real task-token router integration test, and `git diff --check`.

Please take another look at `3d7f3c246`.

multica-eve (changes requested):
Thanks for addressing the previous findings. The agent input budget, human-only route guard, and prompt deduplication are now resolved.

One must-fix rollback issue remains:

**[P1] The rollback guard runs after migrations 414–408 have already been reverted.** The migration runner intentionally executes each file separately (because concurrent index operations cannot share one transaction). On a `down` run, migrations 414 through 408 therefore drop all seven source-context indexes and remove their `schema_migrations` records before the data guard in `server/migrations/407_issue_source_context.down.sql` is reached.

I reproduced this against an isolated fully migrated database containing source-context data: 414–408 completed, 407 failed with the intended guard error, but all seven indexes were gone and only migration 407 remained recorded while the source-context rows and tables were still present. This leaves live data without the identity/uniqueness constraints for context IDs, attached issues, pending origin tasks, and object-intent storage keys, as well as without the lookup/cleanup indexes.

Please move the preflight to the first reverse step (for example, a guard or down hook for migration 414) so it fails before any index is dropped. The migration test should also assert that, after a rejected rollback, all seven indexes and the 408–414 ledger entries are unchanged; its current table/row assertions do not catch the partial rollback.

yyclaw:
@multica-eve Thanks for catching the per-file rollback issue. I reproduced and confirmed that keeping the guard only in migration 407 allowed 414–408 to roll back first.

Commit `7b46d826d` moves the same ownership preflight into the migration runner's pre-rollback hooks and registers it on every step from 407 through 414. Therefore, whether the database is fully applied through 414 or stopped at an intermediate version, the first applied reverse step fails closed before any index or migration-ledger entry is removed. The SQL guard in 407 remains as defense in depth for direct SQL execution.

The migration test now additionally asserts that after a rejected rollback:

- all seven source-context indexes remain valid and ready;
- all eight 407–414 `schema_migrations` entries remain recorded;
- the context, snapshot attachment, and object-intent rows remain intact; and
- after explicit data cleanup, retrying the same down run completes successfully.

The targeted tests and the full `go test ./cmd/migrate -count=1` package pass locally. Please take another look at `7b46d826d`.

Bohan-J (approved):
Thanks for the thorough work on this, @yyclaw — and for turning around all three rounds of review feedback so quickly.

The last pre-merge check was migration numbering. **No conflict:**

- `main` currently ends at `403_runtime_profile_add_zeroclaw`; this PR occupies `407`–`414`, so every numeric prefix is unique.
- `TestMigrationNumericPrefixesStayUniqueAfterLegacySet` and `TestNewMigrationPrefixesStartAfterLegacyRange` both pass on the merge result.
- The `404`–`406` gap is harmless by design: the runner records each version individually (`EXISTS ... WHERE version = $1`) rather than tracking a high-water mark, and `AllVersions()` deliberately verifies every version precisely so a migration numbered below an already-applied one still gets applied when it lands later. The lint enforces uniqueness, not contiguity — so renumbering would be churn without a correctness benefit.
- All eight pre-migration preflight hook registrations in `server/cmd/migrate/main.go` match the on-disk filenames.

Verified against the actual merge of this branch with current `main`, which is 23 commits ahead of the PR base (CI here ran on the older base):

- `go build ./...` clean; `go test ./cmd/migrate/... ./internal/migrations/...` pass
- 63 daemon prompt / source-context tests pass; `execenv -run ByteIdentical` passes
- `pnpm typecheck` 7/7; `packages/core` 457 tests and `packages/views` inbox + source-context 150 tests pass
- The files both sides touched — `internal/daemon/prompt.go`, `packages/core/api/schemas.ts`, `packages/core/issues/mutations.ts`, `packages/views/inbox/components/inbox-page.tsx` — merge cleanly and typecheck
- The remaining local daemon test failures are environment-specific and reproduce identically on unmodified `main`

Heads-up for the other in-flight migration PRs: once this lands, `407`–`414` are taken, so #7491, #7481 and #7400 will need to renumber before they merge.

Merging now — thanks again for the careful work on the rollback preflight in particular.

Issue #6146, opened 2026-07-29
.env.example variable references do not cascade through docker-compose interpolation


mingxin:
## Summary

`.env.example` uses `${FRONTEND_PORT}` inside derived variables:

```env
FRONTEND_PORT=3000
FRONTEND_ORIGIN=http://localhost:${FRONTEND_PORT}
MULTICA_APP_URL=${FRONTEND_ORIGIN}
GOOGLE_REDIRECT_URI=${FRONTEND_ORIGIN}/auth/callback
```

But docker-compose does **not** perform recursive interpolation on `.env` values. When a user changes `FRONTEND_PORT`, the derived variables (`FRONTEND_ORIGIN`, `MULTICA_APP_URL`, `GOOGLE_REDIRECT_URI`) do not update — they receive the literal string `http://localhost:${FRONTEND_PORT}` instead of the resolved URL.

## Reproduction

1. In `.env`, set `FRONTEND_PORT=43927`
2. Run `docker compose -f docker-compose.selfhost.yml up -d`
3. Inspect the backend container's environment:
```bash
docker exec multica-backend-1 env | grep FRONTEND_ORIGIN
# Output: FRONTEND_ORIGIN=http://localhost:${FRONTEND_PORT}  ← literal, not resolved
```

This causes:
- CORS origin mismatch (backend expects `localhost:3000`, frontend is on `localhost:43927`)
- OAuth redirect URI pointing to wrong port
- WebSocket connection failures

## Workaround

Users must manually set all derived variables to absolute URLs:
```env
FRONTEND_PORT=43927
FRONTEND_ORIGIN=http://localhost:43927
MULTICA_APP_URL=http://localhost:43927
GOOGLE_REDIRECT_URI=http://localhost:43927/auth/callback
```

## Suggested fix

**Option A** — Move the derivation into docker-compose YAML where `${VAR}` substitution works:

```yaml
environment:
  FRONTEND_ORIGIN: ${FRONTEND_ORIGIN:-http://localhost:${FRONTEND_PORT:-3000}}
```

**Option B** — Add a comment in `.env.example` warning that `${VAR}` references inside `.env` values are not interpolated by docker-compose, and all derived variables must be updated manually when changing ports.

## Environment

- Multica v0.4.12, Docker self-host, docker-compose v5.1.2

multica-eve:
[...]

PR #7466, opened 2026-08-23
MUL-6611: docs(self-host): route /health through the proxy, add single-origin example and release-tag note

felix021:
## Summary

Documents four traps we hit deploying twice from the quickstart alone, applied to all four doc locales (en / zh / ja / ko):

- **`/health` through the proxy.** `multica setup` probes `<server-url>/health` and needs a 200. Older web releases do not forward the path, so a reverse proxy that sends everything to such a frontend returns 404 and setup fails with "Server not reachable" against a healthy stack. Main now forwards `/health` (#7467); the docs keep the explicit direct-to-backend route as the version-agnostic fix and qualify the diagnosis accordingly. Added the route to the Caddy examples, a single-origin proxy variant (the shape small servers actually use: one domain / one port, with `/health`, `/ws`, and `/api/daemon/ws` called out), a new troubleshooting section, and a hint at the CLI setup step.
- **Release-tag alignment.** The published images track the newest release tag while a plain `git clone` checks out `main`. Anyone building the CLI or images from the checkout should pin the tag first, or built binaries drift from the running containers (daemon protocol included).
- **`/api/config` verification.** After `up -d`, `daemon_server_url` must print the public URL: `MULTICA_PUBLIC_URL` when set, otherwise the app URL — `MULTICA_APP_URL`, falling back to `FRONTEND_ORIGIN`; when that chain is empty the field is omitted from the response entirely, even if `MULTICA_PUBLIC_URL` is set. If it prints `localhost`, set `FRONTEND_ORIGIN`/`MULTICA_APP_URL` explicitly instead of relying on the `${FRONTEND_PORT}` references from `.env.example` (refs #6146).
- **WS probe behavior.** A plain `curl` against `/ws` returning HTTP 400 is expected — the endpoint rejects a handshake without workspace-scoped query parameters before any WebSocket upgrade. That 400 only proves plain HTTP routing reaches the backend; the real connection check is a WebSocket `101` handshake in DevTools → Network → WS (or a WebSocket-aware client).

## Tests

- Docs-only change; verified each locale keeps its own heading anchor (`#3-…` links, the Step 1 self-links, and the new troubleshooting section links resolve within the same locale).
- All commands in the examples were executed against a real single-origin deployment (nginx variant of the same topology) during the deployment these notes came from.
- The `daemon_server_url` precedence and omission condition match `daemonSetupURLsFromEnv` in `server/internal/handler/config.go` and its tests (`TestGetConfigUsesFrontendOriginForSameOriginDaemonSetup`).


multica-eve (dismissed):
This is a useful addition for self-hosted administrators: the single-origin example and release-alignment guidance cover real setup failures. I found three accuracy issues that can misdirect operators, so I am requesting changes before merge.

### Must-fix

1. **The Docker Compose version claim does not match a reproducible test.** `apps/docs/content/docs/self-host-quickstart.mdx:184` and the three localized counterparts say releases older than 2.24 leave the nested `.env` references literal. I tested the official 2.23.3 and 2.24.0 binaries with:

   ```dotenv
   FRONTEND_PORT=3100
   FRONTEND_ORIGIN=http://localhost:${FRONTEND_PORT}
   MULTICA_APP_URL=${FRONTEND_ORIGIN}
   ```

   Both versions resolve `FRONTEND_ORIGIN` and `MULTICA_APP_URL` to `http://localhost:3100`. Please remove the unverified `< 2.24` cutoff. The operational advice can simply say to set these values explicitly when `/api/config` still reports localhost, or it can name a narrower version/invocation if there is a concrete failing reproduction.

2. **The `/health` wording is already stale relative to `main`.** #7467 is merged, and `apps/web/config/runtime-urls.ts:137-144` now forwards `/health` to the backend. The unqualified statements in `self-host-quickstart.mdx:146`, `self-host-quickstart.mdx:249`, and `troubleshooting.mdx:125` that the frontend does not serve the path and that this error is “almost always” a proxy 404 will therefore be false when this PR lands. Keeping an explicit direct-to-backend route is safe and helps older images, but please qualify the diagnosis as applying to older web images or proxies that do not route `/health`, while retaining DNS, TLS, firewall, timeout, and 5xx failures as other causes. Please update all four locales consistently.

3. **A bare `curl` returning 400 does not validate the WebSocket upgrade path.** `troubleshooting.mdx:111` says a backend 400/401 means “the link itself works,” but a plain curl does not send a WebSocket Upgrade. The handler also returns 400 for missing workspace parameters at `server/internal/realtime/hub.go:776-789`, before reaching the upgrader. This proves only that ordinary HTTP routing reaches the backend; it does not prove that the proxy preserves Upgrade headers or that a realtime connection can be established. Please state that boundary in all four locales and keep Browser DevTools → Network → WS (or a WebSocket-aware probe) as the actual connection check.

### Nit

- `apps/docs/content/docs/troubleshooting.ja.mdx:113`: `backend へ転達してください` should be `backend へ転送してください`.

The Caddy route structure and four-locale organization otherwise look clear. Once these factual and diagnostic issues are corrected, the documentation should be ready to merge.

felix021:
Thanks for the thorough review — all points addressed in 50e2072 (en/zh/ja/ko kept in sync):

1. **Compose version claim removed.** The `< 2.24` cutoff is gone everywhere; the guidance now says to set `FRONTEND_ORIGIN` / `MULTICA_APP_URL` explicitly when `/api/config` prints a `localhost` address, then recreate the containers.

2. **`/health` wording qualified.** All statements now say current web releases forward the path while older ones do not (correct after #7467), the explicit direct-to-backend route is kept as the version-agnostic recommendation, and the probe failure causes list DNS / TLS / firewall / timeout / 5xx. This covers the Caddy comment, the `multica setup` section, the FAQ row, and the troubleshooting page in all four locales.

3. **curl-400 boundary stated.** The troubleshooting page now says a bare `curl` 400 only proves plain HTTP routing reaches the backend (the workspace-param rejection happens before the upgrader), and points at DevTools → Network → WS for a real `101` handshake check — in all four locales.

4. **Nit fixed:** 転達 → 転送. I also fixed the stale in-page "Step 1" anchors in the upgrade section (they predated a heading rename).

multica-eve (dismissed):
Thanks for the update. The Compose cutoff is removed, the plain-HTTP-vs-WebSocket boundary is now accurate, the Japanese typo is fixed, and the expanded reachability failure list is helpful. I still see one factual issue in the final documentation.

### Must-fix

**The documented `daemon_server_url` precedence omits a real fallback and an omission condition.** `apps/docs/content/docs/self-host-quickstart.mdx:184` and the other locales say the value is `MULTICA_PUBLIC_URL`, otherwise `MULTICA_APP_URL`, and that empty output means neither is set. The implementation at `server/internal/handler/config.go:127-152` first resolves the app URL as `MULTICA_APP_URL`, falling back to `FRONTEND_ORIGIN`; if that app URL is empty, both daemon URLs are omitted even when `MULTICA_PUBLIC_URL` is set. `server/internal/handler/config_test.go:165-187` explicitly covers the `FRONTEND_ORIGIN` fallback. Please document all three inputs and the actual omission condition in all four locales.

### Follow-up

The PR description still repeats the superseded `< 2.24`, unqualified `/health`, and “400/401 proves the proxy chain works” claims. Please update it as well so the permanent PR record matches the corrected diff.

`git diff --check` passes, the refreshed anchors match the four current Step 1 headings, and CI for `50e2072785f15d973f17fb564c9e9e2bcff4ba16` is green. Once the documentation fact above is corrected, I do not see another merge blocker.

felix021:
Addressed in 5a471a3 (all four locales) and the PR description is updated as well:

- **`daemon_server_url` precedence now matches `daemonSetupURLsFromEnv`:** `MULTICA_PUBLIC_URL` when set, otherwise the app URL — `MULTICA_APP_URL` falling back to `FRONTEND_ORIGIN`; when that chain is empty, the field is omitted from the response entirely (so the grep prints nothing), even if `MULTICA_PUBLIC_URL` is set. Verified against `TestGetConfigUsesFrontendOriginForSameOriginDaemonSetup` and the omitempty behavior of the JSON field. The official-cloud omission branch is intentionally not documented — its host equality (`multica.ai`) cannot match a self-hosted app URL, and the doc's omission sentence stays literally true under it.
- **PR description:** the `< 2.24` claim, the unqualified `/health` statement, and the "400/401 proves the proxy chain works" wording are all replaced with the corrected versions.

Bohan-J:
Thanks for the thorough follow-through here, @felix021 — four locales kept in sync across three rounds of review, and every correction landed with the reasoning spelled out. The `daemon_server_url` precedence and its omission condition now match `daemonSetupURLsFromEnv` exactly, and the release-tag, single-origin, and WebSocket-probe notes all check out against the code.

I pushed one small commit on top (`f08722e`) rather than sending you back around again: the precedence chain starts at `MULTICA_DAEMON_SERVER_URL` (`server/internal/handler/config.go:128`) before `MULTICA_PUBLIC_URL`, so all four locales now name it. `environment-variables.mdx` already documented the full chain, so this just keeps the quickstart consistent with it.

One optional follow-up, not a blocker: the new `/api/config` check sits right after the two-domain example, where its only stated failure signal is `localhost`. In that topology, an operator who sets only `MULTICA_APP_URL` gets the **app** domain printed — which is wrong for daemons and reads as a passing check. Worth a sentence someday if you touch this page again.

Merging. Thanks again.

Bohan-J (approved):
Docs verified against the code across all four locales; all prior review points are resolved. Approving.

PR #7468, opened 2026-08-23
MUL-6661 feat(channels): add /new and /clear conversation controls

### LINKED ISSUE #7135: [Feature]: Add /new and /clear conversation controls for external channels

yyclaw:
## What does this PR do?

Adds a unified conversation-control contract across DingTalk, Feishu/Lark, Slack, Telegram, and WeCom:

- `/new` creates a new Multica Chat; `/new <message>` atomically creates it with the remaining content as its first turn.
- `/clear` resets the agent-visible context while staying in the current Chat.
- `/chat` is not shipped.

The two durable dimensions remain separate by design: route generation determines which Chat receives future external messages, while context generation bounds what an agent can see within one Chat. Queued and running tasks snapshot both where applicable.

Slack registers native `/new` and `/clear` commands because message-leading slash commands are intercepted by Slack. Slash-command names are workspace-global, so the docs call out possible collisions and retain message-style `@Multica /new` and `@Multica /clear` as alternatives.

## Shared message contract

Adapters resolve platform-specific Bot addressing before shared classification. Commands are exact, case-sensitive, token-bounded, and must lead the normalized first non-empty text line. Media does not hide a command, verified current-Bot addressing is removed without deleting other mentions, and command remainders are not reparsed. For example, `/new /issue Fix login` creates a new Chat whose literal first turn is `/issue Fix login`.

A bare `/new` creates an empty Chat without an empty message or task. A pending `/clear` cannot leak into a new Chat, and `/clear` issued after `/new` applies only to that Chat.

## Review-oriented commit split

1. `d1576caae` — `/new` + `/clear` command contract and route rotation, including DingTalk's new-Chat/session binder.
2. `c2cd56ddf` — DingTalk Bot identity resolution at ingest and dispatcher lifecycle only.
3. `77d288d69` — channel-created Chat visibility, realtime list events, and title policy.
4. `ad4deedcc` — immutable `channel_task_delivery` snapshots across all five adapters, including DingTalk outbound routing and cleanup.

Each commit is independently reviewable and matches one production-revert boundary requested in review.

## Migration and rollout

- Rebased onto current `main`; this PR owns migrations `420`–`431`, including corresponding cleanup and rollback-hook identifiers.
- Migration `420` only adds route schema. It does not run a full-table `chat_session` rewrite.
- Chat-origin metadata is backfilled by the migration `431` pre-hook in idempotent, primary-key-cursor pages of 1,000 rows, using short autocommit statements.
- `channel_task_delivery` receives its primary key by first building a unique index concurrently, then attaching the primary-key constraint with `USING INDEX`.
- Existing active channel tasks are backfilled with immutable delivery snapshots.
- Deployment follows the repository contract: migrations run before the server under `Recreate`; arbitrary old/new backend overlap is unsupported, while adjacent-release rollback is guarded.

### Release / rollout note

Daemon claims now derive external delivery metadata exclusively from the task's immutable `channel_task_delivery` snapshot, not from the Chat's current channel binding. A task started later from Web, Desktop, or Mobile in a channel-originated Chat is therefore private unless that task has an explicit external delivery snapshot; it no longer inherits channel/thread/file-delivery behavior merely because the Chat originated from a channel. This prevents replies from leaking to a rotated external route. Existing active channel tasks are backfilled during migration.

## Other implementation details

- Preserves queued/running work on its original Chat, route, context, destination, and file-delivery contract.
- Keeps previous Chats visible and usable from Web, Desktop, and Mobile.
- Reuses the existing automatic-title policy, including manual-rename precedence and attachment-only fallback.
- Resolves DingTalk Bot identity before shared parsing and preserves visible mentions when identity permission is unavailable.
- Updates channel and Slack docs in English, Simplified Chinese, Japanese, and Korean.

## How was this validated?

- Fresh PostgreSQL migration run from `001` through `431`.
- `make sqlc`.
- Race-enabled tests for the shared channel engine, all five adapters, DingTalk, handlers, services, and migration/server entry points.
- Core: 136 files / 1,624 tests.
- Mobile: 18 files / 112 tests plus iOS-run assertions.
- `pnpm typecheck` and `pnpm --filter @multica/mobile typecheck`.
- Repository and Mobile lint; only pre-existing warnings remain.
- `git diff --check` and stale command/path scans.

The full repository Go race script was also attempted. Two existing daemon DSH profile-probe tests timed out at their five-second limit; the affected-package race gate listed above passes.

## Related issue

Closes #7135

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Tests
- [ ] CI / infrastructure

## Checklist

- [x] I have run tests locally and the affected packages pass, including the race detector.
- [x] I have updated relevant documentation.
- [x] I have considered identity, permission, ordering, migration, rollout, and fail-closed risks.

## AI disclosure

**AI tool used:** Codex

**Approach:** Implemented the agreed `/new` + `/clear` contract from the shared engine outward, preserved independent context and route generations, audited cross-channel addressing and delivery snapshots, split the work into requested production-revert boundaries, and verified migrations plus affected-package race behavior.

## Screenshots

Not included. This PR does not add a visual component; the user-visible result appears through existing Chat lists and title surfaces.


Bohan-J:
## Naming: should `/new` be the command that creates a new Chat?

Before we go deeper on the implementation, there's a product question this PR surfaces that we'd like to settle with you first.

The PR draws the line as: `/new` rotates the agent-visible context inside the current Chat, `/chat` creates a new Chat and rotates the route. That split is precise, and the implementation follows it faithfully. But from inside the IM client the two commands are nearly indistinguishable: both make the agent forget, both leave you talking in the same DM or thread, and the only observable difference is whether a new entry appears in the Multica Chat list. Asking users to keep "context generation" and "Chat object" apart is asking them to learn our internal model.

The more basic problem is that `/new` reads as "new session" to most people, and "new session" implies the context is gone as a *consequence*. Today's `/new` delivers the consequence but not the thing itself — and we think that's the part that's actually wrong, not the part `/chat` is fixing.

**Proposal:**

- `/new` → what `/chat` does in this PR: create a new Multica Chat; `/new <message>` atomically creates it with that message as the first turn.
- `/clear` → what `/new` does today: reset the agent-visible context and stay in the current Chat.

`/clear` also matches what the name means in Claude Code, which is where a lot of our users' intuition comes from.

**On backward compatibility.** Redefining `/new` sounds breaking, but in practice we don't think it is. From the IM side the outcome of old `/new` and new `/new` is the same: the agent forgets, and you keep chatting in the same place. The difference only shows up inside Multica, where a conversation goes from one ever-growing Chat to several Chats split by topic — better for finding things, not a regression. So we'd rather redefine it cleanly than carry an alias plus a deprecation notice for a difference users can't perceive. If you see a case where the old and new `/new` diverge in a way a channel user would actually notice, that would change our mind.

The good news is that this PR already implements the behavior we want — it's wired to `/chat`. The change is mostly the command name plus the questions below, not a rewrite.

**Three things we'd like your view on:**

1. **Which commands get native Slack registration.** A Slack message starting with `/` is intercepted by the client and never reaches the app, which is why `/issue` had to become a real slash command (and why this PR registers `/chat`). After the rename the registered command becomes `/new`. If `/clear` also ships, it needs the same treatment or it will only work as `@Multica /clear` — worth deciding deliberately rather than inheriting. Also, Slack slash commands live in a workspace-global namespace, and `/new` is a generic enough name that collisions with other installed apps seem more likely than with `/chat`. Do you see that as a real risk?

2. **Does `/clear` ship in this PR, or later?** Worth being explicit about the cost: keeping an in-place context reset means keeping two parallel generation dimensions. `channel_chat_context_generation` already carries `history_start_message_id` / `history_end_message_id` / `history_boundary_pending`, and this PR adds the same three columns to `channel_chat_session_binding` for the route dimension, with a task snapshot on each. If `/new` creates a Chat and `/clear` is deferred, those two dimensions collapse into one and a good chunk of this PR's machinery goes away. If `/clear` ships now, the rename improves the UX but the machinery stays as-is. Both are defensible — we'd just rather choose it than arrive at it.

3. **Docs.** `/new` is currently documented only for DingTalk and Telegram, even though the shared engine parses it on every channel that reaches the Router. The rename is a good moment to document the command set once, consistently, across all five channels.

---

Unrelated to the naming, one practical heads-up so it doesn't cost you a rebase cycle: the migration numbers in this PR (`404`–`413`) now collide with `main`, where `407`–`413` are taken by `issue_source_context` and the `414`–`419` range is already in use. `main` is also ahead of this branch's base by `403`. When you renumber to `420+`, note that the numbers are hardcoded in `server/cmd/migrate/main.go` (`concurrentIndexCleanups`, `concurrentDownIndexCleanups`, and the `preRollbackHooks["413_channel_outbound_message_binding_index"]` entry), so those need to move together or the invalid-index cleanup and the rollback guard silently stop matching.

yyclaw:
Thanks for laying this out. I did consider the `/new` + `/clear` combination earlier. I chose to add `/chat` because `/new` already had existing users, and preserving its behavior avoided changing an established workflow. `/chat` also maps directly to creating a new Multica Chat, so its product meaning aligns with the Chat concept in the Web, Desktop, and Mobile clients.

That said, I am open to revisiting the naming. If you think `/new` for creating a new Multica Chat and `/clear` for resetting the agent-visible context in the current Chat is the better long-term contract, I can make that change in this PR.

If I take that direction, I would ship `/clear` in the same PR rather than defer it, so the existing reset-in-place capability is not removed. I will also handle the Slack native slash-command path and registration for both commands, including the `/new` namespace-collision consideration, and unify the command documentation across all five channels.

Separately, I will rebase onto current `main`, renumber this PR's migrations to `420+`, and update the hardcoded cleanup and rollback-hook entries together.

If that matches your preferred direction, I will proceed with the rename and simplification on that basis.

Bohan-J:
That matches our preferred direction — please go ahead with `/new` = create a new Multica Chat and `/clear` = reset the agent-visible context in the current Chat.

**On shipping `/clear` in this PR: agreed, and your reasoning is better than ours.** We had leaned toward deferring it to collapse the two generation dimensions, but you're right that deferring would *remove* a capability that exists today — and PR1 built the durable context boundary for it only weeks ago. Deferring would risk a build-delete-rebuild cycle on the same mechanism. Keep `/clear` in this PR.

One consequence worth naming explicitly so nobody expects otherwise: with `/clear` staying, both generation dimensions stay too, so this change is a naming and product-contract fix rather than a simplification. The PR stays roughly its current size.

**Which is why we'd like to ask for one thing in return: please split the work into separate commits.** At ~151 files in a single commit, "which part do we revert if something misbehaves in production" currently has no answer. Roughly one concept per commit:

1. `/new` + `/clear` command contract and the route rotation (the feature itself)
2. DingTalk Bot identity resolution at ingest + the `dispatchSlotRegistry` lifecycle rework
3. Chat visibility model (`explicitly_created_at`) + the title policy changes
4. Replacing the binding lookups with the `channel_task_delivery` snapshot across the five adapters and the daemon claim path

2 and 3 in particular are independently reviewable and independently revertable. Separate PRs would be even better if you think they can be sequenced, but commits would already be a large improvement — you split PR1 from PR2 correctly, and this is the same judgment applied one level down.

Two smaller follow-ups:

- The non-goals in #7135 now contradict this direction — it currently lists "Changing `/new` to mean 'create a new Chat'" and "Replacing `/new` with `/clear`" as explicitly out of scope, and the `/chat` acceptance criteria are written against the old naming. Could you fold that update into this change? We'll also leave a note on the issue so the decision is discoverable from there.
- When you retitle the PR after the rename, please keep the `MUL-6661` prefix so the tracking link survives.

We have a handful of remaining review notes (a behavior change in the daemon claim path that should reach the release notes, the cost of the full-table backfill in the first migration, and a primary key on `channel_task_delivery`). We'll post those once you've rebased onto `main`, so you're not chasing a moving baseline while you renumber the migrations.

yyclaw:
Implemented the agreed direction and force-pushed the rebased branch on current `main` (`617d41602`):

- `/new` now creates a new Multica Chat; `/clear` preserves reset-in-place semantics; `/chat` is not shipped.
- Slack registers native `/new` and `/clear`; the workspace-global collision risk and message-style alternatives are documented.
- The command contract is documented across DingTalk, Feishu/Lark, Slack, Telegram, and WeCom in all maintained locales.
- I split the work into the four requested review/revert boundaries:
  1. `d1576caae` command contract and route rotation
  2. `c2cd56ddf` DingTalk identity/ingest/dispatcher lifecycle only
  3. `77d288d69` Chat visibility/realtime/title policy
  4. `ad4deedcc` immutable delivery snapshots and daemon claim behavior across all five adapters
- I updated #7135's title, body, acceptance criteria, delivery plan, and non-goals to remove the old `/chat` contract and the contradictions called out in the issue comment.

I also addressed the three anticipated implementation notes:

- Added an explicit release/rollout note for the daemon claim behavior: first-party tasks no longer inherit an external destination from the Chat's current binding without a task delivery snapshot.
- Removed the full-table rewrite from migration 420; migration 431 now runs an idempotent primary-key-cursor backfill in 1,000-row autocommit pages.
- Added the `channel_task_delivery` primary key by building its unique index concurrently and attaching the constraint with `USING INDEX`.

Validation includes a fresh 001–431 migration run, affected-package race tests across the shared engine/adapters/handler/service/migrate/server, full Core and Mobile tests, typechecks, lint, SQLC, and diff/stale-command checks. The PR body records the only global-suite caveat: two existing daemon DSH profile probes hit their five-second timeout during the full repository race attempt, while the affected-package race gate passes.

Ready for the next review pass.

yyclaw:
Follow-up on the commit split: I tightened the boundaries once more after auditing the actual per-commit diffs. This is a history-only rewrite; the final tree is byte-for-byte identical to the previous pushed head.

- `d1576caae` now owns the DingTalk `StartSession`/new-Chat replier pieces that belong to the shared command feature.
- `c2cd56ddf` is reduced from 19 files to 9 and now contains only DingTalk Bot identity resolution, ingest normalization, and dispatcher lifecycle.
- `77d288d69` now owns the Chat-created/updated realtime listener changes.
- `ad4deedcc` now owns DingTalk task-delivery lookup, outbound routing, installation cleanup, and generated query changes alongside the other four adapters.

This makes the four commits match the requested review and production-revert boundaries without changing runtime behavior.

Bohan-J:
Thanks for turning this around so thoroughly — merging.

Two of your solutions were better than what we asked for, and worth naming:

- The chat-origin backfill. We asked you to get the full-table rewrite out of the downtime window; you replaced it with a primary-key-cursor hook that pages in short autocommit statements, guards on `explicitly_created_at IS NULL`, and resumes from the cursor if a run dies. That's idempotent and restartable, not just smaller.
- The `channel_task_delivery` primary key. Building the unique index concurrently and promoting it with `ADD CONSTRAINT ... PRIMARY KEY USING INDEX` avoids the index build lock entirely, and registering the invalid-index cleanup hook for `423` closes the failure path too.

Verified on `ad4deedcc`: full `001`–`431` migration run on a clean database, `go build ./...`, and the shared engine, all five adapters, `service`, `handler`, `migrations`, `cmd/migrate`, and `cmd/server` test packages. All 16 CI checks are green, including the frontend and mobile suites.

The renaming came out clean — the shared parser is a straight constant swap with the semantics preserved, and there is no `/chat` residue left in Go, SQL, docs, or TypeScript. Thanks also for updating #7135 rather than leaving the spec contradicting the shipped behavior.

A few non-blocking follow-ups for whoever picks them up next — none worth holding this release for:

- `slashNewThreadGuideText` and `slashClearThreadGuideText` hardcode `@Multica`, but the setup docs tell users to rename the Bot. These should use the configured Bot or agent name.
- In `HandleEnvelope`, `/issue` is matched with `strings.EqualFold` while `/new` and `/clear` use exact comparison. Worth making consistent.
- `MarkChatSessionExplicitlyCreated` runs as a second UPDATE after `CreateChatSession` at four call sites. Folding it into the create parameters would remove the chance that a future creation path forgets it and produces a Chat that never appears in the list.
- The `binding.pending_fresh` queries no longer have a reader now that the generation row owns that state. Worth dropping the column and the three queries, or noting why they stay.

One note on the commit split, since the PR body describes the four commits as production-revert boundaries: they are genuinely good *review* boundaries — the DingTalk commit dropping to 9 files made that part reviewable on its own — but they are not revert boundaries, because `service.PreparedChatTaskEnqueue` is defined in `ad4deedcc` and already referenced from `d1576caae`, so the first three commits do not compile in isolation. This repo squash-merges, so nothing reaches `main` history and it changes nothing here. Flagging it only so the claim isn't carried into a future PR where the sequence might matter: putting the shared service-layer types in the first commit would make the chain buildable end to end.

Nice work on this one.

PR #7474, opened 2026-08-23
fix(execenv): write .gc_meta.json / .managed_env.json atomically

### LINKED ISSUE #7469: execenv: write .gc_meta.json / .managed_env.json atomically
Both metadata writers use plain `os.WriteFile`:

- `WriteGCMeta` (`server/internal/daemon/execenv/execenv.go:960`)
- `WriteManagedEnvProvenance` (`:1029`)

A crash mid-write leaves torn JSON. Consequences:
- Torn `.gc_meta.json` fails `ReadGCMeta` unmarshal → treated as no-meta → dir falls into the 72h orphan-by-mtime path instead of parent-driven cleanup.
- Torn provenance file fails closed on read → reuse declined → session continuity silently dropped (the MUL-4886 bug class).

The in-tree helper `writeFileAtomic` (`hermes_home.go:913`, tmp+rename) already exists and is used elsewhere; these two writers just predate it.

Also: when `WriteGCMeta` fails at finalize time the daemon only Warns (`daemon.go`), so a completed task's env root drops to the 72h path on a transient error. One immediate retry before giving up turns that into a non-event.

PR incoming.

scotthawes:
Closes #7469

Route both metadata writers through the existing `writeFileAtomic` (tmp+rename) helper instead of plain `os.WriteFile`, and add one immediate retry when the finalize-time `.gc_meta.json` write fails (previously Warn-only, silently dropping the dir into the 72h orphan-by-mtime path).

- Torn provenance file = fail-closed on read → session continuity dropped (MUL-4886 bug class) — now impossible.
- Torn GC meta = unmarshal error → treated as no-meta → 72h orphan path — now impossible.
- Retry escalation turns transient write failures into a non-event with an Error log if it still fails.

No behavior change on the happy path; same file modes.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5215,6 +5215,12 @@ func (d *Daemon) handleTask(ctx context.Context, task Task, slot int) {
 			}
 			if err := execenv.WriteGCMeta(result.EnvRoot, meta, taskLog); err != nil {
 				taskLog.Warn("write gc meta failed (non-fatal)", "error", err)
+				// One immediate retry: a dir that silently loses its meta
+				// falls to the 72h orphan-by-mtime path instead of
+				// parent-driven cleanup, pinning disk for days.
+				if retryErr := execenv.WriteGCMeta(result.EnvRoot, meta, taskLog); retryErr != nil {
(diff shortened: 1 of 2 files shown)


scotthawes:

PR #7476, opened 2026-08-23
feat: jitter retry schedules to prevent thundering herd

### LINKED ISSUE #7471: jitter retry schedules to prevent thundering herd on provider recovery
Both retry schedules are deterministic:

- Daemon client: fixed 4/8/16/32/64s backoff, no jitter (`client.go:895-901`).
- Server deferred retries: `retryDelayForAttempt` returns exactly `providerNetworkFinalRetryWait`.

When a provider blip watchdog-kills a wave of tasks simultaneously, their retries fire in lockstep — and re-cluster onto the upstream exactly as it recovers. Same shape fleet-side for deferred children created in the same FailTask transaction burst.

Proposed fix (incoming PR): ±20% jitter in `retrySleep` and in the computed `fire_at`; adjust the schedule test to assert a window.

scotthawes:
Closes #7471

Both retry schedules were deterministic — daemon client backoff (4/8/16/32/64s) and the server-side deferred-retry `fire_at`. Tasks watchdog-killed by a provider blip re-fire in lockstep and re-cluster onto the upstream exactly as it recovers.

Adds ±20% jitter to:
- `retrySleep` (client.go)
- `retryDelayForAttempt`'s provider_network final-tier deferral (keeps the new runtime_offline health-gated deferral untouched)

Test updated to assert a [4s, 6s] window for the jittered tier instead of an exact value; runtime_offline case keeps its exact assertion.

### DIFF
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"runtime"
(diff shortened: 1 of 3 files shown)


scotthawes:

PR #7477, opened 2026-08-23
perf(db): bound claim-candidate scan with LIMIT

### LINKED ISSUE #7472: claim candidates: LIMIT the batch candidate SELECT
`ListQueuedClaimCandidatesByRuntimes` (`queries/agent.sql:1613`) is `SELECT * ... ORDER BY priority DESC, created_at ASC` with no LIMIT. Every non-empty daemon poll sorts its runtimes' entire queued set. The empty-claim cache (TTL 3min) protects idle daemons only — a saturated queue means full scan+sort per daemon per poll, plus every WS nudge.

The batch claim loop stops at `maxTasks` anyway (`task.go:3345`), so fetching more candidates than remaining capacity buys nothing.

Proposed fix (incoming PR): `LIMIT @max_per_poll` bound to remaining claim capacity (`maxTasks - already claimed`). The ORDER BY guarantees the kept head is exactly the priority/FIFO slice the loop would have consumed — fairness unchanged.

scotthawes:
Closes #7472

`ListQueuedClaimCandidatesByRuntimes` had no LIMIT — every non-empty daemon poll sorted its runtimes' entire queued set. Empty-claim cache protects idle daemons only; a saturated queue means full scan+sort per daemon per poll plus every WS nudge.

Change: `LIMIT @max_per_poll` bound to remaining claim capacity (`maxTasks - already claimed this batch`). The existing ORDER BY (priority DESC, created_at ASC) guarantees the kept head is exactly the priority/FIFO slice the claim loop would have consumed (`task.go:3345` stops at maxTasks anyway) — fairness unchanged, scan bounded.

sqlc regenerated.

### DIFF
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -3555,8 +3555,18 @@ func (s *TaskService) ClaimTasksForRuntimes(ctx context.Context, runtimeIDs []pg
 		return claimed, nil
 	}
 
-	// 4. One candidate SELECT across the non-empty set.
-	candidates, err := s.Queries.ListQueuedClaimCandidatesByRuntimes(ctx, nonEmpty)
+	// 4. One candidate SELECT across the non-empty set. Bounded by the
+	// remaining claim capacity: the loop below stops at maxTasks anyway, so
+	// fetching more candidates than that only pays scan+sort cost (the LIMIT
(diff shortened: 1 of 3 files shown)


scotthawes:

PR #7478, opened 2026-08-23
feat(service): auto-retry webhook-triggered autopilot runs

### LINKED ISSUE #7473: auto-retry webhook-triggered autopilot runs (cron stays excluded)
`retryEligible` (`service/task.go:4364`) excludes ALL autopilot runs:

```go
!t.AutopilotRunID.Valid &&
```

Cron runs self-heal (scheduler re-fires next tick). Webhook runs don't: the external poster fired once and is gone. A transient `runtime_offline` / `provider_network` failure therefore kills the run permanently with its trigger payload stranded — `SyncRunFromTask` just records the failure, no human ever re-fires it.

Proposed fix (incoming PR):
- Split eligibility into `retryEligibleForSource(reason, task, autopilotRunSource)`.
- Cron-sourced runs stay excluded (scheduler owns cadence; a retry would double-fire the schedule).
- Webhook-sourced runs become eligible through the existing reason-aware ceilings — trigger payload already persists on the run row (`autopilot.go:1631`).
- Unknown/unreadable run source fails closed to today's behavior.

Happy to hear counterpoints on webhook retry semantics if the team prefers a different recovery path.

scotthawes:
Closes #7473

`retryEligible` excluded ALL autopilot runs. Cron runs self-heal on the next tick; webhook runs do not — the external poster fired once and is gone, so a transient `runtime_offline`/`provider_network` failure killed the run permanently with its trigger payload stranded.

**Change:**
- Split into `retryEligibleForSource(reason, task, autopilotRunSource)`; plain `retryEligible` delegates with "" (behavior unchanged for non-autopilot tasks).
- Cron-sourced runs stay excluded — scheduler owns cadence, a retry would double-fire.
- Webhook-sourced runs become eligible via the existing reason-aware ceilings; trigger payload persists on the run row.
- New `autopilotRunRetrySource` resolves source once per eligibility check; unreadable run fails closed to "cron" (legacy behavior).
- Both retry paths (FailTask in-tx + orphan sweeper) stay in sync.

Rebased on current main including the new `hasRunnableSuccessor` guard — unchanged and composes fine (webhook retries now also respect the single-pending-task slot).

### DIFF
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -4357,7 +4357,7 @@ func (s *TaskService) FailTask(ctx context.Context, taskID pgtype.UUID, errMsg, 
 		if parent, perr := s.Queries.GetAgentTask(ctx, taskID); perr != nil {
 			slog.Warn("fail task auto-retry: load parent failed",
 				"task_id", util.UUIDToString(taskID), "error", perr)
-		} else if retryEligible(failureReason, parent) {
+		} else if retryEligibleForSource(failureReason, parent, s.autopilotRunRetrySource(ctx, parent)) {
 			wantRetry = true
 			// Persist the reason-aware effective budget into the child so the
 			// retry chain self-describes (e.g. provider_network → max_attempts=3),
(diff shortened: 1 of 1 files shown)


scotthawes:

PR #7486, opened 2026-08-24
feat(auth): DingTalk OAuth login via enterprise email

felix021:
## Summary

DingTalk OAuth login for self-hosted Multica. Members of the DingTalk organization that owns the app sign in by scanning a QR code; the login identity is exactly one value — the enterprise mailbox (`org_email`, 企业邮箱) on that organization's member record.

### Email resolution (verified against the live API)

```
authCode → user access token → GET /v1.0/contact/users/me
                                    (nickname / avatar / unionId only)
        → app access token  → getbyunionid → topapi/v2/user/get
                                    (the app's own org member record)
        → org_email → findOrCreateUser(email) → session
```

- **`me` is never used as an email source.** Its consent response carries account-level data that crosses org boundaries — with the app registered under an unrelated 1-member personal org, `me` still returned the user's company-org enterprise email, plus the mobile number in plaintext. Using it would let an app under any personal org log users in via an address that org never assigned, and for accounts with enterprise emails in more than one org the selection is undocumented and can drift.
- **No fallback, by design.** A user who is not a member of the app's org, a member whose record has no `org_email`, or an app missing the required permissions is rejected with 403. There is no other email source.
- Verified end-to-end on live deployments: a personal-org app rejects logins regardless of which identity the user picks on DingTalk's consent screen; an app inside the org resolves the deterministic `org_email` for its members either way. The gate is the app's credentials, not the consent-screen choice.

### Requirements

- The app must be an 企业内部应用 inside the organization whose members should log in, and that org must assign enterprise mailboxes. An app under a personal org cannot work (documented in the guide).
- Permissions, both required: **成员信息读** (`qyapi_get_member`) and the sensitive **企业员工手机号信息和邮箱等个人信息**. Without them `getbyunionid`/`v2/user/get` answer errcode 88 / subcode 60011, or the member record comes back with the email fields omitted entirely.
- Env, passed through `docker-compose.selfhost.yml`: `DINGTALK_CLIENT_ID` / `DINGTALK_CLIENT_SECRET` (login), `MULTICA_DINGTALK_SECRET_KEY` (bot integration, opt-in — encrypts per-workspace BYO bot secrets at rest).

## Implementation

- server: `POST /auth/dingtalk` handler (`server/internal/handler/dingtalk_login.go`), `dingtalk_client_id` exposed in the public `/api/config` only when configured, route registered behind the auth rate limiter
- web: login-page button + authorize URL (scope `openid corpid`, `%20`-encoded), callback page guarded against `authCode` replay (effect re-runs and Suspense remounts no longer re-POST a consumed code)
- docs: `DINGTALK_OAUTH.md` + `DINGTALK_OAUTH.zh.md` — setup, required permissions, email semantics (including why `me.email` is not used), troubleshooting

## Tests

- Existing DingTalk handler tests stay green (`go test ./internal/handler -run DingTalk`).
- Full flow verified manually against the live API and two live deployments (reject/allow matrix above); the resolution chain has no unit coverage yet — the DingTalk endpoints would need base-URL injection to stub, left as follow-up.


felix021:
## Design change: single-source enterprise email (rewritten history, one commit)

The original implementation resolved the login email as `me.email` first, with a server-side `org_email` fallback, and described the result as "the enterprise email". Live-API probing on a real deployment showed that framing was wrong in three ways, so the branch was rewritten around a single-source design.

### Problems with the original approach (all verified against the live API)

1. **`me.email` is account-level data that crosses org boundaries.** With the app registered under an unrelated 1-member *personal* org, `/contact/users/me` still returned the user's **company-org** enterprise email — and the mobile number in plaintext. Using it as the login key meant an app under any personal org could log users in via an address that org never assigned.
2. **The server-side fallback was dead code.** The endpoint was misspelled `getByUnionid` (camelCase); DingTalk answers errcode 22 `Invalid method`. The correct spelling is all-lowercase `getbyunionid`.
3. **Identity-key instability.** For accounts with enterprise emails in more than one org, which address `me` returns is undocumented (plausibly the primary org) and can drift — a changed email would provision a second account for the same person, since login matches users by email. The personal-profile email slot is not a usable escape hatch either: it needs the same sensitive permission, and unverified personal orgs cannot even set the field in the client.

### The new design

The login email is exactly one value: the **`org_email` (企业邮箱) on the app-org member record**, resolved server-side (`app token → getbyunionid → v2/user/get`). `me` supplies only nickname/avatar/unionId. A user outside the app's org, a member without an enterprise mailbox, or an app missing the permissions (成员信息读 `qyapi_get_member` + the sensitive 企业员工手机号信息和邮箱等个人信息) is rejected with 403 — **no fallback email source, by design**.

### Verification matrix (live-tested on two deployments)

| App registration | Consent-screen identity | Result |
|---|---|---|
| personal org (no enterprise mail) | personal org / company org | 403 rejected both ways |
| company org (permissions granted) | personal org / company org | deterministic `org_email` login both ways |

The gate is the app's own credentials and org membership, not the consent-screen choice — the org picker on DingTalk's consent page only affects presentation.

PR #7489, opened 2026-08-24
MUL-6622: fix(server): authorize squad activity by task provenance

### LINKED ISSUE #7487: [Bug]: `squad activity` is rejected on child issues, silently dropping the leader's no_action decisions
### Deployment type

Official App

### Area

CLI

### What happened?

`multica squad activity <issue-id> <outcome>` rejects any issue that is not **itself** assigned to a squad:

```
$ multica squad activity 01a02ba0-7eaa-76a3-9b58-d3e1db42f776 no_action --reason "..."
Invalid request: issue is not assigned to a squad
$ echo $?
5
```

But the squad-leader workflow spends most of its time on issues that are *not* squad-assigned. The runtime instructions injected into a leader agent say, verbatim:

> **Record your evaluation.** After every trigger — whether you delegated, decided no action is needed, or encountered an error — record it: `multica squad activity <issue-id> <outcome> --reason "<short reason>"` […] This is mandatory on every turn.

and, for the `no_action` case:

> If your evaluation outcome is `no_action`, call `multica squad activity <issue-id> no_action --reason "..."` and EXIT. **DO NOT post any comment** […] The squad activity call records your decision in the timeline.

Two ways a leader legitimately ends up on a non-squad-assigned issue:

1. **Child issues it created itself.** The standard fan-out is `issue create --parent <parent> --assignee-id <member> --stage N`. The parent carries `assignee_type: squad`; every child carries `assignee_type: agent`. The leader is then re-triggered on those children (stage barriers, member reports, `@squad` mentions) and every `squad activity` call on them fails.
2. **`@squad` mentions on someone else's issue.** A human @-mentions the squad on an issue owned by an individual agent. The mention wakes the leader — the same runtime block that mandates `squad activity` fires — and the call is rejected.

The `no_action` path is the damaging one, because the instruction makes `squad activity` the *substitute* for a comment. The leader is told not to comment, the recording call errors out, and the decision leaves **no trace anywhere** — not in the timeline, not in the comments. It fails silently from the operator's point of view: nothing in the UI indicates the leader evaluated the trigger at all.

In our workspace this happened on five consecutive turns on the same issue, all with the identical error.

### Steps to reproduce

1. Create an issue and assign it to a squad (`assignee_type: squad`). Confirm the leader is woken.
2. As the leader, create a child issue assigned to an individual member:
   `multica issue create --parent <parent-id> --assignee-id <member-agent-id> --stage 1 --status todo --title "child"`
3. Get re-triggered on that child (a member comment, an `@squad` mention on the child, or a stage barrier).
4. Run `multica squad activity <child-issue-id> no_action --reason "nothing to do"`.
5. → `Invalid request: issue is not assigned to a squad`, exit code 5. Nothing is recorded, and the instructions forbid posting a comment instead.

Contrast: the same command against the **parent** (`assignee_type: squad`) is accepted. The validation keys off the target issue's own assignee, not off whether a squad is involved in the issue at all.

```shell
multica v0.4.29 (commit: c670e0549, built: 2026-08-18T10:09:14Z)
go: go1.26.6, os/arch: darwin/arm64

# child issue, assignee_type=agent  → rejected
$ multica squad activity 01a02ba0-7eaa-76a3-9b58-d3e1db42f776 action --reason "probe"
Invalid request: issue is not assigned to a squad   (exit 5)

# same error with --output json; the message goes to stderr and stdout stays empty
```

Suggested directions, in rough order of how much we'd want them:

- **Accept the call when the squad is *involved* in the issue**, not only when it owns it: the issue's parent is assigned to the caller's squad, or the squad / its leader was @-mentioned on the issue. That matches the set of issues a leader is actually woken on.
- **Failing that, degrade to a no-op with a warning instead of a hard error.** As written, the instruction "ALWAYS call `squad activity` before ending your turn" is unsatisfiable on the majority of a leader's triggers, which teaches agents to ignore it.
- If the restriction is deliberate, the injected runtime instructions should say so and give leaders a sanctioned way to record a `no_action` decision on a child issue — otherwise the current wording actively converts those decisions into silence.

Related but distinct: #4101 asks for server-side telemetry on the *automatic* routing decision. This one is about the leader's own explicit self-report being refused on the issues where it is most needed.

multica-eve:
## Background

`multica squad activity <issue-id> <outcome>` rejected any issue that was not **itself** assigned to a squad:

```
Invalid request: issue is not assigned to a squad   (exit 5)
```

But squad leaders spend most of their turns on issues that are not squad-assigned — a `@squad` mention on an issue owned by an individual agent, or a leader task bound to a child issue the leader created. The injected runtime instructions make that call **mandatory on every turn**, and for `no_action` they make it the *substitute* for a comment. So the leader was told not to comment, the recording call errored, and the decision left no trace anywhere.

The root cause is two sources of truth for "is this run a squad leader turn":

- the claim path keys the squad briefing and the mandatory-recording instruction off the **task row** (`is_leader_task` + `squad_id`) — see the comment at `handler/daemon.go` and `taskIsSquadLeader` in `daemon/prompt.go`;
- `RecordSquadLeaderEvaluation` instead hard-checked `issue.assignee_type == "squad"` and read the squad from `issue.assignee_id`, reading `X-Task-ID` only afterwards and only to check the task belonged to the issue.

MUL-3730 / #4606 migrated the briefing decision to task provenance; this endpoint was not migrated with it.

## Changes

Server-side only. No CLI arguments, request/response payload, or schema changes.

`server/internal/handler/squad.go` — `RecordSquadLeaderEvaluation`:

- load the `X-Task-ID` row **first**; require `task.issue_id == issue.id`, `task.is_leader_task`, and a valid `task.squad_id`. The squad is resolved from `task.squad_id` scoped to `issue.workspace_id`;
- the `issue.assignee_type == "squad"` precondition is gone;
- authorization is `task.agent_id == caller`, not the live `squad.leader_id`. The task row is the provenance — it records that this agent was enqueued as leader of this squad — while `squad.leader_id` is mutable state that can rotate mid-run;
- `activity_log.actor_id` is now `task.agent_id` for the same reason. This was also a latent bug: the `no_action` comment-suppression query matches `actor_id` against `task.agent_id` (`service/squad_no_action.go`), so writing the live leader id there meant a leader rotation silently disabled suppression **and** misattributed the audit record;
- the cross-issue rejection now names the issue the caller should have used. A leader woken by a stage barrier / child-done callback runs on the **parent** (`issue_child_done.go` enqueues on `parent`), so reaching for the child id is the common slip; the error now points at the right issue instead of dead-ending;
- a leader task with no stamped `squad_id` (possible on pre-MUL-3730 rows) is rejected with a `slog.Warn`, not a silent 400.

Injected instructions and docs — the silent-drop path had a second half. The comment rejection in `comment.go` is **conditional** on the `no_action` activity already existing, so a leader whose recording call failed was always allowed to comment; it simply was never told so. Now `daemon/prompt.go`, `execenv/runtime_config_sections.go` (both no_action blocks) and `handler/squad_briefing.go` state that the prohibition lapses when the call fails, and that the leader should post a short comment instead. `cmd_squad.go` help, the builtin `multica-squads` SKILL.md and its source map document which issue the command accepts and why.

## Behavior narrowing (deliberate)

A leader agent running a **non**-leader task on a squad-assigned issue (e.g. a same-squad worker task) used to be accepted here and is now rejected — it is not running as the leader, and the runtime only mandates the call for `is_leader_task`. Pinned by `TestRecordSquadLeaderEvaluation_RejectsNonLeaderTask` so it stays an explicit decision rather than a side effect.

## Testing

New `server/internal/handler/squad_evaluation_provenance_test.go`:

- `AcceptedOnNonSquadAssignedIssue` — the regression: leader task on an agent-owned issue records, with `actor_id` = task agent and `details.squad_id` = task squad;
- `AcceptedOnChildIssueBoundTask` — a leader task bound to a child issue records;
- `RejectsNonLeaderTask`, `RejectsLeaderTaskWithoutSquadID`, `RejectsForeignAgent`;
- `RejectsCrossIssueTaskAndNamesTaskIssue` — asserts the error names the task's issue;
- `SurvivesLeaderChangeAndKeepsNoActionSuppression` — rotates the squad leader after enqueue, then asserts the record still lands under the task's agent **and** that the `no_action` comment stays suppressed.

The shared `runningSquadLeaderTaskFixture` now stamps `is_leader_task` / `squad_id`, so the existing squad-assigned path keeps its coverage.

Local runs against a freshly migrated isolated database:

```
go test ./internal/handler ./internal/service -count=1
ok  github.com/multica-ai/multica/server/internal/handler  72.175s
ok  github.com/multica-ai/multica/server/internal/service  13.132s
```

`go build ./...` clean. Pre-existing, unrelated failures in `internal/daemon`, `internal/daemon/execenv` and `cmd/multica` (HOME/profile-path and CLI config tests) were confirmed identical on an unmodified checkout — `cmd/multica` fails 124 tests both with and without this change.

## Review focus

- the authorization swap from `squad.leader_id` to `task.agent_id` — it is what makes the leader-rotation case work, and it is the security-relevant line;
- whether the cross-issue restriction should stay strict. This PR keeps "record on the issue your task is running on" and improves the error; the alternative is widening to the parent/child chain of the same squad, which I deliberately did not do since the leader is genuinely running on the parent in that flow.

Multica issue: MUL-6622

Fixes #7487


multica-eve:
Pushed `8587f4c86` addressing all four review items.

**1. [P1 security] Tenant leak fixed.** The task is now loaded with `GetAgentTaskInWorkspace` (scoped to `issue.workspace_id`) instead of the global `GetAgentTask`, and the check order is made explicit and load-bearing: tenant scope → `task.agent_id == caller` → only then any message that names a task-derived id. The cross-issue rejection still names the task's issue, but only past the ownership gate. Two negative tests assert the response body does **not** contain the other issue's id: `RejectsForeignWorkspaceTaskWithoutLeakingItsIssue` and `RejectsForeignAgentWithoutLeakingTaskIssue`.

**2. [P1 authz] Live-leader gate restored.** You are right that the row records enqueue-time *intent*, not the delivered role — the claim path clears `resp.IsLeaderTask` on a pre-claim leader swap while `is_leader_task` stays true in the DB. I took the minimal safe fix rather than persisting the delivered role: `squad.leader_id == caller` is checked again, on top of the task-ownership gate. `activity_log.actor_id` stays `task.agent_id`, since that is what the `no_action` suppression lookup matches on. The test that locked the wrong behavior is replaced by `RejectsAfterLeaderChange`, which asserts 403 and no activity row. The code comment records what removing the gate would require (persisting the claim-delivered role), so the next person does not have to re-derive it.

Note the residual cost is now much smaller than when I argued the other way in review: a leader rotated away mid-run is refused here, but item 3's fallback means the decision still leaves a comment rather than vanishing.

**3. [P1 behavior] Fallback no longer collides with one-comment-per-turn.** All four injection sites now scope the fallback to turns that have not already commented, and cap it at one comment — on the `action` path the delegation comment *is* the record. Two tests lock the wording: `TestSquadOperatingProtocolRecordsAgainstTheTurnsIssue` (briefing, both `ownsIssueStatus` variants) and `TestBuildPromptSquadLeaderNoActionFailureFallback` (per-turn prompt).

**4. [P2 repo rule] Tests rebuilt on the shared fixtures.** `squad_evaluation_provenance_test.go` now uses `dbfx.Issue` / `dbfx.Task` / `dbfx.Squad` / `dbfx.User` / `dbfx.Workspace` and `testutil.Call(...).Want(...).Text()`; no open-coded `INSERT ... RETURNING` or recorder/status/decode quartets remain. I also moved the pre-existing `runningSquadLeaderTaskFixture` in `squad_no_action_test.go` over, since I was already editing that insert.

Docs updated to match: the `multica-squads` SKILL.md and source map now describe both gates, the tenant-scoped load, why `actor_id` is `task.agent_id`, and the conditional fallback.

**Verification** (freshly migrated isolated DB): `go test ./internal/handler ./internal/service -count=1` → both `ok` (47.3s / 8.7s); `go test ./internal/daemon -run 'Prompt|Squad'` → `ok`; `go test ./internal/daemon/execenv -run 'ByteIdentical|RuntimeConfig|Brief'` → `ok`; `go build ./...`, `go vet ./...`, `gofmt` and `git diff --check` clean. All 8 provenance tests pass individually.

PR #7493, opened 2026-08-24
TIM-185: rework #7492 PR — reserved headers, httpguts, bounds, cross-origin safety, FlagOrEnvArray removal (MUL-6636 items 1, 2, 3, 5)

### LINKED ISSUE #7492: [Feature]: support attaching custom HTTP headers to outbound server requests (self-hosted Multica behind ZTN / reverse-proxy auth)
### Deployment type

self-host

### Area

CLI

## Summary

Self-hosted Multica operators frequently need to place the daemon behind an authenticating reverse proxy or Zero-Trust Network (ZTN) — Cloudflare Access, Tailscale-fronted `oauth2-proxy`, Authelia, Authentik, Pomerium, Keycloak, Ory Oathkeeper, Traefik `forwardAuth`, Caddy `forward_auth`, etc. These solutions authenticate the client by injecting one or more HTTP headers on the upstream side (e.g. `CF-Access-Client-Id` / `CF-Access-Client-Secret` for a Cloudflare Access Service Token).

Today, the Multica daemon has no way to attach an arbitrary header to its outbound server calls. It only sends its own bearer token + identity headers (`X-Client-Platform/Version/OS`, `X-Client-Capabilities`). As a result, the daemon cannot authenticate to a Multica server protected by any of the solutions above without a workaround.

## Use cases

The motivating case is Cloudflare Access protecting a self-hosted Multica server: the daemon must present a Service Token's `CF-Access-Client-Id` + `CF-Access-Client-Secret` on every HTTP request and every wakeup WebSocket. The same shape applies to any other ZTN or reverse-proxy that uses injected headers as the auth signal.

The list is long enough that we should not single-vendor this:

- **Cloudflare Access** — Service Token `CF-Access-Client-Id` / `CF-Access-Client-Secret`
- **Tailscale** (reverse-proxy-with-identity pattern) — `oauth2-proxy` with the Tailscale OIDC provider, injecting `X-Forwarded-User` / `X-Forwarded-Email`
- **Authelia** — `Remote-User` / `Remote-Groups` / `Remote-Name` / `Remote-Email`
- **Authentik** — `X-Authentik-Username` / `X-Authentik-Groups` / `X-Authentik-Email`
- **oauth2-proxy** — `X-Forwarded-User` / `X-Forwarded-Email` / `X-Forwarded-Groups`
- **Pomerium** — `X-Pomerium-Claim-User` / `X-Pomerium-Claim-Email` / `X-Pomerium-Jwt`
- **Keycloak gatekeeper** — `X-Forwarded-User` / `X-Forwarded-Preferred-Username` / `X-Forwarded-Email`
- **Ory Oathkeeper** — mutator-config-defined headers
- **Traefik `forwardAuth`** / **Caddy `forward_auth`** — operator-defined header names

## Current workaround

Self-hosters currently run a local sidecar proxy (most commonly Traefik) in front of the Multica daemon that injects the auth header. This is fragile, adds a layer of operational complexity, and makes systemd / container deployments strictly harder than necessary.

## Proposed feature

Let the daemon attach a configured set of HTTP headers to every outbound server call (both HTTP and the wakeup WebSocket). Configuration follows the daemon's existing three-tier precedence (CLI flag > env var > config file > unset). The header source must be provider-neutral: the daemon accepts arbitrary `Name: value` pairs, validates them at parse time, and attaches them via `http.Header.Add` so multi-value headers survive.

The proposed scope is **the daemon → server control plane only**. The feature should NOT bleed into MCP / agent-spawn traffic — those go through separate code paths with their own secrets (e.g. the per-runtime `http_headers` block for an MCP server), and a different security model applies there.

## Security considerations

- Parse-time rejection of CR / LF / NUL in name and value, and `:` in name. Hard error at daemon startup, not at first request — a typo should fail the systemd unit, not the next heartbeat.
- No header values in logs. The existing `daemon.Client.requestError` already truncates response bodies and never echoes request headers; this contract must extend to any new code path.
- Persisted config (`~/.multica/config.json`) keeps the existing 0600 perms via `SaveCLIConfig`. Operators who don't want a token on disk use the env-var path.

## Out of scope

- Auto-merge of global headers into per-server MCP `http_headers` for MCP servers behind the same auth realm. Separate concern, different security model.
- Any header injection on agent subprocess → LLM provider, agent → MCP, or daemon → git remote. Those paths have their own auth models and should remain unchanged.

## Implementation draft

There is a draft implementation branch ready for review on this fork that lands the feature in three commits (parser + config, transport wiring on HTTP and WS, and the operator/contributor docs). The PR will link this issue once opened.

## Alternatives considered

- **Bake a per-vendor config block** (e.g. `cloudflare_access: { ... }`). Rejected: the feature is provider-neutral (any `Name: value` works), so a generic headers list is both smaller and more future-proof.
- **Extend the existing bearer-token auth model** to accept tokens from a sidecar header file. Rejected: the auth realm for the daemon → server control plane is bearer; the proxy/ZTN auth is a different concern and conflating them weakens both.

_No response_

timoa:
Closes #7492
Closes TIM-185

Refs TIM-184 (MUL-6636 PR review rework). This PR addresses review items **1, 2, 3, and 5** from the Multica maintainer's review on #7492, plus the "one flag syntax" cleanup folded into the same diff. Items 4 (headless workflow extension) and 6 (docs correctness) ship in separate sub-issues (TIM-151 and TIM-150).

## Item 1 — Reserved / Multica-owned headers rejected at parse time

`server/internal/daemon/headers.go` adds a `reservedHeaders` map and `reservedHeaderPrefixes` slice covering every X-Client-*, every X-Forwarded-*, and every Sec-Websocket-* variant. Operators who try to set Authorization, Content-Type, X-Client-Platform, or any forwarded / WebSocket header via MULTICA_EXTRA_HEADERS / extra_headers / --extra-header now fail daemon startup with a "reserved (managed by Multica)" error — instead of producing a request that carries both a Multica-managed Authorization and an operator-injected `Bearer proxy-token`. Defence-in-depth skip in `setIdentityHeaders` and `buildTaskWakeupHeaders` keeps the wire clean even if a reserved header ever lands in `c.extraHeaders` via a path that bypassed the parser. Tests for HTTP and WS cover every reserved name.

## Item 2 — Cross-origin credential safety

`daemon.Client` installs a `CheckRedirect` on both the control-plane and skill-bundle `http.Client` instances. The handler compares the redirect target's origin (scheme + host + port, captured from `baseURL` in `NewClient`) against the configured origin and strips every configured extra header before following on any change. Go's stdlib only strips Authorization / Cookie / WWW-Authenticate on cross-host redirects — a 302 to a different host would still leak an operator-configured `Cf-Access-Client-Secret`. Tests cover host change, scheme change (http ↔ https), port change, multi-hop ("strip once stay stripped"), and the same-origin positive control. The WS path is unaffected: `gorilla/websocket.Dialer` does not follow redirects.

## Item 3 — Complete HTTP header validation at startup + bounds

`ValidateHeaderNameValue` now layers three independent checks: explicit CR/LF/NUL rejection (clearer per-byte errors), `httpguts.ValidHeaderFieldName` / `ValidHeaderFieldValue` from `golang.org/x/net/http/httpguts` (already an indirect dep), and the reserved-header blocklist. `MaxExtraHeaders = 64` and `MaxExtraHeadersBytes = 16 KiB` surface as package constants, enforced incrementally inside `ExtraHeadersFromSpec`, the CLI flag loop in `cmd_daemon.go`, and the override / config-file branches of `resolveExtraHeaders`. Negative tests cover httpguts-passing-but-invalid names like `X Bad Name` (the documented fail-fast-at-startup contract) and the byte / count bounds.

## Item 5 — Remove `FlagOrEnvArray`

The `--extra-header` flag is wired exclusively through `runDaemonForeground` in `cmd_daemon.go`; the standalone `FlagOrEnvArray` helper in `server/internal/cli` had no callers and is deleted along with its dedicated `flags_test.go`. The comment-only reference at `cmd_daemon.go:runDaemonForeground` is updated to reflect the new state.

## Non-blocking cleanup folded in here

- **One documented flag syntax.** `ExtraHeaderFromFlag` accepts only the documented `Name: Value` form; `Name=value` errors with "expected 'Name: Value'". The flag help text for both `daemon start` and `daemon restart` is updated. Docs update ships separately in TIM-150.
- **Multi-value vs `map[string]string`.** Documented in `resolveExtraHeaders`: env-var and `--extra-header` paths support multi-value (Header.Add); the config-file path is documented as "last-write-wins" because JSON unmarshal collapses duplicate keys. Expanding to `map[string][]string` would be a breaking change for every existing config file; deferred.
- **Concurrency.** `SetExtraHeaders` / `ExtraHeaders` now carry an explicit "startup-only" contract in their doc comments (it was implicit before — the only callers in the daemon follow it).

## Verification

```bash
cd server
go vet ./...                                    # clean
gofmt -l internal/daemon/{headers,client,wakeup}{,_test}.go \
        internal/cli/flags.go \
        cmd/multica/cmd_daemon.go                # clean
go test ./internal/daemon/ \
  -run 'TestExtraHeaderFromFlag|TestExtraHeadersFromSpec| \
        TestValidateHeaderNameValue|TestResolveExtraHeaders| \
        TestIsReservedHeader|TestClient_Redirect| \
        TestClient_SetIdentityHeaders_SkipsReservedFromExtra| \
        TestClient_ExtraHeaders|TestBuildTaskWakeupHeaders| \
        TestTaskWakeupHandshakeCarriesExtraHeaders| \
        TestCLIConfig_ExtraHeaders' \
  -count=1 -timeout 60s                         # PASS
go test ./internal/daemon/execenv/ \
  -run 'TestBriefByteIdenticalAcrossRunsForEveryKind| \
        TestInjectRuntimeConfigByteIdenticalAcrossTriggers' \
  -count=1 -timeout 60s                         # PASS (prompt-prefix stability invariant)
```

Pre-existing baseline failures in `./internal/daemon` (config / identity tests) and `./internal/cli` (config tests) are environmental and reproduce on the unmodified upstream branch; they are not regressions from this change.

## Out of scope (separate sub-issues)

- **Item 4** — Extend the headless workflow to CLI / daemon-spawn / setup probe. Architectural call (refactor vs narrow the promise) and code lands in **TIM-151** (Stage 2).
- **Item 6** — Documentation correctness (drop spoofable identity-assertion providers, drop the false "proxies strip/rewrite" claim, prefer env/secret-file/config over flags for secrets, correct the bearer-not-session-cookie statement). Lands in **TIM-150** (Stage 1, parallel).


multica-eve (changes requested):
Before this change, people who protect a self-hosted Multica installation with an additional sign-in gate must install and maintain a separate helper on every machine that runs work. After this change, Multica can present the machine credential itself, so those people should be able to remove that extra helper and keep the same protected deployment.

## Product review

**User journey:** the operator creates a machine credential at the access provider, configures it locally, restarts Multica, verifies connectivity, and then removes the local proxy workaround.

**Impact:** this is opt-in and should not change default deployments. It affects only self-hosted operators behind an authenticating access layer, but it introduces a new secret lifecycle and therefore needs strict display, storage, rotation, clearing, and destination guarantees.

**Product decision: ✅ This capability belongs in Multica.** However, the current implementation does not yet deliver the stated journey: only one of Multica's outbound client paths receives the credential, and some documented provider examples are upstream identity assertions rather than client credentials. The scope should be narrowed to real machine-to-machine credentials and completed across the headless client journey.

## Implementation review

The default-off behavior, daemon-to-server scoping, HTTP/WebSocket coverage, value redaction in config output, and tests around the new configuration are good foundations. The following items block merge.

### Must fix

1. **Reject reserved and Multica-owned headers at parse time.**

   `server/internal/daemon/client.go:227-232` and `server/internal/daemon/wakeup.go:489-508` append extra values after Multica has set `Authorization`; JSON requests similarly set `Content-Type` before the append at `server/internal/daemon/client.go:1124-1128`. This produces duplicate security/protocol headers rather than making Multica's value win, contradicting `SELF_HOSTING_ADVANCED.md:570`. A local regression test observed both `Bearer multica-real-token` and `Bearer proxy-token` on the same request.

   Please reject at least `Authorization`, `Host`, `Content-Length`, `Content-Type`, `Connection`, `Upgrade`, `Sec-WebSocket-*`, `X-Client-*`, `X-Workspace-ID`, `X-Agent-ID`, `X-Task-ID`, `Forwarded`, and `X-Forwarded-*`. Add negative tests for both HTTP and WebSocket paths.

2. **Do not forward custom secrets across origins.**

   The clients created at `server/internal/daemon/client.go:134-135` have no redirect policy. Go copies custom headers to redirects and only recognizes a small built-in set as sensitive; a local two-server test showed `Cf-Access-Client-Secret` reaching a different host after a 302. Bind extra headers to the configured server origin and either reject a cross-origin redirect or strip every extra header before following it. Cover host, scheme, and port changes with tests.

3. **Perform complete HTTP header validation at startup and add bounds.**

   `server/internal/daemon/headers.go:130-146` only rejects CR/LF/NUL and `:`. A name such as `X Bad Name` passes configuration parsing, lets the daemon start, and then fails every request in `net/http`. Use complete field-name/value validation (for example `httpguts.ValidHeaderFieldName` / `ValidHeaderFieldValue`) so the documented fail-fast contract is true. Also limit header count and aggregate bytes to avoid an unbounded startup/config surface.

4. **Complete the stated headless workflow, or explicitly narrow the product promise.**

   The PR only updates `daemon.Client`. `server/internal/cli/client.go:179-195` still sends no extra credential for `multica login`, normal CLI commands, or the `multica` calls agents use to report work. `server/internal/daemon/daemon.go:151-166` injects the server URL and task token into agent runs but not the proxy credential. `server/cmd/multica/cmd_setup.go:440-455` also probes through a bare client and may treat an access-provider login page as a healthy server after redirects.

   As written, an operator can stop the sidecar and let the daemon claim work, but CLI calls and agent result reporting will then be blocked. To keep the PR's “retire the proxy” promise, put parsing and origin-bound header application in a shared outbound client layer used by daemon HTTP/WS, CLI requests, and the setup probe. Otherwise, remove that promise and document the remaining sidecar requirement precisely.

5. **Remove the unused parallel implementation.**

   `server/internal/cli/flags.go:23-63` adds `FlagOrEnvArray`, but production code never calls it; the only other reference is a comment at `server/cmd/multica/cmd_daemon.go:1107`. `runDaemonForeground` implements separate parsing with different empty-value semantics. Keep one implementation and remove the dead helper and its large test suite.

6. **Correct the security and deployment documentation.**

   - `SELF_HOSTING_ADVANCED.md:559-562` presents `X-Forwarded-User`-style upstream identity assertions as credentials a client should send. Those are normally written by a trusted proxy after authentication and must not be client-spoofable. Keep only verified machine-to-machine credential schemes.
   - `SELF_HOSTING_ADVANCED.md:569-570` claims proxies strip/rewrite the configured value and that a custom `Authorization` is silently dropped. Neither is a general guarantee, and the latter is false in this implementation.
   - `SELF_HOSTING_ADVANCED.md:594-604` recommends command substitution while claiming the secret stays out of `ps`; the shell expands it into argv, and the background daemon keeps that argv visible. Prefer environment/secret-file/config injection and clearly warn against flags for secrets.
   - `SELF_HOSTING_ADVANCED.md:664` says daemon requests fall through to the server's session-cookie layer; daemon requests use Multica bearer authentication instead.

### Non-blocking cleanup

- Prefer one documented flag syntax instead of accepting both `Name=value` and `Name: value` with leftmost-separator rules.
- Align the promised multi-value behavior with the config-file representation, which is currently `map[string]string`.
- Remove stale “PR 1 / PR 2” process narration and replace private `TIM-142` references with `MUL-6636` / `#7492`; reduce comments that describe implementation history rather than invariants.
- Consider keeping `SetExtraHeaders` / `ExtraHeaders` internal or documenting that they are startup-only, since concurrent mutation is not synchronized.

## Verification

At the current head (`105f3c7`):

- `go build ./internal/daemon/... ./internal/cli/... ./cmd/multica/...` — passed.
- Targeted daemon/CLI configuration and wakeup tests — passed.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — passed; this change does not break the prompt-prefix stability invariant.
- GitHub CI is green, but the three security regressions above are not covered by the submitted test suite.

## Recommendation

🔄 **Request changes.** The product direction is valid, but reserved-header conflicts, cross-origin credential leakage, incomplete startup validation, and the incomplete CLI/headless journey must be fixed before merge. Please preserve the good default-off scoping and redaction behavior while addressing these blockers.

multica-eve (changes requested):
Thanks for the substantial update. I re-reviewed the current head `8462847d4ff000eea99ced69c41bf9ad92796a79`. The cross-origin redirect leak and the dead `FlagOrEnvArray` code are fixed, and the `httpguts` validation, 16 KiB aggregate bound, and core reserved-header handling are meaningful improvements. This remains **Request changes**, however, because the end-to-end ZTN workflow and the documented security contract are still not correct.

Must fix before merge:

1. **[P1] The ZTN workflow is still daemon-only, not end to end.** The configured headers are only applied by the daemon client. `server/internal/cli/client.go:154-185` creates the ordinary CLI client without them, and `server/cmd/multica/cmd_setup.go:439-455` probes the protected server without them. Consequently `multica login`, normal CLI commands, and agent result/comment calls made through the CLI are still rejected when the whole server is behind header-based access control. That makes the “retire the old proxy hop” claim in `SELF_HOSTING_ADVANCED.md:666-668` unsafe. Either cover the complete CLI/headless path without exposing proxy credentials to agent subprocesses, or explicitly narrow the product scope, documentation, and `Closes #7492` promise. Deferring this to a private TIM item does not make the current public PR's user journey work.

2. **[P1] The documentation still contradicts both the implementation and the security model.** `SELF_HOSTING_ADVANCED.md:556-570` and `.env.example:186-195` tell users to send `X-Forwarded-User`, while `server/internal/daemon/headers.go:79-82` now correctly rejects every `X-Forwarded-*` header. Those headers are normally identity assertions injected by a trusted proxy after authentication, not credentials that an arbitrary client should self-assert. In addition:
   - `SELF_HOSTING_ADVANCED.md:593-601` says values expanded with `$(cat ...)` do not appear in `ps`, but the expanded secrets become daemon argv and are visible in process listings.
   - `SELF_HOSTING_ADVANCED.md:570` says a custom `Authorization` value is silently dropped, while the new implementation rejects it at startup.
   - `SELF_HOSTING_ADVANCED.md:664` still describes daemon bearer authentication as session-cookie authentication.

   Please rewrite this section around actual client credentials that the proxy accepts (for example Cloudflare Access service-token headers), the actual daemon-only/full-client boundary, and secret-safe configuration guidance.

3. **[P1] The reserved-header set is still incomplete.** `server/internal/daemon/headers.go:60-83` still accepts framing, hop-by-hop, and proxy-authentication fields such as `Transfer-Encoding`, `Trailer`, `TE`, `Keep-Alive`, `Proxy-Authorization`, and `Proxy-Authenticate`. Arbitrary extra headers must not control HTTP framing or connection/proxy semantics. Reject these consistently in parse-time validation and in the HTTP/WS defense-in-depth send paths, with regression tests.

4. **[P2] The override path bypasses the promised common fail-fast validation.** `server/internal/daemon/headers.go:308-323` checks only map length and aggregate bytes for a non-nil override; it does not call `ValidateHeaderNameValue`, and multiple values under one name do not count toward the 64-entry limit. The CLI flag parser currently masks this for its own caller, but exported `Overrides.ExtraHeaders` can still carry invalid values to runtime. Route every active source through the same validation and entry-counting logic.

5. **[P2] The module metadata is not tidy.** This PR directly imports `golang.org/x/net/http/httpguts`, but `server/go.mod:76` still marks `golang.org/x/net` as indirect. `go mod tidy -diff` fails and moves it to the direct dependency block.

Validation on this head: GitHub CI is green (14 passed, 1 skipped); `git diff --check`, the targeted extra-header/redirect/WebSocket tests, the `ByteIdentical` prompt-cache regression test, and `go vet` for the affected packages pass. Please keep the redirect and parser fixes while addressing the remaining items above.

PR #7499, opened 2026-08-24
MUL-6542 fix(mobile): stabilize recycled markdown layout

### LINKED ISSUE #7398: [Bug]: iOS chat user bubble overflows FlashList cell

iiwish:
## What does this PR do?

Fixes the iOS chat-bubble overlap at the two places where layout first becomes incorrect, rather than asking FlashList to relayout from a descendant `onLayout` observer.

The investigation separated all three Markdown paths:

- **Prose / tables:** `react-native-enriched-markdown` 0.5 reports an intrinsically measured width that Yoga can shrink and re-wrap after the height was calculated. Version 0.6 includes the upstream multiline-width fix ([#291](https://github.com/software-mansion/enriched-markdown/pull/291)), Fabric recycle reset ([#324](https://github.com/software-mansion/enriched-markdown/pull/324)), and pending-style measurement fix ([#343](https://github.com/software-mansion/enriched-markdown/pull/343)).
- **Code:** RN's horizontal `ScrollView` defaults to `flexGrow: 1`. In the vertical bubble this allowed the code descendant to measure at 762 pt while the FlashList cell remained 434 pt, so the cell's built-in `onLayout` never received the overflowing size. Setting `flexGrow: 0` makes the bubble and cell share the same geometry.
- **Images:** no source change is needed. Native measurement showed the cell and bubble both move from the 16:9 placeholder (332.67 pt) to the resolved 1:1 image (448 pt); FlashList's built-in path already catches this resize.

The previous bubble observer, exact-float predicate, `useLayoutState` side-effect trigger, and per-recycle full-list layout have all been removed. There is no new feedback loop or duplicate `layout()` call.

## Related Issue

Closes #7398

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Upgrade `react-native-enriched-markdown` from 0.5 to 0.6 for the upstream iOS intrinsic-measurement and Fabric recycling fixes.
- Prevent the horizontal code `ScrollView` from flex-growing beyond its self-sized parent.
- Key async Shiki results by code, resolved language, and theme so a recycled `CodeBlock` never displays the previous message's tokens while the current highlight is pending.
- Update the Markdown ADR's canonical package URL/version and re-audit the 0.6 theme defaults.

## How to Test

1. Build the native iOS app from source and open the temporary 160-row Markdown recycling harness.
2. Verify one-word, multiline prose, table, known/unknown-language code, combined table + async Shiki, and 1:1 image-ratio cases for both user and assistant boundaries.
3. Force 40 alternating top/bottom `scrollToIndex` jumps and verify recycled long bubbles remain equal to their FlashList cells with no overlap or blank bubble.

Measured results on iPhone 17 Pro / iOS 26.5:

| Path | Before | After |
|---|---:|---:|
| Static code | cell 434 pt / bubble 762 pt | cell 258 pt / bubble 258 pt |
| Table + async Shiki, initial | n/a | cell 438 pt / bubble 438 pt |
| Table + async Shiki, highlighted | n/a | cell 454 pt / bubble 454 pt |
| Resolved 1:1 image | placeholder 332.67 pt | cell 448 pt / bubble 448 pt |
| One-word `ok` | n/a | 47.67 x 40 pt |

The 16 pt Shiki change and 115.33 pt image change are both detected by FlashList's existing cell `onLayout -> validateItemSize -> layout()` path after the parent/child geometry is corrected.

Verification completed:

- `pnpm install --frozen-lockfile`
- `pnpm exec turbo typecheck lint test --filter=@multica/mobile --force` (4/4 tasks; 18 files / 113 tests; iOS runner assertions pass)
- `pnpm --dir apps/mobile exec eslint lib/markdown/code-block.tsx` (0 findings)
- Native Expo clean prebuild, CocoaPods install, and from-source iOS compile with `ReactNativeEnrichedMarkdown 0.6.0`
- 160-row native harness with 40 alternating recycle jumps, including valid GFM tables, known/unknown code, async Shiki, image-ratio changes, and assistant boundaries
- Light/dark appearance switching and a cold launch at iOS Accessibility Large; final measured rows remained separated after all 40 jumps
- `pnpm dlx expo-doctor apps/mobile` (17/20 checks; the 3 failures are pre-existing SDK/package-version diagnostics unrelated to this diff)
- `git diff --check`

The full mobile lint has 0 errors and 7 existing warnings, unchanged from `main`.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable)
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (not applicable)

The primary risk is the native dependency upgrade. It stays within the 0.x minor line supported by RN 0.83, its current JS API and Expo plugin usage remain compatible, the theme defaults were audited, and the pod compiled from source in the target simulator build.

## AI Disclosure

**Prompt / approach:** Used Codex to inspect FlashList and React Native layout sources, compare enriched-markdown releases, instrument native cell/bubble measurements, build a focused recycling harness, review the patch, and run the repository verification commands. The harness and logging were removed before commit.

## Screenshots

**Before:** the code descendant outgrows the cell and paints across later rows.

Before: code block overflowing its FlashList cell

**After 24 forced recycling jumps (original evidence capture):** the long table/code bubble ends before the assistant boundary, while the one-word bubble stays tight. A second 160-row run completed 40 jumps in both appearances without overlap or blank bubbles.

After: stable long and one-word bubbles after 24 recycling jumps

### DIFF
--- a/apps/mobile/docs/markdown-rendering-adr.md
+++ b/apps/mobile/docs/markdown-rendering-adr.md
@@ -30,7 +30,7 @@ not a Multica problem.
 
 | Library | Path | Last release | Verdict for Multica |
 |---|---|---|---|
-| [`react-native-enriched-markdown`](https://github.com/software-mansion-labs/react-native-enriched-markdown) | A | v0.5.0 (Apr 2026) | **Selected for prose.** Expo officially recommends it in [Edit rich text](https://docs.expo.dev/guides/editing-richtext/) — A-tier endorsement. Software Mansion (same team as Reanimated / Gesture Handler) |
+| [`react-native-enriched-markdown`](https://github.com/software-mansion/enriched-markdown) | A | v0.6.0 (May 2026) | **Selected for prose.** Expo officially recommends it in [Edit rich text](https://docs.expo.dev/guides/editing-richtext/) — A-tier endorsement. Software Mansion (same team as Reanimated / Gesture Handler) |
 | [`react-native-streamdown`](https://github.com/software-mansion-labs/react-native-streamdown) | A + worklets | active 2026 | Not adopted. Built on enriched-markdown, optimised for AI streaming. Web/desktop don't use a streaming-specific renderer either, mobile streaming isn't currently a top product pain |
 | [`react-native-marked`](https://github.com/gmsgowtham/react-native-marked) | B | v8.1.0 (2026-05-14) | Not adopted. v7 removed `CustomToken`, v8 added "React component embedding" but no token-level customisation. Pure `<Text>` tree → would trigger nested-text bugs |
 | [`amilmohd155/react-native-markdown`](https://github.com/amilmohd155/react-native-markdown) | B | v0.8.5 (Jan 2026) | Not adopted. Same nested-`<Text>` constraint as `react-native-marked`. 14 ⭐, single maintainer, not production-validated |
@@ -106,9 +106,10 @@ enriched-markdown's `normalizeMarkdownStyle.js` carries a frozen table of
 ~30 hardcoded **light-mode** color defaults. Fields not explicitly
 overridden in `useMarkdownStyle()` use those hardcoded values and
 disappear (or render garishly) in dark mode. Every color field must be
-explicitly mapped to a `THEME[scheme]` token. **When upgrading
-enriched-markdown (v0.6+), re-audit `normalizeMarkdownStyle.js` for
-newly-added color fields** — they will also ship light-mode defaults.
+explicitly mapped to a `THEME[scheme]` token. Version 0.6.0 adds a transparent
+link background plus link-variant, superscript, and subscript defaults; none
+introduce a new opaque color. Re-audit `normalizeMarkdownStyle.js` on every
+upgrade because newly-added color fields may ship light-mode defaults.
 
 ---
 
--- a/apps/mobile/lib/markdown/code-block.tsx
+++ b/apps/mobile/lib/markdown/code-block.tsx
@@ -69,16 +69,26 @@ export function CodeBlock({ code, lang, selectable = true }: Props) {
   const { isDarkColorScheme } = useColorScheme();
   const theme = isDarkColorScheme ? SHIKI_THEME_DARK : SHIKI_THEME_LIGHT;
   const resolvedLang = resolveLang(lang);
-  const [lines, setLines] = useState<HighlightedLine[] | null>(null);
+  const [highlighted, setHighlighted] = useState<{
+    code: string;
+    lang: string;
+    theme: string;
+    lines: HighlightedLine[];
+  } | null>(null);
+  const lines =
+    highlighted?.code === code &&
+    highlighted.lang === resolvedLang &&
+    highlighted.theme === theme
+      ? highlighted.lines
+      : null;
 
   useEffect(() => {
-    if (!resolvedLang) {
-      setLines(null);
-      return;
-    }
+    if (!resolvedLang) return;
     let cancelled = false;
     void highlight(code, resolvedLang, theme).then((result) => {
-      if (!cancelled) setLines(result);
+      if (!cancelled && result) {
+        setHighlighted({ code, lang: resolvedLang, theme, lines: result });
+      }
     });
     return () => {
       cancelled = true;
@@ -88,7 +98,14 @@ export function CodeBlock({ code, lang, selectable = true }: Props) {
   return (
     <View className={CODE_BLOCK_CONTAINER_CLASS}>
       <CodeBlockHeader code={code} lang={lang} />
-      <ScrollView horizontal showsHorizontalScrollIndicator={false}>
+      {/* RN's horizontal ScrollView defaults to flexGrow: 1. In a recycled,
+          self-sized chat bubble that lets the child outgrow the FlashList
+          cell instead of sizing to its code. */}
+      <ScrollView
+        horizontal
+        showsHorizontalScrollIndicator={false}
+        style={{ flexGrow: 0 }}
+      >
         {lines ? (
           <HighlightedCode lines={lines} selectable={selectable} />
         ) : (
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -63,7 +63,7 @@
     "react": "19.2.0",
     "react-dom": "19.2.0",
     "react-native": "0.83.6",
-    "react-native-enriched-markdown": "^0.5.0",
+    "react-native-enriched-markdown": "^0.6.0",
     "react-native-gesture-handler": "~2.30.1",
     "react-native-image-viewing": "^0.2.2",
     "react-native-keyboard-controller": "1.20.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -462,8 +462,8 @@ importers:
         specifier: 0.83.6
         version: 0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0)
       react-native-enriched-markdown:
-        specifier: ^0.5.0
-        version: 0.5.0(@expo/config-plugins@55.0.9)(katex@0.16.45)(react-native@0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0))(react@19.2.0)
+        specifier: ^0.6.0
+        version: 0.6.0(@expo/config-plugins@55.0.9)(katex@0.16.45)(react-native@0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0))(react@19.2.0)
       react-native-gesture-handler:
         specifier: ~2.30.1
         version: 2.30.1(react-native@0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0))(react@19.2.0)
@@ -9269,8 +9269,8 @@ packages:
       react-native-svg:
         optional: true
 
-  react-native-enriched-markdown@0.5.0:
-    resolution: {integrity: sha512-CYihNlcg8IjsrVbmypDjFHsDF6Z6vYO0yraw6CkR30SDsg99J6fcBrOkyWOItO8PIDD21jw8yD1eYP/0LBOCdA==}
+  react-native-enriched-markdown@0.6.0:
+    resolution: {integrity: sha512-/ZkKZv9jGPAuc/oXevkf3MpVfcrG3pjHoVMyKzpJhGsmIXxeTRr2dlsIqo66k+Pa4JrChBg+HB6ImQ44Fg3wJw==}
     peerDependencies:
       '@expo/config-plugins': '>=50.0.0'
       katex: '>=0.16.0'
@@ -20902,7 +20902,7 @@ snapshots:
     transitivePeerDependencies:
       - supports-color
 
-  react-native-enriched-markdown@0.5.0(@expo/config-plugins@55.0.9)(katex@0.16.45)(react-native@0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0))(react@19.2.0):
+  react-native-enriched-markdown@0.6.0(@expo/config-plugins@55.0.9)(katex@0.16.45)(react-native@0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0))(react@19.2.0):
     dependencies:
       react: 19.2.0
       react-native: 0.83.6(@babel/core@7.29.0)(@types/react@19.2.14)(react@19.2.0)


Bohan-J (changes requested):
Thanks for picking this up, and for actually reading the FlashList source and docs rather than guessing — the write-up made this reviewable. The refactor half is also a clear improvement over the earlier attempt (more on that below).

I can't merge this version, though. The core problem is that it re-lands the approach that was already tried and replaced for this exact bug, and the mechanism it adds duplicates one FlashList already runs.

## The blocking issue: this is the same approach as `a8eb62fce`

Commit `a8eb62fce` ("fix(mobile): remeasure user chat bubbles") was the first commit on #7399's branch, targeting this same issue. Side by side:

| | `a8eb62fce` (replaced) | This PR |
|---|---|---|
| Observation point | `onLayout` on the user bubble `View` | same |
| Trigger | increment a `useLayoutState` setter | same |
| Dedup condition | exact `messageId` + `height` compare | same, plus finite/non-negative checks |

#7399's own description explains why it was dropped:

> The prior `useLayoutState` callback asked FlashList to lay out again from an observer-only `onLayout`. When native markdown changed height, this formed a recycling feedback loop and could create the large blank bubble reported in build 21.

This PR brings back the feedback loop, the per-recycle `layout()` cost, and the user-only asymmetry, without addressing why build 21's blank bubble won't recur. That needs a direct answer, not just a cleaner implementation of the same idea.

## 1. The observer duplicates FlashList's built-in path

I pulled `@shopify/flash-list@2.0.2` and read it. Every cell container already has an `onLayout`:

```js
// src/recyclerview/ViewHolder.tsx:83
const onLayout = useCallback(
  (event: LayoutChangeEvent) => {
    onSizeChanged?.(index, event.nativeEvent.layout);
  }, [index, onSizeChanged]);
...
<CompatContainer ref={viewRef} onLayout={onLayout} style={style} index={index}>
```

`onSizeChanged` is `validateItemSize`:

```js
// src/recyclerview/RecyclerView.tsx:343
const validateItemSize = useCallback((index, size) => {
  const layout = recyclerViewManager.getLayout(index);
  ...
  if (areDimensionsNotEqual(width, size.width) ||
      areDimensionsNotEqual(height, size.height)) {
    // TODO: Add a warning for missing useLayoutState
    recyclerViewContext.layout();      // <-- here
  }
}, ...);
```

And `useLayoutState`'s setter does:

```js
// src/recyclerview/hooks/useLayoutState.ts:48
recyclerViewContext?.layout();          // <-- the same call
```

So the observer added here terminates in the identical `recyclerViewContext.layout()` that FlashList already calls — just observed one level deeper in the tree.

That built-in path is live in this list, not bypassed. A vertical list uses `LinearLayoutManager`, which only sets `layout.enforcedWidth = !this.horizontal` (`LinearLayoutManager.ts:88`); `enforcedHeight` is never set anywhere in the package. So in `ViewHolder`'s style, `height: layout.enforcedHeight ? layout.height : undefined` resolves to `undefined` — the cell height follows its content. Bubble grows → container grows → container's `onLayout` fires → `validateItemSize` → `layout()`.

The docs say the same thing ([usage.md](https://github.com/Shopify/flash-list/blob/main/documentation/docs/fundamentals/usage.md)):

> Item layout changes will still be detected using `onLayout` callback in the absence of `useLayoutState`, which might not look as smooth on a case-by-case basis.

So if the overlap still reproduces on `main`, the missing ingredient isn't a `layout()` call — that call already happens. Adding a second one is likely to miss the root cause. This is exactly the measurement #7399's review asked for: identify which content path disagrees with the cached cell size, then decide whether the fix belongs in the bubble, in `Markdown`, or in the list's size cache. That measurement is still missing.

## 2. Exact float comparison replaces a deliberate tolerance

`apps/mobile/lib/chat-message-layout.ts:15` compares exactly:

```ts
previousMeasurement.height !== nextHeight
```

FlashList deliberately uses a tolerance for the same comparison (`src/recyclerview/utils/measureLayout.ts:78`, comment: *"with a small tolerance. Used to handle floating-point precision issues in layout measurements"*):

```ts
export function areDimensionsEqual(value1: number, value2: number): boolean {
  return Math.abs(
    PixelRatio.getPixelSizeForLayoutSize(value1) -
      PixelRatio.getPixelSizeForLayoutSize(value2)
  ) <= 1;
}
```

I ran the predicate on its own: two heights differing by one ULP (`48.33333333333333` vs `48.333333333333336`) return `true`, i.e. they trigger a full-list `layout()` — on a difference FlashList itself treats as noise.

That's the concrete mechanism behind "observer-only `onLayout` forms a feedback loop": `layout()` → relayout → sub-pixel jitter in the reported height → this predicate says "changed" → `layout()` again. The guard only stores the *previous* height, with no tolerance, no convergence bound, and no oscillation detection, so any value differing from its immediate predecessor keeps it spinning. That shape matches the build 21 blank bubble.

Minimum fix: compare after `PixelRatio.roundToNearestPixel()`, or match FlashList's ±1 device-pixel tolerance. Note this doesn't resolve point 1.

## 3. `useLayoutState` is used as a pure side-effect trigger

```ts
// apps/mobile/components/chat/chat-message-list.tsx:334
const [, setLayoutRevision] = useLayoutState(0);
```

The state value is discarded; the only thing wanted is the setter's `layout()` side effect. But `useLayoutState` is `useState` + `layout()`, so each trigger costs a React re-render that changes nothing visually, plus a full-list layout pass. The first half is pure waste.

The documented role is *"useful if you want to resize a child component based on a local state"* — i.e. hold the state that **determines** the size, not feed a signal back from `onLayout`. If an observer is kept, `useRecyclerViewContext()?.layout()` expresses it without the dead render — though see the suggestion below.

## 4. Every recycle pays a full-list layout

`hasMeasuredMessageHeightChanged(null, ...)` returns `true`, and the ref is keyed by `messageId`, so when a cell is recycled onto a new message the first `onLayout` always triggers `layout()`. That's at least one full-list relayout per user message per recycle, during scrolling — the "per-recycle `layout()` cost" #7399's review named, and it works against the reason FlashList v2 was adopted here in the first place (the rationale in this file's header comment).

## Suggested direction

`Markdown` (`apps/mobile/lib/markdown/markdown.tsx`) splits content into three kinds, and two of them change height via local state in an async callback — precisely the documented `useLayoutState` case:

- `apps/mobile/lib/markdown/code-block.tsx:72` — `const [lines, setLines] = useState<HighlightedLine[] | null>(null)`; height changes when the async Shiki result swaps `PlainCode` for `HighlightedCode`.
- `apps/mobile/lib/markdown/markdown-image.tsx:45` — `const [aspect, setAspect] = useState<number | null>(null)`; the 16:9 fallback is replaced by the real ratio once `Image.getSize` resolves, which can be a large delta.

Switching those two to `useLayoutState` declares "this state changes my size" at the source of the resize instead of observing it from outside. That:

- covers assistant messages too, not just user ones;
- has no `onLayout` observer, so no feedback loop;
- costs nothing per recycle — it fires only on a real async height change.

Using it outside a FlashList is safe; the implementation guards with optional chaining:

```ts
// src/recyclerview/hooks/useLayoutState.ts:48
recyclerViewContext?.layout();
```

`useRecyclerViewContext()` returns `undefined` with no provider, so the setter degrades to plain `useState`. Call sites outside the list (`issue-description.tsx`, `comment-card.tsx`) are unaffected.

The third kind, `EnrichedMarkdownText`, is a native intrinsically-measured view whose height change doesn't go through JS state — that one needs to be confirmed separately, which is exactly why the measurement in point 1 is worth doing first.

## Tests

The predicate itself is correct — I executed `hasMeasuredMessageHeightChanged` directly and its behavior matches all five assertions. (I didn't run the full Vitest suite; dependencies weren't installed in my checkout.) Extracting it as a pure function with unit tests is a good instinct.

But it's worth being precise about what the coverage means. `apps/mobile/vitest.config.ts` states:

> Mobile vitest is intentionally minimal — Node environment only, scoped to pure-function helpers in `lib/` ... We don't ship jsdom or RN test renderers here

So this lane structurally cannot reach this bug: the real failure mode needs FlashList recycling, native async layout, and `layout()` reentrancy. "17 Vitest files / 114 tests passed" is true but largely orthogonal to whether the patch is correct.

Also, three of the five cases (`NaN`, `Infinity`, `-1`) guard against inputs RN layout never produces. The case that would matter — sub-pixel jitter — isn't there, and it would fail today (point 2). Worth adding; it surfaces the tolerance problem directly.

## What's good here

- Extracting the user branch into `UserMessageRow` (`chat-message-list.tsx:324`) is the right call, and better than `a8eb62fce`'s hoist into `MessageRow`, which registered layout state for every message type. It also matches the existing `AssistantRow` split in this file.
- Isolating the decision into a pure, tested helper is the right structure — the problem is the condition, not the shape.
- The refactor preserves the `isSelecting` short-circuit and long-press behavior, and carries no unrelated churn. The diff reads cleanly, which is a real improvement over the previous round.

## To move forward

1. **Measurement**: which of the three content paths (prose / `CodeBlock` / `MarkdownImage`) disagrees with the cached cell size, and why doesn't FlashList's own `onLayout → validateItemSize → layout()` already cover it?
2. **Loop**: what makes this version safe from build 21's blank bubble, given it matches `a8eb62fce`? At minimum, align the height comparison to a device-pixel tolerance and add a sub-pixel jitter test.
3. **Symmetry**: assistant messages use the same `Markdown` with the same async height changes — why fix only user messages? (See the suggestion above for handling both at once.)
4. **Evidence**: before/after screenshots or a screen recording covering both a long-markdown message and a one-word message. The template asks for this on UI changes, and for a purely visual bug it's what makes the fix reviewable — the linked issue itself has a screenshot.
5. Add `MUL-6542` to the PR title (I've done this).

The underlying bug is real and worth fixing, and the investigation here is more careful than the last round. It's the approach that needs rework, not the goal — please do keep going.

iiwish:
@Bohan-J Thanks for the detailed review. Your objection was correct: the descendant `onLayout -> useLayoutState` observer duplicated FlashList's own invalidation path and did not explain the mismatch. I rewrote the branch from current `main` and removed that approach completely.

The native instrumentation isolated the three paths:

1. **Prose / tables:** the mismatch starts in `react-native-enriched-markdown` 0.5's iOS intrinsic measurement, before FlashList can observe a correct cell size. The branch upgrades to 0.6, which contains the multiline-width fix ([upstream #291](https://github.com/software-mansion/enriched-markdown/pull/291)), Fabric recycle reset ([#324](https://github.com/software-mansion/enriched-markdown/pull/324)), and pending-style measurement fix ([#343](https://github.com/software-mansion/enriched-markdown/pull/343)).
2. **Code:** this was the concrete parent/child disagreement. RN's horizontal `ScrollView` has `flexGrow: 1` by default. Before the fix, the FlashList cell measured 434 pt while its bubble descendant painted at 762 pt. Because the cell itself never became 762 pt, `ViewHolder.onLayout` had no overflowing size to pass to `validateItemSize`. Setting the code scroller to `flexGrow: 0` made the static code cell and bubble both 258 pt.
3. **Images:** I left this path unchanged. The cell and bubble both changed from the 16:9 placeholder (332.67 pt) to the resolved 1:1 size (448 pt), so FlashList's built-in path already handles it.

I also keyed the async Shiki result by `{code, resolvedLang, theme}`. A recycled `CodeBlock` renders its current plain source immediately rather than briefly showing the previous message's highlighted tokens. When highlighting completes, both the cell and bubble move together from 438 to 454 pt; FlashList's existing `onLayout -> validateItemSize -> layout()` catches that 16 pt change.

Loop/performance answer: there is now no `useLayoutState`, descendant layout observer, exact-float comparison, direct recycler `layout()` call, or per-recycle invalidation in this PR. The shared Markdown/CodeBlock fixes apply symmetrically to user and assistant content and to the other Markdown surfaces.

Evidence and verification are in the rewritten PR body, including before/after screenshots. The native harness covered one-word and long prose, tables, known/unknown-language code, async Shiki, images, assistant boundaries, and 24 forced top/bottom recycle jumps. The one-word `ok` bubble stayed 47.67 x 40 pt; long cell/bubble measurements stayed equal with no overlap or blank bubble.

Final checks: frozen install, forced mobile typecheck/lint/test (4/4 tasks, 18 files / 113 tests, iOS runner assertions), affected-file ESLint, `git diff --check`, and a from-source iOS build with the 0.6 native pod. The 7 lint warnings are pre-existing on `main`; there are 0 errors.

Could you please take another look at the rewritten implementation?

iiwish:
@Bohan-J A gentle follow-up on the rewritten implementation. The descendant onLayout/useLayoutState observer, exact-float predicate, and per-recycle relayout path from the reviewed head are gone. The current branch instead fixes the measured sources of the mismatch: enriched-markdown 0.6 for intrinsic/Fabric measurement, flexGrow: 0 for the horizontal code ScrollView, and keyed async highlighting for recycled code blocks. The PR body includes native cell/bubble measurements and recycle-jump evidence for prose, code, images, user, and assistant rows; current CI is green. When convenient, could you please take another look and clear the outstanding changes request if this direction looks right? Thanks.

Bohan-J (approved):
Thank you for the rework — this is a genuinely better fix than either earlier attempt, and the native instrumentation is what made it reviewable.

The `flexGrow: 1` finding is the part that stands out. It explains something my review had left as an open question: why FlashList's built-in `onLayout → validateItemSize → layout()` was live but never corrected the cell. Because the descendant outgrew the cell while the cell itself never changed size, there was no overflowing measurement for `validateItemSize` to act on. That's a parent/child geometry bug, and fixing the geometry is the right layer — much better than adding another invalidation signal on top.

Verified on my side:

- **`flexGrow` claim** — RN 0.83.6 `ScrollView.js:1877`: `baseHorizontal` is `{flexGrow: 1, flexShrink: 1, ...}`, and `:1766` composes `props.style` last, so the override applies. Confirmed.
- **0.6 theme audit** — diffed `normalizeMarkdownStyle.ts` between 0.5.0 and 0.6.0. The only new color default is `transparent` (×2); the other 19 hardcoded colors are byte-identical, and the new `superscript`/`subscript` entries carry only `fontScale`/`baselineOffsetScale`. Your ADR note is accurate, and the dark-mode concern the ADR raises doesn't apply here.
- **Upstream references** — #291, #324, #343 all exist and are merged (May 5 / 18 / 23), with titles matching the fixes you describe.
- **Recycling logic** — `resolveLang` returns `string | null`, so an unknown language can never key-match a stale entry and correctly falls through to `PlainCode`. `highlight()` is typed `Promise<HighlightedLine[] | null>`, so the `&& result` guard is right.
- **Regression surface of `flexGrow: 0`** — `CODE_BLOCK_CONTAINER_CLASS` is auto-height (`px-3 py-2`) at every call site, so nothing was relying on the ScrollView growing.
- **Screenshots** — both load, and the one-word `ok` bubble stays tight, so the `w-[80%]` short-bubble regression from #7399 hasn't returned.

Leaving images alone with a measurement showing the cell and bubble already move together was the right call — declining to change something is easy to argue for and hard to prove, and you proved it.

Two non-blocking notes for whenever you're next in this file:

- `style={{ flexGrow: 0 }}` allocates a new object per render; the rest of this file goes through NativeWind (`CODE_BLOCK_CONTAINER_CLASS`), so `className="grow-0"` or a hoisted `StyleSheet.create` constant would match local convention.
- On a recycled block with an unknown language, `highlighted` keeps the previous message's token array alive (the old `setLines(null)` released it). Bounded by live cell count, so it's negligible — just noting the behavior change.

One heads-up for everyone else rather than for this PR: the enriched-markdown bump is a native dependency change, so anyone on a prebuilt dev client will need to rebuild after this lands.

Merging. Thanks for sticking with this through three iterations and for taking the earlier feedback in the spirit it was meant.

PR #7507, opened 2026-08-24
MUL-6648: feat(cli): add issue comment update command

### LINKED ISSUE #3348: [Feature]: add command [comment update]
### Deployment type

Official App

### What do you want and why?

add command [comment update]

_No response_

_No response_

iiwish:
## Summary

- Add `multica issue comment update <comment-id>` with inline, stdin, and UTF-8 file input.
- Require a positive `--expected-revision` and reuse the server's existing compare-and-swap update path, preventing silent overwrite of concurrent edits.
- Send only `content` and `expected_revision`, so existing attachments remain unchanged.
- Preserve existing CLI output and local-file safety conventions without changing server behavior.
- Document how to read the current revision from `multica issue comment list <issue-id> --output json`.
- Document the command in CLI help, localized CLI documentation, and the on-demand Issues skill reference; keep it out of the resident agent brief.

## Usage

```bash
multica issue comment update <comment-id> --content "Updated text" --expected-revision 3
multica issue comment update <comment-id> --content-file ./comment.md --expected-revision 3
printf 'Updated text' | multica issue comment update <comment-id> --content-stdin --expected-revision 3
```

The three content inputs are mutually exclusive. Missing, zero, and negative revisions fail locally before any request is sent.

## Testing

```bash
cd server
go test ./cmd/multica ./internal/daemon/execenv
go test -race ./cmd/multica -run 'Test(IssueCommentUpdateCommandRegistration|RunIssueCommentUpdate|ResolveTextFlag)' -count=10
go test -race ./internal/daemon/execenv -run 'Test(SlimQuickCreateAvailableCommands|InjectRuntimeConfigAvailableCommandsCoreOnly)' -count=5
go vet ./cmd/multica ./internal/daemon/execenv
```

The request-path regressions run beneath a daemon task marker with a task-scoped test token.

Closes #3348


iiwish:
Synced with `main` at `d44932089` in `f59d56498` and resolved the runtime-brief conflict.

The merged brief preserves both the revision-safe comment update command and main's `repo checkout --fresh` behavior/safety warning. The core-command regression now asserts both together. The built-in Issues reference also documents reading the revision, reconciling stale edits, preserving attachments, and the existing edit-trigger behavior.

Verification:
- Complete `cmd/multica` and `internal/daemon/execenv` test suites pass with the agent-CLI guard.
- `go vet` for both packages and whitespace checks pass.
- Fresh [GitHub CI](https://github.com/multica-ai/multica/actions/runs/34585147813) passes on `f59d56498`, including backend/agent tests, frontend, sqlc, Windows execenv and installers.

The branch is mergeable again; no server API behavior was changed by this refresh.

iiwish:
Hi @Bohan-J, when you have a chance, could you take a look at #7507 and #8202? The first adds revision-safe CLI comment editing through the existing API; the second fixes stale timeline timestamps without rerendering the full page. Both are currently conflict-free with passing checks.

No rush. If either is outside the current priorities, a quick steer would help me focus the follow-up work. Thanks!

multica-eve (changes requested):
Requesting changes: the CLI capability is worth accepting and the core CAS-based implementation is sound, but two issues should be fixed before merge.

The product behavior is the right one: a CLI user who notices a mistake can now update the original comment instead of switching surfaces or posting a second correction. Requiring the current revision prevents silent concurrent overwrites, and omitting `attachment_ids` preserves existing attachments. Reusing the existing update endpoint also keeps the established permission, tenant-isolation, and agent-trigger behavior.

Blocking changes:

1. `server/internal/daemon/execenv/runtime_config_sections.go:278` puts roughly 400 rendered bytes of low-frequency command guidance into the static brief for every issue session. Comment editing is not part of the core agent loop, and the revision, permission, attachment, and trigger semantics are already documented in the on-demand `multica-platform` Issues reference and in CLI `--help`, which has no resident prompt cost. Please remove this brief entry and its always-present command assertions while keeping the CLI help and skill reference. Passing `ByteIdentical` proves stability, but not that this content belongs in the cached brief.

2. The new request-path tests at `server/cmd/multica/cmd_issue_test.go:427` and `:549` inherit `MULTICA_TOKEN=test-token` from `setCLITestServerEnv`. Inside a daemon-managed agent workdir, `newAPIClient` detects the task marker and requires a task-scoped `mat_` token, so both tests fail before making their HTTP request. This makes the PR's own targeted test command fail in the environment used by agent reviewers. Please set `MULTICA_TOKEN=mat_test-token` in those cases, following the existing agent-workdir-safe test pattern, or make the shared helper explicitly provide a task-safe token.

Validation on head `f59d56498`: the new CLI tests and their `-race -count=10` run pass from a normal CLI working directory; the full execenv suite, the required `ByteIdentical` check, the targeted execenv race tests, `go vet`, and `git diff --check` pass. GitHub CI is green, and a local merge-tree against current `main` is conflict-free.

iiwish:
Thanks for the review. Both blockers are addressed in `01dec15f9`, synced with main `9663b4a87`:

- Removed comment editing from the resident brief and its command assertions. The PR now has no execenv/brief diff against main; CLI help and the on-demand Issues reference remain. Updated the PR description accordingly.
- Both request-path tests now create a daemon task marker in an isolated directory and use `mat_test-token`. With the marker present, both failed with the reported token error before the fix and pass after it.

Guarded full CLI/execenv suites, focused `-race -count=10` checks (including ByteIdentical), and vet passed. [CI completed successfully](https://github.com/multica-ai/multica/actions/runs/35050886809).

@multica-eve, could you re-review when convenient? The review-request API returned 404 for my account, so I am requesting it here.

multica-eve (approved):
Thank you for addressing both review points so carefully. I re-verified the updated head: the resident brief change is gone, the daemon-workdir request tests now use task-scoped credentials, the focused and full local checks pass, and CI is green. Approved for merge.

PR #5787, opened 2026-07-22
fix(autopilot): prevent overlapping scheduled runs

fsafey:
## Summary
- serialize scheduled run_only admission per autopilot across server replicas
- record durable skipped receipts when live work already exists
- repair stale partial receipts without consuming same-plan retry slots
- ignore terminal downstream work when a listener missed run synchronization
- preserve explicit manual and webhook overlap behavior

## Verification
- make test ENV_FILE=.env.worktree
- go test -race ./internal/service -run TestScheduledRunOnlyAutopilotDoesNotOverlap -count=10
- go test ./cmd/server -run TestDispatchAutopilotForPlan
- git diff --check
- three Claude Opus review passes; all critical/warning findings resolved

Operational context: VWO-469

PR #7513, opened 2026-08-24
fix(autopilot): mutex gate prevents concurrent scheduled runs (ALL-206)

j499712089:
### Problem

The Autopilot scheduler (cron → `DispatchAutopilotForPlan`) did not check
whether an autopilot already had an in-flight run before dispatching a new
slot. A run that outlived its slot interval let the next slot start a **second
concurrent run for the same autopilot**, duplicating the autopilot's side
effect (a second task/issue on the same schedule).

### Fix

A single admission gate in `shouldSkipDispatch` (`server/internal/service/autopilot.go`)
— the pre-flight choke point shared by every dispatch entry point
(schedule `DispatchAutopilotForPlan`, manual `DispatchAutopilotManual`,
webhook `AdmitAutopilotWebhookDelivery`, API):

- New sqlc query `GetAutopilotRunInFlight` returns the most recent
  `autopilot_run` that has not reached a terminal state. The in-flight status
  set (`pending`, `issue_created`, `running`) mirrors the partial index
  `idx_autopilot_run_status` from migration 042, which is the authoritative
  definition of "in flight".
- When a run is in flight, the slot is recorded as `skipped` with
  `failure_reason = "previous run still in flight (<run_id>)"` and the typed
  reason code `dispatch.ReasonAlreadyActive` — a concurrent run is never
  created.
- Transient DB errors fail open (log + proceed), consistent with the rest of
  the admission gate, so a hiccup never silently swallows a scheduled run.

```go
// server/internal/service/autopilot.go
prior, err := s.Queries.GetAutopilotRunInFlight(ctx, ap.ID)
if err == nil {
    return "previous run still in flight (" + util.UUIDToString(prior.ID) + ")", dispatch.ReasonAlreadyActive, true
}
if !errors.Is(err, pgx.ErrNoRows) {
    slog.Warn("autopilot admission: failed to check for in-flight run",
        "autopilot_id", util.UUIDToString(ap.ID), "error", err)
    return "", "", false // fail open
}
```

```sql
-- server/pkg/db/queries/autopilot.sql
-- name: GetAutopilotRunInFlight :one
SELECT * FROM autopilot_run
WHERE autopilot_id = $1
  AND status IN ('pending', 'issue_created', 'running')
ORDER BY created_at DESC
LIMIT 1;
```

### Regression tests

`server/cmd/server/autopilot_mutex_gate_test.go` (real-DB integration, same
pattern as `autopilot_dispatch_for_plan_test.go`, no new dependencies):

- `TestAutopilotMutexGateSkipsOverlappingSlot` — slot 1 dispatches and is
  `running`; the very next slot while slot 1 is still in flight is recorded as
  `skipped` with a `failure_reason` naming the predecessor, carries no task,
  and the manual "run now" surface returns `ReasonAlreadyActive`. Task count
  across all three attempts is exactly 1 (no concurrent task).
- `TestAutopilotMutexGateAllowsDispatchAfterTerminal` — after slot 1 reaches
  `completed` (the same flip the task listener applies), the next slot
  dispatches normally (`running` + task, no `failure_reason`). Task count is
  exactly 2.

### Out of scope, documented

**Trigger label mismatch** (`label: "guardian-60m"` vs
`cron_expression: "0 */2 * * *"`, trigger `17a136da-9539-4fbc-859c-31023d1430a8`):
this trigger is **not** created or seeded by this repository — no code, seed,
or migration references the trigger id or the `guardian-60m` label (verified by
repo-wide search and against a fresh migrated DB). It is owner-only production
runtime configuration, so it cannot be fixed in this PR. Owner action: update
the trigger's `label` in the production DB to match its actual 2-hour cadence,
or rename the autopilot if "every 60 minutes" was the intent.

### Long-run protection (recommendations, not in this PR)

The mutex gate prevents *overlap*, but a single run that consistently outlives
its slot interval will now be skipped every time — operators should add, as
follow-ups:

1. **Hard cap**: a maximum `autopilot_run` lifetime (e.g. N × slot interval)
   after which the scheduler marks the run `failed`/`skipped` and force-releases
   the slot, instead of skipping forever.
2. **Sharding / job splitting**: for autopilots whose workload legitimately
   exceeds the slot interval, split the unit of work so one run completes within
   one interval (align the cron expression to the real expected duration).
3. **Alerting**: surface `failure_reason = "previous run still in flight (...)"`
   skips in the autopilot health/pause monitor — sustained skips indicate a
   slot interval that is too short for the workload, not just a one-off overlap.

### DIFF
--- a/server/cmd/server/autopilot_mutex_gate_test.go
+++ b/server/cmd/server/autopilot_mutex_gate_test.go
@@ -1,0 +1,212 @@
+package main
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgtype"
+
+	"github.com/multica-ai/multica/server/internal/dispatch"
+	"github.com/multica-ai/multica/server/internal/events"
+	"github.com/multica-ai/multica/server/internal/service"
+	"github.com/multica-ai/multica/server/internal/util"
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+// TestAutopilotMutexGateSkipsOverlappingSlot is the ALL-206 regression for the
+// scheduler mutex/lease gate: while an autopilot still has an in-flight run
+// (autopilot_run.status IN ('pending','issue_created','running') — the exact set
+// of the partial index idx_autopilot_run_status from migration 042), the next
+// scheduled slot MUST NOT start a second concurrent run. It is recorded as a
+// `skipped` run with a clear failure_reason instead.
+func TestAutopilotMutexGateSkipsOverlappingSlot(t *testing.T) {
+	ctx := context.Background()
+	queries := db.New(testPool)
+	bus := events.New()
+	taskSvc := service.NewTaskService(queries, testPool, nil, bus)
+	autopilotSvc := service.NewAutopilotService(queries, testPool, bus, taskSvc)
+
+	var agentID string
+	if err := testPool.QueryRow(ctx,
+		`SELECT id::text FROM agent WHERE workspace_id = $1 ORDER BY created_at ASC LIMIT 1`,
+		testWorkspaceID,
+	).Scan(&agentID); err != nil {
+		t.Fatalf("load fixture agent: %v", err)
+	}
+
+	ap := createAutopilotForMutexTest(t, ctx, queries, "ALL-206 mutex in-flight", agentID)
+	trigger := createTriggerForMutexTest(t, ctx, queries, ap)
+
+	// Slot 1 dispatches normally and leaves a run in flight (running + task).
+	plannedAt1 := time.Now().UTC().Truncate(time.Second).Add(-2 * time.Minute)
+	first, err := autopilotSvc.DispatchAutopilotForPlan(ctx, ap, trigger.ID, "schedule", nil, plannedAt1)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


multica-eve:
Thanks for digging into this and for adding real-DB regression coverage. The underlying problem is real: a long-running scheduled Autopilot can overlap a later occurrence and produce duplicate work, side effects, and cost.

We’re going to close this PR as a duplicate rather than ask you to rework this branch. Two existing PRs are already addressing the same area: #5367 covers an explicit per-Autopilot overlap policy, while #5787 contains the atomic admission and stale-run recovery work. The file overlap is substantial, so keeping a third implementation active would split review and make convergence harder.

There are also two important differences we need to preserve in the consolidated solution:

- Overlap must be a per-Autopilot `skip` / `allow` choice. A global gate would also suppress manual, webhook, and API triggers; those may carry distinct user intent or payloads and must not be silently dropped.
- The admission decision must be atomic. In this version, checking for an active run and creating the next run are separate operations, so two simultaneous dispatches can both pass the check. We also need stale partial-run recovery so one abandoned row cannot suppress every future occurrence indefinitely.

This closure is not a rejection of the problem you identified—the investigation and tests are useful confirmation of it. We just want the work to land through one coherent policy and implementation path. Thanks again for the contribution.

Issue #6865, opened 2026-08-12
[Bug]: bundled multica.exe is unreadable in Windows Codex elevated sandbox


HumbertoCecim:
### Deployment type

Official App

### Summary

On Windows, a Multica-managed Codex task receives the task-scoped Multica environment and the correct bundled CLI directory in `PATH`, but Codex's native `elevated` sandbox cannot initially read or execute the bundled `multica.exe`.

The failing boundary is the Windows ACL on the Desktop installation directory, not credential injection or `PATH` construction.

### Root-cause evidence

- `MULTICA_TASK_CONFIG_ROOT`, `MULTICA_DAEMON_PORT`, task/workspace IDs, and the task credential are present.
- `PATH` contains `<user>\AppData\Local\Programs\@multicadesktop\resources\app.asar.unpacked\resources\bin`.
- Sandboxed commands run as `<machine>\CodexSandboxOnline` or `CodexSandboxOffline`; both are members of `<machine>\CodexSandboxUsers`.
- The bundled CLI directory initially grants access only to SYSTEM, Administrators, and the interactive user. It has no ACE for `CodexSandboxUsers`.
- `Get-Command multica`, `Test-Path`, directory enumeration, or direct execution therefore fails with access denied / command not found although the file exists.
- The target `bin` directory contains only `multica.exe`.

### Steps to reproduce

1. Install Multica Desktop v0.4.24 on Windows x64.
2. Configure Codex with:

```toml
[windows]
sandbox = elevated
```

3. Assign an issue to a local Codex agent.
4. In the task, run:

```powershell
whoami
Get-Command multica
Test-Path '<user>\AppData\Local\Programs\@multicadesktop\resources\app.asar.unpacked\resources\bin\multica.exe'
multica version
```

5. Observe that the task identity is a dedicated Codex sandbox user while the bundled CLI path is inaccessible.

### Verified minimal workaround

Grant only inherited Read & Execute on the exact bundled CLI directory:

```powershell
$multicaBinDir = '<user>\AppData\Local\Programs\@multicadesktop\resources\app.asar.unpacked\resources\bin'
$sandboxGroup = $env:COMPUTERNAME\CodexSandboxUsers
icacls $multicaBinDir /grant ${sandboxGroup}:(OI)(CI)(RX)
```

After this change, a **fresh Multica Codex task** produced:

```text
identity=<machine>\codexsandboxonline
Get-Command multica -> ...\app.asar.unpacked\resources\bin\multica.exe
[...]

PR #7514, opened 2026-08-24
MUL-6654: fix(daemon): stage task-local multica CLI, add sandbox PATH fallbacks for Codex (ALL-219)

j499712089:
## Summary

Upstreams three local daemon patches (ALL-219) that fix real Codex sandbox failures on the desktop build. All three were previously uncommitted working-tree changes on an old release branch; this PR re-applies them on a clean v0.4.33 baseline.

### 1. `stageTaskCLI` — task-local CLI staging

`server/internal/daemon/daemon.go`

Provider sandboxes (Codex on Windows) can be unable to read the daemon's install directory. `runTask` now copies the running multica executable into the task temp dir and prepends that directory to the agent `PATH`, so `multica` commands resolve inside the sandbox.

- `stageTaskCLI` function: `server/internal/daemon/daemon.go:9010`
- `runTask` staging call: `server/internal/daemon/daemon.go:7425`

### 2. `MULTICA_CLI_PATH` — explicit fallback env var

`server/internal/daemon/daemon.go` sets `agentEnv["MULTICA_CLI_PATH"]` to the staged executable path (`server/internal/daemon/daemon.go:7438`) for deterministic fallback when a sandbox filters `PATH`.

`server/internal/daemon/prompt.go` instructs agents to use the absolute CLI path when `multica` is not on `PATH` (`server/internal/daemon/prompt.go:615`).

### 3. `preferWindowsCodexPowerShell` — real PowerShell over WindowsApps alias

The desktop daemon can outlive a user PATH update, leaving its inherited PATH pointed at the zero-byte WindowsApps execution alias for `pwsh.exe`. `windowsCodexPowerShellPath` prepends a verified real PowerShell directory for Codex tasks.

- `preferWindowsCodexPowerShell`: `server/internal/daemon/daemon.go:181`
- `windowsUserPath` (reads user PATH from registry, Windows only): `server/internal/daemon/windows_user_path_windows.go:7`

## Tests

- `TestStageTaskCLI_CopiesExecutableIntoTaskTempDir`
- `TestStageTaskCLI_ReportsMissingSource`
- `TestPreferWindowsCodexPowerShellPrefersRealUserPathOverWindowsAppsAlias`
- `TestPreferWindowsCodexPowerShellKeepsPathWithoutRealPowerShell`
- `TestPreferWindowsCodexPowerShellLeavesOtherProvidersUntouched`
- `TestBuildChatPromptAttachmentIDsCanBeBoundToCreatedIssues` (extended for `MULTICA_CLI_PATH` guidance)

## Verification

```
go build ./cmd/multica        # exit 0
go test ./internal/daemon/ -run 'PreferWindows|TaskCLI' -v   # 5/5 PASS
go test ./internal/daemon/ -run 'TestBuildChatPromptAttachmentIDsCanBeBoundToCreatedIssues' -v  # PASS
```

## Notes

- Patch semantics are preserved exactly as deployed; nothing was deleted or rewritten.
- The PR contains only the six files tied to these three patches — none of the unrelated working-tree changes from the old release branch.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"io"
 	"log/slog"
 	"math/rand"
 	"net"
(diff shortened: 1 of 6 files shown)


multica-eve:
Thanks for the contribution and for documenting the intended Windows failure paths. We reviewed this against the existing Windows Codex reports and are going to close this PR rather than request another revision.

There is a real bug here, but the verified path is narrower than this change: in #6865, a user has explicitly enabled Codex's native Windows sandbox, commands run as `CodexSandboxOnline` / `CodexSandboxOffline`, and the Desktop-bundled `multica.exe` is inaccessible because `CodexSandboxUsers` lacks Read & Execute permission. A durable fix must be scoped to Windows Codex native-sandbox tasks and prove that the staged CLI is executable by that sandbox identity.

This PR does not establish that:

- it stages the daemon binary for every provider and OS, making a Windows-only compatibility fix a mandatory startup dependency for all tasks;
- the task-temp copy receives no `CodexSandboxUsers` RX ACL, and the tests only prove that bytes were copied, not that a real sandbox task can execute the result;
- the `MULTICA_CLI_PATH` prompt points to the same unverified file, so an absolute path cannot solve an ACL denial; platform command availability should be guaranteed by the runtime rather than delegated to prompt fallback;
- the PowerShell/WindowsApps PATH rewrite is a separate behavior without a linked public reproduction, and should not be bundled with the CLI ACL fix.

#7002 is closer to the verified #6865 boundary because it scopes staging to Windows Codex and attempts the minimum RX grant, although that branch now conflicts and still needs a real Windows Desktop sandbox smoke test. We will continue the narrow fix through #6865 / a rebased or maintainer-owned implementation instead of evolving this broader patch. The separate Codex installer-junction failure from #6762 was already fixed by #6802.

Closing because addressing the review would require replacing most of the solution and splitting unrelated behaviors, not making incremental changes to this PR. Thanks again for taking the time to contribute.

PR #7516, opened 2026-08-24
MUL-6655 Fix MCode ACP session startup race (ZIC-199)

### LINKED ISSUE #7168: [Bug]: MiniMax Code ACP exits when session/new is sent immediately after initialize
### Deployment type

Official App

### What happened?

Creating a MiniMax Code agent in Multica Desktop fails with:

```text
mcode initialize failed: mcode process exited

nvironment
Host: macOS on Apple Silicon (Mac mini)
Runtime provider: mcode / MiniMax Code
MiniMax Code version: 0.1.3
Node.js: v22.22.3
Node module ABI: 127
MCode command: /Users/<redacted>/.minimax-code/bin/mcode
Account: MiniMax global account, successfully logged in
ACP command: mcode acp
Multica Desktop: current version that includes the MiniMax Code runtime

Impact
MiniMax Code is detected as an available runtime, but users cannot complete MiniMax Code agent creation in Multica Desktop despite having a valid login and a functioning local ACP CLI.

### Steps to reproduce

Run the following from the same task work directory used by Multica.
Case A — send session/new immediately
printf '%s\n%s\n' \
'{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"multica-agent-sdk","version":"0.2.0"}}}' \
'{"jsonrpc":"2.0","id":2,"method":"session/new","params":{"cwd":"'"$PWD"'","mcpServers":[]}}' \
| mcode acp 2>&1

Actual output
Only the initialize response is returned. No response for session/new is received:
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":1,"agentCapabilities":{"loadSession":true,"mcpCapabilities":{"http":true,"sse":true},"promptCapabilities":{"image":false,"audio":false,"embeddedContext":false},"sessionCapabilities":{"list":{},"resume":{},"close":{}}},"agentInfo":{"name":"minimax-code","title":"MiniMax Code","version":"0.1.3"}}}

This matches the behavior seen in Multica Desktop, which reports:
mcode initialize failed: mcode process exited

### Screenshots (optional)

_No response_

```shell
Root-cause assessment
This is strongly indicative of an initialization timing race or ACP compatibility issue:

mcode acp is found, starts, and returns a valid initialize response.
The same authenticated MCode installation works in the exact Multica task work directory.
The only observed behavioral difference is the delay between initialize and session/new.
Multica's MCode adapter appears to send session/new immediately after receiving the initialize result: server/pkg/agent/mcode.go.
The evidence does not establish whether the underlying defect belongs to MCode or Multica. However, the Multica adapter could make agent creation resilient by waiting briefly for MCode readiness, or by restarting and retrying session/new when the process closes immediately after a successful initialization.
```

vicksiyi:
## What does this PR do?

Adds a short MCode-only readiness gap after ACP `initialize` before the first session RPC, so `session/new` is not sent back-to-back while MiniMax Code is still finishing startup. It also preserves a structured process-exit sentinel so a provider exit after initialize is reported as a `session/new` startup failure instead of being confused with initialize.

## Related Issue

Closes #7168

Multica issue: ZIC-199

## Type of Change

- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/mcode.go`: waits briefly after successful initialize before `session/new` / supported `session/load`, and maps post-initialize `session/new` process exits to an actionable startup error.
- `server/pkg/agent/mcode_test.go`: adds fake ACP coverage for delayed session readiness and provider exit after initialize.

## How to Test

1. `cd server && go test ./pkg/agent -run 'TestMcode'`
2. `cd server && go test ./pkg/agent`
3. `make check-worktree`

Note: `make check-worktree` exited 0 locally, but Docker was not running, so its PostgreSQL container check printed `Cannot connect to the Docker daemon`.

## Checklist

## AI Disclosure

**Prompt / approach:**
Traced the MCode ACP startup sequence from the issue report, kept the compatibility delay scoped to MiniMax Code, and added focused fake-provider tests for the reported initialize-to-session race and for provider exit behavior.

## Screenshots (optional)

N/A

### DIFF
--- a/server/pkg/agent/mcode.go
+++ b/server/pkg/agent/mcode.go
@@ -404,9 +404,10 @@ func mcodeSessionStartupExited(operation string, err error) bool {
 	if operation != "session/new" {
 		return false
 	}
-	if errors.Is(err, errMcodeProcessExited) || errors.Is(err, io.ErrClosedPipe) || errors.Is(err, syscall.EPIPE) {
-		return true
-	}
-	text := strings.ToLower(err.Error())
-	return strings.Contains(text, "process exited") || strings.Contains(text, "broken pipe")
+	// Only trust transport-level signals. A still-running MCode reports
+	// session/new failures as a JSON-RPC error, and session/new is where
+	// mcpServers are launched, so text matching on "process exited" or
+	// "broken pipe" would rewrite a caller's real MCP startup error into
+	// an MCode-exited message and drop its root cause.
+	return errors.Is(err, errMcodeProcessExited) || errors.Is(err, io.ErrClosedPipe) || errors.Is(err, syscall.EPIPE)
 }
--- a/server/pkg/agent/mcode.go
+++ b/server/pkg/agent/mcode.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os/exec"
@@ -41,6 +42,9 @@ type mcodeBackend struct {
 }
 
 var mcodeReaderDrainGrace = 2 * time.Second
+var mcodeSessionStartupReadyDelay = 100 * time.Millisecond
+
+var errMcodeProcessExited = errors.New("mcode process exited")
 
 type mcodeMessageStream struct {
 	ch     chan Message
@@ -183,7 +187,7 @@ func (b *mcodeBackend) Execute(ctx context.Context, prompt string, opts ExecOpti
 				c.handleLine(line)
 			}
 		}
-		c.closeAllPending(fmt.Errorf("mcode process exited"))
+		c.closeAllPending(errMcodeProcessExited)
 	}()
 
 	go func() {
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for this @vicksiyi — nice catch on the root cause. Keeping the readiness gap scoped to MCode and reusing the run context so cancel/timeout still interrupt the wait was the right call, and replacing the `fmt.Errorf("mcode process exited")` string with a proper `errMcodeProcessExited` sentinel is a real improvement on its own.

I pushed one small commit to your branch (40e83ab) rather than sending you back around for it.

**What changed:** `mcodeSessionStartupExited` also matched `process exited` / `broken pipe` anywhere in the error text. A still-running MCode reports `session/new` failures as a JSON-RPC `*acpRPCError`, and `session/new` is the call that launches `mcpServers` — so a genuine error like:

```
failed to start MCP server "filesystem": process exited with code 1
```

got rewritten into `MiniMax Code ACP exited before starting a session; ... upgrade MiniMax Code`. Because that branch returns a fixed string with no `%v`, the original error was dropped rather than relabelled, pointing the member at the wrong subsystem.

I removed the text matching and kept the transport-level signals. Nothing regressed: real provider exits already arrive as `errMcodeProcessExited` via `closeAllPending`, and a stdin write racing the exit arrives wrapped as `syscall.EPIPE`. Your `TestMcodeSessionNewExitAfterInitializeReportsStartupFailure` passes unchanged — it was already exercising the sentinel path, not the substring one. I added a table test pinning both directions so this can't drift back.

Verified: `go test ./pkg/agent -count=1` green; the new tests fail against the pre-fix version and pass after, so they aren't vacuous.

**Non-blocking follow-ups, not for this PR:**

- The 100ms wait is a fixed sleep against a race, so its upper bound on slow machines is unproven. If reports persist, an observable readiness signal or a bounded retry would beat growing the constant.
- `mcodeSessionStartupExited` returns early for `session/load`, so a resumed session hitting the same exit still gets the generic message.
- `retry agent creation` is slightly narrow — members reach `session/new` outside agent creation too.

Merging once CI is green. Thanks again for tracking this one down.

Issue #7520, opened 2026-08-24
repo checkout: on-demand `multica repo checkout` intermittently fails with "requires an active task credential"


utgpay2:
## Bug: `multica repo checkout` intermittently fails with "repo checkout requires an active task credential"

**Environment**
- `multica version`: `0.4.20` (commit `93342d04a`, built `2026-08-06T09:48:39Z`), `go1.26.1`, `darwin/arm64`
- Daemon version reported in logs: `v0.4.33`
- Runtimes affected: both `codex` (codex-cli `0.147.0` and `0.149.1`) and `claude` — not tied to one provider
- Self-hosted desktop app, local daemon profile

**Symptom**

An agent's own in-task invocation of `multica repo checkout <url>` (documented as "Used by agents to check out repos on demand") intermittently fails with:

```
repo checkout requires an active task credential
```

This happens on:
- Freshly dispatched tasks (`resume_session=false reuse_workdir=false`) — not just resumed sessions
- Both a newly-created task and one re-triggered after several hours
- Both runtimes (Codex and Claude)

Meanwhile the daemon's *own* automatic checkout at task dispatch (`repo checkout: worktree created` in daemon.log) succeeds essentially every time for the same repo/workspace — so the repo itself, network access, and the registered repo credentials are fine. Only the manual/on-demand `repo checkout` invocation from inside an agent's own task session hits this.

**Reproduction pattern observed**

Over ~13 hours (same workspace, same registered repo, same daemon instance, no config changes), the same agent repeatedly got this error across a dozen+ separate task dispatches, both fresh and resumed. One task run out of many *did* succeed and completed real work (opened a PR), but the very next dispatch on the same issue hit the identical error again — i.e. it is not a permanent misconfiguration, it's flaky/intermittent, consistent with a race or a credential-issuance step that doesn't always complete before the agent's tool call reaches the daemon.

Restarting the daemon (desktop app relaunch) did not reliably resolve it — failures continued to occur in dispatches after a restart.

No daemon-level ERROR/WARN log line was found for this — the message only surfaces via the CLI's own stderr inside the agent's tool call output, so it was hard to correlate with anything actionable in `daemon.log`.

**Impact**

[...]

PR #7533, opened 2026-08-25
MUL-6662 fix(daemon): inject MULTICA_CLI so agent IPC skips PATH lookup (#7520)

Bohan-J:
Closes the structural half of #7520. [#7524](https://github.com/multica-ai/multica/pull/7524) made the failure self-explanatory; it did not stop the wrong binary from being resolved. This does.

## Problem

daemon and CLI are the same binary speaking a version-coupled private protocol over localhost — but *which* binary an agent ran was decided by `PATH`, a variable the user, their shell profile, and a provider sandbox can all rewrite.

The daemon already prepended its own bin directory to the task `PATH`, and that was never enough. The agent's shell sources the user's profile, and one `eval "$(brew shellenv)"` re-prepends `/opt/homebrew/bin` ahead of it. An unrelated older Homebrew `multica` answers, and the mismatch surfaces as a credential error that names nothing about `PATH`. That is where #7520's intermittency came from: same machine, same workspace, one success in a dozen dispatches, depending on that shell's lookup order.

The daemon knew a compatible path the whole time — `resolveSelfExecutable()`, itself — and threw that certainty into `PATH` to gamble with.

## Change

`applySelfBinaryEnv` exports `MULTICA_CLI` with the absolute path of the binary this daemon process runs, and keeps the `PATH` prepend as the fallback so nothing invoking a bare `multica` breaks.

When the executable cannot be resolved, **neither** layer is applied. An empty `MULTICA_CLI` would survive `${MULTICA_CLI:-multica}` and defeat the very fallback the docs promise, so an unusable value is absent rather than blank.

Guidance to prefer it lands on three surfaces:

- **Runtime brief** — one rule in the section that already says "use the `multica` CLI". That covers every command in every section, including the `--help` surfaces the brief never enumerates, and cannot drift from a per-bullet copy (cf. MUL-5696). Emitted for every task kind: quick-create trims its command list to `issue create`, but that call runs through the same binary. The variable is referenced, never interpolated — its value is per-machine, and this file is the prompt-cache prefix (MUL-5377).
- **`multica --help`** — a `MULTICA_CLI` entry under ENVIRONMENT VARIABLES, reaching every daemon-local subcommand from one place. That block lives in `rootHelpTemplate`, which hardcodes its prose and never renders `rootCmd.Long`; a note on the command struct would have shipped text nobody can see, so a test pins it to the rendered output.
- **`repo checkout --help`** and the `multica-runtimes-and-repos` skill — the path that actually broke. Its old text ("Used by agents to check out repos on demand") was teaching the bare name.

Both POSIX and PowerShell forms appear wherever the invocation does. `$MULTICA_CLI` is a PowerShell *variable*, not the environment, so a Windows agent copying only `"$MULTICA_CLI"` would run an empty command name — a worse failure than the bare name it replaces. The path is quoted for the reason a Windows install path needs it: spaces.

## What this does not claim

`resolveSelfExecutable` reports where the process *started*, not the bytes on disk now — the daemon deliberately supports the binary being replaced out of band with the restart deferred while tasks run (`trySelfReload`). This removes the `PATH` race, not every possible version skew, and the brief, help text, and skill all say so rather than overclaiming. #7524 was careful about the same line.

Splitting the "cloud API client" and "daemon IPC client" roles apart is the larger follow-up and is not in this PR.

## Testing

- `applySelfBinaryEnv`: value present and absolute; `PATH` prepend preserved; nothing injected when `resolveSelfExecutable` fails; relative path rejected while `PATH` still written; `MULTICA_CLI` not overridable via `custom_env`.
- Brief: guidance present for all four task kinds, names the unset fallback, teaches both shell forms, and interpolates no literal path (byte-stable output).
- Help: `multica --help` and `repo checkout --help` both carry the rule — verified to fail when the text is removed, since the rendering path is exactly what made the first attempt dead code.

Local: `go build ./...`, `go vet`, and `internal/daemon/...` + `cmd/multica` + builtin-skill tests all pass. The CLI suite must run outside a daemon task context — its task-boundary tests detect the ambient `.multica/daemon_task_context.json` marker and fail regardless of this change.


Bohan-J:
Pushed `98b95e7` — both must-fixes plus all three non-blocking items.

## Must-fix 1: the deleted-binary window

You were right, and the hole was in the scenario this repo already documents. `restartTargetBinary`'s own comment says brew cleanup deletes the Cellar path `os.Executable()` reports; `auto_update.go` defers the restart while tasks run and gates nothing. So a task genuinely can start under a daemon whose own path is gone.

Injecting there was strictly worse than injecting nothing, exactly as you described: every guidance surface keys its fallback on the variable being **unset**, so a set-but-dead path skips the fallback and fails deterministically, where a bare `multica` still resolves the current version through Homebrew's stable symlink. That breaks the "worst case equals today's behavior" claim I made, so the claim was wrong, not just the code.

`usableCLIPath` now gates injection on absolute + present + executable regular file. The `PATH` prepend stays unconditional — a directory that no longer exists is simply skipped during lookup, which is the pre-change behavior.

Two things I did **not** paper over. The check races the filesystem (the binary can vanish between injection and the agent's first command); the comment states that rather than implying atomicity. What it buys is turning the one *predictable* window into a no-op. And I did not fall back to the stable brew symlink as a second choice — that path is not version-matched to the running daemon either, so it would trade a verified claim for an unverified one. Absent is the honest answer.

Regression tests: deleted binary, directory in place of the binary, and non-executable file — each injects nothing while `PATH` survives.

## Must-fix 2: help contradicted itself

Correct, and worse than missing: root help showed bare `multica` in EXAMPLES a few lines below text demanding `MULTICA_CLI`. Both surfaces now state the unset branch explicitly, and the rendered-help tests assert the fallback rather than just the variable name and PowerShell form.

## Non-blocking

- **`${MULTICA_CLI:-multica}`** — you're right, `:-` substitutes on unset *or* empty; my comments claimed the opposite. The decision to omit stands, but for the actual reason: the guidance keys on *unset*, so any unrunnable value — empty, relative, or dangling — defeats the fallback it was meant to trigger. Comments and test rationale rewritten.
- **Source map** pointed at `rootCmd.Long`; corrected to `rootHelpTemplate`, with the reason recorded so nobody re-adds copy to the dead field.
- **Codex allowlist** now names `MULTICA_CLI` explicitly in the expected set — good catch that the env map passing is not the same as reaching the shell tool.

## Verification

`go build`, `go vet`, `gofmt`, and the full `internal/daemon/...`, `cmd/multica`, and builtin-skill suites all pass. Every new assertion was mutation-checked: reverting `usableCLIPath` to the old `IsAbs`-only gate fails all three new subtests, and deleting the fallback sentence fails the rendered-help test.

Ready for re-review.

PR #7535, opened 2026-08-25
feat(lark): push inbox notifications to a bound member's Lark DM (delivery)

adaiguoguo:
WeCom's smart bot already pushes `inbox:new` to a member who has linked their account (`internal/integrations/wecom`); Lark had no equivalent, so a Lark-bound member only discovered assignments, mentions, comments and status changes the next time they opened the app.

This adds the Lark side: the outbound Patcher subscribes to `inbox:new` and delivers to the member's DM when they have a binding row, addressing them by open_id so Lark opens the 1:1 conversation itself (`SendTextParams` and `SendMarkdownCardParams` gain `ReceiveIDType`; empty keeps existing callers on chat_id). Every miss — agent recipient, unbound member, missing installation, send failure — is a no-op and the in-app inbox is unaffected. Notification preferences need no second check: a muted type never produces an inbox item.

The notification is a schema-2.0 markdown card, not plain text: bodies are member/agent-authored markdown and land in the same conversation as the agent's reply cards, so both render the same way — bold 【type】title header, body verbatim, one 查看详情 link. Assignment-family notifications arrive body-less upstream (`notifyDirect` passes `""`), so the push backfills the issue description — it is exactly what the recipient needs to decide whether to act, and only for that family: repeating it on every status flip would be spam.

**Scope**: this is the delivery half of #7136, re-cut for reviewability (−250 lines, none of the locking/patching subtleties). A follow-up PR adds per-issue aggregation — folding an event burst into one growing card via card patching — which is why every markdown card already declares `update_multi` at send time and `composeInboxCard` accepts follow-up blocks.

**Why binding-by-member rather than per-installation**: `FindChannelBindingForMember` deliberately keys on (member, channel_type), matching the WeCom precedent — the bot is the pipe, the inbox is member-scoped. A workspace-level notification bot would be a cleaner home but is an installation-model change, out of scope here.

Verified end-to-end on a self-hosted deployment: assignment, mention, comment and status-change notifications land in the bound member's Feishu DM as individual cards.

### DIFF
--- a/server/internal/integrations/lark/client.go
+++ b/server/internal/integrations/lark/client.go
@@ -296,7 +296,21 @@ type SendTextParams struct {
 	// ReplyTarget threads the text reply back into a Lark topic; see
 	// ReplyTarget. Empty keeps the chat-level send.
 	ReplyTarget ReplyTarget
+	// ReceiveIDType tells Lark how to read ChatID. Empty means chat_id —
+	// the historical chat-level send. ReceiveIDOpenID addresses a person
+	// instead, which is how the inbox push reaches a bound member without
+	// a pre-existing chat. Ignored when ReplyTarget is set, since a reply
+	// is addressed by message id.
+	ReceiveIDType ReceiveIDType
 }
+
+// ReceiveIDType is Lark's receive_id_type query parameter: it selects
+// which identifier namespace receive_id is read in.
+type ReceiveIDType string
+
+// ReceiveIDOpenID addresses a single user by their per-app open_id. The
+// bot opens (or reuses) a 1:1 conversation with that person.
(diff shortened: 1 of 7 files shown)


adaiguoguo:
Closing to stop the rebase upkeep on an 800-line diff in an active area — two weeks with no review is a fair signal that an external feature PR is not this queue's intake path. The feature itself keeps running in production on our fork, and the code here stays readable for whoever wants it.

The gap this addressed is now tracked as #8152 (inbox push is WeCom-only; Lark/DingTalk/Slack members get nothing), which is the durable form of the ask. Glad to re-cut this against current `main` on any signal of interest.

PR #7542, opened 2026-08-25
fix(billing): make Cloud seat enforcement automatic

ldnvnbl:
## Summary

- replace the separate Fleet/capacity URLs and seat-capacity switches with the single managed `MULTICA_CLOUD_URL` connection
- enable strict seat admission whenever Multica is Cloud-connected; fail request paths closed when the capacity token is absent or invalid
- start the durable recovery worker only when the Cloud URL and capacity machine token form a valid executor, leaving live intents untouched during credential outages
- split invitation throttling into a non-consuming precheck and post-reservation consumption; persistent `capacity_full` and `capacity_overcommitted` rejections consume only the actor budget
- preserve every Cloud/proxy HTTP 429 as retryable, parse both `Retry-After` forms, and defer outbox rows without consuming attempts
- use Cloud's rate-limit scope so a workspace 429 delays only that tenant while global or unscoped proxy 429s stop the current batch
- add `DeferClaimedSeatCapacityIntent` to the seat-capacity SQL and commit regenerated sqlc output
- keep post-purchase invitation failures retryable with the same purchase idempotency key
- update Helm/Compose/environment wiring, UI messages, alerts, and regression coverage

Cloud remains the authority for plans, prices, quotas, purchased capacity, and effective rules. Multica only renders Cloud facts and executes Checkout/Portal/capacity operations.

Companion Cloud PR: https://github.com/multica-ai/multica-cloud/pull/57

## Required deployment migration

This is intentionally breaking and has no compatibility path. Before starting the new Multica binary, **delete/unset** all retired variables from environment files, Secrets, ConfigMaps, and deployment manifests:

- `MULTICA_FLEET_URL`
- `MULTICA_CLOUD_FLEET_URL`
- `MULTICA_CLOUD_FLEET_TIMEOUT`
- `MULTICA_SUBSCRIPTION_CAPACITY_ENABLED`
- `MULTICA_SUBSCRIPTION_CAPACITY_URL`
- `MULTICA_SUBSCRIPTION_CAPACITY_WORKER_ENABLED`

The startup guard treats every non-empty value as retired configuration. In particular, the old documented value `MULTICA_SUBSCRIPTION_CAPACITY_ENABLED=false` **still causes startup to fail**; setting it to `false` is not a migration. Remove it.

Then configure:

- `MULTICA_CLOUD_URL` and `MULTICA_CLOUD_TIMEOUT` for the shared Fleet/Billing/capacity Cloud connection
- `MULTICA_SUBSCRIPTION_CAPACITY_SERVICE_TOKEN` whenever `MULTICA_CLOUD_URL` is set; the token must match Cloud and be at least 32 bytes
- `MULTICA_SUBSCRIPTION_CAPACITY_TIMEOUT` if the default is unsuitable

Entitlements do not share this URL or token; they retain their independent `MULTICA_ENTITLEMENT_POLICY_*` configuration.

Apply the Multica database migrations before rollout. For pre-launch Cloud test databases, follow the companion PR's release note and rebuild from the current migration ladder because Cloud migration `0013` was intentionally revised in place.

For an individual test workspace whose ledger drifted while enforcement was bypassed, revoke accidental pending invitations first, then delete that workspace's Cloud `billing_seat_operations` and `billing_seat_capacity` rows so the next request seeds from a fresh product snapshot.

## Behavior and operational notes

- A normal full-seat purchase flow spends two actor attempts: the persistent full rejection and the final successful invite. Workspace and recipient budgets are preserved for the retry.
- Repeated capacity 429s never dead-letter an intent. They remain pending and are observed through `MulticaSeatCapacityOutboxStalled`; non-429 failures that exhaust attempts use `MulticaSeatCapacityOutboxDeadLettered`.
- The purchase-to-invite handoff remains current-page state. Refreshing or navigating away keeps the purchased seat but stops automatic invitation; the user can return and invite with the available seat.
- The limiter check-to-consume window intentionally permits bounded concurrent overshoot rather than charging shared recipient/workspace budgets before Cloud reserves capacity.

## Verification

Earlier local verification on this branch covered the server build/vet, targeted Go capacity/handler tests, views typecheck/lint, Vitest seat-purchase tests, and Helm config checks. The latest review follow-up was intentionally left to the PR CI without additional local test runs.


ldnvnbl:
Deployment note for this revision: compatibility is intentionally removed. The server now refuses to start when any retired variable has a non-empty value: `MULTICA_FLEET_URL`, `MULTICA_CLOUD_FLEET_URL`, `MULTICA_CLOUD_FLEET_TIMEOUT`, `MULTICA_SUBSCRIPTION_CAPACITY_ENABLED`, `MULTICA_SUBSCRIPTION_CAPACITY_URL`, or `MULTICA_SUBSCRIPTION_CAPACITY_WORKER_ENABLED`. Migrate Fleet/Billing/runtime connectivity to `MULTICA_CLOUD_URL` / `MULTICA_CLOUD_TIMEOUT` and keep the current capacity service token/timeout settings before rollout. This makes the previously silent loss of `mcn_` PAT and Cloud Runtime connectivity explicit.

PR #7554, opened 2026-08-25
MUL-6687 fix(codex): wait for turn/completed before finishing raw turns

golb:
## Bug

Raw Codex runs could finish too early when `final_answer`, `thread/status/changed: idle`, or a non-retrying top-level `error` arrived before the authoritative `turn/completed` notification. This could truncate later lifecycle events and usage accounting.

## Root cause

Those informational or deliverable notifications called `onTurnDone` directly. Completion deduplication also depended on a turn ID, so duplicate ID-less completions could finish a run and count usage more than once.

## Fix

- Keep final answers, idle status, and raw errors non-terminal.
- Use raw `turn/completed` as the sole notification-level completion boundary.
- Preserve final-answer delivery and terminal error capture while waiting.
- Deduplicate completion once per single-turn `codexClient`, with or without a turn ID.

## Tests

Updated unit coverage verifies the three non-terminal notifications, completion and error propagation, duplicate and replayed completions with and without IDs, exact-once usage accounting, subagent isolation, and unchanged legacy completion behavior.

Local verification:

- targeted lifecycle tests, 5 repetitions: pass;
- all Codex tests: pass;
- `git diff --check`: clean.

## Deployment

No deployment, migration, configuration, or infrastructure change is included. This is a backend code-and-tests-only patch.

### DIFF
--- a/server/pkg/agent/codex.go
+++ b/server/pkg/agent/codex.go
@@ -2169,8 +2169,7 @@ type codexClient struct {
 	onDiscardedNotification func(method string, params map[string]any)
 
 	notificationProtocol string // "unknown", "legacy", "raw"
-	turnStarted          bool
-	completedTurnIDs     map[string]bool
+	turnCompleted        bool
 
 	usageMu sync.Mutex
 	usage   TokenUsage // accumulated from turn events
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thank you @golb for the excellent contribution! 🎉

This is a well-executed fix: you correctly identified that `turn/completed` is the sole authoritative completion boundary in the raw protocol, and the implementation preserves exactly what should be preserved — final-answer delivery stays immediate, terminal errors are still captured, and the legacy protocol is untouched. The switch to a per-client boolean for exact-once completion (covering ID-less duplicates and usage accounting) is simpler and strictly safer than the previous turn-ID map for a single-turn client.

The test coverage is particularly appreciated — replay, ID-less duplicates, and the failed-turn error flow are all pinned down.

Reviewed and verified locally (full Codex suite + repeated lifecycle tests). Merging.

PR #7565, opened 2026-08-26
MUL-6689: feat(agents): support MCP config for the Oh-My-Pi runtime

### LINKED ISSUE #7558: [Feature]: supoort UI set agent mcp when bind runtime choose pi or oh my pi , like codex
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

supoort UI set agent mcp when bind runtime choose pi or oh my pi , like codex 
codex:

Image

pi or oh my pi(UI not view mcp settings):

Image

_No response_

_No response_

feiguoL:
## What does this PR do?

Adds MCP configuration support to the Oh My Pi (omp) runtime. Its MCP tab is visible in the agent UI, and managed agent MCP settings are materialized through the daemon execution environment before the runtime starts.

## Related Issue

Closes #7558

## Why Pi is excluded

Pi (pi.dev) explicitly does not support MCP — it is a stated design decision in Pi's README Philosophy section:

> **No MCP.** Build CLI tools with READMEs (see Skills), or build an extension that adds MCP support.

Confirmed against `@mariozechner/pi-coding-agent@0.67.2`: the published `dist/` contains no MCP code, `pi --help` exposes no MCP subcommand or flag, and `~/.pi/agent/` contains only `auth.json`, `bin/`, `sessions/`, and `settings.json`. MCP with Pi is only possible through a user-authored extension, which defines its own configuration format — there is no `.pi/mcp.json` convention for Multica to write into.

OMP (Oh My Pi), a Pi fork, does support MCP and reads `.omp/mcp.json` as its highest-precedence project-level path.

## Type of Change

## Changes Made

- Add `omp` to the frontend MCP-support allowlist (Pi removed — see above).
- Show the MCP tab for Oh My Pi in the agent overview UI.
- Move MCP materialization into `execenv`, writing to the runtime-native path:
  - Oh My Pi: `{workDir}/.omp/mcp.json`
- Use `sidecarManifest` for refuse-to-clobber behavior and daemon teardown/reuse cleanup.
- Add `.omp` to repo-cache git excludes so MCP credentials and runtime sidecars are not committed.
- Report OMP native MCP inventory in the runtime capabilities section.
- `Reuse` now bails (returns nil, forcing fresh Prepare) on `preparePiMcpConfig` failure, matching Cursor's behaviour — prevents stale MCP config from persisting across runs.
- Added a comment documenting that OMP's inventory reads only `~/.omp/agent/mcp.json` (user-scope), not the full precedence chain.
- Added backend, daemon, execenv, repocache, core, and views regression tests.

## How to Test

1. Bind an agent to an Oh My Pi runtime.
2. Open Agent capabilities and confirm the MCP tab is visible.
3. Configure an MCP server and run the agent; the daemon writes the managed config to `.omp/mcp.json` before launch.
4. Verify cleanup through the normal daemon sidecar teardown path.
5. Run `pnpm exec vitest run --config vitest.config.ts agents/mcp-support.test.ts` from `packages/core`.
6. Run `pnpm exec vitest run --config vitest.config.ts agents/components/agent-overview-pane.test.tsx` from `packages/views`.
7. Run the focused Go tests under `server/internal/daemon/execenv`, `server/internal/daemon/repocache`, and `server/internal/daemon`.

## Checklist

- [ ] If this PR touches Chinese product copy, I checked it against conventions.zh.mdx

## AI Disclosure

**AI tool used:** OpenCode

**Prompt / approach:** Investigated the frontend MCP provider allowlist, daemon runtime inventory/merge path, Pi/OMP runtime configuration behavior, execenv sidecar lifecycle, and repo-cache git exclusions. Implemented provider-native MCP materialization with manifest-backed cleanup and overwrite protection, then added focused regression coverage.

## Screenshots (optional)

Not included. The change is backend/runtime behavior plus a small tab-visibility allowlist update; UI tests cover the visibility behavior.


feiguoL:
@jbarket @quake @coderberry @chouti  please review codes

Bohan-J (changes requested):
Thanks for taking this on — the gap you identified in #7558 is real, and closing it is something we want. Pi and Oh-My-Pi being the only mainstream runtimes without an MCP entry point is an inconsistency worth fixing, and the shape of the change (allowlist + a real materialisation path, rather than just unhiding the tab) is the right instinct: unhiding alone would have turned "silently ignored" into "visibly configured but still ignored".

A few things also done well: refusing to clobber a pre-existing file, guarding cleanup with `sync.Once`, and covering all three states (writes / does not write / refuses to overwrite) in tests.

That said, I'd like to request changes before this lands. There are four issues, and the first two can break user tasks outright.

---

## 1. The config is very likely written to a path Pi does not read

The change rests on this premise (`server/pkg/agent/pi.go`, and the comment at `server/internal/daemon/runtime_mcp.go:286`):

> Both Pi and OMP (oh-my-pi) discover MCP servers from a `.mcp.json` file in the project root

Checking both tools' documentation, this does not hold up:

- **Upstream `pi`** reads `~/.pi/agent/mcp.json` (global) and `.pi/mcp.json` (project-level), with project overriding global. The "project-root `.mcp.json`" convention that shows up in search results belongs to `pi-mcp`, a third-party adapter package, not to `pi` itself. As written, `pi` most likely never reads the file we produce.
- **`omp`** does read a project-root `.mcp.json`, but it sits **last (9th)** in its precedence chain, behind `.omp/mcp.json`, `.omp/.mcp.json`, profile-level, user-level, and — at position 7 — **third-party tool configs (Claude Code, Cursor, Codex, …)**. The oh-my-pi docs are explicit: *"Use root `mcp.json` / `.mcp.json` only when you want a portable fallback file that other MCP clients may also read."*

This also makes the comment at `runtime_mcp.go:286` incorrect on its own terms. It asserts:

> There is no runtime-level inheritance to merge […] the agent's managed mcp_config is the sole source of MCP servers.

`omp` has a substantial inheritance chain, and most of it outranks the file we write. On any machine where the user also has Claude Code or Cursor configured with MCP, the servers configured through this new UI would be silently overridden, with no signal to the user.

Worth flagging: **the tests in this PR cannot catch this.** The fake executable in `pi_mcp_test.go` only does `test -f .mcp.json || exit 1`, which verifies that *we wrote the file* — not that *Pi reads it*. That contract can only be established by running the real binaries.

**Suggestion:** write `.pi/mcp.json` for `pi` and `.omp/mcp.json` for `omp`. Those are each tool's highest-precedence project-level path, and they also resolve issue #2 below. Please verify against real `pi` and `omp` installs before merge and note the versions you tested in the PR description — everything else here depends on this being right.

## 2. Colliding with a repo's own `.mcp.json` fails the task permanently

`TestPiExecuteFailsClosedOnExistingMcpJson` shows the fail-closed is deliberate, and refusing to clobber user files is the right principle. The problem is the chosen filename: `.mcp.json` is Claude Code's project-scope MCP convention, so a large number of repositories already ship one at the root. Two consequences:

1. A user configures MCP in the newly-visible tab, and every Pi task on such a repo **stops starting at all** — `Execute` returns before the process is spawned. That is a worse outcome than the pre-PR behaviour of ignoring the config.
2. Task working directories are reused across runs (`repocache`'s `updateExistingWorktreeContext` path). `mcpCleanup` is an in-process `defer`, so it does not run if the daemon crashes, is killed, or the machine restarts. A leftover `.mcp.json` then makes **every subsequent** Pi task on that workdir fail until someone manually deletes it — a self-poisoning state whose failure mode is "fails forever", not "fails once".

The equivalent Cursor path survives this because it goes through `sidecarManifest`, where cleanup is performed by the daemon at task teardown rather than by a `defer` inside the backend. Moving to the `.pi/` / `.omp/` namespaces also drops the collision probability to near zero.

## 3. MCP credentials can end up committed

`mcp_config` is a secret-bearing field — that is what the whole `redactMcpConfig` path exists for. From `server/internal/handler/agent.go:1023`:

```go
// MCP configs (which routinely embed third-party API tokens) — the
```

This PR writes that content in plaintext to the root of the agent's working directory, and `agentGitExcludePatterns` (`server/internal/daemon/repocache/cache.go:58`) does not cover it:

```go
var agentGitExcludePatterns = []string{
	".agent_context", "CLAUDE.md", "AGENTS.md",
	".claude", ".opencode", ".deveco", "CODEBUDDY.md", ".codebuddy",
}
```

A single `git add -A` / `git commit -am` inside a task is enough to put third-party tokens into a commit, and a root-level file is far more likely to be swept up than something nested like `.cursor/mcp.json`.

**Suggestion:** add `.pi` and `.omp` to `agentGitExcludePatterns`. That also protects the existing path that already writes into `{workDir}/.pi/skills/`.

## 4. This logic belongs in `execenv`, not in the backend

`execenv` already has everything needed:

- `params.McpConfig` is plumbed through, and `prepareCursorMcpConfig` is called at `execenv.go:638` and `:926` with a `manifest` already in scope.
- `recordWriteFile` implements refuse-to-clobber via `errPathPreExists`, so the hand-rolled `os.Stat` block is not needed.
- Cleanup runs from the daemon at task teardown, which survives in-process failures (this is what fixes 2.2).
- `context.go:387` already writes Pi files into `{workDir}/.pi/skills/`, so `mcp.json` alongside it is the natural home.
- `envlock` / `gitroot_lock` provide concurrency protection that a direct write from the backend bypasses.

There is one more consequence worth calling out: `sidecar_manifest.go` documents a contract that `local_directory` runs are byte-exactly reversible. A file written outside the manifest breaks that — and for `local_directory` resources the working directory is the user's own folder, so anything left behind is left in their repo.

---

## Smaller points

- The `runtime_mcp.go` change's actual effect is re-marshalling the agent config into the canonical envelope; that does not match what the comment describes. It also has a side effect: with `supported=true`, top-level keys other than `mcpServers` are now dropped, whereas `supported=false` passed the config through untouched. Worth rewriting together with #1.
- `hasManagedMcpConfig(opts.McpConfig)` is evaluated twice, the first time only to reach the `opts.Cwd == ""` check. Folding these into one `if` would read better.
- On non-Windows, `hardenBrowserMcpConfig` returns the input unchanged, so `mcpTempDir` creates an empty directory for nothing. Consider creating it lazily.
- Failing the whole task when `opts.Cwd == ""` is harsh for a defensive branch that should not be reachable; skipping with a `logger.Warn` seems safer.
- **UI inconsistency:** the tab now renders, but its "runtime-inherited MCP servers" section will show `runtime_unsupported` ("This runtime version cannot report its MCP inventory yet") because `listRuntimeLocalMcpServers` (`runtime_mcp.go:377`) was not extended to `pi` / `omp`. That is a confusing first impression for exactly the users this PR is for — worth including.
- The checklist item about syncing landing copy and docs for a new runtime / UI tab is unchecked; please confirm whether anything needs updating.

---

## Suggested direction

One coherent change addresses all four: **move the materialisation into `execenv`, write `.pi/mcp.json` / `.omp/mcp.json` through `sidecarManifest`, and add `.pi` / `.omp` to `agentGitExcludePatterns`.** Please also verify against real `pi` and `omp` binaries and record the versions — that verification gates everything else.

Happy to look again once these are addressed.

feiguoL:
@Bohan-J 

Addressed the requested changes in commit `3c9b60edf`.

- Pi now uses `.pi/mcp.json`; Oh My Pi uses `.omp/mcp.json`.
- MCP materialization moved into `execenv` and is tracked by `sidecarManifest`, including collision protection and daemon teardown/reuse cleanup.
- Added `.pi` and `.omp` to repo-cache git excludes to protect MCP credentials and runtime sidecars from commits.
- Added Pi/OMP runtime MCP inventory support for their native global config paths.
- Added regression coverage for provider-native paths, cleanup, collisions, inventory, git excludes, and UI tab visibility.

Verified installed runtime versions during investigation: `pi 0.84.3` and `omp 17.2.9`. The installed upstream Pi README documents MCP as extension-based rather than core-built-in; OMP supports project MCP configuration. The implementation follows the provider-native paths requested in the review and reports native global inventories where available.

Focused tests pass for execenv, daemon, repocache, core MCP support, and AgentOverviewPane. The unrelated broad local Go run still has pre-existing fixture failures in repocache branch-name reuse and `TestPiExecuteAttachesStdinPipe`; no production path in this PR caused those failures. The PR CI run before this review update had passed backend checks.

Bohan-J (changes requested):
Thanks for the rework — this revision is substantially better than the first one. Three of the four blocking points are cleanly resolved:

- **Layering** — the write moved out of `server/pkg/agent/pi.go` into `execenv/pi_mcp.go`, reusing `recordWriteFile`'s refuse-to-clobber (`errPathPreExists`) instead of a hand-rolled `os.Stat`, and cleanup now runs from the daemon via `CleanupSidecars` rather than an in-process `defer`. That also removes the "crashed run poisons every future task" failure mode.
- **Credential exposure** — `.pi` / `.omp` added to `agentGitExcludePatterns` with a regression guard in `cache_test.go`. The `local_directory` path is covered too, since the file is manifest-recorded and `CleanupSidecars` runs before Finalize.
- **Collision** — moving into the runtime-owned namespace makes the fail-closed path essentially unreachable in practice.

Dropping the `supported=true` shortcut in `loadRuntimeMcpServerConfigs` was also the right call.

I verified the remaining question against real binaries rather than docs this time.

## Verification method

I placed a probe MCP server in a scratch project — a shell script that appends a marker file on launch and then behaves like a dead stdio server. Each CLI was then run with `-p` and a deliberately invalid API key, so the session initialises fully and only the model request fails. If the CLI reads the config, it spawns the command and the marker appears; no model call is needed for the result to be meaningful.

Versions tested: `pi` **0.67.2** (`@mariozechner/pi-coding-agent`) and `omp` **17.2.12**.

## `omp` 17.2.12 — confirmed working ✅

`<cwd>/.omp/mcp.json` is read and its servers are launched. The marker was written, with the scratch project as cwd:

```
1787894506 launched cwd=.../mcp-verify
```

This matches OMP's own changelog, which lists the OMP-owned config locations explicitly:

> set it in an OMP-owned config (`.omp/mcp.json`, `~/.omp/agent/mcp.json`, a project `mcp.json`/`.mcp.json`, or an OMP plugin)

The path table in the binary agrees: project level is `<cwd>/.omp/mcp.json` (plus `.omp/.mcp.json` for compatibility), user level is `~/.omp/agent/mcp.json`. Both the path this PR writes and the new `omp` case in `listRuntimeLocalMcpServers` are correct.

## `pi` 0.67.2 — does not work ❌

The same probe placed at `.pi/mcp.json` was never launched — no marker file, even though the session reached the model request and returned 401, so initialisation definitely completed.

Four independent lines of evidence point the same way:

1. Pi's README states this as an explicit design decision, in the Philosophy section:

   > **No MCP.** Build CLI tools with READMEs (see Skills), or build an extension that adds MCP support.

2. The published `dist/` of `@mariozechner/pi-coding-agent@0.67.2` contains no MCP code at all. The only `mcp` match in the whole bundle is `mcpy` — a language name inside a vendored highlight.js list.
3. `pi --help` exposes no MCP subcommand and no MCP flag.
4. A real `~/.pi/agent/` contains only `auth.json`, `bin/`, `sessions/`, and `settings.json`.

MCP with Pi is only possible through a user-authored extension in `~/.pi/agent/extensions/` or `.pi/extensions/`, and those extensions define their own configuration format. There is no `.pi/mcp.json` convention for Multica to write into.

---

## Remaining blocker: drop `pi` from this PR

Four places need it removed:

- the `"pi"` entry in `packages/core/agents/mcp-support.ts`
- the `provider != "pi"` half of the guard in `preparePiMcpConfig`
- the new `case "pi"` in `listRuntimeLocalMcpServers`
- the `["Pi", "pi"]` rows in the two frontend tests

The `listRuntimeLocalMcpServers` entry deserves a specific note: it returns `supported=true` and reads `~/.pi/agent/mcp.json`, a file Pi never creates or consults. The MCP tab would therefore render "No MCP servers were found in the runtime's user configuration" — which implies the feature works and merely has nothing configured yet. That reads as more trustworthy than the honest "unsupported" state, so it is actively misleading.

This is also what the allowlist's own header comment is guarding against:

> The MCP config tab is hidden for every other provider so a user can't save a value the runtime will silently ignore.

Shipping `omp` alone still closes most of #7558. It would be worth stating in the PR description why Pi is excluded — citing upstream's "No MCP." stance is sufficient, and the reporter asked about both runtimes.

## Should-fix: `Reuse` swallows the error while Cursor's equivalent bails

At `execenv.go:901`, a `preparePiMcpConfig` failure is only logged and execution continues. The Cursor equivalent at `:988` logs and then `return nil`, forcing the environment to be rebuilt.

The normal path is fine — `CleanupSidecars` has already run, so there is nothing to collide with. But if a prior task died between writing the file and persisting the manifest, or if cleanup failed, the leftover `.omp/mcp.json` hits `errPathPreExists`, the warning is swallowed, and **the task silently runs on the previous run's MCP configuration** — a removed server or a rotated token would not take effect. Because the write failed, the file is not in the new manifest either, so this run's cleanup will not remove it and the staleness persists across subsequent runs. Matching Cursor's behaviour here would be safer.

## Nits

- The first revision called `hardenBrowserMcpConfig`; the rewrite dropped it. On Windows, a Playwright MCP server configured for an OMP agent will not get the hardening other runtimes receive. Not a regression for a new feature, but it disappeared during the rewrite, so worth confirming it was intentional.
- OMP's inherited-server inventory is incomplete. Beyond `~/.omp/agent/mcp.json`, OMP also discovers servers from `.claude.json`, `.cursor/mcp.json`, `.vscode/mcp.json`, a project-root `mcp.json`/`.mcp.json`, and plugin `.mcp.json` files, while `listRuntimeLocalMcpServers` reads only the first. I checked the related semantics: with a project `.omp/mcp.json` of `{"mcpServers":{}}`, a user-level server still starts — so clearing every server in the UI does not mean "this agent has no MCP". Comparable to the simplification other providers already make, so not blocking, but the inventory's scope is worth a comment.

## Test results

- `go test ./internal/daemon/execenv/ -run "ByteIdentical|PiMcp" -count=1` → ok
- `go test ./internal/daemon/ -run "Mcp|MCP" -count=1` → ok
- `go test ./internal/daemon/repocache/ -run Exclude -count=1` → ok
- Full `go test ./internal/daemon/execenv/` fails on four Hermes tests (`TestHermesMemoryStorePathLayout`, `TestPruneHermesMemoryStores`, `TestHermesSessionStorePathLayout`, `TestPruneHermesSessionStores`). **These fail identically on `origin/main`**, so they are unrelated to this PR — flagging only because `main` appears to be red there and may deserve its own issue.
- Frontend vitest not run (dependencies not installed in my environment); the frontend diff is allowlist and test-table entries only.

Requesting changes for the `pi` removal. Once that is out, this looks good to merge.

feiguoL:
@Bohan-J 
Thanks for the thorough review. All requested changes are addressed in the latest push (rebased onto `main`):

**Blocker — dropped `pi`:**
- Removed `"pi"` from `MCP_SUPPORTED_PROVIDERS` in `packages/core/agents/mcp-support.ts`
- Removed `case "pi"` from `listRuntimeLocalMcpServers` in `runtime_mcp.go`
- Narrowed the `preparePiMcpConfig` guard to `omp` only
- Removed `["Pi", "pi"]` rows from both frontend tests and the Go test tables
- Updated the PR description explaining why Pi is excluded (upstream's explicit "No MCP" design decision, verified against pi 0.67.2)

**Should-fix — `Reuse` error handling:**
- `preparePiMcpConfig` failure in `Reuse` now `return nil` (forcing fresh Prepare), matching Cursor's behaviour at `:988`. Prevents stale MCP config from persisting when a leftover `.omp/mcp.json` hits `errPathPreExists`.

**Nits:**
- Added a comment on the `omp` case in `listRuntimeLocalMcpServers` documenting that the inventory reads only `~/.omp/agent/mcp.json` (user-scope) and not the full precedence chain (`.claude.json`, `.cursor/mcp.json`, etc.), matching the simplification other providers already make.
- `hardenBrowserMcpConfig`: the rewrite was intentional. OMP materializes its config through the `execenv` path (`preparePiMcpConfig` → `recordWriteFile`), not through `writeMcpConfigToTemp`. The Playwright hardening in `writeMcpConfigToTemp` is specific to Claude's temp-file flow and does not apply to the execenv sidecar model. If we later want browser-MCP hardening for OMP, it would go into `preparePiMcpConfig`, but that's a separate enhancement.

All Go and frontend tests pass locally.

feiguoL:
@Bohan-J

Rebased again from a forced fetch onto the latest `upsteam/main` (`d6ecf4bc8`) in a dedicated git worktree and force-updated the PR branch.

The requested changes are included in `cedba412e`: Pi MCP support remains removed; OMP writes to `.omp/mcp.json` through `sidecarManifest`; `.omp` is git-excluded; `Reuse` returns nil on materialization failure; OMP inventory scope is documented; and regression tests are included.

The new CI run is fully green: backend, frontend, frontend views, Windows execenv, mobile, installers, SQLC, and Go vulnerability scan all pass. Please take another look when convenient.

feiguoL:
@Bohan-J The rebased update is now pushed at `8ed2750a7d3b77303a8b72bdb3b83c7d3aab1010`, based on upstream `d06b6b6e5` (`main`).

The latest CI is fully green: CI run `33149272035` and Mobile Verify run `33149272067`; merge state is `CLEAN`. I also renamed the helper/test files and symbols from `pi_mcp` to `omp_mcp` so the implementation clearly reflects that only OMP remains supported. Please re-review when convenient.

Bohan-J:
Thanks for sticking with this through three rounds — the PR ended up in a genuinely good place.

Worth calling out what made that happen: when the verification showed Pi ships without MCP support, you dropped the Pi half rather than arguing for it. That is the harder call to make on your own PR, and it is why this can merge today instead of shipping a config surface that silently did nothing. The move to `execenv` + `sidecarManifest` and the git-exclude work are both solid, and adding `.pi` to `agentGitExcludePatterns` closes a pre-existing gap that had nothing to do with your change — `.pi/skills/` was already being written into task workdirs unexcluded.

## Two commits pushed directly

To save another round-trip I pushed the last review items myself (maintainer edit on your branch):

- **`27099fdd`** — restores `expect(providerSupportsMcpConfig("pi")).toBe(false)`. Removing `"pi"` from the allowlist was correct, but the negative assertion that had pinned it went with it, so `main` would have lost a guard it previously had. It now carries a comment citing upstream's "No MCP." stance, so the next person to consider adding Pi finds the reasoning rather than an unexplained gap.
- **`7f6581d6`** — three cosmetic items: the inventory-scope comment moved inside the `case "omp":` it documents (above the case it read as a trailing note on the OpenClaw branch), the leftover single-element `for range []string{"omp"}` tables flattened, and two log/error messages made static since the provider guard means only `omp` reaches them.

Nothing behavioural. All Go and frontend tests pass locally, and CI is green on `7f6581d6`.

## Title change

I retitled the PR to say Oh-My-Pi rather than "Pi runtimes". The repository squash-merges, so the PR title becomes the commit message on `main`, and "support MCP config for Pi runtimes" would have been inaccurate in the history for exactly the reason this PR spent two rounds establishing.

## What ships

Oh-My-Pi (`omp`) agents get an MCP tab, and the managed config is materialised as `<workDir>/.omp/mcp.json` — Oh-My-Pi's highest-precedence project-level path.

Pi itself is deliberately excluded. Upstream states the position directly in its README:

> **No MCP.** Build CLI tools with READMEs, or build an extension that adds MCP support.

That was confirmed against pi 0.67.2: no MCP code in the published bundle, no MCP flag or subcommand, and a probe server placed in a project config was never launched. So #7558 is only half-answered by this PR. The remaining half would mean authoring a Pi extension that the daemon injects, which is a separate piece of work rather than something to bolt onto this change.

Merging now. Thanks again.

Bohan-J (approved):
All review items resolved. Verified against real omp 17.2.12; CI green on 7f6581d6.

PR #7570, opened 2026-08-26
feat(skills): add Import from local to the New skill dialog (MUL-6703)

### LINKED ISSUE #5658: [Feature]: Allow creating skills by importing local files
### Deployment type

Official App

### What do you want and why?

新建skill时，可否增加直接导入本地文件的方式

_No response_

_No response_

longer96:
## What does this PR do?

Adds a fourth **New skill** method, **Import from local**, so a workspace member can pick a skill folder on their computer (or a `.skill` / `.zip` archive) and import it into the workspace.

The CLI already supports `multica skill import --file`. This wires the same `POST /api/skills/import` archive path into the shared Skills UI (web + desktop). The selected folder is packed client-side into a STORE zip and uploaded as multipart form data.

## Related Issue

Closes #5658

## Type of Change

- [x] Documentation update

## Changes Made

- New skill chooser now has four methods, in order: Create manually, **Import from local**, Import from URL, Copy from runtime.
- Clicking **Import from local** opens a native folder picker. After a folder is chosen, the dialog shows name / description / file count, then **Import**.
- Footer actions: Cancel, **Choose .skill / .zip**, **Choose folder**, Import.
- Client packs the folder (`SKILL.md` + supporting files) with a dependency-free STORE zip helper and uploads via `api.importSkillArchive`.
- Snapshot `input.files` before resetting the file input. Chromium's `FileList` is live, so clearing `value` was dropping the selection and making Import a no-op.
- i18n for `en` / `zh-Hans` / `ja` / `ko`.
- Skills and tutorial docs updated to describe the fourth method.

Key files:

- `packages/views/skills/components/create-skill-dialog.tsx`
- `packages/core/skills/pack-archive.ts`
- `packages/core/api/client.ts`
- `apps/docs/content/docs/skills.mdx` (and locale variants)

## How to Test

1. Open the workspace **Skills** page → **New skill**. Confirm **Import from local** is the second card.
2. Click it, pick a folder that contains `SKILL.md` (and optional supporting files). Confirm the preview shows the skill name, then click **Import**. The skill appears in the list with its files.
3. Pick a folder without `SKILL.md`. Confirm an error is shown and nothing is created.
4. Use **Choose .skill / .zip** to import a packaged archive.
5. Import the same name twice. Confirm the existing-name conflict message.
6. Repeat on web and desktop.

## Checklist

## Screenshots (optional)

image

image

## Thinking path

The backend already accepts `.skill` / `.zip` uploads on `POST /api/skills/import`. The gap in #5658 is UI-only: the New skill chooser had Create manually, Import from URL, and Copy from runtime. Reusing the archive import path avoids a new API. The dialog lives in `@multica/views`, so one folder picker (`input[webkitdirectory]`) covers web and Electron without new desktop IPC. Packing STORE zip client-side keeps the server parser as the source of truth for `SKILL.md` layout, ignore rules, and size caps.


packages/core/skills/pack-archive.ts
```diff
@@ -0,0 +1,368 @@
...
+      displayName: wrapperName || skillName,
+      skillName,
+      description,
+      fileCount: packed.length,
+      source: "folder",
+    },
+  };
+}
+
+export async function entriesFromFileList(
```
Bohan-J: This reads every file the directory picker returned into memory before anything is filtered. `webkitdirectory` hands back the whole subtree, so a user pointing at a normal project directory pulls in `node_modules`, `dist`, videos and archives — every one of them gets an `arrayBuffer()` here, and all the bytes are retained in `out` at the same time.

Every guard that would have rejected those files — `isIgnoredArchiveEntry`, `isLikelyBinaryFilePath`, `MAX_SKILL_FILE_BYTES`, `MAX_SKILL_BUNDLE_BYTES`, `MAX_SKILL_FILE_COUNT` — runs later, in `prepareSkillArchiveFromEntries`, on data that has already been read. The `too_large` error can only be shown after the memory has already been spent, and on a large enough directory the tab will be unresponsive or die before it gets there.

Suggested shape: decide what to pack from path plus `File.size` alone, which needs no I/O. Locate the `SKILL.md` root, drop ignored and binary paths, drop files over `MAX_SKILL_FILE_BYTES`, accumulate count and total size against their limits, and only then `arrayBuffer()` the survivors.

Please add a test asserting that `arrayBuffer()` is never called for an ignored or oversized entry — a stub `File` whose `arrayBuffer` throws makes that assertion direct.

packages/views/skills/components/create-skill-dialog.tsx
```diff
@@ -433,12 +628,80 @@ export function CreateSkillDialog({
...
+      // staying on the chooser with no way to pick a .skill file.
+      setMethod("local");
+      openFolderPicker();
+      return;
+    }
+    resetLocal();
+    setMethod(next);
+  };
+
+  const onFolderPicked = async (e: ChangeEvent<HTMLInputElement>) => {
```
Bohan-J, line 685: This prepare has no notion of "latest selection", so a stale result can win.

Pick folder A (large, slow). While it is preparing, the header back button is live — `method` is `"local"`, so it renders and nothing disables it. Go back, re-enter local, pick folder B (small, fast). B resolves and sets `localPrepared`; then A resolves and overwrites it. The dialog now previews A while the user believes they chose B, and importing sends A's archive.

`resetLocal()` has the same hole: it clears the state but cannot cancel work already in flight, so the pending prepare writes back over the reset. `onArchivePicked` sets state synchronously and can be clobbered the same way.

Suggested fix: keep a `useRef` generation counter, bump it in `resetLocal` and at the top of each picker handler, capture it locally, and only call `setLocalPreparing` / `setLocalPrepared` when the captured value still matches the current one.

A regression test with two deferred promises resolved in reverse order — B first, then A — pins the behavior.

packages/core/api/client.ts
```diff
@@ -580,6 +581,32 @@ export class PreviewUnsupportedError extends Error {
...
+  }
+  const body = err.body as { reason?: unknown; error?: unknown };
+  const reason = typeof body.reason === "string" && body.reason ? body.reason : "";
+  const error = typeof body.error === "string" && body.error ? body.error : "";
+  const message = reason || error;
+  if (!message || message === err.message) return err;
+  return new ApiError(message, err.status, err.statusText, err.body);
+}
+
+function skillFromImportResult(raw: unknown, endpoint: string): Skill {
```
Bohan-J, line 597: `raw` is cast straight to `SkillImportResult`. `CLAUDE.md` § API Compatibility rules this out: "Parse API JSON with `parseWithFallback` in `packages/core/api/schema.ts` and a zod schema. Do not cast network JSON to `T`", and "When adding or changing an endpoint, add/update the schema and include a malformed-response test."

The nested `result.skill` is parsed correctly, but the envelope around it is not. `status`, `reason` and `existing_skill` are read as trusted `SkillImportResult` fields with no validation, so if the server later renames `status`, returns it as a non-string, or moves `reason`, this silently falls through to `throw new Error(result.reason || "Import failed")` and an installed desktop build shows "Import failed" for what was actually a successful import.

Please add a `SkillImportResultSchema` covering the outer `status` / `reason` / `skill` / `existing_skill` shape, run the response through `parseWithFallback` before branching on `status`, and add a malformed-response test alongside the existing cases in `client.test.ts`.

Bohan-J (changes requested):
Thanks for this, and welcome — a solid first contribution. The product direction is right: it closes the gap between what the CLI can already do and what the **New skill** dialog offers, and it's what #5658 asked for. Reusing the existing import endpoint, keeping one shared entry point for web and desktop, writing STORE zips that the server's `parseSkillArchive` already accepts, not overwriting on name conflict, and shipping all four locales plus targeted tests are all the right calls.

Three things need to change before this can merge — details inline.

1. **Unbounded local file read before any limit applies** (`packages/core/skills/pack-archive.ts:141`). `entriesFromFileList` calls `arrayBuffer()` on *every* file the picker returns and holds all the bytes at once. The ignore rules, per-file size, total size and file-count checks all live in `prepareSkillArchiveFromEntries`, which only runs after everything is already in memory. Pick a real project folder containing `node_modules`, a build output or a video and the page reads hundreds of MB before it can tell you it's too large.

2. **A slower earlier selection can overwrite a newer one** (`packages/views/skills/components/create-skill-dialog.tsx:676`). `onFolderPicked` is async with no request identity, and `resetLocal()` does not invalidate work already in flight.

3. **API response is cast, not parsed** (`packages/core/api/client.ts:596`). `CLAUDE.md` § API Compatibility is explicit: "Do not cast network JSON to `T`", endpoint responses consumed by UI logic must pass through a schema, and a new endpoint needs a malformed-response test. Only the nested `skill` goes through `parseWithFallback` today.

Two non-blocking notes:

- `packages/views/skills/components/create-skill-dialog.tsx:455` — `LocalForm`'s `error` state is not cleared when a new folder or archive is picked. `LocalForm` has no `key` and does not unmount between selections, so a stale "name already exists" message keeps showing over the new preview until submit clears it.
- `apps/docs/content/docs/tutorial.mdx:280` (and the zh/ja/ko copies) — the prose now says four routes, but `tutorial-skill-new-modal.webp` is unchanged and still shows three. Softening the alt text helps, but the screenshot and the text still disagree. Worth regenerating it here.

One longer-term note, no action needed in this PR: the frontend now carries its own copy of the server's size limits, file count and binary-extension list. Those two sides will drift. A contract test or a clear pointer at both definitions would make the sync point explicit.

CI is green on the current head (`image-budget` conditionally skipped). I reviewed against the head commit's source rather than running the suite locally.

Bohan-J (approved):
Thank you for this — and for how you handled the review. All three blocking items came back properly fixed rather than papered over, and each one landed with a test that actually discriminates: the `arrayBuffer()` stub that throws for ignored, oversized and binary entries, and the two deferred promises resolved in reverse order. That is the hard part of both fixes, and you got it right. You also caught the two non-blocking notes without being asked, including regenerating the tutorial screenshot.

I re-verified all of it against `238b86d29` before approving:

- `selectSkillArchiveMembers` decides membership from path + `File.size` with no I/O, so the read is bounded by the 1 MiB / 8 MiB / 256-file caps.
- The generation guard covers all three `setState` paths, and `resetLocal` / `onArchivePicked` invalidate work in flight.
- The envelope goes through `parseWithFallback` before `status` is branched on, with a malformed-response test.

I pushed two small follow-ups myself rather than sending you another round for nits:

- `a21aa40b2` — the envelope's `status` was a `z.enum`, which fails the *whole* parse on a value this client does not know. Since the server field is a bare `string`, a future status would have dropped the server's `reason` and shown a generic "Import failed". It parses as a string now; `skillFromImportResult` already had the default branch. Added a test that fails on the old code with exactly that symptom.
- `fc2cc2827` — the ignore rules and size/count limits were implemented twice, once in `selectSkillArchiveMembers` and once in `prepareSkillArchiveFromEntries`. They agreed, but a one-sided edit would have silently desynced the preview's file count from what actually got packed. Selection happens once now and both prepare paths share one packer.

Local verification of the merged result: core suite 1657/1657, views dialog tests 8/8, typecheck and ESLint clean on both packages.

One thing I deliberately left alone, for a separate issue rather than this PR: `MAX_SKILL_ARCHIVE_BYTES` is checked against `file.size`, while the server's `MaxBytesReader` bounds the whole multipart body — so a file sitting right at 16 MiB passes the client precheck and gets a 400. That is the pre-existing server contract, not something this PR introduced.

Great first contribution. Merging.

Issue #7563, opened 2026-08-26
[Bug]: Squad plan task scans cannot advance tasks due to assignment permissions


ligl007:
### Deployment type

Official App

### Area

Web app

### What happened?

<img width="992" height="385" alt="Image" src="https://github.com/user-attachments/assets/ea730aee-4e97-4f0a-be2f-ff60955692ba" />

<img width="1342" height="914" alt="Image" src="https://github.com/user-attachments/assets/d1d0ebdd-039e-4edd-8e41-80aa5d6017de" />

<img width="1647" height="466" alt="Image" src="https://github.com/user-attachments/assets/5acb9f1a-62fe-4435-b723-86f6c15358f1" />

<img width="1500" height="612" alt="Image" src="https://github.com/user-attachments/assets/9dedaa1d-c790-46be-9b77-2ca3e1873a07" />

### Steps to reproduce

1.后续流转暂停：平台权限拒绝将 DRA-109 指派给 ProjectLeaderAgent-llz 或 llz-team，因此未创建无法正确指派的详细需求子 task
2.DRA-109/DRA-110任务未有负责人
3..DRA-109按照设计应属于ProjectLeaderAgent-llz总任务，Leader分析后根据情况创建子任务分配给应处理的成员Agent
期望是：同一个 Squad 内，Leader 给自己、给本 Squad、给本 Squad 成员 Agent 创建/指派任务，应自动允许，即使这些 Agent 是 private。

### Additional context (optional)

```shell

```

multica-eve:
Fixed by #7572.

PR #7572, opened 2026-08-26
MUL-6691: fix(issues): let verified autopilot runs assign private agents

multica-eve:
Fixes the reported Squad flow in GH #7563 / MUL-6691: a `run_only` autopilot leader finishes its scan, creates DRA-109/DRA-110, and then cannot assign them to `ProjectLeaderAgent-llz` or `llz-team` — the assignment answers 403 and the plan stalls with nobody on the work.

> Updated after review (commit 2): three permission-boundary must-fixes applied — see “Review round 1” at the bottom.

## Background

A schedule/webhook autopilot run has no top-of-chain human originator by design (MUL-4302: `originator_user_id` stays NULL, only the audit-accountable side is set). `canInvokeAgent` judges an agent actor by that originator, so a private agent — the default — is refused. MUL-4857 restored the autopilot's authority for exactly one action, creating a child under the autopilot's own issue, which left two gaps:

- `run_only` has no autopilot issue at all, so a top-level `issue create` had nothing to bind against;
- assigning an **already-created** issue goes through `UpdateIssue`, which passed `nil` for the delegation scope, so even the `create_issue` lineage MUL-4857 accepts was refused there.

The second one is the reported symptom, and it is mode-independent: an identical actor, task and issue are admitted when creating a child but refused when changing the assignee.

## What changed

`validateAssigneePair` now takes an `assignAuthorityScope` — an issue plus an explicit **kind** (`none` / `childOf` / `existingIssue` / `newTopLevelIssue`) — instead of a bare parent issue. The kind matters because those shapes do not admit the same fallbacks. Authority for an unattributed agent request resolves in order:

1. the real top-of-chain originator (unchanged);
2. the speaking task's **own accountable human** (new) — available to all three agent-reachable kinds;
3. the autopilot's member creator (the existing MUL-4857 rule) — **only** for `childOf`, the one surface where it already shipped.

(2) is preferred over (3) because since MUL-4302 the accountable human is the `trigger_owner` — whoever last shaped the firing trigger — which need not be the autopilot's creator. Keying authorization on the creator would let the audit trail name one human while the private-agent / OAuth access came from another.

(3) is deliberately *not* extended to the new surfaces: it performs no liveness or attribution check of its own, so allowing it on the assign verb would let a completed, `owner_fallback`, `delegation`, or entirely unstamped task assign private agents whenever the autopilot's creator happened to have rights. It is kept where it is only so a leader task that predates attribution stamping does not lose the child-creation path it has today.

Wired in at:

| entry point | before | after |
| --- | --- | --- |
| `CreateIssue`, with parent | parent issue | `childOf(parent)` — unchanged behaviour |
| `CreateIssue`, parentless | nothing | `newTopLevelIssue` — verified `run_only` task only |
| `UpdateIssue` | `nil` | `existingIssue(this)` |
| `BatchUpdateIssues` | `nil` | `existingIssue(this iteration's row)` |
| `source_context.go` ×2 | `nil` | `none` — member-only entry points |

## Security bounds

Every one of these must hold, and anything missing, mismatched or unreadable returns no authority and leaves the gate fail-closed:

- the actor **is** the speaking task's agent (server-trusted `X-Task-ID`);
- the task is still `running` — on both branches, so a finished run never lends authority over HTTP;
- the task carries no real originator (one is never silently replaced by this coarser value);
- `originator_source` is `trigger_owner` or `rule_owner`. This **is** the autopilot-lineage proof, since only the autopilot enqueue paths stamp those labels. `owner_fallback` is excluded because that value is the *agent owner* — honouring it would bypass the very white-list `canInvokeAgent` enforces; `delegation` is excluded so authority does not propagate to descendant runs;
- the assignment is bound to work the run verifiably owns: `task.issue_id == issue.id` for `create_issue`, or `issue.origin_id == task.id` **plus** independently verified `run_only` lineage for a run_only-created issue. Both halves are needed: without the second, a create_issue task could escape its same-issue bound by creating a fresh top-level issue, which stamps its own task id. Binding on the creating **task** rather than the autopilot also stops two concurrent runs of one autopilot from borrowing on each other's issues, and needs no migration or backfill;
- the issue is in the request's workspace, and the task's autopilot run resolves to an autopilot in that workspace;
- the accountable human is **still** a member of that workspace.

Explicitly **not** implemented: "same squad implies mutual authority". An admin can put any workspace agent in a squad while not being able to invoke other people's private agents, so treating membership as authorization would reopen a cross-owner OAuth/Composio path.

The resolved id is used for authorization only — the new issue and any task it enqueues are attributed separately and stay unattributed, so "no human directly started this" survives in the audit trail. No schema migration, no API change.

## Testing

`server/internal/handler/issue_autopilot_run_only_assignment_test.go`, 31 subtests through the real handlers.

Positive: `run_only` top-level create dispatching a private agent and a private-leader squad exactly once and unattributed; a sub-issue under an issue the run created; the reported create-then-assign sequence; a squad assigned by update; `create_issue` leader assigning the issue it works on via an explicit accountable stamp while the autopilot's creator has no rights; batch updating a bound issue.

Negative: task not `running` (create, assign, child creation and batch); `owner_fallback`; `delegation`; no attribution stamp; accountable without rights on the target; accountable not a workspace member; actor ≠ task agent; no autopilot run behind the task; run belonging to another workspace; a real originator without rights not being replaced; assigning an issue the run did not create; assigning an issue created by a **sibling** run of the same autopilot; the create_issue laundering path refused on the assign verb, on child creation and in a batch; a foreign issue in a mixed batch left untouched; a plain member still refused. Refused creates are asserted to leave no issue row and no queued task behind.

```
go test ./internal/handler ./internal/service ./internal/attribution ./cmd/server -count=1
ok  github.com/multica-ai/multica/server/internal/handler
ok  github.com/multica-ai/multica/server/internal/service
ok  github.com/multica-ai/multica/server/internal/attribution
ok  github.com/multica-ai/multica/server/cmd/server
```

Run against a scratch database migrated to HEAD (the local dev database lags the checkout, so its stale schema fails the suite wholesale before any of this code runs).

## Review round 1

1. **`agent_create` binding did not prove `run_only` lineage** — a `create_issue` task could launder authority through a fresh top-level issue. Fixed: that branch now also requires the task's own verified autopilot run. Regressions added for the assign verb, child creation and batch.
2. **The creator fallback re-admitted requests the strict rule had declined** — and it checks neither liveness nor attribution. Fixed: confined to `childOf`, plus a `running` requirement on the request path for both branches. The comment-replay resolvers are untouched, keeping historical-task semantics separate from request-path authorization.
3. **`BatchUpdateIssues` is agent-reachable**, not member-only: a task token authenticates as its bound workspace member (`auth.go` sets `X-User-ID` alongside `X-Agent-ID` / `X-Task-ID`), satisfying `requireUserID` while `resolveActor` still classifies the caller as an agent. The comment claiming otherwise was wrong and is corrected; batch coverage added.

## Review focus

- whether excluding `delegation` is the boundary we want (a descendant run of a `run_only` leader currently cannot assign private agents, by design);
- keeping the MUL-4857 creator fallback at all, now that it is confined to child creation and requires a live task;
- the `run_only` binding using `origin_type=agent_create` + `origin_id=<creating task id>` + verified run lineage, which reuses provenance `CreateIssue` already writes (MUL-4305) instead of adding an autopilot stamp.

### DIFF
--- a/server/internal/handler/agent_access.go
+++ b/server/internal/handler/agent_access.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/multica-ai/multica/server/internal/attribution"
 	"github.com/multica-ai/multica/server/internal/util"
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 )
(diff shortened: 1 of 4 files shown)


server/internal/handler/agent_access.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
```
classicsassy34-ops, line 6: ok

classicsassy34-ops (approved):

PR #7573, opened 2026-08-26
MUL-6698: feat(agents): make conversation starters discoverable from the chat they run in

Bohan-J:
## What this changes

The **Suggested first actions** setting named its content but never its surface — you could fill it in and have no idea where it takes effect. From the other side, the chat empty state gave no hint that those three buttons were configuration at all, or that the three you see unconfigured are defaults. The link between the two existed only in the head of whoever built the feature.

This closes the gap from both ends instead of explaining it.

### 1. Rename + copy that names the surface

`Suggested first actions` → **Conversation starters**, in all four locales, with the hint rewritten to say where and when it shows: *"Up to three examples, shown when someone opens a new chat with this agent. Selecting one fills the composer without sending it."*

`Chat Mode Quick Action` was considered and rejected. **Quick Action** is already three other things in this product (chat follow-up suggestions, the issue sidebar section, a settings section), and it promises execution — selecting a starter only prefills the composer. **Mode** isn't a concept here; Chat is a nav destination alongside Issues, so it would send people hunting for a mode switch. `Conversation starters` matches the existing `starter_prompts` field name and the convention users already know from GPT builders.

### 2. An edit entry point where the setting takes effect

Anyone who can edit the agent now gets a quiet **Customize starters** link under the buttons in the chat empty state, deep-linking to that agent's Instructions tab with the editor scrolled into view and briefly ringed (`?view=instructions&focus=starter_prompts`, consumed and stripped on arrival). Readers see nothing — no disabled tease in the DOM.

This is also what answers "are these three the defaults?" without a label: an author sees three suggestions plus a way to change them, and the meaning is self-evident.

### 3. A preview where the setting is configured

The editor now renders the real empty state below it — through the *same component and the same resolver* the chat uses, so the preview can't drift from what it previews. When nothing valid is configured it shows the built-in defaults and says so: *"These are the built-in defaults. Add a suggestion above to replace all three."*

## Notes

- The permission rule is one pure function (`canCustomizeStarterPrompts`) shared by both chat surfaces — the full page and the floating window render the empty state from separate call sites and would otherwise drift. No link for a reader, for an archived agent, or on a backend that drops `starter_prompts` (its Instructions tab renders no editor to land on).
- API field `starter_prompts` is unchanged — this is a copy and UI change only.
- Deliberately **not** in this PR: generating per-agent defaults from the agent's identity instead of the three generic fallbacks. That's a real improvement and a separate piece of work.
- Mobile's chat empty state renders the same feature but keeps its own inlined copy and no i18n; left alone rather than touched without its pre-flight. Its behavior is unchanged.

## Verification

- `pnpm typecheck` — 7/7 pass
- `pnpm lint` — 0 errors
- `pnpm test` — 5/5 packages, 628 files, ~7,240 tests pass
- New coverage: the selection/permission matrix as pure tests in `packages/core/agents/starter-prompts.test.ts`; the customize link's presence and href; the editor preview across configured / default / half-filled states; the deep-link focus (scroll, param strip, and a regression test pinning a flash timer that would otherwise be cancelled by the effect's own cleanup and leave the ring on permanently).

Not visually verified in a running app — no screenshots in this PR.


Bohan-J:
Both must-fixes are in — `fa49070e1`.

**Must-fix 1 — repeat/cross-agent deep links.** Confirmed real, and worse than described: besides not re-focusing, the second click left the `focus` param stranded in the URL because nothing consumed it.

The latch now re-arms when the param disappears (which is exactly when the `replace` has landed) and is keyed by `agent.id` rather than being a bare boolean, so a link aimed at a different agent is never swallowed. The original purpose of the guard — stopping a lagging URL update from looping, since the navigation adapter isn't referentially stable — is unchanged: while the param is still present for the same agent, the effect still short-circuits. A repeat focus also clears the previous flash timer before starting a new one, so it gets a full window instead of inheriting the old one's remainder.

Two regression tests added: `focus → param cleared → focus again`, and an in-place navigation to a second agent while the param is present. I verified both fail against the previous latch before fixing.

**Must-fix 2 — localized docs.** Correct, and my miss: I grepped for the English string, which by construction couldn't match the translated files. All six updated to their shipped UI names (`对话开场建议` / `会話の開始例` / `대화 시작 예시`) across the overview table, the configuration section and the duplication paragraph, plus the English additions they were missing — where it takes effect, the editor preview, and the Customize link. A scan for the old names across `apps/docs`, `packages/views`, `apps/mobile` and `server/` now comes back empty.

**Verification:** `pnpm typecheck` 7/7, `pnpm lint` 0 errors, `pnpm test` 5/5 packages (628 files, 7,239 tests). Still no runtime visual verification, as noted before.

On the non-blocking point: agreed that mobile's empty state deserves the same author entry point. Its own pre-flight process makes it a separate change rather than a late addition here.

PR #7582, opened 2026-08-26
perf(web): preload landing hero background (MUL-6711)

### LINKED ISSUE #7581: perf(web): preload landing hero background
## Problem

On a cold load of https://multica.ai/, the landing Hero can remain on its black fallback for a noticeable interval before landing-bg.webp appears.

## Root cause

The background is above the fold, but its Next Image currently uses default lazy loading. The product preview is already prioritized, so the background request is discovered later.

Live evidence with cache disabled:
- Product preview: start 456.7 ms, end 1062.2 ms
- Background artwork: start 1184.6 ms, end 2051.5 ms

## Proposed change

Add preload to landing-bg.webp with an explanatory comment. Keep ProductImage priority and the Hero fallback color unchanged.

## Controlled benchmark

Local production build at 1280x900, DPR 2, cache disabled, 150 ms latency, 200 KB/s download; one warm-up plus 3 measured runs. Values are medians.

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Background request start | 2391.8 ms | 260.9 ms | 2130.9 ms earlier (-89.1%) |
| Background response end | 5853.2 ms | 4476.2 ms | 1377.0 ms earlier (-23.5%) |
| FCP to background complete | 3421.5 ms | 2082.8 ms | 1338.7 ms shorter (-39.1%) |
| FCP | 2428 ms | 2352 ms | 76 ms earlier |

One before run did not finish the background within the 8-second window; all three after runs did. Afterward, the background and existing product preview enter the preload window together.

## Acceptance criteria

- Background is preloaded with an explanatory comment
- ProductImage is unchanged
- Hero fallback color is unchanged
- Typecheck, lint, and production build pass
- Change is limited to the landing Hero component

Bbbtt04:
Closes #7581

## Summary

Preload the above-the-fold landing background and document why it is critical. This keeps the existing ProductImage priority behavior and Hero fallback color unchanged.

## Root cause

landing-bg.webp used Next Image default lazy loading, so its request was discovered after the already-prioritized product preview. On cold loads this prolonged the interval where the Hero showed its black fallback.

Live cache-disabled evidence before the change:
- Product preview: request start 456.7 ms, response end 1062.2 ms
- Background artwork: request start 1184.6 ms, response end 2051.5 ms

## Before / after

Controlled local production build: 1280x900, DPR 2, cache disabled, 150 ms latency, 200 KB/s download, one warm-up plus 3 measured runs. Values are medians.

| Metric | Before | After | Effect |
| --- | ---: | ---: | ---: |
| Background request start | 2391.8 ms | 260.9 ms | 2130.9 ms earlier (-89.1%) |
| Background response end | 5853.2 ms | 4476.2 ms | 1377.0 ms earlier (-23.5%) |
| FCP to background complete | 3421.5 ms | 2082.8 ms | 1338.7 ms shorter (-39.1%) |
| FCP | 2428 ms | 2352 ms | 76 ms earlier |

One before run did not finish loading the background within the 8-second observation window; all three after runs did. The final HTML contains preload links for both background and product artwork, so they can load in parallel.

## Scope

- One component file
- Two added lines: explanatory comment and preload prop
- No ProductImage changes
- No fallback color changes

## Verification

- pnpm --filter @multica/web typecheck
- pnpm --filter @multica/web lint (passes; one unrelated existing exhaustive-deps warning)
- pnpm --filter @multica/web build
- Production-mode browser A/B under fixed throttling
- Verified both image preload links are present in rendered HTML

### DIFF
--- a/apps/web/features/landing/components/landing-hero.tsx
+++ b/apps/web/features/landing/components/landing-hero.tsx
@@ -111,6 +111,7 @@ function LandingBackdrop() {
         fill
         preload
         className="object-cover object-center"
+        sizes="100vw"
       />
     </div>
   );
--- a/apps/web/features/landing/components/landing-hero.tsx
+++ b/apps/web/features/landing/components/landing-hero.tsx
@@ -126,7 +126,7 @@ function ProductImage({ alt }: { alt: string }) {
           alt={alt}
           width={2640}
           height={1781}
-          priority
+          preload
           className="block h-auto w-full"
           sizes="(max-width: 1320px) 100vw, 1320px"
           quality={85}
--- a/apps/web/features/landing/components/landing-hero.tsx
+++ b/apps/web/features/landing/components/landing-hero.tsx
@@ -104,10 +104,12 @@ export function LandingHero() {
 function LandingBackdrop() {
   return (
     <div className="pointer-events-none absolute inset-0">
+      {/* This artwork is above the fold, so preload it alongside the product preview. */}
       <Image
         src="/images/landing-bg.webp"
         alt=""
         fill
+        preload
         className="object-cover object-center"
       />
     </div>


Bbbtt04:
Screenshots
背景图加载前：
image
背景图加载后：
image

Bohan-J:
Thanks for the thorough writeup, @Bbbtt04 — the before/after screenshots plus the fixed-throttling A/B made this very easy to evaluate. Using `preload` rather than the now-deprecated `priority` is also the right call for Next 16.

I pushed two small follow-ups to your branch instead of sending you round-trips:

- `sizes="100vw"` on the backdrop. Behaviour is unchanged (`fill` already defaults to `100vw`), but it silences the dev-only *"has `fill` but is missing `sizes`"* warning and matches every other `<Image>` on the landing page.
- Switched `ProductImage` from `priority` to `preload`, so the file uses one spelling.

CI is green on the updated head. Merging now.

One thing worth a follow-up rather than blocking this PR: the backdrop and the product preview are now both fetched at high priority, so on a slow link they share bandwidth. Your numbers cover the backdrop, but not LCP or the product preview's new response-end. If you feel like digging further, a slow-network LCP comparison before/after would be a useful addition — and if the preview turns out to be the LCP element and is measurably delayed, `loading="eager"` on the backdrop is the softer alternative.

Issue #4963, opened 2026-07-06
Assignee changes can silently drop agent tasks; self-reassign cancels the current run immediately


Wilson-G:
## Summary

Two related task-lifecycle bugs can silently drop agent work in active Multica workspaces:

1. **A newer issue trigger cancels an earlier task and does not requeue it**, even when the cancelled task produced nothing and belonged to a different agent / different trigger provenance.
2. **A task that reassigns the issue from inside its own run cancels itself immediately**, so any post-handoff cleanup in that same run is interrupted.

We hit both in a production workspace on **2026-07-06 UTC**. We now have local workarounds, but this feels like a platform-level task lifecycle bug rather than a workspace-governance problem.

## Why this matters

This creates a class of "silent dropped baton" failures:

- the UI/comment history shows that a handoff/mention happened,
- but the actual task row was cancelled,
- no replacement task is enqueued,
- and there is no obvious product signal that work was dropped.

In one busy workspace snapshot from **2026-07-06**, two high-traffic agents had unusually high cancellation rates:

- reviewer-like agent: **138 / 523 = 26.4%** cancelled tasks
- builder-like agent: **229 / 794 = 28.8%** cancelled tasks

Many of those cancellations are benign handoff completions, but the same mechanism also hides genuinely dropped work.

### Minimal repro

1. Create an issue assigned to Agent A.
2. Trigger a run for Agent B on that same issue (for example via an `@mention` comment, or any other valid secondary trigger).
3. Within ~30 seconds, perform another issue-triggering write on the same issue (for example, change the assignee again).
4. Observe that Agent B's task is moved to `cancelled` immediately.
5. Observe that no replacement task is automatically enqueued for Agent B, even if the cancelled task had:
   - `result = NULL`
   - no output comment
   - effectively zero useful work produced

### One concrete occurrence

- Task: `796b32d2-8be3-484d-b1b8-4140c56aac63`
- Started: **2026-07-06T00:38:59Z**
- Cancelled: **2026-07-06T00:39:19Z**
- Trigger comment id: `934832c8-514e-4141-bea6-b6783df6c4f7`
- Final state: `cancelled`, `result = NULL`
- There was no automatic retry/requeue; we had to manually hand work back later.

### Minimal repro

1. Start a task for Agent A on an issue.
2. From inside that run, have Agent A reassign the issue to Agent B as part of a normal handoff.
3. Observe that Agent A's current task is marked `cancelled` immediately.
[...]

PR #7588, opened 2026-08-26
server: park terminally-failed issues as blocked, not false-green todo (DP-2271 / I1121.ENS)

digitalpartnersNL:
## Summary

Terminally failed runs left their issue in a false-green state: `HandleFailedTasks` / `FailTask` reset the issue to `todo`, but the daemon only claims *tasks* — a `todo` issue with no queued task can never be dispatched again while still looking queued on the board (I1121.ENS / DP-2271).

- `HandleFailedTasks`: on terminal failure (no retry child, no other active task), park the issue as `blocked` with an explicit reason and a system comment naming the failed task, instead of a bare `todo` reset that bypasses the HTTP enqueue path.
- `FailTask` (daemon-reported failures): same invariant guard — this is the path that produced ~20 live `in_progress` false-greens.
- New helper `blockIssueAfterTerminalFailure` (best-effort fire-and-forget): status write + `issue:updated` broadcast + system comment; errors are logged, never propagated to the failure path.
- `MaybeRetryFailedTask` errors are no longer silently swallowed (`child, _ :=` → logged with issue/agent context).
- `broadcastFailedTasks` fallback (runtime sweeper path, used by tests): parks as `blocked` too.
- Tests updated to the new contract, including `TestRuntimeReconnectRetryHasBoundedTerminalPath` which asserted the old `todo` behavior.

Rebased onto current `main` — merges cleanly with the newer `issuestatus.Effective` custom-status gate (MUL-6243) and the delegated-failure-recovery machinery: the park now keys off *effective* status, so custom statuses that inherit `in_progress` are parked too.

## Testing

- `go build ./...` green
- `go test ./cmd/server/ -run TestSweep` green
- `go test ./internal/service/ -run 'TestFailTask|TestHandleFailedTasks|TestTaskFinalize'` green
- `TestIssuesCRUDThroughRouter` + `TestWebSocketIntegration` fail identically on clean `main` (verified via stash) — pre-existing, unrelated.

Companion ops-side check (separate repo, not in this PR): `multica-daemon-health.py` gained `check_issue_run_invariant` (alert `multica-issue-run-invariant`) which detects executable issues with only failed/cancelled runs and no active task.

### DIFF
--- a/server/cmd/server/runtime_sweeper.go
+++ b/server/cmd/server/runtime_sweeper.go
@@ -584,7 +584,7 @@ func broadcastFailedTasks(ctx context.Context, queries *db.Queries, taskSvc *ser
 		return
 	}
 	// Fallback path used by tests that don't construct a TaskService:
-	// publish task:failed events with workspace IDs and reset stuck issues.
+	// publish task:failed events with workspace IDs and park stuck issues.
 	processedIssues := make(map[string]bool)
 	affectedAgents := make(map[string]pgtype.UUID)
 	for _, t := range tasks {
@@ -598,17 +598,24 @@ func broadcastFailedTasks(ctx context.Context, queries *db.Queries, taskSvc *ser
 				workspaceID = util.UUIDToString(issue.WorkspaceID)
 				issueKey := util.UUIDToString(t.IssueID)
 				// Only issues whose status means "an agent is actively working"
-				// get reset. in_review and blocked are deliberately excluded —
+				// get parked. in_review and blocked are deliberately excluded —
 				// they mean a human or an external dependency owns the issue
 				// now, and resetting those to todo would re-trigger an agent on
 				// work someone else is holding. A custom status resolves to the
 				// canonical status it inherits, so a custom review gate is
 				// excluded for the same reason In Review is. (MUL-6243)
+				// I1121.ENS: park (never reset to todo) — with no retry service
+				// here a task can never be re-enqueued, so a bare todo flip
+				// would leave a false-green issue with no queued task.
 				effectiveStatus := issuestatus.Effective(ctx, queries, issue.WorkspaceID, issue.Status)
 				if effectiveStatus == "in_progress" && !processedIssues[issueKey] {
 					processedIssues[issueKey] = true
 					if hasActive, herr := queries.HasActiveTaskForIssue(ctx, t.IssueID); herr == nil && !hasActive {
-						queries.UpdateIssueStatus(ctx, db.UpdateIssueStatusParams{ID: t.IssueID, Status: "todo", WorkspaceID: issue.WorkspaceID})
+						queries.UpdateIssueStatus(ctx, db.UpdateIssueStatusParams{ID: t.IssueID, Status: "blocked", WorkspaceID: issue.WorkspaceID})
+						slog.Warn("fallback failed-task path parked issue as blocked",
+							"issue_id", issueKey,
+							"failure_reason", failureReason,
+						)
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


Bohan-J (changes requested):
Thanks for this — and welcome. You found a real product bug, the diagnosis is correct, and the fix is in the right place. We want this change.

**We're accepting the direction; five things need to change before we can merge.**

## Why the direction is right

Issue status exists to answer one question: *what happens next, and who owns it.* After a run fails terminally with the retry budget spent, `todo` claims "queued, the system will run this" and `in_progress` claims "someone is working on this" — both are false. The daemon claims **tasks**, not issues (`HasActiveTaskForIssue` / the claim queries in `server/pkg/db/queries/agent.sql`), and we confirmed there is no sweeper anywhere that picks up a `todo` issue with no queued task. So the old reset really did produce a permanently dead issue that looks alive on the board.

`blocked` is the honest answer, and it matches the definition we already ship to agents in their runtime brief — *"You cannot proceed without something you are missing"* (`server/internal/daemon/execenv/runtime_config_sections.go`). What's missing here is a run that completes.

The deeper issue this exposes — issue status and the task queue being two sources of truth that never cross-check each other — is a bigger design problem, and we're not going to hold this PR for it.

## Required before merge

**1. Only park issues owned by an agent or squad.**

The invariant "an executable issue must have a queued task or an explicit blocked reason" holds only for agent-owned issues. A human-owned `todo` is already truthful — a person doesn't need a queued task to be next up.

`HandleFailedTasks` now includes `todo` but never checks `issue.AssigneeType`, so a stale agent task failing can flip a human-held issue to `blocked`. This is reachable, not theoretical: reassignment deliberately does **not** cancel in-flight tasks (#4963 / MUL-4113), so a task can outlive a handoff to a human and then fail.

Please gate the park on `AssigneeType` being `agent` or `squad` — and apply it to the `in_progress` branch too, so "a failed task only ever moves agent-owned issues" is the single rule rather than a per-branch exception.

**2. `FailTask` should use `issuestatus.Effective`, not a raw comparison.**

The new block in `FailTask` compares `issue.Status` directly, while the other two paths go through `issuestatus.Effective`. That makes the PR description's claim — *"the park now keys off effective status, so custom statuses that inherit in_progress are parked too"* — untrue for exactly the path that produced the ~20 live stuck `in_progress` issues. One-line fix.

**3. Point the recovery instruction at something that actually works.**

The system comment says "Rerun manually to start a fresh attempt." The most natural user action — dragging the issue from Blocked back to Todo — is a **no-op**: `WillEnqueueRun` (`server/internal/service/issue_trigger.go`) only fires on `backlog → active`, not `blocked → active`. So the user drags it to Todo, nothing runs, and they're back to a false-green `todo` — the exact invariant this PR is enforcing, broken by following its own advice.

What does work today is **Retry** in the Execution Log (`packages/views/issues/components/execution-log-section.tsx`), which is gated on task status rather than issue status and is clickable from a blocked issue. Please make the comment name that path explicitly.

**4. Drop the internal ticket identifiers from the code.**

`I1121.ENS` appears 9 times across code comments and test assertions; `DP-2271` is in the PR title and commit message. These are identifiers from your own tracker — future readers of this repo can't resolve them. Please replace them with `MUL-6714`.

While you're in there: the test comment referring to "the queue-manager" describes a component that doesn't exist in this codebase — please drop that sentence.

**5. Collapse the two system comments into one, and rename the stale test.**

A terminal failure now posts two system comments on the same issue: the existing `errMsg` one from `FailTask`, plus the new "Issue blocked: ...". Users see that as duplicate noise. Fold the block reason into the existing comment.

Also `TestSweepResetsInProgressIssueToTodo` now asserts `blocked` — please rename it to match.

## Explicitly accepted, no change needed

Two knock-on effects of writing `blocked` instead of `todo`, both of which we reviewed and consider correct:

- Autopilot-created issues get their run marked failed (`server/internal/service/autopilot.go`) — accurate, the run did fail.
- Delegated subscribers now receive an inbox notification, since `blocked` is in `delegatedStatusNotify` and `todo` isn't (`server/cmd/server/notification_listeners.go`) — that's the point of the change.

Worth calling out in the PR description, though, since neither is currently mentioned.

## On our side

The real fix for #3 isn't wording — it's making `blocked → active` trigger `WillEnqueueRun` for agent-owned issues. That's a product decision on our end, we'll open a separate issue for it, and we're not asking you to carry it. Once it lands, dragging Blocked back to Todo will do what users expect.

CI is green and the change is well-scoped. Happy to re-review as soon as the five items above are addressed.

Issue #7457, opened 2026-08-23
[Feature]: Expose retained task environments and add safe manual cleanup


DearZh:
### Deployment type

Official App

### What do you want and why?

Please make the lifecycle of task environments explicit and add a safe, operator-controlled cleanup path in both the app and CLI.

On Multica v0.4.32 (Official App, macOS arm64), `git worktree list` showed that completed project worktrees had already been detached and each repository only had its original working tree. No task root still contained a `worktree/` directory. However, `multica daemon disk-usage --by-task --output json` reported:

- 60 retained task directories
- about 1.57 GB total
- several completed issue task directories around 50-80 MB each
- most retained content was task state such as `codex-home`, logs, output, config, and GC metadata

From the user's perspective these directories look like stale worktrees, even though the actual Git worktree has already been removed. The CLI exposes disk usage but currently has no `multica daemon gc` or supported per-task cleanup command, and the app does not explain why a completed task directory is retained or when it becomes eligible for removal.

The existing behavior is understandable: the latest `(agent, issue)` environment can be a resume candidate while an issue remains open. The problem is that this retention is invisible and has no explicit operator override. Users cannot answer:

- Is this an active Git worktree, a retained resumable environment, or only a branch?
- Why is it being retained?
- When will automatic GC remove it?
- Can I deliberately discard resumability and reclaim the space now?

This creates confusion around worktree cleanup and makes disk growth feel uncontrolled even when the Git worktree cleanup itself succeeded.

### Proposed solution

Add a task-environment lifecycle view and safe cleanup controls.

#### App

- Show separate states for `active worktree`, `retained resumable environment`, `branch retained`, and `cleanup eligible`.
- Show task/issue, agent, size, completion time, retention reason, and the estimated automatic cleanup time.
- Add cleanup actions for selected completed environments and completed environments older than a chosen age.
- Explain that cleaning a retained resume candidate means the next run will start with a fresh environment/session.

#### CLI

Suggested commands:

```bash
multica daemon gc --dry-run
multica daemon gc --now
multica daemon gc --task <task-id>
multica daemon gc --completed-before <duration>
```

The dry-run output should include the retention reason and what will be deleted.

#### Safety requirements

- Never delete an active, queued, or dispatched task environment.
- Never delete a user-owned `local_directory` working directory.
- Refuse cleanup when task changes have not been safely captured to a branch/commit.
- Preserve task branches unless branch deletion is requested separately.
[...]

PR #7590, opened 2026-08-26
MUL-6717: feat(daemon): explain retained task environments

iiwish:
## Summary

- enrich `multica daemon disk-usage` with task identity, completion/local-directory state, resume-candidate state, stable retention reasons, and estimated full-cleanup eligibility
- source GC timing from the owning daemon effective `/health` policy only after its normalized `workspaces_root` matches the scanned root
- identify resume candidates through one additive, workspace-scoped set-based query that preserves canonical poison, retired-session, and resume-overflow filtering
- keep table and JSON output useful when the daemon, server, credentials, metadata, or newer API fields are unavailable, and document the lifecycle semantics next to the command

## Scope

This implements the maintainer-proposed visibility-only first phase of #7457. It does not add deletion behavior, superseded-environment reclamation, manual GC, or app UI.

## Compatibility and safety

- existing periodic GC requests omit `task_ids`, so their query cost and behavior are unchanged
- older servers return no resume-candidate entries; older or stopped daemons, root mismatches, and post-start profile drift leave policy-based estimates explicitly unavailable
- task-environment lookup remains workspace-scoped and resolves up to 500 distinct `(agent, issue)` scopes in one database call
- the set-based candidate selection mirrors `GetLastTaskSession` semantics for poisoned latest states, explicit session retirement, and session-less resume-overflow cutoffs
- a candidate is reported only when the managed workdir still exists locally; in-place local directories and finalized disposable local worktrees are explicitly non-resumable
- missing completion metadata reports `local_directory: null` instead of incorrectly claiming `false`
- estimates describe full-directory eligibility only; active-root guards and the GC scan interval remain authoritative

## Tests

- isolated PostgreSQL 17, all migrations applied
- full handler integration package: `go test ./internal/handler -count=1`
- canonical selection equivalence across healthy, poisoned, retired, and resume-overflow scopes
- maximum batch coverage with 500 distinct task scopes
- root-ownership regression coverage for `MULTICA_WORKSPACES_ROOT` override and profile configuration drift after daemon startup
- focused race checks for health/root policy and handler/database integration
- full daemon package serialized: `go test -parallel 1 ./internal/daemon -count=1`
- guarded agent package: `go test -p 2 -parallel 2 ./pkg/agent/...`
- `go build ./...`
- `go vet ./internal/daemon ./internal/handler ./cmd/multica ./pkg/db/generated`
- `pnpm typecheck`
- `make sqlc` (byte-identical generated diff) and `sqlc vet`
- `git diff --check`

The standard fully parallel Go suite was also exercised. Two unrelated daemon subprocess/deadline tests timed out under host contention; both passed immediately in isolation, and the complete daemon package passed with `-parallel 1`.

Addresses the first phase of #7457.


multica-eve (changes requested):
Today, operators can see how much space old task environments consume but not why they remain or when they may be cleaned up. This change adds that explanation so operators can avoid deleting a resumable environment and know when waiting for automatic cleanup is appropriate. The user-facing direction, including an issue-task-only first phase, is worth accepting.

Requesting changes for two correctness/performance blockers:

1. **The reported GC policy is not proven to own the scanned root.**

   `server/cmd/multica/cmd_daemon.go:1808-1812` disables estimates only for an explicit `--workspaces-root`. The normally resolved root can still come from the CLI process's `MULTICA_WORKSPACES_ROOT`, or from profile configuration changed after the daemon started. `diskUsageGCPolicy` at `server/cmd/multica/cmd_daemon.go:1945-1966` validates the daemon port/profile but never compares the daemon's actual managed root.

   For example:

   ```bash
   MULTICA_WORKSPACES_ROOT=/some/other/root multica daemon disk-usage --by-task
   ```

   can scan `/some/other/root` while applying the default profile daemon's policy, producing `cleanup_eligible` or an estimated cleanup time for a directory that daemon will never scan. Please expose the daemon's normalized workspaces root through health and require it to match the selected scan root before applying estimates. Add regression coverage for an environment override and for profile configuration drifting after daemon startup.

2. **The batch diagnostic introduces an N+1 query path with a worst case of 500 expensive session lookups.**

   The endpoint accepts up to 500 `task_ids`. After the initial bulk subject lookup, `server/internal/handler/daemon.go:5166-5186` calls `GetLastTaskSession` once for every distinct `(agent, issue)` pair. That query performs the full canonical session selection, including poison/retired-session filtering, so one valid CLI invocation can cause one bulk lookup plus 500 database round trips.

   Please replace the loop with a set-based query that returns the canonical session for all requested scopes in a constant number of database calls. Lowering the limit or running the point queries concurrently would only hide the amplification. The bulk query must preserve all existing poisoned, retired, and overflow filtering semantics; please add multi-scope large-batch coverage and an equivalence test against the single-scope selection rules.

The additive response fields and local diagnostic behavior are otherwise compatible, and the change does not alter deletion policy itself. I verified this head (`30de7599`) with `git diff --check`, the targeted daemon/CLI tests, and the `ByteIdentical` prompt-prefix guard. Please address both blockers and request re-review.

iiwish:
@multica-eve Thank you for the precise review. Both requested blockers are addressed in `9efe9d270` and the branch is rebased onto current `main`.

1. **GC policy ownership**
   - `/health` now exposes the running daemon's normalized `workspaces_root`.
   - `daemon disk-usage` applies that daemon's GC policy only when the health root matches the root actually scanned. Missing fields, environment overrides, and profile configuration drift all fail closed to `policy_unavailable`.
   - Regression tests cover both `MULTICA_WORKSPACES_ROOT` overriding a running daemon and profile configuration changing after daemon startup.

2. **N+1 canonical-session lookup**
   - The handler no longer calls `GetLastTaskSession` per scope. `ListIssueTaskEnvironmentSubjects` now resolves all requested subjects and their canonical workdirs in one set-based database query, so task diagnostics stay at one DB call for up to 500 task IDs.
   - The query mirrors the single-scope poisoned-latest-state, retired-session, legacy error-text, authentication, and session-less resume-overflow filters.
   - Database integration tests compare bulk results directly against `GetLastTaskSession` across healthy, poisoned, retired, and overflow scopes, plus exercise the maximum 500 distinct scopes.

Verification includes an isolated migrated PostgreSQL 17 database, full handler integration tests, focused race tests, serialized full daemon tests, guarded agent tests, `go build ./...`, `go vet`, full `pnpm typecheck`, `sqlc vet`, deterministic `make sqlc`, and `git diff --check`. All GitHub CI checks are green, including backend, Windows execenv, sqlc, frontend, vulnerability scan, and installers.

Re-review requested.

iiwish:
@multica-eve Follow-up review hardening is pushed in `9a668f781` (test-only; no production behavior changed).

I expanded the set-based SQL equivalence integration test from four representative scopes to 21 explicit cases. It now asserts the expected canonical workdir as well as comparing against `GetLastTaskSession`, covering:

- every resume-unsafe `failure_reason`;
- all legacy text guards and their required-marker negative boundaries;
- benign failed and cancelled sessions;
- a poisoned session that later recovers;
- both the overflow cutoff and a fresh session established after overflow.

Fresh local verification used a fully migrated PostgreSQL 17/pgvector database: the complete handler suite passed, the 500-distinct-scope endpoint test passed, and focused handler/daemon/CLI race tests passed. I also ran the full daemon and CLI packages, `go build ./...`, affected-package `go vet`, deterministic `make sqlc`, `sqlc vet`, uncached workspace typechecking, Windows test-binary compilation for daemon/CLI, Linux handler compilation, and `git diff --check`.

All current GitHub checks are green, including the 7-minute backend suite, sqlc, Windows execenv, frontend build/tests, vulnerability scan, and all installers. Re-review requested on head `9a668f781`.

iiwish:
@multica-eve A gentle follow-up on the requested re-review. Both blockers from your review are addressed on the current head: daemon GC policy is used only when the running daemon normalized workspaces root matches the scanned root, and the diagnostic now resolves all requested scopes with one set-based query while preserving the canonical session filters. The follow-up commit expands the SQL equivalence coverage across the poison, retirement, legacy-text, and overflow boundaries; current CI is green. When convenient, could you please re-review and clear the outstanding changes request if the fixes look right? Thanks.

iiwish:
Hi @multica-eve, just a gentle check-in on the re-review when you have a moment.

The branch has been kept cleanly synced with upstream main (`a363d26`) and CI remains fully green. Both blockers from the initial review (verifying daemon normalized workspaces_root ownership in `/health` before applying GC estimates, and replacing the N+1 `GetLastTaskSession` lookups with a single set-based query) along with the expanded SQL equivalence test suite are in place.

Whenever convenient, could you please take another look and dismiss/clear the outstanding changes request if the fixes look good? Thank you!

iiwish:
Re-review evidence refreshed on 2026-09-11 for head `a363d2694`.

I rechecked both original blockers against the current code and a conflict-free prospective merge with `main` at `d44932089` (local verification merge `4a467b61`, not pushed):

- The policy probe compares the scanned root to the running daemon's normalized `workspaces_root` before applying GC estimates. The environment-override and post-start profile-drift regressions pass.
- Resume-candidate inspection uses one set-based task-subject query, not per-task `GetLastTaskSession` calls. Its poison, retired-session and overflow predicates still match current main.
- A freshly migrated, isolated PostgreSQL 17/pgvector database passed the targeted GC-check handler tests, including all 21 canonical-session equivalence cases, 500 distinct task scopes, invalid-input limits and custom-status category handling.
- Focused CLI/daemon race tests passed for disk-usage, root resolution, retention estimates, parent-status resolution and health. The prospective merge passes `git diff --check`.

No additional blocking finding or necessary production-code optimization emerged from this re-review, so the PR head is unchanged. Full repository CI was not rerun locally. @multica-eve, please re-review the current head when convenient.

iiwish:
Updated to `d502752ea8f7` against main `7d9613bd2c44`. Retention diagnostics now use the same category-first lifecycle interpretation as GC, including custom done/closed statuses, active-category precedence, and conservative unknown-status handling. This does not change deletion policy.

Added a regression covering the actual status-resolution-to-retention path (confirmed failing before the fix). Guarded daemon/CLI suites, focused database-backed handler tests (21 session-selection cases and 500 scopes), focused race tests, native/Windows vet, and pinned sqlc regeneration passed. Database checks used an isolated migrated PostgreSQL container, now removed. [CI completed successfully](https://github.com/multica-ai/multica/actions/runs/34924611446).

The earlier running-daemon workspace-root validation and set-based batch inspection remain intact. The PR is conflict-free; the existing re-review request remains pending.

iiwish:
Synced with main `8c4f4328f`; current head is `6f2cfd405`.

Resolved the health-test import conflict while retaining both the normalized workspaces-root/GC-policy assertions and the upstream terminal-report queue health coverage. No new retention or deletion policy is introduced; the existing root-ownership check and set-based task-environment query remain intact.

Verification: guarded full daemon and CLI suites; focused health/disk-usage/retention race tests; native and Windows-target vet; deterministic pinned sqlc regeneration; whitespace checks all pass. [Fresh CI](https://github.com/multica-ai/multica/actions/runs/35482843726) passes all applicable checks, including database-backed backend tests. Database integration tests were not rerun locally for this import-only conflict resolution.

GitHub reports no merge conflicts. The existing changes-requested review remains pending; no review was dismissed and no merge was performed.

PR #7599, opened 2026-08-26
MUL-6721: perf(runtime): restore HOT heartbeats and index the stale-task sweep

### LINKED ISSUE #7389: Heartbeat batching defeats HOT updates + unindexed sweeper queries cause seq-scan/autovacuum overload at scale
### Summary
At fleet sizes well beyond the original design assumption (~70 online runtimes), the `agent_runtime` heartbeat batching and background sweepers generate enough write amplification and seq-scan volume to degrade overall database performance (CPU/ACU pressure, slower request latency across unrelated endpoints).

### Root cause 1: `TouchAgentRuntimesLastSeenBatch` update is unconditionally non-HOT
`server/pkg/db/queries/runtime.sql:172-184`:
```sql
UPDATE agent_runtime SET last_seen_at = now() WHERE id = ANY(@ids::uuid[]) AND status = 'online';
```
The comment states this deliberately avoids touching `updated_at` "so the rows stay HOT-eligible." However, migration `115_agent_runtime_last_seen_at_index.up.sql` added a btree index directly on `last_seen_at`:
```sql
CREATE INDEX CONCURRENTLY idx_agent_runtime_last_seen_at ON agent_runtime (last_seen_at);
```
Postgres HOT updates require that *no indexed column* be modified — since `last_seen_at` is itself indexed, every batch flush update is non-HOT regardless of `updated_at`. In production we observed 379.6M cumulative updates on this table with only 0.76% HOT, and 337,181 autovacuum cycles on a table that is otherwise only 22MB / ~18k rows.

At current scale (~18,000 concurrently-online runtimes vs. the ~70 the batching design comment assumes), every 30s batch tick rewrites thousands of rows and touches all 7 indexes on the table per row, driving continuous autovacuum churn and CPU/ACU pressure on the whole cluster.

**Suggested fix:** either drop the plain index on `last_seen_at` and serve the two queries it was built for (`runtime_online`, `runtime_heartbeat_age`) some other way, or accept non-HOT updates but reduce write frequency/fanout (e.g., lean more on the existing Redis liveness cache and flush to Postgres less often as fleet size grows).

### Root cause 2: sweeper queries lack covering indexes for status-only predicates
`runtime_sweeper.go` runs every 30s and includes:
- `FailStaleTasks` (`queries/agent.sql:1199`) — `OR`s across `status='dispatched'` and `status='running'` on `agent_task_queue`. Neither existing partial index (`idx_agent_task_queue_dispatched_prepare`, `idx_agent_task_queue_running_started_at`) matches, because the dispatched branch never states `started_at IS NULL` and the `OR` spans two indexes Postgres can't combine — this forces a full seq scan of the queue table every tick.
- `SelectStaleOnlineRuntimes`, `CountStaleOfflineRuntimesBlockedByTasks`, `ListStaleOfflineRuntimeGCCandidates`, and the runtime-side join in `FailTasksForOfflineRuntimes` all filter `agent_runtime.status` alone. The only status index (`idx_agent_runtime_status`, migration 004) is workspace-scoped and unusable for these workspace-agnostic reads.

At 2,880 ticks/day this fully accounts for the observed seq-scan counts on both tables (158,719 on `agent_task_queue`, ~6.5M on `agent_runtime` over ~410 days).

**Suggested fix:**
- Add `agent_runtime(status, last_seen_at)` (or a partial index on `status IN ('online','offline')`) to cover the sweeper reads.
- Split `FailStaleTasks` into two status-specific statements (or add the missing `started_at IS NULL` predicate) so each branch can use its existing partial index.

### Suggested follow-up
Re-examine the "~70 online runtimes" design assumption noted in `daemon.go`'s `BatchedHeartbeatScheduler` comment — real-world fleets are now ~250x that size, and the batch interval/strategy may need to scale with fleet size (e.g. staggered/sharded flushes) rather than a fixed 30s global tick.

### Environment where this was observed
Aurora PostgreSQL Serverless v2 (engine 16.11, min 0.5 / max 8.0 ACU), ~18,000 concurrently online `agent_runtime` rows across ~60 workspaces, correlating with sustained CPU spikes to 100% and ACU scaling to the 8.0 ceiling.

importcpp:
## What does this PR do?

At fleet scale the `agent_runtime` heartbeat path and the stale-task sweep were the two biggest DB hot spots described in #7389.

**Write amplification.** The batch heartbeat updates only `last_seen_at` and deliberately leaves `updated_at` untouched to keep the row HOT-eligible. That intent never held: migration 115 put a btree index directly on `last_seen_at`, and HOT requires that *no indexed column* change. So every batch flush was non-HOT and rewrote all of the table's indexes, driving continuous write amplification and autovacuum churn on an otherwise tiny table.

**Sequential scans.** `FailStaleTasks` OR'd the `dispatched` and `running` branches in one `UPDATE`. The planner cannot combine the two status-specific partial indexes across an `OR`, so every sweep tick fell back to a full sequential scan of `agent_task_queue`.

This PR fixes both without dropping the `last_seen_at` column or adding a compatibility shim — the column stays, only the index is removed, so there is no multi-version migration dance.

## Related Issue

Closes #7389

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/migrations/432_drop_agent_runtime_last_seen_at_index.{up,down}.sql` — drop `idx_agent_runtime_last_seen_at` (`CREATE INDEX CONCURRENTLY` on rollback). Single-statement per repo convention.
- `server/migrations/433_agent_runtime_fillfactor.{up,down}.sql` — lower `agent_runtime` fillfactor to 90 so heartbeat updates keep page space for HOT chains.
- `server/cmd/migrate/main.go` — register `432` in `concurrentDownIndexCleanups` (its down direction rebuilds the index).
- `server/pkg/db/queries/agent.sql` — split `FailStaleTasks` into `FailStaleDispatchedTasks` (adds `started_at IS NULL` so it matches `idx_agent_task_queue_dispatched_reclaim_v2`) and `FailStaleRunningTasks` (matches `idx_agent_task_queue_running_started_at`).
- `server/pkg/db/generated/agent.sql.go` — regenerated via `make sqlc`.
- `server/cmd/server/runtime_sweeper.go` — `sweepStaleTasks` now runs both queries and unions the results.
- `server/cmd/server/runtime_sweeper_test.go` — test helper that runs both split queries so existing assertions keep testing the union.

## How to Test

Verified on a fresh `pgvector/pgvector:pg17` instance:

1. `go run ./cmd/migrate up` applies cleanly through `433`; after `432`, `last_seen_at` is referenced by **zero** indexes (`SELECT ... FROM pg_index ...`).
2. HOT recovery under a heartbeat workload (150k updates over 5000 online rows, VACUUM between ticks): **0% HOT with the index present → ~72% HOT after removal** on the real migrated schema (`pg_stat_user_tables.n_tup_hot_upd / n_tup_upd`).
3. `EXPLAIN` on the split queries: each uses its partial index (Index Scan) instead of the previous full Seq Scan; `FailStaleTasks` OR version was ~11.6ms → ~0.33ms for the two split statements on 200k rows.
4. `go build ./...`, `gofmt`, `go vet`, `go test ./cmd/migrate/...` (incl. `TestEveryConcurrentUpBuildHasCleanup`), and DB-backed `go test ./cmd/server/... ./pkg/taskfailure/...` all pass.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (N/A — backend only)
- [x] I have updated relevant documentation to reflect my changes (none required)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and docs (N/A)
- [ ] If this PR touches Chinese product copy, I checked it against conventions.zh.mdx (N/A — no product copy)

### Thinking path

Started from #7389's two root causes → confirmed both still present in `runtime.sql` / `agent.sql` against the current `main` → measured on pg17 that any index on `last_seen_at` (including a `(status, last_seen_at)` composite) forces HOT to 0%, so the fix is removal, not a replacement index → confirmed dropping the column is unnecessary (the index, not the column, breaks HOT), which keeps this to two migrations with no multi-version compatibility work → split the OR'd sweep query so each branch hits its existing partial index.

### Risk notes

- Dropping `idx_agent_runtime_last_seen_at` makes the two sampler reads (`runtime_online`, `runtime_heartbeat_age`) scan the table. On a compact ~18k-row table behind an 8s cache this is sub-2ms and acceptable; a covering index cannot be reintroduced without re-breaking HOT.
- `FailStaleDispatchedTasks` adds `started_at IS NULL`. A `dispatched` row has not started, so this never narrows results in practice; it is required for the partial-index match.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** Used Claude Code to triage #7389 against a local checkout, reproduce the write-amplification and seq-scan behavior on a local PostgreSQL 17 instance (measuring HOT ratios and `EXPLAIN` plans before/after), design the minimal fix (drop the index rather than the column to avoid a multi-version migration), and implement + verify the migrations, query split, and tests.

### DIFF
--- a/server/cmd/migrate/main.go
+++ b/server/cmd/migrate/main.go
@@ -293,6 +293,7 @@ var concurrentDownIndexCleanups = map[string]string{
 	"371_comment_content_search_index_strategy":             "idx_comment_content_trgm",
 	"375_drop_issue_last_activity_index":                    "idx_issue_workspace_last_activity",
 	"391_drop_agent_task_queue_dispatched_prepare_index":    "idx_agent_task_queue_dispatched_prepare",
+	"432_drop_agent_runtime_last_seen_at_index":             "idx_agent_runtime_last_seen_at",
 }
 
 var preMigrationHooks = func() map[string]preMigrationHook {
--- a/server/cmd/server/runtime_sweeper.go
+++ b/server/cmd/server/runtime_sweeper.go
@@ -506,18 +506,33 @@ func gcRuntime(ctx context.Context, txStarter runtimeGCTxStarter, queries *db.Qu
 // edge where a runtime row lingers online-with-stale-heartbeat past the
 // wall clock (MUL-4107).
 func sweepStaleTasks(ctx context.Context, queries *db.Queries, taskSvc *service.TaskService, bus *events.Bus, reconnectGrace time.Duration) {
-	failedTasks, err := queries.FailStaleTasks(ctx, db.FailStaleTasksParams{
+	// Dispatched and running tasks are failed by two separate statements rather
+	// than one OR'd UPDATE so each can use its own partial index; an OR across
+	// idx_agent_task_queue_dispatched_reclaim_v2 and
+	// idx_agent_task_queue_running_started_at cannot be combined by the planner
+	// and forced a full sequential scan every tick (GH #7389).
+	dispatched, err := queries.FailStaleDispatchedTasks(ctx, db.FailStaleDispatchedTasksParams{
 		DispatchTimeoutSecs: dispatchTimeoutSeconds,
-		RunningTimeoutSecs:  runningTimeoutSeconds,
 		// Reuse the runtime stale window so the running-task backstop
 		// exactly matches what sweepStaleRuntimes considers "not alive".
 		RuntimeStaleSecs:          staleThresholdSeconds,
 		RuntimeReconnectGraceSecs: reconnectGrace.Seconds(),
 	})
 	if err != nil {
-		slog.Warn("task sweeper: failed to clean up stale tasks", "error", err)
+		slog.Warn("task sweeper: failed to clean up stale dispatched tasks", "error", err)
 		return
 	}
+
+	running, err := queries.FailStaleRunningTasks(ctx, db.FailStaleRunningTasksParams{
+		RunningTimeoutSecs:        runningTimeoutSeconds,
+		RuntimeReconnectGraceSecs: reconnectGrace.Seconds(),
+	})
+	if err != nil {
+		slog.Warn("task sweeper: failed to clean up stale running tasks", "error", err)
+		return
+	}
+
+	failedTasks := append(dispatched, running...)
 	if len(failedTasks) == 0 {
 		return
 	}
--- a/server/cmd/server/runtime_sweeper_test.go
+++ b/server/cmd/server/runtime_sweeper_test.go
@@ -12,6 +12,38 @@ import (
 	db "github.com/multica-ai/multica/server/pkg/db/generated"
 )
 
+// failStaleTasksParams mirrors the fields of the former single FailStaleTasks
+// query. Tests use it to exercise both the dispatched and running paths with
+// one call, matching how sweepStaleTasks fans out to the two split queries.
+type failStaleTasksParams struct {
+	DispatchTimeoutSecs       float64
+	RunningTimeoutSecs        float64
+	RuntimeStaleSecs          float64
+	RuntimeReconnectGraceSecs float64
+}
+
+// failStaleTasks runs both split queries and returns their combined result, so
+// existing tests keep asserting on the union of dispatched + running failures
+// exactly as the pre-split FailStaleTasks statement did.
+func failStaleTasks(ctx context.Context, q *db.Queries, p failStaleTasksParams) ([]db.AgentTaskQueue, error) {
+	dispatched, err := q.FailStaleDispatchedTasks(ctx, db.FailStaleDispatchedTasksParams{
+		DispatchTimeoutSecs:       p.DispatchTimeoutSecs,
+		RuntimeStaleSecs:          p.RuntimeStaleSecs,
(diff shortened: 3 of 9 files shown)


multica-eve:
Thanks for investigating this and for putting together the measurements and patch. We independently confirmed that the underlying issues are real and worth fixing.

We’re going to take the implementation in-house because the remaining changes touch migration ordering and the atomic failure semantics of the stale-task side-effect pipeline. We are narrowing the follow-up to:

- dropping `idx_agent_runtime_last_seen_at` so heartbeat updates can use HOT updates again; and
- atomically rewriting the stale-task sweep so the `dispatched` and `running` branches can use their existing partial indexes without allowing a partial commit.

We will not carry forward the `fillfactor` change. The current migration number also conflicts with `main`, and committing the two task updates independently could strand tasks in `failed` if the second update errors before `HandleFailedTasks` runs.

Rather than ask you to rework the PR around these internal invariants, we’re closing it and tracking the implementation internally. We’ll retain attribution to your analysis and this PR in the follow-up. Thank you again for the useful contribution.

PR #7601, opened 2026-08-26
fix(auth): allow fixed login code for loopback self-hosting

axAilotl:
## Summary

- add an explicit fixed verification code for production-built self-host containers
- accept it only when FRONTEND_ORIGIN resolves to localhost, 127.0.0.1, or ::1
- preserve the normal send-code, rate-limit, expiry, and one-time-use flow
- document and pass the setting through Docker Compose

## Root cause

The self-host stack had no SMTP or Resend provider. The API returned success from send-code but wrote generated codes only to backend logs, while the existing development code is intentionally disabled under APP_ENV=production.

## Validation

- targeted auth regression tests pass, including public-origin rejection
- production backend image built successfully
- live localhost send-code, fixed-code verify, auth-cookie, and GET /api/me smoke passed

## Local deployment

The running loopback stack is configured with MULTICA_LOCAL_VERIFICATION_CODE=888888. No UI or plugin code changed.


multica-eve:
Thanks for taking the time to investigate this, implement the change, and validate the self-hosted login flow.

We are going to close this PR because the current `main` branch already supports the same local/private self-hosting use case through the existing fixed-code configuration. In `.env`, an operator can set:

```dotenv
APP_ENV=development
MULTICA_DEV_VERIFICATION_CODE=888888
```

and then restart/recreate the backend container. `MULTICA_DEV_VERIFICATION_CODE` must contain exactly six digits. These are runtime settings, so they also work with the production-built self-host image; rebuilding the image is not necessary. `docker-compose.selfhost.yml` already passes both settings to the backend.

This preserves the rest of the login flow: the user must first request a verification code for the email address, the server must have a latest unexpired verification-code record, and a successful login consumes that record. For one-off local login without a fixed code, a deployment with neither Resend nor SMTP configured also writes the generated code to the backend logs.

The fixed code is deliberately ignored when `APP_ENV=production`. That boundary was introduced intentionally in #1773 so a public production deployment cannot gain a shared master login code. The new `MULTICA_LOCAL_VERIFICATION_CODE` would add a second, parallel fixed-code path that bypasses that decision. Checking whether `FRONTEND_ORIGIN` contains `localhost` or a loopback IP is not sufficient proof that the API is unreachable through a reverse proxy or another network path: that setting describes the browser-facing origin/CORS configuration, not the server's actual exposure or the request source.

The intended configurations are therefore:

- Local/private self-hosting without email: use `APP_ENV=development` plus `MULTICA_DEV_VERIFICATION_CODE`, while still keeping `JWT_SECRET` strong and private.
- Public production deployment: keep `APP_ENV=production`, leave the fixed code empty, and use Resend, SMTP, or Google sign-in.

Because the requested capability is already available and we do not want a production fixed-code exception, we will not merge this additional configuration path. Thank you again for the contribution and the careful validation.

Issue #5921, opened 2026-07-24
[Bug]: Transcript records are no longer being followed


franciscocpg:
### Deployment type

Official App

### What happened?

After 0.4.10, the agent logs are not following the transcription logs anymore. I need to keep scrolling to see the recent logs.

### Steps to reproduce

Create an issue, run an agent and click `View transcription`.

### Additional context (optional)

```shell

```

NevilleQingNY:
Thanks for the clear report, Francisco. This was fixed in https://github.com/multica-ai/multica/pull/5932.

The fix will be included in the next version, which we expect to ship later today. After updating, live transcript records should follow new output again when you're at the live end, while still letting you scroll back without being pulled down automatically.

PR #7612, opened 2026-08-27
MUL-6739 fix(chat): keep the viewport at the live end while replies stream

tim-smart:
## Summary

- keep the chat pinned when streamed content or the composer changes the viewport
- release only after the chat list confirms reader-driven movement
- carry confirmed touch motion through fractional drag frames and momentum
- discard input that a nested scroller consumed before later system movement can inherit it

## Verification

- focused chat and transcript suites: 5 files, 111 tests passed
- `@multica/views` typecheck passed
- targeted ESLint passed

Closes TIM-55

### DIFF
--- a/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
+++ b/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
@@ -64,6 +64,8 @@ interface FakeObserver {
 }
 
 let observers: FakeObserver[] = [];
+let animationFrames = new Map<number, FrameRequestCallback>();
+let nextAnimationFrameId = 1;
 
 function observedTargets(): Element[] {
   return observers.flatMap((o) => o.targets);
@@ -75,8 +77,14 @@ function fireResizeObservers() {
   });
 }
 
-// The follow latch treats input within its intent window as an ongoing
-// gesture and defers pinning; tests control that clock directly.
+function renderFrame() {
+  const callbacks = [...animationFrames.values()];
+  animationFrames.clear();
+  now += 16;
+  for (const callback of callbacks) callback(now);
+}
+
+// Confirmed motion gets a short settle window; tests control that clock.
 let now = 0;
 function gestureSettles() {
   now += 301;
@@ -84,9 +92,19 @@ function gestureSettles() {
 
 beforeEach(() => {
   observers = [];
+  animationFrames = new Map();
+  nextAnimationFrameId = 1;
   reportContentHeightChanged = undefined;
   now = 0;
   vi.spyOn(Date, "now").mockImplementation(() => now);
+  vi.stubGlobal("requestAnimationFrame", (callback: FrameRequestCallback) => {
+    const id = nextAnimationFrameId++;
+    animationFrames.set(id, callback);
+    return id;
+  });
+  vi.stubGlobal("cancelAnimationFrame", (id: number) => {
+    animationFrames.delete(id);
+  });
   vi.stubGlobal(
     "ResizeObserver",
     class {
@@ -130,6 +148,13 @@ interface Scroller {
   shrinkViewport(px: number): void;
   /** Reader wheel input scrolling up by `px`, and the scroll it causes. */
   readerScrollsUp(px: number): void;
+  /** Starts a one-finger touch gesture at `clientY`. */
+  touchStart(clientY: number): void;
+  /** Moves that finger and then scrolls the list up by `scrollPx`. */
+  touchMove(clientY: number, scrollPx: number): void;
+  touchEnd(): void;
+  /** Touch momentum scrolls the list after the finger lifts. */
+  touchMomentumScrollsUp(px: number): void;
   /** Reader wheel input landing `fromBottom` px above the live end. */
   readerScrollsTo(fromBottom: number): void;
   /**
@@ -166,6 +191,19 @@ function scroller(el: HTMLElement, initialContent = 2000): Scroller {
     act(() => {
       el.dispatchEvent(new Event("scroll"));
     });
+  };
+
+  const touchEvent = (type: string, clientY?: number) => {
+    const event = new Event(type, { bubbles: true }) as TouchEvent;
+    const points =
+      clientY === undefined
+        ? []
+        : [{ identifier: 1, clientY } satisfies Pick<Touch, "identifier" | "clientY">];
+    Object.defineProperty(event, "touches", {
+      value: points,
+    });
+    Object.defineProperty(event, "changedTouches", { value: points });
+    return event;
   };
 
   // Reader input: the wheel delta the hook judges intent from, then the
@@ -222,6 +260,27 @@ function scroller(el: HTMLElement, initialContent = 2000): Scroller {
     },
     readerScrollsUp(px) {
       wheelBy(px);
+    },
+    touchStart(clientY) {
+      act(() => {
+        el.dispatchEvent(touchEvent("touchstart", clientY));
+      });
+    },
+    touchMove(clientY, scrollPx) {
+      act(() => {
+        el.dispatchEvent(touchEvent("touchmove", clientY));
+      });
+      state.scrollTop -= scrollPx;
+      scrollEvent();
+    },
+    touchEnd() {
+      act(() => {
+        el.dispatchEvent(touchEvent("touchend"));
+      });
+    },
+    touchMomentumScrollsUp(px) {
+      state.scrollTop -= px;
+      scrollEvent();
     },
     readerScrollsTo(fromBottom) {
       wheelBy(fromBottom - this.distanceFromBottom());
@@ -365,6 +424,46 @@ describe("ChatMessageList auto-scroll", () => {
     expect(scroll.distanceFromBottom()).toBe(0);
   });
 
+  it("releases when touch momentum carries a flick past the threshold", () => {
+    const { scroll, streamChunk } = renderStreamingChat();
+
+    scroll.touchStart(100);
+    scroll.touchMove(180, 80);
+    scroll.touchEnd();
+    scroll.touchMomentumScrollsUp(80);
+    const parked = scroll.scrollTop;
+
+    streamChunk(1);
+    scroll.grow(180);
+
+    expect(scroll.scrollTop).toBe(parked);
+  });
+
+  it("leaves a fractional touch drag in progress", () => {
+    const { scroll, streamChunk, followsAtBottom } = renderStreamingChat();
+
+    scroll.touchStart(100);
+    scroll.touchMove(180, 80.5);
+    streamChunk(1);
+
+    expect(scroll.distanceFromBottom()).toBe(80.5);
+    expect(followsAtBottom()).toBe("auto");
+    scroll.touchEnd();
+  });
+
+  it("resumes pinning after a sub-threshold touch ends", () => {
+    const { scroll, streamChunk } = renderStreamingChat();
+
+    scroll.touchStart(100);
+    scroll.touchMove(180, 80);
+    scroll.touchEnd();
+    gestureSettles();
+    streamChunk(1);
+    scroll.grow(180);
+
+    expect(scroll.distanceFromBottom()).toBe(0);
+  });
+
   it("releases on incremental upward scrolling during a fast stream", () => {
     const { scroll, streamChunk } = renderStreamingChat();
 
@@ -454,6 +553,16 @@ describe("ChatMessageList auto-scroll", () => {
 
     streamChunk(1);
     scroll.grow(180);
+
+    expect(scroll.distanceFromBottom()).toBe(0);
+  });
+
+  it("pins a system shift after nested scrolling left input unconsumed", () => {
+    const { scroll } = renderStreamingChat();
+
+    scroll.wheelWithoutScroll(300);
+    renderFrame();
+    scroll.browserScrollsListUp(200);
 
     expect(scroll.distanceFromBottom()).toBe(0);
   });
--- a/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
+++ b/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
@@ -89,6 +89,7 @@ interface Scroller {
   distanceFromBottom(): number;
   grow(px: number): void;
   shrinkViewport(px: number): void;
+  readerScrollsUp(px: number): void;
   readerScrollsTo(fromBottom: number): void;
 }
 
@@ -129,6 +130,12 @@ function scroller(el: HTMLElement): Scroller {
     shrinkViewport(px) {
       state.viewportHeight -= px;
       resize();
+    },
+    readerScrollsUp(px) {
+      state.scrollTop -= px;
+      act(() => {
+        el.dispatchEvent(new Event("scroll"));
+      });
     },
     readerScrollsTo(fromBottom) {
       state.scrollTop = state.contentHeight - state.viewportHeight - fromBottom;
@@ -223,6 +230,21 @@ describe("ChatMessageList auto-scroll (TIM-55 regression)", () => {
     scroll.shrinkViewport(72);
 
     expect(scroll.scrollTop).toBe(parked);
+  });
+
+  it("releases the pin on incremental upward scrolling during a fast stream", () => {
+    const { scroll, streamChunk } = renderStreamingChat();
+
+    streamChunk(1);
+    scroll.grow(180);
+
+    for (let seq = 2; seq <= 5; seq++) {
+      scroll.readerScrollsUp(60);
+      streamChunk(seq);
+      scroll.grow(180);
+    }
+
+    expect(scroll.distanceFromBottom()).toBe(960);
   });
 
   it("re-engages when the reader scrolls back down to the live end", () => {
--- a/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
+++ b/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
@@ -132,12 +132,21 @@ interface Scroller {
   readerScrollsUp(px: number): void;
   /** Reader wheel input landing `fromBottom` px above the live end. */
   readerScrollsTo(fromBottom: number): void;
+  /**
+   * Wheel input the list never consumed: it bubbles from a nested scroller
+   * (or hits a list with nothing to scroll), so no scroll event follows.
+   */
+  wheelWithoutScroll(px: number): void;
+  /** PageUp bubbling from a focused control inside a row. */
+  pageUpFromRowControl(): void;
+  /** The browser scrolls the list itself (answering a key), no input seen. */
+  browserScrollsListUp(px: number): void;
 }
 
-function scroller(el: HTMLElement): Scroller {
-  const state = { scrollTop: 0, contentHeight: 2000, viewportHeight: VIEWPORT };
+function scroller(el: HTMLElement, initialContent = 2000): Scroller {
+  const state = { scrollTop: 0, contentHeight: initialContent, viewportHeight: VIEWPORT };
   // Open at the bottom, matching Virtuoso's `initialTopMostItemIndex: LAST`.
-  state.scrollTop = state.contentHeight - state.viewportHeight;
+  state.scrollTop = Math.max(0, state.contentHeight - state.viewportHeight);
 
   Object.defineProperties(el, {
     scrollHeight: { configurable: true, get: () => state.contentHeight },
@@ -189,7 +198,7 @@ function scroller(el: HTMLElement): Scroller {
       return state.scrollTop;
     },
     distanceFromBottom() {
-      return state.contentHeight - state.scrollTop - state.viewportHeight;
+      return Math.max(0, state.contentHeight - state.scrollTop - state.viewportHeight);
     },
     grow(px) {
       state.contentHeight += px;
@@ -215,6 +224,24 @@ function scroller(el: HTMLElement): Scroller {
     readerScrollsTo(fromBottom) {
       wheelBy(fromBottom - this.distanceFromBottom());
     },
+    wheelWithoutScroll(px) {
+      const nested = document.createElement("pre");
+      el.appendChild(nested);
+      act(() => {
+        nested.dispatchEvent(new WheelEvent("wheel", { deltaY: -px, bubbles: true }));
+      });
+    },
+    pageUpFromRowControl() {
+      const control = document.createElement("button");
+      el.appendChild(control);
+      act(() => {
+        control.dispatchEvent(new KeyboardEvent("keydown", { key: "PageUp", bubbles: true }));
+      });
+    },
+    browserScrollsListUp(px) {
+      state.scrollTop -= px;
+      scrollEvent();
+    },
   };
 }
 
@@ -222,7 +249,7 @@ function taskMsg(seq: number, content: string): TaskMessagePayload {
   return { task_id: TASK_ID, seq, type: "text", content } as TaskMessagePayload;
 }
 
-function renderStreamingChat() {
+function renderStreamingChat({ contentHeight = 2000 } = {}) {
   const qc = new QueryClient();
   qc.setQueryData(chatKeys.taskMessages(TASK_ID), [taskMsg(0, "Looking into it. ")]);
 
@@ -244,7 +271,7 @@ function renderStreamingChat() {
   return {
     qc,
     view,
-    scroll: scroller(el),
+    scroll: scroller(el, contentHeight),
     rowCount: () => view.container.querySelectorAll("[data-row-key]").length,
     followsAtBottom: () =>
       view.container
@@ -357,6 +384,51 @@ describe("ChatMessageList auto-scroll", () => {
     streamChunk(1);
     scroll.grow(500);
     scroll.shrinkViewport(72);
+
+    expect(scroll.scrollTop).toBe(parked);
+  });
+
+  // Wheel over a capped code block scrolls the block; the event bubbles to
+  // the list without moving it. Unconsumed input must not release the follow.
+  it("keeps following through wheel input the list never consumed", () => {
+    const { scroll, streamChunk } = renderStreamingChat();
+
+    scroll.wheelWithoutScroll(200);
+    gestureSettles();
+
+    streamChunk(1);
+    scroll.grow(180);
+
+    expect(scroll.distanceFromBottom()).toBe(0);
+  });
+
+  it("keeps following after a flick on a conversation too short to scroll", () => {
+    const { scroll, streamChunk } = renderStreamingChat({ contentHeight: 400 });
+
+    scroll.wheelWithoutScroll(200);
+    gestureSettles();
+
+    for (let seq = 1; seq <= 8; seq++) {
+      streamChunk(seq);
+      scroll.grow(180);
+    }
+
+    expect(scroll.distanceFromBottom()).toBe(0);
+  });
+
+  // Focus sits on a row control, the reader presses PageUp, and the browser
+  // scrolls the list. The scroll confirms the staged key intent: the reader
+  // is released, not pinned back over their own keypress.
+  it("honors keyboard paging from a focused row control", () => {
+    const { scroll, streamChunk } = renderStreamingChat();
+
+    scroll.pageUpFromRowControl();
+    scroll.browserScrollsListUp(VIEWPORT);
+    const parked = scroll.scrollTop;
+    gestureSettles();
+
+    streamChunk(1);
+    scroll.grow(180);
 
     expect(scroll.scrollTop).toBe(parked);
   });
--- a/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
+++ b/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
@@ -8,23 +8,7 @@ import type { ReactElement } from "react";
 import enChat from "../../locales/en/chat.json";
 import { ChatMessageList } from "./chat-message-list";
 
-// TIM-55 regression: the chat viewport must stay at the live end while a reply
-// streams in and while the composer grows underneath it.
-//
-// Both failed for the same reason: Virtuoso's `followOutput` reacts to the ITEM
-// COUNT, and neither event changes it. A streaming reply is one row that keeps
-// growing, and the composer only shrinks the viewport. The list therefore drives
-// its own bottom-stick off a ResizeObserver — which is what these tests drive.
-//
-// This suite covers the WIRING and the pin state (release on scroll-up,
-// re-engagement): that the list measures the right boxes and applies the pin
-// to the real scroller. The pure geometry (thresholds, pin targets, shrinking
-// content) is canonical in stick-to-bottom.test.ts and is not re-run through a
-// DOM mount.
-
-// Real react-virtuoso renders no data rows under jsdom's zero-height viewport,
-// and this suite fakes the scroll geometry anyway; the stub keeps the row count
-// visible so the tests can assert that streaming leaves it unchanged.
+// Virtuoso cannot render rows in jsdom's zero-height viewport.
 vi.mock("react-virtuoso", () => ({
   Virtuoso: ({
     data,
@@ -50,8 +34,6 @@ const TASK_ID = "6af44cbe-80ab-4dfe-b07d-bd3cfd588f4d";
 
 const VIEWPORT = 600;
 
-// ─── A ResizeObserver the test can fire ──────────────────────────────────
-
 interface FakeObserver {
   targets: Element[];
   fire: () => void;
@@ -59,12 +41,10 @@ interface FakeObserver {
 
 let observers: FakeObserver[] = [];
 
-/** Every element the list asked to be measured, across all its observers. */
 function observedTargets(): Element[] {
   return observers.flatMap((o) => o.targets);
 }
 
-/** A box changed size: the browser would run every observer watching it. */
 function resize() {
   act(() => {
     for (const observer of observers) observer.fire();
@@ -101,29 +81,19 @@ afterEach(() => {
   vi.unstubAllGlobals();
 });
 
-// ─── Fake scroll geometry ────────────────────────────────────────────────
-//
-// jsdom has no layout, so scrollHeight/clientHeight are hard 0 and scrollTop
-// never moves. This wrapper gives the scroll container the geometry a browser
-// would, and records the scrollTop the component writes.
-
 interface Scroller {
   el: HTMLElement;
   scrollTop: number;
   contentHeight: number;
   viewportHeight: number;
   distanceFromBottom(): number;
-  /** Content grew (a streamed chunk) — the browser then notifies observers. */
   grow(px: number): void;
-  /** The composer grew, taking `px` off the list's height. */
   shrinkViewport(px: number): void;
-  /** The reader scrolls, leaving `fromBottom` px of content below the fold. */
   readerScrollsTo(fromBottom: number): void;
 }
 
 function scroller(el: HTMLElement): Scroller {
   const state = { scrollTop: 0, contentHeight: 2000, viewportHeight: VIEWPORT };
-  // Open at the bottom, matching Virtuoso's `initialTopMostItemIndex: LAST`.
   state.scrollTop = state.contentHeight - state.viewportHeight;
 
   Object.defineProperties(el, {
@@ -169,8 +139,6 @@ function scroller(el: HTMLElement): Scroller {
   };
 }
 
-// ─── Fixture ─────────────────────────────────────────────────────────────
-
 function taskMsg(seq: number, content: string): TaskMessagePayload {
   return { task_id: TASK_ID, seq, type: "text", content } as TaskMessagePayload;
 }
@@ -199,7 +167,6 @@ function renderStreamingChat() {
     view,
     scroll: scroller(el),
     rowCount: () => view.container.querySelectorAll("[data-row-key]").length,
-    /** One more streamed chunk on the SAME live row. */
     streamChunk: (seq: number) => {
       act(() => {
         qc.setQueryData<TaskMessagePayload[]>(
@@ -211,16 +178,14 @@ function renderStreamingChat() {
   };
 }
 
-// ─── Tests ───────────────────────────────────────────────────────────────
-
+// Pure scroll geometry is covered in stick-to-bottom.test.ts.
 describe("ChatMessageList auto-scroll (TIM-55 regression)", () => {
   it("measures both the viewport and the content, not just the viewport", () => {
     const { scroll, view } = renderStreamingChat();
 
     const targets = observedTargets();
     expect(targets).toContain(scroll.el);
-    // A ResizeObserver reports an element's own box, never its scroll extent,
-    // so watching the container alone would miss a reply growing inside it.
+    // ResizeObserver tracks element dimensions, not scroll extent.
     expect(targets.some((t) => t !== scroll.el && scroll.el.contains(t))).toBe(true);
 
     view.unmount();
@@ -235,7 +200,6 @@ describe("ChatMessageList auto-scroll (TIM-55 regression)", () => {
       scroll.grow(180);
     }
 
-    // The whole point: Virtuoso saw no item-count change to follow.
     expect(rowCount()).toBe(rowsBefore);
     expect(scroll.distanceFromBottom()).toBe(0);
   });
@@ -243,7 +207,6 @@ describe("ChatMessageList auto-scroll (TIM-55 regression)", () => {
   it("keeps the newest content clear of a composer that grew", () => {
     const { scroll } = renderStreamingChat();
 
-    // Three lines of draft text push the composer up into the list.
     scroll.shrinkViewport(72);
 
     expect(scroll.distanceFromBottom()).toBe(0);
--- a/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
+++ b/packages/views/chat/components/chat-message-list.autoscroll.test.tsx
@@ -1,0 +1,271 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, render } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { I18nProvider } from "@multica/core/i18n/react";
+import { chatKeys } from "@multica/core/chat/queries";
+import type { TaskMessagePayload } from "@multica/core/types";
+import type { ReactElement } from "react";
+import enChat from "../../locales/en/chat.json";
+import { ChatMessageList } from "./chat-message-list";
+
+// TIM-55 regression: the chat viewport must stay at the live end while a reply
+// streams in and while the composer grows underneath it.
+//
+// Both failed for the same reason: Virtuoso's `followOutput` reacts to the ITEM
+// COUNT, and neither event changes it. A streaming reply is one row that keeps
+// growing, and the composer only shrinks the viewport. The list therefore drives
+// its own bottom-stick off a ResizeObserver — which is what these tests drive.
+//
+// This suite covers the WIRING: that the list measures the right boxes and
+// applies the controller's decision to the real scroller. The decision matrix
+// itself (thresholds, re-engagement, shrinking content) is canonical in
+// stick-to-bottom.test.ts and is not re-run through a DOM mount.
+
+// Real react-virtuoso renders no data rows under jsdom's zero-height viewport,
+// and this suite fakes the scroll geometry anyway; the stub keeps the row count
+// visible so the tests can assert that streaming leaves it unchanged.
+vi.mock("react-virtuoso", () => ({
+  Virtuoso: ({
+    data,
+    itemContent,
+    computeItemKey,
+  }: {
+    data: unknown[];
+    itemContent: (i: number, item: unknown) => ReactElement;
+    computeItemKey: (i: number, item: unknown) => string;
+  }) => (
+    <div data-testid="virtuoso-rows">
+      {data.map((item, i) => (
+        <div key={computeItemKey(i, item)} data-row-key={computeItemKey(i, item)}>
+          {itemContent(i, item)}
+        </div>
+      ))}
+    </div>
+  ),
+}));
+
+const TEST_RESOURCES = { en: { chat: enChat } };
+const TASK_ID = "6af44cbe-80ab-4dfe-b07d-bd3cfd588f4d";
+
+const VIEWPORT = 600;
+
+// ─── A ResizeObserver the test can fire ──────────────────────────────────
+
+interface FakeObserver {
+  targets: Element[];
+  fire: () => void;
+}
+
+let observers: FakeObserver[] = [];
+
+/** Every element the list asked to be measured, across all its observers. */
+function observedTargets(): Element[] {
+  return observers.flatMap((o) => o.targets);
+}
+
+/** A box changed size: the browser would run every observer watching it. */
+function resize() {
+  act(() => {
+    for (const observer of observers) observer.fire();
+  });
+}
+
+beforeEach(() => {
+  observers = [];
+  vi.stubGlobal(
+    "ResizeObserver",
+    class {
+      targets: Element[] = [];
+      constructor(private callback: () => void) {
+        observers.push(this as unknown as FakeObserver);
+      }
+      observe(target: Element) {
+        this.targets.push(target);
+      }
+      unobserve(target: Element) {
+        this.targets = this.targets.filter((t) => t !== target);
+      }
+      disconnect() {
+        this.targets = [];
+        observers = observers.filter((o) => (o as unknown as this) !== this);
+      }
+      fire() {
+        this.callback();
+      }
+    },
+  );
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+// ─── Fake scroll geometry ────────────────────────────────────────────────
+//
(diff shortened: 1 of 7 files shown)


tim-smart:
## Before / after

Captured in headless Chromium at `84432cd8a` and at this PR's merge base `09a2410e8`, mounting the real `ChatMessageList` (real `react-virtuoso`, real styles) and streaming a real task timeline into the task-messages cache the way `useRealtimeSync` does on `task:message`. Same viewport (960×620), same tick count, same content on both sides.

### 1. A reply streaming in

Captured at the same tick on both builds. Before, the newest rows sit under the composer and the status pill is gone; after, the tail and the pill are flush above it.

| Before (`09a2410e8`) | After (`84432cd8a`) |
| --- | --- |
| before: the last row is cut off at the composer edge and the status pill is not visible | after: the newest rows and the Working status pill are fully visible above the composer |

### 2. A multi-line draft in the composer while the reply streams

The reader types five lines, the composer grows, the list's viewport shrinks. Before, the tail stays under the composer for the rest of the stream.

| Before (`09a2410e8`) | After (`84432cd8a`) |
| --- | --- |
| before: with a five-line draft open the last row is clipped and the status pill is hidden | after: with the same five-line draft open the tail and the status pill remain visible |

## What the numbers say

Distance from the bottom (px of content below the fold) sampled after every streamed chunk.

**Streaming, single-line composer.** On the base, Virtuoso does recover on its own — but only once the drift passes `atBottomThreshold`, so the tail oscillates up to the full 120px band instead of staying put:

```
base:  0,0,0,0,0,0,0,0,0,0,0,0,0,9,31,53,75,97,119,0,0,22,44,66,88,110,0,0,22,44,66,88,110,0, …
PR:    0 on every sample
```

That is 5 samples in every 8 with the newest content partly hidden, worst case 119px — roughly the last two rows plus the status pill.

**Composer grows to five lines.** This one never recovers on the base: with `customScrollParent`, the viewport shrink is not corrected, and continued streaming does not clear it either.

| | base | this PR |
| --- | --- | --- |
| after the composer grows | 80px hidden | 0 |
| after 6 more chunks stream in | 66px hidden | 0 |

## Notes on the harness

The composer is a static stand-in for `ChatInput` with the same geometry (`CHAT_GUTTER` + `CHAT_COLUMN`, same card padding), so the list gets the same viewport it has in the app; everything above it is the real component. Nothing in the harness is committed — the screenshots are stored on a branch of the fork so this comment can embed them.

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!scrollEl) return;
+    lastTop.current = scrollEl.scrollTop;
+
+    const onScroll = () => {
+      const top = scrollEl.scrollTop;
+      const movedUp = top < lastTop.current;
+      lastTop.current = top;
+      // A shrinking layout can clamp upward while remaining at the live end.
+      pinned.current = movedUp
+        ? distanceFromBottom(scrollEl) <= 1
```
Bohan-J: **Blocking.** This releases the pin on any upward movement that leaves the viewport more than 1px from the bottom, and there is no path back.

I drove the hook directly in jsdom (only `ResizeObserver` stubbed) to check:

```ts
const { result } = renderHook(() => useStickToBottom(el, el));
// reader is at the live end; trackpad jitter moves up 2px
s.top -= 2;
act(() => { el.dispatchEvent(new Event("scroll")); });
// -> pinned === false, distanceFromBottom === 2

// 10 chunks stream in
for (let i = 0; i < 10; i++) { s.content += 180; act(() => fire.forEach(f => f())); }
// -> pinned === false, distanceFromBottom === 1802
```

Two things compound here:

1. **The release is at 1px, not at a "small trackpad step".** On the base, `atBottomThreshold={120}` treats a 2px nudge as still-at-bottom and the follow continues. So for a reader who bumps the trackpad while pinned, this PR is a regression, not a fix.
2. **Nothing re-evaluates `pinned` afterwards.** Content growth doesn't move `scrollTop`, so no `scroll` event fires, so `pinned` stays `false` while the reply drifts arbitrarily far below the fold. The only way back is the manual scroll this PR set out to eliminate.

Separately, `<= 1` has no stated basis — it arrived in 84432cd8a to satisfy the clamp case in the harness, with a one-line commit message. And it contradicts the exported constant six lines up: `STICK_EDGE_THRESHOLD`'s doc says "maximum distance from the bottom that still counts as following the live end", but in the upward direction that constant is inert and the real value is 1.

Worth noting that `useStickToBottom` itself has no direct unit test — `stick-to-bottom.test.ts` only covers the three pure helpers, so this branch is only exercised indirectly through the DOM harness.

What I'd want before merging: either a stated basis for the asymmetry, or a single agreed threshold used in both directions.

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,74 @@
+"use client";
+
+import { useCallback, useEffect, useRef } from "react";
+
+/** Maximum distance from the bottom that still counts as following the live end. */
+export const STICK_EDGE_THRESHOLD = 120;
```
Bohan-J: **Blocking (needs an answer in the PR, not necessarily a rewrite).**

`packages/views/common/task-transcript/transcript-follow.ts` is an existing, shipped, unit-tested live-follow latch. Its header comment addresses the exact problem this hook is solving:

> "Am I at the live end right now" therefore cannot distinguish a reader who scrolled away from one the anchoring pushed down.

Its answer is to judge intent from **input events** — accumulated wheel / touchmove / arrow-key deltas plus `pointerdown`, with a 300ms intent window and an explicit guard against fighting text-selection autoscroll — and to treat system displacement as never counting.

This hook uses scroll direction instead, which is the signal that comment argues is insufficient. The commit sequence looks like that playing out: fb20625cd introduced "any upward move releases", and 84432cd8a had to carve out the clamp case eight minutes later. The next special cases are probably already queued up (selection autoscroll, iOS rubber-band, browser scroll anchoring).

Also, `transcript-follow.ts:22` documents `FOLLOW_EDGE_THRESHOLD` as "matching the chat list's atBottomThreshold". Replacing that literal with `STICK_EDGE_THRESHOLD` here turns that into a dangling cross-reference, and leaves two same-valued, same-meaning constants in two packages. Worth either importing one from the other or fixing the comment.

I'm not asking you to unify the modules — the geometry is mirrored (newest-first prepends vs. oldest-first appends). I am asking the PR to say why the existing latch can't be extended.

packages/views/chat/components/chat-message-list.autoscroll.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { I18nProvider } from "@multica/core/i18n/react";
+import { chatKeys } from "@multica/core/chat/queries";
+import type { TaskMessagePayload } from "@multica/core/types";
+import type { ReactElement } from "react";
+import enChat from "../../locales/en/chat.json";
+import { ChatMessageList } from "./chat-message-list";
+
+// Virtuoso cannot render rows in jsdom's zero-height viewport.
+vi.mock("react-virtuoso", () => ({
```
Bohan-J, line 25: **Blocking (coverage).** This file mocks `react-virtuoso` wholesale, and the mock never calls `followOutput` — so all 320 lines here exercise the `ResizeObserver` path only.

The existing `chat-message-list.test.tsx` does call `followOutput(true)`, but only at render time, when `isPinned()` is still its initial `true`.

Net result: `atBottom && isPinned()` — the change this PR leads with — has no test where `isPinned()` returns `false`. Please add a direct case asserting that a released pin turns the follow off even while Virtuoso still reports `atBottom`.

One smaller note on the harness: `scroller()` hardcodes the ordering "scroll event delivered, *then* resize observed" (`readerScrollsUp` dispatches the scroll event before the caller grows the content). Per the HTML spec's update-the-rendering steps that ordering is correct — scroll steps precede the ResizeObserver broadcast — so I don't think there's a live bug here. But it does mean a UA that ordered these differently would show up as "the reader's scroll-up gets reverted" with the suite still green. The input-event approach in `transcript-follow.ts` has no such dependency.

packages/views/chat/components/chat-message-list.tsx
```diff
@@ -327,13 +332,10 @@ export function ChatMessageList({
         // than the viewport, so switching sessions always shows the latest reply.
         initialTopMostItemIndex={{ index: "LAST", align: "end" }}
         increaseViewportBy={{ top: 400, bottom: 600 }}
-        atBottomThreshold={120}
-        // Follow rapid streamed output only while Virtuoso says the reader is
-        // at the live end. An in-flight smooth animation temporarily reports
-        // "not at bottom" on the next append and permanently drops the follow
-        // (#6697), so live growth must use an immediate scroll.
+        atBottomThreshold={STICK_EDGE_THRESHOLD}
+        // Follow only when Virtuoso and resize-driven pinning both report the live end.
```
Bohan-J: **Nit, but it loses information.** e7f53c70d dropped the comment that explained *why* this has to be `"auto"` rather than `"smooth"`:

> An in-flight smooth animation temporarily reports "not at bottom" on the next append and permanently drops the follow (#6697), so live growth must use an immediate scroll.

That took a round to establish in #7018, and the `"auto"` on the next line now looks arbitrary. Someone will flip it back. Please keep the original comment and add the new line after it rather than replacing it.

packages/views/chat/components/chat-message-list.tsx
```diff
@@ -312,6 +315,8 @@ export function ChatMessageList({
       // "near-viewport" against that element rather than the browser viewport —
       // otherwise a diagram only starts loading once it is already on screen.
       <RichContentScrollRootProvider scrollRoot={scrollContainerEl}>
+      {/* ResizeObserver cannot report scroll extent, so observe a content wrapper. */}
+      <div ref={setContentEl}>
```
Bohan-J: **Nit.** `react-virtuoso` already reports this: `totalListHeightChanged?: (height: number) => void` is on the props in the version this repo pins. Using it would give you the content-height signal without adding a DOM node or a second `ResizeObserver` target — the observer on `scrollEl` still covers the composer/viewport half.

Minor: the new wrapper wasn't followed by a re-indent, so the whole `<Virtuoso>` block still sits at its previous level.

Bohan-J (changes requested):
Thanks for this — the problem is real and the evidence you brought (sampled drift numbers and same-tick before/after captures, rather than "feels better") is the right standard. I confirmed the mechanism from the base: nothing in the chat list reacts to its own viewport shrinking, so the composer-growth case genuinely cannot self-correct today.

Splitting the scroll geometry into pure functions is a good call — it makes the threshold and the pin target independently verifiable. The downward-only constraint in `bottomPinTarget` is well reasoned and avoids yanking the reader upward when content shrinks.

I'm requesting changes on three points. The first is a behaviour regression relative to the base; the other two are a design question and a coverage gap.

### 1. A 1px upward move permanently drops the follow, with no recovery path (blocking)

`stick-to-bottom.ts` releases the pin on *any* upward movement that leaves the viewport more than 1px from the bottom. That narrows the effective release window in the upward direction from the base's 120px to 1px, and once released there is nothing that can re-engage it without a manual scroll. Repro and details in the inline comment.

### 2. There is already a live-follow latch in the repo (blocking — needs an answer, not necessarily a rewrite)

`packages/views/common/task-transcript/transcript-follow.ts` solves the same "reader intent vs. system displacement" problem, and its doc comment argues explicitly that scroll position/direction cannot distinguish the two — which is why it drives off input events instead. This PR adds a second latch using the weaker signal and doesn't mention the first one. I don't think the two must be merged (one is newest-first prepends, the other oldest-first appends), but the PR should say why extending the existing latch isn't the move. Inline comment has the details, including a now-dangling cross-reference this creates.

### 3. The PR's headline claim isn't covered by a test (blocking)

`followOutput={... atBottom && isPinned() ...}` is the central change, and no test exercises it with `isPinned()` returning `false`. Inline comment explains why the existing suites miss it.

Plus two nits inline: a lost `#6697` rationale comment, and `totalListHeightChanged` as a way to drop the wrapper `<div>`.

---

Verified locally at `84432cd8a`: the three chat suites pass (39 tests), `@multica/views` typechecks, and ESLint is clean on the four changed files. I did not check CI.

tim-smart:
Thanks for the review — all three blockers and both nits are addressed in 1e6e76af8. You were right on the central point: scroll direction was the wrong signal, and the second special case (`<= 1` for clamps) was the tell.

**1 + 2 (the 1px release, and the existing latch).** These turned out to be the same fix. `transcript-follow.ts`'s state machine is direction-agnostic once you feed it away-positive input deltas and the caller's own distance from the live end (`scrollTop` for the newest-first transcript, distance-from-bottom here), so the chat list now reuses it instead of running a parallel latch on the weaker signal. It is generalized to `createLiveEndFollow` / `onAtEdgeChange` (mechanical rename; the transcript dialog and its suite are updated), and `useStickToBottom` wires wheel/touch/key/pointer input mirroring the dialog's wiring, with signs flipped for a bottom live end. Consequences:

- A 2px trackpad nudge accumulates 2px of intent and keeps following — the base's 120px forgiveness is back, with `FOLLOW_EDGE_THRESHOLD` as the single threshold in both directions (the dangling cross-reference is gone; the chat list now passes the same constant to `atBottomThreshold`).
- Recovery exists: a sub-threshold nudge is pinned back after the 300ms intent window, exactly like the transcript treats sub-threshold residue.
- Incremental scroll-away still works: deltas accumulate across the gesture, so 60px ticks release at the threshold while pins are suppressed mid-gesture.
- The clamp carve-out is deleted, not patched: clamps produce no input, so they can never read as intent, whatever their ordering relative to the ResizeObserver. Same for selection autoscroll and rubber-banding — the latch's `pointerDown`/intent-window guards were built for those.

**3 (`followOutput` coverage).** The suite's Virtuoso stub now surfaces `followOutput` verdicts the same way `chat-message-list.test.tsx` does, and a new case asserts a released follow returns `false` while Virtuoso still reports `atBottom`. Mutation-checked: dropping the `isFollowing()` conjunct fails exactly that test; restoring the old any-upward-move release fails exactly the new trackpad-nudge test.

**Nits.** The #6697 auto-vs-smooth rationale is restored verbatim with the new conjunct explained after it, and the wrapper `<div>` is gone — content growth reaches the stick through `totalListHeightChanged`, so only the container is observed and the indentation issue disappears with the node.

Your harness-ordering observation also resolves itself: reader gestures in the reworked suite are wheel input followed by the scroll it causes, and clamps are position-only moves, so the tests no longer encode a scroll-vs-resize delivery order.

Verified: `packages/views` 405 files / 4823 tests green, `tsc --noEmit` and ESLint clean on the changed files.

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  useEffect(() => {
+    if (!scrollEl) return;
+
+    // Mirror of the transcript dialog's wiring with the live end at the
+    // BOTTOM: away from it is up, so every input sign flips.
+    const onWheel = (e: WheelEvent) => {
+      const scale =
+        e.deltaMode === 1 ? LINE_SCROLL_PX : e.deltaMode === 2 ? scrollEl.clientHeight : 1;
+      follow.input(-e.deltaY * scale);
```
Bohan-J: **Blocking.** This credits the latch with every `wheel` that reaches the container, including ones the chat list never acted on.

Chat rows contain four capped, scrollable `<pre>` blocks (`chat-message-list.tsx:1005`, `:1222`, `:1253`, `:1275`, all `max-h-*` + `overflow-auto`). Wheeling inside one scrolls the `<pre>` — the browser does not chain to the parent while the inner box still has room — but the event bubbles here anyway.

Driving the real hook (only `ResizeObserver` stubbed):

```ts
const { result } = renderHook(() => useStickToBottom(el));
// wheel from a <pre> inside a row; the list's scrollTop never changes
inner.dispatchEvent(new WheelEvent("wheel", { deltaY: -200, bubbles: true }));
// -> isFollowing() === false, el.scrollTop unchanged

for (let i = 0; i < 8; i++) { s.content += 180; result.current.onContentHeightChanged(); }
// -> distanceFromBottom === 1440
```

There's no way back: the list never scrolled, so no `scroll` event fires, so `atEdge` (line 114) still reads `true` and the false→true transition that would call `onAtEdgeChange` never happens. The reader has to scroll the outer list manually — the thing this PR set out to remove.

The variant I'd worry about more is the one that needs no scrollable content at all. On a conversation short enough to fit the viewport, nothing scrolls anywhere, but `wheel` still fires — so a single trackpad flick releases the follow *before the reply starts*, and the whole reply streams off-screen (1240px after 8 chunks in the same probe). That's an ordinary first-message path, not an edge case.

`touchmove` on line 95 has the same shape: a finger drag inside a `<pre>` reports away-displacement the list never took.

Filtering on `e.target` won't work here — unlike `keydown`, `wheel` targets the deepest element under the cursor even when the list is what scrolls. What I'd suggest instead is holding the input as *pending* intent and letting the list's own scroll event confirm it, so a release needs both user input and observed movement of the list. That preserves "system displacement never counts" and adds "input the list didn't consume doesn't count either".

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,159 @@
...
+      const y = e.touches[0]?.clientY;
+      if (y === undefined) return;
+      // Finger moving down scrolls the content up (away from the live end).
+      if (lastTouchY !== null) follow.input(y - lastTouchY);
+      lastTouchY = y;
+    };
+    const onKeyDown = (e: KeyboardEvent) => {
+      // Only keys aimed at the scroller itself; Space/arrows bubbling from
+      // row controls are not scroll intent.
+      if (e.target !== scrollEl) return;
```
Bohan-J: **Blocking (same root cause, opposite direction).** This guard is right in principle, but in the chat list it rejects every real keyboard scroll.

The scroll container (`chat-message-list.tsx:298-304`) has no `tabIndex`, so it can never be `e.target`. Meanwhile focus does land inside it — quick-action buttons, links, copy controls in rows are all focusable. When a keyboard user has focus on one and presses `PageUp`, the browser scrolls the nearest scrollable ancestor, which is the list.

So the list moves, the latch records nothing, and `enforce()` then sees `following === true` with no input in the intent window and pins them straight back:

```ts
// focus on a quick-action button inside a row
button.dispatchEvent(new KeyboardEvent("keydown", { key: "PageUp", bubbles: true }));
s.top -= 600;                                  // the browser scrolls the list
el.dispatchEvent(new Event("scroll"));
// -> scrollTop back to 3400 (reader wanted 2800), isFollowing() === true
```

The keypress is silently undone. This works on `main` today, so it's a regression.

The transcript dialog has the same guard, but its scroller is wired through `scrollerRef` and its rows have no inner scrollers, so I don't think it has this problem — worth confirming rather than assuming, though.

Confirming the release from actual list displacement (as suggested on the `wheel` handler) fixes this case too, without needing to decide which keys count.

packages/views/common/task-transcript/transcript-follow.ts
```diff
@@ -92,20 +102,20 @@ export function createNewestFirstFollow(now: () => number = () => Date.now()): N
...
     },
-    onAtTopChange(atTop: boolean) {
-      if (!active || !atTop) return;
+    onAtEdgeChange(atEdge: boolean) {
+      if (!active || !atEdge) return;
       following = true;
       pendingAway = 0;
     },
-    onScroll(scrollTop: number): boolean {
+    onScroll(distance: number): boolean {
```
Bohan-J, line 185: Context for the two blocking comments on `stick-to-bottom.ts` — the ambiguity is arguably here rather than at the call sites.

Generalising this into a direction-agnostic latch is a good call, and the doc rewrite lands the reasoning well. But the contract now has two inputs describing the same thing at different levels of trust: `input(delta)` means "the reader pushed by this much", while `onScroll(distance)` means "the surface is actually this far from the live end". Nothing ties them together, so a caller can accrue unlimited away-displacement from input that its scroller never took.

That was safe for the transcript, whose rows have no nested scrollers, so it never showed. It isn't safe for chat.

If you decide to fix it here rather than in the chat wiring, the shape I'd expect is for `input()` to stage intent and for `onScroll()` to be what promotes staged intent into a release — which would also let the state machine's existing unit tests express "input arrived but the surface never moved", a case the decision table can't currently represent.

Bohan-J (changes requested):
Re-reviewed at `3f9cbb5af`. All three blockers from the last round are genuinely fixed, and the shape of the fix is better than what I asked for:

- The asymmetric `<= 1` release is gone — one `FOLLOW_EDGE_THRESHOLD` now governs both the latch and Virtuoso's `atBottomThreshold`, so the two judges of "at the bottom" can't disagree. The new "keeps following after a small trackpad nudge near the live end" case covers exactly the repro I posted.
- Rather than justifying a second latch, you generalised the existing one. Making the state machine direction-agnostic — callers feed away-positive deltas and their own distance-from-live-end — is a cleaner boundary than either surface had before, and updating the transcript call site plus the stale cross-reference comment in the same pass is the right thoroughness.
- `totalListHeightChanged` replaced the wrapper `<div>`, the `#6697` rationale is back and extended, and `followOutput` released-state coverage is there.

One new blocker, then I think this is good to go.

### The latch is fed input the list never consumed — and starved of input it did

`input()` is credited from raw `wheel` / `touchmove` / `keydown` on the container, with no check that the chat list actually moved. That breaks in both directions:

**1. Scrolling inside a message's own scroll box.** Chat rows render four `max-h-32/40 overflow-auto` `<pre>` blocks (`chat-message-list.tsx:1005`, `:1222`, `:1253`, `:1275`). Wheeling inside one scrolls the `<pre>`; the event still bubbles to the list, so the latch accrues 200px of "away" and releases — but the list never moved, so no scroll event fires and `atEdge` stays stale at `true`. Since re-engagement only happens on a false→true `onAtEdgeChange` transition, nothing re-engages. I measured the tail at 1440px below the fold after 8 chunks.

**2. A conversation too short to scroll.** This one is worse because it needs no scrollable content at all. On a fresh chat that fits the viewport there is nothing to scroll anywhere, but `wheel` still fires — one trackpad flick releases the follow *before the reply starts*. The reply then streams entirely off-screen (1240px in my probe).

**3. The mirror image: `PageUp` from a focused control inside a row.** The guard on line 101 correctly rejects keys not aimed at the scroller, but the scroll container has no `tabIndex`, so it is never `e.target`. When focus sits on a quick-action button or link in a row, the browser still scrolls the nearest scrollable ancestor — the list. The latch never learns, `enforce()` sees `following === true` with no input recorded, and pins the reader straight back to the bottom. Their keypress is silently undone.

All three are the same root cause: intent is inferred from *events observed*, rather than from *displacement the list actually took*. A target filter won't fix it — `wheel` legitimately targets descendants even when the list is what scrolls.

The direction I'd suggest: treat input as *pending* intent and let the list's own scroll event confirm it, so a release requires both user input and observed movement of the list. That keeps the property the latch exists for (system displacement never counts) while adding the one it's missing (input the list didn't consume doesn't count either), and it fixes case 3 for free. Where that lands — the shared state machine or the chat wiring — is your call; the ambiguity is really in the latch's contract, since `input(delta)` currently means "the reader pushed" rather than "the surface moved".

Worth noting the transcript dialog is **not** affected today: its rows have no inner vertical scrollers, so the same wiring is latent there. No need to change that surface beyond whatever falls out of the shared fix.

Regression tests I'd want: wheel input while the list's `scrollTop` never changes keeps following; the same on a conversation with no scrollable distance; and keyboard paging that does move the list releases rather than being reverted.

---

Verified locally at `3f9cbb5af`: chat + transcript suites pass (91 tests across 5 files, including the transcript dialog since this touches shared code), `@multica/views` typechecks, ESLint clean on all 7 changed files. The three cases above were reproduced against the real hook with temporary probes, since none of them are reachable through the current harness; the probes are deleted and the tree is clean. I did not check CI.

tim-smart:
Fixed in 247523198, taking the contract change you sketched — the fix landed in the shared state machine, since as you said the ambiguity was in `input(delta)`'s meaning, not the call sites.

**The latch now stages intent and lets the surface's scroll promote it.** `input()` only accrues; `onScroll()` (a real scroll of the surface) is what promotes staged intent past the threshold into a release, and the split entry point `onResize()` carries the displacement-without-scroll signals (content growth, viewport resize) and can never promote. The release amount is always the staged input, never the observed distance, so the transcript's original prepend-inside-the-intent-window hazard stays covered. Staged intent also expires with its gesture: input the surface never consumed does not linger into a later gesture's accounting.

That resolves all three cases:

- **Nested `<pre>` scroller / bubbled wheel**: staged, never confirmed, never releases — and the stale-`atEdge` trap can't be reached because the follow never disengages without the list moving.
- **Conversation too short to scroll**: same path; a flick before the first reply stages intent that no scroll ever promotes, and the reply streams pinned.
- **Keyboard paging from a focused row control**: the chat wiring drops the `e.target` guard entirely, which staging makes safe in both directions — a scroll key from anywhere in the container stages intent, and the browser decides: if it scrolls the list, the scroll confirms the release; if a control consumed the key (Space activating a button, Home in a field), nothing promotes. `Home`/`End` now stage `±scrollHeight` instead of disengaging unconditionally, for the same reason. No key-classification logic needed, as you predicted.

**Tests.** The decision table gained the case it couldn't previously express — "input arrived but the surface never moved" — plus gesture-expiry, and the accumulated-release case now asserts the scroll-confirmation step. The DOM suite adds your three regressions: bubbled wheel with an unmoving list keeps following, a flick on a too-short conversation keeps following, and PageUp from a focused row control releases rather than being reverted. All are mutation-checked: restoring release-on-input fails the first two plus both new latch cases, and restoring the keydown target guard fails exactly the keyboard test.

**On the transcript dialog**: confirmed rather than assumed — its Virtuoso scroller renders with `tabIndex: 0`, so unlike the chat container it is focusable and its keydown guard can genuinely match. A focused row control there would still bypass the guard, but that is shipped `main` behavior, unchanged by this PR; its wiring still calls `onScroll` only from real scroll events, so the promotion semantics slot in without any dialog change.

Verified: `packages/views` 405 files / 4828 tests green, `tsc --noEmit` and ESLint clean.

packages/views/common/task-transcript/transcript-follow.ts
```diff
@@ -79,9 +112,11 @@ export function createNewestFirstFollow(now: () => number = () => Date.now()): N
     },
     input(delta: number) {
       if (!active) return;
+      // A fresh gesture starts a fresh accumulation: intent the surface
+      // never consumed must not linger into a later gesture.
+      if (!inputFresh()) pendingAway = 0;
```
Bohan-J: **Blocking — this is the one I'd fix first.**

Resetting on a stale previous input means a release requires crossing `FOLLOW_EDGE_THRESHOLD` **inside a single 300ms burst**. A trackpad flick clears that easily. Someone reading does not: a discrete mouse wheel sends one notch per event, and notches are far more than 300ms apart when you're actually reading.

So `pendingAway` gets reset to 0 on every notch and never accumulates, while the list really does scroll. Driving the real hook at reading pace (500ms between notches, list scrolled each time, then one streamed chunk):

```
Chrome/Win (deltaMode 0, 100px/notch): scrolled up 500px, following=true, after one chunk distance=0
Firefox    (deltaMode 1, 3 lines/notch): scrolled up 600px, following=true, after one chunk distance=0
```

The reader ends up 500-600px back through history, the latch still believes they're following, and the next chunk yanks them to the bottom. On `main` this works — Virtuoso's `atBottom` drops the follow after 120px of real displacement, regardless of how long it took.

Note also that a Firefox notch is exactly 3 × `LINE_SCROLL_PX` = 120px and line 147 tests `> FOLLOW_EDGE_THRESHOLD`, so a single notch can never release even in principle.

The underlying tension: `pendingAway` measures *input pushed*, and the threshold it's compared against (120px) is a *displacement* budget shared with `atBottomThreshold`. Those only line up when the input arrives fast enough to be treated as one gesture.

Since line 147 already requires a real scroll before releasing, the scroll could supply the magnitude instead — accumulate the displacement the surface actually took between scroll events while input is staged, and compare that to the threshold. It releases at 120px of real movement whatever the pacing or device, still ignores system displacement (nothing staged, nothing accumulates), and would make the window-expiry problem on `onResize` moot too.

packages/views/common/task-transcript/transcript-follow.ts
```diff
@@ -92,26 +127,32 @@ export function createNewestFirstFollow(now: () => number = () => Date.now()): N
...
+      // observed distance, which a system shift inside the intent window
+      // could have inflated on its own.
+      if (following && inputFresh() && pendingAway > FOLLOW_EDGE_THRESHOLD && distance > 0) {
+        following = false;
+        return false;
       }
-      return false;
+      return pinVerdict(distance);
+    },
+    onResize(distance: number): boolean {
```
Bohan-J, line 272: **Blocking.** `onResize` declining to pin while input is fresh is correct — but nothing re-evaluates when the window expires, and `onResize`/`onScroll` only run on the next event. If the last event of a stream lands inside the window, the pin is declined and never reconsidered.

```
2px nudge up (list moves, scroll fires)
final chunk +180px at t+100ms -> onResize declines: input still fresh
t+500ms: distance = 182px, following = true, nothing further will fire
```

182px is about the last row plus the status pill sitting under the composer, and it stays there for the rest of the session. The trigger is ordinary — replies end, so there is always a final chunk, and it has a 300ms window to land in after any small nudge.

Worth noting the current tests can't catch this: they either settle the gesture before growing, or advance the clock and then call `onResize` a second time by hand — which is exactly the event that doesn't exist in the real sequence.

A cancellable re-evaluation scheduled at window expiry (reset on new input, cleared on unmount) would close it. If you take the displacement-accumulation approach from the other comment, the pin decision stops depending on `inputFresh()` at all and this case disappears on its own — which is part of why I'd lean that way.

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,179 @@
...
+    // arrive from a focused row control, and the browser answers them by
+    // scrolling the nearest scrollable ancestor — this list. Staging makes
+    // that safe in the other direction too: a key a control consumed (Space
+    // activating a button, Home in a text field) never scrolls the list, so
+    // the staged intent is never promoted.
+    const onKeyDown = (e: KeyboardEvent) => {
+      if (e.key === "ArrowUp") follow.input(LINE_SCROLL_PX);
+      else if (e.key === "ArrowDown") follow.input(-LINE_SCROLL_PX);
+      else if (e.key === "PageUp") follow.input(scrollEl.clientHeight);
+      else if (e.key === "PageDown" || e.key === " ") follow.input(-scrollEl.clientHeight);
```
Bohan-J: **Blocking (small, mechanical).** `Shift+Space` pages **up**, but every `" "` is recorded here as moving toward the live end.

```
Shift+Space from a focused row control -> list scrolls up 600px (top 3400 -> 2800), following=true
next content change                    -> pinned back to the bottom (top 3580, distance 0)
```

Because the staged delta is negative, the real upward movement never promotes, and the reader's page-up is reverted on the next content change. Standard shortcut, so this is a regression against `main` for keyboard users.

Needs `e.shiftKey` to flip the sign — away-positive here, toward-negative in the transcript.

`agent-transcript-dialog.tsx:366` has the same line with the mirrored sign, so the transcript has carried this bug since before this PR. Not something you introduced, but it's the same shared wiring and cheap to fix in both passes.

Bohan-J (changes requested):
Re-reviewed at `247523198`. The staging contract is the right fix for the last round — input is held until the surface's own scroll confirms it, `onScroll` vs `onResize` cleanly separates "the reader moved it" from "the system moved it", and all three cases I raised are genuinely closed: wheeling inside a `<pre>`, flicking a chat too short to scroll, and `PageUp` from a focused row control all behave now. Dropping the `e.target` guard in favour of staging (a key a control swallows never scrolls the list, so its staged intent never promotes) is a nicer resolution than the filter I was imagining.

Three blockers left. They're all in the promotion rule rather than the staging idea, and the first one is the one I'd fix first.

### 1. Release is impossible at reading pace, so scrolling back through history gets undone

`input()` resets `pendingAway` whenever the previous input is older than the 300ms window, so a release requires crossing 120px **within a single 300ms burst**. That's fine for a trackpad flick. It is structurally unreachable for someone reading: a discrete mouse wheel delivers one notch per event, and notches land well over 300ms apart when you're actually reading what scrolls past.

The reader scrolls up, the list really moves, and the latch still says `following === true` — so the next streamed chunk pins them straight back to the bottom:

```
Chrome/Win (deltaMode 0, 100px/notch): scrolled up 500px, following=true, after one chunk distance=0
Firefox    (deltaMode 1, 3 lines/notch): scrolled up 600px, following=true, after one chunk distance=0
```

Five notches, at reading pace, all undone by one chunk. This is the ordinary "scroll back to re-read something while a reply streams" path, and it's a regression against `main`, where Virtuoso's own `atBottom` would have dropped the follow at 120px of actual displacement.

There's a sharp edge on the boundary too: a Firefox wheel notch is exactly 3 lines × `LINE_SCROLL_PX` = 120px, and the check is `> FOLLOW_EDGE_THRESHOLD`, so even two notches inside the window sum to 240 only if both land in time — a single notch can never release no matter what.

### 2. Nothing re-evaluates when the intent window expires

`onResize` correctly refuses to pin while input is fresh, but nothing fires at the 300ms boundary, and `onScroll`/`onResize` only run on the next event. If the last chunk of a reply arrives inside the window after a small nudge, the pin is declined and never reconsidered:

```
2px nudge up -> final chunk +180px at t+100ms -> onResize declines (input still fresh)
t+500ms: distance = 182px, following = true, and nothing further will fire
```

182px is roughly the last row plus the status pill, parked under the composer for the rest of the session. Every existing test avoids this ordering by settling the gesture before growing, or by calling `onResize` a second time after advancing the clock.

### 3. `Shift+Space` is recorded as moving toward the live end

`stick-to-bottom.ts:123` treats every `" "` like `PageDown`, but `Shift+Space` pages **up**. The reader pages up, the staged delta is negative so nothing releases, and the next content change drags them back:

```
Shift+Space -> list scrolls up 600px (top 3400 -> 2800), following=true
next content change    -> pinned back to the bottom (top 3580, distance 0)
```

Same line exists in `agent-transcript-dialog.tsx:366` (mirrored sign), so this one predates the PR on the transcript side — worth fixing in both while the wiring is fresh.

---

For 1 and 2 together, the thing that would settle both is deciding what "confirmed" means. Right now it's "input crossed the threshold inside one gesture, and a scroll happened". The alternative is to let the scroll itself supply the magnitude: stage the fact that input arrived, then measure the displacement the surface actually took between scroll events and accumulate *that* against the threshold, decaying it only when the reader returns to the edge. That releases at 120px of real movement regardless of pacing or input device, keeps system displacement out (no input staged → no accumulation), and removes the need for anything to fire on window expiry, because the pin decision is no longer gated on a timer. Your call whether that's worth the reshuffle versus patching the two cases directly — I don't want to force a redesign this late if you see a smaller fix that covers both.

Tests I'd want alongside: wheel notches spaced past the intent window still release; a growth that lands inside the window still ends up pinned once it expires; and `Shift+Space` from a focused row control releases rather than being reverted.

---

Verified locally at `247523198`: 96 tests pass across the 5 chat + transcript files, `@multica/views` typechecks, ESLint clean on all 7 changed files. All three cases above were reproduced against the real hook with temporary probes (fake timers for the window, real `WheelEvent` `deltaMode` values for the notch units); probes deleted, tree clean. I did not check CI.

tim-smart:
Took the reshuffle — fixed in 3e73087c3. You were right that the two promotion problems shared a root: `pendingAway` measured input pushed while the threshold it met is a displacement budget, and the timer was holding the whole thing together. The scroll now supplies the magnitude.

**New contract.** `input()` only opens a per-gesture budget (split by direction). Each scroll attributes the displacement the surface actually took to the reader, capped by that budget — so a system shift landing inside a gesture (a transcript prepend) attributes at most the gesture's own size and the remainder is pinned immediately. Attributed displacement (`awayTaken`) is what releases past `FOLLOW_EDGE_THRESHOLD`, and it does not expire:

- **Reading pace (blocker 1)**: notches spaced any distance apart accumulate — two 100px Chrome notches or two 120px Firefox notches release, whatever the pacing. The single-Firefox-notch boundary is gone with the burst window; only budgets expire with the gesture, never taken displacement.
- **Window expiry (blocker 2)**: the pin verdict no longer consults the intent window at all — only `mouseHeld`/`scrollbarDrag` suppress it (selection autoscroll intact). Your 2px-nudge-then-final-chunk case pins on the chunk itself; nothing needs to fire at the boundary, and the timer-shaped hole in the test harness you pointed out closes with it (the new tests grow inside the window and assert the pin).
- **Shift+Space (blocker 3)**: `e.shiftKey` flips the sign in both wirings — away-positive in the chat list, toward-negative in the transcript dialog, where the unsigned line predated this PR as you noted.

One deliberate choice to flag, since it goes slightly past your sketch: `awayTaken` decays on the reader's own attributed return and resets at the live-end re-engagement, but **pins do not erase it**. Crediting the system's snap-back against the reader's displacement would re-trap anyone scrolling sub-threshold notches against a fast stream — every chunk would zero their progress, which is the shape of complaint that opened this PR. As written, persistent sub-threshold attempts accumulate into a release (covered by `accumulates displacement across pins during a fast stream`). The mirror cost is that a reader who nudges, gets pinned, and nudges again minutes later releases on cumulative intent — I'd argue repeated upward nudges against a pin *are* a release signal, but if you'd rather pins clear the slate for the quiet-chat case, that's a one-line change and I'll take your call.

Tests: the three you asked for are in, at both layers where they belong — reading-pace notches release (latch + DOM), growth inside the window pins (latch + DOM), Shift+Space from a focused row control releases (DOM). Each blocker fix is mutation-checked: re-expiring `awayTaken` per gesture fails the pacing tests, re-adding the freshness gate to the pin fails the window tests, and reverting the sign fails the Shift+Space test — each mutation fails only its own tests.

Verified: `packages/views` 405 files / 4835 tests green, `tsc --noEmit` and ESLint clean on the changed files.

packages/views/common/task-transcript/transcript-follow.ts
```diff
@@ -92,25 +136,62 @@ export function createNewestFirstFollow(now: () => number = () => Date.now()): N
...
+        const attributed = Math.min(awayBudget, moved);
+        awayBudget -= attributed;
+        awayTaken += attributed;
+        if (following && awayTaken > FOLLOW_EDGE_THRESHOLD) {
+          following = false;
+          return false;
+        }
+        // Wholly the reader's own move: leave them be. Any unattributed
+        // remainder is the system's and falls through to the pin.
+        if (attributed === moved) return false;
```
Bohan-J: **Blocking.** Requiring `attributed === moved` exactly means any movement the budget doesn't fully cover falls through to `pinVerdict` and snaps the reader back to the live end.

Two ways that happens on touch:

**Momentum.** Only `touchmove` funds `awayBudget`, and `touchend` isn't wired anywhere, so once the finger lifts the momentum scroll is entirely unbudgeted:

```
finger travels 80px, surface tracks it 1:1 -> distance 80, following=true   (correct)
momentum carries 80px further              -> distance 0                    (snapped back)
```

A longer flick fares no better — 100px of finger travel plus 720px of momentum also ends at distance 0 with `following` still true. Since finger travel is a small fraction of the scroll a flick is meant to produce, a reader on mobile effectively can't leave the live end while a reply streams.

**Sub-pixel mismatch, finger still down.** Exact equality is fragile even mid-drag:

```
finger travel 80.0        -> awayBudget 80.0
surface moves 80.5        -> distance 0, pinned while the finger is still on the glass
```

`clientY` and `scrollTop` are both fractional on hi-DPI, and rubber-band regions deliberately aren't 1:1, so "the numbers match to the pixel" isn't a safe proxy for "this was all the reader".

Related: the header still promises pinning never happens "while the user is mid-gesture or holding the mouse down", but `pinVerdict` (line 98) now only checks `mouseHeld` / `scrollbarDrag`, and touch has no held-state equivalent — so on touch that guarantee isn't enforced at all.

packages/views/common/task-transcript/transcript-follow.ts
```diff
@@ -92,25 +136,62 @@ export function createNewestFirstFollow(now: () => number = () => Date.now()): N
...
-        following = false;
+      if (scrollbarDrag) {
+        if (distance > FOLLOW_EDGE_THRESHOLD) following = false;
         return false;
       }
-      if (following && !userControlsViewport() && scrollTop > 0) {
-        // System displacement got corrected; drop any sub-threshold residue
-        // so old nudges don't accumulate into a spurious disengage later.
-        pendingAway = 0;
+      if (moved > 0 && inputFresh() && awayBudget > 0) {
```
Bohan-J, line 236: **Blocking (the opposite failure of the one on line 170).** The comment below says the budget cap is what keeps a system shift from counting as reader intent. It bounds the amount, but it doesn't check that the surface moved *because of* the staged input — so leftover budget from input this surface never consumed sits ready to absorb whatever moves next.

```
wheel over a message's <pre>: 300px staged, the list itself does not move
100ms later a 200px system shift moves the list away from the live end
-> min(300, 200) = 200 credited to the reader, awayTaken > 120, follow released
```

That contradicts the invariant stated in the header — "System displacement never counts, no matter how far it moves the viewport" — and it's the case the staging contract was introduced to prevent in the first place.

It also constrains the fix for line 170: you can't resolve the momentum problem by crediting everything that moves inside the intent window, because this is what that would license.

The signal that separates them is whether the preceding input was actually consumed here. In the momentum case the finger moved the list 1:1, so the surface is demonstrably in reader-driven motion. In this case the input produced zero movement, so nothing that follows belongs to the reader. If `input()` recorded a claim and `onScroll` recorded whether that claim was ever honoured, both cases fall out of the same flag.

packages/views/chat/components/stick-to-bottom.ts
```diff
@@ -0,0 +1,181 @@
...
+    const onWheel = (e: WheelEvent) => {
+      const scale =
+        e.deltaMode === 1 ? LINE_SCROLL_PX : e.deltaMode === 2 ? scrollEl.clientHeight : 1;
+      follow.input(-e.deltaY * scale);
+    };
+    let lastTouchY: number | null = null;
+    const onTouchStart = (e: TouchEvent) => {
+      lastTouchY = e.touches[0]?.clientY ?? null;
+    };
+    const onTouchMove = (e: TouchEvent) => {
```
Bohan-J, line 123: `touchmove` is the only thing that funds the budget here, and there's no `touchend` listener on either surface — so there is currently no way to represent "the gesture ended and momentum may follow", which is what the blocking comment on `transcript-follow.ts:170` needs.

The same gap exists in `agent-transcript-dialog.tsx`; whatever shape the fix takes, both wirings need it.

Minor, while you're in here: `onTouchStart` reads `e.touches[0]` and `onTouchMove` tracks a single `lastTouchY`, so a second finger landing mid-scroll retargets the delta to whichever touch happens to be first in the list. Not worth its own round, but a stray jump in `lastTouchY` feeds straight into `input()` as reader intent.

Bohan-J (changes requested):
Re-reviewed at `3e73087c3`. Letting the scroll supply the magnitude is the right call and it closed all three from last round: discrete wheel notches at reading pace now accumulate and release, a final chunk landing inside the window gets pinned once the window expires, and `Shift+Space` is fixed on both surfaces. Splitting `awayBudget` / `towardBudget` / `awayTaken` so a reader's partial retreat decays their accumulated intent is a nicer model than what I sketched.

On the policy call you flagged for a maintainer — keeping `awayTaken` across system pins — I agree with keeping it. Repeated movement toward history is intent even when each individual push is small, the toward-direction decay and the edge reset both bound it, and the alternative silently punishes readers during fast streams. `FOLLOW_EDGE_THRESHOLD` continuing `atBottomThreshold` and 300ms only expiring unconsumed budget are both justified; no unexplained parameters this round.

Two blockers left. They're the two sides of one gap: **attribution is by amount, with no notion of whether the surface is currently in reader-driven motion.** Movement the budget doesn't cover falls through to `pinVerdict` and snaps the reader back; movement the budget *over*-covers is credited as intent and releases.

### 1. Under-attribution: touch is snapped back, both after and during a gesture

Only `touchmove` funds the budget, and nothing wires `touchend`, so momentum after the finger lifts is unbudgeted:

```
finger travels 80px, surface tracks 1:1 -> distance 80, following=true   (correct)
momentum carries 80px more              -> distance 0                    (snapped back)
```

Every flick is truncated to raw finger travel, which on touch is a small fraction of the intended scroll. A longer flick doesn't help — 100px of finger plus 720px of momentum also lands back at distance 0, still `following`. On mobile that means a reader effectively cannot leave the live end while a reply streams.

It isn't only momentum, though. The escape on line 170 requires `attributed === moved` exactly, so any mismatch between finger travel and the scroll the surface actually took falls through to the pin — **with the finger still on the glass**:

```
finger travel 80.0 -> awayBudget 80.0
surface moves 80.5 (fractional scrollTop, hi-DPI) -> distance 0, pinned mid-drag
```

Sub-pixel `clientY` and fractional `scrollTop` are normal on hi-DPI, and rubber-band regions are not 1:1 by design, so exact equality is a fragile thing to hang "leave the reader alone" on.

Worth noting the module header still promises pinning never happens "while the user is mid-gesture or holding the mouse down". `pinVerdict` now only checks `mouseHeld` / `scrollbarDrag`, and touch has no held-state equivalent, so that guarantee currently doesn't hold on touch at all.

### 2. Over-attribution: system displacement is credited when budget is left over

The header's other invariant — "System displacement never counts, no matter how far it moves the viewport" — doesn't hold either, in the opposite direction:

```
wheel over a message's <pre>: 300px staged, list does not move
100ms later, a 200px system shift moves the list away from the live end
-> the full 200px is attributed to the reader, awayTaken > 120, follow released
```

The comment on line 158 says the budget cap is what stops a system shift from counting — but the cap only bounds it, it doesn't require the surface to have moved *because of* the input. When staged input was never consumed, there's leftover budget sitting there ready to absorb whatever moves next.

---

These constrain each other, which is why I'd treat them as one fix: you can't simply credit all in-window movement (that's case 2), and you can't refuse everything the budget doesn't cover (that's case 1).

What distinguishes them is whether the preceding input was actually consumed by this surface. In case 1 the finger moved the list 1:1, so the surface is demonstrably in reader-driven motion and its continued travel in the same direction is that same gesture finishing. In case 2 the input produced *zero* movement here, so nothing about the following displacement is the reader's. Tracking "the last input was confirmed consumed" — and only then letting continued same-direction movement draw on a settle window, rather than a fixed px budget — separates them. `touchend` would need wiring either way; there's currently no signal that a gesture has ended and momentum may follow.

Regression tests I'd want: a flick whose momentum carries past the threshold releases rather than snapping back; a drag frame where the scroll doesn't exactly match finger travel is left alone; and a system shift following unconsumed input still pins.

---

Verified locally at `3e73087c3`: 103 tests pass across the 5 chat + transcript files, `@multica/views` typechecks, ESLint clean on all 7 changed files. All four cases above were reproduced against the real hook with temporary probes (fake timers, real `TouchEvent` / `WheelEvent` sequences); probes deleted, tree clean. I did not check CI.

tim-smart:
Addressed the `3e73087c3` review in `d934726d5`.

- Unconsumed input is discarded after the next animation frame, so later system movement cannot inherit a wheel event consumed by a nested scroller.
- Confirmed touch motion uses the list's measured displacement and continues through fractional drag frames and momentum. Both surfaces now wire `touchend` and `touchcancel` while tracking one touch identifier.
- Added regressions for momentum release, fractional drag, resuming pins after touch end, and system movement after unconsumed input.

Focused chat and transcript suites pass: 5 files, 110 tests. `@multica/views` typecheck and targeted ESLint also pass.

Bohan-J (approved):
Thanks for sticking with this — it went through more rounds than anyone wanted, and the shape it landed in is genuinely better than where it started.

Two things I want to call out, because they're what made the difference:

- You never patched the symptom. Each round you reached for the concept that was actually missing — staging, then displacement-as-magnitude, then confirmed motion — rather than widening a threshold until the failing probe went green.
- When the review pointed at `transcript-follow.ts`, you generalised it into a shared latch and updated the transcript call site, instead of arguing for a second parallel one. That's the outcome we wanted and you got there without being told to.

The before/after sampling in your first comment also set the bar for the whole thread. "5 samples in 8 partly hidden, worst case 119px" is a claim a reviewer can check; "feels smoother" isn't.

Heads up on three commits I pushed to your branch directly (maintainer edits were enabled, and it seemed better than a sixth round trip):

- `dc19407cb` — capped confirmed non-touch motion by its unspent claim. The motion window had no magnitude attached, so a transcript prepend landing inside it was credited to the reader: 30px of real movement released the follow permanently, which is the #5921 case the latch exists to prevent.
- `b6f7d0866` — fixed two ordinary wheel sequences my own change broke: an exact carry comparison rejected the last frame of an unevenly split animation by about one ulp, and a notch arriving mid-animation was discarded before it could be confirmed.
- `0e5cb8658` — a regression test I added was starting from a non-following state, so it passed with the implementation deleted. Now it doesn't.

Please do look those over; if you disagree with any of it, say so and we'll sort it out in a follow-up rather than leaving it.

One known residual, documented rather than fixed: while same-direction motion is still in flight, a wheel claim consumed by a nested scroller can be folded into the carry and spent by that motion. It's bounded by the claim's remaining amount and by the settle window, and it errs toward not interrupting the reader. Closing it properly means giving claims a source identity, which is our debt to pay, not yours.

Merging now.

PR #5792, opened 2026-07-22
Add generic OIDC authentication

drybalka-s:
## What does this PR do?

Adds a separate, discovery-based OpenID Connect authentication path for self-hosted Multica deployments using providers such as Authentik and GitLab. Existing email-code and Google flows remain unchanged.

The provider exchange stays on the server. The flow validates state, nonce, PKCE, issuer, audience, signature, subject, and verified email, then links returning users by stable `(issuer, subject)` identity instead of mutable email.

## Related Issue

Closes #5791

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation update
- [x] Tests
- [ ] CI/CD or infrastructure change

## Changes Made

- Added generic OIDC discovery, authorization, callback, and signed flow-state handling.
- Added stable OIDC identity persistence and concurrent-index migrations.
- Added verified-email and optional group-allowlist enforcement.
- Wired provider discovery into web and desktop login.
- Added iOS system-browser handoff through `multica://auth/callback`.
- Added Docker Compose, Helm, environment, and operator documentation.
- Added coverage for configuration, state, nonce, PKCE, scalar and array group claims, callback routing, and mobile handoff.

## How to Test

1. Configure `OIDC_ISSUER_URL`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`, and `OIDC_REDIRECT_URI`.
2. Sign in through the configured provider on web, desktop, or mobile.
3. Set `OIDC_ALLOWED_GROUPS` and verify allowed and rejected memberships.
4. Change the provider-side email and verify the existing account is reused.

Automated validation completed:

- `go test ./...`
- web: 112 tests, typecheck, ESLint
- mobile: 35 tests, typecheck, lint
- `helm lint deploy/helm/multica`
- `docker compose -f docker-compose.selfhost.yml config --quiet`
- `git diff --check`

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [x] I have updated relevant documentation to reflect my changes
- [x] If this change touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx`
- [x] I have considered and documented any risks above
- [x] I will address all reviewer comments before requesting merge

Risks and remaining manual checks:

- Live Authentik/GitLab end-to-end validation requires deployment credentials and was not run locally.
[...]

Issue #5529, opened 2026-07-16
[Feature]: Enterprise-Level Permission Controls


runzhezhang11:
### Deployment type

self-host

### What do you want and why?

Enterprise level, control panel is very import.
for example
1. User-Level Access Control
Roles such as Admin, Workspace Owner, Editor, and Viewer.
2. Agent Permission Control
Users can define which MCP tools and skills can be configured, and manage bash execution permissions.

iiwish:
Thanks for raising this. After making a few scoped contributions to Multica, including #6830, I began mapping the authorization paths that currently span human, agent, and workspace boundaries. I would like to contribute to this area long-term, but before proposing an implementation I wanted to ask whether the maintainers would find a short umbrella authorization RFC useful.

The first document would be intentionally narrow and do three things:

- map the current principals, attribution paths, and authorization gates;
- define shared authorization terminology and safety invariants around actors, actions, resources, context, explicit grants, denial, revocation, and audit; and
- record backward-compatibility constraints and open product decisions, preserving simple personal use while allowing team and enterprise governance to grow progressively.

It would not prescribe a complete RBAC/ABAC engine or start with schema changes and a broad permission rewrite. Product choices would remain explicit review questions rather than assumptions embedded in code.

I have read #6064 on member, project, and issue permissions, #4981 on runtime ownership, execution consent, and the agent/executor boundary, and #6738 on A2A invocation permissions. I would treat those efforts as inputs and compatibility constraints, not start competing implementations.

For runtimes, I would only cover the points where runtime identity, task admission, capability propagation, owner consent, and audit interfaces meet the shared authorization model. The initial RFC would not redesign runtime sandboxing or replace the product design being shaped in #4981.

If this direction aligns with the project, I can start with a one- or two-page outline and current-state decision matrix for maintainer review before writing any implementation. Would you prefer this kind of proposal as a GitHub Discussion, an issue, or a repository RFC document? If there are existing designs or established terms I should align with, please point me to the right starting place.

iiwish:
Thanks for opening this — it lines up with several related asks: #6446 (runtime permission isolation), #6064 (member-level visibility), #4020 (outbound network allowlist), and #4221 (a viewer-type role).

I've been mapping the current authorization architecture in #7916 (current-state docs only, no behavior changes; the §Open design boundaries section there lists the same enterprise questions). Based on that map, the gap toward enterprise controls splits into three layers:

[...]

PR #7616, opened 2026-08-27
feat(auth): add self-hosted Gitea OAuth login

2233admin:
## Summary

- Add secure self-hosted Gitea OAuth login while preserving the existing email-code flow.
- Map returning users to a stable Gitea provider/issuer/subject identity and verify the primary Gitea email on first login.
- Use signed state, HttpOnly OAuth state, PKCE, exact redirect validation, backend-issued authorization, config redaction, migrations, tests, docs, and localized UI copy.

## Validation

- Direct TypeScript checks pass for the changed frontend packages and generated docs sources.
- Focused login-page tests pass: 35/35.
- Focused Go tests pass: `go test ./internal/handler ./cmd/server ./pkg/db/generated`.
- Full repository checks remain environment-blocked on this Windows host: Docker Desktop is unavailable, Playwright Chromium is not installed, and the repository has unrelated baseline failures in core/web/desktop/Go wrapper checks.

## Base and scope

- Based on upstream `main` at `09a2410e882be8435bd6c4a26e03f7e288038203` (`v0.4.35`).
- This PR intentionally contains only the upstream-oriented Gitea login support. NAS deployment and Docker/3080 validation will be handled separately after review.

Draft while the complete upstream CI and Playwright validation are completed in a Linux/Docker environment.


iiwish:
@2233admin Nice — self-hosted Gitea is a natural pairing for this stack, and this is the first community sign-in provider PR I've seen here.

One piece of context that might help it land: sign-in providers are currently hardcoded in `server/internal/handler/auth.go` — Google's token exchange and userinfo calls are inline. Adding Gitea the same way would mean a second inline implementation, and the enterprise-identity asks in #5529 (SAML/OIDC/SCIM) won't scale that way.

Would you be open to extracting a small provider abstraction first — config-driven issuer/client-ID/secret plus the standard OAuth2/OIDC token + userinfo flow — with Google migrated onto it and Gitea as the second provider? A config-driven OIDC client would likely cover most of the self-host demand in one move (Keycloak, Authentik, Authelia, and Gitea's OAuth2 provider).

Happy to take that abstraction refactor as a separate small PR stacked ahead of yours, or to help within this PR — whichever makes review easier for the maintainers. Your call.

2233admin:
Thanks @iiwish — agreed, a second inline provider isn't the right shape. Please go ahead with the provider-abstraction PR stacked ahead of this one (config-driven issuer / client id / secret, standard OAuth2/OIDC token + userinfo flow, Google migrated onto it first). Once it's up I'll rebase #7616 on top so Gitea becomes the second provider on that abstraction instead of more inline code.

If #5792 (generic OIDC) lands first, that covers our case too — Gitea is an OIDC provider — and I'm happy to close this in its favour. Until then we run this branch as a small patch on top of each upstream tag, so no rush on my side.

PR #7623, opened 2026-08-27
MUL-6734: fix(daemon): accept Windows set-path overrides

### LINKED ISSUE #7613: [Bug]: runtime profile path override is never accepted on Windows — the executable gate requires a Unix permission bit
**Deployment type:** either — the daemon runs on the user's machine in both cases
**Area:** Agents and runtimes
**Multica:** `main` @ `09a2410e8`
**Platform:** any Windows host

## What happened

`multica runtime profile set-path` is accepted by the CLI, but the daemon never uses the
override on Windows. Registration logs

```
custom runtime profile: command path override not executable; falling back to PATH
```

and the profile falls back to a `PATH` lookup. This happens for **every** path, not just
`.cmd` shims — a plain `.exe` is rejected the same way.

Surfaced as a side note while working around #7551 (thanks to the workaround log in that
report, which is what made this findable); filing separately because it is a different code
path and blocks a documented feature (MUL-3284) on its own.

## Root cause

`server/internal/daemon/daemon.go:264`

```go
profilePathExecutable = func(path string) bool {
	info, err := os.Stat(path)
	if err != nil || info.IsDir() {
		return false
	}
	return info.Mode().Perm()&0o111 != 0
}
```

The gate requires a Unix executable bit. Windows has none, and Go does not synthesize one:
`os/types_windows.go` (`fileStat.mode()`, Go 1.26.3 lines 177-196) sets `0666` for a writable
file and `0444` for a read-only one, and adds `0111` **only** for directories:

```go
if fs.FileAttributes&syscall.FILE_ATTRIBUTE_READONLY != 0 {
	m |= 0444
} else {
	m |= 0666
}
...
	if fs.FileAttributes&syscall.FILE_ATTRIBUTE_DIRECTORY != 0 {
		m |= ModeDir | 0111
	}
```

So for any file on Windows, `Perm()&0o111 == 0`, and the `info.IsDir()` guard above rejects
the one case that would have passed. The predicate is unsatisfiable there.

The consumer is `appendProfileRuntimes` at `server/internal/daemon/daemon.go:2786`, and the
comment there describes the intended behaviour — "a per-machine path override wins over the
PATH lookup when it is set AND points at a real executable" — which is correct on Unix and
unreachable on Windows.

## Steps to reproduce

1. On a Windows host running the daemon, install any agent CLI outside `PATH`.
2. `multica runtime profile set-path <profile> C:\path\to\tool.exe`
3. Restart the daemon and watch its log: the override is reported as not executable and the
   profile falls back to `PATH`. If the command name is not on `PATH`, the profile is skipped
   and the runtime never registers.

## Why it matters beyond the override

The same shape is what makes #7551 hard to work around: npm-installed CLIs on Windows are
`.cmd` shims (OpenClaw's own entry point is `openclaw.cmd`), so pinning a wrapper is the
natural escape hatch, and it is closed on the platform where it is most needed.

## Suggested fix

Split the predicate by platform, as the codebase already does for other process concerns:

- Unix: keep the existing `Perm()&0o111` check.
- Windows: existence plus not-a-directory is the meaningful test. Matching the extension
  against `PATHEXT` (`.EXE`, `.CMD`, `.BAT`, `.COM`, …) would additionally reject a path that
  Windows could not execute, which is closer to the intent of the current check than the
  permission bits are.

Worth a regression test per platform: the existing `stubProfilePathExecutable`
(`internal/daemon/runtime_profile_test.go:435`) replaces the predicate wholesale, so no test
today exercises the real one — which is why this went unnoticed.

leilei3167:
## What does this PR do?

`multica runtime profile set-path` is accepted on Windows, but the daemon never uses the override. The gate requires a Unix executable bit, and Go never sets one on Windows files (`0666`/`0444` only; `0111` is directories). Every path — including a plain `.exe` — is treated as not executable and we fall back to PATH.

Unix still checks `Perm()&0o111`. Windows checks that the file exists and its extension is in `PATHEXT` (`.exe` / `.cmd` shims included).

Fixes #7613

## Related Issue

Closes #7613

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/daemon.go`: `profilePathExecutable` now uses PATHEXT on Windows instead of unix exec bits
- `server/internal/daemon/profile_path_executable_test.go`: PATHEXT table tests plus a real-file unix-bits test (the existing override tests stub the predicate, so they never hit this)

## How to Test

1. `go test ./internal/daemon -count=1 -run 'TestExtListedInPathext|TestFileLooksLaunchable|TestRegisterRuntimes_Override'`
2. On a Windows daemon: `multica runtime profile set-path <profile> C:\path\to\tool.exe`, restart, confirm the log uses the override instead of `command path override not executable; falling back to PATH`

## Checklist

- [x] I have updated relevant documentation to reflect my changes, or not applicable

Unix behavior is unchanged. Windows now rejects non-PATHEXT files (e.g. `.txt`) rather than every file.

## AI Disclosure

**AI tool used:** Cursor

**Prompt / approach:**
Looked at the Windows `os.FileMode` behavior called out in #7613 and replaced the unix-bit gate with a PATHEXT check, then added tests that exercise the real predicate instead of the stub.


leilei3167:
@Bohan-J small Windows-only fix for `runtime profile set-path` — the unix exec-bit check was rejecting every override, including a plain `.exe`. Happy to adjust if you'd rather treat existence-only as enough and skip the PATHEXT match.

leilei3167:
Hi @Bohan-J — just a friendly ping on this PR whenever you have a moment. Happy to address any feedback. Thanks!

Bohan-J (changes requested):
Thanks for tracking this down — the diagnosis is right and the fix is worth taking. On Windows the gate is unconditionally false, so a pinned `set-path` either fails registration with a message that blames the user's filesystem, or silently registers whatever `PATH` resolves to — which defeats the point of pinning a path. Restoring it is a small, Windows-only change with no config, API, or data migration.

Two things should change before merge.

## 1. Reuse the existing executable-resolution contract, and keep the path it returns

`server/internal/daemon/daemon.go` (285-330) introduces a second definition of "launchable" (`os.Stat` + a hardcoded PATHEXT list), while the same package already has one: `config.go` (835-861), where `resolveAgentExecutablePath` / `agentExecutablePresent` use `exec.LookPath` — the check the agent backends apply at launch. This bug is essentially two competing notions of executability in one package; adding a third does not close that gap.

They already disagree observably:

- `C:\tools\openclaw` with only `openclaw.cmd` on disk — the npm shim case from #7613. `exec.LookPath` completes the extension and returns `C:\tools\openclaw.cmd`; this PR fails at `os.Stat` first (and `extListedInPathext` rejects an empty extension anyway) and falls back to PATH.
- `defaultWindowsPathext` accepts `.VBS`, `.VBE`, `.JS`, `.JSE`, `.WSF`, `.WSH`, `.MSC`. `CreateProcess` cannot launch those directly, so the new gate admits files that fail at start time while rejecting ones that would have worked.

Suggested shape: call `resolveAgentExecutablePath(override)` (or at minimum `exec.LookPath`) and assign its **return value** to `resolved` at `daemon.go:2846` rather than the raw override — otherwise an extension-less override passes the gate and then fails at launch. Keep the existing fallback and warning on error, and drop the custom PATHEXT parsing. An extension check is still useful as diagnostic wording in the failure log, just not as the gate.

## 2. Make the Windows runner actually execute the new regression

`profile_path_executable_test.go` (65-100) skips unless `runtime.GOOS == "windows"`, and the Windows daemon step in `.github/workflows/ci.yml` (673, "Test Windows agent executable junction resolution") uses an exact `-run` regex that does not list `TestFileLooksLaunchable_WindowsPathext`. The Windows-only root cause is therefore not exercised by the green CI on this PR. No new job needed — adding the test name to that existing regex is enough.

If you move to the resolver in (1), the test should assert the resolved path for `.exe`, `.cmd`, an extension-less override, a directory, and a missing path.

Happy to re-review as soon as those two are in.

leilei3167:
@Bohan-J thanks for the review — both points addressed:

1. `appendProfileRuntimes` now goes through `resolveAgentExecutablePath` / `LookPath` and assigns the returned path to `resolved` (so extension-less pins complete to `.cmd`). The custom PATHEXT gate is gone.
2. `TestResolveAgentExecutablePath_ProfileOverride` covers `.exe` / `.cmd` / extension-less / directory / missing, and it’s in the Windows CI `-run` regex. Also dropped a spurious `.txt` rejection assert that didn’t match `LookPath` on absolute paths (that was what red’d `windows-execenv`).

Happy to take another look whenever you have a moment.

Bohan-J (approved):
Both blockers are addressed — thanks for turning this around so quickly, and for taking the resolver route rather than patching the gate.

Verified at `c48fba721`:

- `appendProfileRuntimes` now goes through `resolveProfileOverridePath` → `resolveAgentExecutablePath` / `exec.LookPath` and assigns the **resolved** path, so an operator who pins `C:\tools\openclaw` while only `openclaw.cmd` exists gets the shim, and the daemon records the same path the launch contract would produce. The custom PATHEXT table is gone, so there's one definition of "launchable" in the package again.
- `TestResolveAgentExecutablePath_ProfileOverride` is in the `-run` regex for "Test Windows agent executable junction resolution", and I confirmed it actually executed on the Windows runner: `=== RUN` / `--- PASS` in the `windows-execenv` log for this commit. The extension-less shim case is asserted there, which is exactly the path the old gate could never reach.
- Dropping the `.txt` rejection assert is the right call — `exec.LookPath` accepts an absolute path to any existing file on Windows, so asserting otherwise would have locked in a behavior the resolver doesn't promise. The comment you left in its place says so, which is better than the assert was.

Fallback semantics, the warning log, the CLI config format and the server API are all unchanged, so the blast radius stays where it was: Windows hosts using a profile path override.

One non-blocking nit, no need to hold the merge: the PR description still describes the PATHEXT implementation and the old test names from the first revision. Worth a quick edit if you want the page to match the code.

Merging. Thanks for the clear reproduction in #7613 and for the follow-through here.

PR #7636, opened 2026-08-27
fix(issue-status): derive a stable key for all-non-ASCII status names

### LINKED ISSUE #7627: [Bug]: Custom status creation fails for non-ASCII-only names
### Deployment type

self-host

### Area

### What happened?

Creating a custom issue status from Workspace Settings fails when the display name contains only non-ASCII characters, for example `客户确认`.

The UI reports:

```text
cannot derive a status key from that name; provide one explicitly
```

The settings form asks for a display name but does not provide a field for an explicit key. As a result, there is no way to create this status through the UI without changing the user-facing name to include ASCII letters or digits.

I expected valid Unicode display names to work. When automatic slug generation produces an empty result, the application could generate a stable internal key, or the UI could offer an explicit-key fallback.

This is not specific to Chinese; it can affect names written entirely in other non-Latin scripts as well.

### Steps to reproduce

1. Open **Workspace Settings** as an owner or admin.
2. Go to **Issue Statuses**.
3. Add a custom status in any category.
4. Set the name to `客户确认` and save.
5. Observe that creation fails with the error above.

### Screenshots (optional)

Not included; the error text and reproduction steps are sufficient.

The create-status API accepts an optional explicit `key`, but the settings UI submits the name, description, category, and color without one. The server then derives a key using ASCII letters and digits; a name made entirely of non-ASCII characters produces no usable key.

Temporary workaround: include ASCII letters or digits in the display name, or call the API with an explicit valid key.

importcpp:
## Problem

Fixes #7627.

Creating a custom issue status from **Workspace Settings → Issue Statuses**
fails when the display name contains only non-ASCII characters (e.g. `客户确认`):

> cannot derive a status key from that name; provide one explicitly

`SlugifyKey` (`server/internal/issuestatus/issuestatus.go`) derives the
status key by keeping only `[a-z0-9]`, so an all-non-Latin name slugifies to
an empty string and is rejected. The settings form submits only
name/description/category/color and exposes **no explicit-key field**, so
there is no way to create such a status through the UI. This affects any name
written entirely in a non-Latin script, not just Chinese.

## Fix

Status keys are ASCII schema-level identifiers (per the conventions doc:
`todo`, `in_progress`), while the display name carries the localized text.
Rather than reject a name that slugifies to empty, derive a **stable
content-based fallback key**:

```
status_<sha256(lower(trim(name)))[:12]>
```

- **Deterministic & case-insensitive** — recreating a status by the same name
  (e.g. after archiving) lands on the same key, matching the existing ASCII
  slug path.
- The `status_` prefix satisfies the leading-letter rule of the key pattern
  and can never collide with a built-in key.
- **Server-only change** — keeps the deliberately thin settings form (one
  name field) intact; no frontend changes.

## Testing

- `go test ./internal/issuestatus/ ./internal/handler/` — green on a freshly
  migrated DB. New coverage:
  - `TestSlugifyKey`: non-ASCII name now yields a valid, stable, non-colliding key.
  - `TestCreateIssueStatusFromNonASCIIName`: `POST /api/issue-statuses` with
    `客户确认` returns **201** with a valid ASCII key.
- **End-to-end** (`make dev`, real HTTP → API → DB):
  - `客户确认` → 201, name preserved, key `status_7d11b1e1782e`.
  - `待审核` → distinct key `status_7bf25421c6df` (no collision).
  - Re-submitting `客户确认` → 409 (same key), confirming determinism.
  - Catalog list endpoint renders both custom statuses correctly.

### DIFF
--- a/server/internal/handler/issue_status_test.go
+++ b/server/internal/handler/issue_status_test.go
@@ -369,6 +369,28 @@ func TestCreateIssueStatusIsNotGatedOnRollout(t *testing.T) {
 	dbfx.Cleanup(t, `DELETE FROM issue_status WHERE id = $1`, parseUUID(created.ID))
 }
 
+// TestCreateIssueStatusFromNonASCIIName pins the #7627 fix: a display name with
+// no slug-able characters (e.g. `客户确认`) must create through the name-only
+// form, since the settings UI offers no explicit-key field. The derived key is
+// a valid ASCII schema identifier.
+func TestCreateIssueStatusFromNonASCIIName(t *testing.T) {
+	seedTestCatalog(t)
+
+	var created IssueStatusResponse
+	testutil.Call(t, testHandler.CreateIssueStatus,
+		newRequest(http.MethodPost, "/api/issue-statuses", map[string]any{
+			"name": "客户确认", "category": issuestatus.InReview, "color": "#123456",
+		})).Want(http.StatusCreated).JSON(&created)
+	dbfx.Cleanup(t, `DELETE FROM issue_status WHERE id = $1`, parseUUID(created.ID))
+
+	if created.Name != "客户确认" {
+		t.Errorf("stored name = %q, want the Unicode display name", created.Name)
+	}
+	if _, err := issuestatus.ValidateKey(created.Key); err != nil {
+		t.Errorf("derived key %q is not a valid status key: %v", created.Key, err)
+	}
+}
+
 // TestIssueWriteStoresCanonicalStatusKey guards the 500 found in review:
 // resolution is case- and whitespace-insensitive, so writing the caller's raw
 // string back would store a value the column's format constraint rejects.
--- a/server/internal/issuestatus/issuestatus.go
+++ b/server/internal/issuestatus/issuestatus.go
@@ -24,6 +24,8 @@ package issuestatus
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"regexp"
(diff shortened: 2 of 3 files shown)


Bohan-J:
Thank you for this — and I owe you a straight account of what happened.

You opened this PR at 09:40 UTC; I opened #7638 at 10:11, half an hour later, without checking whether you had already pushed. You called the bug correctly, you located it correctly, and you got there first. That was my mistake, and I'm sorry for the wasted effort.

#7638 has now been merged (d6ecf4bc8), so I'm closing this one as superseded rather than because there was anything wrong with the diagnosis. Since you did the work, here is what actually differs and why — you should be able to judge the call for yourself.

**Fallback shape.** Both PRs agree on the important thing: the key stays an ASCII machine handle, the Unicode lives on the display name, and no migration is involved. Where they differ is what the fallback looks like. This one produces `status_a3f9c21b4e7d`; the merged change produces `<category>_<n>` — `in_review_2`. The reasoning is that the key is not only stored, it is also what an agent is told to type, and the workspace status catalog is injected into an agent's runtime brief as `` `key` (Name — description) ``. A category-anchored key still states which platform behavior the status inherits, and a model that confuses `in_review_2` with `in_review` lands in the right category rather than nowhere; a hash degrades to nothing. It is also a good deal easier to type on the CLI.

**One thing worth knowing regardless of which shape wins.** The PR description lists as a feature that "recreating a status by the same name (e.g. after archiving) lands on the same key". That property actually backfires here: `idx_issue_status_workspace_key` is **not** a partial index (see migration `335_issue_status_workspace_key_index.up.sql` — no `WHERE archived_at IS NULL`, unlike the name index in `336`), so an archived status still owns its key. Deriving the same key from the same name would hit the unique index and fail with a 409 — on the exact path that has no key field to correct. Any content-derived scheme needs to be disambiguated against the catalog including archived rows.

**Scope that turned out to be needed.** Review surfaced three more things in the same path, which is most of why the merged change is larger:

- Mixed names silently drop their non-ASCII part, so `待客户 Review` and `待供应商 Review` both slug to `review` and the second create fails with a conflict blaming a display name nobody had taken. Derived keys are now disambiguated, archived rows included.
- Derivation reads the catalog to pick a free key, so the read and the insert have to be atomic — every create now takes the exclusive catalog lock, including one that supplies its own key. Leaving explicit-key creates outside it kept the race half-open.
- A derived key is opaque, so the payloads and errors that printed a key alone now carry the display name beside it.

None of that is a criticism of this PR — it is scope that came out of review, and you had no way to know about it up front.

Genuinely: the analysis in your first comment on #7627 was accurate and the fix direction was sound. Please do keep sending patches. If it helps next time, commenting on the issue right before you push (rather than only when you start) makes a collision like this one much easier to catch.

PR #7643, opened 2026-08-27
MUL-6755: fix(search): remove command palette motion

imaffe:
## Summary
- make the high-frequency command palette snap open and closed
- preserve normal modal animation elsewhere
- add regression coverage and the completed animation audit plan

## Verification
- @multica/views tests (4,793 passed)
- @multica/views typecheck
- @multica/views lint (0 errors)

### DIFF
--- a/packages/views/search/search-command.test.tsx
+++ b/packages/views/search/search-command.test.tsx
@@ -304,6 +304,15 @@ describe("SearchCommand", () => {
     });
   });
 
+  it("disables dialog animation in both open and closed states", () => {
+    renderSearch();
+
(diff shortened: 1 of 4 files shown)


multica-eve:
Thanks for the contribution. We reviewed the behavior from first principles and are going to pass on this change. The command palette already responds immediately; the 100ms value is the duration of an intentional visual transition, not an interaction delay, and we do not have user reports or measurements showing a usability problem. A command-palette-specific exception would diverge from the shared modal behavior without demonstrated product benefit. If an accessibility need emerges, we should address it consistently through reduced-motion behavior in the shared UI. Closing as not planned.

PR #7645, opened 2026-08-27
MUL-6759: feat(core): add truncateWithEllipsis utility

### LINKED ISSUE #7544: Add truncateWithEllipsis utility to packages/core/utils.ts
Add a truncateWithEllipsis(text: string, maxLength: number): string utility function to packages/core/utils.ts, following the existing code style in that file. It should truncate text longer than maxLength and append an ellipsis (...), while leaving shorter text unchanged. Add a corresponding test in utils.test.ts following the existing test patterns in that file. Do not change any existing function.

importcpp:
## What does this PR do?

Adds a `truncateWithEllipsis(text: string, maxLength: number): string` utility to `packages/core/utils.ts`, following the existing code style in that file, plus a matching test suite in `packages/core/utils.test.ts`.

## Related Issue

Closes #7544

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Changes Made

- `packages/core/utils.ts`: new exported `truncateWithEllipsis` function.
  - Text with `length <= maxLength` is returned unchanged (no ellipsis).
  - Longer text is sliced to `maxLength - 3` characters with `"..."` appended, so the result's total length equals `maxLength`.
  - The ellipsis is counted toward the budget and the result is clamped so it never exceeds `maxLength`, including when `maxLength < 3` (slice length clamps to 0, trailing dots trimmed to `maxLength`).
- `packages/core/utils.test.ts`: new `describe("truncateWithEllipsis", ...)` block following the file's existing structure.

No existing function, export, or test was modified, reordered, or removed. No new dependencies, no build/lint/config changes.

## How to Test

```bash
cd packages/core
pnpm test        # 19 tests pass, incl. the new truncateWithEllipsis suite
pnpm typecheck   # clean
pnpm lint        # clean
```

New tests cover: shorter-than-limit unchanged, exactly-at-limit unchanged, longer text truncated with ellipsis, total length equals `maxLength`, and the small-`maxLength` edge case (`<= 3`).

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective / my feature works
- [x] New and existing unit tests pass locally with my changes

## AI Disclosure

This change was implemented with the assistance of Claude Code.

### DIFF
--- a/packages/core/utils.test.ts
+++ b/packages/core/utils.test.ts
@@ -1,5 +1,11 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { createRequestId, createSafeId, generateUUID, isImeComposing } from "./utils";
+import {
+  createRequestId,
+  createSafeId,
+  generateUUID,
+  isImeComposing,
+  truncateWithEllipsis,
+} from "./utils";
 
 afterEach(() => {
   vi.unstubAllGlobals();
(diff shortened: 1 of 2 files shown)


multica-eve:
Thank you for taking the time to contribute this and for adding focused tests.

We are closing this PR because it adds a public truncation utility without connecting it to a concrete product flow or an existing caller. The linked issue specifies an implementation, but it does not identify a user-facing problem, an affected surface, or a workflow that this change completes. As a result, merging it would create a shared API and an ongoing maintenance commitment without changing anything users can do or see.

The repository also has several context-specific truncation needs with different semantics: layout-based ellipsis, a single Unicode ellipsis, middle truncation for identifiers and paths, and preview-specific limits. This helper would prematurely make one particular behavior the shared contract—three ASCII periods, UTF-16 code-unit slicing, and partial periods for very small limits—without evidence that those choices are right across those contexts. The tests verify the chosen behavior, but they cannot establish the product need or the correct shared semantics.

If a concrete use case comes up, we would prefer a future change to include that real caller and define the truncation behavior from its product context. A common helper should be extracted only when multiple callers genuinely share the same rules.

Thank you again for the contribution and for the care you put into it.

PR #7646, opened 2026-08-27
MUL-6758 feat(core): add truncateWithEllipsis string helper

### LINKED ISSUE #7544: Add truncateWithEllipsis utility to packages/core/utils.ts

ADanMan:
## What
Adds an exported `truncateWithEllipsis(text, maxLength)` helper to `packages/core/utils.ts`: returns the text unchanged when it fits within `maxLength`, otherwise truncates and appends a trailing `...` so a normal result is exactly `maxLength` chars.

## Why
Closes #7544.

## Tests
New co-located unit tests in `packages/core/utils.test.ts` cover unchanged-short-text, exactly-at-limit, normal truncation, the length bound, and the `maxLength <= 3` edge case. `tsc --noEmit` and eslint are clean; `vitest run utils.test.ts` is green.

### DIFF
--- a/apps/desktop/src/renderer/src/components/daemon-panel.tsx
+++ b/apps/desktop/src/renderer/src/components/daemon-panel.tsx
@@ -15,6 +15,7 @@ import {
   Trash2,
   X,
 } from "lucide-react";
+import { truncateWithEllipsis } from "@multica/core/utils";
 import { cn } from "@multica/ui/lib/utils";
 import { copyText } from "@multica/ui/lib/clipboard";
 import { Button } from "@multica/ui/components/ui/button";
@@ -652,7 +653,9 @@ function EmptyState({
 // ---------- Helpers ----------
 
 function truncateValue(value: string, max = 32): string {
-  return value.length > max ? `${value.slice(0, max)}…` : value;
+  // `max` is a content-character budget; the shared helper counts the ellipsis
+  // toward its own budget, so pass `max + 1`.
+  return truncateWithEllipsis(value, max + 1);
 }
 
 function highlight(text: string, query: string): ReactNode {
--- a/packages/core/utils.test.ts
+++ b/packages/core/utils.test.ts
@@ -69,19 +69,40 @@ describe("truncateWithEllipsis", () => {
     expect(truncateWithEllipsis("hello", 5)).toBe("hello");
   });
 
-  it("truncates longer text and appends an ellipsis", () => {
-    expect(truncateWithEllipsis("hello world", 8)).toBe("hello...");
-  });
-
-  it("keeps the truncated result within maxLength", () => {
+  it("truncates longer text and appends an ellipsis within maxLength", () => {
     const result = truncateWithEllipsis("hello world", 8);
-    expect(result).toHaveLength(8);
-    expect(result.endsWith("...")).toBe(true);
+    expect(result).toBe("hello w…");
+    expect(Array.from(result)).toHaveLength(8);
+    expect(result.endsWith("…")).toBe(true);
   });
 
-  it("returns only the ellipsis when maxLength is at or below the ellipsis length", () => {
-    // No room for visible text, so the truncated output is just the ellipsis.
-    expect(truncateWithEllipsis("hello", 3)).toBe("...");
-    expect(truncateWithEllipsis("hello", 1)).toBe("...");
+  it("trims trailing whitespace before the ellipsis", () => {
+    // The cut lands right after the space; it must not leave "hello …".
+    expect(truncateWithEllipsis("hello world", 7)).toBe("hello…");
+  });
+
+  it("never splits an astral character into a lone surrogate", () => {
+    const result = truncateWithEllipsis("🎉🎉🎉🎉🎉 party", 8);
+    expect(result).toBe("🎉🎉🎉🎉🎉 p…");
+    expect(Array.from(result)).toHaveLength(8);
+    // No unpaired high surrogate left dangling.
+    expect(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/.test(result)).toBe(false);
+  });
+
+  it("measures the limit in code points, not UTF-16 code units", () => {
+    // Five emoji are 10 code units but 5 visible characters, so they fit
+    // within maxLength 8 and must not be truncated.
+    expect(truncateWithEllipsis("🎉🎉🎉🎉🎉", 8)).toBe("🎉🎉🎉🎉🎉");
+  });
+
+  it("truncates CJK content by code point", () => {
+    expect(truncateWithEllipsis("你好世界你好世界", 5)).toBe("你好世界…");
+  });
+
+  it("handles degenerate maxLength values without throwing", () => {
+    expect(truncateWithEllipsis("hello", 1)).toBe("…");
+    expect(truncateWithEllipsis("hello", 0)).toBe("");
+    expect(truncateWithEllipsis("hello", -5)).toBe("");
+    expect(truncateWithEllipsis("hello", Number.NaN)).toBe("…");
   });
 });
--- a/packages/core/utils.test.ts
+++ b/packages/core/utils.test.ts
@@ -1,5 +1,11 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { createRequestId, createSafeId, generateUUID, isImeComposing } from "./utils";
+import {
+  createRequestId,
+  createSafeId,
+  generateUUID,
+  isImeComposing,
+  truncateWithEllipsis,
+} from "./utils";
 
 afterEach(() => {
   vi.unstubAllGlobals();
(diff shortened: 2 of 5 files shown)


Bohan-J (changes requested):
Thanks for the PR, and for the clean, well-documented implementation — the JSDoc explaining that the ellipsis counts toward the budget is exactly the kind of non-obvious decision that deserves a comment.

To be upfront about where this is going: **the implementation faithfully matches the acceptance criteria in #7544. The problem is with #7544 itself.** That issue was filed by a non-maintainer and its "Non-goals" section rules out two things this repo has already decided it needs. So please read the below as "the spec you were handed was wrong", not "you implemented it wrong".

## The main blocker: this helper can't replace any existing truncation site

We already truncate text in 16 places across chat, tasks, runtimes, the editor, desktop, and mobile. As written, this helper doesn't match any of them on two axes:

- **Ellipsis character.** 10 of those 16 use `…` (U+2026); 6 use `...`. The ones with the most careful implementations (`apps/web/platform/document-title.ts`, `packages/views/issues/components/description-preview.ts`, `packages/views/common/github-url.ts`) all use `…`.
- **Length budget.** Almost all existing sites mean "keep N characters of content, then append the ellipsis" (result is N+1 chars). This PR means "total length including the ellipsis is at most N". Both are reasonable, but they're different contracts.

So merging this as-is adds a 17th way to truncate rather than consolidating the existing 16 — and the next person who tries to unify them has to do archaeology first. That's the thing we'd like to avoid.

## Correctness issues (independent of the convention questions)

**1. Splits surrogate pairs.** `text.slice()` cuts on UTF-16 code units, so an astral character straddling the cut becomes a lone surrogate:

```js
truncateWithEllipsis("🎉🎉🎉🎉🎉 party", 8)
// → "🎉🎉\ud83c..."   renders as 🎉🎉�...
```

This isn't theoretical here — issue titles routinely open with an emoji. The repo has already paid for this lesson twice, and both fixes carry comments explaining it: `apps/web/platform/document-title.ts:38-47` iterates code points, and `packages/views/issues/components/description-preview.ts:52-57` detects a trailing high surrogate and drops the character whole. #7544's non-goal "do not add Unicode-grapheme-aware truncation" is directly at odds with those.

**2. No `trimEnd()` before appending.** A cut that lands on a space leaves the ellipsis dangling:

```js
truncateWithEllipsis("hello world", 9)   // → "hello ..."
```

`use-chat-controller.ts:72`, `description-preview.ts:57`, and `document-title.ts:46` all `trimEnd()` first.

**3. The documented contract doesn't hold at the low end.** The JSDoc says a truncated result "has a total length of `maxLength`", but:

```js
truncateWithEllipsis("abcdef", 3)    // → "..."  (len 3, ok)
truncateWithEllipsis("abcdef", 0)    // → "..."  (len 3 > 0)
truncateWithEllipsis("abcdef", -5)   // → "..."
truncateWithEllipsis("abcdef", NaN)  // → "..."
```

A hard total-length bound is the whole point of this semantics — fixed-width layout — and this is precisely where it stops holding. Either tighten the implementation or restate the contract in the JSDoc, but the two shouldn't disagree.

## On the tests

The five cases map one-to-one onto #7544's acceptance list, which means they demonstrate "the spec was implemented" rather than "this behaves well on our content". Nothing covers emoji, CJK, trailing whitespace, or `maxLength` of 0 / negative / `NaN`.

Worth flagging specifically: the `maxLength <= 3` case is asserted as intended behavior, which locks in the contract gap from point 3 above and makes it harder to correct later.

## What we'd like

Recommended shape, which keeps your total-length semantics (it's the more useful contract — a real upper bound) while fitting repo conventions:

```ts
export function truncateWithEllipsis(text: string, maxLength: number): string {
  if (text.length <= maxLength) return text;
  if (maxLength <= 1) return maxLength <= 0 ? "" : "…";
  // Slice by code point so an emoji is never cut into a lone surrogate.
  const chars = Array.from(text);
  if (chars.length <= maxLength) return text;
  return `${chars.slice(0, maxLength - 1).join("").trimEnd()}…`;
}
```

Two notes on that sketch:

- The `text.length` check is only a fast path — it can fail to return early but never truncates wrongly; the code-point check right after it is the authoritative one. That second check also fixes a bug in the current guard: `text.length` overcounts astral characters, so today `"🎉🎉🎉🎉🎉"` reports length 10 and gets truncated at `maxLength = 8` even though it displays as 5 characters.
- This does move the bound from UTF-16 code units to code points, so `truncateWithEllipsis("🎉🎉🎉🎉🎉 party", 8)` returns a string of 8 code points but 13 code units. That's the right tradeoff for a display helper — callers care about what the user sees — but it's worth stating in the JSDoc.

Concretely, for this PR:

1. Fix the three correctness points above.
2. Extend the tests to cover emoji, CJK, a cut landing on whitespace, and the degenerate `maxLength` values.
3. **Convert at least a few existing call sites in the same PR** so the helper earns its place and gets exercised against real content. The closest matches semantically are `apps/desktop/src/renderer/src/components/daemon-panel.tsx:654` (`truncateValue`), the four inline truncations in `apps/mobile/components/chat/chat-timeline.tsx`, and `packages/views/common/task-transcript/trace-event-presenter.ts:92`. Each is `slice(0, N) + ellipsis`, so `truncateWithEllipsis(value, N + 1)` is equivalent — just bump the constant by one. Full migration of all 16 sites is out of scope; two or three is enough to prove the API.

One thing to confirm before you invest in step 3: **is `…` the ellipsis we're standardizing on?** The evidence points that way, but it touches user-visible output in several files, so let's have a maintainer confirm it here first — I don't want you doing the work twice.

Happy to keep reviewing as you iterate.

ADanMan:
Thanks — genuinely useful review, and the "the spec was wrong" framing is fair.

Pushed `f135f449` addressing everything except the call-site conversion:

- **Code-point safe.** Slices via `Array.from` and measures the bound in code points, so an emoji is never cut into a lone surrogate. That also fixes the fast-path over-count you flagged — `"🎉🎉🎉🎉🎉"` (5 visible / 10 code units) is now left untouched at `maxLength` 8.
- **`trimEnd()`** before appending, so a cut landing on whitespace no longer leaves a dangling ellipsis.
- **Low end defined.** `maxLength <= 0` → `""`, `maxLength === 1` → `…`; JSDoc restated to match, and to note the code-point vs code-unit distinction.
- **Ellipsis** is now `…` (U+2026).
- **Tests** extended to cover emoji (no surrogate split), the code-point-vs-code-unit bound, CJK, a cut on whitespace, and the degenerate `maxLength` values (0 / negative / NaN). I adopted your recommended shape almost verbatim — it was the right contract.

I deliberately did **not** touch any call sites yet: you asked to confirm `…` as the standard first, and I'd rather not do the work twice. Once you confirm, I'll convert the closest matches in this PR — `daemon-panel.tsx` `truncateValue`, the inline truncations in `chat-timeline.tsx`, and `trace-event-presenter.ts:92` (each is `slice(0, N) + ellipsis`, so `truncateWithEllipsis(value, N + 1)`) — and leave the full 16-site migration for a follow-up.

So: is `…` the ellipsis to standardize on?

multica-eve:
Confirmed: please standardize this shared helper and the migrated call sites on the single-character Unicode ellipsis `…` (U+2026). The hard total-length budget remains the intended contract.

Please go ahead with the 2–3 existing call-site conversions you proposed (`daemon-panel.tsx`, the matching inline truncations in `chat-timeline.tsx`, and/or `trace-event-presenter.ts`), and update or add tests for any changed observable output. A full migration of every truncation site is not required in this PR.

Once those changes are pushed, let us know and we’ll re-review.

ADanMan:
Done — pushed `68d7865` converting two call sites to the shared helper and standardizing on `…`:

- **`packages/views/common/task-transcript/trace-event-presenter.ts`** `clip()` (was `...`) → `truncateWithEllipsis(value, max + 1)`. This is the visible `...` → `…` change; added a `traceEventSummary` truncation test asserting the single-character ellipsis and the code-point length.
- **`apps/desktop/.../daemon-panel.tsx`** `truncateValue()` (already `…`) → the shared helper, so it's now code-point-safe and trims trailing whitespace as well.

Each site was `slice(0, N) + ellipsis`, so `+1` on the content budget keeps the visible length otherwise unchanged. I left the mobile `chat-timeline.tsx` inline truncations and the full 16-site migration for a follow-up, per "2-3 is enough". Ready for re-review.

Bohan-J (approved):
Thanks for sticking with this through three rounds — the PR ended up somewhere much better than where it started, and that was your work, not the review's.

Worth calling out specifically:

- You took the "the spec you were handed is wrong" framing and ran with it rather than defending the original #7544 acceptance criteria. Switching to `…`, making the cut code-point-safe, and `trimEnd()`-ing before the ellipsis all went against that issue's stated non-goals, and all three were the right call.
- The JSDoc is genuinely good. Spelling out that the bound is measured in code points and that the returned string may therefore exceed `maxLength` in `.length` is exactly the kind of thing that saves the next reader a debugging session.
- The `max + 1` comments at both call sites explain the one thing a future reader would otherwise get wrong.
- And you migrated real call sites instead of leaving a zero-caller utility behind — that's what turned this from "a function nobody calls" into something the codebase actually uses.

One deliberate boundary change that we're accepting knowingly: content of length exactly `max + 1` is now shown in full rather than truncated, since the old code spent an ellipsis to hide a single character. That's an improvement, not a regression.

Merging now. We'll follow up separately with the remaining mechanically-safe truncation sites (the web chat previews, the chat title, the link hover-card fallback, and the mobile chat timeline) so they converge on this helper too. A few sites are staying as they are on purpose — `shortDaemonId` uses a threshold and a cut length that are deliberately different numbers, `truncateUrl`'s main branch preserves the URL origin, and `midTruncate` is middle-truncation — none of those fit this helper's shape.

Thanks again.

Issue #7606, opened 2026-08-26
[Feature]: Model alias map + automatic pricing for runtime-reported model ids (harness provider ≠ serving provider)


adamflagg-neo:
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

When a local Claude Code runtime is pointed at a non-Anthropic serving endpoint (`ANTHROPIC_BASE_URL`, e.g. Fireworks' Anthropic-compatible endpoint) with the model aliases remapped (`ANTHROPIC_DEFAULT_FABLE_MODEL=kimi-fast-latest[1m]` etc.), usage rows are recorded as `(provider = harness type, model = raw model id)` — e.g. `(claude, kimi-fast-latest[1m])`. The cost engine appears to price by looking that composite key up in the pricing table, so aliased/custom models never resolve: cost stays 0 and every token accumulates in `uncosted_*`.

Observed on our workspace (90-day pull via `multica runtime usage`): **every** local-runtime model key is uncosted — `claude/kimi-fast-latest[1m]`, `claude/glm-fast-latest[1m]`, `claude/deepseek-flash-latest`, and even genuine `claude/claude-opus-5` — ~134M uncosted input tokens on a single runtime. So this may be broader than aliases: pricing may not be resolving for local-runtime rows at all.

Two normalization wrinkles visible in the data: Claude Code appends a `[1m]` context-variant suffix (both `deepseek-flash-latest` and `deepseek-flash-latest[1m]` appear), and renamed aliases leave stale forms behind (`claude/deepseek-v4-flash`).

Related: #4021 (manual price entries — same gap, manual-fix angle), #3542 (server-side pricing overrides for Azure Foundry — same seam, different provider), #5954 (Kimi K3 usage missing on issue cards — likely the same unrecognized-model-key family).

### Proposed solution (optional)

1. **Workspace-scoped model alias map**: `(provider, model)` → canonical pricing key, with `[1m]`-style suffix normalization. The map our workspace needs today, for concreteness: `(claude, kimi-fast-latest)` → `fireworks_ai/kimi-k3-fast`; `(claude, glm-fast-latest)` → `fireworks_ai/glm-5p2-fast`; `(claude, deepseek-flash-latest)` → `fireworks_ai/deepseek-v4-flash`.
[...]

PR #7648, opened 2026-08-27
fix(metrics): normalize the [1m] context tag before pricing lookup

ADanMan:
## What
`PriceForModelAlias` now strips a trailing `[...]` context-variant tag and retries the pricing-table lookup when the raw id misses. Claude Code appends a `[1m]` context-window suffix to runtime-reported model ids (e.g. `grok-4.5[1m]`, `kimi-k3[1m]`), and the alias rules are anchored with `$` and carry no optional bracket group, so those ids missed every rule and fell into the uncosted branch.

The raw id is tried first, so already-priced rows are untouched, and rules that already carry an inline optional bracket group still resolve on the first pass, preserving the existing no-false-borrowing guarantees. This mirrors the frontend's `stripContextTag` in `packages/views/runtimes/utils.ts`, so backend and frontend now agree.

## Why
Addresses the `[1m]` context-tag normalization wrinkle called out in #7606. Scoped to that sub-problem only: it does not implement the broader workspace-scoped alias map, automatic pricing, manual overrides, or backfill from that issue, so it does not fully close it.

## Tests
New `TestPriceForModelAliasContextTagStripping` (bracketed genuine ids resolve; empty and unterminated brackets stay unmapped). `go build ./...`, `go vet ./internal/metrics/...`, and `gofmt -l` are clean; `go test ./internal/metrics/...` is green, including the pre-existing no-false-borrowing test.

### DIFF
--- a/server/internal/metrics/pricing.go
+++ b/server/internal/metrics/pricing.go
@@ -172,13 +172,37 @@ var modelAliasRules = []struct {
 	// model identity, so it stays unmapped.
 }
 
-func PriceForModelAlias(model string) (ModelPrice, bool) {
-	model = strings.ToLower(strings.TrimSpace(model))
+// contextTagRe matches a trailing context-window variant tag such as the
+// `[1m]` Claude Code appends to the model id. A complete bracket tag with at
+// least one character inside, anchored at the end — the same shape the
+// frontend's `stripContextTag` strips (`\[[^\]]+\]$` in
+// packages/views/runtimes/utils.ts), so empty tags (`model[]`) and non-tag
+// trailing brackets (`model[`) stay unmapped on both sides.
+var contextTagRe = regexp.MustCompile(`\[[^\]]+\]$`)
+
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks a lot for this, @ADanMan! 🙏

The fix is exactly the shape we'd have picked ourselves: keep the explicit full-name price lookup first, and only strip a complete, non-empty trailing `[...]` tag as a fallback. That mirrors the rule the usage page already uses, so the server-side cost pipeline and the frontend now agree instead of diverging.

We also appreciate the tests covering both the positive cases (Grok / Codex / Kimi) and the negative ones (`model[]`, `model[`, `model[1m]-extra`) — those are the edge cases that keep the normalization from quietly borrowing prices for unknown variants.

One note for tracking: this resolves the context-tag part of #7606, but not the broader alias/pricing story there (custom aliases like `kimi-fast-latest` still won't get a price), so we'll keep #7606 open.

CI is green and this is merging now. Thanks again for the contribution!

PR #7649, opened 2026-08-27
MUL-6757: fix: make task transcript delivery loss-detecting

golb:
## Summary

- replace silent agent message drops with context-aware backpressure and task-scoped abandonment detection across every backend
- make transcript batch delivery retry-safe through an advertised server capability, `(task_id, seq)` uniqueness, atomic idempotent replay, and conflicting-payload rejection
- verify a contiguous transcript high-watermark in the same task-row transaction that completes or fails a task
- drain buffered cancellation tails before returning; fail closed when a stream never closes or delivery is truncated

## Why

The daemon previously cleared each in-memory batch before a one-shot HTTP request. A transient request or insert failure could therefore leave a task terminal while its persisted transcript was incomplete. Agent adapters could also silently drop messages whenever their 256-entry channel filled.

This change fixes the shared delivery boundary instead of adding adapter-specific guards. New daemons retry only when the claimed task advertises `transcript_batch_replay_safe`; old servers retain the existing one-shot behavior, avoiding duplicate rows during mixed-version operation.

## Verification

- pinned `sqlc/sqlc:1.31.1 generate`
- DB-backed replay, conflicting replay rollback, invalid sequence, cancelled final drain, complete/fail receipt, exact terminal replay, and late-write seal tests
- daemon retry, drain, cancellation-tail, and never-closing-stream regressions
- agent backpressure and late-reader regressions across affected stream wrappers
- migration cleanup-map contract tests
- targeted daemon/agent tests under `-race`
- compile-only check and `go vet` for all changed packages
- `git diff --check`

## Upgrade notes

Migration 440 deliberately fails closed if historical duplicates already exist. Before production migration, check:

```sql
SELECT task_id, seq, COUNT(*)
FROM task_message
GROUP BY task_id, seq
HAVING COUNT(*) > 1;
```

If rows are returned, inspect and reconcile them before retrying the migration; do not delete them automatically.

The official Helm backend uses `Recreate`, which avoids a mixed server fleet. Custom multi-replica deployments should roll the server capability fleet-wide before relying on daemon retries.

## Scope boundary

The receipt proves persistence of every `agent.Message` formed during the current process lifetime. Crash-durable daemon restart replay, raw provider-byte preservation, and first-class failed/cancelled receipt storage remain separate follow-up work.

### DIFF
--- a/server/cmd/migrate/main.go
+++ b/server/cmd/migrate/main.go
@@ -274,6 +274,7 @@ var concurrentIndexCleanups = map[string]string{
 	"438_agent_runtime_online_last_seen_index":                  "idx_agent_runtime_online_last_seen",
 	"439_agent_runtime_offline_last_seen_index":                 "idx_agent_runtime_offline_last_seen",
 	"440_github_pr_head_sha_index":                              "idx_github_pull_request_head_sha",
+	"441_task_message_task_seq_unique_index":                    "task_message_task_id_seq_uidx",
 }
 
 // concurrentDownIndexCleanups covers every migration whose down direction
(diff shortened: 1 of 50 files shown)


golb:
Superseded by #7844, rebuilt on the current main with migration conflicts resolved (441/442 -> 444/445). The replacement is patch-equivalent apart from migration numbering.

PR #7656, opened 2026-08-27
test(service): reconciler assertions only speak for their own rows

adaiguoguo:
`TestChannelMediaReconciler_SettlesThreeStates` (and five siblings) flake under a full `go test ./...` run and always pass standalone.

**Root cause.** The tests assert on the fake deleter's *global* key list (`len(deleted) == 1`, "nothing was deleted"). `RunOnce` sweeps the whole `channel_media_pending_object` table — the correct production semantic for a global reclaimer — and the test database is shared by every package the test runner executes in parallel: the wecom, dingtalk and handler suites all write that table. Under load, this test's sweep legitimately claims and deletes another package's due rows mid-assertion, and the count breaks. Reproduced on the first round of a three-package loop:

```
deleted keys = [workspaces/<ws>/wecom/<...> ×4  ws/lark/orphan], want only the orphan
```

**Fix.** `deletedOwn()` filters the recorded deletions to the fixture's `ws/lark/` prefix, preserving order, so per-key sequencing assertions keep their strength. Only the claim "and nothing else happened in the world" is dropped — which was never this suite's to make on a shared table. `CancelledSweepIsQuiet` keeps the global zero deliberately: a sweep with a cancelled context must not have touched anyone's rows, foreign ones included.

**Verification.** Before: reproduced on round 1 of `go test ./internal/service/ ./internal/handler/ ./cmd/migrate/ -count=1` in a loop. After: 6/6 clean rounds of the same loop; `-run TestChannelMediaReconciler -count=2` green.

### DIFF
--- a/server/internal/service/channel_media_reconciler_test.go
+++ b/server/internal/service/channel_media_reconciler_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


Bohan-J:
Thanks @adaiguoguo — this is a well-diagnosed fix, and the writeup made it easy to verify.

We confirmed the root cause independently: `ClaimNextChannelMediaPendingObjectForReconcile` has no workspace filter (correct production semantics for a global reclaimer), and `backend-tests` runs every package against one shared `DATABASE_URL`. Asserting on the deleter's global key list was never something this suite could hold. Scoping to the fixture's own rows while preserving order — and keeping the global zero in `CancelledSweepIsQuiet` — is the right set of trade-offs.

Two notes for the record, neither blocking this PR:

1. **The same root cause still has a reverse direction.** This PR stops other packages from polluting this test's count, but this test's `RunOnce` still sweeps and deletes *other* packages' rows. There's already a live target: `session_db_test.go` seeds a `state='deleting'` row with the default `next_attempt_at = now()` and a NULL lease, which is due the instant it is inserted — a concurrent sweep can tombstone it and fail that test's "untouched 'deleting'" assertion. That one blames an innocent package, so it is harder to trace than the flake you fixed.

2. **The `ws/lark/` prefix is safe by a narrow margin.** Three other test files use `ws/lark/`, but spelled `workspaces/ws/lark/...`, which `strings.HasPrefix(k, "ws/lark/")` does not match. It holds today; a future bare `ws/lark/` key would silently under-count instead of failing loudly.

Both are ours to follow up on — a fixture-owned unique prefix, or real per-package isolation, rather than another prefix convention. No changes needed here.

Merging. Thanks again for taking the time to root-cause this instead of adding a retry.

PR #7658, opened 2026-08-27
MUL-6753: fix(issues): preflight shared agents without runtime access

### LINKED ISSUE #7633: [Bug]: Regular members cannot create tasks via agent - misleading daemon CLI version error, works after promoting to admin
### Deployment type

self-host

### Area

Agents and runtimes

### What happened?

On our self-hosted instance, a **regular member** cannot create a task via the agent-assisted creation path. The UI shows this error:

> 该智能体的守护进程没有报告 CLI版本。通过智能体创建需要 multica CLI≥0.2.21。 请升级守护进猩并重连，或切换到手动创建。
>
> (Translation: "This agent's daemon did not report a CLI version. Creating via agent requires multica CLI ≥ 0.2.21. Please upgrade the daemon and reconnect, or switch to manual creation.")

However, this does **not** appear to be an actual CLI/daemon version problem. The same user, on the **same machine with the same daemon and CLI version**, succeeds immediately after being **promoted to admin**.

This suggests the daemon version probe (or the agent-assisted create path itself) is gated behind an admin-level permission check: for regular members the probe fails or returns empty, and the frontend falls back to the misleading "daemon did not report CLI version" error.

Expected behavior: a regular member with task-creation permission should be able to create tasks via the agent path, as long as their daemon CLI is ≥ 0.2.21.

### Steps to reproduce

1. On a self-hosted instance, connect an agent whose daemon runs CLI ≥ 0.2.21.
2. Log in as a **regular member** (non-admin) and create a task via the agent-assisted creation path.
3. Error appears: "daemon did not report CLI version, requires multica CLI ≥ 0.2.21".
4. Promote the same user to workspace admin.
5. Retry on the same machine/daemon — creation succeeds.

### Additional context

I searched existing issues (CLI version, daemon, quick create, member/non-admin permission) and found no matching report, so this appears to be unreported so far.

Bbbtt04:
## What does this PR do?

Fixes Quick Create for regular members who are allowed to invoke a shared agent but cannot view the private runtime that hosts it.

- Agent-list responses project the bound runtime's CLI version as `runtime_cli_version`.
- Quick Create uses that projection for its CLI compatibility preflight and skips the runtime-list request on upgraded backends.
- Older backends remain supported through the existing runtime-list fallback.
- The fallback now fails closed: Create stays disabled while the runtime query is pending or when it fails.

This does not grant runtime access or broaden agent invocation permission. It only moves the compatibility signal to the resource whose permission already governs the action.

## Root cause and permission contract

Quick Create previously resolved the selected agent's CLI version through `GET /api/runtimes`:

1. Read the selected agent's `runtime_id`.
2. Find that runtime in the caller's permission-filtered runtime list.
3. Read `runtime.metadata.cli_version`.

Agent invocation permission and runtime resource permission are intentionally separate. A member can be included in an agent's `public_to` invocation targets while still being unable to list, inspect, bind to, or manage the owner's private runtime. In that valid state, the runtime lookup returned no row and the UI incorrectly treated "not visible" as "daemon did not report a CLI version." Promoting the same member to admin made the preflight pass only because governance access made the runtime visible.

The new path keeps those boundaries intact:

- `memberAllowedToViewAgent` / the invocation allow-list still decides whether the agent is returned and can be invoked.
- Runtime list/detail/binding permissions are unchanged.
- The agent list exposes only the CLI compatibility signal, not the runtime name, owner, status, configuration, or metadata bag.

## API and compatibility behavior

`GET /api/agents` now returns `runtime_cli_version` on visible list entries:

- A non-empty string means the bound runtime reported a CLI version; the existing minimum-version checks apply.
- An empty string means the upgraded backend checked the runtime but found no valid string version; Quick Create remains blocked.
- An omitted field identifies an older backend or a non-list response; Quick Create falls back to the legacy runtime-list lookup.

The backend uses a single `LEFT JOIN` from `agent.runtime_id` to `agent_runtime.id` and projects only a string-valued `metadata.cli_version`. Missing or malformed metadata becomes an empty string. The join avoids an N+1 query and preserves unbound agents.

For the legacy fallback, loading and failure are separate from the version verdict: loading disables submission without flashing a false missing-version warning, while a failed query blocks both the button and the keyboard-submit path.

## Review-oriented commit split

1. `7423b4a5d` — project runtime CLI compatibility through visible agent-list responses and consume it in Quick Create, with old-backend fallback and permission regression coverage.
2. `38c08edae` — make the legacy fallback fail closed while loading or failed, including button and keyboard-submit guards.

## How was this validated?

- Views: 405 files / 4,796 tests.
- `pnpm --filter @multica/views typecheck`.
- `pnpm typecheck`.
- `pnpm --filter @multica/views lint` (0 errors; pre-existing warnings only).
- `go test ./internal/handler -count=1`.
- `go vet ./internal/handler`.
- `git diff --check`.
- DB-backed regression coverage verifies that a regular member cannot list the private runtime but can receive the projected CLI version for an invocable agent. Local DB execution was unavailable because the shared Docker/containerd store failed with an I/O error; the backend CI job exercises it.

## Related issue

Fixes #7633

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Tests
- [ ] CI / infrastructure

## Checklist

- [x] I have run the affected frontend and backend tests.
- [x] I have added regression coverage for the permission boundary and old-backend fallback states.
- [x] I have preserved agent invocation and runtime resource permissions as separate contracts.
- [x] I have considered installed-client compatibility and fail-closed behavior.

## AI disclosure

**AI tool used:** Codex

**Approach:** Reproduced the member/admin difference through the permission-filtered runtime lookup, traced the independent agent-invocation and runtime-resource gates, projected the minimum compatibility signal onto visible agents, preserved old-backend fallback behavior, and added permission plus loading/error regression coverage.

## Screenshots

Not included. This PR does not add a visual component; it corrects the data source and submission state behind the existing Quick Create version message and Create button.

### DIFF
--- a/packages/core/types/agent.ts
+++ b/packages/core/types/agent.ts
@@ -472,6 +472,13 @@ export interface Agent {
   runtime_id: string;
   /** False exactly when the agent has no runtime. Older backends omit it. */
   runtime_bound?: boolean;
+  /**
+   * CLI version projected from the bound runtime for Quick Create
+   * compatibility checks. This is intentionally separate from runtime-list
+   * access: a member may invoke a shared agent without being allowed to view
+   * its owner's private machine. Older servers and non-list responses omit it.
+   */
+  runtime_cli_version?: string;
   name: string;
   description: string;
   /** What this agent's owner wrote. For a system agent this holds only the
--- a/packages/views/modals/quick-create-issue.test.tsx
+++ b/packages/views/modals/quick-create-issue.test.tsx
@@ -139,6 +139,15 @@ const mockMembersData = vi.hoisted(() => ({
   list: [{ user_id: "user-1", role: "admin" }],
 }));
 
+const mockRuntimesQuery = vi.hoisted(() => ({
+  data: [{ id: "runtime-1", metadata: { cli_version: "1.2.3" } }] as
+    | Array<{ id: string; metadata: Record<string, unknown> }>
+    | undefined,
+  isSuccess: true,
+  isPending: false,
+  isError: false,
+}));
+
 // Per-test override for the squads list so we can flip between "squads
 // exist and one's leader is reachable" and "no squads" cases without
 // re-mocking the whole module.
@@ -164,11 +173,15 @@ vi.mock("@tanstack/react-query", () => ({
       case "agents":
         return { data: mockAgentsData.list };
       case "runtimes":
-        if (!enabled) return { data: undefined, isSuccess: false };
-        return {
-          data: [{ id: "runtime-1", metadata: { cli_version: "1.2.3" } }],
-          isSuccess: true,
-        };
+        if (!enabled) {
+          return {
+            data: undefined,
+            isSuccess: false,
+            isPending: false,
+            isError: false,
+          };
+        }
+        return mockRuntimesQuery;
       case "projects":
         return mockProjectsQuery;
       default:
@@ -274,7 +287,8 @@ vi.mock("@multica/core/runtimes", () => ({
     version
       ? { state: "ok", min: "1.0.0" }
       : { state: "missing", min: "1.0.0" },
-  readRuntimeCliVersion: () => "1.2.3",
+  readRuntimeCliVersion: (metadata?: Record<string, unknown>) =>
+    typeof metadata?.cli_version === "string" ? metadata.cli_version : "",
   MIN_QUICK_CREATE_CLI_VERSION: "1.0.0",
 }));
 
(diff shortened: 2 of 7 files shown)


multica-eve:
Closing in favor of #7655. Both PRs address #7633, but this version adds a runtime CLI-version projection to the agent API and replaces the agent-list query, expanding the backend, API compatibility, and private-runtime information surface without being required for the fix. The existing Quick Create endpoint already performs the authoritative invocation, readiness, and CLI-version checks on the server, so #7655's frontend-only distinction between an unlisted runtime and a runtime with a genuinely missing version is the smaller and safer solution. This branch is also currently conflicting with main. Thank you for the thorough investigation and permission-boundary tests.

PR #7664, opened 2026-08-27
MUL-6771 feat(cli): filter and sort issue list by custom property

### LINKED ISSUE #7663: [Feature]: Filter and sort `multica issue list` by custom property
### Deployment type

self-host

### Area

CLI

### What do you want and why?

We run our backlog (~2,450 issues) entirely through the CLI — our tooling is a set of local coding-agent skills that drive `multica issue list --output json` and friends. No agents or squads on the Multica side, just the tracker as a kanban board, so a capability that only exists in the web UI effectively doesn't exist for us.

We rate every issue on two ordinal axes, impact and effort, currently as label prefixes (`impact:very-low` … `impact:very-high`). Labels are the wrong shape for this: `impact:very-high` sorts alphabetically next to `impact:very-low`, nothing structurally stops an issue carrying two impact labels, and every consumer re-derives the ranking client-side. Select properties with ordered options are exactly the right shape, and we started migrating — then stopped, because the CLI can't filter or sort by a property. The migration would trade away the one thing labels did give us.

The server already implements what we need on `GET /api/issues`: the `properties=` filter (AND across definitions, OR within one, with a `__none__` sentinel for "unset") and `sort=property:<definitionId>`. The web list view uses both. The gap looks like CLI surface only — `validIssueSortColumns` in `cmd_issue.go` rejects `property:` sorts before they ever reach the server, and there's no filter flag at all.

What we want to be able to run:

```bash
multica issue list --property "Impact=High" --property "Impact=Very High" --output json
multica issue list --property "Impact=__none__" --output json   # audit: which issues are unrated
multica issue list --sort property:Impact --direction desc --output json
```

Expose the existing server parameters through `issue list`, resolving names to ids the way `issue property set --name` and its option values already do, so scripts never juggle UUIDs.

One wrinkle I hit while reading the server code: `propertySortExpr` currently sorts a select property by the raw stored value, which is the option-id UUID — an arbitrary order. Nothing user-facing depends on it today (the web UI only offers number/date property sorts), and board grouping already treats option order as canonical, so ordinal sorting needs a small server-side fix there too.

I have a PR ready that does both.

Firer:
## What does this PR do?

Makes custom properties usable from the CLI: `multica issue list` gains a `--property "Name=Value"` filter and `--sort property:<name-or-id>`, backed by the `properties=` / `sort=property:` parameters the server already implements for the web list view.

The motivation is in the linked issue: we run our whole backlog through `issue list --output json`, and we want to move our ordinal impact/effort ratings from label prefixes to select properties — which is blocked while the CLI can't filter or sort by them. The CLI half is thin plumbing over the existing server contract, addressed by name the way `issue property set --name` and its option values already work.

There's one server commit underneath. `propertySortExpr` sorted a select property by the raw stored value, which is the option-id UUID, i.e. an arbitrary order. Nothing sends a select property sort today (the web UI only offers number/date in `sortableProperties`), and board grouping already treats option order as canonical, so the first commit ranks select values by option order instead. That's what makes `--sort property:Impact` mean Low < Medium < High rather than UUID order. Happy to split that commit into its own PR if you'd rather review the shared ORDER BY change separately, since it touches `propertySortExpr`, which ListIssues, ListGroupedIssues, and the table keyset path all consume.

A few calls I made, and why:

- **One `Name=Value` per flag, no comma lists.** Option names may legally contain commas, so repetition is the only unambiguous OR form: repeating a property ORs its values, different properties AND (the server's semantics).
- **Loud failures over silent no-ops**, same reasoning as the existing `--direction` guard. Archived definitions and types the server has no sort order for are rejected up front, and so is any filter value the server could never have stored: a non-numeric or non-finite number, a date outside `YYYY-MM-DD`, a url that is not http(s) with a host, or a text or url value over the store length cap. Sent as-is, those would come back as an empty page that reads like a real answer.
- **`__none__` is documented as reserved.** It's the server's unset sentinel for every type. An option or member literally named `__none__` has to be addressed by UUID.
- The select-sort CASE arms embed each option id's **original config spelling** (validated by `uuid.Parse`, whose grammar admits no quote or backslash) because stored values must equal that spelling exactly, and a canonicalized re-serialization would break explicit non-canonical ids. The `::numeric` cast is load-bearing for the table view's keyset cursor typing. There's a comment pinning both facts.

Deliberately not in this PR: a `--label` filter for `label_ids=` (separable, and label addressing in the CLI is currently UUID/prefix-based, which deserves its own small discussion), the other unexposed ListIssues params, exposing select sorts in the web UI's sort menu, and the server's operator filters (`contains`, `gt`, `gte`, `lt`, `lte`, `before`, `after`). The plan for operators is inline comparison spellings in the same flag (`--property "Effort>=3"`), parsed ahead of the plain `=` so existing `Name=Value` scripts keep working. To keep that path open, a name ending in `<`, `>` or `!` is rejected today.

## Related Issue

Closes #7663

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/property.go` — `propertySortExpr` select branch ranks by option order (`selectPropertySortExpr`); degrade paths (unknown/archived/malformed config) unchanged.
- `server/internal/handler/property_test.go` — DB-backed test with explicit option ids whose lexical order is the reverse of option order (one deliberately uppercase, pinning the original-spelling rule); fails against the old behavior.
- `server/internal/handler/issue_table_query_test.go` — two-page keyset walk over a select-property sort (tied ranks, valueless tail, no duplicated/skipped rows), pinning the `::numeric` cursor typing.
- `server/cmd/multica/cmd_issue.go` — `--property` flag, `property:` sort resolution, updated help and error text.
- `server/cmd/multica/cmd_property.go` — `buildPropertiesFilterQueryParam`, `resolvePropertyFilterValue`, `resolveSortableProperty`; the option resolver lifted out of `encodeIssuePropertyValue` so there's one copy. `resolveActorPropertyRef` now returns the canonical lowercase UUID spelling for explicit `member:<id>` input — stored actor values are normalized on write and the filter matches the stored string exactly, so an uppercase or braced id would have silently missed.
- `server/cmd/multica/cmd_issue_property_filter_test.go` — asserts the exact query params sent (filter shapes, OR merge across name/UUID addressing, sentinel, actor canonicalization, single catalog fetch) and every error guard, including that failing flags never hit `/api/issues`.
- Docs: `issue list` row in `cli.mdx` / `cli.zh.mdx` / `cli.ja.mdx` / `cli.ko.mdx`; the filter list and sort paragraph in `CLI_AND_DAEMON.md`; filtering/sorting section in the `multica-working-on-issues` builtin skill plus its source map.

## How to Test

1. `cd server && go test ./internal/handler -run 'TestListIssuesSelectPropertySortOptionOrder|TestListIssuesPropertyFilterAndSort|TestIssueTableSelectPropertySortKeysetPagination' -count=1` (needs the local Postgres from `make dev`)
2. `cd server && go test ./cmd/multica -count=1`
3. End to end against a dev stack: create a select property (`multica property create --name Impact --type select --option Low --option Medium --option High`), set it on a few issues, then
   `multica issue list --sort property:Impact --direction desc --output json` (option order, unset last),
   `multica issue list --property "Impact=__none__"` (only unrated issues),
   `multica issue list --property Impact=High --property Impact=Medium` (OR within the property).

Risks I thought about: the select sort order change is technically observable through the API, but no shipped client sends one (web `sortableProperties` is number/date only, and the client comparator path can't reach select either); a table-view keyset cursor minted for a select sort before deploy would 400 after (same reach: none today); text properties can legally store the literal string `__none__`, which the filter can't target — that's a server contract property, documented rather than worked around.

## Checklist

- [x] I have run tests locally and they pass (`go test ./internal/handler` and `go test ./cmd/multica`, plus the end-to-end walkthrough above; I did not run the Playwright suite — no frontend code changes)
- [ ] If this change affects the UI, I have included before/after screenshots (no UI changes)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs (N/A)
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` (one table row in `cli.zh.mdx`)

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** I drove this with Claude Code end to end. I gave it our use case (CLI-only tracker workflow, ordinal ratings stuck as labels) and had it verify the server contract against current `main` before writing anything — which is how the select-sort-by-UUID wrinkle surfaced. The implementation was test-first (the select-sort test fails against `main`), and both the plan and the final diff went through a three-way review with Claude, GPT, and Gemini reviewers before I looked at the result; several of their findings shaped the design (no comma lists, the scalar-filter rejection, embedding original option-id spellings). I reviewed and take responsibility for the final diff.


Firer:
Hey @Bohan-J do you have any feedback to help with getting this merged? I'm really interested in the CLI having better feature parity.

Bohan-J (changes requested):
Thanks for this — the product direction is right, and the implementation is careful in the places that matter. Addressing definitions by name so scripts never juggle UUIDs, `StringArray` rather than `StringSlice` so option names containing commas stay unambiguous, one shared catalog fetch, and loud failures instead of silent degradation are all the calls I'd have wanted. The select-sort commit is a genuine fix, not just enabling work: sorting an ordinal scale by option-id UUID was arbitrary, and ranking by option order matches what board grouping already treats as canonical.

One thing to fix before merge.

## Must fix: the "reject what can never match" contract isn't applied to `url` and `text`

`resolvePropertyFilterValue`'s doc comment states the contract:

> A value that could never match is rejected here rather than sent, because an empty result reads like a real answer.

The `number` and `date` branches honour it. Two don't:

- `case "url": return trimmed, nil` — no validation, while `validatePropertyValue` only ever *stores* an http(s) URL with a non-empty host, bounded by `maxPropertyURLValueLen`.
- `case "text": return raw, nil` — unbounded, while the store caps at `maxPropertyTextValueLen` (2000 runes).

Nothing downstream catches either. The server bounds a `contains` operator value at `maxPropertyTextValueLen`, but the plain-string equality path in `parsePropertiesFilterParam` only rejects the empty string — no format check, no length check. So these reach Postgres and come back empty.

Reproduction, dropped into `server/cmd/multica`:

```go
func TestProbeUnstorableFilterValues(t *testing.T) {
	cases := []struct{ name, flag string }{
		{"url missing scheme", "Spec=example.com/foo"},
		{"url garbage", "Spec=not-a-url"},
		{"url scheme the server rejects", "Spec=ftp://example.com/x"},
		{"url no host", "Spec=http://"},
		{"text over the 2000-rune store limit", "Notes=" + strings.Repeat("a", 2500)},
	}
	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			queries, _ := newPropertyFilterTestServer(t)
			if err := listIssuesWithFlags(t, map[string][]string{"property": {tc.flag}}); err != nil {
				t.Logf("REJECTED by CLI (good): %v", err)
				return
			}
			t.Errorf("SENT to /api/issues → silent empty result: properties=%s", (*queries)[0].Get("properties"))
		})
	}
}
```

All five are sent:

```
--- FAIL: TestProbeUnstorableFilterValues/url_missing_scheme
    SENT to /api/issues → silent empty result: properties={"aeaeaeae-…":["example.com/foo"]}
--- FAIL: TestProbeUnstorableFilterValues/url_garbage
    SENT to /api/issues → silent empty result: properties={"aeaeaeae-…":["not-a-url"]}
--- FAIL: TestProbeUnstorableFilterValues/url_scheme_the_server_rejects
    SENT to /api/issues → silent empty result: properties={"aeaeaeae-…":["ftp://example.com/x"]}
--- FAIL: TestProbeUnstorableFilterValues/url_no_host
    SENT to /api/issues → silent empty result: properties={"aeaeaeae-…":["http://"]}
--- FAIL: TestProbeUnstorableFilterValues/text_over_the_2000-rune_store_limit
    SENT to /api/issues → silent empty result: properties={"dddddddd-…":["aaaa…"]}
```

`--property "Spec=example.com/foo"` — dropping the scheme is an easy thing to type — is indistinguishable from "no issue has that spec". That's exactly the failure the comment says the guard exists to prevent, and a silent empty result is the worst shape for it in the scripted use this PR is built for: a human re-reads the command, a script records zero and moves on.

Suggested fix: mirror `validatePropertyValue`'s `url` branch in `resolvePropertyFilterValue` (scheme `http`/`https`, non-empty host, `maxPropertyURLValueLen`) and bound `text` at `maxPropertyTextValueLen`. Then add cases to `TestRunIssueListPropertyFilterErrors` — its existing `len(*queries) != 0` assertion is already the right shape for pinning that a rejected value never reaches `/api/issues`.

## Non-blocking

**The PR description is stale.** It still says value filtering on text/number/date/url errors out, with "(#7240 extends the server side; the CLI guard can be lifted once that lands.)". #7240 has landed and the guard is gone — those types work now. Worth syncing so the description doesn't mislead whoever reads this later.

**Operator filters aren't exposed, and the flag syntax is the part that's hard to change later.** `GET /api/issues` already accepts `{"op","value"}` members — `contains`, `gt`, `gte`, `lt`, `lte`, `before`, `after`. Shipping equality plus `__none__` first is a reasonable scope call and I'm not asking you to widen it here. But `--property "Name=Value"` is the syntax that ships, and `strings.Cut(pair, "=")` splits on the first `=`; adding `--property "Effort>=3"` later means teaching the parser more separators after scripts already depend on the current one. Your own motivating case (ordinal impact/effort) is the one most likely to want `>=` next. A line in the help text reserving the comparison characters, or just confirming that's the intended growth path, would be enough.

## Verified

- I probed the `selectPropertySortExpr` embedding claim with hostile option ids — `' OR 1=1 --`, an arm-breaking `aaaaaaaa-…' THEN 999 WHEN '`, and a trailing backslash. `uuid.Parse` drops all of them before they reach the builder, and the exotic forms it does accept (braces, `urn:uuid:`, uppercase) carry no quote or backslash. The comment is accurate and the `ORDER BY` construction is safe; the `::numeric` note pinning `issueTableOrderBy`'s `strings.Contains` sniff is a good catch to have written down.
- `go test ./cmd/multica -count=1` passes on `fdffc420`. CI is green across all jobs.
- The rebase onto `main` is clean, and the two commits are cleanly separable if you'd still prefer to split the server-side sort change.

Happy to re-review as soon as the value validation is in.

Firer:
Hi @Bohan-J, thanks for the thorough review, and for the probe test. It made the gap very concrete. Pushed the value validation and rebased onto current main.

**Must fix.** `resolvePropertyFilterValue` now mirrors the two store rules it was missing. A `url` value has to parse as http(s) with a non-empty host and stay within 2048 bytes, and `text` is bounded at 2000 runes. The caps live as constants in the CLI next to `propertyNoValueSentinel`, with a comment pointing at `validatePropertyValue`, since `cmd/multica` can't import `internal/handler` without dragging the DB layer into the binary. Your five probe cases are in `TestRunIssueListPropertyFilterErrors` (missing scheme, garbage, ftp, no host, text over the cap) plus a url over the cap, and the existing assertion pins that none of them reach `/api/issues`. There's also a positive case sending exactly 2000 multi-byte runes so the bound is checked in runes, not bytes.

I left `encodeIssuePropertyValue` alone. The set path already gets a 400 from the server for the same inputs, so the silent failure was specific to filtering.

**Operators.** Good question, and you're right that this is the part that's hard to change later. The intended growth path is inline comparison spellings in the same flag, `--property "Effort>=3"`, parsed ahead of the plain `=` so today's `Name=Value` keeps working. Rather than only documenting that, I've reserved it: a name ending in `<`, `>` or `!` is now rejected with "comparison operators are not supported yet", so no script can come to depend on `Impact>` resolving as a property name. The help text, `CLI_AND_DAEMON.md` and the working-on-issues skill say the same. If you'd rather the operator sit in the value (`Effort=gte:3`, which also sidesteps shell redirection on `<` and `>`), now is the cheap time to say so and I'm happy to drop the reservation.

**Description.** Updated. The stale #7240 paragraph is gone and the loud-failures bullet now lists what the guards actually reject.

`go test ./cmd/multica -count=1` is green on the new head. Happy to take another look at anything else you spot.

Bohan-J (approved):
Thanks — this is in good shape, and I appreciate how you handled the review round.

Two things stood out. You fixed the validation gap at the root rather than patching the symptom: mirroring `validatePropertyValue`'s rules into the filter path means the two sides now derive from the same contract, instead of the filter carrying its own drifting copy of "what the server accepts". And the tests pin the shapes that were actually broken, in the place that already asserts a rejected flag never reaches `/api/issues` — so the guard is held by the assertion that matters rather than by a new parallel harness.

The other one: you didn't just reserve `>=`, `<=` and `!=`, you wrote down where the syntax is going and why the reservation keeps that path open. A rejected spelling with no stated plan is a dead end that the next person has to re-litigate; a rejected spelling with the growth path documented in the description is a decision. That's the difference between closing a door and leaving a marked one.

On the select-sort commit — your read that "nothing user-facing depends on it today" is right, but the underlying order is worse than arbitrary in practice, not just in theory. I checked a real workspace with an ordinal `Impact` scale (`XS < S < M < L < XL`) and confirmed `sort=property:<id>` descending currently returns `S, L, XL, M` — exactly the option-id UUID ordering, as you diagnosed. Good instinct to fix it in the same change rather than shipping a CLI flag that surfaces a meaningless order to users for the first time; splitting it out would have meant landing `--sort property:` on top of a known-broken ORDER BY.

I also re-checked the `selectPropertySortExpr` embedding argument, since it's the one hand-built SQL string here. Hostile option ids — `' OR 1=1 --`, an arm-breaking `…' THEN 999 WHEN '`, a trailing backslash — are all dropped by `uuid.Parse` before reaching the builder, and the exotic forms it does accept (braces, `urn:uuid:`, uppercase) carry no quote or backslash. The comment's reasoning holds, and pinning the `::numeric` / `issueTableOrderBy` coupling in a comment is the kind of note that saves someone a confusing afternoon later.

Verified on `ed109195`: every previously-reported input is now rejected before the request, valid inputs including the exact 2000-rune boundary and space-padded text still pass, `go test ./cmd/multica -count=1` is green, and CI is clean.

Approving and merging. Thanks for the thorough follow-up, and for the issue that prompted it.

PR #7669, opened 2026-08-27
fix(scheduler): serialize scheduled autopilot dispatch

jasonbarbee:
## Summary

Scheduled autopilot triggers now enforce single-flight planning. A live RUNNING execution suppresses newer cron occurrences, and a FAILED execution with retry budget remains serialized until its backoff is due. Once retry is due, the original plan time is returned so the existing retry path can reclaim it.

This prevents short-cadence `run_only` autopilots from creating overlapping downstream tasks while preserving stale-lease recovery and terminal-run progression.

## Validation

- `go test ./internal/scheduler`
- `go test ./internal/scheduler ./internal/service`

Both pass from `server/`.

## Regression coverage

`TestAutopilotScheduleSingleFlightWhileRunning` covers RUNNING suppression, FAILED backoff suppression, retry plan-time reuse, and release after SUCCESS.

### DIFF
--- a/server/internal/scheduler/jobs_autopilot.go
+++ b/server/internal/scheduler/jobs_autopilot.go
@@ -224,6 +224,13 @@ func autopilotScopes(
 // missed slot; the lateness guard prevents a paused / newly eligible
 // trigger from firing hours after its configured time.
 //
+// A trigger is also single-flight: while its latest execution lease is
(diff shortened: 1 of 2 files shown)


jasonbarbee:
Closing at the request of the CallTelemetry operator. We are keeping this behavior change in our private deployment/workflow layer rather than proposing an upstream Multica change.

Issue #7653, opened 2026-08-27
[Bug]: JSON with raw \n in description breaks parsing


coreshock:
### Deployment type

self-host

### Area

CLI

### What happened?

Some descriptions contain raw bytes 0x0A inside JSON strings (not escaped \n). json.loads() fails with Invalid control character. --output json should always produce valid JSON.

Works for most issues, breaks on specific ones with multi-line descriptions

### Steps to reproduce

```
multica issue list --output json | python3 -c "import json,sys; json.load(sys.stdin)"
# → json.decoder.JSONDecodeError: Invalid control character
```

### Additional context (optional)

```shell

```

importcpp:
I tried to reproduce this on current `main` and could not — both sides of the path are already safe there, which suggests this is fixed by upgrading. Sharing the verification so you can confirm against your deployment version.

### What I traced

`multica issue list --output json` has two serialization boundaries, and I checked both against `main`:

**1. Server (`/api/issues` → `ListIssues` → `writeJSON`).** `description` is a `text` column, scanned into a Go `string`, placed on `IssueResponse.Description`, and emitted through `writeJSON`, which uses `encoding/json`'s `json.Marshal`. I exhaustively tested every control byte `0x00`–`0x1F` through `json.Marshal`, `Encoder.Encode`, `json.Indent`, and `json.Compact`: **all four escape `0x0A` to `\n`** — none can emit a raw control character inside a JSON string. So a correct server cannot produce the invalid body.

**2. CLI (`GetJSON` decode → `PrintJSON`).** The CLI does **not** stream the response body through to stdout — it decodes the response into a map and re-encodes it with `encoding/json`. I ran the real `runIssueList` against a mock server two ways:
- Server returns properly escaped JSON → CLI prints valid JSON. ✅
- Server returns a body with a **raw** `0x0A` inside the description → the CLI **errors out with empty stdout** (`list issues: invalid character '\n' in string literal`). It never writes the malformed bytes to stdout.

### Why this points at an older build

[...]

PR #7674, opened 2026-08-28
docs(cli): document issue list pagination limit

importcpp:
## What does this PR do?

Documents that `multica issue list --limit` controls one page and that the server caps each page at 100 issues. The previous help text implied that larger values would return that many issues, which could lead callers to mistake a partial page for the complete result set.

Following reviewer feedback, this PR intentionally does not add an unbounded `--all` mode. Callers can use `--offset` for explicit pagination.

## Related Issue

Follow-up to #7653.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/cmd/multica/cmd_issue.go`: clarify the `--limit` help text, document the server's 100-issue page cap, and point users to `--offset` for pagination.

## How to Test

1. Run `go test ./cmd/multica/` from `server/`.
2. Run `go vet ./cmd/multica/` and `go build ./cmd/multica/` from `server/`.
3. Run `multica issue list --help` and confirm that `--limit` describes the 100-issue page cap and `--offset` pagination.

## Checklist

Risk is limited to CLI help output; runtime behavior is unchanged.

## AI Disclosure

**Prompt / approach:** Traced the server-side pagination limit and CLI behavior, initially implemented full pagination, then followed maintainer review to reduce the PR to the documented 100-per-page cap and explicit `--offset` guidance. Verified the final net diff contains only the help-text change and ran the CLI test, vet, and build checks.

## Screenshots (optional)

Not applicable; this change only updates CLI help text.


Bohan-J (changes requested):
Thanks for tracing this — the diagnosis is correct and worth documenting. `ListIssues` does clamp `limit` to 100 (`server/internal/handler/issue.go`), and the flag help never said so, which is a real trap for anyone passing `--limit 1000`.

We'd like to take the help-text half of this PR. We don't want `--all`. Could you reduce this PR to just the `--limit` help change?

## Why not `--all`

The flag is an unbounded paging loop whose cost scales with workspace size, and it is least reliable in exactly the case that motivates it — a large board.

- **Serial request fan-out.** At 30k issues that's 300 sequential requests at 100/page, with no concurrency, no throttling, and no progress output. The command just hangs.
- **Deep offset pagination is quadratic.** The query is `... ORDER BY %s LIMIT %s OFFSET %s`. Reaching page 300 makes Postgres scan and discard the first 29,900 rows; the full walk touches roughly 4.5M rows. One CLI invocation can put visible load on the `issue` table.
- **Whole result set buffered and printed.** The row `SELECT` includes the full `i.description`, and `issuesRaw` accumulates every issue as `map[string]any`. With `--output table` that also floods the terminal with tens of thousands of rows.
- **The snapshot isn't consistent.** Across a walk spanning hundreds of requests, any concurrent create or delete shifts offsets and makes the loop skip or duplicate rows. The default sort is `position` — manual board order — so a single card drag mid-walk reorders the remaining pages. A flag added for result completeness can't guarantee a complete snapshot on the boards where it matters.
- **No bound on the loop.** The only exit is `len(pageIssues) < 100`. If the server ever ignores `offset`, or a filter path keeps returning full pages, this is an infinite loop with unbounded memory growth and nothing to catch it.

There's also a scope point. In #7653 the reporter was already paging correctly with `--offset 0/100/200` on a 146-issue workspace — nobody hit the ceiling unexpectedly. Documenting the cap addresses what was actually observed; adding a new fetch-everything mode goes well past it.

## What we'd merge

Just the flag help, minus the `--all` reference. Something like:

```go
issueListCmd.Flags().Int("limit", 50, "Maximum number of issues to return in one page (the server caps a page at 100; use --offset to page through more)")
```

That means dropping the `--all` flag registration, the paging branch in `runIssueList`, and the three new tests. Happy to look again once it's trimmed.

If a bulk export mode is something you want to pursue separately, it would need a hard page bound, an error rather than silent truncation when that bound is hit, streaming output instead of full buffering, and keyset pagination (`created_at, id`) instead of `OFFSET`. That's a design discussion worth its own issue rather than a flag on this command.

importcpp:
Thanks for the detailed review — agreed. I’ve removed the `--all` flag, the paging branch, and all three related tests. The PR now contains only the requested `--limit` help-text change pointing users to `--offset` (net diff: 1 addition, 1 deletion). I also updated the title and description and reran `go test ./cmd/multica/`, `go vet ./cmd/multica/`, and `go build ./cmd/multica/`; all pass. Could you please take another look?

Bohan-J (approved):
Thanks for trimming this down — the scope is right now, and the wording is accurate.

I verified the claim against the server: `ListIssues` defaults to `limit := 100` and clamps with `if limit > 100 { limit = 100 }`. There is one server path that bypasses that cap (`open_only=true` returns every non-done/cancelled issue with no limit, returning before the clamp), but `runIssueList` never sends `open_only` — so "the server caps a page at 100" holds for every `multica issue list` invocation, with no exception a reader could trip over.

Also confirmed the branch is clean of the previous iteration: no `--all` flag, no paging branch, and `cmd_issue_test.go` is back to its base state.

LGTM.

PR #7683, opened 2026-08-28
MUL-6783: perf(server): set read-header and idle timeouts on main HTTP server

### LINKED ISSUE #7682: [Feature]: Set timeouts on the main HTTP server (Slowloris hardening)
### Deployment type

self-host

### Area

Self-hosting and deployment

### What do you want and why?

The main HTTP server in `server/cmd/server/main.go` is currently constructed with only `Addr` and `Handler`, leaving all timeout fields at their zero value (no timeout).

Without `ReadHeaderTimeout`, a slow or half-open client can keep request headers incomplete indefinitely, retaining a goroutine and read buffer and leaving the server exposed to Slowloris-style resource exhaustion. Without `IdleTimeout`, idle keep-alive connections are not reclaimed.

The metrics server (`server/internal/metrics/server.go`) and profiling server (`server/internal/profiling/server.go`) already configure HTTP server timeouts. The main public server is the only server missing them.

Configure the main `http.Server` with:

- `ReadHeaderTimeout: 5 * time.Second` to bound request-header reads.
- `IdleTimeout: 120 * time.Second` to reclaim idle keep-alive connections while allowing normal browser, CLI, and daemon connection reuse.

Leave `ReadTimeout` and `WriteTimeout` unset because WebSocket endpoints (`/ws` and `/api/daemon/ws`) share this listener and require long-lived connections.

_No response_

importcpp:
## What does this PR do?

The main HTTP server (`server/cmd/server/main.go`) was constructed with only `Addr` and `Handler`, so every timeout defaulted to zero — i.e. **never time out**. Two consequences:

- **Slowloris exposure.** With no `ReadHeaderTimeout`, a slow or half-open client that dribbles request headers one byte at a time pins a goroutine and its read buffer indefinitely. A few thousand such connections can exhaust goroutines/connections at near-zero cost to the attacker. `net/http` gives no header-read deadline by default, and static scanners (gosec) flag exactly this.
- **No idle reclaim.** With no `IdleTimeout`, keep-alive connections are never reclaimed once idle.

The sibling **metrics** (`internal/metrics/server.go`) and **profiling** (`internal/profiling/server.go`) servers already set these timeouts, so the public server was the odd one out — this looks like an oversight rather than intent.

This sets:
- `ReadHeaderTimeout: 5s` — matches the metrics/profiling servers; bounds header reads and closes Slowloris.
- `IdleTimeout: 120s` — deliberately looser than their 30s, because this listener serves browsers, the CLI, and high-frequency daemon polling that heavily reuse keep-alive connections; too short a value would churn TCP/TLS handshakes.

`ReadTimeout` and `WriteTimeout` are **intentionally left unset**: WebSocket upgrades (`/ws`, `/api/daemon/ws`) live on this same listener, and a global read/write deadline would kill those long-lived connections mid-stream. This mirrors the profiling server, which omits the same two timeouts so long captures aren't truncated.

`MaxHeaderBytes` is left at Go's 1MB default (setting it explicitly would only be noise).

## Related Issue

Closes #7682

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

> This intentionally changes the default request-timeout behavior for **every deployment** (unbounded → bounded header reads + idle reclaim), so it is classified as a bug fix / security hardening rather than a no-behavior-change refactor.

## Changes Made

- `server/cmd/server/main.go`: extract the public HTTP server construction into `newMainHTTPServer`, which sets `ReadHeaderTimeout: 5s` and `IdleTimeout: 120s`, with a comment explaining why `ReadTimeout`/`WriteTimeout` are deliberately left zero (WebSocket upgrades on the same listener).
- `server/cmd/server/main_test.go`: add `TestNewMainHTTPServerTimeouts`, a structure-level regression test asserting `ReadHeaderTimeout == 5s`, `IdleTimeout == 120s`, and `ReadTimeout == WriteTimeout == 0`, so the safety defaults can't be silently removed, zeroed, or made non-zero without failing CI.

## How to Test

1. `cd server && go build ./cmd/server/ && go vet ./cmd/server/` → clean.
2. `cd server && go test ./cmd/server/ -run TestNewMainHTTPServerTimeouts` → passes; fails if any of the four timeout invariants is broken.
3. **WebSocket unaffected:** start the server, open a `/ws` (and daemon `/ws`) connection and keep it idle well past 120s → it stays connected (no global write/read deadline).
4. **Slowloris defense active:** open a raw TCP connection and send partial request headers slowly without completing them → the server closes it after ~5s (previously it would hang indefinitely). `slowhttptest -c 200 -H -u http://localhost:8080/` shows connections being dropped instead of held open.

## Checklist

- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs**
- [ ] If this PR touches Chinese product copy, I checked it against the conventions doc

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** As part of a broader server performance/hardening audit, identified that the main `http.Server` set no timeouts while the sibling metrics and profiling servers did. Verified the WebSocket constraint by tracing `/ws` and `/api/daemon/ws` on the same listener, and aligned the values with the existing servers (`ReadHeaderTimeout: 5s`) while keeping `IdleTimeout` looser (120s) for keep-alive-heavy CLI/daemon clients. After review, extracted the construction into a helper and added a structure-level regression test pinning all four timeout invariants. Confirmed with `go build`, `go vet`, and `go test`.

### DIFF
--- a/server/cmd/server/main.go
+++ b/server/cmd/server/main.go
@@ -566,8 +566,14 @@ func main() {
 	})
 
 	srv := &http.Server{
-		Addr:    ":" + port,
-		Handler: r,
+		Addr:              ":" + port,
+		Handler:           r,
+		ReadHeaderTimeout: 5 * time.Second,
+		IdleTimeout:       120 * time.Second,
+		// Bound header reads to stop Slowloris; IdleTimeout is looser than the
+		// metrics/profiling servers' 30s for keep-alive-heavy CLI and daemon
+		// clients. ReadTimeout and WriteTimeout are unset so WebSocket upgrades
+		// (/ws, /api/daemon/ws) on this listener aren't killed mid-connection.
 	}
 	profilingServer := profiling.NewServer()
 
--- a/server/cmd/server/main.go
+++ b/server/cmd/server/main.go
@@ -269,6 +269,23 @@ func jwtSecretBootError(jwtSecret, appEnv string) error {
 	return auth.ValidateJWTSecret(jwtSecret)
 }
 
+// newMainHTTPServer builds the public HTTP server with the production timeout
+// defaults. These values are load-bearing safety settings, not cosmetic tuning,
+// so they live in one helper that main() and the config regression test share.
+//
+// Bound header reads to stop Slowloris; IdleTimeout is looser than the
+// metrics/profiling servers' 30s for keep-alive-heavy CLI and daemon clients.
+// ReadTimeout and WriteTimeout are deliberately left zero so WebSocket upgrades
+// (/ws, /api/daemon/ws) on this listener aren't killed mid-connection.
+func newMainHTTPServer(addr string, handler http.Handler) *http.Server {
+	return &http.Server{
+		Addr:              addr,
+		Handler:           handler,
+		ReadHeaderTimeout: 5 * time.Second,
+		IdleTimeout:       120 * time.Second,
+	}
+}
+
 func main() {
 	logger.Init()
 	// Warn about missing configuration
@@ -565,16 +582,7 @@ func main() {
 		LLMMaxRetries:       llmMaxRetries,
 	})
 
-	srv := &http.Server{
-		Addr:              ":" + port,
-		Handler:           r,
-		ReadHeaderTimeout: 5 * time.Second,
-		IdleTimeout:       120 * time.Second,
-		// Bound header reads to stop Slowloris; IdleTimeout is looser than the
-		// metrics/profiling servers' 30s for keep-alive-heavy CLI and daemon
-		// clients. ReadTimeout and WriteTimeout are unset so WebSocket upgrades
-		// (/ws, /api/daemon/ws) on this listener aren't killed mid-connection.
-	}
+	srv := newMainHTTPServer(":"+port, r)
 	profilingServer := profiling.NewServer()
 
 	// Start background workers.
--- a/server/cmd/server/main_test.go
+++ b/server/cmd/server/main_test.go
@@ -513,3 +513,28 @@ func TestJWTSecretBootError(t *testing.T) {
 		})
 	}
 }
+
+// TestNewMainHTTPServerTimeouts pins the production timeout defaults on the
+// public HTTP server. These are safety settings, not tuning: removing them,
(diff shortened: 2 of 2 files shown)


multica-eve (changes requested):
Self-hosted workspaces should be protected from clients that consume service capacity by never completing a request, while normal connection reuse and long-lived realtime sessions must keep working. This direction is correct, but the safety behavior needs a regression guard before merge.

## Request changes

**Blocking — add regression coverage for the production safety defaults (`server/cmd/server/main.go:568`).** The two timeout values currently exist only as literals inside `main()`, and no test distinguishes the protected configuration from the previous unbounded one. The current test suite and CI would still pass if these fields were removed, reset to zero, or if `ReadTimeout` / `WriteTimeout` were accidentally made non-zero and began disrupting uploads or WebSockets.

Please extract the main HTTP server construction into a small helper and add a unit test in `server/cmd/server/main_test.go` asserting:

- `ReadHeaderTimeout == 5*time.Second`
- `IdleTimeout == 120*time.Second`
- `ReadTimeout == 0`
- `WriteTimeout == 0`

A structure-level configuration test is sufficient; there is no need for a network test that sleeps for five seconds.

**Nit:** Please also classify this as a bug fix / hardening change in the PR description and uncheck “Refactor / code improvement (no behavior change),” because this intentionally changes the default behavior for every deployment.

importcpp:
Thanks! Addressed both items in `baf72ae69`:

- Extracted the main HTTP server construction into `newMainHTTPServer(addr, handler)`.
- Added `TestNewMainHTTPServerTimeouts` in `server/cmd/server/main_test.go`, asserting `ReadHeaderTimeout == 5*time.Second`, `IdleTimeout == 120*time.Second`, and `ReadTimeout == WriteTimeout == 0` as a structure-level regression test.
- Updated the PR description to classify this as a bug fix / hardening change and unchecked `Refactor / code improvement (no behavior change)`.

Validation is still green locally: `gofmt`, `go build ./cmd/server/`, `go vet ./cmd/server/`, and `go test ./cmd/server/`.

PR #7684, opened 2026-08-28
MUL-6784: fix(desktop): share public web URLs

YikaJ:
## Summary

- make project Copy link use the platform navigation adapter instead of the Electron renderer location
- render public web URLs in Desktop AppLink anchors so native copy-link and drag affordances never expose file:// renderer paths
- attach the shareable web page URL to feedback submitted from Desktop
- keep Web anchors route-relative to preserve SSR hydration and native browser navigation

## Root cause

Shared views assumed window.location represented the public application URL. In packaged Electron builds it points at the local app.asar renderer index.html, so direct clipboard and diagnostic uses produced a machine-local file URL. AppLink anchors also kept route-relative hrefs, which Chromium resolved against that file renderer for native link affordances.

The existing NavigationAdapter.getShareableUrl boundary already provides the connected environment Web URL on Desktop and the browser origin on Web. This change routes the remaining current-page uses through that boundary.

## Verification

- pnpm --filter @multica/views typecheck
- pnpm --filter @multica/desktop typecheck
- pnpm --filter @multica/views test (405 files, 4,797 tests)
- pnpm --filter @multica/desktop test (54 files, 518 tests)
- pnpm --filter @multica/desktop build
- targeted ESLint on all six changed files (no errors; one pre-existing project-detail hooks warning)

### DIFF
--- a/apps/desktop/src/renderer/src/platform/issue-window-navigation.test.tsx
+++ b/apps/desktop/src/renderer/src/platform/issue-window-navigation.test.tsx
@@ -10,6 +10,11 @@
 import { describe, expect, it, vi, beforeEach } from "vitest";
 import { act, render, screen } from "@testing-library/react";
 import { MemoryRouter, Route, Routes, useLocation } from "react-router-dom";
+import {
+  currentPath,
+  useNavigation,
+  type NavigationAdapter,
+} from "@multica/views/navigation";
 
 const APP_URL = "https://app.example";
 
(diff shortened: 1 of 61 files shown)


Bohan-J:
Thanks for this, @YikaJ — routing the remaining current-page uses through `NavigationAdapter.getShareableUrl` instead of adding a parallel mechanism is exactly the right shape, and the `AppLink` split between the displayed href and the in-app path it actually routes with is neat.

One blocker came out of review, so I pushed a follow-up commit rather than sending it back to you: rebuilding the current page from `pathname` + `searchParams` alone dropped the fragment. Feedback submitted from a `#comment-…` deep link then reported the whole issue, losing the position the reporter was at — and on web that was a regression, since `window.location.href` used to carry the fragment.

The fragment can't be recovered from `window.location` on desktop (the renderer is a MemoryRouter over the packaged `file://` page), so it had to come from the platform boundary. `NavigationAdapter` now reports `hash` next to `pathname` / `searchParams`, all three adapters (web, desktop shell, issue window) fill it in, and a small `currentPath()` helper composes them so no caller has to remember the fragment again. Regression tests cover web (`hashchange` after mount included), the desktop tab URL, and the issue window's MemoryRouter.

Verified locally: `@multica/views` 406 files / 4801 tests, `@multica/desktop` 54 / 524, `@multica/web` 30 / 250, plus typecheck on all three and ESLint on the changed files (only the pre-existing `project-detail` hooks warning).

Merging — thanks again for the fix.

Bohan-J (approved):
Approving: product direction is right, the implementation reuses the existing navigation boundary, and the fragment blocker is fixed in the follow-up commit. Local suites and typechecks pass.

PR #7689, opened 2026-08-28
MUL-6792: perf(server): batch skill-file loads in LoadAgentSkills

### LINKED ISSUE #7688: [Feature]: Batch skill file loading in LoadAgentSkills
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

`TaskService.LoadAgentSkills` first lists an agent's skills, then calls `ListSkillFiles` once per skill. On the task-claim hot path, an agent with N skills therefore requires N+1 database round trips. For example, 15 skills produce 16 queries.

Each `ListSkillFiles` query also selects complete file rows, including file bodies. The total payload is required by the claim response, but paying pool-acquire, network round-trip, and row-processing overhead separately for every skill adds avoidable latency and database load on a high-frequency path.

The affected method is called from daemon task-claim flows, where the surrounding code documents peak activity of roughly 3,800 acquires per second. The existing `idx_skill_file_skill` index already supports a batched lookup.

Add a sqlc query that loads files for all requested skill IDs in one round trip:

```sql
SELECT * FROM skill_file
WHERE skill_id = ANY(sqlc.arg('skill_ids')::uuid[])
ORDER BY skill_id, path ASC;
```

Group the returned rows by `skill_id` in Go, then assemble results in the original skill order. This reduces `LoadAgentSkills` from N+1 database queries to 2 while preserving output ordering and payload. Keep the existing single-skill query for handlers that legitimately load one skill.

_No response_

importcpp:
## What does this PR do?

`TaskService.LoadAgentSkills` previously listed an agent's skills and then called `ListSkillFiles` once for every skill. On the task-claim hot path, N skills therefore required N+1 database round trips; 15 skills meant 16 queries. Each query also returned complete file rows, so the fixed pool-acquire and round-trip overhead was paid repeatedly for a payload the caller needs as one collection.

This PR adds one batched sqlc query for all skill IDs, groups the returned rows by `skill_id`, and assembles the response in the original skill order. `LoadAgentSkills` now uses 2 database queries regardless of skill count. The total payload is unchanged, each skill's files remain ordered by path, and existing single-skill callers continue using `ListSkillFiles`.

## Related Issue

Closes #7688

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/db/queries/skill.sql`: add `ListSkillFilesBySkillIDs`, backed by the existing `idx_skill_file_skill` index and ordered by `skill_id, path`.
- `server/pkg/db/generated/skill.sql.go`: regenerate sqlc output for the new batch query.
- `server/internal/service/task.go`: replace the per-skill query loop in `LoadAgentSkills` with one batch load and an in-memory grouping pass, while preserving the original skill ordering.

## How to Test

2. Run `go vet ./internal/service/... ./internal/handler/...`.
3. Run the existing service package tests.
4. Confirm sqlc generation is clean and `LoadAgentSkills` returns each skill's files in path order.

## Checklist

Risk is limited to grouping batch-query results by skill ID. The query explicitly orders by `skill_id, path`, the response is assembled in the original skill order, skills without files retain an empty/nil file list as before, and the existing single-skill query remains available to its current callers.

## AI Disclosure

**Prompt / approach:** Audited the task-claim path for repeated database access, identified the per-skill `ListSkillFiles` loop, added a sqlc batch query using the repository's existing UUID-array pattern, regenerated sqlc output, and grouped results in Go while preserving ordering. OpenAI Codex reviewed the final diff, rebased it onto the latest upstream `main`, and prepared the issue/PR metadata.

## Screenshots (optional)

Not applicable; this is a backend database-query optimization with no UI change.

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -1837,6 +1837,28 @@ func (h *Handler) rejectClaimSourceLoad(ctx context.Context, task *db.AgentTaskQ
 	}
 }
 
+// rejectClaimSkillLoad settles a claim whose agent skills could not be read.
+// It mirrors rejectClaimSourceLoad's transient branch — preserve the
+// just-dispatched task so the stale-dispatched reclaim redelivers it — because
+// this is the same kind of failure: a transient read on the claim-build path.
+//
+// Dispatching with whatever loaded is not the safer half-measure it looks
+// like. LoadAgentSkills reads the whole skill set in two queries, so a failure
+// means every skill loses its supporting files or the agent loses every skill,
+// and the ref hashes sent to the daemon are computed over that same truncated
+// content — so the daemon validates the bundle, caches it, and the agent runs
+// with rules silently missing. There is no ErrNoRows branch to mirror: an
+// agent with no skills is a legitimate empty result, not a missing row.
+func (h *Handler) rejectClaimSkillLoad(task *db.AgentTaskQueue, err error) *claimBuildFailure {
+	slog.Error("task claim: agent skill load failed; preserving task for redelivery",
+		"task_id", uuidToString(task.ID), "agent_id", uuidToString(task.AgentID), "error", err)
+	return &claimBuildFailure{
+		outcome: "error_skill_load",
+		status:  http.StatusInternalServerError,
+		message: "failed to load agent skills",
+	}
+}
+
 // rejectClaimOnWorkspaceMismatch enforces the claim's tenant boundary against
 // the workspace that OWNS the task's context (issue / chat session / autopilot
 // / quick-create), which is the only authority for MULTICA_WORKSPACE_ID in the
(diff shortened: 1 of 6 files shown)


Bohan-J:
Thanks for this, @importcpp — clean, well-scoped change with an accurate problem statement, and the batch query you added is exactly the one the next step needs.

One follow-up commit was pushed to your branch (`644f1e6d9`): batching made a single query own *every* skill's files, so the `files, _ :=` discard changed blast radius — one failed read now strips the supporting files from all of an agent's skills at once. That can't be caught downstream, because `BuildAgentSkillBundles` hashes whatever loaded, so the refs sent at claim and the bundle served at resolve agree with each other and the daemon's `validateSkillBundle` accepts and caches the truncated set. `LoadAgentSkills` / `LoadAgentSkillBundles` now return the error, and the three call sites settle it: claim refuses with 500 and leaves the task dispatched for the stale-dispatched reclaim to redeliver, and `ResolveTaskSkillBundles` returns 500, which the daemon's existing resolve retry already treats as transient. Tests cover both read failures plus the batch grouping itself.

Merging now. A follow-up PR will take the next step in the same area: `ResolveTaskSkillBundles` currently calls `LoadAgentSkillBundles` to load and hash *all* of an agent's skills on every resolve request just to return the one that was asked for — with one request per skill, that is O(N²) content reads and hashes per cold dispatch. It will reuse the `ListSkillFilesBySkillIDs` query you added here.

PR #7696, opened 2026-08-28
MUL-6788: perf(server): batch claim-path attachment and runtime lookups

### LINKED ISSUE #7694: [Feature]: Batch remaining task-claim attachment and runtime lookups
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

Two daemon paths still issue one database query per item even though batch queries already exist:

1. `buildClaimedTaskResponse` calls `ListAttachmentsByChatMessage` once for every unanswered chat message while building a claimed task response. A claim containing N unanswered messages therefore performs N attachment queries.
2. `DaemonDeregister` calls `GetAgentRuntime` once for every runtime ID. Deregistering N runtimes therefore performs N point lookups before applying the existing ownership checks and offline updates.

The first path is part of task claiming, where avoidable database round trips directly add dispatch latency. The second adds unnecessary lookup load during daemon shutdown or disconnect handling.

The repository already provides `ListAttachmentsByChatMessageIDs` and `GetAgentRuntimes`, so both N+1 patterns can be removed without adding schema or generated-query changes.

- Collect unanswered chat-message IDs, call `ListAttachmentsByChatMessageIDs` once, group results by `chat_message_id`, and assemble attachments in the original unanswered-message order. The batch query's `created_at ASC` ordering preserves per-message attachment order.
- Parse all runtime IDs, call `GetAgentRuntimes` once, and build a map keyed by canonical UUID. Iterate the original request IDs so logging, offline-reason lookup, ownership verification, state updates, metrics, and missing-runtime handling retain their existing per-runtime semantics.
- Keep the existing single-item queries for callers that legitimately load one message or runtime.

_No response_

importcpp:
## What does this PR do?

Removes two remaining N+1 database patterns from daemon task-claim and deregistration flows by using batch queries that already exist in the repository.

In `buildClaimedTaskResponse`, attachment loading previously called `ListAttachmentsByChatMessage` once per unanswered message. The implementation now collects all unanswered message IDs, loads their attachments once with `ListAttachmentsByChatMessageIDs`, groups by `chat_message_id`, and assembles output in the original message order. Both the single and batch queries order by `created_at ASC`, so per-message attachment ordering is preserved.

In `DaemonDeregister`, runtime validation previously called `GetAgentRuntime` once per requested runtime ID. The implementation now loads all rows once with `GetAgentRuntimes` and builds a canonical UUID-keyed map. It still iterates the original request IDs, preserving offline-reason lookup, logging, ownership verification, state updates, metrics, affected-workspace notifications, and the existing skip behavior for missing runtimes.

The existing single-item queries remain available to their other callers. The lower-value coalesced-comment N+1 identified in the same audit is intentionally left for a separate design discussion because it requires new sqlc queries.

## Related Issue

Closes #7694

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/daemon.go`: batch unanswered-message attachment reads with `ListAttachmentsByChatMessageIDs` and group rows by chat message.
- `server/internal/handler/daemon.go`: batch daemon-deregister runtime reads with `GetAgentRuntimes` and map rows by canonical runtime UUID.
- Preserve original request/message iteration order and all per-item authorization, update, metrics, logging, and missing-row behavior.

## How to Test

2. Run `go vet ./...`.
3. Run the handler tests covering daemon deregistration, attachment-backed task claims, and coalesced claims.
4. Verify deregistration with a mix of existing, missing, and differently-cased runtime UUID strings; existing rows are processed, missing rows are skipped, and offline reasons remain keyed by the original request string.

## Checklist

Risk is limited to grouping batch-query results. Attachment output is rebuilt in the original unanswered-message order, and both queries order attachments by `created_at ASC`. Runtime lookup uses the parsed UUID's canonical form for map keys while retaining the original request string for offline-reason lookup and logs. Batch-query failures fail safely by processing no rows, matching the previous per-item error paths' skip behavior.

## AI Disclosure

**Prompt / approach:** Audited the daemon claim path for repeated database access, reused the existing attachment and runtime batch queries, and grouped results in memory while preserving original iteration order and per-item behavior. Explicitly checked attachment SQL ordering and canonicalized runtime UUID map keys so differently-cased request IDs still resolve. OpenAI Codex reviewed the final diff and prepared the issue/PR workflow.

## Screenshots (optional)

Not applicable; this is a backend database-query optimization with no UI change.


Bohan-J (changes requested):
Thanks for this — the direction is right and I confirmed both N+1 paths are real. The success-path rewrite is also good: keying the deregister map by the canonical UUID (so a differently-cased id still matches) and grouping the globally `created_at ASC` attachment rows by `chat_message_id` while iterating `unanswered` in its original order does reproduce the previous append order exactly.

What blocks merge is the error contract. Batching changes the blast radius of a single failed query, and both new batch reads currently swallow their error and continue as if the rows simply didn't exist.

### 1. `DaemonDeregister`: a failed batch lookup reports success while nothing goes offline

`server/internal/handler/daemon.go:948-954` logs and leaves `runtimesByID` empty when `GetAgentRuntimes` fails. Every id then hits the `!ok` "runtime not found" skip at `:960-963`, and the handler still returns `200 {"status":"ok"}` at `:995`.

Before this change, a transient error skipped one runtime and the rest of the loop could still succeed. Now one hiccup silently no-ops the whole deregister, the daemon logs a successful shutdown, and those runtimes stay online until the liveness sweep reaps them — long enough to keep attracting claims.

This endpoint's own batch sibling already has the right contract: `ListRuntimesForClaim` (`:1655-1658`) returns `500 "failed to load runtimes"` when the same query fails, and only *individually* unknown/unauthorized ids are skipped. Please do the same here — a read error is not "the row does not exist".

### 2. Claim path: a failed attachment batch starts the task with zero attachments

`server/internal/handler/daemon.go:2777-2785` ignores `attErr` from `ListAttachmentsByChatMessageIDs`, and the claim response is then returned as a success. Attachment IDs are the agent's only handle for downloading user files, so the agent gets the text, is never told files exist, and the task is not redelivered.

The per-message version had the same `attErr == nil` shape, so this isn't a new pattern — but it used to cost one message's attachments, and now it costs the entire turn's. Given that widened radius I'd align it with the input contract a few lines above in the same function: the chat-input load failure returns a `claimBuildFailure{outcome: "error_chat_input_load", status: 500}` precisely so an incomplete input is preserved for redelivery instead of dispatched. An attachment read failure should fail closed the same way.

### 3. Rebase onto `main`, and keep the runtime-lookup metric

The PR is `CONFLICTING` against `main`. The conflict is MUL-6884: `main` now reads the row through `h.getAgentRuntime(r.Context(), obsmetrics.RuntimeLookupSourceDaemonAPI, runtimeUUIDs[i])` rather than `h.Queries.GetAgentRuntime`, so this hunk needs a manual resolution.

Please resolve it by keeping the batch read *and* the observability, not by dropping either — `service.RuntimeLookup.Get` is what feeds `multica_agent_runtime_lookup_total`, and a raw `h.Queries.GetAgentRuntimes` call makes deregister lookups disappear from that metric. Recording one `ok`/`not_found` result per requested id from the batch result keeps the counter meaningful without reintroducing N queries.

### 4. Tests

No tests are added, and the only handler-side deregister coverage today is `TestDaemonDeregisterRejectsMalformedRuntimeID` — it can't demonstrate the equivalence this PR claims. Since the whole PR is a behaviour-preserving refactor, the tests are what make it reviewable. Please cover:

- multiple unanswered messages with interleaved attachment `created_at` values → output stays in message order, then per-message attachment order;
- attachment batch read error → claim fails and the task remains eligible for redelivery;
- deregister with a mix of existing / missing / different-cased runtime ids;
- runtime batch read error → not a `200 ok`.

### Summary

Requesting changes: fix the two swallowed batch errors, rebase onto `main` keeping the MUL-6884 lookup metric, and add the regression tests above. The optimization itself (N+1 → 1 read on both paths, writes still per-item on deregister) is one we want, so once those are addressed this should be straightforward to land.

importcpp:
Pushed the review fixes to this PR. The branch was rebuilt on `main` at `11bd18a50`, with head `fcc068be0`; GitHub now reports the branch as merge-clean.

Addressed the requested changes:

- `DaemonDeregister` now fails closed with `500 failed to load runtimes` when the batch runtime lookup fails, before any per-runtime writes occur.
- Attachment batch-load failures now return a `500` claim build failure instead of starting a task with incomplete input; the task remains `dispatched` with `started_at IS NULL`, preserving stale-dispatched redelivery.
- Runtime batching now goes through `RuntimeLookup.GetMany`, preserving one logical `ok` / `not_found` / `error` metric result per requested runtime ID without restoring N point queries.
- Added regression coverage for attachment ordering, attachment-load failure, mixed existing/missing/differently-cased runtime IDs, runtime batch-load failure, and successful/error metric classification (including duplicate IDs).
- New handler tests use the shared `dbfx` fixtures and `testutil.Call` harness.
- Clarified the metric documentation: point and batch reads have separate SQL fingerprints, while `multica_agent_runtime_lookup_total` provides product-source attribution across both and is not SQL query QPS.

This supersedes the original PR-body risk note that treated failed batch reads as empty results: both batch-read failure paths now fail closed.

Validation completed before push:

- the four targeted DB-backed handler regressions passed;
- `go test ./internal/service -run '^TestRuntimeLookup' -count=1`;
- `go test ./internal/metrics -count=1`;
- `go build ./...`;
- `go vet ./internal/handler ./internal/service ./internal/metrics`;
- `gofmt` and `git diff --check` are clean.

Ready for re-review.

Bohan-J (changes requested):
Re-reviewed at `7d2cac95c`. The three items from my last pass are genuinely fixed, and the way you fixed them is better than what I asked for:

- `DaemonDeregister` now returns `500 "failed to load runtimes"` before any write when the batch read fails (`server/internal/handler/daemon.go:961-967`).
- The attachment batch failure now fails closed and leaves the task dispatched for redelivery (`:2936-2947`), matching the chat-input contract in the same function.
- The rebase kept the MUL-6884 attribution rather than dropping it, and `RuntimeLookup.GetMany` (`server/internal/service/agent_runtime_lookup.go:49-80`) mirrors N individual `Get` calls exactly — one `error` per requested id on failure, `ok`/`not_found` per id on success. The new tests cover attachment ordering, both fail-closed paths, mixed existing/missing/case-variant ids, and the GetMany classification itself. That's the coverage I wanted.

One thing still blocks merge.

### Must-fix: the batch-claim path bypasses `RuntimeLookup`, and it's the hot one

`server/internal/service/agent_runtime_lookup.go:15-16` now states that `RuntimeLookup` "is the only way production code reads agent_runtime rows by id — one row via `Get`, or many in one query via `GetMany`", and `server/internal/metrics/business.go:287` documents the counter as "one per requested id, not per SQL query". Two call sites read the rows without going through it, so neither is counted at all:

- `server/internal/handler/daemon.go:1753`, inside `ClaimTasksByRuntime`. This is not a side path — `server/cmd/server/router.go:1457-1458` routes both `/tasks/claim` and `/claim` to it, and `server/internal/handler/daemon_rpc.go:84` replays the WebSocket claim RPC through the same handler. So after this PR, deregister (once per daemon shutdown) is attributed, while the claim path (every daemon, continuously, over both transports) contributes zero. Comparing `source` on `multica_agent_runtime_lookup_total` would make the hottest reader look idle — the opposite of what MUL-6884 built the label for, and exactly the kind of thing that misdirects the next optimization.
- `server/internal/handler/agent.go:1004`, the agent-list availability projection for non-owner/admin viewers. This one predates your PR — I'm raising it because your PR is what promotes "the only way" to a stated contract, so it now has to be true or be worded differently.

The wiring is already in place: `h.getAgentRuntimes(ctx, source, ids)` exists at `server/internal/handler/runtime.go:648`, and `ClaimTasksByRuntime` already de-dups into `ids` and filters authorization after the read, so switching it to `GetMany` and iterating the returned map is mechanical. `RuntimeLookupSourceDaemonAPI` is the right label there — its doc comment already names "task calls". For `agent.go`, classify it rather than letting it land on `other` (which `server/internal/metrics/labels.go:230-231` says should sit at ~0); `runtime_api` fits, or add a dedicated source if you'd rather see it separately, remembering `AllRuntimeLookupSources()` prewarms the grid. If you'd instead prefer to leave `agent.go` alone as out of scope, then narrow the wording at `agent_runtime_lookup.go:15` so the comment doesn't claim more than the code does — but please don't leave the claim-path one, that one is in scope by construction.

To be explicit about a stale suggestion of mine: **do not restore the AST guard.** `main`'s `603519d15` (MUL-6977) deliberately removed that source-scanning test, and re-adding it would fight a decision the repo just made. A handler-level behavior test that claims through the batch endpoint and asserts `daemon_api` `ok`/`not_found` counts is the right shape here — same style as the `GetMany` tests you already wrote, just one level up.

### Summary

Requesting changes on the metric gap only. The batching design, failure semantics, and test coverage are all where I want them — route the claim-path read (and ideally the agent-availability read) through `RuntimeLookup.GetMany` with a handler-level assertion on the counter, and this is ready to land.

importcpp:
Pushed the metric-gap fix. Branch rebuilt on `main` at `d8fa885d2`, head `8109a5ac6`; GitHub reports it merge-clean.

**Claim path now goes through `RuntimeLookup`.** `ClaimTasksByRuntime` reads via `h.getAgentRuntimes(ctx, RuntimeLookupSourceDaemonAPI, ids)`, so every requested id is billed one `ok` / `not_found` on `multica_agent_runtime_lookup_total` and the hottest reader stops looking idle next to once-per-shutdown deregisters.

Two things in that rewrite are load-bearing and worth flagging, since neither is visible from the call-site change alone:

- The authorization loop iterates the requested ids and looks rows up in the returned map, rather than ranging the map. The authorized set is passed to `ClaimTasksForRuntimes`, where `maxTasks` can truncate it, so map iteration order would decide which runtimes get the remaining slots. (`ids` is itself built from a map upstream, so this pins the order to one source rather than making it fully deterministic — noted in a comment, not fixed here.)
- The authorized subset is collected into its own map instead of reusing `GetMany`'s. `GetMany` returns every row it found, while the post-claim loop treats a miss as a stray cross-daemon claim and drops the task — reusing the found-map directly would have silently widened what that loop accepts.

**`agent.go` — took your second option.** `loadAgentRuntimeAvailability` still reads directly, and the `RuntimeLookup` doc comment is narrowed to say what the code actually does. It names the exception rather than hedging with "primary", so the next reader doesn't have to rediscover it: the projection resolves rows for a list of agents rather than a runtime a caller asked for, and folding it into an existing source would make that source's rate stop meaning what its name says. Classifying it needs its own constant (and a slot in `AllRuntimeLookupSources()`'s prewarmed grid), which is a separate change. There's a matching pointer at the call site so the exception is discoverable from both ends.

**Test.** `TestAgentRuntimeLookupBatchClaimIsAttributed` in `runtime_lookup_metrics_test.go` claims through the batch endpoint with one existing and one unregistered id, asserting `daemon_api` deltas of `ok`=1 / `not_found`=1 / `error`=0, plus `other`=0. No AST guard — agreed, `603519d15` removed that deliberately.

I verified the test actually fails without the fix rather than just passing alongside it: reverting the call site to `h.Queries.GetAgentRuntimes` produces `daemon_api/ok delta = 0, want 1` and the same for `not_found`.

Validation on `8109a5ac6`:

- `go test ./internal/handler ./internal/service ./internal/metrics -count=1` — service and metrics green; handler green except `TestClaimTasksByRuntime_ClaimPollHintSchedulesNextDeferredTask`.
- That failure is environmental, not from this change: it reproduces on a clean tree at the same commit, and the cause is my local Postgres container's clock running ~250ms ahead of the host. The test seeds `fire_at = now() + 5s` in the database but bounds the delay against host time, so it lands at 5023–5109ms against a hard 5000ms ceiling. It passes in CI, where the Postgres service shares the runner's clock: `backend-tests` is green on `8109a5ac6`.
- 143 tests pass across the claim / deregister / lookup / attachment filter.
- `go build ./...`, `go vet ./internal/handler ./internal/service`, `gofmt` clean on touched files.

Ready for re-review.

Bohan-J (approved):
Approving at `8109a5ac6`. The metric gap from my last pass is closed, and I verified it rather than taking the diff's word for it.

**Claim path is attributed.** `server/internal/handler/daemon.go:1761` now reads through `h.getAgentRuntimes(..., RuntimeLookupSourceDaemonAPI, ids)`, and `TestAgentRuntimeLookupBatchClaimIsAttributed` asserts the shape that actually matters: one existing id and one unregistered id in a single request produce `daemon_api/ok = 1`, `daemon_api/not_found = 1`, `daemon_api/error = 0`, and nothing lands on `other`. That is a per-requested-id assertion, not a per-query one, so it pins the contract the counter's help string claims. It is also the right test to have written instead of the AST guard.

**The authorization boundary survived the refactor.** `runtimeByID` is built only from rows that pass both `verifyDaemonWorkspaceAccess` and the daemon-ownership check, rather than being reused from `GetMany`'s result — which matters because the post-claim loop reads a miss in that map as a stray cross-daemon claim. The comment at `:1772-1775` says so explicitly, which is what I'd want the next reader to find.

**`agent.go:1004` took the option I offered.** Documented as an explicit exception at `server/internal/service/agent_runtime_lookup.go:20-26` and at the call site, with the reason it isn't folded into an existing source label (it would change what that label's rate means) and a note that classifying it needs its own constant. That's the honest resolution for a call site that predates this PR — the contract comment no longer claims more than the code does.

CI at this head is green including `backend-tests`, which runs against real Postgres and Redis services, so the DB-backed tests here genuinely executed rather than skipping.

Two non-blocking nits, both text only:

1. **The PR description is stale on the most important behaviour.** The risk paragraph still says "Batch-query failures fail safely by processing no rows, matching the previous per-item error paths' skip behavior." That was the first version, and it's the exact thing the review cycle changed: deregister now returns 500 before any write, and an attachment read failure returns `error_chat_attachment_load` and preserves the task for redelivery. Worth fixing before merge since the body tends to become the squash-commit message.

2. **The ordering rationale at `:1766-1771` is inaccurate.** It justifies iterating `ids` by saying `maxTasks` can cut `authorized[]` off partway, so map order would decide which runtimes get the remaining slots. I traced it: `ReclaimStaleDispatchedTasksForRuntimes` applies `LIMIT max_tasks` under `ORDER BY atq.priority DESC, atq.dispatched_at ASC`, `ListQueuedClaimCandidatesByRuntimes` orders by `priority DESC, created_at ASC`, and step 6 of `ClaimTasksForRuntimes` walks candidates in that SELECT order. Nothing downstream consumes the array's order, so truncation is decided by priority/FIFO across the whole set either way. Keep the loop — it's fine and gives stable logs — but the stated reason will mislead whoever tries to preserve it later. The parenthetical hedge about `ids` coming from a map is good instinct; the sentence above it is what's wrong.

Neither blocks. Thanks for working through three rounds on this — the failure semantics and the test coverage ended up meaningfully stronger than the original patch, and the batching itself was right from the start. Leaving the merge to a maintainer.

Bohan-J:
Thanks for this, and for staying with it through three review rounds — the version that's landing is meaningfully better than a straight N+1 fix.

Worth naming what you added along the way: batching two reads is the easy half, and you also worked out what each one should do when the batch read *fails*. Deregister now returns 500 before any write instead of reporting a successful shutdown with every runtime still online, and an attachment read failure preserves the task for redelivery instead of starting an agent that can't see the user's files. Both of those are wider failure modes than the per-item code had, you found the repo's existing contract for each, and you covered them with tests. The claim-path metric attribution in the last round closed the one thing that would have made `multica_agent_runtime_lookup_total` misleading right after this merged.

Merging now.

Two documentation-only follow-ups from my review, neither of which blocks this:

- The PR description's risk paragraph still describes the original "batch failures skip as empty results" behaviour, which the review cycle replaced. The repo squashes using commit messages rather than the PR body, so this won't reach the history — but it's worth correcting if you reference this PR later.
- The ordering rationale at `server/internal/handler/daemon.go:1766-1771` attributes the `ids` iteration order to `maxTasks` truncation; the downstream queries order by `priority`/FIFO, so array order doesn't decide it. The loop itself is fine to keep. Happy for that comment to be fixed in a later pass rather than holding this up.

PR #7700, opened 2026-08-28
MUL-6795: VENU-68: contain analytics leaderboard horizontal overflow

### LINKED ISSUE #7698: [Bug]: Analytics leaderboard does not scroll horizontally on narrow widths
### Deployment type

Official App

### Area

Web app

### What happened?

The Analytics/Usage leaderboard does not handle horizontal overflow locally on narrow viewports or when its content is wider than the card.

The leaderboard uses a fixed multi-column grid for agent, bar, tokens, cost, time, and tasks. When the available width is smaller than the grid, the right-side columns can become clipped or the whole page can overflow horizontally. There is no horizontal scrollbar scoped to the leaderboard, so users cannot reliably reach every column.

Expected behavior:

- The leaderboard should scroll horizontally inside its own card when its columns exceed the available width.
- All columns should remain reachable.
- The page itself should not gain unintended horizontal overflow.
- Existing desktop-width layout should remain unchanged.

I searched the existing issues and pull requests for leaderboard, horizontal scroll, scroll, and overflow, but did not find a matching report or fix. The current `main` implementation in `packages/views/dashboard/components/leaderboard.tsx` also appears to have no local horizontal overflow container.

Is this already being addressed elsewhere? If not, I can prepare a focused PR with a regression test after validating the behavior locally.

### Steps to reproduce

1. Open the Analytics/Usage page in a workspace with leaderboard data.
2. Reduce the browser or app window width until the leaderboard columns exceed the card width.
3. Try to reach the Tokens, Cost, Time, and Tasks columns.
4. Observe that the leaderboard does not provide a scoped horizontal scrollbar; content is clipped or causes page-level horizontal overflow.

### Screenshots (optional)

Image

```shell
Suspected component: packages/views/dashboard/components/leaderboard.tsx
Proposed workflow: reproduce locally, add/update regression coverage, run the relevant frontend tests and type checks, then open a linked PR from iceprosurface/multica.
```

iceprosurface:
Fixes #7698
Fixes VENU-68

Review follow-up tracked in VENU-129.

## Summary

- contain leaderboard columns in a named, keyboard-focusable horizontal scroll region
- replace the detached `44rem` container width with intrinsic column constraints: Agent `10rem`, Progress `6rem`, plus the existing fixed metric tracks
- preserve the full trailing padding with `min-width: fit-content` on each grid while allowing the flexible tracks to expand on wider cards
- reuse the inset focus treatment from `packages/views/editor/styles/zoom-canvas.css` so keyboard focus stays visible without adding an outer halo
- test the region through its role/name contract and pin the product-level CSS constraints instead of Tailwind class strings or DOM parent hops

## Width rationale

The grid now owns its readability policy at the column level:

- Agent: `10rem` for a 24px avatar, an 8px gap, and 128px of readable name
- Progress: `6rem`, keeping the 8px bar at a 12:1 comparison length
- Metrics: `5rem + 5rem + 5rem + 4rem = 19rem`
- Gutters and row padding: `3.75rem + 2rem`

Together these form a natural `40.75rem` / `652px` scroll floor, but the code does not duplicate that total as another magic constant. Above the floor, the `1.6fr / 1fr` tracks continue to expand normally.

## Accessibility

The local scroller now has `role="region"`, the translated leaderboard title as its accessible name, and `tabIndex={0}`. Keyboard users can focus the region directly and use the browser's native horizontal scrolling keys. Its `:focus-visible` state uses `2px solid var(--ring)` with `outline-offset: -2px`, matching the large-interaction-region pattern in `zoom-canvas.css`.

## Verification

- `pnpm --filter @multica/views exec vitest run dashboard/components/dashboard-page.test.tsx` (24 passed)
- `pnpm --filter @multica/views lint` (0 errors; 27 pre-existing warnings outside the changed files)
- `git diff --check`
- `pnpm --filter @multica/views test` was also attempted on the rebased tree: 4,872 passed and 20 unrelated tests failed across 10 files, primarily on existing 5s/30s timeouts. The focused dashboard suite above passes; CI is re-running on the pushed head.

## Responsive recordings

These recordings supersede the earlier single 375px capture. Each enters the named region from the preceding button with Tab, scrolls with the keyboard, and displays live region/page geometry:

- [375×812](https://static.multica.ai/workspaces/093192fd-8085-4b7c-800d-038a539f4998/01a056cf-2f47-71b3-9055-473a5af12ac4.webm): region `325/652px`, max scroll `327px`; page `375/375px`
- [640×812](https://static.multica.ai/workspaces/093192fd-8085-4b7c-800d-038a539f4998/01a056cf-31ae-78f2-9e9c-f55b8051646b.webm): region `590/652px`, max scroll `62px`; page `640/640px`
- [704×812](https://static.multica.ai/workspaces/093192fd-8085-4b7c-800d-038a539f4998/01a056cf-3405-70f3-9519-a433e4e2a406.webm): region `654/654px`; page `704/704px`
- [1024×812](https://static.multica.ai/workspaces/093192fd-8085-4b7c-800d-038a539f4998/01a056cf-364d-7d42-b5e1-8be54f19b137.webm): region `850/850px`; page `1024/1024px`
- [1440×900](https://static.multica.ai/workspaces/093192fd-8085-4b7c-800d-038a539f4998/01a056cf-38b4-7578-9d0e-be748f42bbd1.webm): region `1102/1102px`; page `1440/1440px`

### DIFF
--- a/packages/views/dashboard/components/dashboard-page.test.tsx
+++ b/packages/views/dashboard/components/dashboard-page.test.tsx
@@ -742,14 +742,12 @@ describe("DashboardPage — leaderboard density", () => {
     const user = userEvent.setup();
     renderDashboard();
 
-    // Scoped to the leaderboard card — the trend chart's metric toggle owns
-    // a "Time" button too.
-    const card = screen
-      .getByRole("list", { name: "Leaderboard" })
-      .closest(".rounded-lg") as HTMLElement;
     // Re-ranking must not quietly reveal the tail: the cap belongs to the
     // list, not to one metric.
-    await user.click(within(card).getByRole("button", { name: "Time" }));
+    const sortControls = within(
+      screen.getByRole("group", { name: "Rank agents by" }),
+    );
+    await user.click(sortControls.getByRole("button", { name: "Time" }));
 
     const list = within(screen.getByRole("list", { name: "Leaderboard" }));
     expect(list.getAllByRole("listitem")).toHaveLength(10);
@@ -765,18 +763,25 @@ describe("DashboardPage — leaderboard density", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("contains wide leaderboard columns in a local horizontal scroller", () => {
+  it("exposes wide columns through a named keyboard-focusable local scroller", () => {
     renderDashboard();
 
-    const list = screen.getByRole("list", { name: "Leaderboard" });
-    const grid = list.parentElement;
-    const scroller = grid?.parentElement;
+    const scroller = screen.getByRole("region", { name: "Leaderboard" });
+    const list = within(scroller).getByRole("list", { name: "Leaderboard" });
+    const row = within(list).getByRole("listitem");
 
-    expect(grid).toHaveClass("min-w-[44rem]");
+    expect(scroller).toHaveAttribute("tabindex", "0");
+    scroller.focus();
+    expect(scroller).toHaveFocus();
     expect(scroller).toHaveClass(
       "overflow-x-auto",
       "overscroll-x-contain",
       "[-webkit-overflow-scrolling:touch]",
     );
+    expect(row).toHaveStyle({
+      minWidth: "fit-content",
+      gridTemplateColumns:
+        "minmax(10rem, 1.6fr) minmax(6rem, 1fr) 5rem 5rem 5rem 4rem",
+    });
   });
 });
--- a/packages/views/dashboard/components/dashboard-page.test.tsx
+++ b/packages/views/dashboard/components/dashboard-page.test.tsx
@@ -744,8 +744,9 @@ describe("DashboardPage — leaderboard density", () => {
 
     // Scoped to the leaderboard card — the trend chart's metric toggle owns
     // a "Time" button too.
-    const card = screen.getByRole("list", { name: "Leaderboard" })
-      .parentElement as HTMLElement;
+    const card = screen
+      .getByRole("list", { name: "Leaderboard" })
+      .closest(".rounded-lg") as HTMLElement;
     // Re-ranking must not quietly reveal the tail: the cap belongs to the
     // list, not to one metric.
     await user.click(within(card).getByRole("button", { name: "Time" }));
@@ -762,5 +763,20 @@ describe("DashboardPage — leaderboard density", () => {
     expect(
       screen.queryByRole("button", { name: "Show all" }),
     ).not.toBeInTheDocument();
+  });
+
+  it("contains wide leaderboard columns in a local horizontal scroller", () => {
+    renderDashboard();
+
+    const list = screen.getByRole("list", { name: "Leaderboard" });
+    const grid = list.parentElement;
+    const scroller = grid?.parentElement;
+
+    expect(grid).toHaveClass("min-w-[44rem]");
+    expect(scroller).toHaveClass(
+      "overflow-x-auto",
+      "overscroll-x-contain",
+      "[-webkit-overflow-scrolling:touch]",
+    );
   });
 });
--- a/packages/views/dashboard/components/leaderboard.css
+++ b/packages/views/dashboard/components/leaderboard.css
@@ -1,0 +1,11 @@
+/* Large interaction regions share the inset focus treatment defined by
+   editor/styles/zoom-canvas.css: the indicator stays visible without adding
+   an outer halo around the containing card. */
+.leaderboard-scroll-region {
+  outline: none;
+}
+
+.leaderboard-scroll-region:focus-visible {
+  outline: 2px solid var(--ring);
+  outline-offset: -2px;
+}
--- a/packages/views/dashboard/components/leaderboard.tsx
+++ b/packages/views/dashboard/components/leaderboard.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo, useState } from "react";
+import { useMemo, useState, type CSSProperties } from "react";
 import { EyeOff, Trash2 } from "lucide-react";
 import { ActorAvatar } from "../../common/actor-avatar";
 import { formatTokens } from "../../runtimes/utils";
@@ -13,6 +13,7 @@ import {
   type AgentDashboardRow,
 } from "../utils";
 import { Segmented } from "./dashboard-shared";
+import "./leaderboard.css";
 
 // Which metric ranks the leaderboard. Drives row order, progress bar
 // width, and which column header is emphasised — keeping the three in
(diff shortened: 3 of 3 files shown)


iceprosurface:
## Narrow viewport demo

Leaderboard horizontal scroll demo

https://github.com/user-attachments/assets/22c8eb50-dad1-4421-b5e2-25c85d544f7f

Recorded against PR head `9e0c5963b` at a 375×812 viewport. The leaderboard's local scroll area measured 325px wide with 704px of content and scrolled `0 → 379 → 0 → 379`; the document root remained `375/375px`, confirming that the page itself did not gain horizontal overflow.

Bohan-J (changes requested):
Thanks for the fix and for the 375px capture — the direction is right: keep the horizontal movement inside the leaderboard card, keep every column reachable, and leave the page itself alone. The header and sort controls being left outside the scroller is the right call too, and CI is green with a clean merge onto `main`.

Two things need to be resolved before this can go in.

### 1. `min-w-[44rem]` needs a reviewable rationale — `packages/views/dashboard/components/leaderboard.tsx:139`

The grid's hard floor today is about `24.75rem`: fixed tracks `5rem × 3 + 4rem` = `19rem`, five `gap-3` gutters = `3.75rem`, plus `px-4` = `2rem`. The two `minmax(0,fr)` tracks collapse to `0` before that point, which is why the agent name disappears first and the row only escapes the card below ~396px.

So `44rem` is not the overflow threshold — it is a separate readability policy, and it also changes the ~`24.75rem`–`44rem` band from "compressed agent and bar columns" to "scrolls horizontally". That is a defensible choice, but right now it is an unexplained number and the only evidence in the PR is the 375px case, which never exercises that band.

Please make the constraint auditable: state the minimum space the agent-name and progress columns need to stay readable, either as a comment deriving `44rem` from the column widths or by composing the value from them. The neighbouring constants in this file (`LEADERBOARD_LIMIT`, `LEADERBOARD_GRID`) all carry their "why", so this would match the file's convention. The regression should then pin that product constraint rather than the literal utility class.

### 2. The scroll region is not keyboard-operable and has no accessible name — `packages/views/dashboard/components/leaderboard.tsx:138`

`overflow-x-auto` on its own does not give keyboard users a way to scroll the region. Checking the same structure across engines: with focus on an element inside a row, `ArrowRight` moves the scroll offset in Chromium and Firefox, but WebKit stays at `0`. So the issue's own acceptance line — "All columns should remain reachable" — does not hold on every supported browser yet.

Please make the scroller focusable and announced (`tabIndex={0}` plus `role="region"` and an `aria-label`, reusing the leaderboard title is fine) and add a regression covering it.

Related: the new test at `packages/views/dashboard/components/dashboard-page.test.tsx:767` reaches the scroller through two `parentElement` hops and asserts Tailwind class strings, so it only proves the classes are present — that coupling already forced the neighbouring test to move from `.parentElement` to `.closest(".rounded-lg")`. Once the region has a role and a name, assert through that stable contract instead.

### Notes, not blocking

- The `hash: ""` addition to the test harness is unrelated to this fix, and `main` has since added the same line independently. It still merges cleanly, so no action needed — just flagging that it is not part of the change.
- `errors-tab.tsx` has a same-shaped fixed grid (`OFFENDER_GRID`) with a much narrower floor. It is deliberately out of scope here; please don't widen this PR for it.

iceprosurface:
Addressed both blocking review points and rebased the PR onto current `main` (`0a7992c20`). The new head is `42b6df0ae`.

1. The unexplained `min-w-[44rem]` is gone. Agent and Progress now own explicit `10rem` and `6rem` readability floors through `minmax(...)`; with the existing fixed tracks, gutters, and row padding, the browser naturally forms the complete `40.75rem` / `652px` scroll floor. `min-width: fit-content` preserves both sides' padding without duplicating the total width in code.
2. The scroller is now a translated, named `role="region"` with `tabIndex={0}`, so it can be reached directly and scrolled through native keyboard behavior across engines.
3. The focus treatment no longer uses the earlier outer ring. It is an inset `2px` outline with `-2px` offset, explicitly based on `packages/views/editor/styles/zoom-canvas.css`; the source reference is also documented in `leaderboard.css`.
4. Tests now locate the region and sort controls by role/name, exercise focusability, and assert the product-level grid constraints rather than `parentElement`, `.closest()`, or a Tailwind literal.

Updated recordings (these supersede the earlier single-width demo):

### 375×812

Region `325/652px`, max `327px`; page `375/375px`.

https://github.com/user-attachments/assets/4af83efe-3c6b-4ce5-b961-b1f770c5e9f3

### 640×812

Region `590/652px`, max `62px`; page `640/640px`.

https://github.com/user-attachments/assets/2e51b3db-0f2e-4f65-93c8-13e143714814

### 704×812

Region `654/654px`; page `704/704px`.

https://github.com/user-attachments/assets/d1b00aaf-1a75-4850-8f6c-9b2b2c920cba

### 1024×812

Region `850/850px`; page `1024/1024px`.

https://github.com/user-attachments/assets/7faa3157-c2df-4d82-8a2a-27214d29aff6

### 1440×900

Region `1102/1102px`; page `1440/1440px`.

https://github.com/user-attachments/assets/6a054cbf-02ec-47ce-87b6-a548a021b3d3

Local verification: focused dashboard suite `24/24`, views typecheck passed, views lint has 0 errors (27 pre-existing warnings outside this diff), and `git diff --check` passed. The full local views run reached 4,872 passing tests and 20 unrelated failures across 10 files, mostly existing timeouts; CI has been triggered on this head.

iceprosurface:
@Bohan-J The two blocking items are addressed in head `42b6df0ae`, rebased onto current `main`. Could you please take another look? The preceding follow-up comment maps each review point to the implementation and includes recordings at five viewport widths.

Bohan-J (approved):
Thanks for the follow-up — both blocking items are closed, and the way you closed the first one is better than what I asked for.

1. Rather than documenting `44rem`, you removed the detached total entirely: `minmax(10rem, 1.6fr) minmax(6rem, 1fr)` puts the readability floor on the column that owns it, with the derivation spelled out (24px avatar + 8px gap + 128px name; 8px bar at a 12:1 comparison length), and `min-width: fit-content` lets the intrinsic width fall out of the track definitions instead of being maintained by hand. The floors also keep the header and every row at the same intrinsic width, so nothing drifts out of alignment once the region scrolls.
2. The scroll region is now a named `region` with `tabIndex={0}` and a visible inset focus ring, and the test locates it by role and name and asserts focus plus the column constraint. Replacing the `.closest(".rounded-lg")` lookup in the neighbouring test with the sort group's role/name was a nice extra.

CI is green and it merges cleanly. Merging now — thanks for the clear reproduction, the narrow-viewport capture, and for keeping the change tightly scoped.

PR #7703, opened 2026-08-28
MUL-6796: fix mobile issue header activity status (VENU-69)

### LINKED ISSUE #7702: [Bug]: Agent activity label crowds mobile issue headers
## Deployment type

Official App

## Area

Web app (responsive issue details and inbox)

## What happened?

The issue-detail header shows an Agent avatar and live status label such as “Working” while an Agent handles the issue. On narrow mobile viewports, the non-shrinking label competes with the breadcrumb/title and the header action buttons. The inbox's embedded detail is the first place where the header runs out of room because it also includes Back and Done controls, but the direct issue page is tight for the same reason.

This report originally identified the inbox list row as the failing surface. Follow-up inspection showed that row already gives its detail text `min-w-0` plus ellipsis and degrades gracefully. The real defect is the shared issue-detail header, so the report has been corrected to match the implemented fix.

Expected behavior below the `md` breakpoint:

- hide the visible status copy in every `IssueDetail` host;
- retain the Agent avatar, popover trigger, and full accessible name;
- give the avatar-only trigger a larger touch target;
- keep the full status label on desktop.

## Steps to reproduce

1. Open an issue with a running or queued Agent task.
2. Resize the web app to a 390px-wide viewport.
3. Observe the issue-detail header alongside its breadcrumb/title and action buttons.
4. Compare with a desktop-sized viewport, where the full status label should remain visible.

### Before — mobile header at 390×844

Mobile header before

### After — real application at 390×844

Mobile header after in the real application

## Additional context

The proposed fix is #7703. No inbox-list-row behavior change is intended.

iceprosurface:
## What does this PR do?

Keeps issue-detail headers usable on mobile while preserving the live Agent cue. Below the `md` breakpoint, `IssueAgentHeaderChip` hides its visible status label in every host, retains the avatar and full accessible name, and expands the avatar-only trigger from roughly 32×28px to at least 36×36px. Desktop continues to show the complete status label at its original height.

The original diagnosis pointed to the inbox list row. Review confirmed that the row already truncates its detail text gracefully; the real defect is the shared issue-detail header. #7702 has been updated to record that corrected diagnosis, and this final diff intentionally contains no list-row change.

## Related Issue

Closes #7702

Resolves VENU-69

## Type of Change

- [ ] New feature
- [ ] Documentation update
- [x] Tests

## Changes Made

- Apply `hidden md:inline` directly in the header chip instead of threading opt-in props through `InboxPage` and `IssueDetail`.
- Keep the Agent avatar, popover trigger, and `aria-label` at every breakpoint.
- Increase the mobile trigger to `h-9 min-w-9` while retaining the desktop `h-7` presentation.
- Add focused regression coverage for the responsive label, target size, avatar, and accessible name.
- Squash the PR to one commit and rebase it onto the latest `main`.

## How to Test

```bash
pnpm --filter @multica/views exec vitest run \
  inbox/components/inbox-page.test.tsx \
  inbox/components/inbox-list-item.test.tsx \
  issues/components/issue-agent-header-chip.test.tsx \
  issues/components/issue-agent-activity-indicator.test.tsx
pnpm --filter @multica/views typecheck
pnpm --dir packages/views exec eslint \
  issues/components/issue-agent-header-chip.tsx \
  issues/components/issue-agent-header-chip.test.tsx
git diff --check
```

Local result from the code-changing revision: 4 test files passed, 70/70 tests; typecheck, focused ESLint, and `git diff --check` passed.

## Browser QA

The 390×844 check below uses the complete Next.js application shell and production inbox/issue-detail components. Only backend API responses are mocked to keep a running Agent task deterministic.

- viewport: 390×844;
- Agent label computed `display: none`;
- Agent trigger measured 36×36px;
- real header measured `clientWidth = scrollWidth = 390px` (no horizontal overflow);
- back control, title/breadcrumb, Agent avatar, action buttons, and sidebar toggle are all present.

### Mobile — Before (390×844)

Mobile Before — Agent status text crowds the inbox header

### Mobile — After, real application (390×844)

Mobile After — real inbox header remains within the viewport

### Desktop — Regression check (1024×768)

Desktop regression — full Agent status remains visible

## Checklist

- [x] The issue record and PR description match the final surface being fixed
- [x] Responsive behavior applies to every `IssueDetail` host
- [x] Mobile touch target is larger without changing desktop height
- [x] Screen-reader and keyboard access are preserved
- [x] Real-application mobile evidence is included without adding PR assets to the repository
- [x] Local focused tests and typecheck pass
- [x] PR history is one commit rebased onto current `main`

## AI Disclosure

**AI tool used:** Multica coding agent

**Approach:** Follow the review to remove the unnecessary opt-in seam, correct the issue/PR record, enlarge the mobile touch target, verify the complete application header at 390px, and produce a single rebased commit.

### DIFF
--- a/packages/views/issues/components/issue-agent-header-chip.test.tsx
+++ b/packages/views/issues/components/issue-agent-header-chip.test.tsx
@@ -318,6 +318,20 @@ describe("IssueAgentHeaderChip", () => {
     expect(screen.getByText("Walt 在工作")).toBeInTheDocument();
   });
 
+  it("hides the visible label on mobile while preserving a larger tap target and accessible name", () => {
+    mockState.tasks = [makeTask({ status: "queued" })];
+
+    renderWithI18n(<IssueAgentHeaderChip issueId="issue-1" />);
+
+    const trigger = screen.getByRole("button", { name: "Walt is queued" });
+    expect(trigger.querySelector("[data-slot='avatar']")).not.toBeNull();
+    expect(trigger.className).toContain("h-9 min-w-9");
+    expect(trigger.className).toContain("md:h-7 md:min-w-0");
+    const label = screen.getByText("Walt is queued");
+    expect(label.className).toContain("hidden");
+    expect(label.className).toContain("md:inline");
+  });
+
   it("does not render when the issue has only terminal tasks", () => {
     // The list is issue-scoped by the endpoint, so the chip's only job is to
     // ignore terminal statuses (those are the execution log's story).
--- a/packages/views/issues/components/issue-agent-header-chip.tsx
+++ b/packages/views/issues/components/issue-agent-header-chip.tsx
@@ -178,7 +178,7 @@ function ActiveChip({
               // header. Queued-only state stays calm (no beam) to reserve the
               // motion for work that is genuinely in flight.
               className={cn(
-                "flex h-7 max-w-[11rem] items-center gap-1.5 rounded-md px-1.5 text-muted-foreground outline-none transition-colors hover:bg-accent/60 focus-visible:ring-2 focus-visible:ring-ring",
+                "flex h-9 min-w-9 max-w-[11rem] items-center justify-center gap-1.5 rounded-md px-2 text-muted-foreground outline-none transition-colors hover:bg-accent/60 focus-visible:ring-2 focus-visible:ring-ring md:h-7 md:min-w-0 md:justify-start md:px-1.5",
                 anyRunning && "border-beam bg-brand/5",
               )}
             />
@@ -191,7 +191,10 @@ function ActiveChip({
             opacity={anyRunning ? "full" : "half"}
           />
           <span
-            className={`min-w-0 truncate text-caption ${anyRunning ? "text-info" : "text-muted-foreground"}`}
+            className={cn(
+              "hidden min-w-0 truncate text-caption md:inline",
+              anyRunning ? "text-info" : "text-muted-foreground",
+            )}
           >
             {label}
           </span>


Bohan-J (changes requested):
Thanks for the fix — the header clipping in your "before" screenshot is real, and hiding the status copy while keeping the avatar is the right shape for it. Two things need to be settled before this can merge.

## Blocking: `Closes #7702` no longer matches what this PR changes

#7702 reports the **inbox list row** — the activity label crowding the notification detail and timestamp in the row's lower-right metadata group. The second commit reverted the list-row change, so the final diff only touches the issue-detail header chip, a different surface.

For what it's worth, we agree the header is the real defect: in the row, the detail label is `min-w-0 … text-ellipsis` and degrades gracefully, which we don't consider a bug. But as it stands, merging this would silently close a report about a surface the PR doesn't touch.

Please either restore and verify the list-row fix, or update #7702 and this PR's description so the record says what is actually being fixed and why the original diagnosis was revised.

## Needs an answer: why is this opt-in?

The description justifies the prop chain (`IssueDetail.hideAgentLabelOnMobile` → `IssueAgentHeaderChip.hideLabelOnMobile`) with "direct issue pages and other consumers retain their existing behavior". That premise doesn't hold for this component:

- `IssueAgentHeaderChip` has exactly one call site — `packages/views/issues/components/issue-detail.tsx:2778`. There are no other consumers. The "other consumers of the shared activity indicator have enough room" note in #7702 was about `IssueAgentActivityIndicator` (inbox row / issue list row / board card), which this PR no longer touches.
- The only variation left is between `IssueDetail`'s two hosts, and both are equally tight on a phone. Computed from the classes at a 390px viewport, the direct issue page (no inbox back button) leaves roughly **45px** for breadcrumb + title:
  `390 − 32 (px-4) − 28 (sidebar trigger) − 8 − 277 (chip max-w-[11rem] = 176 + divider + three icon-sm 28px buttons + gap-1) − 8 ≈ 45`
  The chip alone is about 4× the remaining title area, and a parent breadcrumb segment is `shrink-0`, so it overflows there too. The inbox is simply the first host where the number goes negative, because of the extra back and done buttons.

Preference: drop both props and make the label `hidden md:inline` unconditionally. That is one line instead of three files, and it fixes the direct issue page and the desktop issue window at the same time. The `hideAgentLabelOnMobile === true` assertion in `inbox-page.test.tsx` should go with it — it currently locks the opt-in design in place.

If you believe the direct issue page should keep the full label at that width, please say why and we will take that as an explicit decision rather than a side effect.

## Nits

- The two "after" screenshots are component fixtures — empty skeleton body, and the real header also carries the sidebar-toggle button. They show the CSS applies; they don't show the real inbox header fits at 390px. A real-app "after" at 390px would close this.
- With the label hidden, the chip becomes a roughly 32×28px tap target while being the only way to read agent status on touch. Not a regression (it was `h-7` before), but a little more padding below `md` would help.

## What's good

- `hidden md:inline` matches the 32 existing responsive-hide sites in this repo; no new `max-md:hidden` style introduced.
- `aria-label` is preserved, so screen readers and keyboard access are unaffected.
- Clean final diff (5 files), no stray plan or record directories.
- Verified locally on this branch: the four focused test files pass (71 tests) and `pnpm --filter @multica/views typecheck` is clean.

iceprosurface:
Addressed the review in `07f3b7bf9`:

- corrected #7702 and this PR description to identify the shared issue-detail header rather than the inbox list row;
- removed the opt-in prop chain and made `hidden md:inline` unconditional in the header chip;
- increased the mobile avatar-only trigger to 36×36px while retaining the desktop height;
- added a real-application 390×844 screenshot and measured the complete header at `clientWidth = scrollWidth = 390px`;
- squashed the branch to one commit and rebased it onto current `main`.

Local verification passes: 4 focused files / 70 tests, `@multica/views` typecheck, focused ESLint, and `git diff --check`.

iceprosurface:
Added the requested responsive real-app QA contact sheet. It uses the complete Next.js application shell and production inbox / issue-detail components, with only backend responses mocked so every viewport captures the same running-Agent state.

Widths covered: `390`, `640` (`sm`), `768` (`md`), `1024` (`lg`), `1280` (`xl`), `1536` (`2xl`), and `1920` px; each viewport is `844px` tall.

- At 390px and 640px, the visible Agent label is hidden and the trigger measures 36×36px.
- At 768px and above, the full Agent label is restored and the trigger returns to its 28px desktop height.
- At every width, both the page and issue header have equal `clientWidth` / `scrollWidth`, so there is no horizontal overflow.

PR #7703 responsive issue-header QA — 390px through 1920px

Bohan-J (approved):
Thanks for the thorough turnaround here — this addressed both blockers properly, and the extra QA went beyond what was asked.

What landed since the last round:

- **`Closes #7702` now matches the diff.** Rewriting the issue to name the shared issue-detail header, and stating plainly that the inbox list row already truncates gracefully, keeps the record honest instead of letting the original report close silently against an unrelated fix.
- **The opt-in prop chain is gone.** `hidden md:inline` applied directly in the chip takes the diff from five files to two and fixes the direct issue page and the desktop issue window along with the inbox, rather than leaving a known-bad surface behind a flag.
- **Mobile trigger is now 36×36px**, which matters more than usual here since the avatar becomes the only way to read Agent status at that width.

Verified locally on `8dddbca5`:

- focused suite 70/70, and the full `@multica/views` suite at 409 files / 4863 tests passes;
- `@multica/views` typecheck and focused ESLint clean;
- reverting the component change makes the new test fail, so it genuinely guards the responsive behavior;
- no `hideAgentLabelOnMobile` / `hideLabelOnMobile` references remain, and `IssueAgentHeaderChip` still has exactly one call site;
- the responsive contact sheet checks out — 768px is the tightest width that still renders the label, and the title has ample room there, which resolves the `md` vs `lg` breakpoint question I raised earlier.

One non-blocking note for the record: the "before" and desktop screenshots in this PR and in #7702 are hosted from a branch on your fork. If that branch is ever removed or garbage-collected, the permanent issue record loses its before-evidence — the real-app "after" shot and the contact sheet are GitHub attachments and are safe. Worth re-uploading the before image as an attachment at some point.

Approving and merging. Thanks again for the fix and for the careful evidence.

PR #7714, opened 2026-08-28
MUL-6800: perf(handler): reuse HTTP clients instead of per-request allocation

### LINKED ISSUE #7712: [Bug]: Handler hot paths create new HTTP clients per upstream request
### Deployment type

Official App and self-hosted server deployments

### Area

Skills/import integrations and GitHub integration

### What happened?

Several server handler hot paths allocate a new `*http.Client` for each request while calling stable upstream services:

- skill search/import/refresh (`SearchSkills`, `ImportSkill`, `RefreshSkill`)
- GitHub installation repository browsing (`fetchGitHubInstallationRepositories`)

Each throwaway client owns a fresh transport and an empty connection pool, so repeated requests cannot reuse TCP/TLS keep-alive connections. Under normal skill operations or repository browsing, the server pays unnecessary handshake and allocation cost even though these paths use fixed timeout values and talk to stable upstream APIs.

Expected behavior: reuse package-level `http.Client` instances with the same timeout values, matching the existing long-lived client pattern used by the Forgejo integration, so these handlers share connection pools without changing request deadlines or response behavior.

### Steps to reproduce

1. Inspect `server/internal/handler/skill.go`: `SearchSkills` and `ImportSkill` construct `&http.Client{Timeout: 30 * time.Second}` inside the request path.
2. Inspect `server/internal/handler/skill_refresh.go`: `RefreshSkill` constructs the same per-request client before re-fetching the stored skill origin.
3. Inspect `server/internal/handler/github.go`: `fetchGitHubInstallationRepositories` constructs `&http.Client{Timeout: 15 * time.Second}` before the installation-token exchange and repository listing calls.
4. Exercise repeated skill search/import/refresh or GitHub repository browsing; each request starts with a fresh client/connection pool instead of reusing idle upstream connections.

### Additional context

`plugin_hook.go` should stay as-is: that path already creates each outbound request with an explicit per-call `context.WithTimeout`, and hooks can configure `TimeoutMs`. Adding a fixed client-level timeout there would incorrectly truncate hooks whose configured timeout is longer than the default.

importcpp:
## What does this PR do?

The skill search/import/refresh handlers and the GitHub installation-token exchange each constructed a fresh `*http.Client` per request. Every throwaway client carries its own transport with an empty connection pool, so repeated calls cannot reuse TCP/TLS keep-alive connections and pay unnecessary handshake/allocation cost on existing user-facing flows.

This hoists those clients to package-level vars with the same timeout values, matching the existing long-lived client pattern in `integrations/vcs/forgejo.go`, without changing request deadlines or response behavior.

## Related Issue

Closes #7712

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/skill.go`: add `skillHTTPClient` with the existing 30s timeout and reuse it for `SearchSkills` and `ImportSkill`.
- `server/internal/handler/skill_refresh.go`: reuse `skillHTTPClient` for `RefreshSkill` and remove the now-unused `time` import.
- `server/internal/handler/github.go`: add `githubTokenHTTPClient` with the existing 15s timeout and reuse it across the GitHub installation-token exchange, repository listing, and token revocation path.
- Leave `plugin_hook.go` unchanged because hooks already run under per-call `context.WithTimeout`, and a fixed client-level timeout would incorrectly truncate hooks with longer configured `TimeoutMs`.

## How to Test

1. `cd server && gofmt -w internal/handler/github.go internal/handler/skill.go internal/handler/skill_refresh.go`
2. `cd server && go build ./internal/handler/...`
3. `cd server && go vet ./internal/handler/...`
4. `cd server && go test ./internal/handler/ -run 'Skill|GitHub'`

## Checklist

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** Audited outbound HTTP client allocation in existing handler paths, compared against the repository's Forgejo integration pattern, and kept the change scoped to call sites with fixed timeout requirements. Also checked `plugin_hook.go` and intentionally left it unchanged because that path has configurable per-hook context deadlines.

### DIFF
--- a/server/internal/handler/github.go
+++ b/server/internal/handler/github.go
@@ -34,6 +34,11 @@ import (
 // point App-authenticated calls at an httptest server without touching GitHub.
 var githubAPIBase = "https://api.github.com"
 
+// githubTokenHTTPClient is reused across installation-token exchanges (a hot
+// path against api.github.com) so they share one keep-alive connection pool
+// instead of building a throwaway client — and its idle connections — per call.
+var githubTokenHTTPClient = &http.Client{Timeout: 15 * time.Second}
+
 const (
 	githubReturnToGitHub       = "github"
 	githubReturnToRepositories = "repositories"
@@ -824,7 +829,6 @@ func fetchGitHubInstallationRepositories(
 		return GitHubRepositoriesResponse{}, errors.New("github App JWT credentials unavailable")
 	}
 
-	client := &http.Client{Timeout: 15 * time.Second}
 	tokenEndpoint := fmt.Sprintf(
 		"%s/app/installations/%d/access_tokens",
 		strings.TrimRight(githubAPIBase, "/"),
@@ -841,7 +845,7 @@ func fetchGitHubInstallationRepositories(
 	}
 	setGitHubAPIHeaders(tokenReq, appJWT)
 	tokenReq.Header.Set("Content-Type", "application/json")
-	tokenResp, err := client.Do(tokenReq)
+	tokenResp, err := githubTokenHTTPClient.Do(tokenReq)
 	if err != nil {
 		return GitHubRepositoriesResponse{}, fmt.Errorf("create installation token: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


multica-eve:
Thanks for putting this together. We are closing this PR because its main performance premise does not hold for these call sites.

Each existing `http.Client` sets only `Timeout`; its `Transport` remains nil. In Go's `net/http`, a nil transport uses the process-wide `http.DefaultTransport`, whose connection pool is already shared and reuses idle TCP/TLS connections across these client values. Moving the clients to package-level variables therefore does not provide the claimed connection-reuse or handshake benefit.

We also do not have benchmark or production evidence that allocating the small client value itself is a material bottleneck. If such data emerges, we can revisit a focused change against the measured cost. Thanks again for the contribution.

PR #7716, opened 2026-08-28
MUL-6803: perf(handler): pool JSON response encode buffers

### LINKED ISSUE #7715: [Feature]: Reuse buffers in shared JSON response writers
### Deployment type

self-host

### Area

Self-hosting and deployment

### What do you want and why?

`writeJSON` and `writeMeasuredJSON` are shared response paths for most JSON HTTP responses. Both currently call `json.Marshal`, which allocates a fresh result slice for every response, and then append a trailing newline.

Because the slice returned by `json.Marshal` is commonly length-exact, appending that newline can require allocating another slice and copying the entire encoded body. The cost scales with response size and is especially wasteful for large list responses, adding allocation pressure and GC work on a hot server path.

The wire format must remain unchanged: default HTML escaping, exactly one trailing newline, accurate `Content-Length`, and the existing error behavior all need to be preserved.

- Encode into a reusable `*bytes.Buffer` obtained from a `sync.Pool` using `json.NewEncoder(buf).Encode(v)`. `Encode` includes the existing trailing newline and retains the default HTML escaping behavior.
- Compute `Content-Length` from the complete buffered response before writing headers.
- Return ordinary buffers to the pool, but discard buffers whose capacity exceeds 1 MiB so a one-off large response does not keep a large allocation retained indefinitely.
- Add regression coverage for byte identity, the single trailing newline, exact `Content-Length`, encoding failures, and concurrent buffer reuse.

_No response_

importcpp:
## What does this PR do?

Reduces allocation and copy overhead in the shared `writeJSON` and `writeMeasuredJSON` response paths without changing their wire behavior.

Previously, both writers encoded with `json.Marshal` and then appended a newline. The marshal result is commonly length-exact, so adding one byte can allocate a second slice and copy the entire response. This PR instead encodes into a pooled `*bytes.Buffer` with `json.NewEncoder(buf).Encode(v)`, which writes the trailing newline as part of encoding and allows the buffer capacity to be reused by later requests.

The existing behavior is preserved:

- default HTML escaping remains enabled;
- responses end with exactly one newline;
- `Content-Length` is calculated from the complete encoded body;
- encoding failures still return a JSON 500 response; and
- `writeMeasuredJSON` still reports the encoded response size and write errors.

Buffers larger than 1 MiB are not returned to the pool, preventing an unusually large response from retaining oversized backing storage in the pool.

## Related Issue

Closes #7715

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/handler.go`: add a shared `sync.Pool` of JSON encoding buffers and encode responses with `json.NewEncoder`.
- `server/internal/handler/handler.go`: discard pooled buffers above 1 MiB and retain accurate `Content-Length` and error responses.
- `server/internal/handler/handler_writejson_test.go`: update the writer identity documentation and add coverage for one trailing newline, encoding-error responses, and concurrent pool reuse.

## How to Test

2. Run `go vet ./...`.
3. Run `go test -race ./internal/handler/...`.
4. Confirm the writer tests cover byte-identical output, default HTML escaping, exactly one trailing newline, accurate `Content-Length`, encode-error fallback, and concurrent reuse without cross-response corruption.

## Checklist

The main risk is accidental response-format drift or sharing a live buffer across concurrent responses. Existing byte-identity and escaping tests remain in place, while new tests pin the single newline, exact length, error fallback, and concurrent isolation. A buffer is returned to the pool only after the synchronous response write has finished, and oversized buffers are discarded.

## AI Disclosure

**Prompt / approach:** Audited the shared JSON response writers for avoidable allocations, replaced the marshal-plus-append path with pooled buffered encoding, and retained the existing response format and length calculation. Added regression tests for output identity, error handling, and concurrent reuse. OpenAI Codex reviewed the final diff, tightened the performance claims, and prepared the issue and PR workflow.

## Screenshots (optional)

Not applicable; this is a backend response-encoding optimization with no UI change.

### DIFF
--- a/server/internal/handler/handler.go
+++ b/server/internal/handler/handler.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/hex"
@@ -12,6 +13,7 @@ import (
 	"net/netip"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


multica-eve:
Thanks for the contribution and for taking care to preserve the existing wire behavior. After reviewing the performance need independently, we are closing this PR because the measured benefit does not justify adding a global buffer-retention policy to the shared response path.

The main reasons are:

- We could not reproduce the premise that `json.Marshal` commonly returns a capacity-exact slice and that appending the trailing newline therefore copies the whole response. On the repository's Go 1.26.6 toolchain, representative 16 B, 4 KiB, 256 KiB, and 1 MiB payloads all appended in place.
- An exact `Content-Length` requires the complete response to be buffered before headers are written, so the slice returned by `json.Marshal` is already the required output buffer rather than an unnecessary second copy. The proposed change trades that allocation for application-level retained memory.
- The current high-frequency idle claim path is batched per daemon and returns only a tiny empty response. Larger skill payloads are fetched per bundle only on cache misses and then cached, so “large” and “high frequency” do not coincide on the motivating path.
- In a Go 1.26.6 microbenchmark, a 256 KiB payload improved only from about 135.1 µs to 131.6 µs. At 1 MiB, the proposed pool limit causes the buffer to be discarded, and the result slightly regressed from about 531.3 µs / 2 allocs to 534.1 µs / 3 allocs.

Without production profiles, endpoint size/concurrency data, or an end-to-end latency/GC improvement, this does not earn the additional `sync.Pool` lifecycle and the arbitrary 1 MiB retention threshold. If future profiling shows that response-buffer allocation is a material part of the server budget, we would be happy to reconsider a proposal driven by that workload and accompanied by threshold-boundary benchmarks.

This is a product/maintenance decision rather than a wire-correctness concern with your implementation. Thank you again for the thoughtful work.

PR #7723, opened 2026-08-29
MUL-6809: fix(daemon): hand off chat environment on explicit release

### LINKED ISSUE #7722: Chat replacement turn loses session when prior workdir is briefly busy
## Summary

A replacement chat turn can lose its provider-session continuity when it is claimed before the turn it supersedes has finished shutting down.

## Reproduction

1. Start a Codex-backed web chat turn that is still executing tools.
2. Send a replacement message, causing the active turn to be cancelled.
3. Let the replacement task be claimed before the cancelled process releases its managed environment root.

## Current behavior

`lockReusablePriorEnvRoot` receives `execenv.ErrEnvRootBusy` and immediately falls back to a fresh environment. Because the provider session store is tied to the prior environment, `gateResumeToReusedWorkdir` drops `PriorSessionID`. The replacement starts a fresh Codex thread and is told to reconstruct context with `multica chat history`.

This adds latency and token usage and discards the provider's working memory even though the prior environment becomes available shortly afterward.

## Expected behavior

For a chat continuation that already carries a provider resume pointer, allow a short, cancellable wait for the superseded turn to release its environment. Reuse the prior environment and provider session when the lock becomes available; retain the existing fresh-environment fallback when it does not.

mameikagou:
## What does this PR do?

Preserves provider-session continuity when a replacement chat turn is claimed just before the cancelled turn releases its managed environment.

Instead of polling the prior env-root lock every 50ms for an arbitrary five-second window, the daemon now registers every env-root claim with a one-shot release signal. A replacement chat turn carrying a resume pointer waits for the exact in-process holder to release the root, then immediately claims it and resumes the provider session.

Untracked locks may belong to another daemon process, so they keep the existing safe fresh-environment fallback. The wait remains bounded and cancellable through the existing task preparation context.

### Thinking path

1. Runtime logs showed the replacement task being claimed while the prior task was still executing tools.
2. The replacement hit `prior workdir is in use`, created a fresh environment, and invoked Codex with `resume_session=false`; the prior task released the environment shortly afterward.
3. Source tracing connected that sequence through `lockReusablePriorEnvRoot`, `gateResumeToReusedWorkdir`, and `SessionContinuityNoticeChatTranscript`.
4. A fixed polling window reduces the race but cannot express the actual lifecycle boundary: it can be too short, or add needless retry latency.
5. The fix therefore publishes the real env-root claim lifecycle and preserves every existing recovery fallback.

## Related Issue

Closes #7722

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Track daemon-owned env-root claims by canonical filesystem identity.
- Publish a one-shot signal after the exact holder releases its OS lock.
- Make Codex chat continuations wait on that signal instead of polling or sleeping.
- Keep cancellation, external-lock fallback, canonical-path handling, and TOCTOU identity checks intact.
- Add regression tests for explicit handoff, cancellation, and symlink/canonical path identity.

## How to Test

1. `cd server && go test -race ./internal/daemon -run 'Test(LockReusablePriorEnvRoot|EnvRootClaimHandoff)' -count=1`
2. `cd server && env -u MULTICA_TASK_CONFIG_ROOT -u MULTICA_TASK_WORKSPACES_ROOT go test ./internal/daemon -count=1`

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; daemon-only change)
- [x] I have updated relevant documentation to reflect my changes (not applicable; no user-facing contract change)
- [x] If this PR touches Chinese product copy, I checked it against the copy conventions (not applicable)
- [x] I have considered and documented any risks below

### Risk

The replacement may wait until the existing task preparation deadline if a tracked holder never releases. Normal cancellation closes the provider process and releases the claim; explicit preparation cancellation stops the waiter immediately. Locks not tracked by this daemon never enter the wait path.

## AI Disclosure

**Prompt / approach:** The user asked to replace the fixed five-second retry with an explicit old-execution release handoff. Codex traced the env-root lock lifecycle, introduced an in-process one-shot release registry with canonical filesystem keys, and verified the handoff under race detection and daemon package tests.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -574,6 +574,15 @@ type Daemon struct {
 	activeEnvRoots     map[string]int  // env root path -> reference count (handles reuse paths marked twice)
 	deletingEnvRoots   map[string]bool // env roots reserved by GC; new tasks wait until the mutation finishes
 
+	// envRootClaims pairs every environment lock held by this daemon with a
+	// one-shot release signal. A replacement chat turn can therefore wait for
+	// the exact execution it superseded instead of polling the filesystem for
+	// an arbitrary amount of time. The mutex also closes the gap between taking
+	// an OS lock and publishing its signal (and between observing busy and
+	// subscribing to it).
+	envRootClaimsMu sync.Mutex
+	envRootClaims   map[string]*envRootClaimHandoff
+
 	activeStoresMu   sync.Mutex
 	activeStoresCond *sync.Cond      // signalled when an in-flight store deletion finishes, so a blocked markActive can proceed
 	activeStores     map[string]int  // persistent store path (per-conversation Codex sessions, per-agent Hermes memories) -> live-task refcount; guards the store from GC mid-task (MUL-4424)
@@ -645,6 +654,7 @@ func New(cfg Config, logger *slog.Logger) *Daemon {
 		wsHBLastAck:               make(map[string]time.Time),
 		activeEnvRoots:            make(map[string]int),
 		deletingEnvRoots:          make(map[string]bool),
+		envRootClaims:             make(map[string]*envRootClaimHandoff),
 		activeStores:              make(map[string]int),
 		deletingStores:            make(map[string]bool),
 		localPathLocks:            NewLocalPathLocker(),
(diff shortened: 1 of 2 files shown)


multica-eve:
Thanks for tracing this race and for the thorough regression coverage. The reported issue is real, but this PR has been superseded by #7818, which is already merged into `main`.

#7818 preserves session continuity by waiting for the existing environment lock with a bounded 15-second, cancellable budget, then retaining the existing fresh-environment fallback. Since this PR now duplicates that behavior through a separate in-process handoff registry and conflicts with current `main`, we are closing it rather than asking for a rebase.

For the stacked Codex-specific follow-up in #7735, please rebase it onto current `main` and drop this environment-handoff commit before continuing that review.

Thank you again for the investigation and contribution.

PR #7735, opened 2026-08-29
MUL-6821: fix(codex): interrupt active turn before task handoff

mameikagou:
## What does this PR do?

Makes Multica stop an active Codex turn through the native app-server protocol before handing the conversation to a replacement task.

Today the Codex subprocess is created with the task context. Cancelling the task therefore triggers `exec.Cmd.Cancel`, which sends SIGKILL to the whole Codex process group before the backend can tell app-server to interrupt its active turn. The replacement has to recover whatever state survived that hard teardown.

This change separates the task deadline from the app-server process lifetime. On cancellation or timeout Multica now sends `turn/interrupt` with the current `threadId` and `turnId`, waits for the terminal turn notification within a bounded two-second window, and only then runs the existing graceful-close / process-group-kill cleanup. The replacement can resume the same persisted Codex thread and start the appended user message as the next turn.

This is intentionally Codex-only. Other providers keep their existing cancellation behavior.

## Relationship to #7818

This PR is rebased onto current `main`, which includes #7818. That merged change handles the lower-level workdir race by waiting for the previous execution to release its environment lock.

This PR covers the Codex protocol boundary: it interrupts the active Codex turn before the app-server process is closed. It no longer contains the separate environment-handoff registry from #7723.

## Type of Change

- [x] Refactor / code improvement
- [ ] Documentation update
- [x] Tests

## Changes Made

- Split Codex's task context from its subprocess cleanup context so task cancellation cannot SIGKILL app-server before protocol shutdown.
- Capture the active turn ID from both `turn/start` responses and `turn/started` notifications.
- Send native `turn/interrupt(threadId, turnId)` on cancellation and watchdog timeout.
- Wait briefly for `turn/completed` so interrupted state is durable before closing stdio.
- Keep the existing bounded graceful shutdown and process-group SIGKILL fallback for old, crashed, or unresponsive app-servers.
- Add an end-to-end fake app-server regression that fails unless the interrupt RPC arrives before process shutdown with the exact active thread and turn IDs.

## How to Test

1. `cd server && go test ./pkg/agent -run 'TestCodex' -count=1`
2. `cd server && go test -race ./pkg/agent -run 'TestCodexExecute(CancelsWithNativeTurnInterruptBeforeProcessShutdown|TimeoutWinsOverProcessExitDuringActiveTurn|CleansUpWhenScannerOverflowsOnResume)' -count=1`
3. `cd server && go vet ./pkg/agent`

All three commands pass on the rebased head.

## Risk and fallback

Keeping app-server alive after task cancellation is bounded by the two-second interrupt context plus the existing bounded cleanup phases. If the active turn ID is unavailable, the method is unsupported, the process has already exited, or the RPC times out, Multica logs the reason and falls back to the existing process shutdown path. Non-Codex runtimes are untouched.

## AI Disclosure

**Prompt / approach:** The user asked for native Codex stop-and-continue semantics instead of killing a task and cold-starting a conversation. Codex traced task cancellation through the daemon and app-server backend, found that `CommandContext` killed the process group before any protocol interrupt was possible, implemented a bounded native turn handoff, and verified request ordering under race detection.

### DIFF
--- a/server/pkg/agent/codex.go
+++ b/server/pkg/agent/codex.go
@@ -66,6 +66,11 @@ const (
 	// their default separate without slowing failures for lightweight RPCs.
 	defaultCodexThreadHandshakeTimeout = 60 * time.Second
 	codexVersionDiagnosticTimeout      = 2 * time.Second
+	// codexTurnInterruptTimeout bounds the native app-server handoff when a
+	// Multica task is cancelled or times out. The process deliberately outlives
+	// the task context for this brief window so Codex can persist an interrupted
+	// turn before the replacement resumes the same thread.
+	codexTurnInterruptTimeout = 2 * time.Second
 	// codexGracefulShutdownTimeout bounds how long the lifecycle goroutine
 	// waits for codex to exit on its own after stdin is closed, before forcing
 	// a context-cancel kill. A clean exit lets codex run its shutdown path and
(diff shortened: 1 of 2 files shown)


mameikagou:
@multica-eve I’ve updated this PR following your guidance on #7723:

- rebased onto current `main` (`61ea48fd2`), including the merged #7818 workdir-lock fix;
- removed the #7723 environment-handoff commit and its in-process registry entirely;
- kept this PR Codex-only: it sends native `turn/interrupt` for the active thread/turn before bounded app-server shutdown;
- preserved the newer Codex thread handshake timeout changes already on `main` while resolving the rebase conflicts;
- reduced the PR to one commit touching only `server/pkg/agent/codex.go` and `codex_test.go`.

New head: `04cf5569d2d3b1e39c0b09f051379d7951b600da`.

Local verification passed:

- `go test ./pkg/agent -run 'TestCodex' -count=1`
- targeted `go test -race` cancellation/timeout/cleanup regressions
- `go vet ./pkg/agent`

The full GitHub CI is also green, and the PR is now cleanly mergeable. When you have time, could you please take another look?

PR #7742, opened 2026-08-29
feat(daemon): add platform context modes

mameikagou:
## What does this PR do?

Adds host-selectable `full`, `minimal`, and `off` modes for Multica-authored runtime context while keeping the historical `full` behavior as the default.

- `full`: preserves the existing runtime brief, workflow guidance, and task-bound skills.
- `minimal`: keeps only runtime facts plus a neutral catalog of task-bound skills. Server-provided skills remain installed and discoverable, but Multica does not turn their presence, descriptions, or broad topical overlap into activation instructions.
- `off`: removes the managed runtime brief and server-provided built-in skills. User-installed and workspace-assigned skills remain available.

Operators can select a mode with any of:

```bash
multica config set platform_context_mode minimal
multica daemon restart --platform-context minimal
MULTICA_PLATFORM_CONTEXT=minimal multica daemon restart
```

`--no-platform-context` remains an alias for `off`, and environment values `true` / `false` remain aliases for `full` / `off`.

### Thinking path

1. The original opt-out removed both Multica's workflow prompt and all server-provided skills.
2. Gateway-oriented users still need the runtime boundary and capability discovery, but do not want Multica to prescribe their workflow.
3. A three-state host policy separates those concerns without changing server APIs or the default experience.
4. The daemon already owns runtime-brief injection, reused-workdir cleanup, skill hydration, and task CLI exposure, so it can apply the mode locally and atomically.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Tests

## Changes Made

- Added `platform_context_mode`, `--platform-context`, and string-valued `MULTICA_PLATFORM_CONTEXT` with flag > env > profile-config precedence.
- Added a declarative-only minimal brief containing runtime facts and a compact skill catalog with conservative activation guidance.
- Kept server-provided bundles available in `minimal`; only `off` filters them before resolution.
- Kept the authenticated `multica` command available in every mode. If a local daemon build has a custom filename, the daemon now creates a task-private canonical alias instead of requiring a machine-wide symlink.
- Preserved marker-aware cleanup so switching modes on a reused workdir replaces or removes only the Multica-managed block.
- Kept `--no-platform-context` and boolean environment values as compatibility aliases.
- Updated CLI, English/Chinese/Japanese/Korean environment docs, and the built-in runtime troubleshooting source map.

## How to Test

```bash
cd server
go test ./internal/cli ./internal/daemon/execenv ./internal/daemon ./cmd/multica
go vet ./internal/cli ./internal/daemon/execenv ./internal/daemon ./cmd/multica
```

Manual checks:

1. Start with `platform_context_mode=full` and confirm the historical managed brief remains unchanged.
2. Start with `minimal` and confirm task-bound skills remain installed while the managed block contains no Workflow or Always Use CLI section.
3. Rename a local daemon binary, start a task, and confirm `multica version --output json` resolves inside the task environment.
4. Start with `off` and confirm the managed block and server-provided skills are removed while repository-authored instructions and user/workspace skills remain intact.

## Risks

- `minimal` intentionally removes Multica workflow instructions, so users selecting it take responsibility for provider-native workflow behavior.
- Skill descriptions are rendered as single-line, length-bounded text in the minimal catalog; skill files remain the source of truth when a skill is actually selected.
- Renamed local binaries receive an alias only inside each task's private temporary directory; official binaries keep the existing direct PATH behavior.
- The default stays `full`, so existing installations do not change behavior unless an operator opts in.

## Checklist

- [x] Local tests pass in a clean environment
- [x] Static checks pass
- [x] Documentation is updated
- [x] Default behavior is backward compatible
- [ ] Reviewer comments will be addressed before merge

## AI Disclosure

**AI tool used:** Codex

**Approach:** Traced the daemon's skill hydration, runtime-config injection, and task PATH assembly; split the original boolean opt-out into explicit modes; kept platform capabilities available without activation pressure; added mode-transition and renamed-binary tests; and verified the affected Go packages in an environment without task-local Multica variables.


mameikagou:
Merged current main (5f0e0f229) and resolved the conflicts in 32eb9da71.

This preserves platform-context modes alongside the current daemon resume/config behavior, retains the CLI flag-forwarding coverage, and keeps the localized environment-variable docs aligned with main.

Validation passed:
- env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/cli ./internal/daemon/execenv ./internal/daemon ./cmd/multica -count=1
- go vet ./internal/cli ./internal/daemon/execenv ./internal/daemon ./cmd/multica
- git diff --check

The PR is updated and conflict-free. @multica-eve, could you please review when convenient?

PR #7744, opened 2026-08-29
fix(redact): mask custom home directory paths

sooodasama:
## What does this PR do?

Multica already intends to redact the local home directory from agent output, but the current implementation only masks it when os.UserHomeDir() contains os/user.Current().Username as an exact, case-sensitive substring.

That assumption does not hold when Windows returns a differently shaped account name or when a container overrides HOME without changing the process account. In those environments, the full home path remains visible.

Thinking path:

- Keep the existing redaction boundary in Text rather than adding a second caller-specific scrubber.
- Preserve the current exact username replacement when it succeeds.
- Fall back to masking the final home-directory component when the account name and resolved home path differ.
- Cover matching, overridden Unix, and Windows-shaped home paths with a table-driven regression test.

## Related Issue

No public issue exists. This is a small, self-contained bug reproduced by the existing redaction test on Windows.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- server/pkg/redact/redact.go: mask resolved home paths even when the process username is absent from the path.
- server/pkg/redact/redact_test.go: cover matching usernames, overridden HOME values, and Windows paths.

## How to Test

1. On upstream/main before this patch, run go test ./pkg/redact -count=1 on Windows. TestRedactHomeDirectory fails because the home path remains unchanged.
2. On this branch, run go test ./pkg/redact -count=1. The package passes.
3. Run go vet ./pkg/redact and git diff --check upstream/main...HEAD.

Additional validation:

- go test ./pkg/redact -count=1 passed.
- go vet ./pkg/redact passed.
- git diff --check upstream/main...HEAD passed.
- go test ./pkg/... -count=1 was also attempted. Unrelated server/pkg/agent failures involving Windows process behavior and missing fake CLI executables reproduced on an unmodified upstream worktree. pkg/redact failed only before this patch and passed after it.
- make check-worktree was not run because GNU Make is unavailable in this Windows environment.

Risk is low: the change only affects how the already-resolved local home path is replaced in redacted text. It does not change APIs, persistence, or UI behavior.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs
- [x] If this PR touches Chinese product copy, I checked it against the Chinese conventions

The conditional UI, documentation, runtime, and Chinese-copy checklist items are not applicable to this backend-only fix.

## AI Disclosure

**Prompt / approach:**

Codex was used to audit an existing downstream fix, reproduce the failure on the latest upstream main before modification, manually port only the minimal Go helper and regression test, review the complete diff for downstream-specific content, and run targeted tests plus an unmodified-upstream baseline comparison. I reviewed and authorized the resulting contribution.

## Screenshots (optional)

Not applicable; there is no visual change.

### DIFF
--- a/server/pkg/redact/redact.go
+++ b/server/pkg/redact/redact.go
@@ -172,10 +172,31 @@ func Text(s string) string {
 	}
 
 	// Redact home directory paths (e.g. /Users/john/ → /Users/****/).
-	if homeDir != "" && username != "" {
-		masked := strings.Replace(homeDir, username, "****", 1)
+	if homeDir != "" {
+		masked := maskHomeDirectory(homeDir, username)
 		s = strings.ReplaceAll(s, homeDir, masked)
 	}
 
 	return s
+}
+
+func maskHomeDirectory(home, currentUsername string) string {
+	if currentUsername != "" {
+		masked := strings.Replace(home, currentUsername, "****", 1)
+		if masked != home {
+			return masked
+		}
+	}
+
+	// A container may override HOME without changing the process account.
+	// Mask the final path component so that setup remains privacy-safe.
+	trimmed := strings.TrimRight(home, `/\`)
+	if trimmed == "" {
+		return home
+	}
+	separator := strings.LastIndexAny(trimmed, `/\`)
+	if separator < 0 {
+		return "****" + home[len(trimmed):]
+	}
+	return trimmed[:separator+1] + "****" + home[len(trimmed):]
 }
--- a/server/pkg/redact/redact_test.go
+++ b/server/pkg/redact/redact_test.go
@@ -212,6 +212,27 @@ func TestRedactHomeDirectory(t *testing.T) {
 	}
 }
 
+func TestMaskHomeDirectoryWithOverriddenHome(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name     string
+		home     string
+		username string
+		want     string
+	}{
+		{name: "matching username", home: "/home/runner", username: "runner", want: "/home/****"},
+		{name: "different process user", home: "/home/runner", username: "multica", want: "/home/****"},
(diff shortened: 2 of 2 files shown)


Bohan-J:
Thanks for the thorough writeup, and sorry for the outcome after the work you put in — the bug you describe is real and the diff is clean. We're closing this on product grounds rather than code quality, so it's worth spelling out the reasoning.

**Home-directory masking in `redact.Text` isn't a boundary, so we don't want to make it more thorough.** Anyone who can read a task transcript already sees repository paths, file contents, commands, diffs and error output. Masking one path segment out of all that protects nothing, while it does make paths non-copy-pasteable and harder to line up when debugging a run. We treat workspace members as trusted, and the visibility isolation people actually want is planned as a first-class concept rather than as output masking.

Tracing the call sites while reviewing this also showed the mechanism is inconsistent in a way that reinforces the point. `homeDir` / `username` are resolved from whichever process runs `Text`, and the main transcript fields (`Content` / `Output`) are only redacted in the server's ingest handler — so in a hosted deployment those are matched against the *server's* home directory, never the machine actually running the agent. Only tool inputs and stderr diagnostics go through the daemon-side path where the resolved home is the right one.

So instead of merging this we've removed the home-directory branch from `redact.Text` entirely: #7772. That also resolves the Windows failure you hit — `TestRedactHomeDirectory` goes away with the behaviour it was testing, and your diagnosis of *why* it failed (a `DOMAIN\user` account name isn't a substring of the home path, and CI only runs Go tests on ubuntu so `main` stayed green) is quoted in that PR.

To be explicit about what stays: secret redaction — API keys, tokens, PEM blocks, connection strings — is a real boundary and is untouched.

Thanks again, and please do keep contributing.

PR #7745, opened 2026-08-29
MUL-6827: fix(usage): include live token tail in reports

mameikagou:
## Summary

- add a `task_usage_hourly_live` view that keeps materialized historical buckets but rebuilds the boundary and recent buckets from raw `task_usage`
- switch daily, by-agent, and by-runtime usage reports to the live view
- add a regression test proving a completed task's raw usage wins over a stale materialized bucket

## Why

`task_usage_hourly` intentionally excludes the newest five minutes while its hourly rollup refreshes. Reports that read it directly can therefore show zero or stale token usage immediately after a task finishes.

The live view preserves the rollup for older data and scans at most the last 65 minutes of raw events. It replaces whole boundary/recent buckets instead of adding both sources, so it does not double-count usage.

## Validation

- `make sqlc`
- `go test ./internal/handler -run '^TestUsageReportsReadTheLiveTail$' -count=1`
- migration `441_task_usage_hourly_live` applied and tested against a temporary PostgreSQL database

### DIFF
--- a/server/internal/handler/dashboard_test.go
+++ b/server/internal/handler/dashboard_test.go
@@ -703,6 +703,98 @@ func TestRollupTaskUsageHourlyIdempotentAndWatermark(t *testing.T) {
 	}
 }
 
+// TestUsageReportsReadTheLiveTail proves a just-finished task is visible
+// before the five-minute-safe rollup watermark reaches it. A stale materialized
+// row is seeded in the same current-hour bucket as well: the live view must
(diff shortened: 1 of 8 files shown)


mameikagou:
Closing because this addresses report freshness after usage is already persisted, while the actual bug is local Codex usage being lost when an in-flight execution is cancelled/replaced. I will follow up with a narrowly scoped fix in the Codex cancellation/usage reporting path, without the database view or report-wide query changes.

PR #7748, opened 2026-08-29
fix(chat): open first conversation on entry

mameikagou:
## What does this PR do?

Opens the first active conversation automatically when a user enters the bare Chat route, so the detail pane is useful without a second click.

## Behavior

- Waits until the chat-session query has settled before choosing a default.
- Uses the same pinned-first, then recent-activity ordering as the thread list.
- Skips archived conversations.
- Preserves explicit `?session=` deep links.
- Preserves `?agent=` new-chat intents.
- Runs only once per Chat page mount, so compact-layout users can still press Back and remain on the thread list.
- Keeps the selected session URL-addressable through the existing store-to-URL synchronization.

## Validation

- `packages/views`: ChatPage tests (14/14)
- `packages/views`: TypeScript check
- ESLint on all changed files
- `packages/views`: 4,859/4,862 tests passed in the host timezone; the remaining three unchanged Billing tests assert UTC calendar dates and pass with `TZ=UTC` (50/50)
- Desktop production build (`electron-vite build`)

## AI Disclosure

**AI tool used:** Codex

**Approach:** Traced the Chat page's URL/store reconciliation and responsive list-detail behavior, added a one-shot default-selection effect after session loading, and covered loading, archived rows, ordering, and both deep-link forms with page-level regressions.

### DIFF
--- a/packages/views/chat/chat-page.test.tsx
+++ b/packages/views/chat/chat-page.test.tsx
@@ -42,7 +42,17 @@ vi.mock("./components/chat-thread-list", () => ({
   ),
 }));
 vi.mock("./components/chat-session-header", () => ({
-  ChatSessionHeader: () => <div>chat-session-header</div>,
+  ChatSessionHeader: ({
+    session,
+    onArchive,
+  }: {
+    session: ChatSession;
+    onArchive: (session: ChatSession) => void;
+  }) => (
+    <button type="button" onClick={() => onArchive(session)}>
+      archive-current
+    </button>
+  ),
 }));
 vi.mock("./components/chat-empty-state", () => ({
   EmptyState: () => <div>chat-empty-state</div>,
@@ -101,7 +111,6 @@ const agentsSettledRef = vi.hoisted(() => ({ current: true }));
 const sessionsRef = vi.hoisted(() => ({ current: [] as ChatSession[] }));
 const sessionsLoadedRef = vi.hoisted(() => ({ current: true }));
 const mockStartNewChat = vi.hoisted(() => vi.fn());
-const mockHandleSelectSession = vi.hoisted(() => vi.fn());
 const mockToastError = vi.hoisted(() => vi.fn());
 const mockSetActiveSession = vi.hoisted(() =>
   vi.fn((id: string | null) => {
@@ -109,6 +118,19 @@ const mockSetActiveSession = vi.hoisted(() =>
     storeListeners.forEach((l) => l());
   }),
 );
+const mockHandleSelectSession = vi.hoisted(() =>
+  vi.fn((session: ChatSession) => mockSetActiveSession(session.id)),
+);
+const mockAdvanceSelectionAfterArchive = vi.hoisted(() =>
+  vi.fn((archived: ChatSession) => {
+    const next = sessionsRef.current.find(
+      (candidate) =>
+        candidate.status === "active" && candidate.id !== archived.id,
+    );
+    mockSetActiveSession(next?.id ?? null);
+  }),
+);
+const mockArchiveSession = vi.hoisted(() => vi.fn());
 const subscribeToStore = vi.hoisted(() => (cb: () => void) => {
   storeListeners.add(cb);
   return () => storeListeners.delete(cb);
@@ -142,7 +164,10 @@ vi.mock("./components/use-chat-controller", async () => {
         () => storeRef.current,
       ).activeSessionId,
       selectedAgentId: null,
-      currentSession: null,
+      currentSession:
+        sessionsRef.current.find(
+          (session) => session.id === storeRef.current.activeSessionId,
+        ) ?? null,
       isSessionArchived: false,
       isAgentArchived: false,
       isAgentRuntimeBound: true,
@@ -167,8 +192,8 @@ vi.mock("./components/use-chat-controller", async () => {
       handleNewChat: vi.fn(),
       handleStartNewChat: mockStartNewChat,
       handleSelectSession: mockHandleSelectSession,
-      advanceSelectionAfterArchive: vi.fn(),
-      archiveSession: vi.fn(),
+      advanceSelectionAfterArchive: mockAdvanceSelectionAfterArchive,
+      archiveSession: mockArchiveSession,
       setActiveSession: mockSetActiveSession,
       setSelectedAgentId: vi.fn(),
     }),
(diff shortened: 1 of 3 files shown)


Bohan-J (changes requested):
Thanks for this — the direction is sound, and reusing `sortChatSessions` (pinned first, then recent activity, archived filtered out) so "first row" means the same thing here as it does in `ChatThreadList` is exactly right.

I verified the behaviour against `38bb0e31` and hit two state-coordination problems that I think need to be fixed before this lands. Both reproduce with the existing test harness once two fidelity gaps in it are closed (details below).

## 1. Strict Mode replays the mount and leaves nothing selected

`apps/web` is App Router, so React Strict Mode is on by default under `pnpm dev:web`. When the sessions query is already warm — navigate away from Chat and back within the same session, so `sessionsLoaded` is `true` on the very first render — the mount goes:

```
pass 1  URL→store (:110)   urlSession=null  live=null      → no-op
pass 1  default   (:135)   handled=false                   → handled=true, selectSession("recent")
pass 2  URL→store (:110)   urlSession=null  live="recent"  → setActiveSession(null)    ← undoes it
pass 2  default   (:135)   handled=true                    → early return              ← can't recover
final                      activeSessionId = null
```

(that trace is printed straight off your branch with a `console.log` at the top of each effect)

The one-shot ref is set *before* the work it guards is committed, and the sibling URL→store effect is free to undo that work in between the two passes. So in dev the feature looks like it simply doesn't work: the detail pane sits on the empty state, which is the exact thing this PR exists to remove.

The file already treats Strict Mode idempotence as a maintained invariant — see the comment block at `chat-page.tsx:99-107` ("...idempotent under StrictMode's double-invoke") and the existing `?agent=` Strict Mode test. I'd suggest a structural fix rather than another guard: let one reconciliation effect own the whole "which session should be open for this URL" decision (explicit `?session=` → that id; bare and not yet consumed → first active row; bare and already consumed → null) and write the result. Replaying that effect then converges on the same answer instead of two effects racing.

## 2. Not every path that clears the session consumes the one-shot

`handleSelect`, `startNewChat` and the Back button all set `defaultSelectionHandled.current = true`, but arriving via an explicit `?session=` link does not — the effect early-returns on `urlSession` before reaching the assignment, so the ref stays `false` for the whole visit. Two paths then clear `activeSessionId`, which makes the store→URL effect strip the param, which re-fires the default-selection effect on a bare URL with `handled === false`:

**a. Archiving the open chat on a compact layout.** `handleArchive` (`:180-191`) calls `setActiveSession(null)` so the thread list comes back. Verified at a 500px viewport, entering at `?session=deep` with a second `other` session in the list:

```
mount           store="deep", handled=false
archive         setActiveSession(null); replace("/acme/chat")
bare URL lands  default effect fires with handled=false → selectSession("other")
```

The user is dropped full-screen into an unrelated conversation instead of back to the list — the outcome the comment at `:175-179` explicitly says it wants to avoid.

**b. Picking a project.** `changeProjectContext` (`:203-210`) → `handleProjectChange` → the `startFreshChat` plan → `setActiveSession(null)`. Same sequence: the bare URL re-triggers the default selection, an old conversation opens, and the `setComposingNew(true)` that was just set gets reverted by the effect at `:90`. The user asked for a clean project-scoped chat and lands in an unrelated thread. Verified the same way.

Having the explicit-intent branch *consume* the one-shot rather than just returning closes both, as long as the `!sessionsLoaded` branch keeps not consuming it. If you take the single-reconciliation-effect shape from (1), this falls out for free.

## Why the new tests don't catch either

Two gaps in the harness:

- `mockHandleSelectSession` is a bare `vi.fn()`, unlike `mockSetActiveSession`, which writes `storeRef` and notifies subscribers. The real `handleSelectSession` calls `setActiveSession(session.id)`, so with the current mock the store never moves and "was it undone?" can't be asked. The comment above `storeRef` explains why the reactive shape matters — this path just isn't wired to it.
- All four new tests assert that `mockHandleSelectSession` *was called*, and none pass `{ strict: true }` — the option `renderPage` already supports and the `?agent=` test already uses.

To reproduce, make `mockHandleSelectSession` write `storeRef` the way `mockSetActiveSession` does, then:

- **(1)** `sessionsRef.current = [session("recent", ...)]; renderPage("", { strict: true })` → expect `storeRef.current.activeSessionId === "recent"`; it is `null`. The same assertion without `strict` passes, which isolates it to the replay.
- **(2a)** render at `session=deep` with a second `other` session and `layout.width = 500`, fire the session header's `onArchive`, then re-render with a bare search string → `mockHandleSelectSession` is called with `other`.

One thing that cost me a while: `NavigationProvider` memoises on adapter identity, so a URL change in a test has to hand down a **new** adapter object — mutating `searchParams` on the existing one is invisible to the page.

## One product question, not blocking

The effect has no `isCompact` guard, so this also changes phone/foldable behaviour. Those layouts previously landed on the full thread list, which has no empty detail pane to fix; after this change they skip the list and open a conversation, and reaching a different one costs a Back press. Was that intended, or is the target just the desktop two-pane case? Either is defensible — worth being explicit about it.

mameikagou:
Fixed both state-coordination blockers in 5ec6653de.

- URL → store, store → URL, and the one-shot default selection now live in one reconciliation effect. It reads the live Zustand state, so Strict Mode replay converges on the same selected session instead of the old URL effect clearing it.
- An explicit `?session=` is consumed immediately. If archive/project navigation clears or changes the active session while `replace()` is still committing, the consumed URL intent is not re-applied; the URL follows the new store state and the default one-shot cannot open an unrelated chat.
- `!sessionsLoaded` still does not consume the one-shot.
- `mockHandleSelectSession` now updates `storeRef` and notifies subscribers like the real controller. URL-change tests create a new navigation adapter identity.
- Added regressions for a warm bare Chat mount under `StrictMode` and for archiving an explicit compact deep link, committing the bare URL, and remaining on the list without selecting the other session.

I kept the PR's existing compact behavior intentionally: bare Chat entry auto-opens the first row on compact too. The product question remains separable from these correctness fixes.

Validation on this head:

- `vitest run chat/chat-page.test.tsx` — 16/16 passed
- `tsc --noEmit`
- ESLint on `chat-page.tsx` and `chat-page.test.tsx`
- `git diff --check`

@Bohan-J please re-review the updated reconciliation path.

PR #7752, opened 2026-08-29
MUL-6833 fix(web): localize auth callback and error pages

sooodasama:
## What does this PR do?

Localize the web OAuth callback, 404 page, root error boundary, and shared section error boundary using the existing English, Simplified Chinese, Japanese, and Korean resources. The root error boundary uses cookie/browser locale detection because it replaces the normal provider.

Google login failures retain their meaning while being localized. The server attaches stable codes to actionable 4xx failures; the callback translates known codes, preserves unknown or older uncoded 4xx messages, and renders a generic localized message for internal/provider failures.

Thinking path: trace each surface to its localization boundary, then trace each actionable OAuth code back to the actual provider response or signup policy that establishes its meaning. Tests distinguish provider failures from user-actionable failures instead of assuming every failed token exchange is an invalid grant or every userinfo response is successful.

## Related Issue

N/A — follows the review of this PR.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

Key implementation files: `server/internal/handler/auth.go`, `apps/web/app/auth/callback/page.tsx`, `apps/web/app/auth/callback/callback-error.ts`, `apps/web/platform/user-locale-sync.ts`, `packages/core/platform/core-provider.tsx`, and the existing error pages and locale resources. Regression tests live alongside their implementation and in `e2e/auth-callback-locale.spec.ts`.

- Localize callback loading, denial, failure, and Desktop handoff copy. Check OAuth error parameters before requiring a code and translate semantic error state at render time, keeping translation functions out of the single-use code-exchange effect.
- Return stable `account_disabled`, `signup_prohibited`, `email_not_allowed`, `google_account_no_email`, and `oauth_code_invalid` codes while retaining the existing fallback `error` field.
- Restrict `oauth_code_invalid` to a successfully decoded HTTP 400 `invalid_grant`. Other token failures return code-less 502 responses. Reject missing access tokens before fetching userinfo.
- Check userinfo status before parsing a profile. Reject non-200, malformed, or null responses as provider failures; only a successful profile without a usable email returns `google_account_no_email`.
- Describe a missing email as information Google did not provide for this sign-in, rather than asserting that the account has no email. Keep the stable code, and update all four translations and the server fallback together.
- Keep unknown `SignupError` values as actionable 403 responses. Correct an allowlist miss to `ErrEmailNotAllowed` when general signup is enabled. Because this helper is shared, this also corrects the message on the email-code signup path; registration eligibility is unchanged.
- Localize known codes only on 4xx responses, preserve older/unknown actionable 4xx messages, and keep 5xx/transport failures generic even if an unexpected known code is attached.
- Put 404 copy in `common.not_found`; preserve nonempty `ErrorBoundary` messages while localizing titles, recovery actions, and empty-message fallback.
- Share the root HTML language map, resolve emergency locale from cookie then browser languages, and cover actual root hydration.
- Pause automatic account-locale synchronization while Web is on the OAuth callback, then resume it after navigation. A different account language must not reload a URL containing an already consumed code. The optional CoreProvider flag defaults to enabled for other callers and preserves the provider tree.

## How to Test

1. Switch between English, Simplified Chinese, Japanese, and Korean. Check callback failures, 404, root errors, and section errors.
2. Exercise provider denial without a code and known/legacy/unknown actionable 4xx failures. Confirm only genuine actionable failures show a specific localized cause.
3. Simulate token `invalid_client`, `redirect_uri_mismatch`, unknown errors, rate limits, 5xx, malformed payloads, and missing access tokens. Simulate userinfo 401/403/429/5xx, malformed/null payloads, read errors, and transport failures. Confirm code-less 502 responses and generic localized UI.
4. Confirm a successful profile without an email still gets `google_account_no_email` with copy describing the missing response field. Switch locale while that error is displayed and confirm the copy changes without repeating the code exchange. Confirm an existing user can still sign in when new registrations are restricted.
5. With no locale cookie and a Japanese browser preference, hydrate the root error page and check Japanese copy and `html lang=ja-JP`.
6. Run `pnpm exec playwright test e2e/auth-callback-locale.spec.ts` against a running Web server. The test holds workspace lookup open, checks that the callback has not reloaded or exchanged its code twice, then verifies onboarding in the account language. Both a different-language case and a same-language control are included.

Validation began on upstream baseline `9d09902df`. After rebasing this revision onto `d4a712abf`, the 18 targeted Go tests, 47 Web tests, 15 Core tests, 160 locale-parity tests, affected typechecks, lint, Go vet/format checks, production Web build, and both permanent production-browser regressions were rerun successfully. Earlier live Google/native checks and Next dev regressions retain their original scope; they were not rerun after this rebase.

- `go test ./internal/handler -run '^TestGoogleLogin|^TestSignupGating|^TestFindOrCreateUserGating|^TestEmailCodeAllowlistErrors|^TestSendCode|^TestVerifyCode' -count=1`: 18 top-level tests passed against an isolated PostgreSQL 17 database, no skips. Successful login is checked through real authentication middleware and CSRF validation; both email-code endpoints are checked for the shared allowlist error and absence of a user/session.
- Five callback/error-page/platform test files: 47 tests passed. Related Core i18n/auth-initializer tests: 15 passed. Locale parity: 160 passed. A temporary UI harness passed two section-boundary diagnostic/recovery checks.
- Core, Web, Desktop, Views, and UI typechecks passed. Web lint completed with zero errors and its existing login-page dependency warning. Changed Core files passed ESLint.
- `go vet ./internal/handler`, changed-file `gofmt`, and `git diff --check` passed.
- The final production Web build passed, with existing CSS highlight-selector warnings. Eight real-browser probes passed using intercepted API responses and the real API client/store/providers: missing code, denial, specific/legacy/provider failures, Desktop handoff, and Web success with both matching and different account languages.
- The different-account-language browser regression was reproduced on unmodified upstream and failed against this revision before the locale-sync fix. Both permanent browser cases now pass against the fixed production build and Next dev; they require exactly one exchange and callback document, and verify the final HTML language and cookie. The second exchange deliberately fails in the test network handler so replay cannot appear successful.
- Live Google OAuth against the isolated local backend passed without API mocks: cancelling consent returned the localized denial page with no exchange request recorded in the backend logs; retrying signed in with one successful `/auth/google` request recorded; refreshing the destination restored the session. With the isolated local user's language temporarily set to Japanese, signing in from the Chinese login page also completed, with one successful `/auth/google` request recorded and a Japanese destination. The original local user language was restored afterward. This live-provider check complements the deterministic pending-workspace regression; it does not replace its controlled timing assertions.
- Additional checks against a real isolated backend passed for Web email-code login, session restoration after reload, logout, and the native CLI callback receiver. The CLI obtained a server-issued PAT, persisted its isolated profile through its normal login code, and fetched the same user. This receiver check does not include browser UI or Google authorization.
- A separate fresh live Google-to-CLI authentication check also passed. The browser used the native CLI's actual login URL, completed a fresh Google sign-in, and reached its loopback success page. The CLI obtained and saved its own PAT; its native `user profile get` command returned the expected Google user with exit code 0. The account had no workspace, so the separate workspace-creation wait was stopped after authentication verification; workspace setup and a normal exit of the login command are not claimed.
- A real Electron client also accepted a controlled native callback, authenticated against the local backend, and restored the same user after restart without another callback. API responses were not mocked; Windows protocol dispatch and Google authorization were not exercised.
- The earlier full-suite run found two handler failures (`TestStableIDSuffixMatchesDaemon` and `TestParseSkillArchive_RejectsUnsafeSkillMdPath`) and one Web failure (`app/type-scale.test.ts`, with 281 other tests passing). All three were reproduced on an unmodified checkout of `9d09902df` and concern Windows path handling. After the final copy adjustment and retranslation regression test, the affected checks above were rerun; the full suites were not rerun.
- The full `make check-worktree` invocation stopped in its Docker PostgreSQL preflight because the local Docker engine was unavailable; none of the five test stages ran through that invocation. Its cleanup incorrectly printed success and returned zero, so it is not counted as a passing pipeline. The focused checks above were run separately against isolated native PostgreSQL. Full Playwright E2E and the complete Google-to-Desktop flow through Windows protocol dispatch remain unverified. Provider responses in Go tests use an injected HTTP transport; deterministic OAuth browser regressions intercept API responses. The live Google Web/CLI, local email-login, and native receiver checks use the real isolated backend.

## Checklist

- [ ] I have run tests locally and they pass — focused checks pass; broader baseline failures are documented above
- [x] I have updated relevant documentation to reflect my changes — locale resources and owned type declarations are updated
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — not applicable
- [x] If this PR touches Chinese product copy, I checked it against apps/docs/content/docs/developers/conventions.zh.mdx

The change includes backend error classification, a shared signup-error message correction, and deferring Web account-locale synchronization until callback navigation finishes. Existing clients still receive an `error` string. Provider/configuration failures now correctly use 502 instead of implying an actionable 400. Destination selection, deep links, and error-boundary recovery remain the same. The locale fix prevents automatic locale-sync reloads during the callback; it does not make a consumed OAuth code reusable after a manual refresh.

## AI Disclosure

**Prompt / approach:** Trace hardcoded surfaces and provider boundaries, reproduce the review findings before fixing them, audit adjacent token/userinfo failures and all callback consumers, and use real API error objects plus an injected provider transport for regression tests. Validate against isolated PostgreSQL. Exercise the actual browser/provider/store path, reproduce the account-locale race on unmodified upstream, then prove the permanent browser regression fails before the fix and passes afterward. Reproduce unrelated broader-suite failures on the unmodified upstream baseline.

## Screenshots (optional)

Not included; no layout or styling changes. Failure copy and backend error classification are described and tested above.

### DIFF
--- a/apps/web/app/auth/callback/callback-error.test.ts
+++ b/apps/web/app/auth/callback/callback-error.test.ts
@@ -1,6 +1,7 @@
 // @vitest-environment node
 
 import { describe, expect, it } from "vitest";
+import { ApiError } from "@multica/core/api";
 import { callbackErrorFrom } from "./callback-error";
 
 describe("callbackErrorFrom", () => {
@@ -11,18 +12,60 @@ describe("callbackErrorFrom", () => {
     "google_account_no_email",
     "oauth_code_invalid",
   ] as const)("keeps the stable %s error kind for localization", (code) => {
-    expect(callbackErrorFrom(code, "English fallback")).toEqual({ kind: code });
+    const err = new ApiError("English fallback", 403, "Forbidden", { code });
+    expect(callbackErrorFrom(err)).toEqual({ kind: code });
   });
 
   it("keeps an actionable message from an older server that returned an uncoded 4xx", () => {
-    expect(callbackErrorFrom(undefined, "registration is disabled")).toEqual({
+    const err = new ApiError("registration is disabled", 403, "Forbidden");
+    expect(callbackErrorFrom(err)).toEqual({
       kind: "raw",
       text: "registration is disabled",
     });
   });
 
-  it("hides 5xx and transport details behind the localized generic failure", () => {
-    expect(callbackErrorFrom(undefined, undefined)).toEqual({
+  it("preserves unknown actionable codes as their server message", () => {
+    const err = new ApiError("new signup restriction", 403, "Forbidden", {
+      code: "future_restriction",
+    });
+    expect(callbackErrorFrom(err)).toEqual({
+      kind: "raw",
+      text: "new signup restriction",
+    });
+  });
+
+  it("localizes a known 4xx code even without a fallback message", () => {
+    const err = new ApiError("", 400, "Bad Request", { code: "oauth_code_invalid" });
+    expect(callbackErrorFrom(err)).toEqual({ kind: "oauth_code_invalid" });
+  });
+
+  it.each([undefined, null, {}, { code: 42 }, { code: "" }])(
+    "handles malformed error bodies: %j",
+    (body) => {
+      const err = new ApiError("actionable fallback", 400, "Bad Request", body);
+      expect(callbackErrorFrom(err)).toEqual({ kind: "raw", text: "actionable fallback" });
+    },
+  );
+
+  it.each([500, 502, 503])(
+    "hides %s details even when a known code is present",
+    (status) => {
+      const err = new ApiError("internal detail", status, "Server Error", {
+        code: "oauth_code_invalid",
+      });
+      expect(callbackErrorFrom(err)).toEqual({ kind: "login_failed" });
+    },
+  );
+
+  it.each([new Error("Failed to fetch"), null, undefined, { code: "oauth_code_invalid" }])(
+    "handles transport and non-API errors: %j",
+    (err) => {
+      expect(callbackErrorFrom(err)).toEqual({ kind: "login_failed" });
+    },
+  );
+
+  it("uses the localized generic failure for an empty, uncoded 4xx", () => {
+    expect(callbackErrorFrom(new ApiError("", 400, "Bad Request"))).toEqual({
       kind: "login_failed",
     });
   });
--- a/apps/web/app/auth/callback/callback-error.test.ts
+++ b/apps/web/app/auth/callback/callback-error.test.ts
@@ -1,0 +1,29 @@
+// @vitest-environment node
+
+import { describe, expect, it } from "vitest";
+import { callbackErrorFrom } from "./callback-error";
+
+describe("callbackErrorFrom", () => {
+  it.each([
+    "account_disabled",
+    "signup_prohibited",
+    "email_not_allowed",
+    "google_account_no_email",
+    "oauth_code_invalid",
+  ] as const)("keeps the stable %s error kind for localization", (code) => {
+    expect(callbackErrorFrom(code, "English fallback")).toEqual({ kind: code });
+  });
+
+  it("keeps an actionable message from an older server that returned an uncoded 4xx", () => {
+    expect(callbackErrorFrom(undefined, "registration is disabled")).toEqual({
+      kind: "raw",
+      text: "registration is disabled",
+    });
+  });
+
+  it("hides 5xx and transport details behind the localized generic failure", () => {
+    expect(callbackErrorFrom(undefined, undefined)).toEqual({
+      kind: "login_failed",
+    });
+  });
+});
--- a/apps/web/app/auth/callback/callback-error.ts
+++ b/apps/web/app/auth/callback/callback-error.ts
@@ -1,3 +1,5 @@
+import { ApiError, clientErrorMessage, errorCode } from "@multica/core/api";
+
 export type CallbackError =
   | {
       kind:
@@ -12,11 +14,14 @@ export type CallbackError =
     }
   | { kind: "raw"; text: string };
 
-export function callbackErrorFrom(
-  code: string | undefined,
-  clientMessage: string | undefined,
-): CallbackError {
-  switch (code) {
+export function callbackErrorFrom(err: unknown): CallbackError {
+  // An actionable code is meaningful only on a client error. A drifting server
+  // must not turn a provider/internal failure into a specific user diagnosis.
+  if (!(err instanceof ApiError) || err.status < 400 || err.status >= 500) {
+    return { kind: "login_failed" };
+  }
+
+  switch (errorCode(err)) {
     case "account_disabled":
       return { kind: "account_disabled" };
     case "signup_prohibited":
@@ -31,6 +36,7 @@ export function callbackErrorFrom(
       break;
   }
 
+  const clientMessage = clientErrorMessage(err);
   return clientMessage
     ? { kind: "raw", text: clientMessage }
     : { kind: "login_failed" };
--- a/apps/web/app/auth/callback/callback-error.ts
+++ b/apps/web/app/auth/callback/callback-error.ts
@@ -1,0 +1,37 @@
+export type CallbackError =
+  | {
+      kind:
+        | "missing_code"
+        | "access_denied"
+        | "login_failed"
+        | "account_disabled"
+        | "signup_prohibited"
+        | "email_not_allowed"
+        | "google_account_no_email"
+        | "oauth_code_invalid";
+    }
+  | { kind: "raw"; text: string };
+
+export function callbackErrorFrom(
+  code: string | undefined,
+  clientMessage: string | undefined,
+): CallbackError {
+  switch (code) {
+    case "account_disabled":
+      return { kind: "account_disabled" };
+    case "signup_prohibited":
+      return { kind: "signup_prohibited" };
+    case "email_not_allowed":
+      return { kind: "email_not_allowed" };
+    case "google_account_no_email":
+      return { kind: "google_account_no_email" };
+    case "oauth_code_invalid":
+      return { kind: "oauth_code_invalid" };
+    default:
+      break;
+  }
+
+  return clientMessage
+    ? { kind: "raw", text: clientMessage }
+    : { kind: "login_failed" };
+}
--- a/apps/web/app/auth/callback/page.test.tsx
+++ b/apps/web/app/auth/callback/page.test.tsx
@@ -1,22 +1,53 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { render, waitFor } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
+import type { SupportedLocale } from "@multica/core/i18n";
+import { I18nProvider } from "@multica/core/i18n/react";
 import { paths } from "@multica/core/paths";
+import { RESOURCES } from "@multica/views/locales";
 
 const {
   mockPush,
+  mockRouter,
   mockSearchParams,
   mockLoginWithGoogle,
   mockListWorkspaces,
   mockListMyInvitations,
   mockSetQueryData,
-} = vi.hoisted(() => ({
-  mockPush: vi.fn(),
-  mockSearchParams: new URLSearchParams(),
-  mockLoginWithGoogle: vi.fn(),
-  mockListWorkspaces: vi.fn(),
-  mockListMyInvitations: vi.fn(),
-  mockSetQueryData: vi.fn(),
-}));
+  mockQueryClient,
+  mockErrorCode,
+  mockClientErrorMessage,
+} = vi.hoisted(() => {
+  const mockPush = vi.fn();
+  const mockSetQueryData = vi.fn();
+  return {
+    mockPush,
+    mockRouter: { push: mockPush },
+    mockSearchParams: new URLSearchParams(),
+    mockLoginWithGoogle: vi.fn(),
+    mockListWorkspaces: vi.fn(),
+    mockListMyInvitations: vi.fn(),
+    mockSetQueryData,
+    mockQueryClient: { setQueryData: mockSetQueryData },
+    mockErrorCode: vi.fn((): string | undefined => undefined),
+    mockClientErrorMessage: vi.fn((): string | undefined => undefined),
+  };
+});
+
+vi.mock("@multica/core/logger", async () => {
+  const actual =
+    await vi.importActual<typeof import("@multica/core/logger")>(
+      "@multica/core/logger",
+    );
+  return {
+    ...actual,
+    createLogger: () => ({
+      debug: vi.fn(),
+      info: vi.fn(),
+      warn: vi.fn(),
+      error: vi.fn(),
+    }),
+  };
+});
 
 const makeUser = (
   overrides: Partial<{
@@ -36,12 +67,12 @@ const makeUser = (
 });
 
 vi.mock("next/navigation", () => ({
-  useRouter: () => ({ push: mockPush }),
+  useRouter: () => mockRouter,
   useSearchParams: () => mockSearchParams,
 }));
 
 vi.mock("@tanstack/react-query", () => ({
-  useQueryClient: () => ({ setQueryData: mockSetQueryData }),
+  useQueryClient: () => mockQueryClient,
 }));
 
 // Preserve the real sanitizeNextUrl so the "drop unsafe ?next=" behavior is
@@ -66,6 +97,8 @@ vi.mock("@multica/core/workspace/queries", () => ({
 }));
 
 vi.mock("@multica/core/api", () => ({
+  errorCode: mockErrorCode,
+  clientErrorMessage: mockClientErrorMessage,
   api: {
     listWorkspaces: mockListWorkspaces,
     listMyInvitations: mockListMyInvitations,
@@ -75,9 +108,19 @@ vi.mock("@multica/core/api", () => ({
 
 import CallbackPage from "./page";
 
+function renderCallback(locale: SupportedLocale = "en") {
+  return render(
+    <I18nProvider locale={locale} resources={RESOURCES}>
+      <CallbackPage />
+    </I18nProvider>,
+  );
+}
+
 describe("CallbackPage", () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    mockErrorCode.mockReturnValue(undefined);
+    mockClientErrorMessage.mockReturnValue(undefined);
     // Reset the source-backfill dismiss counter so a test that writes
     // it doesn't leak state into the next test (and the next test
     // doesn't inherit a cap-reached state from a previous run).
@@ -98,9 +141,65 @@ describe("CallbackPage", () => {
     mockListMyInvitations.mockResolvedValue([]);
   });
 
+  it("renders callback errors in the selected locale", async () => {
+    mockSearchParams.delete("code");
+
+    renderCallback("zh-Hans");
+
+    expect(await screen.findByText("登录失败")).toBeInTheDocument();
+    expect(screen.getByText("缺少授权码")).toBeInTheDocument();
+    expect(screen.getByText("返回登录")).toBeInTheDocument();
+  });
+
+  it("shows access denied before checking for a missing authorization code", async () => {
+    mockSearchParams.delete("code");
+    mockSearchParams.set("error", "access_denied");
+
+    renderCallback("zh-Hans");
+
+    expect(await screen.findByText("访问被拒绝")).toBeInTheDocument();
+    expect(screen.queryByText("缺少授权码")).not.toBeInTheDocument();
+    expect(mockLoginWithGoogle).not.toHaveBeenCalled();
+  });
+
+  it("renders a localized generic failure instead of a raw English error", async () => {
+    mockLoginWithGoogle.mockRejectedValue(
+      new Error("upstream authentication failed"),
+    );
+
+    renderCallback("zh-Hans");
+
+    expect(await screen.findByText("无法完成登录，请重试。")).toBeInTheDocument();
+    expect(
+      screen.queryByText("upstream authentication failed"),
+    ).not.toBeInTheDocument();
+  });
+
+  it("localizes a stable login error code", async () => {
+    mockLoginWithGoogle.mockRejectedValue(new Error("English fallback"));
+    mockErrorCode.mockReturnValue("signup_prohibited");
+    mockClientErrorMessage.mockReturnValue("English fallback");
+
+    renderCallback("zh-Hans");
+
+    expect(
+      await screen.findByText("此自托管实例已禁止用户注册。"),
+    ).toBeInTheDocument();
+    expect(screen.queryByText("English fallback")).not.toBeInTheDocument();
+  });
+
+  it("preserves an actionable uncoded 4xx message from an older server", async () => {
+    mockLoginWithGoogle.mockRejectedValue(new Error("legacy client error"));
+    mockClientErrorMessage.mockReturnValue("legacy client error");
+
+    renderCallback("zh-Hans");
+
+    expect(await screen.findByText("legacy client error")).toBeInTheDocument();
+  });
+
   it("unonboarded user honors a safe next= (e.g. /invite/{id}) so invitees aren't trapped", async () => {
     mockSearchParams.set("state", "next:/invite/abc123");
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith("/invite/abc123");
     });
@@ -110,7 +209,7 @@ describe("CallbackPage", () => {
   });
 
   it("unonboarded user with no next= and no pending invitations lands on /onboarding", async () => {
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith(paths.onboarding());
     });
@@ -127,7 +226,7 @@ describe("CallbackPage", () => {
         status: "pending",
       },
     ]);
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith(paths.invitations());
     });
@@ -153,7 +252,7 @@ describe("CallbackPage", () => {
         updated_at: "",
       },
     ]);
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith(paths.workspace("acme").issues());
     });
@@ -168,7 +267,7 @@ describe("CallbackPage", () => {
     );
     mockSearchParams.set("state", "next:https://evil.example");
 
-    render(<CallbackPage />);
+    renderCallback();
 
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalled();
@@ -182,7 +281,7 @@ describe("CallbackPage", () => {
     );
     mockSearchParams.set("state", "next:/invite/abc123");
 
-    render(<CallbackPage />);
+    renderCallback();
 
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith("/invite/abc123");
@@ -191,7 +290,7 @@ describe("CallbackPage", () => {
 
   it("falls through to /onboarding when listMyInvitations errors", async () => {
     mockListMyInvitations.mockRejectedValue(new Error("network"));
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith(paths.onboarding());
     });
@@ -216,7 +315,7 @@ describe("CallbackPage", () => {
       );
       mockGoogleLogin.mockResolvedValue({ token: "cli-jwt-token" });
 
-      render(<CallbackPage />);
+      renderCallback();
 
       await waitFor(() => {
         expect(mockGoogleLogin).toHaveBeenCalledWith(
@@ -244,7 +343,7 @@ describe("CallbackPage", () => {
     mockListWorkspaces.mockResolvedValue([]);
     mockListMyInvitations.mockResolvedValue([]);
 
-    render(<CallbackPage />);
+    renderCallback();
 
     await waitFor(() => {
       // Normal web flow: loginWithGoogle is called (not googleLogin)
@@ -276,7 +375,7 @@ describe("CallbackPage", () => {
       );
       mockGoogleLogin.mockResolvedValue({ token: "mixed-jwt" });
 
-      render(<CallbackPage />);
+      renderCallback();
 
       await waitFor(() => {
         expect(mockGoogleLogin).toHaveBeenCalled();
@@ -319,7 +418,7 @@ describe("CallbackPage", () => {
         updated_at: "",
       },
     ]);
-    render(<CallbackPage />);
+    renderCallback();
     await waitFor(() => {
       expect(mockPush).toHaveBeenCalledWith(paths.workspace("acme").issues());
     });
--- a/apps/web/app/auth/callback/page.test.tsx
+++ b/apps/web/app/auth/callback/page.test.tsx
@@ -4,6 +4,7 @@ import type { SupportedLocale } from "@multica/core/i18n";
 import { I18nProvider } from "@multica/core/i18n/react";
 import { paths } from "@multica/core/paths";
 import { RESOURCES } from "@multica/views/locales";
+import { ApiError } from "@multica/core/api";
 
 const {
   mockPush,
@@ -14,8 +15,6 @@ const {
   mockListMyInvitations,
   mockSetQueryData,
   mockQueryClient,
-  mockErrorCode,
-  mockClientErrorMessage,
 } = vi.hoisted(() => {
   const mockPush = vi.fn();
   const mockSetQueryData = vi.fn();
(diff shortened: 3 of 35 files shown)


Bohan-J (changes requested):
Thanks for this — the direction is right, and the three surfaces you picked really are hardcoded English today. Reusing the existing `web.desktop_handoff.*` keys instead of minting duplicates was the right call, and the `renderCallback()` helper is a genuine cleanup of that test file.

I walked each of the three surfaces end to end, and **only the 404 page fully delivers what the description promises**. The other two have gaps I could reproduce, so I'd like to see them addressed before merge.

## Must-fix

### 1. The `document.documentElement.lang` candidate is a dead branch — and it overwrites a correct value

`GlobalError` starts at `locale = "en"`, so its very first render commits `<html lang="en">` onto `document.documentElement`. React applies DOM mutations *before* running effects, so by the time `resolveEmergencyLocale()` reads `document.documentElement.lang`, it always reads `"en"`.

I confirmed this by hydrating the component the way Next does:

```js
document.documentElement.lang = "zh-CN";           // what the root layout rendered
await act(async () => hydrateRoot(document, <GlobalError error={err} reset={noop} />));
// => {"before":"zh-CN","after":"en","title":"Something went wrong"}
```

Two consequences:

- With no locale cookie, the candidate list is effectively `["en", ...navigator.languages]`, and `matchLocale` returns `"en"`. **`navigator.languages` can never win.** A first-time visitor on a Japanese browser — no cookie, served `ja` everywhere else via `accept-language` — still gets an English crash page. That's the case this PR is meant to fix, and those visitors are disproportionately the ones who hit crash pages.
- The new `lang={HTML_LANG[locale]}` actively *replaces* the root layout's correct `zh-CN` with `en`. Before this PR `<html>` carried no `lang` prop here, so it didn't clobber anything.

`app/global-error.test.tsx:20` ("falls back to the document language") passes only because it calls `resolveEmergencyLocale()` directly and never renders the component. As written it locks in behavior that cannot occur, which will make this logic harder to correct later.

Suggested fix: drop the `document.documentElement.lang` candidate entirely — it's redundant anyway, since the root layout's `lang` is itself derived from the same cookie + `accept-language`. Candidates become `cookie → navigator.languages`. Then replace that test with one covering "no cookie, Japanese browser".

### 2. The new `access_denied` copy is unreachable in all four languages

In `apps/web/app/auth/callback/page.tsx`, `if (!code) { setError(missing_code); return; }` runs *before* the `searchParams.get("error")` check. Google's denial redirect is `?error=access_denied&state=...` with **no `code`**, so the first branch always wins and the user sees "Missing authorization code" — a message that doesn't match what they just did.

Reproduced with a throwaway test (`?error=access_denied`, no `code`): asserting `访问被拒绝` fails, asserting `缺少授权码` passes.

The ordering bug predates this PR, but this PR adds translations for that exact branch — translating unreachable code buries it a layer deeper. The fix is a one-line reorder: move the `errorParam` check above the `!code` check, and add a test for it.

### 3. The 404 keys don't belong in the `ui` namespace

`packages/ui/types/i18next.ts` documents why that slice exists: so `packages/ui` can typecheck the selector form **standalone, without depending on `@multica/views`**. It also warns that JSON↔type drift there is *not* caught by the parity test and must be mirrored by hand.

The `ui` namespace's only consumers today are `packages/ui`'s own components (`markdown/CodeBlock.tsx`, `components/ui/sidebar.tsx`, `components/common/file-upload-button.tsx`). Putting `page_not_found` / `page_not_found_description` / `back_to_multica` — copy owned by an `apps/web` page — into `ui` makes `packages/ui` carry manual mirroring cost for keys it will never render, and breaks the boundary that file declares.

`common` is the natural home (`apps/web` already has it), or a new `errors` namespace if you'd rather group the error surfaces.

### 4. Don't put `t` in the deps of the effect that performs the OAuth code exchange

`page.tsx:162` adds `t` to the dependency array of an effect that calls `api.googleLogin(code, ...)` / `loginWithGoogle(code, ...)`. OAuth authorization codes are single-use: any re-run fails and surfaces "Login failed" to a user who actually logged in successfully.

I traced it and it isn't reachable today — language changes in this app go through a full page reload, and the provider's `locale`/`resources` identities are stable on this route, so `t` doesn't change. But this makes the correctness of a non-idempotent action depend on react-i18next's object-identity behavior rather than on an invariant we own. (Note `apps/web/app/(auth)/login/page.tsx:137` deliberately leaves `t` out and carries an `exhaustive-deps` warning — the two are now inconsistent.)

Cleaner: keep `t` out of the effect entirely by storing the error *kind* and translating at render time:

```tsx
type CallbackError =
  | { kind: "missing_code" | "access_denied" | "login_failed" }
  | { kind: "raw"; text: string };

const [error, setError] = useState<CallbackError | null>(null);
// effect: setError({ kind: "missing_code" }) — deps go back to what they were
// render: error.kind === "raw" ? error.text : t(($) => $.web.callback[error.kind])
```

This also makes the error card re-translate on language change, and gives the question below a clean place to land.

## One product question worth settling before merge

All three `.catch` handlers keep `err instanceof Error ? err.message : t(...)`. In practice `err` almost always *is* an `Error`, so the localized branch is the one that rarely runs: the card title becomes Chinese/Japanese/Korean while the sentence that actually explains the failure stays English. Users get a half-translated card, and the part they need to act on is the untranslated part.

The codebase already has a stance on this — `errorCode()` in `packages/core/api/client.ts`:

> so a caller can render its own localized sentence instead of toasting the server's English one

with 15+ call sites (`members-tab.tsx`, `billing-tab.tsx`, `inbox-page.tsx`, `issue-detail.tsx`, …).

I don't think this is strictly blocking — `/auth/google` may not attach a `code` yet, and `login/page.tsx` uses the same `err.message` shape. But swapping it for a generic localized sentence loses diagnostics, while keeping it means this surface is barely localized at all. Could you say which way you want to go rather than inheriting the older shape by default?

## Nits (non-blocking)

- **Duplicated `HTML_LANG`.** `apps/web/app/layout.tsx` already has an identical map with a comment explaining the region-suffix choice. There are now two, and since each independently satisfies `Record<SupportedLocale, string>`, editing one won't produce a type error in the other.
- **Hand-rolled cookie regex.** `createBrowserCookieLocaleAdapter().getUserChoice()` in `packages/core/i18n/browser-cookie-adapter.ts` is a character-for-character equivalent, with a `typeof document` guard. `global-error` is a client component, so `@multica/core/i18n/browser` imports fine.
- **`EMERGENCY_COPY` sits outside `packages/views/locales/`.** Adding a locale is type-guarded, so no key can go missing — but anyone revising existing translations won't see these 12 strings. Worth a pointer comment next to the namespace they belong with.
- **`resolveEmergencyLocale` is exported from a Next special file.** Moving it to a sibling module (e.g. `app/emergency-locale.ts`) keeps `global-error.tsx`'s export surface exactly what Next expects, and lets the test import a pure function instead of a client component.
- **Title casing regressed.** The card title went from `Login Failed` to `Login failed`, which no longer matches its sibling `web.desktop_handoff.failed_title` (`Sign-in Failed`). Reusing one key for both the title and the error body also means neither can be reworded independently later — consider splitting into `failed_title` + `login_failed`.
- **Related surface still English.** `packages/ui/components/common/error-boundary.tsx` (`DefaultFallback`) is the section-level error users hit far more often than a root crash — and it's exactly what the `global-error.tsx` doc comment points at. After this PR a zh-Hans user gets a Chinese full-page crash and an English panel error. It lives in `packages/ui`, which already has the `ui` namespace wired, so it's cheap to include here.

## Root cause, for context

`i18next/no-literal-string` is configured only in `packages/views/eslint.config.mjs`; `apps/web/eslint.config.mjs` has no equivalent. That's why these three surfaces drifted, and `apps/web/app/**` will drift again. Not blocking this PR — I see #7758 is adjacent work and this may already be tracked — just noting that this is a symptom-level fix.

## What I verified locally

On `98c05cee`:

| Check | Result |
| --- | --- |
| `pnpm --filter @multica/web exec vitest run app/auth/callback/page.test.tsx app/global-error.test.tsx app/not-found.test.tsx` | 3 files / 15 tests pass |
| `pnpm --filter @multica/views exec vitest run locales/parity.test.ts` | 160 tests pass |
| `pnpm --filter @multica/web --filter @multica/views --filter @multica/ui typecheck` | pass |
| `pnpm --filter @multica/web lint` | 0 errors (1 pre-existing warning in `login/page.tsx`) |

Everything you listed under "How to Test" does pass — the two issues above are in paths none of the current tests exercise. I also wrote three throwaway probes to confirm findings 1 and 2 (hydrating `GlobalError` into `document` to observe `<html lang>`, and rendering the callback with `?error=access_denied` and no `code`); they've been deleted.

sooodasama:
Thank you for walking these paths end to end and for the concrete hydration probe — it was very helpful. I addressed the four must-fix items and the relevant nits:

- GlobalError now resolves cookie → navigator.languages through the existing browser cookie adapter, starts without an en locale, shares the root HTML lang map, and has a real hydrateRoot regression for a first-time Japanese browser.
- The OAuth error parameter is checked before a missing code, with access_denied coverage.
- Callback failures are stored as semantic error kinds and translated at render time, so t is no longer a dependency of the single-use code exchange.
- The 404 copy moved to common.not_found, removing the manual packages/ui type mirror.
- The emergency helper moved out of the Next special file, failed_title and login_failed are separate, and the shared section ErrorBoundary is localized through its existing ui namespace.

On the product question: I chose localized generic user-facing failure copy and retained the original Error in browser diagnostics via the auth.callback logger. That avoids the half-translated card without losing the raw diagnostic signal.

I kept the apps/web literal-string lint follow-up out of this PR, as suggested. The updated branch passes 3 focused files / 17 tests, Web/Views/UI typechecks, all 160 locale-parity tests, and all three package lints with 0 errors (existing warnings only). New head: 4d5aee459c9bf7d6d56d1787eb5e43101ea682d4.

Thanks again for the thorough and actionable review.

Bohan-J (changes requested):
Thanks for the rework — this is a big improvement. All four must-fixes from the last round are correctly addressed, and you went past the nits:

- `resolveEmergencyLocale` now lives in its own module, reuses `createBrowserCookieLocaleAdapter`, and drops the dead `document.documentElement.lang` candidate. Rendering `lang={locale ? HTML_LANG[locale] : undefined}` also fixes the clobbering half — no attribute on first paint means the root layout's value survives.
- The new `global-error` test (SSR → `hydrateRoot` → assert the Japanese copy renders and the English doesn't) covers exactly the path that was untested before. Good test.
- `errorParam` now precedes the `!code` check, with a named regression test.
- The 404 copy moved to `common.not_found` and `ui` gained `error_boundary.*` — that split is more correct than what I suggested.
- `HTML_LANG` extracted and shared with the root layout; error state is a typed union so `t` is out of the effect deps.

Verified locally on `4d5aee45`: 17 web tests, 160 parity tests, typecheck across web/views/ui, and lint clean in web and ui (remaining warnings are pre-existing on main).

One thing has to change before this can merge.

## Must-fix: don't collapse every login failure into one generic sentence

Right now every failure other than `access_denied` and `missing_code` renders `login_failed` — "We couldn't complete your sign-in. Please try again." The original reason is passed to `authLogger.error()`, which is a `console` wrapper (`packages/core/logger.ts`), so it reaches neither the user nor any collector.

The problem isn't diagnostics, it's that **several of these failures are things the user can act on, and "please try again" is actively wrong advice for them.** From `server/internal/handler/auth.go`, `GoogleLogin` returns these 4xx/actionable reasons:

| Status | Message | Source |
| --- | --- | --- |
| 403 | `account disabled` | `auth.TemporarilyDisabledUserError` |
| 403 | `user registration is disabled on this self-hosted instance` | `ErrSignupProhibited` |
| 403 | `email address or domain not allowed on this instance` | `ErrEmailNotAllowed` |
| 400 | `Google account has no email` | empty `gUser.Email` |
| 400 | `failed to exchange code with Google` | Google rejected the code (expired / already used) |

None of these are fixed by retrying. A user whose account is disabled, or whose domain isn't allowed on a self-hosted instance, now gets an infinite "try again" loop with no way to learn why. The remaining branches (502/503/500 — token-exchange transport failures, `Google login is not configured`, `failed to create user`, `failed to generate token`) genuinely *are* internal and should stay generic.

Either of these is fine; the repo already has a convention for both.

### Option A — render 4xx messages, hide 5xx (frontend-only, small)

`clientErrorMessage()` in `packages/core/api/client.ts` exists for precisely this, and its doc comment states the rule:

> returns the server's message only when it is a CLIENT error (4xx). Handlers write those for the user … A 5xx message is internal detail (Go error chains, pgx table/constraint names, internal ids) that must never reach a toast, and a non-ApiError is a transport failure whose message ("Failed to fetch") says nothing a user can act on. Both return undefined so the caller falls back to its own localized sentence.

So keep the typed error state you built, and add one raw variant:

```tsx
type CallbackError =
  | { kind: "missing_code" | "access_denied" | "login_failed" }
  | { kind: "raw"; text: string };

.catch((err) => {
  authLogger.error("Google OAuth callback failed", err);
  const serverMessage = clientErrorMessage(err);
  setError(serverMessage ? { kind: "raw", text: serverMessage } : { kind: "login_failed" });
});
```

`t` still stays out of the effect deps, 5xx internals still never reach the UI, and the actionable reasons come through. They'll be in English for a zh/ja/ko user, which is a known and accepted trade-off for this route today — `clientErrorMessage` is used this way elsewhere in the codebase.

### Option B — stable codes, fully localized (better end state, touches the server)

`writeErrorCode` already exists next to `writeError` in `server/internal/handler/handler.go`, and its comment describes this exact goal:

> writeError plus a stable machine-readable code, so a UI can translate the failure instead of toasting the English sentence at a user whose console is in another language. The sentence stays as the fallback for anything that has not been given a translation yet.

`GoogleLogin` hasn't adopted it — every branch uses plain `writeError`. Switching the five rows above to `writeErrorCode` with codes like `account_disabled` / `signup_prohibited` / `email_not_allowed` / `google_account_no_email` / `oauth_code_invalid`, then mapping them via `errorCode()` to `auth.web.callback.*` keys, gives fully localized, specific copy. Fall back to Option A's behavior for any code you haven't translated yet.

B is the better destination. A is a legitimate stopping point for this PR if you'd rather not touch Go here — your call, but the current "one sentence for everything" is not.

## Must-fix: same rule for `ErrorBoundary`'s default fallback

`packages/ui/components/common/error-boundary.tsx` dropped the `error` prop from `DefaultFallback` and replaced `{error.message || "An unexpected error occurred."}` with a fixed localized sentence. That's a behavior change well beyond localization — it affects every default-fallback boundary in the app (issues list, issue detail, both attachment previews), and it isn't mentioned in the PR description.

Please restore the message and keep the localized string as the fallback for an empty one:

```tsx
function DefaultFallback({ error, reset }: { error: Error; reset: () => void }) {
  const { t } = useTranslation("ui");
  // …
  <p className="text-muted-foreground">
    {error.message || t(($) => $.error_boundary.description)}
  </p>
```

The title and the "Try again" button are still worth localizing — that part of the change is good.

## Nits (non-blocking)

- **The `errorParam` branch is the only failure path that doesn't log.** All three `.catch` handlers call `authLogger.error`, but `if (errorParam)` returns silently, so `?error=server_error` / `temporarily_unavailable` now vanishes from the UI *and* the console. One `authLogger.warn("OAuth error param", errorParam)` closes it.
- **`.catch((error) => …)` shadows the `error` state variable** declared just above in `CallbackContent`. It works, but reads as if the state is being inspected. `err` avoids it.
- **All three catch handlers log the same `"Google OAuth callback failed"`**, so the CLI, Desktop, and web flows are indistinguishable in the console.
- `authLogger` is console-only. `captureException` from `@multica/core/analytics` is already used in `global-error.tsx` — worth considering for the auth callback so production failures are actually collected. Fine as a follow-up.

sooodasama:
Thank you — the concrete A/B framing and the `ErrorBoundary` callout made the intended boundary very clear. I went with Option B and updated the branch on top of the latest `main`.

- `GoogleLogin` now uses `writeErrorCode` for the five actionable cases you listed: `account_disabled`, `signup_prohibited`, `email_not_allowed`, `google_account_no_email`, and `oauth_code_invalid`. The existing English `error` field remains available as the fallback. I also corrected the allowlist-miss path to return `ErrEmailNotAllowed` instead of collapsing it into `ErrSignupProhibited`.
- The callback uses `errorCode()` for the localized cases, falls back to `clientErrorMessage()` for unknown or older uncoded 4xx responses, and keeps 5xx/transport failures behind `login_failed`. I added both pure mapping coverage and page-level regressions for the localized-code and legacy-4xx paths.
- The default `ErrorBoundary` fallback again renders `error.message || localized description`, so the title and recovery action remain localized without changing the existing diagnostic behavior.
- I also picked up the three logging/readability nits: provider `errorParam` values are logged, catch parameters are named `err`, and CLI/Desktop/web failures have distinct log messages. I left `captureException` for a follow-up as suggested.

Locally, the four focused web files pass 26 tests, locale parity passes all 160 tests, Web/Views/UI typechecks pass, all three package lints have 0 errors (existing warnings only), and `go vet ./internal/handler` passes. The new handler tests compile locally; their execution was skipped by the repository `TestMain` because my local PostgreSQL service is unavailable, so I am relying on this PR's database-backed CI run for that final execution check.

New head: `d7a043ad425673f1ab60d7b14070df35cd37447f`.

Thanks again for the precise guidance here, especially around preserving actionable sign-in errors while keeping internal failures generic. I’d appreciate another look whenever you have time.

Bohan-J (changes requested):
The error-handling rework is the right shape. `callbackErrorFrom(errorCode(err), clientErrorMessage(err))` does exactly what was asked: stable codes render localized specific copy, an un-coded 4xx still shows the server's original message, and only 5xx/transport failures fall back to the generic sentence. Adopting `writeErrorCode` in `GoogleLogin` and extracting `writeGoogleLoginActionableError` is clean, and the `ErrorBoundary` default fallback is restored. The three earlier nits are all closed too — the `errorParam` branch logs now, the shadowed `error` binding is gone, and the three flows have distinct log messages.

Two problems remain, and they're the same failure mode this round was meant to fix — the UI confidently telling the user a wrong cause. Both are now worse than before the PR, because a vague English sentence has been upgraded into a specific localized claim backed by a stable machine code.

## Must-fix 1: every non-200 from the token endpoint is reported as `oauth_code_invalid`

`server/internal/handler/auth.go`:

```go
if tokenResp.StatusCode != http.StatusOK {
    slog.Error("google oauth token exchange returned error", ...)
    writeErrorCode(w, http.StatusBadRequest, googleLoginCodeInvalidOAuthCode, "failed to exchange code with Google")
    return
}
```

The response body is never inspected. Google's token endpoint also returns non-200 for `invalid_client` (wrong `GOOGLE_CLIENT_SECRET`), `redirect_uri_mismatch` (wrong `GOOGLE_REDIRECT_URI`), `unauthorized_client`, and `deleted_client` — all deployment misconfiguration, none of them anything the user did.

I ran this against your own test seam:

```
Google -> 401 {"error":"invalid_client","error_description":"The OAuth client was not found."}
API    -> 400 code="oauth_code_invalid" error="failed to exchange code with Google"
```

So on a self-hosted instance with a mistyped client secret, every user now reads *"The Google authorization code is invalid, expired, or has already been used. Sign in again."* — and signs in again, forever. Before this PR they got a vague English sentence, which at least didn't assert a cause. The operator gets no user-visible signal that the deployment is misconfigured.

Please decode the error body and map only `invalid_grant` to `oauth_code_invalid`. Everything else should keep the detailed `slog.Error` and return a code-less 502/500 so the client falls back to the generic localized sentence:

```go
var tokenErr struct {
    Error string `json:"error"`
}
_ = json.Unmarshal(tokenBody, &tokenErr)
if tokenErr.Error == "invalid_grant" {
    writeErrorCode(w, http.StatusBadRequest, googleLoginCodeInvalidOAuthCode, "failed to exchange code with Google")
    return
}
writeError(w, http.StatusBadGateway, "failed to exchange code with Google")
return
```

`TestGoogleLoginActionableErrorCodes` sends `{"error":"invalid_grant"}` for every non-200 case, so it passes without distinguishing the two — worth adding an `invalid_client` case that asserts the code is *not* `oauth_code_invalid`. Reference: https://developers.google.com/identity/protocols/oauth2/web-server

## Must-fix 2: the userinfo response status is never checked

```go
userInfoResp, err := h.googleHTTPClient().Do(userInfoReq)
if err != nil { /* 502 */ }
defer userInfoResp.Body.Close()

var gUser googleUserInfo
if err := json.NewDecoder(userInfoResp.Body).Decode(&gUser); err != nil { /* 502 */ }

if gUser.Email == "" {
    writeErrorCode(w, http.StatusBadRequest, googleLoginCodeAccountWithoutEmail, "Google account has no email")
    return
}
```

Only the transport error is handled. When Google answers 401/403/5xx, the body is a JSON error object whose fields don't match `googleUserInfo`; `encoding/json` ignores unknown fields, so the decode *succeeds* and leaves `Email` empty — and the empty-email branch fires. Same seam:

```
Google userinfo -> 401 {"error":{"code":401,"message":"Request had invalid authentication credentials.","status":"UNAUTHENTICATED"}}
API             -> 400 code="google_account_no_email" error="Google account has no email"
```

The user reads *"This Google account does not have an email address."*, which is false, and goes off to try a different Google account. The missing status check predates this PR, but it used to surface as a generic English string; now it's a confident localized diagnosis with a machine code the frontend branches on.

A status check before the decode is enough:

```go
if userInfoResp.StatusCode != http.StatusOK {
    body, _ := io.ReadAll(io.LimitReader(userInfoResp.Body, 4096))
    slog.Error("google userinfo returned error", "status", userInfoResp.StatusCode, "body", string(body))
    writeError(w, http.StatusBadGateway, "failed to fetch user info from Google")
    return
}
```

The test transport always returns 200 for `www.googleapis.com`, so no case covers this — please add one.

## Worth calling out in the PR description (not blocking)

`checkSignupAllowed` now returns `ErrEmailNotAllowed` instead of `ErrSignupProhibited` when allowlists are configured but don't match. I think that's a genuine correction — registration isn't disabled in that configuration, the address just isn't on the list — and `TestSignupGating` was properly expanded to cover signup-enabled and signup-disabled separately rather than flipped to pass. But `checkSignupAllowed` has three call sites, so this changes user-visible copy on the email-code signup path too, which is outside "localize auth callback and error pages". Please mention it in the description so it isn't a surprise.

## Nits

- `errors.Is(err, ErrSignupProhibited)` / `errors.Is(err, ErrEmailNotAllowed)` replaced `errors.As(err, &signupErr)`. Identical today since those are the only two `SignupError` values, but `errors.As` caught *any* `SignupError`; a future one would now fall through to the 500 branch instead of the 403.
- `writeGoogleLoginActionableError(w, auth.ErrTemporarilyDisabledUser)` at the `IsTemporarilyDisabledUserEmail` site discards a bool that is always `true` there. A direct `writeErrorCode` call reads more plainly.
- `googleOAuthHTTPClient` is a sensible test seam; a one-line comment noting it's test-only and nil in production would help the next reader of the `Handler` struct.

## Verified locally on `d7a043ad`

| Check | Result |
| --- | --- |
| `go test ./internal/handler -run 'TestGoogleLogin\|TestSignupGating'` | pass |
| `go vet ./internal/handler/` | clean |
| `gofmt -l ./internal/handler/` | only `issue_table_query.go`, untouched by this PR |
| web tests (`app/auth/callback`, `global-error`, `not-found`) | 4 files / 26 tests pass |
| `locales/parity.test.ts` | 160 tests pass |
| typecheck (web / views / ui) | pass |

Both findings above were reproduced through `googleOAuthHTTPClient` with throwaway tests, which I've deleted.

sooodasama:
Thanks for catching these. Updated in [`bdf6ef2ab`](https://github.com/multica-ai/multica/commit/bdf6ef2abd6dd9324e42cff52908b649714b7aef). I’ve corrected both cases so that specific localized errors are backed by the provider response:

- **Token exchange:** `oauth_code_invalid` is returned only for a successfully decoded HTTP 400 `invalid_grant` response. `invalid_client`, `redirect_uri_mismatch`, other provider errors, and malformed responses return a code-less 502, with server-side diagnostics retained.
- **Userinfo:** the HTTP status is checked before decoding the profile. Non-200 responses return a code-less 502 with bounded error-body logging; malformed or null profiles also return 502. `google_account_no_email` is limited to a successful profile without a usable email. The fallback and all four translations now say that Google did not provide an email address for this sign-in.

The callback also keeps 5xx responses generic even if they unexpectedly contain a known actionable code, while preserving the original message for unknown or uncoded 4xx responses.

I’ve addressed the three nits: an `errors.As` fallback preserves 403 responses for unknown `SignupError` values, the disabled-email branch calls `writeErrorCode` directly, and the HTTP override is documented as test-only. The shared allowlist correction also changes the email-code signup message; registration eligibility is unchanged.

Browser testing uncovered a related race: a different account language could reload the callback while workspace lookup was pending, attempting to exchange the same authorization code again. Automatic account-locale synchronization is now paused on the Web callback route and resumed after navigation. Desktop retains its existing default behavior.

Validation for this PR:

The revision is rebased onto upstream `d4a712abf`. The targeted Go/Web/Core/locale tests, affected typechecks, lint, production Web build, and both production-browser regressions were rerun on this base. The earlier Next dev, live Google, and native-client checks below retain their original scope.

- Targeted Go, Web, Core, and locale-parity tests passed. The Go tests ran against isolated PostgreSQL with no skips and include authentication-cookie and CSRF checks.
- Both permanent browser regressions passed in production and Next dev. The different-language regression failed before the fix and now verifies exactly one code exchange, successful navigation, and the final account language. Eight representative production-browser scenarios also passed.
- Live Google sign-in against the isolated local backend passed: cancelling consent returned the localized denial page with no exchange request recorded in the backend logs; retrying completed sign-in; reloading restored the session. Signing in from a Chinese page with a saved Japanese account preference also completed, with one successful `/auth/google` request recorded and a Japanese destination page.
- Additional checks against an isolated local backend passed for Web email-code login, session restoration after reload, logout, and the native CLI callback receiver. The CLI obtained its own server-issued PAT, saved its isolated profile, and successfully fetched the same user.
- A fresh live Google login through the native CLI also passed: the browser returned to the CLI's own loopback receiver, and the CLI saved a PAT and fetched the expected user through its normal profile command. Authentication was verified before stopping its separate workspace-creation wait; workspace setup was not exercised.
- The native Desktop receiver also passed against the real backend: a controlled native callback reached the renderer, and restarting the client restored the same user without another callback. This checks client authentication and persistence, not Windows protocol dispatch.
- The production Web build, affected typechecks, and lint passed; existing warnings remain.

The broader Windows suites still have three failures reproduced on unmodified upstream `9d09902df`: `TestStableIDSuffixMatchesDaemon`, `TestParseSkillArchive_RejectsUnsafeSkillMdPath`, and `app/type-scale.test.ts`. The deterministic OAuth browser regressions use controlled API responses; the live Google, email-login, and native receiver checks use the real backend. The full `make check-worktree` invocation stopped at the Docker PostgreSQL preflight, before its test stages, because the local Docker engine was unavailable; it is not counted as passing despite its cleanup reporting success. The focused checks above ran separately. The complete Google-to-Desktop flow through Windows protocol dispatch remains unverified.

Further testing reproduced a separate race already present on upstream `a6911a952`: a delayed initial `/api/me` response could overwrite a successful sign-in, including clearing the new session on a late 401. I’ve isolated the fix as a separate change, outside this PR. It prevents stale session reads and 401 responses from replacing or clearing newer authentication state, while retaining normal handling of current-session 401s. The separate fix passed 120 targeted Core tests and a production-browser regression that failed before the fix, along with the Web build, Core/Desktop typechecks, and changed-file lint. These regressions use controlled responses; live Google sign-in has not been rerun with that separate fix.

Bohan-J (approved):
Thanks for sticking with this through four rounds — the PR ended up considerably more valuable than its title suggests, and that's down to how you handled the review.

What stood out:

- You didn't just patch the items I listed. On the `ui` namespace question you picked a better split than I proposed (404 copy to `common.not_found`, `error_boundary.*` to `ui` where `packages/ui` genuinely owns it), and you localized the shared `ErrorBoundary` that I'd only mentioned as a nit.
- The `global-error` test — SSR, write to the document, hydrate, assert the copy — covers the exact path that had been invisible, rather than the easy path.
- After the OAuth-classification round you went beyond the two fixes: the empty-`access_token` guard, the nil-profile check, and especially `callbackErrorFrom` refusing to trust an error code unless the response is a 4xx `ApiError`. That last one is the right instinct — it means a future server-side drift can't turn an internal failure into a confident user-facing diagnosis.
- The provider-failure test matrix (`invalid_client`, `redirect_uri_mismatch`, `deleted_client`, rate limits, malformed bodies, userinfo non-200 including a 500 whose body carries a valid email) is thorough, and several of those cases you found yourself.

Beyond the localization itself, this ships a handful of real fixes to paths that were already broken on `main`: the unreachable `access_denied` branch, the locale-sync reload that could replay a consumed authorization code and show "Login Failed" after a successful sign-in, and `checkSignupAllowed` reporting "registration is disabled" for what is actually an allowlist miss.

Verified locally on `bdf6ef2a`: Go handler tests, `go vet` and `gofmt` clean, 47 web tests, 160 locale parity tests, typecheck across web/views/ui/core, and lint clean. CI is green.

One follow-up worth picking up separately, not blocking: `authLogger` is a console wrapper, so auth-callback failures still aren't collected anywhere in production. `captureException` from `@multica/core/analytics` is already used in `global-error.tsx` and would close that gap.

Merging. Thanks again.

PR #7753, opened 2026-08-29
MUL-6834: fix(codex): preserve token usage on cancelled turns

mameikagou:
## Problem

When a running Codex task is cancelled or replaced, `exec.CommandContext` immediately sends the configured process-group SIGKILL. Codex cannot finish its native interrupt flow or emit `turn/completed`, so the daemon receives an empty usage map even though the local turn consumed tokens.

This is the underlying collection gap that #7745 did not address. That PR has been closed.

## Fix

- Keep the app-server process alive independently of the task context during bounded cleanup.
- On cancellation, send the native `turn/interrupt` request with the active thread and turn IDs.
- Wait up to 2 seconds for `turn/completed`, allowing its final usage to enter the existing result and daemon reporting path.
- Preserve the existing process-group kill and bounded cleanup fallback when interruption stalls.

No database views, report queries, migrations, or UI behavior are changed.

## Tests

- Added an app-server integration fixture that cancels an active turn, verifies `turn/interrupt`, and asserts uncached input, cached input, and output tokens are retained.
- `go test -race ./pkg/agent -run TestCodexExecuteCancellationInterruptsTurnAndPreservesUsage -count=1`
- `go test ./internal/daemon -run 'TestHandleTask_ReportsUsage(BeforeCancel|WhenCancelledByPoll)' -count=1`


multica-eve (changes requested):
This change aims to keep already-incurred usage visible when a member cancels or replaces a running task. That behavior is desirable, but some cancellations may now take much longer, and the current evidence does not show that real usage will be preserved. Please address the blockers below before merge.

## Request changes

1. **Preserve prompt cancellation when native interruption cannot complete.**

   `server/pkg/agent/codex.go:1026` detaches the app-server process from the task context for its entire lifetime. Cancellation during initialization/thread setup, cancellation before a turn ID is observed, and `turn/interrupt` failure or timeout all reach `drainAndWait` without first calling `stopProcess` (`codex.go:1382`, `:1421`, `:1659`). Cleanup then waits the 10-second stdin-to-stdout EOF grace before forcing the process tree down, with the existing wait/reader backstops after that. Main currently kills the process tree immediately when the task context is cancelled, so a cancel/replacement can regress from prompt termination to `2s + 10s` or longer.

   Please make `interruptCodexTurn` report whether interruption completed. Missing turn IDs, request failures, and timeouts should call `stopProcess` before drain; task cancellation during the setup RPCs needs the same fast fallback. Add bounded-latency tests for cancellation during initialize, the `turn/start` / `turn/started` race, and a non-responsive interrupt.

   The intended shutdown chains should be:

   - Success: `task cancel -> turn/interrupt -> interrupt response -> turn/completed -> close stdin -> stdout EOF -> cmd.Wait`
   - Failure: `timeout/not interruptible -> stopProcess immediately -> process group/Job Object termination -> reader EOF -> cmd.Wait`

2. **Test the production usage contract rather than a synthetic field.**

   `server/pkg/agent/codex_test.go:3577` injects usage into `turn/completed.params.turn.usage`. Codex 0.144.4's v2 protocol sends the final turn state in `turn/completed`, while token usage is streamed separately through `thread/tokenUsage/updated`; Multica also has a `$CODEX_HOME` rollout `token_count` fallback. The new test exercises neither path: it does not send the real usage notification, configure `CODEX_HOME`, or write a rollout. It also skips Windows, even though this change modifies the Job Object lifecycle there.

   Please use the real wire shape (`turn/completed` without `usage`) and prove `Result.Usage` through a supported source: either handle/test `thread/tokenUsage/updated`, or write a current-thread rollout under the task's `CODEX_HOME` and verify the post-interrupt scan. Add coverage for the Windows ownership path as well. Protocol reference: https://github.com/openai/codex/blob/rust-v0.144.4/codex-rs/app-server/README.md#turn-events

3. **Justify the two-second interruption budget.**

   `server/pkg/agent/codex.go:67` uses one hard-coded two-second budget for both the interrupt response and final completion event. The PR provides no Unix/Windows or active-tool cleanup latency evidence showing that this value preserves usage without making replacement unacceptably slow. Please provide measurements and choose/configure the value from that evidence, with a near-deadline regression test proving the completion event is still consumed.

Validation on the current head passed:

- `git diff --check origin/main...origin/pr/7753`
- `go test -race ./pkg/agent -run '^TestCodexExecuteCancellationInterruptsTurnAndPreservesUsage$' -count=20`
- `go test ./internal/daemon -run 'TestHandleTask_ReportsUsage(BeforeCancel|WhenCancelledByPoll)' -count=1`

The green tests do not cover the protocol and cancellation fallback gaps above.

mameikagou:
Addressed all three requested blockers in e08d10146.

1. Cancellation keeps the native interrupt path only when it actually completes. `interruptCodexTurn` now returns success/failure; missing turn IDs, request errors, and interrupt/completion timeouts call `stopProcess()` before drain. Cancellation during `initialize` and `thread/start`/`thread/resume` also takes the immediate process-tree stop path. Added bounded regressions for initialize cancellation, the `turn/start` → `turn/started` race, and a non-responsive interrupt.

2. The usage test now uses the Codex v2 production contract: `thread/tokenUsage/updated` followed by `turn/completed` with no synthetic usage field. Multica sums `tokenUsage.last` for each accepted current-turn response (not thread-cumulative `total`, which can include resumed history), including reasoning output and cache buckets. Added a multi-response/resumed-history regression and a Windows Job Object test proving a stalled interrupt kills the owned descendant tree.

3. The interruption budget is no longer an unobservable fixed assumption. It is configurable per execution and by `MULTICA_CODEX_TURN_INTERRUPT_TIMEOUT`; logs record the configured budget and observed interrupt latency. The 2s value remains only the compatibility default. A near-deadline regression deliberately delays the response/completion by 120ms under a 250ms budget and proves the final usage notification is consumed; the timeout regression proves the process is force-stopped at the configured deadline. This instrumentation lets Unix/Windows deployments tune from real latency rather than treating fixture timing as production evidence.

Validation on this head:

- `go test -race ./pkg/agent -run '^(TestCodexExecuteCancellationInterruptsTurnAndPreservesUsage|TestCodexCancellationDuringInitializeStopsProcessImmediately|TestCodexCancellationBeforeTurnStartedStopsProcessImmediately|TestCodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline|TestCodexInterruptCompletesNearConfiguredDeadline)$' -count=20`
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/agent`
- `git diff --check`

@multica-eve please re-review the updated shutdown, protocol, and Windows ownership coverage.

multica-eve (changes requested):
This update improves cancellation cleanup and Windows coverage, but it can still record the same usage more than once and can fail unpredictably when cancellation happens exactly as a turn starts. That means members may still see incorrect budget and usage history, so this revision is not ready to merge.

## Request changes

The fast-failure shutdown path and Windows process-tree ownership are materially better on this head. The following blockers remain:

1. **Synchronize access to the active turn ID.**

   `server/pkg/agent/codex.go:2354` reads `c.turnID` from the lifecycle goroutine while `server/pkg/agent/codex.go:3199` writes it from the stdout-reader goroutine, with no lock, atomic operation, or channel happens-before edge. Repeating the PR's own race suite reproduced `WARNING: DATA RACE` in `TestCodexInterruptCompletesNearConfiguredDeadline`. This can make cancellation incorrectly take the no-turn fallback and lose final usage.

   Please protect the value with a synchronized accessor/lock, or pass the active turn ID through an already synchronized gate/channel, and keep a stable `-race` regression for cancellation overlapping `turn/started`.

2. **Make token-usage accounting idempotent for snapshots and resume replay.**

   `server/pkg/agent/codex.go:3374-3396` adds `tokenUsage.last` for every `thread/tokenUsage/updated` notification. Codex v2 models this as a `{total,last}` snapshot, can replay historical usage after resume, and includes `turnId` for attribution. The current code neither checks the active turn ID nor deduplicates repeated/replayed snapshots, so receiving the same snapshot twice charges it twice.

   Please establish a baseline for the current thread/turn and calculate a monotonic delta from `total`, or provide an equivalent idempotent design. Add regressions for an identical duplicate snapshot, a resume replay, and multiple model responses in one turn. Also remove the synthetic `cacheWriteInputTokens` field from `server/pkg/agent/codex_test.go:3648-3679`; it is not part of the upstream `TokenUsageBreakdown` schema. Protocol references: [v0.144.4 usage types](https://github.com/openai/codex/blob/rust-v0.144.4/codex-rs/app-server-protocol/src/protocol/v2/thread.rs#L1372-L1423) and [resume replay coverage](https://github.com/openai/codex/blob/rust-v0.144.4/codex-rs/app-server/tests/suite/v2/thread_resume.rs#L2044-L2097).

3. **Resolve the two-second default as a product/operations decision and document the configuration.**

   `server/pkg/agent/codex.go:63-70` and `server/internal/daemon/config.go:33` still default to two seconds. Making the value configurable and logging future latency does not establish a safe initial default; the 120ms/250ms fixture only validates test scheduling, not Unix/Windows or active-tool cleanup behavior. Please provide measured latency evidence and a selection rationale, or obtain an explicit maintainer decision for the default.

   `MULTICA_CODEX_TURN_INTERRUPT_TIMEOUT` also needs to be documented in `CLI_AND_DAEMON.md` and the applicable `apps/docs/content/docs/environment-variables*.mdx` pages so operators can discover the rollback/tuning control.

4. **Rebase onto current `main` and preserve the separate thread-handshake timeout.**

   This head is currently `CONFLICTING` / `DIRTY`. A local merge-tree check reports conflicts in `server/internal/daemon/config.go`, `server/internal/daemon/daemon.go`, `server/pkg/agent/agent.go`, and `server/pkg/agent/codex.go`. Current `main` includes the independently added Codex thread-handshake timeout (MUL-6862); please retain that behavior while resolving the conflicts and rerun validation.

Validation on `e08d10146b897a3550fa0e3a98aa677a5ff95e72`:

- The new focused tests passed once, but the submitted `go test -race ... -count=20` suite reproduced the data race above.
- `go test ./pkg/agent -count=1` passed.
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1` passed.
- `go test ./internal/daemon/execenv -run ByteIdentical -count=1` passed.
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/agent` passed.
- All 15 GitHub checks passed; the merge conflict and review blockers remain.

mameikagou:
Follow-up for the two-second default / documentation blocker (commit `3e1d07b`).

## Real app-server measurement

I added an opt-in `agentintegration` test that launches an authenticated Codex app-server and measures the exact boundary protected by `TurnInterruptTimeout`: from cancelling the task / sending `turn/interrupt` until Multica receives the authoritative `turn/completed`. It also records cancellation-to-`Session.Result` latency so process cleanup is visible separately.

Environment:

- WSL/Linux amd64
- `codex-cli 0.147.0`
- CLI-configured default model, reasoning effort `low`
- 20 measured samples per scenario, plus 2 discarded warmups per scenario
- Measurement timeout set to 10s so the proposed 2s default does not censor the observations

Scenarios and trigger points:

1. **Generating text:** wait for the first real `item/agentMessage/delta`, allow another 100ms of streaming, then cancel.
2. **Active tool:** ask Codex to execute `sleep 30`, wait for `item/started:commandExecution` / Multica's `exec_command` tool-use event, then cancel while the command is running.

| Scenario | Metric | P50 | P95 | P99 | Max |
|---|---:|---:|---:|---:|---:|
| Generating text | `turn/interrupt` → `turn/completed` | 9ms | 12ms | 15ms | 15ms |
| Generating text | cancel → `Session.Result` | 52.2ms | 57.4ms | 59.6ms | 59.6ms |
| Active tool | `turn/interrupt` → `turn/completed` | 9ms | 10ms | 13ms | 13ms |
| Active tool | cancel → `Session.Result` | 57.1ms | 66.8ms | 68.2ms | 68.2ms |

## Default selection

I recommend retaining **2s** as the compatibility default rather than tightening it to the observed tens of milliseconds:

- it is more than **133x** the worst observed protocol-completion latency;
- it is more than **29x** the worst observed end-to-end cleanup latency;
- that margin covers host contention and cross-platform scheduler/process-cleanup variance that this single WSL host cannot represent;
- it is still short enough to keep replacement bounded, and failures immediately take the existing force-stop path when the grace period expires.

This is intentionally presented as WSL/Linux evidence, not as an unmeasured Windows claim. Operators can override it with `MULTICA_CODEX_TURN_INTERRUPT_TIMEOUT`, and this commit documents that control in `CLI_AND_DAEMON.md` plus all four environment-variable documentation pages.

## Reproduction

The real-agent test remains excluded from ordinary CI and requires explicit account/quota authorization:

```bash
cd server
MULTICA_RUN_REAL_AGENT_SMOKE=1 \
MULTICA_CODEX_INTERRUPT_SAMPLES=20 \
MULTICA_CODEX_INTERRUPT_WARMUPS=2 \
go test -tags=agentintegration ./pkg/agent \
  -run TestCodexRealTurnInterruptLatency -count=1 -v
```

The sample count, warmups, model and post-first-delta settle delay are configurable through the test-only environment variables documented beside the test.

## Usage boundary observed during the probe

The real run also clarified what “preserve usage” can and cannot mean with current Codex:

- cancelling while an `agentMessage` is still streaming produced an interrupted rollout `token_count` event whose `info` was `null`; Codex supplied no usage for that incomplete response, so Multica correctly had nothing to record;
- cancelling during the later active-tool phase preserved the already-finalized preceding model response (one verification sample returned `input=9720`, `output=47`).

So this PR preserves provider-finalized usage already emitted before cancellation, especially in multi-response tool loops. It does not estimate tokens for the partially generated response when Codex itself reports no usage.

Validation after the measurement/docs change:

- `go test -race ./pkg/agent -run 'TestCodex(ExecuteCancellationInterruptsTurnAndPreservesUsage|NonResponsiveInterruptStopsProcessAtConfiguredDeadline|InterruptCompletesNearConfiguredDeadline)$' -count=20`
- `go test ./internal/daemon -run TestLoadConfig_CodexTurnInterruptTimeout -count=1`
- tagged integration test compile/skip path
- `git diff --check`

This addresses the measured-default rationale and configuration-documentation blocker. I am not claiming that it resolves the separate active-turn synchronization or usage-snapshot idempotency blockers from the latest review.

mameikagou:
@multica-eve could you please re-review blocker 3 (the 2s default and operator documentation) on current head `3e1d07bca468be4eb8e9b7a55c6c49455433ed88`?

The full method, environment, quantiles, reproduction command, and the provider-usage boundary are documented in [this evidence comment](https://github.com/multica-ai/multica/pull/7753#issuecomment-5489824332). To make the result independently auditable rather than relying only on summarized percentiles, here are the raw measured samples used for the decision.

All values below are milliseconds. Each scenario has 20 measured samples after 2 discarded warmups. Percentiles use nearest-rank over the sorted samples.

<details>
<summary>Raw samples: generating text</summary>

Cancellation was triggered only after the real app-server emitted `item/agentMessage/delta`, followed by a 100ms settle period. This excludes the earlier `turn/started` startup window and proves the response was actively streaming.

```text
turn/interrupt -> turn/completed:
8, 8, 15, 8, 9, 8, 9, 8, 8, 9, 9, 9, 9, 9, 9, 9, 9, 8, 8, 12

cancel -> Session.Result:
54.917495, 50.683522, 57.365596, 51.672563, 48.816116,
56.819388, 47.710654, 53.451946, 49.972626, 50.690551,
52.182640, 46.543551, 50.674354, 56.122011, 52.721229,
56.130452, 53.110737, 52.217217, 59.566430, 57.344220
```

Nearest-rank summary:

- protocol completion: P50 9ms, P95 12ms, P99/max 15ms
- full session result: P50 52.217217ms, P95 57.365596ms, P99/max 59.566430ms

</details>

<details>
<summary>Raw samples: active tool</summary>

Codex was instructed to run `sleep 30`; cancellation was triggered by the real `item/started:commandExecution` event exposed as Multica's `exec_command` tool-use message, while the command was still running.

```text
turn/interrupt -> turn/completed:
9, 9, 9, 9, 9, 9, 9, 13, 9, 9, 9, 9, 10, 9, 10, 10, 9, 9, 9, 9

cancel -> Session.Result:
57.268657, 55.410797, 68.159524, 53.505283, 56.375795,
53.344856, 52.162540, 66.783701, 59.653824, 57.014892,
59.488169, 57.338365, 59.815795, 55.433816, 58.687471,
64.711295, 51.641256, 56.197405, 62.439023, 57.050424
```

Nearest-rank summary:

- protocol completion: P50 9ms, P95 10ms, P99/max 13ms
- full session result: P50 57.050424ms, P95 66.783701ms, P99/max 68.159524ms

</details>

The measurement itself is checked in as an opt-in, authenticated real-agent integration test in [`server/pkg/agent/codex_interrupt_latency_integration_test.go`](https://github.com/mameikagou/multica/blob/3e1d07bca468be4eb8e9b7a55c6c49455433ed88/server/pkg/agent/codex_interrupt_latency_integration_test.go), guarded by both the `agentintegration` build tag and `MULTICA_RUN_REAL_AGENT_SMOKE=1`. Ordinary CI cannot execute it or consume account quota. Its interrupt budget is 10s, so the 2s candidate does not censor these observations.

Selection calculation:

- `2000ms / 15ms = 133.3x` headroom over the worst observed protocol completion;
- `2000ms / 68.159524ms = 29.3x` headroom over the worst observed end-to-end cleanup;
- a timeout still immediately enters the tested process-tree force-stop path, so 2s remains a strict upper bound rather than being followed by the old unbounded/10s cleanup regression;
- the environment override is now documented in `CLI_AND_DAEMON.md` and all EN/ZH/JA/KO environment-variable pages.

I am deliberately not presenting this as Windows latency evidence: the measurements are WSL/Linux amd64 with `codex-cli 0.147.0`. The retained margin is the cross-platform/host-contention allowance, while the Windows process-tree behavior is covered separately by CI.

The complete CI run for this head is green, including `backend-tests` and `windows-execenv`: [CI run 33477495094](https://github.com/multica-ai/multica/actions/runs/33477495094).

This request is specifically to confirm whether blocker 3 is resolved by the measured evidence, selection rationale, reproducible probe, and documentation. I am not asking you to waive the separate active-turn synchronization and usage-snapshot idempotency blockers.

multica-eve (changes requested):
This revision gives cancelled tasks a measured, bounded chance to preserve usage and now documents the operator control. However, cancellation can still miss usage when it overlaps the start of a task, and resumed tasks can still count historical usage again. Members may therefore see inaccurate budgets and usage history, so this head is not ready to merge.

Product judgment: this is a real problem worth solving, and the measured two-second tradeoff is acceptable. The remaining blockers are implementation correctness.

## Request changes

The measured default and its documentation are now sufficient, and the current head merges cleanly with `main` while preserving the handshake budget. Two must-fix issues remain:

1. **Synchronize access to the active turn ID.**

   The cancellation lifecycle reads `c.turnID` at `server/pkg/agent/codex.go:2419` while the stdout reader writes it at `server/pkg/agent/codex.go:3264`, without synchronization. This is a real race, not only a theoretical concern:

   ```text
   go test -race ./pkg/agent -run '^TestCodexInterruptCompletesNearConfiguredDeadline$' -count=100
   WARNING: DATA RACE
   ```

   The race detector points to those two accesses and fails the test. Please make all active-turn reads and writes use one synchronized mechanism (for example, a mutex-backed accessor or equivalent ownership model), and keep a stable race regression test.

2. **Exclude resume replay and make usage accounting idempotent.**

   Codex v0.147 defines `thread/tokenUsage/updated` with both `threadId` and `turnId`, and the resume path deliberately emits the restored historical token-usage notification after the resume response ([protocol](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/app-server-protocol/src/protocol/v2/thread.rs#L1579-L1634), [resume implementation](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/app-server/src/request_processors/thread_processor.rs#L3421-L3439), [upstream test](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/app-server/tests/suite/v2/thread_resume.rs#L2351-L2398)).

   Multica arms its usage gate after the resume response at `server/pkg/agent/codex.go:1547`. The usage notification then falls through the default-accept path at `server/pkg/agent/codex.go:2324-2327`, while `server/pkg/agent/codex.go:3444-3461` ignores the notification's `turnId` and directly adds every `last` snapshot. Depending on scheduling, the restored history is either discarded before the gate is armed or charged after it is armed; repeated identical snapshots can also be counted repeatedly.

   Please accept raw v2 usage only for the active `turnId` and derive an idempotent monotonic delta from `total`/a baseline (or use an equivalent deduplication design). Add coverage for a historical replay arriving after resume, duplicate identical snapshots, and multiple current-turn responses.

Validation on `3e1d07bca468be4eb8e9b7a55c6c49455433ed88`:

- `go test ./pkg/agent -count=1` passed.
- `go test ./internal/daemon -count=1` passed.
- Targeted interrupt/handshake/config tests passed.
- Windows cross-compilation of `./pkg/agent` passed.
- The tagged real-agent integration test compiles and skips safely without explicit quota authorization.
- All 15 current GitHub checks pass.

Once the race and replay/idempotency cases are fixed, this should be ready for another review.

mameikagou:
Addressed the two remaining blockers on current head 9b6d4c15e. The implementation is in d74d72dd9, followed by a clean merge of current upstream main at 9d09902df.

1. Active turn synchronization

- Added one mutex-backed active-turn accessor and routed every lifecycle read/write through it, including turn/started, cancellation interrupt, timeout diagnostics, and lifecycle logging.
- The exact race shape from the review now passes 100 repeated runs after the main merge.

2. Idempotent v2 usage attribution

- thread/tokenUsage/updated is accepted only when its turnId exactly matches the synchronized active turn ID. A restored historical notification emitted after thread/resume but before the new turn is therefore ignored even though the notification gate is already armed.
- The first accepted current-turn snapshot contributes tokenUsage.last and records tokenUsage.total. Later snapshots contribute only the monotonic delta from the preceding total, so identical delivery contributes zero while multiple model responses in one turn remain accumulated.
- Added regressions for post-resume historical replay, an identical duplicate snapshot, and multiple current-turn responses, including cached input, cache-write input, normal output, and reasoning output.

Validation on the merged head:

- go test -race ./pkg/agent -run ^(TestCodexInterruptCompletesNearConfiguredDeadline|TestCodexThreadTokenUsageUpdatedDropsResumeReplay|TestCodexThreadTokenUsageUpdatedDeduplicatesSnapshot|TestCodexThreadTokenUsageUpdatedAccumulatesCurrentTurnResponses)$ -count=100
- go test ./pkg/agent -count=1
- env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1
- GOOS=windows GOARCH=amd64 go test -c ./pkg/agent
- go vet ./pkg/agent
- the opt-in agentintegration test compiles and takes its safe skip path without quota authorization
- git diff --check upstream/main...HEAD

All passed.

@multica-eve please re-review the active-turn synchronization and replay/idempotency coverage on the current head.

multica-eve (changes requested):
This revision fixes the previously identified cases where cancellation could miss usage or restored history could be counted again. However, input used to create a cache is now counted in two billing categories, while the fallback path loses that category entirely. Members can therefore see inflated usage and estimated cost, so this head is not ready to merge.

Product judgment: preserving finalized usage on cancellation is the right product behavior, and the bounded cleanup policy is acceptable. The remaining blocker is billing correctness.

## Request changes

### Make cache-write input mutually exclusive across the v2 and fallback paths

Codex v0.147.0's own parser test reports `input_tokens=100` with input details of `cached_tokens=40` and `cache_write_tokens=60`, while `total_tokens=110` adds only the 10 output tokens. In the protocol targeted by this PR, cache-read and cache-write tokens are therefore already included in the input total: [upstream test](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/codex-api/src/sse/responses.rs#L813-L838).

Multica charges plain input, cache read, and cache write independently in `server/internal/metrics/business.go:615-620`, and sums those three buckets as total input in `server/internal/handler/daemon.go:4397`, so the buckets must be mutually exclusive.

The current v2 path at `server/pkg/agent/codex.go:3496-3499` subtracts only `CachedInputTokens` from `InputTokens`, then also records `CacheWriteInputTokens` separately. The new multi-response test at `server/pkg/agent/codex_test.go:3834-3869` locks in the resulting overcount: its two raw input deltas total 250, with 80 cache-read and 10 cache-write tokens. Plain input should be `250 - 80 - 10 = 160`, but the test expects 170, making the three stored input buckets total 260.

The fallback has the complementary problem. `server/pkg/agent/codex.go:3752-3758` now parses `cache_write_input_tokens`, but `server/pkg/agent/codex.go:3859-3863` neither puts it into `CacheWriteTokens` nor removes it from plain input. When direct v2 usage is unavailable, the separate cache-write bucket is lost.

Please normalize v0.147 usage into mutually exclusive buckets in both paths: subtract both cache-read and cache-write details from raw input (with non-negative protection), preserve `CacheWriteTokens` in the JSONL fallback, change the multi-response expectation to plain input 160, and add fallback cache-write coverage. The stale comment at `server/internal/metrics/pricing.go:28-31` should also be updated now that this PR consumes the separately reported field.

The two previous blockers are resolved on `9b6d4c15e4a0cc50bb3f7b944570deeba4872a41`:

- All active-turn lifecycle access uses the mutex-backed accessor; the prior race shape passes 100 repeated race-enabled runs.
- Usage is attributed only to the active turn and uses a cumulative baseline, with passing regressions for post-resume history replay, duplicate snapshots, and multiple responses.

Validation on this head:

- The focused `-race -count=100` suite passed in 67.659s.
- `go test ./pkg/agent -count=1` passed in 237.368s.
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1` passed in 37.465s.
- The prompt-prefix byte-identity test, `go vet ./pkg/agent`, Windows cross-compilation, diff check, and merge-tree check passed.
- All 15 current GitHub checks pass, and the PR is mergeable/clean.

Once the cache-write accounting is corrected in both collection paths, this should be ready for another review.

mameikagou:
Addressed the final cache-write billing blocker in `f2cb38b10`; current head `39e91e387` also merges the latest `upstream/main` (`4dfa08917`) cleanly.

- The direct v2 path now subtracts both cache-read and cache-write details from Codex raw input, with the existing non-negative clamp.
- The JSONL fallback applies the same normalization and preserves `CacheWriteTokens` instead of losing that bucket.
- The multi-response regression now asserts the mutually exclusive result: raw input 250 = plain input 160 + cache read 80 + cache write 10.
- Added equivalent JSONL resume/delta coverage and updated the stale v0.144.1 pricing comment for the v0.147 contract.

Validation on the merged head:

- `go test -race ./pkg/agent -run '^(TestParseCodexSessionFileSinceResumeEdgeCases|TestCodexThreadTokenUsageUpdatedDropsResumeReplay|TestCodexThreadTokenUsageUpdatedDeduplicatesSnapshot|TestCodexThreadTokenUsageUpdatedAccumulatesCurrentTurnResponses|TestCodexInterruptCompletesNearConfiguredDeadline)$' -count=100`
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1`
- `go test ./internal/metrics -count=1`
- `go vet ./pkg/agent`
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/agent`
- the opt-in `agentintegration` test compiles and takes its safe skip path without quota authorization

All passed. @multica-eve please re-review the corrected v2 and JSONL cache-write accounting.

mameikagou:
CI follow-up for head `39e91e387`: all 15 checks are complete and green (with `image-budget` correctly skipped), including `backend-tests` (8m18s) and `windows-execenv` (2m6s): https://github.com/multica-ai/multica/actions/runs/33716943576

PR #7754, opened 2026-08-29
MUL-6835 fix(i18n): localize status and priority values

sooodasama:
## What does this PR do?

Fixes localized UI surfaces that still displayed English labels from server/config data:

- Inbox status changes used the built-in catalog's English seed name.
- Inbox priority changes used the English label from `PRIORITY_CONFIG`.
- Project search results used the English label from `PROJECT_STATUS_CONFIG`.

The UI now reuses the existing status, priority, and project-status localization helpers. Unknown or custom values still fall back to their original server value, and existing icons, colors, and behavior are unchanged.

## Related Issue

N/A

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor
- [x] Tests

## Changes

- Localize built-in status labels in Inbox activity details.
- Localize known priority labels in Inbox activity details.
- Localize project status labels in search results.
- Add regression coverage using Simplified Chinese resources.

## How has this been tested?

- `corepack pnpm --filter @multica/views typecheck`
- `corepack pnpm --filter @multica/views exec vitest run inbox/components/inbox-detail-label.test.tsx search/search-command.test.tsx` (37 tests)
- `corepack pnpm --filter @multica/views lint` (0 errors; existing repository warnings only)

Manual scenario covered by tests: in a Simplified Chinese UI, Inbox status/priority changes and project search results show translated values instead of English config seeds.

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my code.
- [x] I have added tests that prove my fix is effective.
- [x] New and existing unit tests pass locally with my changes.
- [x] I checked that the staged diff contains no downstream branding, private paths, or product-specific behavior.

## Screenshots

Not included because this is a copy-only localization correction with no layout changes.

## Risk

Low. The change only affects display labels and preserves raw-value fallback for unknown/custom values. It does not change API payloads, persisted state, status transitions, sorting, or styling.

## AI assistance

Codex was used to inventory the affected UI paths, reuse the repository's existing localization helpers, add focused regression tests, and audit the staged diff for downstream branding.

### DIFF
--- a/packages/views/inbox/components/inbox-detail-label.test.tsx
+++ b/packages/views/inbox/components/inbox-detail-label.test.tsx
@@ -27,6 +27,13 @@ vi.mock("@multica/core/issue-statuses/hooks", () => ({
     isLoaded: true,
   }),
 }));
+// Scope note: stubbing the resolver keeps this suite about the ONE thing this
+// component owns — which resolver it asks. The catalog stub above still hands
+// back the English seed name for `in_progress`, so a component that went back
+// to reading `entryOf(...).name` renders "In Progress" and fails here. What
+// the real `useStatusLabel` does with a key (built-in through i18n, custom
+// through the catalog) is its own contract, covered by `run-confirm.test.tsx`
+// and `issues/utils/status-options.test.tsx`.
 vi.mock("../../issues/utils/status-label", () => ({
   useStatusLabel: () => (key: string) =>
     key === "in_progress" ? "进行中" : key,
--- a/packages/views/inbox/components/inbox-detail-label.test.tsx
+++ b/packages/views/inbox/components/inbox-detail-label.test.tsx
@@ -2,6 +2,7 @@ import { render } from "@testing-library/react";
 import { describe, expect, it, vi } from "vitest";
 import type { InboxItem } from "@multica/core/types";
 import en from "../../locales/en/inbox.json";
+import zhHansIssues from "../../locales/zh-Hans/issues.json";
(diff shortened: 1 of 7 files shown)


Bohan-J:
Thanks for this, @sooodasama — nicely scoped fix.

What made it easy to accept: the inconsistency was already visible inside a single Inbox row, where the status icon's tooltip went through `useStatusLabel` while the detail line right beside it rendered the English catalog seed. You reused the helpers we already had instead of introducing a parallel path, and English output is byte-identical to before (the `en` bundles match the config labels verbatim), so there's no copy change for existing users.

I pushed one follow-up commit to your branch:

- Extracted the guarded priority resolver into `packages/views/issues/utils/priority-label.ts`. The `priority in PRIORITY_CONFIG ? t(...) : raw` shape already lived in the issue activity formatter, and your change made it a second copy — status has `useStatusLabel`, so priority now has its counterpart. It's a plain function rather than a hook because the activity formatter is a plain function holding a `t`.
- Corrected the contract note on the project label hooks: your change made search the last caller to stop reading `PROJECT_STATUS_CONFIG.label`, so that comment no longer described reality.
- Added scope notes to the two new tests — the Inbox suite stubs `useStatusLabel` deliberately, and the search suite's `zh-Hans` bundle is English except for `projects` deliberately. Both read like mistakes without a line saying otherwise.

Verified locally on the merged tree: typecheck clean, `@multica/views` full suite 4861 passing, lint 0 errors. Merging now.

PR #7756, opened 2026-08-29
MUL-6836: fix(codex): allow explicit standard speed

mameikagou:
## Problem

The Codex speed picker currently mirrors only the alternative tiers advertised by `codex debug models`. That catalog advertises `priority` (Fast), but intentionally omits Codex's `default` sentinel. As a result, the UI offers only:

- Runtime default (no override)
- Fast (`priority`)

Users cannot explicitly select standard speed when their local Codex configuration defaults to Fast.

## Fix

- Preserve three distinct states: empty = inherit local Codex config, `default` = explicit Standard, and a catalog tier such as `priority` = explicit Fast.
- Negotiate explicit-Standard support through the daemon model response instead of inferring it from `provider === "codex"`.
- Gate that capability on the installed Codex CLI version. Codex 0.133.0 is the first stable release containing the request-only `default` sentinel from [openai/codex#23537](https://github.com/openai/codex/pull/23537).
- Treat a missing capability from an older daemon, or Codex <0.133.0, as unsupported. The picker does not offer Standard and execution-time validation drops a stale value.
- Keep the capability CLI-wide and model-independent, while alternative tiers remain exact per-model catalog capabilities.
- Carry the additive capability through daemon report, server cache/API schema, and frontend; malformed capability values fail schema parsing.
- Update create/update CLI help, the built-in `multica-creating-agents` Skill and source map, and the Chinese label (`标准速度`).

The backend continues to forward non-empty service tiers through `serviceTier` on thread start/resume and turn start, so no private Multica value or protocol translation is introduced.

## Compatibility coverage

- New frontend + old daemon response (capability omitted): Standard is not offered.
- Current daemon + Codex 0.132.x: capability is false and `default` is rejected.
- Current daemon + Codex 0.133.0+: capability is reported, survives daemon/server/API transport, and Standard is offered.
- No pinned model: Standard remains available because this is a CLI-level capability; Fast remains per-model.

## Tests

- `go test ./cmd/multica -count=1`
- `go test ./internal/service -run '^TestBuiltinSkillsConformToTemplate$' -count=1`
- Windows cross-compilation for `pkg/agent`, `internal/daemon`, and `cmd/multica`
- Core: 139 files / 1,672 tests passed; typecheck and ESLint passed
- Views: 409 files / 4,863 tests passed under `TZ=UTC`; typecheck passed; ESLint completed with no errors (pre-existing repository warnings only)


Bohan-J (dismissed):
**Request changes.** The product problem is real and the direction is right, but the release contract needs to be fixed before this can merge.

## What this changes for users

Before: an agent's speed could only be "runtime default" or Fast. If a member's local Codex config already defaults to Fast, there was no way to pin one agent to standard speed. After: runtime default / Standard / Fast are three distinct choices, and an explicit Standard survives a model change.

That is a real gap and worth a persisted setting, and `default` is the right token — upstream Codex treats it as an explicit standard-routing request, distinct both from unset and from catalog tiers like `priority` (openai/codex#23537).

## Must fix 1 — the picker synthesizes a capability the executing side may not have

`packages/views/agents/components/inspector/service-tier-setting-field.tsx` injects the `default` option from the provider name alone:

```ts
provider === "codex"
  ? [{ id: "default", ... }, ...tiers.filter((tier) => tier.id !== "default")]
  : tiers
```

and `ValidateServiceTierWith` in `server/pkg/agent/thinking.go` accepts it unconditionally in the same commit. But web/server and the local daemon (and the Codex CLI behind it) upgrade independently, and `default` is deliberately absent from the model catalog — so nothing in the persisted value tells the executing side whether it can honor the sentinel.

In the ordinary "new web + older daemon" window, the picker lets a member save Standard, while the older daemon's per-model catalog check does not recognize `default` (it is not an advertised tier, and an empty model fails closed there). The tier gets dropped and the run keeps inheriting the local Fast config. The same silent outcome applies when the daemon is current but the machine's Codex CLI predates the `default` sentinel. The member sees "Standard" selected, pays Fast usage, and no error surfaces anywhere.

Worth noting that `IsKnownServiceTier` is only a safe-token gate, so the API persists `default` on any server version — the mismatch is invisible until execution. Every test added here runs web and server at the same revision, so green CI cannot catch this combination.

Suggested fix: put the capability back on the existing catalog/capability negotiation path. Have the daemon report explicit-standard support (a dedicated field, or a synthesized tier the daemon itself owns) and show the option only when that signal is present; provider name is not proof of capability. Please also add a mixed-version regression test asserting that a new frontend reading an older daemon's model response does not offer `default`.

## Must fix 2 — CLI help and built-in skill docs still state the old contract

This PR makes `default` a special legal value that is not in the model catalog and is accepted with an empty model, but the authoritative entry points still say the opposite:

- `server/cmd/multica/cmd_agent.go:167` and `:190` — "from the selected model's runtime catalog"
- `server/internal/service/builtin_skills/multica-creating-agents/SKILL.md:131` and `:173-176` — "exact model/tier pair checked by daemon", "currently `priority`, shown as Fast"
- `server/internal/service/builtin_skills/multica-creating-agents/references/creating-agents-source-map.md:63` and `:98` — "Accepts only a tier advertised for the explicit Codex model; empty model fails closed"

Root `CLAUDE.md` requires CLI/API/product behavior changes to update the related built-in skill and source map in the same change. Please update the create/update flag help and both skill documents to spell out the three states: empty = inherit local config, `default` = explicit standard, catalog tier (e.g. `priority`) = explicit Fast. Otherwise CLI users and later agents get instructions that contradict the code.

## Nit

`packages/views/locales/zh-Hans/agents.json:235` uses 原始速度 ("original speed"), while en/ja/ko use Standard / 標準 / 표준. 原始 reads as "the previous speed" rather than the product concept. Suggest 标准 or 标准速度.

## Notes

All checks on this PR are green (backend, backend-tests, frontend, frontend-build, frontend-test, frontend-views-test, installer, windows-execenv, sqlc-check).

On rollback risk: reverting the web change would only stop new selections. A `default` already persisted is still ignored by an older daemon, so the fix has to be forward, or the stored value has to be cleaned up.

mameikagou:
@Bohan-J Addressed both must-fix items in 92108f221.

1. Capability is now negotiated, not synthesized from the provider name:
   - the daemon detects the installed Codex version and reports `supports_explicit_standard_service_tier=true` only for Codex >=0.133.0 (the first stable release containing openai/codex#23537);
   - the field is carried through daemon -> server cache/API schema -> frontend;
   - the picker and model-change cleanup require that signal, and execution-time validation uses the same signal before injecting `default`;
   - omission from an older daemon is false by default, and Codex 0.132.x rejects `default`.

   Added regressions for new frontend + old daemon response, current daemon + old Codex, capability wire round-trip, malformed API values, and the CLI-level/no-pinned-model case.

2. Updated both create/update `--service-tier` help strings plus `multica-creating-agents/SKILL.md` and its source map with the three-state contract: empty = inherit, `default` = explicit Standard, catalog tier such as `priority` = explicit Fast.

Also changed the Chinese label from `原始速度` to `标准速度`.

Local verification includes full `pkg/agent`, `internal/daemon`, `internal/handler`, and CLI Go packages; built-in Skill conformance; Windows cross-compilation; 1,672 Core tests; and 4,863 Views tests. Please take another look when CI finishes.

mameikagou:
CI is now fully green on 92108f221: backend, frontend/build/tests, both Views shards, Windows execenv, installers, vulnerability scan, and sqlc-check all passed.

multica-eve (approved):
Thanks for the thorough follow-up and for addressing the compatibility and documentation gaps. The daemon-negotiated capability, mixed-version behavior, and explicit Standard runtime path now look good. Approved.

Issue #3854, opened 2026-06-06
[Bug]: Done-issue revive/rerun passes resume_session from the old task workdir — per-cwd Claude session can't resolve, instant "claude execution failed"


seanpmgallagher:
### Deployment type

self-host

### What happened?

Reviving a **done** issue (via a mention on the issue or `multica issue rerun`) provisions a **fresh task workdir** but passes `resume_session` pointing at the **old task's** Claude session. Claude CLI sessions are stored per-cwd (`~/.claude/projects/<encoded-cwd>/`), so the resume finds nothing in the new workdir → claude exits in <1s → the run fails as `agent_error.unknown` with empty run messages ("claude execution failed").

The tell that it's not an auth/runtime problem: concurrent runs on *other* issues on the same daemon complete fine at the same time.

### Steps to reproduce

1. Let an issue run to `done` (Claude backend; the server persists the task's session, keyed to that task's workdir).
2. Mention-revive the issue (or `multica issue rerun <issue>`).
3. A fresh task workdir is provisioned for the new run, but the run is started with `resume_session` referencing the old task's session id.
4. `claude --resume <old-session>` runs with cwd = new workdir → per-cwd session store has no such session → exits in <1s.
5. Run shows `agent_error.unknown`, no messages. Every retry fails the same way — the issue is permanently unrevivable without manual intervention.

**Expected behavior**
Either don't resume across workdirs (start a fresh session for the new workdir, optionally injecting prior context), or migrate/copy the session directory so the resume target exists for the new cwd.

### Additional context

- Observed 2026-06-06, CLI/daemon 0.3.17, Claude Code backend, self-hosted backend (ghcr.io/multica-ai/multica-backend).
- Likely related to (but distinct from) #1284: that issue was about the resume **fallback** never triggering when `--resume` hits a dead session. Here the dead-session state is *created by the revive flow itself* — the new workdir guarantees the persisted session can't resolve. Even with a working fallback, wiring `resume_session` to a session from a different workdir is the root defect.

hsuanguo:
Independent reproducer (CLI/daemon `0.3.17`, Claude Code backend) — same signature as the report:

- An issue had a completed run several weeks ago in its original task workdir; that run's `session_id` was persisted on the server.
[...]

PR #7760, opened 2026-08-29
MUL-6840 fix(pi): preserve sessions across workdir changes

mameikagou:
## What does this PR do?

Keeps Pi and OMP provider sessions resumable when a follow-up task receives a different workdir.

### Thinking path

Pi's backend uses an absolute JSONL file path as its opaque session ID and passes it directly to `--session`. That session store lives under `~/.multica/pi-sessions`, not under the task workdir.

The daemon's shared resume gate nevertheless required every provider to reuse the previous workdir. When workdir reuse failed or changed, it cleared Pi's still-valid session path, started a new provider conversation, and injected a continuity notice that made the agent reconstruct context through `multica chat history`.

This PR makes reachability match the provider's actual storage model: Pi/OMP resume whenever their recorded session file still exists, independently of workdir identity. Cwd-keyed providers and Hermes retain their existing gates.

## Related Issue

N/A — reported from a live Pi chat where the Multica transcript and workdir survived but the provider session restarted.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Update the daemon resume gate to recognize Pi/OMP's independently addressed JSONL session store.
- Preserve a Pi/OMP `PriorSessionID` across workdir changes when the session file exists.
- Continue dropping the pointer and surfacing the continuity gap when the session file is actually missing.
- Add regression coverage for both Pi and OMP and for the missing-file fallback.

## How to Test

1. Run `cd server && go test ./internal/daemon -run 'TestGate(ResumeToReachableSession|PiResume)' -count=1`.
2. Run `cd server && go test ./internal/daemon -count=1`.
3. Run `cd server && go vet ./internal/daemon`.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (inline runtime documentation updated; no user-facing docs affected)

### Risk

The exception is limited to the two providers backed by `piBackend`. The server already scopes prior session IDs to the same runtime; the daemon additionally requires the recorded path to resolve to a regular file. Every other provider keeps the existing workdir/session-store checks.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Traced a live Pi context-loss report through chat-session selection, daemon resume gating, and Pi's JSONL session implementation; wrote the failing cross-workdir regression first, then made the smallest provider-scoped change and ran focused package tests plus vet.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5769,14 +5769,16 @@ func sameExistingDir(a, b string) bool {
 }
 
 func gateResumeToReachableSession(task *Task, taskCtx *execenv.TaskContextForEnv, provider, envWorkDir string, sessionHomeReachable bool, taskLog *slog.Logger) bool {
-	// Compare the directories, not the spelling. Reuse runs in the canonical
-	// path it validated and locked, which need not be character-identical to
-	// the PriorWorkDir the server sent — a symlinked workspaces root is enough
-	// to make them differ. A string compare would then silently drop the prior
-	// session on every follow-up task in that installation.
-	reachable := task.PriorWorkDir != "" && sameExistingDir(envWorkDir, task.PriorWorkDir) && sessionHomeReachable
-	if provider == "pi" || provider == "omp" {
+	var reachable bool
+	if providerUsesPiSessionFile(provider) {
 		reachable = piSessionFilePresent(task.PriorSessionID)
+	} else {
+		// Compare the directories, not the spelling. Reuse runs in the canonical
+		// path it validated and locked, which need not be character-identical to
+		// the PriorWorkDir the server sent — a symlinked workspaces root is enough
+		// to make them differ. A string compare would then silently drop the prior
+		// session on every follow-up task in that installation.
+		reachable = task.PriorWorkDir != "" && sameExistingDir(envWorkDir, task.PriorWorkDir) && sessionHomeReachable
 	}
 	if !reachable && task.PriorSessionID != "" {
 		taskLog.Info("dropping prior session: session store not reachable from this run",
@@ -5798,12 +5800,24 @@ func gateResumeToReachableSession(task *Task, taskCtx *execenv.TaskContextForEnv
 	return reachable
 }
 
+func providerUsesPiSessionFile(provider string) bool {
+	if provider == "pi" {
+		return true
+	}
+	desc, ok := agent.BuiltinRuntimeByID(provider)
+	return ok && desc.ProtocolFamily == "pi"
+}
+
+// piSessionFilePresent proves there is persisted history to resume. It does
+// not claim the transcript is idle: the Pi backend takes an exclusive lock for
+// the complete child-process lifetime and reports a busy resume as rejected so
+// runTask falls back to its existing one-shot fresh-session path.
 func piSessionFilePresent(sessionID string) bool {
 	if sessionID == "" {
 		return false
 	}
 	info, err := os.Stat(sessionID)
-	return err == nil && info.Mode().IsRegular()
+	return err == nil && info.Mode().IsRegular() && info.Size() > 0
 }
 
 // sessionHomeReachable reports whether a session recorded by a prior task on
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5721,16 +5721,22 @@ func providerNeedsInlineSystemPrompt(provider string) bool {
 	}
 }
 
-// gateResumeToReusedWorkdir clears the task's prior session unless this run
+// gateResumeToReachableSession clears the task's prior session unless this run
 // can actually reach the store the session lives in, and reports whether that
-// held. CLI backends key their session stores to the cwd (Claude Code looks
-// sessions up under ~/.claude/projects/<encoded-cwd>/), so a session id from a
-// different workdir can never resolve: the CLI exits within a second and the
-// run fails before doing any work — permanently, because the failed run
-// records no session and the next claim serves the same stale pointer again.
-// This fires whenever the prior workdir no longer exists (GC'd after the issue
-// went done, daemon reinstall, manual cleanup) and execenv.Reuse fell back to a
-// fresh Prepare (GitHub #3854).
+// held. Most CLI backends key their session stores to the cwd (Claude Code
+// looks sessions up under ~/.claude/projects/<encoded-cwd>/), so a session id
+// from a different workdir can never resolve: the CLI exits within a second
+// and the run fails before doing any work — permanently, because the failed
+// run records no session and the next claim serves the same stale pointer
+// again. This fires whenever the prior workdir no longer exists (GC'd after
+// the issue went done, daemon reinstall, manual cleanup) and execenv.Reuse fell
+// back to a fresh Prepare (GitHub #3854).
+//
+// Pi and OMP are the exception. Their opaque session id is an absolute JSONL
+// path under ~/.multica/pi-sessions, and the backend passes that path directly
+// to --session. The transcript remains resumable when only the task workdir
+// changes, so binding it to workdir reuse discards healthy conversation history
+// and forces the model to reconstruct it through `multica chat history`.
 //
 // A matching workdir is not sufficient on its own. Hermes keys its sessions to
 // HERMES_HOME — the per-task overlay under envRoot — not to the cwd, and the
@@ -5745,8 +5751,8 @@ func providerNeedsInlineSystemPrompt(provider string) bool {
 // HermesSessionStore) — and false drops the resume with the same disclosure as
 // a workdir mismatch.
 // sameExistingDir reports whether two paths name the same existing directory.
-// False when either cannot be stat'd, which is the safe answer here: an absent
-// prior workdir means there is nothing to resume from.
+// False when either cannot be stat'd, which is the safe answer for cwd-keyed
+// providers: an absent prior workdir means there is nothing to resume from.
 func sameExistingDir(a, b string) bool {
 	if a == "" || b == "" {
 		return false
@@ -5762,15 +5768,19 @@ func sameExistingDir(a, b string) bool {
 	return os.SameFile(ai, bi)
 }
 
-func gateResumeToReusedWorkdir(task *Task, taskCtx *execenv.TaskContextForEnv, envWorkDir string, sessionHomeReachable bool, taskLog *slog.Logger) bool {
+func gateResumeToReachableSession(task *Task, taskCtx *execenv.TaskContextForEnv, provider, envWorkDir string, sessionHomeReachable bool, taskLog *slog.Logger) bool {
 	// Compare the directories, not the spelling. Reuse runs in the canonical
 	// path it validated and locked, which need not be character-identical to
 	// the PriorWorkDir the server sent — a symlinked workspaces root is enough
 	// to make them differ. A string compare would then silently drop the prior
 	// session on every follow-up task in that installation.
-	reused := task.PriorWorkDir != "" && sameExistingDir(envWorkDir, task.PriorWorkDir) && sessionHomeReachable
-	if !reused && task.PriorSessionID != "" {
+	reachable := task.PriorWorkDir != "" && sameExistingDir(envWorkDir, task.PriorWorkDir) && sessionHomeReachable
+	if provider == "pi" || provider == "omp" {
+		reachable = piSessionFilePresent(task.PriorSessionID)
+	}
+	if !reachable && task.PriorSessionID != "" {
 		taskLog.Info("dropping prior session: session store not reachable from this run",
+			"provider", provider,
 			"session_id", task.PriorSessionID,
 			"prior_workdir", task.PriorWorkDir,
 			"workdir", envWorkDir,
@@ -5785,7 +5795,15 @@ func gateResumeToReusedWorkdir(task *Task, taskCtx *execenv.TaskContextForEnv, e
 		taskCtx.PriorSessionResumeUnavailable = true
 		task.PriorSessionResumeUnavailable = true
 	}
-	return reused
+	return reachable
+}
+
+func piSessionFilePresent(sessionID string) bool {
+	if sessionID == "" {
+		return false
+	}
+	info, err := os.Stat(sessionID)
+	return err == nil && info.Mode().IsRegular()
 }
 
 // sessionHomeReachable reports whether a session recorded by a prior task on
@@ -5807,9 +5825,8 @@ func gateResumeToReusedWorkdir(task *Task, taskCtx *execenv.TaskContextForEnv, e
 //   - With no store, the transcript is the overlay's own task-local file, which
 //     survives exactly when this run reused the prior task's env root.
 //
-// Every other provider is keyed by cwd (or resolves its own store), so the
-// workdir comparison in gateResumeToReusedWorkdir remains the whole answer and
-// this returns true.
+// Every other provider is keyed by cwd or handles its independently addressed
+// store in gateResumeToReachableSession, so this predicate returns true.
 func sessionHomeReachable(provider string, env *execenv.Environment, envReused bool) bool {
 	if provider != "hermes" {
 		return true
@@ -5904,7 +5921,7 @@ func shouldReusePriorWorkdir(task Task, localAssignment *localDirectoryAssignmen
 
 // gateCodexResumeToRolloutPresence drops the prior Codex session when its
 // rollout is not actually present in the task's CODEX_HOME sessions. A reused
-// workdir keeps PriorSessionID (gateResumeToReusedWorkdir), but Codex session
+// workdir keeps PriorSessionID (gateResumeToReachableSession), but Codex session
 // isolation (MUL-4424) means the rollout may be missing: a migrated legacy home
 // that could not locate it, or a local_directory task whose shared history was
 // pruned. Codex would then silently thread/start from scratch, so we clear the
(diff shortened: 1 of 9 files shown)


Bohan-J (changes requested):
Thanks for this — and especially for the "Thinking path" section. Tracing the gate back to its stated rationale is exactly the right way to approach it, and your reading is correct.

I verified the premise independently rather than taking the description at face value:

- Pi's session id really is a cwd-independent absolute path (`newPiSessionPath` → `~/.multica/pi-sessions/<ts>.jsonl`, `server/pkg/agent/pi.go:750`), passed straight to `--session`.
- The commit that introduced the gate (#4027, fixing #3854) justifies it **specifically** by cwd-keyed session stores — "Claude Code looks sessions up under `~/.claude/projects/<encoded-cwd>/`". It was written as a blanket rule, but its own reasoning never covered Pi.

So this is a legitimate narrowing of an over-broad rule back to its declared scope, not a new parallel path. Renaming `gateResumeToReusedWorkdir` → `gateResumeToReachableSession` makes the predicate say what it actually means, and keeping the missing-file case on the existing continuity-disclosure path (instead of letting `ensurePiSessionFile` silently create an empty file and fake a resume) is the right call — that's the best-judged part of the PR.

That said, I'm requesting changes on one blocking issue.

---

## Blocking: this removes the de-facto mutual exclusion on the Pi session file

The Risk section says the change is safe because prior session ids are runtime-scoped and the path must resolve to a regular file. Both are true — I checked, `prior.RuntimeID == task.RuntimeID` does gate it server-side. But neither addresses concurrency, and the workdir gate was silently providing it.

`lockReusablePriorEnvRoot` takes an **exclusive** lock (`execenv.EnvRootClaim`, an `os.File` lock held via `defer priorClaim.Release()` for the whole of `runTask`). Its comment states an invariant:

> `Declining is always safe: the caller falls back to a fresh Prepare, which costs session continuity, not correctness.`

That invariant holds *because* losing the lock meant a different workdir, which meant the old gate dropped the session. In other words, the workdir gate incidentally guaranteed **at most one live run per Pi session file**. This PR lets pi/omp bypass the gate without putting anything in that guarantee's place.

Concrete path:

1. User sends a message in a Pi chat; task A starts. A publishes its session path mid-run via `PinTaskSession` (`server/internal/handler/task_lifecycle.go:69`) — by design, so a crash doesn't lose the resume pointer.
2. A is still running when a second message arrives (or a second comment lands on an issue). Task B is claimed and the server hands it `PriorSessionID = <A's live S.jsonl>`. Claim is explicitly not serialized — `server/internal/handler/daemon.go:3020`: *"This is advisory context, not a queue gate"*, and the chat path documents *"independent debounce timers race"*.
3. B tries to reuse A's env root, A holds the lock → declines → fresh workdir.
4. **Before this PR:** workdir differs → session dropped → B starts a fresh Pi session. `S.jsonl` has exactly one writer.
5. **After this PR:** `piSessionFilePresent(S.jsonl)` is true → B also passes `--session S.jsonl`. Two `pi` processes append to the same JSONL concurrently.

The result is two conversations interleaved in one transcript, which every later resume then reads. That's worse than the bug being fixed: losing context once is recoverable and visible; a corrupted transcript is persistent and silent.

Two ways out — either is fine, but please say which you're taking:

- Require, in the gate, that no live run currently holds this session file (a `path → run` map in the daemon is enough), and fall back to the old behaviour when it does; or
- Take a real exclusive lock on the pi/omp session file and treat "cannot acquire" as not reachable.

(`markActiveStore`/`unmarkActiveStore` at `daemon.go:8792` is the nearest existing machinery, but it's a GC refcount, not exclusion — it won't work as-is.)

---

## Should fix

**1. Existence is checked, emptiness isn't — a 0-byte session reads as intact context**

`piSessionFilePresent` does `os.Stat` + `IsRegular`. But `ensurePiSessionFile` (`server/pkg/agent/pi.go:763`) creates the file *before* Pi launches, and failed tasks still report their session id — `FailTask` carries it (`daemon.go:5611`), and `GetLastTaskSession`'s SQL comment is explicit: *"We accept 'completed', 'failed' AND 'cancelled' tasks"*.

So: a run creates the empty file, gets killed by a timeout or daemon restart before writing its first event, and the next task receives a 0-byte `S.jsonl`. The gate calls it reachable, the agent is told the conversation continues, and **no** continuity notice fires. That's the same class of failure your missing-file branch carefully avoids, just in a different shape.

An `info.Size() > 0` condition closes it and reuses the disclosure path you already wired up.

**2. The provider family is hardcoded, against the convention in the same file**

`provider == "pi" || provider == "omp"` is a hand-maintained family list. It's correct today, but ~90 lines up in the same file the codebase does the opposite:

```go
func init() {
	// Seed built-in runtime identity display names from the descriptor so
	// adding a new fork doesn't require editing this map by hand.
	for _, desc := range agent.BuiltinRuntimes {
		runtimeDisplayNameOverrides[desc.ID] = desc.DisplayName
	}
}
```

And `builtin_runtimes.go:85` says *"The **first** entry is omp"* — a second pi-family fork is anticipated. When it lands, it silently falls back to workdir gating, this exact bug reappears, and no test goes red.

Deriving the set from `agent.BuiltinRuntimes` where `desc.ProtocolFamily == "pi"` is structurally identical to what's already there.

---

## Nits

- `reachable` is computed from the workdir rule and then wholly overwritten by the pi branch. Minor wasted work, but mainly it reads as though pi participates in the workdir decision. A `switch` or early return would make "the pi family uses a different predicate" visible in the structure.
- `TestGateResumeToReachableSession` still names its variable `reused` (`daemon_test.go:2042`).
- No coverage for a session path pointing at a **directory** — the `IsRegular()` branch is currently unverified.
- No coverage for the concurrency case above. Worth calling out on its own: the three new tests are all single-run, so they read as thorough while leaving the one real risk untested.

---

## Open design question

The PR correctly separates two things the old code conflated — "the transcript is gone" and "the workdir changed" — but only keeps the first one observable. After this change an agent resumes with memories of files at paths that no longer exist, in a freshly prepared workdir, and nothing tells it the ground moved.

It's usually self-correcting (a failed read leads to a re-scan), but it costs a turn and can produce confidently wrong statements about earlier work. Should a resumed-but-relocated run get a short "conversation continues, working directory changed" note? I lean yes — once the two facts are decoupled, the second one has no owner. Happy to hear the counter-argument; it doesn't have to be in this PR, but it shouldn't be lost.

---

## Verification performed

| Check | Result |
| --- | --- |
| `go vet ./internal/daemon/` | pass |
| `go test ./internal/daemon -run 'TestGate(ResumeToReachableSession\|PiResume)' -count=1` | pass |
| `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` | pass |
| `go test ./internal/daemon -count=1` | 7 failures — **pre-existing** |

On the last row: `TestEnsureDaemonID_*`, `TestPrepare{Reasonix,Dsh}Task*` and `TestLoadConfig_BackendOverrides_*` fail with an identical set on `main` in my environment (a config dir made them read real paths instead of temp dirs). Not caused by this PR — no regressions in that package.

I also confirmed the prompt-cache prefix invariant is intact: the `execenv.go` change is comment-only, the brief rendering path is untouched, and `PriorSessionResumed` / `PriorSessionResumeUnavailable` still reach the model through the per-turn prompt (`prompt.go:44`, `prompt.go:67`) rather than the brief.

The direction here is right and the diff is small — fix the concurrency guarantee and this is close.

mameikagou:
Addressed the review in `bea67bdb5`.

- Added a real cross-process exclusive lock on the Pi/OMP JSONL itself, held from before child launch through `Wait()`. Both Unix (`flock`) and Windows (`LockFileEx`) are covered. A concurrent resume now returns structured `ResumeRejected`, so the daemon uses its existing one-shot fresh-session retry and continuity-disclosure path instead of allowing a second writer.
- Release now happens before the terminal result is published, avoiding a follow-up task spuriously seeing a finished run as busy.
- A Pi session is reachable only when it is a non-empty regular file (`Size() > 0`); missing, empty, and directory cases share the existing disclosure path.
- Pi-family detection now derives built-in runtime identities from `agent.BuiltinRuntimes` / `ProtocolFamily == "pi"`; no hand-maintained `pi || omp` list remains.
- Reworked the gate structure so the Pi predicate is visibly separate, renamed the leftover `reused` local, and added coverage for descriptor-derived families, directories, empty files, concurrent writers, and post-run lock release.

I left the resumed-but-relocated informational note out of this PR: it is useful, but it is independent of session correctness and deserves a separately worded surface-aware change rather than expanding this recovery fix.

Verification:

- `go test ./internal/daemon -run "Test(GateResumeToReachableSession|GatePiResume|ProviderUsesPiSessionFile)" -count=1`
- `go test ./pkg/agent -run "^(TestPi|TestBuildPi|TestFilterPi|TestOmp)" -count=1`
- `go test -race ./pkg/agent -run "TestPiExecuteRejectsConcurrentSessionWriter|TestPiExecuteAttachesStdinPipe" -count=1`
- Windows cross-compile of `./pkg/agent`
- `go vet ./internal/daemon/ ./pkg/agent/`

The full daemon suite still has the same seven local-config-path failures listed in the review. The full agent suite also hits two unrelated Codex cleanup fixture failures on this machine; all Pi/OMP coverage passes. Ready for re-review.

Bohan-J (changes requested):
Re-reviewed at `bea67bdb5`. **Every point from the previous round is addressed**, and the fix for the blocking one is better than what I suggested. Requesting changes again only because that fix has a downstream consequence neither of us accounted for.

## What's resolved

| Previous item | Status |
| --- | --- |
| Blocking: concurrent writers on one JSONL | Fixed — real exclusive lock |
| Existence-only check on the session file | Fixed — `info.Size() > 0` |
| Hardcoded `"pi" \|\| "omp"` family list | Fixed — `providerUsesPiSessionFile` derives from `agent.BuiltinRuntimeByID(...).ProtocolFamily` |
| `reachable` computed then overwritten | Fixed — if/else, pi no longer runs the workdir comparison |
| Missing coverage: directory path, empty file | Fixed — `TestGatePiResumeDropsUnusableSessionFile` |
| Missing coverage: concurrency | Fixed — `TestPiExecuteRejectsConcurrentSessionWriter`, plus an assertion that the lock is released after a normal run |

The locking work deserves specific credit, because the easy version of it would have been silently broken:

- `unix.Flock` binds the lock to the **open file description**. The daemon runs tasks as goroutines in one process, so two tasks each `os.OpenFile` the same path and get distinct descriptions — the second `LOCK_EX|LOCK_NB` correctly fails. Had this used `fcntl`/`F_SETLK` (per-process), the second acquisition would have **succeeded** and the guarantee would have been decorative. Getting this right is the whole ballgame here.
- `LockFileEx` + `LOCKFILE_FAIL_IMMEDIATELY` is the correct Windows analogue, and it cross-compiles (`GOOS=windows go build ./pkg/agent/`).
- Releasing explicitly *before* the `resCh` send rather than relying on the defer is right, and the comment explains why (buffered channel, receiver races ahead). I checked that the goroutine has exactly one send path and no early return that bypasses it.

---

## New blocking issue: a busy-lock retry permanently retires the other run's healthy session

Routing "lock is busy" into the existing `ResumeRejected` fallback reuses machinery instead of inventing new machinery, which is the right instinct. The problem is what `ResumeRejected` is wired to downstream: **permanent retirement**. "Currently busy" is transient.

Step by step:

1. Task A is running and holds the `flock` on `S.jsonl`.
2. The user sends a follow-up before A finishes; task B is claimed with `PriorSessionID = S.jsonl`.
3. The gate passes it (file exists, regular, non-empty), so B calls `Execute` with `ResumeSessionID = S`.
4. `tryLockPiSessionFile` fails → `piSessionBusyResult` → `Result{Status: "failed", ResumeRejected: true}`, with `tools == 0`.
5. `shouldRetryWithFreshSession` (`daemon.go:8045`) matches on all four conditions.
6. The retry block's first statement is `retiredSessionID = task.PriorSessionID` (`daemon.go:7717`).
7. B retries on a fresh session, succeeds, and reports `retired_session_id = S`.
8. The `retired_sessions` CTE in `GetLastTaskSession` / `GetLastChatTaskSession` now filters S out. The query comment states the intent plainly:

   > `retired_session_id` records the abandonment itself, so one row retiring a session **removes it from every later lookup no matter how many clean rows still reference it.**

9. A completes normally and records `session_id = S` — a complete, healthy transcript that is now unreachable forever.

A PR whose purpose is preserving conversation history ends up destroying it, irreversibly. The original bug cost one turn of context; this costs the entire conversation.

This isn't a narrow race. The lock is held for the full child-process lifetime, so "user sends a second message while the agent is still working" hits it every time — which is the same Pi chat scenario the PR description opens with.

For contrast, the pre-change behaviour never retired anything here: the old gate cleared `task.PriorSessionID` **before** `Execute`, so `shouldRetryWithFreshSession` saw an empty prior id and returned false. This is new in the second commit; neither `main` nor the first commit had it.

**Suggested fix:** keep busy out of the retirement path. The smallest version is a separate signal — `Result.ResumeBusy`, or `ResumeRejectedTransient` — that still triggers the fresh-session fallback and the continuity notice, but skips the `retiredSessionID` assignment in step 6.

The three existing `ResumeRejected` causes (transcript gone, session belongs to another account, history no longer replayable) are all permanent properties of the session; being momentarily locked is not. Worth noting that this PR had to widen that field's doc comment to admit the new cause — that edit is itself the tell that it doesn't belong in the same bucket.

**Test gap for the fix:** the new tests are all at the backend layer and stop at "returns `ResumeRejected`". Nothing asserts what the daemon then does with it. A test that a successful busy-retry leaves `TaskResult.RetiredSessionID` empty would pin the actual invariant.

---

## Still open from the previous round

The question about disclosure when the conversation resumes into a **different workdir** hasn't been answered. This change correctly decouples "transcript is gone" from "workdir changed" but leaves the second fact unobservable, so a resumed agent can reference paths that no longer exist. It doesn't have to be solved in this PR — I'd just like it acknowledged rather than dropped.

---

## Verification at `bea67bdb5`

| Check | Result |
| --- | --- |
| `go vet ./internal/daemon/ ./pkg/agent/` | pass |
| `GOOS=windows GOARCH=amd64 go build ./pkg/agent/` | pass |
| `go test ./pkg/agent -run TestPiExecute -count=1` | pass |
| `go test ./internal/daemon -run 'TestGate\|TestProviderUsesPiSessionFile' -count=1` | pass |
| `go test ./internal/daemon/execenv -run ByteIdentical -count=1` | pass |

The prompt-cache prefix invariant still holds — `execenv.go` remains a one-line comment change and the brief rendering path is untouched.

Fix the retirement path and I think this is done. The hard part is already behind you.

mameikagou:
Addressed the transient-retirement blocker in `c1703ab0`.

- Split permanent resume rejection from temporary ownership with `Result.ResumeRejectedTransient`.
- A busy Pi/OMP transcript still takes the existing one-shot fresh-session fallback and continuity-disclosure path, but no longer sets `retired_session_id`.
- Added a daemon-level regression that holds the original JSONL lock, runs the real `runTask` fallback to a fresh Pi session, and asserts the healthy original session is not retired.
- Updated the backend test to require the transient signal and reject the permanent one.

Verification:

- `go test ./internal/daemon -run "TestRunTaskPiBusyRetryDoesNotRetireHealthySession|TestShouldRetryWithFreshSession" -count=1`
- `go test ./pkg/agent -run "TestPiExecuteRejectsConcurrentSessionWriter|TestPiExecuteAttachesStdinPipe" -count=1`
- `go test -race ./pkg/agent -run "TestPiExecuteRejectsConcurrentSessionWriter|TestPiExecuteAttachesStdinPipe" -count=1`
- `go test ./internal/daemon ./pkg/agent`
- `go vet ./internal/daemon/ ./pkg/agent/`

The relocated-workdir disclosure remains separate from this correctness fix; I will follow it in a standalone PR.

mameikagou:
Opened the separate relocated-workdir disclosure follow-up as #7783: https://github.com/multica-ai/multica/pull/7783. It is based on current `main`, contains no #7760 code, and merge-tree verification confirms the two branches combine without conflicts.

mameikagou:
@Bohan-J the transient-retirement blocker is fixed in `c1703ab0`, with a daemon-level regression covering the full busy-lock → fresh-session retry path and asserting the healthy original session is not retired. All CI checks are green now. Could you take another look when you have a chance?

Bohan-J (approved):
Approving — thanks for the thorough work on this, and for the patience across three rounds.

What stands out in hindsight is that the original insight was right from the first commit: the resume gate was justified by cwd-keyed session stores, and Pi's transcript was never keyed that way, so the rule was misapplied rather than merely inconvenient. Renaming it to `gateResumeToReachableSession` made the predicate say what it had always meant.

The two follow-ups were the hard part and you landed both:

- **The lock.** Using `flock` (per open file description) rather than `fcntl` record locking is what makes it actually hold inside a single daemon process running tasks as goroutines. The `fcntl` version would have looked identical and been silently useless. Holding it across the full child lifetime and releasing before the buffered result send — with a comment explaining why the defer alone would race — is careful work.
- **`ResumeRejectedTransient`.** Separating "temporarily owned" from "permanently unusable" was the right shape, and reverting the `ResumeRejected` doc comment to `permanently refused` rather than leaving it widened kept the distinction honest.

`pi_busy_retry_unix_test.go` deserves a specific mention: holding a real `flock`, driving the actual `runTask`, and asserting `RetiredSessionID` stays empty pins the invariant end to end rather than restating the implementation. That is the test that would have caught the round-two regression.

Also appreciated that the relocated-workdir disclosure went to #7783 as a separate, `main`-based PR instead of being folded in here. Correct scoping — it is a product decision, not part of this fix — and it will get its own review.

Two things noted for follow-up, neither blocking this merge:

1. `~/.multica/pi-sessions` has no retention policy at all, while Codex session stores, Hermes session stores and Hermes memory stores each have one. That was harmless when the files were unreachable after a workdir change; now that they are the resume mechanism, unbounded growth is worth a tracked issue. Not this PR's debt — this PR just made it matter.
2. In `daemon.go`, the guard reads `if !result.ResumeRejectedTransient`, so a backend violating the documented mutual exclusion by setting both flags would silently skip retirement. Inverting it is not a one-liner (the `UnresumableHistory` and undetectable branches retire while `ResumeRejected` is false), and the contract is documented and asserted in `pi_test.go`, so it is fine as-is — just a note for whoever adds the second producer of this field.

Verified locally at `c1703ab07`: `go vet`, Windows cross-build, the daemon and agent suites, and the `ByteIdentical` prompt-cache tests all pass. CI is 14/14 green, including `windows-execenv` on a real Windows runner, so the `LockFileEx` path has genuine coverage.

Merging now.

PR #7761, opened 2026-08-29
MUL-6841: feat(daemon): add native Codex working directory

mameikagou:
## What does this PR do?

Adds an opt-in machine-local Codex working directory (`codex_native_workdir` / `--codex-native-workdir` / `MULTICA_CODEX_NATIVE_WORKDIR`). Codex can run from a stable user directory while Multica keeps task logs, credentials, generated context, and `CODEX_HOME` in the isolated task env root.

The default remains unchanged. An explicitly assigned project `local_directory` still takes precedence.

This is intended for users who want Multica to remain the transport/task gateway while Codex keeps its native stable-cwd workflow. In native mode Multica does not write `AGENTS.md`, `.agent_context`, task markers, skills, or cleanup sidecars into the selected directory. The runtime brief is sent via Codex app-server `developerInstructions` instead.

Existing conversations can migrate safely: when a prior managed task workdir is proven to belong to the same agent and issue/chat, only the requested rollout is linked into the persistent per-conversation Codex session store. The existing rollout-presence gate remains authoritative, so a missing rollout is never presented as a successful resume.

## Related Issue

No linked issue.

## Type of Change

## Changes Made

- Add native Codex cwd configuration to the CLI config, daemon flags, env overrides, and startup validation.
- Keep native cwd preparation and cleanup sidecar-free while retaining a task-private env root and `CODEX_HOME`.
- Deliver the runtime brief through `developerInstructions` on both `thread/start` and `thread/resume` only when supplied; managed mode remains `nil`.
- Preserve Codex sessions across both task IDs and a one-time managed-workdir-to-native-cwd migration.
- Add coverage for configuration precedence, flag forwarding, cwd ownership, rollout migration, resume gating, and app-server parameters.

## How to Test

1. Configure a writable directory: `multica config set codex_native_workdir /absolute/path/to/code`.
2. Start/restart the daemon and run a Codex chat; verify the reported cwd is the configured directory and no Multica sidecars are created there.
3. Continue a chat whose prior turn used a managed task workdir; verify the same Codex thread resumes after the cwd change.
4. Clear the setting and verify managed per-task workdirs retain their previous behavior.

Automated verification run locally:

- `go test -p=1 ./internal/daemon ./internal/daemon/execenv`
- `go test ./pkg/agent`
- full `cmd/multica` test binary (run outside an existing workspaces-root marker)
- `go vet ./internal/daemon ./internal/daemon/execenv ./pkg/agent ./cmd/multica`
- `go build ./cmd/multica`

## Risks

- Native mode deliberately gives Codex the normal filesystem authority of the configured cwd. It is opt-in and Codex-only.
- A project `local_directory` remains higher priority, preserving explicit project scoping.
- Session migration accepts only an exact prior session rollout after managed provenance, task marker, directory shape, and conversation identity checks.

## Checklist

- [x] If this PR touches Chinese product copy, I checked it against the convention (not applicable)

## AI Disclosure

**AI tool used:** Codex via Multica

**Prompt / approach:** Implemented from a stable-cwd gateway use case. Traced daemon workdir selection, execenv sidecar lifecycle, Codex app-server start/resume parameters, and rollout storage before adding an opt-in path. Kept the upstream default unchanged and verified both fresh and migration paths.

### DIFF
--- a/server/cmd/multica/cmd_config.go
+++ b/server/cmd/multica/cmd_config.go
@@ -35,6 +35,7 @@ var configSetSupportedKeys = []string{
 	"device_name",
 	"runtime_name",
 	"workspaces_root",
+	"codex_native_workdir",
 	"max_concurrent_tasks",
 	"poll_interval",
 	"heartbeat_interval",
@@ -51,11 +52,11 @@ var configSetCmd = &cobra.Command{
 	Short: "Set a CLI configuration value",
 	Long: "Supported keys: " +
 		"server_url, app_url, workspace_id, " +
-		"device_name, runtime_name, workspaces_root, max_concurrent_tasks, poll_interval, " +
+		"device_name, runtime_name, workspaces_root, codex_native_workdir, max_concurrent_tasks, poll_interval, " +
 		"heartbeat_interval, agent_timeout, " +
 		"codex_semantic_inactivity_timeout, codex_handshake_timeout, " +
 		"disable_auto_update, auto_update_check_interval, disable_auto_reload.\n\n" +
-		"The daemon keys (device_name, runtime_name, workspaces_root, max_concurrent_tasks, " +
+		"The daemon keys (device_name, runtime_name, workspaces_root, codex_native_workdir, max_concurrent_tasks, " +
 		"poll_interval, heartbeat_interval, agent_timeout, " +
 		"codex_semantic_inactivity_timeout, codex_handshake_timeout, " +
 		"disable_auto_update, auto_update_check_interval, disable_auto_reload) mirror their " +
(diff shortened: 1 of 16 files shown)


multica-eve:
Today, users choose an existing local folder for a specific project, making the scope and ownership of that folder explicit. This change would instead make one folder a machine-wide fallback for every eligible task that lacks such a project choice. Tasks from different projects, agents, and workspaces could therefore share and modify the same files, while that rule remains invisible from the project itself.

## Product decision: Reject

We do not want to add this behavior in its current form:

- The need has not been established by a linked issue or a verified user journey. Continued conversations already reuse their managed working directory, and users who need an agent to operate in an existing checkout can attach a project-scoped local directory.
- This introduces a second, provider-specific source of truth beside project resources. A machine-wide fallback is too broad: it applies independently of which project, agent, or workspace dispatched the task. If the existing local-directory flow is inconvenient, we should improve that project-scoped flow instead of bypassing it.
- There is also a concrete shared-write risk. The native directory is selected whenever the provider matches and no project local directory is present (`server/internal/daemon/daemon.go:6855`), but it does not participate in the local-directory path lock (`server/internal/daemon/daemon.go:5313`). Since a daemon can run multiple tasks concurrently, unrelated tasks can modify the same directory at the same time.
- The setting is exposed only through daemon configuration, a flag, or an environment variable; there is no Desktop/Web surface or documentation showing users that all eligible tasks on that machine will inherit it.

This is a product rejection rather than a request to keep iterating on the current implementation. If there is a distinct stable-directory use case that the existing conversation reuse and project-local-directory flows do not cover, please open an issue describing that user journey first. The likely product direction would remain explicitly project-scoped, with visible ownership and defined concurrency semantics.

Thank you for the thoughtful implementation and test coverage.

PR #7765, opened 2026-08-30
MUL-6790 feat(issues): support operator matching in scalar property filters

### LINKED ISSUE #7692: [Feature]: Scalar property filter: matching beyond exact equality (contains / range)
### Background

Follow-up from #7240 (merged). Raised during review as product question #1 ("Exact equality only") and deliberately deferred as a documented v1 trade-off.

The custom-property filter for `text` / `number` / `date` / `url` (shipped in #7240) matches by exact equality only:

- For `text` / `url`, users will almost certainly type a fragment and get an empty list with no explanation of why.
- For `number` and `date`, the natural ask is a range (`> 3`, "before a date"), which equality cannot express.

### Current behavior

- Server: `parsePropertiesFilterParam` (`server/internal/handler/property.go`) expands a typed value into every containment form it could match, but all forms are exact-equality containment (`->>` string equality, `::numeric` equality, date-only string equality).
- Client: `issueMatchesPropertyFilters` (`packages/views/issues/utils/filter.ts`) mirrors the same semantics.
- UI: the filter menu (`packages/views/issues/components/issues-header.tsx`) takes a single value; there is no operator selection.

### Goal

Let users filter scalar custom properties with more than exact equality — at minimum:

- `text` / `url`: **contains**
- `number`: comparison operators (`>`, `<`, and ideally `>=` / `<=` or between)
- `date`: **before / after** (ideally between)

Exact equality stays the default so existing saved views and URLs keep their meaning.

### Proposed work

- Decide the operator set per type and the param encoding for saved views / URLs. Must stay backward-compatible: an existing bare value means equality.
- Server: extend `parsePropertiesFilterParam` and the SQL compilation for the new operators.
- Client: mirror the new operators in `issueMatchesPropertyFilters`.
- UI: operator selection in the filter menu, and chips rendering that shows the active operator.

### Acceptance criteria

- [ ] `text` / `url` properties can be filtered by "contains".
- [ ] `number` properties can be filtered by comparison (`>` / `<`).
- [ ] `date` properties can be filtered by before / after.
- [ ] Existing saved views / URLs carrying bare values keep exact-equality semantics.
- [ ] Facet counts keep working (the `__set__` / `__none__` buckets are operator-independent).
- [ ] Round-trips through saved views.

### Related

- #7240 (merged) — scalar property filter, where this was deferred.

deoooo:
Closes #7692

## Summary

Scalar custom-property filters (`text` / `number` / `date` / `url`) previously matched by exact equality only — a user typing a text fragment got an empty list, and a numeric or date range could not be expressed. A property-filter member is now **either** a plain string **or** an operator object:

| Type | Operators | Match semantics |
| --- | --- | --- |
| `text` / `url` | `contains` | case-insensitive substring over the value's text form |
| `number` | `gt` / `gte` / `lt` / `lte` | numeric comparison against stored jsonb numbers |
| `date` | `before` / `after` | lexicographic comparison of the stored `YYYY-MM-DD` strings |

**Backward compatibility:** a bare string member still means exact equality — the default — so every existing saved view, URL, and API call keeps its meaning. The menu's default "is" commits a bare string; only an explicit operator pick commits `{op, value}`.

## What changed

**Backend**
- `parsePropertiesFilterParam` (`server/internal/handler/property.go`) now parses members as raw JSON: strings keep the legacy expansion; `{"op","value"}` objects validate against the per-op rules (finite number for comparisons, `YYYY-MM-DD` for before/after, length-bounded and LIKE-escaped value for `contains`) and compile to `{"__op__","def","value"}` patterns. The marker keys never collide with a UUID-keyed containment pattern or the `__none__` marker.
- `propertiesFilterPredicate` renders operator alternatives as typed comparisons (`ILIKE '%…%'` / `::numeric > $n` / string compare); the static `ListOpenIssues` unroll in `issue.sql` mirrors the same branches, and the generated sqlc code is regenerated. Operators are unindexed (like the `__none__` key-absence check) — scalar ranges cannot use the containment GIN index.
- `IssueTableFilters.properties` carries raw JSON members so both shapes survive the round-trip; the query fingerprint dedupes/orders them byte-exactly.
- Facet buckets (`__set__` / `__none__`) are operator-independent and unchanged.

**Frontend**
- Core types add `PropertyFilterOp` / `PropertyOperatorFilter` / `PropertyFilterValue` and `PROPERTY_FILTER_OPS_BY_TYPE`.
- The filter menu (`issues-header.tsx`) renders an operator row above the scalar input; picking an op with a draft commits immediately, and picking before typing carries the choice through the next Enter/blur commit. Saved-view-locked dimensions disable the row.
- `issueMatchesPropertyFilters` mirrors the server predicates per op (case-insensitive `contains`, numeric-only comparisons, string-only before/after, missing key never matches).
- Chips summarize operator members via locale phrases (`contains {{value}}`…) or locale-independent math symbols; all four locales updated.
- Saved-view baselines accept operator members through the same sanitization rules as other filter groups and key them via `propertyFilterValueKey`.

## Verification

- Go: full server suite with `-race` (two runs; DB-backed) plus new `TestParsePropertiesFilterOperatorUnit` (compilation, SQL rendering, bind types, 400 matrix) and `TestListIssuesPropertyFilterOperators` (dynamic + static unroll agreement, `%`-literal needles, case-insensitivity, OR with `__none__`, AND across definitions, malformed-member 400s).
- TS: `pnpm typecheck` clean; core + views Vitest suites pass, with a new operator matrix in `filter.test.ts`, menu-operator interaction tests in `issues-header.property-filter.test.tsx`, and saved-view baseline sanitization tests.


deoooo:
@Bohan-J This closes #7692 (operator matching for scalar property filters: `contains` for text/url, `gt/gte/lt/lte` for number, `before/after` for date; bare values keep exact-equality semantics so existing saved views and URLs are unaffected). It's ready for review — all CI checks pass, and the full server suite (`-race`) plus `pnpm typecheck` and the core/views Vitest suites are green locally. Happy to adjust anything.

Bohan-J (changes requested):
Thanks for the follow-up on #7692 — this is careful work. The three serving paths (dynamic `ListIssues` predicate, the static `open_only` unroll, and the table-rows endpoint) are all covered, the cross-path agreement test even pins the hex-float case where only the static `::numeric` cast would have blown up, and keeping a bare string as exact equality is the right call for saved views and shared URLs.

One blocking issue, three decisions I'd like to settle before merge, and a handful of non-blocking nits.

## Blocking

### `contains` means different things on the server and on the client

The server predicate is `(i.properties ->> def) ILIKE '%…%'`, and `->>` yields a text form for *any* jsonb type, not just strings. Verified on PG 16:

```sql
SELECT ('{"d":["opt-alpha","opt-beta"]}'::jsonb ->> 'd')            AS array_text,
       ('{"d":["opt-alpha"]}'::jsonb ->> 'd') ILIKE '%alpha%'       AS array_matches,
       ('{"d":3.50}'::jsonb ->> 'd')                                AS number_text;
-- array_text  | ["opt-alpha", "opt-beta"]
-- array_matches | t
-- number_text | 3.50
```

So on the server a `contains` member against a `multi_select` property matches the JSON serialization of the option-id array, and against a `number` property it matches `"3.50"`.

`issueValueMatchesOperator` in `packages/views/issues/utils/filter.ts` does neither: arrays return `false` outright, and numbers go through `String(3.5)` → `"3.5"`. The result is that the server returns rows the client-side instant filter then hides again — the rendered list and the total disagree.

The menu can't produce this combination today (`PROPERTY_FILTER_OPS_BY_TYPE` only offers `contains` for `text` / `url`), but the API, a hand-edited URL, and any future change that opens `contains` to another type all hit it — and the PR's own contract is that the matcher mirrors the server per op.

Smallest fix, and the one that also matches what the menu intends: constrain `contains` to strings on both sides.

- `server/internal/handler/property.go` — add a `jsonb_typeof(i.properties -> def) = 'string'` guard to the `contains` branch of `operatorPatternPredicate`.
- `server/pkg/db/queries/issue.sql` — same guard in the static unroll's `contains` branch (plus regenerated sqlc).
- `packages/views/issues/utils/filter.ts` — drop the `number` / `boolean` `String(value)` branch.

`text` and `url` only ever store strings, so nothing intended is lost. Worth adding a case for it too: neither the Go nor the TS suite currently exercises `contains` against a non-string stored value, which is exactly why the divergence went unnoticed.

## Decisions to settle before merge

### 1. Should `contains` be the default operator for `text` / `url`?

The first pain point listed in #7692 is that users type a fragment and get an empty list with no explanation. This PR makes `contains` reachable, but leaves `is` as the default — so for anyone who doesn't notice the new operator row, the original complaint still stands.

These are two separable things: the **wire format** must stay backward compatible (a bare string means equality — keep that exactly as it is), but the **menu's preselected operator** is a pure UI decision. The menu could default `text` / `url` to `contains` and commit `{op:"contains", value}` explicitly, while bare strings continue to mean equality, leaving every existing URL and saved view untouched.

I'd lean toward defaulting to `contains` for `text` / `url`. Happy to hear the argument for keeping `is`, but it should be a deliberate choice rather than a side effect of the compatibility requirement.

### 2. `date` has no inclusive comparison, `number` does

`number` gets `gt` / `gte` / `lt` / `lte`, but `date` only gets strict `before` / `after`. In everyday use "delivered before September 1" usually includes September 1, and there's no way to express that today (equality ORs rather than ANDs, so you can't compose it either).

Either add `on_or_before` / `on_or_after`, or state why the strict-only set is the right first cut. Right now the asymmetry has no stated rationale.

### 3. Where does `between` land?

#7692 lists `between` as the ideal. Because members OR within a definition, `> 3 AND < 10` isn't expressible in this model, and the scalar menu keeps a single value member — so the "range" half of the issue is only partly delivered. Open-ended comparison covers a lot in practice, so this doesn't have to ship here, but I'd like a decision on record: either pin the shape now (`{op:"between", value:[a,b]}` or similar) or explicitly scope this to open-ended comparison.

Good news is the `{op, value}` union leaves room for a compound operator later without another compatibility break.

## Non-blocking

1. **Numeric bound binds differently on the two paths.** The dynamic predicate binds the bound as a float8 (`addArg(num)`), while the static unroll compares `numeric` against `(alt.pattern->>'value')::numeric`. `numeric > float8` demotes the column value to float8, so the two paths can disagree beyond ~17 significant digits. The agreement test only uses short decimals. Binding the canonical string and casting it (`$n::numeric`) would make them identical.

2. **`jsonb_typeof(...) = 'number' AND (...)::numeric` relies on AND evaluation order,** which SQL doesn't guarantee — the planner is free to reorder quals. I checked this on PG 16 and it does *not* reproduce on either path: the static unroll sits inside an OR branch, so the executor short-circuits in written order, and in the dynamic predicate the planner puts the cheaper `jsonb_typeof` first even when the cast is written ahead of it (the cast alone does raise `invalid input syntax for type numeric`, so the guard is genuinely doing the work). Not a bug, then — but it's a guarantee we don't control. `propertySortExpr` in the same file already solves this with `CASE WHEN … THEN (…)::numeric END`; matching that idiom turns "works today" into "works by construction" for one line.

3. **The operator → symbol map (`> ≥ < ≤`) is duplicated** in `issues-header.tsx` and `filter-chips-bar.tsx`. Worth exporting once from core next to `PROPERTY_FILTER_OPS_BY_TYPE`.

4. **`PROPERTY_FILTER_OPS_BY_TYPE: Record<string, PropertyFilterOp[]>`** keys on `string` rather than `IssuePropertyType`, so adding a scalar type later won't produce a compile-time reminder.

5. **The operator row is a mutually exclusive single-select rendered with `aria-pressed`.** A radiogroup (`role="radio"` / `aria-checked`) is the more accurate semantic, and the group has no accessible label.

## Operational note (not a change request)

Rolling this back isn't clean: after a revert, a saved view or persisted local filter state holding an operator member makes the older server reject the whole request with a 400 (it decodes members as `[]string`), so the issue list errors out rather than the filter silently degrading. Same reason the server needs to ship ahead of the desktop client. Nothing to change in the PR — just worth knowing at merge time.

---

To be clear about the ordering: the `contains` divergence is the only thing I'd call required. The three decisions above are mostly on us, and I'll follow up on 1–3 so you're not blocked guessing.

Bohan-J (approved):
Thanks for the thorough follow-through here — this landed well.

Everything from the last round is addressed: `contains` is now restricted to stored strings on all three serving paths (dynamic predicate, the `open_only` static unroll, and the client matcher), with cases covering the number / boolean / multi_select values that must *not* match. You also picked up every non-blocking note — the numeric bound now compares as `numeric` on both paths, the type guard uses `CASE` so it holds by construction rather than by planner behaviour, the operator symbol map is shared, the type map is tightened, and the operator row became a labelled radiogroup.

Two things worth calling out as good practice, not just done-ness: the cross-path agreement test that pins the hex-float case caught a class of bug that only ever shows up on one of the two SQL paths, and treating the empty needle as an explicit non-match rather than letting it degrade into match-all is the kind of detail that usually gets found in production instead.

I pushed one small commit on top (`9b5bca4f0`): the Japanese `before` / `after` copy used 以前 / 以後, which include the boundary date, while the operators compare strictly. Switched to より前 / より後 so the wording matches the behaviour. Korean 이전 / 이후 has the same literal issue but is also the conventional rendering there, so I left it for a native speaker to decide rather than guess — reasoning is in the commit message.

The three product questions from the last round were ours to settle, and the outcome is: exact equality stays the default, and `between` is out of scope for this PR — the `{op, value}` shape leaves room for it later without another compatibility break.

Merging. Thanks again for closing the loop on #7692.

PR #7767, opened 2026-08-30
MUL-6789: feat(issues): list observed values in scalar property filter facets

### LINKED ISSUE #7693: [Feature]: List observed values in scalar property filter menus (text / url / number / date)
### Background

Follow-up from #7240 (merged). Raised during review as product question #2 ("The input starts empty") and deferred as a documented v1 trade-off.

For `select` properties, the filter menu lists the options as checkboxes, so users can discover which values exist. For the scalar types shipped in #7240 (`text` / `number` / `date` / `url`), the menu shows only a bare input — users cannot discover which values a property actually holds and must guess exact values to get any results.

### Why the data was dropped

#7240 originally grouped the scalar facet by the raw property value, but that had no upper bound: a free-form `text` property on a large workspace produced one facet row per distinct value, all serialized and discarded except `__none__`. To fix that, the facet was bounded to the two buckets the menu actually reads (`server/internal/handler/issue_table_facets.go`, the `text`/`url`/`date`/`number` branch: `CASE WHEN i.properties ? <key> THEN '__set__' ELSE '__none__' END`), which discarded per-value data.

Listing observed values needs that data back — bounded.

### Proposed work

- Server: return bounded per-value facets for scalar types — group by value, `ORDER BY count DESC LIMIT n`, and always union the `__none__` bucket on top (the shape suggested during the #7240 review).
  - If `number` values are listed, define a canonicalization rule first: the facet previously keyed by `->>`, i.e. the stored jsonb literal text, so `3.50` produced a different key than `3.5`.
- Frontend: for `text` / `url` (and possibly number / date), render observed values as checkboxes alongside the free input, the way select properties work. Keep the input for values not in the list.
- Decide the `LIMIT n` and the behavior when values exceed it (the input still covers unlisted values).

### Acceptance criteria

- [ ] Opening a `text` / `url` property filter submenu shows observed values with counts, as checkboxes.
- [ ] The "No value" count is still shown (`__none__`).
- [ ] Typing a value that is not listed still works.
- [ ] The facet response stays bounded (no one-row-per-distinct-value blowup on large workspaces).
- [ ] Saved views / URL round-trips unchanged.

### Related

- #7240 (merged) — scalar property filter, where the facet was bucketed and this was deferred.

deoooo:
Closes #7693

## Summary

For the scalar property types (`text` / `url` / `number` / `date`), the filter menu previously showed only a bare input — users could not discover which values a property actually holds. The scalar facet now returns **bounded per-value rows** and the menu renders them as checkboxes with counts, the way `select` properties work.

## What changed

**Backend** (`server/internal/handler/issue_table_facets.go`)
- The scalar facet replaces the two-bucket `__set__`/`__none__` split with:
  - the top **50** observed values by count (`scalarPropertyFacetLimit`, matched to `maxPropertySelectOptions`), ties broken by key so the result is deterministic;
  - the `__none__` bucket **always unioned on top**, so "No value" can never be crowded out by the limit and unlisted values stay reachable through the free input.
- **Number canonicalization** (the open question from the issue): facet keys go through `trim_scale((i.properties ->> k)::numeric)`. `::numeric` alone keeps the literal's scale (`3.50::text` is `"3.50"`), so `trim_scale` collapses stored jsonb literals `3.50` and `3.5` into one `3.5` key. The API cannot produce that case on its own (numbers round-trip through `float64`), which the test proves with a raw jsonb literal write.
- Text / url / date keys are the stored strings, grouped **exactly** — `Alpha` and `alpha` are distinct observed values with their own counts. Values whose jsonb shape does not fit the definition type are not listed; `__none__` keeps the exact key-absence predicate.
- The response stays bounded by construction: one grouped scan with `LIMIT` plus a single `__none__` count row.

**Frontend** (`packages/views/issues/components/issues-header.tsx`)
- `PropertyFilterOptions` renders the observed values as checkbox rows with counts between the value input and "No value". Toggling commits a **bare equality member** of the same OR-set — exactly the shape the free input commits — so picking an observed value and typing an unlisted one compose, and removal works through the same toggle.
- The client-side facet fallback (when server facets are unavailable) now counts `number` values under `String(value)`, which is already canonical in JS; text/url/date were already counted per raw value there.
- Legacy-backend safety: a new menu ignores the old `__set__` bucket key exactly the way an old client ignores the new per-value rows.

## Compatibility

- The facet response shape is unchanged (`{key, count}[]`); only the keys change for scalar types. Existing clients read only `__none__` for these types, so the extra rows are ignored.
- Saved views / URL round-trips are unchanged: observed values commit bare strings, the same member form that existed before this PR.

## Verification

- New DB-backed tests: `TestIssuePropertyFacetScalarTypes` (mixed-case exact keys with correct per-key counts, `3.50`/`3.5` canonicalization into one count-2 facet, date/url string keys, `__none__` counts) and `TestIssuePropertyFacetScalarBounded` (55 seeded distinct values → exactly 50 value rows + `__none__`, lowest-tie keys dropped, nothing leaks past the limit).
- New menu tests: observed text values render with counts and toggle as bare members; observed number values list under the canonical `3.5` key.
- Full Go suite with `-race`, `pnpm typecheck`, and the core/views Vitest suites pass.


deoooo:
@Bohan-J This closes #7693 (bounded per-value scalar facets: top 50 observed values by count, `__none__` always unioned on top, number keys canonicalized via `trim_scale` so 3.50/3.5 group together; the filter menu renders them as checkboxes). It's ready for review — all CI checks pass, and the full server suite (`-race`) plus `pnpm typecheck` and the core/views Vitest suites are green locally. Happy to adjust anything.

deoooo:
@Bohan-J @forrestchang Following up on #7767 / #7693. The latest head (`c7c075a9`) includes the sentinel-only facet fix, the 52-member filter cap, and the observed-value menu coverage. All required CI checks are green, including backend, frontend, sqlc, installer, and vulnerability checks.

Could you take another look when you have a chance? The branch is currently 88 commits behind `main`, so I’m aware it may need a refresh before merge; I wanted to get the latest implementation back on your review radar first.

PR #7775, opened 2026-08-30
MUL-6849 MUL-7549: fix(runtime): remove duplicate Runtime Brief context

### LINKED ISSUE #7773: Remove duplicate Runtime Brief content to reduce prompt tokens
## Problem

The Multica Runtime Brief contains prompt content that is already represented elsewhere in the model context or repeated by another Brief section:

1. Run-only Autopilot IDs, title, description, trigger source, and trigger payload are present in both the Runtime Brief and the per-turn user prompt.
2. A project GitHub repository is present in both `## Repositories` and `## Project Context → Project resources`.

These duplicates increase input tokens and make the cached prompt prefix larger than necessary. The Autopilot fields also make the Brief vary between runs.

## Goal

Give each piece of context one canonical prompt location while preserving task behavior:

- stable workflow and safety policy belong in the Runtime Brief
- per-run Autopilot data belongs in the per-turn user prompt
- rich project GitHub repository metadata belongs in Project Context

## Scope

- Remove per-run Autopilot fields from the Runtime Brief.
- Keep run ID, Autopilot ID, title, description, source, payload, and configuration lookup guidance in the per-turn prompt.
- Keep the stable Autopilot issue-command guard in the Runtime Brief.
- Do not render a repository again in `## Repositories` when the same URL is already represented by a `github_repo` project resource.
- Preserve unmatched workspace repositories.
- Do not change the runtime repository list used for checkout authorization or caching.
- Keep sidecar context files unchanged.

## Acceptance criteria

- The Autopilot Brief is byte-identical when per-run fields change.
- Autopilot dynamic fields remain present in the user-turn prompt.
- Each project GitHub repository URL appears only once in the Runtime Brief, with ref/default-branch/label metadata preserved in Project Context.
- Workspace repositories that are not represented by project resources still appear.
- Focused regression tests pass.

Bbbtt04:
## Summary

- keep only stable run-only workflow policy and the issue-command guard in the Autopilot Runtime Brief
- keep all per-run Autopilot IDs, instructions, source, and trigger payload in the current user-turn prompt
- render project GitHub repositories once in the richer Project Context section instead of repeating their URLs under `## Repositories`
- preserve unmatched workspace repos and the complete runtime repo list used for checkout authorization and caching
- add regression coverage for Brief byte stability, prompt completeness, project repo deduplication, and workspace repo fallback

## Why

The Runtime Brief is part of the model's prompt prefix. Repeating content there increases input tokens and makes the cached prefix larger than necessary. Per-run Autopilot data also makes the Brief vary between runs, while project GitHub resources were printed twice inside the same Brief.

This change gives each value one canonical prompt location without changing task behavior or repository access.

## Tests

- `go test ./internal/daemon/execenv -run 'Test(ProjectReposRenderedOnceInMetaSkill|ProjectContextKeepsUnmatchedWorkspaceRepoInMetaSkill|PrepareWithRepoContext|ChatProjectContextInjectedIntoRuntimeBrief|BriefByteIdenticalAcrossRunsForEveryKind|InjectRuntimeConfigAutopilotRunOnlyNoIssueWorkflow|BriefOwnsAutopilotIssueCommandsGuard)$' -count=1`
- `go test ./internal/daemon -run '^TestBuildPromptAutopilotRunOnly$' -count=1`

A full `go test ./internal/daemon/execenv ./internal/daemon` run was also attempted. It is blocked on this Windows host by unrelated environment-dependent tests (local CLI and skill discovery, filesystem locking, and long path behavior); the focused affected tests pass.

Closes #7773

### DIFF
--- a/server/internal/daemon/daemon_test.go
+++ b/server/internal/daemon/daemon_test.go
@@ -1093,6 +1093,10 @@ func TestBuildPromptAutopilotRunOnly(t *testing.T) {
 		AutopilotTitle:       "Daily dependency check",
 		AutopilotDescription: "Check dependencies and report outdated packages.",
 		AutopilotSource:      "manual",
+		AutopilotTriggerPayload: json.RawMessage(`{
+			"repository": "multica-ai/multica",
+			"action": "scheduled"
+		}`),
 	}, "claude")
 
 	for _, want := range []string{
@@ -1100,6 +1104,8 @@ func TestBuildPromptAutopilotRunOnly(t *testing.T) {
 		"Autopilot run ID: run-1",
 		"Daily dependency check",
 		"Check dependencies and report outdated packages.",
+		"Trigger source: manual",
+		`"repository": "multica-ai/multica"`,
 		"multica autopilot get autopilot-1 --output json",
 	} {
 		if !strings.Contains(prompt, want) {
--- a/server/internal/daemon/execenv/execenv_test.go
+++ b/server/internal/daemon/execenv/execenv_test.go
@@ -697,15 +697,16 @@ func TestChatProjectContextInjectedIntoRuntimeBrief(t *testing.T) {
 	}
 }
 
-// When the issue's project has its own github_repo resources, those should be
-// the only repos rendered in the meta-skill — workspace-level repos must not
-// leak into the agent prompt to avoid confusing it about which repo to use.
+// When the issue's project has its own github_repo resources, the claim lifts
+// them into Repos for checkout authorization as well as retaining their richer
+// ProjectResources representation. The brief must render each URL only once,
+// in Project Context, without changing the underlying runtime repo list.
 //
 // The handler-side override is exercised in handler tests; this test confirms
 // the rendering side: given a TaskContextForEnv where Repos was already
 // narrowed by the server to project repos only, the meta skill renders just
 // those.
-func TestProjectReposReplaceWorkspaceReposInMetaSkill(t *testing.T) {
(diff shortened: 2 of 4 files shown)


Bohan-J:
Thanks for digging into this, @Bbbtt04. The problem you identified is real, and the analysis in #7773 is clear. We're closing this PR because `main` has moved on since you opened it:

- **The Autopilot half has already landed in #7971.** The runtime brief no longer contains run ID, autopilot ID, title, source, trigger payload, or instructions. Those are rendered once, in the per-turn prompt. `main` also has an Autopilot-specific byte-identity test that covers a second run of the same autopilot and a run of a different autopilot. This part of the PR now only conflicts with `main`.
- **The repository half would now cause a regression.** #8592 lets a project pin a checkout ref on a `github_repo` resource. `## Repositories` now carries the pinned-branch rules: which branch the checkout starts from, `gh pr create --base <that-branch>` when delivering, and a warning not to retarget a kept checkout. When a project has `github_repo` resources, the claim returns only the project's repos and doesn't mix in workspace repos. That means the filter here always removes every entry and drops the whole section, rules included. Project Context only shows `checkout ref: <ref>`. So an agent working on a pinned release branch would go back to opening its PR against the default branch.

The remaining duplication is one URL line per project repo. It sits in the stable, cached part of the brief, so the savings are small. If we revisit it, the likely direction is the reverse of this PR: keep `## Repositories` as the only list of repos and have Project Context point to it.

Thanks again for the contribution and the careful regression tests.

PR #7758, opened 2026-08-29
MUL-6838 fix(i18n): localize squad and editor labels

sooodasama:
## What does this PR do?

Removes hard-coded English from the squad detail and editor surfaces so user-facing labels, feedback, placeholders, enum values, iframe titles, and accessible names follow the selected locale. The change uses the existing typed `useT` selectors and keeps API payloads, state, and behavior unchanged.

The implementation was scoped by tracing each visible literal to its owning component and locale namespace, checking the Chinese copy against the repository terminology contract, and excluding downstream-only branding or product behavior.

## Related Issue

N/A — this is a self-contained localization bug fix.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- Route squad detail toasts, fields, tabs, placeholders, and `agent` / `member` display values through the `squads` namespace.
- Localize editor mention badges, HTML preview iframe titles, and Mermaid accessible names.
- Add matching English, Japanese, Korean, and Simplified Chinese resources.
- Add regression coverage for localized editor output and locale key parity.
- No downstream branding, private paths, API changes, or business-logic changes are included.

## How to Test

1. Run `corepack pnpm --filter @multica/views typecheck`.
2. Run `corepack pnpm --filter @multica/views exec vitest run locales/parity.test.ts editor/extensions/code-block-view.test.tsx editor/extensions/mention-suggestion.test.tsx editor/html-block-preview.test.tsx editor/mermaid-diagram.test.tsx`.
3. Run `corepack pnpm --filter @multica/views lint`.

Local results: typecheck passed; 5 test files / 230 tests passed; lint completed with 0 errors (27 pre-existing warnings).

Risk is limited to presentation copy and accessible labels. Translation key parity and typed selectors cover missing-resource regressions.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [ ] I have updated relevant documentation to reflect my changes
[...]

PR #7776, opened 2026-08-30
MUL-6850 fix(i18n): localize onboarding status label and lint the attribute/toast copy path

Bohan-J:
Two leftovers surfaced by the post-merge review of #7754 / #7758.

## 1. One more mixed-language line — onboarding

`StepWelcome`'s illustration renders mock issue cards, and every string on that screen goes through `useT("onboarding")` — except the status footer, which read `STATUS_CONFIG[...].label`, i.e. the server's English seed. A zh / ja / ko member met the product on a line that read half in each language:

| | before | after |
|---|---|---|
| zh-Hans | `进行中` icon + **In Progress** | `进行中` |
| ja | `進行中` icon + **In Review** | `レビュー中` |

Built-in statuses now resolve from the `issues` namespace — the same keys `useStatusLabel` resolves for real issues (MUL-6243). The illustration's statuses are hard-coded built-in keys, so no workspace status-catalog lookup is involved and `useStatusLabel(wsId)` would be the wrong tool here. `STATUS_CONFIG` still owns the icon color.

`step-welcome.tsx` had no test file (every sibling step does); this adds one that asserts against the real locale bundles, so re-introducing the English seed fails the suite rather than a screenshot review.

## 2. The root cause: lint never looked at this class of string

`packages/views` runs `i18next/no-literal-string` in `mode: "jsx-text-only"`, which only inspects JSX **children**. JSX attributes (`placeholder=` / `title=` / `aria-label=`) and `toast(...)` arguments were never linted at all — which is exactly why #7758 had to hand-fix ~30 of them after they had already shipped. Without closing this, the same PR keeps coming back.

Widening the plugin is not the answer. Measured on this package:

| config | errors |
|---|---|
| `mode: "jsx-only"` | 2639 |
| `mode: "jsx-only"` + `jsx-attributes.include` allowlist | 157 |

`jsx-only` validates *every* literal that appears anywhere inside JSX — `isColVisible("status")`, `{ month: "short" }`, enum keys, option values — and only ~10 of those are copy. So instead this pins the two doors copy actually escapes through with `no-restricted-syntax`, the same technique `apps/desktop/eslint.config.mjs` already uses to guard `shell.openExternal` and `router.navigate`:

- `placeholder` / `title` / `aria-label` with a string literal (`https?:`-prefixed values excluded — a URL is never copy)
- `toast("…")` and `toast.error("…")` with a string literal, in `.ts` as well as `.tsx`, since toasts fire from hooks as often as components

## What the guard caught

8 sites, all in this diff:

**Real copy — localized in en / zh-Hans / ja / ko:**
- `mcp-server-dialog.tsx` — `placeholder="Key"` / `"Value"`, plus two aria-labels that concatenated an English word onto an already-translated label (`` `${label} key ${index + 1}` `` → `进行中 key 1`). Now interpolated keys.
- `chat-message-list.tsx` — `aria-label="Suggested follow-ups"` on the quick-actions row
- `issue-detail.tsx` — `aria-label="Select all sub-issues"` on the sub-issues select-all checkbox

**Technical values — exempted inline with a reason:** `npx` (a command the user types), `xoxb-…` / `xapp-…` (Slack token prefixes), `123456789:AA…` (Telegram token shape). Format hints, not copy.

## Verification

- `pnpm lint` / `pnpm typecheck` / `pnpm test` green at the repo root — 410 test files, 4864 tests
- `locales/parity.test.ts` holds: all 6 new keys present in en / zh-Hans / ja / ko
- New guards proved to fire: a probe file with `toast.error("…")`, `toast("…")`, `placeholder=`, `title=`, `aria-label=` reports 5 errors, while `title="https://example.com"` is correctly exempt
- New test proved to catch the regression: restoring `{cfg.label}` fails the zh-Hans case

## Out of scope

`apps/mobile/components/chat/chat-message-list.tsx:467` has the same `accessibilityLabel="Suggested follow-ups"`. Mobile has its own i18n setup and is excluded from the root lint/test filter, so it is left for a separate change.


Bohan-J:
Thanks for the careful review here — the catch that the first version of the guard only matched a bare `Literal` child was the one that mattered. It would have shipped a rule that looked like it closed the hole while `toast.error(err instanceof Error ? err.message : "Failed to save")` kept sailing through, which is worse than no rule at all: a green `pnpm lint` would have been read as proof the class of bug was handled.

Landing with three commits:

1. The original fix — onboarding status labels resolve from `issues.status.*`, plus the attribute/toast guard.
2. Guard widened to cover JSX expression containers, template literals whose *static* chunks carry letters, and the fallback half of `? :` / `||` / `??` / `&&`. Selectors moved to `eslint-i18n-guard.mjs` so `eslint-i18n-guard.test.ts` can lint fixtures against the exact strings the build enforces — 25 cases, and reverting to the narrow selectors fails 8 of them. Turning it on found 4 real untranslated sites, all localized in en / zh-Hans / ja / ko.
3. `{{label}} 键 {{index}}` → `{{label}}：键 {{index}}` for zh-Hans and ja. A bare space cannot be right for both call sites, since `label` is `环境变量` in one and `HTTP Headers` in the other; a colon reads correctly for both. en and ko are unchanged — both space words natively.

Verified locally on Node 22: root `pnpm lint` 0 errors, root `pnpm typecheck` clean, `@multica/views` 411 files / 4889 tests green. CI is green on all 16 checks.

Issue #7777, opened 2026-08-30
Chat: assistant reply persists but renders invisible — timeline never falls back to chat_messages.content, and the daemon tail flush loses task_messages permanently (no retry)


AzeWZ:
## Summary

In the chat UI, an assistant reply can be **persisted server-side but render as invisible**: the session list preview shows the final reply text, while the conversation detail shows the turn with elapsed time and quick actions but **no reply body**. A page refresh does not recover it. The data needed to render the reply exists in `chat_messages`, so this is not "the agent didn't answer" — it is a rendering contract gap combined with a permanent transcript-tail loss channel.

### 1. The renderer never consults `chat_messages.content` when a timeline exists

`packages/views/chat/components/chat-message-list.tsx` (AssistantMessage):

```tsx
{isNoResponse ? (
  <NoResponseNotice />
) : message && timeline.length === 0 ? (
  <RichContent content={message.content} ... />
) : null}
```

When `timeline.length > 0`, the final answer text is rendered **only** from the task timeline's trailing text segment — `splitTimeline`'s `final` = text items positioned **after the last non-text item** (`packages/views/chat/lib/copy-text.ts`). If that tail is missing from the timeline, the turn renders elapsed + quick actions with no reply body.

Notably, `extractCopyText` (Copy button) **already has** a fallback to `message.content` for exactly this pathological shape ("so Copy never produces an empty string") — the render path lacks the same guard. With the Copy fix, a user can copy the answer they cannot see.

### 2. The timeline tail can be permanently lost: daemon tail flush is fire-and-forget

The daemon batches streamed text into `pendingText` and flushes every 500ms plus a final flush before completion (`server/internal/daemon/daemon.go`, drain goroutine):

```go
if err := d.client.ReportTaskMessages(sendCtx, taskID, toSend); err != nil {
    taskLog.Debug("failed to report task messages", "error", err)
}
```

- `ReportTaskMessages` is a **single-shot `postJSON` with no retry and a 5s timeout** (`server/internal/daemon/client.go:435`).
- The final flush is exactly the one carrying the **closing answer text** (the last ≤500ms of the run).
- Failure is swallowed at Debug level; the task still completes normally.

[...]

PR #7778, opened 2026-08-30
MUL-6851: fix(chat): fall back to message.content when the task timeline has no trailing text

AzeWZ:
## Summary

Fixes the client half of #7777: when a persisted assistant turn's task timeline is non-empty but has **no trailing text segment**, the turn rendered elapsed time and quick actions with **no reply body** — even though the authoritative answer text is right there in `chat_messages.content`.

The renderer only consults `message.content` when `timeline.length === 0`; with a timeline, the answer must come from the timeline's trailing text (`splitTimeline`'s `final`). When that tail is missing — a daemon tail-flush that never persisted (#7777), the answer sandwiched before a later thinking/tool frame, or a hole in the client's task-messages cache — the reply was invisible.

## Change

`AssistantMessage` (`packages/views/chat/components/chat-message-list.tsx`) now falls back to rendering `message.content` when:

- `timeline.length > 0` **and** `final` is empty,
- the turn is not `no_response` (its placeholder still wins),
- `message.content` is non-empty.

This mirrors the guard `extractCopyText` already had ("so Copy never produces an empty string") — previously a user could Copy the answer they could not see. The fallback covers every loss variant: daemon tail loss, sandwiched text, client cache holes — including historical sessions whose tail rows are already gone.

A second commit extends the fallback to the case where `final` is non-empty but **disagrees with `message.content`** (a truncated tail that is only a strict prefix of content, or a stale mid-run narration frame sitting after the last non-text item while the real answer never persisted). The trigger is now a normalized comparison of the timeline's final text vs `message.content`; on mismatch, content is authoritative — the final text segment is replaced, the process fold is kept. The analytics event payload gains a `reason` (`empty_final` | `final_mismatch`) to separate the two shapes.

Also fires an `assistant_reply_missing_final_text` analytics event (once per message id per mount; carries `message_id` / `task_id` / `message_kind` / `timeline_items` / `reason`) so the user-visible symptom is directly measurable and the historical hit rate is observable.

## Self-test

- New unit tests: fallback renders the authoritative text; no duplicate render when trailing text exists; truncated-tail and stale-narration mismatch cases render content without duplicates; `no_response` still shows the placeholder; failed turns still show the failure bubble; telemetry fires exactly once per message with the correct `reason`; whitespace-only differences never fire the fallback.
- Regression: `packages/views` chat + task-transcript suites green; `tsc --noEmit` clean. All run against a fresh `upstream/main` base.

## Notes

- The server/daemon half (retry + `(task_id, seq)` idempotency) is proposed in #7777 and intentionally out of scope here.
- Verified end-to-end on a production workspace: a previously invisible reply (preview showed it, detail did not) renders correctly with this build.

### DIFF
--- a/packages/views/chat/components/chat-message-list.test.tsx
+++ b/packages/views/chat/components/chat-message-list.test.tsx
@@ -1,11 +1,14 @@
-import { describe, expect, it, vi } from "vitest";
+import { describe, expect, it, vi, beforeEach } from "vitest";
 import { act, fireEvent, render, screen } from "@testing-library/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { I18nProvider } from "@multica/core/i18n/react";
 import { chatKeys } from "@multica/core/chat/queries";
-import type { TaskMessagePayload } from "@multica/core/types";
+import { captureEvent } from "@multica/core/analytics";
+import type { ChatMessage, TaskMessagePayload } from "@multica/core/types";
 import type { ReactElement } from "react";
 import enChat from "../../locales/en/chat.json";
+
+vi.mock("@multica/core/analytics", () => ({ captureEvent: vi.fn() }));
 
 // The live timeline is a real list row rather than Virtuoso chrome (MUL-4922),
 // so it shares one identity with the persisted assistant row and keeps its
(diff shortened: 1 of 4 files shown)


AzeWZ:
@Bohan-J could you take a look at this one?

It fixes the client half of #7777: assistant replies persist to `chat_messages.content`, but render invisible whenever the task timeline has no trailing text segment (tail frames lost by the daemon's non-retried flush) — the renderer never fell back to the persisted content, while Copy already did. The fallback covers all loss variants, including historical sessions whose tail rows are already gone.

Verified locally with a packaged build against production; unit + regression suites green on latest `main` (details in the PR description). Happy to adjust per your feedback.

PR #7783, opened 2026-08-30
MUL-6856: fix(daemon): disclose relocated resumed sessions

mameikagou:
## What does this PR do?

Adds a short per-turn notice when a provider conversation resumes successfully in a different working directory from the previous turn.

### Thinking path

PR #7760 separates provider-session reachability from workdir reuse for independently addressed Pi-family transcripts. That preserves conversation history, but it also means the resumed model can remember files and uncommitted changes from a directory this turn no longer uses. The model cannot infer that relocation from the transcript alone.

The daemon already owns both facts after all resume gates finish: whether `PriorSessionID` survived, and the actual prepared workdir. This PR compares directory identity at that point and appends a run-scoped prompt block only when both a resume remains live and the workdir changed. A dropped resume keeps the existing, stronger continuity-loss notice. The block stays out of the cached runtime brief, preserving the prompt-cache prefix invariant.

The detector is provider-agnostic and merge-order independent. On current `main`, cwd-keyed gates clear the prior session before it can fire; once #7760 lands, relocated Pi/OMP resumes naturally activate it without coupling this PR to the Pi gate implementation.

## Related Issue

Follow-up to #7760 review discussion; no separate issue.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Detect a surviving resumed session whose prior and current directories have different filesystem identities.
- Append a concise `Working directory changed` block to the per-turn prompt.
- Tell the agent to inspect the current working tree before relying on remembered paths or file state.
- Cover activation, non-activation, session-loss separation, and cache-prefix placement.

## How to Test

1. `cd server && go test ./internal/daemon -run "TestResumedSessionMovedWorkdir|TestRelocatedWorkingDirectoryBlock|TestGateResumeToReusedWorkdir|TestSharedLocalDirectoryBlock" -count=1`
2. `cd server && go test ./internal/daemon`
3. `cd server && go vet ./internal/daemon/`

## Checklist

- [x] I have updated relevant documentation to reflect my changes (inline contract documentation only; no user-facing docs affected)
- [x] If this PR touches Chinese product copy, I checked the conventions (not applicable)

### Risk

The notice is gated on a non-empty surviving prior session and a recorded prior workdir with different filesystem identity. It changes prompt text only; session selection and environment preparation are untouched.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Traced the review concern through the post-prepare resume gates and per-turn prompt assembly, implemented the smallest provider-independent disclosure, added focused tests, and checked that the branch merges cleanly with #7760.


mameikagou:
@Bohan-J I followed up on your relocated-workdir design question here. This keeps the disclosure separate from #7760, appends it only to the run-scoped prompt after all resume gates, and has been merge-tree checked against #7760 with no conflicts. Would appreciate your thoughts when convenient.

mameikagou:
Merged current main (5f0e0f229) and resolved the prompt-extension conflicts in 5a2c7458d.

The relocated-session notice and main’s new worktree replay-conflict notice now coexist as independent per-turn blocks; both daemon conditions and both regression suites are preserved.

Validation passed:
- targeted relocated/worktree prompt tests
- env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1
- go vet ./internal/daemon
- git diff --check

The PR is updated and conflict-free. @Bohan-J, could you please review when convenient?

Bohan-J:
Thanks for this, and for keeping it rebased through the `main` conflicts — sorry for the slow turnaround.

We reviewed it end to end and are going to close it. The analysis behind it is right, and the mechanism choices are the ones we'd have made ourselves: running the check after all resume gates so a dropped resume still gets the stronger continuity notice, comparing directory identity rather than path spelling, and appending to the per-turn block instead of the cached runtime brief. The reason for closing is that what sits underneath it changed, not a disagreement with any of that.

## Why we're closing

The relocation window #7760 opened is real, but most of what pushed a turn into a different directory in the first place was our own bugs, and we'd rather fix those than describe the symptom in the prompt:

- Worktree mode used to fork a fresh branch from `HEAD` and a fresh tree for every comment, so consecutive turns of one conversation never saw each other's work. `main` now keeps one durable branch per conversation and replays the user's uncommitted state onto it (`server/internal/daemon/execenv/local_worktree.go`). Turn two now stands on turn one's work.
- A comment edit could dispatch a second task for the same trigger while the first was still running; the new run lost the env-root lock and fell back to a fresh environment. That one is being fixed on our side.

With both out, the relocation cases that remain are a GC'd prior directory and genuinely concurrent sibling tasks. We don't think that residual surface earns a permanent prompt block on every Pi/OMP turn — a block that fires routinely reads as background noise and dilutes the rarer notices next to it.

## Two findings worth recording, in case anyone revisits this

Both still hold at `5a2c745`:

**1. The fresh-session retry path emits contradictory text.** `daemon.go` reuses `promptOptions` to build `freshPrompt` after setting `task.PriorSessionResumeUnavailable = true`, so a turn that falls back to a fresh session renders the relocation block ("The provider conversation resumed successfully, but this turn is running in a different working directory…") and the continuity-loss notice in the same preamble. `promptOptions` needs filtering on that path. It isn't an exotic corner — the resume-failure retry is one of the more common ways to reach it.

**2. The copy is aimed at the weaker failure.** "Do not assume files, paths, or uncommitted changes remembered from that directory are present here" assumes the old directory is gone — and that case is self-correcting, because the read fails and the agent looks again. But `DefaultGCCompletedTaskTTLSelfHost = 0` (disabled), so on self-host the remembered absolute path usually still resolves, and what it returns is the *previous* turn's tree: a silent stale read with no error, after which the agent concludes a remembered change is present when it isn't. If this gets redone, the copy should say the remembered path may still be readable but points at the previous turn's tree, and name the current one.

Closing as "not now" rather than "wrong". If Pi/OMP usage grows, or we hit this for real once the two fixes above are in, we'll come back to it with the copy rewritten. Appreciate the work you put into it.

PR #7730, opened 2026-08-29
feat(skills): add Workspace Skill slash picker to Chat, comments, and task creation

medking82:
## What does this PR do?

Adds a Workspace Skill slash picker to Chat, issue comments/replies, and both task-creation editors. Typing `/` lists the current workspace Skills while preserving Quick Actions and reserved `/note` behavior.

A UUID-backed selection grants the Skill only to the exact Run. Task creation freezes member-authored selection before claim, so editing the issue description cannot replace that grant. Later Runs must select their own Skills.

Closes #7729.

## Changes

- Integrates upstream main at `90e0bdf8`, retaining the existing PR history.
- Validates same-workspace, member-owned selections and the 20-Skill bound on the server.
- Preserves agent-scoped builtins, attached/plugin Skills and sparse requested-bundle reads.
- Supports assigned, deferred and squad creation, Chat, quick create and comment-triggered claims.
- Commits selected grants atomically with claim tokens; preserves legacy task context and nil-response handling.
- Removes prior one-Run Codex Skill materialization before the next execution.
- Keeps slash-trigger behavior after Chrome's stale selection following `Ctrl+A` and `Delete`.

## Validation

- 128 focused editor/modal Vitest tests passed.
- `@multica/views` TypeScript typecheck and focused ESLint on all 10 changed TS/TSX files passed.
- 17 top-level PostgreSQL-backed service/handler cases passed with no skips, including selected/unselected resolution, mixed member/agent comments, subsequent Chat Run isolation, create-time freezing, deferred/squad dispatch, scoped builtins, legacy context, nil response and rollback.
- Regenerated SQL using the repository's sqlc v1.31.1 and reran selected DB cases.
- Focused parser, prompt and Codex materialization tests, affected Go vet, and staged whitespace checks passed.
- Broader local Windows provider-fixture failures also reproduce on the unmodified upstream snapshot; the full suite is not claimed green. CI remains required.

## Review and UI

AI tooling: OpenAI Codex implementation and deterministic testing; one completed independent Claude Opus 5.5 review of the frozen authorization diff. Two P2 and one P3 findings were fixed and covered by deterministic regressions. No P0/P1 findings were reported.

The [existing UI screenshot](https://github.com/multica-ai/multica/pull/7730#issuecomment-5462599552) illustrates the picker. It predates this upstream integration and is not a new end-to-end acceptance run.

PR #7786, opened 2026-08-30
fix(desktop): preserve explicit bundled CLI release versions (MUL-6858)

medking82:
## Summary

- Allow the Go CLI bundler to receive an explicit `MULTICA_CLI_VERSION` for a stable or `X.Y.Z-custom.N` Desktop release built from a checkout without reachable CLI tags. Invalid explicit values or missing Go fail closed instead of silently packaging a stale/mis-versioned binary; ordinary builds retain their existing git-derived behavior.
- Keep packaging subprocess windows hidden on Windows and pin the imported package script's hashbang to LF for Vite/Vitest.
- Add a tiny native Go regression fixture and run it in the existing read-only Linux/Windows Desktop Smoke workflow. No real agent, account, network dependency or microphone is exercised by that fixture.

## Reproduction

In a tagless checkout the bundler used `git describe --always`, embedding a bare commit hash as `main.version`. A Desktop whose Electron version was set through `-c.extraMetadata.version=...` could therefore ship with a CLI that fails the runtime/UI semantic-version gate.

The native regression failed against current upstream (`15280617b`): actual CLI version was a bare SHA instead of `0.4.36-custom.123`. It passes with this change. Custom release callers must pass the same immutable release version to the bundler and Electron's metadata override.

## Validation

- `node --test apps/desktop/scripts/bundle-cli.integration.mjs`: 4 passed (Windows x64, Go 1.26.7).
- `pnpm --filter @multica/desktop test scripts/package.test.mjs`: 31 passed.
- actionlint 1.7.12 for the changed Desktop Smoke workflow: passed.
- `git diff --check`: passed.

This is a focused extraction of the packaging repair already validated in a custom Windows installer. A fresh upstream installer was not published or installed for this PR. Publisher/feed configuration, signing behavior, package versions, native dictation and the skill-picker PR #7730 are unchanged.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,6 @@ docker/entrypoint.sh text eol=lf
 
 # Default behavior
 * text=auto
+
+# Vite's SSR hashbang matcher requires LF when this CLI is imported by Vitest.
+apps/desktop/scripts/package.mjs text eol=lf
--- a/.github/workflows/desktop-smoke.yml
+++ b/.github/workflows/desktop-smoke.yml
@@ -48,6 +48,10 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Verify bundled CLI release version
+        working-directory: apps/desktop
+        run: node --test scripts/bundle-cli.integration.mjs
+
       - name: Package Desktop installers (${{ matrix.target }})
         working-directory: apps/desktop
         env:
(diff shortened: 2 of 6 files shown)


multica-eve:
Thank you for the careful reproduction and validation work here.

We are going to close this PR without merging it. Multica's supported Desktop release flow is tag-driven, and we do not plan to support or maintain a separate tagless custom-distribution path. This change would add an explicit version input that publishers must keep synchronized across the CLI bundler and Electron packaging, for a release mode outside that supported workflow.

The problem you identified is understandable, but it does not affect our official release process, so we cannot take on this additional packaging contract. You are welcome to keep the patch in a fork if it is useful for your own custom builds.

PR #7790, opened 2026-08-30
MUL-6862: fix(codex): widen thread setup handshake budget

mameikagou:
## What does this PR do?

Gives Codex `thread/start` and `thread/resume` a separate 60s default handshake budget while keeping lightweight startup RPCs such as `initialize` and `turn/start` at 30s.

Field logs showed both thread RPCs repeatedly reaching the old uniform 30s deadline while Codex was still performing synchronous thread setup. Follow-up measurements on the same host captured healthy `thread/resume` completions at 22.643s, 27.823s, and 34.985s. The 60s budget therefore covers the observed healthy range with 25.015s (about 71%) of headroom over the maximum, without widening lightweight RPC budgets.

A backend-only default change is insufficient because the daemon resolves and passes its 30s default explicitly, so this change carries the separate budget through daemon config and `ExecOptions`.

The legacy `codex_handshake_timeout` override remains global for backward compatibility. Thread setup timeouts remain fail-closed: Multica kills and reaps the process tree but does not replay an RPC whose provider-side outcome is uncertain.

## Redacted field evidence

The incident host was macOS 26.5.2 on arm64. The original failures used Codex CLI 0.145.0; follow-up healthy measurements used Codex CLI 0.151.0 after repairing the standalone installation.

The captured incident window contained four distinct `thread/start` attempts and five distinct `thread/resume` attempts that reached the old 30s boundary. One concurrent pair, normalized to UTC, was:

| Method | Sent | Watchdog fired | Elapsed |
| --- | --- | --- | --- |
| `thread/start` | 2026-08-30 11:26:20.531Z | 2026-08-30 11:26:50.592Z | 30.061s |
| `thread/resume` | 2026-08-30 11:26:22.756Z | 2026-08-30 11:26:52.826Z | 30.070s |

With a larger diagnostic ceiling, three healthy `thread/resume` calls on the same host completed in 22.643s, 27.823s, and 34.985s. The over-30s sample was sent at 2026-08-30 15:01:12.209Z and returned successfully at 2026-08-30 15:01:47.194Z.

Codex's per-task trace for a representative slow `thread/start` showed synchronous work continuing during the delay: its persisted model cache was unusable after a schema change, so Codex refreshed the remote model catalog, then entered session initialization and `session_init.auth_mcp` plugin/MCP configuration before the old watchdog fired. The resume trace used Codex's `resume_thread_with_history` path and then the same `thread_spawn` / session-initialization work.

No prompts, local paths, tokens, credentials, thread identifiers, or host identifiers are included above.

## Field failure

The user-facing task failed after about one minute and exposed the underlying resume handshake error:

```text
codex thread/resume failed: codex app-server handshake timeout: thread/resume did not respond after 30s
```

The daemon log showed the same failure recurring across retries. The incident window also contained `thread/start` timeouts at the same 30s boundary, which is why the fix covers both heavy thread-setup RPCs instead of special-casing only resume.

## Related Issue

Observed in a field incident; no public issue exists yet.

## Type of Change

- [ ] New feature (non-breaking change)
- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `ThreadHandshakeTimeout` and resolve Codex defaults as 30s for lightweight RPCs and 60s for `thread/start` / `thread/resume`.
- Propagate the separate default from daemon config so the backend default is not masked by an explicitly passed 30s value.
- Preserve existing global timeout override behavior.
- Add lifecycle timing and failure diagnostics for `thread/resume`, including explicit process-tree cleanup on timeout.
- Add regression coverage for timeout resolution, daemon propagation, and a stuck resume RPC.

## How to Test

1. `go test ./pkg/agent -run "Test(ResolveCodexHandshakeTimeouts|CodexHandshakeTimeoutFor|CodexExecuteThread(StartTimeoutLifecycleIsFailClosed|ResumeTimeoutUsesThreadBudgetAndLifecycle)|CodexExecuteStartupRPCsHaveBoundedHandshakeTimeout)$" -count=1`
2. `go test ./internal/daemon -run "^TestLoadConfig_CodexHandshakeTimeout$" -count=1`
3. `go build ./cmd/multica`

## Checklist

- [x] I have run the targeted tests locally and they pass
- [x] This change does not affect the UI
- [x] No documentation update is required; the existing global override keeps its documented behavior

## AI Disclosure

**AI tool used:** Codex via Multica

**Prompt / approach:** Reviewed the original daemon and Codex lifecycle logs, disproved the socket/standalone hypothesis against the stdio launch path, traced timeout propagation from daemon config into `ExecOptions`, then implemented and tested the smallest backward-compatible split between lightweight and thread-setup RPC budgets.

## Screenshots (optional)

Not applicable; backend/runtime behavior only.

### DIFF
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -30,6 +30,7 @@ const (
 	DefaultAgentTimeout                   = 0
 	DefaultCodexSemanticInactivityTimeout = 10 * time.Minute
 	DefaultCodexHandshakeTimeout          = 30 * time.Second
+	DefaultCodexThreadHandshakeTimeout    = 60 * time.Second
 	// DefaultOpenCodeIdleWatchdog shortens the no-message budget for OpenCode
 	// runs while they are not executing a tool. OpenCode streams text and tool
 	// events incrementally, so a completely silent interval here covers both a
@@ -134,6 +135,7 @@ type Config struct {
 	// for app-servers that are legitimately slow to their first event (GH #3262).
 	CodexFirstTurnNoProgressTimeout time.Duration
 	CodexHandshakeTimeout           time.Duration
+	CodexThreadHandshakeTimeout     time.Duration
 	OpenCodeIdleWatchdog            time.Duration // OpenCode-specific no-message window; 0 falls back to AgentIdleWatchdog and values above it cannot extend the global bound
 	AgentIdleWatchdog               time.Duration // force-stop a run when the backend goes silent this long with an empty queue (0 = disabled)
 	AgentToolWatchdog               time.Duration // force-stop a run when a single tool call stays in flight (silent) this long (0 = never force-stop during a tool call); defaults to AgentIdleWatchdog, so operators tune one number unless they deliberately want a wider tool budget
@@ -416,6 +418,19 @@ func LoadConfig(overrides Overrides) (Config, error) {
 	if overrides.CodexHandshakeTimeout > 0 {
 		codexHandshakeTimeout = overrides.CodexHandshakeTimeout
 	}
+	// Preserve the legacy global override semantics while giving the heavy
+	// thread RPCs a wider built-in default. The CLI resolves flag, env and
+	// persisted config values into Overrides, while embedded callers may reach
+	// LoadConfig with the environment directly.
+	codexThreadHandshakeTimeout := DefaultCodexThreadHandshakeTimeout
+	if raw, ok := os.LookupEnv("MULTICA_CODEX_HANDSHAKE_TIMEOUT"); ok && strings.TrimSpace(raw) != "" {
+		if parsed, parseErr := time.ParseDuration(strings.TrimSpace(raw)); parseErr == nil && parsed > 0 {
+			codexThreadHandshakeTimeout = codexHandshakeTimeout
+		}
+	}
+	if overrides.CodexHandshakeTimeout > 0 {
+		codexThreadHandshakeTimeout = overrides.CodexHandshakeTimeout
+	}
 
 	maxConcurrentTasks, err := intFromEnv("MULTICA_DAEMON_MAX_CONCURRENT_TASKS", DefaultMaxConcurrentTasks)
 	if err != nil {
@@ -602,6 +617,7 @@ func LoadConfig(overrides Overrides) (Config, error) {
 		CodexSemanticInactivityTimeout:  codexSemanticInactivityTimeout,
 		CodexFirstTurnNoProgressTimeout: codexFirstTurnNoProgressTimeout,
 		CodexHandshakeTimeout:           codexHandshakeTimeout,
+		CodexThreadHandshakeTimeout:     codexThreadHandshakeTimeout,
 		OpenCodeIdleWatchdog:            openCodeIdleWatchdog,
 		AgentIdleWatchdog:               agentIdleWatchdog,
 		AgentToolWatchdog:               agentToolWatchdog,
--- a/server/internal/daemon/config.go
+++ b/server/internal/daemon/config.go
@@ -424,7 +424,7 @@ func LoadConfig(overrides Overrides) (Config, error) {
 	// LoadConfig with the environment directly.
 	codexThreadHandshakeTimeout := DefaultCodexThreadHandshakeTimeout
 	if raw, ok := os.LookupEnv("MULTICA_CODEX_HANDSHAKE_TIMEOUT"); ok && strings.TrimSpace(raw) != "" {
-		if parsed, parseErr := time.ParseDuration(strings.TrimSpace(raw)); parseErr == nil && parsed > 0 {
+		if parsed, parseErr := parseFlexDuration(strings.TrimSpace(raw)); parseErr == nil && parsed > 0 {
 			codexThreadHandshakeTimeout = codexHandshakeTimeout
 		}
 	}
--- a/server/pkg/agent/codex.go
+++ b/server/pkg/agent/codex.go
@@ -59,7 +59,12 @@ const (
 	// while keeping the fast-fail value (MUL-5542).
 	defaultCodexFirstTurnNoProgressTimeout = 60 * time.Second
 	defaultCodexHandshakeTimeout           = 30 * time.Second
-	codexVersionDiagnosticTimeout          = 2 * time.Second
+	// thread/start and thread/resume may refresh the model catalog, initialize
+	// MCP integrations, and restore persisted history. Field evidence shows
+	// healthy calls crossing the 30s budget used for local initialize, so keep
+	// their default separate without slowing failures for lightweight RPCs.
+	defaultCodexThreadHandshakeTimeout = 60 * time.Second
+	codexVersionDiagnosticTimeout      = 2 * time.Second
 	// codexGracefulShutdownTimeout bounds how long the lifecycle goroutine
 	// waits for codex to exit on its own after stdin is closed, before forcing
 	// a context-cancel kill. A clean exit lets codex run its shutdown path and
@@ -947,10 +952,7 @@ func (b *codexBackend) executeOnce(ctx context.Context, prompt string, opts Exec
 	if semanticInactivityTimeout == 0 {
 		semanticInactivityTimeout = defaultCodexSemanticInactivityTimeout
 	}
-	handshakeTimeout := opts.HandshakeTimeout
-	if handshakeTimeout <= 0 {
-		handshakeTimeout = defaultCodexHandshakeTimeout
-	}
+	handshakeTimeout, threadHandshakeTimeout := resolveCodexHandshakeTimeouts(opts)
 	runCtx, cancel := runContext(ctx, timeout)
 
 	// Materialise the agent's MCP config into the per-task
@@ -1108,16 +1110,17 @@ func (b *codexBackend) executeOnce(ctx context.Context, prompt string, opts Exec
 	turnDone := make(chan bool, 1) // true = aborted
 
 	c := &codexClient{
-		cfg:                  b.cfg,
-		stdin:                stdin,
-		pending:              make(map[int]*pendingRPC),
-		processDone:          make(chan struct{}),
-		handshakeTimeout:     handshakeTimeout,
-		pid:                  cmd.Process.Pid,
-		attempt:              attempt,
-		activeLaunches:       activeLaunches,
-		notificationProtocol: "unknown",
-		acceptNotification:   turnNotificationGate.accept,
+		cfg:                    b.cfg,
+		stdin:                  stdin,
+		pending:                make(map[int]*pendingRPC),
+		processDone:            make(chan struct{}),
+		handshakeTimeout:       handshakeTimeout,
+		threadHandshakeTimeout: threadHandshakeTimeout,
+		pid:                    cmd.Process.Pid,
+		attempt:                attempt,
+		activeLaunches:         activeLaunches,
+		notificationProtocol:   "unknown",
+		acceptNotification:     turnNotificationGate.accept,
 		onDiscardedNotification: func(string, map[string]any) {
 			// Any app-server notification proves the process made semantic
 			// progress, even when it is intentionally excluded from the active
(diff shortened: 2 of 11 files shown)


multica-eve:
Thanks for the detailed fix. We traced the current 30-second default back to #5319: at that time, `thread/start` was measured at about 1.4 seconds with a cold catalog and was treated as a quick acknowledgement.

To validate changing that established assumption and choosing 60 seconds, could you share redacted evidence from the incident?

- Codex CLI version, OS, and whether the slow call was `thread/start` or `thread/resume`.
- Sent/response timestamps or the actual elapsed time for a healthy call that eventually completed after 30 seconds, if available.
- If the existing watchdog terminated every attempt at 30 seconds, what evidence indicates that 60 seconds is sufficient rather than only moving the cutoff?
- Which synchronous work was observed during the delay (for example model refresh, MCP initialization, or history restoration).

A couple of samples or a small latency range would be ideal. Please omit prompts, local paths, tokens, credentials, and other sensitive data. This will let us confirm the separate thread budget while keeping lightweight startup calls at 30 seconds.

mameikagou:
Thanks for pointing back to #5319. I recovered the original host's redacted daemon lifecycle records and per-task Codex traces.

**Environment and affected calls**

- Original incident: Codex CLI 0.145.0, macOS 26.5.2, arm64.
- Both `thread/start` and `thread/resume` were affected. The captured incident window contains four distinct `thread/start` attempts and five distinct `thread/resume` attempts that reached the old 30s boundary.
- One concurrent pair, normalized to UTC:
  - `thread/start`: sent `2026-08-30 11:26:20.531Z`; watchdog fired `11:26:50.592Z`; elapsed `30.061s`.
  - `thread/resume`: sent `2026-08-30 11:26:22.756Z`; watchdog fired `11:26:52.826Z`; elapsed `30.070s`.
- Another `thread/start` was sent at `12:09:32.250Z` and reached the watchdog at `12:10:02.320Z` (`30.065s`).

**Healthy completions and the 60s threshold**

After repairing the standalone installation and upgrading to Codex CLI 0.151.0, I temporarily used a larger diagnostic ceiling on the same host. Three healthy `thread/resume` calls completed in **22.643s, 27.823s, and 34.985s**. The over-30s sample was sent at `2026-08-30 15:01:12.209Z` and returned successfully at `15:01:47.194Z`.

That makes 60s a measured boundary rather than an arbitrary doubling: it leaves **25.015s of headroom (about 71%)** over the maximum healthy completion observed, while this PR deliberately keeps lightweight `initialize` / `turn/start` RPCs at 30s.

**Synchronous work observed during the delay**

For the representative slow `thread/start`, Codex accepted the request, found that its persisted model cache was unusable after a schema change, refreshed the remote model catalog, then entered session initialization and `session_init.auth_mcp` plugin/MCP configuration before the old watchdog fired. The resume trace used Codex's `resume_thread_with_history` path and then the same `thread_spawn` / session-initialization work.

I updated the PR description with these redacted measurements. Prompts, local paths, tokens, credentials, thread IDs, and host IDs have been omitted.

server/internal/daemon/config.go
```diff
@@ -416,6 +418,19 @@ func LoadConfig(overrides Overrides) (Config, error) {
 	if overrides.CodexHandshakeTimeout > 0 {
 		codexHandshakeTimeout = overrides.CodexHandshakeTimeout
 	}
+	// Preserve the legacy global override semantics while giving the heavy
+	// thread RPCs a wider built-in default. The CLI resolves flag, env and
+	// persisted config values into Overrides, while embedded callers may reach
+	// LoadConfig with the environment directly.
+	codexThreadHandshakeTimeout := DefaultCodexThreadHandshakeTimeout
+	if raw, ok := os.LookupEnv("MULTICA_CODEX_HANDSHAKE_TIMEOUT"); ok && strings.TrimSpace(raw) != "" {
+		if parsed, parseErr := time.ParseDuration(strings.TrimSpace(raw)); parseErr == nil && parsed > 0 {
```
Bohan-J: This re-parse uses `time.ParseDuration`, but the value was originally read via `durationFromEnv` → `parseFlexDuration`, which also accepts `1d` / `1.5d` / `1d12h`.

With `MULTICA_CODEX_HANDSHAKE_TIMEOUT=1d` I get `CodexHandshakeTimeout=24h0m0s` but `CodexThreadHandshakeTimeout=1m0s` — the heavy RPCs end up with a smaller budget than the lightweight ones, and it regresses current behavior where both are 24h.

This is the live daemon path, not just embedded callers: `resolveDaemonDurationOverride` (`server/cmd/multica/cmd_daemon.go:1665-1680`) returns 0 when the env var is set, so `overrides.CodexHandshakeTimeout` is 0 here and the branch below does not rescue it.

Use `parseFlexDuration`, or have `durationFromEnv` hand back an explicitly-set signal so no second parse is needed. Please add a `1d` regression case to `TestLoadConfig_CodexHandshakeTimeout` as well.

server/pkg/agent/codex.go
```diff
@@ -2293,6 +2337,19 @@ func isCodexHandshakeRPC(method string) bool {
...
 }
 
+func isCodexThreadSetupRPC(method string) bool {
+	return method == "thread/start" || method == "thread/resume"
+}
+
+func (c *codexClient) handshakeTimeoutFor(method string) time.Duration {
+	switch method {
+	case "thread/start", "thread/resume":
+		return c.threadHandshakeTimeout
```
Bohan-J: Nit, non-blocking: this returns `threadHandshakeTimeout` unguarded while `request()` gates on `timeout > 0`, so a zero-value `codexClient` would leave `thread/start` and `thread/resume` completely unbounded.

Only one construction site exists today (`codex.go:1112`, always via `resolveCodexHandshakeTimeouts`), so it is safe as written — but falling back to `c.handshakeTimeout` when the thread budget is non-positive would remove the footgun.

Bohan-J (changes requested):
Thanks for the detailed field evidence — that answered the main open question. I am not blocking on the 60s number. Two other things do need to change before this can merge.

## On the 60s budget: accepted

Widening this bound only ever *delays* a failure; it cannot turn a failure into a success. Thread setup timeouts stay fail-closed and are not retried — `Execute` retries only on `codexInitializeRetrySafe` / `codexStartupRefreshRetrySafe`, and a thread setup timeout carries neither — so the worst case is a single 60s window, not 120s. The cost of erring high is bounded at roughly 30 extra seconds on a genuinely stuck setup, plus one more log line carrying a latency field.

Given that, the 34.985s healthy sample is sufficient. It falsifies the "thread RPCs are fast acks" assumption the original 30s was built on, and that is the part that mattered.

One caveat worth recording in the PR description rather than resolving: the healthy timings came from CLI `0.151.0` after the standalone install was repaired, while the incident was on `0.145.0`, and the slow path was driven by a one-off model-cache schema invalidation. So this is not evidence that a healthy `thread/resume` routinely takes 22–35s. It does not block, precisely because the change is fail-safe.

## Blocking 1: the legacy global override breaks for `d`-unit values

`server/internal/daemon/config.go` re-parses `MULTICA_CODEX_HANDSHAKE_TIMEOUT` with `time.ParseDuration`, but the original parse goes through `durationFromEnv` → `parseFlexDuration`, which accepts `1d`, `1.5d`, `1d12h`. Reproduced locally against this branch:

```
MULTICA_CODEX_HANDSHAKE_TIMEOUT=1d
→ CodexHandshakeTimeout       = 24h0m0s
→ CodexThreadHandshakeTimeout = 1m0s
```

The heavy RPCs end up with a *smaller* budget than the lightweight ones, and it is a regression against current behavior, where both are 24h.

This is not limited to embedded callers. `resolveDaemonDurationOverride` (`server/cmd/multica/cmd_daemon.go:1665-1680`) deliberately returns `0` when the env var is set, leaving the parse to `LoadConfig`, so this is the live `multica daemon` startup path and the `overrides` branch never rescues it.

Suggested fix: call `parseFlexDuration` here, or have `durationFromEnv` return an "explicitly set" signal so the value does not need a second parse at all. Please also add a `1d` case to `TestLoadConfig_CodexHandshakeTimeout`.

## Blocking 2: the documented default is now wrong

The checklist says "No documentation update is required", but after this change the built-in default is 30s for lightweight RPCs and 60s for thread setup, while an explicit override still applies globally. Nine places still state a flat `30s`:

- `apps/docs/content/docs/environment-variables.mdx:234` and `:275`
- the same two rows in the `.zh`, `.ja`, and `.ko` variants
- `CLI_AND_DAEMON.md:255`

Please update all of them, and make the distinction between the built-in default and an explicit override explicit.

## Non-blocking nit

`handshakeTimeoutFor` returns `threadHandshakeTimeout` unguarded, and `request()` gates on `timeout > 0`. A zero-value `codexClient` would therefore leave the two heaviest RPCs entirely unbounded. There is only one construction site today (`codex.go:1112`, which always goes through `resolveCodexHandshakeTimeouts`), so this is safe as written — but a one-line fallback to `c.handshakeTimeout` when the thread budget is non-positive removes the footgun.

## Verified while reviewing

- No production path was missed: `daemon.go:7573` is the only place that sets `HandshakeTimeout`, and it now sets `ThreadHandshakeTimeout` alongside it.
- A `thread/resume` handshake timeout is classified by `isCodexTransportError`, so it correctly fails closed instead of falling back to `thread/start`.
- Targeted tests pass locally: the `pkg/agent` set from the PR description, and `TestLoadConfig_CodexHandshakeTimeout`.

Once the `d`-unit compatibility and the docs are fixed, this is good to merge.

server/internal/daemon/config.go
mameikagou: Fixed in 7d42dd537. The compatibility re-parse now uses parseFlexDuration, and TestLoadConfig_CodexHandshakeTimeout includes a 1d regression asserting that both the base and thread handshake budgets resolve to 24h.

server/pkg/agent/codex.go
mameikagou: Also addressed in 7d42dd537: handshakeTimeoutFor now falls back to the base handshake timeout when the thread-specific value is non-positive, with coverage for both thread/start and thread/resume.

mameikagou:
Addressed the latest review in `7d42dd537`:

- Preserved the legacy global override semantics for flexible day units by using `parseFlexDuration`; added a `1d` regression proving both base and thread budgets resolve to `24h`.
- Updated all 9 documented timeout entries (CLI reference plus EN/ZH/JA/KO env and config tables) to distinguish the built-in `30s` base / `60s` thread defaults and state that an explicit value overrides both globally.
- Applied the non-blocking safety nit: a non-positive thread budget now falls back to the base handshake timeout, with `thread/start` and `thread/resume` coverage.

Validation:
- `go test ./internal/daemon -run "^TestLoadConfig_CodexHandshakeTimeout$" -count=1`
- `go test ./pkg/agent -run "^(TestResolveCodexHandshakeTimeouts|TestCodexHandshakeTimeoutFor|TestCodexExecuteThreadStartTimeoutLifecycleIsFailClosed|TestCodexExecuteThreadResumeTimeoutUsesThreadBudgetAndLifecycle)$" -count=1`
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/agent`

@Bohan-J please re-review when convenient.

mameikagou:
All CI checks are green on `7d42dd537`, including the full backend suite and `windows-execenv`.

Bohan-J (approved):
Thanks for turning this around so quickly — and for the redacted field data, which is what made the 60s number decidable rather than a guess.

Re-reviewed at `7d42dd5` and both blocking items are closed:

- **`d`-unit override.** `parseFlexDuration` is now the single parse path. Verified locally against this head: `MULTICA_CODEX_HANDSHAKE_TIMEOUT=1d` yields `24h` for *both* budgets, and `1d12h` yields `36h` for both. The `1d` regression case in `TestLoadConfig_CodexHandshakeTimeout` locks it in.
- **Docs.** All nine places updated — the two rows across the four `environment-variables` locales, plus `CLI_AND_DAEMON.md:255` — and each now distinguishes the built-in `30s` / `60s` defaults from an explicit value that overrides both globally.

You also picked up the non-blocking nit: `handshakeTimeoutFor` now falls back to the base budget when the thread budget is non-positive, with coverage on both paths. Appreciated.

Targeted `pkg/agent` timeout/lifecycle tests pass here as well.

One cosmetic leftover, not worth another push: the PR description still has "No documentation update is required" checked, which no longer matches the nine doc changes. It does not reach the squashed commit message, so I am approving as-is.

Approving and merging.

PR #7796, opened 2026-08-30
MUL-6868 fix(migrations): repair partial VCS link schema

dixonl90:
## Summary

- add an idempotent repair migration for `issue_vcs_pull_request.reference_only`
- keep the down migration non-destructive because migration 216 already owns the column on healthy installations
- cover a partial pre-release VCS schema, existing-row backfill, repeated application, and rollback behaviour with a PostgreSQL integration test

## Why

A self-hosted deployment that had run an early VCS integration build retained an `issue_vcs_pull_request` table without `reference_only`. Migration 216 is already recorded there and uses `CREATE TABLE IF NOT EXISTS`, so later migration runs cannot reconcile the missing column. Every Gitea pull-request webhook was mirrored into `vcs_pull_request`, then issue linking failed with SQLSTATE `42703`.

The new migration is a no-op on healthy installations and repairs affected partial schemas by adding the column with the same `NOT NULL DEFAULT FALSE` definition as migration 216.

## Test plan

- [x] Confirmed the new integration test failed before the repair migration existed
- [x] `go test ./internal/migrations -count=1` against PostgreSQL 17
- [x] Migration direction and unique-prefix lint tests
- [x] `git diff --check`


multica-eve (approved):
Thanks for identifying and covering this pre-release migration edge case. I renumbered the repair migration to avoid the current main collision, and the PostgreSQL repair test plus migration prefix/direction checks pass locally against the merged tree. Approving for merge.

PR #7798, opened 2026-08-30
MUL-6870 fix(daemon): require proven task ownership before GC mutation

mameikagou:
## What does this PR do?

Makes daemon GC fail closed unless a candidate directory has authoritative task ownership and that ownership matches its exact two-level path under `WorkspacesRoot`.

This was found through a real data-loss incident after `workspaces_root` was intentionally set to a shared project directory for convenient native working-directory workflows. That is a legitimate use of a configurable path. The agent/GPT did not delete these files; Multica daemon GC did. GC treated the configured root as blanket authorization to delete without first proving directory provenance:

```text
WorkspacesRoot = <shared-project-root>

<shared-project-root>/<repository>/data
                       ^             ^
                       workspace     task directory, according to the old GC walk
```

The causal chain was:

1. `runGC` treated every non-dot first-level directory as a Multica workspace.
2. `gcWorkspace` treated every second-level directory as a task root.
3. A normal repository directory had no `.gc_meta.json`, so `shouldCleanTaskDir` classified it through `orphanByMTime(..., "no meta")`.
4. Once its directory mtime exceeded the default 72-hour orphan TTL, the action became `gcActionOrphan`.
5. `cleanTaskDir` called `os.RemoveAll` without first proving that Multica had created or owned the directory.

Sanitized incident log excerpt:

```text
INF gc: orphan directory dir=<shared-project-root>/<repository>/data reason="no meta" age=73h0m0s
INF gc: removed dir=<shared-project-root>/<repository>/data bytes_reclaimed=25063
```

### Incident impact

This bug deleted more than 20 GiB of the affected user's files. The confirmed user-visible impact included:

- 2,009 tracked files removed from a repository working tree; these could be restored from Git.
- Git-ignored data-lake, database, and provider output that could not be restored from Git.
- Prior Qlib and Nautilus run artifacts and `metrics.json` files, preventing direct comparison with the lost runs.
- Source trees, `.git` directories, virtual environments, application directories, and multiple `data` directories recorded in the daemon deletion log.

Separately, the daemon reported approximately 38.3 GiB reclaimed across the affected GC cycles. That broader counter includes regenerable cache data, so it is not used as the confirmed user-data-loss figure; the confirmed incident impact is more than 20 GiB of deleted user files.

A configurable working root is not permission to recursively delete unrelated user data. Multica owns the burden of proving that a path is daemon-managed before mutation. The destructive boundary now requires the Prepare-time `.task_owner` marker, which is written before task content and is already treated elsewhere as authoritative env-root provenance.

## Related Issue

No existing issue. Discovered from a real-world data-loss incident on 2026-08-31.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `execenv.ValidateEnvRootOwnerPath` to prove that an owner names the exact two-level workspace/task path being mutated.
- Require valid `.task_owner` provenance before every GC mutation, including artifact-only cleanup.
- Revalidate ownership after acquiring the env-root exclusion lock, closing the check/mutation race.
- Revalidate immediately before `os.RemoveAll` as defense in depth.
- Preserve compatibility with legacy task-only owner markers only when valid GC metadata supplies the missing workspace identity.
- Intentionally skip directories with no owner marker, malformed ownership, or owner/path disagreement.
- Add an incident-shaped regression test where `WorkspacesRoot` points at a source tree and `<repository>/data/research.sqlite` must survive even with `GCOrphanTTL=0`.
- Add owner/path mismatch and legacy-owner compatibility tests.

## Why this boundary?

Directory age and missing completion metadata are lifecycle hints, not ownership proof. Directory-name validation alone is also insufficient because readable workspace/task names intentionally contain user-controlled prefixes.

`.task_owner` is the right existing primitive:

- it is written before any task content;
- current markers include stable workspace and task IDs;
- the path suffixes can be verified against those IDs;
- partially prepared roots therefore remain collectible;
- unrelated directories under a shared root remain untouched.

The only intentional retention change is that a legacy task-only owner with neither workspace identity nor valid GC metadata is retained instead of recursively deleted. That small possible disk leak is the fail-closed outcome.

## How to Test

1. Run the incident regression:

   ```bash
   cd server
   go test -race ./internal/daemon -run 'TestRunGC_SharedRootPreservesForeignDirectories|TestCleanTaskDir_(RefusesOwnerPathMismatch|AcceptsLegacyOwnerWithGCWorkspaceIdentity|RemovesDirectory|RemovesStableRootRecord)' -count=1
   ```

2. Run the daemon suites and vet:

   ```bash
   cd server
   go test ./internal/daemon/... -count=1
   go vet ./internal/daemon/...
   ```

3. Verify the CLI package still builds and tests:

   ```bash
   cd server
   go test ./cmd/multica -count=1
   ```

All commands pass locally.

## Risks and compatibility

- Normal current task roots retain existing GC behavior because they already have `.task_owner` with both stable IDs.
- Legacy plain task-ID markers remain eligible when `.gc_meta.json` provides the matching workspace ID.
- Unprovable directories are retained and warned about. This trades a bounded potential disk leak for preventing unbounded user-data deletion.
- Validation happens before opening the GC lock, so foreign directories are not modified merely by being scanned.

## Checklist

- [x] UI screenshots are not applicable
- [x] Documentation changes are not required; this hardens an internal destructive boundary

## AI Disclosure

**AI tool used:** Codex GPT-5.6 Sol through Multica

**Prompt / approach:** Traced a real daemon log from `gc: orphan directory` through `shouldCleanTaskDir`, `applyGCAction`, and `cleanTaskDir`; reproduced the legitimate shared-root layout in a regression test; reused the existing Prepare-time `.task_owner` provenance instead of adding a parallel ownership system; then ran the daemon test tree, vet, targeted race tests, and CLI tests.

## Screenshots

The original incident screenshot contains local environment context and is not attached publicly. The causal log lines and impact are reproduced in sanitized form above.

### DIFF
--- a/server/internal/daemon/execenv/root_identity.go
+++ b/server/internal/daemon/execenv/root_identity.go
@@ -167,6 +167,40 @@ func validTaskRootSegment(segment, id string, workspace bool) bool {
 	return strings.HasSuffix(segment, "-"+key)
 }
 
+// ValidateEnvRootOwnerPath proves that envRoot is the two-level task root
+// named by owner under workspacesRoot. GC callers use this before every
+// mutation: configuration may point WorkspacesRoot at an arbitrary user-owned
+// tree, and a directory's age or missing metadata is not proof that the daemon
+// created it.
+func ValidateEnvRootOwnerPath(workspacesRoot, envRoot string, owner EnvRootOwner) error {
+	if strings.TrimSpace(workspacesRoot) == "" || strings.TrimSpace(envRoot) == "" {
+		return errors.New("execenv: workspaces root and env root are required")
+	}
+	if owner.WorkspaceID == "" || owner.TaskID == "" {
+		return errors.New("execenv: env root owner must name both workspace and task")
+	}
+
+	relative, err := filepath.Rel(workspacesRoot, envRoot)
+	if err != nil {
+		return fmt.Errorf("execenv: make env root relative: %w", err)
+	}
+	relative = filepath.Clean(relative)
+	if relative == "." || filepath.IsAbs(relative) {
+		return fmt.Errorf("execenv: env root %s is not a task directory below %s", envRoot, workspacesRoot)
+	}
+	parts := strings.Split(relative, string(filepath.Separator))
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" || parts[0] == ".." || parts[1] == ".." {
+		return fmt.Errorf("execenv: env root %s is not exactly two levels below %s", envRoot, workspacesRoot)
+	}
+	if !validTaskRootSegment(parts[0], owner.WorkspaceID, true) {
+		return fmt.Errorf("execenv: workspace directory %q does not match owner %s", parts[0], owner.WorkspaceID)
+	}
+	if !validTaskRootSegment(parts[1], owner.TaskID, false) {
+		return fmt.Errorf("execenv: task directory %q does not match owner %s", parts[1], owner.TaskID)
+	}
+	return nil
+}
+
 // RemoveRootDirRecord removes the stable index after GC has reclaimed a
 // terminal task root. It verifies that the record still points at envRoot so a
 // stale cleanup can never remove another task's identity.
--- a/server/internal/daemon/gc.go
+++ b/server/internal/daemon/gc.go
@@ -872,12 +872,17 @@ func (d *Daemon) gcTaskDirOwner(taskDir string) (*execenv.EnvRootOwner, error) {
 // reclaimed bytes, and returns that count for the cycle summary. A failed or
 // refused removal reports removed=false.
 func (d *Daemon) cleanTaskDir(taskDir string) (bytes int64, removed bool) {
+	// Measure first, prove ownership second. dirSize walks the entire tree,
+	// which on a large task directory takes long enough for the validated
+	// directory to be replaced underneath us — checking before that walk would
+	// hand RemoveAll a path last proven ours tens of seconds earlier. This is
+	// the defense-in-depth check, so it sits immediately before the removal.
+	bytes = dirSize(taskDir)
 	owner, ownerErr := d.gcTaskDirOwner(taskDir)
 	if ownerErr != nil {
 		d.logger.Warn("gc: refusing to remove unowned task directory", "dir", taskDir, "error", ownerErr)
 		return 0, false
 	}
-	bytes = dirSize(taskDir)
 	if err := os.RemoveAll(taskDir); err != nil {
 		d.logger.Warn("gc: remove task dir failed", "dir", taskDir, "error", err)
 		return 0, false
--- a/server/internal/daemon/gc.go
+++ b/server/internal/daemon/gc.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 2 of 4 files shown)


mameikagou:
Thanks — `0c9c26a` usefully pins the static ownership gate for all four mutating actions. There is still a separate post-gate path-replacement race in managed-artifact cleanup, though.

I reproduced it against the current head (`0c9c26af8d75bc3aaa1098a8da69a7b0defdd335`) through the full `applyGCAction` path, not by calling the cleaner directly. That means both `gcTaskDirOwner` checks and `reserveEnvRootForGC` run successfully before the replacement.

Environment:

```text
Linux 6.6.114.1-microsoft-standard-WSL2 x86_64
go version go1.26.0 linux/amd64
```

Test procedure:

1. Create a valid owned root at `ws/task`.
2. Put 20,000 entries under `codex-home/.sandbox-bin` so `dirSize` has a measurable walk.
3. While that walk is running, rename `codex-home` aside and replace it with a symlink to `foreign-codex-home`, which contains `.sandbox-bin/DO_NOT_DELETE`.
4. Run `applyGCAction(taskDir, gcActionCleanManagedArtifacts, ...)` and assert that the foreign sentinel survives.

Command and result:

```text
go test ./internal/daemon \
  -run TestReviewApplyGCActionManagedArtifactParentReplacement \
  -count=10

10/10 failed: foreign-codex-home/.sandbox-bin/DO_NOT_DELETE was removed
```

Representative output:

```text
INFO gc: artifact removed path=.../ws/task/codex-home/.sandbox-bin bytes=0
--- FAIL: TestReviewApplyGCActionManagedArtifactParentReplacement
    foreign managed artifact was deleted after parent replacement:
    stat .../foreign-codex-home/.sandbox-bin/DO_NOT_DELETE: no such file or directory
```

The remaining sequence is:

```text
applyGCAction: ownership check + reservation + ownership re-check
cleanManagedTaskArtifacts: managedArtifactTarget validates components
dirSize(target): parent is replaced during this walk
os.RemoveAll(target): resolves the same path string through the replacement
```

The new test does not catch this because the filesystem stays unchanged between the owner gate and removal. `reserveEnvRootForGC` coordinates daemon activity in-process, but it cannot prevent another process or the user from moving filesystem paths.

At minimum, please revalidate the root and every target path component after the size walk and immediately before each `RemoveAll`, rejecting symlinks/junctions or changed identity, and add the parent-replacement test as a regression. The same post-walk rule should cover both managed-artifact and pattern-artifact deletion paths.

Bohan-J:
Thanks @mameikagou — merging this.

This is a real fix for a real incident, and the writeup made it cheap to verify: the sanitized log excerpt, the causal chain from `runGC` through `shouldCleanTaskDir` → `applyGCAction` → `cleanTaskDir`, and the incident-shaped regression with `GCOrphanTTL=0` all lined up with what the code actually does. Reusing the Prepare-time `.task_owner` instead of inventing a parallel ownership system was the right call.

Two commits were pushed to the branch during review:

- **`fix(daemon): prove ownership after the size walk, not before it`** — the final revalidation in `cleanTaskDir` ran *before* `dirSize`, which walks the entire tree. On a large task directory that left `os.RemoveAll` acting on a path last proven owned tens of seconds earlier, and it didn't match this PR's own "immediately before `os.RemoveAll`" claim. Reordered to measure → validate → remove.
- **`test(daemon): pin the owner gate for all four mutating actions`** — a table test asserting an unowned directory survives full clean, orphan, pattern artifact and managed artifact cleanup. Worth recording why it earns its place: dropping only the first gate in `applyGCAction` leaves it green, because `cleanTaskDir` revalidates on its own. Dropping both fails exactly the `artifacts` and `managed_artifacts` subtests — artifact-only cleanup was the path with no second line of defense.

Deliberately left out of scope: a regression test for the residual window between the owner read and `RemoveAll` (deterministically testing it needs injectable hooks in production code), and inode identity pinning (`RemoveAll` still resolves the top level by path, so closing that properly means fd-based deletion — a separate change).

One known consequence to flag for whoever picks up the follow-up: directories with no `.task_owner` at all — older layouts that carry only a `.gc_meta.json` with `issue_id` + `workspace_id` and no `task_id` — are now retained permanently, and they can't be backfilled because there is no task identity left to reconstruct. On one real machine that is roughly half the historical task directories. That is the fail-closed trade this PR consciously makes; surfacing the retained bytes in disk-usage output should be tracked separately.

Thanks again for tracing this from the daemon log rather than just reporting the symptom.

PR #7799, opened 2026-08-31
MUL-6872: fix(web): keep inline image blob URLs stable across re-entry to prevent flicker

### LINKED ISSUE #7741: [Bug]: 进入任务详情页返回，在进入任务详情页内容闪烁明显
### Deployment type

Official App

### Area

Web app

### What happened?

https://github.com/user-attachments/assets/4c7d0d5b-4667-4d15-aa8a-532f8949cce8 
进入任务详情页快速返回上一页，在进入任务详情页内容闪烁明显，多次重复操作可以出现闪烁明显，带有图片的详情页闪烁较明显。
系统：Mac，Chrome 版本 151.0.7922.175

### Steps to reproduce

进入任务详情页快速返回上一页任务列表，在进入任意任务详情页，或者上一个任务详情页，会出现内容闪烁情况

### Screenshots (optional)

_No response_

```shell

```

makesomethingshit:
## What does this PR do?

Fixes repeated inline-image flicker when reopening a task detail view that uses the authenticated blob fallback path.

The implementation keeps each attachment's `blob:` URL stable across re-entry while bounding its lifetime. It reference-counts active consumers, starts a five-minute grace-period GC after the last consumer unmounts, cancels that GC if the same attachment re-enters, and revokes and removes the URL when the grace period expires. Attachment-ID guards prevent an old image from surviving an A→B transition when B fails to load.

Cached private bytes remain behind the existing authorization boundary. A cached object URL is returned only when the current QueryClient has successfully fetched authenticated metadata for the same attachment ID. After logout or account switch, a new client's pending or rejected metadata request falls back to the durable attachment URL and cannot reuse the previous account's cached private image, even while that object URL is still within its five-minute GC window.

## Related Issues

Closes #7741

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Refactoring (no functional changes)
- [ ] Documentation update
- [x] Tests

## Changes Made

- Cache blob URLs by attachment ID and reference-count active consumers.
- Delay revocation for five minutes after the final unmount, and cancel pending GC on re-entry.
- Revoke and delete expired cached URLs so long image-heavy sessions remain bounded.
- Guard attachment-ID transitions so a failed new image cannot display the previous image.
- Require authenticated metadata for the same attachment ID before returning a cached private blob URL.
- Add regression coverage for delayed GC, GC cancellation, re-entry reuse, concurrent consumers, ID changes, and cross-account cache isolation.

## How to Test

```bash
pnpm --filter @multica/views exec vitest run editor/attachment.test.tsx
```

The focused regression suite passes locally.

## Screenshots (if applicable)

Not included. The authenticated blob fallback depends on proxy/MinIO or split-origin conditions and is covered by the focused regression suite.

## Checklist

- [x] I have thought through the edge cases and failure modes
- [x] I have added tests that prove my fix is effective
- [ ] If my change affects the UI, I have included screenshots — not included; the environment-specific fallback is covered by regression tests
- [x] I have made corresponding changes to the documentation — N/A, no documented contract changed
- [x] My changes do not introduce a new runtime, tool, or UI tab — N/A
- [x] User-facing text uses Chinese characters with a Traditional/Simplified conversion helper — N/A, no user-facing text added
- [x] I have documented known risks and lifecycle behavior
- [x] I have performed a self-review of my code

### AI tool used

Multica Agent / ChatGPT

### Prompt / approach

Traced the authenticated inline-image lifecycle, preserved the existing authorization boundary, and combined stable re-entry with reference-counted five-minute cleanup and focused regression coverage.


makesomethingshit (commented):
## 차단

### `blob:` URL 캐시가 기존 메모리 해제 계약을 제거합니다

근거: `packages/views/editor/hooks/use-inline-media-url.ts:128-137`의 모듈 `Map`은 항목을 삭제하지 않고, 변경된 effect에는 cleanup도 없어 `URL.revokeObjectURL`이 다시 호출되지 않습니다. 따라서 서로 다른 이미지를 볼 때마다 object URL과 그 Blob 바이트가 renderer 종료까지 누적됩니다. 바로 위 `INLINE_BLOB_GC_MS`가 장시간 이미지 스크롤의 메모리 고정을 막으려는 5분 GC인데 이 캐시가 그 경계를 무력화합니다. 실제 CI도 `editor/attachment.test.tsx:553`의 "unmount 후 revoke" 계약에서 0회 호출로 실패했습니다.

수정안: attachment id별 참조 수를 관리하고 마지막 사용자가 unmount된 뒤 `INLINE_BLOB_GC_MS` 동안만 URL을 유지한 다음 `URL.revokeObjectURL`과 `Map.delete`를 수행해 주세요. 그 전에 재진입하면 예약된 정리를 취소하면 flicker 개선과 bounded memory를 함께 지킬 수 있습니다. 재진입 시 URL 재사용과 지연 정리를 각각 검증하는 작은 테스트도 필요합니다.

## 중요

### attachment id 전환 시 이전 이미지 URL이 실패 경로에 남습니다

근거: `packages/views/editor/hooks/use-inline-media-url.ts:128-134`의 state initializer는 최초 mount에만 실행되고, 이후 id가 바뀌어 새 `blob`이 아직 없거나 요청이 실패하면 effect가 조기 반환해 이전 id의 `url`을 지우지 않습니다. `attachment-preview-modal.tsx:457-464`는 갤러리 이동 때 `PreviewPanel`을 의도적으로 재사용하므로, proxy 모드에서 A→B 이동 후 B의 blob fetch가 실패하면 B 메타데이터 아래 A 이미지가 계속 보이고 `onImageError`도 호출되지 않습니다.

수정안: state/cache 값을 id와 함께 보관하고 현재 id와 일치할 때만 반환하거나, id 변경 시 현재 id의 캐시 값(없으면 빈 값)으로 전환해 주세요. 같은 컴포넌트를 A→B로 rerender하고 B fetch를 reject시키는 회귀 테스트가 최소 검증입니다.

## 전체 판정

수정 요청

makesomethingshit:
Addressed the two earlier review findings in the latest commits:

- blob URLs are now reference-counted and revoked after a five-minute grace period
- pending cleanup is cancelled on re-entry
- cached URLs are guarded by the current attachment id
- regressions cover delayed cleanup, cache reuse, concurrent consumers, and failed A → B transitions

CI is green. The remaining presentation item is a before/after recording from the affected deployment path. The PR description now scopes the fix specifically to the authenticated blob fallback path.

multica-eve (changes requested):
When a member signs out or switches accounts, private images from the previous account must disappear immediately. This change correctly removes the flicker when returning to an image-heavy issue, but the current version can keep a previously viewed private image available to the next account for up to five minutes.

## Product assessment

- **Is the problem real? Yes.** #7741 includes reproducible steps and a recording, and the existing re-entry behavior explains the visible fallback frame.
- **Is a durable mechanism worthwhile? Yes.** Reusing an already loaded image within the same signed-in session removes a frequent reading disruption without adding settings or migration work.
- **Would we solve it this way? Yes, with the existing authorization boundary preserved.** Reuse the existing five-minute resource lifetime within one signed-in session, and invalidate it synchronously when identity changes.

**Product verdict: accept the direction.**

## Must-fix: the new cache survives the logout boundary

`packages/views/editor/hooks/use-inline-media-url.ts:129-135` returns a module-cached `blob:` URL by attachment ID even when the current query cache has no successfully authorized `fresh` metadata. `packages/views/editor/hooks/use-inline-media-url.ts:151-179` then keeps that URL for five minutes after the last consumer unmounts.

Logout explicitly calls `queryClient.clear()` at `packages/views/auth/use-logout.ts:65` to prevent data from crossing into the next account, but the three module-level maps introduced here are outside that lifecycle and are not cleared. As a result, after account A views a private image and signs out, account B can render the same durable attachment URL and receive A's cached bytes synchronously while B's metadata request is still pending or has been rejected. An attachment UUID is an identifier, not an authorization secret.

Please either:

1. reuse a cached object URL only when the current query cache already contains authorized metadata for the same attachment, or
2. synchronously revoke and clear this cache on logout/account switch.

Add a negative regression test for: A loads and unmounts the image → the query cache is cleared to simulate logout → B renders the same attachment ID with a new `QueryClient` while metadata is pending/rejected → A's `blob:` URL must not be returned. The current A→B test changes attachment IDs within one session and does not cover this boundary.

The reference counting, delayed cleanup, cleanup cancellation, concurrent-consumer handling, and different-ID failure guard otherwise address the earlier lifecycle findings.

Validation run against head `3e0e6683`:

- affected test suite: 35/35 passed
- `@multica/views` typecheck: passed
- scoped ESLint: passed
- `git diff --check`: passed

**Verdict: Request changes.** The product direction is correct, but cross-account private-image reuse is a merge blocker.

makesomethingshit:
Hi maintainers, kindly requesting a re-review of PR #7799 for #7741 when you have a moment.

**Context:** Re-entering an image-heavy issue detail page flashes visible content, especially on deployments where inline images fall back to an authenticated `blob:` download (proxy / MinIO, split-origin).

**Root cause:** `useObjectURL` created a new object URL in an effect on every mount. Even with the Blob cached in TanStack Query, the first render used `pickedUrl` and switched to a new `blob:` URL one frame later.

**Fix (head `ffc9575`):**
- Cache one `blob:` URL per attachment id and reuse it synchronously on short re-entry (no 1-frame fallback)
- Ref-counted 5-minute delayed cleanup (`INLINE_BLOB_GC_MS`): `revokeObjectURL` + `Map.delete` 5 min after last consumer unmounts, cancelled on re-entry
- Guard cached URLs by current attachment id so A→B gallery navigation never shows A under B if B's fetch fails
- Gated on authenticated metadata to prevent cross-account cache reuse

**Verification:**
- `pnpm --filter @multica/views exec vitest run editor/attachment.test.tsx` — 35/35 passed (covers delayed revoke, cancel on re-enter, sync reuse, concurrent consumers, A→B rerender)
- `pnpm --filter @multica/views lint` — 0 errors
- Manual: list → detail → back → detail repeatedly with authenticated `blob:` fallback — image URL remains stable, no flash

PR: https://github.com/multica-ai/multica/pull/7799 — ready for review. Thank you!

PR #7805, opened 2026-08-31
MUL-6813: fix: surface private runtime owner mismatch as explicit failure

### LINKED ISSUE #7727: Tasks for agents owned by a different user than the runtime owner are silently stuck in `queued` forever (no error anywhere)
## Summary

When a daemon's authenticated account differs from the `owner_id` of an existing agent, all **issue-assigned/comment tasks** for that agent are never claimed by the runtime. The task sits in `queued` until it eventually fails with "task expired in queue". The daemon logs nothing — no fetch, no error, no rejection reason — so from the user's perspective the agent just "hangs" forever. Creating a **copy** of the same agent (copy is owned by the daemon's account) makes tasks run instantly.

## Environment

- Multica Cloud (api.multica.ai)
- CLI/daemon: multica 0.4.36
- Runtime provider: pi 0.84.3, Linux, daemon running as root
- One runtime, bound to the daemon's account

## Steps to reproduce

1. Account A creates agents and binds them to a runtime. Later, the daemon on that runtime re-authenticates / is re-registered under **account B** (in our case after a fresh `multica login --token` + `multica daemon start` during a machine re-setup). The agents remain owned by A, the runtime is owned by B.
2. Send a chat message to one of A's agents → it runs fine.
3. Assign an issue to A's agent (or @-mention it in a comment) → wakeup is delivered, but the task is never fetched.
4. Create a copy of the same agent while authenticated as B (`multica agent copy`) and assign an identical issue → it is picked up and starts within ~1 second.

## Actual behavior

- Task stays `queued` with `dispatched_at: null, error: null`.
- Server eventually marks it `failed` with error "task expired in queue" (five delivery attempts of the same wakeup were observed over 2 hours, all silently dropped).
- Daemon log shows only: `task wakeup received ... task_id=...` followed by nothing — no `task received`, no `picked task`, no error line.
- Archiving + restoring the agent does **not** help (owner doesn't change).
- Chat tasks for the same agent run normally (they appear to bypass the per-agent claim fence), which makes the bug look intermittent and very confusing to debug.

## Expected behavior

Either the task should run, or the failure should be loud and explicit: an error on the task ("agent owner X is not authorized to use private runtime Y owned by Z"), a visible state in the UI, and/or a daemon-side log line with the rejection reason. A silent `queued`-forever state with zero feedback is the core problem.

## Root cause (from the open-source server code)

`ListQueuedClaimCandidatesByRuntime` in `server/pkg/db/queries/agent.sql` filters claim candidates with an authorization fence:

```sql
AND EXISTS (
    SELECT 1 FROM agent a
    JOIN agent_runtime r ON r.id = atq.runtime_id
    WHERE a.id = atq.agent_id
      AND a.runtime_id = atq.runtime_id
      AND ( r.visibility = 'public'
            OR ( r.visibility = 'private'
                 AND ( r.owner_id IS NULL OR a.owner_id IS NULL
                       OR r.owner_id = a.owner_id ) ) )
)
```

If `agent.owner_id != runtime.owner_id` (private runtime), the agent's queued tasks are excluded from the candidate list. The daemon receives the wakeup hint, finds no candidates, and returns an empty claim — nothing is logged on either side. The binary even contains the string `bound to a different account`, but it is never emitted in this path.

Note: this state is easy to reach legitimately (daemon re-setup / re-login with a different user's PAT, account migration, shared machine), and there is **no CLI command to transfer agent ownership** or change runtime visibility, so the only workaround is recreating/copying every agent and re-pointing squads/autopilots.

## Suggested fixes (any of)

1. Surface a terminal task error when a queued task fails the claim-authorization fence for longer than N minutes, instead of letting it expire silently.
2. Log the rejection (server and/or daemon) with the reason, e.g. "agent owner mismatch: agent owned by X, private runtime owned by Y".
3. Add `multica agent update --owner <user>` (ownership transfer) or make runtime visibility configurable via CLI.
4. Consider delivering chat-vs-issue behavior consistently — currently chat works for the same agent, which masks the issue.

makesomethingshit:
## What does this PR do?

Fixes runtime/agent ownership handling for private runtimes at admission and daemon claim time.

On current `main`, a queued task whose agent owner differs from its private runtime owner is excluded by the claim candidates and can remain queued indefinitely; it does not naturally expire into a terminal state. Queued claim queries now use a deliberate carve-out: they admit the row to the handler, which applies the existing `FailTask` settlement path while continuing the poll. The dispatched stale-reclaim queries intentionally keep their existing owner fence, so an already-dispatched task is never reclaimed onto an incompatible runtime.

Align persisted task failure semantics with admission semantics: queued private-runtime ownership mismatches now settle as `runtime_access_denied`, allowing current Web/Desktop/Mobile clients to show the same actionable recovery guidance instead of "Task identity mismatch" or generic "Failed". Actual task/agent identity violations continue to use `invalid_task_identity`.

## Related Issue

Closes #7727

Internal tracking: PUCK-89, PUCK-132

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Adds the stable `runtime_access_denied` dispatch reason and admission fallback.
- Aligns service admission with daemon runtime ownership: a private runtime with an owner blocks an ownerless or differently-owned agent, while same-owner, public-runtime, and ownerless-runtime behavior remains unchanged. Durable assignment and mention notices include actionable public-runtime and rebind/copy recovery paths.
- Removes the claim-time sweep query and helper. Queued claim candidates are admitted for handler settlement; dispatched stale-reclaim queries retain the owner fence.
- Routes mismatches through `buildClaimedTaskResponse` -> `FailTask`, preserving issue events/comments, chat assistant failure messages, and onboarding release behavior.
- A singular claim returns `200 {task:null}` after successful settlement; settlement/requeue failures still return the existing HTTP error. Batch claims continue past settled failures and return valid tasks from the same poll.
- Regenerates `server/pkg/db/generated/agent.sql.go` with sqlc.
- Removes the earlier unrelated `server/internal/service/task.go` empty-claim version-sampling diff; that file now matches `main` for this behavior.

### Review fixes (3 blockers)

- **Final delivery authorization (TOCTOU):** the claim path reads the runtime before claiming, but a concurrent re-registration can flip `runtime.owner_id` between that read and final delivery. `FinalizeTaskClaim` now runs a caller-supplied `authorize` closure inside its transaction; the shared `finalizeClaimDelivery` gate re-locks the runtime row (`FOR UPDATE`), re-reads the agent, and re-verifies the private-runtime owner fence against CURRENT ownership before the task token commits. The authorization decision and token mint commit atomically, so the owner authorized against is the owner the tokens were minted for. A mismatched task is settled through the existing `failClaimedTaskBeforeLaunch` -> `FailTask` path and never returned to the daemon: singular keeps `200 {task:null}`, batch skips the task and keeps returning valid tasks from the same poll. Regressions cover both paths (`TestClaimTaskByRuntime_RuntimeOwnerChangedAfterClaimNeverDelivered`, `TestClaimTasksByRuntime_OwnerChangedAfterSnapshotSettlesMismatchReturnsValid`).
- **Client copy for `runtime_access_denied`:** current Web/Desktop/Mobile surfaces now resolve the reason to dedicated actionable recovery copy (make the runtime public, or rebind/copy the agent to a runtime its owner can use) via the existing dispatch-reason mapping mechanisms: issues blocked-trigger copy, both chat send-failure toasts, the chat failure-reason map, the mobile dispatch-reason helper, and the autopilot run-now toast. Localized keys added for en/ko/ja/zh-Hans; unknown reasons still degrade to the existing generic fallbacks.

### PUCK-132: persisted failure reason alignment (latest review blocker)

- **New canonical reason `runtime_access_denied`** (`server/pkg/taskfailure/failure.go`): private runtime/agent ownership authorization failure — the agent process never launches, and it is permanent/non-retryable under the existing retry policy. Registered in the platform reasons group.
- **Settlement alignment** (`server/internal/handler/daemon.go`): only the two ownership-denial settlement paths changed — the response-assembly re-check (owner mismatch / ownerless agent) and the delivery-gate authz default branch (owner mismatch, ownerless agent, missing current runtime owner). They now persist `runtime_access_denied` instead of `invalid_task_identity`. The four real identity paths (agent deleted, agent rebound to another runtime, response identity mismatch, `error_agent_runtime_changed`) continue to use `invalid_task_identity`.
- **Client mapping for the persisted reason:** Web/Desktop map persisted `runtime_access_denied` through the shared task/chat failure mapping in `@multica/views`, reusing the existing admission recovery copy semantics (localized keys for en/ko/ja/zh-Hans). Mobile reuses the admission copy through a single shared helper for dispatch reason, persisted chat failure label, and run failure badge. Retry is not presented as the primary remedy; unknown reasons keep generic fallbacks.
- **Regression tests:** (A) pre-existing queued issue task with owner mismatch → `failed`/`runtime_access_denied`; (B) pre-existing queued chat task → same reason + assistant failure message via the existing `FailTask` path; (C) agent rebind (real identity mismatch) → still `invalid_task_identity`, pinning the boundary between the two reasons.

## Behavior / Risk

- New work targeting an ownerless or foreign-owned agent on a private owned runtime is refused during admission with `runtime_access_denied` and is not queued.
- Pre-existing queued mismatches are settled deterministically on the next claim instead of remaining permanently queued, and now persist the same semantic reason as admission (`runtime_access_denied`).
- A valid task sharing that runtime is still claimable in the same singular/batch polling flow.
- The queued carve-out is intentionally asymmetric with dispatched reclaim: the latter continues to enforce runtime/agent ownership.
- Delivery is authorized against the runtime's CURRENT ownership at the finalize boundary, not the claim-time snapshot; a runtime owner change racing a claim cannot deliver the task to a daemon under a stale decision.
- Clients render `runtime_access_denied` with dedicated non-retry recovery copy — both for the admission-time dispatch reason and the persisted task failure reason; old clients and unknown reasons keep the generic fallback.
- The reverse ownership-repair race noted by the maintainer is explicitly out of scope for this PR (non-blocking).

## How to Test

```bash
cd server
sqlc generate
gofmt -w \
  internal/dispatch/reason.go \
  internal/handler/admission.go \
  internal/handler/admission_security_mul4525_test.go \
  internal/handler/admission_test.go \
  internal/handler/agent_access_test.go \
  internal/handler/comment.go \
  internal/handler/daemon.go \
  internal/handler/daemon_runtime_access_test.go \
  internal/handler/issue.go \
  internal/handler/runtime_access_denied_test.go \
  internal/service/agent_ready.go \
  internal/service/issue.go \
  internal/service/reason_code_test.go \
  internal/service/runtime_claim_access_test.go \
  pkg/db/generated/agent.sql.go

go test ./internal/service -run 'TestAgentReadinessVerdict|TestRuntimeUnusableNotice' -count=1
go test ./internal/handler -run 'Test.*RuntimeAccessDenied' -count=1
go test ./internal/service -run 'TestRuntimeAccessGatesQueuedTaskClaims|TestRuntimeAccessGatesStaleDispatchReclaim|TestClaim.*Runtime' -count=1
go test ./internal/handler -run 'Test.*RuntimeAccess|TestClaimTaskByRuntime_Ownerless' -count=1
go test ./internal/handler -run 'TestSendChatMessage_InvokeRevokedAfterSessionCreate' -count=1

# PUCK-132 focused verification
go test ./pkg/taskfailure/ -count=1
go test ./internal/handler -run 'OwnerlessAgent|SettlesPrivateOwnerMismatch|ChatFailure|SettlesMismatchAndReturnsValidTask|AgentOwnerChangedAfterClaim|AgentReboundAfterClaim|RuntimeOwnerChangedAfterClaimNeverDelivered|CurrentRuntimeOwnerMissing|SettlementFailureIsUnsettled|OwnerChangedAfterSnapshot' -count=1
go vet ./...
git diff --check

# Client (PUCK-132)
pnpm --filter @multica/views test -- task-failure chat-message-list blocked-trigger-copy
pnpm --filter @multica/mobile test -- failure-label dispatch-reason run-badge
```

The focused service/handler tests, taskfailure package tests, client vitest suites (Web 44/44 + locale parity 26/26, Mobile 15/15), typechecks, sqlc generation (no unintended diff — no SQL changes), formatting, vet, and diff checks pass. The full `go test ./internal/service ./internal/handler` run is still affected by unrelated baseline fixture drift (built-in skill frontmatter, local skill path normalization, unsafe archive-path expectation, and workspace deletion manifest entries).

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — N/A, copy-only mappings covered by localized tests
- [x] I have updated relevant documentation to reflect my changes — N/A, no documented external contract changed (docs `tasks.*.mdx` failure-reason table row for `runtime_access_denied` noted as follow-up)

## AI Disclosure

**AI tool used:** Multica Agent / ChatGPT

**Prompt / approach:** Implemented the runtime access admission and claim settlement changes, then verified the singular, batch, issue, chat, and mention paths with focused tests. The PUCK-132 follow-up (persisted `runtime_access_denied` alignment + client recovery copy mapping) was implemented, cross-verified, and reviewed through the same Multica agent pipeline with real-DB handler regressions and client vitest suites.

## Screenshots (optional)

N/A — copy-only mappings, covered by localized tests.

### Follow-up review fixes

- Final delivery passes a pointer to the in-transaction task-token params. The shared gate locks the runtime row with the existing `LockAgentRuntime` query, locks the agent second, authorizes against that current snapshot, and sets `task_token.user_id` from the locked current runtime owner before `CreateTaskToken` runs. A missing locked owner rejects delivery and never mints a stale token.
- The singular claim path returns `200 {task:null}` only when the rejected task was successfully settled. If `FailTask` or the exact requeue fails, it preserves the existing HTTP error contract; batch claims continue past either delivery failure and distinguish unsettled failures in logs.
- Removed the duplicate `LockAgentRuntimeByID` SQL/sqlc method; `LockAgentRuntime` is the single repository query. Added regressions for public-runtime owner changes (token identity), owner loss, and singular settlement failure (`500`, no task/token, requeue attempted).

- **Latest delivery identity/race regressions:** the shared finalize gate clears claim-time requesting-user fields and repopulates them from the locked runtime owner (the same owner used for `task_token.user_id`); an owner lookup failure cannot leak stale fields. Singular and batch tests retain the initial runtime snapshot, mutate runtime ownership/visibility after claim, and verify mismatch settlement/omission with valid-task continuation.


Bohan-J (changes requested):
## Request changes — MUL-6813 / #7727

The diagnosis is right and the claim fence is correctly left intact. But the fix sits one layer too low: it settles tasks that should never have been created, and it does the settling from inside the claim hot path. Please land **both** halves below in this PR — together they are smaller than what is here today.

Also, one factual correction for the PR description: on current `main` this state does **not** "remain queued until expiry". Since MUL-6558, `ExpireStaleQueuedTasks` (`server/pkg/db/queries/agent.sql:1396`) only fails queued rows whose runtime stopped heartbeating. On an owner mismatch the runtime is alive and heartbeating, so nothing ever expires the task — it is queued *forever*. The reporter's "task expired in queue" was 0.4.36 behaviour. Worth fixing in the description because it raises the severity.

---

### Required change 1 — reject at admission (the missing half)

Settling at claim time turns "silently stuck" into "fails ~30s later", but the system still accepts work it knows it cannot do. An hourly autopilot on a mismatched agent goes from one silent hang to one failure comment per hour. The invariant needs a representative at the entry, not only at the exit.

`service.AgentReadiness` (`server/internal/service/agent_ready.go:105`) is exactly the right place: it is already the single shared admission gate for issue-assign, comment mention, chat, autopilot, squad-leader and quick-create, and it **already holds both the agent row and the runtime row**, so this costs zero extra queries.

Concretely:

1. `server/internal/dispatch/reason.go` — add `ReasonRuntimeAccessDenied ReasonCode = "runtime_access_denied"` with a doc comment. (`reason_code` is an open string on the wire — `packages/core/api/schemas.ts:2265` — so no client change is required.)
2. `server/internal/handler/admission.go:59` — alias it; `:120` — add its fallback sentence.
3. `server/internal/service/agent_ready.go` — return `AgentBlocked` + the new reason when the runtime is private and its owner cannot run this agent. **Mirror the predicate at `server/internal/handler/daemon.go:2085` exactly**, including the `agent.owner_id IS NULL` arm — if admission checks a narrower condition than enforcement, you have just recreated the same divergence in the opposite direction. `runtimeVerdict` currently only takes the runtime row, so it needs the agent (or do the check in `AgentReadiness` before delegating).
4. Durable trace on assignment. Assignment has no response the assigner reads, so a refusal that vanishes is still a silent failure. Reuse the MUL-6164 mechanism: widen `server/internal/service/issue.go:735` and `server/internal/handler/issue.go:3883` (`verdict.Reason == ReasonRuntimeUnusable`) to cover the new code, and widen `server/internal/handler/comment.go:3073` so a refused @mention leaves the same trace. `RuntimeUnusableNotice` (`agent_ready.go:175`) needs a sibling text for this reason that names the two real repairs: the runtime owner makes the machine public, or the agent is rebound/copied onto a runtime its owner can use.
5. No changes needed in `chat.go:894`, `autopilot.go:952/1316` or `issue_trigger.go:183` — they branch on `verdict.Blocked()` generically and pick this up for free. Please don't touch them.

### Required change 2 — settle existing rows through the path that already exists

Drop `FailQueuedTasksRejectedByRuntimeAccess` and `failQueuedTasksRejectedByRuntimeAccess` entirely, along with both call sites in `ClaimTaskForRuntime` / `ClaimTasksForRuntimes` and the reverted version-sampling comments.

The mechanism is already in the tree, and the file already documents it as the intended pattern. `ClaimAgentTask` at `server/pkg/db/queries/agent.sql:796-799` says, about ownerless rows:

> Ownerless runtime/agent rows remain claimable only so the handler can settle them explicitly before daemon delivery; filtering them here would leave every task silently queued until the TTL.

That is precisely this bug's fix — it was just never extended from "ownerless" to "different owner". So instead of adding a sweep:

- Extend that carve-out to the owner-mismatch case in `ClaimAgentTask` (`:769`), `ListQueuedClaimCandidatesByRuntime` (`:2192`) and `ListQueuedClaimCandidatesByRuntimes` (`:2317`), so a mismatched row can be claimed *for settlement*.
- `server/internal/handler/daemon.go:2085` then catches it — that branch already exists, already carries the user-facing sentence, and already goes through `failClaimedTaskBeforeLaunch` → `TaskService.FailTask`. Nothing is delivered to the daemon.
- Leave the two reclaim queries (`:882`, `:926`) fenced — a dispatched row that turns mismatched is already covered by the dispatch-timeout sweeper — but say so in the PR description so the asymmetry reads as a decision.
- Regenerate `server/pkg/db/generated`.

Why this is strictly better than the current approach:

- **No hot-path write.** The current version runs an unconditional `UPDATE ... RETURNING` on every claim for any runtime that is not idle-cached. Moving the repair after the empty-claim fast path fixed the idle case, but every *busy* machine still pays an extra write transaction per poll. With the carve-out there is no sweep at all: the row simply becomes a candidate, is claimed once, and dies.
- **No `FailTask` bypass.** This is the substantive correctness problem with the current diff. The new SQL is not restricted to issue tasks, so chat and quick-create rows are bulk-flipped to `failed`, and then only `HandleFailedTasks` + `createAgentComment` run. But a chat task's assistant failure message and the onboarding-kickoff release happen *inside* the `FailTask` transaction (`server/internal/service/task.go:4881-4899`). Today's diff therefore produces a failed chat task with nothing at all in the conversation. Going through `failClaimedTaskBeforeLaunch` gets the chat message, the issue comment, events, audit and the non-retry policy for free — `invalid_task_identity` is already correctly non-retryable (`server/pkg/taskfailure/failure.go:134-139`), so there is no settle/retry loop.
- The `UPDATE` also has no `LIMIT`, unlike every other sweep in this file.

### One more thing to fix in the claim response

`buildClaimedTaskResponse` returning a failure makes the **singular** endpoint answer `403` (`server/internal/handler/daemon.go:3304-3308`). That was the right instinct to avoid in the PR description, but the fix belongs here rather than in a sweep: the daemon's legacy per-runtime loop treats an error as fatal for that poll round and abandons the remaining runtimes when nothing has been claimed yet (`server/internal/daemon/client.go:312-318`). With N mismatched rows on a machine you get N error-level poll rounds.

Please give `claimBuildFailure` a `settled bool` (or equivalent) meaning "this task is terminally settled; the poll itself is fine", and have the singular endpoint answer `200 {"task": null}` for it — the same shape the `repaired_stale_comment_plan` branch already uses at `daemon.go:3295`. The batch path already does the right thing (`daemon.go:1706-1712`, settle then `continue`).

### Tests

- `AgentReadiness`: private + mismatched owners → blocked with the new code; private + same owner → available; **public** + mismatched owners → available (the shareable case must not regress); private + ownerless agent → blocked; ownerless runtime → unchanged.
- Admission: assigning a mismatched agent creates **no** `agent_task_queue` row and leaves the system comment; `POST /chat` returns 409 with the reason code; an @mention records a blocked target with the reason.
- Claim: a pre-existing queued mismatched row is now a candidate, is settled via `FailTask` with the user-facing message, and the singular endpoint answers `200 task:null` rather than 403.
- Chat regression (this is the one today's diff would fail): a mismatched **chat** task settles *and* the session gets the assistant failure message.
- Batch: one mismatched row + one valid row on the same runtime → the mismatch is settled and the valid task is still returned in the same response.

The existing `runtime_claim_access_test.go` coverage of the ownerless path should stay.

### Out of scope for this PR

Not asking for them here, but they are the rest of the story and should be filed separately: registration silently overwriting `agent_runtime.owner_id` on re-login (`server/pkg/db/queries/runtime.sql:76` + `server/internal/handler/daemon.go:430`) is what creates this state in the first place; there is a second route through admin agent rebind (`server/internal/handler/agent.go:1822` compares the *actor* against the runtime owner, not the agent's owner); and there is still no CLI for agent ownership transfer or runtime visibility, so the error message we write has no command to point at.

Good find on the root cause — the fence analysis in the description is accurate and the direction is right.

makesomethingshit:
@Bohan-J Addressed the requested changes in `d240b8c5907ecf9cbd88edc1ffe316fc82f2b667`:

- reject private-runtime owner mismatches in the existing `AgentReadiness` admission path with `runtime_access_denied`, including durable assignment and mention notices;
- remove the claim-time sweep and let pre-existing mismatched rows enter the existing `failClaimedTaskBeforeLaunch -> FailTask` settlement path;
- return `200 {"task":null}` after a singular settled mismatch, while batch claims continue to a valid task;
- preserve the dispatched stale-reclaim owner fence and the public, same-owner, and ownerless-runtime behavior;
- restore the unrelated `server/internal/service/task.go` empty-claim sampling code to `main` (no diff remains in that file).

Focused admission, claim, issue, mention, and chat regression tests pass. The latest CI run is green. The PR description now reflects the final architecture and intentional queued/dispatched asymmetry. Please re-review when convenient.

multica-eve (changes requested):
Previously, ownership drift could leave work stuck forever with no explanation. This revision now refuses new incompatible work immediately, turns existing stuck work into an explicit failure, and lets valid work continue. That is the right user outcome, but several user entry points still do not explain how to recover, and one timing window can send work using an outdated private-machine owner decision.

## Product verdict: accept the direction

I independently verified that the problem is real, it is worth a durable fix, and the intended shape is the one we would choose: reject known-invalid work at shared admission, then settle pre-existing rows through the existing terminal failure path. The two blockers from the previous review—claim-hot-path sweeping and bypassing `TaskService.FailTask`—are fixed on this head.

## Must fix before merge

1. **Close the runtime-owner TOCTOU window at the final delivery gate.**

   The singular path reads the runtime before claiming (`server/internal/handler/daemon.go:3266`), `ClaimAgentTask` no longer checks the current owner (`server/pkg/db/queries/agent.sql:796-803`), and the response builder receives that pre-claim runtime snapshot (`server/internal/handler/daemon.go:3305`, checked at `:2086-2103`). The batch path has the same shape: runtimes are loaded at `:1655-1675` and the cached value is passed at `:1706`. Meanwhile `UpsertAgentRuntime` can update `owner_id` concurrently (`server/pkg/db/queries/runtime.sql:69-76`).

   If the owner changes between the initial read and claim/build, a task can pass against the old owner and be delivered. Keep mismatched queued rows claimable for settlement, but re-read the runtime after claim at the terminal gate. Preferably, make the current agent/runtime authorization snapshot and final delivery/token finalization share the same transaction or locked snapshot. Add singular and batch regressions that change the runtime owner after the initial access read and assert that the task is settled and never returned.

2. **Handle `runtime_access_denied` in the current clients with actionable recovery copy.**

   An open string keeps the wire backward-compatible, but it does not complete the current product behavior. The new reason currently falls through to generic handling in:

   - `packages/views/issues/blocked-trigger-copy.ts:29-61`
   - `packages/views/chat/components/use-chat-controller.ts:545-586` and the matching chat-window path
   - `apps/mobile/lib/dispatch-reason.ts:25-33`
   - `packages/views/autopilots/components/run-now-toast.ts:32-64`

   Mobile currently tells the user to retry, but retrying cannot repair a permanent owner mismatch. Keep generic fallback for older clients, and teach current web/desktop/mobile surfaces to recognize the reason and show the same public-runtime or rebind/copy recovery actions as the durable server notice. Please add localized mapping tests.

3. **Use the repository's required DB/handler test helpers.**

   `server/internal/handler/runtime_access_denied_test.go:75-76,105-110` manually builds the recorder flow instead of using `testutil.Call`. `server/internal/handler/daemon_runtime_access_test.go:89-97` manually inserts and cleans up a task instead of using a `dbfx.Task` fixture. These are hard repository conventions; please convert them in this PR.

## Validation

On `d240b8c5907e`, a synthetic merge with current `origin/main` completed without conflicts. `sqlc generate` produced no diff; `git diff --check`, `go vet ./...`, the focused service tests, and the `ByteIdentical` guard passed. The focused DB tests could not reach these assertions locally because the local schema lacks `durable_work_dir` / `conversation_starters`; the current GitHub backend CI is green.

If this change is wrong, the authorization race will surface as a private machine receiving work or credentials based on a former owner, while the client gap will surface as generic failures and futile retries; reverting stops new behavior but does not restore tasks already terminally settled, and the affected population is private-runtime owner changes plus current clients receiving the new reason.

**Verdict: Request changes.** The product direction is accepted and the previous blockers are resolved; the authorization race, current-client recovery path, and required test patterns still block merge.

makesomethingshit:
Test-only follow-up on `bed4c9642` (reviewer's earlier hard-convention note about test patterns).

- `server/internal/handler/daemon_runtime_access_test.go`: `TestFinalizeClaimDelivery_SettlementFailureIsUnsettled` still built its own `httptest.NewRecorder` flow. Converted it to the required `testutil.Call(...).Want(http.StatusInternalServerError)` helper and dropped the now-unused `net/http/httptest` import. No production code changes.

Validation on `809b750c4ef2d333d02a8197488c1b3edfe6f35f`:
- `go test ./internal/handler/ -run 'TestFinalizeClaimDelivery|TestClaimTaskByRuntime_Owner|TestClaimTaskByRuntime_SettlesPrivate|TestClaimTaskByRuntime_RuntimeOwnerChanged|TestClaimTasksByRuntime_OwnerChanged' -count=5` — pass
- `go vet ./...`, `gofmt` (touched file clean), `git diff --check` — pass
- unrelated comment-delivery batch tests that fail on this machine fail identically on the unmodified head (pre-existing local-env flake, not touched by this commit)

Head will stay pinned at `809b750c4` until re-review.

multica-eve (changes requested):
Before this change, an ownership mismatch could leave work waiting indefinitely with no explanation. This revision now refuses new incompatible work, settles already-waiting work, and lets valid work continue, which is the right user outcome; however, users whose queued work is settled still do not reliably see the new recovery guidance, so the product loop is not complete yet.

## Product verdict: accept the direction

The problem is real, it warrants a durable mechanism, and the design is now the one we would choose: shared admission for new work, deterministic settlement for existing work, and a final authorization check against current ownership before delivery. The previous authorization-race and test-convention blockers are resolved on this head.

## Must fix before merge

1. **Persist the same dedicated reason that the clients use for queued owner-mismatch failures.**

   The owner-mismatch settlement paths still call `failClaimedTaskBeforeLaunch` with `taskfailure.ReasonInvalidTaskIdentity` at `server/internal/handler/daemon.go:1905-1910` and `server/internal/handler/daemon.go:2233-2238`. That reason is explicitly defined for a missing/disagreeing task agent or payload at `server/pkg/taskfailure/failure.go:172-177`, not for runtime ownership authorization.

   The newly added web chat copy listens for `runtime_access_denied` at `packages/views/chat/components/chat-message-list.tsx:976-1005`, so it is unreachable for these settled tasks. Web issue history instead shows “Task identity mismatch”; mobile chat and run rows do not map either value to the actionable recovery and fall back to a generic `Failed`. The current backend tests reinforce the mismatch by asserting `invalid_task_identity` for owner-denied tasks.

   Please add a canonical task failure reason for runtime access denial (the same `runtime_access_denied` wire value is appropriate), use it only for owner-mismatch/owner-missing settlement, and retain `invalid_task_identity` for missing/rebound/mismatched agent identity. Add the corresponding web locale/operator mapping, mobile failure and run-badge mappings, and an end-to-end regression from a queued chat/issue task through claim settlement to the user-visible reason. Admission `reason_code` and persisted `failure_reason` should describe the same condition.

## Non-blocking note

The earlier checks at `server/internal/handler/daemon.go:1714-1721`, `:2221-2238`, and `:3471-3484` can still fail closed from an older runtime snapshot if ownership is repaired immediately before the transactional final gate. This is pre-existing behavior and does not create unauthorized delivery, so it does not block this PR, but the final locked gate should eventually become the sole authoritative ownership decision or the reverse race should be documented and tested.

## Validation

Reviewed head `b6a2a35316f900a8c2ca2615b6e859bc07c43bd0`; the PR is mergeable/clean and all current required GitHub checks pass. Locally, `git diff --check`, `sqlc generate` with no generated drift, `go vet ./...`, and the focused pure service tests passed. DB-backed tests could not reach the assertions because the local test schema is stale, and frontend/mobile dependencies are not installed locally; current backend/frontend/mobile CI covers those lanes successfully.

If this assessment is wrong, the first signal will be queued owner-mismatch work showing a generic failure or “Task identity mismatch”; comparing each failed task's `error` and `failure_reason` identifies it, reverting stops new bad records but already-settled tasks must be retried, and affected users are those with queued chat, issue, or autopilot work during a private-runtime ownership mismatch.

**Verdict: Request changes.** The remaining blocker is the persisted-reason/client-copy disconnect; once it and its end-to-end visibility coverage are fixed, this is ready for another review.

makesomethingshit:
## Verification summary (latest review blocker resolved)

Persisted task failure semantics are now aligned with admission semantics: queued private-runtime ownership mismatches settle as `runtime_access_denied`, allowing current Web/Desktop/Mobile clients to show the same actionable recovery guidance instead of "Task identity mismatch" or generic "Failed". Actual task/agent identity violations continue to use `invalid_task_identity`.

### Backend
- `server/pkg/taskfailure/failure.go`: new canonical reason `ReasonRuntimeAccessDenied` (`runtime_access_denied`) — private runtime/agent ownership authorization failure, agent process never launched, permanent/non-retryable under the existing retry policy; registered in the platform reasons group.
- `server/internal/handler/daemon.go`: only the two ownership-denial settlement paths changed — response-assembly re-check (owner mismatch / ownerless agent) and the delivery-gate authz default branch (owner mismatch, ownerless agent, missing current runtime owner). The four real identity paths (agent deleted, agent rebound to another runtime, response identity mismatch, `error_agent_runtime_changed`) still persist `invalid_task_identity`.
- Reuses the existing `FailTask` settlement path and retry allowlist — no parallel settlement path; queued chat tasks still produce the assistant failure message.

### Clients
- Web/Desktop: persisted `runtime_access_denied` mapped via the shared task/chat failure mapping in `@multica/views`, reusing the existing admission recovery copy semantics (make runtime public / rebind or copy the agent); retry is not presented as the primary remedy.
- Mobile: single shared helper maps the persisted reason to dispatch reason, persisted chat failure label, and run badge.

### Regression tests
- A: pre-existing queued issue task with private runtime owner mismatch → `status=failed`, `failure_reason=runtime_access_denied`
- B: pre-existing queued chat task → same reason + assistant failure message via existing `FailTask`
- C: agent rebind (real identity mismatch) → still `invalid_task_identity` (reason boundary pinned)

### Checks
- `go vet ./...` PASS · `go build ./...` PASS · `go test ./pkg/taskfailure/` PASS · focused handler suite 10/10 PASS (real DB)
- Web focused 44/44 · locale parity 26/26 · Mobile focused 15/15 · views/mobile/Desktop typecheck PASS · lint exit 0
- `git diff --check` PASS · no SQL changes (no `sqlc generate` impact)

Maintainer verdict from our internal review: READY FOR MAINTAINER REVIEW — the only blocker from the latest review (persisted reason ↔ client recovery copy disconnect) is closed; the reverse ownership-repair race remains non-blocking and out of scope.

makesomethingshit:
@multica-eve When you have a chance, could you please re-review PR #7805? Thank you!

PR #7813, opened 2026-08-31
MUL-6880 fix(server): hold the execution slot until a cancelled run confirms it stopped

Bohan-J:
## The bug

Editing a comment cancels the run it triggered and enqueues the replacement in the same request. `ClaimAgentTask` already serializes per `(issue, agent)` — a task is claimable only while no sibling row is `dispatched`/`running`/`waiting_local_directory` — but that fence reads DB status, and cancellation flips the row to `cancelled` **immediately**. The agent process keeps running for another 4–10s, until the daemon's own cancel poll notices.

The replacement was claimable inside that window. It could not take the env root lock the dying run still held, so it started a second workdir, and with the workdir went the provider session: the agent came back with `provider session could not be restored` and no memory of the conversation the user was editing.

From a real occurrence (MUL-6856, `~/.multica/profiles/.../daemon.log`):

```
14:31:30.523  starting agent      workdir=...mul-6856-35ad26c54c00/workdir  reused=true
14:31:35.716  prior workdir is in use by another execution; starting a fresh environment
14:31:36.131  dropping prior session: session store not reachable from this run
14:31:36.147  claude started      cwd=...mul-6856-86e5065e9e12/workdir      ← second workdir
14:31:40.325  task reached terminal state server-side, interrupting agent
14:31:40.850  claude finished     status=aborted duration=10.311s           ← 4.6s too late
```

Cost per occurrence: session continuity (user-visible), one wasted agent startup, one extra workdir with its repo checkout.

## The change

A cancelled run that was **actually executing** now keeps owning its `(issue, agent)` execution slot until the daemon proves the process is gone.

- **`agent_task_queue.stop_lease_expires_at`** (migration 443), armed by `trg_arm_task_stop_lease` when a `dispatched`/`running`/`waiting_local_directory` row is cancelled. A `queued` or `deferred` row was never handed to a daemon, so it arms nothing and delays nothing.
- **`ClaimAgentTask`** treats a live lease as an occupied slot, in the same `NOT EXISTS` that already enforces per-`(issue, agent)` serialization.
- **Released** by the daemon's cancel-ack — posted after `runTask` returns, which is the same moment the env root lock is released — followed by a runtime wakeup so the successor is claimed at once rather than on the next poll. Runtime orphan recovery also clears leases (a restarted daemon cannot still be shutting anything down), and a 30s deadline backstops a daemon that never acks at all.

Arming is a trigger rather than a `SET` clause in each cancel query, for the same reason as `trg_clear_runtime_mcp_overlay` (migration 128): the transition lives in a dozen sqlc queries and a future one must not be able to skip it.

The `OR` arm is spelled `active.status = 'cancelled' AND active.stop_lease_expires_at > now()` even though only a cancelled row can carry a lease — every arm then matches on status, so the subquery stays inside `idx_agent_task_queue_agent (agent_id, status)` instead of scanning an agent's whole history.

The chat half of this was GH #6340, which deliberately keeps a cancelled turn's session resumable. This makes that promise hold in practice instead of only in the row.

## Rollout

Additive column plus trigger, no index, no backfill. During a rolling deploy old server instances simply ignore leases the trigger arms and let them expire — same behavior as today. A daemon too old to post a cancel-ack falls through to the 30s deadline instead of stalling.

## Verification

`DATABASE_URL=<scratch db> go test ./internal/handler/ ./internal/service/ ./internal/migrations/ -count=1` — all pass. Full `scripts/test-go.sh` shows the same 4 pre-existing failures with and without this branch (`cmd/multica`, `internal/cli`, `internal/daemon`, `internal/daemon/execenv` — they fail because the suite runs inside an agent execution context). Migration up/down round-tripped on the scratch DB.

New tests:

- `TestUpdateComment_EditedCommentWaitsForTheCancelledRunToStop` — the reported flow end to end: edit → old run cancelled with a live lease → replacement queued but not claimable → cancel-ack → replacement claims and resumes the cancelled run's `prior_work_dir` + `prior_session_id`.
- `TestStopLeaseHoldsTheExecutionSlotUntilTheDaemonConfirms` — which cancellations arm a lease, what a live lease blocks (and what it must not: another issue), release by ack, release by runtime recovery, and the expiry backstop.
- `TestCancelTask_PointerAdvanceIsAtomicWithStatusFlip` was asserting the follow-up could be claimed the instant the cancel committed — that is the timing this PR removes. It now asserts the follow-up is held until the cancel-ack and then resumes the cancelled turn's session.

Both new tests were confirmed to fail without the fix (trigger dropped → no lease; fence arm neutered → the successor claims during the stop window).

## Not in this PR

Two side findings from the same investigation, recorded in MUL-6880 and left open: `shortID` renders both tasks of this race as `task=01a05684` (the daemon log timeline had to be reconstructed from pids), and `lockReusablePriorEnvRoot` leaves an empty env root behind on every successful reuse (42 real environments vs 97 shells on one machine).

---

## Review round 1

`AckTaskCancelled` released the stop lease **before** `FinalizeDeferredCancelledChat`. A direct-chat follow-up claimed in that gap flips to `dispatched`, and `ReanchorNextQueuedDirectChatInput` only reaches a follow-up that is still `queued` — so the settle's `Stopped.` would land after the user's next message and the transcript would read user B before assistant A, the exact inversion that query's contract rules out.

- The release now runs **last**, after every write the ack owes the cancelled run.
- A failed settle **fails closed**: the lease is kept (the sweeper still retries, the deadline still frees the successor) and the ack returns 500 so the daemon retries it.
- `FinalizeDeferredCancelledChat` returned one `false` for both "nothing was deferred" and "the write failed", which cannot express that gate. It now returns `(changed, err)`.
- New concurrency regression test `TestAckTaskCancelled_ReleasesTheSlotOnlyAfterTheChatSettleCommits`: pins the settle on the `chat_session` row lock, fires the cancel-ack asynchronously, asserts the follow-up is unclaimable while blocked, then — after release — asserts it is claimed and that assistant A is ordered before user B. Confirmed to fail with the old ordering.

## Review round 2

The stop lease expires 30s after the cancel, but a cancel-ack that never arrives leaves the deferred chat settle to the sweeper — eligible only after its 60s grace, then on a 30s tick, so it can commit 60–90s after the cancel. In that gap the follow-up starts and the settle can no longer re-anchor it: the same inversion by a slower route.

The lease means "the process may still be running"; it cannot also mean "the outcome is still unwritten", because the two clear at different times. So the unsettled marker is now its own claim barrier — a cancelled chat turn holds its session's slot while `chat_finalize_deferred_at` is set, bounded at 5 minutes so a marker that can never settle cannot wedge the session. The ack path clears the marker before releasing the lease, so this costs nothing there. `FinalizeDeferredCancelledChat` also wakes the runtime after a settle that changed something, since on the sweeper's path nothing else would.

New test `TestSweeperSettle_HoldsTheFollowUpWhenTheCancelAckNeverArrives`: lease expired with the marker still armed → follow-up unclaimable; sweeper settle → claimed, assistant A before user B. Confirmed to fail with the marker arm removed.

Also in this push: merged `main` and renumbered the migration to **444** — `main` took 443 for `issue_project_status_index` while this branch was open, and the prefix-uniqueness lint rejects a collision.

The `backend-tests` failure on the previous run was `TestRelay_ADeliveredReplyIsCountedOnceAndNotAlsoLost` (`internal/integrations/wecom`, added by #7429 shortly before that run). It is timing-sensitive — 120ms lease-settle, a 600ms grace sleep — and unrelated to this branch, which touches no wecom code. 10/10 green locally on the merged tree with `REDIS_TEST_URL` set.


Bohan-J:
Closing in favour of #7818, which fixes the same bug on the daemon side.

This PR made the server hold the `(issue, agent)` claim open until the daemon confirmed the stop: a `stop_lease_expires_at` column, a trigger arming it on cancel, release on cancel-ack, an expiry backstop, a second barrier for the unsettled chat marker, and a sweeper wake. Three review rounds each found a real problem with it — the settle had to commit before the lease was released, the lost-ack path reopened the window between lease expiry and the sweeper's 60–90s fallback, and the new claim predicate turned the hot claim query into a full table scan (9.6ms vs 0.015ms with 100k historical rows for one agent).

None of those questioned *waiting for the predecessor to let go*. All three were the cost of re-deriving a machine-local fact in Postgres and gating the system's hottest query on it. The fact — "the previous process still holds this directory" — is available first-hand in the daemon, as an `ErrEnvRootBusy`, so #7818 waits for the lock there instead: 2 files, no migration, no trigger, no index, and `ClaimAgentTask` untouched.

The reviews were not wasted — they are what established the full timing picture and measured the real cost of the shadow state.

One thing this branch covered that #7818 deliberately does not: a cancelled direct-chat turn whose deferred settle lands after its follow-up was claimed leaves the transcript reading user B before assistant A. That inversion exists on `main` today and is independent of this bug; it was only in scope here because the fence lived in the claim path. Tracked separately for a fix on the reanchor/settle path.

PR #7820, opened 2026-08-31
MUL-6883 perf(server): stop rescanning settled delegated-failure recoveries

Bohan-J:
## Summary

`ListPendingDelegatedFailureRecoveries` is the durable outbox scan that repairs a delegated-failure recovery whose dispatch was lost to a transient error or a process restart. It had no indexable notion of a *finished* recovery: the partial index from migration 343 selected every recovery comment ever written, so each sweeper tick re-proved through four joins and two `NOT EXISTS` subqueries that history was already handled.

`LIMIT` bounds the rows **returned**, never the rows **checked** — so a tick that finds nothing still gets more expensive as history accumulates. #7814 made each execution cheaper; this makes each execution examine less.

## What changed

**`comment.recovery_settled_at` (migration 444) + a partial index over only unsettled recovery comments (migration 445).** The scan gains one predicate, `recovery_settled_at IS NULL`, matching the new index. Everything else it checks is unchanged.

**The marker is written inside the transaction that makes a task terminal.** Every other exclusion the scan applies is reversible — an issue can leave `done`, a cancelled retry can be superseded — so none of them can be frozen into an index predicate. A delivery receipt on a terminal task cannot be taken back, which is what makes that one condition safe to record as durable state.

**Only once the task is terminal.** `SetTaskDeliveredCommentIDs` *replaces* rather than appends a receipt while a task is still `dispatched`, precisely so a reclaim by a differently-capable daemon can drop an id it will not deliver. Settling at receipt-write time would freeze that legitimate transient window into a permanently lost recovery. Exhaustion is the one terminal outcome whose receipt lands on a row that may still be running, so it retires the comment directly in the same transaction.

**A backfill command for history written before the marker existed**, in the shape `backfill_issue_last_activity` already uses. Its keyset watermark advances over every candidate *examined* rather than every candidate marked — a recovery that is genuinely still pending must be left alone, and keying on updated rows would hand back the same watermark forever and never terminate.

## Why an unmarked row is always safe

A `NULL` means "not proven finished", so an unmarked row is scanned exactly as it was before this column existed. That is what makes the rolling-deploy window harmless: an older replica that writes a receipt without the marker only leaves work for the backfill, and no recovery can be lost. The failure direction is missed marks, never false ones.

`TestPlannedButUndeliveredRecoveryStaysPending` is the guard on the other side — a task that only *planned* the comment and was cancelled automatically must leave the obligation open, and the sweeper must still replay it.

## Deploy order

1. Merge and migrate (444 adds a nullable column: metadata-only; 445 builds `CONCURRENTLY`).
2. Let the rolling deploy finish so every replica writes the marker.
3. Run `go run ./cmd/backfill_delegated_failure_settled` off-peak. It is resumable and bounded — `--batch-size`, `--sleep-between-batches`, `--max-batches`, and SIGINT/SIGTERM all stop it cleanly. This is a performance recovery step, not a correctness step.
4. Watch `idx_comment_delegated_failure_unsettled`'s row count fall to the genuinely open backlog, then drop the now-redundant index from 343 in a follow-up.

Index 343 is intentionally left in place through the rollback window.

## Follow-up (not in this PR)

The scan still runs every 30 seconds on every backend replica, inside the runtime-liveness loop. Moving it to its own ~5-minute worker with a Redis single-instance lease, plus one run at startup, is the second half of MUL-6883 and lands after this backfill is verified in production — sequenced so a cadence change cannot mask an index problem.

## Testing

```
go test ./internal/service/ ./internal/delegatedrecoverybackfill/ ./cmd/migrate/ ./internal/handler/   # pass
go test ./...                                                                                          # see note
```

New coverage:

- `TestTerminalCoordinatorTaskSettlesDeliveredRecovery` — complete / fail / cancel each settle a delivered recovery, and the scan then skips it
- `TestPlannedButUndeliveredRecoveryStaysPending` — an automatic cancellation that never delivered leaves it pending, and the sweeper replays it
- `TestUserCancelledRecoveryIsSettled`, `TestExhaustedRecoveryIsSettled`
- `TestBatchSettlesOnlyTerminallyDeliveredRecoveries`, `TestBatchWatermarkAdvancesPastUnsettleableCandidates`

`cmd/multica`, `internal/cli`, `internal/daemon` and `internal/daemon/execenv` fail in my sandbox because it runs inside an agent task context (`daemon_task_context.json` present, task-scoped `MULTICA_TOKEN`). They fail identically on unmodified `main`, so they are environmental, not caused by this change.

### DIFF
--- a/server/cmd/backfill_delegated_failure_settled/main.go
+++ b/server/cmd/backfill_delegated_failure_settled/main.go
@@ -1,0 +1,134 @@
+// backfill_delegated_failure_settled retires delegated-failure recovery
+// comments that were already closed by a terminal delivery receipt before
+// comment.recovery_settled_at existed.
+//
+// Run it once, after every backend in a rolling deployment carries the code
+// that writes the marker. Until it finishes, the outbox scan behaves exactly as
+// it did before — unmarked history is still scanned and no recovery is lost —
+// so this is a performance recovery step, not a correctness step, and it is
+// safe to run off-peak or in several sittings.
+//
+// Each batch commits independently and the walk resumes from an id keyset
+// watermark, so SIGINT/SIGTERM or --max-batches can stop it at any point. The
+// watermark advances over every candidate examined, not only the ones marked:
+// a recovery that is genuinely still pending must be left alone and must not
+// stall the walk. A session advisory lock keeps two operators from walking at
+// once.
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/multica-ai/multica/server/internal/delegatedrecoverybackfill"
+	"github.com/multica-ai/multica/server/internal/logger"
+)
+
+const advisoryLockName = "delegated_failure_recovery_settled_backfill"
+
+func main() {
+	logger.Init()
+	if err := run(); err != nil {
+		slog.Error("delegated failure recovery settled backfill failed", "error", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	batchSize := flag.Int("batch-size", delegatedrecoverybackfill.DefaultBatchSize, "maximum candidate recovery comments examined per statement")
+	delay := flag.Duration("sleep-between-batches", 100*time.Millisecond, "delay between committed batches")
+	maxBatches := flag.Int("max-batches", 0, "stop after N batches (0 = walk the whole history)")
+	flag.Parse()
+	if *batchSize < 1 {
+		return fmt.Errorf("--batch-size must be at least 1")
+	}
+	if *delay < 0 {
+		return fmt.Errorf("--sleep-between-batches must not be negative")
+	}
+	if *maxBatches < 0 {
+		return fmt.Errorf("--max-batches must not be negative")
+	}
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+	dbURL := os.Getenv("DATABASE_URL")
+	if dbURL == "" {
+		dbURL = "postgres://multica:multica@localhost:5432/multica?sslmode=disable"
+	}
+	pool, err := pgxpool.New(ctx, dbURL)
+	if err != nil {
+		return fmt.Errorf("connect to database: %w", err)
+	}
+	defer pool.Close()
+	if err := pool.Ping(ctx); err != nil {
+		return fmt.Errorf("ping database: %w", err)
+	}
+
+	lockConn, err := pool.Acquire(ctx)
+	if err != nil {
+		return fmt.Errorf("acquire advisory-lock connection: %w", err)
+	}
+	defer lockConn.Release()
+	if _, err := lockConn.Exec(ctx, `SELECT pg_advisory_lock(hashtextextended($1, 0))`, advisoryLockName); err != nil {
+		return fmt.Errorf("acquire advisory lock: %w", err)
+	}
+	defer func() {
+		_, _ = lockConn.Exec(context.Background(), `SELECT pg_advisory_unlock(hashtextextended($1, 0))`, advisoryLockName)
+	}()
+
+	remaining, err := delegatedrecoverybackfill.CountRemaining(ctx, pool, "")
+	if err != nil {
+		return err
+	}
+	slog.Info("delegated failure recovery settled backfill started",
+		"unsettled", remaining, "batch_size", *batchSize, "delay", delay.String())
+
+	var scanned, settled int64
+	var afterID *string
+	for batch := 1; *maxBatches == 0 || batch <= *maxBatches; batch++ {
+		result, err := delegatedrecoverybackfill.Batch(ctx, pool, delegatedrecoverybackfill.Options{
+			BatchSize: *batchSize,
+			AfterID:   afterID,
+		})
+		if err != nil {
+			return err
+		}
+		if result.Scanned == 0 {
+			break
+		}
+		if result.LastID == "" {
+			return fmt.Errorf("backfill batch examined %d candidates without a keyset watermark", result.Scanned)
+		}
+		scanned += result.Scanned
+		settled += result.Settled
+		slog.Info("delegated failure recovery settled batch committed",
+			"batch", batch, "scanned", result.Scanned, "settled", result.Settled,
+			"scanned_total", scanned, "settled_total", settled, "last_id", result.LastID)
+		next := result.LastID
+		afterID = &next
+		if *delay > 0 {
+			select {
+			case <-time.After(*delay):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+	}
+
+	remaining, err = delegatedrecoverybackfill.CountRemaining(ctx, pool, "")
+	if err != nil {
+		return err
+	}
+	// The remainder is the genuinely open backlog plus anything a reversible
+	// exclusion is holding, which is what the index is supposed to contain.
+	slog.Info("delegated failure recovery settled backfill finished",
+		"scanned", scanned, "settled", settled, "unsettled_remaining", remaining)
+	return nil
+}
--- a/server/cmd/backfill_delegated_failure_settled/main.go
+++ b/server/cmd/backfill_delegated_failure_settled/main.go
@@ -2,27 +2,44 @@
 // comments that were already closed by a terminal delivery receipt before
 // comment.recovery_settled_at existed.
 //
-// Run it once, after every backend in a rolling deployment carries the code
-// that writes the marker. Until it finishes, the outbox scan behaves exactly as
-// it did before — unmarked history is still scanned and no recovery is lost —
-// so this is a performance recovery step, not a correctness step, and it is
-// safe to run off-peak or in several sittings.
+// Run it after every backend in a rolling deployment carries the code that
+// writes the marker. Until it finishes, the outbox scan behaves exactly as it
+// did before — unmarked history is still scanned and no recovery is lost — so
+// this is a performance recovery step, not a correctness step, and it is safe
+// to run off-peak or in several sittings.
 //
-// Each batch commits independently and the walk resumes from an id keyset
-// watermark, so SIGINT/SIGTERM or --max-batches can stop it at any point. The
-// watermark advances over every candidate examined, not only the ones marked:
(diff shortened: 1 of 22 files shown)


Bohan-J:
Pushed `0d2859a0b` addressing all three must-fix items.

## 1. The backfill reproduced the bug it was written to fix

You were right on both halves. The keyset ordered by `id` while the index is `(created_at, id)`, and the receipt test ran a per-candidate `EXISTS` over `agent_task_queue`.

Fixed: the keyset is now `(created_at, id)`, and the receipt lookup is inverted — one `agent_task_queue` read per **batch**, hash-joined back against the candidates, with both CTEs `MATERIALIZED` so the plan can't be re-inlined into the per-row shape.

`EXPLAIN (ANALYZE, BUFFERS)` on a synthetic 500k comments / 200k tasks / 50k unsettled recoveries, batch size 1000:

| | before | after |
|---|---|---|
| Execution time | **17,393 ms** | **235 ms** |
| Candidate scan | `Index Only Scan … rows=50000` → `Sort` → `Limit` | `Index Only Scan … rows=1000` |
| `agent_task_queue` | `Seq Scan rows=200000` re-read via **`temp read=816,963`** | `Seq Scan … loops=1` |
| temp blocks | 816,963 | 0 |

The `temp read=816963` is the per-candidate re-reading in disguise — it spilled the materialized task set to disk and re-read it for every candidate.

On the defaults: total walk time turns out to be **flat** across batch size — 50×228ms at 1000, 10×1016ms at 5000, 5×2199ms at 10000, all ≈11s — so the per-batch task read is not the dominant cost. Batch size therefore only trades how long one statement holds row locks on `comment`, which makes 1000 the smallest of the equal-throughput options. Kept, now for a measured reason. The rationale is in the `DefaultBatchSize` doc comment so the next person doesn't re-guess it.

## 2. Missed settlement paths

Both of yours are fixed, and auditing the rest of the class found four more in the same position:

| path | fix |
|---|---|
| `RerunIssue` pending-slot clear | settles before the capture loop |
| agent archive | settles via `CaptureCancelledTasks` |
| comment-edit cancellation | settles in the caller's `qtx` |
| chat session archive + delete | settles in the caller's `qtx` |
| workspace revoke | settles in the caller's `qtx` |
| runtime teardown | settles in the caller's `qtx` |

In-transaction where a transaction exists, so those are atomic rather than best-effort. The invariant is now stated on the exported helper, since "every terminal path must reach this" was the thing that wasn't written down anywhere.

Regression tests: `TestManualRerunSettlesDeliveredRecovery` and `TestCaptureCancelledTasksSettlesDeliveredRecovery`. Both verified to fail without the fix.

## 3. `--max-batches` could not resume

Correct, and the failure mode is exactly the one you describe: the front of the index is where the rows this walk *never* settles live, so a bounded rerun starting from zero rescans that prefix forever.

Added `--checkpoint-file` (read on start, atomically replaced after every committed batch) and explicit `--after-created-at` / `--after-id` for operators who would rather not keep a file. The resume cursor is now logged on **every** exit path, including SIGTERM and `--max-batches`, so a run without a checkpoint file is still resumable from the logs. Covered by `TestCheckpointRoundTripsTheWatermark`, `TestCheckpointWriteLeavesNoTempFilesBehind`, and `TestResolveStartPrefersExplicitFlagsOverTheCheckpointFile`.

## Also

- Two more backfill tests: `TestBatchSettlesNothingOutsideTheCurrentBatch` (the per-batch task read sees ids outside the batch — they must be discarded, not settled behind the watermark) and `TestBatchOrdersByCreatedAtNotID`.
- Agreed on both notes: the advisory lock here is operator mutual exclusion for the manual command only, and MUL-6883 stays open after this merges. The startup-scan + 5-minute + Redis-lease worker is #7821.

## Testing

`internal/service`, `internal/handler`, `internal/delegatedrecoverybackfill`, `cmd/backfill_delegated_failure_settled`, `cmd/migrate`, `cmd/server` all pass locally.

One caveat, stated because you flagged local verification: under a full `go test ./...` against a *shared* dev database, `TestCleanupSourceContextObjectIntentsBoundsAttemptsNotSuccesses` (unrelated to this PR) flakes on concurrent load. It passes in isolation on both this branch and unmodified `main`. `cmd/multica`, `internal/cli` and `internal/daemon*` fail in my sandbox because it runs inside an agent task context; also identical on `main`.

Bohan-J:
Pushed `9ae7848c4`. Both must-fixes were correct — the second in particular caught a claim in my own comment that was simply false.

## 1. The settlement SQL now enforces the terminal requirement

"A delivery receipt is only immutable once the task is terminal" is the safety premise of the entire design, and the exported helper enforced none of it. Passing a `dispatched`/`running` task would settle a receipt a reclaiming daemon can still replace — and the marker is monotonic, so nothing downstream could undo it.

The predicate now lives in the SQL:

```sql
WHERE task.id = @task_id
  AND task.status IN ('completed', 'failed', 'cancelled')
  AND recovery.id = ANY(task.delivered_comment_ids)
```

A non-terminal task updates zero rows instead of silently destroying the obligation. `TestNonTerminalTaskWithAReceiptIsNotSettled` covers `dispatched` and `running`; verified to fail on both without the predicate.

## 2. Atomicity — and my "the sweeper replays it" comment was wrong

You're right, and the mechanism is worth stating plainly because it's what makes this unrecoverable rather than merely delayed: the covering task is terminal and already holds the receipt, so `ListPendingDelegatedFailureRecoveries` **excludes** the comment. A failed best-effort settle would neither replay nor settle. It would sit in the partial index forever — exactly the unbounded scan this PR exists to remove.

Every terminal write now commits with its settlement:

| path | how |
|---|---|
| `CancelTasksForIssue` / `ForAgent` / `ByTriggerComment` | wrapped in a transaction |
| rerun pending-slot clear | wrapped in a transaction |
| `CancelQueuedChatTasks` | settles inside the transaction it already had |
| sweeper bulk fails (offline runtimes, reconnect retries, stale, queued-expired) | moved behind `TaskService` wrappers that own the transaction |
| daemon orphan recovery | same wrapper |
| agent archive | new `CancelTasksForArchivedAgent` — cancels and settles atomically, still skipping the per-task broadcast the `agent:archived` event makes redundant |

And the three post-commit helpers — `BroadcastCancelledTasks`, `CaptureCancelledTasks`, `HandleFailedTasks` — **no longer settle at all**. They run after their caller committed, so a failure there could be neither rolled back nor repaired. That was the actual bug, not an oversight at two call sites.

The best-effort helper is **deleted** rather than left available, so the unsafe form cannot come back. Every remaining call site passes a `qtx`:

```
$ grep -rn 'SettleDeliveredDelegatedFailureRecoveries(' --include='*.go' server/ | grep -v _test
… 14 call sites, all with qtx
```

`TestBulkCancelRollsBackWhenSettlementFails` pins it: a trigger rejects the settlement UPDATE, and the test asserts the cancel rolled back with it rather than committing and stranding the receipt.

## A note on how this was implemented

I first tried putting the settlement into the bulk statements as a CTE — atomic by construction and impossible to forget. It changed sqlc's inferred return types from `[]AgentTaskQueue` to five distinct `…Row` types, which would have meant ~300 lines of hand-written conversion boilerplate. I reverted it and used service-level transaction wrappers instead, which keeps one mechanism across the PR and leaves the SQL untouched. Flagging it in case you'd prefer the SQL form despite that cost.

## Testing

`internal/service`, `internal/handler`, `cmd/server`, `internal/delegatedrecoverybackfill`, `cmd/backfill_delegated_failure_settled`, `cmd/migrate` all pass.

Same four sandbox-environmental packages fail as before (`cmd/multica`, `internal/cli`, `internal/daemon*` — agent task context), identical on unmodified `main`.

On your local blocker: the missing `durable_work_dir` column is a stale reused test database, not this branch. A fresh database plus `go run ./cmd/migrate up` runs the service tests clean here.

multica-eve:
Picked up the two non-blocking nits from the last review; pushed as `f460dd08a`. No behavior change, no SQL change.

## 1. Stale invariant comment on `SettleDeliveredDelegatedFailureRecoveries`

The doc still described the contract this PR replaced — "bulk cancellations call it with their own qtx, **or through `CaptureCancelledTasks` when they finalize outside one**", plus "callers that already finalize outside a transaction pass `s.Queries` and log instead". That helper no longer exists and `CaptureCancelledTasks` explicitly refuses to settle, so the comment was the only remaining instruction pointing back at the bug.

Rewritten to state the rule as it now is, and to say *why* rather than just what:

- settlement must run with the **same `qtx`** as the terminal write, per-task and bulk alike;
- a committed-but-unsettled terminal write is **unrepairable**, because `ListPendingDelegatedFailureRecoveries` excludes a comment whose covering task is already terminal and holds its receipt — nothing replays it, nothing else marks it;
- therefore the post-commit helpers (`BroadcastCancelledTasks`, `CaptureCancelledTasks`, `HandleFailedTasks`) must never call it, and a best-effort variant must not be reintroduced;
- the "terminal only" paragraph now notes that `SettleDelegatedFailureRecoveriesForTask` re-checks status in SQL, so a mistaken early call updates 0 rows instead of destroying a recovery obligation.

Cross-checked before writing it: all 15 production call sites pass a `qtx` (`internal/service/task.go` ×9, `internal/handler/{chat,comment,workspace_revoke}.go` ×4, `internal/service/runtime_teardown.go` ×1); `HandleFailedTasks` is post-commit side effects only, downstream of `terminateTasksInTx`/`FailTask`. So the comment now matches the code rather than the other way round.

## 2. `reject_recovery_settlement()` left behind

Cleanup dropped the trigger but not the function. Both are dropped now, trigger first.

Verified in both directions on a clean database rather than assumed:

| | `pg_proc` rows after the test |
|---|---|
| before this commit | **1** (leaked) |
| after | **0** |

The `DROP FUNCTION IF EXISTS` also cleans up a leftover from an earlier run — after re-running on the database that still held the leaked function, the count went 1 → 0.

## Verification

`gofmt` clean, `go vet ./internal/service/`, `go build ./...`, `git diff --check` all pass.

The settlement tests do run here, on a **clean** database — the local blocker earlier in this review was schema drift in the shared dev DB (`postgres://…/multica` fails `migrate up` at `352_autopilot_quota_execution` with `relation "autopilot_quota_period" already exists`, i.e. applied but unrecorded — unrelated to this PR). Migrating a scratch database instead: `go test -run 'Settle|Settlement|DelegatedFailure|NonTerminalTask|BulkCancel' ./internal/service/` → `ok 2.1s`, including `TestBulkCancelRollsBackWhenSettlementFails` and `TestNonTerminalTaskWithAReceiptIsNotSettled`. Scratch DB dropped afterwards.

PR #7821, opened 2026-08-31
MUL-6883 perf(server): run the delegated-failure recovery outbox on its own leased worker

Bohan-J:
Step 5 of MUL-6883. Independent of #7820 — no overlapping files — but see **Merge order** below.

## Summary

The recovery outbox scan sat inside the 30-second runtime-liveness loop, so every backend replica ran the sweeper's most expensive query **2,880 times a day**. That cadence was never its requirement: the normal terminal-failure path dispatches the coordinator handoff immediately, and this sweep exists only to repair the narrow window where the recovery comment committed but its dispatch did not — a transient database error, or a process that exited in between.

It now has its own five-minute loop, runs **once at startup**, and each round is gated on a Redis lease so one round happens per window across the whole deployment. Per replica: **2,880 → 288 calls/day**, and the `× N` amplification is gone rather than reshuffled between replicas.

## The startup round is the one that matters

A process that just started is precisely the aftermath of the exit this sweep repairs, so scanning then covers the dominant case without waiting an interval. Reducing the cadence *without* it would have made the new arrangement strictly worse than the old one.

It is leased too, deliberately: what it repairs is global database state rather than anything this process owns, so during a rolling deploy one scan covers every replica restarted, not one scan per restart.

## The lease: two exclusions, not one

The gate needs **both** halves, and having only the first is the trap:

1. while a round runs, no other replica may join it;
2. after a round **succeeds**, no replica may start another until the cadence window reopens.

Holding a key for the round and `DEL`-ing it at the end gives only (1). Replica phases always differ — each starts its ticker when its own startup round finishes, and a rolling restart staggers them further — so replicas would simply take turns and the deployment would still run one scan per replica per window.

So the protocol is:

| | |
|---|---|
| **acquire** | `SET NX PX` with a per-round holder token |
| **renew** | every 10s while the round runs, token-compared in Lua |
| **succeed** | the admission is **converted** (not deleted) into a `cooldown` marker for the rest of the window, under the same token compare |
| **cut short** | released, token-compared — an interrupted round did not cover its window, so the replica taking over should not wait out a cadence someone else abandoned |

A round has no deadline of its own — the batch size bounds how many recoveries it dispatches, never how long the database takes to answer — so a fixed TTL could expire mid-round and let a second replica re-run the expensive query exactly when the database is already slow. Renewal removes the need to guess: **TTL (30s) is how long the admission survives *without* a renewal**, renewed at 10s so a single missed renewal is retried, and a replica that dies holding it frees the round in seconds rather than minutes.

The tick is a **1-minute poll, finer than the 5-minute cadence**. Ticking at the cadence would let each replica's ticker phase decide who runs, and a window whose claimant then died could sit unclaimed for nearly a full cadence. The cooldown enforces the rate; the poll only decides how promptly a live replica notices. A poll that loses costs one Redis `SETNX` and touches no database.

A **local floor** completes it: a replica will not start a round within a cadence of its own last completed one. That bounds the no-Redis deployment, which has no shared cooldown to consult, and keeps fail-open from turning the finer poll into a hot loop against the database.

**It fails open.** An unreachable Redis must not silently switch off a recovery path: a duplicated round repeats work that is already idempotent — concurrent sweepers are serialised by `AcknowledgeExhaustedDelegatedFailureRecovery`'s `FOR UPDATE` — while a skipped round strands an obligation. With no Redis configured at all (the Helm default `backend.replicas: 1`) every replica runs every round, exactly as today, at a tenth of the rate.

## Tradeoff: worst case is ~6 minutes, and that is accepted

The worst-case age of an obligation stranded by a crash that is **not** followed by a restart goes from 30 seconds to **six minutes** — `delegatedFailureRecoveryWorstCaseHandoff`, which is the cadence plus one poll, not the cadence alone.

The path that produces it: a replica completes a round, converts its admission into the cooldown, and then dies. The cooldown is shared state and outlives it, so nobody may start the next round until the window reopens five minutes later, and the successor notices on its next poll — up to a minute after that. No tighter bound exists while a shared cooldown is what enforces the global rate.

Accepted rather than reshaped, because what it governs is a **double fault**: a dispatch lost after its recovery comment committed **and** no restart afterwards — a restart scans immediately, and that is the dominant shape of this failure. Every other route to a recovery is unchanged and still immediate, and the 30 seconds this replaces was a side effect of sharing the liveness tick, never a requirement of the repair.

It is cheap to tighten if that ever changes: the poll is one line and costs one Redis `SETNX` per replica per interval, so a 30s poll would buy a 5m30s bound.

## Observability

The `delegated_failure_recovery` stage of `multica_runtime_sweeper_stage_duration_seconds` is unchanged, so existing dashboards keep working — it is simply emitted by whichever replica held the round. A skipped round emits nothing rather than a zero-duration, zero-candidate sample, which would be indistinguishable from a round that ran and found nothing.

## Merge order

Sequenced after #7820's backfill is verified in production. Dropping the call rate first would mask whether the settled-marker index actually reduced per-call cost — you would see total load fall either way. Nothing here depends on #7820's code, so the two can merge in either order technically; this is a measurement constraint, not a build one.

## Testing

```
go test -race ./cmd/server/     # pass (clean, migrated database)
go test ./internal/service/     # pass
go vet ./cmd/server/            # pass
```

Coverage of the global property, which is what a mutex-shaped implementation passes and a cadence gate must:

- `TestScanRateDoesNotGrowWithReplicaCount` — 4 staggered replicas vs 1, across several windows; 27 vs 7 rounds under delete-on-finish
- `TestRollingRestartProducesOneStartupScan` — 5 staggered starts, exactly 1 effective scan; 5 under delete-on-finish
- `TestSlowRoundExcludesOtherReplicasDuringAndAfter` — no second replica during a round that outlives the old fixed TTL, **and** none after it in the same window
- `TestWorstCaseRecoveryHandoffIsTheConfirmedBound` — pins cadence + poll = the documented 6m, so moving either input has to move the recorded decision
- `TestDeadHolderHandsOverWithinOnePollOfTheCooldown` — after its holder stops, the window is picked up no earlier than the cooldown and within one poll of it; a cadence-ticking successor takes 799ms against a 460ms bound in the same setup
- `TestAbandonedRoundIsHandedBackImmediately`, `TestFinishAfterLosingTheLeaseDoesNotAffectTheNewHolder`, `TestLocalCadenceFloorHoldsWithoutALease`
- `TestRedisSweepLeaseRenewsDuringTheRoundAndCoolsDownAfter`, `TestRedisSweepLeaseReleasesAnAbandonedRound`, `TestRedisSweepLeaseRefusesACoveredWindow`, `TestRedisSweepLeaseFailsOpen`, `TestUnleasedAlwaysRuns` — the wire protocol behind the above
- `TestDelegatedFailureRecoveryRunsOutsideTheLivenessLoop`, `TestDelegatedFailureRecoveryGlobalCallRate`, `TestDelegatedFailureRecoveryLeaseTimings`, `TestLeasedSweepRunsAtStartupBeforeTheFirstTick`, `TestLeasedSweepSkipsWindowsItDoesNotWin`, `TestLeasedSweepStopsWithItsContext`

Timing-sensitive cases were also run `-count=8 -race` for jitter.

`cmd/multica`, `internal/cli`, `internal/daemon` and `internal/daemon/execenv` fail in my sandbox because it runs inside an agent task context (`daemon_task_context.json` present, task-scoped `MULTICA_TOKEN`). They fail identically on unmodified `main`, so they are environmental, not caused by this change.

### DIFF
--- a/server/cmd/server/delegated_failure_recovery_sweeper_test.go
+++ b/server/cmd/server/delegated_failure_recovery_sweeper_test.go
@@ -141,6 +141,88 @@ func TestDelegatedFailureRecoveryLeaseTimings(t *testing.T) {
 	}
 }
 
+// The bound is six minutes, not the five the cadence suggests, and that is
+// accepted rather than papered over (MUL-6883). Pin the composition so that
+// moving either input has to move the recorded decision with it: a holder that
+// dies straight after a round leaves a cadence-long cooldown behind, and its
+// successor only notices the reopening on its next poll.
+func TestWorstCaseRecoveryHandoffIsTheConfirmedBound(t *testing.T) {
+	const confirmed = 6 * time.Minute
+	if delegatedFailureRecoveryWorstCaseHandoff != confirmed {
+		t.Fatalf("worst-case handoff = %s, but %s is what was confirmed as acceptable",
+			delegatedFailureRecoveryWorstCaseHandoff, confirmed)
+	}
+	if got := delegatedFailureRecoverySweepCadence + delegatedFailureRecoveryPollInterval; got != delegatedFailureRecoveryWorstCaseHandoff {
+		t.Fatalf("cadence %s + poll %s = %s, which no longer matches the documented bound %s",
+			delegatedFailureRecoverySweepCadence, delegatedFailureRecoveryPollInterval,
+			got, delegatedFailureRecoveryWorstCaseHandoff)
+	}
+}
+
+// What makes that bound a cadence plus a POLL rather than a cadence plus a
+// cadence: the successor is watching finer than the rate it is enforcing. A
+// replica ticking at the cadence would miss a reopening it fell a hair short of
+// and wait out another full window — which is the shape this measures, by
+// setting the successor's phase just before the cooldown begins.
+func TestDeadHolderHandsOverWithinOnePollOfTheCooldown(t *testing.T) {
+	const (
+		cooldown = 400 * time.Millisecond
+		poll     = 60 * time.Millisecond
+		// Scheduler slack only. The bound under test is cooldown + poll; a
+		// cadence-ticking successor would need up to cooldown + cooldown.
+		slack = 120 * time.Millisecond
+	)
+
+	for attempt := range 3 {
+		shared := &cadenceLease{cooldown: cooldown}
+		finishA, ok := (replicaLease{shared: shared, name: "replica-a"}).Acquire(context.Background())
+		if !ok {
+			t.Fatalf("attempt %d: the first replica could not acquire", attempt)
+		}
+
+		ctx, cancel := context.WithCancel(context.Background())
+		swept := make(chan time.Time, 1)
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			runLeasedSweep(ctx, poll, cooldown, replicaLease{shared: shared, name: "replica-b"}, func() {
+				select {
+				case swept <- time.Now():
+				default:
+				}
+			})
+		}()
+		// B's startup attempt is refused while A still holds the round, so B's
+		// poll phase is fixed just before the cooldown starts.
+		time.Sleep(2 * time.Millisecond)
+		completedAt := time.Now()
+		// A completes and is then never heard from again: its cooldown is
+		// shared state and outlives it.
+		finishA(true)
+
+		var sweptAt time.Time
+		select {
+		case sweptAt = <-swept:
+		case <-time.After(cooldown + poll + slack + time.Second):
+			cancel()
+			<-done
+			t.Fatalf("attempt %d: the window was never picked up after its holder stopped", attempt)
+		}
+		cancel()
+		<-done
+
+		delay := sweptAt.Sub(completedAt)
+		if delay < cooldown {
+			t.Fatalf("attempt %d: the next round started %s after the last one, inside the %s cooldown",
+				attempt, delay, cooldown)
+		}
+		if delay > cooldown+poll+slack {
+			t.Fatalf("attempt %d: handover took %s, past the cooldown plus one poll (%s)",
+				attempt, delay, cooldown+poll)
+		}
+	}
+}
+
 // The point of the gate: the number of scans the deployment performs must be
 // set by the cadence, not by how many replicas are running. Deleting the key at
 // the end of a round would satisfy the concurrency test below and still fail
--- a/server/cmd/server/delegated_failure_recovery_sweeper_test.go
+++ b/server/cmd/server/delegated_failure_recovery_sweeper_test.go
@@ -1,0 +1,187 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"os"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	redismock "github.com/go-redis/redismock/v9"
+)
+
+type stubSweepLease struct {
+	grant    bool
+	attempts atomic.Int32
+	lastTTL  atomic.Int64
+}
+
+func (l *stubSweepLease) Acquire(_ context.Context, ttl time.Duration) bool {
+	l.attempts.Add(1)
+	l.lastTTL.Store(int64(ttl))
+	return l.grant
+}
+
+// The recovery outbox is a crash backstop, not a liveness signal. Leaving it in
+// the 30-second loop is what made a scan that usually finds nothing the most
+// frequent caller of the most expensive query in the sweeper.
+func TestDelegatedFailureRecoveryRunsOutsideTheLivenessLoop(t *testing.T) {
+	if delegatedFailureRecoverySweepInterval != 5*time.Minute {
+		t.Fatalf("recovery sweep interval = %s, want 5m", delegatedFailureRecoverySweepInterval)
+	}
+
+	source, err := os.ReadFile("runtime_sweeper.go")
+	if err != nil {
+		t.Fatalf("read runtime_sweeper.go: %v", err)
+	}
+	start := strings.Index(string(source), "func runRuntimeSweeper(")
+	end := strings.Index(string(source), "func runRuntimeGCSweeper(")
+	if start < 0 || end <= start {
+		t.Fatal("could not isolate runtime sweeper loops")
+	}
+	if strings.Contains(string(source[start:end]), "sweepPendingDelegatedFailureRecoveries(") {
+		t.Fatal("delegated failure recovery is still invoked from the 30-second liveness loop")
+	}
+}
+
+// The per-replica call rate is the whole point of the change, so pin it rather
+// than leaving it to be re-derived from a constant someone may retune.
+func TestDelegatedFailureRecoveryDailyCallRate(t *testing.T) {
+	const wantCallsPerDay = 288
+	got := int(24 * time.Hour / delegatedFailureRecoverySweepInterval)
+	if got != wantCallsPerDay {
+		t.Fatalf("recovery sweep rate = %d/day/replica, want %d/day/replica", got, wantCallsPerDay)
+	}
+}
+
+// A lease that outlives its round would let a replica that died holding it
+// suppress the next one; a lease far shorter than the round would let a second
+// replica start the same work while the first is still in it.
+func TestDelegatedFailureRecoveryLeaseTTLFitsInsideTheInterval(t *testing.T) {
+	if delegatedFailureRecoveryLeaseTTL >= delegatedFailureRecoverySweepInterval {
+		t.Fatalf("lease TTL %s must be shorter than the %s interval, or a dead holder suppresses the next round",
+			delegatedFailureRecoveryLeaseTTL, delegatedFailureRecoverySweepInterval)
+	}
+	if delegatedFailureRecoveryLeaseTTL < delegatedFailureRecoverySweepInterval/2 {
+		t.Fatalf("lease TTL %s is short enough that a slow round could run twice concurrently", delegatedFailureRecoveryLeaseTTL)
+	}
+}
+
+// The startup scan is the one that matters: a process that just started is the
+// aftermath of the exit this sweep repairs. Waiting a full interval for it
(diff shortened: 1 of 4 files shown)


Bohan-J:
Pushed `ea99ca8c7` — took option 2 (token-compared renewal). The must-fix was correct and the reasoning behind my fixed-TTL note was wrong: a round has no deadline of its own, so **no TTL picked in advance can be proven to cover the slowest round**, and the window opens exactly when it hurts most.

## What changed

`SET NX PX` and forget → a lease held for the whole round:

- **acquire** writes a fresh per-round holder token
- **renew** extends the admission on a timer while the round runs, comparing that token so a replica which already lost its lease cannot extend whatever the new holder wrote
- **release** compares the same token and deletes, so the next round is contested immediately rather than after a timeout

Release is detached from the caller's context, so shutdown still hands the round over instead of leaving the key to expire.

## TTL re-derived, per your note

With renewal, the TTL bounds how long the admission survives **without** a renewal — not how long a round may take. That removes the guess entirely:

| | before | after |
|---|---|---|
| TTL | 4 min (had to cover a whole round) | **30 s** (only has to cover a missed renewal) |
| renewal | none | **10 s** — a third of the TTL, so one missed renewal is retried before expiry |
| dead holder frees the next round in | up to 4 min | up to 30 s |

So the change is strictly better on the failure mode the old TTL was sized for.

## Tests

Replaced the `TTL > interval/2` assertion — you were right that it proves nothing — with the two you asked for:

- **`TestSlowRoundKeepsASecondReplicaOut`** — replica A enters a round that outlives the old TTL; replica B ticks repeatedly throughout and must run zero rounds; when A finishes, B takes the released admission immediately rather than waiting for expiry. Verified to fail without the fix: **21 concurrent rounds**.
- **`TestReleaseAfterLosingTheLeaseDoesNotEvictTheNewHolder`** — the token compare on the release path.
- **`TestRedisSweepLeaseRenewsDuringTheRoundAndReleasesAfter`** — the Redis wire protocol: the renewals arrive *while the round is still running* and carry the acquire token, and the release carries the same token.
- `TestDelegatedFailureRecoveryLeaseTimings` now pins the relationships that actually matter: renewal ≤ TTL/2, and TTL < sweep interval.

## One consequence worth naming

A genuinely *hung* round now renews indefinitely instead of being taken over at TTL. I think that is the correct reading of a single-instance gate — a second replica joining a round stuck on the database would only add load to the thing that is already stuck — and the process context remains what ends it. Happy to add a round timeout on top (your option 1) if you'd rather have both belts; I left it out because a timeout short enough to be useful would need a latency budget we don't have data for yet.

## Verification

- `go test ./cmd/server -race`: pass, and `-count=8 -race` on the lease/concurrency tests for flake
- `go test ./...`: same four sandbox-environmental packages as before (`cmd/multica`, `internal/cli`, `internal/daemon*` — agent task context), identical on unmodified `main`

On your note about the local run: `./cmd/server -race` completed here against a **freshly migrated** database. The missing `conversation_starters` / `durable_work_dir` columns are a stale reused test DB rather than anything in this branch — `go run ./cmd/migrate up` on a clean database fixes it.

Bohan-J:
Pushed `fb9a9833f`. The review is right: the previous commit fixed concurrent overlap and left the actual saving unimplemented. Holding the key for the round and `DEL`-ing at the end is a mutex, not a cadence gate — out-of-phase replicas just take turns, and phases *always* differ because each replica starts its ticker when its own startup round finishes. The `× N` reduction never happened, and my `TestSlowRoundKeepsASecondReplicaOut` asserted precisely the handover that made it wrong.

## The protocol now has both halves

The admission no longer ends at `DEL`. On success it is **converted**, under the same holder-token compare, into a cooldown that blocks every replica for the rest of the window:

```lua
if redis.call('GET', KEYS[1]) == ARGV[1] then
  redis.call('SET', KEYS[1], ARGV[2], 'PX', ARGV[3])   -- cooldown marker
  return 1
end
return 0
```

A round cut short by shutdown still releases — it did not cover its window, and the replica taking over shouldn't wait out a cadence someone else abandoned. So `finish(completed bool)` distinguishes the two outcomes rather than having one exit path.

## The tick is now a poll, not the cadence

This is the part that makes the cooldown correct rather than merely present. If replicas tick *at* the cadence, the phase of whichever replica happens to tick first after the window opens decides everything — and if that replica dies, the window can stay unclaimed for nearly a full second cadence. So:

| | value | role |
|---|---|---|
| cadence | 5 min | the cooldown — the **global** rate |
| poll | 1 min | how promptly a live replica *notices* an open window |
| lease TTL / renew | 30 s / 10 s | in-flight exclusion (unchanged) |

A poll that loses costs one Redis `SETNX` and touches no database.

A **local floor** completes it: a replica won't start a round within a cadence of its own last completed one. That bounds the no-Redis deployment, which has no shared cooldown to consult, and stops fail-open from turning the finer poll into a hot loop.

## Tests now assert the global property

Replacing the mutual exclusion that was standing in for it:

- **`TestScanRateDoesNotGrowWithReplicaCount`** — four staggered replicas vs one, over multiple windows; the count must not scale with replica count. Under delete-on-finish: **27 rounds vs 7**.
- **`TestRollingRestartProducesOneStartupScan`** — five staggered startups must produce exactly one effective scan. Under delete-on-finish: **5**.
- **`TestSlowRoundExcludesOtherReplicasDuringAndAfter`** — the corrected version of the test you flagged. It still requires B out *during* A's slow round, and now also requires B to stay out **after** A finishes. Under delete-on-finish: B runs immediately.
- `TestLocalCadenceFloorHoldsWithoutALease`, `TestAbandonedRoundIsHandedBackImmediately`, plus a redismock test that a completed round issues the **cooldown** command and an abandoned one issues **release**.

Kept: mid-round exclusion, and the stale-finish-does-not-evict-the-new-holder case.

## Consequence worth naming

With the cooldown, if the replica that just completed a window dies, the next scan waits out the remaining cooldown and then up to one poll — so ≤ 6 minutes rather than ≤ 5. The finer poll is what keeps that bounded; at a 5-minute tick it would have been ≤ 10. I sized the poll at 1 minute for that reason, and it's a one-line change if you'd rather tighten it.

## Verification

- `go test ./cmd/server -race`: pass; `-count=10 -race` on the timing-sensitive cadence tests for flake
- `go test ./...`: same four sandbox-environmental packages (`cmd/multica`, `internal/cli`, `internal/daemon*` — agent task context), identical on unmodified `main`

On your local blocker: the missing `conversation_starters` / `durable_work_dir` / `source_context_id` columns are a stale reused test database. `./cmd/server -race` completes here against a freshly migrated one.

multica-eve:
Taking the closing question with **option 2**: the worst case is ~6 minutes, that is accepted, and the three places that still said five have been brought in line. Pushed as `3f4dbd25a` — no scheduling change, so the protocol you reviewed is unchanged.

## The bound, and why 6m rather than 5m

Confirmed by the requester on MUL-6883. The derivation is worth stating precisely because it is not the cadence:

> A replica completes a round → converts its admission into the cadence cooldown → dies. The cooldown is **shared state and outlives it**, so nobody may start the next round until the window reopens a cadence later, and the successor only notices the reopening on its next poll.

So the bound is **cadence + poll = 5m + 1m**, and no tighter bound exists while a shared cooldown is what enforces the global rate. Your review's arithmetic was right; the code and description were repeating the number the cadence alone would give.

Accepted rather than reshaped because of what it actually governs — a **double fault**: a dispatch lost after its recovery comment committed **and** no restart afterwards. A restart runs the scan immediately, and that is the dominant shape of this failure. Every other route to a recovery is unchanged and still immediate, and the 30 seconds this replaced was a side effect of sharing the liveness tick, never a requirement of the repair. If that judgment ever changes it is one line: the poll costs one Redis `SETNX` per replica per interval and no database work, so a 30s poll buys a 5m30s bound.

## Code comments

New `delegatedFailureRecoveryWorstCaseHandoff` constant, **derived** from its two inputs rather than restating a number, carrying the derivation and the acceptance rationale. `runtime_sweeper.go:217-224` (your citation) now names it instead of saying "now five minutes". Two supporting comments point at it: the poll interval, which is the second term, and the `sweepLease` contract, where `finish(true)`'s cooldown is where the extra minute comes from.

## Regression tests

Two, splitting the arithmetic from the property:

- **`TestWorstCaseRecoveryHandoffIsTheConfirmedBound`** — pins `cadence + poll == 6m == the documented constant`, so changing either input has to change the recorded decision with it rather than silently moving the bound.
- **`TestDeadHolderHandsOverWithinOnePollOfTheCooldown`** — the property that keeps the second term a *poll* and not another cadence. The holder completes a round and is never heard from again; the successor's phase is deliberately set just before the cooldown begins — the phase where a cadence-ticking replica misses the reopening by a hair. Handover must land no earlier than the cooldown and within one poll of it.

The second one discriminates, not just documents: swapping the successor's poll for the cadence makes it fail at **799ms against a 460ms bound**, i.e. it waits out a second full window. That is the failure mode your first-round review predicted, now pinned.

## PR description

Rewritten. It still described the protocol from the first commit — `SET NX PX` "is the whole protocol — no renewal, no release", a 4-minute TTL, and a 5-minute worst case — all of which two rounds of review replaced. It now documents the holder-token renewal, cooldown-on-success vs release-on-interruption, the finer poll, the local floor, and the 6-minute bound with its derivation.

## Verification

`go vet ./cmd/server/`, `gofmt` clean, `git diff --check`. `go test -race -count=1 ./cmd/server/` passes **in full** on a clean migrated database (10.2s); the new timing test also ran `-count=8 -race` for jitter. The failure you and Sol-Boy both hit locally is the reused dev database missing existing columns (`conversation_starters`, `durable_work_dir`, `source_context_id`) — on this machine that database also refuses `migrate up` at `352_autopilot_quota_execution` (`relation "autopilot_quota_period" already exists`, applied but unrecorded), so I migrated a scratch database instead and dropped it afterwards. CI is running on the push; I did not wait on it.

Issue #7452, opened 2026-08-23
[Bug]: A missed empty-claim invalidation can stall an idle runtime's queued task for ~3 minutes


yulin7645:
### Deployment type

Official App

### Area

Agents and runtimes

### What happened?

A newly queued issue-comment task remained unclaimed for 205 seconds even though both the runtime host and the target agent had free concurrency capacity. A direct-chat task for the same agent and runtime was created 67 seconds later, but it was dispatched first. The older issue task was then dispatched in the exact second the chat task completed.

This does not appear to be provider serialization or host capacity exhaustion:

- The daemon had a machine concurrency limit of 5, and no other tasks were active during the affected interval.
- The agent had `max_concurrent_tasks = 6`.
- A follow-up test ran two independent 100-second tasks on the same agent and same Kimi runtime with approximately 94 seconds of real process overlap.
- A separate task created shortly after the incident was dispatched in about one second while the older issue task was already running.

The observed timeline, relative to the first enqueue, was:

| Event | Relative time |
|---|---:|
| Issue-comment task queued | T+0s |
| Direct-chat task queued | T+67s |
| Chat task dispatched | T+188s |
| Chat task started | T+190s |
| Chat task completed | T+205s |
| Older issue task dispatched and started | T+205s |

The 205-second delay strongly matches the 3-minute empty-claim cache TTL plus the daemon's next polling opportunity:

- `EmptyClaimCacheTTL` is 3 minutes, and its source comment says the TTL is the safety net for a missed invalidation, for example a transient Redis failure during `Bump`.
- Every enqueue is expected to bump the runtime's empty-claim version before sending the daemon wakeup.
- The batch candidate query orders by `priority DESC, created_at ASC`. Direct chat uses priority 2, while the affected issue task had priority 0, which explains why the later chat was selected first once both rows became visible.
- `ClaimTasksForRuntimes` tracks `triedAgents` and claims at most one fresh task per distinct agent in one batch. This explains why the still-free host did not receive both same-agent tasks in that batch.
- Task completion calls `NotifyTaskFinished`, which bumps/wakes the runtime again. That matches the older issue task starting in the same second the chat completed.

I do not have the hosted server's Redis logs for the incident, so a missed empty-cache invalidation is a strong timing- and source-backed hypothesis, not a log-confirmed root cause. The externally verifiable bug is the liveness failure: a task on an idle, online runtime remained queued for roughly the full negative-cache fallback period.

Expected behavior:

[...]

PR #7824, opened 2026-08-31
MUL-6895, MUL-6586: fix(daemon,tasks): force a server-side re-check for woken runtimes to clear a stale empty-claim stall

ADanMan:
## What

A queued task for an idle runtime can stall for up to ~3 minutes. On enqueue the server bumps an empty-claim verdict cache (`empty_claim_cache.go`, 3m TTL) and the claim path short-circuits any runtime whose `EmptyClaim.IsEmpty` is true. When the invalidating `Bump` is lost to a transient Redis failure, the stale "empty" verdict survives to the TTL and the daemon's 30s idle polls keep getting nothing for that runtime — the failure #7452 reports (a task queued for an online, idle-capacity runtime dispatched only ~205s later).

This wires the issue's suggested direction #1: the daemon's targeted `task_available` wakeup now forces a server-side DB re-check that **bypasses the `IsEmpty` short-circuit for exactly the woken runtimes**, while still `MarkEmpty`-repairing on a zero-candidate result so the cache re-arms. The empty-claim optimization is untouched for every runtime that was not just woken.

## How

- **Server** (`service/task.go`, `ClaimTasksForRuntimes`): a variadic `forceRecheckRuntimeIDs`. At the step-3 empty-cache filter, forced runtimes skip `IsEmpty` but still flow through the step-5 `MarkEmpty` re-arm on zero candidates. Non-forced runtimes are byte-for-byte unchanged.
- **Transport** (`handler/daemon.go`, `daemon/client.go`, `daemon/wsrpc.go`): an **optional, additive** `force_recheck_runtime_ids` on the batch/WS claim request. Parsed defensively (`util.ParseUUID`, skip on error) and **intersected with the daemon's authorized runtime set**, so a daemon can never force a re-check for a runtime it does not own. The WS `tasks.claim` RPC reuses the same HTTP handler, so authorization is not duplicated.
- **Daemon** (`daemon/daemon.go`): a mutex-guarded woken-runtime set. `pollLoop` records the runtime id on each targeted wakeup; `runBatchPoller` drains the set into the next claim and re-notes it on claim error. The set is omitted from the request body entirely when empty, so a steady-state poll is identical to the pre-fix request.

## Testing

RED→GREEN, both directions:
- `service/task_batch_claim_force_recheck_test.go`: a forced runtime bypasses a stale empty verdict and claims while a non-forced runtime still short-circuits; a forced zero-candidate runtime re-arms the cache. (Reverting the step-3 bypass or the step-5 re-arm fails these.)
- `daemon/force_recheck_test.go`: the woken set is tracked, deduped and cleared; an empty wakeup is ignored; `claimTasksBody` omits the field with nothing woken and includes it otherwise.

`go build ./...`, `go vet`, `gofmt -l` clean; the daemon suite and the handler batch-claim tests pass.

## API compatibility

The field is additive/optional and safe in both directions. Newer daemon → older server: the unknown JSON field is ignored, behavior falls back to today's TTL. Older daemon → newer server: the field is absent, decodes to nil, and the non-forced path is unchanged. Empty woken set → the field is omitted, so the request is byte-identical to before.

## Known limitations (narrow, self-healing — flagged for a possible follow-up)

The common-case stall is closed. Two rare interleavings still fall back to the existing TTL for a single runtime (never a correctness regression, never a double-claim, always self-healing on the next wakeup/enqueue or the 3m TTL):

1. If a WS claim returns the "uncertain" outcome (request may not have reached the server), the already-drained woken set is not re-noted (the re-note is on the error path only), so that runtime is not force-rechecked that cycle.
2. If the step-2 stale-dispatch reclaim alone fills the batch's `maxTasks`, `ClaimTasksForRuntimes` returns before the step-3 forced bypass runs, consuming the signal without effect for that cycle.

The singular per-runtime `ClaimTaskForRuntime` path (legacy fallback for servers without the batch route, which therefore lack this feature) is intentionally not wired — threading force there would be dead code.

Addresses #7452.

### DIFF
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -265,20 +265,33 @@ const batchClaimRequestTimeout = 5 * time.Second
 // (batchClaimRequestTimeout) rather than the shared 30s control-plane timeout so
 // one slow claim cannot stall the whole batch; the deadline propagates to the
 // server and cancels the in-flight query there too.
-func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int) ([]*Task, error) {
+func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) ([]*Task, error) {
 	reqCtx, cancel := context.WithTimeout(ctx, batchClaimRequestTimeout)
 	defer cancel()
 	var resp struct {
 		Tasks []*Task `json:"tasks"`
 	}
-	if err := c.postJSON(reqCtx, "/api/daemon/tasks/claim", map[string]any{
-		"daemon_id":   daemonID,
-		"runtime_ids": runtimeIDs,
-		"max_tasks":   maxTasks,
-	}, &resp); err != nil {
+	if err := c.postJSON(reqCtx, "/api/daemon/tasks/claim", claimTasksBody(daemonID, runtimeIDs, maxTasks, forceRecheckIDs...), &resp); err != nil {
 		return nil, err
 	}
 	return resp.Tasks, nil
+}
+
+// claimTasksBody builds the batch-claim request body shared by the HTTP and
+// WS-first transports so both stay byte-identical. force_recheck_runtime_ids
+// (#7452) is an optional, additive field: it is omitted when no runtime was
+// woken so a steady-state poll is exactly the pre-#7452 request, and an older
+// server that does not know the field simply ignores it.
+func claimTasksBody(daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) map[string]any {
+	body := map[string]any{
+		"daemon_id":   daemonID,
+		"runtime_ids": runtimeIDs,
+		"max_tasks":   maxTasks,
+	}
+	if len(forceRecheckIDs) > 0 {
+		body["force_recheck_runtime_ids"] = forceRecheckIDs
+	}
+	return body
 }
 
 // isBatchClaimUnsupported reports whether err is a 404 from the batch claim
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -266,17 +266,19 @@ const batchClaimRequestTimeout = 5 * time.Second
 // one slow claim cannot stall the whole batch; the deadline propagates to the
 // server and cancels the in-flight query there too.
 //
-// The second return value is the server's force_rechecked_runtime_ids echo
-// (#7452): the forced runtimes the server actually scanned this cycle. The
-// caller consumes the wakeup hint only for those and re-notes the rest. An older
-// server omits the field, yielding an empty echo so every forced runtime is
-// re-noted (TTL remains the outer bound).
-func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) ([]*Task, []string, error) {
+// The second return value is the server's force_rechecked_runtime_ids
+// acknowledgement (#7452): the forced runtimes the server observed genuinely
+// idle this cycle. It is a POINTER so absence is distinguishable from
+// emptiness: a non-nil empty slice means an upgraded server acknowledged
+// nothing (keep every forced runtime forced), while nil means the server
+// omitted the field entirely and cannot honour the hint at all, so the caller
+// drops it rather than re-forcing forever.
+func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) ([]*Task, *[]string, error) {
 	reqCtx, cancel := context.WithTimeout(ctx, batchClaimRequestTimeout)
 	defer cancel()
 	var resp struct {
-		Tasks                    []*Task  `json:"tasks"`
-		ForceRecheckedRuntimeIDs []string `json:"force_rechecked_runtime_ids"`
+		Tasks                    []*Task   `json:"tasks"`
+		ForceRecheckedRuntimeIDs *[]string `json:"force_rechecked_runtime_ids"`
 	}
 	if err := c.postJSON(reqCtx, "/api/daemon/tasks/claim", claimTasksBody(daemonID, runtimeIDs, maxTasks, forceRecheckIDs...), &resp); err != nil {
 		return nil, nil, err
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -265,16 +265,23 @@ const batchClaimRequestTimeout = 5 * time.Second
 // (batchClaimRequestTimeout) rather than the shared 30s control-plane timeout so
 // one slow claim cannot stall the whole batch; the deadline propagates to the
 // server and cancels the in-flight query there too.
-func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) ([]*Task, error) {
+//
+// The second return value is the server's force_rechecked_runtime_ids echo
+// (#7452): the forced runtimes the server actually scanned this cycle. The
+// caller consumes the wakeup hint only for those and re-notes the rest. An older
+// server omits the field, yielding an empty echo so every forced runtime is
+// re-noted (TTL remains the outer bound).
+func (c *Client) ClaimTasks(ctx context.Context, daemonID string, runtimeIDs []string, maxTasks int, forceRecheckIDs ...string) ([]*Task, []string, error) {
 	reqCtx, cancel := context.WithTimeout(ctx, batchClaimRequestTimeout)
 	defer cancel()
 	var resp struct {
-		Tasks []*Task `json:"tasks"`
+		Tasks                    []*Task  `json:"tasks"`
+		ForceRecheckedRuntimeIDs []string `json:"force_rechecked_runtime_ids"`
 	}
 	if err := c.postJSON(reqCtx, "/api/daemon/tasks/claim", claimTasksBody(daemonID, runtimeIDs, maxTasks, forceRecheckIDs...), &resp); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	return resp.Tasks, nil
+	return resp.Tasks, resp.ForceRecheckedRuntimeIDs, nil
 }
 
 // claimTasksBody builds the batch-claim request body shared by the HTTP and
--- a/server/internal/daemon/client_batch_claim_test.go
+++ b/server/internal/daemon/client_batch_claim_test.go
@@ -36,7 +36,7 @@ func TestClient_ClaimTasks_PostsRuntimeSetAndParsesTasks(t *testing.T) {
 	c := NewClient(srv.URL)
 	c.SetToken("tok")
 
-	tasks, err := c.ClaimTasks(context.Background(), "daemon-x", []string{"rt-a", "rt-b", "rt-c"}, 3)
+	tasks, _, err := c.ClaimTasks(context.Background(), "daemon-x", []string{"rt-a", "rt-b", "rt-c"}, 3)
 	if err != nil {
 		t.Fatalf("ClaimTasks: %v", err)
 	}
@@ -73,7 +73,7 @@ func TestClient_ClaimTasks_EmptyResult(t *testing.T) {
 	c := NewClient(srv.URL)
 	c.SetToken("tok")
 
-	tasks, err := c.ClaimTasks(context.Background(), "daemon-x", []string{"rt-a"}, 1)
+	tasks, _, err := c.ClaimTasks(context.Background(), "daemon-x", []string{"rt-a"}, 1)
 	if err != nil {
 		t.Fatalf("ClaimTasks: %v", err)
 	}
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -520,6 +520,14 @@ type Daemon struct {
 	pendingWorkInflight map[string]struct{}  // runtime_id -> hint-driven heartbeat in flight
 	pendingWorkLastRun  map[string]time.Time // runtime_id -> when the last hint-driven heartbeat started
 
+	// forceRecheckMu guards forceRecheckRuntimes, the set of runtimes woken by a
+	// targeted `task_available` since the last batch claim (#7452). The next
+	// claim carries them so the server bypasses their cached "empty" verdict,
+	// repairing a stale verdict left by a lost EmptyClaim.Bump instead of
+	// stalling the queued task until the 3-minute TTL.
+	forceRecheckMu       sync.Mutex
+	forceRecheckRuntimes map[string]struct{} // runtime_id -> woken since last claim
+
 	cancelFunc context.CancelFunc // set by Run(); called by triggerRestart
 	rootCtx    context.Context    // set by Run(); used by long-running recoveries that must survive per-runtime ctx cancellation
 	// restartMu guards restartBinary. Two goroutines can reach triggerRestart —
@@ -4750,9 +4758,13 @@ func (d *Daemon) pollLoop(ctx context.Context, taskWakeups <-chan taskWakeup) er
 			// The batch poller re-derives allRuntimeIDs() each cycle; nudge it to
 			// pick up a registered/removed runtime promptly.
 			nudge()
-		case <-taskWakeups:
+		case wk := <-taskWakeups:
 			// Targeted-runtime and catch-up wakeups both trigger one batch claim
-			// across the whole runtime set.
+			// across the whole runtime set. A targeted wakeup also records its
+			// runtime so the next claim forces a server-side re-check that bypasses
+			// a stale cached "empty" verdict (#7452); a catch-up wakeup has no
+			// runtime id and only nudges.
+			d.noteWokenRuntime(wk.runtimeID)
 			nudge()
 		}
 	}
@@ -4815,10 +4827,19 @@ func (d *Daemon) runBatchPoller(pollerCtx, parentCtx context.Context, sem chan i
 			continue
 		}
 
-		tasks, err := d.ClaimTasksWSFirst(pollerCtx, d.cfg.DaemonID, runtimeIDs, len(slots))
+		// Drain the runtimes woken since the last claim and force a server-side
+		// re-check for them so a stale cached "empty" verdict cannot strand their
+		// queued task until the TTL (#7452).
+		forceRecheck := d.drainWokenRuntimes()
+		tasks, err := d.ClaimTasksWSFirst(pollerCtx, d.cfg.DaemonID, runtimeIDs, len(slots), forceRecheck...)
 		if err != nil {
 			d.exitClaim()
 			releaseSlots(slots)
+			// Preserve the force signal across a transient claim failure so the
+			// woken runtime is still re-checked on the next cycle.
+			for _, rid := range forceRecheck {
+				d.noteWokenRuntime(rid)
+			}
 			if pollerCtx.Err() == nil {
 				d.logger.Warn("batch claim failed", "error", err)
 			}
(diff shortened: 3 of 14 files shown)


Bohan-J (changes requested):
Thanks for digging into #7452 — the diagnosis is right and so is the direction. Using the targeted `task_available` wakeup as a second, independent source of truth ("this runtime may have work now") is the correct way to stop a lost `EmptyClaim.Bump` from turning a performance cache into a correctness dependency. The additive/optional `force_recheck_runtime_ids` field is also the right compatibility shape: either side can roll back independently, and a steady-state poll is byte-identical to the pre-#7452 request.

Two things block merge, though. Both are cases where the fix does not actually close the loop on the failure it targets.

## 1. A forced re-check that finds candidates leaves the stale verdict in place

`server/internal/service/task.go:3883` only calls `MarkEmpty` for zero-candidate runtimes. A runtime that *did* have candidates is intentionally not re-marked — but nothing invalidates the **stale verdict that was already there**. `EmptyClaim.Bump` is only called from the enqueue path (`notifyRuntimeMayHaveWork`), never from the claim path, so after a successful forced claim the bad verdict is still live and still tagged with the current version.

That matters because `server/internal/service/task.go:3900` keeps the existing "at most one attempt per agent per batch" rule. So with two queued tasks for the same agent on the same runtime:

1. forced claim → bypass fires, 1 task claimed, second candidate skipped by `triedAgents`;
2. next normal poll → `IsEmpty` still true → short-circuit, 0 tasks;
3. the second task waits for another successful bump/wakeup, or the 3-minute TTL.

This is the exact amplification #7452 describes, just moved one task later. We confirmed it on an isolated Postgres + Redis using the PR's own `batchClaimFixture` (which already queues two same-agent tasks on rt1): forced claim returns 1, the immediately following normal claim returns **0** instead of the remaining task. `server/internal/service/task_batch_claim_force_recheck_test.go:49` stops asserting right before this point.

Fix: on a forced positive result, either invalidate the runtime's verdict, or keep the runtime in the forced state until its remaining candidates have had a normal chance to be claimed.

## 2. The force signal is consumed even when no claim was issued

`server/internal/daemon/daemon.go:4833` drains the woken set *before* the call and only restores it when `err != nil`. But `server/internal/daemon/wsrpc.go:333` returns `(nil, nil)` while the uncertain-claim cooldown is still open — **no WS request, no HTTP request, nothing sent at all**. Any targeted wakeup that arrives during that window is dropped permanently.

The comment at `wsrpc.go:358` justifies not replaying the signal after an *uncertain* claim (don't risk a double-claim), and that reasoning is sound for that branch. It does not apply to the cooldown branch: no claim was attempted, so there is nothing to double. That signal should survive to the HTTP fallback the cooldown itself schedules.

The same consume-without-executing issue exists at `server/internal/service/task.go:3836`: when stale reclaim fills the batch, the function returns before step 3, so the bypass never runs. The comment says the wakeup will recur — but the daemon has already cleared the set, so nothing guarantees that.

Suggested shape: make the claim result distinguish *request not sent* / *server did not run the re-check* / *re-check executed*, and only treat the last one as consuming the hint. The server can echo which runtimes it actually force-re-checked.

## Non-blocking, but worth handling

`signalTaskWakeup` (`server/internal/daemon/wakeup.go:479`) drops on a full channel. That was harmless when every wakeup was an interchangeable nudge; now the wakeup is the sole carrier of a runtime-specific signal, so a drop under burst silently loses that runtime's bypass. Worth at least a debug log, ideally coalescing the runtime id into the pending set instead of dropping.

## Regressions we'd like to see

- two same-agent tasks: forced claim takes one, the **next normal claim takes the other**;
- a targeted wakeup arriving during the uncertain-claim cooldown still forces a re-check on the following cycle;
- reclaim fills the batch → the force signal is not consumed;
- wakeup channel full → the runtime id is not silently lost.

Happy to re-review as soon as these land — the design is fine, it's the signal lifecycle that needs to be airtight. Please ping if you'd like to talk through the "only consume on confirmed execution" contract before implementing it.

ADanMan:
Thanks — this is a great review, and I agree with both blockers and with making the contract "consume the hint only on confirmed execution." Before I implement, here's the concrete shape I'd like to build; please sanity-check it.

**Server echo (drives the consume contract).** The claim response gains an additive `force_rechecked_runtime_ids: []string` — the subset of the request's `force_recheck_runtime_ids` that the server actually included in the candidate scan this cycle (i.e. reached the step-4 SELECT for them). The daemon then clears the woken set for exactly those ids and re-notes every other id it drained. This closes blocker 2 uniformly:
- WS-uncertain **cooldown** branch (`wsrpc.go:333`, returns `(nil,nil)`, nothing sent): no response → confirmed set is empty → daemon re-notes the whole drained set, so the hint survives to the HTTP fallback the cooldown schedules. (The existing "don't replay after an *uncertain* claim" comment still holds only for the branch where a frame actually went out.)
- **reclaim fills the batch** (`task.go:3836`, returns before step 3): server echoes no confirmed ids → daemon re-notes → the bypass runs on the next cycle.

**Blocker 1 (stale verdict after a forced positive).** On a forced runtime whose candidate SELECT returned ≥1 candidate, I'd have the server invalidate that runtime's empty verdict (bump the version) instead of leaving the stale one live. Then the very next normal poll re-checks it rather than short-circuiting on `IsEmpty`, and if it is genuinely empty by then, step-5 `MarkEmpty` re-marks it. That fixes the "second same-agent task waits for TTL" amplification without adding a new forced-state lifetime. Alternative you floated — keep the runtime forced until its candidates drain — also works; I lean toward verdict-invalidation because it's server-local and self-repairing, but happy to do the keep-forced variant if you'd rather the daemon own that state. Which do you prefer?

**Non-blocking — `signalTaskWakeup` (`wakeup.go:479`).** I'll coalesce the runtime id into the pending woken set under the mutex even when the channel is full (so the id is never dropped), and keep a debug log for the dropped *nudge*.

**Regressions I'll add:** two same-agent tasks → forced claim takes one, next **normal** claim takes the other; a targeted wakeup during the uncertain cooldown still forces a re-check next cycle; reclaim fills the batch → the force signal is not consumed; wakeup channel full → the runtime id is not lost.

If that contract looks right (especially the `force_rechecked_runtime_ids` echo and the blocker-1 choice), I'll implement it. Happy to hop on the details if easier.

multica-eve (changes requested):
An already queued task should not remain idle for about three minutes after an available worker has been told that work exists. This is a real reliability gap worth fixing, but the current revision still does not guarantee that outcome.

**Product decision: accept the behavior change. Merge decision: request changes.** The targeted wakeup is the right independent signal; the missing piece is a bounded, compatibility-safe acknowledgement lifecycle.

Thanks for writing out the proposed follow-up. The server echo is the right general shape, and recording the runtime ID before attempting the coalesced nudge is also the right fix for channel saturation. Before implementing it, please tighten the contract in these areas:

1. **Do not acknowledge a forced-positive scan merely because the SELECT ran.** A second `EmptyClaim.Bump` can fail in the same Redis outage that lost the enqueue-side bump; `Bump` currently swallows that error. If the daemon consumes the hint anyway, the stale verdict becomes valid again when Redis recovers and the next same-agent task can still wait for the TTL. My preference is to keep the runtime forced until a scan returns zero candidates. If you use invalidation instead, it must report success and the server must acknowledge that runtime only after invalidation succeeds.

2. **Make acknowledgement support presence-aware.** A newer daemon must distinguish “a newer server explicitly returned an empty acknowledged set” from “an older server omitted the field.” Otherwise an older server or the legacy batch fallback causes the runtime ID to be resent forever. An absent capability/field should deliberately fall back to the existing TTL behavior; a present empty set should retain the pending hint.

3. **Prune pending IDs against the daemon's current runtime set.** Latest `main` now pushes runtime deletion invalidation to connected daemons. A wakeup racing runtime removal must not leave an ID permanently resident in memory and attached to every later claim request.

4. **Treat channel saturation as a blocker now.** Latest `main` allows the healthy-WebSocket safety poll to relax toward three minutes. Record/dedupe the runtime ID before the non-blocking nudge so a full channel cannot turn into the same user-visible stall.

Please rebase first: the current head `180c717e` conflicts with latest `main` in `server/internal/daemon/client.go`, `daemon.go`, and `wsrpc.go`. The acknowledgement should extend the current `claimTasksResult` / claim-poll-hint contract rather than replacing the newer scheduling metadata.

Performance-wise, the bounded design is acceptable: the field is absent in steady state, targeted wakeups already cause a batch claim, and forced runtimes join one index-backed candidate query rather than adding per-runtime queries. Please avoid the unconditional positive-result Redis pipeline, and ensure unsupported/deleted/unacknowledged IDs cannot create unbounded retries.

In addition to the four regressions already listed, please cover:

- both the enqueue-side and forced-positive Redis invalidations failing while a second same-agent task remains queued;
- a new daemon talking to an old server without permanently resending the hint;
- a runtime being removed while its hint is pending.

The current head has no implementation changes since the previous review, so the existing changes-requested decision remains in place. Ping once the rebased implementation and regressions are pushed and I will re-review it.

ADanMan:
Implemented the agreed "consume the hint only on confirmed execution" contract.

**Server echo drives it.** The batch claim now returns `force_rechecked_runtime_ids` — the subset of the request's forced runtimes that actually reached the candidate SELECT this cycle (were in the non-empty set step 4 queried). Threaded through the HTTP handler response, the daemon client (`ClaimTasks`), and the WS path (`ClaimTasksWSFirst` / `tasks.claim`). Additive and optional: an older daemon ignores it and falls back to the TTL.

**Daemon consumes only confirmed.** After a successful claim the poller re-notes every drained runtime the server did NOT echo, so the hint is consumed only for runtimes the server confirmed scanning.

- **Uncertain-cooldown branch:** the send-nothing cooldown scans nothing, so it now echoes an *empty* set (distinct from the send-a-frame branch) and the whole drained set is re-noted — a targeted wakeup arriving during the cooldown still forces its re-check on the following cycle. The "don't replay after an uncertain claim" reasoning stays scoped to the branch where a frame actually went out (which echoes the full set → re-notes nothing).
- **Reclaim fills the batch:** the server returns before step 3, so the echo omits the forced runtimes and the daemon re-notes them next cycle. The echoed set only ever contains runtimes actually scanned.

**Blocker 1 (stale verdict after a forced positive).** A forced runtime whose SELECT returned ≥1 candidate now has its empty verdict invalidated (Bump) on the forced positive, so the next normal poll re-checks it. This fixes the same-agent amplification: the batch's one-attempt-per-agent rule lets a forced claim take one of two same-agent tasks and leave the second behind — without the invalidation that second task waited for the TTL (~3 min).

**Wakeup coalescing.** `signalTaskWakeup` dropped a targeted wakeup's runtime id on a full channel. It now coalesces the id into the pending woken set under the existing mutex (debug-logged), so the runtime-specific bypass signal survives a burst; the already-queued nudge still drives the claim.

**Regressions (4).** Two same-agent queued tasks → forced claim takes one, the next normal claim takes the other; a targeted wakeup during the uncertain-claim cooldown forcing a re-check the following cycle; reclaim filling the batch leaving the force signal re-noted (not consumed); and a full wakeup channel not losing the runtime id.

**Verification.** `gofmt` clean; `go build ./...` and `go vet` clean on the touched packages; the full `internal/daemon` suite passes, including the four new daemon-side regressions. The two same-agent / cache-repair regressions in `internal/service` are DB+Redis-gated and skip in my environment (no multica Postgres or Redis available); they follow the existing `batchClaimFixture` / `newRedisTestClient` skip pattern and run under CI's pgvector service.

One judgment call worth a look: on the candidate-SELECT error path (partial success) I echo an *empty* force set so the daemon re-notes and retries, since the SELECT's outcome for the forced runtimes is unknown there. The alternative — echoing them as scanned — risks consuming the hint on a cycle that may not have repaired the verdict. Happy to flip it if you'd rather treat "reached the SELECT call" as confirmed regardless of the error.

ADanMan:
Adopted both refinements.

**1. Keep a runtime forced until a scan returns zero candidates.** You were right that acknowledging because the SELECT ran leaned on a repair write landing — `Bump` swallows its error, so the same Redis outage that lost the enqueue-side bump can lose that one too and revive the stale verdict. The server now acknowledges a forced runtime only when its scan returned zero candidates, i.e. it is genuinely idle and `MarkEmpty` has just re-armed the cache from a real SELECT. A forced scan that finds candidates is left unacknowledged, so the daemon keeps it forced and the next forced claim picks up the task the one-attempt-per-agent rule left queued. The forced-positive `Bump` is gone along with the logic it supported, so nothing depends on a Redis write succeeding.

**2. Presence-aware acknowledgement.** The acknowledged set is now `*[]string`. Present and empty means an upgraded server acknowledged nothing, so every drained runtime stays forced; absent means a pre-#7452 server that can never acknowledge, so the drained hints are consumed exactly as before #7452 and the TTL stays the outer bound rather than re-forcing forever. The server always emits the field, including when empty. The wire format is unchanged and stays additive in both directions. The send-nothing cooldown returns present-but-empty, uncertain-after-send acknowledges the full drained set, and the legacy per-runtime fallbacks report absent.

Added regressions:

- `TestClaimTasksForRuntimes_ForcedPositiveIsNotAcknowledged` — a forced scan that finds candidates is not acknowledged; a plain poll still short-circuits on the stale verdict; the next forced claim takes the second same-agent task without waiting for the TTL; once the queue drains, the forced scan comes back empty and only then acknowledges.
- `TestClaimTasksForRuntimes_ForceRecheckReArmsCacheOnZeroCandidates` — extended to assert the zero-candidate scan is the thing that acknowledges.
- `TestConsumeForceRecheckHints_PresentButEmptyKeepsAllForced` — present-but-empty keeps all drained runtimes forced.
- `TestConsumeForceRecheckHints_AbsentFieldConsumesHints` — absent field consumes the hints, no infinite re-forcing.
- `TestClaimTasks_AcknowledgementPresenceOverTheWire` — presence survives JSON decoding for `[]`, a populated array, an omitted field and an explicit `null`, and the daemon's decision follows from presence alone.
- `TestConsumeForceRecheckHints_ReNotesEverythingUnacknowledged` and the two `ClaimTasksWSFirst` uncertain-branch tests updated to the new contract.

Bohan-J (changes requested):
Re-reviewed at `b6efb738`. Real progress since the last round — thank you for working through it:

- a forced runtime now stays forced until a scan comes back with zero candidates, so the repair no longer depends on a second `EmptyClaim.Bump` surviving the same Redis window;
- `force_rechecked_runtime_ids` is presence-aware, so an old server's missing field is distinguishable from a new server's explicit empty ack;
- the cooldown and reclaim-full early returns now preserve the hint instead of consuming it;
- a wakeup dropped on a full channel now coalesces its runtime into the force set rather than vanishing.

That closes every blocker from rounds 1 and 2. Two lifecycle bugs remain, and the rebase has become a design task rather than a chore — details below. Still `Request changes`.

## 1. `signalTaskWakeup` records the runtime *after* the send attempt, so the hint can miss the claim it was meant to drive

`server/internal/daemon/wakeup.go:486`:

```go
select {
case taskWakeups <- taskWakeup{runtimeID: runtimeID}:
default:
    if runtimeID != "" {
        d.noteWokenRuntime(runtimeID)   // ← too late
    }
}
```

The `default` branch is only reached because a nudge is already queued. But between the failed send and `noteWokenRuntime`, the poller can consume that queued nudge *and* run `drainWokenRuntimes()`. The id then lands in a set that nobody is going to drain, with no nudge left to trigger a claim — so it waits for the next scheduled poll instead.

Recording the id **before** the `select` fixes it and is strictly safe: if a drain races in between, the drain picks the id up and the subsequent nudge is merely redundant. A concurrency test over 100k iterations fails on the current ordering and passes with the note moved ahead of the `select`.

## 2. A deleted runtime's force id is re-sent forever

`consumeForceRecheckHints` (`server/internal/daemon/daemon.go:4941`) re-notes every drained id that is not in the ack list, and `drainWokenRuntimes` (`:4959`) never intersects the set with the daemon's live runtimes. The server filters `force_recheck_runtime_ids` down to the runtimes the daemon is authorized for, so a runtime deleted after being woken can never appear in the ack — which means it is re-noted on every cycle, permanently.

This is not hypothetical now that `MUL-7002` (#7969) pushes runtime-deletion invalidation to live daemons. Driving the real `runBatchPoller` against a stub server, three consecutive claims go out as `runtime_ids=[live]`, `force_recheck_runtime_ids=[removed]`.

Please filter against the current runtime set both when draining and when re-noting, and cover the case where a deletion interleaves with an in-flight request.

## 3. The rebase is a design change, not a mechanical merge

`main` has moved under this branch. Merging `b6efb738` into `2ed2432a` conflicts in four files: `server/internal/daemon/client.go`, `daemon.go`, `wsrpc.go`, and `server/internal/handler/daemon.go`. Two `main` commits matter beyond the textual conflicts:

**#7983 (relax healthy-WS claim polling to 3 minutes)** is the important one. This branch is built on the old poller, which always slept `PollInterval` (30s) — it has no reference to `WSClaimPollInterval` anywhere. On `main` the sleep is now `taskClaimPollInterval(claimResult)`, which returns up to a jittered `DefaultWSClaimPollInterval` (3 minutes) whenever the claim went over a healthy WS with poll hints negotiated, shortened only by `NextDeferredTaskAfterMillis`. It knows nothing about a pending force hint.

So after a naive rebase: a forced claim finds candidates → dispatches one task → the runtime deliberately stays forced (per this PR's own design) → free slots remain → the poller sleeps up to ~3 minutes. The second same-agent task waits out almost exactly the stall #7452 is about. The rebase must therefore teach `taskClaimPollInterval` that a non-empty force set means "do not take the long safety interval" — otherwise the fix is undone by the branch it lands on.

**#7969 (runtime deletion invalidation)** is the same commit that makes item 2 reachable in practice.

Please also fold the ack into `main`'s existing shape rather than widening the signature: `ClaimTasks` currently returns `([]*Task, *[]string, error)` on this branch, whereas `main` routes optional scheduling metadata through the `claimTasksResult` struct (`client.go:276`) behind the `claim-poll-hints-v1` capability. `ForceRecheckedRuntimeIDs` belongs on that struct, negotiated the same way, instead of as a parallel third return value.

## Regressions we'd like to see with the next push

- concurrent drain vs. a wakeup dropped on a full channel — the id is never stranded without a pending nudge;
- a runtime deleted while it has a pending force hint is dropped from both the set and the request, including when the deletion races an in-flight claim;
- with a force hint still pending, the poller does **not** fall into the ~3-minute WS safety interval;
- the round-1/2 regressions still passing after the rebase.

Once `main` is merged in, please also re-run CI — the green checks on record are from `180c717e` and don't carry over to the current head.

The design is in good shape and the remaining items are all lifecycle/integration, not direction. Happy to look again as soon as the rebase lands.

ADanMan:
Pushed the rebase + the two remaining lifecycle fixes.

1. **Record-before-select.** `signalTaskWakeup` now records the woken runtime in the force-recheck set as its first action, before the channel send is attempted. A drain that races in between just picks the id up and the later nudge is redundant, so a queued nudge can no longer be consumed and drained between a failed send and the note, stranding the id.

2. **Live-runtime filtering.** The force set is intersected with the daemon's live runtime set (`filterLiveRuntimes`) when draining before the claim, on the error re-note, and at consume — re-derived at each use, so a runtime deleted after being woken (reachable via #7969) is dropped from both the set and the request instead of being re-sent forever, including when the deletion races an in-flight claim.

3. **Rebase onto #7983/#7969.** `main` merged in. The `force_rechecked_runtime_ids` ack now rides on `main`'s `claimTasksResult` struct (negotiated the same way as the poll hints), not a parallel third return value; presence-awareness (absent vs explicit-empty) is preserved. `taskClaimPollInterval` now returns the normal poll whenever a force hint is still pending, so a forced claim that dispatches one task and keeps the runtime forced does not fall into the ~3-minute healthy-WS safety interval and re-open the exact stall #7452 targets.

Regressions added (`force_recheck_rebase_test.go`): record-before-send, a 100k concurrent-drain stress (never stranded, race-clean), a deleted runtime dropped before the claim and when the deletion races an in-flight claim, and a pending force hint keeping the normal cadence. Round-1/2 regressions still pass after the merge.

Local verification: `gofmt -l`, `go build ./...`, `go vet`, and `go test ./internal/daemon ./internal/service ./internal/handler` all clean; the daemon-level force-recheck tests run, and the service-level ones skip locally as they are DB+Redis-gated. CI will re-run on this head.

PR #7832, opened 2026-08-31
fix(telegram): prevent duplicate direct-room replies

### LINKED ISSUE #7750: Telegram direct-room replies are delivered twice
## Summary

The Telegram direct-room integration delivers each assistant reply twice.

## Observed

- Date/time: 2026-08-30 around 00:38 (Asia/Bangkok)
- The user sent `so…cannot fix?` once.
- One assistant response was produced, but Telegram rendered two identical messages with the same 00:38 timestamp.
- The user reports this happens for every assistant reply.

Screenshot showing the duplicate messages

## Expected

Each completed assistant response should produce exactly one Telegram message.

## Actual

The same response is delivered twice. The integration should deduplicate delivery or make retries idempotent.

leonzone:
## What does this PR do?

Prevents one completed Telegram direct-room turn from producing two visible replies.

The failure had two reproducible paths:

- repeated chat:done events were queued more than once for the same task;
- an ambiguous editMessageText failure fell back to a new sendMessage even though the streamed message could already contain the final text.

The fix deduplicates terminal delivery by task, ignores late text frames after terminal delivery starts, retries edits against the original Telegram message, and persists outbound Telegram message IDs for restart-safe replay checks.

## Related Issue

Closes #7750

## Type of Change

## Changes Made

- Add bounded in-process terminal-task deduplication and late-frame suppression.
- Make ambiguous and HTML-format edit failures retry the original Telegram message; only a confirmed missing edit target may fall back to a new message.
- Persist task-reply provenance in chat_message and channel_outbound_message.
- Add regression coverage for duplicate completion events, ambiguous edits, HTML fallback, late frames, durable replay suppression, and safe missing-target fallback.

## How to Test

1. Run go test ./internal/integrations/telegram -count=1.
2. Run go test -race ./internal/integrations/telegram -count=1.
3. Run make test.
4. Connect a local Telegram bot and send three direct messages that request distinct exact replies. Verify each turn produces one stable Telegram bubble and one outbound ledger row.

Observed locally: TG7750-A, TG7750-B, and TG7750-C each produced exactly one stable reply bubble, one assistant Telegram message ID, and one task_reply ledger row.

make check reaches the TypeScript unit-test stage but is currently blocked by nine unrelated local @multica/core failures where the test environment exposes localStorage without setItem, removeItem, or clear. The Telegram suite and full Go race suite pass independently.

## Checklist

- [x] I have run the relevant tests locally and they pass

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Reproduced both duplicate-delivery paths against a local fake Telegram Bot API, converted them into regression tests, implemented task-level and provider-message-level idempotency, then completed three real Telegram direct-message acceptance turns against a local Multica environment.

### DIFF
--- a/server/internal/integrations/telegram/outbound.go
+++ b/server/internal/integrations/telegram/outbound.go
@@ -58,6 +58,7 @@ type Outbound struct {
 	terminalInFlight         int
 	queuedTerminalReplyCount int
 	queuedTerminalReplyBytes int
+	terminalTasks            map[string]time.Time
 	workerOnce               sync.Once
 	workerWG                 sync.WaitGroup
 	terminalWorkerWG         sync.WaitGroup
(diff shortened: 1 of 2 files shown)


multica-eve:
关联的 Issue #7750 已由 PR #8545 合入并关闭。#8545 以持久化的跨副本交付所有权覆盖了本 PR 处理的重复回复场景，因此关闭本 PR。

PR #7835, opened 2026-08-31
fix(issues): wake parent when child enters review

valo:
## Summary
- Treat a child transition to `in_review` as a completed stage transition.
- Wake the parent assignee once so its coordinator can evaluate and promote the next ready backlog stage.
- Keep `in_review` → `done` idempotent, preventing a duplicate parent wake.

## Validation
- `go test ./internal/handler -run 'TestChild(InReviewNotifiesParent|InReviewThenDoneDoesNotRefire|CompletedStatusIncludesReview)$' -count=1 -v`
- `git diff --check`

`make test` exercised the handler suite successfully, including the new database-backed regression cases. The aggregate command has unrelated CLI test failures caused by the daemon-managed task marker in this environment.

### DIFF
--- a/server/internal/handler/issue_child_done.go
+++ b/server/internal/handler/issue_child_done.go
@@ -350,15 +350,22 @@ func (h *Handler) postChildDoneComment(ctx context.Context, parent, completed db
 	h.dispatchParentAssigneeTrigger(ctx, parent, comment)
 }
 
-// isTerminalChildStatus reports whether a child issue status counts as
-// "finished" for stage-barrier purposes. Cancelled counts as terminal: a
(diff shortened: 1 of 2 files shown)


valo:
This is an interesting one. The workflow I am trying to implement is to have a parent issue with a bunch of sub-issues. The agents work on the sub-issues and stack the PRs on top of each other. Once a sub-issue reaches `in_review` state, the next sub-tasks should be kicked off to build on top of it. Not sure how to implement this workflow without a change like this.

valo:
I'll close this one. The change is too big imo. I'll tweak my workflow to match the expected behavior. 

I think a much better change will be to allow the parent issue do actions when any of the sub tasks change status

PR #7837, opened 2026-08-31
fix(daemon): restore issue scope for manual reruns

### LINKED ISSUE #7826: [Bug]: Retrying a failed task loses the project's local_directory resource — runs in an empty workdir and reports completed (v0.4.36 self-host)
### Environment

- Multica self-host v0.4.36 (GHCR images via `docker-compose.selfhost.yml`), backend+web+postgres on the same host
- `multica` CLI / daemon 0.4.36, linux/amd64, running in an LXC container (Debian 12)
- Provider: `claude` (Claude Code 2.1.251)
- Project resource: `local_directory` in `worktree` execution mode

### Timeline (from daemon logs)

1. An issue was created in a project whose only resource was `--type local_directory --local-path /root/progetto-prova`. The first run correctly picked it up:
   ```
   INF picked task ... task=01a05721 agent=Mika provider=claude
   INF local_directory: worktree mode, skipping path mutex ... local_directory=/root/progetto-prova
   ```
   That run failed for an unrelated, legitimate reason (`Claude Code refuses bypassPermissions under root/sudo privileges`), and the issue went back to *Todo* as expected.

2. Between the failure and the retry, the resource was updated via CLI (the directory had been moved to a non-root user):
   ```
   multica project resource update <project-id> <resource-id> --local-path /home/agente/progetto-prova
   ```
   `project resource list` confirmed the new path.

3. The user pressed **Retry** on the failed run in the web UI. The retried task ran **without the `local_directory` resource at all** — no `local_directory:` line in its context, a generic scratch workdir, and it **reported `completed`**:
   ```
   INF picked task ... task=01a05727 issue="" agent=Mika provider=claude
   INF starting agent ... task=01a05727 provider=claude workdir=.../task-bea4a9d42bce/workdir
   INF agent finished ... task=01a05727 status=completed duration=11s tools=2
   ```
   The workdir contained only `CLAUDE.md` and `description.md`; the agent's file changes landed nowhere and no error was surfaced.

4. A **freshly created issue** on the same project immediately picked up the updated resource and completed correctly:
   ```
   INF local_directory: worktree mode, skipping path mutex ... local_directory=/home/agente/progetto-prova
   ```

### Expected

A retried task should re-resolve the project's resources at retry time (or at minimum fail loudly if its captured resource snapshot is no longer valid).

### Actual

The retried task runs with no `local_directory` in its context and reports `completed` — a **silent false success**: status is green, but the work was executed in an empty scratch directory.

### Notes

- I could not fully isolate whether the retry *dropped* the resource or carried a *stale snapshot* pointing at the old (now missing) path and silently fell back to a scratch workdir; either way the user-visible outcome is a completed run that did nothing to the project.
- Possibly related but distinct from #6315 (there the task fails with a misleading error; here it *succeeds*).
- Workaround: create a new issue instead of retrying after resources changed.

zengxinyueooo:
## What does this PR do?

Fixes the failure mode reported in #7826, where a manually retried task can reach `ClaimAgentTask` without an `issue_id`.

Without issue scope, the claim path treats the rerun as an issue-less task. As a result, it does not resolve the project's current resources and the daemon starts the task in a generic scratch workdir instead of the configured `local_directory`.

Thinking path:

- Project resources are resolved during task claim from the task's issue and project scope.
- The daemon log in #7826 shows the retried task being claimed with `issue=""`.
- Manual reruns retain `rerun_of_task_id`, which provides durable lineage back to the source task.
- When a manual rerun is missing issue scope, this change restores the issue association from its same-agent source task before the claim response is built.
- The existing claim path can then resolve the project's current resources, including resource path updates made between attempts.
- The recovery is intentionally limited to manual reruns with missing issue scope. Existing issue-bound tasks, chat tasks, autopilot tasks, quick-create tasks, and cross-agent reruns keep their current behavior.

## Related Issue

Closes #7826

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add a conservative predicate for identifying manual reruns whose issue scope can be restored.
- Restore and persist the source task's issue association before resolving the claim response.
- Re-read the task after linking the issue to verify that the association was persisted.
- Add boundary coverage for issue-bound, chat, autopilot, quick-create, and cross-agent reruns.
- Add a database-backed regression test proving that a rerun uses the project's updated `local_directory` resource while retaining its prior workdir lineage.

## How to Test

From `server/`, with PostgreSQL available and migrations applied:

1. Run the focused scope and claim regression tests:

   ```bash
   go test ./internal/handler \
     -run '^(TestRestorableRerunIssueScope|TestClaimTask_ManualRetryReusesWorkdir)$' \
     -count=1

### DIFF
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -1985,6 +1985,71 @@ func rerunSourceMatchesTaskScope(task, source db.AgentTaskQueue) bool {
 	return false
 }
 
+func restorableRerunIssueScope(task, source db.AgentTaskQueue) (pgtype.UUID, bool) {
+	if task.IssueID.Valid || task.ChatSessionID.Valid || task.AutopilotRunID.Valid || !task.RerunOfTaskID.Valid {
+		return pgtype.UUID{}, false
+	}
+	var quickCreate struct {
+		Type string `json:"type"`
+	}
+	if json.Unmarshal(task.Context, &quickCreate) == nil && quickCreate.Type == service.QuickCreateContextType {
+		return pgtype.UUID{}, false
+	}
+	if !source.IssueID.Valid || source.AgentID != task.AgentID {
+		return pgtype.UUID{}, false
+	}
+	return source.IssueID, true
+}
+
+// restoreMissingRerunIssueScope repairs the one manual-rerun state that must
+// never reach the daemon as a generic scratch task: a rerun row whose durable
+// lineage points at an issue task but whose own issue_id is NULL. The source
+// task is server-authored lineage and the agent match prevents a malformed row
+// from borrowing another agent's issue. Persist the repair before assembling
+// the claim so project context is resolved live from the issue's current
+// project resources and later progress/completion stays attached to the issue.
+//
+// Issue-less quick-create reruns are intentionally unchanged: their source has
+// no issue_id, so they continue through the quick-create context path below.
+func (h *Handler) restoreMissingRerunIssueScope(ctx context.Context, task *db.AgentTaskQueue) error {
+	if !task.RerunOfTaskID.Valid || task.IssueID.Valid || task.ChatSessionID.Valid || task.AutopilotRunID.Valid {
+		return nil
+	}
+
+	source, err := h.Queries.GetAgentTask(ctx, task.RerunOfTaskID)
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return nil
+		}
+		return fmt.Errorf("load rerun source while restoring issue scope: %w", err)
+	}
+	issueID, ok := restorableRerunIssueScope(*task, source)
+	if !ok {
+		return nil
+	}
+
+	if err := h.Queries.LinkTaskToIssue(ctx, db.LinkTaskToIssueParams{
+		ID:      task.ID,
+		IssueID: issueID,
+	}); err != nil {
+		return fmt.Errorf("restore rerun issue scope: %w", err)
+	}
+	repaired, err := h.Queries.GetAgentTask(ctx, task.ID)
+	if err != nil {
+		return fmt.Errorf("verify restored rerun issue scope: %w", err)
+	}
+	if !repaired.IssueID.Valid || repaired.IssueID != issueID {
+		return fmt.Errorf("verify restored rerun issue scope: issue_id was not persisted")
+	}
+	task.IssueID = repaired.IssueID
+	slog.Warn("daemon claim: restored missing issue scope from rerun source",
+		"task_id", uuidToString(task.ID),
+		"source_task_id", uuidToString(source.ID),
+		"issue_id", uuidToString(issueID),
+	)
+	return nil
+}
+
 func claimResponseAgentIdentityMatches(resp AgentTaskResponse) bool {
 	return resp.AgentID != "" && resp.Agent != nil && resp.Agent.ID == resp.AgentID
 }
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for digging into this. The conservative guard set, the same-agent check, and the write-then-verify-read are all nicely done, and the core reasoning — that project resources are resolved live at claim time, so restoring issue scope is enough to pick up an updated `local_path` — matches what the code actually does.

I have one concern I think is blocking, plus a few smaller notes.

### The reported run looks like it is excluded by design

`restorableRerunIssueScope` returns false when the task's context is a quick-create context, and the doc comment states that quick-create reruns are intentionally unchanged.

But the evidence in #7826 points at exactly that case. The reporter says the retried run's workdir contained only `CLAUDE.md` and `description.md`. `description.md` is something the agent writes only under the quick-create prompt — `buildQuickCreatePrompt` is what instructs it to write `./description.md` and pass `--description-file`. The daemon selects that prompt on `task.QuickCreatePrompt != ""`, and that field is populated only inside the claim path's quick-create branch, which is gated on `task.Context != nil && !task.IssueID.Valid && !task.ChatSessionID.Valid && !task.AutopilotRunID.Valid` together with `qc.Type == QuickCreateContextType`.

So the failing task most likely did carry a quick-create context, which is the one shape this PR deliberately skips. If that is right, the reported scenario would behave identically after this change and `Closes #7826` would not hold. Is there something that rules that out?

### I could not find a writer that produces the state being repaired

Three queries write `rerun_of_task_id`, and none of them appears able to produce "`rerun_of_task_id` set + `issue_id` NULL + non-quick-create context + issue-bound same-agent source":

- `CreateAgentTask` — a manual rerun reaches it through `enqueueRerunTask` → `enqueueIssueTaskWithCommentPlan` / `enqueueMentionTaskWithCommentPlan`, both of which enqueue against an issue, so `issue_id` is always set. `RerunIssue` additionally rejects a source task whose `issue_id` does not match the issue.
- `CreateDeferredChannelIssueTask` — `issue_id` is a positional parameter, not a nullable arg.
- `CreateManualQuickCreateRetryTask` — its `WHERE` requires `p.issue_id IS NULL`, so the source never has an issue to lend, and it copies `p.context`, so the quick-create exclusion applies anyway.

Relatedly, the database regression test constructs the broken row with a raw `INSERT ... rerun_of_task_id` rather than producing it through a product path. That demonstrates the repair logic works, but not that the state it repairs is reachable. A test that arrives at the state through the rerun API would settle both this and the point above.

### Smaller notes

- `LinkTaskToIssue`'s doc comment says its purpose is attaching the issue a quick-create task produced once the agent has finished, and that "only quick-create tasks land here unset". After this change that is no longer accurate — worth updating so the next reader does not rely on it.
- The claim path now performs a write plus a verifying re-read on a hot path, and neither `GetAgentTask(task.RerunOfTaskID)` nor `LinkTaskToIssue` is workspace-scoped; the same-agent check is the only tenant guard. That is probably sound since agents are workspace-scoped, but it runs against the direction of #7405, which deliberately removed unscoped reads from this path.

### Separately

Whatever the outcome here, the silent half of #7826 is untouched. When a project's `local_directory` is pinned to a different `daemon_id` — plausible for the reporter after moving the daemon from root to a non-root user — `findLocalDirectoryAssignment` skips it silently, and `acquireLocalDirectoryLockIfNeeded` returns early without logging anything at all. The run still finishes green in a scratch workdir. That seems worth addressing on its own, independent of the issue-scope question.

PR #7842, opened 2026-08-31
MUL-6903: fix(processtree): terminate escaped Linux descendants

### LINKED ISSUE #7615: [Bug]: Unix process-tree ownership cannot reach a descendant that leaves the process group (setsid) — measured instance
**Deployment type:** either — this is daemon-side
**Area:** Agents and runtimes
**Multica:** `main` @ `09a2410e8` (i.e. with #7531 / MUL-6658 in)
**Platform:** Unix (Linux/macOS). Windows is **not** affected — see below.

## What happened

#7531 made whole-tree process ownership structural, so that cancelling a task takes down the
CLI's descendants rather than leaving the real agent running (the incident in #7522: an orphan
kept working for 40 minutes and posted comments as the member). On Windows it does that with a
Job Object, which owns the tree irrespective of process groups. On Unix it does it with a
process group and a group-wide `kill(-pgid)`.

A descendant that leaves the group is therefore out of reach on Unix, and that is not
hypothetical — measured on a real OpenClaw host while verifying #6275:

- `openclaw` and its `openclaw-config` helper hold **the same stdout pipe**
  (both `pipe:[135202]`; the same shared inode when stdout is redirected to a file), so the
  helper is a genuine descendant of the invocation;
- but `openclaw-config` runs with its **own PGID and SID**.

`kill(-pgid)` on the leader's group never reaches it. Neither does
`signalProcessGroup`/`releaseProcessGroup` in `server/pkg/agent`, nor
`internal/daemon/processtree`'s `finish`, since all of them address the same group.

## Why this is worth tracking rather than assumed covered

The guarantee as written reads stronger than it is. Examples:

- `TestOnlyOwnedProcessTreesAreStarted` explains that `startOwnedProcessTree` exists so that
  "cancelling or timing out" cannot leave "the CLI's descendants running".
- `internal/daemon/processtree`'s doc says it "runs bounded helper commands whose descendants
  must not survive cancellation".

Both are true for a descendant that stays in the group, which is the common case, and false for
one that calls `setsid`. On the platform where #7531's motivating incident happened (Windows)
the Job Object closes it; on Unix it does not.

## Steps to reproduce

The shape is reproducible without OpenClaw:

1. Write a stub CLI that prints a line and then forks a helper that calls `setsid` while keeping
   the inherited stdout open, e.g.

   ```sh
   #!/bin/sh
   echo ready
   setsid sh -c 'sleep 300' &      # inherits stdout, leaves the group
   exit 0
   ```

2. Run it through any of this repo's owned-start helpers (`runOwned` / `outputOwned` in
   `server/pkg/agent/launch.go`, or `processtree.Output`).
3. After the call returns, the `sleep 300` is still alive and reparented to init.
   `pgrep -f 'sleep 300'` after the run is the check that matters — an assertion cannot see an
   orphan the test never named.

## Impact

Bounded, but real: on Unix, a runtime CLI (or any helper it spawns) that dissociates itself can
outlive task cancellation with the environment it was given. For a read-only probe that is a
leak; for an agent process it is the #7522 shape, which is why this seems worth an issue rather
than a comment on a merged PR.

Not urgent on the evidence available: the only descendant measured doing this is OpenClaw's
config helper, which exits with its parent on 2026.7.1-2, so no orphan actually accumulated in
the E2E runs. No agent CLI is known to do it today.

## Possible directions

1. **Scope the claims.** Cheapest, and worth doing regardless: say "descendants that remain in
   the process group" in the guard tests and the `processtree` doc, so the next reader does not
   inherit a guarantee the code cannot make.
2. **Reap by descent instead of by group on Unix.** Walk the process tree (`/proc/<pid>/stat`'s
   PPID on Linux, `sysctl KERN_PROC_ALL` on macOS) at kill time and signal each pid. This
   catches a `setsid`'d child that is still parented to the CLI, and is the closest Unix analogue
   to what a Job Object gives for free. It does not catch a descendant that also re-parents
   itself, which nothing short of cgroups/PID namespaces would.
3. **cgroup v2 on Linux** would give exact tree ownership, but only there and only with a
   writable cgroup — probably not worth it for helper probes.

Context: found while verifying #6275, which documents the same limitation on
`RunCollect`'s guarantee 2 and does not attempt to fix it, since pipe ownership rather than the
kill is what makes that helper's deadline enforceable.

makesomethingshit:
## What does this PR do?

Linux only, best-effort procfs descendant retention; not a general Unix Job Object equivalent.

This fixes one Linux cleanup gap when a descendant calls `setsid()` and escapes the process group managed by `processtree`. While the root is still alive and `/proc` is readable, the controller remembers observed descendants by PID plus start time and signals them during interrupt, stop, and finish cleanup. Existing process-group signaling remains the portable fallback, and non-Linux Unix behavior is unchanged.

## Related Issues

Closes #7615

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [x] Tests

## Changes Made

- Track Linux descendants observed through `/proc`, including descendants that leave the original process group with `setsid()`.
- Guard descendant signaling with `/proc` start-time identity checks to avoid acting on a reused PID.
- Treat reaped descendant zombies as inactive during cleanup.
- Preserve the existing process-group-only behavior on non-Linux Unix platforms.
- Document the platform-specific, best-effort ownership contract.
- Turn the Linux reproducer into a regression test that requires the observed escaped descendant to terminate.

## How to Test

Run on Linux:

```bash
cd server
go test -race ./internal/daemon/processtree -run '^TestProcesstreeRunSetsidEscape$' -count=1 -v
go test -race ./internal/daemon/processtree -count=1
cd ..
bash scripts/test-go.sh --race
```

## Screenshots (if applicable)

N/A — backend process-lifecycle change.

## Known Limits and Risks

- If `/proc` is unavailable or unreadable, cleanup falls back to the existing process-group behavior.
- A descendant that reparents or exits before it is observed may not be retained.
- PID namespaces can hide or remap descendants and are outside this best-effort contract.
- PID reuse is guarded only for observed processes by matching the recorded start time before signaling.
- This does not provide complete process-tree ownership or a general Unix equivalent of Windows Job Objects.
- Non-Linux Unix behavior is intentionally unchanged.

## Checklist

- [x] I have thought through the edge cases and failure modes
- [ ] I have run tests locally and they pass — the local host is Windows; the Linux-only regression is covered by CI
- [x] I have added tests that prove my fix is effective
- [x] If my change affects the UI, I have included screenshots — N/A, backend process-lifecycle change
- [x] I have made corresponding changes to the documentation
- [x] My changes do not introduce a new runtime, tool, or UI tab — N/A
- [x] User-facing text uses Chinese characters with a Traditional/Simplified conversion helper — N/A, no user-facing text added
- [x] I have documented known limits and risks
- [x] I have performed a self-review of my code

### AI tool used

Multica Agent / ChatGPT

### Prompt / approach

Started from a Linux `setsid()` reproducer, traced the Unix controller lifecycle, then implemented the narrow Linux fix: best-effort retention of procfs-observed descendants by stable process identity while preserving process-group cleanup as the cross-platform fallback.

### DIFF
--- a/server/internal/daemon/processtree/controller_unix.go
+++ b/server/internal/daemon/processtree/controller_unix.go
@@ -13,23 +13,40 @@ import (
 
 const processTreeFinishTimeout = 5 * time.Second
 
-type controller struct{}
+type ownedDescendant struct {
+	pid      int
+	identity uint64
+}
+
+type controller struct {
+	descendants map[int]ownedDescendant
+}
 
 func newController(cmd *exec.Cmd) (*controller, error) {
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
 	cmd.SysProcAttr.Setpgid = true
-	return &controller{}, nil
+	return &controller{descendants: make(map[int]ownedDescendant)}, nil
 }
 
 func (*controller) attach(_ *exec.Cmd) error { return nil }
 
-func (*controller) interrupt(cmd *exec.Cmd) error {
-	if cmd.Process == nil {
-		return os.ErrProcessDone
+func (c *controller) rememberDescendants(rootPID int) {
+	refs, err := discoverOwnedDescendants(rootPID)
+	if err != nil {
+		// Process groups remain the portable Unix fallback. Descendant discovery
+		// is an additional Linux containment layer and must not make cancellation
+		// fail on hosts where procfs is unavailable or restricted.
+		return
 	}
-	if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM); err != nil {
+	for _, ref := range refs {
+		c.descendants[ref.pid] = ref
+	}
+}
+
+func signalProcessGroup(pid int, sig syscall.Signal) error {
+	if err := syscall.Kill(-pid, sig); err != nil {
 		if errors.Is(err, syscall.ESRCH) {
 			return nil
 		}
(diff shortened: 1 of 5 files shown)


multica-eve:
Thanks for the contribution. We’re closing this PR because it does not improve the task-cancellation guarantee members rely on: the measured agent workload can still remain active after cancellation, while this change only protects a repository-maintenance edge case that has not been observed in production.

The production evidence in #7615 is from the OpenClaw/agent runtime path, but this diff only changes the daemon process-tree helper used by Git GC and repository-cache commands. It also does not cover the normal-exit reproducer, and checking a PID’s start time separately from delivering the signal leaves a PID-reuse race. The permanent Linux-specific scanning and signaling mechanism therefore has no demonstrated beneficiary sufficient to justify its maintenance and safety cost.

A future proposal would need to target the shared agent-runtime ownership path, cover both cancellation and normal exit, and use a stable process handle for signaling. That would be a materially different design and should be reviewed on its own evidence.

PR #7844, opened 2026-08-31
MUL-6904: fix: make task transcript delivery loss-detecting

golb:
Supersedes #7649. This branch reapplies the original two commits to the current `main`. The migration conflict is resolved by renumbering `441/442` to the next free versions, `444/445`. It also updates the CodeArts backend added after #7649 was opened to use the same context-aware message delivery instead of the removed `trySend` helper.

## Summary

- replace silent agent message drops with context-aware backpressure and task-scoped abandonment detection across every backend
- make transcript batch delivery retry-safe through an advertised server capability, `(task_id, seq)` uniqueness, atomic idempotent replay, and conflicting-payload rejection
- verify a contiguous transcript high-watermark in the same task-row transaction that completes or fails a task
- drain buffered cancellation tails before returning; fail closed when a stream never closes or delivery is truncated

## Why

The daemon previously cleared each in-memory batch before a one-shot HTTP request. A transient request or insert failure could therefore leave a task terminal while its persisted transcript was incomplete. Agent adapters could also silently drop messages whenever their 256-entry channel filled.

This change fixes the shared delivery boundary instead of adding adapter-specific guards. New daemons retry only when the claimed task advertises `transcript_batch_replay_safe`; old servers retain the existing one-shot behavior, avoiding duplicate rows during mixed-version operation.

## Verification

- pinned `sqlc/sqlc:1.31.1 generate`
- DB-backed replay, conflicting replay rollback, invalid sequence, cancelled final drain, complete/fail receipt, exact terminal replay, and late-write seal tests
- daemon retry, drain, cancellation-tail, and never-closing-stream regressions
- agent backpressure and late-reader regressions across affected stream wrappers
- migration cleanup-map contract tests
- targeted daemon/agent tests under `-race`
- compile-only check and `go vet` for all changed packages
- `git diff --check`

## Upgrade notes

Migration 444 deliberately fails closed if historical duplicates already exist. Before production migration, check:

```sql
SELECT task_id, seq, COUNT(*)
FROM task_message
GROUP BY task_id, seq
HAVING COUNT(*) > 1;
```

If rows are returned, inspect and reconcile them before retrying the migration; do not delete them automatically.

The official Helm backend uses `Recreate`, which avoids a mixed server fleet. Custom multi-replica deployments should roll the server capability fleet-wide before relying on daemon retries.

## Scope boundary

The receipt proves persistence of every `agent.Message` formed during the current process lifetime. Crash-durable daemon restart replay, raw provider-byte preservation, and first-class failed/cancelled receipt storage remain separate follow-up work.

### DIFF
--- a/server/cmd/migrate/main.go
+++ b/server/cmd/migrate/main.go
@@ -298,6 +298,7 @@ var concurrentIndexCleanups = map[string]string{
 	"443_issue_project_status_index":                            "idx_issue_project_status",
 	"445_comment_delegated_failure_unsettled_index":             "idx_comment_delegated_failure_unsettled",
 	"446_issue_properties_bigm_index":                           "idx_issue_properties_bigm",
+	"449_task_message_task_seq_unique_index":                    "task_message_task_id_seq_uidx",
 }
 
 // concurrentDownIndexCleanups covers every migration whose down direction
@@ -321,6 +322,7 @@ var concurrentDownIndexCleanups = map[string]string{
 	"375_drop_issue_last_activity_index":                    "idx_issue_workspace_last_activity",
 	"391_drop_agent_task_queue_dispatched_prepare_index":    "idx_agent_task_queue_dispatched_prepare",
 	"437_drop_agent_runtime_last_seen_at_index":             "idx_agent_runtime_last_seen_at",
+	"450_drop_task_message_task_seq_index":                  "idx_task_message_task_id_seq",
 }
 
 var preMigrationHooks = func() map[string]preMigrationHook {
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -1015,10 +1015,12 @@ var defaultTerminalRetrySchedule = []time.Duration{
 	64 * time.Second,
 }
 
-// taskMessageRetrySchedule is deliberately shorter than the terminal callback
-// schedule: transcript batches are emitted throughout a run, so a long outage
-// must fail the task instead of stacking minutes of blocked flushes. Replays
+// taskMessageRetrySchedule spends at most 1.75s sleeping inside the existing
+// 5s transcript-delivery deadline. That leaves 3.25s for four fast-failing
+// requests while bounding every flush at the same 5s latency ceiling. Replays
 // are safe because the server de-duplicates each message by (task_id, seq).
+const taskMessageDeliveryTimeout = 5 * time.Second
+
 var taskMessageRetrySchedule = []time.Duration{
 	250 * time.Millisecond,
 	500 * time.Millisecond,
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -494,7 +494,17 @@ func (c *Client) ReportTaskMessages(ctx context.Context, taskID string, messages
 	}, nil)
 }
 
+func (c *Client) ReportTaskMessagesWithRetry(ctx context.Context, taskID string, messages []TaskMessageData) error {
+	return c.postJSONWithRetry(ctx, fmt.Sprintf("/api/daemon/tasks/%s/messages", taskID), map[string]any{
+		"messages": messages,
+	}, nil, taskMessageRetrySchedule)
+}
+
 func (c *Client) CompleteTask(ctx context.Context, taskID, output, branchName, sessionID, workDir string, sessionRolloutMissing bool, retiredSessionID, durableWorkDir string) error {
+	return c.CompleteTaskWithTranscript(ctx, taskID, output, branchName, sessionID, workDir, sessionRolloutMissing, retiredSessionID, durableWorkDir, nil)
+}
+
+func (c *Client) CompleteTaskWithTranscript(ctx context.Context, taskID, output, branchName, sessionID, workDir string, sessionRolloutMissing bool, retiredSessionID, durableWorkDir string, expectedMessageSeq *int32) error {
 	body := map[string]any{"output": output}
 	if branchName != "" {
 		body["branch_name"] = branchName
@@ -514,6 +524,9 @@ func (c *Client) CompleteTask(ctx context.Context, taskID, output, branchName, s
 	if retiredSessionID != "" {
 		body["retired_session_id"] = retiredSessionID
 	}
+	if expectedMessageSeq != nil {
+		body["expected_message_seq"] = *expectedMessageSeq
+	}
 	return c.postJSONWithRetry(ctx, fmt.Sprintf("/api/daemon/tasks/%s/complete", taskID), body, nil, defaultTerminalRetrySchedule)
 }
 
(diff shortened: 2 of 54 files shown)


Bohan-J (changes requested):
Thanks for this — the product direction is right, and I want to accept it. Task transcripts that silently lose a segment while the task still reports success are a data-integrity problem: a member reads an incomplete record as complete fact. Making delivery loss-detecting, and refusing to settle a terminal state until completeness is confirmed, is the correct shape for that.

The implementation also gets several structural things right: it extends the existing batch-ingest entry point and the terminal transaction instead of building a parallel channel, and `(task_id, seq)` uniqueness + same-content idempotent replay + conflicting-content rejection + the task row lock + the contiguous-watermark check form a real end-to-end loop. The capability field keeps one-shot sends working when a new runtime talks to an older server.

Four things block merge.

### 1. The cancellation path still sends a positive "flush complete" ack for an unknown transcript

`server/internal/daemon/daemon.go:5178` — the `cancelledByPoll` branch discards nearly every error from `runner.run` and then unconditionally calls `AckTaskCancelled`. Its comment states "runner.run has returned, so the transcript flush is complete", but this PR makes that no longer true: `executeAndDrain` now returns transcript-completeness errors — `transcript delivery failed` (`:8681`), `transcript drain was force-stopped before the message stream closed` (`:8683`), `transcript delivery abandoned during agent cancellation` (`:8687`). When a server-side cancel lands at just the wrong moment, exactly the signal this PR introduces gets swallowed, and we then send an ack whose semantics are "the transcript is fully flushed".

Please separate integrity errors from ordinary cancellation noise: at minimum, do not send a flush-complete ack when completeness is unknown, or extend the cancel ack with a verifiable watermark / incomplete flag. This also needs a regression test through `handleTask` covering `cancelledByPoll` + abandoned/truncated delivery — the current tests stop at `executeAndDrain` and don't cover the swallow at the end.

### 2. When the completeness check genuinely fails, the task is stuck in `running` forever

`verifyTaskTranscript` (`server/internal/service/task.go:4223`) returns an error on a sequence gap; `server/internal/handler/daemon.go:3805` maps it to a 500; and `server/internal/daemon/daemon.go:5549`-`:5562` deliberately leaves the task in `running` once the terminal retry schedule is exhausted on a transient error. The failure callback path only logs (`:5628`). So the check prevents the false completion but never delivers the "explicit anomaly" the change is supposed to give the member — and the task can hold a serialization slot / execution capacity indefinitely.

`server/internal/handler/task_message_batch_test.go:283`-`:294` currently encodes "rejected → stays running" as the expected outcome, which locks in the gap. Please define and persist an explicit transcript-incomplete failure outcome, and add a regression test from the terminal callback through to the final task status. Asserting the 500 at the handler layer isn't enough to show the loop closes.

### 3. The new retry budget has no stated basis

`server/internal/daemon/client.go:1018`-`:1028` explains only that message retries should be shorter than the terminal-callback schedule — not why `250ms / 500ms / 1s`, four attempts, ~1.75s total backoff. Compare `defaultTerminalRetrySchedule` right above it (`:1004`), which cites a concrete incident (MUL-2780) for its window. This budget decides whether a brief blip is recovered or fails the task, for every task on the new build; it deserves the same justification. Please ground the numbers in a checkable latency constraint, incident data, or an existing budget, and have the tests assert the budget boundary rather than restating the implementation via `len(taskMessageRetrySchedule)+1`.

### 4. Migration numbering already collides with `main`

This PR adds `server/migrations/444_task_message_task_seq_unique_index.*` and `445_drop_task_message_task_seq_index.*`, but `main` already has `444_comment_recovery_settled_at.*` and `445_comment_delegated_failure_unsettled_index.*`. Numeric prefixes must be unique; GitHub currently reports this PR as `CONFLICTING`, with the merge tree conflicting in `server/cmd/migrate/main.go`. Please rebase onto latest `main`, renumber to the next free consecutive pair, and update the `concurrentIndexCleanups` / `concurrentDownIndexCleanups` keys accordingly. After that, please re-run the migration static checks and the affected package tests on the actual merge tree.

### Rollout notes (not blocking, but please confirm before merge)

- For multi-replica self-hosted deployments, the server capability needs to be fully rolled out before any runtime relies on replay.
- Safe rollback order is application first, keeping the unique index in place; only consider the down migration once no new runtime depends on it.
- The unique index fails closed if historical duplicates exist — the duplicate check described in the PR body should be a release gate, not a post-merge follow-up.
- If the design is wrong, it will show up as increased `transcript delivery failed`, a higher task failure rate, or tasks parked in `running` — worth watching on the first deploy.

### What I verified

- PR head `0fe397e9d`; migration prefixes 444/445 confirmed taken on `main`; merge state `CONFLICTING`.
- The error-swallow in the `cancelledByPoll` branch and the three transcript errors returned from `executeAndDrain` were confirmed in the head revision, as were the stuck-in-`running` path and the test that pins it.
- CI is green on this head, but a conflict-free merge result can't be produced right now, so merged-state verification is still outstanding.

Requesting changes on the four items above. Happy to re-review as soon as the rebase lands.

golb:
@Bohan-J Thanks for the detailed review. The PR has been rebased onto current main and all four blockers are addressed in `493a5e5bb`:

1. Cancellation now withholds `AckTaskCancelled` when transcript completeness is unknown, with a `handleTask` regression.
2. Sequence gaps persist an explicit terminal `transcript_incomplete` failure; complete/fail callbacks and lost-response replays are covered.
3. The retry schedule is tied to the 5s delivery deadline, with the exact budget asserted.
4. Migrations are renumbered to 446/447 and the complete migration sequence was applied successfully.

CI is green. Local verification also included `go test ./...`, targeted race tests, `go vet`, and the full migration run.

Rollout notes acknowledged: server capability first, application-first rollback while retaining the unique index, duplicate preflight as a release gate, and monitoring transcript delivery failures, task failure rate, and tasks parked in `running`.

Re-review would be appreciated.

multica-eve (changes requested):
Thanks for rebasing. The product direction is still right, and three of the original blockers are resolved: transcript gaps now become an explicit terminal failure with idempotent replay, the retry schedule is tied to the existing delivery deadline, and the migrations use free numbers with concurrent index operations on this branch.

Two issues still block merge at head `0c17c0609`:

### 1. The drain stop-timeout still loses the incomplete-transcript classification

`server/internal/daemon/daemon.go:8950` still returns a plain `errors.New("transcript drain did not stop after cancellation")`. The poll-cancellation path only withholds the positive flush acknowledgement when `errors.Is(err, errTranscriptIncomplete)` succeeds (`:5444`); this plain error falls through to `AckTaskCancelled` at `:5464`.

That means the last-resort timeout path can still acknowledge an unverifiably incomplete transcript as fully flushed—the same integrity failure this PR is intended to eliminate. Please wrap this return with `errTranscriptIncomplete` and add a regression that actually reaches `transcriptDrainStopTimeout` and asserts the classification. The existing tests cover an injected sentinel and `drainTruncated`, but not this branch.

### 2. The PR conflicts with current `main`

The PR is based on `9d09902df`; current `main` is `07430d115`. An actual `git merge-tree` conflicts in `server/pkg/taskfailure/failure_test.go`: main added `environment_prepare_failed`, while this PR adds `transcript_incomplete`, and both sides independently changed the taxonomy totals from 25 to 26.

Please rebase and retain both reasons. The combined totals should be 27 overall and 13 platform-side. Also update the canonical list/count documentation in `server/pkg/taskfailure/failure.go`, which currently says 26 and omits `transcript_incomplete` from the package-level list.

The migration files are currently correctly numbered `449/450`, but the PR description still says `444/445` and “Migration 444”; please update those references to the final numbers after the rebase.

Verification on the PR head: GitHub CI is green; `git diff --check`, the `ByteIdentical` guard, targeted daemon regressions, `pkg/taskfailure`, and migration cleanup-map tests pass. Merged-state tests cannot run until the conflict is resolved.

PR #7845, opened 2026-08-31
fix: surface autopilot anomalies in inbox

AtlieX:
Fixes HBH-563

Summary:
- Add a backend sweep that creates inbox alerts for skipped, failed, unknown, and stale autopilot runs.
- Surface the new alert type in inbox labels and locales.

Checks:
- cd server && /home/godfinns/local/go/bin/go test ./cmd/autopilot_anomaly_report
- cd server && /home/godfinns/local/go/bin/go test ./... (fails in pre-existing server/internal/daemon/repocache TestCreateWorktreeInstallsCoAuthoredByHook)
- corepack pnpm -C packages/core typecheck
- corepack pnpm -C packages/views typecheck
- corepack pnpm -C packages/views test -- inbox-detail-label.test.tsx

### DIFF
--- a/packages/core/types/inbox.ts
+++ b/packages/core/types/inbox.ts
@@ -18,6 +18,7 @@ export type InboxItemType =
   | "task_failed"
   | "agent_blocked"
   | "agent_completed"
+  | "autopilot_anomaly"
   | "reaction_added"
   | "quick_create_done"
   | "quick_create_failed"
--- a/packages/views/inbox/components/inbox-detail-label.tsx
+++ b/packages/views/inbox/components/inbox-detail-label.tsx
@@ -31,6 +31,7 @@ export function useTypeLabels(): Record<InboxItemType, string> {
     task_failed: t(($) => $.types.task_failed),
     agent_blocked: t(($) => $.types.agent_blocked),
     agent_completed: t(($) => $.types.agent_completed),
+    autopilot_anomaly: t(($) => $.types.autopilot_anomaly),
     reaction_added: t(($) => $.types.reaction_added),
     quick_create_done: t(($) => $.types.quick_create_done),
     quick_create_failed: t(($) => $.types.quick_create_failed),
--- a/packages/views/locales/en/inbox.json
+++ b/packages/views/locales/en/inbox.json
@@ -69,6 +69,7 @@
     "task_failed": "Task failed",
     "agent_blocked": "Agent blocked",
     "agent_completed": "Agent completed",
+    "autopilot_anomaly": "Autopilot anomaly",
     "reaction_added": "Reacted",
     "quick_create_done": "Created with agent",
     "quick_create_failed": "Create with agent failed",
--- a/packages/views/locales/ja/inbox.json
+++ b/packages/views/locales/ja/inbox.json
@@ -67,6 +67,7 @@
     "task_failed": "作業失敗",
     "agent_blocked": "エージェントがブロック",
     "agent_completed": "エージェントが完了",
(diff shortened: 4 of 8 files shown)


Bohan-J:
Thanks for taking a swing at this. We're going to close it, so let me be precise about what we're turning down.

The product direction is right, and the notification itself reads fine — a scheduled autopilot that silently stops producing results is a real gap, and the Inbox is the right place to surface it. What we can't take is the implementation.

I ran the branch locally against a seeded database rather than reviewing the diff alone. What came out of that:

**1. The headline scenario doesn't fire.** `evaluateCandidate` returns early for `issue_created / running / completed / cancelled`, so the stale-schedule check is only reachable when the autopilot has *never* produced a run. I set up an hourly autopilot three hours overdue whose last run had succeeded: no alert was created, and the existing one was archived by the sweep. That is the case the PR description opens with.

**2. `next_run_at` isn't a safe signal.** In this repo `autopilot_trigger.next_run_at` is display-only — see the comments around `AdvanceTriggerNextRun` in `server/internal/scheduler/jobs_autopilot.go`. Dispatch dueness is decided from `last_fired_at` and `autopilot_run`, the write is best-effort (`_ = queries.AdvanceTriggerNextRun(...)`), and on a cron parse failure it isn't updated at all — so a healthy autopilot can look overdue. It also skews the interval: because `next_run_at` is floored at plan time rather than a cron boundary, an hourly schedule renders in the alert body as "the cadence is about 59m0s".

**3. A user can't dismiss an alert.** `loadOpenAlerts` only reads `archived = false`. Once someone archives an alert while the anomaly persists, the next sweep no longer sees it and inserts a fresh one. A permanently broken hourly autopilot produces a new unread inbox item every tick.

**4. The status vocabulary is guessed.** `autopilot_run.status` is constrained to `issue_created / running / completed / failed / skipped`. `cancelled` cannot occur, so that arm is dead; meanwhile a run wedged in `running` — a genuine symptom — is classified as healthy. The `unknown_status` arm also writes the literal `"unknown_status"` into `details.last_run_status`, discarding the drifted value the comment says it wants an operator to see.

**5. Nothing in the repo runs it.** The command is built to be driven by an external cron on a host outside this project. Merged as-is, no deployment behaves differently.

**6. It bypasses the write path.** A raw `INSERT` into `inbox_item` from a separate process cannot publish `inbox:new` — `events.Bus` is in-process — so the alert skips WebSocket delivery, notification preferences, and the outbound integrations. `server/cmd/server/autopilot_failure_monitor.go` already does this correctly, and this PR reimplements its recipient resolution (creator → agent owner → workspace member) instead of reusing it.

**7. It doesn't build.** `apps/mobile/components/inbox/detail-label.tsx` holds an exhaustive `Record<InboxItemType, string>`; adding a member to the union without updating it fails `pnpm -C apps/mobile typecheck` with `TS2741: Property 'autopilot_anomaly' is missing`.

If you want to try again, the shape we'd accept is roughly:

- Run it inside the server process — extend `runAutopilotFailureMonitor` or add a JobSpec under `server/internal/scheduler` — using sqlc queries, `CreateInboxItem`, and a published `EventInboxNew`.
- Derive staleness from `last_fired_at` / the most recent `autopilot_run`, and evaluate it regardless of whether the last run succeeded.
- Let the alert link to the autopilot it is about, and stop regenerating it once a user has archived it.
- Settle how it relates to the existing `autopilot_paused` alert — thresholds, and whether one user can receive both.
- A new inbox type also needs the `apps/mobile` label map, `notifTypeToGroup`, and the wecom type label.

Happy to look at a follow-up along those lines.

Issue #7848, opened 2026-08-31
[Feature]: Add Russian (ru) localization


ADanMan:
### Deployment type

Official App

### Area

Web app

### What do you want and why?

Multica supports `en`, `zh-Hans`, `ko`, and `ja`, but not Russian. Requesting `ru` as a supported UI locale to serve Russian-speaking users across the web and desktop apps.

I'm volunteering to own the translation and keep it in parity going forward. Approach/scope is being discussed in #7847.

### Proposed solution (optional)

Tracking checklist (to be delivered in reviewable, per-namespace PRs, each keeping `parity.test.ts` green):

- [ ] Add `ru` to `SUPPORTED_LOCALES` in `packages/core/i18n/types.ts`
- [ ] Create `packages/views/locales/ru/*.json` mirroring the `en` key set
- [ ] Extend the API `language` enum (`packages/core/api/client.ts`) and backend user-language handling to accept `ru`
- [ ] Add `ru` to Mika onboarding language options
- [ ] (Optional) `conventions.ru.mdx` glossary/voice doc mirroring the zh conventions

Discussion for approach/scope: #7847

PR #7849, opened 2026-08-31
MUL-6908: feat(i18n): add Russian (ru) locale

ADanMan:
## What

Adds Russian (`ru`) as a fully selectable UI locale across the web and desktop apps, with translations for all 25 locale namespaces.

Part of #7848 (tracking) — approach discussed in #7847.

## Changes

**Wiring**
- `packages/core/i18n/types.ts`: `ru` added to `SupportedLocale` and `SUPPORTED_LOCALES`.
- `packages/views/locales/index.ts`: imports the `ru` bundles and registers `RESOURCES.ru`.
- `packages/views/settings/components/preferences-tab.tsx`: adds the **Русский** option to the language picker.
- `server/internal/handler/auth.go`: `ru` added to the `supportedLanguages` allow-list. This is required — the picker persists the choice via `updateMe({ language })`, and `user.language` round-trips back through `GetMe` into `i18n.changeLanguage()`. Without the server-side entry, selecting Russian would 400 and never sync across devices. The comment there already asks to keep the list in sync with `SUPPORTED_LOCALES`.
- The `Русский` endonym label is added to the `en`/`zh-Hans`/`ko`/`ja` settings bundles (the picker shows every language by its own name).

**Translations**
- `packages/views/locales/ru/*.json`: 25 namespaces. Each file mirrors the `en` key set 1:1, so the existing `parity.test.ts` guard stays green (verified locally). Interpolation placeholders (`{{count}}`, `<0>…`), URLs and brand names are preserved; plural keys keep the `_one`/`_other` shape.

## Testing

- Locale key parity checked against `en` for all 25 namespaces (matches `parity.test.ts` logic): no missing/extra keys.
- Every `ru/*.json` file validated as parseable JSON.
- `server/internal/handler` builds with the new allow-list entry.

## Follow-ups (intentionally out of scope, to keep this reviewable)

- Fuller Russian plural forms (`_few`/`_many`) — would also need the parity-test normalizer extended; the current `_one`/`_other` shape is correct and i18next falls back sensibly.
- Mika onboarding language (`en|zh|ko|ja`) is a separate value space with its own backend path.
- Documentation site (`apps/docs`) translation is a much larger, separate effort.

I'm happy to own the Russian locale going forward and keep it in parity as new keys land.


ADanMan:
Note on the red checks: `frontend-views-test (1/2)` fails only on `layout/sidebar-resize.test.tsx` ("previews width directly and commits only when the pointer is released" — expected `setItem` to be called once, got 0), and `frontend` is just the aggregate of that shard.

This reproduces on a clean `origin/main` checkout (1 failed / 1 passed) — i.e. it's a pre-existing failure in that pointer-drag test, not something this change touches. This diff doesn't modify `packages/views/layout/` at all.

The locale work itself is green locally:
- `tsc --noEmit` passes across the workspaces
- `parity.test.ts` passes for all 25 namespaces (ru keys mirror en 1:1)
- full `@multica/views` run: 4944 passed / 1 failed — the single failure being the pre-existing sidebar test above

Happy to rebase once that test is sorted.

ADanMan:
Rebased onto latest `main` and restored ru locale parity. Since this branch was opened, `main` added and retired keys across 11 namespaces; `packages/views/locales/parity.test.ts` was failing because ru lagged behind. Filled every missing ru key (notably `settings` `desktop.daemon.*`, `issues` `inline_run`/`reply.annotations`, `agents` `mcp_config`, `auth` `web.callback.*`) and dropped now-stale ru-only keys. ICU placeholders and EN plural suffixes preserved. `pnpm exec vitest run locales/parity.test.ts` → 212 passed, 0 failed locally.

ADanMan:
Merged latest `main`. The French locale (#7141) landed there and touched the same shared locale wiring this PR extends, so it conflicted across ~16 exhaustive records/switches/allow-lists (`SupportedLocale` union + `SUPPORTED_LOCALES`, `HTML_LANG` on web/desktop, landing `localeLabels`, `EMERGENCY_COPY`, use-cases i18n, `RESOURCES` map + imports, the preferences language selector, celestial workspace names, onboarding content-lang map, the task-failure locale test, and the Go `supportedLanguages` allow-list).

Every conflict was resolved by keeping both arms — main's new `fr` entry and this PR's `ru` entry both stay. The four non-ru `settings.json` conflicts (en/ja/ko/zh-Hans) took main's version since I don't own that content.

After the merge, refilled the ru bundle for the new EN keys French locale introduced: agents `success_hint` / `success_unavailable` / `cancelled_count`, skills toolbar `section_labels` / `filter_search_placeholder` / `no_labels` / `no_results`, and the shared `preferences.language` autonym labels so the selector renders every option.

Verified locally: `vitest run locales/parity.test.ts` 264 passed / 0 failed, `pnpm typecheck` 9/9 (confirms no dropped locale arm), `go build ./...` + auth/locale handler tests green.

PR #7850, opened 2026-08-31
test(cli): guard issue list --output json against control chars

ADanMan:
## What

Adds a regression test pinning the JSON-output contract of `issue list --output json`.

## Context

I could not reproduce the raw-`0x0A` defect on current `main`. The list path decodes the server response and re-encodes it through `encoding/json` (`cli.GetJSON` → `cli.PrintJSON`), which always escapes control characters, so a description containing newlines/tabs comes out as valid, escaped JSON. If a server ever sent a body with a raw control character inside a string (invalid JSON on the wire), `GetJSON` fails the decode and the command errors — it never forwards invalid JSON to stdout.

The existing list-json test only asserts the command returns no error, not that its stdout parses. This adds two focused tests:
- valid, parseable JSON output when a description holds `\n` / `\t` / `\r`, with a round-trip check on the value;
- the CLI never forwards an invalid server body to stdout.

## Testing

`go test ./server/cmd/multica/ -run TestIssueListJSON` — both pass.

Refs #7653. If anyone can still reproduce the raw-newline output, a `multica version` and a minimal sample on the issue would help pin which build/path it came from.

### DIFF
--- a/server/cmd/multica/cmd_issue_json_output_test.go
+++ b/server/cmd/multica/cmd_issue_json_output_test.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// Regression guard for #7653: `issue list --output json` must always emit
+// valid JSON. Descriptions can contain newlines, tabs and other control
+// characters; the CLI decodes the server response and re-encodes it through
+// encoding/json (cli.PrintJSON), which escapes those characters. The existing
+// list-json test only asserts the command returns no error, not that its
+// stdout parses — this pins the stronger contract.
+
+func TestIssueListJSONEscapesControlCharsInDescription(t *testing.T) {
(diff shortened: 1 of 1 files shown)


multica-eve:
Thanks for taking the time to investigate this. We're closing this PR because it does not address behavior present in the current product: the report was confirmed to come from an outdated deployment, and upgrading resolved it.

The `issue list` command has used `GetJSON` followed by `PrintJSON` since it was introduced, so there is no raw-response path in this repository's history for these tests to regress. The valid-response case primarily re-tests Go's `encoding/json` control-character escaping, while malformed-response rejection is already covered at the API client layer. The mock also does not exercise the actual server `ListIssues` / `writeJSON` path from the original report.

Without a reproducible failure on current code or a concrete change that introduces this risk, we don't want to add command-specific maintenance coverage for a hypothetical bypass. If the problem becomes reproducible on current `main`, a test that fails before the corresponding fix would be welcome alongside that fix.

PR #6273, opened 2026-08-02
MUL-5625: WS-2867: bound daemon workspace cache growth

tangyuanjc:
## Summary

- create and self-heal `.metadata_never_index` at daemon workspace roots
- treat `codex-home/.tmp` as an exact managed artifact alongside `.sandbox-bin`
- reclaim task-local Codex caches after the provider exits and before terminal reporting, while retaining sessions, logs, workdir, output, auth, and config
- extend GC/disk-usage coverage and document the resulting lifecycle

## Safety and scope

- cleanup continues to reject symlinks and preserves same-named directories inside user repositories unless operators explicitly configure broad basename matching
- whole task roots retain the existing issue/orphan TTL behavior, so issue continuation and session resume are unchanged
- the live `c37d6ffd` / WS-2512 `data/runs` outlier is application-owned scheduled output, not runtime cache; this PR intentionally does not delete it or add a generic size cap

## Test plan

- [x] `go test -count=1 -parallel=2 ./internal/daemon/execenv ./internal/daemon`
- [x] `git diff --check origin/main...HEAD`

Closes WS-2867

PR #7855, opened 2026-09-01
MUL-6914: WS-4609: reclaim per-task Codex temp caches

tangyuanjc:
## Summary

- keep `codex-home/.tmp` reusable between nearby Codex runs and reclaim it only after a dedicated 72-hour idle TTL
- run the cleanup inside each daemon's existing GC loop, covering every configured workspace root without fixed path-depth assumptions
- reserve the env root and revalidate ownership plus last-use timestamps before removing only the exact managed path
- include `.tmp` in disk-usage accounting without treating repository-owned `.tmp` directories as managed

## Review feedback addressed

- Removed task-exit cleanup. Installed marketplaces and plugins now stay warm for resumed and offline runs; a recent second completion refreshes the idle clock.
- The `.tmp` deletion path does not run `dirSize` or any other logical-size pre-scan before `RemoveAll`, avoiding the synchronous double walk and APFS overstatement.
- Rebased onto current `main`; the PR is mergeable again.
- Added regression coverage for active-root exclusion, recent multi-run reuse, exact-path preservation, symlink containment, the 72-hour default/override, and cross-root disk-usage accounting.

## Tests

```bash
go test ./internal/daemon ./internal/daemon/execenv -count=1
go test -race ./internal/daemon -run 'Test(CodexTempGC_|RunGC_CodexTempSkipsActiveTask)' -count=1
git diff --check
```

Supersedes the cache-cleanup portion of #6273; excludes its Spotlight change.

Closes WS-4609


multica-eve (changes requested):
This change reduces disk usage immediately after each agent run, but it also discards reusable plugin content that the same task may need to download again when it continues. The disk-growth problem is real and worth solving; however, the default behavior should preserve short-term reuse and reclaim the content after it has been idle, unless the repeated-download cost is explicitly accepted.

Requesting changes for two blocking issues:

1. `server/internal/daemon/execenv/reclaimable.go:22-28` classifies the whole `codex-home/.tmp` tree as exit-time reclaimable, and `server/internal/daemon/daemon.go:7329-7334` deletes it on every Codex `runTask` return, including `StartTask` failures where the provider never ran. Upstream Codex defines `.tmp/marketplaces` as the installed-marketplace directory, upgrades configured marketplaces at startup, and clones them again when the destination is missing. This turns the approximately 305 MiB marketplace checkout into a per-run network/cold-start cost, while the PR intentionally preserves the approximately 285 MiB `.sandbox-bin` to avoid the same class of repeated download. The new tests use a synthetic `.tmp/plugins` fixture and do not cover the real two-run lifecycle or offline behavior. Please remove unconditional task-exit cleanup and keep this exact path under the existing TTL-based managed GC, or provide an explicit product decision for the repeated-download/offline tradeoff plus a regression test proving `.tmp/marketplaces -> cleanup -> next run rehydrates -> plugins remain usable`.

   Upstream references: [installed marketplace location](https://github.com/openai/codex/blob/e017e93aceafb2fe04bed1c926e448a5fb4f913d/codex-rs/core-plugins/src/installed_marketplaces.rs#L12-L16), [startup upgrade](https://github.com/openai/codex/blob/e017e93aceafb2fe04bed1c926e448a5fb4f913d/codex-rs/core-plugins/src/manager.rs#L2686-L2720), [clone when missing](https://github.com/openai/codex/blob/e017e93aceafb2fe04bed1c926e448a5fb4f913d/codex-rs/core-plugins/src/marketplace_upgrade.rs#L220-L276).

2. `server/internal/daemon/gc.go:904-905` performs a full `dirSize` walk immediately before `os.RemoveAll` walks the tree again. The task-exit caller runs before task completion/result reporting while the environment-root claim is held, so this adds an avoidable synchronous scan of a large Git checkout to the user-visible completion path. If immediate cleanup remains, please use a deletion path that does not calculate bytes first.

The exact managed path, containment checks, active-root protection, and claim ordering are otherwise sound. CI is green, and the targeted daemon tests, race tests, full affected-package tests, `ByteIdentical` guard, and `git diff --check` all pass on `86e801a631ef8308072c9bd49ad79756dc0960b7`.

tangyuanjc:
Addressed the blocking review feedback in `87258dcbc`:

- removed unconditional task-exit cleanup; `codex-home/.tmp` now remains reusable and offline-capable between runs
- added a dedicated 72-hour idle TTL, refreshed by recent task completion metadata
- cleanup takes the active-root reservation and revalidates ownership/timestamps before exact-path removal
- `.tmp` removal does not call `dirSize` or otherwise pre-walk the tree
- rebased onto current `main`; the PR is mergeable
- added active-task, recent second-run reuse, symlink containment, exact-path, TTL-config, and cross-root accounting tests

Local affected-package tests and targeted race tests pass. CI for the replacement head is running/passing.

multica-eve (changes requested):
This revision now keeps recently used plugin content available between runs and reclaims it after a longer idle period, which fixes the previous per-run redownload and offline breakage. However, a recent run that fails or is cancelled does not extend that retention period, and the new default cleanup policy also overrides an existing operator opt-out.

Requesting changes for two blocking issues:

1. The current "last used" calculation misses failed and cancelled runs. `server/internal/daemon/gc.go:1023-1051` considers only the `.tmp` root mtime, GC metadata `CompletedAt`, and the metadata file mtime. Reading or reusing nested marketplace/plugin files does not update the `.tmp` root mtime, while the cancellation and failure returns in `server/internal/daemon/daemon.go:5435-5482` happen before the success-only metadata write at `daemon.go:5511-5535`. Therefore, a cache older than 72 hours can be used by a task, protected while active, and then deleted by the next GC immediately after that task fails or is cancelled. The current test at `server/internal/daemon/gc_codex_temp_test.go:63-97` manually calls `WriteGCMeta`, so it covers only the successful-completion case. Please update a separate lightweight last-used marker when Codex actually claims/uses the environment, independently of task success, and add regressions for `old cache -> failed/cancelled run -> GC keeps cache`.

2. The independent default-on 72-hour policy is not justified and changes the behavior of an existing cleanup opt-out. `server/internal/daemon/config.go:72,118,533-536` adds `MULTICA_GC_CODEX_TEMP_TTL=72h`, but neither the code nor the PR explains why 72 hours is the correct product boundary. More importantly, operators could previously set `MULTICA_GC_ARTIFACT_TTL=0` to disable managed artifact cleanup, including the Codex cache; `server/internal/daemon/gc.go:984-1009` now checks only the new TTL, so those deployments begin deleting `.tmp` after upgrading despite their explicit opt-out. The new variable is also absent from the authoritative configuration table at `CLI_AND_DAEMON.md:263-274`. Please reuse the existing `GCArtifactTTL`/kill switch, or obtain an explicit maintainer decision for the separate policy and document/test the 72-hour rationale, compatibility or migration semantics for `MULTICA_GC_ARTIFACT_TTL=0`, and the new configuration variable.

The previous two blockers are resolved: task-exit cleanup is gone and `.tmp` deletion no longer performs a `dirSize` prescan. Exact-path selection, ownership checks, active-root reservation, timestamp revalidation, and symlink/junction refusal are sound. CI plus targeted, race, full affected-package, `ByteIdentical`, and diff checks pass on `87258dcbc864ad3c8d625bb24de7e019b0251db3`.

PR #7858, opened 2026-09-01
MUL-6918: WS-4612: recover scheduled autopilot resource failures

tangyuanjc:
## Summary

- add an opt-in, single delayed retry for scheduled `run_only` autopilot failures caused by provider quota, entitlement/access, or rate limits
- keep the original autopilot run open and preserve its run ID across the deferred retry
- atomically finalize an exhausted scheduled run and write an idempotent machine-readable failure receipt to a configured issue
- classify and backfill Claude Code's `out of extra usage` failure as provider quota exhaustion
- expose and validate the recovery configuration through the API, core types/schema, CLI, and built-in autopilot skill
- leave `validation_officer_daily_report` unchanged

## Verification

- full migration chain through 445 on a fresh PostgreSQL database
- `go test ./cmd/server -count=1` on that fresh database
- `go test ./internal/service ./internal/handler ./internal/migrations ./pkg/taskfailure -count=1 -p 1`
- focused autopilot CLI tests
- `pnpm exec vitest run packages/core/api/schema.test.ts` (89 passed)
- `make sqlc`
- `git diff --check`

## Rollout

After the new server version is deployed, configure autopilot `61d55f53-159e-4fbb-ae76-db320ef541d1` with a 30-minute resource-failure retry and a `validation_officer_daily` failure receipt on WS-1270. Production acceptance then needs a natural scheduled failure or an explicitly authorized injected failure, with its run ID and receipt comment link recorded on WS-4612.


multica-eve (changes requested):
Today, an unattended scheduled automation that encounters a temporary usage or access limit stops immediately, even if the restriction clears shortly afterward. This proposal gives selected automations one later attempt and leaves a visible notice when recovery still fails.

## Product judgment

The problem is real, and an opt-in recovery path is worthwhile for low-frequency unattended runs. I would keep the proposed boundary of one durable retry on the same run, but only for failures that are demonstrably self-healing and only when replay is safe; the current implementation does not yet establish those guarantees.

## Must fix

1. `server/internal/service/task.go:5008-5071` retries authentication/access, generic quota, and rate-limit failures without distinguishing which ones can recover automatically, and without proving that the failed attempt performed no tools or external side effects. Restrict recovery to structured temporary signals and fail closed when replay safety is unknown.
2. `server/internal/service/task.go:5057-5064` and `server/internal/service/autopilot.go:1144-1195` read the Autopilot's mutable current configuration when a failure occurs. Snapshot the retry policy, delay, and receipt target on the run/rule version so an edit cannot change an in-flight run's behavior or audit destination.
3. `server/internal/service/autopilot.go:1167-1178` and `server/pkg/db/queries/autopilot.sql:417-420,464-481` do not fence state transitions against the current task/run state. A late event from an older attempt can rebind or overwrite the terminal result. Use compare-and-swap conditions and make final run state, quota settlement, and receipt creation atomic.
4. `server/migrations/445_backfill_extra_usage_quota_failure.up.sql:11-21` performs unbounded `%...%` scans/updates on `agent_task_queue` and `chat_message` during startup. Historical rewriting is not required for the new behavior; remove it or move it to a bounded, resumable operational backfill.
5. Migrations 444/445 now conflict with numbers already present on `main` after #7820. Rebase, renumber, regenerate SQLC output, and rerun the migration chain.
6. `server/internal/service/autopilot.go:1264-1273` inserts an unvalidated failure string into a machine-readable marker. Emit only canonical reason codes or safely encode the value so malformed content cannot corrupt the receipt contract.
7. The fixed 30-minute minimum has no verified policy basis. Prefer a structured provider reset/`Retry-After` signal, or obtain an explicit maintainer decision that 30 minutes is the intended product policy.

This PR also combines error classification/backfill, retry policy configuration, run orchestration, and failure receipts in one change. Please narrow or split those responsibilities so the retry core can be reviewed independently.

If this is wrong, opted-in scheduled runs can duplicate external effects or show inconsistent run/receipt state; the rollback is to disable the opt-in policy and revert the application while leaving the additive schema fields in place.

**Merge recommendation: Request changes.**

PR #7863, opened 2026-09-01
fix(storage): observe context cancellation in local uploads

### LINKED ISSUE #7784: [Bug]: LocalStorage upload methods ignore context cancellation
## Deployment type

self-host

## What happened?

`LocalStorage.Upload`, `LocalStorage.UploadStream`, and `LocalStorage.UploadFromReader` accept a `context.Context`, but the local storage implementation does not observe cancellation while reading or writing upload data.

For streamed or large uploads, cancelling the request does not stop the local copy. The operation continues until the source reader reaches EOF or returns another error. This can keep request goroutines and filesystem I/O active after their caller has gone away. `UploadFromReader` also reads the complete source before reaching `Upload`, without checking the context.

Expected behavior:

- a pre-cancelled context prevents the upload from starting;
- cancellation during a copy stops the operation after the current read returns;
- the staging file is removed;
- an existing object at the same key remains intact;
- the returned error preserves `context.Canceled` or `context.DeadlineExceeded` for `errors.Is`.

## Steps to reproduce

1. Create a `LocalStorage` instance using a temporary upload directory.
2. Start `UploadStream` with a reader that returns data in controlled chunks.
3. Wait until the first chunk has been consumed, then cancel the context.
4. Allow the reader to return another chunk.
5. Observe that the upload continues instead of returning a context cancellation error.

The same behavior can be demonstrated by passing an already-cancelled context to `Upload`: the file is still written successfully.

## Additional context

The relevant code is in `server/internal/storage/local.go`:

- `LocalStorage.Upload`
- `LocalStorage.UploadStream`
- `LocalStorage.UploadFromReader`
- `writeAtomic`

`writeAtomic` currently receives only an `io.Reader`, so cancellation information is lost before the copy begins. A context-aware copy path should retain the existing atomic-write guarantee and remove its staging file on cancellation.

adaiguoguo:
Fixes #7784.

`LocalStorage.Upload`, `UploadStream` and `UploadFromReader` take a `context.Context` and never look at it. `io.Copy` and `io.ReadAll` have no context of their own, so a cancelled request keeps reading its source to EOF: the goroutine and its filesystem writes outlive the caller, and a client that hung up still ends up with a stored object. `UploadFromReader` is the worst of the three — it buffers the entire source before `Upload` is even called, so a stalled reader is drained in full before anything consults `ctx`.

Reads now go through a `contextReader` that checks `ctx` before each `Read`, which bounds the overshoot at one in-flight read rather than the whole body. Cancellation is reported as the context error itself rather than wrapped in a storage message, because callers use `errors.Is` to tell "the client went away" from a real storage failure.

Nothing is destroyed on the way out: `writeAtomic` already stages through a temp file, so an aborted copy removes its own staging file and leaves any existing object at that key intact. The pre-flight check means an already-cancelled caller never creates one.

This matches the expected behaviour listed in the issue point for point:

| Expected | Test |
| --- | --- |
| pre-cancelled context prevents the upload from starting | `TestLocalStorage_UploadRejectsCancelledContext` (also asserts the directory stays empty) |
| cancellation during a copy stops after the current read | `TestLocalStorage_UploadStreamCancelMidCopyKeepsExistingObject` |
| staging file is removed | same test (`tempPath` must not exist) |
| existing object at the same key remains intact | same test (content still `original`) |
| error preserves `context.Canceled` / `DeadlineExceeded` | all four, incl. `TestLocalStorage_UploadPreservesDeadlineExceeded` |

`UploadFromReader` gets its own test since its failure mode is distinct (drains before delegating).

**These are real regression tests** — on the current implementation all four fail: two by completing the upload and returning `nil`, two by copying past cancellation for the full 5s timeout.

### DIFF
--- a/server/internal/storage/local.go
+++ b/server/internal/storage/local.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -153,7 +154,7 @@ func (s *LocalStorage) DeleteKeys(ctx context.Context, keys []string) {
 
 func (s *LocalStorage) Upload(ctx context.Context, key string, data []byte, contentType string, filename string) (string, error) {
 	dest := filepath.Join(s.uploadDir, key)
-	if err := writeAtomic(dest, bytes.NewReader(data)); err != nil {
+	if err := writeAtomic(ctx, dest, bytes.NewReader(data)); err != nil {
 		return "", err
 	}
 	// Best-effort sidecar so ServeFile can restore the original filename in
@@ -177,7 +178,7 @@ func (s *LocalStorage) Upload(ctx context.Context, key string, data []byte, cont
 
 func (s *LocalStorage) UploadStream(ctx context.Context, key string, data io.Reader, _ int64, contentType string, filename string) (string, error) {
 	dest := filepath.Join(s.uploadDir, key)
-	if err := writeAtomic(dest, data); err != nil {
+	if err := writeAtomic(ctx, dest, data); err != nil {
 		return "", err
 	}
 	if filename != "" {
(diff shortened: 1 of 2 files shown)


adaiguoguo:
Pushed a test fix — the implementation is unchanged.

`TestLocalStorage_UploadFromReaderStopsReadingOnCancel` was racy: the fake reader blocked on a channel after its first chunk, so when the copy goroutine reached its second `Read` before the test called `cancel()`, it parked there for the full timeout. It passed locally and hit the losing side of the race on CI.

The fake is now an endless reader — every `Read` returns another byte and it signals after the first one, so the test cancels genuinely mid-copy and cancellation is the only thing that can end the copy. It cannot block, so a regression fails as an assertion rather than a timeout (there is a 1 MiB read cap as the backstop).

Verified: 50 runs under `-race` are stable, and all four tests still fail against the current implementation — now in ~1.2s rather than by timing out.

adaiguoguo:
Still applies to `main` at 8c4f4328f — `writeAtomic` takes no context, so a cancelled upload keeps copying to completion and the caller cannot tell "client went away" from a storage failure. Two files, CI green, mergeable.

PR #7865, opened 2026-09-01
refactor(execenv): list a repository in one prompt section, not two

adaiguoguo:
Fixes the repository half of #7773. The autopilot half landed upstream independently as MUL-6984 (#7971) — this PR is rebased on top of it and now carries only what remains.

A `github_repo` project resource is rendered by Project Context with its ref, default-branch hint and label. `## Repositories` then listed the bare URL again — a line that says strictly less than the one above it, and leaves the agent to work out that the two are the same repository. Those repos are skipped there now, and the section is elided when nothing is left. Unmatched workspace repos still appear: a workspace repo is reachable whether or not a project points at it.

URL matching ignores a `.git` suffix, a trailing slash and case, because the two sides store the same repository both ways in practice; it never rewrites host or path, so two genuinely different URLs cannot collapse into one.

**Out of scope, verified untouched**: the runtime repository list used for checkout authorization and caching, and the sidecar context files.

| Acceptance | Test |
| --- | --- |
| project repo appears once, metadata kept in Project Context | `TestRepositoriesOmitsReposAlreadyInProjectContext` |
| section elided when nothing is left | `TestRepositoriesSectionElidedWhenEveryRepoIsAProjectResource` |
| unmatched workspace repos survive | same, plus `TestProjectRepoDedupKeepsDistinctRepos` |
| `.git` / trailing slash / case are the same repo | `TestProjectRepoDedupIgnoresGitSuffixAndTrailingSlash` |
| non-github resources hide nothing | `TestRepositoriesUnaffectedByNonGithubResources` |

### DIFF
--- a/server/internal/daemon/execenv/runtime_config_dedup_test.go
+++ b/server/internal/daemon/execenv/runtime_config_dedup_test.go
@@ -1,0 +1,124 @@
+package execenv
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func githubResource(url, label string) ProjectResourceForEnv {
+	ref, _ := json.Marshal(map[string]string{"url": url})
+	return ProjectResourceForEnv{ResourceType: "github_repo", ResourceRef: ref, Label: label}
+}
+
+// A repo carried by a project resource is already rendered in Project
+// Context with its ref, branch hint and label. The bare URL in
+// ## Repositories says strictly less and costs tokens twice.
+func TestRepositoriesOmitsReposAlreadyInProjectContext(t *testing.T) {
+	t.Parallel()
+
+	ctx := TaskContextForEnv{
+		IssueID: "i1",
+		Repos: []RepoContextForEnv{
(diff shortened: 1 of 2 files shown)


adaiguoguo:
Rebased and narrowed: the autopilot half of this dedup landed upstream as MUL-6984 (#7971) — nice to see it arrive, and it confirms the direction. What remains here is only the repository half: one section per repo, with the richer Project Context line winning. The superseded autopilot tests are dropped in favor of upstream's own in execenv_test.go.

adaiguoguo:
Still applies to `main` at 8c4f4328f — `runtime_config_sections.go` still describes `multica repo checkout` in two separate sections (lines 279 and 425), so an agent reads the same capability twice with different detail. Two files, CI green, mergeable.

multica-eve:
关联的 Issue #7773 已完成决策：Autopilot 部分已由 #7971 覆盖；仓库部分暂不采用，因为移除 Repositories 条目会丢失 #8592 引入的分支交付规则。该 PR 的方案与当前决策冲突，关闭。

PR #7868, opened 2026-09-01
MUL-6921 fix(daemon): honor the Co-authored-by toggle in existing checkouts

Bohan-J:
## Problem

Reported by a Gitea user: after turning the **Co-authored-by** switch off in workspace settings, every new commit in a repo the agent had **already checked out** still carried `Co-authored-by: multica-agent <github@multica.ai>`. Their Gitea renders that trailer as a second commit author with its own avatar, so the stale value is publicly visible — and the address maps to no real account on their instance.

Two independent causes:

1. **The hook froze the decision at checkout time.** `prepare-commit-msg` is installed in the git *common* directory, so it outlives the checkout that created it and applies to every worktree of that bare cache. The script appended the trailer unconditionally; the setting was only consulted when `CreateWorktree` ran again. `git commit --no-verify` doesn't help — it bypasses `pre-commit` and `commit-msg`, not `prepare-commit-msg`.
2. **A settings edit never reached a running daemon.** `UpdateWorkspace` woke daemons only on a *rename*, and the periodic workspace sync deliberately makes no repos/settings request for a workspace it already tracks. So the daemon's cached verdict (and with it the hook) could stay stale until the workspace's next repo checkout.

## Change

**The hook reads the setting at commit time.** It consults a state file (`.repos/<workspace-id>/.multica_co_authored_by`) instead of trusting its own presence on disk; `0` → exit without touching the message. A missing or unreadable file keeps the trailer: the hook only exists because a checkout ran with the setting on, so "no state recorded" must mean what it meant before the file existed.

**The daemon is the sole publisher of that state.** It writes it on every settings refresh and every workspace sync (local, no request, skipped when the value on disk already matches), under a per-workspace lock with the verdict read *inside* the lock — so a publisher that started before a settings update can never overwrite that update's result. A checkout does **not** publish: `CoAuthoredByEnabled` is a snapshot taken before fetch and lock waits, so a slow checkout writing it back could resurrect a trailer the user had turned off. The checkout instead *reads* what was published, and falls back to its snapshot only when nothing has been published yet.

**Hooks from earlier releases are migrated, not waited on.** They read no state, so publishing a value cannot reach them. Publication now reconciles the hooks it owns: enabled rewrites them to the gated script, disabled deletes them, hooks the daemon did not install are left alone. This covers both hook locations — the workspace's bare caches, and checkouts with isolated git metadata (Codex on Linux, the Windows sandbox), which keep `.git` inside the task workdir and are found by walking the daemon's env roots. Hook files are published by rename, since these repos have live checkouts committing against them.

Env roots are attributed by the strongest evidence the release that created them left behind. Since v0.4.35 the owner record names the workspace and an exact match is required. Releases before it recorded a bare task ID, or no marker at all, and put env roots under `<workspaces root>/<workspace ID>/<task key>` — so when the owner record names no workspace, the directory name is the evidence and has to *be* the workspace ID. The readable layout that replaced it always appends a suffix, so a bare UUID directory can only be one of those older roots. Anything that cannot be attributed is skipped: a stale hook in one workdir is cheaper than rewriting another workspace's.

**Settings edits now reach running daemons.** `UpdateWorkspace` notifies members' daemons on a settings change, not just a rename; that hint refreshes settings for the workspaces the daemon already tracks. A reconnect reconcile does the same, so an edit made while the websocket was down is picked up on reconnect instead of being republished stale.

Net effect: existing checkouts converge without any user action, and the manual "move the hook aside, commit, put it back" workaround is no longer needed.

## Verification

Regression tests, each **verified failing with its fix reverted**:

- `TestCoAuthoredByStateStopsTrailerInExistingCheckout` — reproduces the report end-to-end with real git: commit with the setting on → flip the published state with **no new checkout** → the next commit, and a `--no-verify` commit, carry no trailer → flipping back restores it.
- `TestCreateWorktreeDoesNotResurrectDisabledState` — a checkout finishing with a stale `enabled` snapshot must not republish it.
- `TestPersistCoAuthoredByStateReadsVerdictUnderPublishLock` — channel-driven, no sleeps: a publisher is parked between taking the lock and reading, and a second publisher reaching the cache during that window is the failure.
- `TestReconcileCoAuthoredByHooksMigratesReleasedHook` / `…UpgradesReleasedHookInPlace` / `TestReconcileCoAuthoredByHookInCheckoutMigratesReleasedHook` — seed the previous release's hook **verbatim** in a bare cache and in an isolated checkout, then reconcile with **no `CreateWorktree` call** and assert against real commits.
- `TestReconcileCoAuthoredByHooksPreservesUserHook`, `TestReconcileCoAuthoredByHookInCheckoutIgnoresLinkedWorktree` — the sweep stays inside what the daemon owns.
- `TestPersistCoAuthoredByStateReconcilesIsolatedCheckouts` — the env-root walk reconciles this workspace's checkout and not another workspace's.
- `TestPersistCoAuthoredByStateReconcilesLegacyEnvRoots` — both pre-v0.4.35 shapes (no marker, task-ID-only marker) are migrated, while the same shapes under another workspace's directory, and a readable-layout root whose marker names no workspace, are left alone.
- `TestWorkspaceSyncLoop_ReconcilePicksUpSettingsChangedWhileOffline`, `TestRefreshTrackedWorkspaceSettingsAppliesToggle`, `TestSyncWorkspacesPublishesCoAuthoredByState`, `TestUpdateWorkspace_NotifiesDaemonsOnSettingsChange`.

Suites: `go test ./internal/daemon/repocache/` fully green; `./internal/daemon/` green except 7 pre-existing `identity`/`config`/`Prepare*Home` failures that fail identically on `main` in this environment; `-race` over the touched tests in both packages (`-count=5` on the concurrency ones); `go vet ./internal/daemon/...` and `git diff --check` clean.

## Compatibility and rollout

- Old daemon + new server: the extra wakeup is a workspace re-sync it already handles.
- New daemon + old server: no settings wakeup, so a toggle lands on the next reconnect (websocket connect/reconnect) or the workspace's next repo checkout — the periodic sync alone does not re-fetch settings for a tracked workspace by design.
- Upgraded hosts migrate their existing hooks on the daemon's first workspace sync after start; no user action, no re-checkout.
- The state file is a non-directory entry inside the repo cache; GC, eviction and disk-usage accounting all skip non-directories, so nothing else changes.

### DIFF
--- a/server/internal/daemon/coauthor_enabled_test.go
+++ b/server/internal/daemon/coauthor_enabled_test.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/multica-ai/multica/server/internal/daemon/repocache"
 )
@@ -132,8 +133,9 @@ func TestSyncWorkspacesSkipsReposRefreshOnExistingWorkspace(t *testing.T) {
 // coAuthoredByStateCache is a repoCacheBackend that records what the daemon
 // publishes for installed prepare-commit-msg hooks to read.
 type coAuthoredByStateCache struct {
-	mu     sync.Mutex
-	writes []bool
+	mu         sync.Mutex
+	writes     []bool
+	reconciles []bool
 }
 
 func (c *coAuthoredByStateCache) Lookup(string, string) string   { return "" }
@@ -151,6 +153,23 @@ func (c *coAuthoredByStateCache) WriteCoAuthoredByState(_ string, enabled bool) 
 	defer c.mu.Unlock()
 	c.writes = append(c.writes, enabled)
 	return nil
+}
+
+func (c *coAuthoredByStateCache) ReconcileCoAuthoredByHooks(_ string, enabled bool) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.reconciles = append(c.reconciles, enabled)
+	return nil
+}
+
+func (c *coAuthoredByStateCache) lastReconcile(t *testing.T) bool {
+	t.Helper()
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.reconciles) == 0 {
+		t.Fatal("hooks were never reconciled")
+	}
+	return c.reconciles[len(c.reconciles)-1]
 }
 
 func (c *coAuthoredByStateCache) lastWrite(t *testing.T) bool {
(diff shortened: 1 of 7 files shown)


Bohan-J:
Pushed `707a4abd3` addressing all three must-fixes.

**1. A stale checkout could republish `enabled`** — the checkout no longer writes the state file at all. The daemon is the sole publisher (it is the only side that holds the ordering between settings updates), and `applyCoAuthoredBySettingContext` now *reads* what was published, falling back to `params.CoAuthoredByEnabled` only when nothing has been published yet. Publication itself takes a per-workspace lock and reads the verdict **inside** it, so two publishers serialize and the one that writes last is the one that read last — a publisher that started before a settings update can no longer overwrite its result. Tests: `TestCreateWorktreeDoesNotResurrectDisabledState` (stale `true` snapshot against a published `false`, asserted through a real commit) and `TestPersistCoAuthoredByStateReadsVerdictUnderPublishLock` (publisher blocked on the lock, settings flipped underneath, must publish the new value).

**2. Pre-upgrade hooks never migrated** — added `Cache.ReconcileCoAuthoredByHooks(workspaceID, enabled)`, called from `persistCoAuthoredByState`, i.e. on every settings refresh **and** every workspace sync (so also right after daemon start on an upgraded host). Enabled rewrites daemon-owned hooks to the gated script, disabled deletes them, hooks the daemon does not own are untouched. Hook files are now published by rename — these repos have live checkouts committing against them, and a truncate-and-write was visible as a half-written script. Tests build the checkout with the **previous release's hook verbatim** and never call `CreateWorktree` afterwards: `TestReconcileCoAuthoredByHooksMigratesReleasedHook` (disable → real commit has no trailer), `TestReconcileCoAuthoredByHooksUpgradesReleasedHookInPlace` (enable → hook upgraded, then a publish alone stops the trailer), `TestReconcileCoAuthoredByHooksPreservesUserHook`, plus `TestPersistCoAuthoredByStateReconcilesHooks` for the daemon wiring.

Known limit, stated deliberately: only bare caches are enumerable from the cache root. An isolated checkout (Codex / Windows) keeps its hook inside its own task workdir, so a pre-upgrade one migrates at its next checkout. Post-MUL-6921 isolated checkouts are gated at install, so they honor the toggle immediately.

**3. Toggle flipped during a websocket outage** — the reconnect reconcile branch now calls `refreshTrackedWorkspaceSettings` before the sync, so the daemon re-reads settings for workspaces it already tracks instead of republishing the stale cached value. Test: `TestWorkspaceSyncLoop_ReconcilePicksUpSettingsChangedWhileOffline` (cached `true` → server holds `false` → reconcile broadcast → new value read and published, no checkout involved). The now-inaccurate comment in `wakeup.go` about settings refreshing only when a checkout needs them was updated.

Each of the three new regression tests was confirmed **failing** with its fix reverted and passing with it.

Verification: `go test ./internal/daemon/repocache/` all green; `./internal/daemon/` green except the 7 pre-existing `identity`/`config`/`Prepare*Home` failures that fail identically on `main` in this environment; `-race` run over the touched tests in both packages; `go vet ./internal/daemon/...` clean.

Bohan-J:
Pushed `98ada3ead`, and the PR description is rewritten to match the current diff.

**Must-fix — isolated checkouts now migrate too.** Publication walks the daemon's env roots for the workspace and reconciles every checkout that owns its git metadata, so a Codex/Windows workdir prepared by an earlier release stops adding the trailer without waiting for a re-checkout. Details:

- Env roots are attributed by their `.task_owner` record, so a workspace never rewrites another workspace's hooks; an env root with no recorded workspace is skipped rather than guessed at.
- A linked worktree is skipped (its `.git` is a file, and its hook belongs to the bare cache that the other sweep already handles).
- The walk is bounded to the depth repos are checked out at and never descends into a matched checkout.

Tests: `TestReconcileCoAuthoredByHookInCheckoutMigratesReleasedHook` seeds the previous release's hook verbatim in a real isolated checkout, reconciles with **no `CreateWorktree` call**, and asserts against real commits; `TestPersistCoAuthoredByStateReconcilesIsolatedCheckouts` covers the env-root walk including the wrong-workspace and dot-directory cases; `TestReconcileCoAuthoredByHookInCheckoutIgnoresLinkedWorktree` pins the skip. Both new paths verified failing without the fix.

So the "existing checkouts converge with no user action" claim now holds for both checkout modes, and I have not narrowed the issue's promise.

**Nit 1 — the concurrency test is deterministic now.** The verdict read is injected (`publishCoAuthoredByState(workspaceID, verdict)`; production passes `workspaceCoAuthoredByEnabled`), so the test parks a publisher *between taking the lock and reading* and starts a second one after flipping the setting. The second publisher reaching the cache during that window is itself the failure — no sleep decides the outcome. Moving the read back outside the lock fails it immediately (0.00s, 3/3 runs); as written it also passes `-race -count=5`.

**Nit 2 — PR description updated.** It now says the checkout consumes rather than publishes the state, states the isolated-checkout coverage, and corrects the new-daemon-with-old-server line: the toggle lands on the next reconnect or the workspace's next checkout, since the periodic sync deliberately does not re-fetch settings for a tracked workspace.

Verification on `98ada3ead`: `repocache` fully green; `./internal/daemon/` green except the 7 pre-existing `identity`/`config`/`Prepare*Home` failures that fail identically on `main` here; `-race` over the touched tests in both packages (`-count=5` on the concurrency ones); `go vet ./internal/daemon/...` and `git diff --check` clean.

Bohan-J:
Pushed `85d238e4b`. The legacy env-root gap is closed, and the PR description now states the attribution rule.

I verified the history against the tags before changing anything, and it matches your read: `cd169947b` (v0.4.35) introduced both the readable layout and the workspace-aware owner record; `v0.4.34` and earlier used `<workspaces root>/<workspace ID>/<task key>` and wrote `.task_owner` as a bare task ID (`writeEnvRootOwner(envRoot, taskID)` → `os.WriteFile(path, []byte(taskID))`). Those roots are exactly what an upgrade carries forward, and the sweep was skipping all of them.

**The rule now:** if the owner record names a workspace, an exact match is still required. If it names none — task-ID-only marker, or no marker at all — the directory name is the evidence and has to *be* the workspace ID, which is what that era's layout produced. The readable layout always appends a suffix (`<slug>-<id tail>`), so a bare workspace UUID directory can only be one of the older roots and the fallback cannot alias a modern one. Anything unattributable is still skipped, and the daemon hook-signature check still gates every write.

**Regressions** (`TestPersistCoAuthoredByStateReconcilesLegacyEnvRoots`), verified failing without the fallback:

- `v0.4.31`-style env root with **no** `.task_owner` → migrated.
- `v0.4.34`-style env root with a **bare task-ID** marker → migrated.
- Both shapes under *another* workspace's legacy directory → untouched.
- A readable-layout root whose marker names no workspace → untouched (unattributable, not guessed).

Verification on `85d238e4b`: `repocache` fully green; `./internal/daemon/` green except the 7 pre-existing `identity`/`config`/`Prepare*Home` failures that fail identically on `main` here; `-race -count=5` over the Co-authored-by, tracked-refresh and reconnect tests; `go vet ./internal/daemon/...` and `git diff --check` clean.

PR #7178, opened 2026-08-18
feat(desktop): add tab cycling shortcuts

robertn702:
## What does this PR do?

Adds desktop-only tab cycling with the conventional Control+Tab bindings while preserving macOS bracket aliases:

- Next tab on macOS, Windows, and Linux: `Control+Tab`
- Previous tab on macOS, Windows, and Linux: `Control+Shift+Tab`
- macOS next-tab alias: `Shift+Command+]`
- macOS previous-tab alias: `Shift+Command+[` 

Cycling follows the visible tab-strip order, wraps at both ends, updates the existing recent-tab history used when closing tabs, and leaves each tab's navigation history unchanged. The bindings are fixed desktop renderer shortcuts, so the web runtime and shared rebindable shortcut registry remain unchanged.

This PR is intentionally left in draft for Robert's review.

## Related Issue

Closes #6938

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- Added visual-order tab cycling with wraparound to `apps/desktop/src/renderer/src/stores/tab-store.ts`.
- Added exact desktop bindings and guards in `apps/desktop/src/renderer/src/hooks/use-tab-history.ts`.
- Preserved dedicated issue-window, overlay, editable-target, repeat, prevented-event, and IME guards.
- Kept existing unshifted Command/Ctrl bracket navigation-history shortcuts unchanged.
- Added platform-specific fixed shortcut rows to Keyboard Shortcuts settings.
- Preserved the numbered-tab shortcut rows added to upstream while resolving the September 4 refresh conflicts.
- Added Previous tab and Next tab labels in English, Simplified Chinese, Japanese, and Korean.
- Added focused store, keyboard, settings, and locale tests.
- Rebased the single feature commit onto current upstream `main` at `ca47495fc`; no unrelated surface-bridge changes are present.

## How to Test

```bash
corepack pnpm --filter @multica/desktop exec vitest run \
  src/renderer/src/stores/tab-store.test.ts \
  src/renderer/src/hooks/use-navigation-input-bindings.test.tsx

corepack pnpm --filter @multica/views exec vitest run \
  settings/components/keyboard-shortcuts-tab.test.tsx \
  locales/parity.test.ts

corepack pnpm --filter @multica/desktop typecheck
[...]

Issue #6938, opened 2026-08-13
[Feature]: Keyboard shortcuts to switch to the next and previous tab


robertn702:
### Deployment type

Official App

### What do you want and why?

When working across several open Multica tabs, switching between them requires reaching for the tab strip. This interrupts keyboard-focused workflows, particularly when monitoring or moving between multiple issues.

Add keyboard shortcuts to activate the next and previous **Multica tab**. This should work in the desktop app and web app where the application manages its own tab strip.

### Proposed solution (optional)

Provide standard directional shortcuts:

- Next tab: `Cmd+Shift+]`
- Previous tab: `Cmd+Shift+[`

Expected behavior:

- Cycle through currently open Multica tabs, wrapping at either end.
- Keep the shortcut scoped to Multica's tab strip rather than browser tabs.
- Do not interfere with typing in text inputs or editors except where the platform convention requires the shortcut.
- Preserve the current active-tab behavior and accessibility focus management.

Bohan-J:
Worth checking the native browser bindings here, because they overlap with the proposal and constrain what the **web app** can implement.

Browsers already bind next/previous tab — and they already wrap at both ends:

| | Next tab | Previous tab |
|---|---|---|
| Chrome / Edge (Win/Linux) | `Ctrl+Tab`, `Ctrl+PgDn` | `Ctrl+Shift+Tab`, `Ctrl+PgUp` |
| Chrome (macOS) | `Ctrl+Tab`, `Cmd+Option+→` | `Ctrl+Shift+Tab`, `Cmd+Option+←` |
| Safari (macOS) | `Ctrl+Tab`, **`Shift+Cmd+]`** | `Ctrl+Shift+Tab`, **`Shift+Cmd+[`** |

Plus `Ctrl/Cmd+1`–`8` for the Nth tab and `Ctrl/Cmd+9` for the last one.

Two consequences:

**1. The proposed `Cmd+Shift+]` / `Cmd+Shift+[` is Safari's own next/previous-tab binding** ([Apple's Safari shortcut docs](https://support.apple.com/guide/safari/keyboard-and-other-shortcuts-cpsh003/mac)). In the web app on Safari, that keystroke switches the *browser* tab, not the Multica tab.

[...]

PR #7870, opened 2026-09-01
MUL-6923: add numbered desktop tab shortcuts

ASDFGHoney:
## Summary

- intercept fixed `Cmd/Ctrl+1..9` accelerators in Electron main so they work from any focus context
- select tabs 1–8 by current strip order and always map 9 to the final tab without mutating session history or mementos
- reserve the new fixed chords from configurable product shortcuts and validate the IPC payload at the preload boundary

## Testing

- `pnpm --filter @multica/desktop exec vitest run src/renderer/src/stores/tab-store.test.ts src/renderer/src/components/tab-bar.test.tsx src/renderer/src/hooks/use-tab-selection-shortcut.test.tsx src/main/keyboard-shortcuts.test.ts` (167 passed)
- `pnpm --filter @multica/desktop exec vitest run --exclude src/main/runtime-config-loader.test.ts` (610 passed)
- `pnpm --filter @multica/core exec vitest run shortcuts/definitions.test.ts shortcuts/store.test.ts` (36 passed)
- `pnpm --filter @multica/desktop typecheck && pnpm --filter @multica/core typecheck`
- `pnpm --filter @multica/desktop lint && pnpm --filter @multica/core lint` (one existing `tab-content.tsx` hook-dependency warning)

The unfiltered desktop suite cannot collect `runtime-config-loader.test.ts` in this checkout because the installed Electron package lacks its binary. The full core sweep reached 1,671 passing tests but has 10 unrelated failures where this local Node 25 environment exposes an unavailable `localStorage`; affected shortcut suites pass independently.


Bohan-J:
May I ask why you close this PR? Cuz I think the direction could be accepted

ASDFGHoney:
Thanks for checking! I closed this because I saw the discussion in #6938 about next/previous tab shortcuts and thought that approach would also be sufficient for my use case. But if numbered tab shortcuts would still be welcome, I’d be happy to continue with this PR. I’ve reopened it—thanks!

Bohan-J (changes requested):
Thanks for reopening — and to answer your question directly: yes, numbered tab shortcuts are still welcome. #7178 (`Ctrl+Tab` / `Ctrl+Shift+Tab` cycling) and this PR are complementary rather than redundant: cycling walks the strip relative to where you already are, numbered chords jump to an absolute position. Browsers ship both, and we want both here.

The design is a clean extension of what already exists: the main process owns the chord so it fires from any focus context, it reuses the `dispatchToMainRenderer` queue, the payload is revalidated at the preload boundary, and 1–8 by strip order / 9 as last tab matches browser semantics. Routing from a standalone issue window back to the main window is intended behavior on our side — it matches the existing `Cmd/Ctrl+,` path, so please keep it.

Two things need to change before this can merge.

## Must-fix 1 — the new fixed chords never appear in Settings → Keyboard shortcuts

`packages/core/shortcuts/definitions.ts:320` adds `1`–`9` to `PRIMARY_RESERVED_KEYS`. That has a side effect beyond rejecting new bindings: `sanitizeShortcutOverrides` (`packages/core/shortcuts/store.ts:56`) re-runs `isShortcutAllowedForAction` on every load, so an override a user has **already** persisted on `Cmd/Ctrl+<digit>` is silently discarded on upgrade. Meanwhile `packages/views/settings/components/keyboard-shortcuts-tab.tsx` still lists only `Cmd+,`, `Cmd+W`, zoom and `Escape` under "Fixed system shortcuts".

For a user who had customized one of those chords, the net effect is: the binding vanishes with no explanation, and the settings page offers no way to discover what the key does now.

The repo already has a convention here — #7022 shipped `Cmd/Ctrl+,` together with its row in this list and the four locale files. Please do the same:

- add rows to the fixed list in `keyboard-shortcuts-tab.tsx` (e.g. "Select tab 1–8" and "Select last tab")
- add the strings under `shortcuts.fixed` in `packages/views/locales/{en,ja,ko,zh-Hans}/settings.json`
- extend `packages/views/settings/components/keyboard-shortcuts-tab.test.tsx`

## Must-fix 2 — the `!input.shift` guard excludes entire keyboard layouts

`apps/desktop/src/main/keyboard-shortcuts.ts:80`:

```ts
if (!input.shift && /^[1-9]$/.test(input.key)) {
```

Matching on `input.key` while rejecting Shift makes the feature unreachable on layouts where the number row requires Shift. French AZERTY is the common case: unshifted, that row produces `& é " ' ( - è _ ç`, which never matches the regex; with Shift held, `input.key` *is* `1`–`9`, but the guard rejects it. Both paths fail, leaving only the numpad. Chrome works on those layouts because it keys off physical position.

The cleanest fix is to match `input.code` (`Digit1`–`Digit9`, plus `Numpad1`–`Numpad9`) instead of `input.key` — Electron's `before-input-event` input already carries `code`, it just isn't part of the `ShortcutInput` type yet. Simply dropping the Shift guard is a smaller alternative and is a no-op on QWERTY, where `Shift`+digit yields `!@#…` and never matches `/^[1-9]$/`.

Either way, please drop or re-target the `rejects Shift and secondary modifiers` case in `keyboard-shortcuts.test.ts` — as written it locks the layout gap in. If you'd rather keep the Shift rejection deliberately, that's fine too, but then it needs a comment giving the reason.

## Everything else checked out

Payload validation (rejects `0`/`10`/`1.5`/strings), auto-repeat swallowing, missing-position no-op, and tab ordering against the strip render order are all correct, and the three-layer test coverage is good. No other correctness blockers.

Minor nit: `Closes MULT-2` in the description doesn't resolve to anything on GitHub — worth cleaning up.

ASDFGHoney:
Thanks for the thorough review. Addressed both blockers in `32bc14eac`:

- added “Select tab 1–8” and “Select last tab” to Fixed system shortcuts, with all four locales and Settings coverage;
- switched digit detection to physical `input.code` values (`Digit1`–`Digit9` and `Numpad1`–`Numpad9`), with AZERTY, modifier, numpad, zoom-conflict, and auto-repeat regression coverage;
- removed the stray `Closes MULT-2` from the PR description.

Validation on this head:

- Desktop suite excluding the existing Electron-binary-dependent `runtime-config-loader.test.ts`: 621/621 passed;
- affected Views + locale parity suites: 169/169 passed;
- Core shortcut suites: 36/36 passed;
- Desktop / Views / Core typechecks passed, lint reported 0 errors, and `git diff --check` passed.

The full Views sweep reached 4,932/4,933; the sole failure is the existing `sidebar-resize.test.tsx` incompatibility with this checkout's Node 26 global `localStorage` behavior, while every changed suite passes. Ready for re-review.

ASDFGHoney:
One follow-up correction to my previous update after cross-checking the main-process interception against the Settings recorder and persisted shortcut model.

I initially took the physical `input.code` option. However, Settings records `KeyboardEvent.key`, and persisted chords do not retain a physical code. On French AZERTY, that allowed the same physical `Digit1` keypress to be saved as `Ctrl+&` while Electron main consumed it as tab 1, so a shortcut that Settings considered valid could never reach the renderer. Existing `Ctrl+&` overrides had the same problem.

In `8ec4e4c9c`, I switched to the smaller alternative suggested in the review: keep matching the logical `input.key`, but allow Shift when a layout needs it to produce `1`–`9`.

- AZERTY `Ctrl+&` is no longer intercepted and remains usable as a configured shortcut.
- AZERTY `Ctrl+Shift+1` selects tab 1 because the logical key is `1`.
- QWERTY `Ctrl+Shift+Digit1` reports `!`, so it remains a distinct chord.
- Numpad and remapped keys follow their logical value, keeping Electron main consistent with recording, reservation, and persisted sanitization.

Regression coverage now spans the main-process policy, Core reservation/sanitization, and the Settings recorder. On the pushed head, the focused suites pass (Desktop 59, Core 38, Views 10); the full Node 22 verification also passed with Desktop 622 tests (excluding only the existing Electron-binary-dependent `runtime-config-loader` suite), Core 1,697, Views 4,934, all three package typechecks, and lint with 0 errors.

This is now ready for re-review.

Bohan-J (approved):
Thanks for the quick turnaround, and for sticking with this after the earlier close — both must-fixes are resolved and this looks good to merge.

- **Fixed shortcut discoverability:** the "Fixed system shortcuts" list now carries "Select tab 1–8" and "Select last tab", with all four locales and a test for the rows. Users who lose a custom digit binding on upgrade now have somewhere to find out why.
- **Keyboard layout support:** dropping the Shift guard was the alternative I offered and you documented the reasoning in place, which is exactly what I asked for. Threading `code` through `ShortcutInput` and using it to model an AZERTY `Ctrl+&` / `Ctrl+Shift+1` in the tests makes the intent legible.

One known edge we're accepting rather than blocking on: on macOS layouts where the number row requires Shift, the chords for tabs 3–5 land on `Shift-Cmd-3/4/5`, which the system claims for screenshots. Standard QWERTY — the overwhelming majority case — is unaffected, and this is worth a follow-up rather than more churn here.

CI is green. Merging now — thanks again for the contribution.

PR #7883, opened 2026-09-01
MUL-6935: fix(execenv): recognize OpenClaw 2026.8.x 'Unknown config path' wording

doubletree6:
## Problem

OpenClaw `2026.8.x` reworded the missing-config-path error for `openclaw config get agents.list --json`:

- **2026.6.x**: `Config path not found: agents.list`
- **2026.8.x**: `Unknown config path: agents.list. Run openclaw config schema to inspect valid paths.`

`isOpenclawKeyMissingMessage` only matched the older phrasing (`"no value at "`, `"not set"`, `"missing key"`, `"path not found"`). Because `unknown config path` is not in that set, `isOpenclawKeyMissingResult` returned `false`, so the registry fallback in `openclawResolvedAgentsList` never engaged. Result: every OpenClaw `2026.8.x` runtime failed closed during execenv preparation with:

```
read openclaw agents.list: openclaw config get agents.list --json: exit status 1
```

## Second bug

OpenClaw `2026.8.x` also changed the JSON error envelope's `error` field from a **plain string** to an **object carrying `message`**:

```json
{"ok":false,"error":{"type":"cli_error","message":"Unknown config path: agents.list. Run openclaw config schema to inspect valid paths."}}
```

`openclawJSONErrorMessage` unmarshalled `error` into a `string` field, so `json.Unmarshal` failed on the object shape and silently dropped the message even when the exit path looked for it.

## Fix

1. `isOpenclawKeyMissingMessage`: also match `unknown config path` so the graceful registry fallback engages instead of failing closed.
2. `openclawJSONErrorMessage`: accept `error` as either a flat string **or** an object with a `message` field (via `json.RawMessage`).

Both changes are backward-compatible with pre-2026.6 and 2026.6.x phrasing.

## Tests

Added regression cases covering the `2026.8.x` wording to:
- `TestIsOpenclawKeyMissing` (stderr path)
- `TestIsOpenclawKeyMissingResult` (JSON envelope with nested `error.message`)

## Evidence

Verified against a real OpenClaw `2026.8.1` (`ea80657`) install:

- `openclaw config get agents.list --json` returns `{"ok":false,"error":{"type":"cli_error","message":"Unknown config path: agents.list. Run openclaw config schema to inspect valid paths."}}`, exit 1
- `openclaw agents list --json` returns `[{"id":...,"identityName":...,"workspace":...,"model":...}]`, exit 0 (the registry fallback target)

The registry fallback path already existed (`openclawRegistryAgentsList`); these changes restore the intended graceful switch to it.

### DIFF
--- a/server/internal/daemon/execenv/openclaw_config.go
+++ b/server/internal/daemon/execenv/openclaw_config.go
@@ -1443,14 +1443,40 @@ func isOpenclawKeyMissing(err error) bool {
 const openclawJSONErrorMaxRunes = 1024
 
 func openclawJSONErrorMessage(stdout string) (string, bool) {
-	var envelope struct {
-		Error string `json:"error"`
-	}
-	if json.Unmarshal([]byte(strings.TrimSpace(stdout)), &envelope) != nil {
+	// Older OpenClaw builds emitted a flat `{"error":"..."}` envelope. From
+	// 2026.8.x the `error` field became an object with a `message` field
+	// (e.g. {"ok":false,"error":{"type":"cli_error","message":"..."}}).
+	// Unmarshalling a non-string into a string field makes json.Unmarshal
+	// return an error, which would silently drop the message and reclassify a
+	// graceful missing-key fallback as a hard failure. Accept both shapes.
+	trimmed := strings.TrimSpace(stdout)
+	if trimmed == "" {
 		return "", false
 	}
-	message := strings.Join(strings.Fields(envelope.Error), " ")
-	return message, message != ""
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks @doubletree6 for flagging the new wording. Recognition of `Unknown config path: agents.list` shipped in #8678, which also reads the `agents.entries` path that 2026.8+ moved the agents to, and it has now been merged (`d2423aa`). Closing this one as superseded. Thanks for the contribution!

PR #7889, opened 2026-09-01
MUL-6939 fix(util): preserve JSON keys during NUL sanitization

### LINKED ISSUE #7888: [Bug]: PostgreSQL JSON sanitizer can silently drop colliding object keys
### Deployment type

self-host

### Area

Agents and runtimes

### What happened?

`internal/util.SanitizeJSONForPostgres` sanitizes every decoded JSON object key independently before inserting it into a new Go map. Two distinct input keys can therefore normalize to the same PostgreSQL-safe key and one value is silently overwritten.

For example, `"tool"` and `"tool\u0000"` both become `"tool"`. The sanitized object contains one entry instead of two. This is a data-integrity issue on the existing defensive path for malformed agent/runtime payloads.

I expected sanitization to preserve every object entry while still removing PostgreSQL-incompatible NUL bytes.

### Steps to reproduce

1. Call `SanitizeJSONForPostgres` with `map[string]any{"tool": "unchanged", "tool\x00": "contained a NUL"}`.
2. Assert that the returned map has two entries.
3. On current `main`, the returned map has only one `tool` entry because map assignment overwrites the first value.

### Proposed behavior

Reserve unchanged key spellings first, then assign transformed collisions deterministic suffixes such as `#2`, `#3`, and so on. Existing suffix keys must also be respected. Non-colliding payloads should keep their current output.

### Additional context

The bug is deterministic at the normalization boundary, but ordinary Go map iteration makes which value survives unspecified. A regression test should cover both a direct collision and an already-occupied suffix.

eveningb123:
## What does this PR do?

Prevents `SanitizeJSONForPostgres` from silently dropping decoded JSON object entries when distinct keys collide after PostgreSQL NUL sanitization.

Unchanged keys reserve their original spelling. Transformed keys are processed deterministically and receive the first free `#2`, `#3`, ... suffix when needed. The allocator remembers the next candidate per sanitized base, so collision handling is amortized linear after the existing key sort. Non-colliding keys keep the existing output.

### Thinking path

The PostgreSQL defense already recursively sanitizes decoded JSON values, but its map branch writes each normalized key directly into a new Go map. That makes normalization non-injective: `"tool"` and `"tool\u0000"` both target `"tool"`. Preserving the existing recursive contract while allocating collision-free output keys is the smallest change that removes the data-loss path. Sorting raw keys makes the allocation deterministic, and reserving unchanged keys avoids renaming valid input because of malformed input.

Risk is limited to malformed objects whose keys collide after sanitization. Those objects now retain data under suffixed keys instead of losing one value. No database schema or API shape changes are involved.

## Related Issue

Closes #7888

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Update `server/internal/util/text.go` to allocate stable, collision-free sanitized object keys.
- Add regressions for a direct NUL collision, reserved numeric suffixes, deterministic multi-collision assignment across insertion orders, and linear allocation growth in `server/internal/util/text_postgres_test.go`.

## How to Test

1. `cd server && go test ./internal/util ./internal/service ./internal/handler -count=1`
2. `cd server && go vet ./internal/util`

## Checklist

- [x] This change does not affect the UI, so screenshots are not applicable
- [x] This internal defensive fix does not require documentation changes
- [x] This change does not add a runtime, coding tool, or UI tab
- [x] This change does not touch Chinese product copy

## AI Disclosure

**Prompt / approach:** Inspected the current sanitizer and callers, reproduced the key-collision loss with a failing regression test, implemented deterministic collision-free mapping, then followed review feedback by reproducing the quadratic allocation growth, retaining the next suffix per sanitized base, and pinning multi-collision behavior across insertion orders and reserved suffixes. Ran related package tests plus `go vet`.

## Screenshots (optional)

Not applicable.

### DIFF
--- a/server/internal/util/text.go
+++ b/server/internal/util/text.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"fmt"
+	"sort"
 	"strings"
 	"unicode/utf8"
 )
@@ -109,8 +111,8 @@ func sanitizeJSONValue(v any, depth int) any {
 		return SanitizeTextForPostgres(t)
 	case map[string]any:
 		out := make(map[string]any, len(t))
-		for k, val := range t {
-			out[SanitizeTextForPostgres(k)] = sanitizeJSONValue(val, depth+1)
+		for k, safeKey := range sanitizePostgresJSONMapKeys(t) {
+			out[safeKey] = sanitizeJSONValue(t[k], depth+1)
 		}
 		return out
 	case []any:
@@ -123,4 +125,43 @@ func sanitizeJSONValue(v any, depth int) any {
 		// Numbers, bools, nil — nothing a TEXT/JSONB column can choke on.
 		return v
 	}
+}
+
+// sanitizePostgresJSONMapKeys returns a stable one-to-one mapping from raw
+// object keys to PostgreSQL-safe keys. Unchanged keys reserve their spelling
+// first; transformed keys that would collide receive the first free numeric
+// suffix instead of silently overwriting another entry.
+func sanitizePostgresJSONMapKeys(values map[string]any) map[string]string {
+	keys := make([]string, 0, len(values))
+	for key := range values {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	mapped := make(map[string]string, len(keys))
+	used := make(map[string]struct{}, len(keys))
+	changed := make([]string, 0, len(keys))
+	for _, key := range keys {
+		candidate := SanitizeTextForPostgres(key)
+		if candidate == key {
+			mapped[key] = key
+			used[key] = struct{}{}
+			continue
+		}
+		changed = append(changed, key)
+	}
+
+	for _, key := range changed {
+		base := SanitizeTextForPostgres(key)
+		candidate := base
+		for suffix := 2; ; suffix++ {
+			if _, exists := used[candidate]; !exists {
+				break
+			}
+			candidate = fmt.Sprintf("%s#%d", base, suffix)
+		}
+		mapped[key] = candidate
+		used[candidate] = struct{}{}
+	}
+	return mapped
 }
--- a/server/internal/util/text.go
+++ b/server/internal/util/text.go
@@ -140,6 +140,7 @@ func sanitizePostgresJSONMapKeys(values map[string]any) map[string]string {
 
 	mapped := make(map[string]string, len(keys))
 	used := make(map[string]struct{}, len(keys))
+	nextSuffix := make(map[string]int)
 	changed := make([]string, 0, len(keys))
 	for _, key := range keys {
 		candidate := SanitizeTextForPostgres(key)
@@ -154,11 +155,19 @@ func sanitizePostgresJSONMapKeys(values map[string]any) map[string]string {
 	for _, key := range changed {
 		base := SanitizeTextForPostgres(key)
 		candidate := base
-		for suffix := 2; ; suffix++ {
-			if _, exists := used[candidate]; !exists {
-				break
+		if _, exists := used[candidate]; exists {
+			suffix := nextSuffix[base]
+			if suffix == 0 {
+				suffix = 2
 			}
-			candidate = fmt.Sprintf("%s#%d", base, suffix)
+			for {
+				candidate = fmt.Sprintf("%s#%d", base, suffix)
+				if _, exists := used[candidate]; !exists {
+					nextSuffix[base] = suffix + 1
+					break
+				}
+				suffix++
+			}
 		}
 		mapped[key] = candidate
 		used[candidate] = struct{}{}
--- a/server/internal/util/text_postgres_test.go
+++ b/server/internal/util/text_postgres_test.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"encoding/json"
+	"reflect"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


multica-eve (changes requested):
When an agent reports two distinct input names that become identical after invisible characters are removed, users should retain both entries in the execution record. Preserving the original valid name and giving the conflicting entry a stable alternate name is the right product behavior; ordinary records must remain unchanged.

**Product verdict: the direction should be accepted.** The current implementation still needs changes before merge.

### Required changes

1. **Make suffix allocation amortized linear.**

   In `server/internal/util/text.go:157`, every colliding key restarts its search at `#2`. For `k` distinct raw keys that sanitize to the same base, this performs `1 + 2 + ... + k` membership checks: `O(k²)`. This helper processes externally produced, potentially pathological payloads, so the worst case is part of the correctness contract. A local probe took approximately 0.53s for 4,096 colliding keys, 2.00s for 8,192, and 8.16s for 16,384.

   Please keep the good parts of the design—sorted raw keys, unchanged-key reservation, deterministic output—but track the next candidate suffix per sanitized base (or group by sanitized base first), so an already-rejected suffix is never rescanned for later keys. The allocation phase should be `O(n)` amortized; with the existing sort, the helper should remain `O(n log n)` overall.

2. **Pin the deterministic multi-collision contract in tests.**

   The current new tests each contain only one transformed key. Please add coverage where multiple distinct transformed keys sanitize to the same base, including pre-existing `#2`/`#3` keys and different insertion orders. Assert that all values survive, unchanged keys keep their spellings, transformed-key assignment is deterministic, and the output contains no NUL. Avoid a wall-clock assertion; a correctness case with a larger collision set or a benchmark is preferable to a flaky timing test.

### Boundary hardening

`ReportTaskMessages` currently decodes an unbounded request body. A request-size cap would be useful defense in depth, but this endpoint legitimately carries large tool output, so please do not invent an arbitrary limit in this patch. If you include a cap, justify it against the existing payload contract and return an explicit oversized-request response; otherwise keep this PR focused and track that hardening separately.

Please update the implementation and tests, run the relevant Go tests and vet, and re-request review. Once the quadratic allocation is removed and the multi-collision behavior is pinned, this approach is suitable for merge.

eveningb123:
Addressed in 01e94bf69.

- Suffix allocation now retains the next candidate per sanitized base, so rejected suffixes are not rescanned and allocation is amortized linear after sorting.
- Added deterministic multi-collision coverage with reserved `#2`/`#3` keys, forward/reverse insertion order, value preservation, unchanged-key preservation, and NUL-free output.
- Added a non-wall-clock allocation-scaling regression; before the fix, doubling 128 to 256 colliding keys grew allocations from 16,519 to 65,803.
- Kept `ReportTaskMessages` request-size hardening out of this focused patch.

Local verification:
- `go test ./internal/util ./internal/service ./internal/handler -count=1`
- `go vet ./internal/util ./internal/service ./internal/handler`
- `git diff --check`

@multica-eve ready for re-review.

multica-eve (approved):
Users now retain every execution-input entry when cleaned names collide, while valid names keep their original spelling and ordinary records remain unchanged. This is the intended product behavior and closes the silent data-loss path without adding a new user-facing mechanism.

**Product verdict: accept.** I independently verified the original loss, this remains an extension of the existing persistence guard, and the solution now matches the approach we would use: deterministic disambiguation with valid-key priority and bounded allocation work.

The follow-up commit `01e94bf69` resolves the previous blockers:

- `nextSuffix` retains the next candidate per sanitized base, so rejected suffixes are not rescanned. The allocation phase is amortized linear and the existing sort keeps the full helper at `O(n log n)`.
- The new multi-collision test pins reserved `#2`/`#3` keys, multiple transformed keys, forward/reverse insertion order, value preservation, unchanged-key preservation, and NUL-free output.
- The allocation-scaling regression is non-wall-clock and fails the previous quadratic implementation.

Independent validation on the new head:

- `go test ./internal/util -count=1` — passed.
- Sanitizer and allocation regressions with `-count=20` — passed.
- `go test ./internal/service ./internal/handler -run '^$' -count=1` — package compilation passed.
- `go vet ./internal/util ./internal/service ./internal/handler` — passed.
- `git diff --check` — passed.
- A local probe with 4,096 / 8,192 / 16,384 colliding keys completed in approximately 6.1ms / 10.0ms / 14.5ms, versus 0.53s / 2.00s / 8.16s before the fix.
- GitHub CI on `01e94bf69` is green, including backend tests.

Keeping request-size hardening out of this focused patch is acceptable: the quadratic amplification introduced by the first revision is gone, and choosing a cap for legitimately large tool output requires separate evidence rather than an arbitrary limit.

If this change is wrong, it will surface as missing or misnamed entries in task transcripts; rollback is a single backend revert with no migration, and only malformed colliding inputs are affected.

**Merge recommendation: Approve.** The product behavior is desirable, the prior correctness/performance blockers are resolved, and no new blocking issue remains.

Bohan-J:
Thank you @eveningb123 for the careful fix and the quick turnaround on review feedback — first reproducing the collision data loss with a regression test, then landing the amortized-linear allocator with tests that genuinely lock both the determinism and the complexity contract. We verified the fix independently on our side (algorithm trace, allocation-scaling test against the old implementation, full CI). Merging — much appreciated!

Issue #6859, opened 2026-08-12
[Feature]: Add phase-level task latency telemetry and a time-to-first-business-action metric


kinpoe-ray:
## Problem

When an agent run feels slow, Multica currently exposes broad lifecycle timestamps and terminal duration, but not enough phase data to tell whether the delay came from:

- queue pickup / WebSocket fallback;
- claim payload hydration;
- local-directory lock wait;
- skill and execution-environment preparation;
- repository checkout or cache refresh;
- provider process startup and protocol handshake;
- mandatory Multica control-plane tool calls;
- or the actual user task.

That makes performance reports hard to act on. A slow run may lead to a provider-specific fix, a prompt reduction, a repository fast path, or a scheduling change, but today these cases look similar from the run UI.

This is adjacent to, but narrower than:

- #1978: a user-visible slow-response report;
- #1549: persistent gateway mode for OpenClaw cold starts;
- #4798: live output streaming;
- #6144: startup prompt size and prompt-prefix caching.

The missing prerequisite is a provider-agnostic latency breakdown and a metric for when useful task work actually begins.

## Small observed sample

In five historical self-hosted Codex assignment runs from one environment:

| metric | observed range |
|---|---:|
| queue wait | 0 s |
| prepare | 1–2 s |
| first visible output | 43–57 s |
| total runtime | 10m26s–17m02s |
| tool calls | 27–55 |

Four runs overlapped, and the project had no bound code resource. This sample is not proof of a single root cause, and current `main` includes later prompt/cache improvements. It does show why total duration alone is insufficient: queue and prepare were short while the model/tool phase dominated.

### 1. Record stable phase timestamps

Persist or emit structured timings for at least:

1. task created;
2. claim requested / dispatched;
3. daemon picked task;
4. local resource wait started / ended;
5. execution-environment prepare started / ended;
6. provider spawn started;
7. provider protocol ready (for example, Codex `initialize` plus thread start);
8. first provider event;
9. first platform control action;
10. first business action;
11. first visible user output;
12. terminal completion.

Use monotonic durations within the daemon and wall-clock timestamps only where correlation across server and daemon is required.

### 2. Define “first business action”

Track the first action that advances the user's actual task, separate from Multica bookkeeping. Examples:

- first non-Multica tool call;
- first repository inspection/edit/test command;
- or first substantive assistant output for a text-only task.

[...]

PR #7890, opened 2026-09-01
feat(daemon): record task phase latency

eveningb123:
## What does this PR do?

Adds provider-agnostic, structured daemon logs for task phase latency so operators can distinguish preparation delays from runtime startup, first output, first tool use, and completion.

Each phase record contains `task_phase`, `phase_elapsed_ms`, `total_elapsed_ms`, `task_id`, and `runtime_id`. Durations use the daemon's monotonic clock, and duplicate marks are suppressed without affecting task behavior. `phase_elapsed_ms` measures time since the previous recorded phase, so a skipped phase widens the next interval.

### Thinking path

Current `main` already has wall-clock/idle watchdogs and an overall terminal duration, so another timeout would duplicate existing liveness protection. The missing diagnostic is where elapsed time accumulates. The daemon lifecycle already has stable boundaries for claim, preparation, skill availability, environment readiness, successful backend start, message drain, and completion. A task-scoped recorder carried through `context.Context` instruments those boundaries without widening the existing runner/backend method signatures or changing the server state machine.

This is intentionally an incremental observability slice of #6859: structured daemon logs only, with no persistence or UI in this PR. Risk is limited to several additional info-level log records per task; fields contain identifiers and durations, not prompts, paths, tool arguments, or issue content.

## Related Issue

Related to #6859

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add a concurrency-safe, duplicate-suppressing task phase recorder in `server/internal/daemon/task_phase_timing.go`.
- Record claim, prepare, skills-ready, environment-ready, runtime-started, first-output-received, first-tool-use, turn-completed, and finished boundaries in `server/internal/daemon/daemon.go`.
- Record task-level `turn_completed` only after any rejected-resume fallback has finished and its result has been reconciled.
- Add duration, stable-field, message-classification, drain-order, and task-boundary tests in `server/internal/daemon/task_phase_timing_test.go`.

## How to Test

1. `cd server && go test ./internal/daemon -count=1`
2. `cd server && go vet ./internal/daemon`

## Checklist

- [x] This change does not affect the UI, so screenshots are not applicable
- [x] The structured field names and phase definitions are documented in code and this PR
- [x] This change does not add a runtime, coding tool, or UI tab
- [x] This change does not touch Chinese product copy

## AI Disclosure

**Prompt / approach:** Compared the self-hosted diagnostic gap with current upstream `main`, verified that liveness watchdogs already exist, then used test-driven development to add a diagnostic-only phase recorder and wire it into stable daemon lifecycle boundaries. Review follow-up added a rejected-resume fallback regression, moved task completion after final result reconciliation, and renamed the provider-receipt timestamp so it does not imply successful member delivery. Ran the full daemon package test suite and `go vet`.

## Screenshots (optional)

Not applicable.


server/internal/daemon/daemon.go
```diff
@@ -8908,6 +8924,7 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 	select {
 	case result := <-session.Result:
 		waitForDrain()
+		phaseRecorder.Mark(taskPhaseTurnCompleted)
```
Bohan-J: This is inside `executeAndDrain`, which `runTask` can invoke twice — the resume attempt at `daemon.go:8101` and the fresh-session retry at `daemon.go:8157`. Both share the same `ctx`, so both resolve the same recorder, and `Mark` suppresses a repeated phase. On a resume-rejected task this line therefore fires for the attempt that gets discarded, while the real work is still ahead: the retry's `runtime_started`, `first_visible_output`, `first_tool_use` and `turn_completed` are all swallowed, and its whole duration folds into the gap before `finished`.

Suggest moving the task-level `turn_completed` mark to after `reconcileFreshRetryResult`, and — if per-attempt timings are worth keeping — emitting attempt-scoped events with an attempt index rather than reusing the task-level phase names. A regression test for resume-rejected → fresh-retry-succeeds would lock this down.

server/internal/daemon/daemon.go
```diff
@@ -8743,6 +8753,12 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 				if !ok {
 					goto drainDone
 				}
+				if isTaskVisibleOutput(msg) {
+					phaseRecorder.Mark(taskPhaseFirstVisibleOutput)
```
Bohan-J: This marks when the message lands in the drain loop, which is earlier than "visible": the message still has to be batched, wait for the 500 ms ticker flush, and survive `ReportTaskMessages` (see the flush at `daemon.go:8697-8728`, where a failed report is only a debug log). If that report fails the member never sees anything, but the phase record still claims first visible output happened — and text/thinking are buffered until a flush even on the happy path.

The measurement is useful, the name just overstates it. Either rename to `first_provider_event` / `first_output_received` and keep the mark here, or move the mark after the first successful `ReportTaskMessages` and add a report-failure test. I'd prefer the rename — received-time is the better signal for separating provider latency from daemon-side delay.

Bohan-J (changes requested):
Thanks for this — the thinking path and the AI disclosure made the PR much easier to review, and the choice to instrument existing lifecycle boundaries instead of widening runner/backend signatures is the right shape.

## Product direction: accept

Today a self-hosted operator can only see a task's total duration, so "this task is slow" is indistinguishable from "workdir preparation is slow", "the runtime took a long time to start", and "the provider hasn't replied yet". Splitting the lifecycle into phases closes that loop with no new storage, endpoint, or user-facing surface, and shipping it first as daemon-local structured logs is the same order we would pick ourselves. The "related to #6859, does not close it" framing is correct and worth keeping — these records stay on one machine, so cross-machine aggregation and Run details are still open.

Two things need to change before merge. Both are about a phase meaning something other than what its name promises, not about the feature being wrong.

## 1. The fresh-session retry loses all of its phases, and `turn_completed` lands on the discarded attempt

`runTask` can call `executeAndDrain` twice: the resume attempt at `daemon.go:8101`, then the fresh-session retry at `daemon.go:8157`. Both get the same `ctx`, so both resolve the same recorder, and `Mark` suppresses a repeated phase (`task_phase_timing.go:68`). On any resume-rejected task the second attempt therefore records nothing — no `runtime_started`, no `first_visible_output`, no `first_tool_use`, no `turn_completed`.

Two consequences, both landing squarely on the "why was this task slow?" case this feature exists for:

- `turn_completed` is emitted when the *abandoned* attempt ends, while the task is still running. An operator reading the log live sees the turn finish and the task keep going.
- The retry's entire duration silently folds into the gap before `finished`, so the phase most worth measuring is exactly the one that disappears.

Suggested fix: mark the task-level `turn_completed` after `reconcileFreshRetryResult` rather than inside `executeAndDrain`. If per-attempt data is worth keeping, add attempt-scoped events carrying an attempt index instead of reusing the task-level phase names. Please add a regression test for resume-rejected → fresh-retry-succeeds.

## 2. `first_visible_output` measures "the daemon received an event", not "the member can see it"

The mark at `daemon.go:8757` fires as the message comes off `session.Messages` — before batching, before the 500 ms ticker flush, and before `ReportTaskMessages` succeeds (`daemon.go:8697-8728`). When that report fails the daemon only logs at debug level and nothing ever reaches the member, but the phase record still says first visible output happened. Text and thinking are additionally buffered until a flush, so the two timestamps differ even on the happy path.

Received-time is a genuinely useful metric — it is the cleanest split between provider latency and daemon-side delay. The problem is only the name: folding "the provider started responding" and "the member saw something" into one field makes it unreliable in precisely the escalation where someone would quote it. Either:

- rename to something the mark is actually true of (`first_provider_event` or `first_output_received`) and leave it where it is; or
- keep the name and mark after the first successful `ReportTaskMessages`, with a test covering the report-failure path.

I would take the rename — cheaper, and it keeps the more useful number.

## Non-blocking

`phase_elapsed_ms` is "time since the previous *recorded* phase", not "duration of this phase". Worth stating in the doc comment on `Mark`, since a skipped phase — e.g. an early return between `skills_ready` and `environment_ready` — silently widens the next one rather than leaving a gap.

## Verification

Checked out `f590a9e` and ran, all green: `go test ./internal/daemon -count=1`, `go vet ./internal/daemon`, `git diff --check`.

Blast radius here is small — diagnostic-only, no task state touched, and a revert is just reverting this PR. But a diagnostic that misattributes where time went is worse than no diagnostic at all, so I would like both semantics corrected first. Ping me once they are and I will re-review promptly.

server/internal/daemon/daemon.go
eveningb123: Fixed in a260bb004 by renaming the phase to first_output_received and the classifier to isTaskOutputReceived. The mark intentionally stays at daemon receipt time, so the field now describes the provider/daemon boundary without claiming that ReportTaskMessages succeeded or that the member saw the output. The classification and lifecycle-order tests use the new name.

server/internal/daemon/daemon.go
eveningb123: Fixed in a260bb004. turn_completed is now emitted by runTask only after any fresh-session retry has finished and reconcileFreshRetryResult has selected the final task result; executeAndDrain no longer emits that task-level phase. TestExecuteAndDrain_ResumeFailureFallback now asserts that the discarded resume attempt cannot mark completion and that completion is recorded only after the successful fresh retry. I also clarified in Mark that phase_elapsed_ms is measured since the previous recorded phase, so skipped phases widen the next interval.

eveningb123:
Implemented the requested semantic corrections in a260bb004:

- moved task-level turn_completed after fresh-session retry reconciliation;
- added a resume-rejected to fresh-retry-succeeds regression;
- renamed first_visible_output to first_output_received while keeping the provider/daemon receipt boundary;
- documented that phase_elapsed_ms is measured since the previous recorded phase.

Verification: go test ./internal/daemon -count=1, go vet ./internal/daemon, and git diff --check all pass locally.

@Bohan-J ready for re-review.

Bohan-J (approved):
Both blockers from the first review are resolved, and the fixes are the right ones rather than the minimal ones.

- `turn_completed` is now marked once in `runTask` after `reconcileFreshRetryResult` (`daemon.go:8420`) and no longer inside `executeAndDrain` at all, so a resume-rejected task attributes completion to the attempt that actually finished. The regression test in `daemon_test.go` asserts both halves — nothing recorded for the discarded attempt, and nothing recorded by `executeAndDrain` before reconciliation. That's a stronger assertion than I asked for.
- `first_visible_output` is now `first_output_received`, with the predicate renamed to match. The phase claims exactly what it measures.
- The non-blocking note is handled too: `Mark`'s doc comment states that `phase_elapsed_ms` is the interval since the previous *recorded* phase, and that a skipped phase widens the next one.

Verified on `d50a68553` merged with `main`: `go test ./internal/daemon -count=1` and `go vet ./internal/daemon` pass locally, and CI is green on the merge state.

One thing I checked separately, because it's the first question an operator will ask: this doesn't scale with task activity. The dedup map caps output at 9 records per task no matter how many tools a run calls — about 2.4 KB per task, roughly 1% of `daemon.log` volume on a real machine, comfortably inside the existing 20 MB rotation.

Two optional follow-ups, neither worth holding this PR for:

- `taskLog` already carries `task=<full id>`, so the recorder's `task_id` attribute repeats the same UUID on every phase line. Dropping it would leave one field name for one thing.
- `Mark` takes its mutex before the dedup check, so the two call sites in the drain loop lock twice per message for the life of the run. The cost is negligible next to per-message parsing and batching, but an atomic fast path would remove it if it ever shows up in a profile.

Thanks for this — and for how you handled the review. The thinking path in the description, the AI disclosure, and rebasing onto the full-task-ID change from `main` rather than reverting it all made this easy to land. Nice first contribution.

Issue #3980, opened 2026-06-10
[Feature]: Ingest Claude Fable 5 model


fayewong-crypto:
### Deployment type

Official App

### What do you want and why?

Current Issue:
After updating the Claude CLI, we can successfully use the Fable model directly in the local Claude Code CLI. However, Multica’s Claude runtime (provider: claude) does not reflect or expose this new model. As a result, the model is not available for selection when using agents in Multica.

Impact:
Multica’s runtime model list lags behind the actual models supported by the underlying Claude Code CLI. Users cannot leverage newly added models from Claude CLI updates within the Multica workspace.

### Proposed solution (optional)

Multica should dynamically query the locally installed Claude Code CLI for the list of supported models (or provide a configuration option to use any model the CLI supports), instead of relying on a hardcoded/static model list maintained by Multica. This would allow new models like Fable to become available immediately after a CLI update.

Bohan-J:
Thanks for reporting this.

For Claude Code runtimes, Multica follows the local Claude Code CLI configuration by default when no explicit model is selected on the agent. So if you want to use Fable 5 immediately, make sure your local Claude Code CLI is configured/selected to use Fable 5; Multica will let Claude Code resolve that default.

We also added explicit Fable 5 support in #3982, which is shipping today. That makes `claude-fable-5` available in Multica's Claude model picker and adds the corresponding usage pricing.

On dynamic discovery: Multica's model discovery framework can use runtime-specific discovery when the underlying CLI/protocol exposes a structured model catalog. Claude Code currently does not provide a stable machine-readable `model list` / `models --json` command; it only accepts `--model <model>` with aliases or full model IDs. Because of that, the Claude Code runtime currently keeps a short built-in catalog, with manual/custom model input as a fallback. If Claude Code adds a structured model-list command later, we can switch Claude discovery to use that and keep the static list only as a fallback.

Closing this as completed by #3982 for Fable 5 support.

PR #7892, opened 2026-09-01
MUL-6943 feat(agent): add Claude Fable 5.1 model and pricing

yarrumretep:
## What does this PR do?

Adds Claude Fable 5.1 (`claude-fable-5-1`, released 2026-09-01) to the Claude Code runtime's static model catalog and to the backend and frontend pricing tables. Same shape as #3982, which added Fable 5.

Today the model is reachable only by typing the ID into the picker's manual-entry field, and that path has three gaps:

- The daemon's pre-execution guard validates `thinking_level` against the catalog. A model that is not in it has no effort entry, so any persisted effort is dropped with a warning and Claude Code runs at its own default.
- The frontend pricing resolver cannot map the `-1` suffix, so usage rows for Fable 5.1 show as unmapped in the dashboard.
- The backend alias rule for Fable 5 is unanchored, so `claude-fable-5-1` silently resolved to the Fable 5 row. That was harmless for input and output, but Fable 5.1 prices cache reads at $0.25 / MTok (0.025x input) versus Fable 5's $1.00, so cached reads were billed at 4x in metrics.

Pricing source: https://platform.claude.com/docs/en/about-claude/pricing (Fable 5.1: $10 input, $50 output, $12.50 5m cache write, $0.25 cache read).

## Related Issue

No issue filed for Fable 5.1 yet. Fable 5 was tracked in #3980.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/models.go`: add `claude-fable-5-1` to `claudeStaticModels()`. Sonnet 4.6 stays the sole default. No change to `claudeModelEffortAllow`: a model absent from the allow-list gets the full `low..max` superset from `claude --help`, which matches Fable 5.1 (same as Fable 5 today).
- `server/internal/metrics/pricing.go`: add the `anthropic:claude-fable-5-1` row; add a `claude-fable-5[-.]1` alias rule ahead of Fable 5; anchor the Fable 5 rule to `claude-fable-5($|[^-.0-9])` so it keeps matching the bare ID and `[1m]` form but no longer swallows `-1`.
- `packages/views/runtimes/utils.ts`: add the `claude-fable-5-1` pricing row.
- Tests:
  - `server/pkg/agent/models_test.go`: `TestClaudeStaticModelsExposesFable51`.
  - `server/pkg/agent/thinking_test.go`: `TestValidateThinkingLevel_ClaudeFable51AcceptsFullEffortRange` (guards the effort-dropping symptom above).
  - `server/internal/metrics/pricing_test.go`: `TestPriceForModelAliasAnthropicFable51` (bare, `/`- and `:`-prefixed, dotted, and `[1m]` spellings resolve to the 5.1 row; Fable 5 and `claude-fable-5[1m]` still resolve to the Fable 5 row).
  - `packages/views/runtimes/utils.test.ts`: `estimateCost` and `isModelPriced` cases for the dashed, dotted, and provider-prefixed forms.

Not touched here, noted for a follow-up: the Sonnet 5 comments in both pricing tables still describe the $2 / $10 rate as introductory pricing ending 2026-08-31. Anthropic's pricing page now says that rate is permanent and the increase will not occur, so the rows are correct and only the comments are stale.

## How to Test

1. `cd server && go test ./pkg/agent ./internal/metrics`
2. `pnpm -C packages/views exec vitest run runtimes/utils.test.ts dashboard/utils.test.ts`
3. With a daemon built from this branch, open an agent's model picker on a Claude Code runtime: "Claude Fable 5.1" appears with the effort picker enabled, and a task launched with an effort set logs `--effort` on the `claude` command line instead of the `thinking_level: not valid for this (provider, model); skipping injection` warning.

## Checklist

Thinking path: Fable 5.1 shipped 2026-09-01 and no release (v0.4.37 is the latest) or `main` commit mentions it. The Claude catalog is static by design (`ListModels` comment: "Claude's catalog is static by design, not by failure"), so the only path is a catalog entry, and #3982 established the file set for that. Reading `ValidateThinkingLevelWith` and the daemon's `thinking_level` guard showed the user-visible cost of a missing entry is silently dropped effort, hence the extra `thinking_test.go` case. Reading `modelAliasRules` showed the unanchored Fable 5 regex would misprice 5.1's cheaper cache reads, hence the anchoring and the regression cases for both models.

### DIFF
--- a/packages/views/runtimes/utils.test.ts
+++ b/packages/views/runtimes/utils.test.ts
@@ -146,6 +146,20 @@ describe("estimateCost", () => {
     expect(cost).toBeCloseTo(10 + 50 + 1 + 12.5, 5);
   });
 
+  it("prices Claude Fable 5.1 at the Mythos-class tier with quarter-rate cache reads", () => {
+    // Same $10 / $50 and $12.50 cache write as Fable 5, but cache reads are
+    // 0.025x input ($0.25) rather than the usual 0.1x.
+    const cost = estimateCost({
+      ...zeroUsage,
+      model: "claude-fable-5-1",
+      input_tokens: 1_000_000,
+      output_tokens: 1_000_000,
+      cache_read_tokens: 1_000_000,
+      cache_write_tokens: 1_000_000,
+    });
+    expect(cost).toBeCloseTo(10 + 50 + 0.25 + 12.5, 5);
+  });
+
   it("prices Claude Sonnet 5 at Anthropic's intro $2 / $10 tier", () => {
     const cost = estimateCost({
       ...zeroUsage,
@@ -886,6 +900,7 @@ describe("isModelPriced", () => {
   it("recognises both Claude and Codex/GPT families", () => {
     expect(isModelPriced("claude-sonnet-5")).toBe(true);
     expect(isModelPriced("claude-fable-5")).toBe(true);
+    expect(isModelPriced("claude-fable-5-1")).toBe(true);
     expect(isModelPriced("claude-sonnet-4-6")).toBe(true);
     expect(isModelPriced("gpt-5-codex")).toBe(true);
     expect(isModelPriced("gpt-5-mini")).toBe(true);
@@ -899,6 +914,7 @@ describe("isModelPriced", () => {
     // hit the same catalog row, otherwise Copilot-routed usage gets bucketed
     // as "unmapped" and the user has to type the price in by hand.
     expect(isModelPriced("claude-sonnet-5")).toBe(true);
+    expect(isModelPriced("claude-fable-5.1")).toBe(true);
     expect(isModelPriced("claude-haiku-4.5")).toBe(true);
     expect(isModelPriced("claude-sonnet-4.5")).toBe(true);
     expect(isModelPriced("claude-sonnet-4.6")).toBe(true);
@@ -912,6 +928,7 @@ describe("isModelPriced", () => {
     // The provider prefix is routing metadata, not part of the SKU.
     expect(isModelPriced("anthropic/claude-sonnet-5")).toBe(true);
     expect(isModelPriced("anthropic/claude-fable-5")).toBe(true);
+    expect(isModelPriced("anthropic/claude-fable-5-1")).toBe(true);
     expect(isModelPriced("anthropic/claude-opus-4.7")).toBe(true);
     expect(isModelPriced("anthropic/claude-sonnet-4-6")).toBe(true);
   });
--- a/packages/views/runtimes/utils.ts
+++ b/packages/views/runtimes/utils.ts
@@ -195,9 +195,11 @@ const MODEL_PRICING: Record<
   // -- Anthropic: current generation. Sonnet 5 uses Anthropic's published
   //    intro launch rate ($2 / $10 through 2026-08-31). This static map has
   //    no future-dated pricing support yet, so update the row when the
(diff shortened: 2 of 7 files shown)


Bohan-J:
Thanks for the contribution — this was the right file set, and the model ID, pricing, and effort range all check out against Anthropic's docs. Catching that the unanchored Fable 5 alias would misprice 5.1's cache reads at 4x is the part of this PR that only shows up if you actually read `modelAliasRules`, and it would have been easy to miss.

Since maintainer edits are enabled and we want Fable 5.1 available today, we pushed three follow-up commits instead of waiting on a round trip. All three are backend alias matching only — your catalog entry, pricing rows, and frontend table are untouched.

1. **`stop both Fable alias rules at their own version`** — the new `claude-fable-5[-.]1` rule was itself unanchored, so a later sibling (`claude-fable-5-10`, `claude-fable-5-1-preview`) would silently borrow the 5.1 row: the same failure the anchoring was added to prevent. Separately, `[^-.0-9]` on the Fable 5 rule was stricter than intended — it also excluded `-latest` and dated snapshots, which `canonicalCandidates` in `packages/views/runtimes/utils.ts` still strips and prices, so the dashboard and `RecordLLMUsage` disagreed on the same id. Both rules now share a `claudeVersionEnd` terminator whose admitted suffixes mirror `stripContextTag` / `stripDate`.

2. **`anchor claudeVersionEnd to the end of the model id`** — the terminator's alternatives matched only the *start* of each suffix, and the rules are substring matches, so `claude-fable-5-1-latest-preview`, `claude-fable-5-1-20260901x` and `claude-fable-5-1[1m]junk` still resolved.

3. **`strip exactly one context tag before the alias retry`** — `PriceForModelAlias` peeled the trailing tag off `claude-fable-5[1m][2m]` and retried, letting the leftover `[1m]` satisfy a rule that had just correctly rejected the raw form. The frontend strips one tag and does not re-strip, so those ids were unmapped there.

Regression coverage was added for the dated, `-latest`, context-tagged and doubly-tagged spellings on both Fable rows, plus negative assertions that a future `claude-fable-5-2` / `-10` stays unmapped until it gets a row of its own.

Verified before merging: the official model ID and the $10 / $50 / $12.50 / $0.25 rates against Anthropic's pricing page (including the note that 5.1 cache hits are 0.025x input rather than the standard 0.1x); the full `low..max` effort range against the effort docs and against the real Claude Code binary through `ValidateThinkingLevel`, which confirms the dropped-effort symptom your description describes; and `go test ./internal/metrics ./pkg/agent` plus the Views suites.

One thing for a separate follow-up, not a blocker here: Claude Code rejects `claude-fable-5-1` below 2.1.251 (`400 ... version 2.1.251 or newer is required`), while the global floor for the Claude runtime is still 2.0.0. Members on an older daemon will see the model in the picker and hit that error at run time. Raising the global floor would penalize everyone still on older models, so the real fix is a per-model minimum-runtime-version signal that hides or disables the option — we'll track that separately.

Thanks again for getting this in ahead of us.

Bohan-J (approved):
Approving. Model ID, pricing, and the full low..max effort range verified against Anthropic's docs and against the real Claude Code binary through ValidateThinkingLevel. Backend alias matching was tightened in three follow-up commits so Fable 5 and 5.1 cannot inherit each other's or an unknown variant's tier, and so the backend and the dashboard resolve the same suffix forms. All required checks green.

PR #7896, opened 2026-09-01
MUL-6946: fix(cli): stop issue list misreporting its page

### LINKED ISSUE #7895: [Bug]: `issue list` reports the `--limit` you asked for, not the page you got
### Deployment type

self-host

### Area

CLI

### What happened?

`multica issue list --output json` reports the `--limit` you asked for, not the page size the server actually applied. The server caps a page at 100 (`ListIssues` clamps it), but the CLI builds its envelope from the flag value:

```go
limit, _ := cmd.Flags().GetInt("limit")
hasMore := offset+len(issuesRaw) < int(total)
wrapped := map[string]any{
    "issues": issuesRaw, "total": int(total),
    "limit": limit, "offset": offset, "has_more": hasMore,
}
```

So `--limit 500` returns 100 rows next to `"limit": 500` and `"has_more": true`, with nothing on stderr. The paging loop most scripts write, `offset += limit`, then jumps from 0 to 500 and skips 400 rows. On a 145-issue project it collects 100 issues, sees `has_more: false`, and exits 0. Nothing tells you 45 rows are missing.

`--limit 0` and negative values have the same shape. The flag is dropped from the query, the server applies its own default of 100, and the envelope still echoes `0` or `-5`. A negative `--offset` is dropped and echoed back the same way.

Table output, which is the default, has the opposite gap. `total` and `has_more` are only computed inside the JSON branch, so someone running `multica issue list` on a large workspace sees 50 rows and has no way to know there are 2,706.

I saw that #7674 updated the `--limit` help text to say the server caps a page at 100, and I agree with the review there that an unbounded `--all` is the wrong tool. This report is about the narrower problem that is still there: the command reports a page it did not return. I would expect it to either send what the server will honour and report exactly that, or refuse a value it cannot honour, the same way `--direction` on a position sort is refused instead of dropped.

I have a PR ready.

### Steps to reproduce

Any workspace with more than 100 issues. Take a project id from `multica project list --output json`.

```bash
multica issue list --project <PROJECT_UUID> --limit 500 --output json \
  | jq -c '{returned: (.issues|length), total, limit, has_more}'
# {"returned":100,"total":145,"limit":500,"has_more":true}
```

The sweep loop:

```bash
off=0; seen=0
while :; do
  r=$(multica issue list --project <PROJECT_UUID> --limit 500 --offset $off --output json)
  n=$(echo "$r" | jq '.issues|length'); lim=$(echo "$r" | jq '.limit'); hm=$(echo "$r" | jq '.has_more')
  echo "offset=$off limit=$lim returned=$n has_more=$hm"
  seen=$((seen+n)); [ "$hm" = "false" ] && break; off=$((off+lim))
done
echo "collected $seen"
```

Output on a 145-issue project:

```
offset=0   limit=500 returned=100 has_more=true
offset=500 limit=500 returned=0   has_more=false
collected 100
```

Out-of-range flags:

```bash
multica issue list --limit 0 --output json | jq -c '{returned:(.issues|length), limit, has_more}'
# {"returned":100,"limit":0,"has_more":true}
multica issue list --limit -5 --offset -5 --output json | jq -c '{returned:(.issues|length), limit, offset}'
# {"returned":100,"limit":-5,"offset":-5}
```

Table mode:

```bash
multica issue list --limit 3    # 3 rows, nothing on stdout or stderr about the other 2,703
```

### Additional context

```
multica 0.4.35 (commit 09a2410e8); the same code is on main at 2e2974510
server/cmd/multica/cmd_issue.go, runIssueList: envelope built from the flag; total and has_more only computed in the JSON branch
server/internal/handler/issue.go, ListIssues: limit clamped to 100; response body is {issues, total}
Related: #3563 (added the server clamp), #7674 (help text only)
```

Firer:
## What does this PR do?

Makes `multica issue list` report the page it actually returned.

Today the JSON envelope echoes whatever `--limit` you passed, but the server clamps a page to 100. So `--limit 500` returns 100 rows next to `"limit": 500`, and a script that advances `offset += limit` (the natural loop) skips 400 rows and finishes with `has_more: false` and exit 0. On a 145-issue project that loop collects 100 issues and reports a clean sweep. `--limit 0`, negative limits and negative offsets are dropped from the query and echoed back the same way. Table mode, the default, has the opposite gap: it never shows `total` or `has_more` at all, so 50 rows of 2,706 look like the whole list.

I read the review on #7674 before writing this, and I agree an unbounded `--all` is the wrong tool. This PR stays inside the existing one-call-one-page model and just makes that call truthful:

- `--limit` outside 1 to 100 and a negative `--offset` are rejected before any request is made, with the cap named in the error. This follows the `--direction` guard a few lines below it: a flag the server would silently ignore is a footgun in scripts, and the review on #7674 asked for an error over silent truncation. It is a behaviour change twice over: anyone passing `--limit` above 100 today was already getting 100 rows and a wrong envelope, so I think an error is the kinder outcome there; and `--limit 0`, which used to mean "send no limit and let the server default to 100", is now rejected by the same guard. If you would rather clamp to 100 and report that, it is a two-line change and I am happy to switch.
- `limit` is always sent, so the server never applies its own default.
- In the JSON envelope, `has_more` does not rest on `total` alone. An empty page ends a walk whatever `total` says (the page and count queries are separate statements, so they can briefly disagree, and an empty page with `has_more: true` would loop a script forever). A full page carries a walk on unless `total` is trusted, and `total` is trusted only when it is larger than the page it came with. That one comparison rejects both ways the count goes missing: `ListIssues` answers with the row count it just returned when its own count query fails (`server/internal/handler/issue.go:1563`), which is never larger than the page, and an installed CLI can be talking to a backend that no longer sends the field. The cost is a first page that fills exactly. It cannot tell a genuine total from a failed count, so it reports `has_more: true` and the next request comes back empty. One extra request on a list of exactly one page's worth seemed clearly better than a walk that ends one page in whenever the count is unhealthy; the doc paragraph states the trade-off.
- `limit` and `offset` echo the request on every page, so `limit` holds steady across a walk and `offset` is the one you passed. The rows in this page are `issues | length`, and that is what to advance `--offset` by; it equals `limit` on a full page and stays right if a page ever comes back shorter. This is what `fetchIssueColumn` and `fetchAutopilotCandidates` already do internally.
- Table mode prints one line on stderr when the result is a page of a larger set: `Showing 1-50 of 2706 issues. Next page: --offset 50`. It goes to stderr so stdout stays a plain table, the same surface `issue comment list` uses for its next-page cursor. It prints whenever `has_more` would be true or `--offset` is above zero, and stays silent otherwise. The `of N` is dropped whenever `total` is not trusted (`Showing issues 1-100. Next page: --offset 100`), so the footer never claims a page is the whole list beside a pointer to the next one.
- The 100 lives in one constant, `issueListMaxPageSize`, shared with `fetchIssueColumn`, which already paged at 100.

Deliberately not here: any bulk or export mode, keyset pagination, or a server-side echo of the applied limit. The last one would be the durable fix (every client could stop hard-coding the cap), but it is a separate API change.

## Related Issue

Closes #7895

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/cmd/multica/cmd_issue.go`: `issueListMaxPageSize`; `--limit` / `--offset` validated at the top of `runIssueList` before the assignee and project resolvers run; `limit` always sent; `has_more` derived from the returned page and a trusted `total`; `limit` and `offset` echoed as given; table footer on stderr; `fetchIssueColumn` uses the constant; help text.
- `server/cmd/multica/cmd_issue_test.go`: five tests. Out-of-range flags are rejected with zero requests (including with `--assignee` set, to pin the ordering); `limit` and `offset` reach the query; the JSON envelope for a full page, a partial last page, a server that applied a smaller page, empty pages (past the end, and while `total` still claims more), a full page ending exactly on a later page boundary (no probe), a full page whose `total` is missing, undercounts the page, or equals it (on the first page and on later ones), and a short first page whose `total` equals it (the walk ends), plus nothing on stderr in JSON mode; that `--fields` cannot move the paging numbers; the table footer for first, middle and last pages, an empty page at and past the end, a list that fits, no results, and every untrusted-total shape (missing, equal to the page, a failed count on an empty page), plus stdout never carrying footer text.
- `CLI_AND_DAEMON.md`: two short paging paragraphs in the `issue list` section (the cap, the footer and when it drops `of N`, the envelope fields, advance `--offset` by `issues | length`, when `has_more` ignores `total` and what that costs, and use `--sort created_at` for a walk because `position` moves under drags).

## How to Test

1. `cd server && go test ./cmd/multica -run 'TestRunIssueList' -count=1`
2. Against a workspace with more than 100 issues:
   ```bash
   multica issue list --limit 500 --output json      # now errors and names the cap
   multica issue list --limit 3                       # 3 rows, then on stderr: Showing 1-3 of N issues. Next page: --offset 3
   multica issue list --limit 100 --output json | jq -c '{returned:(.issues|length), total, limit, has_more}'
   ```
3. The sweep from the issue, advancing `--offset` by `issues | length`, now collects every row and stops. Against a 3,332-issue workspace it took 34 requests, collected 3,332 rows, kept `limit` at 100 throughout, and ended on `has_more: false`.

Before this change the same steps return 100 rows with `"limit": 500`, print no footer, and the sweep stops at 100.

## Checklist

- [x] I have run tests locally and they pass (`go test ./cmd/multica`, `go vet`, plus the manual steps above against a 3,332-issue workspace; no frontend changes, so no Playwright run)
- [ ] If this change affects the UI, I have included before/after screenshots (CLI only)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs (N/A)
- [ ] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` (no Chinese copy touched)

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** I gave Claude Code the symptom from our own sweep scripts (a project that came back 100 issues short with exit 0) and had it verify every claim against current `main` and a live workspace before touching anything, which is how it found the `--limit 0` and negative-offset variants and the review on #7674. The tests were written first and fail against `main`. The plan and the final diff each went through a three-way review with Claude, GPT and Gemini reviewers; their findings shaped the change (validation hoisted above the resolvers, `has_more` false on an empty page, the footer staying quiet when `total` is missing). I reviewed and take responsibility for the final diff.

Rebased onto current `main` on 7 Sep, after #7946 and #8008 landed in the same function; the conflict was in the JSON branch, where the envelope now sits alongside `--fields` and `--resolve-properties`. The rebase went through the same three-way review, which is what turned up the `has_more` gap fixed above.

The second round, after the review on 7 Sep, went the same way. Claude Code reproduced the blocking case as a failing test before touching the rule, and the result went back through the three-way review. That round caught the empty-page footer printing a failed count as `(0 total)`, and reduced the trust rule to the single comparison shipped here after showing the second clause could only fire inside the concurrent-write race the doc already disclaims. Rebased again onto `main` on 8 Sep.


Firer:
Hey @Bohan-J do you have any feedback for this PR? I'm running to problems using the CLI on a large Multica worskpace due to this bug with pagination.

Bohan-J (changes requested):
Thanks for this — the diagnosis is right and so is the shape of the fix: stay inside one-call-one-page, validate up front, and make the envelope describe the page that actually came back. Hoisting the validation above the assignee and project resolvers so an out-of-range flag costs zero requests, forcing `has_more: false` on an empty page, and putting the table footer on stderr next to the cursor `issue comment list` already prints are all the right calls.

One blocking problem, and one contract question.

## Blocking: the `total` guard still misses the first page

The switch you added handles a `total` that undercounts the walk, but not the specific value the server produces when its count query fails. `ListIssues` degrades like this:

```go
// server/internal/handler/issue.go:1563
if err := h.DB.QueryRow(ctx, countQuery, countArgs...).Scan(&total); err != nil {
    total = int64(len(issues))
}
```

At `offset > 0` that lands in your `default` branch and the walk carries on — your "full page while total undercounts the walk" case covers it. At `offset 0` it does not. `total` comes back equal to `len(page)`, which is exactly `offset+returned`, so `case totalOK && int(total) >= offset+returned` wins and reports `has_more: false` beside a full page.

Concretely: a 145-issue project whose count query fails returns 100 rows and `total: 100`, and the sweep loop from the issue collects 100 of 145 and exits 0. That is the same failure this PR exists to fix, just narrowed to the window where the count query is unhealthy.

It slots straight into your existing table test:

```go
{name: "full page while total equals the page at offset 0", limit: "100", offset: "0", rows: 100, total: 100, wantTotal: 100, wantLimit: 100, wantHasMore: true},
```

On `539d3fd9e` that case fails with `has_more: false`; it is the only failure — the rest of `go test ./cmd/multica` is green.

This edit passes it and every case you already have:

```go
case totalOK && int(total) >= offset+returned:
    // A failed count query answers total = len(page). At offset 0 that is
    // indistinguishable from a genuine single-page total, so a full page
    // still probes rather than ending the walk on it.
    hasMore = offset+returned < int(total) || (offset == 0 && returned == limit)
```

The cost is one extra empty request when a result set genuinely ends on an exact page boundary at offset 0. That looks clearly worth it against silently dropping rows, but it is a real trade-off, so please state it in the doc paragraph whichever way you resolve it.

The root cause is server-side and out of scope here: substituting `len(issues)` for a failed count is itself a silent lie, and it forces every client to guess around it. A follow-up that reports "total unavailable" explicitly would let clients stop inferring — worth its own issue.

## Contract question: `limit` changes meaning with `has_more`

`appliedLimit` is the returned row count when `has_more` is true and the requested limit when it is false, so `--limit 100` on a final page of 45 rows reports `"limit": 100`. That sits awkwardly next to the doc line saying `limit` is the page size the server applied.

There are three distinct numbers here — requested cap, effective cap, rows in this page — and the rows are already `issues | length`. Either contract works as long as it holds on every page: `limit` as the effective page cap (constant across a walk), or `limit` as the rows in this page. I lean toward the first, plus telling scripts to advance by `issues | length` rather than by `limit`. That is what `fetchAutopilotCandidates` already does internally (`offset += pageLen`), and it stays correct regardless of which meaning `limit` carries. Whichever you choose, the doc paragraph should match it exactly — this PR is about the envelope not lying, so it is worth spending a sentence to get right.

Not blocking on its own; I'd fold it in with the fix above.

Happy to re-review as soon as it is updated.

Firer:
Hey @Bohan-J, thanks for such a careful read. You were right on both counts, and the first one was a proper hole.

**The blocking one.** Your case reproduced exactly on 539d3fd9e: `has_more: false` next to a full page, and it was the only failure. I went a slightly different way from the patch you sketched, though it lands in the same place. Rather than special-casing offset 0, the rule is now that `total` is trusted only when it is larger than the page it came with:

```go
totalTrusted := totalOK && int(total) > returned
```

A failed count is never larger than its page, at any offset, so that one comparison catches it on the first page as well as later ones. Your case is in the table, plus two guards in the opposite direction: a short first page whose `total` equals it still ends the walk, and a full page ending exactly on a later boundary still stops without a probe. The doc states the cost the way you asked: a list that fills exactly one page makes one extra request that comes back empty.

The earlier version also required `total >= offset+returned`. A second pass showed that clause was untested and could only fire inside the concurrent-write race the doc already disclaims, so it's gone.

**`limit`.** I took your lean: `limit` and `offset` now echo the request on every page, and the doc tells scripts to advance by `issues | length`, which is what the internal walkers already do. The former `appliedLimit` is deleted.

**Two things that fell out of the same change.** The table footer now drops `of N` whenever `total` isn't trusted (`Showing issues 1-100. Next page: --offset 100`), so it never says "1-100 of 100" beside a pointer to the next page. And the empty-page line had the same blind spot in miniature: with a failed count it would have printed `No issues at --offset 100 (0 total).` on a 100-issue workspace, so that parenthetical is now gated the same way. The footer also stopped going silent when `total` is missing, since a full page with no total now carries `has_more: true`.

I ran the built CLI against a 3,332-issue workspace: `--limit 500` errors and names the cap, the footer and envelope report the real page, and a full walk advancing by `issues | length` collects all 3,332 in 34 requests with `limit` constant at 100 and `has_more` ending it.

Rebased onto current main today, still a single commit, no conflicts. On the server-side follow-up, I agree that reporting "total unavailable" explicitly is the real fix and would let clients stop inferring. Happy to open an issue for that if you'd like. The same pass also noticed that `fetchIssueColumn` (behind `issue reorder`) still ends its walk on `offset >= total`, so it has the same trap in a column of over 100 issues when the count fails. I've left it alone here since it's a separate code path, but it probably wants the same rule in a follow-up.

Bohan-J (approved):
Verified at `8ad180009`. Approving — thanks for sticking with this through two rounds.

The `totalTrusted := totalOK && int(total) > returned` rule is a better fix than the offset-0 special case I sketched: a failed count can never exceed the page it arrived with, so one comparison covers every offset instead of treating "first page" as the defect's signature. Dropping the untested `total >= offset+returned` clause was the right call too.

I re-ran the verification rather than reading the diff alone: 14 adversarial cases against the new rule — degraded counts, missing/negative/non-numeric totals, exact page boundaries, short pages, empty pages — all behaved as specified. Full `go test ./cmd/multica` green, `go vet` clean, and every CI check passing.

The extras you found on your own are what make this land well: the footer dropping `of N` when the count isn't trusted, so it can never print "1-100 of 100" beside a pointer to the next page, and the same gate on the empty-page line. Those weren't in the review.

Both follow-ups you identified are already in flight — the server reporting count failures explicitly, and `fetchIssueColumn` carrying the same trap into `issue reorder`. Thanks for flagging the reorder one; it writes a wrong position rather than just showing a short list, so it mattered more than the symptom suggested.

PR #7902, opened 2026-09-02
fix(autopilot): route accountable member issues

sandeepl337:
## Problem

A `run_only` autopilot can create and route a new issue, but cannot assign a pre-existing canonical issue created by the human who owns the run. Promoter-style automations therefore create duplicate execution-wrapper issues instead of routing the canonical record.

## Fix

Allow the existing-issue assignment scope only when all existing autopilot authority checks pass and either:

- the issue is already bound to this exact task, or
- the issue was created by the run's accountable member and the task independently proves active `run_only` lineage in the same workspace.

The authority remains unavailable for another member's issue, a sibling run's issue, completed tasks, untrusted attribution sources, and unrelated issues in the same batch.

## Verification

Real PostgreSQL 17:

- focused assignment positive/negative suite: PASS
- full `server/internal/handler` package: PASS (`36.716s`)
- mixed batch regression proves an allowed member issue cannot lend authority to another member's issue.

### DIFF
--- a/server/internal/handler/agent_access.go
+++ b/server/internal/handler/agent_access.go
@@ -372,8 +372,9 @@ func (h *Handler) autopilotAssignAuthorityFromRequest(r *http.Request, scope ass
 //   - the assignment is bound to work this run verifiably owns —
 //     issueBoundToAutopilotTask for an existing issue (which for a run_only-created
 //     issue re-proves the run_only lineage, so a create_issue task cannot escape
-//     its same-issue bound via a fresh top-level issue), or the task's own
-//     verified autopilot run when creating a parentless issue;
+//     its same-issue bound via a fresh top-level issue), an existing member-created
+//     issue whose creator is exactly the run's accountable human, or the task's
+//     own verified autopilot run when creating a parentless issue;
 //   - the issue (when there is one) is in the request's workspace;
 //   - the accountable human is STILL a member of that workspace.
 //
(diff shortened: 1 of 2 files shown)


Bohan-J:
Superseded by #7918, which is now merged.

Your diagnosis was right, and the flow you described — a `run_only` autopilot forced to create a duplicate wrapper issue because it cannot route the canonical one — is exactly the problem worth fixing. We ended up taking it at the root instead.

The reason a scheduled run could not assign that issue is that it carried no `originator_user_id` at all. Schedule/webhook dispatch was deliberately attributed with an *accountable* human but no *authorizing* one, so `canInvokeAgent` failed closed against private agents. Every capability such a run legitimately needed therefore had to arrive as a separately-bounded borrow path: first child creation, then issues the run itself created, and this PR would have been the third — with sub-issue dispatch already queued as the fourth. Each is a hand-written scope no operator can see in the UI.

#7918 gives the run a real principal instead: a schedule/webhook dispatch now acts as the human who created the firing trigger, and is judged by the ordinary invoke gate — exactly what that person gets acting directly. Routing a pre-existing member-created issue then needs no special case, and the three existing borrow paths were deleted along with the scope machinery that bounded them.

On the bounds you were careful to enforce: they are still enforced, just expressed once rather than per surface. A run whose human holds no invoke rights on the target is refused, and a run with no resolvable human fails closed. The per-issue binding your version checked is intentionally gone — with a real principal, "which issue is this" stops being the question and "who is this acting as" is the whole answer.

Closing as superseded. Thanks for the report and for the negative cases you spelled out; they made the shape of the problem much clearer.

PR #7905, opened 2026-09-02
MUL-6953: fix(execenv): support OpenClaw agents.entries config

wysl:
## Summary

  - support OpenClaw installations using `agents.entries`
  - retain compatibility with legacy `agents.list`
  - fall back to the registry only when neither config path is available
  - recognize the current `Unknown config path: agents.list` error wording

  ## Verification

  - targeted execenv tests passed
  - git diff --check passed
  - full execenv suite has one unrelated existing failure:
    `TestMigrateHermesTaskMemoriesFailureKeepsSource`

### DIFF
--- a/server/internal/daemon/execenv/openclaw_config.go
+++ b/server/internal/daemon/execenv/openclaw_config.go
@@ -581,6 +581,8 @@ func buildPerTaskOpenclawConfig(activePath string, exists bool, resetPath string
 		if rewritten := rewriteAgentsListWorkspaces(resolvedList, workDir); rewritten != nil {
 			agents["list"] = rewritten
 		}
+	} else if rewritten := rewriteAgentsEntriesWorkspaces(resolvedList, workDir); rewritten != nil {
+		agents["entries"] = rewritten
 	}
 	cfg := map[string]any{
 		"agents": agents,
@@ -650,6 +652,32 @@ func buildGatewayOverride(p OpenclawGatewayPin) map[string]any {
 		return nil
 	}
 	return out
+}
+
+func rewriteAgentsEntriesWorkspaces(list []any, workDir string) map[string]any {
+	entries := make(map[string]any)
+	for _, item := range list {
+		entry, ok := item.(map[string]any)
+		if !ok || entry["__multica_entries_source"] != true {
+			return nil
+		}
+		id, ok := entry["__multica_entries_key"].(string)
+		if !ok || id == "" {
+			return nil
+		}
+		copyEntry := make(map[string]any, len(entry))
+		for key, value := range entry {
+			if key != "__multica_entries_key" && key != "__multica_entries_source" {
+				copyEntry[key] = value
+			}
+		}
+		copyEntry["workspace"] = workDir
+		entries[id] = copyEntry
+	}
+	if len(entries) == 0 {
+		return nil
+	}
+	return entries
 }
 
 // rewriteAgentsListWorkspaces copies every entry of the resolved agents.list
@@ -1132,8 +1160,9 @@ func openclawResolvedAgentsList(bin string, timeout time.Duration) ([]any, bool,
 	out, err := openclawExec(ctx, bin, "config", "get", "agents.list", "--json")
 	if err != nil {
 		if isOpenclawKeyMissingResult(out, err, "agents.list") {
-			// New schema: the config path is gone; the agents live in the
-			// sqlite registry. Resolve them via the subcommand instead.
+			if openclawUnknownConfigPath(out, err, "agents.list") {
+				return openclawResolvedAgentsEntriesOrRegistry(bin, timeout)
+			}
 			list, rerr := openclawRegistryAgentsList(bin, timeout)
 			return list, true, rerr
 		}
@@ -1148,7 +1177,10 @@ func openclawResolvedAgentsList(bin string, timeout time.Duration) ([]any, bool,
 	// key here is what selects the registry, so letting the envelope through
 	// as data would turn a graceful fallback into a failed preparation.
 	if message, isEnvelope := openclawJSONErrorMessage(trimmed); isEnvelope {
-		if strings.Contains(strings.ToLower(message), "agents.list") && isOpenclawKeyMissingMessage(message) {
+		if openclawKeyMissingMessageForPath(message, "agents.list") {
+			if openclawUnknownConfigPathMessage(message, "agents.list") {
+				return openclawResolvedAgentsEntriesOrRegistry(bin, timeout)
+			}
 			list, rerr := openclawRegistryAgentsList(bin, timeout)
 			return list, true, rerr
 		}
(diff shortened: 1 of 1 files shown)


multica-eve (changes requested):
Thanks for working on this compatibility issue. The intended user outcome is right: self-hosted members who upgrade should continue to launch their existing agents, and each task should still use only the content prepared for that task. The current implementation restores part of that compatibility, but it can make some agents read old content from the member's machine and can prevent default single-agent users from starting a task.

## Product verdict

✅ **The product direction should be accepted.** The compatibility break is real (verified against OpenClaw 2026.8.1 upstream), keeping an already-supported provider working across upgrades is worth maintaining, and we would extend the existing discovery path while preserving task isolation and the legacy fallback.

## Must-fix findings

### 1. Preserve task workspace isolation for `agents.entries`

In `server/internal/daemon/execenv/openclaw_config.go:1169-1173`, entries read from `agents.entries` are returned as `agentsFromRegistry=true`. `buildPerTaskOpenclawConfig` then drops all per-agent overrides at lines 550-554 and writes only `agents.defaults.workspace`.

OpenClaw resolves an explicit `agents.entries.<id>.workspace` before the default ([upstream implementation](https://github.com/openclaw/openclaw/blob/b1d6628564bac67a20232710a7896f59f9f9d3e2/src/agents/agent-scope-config.ts#L412-L435)). Therefore, a common config such as `main.workspace=/home/user/...` remains in the included config and wins over the task `workDir`. The task can load host `AGENTS.md` / `SOUL.md` / memory while missing the skills prepared by Multica, violating the existing intentional task-isolation contract.

Please retain the source as an explicit `legacy-list / keyed-entries / registry` state and emit a schema-valid `agents.entries` workspace override for keyed entries. Add regression coverage that asserts the effective workspace of every entry is `workDir`, including both the normal `$include` path and the managed-MCP snapshot path. The current test only asserts that legacy `agents.list` is absent, which locks in the incorrect behavior.

### 2. Handle a valid-but-unset `agents.entries` path

At `server/internal/daemon/execenv/openclaw_config.go:1174`, the code uses `isOpenclawKeyMissingResult`, which is hard-coded to recognize `agents.list`, to classify errors from the `agents.entries` query.

Current OpenClaw reports `Config path is valid but unset: agents.entries...` when the path is valid but not authored ([upstream implementation](https://github.com/openclaw/openclaw/blob/b1d6628564bac67a20232710a7896f59f9f9d3e2/src/cli/config-cli.ts#L119-L145)). With the real JSON error envelope plus a generic `exit status 1`, the helper matches neither `agents.list` nor the existing `not set` phrase, so preparation fails at line 1175 instead of reaching the `agents list --json` fallback. A normal installation relying on the implicit `main` agent can therefore stop working.

Please make the missing-path classifier accept and verify the target path, cover the upstream JSON-envelope shape with a generic process error, and scope `unknown config path` recognition to that target rather than broadening the generic helper used by `agents list` error handling.

## Validation

- `git diff --check` passed.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passed.
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/execenv/ -count=1` passed.
- CI is green, but the added tests do not exercise the two effective behaviors above.

🔄 **Request changes** — the direction is correct, but both issues must be fixed before merge because they break task isolation and current-schema default-install compatibility.

multica-eve (changes requested):
Self-hosted members who upgrade should be able to keep launching their existing agents, and every task should use only the content prepared for that task. This revision fixes the common path, but on slower machines an existing edge path can still make multi-agent tasks read old content from the host and can still stop default-agent tasks before they begin.

## Product verdict

✅ **The product direction should be accepted.** I independently verified the compatibility break against OpenClaw v2026.8.1. Preserving an already-supported runtime across upgrades and keeping task content isolated are both part of the existing product contract; I would extend the same discovery flow, explicitly distinguish the three supported sources, and make both CLI completion shapes produce the same decision.

## Must-fix findings

### 1. Handle an `agents.list` error envelope without an exit status through the `agents.entries` path

`server/internal/daemon/execenv/openclaw_config.go:1215-1222` handles the existing case where a complete JSON answer is accepted before the process exits, so `err == nil`. That branch sends every missing `agents.list` envelope directly to the registry. Unlike the non-zero-exit branch at lines 1184-1207, it never detects `Unknown config path: agents.list` and queries `agents.entries` first.

On OpenClaw 2026.8, an early-accepted `{"error":"Unknown config path: agents.list..."}` therefore produces only `agents.defaults.workspace`. The included config's explicit `agents.entries.<id>.workspace` still wins over that default, so the task can read the host workspace again. The existing `TestPrepareOpenclawConfigFallsBackToRegistryOnEnvelopeWithoutExit` already establishes that an envelope with `err == nil` is a supported execution shape, but the new test covers only a non-zero exit.

Please route both completion shapes through one source-aware decision. Add an `Unknown config path: agents.list` / `err == nil` regression that queries `agents.entries`, does not call the registry, and asserts every entry's effective workspace is `workDir` for both the normal `$include` and managed-MCP snapshot flows.

### 2. Handle an `agents.entries` missing envelope without an exit status, and keep path classification fail-closed

At `openclaw_config.go:1191-1203`, the missing-path classifier is used only when `entriesErr != nil`. If `agents.entries` returns a complete `{"error":"Config path is valid but unset: agents.entries..."}` with `err == nil`, `openclawAgentsEntriesList` turns it into an error at lines 1241-1242 instead of continuing to the registry. A normal installation relying on the implicit `main` agent can still fail before the task starts.

Also, `isOpenclawKeyMissingResultForPath` at lines 1600-1608 concatenates the process error and JSON envelope, then searches the combined text for the target and a missing phrase independently. Production process errors already include the command argument `agents.entries`, so an unrelated envelope containing `not set` or `missing key` can be misclassified as that target being absent and silently fall back. `isOpenclawUnknownConfigPathResult` at lines 1611-1620 has the same cross-fragment risk when stderr names a different path. This does not satisfy the previous request that the error diagnostic itself name the target, and it weakens the existing fail-closed contract.

Please classify structured envelopes and unstructured diagnostics separately, requiring the same diagnostic to contain both the target path and a supported missing-path meaning. Add regressions for: a valid-but-unset `agents.entries` envelope with `err == nil` falling back; an unrelated missing envelope failing closed; and a command string containing the target while the diagnostic names a different path failing closed. An explicit `legacy-list / keyed-entries / registry` source state, instead of `agentsFromRegistry` plus `__multica_agents_entries`, would also prevent these branches from diverging again.

## Validation

- Reviewed current head `ddb5f068dc68eed6261929aced03cb3f46c6e091`.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passed.
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/execenv/ -count=1` passed.
- GitHub CI has 15 passing checks, but the tests do not cover the two no-exit envelope paths above.
- The PR is currently conflicting with the latest `main`; please synchronize the branch after addressing the blockers.

If this is wrong, OpenClaw 2026.8+ tasks on slower hosts will fail before startup or read the host workspace; reverting this PR is the rollback, and self-hosted members using keyed multi-agent configuration or the implicit default agent are affected.

🔄 **Request changes** — the product direction is correct and the ordinary exit path is improved, but the two completion-shape gaps still break task isolation or default-install compatibility.

multica-eve (changes requested):
People who upgrade should keep using their existing agents without tasks reading older files from the machine, and a default setup should still start normally. This revision fixes both user-facing failures. The direction is right, but the change is not ready to merge because it leaves these exact behaviors without lasting protection and the promised upper bound on startup waiting no longer matches the actual flow.

## Product verdict

- Is the problem real? **Yes.** I independently verified the OpenClaw schema change and the previous failure modes.
- Is persistent compatibility support worth maintaining? **Yes.** Existing self-hosted members should not lose configured agents or task isolation when upgrading a supported runtime.
- Would we solve it this way? **Yes.** Extend the existing discovery chain, distinguish each supported source, preserve per-task workspace isolation, and fail closed on ambiguous errors.

✅ **The product direction should be accepted.** The current behavior fix also passes direct reproduction, but the implementation still has two merge blockers.

## Must-fix findings

### 1. Add permanent regression coverage for the new isolation and fallback paths

The PR changes only `server/internal/daemon/execenv/openclaw_config.go`; it adds no tests. The existing tests at `openclaw_config_test.go:1424-1468` and `openclaw_config_test.go:2044-2118` use the older `Config path not found` diagnostic, which goes directly to the registry and never exercises the new branches at `openclaw_config.go:1163-1164`, `1180-1183`, and `1196-1236`.

Please add committed regressions covering all of the following:

- an `Unknown config path: agents.list` envelope with `err == nil` queries `agents.entries` and does not query the registry;
- both the normal include and managed-MCP flows pin every `agents.entries.<id>.workspace` to the task workdir;
- a valid-but-unset `agents.entries` envelope with `err == nil` falls back to the registry;
- a missing-path envelope for an unrelated target remains fail-closed.

I ran temporary review tests for those cases and the current implementation passed. That confirms the code fixes the earlier blockers, but temporary reviewer coverage is not a substitute for repository tests: a later refactor could reintroduce host-workspace access or startup failure while the existing suite remains green.

### 2. Correct the worst-case CLI deadline model

`openclaw_config.go:1196-1203` creates a fresh full deadline for the new `agents.entries` query. The new-schema worst case therefore consumes four serial deadlines: config location, `agents.list`, `agents.entries`, and registry fallback. I measured four in a direct call-graph test, while `openclaw_config.go:126-148` still declares a maximum of three and `openclaw_config_timeout_test.go:212-271` deliberately expects the old path with no fourth query.

Please either share one deadline between the list/entries probes or re-derive the four-step ceiling and update the constant, documentation, and worst-case test. Leaving the invariant stale means its safety test now passes without covering the actual slowest path.

## Validation

- Reviewed head `a290dd1af6716e2832522864c2db935e0a717d23`.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passed.
- Full `execenv` tests passed on the PR head (70.905s).
- The PR merged cleanly with current `main` (`261522e3e3d71516d6b34d7c5ab1ccc8a4669718`); ByteIdentical and the full `execenv` suite also passed on that merge result (141.395s).
- GitHub reports 15 successful checks on the current head, but the committed suite does not cover the new keyed-entry path.

If this is wrong, OpenClaw 2026.8+ self-hosted members are affected: multi-agent tasks can regress to reading host workspace content or default-agent tasks can fail before startup. Reverting this PR is the rollback.

🔄 **Request changes** — the behavior fix is now correct in direct review testing; please commit the missing regressions and restore an accurate CLI deadline invariant before merge.

Bohan-J:
Thanks @wysl for the original work on this. The approach you started here was carried forward in #8678, which also addresses the review items from this PR, and it has now been merged (`d2423aa`). Closing this one as superseded. Thanks again for tracking down the `agents.entries` change.

PR #7906, opened 2026-09-02
MUL-6954 fix(daemon): send private chat deltas on resume

mameikagou:
## What does this PR do?

Stops replaying stable Multica private-chat instructions on every resumed turn.

A web/mobile direct-chat follow-up currently repeats the assistant role, `Audience: direct room`, outbound attachment tutorial, and `Task Initiator` block even though the provider session already contains those facts from the first turn. Besides spending tokens, this places platform boilerplate between every pair of user and assistant messages in the provider transcript.

The daemon now sends only the new turn data when a creator-only web/mobile private chat has a prior provider session. This selection happens before the provider backend, so it applies equally to Codex, Claude, Pi, Cursor, and future providers.

## Scope

The compact prompt is gated by all of these facts:

- the task belongs to a Multica chat session;
- a prior provider session is available;
- the chat has no external channel binding;
- the audience is direct;
- the task is not the initial proactive-introduction turn.

Group rooms and channel-backed conversations keep their existing per-turn audience and delivery instructions.

The first turn keeps the complete chat prompt. When a provider rejects a stale resume, the existing fresh-session retry clears the prior session ID and rebuilds the complete prompt before starting the replacement session.

## Turn-scoped context retained

The resumed private-chat prompt still includes data that can change each turn:

- the new user message;
- explicitly selected skills;
- inbound attachment download metadata;
- connected apps;
- shared-working-directory warnings;
- local worktree replay conflicts;
- a continuity notice when applicable.

Stable role, audience, initiator, and outbound-upload guidance remain in the provider session from the first turn and are omitted from successful follow-ups.

## Verification

- `cd server && go test ./internal/daemon -parallel=1 -count=1`
- `cd server && go test -race ./internal/daemon -run 'Test(BuildResumedWebDirectPrompt|ShouldUseResumedWebDirectPrompt|BuildTaskExecutionPrompt|BuildChatPrompt)' -count=1`
- `cd server && go vet ./internal/daemon`

All pass locally.

Tests also pin the provider-independent behavior, the full first-turn prompt, exclusion of group/channel-backed chats, and preservation of worktree-conflict context added on current `main`.

## Risk and fallback

The change is limited to resumed creator-only web/mobile chats. A missing prior session, a rejected resume, a group room, an external channel, or an intro turn uses the existing complete prompt. No provider session protocol or cancellation behavior changes.

## AI Disclosure

**Approach:** Traced prompt construction through `runTask`, separated stable private-chat context from turn-scoped data, moved the selection above provider backends for provider-independent coverage, and verified both the compact resume path and full-prompt fallbacks.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -7774,7 +7774,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 	if env.LocalWorktree != nil && len(env.LocalWorktree.ReplayConflicts) > 0 {
 		promptOptions = append(promptOptions, WithWorktreeReplayConflicts(env.LocalWorktree.ReplayConflicts))
 	}
-	prompt := BuildPrompt(task, provider, promptOptions...)
+	prompt := buildTaskExecutionPrompt(task, provider, promptOptions...)
 
 	// Pass task-scoped auth credentials and context so the spawned agent CLI
 	// can call the Multica API and the local daemon (e.g. `multica repo checkout`).
@@ -8144,7 +8144,7 @@ func (d *Daemon) runTask(ctx context.Context, task Task, provider string, slot i
 				execOpts.SystemPrompt = runtimeBrief
 			}
 		}
-		freshPrompt := BuildPrompt(task, provider, promptOptions...)
+		freshPrompt := buildTaskExecutionPrompt(task, provider, promptOptions...)
 
 		retryResult, retryTools, retryErr := d.executeAndDrain(ctx, backend, freshPrompt, execOpts, taskLog, task.ID, env.CodexHome, &msgSeq)
 		if retryErr != nil {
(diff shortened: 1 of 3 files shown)


multica-eve (changes requested):
Thanks for the contribution. We have decided not to accept this change. It removes a small amount of repeated guidance from follow-up private chats, but in long conversations it can cause the assistant to silently lose who it is speaking to and which privacy and access boundaries apply. That tradeoff is not acceptable for the modest savings.

The omitted audience and Task Initiator facts remain only in the first-turn transcript, so provider-side compaction can discard or weaken their only copy without triggering the full-prompt fallback. Moving them into the runtime brief is not a safe remedy: MUL-5377 and MUL-5730 deliberately keep these run/session-varying facts in the per-turn prompt so the cached brief remains byte-identical across triggers and issues.

The implementation also creates a parallel prompt assembly path that manually mirrors most of `perTurnContextBlocks` while omitting `BuildTaskInitiatorBlock`. This makes future per-turn additions easy to miss silently.

A narrower change could remove only content that already has an equivalent, compaction-safe copy in the brief, while continuing to send audience and initiator context every turn. That would save roughly 443 bytes per resumed turn and does not justify introducing another maintained prompt path at this time. We can revisit this when there is a provider-independent, non-compacted session-context channel.

PR #7908, opened 2026-09-02
fix(agent): normalise the provider prefix in the Hermes set_model skip-gate

garfield-nd:
## What does this PR do?

The Hermes `session/set_model` skip-gate added in #5690 (MUL-5029) compares the configured model against the model the session reports as current with a raw `==`:

```go
// server/pkg/agent/hermes.go
if opts.Model != "" && effectiveModel == sessionCurrentModel {
```

Those two values come from different id namespaces, so for an agent whose `model` is stored **bare** the comparison does not match whenever the runtime reports the prefixed form. The gate is dead code for those agents and `session/set_model` is replayed on every turn — which is the exact situation #5690 was opened to prevent.

**Why the two sides usually cannot match**

1. **Hermes always reports a provider-encoded id.** `_encode_model_choice` in its ACP adapter returns `f"{provider}:{model}"` whenever a provider resolves, and the provider falls back to `detect_provider()` or `"openrouter"`, so it is effectively never absent. Running that function over ordinary inputs:

   ```
   ('custom',    'claude-sonnet-4-5') -> 'custom:claude-sonnet-4-5'
   ('anthropic', 'claude-sonnet-4-5') -> 'anthropic:claude-sonnet-4-5'
   ('openai',    'gpt-4o')            -> 'openai:gpt-4o'
   ('',          'claude-sonnet-4-5') -> 'claude-sonnet-4-5'
   ```

   Confirmed end to end against a live `hermes acp` session — `initialize` + `session/new`, the same two calls this backend makes: `models.currentModelId` is populated from that encoder, so a bare configured id on a `custom` provider comes back with the `custom:` prefix attached. The last row is the only unprefixed case, and `_build_model_state` cannot reach it.

2. **`agent.model` is persisted verbatim.** `server/internal/handler/agent.go` writes `req.Model` straight into the column on create and on update. Nothing on that path normalises or validates a `provider:` prefix.

3. **The bare spelling is a documented, first-class input.** `server/cmd/multica/cmd_agent.go`'s own flag help is `--model  Model identifier (e.g. claude-sonnet-4-6, openai/gpt-4o)` — no prefix. Only the ACP-catalog dropdown produces the prefixed form, because `acpModelEntry` (`server/pkg/agent/models.go`) uses the advertised id as `Model.ID`.

So dropdown-configured agents match and skip; every agent configured through the CLI, the API, a seed, or a row predating the catalog does not, and never will.

**What the replay costs.** `session/new` already carries `model` (`buildHermesSessionParams`), so on a fresh session any provider mis-resolution has happened before the gate is reached; the gate's value is on the resume path.

Per turn, in the Hermes runtime: `set_session_model` replaces the session's agent with `_make_agent`, which re-runs `load_config` and `resolve_runtime_provider`, joins background MCP discovery for up to `mcp_discovery_timeout` (~1.5s), and constructs a fresh agent whose system prompt is then re-derived from disk.

More importantly it re-runs `_resolve_model_selection` — the provider auto-detection this gate exists to avoid. Running Hermes' own resolver over the ids this daemon sends reproduces the #5690 failure verbatim:

| session provider | id sent to set_model | resolved provider | changed? |
|---|---|---|---|
| `custom` | `deepseek-v4-pro` | `openrouter` | **yes** |
| `anthropic` | `claude-opus-4-5` | *(away from anthropic)* | **yes** |
| `custom` | `claude-sonnet-4-5` | `custom` | no |

> The resolved provider name in the middle column depends on which providers are configured on the host running the resolver, so you may see a different one; what is invariant is that a bare id can resolve away from the session's provider at all.

A provider change makes `_make_agent` drop `base_url` and `api_mode` and re-resolve them, which is how `custom:deepseek-v4-pro` ends up authenticating against OpenRouter. With the gate working, the request is never sent, so nothing re-detects.

**Why this approach**

`acpModelIDsEquivalent` normalises the optional prefix before comparing, rather than rewriting either side:

- either side empty → not equivalent, preserving the existing fall-through for runtimes that report no current model;
- **bare configured id → compare model names only.** It expresses no provider preference, so the session's own provider is authoritative. This is the case the gate exists for;
- **explicit configured provider → equivalent only when the session reports that same provider.** A session reporting a bare id cannot confirm the match, so we fall through and send `set_model` rather than swallow a switch the caller asked for.

The rules are deliberately **asymmetric**. An earlier revision of this PR compared model names whenever *either* side lacked a provider, which silently swallowed `openrouter:X` against a bare current id — a behaviour change from `origin/main`, caught in review and fixed in `7309d21a7`. Bare current ids are real: `acp_effort_test.go` carries unprefixed `currentModelId` values captured from two different runtimes, one of which routes through this same backend.

Only the first colon separates provider from model, so slash-bearing OpenRouter-style names (`nous:moonshotai/kimi-k2.6`) split correctly. Comparison is case-insensitive because Hermes lowercases the provider when it encodes the id.

No migration and no operator action: rows already written in the prefixed form keep working, because both spellings now resolve to the same selection.

**Deliberately not argued on cost.** A cache-hit-rate difference is what first pointed me here, but that measurement was grouped by "does the agent have a configured model", which on the deployment I measured was confounded with the transport wire (Hermes emits `cache_control` only for `api_mode: anthropic_messages` **and** a Claude-named model). I could not separate the two after the fact, so this PR does not claim it. It stands on the comparison being unsatisfiable and on the guard it restores.

## Related Issue

Issue creation is restricted on this repository, so the problem statement is above rather than in a linked issue.

_Issue creation is restricted on this repository for non-members, so the problem statement lives in this PR._

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/hermes.go` — the gate now calls `acpModelIDsEquivalent(effectiveModel, sessionCurrentModel)` instead of `==`, and logs `session_model` alongside the requested model when it skips.
- `server/pkg/agent/hermes.go` — new `splitACPModelID` (first-colon split, mirroring the prefix convention `acpModelEntry` already derives dropdown grouping from) and `acpModelIDsEquivalent`, placed next to `extractACPCurrentModelID`.
- `server/pkg/agent/hermes_test.go` — 12 `acpModelIDsEquivalent` cases and 6 `splitACPModelID` cases.

Scope is one backend. Every other ACP backend (`dim`, `grok`, `kiro`, `qoder`, `traecli`, `kimi`, `reasonix`) sends `session/set_model` unconditionally with no skip-gate at all; whether that costs them what it costs Hermes depends on each runtime's own handler, which I have not investigated and have not touched here.

## How to Test

1. Create a Hermes agent whose model is stored bare, the way the CLI documents it:
   ```
   multica agent create --runtime-id <hermes-runtime> --model claude-sonnet-4-6 ...
   ```
2. Run two turns and watch `~/.multica/daemon.log`.
   - Before: no `skipping redundant set_model` line ever appears; `hermes session model set` appears on every turn.
   - After: the second turn logs `hermes session already on requested model; skipping redundant set_model` with `session_model=custom:claude-sonnet-4-6`.
3. Switch the agent's model to an explicitly different provider (`openrouter:<same-model>`) and confirm `set_model` is still sent — the gate must not swallow a real provider switch.
4. `cd server && go test ./pkg/agent/ -run 'ACPModelIDs|SplitACPModelID'`

## Checklist

- [x] I have included a thinking path that traces from project context to this change — see *What does this PR do?*: the gate's stated purpose in #5690, the two id namespaces that meet at the comparison, the three code paths that write the bare spelling, and the runtime behaviour the replay triggers.
- [x] I have run tests locally and they pass — `go build ./...`, `go vet ./...`, `gofmt -l` clean, and `go test ./pkg/agent/` on a worktree branched from `origin/main`. The only failure in `pkg/agent` is `TestPiExecuteAttachesStdinPipe`, which fails identically on pristine `origin/main` in this environment (verified across two runs) and is unrelated.
- [x] If this change affects the UI, I have included before/after screenshots — N/A, daemon-side only
- [x] I have updated relevant documentation to reflect my changes — N/A; behaviour of a documented feature is restored, no documented contract changes
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — N/A
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` — N/A, no product copy
- [x] I have considered and documented any risks above — the one behavioural risk is swallowing a legitimate provider switch; that case is the second bullet of the equivalence rules and has a dedicated negative test (`same model, different explicit providers` → not equivalent)

## AI Disclosure

**AI tool used:** Multica Agent (Hermes runtime, Claude Opus), driven from a Multica chat session.

**Prompt / approach:**
I hit this while investigating why prompt-cache hit rates differed between agents on my own deployment. The workflow was investigation-first, not code-first: I had the agent read the Hermes ACP adapter source (`_encode_model_choice`, `set_session_model`, `_make_agent`, `_resolve_model_selection`), then verify each claim by *running* the code rather than reading it — a minimal ACP client that performs `initialize` + `session/new` against a real `hermes acp` process to capture `currentModelId`, and direct calls into `hermes_cli.models.parse_model_input` / `detect_provider_for_model` to see what a bare id actually resolves to. Only after the comparison was proven unsatisfiable did we write the helper and its table-driven tests.

The agent also talked me *out* of the original framing: my first draft justified the fix with a cache-hit-rate number, and reading the runtime's caching policy showed that number was confounded with an unrelated transport setting. That claim was removed from both the code comments and this PR — see the last paragraph of *What does this PR do?*

### DIFF
--- a/server/pkg/agent/hermes.go
+++ b/server/pkg/agent/hermes.go
@@ -2159,8 +2159,12 @@ func splitACPModelID(modelID string) (provider, model string) {
 //     that same provider. A session reporting a bare id cannot confirm the
 //     match, so fall through and send set_model rather than silently swallow a
 //     provider switch the caller explicitly asked for. Bare current ids are
-//     not hypothetical: acp_effort_test.go captures unprefixed ids from both
-//     `hermes acp` and jcode, and jcode routes through this same backend.
+//     not hypothetical: acp_effort_test.go captures an unprefixed
+//     `gpt-5.6-sol` from jcode, which routes through this same backend.
+//     Hermes Agent's encoder has a bare branch too — acp_adapter
+//     `_encode_model_choice` returns the model alone when the provider is
+//     empty — but a normally-configured install always resolves one, and a
+//     live `hermes acp` v0.20.0 session reports the prefixed form.
 //
 // Provider and model are compared case-insensitively: Hermes lowercases the
 // provider when encoding the id, so a config written as `Custom:...` would
--- a/server/pkg/agent/hermes.go
+++ b/server/pkg/agent/hermes.go
@@ -2142,53 +2142,34 @@ func splitACPModelID(modelID string) (provider, model string) {
 // acpModelIDsEquivalent reports whether a configured model id and the id the
 // runtime reports as current denote the same selection.
 //
-// Two id namespaces meet here. Hermes always answers session/new and
-// session/resume with a provider-encoded `provider:model` id — acp_adapter
-// server.py `_encode_model_choice` returns f"{provider}:{model}" whenever a
-// provider resolves, and the provider falls back to detect_provider() or
-// "openrouter", so it is effectively never absent. Confirmed against a live
-// `hermes acp` session: session/new populates models.currentModelId from that
-// encoder, so a bare configured `claude-sonnet-4-5` on a `custom` provider
-// comes back as `custom:claude-sonnet-4-5`.
-// agent.model, meanwhile, is persisted verbatim from the API with no prefix
-// normalisation (handler/agent.go CreateAgent/UpdateAgent), and the CLI
-// documents the bare spelling as valid (`--model claude-sonnet-4-6`). A raw
-// == across those two spellings can never be true, so the gate never fired.
+// Normalisation is needed because two id namespaces meet here: agent.model is
+// persisted verbatim (handler/agent.go CreateAgent/UpdateAgent) and the CLI
+// documents the bare spelling as valid, while an ACP runtime commonly answers
+// session/new with a provider-encoded `provider:model` id. A raw == across
+// those two spellings misses, which is why the MUL-5029 skip-gate added in
+// #5690 never fired for bare-configured agents and set_model was replayed
+// every turn.
 //
-// What the redundant call costs, per turn, verified in the runtime source:
-// set_session_model replaces state.agent with SessionManager._make_agent,
-// which re-runs load_config + resolve_runtime_provider, joins background MCP
-// discovery for up to mcp_discovery_timeout (~1.5s), constructs a fresh
-// AIAgent, and re-derives all three system-prompt tiers. It also re-runs the
-// provider auto-detection this gate exists to avoid, and it is the only
-// session mutation that skips _schedule_mcp_late_refresh — so the tool
-// snapshot can drift between turns, and on Hermes' direct-native Anthropic
-// tool-cache layout the tools array sits ahead of the system prompt inside
-// the cached prefix. A drifted tool set therefore invalidates the whole
-// prefix, and a model id remapped by _resolve_model_selection additionally
-// fails _stored_prompt_matches_runtime and forces a full prompt rebuild.
-//
-// Deliberately NOT justified by a cache-hit-rate number. The 6.7%-vs-90.2%
-// split that first pointed here was grouped by "does the agent have a
-// configured model", which on the measured deployment was confounded with the
-// transport wire: Hermes emits cache_control breakpoints only when api_mode is
-// anthropic_messages AND the model is Claude-named (verified against
-// agent.agent_runtime_helpers.anthropic_prompt_cache_policy), and that setting
-// was changed five days before this gate was touched. This fix stands on
-// correctness — a comparison that cannot be true, guarding a call the runtime
-// documents as hazardous — not on a cost figure.
-//
-// Equivalence rules:
+// The rules are deliberately asymmetric, because the risk is:
 //   - Either side empty → not equivalent (caller handles the empty-model case).
-//   - Both sides carry an explicit provider → compare the whole id. A genuine
-//     provider switch (openrouter:X → custom:X) must still send set_model.
-//   - Exactly one side carries a provider → compare model names only. A bare
-//     configured id expresses no provider preference, so the session's own
-//     provider is authoritative and re-selecting it would be pure downside.
+//   - Bare configured id → equivalent on the model name alone. It expresses no
+//     provider preference, so the session's own provider is authoritative.
+//     This is the case the gate exists for.
+//   - Explicit configured provider → equivalent only when the session reports
+//     that same provider. A session reporting a bare id cannot confirm the
+//     match, so fall through and send set_model rather than silently swallow a
+//     provider switch the caller explicitly asked for. Bare current ids are
+//     not hypothetical: acp_effort_test.go captures unprefixed ids from both
+//     `hermes acp` and jcode, and jcode routes through this same backend.
 //
 // Provider and model are compared case-insensitively: Hermes lowercases the
 // provider when encoding the id, so a config written as `Custom:...` would
 // otherwise miss.
+//
+// splitACPModelID cannot distinguish a provider prefix from a colon inside a
+// bare model name (Ollama-style `llama3:8b`). That degradation is safe in the
+// only direction that matters: `llama3:8b` vs `custom:llama3:8b` compares
+// unequal and falls back to sending set_model, never to a false skip.
 func acpModelIDsEquivalent(configured, current string) bool {
 	configured = strings.TrimSpace(configured)
 	current = strings.TrimSpace(current)
@@ -2200,10 +2181,10 @@ func acpModelIDsEquivalent(configured, current string) bool {
 	if !strings.EqualFold(cfgModel, curModel) {
 		return false
 	}
-	if cfgProvider != "" && curProvider != "" {
-		return strings.EqualFold(cfgProvider, curProvider)
+	if cfgProvider == "" {
+		return true
 	}
-	return true
+	return strings.EqualFold(cfgProvider, curProvider)
 }
 
 // extractACPCurrentModelID pulls the model selected by the ACP runtime out of
--- a/server/pkg/agent/hermes.go
+++ b/server/pkg/agent/hermes.go
@@ -574,9 +574,19 @@ func (b *hermesBackend) Execute(ctx context.Context, prompt string, opts ExecOpt
 		// downside. An empty sessionCurrentModel (older runtime or unparsable
 		// state) falls through and still sends set_model, preserving prior
 		// behaviour. See MUL-5029 / NousResearch/hermes-agent#59089.
-		if opts.Model != "" && effectiveModel == sessionCurrentModel {
+		//
+		// The comparison is provider-normalised (acpModelIDsEquivalent), not a
+		// raw string match: Hermes always reports its current model in the
+		// provider-encoded `provider:model` form, while agent.model is stored
+		// verbatim from the API and is routinely bare. A literal == therefore
+		// never matched for those agents, so the gate above was dead code and
+		// the MUL-5029 mis-routing hazard it exists to prevent stayed live for
+		// exactly the agents that hit it. See acpModelIDsEquivalent for the
+		// evidence and for what the redundant call actually costs.
+		if opts.Model != "" && acpModelIDsEquivalent(effectiveModel, sessionCurrentModel) {
 			b.cfg.Logger.Info("hermes session already on requested model; skipping redundant set_model",
 				"model", opts.Model,
+				"session_model", sessionCurrentModel,
 				"session_id", sessionID,
 			)
 		} else if opts.Model != "" {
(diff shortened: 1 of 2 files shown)


Bohan-J (changes requested):
Thanks for this — the investigation is unusually thorough, and I was able to reproduce your core claim exactly. The bug is real and the approach (normalise at the comparison, rewrite neither side, no migration) is the right one.

I verified the following locally against this branch, diffing behaviour against `origin/main` using the repo's own frame-recording stub (`fakeACPRecordingScriptWithCurrentModel`) so I could see whether `session/set_model` actually goes out on the wire:

- **The gate really is dead for bare-configured agents.** On `origin/main`, `Model: "deepseek-v4-pro"` with a session reporting `custom:deepseek-v4-pro` sends `session/set_model` every turn. On this branch it is correctly skipped. Confirmed.
- Your supporting claims check out: `agent.model` is persisted verbatim (`server/internal/handler/agent.go:1420`, `:1887`), the CLI documents the bare spelling (`server/cmd/multica/cmd_agent.go:165`), and `acpModelEntry` uses the advertised id as `Model.ID` (`server/pkg/agent/models.go:2235`).
- All 18 new unit tests pass; CI is green.

There is one thing I think has to change before merge, plus a test gap and some comment cleanup.

---

## Blocking: the third equivalence rule is symmetric, but the risk is not

> exactly one side carries a provider → compare model names only

That rule is safe in the direction this PR is about (bare configured, provider-encoded current). It is not safe in the other direction — explicit configured provider, bare current:

```go
acpModelIDsEquivalent("openrouter:hermes-4", "hermes-4") // => true, set_model is skipped
```

So a member who explicitly asked for `openrouter:` gets no switch, silently, and the session keeps whatever provider it defaulted to. On `origin/main` that same input **does** send `set_model`, so this is a behaviour change introduced here. I confirmed it end to end on both branches with this test:

```go
func TestExplicitProviderVsBareCurrent(t *testing.T) {
	recordPath := filepath.Join(t.TempDir(), "frames.jsonl")
	fakePath := filepath.Join(t.TempDir(), "hermes")
	writeTestExecutable(t, fakePath, []byte(
		fakeACPRecordingScriptWithCurrentModel(recordPath, "ses_new", "hermes-4")))

	backend, _ := New("hermes", Config{ExecutablePath: fakePath, Logger: slog.Default()})
	session, _ := backend.Execute(context.Background(), "prompt-ignored", ExecOptions{
		Timeout: 5 * time.Second,
		Model:   "openrouter:hermes-4", // member picked a provider explicitly
	})
	// ... drain messages, await result ...
	assertNoRecordedFrame(t, recordPath, "session/set_model")
}
```

It fails on `origin/main` (the frame is sent, as it should be) and passes here (the frame is gone). The daemon log line on this branch reads:

```
hermes session already on requested model; skipping redundant set_model
  model=openrouter:hermes-4 session_model=hermes-4
```

This also contradicts the guarantee the PR description gives — that an explicit `openrouter:X` still switches. That holds only when the session *also* reports a prefixed id.

**Why this is reachable, not theoretical.** The rule is justified by "Hermes always reports a provider-encoded id, so the bare-current case cannot happen." This repo contains two counterexamples, both captured from real runtimes:

- `server/pkg/agent/acp_effort_test.go:411` — `hermesAgentSessionResult`, whose own comment says *"Captured from `hermes acp` v0.20.0 on 2026-08-11 — the keys are exactly these four"*. Its `currentModelId` is **`hermes-4`**, bare.
- `server/pkg/agent/acp_effort_test.go:403` — `jcodeEffortSessionResult`, verified by the reporter against jcode v0.71.1 and v0.73.0. `currentModelId` is `gpt-5.6-sol`, also bare. And jcode routes through this exact backend — `hermes.go:253` states it covers "two unrelated binaries — Hermes Agent and jcode".

`_encode_model_choice` returning a prefix is what happens *when a provider resolves*; it isn't a property of every input this backend sees.

**Suggested fix** — make the rule asymmetric. Two lines, and it costs nothing you were trying to buy:

```go
if cfgProvider == "" {
	// A bare configured id expresses no provider preference, so the
	// session's own provider is authoritative. This is the case the
	// gate exists for.
	return true
}
// The configured id names a provider explicitly. Only treat it as
// equivalent when the session reports the same one; if the session
// reports no provider we cannot confirm a match, so fall through and
// send set_model as before.
return strings.EqualFold(cfgProvider, curProvider)
```

Bare-configured + prefixed-current still skips, so the bug you found stays fixed. An explicit provider choice is never swallowed.

## Should have: an end-to-end regression test

The new tests cover `acpModelIDsEquivalent` itself, which is worth having, but nothing asserts that the gate now actually fires — and that is the defect. `fakeACPRecordingScriptWithCurrentModel` (`hermes_test.go:3316`) already exists for exactly this, with two sibling tests next to it (`TestHermesSkipsRedundantSetModelWhenAlreadyCurrent` / `TestHermesSendsSetModelWhenModelDiffersFromCurrent`). A third in the same shape — bare `Model`, `custom:`-prefixed `currentModelId`, assert no `session/set_model` frame — is a few lines and pins the fix at the level the bug lives at. I ran that version too; it passes here and fails on `main`.

Whichever semantics you settle on for the case above should get a table entry as well: the current 12 cases have no `provider:model` vs bare-id row, which is why the regression isn't caught.

## Comment cleanup

- The PR description says the cache-hit-rate claim "was removed from both the code comments and this PR", but the comment on `acpModelIDsEquivalent` still carries the whole `The 6.7%-vs-90.2% split ...` paragraph. Either the deletion was missed or the description overstates it.
- The paragraph immediately above it argues the benefit via `_schedule_mcp_late_refresh` and tool-cache prefix invalidation, and is then followed by "Deliberately NOT justified by a cache-hit-rate number." Those read as arguing against each other.
- ~50 lines of comment on a 15-line function, most of it describing Hermes' Python internals (`_make_agent`, `_stored_prompt_matches_runtime`, the `~1.5s` discovery timeout). This repo can't verify or track those, so they will rot into something misleading. I'd cut it to the equivalence rules, one sentence on why normalisation is needed, and a pointer to #5690 / MUL-5029 — the investigation is already preserved in the commit message.

## Non-blocking notes

- "the comparison can never be true" is stronger than the evidence supports: it holds when the runtime reports a prefixed id, which is not always. Worth softening, since the asymmetric rule above depends on the same premise.
- `splitACPModelID` can't tell a provider prefix from a colon inside a bare model name (Ollama-style `llama3:8b`). I checked the degradation direction and it's safe — `llama3:8b` vs `custom:llama3:8b` compares unequal and falls back to sending `set_model`, i.e. today's behaviour, never a false skip. Worth one line in the comment; no code change needed.
- `session/new` already passes `model` (`buildHermesSessionParams`), so on a fresh session any provider mis-resolution has already happened before the gate is reached. The real value here is the resume path — the description could scope that more precisely.
- Agreed on leaving the other ACP backends alone; scoping that out explicitly was the right call.

Happy to re-review as soon as the first point is addressed — the rest is small.

garfield-nd:
Thank you — this is a better review than the PR deserved, and the blocking
point is correct. I reproduced it before changing anything.

## Blocking point: fixed, and it failed exactly as you described

I wrote your counterexample as a table entry first and watched it fail on the
committed branch:

```
--- FAIL: TestACPModelIDsEquivalent/explicit_configured_provider_vs_bare_current
        acpModelIDsEquivalent("openrouter:hermes-4", "hermes-4") = true, want false
```

Then as an end-to-end frame test, which failed the same way — the frame is
simply gone:

```
--- FAIL: TestHermesSendsSetModelWhenExplicitProviderAndBareCurrent
        no recorded frame for method "session/set_model"
```

I also confirmed the two fixtures you cited, because the whole third rule was
resting on the premise they falsify:

- `acp_effort_test.go` — `hermesAgentSessionResult` carries
  `"currentModelId":"hermes-4"`, bare, and its own comment says it was captured
  from `hermes acp` v0.20.0.
- the same file — `jcodeEffortSessionResult` carries
  `"currentModelId":"gpt-5.6-sol"`, also bare.

So "Hermes always reports a provider-encoded id" was wrong, and the repository
already held the evidence. `_encode_model_choice` returning a prefix is what
happens when a provider resolves; I read that as a property of every response
and it is not.

Adopted your shape verbatim:

```go
if cfgProvider == "" {
	return true
}
return strings.EqualFold(cfgProvider, curProvider)
```

Bare-configured plus prefixed-current still skips, so the original defect stays
fixed. An explicit provider is never swallowed.

## End-to-end regression test: added, both directions

You were right that asserting on `acpModelIDsEquivalent` alone would not have
caught the defect — the gate never firing is the bug, and nothing tested that.

- `TestHermesSkipsRedundantSetModelForBareConfiguredModel` — bare `Model`,
  `custom:`-prefixed `currentModelId`, asserts no `session/set_model` frame.
  This is the one that pins the original defect.
- `TestHermesSendsSetModelWhenExplicitProviderAndBareCurrent` — the guard for
  the rule above. Verified it fails without the asymmetric rule.

Both follow `TestHermesSkipsRedundantSetModelWhenAlreadyCurrent`'s shape and
reuse `fakeACPRecordingScriptWithCurrentModel`. Three table entries added as
well, including the missing `provider:model` vs bare-id row that let this
through.

Full `go test ./pkg/agent/`: the only failure is
`TestPiExecuteAttachesStdinPipe`, unchanged from `origin/main` in this
environment. `go vet` and `gofmt -l` clean.

## Comment cleanup: all three, and one correction to the description

- The `6.7%-vs-90.2%` paragraph is gone. You were right that the description
  overstated it — the deletion had landed on a different branch and never
  reached this one. That was my error, not a wording problem.
- The paragraph arguing the benefit via MCP refresh and tool-cache prefix
  invalidation is gone too. Your reasoning applies: it described Python
  internals this repository cannot verify or track, and it did read as arguing
  against the sentence immediately after it.
- Cut to the equivalence rules, one sentence on why normalisation is needed,
  and the `#5690` / MUL-5029 pointer. The investigation stays in the commit
  messages.

## Non-blocking

- "can never be true" softened. It holds when the runtime reports a prefixed
  id, which is exactly the premise that turned out to be conditional.
- Added the `llama3:8b` line, with the direction you established: it compares
  unequal against `custom:llama3:8b` and falls back to sending `set_model`,
  never to a false skip. No code change.
- Agreed that `session/new` already carries `model`, so the value is on the
  resume path. I will scope the description to that rather than implying the
  fresh-session path benefits.
- Thanks for confirming the other ACP backends were right to leave alone.

Ready for re-review whenever you have time.

Bohan-J (approved):
Thank you — this was a genuinely good exchange. You reproduced the blocking point before changing anything, adopted the asymmetric rule, and added the end-to-end frame tests in both directions, including the `provider:model` vs bare-id table row that was missing. `go vet`, `gofmt -l` and the full `go test ./pkg/agent/` are clean on my machine (`TestPiExecuteAttachesStdinPipe` passes here, so that flake looks environmental on your side rather than anything you introduced).

I pushed one commit to your branch (`34edc4b`) rather than sending you round again for a comment-only change — please sanity-check it.

**It corrects something I got wrong in my review**, which you then took on trust and wrote into the code. I claimed `acp_effort_test.go` proves `hermes acp` reports unprefixed model ids. Only the jcode half of that holds.

I checked against a live `hermes acp` **v0.20.0** — the same version the fixture names — driving `initialize` + `session/new`, the two calls this backend actually makes:

| `session/new` model param | `models.currentModelId` |
|---|---|
| *(none)* | `openai-codex:gpt-5.6-terra` |
| `gpt-5.6-terra` *(bare)* | `openai-codex:gpt-5.6-terra` |
| `openai-codex:gpt-5.6-terra` | `openai-codex:gpt-5.6-terra` |

All nine `availableModels` ids were prefixed too. So the bare `hermes-4` in that fixture is better read as a simplification made while writing a test about `configOptions`, not a verbatim capture.

Two things worth having from that, both now in the comment:

- **Your core claim is confirmed first-hand.** Row two is the defect: a bare configured id comes back provider-encoded, so the old `==` could not match and `set_model` was replayed every turn. That is no longer inferred from reading the adapter — it is what the runtime does.
- **The bare branch is real but not reachable normally.** `_encode_model_choice` does return the model alone when the provider is empty (I called it directly to confirm), but `_build_model_state` defaults the provider via `state.agent.provider or detect_provider() or "openrouter"`, and `normalize_provider("")` falls back to `openrouter`. So a normally-configured install never gets there.

None of this touches the rule, which stands for the reason that mattered all along: this backend serves two unrelated binaries, jcode really does report unprefixed ids, and declining to confirm an explicit provider against a bare current id costs nothing. Your original instinct to fix the comparison rather than normalise either side was right, and the investigation write-up in the commit messages is a good record to leave behind.

Merging once CI is green.

PR #7909, opened 2026-09-02
MUL-6959: WS-4679: block done while linked PRs remain open

tangyuanjc:
## Summary

- reject manual transitions into built-in or custom done-category statuses while any visible GitHub or self-hosted working PR is open or draft
- return HTTP 409 with stable code open_pull_requests_block_done and the blocking PR count
- preflight all batch targets before the first mutation to prevent partial delivery claims
- keep reference-only body mentions non-blocking and document the new contract in the built-in issue workflow skill

## Verification

- focused database regression: PASS
- full server/internal/handler suite: PASS (47.457s)
- broad go test ./...: changed handler, status, VCS integration, service, migration, and all other non-CLI packages passed; server/cmd/multica failed only because this daemon-managed task workdir marker intentionally rejects tests that assume an ordinary human CLI context

Issue: WS-4679

### DIFF
--- a/server/internal/handler/issue.go
+++ b/server/internal/handler/issue.go
@@ -240,6 +240,65 @@ func writeIssueStatusRaceError(w http.ResponseWriter, err error) bool {
 	return false
 }
 
+// issueDoneBlockedByOpenPullRequestsError is returned when a manual status
+// write would claim delivery while a linked working PR is still in flight.
+// Reference-only links are deliberately absent from OpenCount because they are
+// hidden from the issue's PR list and do not represent work on the issue.
+type issueDoneBlockedByOpenPullRequestsError struct {
+	OpenCount int64
+}
+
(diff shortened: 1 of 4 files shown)


multica-eve:
Thanks for the contribution. We are closing this PR for product-direction reasons, not because of implementation or test quality.

An issue status is the team's statement about the work. A linked pull request is supporting evidence and an automation signal; it should not override an explicit user or agent decision. This change makes any visible open or draft link a hard veto, including links created from a title or branch without an explicit `Closes`, `Fixes`, or `Resolves` intent. It also has no in-product unlink or override path, and retained records can become stale after an integration is disconnected.

We want to preserve the existing contract: explicit close intent may automatically move an issue to Done after all linked work settles. If we address accidental completion, the product direction should be a non-blocking warning or confirmation, or a separately designed opt-in policy with an override and stale-link handling.

Rejecting this approach on product grounds. Thank you for the thoughtful implementation.

PR #7911, opened 2026-09-02
MUL-6960: Surface merge-queue state on the PR row and the board card

bustinjailey:
## What

A pull request waiting in a GitHub merge queue was invisible in Multica. This makes it visible in the two places it matters: the issue detail PR row, and the board card.

Closes the gap reported in BUS-231, and carries the earlier BUS-219 work that captured the merge-queue state but was never opened as a PR.

## Why the row still said "Open"

A queued PR is still `state === "open"`, and GitHub reports its `mergeStateStatus` as `blocked` — branch protection routes the merge through the queue. So the detail row painted it the same green "Open" as a PR nobody had touched, and reading `mergeStateStatus` literally would have labelled it "Blocked", the opposite of what it is.

`deriveMergeStatus` therefore ranks merge-queue membership above every `merge_state_status` verdict, but below a conflict — an entry GitHub is about to evict is not on its way in. The row's leading icon and state word now carry that, and the merge element falls silent so the queue is stated exactly once.

Terminal PRs are excluded throughout: the snapshot column keeps whatever the PR last reported while open, so a merged PR can still carry `queued`.

## The board card

The board card payload carries no pull request data at all, and the only PR endpoint is per-issue — a card indicator that way would mean one request per visible card.

This adds `GET /api/workspaces/{id}/github/pull-requests/queued`, returning only the issues whose PR is in a queue. Only a handful of PRs are ever queued, so the response stays small regardless of board size, and the board fetches it once per view. The indicator follows `IssueAgentActivityIndicator`: subscribe to the one shared query, narrow to this issue with a `select`, render nothing otherwise.

The query mirrors `currentGitHubSnapshotAvailable` — current-head snapshot, non-terminal PR, no `reference_only` links — so the board and the detail card can never disagree about the same PR.

## Non-obvious calls

- **The chip sits in card row 1**, not the chip row below, because that row only mounts when some other chip exists and whether this one renders is known only inside the indicator.
- **The indicator renders in every column**, not just In Review. A queued PR is worth seeing wherever the card is.
- **`unmergeable` is styled apart** from the other entry states — it is the one that is not progress.
- **The poll is gated on `useGitHubSettings().enabled`**, since every card mounts one and a workspace with GitHub off would otherwise poll for a permanently empty list.
- **Migration numbered 446.** It was 441 when the branch was written; main has since taken 441–445.

## Test plan

Run against a real PostgreSQL 17 and a locally built Go 1.26.6 toolchain.

- `go build ./...` — clean; `go vet` clean on touched packages.
- `go test ./internal/handler/ ./internal/integrations/ghsnapshot/ ./cmd/server/` — all pass. Note the handler package silently skips its whole suite when no database is reachable; these were run with one up, so they actually executed.
- New `TestQueuedPullRequestQueryFiltering` runs the real query against Postgres. Each case differs from the included one by a single column (not queued / stale head / never fetched / merged / reference-only), so a filter that stops applying surfaces as exactly one failure.
- New `TestQueuedPullRequestsEndpoint` covers what the query test cannot: that the route is registered under the workspace group and survives auth.
- Full migration chain applied from an empty database, ending at 446.
- `pnpm typecheck`, `pnpm lint` (0 errors), `packages/core` (1689) and `packages/views` (4930) suites pass.

Not verified: the rendered UI was not exercised in a running browser — the merge-queue path needs a GitHub App key and a repository with a queue configured, neither available here. Component-level rendering is covered by the jsdom tests instead. The one unrelated failure in this environment is `apps/desktop`, whose Electron binary is not installed (`node_modules/.pnpm/electron@39.8.7/.../dist` missing); this branch touches no desktop file.


multica-eve (changes requested):
Thanks for addressing the visibility gap around GitHub merge queues. I agree that this is useful as a read-only delivery-progress signal, but the current implementation has four blockers:

1. **A missed dequeue event can leave a PR permanently shown as queued.** The stale-snapshot fallback in `server/pkg/db/queries/github_snapshot.sql` excludes queued entries, while the board query in `server/pkg/db/queries/github.sql` has no snapshot-age guard. The UI then promotes that value to the primary status. Please ensure stale queue state expires or is reconciled even when the dequeue webhook is missed.

2. **Polling is owned by every card, and realtime invalidation misses the new query key.** Each `IssueMergeQueueIndicator` mounts its own 60-second observer. Sharing a React Query cache key does not make those observers share one polling timer. In addition, the existing invalidation for `["github", "pull-requests"]` does not match `["github", workspaceId, "pull-requests", "queued"]`. Please move polling to one board-level owner, distribute the result to cards, and invalidate the exact queued-PR query when realtime events arrive.

3. **GitHub App permission and upgrade behavior are not verified.** The new GraphQL fields depend on GitHub's merge-queue data, but the self-hosting permission docs and the PR test plan do not establish that existing installations can read them. Please verify the query with an installation token. If `Merge queues: Read-only` is required, update the App configuration and self-hosting docs, and document the approval/upgrade path for existing installations. Also make sure an authorization error for these fields cannot discard the rest of the PR snapshot.

4. **The migration prefix collides with `main`.** `446_github_pr_merge_queue_state` conflicts with `446_issue_properties_bigm_index`, which is why `backend-tests` and the aggregate `backend` check fail. Please rebase and renumber this migration to the next free prefix.

Once these are addressed, the UI should continue treating queue state as GitHub delivery progress—not as a new Multica issue status and not as a signal to mark an issue Done.

PR #7914, opened 2026-09-02
fix(codex): keep in-place brief out of tracked AGENTS.md

### LINKED ISSUE #7879: local_directory (in-place) + Codex runtime: managed AGENTS.md block dirties the repo's tracked AGENTS.md during runs
## Summary

When a project has a `local_directory` resource (in-place mode) and the agent runs on the **Codex** runtime, the daemon merges its auto-managed instruction block (`<!-- BEGIN MULTICA-RUNTIME (auto-managed; do not edit) -->` … ) directly into the repository's own **tracked** `AGENTS.md`, and creates `.multica/` and `.agent_context/` inside the repo root. For the duration of the run, `git status` shows `AGENTS.md` as modified (+250 lines) plus two untracked directories.

Any agent whose instructions say "stop if the working tree has uncommitted changes" will therefore refuse to work — it mistakes the daemon's own injection for someone else's uncommitted changes. In our workspace this blocked two issues in a row (the QA agent marked both `blocked` and posted "tracked file `AGENTS.md` has uncommitted changes, cannot proceed").

## Environment

- Multica CLI / daemon `v0.4.37`, Linux
- Runtime: `codex app-server`, codex-cli 0.150.1
- Project resource: `local_directory` → `/home/agent/bird-archive-agents` (a git clone whose root already contains tracked `AGENTS.md` and `CLAUDE.md`)

## Steps to reproduce

1. Clone any repo that has a tracked `AGENTS.md` at its root.
2. Add it to a project as a `local_directory` resource (in-place).
3. Assign an issue to an agent bound to a Codex runtime.
4. While the run is active, `git status --short` inside the directory shows:
   ```
    M AGENTS.md
   ?? .agent_context/
   ?? .multica/
   ```
   `git diff --stat AGENTS.md` shows ~254 added lines — the `MULTICA-RUNTIME` managed block.
5. After the run ends the daemon restores the file and removes the dirs, so the tree is clean again — the pollution is only visible from inside the run, which is exactly where the agent checks it.

Daemon log excerpt:

```
INF local_directory: lock acquired local_directory=/home/agent/bird-archive-agents
INF starting agent provider=codex workdir=/home/agent/bird-archive-agents
INF codex lifecycle phase=spawn cwd=/home/agent/bird-archive-agents ...
```

Before we switched to `local_directory`, the same agent ran in the isolated task workdir (`<ws>/nat-17-xxxx/workdir/`), where the managed `AGENTS.md`, `.agent_context/`, `.multica/` lived one level *above* the `bird-archive/` clone, so the repo's git status was untouched.

## Expected

The runtime's managed instruction file and sidecar dirs should not dirty the user's git working tree. Options, roughly in order of preference:

1. In in-place `local_directory` mode, keep the managed `AGENTS.md` / `CLAUDE.md` and sidecar dirs **outside** the repo (e.g. in the task dir) and point the runtime at them — Codex supports `AGENTS.md` discovery from parent directories / `CODEX_HOME`, so a parent-level file or a `codex-home` config should work without touching the repo.
2. If writing into the repo is unavoidable, at least add the sidecar dirs to `.git/info/exclude` automatically, and document that the managed block will appear in tracked `AGENTS.md` during runs so agent authors can exclude it from "dirty tree" checks.

## Workaround we applied

Changed the agents' entry check to `git status --short -- . ':!AGENTS.md' ':!CLAUDE.md' ':!.multica' ':!.agent_context'` and added `.multica/` + `.agent_context/` to `.git/info/exclude`.

iiwish:
> [!NOTE]
> This is a stacked draft on #7123. Review the single commit after `bc545fe0a`; the base PR supplies the task-scoped Git protection for runtime sidecars.

Fixes #7879.

## Problem

A Codex task using `local_directory` in `in_place` mode appended the Multica runtime brief to the repository's tracked `AGENTS.md`. During the run, agent-side Git therefore saw the daemon's own modification and safety-oriented agents refused to start work.

The first inline implementation used app-server `developerInstructions`. Upstream Codex treats that field as a replacement for effective `developer_instructions`, not an additive layer, and the resume override can be stale for the first resumed turn. That would keep `AGENTS.md` clean at the cost of silently dropping user/project instructions.

## Change

- Keep `developerInstructions: null` on both `thread/start` and `thread/resume`, preserving the user's effective Codex configuration.
- For Codex CLI >= `0.121.0`, append the Multica runtime brief as a developer-role Responses API item with `thread/inject_items` after start/resume and before the first turn.
- Keep file delivery for unknown or older Codex versions, where `thread/inject_items` does not exist. This deliberately prefers temporary repository pollution over losing either instruction source.
- Bound brief injection with the startup handshake timeout and fail closed if it is rejected. Diagnostics retain only the method/code/timeout classification, never a provider message or stderr that could echo the brief.
- Factor the daemon's real file-vs-inline decision into one production helper and test that exact boundary against a real Git repository with a tracked `AGENTS.md`.

## Verification

- `go test ./pkg/agent -run 'TestCodex|TestIsCodex|TestApplyCodex|TestExtractThreadID|TestNilIfEmpty' -count=1` (pass, 94.985s)
- `go test ./internal/daemon/execenv -count=1` (pass, 31.508s)
- `go test ./internal/daemon -count=1` (pass, 46.445s)
- `go test -race ./pkg/agent -run 'TestCodex(ThreadStartKeepsDeveloperInstructionsNullWithSystemPrompt|ThreadResumeKeepsDeveloperInstructionsNullWithSystemPrompt|InjectSystemPromptAppendsDeveloperHistory|ExecuteInjectsBriefBetweenThreadStartAndFirstTurn|ExecuteInjectionFailureDoesNotPersistBrief)' -count=1`
- `go test -race ./internal/daemon -run 'TestProviderSupportsInlineSystemPrompt|TestPrepareRuntimeBriefDeliveryKeepsInPlaceCodexAgentsFileExact' -count=1`
- `go vet ./pkg/agent ./internal/daemon/execenv ./internal/daemon`
- `git diff --check`

A first combined daemon run hit the pre-existing timing-sensitive `TestRunTask_PrepareTimeoutStopsLeaseDuringBlockedStartTask`; it passed 10 consecutive isolated runs, and the subsequent full daemon run passed.

Real protocol smoke against Codex CLI `0.151.0-alpha.7.2` passed: `thread/inject_items` accepted a developer-role item, persisted it, and the rollout contained both a user-config `developer_instructions` canary and the Multica runtime-brief canary. Source/tag inspection confirms the method is absent through `0.120.0` and present in `0.121.0`.

Negative controls also discriminate the fixes:

- Restoring non-null `thread/start`/`thread/resume.developerInstructions` fails both config-preservation tests.
- Removing Codex from the version-aware inline branch fails the daemon delivery regression before the tracked-file guarantee can pass.

The broad `go test ./pkg/agent -count=1` was attempted before this re-review and remains timing-sensitive on this host: unrelated 5-second backend fixtures can exceed their setup window under package-wide load. The complete Codex-focused set above passes.

## Stack status

#7123 still conflicts with the latest `main` in its daemon/execenv base changes. This PR remains draft and should be rebased or retargeted only after that foundation is updated or merged.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -5626,20 +5626,26 @@ func providerNeedsInlineSystemPrompt(provider string) bool {
 // flow can choose the delivery that does not write into the user's repository
 // (GitHub #7114).
 //
-// Membership is a property of this repository's backend code, not of the
-// vendor CLI, so it is verifiable by reading it: each provider below prepends
-// SystemPrompt to the prompt it sends (or passes it as --append-system-prompt)
-// whenever the field is non-empty — see grok.go, kimi.go, kiro.go, qoder.go,
-// qwenpaw.go, traecli.go, openclaw.go and codebuddy.go. Everything else is
-// absent deliberately: claude, pi, opencode, deveco and hermes each document
-// at their own call site why they drop the field, and a backend that ignored
-// it would leave the agent with no brief at all — which is why the default is
-// false and a new runtime has to opt in consciously.
-func providerSupportsInlineSystemPrompt(provider string) bool {
+// Membership is a property of this repository's backend code and, for Codex,
+// the installed protocol version. Each provider below consumes a non-empty
+// SystemPrompt through a verified instruction channel. Everything else is
+// absent deliberately: a backend that ignored it would leave the agent with no
+// brief at all, so the default is false and a new runtime has to opt in
+// consciously.
+const minCodexThreadInjectItemsVersion = "0.121.0"
+
+func providerSupportsInlineSystemPrompt(provider, codexVersion string) bool {
 	// Built-in runtime identities inherit their delivery from the protocol
 	// family they run on, the same resolution runtimeConfigPath does.
 	if desc, ok := agent.BuiltinRuntimeByID(provider); ok {
-		return providerSupportsInlineSystemPrompt(desc.ProtocolFamily)
+		return providerSupportsInlineSystemPrompt(desc.ProtocolFamily, codexVersion)
+	}
+	if provider == "codex" {
+		// thread/inject_items first appears in the v0.121.0 app-server
+		// protocol. Unknown and older versions keep file delivery: temporary
+		// repository pollution is preferable to silently dropping the brief or
+		// replacing the user's own developer_instructions.
+		return agent.CheckMinCLIVersionFor(codexVersion, minCodexThreadInjectItemsVersion) == nil
 	}
 	switch provider {
 	case "codebuddy", "grok", "kimi", "kiro", "openclaw", "qoder", "qoderclicn", "qwenpaw", "traecli":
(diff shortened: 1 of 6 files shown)


iiwish:
Hi @Bohan-J, a gentle follow-up on the stacked Codex fix in #7914 when you have a chance.

It currently targets `agent/j/a81be0c9` from #7123, so I wanted to check whether you would prefer to land or update #7123 first and then rebase/retarget #7914, or take another path. There is no urgency from my side; I mainly want to keep the dependency and next step clear.

Happy to rebase or adjust #7914 once you confirm the preferred direction. Thanks!

Issue #6150, opened 2026-07-29
[Bug]: Hermes issue rerun succeeds once, then resumed ACP session loses the named custom provider


shendu831105:
### Deployment type

Official App

## Summary

I have two Linux servers on Tencent Cloud. Each server runs a Multica remote runtime using Hermes, and both Hermes installations use a named custom provider backed by Volcengine Ark / Agent Plane:

```text
provider: custom:ark-code-latest
model: ark-code-latest
```

Hermes Chat works on both servers, and a fresh Multica task works, so the endpoint and credentials are valid.

All per-Agent model overrides in Multica have been cleared. The Agents inherit the runtime default. No model or runtime setting is changed between the successful and failed runs.

The repeatable pattern is:

```text
fresh Hermes ACP session -> task succeeds
next comment / approval / mention -> Multica resumes the session -> immediate failure
```

The exact error is:

```text
hermes provider error: No LLM provider configured. Run hermes model to select a provider, or run hermes setup for first-time configuration.
```

Running `multica issue rerun <issue-id>` creates a fresh task/session and succeeds once. The next invocation fails again when the Hermes ACP session is resumed. Retrying the failed task does not help.

I reproduced two independent cases:

1. **Same-Agent continuation:** a fresh rerun completes, then a simple `continue` comment causes the same Agent to fail immediately.
2. **Cross-Agent handoff:** a fresh ProjectManager run completes, then it mentions the Lead Agent; the Lead resumes its previously persisted Hermes session and immediately fails with the same error.

## Expected behavior

- A resumed Hermes session should re-resolve the current runtime provider and credentials.
- `issue rerun` should not leave the issue in a state where the next invocation deterministically fails.
- If a persisted ACP session cannot be restored, Multica should safely retry once with a fresh Hermes session using the current runtime configuration.

## Actual behavior

- A fresh session works.
- The first resumed invocation loses the usable provider configuration.
- The issue remains stuck until another fresh rerun, which again works for only one invocation.

## Investigation

This appears to be a Hermes ACP session-rehydration bug that Multica exposes through session resume.

In Hermes v2026.7.20:

- `acp_adapter/session.py::_persist()` builds metadata containing provider, base URL, and API mode, but the new-session path initially stores only `{"cwd": ...}` in `model_config`.
- After a successful request, token accounting persists `agent.provider` as `billing_provider`.
- For a named custom provider, the resolved `agent.provider` is the generic string `custom`; the API key is intentionally not persisted.
[...]

PR #5785, opened 2026-07-22
MUL-5154: fix(agent): surface kimi provider.api_error so the task fails visibly (#5760)

kakiuwang-ui:
## Summary

Reduced per @Bohan-J's guidance on the #6083 overlap: this PR is now **only the ACP-sniffer half** — the piece nothing on `main` replaces. Four files, all in `server/pkg/agent/`.

The real #5760 bug: the ACP adapter swallowed the provider error and reported `stopReason=end_turn`, so a kimi `400 … must not be empty` landed as **`status=completed` with empty output**. Everything merged in #6083 keys off `Result.Error`; if the adapter never surfaces the error there, none of it can see the failure. This PR makes the adapter surface it.

## What it does

- **`acpErrorHeaderRe` / `acpErrorDetailRe` / `acpTerminalErrorRe`** now recognise kimi's emoji-less `provider.api_error: NNN` lines. Terminal only for `400/401/403`; `429`/`408` are adapter-retried and must stay `completed`.
- **`messageLocked`** forwards the status prefix (`providerApiErrorStatusRe`) into the surfaced string, so `Result.Error` carries `provider.api_error: 400` — which `taskfailure.UnresumableHistory` on `main` then classifies. This is the connective tissue between the two halves.
- **`isPoisonedHistory()`** detects the `400` + `must not be empty` fingerprint across kimi's two-line stderr format and sets `Result.ResumeRejected` in both the hermes and kimi backends — the positive backend signal the daemon reads alongside the string-based path.

## Dropped as redundant / superseded (per your review)

- The three `classifyPoisonedError` branches → **#6083**'s `taskfailure.UnresumableHistory` covers the same strings (confirmed by your table).
- `GetLastTaskSession` / `GetLastChatTaskSession` `NOT EXISTS` + the daemon carry-forward → **migration 234** `retired_session_id`.
- The agent-rename gate (`gateResumeToSameAgentName` / `ReadPriorAgentName`, MUL-5736) → split out, already up as **#5909**.

## On the 429 question

You asked whether `main`'s code-agnostic predicate needs a transient-status guard. **I don't think it does.** A real provider returns a rate-limit *before* it validates the request body, so `429` + a content-emptiness complaint in one response isn't a combination Moonshot (or others) produce — the kimi adapter also retries `429`/`408` internally before any content check. My sniffer still gates on `400` locally (conservative), but I agree `main`'s reason-string path should stay code-agnostic; #6066's real string carries no code at all. No guard needed from my side.

## Verification

[...]

PR #7920, opened 2026-09-02
MUL-6969: fix(agent): anchor hermes ACP error-sniffer detail regex; terminal markers only on header lines

ouxxyy:
## Problem

The hermes ACP provider error sniffer misclassifies ordinary conversation echo as upstream errors:

1. `acpErrorDetailRe` matched an `Error:` / `Details:` substring **anywhere** in a line, so a tool-output or assistant line that merely contains the word (e.g. an echoed log entry) posed as a diagnostic line.
2. Terminal give-up markers flipped the sniffer's `terminal` flag even when they appeared on a **detail** line that was echoing them, not on an actual header (`⚠️`/`❌`/`[ERROR]`) line — aborting a call that had not actually given up.

## Fix

- Anchor `acpErrorDetailRe` to the line start, allowing an optional marker plus a bare identifier prefix (`RuntimeError:`, `JSONDecodeError:`, …) so real exception lines stay capturable (cf. GH #6150's `RuntimeError: No LLM provider configured`).
- Compute `isHeader` once; only header lines can flip the terminal flag.

## Testing

- `go test ./pkg/agent/ -run 'ACP|ErrorSniffer|Error'` passes; full `./pkg/agent/` suite passes locally.
- Added/updated regex cases cover identifier-prefixed detail lines and terminal-marker echo on detail lines.

By submitting this contribution I agree to condition 2 of the Multica License.

### DIFF
--- a/server/pkg/agent/hermes.go
+++ b/server/pkg/agent/hermes.go
@@ -2755,7 +2755,12 @@ var acpErrorHeaderRe = regexp.MustCompile(`(?:⚠️|❌|\[ERROR\]).*(?:BadReque
 // acpErrorDetailRe pulls the most useful single-line messages out of
 // the subsequent lines of the error block (the one whose "Error:" or
 // "Details:" tag actually spells out what happened).
-var acpErrorDetailRe = regexp.MustCompile(`(?:Error:|detail:|Details:)\s*(.+)`)
+// Anchored to the line start (an optional 📝 marker and a bare identifier
+// prefix like "RuntimeError:"/"JSONDecodeError:" are allowed) so an "Error:"
+// substring in the middle of an echoed conversation/tool line cannot pose as
+// a diagnostic; the identifier prefix keeps real exception lines capturable
+// (GH #6150's "RuntimeError: No LLM provider configured").
+var acpErrorDetailRe = regexp.MustCompile(`^(?:📝\s*)?(?:\w*Error:|detail:|Details:)\s*(.+)$`)
 
 // acpTerminalErrorRe matches markers that only appear when the
 // adapter has *given up* on the upstream call — either after
@@ -2863,10 +2868,13 @@ func (s *acpProviderErrorSniffer) Write(p []byte) (int, error) {
 				continue
 			}
 		}
-		if !(acpErrorHeaderRe.MatchString(line) || acpErrorDetailRe.MatchString(line)) {
+		isHeader := acpErrorHeaderRe.MatchString(line)
+		if !(isHeader || acpErrorDetailRe.MatchString(line)) {
 			continue
 		}
-		if acpTerminalErrorRe.MatchString(line) {
+		// Terminal markers belong to header lines; a detail line that merely
+		// echoes a terminal marker substring must not flip the flag.
+		if isHeader && acpTerminalErrorRe.MatchString(line) {
 			s.terminal = true
 		}
 		if s.seen[line] {
--- a/server/pkg/agent/hermes_test.go
+++ b/server/pkg/agent/hermes_test.go
@@ -1780,8 +1780,7 @@ func TestParseACPTokenUsageCostIsOptional(t *testing.T) {
 // TestParseACPModelIDFromMeta covers the shapes the `_meta` model id can
 // arrive in, and confirms a missing one degrades to empty rather than
 // inventing an attribution.
-func TestParseACPModelIDFromMeta(t *testing.T) {
-	t.Parallel()
+func TestParseACPModelIDFromMeta(t *testing.T) {	t.Parallel()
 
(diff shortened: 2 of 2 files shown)


Bohan-J (changes requested):
Thanks for picking this up — the direction is right. The sniffer decides run status from display text, so tightening what counts as a diagnostic (instead of adding another classifier) is the correct place to fix it, and the detail-regex anchoring plus the `\w*Error:` escape hatch for `RuntimeError:` / `JSONDecodeError:` is well judged.

The first half of the change works. The second half — "terminal markers only on header lines" — does not hold yet, and the new test does not cover it. Details below; all findings are reproduced against head `5f733e4e`.

## 1. `isHeader` is not a header test (blocking)

`acpErrorHeaderRe` is still unanchored, so it matches anywhere in a line. Naming its result `isHeader` in `Write` does not make the line a header — any line that *quotes* a failure banner satisfies header, detail and terminal at once, and still flips `s.terminal`.

Reproduced by feeding these lines to `newACPProviderErrorSniffer("hermes")` at head:

| input line | `terminalMessage()` |
| --- | --- |
| `2026-09-02 10:00:00 [ERROR] tools.vision_tools: Error analyzing image: HTTP 429 from vision API` | non-empty ❌ |
| `Error: copied "[ERROR] API call failed after 3 retries: HTTP 429"` | non-empty ❌ |
| `Details: users often paste ❌ API call failed after 3 retries: HTTP 429 into the chat` | non-empty ❌ |
| `❌ API call failed after 3 retries: HTTP 429` (control) | non-empty ✅ correct |

The first row is this PR's own motivating scenario: a vision-tool error record. It only passes in the test you added because the tool happened to log `Error code: 400` rather than `HTTP 400`; change that one token and a completed run is still marked as a terminal provider failure.

Worth noting the header decision is structural, not textual, so please derive it from structure rather than by anchoring the regex to the physical line start — that would break the legitimate timestamped form. `acpLogRecordPrefixRe` is already parsed a few lines above and gives you level, logger and payload; a header is either a bare banner at the start of the line, or the payload of a real diagnostic record. The logger name is also information worth using: an `[ERROR]` record from `tools.*` quoting an upstream status is not an adapter-level provider failure, which is precisely the class of noise this PR is trying to stop.

## 2. The new test does not exercise the terminal gate (blocking)

`TestHermesProviderErrorSnifferIgnoresToolTracebacks` only proves the detail anchoring. Matching each of its three lines against the regexes at head:

| line | header | detail (new) | detail (old) | terminal |
| --- | --- | --- | --- | --- |
| `[ERROR] mcp.client.stdio: Failed to parse JSONRPC message from server` | no | no | no | yes |
| `pydantic_core.ValidationError: 1 validation error for JSONRPCMessage` | no | no | yes | no |
| `[ERROR] tools.vision_tools: Error analyzing image: Error code: 400` | no | no | no | yes |

Only line 2 changes behaviour (old detail captured it, new one does not). No line reaches `isHeader && terminal`, so the `terminalMessage() == ""` assertion passes on the pre-change code as well — the second fix ships untested. Lines 1 and 3 were never captured before this PR either, so they assert nothing.

Please add a case that lands squarely on the gate — matches detail, contains a terminal marker, is not a header — plus the quoted-banner rows from §1 as negatives, and bare/timestamped real headers as positive controls so the true-positive path stays pinned.

## 3. Rebase onto `main` (blocking)

The branch is based on `9d09902d`; `2e424e06` ("MUL-5154: fix(agent): surface Kimi provider API errors", #5785) landed afterwards and touches the same struct and regex block. GitHub reports the PR as conflicting, and a local merge conflicts in `server/pkg/agent/hermes.go` (`hermes_test.go` merges cleanly).

When resolving, please keep both sides rather than taking this branch's version of the hunk: `kimiStyle`, `kimiProviderApiErrorRe` and `kimiTerminalErrorRe` carry the provider-scoped 400/401/403-terminal and 429/408-non-terminal semantics, and there is also a `Finalize()` / `isPoisonedHistory()` path in `Execute` from the same commit.

## 4. `gofmt` (blocking, trivial)

`server/pkg/agent/hermes_test.go:1783` collapses `{` and `t.Parallel()` onto one line in `TestParseACPModelIDFromMeta`, which is unrelated to this change and makes `gofmt -l` flag the file. Please drop it or run `gofmt -w`.

## Verification performed

- `gofmt -l server/pkg/agent/hermes.go server/pkg/agent/hermes_test.go` → `hermes_test.go`
- `git merge --no-commit --no-ff origin/main` → conflict in `server/pkg/agent/hermes.go`
- Ad-hoc sniffer tests at head producing the two tables above (scratch file, not committed)

Happy to re-review once the header test is structural and the gate has a regression test.

PR #7933, opened 2026-09-02
MUL-6983 WEN-94: reply to triggering Lark messages

wings27:
## Summary
- use native Lark replies for task results in ordinary p2p/group chats
- keep topic replies in-thread and restrict chat-level fallback to unsupported topic replies
- make immediate `/issue` and offline/archived notices quote the triggering message

## Tests
- `go test ./internal/integrations/lark -count=1`
- `go test ./internal/integrations/... -count=1`
- `go vet ./internal/integrations/lark`

Full `go test ./... -count=1` reaches and passes the Lark package but fails unrelated CLI tests because the Multica task runtime marker intentionally blocks normal config-token fallback.

### DIFF
--- a/server/internal/integrations/lark/outbound.go
+++ b/server/internal/integrations/lark/outbound.go
@@ -416,17 +416,16 @@ func outboundChatID(b ChatSessionBinding) ChatID {
 	return ChatID(b.ChannelChatID)
 }
 
-// threadReplyTarget derives the outbound reply target from the chat
-// binding's most-recent inbound trigger. We thread the reply ONLY when
-// that trigger was itself inside a Lark topic (last_lark_thread_id
-// present): normal group / p2p chats keep the unchanged chat-level send
-// path, and only an @-mention that happened inside a thread gets a
-// threaded reply (replying to last_lark_message_id with reply_in_thread).
-// The zero ReplyTarget means "send at the chat level".
+// threadReplyTarget derives the outbound reply target from the binding's
+// most-recent inbound trigger. Every trigger with a message id uses Lark's
+// native reply endpoint; topic messages additionally request an in-thread
+// reply. The zero target is reserved for bindings without a message id.
 func threadReplyTarget(binding ChatSessionBinding) ReplyTarget {
-	if binding.LastThreadID.Valid && binding.LastThreadID.String != "" &&
-		binding.LastMessageID.Valid && binding.LastMessageID.String != "" {
-		return ReplyTarget{MessageID: binding.LastMessageID.String, InThread: true}
+	if binding.LastMessageID.Valid && binding.LastMessageID.String != "" {
+		return ReplyTarget{
+			MessageID: binding.LastMessageID.String,
+			InThread:  binding.LastThreadID.Valid && binding.LastThreadID.String != "",
+		}
 	}
 	return ReplyTarget{}
 }
@@ -451,7 +450,7 @@ func sendWithThreadFallback(log *slog.Logger, op string, target ReplyTarget, sen
 	if err == nil {
 		return nil
 	}
-	if target.IsSet() && isThreadReplyUnsupported(err) {
+	if target.InThread && isThreadReplyUnsupported(err) {
 		log.Warn("lark: thread reply unsupported for target, retrying at chat level",
 			"op", op, "reply_message_id", target.MessageID, "error", err)
 		if fallbackErr := send(ReplyTarget{}); fallbackErr != nil {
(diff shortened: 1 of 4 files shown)


server/internal/integrations/lark/outbound.go
```diff
@@ -451,7 +450,7 @@ func sendWithThreadFallback(log *slog.Logger, op string, target ReplyTarget, sen
 	if err == nil {
 		return nil
 	}
-	if target.IsSet() && isThreadReplyUnsupported(err) {
+	if target.InThread && isThreadReplyUnsupported(err) {
```
Bohan-J, line 453: `threadReplyUnsupportedCodes` isn't thread-specific: `230011` (trigger recalled), `230050` (trigger invisible to the operator) and `230111` (self-destructing message) hit ordinary quote replies just as much as threaded ones.

Gating the fallback on `target.InThread` means a recalled trigger message now drops the task result instead of falling back to a chat-level send — a failure mode that did not exist before, because ordinary chats never routed through the reply endpoint at all.

Suggest keeping `target.IsSet()` here, or splitting the code set into quote-safe vs thread-only, while still not retrying on ambiguous errors (timeouts, 5xx, 230020, 230049).

server/internal/integrations/lark/outbound_test.go
```diff
@@ -683,8 +704,21 @@ func TestPatcherSendsToChatWhenNoThread(t *testing.T) {
 		t.Fatalf("expected one text send; got %d", len(api.textSent))
 	}
 	if api.textSent[0].ReplyTarget.IsSet() {
-		t.Errorf("non-thread trigger must NOT route through the reply endpoint; got %+v",
-			api.textSent[0].ReplyTarget)
+		t.Errorf("missing message id must use chat-level send; got %+v", api.textSent[0].ReplyTarget)
+	}
+}
+
+func TestQuoteReplyDoesNotUseThreadUnsupportedFallback(t *testing.T) {
```
Bohan-J, line 711: This pins the message-loss behaviour as if it were intended. Once the fallback is restored for quote replies, this test should instead assert both halves of the contract: a deterministic failure (e.g. 230011) falls back to a chat-level send, and an ambiguous failure (timeout / 5xx / 230020) does not.

Bohan-J (changes requested):
Product direction looks right: attaching results to the message that triggered them is a real improvement over free-floating chat sends, and it extends the existing `ReplyTarget` path instead of adding a parallel send mechanism. Two things need to change before this can merge.

## Must fix

### 1. Ordinary quote replies lost their safe fallback

`outbound.go:453` narrows the classified-error fallback from `target.IsSet()` to `target.InThread`. But `threadReplyUnsupportedCodes` is not thread-specific — `230011` (trigger message recalled), `230050` (trigger invisible to the operator) and `230111` (self-destructing message) apply to any reply, including the plain quote replies this PR newly introduces for p2p and group chats. The code set's own doc comment states these codes mean nothing was sent and that a chat-level send to the same chat is unaffected, which is exactly the condition that makes the fallback safe.

Concretely: a member runs `/issue` in a group and then recalls that message while the task is still running. The reply endpoint returns 230011, `sendWithThreadFallback` propagates the error, and the task result is dropped entirely. Before this PR the ordinary path never touched the reply endpoint, so this failure mode is new.

Please keep the deterministic-failure fallback for quote replies — restore `target.IsSet()`, or split explicit code sets for quote-reply vs thread-only failures — while still refusing to retry on ambiguous failures (transport timeouts, 5xx, 230020 rate limits, 230049 "message is being sent").

### 2. Rebase required — the PR does not merge

GitHub reports `CONFLICTING` / `DIRTY`; the three-way merge conflicts in `server/internal/integrations/lark/outcome_replier_test.go`. Please rebase onto current `main`, resolve the conflict, and keep the coverage `main` has added in the meantime. Worth re-checking after the rebase that the newer prompts sharing this path (`/new`, `/clear`, bare `/issue`, duplicate `/issue`) also quote ordinary messages correctly — that is a consistent extension of the same product semantics, it just isn't exercised by the tests here.

## Nit

`client.go:244-247` and `:282-283` still describe `ReplyTarget` as topic-only ("so the card lands inside the originating 话题", "threads the text reply back into a Lark topic"). Under the new semantics it is "a message id means quote reply; `InThread` decides whether it also stays in the topic". The `ReplyTarget` type doc at `:255` is already accurate — it is the two param structs that are now stale, and stale docs there are easy for the next caller to misread.

## Blast radius

If this lands wrong, every Lark p2p and group result can surface in the wrong place, or go missing when the target message is no longer replyable. It shows up in outbound error logs and user reports, and reverting the single commit restores the chat-level send — no data rollback involved.

wings27:
Closing at Wen's request: pursue a no-upstream-code-change solution using the existing Alex Hermes profile/gateway instead.

PR #6624, opened 2026-08-08
MUL-5917, YTO-17: detect WorkBuddy’s bundled CodeBuddy CLI

owen-kuai:
## What does this PR do?

Multica already has a CodeBuddy backend, but the daemon only discovers a `codebuddy` executable from its configured path, daemon `PATH`, or login-shell `PATH`. Tencent WorkBuddy installs its headless CodeBuddy CLI inside the desktop app instead of putting it on `PATH`, so WorkBuddy users currently see no CodeBuddy runtime.

The daemon now probes the standard macOS WorkBuddy app locations after the existing sources. When that fallback wins, it also prepends WorkBuddy's staged Node runtime (`~/.workbuddy/binaries/node/versions/*/bin`) so the bundled `#!/usr/bin/env node` entrypoint can launch from a GUI-started daemon. Explicit paths and normal `PATH` installations remain authoritative.

This matches Tencent's installed WorkBuddy 5.3.5 package (`@tencent-ai/codebuddy-code` 2.115.0) and its official [headless-mode documentation](https://copilot.tencent.com/docs/cli/headless), which documents `codebuddy`/`cbc`, stream JSON input/output, MCP, resume, model, and effort flags.

## Related Issue

Closes #6623

Closes YTO-17

Related to #5861 (the broader WorkBuddy-to-Multica integration request).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [x] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/config.go`: add ordered system/user WorkBuddy bundle candidates and discovery of WorkBuddy's staged Node runtime.
- `server/internal/daemon/agents_probe.go`: use the WorkBuddy bundle only after configured/PATH/login-shell resolution, and reject it when no runnable Node runtime is available.
- `server/internal/daemon/agents_probe_workbuddy_test.go`: cover bundle fallback, PATH precedence, explicit-path hard misses, missing Node, and candidate ordering with staged executables.
- `apps/docs/content/docs/install-agent-runtime*.mdx`: document WorkBuddy detection and Tencent's headless CLI reference in all supported locales.

## How to Test

1. Run `go test ./internal/daemon/...` from `server/`.
2. Run `go test ./pkg/agent/...` from `server/`.
[...]

PR #7934, opened 2026-09-02
MUL-5917: feat(agent): add WorkBuddy bundled CodeBuddy CLI as a built-in runtime

### LINKED ISSUE #6623: Detect the CodeBuddy CLI bundled with WorkBuddy on macOS
## Problem

Multica already supports Tencent's `codebuddy` protocol, but daemon discovery only checks `MULTICA_CODEBUDDY_PATH`, the daemon `PATH`, and the user's login-shell `PATH`. Tencent's WorkBuddy desktop app bundles an executable CodeBuddy CLI without installing `codebuddy` globally, so a Mac with WorkBuddy installed can still report no CodeBuddy runtime.

The current WorkBuddy 5.3.5 app ships the CLI at:

```text
/Applications/WorkBuddy.app/Contents/Resources/app.asar.unpacked/cli/bin/codebuddy
```

The bundled package identifies itself as `@tencent-ai/codebuddy-code` 2.115.0. Its `--help` exposes the same headless `-p`, stream-JSON input/output, permission, MCP, resume, model, and effort flags that Multica's existing `codebuddy` backend uses. Tencent's official headless documentation also identifies `codebuddy`/`cbc` as the programmatic interface: https://copilot.tencent.com/docs/cli/headless

## Proposed change

When the default `codebuddy` command is not visible on either daemon or login-shell `PATH`, probe WorkBuddy's bundled CLI in the standard macOS application locations:

- `/Applications/WorkBuddy.app/.../cli/bin/codebuddy`
- `~/Applications/WorkBuddy.app/.../cli/bin/codebuddy`

Keep the existing precedence and safety contract:

1. `MULTICA_CODEBUDDY_PATH`
2. daemon `PATH`
3. login-shell `PATH`
4. WorkBuddy app bundle fallback

An explicitly configured but missing `MULTICA_CODEBUDDY_PATH` must remain a hard miss rather than silently selecting the app bundle.

## Acceptance criteria

- A macOS daemon discovers the executable bundled by WorkBuddy when no global `codebuddy` command is available.
- System `/Applications` is preferred over user `~/Applications`.
- A valid explicit/PATH installation still wins over the bundle fallback.
- A missing explicit `MULTICA_CODEBUDDY_PATH` does not fall back.
- Unit tests use staged executables and do not depend on WorkBuddy being installed.
- Provider documentation explains that Multica can use WorkBuddy's bundled CLI on macOS.

Related to #5861, but scoped to Multica running WorkBuddy's bundled agent runtime rather than WorkBuddy calling Multica.

BuXiuDN:
## What does this PR do?

Tencent's WorkBuddy desktop app bundles a private CodeBuddy CLI (the same `@genie/agent-cli` that CodeBuddy Code ships) but never installs it onto `PATH`. Users who install WorkBuddy — and nothing else — currently get no codebuddy-family runtime, even though a fully capable CLI is sitting inside the app bundle.

This PR makes the daemon discover that bundled CLI and register it as a first-class **`workbuddy`** built-in runtime identity, hosted by the existing `codebuddy` protocol family. It is the "Multica drives WorkBuddy's bundled runtime" half of the WorkBuddy integration discussion in #5861. It supersedes #6624, the stalled macOS-only predecessor (whose tracking issue was #6623), by covering both platforms.

Discovery order: `MULTICA_WORKBUDDY_PATH` (explicit override, hard-miss when missing) → desktop-app bundle locations → not registered.

Key design points:

- **Bundle locations**: Windows — `%ProgramFiles%\WorkBuddy` and `%LOCALAPPDATA%\Programs\WorkBuddy`, plus every existing drive's `\Program Files\WorkBuddy`; macOS — `/Applications/WorkBuddy.app` and `~/Applications/WorkBuddy.app`.
- **Launch shape (every platform)**: the bundled CLI is a `#!/usr/bin/env node` shebang script WorkBuddy never puts on `PATH`. Windows cannot spawn a bare extensionless script, and on unix an `env node` shebang would resolve `node` from the daemon's process PATH — which a GUI-launched daemon may not have even though WorkBuddy staged a runtime. The entry therefore runs the CLI under a pinned Node interpreter on every OS: `Path = <node>`, `LaunchPrefix = [<cli script>]`, spawning `<node> <script> -p …`. `AgentEntry` gains a `LaunchPrefix` field carried per entry — the daemon process environment is never mutated.
- **Node selection**: WorkBuddy stages Node under `~/.workbuddy/binaries/node/versions/<semver>` (Windows: `node.exe` directly in the version dir; macOS: `bin/node`). Both layouts are probed; the highest semantic version wins (22.10.0 > 22.9.0 — a lexicographic reverse sort would get this wrong), with `node` on `PATH` as fallback.
- **Protocol plumbing**: version detection, registration, and task launch all run the entry exactly as it launches (`node <script> --version` reports the bundled CLI's version, not Node's). `codebuddyBackend` implements `backendOverrideApplicator` so the identity can override its fallback executable and error label; heal re-resolution is intentionally not attempted for bundle paths (the app owns them).

This supersedes the stalled macOS-only #6624 and addresses its requested changes: no daemon-wide `PATH` mutation (the interpreter travels per entry as `Path`/`LaunchPrefix`), deterministic semantic-version Node selection, and a real Node-shebang subprocess test proving version detection runs `<node> <cli> --version`.

## Related Issue

Closes #6623
Supersedes #6624
Relates to #5861

## Type of Change

- [x] Documentation update

## Changes Made

- `server/pkg/agent/builtin_runtimes.go` — add `workbuddy` built-in runtime descriptor (protocol family `codebuddy`); `codebuddyBackend` implements `backendOverrideApplicator`
- `server/pkg/agent/codebuddy.go` — runtime-identity overrides (fallback executable, error label)
- `server/internal/daemon/workbuddy_probe.go` — WorkBuddy discovery: bundle locations, staged-node resolution (semver), launch-shape assembly that pins the Node interpreter on every platform (the bundled CLI is never spawned directly), and the Windows Program-Files roots enumeration (kept in the platform-neutral file so the package compiles on every OS)
- `server/internal/daemon/agents_probe.go` — route `workbuddy` through the new probe
- `server/internal/daemon/types.go` — `AgentEntry.LaunchPrefix`
- `server/internal/daemon/daemon.go` — carry `LaunchPrefix` into version probes and task launches
- `server/internal/daemon/workbuddy_probe_test.go`, `server/pkg/agent/workbuddy_test.go` — unit tests: the pinned-node launch shape on every OS, a real Node-shebang subprocess test proving version detection runs `<node> <cli> --version` (CLI version, not Node's), no process-PATH mutation, discovery priority, semver ordering incl. the 22.10.0 vs 22.9.0 trap, hard-miss on missing explicit path, both staged-node layouts, backend overrides
- `packages/views/runtimes/components/provider-logo.tsx` (+ `workbuddy-logo.png`) — provider logo
- `packages/core/runtimes/display.ts` — display-name mapping
- `apps/web/features/landing/i18n/{en,zh,ja,ko}.ts` — supported-tools list (26 → 27) and counts
- `apps/docs/content/docs/install-agent-runtime.{mdx,zh,ja,ko}.mdx` — detection docs for the bundled CLI

## How to Test

1. Install Tencent WorkBuddy desktop (Windows or macOS) and sign in once. Do not install `codebuddy` separately and do not add anything to `PATH`.
2. Start the Multica daemon. Within one discovery cycle (~2 min) a `workbuddy` runtime appears — version equals the bundled CLI's version (e.g. 2.137.1), name "WorkBuddy".
3. Create an agent on that runtime (model empty → provider default) and assign an issue; the agent picks it up and replies via the bundled CLI.
4. Optional overrides: `MULTICA_WORKBUDDY_PATH` pointing at your own install wins over the bundle; a stale/missing explicit path hard-misses (no silent fallback to the bundle).

Unit tests: `go test ./server/internal/daemon/ ./server/pkg/agent/ -run Workbuddy -count=1`. The full backend suite runs in CI (`scripts/test-go.sh --race`).

## Checklist

## AI Disclosure

**AI tool used:** An autonomous AI coding agent

**Prompt / approach:**
This change was produced by an autonomous AI coding agent and human-reviewed before submission. The agent researched the stalled upstream macOS-only proposal (#6624) and its review feedback, verified the bundled CLI's protocol compatibility against a real WorkBuddy install, and implemented the discovery/launch plumbing with unit tests.


multica-eve (changes requested):
People who only installed WorkBuddy should be able to use it directly from Multica; that is a real and worthwhile need. However, this proposal turns an automatic backup choice into a separate product entry, and the experience can still fail on both supported operating systems and after the desktop app updates. **Product verdict: needs clarification before acceptance.**

Product assessment:

- Is the problem real? **Yes.** #6623 documents a WorkBuddy-only installation that Multica cannot currently discover.
- Is a durable mechanism worthwhile? **Yes.** Discovery and isolated use of WorkBuddy's own runtime form a complete user journey.
- Would we solve it this way? **Not without an explicit product decision.** #6623 asks for the bundled CLI as the final fallback of the existing CodeBuddy runtime. This PR instead creates a 27th, independently visible WorkBuddy runtime and shows both entries when standalone CodeBuddy is also installed. If WorkBuddy needs a distinct account/config/lifecycle surface, please record that decision; otherwise extend the existing CodeBuddy discovery path.

The process-wide `PATH` mutation from #6624 is gone, and carrying the interpreter through the existing launch-prefix mechanism is the right isolation direction. The following items still block merge on head `554cfb28a`:

1. **[P1] The production macOS bundle path is incorrect, so the issue's primary scenario is never discovered.**

   `server/internal/daemon/workbuddy_probe.go:50-52` appends `resources/app.asar.unpacked/...` directly below `WorkBuddy.app`. The installed path documented in #6623 is `WorkBuddy.app/Contents/Resources/app.asar.unpacked/cli/bin/codebuddy`. `server/internal/daemon/workbuddy_probe_test.go:41` stages the same incorrect shape, so the test passes while production misses. Please define the real relative path per platform and stage the actual macOS and Windows layouts in separate cases.

2. **[P1] Semantic-version selection does not work for the Windows flat Node layout.**

   `workbuddyStagedNodeGlobs` supports `versions/<version>/node.exe`, but `semverFromStagedNodePath` at `server/internal/daemon/workbuddy_probe.go:307-313` always walks up two directories and therefore only parses `versions/<version>/bin/node`. Every flat Windows candidate becomes unparseable and falls back to lexical order, allowing `22.9.0` to beat `22.10.0`. The current flat-layout test has only one version and cannot catch this. Extract the segment immediately after `versions` for both shapes and add a multi-version flat-layout regression test.

3. **[P1] A WorkBuddy update can leave the runtime permanently pinned to a deleted Node binary until the Multica daemon restarts.**

   `workbuddyEntryWithRuntime` deliberately leaves `Command` empty (`server/internal/daemon/workbuddy_probe.go:206-224`), disabling executable self-heal. The periodic availability refresh preserves the old entry for every known provider (`server/internal/daemon/agents_refresh.go:158-195`), so discovering the newly staged Node does not replace the stale `{Path, LaunchPrefix}` pair. If WorkBuddy removes its previous staged version during an update, version probes and task launches continue failing even though a new usable pair exists. Re-probe and atomically validate/adopt the Node + CLI pair, with a test that removes the old version, stages a new one, and recovers without daemon restart.

4. **[P2] The new identity does not inherit the CodeBuddy capabilities it claims to reuse.**

   The WorkBuddy descriptor has no model-discovery strategy, so `ListModels("workbuddy", ...)` returns an empty catalog. `handleModelList` also drops `entry.LaunchPrefix`, which would execute `node --acp` instead of `node <cli> --acp` if discovery were enabled. `ThinkingControlSupported` and `IsKnownThinkingValue` do not normalize built-in identities to their `ProtocolFamily`, so WorkBuddy cannot use CodeBuddy's supported effort values. If the separate identity remains, inherit these capabilities through the descriptor/common family-normalization layer and cover version detection, model discovery, and a complete task launch with the real subprocess shape.

Validation on `554cfb28a`:

- Targeted WorkBuddy tests — passed
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — passed
- `go test ./internal/daemon/... -count=1` with the run-owned `MULTICA_TASK_CONFIG_ROOT` removed — passed
- `go test ./pkg/agent/... -count=1` — passed
- `git diff --check` — passed
- CI — 16/16 passed

The green checks do not exercise the real macOS path, multiple flat Windows versions, or upgrade recovery.

If this change is wrong, users will detect it as a missing WorkBuddy runtime, the wrong Node version, or task-launch failures after an app update; rollback is to revert the change and restart the daemon, but agents already bound to `workbuddy` must be moved back to `codebuddy`; all users of the new WorkBuddy identity are affected.

Please address the four items above and request another review.

BuXiuDN:
Recording the product decision the review asked for: WorkBuddy stays a separate `workbuddy` runtime identity rather than a fallback path inside the existing `codebuddy` runtime. Rationale below comes from local testing of both products.

Facts observed on a Windows host running WorkBuddy and CodeBuddy CN side by side:

1. **Same underlying CLI, different products.** WorkBuddy bundles a CodeBuddy CLI (`@genie/agent-cli`) under its app resources; on this host the `codebuddy` on PATH is actually an npm shim pointing at that same bundled binary. So the two entries are not two different CLIs — WorkBuddy and CodeBuddy CN are separate desktop products that share the codebuddy-family CLI.

2. **Account separation is realized at the config-dir / spawn-env layer, not in the binary.** WorkBuddy sets `CODEBUDDY_CONFIG_DIR` to a WorkBuddy-owned home (`~/.workbuddy`) before launching its embedded CLI (verified in the app's `configureElectronApp`); a bare launch of the same CLI defaults to `~/.codebuddy`. WorkBuddy's own login is host-injected to the CLI at spawn (`CODEBUDDY_AUTH_TOKEN` / an ACC product-config path), not persisted in the CLI's `settings.json`.

Consequences for the design:

- A single `codebuddy` identity cannot host two accounts: the account is chosen by the env/config-dir the CLI is launched with. The #6623 "final fallback inside codebuddy" would therefore still not reach the WorkBuddy account — it would only discover the same binary under the codebuddy default config. That is the concrete argument for keeping a distinct identity.

- It also means the separate identity is only meaningful if it carries the isolation: the `workbuddy` runtime must launch the bundled CLI against the WorkBuddy config home (`CODEBUDDY_CONFIG_DIR=~/.workbuddy`) and provide an auth path (e.g. a CLI-native login under that isolated config dir). As currently written the PR launches the bundled CLI with neither, so the `workbuddy` entry would read `~/.codebuddy` and be unauthenticated — the opposite of the independent account this identity claims. I will fold this env/config-dir handling into the runtime entry (alongside item 4's capability inheritance) and cover it with a launch test.

When both WorkBuddy and CodeBuddy are installed the two entries are therefore intentional: they are two separately-authenticated runtimes over the same CLI family, matching the existing omp/pi precedent of a distinct identity per product/account within one protocol family.

BuXiuDN:
Re-review requested. New head is `a5471e583` (your review was on `554cfb28a`).

**Product decision, recorded Sep 21, 2026: WorkBuddy and CodeBuddy intentionally share the embedded CLI.** WorkBuddy gets no configuration directory, authentication path, account isolation, or duplicated embedded-CLI setting of its own — it reuses the CodeBuddy protocol family end to end. The `workbuddy` identity stays separate only so a WorkBuddy-only install is discoverable and its bundled CLI is pinned and validated as its own `{node, cli}` pair; it is not a second product surface with its own lifecycle. That decision is now written down under "Scope and non-goals" in `docs/plans/2026-09-21-workbuddy-runtime-review-fixes.md`.

**1. [P1] macOS bundle path** — fixed in `87f9f6164`. `workbuddyBundleRelativeCLIForPlatform(goos)` returns `Contents/Resources/app.asar.unpacked/cli/bin/codebuddy` on darwin and `resources/app.asar.unpacked/cli/bin/codebuddy` on windows, and the tests stage the two real layouts as separate cases. Tests: `TestWorkBuddyBundleRelativeCLIForPlatform`, `TestProbeAgentCLIs_DiscoversWorkBuddyBundle`.

**2. [P1] Windows flat Node semver** — fixed in `87f9f6164`. `semverFromStagedNodePath` now takes the segment immediately after `versions`, so both `versions/<v>/node.exe` and `versions/<v>/bin/node` parse. Multi-version flat-layout regression added (22.9.0 vs 22.10.0): `TestResolveWorkBuddyNode_WindowsFlatLayout_PicksNewestSemver`, plus `TestResolveWorkBuddyNode_WindowsFlatLayout`, `TestWorkBuddyNodeSemverOrdering`, `TestSemverFromStagedNodePath`.

**3. [P1] runtime pinned to a deleted Node after a WorkBuddy update** — fixed in `52e53a9ef`, `4e7d16d19`, `07fcee249`. `workBuddyEntryPairMissing` plus `reprobeWorkBuddyEntry` re-probe and atomically validate/adopt a fresh `{Path, LaunchPrefix}` pair; `resolveAgentEntryWithHeal` and `probeBuiltinRuntime` call it when the pinned pair fails, so recovery no longer waits for a daemon restart. Tests remove the old staged version, stage a new one, and assert recovery without a restart: `TestResolveAgentEntry_WorkBuddyReprobesReplacedStagedNode`, `TestProbeBuiltinRuntime_WorkBuddyReprobesWhenCurrentPathProbeFails`, `TestResolveAgentEntry_WorkBuddyVersionProbeFailureDoesNotAdopt` (never adopts an unvalidated pair), `TestProbeWorkBuddyAgent_MissingBundleCLIHardMisses`, `TestProbeWorkBuddyAgent_ExplicitMissingOverrideDoesNotUseBundleOrPATH`.

**4. [P2] new identity did not inherit the CodeBuddy capabilities it claims to reuse** — fixed in `7935839ac` and `3283826d5`.

- Model discovery: the descriptor now carries `CatalogDiscovery` (`discoverCodebuddyModels`), so `ListModels("workbuddy", ...)` returns CodeBuddy's catalog including `Fallback` / unavailable-model metadata instead of an empty one. Tests: `TestWorkbuddyModelDiscoveryPreservesFallbackCatalog`, `TestListModelsWorkBuddyPreservesCatalogAndLaunchPrefix`, `TestListModelsWorkBuddyFallbackRetries`.
- `LaunchPrefix` in discovery: `handleModelList` now passes the resolved entry's filtered `LaunchPrefix`, so discovery runs `node <cli script> --acp` rather than `node --acp`. Tests: `TestHandleModelList_BuiltinEntryCarriesLaunchPrefix`, `TestWorkBuddyDiscoveryPreservesNodeScriptLaunchPrefix`.
- Thinking: `thinkingCapabilityProvider` normalizes a built-in identity to its `ProtocolFamily` before the capability lookup in `usesDynamicThinkingCatalog`, `UsesACPCatalogThinking`, `ThinkingControlSupported`, and `IsKnownThinkingValue`; unknown and custom runtime IDs are deliberately left unchanged. Tests: `TestWorkBuddyThinkingCapabilityUsesCodeBuddyFamily`, `TestThinkingCapabilityNormalizationLeavesUnknownAndCustomProvidersUnchanged`, `TestIsKnownThinkingValue`, `TestThinkingControlSupported`.
- Complete task launch in the real subprocess shape: new `server/pkg/agent/workbuddy_launch_test.go` re-execs the test binary as a fake Node interpreter, runs an actual task through `Execute`, and reads back the argv the spawned process really received — asserting `argv[0] == <cli script>`, `argv[1] == "-p"`, and the presence of `--output-format stream-json`, `--input-format`, and `--permission-mode bypassPermissions`, then asserting the parsed assistant message, `SessionID`, and `Status=completed` (`TestWorkBuddyTaskLaunchKeepsBundledCliPrefix`). I checked that it actually catches the bug it guards: setting `LaunchPrefix` to nil makes it fail with `launched argv = ["-p", "--output-format", ...]`.

**Validation on `a5471e583`:**

- `go test ./pkg/agent -run 'Work[Bb]uddy|Codebuddy|CodeBuddy|TestIsKnownThinkingValue|TestThinkingControlSupported|TestLaunchPrefix|TestBuiltinRuntimeIdentitiesFilterLaunchPrefix|TestFilterLaunchPrefix|TestDiscoveryCacheKeySeparatesLaunchPrefixes' -count=1` — passed
- `go test ./internal/daemon -run 'Work[Bb]uddy|ModelList|LaunchPrefix|AgentEntry|Probe' -count=1` — passed
- `go vet ./pkg/agent/... ./internal/daemon/...` — passed, and `GOOS=darwin GOARCH=arm64 go vet ./pkg/agent/... ./internal/daemon/...` — passed, so the new cross-platform launch fixture also compiles for the real macOS path
- `gofmt -l` and `git diff --check` — clean

One caveat, stated plainly rather than left implicit: this box is Windows and the full `./internal/daemon` and `./pkg/agent` suites do not run clean here because of sandbox limitations (`symlink ... A required privilege is not held by the client`, and `resolve symlinks ... Access is denied` on a temp directory), which fail pre-existing tests unrelated to this change. I confirmed the failing set is identical when running the same tests against `9a4e42f60`, the previous head — including `TestAcquireLocalDirectoryLockSkipsWorktreeMode`, `TestChatTaskOnWorktreeResourceKeepsAssignment`, `TestChatTaskSkipsPathMutexButKeepsAssignment`, and `TestShouldReusePriorWorkdirSquadLeaderRejectsSymlinkEscape`, each failing 3/3 in both trees. So CI stays the authority for the macOS path and the POSIX shapes.

One thing I have not changed, and do not want to slip past you: WorkBuddy reuses the CodeBuddy backend through `ProtocolFamily`, so it inherits the existing `--permission-mode bypassPermissions` headless policy. This PR adds no new bypass surface, but the inheritance is implicit. Say the word if you would rather that become an explicit identity-level policy, or if you would prefer the separate-identity question split out into its own issue.

@multica-eve could you take another look?

PR #7935, opened 2026-09-02
MUL-6974: warn before overwriting a write-only MCP server entry on edit (#7923)

### LINKED ISSUE #7923: [Bug]: Original configuration is lost when editing an existing MCP server (Settings → MCP Configuration)
### Deployment type

self-host

### Area

Web app

### What happened?

在「设置 → MCP 配置」中，添加一个 MCP Server（例如名称 `open-design`）并填写完整的启动命令、参数、环境变量后保存，正常可用。随后点击「编辑」重新打开该 MCP 的编辑表单时，原有的配置信息（启动命令 / 参数 / 环境变量）没有正确回填，字段变为空，导致再次保存会覆盖并丢失原始配置。

期望行为：编辑已有 MCP 时，表单应完整回填原有的 名称 / 启动命令 / 参数 / 环境变量 等信息。

附上编辑时的截图：编辑名为 `open-design` 的 MCP 时，启动命令、参数、环境变量均显示为空（仅残留 `/ 占位提示`）。

### Steps to reproduce

1. 打开 设置 → MCP 配置。
2. 添加一个新的 MCP Server，填写：名称（如 `open-design`）、连接方式（STDIO / Streamable HTTP）、启动命令、参数、环境变量，然后保存。
3. 点击该 MCP 的「编辑」，重新打开编辑表单。
4. 观察：启动命令、参数、环境变量等字段是否仍保留原值。

### Screenshots (optional)

Image

Frankli9986:
Fixes #7923

## Problem

Editing an existing server in **Settings → MCP Configuration** opens an empty form: the workspace MCP library is write-only, so the API never returns the stored configuration (name/transport only). Saving then **silently replaces the stored configuration** — the flow looks like a harmless "open the form, fix the name, hit Save", and the original command, args, and env are lost.

## Fix

Keep the write-only design; surface the replacement where the decision is made:

- The edit dialog shows an inline warning (`role="alert"`, warning-tone styling matching the existing `delete-runtime-dialog` notice) when the caller could not hand the dialog a readable config: the saved configuration can't be read back, the form starts empty, and saving replaces the existing configuration for this server.
- In that state the save button reads **"Replace Configuration"** instead of "Save Changes", so the action is named for what it does.
- The judgment is `server && Object.keys(server.config).length === 0` — entries whose config the caller *can* hand over (the agent's own MCP tab) keep the quiet "Save Changes" path, and **Add** never shows the warning.
- New keys under `tab_body.mcp_config` in `en` / `ja` / `ko` / `zh-Hans`.

## Testing

- `packages/views` targeted tests (settings `mcp-tab`, agents `mcp-config-tab`, both i18n parity suites): the existing edit test now asserts the warning appears and submits via "Replace Configuration"; new tests cover "add shows no warning" and "readable config keeps Save Changes".
- Full `packages/views` suite: **413 files / 4939 tests passing** on top of current `main`.
- `tsc --noEmit` and eslint clean on the touched files.

### DIFF
--- a/packages/views/agents/components/tabs/mcp-config-tab.test.tsx
+++ b/packages/views/agents/components/tabs/mcp-config-tab.test.tsx
@@ -274,6 +274,27 @@ describe("McpConfigTab", () => {
     });
   });
 
+  // GH #7923: the write-only replace warning is for entries whose saved
+  // config the caller could NOT read back. Here the real config is handed to
+  // the dialog, so no warning appears and the save button keeps its label.
+  it("edits a readable entry without the write-only replace warning", async () => {
+    const user = userEvent.setup();
+    renderTab({
+      mcp_config: {
+        version: 1,
+        mcpServers: { fetch: { command: "uvx" } },
+      },
+    });
+
+    await user.click(
+      screen.getByRole("button", { name: /edit mcp server fetch/i }),
+    );
+    expect(screen.queryByTestId("mcp-write-only-warning")).toBeNull();
+    expect(
+      screen.getByRole("button", { name: /save changes/i }),
+    ).toBeInTheDocument();
+  });
+
   it("deletes the last managed server only after confirmation", async () => {
     const user = userEvent.setup();
     const { onSave } = renderTab({
--- a/packages/views/agents/components/tabs/mcp-server-dialog.tsx
+++ b/packages/views/agents/components/tabs/mcp-server-dialog.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useEffect, useMemo, useState } from "react";
-import { Loader2, Plus, Trash2 } from "lucide-react";
+import { Loader2, Plus, Trash2, TriangleAlert } from "lucide-react";
 import { Button } from "@multica/ui/components/ui/button";
 import {
   Dialog,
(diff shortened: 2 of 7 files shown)


Bohan-J:
Thank you for this, and sorry it sat open while a broader change landed on top of it.

Your diagnosis was exactly right, including the part that mattered most: the flow reads as "open the form, fix the name, hit Save", and the original command, args and env are silently replaced. Naming the action for what it actually does was the correct instinct.

#8021 has now merged (`d5cd0f300`) and covers the same ground with a wider fix, so I'm closing this as superseded rather than because anything here was wrong:

- The dialog is now titled **Replace configuration for {name}** and states that saved values aren't shown again for security, that a complete new configuration is required, and that agent assignments are preserved. The submit button reads **Replace configuration** — the same rename you proposed.
- It also removes the need to take that path at all for a rename: renaming is a separate inline action that submits only the name, so the destructive overwrite you identified can no longer happen as a side effect of relabelling a server.

One thing your version had that the merged change does not is the warning-toned `role="alert"` callout. That was a deliberate call rather than an oversight — with replacement split into its own explicitly labelled action, with its own dialog title and button, the salience is carried by the structure instead of by a coloured box. If replacement still turns out to be too easy to walk into, your inline warning is the obvious next thing to add, and I'd take it.

Closing as superseded by #8021. The report you fixed against, #7923, is closed too. Thanks for taking the time on this — it was a real bug and you found it.

PR #7942, opened 2026-09-02
WEN-92: handle Feishu document comment mentions

wings27:
## Summary
- route `drive.notice.comment_add_v1` through the existing Feishu installation and Multica agent task pipeline
- fetch anchored document comment context with the installed app credentials
- persist the original document comment target and deliver/read back the final reply through Drive APIs, never IM
- filter non-mentions, self-authored events, and unsupported notice types while reusing existing dedup and identity checks

## Verification
- `go test ./internal/integrations/lark -count=1`
- `go vet ./internal/integrations/lark`
- `git diff --check`

## Deployment dependency
The same Feishu app must subscribe to `drive.notice.comment_add_v1`, hold the required Drive/comment scopes, publish a new version, and have access to the target document. No second bot or Hermes gateway is required.


wings27:
Closing at Wen's request. We will not modify Multica official code; migrate the existing Alex Bot connection from Multica to the wen-alex Hermes Gateway instead.

PR #1123, opened 2026-04-16
MUL-4914: fix(daemon): expand isBlockedEnvKey with code-injection vectors

shaun0927:
## What does this PR do?

`isBlockedEnvKey` only rejected 6 specific keys plus the `MULTICA_` prefix. Well-known variables that lead to code execution (`LD_PRELOAD`, `NODE_OPTIONS`, `PYTHONPATH`) or traffic interception (`HTTP_PROXY`, `SSL_CERT_FILE`) on agent CLI child processes were not blocked, meaning any agent owner could achieve RCE on the daemon host via `custom_env`.

This expands the blocklist to cover dynamic-loader injection, interpreter-level injection, proxy/TLS hijacking, git-driven command execution, and editor-as-shell vectors. Legitimate keys like `ANTHROPIC_API_KEY`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_USE_BEDROCK` are unaffected.

## Related Issue

Closes #1114

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **`server/internal/daemon/daemon.go`**: Replace the `switch` statement with a `map[string]struct{}` covering ~25 dangerous keys across 5 categories. Move to package-level var for easier future additions.

## How to Test

1. Set `custom_env` on an agent to `{"LD_PRELOAD": "/evil.so"}` → verify it is blocked and a warning is logged.
2. Set `custom_env` to `{"NODE_OPTIONS": "--require /evil.js"}` → blocked.
3. Set `custom_env` to `{"HTTP_PROXY": "http://attacker:8080"}` → blocked.
4. Set `custom_env` to `{"ANTHROPIC_API_KEY": "sk-..."}` → passes through (legitimate use case).
5. Verify existing agents with legitimate `custom_env` are unaffected.

## Checklist

- [x] I have run tests locally and they pass
- [x] I have considered and documented any risks above

## AI Disclosure

**AI tool used:** Claude Code
**Prompt / approach:** Audited isBlockedEnvKey against known code-injection environment variables for Node.js, Python, Go, and OS-level dynamic linkers.

PR #7944, opened 2026-09-02
MUL-6993: fix(daemon): block code-injection and trust-redirection env vars in isBlockedEnvKey

### LINKED ISSUE #1114: [Bug]: Security: isBlockedEnvKey allows LD_PRELOAD/NODE_OPTIONS/HTTP_PROXY override via custom_env
### Deployment type

Self-hosted

### What happened?

`isBlockedEnvKey` in `server/internal/daemon/daemon.go` (v0.2.0, lines 1380–1388) only rejects `HOME`, `PATH`, `USER`, `SHELL`, `TERM`, `CODEX_HOME`, and anything prefixed with `MULTICA_`. Several well-known variables that lead to code execution or TLS/proxy hijacking in the agent CLIs multica spawns are **not** on the blocklist and therefore pass through `custom_env` into the child process environment.

```go
func isBlockedEnvKey(key string) bool {
    upper := strings.ToUpper(key)
    if strings.HasPrefix(upper, "MULTICA_") {
        return true
    }
    switch upper {
    case "HOME", "PATH", "USER", "SHELL", "TERM", "CODEX_HOME":
        return true
    }
    return false
}
```

### Impact

`custom_env` is set by agent owners (any workspace member, post-#1013) and is injected directly into the environment of `claude`, `codex`, `opencode`, `openclaw`, `gemini`, and `hermes` child processes running on the **daemon host** — which is generally a workstation belonging to a workspace owner. An agent owner can today set any of the following and obtain code execution or traffic interception on the daemon host:

| Variable | Effect |
|---|---|
| `LD_PRELOAD` / `LD_LIBRARY_PATH` | Arbitrary shared library injection (Linux) |
| `DYLD_INSERT_LIBRARIES` / `DYLD_LIBRARY_PATH` | Arbitrary dylib injection (macOS, unsigned/entitled binaries) |
| `NODE_OPTIONS` | `--require /path/to/evil.js` runs arbitrary JS in Claude Code / OpenCode |
| `PYTHONPATH` / `PYTHONSTARTUP` | Arbitrary Python on any Python-backed subtool |
| `GIT_SSH_COMMAND` | Arbitrary command during repocache `git` operations |
| `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` | Silently route all agent HTTP(S) traffic through attacker-controlled proxy → token/API-key interception |
| `SSL_CERT_FILE` / `SSL_CERT_DIR` / `NODE_TLS_REJECT_UNAUTHORIZED` | Disable or redirect TLS validation |
| `EDITOR` / `VISUAL` | Arbitrary command triggered by any subcommand that opens an editor (e.g. git commit flows) |
| `GIT_TERMINAL_PROMPT` | Re-enables the prompt that repocache (#1029) explicitly disables |

Because the daemon runs with the ambient permissions of the user who launched it, this is full user-level RCE on the daemon host, scoped to any workspace where the attacker can create an agent.

### Steps to reproduce

1. Create an agent (any runtime) in a workspace you're a member of.
2. Open the Environment tab and set `NODE_OPTIONS` to `--require /tmp/evil.js`.
3. Write `/tmp/evil.js` on the daemon host (or trick a workspace owner to do it — e.g. through any normal agent-authored file write) with `require('child_process').execSync('id > /tmp/pwned')`.
4. Assign the agent a task. Observe `/tmp/pwned` created on the daemon host.

### Suggested fix

Extend the blocklist to cover the categories above. An allowlist-first redesign is even better long term, but at minimum:

```go
var blockedEnvKeys = map[string]struct{}{
    // Daemon-managed core
    "HOME": {}, "PATH": {}, "USER": {}, "SHELL": {}, "TERM": {}, "CODEX_HOME": {},

    // Code-injection vectors
    "LD_PRELOAD": {}, "LD_LIBRARY_PATH": {},
    "DYLD_INSERT_LIBRARIES": {}, "DYLD_LIBRARY_PATH": {},
    "NODE_OPTIONS": {}, "NODE_PATH": {},
    "PYTHONPATH": {}, "PYTHONSTARTUP": {},

    // Proxy / TLS hijacking
    "HTTP_PROXY": {}, "HTTPS_PROXY": {}, "ALL_PROXY": {}, "NO_PROXY": {},
    "SSL_CERT_FILE": {}, "SSL_CERT_DIR": {},
    "NODE_TLS_REJECT_UNAUTHORIZED": {},

    // Git-driven command execution
    "GIT_SSH_COMMAND": {}, "GIT_TERMINAL_PROMPT": {},
    "GIT_EXEC_PATH": {}, "GIT_TEMPLATE_DIR": {},

    // Editor-as-shell
    "EDITOR": {}, "VISUAL": {},
}

func isBlockedEnvKey(key string) bool {
    upper := strings.ToUpper(key)
    if strings.HasPrefix(upper, "MULTICA_") {
        return true
    }
    _, ok := blockedEnvKeys[upper]
    return ok
}
```

Legitimate `custom_env` use cases (`ANTHROPIC_API_KEY`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_USE_BEDROCK`, etc.) are unaffected.

xiaoyaner0201:
## What does this PR do?

`isBlockedEnvKey` is the only gate between agent `custom_env` and the environment of the CLI child processes the daemon spawns. It currently rejects the `MULTICA_` prefix and 14 named keys. Well-known variables that grant code execution or redirect trust — `NODE_OPTIONS`, `LD_PRELOAD`, `GIT_SSH_COMMAND`, `HTTPS_PROXY`, `SSL_CERT_FILE` — are not among them, so an agent owner can set them and have the daemon hand them to `claude`, `codex`, `opencode`, `pi` and the rest on the daemon host.

This expands the denylist to cover dynamic-loader injection, interpreter and shell startup hooks, JVM launcher options, proxy/TLS redirection, git command sinks, editors and pagers, and npm configuration.

The change is **purely additive**. The existing `switch` keeps its original case arm untouched and new arms are appended, so no key that is blocked today becomes allowed. That property is the main design constraint here, not an incidental one — see below.

### Why prefixes and not just names

Four families are matched by prefix rather than enumerated: `LD_`, `DYLD_`, `NAPI_RS_`, `GIT_CONFIG_`, plus `BASH_FUNC_`.

A denylist fails whenever the attacker picks the family member nobody listed. `LD_PRELOAD` is the famous one, but `LD_AUDIT` loads a library through the rtld-audit interface just as effectively, and `DYLD_FALLBACK_LIBRARY_PATH` satisfies an otherwise-missing dylib. `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n` inject arbitrary git config — including `core.sshCommand` — without ever naming `GIT_CONFIG`. `BASH_FUNC_x%%` exports a shell function that a child `bash` executes on startup.

Prefix matching closes each family in one rule and keeps closing it as new members appear upstream. `GCONV_PATH` is listed by name because it is a loader-adjacent vector that does not carry the `LD_` prefix.

Over-blocking case variants is safe: the loader honours only exact-case names, so there is no legitimate lowercase counterpart to preserve.

### What is deliberately not blocked

Provider credentials (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GEMINI_API_KEY`), provider base URLs, and feature flags like `CLAUDE_CODE_USE_BEDROCK` are untouched. They are the legitimate use case for `custom_env` and the existing test asserts `ANTHROPIC_API_KEY` stays allowed.

## Related Issue

Closes #1114

This is the same problem #1123 set out to fix, and that PR's analysis and issue write-up did the work of identifying the vector classes. The difference is structural: #1123 replaces the `switch` with a new map, and in doing so drops eight keys `main` blocks today — `TMPDIR`, `TMP`, `TEMP`, `REASONIX_STATE_HOME`, `CURSOR_DATA_DIR`, `CURSOR_MCP_AUTH_SOURCE`, `OPENCLAW_CONFIG_PATH`, `OPENCLAW_INCLUDE_ROOTS` — which @ADanMan identified in review as a net regression and the reason to rebase and union rather than replace.

This PR takes the union approach from the start, and folds in the follow-up families from that same review: the glibc loader siblings, the dyld fallback paths, the interpreter module-search paths, `JDK_JAVA_OPTIONS`, and the git sinks that override the ones already listed (`GIT_SSH` over `GIT_SSH_COMMAND`, `GIT_EDITOR` over `EDITOR`).

Credit to @shaun0927 for the original report and PR, and to @ADanMan for the family-by-family review that this list is largely built from. If @shaun0927 would rather carry this forward in #1123, the diff here can be taken over wholesale — the goal is the hole closed, not this particular PR merged.

## Type of Change

## Changes Made

- **`server/internal/daemon/daemon.go`** — `isBlockedEnvKey`: add five prefix families (`LD_`, `DYLD_`, `NAPI_RS_`, `GIT_CONFIG_`, `BASH_FUNC_`) alongside the existing `MULTICA_` check, and append grouped `case` arms for named vectors. The original case arm is unchanged, with a comment recording that additions must stay additive.

  Groups, and why each is an execution or trust sink:

  | Group | Representative keys | Effect if attacker-controlled |
  |---|---|---|
  | CLI config / package injection | `XDG_*`, `OPENCODE_CONFIG_CONTENT`, `PI_PACKAGE_DIR`, `NODE_OPTIONS`, `NODE_PATH`, `BUN_OPTIONS` | Redirect a CLI's config root, or preload arbitrary JS before the CLI's own entrypoint |
  | Proxy / trust store / loader | `HTTP(S)_PROXY`, `SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `OPENSSL_CONF`, `SSLKEYLOGFILE`, `GCONV_PATH` | Route agent traffic through an attacker proxy and make it validate against an attacker CA |
  | Interpreter / shell / JVM | `PYTHONPATH`, `PYTHONHOME`, `PERL5LIB`, `RUBYLIB`, `BASH_ENV`, `ZDOTDIR`, `JAVA_TOOL_OPTIONS`, `JDK_JAVA_OPTIONS`, `CLASSPATH` | Shadow a module that gets imported at startup, or inject `-javaagent:` into any JVM launch |
  | Git command sinks | `GIT_SSH`, `GIT_PROXY_COMMAND`, `GIT_EXTERNAL_DIFF`, `GIT_ASKPASS`, `GIT_EXEC_PATH`, `GIT_TEMPLATE_DIR`, `GIT_TERMINAL_PROMPT` | Run an arbitrary command during ordinary repocache git operations; `GIT_TERMINAL_PROMPT` re-enables the prompt repocache disables on purpose |
  | Editors / pagers | `GIT_EDITOR`, `EDITOR`, `VISUAL`, `GIT_PAGER`, `PAGER` | Spawned via `sh -c` by commit/rebase flows and by any paginated output |
  | npm | `NPM_CONFIG_SCRIPT_SHELL`, `NPM_CONFIG_CAFILE`, `NPM_CONFIG_USERCONFIG` | Replace the shell npm runs scripts with, or its trust anchor |

- **`server/internal/daemon/daemon_test.go`** — extend `TestIsBlockedEnvKey` with 91 assertions covering every added key plus lowercase variants and prefix-family members that are not named individually (`NAPI_RS_FORCE_WASI`, `GIT_CONFIG_COUNT`, `LD_AUDIT`, `DYLD_FALLBACK_LIBRARY_PATH`, `BASH_FUNC_x%%`). Extend the `TestLayerCustomEnvAndHermesHome` blocklist case with a realistic attacker payload for each family and assert every one is absent from the resulting child environment, while the Hermes home overlay still applies.

## How to Test

```
cd server && go test ./internal/daemon/ -run 'TestIsBlockedEnvKey|TestLayerCustomEnvAndHermesHome' -v
```

End to end, on a self-hosted deployment:

1. Set `custom_env` on any agent to `{"NODE_OPTIONS": "--require=/tmp/evil.js"}`.
2. Write `/tmp/evil.js` on the daemon host containing `require('child_process').execSync('id > /tmp/pwned')`.
3. Assign the agent a task. Before this change `/tmp/pwned` appears; after it, the key is dropped and the task runs normally.
4. Set `custom_env` to `{"ANTHROPIC_API_KEY": "sk-..."}` and confirm it still reaches the CLI — credential passthrough is unaffected.

## Checklist

## AI Disclosure

**AI tool used:** Hermes Agent (Claude Opus)

**Prompt / approach:** The vector list was assembled by reading #1114 and the review on #1123, then working outward family by family: for each named key already under discussion, ask which siblings share the same execution or trust primitive and would be missed by an enumeration. Prefix matching was chosen where a family is open-ended upstream. The additive-union constraint came from re-reading the review's regression analysis against `main` rather than from the PR diff alone. Tests were written to assert the boundary in both directions — blocked keys absent, credential keys still present — and run against the daemon package before opening this PR.


multica-eve:
Thanks for the careful analysis and test coverage. We’re closing this PR for product/security-model reasons, not because the implementation is broken.

Multica’s documented security boundary is the operating-system user that runs the daemon. Tasks are treated as unsandboxed and already have that user’s file, credential, and network access. An agent owner who can configure `custom_env` can also invoke that agent, so variables such as `NODE_OPTIONS` or `LD_PRELOAD` change when code runs but do not grant a capability the task did not already have. The reproduction also requires first placing the payload on the daemon host (or having a normal task place it), which already exercises the same authority.

A denylist cannot create the missing isolation boundary—a task can set these variables again in commands—while this one would silently break legitimate proxy, private-CA, XDG, npm, and Pi configuration. We will keep the existing blocklist for daemon-owned invariants such as `MULTICA_*`, `PATH`, `HOME`, and task-scoped state.

If we later support untrusted workspace members sharing another user’s runtime, that needs a separate design around runtime invocation authorization and process/container/VM isolation, not an expanding environment-variable denylist.

Closing as not planned. Related issue: #1114.

PR #7946, opened 2026-09-02
MUL-6994 feat(cli): add --fields to issue list JSON output

eduardojmatos:
## Summary
- Adds `--fields` to `multica issue list` (JSON output only): a comma-separated whitelist of top-level issue keys to return.
- Description makes up ~81% of a typical issue payload (measured across 457 real outputs), so `--fields=id,title,status,priority` lets agent-driven reads skip pulling it.
- Omitting `--fields` returns the full issue object exactly as before — fully backward compatible, no default behavior change.
- Unknown field names are rejected up front with the valid list (same style as the existing `--sort` validation), rather than silently dropped.
- `--output table` is completely unaffected.

Field names are the real API field names (`server/pkg/publicapi/v1/types.go`), not table-display labels — e.g. assignee is `assignee_type`/`assignee_id`, and there is no `labels` field on an issue (labels/properties are separate resources), so those aren't accepted as field names.

## Test plan
- [x] `go test ./cmd/multica/...` — 924 passed, including new tests covering: multiple `--fields` combos (id/title/status/priority, assignee_type/assignee_id, single field, description explicitly requested, whitespace-tolerant parsing), default output unchanged (description still present), invalid field name rejected with helpful error, `--output table` unaffected by `--fields`.
- [x] `go vet ./cmd/multica/...` clean, `gofmt -l` clean.
- [x] Manually built the CLI and exercised it end-to-end against a fake `/api/issues` server: `--help` shows the flag, `--fields=id,title,status,priority` returns only those keys, no `--fields` still returns the full object, an invalid field name errors clearly, `--output table` with `--fields` is unaffected.
- [x] Updated `CLI_AND_DAEMON.md`'s Issues section with usage and semantics.

Ref: EDU-121


Bohan-J (changes requested):
Thanks for this — the product direction is right, and the shape is the one we'd want: JSON-only, opt-in, no default behavior change, unknown names rejected up front rather than silently dropped. Two things need fixing before merge.

### 1. The field whitelist is derived from the wrong contract

`validIssueFields` (`server/cmd/multica/cmd_issue.go:408`) is documented as the top-level keys of `server/pkg/publicapi/v1/types.go`, but `/api/issues` doesn't serialize that type — it serializes `handler.IssueResponse` (`server/internal/handler/issue.go:37`). That response carries two keys the whitelist is missing:

- **`status_name`** — always emitted (`issue.go:64`), and deliberately non-`omitempty`: MUL-6749 made it so precisely because a consumer holding only `status` is otherwise left with an opaque key like `in_review_2`.
- **`labels`** — bulk-attached by the list endpoint (`issue.go:1181`, inside `ListIssues`) so clients don't pay an N+1 per row.

Both are in the live payload today: a real `multica issue list --output json` returns 27 top-level keys; the whitelist has 25. So `--fields=labels` and `--fields=status_name` are rejected as invalid names even though they're exactly what a triage-style read wants. The PR description and `CLI_AND_DAEMON.md:510` also state there is no `labels` field on an issue, which isn't true for this endpoint.

`fullTestIssue()` (`server/cmd/multica/cmd_issue_test.go:3291`) inherits the same wrong source — it claims to carry "every field the Issue struct exposes" — so the tests lock the incorrect model in instead of catching it.

Please derive the list from the response this command actually parses, and add a guard that fails when the two drift apart. `TestIssueToMap_KeysMatchIssueResponse` in `server/internal/handler/issue_payload_shape_test.go` is the existing pattern for exactly this class of bug.

### 2. The bandwidth claim doesn't match where the filtering happens

`runIssueList` downloads and decodes the full response (`cmd_issue.go:771`, `client.GetJSON`) and only then drops keys locally (`:780` → `filterIssueFields`). That's a genuine win for stdout size and agent token cost, but the HTTP transfer, the server-side read, and the JSON parse are all unchanged.

The PR description ("lets agent-driven reads skip pulling it"), the flag help (`:515`), and `CLI_AND_DAEMON.md:510` (both "cuts token/bandwidth cost") promise something the implementation doesn't deliver. Please pick one:

- pass the projection to the server and apply it there, or
- narrow all three wordings to what this actually does — smaller CLI output and lower agent context cost.

---

Everything else looks good: no default behavior change, no persistence or migration, `--output table` untouched, and rollback is a single flag removal. `go test ./cmd/multica/...` and `go vet` pass on the branch here.

eduardojmatos:
Both fixed in 5b86a4580.

**1. Whitelist source.** Confirmed `/api/issues` serializes `handler.IssueResponse`, not `publicapi/v1.Issue` — `status_name` (always emitted) and `labels` (bulk-attached in `ListIssues`) were missing. Fixed `validIssueFields` to match, updated the doc comment to point at the real type, and added `TestValidIssueFieldsMatchListEndpointShape` — it builds a `handler.IssueResponse` the way `ListIssues` actually populates one (Reactions/Attachments/SourceContext deliberately left unset, since the list endpoint never sets them) and fails if the whitelist and the real response shape diverge again. Also refactored `fullTestIssue()` to derive from the same `handler.IssueResponse` fixture instead of a second hand-written key list, so the two can't drift from each other either.

**2. Bandwidth wording.** You're right that filtering happens client-side after the full response is already fetched over the wire — no server or network cost is actually saved. Rather than push a projection into the API contract for this, I narrowed the claim everywhere it appeared (flag help, `CLI_AND_DAEMON.md`) to what the flag actually does: reduces CLI output size and agent context cost, not network/server-side cost.

`go build`, `go vet`, and `go test ./cmd/multica/...` all pass on the updated branch.

Bohan-J (approved):
Both items are properly addressed — thanks for the thorough turnaround, especially for going after the root cause rather than just widening the list.

I re-verified the fix rather than just reading it:

- **Whitelist** now matches the real `/api/issues` list payload field-for-field (27 keys), and I confirmed end-to-end against a live server that `--fields=identifier,status,status_name,labels` returns those keys where it previously errored.
- **The drift guard earns its place.** I ran two mutations against it: dropping `labels` from the whitelist fails it, and — more importantly — adding a new field to `handler.IssueResponse` fails it too. That second case is exactly the drift that caused the original bug, and it now gets caught automatically. Deriving `fullTestIssue` from the same `handler.IssueResponse` source was the right call.
- **Wording** now matches where the filtering actually happens. Narrowing the claim instead of pushing a projection into the API contract was the right scope decision for this change.

One non-blocking note for whoever touches this next: the guard builds `sampleIssueResponse()` by hand, so it catches new non-`omitempty` fields automatically but would miss a future `omitempty` field that `ListIssues` populates. That's the same tradeoff as deliberately excluding `reactions`/`attachments`/`source_context`, and it's documented in the test — just worth knowing the boundary.

`go build`, `go vet`, and the full `./cmd/multica/...` suite pass here; CI is green.

Issue #5931, opened 2026-07-25
[Feature]: Add simple approval controls for sensitive actions


ystimen-beep:
### Deployment type

Official App

### What do you want and why?

When an agent needs to deploy to production, change or modify production data, or perform a destructive action, Multica should show a clear approval card in the issue.
The card should briefly explain what will be changed.
Then the owner can simply click **Approve** or **Reject**.
The approval must apply only to that exact action. If scope changes, the agent should request approval again.
For deploys, it would be useful to approve the first deploy in an issue and allow later redeploys of the same service without asking again. Secrets, production data, and destructive actions should always require a separate approval.

This would make production control clear and safe without relying on exact words in comments.

coderlfm:
同求这个功能，例如飞书的card中可以直接支持 接入按钮，用户只需要点击按钮进行决策

Issue #5528, opened 2026-07-16
[Feature]: Proactive Lark/Slack notifications for issue approval gates, with reply-to-advance


em998:
### Deployment type

Official App

### What do you want and why?

Problem

The Lark/Feishu (and Slack) bot integration is currently chat-reactive: it only responds when a user initiates a chat with the bot. Issue-level
events—especially approval gates where an agent @-mentions the user (e.g. "plan confirmation" or "merge to main")—are NOT pushed to Lark/Slack, and a
reply in Lark/Slack cannot flow back to advance the issue.

For self-hosted / single-user / mobile scenarios, this means the user must open Multica to see and act on approval gates, even though they already have a
Lark bot bound.

Use case

- ProductAgent reaches an approval gate in an issue and @-mentions the owner.
- The owner receives a proactive Lark/Slack notification (card) with the issue + what's being approved.
- The owner replies in Lark/Slack (e.g. "approve" / "allow merge to main") and the reply posts back to the issue and advances the flow—without opening
Multica.

Current workaround

User-initiated Lark chat: the owner messages the bot ("pending approvals?" / "approve issue-x") and the agent checks/acts in a chat session. Functional, but
requires the user to initiate—no proactive push.

Proposal

1. Proactive outbound: when an agent @-mentions a user in an issue comment (or an issue enters an approval-waiting state), push a notification to that
user's bound Lark/Slack bot (beyond the current chat-reply-only outbound).
2. Reply-to-advance: allow a Lark/Slack reply (keyword or free text) to post back as an issue comment and trigger the next flow step (e.g. approval ->
agent continues; rejection -> agent adjusts).
3. Reuse existing infra: lark_user_binding already links Lark identity to a Multica member; outbound.go already sends to Lark. Extending outbound to issue
events + bridging inbound replies to issue comments seems feasible.

Scope

- Applies to both Lark/Feishu and Slack (channel-generalized).
- Opt-in per workspace/user (notification preference), to avoid noise.

Tenember:
+1, and one detail that might lower the perceived cost of this feature.

The Slack bot integration already requests chat:write in its scope list (app_mentions:read, chat:write, reactions:write, users:read, commands, per the Slack Bot integration docs). So the outbound capability is already granted at the OAuth layer. Existing installs would not need a re-consent flow, which is usually the expensive part of adding an outbound path. What is missing is only the surface that exposes it to non-chat runs.

[...]

PR #7949, opened 2026-09-02
MUL-6996: fix(lark): preserve installation identity

YJ-Hao:
## What does this PR do?

Feishu channels receive the durable channel installation UUID through `channel.Config.ID`, but `newFeishuFactory` rebuilt its temporary `db.ChannelInstallation` from only `cfg.Raw`. That dropped the UUID before `installationFromRow` constructed the connector installation.

Card actions use the connector installation ID as an authority fence against the installation stored with the card. Losing the ID therefore makes valid callbacks look cross-installation and rejects them.

This PR carries `cfg.ID` into the reconstructed row. Workspace, agent, and installer identity continue to be resolved by the existing router path; only the durable row ID is preserved.

## Related Issue

Related to #5528 and #5931. This narrow Feishu installation-identity fix does not close either broader feature request.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Preserve `channel.Config.ID` in the Feishu factory when constructing the connector installation.
- Clarify that Feishu consumes the config ID for callback authority checks.
- Add a factory regression test with a nonzero installation UUID using the existing test helper.

## How to Test

1. Run `go vet ./internal/integrations/channel ./internal/integrations/lark`.
2. Run `go test ./internal/integrations/channel ./internal/integrations/lark -count=1`.
3. Confirm `TestFeishuFactory_PreservesInstallationID` fails without the production line and passes with it.

Local result in Go 1.26.7 Debian, non-root:

- `gofmt`: clean
- changed-package `go vet`: pass
- changed-package `go test`: pass
- `go test ./...` was also attempted in a container. All integration packages passed, while two unrelated daemon process-group tests failed under nested container PID semantics: `TestCombinedOutputKillsDescendantsHoldingOutput` and `TestCreateWorktreeContextCancelsRunningGitProcessTree`. Upstream CI remains the full-suite authority.

## Thinking Path and Risk

The channel engine already provides the database row UUID in `channel.Config.ID`. Other Feishu identity fields intentionally remain resolver-owned. Propagating that existing UUID at the factory boundary is smaller and safer than weakening the callback installation comparison or resolving by App ID.

Risk is limited to Feishu channel construction. Invalid or absent IDs remain invalid and continue to fail closed. No schema, API, credential, callback payload, or authorization rule changes.

## Checklist

- [x] I have run relevant tests locally and they pass
- [x] This change does not affect UI
- [x] No user-facing documentation changes are required

## AI Disclosure

**AI tool used:** AI coding assistant

**Prompt / approach:** Assisted with tracing the callback authority mismatch, reducing the fix to the factory boundary, adding the regression test, and running repository-aligned verification.

### DIFF
--- a/server/internal/integrations/channel/channel.go
+++ b/server/internal/integrations/channel/channel.go
@@ -91,7 +91,7 @@ type Config struct {
 	// in-tree does today, but Factory implementations should tolerate it
 	// rather than assume Valid). WeCom uses it to key its per-connection
 	// wsSender into a shared registry the OutboundReplier looks up by;
-	// Feishu and Slack don't currently read it.
+	// Feishu uses it for callback authority checks; Slack does not currently read it.
 	ID pgtype.UUID
 
 	// Handler is the shared inbound entry point the engine injects so the
--- a/server/internal/integrations/lark/feishu_channel.go
+++ b/server/internal/integrations/lark/feishu_channel.go
@@ -26,10 +26,9 @@ import (
 // through the Lark HTTP API. One instance is built per channel_installation by
 // the registered Factory; the connector is shared across instances.
 //
-// The Channel holds only the credentials it needs for Connect/Send (decoded
-// from the per-installation config blob). The installation IDENTITY
-// (workspace / agent / installer) is resolved per message by the Router's
-// InstallationResolver, so it is deliberately absent here.
+// The Channel holds connection credentials plus the durable installation row
+// ID. Workspace, agent, and installer identity are resolved per message by the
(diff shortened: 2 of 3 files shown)


multica-eve:
Thanks for the contribution. We are closing this because the current Feishu connector does not handle card-action events, so propagating the installation ID only changes log correlation and cannot fix the described callback authorization behavior. A future proposal should either implement card-action ingestion with same-installation success and cross-installation rejection coverage, or be scoped explicitly as an observability change.

PR #7951, opened 2026-09-02
fix(wecom): re-read the installation's permission at every stage of a file delivery

seacen:
## What

A file delivery re-reads the installation's permission at every stage that is about to put bytes toward the chat: after the object read, before **each** chunk, and after the upload before the `media_id` is sent.

## Why

A revocation is a person withdrawing the connection, and a file delivery is minutes of work between the top-of-loop gate and the send — the object read, the chunked upload, the finish. A withdrawal landing anywhere in there reached a worker with nothing left to consult, so the file went into a chat whose owner had removed the bot. This is the independent piece from #6606 that you pointed at.

## How

- `uploadMedia` takes an optional `BeforeChunk func(ctx) error`, asked **inside the goroutine that writes**, ahead of every network attempt including retries — so nothing waits between asking and writing. The dispatch loop keeps only a `gctx.Err()` break.
- `sendAttachment` re-reads the row through `mayStillWrite`, which has three outcomes rather than a bool: a row that says revoked is final, a read that failed is this attempt only. Both stop this delivery; the log line says which.
- The `upload_media_finish` frame is deliberately unchecked: it carries no file bytes, and a revocation landing between the last chunk and finish produces a `media_id` the post-upload check then refuses to send.

## Tests

- `TestUploadMediaChunks_AWithdrawalDuringTheSlotWaitStopsTheNextChunk` — five chunks at parallelism three, three held in flight, revoke, release: three chunk frames and no finish.
- `TestUploadMediaChunk_AWithdrawalBeforeTheRetryStopsTheSecondOffer` — one chunk, first verdict dropped, permission withdrawn the moment the first frame exists: one frame on the wire.
- `TestSendAttachment_ARevokeInTheDatabaseMidUploadStopsTheFile` — the real query layer against Postgres: an `UPDATE channel_installation SET status = 'revoked'` while three chunks are held, asserting no finish frame and no send.

Each guard was reverse-verified on its own: with the check hoisted out of the attempt loop the first test passes and the second still fails; with the three re-checks stripped from `sendAttachment` the database test reports the file delivered.

### DIFF
--- a/server/internal/integrations/wecom/attachment_revocation_db_test.go
+++ b/server/internal/integrations/wecom/attachment_revocation_db_test.go
@@ -1,0 +1,95 @@
+package wecom
+
+// attachment_revocation_db_test.go — the attachment write path re-reads the
+// installation's permission from the DATABASE, not from a mock, at each stage
+// that is about to put bytes toward the chat. The unit tests beside it
+// (media_upload_test.go) prove the BeforeChunk hook stops the next chunk; this
+// one proves the hook is wired to the real row, so a person revoking the bot
+// mid-upload — an UPDATE on channel_installation — is what stops the file.
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"testing"
+	"time"
+
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+func TestSendAttachment_ARevokeInTheDatabaseMidUploadStopsTheFile(t *testing.T) {
+	pool := twoReplicaDB(t)
+	ctx := context.Background()
+
+	wsID, userID, agentID, instID := mustTestUUID(t), mustTestUUID(t), mustTestUUID(t), mustTestUUID(t)
+	tag := uuidStringPub(instID)[:8]
+	exec := func(sql string, args ...any) {
+		t.Helper()
+		if _, err := pool.Exec(ctx, sql, args...); err != nil {
+			t.Fatalf("seed: %v\n%s", err, sql)
+		}
+	}
+	t.Cleanup(func() {
+		_, _ = pool.Exec(ctx, `DELETE FROM channel_installation WHERE id = $1`, instID)
+		_, _ = pool.Exec(ctx, `DELETE FROM agent WHERE id = $1`, agentID)
+		_, _ = pool.Exec(ctx, `DELETE FROM "user" WHERE id = $1`, userID)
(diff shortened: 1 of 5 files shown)


seacen:
@Bohan-J A nudge on this one — it is the revocation re-check you named on #6606 as worth its own PR, and it has been open since the 2nd with no conflicts.

Re-verified against today's `main` (`3551e72e7`) rather than the base it was cut from: cherry-picks clean, and `go test -race ./internal/integrations/wecom` is green against a database migrated through 467. 60 lines of production code, three tests, one of them database-backed.

seacen:
Rebased onto `8c4f4328f` and pushed as `14605f2d8`. Doing that turned up something worth knowing before this gets reviewed, because it is the shape of problem this repo's CI cannot see.

**This PR was green, mergeable, and would have broken `main`.** Its checks ran on 09-02. #8318 landed on 09-14 and removed `mediaConn.dropAcks` along with the tests that used it, and one test here still used it. The rebase applied without a conflict — git had nothing to report, since neither side edited the same lines — and the package then failed to build. Nothing re-ran the checks in between, so the PR page kept saying what it said twelve days ago.

I have removed that one case rather than restoring the field. #8318 deleted those tests because each stood still for a whole `ackTimeout`, and `ackTimeout` is a package constant with no seam, so bringing the harness back would bring back exactly the test time that PR was cutting. The gap is recorded where the test was, including what closing it properly would cost: one seam, `ackTimeout` as a field on `wsSender` rather than a constant, which is a change to production code for a test's benefit and is not something I wanted to slip into a PR about revocation without asking.

**Say the word and I will add that seam** — here or separately. The property it would restore is narrow and real: a withdrawal landing between a lost verdict and the chunk's second offer. The rest of the revocation coverage is unaffected, and `uploadMediaChunk` calls `beforeChunk` at the top of every attempt, so the check the retry needs is the same call the first attempt still exercises.

Everything else is unchanged from what you have not yet seen: the re-read at each stage of a file delivery, which is the piece you named on #6606 as worth its own PR.

`go build`, `go vet` untagged and under `agentintegration`, `-race -p 1` on the touched package differenced against the merge base, the DB-backed test proven to have RUN rather than skipped, and the merge with `main` compiled.

No rush on my side, but I would rather it not sit at a base that old again — if it is going to wait, I will keep rebasing it so the green on the page means something.

PR #7952, opened 2026-09-02
MUL-7004: feat(wecom): report a failed run to the chat that asked

seacen:
## What

WeCom subscribes to `task:failed` and `task:cancelled`. A failed run now reaches the chat that asked; until now it said nothing and logged nothing, because no handler ran.

## Why

`wecom/outbound.go` subscribed to `EventChatDone` and `EventInboxNew` only, while DingTalk, Lark, Slack and Telegram all handle `EventTaskFailed`. You confirmed this on #6606 as a plain bug, independent of anything else there.

## How

- `task:failed` goes out on the **same route a reply takes**: task delivery row, origin gate, installation status, socket or relay. There is no second path to reason about.
- The text is the platform's own redacted failure text behind a warning mark — `⚠️ <error>` — the way DingTalk delivers it. Nothing is sent while an auto-retry is pending: the retry attempt reports its own outcome, and announcing a failure the next attempt may undo would be noise.
- `task:cancelled` is subscribed for the log line only. A cancellation ends the run without an answer, and WeCom has nothing to take back — no typing badge, no placeholder. Lark makes the same choice.

## Tests

- the failure reaches the bound chat, as one `aibot_send_msg`;
- a retry-pending attempt, a failure with no text, and a web-UI run's failure on a session that also lives in a WeCom group all stay silent (the origin gate is the reply path's);
- the wiring through a real `events.Bus`, with a cancellation publishing nothing;
- the same failed run through the real query layer against Postgres (`TestTaskFailed_ReachesTheChatThroughTheRealQueryLayer`).

Reverse-verified: with the two subscriptions removed the bus-wired tests fail; with the content helper reverted to `chatDoneContent` the `processEvent` test fails.

### DIFF
--- a/server/internal/integrations/wecom/outbound.go
+++ b/server/internal/integrations/wecom/outbound.go
@@ -33,6 +33,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


multica-eve (approved):
Thanks for splitting this into a focused fix and for the careful regression coverage. This closes the silent-failure gap cleanly by reusing the existing WeCom reply route while preserving the retry and origin safeguards. Approved — appreciate the thoughtful work here.

PR #7953, opened 2026-09-02
MUL-7005 fix(wecom): settle a relayed reply's outcome once, after its last offer

seacen:
## What

`deliverRelayed` settles a relayed reply's outcome once, after its last offer, instead of once per attempt.

## Why

It recorded the outcome before deciding whether the frame would be offered again. A frame that fails before the write is re-offered across the whole retry chain (eleven entries on the production defaults) and the publisher's `watchOutcomes` settles it once more, so one reply landed on `outbound_dropped` twelve or thirteen times — and when a later offer succeeded, on `outbound_delivered` as well. That is the "delivered and dropped at once" defect the package's own comment says it no longer has (#7428's counters, #7429's relay).

## How

`provablyNotSent` is asked first. A frame still owed an offer returns `outcomeProvablyNotSent` with no counter moved; only a frame that will not be offered again is counted here. The publisher's `watchOutcomes` remains the single owner that settles a routed reply nobody took.

## Tests

Both drive the **real** `deliverRelayed` through the real dispatcher and retry plan, on a connection double whose `SetWriteDeadline` can fail — the existing doubles always return nil there, which is why no existing test could produce a provably-unsent error on this path.

- `TestRelayedReply_AFrameStillInFlightRecordsNoReplyOutcome` — re-offered eight times, no reply outcome recorded (was `outbound_dropped = 8`).
- `TestRelayedReply_ARetryThatSucceedsIsNotAlsoADrop` — one delivered, zero dropped (was `outbound_dropped = 1`).

### DIFF
--- a/server/cmd/server/main.go
+++ b/server/cmd/server/main.go
@@ -47,6 +47,26 @@ func newNamedRedisClient(base *redis.Options, suffix string) *redis.Client {
 	return redis.NewClient(&opts)
 }
 
+// newClaimRedisClient is a named client that honours its callers' context
+// deadlines.
+//
+// go-redis discards them by default (Options.ContextTimeoutEnabled): a command
+// is bounded by the socket timeout instead, so a caller's deadline reaches the
+// wire only as a suggestion. That is the right default for the relay's own
+// publish traffic, where nobody is holding a stopwatch, and the wrong one for
+// the WeCom claim store, whose callers spend budgets they have promised to
+// keep — DedupeStore.ClaimBudget is what sizes the dispatcher's outcome grace,
+// and a shutdown drain gives its whole sequence of round trips one DrainBudget.
+//
+// Hence a dedicated client rather than the flag on the shared relay client:
+// setting it there would change the timeout behaviour of every publish that
+// runs through it, which is a far wider blast radius than this store needs.
+func newClaimRedisClient(base *redis.Options, suffix string) *redis.Client {
+	opts := *base
+	opts.ContextTimeoutEnabled = true
+	return newNamedRedisClient(&opts, suffix)
+}
+
 func redisClientName(existing, suffix string) string {
 	if suffix == "" {
 		return existing
@@ -395,6 +415,7 @@ func main() {
 	var storeRedis *redis.Client
 	var channelLeaseRedis *redis.Client
 	var relayWriteRedis *redis.Client
+	var wecomClaimRedis *redis.Client
 	var relayReadRedis *redis.Client
 	var shardedReadRedis *redis.Client
 	var legacyReadRedis *redis.Client
@@ -427,6 +448,7 @@ func main() {
 		closeRedisClient("realtime-read-legacy", legacyReadRedis)
 		closeRedisClient("realtime-read-sharded", shardedReadRedis)
 		closeRedisClient("realtime-read", relayReadRedis)
+		closeRedisClient("wecom-claim", wecomClaimRedis)
 		closeRedisClient("realtime-write", relayWriteRedis)
 		closeRedisClient("channel-lease", channelLeaseRedis)
 		closeRedisClient("store", storeRedis)
@@ -485,8 +507,9 @@ func main() {
 						"error", err)
 					leaseSettle = 0
 				}
+				wecomClaimRedis = newClaimRedisClient(opts, "wecom-claim")
 				wecomRelayOutbound = wecom.NewRelayOutbound(wecomRelay,
-					wecom.NewRedisDedupe(relayWriteRedis, 0, slog.Default()),
+					wecom.NewRedisDedupe(wecomClaimRedis, 0, slog.Default()),
 					wecom.RelayConfig{
 						ReplayGrace: relayConfig.ReplayGrace,
 						LeaseSettle: leaseSettle,
--- a/server/cmd/server/wecom_claim_redis_wiring_test.go
+++ b/server/cmd/server/wecom_claim_redis_wiring_test.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/redis/go-redis/v9"
+)
+
+// The WeCom claim store is the one Redis user here whose callers spend budgets
+// they have promised to keep: DedupeStore.ClaimBudget sizes the dispatcher's
+// outcome grace, and a shutdown drain gives its whole sequence of round trips
+// one DrainBudget. Neither is real unless the client applies those deadlines
+// to the wire, and go-redis does not by default — it waits out ReadTimeout and
+// ignores the context (baseClient.context).
+//
+// So the claim store gets its own client with the flag set, and the shared
+// relay client keeps the default: turning it on there would change the timeout
+// behaviour of every realtime publish that runs through it.
+//
+// This pins the two constructors, not main()'s choice between them; that the
+// dedupe store is handed the right one is guarded at runtime instead, by the
+// warning NewRedisDedupe logs for a client that ignores deadlines.
+func TestClaimRedisClientHonoursContextDeadlines(t *testing.T) {
+	base := &redis.Options{Addr: "localhost:6379", ClientName: "multica", ReadTimeout: 3 * time.Second}
+
+	claim := newClaimRedisClient(base, "wecom-claim")
+	defer claim.Close()
+	if !claim.Options().ContextTimeoutEnabled {
+		t.Fatal("the claim client ignores context deadlines: ClaimBudget and the drain budget " +
+			"would be numbers nothing enforces")
+	}
+
+	// The client it is derived from keeps go-redis's default, so this stays a
+	// scoped change rather than a fleet-wide one.
+	shared := newNamedRedisClient(base, "realtime-write")
+	defer shared.Close()
+	if shared.Options().ContextTimeoutEnabled {
+		t.Fatal("the shared relay client now enforces context deadlines too: that is a wider " +
+			"blast radius than the claim store asked for")
+	}
+
+	// Everything else about the client is still the deployment's.
+	if got := claim.Options().ReadTimeout; got != base.ReadTimeout {
+		t.Fatalf("ReadTimeout = %v, want the configured %v", got, base.ReadTimeout)
+	}
+	if got := claim.Options().ClientName; got != "multica:wecom-claim" {
+		t.Fatalf("ClientName = %q, want the suffixed name", got)
+	}
+	if base.ContextTimeoutEnabled {
+		t.Fatal("newClaimRedisClient mutated the options it was handed")
+	}
+}
--- a/server/internal/integrations/wecom/dedupe_redis.go
+++ b/server/internal/integrations/wecom/dedupe_redis.go
@@ -72,11 +72,13 @@ func (d *redisDedupe) Held(ctx context.Context, key string) (bool, error) {
 // Best effort by design: a Release that fails leaves a key that expires on its
 // own, which costs one un-retried delivery in the replay window rather than a
 // duplicate in somebody's chat.
-func (d *redisDedupe) Release(ctx context.Context, key string) {
+func (d *redisDedupe) Release(ctx context.Context, key string) error {
 	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), d.budget)
 	defer cancel()
 	if err := d.rdb.Del(ctx, key).Err(); err != nil {
 		d.log.WarnContext(ctx, "wecom relay: could not release a delivery claim",
 			"error", err, "key", key)
+		return err
 	}
+	return nil
 }
--- a/server/internal/integrations/wecom/dedupe_redis.go
+++ b/server/internal/integrations/wecom/dedupe_redis.go
@@ -7,9 +7,11 @@ package wecom
 // cross-replica routing and nothing to deduplicate. A deployment that never
 // needed Redis is not asked for it now.
 //
-// SET NX PX is the claim. One key per turn, keyed on the same event id the
-// stream entry carries, so a frame replayed after a restart and a frame read
-// by two replicas mid-lease-move meet the same key.
+// One key per turn, keyed on the same event id the stream entry carries, so a
+// frame replayed after a restart and a frame read by two replicas
+// mid-lease-move meet the same key. The value is the owner's token while the
+// delivery is in flight, then "settled" (its holder recorded the outcome) or
+// "lost" (the publisher did).
 
 import (
 	"context"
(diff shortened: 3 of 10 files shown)


Bohan-J (changes requested):
The product direction is right: a relayed reply should land on exactly one outcome, and settling it after the last offer is the correct place to do that. One blocking issue before this can merge.

## Must fix: a failed claim release leaves the reply with no outcome at all

`deliverRelayed` now returns `outcomeProvablyNotSent` before any counter moves (`server/internal/integrations/wecom/relay_outbound.go:1021`), handing the outcome to the single owner that settles it afterwards. That handoff assumes the claim actually goes back — but `DedupeStore.Release` is best-effort and reports nothing to its caller: `redisDedupe.Release` logs a failed `DEL` and returns void (`server/internal/integrations/wecom/dedupe_redis.go:75`).

If that `DEL` fails, the key survives to its TTL, and from there:

- every re-offer loses the claim at `relay_outbound.go:784` and never reaches the socket again;
- `settle` then reads the key as still held at `relay_outbound.go:923` and returns — "somebody took it";
- so the reply reaches nobody while `delivered`, `dropped` and `unconfirmed` all stay at zero.

Before this PR the same run recorded at least one `transport_error` drop on the first local failure. This PR removes that last signal, so a Redis blip on `DEL` turns a lost reply into a silent one — observable only indirectly, through the `could not release a delivery claim` warning.

Suggested shape of the fix: let `Release` return success/error to its caller. Released cleanly → keep the new behaviour and wait for the next offer. Release failed → the claim owner, the only party that can, records one terminating outcome before returning. An explicit state instead of a bool is equally fine; what cannot stand is a key whose presence means both "delivered" and "failed before the write, and the release failed too".

Please cover it with a regression test: release fails, key stays, assert exactly one dropped and zero delivered. The new tests run with `dedupe = nil` (`server/internal/integrations/wecom/relay_relayed_send_test.go:153`), so this production path is untested.

## Otherwise

The change reuses the existing relay, claim and outcome-watcher machinery — no parallel mechanism and no unexplained tuning knobs. Moving the "will this frame be offered again?" decision ahead of the counters matches how this package already assigns outcome ownership, and the two new tests drive the real `deliverRelayed` through the real dispatcher, which is the right level for these properties.

Close the release/outcome hole and add that regression test, and this looks good to merge.

seacen:
@Bohan-J Fixed in `a595d550f`. `Release` reports failure now. A clean release keeps the new behaviour — the frame is owed another offer. A failed one is ended by the replica holding the claim, the only party that still can: one `transport_error` drop for a reply (an inbox push logs and moves no reply counter), and no further offer.

`TestRelayedReply_AClaimThatCannotBeReleasedIsCountedOnceAndNotReoffered` drives the real dispatcher with a claim store whose `Release` fails and keeps the key: exactly one dropped, zero delivered, one offer, the key still held. With the error ignored it fails on `outbound_dropped = 0`.

Bohan-J (changes requested):
Re-reviewed at `a595d55`. The round-1 hole is genuinely closed: `Release` now reports failure to its caller, and `strandedClaim` records one terminating outcome for a claim that provably stayed put (`server/internal/integrations/wecom/relay_outbound.go:808-822`), with a test that covers it. The product direction is unchanged and still right.

One blocking issue remains, in the new code.

## Must fix: a `DEL` error does not prove the key survived

`redisDedupe.Release` returns `Del(...).Err()` verbatim (`server/internal/integrations/wecom/dedupe_redis.go:78`), and the caller reads any error as the single fact "the key survives to its TTL" (`relay_outbound.go:808-817`), records a drop, and stops offering. But a network error or a read timeout can be raised *after* the server already executed the `DEL`. The client error means the outcome is **unknown**, not that the key is still there.

Two concrete ways that breaks the invariant this PR exists to restore:

1. **Dropped and delivered for the same reply.** Server deletes the key, the response is lost. This replica records `transport_error` in `strandedClaim` (`relay_outbound.go:829`). Meanwhile another replica that lost the claim earlier is still inside its backoff chain; its next `Claim` now succeeds against a key that is really gone, it owns the socket, and it delivers — `outbound_delivered` +1 against an `outbound_dropped` that is already counted. A relay replay after a restart reaches the same state.
2. **Dropped twice for the same reply.** Even with no redelivery, the publisher's `settle` runs at the end of the grace and reads the key as absent, so it records a second drop under `no_connection` (`relay_outbound.go:946-958`) on top of `strandedClaim`'s `transport_error`.

There is also a fencing gap behind this. The claim is a bare `SetNX(key, "1", ttl)` with no owner token (`dedupe_redis.go:55-58`), and the relay Redis client is built without a `MaxRetries` override (`server/cmd/server/main.go:489`), so go-redis retries network errors by default: a re-sent `DEL` can delete a claim that a *later* owner has since taken, converting one unlucky release into someone else's duplicate-delivery window.

The test that covers this path locks in the single interpretation rather than exposing it: the fake's `Release` returns an error and keeps the key (`server/internal/integrations/wecom/outbound_two_replica_db_test.go:380-387`), which is exactly the branch `TestRelayedReply_AClaimThatCannotBeReleasedIsCountedOnceAndNotReoffered` asserts (`server/internal/integrations/wecom/relay_relayed_send_test.go:286-313`). Nothing models "the delete landed and the answer was lost."

### What I would do instead

Do not treat a plain `error` as a determinate "not released". This repo already has the pattern for it — token-fenced Redis state via single Lua compare-and-delete in `server/internal/integrations/channel/engine/redis_lease_store.go:17-46`. Mint a unique owner token per claim, make release a compare-and-delete that is safe to retry, and let the final settle distinguish three states rather than two: delivered, provably not delivered, and *release outcome unknown*. The unknown case should resolve to exactly one record without racing a replica that may still legitimately win the claim.

Please add regression coverage for the case the current fake cannot produce: `Release` deletes the key **and** returns an error, then another replica (or a replay) claims and delivers. On the current implementation that ends with `dropped = 1` and `delivered = 1`; the fix should make it exactly one.

## Non-blocking

`strandedClaim` logs `wecom outbound: reply not delivered` for inbox frames too (`relay_outbound.go:831`), even though the counters correctly stay put for them. The rest of this package keeps an inbox-specific message; worth restoring here so the log does not read as a lost agent reply during triage.

Close the unknown-release case and add that cross-replica regression test, and this is ready.

seacen:
@Bohan-J You were right that a `DEL` error is not a verdict, and round two read it as one. Fixed in `a238c1b8b`, along the lines you pointed at.

**The claim carries an owner token, and every mutation is one Lua operation on it** — the `redis_lease_store.go` shape:

- `Claim` takes the key for the token, or **re-takes** it when the key already holds that token: the same owner coming back after an unknown release. A re-sent `SET NX` can no longer make an owner lose its own claim.
- `Release` is compare-and-delete, safe to retry: a retry after the delete landed reports "not ours" rather than deleting a later owner's claim. An error is read as exactly what it is — unknown — and **nothing is recorded on it**. The next offer's `Claim` decides: it re-takes a key that still holds the token, takes a deleted one fresh, or loses to the replica that got there first, which then delivers and records once.
- `Settle` is compare-and-set token → `settled`, made by the holder **before** it records the outcome; a refusal means the publisher already resolved the reply as lost, and the holder then records nothing. `deliverRelayed` now hands the dispatcher its record as a closure so the record can wait for `Settle`.
- `Resolve` is the publisher's end-of-grace read, and it fences a key still held by a token as `lost` in the same operation. So `settle` sees four states — absent, held, settled, lost — and records only for absent (nobody took it) and held (a holder that never managed to record anything); a late holder finds its `Settle` refused.

The grace grows by one delivery's budget (`RelayConfig.DeliveryBudget`, default `ackTimeout`), so the last offer's own ack wait cannot be fenced from under it. With the token on every mutation, go-redis's default retries are safe: a re-sent command can never act on a claim a later owner holds.

**Tests, each reverse-verified:**

- `TestTwoReplicas_AReleaseThatLandedButErroredEndsWithExactlyOneRecord` — the case you asked for: release deletes the key and returns an error, then another replica (a second socket holder in the rig) or the holder's own re-offer claims and delivers. One delivered, zero dropped, across every replica and the publisher. With a drop recorded on the release error it fails on `outbound_dropped = 1` beside the delivery.
- `TestTwoReplicas_AStrandedClaimIsResolvedOnceByThePublisher` — every write fails before the frame leaves and every release errors with the key kept: the publisher records one `transport_error`, the holder nothing, and the key is left fenced as `lost`. With `Resolve` reporting a held key as settled, the reply ends with no record.
- `TestRelayedReply_AReleaseWhoseResultIsUnknownIsReclaimedByTheNextOffer` / `…ThatLandedButErroredIsTakenFreshByTheNextOffer` — the two faces of an unknown release on one replica, both ending delivered once. Without the re-entrant `Claim` nothing is ever delivered.
- `TestRedisDedupe_TokenFencedOperations` — the four Lua operations against a real Redis, one assertion per rule (needs `REDIS_TEST_URL`, like the ordering tests).

The fake now models both faces of a failed release (`releaseFails`, `releaseErrAfterDelete`) rather than one.

**Non-blocking:** `strandedClaim` is gone with the approach; the publisher's `settle` is reply-only (`publish` enrols only `relayKindReply`), and the dispatcher's release log names the frame kind.

Bohan-J (changes requested):
Re-reviewed at `a238c1b8b`. The round-2 blocker is properly closed: claims are token-fenced through Lua compare-and-act, `Release` is now safe to retry, and the asymmetric case I asked for is covered by `releaseErrAfterDelete` (`server/internal/integrations/wecom/outbound_two_replica_db_test.go:405-407`). That is the right shape.

Two things left, both small. I am deliberately **not** asking for a redesign here — see the last section.

## 1. `Settle` has the same unknown-outcome hole that `Release` just had

When `Settle` returns an error the holder returns without recording (`server/internal/integrations/wecom/relay_outbound.go:875-888`), so `res.record()` at `:896` never runs. If Redis executed the `SET` and only the response was lost, the publisher's `Resolve` later reads `claimSettled` and returns at `:1061-1062`. The reply reached the user and `delivered`, `dropped` and `unconfirmed` are all zero.

This is the same failure mode this PR already fixed one level up: an error from the store means *unknown*, not *did not happen*. `Release` got that treatment; `Settle` did not.

The fix should be cheap, because `redisSettleSource` is already idempotent and token-fenced (`dedupe_redis.go:81-91`): a bounded retry of `Settle` resolves it correctly in both directions. If the first call did land, the retry matches `v == settled` and returns 1, so the holder records once. If it never landed and the publisher has since fenced the key as `lost`, the retry returns 0 and the holder takes the existing "delivered after the publisher resolved the reply as lost" branch — no double count either way.

Please also add the test symmetric to `releaseErrAfterDelete`: a `Settle` that writes the state **and then** returns an error, asserting exactly one outcome. Today `settleFails` (`outbound_two_replica_db_test.go:411-416`) only models "did not execute", and no test in the repo sets it at all.

## 2. `outcomeGrace` omits a full round trip per offer

`outcomeGrace()` counts one `Claim` budget per offer, the backoffs, and one delivery budget (`relay_outbound.go:952-973`). Each offer also runs `Release` synchronously (`:920`), and the final one runs `Settle` — each on the same store budget, none of it counted.

At production defaults (`LeaseSettle` = `engine.DefaultPollInterval` = 30s, `RetryBackoff` = 200ms) the plan is 11 entries / 12 offers, backoffs total 50.1s, and the grace is 79.1s — against up to **24.0s** of unaccounted release/settle round trips.

That matters because the `absent` state is not fenced: if the watcher's `Resolve` lands after a successful `Release` and before the next `Claim`, it records `no_connection` at `:1080`, and a later offer can still claim, deliver and settle — recording `delivered` for a reply already counted dropped. That is exactly the contradiction this PR exists to remove. Queue residency pushes the same way, since the grace clock starts at publish while the frame can still be waiting in a shard queue.

Including the second round trip per offer in the grace is a one-line change and closes the arithmetic gap.

One thing to avoid: please do **not** fix this by fencing `absent` in `Resolve`. That would turn a premature grace expiry into a claim no later offer can take, converting a miscount into a reply the user never receives — strictly worse for the user than the bug being fixed.

## Note, not a blocker: legacy claim values

This changes the value stored under the existing key namespace from `"1"` to a token / `settled` / `lost`, and `redisResolveSource` fences any unrecognised value as a live token, so a legacy `"1"` would be recorded as a drop.

I checked the exposure and it is small: the chart deploys with `strategy: Recreate` and `replicas: 1` (`deploy/helm/multica/templates/backend.yaml:34-35`), so no mixed-version overlap there, and `awaitOutcome` is only called from `publish` (`relay_outbound.go:786`) — the consume/replay path enrols nothing, so a restarted process does not resolve legacy keys for replayed frames. What remains is a deployment that scales replicas up *and* switches to a rolling strategy.

Worth either a guard in `redisResolveSource` that only fences token-shaped values (`tokenFor` is `owner + "/" + eventID`, so it is recognisable) or a line in the PR description recording the boundary. Either is fine; it should not hold the merge.

## What I am not asking for

An earlier round suggested making the publisher the sole counter, with the holder only persisting a terminal state. It would close item 1 more completely, including the process-death window between a successful `Settle` and `record()`. I do not think it is worth it here: it delays every `delivered` by a full grace (~79s at defaults) to close a window that is much narrower than the one the bounded retry already handles, and the residual is the ordinary non-atomicity of any design without a transactional outbox.

Land the two changes above and this is good to merge.

seacen:
@Bohan-J Both landed in `3119a1f4e`, and the legacy-value note is closed in code rather than in the description.

## 1 — `Settle` gets the same treatment `Release` got

An error is unknown, so it is retried rather than read as a refusal. Three attempts, paced by the claim layer's own backoff, which works in both directions exactly as you set out: a retry after a settle that DID land matches `v == settled` and the holder records once; one after a settle that never landed either settles it now or meets the publisher's fence and takes the existing "delivered after the publisher resolved the reply as lost" branch.

Attempts exhausted means still unknown. Nothing is recorded there: the claim is then held by this token, which is the one state `Resolve` turns into a record of its own, so counting here as well would be the double record this path exists to prevent.

Tests, both reverse-verified:

- `TestTwoReplicas_ASettleThatLandedButErroredEndsWithExactlyOneRecord` — the symmetric case you asked for. The fake now has `settleErrAfterWrite` beside `settleErrBeforeWrite`, so "the SET ran and the response was lost" and "the SET never reached the server" are separate shapes rather than one. Exactly one record, delivered. Without the retry it fails with **no** record at all — the holder stays quiet and `Resolve` reads `claimSettled` and stays quiet too, which is the hole you described.
- `TestTwoReplicas_ASettleNobodyCanCompleteIsEndedOnceByThePublisher` — the store answers nothing at all: the holder records nothing, the publisher's `Resolve` finds the claim still held and ends it once. With a record on exhaustion it fails with two.
- `TestRelayedReply_ASettleThatNeverExecutedIsRetried` — one delivery, one record, and one offer: a settle retry is not a re-delivery.

## 2 — the grace counts both round trips per offer

`outcomeGrace` now counts a Claim **and** the Release-or-Settle each offer ends in, plus the settle retries and their pauses. At the defaults that moves it from 79.1s to 107.5s against a worst case of 107.5s; `TestRelayOutcomeGrace_CoversEveryStoreRoundTripTheChainCanMake` pins the arithmetic and fails on the old shape with the numbers in the message.

`absent` is left unfenced, for the reason you gave: fencing it would turn a premature expiry into a claim no offer can take, which costs the user the reply rather than the count. The comment says so where the decision lives.

## 3 — legacy claim values

Guarded in `redisResolveSource` rather than left to the description. Every token carries the `/` that `tokenFor` puts there, so a value without one is a pre-token claim whose holder recorded its own outcome inline: reported as `claimSettled` and left untouched. `TestRedisDedupe_TokenFencedOperations` seeds a literal `"1"` against a real Redis and asserts both halves — reported settled, and not rewritten. Dropping the guard fails it.

## Verification

`go test -race` green on `./internal/integrations/wecom ./cmd/server ./internal/integrations/channel/engine ./internal/events`, with `REDIS_TEST_URL` set so the Lua tests run rather than skip. Whole tree: 64 packages ok; `internal/service` and `pkg/agent` fail only under the full parallel run, pass on their own, and this branch touches neither (7 files, all under `integrations/wecom`). CI is green.

Bohan-J (approved):
Approving — thanks for sticking with this one, @seacen. It went four rounds, and each round the fix got narrower and better-founded rather than bigger, which is exactly how you want a concurrency change to converge.

Worth recording what the review actually surfaced, because none of it was visible from the original diff:

- A claim release whose result was unknown could leave a reply with no outcome at all. Closed by token-fenced claims with a compare-and-delete that is safe to retry.
- The same "an error means unknown, not failed" reasoning had to be applied one step later, to `Settle`. Closed by a bounded, idempotent retry.
- The retry then had to live inside the shutdown budget — and the deadline had to actually reach the wire, which took a dedicated Redis client, since go-redis discards context deadlines by default.

The last one is the kind of thing that would have been very hard to find in production: the deadline was present in the context, the code read correctly, and the round trip still waited out the socket timeout. Scoping the client rather than flipping the flag on the shared relay client was the right call.

I also appreciate that the comments explain the *reasoning* and not just the mechanics, and that the tests carry reverse-verification notes. The boundary that remains — a store that keeps losing settle responses can cost one monitoring record — is now stated in the code, the log and a test, instead of being implied. That is the right way to leave a deliberate trade.

CI is green. Merging.

PR #7954, opened 2026-09-02
feat(wecom): a one-frame receipt on ingest, over the aibot_respond_msg transport

seacen:
## What

The `aibot_respond_msg` transport plus a single `finish=true` receipt on ingest — "收到，正在处理。" — addressed by the callback's own `req_id`. The answer stays exactly where it is today, on the `aibot_send_msg` path. No stream store, no ledger, no in-place replacement.

This is the shape you specified on #6606. One commit, rebased on today's `main`.

The live bubble that used to ride along as a second commit has been **taken out of this PR**. You left it at "not never, but argue it from someone complaining about the second message" — so it stays on our branch until there is such a complaint, and this PR is only the receipt.

## Why a stream frame rather than a plain message

They cost the same: document 101463 says replies and pushes share one per-conversation budget of 30 per minute, and a single-frame receipt is one message under it. The reasons to prefer it are the `req_id` addressing inside the 24-hour reply window, and that it is the transport a bubble would need later.

## The transport, and what was measured

Every frame of one turn must echo the callback's `req_id`, and the server's ack carries nothing but that `req_id` — no stream id, no sequence. A verdict is therefore identifiable only by its position, and only while at most one frame of a `req_id` is on the wire. Against the live bot: 24 frames of one `req_id` written back-to-back were acked grouped by outcome, not in write order, and eleven of twelve concurrent refreshes to one stream were refused with `6000` ("data version conflict").

So the sender:

- keeps **one frame in flight per `req_id`** — a closing frame queues behind an unacked one rather than jumping it;
- counts verdicts in and frames out, so a late verdict for a frame whose caller gave up is never handed to the next frame (the misattribution you named);
- seals a stream once its closing frame is written, so a straggler never paints over it;
- resolves the socket **per frame** in the registry, because a `req_id` belongs to the turn and not to the connection it arrived on (a stream opened before a reconnect is finished after it — measured twice).

For the receipt itself none of this can lose anything: a receipt whose verdict is lost or misattributed costs the user a receipt. The transport is here because it is the part that has to be right before anything else is built on it, and a one-frame caller is the cheapest place to review it.

## Behaviour

- One receipt per turn; a burst that flushes into one run gets one (5-second coalescing per session, as DingTalk's ack notifier does).
- No `req_id` (an event callback) or no socket: nothing is sent, and the answer is unaffected.
- Wired through `NewResolverSet` as the WeCom `TypingNotifier`; the Router already treats it as optional.

## Tests

- Receipt shape and coalescing (`receipt_test.go`).
- Ack attribution (`stream_ack_test.go`): a late verdict is not handed to the next frame; a closing frame waits for an unacked one, and goes out once that frame's budget runs out; a frame after the seal never reaches the wire; a failed write gives back its place; the sweep keeps live turns even over its cap.
- Closing-frame content rules (`stream_frame_test.go`).
- The whole path through the **real connection loop** (`receipt_loop_test.go`): callback in over the socket, receipt out on the same socket with the callback's `req_id`, the server's anonymous ack routed back to the waiting frame.

Reverse-verified: with `routeResponse` reverted the loop test fails on a lost verdict; with coalescing removed the burst test fails on two receipts.

`go test -race ./internal/integrations/wecom ./internal/integrations/channel/engine ./cmd/server` green against a database migrated through 467.


seacen:
@Bohan-J Rebased on today's `main` and **cut down to the one commit you asked for** — the transport plus the single-frame receipt. The bubble that used to be commit 2 is off this PR entirely; it stays on our branch until there is a complaint about the second message to argue it from, which is where you left it.

The conflict that had appeared was between the bubble commit and our own #7952 (both wanted to own the failure notice). Removing the bubble removes it: the receipt commit applies clean to `3551e72e7`, and the only adjustment main needed was the fifth argument at `NewResolverSet`'s new call site in `channel_title_adapters_db_test.go`.

Green on `./internal/integrations/wecom ./internal/integrations/channel/engine ./cmd/server` against a database migrated through 467. No rush, but it has been sitting since the 2nd and this is the smallest it will ever be.

Bohan-J:
Closing this — we are taking the bubble direction instead, and the receipt is superseded by it.

To be clear about what happened here: **this PR is our design, not yours.** You opened #6606 with the bubble, we asked for the receipt instead, and you built exactly what we specified. Closing it now means the receipt work was spent on a shape we asked for and are no longer taking. That cost is ours, and we are not going to pretend otherwise.

## Why we changed our minds

The gate we set on #6606 was "argue the bubble from someone complaining about the second message". That gate was not a fair one, and it took us too long to see why: the measurements can only come from your deployment, so the complaint we were waiting for could only ever reach us through you. We were asking you to produce evidence that was independent of you, about a tenant that is yours. That is not a condition anyone can meet.

The substance also moved. Rotation is gone, so the message count is bounded at two and the ordinary case is one — never worse than the receipt. The ack-attribution fix in this PR's transport answered the objection we actually cared about on 2026-09-02. And the `channel/engine` objection we closed #8355 on is genuinely gone: the diff against the shared engine and the other four adapters is empty. Each time we named a reason, you removed it. At some point that stops being a coincidence and starts being an argument, and we should have registered that a round earlier.

## Where the work in this PR goes

The transport is not lost — `ws_sender.go`, `ws_frame.go`, `senders_registry.go` and the ack layer all live on in #8355, which carries them. What this PR uniquely holds is `receipt.go` and its tests, and that is the part the bubble replaces. Those two cannot both occupy `ResolverSet.Typing`.

We are reviewing #8355 now and will post that review shortly. It is not a rubber stamp — there are findings in the ack path and in the seal/fallback budget that we think are blocking, and some of them touch the transport this PR also carries, so they are worth your attention either way.

Thank you for building the receipt when we asked for it, and sorry for spending your time on a shape we then walked away from.

PR #7956, opened 2026-09-03
MUL-6987: docs: document task runtime environment contract

### LINKED ISSUE #7936: [Docs/Feature]: Document the runtime env injection contract (MULTICA_TOKEN/TASK_ID) + nested-runtime guidance
## Summary

While integrating Multica with an external normalization runner (bb, https://github.com/get-bb/bb), we verified the runtime env injection contract experimentally and found it works well — documenting it would help other integrators.

**Observed contract** (custom runtime profile, pi protocol family): the daemon injects into the runtime process:

- `MULTICA_TOKEN` (task-scoped token)
- `MULTICA_TASK_ID` / `MULTICA_TASK_CONFIG_ROOT` / `MULTICA_TASK_WORKSPACES_ROOT`
- `MULTICA_AGENT_ID` / `MULTICA_AGENT_NAME` / `MULTICA_WORKSPACE_ID`
- `MULTICA_SERVER_URL` / `MULTICA_DAEMON_PORT` / `MULTICA_TASK_SLOT`

Two behaviors worth documenting:

1. **Attribution**: comments posted by the agent's own `multica` CLI calls inside the task process are attributed to the assigned **agent** (not the human member) — we verified this is server-side attribution tied to the task context, not the token identity itself (the token's `/api/me` still returns the owning member).
2. **Nested runtimes**: when the custom runtime itself spawns *another* agent runtime (bb spawning a harness), the env must be forwarded explicitly by that layer. Currently there's no documented guidance for this nested case; integrators discover the contract only by reading process environments.

## Request

- Document the injection contract (which vars, when, attribution semantics) in the docs
- Optionally: guidance for nested-runtime scenarios (forward `MULTICA_TASK_ID`/`MULTICA_TOKEN` down), and whether an agent-scoped token (read/write limited to the assigned issue) is planned — that would make nested integrations safe by construction

Scorbunny2:
## What does this PR do?

Documents the environment contract that the daemon injects into agent task runtimes. The new section:

- lists the ten daemon-owned task context variables and their meanings
- explains server-side agent/task attribution and human-only endpoint behavior
- adds safe forwarding guidance for nested runtimes
- keeps the English, Chinese, Japanese, and Korean docs in sync

## Related Issue

Closes #7936

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Traced the documented values to `taskMulticaEnvironment` and the task launch path in `server/internal/daemon/daemon.go`.
- Documented which variables are injected, when they are available, and why custom agent environments cannot replace them.
- Clarified that the server derives authorship from the task token binding, regardless of request headers or the token owner's `/api/me` response.
- Added a warning for nested processes to forward only required variables and never persist `MULTICA_TOKEN`.

## How to Test

- Run the Fumadocs MDX generator.
- Run the docs TypeScript check with `tsc --noEmit`.
- Run `git diff --check`.

## Checklist

- [x] I have added or updated tests where applicable (documentation-only; no runtime tests required)
- [x] If this change affects the UI, I have included before/after screenshots (not applicable; documentation-only)

## AI Disclosure

**Prompt / approach:** Selected an unclaimed, low-complexity documentation issue; inspected the daemon injection and authentication source paths; documented the verified contract; synchronized the existing translations; and ran the repository's MDX and TypeScript documentation checks.

## Screenshots (optional)

Not applicable; documentation-only change.


Bohan-J (dismissed):
Thanks for picking this up — the underlying request is legitimate and we want this documented.

Before the specifics: you clearly read the source rather than guessing, and the hardest part of this PR is correct. I verified each of these against the code and they hold:

- **Attribution.** "The server, not request headers or the token owner's `/api/me` response, determines authorship" is exactly right. The auth middleware overwrites `X-Agent-ID` / `X-Task-ID` from the `mat_` token row and sets the server-only `X-Actor-Source: task_token`; `resolveActor` keys off that. And yes, `/api/me` does return the owning human — that surprises people, and it was worth writing down.
- **"Custom env cannot override these."** Correct, and stronger than you wrote: `isBlockedEnvKey` (`server/internal/daemon/daemon.go:9489`) rejects the entire `MULTICA_` prefix, not just these ten.
- **The zh translation** gets the `issue` / `task` distinction right, which is the thing most contributors miss on their first docs PR.

That said, there are five things to fix before this can merge. #1 is the important one.

---

## 1. The nested-runtime callout states the opposite of what actually happens

This is the blocking issue, because it is security guidance pointing the wrong way.

The callout currently says Multica "does not automatically inject these values into that second layer" and asks the integrator to forward what they need. In practice the reverse is true:

- `buildEnv` (`server/pkg/agent/claude.go:877`) is `mergeEnv(os.Environ(), extra)`, and every backend — including `pi.go:250`, the protocol family in the originating report — assigns `cmd.Env` from it.
- So these values land in the runtime process's **real OS environment**. Any process it forks inherits all of them by default, `MULTICA_TOKEN` included.

A reader following the current text draws two wrong conclusions: that they must forward the variables (they usually don't), and that not forwarding keeps a child process credential-free (it doesn't — the token is already there).

The genuinely useful fact runs in the opposite direction, and it is worth documenting: some tools filter their own subprocess environment. Codex's shell tool drops names containing `TOKEN` / `KEY` / `SECRET`, so that layer *loses* the credential, and the daemon writes a managed shell policy to put it back (`server/internal/daemon/execenv/codex_shell_env.go:16`).

The behavior you observed with `bb` — needing to forward explicitly — is `bb`'s own environment handling, not a Multica contract. Please don't generalize it into one.

Rather than have you re-derive this, here is the replacement text. Adapt the wording, keep the facts:

> These values live in the runtime process's real environment. Any child process your runtime starts inherits all of them by default, including `MULTICA_TOKEN` — if a child should not hold the credential, you must strip it explicitly; do not rely on isolation that isn't there. The exception runs the other way: tools that filter their own subprocess environment (Codex's shell tool drops names containing `TOKEN`, `KEY`, or `SECRET`) may need the variable explicitly allowed instead. Keep the token in process environments only — never in a prompt, log, repository file, or persistent configuration. A child process shares the parent task's identity and permissions; it does not receive a new, independently scoped identity.

Please also cross-link the token's lifecycle from [`auth-tokens.mdx`](apps/docs/content/docs/auth-tokens.mdx) (valid at most 24 hours, cleaned up when the task ends). Those two facts are the ones an integrator actually needs before deciding whether to hand the credential to a child process, and the section currently omits both.

## 2. Move the section to `daemon-runtimes.mdx`

`environment-variables.mdx:10` scopes itself explicitly, in all four languages: *"This page lists deployment-facing configuration only; test variables and internal task variables are not covered here."* This PR adds exactly those variables and leaves the sentence untouched.

Rather than amend that scope statement, please move the section to `daemon-runtimes.mdx` under **Custom runtime profiles** — that is where we already invite people to plug in their own executable, so it is where the contract belongs, and it puts the security guidance next to the integration instructions. Leave a one-line pointer in `environment-variables.mdx` so search still lands readers correctly.

This also resolves a side effect of the current placement: the new `###` splits `## Daemon configuration` in half, so the paragraph that follows it — ending in *"Precedence is command-line flag → environment variable → `~/.multica/config.json` → built-in default"* — now reads as if it applies to the injected variables, whose semantics are the exact opposite.

## 3. The list reads as exhaustive but isn't

`taskMulticaEnvironment` (`server/internal/daemon/daemon.go:163`) returns thirteen entries — the ten you list plus `TMPDIR`, `TMP`, and `TEMP` — and the launch path conditionally adds `MULTICA_AUTOPILOT_ID`, `MULTICA_AUTOPILOT_RUN_ID`, `MULTICA_QUICK_CREATE_TASK_ID`, `MULTICA_QUICK_CREATE_ATTACHMENT_IDS`, and `MULTICA_REPO_CHECKOUT_MODE`.

Please frame the list as what a custom runtime can rely on today, state plainly that it is not exhaustive and not a versioned API surface, and mark `MULTICA_TOKEN`, `MULTICA_TASK_ID`, `MULTICA_AGENT_ID`, `MULTICA_WORKSPACE_ID`, and `MULTICA_SERVER_URL` as the ones to build integrations on. The rest are informational and may change.

`TMPDIR` is worth adding explicitly — `MULTICA_AGENT_TEMP_BASE` is documented as its parent directory, so a reader arrives with that question already formed.

## 4. Link to the existing token section instead of restating it

`auth-tokens.mdx:78` ("Temporary tokens for agent runs") already covers the binding, the attribution outcome, and the human-only endpoint behavior. Two copies of the same explanation will drift. Keep one or two sentences here and link out for the rest.

## 5. The Japanese and Korean versions collapse `issue` and `task` into one word

`apps/docs/content/docs/developers/conventions.zh.mdx` sets the terminology for all localized docs, not only Chinese: `issue` is **タスク** / **태스크**, `task` is **作業** / **작업**, and the table states outright that no language may render the two as the same word.

On `main`, both files use 作業 / 작업 fifteen times each; the only lowercase `task` is inside the identifier `max_concurrent_tasks`. This PR introduces roughly eleven prose occurrences of bare `task`, which collide with the existing translation on adjacent lines.

One sentence is a substantive mistranslation rather than just a terminology slip. The English reads:

> Writes such as **issue** comments are attributed to the assigned agent and active **task**.

The Japanese renders both as `task`:

> **task** のコメントなどの書き込みは、割り当てられたエージェントと実行中の **task** に帰属します。

Two different entities, one word — precisely what the terminology table prohibits. Korean has the same problem. Your Chinese version already gets it right (「任务评论…归属到…当前 task」); use it as the model.

---

## Summary

Requesting changes on #1, #2, #3, #4, and #5. Directions for the two open questions from the original issue — where this lives, and how firm the list is — are answered in #2 and #3, so you shouldn't have to guess at either.

Worth noting that CI being green is not evidence here: this is a documentation-only change, and nothing in the pipeline checks factual accuracy against the daemon or terminology against the conventions doc. That verification is what this review is for.

Happy to take another look once these are addressed.

Scorbunny2:
@Bohan-J Addressed all five review points in d4927f17b:

- corrected the child-process environment inheritance and credential stripping guidance, including the Codex filtering exception
- moved the full contract under Custom runtime profiles and left search pointers in the environment-variable pages
- marked the five integration-contract variables, framed the list as non-exhaustive/non-versioned, and added TMPDIR/TMP/TEMP
- linked the existing temporary-token lifecycle and permissions documentation
- restored the issue/task distinction in Japanese and Korean

All CI checks on the new head are green. Ready for re-review.

multica-eve (approved):
Thank you, @Scorbunny2, for taking this on and for the careful follow-up. The revision now puts the contract in the right place, correctly explains child-process inheritance and token handling, distinguishes integration-contract variables from informational details, and keeps all four locales aligned. Approved — we appreciate the contribution!

PR #6919, opened 2026-08-13
MW-50: fix Antigravity token usage accounting

mikey-wang:
## What does this PR do?

Antigravity tasks now report provider token usage through Multica's existing task-usage pipeline. Previously, the adapter consumed plain text and always returned an empty usage map, so completed Gemini work appeared as 0 tokens and $0.00.

The adapter now consumes agy 1.1.12 `stream-json`, treats the final result as the authoritative aggregate, and keeps completed step usage as a fallback for failed or cancelled runs. The dashboard distinguishes recorded-but-unpriced usage from no usage across the KPI, trend, and leaderboard; mixed priced/unpriced totals render as `$X+`.

## Related Issue

Fixes MW-50

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [ ] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/antigravity.go`: parse structured output, session IDs, model attribution, terminal states, and token usage without double-counting step and result events.
- `server/pkg/agent/version.go`: require agy 1.1.12, the first supported version of this structured usage contract.
- `packages/views/dashboard/`: propagate an unpriced-cost state and render `Unpriced` or `$X+` instead of a misleading `$0.00`.
- Add unit, page, lifecycle, and opt-in real-provider smoke coverage.

## How to Test

1. Run `go vet ./pkg/agent` and `go test ./pkg/agent -run 'Antigravity|CheckMinVersion' -count=1` from `server/`.
2. Run `MULTICA_RUN_REAL_AGENT_SMOKE=1 go test -tags=agentintegration ./pkg/agent -run TestAntigravityRealStreamJSONUsage -count=1 -v` from `server/`.
3. Run the dashboard tests, type check, and lint for `@multica/views`.

Validation on this branch:

- Real agy smoke: `Gemini 3.6 Flash (High)`, 20,262 input tokens and 35 output tokens.
- Dashboard regression suite: 74 tests passed.
- Type check passed; lint completed with 0 errors (21 pre-existing warnings).

## Risks and Historical Data

- Model attribution depends on agy's selected-model log field. If that upstream field changes, token totals remain recorded but the model falls back to the configured value or `unknown`.
[...]

PR #7961, opened 2026-09-03
MUL-7000: fix(agent): record Antigravity token usage

mameikagou:
## What does this PR do?

Antigravity tasks now report provider token usage through Multica's existing task-usage pipeline. The adapter previously consumed plain text and always returned an empty usage map, so completed Gemini work appeared with no token usage.

The adapter now consumes `agy --output-format stream-json`. It sums deduplicated completed-step usage for the current execution and uses terminal `result.usage` only when no completed step is available. This matters for resumed conversations because agy's terminal result is cumulative across the conversation. It also distinguishes an interrupted partial reply from a reply that reached `agent_response=DONE` before agy reported a trailing network error. The trace is: agy stream-json -> `Result.Usage` and terminal state -> the daemon's existing task result and usage persistence.

This is a focused follow-up to #6919. That PR was closed without merging after referencing #6654, but #6654 addressed Antigravity model-catalog parsing only; current `main` still returns an empty usage map from the Antigravity adapter.

## Related Issue

Follow-up to #6919.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/antigravity.go`: parse structured deltas, session IDs, terminal states, selected-model attribution, and token usage.
- Prefer completed steps keyed by `step_index` for current-execution usage; fall back to terminal usage when no completed step exists.
- Map input, output, cache-read, and cache-write buckets directly; thinking remains included in output.
- Prefer a structured provider error over the less useful non-zero process exit status.
- For Antigravity's exact trailing network error, keep a non-empty reply as completed only when an `agent_response` step reached `DONE`. A stale ACTIVE tool step no longer vetoes a completed response because agy can cancel background work through a separate step without closing the original step index.
- Require agy 1.1.10, where headless `--model` selection is reliable, and update the four localized runtime-install docs.
- Add success, non-zero failure, cancellation, direct cache mapping, model attribution, version, trailing-network, stale-background-step, partial-response, and opt-in real two-turn resume coverage.

## How to Test

1. `go test ./pkg/agent -run 'Antigravity|CheckMinVersion' -count=1`
2. `go vet ./pkg/agent`
3. `go test ./pkg/agent -count=1`
4. `MULTICA_RUN_REAL_AGENT_SMOKE=1 go test -tags=agentintegration ./pkg/agent -run TestAntigravityRealResumedUsageIsPerExecution -count=1 -v`
5. `go build -buildvcs=false ./cmd/multica`

Validated against agy 1.1.25. A real resumed two-execution smoke used the same conversation ID and attributed current-turn usage to `Gemini 3.8 Flash (High)`: first execution 13,979 input / 36 output, second execution 14,227 input / 60 output. Both runs reported cache read/write as zero directly from the provider.

Raw three-turn wire verification established the accounting scope: terminal totals grew from 13,909/59 to 28,089/96; on the third execution, the DONE step was 14,429/37 and the terminal became 42,518/133, exactly the prior terminal plus the current step. Therefore terminal usage is cumulative and completed-step usage is per execution.

## Incident evidence and completion boundary

A real Antigravity 1.1.26 incident on 2026-09-05 showed a coherent, complete reply in chat followed by:

```text
There was a network issue connecting to the server, please try again.
```

A representative daemon result had `status=failed`, `duration=1m30s`, `output_bytes=2171`, a conversation ID, and one model usage entry before the task was blocked as `agent_error.unknown`. Across the same connectivity episode there were 11 records with this exact error; 10 already contained 260-2638 output bytes.

A second real incident on agy 1.1.27 on 2026-09-07 exposed a missing completion branch. The daemon recorded `status=failed`, `output_bytes=2119`, a valid conversation ID, and one model usage entry even though the durable transcript ended with a non-empty `PLANNER_RESPONSE / MODEL / DONE`. Earlier in that turn, background command step 2120 reported RUNNING; the agent then cancelled it through `manage_task` steps 2121-2124 and received a cancellation system message, but agy never emitted a same-index DONE update for step 2120. Treating every historical ACTIVE step as a completion gate therefore turned a complete reply into a false failure.

Output length alone is not used as proof of completion. The narrow recovery now requires all of the following:

- provider error exactly matches the Antigravity network message above;
- terminal `result.response` is non-empty;
- the structured `agent_response` step reached `DONE`.

If the network drops during generation, the response step remains `ACTIVE`, so Multica still returns `failed` with the partial output and the provider error. Cancellation and timeout also continue to win. A stale ACTIVE tool step is not a reliable sign of an incomplete response because agy may cancel background work through a separate `manage_task` step instead of closing the original step index. This follows Antigravity's documented [ACTIVE/DONE response contract](https://antigravity.google/docs/cli/headless/). Google's own [complete-response/error report](https://github.com/google-antigravity/antigravity-cli/issues/848) independently shows why a terminal `ERROR` can describe post-response work rather than an incomplete answer.

The regression fixtures cover all three boundaries:

- `DONE response + exact network error + exit 1` becomes completed;
- `stale ACTIVE tool + separately completed cancellation + DONE response + exact network error` also becomes completed;
- `ACTIVE response + partial output + the same error + exit 1` remains failed.

## Risks and historical data

- Runtimes below agy 1.1.10 no longer register because earlier releases do not provide both the required structured contract and reliable headless model selection.
- Historical Antigravity tasks cannot be backfilled exactly because the old adapter discarded the provider counters.
- Model attribution prefers the stream, then agy's selected-model log, then the configured model, and finally `unknown`.
- Trailing-error recovery is intentionally limited to Antigravity's exact network message, a non-empty canonical result response, and a DONE agent-response step; other provider errors and incomplete responses are not suppressed.

## Checklist

- [x] This backend-only change does not require UI screenshots
- [x] This does not add a new runtime, coding tool, or UI tab
- [x] I checked the Chinese documentation change against `apps/docs/content/docs/developers/conventions.zh.mdx`

## AI Disclosure

**AI tool used:** Multica Agent (Codex)

**Prompt / approach:** Reproduced the missing usage against agy 1.1.25, established resumed-session accounting from raw provider events, and correlated two real trailing-network failures with daemon and durable transcript state. The 1.1.27 follow-up specifically compared stream step lifecycle with the separately completed background-task cancellation and added a regression fixture for the stale ACTIVE index without weakening the partial-response boundary.

## Screenshots (optional)

Not attached because the incident screenshot contains unrelated private conversation content; the exact provider error and sanitized daemon/transcript evidence are included above.


multica-eve (changes requested):
Before this change, people using Antigravity can see completed replies but cannot see how much each run consumed. After it, those runs contribute to the existing usage views, while people on unsupported older installations must upgrade before continuing to use Antigravity. This is a product problem worth solving, and the proposed user journey is the right one.

**Product verdict: accept the direction.** The current implementation has four merge blockers:

1. **`server/pkg/agent/antigravity.go:92-95` undercounts ordinary input whenever cache reads are smaller than input.**

   Google's [official structured-output examples](https://antigravity.google/docs/cli/headless/) include a turn with `input_tokens=278`, `cache_read_tokens=30214`, `output_tokens=4`, and `total_tokens=282`. This shows that cache reads are a separate bucket and that `total_tokens == input_tokens + output_tokens` does not prove cache is included in input. The current condition treats that equality as proof of inclusion. For the official one-turn example `10418 / 589 / 8113 / 11007`, it would reduce ordinary input from `10418` to `2305`. Please map the fields directly without changing their meaning based on relative magnitude, and update `server/pkg/agent/antigravity_test.go:875-901`, which currently locks in the incorrect assumption.

2. **`server/pkg/agent/antigravity.go:314-316` discards the structured provider error on the documented failure path.**

   Antigravity documents that a failed run exits non-zero and also supplies `status` and `error` in the terminal result. Because `cmd.Wait()` is handled first, an ordinary `exit 1 + result.error` becomes only `agy exited with error: exit status 1`; the later `streamResultError` branch cannot run. The new failure fixture exits zero, so it does not cover the provider contract. Please prefer the terminal structured error when present, use the process error only as a fallback, and add a regression case whose fixture exits non-zero.

3. **`server/pkg/agent/version.go:14` should gate this contract at `1.1.10`, not `1.1.8`.**

   The [Antigravity 1.1.10 release notes](https://github.com/google-antigravity/antigravity-cli/releases/tag/1.1.10) state that earlier versions ignored `--model` in headless runs and silently used the persisted/default model. This PR then attributes usage using `init.model`, the log, or the configured value. Accepting 1.1.8-1.1.9 can therefore run one model while recording usage under another, affecting model-level pricing. Please raise the floor and align the version test and all four localized installation docs, or provide reproducible evidence that these versions execute and attribute an explicitly selected model correctly.

4. **The accounting scope for resumed conversations is unverified and may double-count prior turns.**

   Multica commonly resumes with `--conversation`, but `server/pkg/agent/antigravity_integration_test.go:16-64` exercises only the first turn. Antigravity also documents cumulative `usage` for multi-turn structured sessions. Please add an opt-in real two-execution check where the second call uses the first call's session ID, and establish whether its terminal usage is per-turn or cumulative. If it is cumulative, report a reliable current-turn delta rather than persisting the aggregate again.

The choice to reuse the existing `Result.Usage` persistence path is good, and preferring the terminal result with a deduplicated completed-step fallback is otherwise aligned with the documented event shape.

Local validation on `f86fde36`: targeted Antigravity/version tests, `go vet ./pkg/agent`, the full `go test ./pkg/agent -count=1`, and `git diff --check` all pass. I did not rerun the authenticated provider smoke test because it consumes account quota; the reported smoke covers only the first turn.

mameikagou:
Updated in `61d869a68`; all four blockers are addressed:

1. Provider usage buckets are now mapped directly. The regression table includes both official shapes (`278/4/30214/282` and `10418/589/8113/11007`).
2. A structured terminal error now takes precedence over `cmd.Wait()`, and the failure fixture emits `result.error` then exits 1.
3. The Antigravity minimum is now 1.1.10 in code, tests, and all four localized docs.
4. I verified resumed accounting against agy 1.1.25. Raw terminal usage is cumulative: turn 1 was `13909/59`, turn 2 was `28089/96`; on turn 3 the DONE step was `14429/37` and terminal usage became `42518/133`, exactly prior terminal plus the current step. The adapter therefore prefers deduplicated DONE-step usage for the current execution and uses terminal usage only as a fallback.

I also changed the opt-in provider test to make two executions with the second resuming the first session ID. It passed against `Gemini 3.8 Flash (High)` with per-execution usage `13979/36` and `14227/60`; cache was provider-reported zero on both turns.

Local validation: targeted tests, `go vet ./pkg/agent`, full `go test ./pkg/agent -count=1`, `git diff --check`, and the authenticated two-turn smoke all pass. Please re-review when convenient.

mameikagou:
Follow-up pushed in `681ae11bd` after merging current `upstream/main`.

This addresses a real Antigravity 1.1.26 failure mode observed on 2026-09-05:

```text
There was a network issue connecting to the server, please try again.
```

A representative daemon result had `status=failed`, `duration=1m30s`, `output_bytes=2171`, a conversation ID, and one model usage entry before the task was blocked as `agent_error.unknown`. The same connectivity episode produced 11 records with this exact error; 10 already had 260-2638 output bytes.

The fix does **not** infer success from non-empty output. It is Antigravity-only and requires the exact network error, a non-empty terminal `result.response`, `agent_response=DONE`, and zero remaining `ACTIVE` steps. A mid-generation disconnect therefore remains failed with partial output; cancellation, timeout, and all other provider errors are unchanged.

Paired fake-provider regressions prove both boundaries:

- `DONE + complete response + exact network error + exit 1` -> completed, answer preserved
- `ACTIVE + partial response + same error + exit 1` -> failed, partial answer and error preserved

Local validation after merging current main:

- targeted complete/partial/usage tests: pass
- `go test ./pkg/agent -count=1`: pass (98.0s)
- `go vet ./pkg/agent`: pass
- `git diff --check`: pass

I did not rerun the authenticated smoke test because this follow-up is deterministic terminal-state classification and a real run consumes provider quota. The earlier resumed-session smoke evidence remains in the PR body. Please re-review when convenient.

mameikagou:
Follow-up pushed in `8a55f080b` for the newly observed agy 1.1.27 false-failure branch.

This was not a binary rollback. The affected daemon was already running the prior fix, but the real turn exposed a lifecycle shape absent from the original fixture:

- daemon result: `status=failed`, `output_bytes=2119`, valid conversation ID, one model usage entry;
- durable transcript: final non-empty `PLANNER_RESPONSE / MODEL / DONE`;
- background command step 2120 reported RUNNING;
- `manage_task` then killed it through separate steps 2121-2124 and agy delivered the cancellation system message;
- agy never emitted DONE for the original step-2120 index.

The old `activeSteps == 0` gate therefore treated a cancelled background command as unfinished work even after `agent_response=DONE` and a complete terminal response.

The recovery boundary now requires the exact Antigravity network error, a non-empty canonical `result.response`, and `agent_response=DONE`. It no longer treats stale tool-step bookkeeping as proof that the response is partial. `agent_response=ACTIVE` still remains failed, so a mid-generation disconnect is not hidden.

New regression coverage reproduces the stale ACTIVE tool index plus separately completed cancellation and complete response. Validation:

```text
go test -race ./pkg/agent -run 'TestAntigravityBackend(IgnoresTrailingNetworkErrorAfterDoneResponse|IgnoresStaleActiveStepAfterDoneResponse|KeepsNetworkFailureForPartialResponse)$' -count=20
go test ./pkg/agent -run 'Antigravity|CheckMinVersion' -count=1
go vet ./pkg/agent
git diff --check
```

All pass locally. The screenshot is omitted because it contains unrelated private conversation content; the evidence above is sanitized from the daemon log and agy transcript.

multica-eve (changes requested):
Before this change, people could receive a finished answer yet see no usage for the run, and a connection problem after the answer could make finished work appear failed. After it, usage appears in the existing views and genuinely finished answers are preserved, while older installations must be upgraded. These outcomes are worth shipping, but the latest recovery rule can still mark an unfinished answer as complete.

**Product verdict: accept the direction.** The missing-usage problem is present on `main`, the result closes the loop through the existing usage views, and reusing the existing accounting path is how I would solve it. The additional completed-answer recovery is also a real user problem, independently reflected in upstream [issue #848](https://github.com/google-antigravity/antigravity-cli/issues/848), but its completion boundary must remain fail-closed.

The four blockers from my previous review are resolved at `8a55f080b531`: provider usage buckets are mapped directly; a structured provider error wins over a non-zero process exit; the minimum version and localized docs now use 1.1.10; and completed-step usage is preferred over cumulative terminal usage, with a two-turn opt-in smoke test and raw multi-turn evidence. The official headless documentation also confirms that terminal usage is cumulative while step usage is per-step.

## Must fix

**`server/pkg/agent/antigravity.go:240,263-265,330` remembers whether _any_ response step reached `DONE`, not whether the latest response step finished.**

`streamAgentResponseDone` is a sticky boolean. A normal tool-using turn may contain more than one response step, so this valid sequence is misclassified:

```text
agent_response step 2: DONE
tool step 3: DONE
agent_response step 4: ACTIVE with partial text
result: ERROR + exact network message + non-empty partial response
```

After step 2 the boolean remains true. The terminal error therefore enters `antigravityCompletedDespiteTrailingNetworkError` and returns `completed`, suppressing both the provider error and non-zero exit even though the latest response never reached `DONE`. The current partial-response regression only has a single `ACTIVE` response, so it does not exercise this branch. The upstream stream contract explicitly uses `ACTIVE` fragments followed by the final `DONE`; an earlier completed response step cannot prove a later one finished.

Please track response state by step index (or otherwise bind the terminal response to its final response step) and require the latest/current response step to be `DONE`. Add the sequence above as a regression. This still permits the `8a55f080b` incident shape—stale `ACTIVE` tool bookkeeping with a final `DONE` response—without converting a genuinely partial final response into success.

The exact network-message match is narrow and fail-closed when the wording changes, so it is acceptable as a provider-specific workaround while upstream exposes no structured cleanup-error cause. The response-completion signal itself must be exact before overriding an explicit `ERROR` result.

Validation on `8a55f080b531`:

- targeted Antigravity/version tests: pass
- the three trailing-network tests under `-race -count=20`: pass
- `go vet ./pkg/agent`: pass
- `go build -buildvcs=false ./cmd/multica`: pass
- `git diff --check`: pass
- full `go test ./pkg/agent -count=1`: one initial run failed after 368.689s with the specific failure lost in truncated parallel logs; an immediate structured-output rerun completed cleanly
- current GitHub CI: all reported checks pass

If this remains unfixed, a transient connection failure after intermediate narration can be recorded as a successful task with a truncated answer; rollback stops future misclassification but cannot repair already-persisted task state automatically. It affects Antigravity users on the new default structured-output path.

**Merge recommendation: Request changes** — the original four blockers are fixed, but completion recovery must key off the final response step before this can merge.

mameikagou:
Addressed the remaining review blocker in `6b5958eb1`.

The completion guard no longer uses a sticky "some response was DONE" boolean. It now tracks the greatest `agent_response` step index observed and derives completion only from that step's latest state. A late update for an older response step cannot overwrite the newer response state.

The regression reproduces the exact counterexample from the review:

1. response step 2: `DONE`
2. tool step 3: `DONE`
3. response step 4: `ACTIVE` with partial text
4. terminal `ERROR` with the exact known network message and a non-empty partial response

It now remains `failed`, preserves the partial response, and keeps the provider error. The existing completed-response and stale-active-tool cases still complete successfully.

Validation:

- focused `go test -race` for all four trailing-network response-state cases, 20 repetitions
- `go vet ./pkg/agent`
- Windows amd64 package cross-compilation
- `git diff --check`

multica-eve (approved):
**Merge — the latest-response completion guard closes the only remaining false-success path, and the Antigravity usage accounting is now correct and sufficiently covered.**

Product verdict: ✅ Accept. Antigravity runs now persist provider-reported token usage, resumed runs avoid double-counting earlier turns, and terminal behavior remains honest: a fully completed answer survives the known trailing transport error, while a newer partial answer is still reported as failed.

The new maximum-step tracking fixes the sticky completion flag from my previous review. A newer `agent_response` step replaces the completion state of an earlier response, while late updates for older response steps cannot overwrite it. The added regression reproduces the exact `DONE response → tool → newer ACTIVE response → network error` sequence and correctly keeps it failed; the completed-response and stale-tool cases remain covered.

The earlier review findings are also resolved: provider token buckets are mapped directly, per-step usage avoids cumulative resumed-session totals, structured provider errors take precedence over process exit status, and the minimum supported Antigravity version is 1.1.10.

Validation on `6b5958eb1743d61a096e83d320e8416f95a411dd`:

- targeted Antigravity/version tests
- the four trailing-network-error regressions under `-race -count=20`
- `go vet ./pkg/agent`
- `go build -buildvcs=false ./cmd/multica`

All passed locally, and the GitHub CI suite is green. I did not rerun the authenticated real-provider smoke test because it consumes external provider quota; the protocol fixtures cover the changed state machine deterministically.

PR #7963, opened 2026-09-03
feat(agent): add Command Code as a native runtime

foXaCe:
## What does this PR do?

Adds **Command Code** (`commandcode`, npm `command-code`) as a native runtime.

Command Code is a Node coding CLI with a headless mode. Today it can only reach Multica indirectly — as an OpenAI-compatible provider configured inside OpenCode's `provider` block — which means its own session handling, model catalog and tool events are invisible to the daemon. You get the models, but not the agent.

**Why a new protocol family rather than a `BuiltinRuntimes` descriptor.** `builtin_runtimes.go` says adding a compatible fork is a descriptor entry, and that was the first thing I checked. Command Code does not qualify: it speaks neither Claude Code's `stream-json`, nor pi's `--mode json`, nor ACP. `commandcode -p --output-format json` emits NDJSON where every line is one of two envelopes:

```json
{"type":"event","event":{"type":"run_start","sessionId":"…"}}
{"type":"event","event":{"type":"model_request_start","model":"…"}}
{"type":"result","subtype":"success","sessionId":"…","usage":{…},"durationMs":1834,"finalText":"…"}
```

The signatures `run_start`, `finalText` and `model_request_start` appear in none of the 26 existing families, and `applyBuiltinRuntimeOverrides` is implemented only by `piBackend`, so no descriptor could host it.

**Notable protocol traits, and how the backend handles them:**

- **The prompt goes on stdin.** `-p` with no positional makes Command Code read the run message from stdin. That keeps a large prompt off argv — the Windows `CreateProcess` command-line cap is reachable with a prompt carrying workspace skills — and keeps it out of process listings.
- **Assistant text does not stream token by token.** It arrives whole at `message_end.content`, as Anthropic-shaped blocks. A turn's prose therefore surfaces when the turn closes. Tool activity *does* stream, through `tool_running` before execution, so tool calls are not held back until the turn ends.
- **Usage is per model.** `model_request_end` names the model that incurred it, so a run that switches models bills each one separately rather than folding everything under the configured model.
- **The resume pointer is pinned at `run_start`**, so a run killed before its result line still leaves a session id the daemon can resume from.
- **The result line is the only proof the run finished.** A stream that simply stops is not a success — that is what the prompt-write diagnosis keys on, following the reasoning already in the OpenCode backend.
- **Exit code 8 is a bounded stop**, not a crash: that is how Command Code reports reaching the `--max-turns` cap the daemon itself set.

Model discovery parses `commandcode --list-models`. The command has no JSON mode, so the aligned catalog table is the only source; section headers and the trailing help block fall out because neither matches the provider-qualified id shape.

The runtime brief reaches the agent through **AGENTS.md**, which the Command Code bundle reads.

## Related Issue

No existing issue — happy to open one if you would rather discuss the approach before reviewing the code.

## Type of Change

## Changes Made

**Backend**
- `server/pkg/agent/commandcode.go` — the protocol family: process lifecycle, NDJSON scanner, event→message mapping, model discovery.
- `server/pkg/agent/commandcode_test.go` — 14 tests over captured transcript shapes.
- `server/pkg/agent/agent.go` — `SupportedTypes`, `New()` dispatch, launch header.
- `server/pkg/agent/launch.go`, `models.go` — blocked-args registry, `ListModels` dispatch.
- `server/pkg/agent/agent_supported_types_test.go` — whitelist kept in lockstep.

**Migration**
- `server/migrations/450_runtime_profile_add_commandcode.{up,down}.sql` — extends the `runtime_profile.protocol_family` CHECK, `NOT VALID`, matching the shape of 441.

**Daemon**
- `agents_probe.go` — probes `MULTICA_COMMANDCODE_PATH` / `MULTICA_COMMANDCODE_MODEL`.
- `config.go`, `daemon.go` — default command list, error text, display name.
- `execenv/runtime_config.go` — AGENTS.md as the context file.
- `internal/metrics/labels.go` — metric label.

**Frontend**
- `packages/core/types/agent.ts`, `packages/core/runtimes/display.ts` — protocol family and display name.

**Landing copy and docs** (per the checklist)
- `apps/web/features/landing/i18n/{en,ja,ko,zh}.ts` — added to all runtime lists, counter 26 → 27.
- `apps/docs/content/docs/environment-variables{,.ja,.ko,.zh}.mdx` — the two environment variables.

## How to Test

```bash
# Unit tests over the captured protocol fixtures
cd server && go test ./pkg/agent/ -run "Commandcode|CommandCode" -count=1 -v

# Wiring guards
go test ./pkg/agent/ -run "SupportedTypes|LaunchHeader" -count=1

# End to end, with Command Code installed and logged in
npm i -g command-code && commandcode login
multica daemon start     # the probe registers a "Command Code" runtime
# create an agent on it, assign an issue
```

Model discovery can be checked without an account balance:

```bash
commandcode --list-models
```

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — no bespoke UI; the runtime uses the existing runtime cards
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** and **relevant docs**
- [ ] If this PR touches Chinese product copy — only the runtime name (an untranslated brand) and the two environment variables were added to the existing zh sentences

## Risks and limits, stated plainly

**The protocol was reconstructed, and one path is not exercised end to end.** My Command Code account ran out of credits partway through, so every run I could capture ends at `run_error: insufficient credits` rather than a completed turn. The event vocabulary, the envelope shapes, the `buildPrintResultLine` field set and the content-block shapes were therefore read out of the shipped `dist/cli.mjs` of Command Code 1.44.0, and the tests replay those shapes as fixtures. Captured live and asserted: `run_start`, `turn_start`, `message_start`, `model_request_start`, `model_trace`, `run_error`, the result line, and the full `--list-models` catalog. Reconstructed from the bundle and covered only by fixtures: `message_end` content blocks, `tool_running` / `tool_completed` / `tool_errored` / `tool_denied`, and `model_request_end` usage. If a reviewer with a funded account runs one real task, that closes the gap — I would rather say this than let it be found in review.

**No logo.** The runtime falls back to the generic monitor icon rather than shipping a third-party mark I would have had to invent.

**No ticket number.** External contribution, so the commit carries no `MUL-` reference.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** The owner of this checkout asked for Command Code to be supported natively. The work was done by reading the existing backends for the house pattern (the OpenCode backend is the closest analogue and its cancellation and prompt-write reasoning is followed deliberately), measuring the real CLI where an account balance allowed, and reverse-engineering the remaining protocol from the shipped bundle. Every claim above about the protocol is either a captured transcript or a quoted function from `dist/cli.mjs`; where I could not verify something by running it, it is listed under Risks rather than asserted.


foXaCe:
Pushed 5f1b4b7a8 — a parsing bug in this PR, found while running the runtime against a current Command Code.

`commandcodeModelLine` required a provider-qualified id (`^(\S+/\S+)\s{2,}…`). Only the open-source sections of the catalog use that shape; Anthropic and OpenAI list bare ids:

```
claude-sonnet-5    best combo of speed & intelligence (recommended)
gpt-6-astra        most capable OpenAI model for demanding reasoning & agents
```

So all 16 of their models were silently dropped. Command Code 1.54.2 advertises 70; this PR enumerated 54.

The pattern is now `^(\S*[^\s:])\s{2,}(\S.*)$` — no slash required, and the trailing `Docs:  https://…` line is excluded by the id not ending on a colon, which still admits suffixed ids like `meituan/longcat-2.0:free`. Section headers are single-spaced and never reach the two-space column, so they keep falling out on their own.

Worth naming the root cause: the test fixture was mine and it was wrong. It wrote Anthropic as `anthropic/claude-fable-5-1`, a shape the CLI does not emit, so the test passed by confirming a false premise about the very format it was meant to pin. The fixture is now the real output of 1.54.2, carrying both id shapes, the colon suffix and the `Docs:` line. Verified against a live catalog: 70 rows parsed out of 70 advertised, nothing else caught.

PR #7964, opened 2026-09-03
feat(agents): support Pi runtime MCP settings

### LINKED ISSUE #7957: [Feature]: pi runtime support mcp UI setting and use with issue
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

like codex and opencode cli runtime mcp setting in UI

_No response_

opencode runtime mcp UI setting

Image

pi runtime no mcp setting

Image

feiguoL:
## What does this PR do?

Adds real task-scoped MCP support to the Pi runtime through a daemon-generated Pi extension. Pi does not natively read `.pi/mcp.json`, so the extension is loaded explicitly with `--extension`, performs MCP initialization and tool discovery, registers the discovered tools with Pi, and forwards tool calls to the configured stdio or HTTP MCP server.

## Related Issue

Closes #7957

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Add `pi` to the shared MCP-supported provider list and MCP tab visibility coverage.
- Generate task-local `.pi/mcp.json` and `.pi/extensions/multica-mcp.js` sidecars with restrictive permissions and manifest cleanup.
- Pass the task extension to Pi as a daemon-owned `--extension` argument; block custom args from replacing it.
- Implement MCP `initialize`, `tools/list`, and `tools/call` in the extension for stdio and HTTP-style servers, including configured env and headers.
- Refresh the extension on task reuse without changing the existing OMP MCP path.

## How to Test

1. Configure a Pi agent with an MCP server in the agent MCP tab.
2. Run an issue with that agent and verify the task starts Pi with `--extension .pi/extensions/multica-mcp.js`.
3. Verify the extension performs MCP initialization and tool discovery, and that a discovered tool call reaches the configured MCP server.
4. Run `go test ./pkg/agent -run 'TestBuildPiArgsIncludesTaskMcpExtension|TestBuildPiArgsOwnsTaskMcpExtensionFlag' -count=1`.
5. Run `go test ./internal/daemon/execenv -run 'TestPrepareOmpMcpConfig|TestPreparePiMcpExtension' -count=1`, `pnpm exec vitest run packages/core/agents/mcp-support.test.ts`, and `pnpm typecheck`.

## Checklist

## AI Disclosure

**AI tool used:** Codex CLI

**Prompt / approach:**
Reviewed the upstream Pi behavior and existing OMP MCP path, reproduced the missing runtime-consumption gap, added regression coverage first, then implemented a task-isolated Pi extension that performs the MCP protocol handshake, discovers tools, and routes calls to configured servers.

## Risks

- Pi has intentionally left MCP outside its core, so this extension is maintained against Pi's public `ExtensionAPI` and MCP JSON-RPC interfaces.
- Managed MCP configuration remains task-local and is never placed on the command line; only the extension path is passed to Pi.

## Screenshots (optional)

Not included; the existing MCP tab layout is unchanged.

### DIFF
--- a/packages/core/agents/mcp-support.test.ts
+++ b/packages/core/agents/mcp-support.test.ts
@@ -5,14 +5,11 @@ import { providerSupportsMcpConfig } from "./mcp-support";
 describe("providerSupportsMcpConfig", () => {
   it("accepts a provider whose runtime consumes mcp_config", () => {
     expect(providerSupportsMcpConfig("claude")).toBe(true);
+    expect(providerSupportsMcpConfig("pi")).toBe(true);
   });
   it("rejects providers whose runtime ignores mcp_config", () => {
     expect(providerSupportsMcpConfig("antigravity")).toBe(false);
     expect(providerSupportsMcpConfig("copilot")).toBe(false);
-    // Pi ships without MCP by design: upstream's README states "No MCP." and
-    // directs users to extensions instead, so there is no config file Multica
-    // could write that pi would read. Only its omp fork consumes mcp_config.
-    expect(providerSupportsMcpConfig("pi")).toBe(false);
     // ZeroClaw's ACP server never reads `params.mcpServers` — MCP lives in
     // ZeroClaw's own config-dir, so a value saved here could not be honoured.
     expect(providerSupportsMcpConfig("zeroclaw")).toBe(false);
--- a/packages/core/agents/mcp-support.ts
+++ b/packages/core/agents/mcp-support.ts
@@ -26,6 +26,7 @@ const MCP_SUPPORTED_PROVIDERS = new Set([
   "mcode",
   "traecli",
   "dim",
+  "pi",
   "omp",
 ]);
 
--- a/packages/views/agents/components/agent-overview-pane.test.tsx
+++ b/packages/views/agents/components/agent-overview-pane.test.tsx
@@ -190,6 +190,7 @@ describe("AgentOverviewPane MCP tab visibility", () => {
     ["Kiro", "kiro"],
     ["OpenCode", "opencode"],
     ["OpenClaw", "openclaw"],
+    ["Pi", "pi"],
     ["Oh My Pi", "omp"],
   ])("renders the MCP tab when the agent runs on the %s runtime", (_label, provider) => {
     renderPane([makeRuntime(provider)]);
--- a/server/internal/daemon/execenv/execenv.go
+++ b/server/internal/daemon/execenv/execenv.go
@@ -61,7 +61,7 @@ type PrepareParams struct {
 	OpenclawBin  string // resolved openclaw CLI path (only used when Provider == "openclaw"); empty = look up on PATH
 	// McpConfig is the agent's saved `mcp_config` JSON, forwarded to the
 	// provider-specific config preparer when that provider materialises MCP
-	// via a per-task config file. Cursor, OpenClaw, and OMP consume it here;
+	// via a per-task config file. Cursor, OpenClaw, Pi, and OMP consume it here;
 	// other providers wire MCP via ExecOptions.McpConfig in the agent backend.
 	McpConfig json.RawMessage
 	// CursorMcpAuthSource is an explicit opt-in path to a Cursor mcp-auth.json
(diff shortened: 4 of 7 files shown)


Bohan-J:
Thanks for putting this together — and for #7565 before it, which is what makes this a reasonable thing to try. Unfortunately this one can't work, and the reason is upstream of the code: **it writes a file that pi never opens.**

## The check

Same temp directory, same deliberately corrupt file, pi 0.73.1:

```sh
# A — corrupt .pi/mcp.json
T=$(mktemp -d); mkdir -p "$T/.pi"
echo 'NOT JSON {{{' > "$T/.pi/mcp.json"
cd "$T" && pi --offline --no-session --no-extensions -p "hi" </dev/null
# → total silence about the file; startup runs all the way to "No API key found"

# B — corrupt .pi/settings.json
T=$(mktemp -d); mkdir -p "$T/.pi"
echo 'NOT JSON {{{' > "$T/.pi/settings.json"
cd "$T" && pi --offline --no-session --no-extensions -p "hi" </dev/null
# → Warning: (runtime creation, project settings) Unexpected token 'T', "NOT JSON "... is not valid JSON
```

B proves the harness is sound — pi does surface a broken file in `.pi/` when it reads it. A is silent because `.pi/mcp.json` is not in its read path at all.

The static side agrees: the string `mcp.json` occurs **zero** times in pi's published `dist/`, and the only two `mcp` matches in the whole package tree are noise (an Anthropic OAuth scope in `@mariozechner/pi-ai`, and `multiset`/`multimap` in a vendored highlight.js keyword list). Under `.pi/` pi uses `settings.json`, `agent/sessions/`, and `skills/`. `~/.pi/agent/mcp.json` — the inventory path here — doesn't exist either; `.pi/agent/` is where sessions live, so that list will always be empty while the runtime reports `mcp_supported: true`.

## Why the omp path doesn't transfer

omp is a fork that **added** MCP; `.omp/mcp.json` is a real, highest-precedence project path, which is why #7565 worked. Reusing the same writer for pi assumes the upstream has the capability its fork added. It doesn't — that is exactly why #7565 covered `omp` only, and why `mcp-support.test.ts` carried a negative assertion for `pi` plus a comment recording the reason. This PR removes both, which is the guard that was meant to catch this shape of change.

## Green CI doesn't cover the claim

Every test here asserts that *Multica wrote a file*, never that *pi consumed a server* — and none can, because pi won't. Note also that `TestPreparePiMcpConfigCleanupRemovesManagedFiles` still passes `"omp"`, so despite the name it doesn't exercise pi. The PR's own How-to-Test step 2, "verify Pi can use the configured server", is the one step that cannot pass.

## One change here would regress omp

```go
case "pi", "omp":
    return map[string]any{}, true, nil
```

This flips `supported` to `true` with an empty server map. In `mergeRuntimeAndAgentMcpConfig`, `!supported` returns the agent config **verbatim**, while `supported` unmarshals it, keeps only the `mcpServers` sub-map, and re-marshals. That result is `effectiveMcpConfig`, which feeds both `PrepareParams.McpConfig` (daemon.go:7506/7556) and `ExecOptions.McpConfig` (8032/8079).

So for omp — a shipped, working feature — a saved `mcp_config` would stop being written byte-for-byte to `.omp/mcp.json` and get normalised instead, silently dropping any top-level key other than `mcpServers`. The case contributes nothing (the map is empty), so it is pure downside. And it wouldn't be caught: `TestPrepareOmpMcpConfigUsesProviderNamespace` asserts byte-exact passthrough at the execenv layer, which sits below this change. It also leaves the merge path claiming omp has no runtime servers while `listRuntimeLocalMcpServers` still reads `~/.omp/agent/mcp.json` and shows them in capabilities.

## Where this could go

The only route upstream leaves open is the one its README names: author a pi **extension** that adds MCP support, and have the daemon inject it per task. That's a different change, and it means owning the extension against an interface pi has deliberately not provided. If you want to attempt it, the acceptance criterion should be a recording of pi actually invoking a tool from a managed server — not a materialised file.

Closing this one; #7957 stays closed for the same reason. The tab being hidden for `pi` is the honest state today: showing it would let someone save a value that silently does nothing. Genuinely appreciate the contribution, and #7565 remains the right precedent — it just happens to be the case where the runtime really did read the file.

PR #7978, opened 2026-09-03
fix(wecom): isolate a group chat's session per sender

ii-jj:
## Problem

In a WeCom (企业微信) group, every member shares one chat session.

`channel_chat_session_binding` keys the live session on
`(installation_id, channel_chat_id)`, and the wecom adapter passes
`Source.ChatID` straight through as the binding key
(`wecom_resolvers.go`). For a group that id is the room, so:

- the agent reads a context in which several people's turns are interleaved;
- the chat that surfaces in Multica belongs to whoever spoke to the bot first,
  and everyone else's messages land inside it.

Reproduce: `@bot` from user A in a group, then `@bot` from user B. B's turn is
appended to A's session.

## Why sender, and not a thread

Every other adapter already isolates — Slack on `channel:threadRoot`, Lark on
`chat:topic`, and `EnsureSessionInput` is documented for a composite key.
WeCom was the only one passing the raw chat id.

It cannot copy them: an `aibot_msg_callback` carries `msgid` / `aibotid` /
`chatid` / `chattype` / `from.userid` / `response_url` / `msgtype` / a content
body / `quote` / `stream` — and nothing else. There is no thread or branch id.
`quote` is the closest thing and it only mirrors the quoted message's
*content* (`{"msgtype":"text","text":{"content":"…"}}`) — no msgid, no
userid — so no thread root is recoverable from it either.

The sender is the only dimension the wire offers, and it happens to be the one
a member expects: their own conversation with the bot.

## Change

- group binding key becomes `chatid:senderid`; the real chatid moves onto the
  binding config, the same shape Slack and Lark use for a composite key
- outbound reads it back through `wecomOutboundChatID`, which falls back to
  `channel_chat_id` so 1:1 bindings — and group bindings written before this
  change — stay deliverable
- a group message with no `from.userid` falls back to the room key
- 1:1 is untouched: the wecom ChatID already IS the userid there

The bot's reply still goes to the room — aibot has no per-member visibility —
so only the CONTEXT becomes private, not the answer.

## Migration

None. A pre-existing group binding is simply never looked up again; each
member's next message opens their own session. The cost is a one-time context
reset per group, and the old binding row stays addressable for any reply still
in flight against it.

## Tests

`TestWecomSessionRouting` pins the isolation contract (p2p key, two senders in
one room getting two keys, the no-sender fallback) and `TestWecomOutboundChatID`
pins outbound addressing across all four binding shapes, including the
pre-split and unparsable-config rows. Existing group-key assertions updated.

`go test ./internal/integrations/wecom/ ./internal/integrations/channel/...`
passes; `go vet` and `gofmt` clean.


ii-jj:
Rebased onto `a9e82c797`. Two conflicts, both from `main` editing the lines this branch moved:

- `outbound.go` — `main` replaced `content != ""` with `hasVisibleChar(content)` (a completion of whitespace is not a message, MUL-7319). Kept that test; the argument it guards is still `chatID` from `wecomOutboundChatID(binding)` rather than `binding.ChannelChatID`, which is what this branch is about.
- `resolvers_test.go` — `main` added `CommandText` to the group fixture and its assertion. Kept both, with this branch's `SenderID` on the source and the `GROUP_1:TUSER_A` binding key.

`./internal/integrations/wecom` and `./internal/integrations/channel/...` green against a database migrated through 468.

PR #7980, opened 2026-09-03
MUL-7019 fix(wecom): show the agent the message a sender quoted

ii-jj:
## Problem

`aibotMsgCallback` does not declare `quote`, so the message a WeCom sender
replied to is dropped at the JSON boundary.

The callback carries it on every reply — [接收消息](https://developer.work.weixin.qq.com/document/path/100719)
documents `quote` as "引用内容，若用户引用了其他消息则有该字段":

```json
"quote": { "msgtype": "text", "text": { "content": "这是今日的测试情况" } }
```

Replying to a message is how people ask about it. "这个怎么处理" quoting an
alert is a complete question in the chat and an empty one to the agent, which
receives the four words and none of what they point at. The sender sees a bot
that ignored what they were obviously asking about, with nothing anywhere
saying why.

## Change

The quoted content is rendered as a labelled blockquote ahead of the sender's
own words:

```
> [Quote] 生产库连接数打满了

这个怎么处理
```

- `[Quote]` is spelled like `mediaPlaceholder`'s existing `[Image]` / `[File]` /
  `[Video]`, which are already byte-identical to Lark's and DingTalk's, so an
  agent reading every channel through one prompt meets one vocabulary
- the blockquote wraps every line — a multi-line quote would otherwise come
  apart from its label
- a quoted attachment renders the same placeholder a sent one does
- `quotedMessage` exists only because a quoted 图文混排 nests one level deeper
  than a mixed run; everything else comes off `mixedItem`

## What it deliberately does not do

**The quote never reaches `ownCommandSource`.** The command parsers read the
first non-empty line, and a quoted line is not one the sender typed here.
Prefixing it would let a quote of somebody else's `/issue …` file an issue
nobody asked for — pinned by a test. The rendering happens last in
`channelMessageFromCallback`, after `normalizeWeComControlLayout`, so the
control-layout rewrite and the command source it can hand back still read the
body the sender actually composed. Only the stored, agent-visible text grows.

**The quote is rendered, not resolved.** WeCom sends no msgid and no author
userid alongside the quoted content, so there is nothing to resolve it against
— not a stored `chat_message`, not a channel user binding. Rendering it into
the body is the whole of what the wire supports.

## Tests

`TestQuotedContext` covers the label, multi-line, quoted attachment, quoted
mixed message, absent quote, and an unknown quoted kind.
`TestChannelMessageFromCallback_QuoteLeadsBodyButNotCommand` pins the three
things that make it safe: the quote leads the body, a quoted `/issue` does not
become this sender's command, and a `/new …` keeps its quote in the new
session's first turn while staying parseable.

`go test ./internal/integrations/wecom/ ./internal/integrations/channel/...`
passes; `go vet` and `gofmt` clean.

## Not in this PR

The same callback also carries `response_url` (a temporary URL for replying
over HTTP) and `stream` (a streaming-message id), both likewise undeclared.
Neither is a parsing gap: `response_url` is a second outbound transport
alongside the WebSocket `aibot_send_msg` this adapter uses — interesting as a
fallback for `errNoLiveConnection`, but a transport, not a field — and `stream`
needs progressive output the engine has no concept of, since a reply is one
complete message at `EventChatDone`. Happy to open either separately.


seacen (changes requested):
The approach is right — it mirrors Lark's enricher, `[Quote]` joins the placeholder vocabulary, and keeping the quote out of `ownCommandSource` is the correct boundary. One blocking case, found by running it through the real Router:

**A bare `/new` behind a quote leaks the directive into the new session.** Stored body: `"> [Quote] Q3 毛利率 42.1%\n\n/new"`. `normalizeWeComControlLayout` declines a bare directive on purpose (`control.Body == "" && !hasMedia` → false), leaving `Text = "/new"` for the Router to strip through its `msg.Text == msg.CommandText` check — and prepending the quote breaks that equality, so the Router no longer rewrites and `/new` becomes the first line of context every later turn inherits. A bare `/clear` goes the other way: `ForceFresh` is not set, the Router rewrites `Text` to `""`, and the quote is dropped silently — which contradicts the "keeps its quote" case the PR tests.

Suggested fix: handle the bare-directive-behind-a-quote case in the adapter rather than relying on the Router's equality — strip the directive there, leave the quote block (or nothing) as the body for `/new`, and set `ForceFresh` for `/clear`. The regression should drive `engine.Router`, since `channelMessageFromCallback` alone cannot see this.

Non-blocking: the quoted content has no length bound; a few hundred runes with an ellipsis would keep a quoted document from becoming the body.

ii-jj:
Confirmed both, fixed both, and the trace turned up a third defect underneath them. Pushed as `ebcb4b08d`.

**The blocking case.** Reproduced exactly as described, and the diagnosis holds: `normalizeWeComControlLayout` declines a bare directive because `control.Body == "" && !hasMedia`, and prepending the quote then breaks the Router's `msg.Text == msg.CommandText` equality, so nothing strips it.

Fixed in the adapter as suggested. A quote now counts as content for the same reason media does — it is why the directive-only layout is not the empty pending sentinel — so the parameter is `hasOtherContent` rather than `hasMedia`:

```go
media := mc.attachments()
quoted := mc.quotedContext()
normalizedText, control, controlNormalized := normalizeWeComControlLayout(
    mc, text, command, chatType, botDisplayName, len(media) > 0 || quoted != "",
)
```

Bare `/new` behind a quote now leaves the quote as the body; bare `/clear` keeps the quote and sets `ForceFresh` through the existing `control.Kind == ControlCommandFreshSession && control.Body == ""` path.

**The third one, which no adapter change could have fixed.** With the above in place a bare `/new` behind a quote still lost the turn, one layer down — `engine/router.go:402`:

```go
persistMessage := msg.CommandText != "" || hasMedia
```

A turn whose entire content is a quote has an empty `CommandText` and no media, so the Router reads it as empty and never persists it. `msg.Text` is now part of that test. This is invisible from `channelMessageFromCallback`, which is why the regression had to drive the Router — the point you made about where the test belongs is what surfaced it.

**Length bound.** `maxQuotedRunes = 500`, truncated on a rune boundary with a trailing ellipsis, so a quoted document cannot become the body.

**Tests.** Four in the adapter package and three driving `engine.Router` end to end, including the sentinel case (a bare directive with *no* quote must stay the pending sentinel) so the fix cannot over-strip. Each change was reverted in isolation to prove it load-bearing: reverting the adapter fails the two adapter cases, reverting the Router line fails the `/new` case, and the sentinel tests stay green throughout.

seacen (commented):
The blocker is fixed, and the third defect you found underneath it is the more valuable half of this round. I re-ran my own Router-driving regressions against `ebcb4b08d` — a `/clear` behind a quote still starts a fresh session, a quoted `/clear` does not restart somebody else's, the binder still hands the unquoted line to the classifier, and neither directive survives into the body. All green.

`persistMessage` is worth calling out in the description: it is not a WeCom bug. Any adapter that puts content in `Text` the sender did not type here hits it, and Lark's enricher already prepends quoted-reply context (`lark/content_flatten.go`). This fixes that latent case too, and a reader of the diff will not guess that from the WeCom-shaped tests around it.

## One thing I would want decided before merge

A bare directive behind a quote now starts an agent run whose entire prompt is the quoted message. Your own test states it: `TestBareClearBehindAQuoteKeepsTheQuote` asserts `h.tasks.wasCalled`, and for `/new` `PersistMessage` is true so `processClaimed` prepares and enqueues.

On `main` today a bare directive never does that. `bareFresh` sends it to `MarkPendingFresh` and returns `OutcomeFreshPending` — the route opens, nothing runs, and the person's next message is the turn. This PR moves that decision into the adapter (the directive is consumed there, so `ParseControlCommand(msg.CommandText)` no longer fires and `bareFresh` is false), and the adapter answers it differently.

The case that made me stop: quoting the **bot's own last answer** and typing `/clear` is an ordinary way to say "forget that, start over". Under this change the bot starts a run whose prompt is its own previous answer and replies to it — in a group, in front of the room, with no unsend on the platform. The same shape applies to quoting a colleague: an unrequested answer to words they wrote and the sender never asked about.

The reading behind the current behaviour is defensible — the sender attached the quote deliberately, so treat it as the turn. Mine is that a bare directive means "no question yet", and the quote is the subject of the question they are about to type. I do not think a reviewer should pick that for you, but it should not be settled by a `!hasMedia` parameter growing a second meaning.

If you want the conservative shape, it does not cost the fix: let a quote license *stripping* the directive exactly as it does now, and keep "a bare directive is not a turn" where Router already decides it — the sentinel survives, the quote is dropped with the directive, and nothing runs until the person types their question. If you keep the current behaviour, the description is where to say so, because `TestABareDirectiveWithNothingElseIsStillTheSentinel` reads like the sentinel is intact when what is intact is the no-quote half of it.

## Not yours

`cmd/server`'s `TestReadinessEndpoints` fails on this branch, and also at its merge base `035abddfe` — 67 commits behind `main`, where it passes. Nothing to chase.

seacen:
One follow-up I'd like to claim, because it is the other half of the same message and I have it working already.

A quoted **picture** currently reaches the agent as the four characters `[Image]`. The most common shape of this in a room is someone quoting a screenshot of an error and typing "这个怎么处理" underneath — after this PR the agent gets the words and a placeholder where the error is, which answers the label rather than the picture.

The callback carries the quoted attachment the same way it carries a sent one: a pre-signed url and its `aeskey`, under `quote.image` (or inside `quote.mixed.msg_item`). So the fix is small and lives entirely in the two places this PR already touches:

- collect the quoted message's attachments ahead of the sender's own, so the Nth placeholder in the body is the Nth entry of `InboundMedia` — the correspondence 图文混排 already relies on;
- render a quoted attachment with a marker that says it is quoted, so an agent can tell a picture it is being shown from one it is being handed.

I'd rather not push it into your branch mid-review. Two options, whichever suits you: I open it as a follow-up PR the day this one lands, or I open it now against your branch so you can see the whole shape before merging. Say the word and I'll do either — happy for it to be a separate review in any case, since it pulls in the media resolver and yours does not.

(Unrelated to the above: `TestReadinessEndpoints` fails on this branch and also on its merge base `035abddfe`, where `main` passes — 67 commits of drift rather than anything you changed.)

ii-jj:
Rebased onto `a9e82c797`. The Router edit is gone, and what replaced it answers the question you left open.

**`main` landed the general form of the change this branch was making.** `channel.InboundMessage.HasSelectedContext` (message.go:149) says "Text includes a quote or forward explicitly selected by the sender … a selected context is input even when a control command has no own body", and Router now derives both `persistStartedMessage` and `bareFresh` from it. That is the rule my `msg.Text != ""` line was reaching for, stated properly and with the automatic-history case excluded — which my version could not distinguish. DingTalk, Lark and Telegram already set it; WeCom did not, so the adapter sets it now (`quoted != ""`) and the Router hunk is dropped.

Behaviour is what you verified on `ebcb4b08d`: a bare directive behind a quote is a turn whose prompt is the quoted message.

**On the decision.** I was ready to take the conservative shape you described, but I do not think it is mine to take any more: `main` has since decided that sender-selected context is input even when the directive has no body, for every channel. A WeCom quote is sender-selected context. Deciding otherwise here would make WeCom the one adapter where a quote does not count, on a branch whose subject is quotes. If the decision is revisited, `hasSelectedContext` in router.go:390 is one line and it moves for everyone at once — which is the outcome your review was asking for when you said it should not be settled by a `!hasMedia` parameter growing a second meaning.

Pinned on both sides: the adapter test asserts the flag is set behind a quote and unset for a bare directive with nothing else, and the Router regression sends it. Hardcoding it `false` fails `TestABareNewBehindAQuoteLeavesOnlyTheQuote`.

**Your quoted-picture follow-up: please take it, as a separate PR once this lands.** It is the right fix and the ordering contract you describe is already settled upstream — DingTalk does exactly that shift in `dingtalk/inbound.go:400-415`, prepending the quoted resources and advancing every current resource's `InlineIndex` by the placeholder count the quoted block introduces, so the Nth placeholder stays the Nth entry. WeCom's `attachments()` will need the same treatment against `quotedContext`'s 500-rune elision, since a truncated block must not drop a placeholder it still counts. Separate review is right: it pulls in the media resolver, and this branch deliberately does not.

`TestReadinessEndpoints` passes on the rebased branch — it was the 67-commit drift, as you said.

seacen (approved):
Approving. The open question is answered, and better than by either of us picking a side.

I checked the claim rather than taking it: `channel.InboundMessage.HasSelectedContext` is on `main` at `message.go:149-153`, with "a selected context is input even when a control command has no own body" in the doc itself, and `router.go:390-392` derives both `persistStartedMessage` and `bareFresh` from it. So the decision I said a reviewer should not make for you has been made upstream, for every channel at once — and a WeCom quote is sender-selected context by any reading. Making it the one adapter where a quote does not count, on a branch about quotes, would be the odd choice.

The part I most wanted is that the lever is now one line in a shared place rather than a parameter here growing a second meaning. If the behaviour is ever revisited it moves for Lark, DingTalk and Telegram in the same commit.

Dropping the Router hunk for the upstream flag is the right trade too: your `msg.Text != ""` could not tell sender-selected context from automatic history, and `HasSelectedContext` says so explicitly.

On the quoted picture: taking it, as its own PR once this lands. Thanks for the ordering contract — `dingtalk/inbound.go:400-415` shifting every current resource's `InlineIndex` by the count the quoted block introduces is exactly the invariant I needed, and you are right that the 500-rune elision is where WeCom differs: a truncated quote must not drop a placeholder it still counts. That pulls in the media resolver, so it does not belong here.

Bohan-J (changes requested):
The rebase is the right resolution — deferring to `HasSelectedContext` instead of growing a second meaning on the `hasOtherContent` parameter means the lever moves for every channel at once, and pinning both sides of the sentinel is what makes the directive handling safe to read. Keeping the quote out of `ownCommandSource` is still the correct boundary.

One blocking case, which the rebase did not introduce but which the branch now owns: this makes WeCom the third adapter that enriches `Text`, and the first one that does not satisfy the invariant at `engine/router.go:200-208`.

## A reply that is only a screenshot leaves `CommandText` empty behind an enriched `Text`

**Trigger:** quote a message, then send only an image, file or video, typing nothing. In WeCom that is an ordinary way to reply — it is the same shape as the quoted-screenshot case, seen from the sender's side.

`ownCommandSource` answers `""` for a standalone photo, file or video by design (`wecom/ws_frame.go:402`, and the doc comment above it says so deliberately). `quotedContext` then prepends the quote to `Text`. So the adapter hands Router:

```
Text        = "> [Quote] 生产库连接数打满了\n\n[Image]"
CommandText = ""
```

and `router.go:209-211` fills `CommandText` from the **already enriched** `Text`. That is the exact failure the comment above it names:

> INVARIANT: an adapter that enriches Text with content the member did not type — a quoted reply, recent group history — MUST set CommandText itself before its message reaches Router. This fallback assigns the ALREADY enriched Text, so an enriching adapter that leaves CommandText empty silently reinstates #8058 (the enrichment prefix becomes the Chat title) while every title test stays green. lark and telegram are today's only enriching adapters and both comply.

`engine/session.go:365` states it from the other end: *"CommandText is the user's OWN typed text … the adapter supplies it because enrichment is platform-specific."*

**What the person sees:** the Chat is named after the message they quoted, carrying the adapter's own label into the name. Feeding the adapter's output through `chatTitleSource` / `deriveFirstMessageTitle`:

```
screenshot only, no quote (main)     → ""  → falls through to the media title path ("Image chat" / filename)
screenshot as a reply (this branch)  → "[Quote] 生产库连接数打满了"
```

The last clause of the invariant held, too: CI is green and three rounds of review did not surface this, because no title test drives an enriching WeCom message.

**Suggested fix** — lark's shape. It snapshots the visible body before enrichment at `lark/ws_frame_decoder.go:113` for exactly this reason:

```go
ownBody := text
if quoted != "" {
    if text == "" {
        text = quoted
    } else {
        text = quoted + "\n\n" + text
    }
}

// ... the ForceFresh command-source assignment stays where it is ...

// Router fills an empty CommandText from Text (router.go:209), and Text is
// enriched by the time it gets there. Hand it the body the sender actually
// sent instead of the one this adapter wrote.
if command == "" && quoted != "" {
    command = ownBody
}
```

I ran this against `6b8bf854e` locally: no new failures in `./internal/integrations/wecom/` or `./internal/integrations/channel/engine/`, and the case `ownCommandSource` exists for still works — a screenshot followed by `/issue 登录坏了`, sent behind a quote, keeps `CommandText = "/issue 登录坏了"` and `SkipAgentRun = true`.

## Non-blocking

1. **`[Quote]` leaks into the Chat title for a bare directive behind a quote.** A `/new` or `/clear` sent as a reply has the quote as its whole content, so the title comes out as `[Quote] Q3 毛利率 42.1%` — `chattitle.Derive` strips `>` but not `[Quote]`. The fix above does not cover this one, because the quote genuinely is all the person sent; stripping the label would have to happen in title derivation. Cosmetic, but worth a line in the description if you leave it as is.

2. **Confirming for the follow-up, not a defect here:** the quoted `[Image]` placeholder cannot misalign anything today, because WeCom never sets `InlinePlaceholder` / `InlineIndex` — dingtalk is the only adapter in the tree that does. So the placeholder-to-media correspondence is something the quoted-picture PR will establish for WeCom, not something this branch breaks.

3. `maxQuotedRunes = 500` is argued rather than assumed, and bounding by rune rather than byte is the right call for the content this actually carries. No concern.

## What I verified, and what I did not

I drove `channelMessageFromCallback` over text / voice / image / 图文混排 × quoted / not quoted × bare directive / directive with a body / group @-mention, and pushed the adapter output through the `engine.Router` test harness to observe the title source. The seven tests this branch adds pass.

DB-backed tests do not run on my machine (stale local schema); the failing set is identical on `main`, so nothing here regressed them, and CI covers them. I did not capture a real WeCom callback — the shape of `quote` is checked against the platform docs and your description only.

ii-jj:
Confirmed and fixed in `4404bfc7a`. You are right on every step of it, and I reproduced the whole chain before touching anything rather than taking the diagnosis on trust.

**The observed shape**, driving `channelMessageFromCallback` with an image-only callback carrying a quote:

```
ownText     = "[Image]"
Text        = "> [Quote] 生产库连接数打满了\n\n[Image]"
CommandText = ""
```

and then through the title hop itself:

| what the person sent | `chatTitleSource` | title |
|---|---|---|
| screenshot, no quote (`main`) | `"[Image]"` | `""` → media path |
| screenshot as a reply (before this commit) | `"> [Quote] 生产库连接数打满了\n\n[Image]"` | **`"[Quote] 生产库连接数打满了"`** |
| screenshot as a reply (now) | `"[Image]"` | `""` → media path |

So the Chat was named after the message the sender replied to, and the label went with it.

**The fix is the shape you suggested**, snapshotting the body before enrichment and filling only an empty command source. The reason it is safe to hand a placeholder to the command source — rather than something cleverer — is the last row above: `deriveFirstMessageTitle` drops placeholder lines when the message has media, so `"[Image]"` lands the title on exactly the path the same screenshot takes with no quote. That equality is what the engine test asserts, instead of hardcoding what the title should be.

**Tests, both sides, as with the rest of this branch:**

- `wecom/TestAQuotedScreenshotHandsRouterTheSendersOwnBody` — fails without the fix with `CommandText is empty behind an enriched Text`. I checked that by neutering the assignment, not by reasoning about it.
- `wecom/TestAQuotedScreenshotWithWordsKeepsTheTypedCommand` — the over-application guard you also checked: a screenshot with `/issue 登录坏了` under it, behind a quote, keeps `CommandText = "/issue 登录坏了"` and `SkipAgentRun`.
- `engine/TestAnEnrichedWeComBodyDoesNotBecomeTheChatTitle` — compares the quoted and unquoted titles, and refuses to pass if the empty-`CommandText` fallback ever stops leaking the quote, so it cannot quietly become a test of nothing. This is the title test the invariant's own comment predicted was missing.

**Non-blocking 1 — `[Quote]` in the title of a bare directive behind a quote.** Left as is and now stated in the commit message: there the quote genuinely is all the person sent, so the body is right and only the label is noise; stripping it belongs in title derivation, where it would apply to every adapter that renders one.

**Non-blocking 2** — agreed, and thanks for confirming it from your side. `InlinePlaceholder` / `InlineIndex` are untouched by this branch; establishing that correspondence for WeCom is the quoted-picture PR's job.

Rebased onto `7e4758ac1` while I was here. `./internal/integrations/wecom/` and `./internal/integrations/channel/...` are green; the 61 DB-backed tests skip on my machine (no local database), so CI is the authority on those.

Bohan-J (approved):
Approving. `4404bfc7a` fixes the case exactly, and the way it is pinned is better than what I asked for.

I re-ran it: a quoted screenshot now hands Router `CommandText = "[Image]"` instead of an empty one, so the enriched `Text` no longer gets assigned back through the fallback, and the Chat titles off the media path the same screenshot takes with no quote. A quoted `/issue` still never reaches the command source, a screenshot with `/issue 登录坏了` under it still files that issue and keeps `SkipAgentRun`, and the no-quote sentinel is untouched. Build, `gofmt` and the branch's own tests are clean; the two packages show no new failures against `main` (the DB-backed failures on my machine are a stale local schema and fail identically on `main`).

The part I did not expect: `TestAnEnrichedWeComBodyDoesNotBecomeTheChatTitle` asserts the leak is still reachable with an empty `CommandText`, so the test fails if the invariant it guards ever moves rather than quietly passing for the wrong reason. That is the failure mode the invariant's own comment describes — CI green through three rounds — and the test now refuses to be the thing that stays green.

Leaving the `[Quote]` label in the title for a bare `/new` / `/clear` behind a quote is the right call, and the commit message saying so is enough: the quote genuinely is all the person sent there, and stripping the label belongs in title derivation.

Thank you for this one. Four rounds, two defects found underneath the original fix — the `persistMessage` gap in Router and this command-source case — and every round came back with the diagnosis reproduced, the fix scoped to the layer that owns it, and both sides pinned. The rebase call was the best of them: dropping your own Router hunk once `main` landed `HasSelectedContext` left WeCom using the shared rule instead of a second meaning growing on this branch. That is a genuinely good first contribution to this repo.

Merging now. Looking forward to the quoted-picture follow-up.

PR #7985, opened 2026-09-03
MUL-7024 fix(wecom): clear the previous bot's rows when an agent swaps bots

### LINKED ISSUE #6547: fix(wecom): swapping a bot on the same agent leaves the previous bot's bindings behind
Follow-up from #5833 (WeCom smart-bot integration). Found in review; confirmed against the merged code.

cc @leroy-chen @seacen — this one is yours as the directory's code owners, no urgency.

## What happens

Swapping the bot on an agent that already has one reuses the existing `channel_installation` row but leaves every dependent row from the previous bot in place.

`InstallationService.Upsert` (`server/internal/integrations/wecom/installation.go`) calls:

```go
ReclaimDeadChannelInstallationByAppID(ctx, {ChannelType: "wecom", AppID: p.BotID, ...})
```

`AppID` here is the **new** bot's id. That reclaims a dead installation elsewhere holding the *new* bot's routing slot — it does nothing about the rows the *current* installation accumulated under the *old* bot.

`UpsertChannelInstallation` then conflicts on `(workspace_id, agent_id, channel_type)`, so the **same installation row and id** is reused with only `config` replaced (bot_id + sealed secret).

Everything keyed on `installation_id` therefore survives the swap: `channel_user_binding`, `channel_chat_session_binding`, binding tokens, and inbound dedup rows.

## Why it's wrong

WeCom aibot userids are anonymized **per (bot, user)** — that's the premise the whole binding flow rests on (see the comment at the top of `wecom/binding.go`). So the surviving `channel_user_binding.channel_user_id` values belong to a namespace the new bot does not share. They are not merely stale, they are meaningless to the new bot.

The most concrete symptom is on the inbox push path. `Outbound.tryDeliverInbox` looks the binding up by `(workspace_id, multica_user_id, channel_type)` — not by bot — then does:

```go
sender := o.senders.get(binding.InstallationID)   // resolves: installation id was reused → live NEW bot
sender.sendText(binding.ChannelUserID, chatTypeSingleInt, content)  // OLD bot's userid
```

So a notification is addressed to an old-bot userid over the new bot's connection. Chat session bindings carry over the same way.

## Repro

1. Connect bot X to agent A; have a member bind their account by messaging the bot.
2. Disconnect bot X from agent A.
3. Connect bot Y to the same agent A.
4. Trigger an inbox notification for that member — it is addressed to their bot-X userid over bot Y's connection.

## Suggested fix

Two workable options; the first is preferred:

1. When `Upsert` reuses an installation whose stored `bot_id` differs from the incoming one, clear that installation's dependent rows in the same transaction (`channel_user_binding`, `channel_chat_session_binding`, binding tokens, inbound dedup) before the upsert — the application-layer cleanup the no-FK rule requires. `ReclaimDeadChannelInstallationByAppID` already does exactly this shape of cleanup for the dead-owner case, so it is mostly a matter of covering the reuse case too.
2. Or refuse a bot change on an existing installation outright and require disconnect-then-reconnect, with the disconnect doing the cleanup.

Worth a test that binds a user under bot X, swaps to bot Y on the same agent, and asserts no binding rows survive for the installation.

## Notes

Deliberately not held as a merge blocker for #5833 — it needs an admin to swap bots on an agent, and the integration ships community-maintained. Filing it so it isn't lost.

ii-jj:
Closes #6547.

## Problem

Swapping which bot an agent uses reuses the `channel_installation` row.
`UpsertChannelInstallation` conflicts on `(workspace_id, agent_id,
channel_type)` and only rewrites `config`, so the row and **its id** survive —
and with them every dependent row keyed on `installation_id`.

They are not merely stale. A WeCom aibot userid is anonymized per `(bot, user)`
— the premise the whole binding flow rests on (`wecom/binding.go`) — so a
carried-over `channel_user_binding` holds an id from a namespace the new bot
does not share.

The concrete symptom the issue names, in `Outbound.tryDeliverInbox`:

```go
sender := o.senders.get(binding.InstallationID)  // installation id reused → the LIVE NEW bot
sender.sendText(binding.ChannelUserID, chatTypeSingleInt, content)  // the OLD bot's userid
```

A p2p `channel_chat_session_binding` carries that same userid as its
`channel_chat_id`, and a queued `channel_task_delivery` carries it again.

## Change

`ClearChannelInstallationBotScopedRows` — the reuse-case counterpart to the
cleanup `ReclaimDeadChannelInstallationByAppID` already performs for the
dead-owner case, which is the shape the issue's option 1 pointed at.

It runs inside the install transaction, **before** the upsert, so a swap that
fails afterwards leaves the old bot's rows intact rather than half-cleared. It
is gated on `carried.BotID != "" && carried.BotID != p.BotID`, right beside the
existing display-name carry-forward that already reasons about the same swap.

Cleared: `channel_user_binding`, `channel_chat_session_binding`,
`channel_binding_token`, `channel_inbound_message_dedup`,
`channel_task_delivery`, `channel_outbound_message`, and
`channel_outbound_card_message` (no `installation_id` and no FK, so it is
reached through the just-removed bindings — the same link the reclaim uses).

The issue named the first four. The outbound three are the same defect one step
later: a queued delivery carries the old bot's `channel_chat_id` and would be
sent over the new bot's connection, which is the inbox symptom with a different
entry point. Happy to drop them if you'd rather keep the change to the named
set.

## What deliberately stays

- **the installation row and its `channel_inbound_audit`** — the row is the
  same connection under new credentials, and an audit row still records
  something an operator can act on. Reclaim DETACHes audit only because the
  installation is going away; here it is not.
- **`chat_session` and `channel_chat_context_generation`** — deleting the
  binding detaches the WeCom route from the Chat, the same shape a disconnect
  leaves behind. The Chat and its history are the user's work product and
  outlive whichever bot carried them.
- **everything, on a same-bot secret rotation.** An admin pasting a fresh
  secret for the same bot is rotating a credential, not swapping. Clearing
  there would silently unbind every member and make them all link their
  account again — a worse bug than the one being fixed, so it has its own test.

## Tests

`installation_bot_swap_test.go`, DB-backed in the shape of
`installation_bot_name_test.go` next to it (same `reclaimTestDB` skip):

- `TestSwappingTheBotClearsThePreviousBotsRows` seeds one row of each kind
  under bot A, swaps to bot B, and asserts none survive. It also asserts the
  installation id was in fact reused — if that ever stops being true the test
  would be asserting nothing, so it says so rather than assuming it.
- `TestRotatingTheSecretKeepsTheBindings` is the other direction: re-installing
  the SAME bot must leave all of them in place.

Verified against a real migrated Postgres 17, not just by skipping: both pass
with the fix; with the gate forced off, `TestSwappingTheBotClearsThePreviousBotsRows`
fails on all four seeded tables while `TestRotatingTheSecretKeepsTheBindings`
still passes — so the pair is guarding both over- and under-purging rather than
mirroring the implementation.

Full `./internal/integrations/... ./internal/handler/...` passes with
`DATABASE_URL` set; `sqlc generate` is clean (the regenerated
`pkg/db/generated/channel.sql.go` diff contains only the new query);
`go vet` and `gofmt` clean.


seacen (approved):
Approving — this is the fix #6547 asked for, in the shape option 1 pointed at, and it lines up with `ReclaimDeadChannelInstallationByAppID`'s set rather than inventing one. Both DB tests pass here against a database migrated through 449.

Three non-blocking notes and one question:

1. The slot lock is the NEW bot's. The old bot's socket is live until the change propagates, so a message arriving under it between the DELETE and COMMIT can insert a binding that survives the sweep. Small window; worth a sentence in the comment, or a second clear once the Supervisor has torn the old connection down.
2. Deleting `channel_task_delivery` for a task still running means its answer is skipped silently — `processEvent` returns nil on no row, with no counter and no log line. Right direction (the old userid is unreachable either way), but a log line would help whoever wonders where the answer went.
3. The tests seed and assert four of the seven tables; the three outbound ones are cleared but never seeded. One row each would make the query's full set load-bearing.

Question: is a group `chatid` stable across bots? If it is, dropping group session bindings restarts every group conversation (history kept, route detached). If it is per-bot like the userid, the delete is right. I don't have a measurement either way.

ii-jj:
All three notes addressed in `df20895d9`. None of them changes what the sweep deletes.

**1 — the race window.** Added as a sentence in the comment rather than a second clear, because this transaction cannot observe the moment you would key the second clear off:

> The sweep is not airtight, and cannot be from here. The slot lock taken above is the NEW bot's; the OLD bot's socket stays live until the Supervisor tears it down, so an inbound message arriving under it between this DELETE and the COMMIT re-inserts a binding that outlives the sweep. The window is short and self-healing — that binding is the old bot's namespace, so the next delivery over it fails to address and the user re-binds — and closing it properly means a second clear once the old connection is confirmed down, which this transaction cannot observe.

If the Supervisor grows a "old connection torn down" signal, that is the hook to hang the second clear on.

**2 — the dropped answer.** `ClearChannelInstallationBotScopedRows` is now `:one`; every CTE carries a `RETURNING` and the statement ends with a count per swept table. `InstallationService` gained a nil-safe logger in the shape `channel_media_reconciler.go` already uses, and the swap logs:

```
wecom: bot swap cleared the previous bot's rows
  installation_id=… previous_bot_id=… user_bindings=N
  chat_session_bindings=N queued_task_deliveries_dropped=N
```

You are right that this is the only place that can say it — `processEvent` finds no row, returns nil, and has no counter of its own.

**3 — the three unseeded tables.** Now seeded: `channel_task_delivery`, `channel_outbound_message`, and `channel_outbound_card_message`. The card table carries no `installation_id`, so `countBotScopedRows` counts it by `chat_session_id`. With the swap gate forced off, all seven tables report surviving rows; with it on, both directions pass. Verified against a database migrated through 450.

**The `chatid` question — I could not answer it, and I do not want to imply otherwise.** WeCom's 接收消息 doc (`path/100719`) does not document `chatid`'s scope at all. It documents the userid as an 企业主体下的加密 userid, which is not obviously the same statement as the per-`(bot, user)` premise the binding flow already rests on (`binding.go:10`, `wecom_resolvers.go:147-155`) — that premise predates this PR, and I did not find its source either. There is nothing in the repo to measure against.

What I can argue is the asymmetry, which is why I left the group delete in. If `chatid` is stable across bots, deleting costs one detached route: history stays and the next message in the group re-binds. If it is per-bot and we keep the binding, we address the old bot's namespace and deliveries land nowhere, silently. A re-bind is the cheaper wrong answer, so deleting is the safer default under uncertainty — but if someone measures it and it turns out stable, narrowing the sweep to p2p bindings is a small follow-up and I am happy to take it.

seacen (approved):
All three addressed, and the answer to the `chatid` question is the right one — saying "I could not measure it" and arguing the asymmetry beats asserting a scope neither of us can source. I have no measurement either. Keeping the group delete under uncertainty is the direction I would take too: a re-bind is visible and self-healing, a silently unaddressable delivery is not.

Verified locally against a database migrated through 450:

- both tests pass;
- with the swap gate forced to `false`, all **seven** tables report surviving rows — so the seeded set is load-bearing now, not just the four;
- the same-bot rotation stays green throughout, which is the half that a too-eager purge would break;
- `./internal/integrations/wecom` and `./internal/handler` green.

The `:one` rewrite is sound: data-modifying CTEs run to completion whether or not the primary query reads them, so moving `channel_user_binding` into a CTE and ending on a `SELECT` of three counts deletes exactly what it did before.

Two nits, neither blocking:

1. **The log covers three of the seven tables.** The argument for logging `channel_task_delivery` — a queued answer disappears and `processEvent` returns nil with no counter and no line — applies word for word to `channel_outbound_message`. It is likely zero rows for WeCom in practice, which is why I would just add it to the count rather than argue about it. The PR description says "a count per swept table", so the mismatch is with your own sentence rather than with the code.
2. **The race-window comment says the binding is self-healing** because the next delivery over it fails to address. Worth naming what the user sees while it heals: a message that gets no answer, and no log line saying why, until they re-bind. That is the same silence item 2 just fixed one line up, so a reader may fairly ask why one gets a log and the other a comment.

Approving; neither nit needs another round from my side.

ii-jj:
Both nits taken, in `0123d5637`, and rebased onto `a9e82c797`.

**1 — the fourth count.** `channel_outbound_message` now has its own CTE count and its own field on the swap line (`queued_outbound_messages_dropped`). You were right that it is the same argument as `channel_task_delivery` word for word, and that "likely zero for WeCom" is a reason to count it rather than to argue about it. The PR description's sentence and the code now agree.

**2 — the race window.** The comment says what healing looks like from the user's side: one message that gets no answer and no line saying why, until they re-bind. Same silence as item 2 one line up, which is the comparison you were making.

**And the counts now have a test.** They go nowhere but the log — `Upsert` returns an `Installation`, not a report — so `TestSwappingTheBotClearsThePreviousBotsRows` captures the line through the service's own logger and asserts one count per swept table. Dropping the new field from the log fails it with `queued_outbound_messages_dropped = 0, want the 1 seeded row reported`; that is how I checked it is load-bearing rather than decorative.

Verified against a database migrated through 468: both bot-swap tests, `./internal/integrations/wecom` and `./internal/handler` green.

seacen (approved):
Re-approving on `0123d563`. Both nits are taken and the fourth count came back with something I did not ask for and should have: a test.

Counts that go only to a log line are the easiest thing in a PR to leave decorative, and `TestSwappingTheBotClearsThePreviousBotsRows` capturing the service's own logger and asserting one count per swept table is the right shape — `Upsert` returns an `Installation`, so there is nowhere else to read them. That "dropping the field fails it with `queued_outbound_messages_dropped = 0, want the 1 seeded row reported`" is the part that makes it worth having.

The race-window comment reads right now. Saying what healing looks like from the user's side — one message that gets no answer and no line saying why, until they re-bind — puts it next to the silence one item up, which is the comparison that makes the window's size legible rather than just stated.

Bohan-J:
Thanks for this, @ii-jj — and for carrying it through several rounds of review.

This is a real bug and the fix lands in the right place. Before the change, pointing an agent at a different bot kept the installation row and its id, so every dependent row keyed on `installation_id` — user bindings, p2p chat session bindings, queued task deliveries and queued outbound sends — survived into a bot whose anonymized userid namespace does not contain them. Clearing them inside the same transaction, before the upsert, is the right shape, and keeping bindings intact for a same-bot secret rotation is the right distinction to draw. The table coverage checks out too: the one bot-scoped table left alone, `channel_media_pending_object`, is documented in its own migration as diagnostics-only with no logic keyed on it, so leaving it is consistent with how you treat the audit trail.

Two things we looked at closely and are merging with, rather than holding the PR for:

- The residual window your comment in `installation.go` documents is real, but it is bounded by the commit rather than by supervisor teardown: inbound resolves the installation by `bot_id` (`wecom_resolvers.go`), so once the swap commits, a frame arriving on the old bot's still-live socket finds no row and is dropped. Against `main`, where every swap leaks all of the old bot's state permanently, this is a strict improvement — we would rather ship it than sit on it.
- The cleanup log line is emitted before `tx.Commit`, so a failed upsert or commit would roll the deletes back while the log still claims they happened. Cosmetic, and we will clean it up on our side.

We are filing a follow-up to move WeCom onto the same installation-identity model DingTalk already uses — an advisory lock on the logical `(workspace, agent, channel)` slot plus a fresh `installation_id` when the bot changes — which closes the remaining race properly. That is our work, not yours.

Merging now. Thanks again for the careful comments explaining *why* these rows are invalid rather than merely stale — that context is worth as much as the fix.

PR #7993, opened 2026-09-03
fix(auth): send the user to login when the session expires (MUL-7028)

Bohan-J:
## Problem

When a client's session expires, every action fails with `missing authorization`
in a toast and the app never leaves the workspace — the user is stuck on a screen
that no longer works, reading an error most people can't interpret.

The 401 handler dropped the stored token and nothing else:

```ts
// packages/core/platform/core-provider.tsx
onUnauthorized: () => {
  storage.removeItem("multica_token");
},
```

The auth store kept `user` set and `status: "authenticated"`, so neither shell's
login gate ever fired — desktop `App.tsx` (`user ? <DesktopShell/> : <DesktopLoginPage/>`)
and web `[workspaceSlug]/layout.tsx` (`if (!isAuthLoading && !user) router.replace(paths.login())`)
both wait on `user` going null. And because the token was now gone, every request
after the first went out with no `Authorization` header, so the server's auth
middleware answered `{"error":"missing authorization"}` — the message in the toast.

Only `AuthInitializer` handled a 401, and only at boot. Mobile already does the
right thing (`apps/mobile/app/_layout.tsx` → logout + `router.replace("/login")`);
web and desktop share the core provider that did not.

## Change

**`sessionExpired()` on the auth store** — an idempotent teardown that lands in
exactly the state a cold boot with a dead token reaches: clear the credential,
reset analytics, run `onLogout`, publish `unauthenticated`. No server round-trip;
the credential is already dead. Idempotent because a session dies once but a
screenful of in-flight requests each learn about it separately.

The api client's 401 hook drives it, and `AuthInitializer`'s boot rejection now
delegates to it too, so both dead-session paths converge on one implementation
instead of two that drifted.

**A reason on the login page.** Bouncing someone to login with no explanation
trades one confusing state for another. The notice is only shown when this client
actually presented a credential the server rejected — a live session, or a stored
token from an earlier one. A first visit's identity probe 401s as well, and
telling that person their session expired would be a lie.

Two follow-ons the newly-reachable path exposed:

- **Drop the query cache on an involuntary session end.** Every query runs
  `staleTime: Infinity`, so whoever signed in next — the same user, or a colleague
  on a shared machine — would render the dead session's data and never refetch it
  away. Explicit logout already cleared the cache (`views/auth/use-logout`); the
  involuntary path did not, and before this change it could not be reached.
- **Snapshot the CLI-authorize token before probing the cookie.** The probe is
  *expected* to 401 for a token-mode session, and the teardown clears the very key
  the localStorage fallback then looks for — so token-mode users could never
  authorize a CLI and landed on the email step instead. Pre-existing, but this
  change makes the teardown louder, and the fix is four lines in the same effect.

## Session end: what goes, what stays

An expiry erases exactly the client state an explicit logout erases — in-memory
drafts, per-workspace persisted state, the desktop tab layout, the
last-workspace cookie, the query cache — because the screen it lands on offers a
login form and the next person to use it is not guaranteed to be the one who was
just signed in. Sharing a workspace with them means the post-login stale-tab
validator finds nothing to prune, so nothing else would catch it. The teardown
lives in core's `platform/session-cleanup` and both paths call it.

The cost is that unsent drafts do not survive an expiry, same as a logout.
Keeping them would mean gating the restore on user identity, which the login
page cannot decide before it has one.

The daemon is the one thing an expiry leaves alone. It authenticates with its
own PAT, minted separately from the window's session token, and may be running
agent work; only this window has to sign in again. An account switch stays safe
because `daemon:sync-token` mints a fresh PAT and restarts the daemon whenever
the user id changes.

## What ends a session

Only a 401 on a credential we presented. `AuthInitializer`'s catch narrows to
`ApiError` with status 401 before rejecting anything, and the api client's
hook fires on `res.status === 401` — a rejected `fetch()` never reaches it. A
client that cannot reach the server (offline, DNS, 5xx, timeout) settles on
`recovering`, keeps its token, and retries with backoff; it is never signed
out. Regression tests cover offline, 500 and 504.

Client-state cleanup is keyed on *being* unauthenticated rather than on leaving
`authenticated`, because the worst leak never passes through `authenticated`:
the app starts, the stale token on disk is rejected at the identity probe, and
the previous session's drafts wait for whoever signs in next. That trigger
cannot sign anyone out either — reaching `unauthenticated` is itself the
decision to end the session, made only by the 401 above or by having no
credential at all.

The keys to remove are enumerated from storage, not from the workspace list:
a session rejected at the probe never loaded one. `StorageAdapter` gains an
optional `keys()`; workspace-scoped entries are `${base}:${slug}` and every
base is known, so the slugs no longer have to be.

## Testing

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` (638 files, all passing)
run locally. New coverage:

- `packages/core/auth/store.test.ts` — teardown, idempotence under a 401 burst,
  the expired flag cleared on re-login, and the two cases that must *not* claim
  expiry (first visit) vs must (stale stored token).
- `packages/core/api/client.test.ts` — real `ApiClient` + real store: a 401
  response ends the session, which is the wiring `CoreProvider` installs.
- `packages/core/platform/auth-initializer.test.tsx` — the query cache is dropped
  when a live session is rejected.
- `packages/views/auth/login-page.test.tsx` — the notice appears only when the
  session expired, and the CLI fallback survives the probe's teardown (verified
  failing without the fix).
- `packages/core/platform/session-cleanup.test.ts` — what an ended session
  leaves behind, across all three layers (persisted, in-memory, query cache),
  plus the reset-before-remove ordering the persist middleware demands.
- `apps/desktop/src/renderer/src/platform/session-teardown.test.ts` — logout
  stops the daemon, expiry does not (verified failing when it does).
- `packages/core/auth/store.test.ts` — a rejected token is dropped even when
  there was no session to end, so Desktop's deep link cannot leave an invalid
  credential at rest (verified failing without the fix).
- `packages/core/platform/auth-initializer.test.tsx` — a cold start with a
  stale token erases the previous session with an empty Query cache, and
  offline / 500 / 504 keep the session, the token and every draft. The
  cold-start case was verified failing against each half of its fix
  separately.
- `packages/core/platform/storage-cleanup.test.ts` — the sweep removes
  workspace-scoped keys for slugs the caller never knew, and reports rather
  than silently no-ops when an adapter cannot enumerate.

E2E was not run locally (needs a running stack); CI covers it.


Bohan-J:
Updated: the daemon no longer stops when the session expires — only this window has to sign in again. Pushed as `8164da8`.

`onSessionExpired` defaults to `onLogout`, so a shell only needs its own handler when part of its logout teardown is too destructive for an expiry it did not ask for. On desktop that's the daemon (it authenticates with its own PAT and may be running agent work) and the tab layout with it — the same person is usually about to sign straight back in, and if someone else signs in instead, `AppContent` already re-validates every tab group against the new user's workspace list. An account switch stays safe without stopping anything here, because it's handled at the other end: `daemon:sync-token` mints a fresh PAT and restarts the daemon whenever the user id changes.

One hole this exposed: the no-token boot exit in `AuthInitializer` ran the *logout* teardown. So the daemon would survive the expiry and then die on the next app start, when a missing token got read as "the user signed out". That path now runs the same teardown as every other unauthenticated exit.

Both desktop teardowns moved to `platform/session-teardown` with their side effects injected, so the difference between them is under test without needing an Electron window — the "leaves the daemon running" case was verified failing when expiry stops the daemon. `AuthInitializer` no longer takes `onLogout` at all now that every unauthenticated exit delegates to the store.

`pnpm typecheck` (9/9), `pnpm lint` (0 errors) and `pnpm test` (637 files) pass locally.

Bohan-J:
Both must-fixes addressed in `c559c89`.

**Must-fix 1 — cross-account residue.** Took the first option: an expiry now erases exactly what an explicit logout erases. The teardown moved out of `use-logout` into core's `platform/session-cleanup`, and both paths call it — in-memory drafts (with uploads aborted), per-workspace persisted keys for every workspace the session had, `multica_tabs`, `last_workspace_slug`, and the query cache. Desktop's expiry path now resets the tab store too, which leaves the daemon as the only difference between the two teardowns — the point of having split them.

You were right that "keeps the user's tabs" locked in an unproven claim; that test is gone. The `A → 401 → B` case is now covered at two layers: `session-cleanup.test.ts` pins what an ended session leaves behind across the persisted / in-memory / query-cache layers (including the reset-before-remove ordering the persist middleware demands, moved down from the logout suite so it has one home), and `auth-initializer.test.tsx` asserts a 401 on a live session is actually wired to it. Verified failing against the old narrow `qc.clear()`.

The cost is that unsent drafts don't survive an expiry now, same as a logout. The identity-gated alternative needs the previous user's workspace slugs to clear their persisted keys, and after a restart there's nothing left to enumerate them from — it'd need a persisted cleanup ledger to be sound, which seemed worse than matching logout.

**Must-fix 2 — credential survived the guard.** Dropping the rejected credential now happens first and unconditionally; the guard below it only suppresses the parts that actually end a session. Added the test you asked for — a `loginWithToken` that 401s leaves storage empty — verified failing with the guard back on top.

**Non-blocking.** PR description updated; the "Note for review" section contradicting `8164da8` is replaced with what the current code does.

`pnpm typecheck` (9/9), `pnpm lint` (0 errors), `pnpm test` (638 files) pass locally.

Bohan-J:
Cold-start leak fixed in `fc607e7a`.

**The two halves.** The trigger was keyed on leaving `authenticated`, which a cold start never does, and the sweep enumerated slugs from a Query cache that a rejected identity probe never fills. Both had to change: the cleanup is now keyed on *being* unauthenticated, and the keys come from storage rather than from the workspace list. `StorageAdapter` gains an optional `keys()` — workspace-scoped entries are `${base}:${slug}` and every base is known, so the slugs no longer have to be. That also catches workspaces the last session had already left, which the ledger approach would not.

I verified each half is load-bearing separately: reverting the trigger fails the cold-start test, and so does reverting the sweep.

You were right about the test gap — `never resolved a workspace list` asserted only the global `multica_tabs`, so it passed while every per-workspace key survived. It now seeds `multica_comment_drafts:acme` and `multica:chat:activeSessionId:acme` and asserts storage is empty.

**On the trigger being safe.** Widening it is the part worth scrutinising, so to state the invariant explicitly: reaching `unauthenticated` *is* the decision to end a session, and only two things get there — a 401 on a credential we presented, or having no credential at all. `AuthInitializer`'s catch narrows to `ApiError` with status 401 before rejecting, and the api client's hook fires on `res.status === 401`, which a rejected `fetch()` never reaches. Offline, DNS failure, 5xx and timeouts all settle on `recovering`, keep the token, and retry with backoff. That is now pinned by regression tests for offline / 500 / 504 asserting the session, the token and every draft survive — an unreachable server must never cost someone their login.

`pnpm typecheck` (9/9), `pnpm lint` (0 errors), `pnpm test` (638 files) pass locally.

PR #8008, opened 2026-09-03
MUL-7037 feat(cli): resolve custom property names in issue JSON

### LINKED ISSUE #8007: [Feature]: Resolve custom property names in CLI issue JSON
### Deployment type

self-host

### Area

CLI

### What do you want and why?

`multica issue list --output json` and `multica issue get --output json` return custom properties as an opaque map: definition UUID to option UUID. Labels on the same payload carry their names. Measured on a v0.4.39 server and CLI:

```
$ multica issue list --output json | jq '.issues[0] | {identifier, properties}'
{
  "identifier": "DEV-1",
  "properties": {
    "3237807a-6332-44f8-94d1-0223ad8d25eb": "462c028e-f5ef-4794-868e-0bd4bd429188",
    "dc3e070d-b36e-4fc5-84d4-0f3be27146ca": ["493ac4d0-22ae-4c15-9228-daca867d37f2", "7d9a805f-48cd-4f9c-b2cd-c8a69d0a9e5f"]
  }
}
```

To learn that this issue has `Impact = High`, a consumer has to call `multica property list` as well and join on both the definition id and the option id. For a sweep of 100 issues that is a second request plus a two-level join in every script. The CLI's main audience is coding agents, and this is the one place where properties are harder to consume than the labels they otherwise replace.

The resolved shape already exists in the CLI. `multica issue property list <id> --output json` prints one row per set property:

```
[{"property_id":"3237807a-...","name":"Impact","type":"select","value":"462c028e-...","display":"High"}]
```

`issue list` and `issue get` cannot emit it, so a consumer that wants names pays one extra request per issue.

A related wart: that per-issue command renders a `multi_select` value as one comma-joined string (`"iOS, Android"`), which is ambiguous as soon as an option name contains a comma. The CLI's own `issue property set --value` splits on commas, so such an option cannot even be set from the CLI today.

Add an opt-in flag, `--resolve-properties`, to `issue list` and `issue get` in JSON mode. When set, `properties` is rewritten from the id map into the same rows `issue property list` prints, so names, option labels and member names arrive with the issue. Omitting the flag leaves today's output unchanged, and table output is unaffected.

Details worth settling up front:

- One catalog request per command, shared with the existing `--property` and `--sort property:` fetch.
- `actor` and `multi_actor` values resolve through the same members lookup `issue property list` uses.
- A stored option id that is no longer in the definition falls back to the raw id, as the server's sort path already treats such values as null rather than failing.
- Archived definitions still resolve, since their values are preserved on the issue.
- Multi-value rows gain a `display_values` array beside the flat `display` string, so a comma in an option name is no longer ambiguous.

This composes with the open `--fields` proposal (#7946): the resolved rows stay under the `properties` key, so `--fields=identifier,properties --resolve-properties` needs no new field name as long as projection runs after resolution.

Firer:
## What does this PR do?

`multica issue list --output json` and `multica issue get` return custom properties as a map from definition id to option id. Labels on the same payload carry their names, so properties are the one field an agent cannot read without a second call to `property list` and a two-level join. The resolved shape already exists in `issue property list --output json`; this PR lets `issue list` and `issue get` emit it.

Two commits:

1. **`display_values` on property rows.** `issue property list` joins a `multi_select` value into one string (`"iOS, Android"`), which cannot be split back once an option name contains a comma. Multi-value rows now also carry `display_values`, the per-item names in stored order. `display` is unchanged, the key is absent on single-value rows, and table output is untouched.
2. **`--resolve-properties` on `issue list` and `issue get`.** In JSON output the flag replaces the `properties` id map with the rows `issue property list` prints: `property_id`, `name`, `type`, the stored `value`, `display`, `display_values` for multi-value types, and `archived` on an archived definition. Rows follow catalog order.

Design decisions, stated so they can be disagreed with:

- **Opt-in, JSON only.** Without the flag the output is unchanged (verified byte-identical against the installed v0.4.39 CLI, see below). With `--output table` the flag is a no-op and no extra request is made, matching how `--fields` (#7946) and `issue comment list --compact` treat table output.
- **`properties` is replaced, not joined by a sibling key.** The rows already carry `property_id` and the raw `value`, so a sibling would duplicate every id. Replacing keeps the top-level key set identical to `handler.IssueResponse`, which is what #7946's `--fields` whitelist is being asked to derive from, and keeps one shape for property values across `issue property list`, `issue property set`, `issue list` and `issue get`. The cost is honest: resolved rows are about 2.2x the raw bag per value (509 vs 1,116 compact bytes for a seven-value issue). The flag saves a request and a join, not bytes. A typed client that decodes `properties` as a map should not pass the flag.
- **Actor values resolve.** `actor` and `multi_actor` go through the members lookup `issue property list` already uses. Each command loads the member list once, on first use, and shares it between an actor `--property` filter, an actor `--value` on `issue property set` and the resolution, so the request is made at most once and only when something needs it. `--assignee` keeps its own lookup, since it also searches agents and squads. A member who has left the workspace keeps the raw `member:<uuid>` in `display`.
- **Failures are errors, not unresolved ids.** A failed catalog or member-list request fails the command (exit 1). A CLI that advertised resolution and printed ids at exit 0 is the kind of bug that let a select sort defect survive several releases in this subsystem. `issue property list` and `issue property set` keep their current behaviour (raw reference on a failed lookup); the second one has already written the value and must not report failure.
- **Stale option ids.** A stored option id that is no longer in the definition keeps the raw id in `value` and `display`. The server's sort path treats that case as NULL rather than failing (`selectPropertySortExpr`), so resolution degrades the same way. The server refuses to drop an option that is still in use, so this only arises from data written before that guard or outside the API.
- **Archived definitions resolve** and their rows carry `archived: true`; values are preserved on archive. The `--property` filter's rejection of archived definitions is deliberately not copied.
- **Catalog fetch order.** `--property` and `--sort property:` already fetch the catalog before the page; resolution reuses it. Otherwise the catalog is fetched after the page, and only when some bag holds a value, so a server old enough to lack the endpoint still serves pages with nothing to resolve. Because the pre-page fetch can be older than the page, a value whose definition is missing from the catalog is an error naming the issue and the id rather than a silently dropped row.
- **Client-side, not an API change.** An installed CLI must work against an older server, which a new query parameter cannot give it, and this needs no deploy or version gate.
- **Not in scope.** `issue search`, `issue children` and the create/update responses return the same bag; if the flag is wanted there later, the same name and helper apply.

## Related Issue

Closes #8007

Coordinates with two open PRs on the same output path:

- #7946 (`--fields`) whitelists top-level keys, and landed first. Because resolution rewrites `properties` in place, `--fields=identifier,properties --resolve-properties` needs no new field name. Resolved on rebase with projection running *before* resolution, the opposite of what this description originally proposed: resolving first spends a catalog and a member request on output the next line deletes, and fails the command outright on a definition id missing from the catalog — for a key the user asked to drop. Projecting first costs nothing, but on its own it would leave `--resolve-properties` silently doing nothing, so a `--fields` list that omits `properties` is now rejected up front instead, matching how `--fields` already rejects `assignee` rather than ignoring it. In `--output table` mode both flags stay no-ops and nothing is rejected. Covered by `TestRunIssueListResolvePropertiesWithFields`.
- #8001 (`--property` comparison operators) touches the same docs and skill files in different sections. Whichever lands second rebases.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/cmd/multica/cmd_property.go`: `issuePropertyValueRow` gains `display_values`; `issuePropertyDisplayValues` resolves per-item names and stays index-parallel with the stored array; `fetchActorPropertyNames` takes any number of bags and returns its error; new `resolveIssueProperties` rewrites the bags of a page in place; new `memberDirectory` loads the member list once per command and serves the actor filter, `issue property set --value` and the resolution from it.
- `server/cmd/multica/cmd_issue.go`: `--resolve-properties` on `issue list` and `issue get`, applied in their JSON branches; the matching rules of `resolveAssignee` move into `matchAssignee`, which takes an already fetched candidate list, so the actor property path applies them to the shared member list and `--assignee` is unchanged.
- `server/cmd/multica/cmd_property_test.go`: row shape for multi-value types, a comma in an option name, a non-string item, an unknown member; one member request for `issue property set` with an actor value.
- `server/cmd/multica/cmd_issue_resolve_properties_test.go`: every property type, empty and absent bags, stale option id, archived definition, unknown type, wrapper keys; no-flag output unchanged and `--resolve-properties=false` identical to the default; table output makes no catalog request; catalog and member failures error; one catalog fetch shared with `--property`; no catalog fetch when nothing needs resolving; unknown definition and non-object bag error; one member request per page; one member request shared by two actor filters and the flag; `issue get` with and without the flag.
- `CLI_AND_DAEMON.md`, `apps/docs/content/docs/cli.mdx` and the `.ja` / `.ko` / `.zh` siblings: the flag and the row shape.
- `server/internal/service/builtin_skills/multica-platform/references/issues.md`: the read pattern under Custom properties, next to the `--property` and `--sort property:` bullets; `server/internal/service/builtin_skills_test.go`: anchors for `--resolve-properties`, `display_values` and the guarantee that `value` keeps the stored ids.

## How to Test

1. `cd server && go test ./cmd/multica ./internal/service -count=1` (15 new tests in the CLI package, plus the skill anchors), `go vet ./cmd/multica`, `gofmt -l` on the changed files.
2. Against a server at v0.4.39 or newer, create a select, a multi_select with an option name containing a comma, a number, a checkbox, an actor and a text property; set them all on one issue, nothing on a second, and archive one definition after setting it. Then:
   ```
   multica issue list --output json --resolve-properties | jq '.issues[] | {identifier, properties: [.properties[]? | {name, display, display_values}]}'
   multica issue get <id> --resolve-properties
   multica issue list --output json | jq '.issues[0].properties'   # unchanged id map
   multica issue list --resolve-properties                          # table, unchanged
   ```
3. What I ran: a local API built from `5f0e0f229` (the commit after the v0.4.39 release entry), seeded as above plus a stale option id planted directly in the test database, since the API refuses to remove an option in use. The no-flag `issue list`, `issue get` and table output from this branch are byte-identical to the installed v0.4.39 CLI; `issue property list` differs only by the added `display_values`. With the flag: every type resolves, the archived definition resolves with `archived: true`, the empty bag becomes `[]`, the stale id surfaces raw, and `--property Impact=High --resolve-properties` makes one catalog request. A read-only check against a production v0.4.39 workspace gave the same result and the same no-flag byte identity.

## Checklist

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** I wrote the problem statement and the design constraints (opt-in, byte-identical when omitted, table untouched, mirror the server's stale-id and archived semantics, coordinate with #7946) and had Claude Code do the implementation. The plan went through a three-way review by Codex, Antigravity and a separate Claude session; that review is what turned a silent `member:<uuid>` fallback into an error, moved the catalog fetch after the page, and added the missing-definition check. The diff was reviewed by Codex and Antigravity (the Claude leg did not complete), and their findings tightened the docs wording and one test assertion. I verified the result myself against a local server and a production v0.4.39 workspace as described above.


Bohan-J (changes requested):
Thanks for this — the product direction is right. `properties` returning bare definition/option UUIDs means every script and agent has to re-implement the same id→name join, and making the resolved rows opt-in (raw map unchanged by default, table output untouched, `value` still carried alongside `display`) is the shape we'd have picked ourselves. `display_values` in particular is the correct answer to the comma ambiguity you raised in #8007: `"iOS, Android, tablets"` can't tell you whether that's two options or three.

Two things to fix before this can merge.

## 1. Rebase onto current `main` — the skill docs moved

`main` (#7968) merged `multica-working-on-issues` into `server/internal/service/builtin_skills/multica-platform/references/issues.md` and deleted the per-skill source maps. This PR still edits two files that no longer exist on `main`:

- `server/internal/service/builtin_skills/multica-working-on-issues/SKILL.md`
- `server/internal/service/builtin_skills/multica-working-on-issues/references/working-on-issues-source-map.md`

GitHub currently reports the PR as conflicting, and it's a modify/delete conflict on exactly these paths.

Please:

- Drop the `working-on-issues-source-map.md` hunk entirely. Re-adding that file would reintroduce the source-reference maintenance burden #7968 deliberately removed.
- Move the `--resolve-properties` contract into `references/issues.md`, under **Custom properties: typed workflow state**, next to the existing `issue list --property "Name=Value"` / `--sort property:` bullets.
- Pin it in `server/internal/service/builtin_skills_test.go`. The `references/issues.md` case in the anchor table is how every relocated contract is kept from silently rotting; please add anchors for `--resolve-properties`, `display_values`, and the guarantee that raw `value` is still present.

## 2. "Member list fetched once per command" isn't true when combined with actor filtering

`CLI_AND_DAEMON.md:527` (and the PR description) says:

> The catalog, and the member list when some issue holds an actor value, are fetched once per command

The catalog half holds — `runIssueList` passes the already-fetched catalog into `resolveIssueProperties`, and `TestRunIssueListResolvePropertiesSharesCatalogFetch` covers it. The member-list half does not:

- `--property "Reviewer=Ada"` goes `resolvePropertyFilterValue` → `resolveActorPropertyRef` → `resolveAssignee`, which issues a fresh `GET /api/workspaces/<id>/members` on every call, with no caching or sharing.
- `resolveIssueProperties` → `fetchActorPropertyNames` then fetches the member directory again.

So `multica issue list --property "Reviewer=Ada" --resolve-properties` makes two member requests, and N actor filter tokens make N+1 — each an extra round trip and an extra failure point. This is the performance bound #8007's triage specifically asked to have pinned down, so it shouldn't ship stated one way and behaving another.

Either fix is acceptable:

- **Preferred:** resolve the member directory once per command and share it between filter resolution and property resolution, so the documented bound becomes true.
- Otherwise: correct `CLI_AND_DAEMON.md` and the PR description to state the actual bound.

Either way, please add a test for the combination. `cmd_issue_property_filter_test.go` and `cmd_issue_resolve_properties_test.go` each cover one side, and `TestRunIssueListResolvePropertiesOneMembersRequestPerPage` asserts `membersCalls == 1` only when no actor filter is present — nothing currently exercises `--property <actor-property>` together with `--resolve-properties`, which is why the gap wasn't caught.

## Notes

- Targeted property-resolution tests pass, and CI was green on the current head — but that head predates #7968, so CI has to re-verify against the new baseline after the rebase.
- Everything else reads well: the implementation extends the existing `issue property list` row construction rather than adding a parallel resolver, and archived definitions, stale options, unknown types and failure semantics all have deliberate coverage.

Happy to take another look once the rebase and the member-request contract are sorted.

Firer:
Hi @Bohan-J thanks for the review!

Rebased onto `main`. The `SKILL.md` bullet now lives in `references/issues.md` under "Custom properties: typed workflow state", after the `--sort property:` bullet, and the source map hunk is gone. `TestPlatformSkillCoversPlatformContracts` anchors `--resolve-properties`, `display_values` and "`value` keeps the stored ids" in the `references/issues.md` case.

The member request is now shared. Each command gets one `memberDirectory` that loads the member list on first use; the actor `--property` filter, an actor `--value` on `issue property set`, and `--resolve-properties` all read from it. To do that without a second matcher, the ranking rules moved out of `resolveAssignee` into `matchAssignee`, which takes an already fetched candidate list. `resolveAssignee` still builds its candidates from members, agents and squads and calls the same function, so `--assignee` behaviour is unchanged and its existing tests pass as they were.

`TestRunIssueListActorFilterSharesMembersRequest` runs `--property Reviewer=Ada --property Owners=Ada --resolve-properties` and asserts one members request, the filter that was sent, and the resolved rows. `TestRunIssuePropertySetSharesMembersRequest` covers the set path, which had the same double fetch. `CLI_AND_DAEMON.md` now says the flag adds at most two requests, the catalog and the member list, each shared with the filter.

One thing I left alone: `--assignee` keeps its own lookup, since it also searches agents and squads. So `--assignee Ada --property Reviewer=Ada --resolve-properties` still makes two member requests, and the doc sentence is scoped to what the flag adds rather than the whole command. I can route `--assignee` through the directory as well if you would rather the bound be per command.

Bohan-J (changes requested):
Thanks — both must-fix items from the last round are properly resolved, and I verified them rather than taking the summary at face value.

- **Skill migration:** the source-map hunk is gone, the `--resolve-properties` contract lives in `references/issues.md` under "Custom properties", and `builtin_skills_test.go` anchors `--resolve-properties`, `display_values` and the raw-`value` guarantee.
- **Member directory:** genuinely fixed rather than documented around. `memberDirectory` loads once and serves the actor `--property` filter, `issue property set --value` and the resolution. I compared the `matchAssignee` extraction against the old `classify` bucket-by-bucket — UUID/short-id/email alias, then exact name, then substring — and `--assignee` behaviour is unchanged. A second reviewer walked the same path independently and agreed.
- **Request-count bound:** `--assignee X --property Reviewer=X --resolve-properties` is two member requests, and it does not grow with the number of actor filters or issues. The doc wording scoped to what the flag adds is accurate. Sharing with `--assignee` too is a fine follow-up, not a blocker.

## Heads-up: we rebased this branch for you

`main` had moved 17 commits past your 9/4 rebase and the PR was conflicting again, so — with maintainer edits enabled — we rebased onto `440ef8aa1` and force-pushed (`--force-with-lease` against your head each time, nothing of yours was lost). Your three commits are preserved with authorship intact. **You'll want to reset your local branch to the remote rather than pull.**

Two of the three conflicts were one-line-each-side. The third is the one you predicted.

## One commit added on top — and it does the opposite of what the description says

`--fields` (#7946) landed first, so the two flags now meet in the same JSON block of `runIssueList`. The description says:

> `--fields=identifier,properties --resolve-properties` needs no new field name, as long as projection runs after resolution. Whichever PR lands second should place the call in that order and add a test for the combination

We went the other way, and added `MUL-7037 fix(cli): reject --resolve-properties when --fields drops properties` to do it. Measuring both orderings:

| ordering | `--fields id,title --resolve-properties` | `--fields id,title,properties --resolve-properties` |
|---|---|---|
| projection after resolution | output discarded, but 1 catalog + 1 member request wasted | resolves correctly |
| projection before resolution | no requests | resolves correctly |

Resolution-first also fails the whole command when a bag holds a definition id missing from the catalog — for a key the user explicitly excluded. But projection-first alone leaves `--resolve-properties` silently doing nothing, which is the footgun `--fields` already refuses elsewhere by rejecting `assignee` instead of ignoring it.

So the commit projects first **and** rejects the combination up front when `--fields` omits `properties`, with tests for all three cases. The pair still works where it means something.

One correction against our own commit, caught in final review: the guard was initially unscoped and fired in table mode too, breaking your documented "with `--output table` the flag is a no-op" invariant. That is fixed in the current head (`11576dc5a`), with a regression test that asserts table output still prints its rows and makes no requests.

Bohan-J (approved):
Approving and merging — thank you for this one.

You filed a clear issue, brought the fix yourself, and then turned around two rounds of review feedback quickly and thoroughly: the skill-directory migration landed with proper test anchors, and the duplicate member lookup got a real fix — a shared `memberDirectory` with the ranking logic lifted into `matchAssignee` — rather than a doc edit papering over the bound. That refactor touches the assignee resolution every command in the CLI goes through, and it came out behaviour-preserving; two reviewers walked it independently and agreed. The care shows in the details too: `display_values` resolving the comma ambiguity, archived definitions still resolving, departed members keeping their raw reference rather than silently vanishing, and failures being errors instead of ids printed at exit 0.

Housekeeping on my side, so the record is straight:

- I rebased and force-pushed your branch twice (`main` had moved 17 commits), and added the `--fields` interaction commit on top. That commit's first version had a bug of my own — the guard fired in `--output table` mode too, breaking the no-op invariant you had documented. Caught in final review and fixed in `11576dc5a`, with a regression test.
- I also updated the `#7946` bullet in the description, since the branch ended up doing the opposite of what it proposed. My edit, not yours.

Nothing left on you. Merging now.

PR #8018, opened 2026-09-03
fix(server): keep the parent workdir on codex_semantic_inactivity retries (GH #7998)

### LINKED ISSUE #7998: Preserve workdir continuity when failed retries require a fresh agent session
## Summary

When the daemon auto-retries a task that ended in `codex_semantic_inactivity`, the retry child gets **both** a fresh provider session **and** a fresh workdir. The workdir discard is a separate decision that the rest of the code does not make, and it costs users their in-progress project state on the very failure class the poisoned-session guard exists to recover from.

## Invariant

> A resume-unsafe provider session does **not** imply a resume-unsafe workdir.

The session and the workdir are independently verified throughout the codebase:

- `server/internal/daemon/daemon.go` `reportTaskResult` always forwards `sessionID` and `workDir` as separate fields on both `CompleteTask` and `FailTask`.
- `server/internal/daemon/daemon.go` `shouldRetryWithFreshSession` (in-turn recovery) clears only `task.PriorSessionID` / `execOpts.ResumeSessionID`; it never touches the workdir.
- `server/internal/daemon/daemon.go` `gateResumeToReachableSession` already evaluates "is the prior session reachable from this run?" independently of the workdir match.
- The terminal `UPDATE` in `server/pkg/db/queries/agent.sql` (`UpdateAgentTaskOnComplete`/`UpdateAgentTaskOnFail`) writes `session_id` and `work_dir` from independent parameters.
- `GetLastTaskSession` and `GetLastChatTaskSession` filter on the poisoned reason for the **session**; the workdir they return is just whatever the most recent surviving row recorded.

Only the retry-creation path violates the invariant.

## Root cause

`server/pkg/db/queries/agent.sql` `CreateRetryTask` (lines 570-571 on current `main`, commit `5f0e0f2`):

```sql
CASE WHEN p.failure_reason IS NOT DISTINCT FROM 'codex_semantic_inactivity' THEN NULL ELSE p.session_id END,
CASE WHEN p.failure_reason IS NOT DISTINCT FROM 'codex_semantic_inactivity' THEN NULL ELSE p.work_dir END,
```

Both columns are NULLed together. The session NULL is correct (`force_fresh_session=true` on the next line reflects the same intent). The workdir NULL is not — the retry child loses the prior `work_dir` even when the directory itself is still on disk and valid.

The existing test `TestCreateRetryTaskFreshensCodexSemanticInactivity` in `server/cmd/server/rerun_session_test.go` (lines 587-589) locks this behavior in:

```go
if child.WorkDir.Valid {
    t.Fatalf("expected retry child to drop poisoned work_dir, got %q", child.WorkDir.String)
}
```

That assertion encodes the bug as a contract.

## Proposed minimal patch

Scope: this fix touches only the `codex_semantic_inactivity` auto-retry path. Other poisoned reasons (`iteration_limit`, `agent_fallback_message`, `api_invalid_request`, `codex_resume_oversized`, `agent_error.context_overflow`) do **not** have a NULL-on-failure clause in `CreateRetryTask` today, so they already inherit workdir. The broader policy audit (whether some of those *should* also be retired, and whether workdirs should be independently validated before reuse) is a follow-up — see "Follow-up" below.

```diff
--- a/server/pkg/db/queries/agent.sql
+++ b/server/pkg/db/queries/agent.sql
@@ -567,8 +567,8 @@ SELECT
   p.trigger_comment_id, p.coalesced_comment_ids, p.trigger_summary, p.context,
   CASE WHEN p.failure_reason IS NOT DISTINCT FROM 'codex_semantic_inactivity' THEN NULL ELSE p.session_id END,
-  CASE WHEN p.failure_reason IS NOT DISTINCT FROM 'codex_semantic_inactivity' THEN NULL ELSE p.work_dir END,
+  p.work_dir,
   p.attempt + 1, COALESCE(sqlc.narg(max_attempts)::int, p.max_attempts), p.id,
   p.failure_reason IS NOT DISTINCT FROM 'codex_semantic_inactivity',
   p.is_leader_task,
```

The session-`NULL` clause and the `force_fresh_session` flag both stay exactly as they are; only the workdir column reverts to "carry the parent's workdir through."

Update the existing test to assert the new contract:

```diff
--- a/server/cmd/server/rerun_session_test.go
+++ b/server/cmd/server/rerun_session_test.go
@@ -583,9 +583,11 @@ func TestCreateRetryTaskFreshensCodexSemanticInactivity(t *testing.T) {
 	if child.SessionID.Valid {
 		t.Fatalf("expected retry child to drop poisoned session_id, got %q", child.SessionID.String)
 	}
-	if child.WorkDir.Valid {
-		t.Fatalf("expected retry child to drop poisoned work_dir, got %q", child.WorkDir.String)
+	if !child.WorkDir.Valid || child.WorkDir.String != "/tmp/codex-stuck" {
+		t.Fatalf("expected retry child to inherit parent work_dir, got %+v", child.WorkDir)
 	}
 	if !child.ForceFreshSession {
 		t.Fatal("expected retry child to force a fresh session")
 	}
```

After `make sqlc` regenerates `pkg/db/queries/*.sql.go`, no other Go code changes — the daemon already treats `task.PriorWorkDir` and `task.PriorSessionID` as independent fields, and `execenv.Reuse` plus `gateResumeToReachableSession` already handle the case where the inherited workdir is no longer reachable (GC'd, deleted, symlink drift, etc.) by transparently falling back to `Prepare` and dropping only the unreachable session.

## Verification

- `make sqlc` — regenerate sqlc bindings.
- `go test ./cmd/server/... -run 'TestCreateRetryTask'` — updated test passes.
- Spot-check that `gateResumeToReachableSession` logs `"dropping prior session: session store not reachable from this run"` rather than rebuilding a fresh workdir if the inherited workdir is stale — it already does, because `task.PriorSessionID = ""` is the only field it mutates in that branch.

## Tests requested in this issue

- `TestCreateRetryTaskFreshensCodexSemanticInactivity`: update as above (workdir retained).
- New: `TestCreateRetryTaskRetainsWorkdirForResumeSafeFailure` — confirm an `iteration_limit`/`agent_fallback_message`/`api_invalid_request` retry keeps both pointers (regression net for the sibling reasons).
- New: end-to-end claim-side test that the daemon's first claim on the retry child does **not** invoke `execenv.Prepare` (workdir reused via `Reuse`) when the parent's workdir still exists, even though `force_fresh_session=true`.
- New: when the parent's workdir has been GC'd between the original failure and the retry claim, the daemon logs the existing `"session store not reachable from this run"` path and falls back to a fresh workdir without losing the resume-pointer abstraction (already covered by `shouldReusePriorWorkdir` + `gateResumeToReachableSession`, but worth a regression test).

## Broader follow-up (not required for this fix)

The repo's recovery policy is currently binary: `unresumable → new session`. A staged policy worth considering once enough of the provider SDKs expose enough control:

1. `try_resume_original`
2. `try_resume_with_repair` — sanitize known-invalid empty transcript messages if the provider/runtime supports it; for `AuthMethodUnresolved` re-resolve provider/auth config once before abandoning the session.
3. `fallback_create_new_session`
4. Preserve the prior workdir independently throughout, replacing it only when the workdir itself is independently invalid.

The transcript-repair step is not currently possible from Multica's side because the daemon only holds an opaque session id for most providers; framing it as a future architecture option rather than a near-term fix.

## Acceptance criteria for this issue

1. Confirm the exact current `CreateRetryTask` SQL on `main` matches the snippet above (or update the issue if the predicates have moved).
2. Apply the diff in `server/pkg/db/queries/agent.sql` and update `TestCreateRetryTaskFreshensCodexSemanticInactivity`.
3. Run `make sqlc` and the targeted Go tests above.
4. Decide whether to ship the test-list additions as part of this fix or as a follow-up PR.

## Environment

- Verified against `multica-ai/multica` `main` at commit `5f0e0f2293293fd6e7a9b37c4a3685d0a9ada8ef` ("fix(daemon): stop rejecting preparation fields from an older parent (MUL-7029) (#7994)").
- Filed via `gh issue create` from a local checkout because the upstream ChatGPT integration has read-only access to this repo.

makesomethingshit:
## What does this PR do?

When a task fails with `codex_semantic_inactivity`, `CreateRetryTask` discarded the parent's provider session **and** its `work_dir` together, then forced a fresh session. The problem: that failure reason tells us the provider *session* is unsafe to resume — it says nothing about the filesystem state. A resume-unsafe session does not imply a resume-unsafe workdir, so this PR stops throwing away the user's already-progressed worktree state along with the poisoned session.

The fix keeps the existing retry semantics intact: the poisoned `session_id` is still dropped and `force_fresh_session` stays `true`, but the retry child now **inherits the parent's `work_dir`**. The claim handler is extended so an automatic retry (a `retry_of_task_id` child with `force_fresh_session`) forwards that `work_dir` to the daemon as `PriorWorkDir` with an empty `PriorSessionID` for both issue and chat tasks, so the daemon reuses the existing filesystem and starts a fresh provider session — no workdir is thrown away, and no poisoned session is resumed. When the workdir is stale or unusable, the daemon's existing reuse validation (`shouldReusePriorWorkdir` → `execenv.Reuse`) declines and falls back to a fresh `Prepare` — no new filesystem-existence check or retry logic is added.

## Related Issue

Closes https://github.com/multica-ai/multica/issues/7998

## Type of Change

## Changes Made

- `server/pkg/db/queries/agent.sql` — `CreateRetryTask`: for `codex_semantic_inactivity`, keep nulling `session_id` and forcing a fresh session, but remove the `CASE` that also nulled `work_dir` so the retry child inherits `p.work_dir`. No other failure-reason path is touched. Query doc comment updated to state the new contract.
- `server/pkg/db/generated/agent.sql.go` — regenerated via `sqlc generate` (v1.31.1); diff is confined to the SQL change above (embedded query text + doc comment).
- `server/internal/handler/daemon.go` — claim handler: issue and chat automatic-retry routes now forward the child's own `task.WorkDir` as `resp.PriorWorkDir` when `retry_of_task_id` is set and `force_fresh_session=true`, while leaving `PriorSessionID` empty. The existing `shouldReusePriorWorkdir` provenance validation and the fresh-`Prepare` fallback — and `execenv.Reuse`'s "empty means a fresh thread" rule on the empty session — are reused unchanged; manual-rerun and follow-up resume semantics are untouched.
- `server/cmd/server/rerun_session_test.go` — `TestCreateRetryTaskFreshensCodexSemanticInactivity` now asserts the full DB contract: `session_id` dropped, `work_dir` inherited (`/tmp/codex-stuck`), `force_fresh_session` true, attempt 2.
- `server/internal/handler/daemon_test.go` — issue and chat claim-layer regressions use a `codex_semantic_inactivity` parent → real `CreateRetryTask` → actual claim and assert `PriorWorkDir == parent.WorkDir` with `PriorSessionID == ""` (fresh provider session over the preserved worktree).

## How to Test

1. `cd server && make sqlc` — regenerate; confirm only `agent.sql.go` changes (LF/CRLF churn from a local Windows regeneration is not part of the diff).
2. `go vet ./...` — passes.
3. `go build ./cmd/server` — passes.
4. `go test -count=1 ./cmd/server/ -run 'TestCreateRetryTask' -v` — requires a live PostgreSQL (schema migrated to latest); both `TestCreateRetryTaskFreshensCodexSemanticInactivity` and `TestCreateRetryTaskKeepsOrdinaryTimeoutSession` pass.
5. `go test ./internal/handler -run 'TestClaimTask_(AutoRetryOffersParentWorkdirForFreshSession|ChatAutoRetryOffersParentWorkdirForFreshSession)' -count=1` — requires a live PostgreSQL; the issue and chat claim-layer regressions both pass. These cover an automatic retry parent that failed with `codex_semantic_inactivity`, a real `CreateRetryTask` child, and the final claim response (`PriorWorkDir == parent.WorkDir`, `PriorSessionID == ""`). In a live run, watch the daemon reuse the existing worktree (`shouldReusePriorWorkdir` passes so `execenv.Reuse` mounts the same workdir) while starting a fresh provider session. To sanity-check the fallback, delete the recorded workdir and repeat — the daemon declines reuse and prepares a fresh environment.

## Checklist

## AI Disclosure

**AI tool used:** Multica Agent (multi-agent: worker-opencode implementation, Worker-codex independent cross-check, cheap-models pre-flight/review)

**Prompt / approach:**
Implementation spec (upstream #7998) was executed as a minimal production change: drop the `work_dir` CASE in `CreateRetryTask` (keep `session_id` NULL + `force_fresh_session`), then extend the claim handler's issue and chat automatic-retry routes to forward the child's preserved `work_dir` as `PriorWorkDir` with an empty `PriorSessionID`, so the daemon reuses the worktree and starts a fresh provider session. `sqlc` regenerated the binding; the existing DB-level regression test was flipped to pin the new contract and issue/chat claim-layer regressions prove the workdir flows into the daemon on automatic retries. Two independent agents cross-checked the diff, SQL semantics, generated output, test results and the daemon reuse-path trace; maintainer re-review was requested after that cross-check.

### DIFF
--- a/server/cmd/server/rerun_session_test.go
+++ b/server/cmd/server/rerun_session_test.go
@@ -552,6 +552,13 @@ func TestGetLastTaskSessionExcludesCodexSemanticInactivity(t *testing.T) {
 	}
 }
 
+// TestCreateRetryTaskFreshensCodexSemanticInactivity pins the full retry
+// contract for Codex semantic inactivity timeouts: the poisoned provider
+// session is discarded and a fresh session is forced, but the parent's
+// work_dir is retained — a resume-unsafe provider session does not imply a
+// resume-unsafe workdir (GH #7998). When the stale workdir itself is
+// unusable, the daemon's reuse validation still falls back to a fresh
+// Prepare.
 func TestCreateRetryTaskFreshensCodexSemanticInactivity(t *testing.T) {
 	if testPool == nil {
 		t.Skip("no database connection")
@@ -584,8 +591,8 @@ func TestCreateRetryTaskFreshensCodexSemanticInactivity(t *testing.T) {
 	if child.SessionID.Valid {
 		t.Fatalf("expected retry child to drop poisoned session_id, got %q", child.SessionID.String)
 	}
-	if child.WorkDir.Valid {
-		t.Fatalf("expected retry child to drop poisoned work_dir, got %q", child.WorkDir.String)
+	if !child.WorkDir.Valid || child.WorkDir.String != "/tmp/codex-stuck" {
+		t.Fatalf("expected retry child to inherit parent work_dir, got %+v", child.WorkDir)
 	}
 	if !child.ForceFreshSession {
 		t.Fatal("expected retry child to force a fresh session")
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -2619,6 +2619,23 @@ func (h *Handler) buildClaimedTaskResponse(r *http.Request, task *db.AgentTaskQu
 				)
 				resp.PriorSessionResumeUnavailable = true
 			}
+		} else if task.RetryOfTaskID.Valid && task.ForceFreshSession {
+			// Auto-retry child with a forced fresh session (codex_semantic_inactivity
+			// is the only reason CreateRetryTask sets the flag): the poisoned
+			// provider session must NOT be resumed, but the child inherits the
+			// parent's work_dir (GH #7998) — offer it so the daemon's reuse
+			// validation (shouldReusePriorWorkdir) can continue in the same
+			// filesystem/worktree while starting a fresh provider session. A
+			// resume-unsafe provider session says nothing about the filesystem;
+			// when the stale workdir fails reuse validation the daemon falls back
+			// to a fresh Prepare (best-effort, same contract as the rerun path).
+			// PriorSessionID stays empty: a forced-fresh auto-retry never resumes a
+			// session through this path. Auto-retries without the flag (e.g. plain
+			// timeout) still take the GetLastTaskSession branch below, which picks
+			// the freshest resume-safe row rather than this child's own pointer.
+			if task.WorkDir.Valid {
+				resp.PriorWorkDir = task.WorkDir.String
+			}
 		} else if !task.ForceFreshSession {
 			// Non-rerun follow-up on the same issue: resume the most recent
 			// (agent, issue) session so the agent keeps the issue's conversation
--- a/server/internal/handler/daemon.go
+++ b/server/internal/handler/daemon.go
@@ -2730,6 +2730,11 @@ func (h *Handler) buildClaimedTaskResponse(r *http.Request, task *db.AgentTaskQu
 			}
 		}
 		projectCtx.applyTo(&resp)
+		// Auto-retry children keep their filesystem even when the provider
+		// session must start fresh (codex_semantic_inactivity).
+		if task.RetryOfTaskID.Valid && task.ForceFreshSession && task.WorkDir.Valid {
+			resp.PriorWorkDir = task.WorkDir.String
+		}
 		if !task.ForceFreshSession && !task.ChannelContextRevision.Valid {
 			// Resume chat sessions only when the stored pointer was produced
 			// by the same runtime as the claiming task. When the chat_session
--- a/server/internal/handler/daemon_test.go
+++ b/server/internal/handler/daemon_test.go
@@ -3170,6 +3170,50 @@ func TestClaimTask_AutoRetryOffersParentWorkdirForFreshSession(t *testing.T) {
 	}
 }
 
+func TestClaimTask_ChatAutoRetryOffersParentWorkdirForFreshSession(t *testing.T) {
+	if testHandler == nil {
+		t.Skip("database not available")
+	}
+	ctx := context.Background()
+
+	agentID, runtimeID, daemonID := createRuntimeGuardAgent(t, ctx)
+	sessionID := dbfx.ChatSession(t, agentID, testutil.Cols{
+		"title":      "chat auto-retry workdir reuse fixture",
+		"session_id": "chat-pointer-session",
+		"work_dir":   "/tmp/chat-pointer-workdir",
+		"runtime_id": runtimeID,
+	})
+	dbfx.Insert(t, "chat_message", testutil.Cols{
+		"chat_session_id": sessionID,
+		"role":            "user",
+		"content":         "retry this chat turn",
+	})
+	parentID := dbfx.Task(t, agentID, testutil.Cols{
+		"runtime_id":      runtimeID,
+		"chat_session_id": sessionID,
+		"status":          "failed",
+		"failure_reason":  "codex_semantic_inactivity",
+		"session_id":      "poison-chat-session",
+		"work_dir":        "/tmp/auto-retry-chat-parent-workdir",
+	})
+	child, err := db.New(testPool).CreateRetryTask(ctx, db.CreateRetryTaskParams{ID: parseUUID(parentID)})
+	if err != nil {
+		t.Fatalf("CreateRetryTask: %v", err)
+	}
+	if uuidToString(child.ChatSessionID) != sessionID || child.WorkDir.String != "/tmp/auto-retry-chat-parent-workdir" || child.SessionID.Valid || !child.ForceFreshSession {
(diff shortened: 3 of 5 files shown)


Bohan-J:
Thanks for working on this, @makesomethingshit. The approach here is what landed in #8291, which is now merged and closes #7998: keep `work_dir` in `CreateRetryTask`, and hand it back from the claim handler for automatic retries that must start fresh.

#8291 also adds one thing on top. When the daemon reuses a prior workdir without resuming the session, it now warns the fresh session, so the agent doesn't re-run `multica repo checkout` and reset the checkout being kept. The server only offers the workdir to daemons that advertise that warning.

PR #8019, opened 2026-09-04
fix(codex): force the supported remote compaction path

### LINKED ISSUE #8000: [Bug]: Error running remote compact task
### Deployment type

Official App

### Area

Web app

### What happened?

Error running remote compact task: unexpected status 404 Not Found: {"detail":"Not Found"}, url: https://chatgpt.com/backend-api/codex/responses/compact, cf-ray: a35507973eee3d57-SJC, request id: 9eed88ee-822d-446e-9b73-df49d56fe7e0

### Steps to reproduce

Error running remote compact task

### Screenshots (optional)

_No response_

```shell

```

Scorbunny2:
## Summary

- force `remote_compaction_v2` for daemon-managed built-in Codex versions that support it
- remove lower-priority launch-prefix and argument overrides that would re-enable the retired `/responses/compact` path
- leave custom Codex-compatible executables, older Codex versions, and the shared user config untouched
- add focused regression coverage for stale config overrides and the supported version floor

Fixes #8000

Upstream context: openai/codex#42468

## Test plan

- `go test ./pkg/agent -run 'TestEnforceCodexRemoteCompactionV2OverridesStaleConfig|TestCodexRemoteCompactionV2VersionFloor|TestBuildCodexArgsExplicitFastOverridesLowerPriorityDisable' -count=1`\n- `go test ./pkg/agent -count=1`\n- `go vet ./pkg/agent`\n- `go test ./internal/daemon -run '^$' -count=1`\n- `codex app-server --listen stdio:// --enable remote_compaction_v2 --help`

### DIFF
--- a/server/pkg/agent/codex.go
+++ b/server/pkg/agent/codex.go
@@ -34,9 +34,11 @@ var codexBlockedArgs = map[string]blockedArgMode{
 }
 
 const (
-	codexFastServiceTier     = "priority"
-	codexStandardServiceTier = "default"
-	codexFastModeFeature     = "fast_mode"
+	codexFastServiceTier              = "priority"
+	codexStandardServiceTier          = "default"
+	codexFastModeFeature              = "fast_mode"
+	codexRemoteCompactionV2Feature    = "remote_compaction_v2"
+	minCodexRemoteCompactionV2Version = "0.146.0"
 )
 
 // codexStderrTailBytes bounds the stderr tail captured for inclusion in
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for the unusually thorough investigation here — the diagnosis is correct, and the write-up in openai/codex#42468 saved us a lot of time. We're going to decline this change, but not because the analysis is wrong.

To be clear about what we're *not* saying: we have no objection in principle to overriding a Codex feature flag in daemon-managed sessions. We already do exactly that for `fast_mode` in this same file, including stripping lower-priority `--disable` and `-c features.*=false` overrides. So the mechanism your PR generalizes is one we already ship, and the refactor into `stripCodexFeatureConflicts` is clean.

The distinction that decides it is *why* the override exists:

- `fast_mode` enforces **our own product semantics**. A user selected (and paid for) a service tier, and we're obligated to guarantee the process actually runs on it. Overriding a conflicting local config is us honoring our own contract.
- `remote_compaction_v2` is purely a **workaround for an upstream defect**. It carries no Multica semantics, and upstream has already stated its intent to fix it — either marking the flag `Stage::Removed` or adding a 404 fallback. Once either lands, this override becomes permanently dead code we have to carry, and worse, it actively blocks anyone who later has a legitimate reason to disable the flag.

Adding a version-gated override plus a removal condition that nobody will remember to act on is a bad trade for a bug whose user-facing fix is deleting one line from `~/.codex/config.toml`. That fix also propagates to Multica task sessions on its own — we refresh the copied Codex config from the shared source between runs, so there's nothing for the user to do on our side.

There is a real gap on our side, though, and it's the one your report exposed: this failure is completely unclassified in our pipeline, so the user sees raw Codex error text with no indication of which file or line to change. We're tracking that separately — turning an opaque 404 into an actionable message is squarely our responsibility, and it's the part that would still matter after upstream fixes the flag.

Genuinely good work on the root cause. If you're interested in the error-classification piece instead, that one we'd take.

PR #8020, opened 2026-09-04
MUL-7038: fix(repocache): restore resolved base commit before isolated checkout

### LINKED ISSUE #8011: [Bug]: isolated repo checkout loses the fetched tip when the shared bare cache is shallow
### Deployment type

Managed Linux daemon runtime using isolated Git metadata for Codex.

The observed build is a downstream build at commit `4d06a77`. The relevant `gitFetchContext`, `localCloneArgs`, and `createIsolatedCheckoutContext` implementations are byte-for-byte identical to current upstream `main`.

### What happened?

`multica repo checkout <url> --ref master` consistently failed after spending about two minutes repacking a multi-gigabyte shared cache:

```text
checkout failed: create isolated checkout: git checkout --detach:
fatal: reference is not a tree: <CURRENT_MASTER_SHA>
```

The shared bare cache was shallow. A fetch correctly updated `refs/remotes/origin/master`, and the new commit object existed in the cache, but the stale snapshot under `refs/heads/master` was not updated:

```text
git rev-parse --is-shallow-repository  true
refs/heads/master                      <OLD_MASTER_SHA>
refs/remotes/origin/master             <CURRENT_MASTER_SHA>
git cat-file -t <CURRENT_MASTER_SHA>   commit
```

The isolated checkout path then ran the equivalent of:

```text
git clone --local --no-checkout --no-tags --origin multica-cache <cache> <checkout>
git -C <checkout> checkout --detach <CURRENT_MASTER_SHA>
```

Git ignores `--local` for a shallow source and transfers commits reachable from the source's fetched heads. Since the current tip only has a remote-tracking ref, it is omitted from the isolated clone. The following checkout therefore fails with `reference is not a tree`.

A direct single-branch shallow clone from the real remote succeeded, ruling out repository access and a missing remote object.

The origin of the field cache's shallow state is not yet known. Its `shallow` file appeared after the cache was originally created, so this report does not claim that current Multica intentionally creates shallow caches. The bug is that an existing shallow cache is accepted, refreshed, and then used in a way that deterministically produces an unusable isolated checkout.

### Minimal reproduction

This standalone script uses only temporary local repositories and reproduces the same Git sequence without private repository data:

```bash
set -eu

tmp=$(mktemp -d)
trap 'rm -rf -- "$tmp"' EXIT

origin="$tmp/origin.git"
work="$tmp/work"
cache="$tmp/cache.git"
checkout="$tmp/checkout"

git init -q --bare "$origin"
git clone -q "$origin" "$work"
git -C "$work" config user.name repro
git -C "$work" config user.email repro@example.com

git -C "$work" commit -q --allow-empty -m A
git -C "$work" branch -M master
git -C "$work" push -q -u origin master
old_tip=$(git -C "$work" rev-parse HEAD)
git --git-dir="$origin" symbolic-ref HEAD refs/heads/master

# Model an existing shallow shared bare cache.
git clone -q --bare --depth=1 "file://$origin" "$cache"
git --git-dir="$cache" config --unset-all remote.origin.fetch || true
git --git-dir="$cache" config --add remote.origin.fetch \
  '+refs/heads/*:refs/remotes/origin/*'

# Advance the remote and run the same fetch layout used by repocache.
git -C "$work" commit -q --allow-empty -m B
git -C "$work" push -q origin master
new_tip=$(git -C "$work" rev-parse HEAD)
git --git-dir="$cache" fetch -q origin

test "$(git --git-dir="$cache" rev-parse refs/heads/master)" = "$old_tip"
test "$(git --git-dir="$cache" rev-parse refs/remotes/origin/master)" = "$new_tip"
git --git-dir="$cache" cat-file -e "$new_tip^{commit}"

# Model createIsolatedCheckoutContext.
git clone --local --no-checkout --no-tags \
  --origin multica-cache "$cache" "$checkout"
git -C "$checkout" checkout --detach "$new_tip"
```

Actual final output:

```text
warning: source repository is shallow, ignoring --local
warning: --local is ignored
fatal: reference is not a tree: <NEW_TIP>
```

The reproduction is deterministic on Linux and completes in under a second.

### Expected behavior

The daemon should not proceed into an isolated checkout that cannot contain `baseCommit`. Any of these would be safe outcomes:

- detect a shallow cache and fetch `baseCommit` into the isolated checkout before checking it out;
- rebuild or repair the cache into a supported shape;
- reject the cache early with an actionable error explaining how to recover it.

It should not repack gigabytes and then fail with a misleading missing-tree error.

### Suggested regression test

Add an isolated-checkout test with a shallow bare cache where:

1. `refs/heads/master` points to commit A;
2. `refs/remotes/origin/master` points to commit B;
3. B exists in the cache but is not reachable from any local head;
4. `createIsolatedCheckoutContext(..., baseCommit=B)` must produce a checkout containing B.

### Related but not duplicate

- #6153 / #7856 handles requested refs or commits that were never fetched into the cache. Here the commit is already present and resolves successfully; it is lost during the subsequent shallow local clone.
- #5772 repairs promisor configuration for blobless partial clones. A shallow cache is not a promisor partial clone and fails earlier with a missing commit.
- #5639 / #5762 discusses whether Multica should create shallow caches. This report only asks the daemon to detect or recover a shallow cache that already exists.

makesomethingshit:
## What does this PR do?

`Cache.CreateWorktree` with `IsolatedGitMetadata: true` can silently lose the commit it just resolved as its base when the shared bare cache is shallow. With a shallow source, Git ignores `git clone --local` and clones over the pack protocol instead; such a clone transfers only the objects reachable from the heads it fetches (`refs/heads/*`), so a tip that arrived through the cache's remote-tracking refspec (`refs/remotes/origin/*`) is absent from the fresh checkout even though it resolved in the cache moments earlier. The subsequent `git checkout --detach <sha>` then fails with `fatal: reference is not a tree` / `unable to read tree` (multica-ai/multica#8011).

This change makes isolated checkout creation enforce the resolved-base-commit invariant: after the clone and before the detached checkout, the checkout's object store must actually contain the resolved base commit; if it does not, fetch exactly the resolved ref from the local bare cache and re-verify. Recovery is bounded to the requested base ref, and the failure mode refuses to substitute the stale head.

**Non-goals** (intentionally out of scope for this PR):

- No cache policy change — how the shared cache originally became shallow is not investigated or changed here
- No full unshallow — caches are not unshallowed as part of this fix
- No refspec widening — the isolated checkout's standing fetch refspec is unchanged
- No shared-cache ref rewrite — the cache's stale `refs/heads/*` head is left untouched
- No UI / schema change — no user-facing, API, or database surface is touched

## Related Issue

Fixes #8011

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/repocache/cache.go`: new `ensureBaseCommitInCheckoutContext` — called from `createIsolatedCheckoutContext` directly before `git checkout --detach <baseCommit>`. It verifies the fresh clone's object store contains the resolved base commit with a single `rev-parse --verify <sha>^{commit}` (one cheap call on the healthy path, where `--local` hard-links the full object store, so healthy behavior is unchanged). If the commit is missing, it fetches exactly the resolved ref from the **local bare cache** (`git fetch --no-tags <cache> <baseRef>` — single ref, FETCH_HEAD only; no network origin, no shared-cache ref rewrite, no unshallow, no refspec widening) and re-verifies. If the local cache cannot serve the commit either, it fails loudly with an error that names the resolved commit, the shallow cache path, and the failed recovery — it never falls back to the stale head.
- `server/internal/daemon/repocache/shallow_cache_test.go`: new regression test reproducing the exact stale-head/shallow-cache topology from #8011: bare origin → working clone → commit A → shared bare cache seeded as a depth-1 shallow clone → modern remote-tracking layout → advance `origin/master` with commit B → fetch through the production cache-fetch path. Mandatory preconditions are asserted before the code under test runs (cache is shallow; `refs/heads/master == A`; `refs/remotes/origin/master == B`; `B != A`; `B^{commit}` resolvable in the cache). The real `CreateWorktree(IsolatedGitMetadata: true)` path (the mode exercised by Linux/Windows Codex isolated checkouts) is used, and the assertions cover: the operation succeeds; checkout HEAD == B; the content introduced by B is visible; and the shared cache's stale `refs/heads/master` was **not** rewritten by the repair. The test fails on unmodified upstream `main` at `git checkout --detach` with the missing commit and passes with this change.

## How to Test

1. From `server/`: `go test ./internal/daemon/repocache/ -run 'TestCreateIsolatedCheckoutFromShallowCacheResolvesFetchedTip' -count=1 -v` — fails on unmodified `main`, passes with this change.
2. `go vet ./internal/daemon/repocache/...`
3. `git diff --check`

## Checklist

- [x] I have run tests locally and they pass (focused regression suite; the 6 tests failing on this Windows host fail identically on unmodified `main` — pre-existing Windows-only `MAX_PATH` test-infra issue, unrelated to this PR; CI runs this package on ubuntu/macos)
- [ ] If this change affects the UI, I have included before/after screenshots (no UI change)
- [ ] I have updated relevant documentation to reflect my changes (no docs changed)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) (not applicable)
- [ ] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (not applicable)

## AI Disclosure

**AI tool used:** Multica Agent (worker-opencode + multica-agent)

**Prompt / approach:** The worker-opencode coding agent, orchestrated by multica-agent in a Multica workspace, developed this fix against upstream `main`: it first reproduced the #8011 topology with a deterministic local-only regression test and captured the pre-fix failure at the real production path (`CreateWorktree` with `IsolatedGitMetadata: true` fails at `git checkout --detach` on the just-resolved fetched tip), then applied the minimal patch enforcing the resolved-base-commit invariant with a bounded single-ref local recovery fetch. `go vet ./internal/daemon/repocache/...`, the focused regression suite, and `git diff --check` were run locally on the PR head; the change was then cross-validated by a second worker and an independent pre-flight check comparing the PR head with unmodified main (test PASS on head, FAIL on main).

### DIFF
--- a/server/internal/daemon/repocache/cache.go
+++ b/server/internal/daemon/repocache/cache.go
@@ -1110,6 +1110,13 @@ func createIsolatedCheckoutContext(ctx context.Context, barePath, repoURL, check
 		}
 	}
 
+	// Enforce the resolved-base-commit invariant before the detached checkout:
+	// the fresh clone's object store must contain the commit that was resolved
+	// from the shared cache before the clone started (multica-ai/multica#8011).
+	if err := ensureBaseCommitInCheckoutContext(ctx, barePath, checkoutPath, baseRef, baseCommit); err != nil {
+		return "", err
+	}
+
 	if out, err := runGitCombinedOutputContext(ctx, "-C", checkoutPath, "checkout", "--detach", baseCommit); err != nil {
 		return "", fmt.Errorf("git checkout --detach: %s: %w", strings.TrimSpace(string(out)), err)
 	}
@@ -1129,6 +1136,41 @@ func createIsolatedCheckoutContext(ctx context.Context, barePath, repoURL, check
 	}
 	cleanup = false
 	return actualBranch, nil
+}
+
+// ensureBaseCommitInCheckoutContext verifies that a fresh isolated checkout's object
+// store contains the commit resolved from the shared bare cache before the
+// clone started, and performs a bounded local recovery when it does not.
+//
+// The healthy path is a single `rev-parse --verify`: a `git clone --local`
+// from a full cache hard-links the whole object store, so the commit is there
+// and nothing else happens. When the cache is shallow, Git ignores --local
+// and clones over the pack protocol instead; such a clone transfers only the
+// objects reachable from the heads it fetches (refs/heads/*), so a freshly
+// fetched tip that arrived through the remote-tracking refspec
+// (refs/remotes/origin/*) can be missing even though it resolved in the
+// cache a moment earlier (multica-ai/multica#8011).
+//
+// Recovery fetches exactly baseRef from the local bare cache — never the
+// network origin, never a wider refspec, no cache mutation — into FETCH_HEAD
+// only. Object transfer alone is sufficient because the caller checks out
+// baseCommit by SHA. If the local cache cannot serve the commit either, fail
+// loudly instead of falling back: a shallow cache's stale refs/heads/* head
+// must never be checked out in place of the resolved base commit.
+func ensureBaseCommitInCheckoutContext(ctx context.Context, barePath, checkoutPath, baseRef, baseCommit string) error {
+	if gitRefExistsContext(ctx, checkoutPath, baseCommit+"^{commit}") {
+		return nil
+	}
+	out, err := runGitCombinedOutputContext(ctx, "-C", checkoutPath, "fetch", "--no-tags", barePath, baseRef)
+	if err != nil {
+		return fmt.Errorf("isolated checkout is missing resolved base commit %s (shallow cache at %s): recovery fetch of %s failed: %s: %w",
+			baseCommit, barePath, baseRef, strings.TrimSpace(string(out)), err)
+	}
+	if !gitRefExistsContext(ctx, checkoutPath, baseCommit+"^{commit}") {
+		return fmt.Errorf("isolated checkout is still missing resolved base commit %s after recovery fetch of %s from cache %s; refusing to check out a stale commit instead",
+			baseCommit, baseRef, barePath)
+	}
+	return nil
 }
 
 func resolveCommit(repoPath, ref string) (string, error) {
--- a/server/internal/daemon/repocache/shallow_cache_test.go
+++ b/server/internal/daemon/repocache/shallow_cache_test.go
@@ -1,0 +1,153 @@
+package repocache
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestCreateIsolatedCheckoutFromShallowCacheResolvesFetchedTip reproduces
+// multica-ai/multica#8011: an isolated checkout created from a shallow shared
+// cache loses the commit CreateWorktree just resolved as its base.
+//
+// The shared cache below has the reported topology: it is shallow, its stale
+// refs/heads/master still points at commit A, and the freshly fetched tip B is
+// reachable only through refs/remotes/origin/master. `git clone --local` from
+// a shallow source is silently downgraded to a transport-based clone, which
+// transfers only the objects reachable from the heads it fetches
+// (refs/heads/*) — so B is absent from the fresh checkout and the subsequent
+// `git checkout --detach B` fails (reported as "reference is not a tree";
+// "unable to read tree" on newer Git).
+func TestCreateIsolatedCheckoutFromShallowCacheResolvesFetchedTip(t *testing.T) {
+	t.Parallel()
+
+	// A bare origin plus a working clone: commit A, then advance origin/master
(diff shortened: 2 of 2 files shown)


multica-eve:
Thanks for the thorough fix and regression coverage. We are keeping #8024 because it fixes the same root cause by moving the existing cache-ref/base synchronization before the first checkout, avoiding a parallel recovery helper and an extra verification subprocess on every fresh isolated checkout. Both approaches pass the regression and package checks; this is being closed as superseded by #8024, not because of a correctness issue.

PR #8021, opened 2026-09-04
MUL-7033 fix(mcp): improve server configuration and management

### LINKED ISSUE #7996: [Feature]: Improve MCP server configuration and management
### Deployment type

Official App

### Area

Agents and runtimes

### What do you want and why?

Improve MCP server configuration and management in agent settings and the workspace MCP library, while preserving their existing data and permission boundaries.

Workspace MCP configuration is intentionally write-only: saved connection details are not returned to the editor, including for workspace owners. Showing an ordinary edit form with empty fields makes that security behavior look like data loss. Renaming should be a separate inline action; updating connection details should clearly ask for a complete new configuration. This changes the interaction, not the write-only contract or the existing API overwrite behavior.

The two pages also need consistent list rows, transport icons, action styles, and concise tooltips. The configuration dialog should keep its footer reachable after closing and reopening, provide clear validation, and preserve meaningful values such as leading, middle, and trailing empty arguments.

- Share list-row presentation across agent and workspace MCP surfaces, with transport labels in icon tooltips and consistent normal/hover states for destructive actions.
- Rename servers inline, separately from connection configuration. Hide the name input when configuring an existing server.
- Preserve the workspace library's write-only contract. Explain that saved connection details cannot be displayed and a complete new configuration is required to update them; keep the server ID and existing agent assignments unchanged.
- Continue to prefill readable agent-managed configurations. Sharing UI must not change which data either surface can read.
- Provide a guided STDIO/Streamable HTTP editor and a JSON path for advanced or provider-native configurations. Keep unsupported existing configurations JSON-only when editing them; allow an explicit full workspace configuration update to choose a supported transport.
- Keep the header and footer non-shrinking and scroll only the form body, including after reopening the dialog.
- Validate after an explicit save attempt, show localized field errors, and focus the first invalid control.
- Preserve empty arguments at every position and empty/whitespace environment or header values; retain drafts across parent rerenders and use current agent configuration when renaming.
- Cover the shared interactions, write permissions, configuration round-tripping, dialog lifecycle, and English, Simplified Chinese, Japanese, and Korean copy with regression tests.

Not included.

yyclaw:
## What does this PR do?

Makes MCP server management consistent across agent settings and workspace settings, with separate actions for renaming a server and changing its connection configuration.

Previously, workspace configuration editing opened empty fields because saved connection details are write-only, making the form look as though it had lost data. The new **Replace configuration** action explains that a complete new configuration is required. Inline renaming leaves the existing configuration and agent assignments intact. Agent configuration editing continues to prefill readable values.

The shared editor also moves validation to explicit save attempts and keeps its actions outside the scrolling form body, so users can correct errors and reach the save button after reopening or adding fields.

## Related Issue

Closes #7996

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `packages/views/common/mcp-server-row.tsx` and `mcp-transport.ts`: share server rows, protocol icons and labels, accessible action buttons, and inline rename controls.
- Agent `mcp-config-tab.tsx` and settings `mcp-tab.tsx`: separate rename, configure, and remove actions; enforce existing UI write permissions; use current agent configuration during rename and preserve workspace replacement drafts across rerenders.
- `packages/views/agents/components/tabs/mcp-server-dialog.tsx`: organize STDIO/HTTP fields, retain advanced JSON editing, hide the name when configuring an existing server, show field errors after submission, and focus the first invalid input. Keep the header/footer fixed within the dialog while its body scrolls.
- `packages/views/locales/{en,zh-Hans,ja,ko}/{agents,settings}.json`: update labels, validation messages, and configuration guidance for these interactions.
- Add regression coverage for renaming, permissions, dialog reopening, shared controls, and meaningful empty values, including an agent-update HTTP test in `server/internal/handler/handler_test.go`.

The change retains existing APIs and the workspace write-only boundary. Unsupported existing configurations keep their JSON editing path. Browser verification of footer visibility remains outstanding.

## How to Test

1. Run `pnpm --filter @multica/views typecheck`.
2. Run `pnpm --filter @multica/views test agents/components/tabs/mcp-config-model.test.ts agents/components/tabs/mcp-config-tab.test.tsx common/mcp-server-row.test.tsx settings/components/mcp-tab.test.tsx agents/components/tabs/agent-mcp-tab.test.tsx`.
3. Rename servers on both pages; check validation, cancellation, failure handling, and viewer permissions. Verify workspace rename preserves the connection configuration and assignments.
4. Configure an agent server and replace a workspace configuration. Verify their respective prefilled/empty forms, switch between form and JSON modes, and round-trip arguments such as `["", "2222", "", "333", ""]`.
5. Reopen the dialog at a small viewport height and add fields until it scrolls; verify save/cancel remain reachable.

On commit `861b5e692`, TypeScript and the five listed test files passed: 78 tests. The full views suite, database-backed Go test, and browser checks were not run on this commit.

## Checklist

- [x] Separate inline renaming from configuration updates on both MCP pages.
- [x] Preserve workspace write-only configuration and existing agent assignments.
- [x] Add regression tests for MCP management and dialog interactions.
- [x] Pass views TypeScript checks and 78 targeted tests on `861b5e692`.
- [x] Check the changed Chinese MCP terminology against current conventions: 智能体, 工作区, 运行, and lowercase `owner` / `admin`.

## AI Disclosure

**Prompt / approach:**
Unify MCP management on the agent and workspace pages while preserving their different configuration-read boundaries. Separate renaming from configuration updates, improve editor validation and scrolling, and cover the resulting interactions with regression tests.

## Screenshots

Before:
image

After:
image
image

Agent detail:
image
image


Bohan-J (changes requested):
Thanks for this — the product direction is right and I'd like to merge it.

The core insight is correct: the workspace library's write-only storage was leaking into the UI as apparent data loss, and the old flow made it worse than a cosmetic problem. Because the previous save path submitted `{ serverId, name, config }` together, **renaming a shared server necessarily overwrote its entire connection configuration** — you had to retype the URL and headers from memory just to change a label. Splitting rename from replacement fixes a real destructive-edit bug, not just a confusing form. Explaining the write-only boundary in the replace dialog, unifying transport terminology across both surfaces, and moving to field-level errors with focus on the first invalid input are all clear wins.

Three things I'd like fixed before merge. All are small.

## 1. Workspace rename/replace leaves the agent page showing stale data

`useUpdateWorkspaceMcpServer` only invalidates the workspace library:

https://github.com/multica-ai/multica/blob/861b5e692898730153a9045f812637630c0d7df5/packages/core/workspace/mutations.ts#L113-L114

But an agent's assignment list is a separate key (`packages/core/workspace/queries.ts:199`, `["agents", agentId, "mcp-servers"]`) and the global `staleTime` is `Infinity` (`packages/core/query-client.ts:7`). So: visit an agent's MCP tab, go to Settings → MCP, rename a shared server (or replace it with a different transport), return to the agent — the old name and old transport are still on screen, while the next run already uses the new configuration.

`useDeleteWorkspaceMcpServer` already handles this correctly, with a comment explaining why:

https://github.com/multica-ai/multica/blob/861b5e692898730153a9045f812637630c0d7df5/packages/core/workspace/mutations.ts#L122-L127

The same reasoning now applies to update, since this PR makes rename a first-class action people will actually use. Please invalidate `["agents"]` there too, with regression coverage in `packages/core/workspace/mutations.test.tsx` for both rename and a transport-changing replacement.

Worth weighting this above a normal cache nit: a rename that appears not to have taken effect recreates exactly the "did the system just lose my change?" doubt this PR sets out to remove, on the same user path.

## 2. Transport is now hover-only, which is a regression

The old row rendered the transport as a caption under the name. The new shared row reduces it to an icon with a tooltip. For the workspace library specifically, the configuration is unreadable by design, so **name + transport is the entire visible content of a row** — this hides half of it behind hover, and touch users can't reach it at all.

It also works against this PR's own consistency goal: the runtime list still renders `mcpTransportLabel(...)` as text, so the same page now shows transport two different ways. In the screenshots, three rows sit next to each other distinguished only by a small glyph.

Please keep the text label next to the icon. The tooltip is a good addition, just not as the sole carrier.

## 3. Field placeholders were dropped from Command and URL

`npx` and `https://mcp.example.com/mcp` were both removed. Arguments, environment variables and headers gained descriptive hints in exchange, but Command and URL are now bare empty inputs with only a label.

That loss lands precisely where guidance matters most: in **Replace configuration** the fields are empty by design and the user is reconstructing a config from memory with no reference to the saved value. Losing the format example there is the opposite of what that flow needs. Please restore an example for both (as placeholder or `FieldDescription`).

## Non-blocking

- **Three icon-only actions per row, with rename (pencil) and replace (refresh) sitting adjacent and semantically close.** Mis-clicking replace is recoverable — cancel loses nothing — so this isn't a safety issue, just a wasted round trip. Still, `Replace configuration` is a low-frequency, heavyweight action; I'd give it a text label rather than a glyph, at least in workspace settings.
- The PR notes browser verification of footer visibility at the new `max-h-[88vh]` is still outstanding. Please confirm at a small viewport with the form scrolled, after closing and reopening the dialog.
- Longer term, and explicitly **not** something to solve here: treating the whole config as one secret is coarser than reality — a URL usually isn't sensitive, header and env *values* usually are. If we later mark sensitivity per field, "replace everything" can go away entirely. This PR doesn't block that path, which is part of why I'm happy to take it as the intermediate step.

Happy to re-review as soon as the three items above are in.

yyclaw:
@Bohan-J addressed in the squashed commit `ae0e29dd8`.

Implemented the three requested changes:

- Workspace rename and configuration replacement now invalidate both the workspace library and agent assignment caches. Regression tests use a real QueryClient and cover rename, HTTP-to-STDIO replacement, and recovery when the request committed but its response was lost.
- Restored visible transport text across the affected MCP lists while retaining the tooltips.
- Restored the `npx` and `https://mcp.example.com/mcp` examples for Command and URL.

Also fixed configuration editing for existing server names containing dots or spaces: hiding the name field could prevent saving, or silently trim an existing name. Configuration edits now preserve the exact saved name; new names and renames retain their validation. The inline rename input now also has more space above the transport caption.

Regarding the non-blocking suggestion to add a persistent text label to the replacement action, I have kept the three row actions as consistent icon buttons. I consider the existing cues sufficiently clear: the “Replace configuration” tooltip, the dialog title naming the target server, the explanation that a complete new configuration is required, and the final “Replace configuration” button. Clicking the row action only opens a cancellable form; the saved configuration is changed only on submission. I therefore prefer to avoid repeating this label on every row.

Verified in Chrome at 800×600 and 390×568: after adding eight header rows, scrolling to the bottom, closing and reopening the dialog, Save and Cancel remain fully visible and clickable, and the final save succeeds. This check mounted the actual shared components and project styles in a local test page.

Related TypeScript tests, core/views typechecks, and two handler regressions against an isolated PostgreSQL database have completed. The full repository suite and native Desktop application were not run. Field-level sensitivity remains a follow-up discussion.

Bohan-J (approved):
Thanks — this addresses all three blocking items, and the revision is tighter than what I asked for.

I verified the delta between `861b5e69` and `ae0e29dd8` file by file rather than reading it off the summary:

- **Cache invalidation.** `onSettled` now invalidates both the library and `["agents"]`, matching what delete already did. The regression tests use a real `QueryClient` with `staleTime: Infinity` instead of a mock, cover rename and HTTP→STDIO replacement, and distinguish an active (refetching) assignment query from an inactive (invalidated) one. The case I hadn't thought to ask for — update commits but its response is lost, assignments still refetch, and the rejection still propagates — is a good catch, and it's what proves returning a promise from `onSettled` doesn't swallow the error.
- **Transport text** is back on every affected list, asserted via visible text within the row rather than by class.
- **`npx` and `https://mcp.example.com/mcp`** are restored, with tests asserting both the empty value and the placeholder — exactly the replace-flow state where the hint matters.

The additional `preservesServerName` fix is the right call and correctly scoped: only a byte-identical existing name skips validation, while create and rename keep the strict grammar, and `mcp-server-dialog.test.tsx` covers both directions. Without it, an entry that arrived from a runtime or the API with a dot or a space in its name would have been unopenable and unsavable.

On the icon-only replace action: your reasoning holds. The row action only opens a cancellable form, and nothing is committed until the explicitly labelled button inside a dialog that names the server and explains what will happen. I withdraw the suggestion.

One small follow-up, not for this PR: in both inline `handleRename` implementations the `name === renamingServer.name` no-op check sits after the grammar check, so opening rename on a historical name like `context7.dev` and confirming it unchanged reports a format error instead of quietly closing. It only affects agent-side entries — the workspace library enforces the same grammar server-side in `validateWorkspaceMcpServerName` — but it's the same class of case this PR just taught the dialog to tolerate.

Approving. Thanks for the thorough browser verification write-up as well.

PR #8027, opened 2026-09-04
feat(cli): add Gitea release source support (MUL-7058)

### LINKED ISSUE #8035: [Feature]: Support self-hosted Gitea release sources
## Deployment type

- [x] self-host

## Area

- [x] Self-hosting and deployment

## What do you want and why?

Self-hosted Multica deployments may have access to an internal network or Gitea instance but no reliable direct access to GitHub. In that setup, `multica update` cannot discover or download the latest CLI release even though the deployment's own release infrastructure is available.

The current CLI updater hard-codes GitHub Releases for release metadata and binary downloads. This creates an operational gap for self-hosted installations and requires operators to manually copy binaries or reimplement update logic in their deployment agent.

## Proposed solution

Allow the Multica CLI and daemon updater to use a configurable self-hosted release source while preserving GitHub as the default.

Suggested configuration:

```bash
MULTICA_RELEASE_API_BASE_URL=https://gitea.example.com/api/v1
MULTICA_RELEASE_DOWNLOAD_BASE_URL=https://mirror.example.com/releases
```

The source should provide GitHub Releases API-compatible metadata, release assets, and `checksums.txt`. Checksum verification must remain mandatory. Self-hosted daemon auto-update should remain explicitly opt-in through `MULTICA_DAEMON_AUTO_UPDATE=true`.

Gitea is the primary use case; N8N or another automation system may synchronize upstream releases into Gitea but should not be a Multica-specific dependency.

Related implementation PR: #8027

## Screenshots / mockups

Not applicable; this is a CLI and self-hosting feature.

2233admin:
## What does this PR do?

Self-hosted Multica deployments often cannot reach GitHub directly, which makes `multica update` fail even when the deployment has a healthy internal network. Gitea is a natural self-hosted release source because it provides release metadata, release assets, and checksums through an HTTP API.

This PR adds configurable release sources so a deployment can update from a Gitea instance or another GitHub Releases API-compatible mirror. GitHub remains the default when no override is configured.

## Related Issue

Closes #8035

Deployment tracking: CERE-322.

## Type of Change

- [x] Documentation update

## Changes Made

- Added `MULTICA_RELEASE_API_BASE_URL` for release metadata discovery.
- Added `MULTICA_RELEASE_DOWNLOAD_BASE_URL` for an optional artifact mirror serving `/<tag>/<asset-name>`.
- Preserved GitHub as the default release source.
- Preserved mandatory SHA-256 verification through `checksums.txt`.
- Documented Gitea configuration and the existing opt-in self-hosted daemon auto-update setting.
- Added regression coverage for default and overridden release endpoints.

The implementation stays provider-neutral: Gitea is supported through its compatible `/api/v1/repos/.../releases/...` endpoints without adding a Gitea-specific dependency or code path.

## How to Test

1. Configure `MULTICA_RELEASE_API_BASE_URL` against a Gitea or HTTP test server exposing the Releases API-compatible endpoints.
2. Run the focused updater tests:
   `go test ./internal/cli -run "Test(Release|Find|Parse|Verify|FetchLatest)" -count=1`
3. Compile the CLI package and run static checks:
   `go test ./cmd/multica -run ^$ -count=1`
   `go vet ./internal/cli ./cmd/multica ./internal/daemon`

The focused tests, CLI compile check, `go vet`, and `git diff --check` pass. The repository-level `make test` command was also attempted but is gated by the repository's missing local `.env` file.

## Checklist

- [x] I have run targeted tests locally and they pass

## AI Disclosure

**Prompt / approach:**

Traced the existing CLI updater and self-hosted daemon update policy, reproduced the GitHub connectivity failure on a self-hosted runtime, implemented configurable release metadata and artifact sources while preserving checksum verification, added focused `httptest` regression coverage, and ran targeted Go tests, compile checks, `go vet`, and whitespace validation.

## Screenshots (optional)

Not applicable; this is a CLI/backend change.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -31,7 +31,23 @@ For install script or manual installs, use:
 multica update
 ```
 
-`multica update` auto-detects your installation method and upgrades accordingly.
+`multica update` uses GitHub Releases by default. Self-hosted installations can
+point the CLI at a GitHub Releases-compatible metadata mirror and an artifact
+mirror without changing the command. To let a daemon poll that source, also
+enable self-update explicitly because self-hosted auto-update is disabled by
+default:
+
+```bash
+export MULTICA_RELEASE_API_BASE_URL=https://updates.example/api
+export MULTICA_RELEASE_DOWNLOAD_BASE_URL=https://updates.example/releases/download
+export MULTICA_DAEMON_AUTO_UPDATE=true
+multica update
+```
+
+The metadata mirror must serve `/repos/multica-ai/multica/releases/latest` and
+`/repos/multica-ai/multica/releases/tags/<tag>`. The artifact mirror must serve
+`/<tag>/<asset-name>` and preserve the published `checksums.txt` contents.
+When these variables are unset, the GitHub defaults remain unchanged.
 
 ## Quick Start
 
--- a/server/cmd/multica/cmd_update.go
+++ b/server/cmd/multica/cmd_update.go
@@ -30,7 +30,7 @@ func runUpdate(_ *cobra.Command, _ []string) error {
 
 	fmt.Fprintf(os.Stderr, "Current version: %s (commit: %s, built: %s)\n", version, commit, date)
 
-	// Check latest version from GitHub.
+	// Check the configured release source.
 	latest, err := cli.FetchLatestRelease()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: could not check latest version: %v\n", err)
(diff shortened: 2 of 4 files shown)


Bohan-J (approved):
Thanks a lot for this, @2233admin! This is a clean and nicely scoped change: GitHub stays the default, checksum verification is still mandatory, and daemon auto-update on self-hosted stays opt-in. It also works for both manual `multica update` and the daemon's update path.

I re-verified it on top of current `main`. `go build ./...`, `go vet`, and the `internal/cli`, `internal/daemon`, and `cmd/multica` test suites all pass. Merging now.

If you'd like to do a small follow-up, two things would help Gitea users:
- A concrete Gitea example in `CLI_AND_DAEMON.md`. The API base needs the `/api/v1` suffix, e.g. `MULTICA_RELEASE_API_BASE_URL=https://gitea.example/api/v1`, and `MULTICA_RELEASE_DOWNLOAD_BASE_URL` can be left unset when the release asset URLs are already reachable. Also a note that a running daemon has to be restarted from an environment where these variables are set.
- An end-to-end `httptest` case that runs `UpdateViaDownloadWithTimeout` against a mirror, covering both a checksum match and a mismatch.

Thanks again for the contribution!

Issue #6883, opened 2026-08-13
[Bug]: v0.4.24 repeatedly spawns codex-windows-sandbox-setup.exe dialogs with task-scoped elevated sandboxes


hhhappb:
## Summary

On Windows 11, Multica Desktop v0.4.24 repeatedly launches `codex-windows-sandbox-setup.exe` while Codex tasks are running. Multiple identical native dialogs can appear concurrently:

```text
...\codex-windows-sandbox-setup.exe
The specified module could not be found.
(找不到指定的模块。)
```

The helper exists and has a valid OpenAI Authenticode signature. Updating the global npm Codex installation from 0.145.0 to 0.147.0 did not permanently resolve the problem.

This is related to #5876, but is filed separately so the current v0.4.24 native-dialog regression can be tracked independently.

## Environment

- Multica Desktop / bundled daemon: `v0.4.24` (2026-08-12)
- OS: Windows 11, amd64, zh-CN
- Codex CLI: `0.147.0`
- Installation: global npm package `@openai/codex`
- Multica profile: Desktop-managed named profile
- Task policy: `sandbox_mode = "workspace-write"`
- Native sandbox: `[windows] sandbox = "elevated"`
- Multica creates a separate task-scoped `CODEX_HOME` for every task

Usernames, task IDs, workspace IDs, tokens, and absolute user paths are omitted.

## Actual behavior

1. Multica starts the globally installed npm Codex app server.
2. Every task receives an independent `codex-home`.
3. The task config retains the elevated Windows sandbox setting.
4. Codex repeatedly launches the elevated sandbox setup helper.
5. Multiple helpers can run in the same short time window.
6. One or more helpers display the native "specified module could not be found" dialog.
7. Successful helpers are recorded in task logs, while a competing failure may produce only the native dialog and no actionable Multica task error.

## Confirmed process chain

```text
multica.exe
  -> cmd.exe /c %APPDATA%\npm\codex.cmd app-server --listen stdio://
    -> node.exe
      -> npm-packaged codex.exe
        -> codex-windows-sandbox-setup.exe
```

Process inspection confirms that the dialogs originate from Multica-launched Codex processes, not the separate Codex Desktop app. A direct elevated Codex sandbox smoke test can succeed on the same machine outside the Multica lifecycle.

## Reproduction steps

1. Install Multica Desktop v0.4.24 on Windows 11.
2. Install Codex globally through npm.
[...]

PR #8028, opened 2026-09-04
fix(cli): compare workdir paths in one namespace (MUL-4252, MUL-7193)

### LINKED ISSUE #8213: CLI workdir containment guard compares paths in different namespaces
## Summary

The CLI workdir containment guard behind `--content-file`, `--description-file`, and `--attachment` (`fileWithinWorkingDir` in `server/cmd/multica/cmd_issue.go`) canonicalizes the workdir but often cannot fully canonicalize a candidate path that does not exist yet. The two sides are then compared in different namespaces.

This was the guard introduced for MUL-4252 (stale `/tmp` files from another run). The fix made the rule louder, but the predicate itself is wrong in both directions.

## Bug

**A path inside the workdir reads as outside.**

On hosts whose workdir path crosses a symlink (e.g. macOS `/var` → `/private/var`, or a logical `$PWD`), the base is canonical while the candidate stays logical:

```text
$ multica issue comment add MUL-123 --content-file reports/summary.md
Error: --content-file path "reports/summary.md" resolves outside the current
working directory; ...
Pass --allow-external-file to override.
```

`reports/summary.md` is plainly inside the workdir; `reports/` simply does not exist. The only remedy the error offers is to disable the guard, so the next retry fails with an unrelated not-found.

**A path outside the workdir reads as inside.**

With an already-canonical workdir, a symlink planted inside the workdir pointing out of it (`escape/sub/report.md`) is judged lexically and admitted.

**`..` is applied to the string, not to the followed path.**

`filepath.Join` / `filepath.Clean` collapse `escape/..` before any symlink is followed, while the kernel follows the link first. A guard-true / read-outside window remains for an existing file.

## Expected

Both sides of the comparison should be canonicalized the same way (resolve as far as the filesystem allows, then re-attach the missing tail), so:

- a missing intermediate directory is not reported as a containment violation
- a symlinked escape hatch is rejected even when the leaf is missing
- `..` is applied after following symlinks

## Context

- Related to the MUL-4252 guardrail
- Same predicate also backs the runtime-local-link lint (`local_path_link_lint.go`)
- Fix is implemented in #8028

GODVvVZzz:
## What does this PR do?

`--content-file`, `--description-file` and `--attachment` all pass their path
through one containment guard (`fileWithinWorkingDir`) before reading it.
MUL-4252 put that guard there: an agent wrote its description to a fixed
`/tmp/desc.md`, the write silently failed, a *different* environment's run had
left a stale file at that path minutes earlier, and the CLI fed that stale
content into a new issue. Requiring the path to live under the task workdir
turned "silently read another run's file" into a loud command failure.

The guard canonicalizes the workdir, but cannot canonicalize a candidate path
that does not exist yet, and what it does instead compares the two sides in
different namespaces. That is wrong in four ways, and only the first is merely a
wrong answer.

**1. A path inside the workdir reads as outside.** `os.Getwd()` returns the
LOGICAL working directory whenever `$PWD` names the current directory, and
everything that writes `$PWD` carries the path as typed: a shell's `cd`, and the
`PWD=` this repo exports to agent processes
(`pkg/agent/opencode.go:143`, `deveco.go:142`, `codearts.go:138`). So on any host
whose workdir path crosses a symlink, the base is `/private/var/...` while the
candidate stays `/var/...`:

```
$ multica issue comment add MUL-123 --content-file reports/summary.md
Error: --content-file path "reports/summary.md" resolves outside the current
working directory; write agent temp files inside the task workdir ...
Pass --allow-external-file to override.
```

`reports/summary.md` is plainly inside the workdir. The real problem is that
`reports/` does not exist. The only remedy the error offers is the flag, so the
agent's next move is to disable the guard and retry — earning a second,
unrelated `no such file or directory`. The diagnosis points away from the problem
and teaches the agent to turn the guard off.

**2. A path outside the workdir reads as inside.** With an already-canonical
workdir the same fallback inverts. On `main`:

```
escape/sub/report.md     within=true    (`escape` is a symlink out of the workdir)
escape/a/b/report.md     within=true
```

The unresolvable tail falls back to a lexical clean that never sees the symlink.

**3. `..` is applied to the string, not to the followed path.**
`filepath.Join`/`Clean` collapse `escape/..` before any symlink is followed,
while the kernel follows `escape` out of the workdir first and then goes up. So
the guard judges a different path than `os.ReadFile` opens:

```
guard says within=true ; os.ReadFile got "another run's file"
```

That is the MUL-4252 failure itself, reachable with a file that exists and no
race.

**4. A path on another volume reads as an internal failure.** Windows only:
`filepath.Rel` cannot relate two absolute paths on different volumes and returns
an error, which the guard surfaced as

```
resolve --content-file path "Z:\report.md": Rel: can't make Z:\report.md
relative to C:\Users\...\workdir
```

Same shape as (1): the caller is told about path algebra instead of the rule it
broke.

### Why this approach

`util.ResolveSymlinksBestEffort` canonicalizes as far as the filesystem allows:
it resolves the path as given first — the only step that can see `..` the way the
kernel does — then walks up to the deepest existing ancestor, resolves that, and
re-attaches the missing tail lexically. Both sides of the comparison go through
it. A mismatch inside the missing tail is inert: a path whose intermediate
directories do not exist cannot be opened at all.

- **Not "compare cleaned paths on both sides".** `filepath.Clean` does not follow
  symlinks, so a symlink planted in the workdir walks a candidate straight past
  the guard. 5 tests / 15 subtests fail under that mutation.
- **Not "resolve one level up".** That is what `main` does, and it breaks as soon
  as an intermediate directory is missing — in both directions above. 3 tests / 8
  subtests fail.
- **Not `os.Root` / `os.OpenInRoot`.** It would close the check-then-read window
  too, but it cannot serve the third caller: `classifyLocalPathTarget` needs a
  predicate for paths that do not exist, and `root.Stat` cannot distinguish
  "inside but missing" from "outside". Left as a non-goal rather than half-done.
- **Not "fail closed when a component exists but cannot be canonicalized".** On
  Windows that would reject every path under a pnpm-style `node_modules`
  junction, and on a host whose workdir is itself reached through a junction it
  would reject every `--content-file` and `--attachment` on that host. A larger
  break than the one it prevents.

`internal/daemon/execenv` already had this algorithm as `resolvePathBestEffort`,
unexported, documented as Python's `Path.resolve(strict=False)`. It now
delegates, so there is one owner, and the `..` handling makes it more faithful to
that documented contract rather than less.

The rest of `server/` was swept for the same asymmetry (64 `EvalSymlinks` sites,
23 `filepath.Rel` sites). Every other containment check either resolves both
sides or requires the path to exist, so this predicate is the only one affected.

### Verified on three platforms

| | macOS 26.5.1 arm64, go1.26.6 | Linux amd64, go1.26.8 | Windows 10.0.19045, go1.26.6 |
| --- | --- | --- | --- |
| `./cmd/multica` `./internal/util` `-race` | ok | ok | ok (`internal/util`; `cmd/multica` scoped, below) |
| `./internal/daemon/execenv` | ok | ok | n/a |
| skipped subtests | 0 | 0 | symlink cases skip on a non-admin host |
| `one-level-fallback` mutation | 3 tests / 8 subtests | identical list, line for line | n/a |
| cross-volume branch | unreachable | unreachable | red under mutation, green with the fix |

## Related Issue

Closes #8213

Implements the containment fix for the guard MUL-4252 introduced. The same
predicate backs the runtime-local-link lint
(`cmd/multica/local_path_link_lint.go:114`), where the identical false negative
flips direction: a workdir-local path with a missing intermediate directory
matches neither of the lint's signals, so the link is published instead of
blocked.

## Type of Change

- [ ] Documentation update
- [x] CI / infrastructure

## Changes Made

- `server/internal/util/path.go` (new) — `ResolveSymlinksBestEffort`. Resolves
  the path as given before any lexical cleaning, then walks up to the deepest
  existing ancestor and re-attaches the missing tail. Any `EvalSymlinks` error
  stops the descent, not just "not exists": permission-denied on an ancestor is
  equally "cannot canonicalize here".
- `server/cmd/multica/cmd_issue.go` — both sides of `fileWithinWorkingDir` go
  through the helper; the candidate is handed over as typed, because pre-joining
  it would clean the string and collapse a `..` across a symlink. A
  `filepath.Rel` error now reads as "outside". `pathExists` splits the two facts
  the rejection message used to merge, so a path that is outside AND missing
  leads with the missing file; any stat error other than "not exists" keeps the
  guard's own wording and stays fail-closed.
- `server/internal/daemon/execenv/hermes_home.go` — `resolvePathBestEffort`
  delegates to the shared helper.
- `server/cmd/multica/workdir_guard_windows_test.go` (new, windows-tagged) — the
  guard's Windows-only judgments, and the junction gap pinned with the
  observations it rests on.
- `server/internal/util/path_test.go` (new), `cmd_issue_test.go`,
  `local_path_link_lint_test.go` — `TestFileWithinWorkingDir` runs its table
  under two working-directory shapes, and both are load-bearing: under a
  symlinked workdir an unresolved candidate always looks outside, so the
  symlink-escape rows can only fail under a canonical workdir.
- `.github/workflows/ci.yml` — one scoped step in `windows-execenv`.
  `cmd/multica` has no Windows job otherwise, so these tests would run nowhere;
  it matters twice over on a non-admin host, where the package's cross-platform
  cases skip for want of the symlink privilege while a junction needs none. `-v`
  so a silent skip is visible instead of passing as `ok`.

## How to Test

Each step removes one part of the mechanism and names the test that must go red.

1. `cd server && go test -race -count=1 ./cmd/multica/ ./internal/util/` and
   `go test -count=1 ./internal/daemon/execenv/` — all pass. The last one is
   included because `hermes_home.go` now delegates to the shared helper.
2. Replace `abs := util.ResolveSymlinksBestEffort(filePath)` in
   `fileWithinWorkingDir` with `main`'s fallback (EvalSymlinks, else one
   `filepath.Dir` step, else `filepath.Clean`). 3 tests / 8 subtests fail,
   including `canonical_workdir/symlinked_escape_hatch_with_a_missing_directory`
   — the direction where the old code admits a path that leaves the workdir.
3. Replace it with `filepath.Clean` instead: 5 tests / 15 subtests fail. The
   extra rows are the escape hatches, which is what keeps this fix from
   degenerating into "compare unresolved paths".
4. `base := util.ResolveSymlinksBestEffort(cwd)` → `base := cwd`: 5 tests / 9
   subtests. The asymmetry the bug started from.
5. In the helper, wrap the first call as
   `filepath.EvalSymlinks(filepath.Clean(absNoClean(p)))`: 2 tests / 5 subtests,
   all of them the `..`-across-a-symlink rows.
6. In the helper, return early after one upward step: 4 tests / 9 subtests.
7. Delete the `if !pathExists(filePath)` branch, then separately make
   `pathExists` return false unconditionally.
   `TestExternalFileErrorLeadsWithMissingFile` fails on the missing-file
   assertion in the first case, and on "an existing external file must still be
   told about `--allow-external-file`" in the second.
8. Windows, `go test ./cmd/multica -v -run '^TestFileWithinWorkingDirWindows'`:
   9 subtests plus the junction test, no skips. Then change the cross-volume
   branch back to `return false, err`:
   `a_path_on_another_volume_reads_as_outside_the_workdir` fails with
   `must report a cross-volume path as outside, not as an error: Rel: can't make
   Z:\... relative to C:\...`, and nothing else changes.

### Windows coverage, as of the CI run on this head

The new step ran on `windows-latest` and the junction observation reproduces
there, on a different host from the one the notes were measured on
([`windows-execenv`](https://github.com/multica-ai/multica/actions/runs/33849149366/job/100947778344),
2m5s):

```
=== RUN   TestFileWithinWorkingDirWindowsPaths
--- PASS: TestFileWithinWorkingDirWindowsPaths (0.02s)          (9 subtests, no skips)
=== RUN   TestFileWithinWorkingDirWindowsJunctionIsAKnownGap
    junction ...\escape: Lstat mode=?rw-rw-rw- symlinkBit=false irregular=true
    EvalSymlinks("escape\\stale.md") = "" err=The system cannot find the path specified.
    known gap pinned: EvalSymlinks refuses the junction ..., so the guard judges
    "escape\\stale.md" lexically and admits it
--- PASS: TestFileWithinWorkingDirWindowsJunctionIsAKnownGap (0.02s)
```

Step 8's mutation (the cross-volume branch reverted to `return false, err`) was
run on the Windows 10.0.19045 host rather than in CI, since a red CI run is not a
useful artifact to leave behind.

## Deliberate non-goals

- **The check-then-read window.** The predicate answers, then the caller opens
  the path; a path can change underneath in between. Closing that means reading
  through `os.Root`, which the lint caller cannot use (see above). Unchanged by
  this PR in either direction.
- **Windows directory junctions.** `filepath.EvalSymlinks` does not descend
  through a junction — it resolves the junction itself to its own name, so the
  ancestor walk stops there and judges the tail lexically. A junction inside the
  workdir pointing out of it therefore reads as inside, before this change and
  after it. The windows-tagged test pins that deliberately and fails with
  instructions if the platform ever starts resolving them.

  Which semantics apply is not a property of the host: it is GODEBUG
  `winsymlink`, whose default follows the main module's go directive
  (`os/types_windows.go` files a mount point under `ModeIrregular` in `mode()`
  and under `ModeSymlink` in `modePreGo1_23()`; `internal/godebugs/table.go`:
  `winsymlink, Changed: 23, Old: "0"`). This module declares go 1.26.6, so the
  CLI ships with the modern semantics, and so does the daemon-side test that
  asserts the same refusal (`internal/daemon/config_windows_test.go`).

  Closing it does not need `GetFinalPathNameByHandle`: `os.Readlink` answers on a
  junction under both settings, returns a clean absolute target with no `\??\`
  prefix, and answers even when the target is gone. Measured alongside the two
  facts that decide whether resolving is safe — a junction pointing inside the
  workdir resolves to a path still inside it, so resolving costs no false
  rejections, and a chain needs a bounded follow loop because one `Readlink` step
  only reaches the next link. That is a Windows behaviour change with its own
  blast radius, so it is its own PR.
- **`..` inside the missing tail.** Still lexical. A path whose intermediate
  directories do not exist cannot be opened, so a divergence there leaves nothing
  to read.
- **The walk's termination guard.** No test reaches it and none cheaply can: `/`
  always resolves, and `filepath.EvalSymlinks("Z:\\")` returns `Z:\` with a nil
  error for a drive letter with no volume behind it while `os.Stat` on the same
  root fails. It is there so the loop does not depend on what `EvalSymlinks` does
  at a root, which is unspecified and demonstrably disagrees with `os.Stat`. The
  branch says so rather than implying a test covers it.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, CLI only
- [x] I have updated relevant documentation to reflect my changes — the guard's
      doc comments; no docs page or prompt quotes these error strings (checked)
- [ ] If I added a new runtime / coding tool / UI tab, I synced landing copy and docs — N/A

### Risks

- The rejection message for a path that is outside AND missing changes. Existing
  tests that assert on the old wording use files that exist, so they are
  unaffected; the wording for an existing external file is unchanged, and
  `TestExternalFileErrorLeadsWithMissingFile` pins both halves.
- `resolvePathBestEffort` in `execenv` changes behaviour for one input shape: a
  path containing `..` after a symlink now resolves physically rather than
  lexically. That is the documented contract of the function
  (`Path.resolve(strict=False)`) and the full package passes, but it is a
  behaviour change in the daemon reached from a CLI fix.
- Unrelated to this change, flagged because it was hit while verifying on Linux:
  three `execenv` tests match git's English output
  (`gitroot_lock_test.go:250` requires `fatal:`; `local_worktree.go:1312` keys on
  `already used by worktree`) and fail on a host with a non-English locale. CI
  runs in English so it never sees this. Not touched here.

## AI Disclosure

**AI tool used:** Kiro CLI

**Prompt / approach:** The bug was found by reading this guard rather than from a
report: its error text pointed at `--allow-external-file` for a path that was
plainly inside the workdir. The loop was to reproduce each failure against
current `main` before changing anything, sweep the rest of `server/` for the same
asymmetry, then mutation-verify every new assertion by reverting the mechanism it
guards — the numbers in "How to Test" are those runs. The Linux and Windows rows
of the platform table were produced by agents on those hosts, given the mutation
recipes to run rather than conclusions to confirm.

---

## Update addressing the review (d3d370377)

Both blockers are closed. The review was validated against `3ec1781e`; this section describes the tree as of `d3d370377`.

### 1. Windows path kinds are preserved (`server/internal/util/path.go` split per platform)

`absNoClean` prefixed the working directory onto every path `filepath.IsAbs` rejects — but on Windows that predicate is true of two kinds the loader does **not** resolve against the working directory: a root-relative `\tmp\desc.md` resolves against the volume root, and a drive-relative `C:tmp\desc.md` against the current directory on that drive. The guard inspected `C:\task\workdir\tmp\desc.md` while `os.ReadFile` opened `C:\tmp\desc.md`; with the shadow present, an outside read passed.

- `path_other.go` / `path_windows.go` now carry platform-specific `absNoClean`. The Windows version preserves all four kinds: drive-absolute and UNC pass through, root-relative resolves against `filepath.VolumeName(cwd)` (the UNC share when the cwd is one), and drive-relative resolves against the working directory only when the drive **is** the cwd's drive — the one case where the OS actually resolves it there.
- A drive-relative path on any other drive has a per-drive current directory this process cannot observe, so it is handed back uncanonicalized. `ResolveSymlinksBestEffort` returns it unchanged, and the guard's `filepath.Rel` branch fails closed across volumes.
- Regression tests in `workdir_guard_windows_test.go`: a real root-relative target (`\tmp\desc.md`, forward-slash form too) with the in-workdir shadow present must be **rejected**; a same-drive drive-relative path must still resolve into the workdir; a cross-drive drive-relative path must fail closed. The existing cross-volume case is unchanged and still passes.

### 2. Junctions resolve; the known-gap pin is gone

`filepath.EvalSymlinks` refuses to descend through a junction under the `winsymlink` semantics this module's go directive selects — that was the gap the old test pinned. Resolution on Windows is now component-wise (`evalPath` in `path_windows.go`) and follows junctions via `os.Readlink`, which answers under both `winsymlink` settings, with a bounded follow budget (255, EvalSymlinks' own scale). A reparse point `os.Readlink` cannot speak for — a OneDrive placeholder — is treated as an opaque regular component, because the kernel opens it where it stands.

- A junction out of the workdir is **rejected** (`escape\stale.md` → resolved target outside). A junction chain out of it is rejected too.
- A junction into the workdir — the pnpm `node_modules` shape — is **still admitted**; resolving costs no false rejections there (the fact measured alongside the escape behavior when the gap was first characterized).
- A junction whose target is gone: `Readlink` answers but the walk cannot resolve through the dead target, so the junction becomes the first unresolvable component and the tail is re-attached lexically — inside, and inert (the kernel stops there too; `os.ReadFile` surfaces the not-found). Pinned with that explanation.
- The "Deliberate non-goals" entry about Windows junctions above is **superseded** by this change. The consequence is intended: paths under an out-of-workdir junction are now rejected, which is the review's direction.

### The shared walk is now component-wise on both platforms

Restructuring `evalPath` for junctions made the second phase of `ResolveSymlinksBestEffort` component-wise everywhere. Two consequences, both pinned by tests:

- `..` inside the walk is applied to the **resolved** prefix even when the tail is missing. Previously that phase started from `filepath.Abs`, which cleans first, so `esc/../missingdir/x.md` (esc a live symlink) was judged under the workdir while the kernel would place the tail next to the link's target. New subtest in `path_test.go`. Behavior for every shape the existing tests cover is unchanged; the full suites pass unmodified.
- The walk starts from the root without needing it to resolve, so the old `parent == cur` termination branch and its reliance on what `EvalSymlinks` does at a root are gone; the cross-volume case (`Z:\...`) exercises the lexical-from-root shape.

### Verified

- macOS arm64, go1.26.x: `go test -race -count=1 ./cmd/multica/ ./internal/util/ ./internal/daemon/execenv/` — all pass. `GOOS=windows go build && go vet` for `internal/util` and `cmd/multica` — clean.
- Windows assertions (root-relative shadow, drive-relative both directions, junction escape/admit/chain/dead-target) run in the `windows-execenv` job on this head — junctions need no elevation, so they execute on `windows-latest` rather than skipping.

---

## Update addressing the second review (ca96a79a0)

Both blockers from the review of `d3d370377` are closed.

### 1. Link targets are classified with the full Windows path grammar

`evalPath` used to splice a target onto the link's parent unless `filepath.IsAbs` said otherwise — wrong for two kinds the kernel resolves by the same grammar as any path:

- A root-relative target (`\outside`) now resolves against the root of the volume the link sits on (`classifyTarget`), not against the link's directory. With the in-workdir shadow present, the guard rejects; the resolver's answer matches what `os.ReadFile` opens.
- A volume-GUID target (`\\?\Volume{GUID}\...`) is absolute in the device namespace and is kept **verbatim**: `trimDeviceNamespace` (which stripped `\\?\` and thereby turned the absolute mount target into a spliced "relative" path) is gone. A device-namespace target without a Win32 spelling ends the walk — the remaining components are joined lexically and the containment comparison fails closed across the namespace. Device-qualified (`\\?\C:\...`) and UNC (`\\?\UNC\...`) forms are reduced to their Win32 spelling so they can still be walked and compared; the conversion ignores path length on purpose, since the form is only compared, never handed to an API. The prefix rules mirror `internal/daemon/canonical_path_windows.go` (#6883).

Negative tests for both shapes are in `workdir_guard_windows_test.go`, each verifying the kernel's target **by content** (shadow and target carry different bytes; the read through the link must return the target's) before asserting rejection, so the fixture cannot silently invert on a different host.

### 2. Readlink failure no longer means "does not redirect"

`os.Readlink` decodes only the symlink and mount-point tags; other name-surrogates (container-isolation links, for instance) redirect through a filter while the lexical location stays behind. The opaque-continue now requires confirmation from `GetFinalPathNameByHandle` (via `golang.org/x/sys/windows`, already a dependency — the same pattern `internal/daemon/canonical_path_windows.go` uses):

- The component is opened with **zero desired access** (no reads, no writes, a cloud placeholder is not hydrated) plus `FILE_FLAG_BACKUP_SEMANTICS` for directories.
- Handle final path == the component → it redirects nothing (OneDrive placeholder, deduplicated file) → opaque, as before.
- Handle final path ≠ the component → it redirects → the walk continues from the final path through the same target classification.
- The handle cannot name the location → `evalPath` fails with the new `util.ErrUnresolvablePath` instead of judging the lexical form.

### The fail-closed case is now an explicit error

`ResolveSymlinksBestEffort` returns `(string, error)`; `ErrUnresolvablePath` covers what the kernel may open somewhere this process cannot name: an unknown redirecting reparse point that survives both `Readlink` and a handle query, and a drive-relative path on an unobservable drive (previously returned uncanonicalized and distinguished only implicitly by `filepath.Rel` erroring). Callers:

- `fileWithinWorkingDir` treats the error as **outside** — fail closed, with the reasoning in the comment.
- `execenv`'s `resolvePathBestEffort` falls back to the input unchanged; its `isPathUnder` comparison then fails to relate it and treats it as under nothing.
- The Unix behavior is unchanged for every shape the existing tests cover; the full `internal/util`, `cmd/multica` (race), `internal/daemon` and `internal/daemon/execenv` suites pass on macOS, and both packages cross-compile for Windows.

### Verified

- macOS arm64: `go test -race -count=1 ./cmd/multica/ ./internal/util/` and `go test -count=1 ./internal/daemon/ ./internal/daemon/execenv/` — all pass. `GOOS=windows go build && go vet` for the affected packages — clean.
- Windows assertions run in the `windows-execenv` job on this head: the two new negative cases (root-relative symlink target + shadow, volume-GUID junction target + shadow) plus the existing junction/rooted/drive-relative matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -716,6 +716,29 @@ jobs:
         # silently as "ok".
         run: go test ./internal/daemon/execenv -v -run '^TestWindowsOpenclawShim' -count=1 -timeout=5m
 
+      - name: Test Windows workdir containment guard in the CLI
+        if: ${{ needs.changes.outputs.backend == 'true' }}
+        working-directory: server
+        # The CLI's --content-file / --attachment guard decides whether a path is
+        # inside the task workdir by canonicalizing both sides. Two of the things
+        # that decision rests on are Windows-only and covered nowhere else: a
+        # candidate on another volume, where filepath.Rel returns an error rather
+        # than a "..", so the guard has to read that as "outside" instead of
+        # surfacing `Rel: can't make Z:\... relative to C:\...` as a resolve
+        # failure; and a directory junction — the containment-relevant link shape
+        # that needs no elevation, so the one real hosts have — which resolves to
+        # itself rather than to its target under this module's go directive. The
+        # junction test pins that gap deliberately, records the observations it
+        # rests on, and fails with instructions if the platform ever starts
+        # resolving them.
+        # cmd/multica has no Windows job otherwise, so these windows-tagged
+        # tests run nowhere else. It matters twice over on a non-admin host,
+        # where the package's cross-platform cases skip for want of the symlink
+        # privilege; the ubuntu backend job covers those shapes instead.
+        # -v so a silent skip (or a -run pattern that stops matching) is visible
+        # in the log instead of passing as "ok".
+        run: go test ./cmd/multica -v -run '^TestFileWithinWorkingDirWindows' -count=1 -timeout=5m
+
       - name: Test Windows isolated repo checkout is committable
         if: ${{ needs.changes.outputs.backend == 'true' }}
         working-directory: server
--- a/server/cmd/multica/cmd_issue.go
+++ b/server/cmd/multica/cmd_issue.go
@@ -121,41 +121,52 @@ func ensureFileFlagWithinWorkdir(cmd *cobra.Command, fileFlag, flagName, filePat
 }
 
 // fileWithinWorkingDir reports whether filePath resolves to a location inside
-// the process working directory. Both sides are symlink-resolved so aliased
-// roots (e.g. macOS /tmp -> /private/tmp) and symlinks planted inside the
-// workdir fail closed. A path that does not exist yet is judged on its cleaned
-// absolute form so the caller's os.ReadFile still surfaces the real not-found
+// the process working directory. Both sides are canonicalized the same way, so
+// aliased roots (e.g. macOS /tmp -> /private/tmp) and symlinks planted inside
+// the workdir fail closed. A path that does not exist yet is resolved as far as
+// it exists, so the caller's os.ReadFile still surfaces the real not-found
 // error afterwards.
 func fileWithinWorkingDir(filePath string) (bool, error) {
 	cwd, err := os.Getwd()
 	if err != nil {
 		return false, err
 	}
-	base := cwd
-	if resolved, err := filepath.EvalSymlinks(cwd); err == nil {
-		base = resolved
-	}
-	abs := filePath
-	if !filepath.IsAbs(abs) {
-		abs = filepath.Join(cwd, abs)
-	}
-	if resolved, err := filepath.EvalSymlinks(abs); err == nil {
-		abs = resolved
-	} else {
-		// The file may not exist yet (the caller's os.ReadFile surfaces that).
-		// Resolve symlinks on the parent directory instead so the comparison
-		// base and the candidate share the same canonical prefix — otherwise a
-		// workdir under a symlinked root (e.g. macOS temp dirs) would falsely
-		// read as "outside". A missing parent falls back to a plain clean.
-		if resolvedParent, perr := filepath.EvalSymlinks(filepath.Dir(abs)); perr == nil {
-			abs = filepath.Join(resolvedParent, filepath.Base(abs))
-		} else {
-			abs = filepath.Clean(abs)
-		}
-	}
+	base := util.ResolveSymlinksBestEffort(cwd)
+	// Canonicalize the candidate exactly the way the base was, and hand it over
+	// as typed: util.ResolveSymlinksBestEffort makes it absolute itself, because
+	// pre-joining it here would clean the string first and collapse a ".."
+	// across a symlink — judging a different path than the one os.ReadFile then
+	// opens.
+	//
+	// os.Getwd() returns the LOGICAL working directory when $PWD names it (a
+	// shell's `cd`, and the PWD the daemon exports to agent processes, both
+	// carry unresolved symlinks), so the two sides only agree once both are
+	// resolved. Resolving best-effort is what makes that possible for a
+	// candidate that does not exist yet — a typo, or an artifact a build step
+	// never produced — and every cheaper approximation gets one direction wrong:
+	//
+	//   - Leaving the candidate unresolved (filepath.Clean) reads a path inside
+	//     the workdir as outside it, which reports a missing file as a
+	//     guardrail violation and points the caller at --allow-external-file
+	//     instead of at the real error.
+	//   - Resolving only one level up (filepath.Dir) still breaks once an
+	//     intermediate directory is missing, and worse: with an already-
+	//     canonical cwd it reads `escape/sub/x.md` — under a symlink pointing
+	//     out of the workdir — as INSIDE it, because the unresolvable tail
+	//     falls back to a lexical clean that never sees the symlink.
+	abs := util.ResolveSymlinksBestEffort(filePath)
 	rel, err := filepath.Rel(base, abs)
 	if err != nil {
-		return false, err
+		// Two absolute paths that filepath.Rel cannot relate are on different
+		// volumes, which is as far outside the workdir as a path can get. Only
+		// Windows produces this: with a workdir on C:, `--content-file
+		// Z:\report.md` used to surface `Rel: can't make Z:\report.md relative
+		// to C:\...` as an internal resolve failure instead of the guardrail's
+		// own explanation — the same misleading-diagnosis shape this guard is
+		// being fixed for. Measured on 10.0.19045 / go1.26.6. There is no Unix
+		// input that reaches this branch, so the windows-tagged test in this
+		// package is the only thing that covers it.
+		return false, nil
 	}
 	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
 		return false, nil
--- a/server/cmd/multica/cmd_issue.go
+++ b/server/cmd/multica/cmd_issue.go
@@ -110,6 +110,13 @@ func ensureFileFlagWithinWorkdir(cmd *cobra.Command, fileFlag, flagName, filePat
 		return fmt.Errorf("resolve --%s path %q: %w", fileFlag, filePath, err)
 	}
 	if !within {
+		if !pathExists(filePath) {
+			return fmt.Errorf(
+				"--%s path %q does not exist; it also resolves outside the current working directory, "+
+					"so --allow-external-file would not make this read succeed. Write the file inside the "+
+					"task workdir (e.g. ./%s.md) and pass that path.",
+				fileFlag, filePath, flagName)
+		}
 		return fmt.Errorf(
 			"--%s path %q resolves outside the current working directory; "+
 				"write agent temp files inside the run workdir (e.g. ./%s.md) rather than machine-shared "+
@@ -118,6 +125,18 @@ func ensureFileFlagWithinWorkdir(cmd *cobra.Command, fileFlag, flagName, filePat
 			fileFlag, filePath, flagName)
 	}
 	return nil
+}
+
+// pathExists reports whether filePath names something on disk. It exists to
+// separate the two facts the containment guard used to merge: a rejected path
+// that is ALSO missing has no stale file behind it for anyone to read by
+// mistake, so leading with --allow-external-file sends the caller to disable
+// the guard and hit an unrelated not-found error on the retry. Any stat error
+// other than "not exists" (a permission wall, a broken mount) is treated as
+// "exists" so the guard keeps its own wording and stays fail-closed.
+func pathExists(filePath string) bool {
+	_, err := os.Stat(filePath)
+	return !errors.Is(err, os.ErrNotExist)
 }
 
 // fileWithinWorkingDir reports whether filePath resolves to a location inside
@@ -1086,6 +1105,13 @@ func ensureAttachmentWithinWorkdir(cmd *cobra.Command, filePath string) error {
 		return fmt.Errorf("resolve --attachment path %q: %w", filePath, err)
 	}
 	if !within {
+		if !pathExists(filePath) {
+			return fmt.Errorf(
+				"--attachment path %q does not exist; it also resolves outside the current working "+
+					"directory, so --allow-external-file would not make this upload succeed. Generate the "+
+					"file inside the task workdir and attach that path.",
+				filePath)
+		}
 		return fmt.Errorf(
 			"--attachment path %q resolves outside the current working directory; "+
 				"attach files generated inside the run workdir rather than machine-shared "+
--- a/server/cmd/multica/cmd_issue_test.go
+++ b/server/cmd/multica/cmd_issue_test.go
@@ -455,6 +455,48 @@ func TestEnsureFileFlagWithinWorkdirReportsMissingFileAsMissing(t *testing.T) {
 	}
 }
 
+// TestExternalFileErrorLeadsWithMissingFile pins the remaining half of the
+// misleading diagnosis. For a path that is genuinely outside the workdir AND
+// does not exist, the guard used to advise --allow-external-file and nothing
+// else, so the agent's reasonable next move was to disable the guard and retry
+// — earning a second, unrelated "no such file" error. There is no stale file to
+// protect anyone from when the path does not exist, so the missing file is the
+// fact that leads.
+func TestExternalFileErrorLeadsWithMissingFile(t *testing.T) {
+	outside := withWorkdirShape(t, false)
+	cmd := &cobra.Command{Use: "test"}
+	cmd.Flags().Bool("allow-external-file", false, "")
+
+	missing := filepath.Join(outside, "desc.md")
+	err := ensureFileFlagWithinWorkdir(cmd, "description-file", "description", missing)
+	if err == nil {
+		t.Fatal("expected an error for a missing path outside the workdir")
+	}
+	if !strings.Contains(err.Error(), "does not exist") {
+		t.Errorf("error should lead with the missing file, got: %v", err)
+	}
+	if !strings.Contains(err.Error(), "outside the current working directory") {
+		t.Errorf("error should still record that the path is external, got: %v", err)
+	}
+
+	// An existing external file is the case the guard was built for: the
+	// wording must keep pointing at the escape hatch.
+	stale := filepath.Join(outside, "stale.md")
+	if err := os.WriteFile(stale, []byte("x"), 0o644); err != nil {
+		t.Fatalf("write fixture: %v", err)
+	}
+	err = ensureAttachmentWithinWorkdir(cmd, stale)
+	if err == nil {
+		t.Fatal("expected an error for an existing path outside the workdir")
+	}
+	if strings.Contains(err.Error(), "does not exist") {
+		t.Errorf("an existing file must not be reported as missing, got: %v", err)
+	}
+	if !strings.Contains(err.Error(), "allow-external-file") {
+		t.Errorf("error should point at --allow-external-file, got: %v", err)
+	}
+}
+
 // TestEnsureAttachmentWithinWorkdir covers the MUL-4252 guardrail extended to
 // --attachment: a local attachment path outside the task workdir is rejected
 // (so an agent can't attach another run's stale /tmp file), with the same
--- a/server/cmd/multica/cmd_issue_test.go
+++ b/server/cmd/multica/cmd_issue_test.go
@@ -300,6 +300,161 @@ func TestResolveTextFlag(t *testing.T) {
 	})
 }
 
+// withWorkdirShape chdirs into a workdir that is either reached through a
+// symlink (logical) or is its own canonical path, and returns the temp dir the
+// case may use as "outside". BOTH shapes matter, and for opposite reasons:
+//
+//   - logical: os.Getwd() prefers $PWD when it names the current directory, and
+//     everything that sets $PWD carries the path as typed — a shell's `cd`,
+//     testing.T.Chdir, and the PWD the daemon exports to agent processes (see
+//     pkg/agent/opencode.go). This is the shape where an unresolved candidate
+//     reads as "outside the workdir" though it plainly is not.
+//   - canonical: this is the shape where an unresolved candidate reads as
+//     "inside the workdir" although a symlink takes it out — so it is the only
+//     shape in which the escape-hatch cases below can fail. Run them only under
+//     the logical shape and they pass no matter what the code does, because
+//     there everything unresolved looks outside.
+func withWorkdirShape(t *testing.T, canonical bool) (outside string) {
+	t.Helper()
+	root := t.TempDir()
+	physical := filepath.Join(root, "physical")
+	if err := os.MkdirAll(physical, 0o755); err != nil {
+		t.Fatalf("mkdir workdir: %v", err)
+	}
+	workdir := filepath.Join(root, "logical")
+	if err := os.Symlink(physical, workdir); err != nil {
+		t.Skipf("symlinks unavailable: %v", err)
+	}
+	if canonical {
+		resolved, err := filepath.EvalSymlinks(physical)
+		if err != nil {
+			t.Fatalf("resolve workdir: %v", err)
+		}
+		workdir = resolved
+	}
+	t.Chdir(workdir)
+	return t.TempDir()
+}
+
+// TestFileWithinWorkingDir covers the containment predicate behind the
+// MUL-4252 guardrail directly, because the callers only ever exercise it with
+// files that already exist — and existence is precisely what used to hide the
+// bug. A candidate that does not exist yet cannot go through
+// filepath.EvalSymlinks, and the fallbacks it used to have (one filepath.Dir
+// step, then filepath.Clean) left it unresolved as soon as an intermediate
+// directory was missing too. That failed in both directions:
+//
+//   - Against a workdir reached through a symlink, `subdir/report.md` read as
+//     outside the workdir. The command then reported a missing file as a
+//     guardrail violation and advised --allow-external-file — the one move that
+//     makes things worse, since it only disables the guard.
+//   - Against a canonical workdir, `escape/sub/report.md` — where `escape` is a
+//     symlink out of the workdir — read as INSIDE it, so the guard admitted a
+//     path that resolves to a machine-shared directory.
+func TestFileWithinWorkingDir(t *testing.T) {
+	for _, shape := range []struct {
+		name      string
+		canonical bool
+	}{
+		{name: "workdir reached through a symlink", canonical: false},
+		{name: "canonical workdir", canonical: true},
+	} {
+		t.Run(shape.name, func(t *testing.T) {
+			outside := withWorkdirShape(t, shape.canonical)
+			sep := string(filepath.Separator)
+			if err := os.WriteFile("exists.txt", []byte("x"), 0o644); err != nil {
+				t.Fatalf("write fixture: %v", err)
+			}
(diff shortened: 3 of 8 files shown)


GODVvVZzz:
@multica-eve request review

multica-eve (changes requested):
Request changes: the product direction is correct—the CLI must reject any file that the OS will actually read outside the task workdir—but the current Windows implementation can still validate one path and read a different one.

The fix is necessary and should be accepted once these two correctness gaps are closed:

1. **Must fix — preserve Windows rooted-path semantics.** In `server/internal/util/path.go` (`absNoClean`, around lines 85–94), a root-relative path such as `\\tmp\\desc.md` is not considered absolute by `filepath.IsAbs`, so it is prefixed with the current workdir. With a workdir of `C:\\task\\workdir`, the guard can inspect `C:\\task\\workdir\\tmp\\desc.md`, while `os.ReadFile("\\tmp\\desc.md")` reads `C:\\tmp\\desc.md`. If both files exist, the in-workdir shadow can make an outside read pass the guard. Preserve Windows path kinds (root-relative, drive-relative, drive-absolute, and UNC) and add a regression test with a real root-relative target plus an in-workdir shadow; the guard must reject it.

2. **Must fix — junction escapes must not be treated as an accepted fallback.** In `server/cmd/multica/workdir_guard_windows_test.go` (around lines 181–210), the junction case explicitly accepts `within == true` when `EvalSymlinks` fails. Directory junctions are common and do not require administrator privileges, so this test locks in a known contract violation: `workdir/escape/file` may resolve outside the workdir and still be allowed. Resolve junctions reliably and flip the test to require rejection.

Validation was performed against head `3ec1781e`: the synthetic merge with current `main` was clean, `git diff --check` passed, and the targeted CLI/util/execenv tests passed. The full `cmd/multica` suite could not complete because the task environment's running daemon owns its marker; the same representative failure reproduces on current `main`, so it does not change the two blockers above.

GODVvVZzz:
Both blockers from the review (validated at `3ec1781e`) are addressed in d3d370377; the PR description now carries a section detailing each.

**1. Windows rooted/drive-relative paths** — `absNoClean` no longer prefixes the workdir onto them. `\tmp\desc.md` resolves against the volume root and `C:tmp\desc.md` against the current directory on C:, which is what `os.ReadFile` actually opens; the in-workdir shadow can no longer pass. A drive-relative path on an unobservable drive is handed back uncanonicalized and fails closed through the guard's `filepath.Rel` branch. Regression tests: root-relative with the shadow present must be rejected, same-drive drive-relative must still resolve into the workdir, cross-drive must fail closed.

**2. Junctions resolve** — resolution is component-wise on Windows (`evalPath`) and follows junctions via `os.Readlink` under both `winsymlink` settings, with a bounded follow budget. The known-gap test is replaced: junction out of the workdir is rejected, junction into it (pnpm shape) is still admitted, a chain is rejected, a dead target stays inert (the kernel cannot open it either — the not-found surfaces from `os.ReadFile`). Reparse points `Readlink` cannot speak for (OneDrive placeholders) are treated as opaque, since the kernel opens them where they stand.

One deliberate side effect of the restructure, pinned by a new test: the walk now applies `..` to the resolved prefix even when the tail is missing, where it previously started from `filepath.Abs` (which cleans first). Behavior for every shape the existing tests cover is unchanged — the full `cmd/multica`, `internal/util` and `internal/daemon/execenv` suites pass unmodified with `-race` on macOS, and `GOOS=windows` build+vet are clean.

Re-requesting review.

multica-eve (changes requested):
Request changes: head `d3d370377` closes the two blockers from the previous review, but the new Windows resolver still has cases where the guard validates an in-workdir spelling while the kernel follows a different, outside path.

The product change remains necessary and should be accepted once these remaining fail-open cases are closed:

1. **Must fix — classify `os.Readlink` targets with the full Windows path grammar.** In `server/internal/util/path_windows.go:119-126`, `resolved` is reset only when `filepath.IsAbs(target)` is true. A root-relative link target such as `\outside` is not `IsAbs`, but Windows resolves it from the current volume root. The resolver instead keeps the link's parent as the base. With a same-named shadow under the workdir, the guard can approve the shadow while `os.ReadFile` opens `C:\outside`. The same normalization path corrupts volume-GUID targets at lines 137-145: stripping `\\?\` from `\\?\Volume{GUID}\...` turns an absolute mount target into a relative path. Reuse the complete Windows path-kind handling for link targets, preserve the volume-GUID namespace, and add negative tests for a root-relative symlink target plus shadow and a volume-GUID mount target.

2. **Must fix — do not treat every unsupported `ModeIrregular` reparse point as non-redirecting.** Lines 96-104 continue lexically whenever `os.Readlink` rejects an irregular reparse point. That set is broader than OneDrive placeholders: for example, `IO_REPARSE_TAG_WCI_LINK` is a name-surrogate representing another named entity, while Go's `os.Readlink` decodes only symlink and mount-point tags. A Windows container/filter can therefore follow the component elsewhere while the guard keeps comparing its lexical workdir location. This is an access-control boundary: inspect the reparse tag and either resolve name-surrogates through a handle/final path or fail closed for an unknown surrogate. `Readlink` failure cannot mean “does not redirect.” See Microsoft's [reparse-tag contract](https://learn.microsoft.com/en-us/windows/win32/fileio/reparse-point-tags).

Validation against this head: synthetic merge with current `main` is clean; `git diff --check` passes; targeted `cmd/multica` race tests, `internal/util` race tests, the full `internal/daemon/execenv` package, and the required `ByteIdentical` tests pass on macOS; both affected packages cross-compile for Windows. The latest Windows CI executed the existing root-relative-input and junction cases successfully and all 19 checks are green, but none of the link/reparse-target cases above is covered, so green CI does not clear these blockers.

GODVvVZzz:
Both blockers from the review of `d3d370377` are addressed in ca96a79a0; the PR description carries the details.

**1. Link targets classified with the full grammar** — `classifyTarget` applies the same path-kind rules to targets that `absNoClean` applies to input: a root-relative target resolves against the volume root the link sits on (never the link's parent), and `\?\Volume{GUID}\...` is kept verbatim in the device namespace so the comparison fails closed instead of splicing a mangled relative path. `trimDeviceNamespace` is gone; device-qualified and UNC extended-length forms are reduced to their Win32 spelling so they can still be walked and compared (the prefix rules mirror daemon's #6883 handling). Both negative tests the review asked for are in, and each verifies the kernel's target **by content** — shadow and target carry different bytes, so the fixture cannot silently invert.

**2. Readlink failure is not "does not redirect"** — a Readlink-failing reparse point is resolved through `GetFinalPathNameByHandle` (zero desired access — no hydration — plus `FILE_FLAG_BACKUP_SEMANTICS` for directories, via `x/sys/windows` like the rest of the repo). Equal to the component → opaque (OneDrive placeholder). Different → the walk continues from the final path. Unknowable → the new `util.ErrUnresolvablePath`, which `fileWithinWorkingDir` treats as outside and `execenv`'s delegate reduces to a form its containment comparison rejects.

The signature change makes the fail-closed case explicit rather than a look-alike path; Unix behavior is unchanged for every shape the existing tests cover, and the full affected suites pass with `-race` on macOS.

GODVvVZzz:
The windows-execenv failure on ca96a79a0 is fixed in 3ee1a956c: `classifyTarget` asked `VolumeName` before `IsAbs`, and `VolumeName` is non-empty for a drive-absolute path (`C:\dir`) too — so a junction's absolute target was misread as drive-relative, resolved against the working directory (a nonexistent spelling), errored, and the escape fell back to a lexical judgment inside the workdir. The two expect-true junction cases were passing through that same wrong path.

The IsAbs branch is now asked first, and the ordering is pinned by `TestClassifyTarget` — a fixture-free table over the full link-target grammar (drive-absolute, UNC, root-relative, drive-relative both ways, extended-length and object-manager forms, volume-GUID verbatim) that catches this class of bug without needing any filesystem fixture.

GODVvVZzz:
CI is green on 25affcaec (17 pass), including a new windows-execenv step that runs `TestClassifyTarget` — the fixture-free link-target grammar table — in `internal/util`, which previously had no Windows execution point. The junction, root-relative-symlink-target and volume-GUID cases in cmd/multica all pass with the target verified by content.

GODVvVZzz:
Head moved past the closure comment above — three commits landed after it, so the head under review is now `25affcaec`. Both blockers from this review remain closed on it.

The one that needs explaining is `3ee1a956c`: it fixes a fail-open regression `ca96a79a0` itself introduced. In `classifyTarget`, the drive-relative branch ran before the `IsAbs` branch. `filepath.VolumeName` is non-empty for a drive-absolute path (`C:\dir`) too, so a junction's absolute target was misread as drive-relative and resolved against the working directory — the same "the guard approves one path, the kernel opens another" shape as blocker 1, and it read a junction escape as inside the workdir on one CI run of `ca96a79a0`. The `IsAbs` branch now precedes the `VolumeName` check. The fixture-free `TestClassifyTarget` table pins that ordering, `527aec0b5` runs the table in the `windows-execenv` job so the branch order is exercised on Windows, and `25affcaec` corrects two of the table's own expectations (the walk base for drive-relative and plain-relative targets) that had been written against the wrong shapes.

Issue #7833, opened 2026-08-31
[Bug]: Cursor background shell can prevent terminal result and leave task running after delivery


WitMiao:
### Deployment type

self-host

### Area

Agents and runtimes

### What happened?

A Cursor-backed one-shot task can remain `running` after its user-visible work has already been delivered when the agent started a long-lived background shell, such as a Vite development server.

In the affected sequence:

1. Cursor calls its `Shell` tool with `block_until_ms: 0` and receives an `isBackground: true` result.
2. The agent completes the requested verification and posts its final issue comment.
3. The top-level Cursor turn ends, but `cursor-agent -p --output-format stream-json` does not emit a terminal `{"type":"result"}` event or exit while the background shell is alive.
4. Multica therefore keeps the task in `running` and continues occupying a runtime slot. The only generic liveness bound is the two-hour idle watchdog.

This was reproduced on three independent tasks with Cursor Agent `2026.08.25-3e8eec8`. In each case the long-lived tool process was a background development server. The final comment being visible is only a tool side effect and should not itself be treated as proof of successful task completion.

There is also a secondary cleanup problem: a background shell can create a new process group. If the Cursor root process later exits, that shell can be reparented and continue running outside the process group Multica originally owned.

Expected behavior:

- A tool-started background process should not silently keep a completed top-level Cursor turn in `running` for up to two hours.
- Multica should either prevent this background launch mode for managed Cursor runs, or track the background process explicitly and surface/clean it up during cancellation or terminal handling.
- Any fallback must remain fail-closed: an assistant message, `step_finish`, or a successfully posted comment is not a safe substitute for Cursor's terminal protocol result because more tool calls may still follow.

### Steps to reproduce

1. Register a Cursor Agent runtime and use Cursor Agent `2026.08.25-3e8eec8` (or another version with the same behavior).
2. Trigger an issue task that asks the agent to start a Vite development server in the background, verify the page, and report the result.
3. Confirm that Cursor uses `Shell(block_until_ms: 0)` and reports the shell as background work.
4. Let the agent finish its verification and post its final issue comment without explicitly stopping the development server.
5. Observe that the issue comment is present, but the Multica task remains `running` and the `cursor-agent` process does not exit.
6. Inspect the process tree after the Cursor root eventually exits or the task is cancelled; depending on how Cursor launched the shell, the development server may remain alive in a separate process group.

### Screenshots (optional)

[...]

PR #8042, opened 2026-09-04
fix(cursor): fail managed runs on background shell launch

makesomethingshit:
## What does this PR do?

Cursor can report a successful background `shellToolCall` and then never emit its terminal result while the shell remains alive. Multica currently waits for that result/process exit, so the task can retain a runtime slot until the global idle watchdog.

**Root cause:** the Cursor stream already reports `"isBackground": true` on the completed shell result, but `parseCursorToolCall` currently discards that lifecycle meaning. The backend only cancels Cursor proactively after a terminal result — precisely the event missing in this failure mode.

**Fix:** parse the authoritative `shellToolCall.result.isBackground` (JSON root boolean only, no string/command heuristics). When a completed shell call reports `true`, mark a provider-specific foreground-execution guard, cancel the Cursor managed process using the existing runtime ownership path (`cancel()` → owned process group SIGKILL), and preserve the guard as the final failure reason instead of letting context cancellation rewrite it to generic `aborted`. Partial assistant output is never exposed as `Result.Output`, and completion is never synthesized from assistant text or tool side effects.

This PR changes only the Cursor adapter rather than weakening global watchdog semantics or adding generic process-tree machinery.

The primary fix closes the task-slot/liveness failure by failing the managed Cursor run as soon as Cursor confirms that a shell was launched in background mode and by cancelling the runtime through the existing owned-process-tree path.

A Unix descendant that deliberately leaves that process group is the broader process-ownership limitation and is not solved here without a provider-supplied process identity (see #7615 as a related follow-up).

## Related Issue

Addresses #7833 (deliberately not auto-close: the secondary process-group-escape limitation tracked in #7615 remains open)

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/cursor.go`
  - Add `Background bool` to `cursorToolCall`; `parseCursorToolCall` structurally reads the JSON root `isBackground` boolean of the completed shell result (raw `Result` preserved; malformed results never fail the whole parse; no string/command heuristics).
  - In the `completed` tool_call handler, after the normal `MessageToolResult` is forwarded, a `shell` call with `Background == true` sets `backgroundShellDetected` and invokes the existing `cancel()` path (no new process-tree helpers; `proc_*.go`/`launch.go` untouched).
  - Finalization precedence: background guard > terminal result > timeout > external cancellation > scanner/protocol/prompt-write/exit errors > missing result. Guarded runs end `failed` with the new constant `cursorBackgroundShellGuardError` ("cursor-agent launched a background shell; Multica-managed runs require foreground execution") instead of generic "execution cancelled"; `cursorFailureDiagnostic` and stderr tail sanitization are preserved; `Result.Output` stays empty and the session ID is preserved.
- `server/pkg/agent/cursor_test.go`
  - Extend the `TestParseCursorToolCall` table with `wantBackground`: background result (`isBackground: true`), explicit false, false-positive guard (stdout containing the text `"isBackground":true` must not trip detection), malformed/missing result.
  - Add end-to-end `TestCursorExecuteFailsPromptlyOnBackgroundShellWithoutTerminalResult` using a fake Cursor CLI (outer ctx ~10s, exec timeout ~8s, asserts return well within 5s): `Status == "failed"`, error mentions background shell/foreground execution, not `aborted`/`timeout`, `Result.Output == ""`, `SessionID == "sess-cursor-bg"`. This test is RED on unpatched main and GREEN after the fix.
- `server/pkg/agent/claude_deadlock_test.go`
  - Package-wide `TestMain`: add `CURSOR_FAKE_MODE=background_shell_no_result` dispatch and `runFakeCursorBackgroundShellNoResult()` helper (reads stdin to EOF, emits init → tool_call started → completed with `result.isBackground: true`, then blocks without a terminal result — reproducing the exact #7833 state).

## How to Test

1. Red check on unpatched main (with the new tests): `go test ./pkg/agent -run 'TestCursorExecuteFailsPromptlyOnBackgroundShellWithoutTerminalResult' -count=1` → FAIL (task does not return before timeout).
2. After the fix: `go test ./pkg/agent -run 'TestParseCursorToolCall|TestCursorExecuteFailsPromptlyOnBackgroundShellWithoutTerminalResult' -count=1` → PASS.
3. Full package + vet + whitespace: `go test ./pkg/agent -count=1`, `go vet ./pkg/agent`, `git diff --check` → PASS (on Windows runners a pre-existing set of unrelated environment failures exists; the failure set is identical before/after the patch, added=0/removed=0).

## Checklist

## AI Disclosure

**AI tool used:** Multica Agent (multi-agent squad pipeline)

**Prompt / approach:**
Implemented by a Go-focused agent worker from a written implementation spec, independently cross-verified by a second agent worker (diff review against the spec + local re-run of vet/tests with a baseline failure-set comparison), followed by an automated pre-flight check of conventions and this PR template. Verification evidence: commit `9637a36b384b834100816d82306eb9f6eb8583ea`, RED→GREEN for `TestCursorExecuteFailsPromptlyOnBackgroundShellWithoutTerminalResult`, `go vet ./pkg/agent` PASS, `git diff --check` PASS.

### DIFF
--- a/server/pkg/agent/claude_deadlock_test.go
+++ b/server/pkg/agent/claude_deadlock_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"strings"
@@ -27,6 +28,15 @@ func TestMain(m *testing.M) {
 	if os.Getenv(opencodeStdinHelperEnv) == "1" {
 		runFakeOpencodeStdinHelper()
 		os.Exit(0)
+	}
+	switch mode := os.Getenv("CURSOR_FAKE_MODE"); mode {
+	case "":
+	case "background_shell_no_result":
+		runFakeCursorBackgroundShellNoResult()
+		os.Exit(0)
+	default:
+		fmt.Fprintf(os.Stderr, "unknown CURSOR_FAKE_MODE: %q\n", mode)
+		os.Exit(2)
 	}
 	switch mode := os.Getenv("CLAUDE_FAKE_MODE"); mode {
 	case "":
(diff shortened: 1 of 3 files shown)


makesomethingshit:
**Superseded notice:** This PR implements the opposite behavior (immediate cancel + forced `failed` on background-shell observation) from what we converged on after further analysis in #7833: a background shell itself is not the failure — the pathological case is a stalled run with no semantic progress and no terminal `result`. The redesign (background liveness guard, terminal-result precedence, fail-closed semantics with regression tests A–F+G) is in **#8050**, which supersedes this one. Closing as superseded once #8050 is ready for review — please direct review attention to #8050.

makesomethingshit:
Closing as superseded by #8050 (approved by the contributor team lead; see the superseded notice above).

PR #8046, opened 2026-09-04
MUL-7065 CODI-10 feat(daemon): make tool_result truncation explicit, UTF-8 safe, and visible

importcpp:
Implements CODI-10.

## The bug

`daemon.go` cut tool results at a raw 8192-byte boundary and recorded nothing about having done so. The execution log then rendered a prefix that ends naturally, so a truncated result reads as a complete one.

The harm was unevenly distributed. The UI's truncation hint fired on a character count (8000) while the cut ran on bytes: 8192 bytes of ASCII crosses that threshold and shows the hint, but the same byte budget holds only 2730 characters of Chinese, which does not. A Chinese-language user lost roughly two thirds of an output and saw no marker at all.

The same line of code produced two more consequences. Screenshot base64 was cut mid-string and rendered as 8 KB of garbage text instead of an image. And credentials spanning the cut escaped redaction, because redaction runs server-side, after truncation, and several rules only match as a whole — a connection string needs its `@`, a JWT needs its second `.`.

## The change

A preview helper on the daemon side: each redaction rule declares `partialAtEOF`, and the window rolls back to before any position that could begin a cross-boundary secret. Two nullable columns carried end to end. Three display surfaces that distinguish "the source was truncated" from "only part is shown here".

Three states stay distinguishable end to end: absent / false / true. Collapsing absent into false would assert completeness for every historical row — exactly the bug being fixed. Hence a pointer rather than `bool` with `omitempty`, nullable columns with no default, a `text[]` batch array (sqlc maps `boolean[]` to `[]bool`, which cannot carry per-row NULL), and `=== true` at the render sites. Historical rows cannot be backfilled: the original length was never recorded.

## Verification

The security argument rests on a falsifiable oracle. 3048 cut points are enumerated, and the criterion is that a preview must not expose more than redacting the full text and then truncating it would. `TestOracleRejectsTrivialTruncation` proves the assertion can fail: a drop-one-byte stub implementation is caught at 877 cut points.

Test counts: views 4922, core 1720, mobile 152. Typecheck clean across all workspaces; ESLint reports no errors in the touched packages.

## Two notes for review

Redaction ordering is outside the issue's written scope, which says only "preserve existing redaction semantics". It was kept in because the vulnerability exists in production today, and moving the truncation boundary without fixing it would merely reshuffle which offsets leak.

The `*bool` three-state handling is the highest-risk part of the diff. Getting it wrong produces permanent data ambiguity, and it would only surface when someone queries a field that was never written.

## Out of scope

Recovering the discarded content. That needs artifact storage (CODI-13 / CODI-14), so an over-budget image is a placeholder notice rather than a restored image.


Bohan-J:
Thanks for this, and for the review rounds visible in the commit history — the progression from a vacuous oracle to `TestOracleRejectsTrivialTruncation` is genuinely good work, and finding your own `{name}` vs `{{name}}` bug by writing render assertions is exactly the right instinct.

We're going to close this rather than merge it. Not because the problems aren't real — I verified all three independently before writing this:

- `daemon.go` did cut at a raw 8192-byte boundary with nothing recording it.
- The byte/character mismatch is real and it is not symmetric. I measured it: 306 KB of Chinese truncates to 8190 bytes = **2752 characters**, well under the UI's 8000-character hint threshold, so a CJK user lost ~97% of an output and saw no marker; the same budget of ASCII is 8189 characters and does trigger the hint.
- Redaction running after truncation does leave a straddling credential's visible head in plaintext.

The problem is scope, and it's worth being precise about why that's disqualifying rather than a style preference.

## This is three changes, not one

By line count the diff is 2744, but 1572 of that is tests and 42% of the production lines are comments — the actual new logic is roughly 430 lines. Those 430 lines do three separable things:

1. **UTF-8 correctness** — the byte cut slicing into a multi-byte rune (~60 lines).
2. **Truncation visibility** — two nullable columns, the protocol fields, three render surfaces (~250 lines).
3. **Redaction ordering** — `redact.go` (+305) with `preview_test.go` (+346).

Item 3 is ~24% of the diff, and the PR description already acknowledges it is outside the issue's written scope ("preserve existing redaction semantics"). It rewrites a globally shared security package, and it establishes a standing invariant: **every future redaction rule must ship a matching `partialAtEOF` matcher, and a matcher that is narrower than its rule is a hole** — your own comment says this. `TestEverySecretPatternHasPartial` catches "forgot to write one"; nothing catches "wrote it too narrow."

That is a decision we need to make deliberately, on its own, with its own revert story. Bundled with a UI and schema change, reverting the truncation work drags the security fix out with it, and vice versa. That coupling is the blocker.

## Findings that stand on their own

Independent of scope, four things need fixing:

1. **`redact.Text()` on the daemon's hot path, result discarded.** The fast path is `len(norm) <= budget && len(redact.Text(norm)) <= budget` — the redacted string is used only to measure length, then thrown away. Measured on an M-series machine: **6.5 ms per 8 KB** of ordinary log text (1.26 MB/s), which matches your own ~0.6 µs/byte figure. It sits inline in `executeAndDrain`'s drain loop, and `trySend` (`pkg/agent/claude.go:692-698`) drops messages when the channel is full, so slowing the drain costs live transcript rows. The server then redacts the same bytes again at `handler/daemon.go:4681`.

   The check may not be needed at all: `task_message.output` is `TEXT` with no length constraint (`migrations/026_task_messages.up.sql:9`), so 8192 is a soft budget and overshooting it by a few KB is harmless. It also has a semantic cost — output that was never truncated, only redaction-expanded, gets `truncated=true` and renders as "source truncated", which is the same class of false statement the PR exists to remove. `TestToolOutputPreviewAccountsForRedactionGrowth` locks that in.

2. **The in-band notice duplicates or vanishes.** Nothing strips `[… truncated, N bytes total]` client-side (`grep "bytes total"` across `packages/` and `apps/` is empty). On CJK the stored preview is ~2752 characters, under every display-clip threshold, so the raw English marker renders inside the code block *alongside* the localized badge. On ASCII the preview is ~8189 characters, so the display clip removes the marker entirely and it is never seen. Neither behavior is tested.

3. **`TruncationUnknownNotice` is per-turn in web chat.** `chat-message-list.tsx:1081` puts it inside `TimelineView`, which renders per message (`:611`, `:1041`), as the first element in the bubble. With no backfill, every historical turn gets the disclaimer on deploy — and since the daemon is user-installed and lags, new turns from older daemons get it too. The transcript dialog does it once per transcript, which is what the component's own doc comment describes.

4. **`text-warning-foreground` is not a token in this repo.** `output-truncation.tsx:48`. Only `--color-warning` exists (`packages/ui/styles/tokens.css:35`); Tailwind v4 emits nothing for an undefined `--color-*`, so the badge text inherits `text-faint-foreground` — grey on amber. The sibling badge at `agents/components/agent-live-peek-card.tsx:233` uses `text-warning`.

Two smaller ones worth carrying forward: `ToolResultOutput`'s doc diagnoses the `json.Unmarshal`-into-string / `null` bug precisely, but the live instance of that exact pattern is still at `qwen.go:391-397`. And `ToolResultOutput` was wired into only `claude.go` and `codebuddy.go` — the other ~10 backends neither unwrap nor set `OutputIsImage`, so "the size means the same thing regardless of which provider produced it" doesn't hold as shipped, and oversized images still get byte-sliced everywhere else.

## If you want to pick this back up

Please do — the underlying work is sound. Split it:

- **PR A — UTF-8 safe truncation.** The rune-boundary fix and the preview helper, with the redaction-growth check dropped (see #1). Small, self-contained, mergeable on its own.
- **PR B — truncation metadata and UI.** Columns, protocol fields, render surfaces. Depends on A. Please settle the in-band notice (#2) and the unknown-notice granularity (#3) first — both are product calls, and we'd rather answer them than have you guess.
- **PR C — redaction ordering.** Its own PR, reviewed as a security change. Open an issue first and let us weigh the standing `partialAtEOF` invariant before you write it; if we decide the maintenance cost isn't worth it, we'd rather tell you that than have you rewrite 650 lines twice.

For B, one thing to preserve: the `*bool` / nullable-column / `text[]` tri-state reasoning is correct and the comments explaining it are worth keeping verbatim. That part we'd merge as-is.

PR #8050, opened 2026-09-04
MUL-7063 fix(cursor): bound stalled runs after background shell launch

makesomethingshit:
## What does this PR do?

Cursor reports a background shell's launch `tool_call` as completed while its process remains alive. Previously, that completion decremented message-based tool accounting and moved still-running work onto the shorter idle watchdog budget.

Keep verified background work represented in backend-native `Session.ToolActivity`. The existing daemon tool watchdog owns the budget. At expiration, stop verified background work, release its original launch result once, and give Cursor a fresh normal watchdog budget to emit its authoritative terminal result. No Cursor-private inactivity timer or new global timeout is introduced.

Related: #7833 and #8042. A background launch is not itself a fatal error.

## Implementation

- Recognize only the root boolean `shellToolCall.result.isBackground == true`; read `result.success.pid` separately and preserve the raw Cursor tool result. No command-string, stdout or assistant-text heuristics.
- Native tool accounting is independent of the optional transcript. Verified live background work remains in flight until it ends or cleanup is confirmed. Ordinary background silence uses the existing tool budget.
- A launch whose ownership **was never captured** returns its raw launch result immediately. Once no other tools are in flight, the ordinary idle budget applies even when `MULTICA_AGENT_TOOL_WATCHDOG=0`. This is launch completion, not a claim of successful process cleanup, and it grants no cleanup recovery window.
- A failure while checking or cleaning **already-claimed** work does not decrement native accounting. `Close` retries the shared cleanup path once before releasing resources. Persistent failures are logged as unconfirmed; the raw launch payload is preserved without fabricating successful native cleanup.
- Final watchdog revalidation refreshes native tool state before reading its timestamp. Fresh activity aborts the stale timeout attempt even if the count remains positive. Later genuine inactivity still fires.
- Capture verifies ancestry and target identity. Post-capture validation still rejects a changed target; it no longer discards a valid target solely because the Cursor root exited meanwhile. Darwin can also prove the original ancestry of an unreaped root while draining its final events.
- Cancellation and normal finalization use the same process cleanup path. **User-visible change:** a background server started by the run is stopped when that run finishes normally, rather than intentionally surviving its completion.

### Process ownership

**Linux:** retain a process-group pidfd after ancestry/start-identity validation. `PIDFD_SIGNAL_PROCESS_GROUP` targets the original kernel group, including late children after leader exit. Linux 6.9+ and available pidfd syscalls are required. Unsupported capture returns the launch result for the ordinary idle fallback.

**macOS:** no `/bin/cat` group anchor and no requirement to join Cursor's session. The real `setsid()` topology is covered by the fixture. Query XNU process unique IDs and original-parent unique IDs, retain observed birth identities, and admit only members whose identity or parent lineage links to that claim. Send each signal using `PROC_INFO_CALL_SIGNAL_AUDITTOKEN`: the kernel validates PID plus pidversion and holds the matching process reference while signalling. PID reuse or exec between observation and signal rejects the stale request instead of targeting a replacement. Cleanup re-scans and re-signals newly proven members within the existing one-second cleanup bound.

This uses the existing Go syscall facility with no cgo or new dependency. The private 56-byte process-info ABI and 32-byte audit token are checked by regression tests. Sources: [XNU process-info ABI](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_info_private.h), [kernel audit-token signalling](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/proc_info.c), [original parent identity](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_internal.h).

**Windows:** the existing nested Job Object implementation is retained. Parents are assigned before pre-existing children are enumerated; children created after assignment inherit the job. The real late-child regression covers cleanup after leader exit.

### Limits and risks

- The macOS process-info and audit-token calls are **private OS interfaces**. Capture probes availability using a non-existent PID; unsupported kernels or denied queries fail closed. This PR does not change release builds to require cgo and does not fall back to a numeric PID/PGID kill.
- A late direct child can be identified after its parent exits if that parent's unique ID was retained. However, if an **unobserved intermediate generation exits before it can be linked**, the remaining descendant's origin cannot be proven. An orphan that execs before its original lineage is observed can also lose that link. Such work remains un-signalled. Complete cleanup of arbitrary detached trees on macOS is not claimed.
- Work that deliberately leaves the tracked Unix group is outside the group scan. Persistent identity, permission or termination failures can also leave work alive after finalization. Cleanup is then explicitly logged as unconfirmed; no recovery window or successful cleanup is fabricated.
- Never-claimed launches use the ordinary idle fallback. Verified active tools still honor the documented disabled tool budget, and `MULTICA_AGENT_IDLE_WATCHDOG=0` still disables the watchdog suite. Global defaults and other agent backends are unchanged.

## Changed files

- `server/pkg/agent/cursor.go`, `agent.go`, `cursor_background.go`: structural parsing, native activity, watchdog interruption and shared finalization/accounting.
- `server/pkg/agent/cursor_background_process*.go`: shared ownership checks, Linux pidfds, Darwin birth-identity lineage and kernel-validated signals, retained Windows Jobs and unsupported-platform behavior.
- `server/internal/daemon/daemon.go`: tool-budget integration and refreshed final revalidation; remove the internal marker comment.
- Cursor lifecycle/process tests, `cursor_test.go`, `claude_deadlock_test.go`, `cursor_background_watchdog_test.go`, and `cursor_uncaptured_watchdog_test.go`: real subprocess topology, native result/accounting, root-exit race, stale PID-version rejection and disabled-budget regression.
- `.github/workflows/ci.yml`: Linux/macOS/Windows focused race coverage, non-race and race macOS finalization repeated five times, and cgo-disabled macOS ownership checks. (The PR-only reviewed-head reproduction steps were removed before merge; their results are recorded under Verification below.)
- `apps/docs/content/docs/environment-variables.mdx`: behavior, normal-finalization cleanup and platform limits.

## How to test

```sh
cd server
go test ./pkg/agent -run 'TestCursorBackground|TestCaptureCursorBackground' -count=1
go test ./internal/daemon -run 'Test.*Background.*Watchdog|Test.*IdleWatchdog' -count=1
go test ./pkg/agent -run 'TestCursor|TestParseCursor' -count=1
go test ./internal/daemon -count=1
go vet ./pkg/agent ./internal/daemon
go test -race ./pkg/agent ./internal/daemon
cd ..
git diff --check
```

The subprocess matrix covers natural completion, work beyond the idle budget, owned cleanup at the tool boundary, authoritative Cursor results, multiple processes, external cancellation, normal finalization and unverified ownership. The late-child test gates the fork until after capture, kills the leader without refreshing, then requires child cleanup and survival of an unrelated process. The root-exit test forces root termination during capture. The Darwin native test sends SIGKILL with a deliberately wrong pidversion and requires kernel rejection and survival, then verifies the correct version terminates the fixture.

The watchdog boundary test uses `testing/synctest`: native activity advances during final revalidation while count stays 1; the stale tick must abort and a later inactive budget must fire. The disabled-tool-budget test runs the actual Cursor adapter with an uncapturable launch and requires the idle watchdog to end it, rather than the caller deadline.

## Validation evidence

Current head: `3ac7f6f0b`, rebased onto `main`. Production fixes: `c7f8754e9` (Darwin ownership), `46bd420ff` / `9696f6b1d` / `3ac7f6f0b` (terminal boundary and its hand-off), `3d9d8bafb` (bounded background cleanup); the remaining commits cover CI, docs and tests.

- Windows focused background/process, watchdog and Cursor/parser tests, `go vet`, and `git diff --check`: passed locally. The new uncaptured-accounting test failed on the old behavior (`count=1`, no launch result) and passed after the fix.
- Darwin arm64 agent/daemon cross-compilation and vet passed locally; Darwin amd64 agent compilation also passed with cgo disabled. This is compilation evidence, not macOS execution.
- [CI at `257fa50`](https://github.com/multica-ai/multica/actions/runs/34110918122): Linux full backend and Linux/Windows focused lifecycle race jobs passed. macOS focused lifecycle/watchdog race tests, stale pidversion rejection, finalization repeated five times both with and without race instrumentation, and cgo-disabled native ownership tests passed. The full macOS agent race suite failed three Codex timing fixtures: `TestCodexExecuteSemanticInactivityAllowsContinuousDeltaProgress`, `TestCodexExecuteLegacyFirstTurnMessageSatisfiesProgress`, and `TestCodexExecuteSemanticInactivityAllowsContinuousMessages`. No full macOS pass is claimed; no Kiro failure was reported in this run.
- [Diagnostic CI](https://github.com/multica-ai/multica/actions/runs/34111883779): Linux and Windows lifecycle jobs passed. On macOS, all focused lifecycle/watchdog tests, five-run finalization checks and cgo-disabled ownership tests passed. Both original regressions failed with the intended assertions on `c5cc30e`; the disabled-budget and detached-session defects likewise reproduced on `96915a2`. An extra full macOS package run failed three unchanged Codex timing fixtures that also failed on isolated `96915a2`, plus one Mcode timing fixture. That extra package run is not part of the repository's normal CI matrix; the normal full Linux race suite passed. No backend timeout threshold was changed to mask those failures.
- [Current CI](https://github.com/multica-ai/multica/actions/runs/34113848244) passed. This includes the normal full Linux race suite, Linux/macOS/Windows Cursor lifecycle jobs, macOS cgo-disabled ownership and repeated finalization checks, Windows execution-environment tests, vet, generated SQL verification, vulnerability scanning, and frontend validation.
- Linux and macOS CI executed the real late-child and stale-watchdog regressions and verified their expected failures on isolated `c5cc30e` sources. Round 2 reproductions confirmed the disabled-tool-budget failure on both platforms and the detached-session failure on macOS at `96915a2`. These were assertion failures, not build errors or skipped tests.
- The current full Windows daemon run and the saved `c5cc30e` baseline have exactly the same 68 failing test/subtest names (home/path/symlink and runtime-discovery assumptions), with no new failing names. No full local Windows pass is claimed. The full Linux backend job passed in the linked CI run.
- Local race execution is unavailable because the Windows environment has no supported cgo/C compiler. No local Linux/macOS runtime or real Cursor CLI/browser/authenticated manual end-to-end validation was run.

Codex implemented the changes and integrated read-only reviews of the native ownership/signalling semantics. Main-agent checks and CI execution are distinguished above.


Bohan-J (changes requested):
## Request changes

Thanks for the careful work here — the concurrency design in `cursorBackgroundGuard` is genuinely well built, and the structural `isBackground` read is exactly the right instinct. I verified several of your claims against the real CLI and they hold up.

But I have to block this one, because the guard as designed can fail a run that Cursor would have completed successfully, and I was able to reproduce that with your own test fixture.

I verified everything below against `cursor-agent 2026.08.25-3e8eec8` — the same version reported in #7833 — on macOS/arm64.

---

### What I confirmed the PR gets right

- **`shellToolCall.result.isBackground` is real and is where you say it is.** Live payload confirms the boolean sits at the root of `result`, and the tool key is `shellToolCall` (so `call.Name == "shell"` is correct). Reading it structurally rather than sniffing the command string or stdout is the right call.
- **#7833 reproduces.** Backgrounding `sleep 3000` keeps `cursor-agent` alive indefinitely with no terminal `result`. The problem you are fixing is real.
- **The guard's concurrency is sound.** Single timer owner, coalesced resets via a buffered channel instead of racing `Reset`, `fire()` re-reading the observation under the lock rather than trusting the tick. `go test ./pkg/agent -run 'TestCursor|TestParseCursor' -count=1 -race` passes cleanly here — you mentioned `-race` was unavailable on your machine, so consider that confirmed.
- Fail-closed on `Result.Output` is right, and `fire()` declining to claim a run the deadline or the caller already ended is the correct precedence.

---

### Blocking issue: after a background shell, the stream is silent by design — so the guard is a cap on background-process lifetime, not a progress guard

This is the finding that blocks the PR. Timestamped capture, backgrounding `sleep 90`:

```
t=  8.1s  tool_call completed   isBackground=true    <- guard arms
t= 10.9s  thinking completed                         <- last "semantic progress" event
          ......  87.3 s of complete silence  ......
t= 98.2s  system task_notification                   <- background process finished
t=102.0s  result success                             <- normal terminal result, run succeeds
```

I also checked the interleaving case explicitly (background `sleep 120` followed by a foreground `echo`). Foreground tool calls do interleave normally — but once the model's turn is done, `cursor-agent` still blocks the turn's completion on the background shell and goes silent for **117.5 s**, then emits `task_notification` → `result success`.

So: any Cursor run that launches a background command **necessarily** ends with a silent window equal to the remaining lifetime of that background process. And `cursorSemanticProgressEvent` deliberately excludes `system`, so nothing refreshes the deadline during that window.

That makes `defaultCursorBackgroundGuardTimeout` a **hard 10-minute cap on how long any background command may live**, not a measure of whether the agent is still working. This line in the PR description:

> legitimate workflows that start a background server, test against it, and stop it continue to work

only holds when the background process happens to live less than 10 minutes.

**Reproduced with your own fixture.** `cursorFakeResultWins` is exactly this shape: background shell observed → silence → valid terminal result. Test F runs it with `guard=30s` (longer than the gap) and gets `completed`. Run the identical stream with the guard shorter than the gap — which is what "the background job took more than 10 minutes" means in production — and you get:

```
status="failed"
error="cursor-agent stopped producing semantic progress after launching a background shell and did not emit a terminal result"
output=""
```

A run that Cursor completed successfully, reported as failed, with its terminal result discarded. The only variable is how long the background job ran.

That failure message is also factually wrong in this case: the agent had not stopped producing progress, and it did emit a terminal result.

---

### Root cause is one level up: a background shell stops counting as a tool in flight

The daemon picks its silence budget from an in-flight tool count: `MessageToolUse` increments, `MessageToolResult` decrements. Count > 0 uses the larger tool budget; count == 0 uses the idle budget.

- **1 h foreground test** — `started` sets the count to 1, silence for an hour, `completed` clears it. Covered by the tool budget the whole time. Safe.
- **1 h background test** — `started` sets the count to 1, then `completed` arrives ~2 s later because the shell returned immediately, so the count drops to 0, and *then* the stream goes silent for an hour. The work is still running but nothing is counted as in flight.

That early decrement is the actual defect. And the invariant is already written down in this very file, eight lines above the `completed` case this PR modifies (from the MUL-5231 review):

> A non-terminal (e.g. a future `progress`) or missing subtype must NOT synthesize a result — that would decrement the daemon's in-flight tool count early and **drop a still-running long tool from the tool watchdog onto the shorter idle watchdog, which can force-stop it as falsely stuck**.

A background shell is precisely "a still-running long tool", and it does decrement early. This PR adds a third, even shorter budget on top of that, with no in-flight-tool exemption at all.

Related: `server/internal/daemon/config.go` already reasons this through for the equivalent Codex timer, and reaches the opposite conclusion —

> That one timer therefore stands in for BOTH daemon watchdogs on a Codex run, so it has to be sized like the larger of them — otherwise a quiet test suite or an output-buffering `docker build` dies at the Codex ceiling long before the daemon budget that was supposed to protect it […]

---

### The guard does not clean up the process it fails the run for

Verified process relationships:

```
cursor-agent   pid 38039  pgid 38036
  background shell  pid 38447  pgid 38447   <- its own process group
    sleep 3000      pid 38454  pgid 38447
```

The background shell creates a new process group, so `signalProcessGroup`'s `kill(-cursorPid)` cannot reach it. After killing the `cursor-agent` root, the shell and its child survive, reparented to init. So once the guard fires the real end state is: task failed, runtime slot released, **and the dev server still running with its port held, permanently**.

Note that `result.success` carries `pid` and `shellId` right next to the `isBackground` flag you already parse — and the `pid` it reports **is the leader of that escaped group** (38447 == pgid 38447 in my capture). The handle needed to clean this up is in the object this PR reads and discards.

---

### A direction that does not have the false-positive cost

I tested the inverse action: instead of cancelling the run when the window elapses, terminate only the tracked background process. Backgrounding `sleep 600`, then `kill -9` on the reported pid and nothing else:

```
t=  6.4s  tool_call completed  isBackground=true    (pid 44970, pgid 44970)
          <- kill -9 44970 here
t= 11.7s  system task_notification                  <- cursor-agent notices within ~5 s
t= 19.1s  assistant
t= 19.1s  result success                            <- authoritative terminal result
cursor-agent exit=0, total 19 s instead of 600 s
```

The agent's own final result was:

> DONE. The background `sleep 600` ended with an error (it did not run for the full 600 seconds). Nothing else was requested after that, so no further action was taken.

No stray processes afterwards. That gets all three properties at once: the slot is released promptly, the task ends **completed** with Cursor's authoritative result and an honest explanation instead of a bogus failure, and the escaped process is actually cleaned up.

Concretely, the shape we would like to see:

1. **Keep a background shell counted as in flight.** Observing `isBackground: true` should not let the in-flight tool count drop to zero, so the run stays on the tool budget. After this, a background test and a foreground test behave identically — whatever budget you give the foreground case is what the background case gets.
2. **When the budget is exhausted** (or the task is cancelled, or the run is finishing normally), **terminate the tracked background process group and let Cursor produce its own terminal result** — rather than cancelling the whole run and synthesizing a failure.
3. **Do not hard-code the window in the backend.** Derive it from the existing budgets and make it operator-overridable, following the `CodexSemanticInactivityTimeout` precedent in `daemon/config.go`. As written, `backgroundGuardTimeout` is never assigned anywhere in production code — only in tests — so there is no way to tune or disable it, and it is not covered by `MULTICA_AGENT_IDLE_WATCHDOG=0`, which `config.go` documents as disabling "the whole watchdog suite".

To be explicit about the trade-off this accepts: at default settings this does **not** shorten #7833 from ~2 h to 10 minutes. It changes what happens at the boundary — from "aborted, result lost, process leaked" to "background process stopped, authoritative result, task completed, nothing leaked". Freeing the slot sooner is a separate, deliberate, configurable decision. There is no in-band signal that distinguishes "genuinely stuck" from "legitimately slow" — I checked, the stream is identical in both cases — so whatever number is chosen will cut off background work that exceeds it. The point is that exceeding it must not produce a false failure.

If you take this direction, please also treat these as must-fix:

- **Background tool accounting and cleanup must share one lifecycle.** The daemon's generic watchdog must not race the "kill only the background process" action and end up cancelling the whole Cursor run anyway.
- **Track each background process and validate ownership before signalling.** Do not trust a bare PID from the payload unconditionally — check process-group ownership, handle a PID that has already exited or been reused, and define the Windows-equivalent behaviour.
- **Regression coverage for the real shapes**: a background job that outlives the idle budget but stays under the tool budget completes normally; budget exhaustion cleans up only the background process and still yields Cursor's terminal result; cancellation, normal completion, and multiple background processes all leave nothing behind.

---

### Test coverage note

Test B (`cursorFakeWorkThenResult`) and Test C (`cursorFakePulses`) both establish "legitimate workflows are unaffected" by having thinking / step_finish events continue to arrive after the background shell. The real CLI does not emit anything during that window. These tests encode a stream shape that does not occur, and they will make it harder to correct this behaviour later.

The case that is missing is the important one: a background job that legitimately outlives the guard window and then produces a valid terminal result.

---

### Smaller notes (non-blocking)

- `cursorTerminalResultObserved(g)` is a one-line wrapper around `g.Stop() == ""` carrying a 12-line comment, and the name ("Observed") suggests it records something when it actually disarms the guard. Inlining `guard.Stop() == ""` at the call site would read more clearly.
- Several new comments in `cursor.go` argue with review feedback on #8042 (e.g. "that is not #8042's rule returning through another door"). That reasoning belongs in the PR description; the surrounding file does not do this, and the added comment density is well above the rest of `cursor.go`.
- `cursorSemanticProgressEvent` excludes `system` as "(startup/control)", but `system` also carries `task_notification`, which is a real lifecycle signal. The stated reason does not hold, even though it arrives too late to help in the current shape.
- The branch is 29 commits behind `main`. Not a blocker, worth rebasing before the next round.

CI is green and `go vet ./pkg/agent` is clean — the problems above are design-level, not build-level.

Happy to look at the next revision.

Bohan-J (changes requested):
## Request changes (round 2, at `96915a2`)

The redesign is the right one. You removed the Cursor-private timeout, kept background work counted as in flight, gave the budget back to the daemon's existing tool watchdog, and made cleanup target the specific process. That is exactly the direction, and several details are better than what I asked for.

I still can't approve it, because I verified on macOS that the ownership capture can never succeed against the real Cursor CLI, and that one configuration now goes from bounded to unbounded.

All findings below were verified locally against `cursor-agent 2026.08.25-3e8eec8` on macOS/arm64.

---

### What's good

- **`Session.ToolActivity` is a genuine improvement over the message-based count.** Native accounting no longer depends on the best-effort transcript, so a dropped message can't corrupt the budget decision. `TestExecuteAndDrain_BackgroundNativeCountSurvivesDroppedTranscript` covers it well.
- **Fail-closed ownership discipline.** An unverified PID is never signalled, `Alive`/`Terminate`/`Close` are all nil-safe, and capture checks ancestry, leader identity and process start time with a post-capture revalidation. This is careful work.
- **Linux's `pidfd_open` + `PIDFD_SIGNAL_PROCESS_GROUP` is the right primitive** — it retains the kernel group identity and has no same-session restriction.
- Reusing the existing watchdog and leaving the global defaults and zero-disable semantics alone is what the previous round asked for.

---

### Blocking 1 — on macOS, capture can never succeed against the real CLI

The Darwin implementation anchors the group by starting a `/bin/cat` child with `SysProcAttr{Setpgid: true, Pgid: pgid}`, i.e. it **joins** the shell's process group. `setpgid` returns `EPERM` across sessions.

I drove the real `cursor-agent` and called this PR's own `captureCursorBackgroundProcess` against the PID it reports:

```
cursor-agent : pid=71689 pgid=71689
bg shell     : pid=72113 pgid=72113   descendant=true
SESSION      : cursor-agent sid=71664   background shell sid=72113   same=false
CAPTURE FAILED: retain Cursor shell process group: fork/exec /bin/cat: operation not permitted
```

**Cursor's background shell calls `setsid()`** — it is its own session leader (`sid == pgid == pid`). I separately confirmed that the same `/bin/cat` anchor joins a *same-session* group without error, so this is specifically the cross-session restriction and not a sandbox or exec problem.

Your own Darwin unit test already encodes this: `TestCaptureCursorBackgroundRejectsDetachedSession` asserts that a `Setsid: true` child is rejected. That test is correct about the mechanism — but the shell it describes is the production case, not an edge case. On macOS the result is that capture always fails, nothing is cleaned up, nothing is released early, and the background process is still leaked.

**Suggested fix:** don't require a joinable group. Since the target is necessarily a session leader, re-verify identity by `(pid, start time)` at signal time and signal the enumerated group members individually by that identity. That preserves the fail-closed and PID-reuse-safe properties without needing the same session. You lose the durable reference after leader exit — but that is macOS's actual ceiling, and stating it is better than a path that silently never engages.

---

### Blocking 2 — `MULTICA_AGENT_TOOL_WATCHDOG=0` becomes unbounded (regression)

When capture fails, the background tool stays in flight permanently. In `runIdleWatchdog`:

```go
if toolInFlight {
    if toolWindow <= 0 { continue }   // "never force-stop during a tool call"
    threshold = toolWindow
}
```

so with the tool budget disabled, no budget can ever apply. A daemon-level test with an uncapturable background tool and `AgentToolWatchdog = 0`:

```
status="timeout" err=<nil> elapsed=2s interrupts=0
```

`interrupts=0` — the watchdog never fired at all. The run only ended because the caller's context expired, and production `AgentTimeout` defaults to `0`, so no such bound exists there.

On `main` the same run *is* bounded: the Cursor adapter emits the tool result at `tool_call completed`, the count returns to zero, and the idle budget applies. So for deployments that set the tool watchdog to `0` — a documented, supported value — this PR turns #7833 from "holds a slot for up to 2 h" into "holds a slot indefinitely". Combined with Blocking 1, that is every Cursor run that launches a background shell on macOS, not a corner case.

**Suggested fix:** a background tool whose ownership was never captured should not count as in flight forever. If it is still uncaptured past the idle budget, fall back to the idle budget rather than leaving the run unbounded.

---

### Blocking 3 — a new test in this PR fails deterministically on macOS

`TestCursorBackgroundLifecycle/finish` fails 5 out of 5 runs here:

```
--- FAIL: TestCursorBackgroundLifecycle/finish
    cursor_liveness_test.go:287: background process 74671 survived
WARN cannot own Cursor background shell; tool watchdog will use normal cancellation
     error="cursor background process is not an owned descendant"
```

That is the normal-completion cleanup path. Capture is rejected in the post-capture revalidation block (I instrumented the first ancestry check and it is not the one firing) because the `cursor-agent` root is on its way out at that moment, so the handle is lost and cleanup never runs.

This went unnoticed because CI runs the backend Go tests on `ubuntu-latest` (plus one `windows-latest` job) and **has no macOS Go test job**. Multica's desktop app and local daemon both run on macOS, so please add macOS backend coverage — at minimum for these new process-ownership tests. Otherwise this class of bug can only surface on user machines.

---

### Non-blocking

- `server/internal/daemon/daemon.go:9090` still carries an internal marker comment: `// ponytail: fixed shortest-budget cadence; use an adaptive timer if subsecond production budgets are needed.` Please drop it.
- In a full-package run, `TestKiroBackendPreservesCompletionOnRealGPT56SolFrames` timed out (`kiro initialize failed: context deadline exceeded`); it passes in isolation and passes on `main`. The new process-spawning tests likely add enough contention to matter — worth a look.
- Worth calling out in the PR description: a background shell is now killed when the run finishes normally, where previously it survived. I think that is the right behaviour, but it is a user-visible change.
- For reference, the daemon test failures I see locally (`TestEnsureDaemonID_*`, `TestPrepareReasonixTaskStateHome`, `TestPrepareDshTaskSessionRoot`, `TestLoadConfig_BackendOverrides_BackwardCompat_NoConfigFile`) reproduce identically on `main` and are unrelated to this PR.

---

### Summary

The design question is settled and the answer in this revision is right. What remains is bounded engineering: a macOS ownership primitive that doesn't depend on joining the group, closing the disabled-tool-budget hole, and CI coverage on macOS so both are actually verified. Happy to re-review as soon as those land.

Bohan-J:
I pushed three small maintainer-side commits to this branch (`79caaa4` → `0df812f`) rather than sending you round-4 review notes for two items that were faster to just do. Nothing in the ownership or watchdog behaviour changed.

**`chore(ci)` — dropped the PR-specific reproduction steps.** The two steps gated on `github.event.pull_request.number == 8050` did a genuinely good job of proving the new regression tests were RED on earlier heads, but they can't survive the merge: on `main` the condition can never hold again, and both pin commit SHAs that only exist on this branch. The permanent coverage is untouched — the lifecycle job still runs on ubuntu/macOS/Windows, and macOS still repeats the finalization test and checks the cgo-free path.

**`docs(daemon)` — wrote down what a zero tool budget now means.** Keeping a background shell in flight for the launched process's whole lifetime is exactly what puts a long background job on the tool budget instead of the shorter idle one, and I'm not asking you to change it. But it also changes what `MULTICA_AGENT_TOOL_WATCHDOG=0` does in practice: such a run is now bounded only by `MULTICA_AGENT_TIMEOUT`, which is `0` by default. That follows from the invariant rather than contradicting it — a foreground tool that never returns already behaves this way — but an operator reading "never force-stop during a tool call" would not predict it, so it's now stated where the knob is defined.

**`test(daemon)` — pinned the mirror case.** `TestCursorBackgroundUncapturedDisabledToolWatchdog` covers capture *failure* and releases the tool so the idle budget still applies. The capture-*success* case was untested and looks like an inconsistency next to it, which invites a later "fix" that quietly changes what the knob promises. Now both are pinned with a comment saying why they differ.

---

On the round-3 changes themselves: the Darwin rewrite works against the real CLI. I ran `cursor-agent 2026.08.25-3e8eec8` on macOS/arm64, had it launch a background `sleep 400`, and called your `captureCursorBackgroundProcess` on the PID it reports:

```
bg shell  pid=28452 pgid=28452 sid=28452   (same session as cursor-agent = false)
CAPTURE OK
Alive() before terminate = true
Terminate() err=<nil> took=11ms
Alive() after terminate  = false
cursor-agent emitted its terminal result, exited 0
total 20.7s instead of 400s, no stray processes
```

The shell is still cross-session, so the identity-based approach is doing real work that the group-join approach could not. `TestCursorBackgroundLifecycle` now passes here (it failed 5/5 before), the full `./pkg/agent` package is green, and `./internal/daemon` shows only failures that reproduce identically on `main`.

Two smaller things I checked and decided were fine, so no action needed: the capture probe's use of `SIGKILL` against a PID above `PID_MAX` is correct — I tried signal `0` and it returns `EINVAL`, so your choice is the working one. And the identity scan cost is negligible: a full process-table read is ~176 µs with 756 processes here, against a 1 Hz reap.

Nice work on this one — the redesign landed well.

Bohan-J:
Pushed two more maintainer-side commits (`0df812f` → `3962fac`) covering the two blocking findings from the last review round, plus the stale references in the PR description.

## `fix(daemon)` — an observed terminal result now outranks the idle watchdog

This was a real defect: a run that had already succeeded could still be reported as a hang.

On reading its authoritative `result`, `cursor.go` called `background.Close()` synchronously, before `Session.Result` reached the daemon. That call takes the tracker lock, so it can queue behind a tool-watchdog `Interrupt()` already in progress — and `Interrupt()` is allowed to return `false` without moving native accounting or its timestamp. The watchdog's revalidation then saw an unchanged count, an unchanged timestamp and an empty queue, cancelled the run, and `executeAndDrain` re-tagged the completed result as `idle_watchdog`.

The missing piece was a lifecycle boundary the daemon could see. *Cleanup succeeded* and *the outcome is decided* are different facts, and only the second one should silence a liveness policy:

- `Session.TerminalObserved` carries it (nil for backends that offer no such boundary).
- The Cursor backend publishes it via a lock-free flag **before** any cleanup that can block or fail.
- The watchdog checks it both before interrupting and immediately before force-stopping.
- The re-tag at `Session.Result` is gated on it too, so even a watchdog that wins the last instant cannot rewrite a decided outcome.

Two regression tests drive the exact interleaving with channel handshakes rather than sleeps — the cleanup callback itself publishes the terminal observation, so the terminal result provably lands while cleanup is in progress and returns `false`:

- `TestIdleWatchdogYieldsToTerminalObservedDuringCleanup` (`synctest`, no wall-clock) — asserts the interrupt boundary was actually reached, then that nothing fires, and that it stays decided across further budgets.
- `TestExecuteAndDrainKeepsTerminalResultObservedDuringCleanup` — covers the re-tagging half end to end.

Both fail on the parent commit, with `fired=true` and `Status:idle_watchdog Output:Cursor terminal result` respectively.

## `docs` — the operator contract is now the same in all four locales

The English page had gained three contracts that `ja`/`ko`/`zh` still described with the pre-change one-liner: that a Cursor background shell stays in flight until its owned processes exit, that normal finalization stops those processes, and what the per-platform fail-closed limits are. Those decide whether an operator sets an overall timeout, sets a tool budget, and expects a background server to survive a run, so they can't ship in one locale only.

While syncing I also made one thing explicit in all four, including English: with the tool budget at `0`, claimed background work keeps the run in flight for as long as it lives, so the run is bounded only by `MULTICA_AGENT_TIMEOUT` — which is itself `0` by default. That follows from the invariant rather than contradicting it, but an operator reading "never force-stops a tool call" would not predict it.

## PR description

Updated the current-head line and dropped the reference to the reviewed-head reproduction steps, which no longer exist. Their results are still recorded under Verification.

## Verification

- `gofmt` and `go vet ./pkg/agent ./internal/daemon` clean.
- `-race` on both focused CI selections, plus the two new tests: pass.
- macOS steps: `TestCursorBackgroundLifecycle/finish` ×5 and the `CGO_ENABLED=0` identity tests pass.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passes.
- Full `./pkg/agent` package green. One earlier full-package run showed six `TestKimi*` fixtures failing on 5s timeouts; they pass in isolation, pass on a re-run, and the same package is green on the unmodified parent commit, so that was machine contention rather than a change in behaviour.
- `./internal/daemon` shows only the seven environment-dependent failures that reproduce identically on `main`.

Bohan-J:
Pushed three more commits (`3962fac` → `fe8982f`) for the two blocking findings from the last round.

## `fix(daemon)` — a decided outcome now survives the drain-timeout classifier

The previous round added the terminal boundary but checked it in only two of the three places that can classify a run. The watchdog reads `terminalObserved` and *then* writes `fired` and cancels, so Cursor can publish its terminal result in between. The cancellation that follows does not surface on the `Session.Result` arm — the backend is still finalizing and cannot deliver `Result` yet — it surfaces on `drainCtx.Done()`, which classified purely on the `fired` flag. A run that had already succeeded came back as `idle_watchdog` with an emptied output.

That branch now checks the boundary too and waits, bounded, for the real result. Falling through on expiry keeps the previous behaviour, so the backstop cannot hang a run.

The earlier regression test could not reach this path: there the watchdog returned at its final gate, so `fired` was never set and neither the rewrite guard nor this classifier ran. `TestExecuteAndDrainKeepsTerminalResultPublishedAfterTheFinalGate` makes `TerminalObserved` publish on the very read the final gate performs while still answering `false` to that reader — the losing race exactly — and asserts the watchdog really did cancel before checking the outcome. On the parent commit it fails with:

```
a decided outcome was reclassified after the final gate:
{Status:idle_watchdog Output: Error:agent produced no new messages for 50ms ...}
```

## `fix(agent)` — the whole of `Close()` is now bounded, not just each process

Each `Terminate()` was bounded, but the number of background shells a run may launch is not, so finalization as a whole was not. That only matters after a terminal result — precisely when the watchdog has deliberately stopped supervising the run and `MULTICA_AGENT_TIMEOUT` is `0` by default, leaving nothing else to bound it.

`Close()` now has one budget for the whole pass. Work still unconfirmed when it expires takes the existing unconfirmed-at-close path, so its launch result is still surfaced and its cleanup is still logged as unconfirmed rather than reported as successful. `Reap` and `Interrupt` keep their previous behaviour.

`TestCursorBackgroundCloseIsBoundedAcrossTools` drives 40 tools whose termination never confirms. On the parent commit:

```
Close took 8.075660875s, want under 2.4s
```

## `docs(ja)`

Applied the wording suggestions: `最終結果` as used elsewhere in the docs, both signalling clauses now say シグナル送信 explicitly, and プロセス捕捉時 / the background process rather than "the launch".

## Verification

- `gofmt`, `go vet ./pkg/agent ./internal/daemon` clean.
- Both focused CI selections under `-race`: pass. The terminal-boundary tests ×50: pass.
- macOS `TestCursorBackgroundLifecycle/finish` ×5 and the `CGO_ENABLED=0` identity tests: pass.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1`: pass.
- `./internal/daemon`: only the seven environment-dependent failures that reproduce identically on `main`.
- `./pkg/agent`: full-package runs on this machine intermittently fail a *different* random set of provider fixtures each time, always at exactly 5.00s — Codex, Opencode, Hermes, Kimi, Cursor, Pi, Deveco have all appeared. I checked this rather than assuming: they pass in isolation, and a full-package run on the unmodified parent commit under the same load fails its own unrelated set the same way. It is local contention, not a change in behaviour. The focused selections above are the ones that actually cover this PR and they are stable.

Bohan-J:
Pushed the two blocking findings from the last round, plus a rebase onto `main` (the branch was `CONFLICTING` and 62 commits behind; the only conflict was `config.go`, where `main` added `CodexTurnInterruptTimeout` and realigned the struct — resolved in favour of `main`'s layout with just the `AgentToolWatchdog` comment change reapplied). Head is now `3ac7f6f0b` and the PR is `MERGEABLE`.

## Must-fix 1 — the verdict is now linearized on the hand-off, not on a flag read

You were right that adding a third read site only moved the window. The fix stops reading-then-acting: once the watchdog has fired, the drain arm enters the hand-off and decides from **what arrives**. The result's delivery is the linearization point, and the backend contract — publish the observation *before* sending `Result` — is what makes the check after delivery reliable rather than lucky. The `Result` arm already read in that order; this brings the drain arm in line. That contract is now written into `Session.TerminalObserved` rather than left implicit.

An expired budget still linearizes to the liveness verdict, so the branch stays bounded whatever a backend does.

On your test point: you were right that the old test proved nothing about which arm ran, because its backend sent `Result` immediately on cancellation. The new one withholds `Result` until the daemon has entered the hand-off, using the hand-off's own log record as both the gate and the evidence that the branch executed. On the previous head it fails with `the post-force-stop hand-off never ran; the drain arm was not exercised`.

**One thing I got wrong on the first attempt, and what it changed.** Entering the hand-off unconditionally taxed exactly the case the watchdog exists for: three existing idle-watchdog latency tests went from ~50 ms to the full 30 s budget, because a wedged backend has no result to hand over. `TerminalObserved == nil` is the signal for that and is now kept distinguishable instead of collapsed into a false-returning shim at the point of use, so backends without a boundary keep their previous behaviour exactly. Cursor always closes `Result`, so even a wedged Cursor run ends the wait through the closed channel rather than the budget. There is a new test pinning that fast path so the hand-off cannot quietly become a tax on every force stop again.

## Must-fix 2 — the bound now covers the lock wait

Correct diagnosis: `Close()` took its deadline before the lock, but nothing shortened the wait *for* the lock, and a concurrent `Interrupt()` still scanned an unbounded number of tools with no deadline of its own. Every pass that holds the lock is now bounded by the same budget. If waiting for the lock consumed it, the closing pass takes a fresh one rather than pushing every tool onto the unconfirmed path without trying.

`terminalResultHandoffBudget` is now derived from that ceiling rather than picked: a concurrent cleanup pass we may wait behind (10 s), the closing pass itself (10 s), the reaper tick (1 s) and the process `WaitDelay` (0.5 s) — about 21.5 s, with 30 s leaving margin. The derivation is in the comment.

The new test puts an `Interrupt()` pass on the lock first, as you described: ~1.1 s on the previous head against a 100 ms budget.

Semantics after a budget expiry are unchanged, as you approved: the original launch result is returned, cleanup is logged as unconfirmed, and nothing is reported as successful cleanup.

## Verification

- `gofmt`, `go vet ./pkg/agent ./internal/daemon`, `go build ./...` clean.
- Both focused CI selections under `-race`, plus the full `TestExecuteAndDrain*` set: pass.
- Terminal-boundary tests `-race -count=50`; close-bound tests `-race -count=20`: pass.
- macOS `TestCursorBackgroundLifecycle/finish` ×5 and the `CGO_ENABLED=0` identity tests: pass.
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1`: pass.
- Full `./pkg/agent`: green. `./internal/daemon`: only the seven environment-dependent failures that reproduce identically on `main`.
- CI will run fresh on the rebased head.

Bohan-J:
Fixed the test-ordering must-fix and the budget-derivation wording. Head is now `fe5d22eb4`.

**The gate now proves the lock is held.** You were right that closing a channel just before calling `Interrupt()` only says the goroutine was about to try, and that `Close()` winning the lock would let an unbounded `Interrupt()` pass the test. The gate is now closed from inside the first process's `terminate()`, which `finish()` reaches while holding `mu` for the whole pass, so reaching it is proof the pass owns the lock.

Worth recording what I measured before changing it, because it affects how much this was worth: with the old gate the test still failed **110/110** against an unbounded `Interrupt()` — 50 runs at default parallelism, then 30 each at `GOMAXPROCS=1` and `GOMAXPROCS=2` — because `close()` does not yield the P, so the interrupting goroutine reliably reached the lock first. So the false green was structural rather than observed. The change is still worth making: a regression test should not depend on a scheduler property to exercise the thing it guards. It is now deterministic by construction — 20/20 RED against an unbounded `Interrupt()` at `GOMAXPROCS` 1, 2 and 8.

**Budget derivation corrected as you described.** Closing the reaper's stop channel wakes it immediately, so its tick was not a real term; what deserves the margin is an already-started termination overshooting its pass's deadline by that termination's own bound. The derivation now reads: two bounded passes (10s each), one in-flight termination overshoot per pass (~1s each), plus the process `WaitDelay` (0.5s) — about 22.5s, with 30s as margin. The constant is unchanged, and the comment now says explicitly that the reaper tick is deliberately not a term.

**Verification:** `gofmt`, `go vet` clean; both focused CI selections and the full `TestExecuteAndDrain*` set under `-race`; close-bound tests `-race -count=20`; terminal-boundary tests `-race -count=50`; macOS `finish` ×5 and the `CGO_ENABLED=0` identity tests; `ByteIdentical`. Full `./pkg/agent` green; `./internal/daemon` shows only the seven environment-dependent failures that reproduce identically on `main`.

Bohan-J (approved):
Approving. Thank you for sticking with this one — it went through a redesign and four review rounds, and the result is materially better than what any single round would have produced.

What you shipped, from where it started:

- The first version failed a run as soon as Cursor reported a background shell. That would have reported successful work as a failure, and you rebuilt it around the opposite idea: a background shell is work still in flight, so it belongs on the daemon's existing tool budget rather than a new provider-private timer. That is the right shape, and it reuses a mechanism instead of adding a parallel one.
- When I showed that the macOS ownership capture could never succeed against the real CLI — Cursor's background shell is its own session leader, so the group-join approach always returns `EPERM` — you replaced it with kernel birth identities and `(pid, pidversion)` signalling instead of arguing the edge case away. I verified the replacement end to end against `cursor-agent 2026.08.25-3e8eec8`: capture succeeds, termination lands in 11 ms, Cursor emits its authoritative result and exits 0, and nothing is left behind.
- The fail-closed discipline held the whole way through. An unverified PID is never signalled, unconfirmed cleanup is logged as unconfirmed rather than reported as success, and unsupported kernels degrade to ordinary cancellation. That is the part that makes this safe to run on other people's machines.
- You also added CI coverage on all three platforms, which is what turned "works here" into something the project can keep.

Two things I did on the branch rather than sending you another round: removed the PR-scoped CI reproduction steps, which could not survive the merge, and closed the terminal-result / watchdog race plus the cleanup bound that the last two rounds surfaced. All of it is described in the comments above.

Merging now. Thanks again — this was a genuinely hard problem in someone else's process tree, and you took the hard road on it every time.

PR #8057, opened 2026-09-05
MUL-7073: fix(usage): distinguish unknown model costs from zero

mameikagou:
## Problem

The workspace leaderboard currently renders an unpriced usage row as `$0.00`. That makes two materially different states look identical:

- a model with an explicit zero-dollar rate;
- a model whose price Multica cannot determine.

This was observed in a real workspace rather than a synthetic edge case:

| Runtime-reported model | Tokens shown | Previous cost | What the value actually meant |
| --- | ---: | ---: | --- |
| `antigravity/gemini-3.8-flash-high` | 263.1M | `$0.00` | No matching rate and no provider-reported cost |
| `zhipu/glm-4.7-flash` | 25.7M | `$0.00` | Explicitly free rate |
| `antigravity/claude-opus-4-6-thinking` | 495.5K | `$0.00` | Same priced Claude Opus 4.6 SKU, but the routing CLI appended a mode suffix |

The first and third rows looked free even though Multica only lacked enough information to calculate them. At this volume, `$0.00` is a false financial claim rather than a harmless rounding artifact.

## What changed

- Track whether each per-agent total contains tokens that have neither provider-reported cost nor a matching local rate.
- Render a fully unknown cost as `—` instead of `$0.00`.
- Render a mixed total as `$known+`, making the displayed number an explicit lower bound.
- Keep `$0.00` for models that really have an explicit zero rate.
- Normalize the exact Claude `-thinking` / `-non-thinking` routing suffixes to the underlying Claude pricing SKU.
- Add the explanatory tooltip in all four shipped locales.

Provider-reported cost remains authoritative. This does not guess rates for private or subscription-only models, change stored usage, or change the API.

## Tests

- `pnpm --filter @multica/views exec vitest run dashboard/utils.test.ts runtimes/utils.test.ts dashboard/components/leaderboard.test.tsx` — 143 passed
- `pnpm --filter @multica/views exec vitest run locales/parity.test.ts` — 160 passed
- `pnpm --filter @multica/views typecheck` — passed
- `pnpm --filter @multica/views test` — 4,897 passed
- ESLint on the six changed TypeScript files — 0 errors

### DIFF
--- a/packages/views/dashboard/components/dashboard-page.tsx
+++ b/packages/views/dashboard/components/dashboard-page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useMemo, useState } from "react";
-import { BarChart3, RefreshCw } from "lucide-react";
+import { AlertCircle, BarChart3, RefreshCw } from "lucide-react";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { Button } from "@multica/ui/components/ui/button";
 import { Skeleton } from "@multica/ui/components/ui/skeleton";
@@ -29,10 +29,12 @@ import { useViewingTimezone } from "../../common/use-viewing-timezone";
 import { PAGE_GUTTER } from "../../layout/page-header";
 import { CollectionPageHeader } from "../../layout/collection-page";
 import { KpiCard } from "../../runtimes/components/shared";
+import { CustomPricingDialog } from "../../runtimes/components/custom-pricing-dialog";
 import { useNavigation } from "../../navigation";
 import {
   addDaysIso,
   aggregateByWeek,
+  collectUnmappedModels,
   formatTokens,
   todayIso,
 } from "../../runtimes/utils";
@@ -56,6 +58,8 @@ import {
   computeFailureTotals,
   isSyntheticAgentRow,
   mergeAgentDashboardRows,
+  summarizeCost,
+  type CostEstimate,
 } from "../utils";
 import {
   ALL_PROJECTS,
@@ -173,7 +177,7 @@ export function DashboardPage() {
 
   // The user can save model prices from the runtimes page; re-render when
   // they do so the dashboard reflects the new rates.
-  useCustomPricingStore((s) => s.pricings);
+  const pricings = useCustomPricingStore((s) => s.pricings);
 
   const { data: projects = [] } = useQuery(projectListOptions(wsId));
   const agentsQuery = useQuery(agentListOptions(wsId));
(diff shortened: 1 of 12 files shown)


multica-eve (changes requested):
This change addresses a real trust problem: when usage cannot be priced, showing it as free can cause workspace administrators to make incorrect budget decisions. Distinguishing unknown, partial, and genuinely free costs is the right direction, but the current version still gives contradictory cost information on the same page and can hide high-usage unknown-cost agents in the ranking.

## Request changes

1. **[P1] Propagate cost completeness to every cost surface.** `packages/views/dashboard/utils.ts:161-172` discards completeness in the whole-window total, so `dashboard-page.tsx:310-317,539-542` still renders an exact `$0.00`. Meanwhile, `usage-trend-card.tsx:81-95` treats zero known cost as no data. With only unpriced usage, the leaderboard shows `—`, the KPI shows `$0.00`, and the trend says there is no data for the same usage.

2. **[P1] Define sorting semantics for incomplete costs.** `leaderboard.tsx:23-26,79-94,201-207` sorts, truncates, and scales rows by the known numeric lower bound. An agent with substantial unpriced usage is therefore ranked as zero-cost and can be hidden below the first ten rows; a partially priced row is also presented as if it had a valid position in a strict cost ordering. Please group incomplete rows separately or otherwise avoid claiming an ordering that the available data cannot support.

3. **[P2] Recompute derived costs when custom pricing changes.** `dashboard-page.tsx:174-176` subscribes to the pricing store, but the relevant memo dependencies at `dashboard-page.tsx:310-317,406-409` do not include the pricing data, so a re-render can reuse stale cost and `hasUnpricedUsage` values.

4. **[P2] Make the explanation accessible.** `leaderboard.tsx:255-263` puts the incomplete-cost explanation only in a native `title` on non-focusable text. That is not reliably available to keyboard or touch users. Please use a visible explanation or the project's accessible Tooltip pattern, ideally with the existing custom-pricing path as the next action.

5. **[P2] Do not broaden financial model normalization without evidence.** `runtimes/utils.ts:531-536` strips both `-thinking` and `-non-thinking`, while `runtimes/utils.test.ts:138-151` only demonstrates and tests the observed `-thinking` identifier. Please remove the unverified branch or add a real source and regression fixture for it.

I recommend carrying a shared structured result such as `{ knownCost, completeness: "exact" | "partial" | "unknown" }` through daily totals, KPIs, trends, and per-agent rows, then applying one display policy everywhere. The focused tests, typecheck, diff check, and current CI pass, but they do not cover the cross-surface semantics and ranking behavior above.

mameikagou:
Addressed all five requested changes in c9846eb53:

1. Dashboard cost now carries a shared `{ knownCost, completeness }` contract. The KPI renders exact zero as `$0.00`, partial cost as the known lower bound with `+`, and fully unknown cost as `—`. Daily/weekly trends use token presence rather than a numeric zero to decide whether data exists, and show the same partial/unknown explanation.
2. Cost sorting no longer treats incomplete rows as strict dollar rankings. Incomplete rows form a separate first cohort ordered by complete token usage so high-usage unknown-cost agents cannot disappear below Top 10; fully priced rows follow by cost. Incomplete rows are not drawn against the priced dollar bar scale, and the rule is stated visibly.
3. Every price-derived memo (KPI, daily cost, weekly cost, per-agent aggregation, completeness, and unmapped-model list) is tied to the subscribed custom-pricing snapshot, so saving an override updates the page immediately.
4. The explanation is now visible instead of title-only, including KPI/trend/ranking context and a keyboard/touch-accessible `Set custom pricing` action that opens the existing pricing dialog. The existing title remains only as supplemental hover text.
5. Removed the unsupported `-non-thinking` normalization. Only the observed and tested `-thinking` suffix is stripped.

Validation completed:
- `packages/views` typecheck
- targeted ESLint on every changed TS/TSX source file
- 167 focused dashboard/pricing tests passed
- locale JSON parse checks and `git diff --check`

This makes unknown, partial, and genuinely free cost states consistent across every dashboard surface without inventing a financial ordering from incomplete data.

PR #8059, opened 2026-09-05
MUL-7074 fix(channels): derive chat titles from the current user instruction

### LINKED ISSUE #8058: [Bug]: Channel Chat titles can use enriched context instead of the current input
### Deployment type

Official App / self-host

The issue is in the shared backend, so it is not specific to a deployment type. The reproduction below was verified in a local source checkout with PostgreSQL; the deployed Official App was not separately tested.

### Area

Chat integrations — shared channel engine: Feishu / Lark, Slack, DingTalk, Telegram, and WeCom

### What happened?

A channel-created Chat can get its first title from context attached to the message instead of the member's current input. The backend uses the enriched, agent-visible body to derive the fallback title and as input to Router-triggered automatic title generation. That body can begin with a quoted message or recent group history.

For example, a member quotes an older discussion and asks `Compare current alternatives`. The fallback title can start with `<quotedmessage messageid="par…` or `<recentcontext count="1"`, and an automatically generated title can describe the older discussion. A media opening with a current media placeholder can similarly receive a context-derived title before its filename is available.

Expected behavior:

- Use the member's current input for the first title, while preserving the complete quote/history in the message supplied to the agent.
- Apply the same source rule to an ordinary first message and the first message supplied with `/new`.
- Exclude an executed `/clear` directive from the title source, while preserving literal command text in a remainder such as `/new /clear literal`.
- Keep the existing title-preservation rules when clearing or continuing a Chat. When media is pending and the known current input contains only a media placeholder, leave naming to the existing attachment-title path.

### Steps to reproduce

Concrete Feishu/Lark quote scenario:

1. In a conversation with a configured bot, reply to an older human message about an unrelated topic.
2. Address the bot with `/new Compare current alternatives` in that reply. The explicit quote is included in the new Chat's first message.
3. Inspect the new Chat title: it can come from the quoted-context prefix or historical topic instead of `Compare current alternatives`.
4. Inspect the stored first message: the quote should still be present after the title-source bug is fixed.

Recent-context variant: with recent-context enrichment enabled, make the first ordinary bot-addressed message on a group route that has not yet received a public user message. The title can come from the prepended recent-history block. This variant uses an ordinary first message because `/new` intentionally excludes automatic recent history.

These paths were exercised through server-side tests using the real Feishu enricher and session binder with PostgreSQL. No live channel UI smoke test was performed.

### Screenshots (optional)

Not available; the reproduction asserts persisted titles and message content.

- Verified against upstream `main` at `ca47495fcc3c4d2d0d6c7a421e45ddd0de98f0fd`.
- The existing input contract separates `CommandText` (current member input) from `Text`/`Body` (agent-visible content). The problem is that title derivation uses the latter.
- The implementation scope covers the shared engine and all five `/new` session adapters. Feishu provides the database-backed enrichment reproduction; Telegram's quote normalization is also covered. Shared adapter tests exercise all five binders with normalized inputs. This is not a claim that the same live-platform failure was reproduced separately on every channel.
- Existing titles are not being backfilled, and this issue does not request a change to automatic-title enablement. Missing-current-input fallback behavior is a separate limitation; this fix is not intended to resolve every media-only input case.

yyclaw:
## What does this PR do?

Fix first-title selection in the shared channel engine. Quoted messages and recent group history belong in the agent's context, but can currently determine the Chat title because the fallback and Router's LLM-title request use the enriched body. A quoted opening such as `Compare current alternatives` should be titled from that request while retaining the quote in the stored message.

The inbound contract already provides the current member input as `CommandText`. This PR uses it for title selection and forwards it through all five `/new` adapters: Feishu/Lark, Slack, DingTalk, Telegram, and WeCom. This is shared backend behavior for Official App and self-hosted deployments running the affected code.

## Related Issue

Closes #8058

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

| Surface | Change and purpose |
| --- | --- |
| `server/internal/integrations/channel/engine/title.go`, `session.go`, `router.go` | Select the current input for ordinary first-message and `/new` fallback titles, and pass the same source to Router-triggered LLM naming. Keep the enriched Body for persistence and agent input. |
| `server/internal/integrations/lark/feishu_resolvers.go`; `{slack,dingtalk,telegram}/resolvers.go`; `wecom/wecom_resolvers.go` | Forward each adapter's existing `CommandText` into `StartSessionInput`. Without this mapping, `/new` would fall back to Body whenever the current source was omitted. Where the two inputs are identical, the visible result is unchanged. DingTalk and WeCom's production changes are this one-line mapping, rather than separate platform-specific title logic. All paths share the `server/internal/integrations/` prefix. |
| `server/internal/integrations/slack/slash_control.go` | Native `/new` passes its body as the current input. Native `/clear` passes the directive plus body so the shared selector can consume exactly one directive and retain the first-turn title; its command body remains an ordinary message. |
| Regression tests in the engine, adapters, and handler | Verify title/body separation, command remainders, media filename initialization, native Slack controls, and LLM input/CAS publication. Add explicit stored/returned-title assertions for empty and nonempty `/new` sessions. |

Behavior boundaries:

- Consume only an executed `/clear`. `/new /clear literal` keeps the inner `/clear`; `/clear /issue literal` does not become an issue command.
- With pending media and only placeholder text in the known current source, allow the existing attachment path to initialize the title from the filename. Missing `CommandText` retains the existing Body fallback, so some media-only cases remain outside this fix.
- Existing/manual titles, message content, routing, task attribution, and deduplication keep their existing rules. There is no historical-title rewrite, schema/API/event-shape change, or deployment step.
- Automatic-title enablement is unchanged. This does not add LLM naming or title-event hooks to native Slack `/clear`.

## How to Test

1. Follow the linked issue's Feishu quote or ordinary-first-message recent-context reproduction. The title should use the current request, while the stored message keeps the context. Also check `/new /clear literal`, consumed `/clear`, and a media-placeholder opening.
2. With `DATABASE_URL` pointing to a migrated local test database, run from `server/`:

   ```bash
   ../scripts/go-test-with-agent-cli-guard.sh -- go test -race \
     ./internal/integrations/channel/engine ./internal/integrations/lark \
     ./internal/integrations/telegram ./internal/integrations/dingtalk \
     ./internal/integrations/slack ./internal/integrations/wecom -count=1

   ../scripts/go-test-with-agent-cli-guard.sh -- go test ./internal/handler \
     -run 'Test(ChannelChatTitle_UsesCurrentSourceForLLMAndPublishesAfterCAS|SlackNativeNewCommandCreatesMessageTaskAndDeliveryAtomically|SlackNativeClearCommandKeepsChatAndAdvancesContextAtomically|ChatTitle_DoesNotClobberManualRename|ChatTitle_SilentFallbackOnUpstreamError|ChatTitle_FallsBackWhenLLMDisabled|ChatTitle_FallsBackOnEmptyModelOutput|ChatTitle_AutoTitlesOnlyOnce)' \
     -count=1 -v

   go vet ./internal/integrations/channel/engine ./internal/integrations/lark \
     ./internal/integrations/telegram ./internal/integrations/dingtalk \
     ./internal/integrations/slack ./internal/integrations/wecom ./internal/handler
   ```

3. Local verification passed, with the following evidence boundaries:

   | Coverage | What it verifies |
   | --- | --- |
   | Real Feishu enrichment → binder → PostgreSQL | Quote/recent-context, `/new`, and consumed `/clear` titles; preserved message body |
   | Real Telegram normalization → captured binder inputs | Current input and quoted body remain separate for append and `/new` |
   | All five real binders → PostgreSQL, using normalized fixtures | `/new` forwards the current source and persists the full body; this does not simulate each platform's live transport |
   | Shared engine media and Router tests | With pending media and a known current placeholder, naming waits for metadata; the title lifecycle receives the intended current source |
   | Native Slack and handler tests | Title, route/context, task/delivery, and deduplication results; exact LLM input, CAS-before-event behavior, and existing manual-rename/error fallbacks |

   The six channel packages passed with `-race`; targeted tests, `go vet`, formatting, and diff checks passed. The two strengthened `/new` title tests were rerun after adding their assertions. Restoring the three old title-source expressions in an isolated copy made 16 scenarios fail on the intended assertions, while the missing-source control passed. The LLM test uses a local fake HTTP server. Live channel accounts, the deployed Official App, a real LLM, and the full monorepo suite were not exercised locally.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — title values are asserted in tests; no UI screenshots were captured
- [ ] I have updated relevant documentation to reflect my changes — no new configuration, commands, or documented workflow
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy** (`apps/web/features/landing/i18n/`) and **relevant docs** (`apps/docs/content/docs/`) — not applicable
- [ ] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (terminology, mixed-rule for `task` / `issue` / `skill`) — not applicable

## AI Disclosure

**Prompt / approach:**

Codex reviewed and refined the changes against the existing input contract and repository history. Independent read-only checks, targeted PostgreSQL tests, race checks, and an old-behavior comparison were used to verify the fixes and regression coverage.

## Screenshots (optional)

Not included; automated tests assert the stored title, preserved message content, and title-generation input.

### DIFF
--- a/server/internal/handler/channel_new_e2e_test.go
+++ b/server/internal/handler/channel_new_e2e_test.go
@@ -704,6 +704,11 @@ func TestSlackNativeClearCommandKeepsChatAndAdvancesContextAtomically(t *testing
 	if sessionID == "" || routeRevision != 1 || contextRevision != 2 || content != "/issue investigate native clear" || messageKind != "message" || !forceFresh || !boundaryPending || taskCount != 1 || deliveryCount != 1 || routeCount != 1 || !dedupProcessed {
 		t.Fatalf("native /clear message state: session=%s route=%d context=%d content=%q kind=%q fresh=%t pending=%t tasks=%d deliveries=%d routes=%d dedup=%t", sessionID, routeRevision, contextRevision, content, messageKind, forceFresh, boundaryPending, taskCount, deliveryCount, routeCount, dedupProcessed)
 	}
+	var title string
+	dbfx.QueryRow(t, `SELECT title FROM chat_session WHERE id = $1`, sessionID).Scan(&title)
+	if title != "/issue investigate native cle…" {
+		t.Fatalf("native /clear first-turn title = %q, want the literal command body", title)
+	}
 
 	if err := starter.ClearSlackDMContext(ctx, installation, util.MustParseUUID(testUserID), slackapi.SlashCommand{
 		ChannelID: "D-native-clear",
@@ -740,6 +745,10 @@ func TestSlackNativeClearCommandKeepsChatAndAdvancesContextAtomically(t *testing
 	}
 	if currentSessionID != sessionID || currentContextRevision != 3 || !currentBoundaryPending || currentTaskCount != 1 || currentMessageCount != 1 || currentRouteCount != 1 || !bareDedupProcessed {
 		t.Fatalf("bare native /clear state: session=%s/%s context=%d pending=%t tasks=%d messages=%d routes=%d dedup=%t", currentSessionID, sessionID, currentContextRevision, currentBoundaryPending, currentTaskCount, currentMessageCount, currentRouteCount, bareDedupProcessed)
+	}
+	dbfx.QueryRow(t, `SELECT title FROM chat_session WHERE id = $1`, sessionID).Scan(&title)
+	if title != "/issue investigate native cle…" {
+		t.Fatalf("bare native /clear changed the existing title to %q", title)
 	}
 }
 
--- a/server/internal/handler/channel_title_regression_test.go
+++ b/server/internal/handler/channel_title_regression_test.go
@@ -1,0 +1,76 @@
+package handler
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/multica-ai/multica/server/internal/events"
+	"github.com/multica-ai/multica/server/internal/testutil"
+	"github.com/multica-ai/multica/server/pkg/llm"
+	"github.com/multica-ai/multica/server/pkg/protocol"
+)
+
+func TestChannelChatTitle_UsesCurrentSourceForLLMAndPublishesAfterCAS(t *testing.T) {
+	requireDB(t)
+	fx := testutil.New(testPool, testWorkspaceID, testUserID)
+	const current = "Investigate the login redirect"
+	sessionID := parseUUID(fx.ChatSession(t, uuidToString(chatTitleTestAgentID(t)), testutil.Cols{"title": current}))
+	requests := make(chan string, 1)
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var payload struct {
+			Messages []struct {
+				Role    string `json:"role"`
+				Content string `json:"content"`
+			} `json:"messages"`
+		}
+		if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
+			t.Error(err)
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		for _, message := range payload.Messages {
+			if message.Role == "user" {
+				requests <- message.Content
+			}
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = io.WriteString(w, `{"choices":[{"message":{"role":"assistant","content":"Login redirect investigation"},"finish_reason":"stop"}]}`)
+	}))
+	defer upstream.Close()
+	h := *testHandler
+	h.Bus = events.New()
+	h.LLM = llm.New(llm.Config{APIKey: "test-key", BaseURL: upstream.URL})
+	updates := make(chan protocol.ChatSessionUpdatedPayload, 1)
+	h.Bus.Subscribe(protocol.EventChatSessionUpdated, func(event events.Event) {
+		payload, ok := event.Payload.(protocol.ChatSessionUpdatedPayload)
+		if ok && payload.ChatSessionID == uuidToString(sessionID) {
+			stored, err := h.Queries.GetChatSession(context.Background(), sessionID)
+			if err != nil || stored.Title != payload.Title {
+				t.Errorf("title event preceded committed CAS: stored=%q event=%q err=%v", stored.Title, payload.Title, err)
+			}
+			updates <- payload
+		}
+	})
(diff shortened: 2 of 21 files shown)


yyclaw:
Please review and merge #8061 before proceeding with this PR.

Bohan-J:
Thanks @yyclaw — merged.

This was a well-scoped fix, and a few things made it straightforward to accept:

- It extends an existing path rather than adding a parallel one. The inbound contract already separated the current instruction from the agent-visible body, so the fix is ~30 lines of production code, five of them one-line adapter mappings. The alternative — teaching the shared engine to strip each platform's enrichment markup — would have been the worse layering, and you avoided it.
- The evidence boundaries in the description held up. We reproduced the broken titles independently (`<quotedmessage messageid="om8…` for the quoted-reply case, `<recentcontext count="5"` for the recent-context variant), and reverting the three title-source expressions made six scenarios fail on their intended assertions — the regression tests genuinely guard the behavior instead of restating it.
- You were explicit about what you had *not* verified (live channel UI, media-only openings with a missing current source) instead of overclaiming. That is worth more to a reviewer than extra green checkmarks.

We pushed one comment-only commit to your branch (`faacb2d8`, no behavior change). It records three invariants the fix depends on that were true but implicit, each of which would have failed silently:

- `Router.Handle`'s `CommandText == "" → Text` fallback assigns the *already enriched* text. A future enriching adapter that leaves `CommandText` empty would reinstate this bug with every title test still green, so the invariant now sits on the fallback itself.
- `AppendInput`'s doc still described `CommandText` as `/issue`-parsing input only, and did not state that a `ForceFresh` caller must pass the directive together with its body — a bare `/clear` is consumed to an empty string and the chat silently keeps an empty title.
- The `missing current text` case is now marked as a recorded gap rather than a contract, with the correct future fix noted (return `""` and let the attachment path name the chat).

On merge order: you asked for #8061 to land first. We merged this one ahead of it because it stands on its own review, and we verified the two heads auto-merge cleanly — `router.go` is the only file both touch, and it merges without conflict — so this should cost you no rebase. #8061 is under separate review and is not blocked by this merge.

One non-blocking follow-up whenever you are next in this area: the native dedup fixture at `server/internal/integrations/channel/engine/session_test.go:529` still passes `Body: "start clean", CommandText: "/clear", ForceFresh: true`, which now contradicts the contract documented above. It does not correspond to any real caller, so it is a consistency nit rather than a defect.

PR #8061, opened 2026-09-05
fix(channels): preserve selected DingTalk input and control-command context

### LINKED ISSUE #8060: [Bug]: Channel input loses DingTalk context, emits XML quotes, and consumes /new twice
### Deployment type

Deployment-independent backend behavior. The affected scope is any deployment running the affected code with the relevant channel enabled. Reproduced in a local source checkout using callback fixtures and test doubles; the Official App version and live provider behavior have not been verified.

### Area

- Chat integration — DingTalk (community-maintained)
- Chat integration — Feishu / Lark
- Chat integration — Telegram (community-maintained)
- Shared channel command routing, also used by Slack and WeCom
- Web/desktop chat rendering of the resulting message body (production renderer unchanged)

### What happened?

Channel input should keep the explicitly selected reply readable, retain the current message and its media, and consume a current control command once. The following local reproductions identify different parts of that input path:

| Area and trigger | Behavior on the base revision | Expected behavior |
| --- | --- | --- |
| DingTalk callback containing a selected text, picture, rich-text, or bot-card snapshot | Selected content is omitted from the normalized message; the current instruction remains | Include a readable quote with a best-effort author label, and pass supported quoted images through the existing media pipeline |
| DingTalk current rich text with a structured text node, such as `{"text":{"content":"show details"}}` | Visible text can become `[Image unavailable]` | Keep the text and its order relative to current images |
| Lark parent-message enrichment; Telegram group message that explicitly mentions the bot and replies to a non-bot author | Existing reply context is emitted in a private `<quoted_message message_id=... sender=... type=...>` wrapper | Emit a standard Markdown blockquote with a readable author label while retaining literal body content |
| Any shared-router input whose adapter already consumed the first `/new`, leaving a body that begins with literal `/new` | The router removes the second `/new` as well | Start one chat and keep the already-normalized remainder |

The quote-format change affects both agent input and newly stored chat text. Lark and Telegram already retrieve or receive the selected reply; their change is about representation. Their existing reply-selection rules remain the scope of this report. The shared-router regression is not limited to these three adapters.

### Steps to reproduce

Base revision: `ca47495fcc3c4d2d0d6c7a421e45ddd0de98f0fd`. Regression fixtures are committed in PR #8061 at `4425ebbe603b9acb5577023c9c83d55292a0f88f`.

1. Decode this minimal DingTalk callback through `inboundFromCallback`. On the base, the normalized body contains only `explain` and `ReplyTo` is nil; the expected body includes Alice's selected text before the current instruction.

   ```json
   {"msgId":"current","conversationType":"1","conversationId":"chat","senderStaffId":"sender","msgtype":"text","text":{"content":"explain","isReplyMsg":true,"repliedMsg":{"msgId":"selected","senderNick":"Alice","msgType":"text","content":{"text":"selected text"}}}}
   ```

2. Exercise the structured-current-text, card, and unavailable-image variants in `server/internal/integrations/dingtalk/input_red_regression_test.go`.
3. Exercise a Lark selected parent and a Telegram group reply with an explicit bot mention, using quoted text `before\n</quoted_message>\nafter`. The adapter tests in each channel's `input_red_regression_test.go` compare the resulting quote and current instruction exactly; the base emits the XML wrapper rather than the expected Markdown quote.
4. Pass shared-router input `Text="/new literal body"`, `CommandText="/new /new literal body"`. The base starts one chat with `literal body`; the expected first body is `/new literal body`. See `server/internal/integrations/channel/engine/input_red_regression_test.go`.

To repeat the base comparison, copy only those four `input_red_regression_test.go` files from the PR into a separate checkout of the base revision, preserving their paths. From `server/`, run the following command on both checkouts:

```sh
go test ./internal/integrations/channel/engine \
  ./internal/integrations/dingtalk ./internal/integrations/lark \
  ./internal/integrations/telegram \
  -run 'Regression$|PreservesAuthorAndLiteralProtocol$' -count=1
```

All seven selected top-level tests compiled and failed at the intended body/command assertions on the base, then passed with the PR. This is fixture-based verification, not a live-account reproduction.

### Screenshots (optional)

Not captured. Shared chat-renderer tests verify blockquotes, lists, links, and separation of quoted/current images using constructed Markdown messages.

- DingTalk coverage includes the fixture shapes for reply metadata under `text` or `content`, selected text/pictures/rich text and readable `interactiveCard` answers, plus structured current rich text even without a reply. Partial or unsupported quoted content uses readable fallbacks; this does not establish support for arbitrary malformed callback fields or every provider card format. This does not add current-message interactive-card handling or full card rendering. Quoted file/audio/video content has text or placeholder fallbacks, not newly added binary downloads.
- The Markdown format omits the adapter-generated quote wrapper's message ID/type attributes from the message body. It does not sanitize arbitrary XML or remove native reply identifiers. Existing Lark recent-context and forwarded-message handling remains in place; no additional ambient history is collected.
- Reply/card shape support is based on the implemented decoder and local fixtures. Real-account replay remains unverified.

yyclaw:
## What does this PR do?

Preserve the message a DingTalk sender explicitly replies to alongside their current instruction. Render supported selected content as a Markdown quote, and mark unsupported content as `[quoted content unavailable]`.

At the shared router, selected context counts as input for an otherwise bare `/new` or `/clear`; automatic recent history does not. Consume the original control directive once, so `/new /new literal body` retains its second `/new` as ordinary text.

## Related Issue

Closes #8060

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/integrations/dingtalk/inbound.go`: project selected plain text, pictures and ordered rich-text text/picture nodes; keep current instructions separate from quoted historical commands. Preserve image occurrence indexes across quoted/current media and unavailable nodes. Malformed optional quote metadata cannot reject an otherwise valid current message.
- Bound the fallback deliberately: raw quoted text containing `||` becomes unavailable, without length, alphabet, version or trailer-field assumptions. This also hides legitimate quoted code/prose containing `||`; current input is unaffected. It is a conservative display policy, not a decoder or a complete classifier for arbitrary opaque formats. The [public sample](https://github.com/open-dingtalk/dingtalk-stream-sdk-go/issues/22) supplies a regression fixture, not a vendor envelope contract.
- Unknown message types and interactive cards become unavailable. Picture download references remain usable, but supplementary `content.text` produces an unavailable marker rather than an inferred caption; no `[Image]` prefix stripping remains. Rich text uses ordered string-text/picture nodes, without recursive aliases, JSON-encoded arrays or summary-to-image pairing. The [receive-message schema](https://open.dingtalk.com/document/orgapp/receive-message) does not establish those reply-summary semantics; [card parameters](https://open.dingtalk.com/document/orgapp/create-and-deliver-cards) are template-specific. File/video labels and available audio recognition remain bounded projections, without new binary downloads.
- `server/internal/integrations/channel/{message.go,engine/router.go}`: carry `HasSelectedContext` through the existing adapter-to-router path. Preserve selected text/media on bare controls, including unavailable placeholders; retain task preparation, transaction/retry and `SkipAgentRun` behavior.
- `server/internal/integrations/lark/` and `telegram/inbound.go`: propagate selected-context status while preserving the existing XML quote representation and selection conditions. There is no Lark/Telegram Markdown migration.
- Channel/integration tests and `packages/views/chat/components/chat-message-list.test.tsx` cover command boundaries, quote fallbacks, ordered media and stored-message rendering. No frontend production code, daemon code, migrations, public API or deployment configuration changes; no historical-body rewrite or new message-history lookup.

## How to Test

1. Exercise production callback conversion with the public opaque body fixture and synthetic callbacks: ambiguous short/extended `||` text must become unavailable, unknown kinds must not expose `content.text`, and picture supplementary text must not become a caption. Current input and neighboring image slots must survive. The new regression assertions fail against the previous production implementation and pass with this change.
2. From `server/`, run `go test -race ./internal/integrations/channel ./internal/integrations/channel/engine ./internal/integrations/dingtalk ./internal/integrations/lark ./internal/integrations/telegram ./internal/integrations/slack ./internal/integrations/wecom -count=1`. Passed all seven packages. Run `go vet` over the same package list; passed. The mapped control/quote regression selection also passed.
3. Run `pnpm --filter @multica/views test chat-message-list`. Passed 49 tests in two files, including quoted text/images and existing stored-message rendering.

## Checklist

No runtime/tool/tab or Chinese product-copy changes. Compatibility and fallback behavior are described above; no repository documentation files are changed.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Investigated the reported channel-input failures, traced adapter-to-router behavior, and revised the implementation against reviewer feedback. Used provider documentation to narrow supported projections, reproduced regressions against the previous implementation, and checked the final diff against targeted and package-level tests.

## Screenshots (optional)

No frontend production changes. Message rendering is covered by the automated tests above.


yyclaw:
Please review and merge #8059 before #8061 based on the functional dependency, not submission time. Neither PR is DingTalk-only: #8059 fixes shared title selection across Feishu/Lark, Slack, DingTalk, Telegram, and WeCom; #8061 covers DingTalk input parsing, Lark/Telegram Markdown quotes, and shared-router `/new` handling.

The title fix should land first so titles use the current instruction while preserving contextual message bodies. It fixes existing contextual-title issues and prevents #8061's newly supported DingTalk quotes from producing titles based on quoted history.

Bohan-J (changes requested):
Thanks for splitting this out — the problem is real and worth fixing. I reproduced all three symptoms on the baseline (`ca47495f`): DingTalk selected replies losing their content, structured quote markup reaching the agent instead of readable text, and `/new` being consumed twice. The corresponding regressions pass on this branch, and the direction — one selected reply, rendered as a Markdown quote, with control commands interpreted exactly once — is the right one.

I'm requesting changes on three points before merge, plus one design decision that needs a maintainer call.

## Must-fix 1: real multi-line DingTalk quotes may be passed through as if they were readable text

`server/internal/integrations/dingtalk/inbound.go:550` writes `replied.Content.Text` straight into the quote block for `msgType == "text"`.

Per the DingTalk Stream SDK maintainers, a multi-line `repliedMsg.content.text` can arrive as opaque content with a `||version||type||len` suffix, and there is no public algorithm or API to recover the plaintext: [open-dingtalk/dingtalk-stream-sdk-go#22](https://github.com/open-dingtalk/dingtalk-stream-sdk-go/issues/22#issuecomment-5053446899).

As written, that opaque payload is handed to the agent as if it were the member's own words — the opposite of what this PR is trying to achieve. The synthetic fixtures here can't catch it, because they only ever contain the plaintext form.

Please either recover the original text through a verifiable source (e.g. message history lookup), or detect the opaque form and degrade explicitly to something like "quoted content unavailable". Fabricating context is worse than admitting it's missing. A redacted real callback pinned as a fixture would settle this.

## Must-fix 2: a bare `/new` or `/clear` on a quoted message still drops the quote

The adapter correctly produces "`Text` is the quote block, `CommandText` is the directive", but the shared router decides emptiness from the directive remainder only and never looks at `Text`:

**`/new` + text-only quote** — `server/internal/integrations/channel/engine/router.go:216` sets `CommandText = control.Body` (empty). `Text` is preserved (the `Text == CommandText` guard at `:211` correctly declines to strip), but `:395` computes `persistMessage := msg.CommandText != "" || hasMedia` → false, and `:451` returns `OutcomeChatStarted` early. The new chat is created and the quote is silently discarded. (A quoted *image* survives this path, since `hasMedia` is true — it's the text-only case that's lost.)

**`/clear` + quote** — `:224` sets `bareFresh = control.Body == ""` → true, so `:465` calls `MarkPendingFresh` and returns `OutcomeFreshPending` without ever reaching `AppendMessage` at `:488`. Here both the quoted text *and* the quoted media are dropped.

Worth noting that `normalizeDingTalkRichTextControlLayout` (`inbound.go:996-999`) already works around exactly this for media-bearing rich-text `/clear` by reassigning `CommandText = msg.Text`. That confirms the sentinel is the problem — it just isn't applied on the other paths.

The fix belongs in the shared layer: base the "bare command" decision on the actual visible body and media after the directive is removed, not on `control.Body` alone.

`TestInboundFromCallback_QuotedControlStripsOnlyCurrentDirective` does cover `bare new` / `bare clear`, but it asserts on `msg.Text` at the adapter boundary and never crosses the router, so both drops are invisible to it. Please add router-level or end-to-end coverage for `/new` and `/clear` with both a text quote and an image quote.

## Must-fix 3: the undocumented parsing surface needs provider evidence

`inbound.go:723-943` infers `interactiveCard` structure from field-priority chains, node aliases and recursive rules, and `:605-715` (`renderDingTalkQuotedRichText` / `selectDingTalkRichTextLayout`) picks between two competing sources of truth for rich text. All of this is on by default, and the PR description states there's no real-account replay.

The tests are thorough, but they're written against the same assumptions as the implementation — they prove internal consistency, not that DingTalk actually emits these shapes. For each branch this PR claims to support, could you attach a redacted real payload or a link to the official contract? Otherwise I'd rather narrow the implementation to the shapes we've actually observed and degrade explicitly on everything else.

## Design decision needed: Lark/Telegram quote representation

This PR removes the quoted message ID/type from the outgoing body for Lark and Telegram in favour of the Markdown quote block. The existing `<quoted_message>` form was introduced deliberately in `5eba94ee2` to give the agent explicit context and command boundaries — it isn't an accidental format.

So this is a product call, not a cleanup: do we accept dropping the structured source metadata, or do we keep it and add the readable projection alongside? Historical messages aren't migrated either way. I'd suggest landing the DingTalk fix and the control-command boundary first, and deciding the Lark/Telegram migration separately.

## What's working well

- `channel.FormatQuotedMessage` gives one consistent escaping and multi-line quote boundary, and commands still read from a separate field — so a historical quote can't be reinterpreted as a current directive. That's the right split.
- The `/new` fix is appropriately minimal: it stops re-parsing a body the adapter already normalized, which correctly preserves the literal second segment in `/new /new literal body`.
- Quoted images reuse the existing occurrence-based `InlineIndex` with a placeholder offset instead of introducing a media side-channel.
- No `server/internal/daemon/` changes, no migrations, no public API or deployment config changes.

## Verification notes

- Baseline `ca47495f`: the four regression files fail on the expected context/format/command assertions — the problem is confirmed, not theoretical.
- This branch `3ad01cfe`: targeted `go test -race` passes on the five relevant packages; `go vet` passes on seven; `git diff --check` clean.
- Caveats on my side: a wider seven-package race run was blocked by an unmigrated local database (missing `route_revision`, `channel_chat_context_generation`), so I'm not counting it as passing, and I didn't run `vitest`. CI is green here, but no run exercises a real DingTalk callback.

Happy to re-review once 1 and 2 are addressed — 3 can be narrowed rather than fully evidenced if that's faster.

yyclaw:
@Bohan-J Addressed the three must-fix items and kept the existing Lark/Telegram quote representation.

Fix commit: [a279f0c18](https://github.com/multica-ai/multica/pull/8061/commits/a279f0c18754c0d3f6017c9dfa30deb8d4be3317).

1. **Opaque DingTalk quotes:** the long encoded envelope matching the public sample in [SDK issue #22](https://github.com/open-dingtalk/dingtalk-stream-sdk-go/issues/22) now becomes `[quoted content unavailable]`. Detection applies only to selected content; ordinary multiline text, literal JSON, and current input stay intact. The regression identifies the public body sample's provenance and explicitly distinguishes it from the synthetic callback envelope.
2. **Bare `/new` and `/clear`:** the shared router uses `HasSelectedContext` to distinguish an explicit selection from automatic history, together with actual media presence. Selected input is persisted and processed; bare commands and automatic recent context still do not create empty runs. Removed DingTalk's workaround that replaced `/clear`'s `CommandText` with an image marker. Router coverage includes text, image, XML, and unavailable quotes, plus retries and `SkipAgentRun`.
3. **Undocumented parsing:** took the narrowing option. Kept ordered string-text/picture nodes from the [receive-message documentation](https://open.dingtalk.com/document/orgapp/receive-message), and removed recursive card rendering, nested text aliases, recursive decoding of JSON-encoded arrays, and summary-to-image matching. Complex cards and unknown structures degrade explicitly while supported neighboring text and images survive. Malformed optional quote snapshots no longer reject an otherwise valid current message.
4. **Lark/Telegram:** restored the original XML quote and its `message_id`/`sender`/`type` metadata. The existing paths only gain the explicit-selection flag needed for item 2; a format migration can be considered separately.

The [card API documentation](https://open.dingtalk.com/document/orgapp/create-and-deliver-cards) makes `cardParamMap` template-dependent; the documentation checked does not define a universal body-recovery contract for complex reply snapshots. This narrowed scope supersedes the broader parsing and Lark/Telegram format-migration claims in the original PR description.

Validation: full race tests in the seven relevant Go packages, matching `go vet`, 49 chat-message-list frontend tests, and `git diff --check` passed. The full PR behavior/test inventory has also been reconciled. Synthetic boundary tests are not presented as evidence of an undocumented provider contract. Ready for another review of this narrowed scope.

Bohan-J (changes requested):
Thanks for the rework — this is a real improvement, and the central fix is now in the right place.

**Previous round's blocking issue is properly resolved.** The bare `/new` / `/clear` drop is fixed at the shared layer rather than patched per-adapter: `channel/message.go` adds `HasSelectedContext`, and `channel/engine/router.go:379-381` folds it into both `persistStartedMessage` and `bareFresh`. Distinguishing sender-*selected* context from automatic recent history is exactly the right distinction to draw there, and it means an otherwise-bare control command with a quote now becomes a real turn. Router-level coverage for text, image, unavailable-placeholder and `SkipAgentRun` is the coverage I asked for.

**The Lark/Telegram format migration is withdrawn**, which removes the open design question from last round — both now only propagate the signal (`lark/inbound_enricher.go` +2, `telegram/inbound.go`), keeping the existing `<quoted_message>` representation.

Three things still block, all narrower than last time.

## Must-fix 1: the opaque-envelope matcher is an unproven heuristic

`server/internal/integrations/dingtalk/inbound.go:640-661` classifies a body as opaque when it splits into exactly four `||` parts, the last three are entirely numeric, and the first is ≥64 characters drawn from the Base64 alphabet.

The function's own doc comment concedes the case: *"Even the published sample is not valid padded Base64, so decoding is not a reliable detector."* That's an accurate read of the evidence — which is precisely why the surrounding thresholds shouldn't be load-bearing. This is pattern-matching a single community-reported sample, not a vendor contract.

To be fair about which direction actually hurts: the matcher is tight enough that false positives (hiding legitimate text) look unlikely in practice. The problem is **false negatives**. A shorter body, a different version prefix, or a fourth numeric field all fall through and get handed to the agent as the member's own words — which is the exact failure #8060 reports. A heuristic that silently degrades to the original bug in the cases it doesn't recognize isn't a fix for it.

`inbound_opaque_test.go:65-67` also pins the arbitrary boundary as product behavior by asserting a 63-character body is preserved and a 64-character one is not. That locks in a number nothing justifies.

Please either back the matcher with a vendor contract or several redacted real callbacks, or match only the exact form we have evidence for and degrade explicitly on the rest. Recovering the text through message history remains the durable answer if it's available.

## Must-fix 2: two branches still contradict the "explicit degradation" claim

- **Unknown `msgType`** — `inbound.go:516-517`: `default: appendText(replied.Content.Text)`. This does pass through the opaque filter, so it isn't completely unguarded, but any present or future message type we haven't modelled still has its `content.text` presented to the agent as the member's prose. `inbound_boundary_test.go:75` (`"unknown kind with text"` → `"> selected"`) pins that as intended behavior.
- **Picture caption inference** — `inbound.go:495-497` treats `content.text` on a `picture`/`image` snapshot as a caption, and `dingTalkRichTextSummary` (`:536-545`) strips leading `[Image]` markers off it. The public [receive-message doc](https://open.dingtalk.com/document/orgapp/receive-message) doesn't define `repliedMsg.content.text` for pictures at all, so both the caption meaning and the marker-stripping are inferred. `inbound_test.go:391` (`TestInboundFromCallback_QuotedPictureKeepsTextSummary`) pins it.

Both read as leftovers from the previous approach rather than deliberate choices — the `interactiveCard` branch now degrades cleanly to `[quoted content unavailable]`, and these two should follow unless there's a contract behind them. The verified plain text and ordered rich-text node paths are fine as they stand.

## Must-fix 3: the PR description no longer matches the code

To be explicit, since it came up in review: **`Closes #8060` is fine — I'm not asking you to remove it.** The scope here is a reasonable close for that issue.

The description itself is stale, though, and it's what reviewers and the changelog read:

- "render selected DingTalk/**Lark/Telegram** replies as Markdown" and "This includes a deliberate Lark/Telegram output-format change" — withdrawn.
- "The generated outer `<quoted_message>` wrapper and its `message_id`/`type` attributes disappear from Lark/Telegram message bodies" — no longer true; that representation is retained.
- "readable `interactiveCard` answers" — cards now return `[quoted content unavailable]`.
- "decode current rich-text structured `text`/`content`/`data` nodes, `msgType` aliases, and direct/JSON-encoded arrays" — removed in the narrowing.
- The Changes Made table and How to Test still describe commit `4425ebbe`; head is `a279f0c1`.

Please refresh the body, the compatibility notes and the verification section to match what actually ships. A comment further down explaining the change doesn't substitute for the canonical description.

## What's good

- The new signal extends the existing adapter → Router path instead of standing up a parallel control plane, and automatic history stays clearly separated from member-selected context. That's the design I'd want.
- `/new` still consumes only the untouched original body, so `/new /new literal body` keeps its second segment literal.
- Cards degrade explicitly rather than guessing; unknown rich-text nodes degrade per-node instead of failing the whole snapshot; media keeps the occurrence-based `InlineIndex` contract.
- The diff is meaningfully smaller than last round — `inbound.go` dropped from 1132 to 797 lines. Removing the speculative card recursion and alias handling was the right call.
- No `server/internal/daemon/` changes, no migrations, no public API or deployment config changes.

## Verification notes

- Confirmed at head `a279f0c1`: targeted `go test -race` passes on the five relevant packages; `go vet` passes across the seven integration trees; `git diff --check` clean.
- Same caveats as before: the wider channel/DingTalk/Lark/Telegram race suite is blocked locally by an outdated DB schema (missing `route_revision`, `channel_chat_context_generation`, `dingtalk_group_presence`), so I'm not counting it as passing, and local frontend deps aren't installed so `vitest` didn't run. CI is green, but nothing exercises a real DingTalk callback.

Must-fix 3 is a five-minute edit. If 1 and 2 are hard to evidence, narrowing to the proven shapes and degrading elsewhere is a perfectly good outcome — I'd rather ship a smaller correct surface than a broad inferred one.

Bohan-J (changes requested):
Re-requesting review didn't bring anything new to look at — head is still `a279f0c1` (last commit 06:00 UTC, before my review at 07:52 UTC), and the description is unchanged. All three items from [my previous review](https://github.com/multica-ai/multica/pull/8061#pullrequestreview-5129387522) stand exactly as written, so I won't restate the reasoning here:

1. **Opaque-envelope matcher** — the unproven ≥64-byte / numeric-triple boundary at `dingtalk/inbound.go:640-661`, with `inbound_opaque_test.go:65-67` pinning 63-vs-64 as product behavior. The live risk is unmatched shapes falling back into the original bug.
2. **Two branches still trusted as the member's prose** — unknown `msgType` at `inbound.go:516-517`, and the picture-caption inference at `:495-497` / `:536-545`.
3. **Stale description** — still describes the withdrawn Lark/Telegram Markdown migration, readable `interactiveCard` answers and rich-text alias/array decoding, and still records verification against `4425ebbe`. To repeat: `Closes #8060` is fine and should stay — only the delivered scope and the verification record need to match `a279f0c1`.

If you pushed something that didn't make it here, worth double-checking the branch — nothing new landed on `split/channel-input`.

Item 3 is a description edit with no code behind it, so it can come off the list on its own. For 1 and 2, narrowing to the shapes we have evidence for and degrading explicitly elsewhere is still a perfectly good resolution — it doesn't need vendor documentation to land. Happy to turn a re-review around quickly once there's a commit.

yyclaw:
Addressed the three must-fix items from [this review](https://github.com/multica-ai/multica/pull/8061#pullrequestreview-5129387522) in 7ba2e8a8e and the canonical PR description:

1. Removed the opaque matcher's length, alphabet and trailer-field thresholds. Any raw quoted text containing `||` is now explicitly unavailable; legitimate quoted code/prose containing that separator is also withheld, while current input remains unchanged. This is a conservative projection policy, not a vendor envelope contract or a classifier for arbitrary ciphertext. Tests no longer pin the 63/64 boundary and cover short bodies, additional fields, different prefixes and legitimate double-pipe text as policy examples, not claimed observed provider variants.
2. Unknown `msgType` values now degrade even when `content.text` is present. Pictures retain their download references; unverified supplementary text produces an unavailable marker instead of an inferred caption, with no `[Image]` prefix stripping. Filtering happens only on raw text nodes, preserving neighboring images and current-media occurrence indexes.
3. Updated the PR title, canonical body, compatibility notes and verification section, keeping `Closes #8060`. They now describe retained Lark/Telegram XML, unavailable cards and bounded standard nodes, without the old `4425ebbe` verification or recursive-parser claims.

Validation: the new regression assertions fail against the previous production implementation. The final seven-package Go race suite and matching `go vet` passed, as did 49 message-list tests and all 58 named tests mapped to the full PR. The previous shared bare-control/`SkipAgentRun` fix, single command consumption, and original Lark/Telegram representation remain covered and passing.

Bohan-J (approved):
Thanks for working through three rounds of this — the result is meaningfully better than where it started, and I appreciate the patience.

All three blocking items are closed, and I verified each against `7ba2e8a8` rather than taking the summary on faith:

1. **Opaque handling** — the unjustified 64-byte / Base64-alphabet / numeric-triple thresholds are gone. `dingTalkReadableQuotedText` is now an honest conservative *policy* rather than a detector pretending to be one, and the doc comment states the cost plainly: legitimate quoted prose containing `||` will be withheld. That's the right direction to fail in, and naming the tradeoff in the code is worth more than a cleverer heuristic would have been.

   Moving the filter off the shared `appendText` and onto the three provider-text call sites (`:480`, `:504`, `:550`) was a good catch on your part — applied blanket, a fallback could have swallowed generated `[Image]` markers and desynced the media `InlineIndex` slots. The comment explaining exactly that is appreciated.

2. **The two inferred branches** — unknown `msgType` now degrades explicitly (`:510-511`), and the picture-caption inference is gone along with `dingTalkRichTextSummary`; the image is kept while the undocumented text is withheld. Tests moved from asserting the inferred content to asserting the degradation, which is the part that keeps this honest over time.

3. **Description** — title and body now match what ships, including an explicit "There is no Lark/Telegram Markdown migration." `Closes #8060` correctly retained.

Also confirmed: CI green across all checks on this head, `mergeable`/`CLEAN`, and the incremental diff is scoped to DingTalk parsing plus its tests.

The narrowing across these rounds — `inbound.go` went from 1132 lines at its widest down to a focused change — landed in a good place: a smaller surface we can actually stand behind, with the unproven parts degrading loudly instead of guessing. Merging now.

PR #8063, opened 2026-09-05
fix(lark): fallback to group notice when binding card unavailable

m2keloop:
## What does this PR do?

When Lark users trigger a binding flow in group chat, if the bot cannot send the private binding card because the user has no chat availability (error 230013), the bot currently returns silently after @ mention with no feedback.

This PR makes the failure user-visible by sending a group fallback notice in this specific availability-only case, while preserving existing behavior for other errors.

## Related Issue

Closes #8062

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- server/internal/integrations/lark/http_client.go: return structured APIError with code for binding prompt failures.
- server/internal/integrations/lark/http_client_test.go: add regression test for 230013 mapping.
- server/internal/integrations/lark/outcome_replier.go: detect private-card unavailability and fallback to group notice for this case only.
- server/internal/integrations/lark/outcome_replier_test.go: add tests for fallback and non-fallback branches.

## How to Test

1. Run targeted Go tests under server/internal/integrations/lark.
2. Validate:
   - 230013 sending binding prompt -> private attempt + group fallback notice.
   - other API errors -> no fallback notice.

## Risks

- Only changes user-facing behavior for Lark binding failures; no schema/migration changes.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Implemented from issue reproduction + code-path audit, then added regression tests and tuned handling to isolate fallback condition.

PR #8064, opened 2026-09-05
MUL-7076 fix(lark): fallback to group notice when binding card unavailable

### LINKED ISSUE #8062: fix(lark): fallback to group notice when binding prompt unavailable
### Issue Description
When triggering a binding flow in a Lark group, the bot may fail to send the private binding card to the user (commonly when returning code `230013` with message `Bot has NO availability to this user.`). The group only shows an @mention without any follow-up message.

### Expected
If private-card delivery fails because the user is unavailable for private chat, the bot should send a fallback message in the group to guide the user.

### Repro Steps
1. Trigger a scenario that requires a binding card in a group.
2. Make sure the target user has not established or can not receive private chat with the bot.
3. Observe the returned error code `230013`.
4. Confirm that no feedback is posted in the group.

### Fix Notes
- Change `SendBindingPromptCard` to return a structured `APIError` that carries an error code.
- Add private-unavailability detection (error code `230013` or message contains `no availability`).
- In this case, fallback to sending a group notice.
- Keep other error paths unchanged.

### Verification
- 230013 private-unavailable: fallback notice in group.
- Non-230013 (for example `230002`): no fallback.

m2keloop:
## What does this PR do?

When Lark users trigger a binding flow in group chat, if the bot cannot send the private binding card because the user has no chat availability (error 230013), the bot previously returned silently after @ mention with no feedback.

This PR makes that failure user-visible by sending a group fallback notice in this specific availability-only case, while preserving existing behavior for other errors.

## Related Issue

Closes #8062

## Type of Change

## Changes Made

- server/internal/integrations/lark/http_client.go: return structured APIError with code for binding prompt failures.
- server/internal/integrations/lark/http_client_test.go: add regression test for 230013 mapping.
- server/internal/integrations/lark/outcome_replier.go: detect private-card unavailability and fallback to group notice for this case only.
- server/internal/integrations/lark/outcome_replier_test.go: add tests for fallback and non-fallback branches.

## How to Test

1. Run targeted Go tests under server/internal/integrations/lark.
2. Validate:
   - 230013 sending binding prompt -> private attempt + group fallback notice.
   - other API errors -> no fallback notice.

## Risks

- Only changes user-facing behavior for Lark binding failures; no schema/migration changes.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Implemented from issue reproduction + code-path audit, then added regression tests and tuned handling to isolate fallback condition.

### DIFF
--- a/server/internal/integrations/lark/http_client.go
+++ b/server/internal/integrations/lark/http_client.go
@@ -71,24 +71,19 @@ const (
 	// transfers off the connector ACK path.
 	maxMessageResourceBytes = 100 << 20
 
-	// Access-credential rejections. Seeing one means the token we
-	// presented is not usable, so drop the cached entry and re-mint from
-	// /tenant_access_token/internal. Both are from Lark's generic
-	// error-code table:
-	// open.feishu.cn/document/server-docs/api-call-guide/generic-error-code.
-	//
-	// 99991663 covers BOTH halves for the only credential this client
-	// mints — "tenant_access_token 已过期" and "凭证无效" share the one
-	// code — and is the code the wedged-cache bug reported in #7611 hit.
-	//
-	// 99991664 is app_access_token, which this client never mints. It
-	// stays in the class deliberately: it is unambiguously "the
-	// credential you presented was rejected", and if Lark ever answers it
-	// for a call we authenticated with a tenant token, re-minting is the
-	// only recovery. Being wrong costs one bounded refresh and one
-	// replay, while leaving it out would cost another wedged cache.
-	codeTenantTokenInvalid = 99991663
-	codeAppTokenInvalid    = 99991664
+	// Lark's "invalid tenant_access_token" / "tenant_access_token
+	// expired" error codes. When we see either, drop the cached token
+	// so the next call refreshes from /tenant_access_token/internal.
+	// 99991663 = expired, 99991664 = invalid. Documented at:
+	// open.feishu.cn/document/server-docs/api-call-guide/server-error-codes.
+	codeTokenExpired = 99991663
+	codeTokenInvalid = 99991664
+
+	// codeNoAvailability means the bot cannot send private messages to the user
+	// because they are not in the bot's visible scope. In that case, the
+	// connector should keep the group-level reply path visible so users still get
+	// guidance.
+	codeNoAvailability = 230013
 )
 
 // HTTPClientConfig configures the production Lark HTTP APIClient.
@@ -272,48 +267,6 @@ func (c *httpAPIClient) invalidateToken(appID string) {
 	c.mu.Unlock()
 }
 
-// InvalidateTokenCache drops the cached tenant_access_token for an
-// app_id from outside the client. It exists for credential rotation:
-// re-registering a Bot issues a new app_secret, which makes Lark revoke
-// every token minted from the previous one, and the cache — keyed by
-// app_id, which does NOT change — would otherwise keep replaying a
-// token Lark now rejects.
-func (c *httpAPIClient) InvalidateTokenCache(appID string) {
-	c.invalidateToken(appID)
-}
-
-// doAuthedJSON performs one tenant_access_token-authenticated call. When
-// Lark rejects the token itself, it drops the cached entry, mints a
-// fresh token and replays the request exactly once.
-//
-// Replaying is safe for every caller here: Lark rejects a bad token
-// during authentication, before the request is processed, so nothing was
-// delivered and the retry cannot duplicate a message. Without it the
-// refresh would only help the NEXT call and the reply that triggered it
-// would still be lost.
-//
-// Only token errors are retried. Business codes, 5xx and network
-// failures are returned untouched — those are either definitive or
-// ambiguous about delivery, and neither is fixed by a new token.
-func (c *httpAPIClient) doAuthedJSON(ctx context.Context, creds InstallationCredentials, method, path string, body, out any) error {
-	token, err := c.tenantAccessToken(ctx, creds)
-	if err != nil {
-		return err
-	}
-	err = c.doJSON(ctx, c.resolveBaseURL(creds), method, path, token, body, out)
-	if !isTokenError(larkErrorCode(err)) {
-		return err
-	}
-	c.cfg.Logger.Warn("lark http client: tenant_access_token rejected; refreshing and retrying once",
-		"app_id", creds.AppID, "path", path, "err", err)
-	c.invalidateToken(creds.AppID)
-	fresh, refreshErr := c.tenantAccessToken(ctx, creds)
-	if refreshErr != nil {
-		return fmt.Errorf("%w (token refresh failed: %v)", err, refreshErr)
-	}
-	return c.doJSON(ctx, c.resolveBaseURL(creds), method, path, fresh, body, out)
-}
-
 // outboundMessageRequest builds the (path, body) the three send methods
 // share. When target.IsSet() the message is routed through Lark's reply
 // endpoint (POST /im/v1/messages/{message_id}/reply) so it threads back
@@ -350,6 +303,10 @@ func (c *httpAPIClient) SendInteractiveCard(ctx context.Context, p SendCardParam
 	if p.CardJSON == "" {
 		return "", errors.New("lark http client: missing card json")
 	}
+	token, err := c.tenantAccessToken(ctx, p.InstallationID)
+	if err != nil {
+		return "", err
+	}
 	path, body := outboundMessageRequest(p.ChatID, "interactive", p.CardJSON, p.ReplyTarget)
 	var resp struct {
 		Code int    `json:"code"`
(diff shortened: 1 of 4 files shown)


Bohan-J (changes requested):
Thanks for the follow-up on #8063 — the direction here is right. Today a member who @-mentions the bot in a group and cannot receive the binding card gets nothing back but the mention, and this PR closes that dead end by reusing the existing `sendChatNotice` path instead of building a parallel sender. No migration, no new config, no retry machinery — good.

Two things need to change before merge, both about how narrow the trigger is. The new behavior publishes "this member is not bound yet" into a shared chat, so the classifier has to be exact.

## Must fix

**1. Use the existing structured error-code helper; drop the text match.** `server/internal/integrations/lark/outcome_replier.go:258`

`isBindingPromptUnavailable` only inspects `*APIError` and then falls back to matching `"no availability"` in the error string. That misses one real case and invents another:

- Lark reports some failures as **non-2xx HTTP with the business code in the body** — that shape is `larkAPIStatusError`, not `APIError` (see the `#7611` comment in `http_client.go`). A `230013` delivered that way is skipped by the `errors.As` branch, and only caught if the message text happens to match. If Lark changes or localizes the copy, the fallback silently stops firing.
- Conversely, a transport/proxy/gateway error whose text happens to contain `no availability` will trigger a public group notice — including cases where the DM may actually have been delivered.

The repo already has one helper covering both shapes: `larkErrorCode(err)` (`http_client.go:1242`) reads the code from either `larkAPIStatusError` or `APIError`, and deliberately returns `0` for network failures and timeouts so ambiguous transport errors can never match a code-keyed lookup. This is exactly the distinction the threaded-reply fallback already relies on.

```go
func isBindingPromptUnavailable(err error) bool {
	return larkErrorCode(err) == codeNoAvailability
}
```

Please add two tests to pin it: non-2xx carrying `230013` with arbitrary `msg` **does** classify; an error with no business code whose text contains `no availability` **does not**.

**2. Gate the fallback on group chats.** `server/internal/integrations/lark/outcome_replier.go:247`

`OutcomeNeedsBinding` is produced by a shared identity check that runs for p2p as well as group messages, so as written a `230013` in a **direct** chat also sends the notice — beyond the "group notice" scope stated in this PR and in #8062, and a little odd (telling someone in their DM that we could not DM them).

Please add a `msg.ChatType == ChatTypeGroup` guard (`ChatTypeGroup` is in `lark/types.go`) and cover both sides — group triggers, p2p does not. Note `TestLarkOutcomeReplierNeedsBindingFallsBackToGroupNoticeWhenPrivateUnavailable` (`outcome_replier_test.go:386`) does not set `ChatType`, so the test name does not currently pin group behavior.

If you actually want the p2p fallback too, say so and let a maintainer accept that wider scope explicitly rather than having it arrive implicitly.

## Nits

- `outcome_replier.go:248-249`: when the fallback send itself fails, the original DM error is dropped and only the second error is returned. Consider keeping both (as `sendWithThreadFallback` does) so logs still show why we entered the fallback.
- The new copy mixes 私聊 and 私信, and "该用户侧的可见性" reads like ops jargon. Suggest settling on 私聊 and making the action concrete: 请先打开机器人对话并发送一条消息，再回到群里重试；仍失败请联系管理员检查应用可用范围.

## Verification request

Could you attach a screenshot or short screen recording from the real tenant where you hit `230013`? Specifically: the group message that triggers binding, and the fallback notice as it appears in the group. This path is hard to reproduce without a tenant whose visibility scope actually excludes the bot, so a real capture is the only way we can confirm the rendering and wording land as intended.

## Local verification on my side

- Head merges cleanly with current `main`; `git diff --check` clean.
- Relevant `SendBindingPromptCard` / `NeedsBinding` / existing success-path tests pass on a local merge with `main`.
- `go vet ./internal/integrations/lark` clean.
- I could **not** complete a full `go test ./internal/integrations/lark -count=1`: my local shared test database is missing later schema (`dingtalk_group_presence`, `channel_task_delivery`), so DB-backed tests fail for reasons unrelated to these four files. Flagging so my "tests pass" is not read as broader than it is. CI on the PR is green.

m2keloop:
Thanks for the detailed review. The requested fixes have been implemented and verified in our internal repository:

- Classify private-card unavailability using structured business code `230013` only, including non-2xx responses; remove error-text matching.
- Restrict the fallback to `ChatTypeGroup`; p2p and unspecified chat types do not trigger a notice.
- Preserve both the original binding error and the fallback-send error.
- Update the Chinese recovery instructions.
- Add regression coverage for non-2xx `230013` with arbitrary message text, errors without business codes, group/p2p behavior, and preservation of both errors.

Verification passed:
```sh
cd server
go test ./internal/integrations/lark -count=1
go vet ./internal/integrations/lark
```

I confirm that the screenshot below was captured in a real group chat and shows the updated fallback notice. It does not include the triggering message. Identifying information has been redacted for privacy.



Note: the fixes above have been merged and deployed internally; this comment alone does not update this GitHub PR's code.
f5ad833d68969098755380ac36f7c9e3

multica-eve (changes requested):
When an unbound member mentions the bot in a group but cannot receive the private binding card, they should get clear instructions to open the bot conversation, send a message, and retry in the group. Today that failure leaves them without a next step.

That recovery is worth adding, provided the notice is limited to group conversations where private delivery is definitively unavailable. The current version still exceeds that boundary, so **changes remain required**.

Thanks for the screenshot and the clarification that your fixes have been deployed internally. This PR still points to `6f2293ae21558f8e171a5652cb9bd08b9fc72096`, the same commit as the previous review. Please push those fixes and their regression tests to this PR so they can be reviewed here.

Both blockers remain in the public code, and I reproduced them on that commit:

1. **Use structured error codes only.** `server/internal/integrations/lark/outcome_replier.go:258` still inspects only `APIError` before matching `no availability` in the error text. A non-2xx `larkAPIStatusError` carrying `230013` with arbitrary wording produces zero notices when one is expected. Conversely, an unstructured transport error containing that phrase produces a notice when none is expected. Use the existing `larkErrorCode(err) == codeNoAvailability`, which covers both structured response shapes, and remove text matching.
2. **Restrict the fallback to group chats.** `server/internal/integrations/lark/outcome_replier.go:247` still lacks `msg.ChatType == ChatTypeGroup`. Both p2p and unspecified chat types produce a notice on `230013`. Add the guard and tests for group delivery plus p2p/unspecified suppression. The shared identity check can produce `OutcomeNeedsBinding` for direct conversations too.

The existing notice sender and topic routing are suitable to reuse. Please also bring over the original-error preservation and revised wording described in your reply; both previous nits remain unchanged in this PR.

The screenshot makes the revised instructions readable, but it omits the triggering message. Please supplement it with a redacted real-group capture or short recording showing the triggering message and resulting notice together, and identify the PR commit used for that verification.

Validation on this commit: five existing binding-related tests and `go vet ./internal/integrations/lark` passed. Temporary regression probes reproduced all four failing boundary cases above; the group `230013` and unrelated-business-error controls passed. The probes were removed afterward. I did not run the full database-backed package suite or verify against a live tenant.

If shipped as-is, unbound members could see missing or unintended notices; reverting the service release would stop the added behavior. Please update this PR before requesting another review.

m2keloop:
@multica-eve The requested fixes and regression tests have now been pushed to this PR.

Commit: 82e2ab460b45319e2fac134f44a1b9abd5313229

Updated:
- Use `larkErrorCode(err) == codeNoAvailability` and remove error-text matching.
- Restrict the fallback to `ChatTypeGroup`; p2p and unspecified chat types no longer send the notice.
- Preserve both the original private binding prompt error and the fallback-send error.
- Update the Chinese recovery instructions.
- Add regression coverage for non-2xx `230013`, unstructured `no availability` text, group/p2p/unspecified behavior, and fallback error preservation.

Verified:
```sh
cd server
go test ./internal/integrations/lark -run 'TestLarkOutcomeReplierNeedsBinding' -count=1
go vet ./internal/integrations/lark
```

I'll supplement the real-tenant redacted group capture showing both the triggering message and the fallback notice.

Bohan-J (approved):
Thanks for the thorough follow-through here — and for pushing the fixes and the regression tests to this PR rather than leaving them internal. Re-verified everything at `82e2ab460`:

- `outcome_replier.go:259` now classifies on `larkErrorCode(err) == codeNoAvailability`, so a non-2xx reply carrying `230013` is caught regardless of `msg` wording, and a transport error with no business code can never trigger a public notice. The text match is gone.
- `outcome_replier.go:247` gates the fallback on `msg.ChatType == ChatTypeGroup`, matching the scope stated in #8062.
- Both nits addressed: the fallback error now wraps the original binding-prompt error alongside the send failure, and the copy reads as a concrete recovery step.
- Test coverage lines up with what was asked: group-chat positive, non-2xx `230013` positive, unstructured `no availability` text negative, a p2p / unspecified-chat-type suppression table, and the error-preservation case.

The real-tenant capture also looks right. CI is green on this commit. Merging now — thanks for the contribution and for the patience through two review rounds.

PR #8074, opened 2026-09-05
[MUL-7086] feat(auth): add workspace scope to personal access tokens

v1xingyue:
## Summary

Personal access tokens are currently account-scoped, which makes it impossible to issue a least-privilege token for one workspace. This change adds an optional workspace binding while preserving existing account-level tokens.

## Changes

- Add nullable `workspace_id` to `personal_access_token` with an index and rollback migration.
- Accept and return `workspace_id` when creating/listing tokens.
- Reject workspace-bound tokens when a request selects a different workspace.
- Keep workspace-bound tokens off the legacy user-only auth cache.
- Add the field to generated SQLC models and shared API types.

## Compatibility

Existing tokens with `workspace_id = NULL` retain current account-level behavior. New workspace-bound tokens are enforced server-side and the bound workspace is propagated to downstream workspace resolution.

## Validation

`go test ./internal/handler ./internal/middleware`

### DIFF
--- a/packages/core/types/api.ts
+++ b/packages/core/types/api.ts
@@ -554,11 +554,13 @@ export interface PersonalAccessToken {
   expires_at: string | null;
   last_used_at: string | null;
   created_at: string;
+  workspace_id?: string | null;
 }
 
 export interface CreatePersonalAccessTokenRequest {
   name: string;
   expires_in_days?: number;
+  workspace_id?: string;
 }
 
 export interface CreatePersonalAccessTokenResponse extends PersonalAccessToken {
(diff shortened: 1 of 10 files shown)


multica-eve (changes requested):
Before this change, a personal access token can be used across every workspace its owner may access. After this change, an administrator could restrict a new token to one workspace while existing tokens retain their current reach.

**Product judgment: ❌ Reject this proposal.** The workspace-scoped token capability may be worth pursuing, but this PR makes workspace deletion automatically delete related tokens through the database. That lifecycle behavior does not match Multica's approved architecture, so we should not accept this proposal as designed.

**Blocking implementation issue:** the migration adds a foreign key from `personal_access_token.workspace_id` to `workspace.id` with `ON DELETE CASCADE`. Multica explicitly forbids database foreign keys and cascading constraints; entity relationships, validation, and deletion semantics must be enforced in the application layer.

**Merge recommendation: ❌ Reject.** A replacement proposal is welcome if it models the relationship without a foreign key or cascade and explicitly handles workspace validation and deletion behavior in application code.

PR #8075, opened 2026-09-05
fix(claude): preserve results and keep native loops running (MUL-7087)

mameikagou:
## Reported failure

A native Claude `/loop` run failed with:

```text
claude stream ended without terminal result; claude stderr: [claude-code:unrecognized_model] {"model":"k3-256k","query_source":"sdk"}
```

The incident logs identify the regression:

| Log time | What happened |
| --- | --- |
| 22:18:13 | `ScheduleWakeup` reported success for a wakeup requested 1,400 seconds later. |
| 22:18:28 | Claude produced a complete answer and a normal terminal `result`. Multica then cleared `sawResult` and the final text while waiting for another iteration. |
| 22:18:33 | The CLI exited with `exit_code=0`, without a second result. Multica finalized with no retained terminal result, losing the answer and reporting the error above. |

The custom-model warning also appeared in successful runs, so it did not establish the cause of this failure. The decisive evidence was the valid result followed by a clean exit.

A follow-up correction retained the answer but always closed stdin after the first result. That prevented native scheduling from reaching its next iteration. The adapter needed to distinguish a completed iteration, a scheduled wait, and the start of a subsequent turn.

## Code changes

- **Native scheduling receipts — `server/pkg/agent/claude_schedule.go`:** track main-thread `ScheduleWakeup` and `CronCreate` calls together with their structured results. Require a positive `scheduledFor` or an accepted cron receipt; a non-error tool result alone is insufficient because disabled scheduling can return `scheduledFor: 0`. Failed, text-only and subagent receipts do not arm a wait. Native stop and cron-deletion receipts clear the corresponding schedules.
- **Stream lifetime — `server/pkg/agent/claude.go`:** keep stdin open while a confirmed schedule remains, allowing Claude to own timer dispatch and session reuse. Close input after ordinary completion, terminal failure, or removal of the remaining schedule.
- **Result and usage handling — `claude.go`:** retain the completed answer throughout the wait, resetting terminal state only when the next main-thread turn starts. Clean EOF while waiting preserves the answer; a subsequent turn that starts but never returns a terminal result still fails. Terminal usage snapshots replace accumulated totals to avoid double-counting across iterations.
- **Daemon liveness — `server/pkg/agent/agent.go` and `server/internal/daemon/daemon.go`:** carry the expected wakeup time in `Message.WaitingUntil`. The idle watchdog allows silence until that time, then applies its ordinary idle budget. Incidental logs preserve the scheduled wait; a running transition restores normal hang detection immediately. Cancellation and hard task deadlines still apply.

## Why Codex `/goal` worked while Claude `/loop` failed

| | Codex `/goal` | Claude `/loop` |
| --- | --- | --- |
| Continuation policy | Work across turns toward an objective and a verifiable stopping condition. | Run another iteration when a scheduled wakeup becomes due. |
| Typical use | Finish a migration and validate the result. | Check CI repeatedly and address failures. |
| Cadence | Driven by progress toward the objective. | Fixed cron interval or a delay Claude selects each iteration. |
| Multica integration | `codex app-server --listen stdio://`, with its own notification/turn handling. | `claude -p` with stream-json input/output and per-iteration results. |

The working Codex path never executed the faulty Claude-specific reset/close logic. This was an adapter lifecycle bug, rather than evidence that Claude cannot remain active. The transport difference alone does not guarantee correct continuation; each backend must respect its runtime's completion boundaries.

Official references: [Codex goal workflow](https://learn.chatgpt.com/use-cases/follow-goals) and [Claude native scheduling](https://code.claude.com/docs/en/scheduled-tasks).

## Verification

- Full `pkg/agent` and `internal/daemon` suites and `go vet` passed on both the PR branch and the maintained personal branch.
- Regression and focused race tests cover confirmed/invalid scheduling receipts, two-turn execution, stop, cancellation, clean EOF during a wait, and incomplete subsequent turns. Daemon tests check the scheduled idle extension, restoration of ordinary hang detection, and cancellation.
- The opt-in integration test runs **real Claude Code 2.1.246 against a local mock Messages API**. Dynamic `/loop` and fixed `/loop 1m` each enter a second iteration in the same process/session, then stop through `ScheduleWakeup(stop: true)` and `CronDelete`, respectively. Native CLI expansion, timers, tool execution and stream handling are exercised; model responses are mocked and no authenticated model account is used.
- Real-CLI execution requires the `agentintegration` build tag and `MULTICA_RUN_REAL_AGENT_SMOKE=1` before executable lookup. `MULTICA_CLAUDE_EXECUTABLE` optionally selects the binary.

Coverage establishes the confirmed dynamic-wakeup and cron paths during an active execution; it does not establish compatibility with every Claude background mechanism or CLI version. Native scheduling availability remains controlled by Claude Code.

The implementation is mirrored to `mameikagou/multica`'s `local/native-context-chat-handoff` branch at `bcc74f93ed`, and a clean Linux amd64 daemon build was produced from that commit. That build step did not restart the running daemon.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -8749,7 +8749,8 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 
 	var toolCount atomic.Int32
 	// lastActivityAt records (as unix nanos) when the drain loop most
-	// recently received a message from the backend. The idle watchdog
+	// recently received a message from the backend, or the end of a confirmed
+	// native scheduled wait, whichever is later. The idle watchdog
 	// reads this to decide whether the agent has gone silent for too long.
 	// Initialise to the start so a backend that never emits a single
 	// message also trips the watchdog.
@@ -8844,6 +8845,7 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 		}()
 
 		var sessionPinned atomic.Bool
+		var waitingUntil time.Time
 		for {
 			select {
 			case msg, ok := <-session.Messages:
@@ -8855,7 +8857,14 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 				// gone silent — stamping before processing makes sure a
 				// slow downstream call (mu.Lock contention, batch resize)
 				// can't be misattributed to backend silence.
-				lastActivityAt.Store(time.Now().UnixNano())
+				activityAt := time.Now()
+				if msg.Type == agent.MessageStatus {
+					waitingUntil = msg.WaitingUntil
+				}
+				if waitingUntil.After(activityAt) {
+					activityAt = waitingUntil
+				}
+				lastActivityAt.Store(activityAt.UnixNano())
 				switch msg.Type {
 				case agent.MessageStatus:
 					// Persist the session/work_dir as soon as the backend
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -8748,10 +8748,10 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 	defer drainCancel()
 
 	var toolCount atomic.Int32
-	// lastActivityAt records (as unix nanos) when the drain loop most
-	// recently received a message from the backend, or the end of a confirmed
-	// native scheduled wait, whichever is later. The idle watchdog
-	// reads this to decide whether the agent has gone silent for too long.
+	// lastActivityAt records (as unix nanos) when the drain loop most recently
+	// received a message from the backend. Confirmed native scheduled waits live
+	// only in Session.Liveness; Message.WaitingUntil is a display field and must
+	// not become a second, uncleared watchdog-control source.
 	// Initialise to the start so a backend that never emits a single
 	// message also trips the watchdog.
 	var lastActivityAt atomic.Int64
@@ -8845,7 +8845,6 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 		}()
 
 		var sessionPinned atomic.Bool
-		var waitingUntil time.Time
 		for {
 			select {
 			case msg, ok := <-session.Messages:
@@ -8858,12 +8857,6 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 				// slow downstream call (mu.Lock contention, batch resize)
 				// can't be misattributed to backend silence.
 				activityAt := time.Now()
-				if msg.Type == agent.MessageStatus {
-					waitingUntil = msg.WaitingUntil
-				}
-				if waitingUntil.After(activityAt) {
-					activityAt = waitingUntil
-				}
 				lastActivityAt.Store(activityAt.UnixNano())
 				switch msg.Type {
 				case agent.MessageStatus:
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -8777,7 +8777,7 @@ func (d *Daemon) executeAndDrain(ctx context.Context, backend agent.Backend, pro
 	var idleWatchdogThreshold atomic.Int64
 	idleWatchdogThreshold.Store(int64(idleWindow))
 	if idleWindow > 0 {
-		go d.runIdleWatchdog(agentCtx, idleWindow, d.cfg.AgentToolWatchdog, &lastActivityAt, &inFlightTools, &idleWatchdogFired, &idleWatchdogThreshold, agentCancel, session.Messages, taskLog)
+		go d.runIdleWatchdog(agentCtx, idleWindow, d.cfg.AgentToolWatchdog, &lastActivityAt, &inFlightTools, &idleWatchdogFired, &idleWatchdogThreshold, agentCancel, session.Messages, session.Liveness, taskLog)
 	}
 
 	// drainFinished closes after the drain goroutine has flushed the last
@@ -9122,7 +9122,7 @@ func idleWatchdogTickInterval(window time.Duration) time.Duration {
 //
 // Polling rate comes from idleWatchdogTickInterval, so a run is force-stopped
 // somewhere between its budget and budget + tick, never earlier.
-func (d *Daemon) runIdleWatchdog(agentCtx context.Context, window, toolWindow time.Duration, lastActivityAt *atomic.Int64, inFlightTools *atomic.Int32, fired *atomic.Bool, firedThreshold *atomic.Int64, cancel context.CancelFunc, messages <-chan agent.Message, taskLog *slog.Logger) {
+func (d *Daemon) runIdleWatchdog(agentCtx context.Context, window, toolWindow time.Duration, lastActivityAt *atomic.Int64, inFlightTools *atomic.Int32, fired *atomic.Bool, firedThreshold *atomic.Int64, cancel context.CancelFunc, messages <-chan agent.Message, liveness *agent.SessionLiveness, taskLog *slog.Logger) {
 	ticker := time.NewTicker(idleWatchdogTickInterval(window))
 	defer ticker.Stop()
 	for {
@@ -9143,6 +9143,9 @@ func (d *Daemon) runIdleWatchdog(agentCtx context.Context, window, toolWindow ti
 				threshold = toolWindow
 			}
 			last := time.Unix(0, lastActivityAt.Load())
+			if waitingUntil := liveness.WaitingUntil(); waitingUntil.After(last) {
+				last = waitingUntil
+			}
 			idleFor := time.Since(last)
 			if idleFor < threshold {
 				continue
--- a/server/internal/daemon/daemon_test.go
+++ b/server/internal/daemon/daemon_test.go
@@ -3323,11 +3323,20 @@ func TestExecuteAndDrain_ContextCancelled_ReportsCancelled(t *testing.T) {
 // idle watchdog is the only thing that ends this otherwise-forever-silent run.
 type idleWatchdogBackend struct {
 	emitOne bool // when true, emit one message before going silent; when false, never emit anything
+	waitFor time.Duration
+	resume  bool
 }
 
 func (b idleWatchdogBackend) Execute(_ context.Context, _ string, _ agent.ExecOptions) (*agent.Session, error) {
-	msgCh := make(chan agent.Message, 1)
+	msgCh := make(chan agent.Message, 3)
 	resCh := make(chan agent.Result)
+	if b.waitFor > 0 {
+		msgCh <- agent.Message{Type: agent.MessageStatus, Status: "waiting", WaitingUntil: time.Now().Add(b.waitFor)}
+		msgCh <- agent.Message{Type: agent.MessageLog, Content: "still waiting"}
+		if b.resume {
+			msgCh <- agent.Message{Type: agent.MessageStatus, Status: "running"}
+		}
+	}
 	if b.emitOne {
 		msgCh <- agent.Message{Type: agent.MessageText, Content: "hello"}
 	}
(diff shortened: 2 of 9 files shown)


multica-eve (changes requested):
以前，成员启动定时循环后，第一次检查结束就可能显示任务已完成，之后承诺的检查不再发生；改动后，已确认安排下一轮的任务会继续等待并按时重复，直到成员或助手停止循环，同时保留最后答复与累计用量。

## 产品判断

✅ **产品上应该接受。** `/loop` 是已有的正式用户入口，这个 PR 选择只在拿到结构化的“下一轮已成功安排”回执后保持任务等待，产品边界正确；普通完成、拒绝安排、错误、取消和明确停止仍立即结束。

典型路径是：成员输入 `/loop 25m 检查部署` → 第一轮检查完成并确认下一轮 → 任务继续等待 → 25 分钟后再次检查 → 停止循环后才显示完成。影响仅限使用 Claude 原生 `/loop`/定时任务的成员；普通任务不变，无数据或配置迁移。

## Must-fix

1. **合法的周日 cron 会被误判为无调度，导致第一轮后直接结束。** `server/pkg/agent/claude_schedule.go:81` 使用 `robfig/cron/v3` 标准解析器，其星期范围是 `0–6`；但 [Claude 官方 cron 语法](https://code.claude.com/docs/en/scheduled-tasks#cron-expression-reference) 允许 `0` 或 `7` 表示周日。Claude 接受 `0 9 * * 7` 后，本地解析会失败，随后 `server/pkg/agent/claude.go:298` 关闭输入，刚创建的任务永远不会执行。请按 Claude 的完整语法归一化/解析，并补 `7`、范围和列表的回归测试。

2. **已确认下一轮后进程提前退出仍会被报告为成功。** `server/pkg/agent/claude.go:287-319` 在等待期间保留 `sawResult=true`，流结束后没有检查“仍有未兑现的调度”，最终会把上一轮结果当作完整成功；`server/pkg/agent/claude_schedule_test.go:174-176` 还固化了这个结果。官方说明定时任务[只在会话保持运行时触发](https://code.claude.com/docs/en/scheduled-tasks#limitations)，因此确认调度后却在下一轮前退出意味着循环已经丢失，应显式失败。

3. **投递 waiting 状态的阻塞发送会引入管道死锁。** `server/pkg/agent/claude.go:292-295` 对 `msgCh` 阻塞发送，但 `server/pkg/agent/agent.go:20` 明确约定 `Session.Messages` 可以不消费。第一轮产生至少 256 条消息且调用方只等待 `Session.Result` 时，这里会无限期阻塞，scanner 停止读取后子进程也可能随管道填满而停住。请把 watchdog 控制信号放到不依赖可选消息消费者的专用通道/状态上，并补“填满消息缓冲、只读取 Result”的回归测试。

4. **7 天到期时间没有约束实际 watchdog 豁免。** `server/pkg/agent/claude_schedule.go:88` 保存了 `expires`，但 `waitingUntil()`（`:119-133`）只返回下一次 cron 时间加抖动。对每年一次的 recurring cron，它会把 daemon 活跃时间推进到下一年；若 Claude 的到期触发或调度器卡住，任务会远超官方 [7 天生命周期](https://code.claude.com/docs/en/scheduled-tasks#seven-day-expiry) 持有 daemon 并发槽和工作目录。等待上限应取下一次触发边界与到期边界的较早者，并覆盖“下一次触发晚于到期”的测试。

实现中复用现有 Claude 生命周期与 daemon idle watchdog、只信结构化调度回执、排除失败回执和子代理回执，以及保留上一轮结果的方向是合理的。用量处理也保持了同一 stream-json 进程的累计语义；没有数据库、权限、租户隔离或持久格式变化。

## 验证

- `go test ./pkg/agent -count=1`：通过。
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1`：通过。
- 相关 `go test -race`、`go vet ./pkg/agent ./internal/daemon`、execenv byte-identical 测试和 `git diff --check`：通过。
- 当前 GitHub CI：15/15 通过。
- 未运行真实 Claude CLI `agentintegration` smoke：仓库规则要求显式授权，本次没有该授权。

## Merge Recommendation

🔄 **Request changes** — 产品方向正确，但以上 4 个状态机/协议边界会造成合法循环被截断、失败被误报为成功或 daemon 任务长期卡住；修复并补回归测试前不应合入。

mameikagou:
Addressed all four requested changes in 68122ef1b:

1. Claude's Sunday value `7` is normalized before robfig parsing. Literal, list, range, and stepped-range forms are covered (`7`, `1,3,7`, `5-7`, `5-7/2`), while normal `0-6` expressions pass through unchanged.
2. A process exit after a confirmed scheduled wait now fails the run explicitly instead of returning the previous turn's checkpoint as success.
3. Watchdog control no longer depends on the optional `Session.Messages` stream. Confirmed waiting state is published through dedicated atomic session liveness state, and the UI status send is non-blocking. A regression test fills the 256-message buffer, never consumes it, and still receives the final result without deadlocking.
4. The native-wait exemption is capped at the earlier of the next trigger boundary and the existing seven-day schedule expiry, after which the ordinary idle watchdog applies.

Validation completed:
- `go test ./pkg/agent`
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon`
- focused race run, 20 repetitions
- `go vet ./pkg/agent ./internal/daemon`
- Windows amd64 cross-compilation for both packages

This keeps Claude's native `/loop` ownership intact while bounding daemon liveness and making lost loops fail visibly.

multica-eve (changes requested):
Members expect scheduled checks to keep running until they are deliberately stopped. This update fixes the four failure modes from the previous review, but one-time reminders can still be terminated days before their promised time, remain stuck after they have already fired, or delay recovery when a newly started iteration hangs.

## Product judgment

✅ **The product change should be accepted.**

1. **Is the problem real? Yes.** I independently verified the original lifecycle failure and reproduced the remaining one-time reminder and liveness failures on the current head.
2. **Is it worth maintaining? Yes.** Native `/loop` and one-time reminders are existing Claude user journeys; keeping the Multica task alive for their actual lifecycle closes those journeys rather than introducing a separate feature surface.
3. **Would we solve it this way? Yes, with the remaining boundaries fixed.** We should extend the existing Claude session and daemon idle-watchdog path only after a structured scheduling receipt, while preserving the distinct recurring and one-shot semantics.

The affected scope is Claude-native scheduled work only. Ordinary agent tasks, persisted formats, configuration, permissions, and tenant isolation are unchanged.

## Re-review of `68122ef1b954`

The new commit correctly addresses all four findings from the previous review:

- Sunday value `7` is normalized for literal, list, range, and stepped-range forms.
- Exiting before a confirmed wakeup now fails instead of returning the previous checkpoint as success.
- Watchdog control uses dedicated atomic session liveness, and the display-message send is non-blocking.
- Recurring cron waits are bounded by the seven-day expiry.

Three must-fix issues remain in the complete diff:

1. **[must-fix] One-shot cron entries incorrectly receive the recurring-only seven-day expiry.** `server/pkg/agent/claude_schedule.go:95` assigns `now+7d` to every cron entry, `:115` deletes every entry after that deadline, and `:136-138` unconditionally caps every wait at it. Claude documents that [the seven-day expiry applies to recurring tasks](https://code.claude.com/docs/en/scheduled-tasks#seven-day-expiry); an unexpired one-shot remains valid until its scheduled time. A reminder scheduled one month ahead therefore resumes the ordinary idle watchdog on day seven and is terminated early. A minimal regression on this head returned `2026-09-15` instead of the actual `2026-10-01` fire time. Apply expiry only to recurring entries and add a `one-shot > 7d` regression.

2. **[must-fix] An early one-shot fire remains recorded as pending after it has completed.** Claude's [jitter rules](https://code.claude.com/docs/en/scheduled-tasks#jitter) allow one-shots at `:00` or `:30` to fire up to 90 seconds early, but `beginTurn()` in `server/pkg/agent/claude_schedule.go:110-123` consumes a one-shot only when `now >= entry.next`. If a 13:00 reminder legally fires at 12:58:31, the final result reaches `server/pkg/agent/claude.go:289-296` while the entry is still pending, so stdin stays open and the completed reminder waits until the watchdog eventually stops it. I reproduced this with an 89-second-early fire. Prefer consuming the exact task from a structured scheduled-fire identifier. If upstream exposes no such identifier, any 90-second tolerance also needs a regression proving that an unrelated pre-due notification cannot consume the reminder.

3. **[must-fix] Clearing dedicated liveness still depends on a best-effort `running` display message to clear the daemon's cached deadline.** `server/pkg/agent/claude.go:222-230` clears `Session.Liveness` and then publishes `running` with `trySend`, which may drop it when the optional message buffer is full. However, `server/internal/daemon/daemon.go:8848-8867` also caches the prior waiting message's future deadline in `lastActivityAt`, and `:9145-9148` takes the later of that value and dedicated liveness. If `running` is dropped, a newly started turn that subsequently hangs retains the old exemption—up to another 30 minutes for recurring cron—despite liveness already being clear. A cross-layer reproduction with a 50 ms idle budget fired after 354 ms because the old 300 ms wait remained active. The current daemon test always delivers `running`, while the buffer-full test does not pass through the daemon, so neither covers this intersection. Make `Session.Liveness` the sole watchdog-control source and keep `Message.WaitingUntil` display-only, or otherwise ensure clearing never depends on an optional message.

The EOF dependency chain remains acyclic: during a confirmed wait, the adapter keeps stdin open; Claude's timer emits the next turn on stdout; a terminal result with no further schedule closes stdin; Claude exits and closes stdout; the scanner ends; and `cmd.Wait` completes before the result is delivered. Cancellation closes stdin, terminates the process group, and closes stdout to unblock the scanner. The new commit removes the message-buffer deadlock, but finding 3 still delays cleanup after a new turn hangs.

If these boundaries are wrong, users will see distant reminders disappear early, completed one-time reminders continue occupying a task, or hung iterations retain daemon capacity longer than intended. Reverting this PR restores the previous lifecycle but also restores the original `/loop` first-iteration failure; the affected population is every Multica task using Claude-native scheduling.

## Validation

- Relevant `go test -race` suites for Claude scheduling, buffer saturation, and daemon scheduled-wait/cancellation paths passed.
- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1` passed.
- `go vet ./pkg/agent ./internal/daemon`, the daemon `ByteIdentical` invariant test, and `git diff --check` passed.
- GitHub CI is green (15/15 checks).
- `go test ./pkg/agent -count=1` hit one unrelated timing assertion in a Codex handshake test (8.20 s vs `<8 s`); the same test passed in isolation.
- Three minimal reviewer regressions reproduce the findings above on this head; the temporary tests were removed and the worktree is clean.
- I did not run the real Claude CLI `agentintegration` smoke because repository policy requires explicit authorization.

## Merge recommendation

🔄 **Request changes** — the previous four findings are resolved, but the one-shot lifecycle/jitter behavior and the liveness-clear watchdog path still have three reproducible merge blockers.

mameikagou:
Addressed all three remaining review blockers in `2666c08`.

1. Seven-day expiry is now assigned and enforced only for recurring cron entries. A regression creates a one-shot on September 5 for October 1, advances past day 7, and verifies the wait still ends at the October 1 fire time.
2. Early one-shot completion is now tied to Claude's dedicated `system/scheduled_task_fire` event. Only that structured marker enables the documented 90-second early window, and only for one-shots scheduled at `:00` or `:30`; an ordinary pre-due main-thread notification cannot consume the reminder. The regression reproduces a 13:00 one-shot firing at 12:58:31 and first proves the same timestamp without the fire marker leaves it pending.
3. `Session.Liveness` is now the sole watchdog-control source. `Message.WaitingUntil` remains display-only and is never copied into `lastActivityAt`. The cross-layer regression fills the message buffer so the `running` display event is definitely dropped, clears dedicated liveness, and verifies a 50 ms watchdog does not retain the old 300 ms exemption. This path passed 20 repetitions under the race detector.

For item 2, I also checked the locally installed Claude Code 2.1.263 executable rather than assuming the wire shape. Its stream schema declares `scheduled_task_fire` with only `content`, `uuid`, and `session_id`; the internal transcript object has `taskId`, but the stream serializer intentionally strips it. Therefore Multica cannot match an exact cron ID from stream-json today. Gating the tolerance on the dedicated fire event is the narrowest observable fallback and keeps unrelated notifications safe.

Validation:

- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon -count=1`
- focused Claude scheduling/native-loop race suite, 10 repetitions
- focused cross-layer daemon liveness race suite, 20 repetitions
- `go vet ./pkg/agent ./internal/daemon`
- Windows amd64 cross-compilation for both packages
- `git diff --check`

PR #8081, opened 2026-09-05
MUL-7092: fix(runtime): show configured Qwen models

Biyocon:
## What does this PR do?

An online, authenticated Qwen Code runtime currently returns an empty model catalog, so agent settings show "No models available" even though Qwen's `/model` picker has configured models. This change reads the account's configured model IDs from Qwen's local `settings.json`, marks the active route as default, and preserves the existing Default/manual-entry behavior when no settings file exists.

The discovery path parses only `modelProviders` and the selected `model`. It does not expose Qwen's `env` block, invoke a model, or log credentials. ACP route IDs are deliberately not used because Multica launches Qwen headlessly with `qwen --model`, which accepts the configured model ID.

## Related Issue

Observed with Multica Desktop v0.4.40 and Qwen Code v0.23.0.

## Type of Change

## Changes Made

- Discover Qwen models from `QWEN_HOME/settings.json` or `~/.qwen/settings.json`.
- Deduplicate identical model IDs and prefer the currently selected provider route for the label.
- Keep the catalog empty when settings are absent so Default and manual model entry remain available.
- Add unit, malformed-config, missing-config, and opt-in real-account smoke coverage.

## How to Test

1. Run `go test ./pkg/agent -run '^Test(ListModelsQwen|ReadQwenModels|ListModels)' -count=1 -v`.
2. Run `go test ./pkg/agent -run '^Test(NewReturnsQwenBackend|BuildQwenArgsKeepsProtocolManaged|BuildQwenArgsYoloAlwaysPresent|QwenCode020FixtureParses|QwenCode020ErrorFixturePreservesErrorMessage)$' -count=1 -v`.
3. With Qwen configured locally, run `$env:MULTICA_RUN_REAL_AGENT_SMOKE='1'; go test -tags=agentintegration ./pkg/agent -run '^TestQwenRealModelCatalogSmoke$' -count=1 -v`.
4. Run `go vet ./pkg/agent` and `go build -buildvcs=false ./...` from `server/`.

## Checklist

Risk: duplicate IDs configured against different base URLs cannot be disambiguated by Qwen's headless `--model` flag. The catalog therefore exposes each ID once and uses the active route's label when available.

## AI Disclosure

**Prompt / approach:** Reproduced the empty catalog, traced Qwen selection through Multica and Qwen Code 0.23.0, rejected ACP-only route IDs as incompatible with headless execution, and implemented settings-based discovery with unit and opt-in real-account validation.

## Screenshots (optional)

The reported before-state shows an online Qwen runtime with an empty model picker. The runtime UI can be verified after this daemon change reaches a Desktop release.

### DIFF
--- a/server/pkg/agent/models.go
+++ b/server/pkg/agent/models.go
@@ -5,11 +5,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
@@ -265,10 +267,13 @@ func ListModels(ctx context.Context, providerType string, runtimeCmd Command) (C
 			return discoverCodebuddyModels(ctx, runtimeCmd)
 		})
 	case "qwen":
-		// Qwen Code has no account-independent headless model catalog. An
-		// empty list keeps the runtime default and manual model entry available
-		// without advertising a Token-Plan-specific model to other accounts.
-		return Catalog{Models: []Model{}}, nil
+		// Qwen Code keeps its account-scoped headless model catalog in the user
+		// settings file. Read only the model provider definitions instead of
+		// publishing a static Token-Plan-specific list that would be wrong for
+		// other accounts.
+		return cachedDiscovery(discoveryCacheKey(providerType, runtimeCmd), func() (Catalog, error) {
+			return discovered(discoverQwenModels(ctx, runtimeCmd))
+		})
 	case "qwenpaw":
 		// QwenPaw's model selection is unsupported (session/set_model
 		// persists to agent scope, not session scope), so there is no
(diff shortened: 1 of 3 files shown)


multica-eve (changes requested):
This change would let members see and select the models already configured in Qwen instead of relying on the runtime default or manually typing a model name. The problem is real and account-scoped discovery is worth supporting, but the current behavior cannot be accepted until a provider-labeled selection is guaranteed to execute through that same provider route.

Requesting changes for the following blockers:

1. **[P1] Duplicate IDs can display one provider while execution uses another.** `server/pkg/agent/models.go:725-732` replaces the deduplicated row with the route matching the saved `baseUrl`, but `server/pkg/agent/qwen.go:62-66` launches Qwen with only `--model <id>`. In Qwen Code v0.23.0, a model supplied through argv deliberately ignores the persisted `baseUrl` and resolves to the first same-ID entry for the active auth type. The fixture in `server/pkg/agent/models_test.go:53-76` therefore labels and marks the DeepSeek route as default while the actual invocation can use the earlier ModelStudio route. This can silently change credentials, cost, and data destination.

2. **[P1] The catalog is not scoped to Qwen's active authentication protocol.** `server/pkg/agent/models.go:703-724` enumerates every `modelProviders` group, while Qwen resolves models only within `security.auth.selectedType`, including custom mappings from `providerProtocol`. The picker can therefore advertise entries that the current session will route differently or reject. Qwen OAuth is also missed in the other direction: its built-in `coder-model` is not stored in `modelProviders`, so an authenticated OAuth runtime can still get an empty catalog.

3. **[P2] Discovery does not read the effective configuration of the runtime being listed.** `discoverQwenModels` discards the supplied `Command` at `server/pkg/agent/models.go:666` and reads one settings file from the daemon environment. Qwen itself resolves `QWEN_HOME` from `.env`, expands `~`, and merges system, user, and trusted-workspace settings. Custom executable/fixed-argument profiles can therefore receive a cache entry keyed to their command but populated from a different configuration.

Please either carry a stable route identity from discovery through execution, or make the first version fail closed: expose only entries for the active protocol whose IDs are unique and executable with `--model <id>`, while leaving ambiguous/custom-wrapper cases on Default/manual entry. Add contract tests for active auth, `providerProtocol`, duplicate ID/base URL routes, and custom runtime commands; selecting a catalog row should be verified against the identity ultimately used by a fake Qwen invocation.

Validation on head `1c2ad44b6`: `git diff --check`, `go vet ./pkg/agent`, targeted Qwen tests, and an isolated `go test ./pkg/agent -count=1` pass; CI is green. The existing real-account smoke only proves the settings file can be parsed and does not cover the discovery-to-execution contract above.

Issue #8082, opened 2026-09-06
[Bug]: pi resume gate checks the session file but not its recorded cwd — regression from #7760 causes a sub-200ms exit-1 crash loop after worktree GC


libingtong:
### Deployment type

self-host (desktop app + locally running daemon). The defect is in the local daemon binary and is server-independent.

### Area

Agents and runtimes

### What happened?

After the task workdir of an earlier `pi` run is reclaimed, every later dispatch of that same (agent, issue) pair fails in ~165 ms with `exit status 1`, `tools=0`, `output_bytes=0` — permanently, with no self-healing and no useful error surfaced to the platform.

The resume gate and `pi` itself disagree about what makes a session resumable:

- The daemon (since #7760) decides reachability from **whether the session JSONL file exists**.
- `pi` refuses to start based on **whether the `cwd` recorded on the session's first line still exists**.

When the session file survives but its recorded `cwd` has been removed, the daemon logs `resume_reachable=true`, logs `resuming session`, and hands `--session <path>` to a process that cannot use it.

**This is a regression from #7760** (`MUL-6840 fix(pi): preserve sessions across workdir changes`, v0.4.37, merged 2026-08-31). Observed on daemon `0.4.39`.

#7760 deliberately decoupled pi/OMP resume reachability from workdir identity — in `server/internal/daemon/daemon.go`, `gateResumeToReachableSession` now takes

```go
if providerUsesPiSessionFile(provider) {
    reachable = piSessionFilePresent(task.PriorSessionID)
} else {
    reachable = task.PriorWorkDir != "" && sameExistingDir(envWorkDir, task.PriorWorkDir) && sessionHomeReachable
}
```

The PR kept the fallback for a **missing session file**, but nothing checks the session's **recorded cwd**. That is the one thing pi actually validates at startup.

Notably, the doc comment on that same function already describes this exact failure shape as the thing the gate exists to prevent — "the CLI exits within a second and the run fails before doing any work — permanently, because the failed run records no session and the next claim serves the same stale pointer again". #7760 removed that protection for pi/OMP without replacing it with the key pi actually uses.

Before #7760 a changed workdir dropped the session pointer and started a fresh conversation — lossy, but the run worked. After #7760 the run cannot start at all.

### Steps to reproduce

1. Run a `pi` task in worktree A. The session JSONL under `~/.multica/pi-sessions/` records `{"cwd": "<A>", ...}` on its first line.
[...]

PR #8085, opened 2026-09-06
fix(pi): skip resuming sessions with missing recorded cwd

makesomethingshit:
## What does this PR do?

When a Pi session file still exists but its recorded working directory (cwd
in the session header) has been deleted, Pi refuses to start with
`Stored session working directory does not exist`. The daemon previously
treated any present, non-empty session file as resumable, so it kept
handing the stale session path to Pi and hit the same failure on every
start.

This change adds a narrow, Pi-only pre-check: if the session file's
header records an absolute cwd that is confirmed missing (os.IsNotExist),
the gate marks the session unreachable and reuses the existing
fresh-session path. Normal cross-workdir resumes (PR #7760) are
preserved — only a confirmed-missing recorded cwd narrows the verdict.

The new helper reads only the first line of the session file (bounded to
64 KiB), so a long transcript body can never push a healthy header over
the limit. Malformed, empty, relative, or non-string cwd values, and
headers the decoder cannot fully parse, are treated as undecidable and
stay resumable — this check never invents new reasons to discard a
session beyond the one Pi itself refuses to start with.

## Related Issue

Addresses #8082 — startup recovery only.
The separate stderr-diagnostics request is not included.

## Type of Change

## Changes Made

- `server/internal/daemon/daemon.go`: added
  `piSessionRecordedCwdMissing` helper that reads only the first line of
  the session file, decodes the `type`/`cwd` fields, and returns true
  only when the cwd is a confirmed-missing absolute path. Wired into
  `gateResumeToReachableSession` behind `provider == "pi"`. Added
  `reason: "recorded_cwd_missing"` to the existing "dropping prior
  session" log.
- `server/internal/daemon/daemon_test.go`: added `TestGH8082*`
  regression tests — core repro (A), cross-workdir preservation (B),
  header-cwd vs PriorWorkDir (C), undecidable-header tolerance (E),
  format compatibility (F), long-transcript bounded-read guard, and
  OMP/other-provider non-regression (G).
- `TestGH8082GateToBackendHandoff` additionally runs the real production
  `d.runTask(...)` with a fake Pi CLI (extended from the existing
  busy-retry fixture): it observes that the stale session `S` is not
  passed, that the CLI receives `--session <fresh S2>` with the prepared
  cwd, that stdin carries the continuity notice, and that the resulting
  `TaskResult.SessionID` equals the observed fresh session — closing the
  gate → `agent.ExecOptions` → Pi backend execution boundary end-to-end
  (no manual injection of `ResumeSessionID`).

## How to Test

```
cd server
go test ./internal/daemon/ -run 'TestGH8082|TestGatePiResume|TestGateResumeToReachableSession' -count=1 -v
go test ./pkg/agent/ -run 'Pi|Omp' -count=1
go vet ./internal/daemon/ ./pkg/agent/
git diff --check
```

Repro case: create a Pi session file whose header `cwd` points to a
directory, delete that directory, then run the gate with
provider="pi" — reachable is now false and the prior session ID is
cleared, so Pi mints a fresh session instead of refusing to start.

The `TestGH8082GateToBackendHandoff` test additionally drives the real
production `d.runTask(...)` with a fake Pi CLI and asserts the fresh
`--session` argument, the prepared working directory, the continuity
notice on stdin, and the returned `SessionID` — i.e. the full
execution handoff, not just the gate decision.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (N/A)
- [ ] I have updated relevant documentation to reflect my changes (N/A)
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change (N/A)
- [ ] If this PR touches Chinese product copy, I checked it (N/A)

## Risk

- The check reads only the first line of the session file (bounded) and
  never modifies the session file, the current workdir, or GC/lock
  policy.
- Undecidable headers (malformed, empty/relative/non-string cwd, read
  limits) stay resumable — no new forced-reinit paths.
- A race where the recorded cwd is deleted right after the check is not
  addressed atomically (out of scope per issue spec).
- Verified at gate level plus the production `runTask` → Pi backend →
  fake CLI execution boundary (argv, cwd, stdin, returned SessionID).
  Real Pi CLI/model runs and the Unix fake-CLI path were not exercised
  on this Windows runtime (no credentials); `-race` was skipped locally
  (no gcc) and is left to CI.

## AI Disclosure

**AI tool used:** Multica Agent (worker-opencode implementation + Worker-codex
independent cross-validation over multiple review rounds).

**Prompt / approach:** Implemented per the issue spec (design §5, required
regression tests §6). Cross-validation found and fixed: a whole-file bounded
read misclassifying long transcripts, a permissive single-Decode accepting
trailing garbage after the header, and an initial §6H test that bypassed the
production `runTask` glue — the final regression runs the real `d.runTask(...)`
with a fake Pi CLI. Verification evidence: focused tests / go vet /
git diff --check all PASS; `-race` skipped locally (no gcc) but confirmed via
upstream CI (run 34013682159: backend-tests runs `scripts/test-go.sh --race`,
daemon package PASS; 14 SUCCESS / 1 SKIPPED / 0 failed).

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -1,11 +1,13 @@
 package daemon
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"io"
 	"log/slog"
 	"math/rand"
 	"net"
@@ -6125,8 +6127,19 @@ func sameExistingDir(a, b string) bool {
 
 func gateResumeToReachableSession(task *Task, taskCtx *execenv.TaskContextForEnv, provider, envWorkDir string, sessionHomeReachable bool, taskLog *slog.Logger) bool {
 	var reachable bool
+	reason := "session_store_unreachable"
 	if providerUsesPiSessionFile(provider) {
 		reachable = piSessionFilePresent(task.PriorSessionID)
+		// Pi refuses to start when the recorded cwd of the session file it
+		// is given no longer exists (GH #8082), so a present file alone
+		// must not count as resumable. Pi-only: OMP shares the session-file
+		// plumbing but has no such recorded-cwd refusal evidence, and #7760
+		// keeps Pi-family resumes independent of workdir equality — only a
+		// confirmed-missing recorded cwd narrows the verdict here.
+		if reachable && provider == "pi" && piSessionRecordedCwdMissing(task.PriorSessionID) {
+			reachable = false
+			reason = "recorded_cwd_missing"
+		}
 	} else {
 		// Compare the directories, not the spelling. Reuse runs in the canonical
 		// path it validated and locked, which need not be character-identical to
@@ -6138,6 +6151,7 @@ func gateResumeToReachableSession(task *Task, taskCtx *execenv.TaskContextForEnv
 	if !reachable && task.PriorSessionID != "" {
 		taskLog.Info("dropping prior session: session store not reachable from this run",
 			"provider", provider,
+			"reason", reason,
 			"session_id", task.PriorSessionID,
 			"prior_workdir", task.PriorWorkDir,
 			"workdir", envWorkDir,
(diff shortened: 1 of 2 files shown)


Bohan-J:
Thanks for this — and sorry it sat here without a response.

The fix for #8082 has landed in #8097 (merged as 33222ff5), so I'm closing this PR. I want to be straight about the order of events though: **your PR came first.** You opened this on Sep 6, and #8097 was opened about 25 hours later. This sat unreviewed in the meantime and we ended up writing a parallel fix instead of reviewing yours. That's on us, not on you.

Your diagnosis was correct and matched ours independently: the gate treated a present session file as resumable, while Pi refuses to start when the `cwd` recorded in the session header is gone.

## Why the merged version rather than this one

Two things #8097 does that this PR doesn't:

1. **Runtime provenance, not just the provider name.** This PR gates on `provider == "pi"`. A custom runtime profile registers its *protocol family* as the provider, so `protocol_family: pi` with any command also arrives as `"pi"` while being an unrelated implementation. #8097 threads the same provenance the backend gets (`agent.Config.BuiltinRuntime`) so only Pi's own discovered binary is subject to the check. This one was caught in review of #8097 too — the first revision had the same class of bug against OMP.
2. **A backend-side signal.** Pi writes its refusal to stderr and exits before emitting any JSON event, so the daemon only saw `exit status 1` and classified a non-retryable `process_failure`, keeping the same session pointer. #8097 sets `Result.ResumeRejected` on that refusal, which routes into the existing single fresh-session retry. Without it, any refusal the gate fails to predict is still a permanent loop rather than one wasted run. You scoped this out deliberately, which is reasonable — just noting why the merged change is wider.

## One piece of review feedback

Your `piSessionRecordedCwdMissing` drops only on `os.IsNotExist`, treating every other `stat` error as undecidable. That reads as the more conservative choice, but Pi's reference behaviour is Node's `existsSync`, which returns `false` on `EACCES` as well:

```
existsSync(exists-but-EACCES-parent) = false
accessSync error code               = EACCES
```

So when the directory exists but its parent lost traverse permission, Pi still refuses. This PR would keep the session, Pi would refuse, and with no backend fallback that is the same permanent loop the PR is fixing. Narrow, but worth knowing.

## What we're taking from your PR

Two things here are more correct than what we merged, and we'll port them with credit to this PR:

1. **Exact-key JSON decode.** You decode into `map[string]json.RawMessage` and look up `cwd` by exact key. `encoding/json` struct matching is case-insensitive, so the merged code reads `{"type":"session","CWD":"/gone"}` as a recorded cwd, while Pi's `header?.cwd` sees nothing there and starts normally — meaning we'd discard a healthy session. Your version is right.
2. **Relative `cwd` treated as undecidable.** The merged code calls `os.Stat` on it, which resolves against the *daemon's* working directory, whereas Pi's `existsSync` resolves against *Pi's*. Different base, possibly wrong answer. Treating it as undecidable avoids that.

We'd also like to port `TestGH8082GateToBackendHandoff`. Driving the real `runTask` with a fake Pi CLI and asserting the fresh `--session` argument, the prepared cwd, the continuity notice and the returned `SessionID` covers the gate → `ExecOptions` → backend boundary end to end, which our unit-level gate tests don't.

Closing as superseded by #8097, with genuine thanks — the analysis and the test design here were both good, and we'd welcome more contributions from you.

PR #8092, opened 2026-09-06
MUL-7095: fix(issues): keep task detail content stable across re-entry

makesomethingshit:
## What does this PR do?

Refs #8083 — fixes task-detail description flicker/blank re-entry and preserves the first edit or file drop while the deferred editor starts.

The final implementation:

1. Keeps the most recently opened issue detail mounted while the issue list is visible, so list → detail re-entry reuses the populated editor instead of rebuilding it during navigation.
2. Initializes large Markdown documents before the ProseMirror view mounts while keeping Tiptap's deferred `immediatelyRender: false` path.
3. Preserves clicks and file drops that arrive before the deferred editor instance exists.
4. Restores issue scroll only after the description editor is ready, keeps retrying while images are pending, and stops immediately on wheel/touch/pointer/keyboard intent. A hidden retained surface can no longer overwrite a valid offset with zero, and a later platform memento remains authoritative.
5. Makes the navigation A/B runner executable and strict: unambiguous Playwright arguments, zero-exit samples in both modes, sampler-error rejection, uncached builds, and empirically documented 25% relative guardrails for click→commit and maximum overlapping Long Task. The runner is Chromium-only, matching the Chromium-only default project matrix.
6. Defines an explicit zero-baseline policy for the relative guardrail. A baseline with no overlapping Long Task and a head with none is a valid clean result, not an unavailable comparison: `0 → 0` is an available **passing** comparison, `0 → >0` is an available **failing** comparison reported as "baseline was zero, head introduced a non-zero value", and missing/invalid/non-finite data stays unavailable and fail-closed. Reporting and the process exit code consume the same helper result, so the policy is not duplicated between the Markdown and the gate.
7. Defaults `--repeats` to five, matching the sample size the 25% allowance is justified with, so the documented invocation samples five times without extra flags. Explicit `--repeats N` still works.
8. Keeps the default Playwright matrix Chromium-only and runs the MUL-7095 description re-entry acceptance under WebKit through `playwright.webkit.config.ts` and the explicit final step in `scripts/check.sh`. The navigation-trace spec stays Chromium-only because it measures the Long Tasks API, which WebKit does not implement; it asserts support and fails clearly on an unsupported engine instead of recording empty 0 ms samples.
9. Removes revision-only production surface (`eagerClientRender` and `mul7095-*` performance marks/measures).

## Related Issue

Refs #8083

## Type of Change

- [x] Bug fix (non-breaking change)
- [x] Tests / verification tooling

## Key Files

- `apps/web/app/[workspaceSlug]/(dashboard)/issues/layout.tsx` — retains one detail surface across list re-entry.
- `packages/views/editor/content-editor.tsx` — populated-first deferred initialization and startup-input preservation; no eager-render opt-in or production-only instrumentation remains.
- `packages/views/issues/components/issue-detail.tsx` — pre-instance click forwarding and editor-readiness handoff for scroll restoration.
- `packages/views/issues/hooks/use-issue-detail-scroll-restore.ts` — image-driven bounded retry, user-intent cancellation, hidden-surface save guard, and late memento precedence.
- `e2e/description-reentry.spec.ts` — retained-surface, first-edit/drop, geometry, scroll, blank-frame, and ordering coverage; the only description spec run under WebKit.
- `e2e/description-navigation-trace.spec.ts` — Chromium-only navigation/performance trace. It requires Long Tasks API support, records the retained surface's pre-click state, and requires the initialized/populated invariant to be satisfied by a committed post-click sample.
- `scripts/nav-trace-compare.mjs` — clean base/head runner; zero-baseline-aware guardrail reporting and the five-repeat default.
- `scripts/nav-trace-timing.mjs` — sample-validity gate plus the shared `relativeRegression` policy (relative, zero-baseline, unavailable).
- `playwright.webkit.config.ts` / `scripts/check.sh` — scoped WebKit gate without doubling the canonical suite.

## Verification

Final local head: `41e822b68` (runner-policy revision on top of `affd270dc`; no product, editor, or navigation code changed in this revision).

- Navigation-runner unit tests: **11/11 passed** (`node --test scripts/nav-trace-timing.test.mjs`), including the six new zero-baseline/relative cases.
- Views focused suites: 4 files / 144 tests passed (`content-editor`, `content-editor.ready`, `issue-detail`, `use-issue-detail-scroll-restore`).
- Web focused suite: 1 file / 1 test passed (issues `layout.test.tsx`).
- Typecheck for the affected workspaces: `turbo typecheck --filter=@multica/views --filter=@multica/web` → 6/6 tasks passed.
- `git diff --check` passed. No changed file is covered by the repo's ESLint configs (the revision touches only `scripts/*.mjs`, which have no ESLint project).
- Browser evidence gathered for the accepted revision still applies unchanged: Chromium re-entry + startup 6/6, scoped WebKit gate (description re-entry only) 2/2, WebKit first-edit + file-drop 10/10, and the Long-Tasks guard failing loudly on WebKit instead of reporting empty samples.

### Zero-baseline policy

`relativeRegression(baseValue, headValue, allowance)` now returns one of three explicit modes:

| Case | Result | Gate |
| --- | --- | --- |
| base `0`, head `0` | `mode: "zero-baseline"`, `available: true`, `ratio: null`, `failed: false` | **pass** |
| base `0`, head `> 0` | `mode: "zero-baseline"`, `available: true`, `ratio: null`, `failed: true` | fail, reported as *baseline was zero, head introduced a non-zero value* |
| base `> 0` | `mode: "relative"`, `failed: ratio > 1.25` | existing 25% relative rule (exact `1.25x` passes) |
| missing / non-finite | `mode: "unavailable"`, `available: false` | fail-closed |

No absolute millisecond fallback was introduced, so no second machine-time policy needs its own evidence. The generated report says `zero baseline; head remained zero — passed` or `zero baseline; head introduced a non-zero value — FAILED` instead of `not evaluated`, while invalid measurements still report as unavailable and still fail the gate. **`0 → 0` is now a valid, available, passing zero-baseline comparison.**

### Clean navigation A/B

Official runner, Chromium, **default repeat count (no `--repeats` flag → five repeats, recorded as `repeats: 5` in both reports)**, uncached production builds, live backend. The base side runs in `collect` mode, the head in `accept` mode.

| Metric | base `2ae2dbbb8` | head `41e822b68` | change |
| --- | ---: | ---: | ---: |
| Click → first committed detail frame | 57.5 ms | 48.4 ms | -15.8% (ratio `0.842`, limit `1.25`, passed) |
| Maximum overlapping Long Task | 63.6 ms | 0.0 ms | ratio `0.000`, passed |
| Click → first populated | 147.6 ms | 48.4 ms | derived — same commit frame on the head |
| Total overlapping Long Task | 63.6 ms | 0.0 ms | same window caveat as above |

- Base: **5/5 usable**; head: **5/5 usable** — all ten scenario runs exited `0`.
- Both builds were real `--force` builds: `build_cached: false` on both sides.
- Base max-overlap per repeat: 62, 67, 65, 60, 64 ms. Head max-overlap per repeat: 0, 0, 0, 0, 0 ms. On this machine the base's post-commit rebuild produces a clipped Long Task just above the 50 ms threshold in every repeat, so the base-vs-head comparison exercises the normal relative branch.

### Zero-baseline result (the reported blocker)

Because this machine never produces a `0` baseline from the base side, the reported failure mode was reproduced and then cleared with the official command by comparing the clean revision against itself (same command, same defaults, uncached builds, five repeats):

- Base `5/5` usable, head `5/5` usable, all scenario exits `0`, `build_cached: false`.
- `maxOverlapMs`: base `0.0`, head `0.0` → report line: `maxOverlapMs relative guardrail: 25.0% (zero baseline; head remained zero — passed)`.
- `clickToCommitMs`: ratio `0.980` (passed).
- Command exit code: `0`.

Before this revision that identical clean result was reported as `not evaluated` and failed the gate with exit `1`, which was the merge blocker.

## Confirmed Fixed

- Re-entry reuses a populated detail surface without a blank frame.
- First click and file drop survive deferred startup on Chromium and WebKit.
- Scroll restoration no longer fights trusted user input or loses the previous offset when the retained surface is hidden.
- The click→commit performance guardrail passes for the measured implementation (`0.842`), and re-entry no longer blanks the description.
- The A/B gate now passes a valid `0 → 0` Long Task result instead of treating a zero baseline as an unavailable comparison.
- The documented invocation samples five repeats by default, matching the evidence behind the 25% allowance.
- WebKit coverage is scoped to `description-reentry.spec.ts`; the Long Tasks navigation trace is Chromium-only and fails clearly when the entry type is unavailable instead of reporting empty samples.
- Revision-only eager-render and performance-mark paths are removed.

## Remaining Scope / Risks

- Only the most recent detail surface is retained; opening another issue replaces it.
- Native Markdown images without intrinsic dimensions can still shift layout while decoding. Restoration is held through editor readiness and pending image load, but this PR does not claim every image-heavy case in #8083 is resolved.
- The scroll retry remains bounded to 30 animation frames and yields immediately to user intent.
- Playwright WebKit is not a substitute for a real macOS Safari run.

## AI Disclosure

**AI tool used:** Multica Agent — muse-spark-1.3, GPT-5.6 Luna/Sol, deepseek-4.1-flash


multica-eve (changes requested):
When someone returns to a task they just viewed, the description currently disappears briefly before coming back. This PR aims to keep that known content visible until it can be edited again, which is the right user outcome. However, the current handoff can replace the blank flash with a position jump, and the visible description temporarily ignores the user's first attempt to edit or attach a file, so this version is not ready to merge.

## Phase A — Product decision

- **Before → After:** the description changes from briefly disappearing on re-entry to remaining visible while the page becomes editable.
- **Journey:** this affects anyone moving quickly from task detail to the list and back, especially on long or image-heavy descriptions in Web and Desktop.
- **Scope:** no migration, setting, or relearning is required; the behavior changes by default for every task detail view.
- **Decision:** 🤔 **The product direction is correct, but the interaction design still needs a maintainer decision.** We need to choose whether descriptions remain immediately editable on open or stay stable in reading mode until the first editing intent. Either choice must avoid both a visible jump and a temporarily inert description.

## Phase B — Implementation review

### Must fix

1. **The new handoff reintroduces a layout shift that the existing design explicitly avoids, while the tests cannot validate the visual requirement.**

   `packages/views/issues/components/issue-detail.tsx:2018-2022` documents that swapping a long description between `ReadonlyContent` and ProseMirror accumulates per-block height differences into a visible scroll/layout jump. The new path at `issue-detail.tsx:3061-3070` performs that exact swap.

   `packages/views/editor/content-editor.ready.test.tsx:185-275` verifies nodes, text, and image `src`; `packages/views/issues/components/issue-detail.test.tsx:867-920` verifies that JSDOM's `scrollTop` property was not reassigned. JSDOM does not calculate layout, paint, or image decoding, and the PR explicitly says browser validation was skipped. These tests therefore prove the wiring, but not that the reported flicker is gone rather than replaced by another one.

   Before merge, please validate Chrome and Safari with short, long, mixed-block, and image descriptions, covering same-task re-entry and A→B switching, and provide evidence that there is no blank frame or geometry jump. Prefer a browser-level regression. If the handoff still shifts, keep one rendering surface continuously responsible for the visible description instead of swapping automatically.

2. **The visible fallback drops the first edit or file-drop intent before the editor is ready.**

   The fallback at `issue-detail.tsx:3061-3070` has no pointer/keyboard activation path. The drop zone at `issue-detail.tsx:2038-2040` forwards files directly to the ref, while `packages/views/editor/content-editor.tsx:934-937` returns when the editor does not exist yet. During the documented 70–460 ms editor setup window, a user can click content that looks ready or drop a file into the now-visible target, but nothing is focused or queued.

   `packages/views/editor/use-lazy-editor.ts:52-71` already owns this focus-transfer and upload-queue behavior. If the fallback design remains, please reuse that controller (including `resetKey: id`) or provide equivalent intent capture and replay, with tests for the first click and file drop before readiness.

### Please address with the revision

- `packages/views/editor/content-editor.tsx:560,597,691-697` unconditionally adds a readiness state update to every `ContentEditor`, including callers without `onReady`. Since Tiptap emits `create` from a timer, this adds a separate render across a shared hot path. Restrict that bookkeeping to consumers that need readiness, or signal readiness without global component state.
- The same short/long/mixed/image matrix is repeated in both the real editor suite and the mocked `IssueDetail` suite. Keep the behavior matrix in one canonical layer, and leave the host suite with one wiring case plus one cross-issue reset case; use the saved coverage budget for real browser layout and interaction checks.

If this change is wrong, all Web and Desktop task-detail users can see description/image movement or lose their first edit action; reverting this PR restores the previous behavior but also restores the original blank flash. The targeted tests, attachment regressions, typecheck, lint, diff check, and current CI pass, but they do not cover the two browser interaction blockers above.

**Recommendation: Request changes.** Please resolve both must-fix items and add real-browser evidence before re-review.

makesomethingshit:
@multica-eve Revision pushed in faeb200 for another review.

The description now retains one populated ProseMirror surface from its first client paint. The readonly handoff is removed; chunked initial parsing happens before view construction, and readiness no longer adds an unconditional state/render. The PR body includes the browser reproduction and validation results: 149 focused/related unit tests, 3 Chromium regressions, and the same 3 regressions in Windows Playwright WebKit, plus typecheck/lint.

Two description-only presentation changes are explicitly documented for review: permanent 16:9 image preview frames using `object-fit: contain` (full-image viewing preserved), and system monospace code typography to avoid WebKit's delayed font-metric shift.

**Actual macOS Safari remains untested because no macOS/Safari environment is available. Windows WebKit is not represented as Safari validation.** This is a re-review request, not a claim of complete acceptance or merge readiness.

makesomethingshit:
Please wait a moment. I am making additional revisions.
Sry :(

makesomethingshit:
I have finished the revisions. Please review them. @multica-eve

multica-eve (changes requested):
Request changes: the single-editor revision fixes the previous renderer handoff, but it now blocks task-detail navigation while Tiptap is constructed synchronously, and the Safari/WebKit first-input acceptance is still not reliable.

## Product decision

The intended before → after remains correct: a cached task description should stay visible, editable, and at its restored position when the user returns to the task. The new revision does remove the readonly/ProseMirror swap and its inert first-click/drop window, so the earlier two design blockers were addressed rather than papered over.

However, the current trade replaces a visible blank with a much longer frozen navigation for large descriptions. That is the same core journey—rapid list/detail movement—and is a worse default for every Web and Desktop task detail. 🤔 **Revise before merge.** No migration or compatibility issue is involved, but the latency and browser-interaction regressions below are must-fix.

## Must fix

1. **Do not build the task-description editor synchronously in the route render.**

   `packages/views/editor/content-editor.tsx:570-583` passes `immediatelyRender: true` for this host and performs the large-document parse in `onBeforeCreate`; `packages/views/issues/components/issue-detail.tsx:3044` enables it for every task description. Tiptap constructs its schema, document, state, and ProseMirror view in that render. The repository's existing measurement at `packages/views/editor/use-lazy-editor.ts:52-56` already records 70–460 ms per editor and explains why editor construction was kept off the navigation-critical path.

   I A/B-tested the current head and its merge base with the same real Web/API stack, isolated database, Chromium, and the PR's 100-section fixture. After discarding each run's first cold sample, click → first committed detail frame was **482–806 ms on `84f43d5` versus 84–114 ms on base**; the head produced a **514 ms** maximum Long Task versus 148 ms on base. This is not covered by the submitted regression: `e2e/description-reentry.spec.ts:102-129` only appends a frame after `[data-testid="issue-description"]` exists, so it cannot observe the frozen interval before the first detail commit.

   Please retain the blank-free/single-surface outcome without constructing ProseMirror inside the route render (for example, preserve or prepare the populated surface outside that critical render), and add an A/B browser gate that starts at the issue-link click and records first committed frame plus Long Tasks. The new path should not regress the base navigation budget.

2. **Make the reported WebKit/Safari interaction path reproducible and stable.**

   `playwright.config.ts` still defines only Chromium, so the documented WebKit command is not runnable as submitted (`--browser=webkit` is rejected when configured projects exist). With a temporary WebKit project, the stability/navigation scenario passed, but the new “first edit and file drop” case lost `FIRSTEDIT` while retaining `first-drop.txt` in **2 of 8 executions** (failure at `e2e/description-reentry.spec.ts:270`). This also occurred on a warm repeated run, contrary to the PR body's 2/2 warm claim.

   Please commit a reproducible WebKit project and make this acceptance deterministic. Instrument whether the heading click actually establishes DOM focus before `insertText`; if it does not, fix the first-click path, and if it does, fix the document mutation being lost. Then run the same test repeatedly on WebKit. Playwright WebKit is not literal macOS Safari, but it is the minimum executable coverage for a report reproduced in Safari.

## Acceptance scope and cleanup

- `e2e/description-reentry.spec.ts:63-79` deliberately excludes every pre-decode image frame from geometry assertions. The PR body itself reports the anchor moving from −87.8 px to 218.2 px—about **306 px**—before image decode, while #8083 says image-heavy descriptions make the problem most visible. If this PR remains a partial `Refs #8083` fix, rename/narrow the geometry claim and keep the image behavior explicitly outstanding. If it is meant to resolve #8083, the assertion must cover and bound first-visible-frame → settled image movement.
- Remove `.github/pr-evidence/*.png` from the product repository. These are PR-only evidence binaries, and two still frames cannot demonstrate a transition; attach the evidence to the PR/run instead.

## What passed

- The previous readonly handoff and readiness-only render were removed; lifecycle ordering and canonical test ownership now look sound.
- Focused Vitest: 5 files / 127 tests passed.
- Chromium regression: 2/2 passed.
- Typecheck: 9/9 tasks; changed-file ESLint and `git diff --check`: passed.
- Current GitHub checks are green.

The performance A/B and WebKit failures are the remaining merge blockers; the image statement and PR-only artifacts also need to be made accurate/clean in the revision.

multica-eve (changes requested):
Request changes: the deferred, populated-first editor mount is the right product direction and it fixes the previous eager-render regression, but this head still cannot produce its required performance evidence and introduces two broad behavioral risks that should not ship as-is.

The user-facing change is worthwhile: reopening a long issue should show populated content on the first editor frame, while clicks and drops during deferred startup are preserved. I verified the focused Chromium/WebKit flows locally, including 10 consecutive WebKit first-edit runs. The remaining blockers are:

1. `scripts/nav-trace-compare.mjs:227-237` cannot run the advertised A/B gate. It invokes Playwright as `--project chromium e2e/description-navigation-trace.spec.ts`; with the repository's Playwright 1.58.2, the spec is consumed as a second project name. A clean 3-repeat run at `60ecf3637` therefore produced `Project(s) "e2e/description-navigation-trace.spec.ts" not found` for every base and head repeat, leaving both sides at 0/3 usable samples. Please use an unambiguous argument form (for example `--project=${project}` or put the spec before the option) and commit a successful clean A/B result. While fixing this, do not let `collect` mode accept every non-zero scenario exit (`:262-280`, `:425-434`): because the spec already skips the blank-frame assertion in collect mode, an exit must still be zero or an explicitly identified expected failure; otherwise teardown/worker/assertion failures after report creation become usable samples. The default 25% regression allowance also needs an empirical basis or a project-owned policy rather than an unexplained one-repeat threshold.

2. `playwright.config.ts:16-30` adds WebKit to the global project matrix. Contrary to the comment, project order does not preserve the old default: bare `playwright test` runs both projects. `scripts/check.sh:135` uses exactly that bare command, so this doubles every canonical E2E spec and makes the suite require WebKit, while the checked-in workflow only installs Chromium. Scope WebKit to the description startup specs (or a separate config/command) and wire that command into the intended gate.

3. `packages/views/issues/hooks/use-issue-detail-scroll-restore.ts:83-115` now rewrites `scrollTop` for eight stable frames. Any user scroll during that roughly 133 ms window is indistinguishable from layout drift and is snapped back to the stored target; the new unit expectation actually locks in that overwrite. Please stop restoration on trusted wheel/touch/pointer/keyboard intent, or drive the retry from the image/layout event being handled. Also replace the unexplained `8` with evidence tied to the real late-layout boundary.

4. Please remove the revision-only surface that no longer supports the production strategy. `eagerClientRender` (`packages/views/editor/content-editor.tsx:300,463,665`) has no production caller and leaves an opt-in path back to the regression this revision intentionally reverted. The `mul7095-*` production marks/measures (`:105-166`, `:675-760`) are not consumed by the committed runner, reuse global mark names across possible concurrent editors, and clear marks but never measures, so every mount leaves Performance Timeline entries behind. Keep attribution inside the E2E harness or scope/consume/clear it safely.

Validation at this head: focused Vitest 144/144; navigation/re-entry Chromium+WebKit 6/6; WebKit first-edit 10/10; script unit tests 3/3; typecheck, changed-file ESLint, and `git diff --check` passed. The new A/B gate failed deterministically as described above. The image decode shift remains a documented partial scope, so this PR should continue to use `Refs #8083` rather than claim the entire upstream issue is fixed.

multica-eve (changes requested):
Request changes: the retained-detail implementation is now a sound product fix and the four previous blockers are addressed, but the checked-in A/B gate still fails on a valid, fully clean result when the baseline has no Long Tasks.

Product judgment: this change should exist. On a warm list → issue re-entry, the latest detail surface and its populated editor remain mounted but hidden, so the first visible commit contains the content instead of rebuilding it after navigation. Retaining exactly one surface is bounded, needs no user migration, and matches the reported workflow. The scoped WebKit gate, user-intent cancellation during scroll restoration, and removal of revision-only editor APIs also resolve the prior design risks.

One merge blocker remains in `scripts/nav-trace-compare.mjs:316-327,397-409`. `relativeRegression()` returns `available: false` whenever the base value is `0`, and `regressionFailed` treats every unavailable result as a failure. On this exact head I ran the official command with `--repeats 5` and uncached production builds. Both base and head produced 5/5 usable, zero-exit samples; click→commit improved from 86.0 ms to 32.9 ms (ratio 0.382), and both sides had `maxOverlapMs = 0`. The runner nevertheless exited non-zero because the max-overlap guardrail was “not evaluated.” A machine fast enough to produce no ≥50 ms Long Task therefore cannot pass the gate even when the head is equally clean.

Please define and test the zero-baseline policy. At minimum, base `0` / head `0` should pass; base `0` / head `>0` should follow an explicit fail or absolute-fallback rule. Also align the sampling contract with the threshold evidence: the 25% allowance is justified using five repeats, while `--repeats` still defaults to one and the usage example omits it. Default to five or reject undersampled gate runs so the documented policy is the behavior users get.

Validation on `affd270dc`: views focused tests 145/145; web layout test 1/1; runner unit tests 5/5; Chromium navigation/re-entry 3/3; scoped WebKit 2/2; WebKit first-edit/drop 10/10; full monorepo typecheck, changed-file ESLint, diff check, and uncached base/head production builds passed. The official A/B run itself ended non-zero only for the zero-baseline bug above. The remaining native image decode shift is still honestly scoped under `Refs #8083`.

PR #8103, opened 2026-09-07
feat(projects): add markdown notepads to projects

### LINKED ISSUE #8094: [Feature]: Markdown notepads scoped to a project
### Deployment type

self-host

### Area

### What do you want and why?

A project today has issues, resources and a description, but nowhere to keep the
free-form working notes a team accumulates around it: meeting takeaways, a running
decision log, links that are not formal resources, the "why did we do it this way"
that never fits in an issue.

Today that content ends up in one of three places, all of them lossy:

- The project description, which is a single field everyone overwrites.
- An issue that is not really an issue, so it distorts the board and never closes.
- An external doc, which drops out of the workspace entirely — agents cannot see
  it, and it rots.

The last point is what makes this more than a convenience. An agent working on a
project can read the project's issues and resources, but a decision recorded in an
external doc is invisible to it. Notes that live in the workspace are notes an
agent can be told to read and append to.

Markdown notepads scoped to a project:

- Multiple named notes per project, each an ordinary markdown document.
- A section on the project detail page to read and edit them (no new route; it
  is inherited by web and desktop through project-detail.tsx).
- CLI access (`multica project-note list|get|create|update|append|delete`) so an
  agent can maintain a running log. `append` is a distinct verb precisely so a
  long-running agent does not have to read-modify-write and race itself.
- The agent brief carries the instructions for those commands, not the note
  contents, so adding notes does not inflate every prompt.

I have this implemented and running in a downstream fork, and can open a PR
against `main` if the shape sounds right. Happy to adjust the surface — in
particular whether this warrants its own table or should fold into
`project_resource`, which is the closest existing concept.

Image

sunyang-16:
## What does this PR do?

Adds markdown notepads scoped to a project.

A project has issues, resources and a description, but nowhere to keep the free-form working notes a team accumulates around it: meeting takeaways, a running decision log, the "why did we do it this way" that never fits in an issue. Today that lands in the project description (one field, everyone overwrites), in an issue that is not an issue (distorts the board, never closes), or in an external doc — which drops out of the workspace entirely, so an agent working the project cannot see it.

That last point is why this is more than a convenience. An agent can read a project's issues and resources; a decision recorded in an external doc is invisible to it. Notes that live in the workspace are notes an agent can be told to read and append to.

**On the design question from the issue** — whether this should fold into `project_resource` instead of getting its own table: `project_resource` models a *pointer* to something (url, kind, position), while a note *is* the content, with body/body_size/append semantics and its own size budget. Happy to revisit if you would rather see them unified.

## Related Issue

Closes #8094

## Type of Change

## Changes Made

**Data layer**
- `server/migrations/456_project_note.{up,down}.sql` — the table, no foreign keys per the repository rule.
- `457_*`, `458_*` — both indexes `CREATE INDEX CONCURRENTLY`, each alone in its own migration, both registered in `concurrentIndexCleanups` (`server/cmd/migrate/main.go`).
- `server/pkg/db/queries/project_note.sql` + generated code (sqlc v1.31.1 via `make sqlc`).
- `workspace_delete.sql` — a `deleted_project_notes` arm in the teardown CTE. With no FK, teardown must delete notes explicitly; `project_note` is classified `workspaceDelete` in the manifest test, matching its sibling `project_resource`.

**Backend**
- `server/internal/handler/project_note.go` + tests — CRUD plus a distinct `append` endpoint, so a long-running agent can extend a log without a read-modify-write race against itself.
- `server/cmd/server/router.go` — routes under `/api/projects/{id}/notes`.
- `server/pkg/protocol/events.go` — `project_note:created|updated|deleted`.
- `execenv/runtime_config_sections.go` — the brief carries a fixed-size instruction block naming the commands. Note **contents are deliberately excluded**, so adding notes does not inflate every prompt.

**CLI** — `multica project note list|get|create|update|append|delete`.

`project resource` uses `add`/`remove` because those are *association* verbs; a notepad is an owned entity, so `create`/`delete`. The rest follow this repo's dominant convention (`get` 18, `update` 28, `delete` 14 occurrences across `cmd/multica`).

**Frontend**
- `packages/views/projects/components/project-notes-section.tsx` — a section on the project detail page. No new route: web and desktop both inherit it through `project-detail.tsx`.
- `packages/core/projects/note-queries.ts`, types, and 6 API methods in `client.ts`, each through `parseWithFallback` with a zod schema.
- All four locales, 20 keys each.

**Docs** — folded into `builtin_skills/multica-platform/references/projects.md` as a `## Notes` section rather than a new skill directory, following #7968.

Three details that are easy to get wrong:
- `body_size` uses `octet_length` / Go `len()`, not a UTF-16 length, so CJK sizes are correct.
- The title limit counts runes, so 200 characters means 200 characters in any script.
- `ListProjectNotes` projects `octet_length` instead of `SELECT *`, so listing does not pull every body just to discard it.

## How to Test

1. `make sqlc` — should produce no diff.
2. `cd server && go build ./... && go vet ./...`
3. `go test ./internal/handler/ -run 'ProjectNote|RemovesNotes'` — the second pattern matters: `TestDeleteProjectRemovesNotes` and `TestDeleteWorkspaceRemovesNotes` do not match `ProjectNote`, and they are the two covering teardown. Needs a migrated database — `internal/handler`'s `TestMain` exits 0 when Postgres is unreachable, so check the RUN count rather than trusting `ok`.
4. `go test ./cmd/migrate/ ./cmd/multica/`
5. `pnpm --filter @multica/core test` and `pnpm --filter @multica/views exec vitest run projects/`
6. Manually: add a note in the project's Notes section, then `multica project note list <project>` and `multica project note append <project> <note> --body "..."`; confirm both surfaces agree.

## Checklist

- [x] If I added a new runtime / coding tool / UI tab, I synced landing copy and docs — n/a: a section inside an existing page, not a new tab or runtime
- [x] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` — zh-Hans follows §3 punctuation (`...` not `……`); 项目 matches the glossary

**Verification actually run:** `go build ./...` and `go vet ./...` clean; `gofmt` clean on every file in this commit; `make sqlc` reproduces the generated code with no diff; the note suite is 15 top-level RUNs / 22 with subtests, all passing, against a database confirmed at `schema_migrations` max `453_project_note_workspace_index`; `cmd/migrate` and `cmd/multica` green; the CLI verb naming verified by running the built binary, not just by compiling; migrations checked up/down on a real Postgres 16 with `pg_constraint contype='f'` count 0 and both indexes `indisvalid`; frontend `@multica/core` 1722 and `@multica/views` `projects/` 51 passing, `tsc --noEmit` clean.

**Two gaps I am flagging rather than hiding:**
1. `appendProjectNote` and title-rename exist in the client and hooks, but no component calls them yet — the UI covers list/read/create/replace/delete. The CLI and API cover all six, which is what the agent path needs. Happy to wire the UI in this PR if you would rather it ship complete.
2. `project-notes-section.tsx` has no direct test file, and `project-detail.test.tsx` stubs it to `null` — the established pattern there, matching `ProjectResourcesSection`. Per the testing table in CLAUDE.md a shared view component should have one; say the word and I will add it here.

**One pre-existing failure, not from this change:** `TestMigrateHermesTaskMemoriesFailureKeepsSource` in `internal/daemon/execenv` chmods a file unreadable and asserts the read fails, but as uid 0 the read succeeds. I confirmed it fails identically at this commit's parent with `execenv/` checked out from clean `main`. Two files also carry pre-existing `gofmt` violations I left alone: `internal/handler/issue_table_query.go` and `internal/service/empty_claim_cache.go`.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** The feature was built and has been running in a downstream fork; porting it upstream was done with Claude Code. The approach was to inventory the feature's true boundary first — including the wiring scattered through existing files, which a file list alone misses — then port it against *this* repository's rules rather than copying the fork verbatim. That meant renumbering migrations, dropping the fork's foreign keys in favour of explicit application-layer teardown, folding the skill docs into `multica-platform` instead of adding a skill directory, renaming the CLI verbs to this repo's dominant convention, and replacing a duplicate byte formatter with the `formatBytes` already in that package.

Every claim in the verification section was re-run and checked independently rather than taken from a summary. That is how the pre-existing `execenv` failure was confirmed (reproduced on clean `main`) and how a mismatch between the documented CLI verbs and the actual ones was caught before this PR was opened.

Porting from the fork's current `master` rather than the original commit was deliberate: the introducing commit had since been fixed several times there (a UTF-16 `body_size` bug, a byte-vs-rune title limit, a `SELECT *` listing), and two of its queries had already been deleted as dead. Those fixes are folded in here.


sunyang-16:
First-time contributor here, so both workflows are sitting at action_required. Whenever someone has a moment to approve them, I'd appreciate it — happy to rebase if main has moved on by then. Everything I could verify locally is in the description.

Issue #8095, opened 2026-09-07
[Feature]: Add contextual DingTalk replies and task lifecycle reactions


yyclaw:
## Deployment type

Not established for a specific instance; this feature request is not limited to a deployment type.

## Area

Chat integration — DingTalk (community-maintained)

## What do you want and why?

This proposal addresses four gaps in the upstream DingTalk integration:

- **Processing feedback adds chat noise.** Acknowledgements are separate messages rather than status attached to the request. They do not show whether that request is still processing or has finished.
- **Group replies lack clear attribution.** Without a visible quote of the triggering message or a native sender mention, answers can be difficult to follow in a busy conversation.
- **Quoting a bot answer can lose context.** The outgoing Markdown title contains only a heading or generic preview, so a subsequent quoted reply may omit the answer text needed to continue the discussion.
- **Command results are less useful than they could be.** Created or duplicate issues are shown as plain identifiers, and command confirmations do not share a consistent reply format.

## Proposed solution (optional)

- **Use reactions for task status.** Replace separate acknowledgements with a processing reaction on the source message, in groups and direct conversations. Clear it when processing ends; add Done only after the final answer is successfully delivered. Handle cancellation, failure, retries and late callbacks without marking another request as finished.
- **Make group replies contextual.** Visually quote the current user message and use the callback's session webhook to mention its sender. Encrypt persisted webhook credentials. When the webhook is unavailable, retain conversation delivery without the mention. After sending a request, switch transports only on an explicit rejection; an ambiguous result must not trigger a duplicate send.
- **Preserve complete answer context.** Keep each answer chunk's original Markdown in its title, within the serialized message-size budget, so a subsequent quote retains the answer text, code and links. Keep the displayed source-message excerpt separately bounded, and preserve supported selected RichText, literal separators and visible HTML source without broadening parsing of the current user input.
- **Unify command feedback.** Apply the same group reply context to command results, remove decorative success checkmarks from generated group confirmations, and link created or duplicate issues to their workspace-scoped pages. Keep plain identifiers when a link cannot be built.

Reply quotes, mentions and Done must refer to messages actually included in the task. The shared engine and task service should record that association without changing existing batching, media waiting, `/new`, `/clear` or retry isolation. Other channels retain their existing behavior. Persisted webhook credentials must be encrypted and cleaned up after expiry; this includes index migrations to support bounded cleanup.

[...]

PR #8073, opened 2026-09-05
MUL-7085: fix(chat): publish cancellation events when archiving agents

yyclaw:
## What does this PR do?

Archiving an agent cancels its chat tasks but skips `task:cancelled`, leaving existing chat lifecycle consumers uninformed. For example, Slack keeps an already-added processing reaction after the task is cancelled.

Publish the existing event for each chat task actually cancelled by archive, after the cancellation and recovery receipt settlement transaction commits. This repairs the producer shared by Slack/Lark cleanup, Telegram stream cleanup and web/desktop/mobile chat cancellation handlers. Non-chat tasks still receive no per-task archive event, and `agent:archived` remains unchanged.

The approach follows the existing cancellation contract: #6631 added the channel subscribers and identified archive's missing broadcast as a follow-up; #7820 established the atomic cancellation/settlement wrapper. Keeping that wrapper and reusing the existing event producer addresses the missing trigger without a new cancellation path or protocol.

Reported deployment: Not established (source-level report; verified with local automated tests).

Code applicability: Deployment-independent (Official App and self-host deployments running the affected backend version). The archive route and cancellation service have no deployment-mode gate. Which affected versions are currently deployed has not been verified.

## Related Issue

Closes #8072

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Refactor / code improvement (no behavior change)
- [x] Documentation update
- [x] Tests (adding or improving test coverage)
- [ ] CI / infrastructure

## Changes Made

- `server/internal/service/task.go`: publish chat-only cancellation events after the existing transaction succeeds; propagate errors before any event is emitted.
- `server/internal/service/task_archive_cancel_test.go`: cover all five active statuses, committed state visible from another connection, event scope/payload, non-chat exclusion, terminal/other-agent preservation, repeat cancellation, and rollback without publication when commit fails.
- `server/internal/integrations/slack/typing_indicator_archive_test.go`: verify archive cancellation removes the reaction through the real service, bus subscriber and Slack SDK, using a local HTTP endpoint and a real database.
- Update the explanatory comments in `server/internal/handler/agent.go`, `server/internal/integrations/slack/typing_indicator.go` and `server/internal/integrations/lark/outbound.go`; their executable code is unchanged.

No API fields, event shapes, schema, migrations or client implementations change. Installed clients can use their existing `task:cancelled` handlers. WeCom retains its existing no-reply behavior on cancellation.

[...]

PR #8125, opened 2026-09-07
MUL-7106 feat(dingtalk): add contextual replies and task lifecycle reactions

yyclaw:
## What does this PR do?

DingTalk group replies gain a source quote, and processing acknowledgements use message reactions instead of separate chat messages. Command confirmations preserve the user's current command and link created or duplicate issues to their workspace pages.

The implementation follows the review's narrower design: quote text comes from task-owned sealed input; reaction coordinates are captured during local ingest. Shared channel routing and persistence remain unchanged.

## Related Issue

Related to #8095. Native sender mentions and full-answer Markdown titles remain separate followups.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- **Group reply quotes** (`outbound.go`, `outbound_quote.go`, `markdown.go`): quote the last channel-ingested message in the task's sealed input. Internal attachment images in that quote become `[Image]`; stored input and the agent's answer are not rewritten. Bound previews to 256 UTF-8 bytes while preserving line breaks and complete URLs. If sealed input cannot be read, send the answer without attribution. Escape block markers only at the start of source lines; ordinary punctuation stays literal. Preserve URL bytes and hard line breaks, including the earlier fixes for escaped URL underscores and parentheses.
- **Selected rich-text quotes** (`inbound.go`): recognize ordered snapshot nodes with supported `text`/`content` wrappers and `msgType` aliases, preserving neighboring text and picture order. This fixes supported quoted nodes appearing as `[rich-text content unavailable]`; arbitrary nested objects are not interpreted as prose or commands.
- **Processing and completion reactions** (`ack.go`, `ack_batch.go`, `emotion.go`, `reply_source.go`, `outbound.go`): use existing input task/context ownership and input ordering to keep the processing reaction on the last message in a batch. Task completion, failure and cancellation retire their owned inputs. Agent archive also immediately retires locally registered receipts before a task exists. In-flight adds are recalled after returning if their state was retired. Add `Done` only after every final-answer chunk is delivered, using the locally captured provider anchor. Gate terminal reaction reads on local session interest, registered before input commit and retained through source capture and ingest, so unrelated channels incur no reaction queries without losing protection against late hooks. Reuse successful task and sealed-input reads from reply delivery; retry failed reads in the cleanup context.
- **Command feedback** (`replier.go`, `resolvers.go`): capture the current command before routing rewrites it; retain success checkmarks in both group and direct confirmations. Link issue identifiers when the app URL, workspace slug and issue ID are usable, and otherwise keep plain identifiers. Removing the decorative group checkmark belongs with the native-mention followup.
- **Delivery and coverage** (`outbound_send.go`, `server/cmd/server/router.go`, tests): keep serialized Markdown payloads within 15,000 bytes while preserving answer chunks and code fences; wire the same client and existing queries into the reaction lifecycle. Cover input ordering, sealing races, archive, lookup failures, invalid ownership, partial delivery, source-cache isolation, and real router/database integration. Retain the four DingTalk guide updates.

Paths above are under `server/internal/integrations/dingtalk/` unless stated otherwise. The incoming HTML/source-text preservation and full-answer title work remain in the title followup; webhook-based native mentions remain in the mention followup. No shared engine, service, SQL, schema, persisted callback credentials, or reconciliation worker is added here.

### Known limitations

- Task quotes use sealed input content, which can include the user's quoted history; this does not restore a strict current-turn-only preview.
- The shared failed-enqueue hook supplies only a session ID. An older failed flush can still clear a newer batch's processing reaction, and a hook that starts after that cleanup can re-add an old reaction. This differs from the input-scoped task-terminal cleanup above.
- Offline and archived notices are sent without a quote, including live requests, because the result does not distinguish a recovered generation from the current callback. This avoids attributing an older request to a later message.
- Reactions remain best-effort and process-local. Restarts, events handled in another process, failed provider calls, or hooks not registered before archive can leave receipts behind. Source and settled-input caches are bounded to 1,024 entries; missing or evicted anchors suppress `Done`. Active receipts have no independent expiry/reconciler and can accumulate when terminal events never arrive.

These limitations are disclosed rather than addressed by expanding the shared engine. The Slack/Lark implementation comparison confirms the session-wide and late-add cleanup limitations there as well; it is not a claim of exact per-message or restart-safe cleanup.

## How to Test

From `server/`, with an isolated, migrated PostgreSQL test database configured:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -p 1 -count=1 \
  ./internal/integrations/dingtalk ./cmd/server
```

- Race tests: 715 DingTalk and 473 server tests/subtests passed, with no failures or skips.
- Named behavior verification: 319 tests/subtests passed, covering unrelated-channel query counts, terminal-before-ingest races, source eviction, reference cleanup, successful/empty input-read reuse, and failed-read retries.
- `go build ./...`, `go vet ./internal/integrations/dingtalk ./cmd/server`, and `git diff --check` passed.
- Quote regression tests cover literal punctuation, line-leading markers, URL underscores and parentheses, preserved line breaks, and partial-URL truncation. These verify outbound payloads and local Markdown rendering; they do not certify the DingTalk client renderer.
- Manual smoke steps: send quoted text/images in a group, exercise `/new`, `/clear` and `/issue`, and verify quotes and workspace links. Send multiple inputs in one batch and check the last-message receipt; archive before enqueue, cancel a task, and simulate a failed later answer chunk to verify that it never earns `Done`.

I tested the three quote-rendering cases in a real DingTalk client and confirmed they pass: inline punctuation and URL preservation, line-leading markers and line breaks, and truncated source quotes. Screenshots are attached below. The test counts above refer to local automated tests; the screenshots do not cover the full manual lifecycle checklist.

## Checklist

- [x] I have included real DingTalk client screenshots of the updated behavior

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Used Codex to implement and test the DingTalk-scoped reply and reaction behavior, compare the split work with the original combined implementation, and check the Slack/Lark lifecycle model suggested in review. Shared reliability changes were kept out; the remaining differences and local-test limits are documented above.

### Inline punctuation and URL preservation

image

### Line-leading markers and line breaks

image

### Truncated source quote

image


Bohan-J (changes requested):
Thanks for the thorough write-up — the direction is right. The separate "👀 On it" message really is noise in a busy group, and an answer with no visible link back to the question it answers really is hard to follow. The prerequisite PRs (#8061, #8059, #8073) are all merged, so the dependency story checks out.

I can't approve this as it stands. There is one issue that can silently drop a user's reply entirely, a few smaller correctness problems, and four decisions that a maintainer needs to make before this shape gets locked in.

Verified locally: `go build ./...` and `go vet` over `internal/integrations/dingtalk/...`, `internal/integrations/channel/engine/...` and `internal/service/...` all pass; CI is green.

---

## Blocking

### 1. The Markdown title has no length cap, and it roughly doubles every message payload

`markdownTitle` changed from "first ATX heading" to "every readable block of this chunk", and nothing truncates the result:

```go
chunk.title = markdownTitle(answer)   // up to ~16000 bytes, never truncated
...
param, _ := json.Marshal(markdownParam{Title: title, Text: chunk.withoutMention()})
```

The comment at the top of `markdown.go` is explicit about why the budget exists: DingTalk hard-rejects a `sampleMarkdown` payload over ~20000 bytes and **drops the whole message**, which is why the body budget sits at 16000 and even subtracts the 4 synthetic bytes of `"\n```"`. What actually goes on the wire is `msgParam` — the JSON of `{title, text}` — and this change takes it from ~16000 to ~32000 without recomputing anything.

Concrete failure: an agent returns a 12000-byte answer (one moderately long review summary). Title 12000 + text 12000 ≈ 24000 bytes of `msgParam`, over the limit, DingTalk rejects it, **the user gets nothing**, and the only trace is a `warn` line server-side. Short answers keep working, so this will most likely be reported as "the bot stopped replying" and misattributed to the agent.

Even if the limit turns out to apply only to `text`, the `title` field has its own limit — and the PR description says as much: "long-title acceptance limits ... remain unverified".

`outbound_preview_test.go:69` (`TestSender_CompleteQuoteTextFollowsAnswerChunks`) asserts that a 9000-byte answer chunk produces a 9000-byte title, so the current test suite actively pins this behavior in place.

Please give the title an explicit byte budget with a stated basis, count it against the overall `msgParam` budget, and update that test accordingly.

### 2. A cosmetic capability can reject an inbound user message

`dingtalkSessionRouting` went from infallible to returning an error, and that error propagates all the way up through `StartSession` / `AppendMessage`:

```go
cfg.SessionWebhookEncrypted, err = encryptToken(strings.TrimSpace(raw.SessionWebhook), encrypt)
if err != nil {
    return "", nil, fmt.Errorf("encrypt session webhook: %w", err)
}
```

`AppendMessage` fails → `processClaimed` fails → `Handle` returns an error → the connector NACKs → DingTalk redelivers → it fails deterministically again. Failing to encrypt a short-lived credential used only for a decorative mention should not be able to keep the user's message out of the system.

In production `box.Seal` only fails on an RNG failure, so this is unlikely to fire today. The direction is what I'd like changed: the quote and the mention are presentation, so they should degrade (send without the mention) rather than block ingestion. Note also that `NewDingTalkResolverSet`'s `encrypt` parameter is nillable (`channel_title_adapters_db_test.go:72` passes nil) — a second production call site that forgets it would silently take down group message handling.

Nice touch, though: `cmd/server/dingtalk_wiring_test.go` asserting via AST that argument 6 is `box.Seal` is exactly the right guard against that.

### 3. `attempts[0]` is indexed without a length check

```go
input := attempts[0]
```

The comment ("The SQL VALUES guard returns one row even when the input or owner is gone") is accurate today — `FROM (VALUES (1)) guard(n) LEFT JOIN attempts t ON true` does guarantee a row. But that is an implicit contract spanning SQL and Go, and breaking it produces a panic rather than an error. `if len(attempts) == 0 { return true, nil }` costs nothing.

### 4. `has_final_failure_outcome` correlates rows by timestamp equality

```sql
CASE WHEN t.status = 'failed' THEN EXISTS (
    SELECT 1 FROM chat_message outcome
    WHERE outcome.chat_session_id = t.chat_session_id
      AND outcome.created_at = t.completed_at
      AND outcome.role = 'assistant' AND outcome.task_id = t.id
) ELSE false END
```

`outcome.task_id = t.id` already identifies the row. The extra `created_at = completed_at` relies on an invariant that is declared and guarded nowhere: `FailTask` writes `completed_at = now()` and the chat message in one transaction, so the Postgres transaction timestamps happen to match. Switch either side to `clock_timestamp()`, or move that message into a different transaction, and the predicate is permanently false with no test to catch it.

More immediately: on the sweeper path it is **already** always false. `HandleFailedTasks` — the funnel for `FailStaleTasks`, `FailTasksForOfflineRuntimes` and recover-orphans — never writes that assistant message; only `FailTask` does. So for any sweeper-failed task, "the final failure outcome has landed" always reads as false, and `ChannelReactionSettled` keeps reporting the input as unsettled while `retryEligible` holds, leaving the 收到 reaction pinned.

In practice `OnDeliverySettled` (the terminal-event `defer`) usually covers it — but covering the case where the terminal event fires on a *different replica* is the entire reason this query exists, and that is exactly where it doesn't work.

A stable predicate such as `outcome.task_id = t.id AND outcome.role = 'assistant' AND outcome.failure_reason IS NOT NULL` would avoid the coupling. If you do need to distinguish the final failure output from other assistant messages on the same task, please document the invariant on `FailTask` and add a test that guards it.

### 5. `retryEligible` is called with a hand-assembled partial struct

```go
retryEligible(attempt.FailureReason.String, db.AgentTaskQueue{
    ChatSessionID: attempt.ChatSessionID, AutopilotRunID: attempt.AutopilotRunID,
    Attempt: attempt.Attempt, MaxAttempts: attempt.MaxAttempts,
})
```

I checked `task.go:5262`: for chat tasks this is currently correct — a valid `ChatSessionID` short-circuits the OR, and an empty `Context` makes `isSourceContextQuickCreateTask` return false early. But `retryEligible` is shared between `FailTask`'s in-transaction retry and the orphan sweeper, and the moment it starts reading a fifth field this call misjudges silently: no compile error, no failing test, just a stuck reaction. Please either note the partial-struct caller on `retryEligible`, or select the full set of columns it reads.

---

## Design questions for the maintainers

These are judgement calls I don't think a reviewer should make unilaterally, and they shape the code, so I'd like answers before this merges.

**Q1 — Is putting the whole answer in the push preview the behavior we want?** It turns every mobile push notification into the full reply text. It also contradicts a recorded note: `markdown.go` on `main` says "DingTalk shows the title only in the push preview, **not in the message body**", while this PR asserts "Observed DingTalk text quote callbacks carry this title as their body". One of those is wrong, and the PR doesn't reconcile them or show the observation.

**Q2 — Is permanent retention of the session webhook and the verbatim user text acceptable?** `dingtalkBindingConfig` now carries `SessionWebhookEncrypted` and `QuoteText`; both land in `channel_chat_session_binding.config` and are copied into every `channel_task_delivery.config` row, which is an append-only per-task snapshot. That means (a) encrypted bearer URLs accumulate with no cleanup path and become key-rotation baggage long after they expire, (b) user message text now exists in N copies outside `chat_message`, untouched by `/clear`, `/new` or session deletion, and (c) `QuoteText` is unbounded, so one long message inflates the binding config and every delivery snapshot. If the answer is "acceptable", `QuoteText` should at least get a length cap and the credential retention story should be written down.

**Q3 — May a community-maintained channel push channel-specific logic into shared core?** Concretely: `channel_type = 'dingtalk'` hardcoded in the shared `channel.sql`, a `DingTalkReactions` field on the shared `handler.Handler`, and a `runDingTalkReactions` goroutine in `main.go` (we already have `ChannelSupervisor` as the shared channel-worker abstraction). Related: `ContextTypingNotifier`, `MessageTypingSettler` and `ExactChannelTaskEnqueuer` are three new optional interfaces discovered by runtime type assertion, and the comment on the last one says the seam exists for "test and platform enqueuers". `CLAUDE.md` is explicit here — "For internal, non-boundary code, do not add compatibility layers, fallback paths, dual writes, legacy adapters, or temporary shims unless explicitly requested" and "Prefer existing patterns/components over new parallel abstractions". Beyond style, it has a real consequence: production uses `*service.TaskService` and takes the `EnqueueChannelChatTaskWithReplyTarget` branch, while a test double that doesn't implement the interface takes the other one — so the shared-path regression tests may not cover what production actually runs.

**Q4 — Should all four features ship together?** Reaction lifecycle, contextual replies, readable previews and command feedback carry very different risk. The reaction lifecycle is close to purely additive; the readable preview (Q1) and the webhook persistence (Q2) are the parts that need a decision and live verification. Splitting into 2–3 PRs would let each be validated and rolled back independently, and would make the 579-line `ack.go` much easier to review on its own.

---

## Nits

- `outbound_send.go:270` — `escapeMarkdownText` unconditionally escapes `-`, `+`, `#`, `|`, `!` and friends. If DingTalk's Markdown subset doesn't honor backslash escapes (quite likely, given it is a restricted subset), every hyphen in a quoted message renders as `\-` and the user sees their own words peppered with backslashes. This lands squarely in the "actual DingTalk rendering ... remain unverified" area. Consider escaping only the leading characters that would break block structure.
- `outbound_send.go:207` — `strings.ContainsRune("-*+", rune(line[0]))` treats a byte as a rune; a multi-byte UTF-8 leading character yields a meaningless rune. It happens not to misfire today (the result is always false), but `strings.HasPrefix` or decoding the first rune would say what's meant.
- `ack.go` — 579 lines replacing 136, with one mutex plus a per-state `cleanupMu`, a chain of `ready` channels, three TTL'd fence maps, a DB reader and a reconcile worker. I traced it and did not find a deadlock or a data race (removal from `active` and the `ready` write both happen under `n.mu`, which gives the needed happens-before), and `-race` passes. The complexity itself is the risk though: three separate fence maps (`preSettled`, `preSettledAgents`, `settledBatches`) suggest the race between "ingest detected" and "settlement detected" was fenced off incrementally rather than designed out.
- `moveAcknowledgement` chains every anchor move, and each link is "wait for the previous link → recall → add", i.e. two HTTP calls. Five messages in quick succession means the fifth 收到 only appears after four serialized round trips. Correct, but possibly a visible lag in a busy group — worth confirming that's acceptable.

## Things worth keeping

- Moving `CreateChannelTaskDeliveryFromSession` from before to after the input seal is necessary and correct — the new CTE reads `chat_message.task_id`, which the seal writes. I checked that `SetChatTaskInputOwnerSelf` and `SealChannelChatTaskInput` don't read `channel_task_delivery`, so the reorder is safe within the transaction.
- With `override_reply_target = false` the priority is always 1, which preserves every other channel's existing behavior exactly. Backward compatibility is handled cleanly, including legacy configs without `source_message_id`.
- `sessionWebhookRejected` distinguishing an explicit rejection from an ambiguous outcome, and choosing not to switch transports on ambiguity so a reply can never be duplicated, is the best-argued part of this PR.
- `postSessionWebhook` refusing to put provider text or the bearer URL into any loggable error — including the `url.Error` path — is careful work.
- The AST-based wiring test pinning `box.Seal`.

---

One more thing: the PR states the live smoke test wasn't run, and a meaningful share of this change's correctness rests on actual DingTalk behavior — title length acceptance, what a quote callback carries, and Markdown escaping. Before merge I'd like either a run against a real DingTalk workspace, or the parts that depend on unverified provider behavior split out.

Happy to re-review as soon as #1 is addressed.

yyclaw:
@Bohan-J I pushed the review follow-up in a5a90c118 and updated this PR and #8095.

- **Payload size and quoting:** Every chunk is checked against a 15,000-byte serialized payload budget, including title, text and webhook mention metadata, before sending starts. The title retains the chunk's complete Markdown for subsequent quotes; source-message previews have a separate 256-byte cap.
- **Input and terminal handling:** Missing or failed webhook encryption preserves ingestion and falls back to proactive delivery. Empty attempt results return an error, and final-failure detection uses the task-bound failure marker instead of timestamp equality. The partial retry projection is covered against the current policy.
- **Retention and shared enqueueing:** Expired webhook credentials are cleaned up in bounded batches, supported by two concurrent partial indexes. Per-message reply configuration requires the exact-source capability; missing support rejects the enqueue instead of silently dropping the source metadata.
- **Rendering:** Added regressions for UTF-8 preview boundaries, long answer blockquotes, mention placement, table delimiters and issue-link fallback.

The four requested behaviors and the DingTalk quoted-RichText repair remain together.

The 195 targeted regressions, focused package race runs, frontend parity cases and `go vet` passed. Live DingTalk end-to-end validation has also been performed on the current version. Full validation details are recorded in the PR description.

This is ready for another review.

Bohan-J (changes requested):
Thanks for the fast and complete follow-up in a5a90c1. Every item flagged as blocking in the previous review is fixed, and fixed carefully: the serialized 15 KB budget, the best-effort webhook encryption, the failure-marker predicate, the guard-row check and the `retryEligible` note are all what I asked for. Build, vet and CI are green on my side too.

This review is about the open maintainer questions from last time (Q3 and Q4), which I can now answer. The answer is that the PR needs to change shape, not just details, so I am requesting changes again. I want to be clear that this is a scope decision on our side, not a verdict on the quality of the work.

## Decision

**The shared core stays channel-agnostic.** DingTalk is a community-maintained channel, and the bar for touching `channel/engine`, `service`, `channel.sql`, `main.go` or migrations is "another channel needs the same thing", not "this channel wants to be more precise". Nothing in this PR meets that bar, so I am not accepting:

- the three runtime-asserted optional interfaces (`ContextTypingNotifier`, `MessageTypingSettler`, `ExactChannelTaskEnqueuer`) and the batcher batch/sequence identity;
- the reply-source priority CTE in `CreateChannelTaskDeliveryFromSession` and the `ChannelReplyTarget` enqueue path;
- `GetChannelReactionInputAttempts`, `ChannelReactionSettled`, the `DingTalkReactions` field on `Handler`, and the two DingTalk workers started from `main.go`;
- the two `channel_type = 'dingtalk'` partial indexes, the cleanup queries, and persisting encrypted session webhooks in binding and delivery config.

**The precedent to follow already exists in the repo.** Slack and Feishu ship exactly this feature today: `server/internal/integrations/slack/typing_indicator.go` and `server/internal/integrations/lark/typing_indicator.go` add a reaction on ingest and remove it on `EventChatDone`, `EventTaskFailed` and `EventTaskCancelled`. Each is under 300 lines, lives entirely in its own package, keeps state in memory, and is best-effort by design. Their comments spell out the accepted gaps (add/clear race, process-local bus, restart) and we live with them. DingTalk should get the same treatment, and `dingtalk/emotion.go` is already the right primitive for it.

## What I would like the PR to become

Keep, inside `server/internal/integrations/dingtalk/` only:

1. **Reaction lifecycle.** Replace the "On it" text send in `ack.go` with `emotion/reply`, subscribe to the three terminal events and recall it, swap to Done only after the final answer chunks are accepted. Model it on the Slack manager. Session-granularity clearing is fine; a debounced batch does not need per-message exactness.
2. **Quoted source in group replies.** The `channel_task_delivery` snapshot already freezes the trigger message ID and binding config at enqueue, and the task's sealed `chat_message` rows already hold the user's text. Read the quote from there at send time instead of persisting a `QuoteText` copy. A bounded blockquote is enough for attribution.
3. **Command feedback and issue links** in `replier.go`, which is self-contained and fine as is.

Move to separate PRs, each with its own evidence:

- **Native `@` via the session webhook.** This single feature is what forces credential persistence, expiry cleanup, indexes and a worker. The webhook also expires within hours, so any run longer than that falls back to proactive send and loses the mention anyway. If we want it, it needs a design that does not persist the bearer URL at all, or a product decision that the retention is acceptable. Please open that as its own PR after this one lands.
- **Full answer text in the Markdown title.** This is a product-visible behavior (push preview becomes the whole answer) that hinges on the quote-callback observation. It needs redacted real callback and client evidence, and it is independent of everything above.

Expected size is a few hundred lines of production code, a small fraction of the current diff. Tests should shrink proportionally; the ones for `emotion.go`, the reply layout and the replier links are worth keeping.

If it helps land something sooner, the reaction lifecycle alone is a clean first PR and I will review it quickly. Happy to discuss any of this in the thread if you see a constraint I am missing.

yyclaw:
@Bohan-J — could you verify one narrow point in the proposed quoted-source simplification?

**[SRC-001] Freezing the delivery snapshot does not, by itself, prove that its source message belongs to the task's sealed input.** I reproduced a generation-1 task whose input contains only message A, while its delivery snapshot names generation-2 message B. This uses the ordinary `EnqueueChannelChatTask` path. The current PR's guarded `EnqueueChannelChatTaskWithReplyTarget` path prevents this mismatch; I am not reporting it as still broken in that DingTalk path.

The distinction is visible in the code:

- The [base snapshot query](https://github.com/multica-ai/multica/blob/9fab6da917953c6f1a967a17beef045aff3d0ce5/server/pkg/db/queries/channel.sql#L822-L835) copies `binding.last_message_id`, `last_thread_id`, and `config` at enqueue time.
- At [the reviewed head](https://github.com/multica-ai/multica/blob/a5a90c1185416c35b69e817f532bf32c1f093892/server/internal/service/task.go#L1875-L1890), ordinary enqueue passes no reply target. Consequently, [the query](https://github.com/multica-ai/multica/blob/a5a90c1185416c35b69e817f532bf32c1f093892/server/pkg/db/queries/channel.sql#L836-L865) takes priority 1 via `NOT override_reply_target`, retaining the base's source-selection behavior.

The sequence to check is:

1. A is persisted in context revision 1; that revision has not enqueued yet.
2. A `/clear` boundary and a newer input B advance the same session to revision 2 and update its binding's latest source to B.
3. The pending revision-1 work enqueues through the ordinary path. Input sealing selects A by revision, but the delivery copies the binding's latest message ID, B.

The test below materializes the database state after step 2, then calls the production enqueue service and generated SQL. It is a deterministic database-level reproduction, not a live DingTalk or full Router/callback replay. Please also check whether this pending-generation sequence matches the intended `/clear`/debounce contract.

**Reproduce on `a5a90c1185416c35b69e817f532bf32c1f093892`:** in an isolated checkout with a migrated test PostgreSQL database, append this function to `server/internal/service/task_channel_reply_target_test.go`. It reuses that file's existing imports and `newChannelReplyFixture`, which creates A as `source-1` in revision 1. Set `DATABASE_URL` to the test database; the fixture creates and cleans its own rows.

```go
func TestReview8125LegacySnapshotCanNameAnotherGeneration(t *testing.T) {
	f := newChannelReplyFixture(t)
	ctx := context.Background()
	f.InsertNoID(t, "channel_chat_context_generation", dbfx.Cols{
		"chat_session_id": f.session.ID, "revision": 2, "pending_fresh": true,
		"initiator_user_id": f.UserID,
	}, "chat_session_id = $1 AND revision = $2", f.session.ID, 2)
	second := util.MustParseUUID(f.Insert(t, "chat_message", dbfx.Cols{
		"chat_session_id": f.session.ID, "role": "user", "content": "second generation input",
		"channel_ingested": true, "channel_context_revision": 2,
	}))
	f.Exec(t, "UPDATE channel_chat_session_binding SET context_revision=2,last_message_id='source-2' WHERE id=$1", f.binding.ID)
	task, err := f.service.EnqueueChannelChatTask(ctx, f.session, util.MustParseUUID(f.UserID), false, 1, f.binding.ID, f.binding.RouteRevision)
	if err != nil {
		t.Fatal(err)
	}
	delivery, err := f.service.Queries.GetChannelTaskDelivery(ctx, task.ID)
	if err != nil {
		t.Fatal(err)
	}
	input, err := f.service.Queries.ListChatInputMessages(ctx, task.ChatInputTaskID)
	if err != nil {
		t.Fatal(err)
	}
	if len(input) != 1 || input[0].ID != f.message || input[0].ID == second {
		t.Fatalf("unexpected sealed input: %+v", input)
	}
	if delivery.ChannelMessageID.String != "source-2" {
		t.Fatalf("expected latest binding snapshot: %+v", delivery)
	}
	t.Log("CONFIRMED: generation-1 task owns only source-1 input; immutable delivery names generation-2 source-2")
}
```

Run from `server/`:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -count=1 \
  ./internal/service \
  -run '^TestReview8125LegacySnapshotCanNameAnotherGeneration$' -v
```

Observed output:

```text
CONFIRMED: generation-1 task owns only source-1 input; immutable delivery names generation-2 source-2
--- PASS: TestReview8125LegacySnapshotCanNameAnotherGeneration
```

This is a **characterization test**: `PASS` means it reproduced the mismatch, not that the desired source-ownership invariant holds. A database-unavailable `SKIP` does not verify it. I reran it with `-race`, along with `TestEnqueueChannelChatTaskWithReplyTarget_ConcurrentGenerationsKeepOwnDeliveryAndInput`; both passed without skips.

The scope of the result is source association in the ordinary enqueue path. I have not established a user-visible Slack/Feishu incident from it. Your suggestion to read the quote from sealed task input, using `chat_input_task_id` (including retry ownership), may avoid depending on these delivery coordinates. This observation alone does not establish that the current shared CTE/interface must remain. Could you verify the reproduction and confirm whether that distinction changes how we should derive attribution in the smaller PR?

Bohan-J:
Thanks for [SRC-001] — that's a good catch, and it's worth separating two things: whether the observation is correct (it is), and whose problem it is (not yours).

## The observation reproduces

I verified it structurally rather than by running your test — reading the query and its call sites is enough to confirm the mechanism:

- `CreateChannelTaskDeliveryFromSession` on `main` copies `binding.last_message_id` / `last_thread_id` / `config` as they stand at enqueue time, with nothing tying them to the task's sealed input.
- It has exactly one production caller: `enqueueChatTaskTx` in `server/internal/service/task.go`, on the shared channel enqueue path.

So yes — a pending generation that enqueues after a `/clear` boundary can end up with a delivery snapshot naming a message the task never owned. Your reading of the priority-1 branch is right too.

## It is not a DingTalk problem, so please don't carry it here

`/clear` and `/new` live in `channel/engine/fresh_command.go`, and every channel reads those snapshot fields back out for outbound routing:

- `slack/outbound.go:192` — `LastMessageID: delivery.ChannelMessageID, LastThreadID: delivery.ChannelThreadID`
- `lark/outbound.go:355` — same
- `wecom/outbound.go:312` — same
- `telegram/outbound.go:1075` — same

That means the mismatch you found can put a reply on the wrong thread anchor on Slack, Lark, WeCom and Telegram as well, entirely independent of DingTalk. This is a core defect on `main` and it is ours to investigate and fix; we'll track it separately. Please don't treat it as scope for this PR, and please don't feel obliged to fix it — you've done the useful part by finding and characterising it.

## To your actual question

> Could you verify the reproduction and confirm whether that distinction changes how we should derive attribution in the smaller PR?

Yes, it changes it — thank you for pushing on this. The suggestion to anchor the reaction on `channel_task_delivery.channel_message_id` was wrong, for exactly the reason you give. Concretely, for the smaller PR:

- **Quote text** — read it from the sealed input (`chat_input_task_id` → the owned `chat_message` rows). That part of the earlier suggestion stands, and it removes the need for a persisted `QuoteText` copy.
- **Reaction anchor** — do *not* derive it from the delivery snapshot, and note it can't come from sealed input either: `chat_message` has no provider message-id column (the table carries `channel_ingested` and `channel_context_revision`, but never the platform `msgId`). So the anchor should be the one the notifier captured in-process at ingest, cleared on the terminal event — the Slack/Lark typing-indicator shape. It is best-effort and lost across a restart, which is the same guarantee Slack and Lark already accept and document.

That last point is the crux of the disagreement, so let me state it plainly: the reason the current design needs `ChannelReplyTarget`, the priority CTE and the engine interfaces is that it aims for a *stronger* guarantee than any other channel provides — exact per-message cleanup that survives replica changes and restarts. We don't think a reaction should buy that guarantee at the cost of channel-specific logic in shared code. If the cleanup is best-effort, SRC-001 stops mattering for attribution and the shared changes disappear with it.

So the requested shape is unchanged: reaction lifecycle, group quote block and command issue links, contained in the `dingtalk` package, with the native mention and the full-answer title as separate PRs. And as noted before — if you'd rather land just the reaction lifecycle first as a small PR, that's welcome and we'll review it quickly.

yyclaw:
@Bohan-J I've narrowed this PR to the shape you requested: group quote blocks, best-effort reactions, and command issue links, contained in the DingTalk adapter with its server wiring. The shared engine/query changes and migrations are gone; SRC-001 is left to the separate core fix.

Task replies now take their quote from sealed input, so the displayed excerpt can include history the user quoted. Reaction anchors are captured at ingest and kept only in-process. This accepts the restart/replica limitations you described, without adding durable recovery. `Done` still requires successful delivery of the complete answer; losing the local anchor simply skips that reaction.

Native mentions and full-answer titles are separate followups. I've also moved the incoming quote-parser extensions out with the title work, so this PR preserves the existing input parsing behavior.

The PR description now reflects these boundaries. Could you take another look at whether the remaining scope matches what you had in mind?

yyclaw:
@Bohan-J This PR covers DingTalk group reply quotes, best-effort lifecycle reactions, and command issue links. It follows the adapter-local approach you suggested: quotes come from sealed input, and reaction anchors are captured in process at ingest. There are no shared engine or schema changes; SRC-001 remains separate core work.

The quote handling includes a focused parser fix to retain supported rich-text text/image order and renders internal images as `[Image]`. That parser fix belongs here because quoted replies depend on it; broader HTML/source-text preservation stays with the full-answer-title followup. Native mentions also remain separate, including removal of the success checkmark from group command confirmations.

Reactions stay on the last message in a batch, and archiving clears receipts already tracked locally, including before enqueue. The remaining boundary is best effort: session-scoped failed-enqueue cleanup can affect an overlapping batch or race a late ingest hook. Task quotes can include quoted history, and offline/archive notices omit quotes because their source cannot be reliably attributed. Handling that attribution precisely appears to require changes to the shared engine to expose reliable source information; do you see a better way to do this within the current adapter-local scope? Local race-enabled tests pass; live DingTalk client verification has not been repeated.

Could you take another look at this scope, particularly the focused quote-parser fix and the cleanup/attribution limits above?

Bohan-J (changes requested):
This is the shape we asked for, and it is close to mergeable. Thank you for the rewrite: 43 files and +3704/−249 against the previous 70 files and +9451/−407, with every production change inside `server/internal/integrations/dingtalk/` and a two-line wiring change in `cmd/server/router.go`. No engine interfaces, no service method, no shared SQL, no migrations, no worker, no persisted webhook. The known-limitations section is honest and matches what Slack and Lark already accept.

Verified locally on `9f1cabfa`: `go build ./...`, `go vet` over the DingTalk package and `cmd/server`, `go test -race` on both, and `git diff --check` all pass. CI is green.

Checked against the earlier request:

- Reaction lifecycle in the Slack/Lark shape, anchor captured in-process at ingest, retired on task terminal events by sealed input, `Done` only after every answer chunk is accepted: yes.
- Group quote read from sealed input via `ListChatInputMessages`, 256-byte preview, no persisted copy: yes.
- Issue links in command feedback: yes.
- Native mention and full-answer title split out: yes, and `markdownTitle` is back to "first heading of the answer".

I am requesting changes for one small fix and one verification. Once those land I will approve.

## Blocking

### 1. Reaction settle runs two queries for every chat task on every channel (`outbound.go:119`)

`processEvent` registers the `settleTaskReactions` defer at line 119, before the `GetChannelTaskDelivery` lookup at 129 and the `ChannelType != dingtalk` check at 136. The defer runs on every return path, so whenever DingTalk is configured, every chat-task terminal event in the deployment, Slack, Lark, WeCom, Telegram and plain web chats included, now executes `GetAgentTask` and `ListChatInputMessages` (lines 199 and 210), finds no local reaction state for that session, and returns.

That is the same "DingTalk imposes cost on everyone" concern from the previous rounds, moved from shared SQL to query volume. Reaction state is process-local by design, so the gate can be free: check whether `ackNotifier.active` holds any state for this session before touching the database, and return immediately when it does not. That is a handful of lines and it restores the pre-PR behavior of one cheap lookup per non-DingTalk event.

### 2. Quote-block escaping has not been seen on a real client (`outbound_send.go:160`)

The PR states no live DingTalk smoke test was run for this revision. The quote block is rendered on every group reply, and `escapeMarkdownText` backslash-escapes `-`, `#`, `+`, `!`, `>` and `|` in the user's text. DingTalk's Markdown is a restricted subset; if it does not honor backslash escapes, every hyphen in a quoted message renders as `\-`. Please attach one screenshot of a real group reply quoting a message that contains `-` and `#`, or narrow the escaping to the line-leading characters that can actually break the blockquote structure.

## Nits, not blocking

- `ackNotifier.active` (`ack.go:40`) has no upper bound, while `settledInputs` and the reply-source cache are both capped at `maxReplySources`. In a multi-replica deployment, terminal events that land in another process leave entries here forever, each carrying an installation config. Slack has the same property, but since the 1024 cap already exists, applying it here is cheap.
- Keeping the reaction on the last message of a batch costs at least two `GetChatMessage` reads per ingest plus the CAS loop in `ack_batch.go` (lines 46, 53, 67). Slack reacts on every message and clears them all at the end, with no reads. The current behavior is visually cleaner and acceptable; if you ever want to cut further, this is the next place.
- `ListChatInputMessages` runs twice per DingTalk terminal event, once for the quote at line 154 and once in settle at line 210. Worth sharing the result.
- The quoted rich-text parser fix is scoped to the `quoted` flag and does not touch current-message parsing. Fine to keep here.

## Your two questions

- Offline and archived notices without a quote: keep it that way. Those notices are not answers to a specific message, so omitting the quote is correct, and it is not worth opening the engine for.
- Session-scoped failed-enqueue cleanup that can touch an overlapping batch: accepted. Slack clears at session granularity too, and DingTalk does not need a stronger guarantee.

yyclaw:
I addressed the two blocking points in this revision:

- Terminal reaction cleanup now checks local session interest before reading the task or sealed input. I register that interest before input commit and retain it through source capture and ingest, so the gate also handles a terminal event that arrives before the reaction hook. Tests cover unrelated Slack, Lark, WeCom, Telegram and web sessions, late hooks, source eviction, and failed input writes.
- I narrowed quote escaping as suggested: ordinary `-`, `#`, `+`, `!`, `>` and `|` stay literal, with block-marker protection limited to line starts. URL bytes, unescaped parentheses and hard line breaks are preserved. Payload and local Markdown tests pass; I also tested these cases in a real DingTalk client and confirmed they pass. Three screenshots are attached in the PR description: inline punctuation and URL preservation, line-leading markers and line breaks, and truncated source quotes.

For the nits, I reused successful task and sealed-input reads between reply delivery and reaction settlement, including successful empty results. Failed reads can still retry in the cleanup context. The last-message-of-batch behavior is unchanged.

I also checked Slack and Lark and reproduced their add-versus-terminal cleanup race. A fixed cap would bound memory, but would not fix missed cleanup and could make workspaces compete for slots. I suggest addressing lifecycle cleanup and capacity together in a separate cross-channel follow-up, keeping this PR scoped to DingTalk. The process-local limitations remain documented.

Race tests passed: 715 DingTalk and 473 server tests/subtests, with no failures or skips. Build and vet also passed. Please take another look.

Bohan-J (approved):
Approving — thank you for the work on this one.

This went through four substantial rounds and you engaged with every piece of feedback rather than defending the original shape. The rework from 70 files / +9451 down to 44 files / +4173, with all production code back inside the `dingtalk` package and the shared surface reduced to two lines of wiring in `router.go`, is exactly what we asked for and it landed cleanly. Two things in particular were above the usual bar:

- The `[SRC-001]` investigation. You found a real pre-existing defect in the shared enqueue path while arguing against one of our suggestions, characterised it precisely, and were careful to state what you had and hadn't established. We've confirmed it and are tracking it on our side — it affects Slack, Lark, WeCom and Telegram equally, so it's ours, not yours, and it is deliberately out of scope here.
- The honesty in the docs and the known-limitations section. Writing down that reaction cleanup is best-effort, process-local and lost across a restart is what makes this mergeable; it sets the guarantee at the same level Slack and Lark already provide instead of implying more.

## One follow-up, deliberately not blocking this PR

We don't want to grow this PR any further, so please treat this as a separate small change whenever convenient.

In `replyMarkdownChunksWithBudget`, the quote prefix and the answer are chunked independently and only merged when the first answer chunk happens to fit alongside the quote. When it doesn't, the quote goes out as its own message and none of the answer messages carry attribution.

We measured the boundary so you don't have to re-derive it — with a short quote (33-byte rendered prefix) and a normal multi-line answer:

| answer size | messages | first message is quote-only |
|---|---|---|
| 13,800 B | 1 | no |
| 14,000 B | 2 | **yes** (33 bytes) |
| 30,000 B | 4 | **yes** |

So it triggers once the answer crosses the 14,000-byte budget with ordinary line structure; an answer made of a few very long lines doesn't trigger it, because the first chunk lands well under the budget.

The fix is to reserve the quote prefix length when chunking the first answer segment, so the first outbound message always carries both the quote and a non-empty piece of the answer; later chunks needn't repeat the quote. `TestSenderQuotedChunksPreserveOneSourceAndWholeAnswer`'s `standalone source` case currently pins the present behaviour, so that assertion would flip.

Worth being clear about the severity: for answers over ~14 KB this doesn't regress anything relative to today's behaviour — the message count is the same as the current separate acknowledgement — it just means the attribution improvement doesn't reach that case. That's why it's a follow-up rather than a blocker.

## On the red CI

The two red backend checks are not yours. The only failing test is `TestMigrationNumericPrefixesAreUnique`, reporting `468_comment_deleted_at` and `468_drop_reference_only_column` sharing a prefix — both existed on the base commit this run was built against, and the collision has since been resolved on `main`. This PR adds no migration files at all, and every other package in that run passed, including `integrations/dingtalk`, `integrations/channel/engine` and `internal/service`. Merging.

Thanks again — this is a good contribution and the DingTalk integration is better for it.

PR #8128, opened 2026-09-07
fix(agent): recover from ghost ACP sessions — dead resume pointers self-heal instead of looping forever

### LINKED ISSUE #8116: Chat conversation permanently stuck: daemon retries a never-persisted session id on session/resume after an early task failure (qoder/ACP)
## Summary

A chat conversation can become **permanently broken** after an early task failure: the daemon keeps trying to `session/resume` a backend session id that was never persisted by the agent CLI, and every subsequent message in that conversation fails with `Invalid session identifier`. The daemon never falls back to creating a fresh session, and there is no CLI command to reset the binding, so the only recovery is to abandon the conversation and start a new one.

## Environment

- Multica daemon: **v0.4.35** (local runtime, macOS)
- Provider: **Qoder CLI 1.1.25** (`/Users/hawk/.qoder/bin/qodercli/qodercli-1.1.25`, driven over ACP)
- Agent model at the time of the first failure: `qmodel_latest` (invalid); later corrected to `qmodel_38max`
- Self-hosted workspace

## What happened

Timeline from `~/.multica/daemon.log` (all in the same chat session, agent bound to provider `qoder`):

**1. First message — task fails on an invalid model id, after the session was created:**

```
16:44:00.553 INF qoder session created component=daemon session_id=27d8031c-9fea-4bda-9d42-37c36fa9aebf
16:44:00.566 WRN qoder set_session_model failed component=daemon
  error="session/set_model: Invalid params: Invalid or unavailable model: qmodel_latest
  (code=-32602, data={\"modelId\":\"qmodel_latest\"})" requested_model=qmodel_latest
16:44:10.567 INF agent finished component=daemon task=… status=failed duration=58s
  agent_error="qoder could not switch to model \"qmodel_latest\": …"
```

The ACP `session/new` succeeded (id `27d8031c…`), but `session/set_model` failed, the daemon aborted the task, and qodercli exited **without ever persisting the session to disk**. Confirmed on the machine: the corresponding project directory under `~/.qoder/projects/` for this task's workdir does not exist.

However, the daemon had already recorded `session_id=27d8031c…` as the session for this conversation (it appears in the failed run's `agent result detail`).

**2. Every subsequent message in the same conversation fails at resume — before the model is even consulted:**

```
16:49:35.921 INF starting agent component=daemon task=… provider=qoder
  model=qmodel_38max reused=true          ← corrected model IS picked up
16:49:35.922 DBG invoking backend component=daemon … resume_session=true
16:49:49.784 DBG agent result detail component=daemon status=failed
  agent_error="qoder session/resume failed: session/resume: Invalid session identifier
  \"27d8031c-9fea-4bda-9d42-37c36fa9aebf\".
  Searched for sessions in /Users/hawk/.qoder/projects/-Users-hawk-multica-workspaces-…-task-…-workdir.
  Use --list-sessions to see available sessions, then use --continue …"
```

Same failure repeats for every new message (observed at 16:45, 16:49, 16:52 — 3-for-3).

Note the corrected model (`qmodel_38max`) is correctly read from agent config on each run — the model fix works, but it can't help because the run dies at the **resume** step, before any model handling.

**3. The agent itself is healthy** — a *new* conversation with the same agent (fresh workdir, `resume_session=false`) succeeds end-to-end:

```
16:50:45.446 INF starting agent … model=qmodel_38max reused=false
16:50:46.801 INF qoder session created …
16:51:06.897 INF qoder finished … status=completed duration=21.446s
```

## Root cause

Two issues compound:

1. The daemon persists/records the backend session id from a run that **failed before producing any output or being persisted by the CLI**. That id is a "ghost": valid at creation time, gone immediately after.
2. When `session/resume` later fails with `INVALID_SESSION_IDENTIFIER` (or the equivalent for other providers), the daemon treats it as a fatal task error and **retries the same dead id forever** instead of discarding it and creating a new session.

## Suggested fix

- On `session/resume` failure with an invalid/unknown session error, fall back to `session/new` (accepting the loss of in-conversation memory) instead of hard-failing the task.
- Alternatively (or additionally): only record the session id for future resume after the run completes successfully (or after at least one successful prompt round-trip).

## Workaround

Start a new chat conversation with the agent (fresh workdir + fresh backend session). The old conversation cannot be repaired from the CLI — `multica chat` only offers `history`/`thread`, and `multica daemon logs` is not available inside a daemon-managed task.

hawkproject:
Fixes #8116

## Problem

When the first turn of a chat dies at `session/set_model` (invalid model id), qodercli exits **without persisting the session** — yet the id was still reported by the backend and pinned as the conversation's resume pointer (the poison survives the COALESCE in `UpdateChatSessionSession`). Every follow-up then fails at `session/resume` with `Invalid session identifier "…"` forever:

- the phrase was not in `isACPSessionNotFound`'s match set, so the rejection was invisible;
- the `session/resume` / `session/load` failure branches did not set `ResumeRejected`, so the daemon's fresh-session retry never fired;
- no CLI escape hatch exists — the only workaround was abandoning the conversation.

Wording verified by hand-driving `qodercli --yolo --acp` 1.1.25: invalid_params (-32602) `Invalid session identifier <id>` with a structured data object. The code was already in the accepted error-code set; only the phrase was missing.

## Fix — two commits

**1. Backend/daemon self-heal (new rows)**

- `qoder`: on `set_model` failure for a *fresh* session, suppress the session id — a session that never ran a prompt has no conversation worth resuming, and publishing it can only publish a ghost pointer.
- `isACPSessionNotFound` (shared classifier): also match `"invalid session identifier"`.
- `qoder`/`hermes`/`kimi` (`session/resume`) and `kiro`/`traecli`/`grok` (`session/load`): flag `ResumeRejected` when the runtime rejects the recorded id outright, so the daemon's in-turn fresh-session retry fires and retires the dead pointer (`ClearChatSessionSessionIfMatches`). The next message recovers by itself. Previously only zeroclaw/qwenpaw/mcode/dim/reasonix carried the flag on this path.

**2. Server-side guard for legacy rows** (a daemon older than the fix writes the failure with no flag)

- `taskfailure.InvalidSessionIdentifier`: single-source predicate; only the full phrase matches so lookalike errors don't discard healthy pointers.
- `ResumeUnsafeFailure` consults it → a manual rerun starts fresh instead of replaying the rejection.
- `GetLastTaskSession` / `GetLastChatTaskSession` gain the same error-text guard as the existing auth-resolution one; sqlc regenerated.

## Tests

- Per-backend tests pin each runtime's real rejection shape (qodercli frame, kimi `Invalid params` + data, kiro `-32603` data, traecli/grok plain, hermes) and assert `ResumeRejected=true`.
- `TestIsACPSessionNotFound` gains the qodercli frame case.
- Fresh-session `set_model` failures must report an empty session id (qoder).
- `TestShouldRetryWithFreshSession` gains an end-to-end case matching the incident.
- `ResumeUnsafeFailure` table gains positive and negative cases (shorter "invalid session id" wording must NOT flip resume).

Verified on this branch: `go build ./...`, `go vet`, and the touched packages' tests pass.

Co-Authored-By: Claude Code <noreply@anthropic.com>

### DIFF
--- a/server/internal/daemon/daemon_test.go
+++ b/server/internal/daemon/daemon_test.go
@@ -2974,6 +2974,21 @@ func TestShouldRetryWithFreshSession(t *testing.T) {
 			want:           true,
 		},
 		{
+			// The ghost-session incident: qodercli rejects, at session/resume,
+			// an id from a session that died at set_model before its first
+			// prompt (never persisted by the runtime). The backend fix flags
+			// ResumeRejected on that branch; this pins that the incident shape
+			// actually clears the daemon's fresh-session retry gate.
+			name: "qoder invalid session identifier at resume retries",
+			result: agent.Result{
+				Status:         "failed",
+				Error:          `qoder session/resume failed: session/resume: Invalid session identifier "27d8031c" (code=-32602)`,
+				ResumeRejected: true,
+			},
+			priorSessionID: "27d8031c",
+			want:           true,
+		},
+		{
 			name:           "no resume requested never retries",
 			result:         agent.Result{Status: "failed", Error: "boom", ResumeRejected: true},
 			priorSessionID: "",
--- a/server/internal/service/resume_unsafe_test.go
+++ b/server/internal/service/resume_unsafe_test.go
@@ -67,6 +67,27 @@ func TestResumeUnsafeFailureEmptyHistoryMessage(t *testing.T) {
 			want:          true,
 		},
 		{
+			// A runtime rejecting the recorded session id outright (qodercli
+			// "Invalid session identifier ..." at session/resume) can only be
+			// cured by a fresh session. Rows a fixed daemon writes carry
+			// ResumeRejected and never reach here, but an un-upgraded daemon
+			// writes agent_error.unknown with this text — the text guard is the
+			// only thing that stops a rerun from replaying the dead pointer.
+			name:          "invalid session identifier text flips resume even when the reason is unknown",
+			failureReason: "agent_error.unknown",
+			errorText:     `qoder session/resume failed: session/resume: Invalid session identifier "27d8031c". Searched for sessions in ~/.qoder/projects/-tmp. (code=-32602)`,
+			want:          true,
+		},
+		{
+			// The predicate is deliberately the full phrase: a session-shaped
+			// complaint that is NOT about the runtime refusing the recorded id
+			// must stay resumable.
+			name:          "shorter invalid-session wording stays resumable",
+			failureReason: "agent_error.unknown",
+			errorText:     "invalid session id supplied by client",
+			want:          false,
+		},
+		{
 			// Transient failures must stay resumable: the whole point of
 			// reusing the session on retry is to keep the conversation.
 			name:          "provider network drop stays resumable",
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -5247,6 +5247,16 @@ func ResumeUnsafeFailure(failureReason, errorText string) bool {
 	if taskfailure.AuthMethodUnresolved(errorText) {
 		return true
 	}
+	// Same defense-in-depth for a runtime rejecting the recorded session id
+	// outright ("Invalid session identifier ...", qodercli at session/resume):
+	// the daemon-side fix flags ResumeRejected and retires the pointer, but
+	// rows an older daemon wrote carry only the error text. Keep it in sync
+	// with the GetLastTaskSession / GetLastChatTaskSession /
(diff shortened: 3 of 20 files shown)


hawkproject:
Saw #8155 got merged and covers the same defects (both A and B, plus more backends). My PR is a subset — happy to close this. Thanks for the comprehensive fix!

multica-eve:
感谢贡献。#8155 已合入 main，覆盖相同的 ACP ghost session 恢复问题，并扩展处理更多后端及恢复边界。正如本 PR 的说明所述，当前实现是 #8155 的子集，因此以已被替代关闭此 PR。

PR #8130, opened 2026-09-07
MUL-7127 fix: authenticate GitHub release updates

paolurso:
## Summary

- authenticate GitHub release metadata requests with `GH_TOKEN`, falling back to `GITHUB_TOKEN`
- download release assets through their authenticated GitHub API URLs with `application/octet-stream`
- strip authorization on untrusted redirects and reject non-HTTPS/lookalike asset URLs
- preserve the existing checksum verification and atomic replacement flow

## Why

The dashboard updater receives HTTP 403 when release metadata or assets require authenticated GitHub access, even when a valid token is available in the Multica environment. This makes in-app updates fail while manual authenticated GitHub requests succeed.

## Security

- credentials are attached only to exact HTTPS `api.github.com` or `github.com` origins on the default port
- credentials are removed before redirects to CDN or other hosts
- release asset API URLs fail closed unless they are exact trusted `api.github.com` URLs
- credentials never enter request URLs or error strings
- redirects remain limited to 10

## Verification

Per deployment policy, verification ran in the explicitly identified Railway production service using an isolated temporary directory; it did not replace the running binary.

- full `./internal/cli` package suite: PASS
- live authenticated GitHub flow: PASS (`FetchLatestRelease` → archive/checksum API assets → CDN redirects → SHA-256 verification)
- `gofmt` check: PASS
- `git diff --check`: PASS

The temporary live-network test was removed before commit.

### DIFF
--- a/server/internal/cli/update.go
+++ b/server/internal/cli/update.go
@@ -10,9 +10,11 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -31,6 +33,10 @@ import (
 const ChecksumManifestName = "checksums.txt"
 
 const DefaultUpdateDownloadTimeout = 120 * time.Second
+
+const githubAPIVersion = "2022-11-28"
+const githubJSONAccept = "application/vnd.github+json"
+const githubBinaryAccept = "application/octet-stream"
 
 // GitHubRelease is the subset of the GitHub releases API response we need.
 type GitHubRelease struct {
@@ -124,6 +130,7 @@ func parseReleaseVersion(v string) ([3]int, bool) {
 
 type GitHubReleaseAsset struct {
 	Name               string `json:"name"`
+	APIURL             string `json:"url"`
 	BrowserDownloadURL string `json:"browser_download_url"`
 }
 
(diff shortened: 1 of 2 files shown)


paolurso:
@Bohan-J Could you please review this updater fix when convenient? It adds authenticated GitHub release API/asset downloads when a token is configured, while preserving the unauthenticated fallback. All upstream CI checks are passing.

Bohan-J:
Thanks for the careful work here — the host pinning, port/userinfo checks, and redirect handling are more rigorous than most external PRs we see, and the lookalike-host test coverage is appreciated.

We're going to close this rather than iterate on it, because the premise doesn't hold for this project: the updater points at a hardcoded public repo (`multica-ai/multica`, see `server/internal/cli/update.go`), and unauthenticated requests to `releases/latest` return 200. There is no configuration that can point it at a repository whose releases require authentication. The 403 you hit — likely from a shared egress IP — is GitHub's 60 req/hour anonymous **rate limit**, not an auth requirement.

Two concrete problems with the change as it stands:

- **It makes the real problem worse for everyone without a token.** Moving downloads from `browser_download_url` to the assets API moves the checksum manifest and the archive from the public CDN into the metered API. We measured it: a request to `https://api.github.com/repos/.../releases/assets/<id>` decrements `x-ratelimit-remaining`, while the equivalent `https://github.com/.../releases/download/...` request carries no rate-limit headers and consumes no quota. So each update now costs ~2 extra units against a 60/hour budget — on exactly the rate-limited hosts this PR is meant to help. There's also no fallback when an asset's `url` is absent; `releaseAssetAPIURL` fails closed on a path that previously could not fail.
- **It conflicts with existing conventions in this repo.** Server-side GitHub auth already exists as a host-gated helper (`addGitHubAuthHeader` / `newRawFileRequest` in `server/internal/handler/skill.go`), keyed on `GITHUB_TOKEN` and documented specifically as a rate-limit lift. This PR adds a second, differently-scoped implementation and introduces `GH_TOKEN` with higher precedence — a variable this repo already treats as a host secret that must not be inherited (see the `execenv` shell-env allowlist).

If we do address rate limiting in the updater, the fix we'd want goes the other direction: stop touching the API entirely for a public repo. `https://github.com/multica-ai/multica/releases/latest` 302-redirects to the tag page (currently `/releases/tag/v0.4.41`) with no rate-limit headers, and assets are already served from `releases/download/`. That needs no token, no new configuration surface, and helps every deployment rather than only the ones with a credential configured. We'll track that separately.

Sorry for the outcome after the effort you put in — the analysis and the security posture here were genuinely solid, they were just applied to a problem this codebase doesn't have.

PR #6173, opened 2026-07-30
feat(agents): pick an emoji as the agent avatar (MUL-5534)

forrestchang:
## What

Agent avatars can now be an emoji, picked in the UI.

The server has always seeded a new agent with a random `emoji:<char>` avatar (`agentEmojiAvatars` in `server/internal/handler/agent_avatar.go`) and every renderer already parsed the marker — but the only way for a user to *change* one was to upload an image, so an agent could never get back to an emoji once its avatar was replaced.

Clicking an agent avatar now opens a picker with both options:

- **Upload image** — the crop + upload flow the control always had, unchanged.
- **Emoji** — the same 24 faces the server hands out, one click each, with the full searchable emoji-mart picker behind "More emojis…". The agent's current emoji is marked as selected.

## Scope

Emoji is opt-in per call site via the new `onEmojiSelected` prop, wired only into the three agent surfaces (create dialog, creation studio, detail → Settings → General). User, workspace, and squad avatars keep their click-straight-to-upload behavior.

No backend change: `acceptAvatarURL` already passes an emoji marker through untouched, and it never reaches the storage-object authorization path.

## Notes

- `formatAvatarEmoji` joins `parseAvatarEmoji` in `packages/ui/lib/avatar-emoji.ts` so the `emoji:` prefix keeps one owner on the client.
- No success toast on an emoji pick — the avatar swaps in place, so the change is already visible. Failures still surface.
- The emoji data set (~1MB) stays lazy; only the suggestion grid renders eagerly.

## Verification

- `pnpm typecheck` — 6/6 pass
- `pnpm --filter @multica/views test` — 3355 pass; the one failure (`layout/sidebar-resize.test.tsx`) reproduces on a clean checkout and is unrelated
- New `packages/views/common/avatar-upload-control.test.tsx` covers both paths, the emitted marker value, the selected-state marker, and the disabled case
- `pnpm lint` — 0 errors
- Verified end to end against a local stack: picked an emoji on an existing agent, saw it persist and propagate to the page header

Issue #381, opened 2026-04-03
feat(agents): allow changing agent emoji when creating an agent


howardpen9:
## Summary

When creating a new agent, users should be able to pick a custom emoji to represent the agent, instead of being assigned a default one.

## Motivation

Emojis help distinguish agents visually — especially useful when a workspace has multiple agents. Being able to set the emoji at creation time makes the onboarding experience feel more intentional and polished.

## Proposed Behavior

- Add an emoji picker to the "Create Agent" form/modal
- The selected emoji is stored and displayed alongside the agent name everywhere agents appear (sidebar, issue assignee, etc.)
- If no emoji is selected, fall back to a sensible default (e.g. 🤖)

## Acceptance Criteria

- [ ] Emoji picker is shown in the Create Agent dialog
- [ ] Selected emoji is saved to the agent record
- [ ] Agent emoji is rendered correctly in all places agents are displayed

PR #8138, opened 2026-09-08
Add CLI support for explicit agent avatar emoji selection

kylerjensen:
## What does this PR do?

Adds `--avatar-emoji` to `agent create` and `agent update`, accepting a bare emoji glyph. The CLI prefixes the internal `emoji:` marker before sending `avatar_url`; empty and whitespace-only values are rejected before any request is made. Omitting the flag preserves the random default on creation and the current avatar on updates. Existing local-image upload remains available.

## Type of Change

- [x] Documentation update

## Changes Made

- `server/cmd/multica/cmd_agent.go`: register `--avatar-emoji` flag, forward its value with `emoji:` prefix, add whitespace validation, and update help.
- `server/cmd/multica/cmd_agent_test.go`: cover create/update requests with emoji glyph, empty string, whitespace-only, and omitted values; verify error messages for empty/whitespace.
- `apps/docs/content/docs/cli.mdx`, `cli.zh.mdx`, `cli.ja.mdx`, `cli.ko.mdx`: document emoji syntax, omission behavior, and the upload command.
- `server/internal/service/builtin_skills/multica-platform/references/agents.md`: document emoji syntax and omission behavior.

## How to Test

From `server/`:

```sh
go test ./cmd/multica -run '^TestAgentAvatarEmojiRequests$' -count=1
go vet ./cmd/multica
```

Both passed locally.

## Checklist

### DIFF
--- a/apps/docs/content/docs/cli.ja.mdx
+++ b/apps/docs/content/docs/cli.ja.mdx
@@ -178,6 +178,17 @@ multica runtime activity <runtime-id>
 
 `multica chat` が読み取るのは、**エージェントが現在処理している外部チャットセッション**です。主にチャット統合の中のエージェントが使うもので、任意のワークスペースの会話を閲覧するための汎用コマンドではありません。
 
+## エージェントのアバター
+
+エージェント作成時にアバターを設定するか、後から `--avatar-emoji` で更新します:
+
+```bash
+multica agent create --name Fox --runtime-id <runtime-id> --avatar-emoji '🦊'
+multica agent update <agent-id> --avatar-emoji '🦊'
+```
+
+このオプションは絵文字を受け付け、空文字列や空白のみの文字列には設定できません。作成時に省略するとランダムなデフォルトアバターのまま、更新時に省略すると現在のアバターのままになります。ローカル画像をアップロードするには `multica agent avatar <agent-id> --file <image>` を使用します。
+
 ## ID と出力形式
 
 タスクには `MUL-123` のようなキーまたは完全な UUID を使います。UUID の短いプレフィックスは受け付けません。
@@ -249,7 +260,7 @@ CLI の設定ファイルには、あなたとして Multica にアクセスで�
 | | `resource list/add/update/remove` | プロジェクトの追加リソースを管理 | `--type`、`--url`、`--local-path`、`--daemon-id`、`--execution-mode`（ローカルディレクトリの `in_place` / `worktree`） |
 | `label` | `list/get/create/update/delete` | ワークスペースのラベルを管理 | |
 | `property` | `list/get/create/update/archive/unarchive` | ワークスペースのカスタムプロパティを管理 | `create`: `--name`、`--type`（`text`、`number`、`select`、`multi_select`、`date`、`checkbox`、`url`、`actor`、`multi_actor`）、`--option`（繰り返し可、select 系のみ）；`list`: `--include-archived`；タイプは作成後に変更できません |
-| `agent` | `list/get/create/update/archive/restore` | エージェントを管理 | `--name`、`--runtime-id`（`create` では必須）、`--instructions`、`--conversation-starters`、`--model`、`--thinking-level`、`--mcp-config`、`--permission-mode`、`--max-concurrent-tasks` |
+| `agent` | `list/get/create/update/archive/restore` | エージェントを管理 | `--name`、`--avatar-emoji`（絵文字）、`--runtime-id`（`create` では必須）、`--instructions`、`--conversation-starters`、`--model`、`--thinking-level`、`--mcp-config`、`--permission-mode`、`--max-concurrent-tasks` |
 | | `copy <agent-id>` | 新しいエージェントとして複製。元のエージェントは変更されません | `--name`（デフォルトは元の名前 + ` (copy)`）、`--runtime-id`（別のランタイムへ複製する場合は `--model` の同時指定が必須）、`--no-skills`；`custom_env`、`mcp_config`、`runtime_config` などの機密設定は複製されないため、`create` と同じフラグで再指定してください |
 | | `tasks <id>` | エージェントの実行を表示 | |
 | | `avatar <id>` | アバターをアップロード | |
--- a/apps/docs/content/docs/cli.ko.mdx
+++ b/apps/docs/content/docs/cli.ko.mdx
@@ -178,6 +178,17 @@ multica runtime activity <runtime-id>
 
 `multica chat`은 **에이전트가 현재 처리 중인 외부 채팅 대화**를 읽으며 주로 채팅 연동의 에이전트가 사용합니다. 워크스페이스의 임의 대화를 탐색하는 일반 명령이 아닙니다.
 
+## 에이전트 아바타
+
+에이전트를 생성할 때 아바타를 설정하거나 이후 `--avatar-emoji`로 업데이트합니다:
+
+```bash
+multica agent create --name Fox --runtime-id <runtime-id> --avatar-emoji '🦊'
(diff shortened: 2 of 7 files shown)


Bohan-J (changes requested):
## Product review

Accepting the product direction. The problem is real — `agent create` / `agent update` have no avatar input today, and the only CLI avatar entry point (`agent avatar <id> --file`) requires a second command after creation. Letting the same call that configures an agent also set its identity is worth a durable capability, and this is roughly how we'd solve it: extend the existing create/update request path and reuse the current avatar rules, rather than adding a parallel flow.

The implementation shape is right, too. `cmd.Flags().Changed("avatar-url")` correctly distinguishes "omitted" from "explicitly set", so the default behavior (random avatar on create, current avatar preserved on update) is untouched for existing automation. The `emoji:<glyph>` prefix and the flag naming follow existing conventions, and avatar validation/authorization stays with `acceptAvatarURL` on the server.

Two things need to be settled before merge.

## Must-fix

### 1. `--avatar-url ""` has divergent, undocumented semantics — and the test pins it

`server/cmd/multica/cmd_agent_test.go` asserts that `--avatar-url ""` is *sent* as `avatar_url: ""` for both create and update, but the server does two different things with it:

- **Create** — `newAgentAvatar` (`server/internal/handler/agent_avatar.go:34`) checks `strings.TrimSpace(*avatarURL) != ""`, so an empty value falls through to `randomAgentEmojiAvatar()`. The explicit empty string is silently ignored.
- **Update** — `server/internal/handler/agent.go:1845` only checks `req.AvatarURL != nil`, so `""` reaches `acceptAvatarURL`, comes back as `""`, and is written as `pgtype.Text{String: "", Valid: true}` — i.e. it **clears** the avatar.

So the same flag value means "no-op" on create and "clear the avatar" on update. Neither the flag help (`omitted = random avatar` / `omitted = keep current avatar`) nor the new docs mention what happens when the value is supplied but empty, yet the test locks in the current wire behavior as intended.

Please pick one and make it explicit:

- **Support clearing** — document that `--avatar-url ""` clears the avatar on update and is ignored on create, and say so in the flag help. Worth confirming that a stored empty avatar renders acceptably everywhere before committing to this as a supported state.
- **Don't support it** — reject an empty value in the CLI with a clear error, and drop the `empty` case from `TestAgentAvatarURLRequests` (or flip it to assert the rejection).

Either is fine; what shouldn't ship is a test that freezes an undefined contract.

### 2. The localized CLI command reference wasn't updated

`apps/docs/content/docs/cli.mdx` gets both the new "Agent avatars" section and the `--avatar-url` entry in the `agent` flag table, but `cli.zh.mdx`, `cli.ja.mdx`, and `cli.ko.mdx` still carry the old flag list (line 252 in each) and none of them mention the new option.

This doc tree keeps all four languages in sync for CLI changes — 7 of the last 8 commits touching `cli.mdx` updated `cli.zh.mdx` / `cli.ja.mdx` / `cli.ko.mdx` in the same commit. Please add the equivalent section and table entry to the three localized files so the published reference doesn't describe a stale command surface.

## Notes (non-blocking)

- Syncing `server/internal/service/builtin_skills/multica-platform/references/agents.md` alongside the CLI change is the right call.
- Test coverage over create/update × emoji/URL/empty/omitted is good; only the `empty` case needs to follow whatever contract you settle on above.

## Verification

`TestAgentAvatarURLRequests` passes, `go vet ./cmd/multica` is clean, `git diff --check` is clean. A full `go test ./cmd/multica` run in my workspace hits pre-existing failures unrelated to this PR (an ancestor-directory daemon task marker makes many CLI tests resolve an agent context). CI on this PR is green.

## Risk if this is wrong

Blast radius is limited to members who explicitly pass the new flag: the failure mode is an avatar not set as expected, or unintentionally cleared. Reverting removes the entry point, and any affected avatar can be fixed through the existing update or upload path.

🔄 **Request changes** — good feature, right approach; needs the empty-value contract defined and the three localized docs brought in line.

kylerjensen:
Thanks for the review — both points are addressed in the latest commit (`f41be5439`).

1. `--avatar-url ""` is now rejected with a clear CLI error on both `agent create` and `agent update`, before any request is sent to the server. This avoids the divergent create/update semantics you flagged and avoids introducing a stored-empty-avatar state that hasn't been verified to render correctly everywhere. Flag help and docs now say the value must not be empty, and `TestAgentAvatarURLRequests` asserts the rejection (and that no HTTP request is made for the empty case) instead of pinning the old wire behavior.
2. Added the "Agent avatars" section and the `--avatar-url` flag-table entry to `cli.zh.mdx`, `cli.ja.mdx`, and `cli.ko.mdx`, matching the English doc.

Local verification: focused `TestAgentAvatarURLRequests` and `go vet ./cmd/multica` pass; `git diff --check` is clean. CI is running on the new commit.

kylerjensen:
@Bohan-J Both issues have been addressed. Would you mind taking another look at this?

multica-eve (changes requested):
Request changes: the CLI should expose the user concept (`--avatar-emoji 🦊`), not the persisted `avatar_url` representation or arbitrary remote-image URLs; the branch also needs to be rebased onto current `main`.

Thank you for addressing the two findings from the first review: the explicit empty-string case is now rejected, and all four CLI-doc languages are synchronized. On a deeper product review, however, I need to correct my earlier acceptance of the `--avatar-url` shape.

The real gap is that CLI users cannot choose an emoji while creating or ordinarily updating an agent. That capability is worth adding. The established Agent-avatar model, as implemented for #381 / #6165 by #6173, is specifically **choose an emoji or upload an image**. It does not offer “paste an image URL.” Requiring users to type `emoji:🦊` leaks a persisted marker into the public interface, while accepting arbitrary URLs creates a second product state that the web flow deliberately does not produce.

This is not just naming. `packages/ui/lib/avatar-emoji.ts:23-29` keeps the `emoji:` marker private to its formatter. Meanwhile, `server/internal/handler/avatar.go:254-264` is not general URL validation: non-owned values—including third-party URLs and `data:` URIs—pass through unchanged. Those values are then loaded directly by avatar renderers. Promoting that permissive storage field through CLI help and the built-in agent skill would make external availability, viewer-side requests to third parties, and potentially large inline values part of the supported Agent-avatar contract.

Please make these changes:

1. Replace `--avatar-url` with `--avatar-emoji <glyph>` for `agent create` and `agent update`; have the CLI add the `emoji:` prefix internally. Update the tests, four localized docs, and `multica-platform` agent reference accordingly. The existing `multica agent avatar <id> --file <image>` remains the image-upload path. A create-time `--avatar-file` would be a reasonable optional follow-up, but is not required here.
2. Validate the revised value after trimming. The current checks at `server/cmd/multica/cmd_agent.go:660` and `:762` only reject `v == ""`; a whitespace-only value still reaches the server, where create chooses a random avatar but update clears the current avatar. Add a whitespace regression case.
3. Rebase onto current `main` and resolve the conflicts. GitHub reports this PR as conflicting; the overlap is in `cmd_agent.go`, `cmd_agent_test.go`, and the four `cli*.mdx` files after the recent conversation-starters CLI work.

The focused `TestAgentAvatarURLRequests`, `go vet ./cmd/multica`, and `git diff --check` pass on `f41be5439`, and CI is green. Those checks establish that the current pass-through works, but they do not resolve the product-boundary and merge-conflict blockers above.

If this ships in its current shape, failures would surface as broken avatars or viewer devices requesting creator-controlled third-party resources. Reverting the CLI entry point would stop new writes but would not repair already persisted remote or inline avatar values.

kylerjensen:
@multica-eve The requested changes have been implemented.

kylerjensen:
@Bohan-J the requested changes have been implemented.

PR #8147, opened 2026-09-08
[MUL-7136] fix(lark): include workspace slug in issue links

### LINKED ISSUE #8146: Lark issue creation reply links should include workspace slug
## Description

Lark issue creation replies currently generate issue links without the workspace slug:

```text
/issues/{identifier}
```

However, current web routes are workspace-scoped and expect:

```text
/{workspaceSlug}/issues/{identifier}
```

This makes the generated link point to the wrong route when an issue is created from a Lark channel message.

## Expected behavior

Issue creation and duplicate issue replies should link to:

```text
/{workspaceSlug}/issues/{identifier}
```

## Actual behavior

The reply links to:

```text
/issues/{identifier}
```

## Proposed fix

Carry the workspace slug in the channel issue result and use a shared helper to generate workspace-scoped issue web links, while keeping the previous path as a fallback when the slug is unavailable.

m2keloop:
## Summary

Fix Lark issue creation replies to generate workspace-scoped issue links.

Previously the reply linked to:

```text
/issues/{identifier}
```

Current web routes are workspace-scoped, so the correct path is:

```text
/{workspaceSlug}/issues/{identifier}
```

This change carries the workspace slug in the channel engine issue result and uses a shared channel helper to render Lark issue-created and duplicate replies. If the slug is unavailable, the previous `/issues/{identifier}` path is retained as a fallback.

Fixes #8146

## Tests

```bash
cd server
go test ./internal/integrations/channel ./internal/integrations/lark ./internal/integrations/channel/engine
```

### DIFF
--- a/server/internal/integrations/channel/engine/resolvers.go
+++ b/server/internal/integrations/channel/engine/resolvers.go
@@ -58,11 +58,12 @@ type Result struct {
 	ChannelRouteRevision int64
 	// Sender is the platform-native sender id (e.g. Lark open_id), so the
 	// replier can target a binding prompt back to the sender.
-	Sender          string
-	IssueID         pgtype.UUID
-	IssueNumber     int32
-	IssueIdentifier string
-	IssueTitle      string
+	Sender             string
+	IssueID            pgtype.UUID
+	IssueNumber        int32
+	IssueIdentifier    string
+	IssueWorkspaceSlug string
+	IssueTitle         string
 	// IssueDuplicate marks an /issue command that did not create a new issue
 	// because the shared duplicate guard found the active IssueID above.
 	// Repliers render this as a business conflict, never as an internal error.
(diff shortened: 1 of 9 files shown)


Bohan-J:
Thanks for the fix @m2keloop — workspace-scoping the link is exactly right, and reading the slug from the existing `GetWorkspace` call instead of adding a second query is the shape we'd have picked too.

I pushed one follow-up commit rather than sending this back for another round.

The "keep the no-slug fallback" requirement from #8146 doesn't survive a closer look. `workspace.slug` is `UNIQUE NOT NULL` and validated on create, and the router now takes the slug and the issue prefix from the same `GetWorkspace` read. So the only way an empty slug reaches `IssueWebLink` is that read failing — which empties the prefix too and degrades the identifier to `#42`. The fallback branch could therefore only emit a workspace-less path the web app can't route, or `/issues/%2342`. Neither is a usable best-effort URL, and the `TestIssueWebLink` case for it pinned a state production can't reach.

That matters mostly for the contract: `IssueWebLink` sits in the shared `channel` package precisely so the other adapters can reuse it, and "no slug → legacy link" would let a future adapter that forgets to plumb the slug through pass its tests while silently shipping the same broken link this PR fixes. The follow-up makes a missing slug return an empty string (the caller then replies without a link), builds the link from `IssueIdentifier` instead of the `#42` display fallback, and adds a regression test for the degraded-read path.

`channel`, `channel/engine` and `lark` suites pass locally. Merging once CI is green.

PR #8150, opened 2026-09-08
MUL-7145: perf(realtime): stop task:progress ticks from invalidating task caches

adaiguoguo:
## Problem

`task:progress` is broadcast-only narration: `ReportProgress` persists nothing, so the tick itself never signals that any of the eight task-prefix query families changed. Yet every tick fell through to the task-prefix invalidation in `useRealtimeSync` and marked those families stale per debounce window, refetching data that had not changed.

How often that costs anything depends on the runtime. The upstream daemon reports progress twice per run — at launch (`daemon.go` "Launching %s") and at the successful wrap-up ("Finishing task") — so on upstream the 100ms debounce coalesces most of it with the adjacent lifecycle invalidation and the waste is small (measured below). The every-few-seconds stream described in the first version of this PR is an observation from our self-hosted deployment, whose runtime reports progress continuously during a run; there, each tick used to refetch the task families for every viewer with the workspace open.

## Change

Add `task:progress` to `specificEvents` next to `task:message`, which was excluded for the same reason (MUL-6396). Real lifecycle transitions (`queued` / `dispatch` / `running` / `completed` / `failed` / `cancelled`) keep flowing through the prefix invalidation unchanged — a test pins both behaviors.

## Upstream before/after measurement

Setup: this branch's checkout (`ca83809` + review follow-ups, base `028f3b4`), single-node dev server, one viewer with `MEAS-1`'s issue detail (execution log visible) open and settled. The task run replays the exact report sequence the upstream daemon emits for one successful run — `start` → `ReportProgress("Launching claudecode", 1, 2)` → three message batches → `ReportProgress("Finishing task", 2, 2)` → `complete` over ~6.5s — via the daemon HTTP API. Requests counted from the browser's network capture, window = run start until 5s after `complete`, initial page load excluded, preflights excluded. "Before" is the same checkout with the one-line `specificEvents` entry commented out, full page reload between variants.

| Endpoint (GET) | before (main behavior) | after (this PR) |
|---|---|---|
| `/api/agent-task-snapshot` | 3 | 2 |
| `/api/issues/{id}/task-runs` | 3 | 2 |
| `/api/agents?workspace_id=…` | 2 | 2 |
| `/api/issues?limit=50` | 7 | 7 |
| `/api/issues/{id}` | 1 | 1 |
| `/api/inbox`, `/api/inbox/unread-summary`, `/api/notification-preferences` | 3 | 3 |
| **total** | **19** | **17** |

Reading of the delta: the launch report lands ~50ms after `task:running`, inside the debounce window, and coalesces in both variants — exactly as the review predicted. The "Finishing task" report, however, precedes `task:completed` by more than the 100ms debounce (upstream inserts a `GetTaskStatus` round-trip between them), so on main it produces one extra mid-run flush; only the actively observed families actually refetch (snapshot + task-runs here), which is why the saving is 2 requests and not 8 — counting actual requests, not invalidations, as requested. So the upstream saving is small and bounded (~2 requests per run per viewer), and the change is primarily about runtimes that stream progress continuously.

Status promptness and log delivery were checked in the same session: the "agent working" indicator and the running entry in the execution log appeared immediately on `task:running`, message batches arrived live (`task:message` has its own subscription channel, untouched by this PR), and the completed state rendered immediately on `task:completed`.

## Recovery tradeoff (documented, tested)

On main, a `task:progress` tick doubles as a repair signal: a mounted view that missed `task:running` (cross-node publication can fail and the broadcaster drops the error without disconnecting the browser) would be healed by the next tick's refetch within ~100ms. With this change that repair channel is gone: the view stays on its stale status until the next event that still hits the task-prefix invalidation — in the common case the run's own `completed`/`failed`/`cancelled` transition, otherwise a user-driven refetch (remount, window focus). `staleTime: 30s` does not schedule anything by itself.

A new test (`use-realtime-sync-task-progress.test.tsx`, mounted query with a fetching `queryFn`) pins both halves: the tick no longer fetches after a suppressed `task:running`, and the terminal transition still restores the displayed state. Verified in both directions — with the entry removed, the same test fails because the tick does refetch. For upstream's two-report sequence the exposure window is one run's duration; for continuously-reporting runtimes it is the same window this PR removes the traffic from, which is the tradeoff to weigh.

### DIFF
--- a/packages/core/realtime/use-realtime-sync-task-progress.test.tsx
+++ b/packages/core/realtime/use-realtime-sync-task-progress.test.tsx
@@ -1,0 +1,110 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { renderHook } from "@testing-library/react";
+import type { ReactNode } from "react";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { agentTaskSnapshotKeys } from "../agents/queries";
+import type { WSMessage } from "../types/events";
+import type { WSClient } from "../api/ws-client";
+import { useRealtimeSync, type RealtimeSyncStores } from "./use-realtime-sync";
+
+vi.mock("../platform/workspace-storage", () => ({
+  getCurrentWsId: () => "ws-1",
+  getCurrentSlug: () => "test-ws",
+  createWorkspaceAwareStorage: (adapter: unknown) => adapter,
+  registerForWorkspaceRehydration: () => {},
+}));
+
+vi.mock("../paths", () => ({
+  useHasOnboarded: () => true,
+  resolvePostAuthDestination: () => "/",
+}));
+
+const FLUSH_MS = 200;
+
+function createMockWs(anyHandlers: ((msg: WSMessage) => void)[]): WSClient {
+  return {
+    on: vi.fn(() => () => {}),
+    onAny: vi.fn((handler: (msg: WSMessage) => void) => {
+      anyHandlers.push(handler);
+      return () => {};
+    }),
+    onReconnect: vi.fn(() => () => {}),
+    getLastSyncId: vi.fn(() => null),
+    dispatch: vi.fn(),
+  } as unknown as WSClient;
+}
+
+function createStores(): RealtimeSyncStores {
+  return {
+    authStore: Object.assign(() => ({}), {
+      getState: () => ({ user: { id: "u1" } }),
+      subscribe: () => () => {},
+      setState: () => {},
+      destroy: () => {},
+    }),
+  } as unknown as RealtimeSyncStores;
+}
+
+function createWrapper(qc: QueryClient) {
+  return function Wrapper({ children }: { children: ReactNode }) {
+    return <QueryClientProvider client={qc}>{children}</QueryClientProvider>;
+  };
+}
+
+describe("useRealtimeSync — task:progress does not trigger the task-prefix invalidation", () => {
+  let qc: QueryClient;
+  let anyHandlers: ((msg: WSMessage) => void)[];
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    qc = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+    anyHandlers = [];
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  function mount() {
+    renderHook(() => useRealtimeSync(createMockWs(anyHandlers), createStores()), {
+      wrapper: createWrapper(qc),
+    });
+    if (anyHandlers.length === 0) throw new Error("onAny handler was not registered");
+  }
+
+  function fire(type: string) {
+    for (const h of anyHandlers) h({ type, payload: {} } as WSMessage);
+  }
+
+  function snapshotInvalidated(spy: { mock: { calls: unknown[][] } }) {
+    return spy.mock.calls.some(([f]) => {
+      const key = (f as { queryKey?: unknown[] })?.queryKey;
+      return JSON.stringify(key) === JSON.stringify(agentTaskSnapshotKeys.list("ws-1"));
+    });
+  }
+
+  it("ignores progress ticks: broadcast-only payload, no cache changed", () => {
+    mount();
+    const spy = vi.spyOn(qc, "invalidateQueries");
+
+    // A long run's steady progress stream — this used to refetch eight query
+    // families every debounce window for data that never changed.
+    for (let i = 0; i < 10; i++) fire("task:progress");
+    vi.advanceTimersByTime(FLUSH_MS);
+
+    expect(snapshotInvalidated(spy)).toBe(false);
+  });
+
+  it("still refreshes the snapshot on real lifecycle transitions", () => {
+    mount();
+    const spy = vi.spyOn(qc, "invalidateQueries");
+
+    fire("task:running");
+    vi.advanceTimersByTime(FLUSH_MS);
+
+    expect(snapshotInvalidated(spy)).toBe(true);
+  });
+});
--- a/packages/core/realtime/use-realtime-sync-task-progress.test.tsx
+++ b/packages/core/realtime/use-realtime-sync-task-progress.test.tsx
@@ -1,8 +1,8 @@
 /**
  * @vitest-environment jsdom
  */
-import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import { renderHook } from "@testing-library/react";
+import { QueryClient, QueryClientProvider, useQuery } from "@tanstack/react-query";
+import { act, renderHook } from "@testing-library/react";
 import type { ReactNode } from "react";
 import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
 import { agentTaskSnapshotKeys } from "../agents/queries";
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
When a user watches a task, this change is intended to remove redundant background refreshes while keeping the working/completed status and execution log current. That is a reasonable direction for web and desktop.

Requesting changes on `ca838094`. Please keep the follow-up small:

1. **Correct the rationale in the PR description and added code comment.** The upstream daemon reports progress at [launch](https://github.com/multica-ai/multica/blob/ca838094332069ee1ea4699cd8718752d26a0865/server/internal/daemon/daemon.go#L7981) and [successful wrap-up](https://github.com/multica-ai/multica/blob/ca838094332069ee1ea4699cd8718752d26a0865/server/internal/daemon/daemon.go#L5553), rather than continuously every few seconds. Both reports can be coalesced with adjacent lifecycle events by the existing 100ms debounce. Also, `ReportProgress` not writing to the database does not prove that the browser already has current data. Please revise `use-realtime-sync.ts:978–984` accordingly and distinguish internal-deployment observations from upstream measurements.

2. **Add one small upstream before/after comparison.** Use the same setup, one viewer, the same open page (for example, an issue detail showing its execution log), and a comparable task run with and without this change. After initial page loading settles, record actual HTTP request counts for the affected endpoints and check how promptly the running/completed status appears and that log messages still arrive. A short table plus reproduction steps, revisions, and a DevTools capture or equivalent is sufficient. Count actual requests, not query-family invalidations; small or zero savings in that scenario is a valid result.

3. **Check one recovery scenario.** With a mounted view and an otherwise healthy connection, suppress one `task:running` notification, then deliver the following `task:progress`. Compare before/after and report what actually restores the displayed state, and when. A focused test with a mounted query and a fetching `queryFn` is sufficient; the current invalidation spies do not exercise this behavior. This is a plausible failure path: [cross-node publication can fail](https://github.com/multica-ai/multica/blob/ca838094332069ee1ea4699cd8718752d26a0865/server/internal/realtime/sharded_stream_relay.go#L284), and [the broadcaster ignores that error](https://github.com/multica-ai/multica/blob/ca838094332069ee1ea4699cd8718752d26a0865/server/internal/realtime/redis_relay.go#L675), without necessarily disconnecting the browser. `staleTime: 30s` does not itself schedule a refetch. Please document any loss of recovery opportunity so we can assess the tradeoff.

A short verification section and this focused recovery check are enough for the follow-up.

adaiguoguo:
All three items addressed in `d365a43` plus a rewritten PR description.

**1. Rationale corrected.** You're right about the upstream daemon: two `ReportProgress` calls per run (launch and successful wrap-up), not a continuous stream — that stream is our self-hosted runtime, and the code comment plus the PR description now label it as such and describe the coalescing behavior. The "writes nothing to the DB" claim is now scoped to what it actually proves (the tick itself signals no cache change), not to what the browser holds.

**2. Upstream before/after measurement** — table, reproduction steps, and revisions are in the updated PR description. Summary: replaying the exact upstream report sequence against a settled issue-detail viewer, before = 19 GETs, after = 17. The launch report coalesces with `task:running` in both variants, as you predicted; the "Finishing task" report precedes `task:completed` by more than the debounce window (there's a `GetTaskStatus` round-trip between them), so main pays one extra mid-run flush of the actively observed families (snapshot + task-runs). Small and bounded, as you said it might be — the real savings are on continuously-reporting runtimes.

**3. Recovery scenario** — new focused test with a mounted query and a fetching `queryFn`: suppress `task:running`, deliver `task:progress`. On main the tick repairs the stale status within ~100ms (verified by running the test with the entry removed — it fails there because the fetch happens); with this PR the tick no longer fetches and the displayed state is restored by the run's own terminal transition. The tradeoff is documented in the code comment and the PR description: the exposure window is one run's duration, and `staleTime` alone schedules nothing. If you'd rather keep the repair channel, an alternative is a cheap targeted patch on `task:progress` (touch only the snapshot family) instead of full exclusion — happy to switch if you prefer that shape.

adaiguoguo:
You are right, and thank you for checking the actual call sites — closing this.

**Fact check confirms your point 1**: `ReportProgress` fires exactly twice per task lifetime (launch 1/2 at daemon.go:7817, finishing 2/2 at daemon.go:5510), not "every few seconds". I conflated it with `task:message` — the actually-high-frequency stream, which is already excluded. Both progress reports land adjacent to lifecycle events and coalesce into the same 100ms debounce window, so the upstream savings this PR claimed are approximately zero.

That also means your point 3 tips the balance the other way: the progress-triggered invalidation is worth more as an incidental recovery path for a lost `task:running` than the near-zero refetch cost it incurs. Not worth a before/after study to confirm a ~0 result — withdrawing rather than spending your review bandwidth on it.

(Our internal deployment keeps a variant only because a separate reconnect catch-up mechanism replays lost lifecycle events there; that reasoning doesn't transfer upstream.)

adaiguoguo:
Reopening — I closed this by mistake before seeing that the review's three asks had already been addressed in the follow-up commit and comment above (rationale corrected to the actual two-reports-per-task behavior, real before/after request counts on the same page, and the lost-`task:running` recovery check both ways). Please disregard my closing comment.

Note on the force-push since: it only rewrote commit author/committer metadata (a wrong git identity on this branch); the trees are byte-identical to what was reviewed.

adaiguoguo:
@multica-eve requesting a re-review — all three asks from your review are delivered on the current head (`9369816`), and the PR description was rewritten to match:

1. **Rationale corrected**: the upstream daemon reports progress twice per run (launch and successful wrap-up), which coalesces with the adjacent lifecycle invalidation under the 100ms debounce; the every-few-seconds stream is our self-hosted runtime and is labelled as such in both the code comment and the description.
2. **Upstream before/after measurement**: real browser network capture against a settled issue-detail viewer, replaying the exact upstream report sequence — 19 GETs before, 17 after; table, reproduction steps and revisions are in the description. The saving is the one mid-run flush the "Finishing task" report causes because it precedes `task:completed` by more than the debounce window.
3. **Recovery scenario**: a focused test with a mounted query and a fetching `queryFn` — suppress `task:running`, deliver `task:progress` — verified in both directions (the same test fails on `main` because the tick does refetch). The tradeoff is documented in the comment and the description, and I offered a targeted-patch alternative if you would rather keep the repair channel.

The head SHA differs from the one in my 2026-09-08 comment only because the commits were re-authored to my usual identity; the tree is unchanged. Sorry for the noise from the accidental close/reopen in between.

adaiguoguo:
@multica-eve following up on the re-review request from the 17th — head is still `9369816`, with all three changes you asked for in it. Is there anything outstanding, or can this go back in the queue?

Bohan-J:
Thanks for the thorough follow-up on `9369816`. The upstream measurement and the recovery test are exactly what was needed to decide this, and they settled it cleanly. We're closing without merging, for the reason your own numbers point to.

**Benefit upstream is about 2 GETs per run per viewer.** The launch report coalesces into the `task:running` debounce window, and only the "Finishing task" report produces an extra flush. That is bounded, unchanged data, and not worth a behavior change. The every-few-seconds stream that motivated the PR is specific to your self-hosted runtime.

**Cost upstream is not zero.** On a multi-node deployment the relay publish can fail for a single frame (`sharded_stream_relay.go` returns the XADD error, `DualWriteBroadcaster` discards it) while the browser stays connected, so the reconnect backstop never fires. Today the launch tick repairs a lost `task:running` roughly 50ms later. With this change, a viewer on another node stays on the stale status for the whole run. Your tradeoff test pins that precisely, and your 2026-09-09 comment reached the same conclusion we did.

We also considered the targeted-patch alternative (touch only the snapshot family on progress). It keeps the "no-op event triggers a refetch" pattern this PR set out to remove, for roughly one GET, so it doesn't change the answer.

Two things this PR surfaced that we will carry forward separately:

1. **Relay publish failures are swallowed with no retry or reconciliation.** Any lifecycle frame can be lost cross-node today. A lost `task:completed` leaves other nodes showing "working" until the viewer navigates. That is a real latent bug independent of this PR and we will track it on our side.
2. **If upstream ever adds continuous progress reporting**, `task:progress` should get a dedicated handler that patches a progress field rather than a blanket exclusion. At that point your runtime's experience becomes directly relevant, and we'd welcome a reopen.

You mentioned your deployment has a reconnect catch-up mechanism that replays lost lifecycle events. That is the piece that would make this exclusion safe upstream. If you'd ever like to upstream that, it would be a much more valuable contribution than this one and we'd be glad to review it.

adaiguoguo:
Taking you up on that — and saying so here before writing code, since the last time around nobody said a PR was in flight.

**What exists.** Since 2026-09-08 every outbound realtime frame on our deployment goes through one writer that appends it to a `workspace_event` table and only then hands it to the `Broadcaster`; the row id rides on the frame as `sync_id`. On reconnect the client sends its highest seen id to `GET /api/events?since=` and replays only what it missed, through the normal dispatch path, instead of invalidating ~30 workspace-scoped query families wholesale. Whenever completeness cannot be proven — no id seen yet, `since` older than the oldest surviving row, more than 500 behind, request failed — the server says `resync` and the client falls back to today's behaviour. A per-workspace advisory lock keeps ids in commit order across instances; pruning deletes by id watermark rather than raw age so that rule survives the cutoff boundary. `task:progress`, `task:message` and `chat:message` are delivered but never logged. ~1500 lines including tests.

Field data from this morning, read straight off the table: the 24h retention window holds 132 events across 3 workspaces with ids 1709–1840 and **no gaps**, so the watermark rule has held in production; 1,840 ids assigned since launch on 09-08; 6 of the 132 are personal events. To be clear about what that does and does not prove — this is a small deployment, peak 47 events/hour. The invariants are field-verified; throughput at your scale is not. The log also taught us one thing on its own: `daemon:register` is a third of our rows (it fans into every workspace's log on each registration) and is worthless to replay, so it belongs with the delivery-only set — I'd fold that into slice 1.

**How it sits against your relay.** The writer takes a `realtime.Broadcaster` and calls only `BroadcastToWorkspace` / `SendToUser(…, excludeWorkspace)` / `Broadcast` — the interface `DualWriteBroadcaster` already implements — so the relay is untouched. The row is written before the `_ = d.relay.PublishWithID(...)` you pointed at, which is what turns a dropped XADD from "lost" into "not yet delivered".

**What it does not fix yet, to be precise about it.** Replay only runs on reconnect. Your cross-node case is a browser that *stays* connected, so today's version would leave it stale exactly as `main` does. The piece that closes it is small and client-side: the frame after the dropped one carries `sync_id` N+1 while the client last saw N−1, so a gap is detectable and can trigger the same catch-up without a reconnect. That slice is what would make the #8150 exclusion safe; it is not in production yet, so I'd build and measure it rather than claim it.

**Two calls I'd want from you before coding.**
1. The v1 insert runs after the producing transaction commits, not inside it — a crash in between loses the event from wire and log alike. Acceptable as a first landing with a transactional outbox as follow-up, or do you want the outbox from the start? It changes the shape of the first PR considerably.
2. Personal events (inbox, invitations, chat lifecycle) consume ids from the same sequence, so under gap detection a client will see legitimate gaps for other members' personal frames and make cheap, empty catch-up calls. They are ~5% of our log, so I'd accept that over partitioning the ids — unless your ratio looks very different.

**Proposed slicing**, each behaviour-preserving until the last:
1. Table + writer + `sync_id` on frames + pruner. Additive field, no client change. (~600 lines with tests)
2. `GET /api/events` + client reconnect replay with the fallback rules above. (~500)
3. Gap-triggered catch-up. Then the #8150 exclusion as its own small PR on top.

If the shape works I'll open a tracking issue with the contract written down and start on slice 1.

PR #8151, opened 2026-09-08
MUL-7144: feat(comment): client_request_id idempotency key for retry-safe creation

adaiguoguo:
## Problem

Agents retry far more than humans. A re-run `multica issue comment add` — an agent retrying after a timeout, a flaky gateway, or its own session being restarted — posts a duplicate comment, because comment creation is not idempotent. Duplicate agent comments then re-trigger mentions and clutter the thread.

## Change

`POST /api/issues/{id}/comments` accepts an optional `client_request_id` (max 128 chars, unique per issue via a partial unique index):

- A retry carrying the same key gets the **already-created comment back with 200** instead of creating a duplicate — via a replay pre-check, plus unique-violation recovery for the concurrent race.
- Scoped **per issue** rather than per workspace, so one key scheme (e.g. `<task>:final-report`) stays reusable across the issues an agent works on.
- No key → no dedup: existing behavior is unchanged, and a test pins that.
- CLI: `multica issue comment add --idempotency-key <key>`.

Migrations follow the repo rules: no FK, the unique index builds with `CREATE INDEX CONCURRENTLY` in its own single-statement file, and it is registered in `concurrentIndexCleanups`.

## Verification

- 4 handler tests: same-key replay returns the same comment id with exactly one row; distinct keys create distinct comments; the same key on different issues stays independent; over-length key → 400; no-key duplicates pinned as unchanged behavior.
- Running in production on our internal deployment; verified live that a same-key retry returns the original comment (201 → 200, one row) and that the idempotent replay does not re-broadcast a `comment:created` event.


multica-eve (changes requested):
Before this change, retrying a comment after its acknowledgement is lost can create a duplicate comment and notify people twice. After this change, a person or agent that reuses the same stable identifier gets the original comment back, while submissions without an identifier keep their current behavior.

## Product decision

The problem is real: the current create path cannot distinguish a retry after a server-side success from a new submission. It is worth adding a durable idempotency mechanism because duplicate comments can also duplicate mentions and downstream work. However, I would not ship it with the current boundary: the actual agent delivery path should use it by default, the identifier must be bound to the submitter and exact operation, and a replay should only report success once the whole operation is complete or recoverable.

So the direction is welcome, but the current design still needs maintainer decisions on default adoption, key scope, payload mismatch behavior, and the operation's completion boundary.

## Must-fix issues

1. **The migration versions now collide with `main`.**

   This PR uses `server/migrations/454_comment_client_request_id.up.sql` and `455_comment_client_request_id_index.up.sql`, but #8148 has since landed different migrations using versions 454 and 455. A migration source cannot safely contain two migrations with the same versions, and an environment that has already recorded the versions will not receive this column/index. The new `CreateComment` query writes `client_request_id` even when the request has no key, so deploying the binary without the column makes all comment creation fail with 500.

   Please rebase, renumber both migrations after the current tip, update the cleanup registration in `server/cmd/migrate/main.go`, regenerate sqlc, and rerun migration and handler tests.

2. **Row insertion is treated as completion before the complete comment operation has finished.**

   `server/internal/handler/comment.go:1744-1755` returns a replay as soon as a comment row exists. Attachment linking, realtime publication, thread reopening, deferred-escalation cancellation, and mention-triggered work happen later at `:1924-1956`. If the process exits after the insert, or a concurrent replay returns before the first request completes those steps, every later retry skips recovery permanently.

   The replay response at `:1970-1977` also cannot reproduce the original `trigger_outcomes`, so a caller whose first response was lost no longer learns that a mention was blocked or deferred. Please persist a recoverable operation state/original result, or use the project's accepted transaction plus durable-outbox pattern so the idempotency record represents the complete operation rather than only the comment row.

3. **The key is not bound to the authenticated actor or the original payload.**

   The lookup and unique index use only workspace/issue/key (`server/pkg/db/queries/comment.sql:405-410`, `server/migrations/455_comment_client_request_id_index.up.sql:4-6`). The early replay also runs before parent/attachment parsing, actor resolution, and the task reply-parent / `no_action` guards (`server/internal/handler/comment.go:1744-1755` versus `:1759-1863`).

   As a result, another actor—or the same actor sending different content, parent, attachments, or suppression choices under a reused key—receives the old comment with 200 and can mistake that for a successful delivery. Bind the key to the authenticated actor/task and a canonical request fingerprint; the same key with a different operation should return 409.

4. **CLI retries with attachments still create orphaned uploads.**

   `server/cmd/multica/cmd_issue.go:2157-2178` uploads every attachment before posting the keyed comment. If the first comment request succeeds but its response is lost, a retry uploads every file again; the server then replays the old comment and ignores the new attachment IDs. Those new uploads remain unbound, contradicting the flag's “safe to retry” promise. Attachment upload must participate in the same idempotent operation, or the CLI must durably reuse the first upload IDs.

5. **The concurrency branch and the real agent entry point are not covered.**

   `server/internal/handler/comment_idempotency_test.go:21-43` only performs sequential replay; it never drives two requests past the pre-check concurrently, so the unique-violation recovery path is untested. It also omits same-key/different-payload, different-actor, attachments, trigger outcomes, and interruption after insertion. The CLI test helper at `server/cmd/multica/cmd_issue_test.go:356-365` does not register the new flag, so existing CLI tests do not prove it is sent.

   More importantly, the runtime still instructs agents to post the mandatory final comment without a stable key (`server/internal/daemon/execenv/runtime_config_sections.go:277` and the per-turn delivery command). A restarted session following today's instruction will still duplicate the comment. Please decide whether the CLI derives a key from trusted task context or the dynamic turn instruction supplies one, then add an end-to-end regression for that actual path.

6. **The new key contract diverges from the existing idempotency convention without a rationale.**

   Manual autopilot triggering already uses the `Idempotency-Key` header with a 255-byte limit (`server/internal/handler/autopilot.go:2387-2395`). This PR adds a JSON `client_request_id` with a 128 “characters” limit without explaining why the existing convention cannot be extended. `server/internal/handler/comment.go:1739-1742` uses Go's `len`, so the implementation actually limits UTF-8 bytes, not characters. Please settle the transport, unit, and limit and document the reason.

## Validation

- `git diff --check 028f3b4...origin/pr-8151` passed.
- `go test ./cmd/multica -run 'TestRunIssueCommentAdd' -count=1` passed, but does not exercise the new flag because of the missing test wiring above.
- `go test ./internal/handler -run 'TestCreateComment(IdempotencyKey|WithoutIdempotencyKey)' -count=1` failed against a database at current `main` migration state with `column "client_request_id" does not exist`, which is the unsafe mixed-version path caused by the 454/455 collision.

The nullable column, opt-in behavior for old clients, concurrent unique-index build, and absence of foreign keys are good choices to preserve. The blockers above need to be resolved before merge.

adaiguoguo:
Thanks for the thorough review — the mixed-version catch in particular was exactly the kind of thing that only shows up when you test against a database at current `main`. I've pushed the mechanical fixes and would like maintainer direction on the design items before reworking the boundary.

## Fixed in 0d37512

**1. Migration collision** — rebased onto `main`, renumbered 454/455 → 456/457, updated the cleanup registration in `cmd/migrate/main.go`, regenerated sqlc (no drift), and verified by migrating a fresh database from zero through 457. `TestCreateComment(IdempotencyKey|WithoutIdempotencyKey)` and the migration run both pass against the post-#8148 schema.

**6 (unit half)** — the implementation limits UTF-8 bytes via Go `len()`, so the error message and CLI help now say "128 bytes", not characters.

**5 (CLI wiring half)** — `newIssueCommentAddTestCmd` now registers the flag, and `TestRunIssueCommentAddSendsIdempotencyKey` proves `--idempotency-key` reaches the server as `client_request_id` and that an unkeyed request omits the field entirely.

## Design items — my read, pending your call

**3. Key binding (actor + fingerprint)** — agreed, and I'm ready to implement it in this PR: move the replay check after auth/actor resolution and request validation, bind the unique key to the authenticated actor, store a canonical fingerprint (content, parent, attachment IDs, suppression flags), and return 409 when the same key arrives with a different operation.

**2. Completion boundary** — agreed that "row exists" is not "operation completed". The clean fix is your suggestion: make the idempotency record represent the whole operation (insert + attachment linking + triggers + original `trigger_outcomes`) via the accepted transaction/outbox pattern, and replay the persisted original result. That is a substantially bigger change than the current PR. Before I build it: do you want that full operation-record design in this PR, or would you rather I narrow this PR's contract (documented as "no duplicate comment row; side effects best-effort") and do the operation record as a follow-up? I'll take either direction.

**4. CLI attachments** — agreed it currently breaks the flag's promise. Unless you prefer durable client-side upload reuse, my proposal for v1 is to reject `--idempotency-key` combined with `--attachment` with a clear error, and lift the restriction when attachment upload participates in the operation record from item 2.

**5. Agent path adoption** — the motivating case is exactly the one you describe: the mandatory final comment posted by a restarted session. My preference is that the server-side per-turn delivery instruction supplies the key derived from trusted task context (e.g. `task-<id>:final`), so agents don't have to invent one. If you agree, I'd wire that into `runtime_config_sections.go` plus an end-to-end regression in this PR once the boundary from item 2 is settled.

**6 (transport half)** — happy to switch to the existing `Idempotency-Key` header convention with its 255-byte limit for consistency with autopilot; the JSON field was chosen only because the CLI posts through the JSON body helpers. Say the word and I'll align it.

If you'd rather see this split — a small PR that only hardens the key contract (items 3/6) and a second one for the operation-record boundary (items 2/4/5) — that also works for me.

adaiguoguo:
Converting to draft rather than renumbering again: `main` has moved past 456/457 a second time (now at 467), and the same collision would recur every week while the design questions from the review — replay completion boundary (outbox vs. persisted operation state), binding the key to the actor + request fingerprint, attachment participation, and whether to extend the existing `Idempotency-Key` header convention instead of a body field — are still waiting on a maintainer direction.

Once there is a direction on those, I'll rebase, renumber the migrations to the current tip, and address the design items in one pass. The mechanical fixes from the first round (cleanup registration, byte-limit wording, CLI flag test wiring) stay in the branch. Happy to go whichever way maintainers prefer on the design points.

adaiguoguo:
Closing this draft rather than renumbering the migrations a third time.

`main` has moved past this branch's migration numbers twice since the PR was opened (454/455, then 456/457, and `main` is now at 467+). A conflicting draft with no maintainer input yet on the open design questions is not worth the weekly renumbering churn, so I am taking it off the board. The branch stays on the fork and the first-round review fixes are preserved there.

Still looking for direction on four points before resubmitting:

1. **Replay boundary** — should a replayed key return the existing comment only once the original request has fully completed (attachments, mentions, notifications), or as soon as the comment row exists?
2. **Key scope** — bind `client_request_id` to the actor plus a content fingerprint, or keep it a plain per-issue unique key?
3. **Attachments** — should attachment IDs participate in the fingerprint, or stay out of it?
4. **Transport** — keep it as a JSON field on the request body, or follow the `Idempotency-Key` header convention?

Once those are settled I will open a fresh PR rebased on current `main` with the migrations renumbered once. Related design context for the wider sync-log work is in #8226.

PR #8161, opened 2026-09-08
[MUL-7148] fix(lark): distinguish current requests from prior group context

ykwok:
## What does this PR do?

Prevent a standalone `@Bot` in a Feishu/Lark group from turning earlier, already answered questions into the current request.

A real conversation exposed this sequence (generalized here):

1. A member asks about a feature's value and free quota; the bot answers in an interactive card.
2. The same member asks a new question: “Which two scenarios should we prioritize, and what should the entry UI look like?”
3. The member sends `@Bot` as a separate message.
4. Mention removal leaves the trigger's body empty. The enricher supplies recent user questions but drops bot cards before request selection, so the model can answer the old quota question again before addressing the new question.

The issue is the agent input's turn boundary, not a duplicated outbound card. Recent history is useful for continuity, but it should not implicitly become a backlog. This change explicitly separates `current_request` from `reference_context`, and conservatively identifies the intended preceding request for a mention-only trigger. It uses the existing history fetch, with no additional model or API calls.

This builds on the current-input title selection already merged in #8059. The existing explicit `CommandText` path stays preferred; when the body fallback contains the new generated envelope, both deterministic and LLM title inputs use its current request instead of internal framing.

## Related Issue

No separate issue. The generalized reproduction and expected behavior are included above. Related: #8059 (chat title source selection).

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/integrations/lark/inbound_current_request.go`: frame current instructions and reference context separately. For an empty text/post trigger, prefer its explicit quoted message; otherwise select at most three immediately preceding same-sender messages within five minutes. Bot messages (including cards), prior mentions, other speakers, topic changes, missing timestamps and unsupported content stop inference. Ambiguous, unavailable or overlong candidates request clarification instead of silently truncating or restarting old work.
- `server/internal/integrations/lark/inbound_enricher.go`: keep card records as selection boundaries while still omitting opaque card placeholders from rendered history. Include the trigger's second in the history query and apply millisecond filtering, so a slightly earlier message in the same second is not lost. Control commands and forwarded-message paths retain their existing handling.
- `server/internal/integrations/channel/engine/title.go`: unwrap only the known generated envelope for title fallback; preserve current `CommandText` preference.
- Colocated tests: cover the reported sequence, explicit recap/quote requests, missing quotes, repeated bare mentions while prior work may still be running, speaker/topic/card/mention/time boundaries, overlong message runs, same-second messages and title framing.
- `apps/docs/content/docs/lark-bot-integration{,.zh}.mdx`: document separate-message mentions, bounded selection and explicit quoting for older requests.

Behavior boundaries and risks:

- This is conservative request inference, not a durable processed-message ledger or a guarantee that a model can never repeat itself. Selected source IDs are recorded in the enriched input; existing trigger IDs, routing, command classification and delivery attribution are preserved.
- The five-minute / three-message limit and boundary checks intentionally prefer clarification in ambiguous conversations. Quoting the intended message or including the request with the mention remains the explicit path.
- Interactive cards establish a turn boundary; this change does not extract their full rendered answer text or rebuild conversation memory. Regular requests keep history as reference, while mention-only inferred requests exclude unrelated history from the generated input.
- This applies to the Lark group-enrichment path with recent context enabled. No database migration, deployment change, bot-specific configuration, new agent or paid model call is introduced.

## How to Test

1. Reproduce the four-message sequence above in a test group. The next response should address scenario selection and UI, without restarting the quota explanation. This live check has **not** been performed for this PR.
2. From `server/`, run:

   ```bash
   ../scripts/go-test-with-agent-cli-guard.sh -- go test -race \
     ./internal/integrations/lark ./internal/integrations/channel/engine -count=1
   go vet ./internal/integrations/lark ./internal/integrations/channel/engine
   GOOS=linux GOARCH=amd64 go build -o /tmp/multica-current-request-server ./cmd/server
   ```

3. Local validation on upstream main `7a9dc633d26b2b0d6c3c94825e2716c27ed2b712` plus this patch:
   - Both package suites passed with `-race`: 647 test/subtest pass events; 28 database-dependent tests skipped because a test PostgreSQL / `DATABASE_URL` was unavailable. No failures.
   - `go vet`, Linux amd64 server compilation, `gofmt` and `git diff --check` passed.
   - Tests use fake Lark responses and do not send live messages, call a real model, or exercise a production deployment. Database-backed tests and the full monorepo suite remain for CI / maintainer validation.

## Checklist

- [x] I have run tests locally and they pass (database-dependent skips documented above)
- [ ] If this change affects the UI, I have included before/after screenshots — no layout change; title values are covered by tests
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — not applicable
- [x] If this PR touches Chinese product copy, I checked it against the Chinese conventions document

## AI Disclosure

**Prompt / approach:** Investigate repeated answers when a member sends a follow-up and then a separate bot mention; preserve useful context without treating old questions as new work. Codex traced the inbound enrichment path, added a bounded selection rule and regression fixtures, adapted the patch to current upstream main, and ran the checks above. The public reproduction and fixtures are generalized; no private group transcripts, identifiers or credentials are included.


Bohan-J (changes requested):
## Product read

Previously, when a member posted a new question and then sent a bare `@Bot` reminder, the bot could treat an older, already-answered question as this turn's request and re-answer it. With this change it prefers the immediately preceding new question, asks for clarification when the boundary is unclear, and treats older discussion as background only. That direction directly reduces off-topic answers and the need for users to re-send their question.

**Verdict: 🤔 direction should be accepted, but the policy values need a maintainer decision.**

- The problem is real: a blank trigger does show up alongside multiple historical questions, and bot answer cards are hidden, so the boundary between "current question" and "already answered" genuinely gets lost.
- Worth solving: separating the current request from background context and failing closed on ambiguity keeps the whole loop inside the group chat, no context switch for the user.
- If we built this ourselves, we'd also extend the existing group-context path and default to conservative clarification. But `5 minutes / at most 3 messages` is a long-lived product contract that will end up in user docs, and the PR gives no data, platform constraint, or prior decision behind those two values. Please confirm both thresholds explicitly, or replace them with bounds that have a stated rationale. Until that's confirmed this is a product gate.

Blast radius is Lark/Feishu group chats with recent context enabled, on by default in production. No migration or config change; users don't need to migrate, but an ambiguous bare `@` goes from "may answer directly" to "asks for clarification".

## Must fix

**1. Resolved sender names are dropped on the blank-`@` + quoted-message path**

`server/internal/integrations/lark/inbound_current_request.go:29` re-renders the quote with `nil`, discarding the names already resolved at `server/internal/integrations/lark/inbound_enricher.go:210`. This breaks the existing contract from MUL-3084. A regression test reproduces the quoted sender changing from `Alice` to `User 1`. Please pass the resolved names through, and add a name assertion for this path; merge-forward quotes should be covered too.

**2. The title model still receives the full history on a bare-`@` trigger**

`server/internal/integrations/channel/engine/title.go:21-29` only unwraps when `commandText` is empty, but `server/internal/integrations/channel/engine/router.go:209-210` replaces empty raw input with the fully enriched text first. So for a new conversation started by a bare `@`, the deterministic title is fine, yet the text handed to the title model is still the whole history plus the wrapper. A router-level regression test reproduces this — the actual model input contains `old quota`. The helper unit test added in this PR bypasses that production fallback. Please add a router-level test and have the deterministic title and the model title read from the same trusted current-request source.

Separately, having the shared `DeriveChatTitle` recognize a Lark-specific wrapper via a user-typable English prefix isn't advisable: a user on another channel who types the same prefix and labels would get their message unwrapped too. Carrying a structured, trusted title source on the channel message is safer than parsing it back out of the body.

**3. Per-turn input cost is uncontrolled**

`server/internal/integrations/lark/inbound_enricher.go:256-258` attaches the long wrapper to every ordinary group request, not just blank `@`s, and the fixed instructions at `server/internal/integrations/lark/inbound_current_request.go:45-53` are persisted with the message and accumulate turn over turn. Measured on one ordinary message from this PR, the body grows from 88 bytes to 532 bytes — +444 bytes net. Twenty messages in the same conversation add roughly 8.9 KB of persisted content and about 93 KB of replayed input across turns. Please trim the instructions to the minimum semantics needed, apply them only on the ambiguous path, or move them to an appropriate caching surface — and report measured overhead for whatever the final approach is.

## What's good

The change reuses the existing history fetch and adds no external calls. Keeping bot cards as selection boundaries without rendering placeholders, the millisecond-level filter for same-second messages, and failing closed on ambiguity are all sound choices. The test coverage for speaker, topic, card, stale `@`, missing timestamp, and over-long consecutive-message boundaries is valuable.

## Verification

- Relevant non-database tests: `go test -race` passes.
- `go vet`, Linux amd64 server build, and `git diff --check` all pass.
- Two temporary negative regression tests reliably reproduce the name loss and the title-model input pollution; those test files were removed afterwards.
- The full-package race command is blocked by a locally unmigrated test database (missing pre-existing tables/columns) — not an assertion failure from this PR. CI is green. The real group-chat verification the PR describes has not been run.

If this lands wrong, the first symptoms will be wrong answers in group chats, over-eager clarification requests, or titles carrying history. Reverting the server commit is enough; the exposure is Lark/Feishu group chats with recent context on by default.

## Merge recommendation

🔄 **Request changes** — the product direction is right, but there are two reproduced behavior regressions, one unconfirmed product-policy gate, and uncontrolled per-turn input cost. Not ready to merge as-is.

Issue #8157, opened 2026-09-08
[Bug]: Pi task can remain running after work has already completed and provider activity has stopped


makesomethingshit:
### Deployment type

Official App

### Area

Desktop app

### Summary

A Pi task completed its actual work and posted its final report, but the server-side task remained in `running` for about 11 minutes with no further provider/run activity.

The task was eventually transitioned to `cancelled`, but both `result` and `error` were `null`.

This appears to be a daemon terminal-disposition synchronization/lifecycle issue somewhere between post-run teardown and the final `CompleteTask` / `FailTask` reporting path.

This may be related to #4578, but I’m reporting a current reproduction because the same class of failure still appears observable after that issue was closed.

### Observed timeline

- The agent completed the requested work.
- The final task/report comment was successfully posted before the provider stopped.
- Last provider message:
  - `04:33:49.249Z`
- After `04:33:49.249Z`:
  - no additional provider activity
  - no additional run messages
- The server-side task nevertheless remained:
  - `status = running`
- At approximately:
  - `04:44:53Z`
- the task transitioned to:
  - `status = cancelled`
  - `result = null`
  - `error = null`

No idle-watchdog termination was observed.

### Expected behavior

Once the agent has completed execution, daemon/server state should eventually converge to an appropriate terminal state.

A task whose execution has already finished should not remain indefinitely or unexpectedly in `running` solely because delivery of its terminal disposition was interrupted or delayed.

### Actual behavior

The work was already complete and externally observable through the final report, and provider activity had stopped, but the server task remained `running`.

The eventual `cancelled` state appears to have come through a separate later cancellation path rather than from the original execution result.

Observed state progression was effectively:

```text
Pi performs task
→ final report/comment is posted
→ provider activity stops
→ daemon does not settle the terminal state on the server
→ server row remains `running`
→ later, a separate cancellation path changes it to `cancelled`
```

### Why this looks like a daemon terminal-state lifecycle issue

Current daemon behavior allows a terminal reporting attempt to exhaust its retry path without falling back to reporting the task as failed, because converting a successfully completed task into a failure would also be incorrect.

That means there is a legitimate path where:

```text
agent execution succeeds
→ terminal callback cannot be settled
→ daemon gives up reporting
→ server-side task remains `running`
```

This is consistent with the observed state.

I cannot confirm from the available data whether the exact failure was:

- `CompleteTask` receiving transient 5xx responses or transport timeouts,
- post-run teardown blocking before terminal reporting,
[...]

PR #8166, opened 2026-09-08
MUL-7147: fix(daemon,server): fence terminal reports to their claim

makesomethingshit:
## What does this PR do?

Fixes terminal-report ownership across task reclaims.

A terminal result now belongs to `(task id, claim dispatched_at)`, not only the task id. The daemon persists that generation with the durable terminal report, and current servers compare it inside the same UPDATE that performs complete/fail.

Generation-aware reports use dedicated v2 terminal endpoints. This matters during rolling deploys: an old replica has no v2 route and therefore answers 404 instead of silently applying the report through the legacy unfenced path.

## Safety model

- Claim responses explicitly advertise `terminal_report_generation_fence_v1`; `dispatched_at` alone is never treated as fence support.
- Generation-aware durable records are v2 and require `claim_dispatched_at`.
- Queue/in-flight identity is `(task id, generation)`.
- Generation-aware reports always use `/api/daemon/v2/tasks/...`; there is no fallback to the legacy terminal routes.
- The v2 route requires a valid non-zero `expected_dispatched_at` (missing/empty/malformed → 400) and has no unfenced mode.
- The generation comparison stays inside the terminal UPDATE (`dispatched_at = @expected_dispatched_at`), so ownership cannot race a reclaim.
- Plain v2 404 means an old/unsupported replica and stays pending: never acknowledged, never quarantined, never compensated.
- Structured `task_not_found` means the task genuinely no longer exists.
- `task_claim_generation_mismatch` retires the stale report without complete→fail compensation.
- Legacy daemons keep using the existing unfenced terminal endpoints.

## Version skew

- An old-server claim carries `dispatched_at` but no capability: it stays legacy — unfenced live callback, no `expected_dispatched_at`, no v2 record.
- A capability-advertising claim with a missing, empty or malformed `dispatched_at` is a protocol error: terminal reporting is refused rather than degraded into an unfenced write.
- v1 records are retained and never replayed; a v2 record without a generation is corrupt and is neither replayed nor downgraded.

## Validation

- old-server claim payload with `dispatched_at` stays legacy;
- matching v2 generation completes/fails;
- stale generation cannot settle a reclaim (and a stale completion after a newer settlement returns the stable 409);
- same-generation replay remains idempotent;
- two generations of one task coexist independently in the durable queue, including in `failed/`;
- v1 records are retained but never replayed;
- old replicas return a plain v2 404 and the report remains pending until a current replica accepts it, with zero legacy calls;
- structured `task_not_found` stays distinguishable from an unmatched route;
- router registration covers both terminal surfaces;
- CI green.

## AI disclosure

Implementation written with Codex assistance.


makesomethingshit:
Quick reviewer map in case the size of this PR makes it awkward to enter:

1. Ownership/lifecycle invariant — once the daemon has produced a terminal result, it keeps responsibility for delivering that result while the daemon is alive, but only while the same claim still owns an eligible server task. The recovery loop does not create a new execution timeout/watchdog or resurrect terminal tasks.
2. Stale-claim safety — terminal callbacks carry the server-provided dispatched_at, and the server fences the terminal UPDATE against it. A reclaim that happens before or during callback retries therefore rejects the old callback instead of letting an earlier claim settle the newer one.
3. Failure boundary / evidence — this intentionally fixes the reproduced transient-callback-exhaustion path from #8157 without claiming the original production incident was reproduced end to end. The final head has remote CI green, plus focused daemon/handler and PostgreSQL regressions around stale reclaim and terminal settlement.

The part I’d especially value review on is whether the claim-generation fence + live-daemon recovery ownership boundary matches the lifecycle model you want upstream. If that boundary is right but the patch is too broad to review comfortably, I’m happy to split follow-up pieces rather than expand the scope further.

makesomethingshit:
I’m going to narrow the scope of this PR before the next revision.

Since #8533 has now landed and introduced a durable daemon-side terminal report queue with restart/reconnect replay, the recovery mechanism originally proposed here would overlap with infrastructure that now already owns terminal-report durability and delivery.

Rather than keeping a second in-memory recovery loop, I’m going to rebase this PR onto the latest main and reduce it to the remaining safety issue: claim-generation fencing for terminal reports.

The revised design will:

* keep #8533’s durable terminal report queue as the single replay/delivery mechanism;
* remove the separate in-memory pending-report/recovery loop from this PR;
* carry the claim’s dispatched_at through the terminal report and persisted queue record;
* fence /complete and /fail atomically against that expected claim generation;
* ensure a terminal report from an older claim cannot settle a later reclaim of the same task ID;
* treat an explicit stale-generation rejection as a terminal outcome for that old report, rather than retrying it indefinitely or triggering compensation against the newer claim.

The reason for this change is that #8533 now solves the durability/replay part of the original problem, but it also makes generation ownership more important: a persisted terminal report may be replayed much later, after the same task ID has been reclaimed with a new dispatched_at.

So the updated PR will focus only on closing that ownership race, without introducing a competing recovery subsystem.

makesomethingshit:
Final scope note: after #8533 landed, this PR was reduced to claim-generation ownership only. The current head has no second recovery loop or watchdog. #8533 remains the sole durable terminal-report replay mechanism; this PR adds generation-aware identity, the atomic server fence, and a versioned terminal surface so stale reports remain safe across reclaim and mixed-version replicas.

PR #8173, opened 2026-09-08
MUL-7161: fix(dashboard): axis tick labels use foreground color in dark mode

linhdmn:
Chart x/y axis tick (and unit) labels render `fill-muted-foreground` — a mid-grey (`oklch(0.705 …)`) in dark mode, which reads as washed-out grey rather than white on dark surfaces.

This adds a dark-variant override on `ChartContainer` so `.recharts-cartesian-axis-tick text` uses `fill-foreground` (`oklch(0.985` ≈ white) in dark mode only. Light mode is untouched — ticks keep the muted tone.

- Works for all charts: `ChartContainer` is the single styling site for axis tick text (`packages/ui/components/ui/chart.tsx`), and unit strings (`$`, token counts) come from `tickFormatter` inside the same tick `<text>`, so one rule covers x and y.
- The `dark:` class variant is defined (`@custom-variant dark (&:is(.dark *))`) in all consuming apps (web, desktop, docs).
- Specificity: the variant rule carries the extra `:is(.dark *)` class, so it wins over the base rule without order dependence.


multica-eve:
Thanks for the contribution. We are closing this PR because the existing muted axis labels are intentional secondary text, and the dark-mode color already meets WCAG AA contrast requirements. Changing only dark-mode labels to the primary foreground color would make the visual hierarchy inconsistent across themes and all shared charts without evidence of an actual readability failure. If you can reproduce a rendering issue, please share a screenshot, browser and OS, zoom level, and measured contrast so we can reassess it.

PR #8180, opened 2026-09-08
MUL-7164 fix(tasks): retry transient provider rejections

mameikagou:
## What does this PR do?

Claude Code can surface an Anthropic-compatible provider response as authentication_failed / HTTP 403 while the response text says concurrent request limit. Retrying the same prompt with the same credentials and session can succeed, but Multica currently lets the generic 403 classifier win, persists provider_auth_or_access, tells the user to sign in again, and terminates the task.

This PR recognizes that exact transient wire shape before the generic auth fallback, upgrades stale auth reasons reported by older installed daemons, and lets provider capacity/rate-limit failures use the existing bounded retry path. The retry keeps the same session and existing max_attempts ceiling, so the default behavior is one automatic resend rather than an unbounded loop.

## Related Issue

Local incident evidence is included as a verbatim regression fixture; no public issue exists yet.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Make the exact concurrent request limit witness outrank the generic HTTP 403 auth rule.
- Normalize provider_auth_or_access from older daemons to provider_capacity_or_rate_limit for the same exact wire shape.
- Add provider_capacity_or_rate_limit to the bounded, resume-safe task retry allowlist.
- Cover current-daemon classification, mixed-version normalization, same-session retry, and the existing attempt ceiling.

## How to Test

1. Run: cd server && go test ./pkg/taskfailure ./internal/service ./internal/daemon -count=1
2. Run: cd server && go vet ./pkg/taskfailure ./internal/service ./internal/daemon
3. Run: cd server && go test ./... -count=1

All commands pass locally.

## Risks

The capacity/rate-limit bucket now gets one retry under the existing default max_attempts=2. Tasks that explicitly set max_attempts=1 remain non-retryable, autopilot tasks remain excluded, and the retry ceiling prevents loops. The session is preserved because this failure does not poison conversation history.

## Checklist

- [ ] If this PR touches Chinese product copy, I checked it against the product conventions

UI, documentation, runtime inventory, and Chinese copy are unchanged, so those unchecked items are not applicable.

## AI Disclosure

**Prompt / approach:** Started from the local daemon timeline and Claude Code rollout record, converted the exact observed error into a failing classifier test, traced it through daemon/server mixed-version normalization and retry eligibility, then implemented the smallest bounded fix and ran the complete Go test suite.

### DIFF
--- a/server/internal/service/task.go
+++ b/server/internal/service/task.go
@@ -5116,25 +5116,28 @@ func (s *TaskService) FailTask(ctx context.Context, taskID pgtype.UUID, errMsg, 
 // etc.) are intentionally excluded — those are real problems that the user
 // should see, not infrastructure flakiness.
 //
-// The one agent_error.* exception is provider_network: a mid-stream provider
-// disconnect (e.g. Claude Code's "API Error: Connection closed mid-response")
-// is transient infrastructure flakiness, not an agent decision. Unattended
-// issue runs otherwise terminate on it, while interactive chat only survives
-// because the CLI's own in-process retry happens to recover first — so we make
-// the platform retry it directly (MUL-4910). It is resume-safe (not in
+// Two agent_error.* exceptions are transient provider failures:
+// provider_network covers a mid-stream disconnect (e.g. Claude Code's "API
+// Error: Connection closed mid-response"), while
+// provider_capacity_or_rate_limit covers a provider that temporarily refuses a
+// request. Unattended runs otherwise terminate on either one, while manually
+// sending the same prompt again often succeeds. Both are resume-safe (not in
 // resumeUnsafeFailureReason), so the retry child inherits the session and
-// continues the truncated conversation rather than restarting from scratch.
+// continues the conversation rather than restarting from scratch. Capacity
+// keeps the task's existing attempt ceiling: by default that is one automatic
+// resend, the same operation that currently succeeds when the user retries.
 // skill_bundle_unavailable is retryable for the same reason: the agent process
 // never started, so there is nothing to be idempotent about, and every bundle
 // that did download is already cached on disk — a retry resumes from there
 // instead of re-fetching the whole set (MUL-5370).
 var retryableReasons = map[string]bool{
-	string(taskfailure.ReasonRuntimeOffline):         true,
-	string(taskfailure.ReasonRuntimeRecovery):        true,
-	string(taskfailure.ReasonTimeout):                true,
-	"codex_semantic_inactivity":                      true,
-	string(taskfailure.ReasonAgentProviderNetwork):   true,
-	string(taskfailure.ReasonSkillBundleUnavailable): true,
+	string(taskfailure.ReasonRuntimeOffline):                   true,
+	string(taskfailure.ReasonRuntimeRecovery):                  true,
+	string(taskfailure.ReasonTimeout):                          true,
+	"codex_semantic_inactivity":                                true,
+	string(taskfailure.ReasonAgentProviderNetwork):             true,
+	string(taskfailure.ReasonAgentProviderCapacityOrRateLimit): true,
+	string(taskfailure.ReasonSkillBundleUnavailable):           true,
 }
 
 // runtime_offline retries start deferred, not queued: their positive fire_at
--- a/server/internal/service/task_complete_race_test.go
+++ b/server/internal/service/task_complete_race_test.go
@@ -257,6 +257,9 @@ func TestTaskFailureClassifiers(t *testing.T) {
 		// Transient mid-stream provider disconnect (MUL-4910): retryable, and
 		// resume-safe so the retry continues the truncated conversation.
 		{reason: "agent_error.provider_network", wantType: "agent_error", wantResumeOK: true, wantRetry: true},
+		// A provider can report a transient concurrency rejection as HTTP 403.
+		// Retrying the same session is safe and matches a manual resend.
+		{reason: "agent_error.provider_capacity_or_rate_limit", wantType: "agent_error", wantResumeOK: true, wantRetry: true},
 		{reason: "runtime_recovery", wantType: "runtime", wantResumeOK: true, wantRetry: true},
 		{reason: "iteration_limit", wantType: "agent_output", wantResumeOK: false, wantRetry: false},
 		{reason: "api_invalid_request", wantType: "agent_error", wantResumeOK: false, wantRetry: false},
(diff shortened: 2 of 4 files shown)


Bohan-J (changes requested):
Thanks for this — you've found a real problem. Telling a member "sign in again" when their credentials are perfectly fine is a genuinely bad experience, and it's worse than a cosmetic wording issue: it pushes people into destructive troubleshooting (re-authenticating, rotating a key, checking account state) when the real answer was just "wait a moment."

We want the classification half of this PR. We'd like the auto-retry half taken out for now. Details below.

## What we want to keep

Making the concurrent-request rejection outrank the generic 403 auth rule, plus `NormalizeDaemonReason` upgrading the stale label from older daemons. That part is correct and lands every time. The mixed-version normalization in particular is the right call — the call path is real (`FailTask` normalizes before the retry gate reads the reason), and the regression assertion that a plain `API Error: 403 Forbidden` still classifies as auth is exactly the guard we'd want.

**Ask: please split this into its own PR.** We can merge it quickly.

## Why the retry half needs to come out

### 1. It retries with zero delay, but the remedy for this failure is to wait

The provider's own message is "Please wait for your ongoing requests to finish and try again." Our user-facing copy for this bucket says "Please try again in a moment." And `freshSessionMayHelp` in `server/internal/daemon/daemon.go` classifies this reason as one with a defined non-session remedy, listing that remedy verbatim as **"wait, back off"**.

`retryDelayForAttempt` returns `0` for the new bucket, so the retry child is created with a NULL `fire_at` and is immediately claimable. At the instant of failure the requests that consumed the concurrency slot are still in flight — typically other agents' tasks in the same workspace, running for minutes. So the resend most likely hits the same rejection.

For comparison, `provider_network` — a failure that is *less* dependent on waiting — got a bespoke schedule with a ~5s cooldown before its final attempt, precisely so a blip doesn't get two back-to-back shots. This bucket gets none, and the PR doesn't say why.

The practical effect in the common case: the member is no longer misled (good — that's the half we're keeping), one attempt is silently consumed, the task still fails, and they still resend by hand. The retry mostly spends the budget without buying anything.

### 2. The gate is the whole shared reason, not the wire shape you're fixing

The PR argues for one specific wire shape but implements `retryableReasons[ReasonAgentProviderCapacityOrRateLimit] = true`. That makes every member of rule 5 auto-retryable: `429`, `529`, `rate limit`, `overloaded`, `no capacity available`.

That's a much bigger behavior change than the description covers, and its worst case is the one we least want: during a provider-wide overload, every failing task in a workspace re-fires at the same instant, fails again, and adds load to a provider that is already overloaded. No member is helped by that second attempt.

There's a direct precedent. #6747 (MUL-6016) did the same thing for Codex capacity failures and deliberately kept the gate on an exact raw-error predicate rather than the shared reason — the comment there says it explicitly, so that "the shared capacity/rate-limit reason does not broaden ... to unrelated provider failures." It was also opt-in and bounded, with a configurable count and a hard ceiling. That PR was reverted in #7586 without a stated reason, and we need to confirm internally why before re-landing anything in this area.

If we do want an automatic resend here, the shape should be: a delay greater than zero, and a gate narrow enough to cover "your own requests collided" without covering "the provider is down."

### 3. A load-bearing invariant comment is falsified but not updated

`server/pkg/taskfailure/classify.go:28` argues that the loose matching in `httpAuthCodeRe` / `httpQuotaCodeRe` / `httpCapacityCodeRe` is safe because "the auth / quota / capacity buckets these regexes guard are all non-retryable," and that "the only `agent_error.*` reason on ... `retryableReasons` is `provider_network`." Both sentences stop being true with this PR. That comment is the stated reason those regexes are allowed to be loose, so it needs to be updated in the same change.

### 4. Public docs state the opposite behavior

`apps/docs/content/docs/tasks.mdx:113` (and the `.zh` / `.ja` / `.ko` versions) lists `provider_capacity_or_rate_limit` as "Retry later" — i.e. the member retries — while `provider_network` is documented as "Retries automatically." This PR changes exactly that row. The checklist note that documentation is unchanged and therefore not applicable isn't accurate here.

### 5. The new tests pin down the part that's in question

`TestConcurrentRequestLimitRetries` and the new `TestTaskFailureClassifiers` row assert that the whole capacity reason is retryable. If the gate narrows, those assertions have to change with it — worth keeping in mind so we don't lock the open decision into the suite.

## One thing worth fixing in the half we're keeping

The new rule sits after rule 1, whose condition is `contains("token") && contains("limit")`. Your fixture has no "token" in it, so it passes — but a small variation doesn't:

| raw error | classified as |
|---|---|
| `Failed to authenticate. API Error: 403 You've reached your concurrent request limit. ...` | `provider_capacity_or_rate_limit` ✅ |
| `Failed to refresh access token. API Error: 403 You've reached your concurrent request limit.` | `context_overflow` ❌ |

The second lands in a bucket that is both non-retryable *and* resume-unsafe, which is worse than the status quo. Rule 1 catching these is a pre-existing issue rather than something you introduced, but it means the fix only holds for one exact phrasing. Worth either hoisting the witness above rule 1 or adding a variant to the test table so the fragility is visible.

Minor: rule 7's comment asks for new witnesses to be mirrored into the MUL-1949 offline backfill SQL, so in-flight and backfilled rows stay in the same taxonomy. Worth confirming whether that's needed here.

## Summary of what we're asking for

- [ ] Split the classifier fix (`Classify` ordering + `NormalizeDaemonReason`) into its own PR — we'll merge that.
- [ ] Drop `provider_capacity_or_rate_limit` from `retryableReasons` in this PR.
- [ ] If you'd like to pursue the retry separately: give it a non-zero cooldown, and gate it on the specific wire shape rather than the shared reason. We'll also need to resolve the #7586 revert question on our side first, so please don't invest in it until we come back to you on that.
- [ ] While you're in the classifier: consider the `"token"` ordering case above.

Requesting changes on the retry portion only. The classification fix is good work and we'd like it in.

mameikagou:
@Bohan-J, split as requested in #8200.

Following your review:

> We want the classification half of this PR. We'd like the auto-retry half taken out for now.

#8200 contains only the classifier ordering and mixed-version `NormalizeDaemonReason` fix. It also covers the `refresh access token` + `concurrent request limit` counterexample, pins ordinary auth/context cases as unchanged, and explicitly asserts that `provider_capacity_or_rate_limit` remains non-auto-retryable.

I confirmed the MUL-1949 offline backfill SQL is not checked into this repository and documented the exact CASE witness/order needed if historical rows are reprocessed. The auto-retry proposal is intentionally not being pursued until the #7586 revert question is resolved, so this PR is superseded by #8200.

Issue #8184, opened 2026-09-08
[Feature]: Discuss redaction ordering and boundary-safety design (CODI-10 / PR C)


importcpp:
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

Opening the focused design issue requested for PR C in the [review of #8046](https://github.com/multica-ai/multica/pull/8046#issuecomment-5567223297), before reapplying or rewriting the security implementation.

The review publicly confirmed that truncating tool output before server-side redaction can leave the visible prefix of a credential whose full match crosses the cut. This issue discusses that already identified ordering problem; it includes no real credentials, private logs, or operational exploit details.

#8046 was closed without merging because UTF-8 correctness, metadata/UI, and a globally shared security-package change were coupled. PR A (#8181) now isolates UTF-8 correctness. PR C must be independently reviewable and revertible, without depending on PR B's schema or UI.

### Proposed solution (optional)

No implementation is selected yet. The main options to evaluate are:

1. Redact the complete logical output before forming a preview. This avoids losing the suffix required by a full rule, but its CPU/memory cost, placement, and interaction with existing server redaction need evaluation.
2. Use the old proposal's boundary-aware preview with a `partialAtEOF` matcher for each full redaction rule. This creates a permanent coverage obligation: a partial matcher narrower than its full rule can leak a prefix. `TestEverySecretPatternHasPartial` checks presence, not coverage.
3. Explore a shared rule specification or another design that reduces the risk of full/partial recognizers drifting. This is a discussion option, not a claim that a general automatic derivation is already available.

### Questions for maintainers

1. Is the standing per-rule `partialAtEOF` obligation acceptable? What coverage evidence would be required beyond checking that a matcher exists?
2. Would the team prefer full-output redaction before truncation or another design? Where should it run, and how should we handle newer servers with older installed daemons? Once a daemon has discarded a suffix, the server cannot recover it.
3. What performance budget and message-drain/backpressure constraints should gate the design? Expensive synchronous work in `executeAndDrain` can slow draining, and some provider channels drop messages when full.
4. Should the fix remain preview-specific, or may it change shared redaction behavior? What is the expected rollout and independent revert boundary?

### Proposed validation bar, subject to the design decision

- Use synthetic fixtures only; test many cut positions, adversarial inputs, and fuzz cases.
- Preserve the falsifiable oracle comparing exposure against full-text redaction followed by truncation, and its negative control `TestOracleRejectsTrivialTruncation`.
[...]

Issue #8183, opened 2026-09-08
[Feature]: Decide the scope of unknown truncation-status notices (CODI-10 / PR B)


importcpp:
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

This is the second product decision requested before PR B in the [review of #8046](https://github.com/multica-ai/multica/pull/8046#issuecomment-5567223297), specifically finding 3. #8046 was closed without merging. The standalone UTF-8 correction is PR A, #8181; it does not add metadata or these notices.

The original proposal rendered `TruncationUnknownNotice` inside web chat's per-message `TimelineView`. Historical messages have no recorded truncation metadata, and new messages from older installed daemons can also lack it. Rendering a disclaimer for every such turn would fill conversations with repeated notices. The transcript dialog already placed the notice once per transcript.

This describes a problem in the unmerged proposal, not an already deployed warning. The product choice is where and how often to explain unknown completeness without overwhelming users.

### Proposed solution (optional)

Prefer one contextual notice at the conversation level when the loaded conversation contains tool results with unknown status, and one per opened transcript in the transcript viewer. Do not repeat it inside every turn. Mobile should preserve the same semantics.

An alternative is to show the unknown-status notice only in the transcript viewer and keep ordinary chat free of this disclaimer. We would like the team's choice before implementing either direction.

### Questions for maintainers

1. Should ordinary chat show a conversation-level notice, or should the notice be confined to the transcript viewer?
2. If chat shows one, is the intended scope the currently loaded history, and should the notice appear when pagination first loads an unknown-status result?
3. Should it be dismissible? If so, what should dismissal cover: the current view, the conversation, or future visits too?

### Constraints and validation after agreement

- Preserve the review-approved `*bool`, nullable columns with no default, and `text[]` batch handling and explanatory comments: NULL/absent = unknown, false = complete, true = truncated.
- Do not backfill historical NULL to false: the original lengths cannot be reconstructed.
- Known-truncated output still gets its own source-truncation indicator; the conversation notice only explains unknown rows.
- Test multiple historical turns, mixed old/new daemon messages, NULL/false/true, pagination, and notice counts on web/desktop chat, transcript viewer, and mobile. Add dismissal tests only if that behavior is selected.
- Retain rendered localization assertions, including correct `{{name}}` interpolation.

[...]

Issue #8182, opened 2026-09-08
[Feature]: Decide how tool-output truncation notices should be presented (CODI-10 / PR B)


importcpp:
### Deployment type

self-host

### Area

Agents and runtimes

### What do you want and why?

This is a product-design question before implementing PR B from the [review of #8046](https://github.com/multica-ai/multica/pull/8046#issuecomment-5567223297), specifically finding 2. #8046 was closed without merging; the independently scoped UTF-8 fix is now PR A, #8181. This issue does not ask to reopen or merge the original combined implementation.

The original proposal appended an English `[… truncated, N bytes total]` marker to stored output and also rendered a localized badge from structured metadata. The review identified inconsistent presentation:

- A byte-bounded CJK preview remains below the UI's character-based display limit, so both the English marker and localized badge are visible.
- An ASCII preview crosses the display limit, so client-side clipping hides the marker.

Those are problems in the unmerged proposal, not a claim that this metadata UI has shipped. We need one coherent policy that distinguishes source truncation from client-side display clipping, across web/desktop chat, the transcript viewer, and mobile.

### Proposed solution (optional)

Prefer keeping stored output free of presentation notices and rendering a localized source-truncation indicator solely from structured metadata. Client-side "show more" remains a separate indication of display clipping.

The compatibility tradeoff needs an explicit decision: a new daemon can send metadata to an older server that drops unknown fields, or an older client can ignore new fields. Without an in-band marker, those combinations do not gain a source-truncation indication.

### Questions for maintainers

1. Is metadata-only presentation acceptable, or should any in-band notice remain?
2. For mixed versions, is the existing lack of a reliable truncation indication acceptable until server/client upgrades, or is an explicit compatibility fallback required? If required, where should it live and how should duplication be avoided?
3. Should copy/export contain only tool output, or include truncation status in a separate presentation layer?

### Implementation boundaries after agreement

- Keep source truncation and display clipping distinct.
- Preserve the review-approved tri-state contract: unknown (NULL/absent), complete (false), truncated (true); do not label historical unknown data complete.
- Test actual rendered CJK and ASCII previews, localization, copy behavior, and the chosen mixed-version policy across the affected surfaces.
- Use the existing `text-warning` token, not the undefined `text-warning-foreground`.
- PR B depends on PR A (#8181); redaction ordering and provider normalization remain separate.

The companion PR B issue #8183 concerns the granularity of unknown-status notices. Both product decisions should be settled before the metadata/UI implementation resumes.

[...]

PR #8185, opened 2026-09-08
MUL-7169: fix(agent): preserve JSON null in Qwen tool results

importcpp:
## What does this PR do?

Preserve JSON `null` in Qwen tool-result output instead of silently turning it into an empty string.

Qwen's user-event adapter calls `qwenToolResultOutput` for each tool result. The helper previously treated successful `json.Unmarshal` into a string as proof that the input was a JSON string. Go also accepts `null` into a string without error, leaving its zero value unchanged, so an explicit null result became indistinguishable from empty output.

Check the first non-JSON-whitespace byte before decoding. Decode exactly one layer only for JSON strings; pass other values through verbatim. This fixes the live Qwen call path without importing the broader provider/image helper from the original proposal.

## Related Issue

Addresses the Qwen null-parsing finding in the [review of #8046](https://github.com/multica-ai/multica/pull/8046#issuecomment-5567223297).

Independent of PR A (#8181) and the PR B/C design discussions (#8182, #8183, #8184). This PR is based directly on upstream main and does not close the broader CODI-10 work.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/pkg/agent/qwen.go`: only attempt string unwrapping when the first non-whitespace byte is a double quote; preserve null and other non-string JSON bytes.
- `server/pkg/agent/qwen_test.go`: add a 15-case parser matrix for strings, empty input, null, whitespace, escapes, one-layer decoding, objects, arrays, numbers, booleans, and malformed input.
- Add a regression through `handleQwenEvent -> handleQwenUser` that verifies output and call IDs for explicit null, empty string, and missing content.

Scope/risk: explicit null now appears as the text `null`. Empty strings and missing content remain empty, valid JSON strings still decode once, and other payloads retain their bytes. No protocol, schema, UI, redaction, or other-provider changes; no recovery of previously lost null output.

## How to Test

Run from `server/`:

1. Focused regression:
   `../scripts/go-test-with-agent-cli-guard.sh -- go test ./pkg/agent -run '^(TestQwenToolResultOutput|TestHandleQwenEventPreservesNullToolResult)$' -count=1`
2. All Qwen tests with race detection:
   `../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run 'Qwen' -count=1`
3. Build and static analysis:
   `go build ./...` and `go vet ./pkg/agent`.

Regression proof: with the new tests and the original implementation, the null and whitespace-wrapped null cases fail, and the real event-handler regression reports empty output instead of `null`. All focused tests pass after the fix.

Validation passed: focused regression tests, all Qwen tests with `-race` (7.904s), `go build ./...`, `go vet ./pkg/agent`, gofmt, and diff whitespace checks.

Broader-check limitation: `go test -race ./pkg/agent -count=1` is not green on this machine. A separate baseline run, with `pkg/agent` unchanged from upstream main (`2ed2432a2`), also failed in four existing Codex lifecycle tests: `TestCodexCancellationBeforeTurnStartedStopsProcessImmediately`, `TestCodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline`, `TestCodexExecuteTimesOutWhenTurnStopsAfterToolResult`, and `TestCodexExecuteFirstTurnRetryErrorDoesNotSatisfyProgress` (cleanup-duration/result-timeout assertions). Those failures are not being addressed in this Qwen-only PR; no full-suite pass is claimed. A focused rerun of those four tests on this branch reproduced the two cancellation/interrupt cleanup-duration failures; the two result-timeout tests passed when run in that smaller selection.

`make check-worktree` was attempted but stopped before checks because this isolated checkout has no `.env.worktree`. No database/frontend/E2E or real-agent-account smoke test was run; the agent CLI guard prevents default tests from accidentally invoking installed agent CLIs.

## Checklist

UI screenshots, user-facing documentation, new-runtime copy, and Chinese product-copy items are not applicable to this adapter-only correction.

## AI Disclosure

**Prompt / approach:**
Asked to fix the reviewer's Qwen null-parsing finding in a standalone PR. Inspected the live adapter and the original PR's string-type guard, wrote failing parser/event-handler regressions first, applied the guard directly to Qwen's existing helper, and ran guarded Go tests plus build/vet. Kept this separate from truncation metadata, UI decisions, image handling, and redaction design.

## Screenshots (optional)

Not applicable: no UI changes.

### DIFF
--- a/server/pkg/agent/qwen.go
+++ b/server/pkg/agent/qwen.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -389,9 +390,14 @@ func qwenResultUsage(usage *qwenUsage, model string) map[string]TokenUsage {
 }
 
 func qwenToolResultOutput(raw json.RawMessage) string {
-	var text string
-	if json.Unmarshal(raw, &text) == nil {
-		return text
+	// Unmarshal accepts null into a string without error, leaving it empty.
+	// Only unwrap JSON strings; preserve every other result verbatim.
+	trimmed := bytes.TrimLeft(raw, " \t\r\n")
+	if len(trimmed) > 0 && trimmed[0] == '"' {
+		var text string
+		if json.Unmarshal(raw, &text) == nil {
+			return text
+		}
 	}
 	return string(raw)
 }
--- a/server/pkg/agent/qwen_test.go
+++ b/server/pkg/agent/qwen_test.go
@@ -76,6 +76,68 @@ func TestBuildQwenArgsYoloAlwaysPresent(t *testing.T) {
 	}
 }
 
+func TestQwenToolResultOutput(t *testing.T) {
+	t.Parallel()
+	for _, tc := range []struct {
+		name string
+		raw  string
+		want string
+	}{
+		{"missing", "", ""},
+		{"string", `"hello"`, "hello"},
+		{"empty string", `""`, ""},
+		{"quoted null", `"null"`, "null"},
+		{"null", "null", "null"},
+		{"null with whitespace", " \t\r\nnull \t\r\n", " \t\r\nnull \t\r\n"},
+		{"string with whitespace", " \t\r\n\"hello\" \t\r\n", "hello"},
+		{"escaped string", `"line\n\u4e2d\""`, "line\n中\""},
+		{"decode once", `"\"hello\""`, `"hello"`},
+		{"object", `{"ok":true}`, `{"ok":true}`},
+		{"array", `["hello",null]`, `["hello",null]`},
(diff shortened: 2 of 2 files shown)


multica-eve:
Thanks for the contribution. We're closing this because the user-facing problem is not reachable through Qwen Code's supported output contract: both v0.20.0 and the current upstream version define tool-result content as a string or content-block array, and the emitting path produces a string or omits the field. The regression constructs content: null directly, so it demonstrates defensive handling for an out-of-contract event rather than a verified user-facing bug. If a real stream-json capture or an upstream contract change shows that null can be emitted, we're happy to revisit this.

PR #8187, opened 2026-09-08
fix(daemon): retry terminal task reports

vihang:
## Summary

- retry terminal complete/fail callbacks across bounded network, 5xx, timeout, rate-limit, and generic proxy-404 failures
- stop immediately when the daemon API explicitly reports that the task no longer exists
- preserve a successful agent outcome instead of converting delivery failure into a failed task
- add focused retry, exhaustion, and permanent-error tests

## Why

A short backend or same-origin proxy interruption can currently strand an already-finished task in `running`: streamed messages may fail, the one-shot completion callback may receive a transient response, and the fallback fail callback can fail as well. Completion is server-side idempotent, so a bounded retry window safely closes that gap, including the case where the first request committed but its response was lost.

## Validation

- `git diff --check`
- focused Go tests are included; repository CI is the executable validation because Go 1.26.1 is not installed on the authoring host

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math/rand"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


vihang:
Closing as redundant. The retry behavior is already implemented upstream by #3443 (commit 90a737fc7e89768eb5dd57e003690d9786773c29) and is contained in v0.4.41. I will diagnose the deployed daemon/runtime version instead of maintaining a duplicate patch.

PR #8188, opened 2026-09-08
fix(daemon): retry proxy 404 terminal callbacks

vihang:
## Summary

- retry the reverse-proxy-shaped plain-text `404 page not found` response for complete/fail callbacks
- preserve JSON task-not-found 404 responses as permanent
- keep the exception scoped to terminal callbacks so compatibility probes retain their existing behavior
- avoid downgrading a successful result through `/fail` when edge-404 retries are exhausted

## Why

A backend interruption can present as a short 5xx burst followed by the proxy/router plain-text unmatched-route 404. The daemon currently treats that final 404 as permanent, so it stops the retry schedule and falls back from `/complete` to `/fail`. If the same edge state rejects `/fail`, the authoritative task remains running with no durable result.

The backend task handler returns JSON for a real missing task, so the response body distinguishes the transient edge response from the permanent application response.

## Tests

- callback predicate classification
- plain-text edge 404 retries and succeeds
- JSON task-not-found bails after one attempt
- exhausted edge 404 does not fall back to `/fail`

Local Go tooling is unavailable in this environment; repository CI is the executable verification.

### DIFF
--- a/server/internal/daemon/client.go
+++ b/server/internal/daemon/client.go
@@ -542,7 +542,7 @@ func (c *Client) CompleteTask(ctx context.Context, taskID, output, branchName, s
 	if retiredSessionID != "" {
 		body["retired_session_id"] = retiredSessionID
 	}
-	return c.postJSONWithRetry(ctx, fmt.Sprintf("/api/daemon/tasks/%s/complete", taskID), body, nil, defaultTerminalRetrySchedule)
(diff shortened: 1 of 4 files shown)


vihang:
Closing this external contribution. We will carry the terminal-callback resilience fix in our private, release-managed control-plane integration instead of depending on an upstream merge.

PR #8192, opened 2026-09-08
MUL-7176 perf(issues): reuse status catalogs across child completion checks

importcpp:
## What does this PR do?

Child-completion notifications repeatedly resolve custom statuses while checking the changed child, parent, siblings, stage gates, and progress summary. Reuse one lazy `issuestatus.Resolver` per workspace for one notification pass, including all parent groups in a batch. Built-in-only passes still make zero catalog reads; custom-status passes make at most one catalog read per workspace.

The review fixes make this side-effecting path fail closed: a catalog-read error or an unresolved custom key skips the affected parent's notification and wake. The latter also covers a parent row whose newly created status is newer than the cached catalog. The stage algorithm and successfully resolved status behavior remain unchanged; this is deliberately stricter for unknown/deleted keys, not a zero-behavior-change claim.

## Related Issue

Tracks CODI-19 in the contributor's coding workspace. Based directly on `main`, independently of CODI-18 and CODI-20.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/issue_child_done.go`: share workspace-keyed, pass-local resolvers; require successful category resolution before notification side effects. Resolve participating sibling statuses before stage/progress helpers run. Single notifications return on an error; batch notifications skip the affected parent group. Unknown-key errors do not poison valid keys in the same catalog.
- `server/internal/issuestatus/issuestatus.go`: expose the cached read error through `Resolver.Err()`. The existing display-oriented `Effective`/`Name` fallback behavior is unchanged; category validation belongs to the private notification resolver.
- `server/internal/handler/issue_child_done_resolver_test.go` and `server/internal/issuestatus/issuestatus_test.go`: database-backed notification/query-count regressions and resolver error coverage. Include deterministic catalog-then-parent mutation cases for new backlog/done/cancelled statuses in single and batch paths; unknown/deleted keys; sibling/progress checks; parent-group/workspace isolation; and fresh-pass resolution.
- `server/internal/service/builtin_skills/multica-platform/references/issues.md`: document best-effort notification semantics.

Risks and scope:

- Stored status keys and committed status updates are never rewritten or rolled back. Notifications skipped for resolution uncertainty are not automatically replayed; a subsequent pass gets a fresh catalog but does not reconstruct a lost completion event.
- No additional point lookup or catalog refresh on a miss. No global cache, new API, migration, or frontend change. Archived custom statuses remain resolvable.
- The O(n²) batch sibling traversal and the separate transition lookups in `issue.go` remain outside this PR, tracked separately as CODI-23 and CODI-22.

## How to Test

Use `DATABASE_URL` pointing to an isolated, migrated PostgreSQL 17 database. From the repository root:

1. Focused regressions, including both review fixes:

   ```bash
   scripts/go-test-with-agent-cli-guard.sh -- go -C server test -race ./internal/handler ./internal/issuestatus -run 'Test(Child|Stage|Batch.*(Child|Terminal)|Resolver|Effective)' -count=3 -timeout=5m
   ```

2. Full relevant packages:

   ```bash
   scripts/go-test-with-agent-cli-guard.sh -- go -C server test -race ./internal/handler ./internal/issuestatus -count=1 -timeout=5m
   ```

3. Build/vet:

   ```bash
   cd server && go build ./cmd/server/ && go vet ./internal/handler ./internal/issuestatus
   ```

Latest local results (2026-09-09):

- Red/green verification: before the new guard, all six catalog-snapshot regressions produced one wrong notification and one queued parent run; after it, all six produce neither. The focused suite passed three consecutive race-enabled runs, including existing zero/one-read assertions. Build/vet, `gofmt`, and `git diff --check` passed.
- The unfiltered full handler run failed only `TestClaimTasksByRuntime_ClaimPollHintSchedulesNextDeferredTask`: 5074ms versus a 5000ms upper bound. An untouched `b5a7ee1e0` checkout reproduced that assertion in 3/3 focused runs (5076–5084ms). The full relevant packages passed with only that test excluded via `-skip '^TestClaimTasksByRuntime_ClaimPollHintSchedulesNextDeferredTask$'` (handler: 65.861s). No unrelated test or production logic was modified.
- `make check-worktree` stopped because this backend-only checkout has no `.env.worktree`; the full-stack frontend/E2E checks were not run locally. The isolated database lacks optional `pg_bigm`, so its guarded migrations were skipped. No end-to-end latency claim is made.

## Checklist

- [x] I have run tests locally and they pass (focused and filtered full Go suites; unfiltered failure and baseline reproduction documented above)

## AI Disclosure

**Prompt / approach:** Traced the existing child-completion path and reused its request-local status resolver. Followed both reviewer reproductions with database-backed red/green tests, deterministic mutations after catalog materialization, and explicit checks for both notifications and queued parent runs. Kept strict category validation local to the side-effecting path, preserved display fallback behavior, and checked race-enabled regressions in an isolated database. Compared the unrelated timing assertion against an untouched baseline and documented the verification limits.

## Screenshots (optional)

### DIFF
--- a/server/internal/handler/issue_child_done.go
+++ b/server/internal/handler/issue_child_done.go
@@ -82,8 +82,18 @@ func (h *Handler) notifyParentOfChildDone(ctx context.Context, prev, issue db.Is
 	// inherit, so a move into a custom done/cancelled status fires the barrier
 	// exactly like a move into Done or Cancelled. (MUL-6243)
 	effective := h.childStatusResolver(ctx)
-	prevTerminal := isTerminalChildStatus(effective(prev))
-	nowTerminal := isTerminalChildStatus(effective(issue))
+	prevStatus, err := effective(prev)
+	if err != nil {
+		slog.Warn("child done: failed to resolve previous child status", "error", err, "child_id", uuidToString(issue.ID))
+		return
+	}
+	nowStatus, err := effective(issue)
+	if err != nil {
+		slog.Warn("child done: failed to resolve child status", "error", err, "child_id", uuidToString(issue.ID))
+		return
+	}
+	prevTerminal := isTerminalChildStatus(prevStatus)
+	nowTerminal := isTerminalChildStatus(nowStatus)
 	if prevTerminal || !nowTerminal {
 		return
 	}
@@ -98,7 +108,11 @@ func (h *Handler) notifyParentOfChildDone(ctx context.Context, prev, issue db.Is
 	// Custom statuses inherit the canonical status they name, so a custom
 	// terminal status closes this out and a custom backlog status parks it,
 	// exactly like Done/Cancelled and Backlog do. (MUL-6243)
-	parentStatus := effective(parent)
+	parentStatus, err := effective(parent)
+	if err != nil {
+		slog.Warn("child done: failed to resolve parent status", "error", err, "parent_id", uuidToString(parent.ID))
+		return
+	}
 	if parentStatus == "done" || parentStatus == "cancelled" {
 		return
 	}
@@ -133,7 +147,11 @@ func (h *Handler) notifyParentOfChildDone(ctx context.Context, prev, issue db.Is
 			"parent_id", uuidToString(parent.ID))
 		return
 	}
-	isTerminal := func(c db.Issue) bool { return isTerminalChildStatus(effective(c)) }
+	isTerminal, err := resolveTerminalChildren(children, effective)
+	if err != nil {
+		slog.Warn("child done: failed to resolve sibling statuses", "error", err, "parent_id", uuidToString(parent.ID))
+		return
+	}
 	if !stageBarrierClosed(children, issue, isTerminal) {
 		return
 	}
@@ -191,7 +209,6 @@ func (h *Handler) notifyParentsOfBatchChildDone(ctx context.Context, completed [
 	}
 
 	effective := h.childStatusResolver(ctx)
-	isTerminal := func(c db.Issue) bool { return isTerminalChildStatus(effective(c)) }
 	for _, g := range groups {
 		parent, err := h.Queries.GetIssue(ctx, g.parentID)
 		if err != nil {
@@ -200,7 +217,11 @@ func (h *Handler) notifyParentsOfBatchChildDone(ctx context.Context, completed [
 			continue
 		}
 		// Same parent guards as the single path (see notifyParentOfChildDone).
-		parentStatus := effective(parent)
+		parentStatus, err := effective(parent)
+		if err != nil {
+			slog.Warn("batch child done: failed to resolve parent status", "error", err, "parent_id", uuidToString(parent.ID))
+			continue
+		}
 		if parentStatus == "done" || parentStatus == "cancelled" {
 			continue
 		}
@@ -218,6 +239,11 @@ func (h *Handler) notifyParentsOfBatchChildDone(ctx context.Context, completed [
 			continue
 		}
 
+		isTerminal, err := resolveTerminalChildren(children, effective)
+		if err != nil {
+			slog.Warn("batch child done: failed to resolve sibling statuses", "error", err, "parent_id", uuidToString(parent.ID))
+			continue
+		}
 		batch := len(g.children) > 1
 		if !siblingsAreStaged(children) {
 			// Unstaged: one implicit stage. Fire once iff every child is terminal
@@ -370,19 +396,40 @@ func isTerminalChildStatus(status string) bool {
 //
 // Resolve without rewriting the issue rows, which also supply the original
 // user-selected status to downstream rendering. Built-in keys need no I/O.
-func (h *Handler) childStatusResolver(ctx context.Context) func(db.Issue) string {
+func (h *Handler) childStatusResolver(ctx context.Context) func(db.Issue) (string, error) {
 	resolvers := make(map[pgtype.UUID]*issuestatus.Resolver)
-	return func(c db.Issue) string {
+	return func(c db.Issue) (string, error) {
 		if issuestatus.IsBuiltIn(c.Status) {
-			return c.Status
+			return c.Status, nil
 		}
 		resolver := resolvers[c.WorkspaceID]
 		if resolver == nil {
 			resolver = issuestatus.NewResolver(c.WorkspaceID)
 			resolvers[c.WorkspaceID] = resolver
 		}
-		return resolver.Effective(ctx, h.issueStatusCatalog(), c.Status)
+		status := resolver.Effective(ctx, h.issueStatusCatalog(), c.Status)
+		return status, resolver.Err()
 	}
+}
+
+// resolveTerminalChildren checks every status needed by the stage barrier and
+// progress summary before either can produce a notification. The returned
+// predicate only reads this snapshot, so a late catalog failure cannot be
+// hidden by the bool-only stage helpers. Stored status keys stay untouched.
+func resolveTerminalChildren(children []db.Issue, effective func(db.Issue) (string, error)) (func(db.Issue) bool, error) {
+	terminal := make(map[pgtype.UUID]bool, len(children))
+	staged := siblingsAreStaged(children)
+	for _, child := range children {
+		if staged && !child.Stage.Valid {
+			continue // Neither stage helper considers unstaged siblings.
+		}
+		status, err := effective(child)
+		if err != nil {
+			return nil, fmt.Errorf("resolve child %s status %q: %w", uuidToString(child.ID), child.Status, err)
+		}
+		terminal[child.ID] = isTerminalChildStatus(status)
+	}
+	return func(child db.Issue) bool { return terminal[child.ID] }, nil
 }
 
 // siblingsAreStaged reports whether any child in the set carries an explicit
--- a/server/internal/handler/issue_child_done.go
+++ b/server/internal/handler/issue_child_done.go
@@ -396,6 +396,9 @@ func isTerminalChildStatus(status string) bool {
 //
 // Resolve without rewriting the issue rows, which also supply the original
 // user-selected status to downstream rendering. Built-in keys need no I/O.
+// Unlike display-oriented Resolver callers, this side-effecting path must
+// reject unresolved custom keys: parent or sibling rows can be newer than the
+// catalog snapshot. A miss must not bypass a parked/terminal parent's guard.
 func (h *Handler) childStatusResolver(ctx context.Context) func(db.Issue) (string, error) {
 	resolvers := make(map[pgtype.UUID]*issuestatus.Resolver)
 	return func(c db.Issue) (string, error) {
@@ -408,7 +411,13 @@ func (h *Handler) childStatusResolver(ctx context.Context) func(db.Issue) (strin
 			resolvers[c.WorkspaceID] = resolver
 		}
 		status := resolver.Effective(ctx, h.issueStatusCatalog(), c.Status)
-		return status, resolver.Err()
+		if err := resolver.Err(); err != nil {
+			return "", err
+		}
+		if !issuestatus.IsCategory(status) {
+			return "", fmt.Errorf("unresolved status %q in workspace %s", c.Status, uuidToString(c.WorkspaceID))
+		}
+		return status, nil
 	}
 }
 
--- a/server/internal/handler/issue_child_done.go
+++ b/server/internal/handler/issue_child_done.go
@@ -81,8 +81,9 @@ func (h *Handler) notifyParentOfChildDone(ctx context.Context, prev, issue db.Is
 	// Both sides of the transition are resolved to the canonical status they
 	// inherit, so a move into a custom done/cancelled status fires the barrier
 	// exactly like a move into Done or Cancelled. (MUL-6243)
-	prevTerminal := isTerminalChildStatus(issuestatus.Effective(ctx, h.Queries, prev.WorkspaceID, prev.Status))
-	nowTerminal := isTerminalChildStatus(issuestatus.Effective(ctx, h.Queries, issue.WorkspaceID, issue.Status))
+	effective := h.childStatusResolver(ctx)
+	prevTerminal := isTerminalChildStatus(effective(prev))
+	nowTerminal := isTerminalChildStatus(effective(issue))
 	if prevTerminal || !nowTerminal {
 		return
 	}
@@ -97,7 +98,7 @@ func (h *Handler) notifyParentOfChildDone(ctx context.Context, prev, issue db.Is
 	// Custom statuses inherit the canonical status they name, so a custom
 	// terminal status closes this out and a custom backlog status parks it,
 	// exactly like Done/Cancelled and Backlog do. (MUL-6243)
-	parentStatus := issuestatus.Effective(ctx, h.Queries, parent.WorkspaceID, parent.Status)
+	parentStatus := effective(parent)
 	if parentStatus == "done" || parentStatus == "cancelled" {
 		return
 	}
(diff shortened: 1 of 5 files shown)


Bohan-J (changes requested):
Thanks for this — the duplicate catalog reads are real, worth removing, and reusing the existing `issuestatus.Resolver` for the span of one notification pass is the right shape for the fix. I reproduced the counts on a database-backed run against baseline `b5a7ee1e0`: one parent, 100 same-stage children, parent and children all on custom statuses.

| notification pass | before | after |
| --- | ---: | ---: |
| last child completes | 202 status queries | 1 |
| batch completing all 100 children | 10,101 | 1 |
| the same shapes with built-in statuses only | 0 | 0 |

Both revisions produce the same single, correct `Stage 1: 100/100 done` notification. Requesting changes on one issue below.

## Blocking: a cached catalog-read failure lets a parked parent be woken

`Resolver.load` sets `loaded` *before* the read, so a failed read is cached as "loaded with nil maps" for the resolver's whole lifetime (`server/internal/issuestatus/issuestatus.go:464-482`). That was safe while a `Resolver` served one list endpoint's rows. Now that a single resolver is shared across the child-transition guards *and* the parent guard, one transient failure while resolving the **child's** status leaves the **parent's** custom status unresolved at `server/internal/handler/issue_child_done.go:101`: `parked` never resolves to `backlog`, so the "a parent parked in backlog stays inert" guard (#4320 / MUL-3497) is skipped, and the notification plus the parent-assignee wake both fire.

On `main` each `issuestatus.Effective` call is an independent point query, so a single failure only affects the call that failed — the parent lookup still succeeds and the parent is correctly recognized as backlog.

Repro: parent in custom backlog status `parked`; child moves from custom `working` into built-in `done`; the first status read fails and every later read succeeds.

```go
func TestPR8192ReviewTransientFailure(t *testing.T) {
	ctx := context.Background()
	ws := dbfx.Workspace(t, "Transient catalog failure", "review-transient")
	fx := testutil.New(testPool, ws, testUserID)
	for key, category := range map[string]string{"working": "in_progress", "parked": "backlog"} {
		fx.Insert(t, "issue_status", testutil.Cols{"workspace_id": ws, "key": key, "name": key, "category": category, "color": "#123456"})
	}
	parentID := fx.Issue(t, "Deliberately parked parent", testutil.Cols{"status": "parked"})
	fx.Cleanup(t, "DELETE FROM comment WHERE issue_id = $1", parentID)
	id := fx.Issue(t, "Completed child", testutil.Cols{"parent_issue_id": parentID, "status": "done"})
	last, err := testHandler.Queries.GetIssue(ctx, parseUUID(id))
	if err != nil {
		t.Fatal(err)
	}
	prev := last
	prev.Status = "working"

	// counter fails only the FIRST status read (point or catalog), then passes through.
	counter := &pr8192ReviewDB{DBTX: testPool, failFirst: true}
	h := *testHandler
	h.Queries = db.New(counter)
	h.IssueStatusCatalog = nil

	h.notifyParentOfChildDone(ctx, prev, last)

	if got := countSystemCommentsOn(t, parentID); got != 0 {
		t.Fatalf("a transient child-status read failure must not bypass the parked-parent guard")
	}
}
```

- `main`: 0 system comments — the parent is recognized as backlog and stays inert.
- this branch: 1 comment on the parked parent, plus the existing wake path when that parent is assigned to an agent or squad.

The new `unavailable` cases don't cover this combination: they give the child a custom status too, so the pass returns at the terminal-transition guard before it ever reaches the parent check.

Suggested direction: let a failed catalog read abort the pass explicitly instead of being reused as "unresolved" — surface the load failure from the resolver, return early in `notifyParentOfChildDone`, and skip that parent's group in the batch path — while keeping the reuse for successful reads. A regression case for `custom backlog parent + built-in terminal child status + one failed read` would pin it.

## Non-blocking

- Most of the batch win (10,000 of the 10,101) comes from `stageBarrierClosed` scanning every sibling once per completed child, i.e. an O(n²) traversal. This PR makes each probe free but leaves the traversal; fine to leave here, worth a follow-up if batch sizes grow.
- `server/internal/handler/issue.go:4379` still resolves both sides of each row's transition one point query at a time during the same batch request. Out of scope for this PR, but it is the same pattern on the same request path and would be a natural next step.

Otherwise the approach looks right to me: workspace-keyed, lazy on the first custom key, discarded at the end of the pass, archived statuses still resolvable, and stored status keys never rewritten.

importcpp:
@Bohan-J Thank you for the thorough review, query-count verification, and concrete reproduction. You were right: the original tests missed the combination where a failed child-status lookup lets a custom-backlog parent bypass its guard.

### Fix in 2440753ef9ec60498c2560d1ef08a8f7b6ecb023

I've followed your suggested approach:

- `Resolver.Err()` exposes the cached catalog-read failure without changing the existing `Effective`/`Name` fallback behavior for other callers. The single notification path returns early on a resolution error; the batch path skips affected parent groups.
- Sibling statuses needed by the stage checks and progress summary are resolved before either can create a notification. Those helpers then use an in-memory terminal-status snapshot, so a later lookup failure cannot silently pass through a bool-only helper and enqueue a parent run.
- Successful reads remain shared per workspace for the notification pass. Failed reads do not affect other workspaces or built-in-only groups, and a subsequent pass gets a fresh resolver. This does not add automatic notification retries or roll back an already-committed status change.

Added database-backed regressions for your parked-parent reproduction, cancellation, terminal-parent protection, sibling/progress-summary failures, batch failure isolation, and next-pass recovery. They assert both no notification and no queued parent run on failure. The existing built-in zero-read/custom one-read assertions still pass. Focused regressions passed three consecutive `-race` runs, and build/vet passed. The latest CI's **backend-tests and backend checks both passed**.

The O(n²) stage traversal and the separate transition lookups in `issue.go` remain unchanged for follow-up, as suggested.

### Could you also take a look at a suspected flaky Windows test?

The only failed check in [the latest run](https://github.com/multica-ai/multica/actions/runs/34321367597) is `cursor-background-lifecycle (windows-latest)`, specifically `TestCursorBackgroundLifecycle/multiple`. I investigated it and suspect an intermittent Windows process-ownership failure independent of this patch:

- The [failing log](https://github.com/multica-ai/multica/actions/runs/34321367597/job/102368568591) reports `cannot own Cursor background shell ... Access is denied` about 0.12 seconds after startup, followed by `status="timeout"` at the test's 10-second context deadline. This looks different from a runner simply being too slow.
- The fixture re-executes the test binary as a fake Cursor, launches two real background shells with 30-second child lifetimes, and expects an explicit interrupt to terminate them. Its `ready` event means the launch events were processed, not that both shells were successfully claimed. `InterruptBackgroundTools()` returning `true` only confirms that at least one tracked tool was cleaned up. If one capture fails, the other can be cleaned up successfully while the unclaimed shell keeps the fake Cursor waiting until the deadline. This agent-only test does not run the daemon idle watchdog that the fallback log refers to.
- The same subtest passed in **0.21 seconds** on both [current main](https://github.com/multica-ai/multica/actions/runs/34319858877/job/102363904029) and [this PR's previous run](https://github.com/multica-ai/multica/actions/runs/34287950875/job/102267803486). The runner image and Go versions match. I verified that current main and the failing merge revision have identical `pkg/agent` and in-repository test dependency trees; the previous and current PR runs also have identical agent code, module files, and CI workflow. This test does not depend on the handler/issuestatus packages changed here. Linux/macOS passed as well.

The remaining uncertainty is where the access-denied error originates: the capture path returns several Windows API errors without identifying the operation or PID. I haven't established whether this is an environmental transient or a real process-capture race, so I wouldn't dismiss it as a harmless test flake or mask it by extending the timeout/skipping the test.

Would you mind rerunning the failed Windows job and taking a look when convenient? My current upstream access doesn't permit rerunning Actions. If it recurs, adding operation/PID details and repeating the focused test on Windows would help distinguish a fixture/environment issue from an ownership bug.

Thanks again — I'd appreciate another look at the blocking fix when you have time.

Bohan-J (changes requested):
Thanks for the quick turnaround — `2440753ef` fixes the blocker from the last round. `Resolver.Err()` makes the cached read failure explicit, `resolveTerminalChildren` resolves the whole sibling set up front so a late failure cannot hide behind the bool-only stage helpers, and both the single and the batch path bail out instead of proceeding on an unresolved status. My earlier repro passes on this branch, the query counts still hold (one catalog read per pass with custom statuses, zero with built-ins only), and CI is green.

One more case in the same guard, so still requesting changes.

## Blocking: a successfully loaded catalog snapshot that predates the parent row

`Err()` only reports whether the catalog *read* failed. When the read succeeded but the catalog does not contain a key, `Resolver.Effective` returns the raw key with a nil error (`server/internal/handler/issue_child_done.go:410`), and `server/internal/handler/issue_child_done.go:111` then compares that raw key against `done` / `cancelled` / `backlog`. An unresolved custom status therefore reads as "none of the guarded categories" and the pass proceeds.

The ordering is what makes this reachable: the catalog snapshot is taken while resolving the *child's* status, which happens before `GetIssue(parent)`. So the parent row can be newer than the snapshot it is resolved against. On `main` every `Effective` call is a fresh point query issued after the row was read, so a row and its resolution are always consistent with each other.

Repro — every read succeeds; the mutation commits right after the first status read returns, i.e. before the parent row is loaded:

```go
queries := &pr8192CatalogMutationDB{DBTX: testPool, mutate: func() {
	fx.Insert(t, "issue_status", testutil.Cols{"workspace_id": ws, "key": "new_parked", "name": "New parked", "category": "backlog", "color": "#123456"})
	fx.Exec(t, "UPDATE issue SET status = 'new_parked' WHERE id = $1", parentID)
}}
h := *testHandler
h.Queries = db.New(queries)
h.IssueStatusCatalog = nil

h.notifyParentOfChildDone(ctx, previous, child)

if countSystemCommentsOn(t, parentID) != 0 || countPendingTasksForAgent(t, parentID, agentID) != 0 {
	t.Fatalf("a parent parked after the catalog snapshot must not receive a notification or queued run")
}
```

| | wrong notifications | queued parent runs |
| --- | ---: | ---: |
| `b5a7ee1e0` | 0 | 0 |
| `2440753ef` | 1 | 1 |

On scope, to be fair to the change: this needs a custom status to be created *and* the parent parked into it within one notification pass, so it is not something every completion hits. What makes it worth fixing rather than accepting is that the parent is already parked before its row is read — this is not the pre-existing "parent changes after the guard ran" race, it is the guard applying stale category data to a row it has already loaded.

### Two ways out

I would take the first:

1. **Fail closed in this path.** Treat "custom key that did not resolve to a built-in category" exactly like a read error and skip that parent's notification. The guard's real question is "do I positively know this parent may be woken?", and on a side-effecting path the answer to "I don't know" should be "don't fire". It costs no extra query and keeps the one-read-per-pass win intact. Worth naming explicitly: this is slightly more than restoring parity — on `main` a genuinely unknown or deleted key bypasses this same guard — so it also closes that older hole. Leave the existing lenient fallback alone for the display-oriented `Resolver` callers.
2. **Re-check on a miss.** On a catalog miss, fall back to a single point read for that one key before deciding. Smaller behavioral delta and it fixes the staleness precisely, but it reintroduces a per-key query in the miss case, so it needs a guard against a sibling set full of unknown keys undoing the optimization.

Either way, a regression case for `parent parked into a custom backlog status created after the snapshot` would pin it.

Everything else from the last round still looks right to me. The two non-blocking notes I left then — the O(n²) stage traversal and `server/internal/handler/issue.go:4379` — remain out of scope here.

importcpp:
@Bohan-J Thank you for catching this and for distinguishing the stale-snapshot case from the pre-existing race after the parent guard. I reproduced it on `2440753ef` and followed your first option in d852263c3c6c018393b1e4e4dcebc1fd573f08c2.

The private `childStatusResolver` now requires a valid category after checking `Resolver.Err()`. A successful catalog read that cannot resolve a custom key returns an error to the existing single/batch guards, so the affected parent's notification and wake are skipped. There is no point-read fallback or catalog refresh on a miss. The public display-oriented `Resolver.Effective`/`Name` behavior remains unchanged, and a missing key does not poison other valid keys in the same workspace's snapshot.

Added deterministic database-backed regressions that create a status and update the parent after the catalog rows have been read, but before the affected parent row is loaded. They cover new backlog, done, and cancelled categories in both single and batch paths; the batch case primes the catalog through an earlier parent group. All six cases produced **one wrong notification and one queued parent run** before the guard, and **neither** after it. Additional coverage pins genuinely unknown/deleted keys, previous/current child and sibling/progress checks, isolation of unaffected parent groups and workspaces, and resolution with a fresh pass. Existing archived-status and zero/one-catalog-read assertions still pass.

As you noted, this intentionally tightens behavior for genuinely unknown/deleted keys too. The PR description and built-in issue reference now document that distinction: committed status updates are preserved, but skipped notifications are not automatically replayed. The O(n²) traversal and separate `issue.go` transition lookups remain out of scope, tracked as CODI-23 and CODI-22 respectively.

Validation: the focused suite passed three consecutive `-race` runs; build/vet and formatting checks passed. The unfiltered full handler suite hit only the existing `TestClaimTasksByRuntime_ClaimPollHintSchedulesNextDeferredTask` timing assertion (5074ms versus a 5000ms bound). I reproduced it 3/3 times on untouched `b5a7ee1e0` (5076–5084ms); the full relevant packages passed with only that case excluded, without changing the test. The full-stack check remains unavailable locally because this checkout has no `.env.worktree`.

I also saw that the previous Windows check passed on rerun — thank you for taking another look. [This new commit's CI run](https://github.com/multica-ai/multica/actions/runs/34329778737) is now green too, including `backend-tests` and the Windows lifecycle check. I'd appreciate another review when convenient.

Bohan-J (approved):
Approving. `d852263c3` closes the last blocker, and I re-checked the whole change as it will land rather than just the delta.

**Both blockers are fixed.** `Resolver.Err()` separates a failed catalog read from a miss; `childStatusResolver` now also rejects any custom key that does not resolve to a real category (`server/internal/handler/issue_child_done.go:411-420`), so a parent row newer than the snapshot can no longer slip past the parked/terminal guards; `resolveTerminalChildren` resolves the sibling set up front so neither stage helper can swallow a late failure; and the batch path skips only the affected parent group. Both of my repros now come out at 0 notifications and 0 queued runs, and the query counts are untouched — one catalog read per pass with custom statuses, zero with built-ins only.

One thing I wanted to establish before agreeing to fail closed here, since failing closed on a notification path can silently swallow real work: **how reachable is an unresolved key in normal operation?** It is not. Custom statuses are only ever archived, never hard-deleted — `ArchiveIssueStatusEntry` retires a key while issues keep sitting on it, and the resolver loads with `IncludeArchived: true`, so archived statuses still resolve to their category. The only hard delete is `DeleteIssueStatusEntriesForWorkspace` during workspace teardown, where these notifications are moot anyway. So "unresolved" really does mean the create-after-snapshot race or genuine corruption, not an ordinary admin workflow. That is what makes fail-closed the right trade here rather than a source of quietly dropped notifications, and it is worth having in the record for whoever revisits this.

Two non-blocking observations, neither worth holding the merge:

- An unresolvable key on a *higher* stage now skips the notification for an already-closed lower stage, where `main` would have fired it and just counted that sibling as non-terminal in the summary. Same narrow window as the bug being fixed, and it fails in the safe direction — a missed notification rather than a wrongly woken parent — which is the trade this guard exists to make. The best-effort semantics you added to `references/issues.md` cover it.
- `resolveTerminalChildren` skips unstaged siblings in a staged set, and the returned predicate defaults to `false` for anything it did not resolve. That is correct today because `stageBarrierClosed` and `stageProgressSummary` independently skip unstaged children too, and your comment says exactly that — but it is an invariant held by convention rather than by the types, so a future change to either helper needs to remember it.

Verified on `d852263c3`: `go vet ./internal/handler ./internal/issuestatus` and `gofmt` clean locally, CI 17/17 green. I did not re-run the database-backed suites myself — no isolated database in this environment — so I am relying on CI plus your `-race` runs for those.

Thanks for sticking with this through three rounds. Both fixes landed as real fixes rather than test adjustments, the regression cases you added pin the exact orderings that were broken, and updating `references/issues.md` with the best-effort semantics was the right call — that behavior is the kind that is invisible until someone is debugging a notification that never arrived.

Issue #6144, opened 2026-07-29
[Feature]: Reduce Agent Startup Prompt Size with Lazy Loading, Distillates, and Caching


maxivillus:
### Deployment type

self-host

### What do you want and why?

# Feature Request: Reduce Agent Startup Prompt Size with Lazy Loading, Distillates, and Caching

## Summary

The first agent invocation in Multica can spend a significant number of tokens on startup context: a large `AGENTS.md`/runtime block, skill and plugin lists, tool schemas, workflow rules, developer/system instructions, and mandatory issue/comment reads. For simple tasks, such as “who are you / what model are you?”, most of that context is not needed.

This feature request proposes replacing the “load all important context up front” model with:

`minimal stable kernel + embedded distillate + content-addressed index + lazy body fetch + stable prompt-prefix caching + SHA-backed compression gate`.

The goal is to reduce cold-start prompt size, improve warm-start cache hit rate, and preserve Multica workflow guarantees.

## Observed Problem

In the test run, full skill bodies were mostly unnecessary. The agent only loaded a full skill after a routing decision, while most of the initial token cost came from:

- the large Multica runtime / `AGENTS.md` block;
- Codex system/developer instructions;
- available tool schemas;
- recommended plugin list;
- available skill list;
- mandatory `multica issue get`, `metadata list`, and `comment list` outputs;
- the selected `SKILL.md`, once the agent decided it was applicable.

For a simple user-facing answer, the agent only needs a small core: identity, event type, required issue/comment reads, safe final delivery via `--content-file`, and background-task rules.

### 1. Stable Prompt Prefix

Put only byte-stable context at the beginning of the prompt:

- agent identity and execution surface;
- tiny Multica delivery kernel;
- embedded distillate of mandatory workflow/runtime rules;
- sorted compact index of runtime docs, skills, tools, and plugins.

Example index entry:

```yaml
- name: multica-working-on-issues
  trigger: "working on a Multica issue after runtime trigger context"
  locator: "skill://multica-working-on-issues"
  source_sha256: "..."
  updated_at: "2026-07-29T..."
  token_size: 3559
  priority: normal
```

The prefix should be byte-for-byte stable. All volatile run-specific data should be appended after it.

### 2. Volatile Tail

Place run-specific data at the end:

- issue id;
- trigger comment id;
- run type: assignment, comment, autopilot, etc.;
- minimal event envelope;
- task-specific context.

Issue/comment changes should not invalidate the stable prompt prefix.

[...]

PR #8196, opened 2026-09-09
MUL-7182 feat(cli): add `issue get --compact` for agent reads

ii-jj:
## What does this PR do?

Adds `--compact` to `multica issue get`, the opt-in `multica issue comment list` has carried since #5999 / MUL-5442.

`issue get` is step 1 of every issue turn the runtime brief prescribes, so its response is on the critical path of every run — and unlike the comment reads it had no way to shed the bookkeeping half of its payload. On an issue with a screenshot thread the attachment array alone is the largest non-content field in the response, and none of it is actionable for the caller the flag is aimed at.

**Dropped:** `workspace_id` and `number` (the caller already addressed this issue, and `identifier` is the handle used everywhere else in the CLI), `position` and `revision` (board ordering and optimistic-concurrency state), `status_category` when it only repeats `status`, `updated_at` when it only repeats `created_at`, null-valued fields, and empty arrays.

**Attachments** are slimmed per record rather than per response, which is where the payload actually is. Of the three URLs on each record only `markdown_url` survives: the API contracts it as durable and persistable, and it is what an agent embeds when it references a file in a comment. `url` and `download_url` are render-time values for a browser — and the runtime brief already tells agents to fetch attachments through `multica attachment`, never by opening a Multica resource URL, so for this caller they are bytes that exist only to be ignored.

**Deliberately NOT dropped:**

- `metadata` and `properties` when empty. The brief tells agents an empty `{}` is a normal answer, so pruning the key would turn "no metadata" into "this response did not say".
- Zero values. `stage: 0` and `priority: "none"` are answers; null pruning must never generalize to zero pruning — the mistake #6546's review caught in the comment path.

The default output is byte-for-byte untouched, and a test pins that.

### Measured impact

Realistic issue (2 KB description, pretty-printed JSON as the CLI emits it):

| attachments | before | after | saved |
| ---: | ---: | ---: | ---: |
| 0 | 3009 B (~752 tok) | 2692 B (~673 tok) | **-11%** |
| 3 | 5504 B (~1376 tok) | 3957 B (~989 tok) | **-28%** |
| 6 | 7997 B (~1999 tok) | 5199 B (~1299 tok) | **-35%** |
| 21 | 20506 B (~5126 tok) | 11442 B (~2860 tok) | **-44%** |

21 attachments is the shape #5999 measured on a production issue. Note this is the *residual* after MUL-5372 — the CLI already declares `stable_attachment_urls`, so the ~800-char CloudFront-signed URLs are long gone; what is left is field noise, which is smaller but paid on every single turn.

### Thinking path

1. #5999 established that CLI JSON payloads are a primary driver of agent context exhaustion, and closed with two fixes: `stable_attachment_urls` on the server and `--compact` on `issue comment list`.
2. `issue get` got neither, yet the brief's own workflow makes it mandatory on every turn — so the one read no agent can skip is the one read with no way to shrink.
3. The rules `compactComments` already encodes are the right rules; the issue record just needs its own echoed-identity list, and attachments need per-record treatment neither path had.
4. The flag is only worth having if the agent knows it exists, so the brief's `## Available Commands` line advertises it — one line, and a test pins that it stays advertised.

## Related Issue

Follow-up to #5999 (`issue get` half). Also reduces the startup/turn payload #6144 is about.

## Type of Change

## Changes Made

- `server/cmd/multica/cmd_issue.go`
  - `--compact` flag on `issueGetCmd`, applied in `runIssueGet` **after** `--resolve-properties` so the two compose (resolve rewrites `properties` in place; pruning first would hide the key it resolves).
  - New `compactIssue`, and `compactComments` refactored onto the same three shared helpers: `compactRedundantUpdatedAt`, `compactEmptyValues`, `compactAttachments`. `compactComments`' external behaviour is unchanged except that it now slims attachments too — the same defect, the same flag.
- `server/internal/daemon/execenv/runtime_config_sections.go` — the brief's `issue get` line advertises the flag.
- Tests: `server/cmd/multica/cmd_issue_get_compact_test.go` (new), plus the two brief-content assertions updated for the new line.

## How to Test

```bash
cd server
go test ./cmd/multica/ ./internal/daemon/execenv/ ./internal/daemon/
go vet ./cmd/multica/ ./internal/daemon/execenv/
```

All three packages pass locally.

Manually:

```bash
multica issue get <id> --output json | wc -c
multica issue get <id> --compact --output json | wc -c
```

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — n/a
- [x] I have updated relevant documentation to reflect my changes (the runtime brief; `references/issues.md` documents contracts the brief does not encode, and a flag the brief now states is not one of them)

## AI Disclosure

**Prompt / approach:** Started from "why is Multica's runtime context so large, and what is actually optimizable" — measured the runtime brief section by section with a throwaway Go test (14.6 KB / ~3.7 k tok for the issue kind, already slim after MUL-3560/4297), then walked the per-turn cost outward until the CLI reads showed up as the part with no lever. Read #5999 for the prior art, confirmed `stable_attachment_urls` had already landed so the remaining cost was field noise rather than signed URLs, and sized the change against a realistic fixture before writing it.


multica-eve:
Thank you for the thoughtful contribution and the tests. After reviewing the expected impact, we have decided not to move forward with this change.

The remaining savings are small for typical issues after the existing server-side attachment URL handling, while maintaining a command-specific list of fields to omit introduces an ongoing compatibility risk: a field that seems redundant today may be needed by future agent workflows. We also do not yet have production data showing enough benefit to justify that new contract.

If reading payload becomes a measurable issue, we would prefer to address it through a server-defined agent-oriented view or projection with explicit compatibility guarantees rather than expanding client-side blacklists.

We will close this PR for now, but we appreciate the work and the careful implementation.

Issue #5653, opened 2026-07-20
[Feature]: User-controlled context reset (threshold-based) + external spec-file continuity


ShaoSK:
> Related, but not covered by, existing issues: context-usage **observability** is already proposed in #2038 and #4984. This issue is specifically about **user-controlled context *reset*** and how work continuity survives that reset — a workflow angle those issues don't cover.

## Deployment type

Official App

### Background — how I work with a single coding agent today

When I use a coding agent (e.g. Claude Code) directly, my standard workflow is:

1. I keep an eye on the session's context-window utilization.
2. Once it reaches ~30%, I deliberately **restart a fresh session** — a full context reset, not a compaction — because response quality degrades well before the hard limit.
3. Project progress does **not** live in the agent's conversation memory. It lives in **local spec files** (open-spec style: plan / tasks / decisions written to disk). Each fresh session re-reads those files and continues exactly where the previous one left off.

So for me, a large context window is not something to "fill up" — I actively want short, clean sessions, with continuity handled by external files rather than by an ever-growing transcript.

### The problem on Multica today

In the current version I cannot reproduce this workflow:

- **No context-usage visibility per agent run.** I cannot see what % of the context window a running agent is consuming, so I have no signal for *when* to reset. (This half overlaps #2038 / #4984.)
- **No user-controlled reset.** I cannot say "start this next run from a clean context." Comment follow-ups appear to resume the prior session by default (see #4754), and the only reset-ish mechanism is the `/fresh` directive PR (#1298), which is a manual per-comment flag with no threshold and no UI.
- **No first-class continuity contract.** There is no supported "the agent's durable progress lives in these spec files, the transcript is disposable" model. Resetting the session today risks losing progress, because progress and transcript are conflated.

The net effect: I'm forced to let sessions grow unbounded and hope compaction/resume behaves, instead of running the short-session + external-spec loop that I trust.

## Proposed solution

Three parts, roughly in priority order. Part A depends on the observability work in #2038 / #4984; B and C are the new asks.

**A. Context-usage signal (prerequisite — may be satisfied by #2038 / #4984)**
Surface current-run context utilization (used / limit, %) somewhere I can see it before/while an agent runs.

**B. User-controlled reset, ideally threshold-based**
- A manual "reset context / start fresh" control on the issue or agent-run panel (not only a typed `/fresh` comment).
[...]

Issue #3435, opened 2026-05-28
[Feature]: Codex issue comment follow-ups should optionally resume the previous session or provide a low-token follow-up mode


jianxunzhang1101:
### Deployment type

Official App

### What do you want and why?

When using a Codex-backed agent in an issue, each issue comment/reply creates a new task. In practice, follow-up questions inside the same issue feel like a continuation of the prior agent run, but Multica starts a fresh Codex task/session and reconstructs context from the issue/comments.

This has two problems:
1. It burns significantly more tokens because the issue, comments, instructions, and environment context are re-injected every time.
2. It loses the conversational continuity users expect when replying to an agent's previous answer.

This is especially noticeable for operational/debugging workflows:
- Agent posts an investigation summary.
- User replies: "What does this error mean?"
- User replies again: "Check the raw prompt/log around that error."
Each reply is logically a continuation, but Codex behaves like a new run.

Expected:
For issue comment follow-ups, Multica should provide one of:
- an option to resume the previous Codex session/thread for the same issue + agent + runtime,
- a "follow-up mode" that keeps a compact rolling summary / decision memory,
- or a UI distinction between "new task" and "continue conversation".

Current behavior:
Each reply/comment creates a new task. Codex session resumption appears unavailable or ineffective for this path, so repeated follow-ups are expensive and less coherent than continuing directly in Codex.

Related:
- #1136 describes repeated sandbox/session initialization with Codex in the same issue card.
- Docs say Codex session resumption is currently "code exists but unusable".

Bohan-J:
Noted, we are currently working on the session resume logic. Estimated to be published tomorrow

Bohan-J:
@jianxunzhang1101 good news — comment-driven session resume shipped 🎉

**What changed:** comment-triggered tasks now **resume the prior agent session by default** (same runtime), instead of starting a fresh one and re-injecting the entire issue context. Applies across Codex and the other resume-capable runtimes (Claude, Hermes, Kimi, Kiro).

This directly addresses both problems you raised:

1. **Token cost** — the model isn't re-reading the issue, comments, instructions, and environment every reply; it picks up the conversation where it left off.
2. **Conversational continuity** — replies on the agent's prior answer feel like a continuation, not a separate task.

Cold start (no prior run) still does a full-context read.

A small follow-on also shipped: the per-turn comment hint is now a since-delta — "N new comment(s) since your last run, `--since <ts>`" — so the agent only fetches what's new instead of re-reading the whole thread.

[...]

Issue #2038, opened 2026-05-03
[Feature]: Agent Context Window Observability


pangzitudou:
### Deployment type

multica.ai (hosted)

### What do you want and why?

<h3 data-heading="Background">Background</h3>
<p>Multica's current token tracking focuses on the <strong>cost accounting</strong> dimension (how many tokens / how much money spent), with no observability into <strong>runtime health</strong>. Users cannot determine:</p>
<ul>
<li>What percentage of an agent's session context window is consumed</li>
<li>Whether a session needs manual compaction or restart</li>
<li>Whether a session is about to hit the model's context limit</li>
</ul>
<h3 data-heading="Current State">Current State</h3>

Capability | Web UI | CLI
-- | -- | --
Daily runtime token summary | ✅ Charts | ✅ runtime usage
Per-issue token consumption | ✅ Sidebar (historical data may be missing #728) | ❌ N/A
Agent context window utilization | ❌ | ❌
Session compaction status/advice | ❌ | ❌

<h3 data-heading="Desired Features">Desired Features</h3>
<p><strong>P0 — Context Window Utilization Visibility</strong></p>
<p>Display current session context consumption on the Issue detail page (or Agent run status panel), e.g.:</p>
<p>plain text</p>
<pre><code>Context usage: 68,000 / 200,000 tokens (34%)
Status: Normal | Compaction Recommended | Near Limit
</code></pre>
<p>Data sources: Codex runtime <code>state_*.sqlite</code> → <code>threads.tokens_used</code>, Claude Code <code>result.message.modelUsage</code>, etc.</p>
<p><strong>P1 — Compaction / Restart Suggestions</strong></p>
<p>When context usage exceeds a threshold (e.g. 80%), proactively alert the user:</p>
<blockquote>
<p>"This session's context window is over 80% full. Consider compacting or restarting for better response quality."</p>
</blockquote>
<p><strong>P1 — CLI Access to Per-Issue Token Data</strong></p>
<p>The backend API <code>GET /api/issues/{id}/usage</code> already exists (#581), but <code>multica issue get</code> does not return token data. The CLI should expose this.</p>
<h3 data-heading="Technical Feasibility">Technical Feasibility</h3>
<p>Confirmed data sources exist across runtimes:</p>
<ul>
<li><strong>Codex</strong>: <code>state_*.sqlite</code> → <code>threads.tokens_used</code> column</li>
[...]

PR #8198, opened 2026-09-09
feat(daemon): stop resuming a session that has outgrown its budget (#4754)

### LINKED ISSUE #4754: Comment resume should fall back to low-token mode for long or tool-heavy sessions
## Summary

Comment-triggered follow-up runs currently appear to resume the prior provider/Codex session by default. That helps the case described in #3435, where fresh comment tasks repeatedly re-read issue context, but it creates the opposite failure mode for long or tool-heavy sessions: a small new comment can inherit a very large prior transcript and become more expensive than a fresh low-token follow-up.

This is not a request to remove resume support. It is a request to make resume conditional and to fall back to a low-token follow-up mode when the prior session is already large, old, compacted, or polluted by large tool outputs.

## Reproduction shape

In a private Multica workspace, we observed this pattern with Codex runs instrumented into Langfuse:

1. Run an issue-assignment task that performs a substantial amount of coding work.
2. The task accumulates many tool calls and outputs, including build/test logs and image or browser-preview payloads.
3. Add a short issue comment after that task.
4. The `new_comment` run resumes the prior session.
5. The first generation of the comment follow-up starts with a very large input context, even though the triggering comment is small.

No private issue titles, logs, screenshots, URLs, users, or customer/workspace identifiers are included here.

## Observed data

From a deduplicated Langfuse sample, grouped by `metadata.codex.turn_id`:

| Trigger category | Turns | Total tokens | Share |
| --- | ---: | ---: | ---: |
| `new_comment` | 34 | 120.81M | 54.3% |
| `assigned_issue` | 11 | 95.39M | 42.8% |
| `autopilot` | 16 | 4.15M | 1.9% |

Issue/comment-triggered work accounted for about 97.1% of the measured tokens. Scheduled/no-op autopilot work was not the main driver.

Sanitized examples:

- A comment follow-up with an attachment resumed a long prior session: 85 generations, 155 tool calls, 11.53M total tokens. The first generation was already about 190k input tokens, with later generations around 220k.
- A short text comment follow-up on another long prior session: 70 generations, 115 tool calls, 10.97M total tokens. Later generations were again around 220k input tokens.
- A no-op scheduled scan in the same environment: 7 generations, 7 tool calls, 330k total tokens.

Local session records for the resumed runs showed hundreds of function call/output items, large image/browser-preview-style payloads in the hundreds of KB, and compacted summaries in the tens of KB. The expensive part was not the new comment body; it was the inherited provider session context.

## Expected behavior

A comment follow-up should resume only when the prior session is likely to be useful and cheap enough to continue.

For long/tool-heavy sessions, Multica should use a lower-token follow-up context, for example:

- issue metadata and current title/description summary
- the triggering comment
- a bounded active thread tail
- linked PR metadata if needed
- a structured rolling summary of decisions, changed files, open questions, and latest verification status

## Actual behavior

The comment follow-up resumes the prior session by default, so a short comment can inherit a huge transcript containing prior tool outputs, screenshots/previews, logs, and compaction summaries. This can make `new_comment` turns the dominant token consumer.

## Why #3435 is not enough

#3435 reported the fresh-task failure mode: every comment created a new task, re-read context, and lost continuity. The shipped fix appears to resume the prior session by default, which is useful for short in-flight follow-ups.

This issue is the opposite edge case: for long, stale, compacted, or tool-heavy sessions, default resume can be more expensive and less predictable than a bounded low-token follow-up. The close-out comment on #3435 explicitly asked for a new/reopened issue if resume happens when it should not; this is one of those cases.

Related context: #3047, #2038, #1796.

## Suggested fix

Add a resume eligibility policy for comment-triggered tasks. Possible inputs:

- prior session age
- estimated current context size
- number of prior generations/tool calls
- compaction count or compacted-session marker
- total persisted tool-output size
- large image/browser-preview payload presence
- runtime/model/config mismatch
- failed or poisoned session marker

If eligibility fails, start a fresh low-token follow-up instead of resuming the provider transcript. Also emit the selected mode and reason in run metadata so this can be audited in Langfuse or similar observability tools.

ii-jj:
## What does this PR do?

Adds a resume budget to the issue claim path: when the provider session a comment follow-up would inherit has already billed more input than the budget allows, the run starts a fresh conversation instead. Implements the "fall back to low-token mode for long or tool-heavy sessions" ask in #4754.

**Today** a non-rerun follow-up resumes the most recent `(agent, issue)` session unconditionally. That is right for the case it was built for — a short exchange keeps its context instead of re-reading the issue every turn (#3435) — and wrong for the opposite one. Once a session has run substantial tool-heavy work, the next one-line comment inherits the entire transcript, so answering it costs more than answering it cold would, and eventually costs more than the context window has.

#4754 measured it in a self-hosted workspace: two sessions at **11.0M and 11.5M** billed tokens where a short comment's *first* generation already carried **~190k input tokens**, against **330k for an entire no-op scheduled scan** in the same deployment. `new_comment` runs were **54.3% of all tokens measured**.

The gate is **preventive, not reactive**. `agent_error.context_overflow` and `codex_resume_oversized` are already in `GetLastTaskSession`'s poison list, so a saturated session stops being offered — but only *after* a run has failed on it (#6402). This stops offering it before.

### What it deliberately does not do

- **It does not drop the workdir.** The repository is still checked out and the branch is still this conversation's, so only the transcript resets. `PriorWorkDir` is set exactly as before.
- **It does not lose the thread.** Setting `PriorSessionResumeUnavailable` makes the daemon render the continuity notice it already owns (`sessionContinuityNoticeFor`), and the brief's mandatory comment catch-up is what rebuilds context — from the durable record rather than from a transcript. This is the same mechanism MUL-4424 uses to turn a silent context loss into a visible one.
- **It does not touch chat sessions.** A reset there is directly visible to the person typing; the issue path is the one #4754 is about.
- **It does not fire on ordinary work.** See the default below.

### The signal, and its limits

The precise number — peak input tokens on a single request — is recorded nowhere. What *is* recorded is per-task usage, so the new query sums `input_tokens + cache_read_tokens` across every task that ran on the session. That grows with both transcript length and tool-call count, which are the two things that make a session expensive to inherit, so the proxy moves with what we care about even though its unit is not context-window tokens. Cache reads are counted deliberately: a cached prefix is cheaper, not smaller — it still occupies the window the next turn must fit inside.

This is documented as a proxy in the query comment rather than dressed up as a measurement. The precise version needs the daemon to report peak per-request input on the task row, which is a separate change and, I think, the right follow-up if you want the gate tightened.

### The default is set to not fire

`8_000_000` is an order of magnitude above ordinary work and just below the observed failure region, so on a normal issue this PR changes nothing. Operators who want the short-session workflow #5653 describes set `RESUME_CONTEXT_BUDGET_TOKENS` far lower; an explicit `0` restores the previous always-resume behaviour exactly.

That last part is why `Config.ResumeContextBudgetTokens` is a `*int64`: `0` is a setting, not an absence, and collapsing the two onto one int is how a deliberate "never gate a resume" would silently re-enable itself. Same reasoning as `llm.Config.MaxRetries`.

It also **fails open**: a usage read that errors, or a deployment whose daemons never reported usage, resumes as before. Losing continuity on a working issue because telemetry is missing would be a worse outcome than one more expensive turn, and it would be invisible.

## Related Issue

Closes #4754. Partially addresses #1796 (a run that failed at 208k requested against a 204k limit) and is the reset half of #5653, whose part A (visibility, #2038 / #4984) is still open and would pair well with this.

## Type of Change

## Changes Made

- `server/pkg/db/queries/task_usage.sql` — new `GetSessionBilledInputTokens`; `server/pkg/db/generated/task_usage.sql.go` regenerated with `make sqlc`.
- `server/internal/handler/daemon_resume_budget.go` (new) — the default, the tri-state config read, the pure policy predicate, and the lookup that logs when it fires.
- `server/internal/handler/daemon.go` — the gate, in the non-rerun resume branch only.
- `server/internal/handler/handler.go` — `Config.ResumeContextBudgetTokens`.
- `server/cmd/server/{main,router}.go` — `RESUME_CONTEXT_BUDGET_TOKENS` via a new `envOptionalNonNegativeInt64` (tri-state; the existing `envNonNegativeInt64`-shaped helpers cannot express "explicitly 0").
- `.env.example` — documented. Deliberately **not** added to the four-language `environment-variables.mdx` reference, matching how the other server-side operational tunables (`AUTOPILOT_FAIL_MONITOR_*`) are handled — those four files are byte-parallel and I did not want to half-sync them. Say the word and I will add all four rows.

## How to Test

```bash
cd server
go test ./internal/handler/ ./cmd/server/
go vet  ./internal/handler/ ./cmd/server/
```

Both packages pass locally against a real PostgreSQL (migrations applied from `./cmd/migrate up`), including the DB-backed test.

`TestResumeExceedsContextBudgetReadsWholeSession` is the #4754 shape in miniature: one heavy task, then a cheap one, both on the same session. Neither exceeds the budget alone; together they do — the assertion that fails if the query is ever scoped to a single task. It also covers an unused session, an explicit `0`, and a different session on the same issue.

Manually, to watch it fire:

```bash
RESUME_CONTEXT_BUDGET_TOKENS=1 ./bin/server   # gate every follow-up
RESUME_CONTEXT_BUDGET_TOKENS=0 ./bin/server   # previous behaviour
```

The server logs `resume context budget exceeded; starting a fresh session` with the billed total and the budget.

## Risks

- **Behaviour change on by default.** Mitigated by the default being calibrated above the observed failure region rather than at a round number, and by `0` being a complete opt-out. If you would rather ship it off by default I will flip the default to `0` and let operators opt in — say which you prefer.
- **The proxy can be wrong in one direction.** A session with many cheap tool round trips accumulates billed tokens without a proportionally large transcript, so it could be reset earlier than strictly necessary. The cost of that is one cold start on an issue whose comment history is intact; the cost of the opposite error is the failure #1796 reports.
- **Old rows.** A session whose tasks predate usage recording reads as 0 and resumes, which is the fail-open behaviour above.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — n/a
- [x] I have updated relevant documentation to reflect my changes (`.env.example`; see the note above about the four-language reference)

## AI Disclosure

**Prompt / approach:** Started from "why is Multica's runtime context so large" — measured the runtime brief section by section (14.6 KB / ~3.7k tokens for the issue kind, already slim after MUL-3560/4297) and the per-turn prompt (~300 tokens), which ruled both out as the driver, then followed the remaining cost to unbounded session inheritance. Read #4754/#1796/#5653/#6402 for the prior art and the production numbers before designing the gate, and deliberately built it on top of the continuity-notice mechanism MUL-4424 already established rather than inventing a second way to tell an agent its context is gone.


ii-jj:
Rebased onto `a9e82c797` (clean, no conflicts). Tests re-run on the rebased branch against a database migrated through 468.

ii-jj:
Rebased onto `7e4758ac1`. The conflict was with MUL-7344, and resolving it needed a decision worth flagging rather than a mechanical merge.

`main` now sets `resumeAnchor = newResumedRunAnchor(...)` on the same line this branch guards, so the two changes meet on one statement: main dates the issue/comment deltas from the run being continued; this branch sometimes declines to continue it.

**Resolution: the anchor is left nil on the budget path.** The anchor means "what changed since the run you are continuing", and past the budget this turn continues nothing — dropping the transcript is the whole point. Claiming a delta there would be worse than claiming none, because `NewCommentsDeltaKnown` is what lets the workflow skip the comment scan, and the agent it would be skipped for has never read those comments. That is also the answer `main` already gives every other non-adopting path (`no prior session`, `another runtime`, `force fresh`, `poisoned rerun`), so the budget path joins the set rather than inventing a case.

Pinned by `TestClaimTaskByRuntime_OverBudgetResumeReportsNoDeltas`: a prior run with 9M billed against the 8M default is not resumed, says so through `PriorSessionResumeUnavailable`, keeps its workdir, and reports neither delta. It is written against the default budget on purpose, so it does not reach into the handler's configuration.

That test is DB-backed and my machine has no local database right now, so it compiles and `go vet` is clean here but **CI is the first place it actually runs** — worth a glance when the checks come back. The non-DB suites pass.

Issue #6283, opened 2026-08-02
[Bug]: agent mcp_config is silently ignored on the claude runtime — every host MCP server stays reachable


ivrylobs:
## Summary

An agent's saved `mcp_config` has no effect on the **claude** runtime. The field accepts a
value, the API and CLI echo it back, the UI presents it as configuration — and the agent
still receives every MCP server from the operator's user-global Claude config.

This is worse than the feature not existing. An operator who sets `mcp_config` to restrict an
agent will believe the restriction is in force. Combined with `--permission-mode
bypassPermissions` (which the daemon passes on every run), there is then no boundary at all
between an agent and every MCP server on the host — including servers belonging to unrelated
clients or pointing at production.

## Environment

- Multica `v0.4.16`, self-hosted, macOS daemon
- Runtime: `claude` (Claude Code `2.1.220`)
- Agent created with `--mcp-config-file` containing `{"mcpServers": {}}`

## Steps to reproduce

1. Create an agent on a `claude` runtime with an empty MCP set:

   ```bash
   echo '{"mcpServers":{}}' > mcp-none.json
   multica agent create --name "Probe" --runtime-id <claude-runtime> \
     --instructions "Read only. Report raw output." \
     --mcp-config-file ./mcp-none.json
   ```

2. Confirm it saved — `multica agent get <id> --output json` returns
   `"mcp_config": {"mcpServers": {}}`.

3. Assign it an issue asking it to list the tools available to it whose names begin with
   `mcp__`, without calling any.

## Expected

No `mcp__*` tools, because the agent's declared MCP set is empty.

## Actual

**1,301 `mcp__*` tools across 14 MCP servers**, every one reachable. They arrive as deferred
tools — schemas unloaded — which is a lazy-loading optimisation, not a restriction: one
`ToolSearch` call with `select:<name>` makes any of them callable.

The set is exactly the operator's user-global MCP configuration, unfiltered. In our case that
included several servers belonging to an unrelated client, one of them pointing at that
client's production database.

## `--strict-mcp-config` via `custom_args` does not close it either

Claude Code has a `--strict-mcp-config` flag that ignores all MCP configuration except what is
passed explicitly. Setting it on the agent:

```bash
multica agent update <id> --custom-args '["--strict-mcp-config"]'
```

changes nothing — the same 1,301 tools are present on the next run. We could not confirm from
inside the task whether the flag reached the CLI at all: `ps` truncates the command line at
[...]

PR #8199, opened 2026-09-09
feat(daemon): let a host stop lending its own MCP servers to agents

ii-jj:
## Problem

`mergeRuntimeAndAgentMcpConfig` folds this host's own MCP configuration into the task-local config of every agent that carries a Multica-managed `mcp_config`: `~/.claude.json`, `~/.cursor/mcp.json`, `$CODEX_HOME/config.toml`, OpenClaw servers, Claude plugin servers. `--strict-mcp-config` is then applied **to the merged result**, so strict mode enforces the merged set rather than the declared one — an agent configured with `{"mcpServers":{}}` still starts every server the machine happens to have.

That is the gap behind #6283: 1,301 tools across 14 servers reaching every run, one of them pointing at an unrelated client's production database. Trimming the agent's own config cannot close it, because the agent's config is not what strict mode is enforcing.

Two costs, one cause:

- **Blast radius** — servers the operator never intended for a given workspace are reachable from it.
- **Context** — every server's tool schemas are injected into the model's prompt. #1796 measured 17,587 tokens of tool input on a single run; that cost is paid on every turn, and it is the largest slice of per-turn overhead that is neither the task nor the conversation.

## Change

A host-level opt-out for the base layer:

```bash
multica daemon start --no-runtime-mcp
MULTICA_DAEMON_RUNTIME_MCP=false multica daemon start
multica config set disable_runtime_mcp_inherit true
```

Off, the agent's config passes through **byte-identical** — no re-marshalling, which also preserves provider-native shapes such as OpenCode's top-level `mcp` map — and `--strict-mcp-config` then enforces exactly what the agent declares.

Precedence follows the existing single-direction disable signal shared by `--no-auto-update` / `--no-auto-reload`: flag > env > config file > default, resolved through `resolveDaemonDisableSignal`.

## Decisions worth reviewing

- **Default stays on.** The merge exists so that adding one managed server does not silently disable an operator's unrelated ones. Moving the default would change behavior for every existing deployment; #6283 asks for a way off, not for a different default.
- **Host-level, not per-agent.** These servers come from the machine, not from any workspace, so no per-agent field is the honest place to refuse them. A workspace-side toggle would also let a workspace reach into a machine's private configuration.
- **Agents without a managed `mcp_config` are untouched in both modes.** Multica writes no config for them, so the runtime keeps its native inheritance — the opt-out withholds the base layer of a merge, it does not start writing configs for agents that had none. Pinned by a test.

## Tests

- `TestMergeRuntimeAndAgentMcpConfigWithoutInheritanceKeepsAgentSetOnly` — host `~/.claude.json` + a Claude plugin server present, merge off, result is the agent config verbatim.
- `TestMergeRuntimeAndAgentMcpConfigWithoutInheritanceStillSkipsUnmanagedAgents` — nil / empty / `null` agent config keeps the passthrough.
- `TestLoadConfig_RuntimeMcpInherit_DefaultsOn` and `..._OffSwitches` — mirror the existing auto-reload resolution tests across env, flag and config-file layers.
- Existing merge tests updated to pass `true` (unchanged behavior).

`go vet ./...` clean; `./internal/daemon/`, `./cmd/multica/`, `./internal/cli/` pass.

## Docs

`CLI_AND_DAEMON.md` gains a "Host MCP servers in agent tasks" section; `MULTICA_DAEMON_RUNTIME_MCP` added to the environment-variable table in all four locales.

Refs #6283, #1796.


ii-jj:
Rebased onto `a9e82c797` (clean, no conflicts). Tests re-run on the rebased branch against a database migrated through 468.

PR #8200, opened 2026-09-09
MUL-7164 fix(tasks): classify concurrent request rejections

mameikagou:
## What does this PR do?

Claude Code can surface an Anthropic-compatible provider rejection as an authentication or access-token failure with HTTP 403 even though the specific response says the account has reached its concurrent request limit. Multica currently persists that as either `provider_auth_or_access` or, when the prefix contains both `token` and `limit`, `context_overflow`. The member is then told to sign in again or reduce context even though neither action addresses the failure.

This PR classifies the specific `concurrent request limit` witness as `provider_capacity_or_rate_limit` before the broader context and auth rules. It also upgrades the stale labels reported by older installed daemons at the server boundary.

Per @Bohan-J's review on #8180:

> We want the classification half of this PR. We'd like the auto-retry half taken out for now.

This split contains classification only. It does **not** add the capacity/rate-limit reason to `retryableReasons`, add a retry delay, or change task retry behavior. A service-layer regression explicitly pins `provider_capacity_or_rate_limit` as non-auto-retryable.

## Related context

- Split from #8180 after review.
- The original incident is represented verbatim in the regression fixtures.
- No public issue exists for the Multica-side misclassification.

## Type of Change

- [ ] New feature (non-breaking change)
- [ ] Documentation update
- [x] Tests
- [ ] CI / infrastructure

## Changes Made

- Match `concurrent request limit` before both the broad `token` + `limit` context rule and generic 403/auth rule.
- Normalize legacy `context_overflow`, `provider_auth_or_access`, `unknown`, and coarse `agent_error` labels only when the same narrow raw-error witness is present.
- Cover both observed phrasings:
  - `Failed to authenticate. API Error: 403 ... concurrent request limit`
  - `Failed to refresh access token. API Error: 403 ... concurrent request limit`
- Prove ordinary 403 auth failures and ordinary token-limit overflows are unchanged.
- Prove the corrected capacity reason remains resume-safe and non-auto-retryable.

## Offline backfill note

The MUL-1949 offline backfill SQL referenced by the classifier comments is not checked into this repository, so there is no migration file to update in this PR. If historical rows are reclassified, its CASE expression should add `%concurrent request limit%` before the token/context and 403/auth branches. `NormalizeDaemonReason` covers new submissions from mixed-version installed daemons; it does not rewrite already-persisted historical rows.

## Validation

- `go test ./pkg/taskfailure -count=1`
- focused classifier and mixed-version normalization tests under `-race`, 50 repetitions
- `go test ./internal/service -run '^TestTaskFailureClassifiers$' -count=1`
- `go vet ./pkg/taskfailure ./internal/service`
- `git diff --check`

## Risks

The witness is intentionally narrow and only changes taxonomy and member-facing recovery guidance. It cannot make unrelated capacity errors retry, and negative regressions preserve the existing auth and context classifications.

## Checklist

- [x] I have run targeted tests locally and they pass
- [x] I have added regression coverage
- [x] I have considered mixed-version daemon behavior
- [x] I have documented the external backfill boundary
- [ ] UI screenshots are not applicable
- [ ] Documentation changes are not applicable because retry behavior is unchanged

## AI Disclosure

**Approach:** Split the accepted classification portion from #8180, added the exact ordering counterexample from review, kept retry behavior unchanged, and verified both current- and older-daemon reason paths.

### DIFF
--- a/server/internal/service/task_complete_race_test.go
+++ b/server/internal/service/task_complete_race_test.go
@@ -257,6 +257,9 @@ func TestTaskFailureClassifiers(t *testing.T) {
 		// Transient mid-stream provider disconnect (MUL-4910): retryable, and
 		// resume-safe so the retry continues the truncated conversation.
 		{reason: "agent_error.provider_network", wantType: "agent_error", wantResumeOK: true, wantRetry: true},
+		// Capacity/rate-limit failures keep their existing user-retry posture.
+		// Correcting a misleading auth label must not enable an automatic resend.
+		{reason: "agent_error.provider_capacity_or_rate_limit", wantType: "agent_error", wantResumeOK: true, wantRetry: false},
 		{reason: "runtime_recovery", wantType: "runtime", wantResumeOK: true, wantRetry: true},
 		{reason: "iteration_limit", wantType: "agent_output", wantResumeOK: false, wantRetry: false},
 		{reason: "api_invalid_request", wantType: "agent_error", wantResumeOK: false, wantRetry: false},
--- a/server/pkg/taskfailure/classify.go
+++ b/server/pkg/taskfailure/classify.go
@@ -35,6 +35,14 @@ var (
 	httpCapacityCodeRe = regexp.MustCompile(`(^|[^0-9])(429|529)([^0-9]|$)`)
 )
 
+// concurrentRequestLimitWitness is emitted by Anthropic-compatible providers
+// that use HTTP 403 for a transient concurrency rejection. Claude Code may
+// prefix it with an authentication or access-token failure, but credentials
+// remain valid and a later request can succeed. Match the semantic witness
+// before both token-window and generic auth rules so the persisted reason and
+// member-facing recovery guidance describe the actual failure.
+const concurrentRequestLimitWitness = "concurrent request limit"
+
 // Classify maps a free-form error string from the agent runtime / CLI
 // to one of the 14 agent_error.* sub-reasons. Always returns a valid
 // Reason; falls back to ReasonAgentUnknown when no rule matches and for
@@ -71,6 +79,12 @@ func Classify(rawError string) Reason {
 	lower := strings.ToLower(trimmed)
 
 	switch {
+	// A concurrent-request rejection can contain both "access token" and HTTP
+	// 403. Its specific semantic witness must beat the broader context and auth
+	// rules below; this classification does not itself make the reason retryable.
+	case strings.Contains(lower, concurrentRequestLimitWitness):
+		return ReasonAgentProviderCapacityOrRateLimit
+
 	// 1. Context / token window overflow. Checked early so "token
 	//    limit" doesn't get swallowed by the broader "limit" / "quota"
 	//    rule below.
@@ -453,6 +467,17 @@ var legacyOpencodeStreamEndedReasons = map[string]bool{
 	"agent_error":                     true,
 }
 
+// legacyConcurrentRequestLimitReasons are the stale buckets emitted by daemons
+// whose classifiers let a generic token/context or HTTP 403 rule win over this
+// more specific wire shape. The raw witness keeps the upgrade narrow; unrelated
+// context overflows and authentication failures retain their original reason.
+var legacyConcurrentRequestLimitReasons = map[string]bool{
+	string(ReasonAgentContextOverflow):      true,
+	string(ReasonAgentProviderAuthOrAccess): true,
+	string(ReasonAgentUnknown):              true,
+	"agent_error":                           true,
+}
+
 // NormalizeDaemonReason upgrades a failure_reason reported by an older daemon
 // onto the taxonomy this server understands, using the raw error text as the
 // witness. It returns the reason unchanged when nothing applies.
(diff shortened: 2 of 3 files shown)


Bohan-J (approved):
This addresses everything from the #8180 review. Approving.

Verified locally on `801d5977`:

**The split is clean.** `retryableReasons` is untouched, `retryDelayForAttempt` is untouched, and nothing else in `internal/service` changes behavior. The only service-side edit is the new assertion row.

**The ordering fix is the stronger of the two options I suggested.** Hoisting the witness above rule 1 rather than only papering over it with a test means both observed phrasings land correctly, and I confirmed the neighbours are unchanged:

| raw error | classified as |
|---|---|
| `Failed to authenticate. API Error: 403 ... concurrent request limit. ...` | `provider_capacity_or_rate_limit` |
| `Failed to refresh access token. API Error: 403 ... concurrent request limit.` | `provider_capacity_or_rate_limit` |
| `API Error: 403 Forbidden` | `provider_auth_or_access` |
| `prompt is too long: 250000 tokens > 200000 token limit` | `context_overflow` |
| `API Error: 429 rate limit reached` | `provider_capacity_or_rate_limit` |
| `missing api_key for openai (401 returned downstream)` | `missing_config` |

**Adding `ReasonAgentContextOverflow` to the legacy normalization set is right, and the consequence is intended.** `context_overflow` is resume-unsafe while the corrected reason is not, so this normalization also restores session resumption for those rows — which is correct, since the conversation was never actually overflowed. The independent text guards in `ResumeUnsafeFailure` still apply on top.

**The `wantRetry: false` row is a good addition.** It turns "this PR does not enable a retry" into something the suite enforces, so the open question can't be answered by accident later.

**The invariant comment I asked about no longer needs a change.** `classify.go:26-31` argues those regexes are safe because the auth/quota/capacity buckets are all non-retryable and `provider_network` is the only `agent_error.*` reason on the allowlist. Both sentences are still true under this PR (`skill_bundle_unavailable` is not an `agent_error.*` reason), so the comment stands as-is. Good that you didn't touch it.

**Docs stay correct.** Since retry behavior is unchanged, the `Retry later` row in `tasks.mdx` remains accurate — the "not applicable" note on that checklist item is right this time.

**On the backfill SQL:** thanks for chasing that down and writing up the exact CASE witness and ordering. That's the correct answer given it isn't in this repository; the follow-up is ours to carry if we ever reprocess historical rows.

`go build ./...`, `go test ./pkg/taskfailure ./internal/service -count=1` pass locally.

The auto-retry question stays parked on our side until we resolve why #7586 reverted #6747 — we'll come back to you rather than the other way around. Thanks for splitting this cleanly and for turning the counterexample into real coverage instead of just a passing case.

Bohan-J:
Merging this. Thank you — this was a good contribution, and worth saying why.

You found a real defect that nobody had filed: a provider rejection that has nothing to do with credentials was telling members to sign in again, which pushes people into re-authenticating or rotating a key when the actual answer is to wait a moment. That's a bug with a real cost, and it took reading an incident carefully to notice it.

You also handled the review well. Being asked to split a PR and drop half of your work is not fun, and you did it cleanly and without argument — #8200 is genuinely just the classification, with the retry path untouched. When given two ways to fix the `token` + `limit` ordering hazard, you took the one that actually fixes it rather than the one that only makes the test pass. Adding the assertion that `provider_capacity_or_rate_limit` stays non-retryable was your own idea and a good one: it turns "we deliberately didn't do this yet" into something the suite enforces. And on the offline backfill SQL, you checked, found it isn't in this repository, and wrote down exactly what would be needed instead of quietly skipping the question — that's the right way to handle a boundary you can't cross.

On the auto-retry proposal: it's parked on our side, not rejected. We need to work out why #7586 reverted #6747 before we re-approach that area. We'll come back to you on it rather than leaving you to guess — if we do pursue it, the shape we'd want is a non-zero cooldown and a gate on the specific wire shape rather than the shared reason.

Thanks again, and please do send more.

mameikagou:
> Merging this. Thank you — this was a good contribution, and worth saying why.
> 
> You found a real defect that nobody had filed: a provider rejection that has nothing to do with credentials was telling members to sign in again, which pushes people into re-authenticating or rotating a key when the actual answer is to wait a moment. That's a bug with a real cost, and it took reading an incident carefully to notice it.
> 
> You also handled the review well. Being asked to split a PR and drop half of your work is not fun, and you did it cleanly and without argument — #8200 is genuinely just the classification, with the retry path untouched. When given two ways to fix the `token` + `limit` ordering hazard, you took the one that actually fixes it rather than the one that only makes the test pass. Adding the assertion that `provider_capacity_or_rate_limit` stays non-retryable was your own idea and a good one: it turns "we deliberately didn't do this yet" into something the suite enforces. And on the offline backfill SQL, you checked, found it isn't in this repository, and wrote down exactly what would be needed instead of quietly skipping the question — that's the right way to handle a boundary you can't cross.
> 
> On the auto-retry proposal: it's parked on our side, not rejected. We need to work out why #7586 reverted #6747 before we re-approach that area. We'll come back to you on it rather than leaving you to guess — if we do pursue it, the shape we'd want is a non-zero cooldown and a gate on the specific wire shape rather than the shared reason.
> 
> Thanks again, and please do send more.

Thanks for the detailed review and for merging this. I will do this more.

PR #8205, opened 2026-09-09
MUL-7160: fix(agent): discover Codex models from the effective catalog

### LINKED ISSUE #8177: [Bug]: Codex capability discovery ignores the configured model catalog
### Deployment type

self-host

### Area

Agents and runtimes

### What happened?

Codex model discovery uses `codex debug models --bundled`. That intentionally ignores a configured `model_catalog_json`, so Multica's model/effort/speed pickers and execution-time capability checks can use a different catalog from the one Codex actually executes with.

For example, a custom catalog can expose `gateway/custom-codex` with reasoning level `high` and service tier `priority` (Fast). The effective Codex catalog includes that exact ID and metadata, but Multica receives only the bundled native IDs. The custom model can be entered manually, yet Fast is unavailable; a persisted `service_tier=priority` or `thinking_level=high` can also be skipped by the daemon because the model does not match the discovery catalog.

The inverse matters too: if a custom catalog restricts the capabilities of an existing native ID, Multica can advertise the bundled model's capabilities instead of the configured subset.

Expected: use the selected runtime command's effective catalog, preserving its exact model IDs and per-model capabilities. This should retain bounded discovery and sensible fallback behavior; it should not globally strip provider prefixes or assume all custom models support Fast.

### Steps to reproduce

1. Configure a Codex runtime with a valid `model_catalog_json` containing a visible custom model ID, such as `gateway/custom-codex`, with a supported reasoning level and service tier. The catalog can be supplied through a runtime command prefix such as `-c model_catalog_json=/path/to/catalog.json`.
2. Compare the same binary and configuration:
   ```sh
   codex -c model_catalog_json=/path/to/catalog.json debug models
   codex -c model_catalog_json=/path/to/catalog.json debug models --bundled
   ```
   The first returns the configured custom model; the second returns the bundled native IDs without the custom model.
3. Request the Multica runtime's model list and configure an Agent to use the custom ID. The runtime catalog omits the ID, so its advertised effort/Fast options are unavailable.
4. If the corresponding overrides were saved through the API, the daemon's per-model validation rejects the mismatch and omits those overrides at execution time.

### Additional context

- Reproduced on Multica v0.4.38; the discovery code is unchanged on current `main`, `461479a9e1146788a495717330b3e16ade965485`.
- Independently reproduced the native CLI distinction with Codex `0.151.0-alpha.7.2`, an isolated empty home, no credentials, and a custom catalog. This is catalog discovery only, not a paid model invocation.
- A regression fixture against current `main` calls the real public `ListModels` entry point with a stand-in CLI that distinguishes its bundled and configured catalogs. Both normal runtime configuration and a `-c model_catalog_json=...` runtime prefix return the wrong bundled catalog today.
- Source: [`discoverCodexModels` / `codexDebugModelsArgs`](https://github.com/multica-ai/multica/blob/461479a9e1146788a495717330b3e16ade965485/server/pkg/agent/thinking.go#L319-L388), [`ValidateServiceTierWith`](https://github.com/multica-ai/multica/blob/461479a9e1146788a495717330b3e16ade965485/server/pkg/agent/thinking.go#L740-L773).
- Related but distinct: #4868 fixed copying a referenced custom catalog into the task-local Codex home. This report concerns discovery and capability validation before launch. #5821 already added Fast support itself.

ADanMan:
## Problem

Codex model discovery ran `codex debug models --bundled`. The `--bundled` flag reports the binary's baked-in native catalog and **intentionally ignores a configured `model_catalog_json`**, so Multica's model/effort/service-tier pickers and the daemon's execution-time capability checks could use a different catalog from the one Codex actually runs with.

Consequences (all in #8177):

- A custom catalog model (e.g. `gateway/custom-codex` with reasoning `high` and a `priority`/Fast service tier) was omitted, so Fast was unavailable and persisted `service_tier=priority` / `thinking_level=high` overrides got dropped by daemon validation.
- The inverse: a catalog that *restricts* a native ID's capabilities was over-advertised from the bundled set.

## Fix

Drop `--bundled` so `codex debug models` reports the **effective** catalog Codex executes with. The runtime launch prefix (`Command.Prefix`) is already applied ahead of the subcommand by `Command.exec`, so a catalog supplied via a `-c model_catalog_json=/path` runtime prefix is reflected as well — no provider prefixes are stripped, and capabilities come from the effective per-model entry rather than assuming Fast.

Discovery stays bounded and falls back safely:

- The plain command can refresh over the network (which `--bundled` used to skip), so `runCodexDebugModels` caps it with a 15s timeout, matching the opencode/pi discovery paths.
- A timeout, error, or empty/malformed payload still falls back to `codexStaticModels()`, so behaviour with **no** configured catalog is unchanged.

The code change is two lines (argv + bounded timeout); the rest is the doc comment and tests.

## Test

Adds `TestListModelsCodexUsesEffectiveCatalog`, a regression through the real public `ListModels` entry point using a test-created stand-in `codex` that returns a **different** payload for `debug models --bundled` (bundled native IDs only) than for a plain `debug models` (the configured catalog, which adds `gateway/custom-codex` with `high` + `priority`/Fast). It asserts Multica now receives the configured custom model and its capabilities, covering both:

- plain runtime config, and
- a `-c model_catalog_json=…` runtime prefix (also asserting the prefix reached the binary and `--bundled` did not).

The test fails against the old `--bundled` argv and passes with the fix. The two existing argv-pinning/fallback tests were updated to the new `{"debug", "models"}` contract.

Verification (macOS):

- `gofmt -l` clean on both files
- `go build ./...`
- `go vet ./pkg/agent`
- `go test ./pkg/agent -run 'Codex|ListModels|ServiceTier|ThinkingLevel|Thinking' -count=1 -p 1` — ok

(The full `./pkg/agent` run surfaced only unrelated traecli/kimi ACP handshake timeouts that pass in isolation — load-induced flakes in a 223s fully-parallel run, not this change.)

Closes #8177

### DIFF
--- a/server/pkg/agent/thinking.go
+++ b/server/pkg/agent/thinking.go
@@ -245,11 +245,11 @@ func projectClaudeLevels(superset []string, allow map[string]bool) []ThinkingLev
 
 // ── Codex ────────────────────────────────────────────────────────────
 //
-// `codex debug models --bundled` is the structured discovery hook for the
-// visible model catalog, each model's reasoning catalog, and service tiers. OpenAI added
-// the command and `--bundled` flag together in Codex 0.122.0 (openai/codex
-// #18625). Older versions, failed invocations, and malformed/empty payloads
-// use codexStaticModels so the picker remains usable.
+// `codex debug models` is the structured discovery hook for the visible model
+// catalog, each model's reasoning catalog, and service tiers. OpenAI added the
+// command in Codex 0.122.0 (openai/codex #18625). Older versions, failed
+// invocations, and malformed/empty payloads use codexStaticModels so the
+// picker remains usable.
 //
 // We prefer this over the older config-error probe trick because:
 //   1. It gives us per-model subsets without hand-maintained tables.
@@ -258,14 +258,30 @@ func projectClaudeLevels(superset []string, allow map[string]bool) []ThinkingLev
 //
 // The subcommand emits JSON on stdout by default — there is no
 // `--output json` flag (a prior version of this code passed one and
-// silently failed on 0.131.0). We add `--bundled` to skip the network
-// refresh: discovery runs on every daemon poll and a network hop here
-// would block the picker behind whatever the user's connection allows.
-// The bundled catalog is what determines which `model_reasoning_effort`
-// tokens the local binary actually accepts, which is the only thing we
-// need for validation.
+// silently failed on 0.131.0).
 //
-// The static fallback deliberately mirrors a recently verified bundled
+// We deliberately do NOT pass `--bundled`. That flag reports the binary's
+// baked-in native catalog and intentionally ignores a configured
+// `model_catalog_json`, so a runtime whose catalog exposes a custom model
+// (e.g. `gateway/custom-codex` with its own reasoning levels and service
+// tiers) — set in config.toml or supplied through a runtime command prefix
+// like `-c model_catalog_json=/path` — would be discovered with the wrong IDs
+// and capabilities. Multica would then advertise Fast/effort options the
+// effective catalog does not have and drop valid persisted service_tier /
+// thinking_level overrides at launch, while a custom catalog that *restricts*
+// a native ID's capabilities would be over-advertised from the bundled set
+// (#8177). The runtime command's launch prefix (carried by Command.Prefix) is
+// applied ahead of the subcommand by runtimeCmd.exec, so the plain command
+// reports the exact catalog Codex will execute with, including a
+// `-c model_catalog_json=…` prefix.
+//
+// Dropping `--bundled` reintroduces the network refresh the flag used to skip,
+// so runCodexDebugModels caps the call with codexDebugModelsTimeout to keep
+// discovery bounded on the daemon poll path. A timeout, error, or empty/
+// malformed payload falls back to codexStaticModels rather than blocking or
+// emptying the picker.
+//
+// The static fallback deliberately mirrors a recently verified visible
 // model/thinking catalog. It does not guess service-tier availability.
 
 // codexEffortLabel is the human display string for each Codex effort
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
Request changes: using the effective Codex catalog is the right product behavior, but the current fallback path is still reported as authoritative. That can cache or persist bundled capabilities after a discovery failure and can still silently discard settings for custom models.

1. When `codex debug models` times out, exits unsuccessfully, or returns an invalid/empty payload, `discoverCodexModels` returns `codexStaticModels()`. `ListModels` then wraps that result through `discovered(..., nil)`, so `Catalog.Fallback` remains false. The daemon may cache this static result for 60 seconds, and the server may replace its last-known-good effective catalog with it. Removing `--bundled` also introduces the remote-refresh path for runtimes without a custom catalog, making this failure mode materially more likely. Please preserve the distinction between an effective discovery result and a static fallback, so fallback results are not cached or persisted as authoritative.

2. Setting `Catalog.Fallback` alone is not sufficient. The execution-time thinking-level and service-tier validators currently consume the returned model list without checking that flag. If effective discovery is unavailable and the selected model exists only in `model_catalog_json`, the bundled fallback does not contain it; validation then treats the configured capability as unsupported and drops the persisted override before launching Codex. Please route fallback through the existing lookup-unavailable/error semantics for explicit model overrides, while preserving the current fail-closed behavior for an empty Codex model. Add coverage for failure followed by recovery and for custom-model `thinking_level` / `service_tier` settings surviving a fallback.

I verified the PR merged onto current `main`: the added effective-catalog test passes, the targeted agent test suite passes, and `go vet ./pkg/agent` plus `go build ./...` pass. A failure-then-recovery regression reproduces the first issue: the failed discovery returns `Fallback=false`, and the next successful discovery remains masked by the cached static catalog.

ADanMan:
Addressed the review — the fallback path is no longer reported as authoritative.

1. **Fallback is flagged, not cached/persisted.** `discoverCodexModels` now returns a `Catalog` and marks every static-fallback path (unknown/too-old version, failed or timed-out `debug models`, empty/malformed payload) `Catalog.Fallback=true`. `ListModels` returns it directly instead of `discovered(..., nil)`, so `cachedDiscovery` skips it (no 60s pin) and the server won't replace its last-known-good effective catalog with a transient failure.

2. **Validators respect the flag for explicit overrides.** For an explicit codex model absent from a fallback catalog, `ValidateThinkingLevelWith` / `ValidateServiceTierWith` now return a lookup-unavailable error (`errCatalogUnavailable`) instead of `(false, nil)`. The daemon already treats a catalog-lookup error as "pass the override through", so a persisted `thinking_level` / `service_tier` for a custom `model_catalog_json` model survives a transient discovery failure rather than being dropped before launch. The guard is scoped to codex — other providers' curated static fallbacks (and a real, non-fallback codex catalog) still fail closed on a genuinely unknown model — and an empty codex model still fails closed since no catalog can resolve which model `config.toml` selects.

**Tests:** `TestListModelsCodexFallbackNotCachedThenRecovers` (failed discovery reports `Fallback=true` and does not mask the next successful poll via the cache) and `TestValidateCodexCustomModelSurvivesFallback` (custom-model `thinking_level` / `service_tier` preserved on a fallback; empty-model fail-closed retained). The existing version-gate/fallback test now also asserts the `Fallback` flag.

Verified on macOS: `gofmt -l` clean, `go build ./...`, `go vet ./pkg/agent`, and `go test ./pkg/agent -run 'Codex|ListModels|ServiceTier|Thinking' -count=1 -p 1` all pass. Rebased onto current `main` (resolved a `thinking_test.go` conflict where the static default model was renamed to `gpt-6-astra`).

ADanMan:
Closing this as superseded by #8722 (MUL-7613, "refresh live model catalogs"), now merged to main. That change already delivers everything this PR aimed at, and more:

- Codex discovery reads the effective/live catalog via `discoverCodexCatalog` (`codex debug models` first), which is the core of #8177.
- The same non-authoritative `Catalog.Fallback` flag is set on every static/version-gate/failure/empty path, and `ListModels`' codex case returns the discovery result directly, so a discovery failure is neither cached nor persisted as authoritative.
- `missingFromFallbackCatalog` gates both the thinking-level and service-tier validators, preserving a configured override for an explicit custom model that is only absent because discovery was unavailable, while keeping empty-model fail-closed.

#8722 also covers cases this PR didn't (explicit standard service-tier resolved before the missing-model check; broader live-discovered provider-prefix allowances). No guarantee here is left uncovered, so there's nothing to rebase in. Thanks to the #8722 author. (If useful, the only leftover is cosmetic — a named `errCatalogUnavailable` sentinel in place of the inline error for `errors.Is` callers — which I'm happy to open separately if wanted.)

PR #8220, opened 2026-09-09
fix(codex): accept gpt-6-astra in the static model catalog

### LINKED ISSUE #8219: [Bug]: Codex gpt-6-astra is rejected as model unavailable
### Deployment type

Official App

### Area

Agents and runtimes

### What happened?

Selecting Codex model `gpt-6-astra` (or saving a personal runtime override with that id) fails with `resolution_status=model_unavailable` ("Runtime model unavailable"). The local Codex CLI catalog already lists `gpt-6-astra` with `visibility: list`, and `codex` can run it.

Root cause: `ModelKnownIncompatibleWithProvider` treats Codex compatibility as an exact match against `codexStaticModels()`. `gpt-6-astra` is a live Codex slug that is not in that static fallback. Because it has the `gpt-` prefix, it is classified as a known runtime-specific id and rejected, even though unknown/custom ids are supposed to fail *open*.

The settings toast is only displaying that server status. Daemon discovery (`codex debug models --bundled`) can still report the model from the installed CLI; the save/resolve path never consults that live list.

Related but not a substitute: #8177 / #8205 switch discovery to the effective (non-`--bundled`) catalog. That helps the picker; it does not change the static allowlist used by `ModelKnownIncompatibleWithProvider`.

Expected: when the installed Codex CLI advertises `gpt-6-astra`, selecting it should resolve as `ready` and persist.

### Steps to reproduce

1. Bind an agent (or a personal runtime override) to an online Codex runtime whose local CLI catalog includes `gpt-6-astra`.
2. Set the model to `gpt-6-astra` (or leave an agent recommended model of `gpt-6-astra`).
3. Save, or reload the agent runtime-config summary.
4. Observe `resolution_status=model_unavailable` and the "Runtime model unavailable" toast.

```
Codex CLI 0.153.x live catalog includes gpt-6-astra (visibility: list).
Static fallback in server/pkg/agent/models.go still starts at gpt-5.6-sol.
```

WitMiao:
## Summary

Codex CLI live catalogs already advertise `gpt-6-astra`. Multica still treats Codex compatibility as an exact match against `codexStaticModels()`, so a `gpt-*` id that is not in that snapshot is rejected as `model_unavailable`.

This adds `gpt-6-astra` to the static Codex catalog (current flagship `Default`; thinking `low`…`ultra`) and syncs the front/back pricing tables to OpenAI's published $10 / $50 rates (cache read $1, cache write $12.50).

Related: #8177 / #8205 switch discovery to the effective catalog. That helps the picker; it does not change this server-side allowlist.

## Test plan

- [ ] `go test ./pkg/agent ./internal/metrics -run 'TestModelKnownIncompatibleWithProvider|TestParseCodexModelCatalog$|TestDiscoverCodexModelsVersionGateAndFallback|TestValidateThinkingLevelCodexPerModelFallbackCatalog|TestStaticModelCatalogsAreValid|TestPriceForModelAliasCodexGPT56' -count=1`
- [ ] On a Codex runtime whose CLI catalog includes `gpt-6-astra`, save that model and confirm `resolution_status=ready` (no "Runtime model unavailable" toast)
- [ ] Settings picker for `gpt-6-astra` lists `ultra`

Fixes #8219

### DIFF
--- a/packages/views/runtimes/utils.test.ts
+++ b/packages/views/runtimes/utils.test.ts
@@ -284,6 +284,7 @@ describe("estimateCost", () => {
     // can't hide behind an input-only assertion. `total` is 1M of each of the
     // four categories priced at its own rate.
     const cases = [
+      { model: "gpt-6-astra", input: 10, cacheRead: 1, cacheWrite: 12.5, output: 50, total: 73.5 },
       { model: "gpt-5.6-sol", input: 5, cacheRead: 0.5, cacheWrite: 6.25, output: 30, total: 41.75 },
       { model: "gpt-5.6-terra", input: 2.5, cacheRead: 0.25, cacheWrite: 3.125, output: 15, total: 20.875 },
       { model: "gpt-5.6-luna", input: 1, cacheRead: 0.1, cacheWrite: 1.25, output: 6, total: 8.35 },
@@ -340,6 +341,8 @@ describe("estimateCost", () => {
     // literal-dot alias in server/internal/metrics/pricing.go (MUL-4347).
     expect(isModelPriced("gpt-5-6-luna")).toBe(false);
     expect(isModelPriced("gpt-5-6-sol")).toBe(false);
+    expect(isModelPriced("gpt-6-astra")).toBe(true);
+    expect(isModelPriced("gpt-6-astra-pro")).toBe(false);
     expect(
       estimateCost({
         ...zeroUsage,
--- a/packages/views/runtimes/utils.ts
+++ b/packages/views/runtimes/utils.ts
@@ -223,12 +223,13 @@ const MODEL_PRICING: Record<
   // -- OpenAI: dotted-minor Codex catalog SKUs. Each generation is priced
   //    independently — no fallback to `gpt-5`. Entries track
   //    `server/pkg/agent/models.go` (Codex provider list).
-  //    gpt-5.6 (sol/terra/luna) uses OpenAI's official announcement rates.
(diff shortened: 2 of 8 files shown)


Bohan-J:
Closing this together with #8219 — that report turned out to be an older Codex CLI rather than a server-side rejection, so this PR isn't fixing the bug it claims to. `ModelKnownIncompatibleWithProvider` never rejects a model save; details are in #8219.

Two pieces of the change are independently real, though, and worth reopening as their own PR if you'd like to pick them up:

- **Pricing.** `gpt-6-astra` is unmapped in both `server/internal/metrics/pricing.go` and `packages/views/runtimes/utils.ts`, so it gets no cost attribution on either side. That's a genuine gap and stands on its own.
- **Static catalog.** Adding the entry to `codexStaticModels()` would stop the fallback path from silently dropping `thinking_level` for astra when discovery fails (offline, logged out, or CLI older than 0.122.0).

One caveat for that second one: the static catalog's convention is to mirror the newest locally verified **bundled** catalog, so the entry should be confirmed against `codex debug models --bundled` on 0.153.x before landing — the reported `visibility: list` came from the live catalog. If astra only appears in the effective catalog and not the bundled one, then #8205 (dropping `--bundled` from discovery) is the change that actually matters, and a static entry would over-advertise `ultra` on binaries that can't run it.

Framed as pricing plus catalog hygiene rather than a bug fix, both parts are easy to review. Thanks for digging into this and for the thorough issue write-up.

PR #8225, opened 2026-09-09
MUL-7201: feat(cli): expose agent conversation starters on create and update

### LINKED ISSUE #8224: [Feature]: Expose agent conversation starters in the CLI (`agent create`/`update` have no flag)
### Deployment type

Official App

### Area

CLI

### What do you want and why?

`POST /api/agents` and `PUT /api/agents/{id}` already accept `conversation_starters` (at most 3 items; each `label` ≤ 80 characters, each `prompt` ≤ 4000). The web UI can set them from Build with AI and from the agent's Instructions tab. `multica agent get` already returns the field.

`multica agent create` and `multica agent update` have no matching flag. `multica agent copy` only copies an existing value — it cannot set, replace, or clear the field on its own. This is a CLI surface gap, not a server rejection.

Anyone who manages agents from scripts therefore has to drop to raw HTTP (or the web UI) for this one field, while name, instructions, model, thinking level, and custom args already have CLI flags.

Omit vs clear already have a server contract: omitting the field on create defaults to `[]`; omitting it on update preserves the stored value; sending an explicit `[]` clears it. Copy should keep carrying the source value with no dedicated override flag.

I searched open and closed issues/PRs in this repo for `conversation_starters`, `conversation starters`, and CLI `agent create` / `agent update` flags. Related merged work added the field and renamed it (`#7261`, `#7592`, `#7573`) but did not expose create/update flags. No open issue or in-review PR covers this CLI gap.

Add `--conversation-starters` to `multica agent create` and `multica agent update`, as a JSON array of `{ "label", "prompt" }` objects, in the same Changed-gated style as `--custom-args` / `--thinking-level`:

- send the field only when the flag is provided
- `--conversation-starters '[]'` clears on update
- omitted flags leave create at the server default and leave update unchanged
- `agent copy` keeps copying the source value and does not gain an override flag

Help text and tests should cover create, update (set + clear), copy-still-copies, and invalid JSON.

WitMiao:
## What does this PR do?

The server already accepts `conversation_starters` on agent create/update, and the web UI can write them. `multica agent create` / `agent update` had no flag, so CLI users had to drop to raw HTTP. This wires `--conversation-starters` in the same Changed-gated style as `--custom-args`.

`agent copy` still copies the source value and does not gain an override flag.

## Related Issue

Closes #8224

## Type of Change

- [x] Documentation update

## Changes Made

- `server/cmd/multica/cmd_agent.go`: add `--conversation-starters` to create/update, parse a JSON array of `{label, prompt}` objects, send `[]` (never `null`) so update can clear.
- `server/cmd/multica/cmd_agent_test.go` / `cmd_agent_copy_test.go`: cover create, omit, update set/clear, help, invalid JSON, and copy-has-no-override.
- Docs and the platform agents skill: document the flag; copy remains carry-only.

## How to Test

From `server/`:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test ./cmd/multica -count=1 -run 'TestAgentCreateSendsConversationStarters|TestAgentCreateOmitsConversationStartersWhenUnset|TestAgentUpdateSendsConversationStarters|TestAgentCreateAndUpdateExposeConversationStartersFlag|TestParseConversationStarters|TestAgentCopyDoesNotExposeConversationStartersOverride|TestAgentCopySameRuntimeCopiesPortableFields'
go vet ./cmd/multica
```

These passed locally.

Manual:

```sh
multica agent create --name "Reviewer" --runtime-id <runtime-id> \
  --conversation-starters '[{"label":"Review a PR","prompt":"Review the most relevant open pull request."}]'
multica agent update <id> --conversation-starters '[]'   # clear
multica agent copy <id>                                 # still copies starters; no override flag
```

## Checklist

## AI Disclosure

**AI tool used:** Cursor / Multica Agent

**Prompt / approach:**
Looked up whether this CLI gap already had an upstream issue/PR, then added a thin flag over the existing HTTP field.

## Thinking path

HTTP create/update already persist conversation starters with omit-vs-`[]` semantics. The CLI already copies the field on `agent copy`. The missing piece is the write flags, matching `--custom-args` / `--thinking-level`: send the key only when the flag is set, keep JSON-shape validation in the CLI, leave label/prompt length to the server.

### DIFF
--- a/apps/docs/content/docs/agents-create.ja.mdx
+++ b/apps/docs/content/docs/agents-create.ja.mdx
@@ -71,6 +71,8 @@ Build with AI には、Builder との対話のためにオンラインのラン�
 
 空のままなら、チャットにはローカライズされた汎用の初期値が表示されます。エージェントの役割と境界に合わせた例のほうが通常は役立ちます。
 
+CLI では `agent create` / `agent update` の `--conversation-starters` で設定します（`{"label","prompt"}` オブジェクトの JSON 配列。更新時に `'[]'` を渡すとクリア）。`agent copy` は元の値をコピーするだけで、上書き用のフラグはありません。
+
 ## ランタイム・モデル・思考レベルの選択
 
 ランタイムは、それぞれ 1 つの AI コーディングツールに対応しています。ランタイムを選ぶと、そのツールがサポートするモデルと思考レベルを続けて選べます。一部のツール（Codex など）ではサービスティアも選べます。
@@ -146,7 +148,8 @@ multica agent create \
   --name "Frontend Reviewer" \
   --runtime-id <runtime-id> \
   --description "フロントエンドの Pull Request をレビューする" \
-  --instructions "まず変更とテストを読み、レビュー結果は issue のコメントだけに投稿する。"
+  --instructions "まず変更とテストを読み、レビュー結果は issue のコメントだけに投稿する。" \
+  --conversation-starters '[{"label":"PR をレビュー","prompt":"今いちばん関連する未マージの Pull Request をレビューする。"}]'
 ```
 
 コマンドライン引数の平文はシェル履歴に残ります。stdin と権限を制限したファイルは残りません。
--- a/apps/docs/content/docs/agents-create.ko.mdx
+++ b/apps/docs/content/docs/agents-create.ko.mdx
@@ -71,6 +71,8 @@ AI로 만들려면 Builder 대화를 진행할 온라인 런타임이 필요합�
 
 목록을 비워 두면 채팅에 현지화된 범용 기본값이 표시됩니다. 에이전트의 역할과 경계에 맞춘 예시가 보통 더 유용합니다.
 
+CLI에서는 `agent create` / `agent update`의 `--conversation-starters`로 설정합니다(`{"label","prompt"}` 객체의 JSON 배열, 업데이트 시 `'[]'`를 넘기면 지워집니다). `agent copy`는 원본 값만 복사하며 덮어쓰기 플래그는 없습니다.
+
 ## 런타임, 모델, 사고 강도 선택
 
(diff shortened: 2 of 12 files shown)


Bohan-J:
Thanks for the clean, well-scoped contribution @WitMiao — merging.

Second review pass on `6d53b4a`:

- The flag follows the existing `--custom-args` shape exactly: `Changed`-gated body key, so omitting it leaves the server default on create and the stored value on update. `TestAgentCreateOmitsConversationStartersWhenUnset` pins that.
- `parseConversationStarters` rejecting `null`/empty and coercing a nil slice to `[]` is the right call — it keeps "clear" (`[]`) distinct from "omit", which is exactly the distinction the update handler's `*[]AgentConversationStarter` depends on.
- `agent copy` stays carry-only, and the misconception entry in the platform skill reference was updated rather than left stale.
- Docs and the skill reference are updated in the same PR, all four locales.

The earlier concern about a newer CLI silently dropping the field against an older backend does not apply: our supported configuration is that the backend is never behind the CLI/daemon/clients, so the `/api/config` capability handshake is not needed here. Retiring that capability bit is separate internal cleanup and is not this PR's job.

Verified locally on the PR head: `go build ./...`, `go vet ./cmd/multica`, `go test ./cmd/multica -count=1` (full package, via `scripts/go-test-with-agent-cli-guard.sh`), and `go test ./internal/service -run 'Skill|Builtin' -count=1` for the skill-reference assertions — all pass. CI is green.

One optional nit, not blocking: the CLI duplicates the server's max-3 limit while deferring the label/prompt length limits to the server. If the server limit ever changes, an older CLI would reject a now-valid input with a misleading message. Dropping the client-side count check (or keeping only the shape validation) would be more consistent — happy to leave as is.

PR #8229, opened 2026-09-09
fix(migrations): backfill pre-451 NULL comment_thread_id on pending/in-flight tasks (457) (HYP-1754)

zezhong-zhang:
## Summary

Implements the HYP-1750 ruling (Plan A, debate-approved 3/3) via **migration 457**: a one-shot, idempotent, re-runnable conditional UPDATE that backfills `comment_thread_id` for pre-451 legacy rows that kept `NULL` thread scope when migration 451 deliberately skipped the existing-row backfill (`76f59f5f1`).

Under migration 452's partial unique index, `NULL` folds into the zero-UUID bucket, so every `IS NOT DISTINCT FROM` thread fence in `agent.sql` was blind to those legacy rows:

- a new trigger in any thread slipped past the pending dedup fence → duplicate **sequential** run (`ClaimAgentTask` still serializes per issue+agent, so never concurrent);
- `MergeCommentIntoPendingTask` would not fold new comments into the legacy row;
- `CancelPendingTasksByIssueAndAgentInThread` could not cancel it;
- a legacy `running` / `waiting_local_directory` row missed `RegisterPlannedCommentForActiveTask`'s fence → same-thread comments fell back to enqueue → duplicate sequential run.

## What the migration does (per the ruling, point by point)

1. **Scope** — `comment_thread_id IS NULL AND trigger_comment_id IS NOT NULL` (assignment-level rows are NULL by design via the STRICT trigger and are never touched) AND status in `queued` / `dispatched` / `deferred`+`channel_issue_media_pending` / `running` / `waiting_local_directory`; terminal rows (`completed` / `cancelled` / `failed`) excluded.
2. **Collision guard** — `NOT EXISTS` predicate **verbatim-mirrors** the 452 partial-index WHERE clause (`status IN ('queued','dispatched') OR (status='deferred' AND context->>'channel_issue_media_pending'='true')`) plus `id <>` self-exclusion, so the backfill can never trip the unique index. Colliding legacy rows **stay NULL and are NOT cancelled**, and the migration reports the collision count via `RAISE NOTICE` alongside the backfill count.
3. **Migration-file comments** document: concurrency semantics (candidate scan + guard + UPDATE share one statement snapshot; concurrent INSERTs are backstopped by the 452 unique index), the "fail loud, safe to re-run, never skip silently" policy, the reconcile-narrowing behavior increment (backfilled running rows reconcile at thread scope — standard post-451 semantics), and the deleted-trigger-comment COALESCE fallback (anchor missing → thread = the dead comment id, row leaves the NULL bucket deterministically).
4. **Down migration** — explicit no-op (`SELECT 1;`) with a "backfill is irreversible, no rollback needed" comment, following the `065` convention.
5. **Backfill value** — the thread **root** comment id via `comment_thread_root_id(trigger_comment_id)` (recursive CTE; nested replies get the root, not the direct parent).

## Regression tests

`server/internal/handler/comment_thread_backfill_test.go` — 8 test functions (10 cases) covering the ruling's §4 list plus the deleted-trigger addendum. Legacy rows are simulated faithfully: insert via the normal path (451 trigger fills the scope), then `UPDATE ... SET comment_thread_id = NULL` (the trigger only listens to `UPDATE OF trigger_comment_id`). Tests execute the **real migration file** against the shared test DB:

1. unique index intercepts a same-(issue, agent, thread) pending insert after backfill (23505 from `idx_one_pending_task_per_issue_agent_thread`; different thread not over-blocked);
2. `MergeCommentIntoPendingTask` hits the backfilled row (misses it pre-backfill — pins the regression);
3. `CancelPendingTasksByIssueAndAgentInThread` cancels the backfilled row (misses pre-backfill);
4. assignment-level rows stay NULL;
5. guard fidelity ×3: `dispatched` occupant blocks, `deferred`+media occupant blocks, plain `deferred` occupant does not;
6. nested reply (3rd level) backfills to the thread root id;
7. idempotency: second run reports `backfilled 0 / skipped 0` via captured NOTICE and changes nothing;
8. `deferred`+media legacy row backfilled + terminal rows excluded + backfilled `running` row hit by `RegisterPlannedCommentForActiveTask` (misses pre-backfill);
9. hard-deleted trigger comment: `comment_thread_root_id` COALESCE fallback lands the row in a deterministic singleton bucket (FK bypass simulated with `session_replication_role`; skips gracefully without privileges).

Mutation-checked: dropping `running` from the scope fails test 8; dropping the `deferred`+media arm from the guard fails test 5 with exactly the unique violation the guard exists to prevent.

## Verification (backfill / collision counts)

The migration emits one NOTICE, e.g. `457_agent_task_comment_thread_backfill: backfilled N legacy row(s); skipped M collision row(s) left NULL ...` — visible in the migrate output / Postgres log at deploy time, and asserted in test 7. Post-deploy audit query:

```sql
SELECT count(*) FROM agent_task_queue
WHERE comment_thread_id IS NULL AND trigger_comment_id IS NOT NULL
  AND (status IN ('queued','dispatched','running','waiting_local_directory')
       OR (status = 'deferred' AND context->>'channel_issue_media_pending' = 'true'));
-- expect: the M collision rows reported by the NOTICE (or 0)
```

## Test evidence (local, Postgres 17 via docker compose)

- `go test ./internal/handler/ -run TestThreadBackfill457 -count=1` — **ok** (also with `-race`)
- `go test ./internal/handler/` — **ok** (full package, 34s)
- `go test ./internal/service/ ./internal/migrations/ ./cmd/migrate/` — **ok**
- `go vet ./internal/handler/ ./internal/migrations/ ./cmd/migrate/` and `go vet -tags agentintegration ./internal/handler/` — clean; new file gofmt-clean
- `bash scripts/test-go.sh` (full server suite): 62 packages **ok**; the only failing package, `cmd/multica`, fails **identically on the base commit** (159 tests) because the run environment is itself a Multica daemon task — the CLI suite fail-closes on the runtime's own `.multica/daemon_task_context.json` marker and `MULTICA_*` env vars. Unrelated to this diff (no shared dependencies touched). `internal/daemon/execenv` similarly passes once `MULTICA_TASK_CONFIG_ROOT` is unset.

## Risk & rollback

Low: a single conditional UPDATE over a bounded row set (pending/in-flight legacy rows only; terminal history untouched), no locks held across statements, idempotent and safe to re-run on failure. Rollback is intentionally a no-op — the backfill restores derived fence data to what the 451 trigger would have written.

## Limitations

- Legacy rows that collide with a newer same-thread pending row are intentionally left NULL (counted in the NOTICE); their merge/cancel fence gap remains for that row only — the newer row covers the thread's work.
- `cmd/multica` test package could not be validated locally for the environmental reason above; it is untouched by this diff and covered by CI.

Links: HYP-1754 (implementation), HYP-1750 (audit + ruling).

### DIFF
--- a/server/internal/handler/comment_thread_backfill_test.go
+++ b/server/internal/handler/comment_thread_backfill_test.go
@@ -1,0 +1,418 @@
+package handler
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/multica-ai/multica/server/internal/testutil"
+	db "github.com/multica-ai/multica/server/pkg/db/generated"
+)
+
+// Regression tests for migration 457 (HYP-1750 / HYP-1754): the one-shot
+// backfill of comment_thread_id for pre-451 legacy pending/in-flight rows.
+// Each test simulates a legacy row the way the Tech Lead verified is
+// faithful: INSERT through the normal path (the 451 trigger fills the thread
+// scope), then UPDATE ... SET comment_thread_id = NULL — the trigger only
+// listens to INSERT / UPDATE OF trigger_comment_id, so it does not re-fire —
+// and executes the real migration file against the shared test database.
+// Only this file ever NULLs comment_thread_id on a trigger-bearing row, and
+// tests in a package run sequentially, so the migration's global UPDATE only
+// ever touches the fixture rows of the test currently running.
+
+const backfill457File = "457_agent_task_comment_thread_backfill.up.sql"
+
+func backfill457SQL(t *testing.T) string {
+	t.Helper()
+	_, self, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("resolve test file path")
+	}
+	contents, err := os.ReadFile(filepath.Join(filepath.Dir(self), "..", "..", "migrations", backfill457File))
+	if err != nil {
+		t.Fatalf("read migration %s: %v", backfill457File, err)
+	}
+	return string(contents)
+}
+
+// runThreadBackfill457 executes the real 457 up migration on the shared pool.
+func runThreadBackfill457(t *testing.T) {
+	t.Helper()
+	if _, err := testPool.Exec(context.Background(), backfill457SQL(t)); err != nil {
+		t.Fatalf("run migration 457: %v", err)
+	}
+}
+
+// backfillNoticeConn opens a dedicated connection that records NOTICE
+// messages, so tests can assert the migration's backfill/collision counts.
+func backfillNoticeConn(t *testing.T, notices *[]string) *pgx.Conn {
+	t.Helper()
+	cfg, err := pgx.ParseConfig(testPool.Config().ConnString())
+	if err != nil {
+		t.Fatalf("parse pool conn string: %v", err)
+	}
+	cfg.OnNotice = func(_ *pgconn.PgConn, n *pgconn.Notice) {
+		*notices = append(*notices, n.Message)
+	}
+	conn, err := pgx.ConnectConfig(context.Background(), cfg)
+	if err != nil {
+		t.Fatalf("connect for notice capture: %v", err)
+	}
+	t.Cleanup(func() { conn.Close(context.Background()) })
+	return conn
+}
+
+// legacyNullThreadTask inserts a task through the normal path, asserts the
+// 451 trigger filled its thread scope, then NULLs the scope to simulate a
+// pre-451 legacy row.
+func legacyNullThreadTask(t *testing.T, agentID string, over testutil.Cols) string {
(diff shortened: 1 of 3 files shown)


multica-eve:
Thanks for the detailed investigation and test coverage. We are closing this PR because we do not want to reintroduce this data backfill through the startup migration path.

The original migration 451 backfill was deliberately removed in #8216 before the final v0.4.42 rollout because it coupled a recursive historical rewrite to deployment and carried unacceptable production locking risk. Migration 457 narrows the logical candidate set, but it is still a one-shot update of active task rows without verified production candidate counts or a production-statistics query plan, batching, rate limiting, or migration-local lock and statement timeouts.

The remaining legacy behavior is transitional and drains as those pre-451 tasks finish. Without evidence that a material affected set still exists, the benefit does not justify adding another automatic production migration. If a repair is needed later, it should begin with read-only candidate counts and `EXPLAIN`, then use a bounded, rerunnable operational backfill outside the startup path when the set is nontrivial.

Issue #6936, opened 2026-08-13
Publish dsh-multica-runtime to npm so self-hosted users can install the dsh profile


Frankli9986:
### Deployment type

- [x] self-host (local workspace agent)

### What do you want and why?

PR #6923 added DeepSeek Harness (dsh) as a first-party Multica runtime. That
integration is great — but for self-hosted/external users it depends on the
`multica` DSH profile coming from the `dsh-multica-runtime` plugin, and that
plugin is **not published to npm**: its `package.json` still has
`"private": true`, version `0.1.0-private.1`, a description that says "Private
DeepSeek Harness runtime bridge", and a stale `repository` URL pointing at
`dsh-external/dsh-multica-runtime` (the actual public repo is
`forrestchang/dsh-multica-runtime`). The README also still describes itself as
"Private, out-of-tree runtime bridge".

So although the source repository is now public, an external user who wants the
full dsh runtime (session resume, MCP injection, model/thinking discovery) has
no way to install the profile in one step:

```bash
# what we'd like to be able to do:
dsh plugin --profile multica add @multica-ai/dsh-runtime
# what we actually have to do today:
git clone https://github.com/forrestchang/dsh-multica-runtime
cd dsh-multica-runtime && pnpm install && pnpm build
dsh plugin --profile multica add /absolute/path/to/dsh-multica-runtime
```

This is a real usability gap for everyone who is not inside the Multica org:
the merged runtime feature is effectively unreachable unless you build the
profile bridge yourself.

### Proposed solution

1. Publish the plugin to npm under a stable name (e.g. `@multica-ai/dsh-runtime`
   or `@multica-ai/dsh-multica-runtime`) with a proper non-`-private` version.
2. Flip `"private": true` → `false` (or just drop the field), fix the
   `repository` URL in `package.json`, and update the README to describe the
   published install path:
   `dsh plugin --profile multica add @multica-ai/dsh-runtime`.
3. Consider documenting the self-host dsh runtime setup in the main Multica
   docs (the runtime discovery contract requires `--probe` to return protocol
   version 1, which the current code does).

For reference, I verified locally that the build-and-add path works on dsh
0.1.0-rc.6: `pnpm build` → `dsh plugin --profile multica add <path>` →
`dsh --profile multica --probe` returns `{"v":1,"type":"probe","runtime":"dsh",
"plugin_version":"0.1.0-private.1","protocol_version":1}` and
[...]

Issue #5205, opened 2026-07-10
fix(macOS): detect Codex CLI after Codex app moved to ChatGPT.app


kinpoe-ray:
## Bug

After updating Codex Desktop on macOS, the Multica daemon can retain and execute the old bundled CLI path:

`/Applications/Codex.app/Contents/Resources/codex`

The updated app now provides the bundled CLI at:

`/Applications/ChatGPT.app/Contents/Resources/codex`

Because the old path no longer exists, Multica agent chat/tasks fail before Codex starts.

## Error

```text
codex executable not found at "/Applications/Codex.app/Contents/Resources/codex":
exec: "/Applications/Codex.app/Contents/Resources/codex":
stat /Applications/Codex.app/Contents/Resources/codex: no such file or directory
```

## Environment

- macOS Apple Silicon
- Multica CLI: `0.3.30`
- Multica daemon: `0.3.28`
- Codex app bundle identifier: `com.openai.codex`
- Codex app bundle version: `26.707.31428` (build `5059`)
- Bundled Codex CLI: `codex-cli 0.144.0-alpha.4`
- Current executable: `/Applications/ChatGPT.app/Contents/Resources/codex`
- Multica daemon launched by macOS `launchd`

## Reproduction

1. Install/update the Codex Desktop app.
2. Start or keep running a Multica daemon that previously detected `/Applications/Codex.app/Contents/Resources/codex`.
3. Assign a Codex-backed agent task or send a chat message.
4. The daemon attempts to execute the removed old path and the agent fails immediately.

The old path is absent, while the new path runs successfully:

```text
/Applications/ChatGPT.app/Contents/Resources/codex --version
codex-cli 0.144.0-alpha.4
```

## Expected behavior

Multica should detect the current Codex App bundle path on macOS and avoid retaining a stale absolute executable path after an app update. A daemon restart should re-resolve the executable correctly.

## Workaround

Set the daemon environment variable to the new executable path:

```text
MULTICA_CODEX_PATH=/Applications/ChatGPT.app/Contents/Resources/codex
```

## Suggested fix

- Add `/Applications/ChatGPT.app/Contents/Resources/codex` to the macOS Codex Desktop fallback paths.
- Re-detect or validate the executable path when the daemon starts or before launching a task.
- Prefer a configurable/per-machine executable override over a stale hard-coded absolute path.

PR #8237, opened 2026-09-10
MUL-7232 feat(daemon): add dsh desktop(self-hosted) support

yswang1024:
## What does this PR do?

DeepSeek Harness (dsh) is the one built-in runtime whose CLI is not enough on its own: the
backend drives `dsh --profile multica --stdio`, and that profile — a DSH-side bundle that
implements Multica's JSONL protocol — is what supplies the protocol. Every other provider
drives a protocol its CLI already has. So dsh has two failure modes no other provider has,
and both were invisible:

1. **The binary was never found** when only DSH Desktop is installed (it ships its CLI inside
   the `.app` and never puts `dsh` on PATH), so a GUI-launched daemon and the login-shell
   fallback both missed it.
2. **The profile was missing.** `probeAgentCLIs` is documented as pure discovery, but dsh's
   runtime-profile gate lived *inside* it — one layer above the verdict machinery that turns
   a drop into a `/health` `skipped_agents` reason and a daemon-log line. The provider simply
   vanished from the runtime list with nothing anywhere saying why, which is the shape both
   #7721 and #6936 run into.

This PR fixes both, makes the daemon install the profile when an operator configures what to
install, and keeps the registered runtime in step with the profile on disk.

The approach is: **a gate belongs where its verdict can be reported, and a local precondition
whose changes are invisible to the provider set needs to be compared, not polled for changes.**

## Related Issue

Refs #7721, #6936 — this does not close either.

- #7721 (desktop, dsh on PATH, never detected): the silent drop both reports share is fixed.
  The Windows report itself was not reproduced here.
- #6936 (publish `dsh-multica-runtime` to npm): still required for the install step to work
  out of the box. This PR makes the daemon able to install *whatever* bundle the operator
  names, so #6936 becomes a one-line default change instead of a code change.

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/daemon/config.go` — `dshDesktopAppBundlePaths`: candidate macOS app-bundle
  locations for the CLI bundled in DeepSeek Harness Desktop, mirroring
  `codexDesktopAppBundlePaths` (#5205).
- `server/internal/daemon/agents_probe.go` — the app-bundle fallback in `probe()`, plus
  `executableCandidate`. The candidate is a Node script rather than a native binary, so it
  counts only while it is executable; an entry discovered without the executable bit would be
  advertised as healthy and fail on every spawn. The runtime-profile gate is removed from
  registration: discovery is now profile-blind (`probeAgentCLIs` is documented as pure
  discovery, and a gate is not discovery).
- `server/internal/daemon/daemon.go` — new verdict `builtinProbeMissingProfile` and its two
  reasons; the gate moved into `probeBuiltinRuntime`, ahead of version detection and behind an
  `agentExecutablePresent` check (the probe cannot tell "the profile refused" from "there was
  nothing to run", and would otherwise blame a missing profile for a path that vanished under
  us). The verdict is deterministic, so it is demotable immediately.
- `server/internal/daemon/dsh_profile.go` (new) — `MULTICA_DSH_PROFILE_BUNDLE` (npm spec,
  directory or tarball; candidates tried in order) and `MULTICA_DSH_PLUGIN_PATH`; the install
  itself; and the cheap profile-presence stat.
- `server/internal/daemon/agents_refresh.go` — `convergeAgentRuntimes` extracted so the ticker
  and a new kick share one backoff; a buffered, non-blocking `agentDiscoveryKick`; and
  `dshRuntimeProfileInconsistent`, which forces a round when the profile state and the
  registered runtime disagree. `convergeRuntimeRegistrations` now consumes the verdict map it
  used to discard, so a forced round reaches demotion.
- Tests: `agents_probe_dsh_test.go`, `dsh_profile_test.go` (new), `agents_refresh_test.go`.

## How to Test

1. **Discovery without a PATH entry.** On a machine with DSH Desktop installed and no `dsh` on
   PATH, `probeAgentCLIs()["dsh"]` resolves to the bundled CLI. Test:
   `TestProbeAgentCLIsUsesDshDesktopAppBundleFallback` (and the negative case:
   `TestProbeAgentCLIsIgnoresNonExecutableDshBundle`).
2. **The drop is reported.** With a dsh binary and no `multica` profile,
   `GET /health` carries
   `"skipped_agents": {"dsh": "the Multica runtime profile is not installed; run ..."}`, and
   the daemon logs one `WRN`. Test: `TestProbeBuiltinRuntime_DshWithoutMulticaProfile`.
3. **A removed profile takes the runtime offline within one discovery tick.** Reproduced on a
   real macOS host with the packaged desktop app: profile removed at `13:12:03` while dsh held
   a runtime; at `13:13:44` (first tick) the log shows
   `WRN skip registering runtime: DSH Multica runtime profile is not installed` and
   `WRN agent CLI is no longer usable on this machine; taking its runtimes offline
   runtime_ids=[...]`, and the workspace went from `runtimes=3` to `runtimes=2`.
   Tests: `TestDshRuntimeProfileInconsistent`,
   `TestConvergeAgentRuntimes_ForcedRoundWithNothingMissingKeepsTheBackoffClear`.
4. **A manual install is noticed just as fast** (the converge backoff used to grow to
   `agentConvergeMaxBackoff` = 30m over the rounds that could not register):
   `TestConvergeAgentRuntimes_ForceIgnoresThePendingBackoff`.
5. **Automatic install.** With `MULTICA_DSH_PROFILE_BUNDLE=<spec>`, a daemon that finds a dsh
   without a profile installs it and re-probes immediately. Verified end-to-end on macOS with
   the real dsh CLI, the real DSH-bundled pnpm and the real registry: install finished 2.3s,
   `provider=dsh` registered 0.85s later, no manual action. Tests:
   `TestProvisionDshMulticaProfile_*`, `TestStartDshProfileProvision_*`.

Full suite: `go test ./internal/daemon/ -count=1` passes on each commit's tree (38.8s / 41.2s
/ 40.5s), `go vet ./internal/daemon/` and `gofmt -l` clean, and the new tests pass 5× repeated.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — **N/A, no UI
      change.** Verified end-to-end in the packaged desktop app instead.
- [ ] I have updated relevant documentation to reflect my changes — **NOT DONE, see Risks.**
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to **landing copy**
      and **relevant docs** — **N/A, no new runtime or tab.** dsh already existed.
- [ ] If this PR touches Chinese product copy, I checked it against
      `conventions.zh.mdx` — **N/A, no user-facing copy.**

## AI Disclosure

**AI tool used:** Claude Code (model: Fable 5.1)

**Prompt / approach:** Started from a user report ("dsh runs in my terminal but Multica never
lists it"), traced it to the profile gate in `probeAgentCLIs` by reading the drop path and
comparing it with `/health`'s `skipped_agents` machinery, then verified every claim against a
real DSH Desktop install on macOS rather than against the code alone. Every fix in this PR
carries a test that fails before it — each one was checked by reverting the change and
confirming the red, and each commit's intermediate tree was compiled and tested before the
next was built.

## Screenshots (optional)
image
image

---

## Risks and open questions

**1. Documentation is missing for the two new variables.** `MULTICA_DSH_PROFILE_BUNDLE` and
`MULTICA_DSH_PLUGIN_PATH` are not in `apps/docs/content/docs/environment-variables.mdx`, which
documents the other `MULTICA_DSH_*` variables at line 252. That file exists in four languages
(en/zh/ja/ko). I left this out rather than editing one language and creating drift — happy to
do all four if that is preferred. `install-agent-runtime.mdx:58` also still has an unfilled
`<bundle>` placeholder that this PR's `MULTICA_DSH_PROFILE_BUNDLE` now makes concretely
nameable.

**2. `d.agents()` now always contains `dsh`.** Discovery no longer depends on the profile, so
`/health.agents`, `daemon status`'s `Agents:` line and `daemon probe-runtimes`'s
`provider_summary` will list a dsh that cannot run. Reproduced: a fake dsh that only refuses
`--probe` still reports `{"dsh":1}` from `probe-runtimes`. The authoritative "usable or not"
signals are now the workspace runtime list and `skipped_agents`. If the CLI surfaces should
carry `skipped_agents` too, that is a small follow-up — `daemon status` already fetches
`/health` (`checkDaemonHealthPort`).

**3. `LoadConfig` no longer errors on a machine whose only CLI is an unusable dsh.**
Previously `len(agents) == 0` triggered `"no agent CLI found: install ..."`. Such a machine now
starts and reports the reason on `/health` instead of failing to start. I think that is the
better trade, but it is a behavior change.

**4. The missing-profile verdict has no confirmation window**, unlike
`builtinProbeNotExecutable`, which requires `confirmNotExecutable`. A transient `--probe`
failure (5s timeout, `bin.js` being replaced during a DSH upgrade) is therefore treated as
confirmed. It is not rapid flapping — demotion is reached from the version-refresh tick or a
forced round, and recovery needs a later registration — but a confirmation window is
defensible if reviewers prefer symmetry.

**5. Demotion still takes up to one version-refresh tick (10m) when the profile is removed
*and* nothing is missing.** The forced round added here closes the case where the round would
otherwise never run, and makes it the discovery tick (2m) rather than the version tick, but
the underlying demotion cadence is unchanged and deliberately shared with the other confirmed
verdicts.

**6. `MULTICA_DSH_PLUGIN_PATH` defaults to a hardcoded macOS path** derived from the DSH
Desktop product name (`~/Library/Application Support/DSH Desktop/runtime-commands/bin`). It is
needed because DSH Desktop ships the pnpm that `dsh plugin` forwards to inside its own
runtime-commands directory and injects it only for processes it spawns; without it the install
fails with `pnpm not found on PATH` on exactly the machines this feature is for. Non-macOS
installs and renamed installations need the override.

**7. The install runs once per daemon.** A registry or network failure is logged with the
CLI's own output and not retried until a restart, so a transient failure leaves the profile
uninstalled rather than retrying every discovery round.

**8. One unreproducible test failure during local staging.** One full-suite run failed while I
was validating the third commit's tree; the failure was not captured and did not reproduce in
two subsequent full runs, and all of this PR's tests passed 5× repeated. Flagging it rather
than claiming a clean sheet — I could not attribute it.


Bohan-J (changes requested):
**Request changes: the product direction should land, but the current implementation treats every transient DSH probe failure as a definitive "profile not installed", which produces wrong behavior in three places — runtime demotion, automatic provisioning, and server-side task admission.**

The problem is real. On `main`, `dsh` is dropped from the discovery result entirely when the `multica` profile is missing, which is exactly the "the CLI is right there but the runtime disappeared" symptom in #7721; #6936 confirms there is no one-shot install path for the official profile yet. Separating "found the CLI" from "the CLI is usable by Multica", reusing the existing `skipped_agents` / runtime-demotion machinery, and adding the macOS Desktop bundle fallback all fit how the rest of the daemon already works. Net effect: macOS self-hosted members who only installed DSH Desktop get discovered; when the profile is missing, raw `/health` and the logs give a reason instead of silence; once an operator configures a bundle the daemon attempts the install; and adding/removing the profile re-converges the runtime. The blast radius is every self-hosted host with `dsh` installed — installation itself stays opt-in — but `health.agents` and plain `daemon status` will now list a `dsh` that is not yet usable, and a daemon whose only provider is an unusable `dsh` changes from "fails to start" to "starts successfully".

### Must fix

**1. Don't classify every `--probe` failure as "profile not installed."**

`probeDshMulticaProfile` returns a bare bool (`server/internal/daemon/agents_probe.go:344`): a timeout, a non-zero exit, malformed output, a future protocol version, and a genuinely absent manifest all collapse to `false`. That single `false` then flows into `builtinProbeMissingProfile` (`server/internal/daemon/daemon.go:2409-2412`), which is demotable immediately — unlike `builtinProbeNotExecutable`, which must clear `confirmNotExecutable` first (`daemon.go:2560-2561`) — and which may also kick off an install. So a momentary failure during a DSH upgrade, or an installed-but-protocol-incompatible profile, gets the runtime taken offline; with a bundle configured, it can also overwrite the user's existing installation. When the manifest is present but the probe fails, `dshRuntimeProfileInconsistent` (`server/internal/daemon/agents_refresh.go:157`, forced at `:88`) additionally forces a converge round every `agentDiscoveryInterval` (2m), bypassing the 30m `agentConvergeMaxBackoff`.

Please return a structured probe result: only a confirmed-absent profile should enter the missing/install path, while timeout, exec/parse failure, and protocol incompatibility get their own transient/incompatible verdicts, and destructive demotion at minimum reuses the existing confirmation window. Please add regression tests for "manifest present + timeout / non-zero exit / malformed output / future protocol version ⇒ no demotion, no provisioning".

**2. A missing-profile demotion looks to the server like a machine that is merely offline.**

`newRuntimeVerdict` only produces a structured `offline_reason` for `builtinProbeNotExecutable` (`server/internal/daemon/daemon.go:2310-2311`), so the missing-profile demotion calls Deregister with no reason attached. Server-side `runtimeVerdict` then classifies such a runtime as `AgentWaitable` / `runtime_offline` (`server/internal/service/agent_ready.go:129,144`), so assignments and @mentions bound to that agent keep queuing — even though, with no automatic install configured, that wait never resolves on its own. That conflicts with the semantics settled in MUL-6164: when a human has to intervene, refuse the trigger and say how to fix it. Please define a stable, structured offline cause for missing/incompatible profiles and plumb it through readiness, dispatch, and the localized messages; whether an in-flight automatic install is waitable should be modeled explicitly rather than falling out of an absent reason.

**3. Automatic provisioning is not tied to the daemon lifecycle and does not own its process tree.**

`startDshProfileProvision` launches its goroutine with `context.Background()` (`server/internal/daemon/dsh_profile.go:192`), and `provisionDshMulticaProfile` calls `exec.Cmd.CombinedOutput` directly (`:163`). On daemon shutdown or the 3-minute timeout, only the direct `dsh` process is signaled; the pnpm/node descendants it spawned can keep mutating `~/.dsh`, and rolling the daemon back does not undo that external write. The repo already has `server/internal/daemon/processtree.CombinedOutput` for exactly this — Unix process group and Windows Job Object whole-tree termination. Please wire the provisioning context to the daemon lifecycle, reuse that helper, and add a test asserting that after cancellation no grandchild survives and no further profile writes happen.

**4. Install inputs and the cross-platform environment aren't safe yet.**

Both the success and failure logs record the `bundle` spec verbatim, and the error string concatenates the full spec plus unbounded package-manager output (`server/internal/daemon/dsh_profile.go:165-171`). An npm URL can carry credentials, and a local path can leak a username — this should follow the existing agent-command logging convention of redacting user-supplied arguments and bounding captured output. Separately, `dshPluginEnv` matches only a case-exact `PATH=` prefix (`dsh_profile.go:125-138`); on Windows the common `Path=` is missed, and the appended `PATH` then wins under Go's case-insensitive Windows de-duplication, leaving only the plugin directory. Please merge the environment case-correctly so the original PATH survives, and add Windows coverage — every provisioning test currently skips on Windows (`dsh_profile_test.go:52, 68, 112, 130, 153, 192`).

**5. The new operator configuration has no product surface.**

`MULTICA_DSH_PROFILE_BUNDLE` and `MULTICA_DSH_PLUGIN_PATH` exist only in source; they appear in none of the four localized environment-variable docs, and the install docs still show a non-executable `<bundle>` placeholder. Automatic provisioning is a new, persistent feature surface, so at minimum it needs documented legal values, the opt-in / once-per-daemon retry semantics, how to diagnose a failure, how to clean up or roll back, and the Windows/macOS differences. Without that there is no supported way for an admin to configure it.

### If this judgment is wrong

The first symptoms would be a DSH runtime demoted during an ordinary upgrade, triggered tasks queueing until they expire, or pnpm still writing the profile after the daemon has stopped. Affected: self-hosted hosts with `dsh` installed; the automatic writes only affect operators who set a bundle. A code rollback stops the new behavior but cannot undo a profile already written into the DSH home.

### Verification (PR head `e91ca8940`)

- `env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/ -count=1` — pass (38.852s; the first run failed only because this agent's injected task config root polluted the env-isolation tests)
- `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` — pass
- `go test ./internal/daemon/processtree/ -count=1` — pass
- `env -u MULTICA_TASK_CONFIG_ROOT go test -race ./internal/daemon/ -run 'Dsh|ConvergeAgentRuntimes' -count=1` — pass
- `go vet ./internal/daemon/`, `git diff --check` — pass
- CI is currently green. Real DSH Desktop and Windows provisioning were not verified locally. The existing green tests do not cover the misclassification, server admission, grandchild cancellation, or `Path=` cases above.

yswang1024:
# Review reply

Thank you — the classification critique was right, and following it turned up three further
defects that the original design hid. All five must-fix items are addressed; one deviates from
the literal wording and is argued below rather than quietly changed.

Seven commits on top of the previous head:

```
af4ca3c80  refactor(agent): export the DSH stdio protocol version
e475b9230  fix(redact): mask credentials carried in URL userinfo
4141b6a69  fix(daemon): decide the DSH runtime-profile verdict on evidence
666ba52de  feat(daemon): discover the CLI shims DSH Desktop generates
5be583e2a  fix(server): refuse a missing DSH runtime profile under its own reason
e8f4746f8  docs(runtime): document DSH profile provisioning and shim discovery
fcda24327  fix(daemon): find DSH Desktop's bundled pnpm on Windows too
```

---

## 1. Don't classify every `--probe` failure as "profile not installed"

`probeDshMulticaProfile` now returns `dshProbeVerdict` — `OK` / `MissingProfile` /
`Incompatible` / `Unavailable` — and the rule is:

| Observation | Verdict |
|---|---|
| A frame with the protocol this backend drives | `OK` |
| A frame with any other protocol or runtime | `Incompatible` |
| No usable frame, round cancelled | `Unavailable` |
| No usable frame, manifest **absent** | `MissingProfile` |
| No usable frame, manifest **present** | `Unavailable` |

**A manifest that is present keeps every failure transient**, which is the protection you asked
for: a probe that times out or exits non-zero during a DSH upgrade cannot demote the runtime and
cannot start an install over an existing profile.

Two details differ from what I would have written from the code alone, and both came out of
running the real CLI:

**The frame decides, not the exit status.** A profile that printed a valid frame has answered,
whatever the process did on its way out. This also removes a latent bug: `processtree` reports a
lifecycle error for a tree that outlived its leader, which says nothing about the protocol and
would have been read as a failed probe.

**A timeout is not transient by itself.** Asked for a profile it does not have, DSH does not
refuse `--probe` — it hangs, printing nothing, until something kills it. Measured on a real
install (macOS 2.0.5, profile absent): still running after 10s, zero bytes on both streams.
Classing a timeout as transient therefore made a missing profile *undetectable* on exactly the
host the install exists for: the daemon reported "the probe returned no usable answer", never
started the install, and exited for want of a runtime. That is what the first Windows run did.
The manifest is the same file DSH's own `loadProfile` consults before refusing to boot, so a
confirmed-absent one is evidence rather than a guess. `dshProbeTimeout` also went 5s → 15s: the
healthy path boots the Desktop app's Electron binary as node, and 5s cut that short on a cold VM.

**Demotion.** `notExecutableConfirmWindow` became `condemnedConfirmWindow` and now gates every
condemnable verdict, keyed per verdict so a provider moving between states starts a fresh clock
instead of inheriting one earned by a different problem. The demotable set is now a named
predicate (`demotableBuiltinProbeVerdict`) rather than an inline disjunction, so a verdict added
later has to be considered rather than inherited.

**Deviation — `Incompatible` is not demotable at all**, not merely confirmation-gated. Your test
spec says "future protocol version ⇒ no demotion", and I agree, but for a reason worth stating:
a protocol this daemon does not drive can equally mean *this daemon* is the stale side of the
skew. Tearing down a live runtime because this build is behind is a destructive answer to a
version question. It still blocks a fresh registration and is reported on `/health`. The cost is
real and I want it on the record: a registered dsh whose profile is *downgraded* below the
protocol keeps taking work and fails those tasks loudly, rather than going offline. If you would
rather it demote after confirmation, that is a one-line change to the predicate.

**The 2-minute forced round.** Fixed, and fixing it naively deadlocked. `dshRuntimeProfileMismatch`
now returns the *kind* of disagreement, and only `profile-without-runtime` — the direction no
round is guaranteed to resolve — is throttled to one forced round. `runtime-without-profile` must
force on **every** tick, because condemning it takes two probe rounds and nothing else schedules
the second: with dsh still registered nothing is missing a runtime, so an unforced tick returns
from `convergeAgentRuntimes` without probing at all. Throttling both directions (my first attempt)
left a registered dsh whose profile was deleted online forever — it passed every existing test.
`TestDshRuntimeProfileMismatch_ForcingConvergesEachDirection` asserts the force sequence, not the
mismatch kind, so that trap cannot be re-entered silently.

Regression tests for the matrix you named:

- `TestProbeDshMulticaProfile` — 13 cases including manifest-present timeout / non-zero exit /
  unparseable output / exit-0-silent / future protocol / foreign runtime, manifest-absent hang and
  silent exit, a valid frame followed by a non-zero exit, and a cancelled round.
- `TestProbeBuiltinRuntime_DshFailureWithProfileInstalledIsNotCondemned` — the four failure modes
  with the profile installed **and a bundle configured**, asserting via
  `demotableBuiltinProbeVerdict` that no route reaches a demotion, and that nothing was installed.
  It asserts through the predicate rather than naming verdicts, so a verdict added to the demotable
  set later fails this test instead of slipping past a list of names.
- `TestProbeBuiltinRuntime_DshThatCannotRunIsNotAProfileVerdict` — see item 1b below.

### 1b. A CLI that cannot run is not a profile verdict

Not in your list, but the same defect class, and Windows hit it first. The profile probe ran
*before* version detection, behind an `agentExecutablePresent` guard — and `exec.LookPath` is far
too weak a proxy for "this CLI works". On Windows the CLI is a `.cmd` shim that LookPath resolves
happily and that exits **9009** (cmd.exe's "command not found") when what it forwards to is
missing. The daemon reported that machine as *the Multica runtime profile is not installed* — a
repair for a problem it did not have, and with a bundle configured it would have started
installing into a DSH that cannot execute.

The profile probe now runs only after version detection succeeds. A CLI that cannot answer
`--version` is not one this daemon has any business holding an opinion about the profile of; its
failure is reported as the version-probe failure it is. Side benefit: a too-old dsh without a
profile now reports "version too old" instead of "missing profile".

---

## 2. A missing-profile demotion looks like a machine that is merely offline

New offline code `dsh_profile` on the daemon side, and a new dispatch reason
`runtime_profile_missing` on the server side, plumbed through readiness, dispatch and the
localized copy in all four languages.

It is **not** folded into `runtime_unusable`, and that matters more than it first looks. The two
causes have opposite repairs: an unrunnable CLI is reinstalled, while a CLI missing its runtime
profile is working perfectly and reinstalling it changes nothing. Sharing one code meant every
client told DSH users to reinstall a CLI that was never the problem — the chip and toast read
"reinstall it there", and the durable system comment left on the issue opened with *"its CLI is
installed but cannot be executed on that machine"*, blamed a blocked npm postinstall, and rendered

```bash
dsh plugin --profile multica add <bundle>
```

inside a fenced block as the line to run on that machine. A placeholder presented as a command is
the shape of instruction people paste verbatim and then report as broken, so `Repair.Command` is
now empty for this cause and `Repair.Package` names what is missing; the prose that explains it
lives in the notice, where it can name both ways to supply a real bundle without printing
something to copy.

**Waitability is modelled explicitly**, as you asked: `RuntimeOfflineReason.Installing` says the
daemon is installing the profile right now, which is the one case where queueing is correct.

**And it is withdrawn when the install gives up.** Writing that flag surfaced a defect in my own
fix: the runtime row is written exactly once, by the demotion — which also removes the runtime
from the index, so no later round has a runtime to condemn and no later deregistration is sent. An
install that failed after that point left every trigger for that agent queueing forever behind an
install that had already stopped, which is precisely the failure the structured reason exists to
prevent. `withdrawDshInstallWait` re-sends the corrected reason, skipping any runtime that has
since been re-registered.

Tests: `TestNewRuntimeVerdict_DshProfileCarriesAStructuredCause`, `TestAgentReadinessVerdict`
(blocked/`runtime_profile_missing`, the in-flight-install exception, and the notice text asserting
it contains no `<bundle>`, no fenced block and no "cannot be executed"),
`TestWithdrawDshInstallWait_*`, and `blocked-trigger-copy.test.ts` asserting the label is distinct
from `runtime_unusable` and contains no "reinstall".

---

## 3. Provisioning is not tied to the daemon lifecycle and does not own its process tree

Both fixed. `startDshProfileProvision` uses the context `Run` was handed (kept behind `d.mu` with
an accessor, so "the write happens before every reader starts" is not a property a later edit can
quietly break), and both the install and the `--probe` call go through `processtree`.

`TestProvisionDshMulticaProfile_CancellationKillsTheWholeTree` publishes the grandchild's pid
before the parent settles into its own wait, so the assertion is about survival rather than about
a race — an unsynchronised grandchild can be killed before it publishes anything, which would let
the test pass without testing anything. It then asserts both that the pid is gone and that the
late write never landed.

Two further things came out of this:

**Success is the manifest on disk, not the exit status.** A candidate that exits 0 without
producing a profile leaves the operator exactly where a failure does, so it is a failed candidate
— logged as one, and the next candidate still gets its turn (previously the first exit-0 ended the
loop). `dshMulticaProfilePresent` is the fact the rest of the daemon judges by; an install allowed
to disagree with it would log "installed the DSH runtime profile" and leave every later round
reporting that same profile as missing.
(`TestProvisionDshMulticaProfile_ExitZeroWithoutAProfileIsAFailedCandidate`)

**Startup could never let the install finish.** On a host whose only provider is a dsh without the
profile — the host this feature exists for — the sequence is 32ms long: the probe reports the
missing profile and starts the install, registration finds nothing to register, `Run` returns that
error, and the process exits, killing the install's process tree with it. Every restart repeats
it. `startupMayProceedWithoutRuntimes` allows startup to continue for that one case and nothing
else; an unreachable server, a rejected token and a genuinely empty machine all still fail fast.
Because such a daemon tracks no workspace, a finished install picks the workspace up before
converging, or the runtime would wait for the 30-minute consistency sync.
(`TestStartupMayProceedWithoutRuntimes`, `TestRegisterAfterDshProfileInstall_*`)

---

## 4. Install inputs and the cross-platform environment

**Logging.** The bundle spec no longer reaches any log line — candidates are named by position
("candidate 2 of 2") — and captured output is redacted and bounded to 2 KiB from the tail, cut on
a rune boundary because package-manager output is full of box drawing and non-ASCII paths and a
byte cut produces invalid UTF-8 that downstream JSON encoders rewrite.

**Redaction had a hole worth flagging on its own.** `redact.Text` did *not* mask
`https://user:token@registry/...` — its connection-string rule covers postgres/mysql/mongodb/
redis/amqp only. Measured before the fix: the token survived verbatim. That shape is exactly what
your review was about, since a registry echoes the request URL back in its error text. Added a
scheme-agnostic userinfo rule that keeps the scheme and host (which registry refused is what an
operator reads, and is not the secret) with tests in both directions — `https://host:8443/p`,
`ssh://git@host/p` and a bare `a:b` in a path must not trip it. This is the one change in the PR
that affects every provider's output, not just dsh, so it is worth a second look.

**PATH merge.** Matched case-insensitively, preserving the existing entry's own spelling so the
merge cannot itself introduce a second entry.

**Windows coverage.** The provisioning tests that need a shell fixture still skip on Windows —
a `.cmd` cannot stand in for pnpm — so instead the logic that was Windows-specific was made
host-independent and is now asserted from any host: `TestMergeDshPluginPath` (PATH / Path / path),
`TestDshProvisionCommand` (argv and env without a fixture), `TestDshDesktopShimCandidates_*` and
`TestDshPluginPathDirs_*` (the Windows layouts, built in a temp directory with `goos` passed as an
argument rather than read from `runtime.GOOS`). The alternative leaves those branches testable only
on Windows, which in this package has meant untested.

---

## 5. The new operator configuration has no product surface

Documented in all four languages, on both pages.

**Environment variables** — the install semantics, which previously existed only in source, and
which on the first pass I attached to the *wrong variable* (the sentences describing what
`MULTICA_DSH_PROFILE_BUNDLE` does followed the `MULTICA_DSH_PLUGIN_PATH` sentence, so "the variable
is unset by default and the install it enables is bounded" read as being about a variable that
enables no install). Now: at most once per daemon lifetime, only on a confirmed-absent profile,
failures logged with the package manager's own output and retried only after a restart, a timeout
or an unsupported protocol version reported and never installed over. Plus how to diagnose (search
the daemon log for `DSH runtime profile`) and how to roll back (delete `$DSH_HOME/profiles/multica`).

**Install guide** — what to actually put in the variable, which is the part that was missing.
The bridge is not on a public npm registry yet (#6936), so a self-hosted team builds it from the
repository and points at the build directory, or at the tarball `npm pack` produces when the daemon
host is not the build host; a package name becomes the simplest option once it is published. With a
warning specific to this value: it is installed into every daemon host's DSH home without further
confirmation, so an unvetted third-party package is a supply-chain decision, not a convenience.

The `<bundle>` placeholder is gone from the repair command entirely (see item 2). It remains in the
install docs, where it is prose with the shapes it can take spelled out next to it.

---

## Discovery: Windows, and a macOS gap it exposed

The macOS-only app-bundle fallback is replaced by one layout that serves both platforms, because
the original could not have worked on Windows — every candidate was an `/Applications` path.

DSH Desktop generates CLI shims under its own per-user data directory
(`~/Library/Application Support/DSH Desktop` on macOS, `%APPDATA%\DSH Desktop` on Windows). On
Windows that is roaming application data even though the app installs to Program Files or
`%LOCALAPPDATA%\Programs`, which is the detail that makes this un-guessable; one lookup covers both
install modes, since a per-machine install still generates these shims per user. Payload
directories are content hashes (and, under `host-commands`, a hash plus a generation uuid), so they
are enumerated rather than composed, newest first, with the name breaking timestamp ties because
`ReadDir` order is not a promise and the daemon must not pick a different CLI on different rounds.

**`host-commands` outranks `cli`, and that ranking is load-bearing.** Nothing at this layer executes
a candidate, so a broken-but-present shim is indistinguishable from a working one — and on the
Windows install we tested, the `cli` shim was exactly that: resolvable by `exec.LookPath`, exiting
9009. Ranking is the only lever available.

**A macOS gap fell out of the same work.** Generated shims now outrank the app-bundle script on
macOS too, because that script is entered through `#!/usr/bin/env node` and so only works where node
happens to be installed, while a generated shim runs the Desktop app's own runtime by absolute path
and needs nothing on PATH. That was latent on macOS and would have surfaced as the same class of
report.

`executableCandidate` had to learn the platform difference for any of this to work: Go synthesises a
regular file's mode from the read-only attribute and never sets an executable bit on Windows, so the
Unix test rejected *every* candidate and the whole fallback would have been dead code there. Windows
uses `exec.LookPath`, the same check the launcher makes — which is also why the Node script has no
Windows counterpart and `.js` never appears in that list.

The same enumerator now serves the pnpm lookup, which was composing a single macOS path and was
commented as macOS-only. The two platforms do not agree on the shape — macOS 2.0.5 keeps a flat
`runtime-commands/bin`, a Windows install keeps `runtime-commands/generations/<id>/bin` — so
composing either leaves the other failing with "pnpm not found on PATH" on exactly the hosts the
automatic install is for.

Only default locations are covered, and nothing reads an install receipt or the registry, so a
relocated install still needs `MULTICA_DSH_PATH`.

---

## Behaviour changes (blast radius)

Your original note covered discovery; these are the ones this round adds. The first two apply to
**every** dsh, npm-installed or Desktop:

1. **A missing profile no longer demotes immediately** — it now clears the same confirmation window
   as `not_executable`, so a removed profile takes two probe rounds (~2–4 min) to take the runtime
   offline instead of one.
2. **An incompatible profile no longer demotes at all** (argued in item 1).
3. **`dshProbeTimeout` 5s → 15s.** A healthy profile answers immediately and is unaffected; a host
   *without* the profile now spends 15s per discovery round in a hung probe, because DSH hangs
   rather than refusing. Bounded by construction — the state ends when the profile is installed.
4. **A daemon can now start with zero runtimes**, but only while it is installing the profile
   itself. Documented in the install guide's startup callout.
5. **`redact.Text` masks URL userinfo for every provider**, not just dsh.

The Desktop-specific additions — shim discovery and the pnpm lookup — cannot affect an
npm-installed `dsh`: the shim fallback is reached only when `exec.LookPath` *and* the login-shell
resolution both miss, and the pnpm lookup runs only under `MULTICA_DSH_PROFILE_BUNDLE`.

---

## Verification

Host: macOS 26 (arm64), Go 1.26.8.

```
env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/ -count=1                 pass  38.889s
env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/processtree/ -count=1     pass   0.661s
env -u MULTICA_TASK_CONFIG_ROOT go test ./internal/daemon/execenv/ -count=1         pass  65.882s
env -u MULTICA_TASK_CONFIG_ROOT go test ./pkg/agent/ ./pkg/redact/ -count=1         pass
env -u MULTICA_TASK_CONFIG_ROOT go test -race ./internal/daemon/ \
  -run 'Dsh|Startup|RegisterAfter|Provision|Withdraw|Probe|Converge' -count=1       pass  10.162s
go test ./internal/service/ -run 'AgentReadinessVerdict|DispatchFailReasonCode' -count=1  pass
go build ./... && GOOS=windows go build ./... && go vet ./...                       pass
gofmt -l, git diff --check                                                          clean
pnpm typecheck                                                                      9/9
pnpm --filter @multica/views exec vitest run issues/                                935 tests
pnpm --filter @multica/docs build                                                   pass
```

`internal/service` as a whole needs a database and was not run; the three tests above are the ones
this PR touches.

**Windows** — verified end to end on a real DSH Desktop install: Windows 11 x64,
with a packaged desktop build of this branch (`0.4.42-12-gfcda24327`, x64 NSIS), pointed at a
local API server.

- Discovery finds the generated shim at
  `%APPDATA%\DSH Desktop\host-commands\desktop\generations\<hash>-<uuid>\bin\dsh.cmd`, and
  version detection succeeds against it.
- With the profile absent, the daemon reports the missing profile rather than a transient probe
  failure, and starts rather than exiting.
- With `MULTICA_DSH_PROFILE_BUNDLE` set, the install runs with the bundled pnpm found under
  `runtime-commands\generations\<id>\bin`, and dsh registers once it finishes.

Three defects were found *on that host*, and each was reproduced from its daemon log before the fix
and re-checked after: the 9009 misclassification (item 1b), the hung-probe misclassification
(item 1), and the startup deadlock (item 3). None of them were reachable from the test suite as it
stood — the first two need a `.cmd` shim, the third needs a host whose only provider is dsh — which
is why each now has a regression test built from what the log showed.

Bohan-J (changes requested):
**Request changes: two items left.** Thanks for the thorough rework and the end-to-end run on Windows — it addresses everything in the previous review. What remains is one correctness issue in the new `installing` state and keeping the PR scoped to DSH. Nothing else is required for merge.

### 1. The `installing` claim can outlive the install

`Installing: true` tells the server to queue triggers instead of refusing them, but three paths can leave it on a runtime row after the install has already ended:

- **Race between verdict and wait.** `detectBuiltinRuntimes` reads `dshInstallInFlight` when it builds the verdict (`server/internal/daemon/daemon.go:2762`), but the wait is only recorded later, in `demoteUnusableRuntimes` (`server/internal/daemon/agents_refresh.go:593`). If the install fails in between, `withdrawDshInstallWait` (`server/internal/daemon/dsh_profile.go:345`) finds an empty list, and the `installing: true` written afterwards is never withdrawn.
- **Best-effort withdraw.** A failed Deregister in the withdraw is only logged (`agents_refresh.go:687`), and during shutdown it runs on the already-cancelled lifecycle context.
- **Restart mid-install.** The row keeps `installing: true`, and the new process doesn't track that runtime, so nothing withdraws it.

In all three cases `runtimeVerdict` keeps treating the agent as waitable (`server/internal/service/agent_ready.go:175`), and triggers queue behind an install that is no longer running — the outcome `Installing` was introduced to prevent.

Suggested fix: make the claim bounded instead of trying to guarantee the withdraw. `SetAgentRuntimeOfflineWithReason` already stamps `updated_at = now()`, so `runtimeVerdict` can honour `Installing` only while `rt.UpdatedAt` is within a window a bit longer than `dshProvisionTimeout` (say 5 minutes), and fall back to `runtime_profile_missing` after that. That covers all three paths in a few lines plus a test; the withdraw can stay as the fast path. Dropping the `Installing` exemption entirely — refusing during the install, which the notice already tells users to retry after — is also acceptable, and would let `dshInstallWaits` / `withdrawDshInstallWait` go.

### 2. Keep the PR scoped to DSH

Two changes reach beyond DSH. Both were introduced for DSH, but as written they change behaviour for every provider:

- **Global URL-credential redaction** (`server/pkg/redact/redact.go:82`). `redact.Text` runs over every agent's task messages, comments and error text, so this alters output for all providers. The gap you found is real (and the previous review prompted it), but a change to shared redaction deserves its own PR and review. Here, please scrub URL userinfo locally for the provisioning output in `dshProvisionOutput` (`server/internal/daemon/dsh_profile.go:190`) and leave the global pattern list unchanged — or land the rule as a separate PR first and rebase on it.
- **Demotion from the converge path** (`server/internal/daemon/agents_refresh.go:824-834`). On `main` only `refreshAgentVersions` acts on demotable verdicts; with this change every converge round does, for every provider, so Claude Code, Codex and the rest would be taken offline on a different schedule. The forced DSH rounds only need this for `dsh`, so please restrict that call to the dsh verdict and leave other providers on `main`'s behaviour.

### On your open question

Keeping `Incompatible` non-demotable is fine as argued; no change needed.

Checked at `fcda24327`.

yswang1024:
# Review reply

Both items done.

```
207109080  fix(server): bound the in-flight DSH install claim instead of trusting it
a87481daf  refactor(daemon): scrub URL credentials in provisioning output, not globally
8d127682e  fix(daemon): let a converge round condemn dsh only
```

## 1. The installing claim can outlive the install

Took the bounded option. `SetAgentRuntimeOfflineWithReason` does stamp `updated_at = now()`
(`runtime.sql:246`), so:

```go
const runtimeInstallClaimWindow = 5 * time.Minute

func installClaimHolds(reason runtimeOfflineReason, rt db.AgentRuntime) bool {
	if !reason.Installing || !rt.UpdatedAt.Valid {
		return false
	}
	return time.Since(rt.UpdatedAt.Time) < runtimeInstallClaimWindow
}
```

All three paths you listed are covered by the floor, and `withdrawDshInstallWait` stays as the
fast path. The window is longer than `dshProvisionTimeout` (3 min) on purpose; that coupling is
now noted in Risks, since a longer provisioning timeout would have to move the window with it.

One branch you did not name, decided fail-closed: **a row with no usable `updated_at` does not
hold the claim.** A claim that cannot be bounded, if honoured, degrades to the original failure —
work queued until it expires. Refusing during a genuine install costs one retry, which the notice
already asks for.

I did not take the "drop the exemption entirely" option. Refusing a trigger with "install the
profile by hand" *while the daemon is installing it* is the same class of wrong-message problem
this review has been correcting, and the queue is the whole point of knowing an install is running.

`TestAgentReadinessVerdict` now asserts three states: fresh → waitable/`runtime_offline`, stale →
blocked/`runtime_profile_missing`, no timestamp → blocked.

## 2. Keep the PR scoped to DSH

**Redaction.** `pkg/redact` is back to `main` byte for byte (`git diff main -- server/pkg/redact/`
is empty). The rule moved into `dshProvisionOutput`, where the shared redactor still runs after it
for the token shapes it already knows. Log behaviour for provisioning is unchanged; every other
provider's output returns to `main`'s. The tests moved with it, plus the false-positive cases a
local rule still has to get right — a port, an scp-style git remote, a bare colon in a path — and
one asserting the host survives, since which registry refused the request is what an operator reads
and is not the secret.

The gap in the shared list is still real, and I would rather not lose it: I have left it as an open
risk on the PR so it can be picked up as its own change.

**Converge demotion.** Narrowed to dsh via `dshOnlyVerdicts`. Every other provider is back on
`main`'s behaviour exactly — demoted from `refreshAgentVersions` alone, on its own ten-minute
cadence. The comment records why dsh is the exception rather than leaving it to be re-derived: it
is the only built-in whose precondition can change without any version changing, which is why it is
the only one that needs a round it can be forced into. `TestDshOnlyVerdicts` pins it.

## Verification

Full set re-run at `8d127682e`; commands and results are here:

```
go test ./internal/daemon/ -count=1                                    pass  38.6s
go test ./internal/daemon/processtree/ -count=1                        pass   0.7s
go test ./internal/daemon/execenv/ -count=1                            pass  65.9s
go test ./pkg/agent/ ./pkg/redact/ -count=1                            pass
go test -race ./internal/daemon/ \
  -run 'Dsh|Startup|RegisterAfter|Provision|Withdraw|Probe|Converge'   pass  10.2s
go test ./internal/service/ \
  -run 'AgentReadinessVerdict|DispatchFailReasonCode|RuntimeUnusableNotice'  pass
go build ./... && GOOS=windows go build ./... && go vet ./...          pass
gofmt -l (touched dirs), git diff --check                              clean
pnpm typecheck                                                          9/9
pnpm --filter @multica/views exec vitest run issues/                    935 tests
pnpm --filter @multica/docs build                                       pass
```

multica-eve (approved):
Thank you for the thorough implementation and for addressing the review feedback so carefully. The final head keeps the behavior scoped to DSH, bounds the in-flight install claim, and preserves existing behavior for other providers. Approved for merge.

PR #8238, opened 2026-09-10
MUL-7209 fix(lark): reply natively and mention the asker in ordinary groups

### LINKED ISSUE #8234: [Bug]: Lark ordinary group responses lack native reply and sender mention
**Deployment type:** Official Multica Cloud (`multica.ai`)

## Problem

In an ordinary Feishu/Lark group chat, a member can @-mention a Multica bot and the agent executes successfully and sends the result back to the same group. However:

1. The outbound message is not a native reply to the triggering message.
2. The displayed `@username` is plain text, not a native Feishu mention, so the member receives no @ notification.

Inbound routing and agent execution are working. The failure is limited to outbound reply/mention semantics.

## Reproduction

1. Bind a Multica agent to Feishu using the official Cloud integration.
2. Add the bot to an ordinary group chat (not a topic/thread scenario).
3. From a bound Multica member account, send `@Bot <request>` using a real Feishu mention.
4. Wait for the successful agent result.

## Actual result

- The result lands in the group as a standalone message rather than a native reply to the triggering message.
- The leading `@member` is ordinary black text and does not trigger a Feishu mention notification.

## Expected result

- Preserve the inbound triggering `message_id` as the outbound reply target and use Feishu's message reply API for ordinary group-chat responses.
- Render the triggering member as a native mention using the sender's Feishu `open_id` (or the identifier required by the selected API), rather than concatenating `@display_name` into text.
- Keep existing topic/thread routing behavior intact.
- If sender identity is unavailable, degrade safely without mentioning a different member.

## Suspected implementation gap

Observed behavior suggests that a reply target is currently set for topic/thread scenarios but not for ordinary group messages. Please verify this in source before treating it as the root cause.

The fix should live in the Lark outbound adapter/message serialization layer: carry the original `message_id` and sender identity from the inbound event, call the native reply path, and serialize a structured mention. Prompt-generated `@name` text is not equivalent.

## Relation to #6266

Related, but not a duplicate of #6266. That issue asks to expose channel identifiers to the agent runtime. This report concerns the platform-controlled outbound patcher/adapter: Multica already knows enough to route the result back to the originating chat, but the returned message lacks native reply and mention semantics. The platform should be able to implement this without exposing credentials to agents.

## Acceptance criteria

- In an ordinary Feishu group, the result appears as a native reply to the triggering message.
- The triggering member is shown as a clickable native mention and receives the @ notification.
- Topic/thread replies continue to land in the correct topic.
- Tests cover a resolved sender `open_id`, missing sender identity, and concurrent triggers from different members without cross-mentioning.

Bohan-J:
Fixes #8234.

In an ordinary Feishu group, the agent's answer arrived as a standalone message opening with a plain-text `@name` the model had written itself. It neither attached to the question that produced it nor notified the member who asked. Two independent gaps.

## 1. Native reply

The trigger message id was already recorded on every inbound turn, but `threadReplyTarget` required a thread id *alongside* it. An ordinary group has no thread id, so it discarded a usable target and fell through to the chat-level send.

Now the thread id only decides `reply_in_thread`; a trigger message id on its own is enough to reply. The transport already supported this (`outboundMessageRequest` switches to `/im/v1/messages/{id}/reply` whenever the target is set), so this is a routing-condition change, not new plumbing.

**p2p is deliberately unchanged** on the chat-level send: a 1:1 has no competing questions for a quote to disambiguate. `inboundReplyTarget` (the synchronous `OutcomeReplier` — `/issue` confirmations, offline/archived notices) gets the same three cases, so a user cannot tell which path answered them from where the reply lands.

## 2. Native mention

The sender's `open_id` was parsed on inbound but never persisted; by the time the outbound patcher ran, identity had been collapsed to a Multica UUID.

Getting this right took three attempts, and the two rejected ones are the useful part of the review history:

- **Resolve the open_id at send time from the task's `initiator_user_id`.** Unsound: `channel_user_binding` is unique on `(installation_id, channel_user_id)` and *not* on `(installation_id, multica_user_id)`. The redeem path plain-INSERTs when a member binds a second Feishu account on one installation (the `ON CONFLICT` only fires for a repeat of the same `channel_user_id`), so the reverse lookup had two rows and no ordering — a reply to account B could mention account A.
- **Record the trigger on `channel_chat_session_binding`.** Also unsound: that table holds one row per `chat_session` and remembers only the *latest* trigger. The debouncer keys timers on `(chat_session, context revision)`, so a run batched in revision 1 is enqueued *after* a `/clear` has opened revision 2 and revision 2's message has committed — freezing the newer generation's message and sender onto the older run. A's answer quotes and @-mentions B. No unusual interleaving required: only B's append committing inside A's 3s debounce window.

**What actually shipped:** the trigger — the message an answer quotes, the thread it replies into, and the channel-native sender it @-mentions — is snapshotted on `channel_chat_context_generation`, keyed `(chat_session_id, revision)`. That table already snapshots `initiator_user_id` against exactly this hazard; `SetChannelChatContextInitiator`'s comment names it for crash recovery. `CreateChannelTaskDeliveryFromSession` takes the context revision and freezes that generation's trigger onto the task.

All three values move in one statement, so a sender or thread can never describe a different message than the reply targets.

**The thread sits with the trigger, not the route** — a distinction worth stating because it is easy to get backwards, and I did get it backwards once mid-review. For a thread-*isolated* session (Lark topic, Slack channel thread) there is one binding per thread, so binding and trigger coincide and the thread looks like route data. Slack DMs break that: `slackSessionRouting` keeps ONE binding for the whole DM channel while replying into whichever thread the member used, so the binding cursor names the thread that spoke *last*. Reading the thread from the binding hands a debounced revision-1 run a later revision's thread. Taking it from the generation is correct for both shapes.

The route is what remains: chat, type, config, revision — properties of the session that do not vary by generation.

Two properties follow, and both are load-bearing:

- The mention names the account that sent *this* trigger — not the member it resolved to, so a second bound account is unreachable from this path; and not the session's newest trigger, so a later generation cannot claim an older run's answer.
- The same holds for the message quoted and the thread replied into.

The snapshot is written under the same `cmd == nil` guard as the initiator: a message excluded from agent input is not the question being answered, so it must not become the message the answer quotes.

The binding keeps `last_message_id` / `last_thread_id` — those remain a genuine per-session cursor driving the history-boundary bookkeeping.

Identity is never inferred from the `@name` in the agent's prose; duplicate names, nicknames and a model that guesses wrong would all ping the wrong colleague. With no usable `open_id` the answer goes out **unmentioned**. Mentions are group-only, for the same reason replies are.

The two outbound wire shapes spell mentions differently — `<at user_id="ou_x"></at>` for `msg_type=text`, `<at id=ou_x></at>` for the schema-2.0 card markdown. The wire shape is picked from the agent's own content *before* the mention is attached, so a mention can never flip a plain-prose answer onto the card path.

## Migration 460

Adds `last_message_id` / `last_thread_id` / `last_sender_id` to `channel_chat_context_generation`, plus `channel_sender_id` on `channel_task_delivery`. Purely additive DDL — nullable, no index, no data rewrite.

**Pre-migration rows intentionally keep a NULL trigger**, matching `451_agent_task_comment_thread`'s convention that pre-migration rows drain without rewriting historical data. An earlier revision seeded them from the binding cursor; that was removed because it could not be proven correct (the cursor advances for channel commands, which the trigger deliberately excludes) and bought almost nothing (the trigger is written during `AppendUserMessage`, which commits before the flush that creates the delivery, so the first post-deploy turn already supplies a correct value).

A NULL trigger costs different things depending on where the answer is going, and the delivery query is written around that:

- **Ordinary group / p2p / Slack DM** — a chat-level send lands in the very conversation the session belongs to: no quote, no mention, right place.
- **Thread-isolated session (Slack channel thread, Telegram forum topic, Lark topic)** — the thread is recovered from the binding, which for these bindings is a stable property of the session rather than a cursor. Isolation is detected from the marker every adapter already writes: a composite `chat:thread` key carries the real chat id in its config, so a config chat id differing from `channel_chat_id` *is* the flag. The answer still reaches its thread.
- **Lark topic specifically** — a recovered thread is not enough, because Lark can only enter a topic by replying to a message inside it. The Patcher declines to send rather than posting into the parent group, on both the reply and error-card paths. Not a confidentiality break (group members can open the topic either way) but the wrong place, and it silently undoes the session isolation topic routing exists to provide. Deliberately narrower than `sendWithReplyFallback`'s chat-level retry, which fires only when Lark reports the topic itself cannot receive the reply.

When the generation *does* have a trigger, its thread wins unconditionally — including when that thread is legitimately empty, which is how a Slack DM answered at top level stays distinct from one answered inside a thread.

Reachable only for generations predating migration 460 that are recovered after deploy without a new inbound turn, since every turn after deploy records a trigger before its task is enqueued.

## Beyond Lark

The trigger snapshot lives in the shared channel engine, so this also fixes the same latent mis-attribution on other channels — Telegram's quote-reply and Slack's thread targeting had it too; only Lark's new ordinary-group reply made it visible. Only the Lark adapter populates `SenderChannelID` today; other channels leave it NULL, which reads as "no identity" and sends unmentioned.

## Testing

`go build ./...`, `go vet`, and `make test` against a migrated database. `internal/integrations/lark`, `internal/integrations/channel/engine` and `internal/service` pass.

Local-only failures, none of them from this diff: `cmd/multica`, `internal/cli`, `internal/daemon` and `internal/daemon/execenv` abort on a `.multica/daemon_task_context.json` marker present in the agent workspace this was developed in. `internal/handler` and `internal/service` pass. (An earlier round saw `TestCommentSourceContextLifecycle` fail here; it was verified identical at the merge base `9fab6da91` with none of these changes applied, and on a second database, so it was never from this diff.)

Coverage added:

- ordinary group replies to the trigger with `reply_in_thread=false`; topic replies unchanged; p2p and no-trigger-id stay chat-level (patcher and outcome replier both)
- mention rendered in the text path and in the markdown card path; the two spellings pinned apart; malformed ids degrade to no mention
- unbound / unknown sender and a failed lookup both send the body unchanged; p2p issues no mention
- **`TestChannelTaskDeliveryFreezesTriggerPerGeneration`** — DB-backed. Two generations are appended with *no* delivery created in between, then both enqueued, and each delivery keeps its own message and sender. It first asserts the session cursor holds the *other* generation's message, so the wrong answer is demonstrably reachable rather than merely hypothetical.
- `TestChannelTaskDeliveryIsImmutableAfterCreation` — the separate "already-created row does not move" case
- `TestPatcherMentionsEachTaskOwnSender`, `TestPatcherMentionsTriggeringAccountNotSiblingIdentity` — cross-mention and second-bound-account regressions
- `TestAppendUserMessage_ReplyTargetCarriesSender` / `...SenderNullWhenAbsent` — the triple is written to the generation this message belongs to, and NULL rather than empty when an adapter supplies no sender

Note for reviewers running the DB tests: they fall back to `postgres://…/multica` when `DATABASE_URL` is unset, and this PR adds a migration — an unmigrated local database will fail on the new columns.

**Not verified against a live tenant** — no Official Cloud access from here. Two things are worth a manual check on staging before production, because they decide whether the fix delivers the notification the issue asks for:

1. whether `<at>` inside a **schema-2.0 card** fires a real Feishu @ notification (most agent answers contain markdown, so the card is the common path — if cards don't notify, the mention needs to move to a text message);
2. that the reply endpoint behaves as expected in an ordinary group with `reply_in_thread=false`.

### DIFF
--- a/server/internal/integrations/channel/engine/session.go
+++ b/server/internal/integrations/channel/engine/session.go
@@ -74,6 +74,7 @@ type SessionQueries interface {
 	ResolveChannelChatContextHistoryStart(ctx context.Context, arg db.ResolveChannelChatContextHistoryStartParams) error
 	SetChannelChatContextInitiator(ctx context.Context, arg db.SetChannelChatContextInitiatorParams) (pgtype.UUID, error)
 	UpdateChannelChatSessionBindingReplyTarget(ctx context.Context, arg db.UpdateChannelChatSessionBindingReplyTargetParams) error
+	SetChannelChatContextReplyTarget(ctx context.Context, arg db.SetChannelChatContextReplyTargetParams) error
 	MarkChannelInboundDedupProcessed(ctx context.Context, arg db.MarkChannelInboundDedupProcessedParams) (int64, error)
 }
 
@@ -204,6 +205,9 @@ func (a dbSessionQueries) SetChannelChatContextInitiator(ctx context.Context, ar
 }
 func (a dbSessionQueries) UpdateChannelChatSessionBindingReplyTarget(ctx context.Context, arg db.UpdateChannelChatSessionBindingReplyTargetParams) error {
 	return a.q.UpdateChannelChatSessionBindingReplyTarget(ctx, arg)
+}
+func (a dbSessionQueries) SetChannelChatContextReplyTarget(ctx context.Context, arg db.SetChannelChatContextReplyTargetParams) error {
+	return a.q.SetChannelChatContextReplyTarget(ctx, arg)
 }
 func (a dbSessionQueries) MarkChannelInboundDedupProcessed(ctx context.Context, arg db.MarkChannelInboundDedupProcessedParams) (int64, error) {
 	return a.q.MarkChannelInboundDedupProcessed(ctx, arg)
@@ -549,9 +553,19 @@ func (s *ChatSession) StartSession(ctx context.Context, in StartSessionInput) (S
 	if in.MessageID != "" {
 		if err := qtx.UpdateChannelChatSessionBindingReplyTarget(ctx, db.UpdateChannelChatSessionBindingReplyTargetParams{
 			ReplyChatSessionID: session.ID, LastMessageID: textOrNull(in.MessageID), LastThreadID: textOrNull(in.ThreadID),
-			LastSenderID: textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return StartSessionResult{}, fmt.Errorf("set started chat reply target: %w", err)
+		}
+		// A started session opens at revision 1, and the message that started
+		// it is that generation's trigger.
+		if err := qtx.SetChannelChatContextReplyTarget(ctx, db.SetChannelChatContextReplyTargetParams{
+			ChatSessionID: session.ID,
+			Revision:      1,
+			LastMessageID: textOrNull(in.MessageID),
+			LastThreadID:  textOrNull(in.ThreadID),
+			LastSenderID:  textOrNull(in.SenderChannelID),
+		}); err != nil {
+			return StartSessionResult{}, fmt.Errorf("snapshot started chat reply target: %w", err)
 		}
 	}
 	dedupMessageID := in.DedupMessageID
@@ -713,6 +727,21 @@ func (s *ChatSession) AppendUserMessage(ctx context.Context, in AppendInput) (Ap
 		}); err != nil {
 			return AppendResult{}, fmt.Errorf("snapshot channel context initiator: %w", err)
 		}
+		// The trigger an outbound reply answers, snapshotted on the same
+		// generation and under the same command guard as the initiator above:
+		// a message excluded from agent input is not the question being
+		// answered, so it must not become the message the answer quotes.
+		if in.MessageID != "" {
+			if err := qtx.SetChannelChatContextReplyTarget(ctx, db.SetChannelChatContextReplyTargetParams{
+				ChatSessionID: in.SessionID,
+				Revision:      contextRevision,
+				LastMessageID: textOrNull(in.MessageID),
+				LastThreadID:  textOrNull(in.ThreadID),
+				LastSenderID:  textOrNull(in.SenderChannelID),
+			}); err != nil {
+				return AppendResult{}, fmt.Errorf("snapshot channel context reply target: %w", err)
+			}
+		}
 	}
 	// channel_ingested is the immutable provenance the cancel path gates on:
 	// it must be stamped in the same transaction as the message so no later
@@ -738,16 +767,15 @@ func (s *ChatSession) AppendUserMessage(ctx context.Context, in AppendInput) (Ap
 		return AppendResult{}, fmt.Errorf("touch chat session: %w", err)
 	}
 
-	// Record the latest trigger so the decoupled outbound patcher can thread
-	// its reply back into the originating topic and @-mention its sender. All
-	// three move together: the sender must describe the same message the reply
-	// targets, or a reply to one member could carry another member's mention.
+	// Advance the session's latest-trigger cursor, which the history-boundary
+	// bookkeeping reads. The outbound reply target is NOT taken from here — it
+	// is frozen per generation above, because this cursor belongs to the
+	// session and a debounced run answers one generation of it.
 	if in.MessageID != "" {
 		if err := qtx.UpdateChannelChatSessionBindingReplyTarget(ctx, db.UpdateChannelChatSessionBindingReplyTargetParams{
 			ReplyChatSessionID: in.SessionID,
 			LastMessageID:      textOrNull(in.MessageID),
 			LastThreadID:       textOrNull(in.ThreadID),
-			LastSenderID:       textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return AppendResult{}, fmt.Errorf("update reply target: %w", err)
 		}
--- a/server/internal/integrations/channel/engine/session.go
+++ b/server/internal/integrations/channel/engine/session.go
@@ -562,6 +562,7 @@ func (s *ChatSession) StartSession(ctx context.Context, in StartSessionInput) (S
 			ChatSessionID: session.ID,
 			Revision:      1,
 			LastMessageID: textOrNull(in.MessageID),
+			LastThreadID:  textOrNull(in.ThreadID),
 			LastSenderID:  textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return StartSessionResult{}, fmt.Errorf("snapshot started chat reply target: %w", err)
@@ -735,6 +736,7 @@ func (s *ChatSession) AppendUserMessage(ctx context.Context, in AppendInput) (Ap
 				ChatSessionID: in.SessionID,
 				Revision:      contextRevision,
 				LastMessageID: textOrNull(in.MessageID),
+				LastThreadID:  textOrNull(in.ThreadID),
 				LastSenderID:  textOrNull(in.SenderChannelID),
 			}); err != nil {
 				return AppendResult{}, fmt.Errorf("snapshot channel context reply target: %w", err)
--- a/server/internal/integrations/channel/engine/session.go
+++ b/server/internal/integrations/channel/engine/session.go
@@ -379,14 +379,20 @@ func (s *ChatSession) createSessionAndBinding(ctx context.Context, in EnsureSess
 // its own binding row, recording the real thread here per session does not clash
 // across sibling threads.
 type AppendInput struct {
-	SessionID           pgtype.UUID
-	Sender              pgtype.UUID
-	InstallationID      pgtype.UUID
-	Body                string
-	CommandText         string
-	MessageID           string
-	DedupMessageID      string
-	ThreadID            string
+	SessionID      pgtype.UUID
+	Sender         pgtype.UUID
+	InstallationID pgtype.UUID
+	Body           string
+	CommandText    string
+	MessageID      string
+	DedupMessageID string
+	ThreadID       string
+	// SenderChannelID is the platform-native id of whoever sent THIS message
+	// (Lark open_id, Slack user id, ...). Sender above is the Multica user it
+	// resolved to; both are recorded because the outbound side needs the
+	// platform id to render a native @-mention, and a Multica user can hold
+	// more than one platform identity on the same installation.
+	SenderChannelID     string
 	ClaimToken          pgtype.UUID
 	MediaPendingSeconds float64
 	ForceFresh          bool
@@ -407,10 +413,12 @@ type StartSessionInput struct {
 	// CommandText is the current member-authored instruction before adapter
 	// context enrichment. It is used only for the initial Chat title; Body
 	// remains the canonical persisted/agent-visible content.
-	CommandText            string
-	MessageID              string
-	DedupMessageID         string
-	ThreadID               string
+	CommandText    string
+	MessageID      string
+	DedupMessageID string
+	ThreadID       string
+	// SenderChannelID mirrors AppendInput.SenderChannelID.
+	SenderChannelID        string
 	ClaimToken             pgtype.UUID
 	MediaPendingSeconds    float64
 	PersistMessage         bool
@@ -541,6 +549,7 @@ func (s *ChatSession) StartSession(ctx context.Context, in StartSessionInput) (S
 	if in.MessageID != "" {
 		if err := qtx.UpdateChannelChatSessionBindingReplyTarget(ctx, db.UpdateChannelChatSessionBindingReplyTargetParams{
 			ReplyChatSessionID: session.ID, LastMessageID: textOrNull(in.MessageID), LastThreadID: textOrNull(in.ThreadID),
+			LastSenderID: textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return StartSessionResult{}, fmt.Errorf("set started chat reply target: %w", err)
 		}
@@ -730,12 +739,15 @@ func (s *ChatSession) AppendUserMessage(ctx context.Context, in AppendInput) (Ap
 	}
 
 	// Record the latest trigger so the decoupled outbound patcher can thread
-	// its reply back into the originating topic.
+	// its reply back into the originating topic and @-mention its sender. All
+	// three move together: the sender must describe the same message the reply
+	// targets, or a reply to one member could carry another member's mention.
 	if in.MessageID != "" {
 		if err := qtx.UpdateChannelChatSessionBindingReplyTarget(ctx, db.UpdateChannelChatSessionBindingReplyTargetParams{
 			ReplyChatSessionID: in.SessionID,
 			LastMessageID:      textOrNull(in.MessageID),
 			LastThreadID:       textOrNull(in.ThreadID),
+			LastSenderID:       textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return AppendResult{}, fmt.Errorf("update reply target: %w", err)
 		}
--- a/server/internal/integrations/channel/engine/session.go
+++ b/server/internal/integrations/channel/engine/session.go
@@ -562,7 +562,6 @@ func (s *ChatSession) StartSession(ctx context.Context, in StartSessionInput) (S
 			ChatSessionID: session.ID,
 			Revision:      1,
 			LastMessageID: textOrNull(in.MessageID),
-			LastThreadID:  textOrNull(in.ThreadID),
 			LastSenderID:  textOrNull(in.SenderChannelID),
 		}); err != nil {
 			return StartSessionResult{}, fmt.Errorf("snapshot started chat reply target: %w", err)
@@ -736,7 +735,6 @@ func (s *ChatSession) AppendUserMessage(ctx context.Context, in AppendInput) (Ap
 				ChatSessionID: in.SessionID,
 				Revision:      contextRevision,
 				LastMessageID: textOrNull(in.MessageID),
-				LastThreadID:  textOrNull(in.ThreadID),
 				LastSenderID:  textOrNull(in.SenderChannelID),
 			}); err != nil {
 				return AppendResult{}, fmt.Errorf("snapshot channel context reply target: %w", err)
--- a/server/internal/integrations/channel/engine/session_db_test.go
+++ b/server/internal/integrations/channel/engine/session_db_test.go
@@ -1346,6 +1346,126 @@ func TestChannelTaskDeliveryFreezesThreadPerGeneration(t *testing.T) {
 	}
 }
 
+// TestChannelTaskDeliveryRecoversThreadForIsolatedBinding covers the shape
+// migration 460 leaves behind on every thread-isolated channel: a generation
+// with no trigger at all, recovered after deploy.
+//
+// The trigger is legitimately unknown there. The THREAD is not: a Slack
+// channel thread, Telegram forum topic or Lark topic keeps one binding per
+// thread, so its last_thread_id is a stable property of the session rather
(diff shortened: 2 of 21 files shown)


Bohan-J:
Pushed `0117ff1`, which addresses the identity blocker.

## The blocker was real

Confirmed the reachability claim: `CreateLarkUserBinding`'s `ON CONFLICT (installation_id, channel_user_id)` only fires when the *same* platform account is re-bound. A member binding a *second* Feishu account on the same installation is a plain INSERT (`binding_token.go:178`), so `(installation_id, multica_user_id)` genuinely holds two rows. `GetChannelUserBindingByMember` was a `:one` with no ordering and nothing to make it deterministic, so it could name either account — "reply to B, @ A". Normal binding flow, not a corrupt DB.

## What changed

The sender is per-trigger data, exactly like the message id it accompanies, so it now travels the same path rather than being re-derived:

- `channel_chat_session_binding.last_sender_id` and `channel_task_delivery.channel_sender_id` (migration 457, additive, nullable, no backfill, no index — both tables are only ever read by their existing key).
- `UpdateChannelChatSessionBindingReplyTarget` writes the sender in the **same statement** as `last_message_id` / `last_thread_id`. That co-write is the actual guarantee: the sender can never describe a different message than the one the reply targets.
- `CreateChannelTaskDeliveryFromSession` (and `CopyChannelTaskDelivery`, for child tasks) freeze it per task.
- The Patcher reads the snapshot. `GetChannelUserBindingByMember` and its store wrapper are deleted.

The Multica member is no longer an input to the mention at all, so the sibling-identity case is now unrepresentable rather than merely unlikely — which is why I took the snapshot rather than adding a reverse uniqueness constraint. As you noted, a "one channel identity per member per installation" rule is a separate product decision with its own migration and existing-duplicate story; nothing here needs it.

`SenderChannelID` is populated only by the Lark adapter today. Other channels leave it NULL, which reads as "no identity" and sends unmentioned — the same degradation as pre-migration rows.

## On Must-fix 2

You were right that the old `map[member]openID` fake could not express the state. It is gone; the fake now carries per-task delivery snapshots. Three tests cover the gap:

- `TestPatcherMentionsTriggeringAccountNotSiblingIdentity` — the task's initiator is a member whose *other* open_id is the one the fixture would have found; the reply must mention the account that sent the trigger.
- `TestChannelTaskDeliveryFreezesTriggerSender` — DB-backed, against real Postgres: member A's delivery row keeps A's sender and message id after member B messages the same group, and B's own run freezes B's.
- `TestAppendUserMessage_ReplyTargetCarriesSender` / `...SenderNullWhenAbsent` — the sender rides along in the reply-target write, and is NULL rather than empty when an adapter does not supply one.

## Verification

`go build ./...`, `go vet`, and `make test` against a migrated worktree DB. `internal/integrations/lark`, `internal/integrations/channel/engine`, `internal/service` and `internal/handler` all pass.

On the shared-test-DB failures you hit: those are a stale local database, not the diff. The engine DB tests fall back to `postgres://…/multica` when `DATABASE_URL` is unset; pointing it at a migrated database makes them pass. Worth knowing since this PR adds migration 457 — an unmigrated local DB will now fail on `last_sender_id` too.

Still failing here, unchanged and unrelated: `cmd/multica`, `internal/cli`, `internal/daemon`, `internal/daemon/execenv` all abort because the tests detect a `.multica/daemon_task_context.json` marker in the agent workspace I run in. Not present in CI.

Real-tenant verification is still outstanding and I still can't do it from here — the two open questions from the PR description (does `<at>` inside a schema-2.0 card actually fire an @ notification, and how a non-thread reply renders in an ordinary group) are unchanged by this commit.

Bohan-J:
Pushed `d179885`: the seed `UPDATE` is gone, so migration 457 is now purely additive DDL.

**On the two reasons given — I agree with the conclusion, but for the second reason, not the first.**

The provenance objection is correct and it is my bug. `AppendUserMessage` advances the binding cursor unconditionally while the generation trigger is written only when `cmd == nil`; the seed copied the former into the latter, so a session whose last pre-deploy message was `/issue` would have had that command seeded as a generation's reply target — precisely the value the guard exists to exclude. I wrote both halves and did not notice they disagreed.

The deployment-risk framing I think overstates this particular statement. `channel_chat_context_generation` holds one row per (channel-bound chat session, revision), the `UPDATE` touches at most one row per session and is filtered on `last_message_id IS NULL`. That is a different order of magnitude from the historical task-queue rewrite `451_agent_task_comment_thread` declined, and I do not think this one warranted a lock budget.

**But the seed should go regardless, for a reason neither of us led with: it buys almost nothing.** The trigger is recorded during `AppendUserMessage`, which commits well before the debounced flush that enqueues the task and creates its delivery row — so the first inbound turn after deploy already supplies a correct value for that generation, without any backfill. The only rows a seed could serve are generations enqueued with *no* post-deploy append, i.e. the recovered-run path. And for exactly those the binding cursor has already advanced past the generation, so the seed would hand them a confidently wrong message and sender instead of no answer. The set it helps and the set where it is wrong are nearly the same set.

That is also why I did not take the suggested alternative (a provable agent-visible-trigger source, a batching/resume hook, a legacy-fixture migration test). Building that machinery to recover a value that recreates itself on the session's next message would be effort spent on the wrong thing. Deleting is the right size of fix.

Leaving these rows NULL degrades to a chat-level send with no mention, which for the recovered-run case is strictly better than the pre-change behaviour of quoting the session's newest trigger — some other member's message. It also matches 451's convention directly.

PR description updated with the removal and the reasoning.

## A correction to my own earlier verification claims

I reported `internal/handler` as passing in the last two rounds. That was a cached `ok` — I had not actually exercised it. Running it fresh, `TestCommentSourceContextLifecycle` fails on this machine.

It is not from this PR: it fails identically at the merge base `9fab6da91` with none of my changes applied, and on a second, separate database. CI's `backend-tests` passes on this branch. So the conclusion is unchanged, but the earlier claim was weaker than I stated it.

Locally still failing, all pre-existing or environmental: `cmd/multica`, `internal/cli`, `internal/daemon`, `internal/daemon/execenv` (they abort on the `.multica/daemon_task_context.json` marker in the agent workspace) plus the handler test above. `internal/integrations/lark`, `internal/integrations/channel/engine` and `internal/service` pass.

Bohan-J:
Pushed `73de0fe`. The topic-leak finding is correct and I've taken it — with a different fix than either option suggested.

## The finding holds, and it is broader than Lark

Verified the sequence. What it exposed is that moving the whole message/thread/sender triple onto the generation conflated two different things:

- **Trigger** — which message an answer quotes, which account it mentions — is genuinely per-generation.
- **Route** — which chat or topic the answer goes *to* — is per-binding. A thread-isolated session has one binding per topic, so `last_thread_id` names that topic for every generation of it.

Sourcing the route from the generation left pre-migration generations with no thread. And this is not Lark-only: `slack/outbound.go:218` resolves the real channel from config and takes `thread_ts` from `LastThreadID`, so a recovered run posts to the parent channel instead of its thread. Telegram is the same shape. My change introduced that for all three; I should have caught it when I moved the column.

## The fix: put the route back where it belongs

`channel_thread_id` goes back to being sourced from the binding — exactly where it was before this PR — and the generation carries only `last_message_id` and `last_sender_id`. **Slack and Telegram are fixed by that alone, with no adapter change**, because both can place a message in a thread from the thread id by itself.

Lark cannot: its only route into a topic is replying to a message inside it. So for the one remaining combination — topic route, no trigger message — the Patcher now declines to send, on both the reply and the error-card paths. Silence is recoverable (the member asks again, and the answer is still in Multica); a topic conversation surfacing in front of the whole group is not.

That guard is deliberately narrower than `sendWithReplyFallback`'s chat-level retry. That path fires only when Lark reports the topic itself cannot receive the reply — recalled, gone, topics disabled — where the topic is unusable and delivering beats losing the reply. Here the topic is fine and only our own bookkeeping is missing, so falling back would be a self-inflicted leak rather than an unavoidable one.

## Where I landed differently

I did not add a fail-closed branch for Slack/Telegram, because with the route restored they have nothing to fail closed *about* — they route correctly. Fail-closed is only right for Lark, where the platform genuinely offers no way to reach the topic. Defining a shared cross-channel fail-closed policy would have been solving a problem the route/trigger split removes.

I also want to be accurate about reachability rather than lean on it: this needs a generation predating migration 457, recovered after deploy with no new inbound turn. Every post-deploy turn records a trigger before its task is enqueued. That is a narrow, one-time window — but "narrow" is not an argument for leaking, which is why the guard is in rather than a note in the description.

## Tests

- `TestChannelTaskDeliveryKeepsRouteWhenTriggerIsMissing` (DB-backed) — a thread-isolated binding with a trigger-less generation: the delivery keeps the thread and both trigger columns stay NULL. This is the Slack/Telegram regression too, at the shared layer.
- `TestPatcherSkipsTopicReplyWithoutTrigger` / `TestPatcherSkipsTopicErrorCardWithoutTrigger` — a composite `chat:thread` binding with no trigger sends *nothing*, on both paths.
- `TestPatcherStillRepliesInOrdinaryGroupWithoutTrigger` — the boundary: an ordinary group still replies, unquoted and unmentioned, because a chat-level send lands in the chat the session belongs to.

## PR description

Corrected on both counts — it now describes the route/trigger split, and the local-verification section no longer claims `internal/handler` passes. You were right to flag that; the accurate statement is that it fails locally, identically at the merge base with none of my changes, while CI's `backend-tests` is green.

One addition to that list: a source-context cleanup test in `internal/service` is flaky against the shared dev database — it failed in a full-package run and passes in isolation and on re-run. Also unrelated to this diff, but I would rather name it than have it show up as a surprise.

Bohan-J:
Pushed `94306b4`. Both must-fixes taken; the first one was a live regression I introduced, not an edge case.

## Must-fix 1 — Slack DM

Verified against `slackSessionRouting`: a DM returns `bindingKey = chatID` (**not** composite) with a per-message `replyThread`, so one binding spans every thread in that DM. My "the thread is route data, read it from the binding" generalization is simply false there, and the consequence is ordinary traffic — A asks inside T1, a `/clear` within the 3s window opens revision 2 at top level, and revision 1's answer lands wherever revision 2 spoke.

What I got wrong is worth naming precisely: for thread-*isolated* sessions binding and trigger coincide, which made the thread look like route data. It generalizes to exactly one of the two shapes. So `last_thread_id` goes back onto the generation with the message and sender — the three are one trigger and travel together — and the route is what genuinely does not vary by generation: chat, type, config, revision.

Lark's fail-closed guard is re-keyed accordingly. It now identifies a topic-isolated binding by its **config**, which is what actually marks one (`larkSessionRouting` writes a config only for topic sessions), instead of by a thread id that is now trigger data and legitimately absent for pre-migration generations. Without that change the guard would have silently stopped firing.

## The test you flagged

You were right that it proved nothing: it ran on the `p2p`, non-composite fixture with a thread inserted by hand, so it asserted "binding thread survives" on precisely the shape where the binding thread must *not* be used. It would have locked in the wrong generalization.

Replaced rather than repaired, with the DM shape as the subject: T1 in revision 1, top level in revision 2, both appended before either delivery is created, each keeping its own thread — and the binding cursor asserted to hold the other one first, so the wrong answer is demonstrably reachable.

## Must-fix 2 — migration prefix

Renumbered `457` → `460`; `main` has since taken 457, 458 and 459. up/down renamed as a pair, and `git merge-tree` against current `origin/main` reports no conflicts.

## Documentation

Corrected the drift you found. The description now states the trigger as message + thread + sender on the generation, and — since this is the part that is easy to get backwards, and I did get it backwards — explains *why* the thread belongs with the trigger rather than the route, with the Slack DM counterexample. The migration section is rewritten for 460 and no longer claims NULL uniformly degrades to a chat-level send; it now lists the per-destination behaviour including the Lark topic case.

One correction to my own framing from the previous round: I described the Lark topic case as a leak "in front of the whole group". That overstated it — Feishu topics are visible to group members either way, so it is misplacement and a silent loss of session isolation, not a confidentiality breach. The guard still stands, but on the accurate justification, and the code comment says so.

## Verification

`go build ./...`, `go vet`, full `make test` against a migrated database. `internal/integrations/lark`, `internal/integrations/channel/engine`, `internal/service` and `internal/handler` all pass this run.

Still failing locally, unrelated and pre-existing: `cmd/multica`, `internal/cli`, `internal/daemon`, `internal/daemon/execenv` abort on the `.multica/daemon_task_context.json` marker in this agent workspace.

The two real-tenant checks — card `<at>` notification, ordinary-group `reply_in_thread=false` — remain unverified from here and stay staging gates.

Bohan-J:
Pushed `ecec60a`. The finding is right — putting the thread back on the generation fixed Slack DMs and reopened the legacy hole on exactly the channels that route by thread.

## The fix, and why it is one change rather than three

`CreateChannelTaskDeliveryFromSession` now expresses **both** semantics instead of one:

```sql
CASE
    WHEN generation.last_message_id IS NOT NULL THEN generation.last_thread_id
    WHEN COALESCE(binding.config ->> 'chat_id', binding.config ->> 'channel_id', '')
         NOT IN ('', binding.channel_chat_id) THEN binding.last_thread_id
    ELSE NULL
END
```

- A generation that recorded a trigger keeps its own thread, **including when that thread is legitimately empty** — which is what keeps a Slack DM answered at top level distinct from one answered inside T1 or T2.
- Only a generation with no trigger at all falls back, and only for a thread-isolated binding.

Isolation is read from the marker every adapter already writes rather than inferred: a composite `chat:thread` key carries the real chat id in its config precisely *because* the key is composite, so a config chat id differing from `channel_chat_id` **is** the flag (`larkSessionRouting`, `slackSessionRouting`, `telegramSessionRouting`). A DM or plain chat stores a config id equal to its key, so it recovers nothing — correct, since its cursor only ever names whoever spoke last.

I did this in the shared query rather than adding fallbacks to the Slack and Telegram adapters. The delivery row is where route and trigger are frozen, so one change covers both channels and any future thread-routed one, with no adapter churn. `historyTarget` is a good reference for the *idea* — it recovers a thread root from the composite key on the read path — but replicating its key parsing per adapter would put the same rule in three places, and this version needs no key parsing at all since the isolated binding's `last_thread_id` already holds the root.

Lark still declines to send in this case regardless: a recovered thread is not enough when the only way into a topic is replying to a message inside it.

## Tests

- `TestChannelTaskDeliveryRecoversThreadForIsolatedBinding` (DB-backed, table-driven) — a pre-460 generation with no trigger. The isolated binding still delivers into `thread_T1`; the DM/plain binding recovers nothing; neither smuggles in a trigger message or sender.
- `TestChannelTaskDeliveryFreezesThreadPerGeneration` kept unchanged — the DM cross-generation case from last round, which is exactly what the first `WHEN` branch protects.

On asserting the Slack and Telegram *send* targets specifically: both adapters read `delivery.channel_thread_id` straight through (`slackBindingFromTaskDelivery` → `outboundTarget`, `telegramBindingFromTaskDelivery` → `outboundTarget`), so that column is the canonical layer for this guarantee and where the fix lives. Adapter-level tests would re-assert a pass-through rather than exercise the fix, so I asserted at the delivery layer and named the consumers in the test comment instead. Happy to add the thin adapter tests if you would rather have them pinned there too.

## Documentation

Both corrections made. The migration's closing note no longer says NULL degrades uniformly to a chat-level send — it now lists the three destination cases, matching the query and the Lark guard. The PR description's NULL section is rewritten the same way, and the local-verification paragraph now reflects this round: `internal/handler` and `internal/service` pass, with a note that the earlier `TestCommentSourceContextLifecycle` failure was verified identical at the merge base and on a second database, so it was never from this diff.

## Verification

`go build ./...`, `go vet`, full `make test` against a migrated database. `lark`, `channel/engine`, `slack`, `telegram`, `service` and `handler` all pass; the DB-backed regressions above were also run individually.

Still failing locally, unrelated: `cmd/multica`, `internal/cli`, `internal/daemon`, `internal/daemon/execenv` abort on the `.multica/daemon_task_context.json` marker in this agent workspace.

Card `<at>` notification and ordinary-group `reply_in_thread=false` remain staging gates.

Issue #6857, opened 2026-08-12
[Feature]: Preserve Codex-native AGENTS.md, Hooks, and user extensions in task-scoped CODEX_HOME


n374:
## Deployment type

self-host

Self-hosted Multica daemon with Codex agents. This may also affect any deployment that creates an isolated, task-scoped `CODEX_HOME`.

---

### English

Multica creates a task-scoped `CODEX_HOME` for Codex isolation. On current `main`, `server/internal/daemon/execenv/codex_home.go` inherits only a selected set of files from the shared Codex home, including:

```go
var codexCopiedFiles = []string{
    "config.json",
    "config.toml",
    "instructions.md",
}
```

There is also separate handling for `auth.json`, user Skills, plugin cache exposure, and selected files referenced from `config.toml`. This helps preserve task isolation, but it can silently remove Codex-native user-level configuration and extension points.

A concrete example is global `AGENTS.md`. Codex officially supports user-level instructions from:

```text
$CODEX_HOME/AGENTS.override.md
$CODEX_HOME/AGENTS.md
```

Codex uses `AGENTS.override.md` before `AGENTS.md` at the global level, then combines the selected global instructions with project-level `AGENTS.md` files discovered from the project root down to the current working directory. Project-level discovery also prefers `AGENTS.override.md` before `AGENTS.md` in each directory.

When Multica replaces `CODEX_HOME` with a task-scoped directory but does not preserve these entries, global Codex instructions that work in normal Codex sessions disappear inside Multica.

Hooks show the same class of compatibility issue. Codex supports user-level hooks through `hooks.json` and inline `[hooks]` configuration in `config.toml`. Hook behavior depends not only on the hook config file, but also on referenced scripts, relative paths, executable permissions, the hook source path, and trust state.

The related Multica history points to a recurring pattern:

- #3112 attempted to sync Codex hook files into task homes, but was closed without merge.
- #5174 is currently open and addresses inheriting user Codex hooks plus hook trust-path remapping.
- #1922 reported missing user Skills; #2519 later seeded user-installed Codex Skills into per-task `CODEX_HOME`.
- #6271 reported missing files referenced by Codex config; #6291 added provisioning for selected config-referenced instruction files.
- #6742 changed user Codex Skills inheritance to linking instead of copying.
- #6233 restored the real Unix `HOME`, while leaving task-scoped `CODEX_HOME` in place.

[...]

PR #8273, opened 2026-09-10
MUL-7261: fix(daemon): inherit Codex native instructions and agent roles

matthyx:
## What does this PR do?

Codex tasks currently inherit the user's configuration and plugin skills but lose global instruction files and native agent definitions when Multica creates an isolated `CODEX_HOME`. A workflow can therefore require an installed reviewer role that the task cannot discover or launch.

Copy `AGENTS.md`, `AGENTS.override.md`, and direct `agents/*.toml` definitions into the task home on preparation and reuse. Preserve file contents and Codex's instruction precedence while refreshing host-owned copies and removing stale inherited definitions.

## Related Issue

Refs #6857. This addresses the global-instructions/native-role subset; hooks and arbitrary referenced assets remain follow-up work.

## Type of Change

- [x] Bug fix
- [x] Documentation update
- [x] Tests

## Changes Made

- Add native-configuration synchronization before other Codex-home preparation steps. Source-read or manifest errors stop preparation instead of silently dropping configuration.
- Track inherited role filenames so deleted host definitions are removed on reuse while unrelated task role files remain. Matching host definitions and both global instruction filenames are authoritative.
- Use a verified `os.Root`, atomic replacement, and owner-only file permissions. A task cannot redirect the new writes or stale deletions outside its home through destination symlinks; task edits do not mutate shared files.
- Keep Multica's existing native multi-agent policy and daemon opt-in unchanged.
- Document inheritance, refresh behavior, and scope limits in the CLI/daemon guide and built-in agent reference.

## How to Test

From `server/`:

```sh
bash ../scripts/go-test-with-agent-cli-guard.sh go test -race ./internal/daemon/execenv -count=1
go vet ./internal/daemon/execenv
go build ./internal/daemon/execenv
```

The new preparation regression failed before the fix because `AGENTS.md` was missing. Coverage includes refresh and removal, unrelated task roles, linked sources, destination symlinks/hardlinks, owner-only permissions, invalid manifests, interrupted-refresh retry, and unchanged default/opt-in policy. The full package race suite passes after the change. Default tests use temporary fixtures and the agent CLI guard, without launching authenticated agents.

The existing symlinked-home regression now expects the earlier native-configuration guard; its assertion that external files remain unchanged is preserved.

## Scope and risks

Only direct native role TOMLs are copied. Files referenced by those definitions, custom `agents.<name>.config_file` locations outside that directory, hooks, and hook trust state are not provisioned by this change. It does not change subagent lifecycle tracking or automatically enable multi-agent execution. Invalid native configuration sources now fail preparation visibly. Unix mode assertions are skipped on Windows; Windows ACL behavior was not exercised locally.

## Checklist

- [x] Included the project context and reasoning for the change above
- [x] Ran local tests and added regression coverage
- [x] Updated relevant documentation
- [x] Documented scope and risks
- [ ] Address upstream reviewer feedback before requesting merge

## AI Disclosure

**AI tool used:** Codex with oh-my-codex.

**Approach:** Reproduced the missing-inheritance behavior, implemented a focused fix using existing filesystem-boundary utilities, added regression tests, and ran cleanup plus independent code and architecture reviews. The PR is intentionally a draft for maintainer feedback on the inheritance contract.

### DIFF
--- a/CLI_AND_DAEMON.md
+++ b/CLI_AND_DAEMON.md
@@ -240,6 +240,29 @@ You need at least one installed. The daemon registers each detected CLI as an av
 4. Heartbeats are sent periodically (default: 15s) so the server knows the daemon is alive
 5. On shutdown, all runtimes are deregistered
 
+### Codex configuration inheritance
+
+Each task has an isolated `CODEX_HOME`. The daemon copies global `AGENTS.md`,
+`AGENTS.override.md`, and direct `agents/*.toml` native role definitions from its
+own `CODEX_HOME` (or `~/.codex` when unset). Contents are copied unchanged; Codex
+continues to decide instruction precedence, including `AGENTS.override.md`.
+Task edits to these copies do not modify the host files.
+
+On prepare and reuse, the shared source is authoritative: instruction files and
+matching native roles are refreshed, and missing shared instruction files remove
+their task copies. A task-local `.multica-inherited-agents.json` manifest tracks
+copied roles so removed host roles are deleted while unrelated task role files
+remain. Source read errors and invalid manifests fail preparation rather than
+silently continuing with incomplete inheritance.
+
+This copies native role definitions only; it does not recursively copy assets
+referenced by those definitions or install hooks. Native subagent execution still
+uses the existing daemon opt-in, `MULTICA_CODEX_MULTI_AGENT=1`, which stops Multica
+from forcing `features.multi_agent=false` and preserves the user's Codex setting.
+Set it in the daemon environment before starting a new task. The parent agent
+must await its children before finishing because task completion follows the
+parent's lifecycle.
+
 ### Configuration
 
 Daemon behavior is configured via flags or environment variables:
--- a/server/internal/daemon/execenv/codex_home.go
+++ b/server/internal/daemon/execenv/codex_home.go
@@ -1027,9 +1027,10 @@ func materialiseInCodexHome(codexHome, relPath, src, key string) error {
 	mode := os.FileMode(0o644)
 	if info, err := root.Lstat(relPath); err == nil {
 		// A reference may alias a private native instruction or role copy.
-		// Refresh regular files without broadening their existing permissions.
+		// Preserve restrictions for group/other, but restore owner read/write
+		// access if a previous task left an unreadable copy.
 		if info.Mode().IsRegular() {
-			mode &= info.Mode().Perm()
+			mode = (mode & info.Mode().Perm()) | 0o600
 		}
 		if err := root.Remove(relPath); err != nil {
 			return fmt.Errorf("remove stale %s copy %s: %w", key, relPath, err)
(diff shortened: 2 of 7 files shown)


multica-eve (changes requested):
🔄 Request changes：产品方向应该接受，但当前实现有两个可复现的阻塞性回归；在复用 `$CODEX_HOME` 时，Codex 可能丢失角色定义，或无法读取继承的 instructions。

1. **大小写重命名可能删除刚写入的角色文件**

   在 macOS/Windows 的大小写不敏感文件系统上，如果宿主角色从 `Reviewer.toml` 重命名为 `reviewer.toml`，当前代码会先写入新名称，再按旧名称清理 stale role。`server/internal/daemon/execenv/codex_native_config.go:118` 写入当前角色，而 `:124` 使用大小写敏感的 `slices.Contains` 判断旧角色；最终旧名称的删除会命中同一个文件系统对象，把刚写入的角色删除。我已在 macOS 复现。

   请改为先清理 stale entries 再写当前角色，或采用能识别文件系统同一性的比较方式，并补充 macOS/Windows 回归测试。

2. **旧目标文件的不可读权限会持续污染新副本**

   `server/internal/daemon/execenv/codex_home.go:1027-1033` 用 `0644 & old destination permissions` 计算新 mode。若上一次任务留下的相对 `model_instructions_file` 是 `0000`（或只有执行位），后续复用虽然会刷新内容，却仍把文件创建成不可读，导致 Codex 在读取 instructions 时失败。我已用 `0000` 目标文件复现。

   daemon 生成的文件应至少保持 owner-readable；对于与 native 配置重叠的路径，可使用受控的 `0600`。请补充 `0000`/仅执行权限的复用回归测试。

产品判断是 ✅ 应接受这项能力；修复以上两项后可以重新 review。现有 execenv race 测试和 `ByteIdentical` 缓存前缀测试均通过，但没有覆盖这两个 OS/权限边界。

multica-eve (approved):
✅ Approve: commit `2ed6421` resolves both previously reported blockers, and I found no new merge-blocking issues.

The product verdict remains **accept**: Multica-managed Codex tasks should inherit the user's native instructions and role definitions so daemon tasks behave consistently with locally launched Codex, without requiring user migration or changing the existing multi-agent opt-in.

The follow-up fixes are effective:

- Case-only role renames now remove stale names before publishing current roles, so a case-insensitive filesystem cannot delete the newly written role. I verified the regression on macOS, including repeated reuse.
- Refreshed config-referenced files now restore owner read/write access while preserving group/other restrictions, so a previous `0000` or execute-only destination cannot poison later tasks.

Validation completed on head `2ed6421d1b55376f961fd730968c17abd368e699`:

- Targeted regressions under `-race`: pass
- Full `server/internal/daemon/execenv` suite under `-race`: pass
- `ByteIdentical` prompt-prefix invariant: pass
- `go vet ./internal/daemon/execenv`: pass
- `git diff --check`: pass

If this behavior regresses, affected users will see missing or unreadable inherited instructions/roles when Codex starts; rollback is the PR revert plus invalidating affected reusable task-home copies.

PR #8275, opened 2026-09-10
MUL-7262: BHI-16123: refuse duplicate workspace daemons

bhirstmedia:
Closes BHI-16123

## What changed
- Adds a daemon start guard that scans known local profiles and refuses to start a second live daemon for the same workspace_id.
- Applies the guard to both background and foreground daemon starts.
- Adds regression coverage for same-workspace profile co-tenancy, including default-profile starts.

## Verification
- `go test ./cmd/multica -run 'TestRequireNoSameWorkspaceDaemon|TestDaemonLifecycleRefusesForeignDaemon|TestDaemonStatusReportsPortConflictInsteadOfClaimingItsOwn' -count=1 -v`
- `go build -buildvcs=false -ldflags "-X main.version=$VERSION -X main.commit=$COMMIT -X main.date=$DATE" -o bin/multica ./cmd/multica`
- `strings ./bin/multica | /usr/bin/grep -c requireNoSameWorkspaceDaemon` returned `1`

## Notes
- Full `go test ./cmd/multica -count=1` currently fails under this daemon-managed task shell because command tests detect the active task environment as agent execution context. The new daemon regression tests pass in the targeted run.

### DIFF
--- a/server/cmd/multica/cmd_daemon.go
+++ b/server/cmd/multica/cmd_daemon.go
@@ -375,6 +375,90 @@ func daemonIdentityMismatch(health map[string]any, profile string, port int) err
 	return &daemonProfileMismatchError{Want: profile, Got: got, Port: port}
 }
 
+type sameWorkspaceDaemonConflictError struct {
+	WorkspaceID string
+	WantProfile string
+	GotProfile  string
+	Port        int
+	PID         int
+}
+
+func (e *sameWorkspaceDaemonConflictError) Error() string {
+	pid := "unknown"
+	if e.PID > 0 {
+		pid = strconv.Itoa(e.PID)
+	}
+	return fmt.Sprintf(
+		"refusing to start %s: %s already has a live daemon for workspace_id %s on port %d (pid %s)",
+		describeProfile(e.WantProfile), describeProfile(e.GotProfile), e.WorkspaceID, e.Port, pid)
+}
+
+func healthIncludesWorkspace(health map[string]any, workspaceID string) bool {
+	workspaces, ok := health["workspaces"].([]any)
+	if !ok {
+		return false
+	}
+	for _, rawWorkspace := range workspaces {
+		workspace, ok := rawWorkspace.(map[string]any)
+		if !ok {
+			continue
+		}
+		if id, _ := workspace["id"].(string); id == workspaceID {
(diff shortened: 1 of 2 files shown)


multica-eve (changes requested):
Request changes: the product problem is real, but this one-time health scan does not establish the single-owner invariant and can both miss real conflicts and reject valid profile combinations.

The concrete failure is stronger than duplicate polling. `daemon_id` is machine-scoped across profiles, and built-in runtimes are upserted by `(workspace_id, daemon_id, provider)`. When a second process registers the same runtime, its initial sync calls `RecoverOrphans`, which marks every `dispatched`, `running`, or `waiting_local_directory` task on that shared runtime as failed. It can therefore interrupt work that the first process is still executing. The two processes also have independent machine-level concurrency semaphores and share heartbeat/deregister state, so the runtime model requires one process owner.

The current guard has four blocking correctness problems:

1. It checks only `fileCfg.WorkspaceID`, while the daemon calls `ListWorkspaces` and registers every workspace the authenticated user belongs to. An overlap on any non-default workspace is not detected.
2. The check-then-start sequence is not atomic. Two concurrent starts can both observe no conflict and proceed. A daemon in the `starting` state also serves health before initial workspace sync, so its workspace list can be empty during the exact race window.
3. The comparison is not scoped by server URL. Profiles intentionally isolate backend/account state; the same workspace UUID on different deployments must not conflict.
4. Each known profile is probed sequentially with an independent two-second timeout, so startup latency grows linearly with the number of profiles in the worst case.

Please enforce ownership with a crash-safe cross-process lock in the machine-level state directory, keyed by canonical `(server URL, workspace_id)`. After authentication and `ListWorkspaces`, but before any register or `RecoverOrphans` side effect, acquire the complete lock set in stable order; on any conflict, release locks acquired by that attempt and refuse startup. Hold the handles for the process lifetime, and acquire/release locks before runtime registration/removal when workspace membership changes. Health probing can remain as a compatibility or diagnostic aid, but it cannot be the correctness mechanism.

Regression coverage should include simultaneous starts, overlap on a non-default workspace, an existing daemon still reporting `starting`, identical UUIDs on different servers, crash-based lock release, and a workspace added while both daemons are already running.

Issue #6145, opened 2026-07-29
make selfhost health check uses wrong port variable ($PORT instead of $BACKEND_PORT)


mingxin:
## Summary

The `selfhost` Makefile target's health check uses `$PORT` (container internal port 8080) instead of `$BACKEND_PORT` (host-mapped port). When a user customizes `BACKEND_PORT` in `.env`, the health check silently fails even though the service is running correctly on the custom port.

## Reproduction

1. Set `BACKEND_PORT=42817` in `.env`
2. Run `make selfhost`
3. Services start correctly, but the health check loops for 60 seconds (hitting `localhost:8080` which has nothing listening) and never reports success

## Root cause

`Makefile:112`:

```makefile
@if curl -sf http://localhost:$${PORT:-8080}/health > /dev/null 2>&1; then
```

`$PORT` is the container's internal port (always 8080). The host-mapped port is `$BACKEND_PORT`. The same issue is on line 121 where the success message prints the backend URL.

## Suggested fix

```makefile
# Line 112: use BACKEND_PORT
@if curl -sf http://localhost:$${BACKEND_PORT:-8080}/health > /dev/null 2>&1; then

# Line 121: same fix
echo "  Backend:  http://localhost:$${BACKEND_PORT:-8080}";
```

## Environment

- Multica v0.4.12, macOS, Docker self-host

multica-eve:
Fixed by #6168, which is now being merged.

PR #8277, opened 2026-09-10
MUL-7263: feat(selfhost): refuse to start when a host port is already in use

luizvidoto:
## What does this PR do?

`make selfhost` had no defence against a host port it publishes already being in use.

Today the first sign of a conflict is Docker's own `Bind for 127.0.0.1:3000 failed: port is already allocated` — raised only *after* `docker compose pull` has downloaded several hundred megabytes of images. And when `up -d` fails only partially, `scripts/selfhost-wait.sh:71-75` still ends on `Services are still starting. Check logs` and **exits 0**, so a port conflict can read as a successful install.

Manually working around it did not stick either. `make selfhost PORT=9090 FRONTEND_PORT=9071` works for that run, but the `.env` the recipe generates is a straight copy of `.env.example` with `PORT=8080` / `FRONTEND_PORT=3000`, so the next bare `make selfhost` walks right back into the conflict the operator just worked around.

This PR adds a preflight that runs before the `.env` bootstrap, the pull and the up, and makes a manually chosen port persist.

**Which ports get checked comes from Compose, not from re-deriving the chain.** `scripts/selfhost-preflight.sh` asks `docker compose config` which host ports the stack would publish. Re-deriving `BACKEND_PORT → API_PORT → SERVER_PORT → PORT → 8080` a third time is #6145, and the alias order is only half of that bug — which source wins (env file, environment, make command line) differs per entry point too. `scripts/selfhost-wait.sh` already carries this rule for the health probe; the preflight follows it.

**How a port is tested.** A connect probe on `127.0.0.1` via bash's `/dev/tcp`, because that is the question being asked: `docker-compose.selfhost.yml` publishes on `127.0.0.1` only, so a listener on another interface does not collide and must not be reported. `port_free()` in `scripts/dev-env.sh` uses `lsof -sTCP:LISTEN`, which lists every interface — right for that script, wrong here — and silently reports "free" on a host with no `lsof`, the common case on a small self-host server. `lsof` is used only to decorate the message with a name.

**What is deliberately not a conflict:** a port already published by this project's own container. Re-running `make selfhost` is how the stack picks up a new image, so it stays idempotent. Checked with `docker compose port`, the same primitive `selfhost-wait.sh` uses.

The message lists every conflicting port at once — a machine with both 8080 and 3000 taken should not discover them one run at a time:

```
  Port 8080 (backend) is already in use by the Docker container qbittorrent.
  Port 3000 (frontend) is already in use by the Docker container gitea.

Multica was not started; no images were pulled.

Pick free ports and run again:
  make selfhost PORT=<free port> FRONTEND_PORT=<free port>

They are written to the .env that run creates, so later runs keep them.
```

Docker is asked to name the holder before `lsof` is, because that is the case `lsof` describes worst: the listening socket belongs to root's `docker-proxy`, so a non-root `lsof` sees nothing at all and a root one reports `docker-proxy`, which names no container.

An escape hatch, `MULTICA_SELFHOST_SKIP_PORT_CHECK=1`, exists for the test suite and is documented in the script header.

The `.env` bootstrap in both recipes now also records `PORT` and `FRONTEND_PORT` as the recipe resolved them. An existing `.env` is still never rewritten. If maintainers would rather not persist the ports, that part is a self-contained hunk and can be dropped without touching the rest.

## Related Issue

No existing issue — happy to open one if you would like this tracked. Context is #6145, whose rule ("nothing re-derives the published port from the inputs") this change extends from the health probe to the pre-start check.

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `scripts/selfhost-preflight.sh` (new) — reads the host ports from `docker compose config`, probes each on `127.0.0.1`, skips ports this project already publishes, and exits 1 listing every conflict with its holder.
- `Makefile` — `selfhost` and `selfhost-build` call it right after `$(REQUIRE_COMPOSE)`, before the `.env` bootstrap, the pull and the up. Both `.env` bootstraps now write `PORT` and `FRONTEND_PORT`.
- `scripts/selfhost-config.test.sh` — the `docker` stub learns `config`; the throwaway checkout gets the new script; `run_recipe` gains a no-`.env` mode and skips the check by default; three new cases.
- `.github/workflows/ci.yml` — `scripts/selfhost-preflight.sh` added to the path filter that runs the suite.
- Docs — `apps/docs/content/docs/troubleshooting.mdx` (the canonical "Port already in use" section), `apps/docs/content/docs/self-host-quickstart.mdx`, `SELF_HOSTING_AI.md`, `SELF_HOSTING_ADVANCED.md`.

## How to Test

1. `bash scripts/selfhost-config.test.sh` — the suite CI already runs; the three new cases claim their own loopback ports from the kernel, so they assert the behaviour without assuming anything about what the machine has free.
2. Occupy a default port (`docker run -d -p 3000:80 --name port-hog nginx`) and run `make selfhost`. It must fail in about a second, name port 3000 and `port-hog`, and pull nothing. `.env` must not be created.
3. `make selfhost PORT=<free> FRONTEND_PORT=<free>` starts normally; `grep -E '^(PORT|FRONTEND_PORT)=' .env` shows the chosen ports, and a later bare `make selfhost` stays on them.
4. With the stack up, `make selfhost` again must still work — its own published ports are not treated as a conflict.

Verified on a machine where 8080 and 3000 were genuinely taken (by qbittorrent and gitea), which is what prompted this.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A
- [ ] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs — N/A
- [ ] If this PR touches Chinese product copy, I checked it against `conventions.zh.mdx` — N/A, English pages only

### Notes for reviewers

- **Localized docs.** The `zh`/`ja`/`ko` twins of `troubleshooting.mdx` and `self-host-quickstart.mdx` are untouched, on the assumption they follow your translation flow. Happy to include them if you would rather they land together.
- **Out of scope, but related:** `scripts/install.sh:389` and `scripts/install.ps1:445` bring the stack up without a port check either — and `install.ps1` does not check `$LASTEXITCODE` after `up -d`, so on Windows a failed `up` continues to the wrong message ("could not read the backend host port"). Pre-existing and independent of this change; I can send a follow-up.

## AI Disclosure

**Prompt / approach:** I asked for a plan first: check the port before anything else, exit immediately with an explanation if it is taken, then let the port be chosen manually — and update the docs if needed. I reviewed and approved the plan (fail-fast with instructions, no interactive prompt and no silently picking a free port, so it stays deterministic in CI and under `curl | bash`) before any code was written. Each behaviour was verified against the real `make` recipes, and the two Makefile changes were mutation-checked: removing the preflight call and removing the `PORT` persistence each make a new test fail.

https://claude.ai/code/session_01D7WdJxpcXN6i8MMzfBpNX7

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,7 @@ jobs:
               - 'scripts/dev-env.sh'
               - 'scripts/dev-env.test.sh'
               - 'scripts/selfhost-wait.sh'
+              - 'scripts/selfhost-preflight.sh'
               # selfhost-config.test.sh asserts on the installers' side of
               # the port contract, so a change to either must run it.
               - 'scripts/install.sh'
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,17 @@ define REQUIRE_COMPOSE
 	esac
 endef
 
+# Both self-host recipes check the host ports before anything else costly, via
+# scripts/selfhost-preflight.sh. A busy port used to surface as Docker's own
+# "port is already allocated" after the image pull, or worse as a
+# successful-looking run (scripts/selfhost-wait.sh reports "still starting" and
+# exits 0), so the check pays for itself the first time 8080 or 3000 is taken.
+#
+# The .env bootstrap then records PORT and FRONTEND_PORT as the recipe resolved
+# them, so `make selfhost PORT=9090` is remembered. Without it the generated
+# .env always says 8080/3000 and the next bare `make selfhost` walks back into
+# the conflict the operator just worked around.
+
 # Default target changed from selfhost to help: bare `make` now prints this help
 # instead of launching a full Docker Compose build, which is safer for onboarding.
 .DEFAULT_GOAL := help
@@ -81,6 +92,7 @@ makehelp: help ## Alias for `make help`
 
 selfhost: ## Create .env if needed, then pull and start the official self-hosted images
 	$(REQUIRE_COMPOSE)
+	@bash scripts/selfhost-preflight.sh official
 	@if [ ! -f .env ]; then \
 		echo "==> Creating .env from .env.example..."; \
 		cp .env.example .env; \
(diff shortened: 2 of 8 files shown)


multica-eve:
Thanks for the thoughtful contribution. We're closing this because we don't want to add a separate port-preflight mechanism here. Docker Compose already returns a non-zero status and identifies the allocated port, Make stops at that failure, and the pulled images are cached for the retry after the operator changes the port. Moving that error earlier is not enough benefit to justify another default, cross-platform detection path and its maintenance surface. We can revisit a smaller improvement around the actual failure message if user evidence shows Docker's error is insufficient.

PR #8278, opened 2026-09-10
MUL-7239 fix(agents): preserve provider-qualified model ids

### LINKED ISSUE #8246: WebUI: agent model field silently stripped of provider prefix (fully-qualified id saved as bare id), causing runtime routing failures
### Summary

When editing an agent's **model** field through the Multica Web UI, a fully-qualified model id that the user types (e.g. `commandcode/deepseek/deepseek-v4.1-flash`) is persisted to the server as a **bare id without the provider prefix** (`deepseek/deepseek-v4.1-flash`). The same value saved through the CLI (`multica agent update --model ...`) is stored verbatim. For runtimes whose model ids require the provider prefix (oh-my-pi in our case), the silently-stripped value makes the agent un-routable and every task fails with `exit status 1`.

### Environment

- Multica workspace: hosted SaaS (`api.multica.ai`)
- Runtime: oh-my-pi (OMP) v18, local runtime on Ubuntu 24.04
- Agent model field edited via Web UI agent settings
- Date observed: 2026-09-10

### Reproduction / evidence chain

1. Agent previously worked with model `commandcode/deepseek/deepseek-v4-flash` (set via CLI).
2. User edited the model field in the Web UI, typing the fully-qualified `commandcode/deepseek/deepseek-v4.1-flash`. Immediately after saving:
   - `multica agent get <id>` returned `"model": "deepseek/deepseek-v4.1-flash"` — **the `commandcode/` prefix was gone**.
   - A test task dispatched to the agent failed: `pi exited with error: exit status 1`. The OMP runtime routed the bare id to its built-in openrouter provider (no key for it), producing a 404 / exit 1. Session log had no `model_change` record consistent with the intended provider.
3. **Server-side storage is verbatim** (verified by controlled round-trips on a throwaway agent via CLI):
   - create with `commandcode/deepseek/deepseek-v4.1-flash` → reads back identical
   - update to `deepseek/deepseek-v4.1-flash` → reads back identical (bare)
   - update back to fully-qualified → reads back identical
   So the API does **not** normalize/strip; the transformation happens in the Web UI layer (most plausibly the model input/picker mapping the typed text onto a catalog entry whose `id` lacks the provider prefix, or a free-text field being rewritten on submit).
4. After re-saving the fully-qualified value via `multica agent update`, the next dispatched task ran with the correct provider: session log recorded `{"type":"model_change","model":"commandcode/deepseek/deepseek-v4.1-flash","resolvedModelIsFallback":false}` and the task completed successfully.

### Expected behavior

The model value saved from the Web UI must be stored **exactly as entered**. If the UI wants to offer catalog/autocomplete suggestions, it must not rewrite user input that already matches a valid fully-qualified id (or at minimum it must show/confirm the value that will actually be saved).

### Actual behavior

Web UI silently persisted the model id without the provider prefix; the agent then failed on every dispatch with a confusing `exit status 1` while the user believed they had configured a valid model.

### Impact

- Silent misconfiguration: nothing in the UI warns that the saved value differs from the entered one.
- Failure mode is opaque to users: tasks fail with a generic runtime exit code; connecting it to the stripped prefix requires comparing the stored field.
- Affects any runtime that needs provider-qualified model ids (oh-my-pi with `provider/model` routing, e.g. `commandcode/...`, `zen/...`); runtimes that accept bare ids (Codex, Claude) mask the bug, which is why it is easy to miss.

### Workaround

Edit the agent model via the CLI instead of the Web UI (`multica agent update <id> --model '<provider>/<model>'`), then verify with `multica agent get <id>`.

### Suggested fixes (any one would have caught this)

1. Store free-typed model strings verbatim (no catalog normalization), or preserve the exact input when it already contains a `/`-separated provider prefix.
2. After save, echo the persisted value in a confirmation toast/banner so a silent rewrite is visible.
3. Server-side validation endpoint: warn (UI banner) when a model id is not resolvable by the target runtime — the daemon already knows which models its runtime resolves.

multica-eve:
## Summary

- keep the verbatim custom-model option when typed text matches only a catalog label
- continue suppressing the custom option when the typed text exactly matches the canonical model ID
- cover both the create-agent dropdown and inspector picker with the provider-qualified ID regression

Fixes #8246

## Verification

- `pnpm --filter @multica/views exec vitest run agents/components/model-dropdown.test.tsx agents/components/inspector/model-picker.test.tsx` (12 tests)
- `pnpm --filter @multica/views test` (419 files, 5035 tests)
- `pnpm --filter @multica/views lint` (0 errors; 26 warnings outside the changed files)

### DIFF
--- a/packages/views/agents/components/inspector/model-picker.test.tsx
+++ b/packages/views/agents/components/inspector/model-picker.test.tsx
@@ -87,6 +87,34 @@ describe("ModelPicker (inspector)", () => {
     expect(onChange).toHaveBeenCalledWith("claude-fable-5");
   });
 
+  it("keeps an exact label match available as a verbatim custom model id", async () => {
+    const qualifiedId = "commandcode/deepseek/deepseek-v4.1-flash";
+    discovery = async () => ({
+      models: [
+        {
+          id: "deepseek/deepseek-v4.1-flash",
+          label: qualifiedId,
+          provider: "commandcode",
+        },
+      ],
+      supported: true,
+    });
+
+    const { container, onChange } = renderPicker();
+    openPicker(container);
+    fireEvent.change(
+      await screen.findByPlaceholderText(
+        enAgents.pickers.model_search_placeholder,
+      ),
+      { target: { value: qualifiedId } },
+    );
+
+    fireEvent.click(
+      await screen.findByRole("button", { name: `Use "${qualifiedId}"` }),
+    );
+    expect(onChange).toHaveBeenCalledWith(qualifiedId);
+  });
+
   it("shows an unavailable model with its reason and no way to select it", async () => {
     const { container, onChange } = renderPicker();
     openPicker(container);
--- a/packages/views/agents/components/inspector/model-picker.tsx
+++ b/packages/views/agents/components/inspector/model-picker.tsx
@@ -80,10 +80,11 @@ export function ModelPicker({
   }, [models, search]);
 
   const trimmedSearch = search.trim();
-  const exactMatch = models.some(
-    (m) => m.id === trimmedSearch || m.label === trimmedSearch,
-  );
-  const canCreate = trimmedSearch.length > 0 && !exactMatch;
+  // Labels are display copy, not aliases for the persistable model ID. Keep
+  // the verbatim option when only a label matches so a provider-qualified
+  // value cannot be silently replaced with that row's different ID (#8246).
+  const exactIdMatch = models.some((m) => m.id === trimmedSearch);
+  const canCreate = trimmedSearch.length > 0 && !exactIdMatch;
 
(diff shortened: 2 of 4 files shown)


Bohan-J:
Closing this. The environment details the reporter provided on #8246 show this change cannot affect that bug.

Their runtime is a custom profile with `protocol_family: "pi"` running `omp`, and on that path model discovery returns an empty catalog — so there is no row whose label could match the typed text, and the condition this PR changes never fires. Even on the built-in Oh-My-Pi path, the catalog names for that model (`CommandCode DeepSeek V4.1 Flash`, `DeepSeek V4.1 Flash`) never equal the id being typed, so it would not fire there either.

The root cause of #8246 is still being investigated, and the issue stays open. If we later find a runtime whose catalog reports a display name equal to some other row's id, this change can be revisited on its own merits.

Issue #8221, opened 2026-09-09
[Bug]: Network partition strands a finished run in `running` forever and silently drops its comment deliverable, deadlocking the delegating agent


karlarso:
### Deployment type

Official App

### Area

Agents and runtimes

### What happened?

A ~17-minute network partition between the local daemon and `api.multica.ai` produced **two** failures from one root cause. The second one is the damaging half and I don't think it's covered by the existing reports.

**1. The task is permanently stuck in `running`.**

The agent finished successfully at `18:07:53` local (`agent finished ... status=completed duration=4m8s tools=19`). The `/complete` callback then exhausted its retries and the daemon deliberately gave up:

```
18:14:27.684 ERR complete task failed after retries; leaving task in running rather than
falling back to fail  component=daemon task=01a0859f-9b89-71ea-99ab-1442f3acd070
error="Post \"https://api.multica.ai/api/daemon/tasks/.../complete\": context deadline
exceeded (Client.Timeout exceeded while awaiting headers)"
```

Not falling back to `fail` is the right call on its own — a successful run must not be recorded as a failure. But nothing reconciles it afterwards. Connectivity came back ~8 minutes later and the daemon heartbeated continuously for the next 3.5 hours; the server row stayed `running` the whole time. The local process was long gone (nothing matching the run's workdir in `ps`). `attempt=1 / max_attempts=2`, so the retry never fires either — retry is gated on the task reaching `failed`, which by design it never does.

**2. The agent's actual deliverable was silently dropped, and the run still reported success.**

This run's entire job was to post one comment (a plan review verdict) back to the issue. The `comment add` call landed inside the partition window and timed out. The agent noticed and tried to self-check before giving up:

```
18:06:48.570 DBG agent task=01a0859f-... text="评审已完成，结论为「需改进」，包含两项 P1。
平台提交请求超时，我正在核对评论是否已落库，避免重复发布。"
18:07:48.543 DBG agent task=01a0859f-... text="…评论提交及回读均超时…"
```

Both the write and the read-back timed out, so the agent correctly refused to double-post — and then the run terminated as `status=completed`. The comment never reached the server. The work product only survived because the agent happened to have staged it in a scratch file in its workdir; had it composed the comment inline it would have been gone entirely.

**3. In a multi-agent setup this deadlocks the delegating agent, with no timeout and no alert.**

[...]

Issue #8272, opened 2026-09-10
[Bug]: Transient 404 on task status kills a healthy run; server row then stays 'running' forever (agent stuck 'working', no retry)


karlarso:
### Deployment type

Official App

### Area

Agents and runtimes

### What happened?

During a ~5 minute API degradation on api.multica.ai (2026-09-10 10:34–10:39 UTC), a healthy in-flight issue task was killed by the local daemon and its run was then left in `running` on the server for 4+ hours. The agent showed as `working` the whole time, the run never got re-dispatched (attempt 1 of max 2), and the agent's closing work (a description update + final comment) was lost. `multica issue cancel-task` was needed to clear it manually.

Two defects combine to produce this:

**1. Server can return `404 task not found` for a task that still exists, and the daemon treats that as "row deleted → kill agent".**

`requireDaemonTaskAccessWithWorkspace` (`server/internal/handler/daemon.go`) is careful to map only `pgx.ErrNoRows` to 404 — the comment there says "a transient DB error must not be reported as a deletion". But a few lines later it also emits `404 task not found` when `ResolveTaskWorkspaceID` returns `""`. `ResolveTaskWorkspaceID` (`server/internal/service/task.go`) swallows *any* error from `GetIssue` / `GetChatSession` / `GetAutopilotRun` and falls through to `""`. So a DB timeout on the issue lookup becomes a 404 that `shouldInterruptAgent` / `isTaskNotFoundError` (`server/internal/daemon/daemon.go`, `client.go`) accepts as a deletion signal, and `watchTaskCancellation` cancels the run context.

Evidence that this was the branch taken: within 0.5 s of the 404, `POST .../messages` for the same task returned `500 failed to load task` (i.e. `GetAgentTask` was erroring, not `ErrNoRows`), and the agent's own CLI calls (`multica issue get` / `comment list`) had been timing out and then got a 404 on the *issue* at the same second. The task row was never deleted — `multica issue runs` still listed it as `running` hours later.

**2. After a daemon-initiated interrupt, nothing ever finalizes the server-side row.**

[...]

PR #8281, opened 2026-09-10
MUL-7259, MUL-7199: fix(server,daemon): converge tasks whose runtime stopped executing them

Bohan-J:
> Part 2 of MUL-7259. Part 1 (#8279, error classification) is merged; this branch was rebased onto `main` after it, so the diff is Part 2 only.

## Summary

A terminal callback is the only thing that moves a task out of `running`. When one is lost or never sent, the row stays `running` **forever** — `FailStaleTasks` deliberately skips rows whose runtime is still heartbeating, and a healthy daemon heartbeats indefinitely. The perverse property is that the healthier the daemon, the less recoverable the row.

The damage is not just an untidy row:

- the agent stays `working` (`RefreshAgentStatusFromTasks` derives status from `dispatched`/`running`),
- **one of the agent's `max_concurrent_tasks` slots is consumed permanently** (`CountRunningTasks` gates claiming on the same statuses) — this accumulates across incidents, so with the default of 6 an agent stops claiming entirely after six,
- the retry budget is unreachable, because `MaybeRetryFailedTask` opens with `if parent.Status != "failed"`.

Manual `cancel-task` was the only recovery. There are two entrances to that state; this fixes both, because closing only one leaves the other wide open.

## Server — cancel-ack converges an abandoned row

`AckTaskCancelled` assumed cancellation is always server-initiated, so by ack time the row is terminal and only the branch/error pointers remain to record. All three of its writes CAS on `status='cancelled'`. Against a `running` row they were silent no-ops and the endpoint returned `200` having changed nothing — the reproduction probe shows two consecutive 200s leaving a five-hour-old task untouched.

A daemon only acks after its run has stopped and its transcript is flushed, so a non-terminal row at that moment is stale by definition. `FinalizeAbandonedTask` converges it to `failed` with a new `runtime_abandoned` reason, carrying branch and durable workdir in the same statement (the pointer CASes below it would miss the row we just failed).

**The status CAS is the whole safety argument, in both directions:**

- a user cancel commits `cancelled` *before* the daemon acks, so this converges nothing and falls through to the existing record-only path — a real cancel can never be rewritten as a failure;
- a replayed or duplicated ack finds a terminal row and is a no-op, so at-least-once delivery stays safe.

`runtime_abandoned` is in `retryableReasons`, and that is what makes this a recovery rather than a tidier tombstone: converging to `cancelled` would free the slot and still lose the run. `HandleFailedTasks` then applies the existing rules unchanged — exhausted budget or a non-retryable reason still does not retry.

New reason rather than reusing `runtime_recovery`, which is documented as specifically a daemon *restart*. Here the runtime never went away, which is exactly why no sweeper reclaims the row.

## Daemon — durable outbox for terminal reports

The cancel-ack fix alone would not touch #8221, where **no 404 is involved at all**: a partition exhausted the `/complete` retries and the daemon deliberately left the task running rather than reporting a successful run as a failure. That call is right on its own; nothing reconciled it afterwards. The existing code comment said as much — *"leave the task in running instead so a future fix (server-side stuck-task reaper, or a daemon-side persistent pending queue) can recover it"*.

`reportTerminalTask` now persists a transiently-failed report to disk and the poll loop replays it once the server is reachable, including across a restart. `terminalTaskReport` already existed as the single choke point for exactly this ("gives the durable outbox one insertion point without revisiting every task exit when it is added").

- Atomic writes: temp file, fsync, rename — a half-written report on restart would be the same lost callback this prevents.
- Replay is safe because `CompleteTask` already treats an already-finalized callback as idempotent success.
- Permanent 4xx are dropped rather than retried forever; the caller's existing complete→fail downgrade gets its own durability.
- Reports are abandoned after 7 days so the queue cannot grow without bound.
- One file per task, so the complete→fail downgrade replaces rather than queues behind.

## Testing

Postgres 17 via `docker compose`, migrations applied to a scratch database. Both probes attached to MUL-7259 confirmed reproducing on `b36906e6d` first; both now fail against these branches.

Server (`internal/handler`):

- converges an abandoned `running` row to `failed`, with `completed_at`, branch and workdir preserved
- **releases the concurrency slot** (`CountRunningTasks` 1 → 0) and returns the agent to `idle`
- spends the remaining retry (attempt 1/2 → a child at attempt 2)
- exhausted budget does not retry; a daemon-reported non-retryable reason (`local_directory_error`) does not retry and wins over the default
- **a user cancel stays `cancelled`**, still receives its branch/error pointers, and is not retried
- a `completed` row is untouched — convergence never downgrades a success
- three replayed acks produce exactly one failure and one retry

Daemon (`internal/daemon`):

- persist/list/remove round-trip preserving every field; later report replaces earlier
- corrupt entries are discarded, not retried forever; unsafe task IDs (`../escape`) are refused as filenames
- transient failure is queued; permanent rejection is not; success clears a stale queued copy
- **end-to-end recovery**: queued while the server returns 503, kept (not dropped) on a drain during the outage, delivered intact to the right endpoint after recovery, and a second drain does not re-send
- expired reports are abandoned without being sent

```
go build ./...                                     ok
go vet ./...                                       ok
go test ./internal/handler/ -count=1               ok (full suite)
go test ./internal/service/ -count=1               ok
go test ./pkg/taskfailure/ -count=1                ok
go test -race ./internal/daemon/ -count=1          7 failures, identical set on b36906e6d
```

The 7 daemon failures are pre-existing sandbox issues (HOME/profile-dir resolution), verified by running the same suite on the base commit. An earlier run of this suite surfaced a genuine 8th failure — `TestReportTaskResult_TransientCompleteExhaustedDoesNotFallback` builds a `Daemon` literal with no outbox and my persist path lacked the nil guard I had added elsewhere. Fixed; the count is back to the baseline set.

CI will run these against its own PostgreSQL service.

## Risks

- Tasks that would previously hang in `running` now become `failed` and may auto-retry. That is the intent, and the retry stays behind the existing budget and reason gates.
- Convergence is reachable only through a daemon's own cancel-ack on a non-terminal row; a user cancel and a completion are both protected by the CAS and covered by tests.
- The outbox writes under `WorkspacesRoot/.terminal-outbox`, alongside the existing `.repos` and `.skill-cache`. No migration; the new column values are existing columns.

### DIFF
--- a/server/internal/daemon/daemon.go
+++ b/server/internal/daemon/daemon.go
@@ -482,6 +482,13 @@ type Daemon struct {
 	// (watchTaskCancellation, workspaceSyncLoop) so the WS connect/reconnect
 	// path can shrink coarse fallback reconciliation gaps to sub-second. See
 	// reconcile.go and runTaskWakeupConnection.
+	// terminalOutbox holds complete/fail reports whose in-process retry
+	// schedule was exhausted while the server was unreachable. Without it those
+	// reports are dropped and the task row stays 'running' forever, because a
+	// still-heartbeating daemon is exactly what the stale-task sweeper skips
+	// (GH #8221).
+	terminalOutbox *terminalOutbox
+
 	reconcile *reconcileBroadcaster
 	// workspaceChanges is the account-scoped server hint for membership-set
 	// changes. It stays separate from reconcile because a membership hint only
@@ -667,6 +674,7 @@ func New(cfg Config, logger *slog.Logger) *Daemon {
 		cancelPollInterval:        5 * time.Second,
 		envRootBusyWait:           15 * time.Second,
 		taskPrepareTimeout:        defaultTaskPrepareTimeout,
+		terminalOutbox:            newTerminalOutbox(filepath.Join(cfg.WorkspacesRoot, ".terminal-outbox")),
 		reconcile:                 newReconcileBroadcaster(),
 		workspaceChanges:          newWorkspaceChangeSignal(),
 		wsRPC:                     newWSRPCClient(wsRPCResponseGrace),
@@ -5067,6 +5075,13 @@ func (d *Daemon) runBatchPoller(pollerCtx, parentCtx context.Context, sem chan i
 			return
 		}
 
+		// Replay any terminal report a previous run could not deliver. This runs
+		// before the claim, and on the same loop that proves the server is
+		// reachable, so a task finished during an outage is finalized as soon as
+		// connectivity returns instead of sitting 'running' until someone
+		// notices. Cheap when the queue is empty: one directory read.
+		d.drainTerminalOutbox(pollerCtx)
+
 		runtimeIDs := d.allRuntimeIDs()
 		if len(runtimeIDs) == 0 {
 			if err := sleepWithContextOrWakeup(pollerCtx, d.cfg.PollInterval, wakeup); err != nil {
@@ -5917,15 +5932,18 @@ func (d *Daemon) reportTaskResult(ctx context.Context, taskID string, result Tas
 		// error reaching us here means the schedule was exhausted while
 		// the upstream was still 5xx / unreachable. Converting that into
 		// a fail would lose the agent's actual result and surface a
-		// misleading red badge in the UI — leave the task in running
-		// instead so a future fix (server-side stuck-task reaper, or a
-		// daemon-side persistent pending queue) can recover it. Only
-		// permanent server-side rejections (4xx other than 408/429)
-		// warrant the legacy fallback, because at that point the server
-		// has already refused this task and the only useful UI signal
-		// left is a concrete failure.
+		// misleading red badge in the UI. Only permanent server-side
+		// rejections (4xx other than 408/429) warrant the legacy
+		// fallback, because at that point the server has already refused
+		// this task and the only useful UI signal left is a concrete
+		// failure.
+		//
+		// reportTerminalTask has already queued this report durably, so the
+		// run is no longer lost when the schedule runs out: the poll loop
+		// replays it once the server is reachable. This used to be the point
+		// where a finished run became a permanently 'running' row (GH #8221).
 		if isTransientError(err) {
-			taskLog.Error("complete task failed after retries; leaving task in running rather than falling back to fail", "error", err)
+			taskLog.Error("complete task failed after retries; queued for replay rather than falling back to fail", "error", err)
 			return
 		}
 		taskLog.Error("complete task rejected by server, falling back to fail", "error", err)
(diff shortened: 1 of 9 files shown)


multica-eve (changes requested):
Request changes: the outbox is not write-ahead, and startup runs orphan recovery before replay, so a daemon restart can still discard a successful terminal result.

The underlying problem is real and worth fixing, but this PR does not yet close the user-visible failure in #8221: a finished delegated run can still lose its actual comment and fail to wake its coordinator.

Must-fix findings:

1. **P0 — Persist before the first network attempt.** `reportTerminalTask` calls `sendTerminalTask` first and only persists after the entire transient retry schedule returns (`server/internal/daemon/daemon.go:6022-6042`). Shutdown only waits 30 seconds for task goroutines while this send can run for up to six minutes (`daemon.go:5029-5043`), so an update, crash, or power loss during the request/retry window leaves no outbox record. A persist failure is also only logged (`daemon.go:6048-6055`). Write and fsync the envelope (including the parent directory) before sending; only delete it after an exact server acknowledgement, and fail closed/stop claiming if the terminal result cannot be persisted. Add kill-point tests for pre-send, in-flight send, and committed-response-lost cases.

2. **P0 — Fix startup ordering and terminal conflict semantics.** `Run -> preflightAuth -> syncWorkspacesFromAPI` calls `RecoverOrphans` first (`daemon.go:2042-2044,4040-4048`); the outbox is not drained until the poller starts (`daemon.go:5078-5083`). After a restart, the server therefore marks the original row `failed/runtime_recovery` and may create a retry before the pending completion is replayed. `CompleteTask` then treats *any* existing terminal state as idempotent success (`server/internal/service/task.go:4375-4386`), causing the daemon to delete the successful result without applying it. Either exclude pending task IDs from orphan recovery or drain them after authentication but before recovery, and return structured `applied / same-result duplicate / conflict` outcomes. Only the first two may acknowledge deletion. Cover the real startup sequence in an integration test.

3. **P0 — Make the final comment and coordinator wake-up durable.** This outbox persists only complete/fail envelopes, not the failed `multica issue comment add` that was the actual deliverable in #8221. The full review body in the incident lived only in a scratch file; terminal stdout was not an equivalent copy. Even the completion fallback is post-transaction (`task.go:4408-4451`), and `createAgentComment` silently returns on lookup/insert failure (`task.go:7257-7291`) while the endpoint still returns success. When insertion succeeds, it bypasses `computeCommentAgentTriggers` / `EnqueueTaskForSquadLeader`, so the delegating leader is still not awakened. Put task-owned final comments behind a stable idempotency key and durable delivery contract, and reuse the normal trigger semantics. The acceptance test should assert exactly one terminal state, one comment, and one leader wake-up after recovery.

4. **P1 — Do not silently destroy the only copy of a result.** Corrupt entries are deleted (`terminal_outbox.go:170-199`), and permanent 4xx or the hard-coded seven-day age also delete reports after a local-only log (`daemon.go:6101-6117`). A healthy heartbeat plus a terminal-endpoint/DB-specific outage means the server sweeper is not a backstop. Preserve a visible dead-letter/conflict state and bound growth through capacity/backpressure, not expiry of the only result.

5. **P1 — Do not block the machine-wide claim loop or shutdown.** The drain runs synchronously in the sole batch poller (`daemon.go:5078-5083`), while `sendTerminalTask` strips poller cancellation with `context.WithoutCancel` (`daemon.go:6058-6060`). One endpoint-specific failure can delay every claim for roughly six minutes, and shutdown waits for `pollerDone` without a bound. Use a separate single-flight drainer with cancellable, short replay attempts.

6. **P1 — The cancel acknowledgement remains lossy, and fail-to-retry is not atomic.** Both daemon paths still call `AckTaskCancelled` directly with finite in-memory retries (`daemon.go:5542,5591`), so the same partition can prevent the new convergence path from running at all. `ConvergeAbandonedTask` also commits `failed` before calling `HandleFailedTasks` outside the transaction (`task.go:2574-2621`); a crash or lost response after the commit makes a replay a terminal no-op, potentially losing the retry and notifications. Include cancel-ack in the durable terminal protocol and atomically persist failure plus retry/server-side follow-up, as `FailTask` already does.

Verification on base `7cd4d9c`, head `a7d2954`: `git diff --check`, targeted daemon outbox tests, `pkg/taskfailure`, service compilation, and `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` passed. The handler tests could not be executed against this checkout's stale local test database because it lacks the pre-existing `durable_work_dir` column; that is a local validation gap, not a claimed PR regression.

Bohan-J:
Closing without merging.

Review showed this doesn't deliver the durability it claims:

- The outbox is not write-ahead. A report is only persisted after the in-process retry schedule is exhausted, so a restart or crash during the first attempt or its backoff still loses it.
- On startup, orphan recovery runs before the outbox is replayed. A persisted successful result can therefore be replayed against a row that has already been failed, get accepted as an idempotent no-op, and be deleted — so "survives a restart" does not hold.
- Cancel-acks are not routed through the outbox, so the server-side convergence never gets the signal after an outage longer than the retry window.
- Replay runs inline on the claim loop and ignores cancellation, so one undeliverable report can stall task claiming and delay shutdown.
- Failing the task and creating its retry are not committed atomically, and reports are deleted after 7 days without the server confirming anything.

#8279 has already removed the trigger in #8272 (a failed workspace lookup reported as `404 task not found`). The remaining problem — a finished run whose terminal report never reaches the server — is tracked in #8221, and will be redone there with write-ahead persistence, recovery ordering, and a replay worker that runs outside the claim loop. The branch is kept for reference.

Issue #5731, opened 2026-07-21
[Feature]: multica CLI: expose the current agent's Feishu channel context (bot credentials + session chat_id)


lwydyby:
### Deployment type

Official App

### What do you want and why?

Background and motivation

When an agent runs inside a Multica runtime, it needs to operate Feishu through larkcli (https://github.com/larksuite/cli). Today, multica's built-in Feishu notification only supports simple markdown message replies, which falls short of real agent workflows:

Chat history access is inflexible: the platform can fetch chat history, but only as a fixed, passive injection — the agent cannot query history on demand mid-task (choosing time ranges, paginating, searching, reading related threads or chats, or re-fetching as the conversation evolves). The initiative over context retrieval isn't in the agent's hands.
Limited reply forms: all results pile into one flat conversation — long-running tasks and multi-turn interactions flood the chat. Feishu threads work far better in many scenarios: one task aggregates under one thread, keeping the main chat clean and context naturally grouped.
Limited output forms: complex scenarios like analysis results or reports are far better delivered as Feishu docs / bitables, and rich text, interactive cards, and file messages are unavailable.
larkcli already covers all of these capabilities, but it needs two pieces of information to work:

Bot credentials: larkcli accepts credentials via the LARKSUITE_CLI_APP_ID / LARKSUITE_CLI_APP_SECRET environment variables, but the agent currently has no way to retrieve its own bound Feishu bot's app_id / app_secret from the multica CLI. Credential injection requires manual user setup, breaking the automation loop. Credentials should be managed and issued by the platform; querying them via the CLI is the zero-config, least-privilege way for agents to obtain them.
Chat context: larkcli already supports sending messages to a specific chat as the bot (lark-cli im +messages-send --as bot --chat-id <id>), but the agent cannot learn the current session's Feishu chat_id from the multica CLI — without knowing where to send, thread replies, on-demand history fetches, and everything else are impossible, and it falls back to the built-in capabilities.
multica doesn't need to anticipate what the agent will do with Feishu; it only needs to tell the agent "who am I (bot credentials)" and "where am I (chat_id)", and leave the rest to larkcli — no per-scenario wheels to reinvent on the platform side.

背景与动机

Agent 在 Multica 运行时中执行任务时，需要通过 larkcli（https://github.com/larksuite/cli）操作飞书。目前 multica 内建的飞书通知能力只支持简单的 markdown 消息回复，对 agent 的真实工作场景覆盖不足：

[...]

PR #8282, opened 2026-09-10
MUL-7265: feat(lark): add agent-scoped Feishu document tools

YJ-Hao:
## Summary

- expose six bounded Feishu document operations through a task-scoped built-in MCP endpoint
- bind every capability to the exact workspace, Agent, runtime, daemon, installation, and installation revision
- automatically mount the tools for all task sources when the Agent has an active Feishu bot connection
- keep Feishu application credentials on the backend and use the bot document ACL as the authorization boundary

## Safety

- accept only explicit Feishu or Lark document URLs and fixed tool schemas
- require fresh reads, unique text replacement, bounded content and block ranges, and confirmation for destructive or large edits
- revalidate task and installation state on every call so disconnects and reconnects revoke stale capabilities
- avoid blind retries when a write result is uncertain

## Verification

- go test ./... -count=1 -p=1
- go build ./...
- go vet -tags agentintegration ./...
- sqlc v1.31.1 generate with a clean worktree afterward

### DIFF
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -601,6 +601,7 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 				// backfills) take it directly; the constructor-based services
 				// wrap *db.Queries internally, so they keep taking queries.
 				cs := lark.NewChannelStore(queries)
+				h.LarkDocuments = lark.NewDocumentService(cs, installSvc, larkClient, slog.Default())
 				patcher := lark.NewPatcher(cs, installSvc, larkClient, lark.PatcherConfig{})
 				patcher.Register(bus)
 
@@ -1449,6 +1450,10 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 		// The broker asks for an mcp hook's credential at connection time, so
 		// a secret never sits in a task record.
 		r.Get("/tasks/{id}/plugin-mcp/{contributionId}/credential", h.ResolvePluginMCPCredential)
+		// Built-in Feishu document MCP: both the credential exchange and every
+		// call revalidate the exact live task, daemon, Agent and installation.
+		r.Get("/tasks/{id}/remote-mcp/{contributionId}/credential", h.ResolveRemoteMCPCredential)
+		r.Post("/tasks/{id}/feishu-documents/{contributionId}/mcp", h.ServeFeishuDocumentsMCP)
 
 		r.Post("/runtimes/{runtimeId}/tasks/claim", h.ClaimTaskByRuntime)
 		// Canonical machine-level batch claim (MUL-4257). `/claim` is a
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -601,6 +601,7 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 				// backfills) take it directly; the constructor-based services
 				// wrap *db.Queries internally, so they keep taking queries.
 				cs := lark.NewChannelStore(queries)
+				h.LarkDocumentInstallations = cs
 				h.LarkDocuments = lark.NewDocumentService(cs, installSvc, larkClient, slog.Default())
 				patcher := lark.NewPatcher(cs, installSvc, larkClient, lark.PatcherConfig{})
 				patcher.Register(bus)
--- a/server/internal/daemon/execenv/runtime_config_sections.go
+++ b/server/internal/daemon/execenv/runtime_config_sections.go
@@ -204,6 +204,7 @@ func BuildConnectedAppsBlock(apps []runtimeapps.ConnectedApp) string {
 	}
 	var b strings.Builder
 	var lines strings.Builder
+	hasFeishuDocuments := false
 	for _, app := range apps {
 		serverName := sanitizeBriefCodeToken(app.ServerName)
 		toolkitSlug := sanitizeBriefCodeToken(app.ToolkitSlug)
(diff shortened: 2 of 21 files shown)


Bohan-J (changes requested):
Thanks for this. We want agents to be able to work with Feishu docs through their own bot, and keeping the app credentials on the server is the right call (see also #5731). We're fine with two things as they are:

- **On by default.** Any agent with a Feishu bot connected gets the tools automatically.
- **The bot's document access as the permission boundary.** Whoever can invoke the agent gets whatever access the bot has, the same as other agent capabilities.

A few things need to change before this can merge, though.

## Must fix

### 1. Show that the upstream API works, or switch to documented APIs

All reads and writes go through `/open-apis/docs_ai/v1/documents/...` (`document_client.go:83`, `:122`). We couldn't find `docs_ai` in the public OpenAPI docs index for either Feishu or Lark. The documented APIs are the `docx/v1` block endpoints and `docs/v1/content`, and both support revision IDs. Every document test uses a fake client, and the PR's verification is `go test`, `build` and `vet` only.

Please do one of the following:
- Link the official docs for `docs_ai/v1` and share an end-to-end run on a real Feishu tenant and a real Lark tenant. It should cover a fetch, each write operation, a `/wiki/` link, a document the bot can't access, and an app that lacks the doc scope.
- Move to the documented APIs.

Related: `/wiki/<token>` is passed straight through as a document token. A wiki node token has to be resolved to its `obj_token` first, and the node may be a sheet or a bitable rather than a docx. `DocumentKindWiki` is defined but never used.

### 2. Don't use `channel_installation.updated_at` as the capability revision

The claim pins `UpdatedAt.UnixNano()` into the contribution ID (`feishu_documents_mcp.go:45-50`). Every call then requires an exact match (`feishu_documents_mcp.go:231-234`, `document_service.go:164-166`). But this column changes often, and for reasons unrelated to the bot:

- `AcquireChannelWSLease` sets `updated_at = now()` on every lease renewal (`channel.sql:412-419`).
- The default lease backend is Postgres (`cmd/server/router.go:272-275`), and leases renew every min(60s, TTL/3) (`channel/engine/supervisor.go:204-205`).
- On the Redis backend, the config and region backfills (`SetChannelInstallationConfig`, `BackfillChannelInstallationRegionToFeishuLark`) still write the same column.

With the default setup, every tool call made more than about a minute after the claim returns `capability_denied`.

Every call already re-checks that the installation is active and belongs to the agent, so the revision adds no security. Please drop it and bind the contribution to the installation ID plus the `app_id`.

### 3. Narrow v1 to non-destructive operations

`confirmed` is a boolean the model fills in itself. The server only checks that it is true, so no person ever confirms anything, and unattended runs have nobody to ask. For v1, please ship only:
- `fetch`
- `append`
- `replace_text`
- `insert_after`
- single-block `replace_block`

Drop `delete_block` and the range variants of replace. Remove the `confirmed` field and the unused `DocumentConfirmBlocks` constant (`document_protocol.go:17`), and update the PR description so it no longer says destructive edits require confirmation. Delete and range operations can come later, together with a real human-approval step.

### 4. Don't require a public HTTPS `MULTICA_PUBLIC_URL`

The MCP endpoint is built from `PublicURL` and goes through the broker's `ValidatePublicHTTPSEndpoint`, which only accepts HTTPS URLs that resolve to public IPs. As a result, self-hosted deployments on http or a private network get no tools, and the only sign is a warn log.

The credential lookup already uses the daemon's own API connection, so the MCP endpoint can too:
- Make the built-in contribution's endpoint a relative path that the daemon resolves against its own API base URL.
- Authenticate with the daemon token directly.
- Gate this behind a new daemon capability, the same way the PR already gates on `remote-mcp-v1`.

Please also revert the change that registers `/tasks/{id}/remote-mcp/{contributionId}/credential` as a Feishu-only handler (`router.go:1456`). The daemon uses that route by default for non-plugin connections (`daemon/client.go:244-247`), and it should stay available for that.

### 5. Separate "missing app scope" from "bot has no access to this document"

Feishu error `99991672` means the app hasn't been granted the API scope. The PR maps it to `permission_denied` (`document_service.go:307-309`), and the message tells the user to share the document with the bot, which won't fix it. The scan-to-install flow only sends `archetype=PersonalAgent` and requests no scopes (`registration.go:270-274`), so this is likely the first error new users hit.

Please return two distinct errors, each with the right fix:
- Missing scope: enable the doc scopes for the app and publish a new version.
- No document access: add the bot to the document.

### 6. Add a server-side kill switch

The feature is on by default and writes real documents in every workspace at once. Please add an env var that turns it off without rolling back the server. No UI is needed.

### 7. Only describe tools that are actually mounted

The "Feishu Documents" Connected Apps entry is added whenever the daemon advertises the capability (`handler/daemon.go:2195-2203`). It doesn't check whether the provider supports the broker, which currently covers only codex, claude, hermes, qoder and mcode (`daemon/remote_mcp_broker.go:159-166`), or whether the broker actually started.

The prompt also names the server `feishu-documents`, while the broker registers it as `plugin-feishu-documents-builtin:` (`remote_mcp_broker.go:193-205`). A model that is told about tools it can't reach will keep trying to call them.

## Should fix

- **State the content format for each tool.**
  - Writes: `append` sends markdown, while `replace_text`, `insert_after` and `replace_block` send XML (`document_service.go:209-231`).
  - Reads: a `full` fetch returns markdown without block IDs, and the other scopes return XML with block IDs (`document_service.go:177-190`).

  The schemas and tool descriptions don't mention either, so models will send markdown to the XML operations.
- **Keep the server's time budget under the broker's.** The server has 30s in total for fetch, update and verify (`feishuDocumentsCallTimeout`), and the broker's `remotemcp.CallTimeout` is also 30s. On a slow write the broker can time out first. The agent then sees a transport error instead of `timeout_uncertain` and may retry `append`, which is not idempotent.
- **Don't return the whole document after a write.** `append` and `replace_text` put the full document markdown in their result. That wastes context, and for large documents it can exceed the broker's 4 MB limit once JSON-encoded twice.
- **Nits.**
  - `Verified: true` only means the re-read succeeded; it doesn't check the content.
  - `formatDocumentRevision` is only used by tests.
  - `DocumentToolInput.RevisionID` is never used.

## What we verified

On `38d3820`:
- `go build ./...` and `go vet` on the touched packages pass.
- The lark package tests pass.
- The `execenv` ByteIdentical tests pass. The brief is unchanged, and the Feishu guidance only goes into the per-turn prompt, which is correct.
- The handler tests pass against a real Postgres, including all 13 new tests.
- `sqlc generate` reproduces the committed output.

We did not make any real Feishu or Lark calls. Finding 2 comes from reading the code path, not from an end-to-end run.

## Summary

Show that the API works, replace `updated_at` with a stable identity, narrow v1 to the non-destructive operations, and make the endpoint work without a public URL. With those done, turning it on by default is fine.

PR #8285, opened 2026-09-11
MUL-7268: fix(server): reject claims with unreadable comment inputs

importcpp:
## What does this PR do?

A database error while reading a primary comment, coalesced handoff instruction, or reply thread can otherwise dispatch a task without the sender's constraints. Reject that claim: the single-runtime endpoint returns 500, and the batch endpoint omits the affected task. The unstarted task remains dispatched for existing lease-based redelivery, without committing a delivery receipt or task token.

Missing comments (`pgx.ErrNoRows`) retain their existing skip behavior, and author display names remain optional enrichment.

## Related Issue

MUL-7268.

## Type of Change

- [x] Bug fix
- [x] Tests
- [x] Documentation update

## Changes Made

- Propagate required comment and reply-thread read errors from the shared claim builder.
- Add database-backed fault/recovery coverage for both claim endpoints and all three input-read paths, plus missing-input, optional-author, and tombstone boundaries.
- Document the existing redelivery behavior in the built-in mention reference.
- Keep the comment-error return aligned with the builder's six-value contract, including `issueSnapshot`. Both callers handle the failure before claim finalization, so failed reads still commit no receipt or task token.

## How to Test

Validation on 2026-09-22 uses PR head `d42fa589b` plus the one-line return-contract correction. The previous head reproduced `not enough return values` in `go build ./cmd/server`; the correction restores compilation.

Passed locally:

- `go build ./...` from `server/`.
- `go vet ./internal/handler` from `server/`.
- All 47 top-level tests and their subtests selected from `claim_comment_load_test.go`, `daemon_comment_delivery_test.go`, `daemon_batch_claim_finalize_test.go`, `daemon_issue_snapshot_test.go`, and `daemon_comment_workspace_scope_test.go`, with `-race -count=1 -timeout=3m`. No selected tests were skipped.
- `git diff --check`.

The regression run covers all six single/batch × trigger/coalesced/thread read failures and subsequent redelivery, missing inputs, optional author lookups, deleted comments, receipt/token finalization, issue snapshots, and workspace isolation. It used the repository's agent CLI guard and a new isolated worktree PostgreSQL database, initialized using the repository environment and migration scripts through migration 534. No real agent CLI was invoked. Optional migrations retain the runner's environment gates.

The full backend suite and Windows runtime checks were not rerun locally for this one-line correction; the updated PR triggers CI for those checks. Earlier CI runs do not validate the new head.

## Risks

A transient required-input read error delays execution until the existing dispatched-task recovery window and prepare lease allow redelivery. There are no schema migrations or wire-format changes. The change does not add a durable dispatch outbox or alter fail/cancel reconciliation.

## Checklist

- [x] Included the triggering scenario and resulting behavior
- [x] Added regression tests and documented the checks run and their limits
- [x] Updated the relevant built-in skill reference
- [x] Considered recovery latency and compatibility
- UI screenshots, localization, and runtime landing copy: not applicable

## AI Disclosure

**AI tool used:** Codex

**Approach:** Traced comment loading and both claim endpoints, reproduced the input-read failure with PostgreSQL fault injection, and verified lease-based recovery. For the return-contract correction, reproduced the compiler error on the current PR head, checked all builder returns and both callers, and reran compilation, vet, and the existing claim/delivery/snapshot regressions with independent code review.


ADanMan:
Nice fix — this closes a real correctness gap. Starting an agent run after silently dropping a coalesced handoff comment (a transient read error looking the same as a deleted comment) means the agent runs without the sender's constraints, and failing closed onto lease redelivery is the right call.

What I like:
- The core distinction is exactly right: `pgx.ErrNoRows` stays a skip (genuinely missing / wrong-workspace / tombstoned input), while every other read error rejects the whole claim. Tombstone handling (`DeletedAt.Valid` → skip) is correctly kept as a separate case rather than folded into the error path.
- Clean failure path: `buildClaimedTaskResponse` returns a `claimBuildFailure{outcome:"error_comment_load", …}` consistent with the existing pattern, and the test proves the task stays `dispatched` with no delivery receipt and no task token minted, then the normal endpoint reclaims and delivers the full handoff after lease expiry. That end-to-end reclaim assertion is the convincing part.
- Thorough test matrix — single vs batch × trigger/coalesced/thread, plus the tombstone-keeps-replies and optional-author cases. The DBTX wrapper keyed on the exact sqlc query name + arg is a clean, targeted fault-injection technique.

One non-blocking note worth a line in the PR: the author-name lookup (`GetUser`) error is still swallowed — the comment stays deliverable without a display name — while comment/thread read errors now reject. That's defensible (a display name isn't a constraint), but it's the one place a transient DB error still lets a run proceed with degraded data, so it's worth stating that it's intentional.

CI is green on `d328459e` (backend / backend-tests / backend-agent-tests all pass), so the earlier red run looks resolved. Reads solid.

PR #8288, opened 2026-09-11
MUL-7272: fix(comments): preserve deferred worker reply handoffs

importcpp:
## What does this PR do?

A worker's progress reply can wake its squad leader, but a later ordinary reply in the same thread is lost as an execution input if the leader has already claimed that run. Creation cannot merge into the dispatched run and defers by timestamp; completion then discards the implicit agent route. The comment remains visible, but the leader never receives another run for it.

Record accepted worker replies in the existing planned-input set when a pending run can no longer be merged. Completion replays an undelivered worker reply only when that run already carries its input obligation, after recomputing routing and permissions. Timestamp-only ordinary agent replies remain excluded. This change is independent of the stage-notification work in #8286.

## Changes Made

- Mark the existing non-self agent-to-assigned-squad-leader route internally, and use planned-input registration when its queued merge misses. Explicit mentions and member routing retain their existing rules.
- Permit completion replay of planned worker inputs, preserving the claim receipt, completing-agent scope, human delegation provenance, and serialized execution.
- Recheck active status on the registration UPDATE itself: if completion wins while the query waits for a row lock, return a miss and let the enqueue path create a successor. A subquery-only status check could append an obligation to an already completed run.
- Add PostgreSQL-backed delivery, replay-boundary, and deterministic row-lock regressions; update the built-in squad reference and query documentation.

## Validation

- On unchanged main `b36906e6d`, the six-case queued/dispatched/running × plain/explicit-squad reply matrix fails only for dispatched + plain reply: no successor receives the result. A separate deterministic row-lock probe also reproduces registration incorrectly succeeding after completion commits.
- After the fix, all 16 scenarios in `TestWorkerReplyDelivery`, `TestWorkerReplyReconcileBoundaries`, and `TestWorkerReplyRegistrationLosesCompletionRace` pass under `-race`. They drive actual comment creation and claim/start/completion, including multiple results, suppressed and unplanned inputs, self-replies, notes, archival/reassignment, registration failure, and completion winning before registration.
- The full `go test -race ./internal/handler -count=1` run fails only `TestClaimTasksByRuntime_ClaimPollHintSchedulesNextDeferredTask` and `TestReportTaskMessagesFallsBackWholeBatchForClockSkew`. Both fail again with all modified production Go files overlaid from main; the database clock is ahead of the host. With only those two baseline cases excluded via `-skip`, the entire handler suite passes (71.8 s).
- `go vet ./internal/handler` and `git diff --check` pass; SQL code was regenerated with the repository's pinned sqlc version. Tests use an isolated PostgreSQL 17 / pgvector database and the agent CLI guard.

## Risks and scope

This deliberately permits a follow-up for an accepted, planned worker input that the previous run did not receive. It does not authorize replay of arbitrary ordinary agent comments. Suppressed triggers, notes, self-replies, current routing, and invocation permissions still apply. The tests check actual claim payloads and receipts, delegation provenance, coalescing, and the absence of an extra run after the successor completes.

This uses the existing task input plan, with no schema or API changes. It does not add a general notification outbox or recover failed registration writes, failed completion reconciliation, or historical replies that never acquired a planned obligation. Stage notification generation and replay remain separate work.

## AI Disclosure

**AI tool used:** Codex.

**Approach:** Reproduced the missing worker handoff on main with real PostgreSQL and handler/claim/start/completion calls, narrowed replay to accepted input obligations, then exercised negative routing cases and a concurrent completion/registration schedule. No real agent CLI was invoked.

### DIFF
--- a/server/internal/handler/comment.go
+++ b/server/internal/handler/comment.go
@@ -1504,9 +1504,11 @@ type commentAgentTrigger struct {
 	Source         commentAgentTriggerSource
 	Squad          *db.Squad
 	AlreadyPending bool
-	// WorkerReply is the non-self agent reply routed to the assigned squad's
-	// leader. Completion may replay it only if creation recorded a planned input.
-	WorkerReply bool
+	// NonLeaderAgentReply marks an agent comment, not authored by the leader,
+	// that the assigned-squad-leader fallback routes to that leader. The author
+	// need not be a squad member. Completion may replay it only if creation
+	// recorded it as a planned input.
+	NonLeaderAgentReply bool
 }
 
 type commentTriggerComputeOptions struct {
@@ -2097,7 +2099,7 @@ func (h *Handler) resolveCommentTriggerEnqueue(ctx context.Context, issue db.Iss
 			}
 			// No same-head QUEUED row to fold into (merge missed). The two paths
 			// resolve differently.
-			if !lostRace && !trigger.WorkerReply {
+			if !lostRace && !trigger.NonLeaderAgentReply {
 				// (b) AlreadyPending path: this comment arrived AFTER its task, so
 				// it is newer than the task and completion reconcile covers it by
 				// timestamp; MUL-4195 leaves the claimed task untouched. Defer to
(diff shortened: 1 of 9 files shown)


Bohan-J (approved):
Thanks @importcpp, this is a clean fix with a very thorough reproduction. Reusing the existing planned-input path and only replaying replies that creation actually recorded closes the dispatched-window gap without letting timestamp-only agent replies start new runs. The status recheck on the registration `UPDATE` also closes the same completion race on the existing lost-race path.

I pushed one small follow-up commit before merging:

- Renamed `WorkerReply` to `NonLeaderAgentReply`. The assigned-squad-leader fallback doesn't require the author to be a squad member, so the old name was a bit narrower than what it covers.
- Trimmed the new `squads.md` paragraph down to the behavior agents rely on.

Merging now. Thanks again for the contribution!

PR #8297, opened 2026-09-11
MUL-7275: fix(skills): normalize Windows archive entry paths

### LINKED ISSUE #8292: [Bug]: Skill archive import stores Windows backslash paths verbatim, so every run of the agent fails with skill_bundle_unavailable
### Deployment type

Official App

### Area

Agents and runtimes

### What happened?

A skill imported from a `.zip` made on Windows stores its supporting-file paths with backslashes (for example, `agents\openai.yaml`). The server accepts that path. The daemon then rejects the skill bundle, so every run of any agent that has the skill attached fails before it starts:

```
skill bundle unavailable: resolve skill bundle returned invalid bundle: skill_id=<id> source=workspace hash=sha256:<hash>
failure_reason: skill_bundle_unavailable
```

Expected: the archive importer should normalize `\` to `/` in entry names, or reject the archive with a clear error. It should not store a path that the daemon will always reject.

**Why it happens (main @ 885ef12)**

- The ZIP spec (APPNOTE 4.4.17.1) requires forward slashes. Windows PowerShell 5.1 `Compress-Archive` writes backslashes anyway, e.g. `agents\openai.yaml`.
- `parseSkillArchive` runs `path.Clean(f.Name)` on the raw entry name ([skill_import_archive.go#L152](https://github.com/multica-ai/multica/blob/885ef12232bf191753521a6ea617b12fc33fcc05/server/internal/handler/skill_import_archive.go#L152)). `path` treats `\` as an ordinary character.
- `validateFilePath` uses `filepath` and only rejects absolute paths and `..` ([skill.go#L304](https://github.com/multica-ai/multica/blob/885ef12232bf191753521a6ea617b12fc33fcc05/server/internal/handler/skill.go#L304)). On a Linux server `agents\openai.yaml` passes, and it is stored as-is.
- `safeSkillFilePath` in the daemon rejects any path containing `\` ([skill_cache.go#L157-L166](https://github.com/multica-ai/multica/blob/885ef12232bf191753521a6ea617b12fc33fcc05/server/internal/daemon/skill_cache.go#L157-L166)). The whole bundle is invalid, and the run fails.

**Relation to existing work**

- #6893 reports the same symptom through the runtime-local import flow. This report covers a different entry point: archive upload (`multica skill import --file`).
- #6899 only touches `runtime_local_skills.go`, so it does not cover archive import.
- #7743 changes `validateFilePath` to the shared strict contract. For archive import, that turns the corrupt write into a silent drop: the `continue` at [L171-L173](https://github.com/multica-ai/multica/blob/885ef12232bf191753521a6ea617b12fc33fcc05/server/internal/handler/skill_import_archive.go#L171-L173) skips backslash entries without any warning. A wrapper-layout archive, where every entry uses `\`, already fails with `archive does not contain a SKILL.md`, because `path.Base("my-skill\SKILL.md")` is not `SKILL.md`. That message doesn't help the user find the real problem.

A possible fix is to apply `strings.ReplaceAll(f.Name, "\\", "/")` before `path.Clean` in both loops of `parseSkillArchive`.

### Steps to reproduce

1. On Windows, in Windows PowerShell 5.1, create a skill folder that contains `SKILL.md` and a supporting file in a subfolder, e.g. `agents/openai.yaml`.
2. `Compress-Archive -Path '.\my-skill\*' -DestinationPath '.\my-skill.zip'`
3. `multica skill import --file .\my-skill.zip`. The import succeeds.
4. `multica skill get <skill-id>` lists the file path as `agents\openai.yaml`.
5. Attach the skill to an agent and mention the agent in an issue comment. The run fails immediately with `skill_bundle_unavailable`, and so does every retry and every later run of that agent.

Variant: with `-LiteralPath '.\my-skill'` (wrapper folder layout), step 3 fails with `archive does not contain a SKILL.md` even though the archive contains `my-skill\SKILL.md`.

### Additional context

```shell
# Client that ran the import
OS: Windows
Multica CLI / daemon: v0.4.42 (launched by Desktop)
Agent runtime: codex-cli 0.154.0
Server: multica.ai (Official App)

# Entries in the uploaded archive
agents\openai.yaml
SKILL.md

# Workaround: delete the bad file, or re-upload it with a forward-slash path
multica skill files delete <skill-id> <file-id>
multica skill files upsert <skill-id> --path agents/openai.yaml --content-file ./openai.yaml
```

Two side effects made this harder to diagnose:

- The error doesn't say which check failed, only "invalid bundle".
- The UI text for `skill_bundle_unavailable` says "This is usually a connection problem — please try again", and the task is retried automatically. The failure is deterministic, so retrying can't help.

Dante-dan:
## What does this PR do?

Windows PowerShell 5.1 can produce ZIP entries with backslash separators even though the ZIP format specifies forward slashes. The archive importer previously passed those raw names to `path.Clean`, which treats backslashes as ordinary characters on the server. Root-layout archives therefore persisted daemon-invalid supporting-file paths, while wrapper-layout archives could fail to find their existing `SKILL.md`.

Both passes through `parseSkillArchive` now translate archive-entry backslashes to `/` before applying `path.Clean`. The existing validation remains after cleaning, so traversal and absolute paths encoded with either separator are still rejected; the daemon's strict bundle validator is unchanged.

## Related Issue

Fixes #8292

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- Normalize non-standard Windows separators before cleaning ZIP entry names in both archive parsing loops.
- Add regressions for root and wrapper archive layouts with backslash entry names.
- Add regressions proving backslash-encoded traversal and absolute paths remain excluded.
- Skip trailing-backslash directory markers, including entries without directory attributes produced by PowerShell `Compress-Archive`, in both parsing passes.
- Reject the entire import when entries resolve to the same path. This replaces nondeterministic duplicate-file selection with a clear error.
- Drop drive-prefixed and NUL-containing paths before persistence so imported paths remain compatible with the daemon.

## How to Test

1. All `TestParseSkillArchive` tests — passed, including root/wrapper Windows directory markers without directory attributes, nested supporting files, path collisions and unsafe paths. The pure parser suite used a Go overlay to bypass the package’s database-only `TestMain`; archives were synthesized locally with `zip.Writer.Create`, not generated on a Windows machine.
2. `go vet ./internal/handler` — passed.
3. `go build ./internal/handler` — passed.
4. `gofmt` and `git diff --check` — passed.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (not applicable; backend-only)
- [x] I have updated relevant documentation to reflect my changes (not applicable; no documented interface changed)

## AI Disclosure

**Prompt / approach:**
I used Codex to trace both archive parsing passes and the daemon/import path contract, compare the archive-upload scope with the related runtime-local and direct-authoring fixes, implement normalization at the archive trust boundary, and add focused regression coverage for Windows layouts and path-safety cases.

## Screenshots (optional)

### DIFF
--- a/server/internal/handler/skill_import_archive.go
+++ b/server/internal/handler/skill_import_archive.go
@@ -105,6 +105,7 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 	// primary content — keeping every accepted entry zip-slip-safe.
 	var skillMd *zip.File
 	rootPrefix := ""
+	skillMdEntries := make(map[string]string)
 	for _, f := range zr.File {
 		if f.FileInfo().IsDir() {
 			continue
@@ -113,9 +114,13 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 		if !strings.EqualFold(path.Base(clean), skillpkg.ContentFilename) {
 			continue
 		}
-		if !validateFilePath(clean) {
+		if !validateArchiveFilePath(clean) {
 			continue
 		}
+		if previous, exists := skillMdEntries[clean]; exists {
+			return nil, fmt.Errorf("archive entries %q and %q resolve to the same path %q", previous, f.Name, clean)
+		}
+		skillMdEntries[clean] = f.Name
 		prefix := archiveEntryPrefix(clean)
 		if skillMd == nil || len(prefix) < len(rootPrefix) {
 			skillMd = f
@@ -145,6 +150,7 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 		content:     content,
 	}
 
+	seenFiles := make(map[string]string)
 	for _, f := range zr.File {
 		if f.FileInfo().IsDir() {
 			continue
@@ -168,9 +174,13 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 			continue
 		}
 		// zip-slip / absolute-path guard.
-		if !validateFilePath(rel) {
+		if !validateArchiveFilePath(rel) {
 			continue
 		}
+		if previous, exists := seenFiles[rel]; exists {
+			return nil, fmt.Errorf("archive entries %q and %q resolve to the same path %q", previous, f.Name, rel)
+		}
+		seenFiles[rel] = f.Name
 		fileContent, ferr := readZipFile(f, maxImportFileSize)
 		if ferr != nil {
 			// An oversize or unreadable individual asset is skipped rather than
@@ -196,6 +206,24 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 // absolute and traversal entries cannot bypass the zip-slip guards.
 func cleanArchiveEntryName(name string) string {
 	return path.Clean(strings.ReplaceAll(name, "\\", "/"))
+}
+
+// validateArchiveFilePath extends the existing import validation with portable
+// archive constraints. Windows drive syntax is not recognized by filepath.IsAbs
+// on a Unix server, and NUL bytes would be removed before persistence. Neither
+// can safely be passed to the daemon as a bundle path.
+func validateArchiveFilePath(p string) bool {
+	if strings.ContainsRune(p, '\x00') {
+		return false
+	}
+	if len(p) >= 2 && isASCIIAlpha(p[0]) && p[1] == ':' {
+		return false
+	}
+	return validateFilePath(p)
+}
+
+func isASCIIAlpha(c byte) bool {
+	return c >= 'a' && c <= 'z' || c >= 'A' && c <= 'Z'
 }
 
 // archiveEntryPrefix returns the directory prefix (with trailing slash) of a
--- a/server/internal/handler/skill_import_archive.go
+++ b/server/internal/handler/skill_import_archive.go
@@ -107,7 +107,8 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 	rootPrefix := ""
 	skillMdEntries := make(map[string]string)
 	for _, f := range zr.File {
-		if f.FileInfo().IsDir() {
+		// Compress-Archive directory markers may have no directory attributes.
+		if f.FileInfo().IsDir() || strings.HasSuffix(f.Name, `\`) {
 			continue
 		}
 		clean := cleanArchiveEntryName(f.Name)
@@ -152,7 +153,8 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 
 	seenFiles := make(map[string]string)
 	for _, f := range zr.File {
-		if f.FileInfo().IsDir() {
+		// Compress-Archive directory markers may have no directory attributes.
+		if f.FileInfo().IsDir() || strings.HasSuffix(f.Name, `\`) {
 			continue
 		}
 		clean := cleanArchiveEntryName(f.Name)
--- a/server/internal/handler/skill_import_archive.go
+++ b/server/internal/handler/skill_import_archive.go
@@ -109,7 +109,7 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 		if f.FileInfo().IsDir() {
 			continue
 		}
-		clean := path.Clean(f.Name)
+		clean := cleanArchiveEntryName(f.Name)
 		if !strings.EqualFold(path.Base(clean), skillpkg.ContentFilename) {
 			continue
 		}
@@ -149,7 +149,7 @@ func parseSkillArchive(data []byte, filename string) (*importedSkill, error) {
 		if f.FileInfo().IsDir() {
 			continue
 		}
-		clean := path.Clean(f.Name)
+		clean := cleanArchiveEntryName(f.Name)
 		// Only files under the resolved skill root belong to this skill.
 		if rootPrefix != "" && !strings.HasPrefix(clean, rootPrefix) {
 			continue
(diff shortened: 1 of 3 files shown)


Bohan-J (changes requested):
Thanks for picking this up. Normalizing at the archive boundary is the right fix, and it matches what the web/desktop folder packer already does (`packages/core/skills/pack-archive.ts`). I checked it locally: root-layout and wrapper-layout archives with `\` entries now import through `POST /api/skills/import` as `agents/openai.yaml`, and every stored path passes the daemon's `safeSkillFilePath` rule. I also ran a randomized check over ~480k generated entry names: no accepted path violates the daemon rule, and none are duplicated.

There is one blocking gap and one description update.

### 1. Blocking: `Compress-Archive` directory entries become empty files

Windows PowerShell's `Compress-Archive` writes an explicit `dir\` entry for every directory that has no files *directly* inside it. That covers empty directories and also directories that only contain subdirectories. In [`Microsoft.PowerShell.Archive.psm1` v1.0.0.0, L686–L693](https://github.com/PowerShell/Microsoft.PowerShell.Archive/blob/v1.0.0.0/Microsoft.PowerShell.Archive/Microsoft.PowerShell.Archive.psm1#L686-L693), `FullName + "\"` is added when `GetFiles()` is empty, and [L781–L849](https://github.com/PowerShell/Microsoft.PowerShell.Archive/blob/v1.0.0.0/Microsoft.PowerShell.Archive/Microsoft.PowerShell.Archive.psm1#L781-L849) writes it with `ZipArchive.CreateEntry`. The entry ends in `\` rather than `/` and carries no directory attribute, so `f.FileInfo().IsDir()` returns false.

With this PR, a skill laid out like this:

```
SKILL.md
references/
  guides/
    setup.md
```

produces the entries `SKILL.md`, `references\` and `references\guides\setup.md`. The `references\` entry is normalized to `references` and imported as an empty *file* alongside `references/guides/setup.md`. The import succeeds (201), but the daemon cannot write the bundle to disk. `writeSkillFiles` (`server/internal/daemon/execenv/context.go`) fails with `mkdir .../references: not a directory`, or with `is a directory` if the files are written in the other order. So every run of an agent using this skill still fails before it starts, which is the same symptom as #8292. An empty `assets\` directory likewise becomes an empty file named `assets`.

Suggested fix: treat a trailing backslash as a directory marker in both loops:

```go
if f.FileInfo().IsDir() || strings.HasSuffix(f.Name, `\`) {
	continue
}
```

Please also add a regression test with a zero-byte `references\` entry written via `zip.Writer.Create`, which has no directory attribute, just like the entries `Compress-Archive` writes. I tried this change locally: the layout above then imports only `references/guides/setup.md`, and all tests in this PR still pass. My reproduction used a synthesized archive, not a real Windows machine.

### 2. Please document the collision rule in the PR description

The second commit adds behavior that isn't listed under "Changes Made":

- Entries that resolve to the same path now fail the whole import.
- Drive-prefixed and NUL-containing entries are dropped.

The collision rule is visible to users. An archive that contains the same entry name twice used to import: one of the two files won, and which one was not deterministic. It now returns 400. I agree with rejecting it; please just call the change out in the PR description and give the reason.

### Non-blocking

- The collision error uses `%q`, so the web dialog and the CLI show `"agents\\openai.yaml"` with doubled backslashes. Plain quoting (`"%s"`) would read better.
- The collision check runs on the raw entry name, but paths go through `sanitizeNullBytes` before they are persisted (`skill_create.go`), which replaces invalid UTF-8 with U+FFFD. Two different non-UTF-8 names such as `a\xff.md` and `a\xfe.md` therefore both become `a�.md`: the response lists 2 files while only 1 row is stored. This predates the PR, but keying `seenFiles` by `sanitizeNullBytes(rel)` would close it. A follow-up is fine too.
- `validateArchiveFilePath` overlaps with the shared path rule proposed in #7743. I test-merged the two PRs with current `main`: there were no conflicts and the tests pass, so the two can be consolidated once both have landed.

Dante-dan:
Fixed in 22de05278. Both parsing passes now skip entries ending in a backslash even when the ZIP header has no directory attribute. The regression uses zero-byte entries from `zip.Writer.Create` and covers root and wrapper layouts, an empty directory, a directory containing only nested files, and a directory named `SKILL.md`. Only the nested supporting file is imported.

All archive parser tests, `go vet ./internal/handler`, `go build ./internal/handler`, and `git diff --check` passed. These are synthesized-archive tests; I did not run PowerShell on Windows.

I also updated the description to explain that canonical-path collisions fail the entire import to avoid nondeterministic overwrites, and that drive-prefixed and NUL-containing paths are dropped. The pre-existing invalid-UTF-8 persistence collision and shared-validator consolidation remain separate follow-ups as you suggested.

Bohan-J (approved):
Thanks for the quick turnaround, and for a well-scoped fix — this is a clean first contribution.

The directory-marker gap is resolved, and the new regression is a good one: it asserts the fixture carries no directory attributes, so it fails if the premise ever stops holding, and it covers `SKILL.md\` being skipped as a primary-content candidate. The PR description now documents the collision rejection and the drive-prefix / NUL drops.

Verified locally on `22de0527` (macOS, Postgres-backed handler tests):

- All `TestParseSkillArchive*` and `TestImportSkill_Archive*` tests pass, plus `go vet ./internal/handler`.
- End to end through `POST /api/skills/import` with a PowerShell-shaped archive (`SKILL.md`, `references\`, `references\guides\setup.md`, `assets\`, `agents\openai.yaml`): the skill stores exactly `agents/openai.yaml` and `references/guides/setup.md`, with the directory markers dropped.
- Feeding those two paths to the daemon's `writeSkillFiles` now succeeds, where it previously failed with `not a directory`.

The remaining non-blocking items from the last round — the `%q` escaping in the collision error, the non-UTF-8 name collision that predates this PR, and consolidating `validateArchiveFilePath` with #7743 — are fine as follow-ups.

Note that this fixes the importer going forward; skills already stored with backslash paths still need to be re-uploaded or corrected.

Approving and merging. Thanks again!

PR #8314, opened 2026-09-11
MUL-7288: fix(server): wake guest squad leaders on worker replies

### LINKED ISSUE #8301: [Bug]: Guest squad worker comments do not reawaken the leader on member-assigned issues
### Deployment type

self-host

### Area

Agents and runtimes

### What happened?

A squad can participate in an issue as a guest when the issue remains assigned to a human member and the squad is invoked through an `@squad` comment.

In this mode, the squad leader is correctly started with `is_leader_task=true` and receives the Squad Operating Protocol. The leader then delegates work to a squad member using an `@agent` mention.

When that worker finishes and posts a normal progress/result comment without an explicit mention, the leader is not reawakened. No new task is created, and the issue remains stalled until a human manually mentions the squad again.

This contradicts the leader briefing, which says:

> You will be re-triggered automatically when:
>
> - a delegated member posts an update or asks you a question;
> - a delegated member finishes and the issue moves forward.

The behavior works when the issue itself is assigned to the squad, but not when the squad participates as a guest on an issue assigned to a member.

### Expected behavior

A plain result/progress comment from a worker delegated by a guest squad leader should enqueue exactly one new task for that same leader:

- `is_leader_task = true`
- `squad_id` set to the guest squad
- Squad Operating Protocol and roster injected
- existing permission, attribution, self-trigger, and pending-task deduplication rules preserved

### Actual behavior

The worker comment is persisted but produces no trigger and no task. The leader→worker→leader coordination loop stops after every delegation round.

### Root cause

In the current `computeCommentAgentTriggers` implementation, agent-authored comments retain only the assigned-squad fallback:

https://github.com/multica-ai/multica/blob/676245af632db866dcba2afb47f5e7d99abf53b5/server/internal/handler/comment.go#L2619-L2636

The fallback is guarded by:

```go
if issue.AssigneeType.Valid && issue.AssigneeType.String == "squad" {
    if trigger, ok := h.routeAssignedSquadLeaderFallback(...); ok {
        return []commentAgentTrigger{trigger}, nil
    }
}
return nil, nil
```

For a guest squad, `issue.AssigneeType == "member"`, so the worker comment reaches the final `return nil, nil`. Parent-author and conversation routing are not considered because the agent-authored branch returns first.

The existing worker-comment regression coverage explicitly constructs only squad-assigned issues:

https://github.com/multica-ai/multica/blob/676245af632db866dcba2afb47f5e7d99abf53b5/server/internal/handler/squad_comment_trigger_test.go#L226-L273

MUL-7272 / PR #8288 does not fix this case. It preserves delivery of worker replies already accepted by the assigned-squad fallback, but does not extend routing to guest squads:

https://github.com/multica-ai/multica/pull/8288

### Suggested direction

Resolve the guest leader from trusted task/comment provenance rather than from the issue assignee.

The existing `squadLeaderRoleOfAuthoringTask` helper introduced by MUL-7006 may be reusable when the worker is replying to a delegation comment authored by a leader-role task:

https://github.com/multica-ai/multica/blob/676245af632db866dcba2afb47f5e7d99abf53b5/server/internal/handler/comment.go#L2712-L2760

The route should fail closed if provenance is missing, the squad is archived, or the leader has changed. It should not infer a squad merely from membership because one agent may belong to multiple squads.

### Steps to reproduce

1. Create an issue and assign it to a human member, not to a squad.
2. Add a comment containing `[@Squad](mention://squad/<squad-id>)`.
3. Confirm that the squad leader receives a task with `is_leader_task=true`.
4. Have the leader delegate work by posting a comment that explicitly mentions a worker agent.
5. Have the worker complete the task and post a plain progress/result comment without any `@agent` or `@squad` mention.
6. Inspect the issue runs or task queue.
7. Observe that no new leader task is created.
8. Post another human comment explicitly mentioning the squad.
9. Observe that the leader starts again, confirming that the leader and runtime are otherwise available.

### Screenshots (optional)

_No response_

```shell

```

bojackli:
## Summary

Fix the leader → worker → leader handoff when a squad participates as a guest through an `@squad` mention instead of owning the issue.

A worker's plain reply to the guest leader's delegation now wakes the exact squad leader without requiring another explicit mention.

Fixes #8301.

## Problem

Agent-authored comments previously retained only the assigned-squad fallback. On a member-assigned issue, a guest squad worker's progress comment therefore produced no trigger and silently stalled the coordination loop.

The existing fallback was also insufficient when an issue was assigned to squad A while guest squad B delegated work: blindly routing to the assigned squad could wake the wrong leader.

## Solution

- Resolve guest squad handoffs from server-trusted task lineage:
  - the worker's authoring task must match the comment author and issue;
  - the task must cover the replied-to delegation comment;
  - the parent comment's source task must prove the squad leader role.
- Prefer this exact guest delegation route before the assigned-squad fallback.
- Fail closed when the guest squad is archived, its leader changed, or invocation permission is no longer valid.
- Preserve the existing assigned-squad behavior when the delegation belongs to the currently assigned squad.
- Carry the authoring task through preview, cancellation recovery, and completion reconciliation.
- Allow planned guest worker replies to replay after a dispatched leader task completes.
- Update the built-in squad behavior documentation.

## Tests

Added regression coverage for:

- member-assigned issues with guest squad delegation;
- immediate wake-up when the leader is idle;
- dispatched leader tasks with planned reply registration and completion replay;
- assigned squad A versus guest squad B routing;
- archived guest squads;
- changed guest squad leaders;
- revoked invocation permission;
- rejection of timestamp-only, unplanned replay.

Local verification:

- `go vet ./internal/handler`
- targeted handler test package compilation
- `git diff --check`

The local sandbox could not connect to PostgreSQL, so the database-backed assertions are expected to run in GitHub Actions.


multica-eve (changes requested):
Request changes: the core guest-squad handoff fix is directionally correct, but a deleted delegation tombstone can still wake the guest leader on current `main`.

`computeCommentAgentTriggers` calls `routeGuestSquadLeaderFallback` in the agent-authored branch before the existing deleted-parent guard, and the new helper does not reject `parent.DeletedAt`. A concrete failure sequence is:

1. A worker reply is registered as a planned, undelivered input on a dispatched leader task.
2. The original leader delegation is deleted and becomes a tombstone.
3. Completion reconciliation recomputes the worker reply route and enqueues another leader task through that tombstone.

This makes deleted content participate in agent routing, conflicting with the established #8323 contract that tombstones only preserve the reply tree and are never agent input.

Please treat a tombstoned parent as handled with no trigger (and do not fall through to the assigned-squad fallback), then add a completion-reconciliation regression proving that deleting the delegation prevents a successor leader task.

The primary idle/busy guest handoff, exact guest-vs-assigned-squad routing, permission/archival/leader-change cases, existing assigned-squad behavior, race-enabled targeted tests, `go vet ./internal/handler`, the runtime-brief `ByteIdentical` test, and `git diff --check` otherwise pass against current `main`.

bojackli:
Thanks for the review. Addressed in `3e4791a5e`:

- rebased the branch onto the latest `main`;
- treat a tombstoned delegation parent as handled with no trigger, preventing fallback to the assigned squad;
- added a completion-reconciliation regression covering a planned worker reply whose delegation is deleted before the leader run completes.

Could you please take another look when the checks finish?

multica-eve (changes requested):
Request changes: the guest-squad handoff is directionally correct, but the new tombstone guard regresses the existing assigned-squad worker-to-leader loop.

`computeCommentAgentTriggers` calls `routeGuestSquadLeaderFallback` before `routeAssignedSquadLeaderFallback`. In the guest helper, `parent.DeletedAt.Valid` currently returns `(nil, true)` before validating the authoring task, guest lineage, or whether the delegation belongs to the assigned squad. Therefore any agent reply whose parent is a tombstone is treated as fully handled and can never reach the assigned-squad fallback, even though that fallback derives its leader from `issue.assignee_id` and does not rely on the deleted parent as routing authority.

A concrete regression is an issue assigned to squad A where a worker posts a result under a comment that was deleted while the worker was running: current `main` still wakes A's leader, while this head queues no leader task. The same suppression occurs when the tombstone was authored by a member and has no guest-squad lineage at all.

Please change the tombstone branch to `return nil, false`, so it rejects guest routing without consuming the independent assigned-squad path. The existing `deleted_delegation` guest regression remains valid because its issue is member-assigned and has no assigned-squad fallback. Please also add a regression proving that, on a squad-assigned issue, a worker reply under a tombstone still wakes the assigned leader.

bojackli:
Thanks for clarifying the distinction between guest delegation routing and the issue's independent squad assignment.

Addressed in `d0a625212`:

- A tombstoned parent now rejects guest routing without consuming the assigned-squad fallback.
- Added regressions for worker replies under both agent-authored and member-authored tombstones on squad-assigned issues, plus the assigned-A / guest-B case.
- Retained the guest completion-reconciliation regression: deleting the delegation on a member-assigned issue prevents a successor leader task.
- Preserved the existing permission, triage, self-trigger, and replay-target checks.

The branch is rebased onto current main. All CI checks pass, including PostgreSQL-backed race tests and vet. An independent review found no further issues.

PR #8317, opened 2026-09-11
MUL-7287: fix(agent): guard process-group signals against pid reuse

### LINKED ISSUE #8306: daemon silently killed by its own SIGKILL to a related PID within seconds/minutes of a clean start, when run under a process supervisor (no TTY, shared process group)
## Summary

`multica daemon start` (both with and without `--foreground`) starts cleanly, authenticates, detects all agent CLIs, and successfully registers 4 runtimes against the workspace — then dies silently (no shutdown/error log line at all) anywhere from ~10s to ~90s later. Under a process supervisor that respawns on exit (s6-overlay in this case), this repeats forever: healthy run → silent death → respawn → repeat.

Traced with `strace -f -p <daemon-pid> -e trace=signal`, the daemon's own PID is seen issuing a `kill(SIGKILL)` against a *different, related* PID right after receiving a `SIGCHLD` from a subprocess exiting cleanly:

```
1513  02:40:21.639367 --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=1524, si_uid=0, si_status=0} ---
1513  02:40:21.639494 rt_sigreturn({mask=[]}) = 1524
1513  02:40:21.639709 kill(1523, SIGKILL
```

(1513 = the daemon's own PID at the time; 1524 = a subprocess that just exited normally with status 0; 1523 = a *different* PID that gets SIGKILLed immediately after — the daemon process itself disappears at this point.)

An earlier trace of a different run caught what looks like the same mechanism at the process-group level:

```
11719 02:25:37.914313 kill(-11725, SIGKILL) = -1 ESRCH (No such process)
```

(negative PID = process-group kill; ESRCH here just means that particular target had already exited — but the pattern of "kill a PID/PGID right after a sibling subprocess's SIGCHLD" is the same.)

This strongly suggests the agent-version-detection code (which shells out to each configured agent CLI — `claude --version`, `codex --version`, `copilot ...`, `agy --version` — with some kind of timeout) is, on at least some of those subprocess completions, computing the wrong PID/PGID to signal — plausibly a race between `os/exec`'s `Setpgid`/child-tracking and the timeout-kill path, where the "kill this subprocess (or its process group)" logic ends up targeting the daemon's own PID or a PID/PGID it shares with the daemon, instead of only the intended subprocess. When the daemon isn't isolated into its own session (no `setsid`), this is fatal to the daemon itself.

## Supporting log evidence

Before the silent kill, we did see two adjacent log lines from a related run that point at the same subsystem (agent version probing + immediately-following registration HTTP call both getting torn down by a signal in the same window):

```
DBG agent version probe failed; retrying component=daemon name=copilot attempt=1 error="detect version for /usr/lib/node_modules/@github/copilot/npm-loader.js: signal: killed"
ERR failed to register runtimes component=daemon workspace_id=... error="register runtimes: Post \"http://.../api/daemon/register\": terminated signal received"
```

Both errors ("signal: killed" for the copilot subprocess check, "terminated signal received" for the concurrent HTTP POST) happened at the exact same timestamp, consistent with something delivering a signal broadly enough to hit both the subprocess-check goroutine and the HTTP-registration goroutine at once.

## Environment

- Multica CLI/daemon version: **0.4.42**
- OS: Debian 12 (bookworm-slim) container, `linux/amd64`
- Agents detected: `claude`, `codex`, `copilot`, `antigravity` (all 4 register successfully every time, before the crash)
- Self-hosted server (own Helm-deployed instance), reached over an in-cluster Kubernetes Service — not multica.ai
- Daemon is launched by an s6-overlay v3 "legacy longrun" service (no TTY, no `setsid` — the daemon process shares its process group with the service script that execs it)
- Reproduced identically with `multica daemon start --foreground` and with the documented default (backgrounded) `multica daemon start`
- Reproduced on a freshly recreated pod/container with zero other processes or manual interference — not specific to a particular machine state
- Ruled out: OOM (cgroup `memory.events` shows `oom_kill 0`), PID limits (`pids.current` far below `pids.max`), port conflicts (confirmed clean via `ss -tlnp`), container/pod restarts (0 restarts throughout), and any external supervisor-side kill signal (s6 itself was not observed sending anything — the kill originates from the daemon's own process, per the trace)

## Steps to reproduce

1. `multica login --token <PAT>` (or browser login) against a self-hosted instance
2. `multica daemon start` (or `--foreground`) in a container/session where the daemon is **not** given its own session (i.e., no `setsid`) — e.g. launched directly by a process supervisor's exec chain rather than via an interactive shell
3. Watch `daemon.log`: full successful start, auth, agent-version detection, workspace registration for all runtimes, then the log simply stops — no shutdown message, no panic, no error — anywhere from ~10s to ~90s in
4. `ps`/`strace` at the moment of death shows the daemon process itself issuing a `kill(SIGKILL)` against a PID it does not appear to own exclusively

## What we've tried that does *not* fix it

- Switching from `--foreground` to the documented default background mode + waiting on `daemon.pid` — same failure, same silent-death pattern
- Confirming the token, server_url/app_url config, and self-hosted backend connectivity are all correct (manual `multica login` + one-off `multica daemon start` runs the same commands successfully every time when run interactively via `su <user> -c '...'`, only reliably fails when supervised without a controlling terminal/session)

Happy to provide the full `daemon.log` and strace output, or run additional diagnostics (this is a self-hosted deployment, easy to reproduce on demand).

israoo:
Fixes/closes #8306

## Summary

`reapProcessTree`/`signalProcessGroup` (`server/pkg/agent`) and the `processtree` controller (`server/internal/daemon/processtree`) both signal a process group by the pid `Wait()` has already reaped, on the assumption documented in `reapProcessTree`'s own comment: *"sequential pid allocation makes reuse inside that window (microseconds, and only after wrapping the whole pid space) not a practical concern."*

That assumption doesn't hold inside a container: the pid namespace starts small and few processes are live at once, so a freshly-freed low pid number can be — and in practice is — handed to an unrelated process within microseconds, well inside the delay between a probe subprocess exiting and this package's own cleanup kill firing.

## Repro (full writeup + strace in #8306)

Self-hosted deployment, daemon run under s6-overlay (a process supervisor, no TTY). The daemon reliably launches 4 concurrent agent-version-detection subprocesses on startup (`claude`/`codex`/`copilot`/`antigravity --version`) and died silently — `SIGKILL`, no shutdown log line at all — anywhere from under a second to a few minutes later. `strace` attached live caught the daemon's own pid issuing `kill(-pid, SIGKILL)` immediately after a `SIGCHLD` from a *different*, already-reaped process — the cleanup kill had been handed a stale, already-reused pid.

## Fix

Tags every process this package starts with its `/proc/<pid>/stat` `starttime` the instant `Start()` returns (`pidtag_unix.go` for `server/pkg/agent`; the `controller` struct itself for `processtree`, which can't import the former without a cycle — see the module notice below on why the check is duplicated rather than shared). Before delivering a signal to a pid this package previously owned, it re-reads that pid's current `starttime` and compares:

- No tag recorded (non-Linux, `/proc` unreadable that one time): falls back to the prior, unguarded behavior — no false negatives introduced.
- Tag present but the pid is gone, or its `starttime` no longer matches: the process this package started has already exited and the number may have been reused. Skip — signalling now would land on whoever holds it, not the intended target.
- Tag present and `starttime` matches: still the same process (or the pid hasn't been reused yet either way). Signal as before.

This doesn't change behavior on any platform/setup where the original "not a practical concern" assumption actually holds — it only skips signals that would otherwise land on the wrong target.

## Verification

With **only** this patch applied — deliberately not combined with any other mitigation (no `setsid`, daemon left sharing its process group with whatever execs it) — the daemon ran stably for 6+ minutes under the exact s6-supervised, no-TTY setup that previously killed it in under a second, every time, without exception.

`go test ./server/pkg/agent/... ./server/internal/daemon/processtree/...` and `go vet ./...` both pass unchanged (existing suites, no new test added yet — happy to add a `/proc`-backed unit test for `readProcStartTime`/`stillOurProcess` if that's wanted before merge, or if there's a preferred way to fake `/proc` in this codebase's existing test style).

## Notes for reviewers

- `server/internal/daemon/processtree` and `server/pkg/agent` don't share code today (the latter imports patterns the former doesn't, to avoid a cycle per the existing package boundary), so the `/proc/<pid>/stat` parsing is duplicated in both files rather than factored into one helper. Happy to extract a small shared internal package instead if that's preferred.
- The starttime tag map in `pidtag_unix.go` is keyed by the `*exec.Cmd` pointer (not the pid), and cleared in `releaseProcessGroup` — which every existing call site already calls at the end of a command's lifecycle — so this shouldn't add unbounded memory growth over a long daemon uptime.
- Windows is untouched; the existing Job Object-based ownership model there doesn't have this class of pid-reuse race.

### DIFF
--- a/server/internal/daemon/processtree/controller_unix.go
+++ b/server/internal/daemon/processtree/controller_unix.go
@@ -44,18 +44,23 @@ func (c *controller) attach(cmd *exec.Cmd) error {
 	return nil
 }
 
-// stillOurs reports whether pid still refers to the process tagged with
-// startTime in attach. See the equivalent (and the reasoning behind it) in
+// stillOurs reports whether it is safe to signal pid, using the tag laid
+// down in attach. See the equivalent (and the reasoning behind it) in
 // server/pkg/agent/pidtag_unix.go — this package cannot import that one
 // without an import cycle, so the same small check is duplicated here rather
 // than shared.
+//
+// A pid that no longer exists at all answers true (harmless no-op signal;
+// also how a surviving grandchild left in the group after the leader exited
+// still gets reached). Only a pid that exists but is now a *different*
+// process — proven by a starttime mismatch — answers false.
 func (c *controller) stillOurs(pid int) bool {
 	if c.startTime == 0 {
 		return true
 	}
 	current, err := procStartTime(pid)
 	if err != nil {
-		return false
+		return true
 	}
 	return current == c.startTime
 }
--- a/server/internal/daemon/processtree/controller_unix.go
+++ b/server/internal/daemon/processtree/controller_unix.go
@@ -3,17 +3,27 @@
 package processtree
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
+	"strings"
 	"syscall"
 	"time"
 )
 
 const processTreeFinishTimeout = 5 * time.Second
 
-type controller struct{}
+// startTime is the /proc/<pid>/stat "starttime" observed right after
+// Start() returned, tagging which process actually owns this controller's
+// pid number. Zero means it could not be read (non-Linux, restricted /proc)
+// — stillOurs then falls back to the pre-patch behavior of trusting the pid
+// unconditionally, rather than reporting a false negative.
+type controller struct {
+	startTime uint64
+}
 
 func newController(cmd *exec.Cmd) (*controller, error) {
 	if cmd.SysProcAttr == nil {
@@ -23,11 +33,39 @@ func newController(cmd *exec.Cmd) (*controller, error) {
 	return &controller{}, nil
 }
 
-func (*controller) attach(_ *exec.Cmd) error { return nil }
+// attach tags the just-started process the instant it is available, so the
+// later interrupt/stop/finish calls — which may run long after this pid has
+// exited and potentially been reused — can tell whether they are still
+// signalling the process this controller started.
+func (c *controller) attach(cmd *exec.Cmd) error {
+	if cmd.Process != nil {
+		c.startTime, _ = procStartTime(cmd.Process.Pid)
+	}
+	return nil
+}
 
-func (*controller) interrupt(cmd *exec.Cmd) error {
+// stillOurs reports whether pid still refers to the process tagged with
+// startTime in attach. See the equivalent (and the reasoning behind it) in
+// server/pkg/agent/pidtag_unix.go — this package cannot import that one
+// without an import cycle, so the same small check is duplicated here rather
+// than shared.
+func (c *controller) stillOurs(pid int) bool {
+	if c.startTime == 0 {
+		return true
+	}
+	current, err := procStartTime(pid)
+	if err != nil {
+		return false
+	}
+	return current == c.startTime
+}
+
+func (c *controller) interrupt(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
+	}
+	if !c.stillOurs(cmd.Process.Pid) {
+		return nil
 	}
 	if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM); err != nil {
 		if errors.Is(err, syscall.ESRCH) {
@@ -38,9 +76,12 @@ func (*controller) interrupt(cmd *exec.Cmd) error {
 	return nil
 }
 
-func (*controller) stop(cmd *exec.Cmd) error {
+func (c *controller) stop(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
+	}
+	if !c.stillOurs(cmd.Process.Pid) {
+		return nil
 	}
 	if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
 		if errors.Is(err, syscall.ESRCH) {
@@ -57,6 +98,12 @@ func (c *controller) finish(cmd *exec.Cmd) error {
 	}
 	pid := cmd.Process.Pid
 	if err := syscall.Kill(-pid, 0); errors.Is(err, syscall.ESRCH) {
+		return nil
+	}
+	if !c.stillOurs(pid) {
+		// The pid we started has already exited and either no longer exists
+		// or now belongs to something this controller never launched.
+		// Nothing left here is ours to reap.
 		return nil
 	}
 	// A normally-exited leader can still leave a descendant holding inherited
@@ -76,3 +123,29 @@ func (c *controller) finish(cmd *exec.Cmd) error {
 }
 
 func (*controller) close() {}
+
+// procStartTime parses field 22 (starttime, in clock ticks since boot) from
+// /proc/<pid>/stat. The process name in field 2 is parenthesized and may
+// itself contain spaces or parentheses, so field boundaries are found from
+// the *last* ')' rather than by a fixed-width split.
+func procStartTime(pid int) (uint64, error) {
+	data, err := os.ReadFile("/proc/" + strconv.Itoa(pid) + "/stat")
+	if err != nil {
+		return 0, err
+	}
+	idx := bytes.LastIndexByte(data, ')')
+	if idx < 0 || idx+2 > len(data) {
+		return 0, os.ErrInvalid
+	}
+	// Fields from here on: state(3) ppid(4) pgrp(5) session(6) tty_nr(7)
+	// tpgid(8) flags(9) minflt(10) cminflt(11) majflt(12) cmajflt(13)
+	// utime(14) stime(15) cutime(16) cstime(17) priority(18) nice(19)
+	// num_threads(20) itrealvalue(21) starttime(22) ...
+	// so starttime is at 0-based index 22-3 = 19 in this slice.
+	const starttimeIndex = 19
+	fields := strings.Fields(string(data[idx+1:]))
+	if starttimeIndex >= len(fields) {
+		return 0, os.ErrInvalid
+	}
+	return strconv.ParseUint(fields[starttimeIndex], 10, 64)
+}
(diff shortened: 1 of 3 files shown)


israoo:
Pushed a fix for a real regression CI caught: my first version blocked the signal on *any* failure to read the pid's starttime, which also fired for a pid that simply no longer exists — breaking 5 existing tests that rely on this same kill still reaching a grandchild process left forked into the group after the direct child already exited (`TestRunCollectReapsForkedHelper`, `TestDetectCLIVersionReapsForkedHelper`, `TestCheckOpenclawVersionReapsPipeHoldingGrandchild`, `TestRunCollectRetriesTheTreeKill`, `TestProbeReturnsWhenLeaderExitsHoldingPipes`).

Fixed by distinguishing the two cases explicitly:
- pid doesn't exist at all → still signal (harmless no-op / reaches a surviving grandchild, exactly as before this patch)
- pid exists but its `/proc/<pid>/stat` starttime doesn't match what was recorded at `Start()` → skip (this is the actual bug: a different process now holds that number)

`go test ./server/pkg/agent/... ./server/internal/daemon/processtree/...` passes in full now (184.9s), `go vet` clean. Re-verified live on the same self-hosted setup afterward — no regressions observed.

multica-eve (changes requested):
Request changes: the hardening is worthwhile, but this implementation still permits the exact class of unsafe group signal it is intended to prevent and does not prove that it fixes #8306.

The observable goal is correct: when a daemon-owned probe exits or times out, cleanup must terminate only that probe's process group, never the daemon or an unrelated process that later reused the PID. The current patch cannot guarantee that outcome:

1. Identity verification fails open. In `server/pkg/agent/pidtag_unix.go:30-36,71-80`, failure to capture a tag leaves verification disabled, and a later `/proc` read/parse error is treated as a match. `server/internal/daemon/processtree/controller_unix.go:40-44,57-65` has the same behavior. A restricted or unexpectedly mounted `/proc`, permission failure, or parser error therefore falls back to the raw `kill(-pid, ...)` path. Please distinguish a missing leader (`ENOENT`, where signalling surviving descendants may be intended) from an unknown verification error, and fail closed with a diagnostic for the latter. Non-Linux behavior should be explicit rather than silently unverified.

2. The check and signal are not an atomic ownership guarantee. `/proc/<pid>/stat` start time is measured in clock ticks, and another process can replace the leader between verification and `kill(-pid, ...)`; same-tick reuse is also possible. This repository already has a stronger Linux mechanism in `server/pkg/agent/cursor_background_process_linux.go:16-40`: retain a pidfd as an ownership handle and use it when signalling the process group. Please generalize the existing owned-process-group mechanism (and the Darwin birth-identity equivalent), with an explicit fallback policy for unsupported kernels/platforms. If a best-effort fallback is intentionally accepted, document the residual race rather than claiming it is eliminated.

3. The new safety behavior has no regression coverage. Add injected identity-reader/signal-sender tests covering: mismatch blocks the syscall; capture failure; leader disappearance (`ENOENT`); other read/parse errors; concurrent release; and both affected implementations. Keep the existing grandchild-cleanup and EOF/lifecycle coverage intact.

4. The causal chain for #8306 is still unproven. v0.4.42 already starts the background daemon in a new session and probes `Setpgid`; Linux PID allocation is cyclic up to `pid_max`, so a container with few processes does not by itself establish immediate PID reuse. The supplied trace shows the daemon sending a signal, but not that the daemon receives that fatal signal. Please provide either a deterministic reproduction or a complete trace/harness recording PID, PPID, PGID, SID, signal target, and final exit status.

Verification on the PR head: `git diff --check`; `go test ./pkg/agent/... ./internal/daemon/processtree/... -count=1`; and `go test ./internal/daemon/execenv/ -run ByteIdentical -count=1` all passed locally. That preserves existing behavior but does not exercise the new Linux `/proc` mismatch/error paths; the local host was Darwin. If this implementation is wrong, cancellation, version probes, and helper cleanup across Unix runtimes can leave descendants holding pipes/locks or signal an unrelated process. The immediate rollback is to revert this PR.

multica-eve (changes requested):
Request changes: the latest head (`2b268ff5`) only merges `main`; it does not change the three production files or close any of the safety blockers from the previous review.

The product direction remains valid: cleanup must terminate only the probe/agent process tree owned by the daemon, never the daemon itself or an unrelated process after PID/PGID reuse. The current implementation still cannot guarantee that behavior:

1. Identity verification remains fail-open. `server/pkg/agent/pidtag_unix.go:24-36,67-81` and `server/internal/daemon/processtree/controller_unix.go:19-23,40-65` treat both initial capture failure and any later `/proc` read/parse failure as permission to continue with a negative-PGID signal. Please distinguish a genuinely missing leader (`ENOENT`, where surviving descendants may still need cleanup) from permission, mount, and parse failures; unknown failures must fail closed and produce a diagnostic.

2. The identity read and `kill(-pid, ...)` remain separate operations, so PID reuse between the check and signal is still possible; clock-tick `starttime` also cannot disambiguate same-tick reuse. Please generalize the repository's existing owned-group mechanisms—the persistent Linux pidfd group handle in `cursor_background_process_linux.go` and Darwin birth identity—and define an explicit fallback/residual-risk policy for unsupported kernels and platforms.

3. There are still no tests for the new safety logic. Add injected identity-reader/signal-sender coverage for mismatch blocking the syscall, capture failure, `ENOENT`, other read/parse errors, concurrent release, and both the agent and processtree implementations. Existing grandchild/EOF tests only show that prior cleanup behavior remains intact.

4. The #8306 causal chain remains incomplete. The available trace shows the daemon sending a group signal, but not the daemon receiving the fatal signal. Please provide a deterministic reproducer or a complete trace/harness recording PID, PPID, PGID, SID, signal target, and final exit status.

Latest-head verification: `git diff --check` passed; the execenv `ByteIdentical` test passed; the processtree suite passed; and the five directly related grandchild/pipe cleanup regressions passed. The full agent suite hit unrelated Codex/Kimi timeout flakes that passed when rerun in isolation. GitHub CI is also currently red because `TestZeroclawSessionNew` hit `text file busy` (with the aggregate `backend` job failing as a consequence). None of those results exercises the missing Linux `/proc` mismatch/error paths; the local host was Darwin.

If this is wrong, Unix cancellation, version probing, and helper cleanup can still signal an unrelated process or leave descendants holding pipes and locks. The immediate rollback remains reverting this PR.

PR #8321, opened 2026-09-11
MUL-7299: feat(issues): filter issues by the parent project's status

tempo22:
## What

Adds a **Project status** dimension to the issue filter panel, next to the existing **Project** filter: pick `In Progress` once instead of ticking every active project by hand. The two dimensions AND together, and both stay optional and off by default.

## Server

`filters.project_statuses` on the issue-table query spec, as an `EXISTS` over `project` modelled on the label filter. Notable details:

- Values are validated against `validProjectStatuses`, so an unknown one is a clean 400 rather than a DB error.
- The `EXISTS` is bound with `p.workspace_id = i.workspace_id`. That is not redundant: the schema has no foreign keys by design, so a stale `issue.project_id` can name a project in another workspace, and without the bound that tenant's status would decide this row's membership.
- An issue with no project never matches — `project_id IS NULL` makes the `EXISTS` false. `include_no_project` widens the project-id dimension only.
- The predicate joins the shared `where`, so rows / groups / facets inherit it with no per-consumer change. The new field is normalized into the query fingerprint like the other set-like arrays, or equivalent filter sets would produce different cursors.

No migration, and no index: `project(status)` is worth adding only if a perf need shows up.

## Client

`projectStatusFilters` in the view store — persisted, resettable per dimension, with its own filter chip, and part of a saved view's snapshot.

The client-side predicate in `filter.ts` serves Gantt and the swimlane extra-children merge (every other surface filters server-side). It needs the project catalog, which reaches it through `IssueFilterContext` and is pulled in by `loadProjects` whenever the filter is on. Two states matter and neither is "no filter":

- **Catalog absent** (surface never loads it): the predicate is a no-op. A defined-but-empty map would drop every issue and blank the board.
- **Catalog unresolved** (cold load or failed request) on a surface that *does* apply the predicate: the surface holds in loading and reports the failure, mirroring `statusFilterPending` / `statusFilterError` for a custom status filter. Returning unfiltered rows under an active chip is as wrong as blanking the board.

Project writes invalidate the issue-table caches, locally and over realtime. Nothing in an issue's payload changes when its *project's* status does, and the global `staleTime` is `Infinity`, so without it a filtered window stayed stale until reload.

Saved views written before this dimension carry no key; every read path defaults to an empty list, and `mergeViewStatePersisted` drops members the client cannot represent.

i18n keys added in **en, ja, ko, zh-Hans**.

## Tests

| Layer | What it pins |
| --- | --- |
| `server/internal/handler/issue_table_project_status_test.go` | The SQL predicate: single status, OR within the dimension, AND with the project-id filter, `include_no_project` not bypassing it, the cross-workspace bound, unknown status → 400 |
| `packages/views/issues/utils/filter.test.ts` | The client predicate, including the absent-catalog no-op |
| `packages/core/issues/stores/view-store-project-status.test.ts` | Toggle / reset / persist sanitization |
| `packages/core/issue-views/baseline.test.ts` | Saved views predating the dimension |
| `packages/views/issues/surface/use-issue-surface-controller.test.tsx` | Store → `project_statuses` in the request body |
| `packages/views/issues/surface/use-issue-surface-catalog.test.tsx` | Hold / error while the project catalog is unresolved |
| `e2e/issue-project-status-filter.spec.ts` | The filter in the running app, and a project moving into the selected status refreshing the list |

Verified locally: `pnpm typecheck`, `pnpm lint` (0 errors), the TS suites per package, `go test ./internal/handler`, and Playwright.

## Review notes

- `decodeIssueTableJSON` uses `DisallowUnknownFields()`, so a new client against an older backend gets a 400 on this field. The direction the repo's compatibility rules care about — old client, new backend — is fine, since the field is omitted.
- The sub-menu shows no per-status counts, unlike Status / Priority / Project. That would need a `project_status` server facet, which is out of scope here.

Two adversarial review passes (OpenAI Codex, read-only) ran against this diff; the findings they raised are fixed in the last two commits, and each fix has a test that was verified to fail without it.


multica-eve (changes requested):
The product direction is valid, but the current implementation has a cache-correctness blocker and cannot merge cleanly yet. Please address the following before merging:

1. In `packages/core/projects/mutations.ts`, the Issue Table invalidation for project deletion was added to `useCreateProject().onSettled` (the comment itself describes deletion), while `useDeleteProject().onSettled` only invalidates the project list. If the realtime event is missed, queries with `staleTime: Infinity` can keep showing issues from the deleted project indefinitely. Move the `issueKeys.tableAll(wsId)` invalidation to the delete mutation, retain the status-update invalidation, and add a mutation regression test.
2. Rebase onto the current `main` and resolve the conflicts in `packages/core/issues/stores/view-store.ts`, `packages/views/issues/surface/use-issue-surface-data.ts`, and `server/internal/handler/issue_table_query.go`, adapting the change to the latest issue-lifecycle model.
3. Reuse the existing canonical `useProjectStatusLabels()` mapping instead of duplicating the five project-status translations in every `issues.json` locale. Only the new “Project status” filter-dimension label should need new copy.

The workspace-bounded, parameterized `EXISTS` predicate and the separate opt-in filter dimension are otherwise the right shape for this feature. Please rerun the targeted backend/frontend tests and the Playwright scenario after rebasing.

Issue #8329, opened 2026-09-11
Skill changes in the agent UI silently fail to save unless you also edit another field


asdf8675309:
### Deployment type

Self-host

### What happened?

I removed a skill from an agent in the web UI. The panel showed it gone. The server still had it.

Reading the agent back over the CLI showed the old skills, and `updated_at` sat at its old value. The save stopped short of the server, and the UI stayed quiet about it.

It works if you edit something else at the same time. Change the agent's description in the same save and both changes land.

One more thing I noticed, which may be separate: the per-skill control seems to set `enabled: false` rather than drop the assignment. A skill I thought I removed was still in the `skills` array, just disabled. That may be intended — disable and unassign are different operations — so I'm flagging it as an observation.

### Steps to reproduce

1. Open an agent that has skills assigned.
2. Remove or toggle a skill under Capabilities, then save.
3. Run `multica agent get <agent-id> --output json`.
4. The `skills` array and `updated_at` both still show the old state.
5. Do it again, and edit the description in the same save.
6. Read it back. Both changes are there.

### Expected behavior

A skill change should save on its own.

### Version info

```
app binary: v0.4.39 (commit 0eff894b5, built 2026-09-03T10:44:39Z, darwin/arm64)
daemon:     0.4.39 (commit 0eff894b5, built 2026-09-03T10:38:43Z, linux/amd64)
```

### Additional context

#3459 and PR #3464 may be relevant. That one found `UpdateAgent` leaving the `agent_skill` table alone while the response wrongly reported empty skills — the DB writes were fine. It establishes that skill bindings go through a separate path from the general agent update, which might explain why a skills-only change gets dropped while a combined one lands.

Why this stings: you think a skill is gone and the agent can still use it. It also breaks definitions-as-code tooling, since live state disagrees with what you set and nothing warns you.

Issue #3459, opened 2026-05-28
agent update clears existing skill assignments


NewmanXBT:
## Summary

`multica agent update` clears an agent's existing skill assignments when updating only agent metadata/instructions.

## What happened

I updated an existing workspace agent's description and instructions:

```sh
multica agent update <agent-id> \
  --description "..." \
  --instructions "..." \
  --output json
```

The update succeeded, but the returned agent payload showed:

```json
"skills": []
```

Before the update, the agent had five assigned skills:

- `analyzing-user-feedback`
- `doc-coauthoring`
- `founder-knowledge-systems`
- `obsidian`
- `research-monitoring-and-source-retrieval`

I had to restore them manually with:

```sh
multica agent skills set <agent-id> --skill-ids <original-skill-ids>
```

## Expected behavior

Updating description/instructions should preserve existing skill assignments unless the update request explicitly changes skills.

## Actual behavior

Existing skills were cleared even though no skill update was requested.

## Impact

This is easy to miss in automation: an agent instruction update can silently degrade the agent by removing tool/skill capabilities. It also makes safe operating-system updates harder because every `agent update` needs an immediate read-back and manual skill restore guard.

## Suggested fix

- Make `agent update` preserve existing skills by default.
- Only mutate skills through `multica agent skills set` or an explicit skills field.
- Add a regression test that updates an agent's description/instructions and verifies its skill assignments are unchanged.

## Workaround

After every `multica agent update`, run:

```sh
multica agent get <agent-id> --output json
```

and restore skills with `multica agent skills set` if they were cleared.

PR #8330, opened 2026-09-11
MUL-7303 fix(agents): touch last-modified time on skill changes

Dante-dan:
## What does this PR do?

Skill assignments already persisted through the dedicated skill endpoints, and dispatch already respected disabled and removed assignments. Those writes left the parent Agent's `updated_at` unchanged. This patch touches that last-modified timestamp in the same transaction as each replace, add, toggle, or remove operation.

Disable still retains the assignment with `enabled: false`; remove still deletes it. The related #3459 / #3464 fix addressed response hydration rather than persistence. This change likewise does not establish or fix the silent persistence loss reported in #8329, so it references that issue without closing it.

`updated_at` remains a last-modified timestamp, not a strictly monotonic revision or a guarantee that multiple writes within one second are distinguishable through the public API. The transaction keeps the assignment write and timestamp touch atomic: a failed touch or commit rolls both back.

## Related Issue

Refs #8329

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/skill.go`: touch the Agent inside each workspace-skill mutation transaction.
- `server/pkg/db/queries/skill.sql` and generated code: add `TouchAgentForSkillChange`, which updates the Agent row.
- `server/internal/handler/handler_test.go`: use a deterministic backdated baseline for all four mutation endpoints; retain disable-versus-unassign assertions.
- `server/internal/handler/agent_skill_transaction_test.go`: exercise real PostgreSQL relation writes with injected touch/commit failures for toggle and remove; verify the assignment and Agent timestamp remain unchanged.

## How to Test

Run from `server/` with an isolated PostgreSQL database:

1. `go test ./internal/handler -run '^TestAgentSkill(MutationsTouchAgentUpdatedAt|ToggleAndRemoveRollback)$' -count=1 -v` — passed, including all eight new/expanded subcases, using the repository agent-CLI guard.
2. `go test ./internal/handler ./pkg/db/generated -run '^(TestAgentSkill|TestSetAgentSkills|TestAddAgentSkills)' -race -count=1 -v` — passed using the repository agent-CLI guard.
3. `go vet ./internal/handler ./pkg/db/generated` — passed.
4. From the repository root, `make sqlc` and `git diff --check` — passed.

## Checklist

## AI Disclosure

**Prompt / approach:** Used Codex to trace the dedicated skill handlers and transaction boundaries, compare #3464's response-hydration fix, and implement the parent timestamp touch. For review feedback, narrowed the contract to last-modified semantics and added the requested endpoint and rollback regressions against PostgreSQL.

## Screenshots (optional)

### DIFF
--- a/server/internal/handler/handler_test.go
+++ b/server/internal/handler/handler_test.go
@@ -1909,6 +1909,73 @@ func TestAddAgentSkillsRejectsCrossWorkspaceSkillID(t *testing.T) {
 	assertAgentSkillRowCount(t, agentID, 0)
 }
 
+// A workspace-skill binding is part of the persisted agent definition even
+// though it lives in agent_skill. Before #8329 these endpoints changed the
+// junction row without advancing agent.updated_at, so skill-only edits were
+// invisible to consumers that use the parent timestamp as their version.
+// Keep disable (the row remains with enabled=false) distinct from unassign
+// (the row is deleted) while proving both changes advance the parent version.
+func TestAgentSkillToggleAndRemoveAdvanceAgentUpdatedAt(t *testing.T) {
+	agentID := createHandlerTestAgent(t, "Handler Skill Timestamp", nil)
+	skillA := insertHandlerTestSkill(t, "timestamp-a", "a body")
+	skillB := insertHandlerTestSkill(t, "timestamp-b", "b body")
+	dbfx.Exec(t,
+		`INSERT INTO agent_skill (agent_id, skill_id) VALUES ($1, $2), ($1, $3)`,
+		agentID, skillA, skillB,
+	)
+
+	backdate := func() time.Time {
+		t.Helper()
+		past := time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC)
+		dbfx.Exec(t, `UPDATE agent SET updated_at = $2 WHERE id = $1`, agentID, past)
+		return past
+	}
+	readUpdatedAt := func() time.Time {
+		t.Helper()
+		var got time.Time
+		dbfx.QueryRow(t, `SELECT updated_at FROM agent WHERE id = $1`, agentID).Scan(&got)
+		return got
+	}
+
+	past := backdate()
+	toggleReq := newRequest("PUT", "/api/agents/"+agentID+"/skills/"+skillA+"/enabled", map[string]any{
+		"enabled": false,
+	})
+	toggleReq = withURLParam(toggleReq, "id", agentID)
+	toggleReq = withURLParam(toggleReq, "skillId", skillA)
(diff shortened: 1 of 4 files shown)


multica-eve (changes requested):
Request changes — the transaction wiring is directionally correct and preserves the existing Disable versus Unassign behavior, but the PR currently describes a stronger contract than it implements and the regression coverage does not yet protect all four mutation paths or the new rollback behavior.

Before this patch, skill assignments were already persisted correctly and task dispatch already respected both `enabled = false` and deleted assignments; only the parent Agent's last-modified timestamp stayed unchanged after a skill-only edit. After this patch, each skill mutation should touch that parent timestamp in the same transaction, without changing those assignment semantics.

Please address these items before merge:

1. Narrow the wording to the actual last-modified contract. Update the title/body and the source/test comments that call `agent.updated_at` a "version" or "parent version." State explicitly that the skill mutation itself was already persisted and that this patch only makes skill-only edits touch the parent Agent's last-modified timestamp. Do not describe this as a fix for silent save loss, and do not promise strict monotonicity or that multiple writes within one second are distinguishable through the public API.

2. Complete the regression coverage for the accepted contract. Keep the deterministic backdated timestamp baseline, but cover all four endpoints: replace, add, toggle, and remove. For the newly transactional toggle and remove paths, also add failure-path assertions showing that the `agent_skill` mutation is rolled back when the timestamp touch or transaction commit fails. Please do not add concurrency, same-second wire-format, or strict-monotonicity assertions; those are outside this PR's last-modified semantics.

3. Change `Closes #8329` to `Refs #8329`. The upstream report claims silent persistence loss, which this patch does not establish or fix, so merging must not automatically close it.

Non-blocking nit: `TouchAgentForSkillChange` would describe the query more accurately than `TouchAgentSkills`, since the query updates the Agent row rather than the Skill rows.

Dante-dan:
Addressed the review in 28549fc0a (https://github.com/multica-ai/multica/commit/28549fc0a4f2b275bc0317c4d9400ffaea125e89).

The title, description, and source/test comments now describe only the Agent's last-modified timestamp. Skill assignments were already persisted; I changed the issue reference to `Refs #8329` and renamed the query to `TouchAgentForSkillChange`.

The backdated-baseline regression now covers replace, add, toggle, and remove, while preserving disable versus unassign behavior. For toggle and remove, separate touch-failure and commit-failure cases execute the relation write against PostgreSQL and verify that the enabled assignment and original timestamp survive the rollback.

All eight requested subcases passed. The related skill-handler tests also passed with `-race` under the repository's agent-CLI guard; `make sqlc`, `go vet ./internal/handler ./pkg/db/generated`, and `git diff --check` passed. No concurrency, same-second API, or strict-monotonicity assertions were added.

PR #8331, opened 2026-09-11
MUL-7304: feat(daemon): per-provider concurrency caps via `MULTICA_<PROVIDER>_MAX_CONCURRENT_TASKS`

mmontes11:
## What does this PR do?

**The problem: one machine-wide cap, providers with very different capacities.**

Multica's value is mixing providers on one machine — high-concurrency cloud agents (Claude) alongside local ones (OpenCode on your own GPU). Today concurrency has one machine-wide knob, `MULTICA_DAEMON_MAX_CONCURRENT_TASKS`. Per-agent settings don't fix the mismatch: several agents on the same local provider can still be scheduled together under a cap > 1.

Concrete example: `MULTICA_DAEMON_MAX_CONCURRENT_TASKS=4` with Claude (which happily runs 4 in parallel) and a local OpenCode whose model occupies the whole GPU (which can really only run 1). The cap applies to both — 4 is fine for Claude, but four simultaneous local OpenCode runs thrash the GPU and all four get slow. The only way out today is dropping the machine-wide cap to 1, which means giving up Claude's parallelism (e.g. fanning a planning issue into sub-issues across agents). You shouldn't have to renounce cloud concurrency because a local provider can't handle it.

**The change:** `MULTICA_<PROVIDER>_MAX_CONCURRENT_TASKS` — a per-provider cap on top of the global slot pool, keyed by the runtime's provider field uppercased verbatim (`MULTICA_CLAUDE_MAX_CONCURRENT_TASKS`, `MULTICA_OPENCODE_MAX_CONCURRENT_TASKS`, `MULTICA_CODEX_MAX_CONCURRENT_TASKS`, …). The example above becomes:

```
MULTICA_DAEMON_MAX_CONCURRENT_TASKS=4
MULTICA_OPENCODE_MAX_CONCURRENT_TASKS=1
```

Claude keeps its 4 parallel runs; the local provider is hard-capped at 1. Providers without the variable are unlimited — behavior is unchanged unless you opt in. A custom runtime profile is capped through its protocol family's variable, not its display name.

## Why park-and-hold, not re-queue

A claimed task whose provider is at its cap parks in `handleTask` — still holding its global slot — until a provider token frees:

- While parked, the task's prepare lease is extended on a fixed cadence so the server never reclaims and re-delivers the dispatched task mid-wait, and a cancellation watcher aborts the wait the moment the server moves the task to a terminal state. Parent-context cancellation (daemon shutdown) parks nothing and fails nothing: the task stays dispatched and the runtime-offline recovery path retries it.
- The wait runs **before** the local_directory path lock, so a parked task never holds a directory mutex.
- Unset means unlimited; invalid values are logged and treated as unlimited. Limits are read lazily per provider (first dispatch) and memoized for the daemon's lifetime; the daemon's environment is fixed for the life of the process, so memoization cannot mask a changed value. No `Config` field, no CLI flag.

Re-queueing (release the global slot, promote on free) needs promotion machinery that does not exist today and would re-deliver the task through the whole claim path. Holding the slot reuses the existing lease-extender + cancellation-watcher pattern from the local_directory wait and keeps the code path small. The documented trade-off is head-of-line blocking: if 30 GPU tasks queue behind a cap of 1, they occupy 30 global slots while parked. That matches the "this machine has N GPUs" mental model and can be bounded by the existing machine-wide cap.

## Related Issue

None. This is a new feature request.

## Type of Change

- [x] Documentation update

## Changes Made

- `server/internal/daemon/provider_concurrency.go` (new) — per-provider token pool: env-var mapping, memoized pool lookup, `waitForProviderCapacity`
- `server/internal/daemon/daemon.go` — `handleTask` waits for provider capacity after the global slot claim, before the local_directory path lock
- `server/internal/daemon/provider_concurrency_test.go` (new) — 9 tests
- `CLI_AND_DAEMON.md` — env-var row under "Agent-specific overrides"
- `apps/docs/content/docs/daemon-runtimes.mdx` + `.ja.mdx` / `.ko.mdx` / `.zh.mdx` — per-provider cap paragraph in "Concurrency limits"

## How to Test

1. `cd server && go test ./internal/daemon/ -run 'TestProviderLimitEnvVar|TestProviderPool|TestWaitForProviderCapacity|TestHandleTaskEnforcesProviderConcurrency'`
2. Manual: with `MULTICA_DAEMON_MAX_CONCURRENT_TASKS=2` and `MULTICA_OPENCODE_MAX_CONCURRENT_TASKS=1`, dispatch two runs to an OpenCode agent and one to a Claude agent. The second OpenCode run parks (still holding a machine slot) until the first finishes; the Claude run proceeds immediately.
3. `cd server && go build ./... && go vet ./internal/daemon/...`

## Checklist

- [x] I have included a thinking path that traces from project context to this change (see "Why park-and-hold" above)
- [x] If this change affects the UI, I have included before/after screenshots (N/A — daemon only)
- [x] If I added a new runtime / coding tool / UI tab, I synced the change to landing copy and relevant docs (N/A — new env var, not a runtime)
- [x] If this PR touches Chinese product copy, I checked it against `apps/docs/content/docs/developers/conventions.zh.mdx` (zh docs paragraph follows that page's existing terminology)
- [x] I have considered and documented any risks above (head-of-line blocking, documented in "Why park-and-hold")

## AI Disclosure

**AI tool used:** Multica agent

**Prompt / approach:** Feature requested via community issue (LAB-128). The agent implemented the per-provider token pool, tests, and docs from the request, then iterated on review feedback (docs wording, test cases, localized docs, table placement).


multica-eve (changes requested):
Thanks for the contribution. We’re closing this PR because this is not the configuration model we want to adopt.

The need is valid: one machine can run cloud and local providers with different capacity constraints. However, adding a public environment variable for every provider creates a growing daemon-local configuration surface that users cannot manage or inspect from Multica, and it introduces a second source of truth.

Our intended direction is to configure capacity in the existing Runtime machine UI, only for providers actually registered on that machine, with server-side persistence and owner permissions. Effective concurrency should be the minimum of the machine, runtime/provider, and Agent limits, and the daemon should report whether the desired configuration has been applied. A single structured override may remain useful for headless or Kubernetes deployments, but environment variables should not be the primary product model.

The current implementation also claims provider-saturated tasks into the global worker pool before waiting. Older tasks from one capped provider can therefore occupy every global slot and prevent unrelated providers from starting. Fixing that within the environment-variable design would still leave the wrong product surface, so this PR should not be merged as-is.

PR #8332, opened 2026-09-12
MUL-7305 Feat/lark outbound image message

zprial:
## What does this PR do?

Feishu/Lark conversations can now receive the files an agent attached to its reply. Previously only the text arrived — the Lark outbound path had no attachment hop at all.

`multica attachment upload` already puts files in object storage, and task completion already binds them to the assistant `chat_message` (`BindChatAttachmentsToMessage`). Everything downstream of that bind assumed a chat window in a browser, so a Feishu conversation was told it could not take files — the same gap the WeCom adapter closed in `integrations/wecom/outbound_media.go`. This closes that hop for Lark, following the decisions that file settled.

Thinking path: the reply already reached Lark (`sendChatReply`), the binding already existed in the DB, and WeCom already solved the identical last hop — so the change is scoped to the one missing link plus the transport calls it needs, not to the reply or binding paths.

Why this shape:
- The answer goes first, always. An upload is megabytes and round trips and can fail; the sentence the agent wrote must not wait behind one and must not be lost to one. Delivery runs after the reply is out, on its own goroutine and budget, and its worst outcome is one extra line saying a file did not make it.
- Images and non-images split on the attachment row's own content type, not the extension: Lark validates an upload against the declared kind and refuses a mismatch instead of downgrading it, so `larkFileTypeFor` maps only extensions that name a kind exactly and sends everything else as `stream`.
- The four transport methods live on the concrete client and are requested through a narrow interface, so `APIClient` — and the production stub plus four test fakes behind it — is untouched.

## Related Issue

No issue exists for this; opening one first would be reasonable if the maintainers prefer that.

## Type of Change

## Changes Made

- `server/internal/integrations/lark/http_client_media.go` (new) — `UploadImage` / `UploadFile` / `SendImageMessage` / `SendFileMessage`: multipart upload, a token-refresh retry mirroring `doAuthedJSON`, and the per-endpoint size ceilings.
- `server/internal/integrations/lark/outbound_media.go` (new) — the delivery hop: attachment lookup by assistant message id, object read, upload, send, and the failure notices.
- `server/internal/integrations/lark/outbound.go` — `Patcher` gains `media` (type-asserted off the client) and `objects`; `sendChatReply` starts the hop once the reply has landed. An attachment-only reply (empty text, which task.go supports as "the attachment cards ARE the response") now runs the delivery instead of returning early.
- `server/internal/integrations/lark/params.go` — `UploadImageParams`, `UploadFileParams`, `SendImageParams`, `SendFileParams`.
- `server/internal/integrations/lark/outbound_media_test.go` (new) — 10 tests.
- `server/cmd/server/router.go` — one `if store != nil` calling `patcher.SetAttachments(store)` and `h.DeclareChannelFileDelivery(string(channel.TypeFeishu))`.

## How to Test

1. `go test -race ./internal/integrations/lark/`
2. `go vet ./...` (compiles every test file in the repo)
3. Not runnable in my environment — no bot credentials. Against a real bot: in a bound Feishu chat, have an agent run `multica attachment upload chart.png` and reply; expect the text reply, then the image as a following message in the same thread. A non-image attachment should arrive as a file message.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, backend only
- [ ] I have updated relevant documentation to reflect my changes — nothing documents the Lark attachment path
- [ ] runtime / coding tool / UI tab sync — N/A
- [x] Chinese product copy checked — the two new user-facing strings are English, matching this adapter's existing strings (`replier.go`, `outbound.go`); the Chinese wording lives only in the China-only WeCom adapter

## AI Disclosure

**AI tool used:** Claude Code (running as a Multica agent)

**Prompt / approach:** The gap was identified from a conversation about files not arriving in Feishu. The agent cloned the upstream repo, found the missing hop by comparing the Lark outbound path against the WeCom adapter, implemented it, and self-reviewed it — including a test verified to fail without the fix, `-race`, and a full-repo `go vet`.

## Screenshots (optional)

N/A

### DIFF
--- a/server/cmd/server/router.go
+++ b/server/cmd/server/router.go
@@ -604,6 +604,25 @@ func NewRouterWithOptions(pool *pgxpool.Pool, hub *realtime.Hub, bus *events.Bus
 				patcher := lark.NewPatcher(cs, installSvc, larkClient, lark.PatcherConfig{})
 				patcher.Register(bus)
 
+				// Second hop for the agent's files: the files an agent
+				// bound to its reply are read back out of object storage,
+				// uploaded to Lark, and posted behind the answer. Mirrors
+				// the WeCom branch below.
+				//
+				// SetAttachments and DeclareChannelFileDelivery sit on one
+				// `if` on purpose. The first builds the hop; the second is
+				// the same fact told to the agent, and a run must not be
+				// promised a delivery this branch never built. A deployment
+				// with no object storage has nothing to read an attachment
+				// out of, and answering false is the safe direction: an
+				// agent that says "见附件" into a room where nothing is
+				// attached leaves the reader hunting for a file that was
+				// never sent.
+				if store != nil {
+					patcher.SetAttachments(store)
+					h.DeclareChannelFileDelivery(string(channel.TypeFeishu))
+				}
+
 				// Typing indicator: shows a "processing" reaction on the user's
 				// message while the agent is working, then removes it before the
 				// reply is sent. Best-effort; failures are logged only.
--- a/server/internal/integrations/lark/http_client_media.go
+++ b/server/internal/integrations/lark/http_client_media.go
@@ -1,0 +1,320 @@
+package lark
+
+// http_client_media.go — the transport half of agent-file delivery.
+//
+// Four methods, and they are declared on the concrete client rather than
+// added to APIClient on purpose. APIClient is the seam every consumer in
+// this package is wired through, including the production stub and four
+// test fakes; four more methods there would mean four more no-op stubs in
+// each of them for a capability exactly one call site uses. The delivery
+// path asks for the narrow interface in outbound_media.go instead and
+// degrades to "attachments are not deliverable here" when the client does
+// not implement it — which is precisely what the stub means.
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"mime/multipart"
+	"net/http"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	// larkImageUploadMaxBytes is Lark's ceiling for image_type=message.
+	// Enforced here rather than left to Lark because a file that is too
+	// big has to reach the user as a sentence, and by the time Lark
+	// refuses the body the delivery path has already told itself the
+	// file was sendable.
+	larkImageUploadMaxBytes = 10 << 20 // 10 MiB
+
+	// larkFileUploadMaxBytes is Lark's ceiling for POST /im/v1/files.
+	larkFileUploadMaxBytes = 30 << 20 // 30 MiB
+
+	// The two upload endpoints. Both take multipart/form-data and both
+	// answer with the usual {code,msg,data} envelope.
+	larkUploadPathImage = "/open-apis/im/v1/images"
+	larkUploadPathFile  = "/open-apis/im/v1/files"
+)
+
+// mediaUploadForm is one multipart POST to a Lark upload endpoint: a few
+// text fields followed by a single file part. Both endpoints share that
+// shape (image_type + image; file_type + file_name + file), so one
+// builder serves both.
+type mediaUploadForm struct {
+	// textFields are written in order, ahead of the file part.
+	textFields [][2]string
+	fileField  string
+	filename   string
+	data       []byte
+}
+
+// build renders the form into a fresh buffer.
+//
+// Called once per attempt, not once per upload: the token-refresh retry
+// in doAuthedUpload re-sends, and a multipart body is a one-shot reader.
+func (f mediaUploadForm) build() (*bytes.Buffer, string, error) {
+	var body bytes.Buffer
+	w := multipart.NewWriter(&body)
+	for _, kv := range f.textFields {
+		if err := w.WriteField(kv[0], kv[1]); err != nil {
+			return nil, "", fmt.Errorf("write field %s: %w", kv[0], err)
+		}
+	}
+	// CreateFormFile rather than CreatePart: it escapes the filename for
+	// us, and Lark reads the declared content type as a hint only — the
+	// bytes are what it validates a file against.
+	part, err := w.CreateFormFile(f.fileField, filepath.Base(f.filename))
+	if err != nil {
+		return nil, "", fmt.Errorf("create file part: %w", err)
+	}
+	if _, err := part.Write(f.data); err != nil {
+		return nil, "", fmt.Errorf("write file part: %w", err)
+	}
+	if err := w.Close(); err != nil {
+		return nil, "", fmt.Errorf("close multipart writer: %w", err)
+	}
+	return &body, w.FormDataContentType(), nil
+}
+
+// doAuthedUpload posts one multipart form to a Lark upload endpoint,
+// minting a tenant_access_token first and refreshing it once on a token
+// rejection.
+//
+// That contract is copied from doAuthedJSON because it is needed for the
+// same reason: an app_secret can be rotated under a long-lived process,
+// and the cached token then fails every call until something invalidates
+// it. An upload that skipped the retry would report a credential event to
+// the user as "your file could not be sent".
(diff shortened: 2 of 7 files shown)


Bohan-J (changes requested):
Thanks for this — the gap is real and the product direction is right. Feishu/Lark conversations currently drop the files an agent bound to its reply, and closing that last hop the way `integrations/wecom/outbound_media.go` closed it is the shape I'd pick too. The commit hygiene and the reasoning in the comments are genuinely good.

I went through the implementation against the WeCom pipeline and the Lark API docs and I'm requesting changes: four issues would show up in normal use, and each one has an established answer already in this repo.

## 1. Capacity protection runs too late, and sheds instead of waiting

`server/internal/integrations/lark/outbound_media.go:151-208`. `deliverAttachmentsAsync` spawns the goroutine unconditionally, `deliverAttachments` then runs `ListAttachmentsByChatMessage`, and only *after* the lookup does it contend for the 3-slot semaphore. Two consequences:

- Goroutines and DB lookups are unbounded. The semaphore bounds neither, because both happen before it.
- The `default:` branch drops a delivery immediately. The 4th concurrent turn that actually has a file gets `mediaSendFailedText` without a single upload attempt.

WeCom settled this with three tiers rather than one, and the reasoning is written out at `wecom/outbound_media.go:208-222` and `:600-630`: pre-spawn admission (`admitAttachmentDelivery`, cap 64) bounds the goroutines and the lookups; a post-lookup pending cap (32) is the gate that can honestly name a dropped file; and the concurrency semaphore waits on `ctx.Done()` instead of shedding on `default`. Worth mirroring, along with tests that pin both bounds — `TestDeliverAttachments_AdmissionBoundsTheLookupStage` is the model.

The value `3` also needs a basis. `mediaUploadForm.build()` copies `data` into a fresh `bytes.Buffer`, so each in-flight delivery holds roughly twice the file size — 3 × 2 × 30 MiB ≈ 180 MiB resident. And these uploads go through `c.cfg.HTTPClient`, whose default is `defaultRequestTimeout = 10 * time.Second` (`http_client.go:60`): a 30 MiB upload has to sustain ~24 Mbps to finish inside it. This package already decided media transfers need their own budget — `ResourceHTTPClient` / `DefaultResourceDownloadTimeout = 45s` (`http_client.go:37-40`) exists for exactly that on the inbound side. The upload path should use the same seam rather than the 10s JSON client.

## 2. An uncertain outcome is reported as a definite failure

`outbound_media.go:211-224` and `:273-303`. Every error from `SendImageMessage` / `SendFileMessage` increments `failed`, and the user is then told **"A file didn't make it. I still have it — say the word and I'll resend."** An HTTP timeout or a connection reset on that POST means the request may well have been accepted and only the response was lost — so following that instruction produces a duplicate attachment in the chat.

To be precise about where this does and doesn't apply: a failed *upload* is genuinely definite (no `image_key`/`file_key` was returned, so no message was ever addressed), and it's fine to keep reporting those as failures. It's the final send POST that is ambiguous. WeCom models this explicitly with a tri-state — `deliveryDefinitelyFailed` / `deliveryUnknown` / delivered, `sendOutcome`, and a separate `mediaSendUnknownText` (`wecom/outbound_media.go:340-400`) — so an unconfirmed send never borrows the definite wording.

Lark also gives you a cleaner fix than a wording change. Both `POST /open-apis/im/v1/messages` and `POST /open-apis/im/v1/messages/:id/reply` accept an optional `uuid` (≤50 chars) that dedupes for one hour: [send](https://open.feishu.cn/document/server-docs/im-v1/message/create), [reply](https://open.feishu.cn/document/server-docs/im-v1/message/reply). Threading the attachment row id through `outboundMessageRequest` as `uuid` makes a retry after an ambiguous timeout safe, which is a better outcome than asking the user to decide.

Separately, and this one is unconditional: when the 5-minute `attachmentBudget` expires, `tellUser` at `:223` is handed the same cancelled `ctx`, so `SendTextMessage` fails immediately and the notice cannot be sent — the path that most needs to speak is the one that's guaranteed silent. WeCom hits this and uses `context.WithoutCancel(ctx)` (`wecom/outbound_media.go:310`).

## 3. `image/*` is not the same as "an image Lark will accept"

`outbound_media.go:261-263` and `:333-342`. `isImageAttachment` returns true for any `image/*` prefix, and anything it accepts is hard-routed to the image endpoint.

`server/internal/handler/file.go:30` maps `.svg` → `image/svg+xml` and writes that straight into `attachment.content_type`. So an agent that uploads an SVG chart gets its file sent to the image endpoint, which [does not accept SVG](https://open.feishu.cn/document/server-docs/im-v1/image/create) (JPG, JPEG, PNG, WEBP, GIF, BMP, ICO, TIFF, HEIC only). I confirmed the routing against the branch:

```
isImageAttachment("image/svg+xml", "chart.svg") == true
isImageAttachment("IMAGE/SVG+XML; charset=utf-8", "chart.svg") == true
```

That file would fail 100% of the time, and under item 2 the user is told it's gone. This repo has already made this exact call once: `server/internal/storage/util.go:105-120` carves SVG out of `image/*` on purpose, with a comment explaining why the prefix test alone is wrong.

The second half is the 10–30 MiB band. `:262` returns an error for an image over 10 MiB, so it's reported as a failed delivery — even though the file endpoint takes it up to 30 MiB. `wecomMediaKind` (`wecom/outbound_media.go:533-550`) guards on `size <= maxOutboundImageBytes` and falls through to `mediaTypeFile`, so an oversize image degrades to a file card instead of disappearing.

Suggested shape: an explicit allowlist of the formats Lark's image endpoint accepts, and everything else — SVG, >10 MiB images, unrecognized types — downgraded to the file path rather than failed. Tests covering at least SVG, the oversize image, and a format refusal.

Minor, but cheap to settle while you're here: the docs say 10 MB / 30 MB and the constants are `10 << 20` / `30 << 20`. If Lark means decimal MB, a file between 10,000,000 and 10,485,760 bytes passes the local check and is then refused upstream.

## 4. The new transport has no boundary tests

`server/internal/integrations/lark/http_client_media.go` is ~320 new lines carrying the multipart field names (`image_type`/`image`, `file_type`/`file_name`/`file`), the two endpoint paths, a token-refresh retry, the response envelopes, and reply-vs-send routing. All 10 tests in `outbound_media_test.go` fake `UploadImage`/`UploadFile`/`SendImageMessage`/`SendFileMessage`, so none of that is exercised — `git grep` confirms the four methods appear nowhere outside `outbound_media.go`'s interface and call sites. A typo in a field name compiles, passes CI, and fails only against a real bot.

`http_client_test.go` already uses `httptest` for this package, so the pattern is there to follow.

The credential side can't be checked statically either: the repo carries no scope manifest, so whether the app template actually holds `im:resource:upload` (or `im:resource`) is only answerable against a real app — both upload endpoints require one of those, and the app needs bot capability enabled ([image](https://open.feishu.cn/document/server-docs/im-v1/image/create), [file](https://open.feishu.cn/document/server-docs/im-v1/file/create)).

## One more thing, on the copy

The checklist marks Chinese product copy as checked, on the grounds that the two new notices are "English, matching every other user-facing string this adapter sends (`replier.go`, `outbound.go`)". That premise doesn't hold: `replier.go` is in the **wecom** package, not `lark`. Every user-facing string this adapter actually sends is Chinese — `outcome_replier.go:439-445` (all seven) and the binding card at `http_client.go:1349,1357`. `mediaSendFailedText` and `mediaLookupFailedText` would be the only English strings a Feishu user ever sees from this bot. I'd make them Chinese to match.

## Could you add a demo?

The PR notes this wasn't runnable without bot credentials, which is completely understandable. But this change's whole value is what lands in the chat window, and that's the one thing neither the diff nor the tests can show. If you can get a Feishu/Lark test bot pointed at this branch, a short screenshot or screen recording would help a lot — ideally covering: the text reply followed by the image, a non-image arriving as a file card, and (once item 3 is addressed) an SVG or an oversize image degrading to a file instead of vanishing. That would also be the natural place to confirm the upload scopes are in fact granted.

Happy to re-review once these are addressed — the hard part, working out where the hop belongs, is already done here.

PR #8340, opened 2026-09-12
MUL-7312 LARI-10: Desktop PATH fix + CodeBuddy Claude-parity stream-json guards

SeaSand1024:
## Summary

Two related fixes so local CodeBuddy behaves like other Multica IDE runtimes:

1. **Desktop PATH** (`apps/desktop/src/main/index.ts`): after `fix-path`, append only *missing* fallback dirs — never prepend `/usr/local/bin` over nvm/fnm Node. That was causing `#!/usr/bin/env node` CLIs (CodeBuddy / OpenClaw) to fail `--version` with Node 12 and land in `skipped_agents`.
2. **CodeBuddy headless background guard** (`server/pkg/agent/codebuddy.go`): Multica closes stdin after the first result and cannot wait for cross-turn `task_notification` events. Force the official `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1` (hides/`ignores` `run_in_background` before tool execution). As a belt-and-suspenders check, fail the run if CodeBuddy still emits `system` subtypes `task_started` / `task_progress` / `task_updated` / `task_notification` (the real background lifecycle in `@tencent-ai/codebuddy-code@2.150.0`).

## Intentional differences (not bugs)

- No `--strict-mcp-config` (CodeBuddy keeps user/project/local MCP scopes)
- Extra `--disallowedTools` for plan-mode tools (headless stall)
- Control response sends both `allowed` and `behavior` (CodeBuddy SdkPermissionClient)
- No Claude-style `terminal_reason=prompt_too_long` handling — that field is not emitted by CodeBuddy 2.150.0; failures use `is_error` + `errors` / `errors_info`

## Test plan

- [x] Local: CodeBuddy runtime online after PATH/host fix; smoke issue completed with `provider=codebuddy`
- [ ] CI: `backend-agent-tests` / CodeBuddy unit tests green on this PR
- [ ] Mac with nvm Node ≥18 and stale `/usr/local/bin/node`: Desktop daemon registers CodeBuddy (not skipped)
- [ ] Unit: `TestBuildCodebuddyEnvDisablesBackgroundTasks` forces env=1
- [ ] Unit: `TestCodebuddyExecuteFailsOnBackgroundTaskStarted` replays real `system/task_started` shape and fails (not completed)

Part of LARI-10 / SeaSand1024 fork CodeBuddy support.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -107,15 +107,21 @@ const BUNDLED_ICON_PATH = join(__dirname, "../../resources/icon.png").replace(
 // or any daemon-manager spawn.
 if (process.platform !== "win32") {
   fixPath();
-  // Fallback: prepend common install locations in case fix-path came up
-  // short (broken shell rc, non-interactive $SHELL, missing entries). Safe
-  // to duplicate — PATH lookups short-circuit on first match.
+  // Fallback: ensure common install locations are on PATH when fix-path came
+  // up short (broken shell rc, non-interactive $SHELL, missing entries).
+  // Append only missing dirs — never prepend. Prepending /usr/local/bin over
+  // a recovered login PATH shadows nvm/fnm Node with a stale system binary
+  // (e.g. Node 12), which breaks shebang CLIs (`#!/usr/bin/env node`) such as
+  // CodeBuddy and OpenClaw during daemon --version probes.
   const fallbackPaths = [
     "/opt/homebrew/bin",
     "/usr/local/bin",
     join(homedir(), ".local/bin"),
   ];
-  process.env.PATH = `${fallbackPaths.join(":")}:${process.env.PATH ?? ""}`;
+  const current = (process.env.PATH ?? "").split(":").filter(Boolean);
+  const existing = new Set(current);
+  const missing = fallbackPaths.filter((p) => !existing.has(p));
+  process.env.PATH = [...current, ...missing].join(":");
 }
 
 const PROTOCOL = "multica";
--- a/server/pkg/agent/codebuddy.go
+++ b/server/pkg/agent/codebuddy.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/multica-ai/multica/server/pkg/taskfailure"
 )
 
 // codebuddyBackend implements Backend by spawning the CodeBuddy CLI
@@ -201,6 +203,8 @@ func (b *codebuddyBackend) Execute(ctx context.Context, prompt string, opts Exec
 		var finalResultText string
 		sawResult := false
 		resultIsError := false
+		var terminalReasonError string
+		sawAsyncLaunch := false
 		var sessionID string
 		usage := make(map[string]TokenUsage)
 		eventCount := 0
@@ -241,7 +245,9 @@ func (b *codebuddyBackend) Execute(ctx context.Context, prompt string, opts Exec
 				}
 				lastAssistantText = turn.resolveFallback(lastAssistantText)
 			case "user":
-				b.handleUser(msg, msgCh)
+				if b.handleUser(msg, msgCh) {
+					sawAsyncLaunch = true
+				}
 			case "system":
 				if msg.SessionID != "" {
 					sessionID = msg.SessionID
@@ -251,6 +257,7 @@ func (b *codebuddyBackend) Execute(ctx context.Context, prompt string, opts Exec
 				sawResult = true
 				finalResultText = msg.ResultText
 				resultIsError = msg.IsError
+				terminalReasonError = codebuddyTerminalReasonFailure(msg.TerminalReason, msg.ResultText)
 				sessionID = msg.SessionID
 				if resultUsage := codebuddyResultUsage(msg, opts.Model); len(resultUsage) > 0 {
 					usage = resultUsage
@@ -286,6 +293,10 @@ func (b *codebuddyBackend) Execute(ctx context.Context, prompt string, opts Exec
 		// broken pipe, or been unblocked by the kill that ended cmd.
 		writeErr := <-writeDone
 
+		completionGuardError := ""
+		if sawAsyncLaunch {
+			completionGuardError = "codebuddy launched an async background task; Multica-managed runs require foreground execution"
+		}
 		finalStatus, finalOutput, finalError := finalizeStreamResult(
 			"codebuddy",
 			timeout,
@@ -294,17 +305,21 @@ func (b *codebuddyBackend) Execute(ctx context.Context, prompt string, opts Exec
 			exitErr,
 			sessionID,
 			streamTerminalState{
-				lastAssistantText: lastAssistantText,
-				finalResultText:   finalResultText,
-				sawResult:         sawResult,
-				resultIsError:     resultIsError,
-				scanErr:           scanErr,
+				lastAssistantText:   lastAssistantText,
+				finalResultText:     finalResultText,
+				sawResult:           sawResult,
+				resultIsError:       resultIsError,
+				scanErr:             scanErr,
+				terminalReasonError: terminalReasonError,
 			},
-			"",
+			completionGuardError,
 		)
 
+		// cmd.Wait() has returned — stderrBuf.Tail() is complete. Resume
+		// rejection phrases often land only on stderr (mirror Claude MUL-4966).
+		stderrTail := stderrBuf.Tail()
 		if finalError != "" {
-			finalError = withAgentStderr(finalError, "codebuddy", stderrBuf.Tail())
+			finalError = withAgentStderr(finalError, "codebuddy", stderrTail)
 		}
 		logStreamProtocolObservation(b.cfg.Logger, streamProtocolObservation{
 			provider:                   "codebuddy",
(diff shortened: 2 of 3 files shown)


Bohan-J (changes requested):
## Request changes

The Desktop PATH fix is correct and I'd take it as-is. The two CodeBuddy stream-json guards, however, key off fields that the shipped CodeBuddy CLI does not emit, so the "background task reported as success" failure mode this PR claims to close is still reproducible.

Direction is right — both problems are real and the fixes belong here. What needs changing is the protocol the CodeBuddy half is written against. No new product decision required.

Evidence base: static read of the published `@tencent-ai/codebuddy-code@2.150.0` package (published 2026-09-11, before this PR) plus a protocol replay. No live logged-in CLI run, no full Electron GUI run.

### 1. [P1] The background guard doesn't cover the path that actually executes

`server/pkg/agent/codebuddy.go:438` and `:466`.

- `forceClaudeToolInputForeground` only runs inside `handleControlRequest`. Under the default `bypassPermissions` mode, ordinary tools are allowed through without a permission round-trip, so there is no guarantee a `control_request` is ever emitted for the Bash call you're trying to rewrite.
- `claudeToolResultHasAsyncLaunch` looks for `status=async_launched`. That literal does not appear anywhere in `dist/codebuddy-headless.js` in 2.150.0. `StreamJsonMessageTransformers.transformToUserMessageFromFunctionResult` converts ordinary tool results into a text block even when the raw result is an object, so `content.status` isn't there to read.
- The real background lifecycle is carried by `system` events — `task_started`, `task_updated`, `task_notification`. This PR's `system` branch only reads `session_id` and ignores those.

Replay of `init → Bash tool_use(run_in_background=true) → system/task_started → text tool_result → success result` (no task-completion event) still yields `status="completed"`, `output="Started the build in the background."`, empty error.

Suggested fix: use the officially supported `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1` — the 2.150.0 package ships the corresponding Bash/PowerShell/Agent handling, and it disables the background path *before* tool execution, which fits this adapter's "close stdin after the first result" lifecycle. Please add a regression test built on the actual event shapes rather than the Claude ones.

Ref: https://www.codebuddy.ai/docs/cli/headless

### 2. [P2] The context-exhaustion branch has no applicable upstream signal

`server/pkg/agent/codebuddy.go:260` and `:552`.

`terminal_reason` and `prompt_too_long` have zero occurrences in the 2.150.0 published bundle — there is no code path that produces them. CodeBuddy's failure result construction uses `is_error=true` with `errors` / `errors_info`; the success path uses `is_error=false` with `result`.

The `terminal_reason` precedent (#6422) came from Claude Code 2.1.220/2.1.221, which isn't evidence about CodeBuddy's protocol. `TestCodebuddyExecuteFailsOnPromptTooLong` hand-assembles a Claude-shaped frame, so it passes without demonstrating anything about CodeBuddy.

Please either supply a real context-exhaustion sample from a named CodeBuddy version and implement against whatever signal it actually carries, or drop this branch along with the corresponding claim in the PR description.

### What I verified

- Targeted Go tests over `pkg/agent` (CodeBuddy + shared result handling) pass; PATH-ordering checks pass.
- PATH change replayed through Node `vm` against both baseline and PR code with a fixed `fixPath` return: nvm precedence preserved, only missing dirs appended, no duplicate `/usr/local/bin`, fallback still added when PATH is unset. Electron GUI not launched; no real `--version` probe against CodeBuddy/OpenClaw.
- Reverting only this PR's `stderrTail` argument to the two resume helpers via a Go overlay leaves `TestCodebuddyExecuteResumeRejectedFromStderr` passing — existing `withAgentStderr` already folds the tail into `finalError`, so that test isn't a regression guard for new resume behavior.
- Existing CI (backend-agent-tests, backend-tests) is green; it does not cover either protocol gap above.

### Scope / rollback

PATH change affects every CLI subprocess spawned from macOS/Linux Desktop; the guards affect all CodeBuddy daemon tasks. No migration. If wrong, it shows up as changed CLI version selection or misreported task status — revert the relevant commit and restart Desktop/daemon.

SeaSand1024:
Thanks @Bohan-J — addressed in f75c46f3b per your review:

1. **P1**: force `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1`, fail on real `system/task_*` background events, drop Claude `async_launched` / control-request rewrite. Regression: `TestCodebuddyExecuteFailsOnBackgroundTaskStarted`.
2. **P2**: dropped `prompt_too_long` / `terminal_reason` branch and PR claim (not in CodeBuddy 2.150.0).

PATH commit left as-is.

Bohan-J (changes requested):
## Request changes

Both findings from the previous round are properly fixed — I verified them rather than taking the description's word for it. One new blocking issue remains, in the same function the round-1 fix introduced.

### Verified fixed

- **Background guard.** Replaying the 2.150.0 background shape (`init` → `Bash tool_use(run_in_background=true)` → `system/task_started` → text `tool_result` → success `result`) against both commits with the same test: `12424cbb` returns `status="completed"`, `1a9589a3` returns `status="failed"` with a readable reason. The forced env var also genuinely reaches the child — a fake CLI echoing the variable back reports `1` even when the caller passes `0`.
- **Choice of mechanism.** `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS` is the right lever, and the bundled headless docs confirm why: it hides `run_in_background` from the Bash/PowerShell/Agent tool schema and degrades it to foreground if the model sends it anyway, and the vendor's own SDK `query()` injects the same variable for exactly this lifecycle (stops at the first `result`, cannot receive cross-turn push-back). That matches this adapter.
- **No false-positive risk in the subtype guard.** All four subtypes exist in the shipped 2.150.0 bundle, and `emitTaskEvent` is fed only from `getBackgroundTasks()` / `backgroundTaskSubject` — so a normal foreground run using the Task/Agent tool cannot trip it.
- **`terminal_reason` branch** and its field are gone, and the PR description now states the correct upstream behavior instead of claiming a fix.

### [P1, blocking] The forced disable can be overridden on Windows

`server/pkg/agent/codebuddy.go:538` and `:541`.

`buildCodebuddyEnv` prepends `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1` and then drops duplicates with an exact, case-sensitive comparison. Windows environment keys are case-insensitive, and `os/exec` dedups with `dedupEnvCase(runtime.GOOS == "windows", …)`, which lowercases the key and keeps the **last** occurrence.

So a lowercase key — from the daemon's inherited environment, or from an agent's `custom_env` — survives the filter. Running the current code with `custom_env` = `codebuddy_code_disable_background_tasks: "0"` produces both entries:

```
CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1     ← forced, first
codebuddy_code_disable_background_tasks=0     ← survives the filter, last
```

On Windows the child then receives `0`. Background tasks can start again; the `system/task_*` guard turns that into a failure instead of a false success, so no work is silently lost — but the PR's stated goal (keep execution in the foreground) is not met, and the user gets a failed task instead of a completed one.

Suggested fix — stop hand-rolling platform env semantics and let `os/exec` do its own dedup, which is already platform-correct:

```go
func buildCodebuddyEnv(extra map[string]string) []string {
	return append(buildEnv(extra), codebuddyDisableBackgroundTasksEnv+"=1")
}
```

Please add a regression test with a lowercase key alongside the existing one.

### [nit] The comment's rationale is what led here

The same block says *"Unix getenv is first-wins; put our forced value first"*. That is not the mechanism in play: `os/exec` dedups before `exec`, keeping the last value, so ordering alone never protects the forced value — the explicit filter does, and only on case-sensitive platforms. Worth correcting whichever way the fix goes, since this comment is what makes the hand-rolled dedup look sufficient.

### [nit, not blocking] The PATH change has no test coverage

`apps/desktop/src/main/index.ts` is a module-level side effect with no `index.test.ts`, while `apps/desktop/src/main/` has a dozen other `.test.ts` files. The PATH computation itself is pure — extracting it to a function taking the current PATH plus the fallback list, with cases for "all dirs present", "one missing" and "PATH empty", would be cheap and would cover a block that affects binary resolution for every agent CLI the Desktop-launched daemon spawns, not just CodeBuddy. Pre-existing gap, not introduced here, but this PR is the natural place to close it.

### Verification scope

Targeted and full-package `go test ./pkg/agent` pass on `1a9589a3` (macOS arm64). The Windows behavior above is derived from the `os/exec` source and a direct reproduction of the precondition (both entries reaching `cmd.Env`); no Windows machine was used. No live logged-in CodeBuddy CLI run and no full Electron GUI run — the CodeBuddy protocol evidence is static analysis of the published 2.150.0 package plus replay. The PATH commit is unchanged since the last round.

SeaSand1024:
Thanks @Bohan-J — addressed in 1fa3ca811:

1. **P1 Windows env**: `buildCodebuddyEnv` now `append(buildEnv(extra), …=1)` so `os/exec` last-wins (case-insensitive on Windows) keeps the force. Regression covers lowercase `codebuddy_code_disable_background_tasks=0`. Comment updated.
2. **PATH nit**: extracted `appendMissingPathDirs` + `path-fallback.test.ts` (all present / one missing / empty / never-prepend).

Bohan-J (approved):
## Approve

Both rounds of feedback are fully addressed, and I verified the final commit rather than relying on the description. Thanks for the thorough follow-through here — particularly for replacing the hand-rolled env dedup with the stdlib's instead of patching around the symptom, and for picking up the PATH test suggestion that wasn't blocking.

### Verified on `1fa3ca81`

- **Windows override is fixed.** `buildCodebuddyEnv` now appends the forced entry last and lets `os/exec` do its own platform-aware dedup. Running the current code with `custom_env` = `codebuddy_code_disable_background_tasks: "0"` puts the entries in the order `…=0`, then `CODEBUDDY_CODE_DISABLE_BACKGROUND_TASKS=1` — so the case-insensitive, last-wins dedup on Windows now resolves to `1`. That is the inverse of the previous ordering, and the corrected comment states the real mechanism.
- **The PATH extraction is behaviour-preserving.** I diffed `appendMissingPathDirs` against the pre-extraction inline implementation over 262 PATH inputs (all subsets of a representative dir set in both orderings, plus empty / trailing-separator / repeated-separator edge cases): zero mismatches. The four behaviours the new test file asserts hold independently.
- **No regression in the earlier guards.** The background replay still returns `failed`, the resume-rejection and subtype tests pass, and the full `go test ./pkg/agent` package is green locally (macOS arm64).
- CI is green on this commit, including `windows-execenv`.

### Known scope, recorded for the merge

The PATH commit affects command resolution for every runtime launched from a macOS/Linux Desktop daemon, not only CodeBuddy — that is the intended fix, since the previous prepend was overriding the user's own login-shell ordering and making Desktop-launched daemons resolve binaries differently from terminal-launched ones. One consequence worth knowing: when shell recovery fails entirely and only a minimal PATH remains, system tools can now take precedence over Homebrew ones. Rollback is reverting the relevant commit and restarting Desktop/daemon; no migration involved.

Not covered by any of this: a live logged-in CodeBuddy CLI run, a full Electron GUI run, and a real Windows machine — the Windows reasoning comes from the `os/exec` source plus a direct reproduction of the ordering.

PR #8342, opened 2026-09-12
ZIC-224 Fix quick create preview-only completion

### LINKED ISSUE #8328: [Bug]: Quick Create treats a successful preview-only response as a failure and hides the agent output
### Deployment type

self-host

### Area

Web app

## Summary

When “Create with agent” is asked to draft an issue without creating it, the agent can complete successfully and return the requested draft, but Multica reports the operation as failed:

> Quick create failed, but the agent's output was too large to show the reason safely. Check the task's execution log for details.

In this case, the agent did not fail. It followed the instruction correctly and returned a complete issue draft. The failure appears to be caused by Quick Create assuming that every successful run must create an issue.

The useful response is then hidden because it exceeds the failure-detail length limit.

## Expected behavior

One of the following would be appropriate:

* If preview-only responses are supported, show the agent's draft and mark the run as completed without creating an issue.
* If Quick Create must always create an issue, reject preview-only instructions before starting the run and explain that the user should use regular chat or manual issue creation instead.

In either case, a successfully completed agent response should remain accessible.

## Actual behavior

The agent completes successfully and returns the requested draft, but the UI reports:

> Create with agent failed
> Quick create failed, but the agent's output was too large to show the reason safely.

No draft is displayed in the UI.

## Execution evidence

The daemon logs show that the provider run completed successfully:

```text
status=completed
result_is_error=false
result_bytes=2534
agent_error=""
```

The beginning of the returned result is also visible in the debug log:

```text
Here is the issue draft. I have not created it yet.
```

This confirms that the provider produced the expected result and that the failure occurred while interpreting the Quick Create outcome.

## Environment

* Multica: self-hosted `v0.4.42`
* Provider: Cursor Agent
* Cursor Agent CLI: `2026.09.10-fd3934a`
* macOS
* Project resource: `github_repo`

## Suspected cause

The Quick Create completion path appears to use the existence of a newly created issue as the success condition. A completed task that returns text but creates no issue is therefore routed through the failure notification path.

The related logic appears to be in:

`server/internal/service/task.go`

Relevant functions/constants appear to include:

* `notifyQuickCreateCompleted`
* `notifyQuickCreateFailed`
* `quickCreateFailureDetail`
* `maxQuickCreateFailureDetailRunes`

The current assumption that “no issue was created” means “the agent failed” does not hold when the agent was explicitly asked to return a preview only.

## Suggested improvement

It may help to represent these outcomes separately:

1. Agent execution succeeded or failed.
2. An issue was created or was not created.

For a successful run with no created issue, Multica could:

* display the returned response;
* use a neutral state such as **Completed without creating an issue**;
* provide a button to create an issue from the returned draft.

At minimum, the full agent output or task transcript should remain accessible from the Quick Create result instead of being replaced by the oversized-failure message.

## Steps to reproduce

1. Open a Multica project connected to a repository.

2. Select **Create with agent**.

3. Enter a request such as:

   > Draft an issue for establishing the architecture baseline. Do not create or start the issue. Show me the draft first.

4. Wait for the agent run to finish.

### Screenshots (optional)

_No response_

```shell

```

vicksiyi:
## What does this PR do?

Fixes Quick Create completion handling when the agent successfully returns a draft/preview but does not create an issue. The task completion path now treats "agent completed, no created issue found" as a completed no-issue result and stores the returned output on a `quick_create_done` inbox item instead of reporting `quick_create_failed`.

The inbox detail label now shows the returned body for this no-issue success case, while preserving the existing created-issue identifier label for normal quick-create success.

## Related Issue

Closes #8328

Multica issue: ZIC-224

## Type of Change

- [ ] CI / infrastructure

## Changes Made

- `server/internal/service/task.go`: pass completion output into quick-create reconciliation and write a no-issue success inbox row for completed preview-only runs.
- `server/internal/service/task_notify_test.go`: add a lifecycle regression for completed quick-create output with no created issue.
- `packages/views/inbox/components/inbox-detail-label.tsx`: show the returned body for no-issue `quick_create_done` rows.
- `packages/views/inbox/components/inbox-detail-label.test.tsx`: cover no-issue success rendering and existing identifier rendering.
- `apps/mobile/components/inbox/detail-label.tsx`: mirror the no-issue success body fallback.

## How to Test

1. `set -a; . ../.env.worktree; set +a; go test ./internal/service -run 'TestCompleteQuickCreateWithoutCreatedIssueWritesSuccessWithOutput'`
2. `pnpm --dir packages/views exec vitest run inbox/components/inbox-detail-label.test.tsx`
3. `make check-worktree`

Note: `make check-worktree` exited 0, but printed that Docker was unavailable while checking PostgreSQL: `Cannot connect to the Docker daemon at unix:///Users/icezero/.docker/run/docker.sock`.

## Checklist

## AI Disclosure

**Prompt / approach:**
Traced the quick-create completion path from `CompleteTask` into `notifyQuickCreateCompleted`, separated successful agent execution from created-issue discovery, added a focused backend regression, and updated inbox rendering so the returned draft remains visible.

## Screenshots (optional)

N/A

### DIFF
--- a/apps/mobile/components/inbox/detail-label.tsx
+++ b/apps/mobile/components/inbox/detail-label.tsx
@@ -138,9 +138,8 @@ export function InboxDetailLabel({
           ? `Reacted with ${details.emoji}`
           : TYPE_LABEL[item.type];
       case "quick_create_done":
-        return details.identifier
-          ? `Created with agent: ${details.identifier}`
-          : TYPE_LABEL[item.type];
+        if (details.identifier) return `Created with agent: ${details.identifier}`;
+        return singleLine(item.body) || TYPE_LABEL[item.type];
       case "quick_create_failed": {
         const detail = singleLine(details.error) || singleLine(item.body);
         return detail ? `Failed: ${detail}` : TYPE_LABEL[item.type];
--- a/packages/views/inbox/components/inbox-detail-label.test.tsx
+++ b/packages/views/inbox/components/inbox-detail-label.test.tsx
@@ -1,0 +1,78 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import type { InboxItem } from "@multica/core/types";
+import { InboxDetailLabel } from "./inbox-detail-label";
+
+vi.mock("../../issues/components", () => ({
+  StatusIcon: () => null,
+  PriorityIcon: () => null,
+}));
+
+vi.mock("@multica/core/workspace/hooks", () => ({
+  useActorName: () => ({ getActorName: () => "Someone" }),
+}));
+
+vi.mock("../../i18n", () => ({
+  useT: () => ({
+    t: (accessor: (dict: unknown) => string, params?: Record<string, string>) => {
+      const template = accessor({
+        labels: {
+          created_with_agent: "Created with agent: {{identifier}}",
+          failed_with_detail: "Failed: {{detail}}",
+        },
+        types: {
+          quick_create_done: "Quick-create done",
+          quick_create_failed: "Quick-create failed",
+        },
+      });
+      if (!params) return template;
+      return template.replace(/\{\{(\w+)\}\}/g, (_, key: string) => params[key] ?? "");
+    },
+  }),
+}));
+
+function item(overrides: Partial<InboxItem> = {}): InboxItem {
+  return {
+    id: "inbox-1",
+    workspace_id: "workspace-1",
+    recipient_type: "member",
+    recipient_id: "member-1",
+    actor_type: "agent",
+    actor_id: "agent-1",
+    type: "quick_create_done",
+    severity: "info",
+    issue_id: null,
+    title: "Quick create completed",
+    body: null,
+    issue_status: null,
+    read: false,
+    archived: false,
+    created_at: "2026-09-12T08:00:00Z",
+    details: null,
+    ...overrides,
+  };
+}
+
+describe("InboxDetailLabel quick-create success", () => {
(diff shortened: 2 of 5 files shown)


Bohan-J:
Thanks for picking this up, and for adding tests on both sides of the stack. We're not going to merge it, though.

**Product reason:** the change makes "agent completed, no issue created" a success outcome, and we've decided against adding a preview-only outcome to Quick Create — reasoning is in #8328. The most common real cause of that state isn't a preview request, it's a rejected `multica issue create` (duplicate guard, validation error). With this change, a duplicate rejection would render as a `quick_create_done` success row whose body is the error text.

**Mechanics, in case any of it is useful next time:**

- The branch is based on a commit from June 30 and is roughly 1439 commits behind `main`; GitHub already reports it as conflicting. Worth rebasing before writing the fix — some of what this PR adds already exists upstream.
- `main` already passes the completion result into `notifyQuickCreateCompleted` and extracts the CLI error as the user-facing failure reason via `quickCreateFailureDetail`.
- `main` also has a dedicated `quick_create_unconfirmed` inbox type for the branch where the issue lookup itself fails, specifically so an unverified outcome isn't labelled `Failed:`. This PR routes that branch back into `notifyQuickCreateFailed`, which would undo the distinction.
- The new no-issue row stores the whole agent output in the inbox `body` with no length bound, and the list label renders `body` directly. A runaway output would land in the inbox list — the failure mode the existing rune cap on the failure detail exists to avoid.

Closing this one. Thanks again for the contribution, and sorry for the wasted effort on a direction we ended up declining.

PR #8344, opened 2026-09-12
fix(wecom): send a long answer in pieces instead of losing it

seacen:
## Problem

An `aibot_send_msg` body is capped at 20480 utf8 bytes ([101138](https://developer.work.weixin.qq.com/document/path/101138)), and WeCom refuses anything past it **whole** — it does not clip. So an agent's answer over that size is answered `45002` and never appears in the chat.

From the person's side: they ask a question, nothing comes back, and there is no error and no partial answer to go on. The only record is a log line. A code review, a pasted log and a document draft all reach that size routinely, so this is not an edge case for an agent that writes for a living.

## Change

`sendTextCtx` cuts the body into pieces the server will accept.

- The cut prefers a line break in the last quarter of the budget, then a rune boundary — a piece never ends mid-character and rarely ends mid-line.
- Every piece but the last carries a bare `(n/total)` counter. This is the one place the adapter adds words to an agent's own text, which is why it is a counter rather than a sentence: it belongs to no language, so it needs no translation and cannot contradict an answer written in one.
- An answer that already fits is returned untouched and allocates nothing — the common path is unchanged.

It lives in the sender rather than at a call site because the cap applies to **every** caller that pushes plain text — the reply, an inbox card, a relayed frame — and the one that forgot the rule would lose its message silently.

## The one subtle part: a piece that fails after an earlier one landed

A failing piece stops the rest (the pieces after it are the tail of an answer whose head never arrived, and sending them alone reads as the bot replying to nothing).

A failure past the **first** piece is wrapped in `errPartiallySent`, because the caller's question — may this send be tried again? — has a different answer once part of the answer is in the chat:

- `provablyNotSent` (the relay's retry gate) says no, so a routed reply is not re-offered and re-printed;
- the reply path counts it delivered and logs that only part arrived, rather than recording a drop an operator would read as "resend it".

WeCom has no unsend, so printing the opening twice is the outcome worth avoiding.

## Tests

They drive the real sender against a connection that enforces the server's own rule — a body past the cap is answered `45002` and written nowhere — so a test cannot pass by writing a frame nobody would have seen:

- a long answer is readable end to end after the markers are stripped;
- a short answer is untouched, with no extra frame and no counter appended;
- each piece says the answer continues, and the last one does not;
- no piece ends mid-character (multi-byte runes throughout, on purpose);
- the cut prefers a line boundary when there is a usable one.

Reverse-verified: with the split removed, the first test fails exactly the way production does — `errcode=45002 content exceed max length`, nothing delivered.


Bohan-J (changes requested):
The product direction is right, and I'd like to see this land. WeCom refuses an oversized `aibot_send_msg` body **whole** (20480 utf8 bytes, errcode 45002), so today a long answer never reaches the chat at all — and splitting inside the adapter rather than at a call site is already the house pattern here (`slack/channel.go` `chunkMessage`, `telegram/sender.go` `chunkMessage`). A reply that already fits stays a single untouched frame, which is the right default.

Three things in the current implementation break the "complete, in order" promise the PR makes. All three reproduce on `e3ad38e`; the five tests this PR adds pass while all three are live.

## 1. Cutting at a line boundary deletes the newline

`wireCutPoint` returns the index **of** the `\n` (`ws_frame.go:793`), and `splitForWire` then does `strings.TrimLeft(remaining[cut:], "\n")` from that same index (`ws_frame.go:769`). So the newline at every seam disappears, and a run of them can go too. After stripping the counters, `lineA\nlineB` reassembles as `lineAlineB`.

Reassembling the pieces (markers stripped, same way `reassemble` in the new test does it):

```
newline fixture:    orig 44160 bytes, got 44158 bytes, 3 pieces — 2 bytes missing
blank-line fixture: orig 40950 bytes, got 40948 bytes, 3 pieces — 2 bytes missing
```

This contradicts the invariant the test file states in its own header ("whatever the agent wrote, the person can read all of it"), and it bites hardest on exactly the content the PR is aimed at: pasted logs and code, where a swallowed paragraph break changes how the text reads.

The existing tests miss it for a specific reason worth fixing together with the bug:

- `TestALongAnswerReachesTheChatWhole` is the only test that reassembles, and its fixture (`aLongAnswer`) contains no newlines, so the line-boundary branch never runs.
- `TestTheCutPrefersALineBoundary` exercises that branch but only checks line lengths *within* each piece — it never reassembles.

Please keep the newline on one side of the cut, and add a byte-exact reassembly test over a fixture with single newlines, blank-line runs, and multi-byte runes.

## 2. The same partial failure is recorded as delivered or as dropped depending on deployment topology

- Direct path (`outbound.go:288-301`): `errPartiallySent` is logged, then `o.delivered()` runs.
- Relay path (`relay_outbound.go:1252-1290`): `provablyNotSent` returns `false` — correct, no retry — but execution then falls through to `record`, where `unconfirmedReason(sendErr)` / `droppedFor(...)` fires. A wrapped `*wecomAPIError` isn't in `unconfirmedReason`'s switch (`outbound_outcome.go:114-126`), so a refused second piece lands on `droppedFor(dropPlatformRefused)`.

Same user-visible event — piece 1 in the chat, piece 2 refused — counts as `outbound_delivered` on a single-replica deployment and `outbound_dropped` once the frame goes through the relay. That's the "one reply counted as delivered and dropped at the same time" class of defect the comment block right above that code says this package no longer has, and it means a partial delivery pages someone or doesn't depending on which replica held the socket.

Please define the partial-send outcome in one place and have both paths consume it, with paired tests on each path for a refused second piece and for a lost ack.

## 3. The pieces of one answer are not a contiguous unit on the wire

`sendTextCtx` calls `sendOneTextCtx` once per piece (`ws_sender.go:310-323`), and `request` releases `s.mu` before waiting for each ack (up to `ackTimeout`, 5s). `s.mu` orders a single *frame write*, not a logical message, and nothing above `wsSender` serializes the several independent callers of `sendText`/`sendTextCtx` (agent reply, inbox push, media receipt, unsupported-type receipt, relay).

One long answer plus one unrelated send on the same connection, with a 30ms ack:

```
frame 1: "答答…" (len 20476)   ← (1/3)
frame 2: "INTERLOPER" (len 10)
frame 3: "答答…" (len 20476)   ← (2/3)
frame 4: "答答…" (len 1518)    ← (3/3)
```

The reader sees an unrelated message wedged into the middle of one answer. With two long answers in flight at once, the `(n/total)` counters can't even be matched back to their own text — which is worse than the status quo, where a long answer simply didn't arrive.

Please serialize user-visible logical messages — per target chat is probably enough and keeps the ping loop unblocked — and pin wire order with a delayed-ack + concurrent-send test.

## What I verified

- `git diff --check`: clean.
- The five tests added by this PR: `go test -race` passes.
- The three repros above: written against this branch, all three fail.
- `go test -race ./internal/integrations/wecom ./cmd/server` as a whole: couldn't complete — my local test database is missing current-schema objects (`context_revision`, `channel_task_delivery`, `dingtalk_group_presence`). This PR carries no migration and those failures look unrelated to the diff.
- CI is green, but none of the three paths above is covered by a test, and there's no real WeCom end-to-end run behind this.

Scope note: this changes every WeCom text send by default, so if the split is wrong it shows up as missing content, wrong order, or misreported delivery metrics rather than as an error. Rollback is a revert of this commit and is confined to over-cap WeCom messages.

seacen:
All three are fixed. Each one has a test that fails without the fix, and I've quoted those failures rather than describing them.

## 1. The seam ate a line break

`wireCutPoint` returned the index OF the newline and `splitForWire` then trimmed leading newlines from that same index. The trim is gone and the cut goes after the break, so the break stays with the line it ended.

Two properties, two tests, because keeping every byte and putting the break on the right side of the seam are not the same claim:

- `TestReassemblingThePiecesGivesBackEveryByte` — byte-exact reassembly over four fixtures: single newlines, blank-line runs, multi-byte runes either side, a break landing exactly at the budget, and one that is nothing but blank lines. Reverse verification with the trim restored: `reassembled 40947 bytes from 3 pieces, want the original 40950 — they part at byte 20474`.
- `TestNoPieceOpensWithTheBreakThatEndedTheLastOne` — reverse verification with `return nl`: `piece 2 opens with "\n连续文\xe5\xad" — the break that ended piece 1 fell into the gap`.

Worth saying: only the trim was losing bytes. `nl` vs `nl+1` is a readability property, and it needed its own test to be load-bearing at all — with the trim removed, the first test passes either way.

`TestTheCutPrefersALineBoundary` is updated: it asserted a piece does NOT end in a break, which is the behaviour that changed.

## 2. The same partial send counted two ways

`recordSend` (outbound_outcome.go) is now the single definition, and both paths consume it — `processEvent` records rather than returning, so `handleChatDone` cannot classify the same send a second time.

A partial send counts as delivered and WARNs with what did not land. The reasoning is in the comment: the action a drop invites is a resend, a resend prints the first piece again, and WeCom has no unsend. I considered a separate counter and did not add one — a partial send is visible in the log and not in `/metrics`. If you want it countable, say so and it is a small addition.

Four tests, a refused second piece and a lost verdict on each path: `TestDirectPath_ARefusedSecondPieceCountsDelivered`, `TestDirectPath_ALostAckOnTheSecondPieceCountsDelivered`, `TestRelayedReply_ARefusedSecondPieceCountsDeliveredAndIsNotReoffered`, `TestRelayedReply_ALostAckOnTheSecondPieceCountsDelivered`.

Reverse verification, dropping the `errPartiallySent` arm, reproduces exactly the split you described:

```
--- FAIL: TestDirectPath_ARefusedSecondPieceCountsDelivered
    outbound_delivered = 0, want 1
    outbound_dropped = 1, want 0
--- FAIL: TestRelayedReply_ALostAckOnTheSecondPieceCountsDelivered
    outbound_unconfirmed = 1, want 0
```

## 3. The pieces were not contiguous

`sendTextCtx` now holds a lock across the whole logical message, keyed on the target chat. Per chat rather than per socket: an answer to one room must not hold up another, and the ping loop takes no chat lock. The lock is created on demand and dropped by its last holder, so a process that has talked to many chats keeps no entry for each.

Held for every send, not only a split one — a single-frame push from another caller is exactly what used to land between piece one and piece two.

`TestThePiecesOfOneAnswerAreContiguousOnTheWire` uses a 30ms ack delay and a concurrent unrelated send. Reverse verification without the lock: `an unrelated message landed at position 2, between pieces of one answer`.

`TestAnAnswerToOneChatDoesNotHoldUpAnother` is the guard against over-fixing. Reverse verification with the lock keyed on a constant: `the answer to one chat finished before a push to another got through — the lock is not per chat, it is the whole socket`.

## One thing I found and did not fix here

`RelayConfig.DeliveryBudget` is documented as the bound on "the send and its ack wait", and `outcomeGrace` adds it as the last offer's own delivery. It is never applied: `perform` hands `deliverRelayed` the dispatcher's context, so the send's only bound is `ackTimeout`, the constant. A deployment that lowers the budget shortens the grace without shortening the delivery it is waiting on, and a Resolve landing inside an ack wait fences a reply that is on its way.

I found it writing the relay test above — a 120ms budget still took five seconds. The fix is three lines plus a test, and it is ready on a branch off `main`; I have kept it out of this PR because it touches `relay_outbound.go`, which this PR also touches. I will open it as soon as this merges, or now if you would rather have it first.

Rebased on `27d0529fa`. #8347 landing put `hasVisibleChar` on `main`, so the send guard here is now `hasVisibleChar(content)` rather than `content != ""`. `go test -race` green on `./internal/integrations/wecom ./cmd/server`.

multica-eve (changes requested):
Requesting changes on `e3be833`: the product direction remains right, and the newline preservation plus direct/relay partial-send accounting issues from the first review are fixed, but the current implementation still does not make one user-visible reply contiguous across all send types and introduces a definitely-unsent timeout path that is treated as possibly delivered.

The intended user behavior is still worth landing: an answer beyond WeCom's 20,480-byte markdown limit should arrive completely, in ordered pieces, while a short answer remains untouched. The current head closes the byte-reassembly and topology-dependent accounting gaps, and its text-to-text per-chat locking tests pass. Three correctness blockers remain.

## 1. Media can still interleave with a split text reply in the same chat

`sendTextCtx` acquires the per-chat lock at `ws_sender.go:390`, but `sendMedia` calls `request` directly at `media_upload.go:323-328`. Attachment delivery is asynchronous (`outbound_media.go:223-227`), so this is not theoretical: with a 30 ms delayed ACK, I reproduced this wire order in one chat:

```
text piece 1 -> media -> text piece 2
```

That still breaks the PR's “one answer arrives as a contiguous sequence” promise. The comment above `sendTextCtx` says the lock is held for every send and specifically names media, but the media path bypasses it.

Please make all user-visible `aibot_send_msg` sends for a chat share the same logical-message lock (at minimum text and media), and add a delayed-ACK text-vs-media concurrency test. The lock should cover the chat push, not unrelated upload work.

## 2. Timing out while waiting for the new chat lock drops a message without retry

`chatLocks.acquire` returns a bare `ctx.Err()` at `ws_sender.go:152-157`. At that point no frame has been built or written. However:

- `unconfirmedReason` classifies that error as `interrupted` (`outbound_outcome.go:118-123`), so the direct path records an unknown delivery rather than a definite transport failure.
- `provablyNotSent` returns false for it (`relay_outbound.go:1358-1359`), so the relay settles the claim and does not re-offer a message that definitely never reached the socket.

I held the chat lock, gave a second send a 10 ms deadline, and observed zero wire frames while the current classifiers returned `provablyNotSent=false` and `unconfirmedReason="interrupted"`. The user receives nothing, and the retry owner is told not to retry.

Please wrap lock-acquisition cancellation in a distinct before-write/not-attempted sentinel, check it before generic context errors in both classifiers, and add direct-outcome plus relay-reoffer regression tests.

## 3. `DeliveryBudget` must bound the complete logical relay delivery in this PR

`RelayConfig.DeliveryBudget` is documented as the maximum duration of one claimed delivery, including send and ACK, but it currently only sizes `outcomeGrace` (`relay_outbound.go:1062-1066`). `perform` passes its existing context straight to `deliverRelayed` at `relay_outbound.go:866`; the configured budget never bounds the operation.

This can no longer be treated as unrelated follow-up work. Before splitting, one text send waited for at most one ACK. This PR makes one logical send wait for the chat lock and then for N sequential ACKs, while `outcomeGrace` still reserves only one delivery budget. A slow multi-piece send can therefore outlive the publisher's grace; `Resolve` may fence the holder while it is still sending, and the holder's later `Settle`/outcome record is then refused.

Please land the prepared budget fix before rebasing this PR, or include it here. One budget should cover the whole logical delivery (chat-lock wait plus all pieces), with a slow multi-piece ACK test proving that publisher grace cannot expire before that bound.

## Validation scope

- 15 added/related WeCom tests pass under `go test -race` on this head.
- Two focused review reproductions consistently demonstrate blockers 1 and 2; those temporary tests were removed after verification.
- `git diff --check` passes, and the current GitHub checks are green.
- I did not run a real WeCom end-to-end test.

There are no schema or configuration changes. If the split behavior is wrong in production, it affects every over-limit WeCom text reply by default; rollback is a revert of this PR.

seacen:
All three are fixed, each with a test written before the change and reverse-verified after it.

## 1. Media no longer lands between the pieces

`sendMedia` takes the same per-chat logical-message lock `sendTextCtx` does, so both producers of an `aibot_send_msg` in this package are behind one door. The lock covers the chat push only — `uploadMedia` stays outside it, so a multi-megabyte upload does not serialize unrelated messages.

`TestAFileDoesNotLandBetweenThePiecesOfOneAnswer` reproduces your wire order with a 30ms delayed ACK. Before the fix:

```
"image MEDIA_1" landed at position 2, between pieces of one answer:
  [20476 bytes ending … (1/4) | image MEDIA_1 | 20476 bytes ending … (2/4) | …]
```

## 2. Giving up on the chat's turn is a definite non-delivery

`chatLocks.acquire` now returns `errChatBusy` wrapping the context error, and all three classifiers check it before the generic context arm: `unconfirmedReason` → `""`, `provablyNotSent` → `true`, and `sendOutcome` → `deliveryDefinitelyFailed` (that third one because media now takes the same lock).

Two tests, one per path — the direct outcome and the relay re-offer. Before the fix, with your numbers:

```
provablyNotSent(context deadline exceeded) = false, want true
unconfirmedReason(context deadline exceeded) = "interrupted", want ""
outbound_dropped = 0, want 1 — the reply is definitely not delivered
```

and on the relay path, `timed out waiting for the reply to reach the chat on a later offer`.

One thing that came out of building it: the first version of the sentinel made `TestDeliverRelayed_ContextErrorsDoNotRelease` non-deterministic. With an already-cancelled context and a free lock, Go's `select` picks at random between taking the lock and taking `ctx.Done()`, so the same situation produced two different verdicts. `acquire` now tries the lock with a non-blocking select first, so `errChatBusy` means what it says — the chat was genuinely held — and that test is deterministic again at `-count=20`.

## 3. DeliveryBudget bounds the whole logical delivery, in this PR

The prepared fix is folded in: `perform` wraps `deliverRelayed` in `context.WithTimeout(ctx, cfg.deliveryBudget())`, so one budget covers the chat-lock wait and every piece. `TestASlowMultiPieceDeliveryFitsInTheOneBudgetTheGraceReserves` holds the chat 60ms, sends three pieces at 40ms an ACK with the third never answered, and asserts the delivery ends inside the budget and that `outcomeGrace()` exceeds what it took. Before the budget was applied: `timed out waiting for the delivery to end inside one budget rather than on the ack timeout of one piece`.

Three comments that were describing the old behaviour are corrected with it, including `outcomeGrace`'s "The send waits at most ackTimeout".

Worth flagging, because it is the kind of thing that hides: my first version of that relay test was passing for the wrong reason. It was written against code where the budget was not yet applied, so the delivery context had no deadline at all and the test simply waited for the lock and succeeded. Reordering the commits so the budget lands first — and giving the test a chain long enough to matter — is what made its reverse verification actually red.

Rebased on `a843b44a4`. `go vet` clean under both tag sets, `go test -race` green on the touched packages, and the DB-backed tests run rather than skip.

Bohan-J (changes requested):
Reviewed at head `ba50fef`. Product direction is still a clear accept, and the three items from the last round are genuinely closed: media now takes the same chat lock, `errChatBusy` exists and is correctly `provablyNotSent == true`, and `perform` finally applies `DeliveryBudget` to a whole logical delivery. The newline/reassembly and direct-vs-relay accounting fixes from earlier rounds hold up too.

Two blocking items left. I reproduced both on this head before writing this.

## 1. `outcomeGrace` counts one delivery budget for a chain that can spend one per offer

`perform` gives every claimed delivery its own full `DeliveryBudget` (`relay_outbound.go:883`). `errChatBusy` is `provablyNotSent == true`, so an offer that waits out that entire budget on a busy chat releases its claim and the frame is re-offered. Every offer in the chain can therefore burn a full delivery budget. But `outcomeGrace` computes the offer count at `relay_outbound.go:1073` and then adds delivery time exactly once, at `relay_outbound.go:1091`.

With `ClaimBudget=1ms`, `DeliveryBudget=100ms`, `LeaseSettle=40ms`, `RetryBackoff=5ms`:

```
offers=8  retryPlan=[5ms 10ms 10ms 10ms 10ms 10ms 10ms]  chain=65ms
outcomeGrace = 193ms
worst case with a busy chat = 893ms   (grace counts 100ms of delivery; the chain can spend 800ms)
```

The watcher then Resolves while the chain is still retrying: if a claim is held at that moment it fences a send that is still in progress, and if the chain happens to be in release/backoff the reply gets recorded dropped and then delivered. That is the "one reply counted as delivered and dropped at the same time" outcome the comment block at `relay_outbound.go:1285-1300` says this package no longer has.

Worth being explicit that this one is **introduced by this PR**, not pre-existing. On `main` the single-budget term is correct, because there is no chat lock and `perform` never applied the budget at all — an offer was one frame and one ack. This PR adds both the lock and the enforcement, and the arithmetic did not move with them.

`TestRelayOutcomeGrace_CoversEveryStoreRoundTripTheChainCanMake` passes because it encodes the same one-budget assumption it is meant to check (`relay_outbound_attribution_test.go:145-149`). Please scale the delivery term by offer count, and pair the arithmetic test with a "chat stays busy + publisher watcher" case so the assumption is checked against behaviour rather than against itself.

## 2. The new chat lock hands out an already-dead context, and the pre-write path calls that "maybe sent"

`request` returns a bare `ctx.Err()` from its pre-write check at `ws_sender.go:306-308`, before a waiter is registered or a frame is built. `unconfirmedReason` files that as `"interrupted"` and `provablyNotSent` returns false, so zero frames went out and the relay still settles the claim and stops re-offering.

Elon's repro for this was an already-cancelled caller on an idle chat, and it does reproduce — 0 frames, `provablyNotSent=false`, `unconfirmedReason="interrupted"`. I want to be precise about that case though: **that shape is pre-existing on `main`** and I would not block this PR on it alone.

What makes it blocking here is that the new lock opens a second door into the same misclassification, and this is the door `errChatBusy` was written to close. `acquire`'s blocking select at `ws_sender.go:176-181` has two ready cases when the lock frees at the same moment the budget expires, and Go chooses among ready cases at random — so it returns the lock, with a context that is already over, and the very next thing that runs is the pre-write check above. In a harness that forces the release and the cancellation to coincide, 1930 of 2000 contended acquires came back holding a live lock and a dead context rather than `errChatBusy`.

That ratio is an artifact of forcing simultaneity and says nothing about production frequency — in production the two events are independent, so this is rare. What it establishes is that the case is reachable through the new lock, that `errChatBusy` covers only part of the race it was introduced for, and that when it does happen the answer is the wrong one. Combined with the fact that this PR is the first to make the delivery context actually expire (`perform` at `relay_outbound.go:883`), the exposure is new even though the misclassification is not.

Please return one not-attempted sentinel from `request`'s pre-write check and have the direct, relay and media classifiers test for it ahead of their generic context branch. `TestDeliverRelayed_ContextErrorsDoNotRelease` (`outbound_two_replica_db_test.go:814`) currently asserts the wrong result for this case — its own comment says "already expired: sendTextCtx fails on its pre-write check" — so it should become two tests: pre-write, which wrote nothing and is safely retryable, and post-write while waiting for an ack, which is genuinely unknown and must not be retried.

## What I verified

- Both items above reproduce on `ba50fef` with minimal cases; the grace numbers are the ones quoted.
- `errChatBusy` really is `provablyNotSent == true`, so the re-offer loop in item 1 is real and not hypothetical.
- Compared against the merge base `a843b44` to separate regression from pre-existing: no `chatLocks`/`errChatBusy` there, and `deliveryBudget()` appears only in the grace arithmetic, never as an applied bound.
- `go vet ./internal/integrations/wecom`: clean. `git diff --check`: clean.
- Long-answer, split, partial-send, chat-lock, chat-busy and grace tests: `go test -race` passes on this head.
- Full `go test -race ./internal/integrations/wecom` does not complete here — my local test database is behind the current schema (`route_revision`, `context_revision`, `channel_task_delivery`). This PR carries no migration and those failures are unrelated to the diff. GitHub checks are green.
- No real WeCom end-to-end run behind this review.

No schema or config migration in this PR; rollback stays a plain revert, confined to over-cap WeCom messages.

seacen:
Both fixed at head `2cc3bf59c`, rebased onto `8c4f4328f`. Same method as the bubble PR: an invariant in one sentence, a query that enumerates every site it applies to, then all of them.

The rebase had one conflict, in `media_upload.go`'s `sendMedia` — `main` took the send rate gate from #8345 and routed the media push through `sendMsgFrame`, while this branch added the chat lock at the same spot. Both belong: acquire, then frame.

## 1. `outcomeGrace` counts one delivery budget for a chain that can spend one per offer

Your arithmetic reproduces here exactly, including the 893ms.

**Invariant: the grace covers everything the re-offer chain can spend, and every offer in it can hold a full `DeliveryBudget`.**

The sites for this one are the places that describe the arithmetic rather than call sites, so the enumeration is `grep -rn "DeliveryBudget\|deliveryBudget"` over the package. Four places had it wrong in the same direction: the term itself, `perform`'s budget comment, the `RelayConfig.DeliveryBudget` field doc, and one stale sentence in a relay test's comment. The term now scales by offer count and the three comments say "per offer" and why.

```
reviewer's numbers:   offers=8   chain=65ms    deliveryBudget=100ms  grace=893ms
production defaults:  offers=12  chain=50.1s   deliveryBudget=5s     grace=2m42.5s
```

**Worth knowing before you approve: the default grace goes from 1m47.5s to 2m42.5s.** Production sets only `ReplayGrace` and `LeaseSettle`, so `DeliveryBudget` falls to `ackTimeout`'s 5s. The direction is safe — a loss is counted later, never sooner — but it is a real change to how long the watcher waits.

`TestRelayOutcomeGrace_CoversEveryStoreRoundTripTheChainCanMake` encoded the one-budget assumption it was meant to check, as you said. It is fixed, and it is no longer alone: `TestRelayOutcomeGrace_OutlastsAChainThatSpendsADeliveryBudgetPerOffer` holds the chat busy for the whole run and watches the publisher, so the arithmetic is checked against behaviour. Without the fix:

```
the watch resolved the reply on offer 5 of 8 — the chain was still running, and Resolve
fences the key, so a later offer that delivered would have recorded nothing while the loss
was already counted (grace 535ms, 8 offers × 100ms of delivery alone)
```

Clean over `-count=8`.

## 2. The chat lock hands out an already-dead context

**Invariant: a failure that ended before any byte could leave this process is read as "provably not sent" by every classifier that decides whether the answer may go out again.**

```
grep -rn --include='*.go' -E 'errors\.Is\([^,]+, context\.(Canceled|DeadlineExceeded)\)' \
  internal/integrations/wecom/ | grep -v _test
```

Four sites. Three are the classifiers you named — `unconfirmedReason` (direct), `sendOutcome` (media), `provablyNotSent` (relay) — and all three now test the sentinel ahead of their generic context branch. The fourth is `sendMsgFrame`'s closing switch in `rate_limit.go`, which is **not** a delivery classifier: it chooses between a stated refusal and the outcome of a second attempt. Its context branch still fires, because the sentinel wraps `ctx.Err()` rather than replacing it, and that dependency is now pinned by an assertion rather than left to be rediscovered.

`request`'s pre-write check returns `errNotAttempted`. One consolidation I made beyond what you asked, so you can push back on it: **`errChatBusy` now wraps `errNotAttempted`** instead of being a second independent sentinel. Otherwise each of those three classifiers carries its own list of "sentinels that mean nothing was written", and the next one added has to find all three. Its error string is unchanged.

`TestASendThatNeverStartedIsProvablyNotSent` covers all three classifiers in one case. Without the fix:

```
provablyNotSent(context canceled) = false, want true
unconfirmedReason(context canceled) = "interrupted", want "" — "interrupted" says the
  message may be on the person's screen
sendOutcome(context canceled) = unknown, want definitely_failed
```

`TestDeliverRelayed_ContextErrorsDoNotRelease` is split the way you asked. Its own comment said "already expired: sendTextCtx fails on its pre-write check" while asserting the claim must be held, so the pre-write half was asserting the wrong result. The post-write half keeps the property and is stronger now — it also asserts exactly one frame reached the wire and that the count lands on `outbound_unconfirmed` rather than `outbound_dropped`.

## Existing tests changed

- **`TestRelayOutcomeGrace_CoversEveryStoreRoundTripTheChainCanMake`** — the `worst` expression scales by offer count. This is the assumption-encoding one.
- **Two Redis outcome-watch tests** gained `DeliveryBudget: 20ms`. They already shrink every other grace term; `DeliveryBudget` became a per-offer term, so leaving it at the default takes that chain's grace from 5.6s to 40.6s and starves a 10s wait. Configuration following the arithmetic, not a loosened assertion.
- **`TestRelay_ADeliveredReplyIsCountedOnceAndNotAlsoLost`** — a hardcoded `time.Sleep(600ms)` commented "Past the whole grace" is now `outcomeGrace() + 200ms`. **That comment was already false before this PR**: the grace was 5.6s, so the publisher assertion had been running before the watcher ever woke. Demonstrated by breaking the holder's settle — the 600ms version reports only the holder's failure, the grace-derived one also reports the publisher's `outbound_dropped = 1`.

## Checks

`go build`, `go vet` untagged and under `agentintegration`, `-race -p 1` on the touched package differenced against the merge base, DB-backed tests proven to have RUN rather than skipped, and the merge with `main` compiled. `VERDICT:FIT`.

Two things about the local rig rather than the diff, since they affected what I could claim: the throwaway test database was 33 migrations behind `main` and 12 DB tests were failing on a missing column until I migrated it, and the Redis container was down, which was silently skipping four tests. Both now run.

Bohan-J (approved):
Thanks @seacen for the fix and for working through the review feedback. Re-reviewed at `2cc3bf59c`: both remaining blockers are resolved. The outcome grace now accounts for every offer's delivery budget, and cancellation before the first write is correctly distinguished from cancellation while awaiting an ACK.

The regression tests also catch the old behavior when those fixes are temporarily reverted. Focused tests passed under `-race -count=10`; the non-DB/Redis WeCom and server tests, `go vet`, and diff checks passed. DB/Redis integration tests were skipped locally, and no live WeCom E2E was run. CI is green on this head.

Approved — thanks for adding the regression coverage.

PR #8345, opened 2026-09-12
MUL-7314 fix(wecom): pace outbound pushes at WeCom's published rate, retry one throttle, best effort past it

seacen:
## What this delivers

Three things, in the order they matter. Outbound pushes are paced at WeCom's published rate for a conversation — 30 a minute, 1000 an hour, counted per chat because that is the unit the doc names. A push WeCom throttles (`45009` or `45033`) gets one jittered retry, when the caller can afford it and when our own window says a slot is free. And past that, this is best effort: a burst far over the rate has its extra frames refused before the write, not delayed into the next window.

That last sentence is the honest limit and it is worth being explicit about. At the published figures the 31st frame of a burst is not held back — the next slot is most of a minute away and no caller on this path has a minute to give — so it is refused where it stands, filed as `transport_error`, with nothing on the person's screen to say so. What changes is where the failure happens: on our side, where it costs no quota, stays provably unsent so the relay can offer it elsewhere, and shows up in a counter, instead of on WeCom's side as a `45009` that every caller downstream reads as final.

## Problem

WeCom refuses an over-quota push with errcode `45009`, and every caller on this path reads a stated refusal as final — `provablyNotSent` says no re-offer, `classifyDrop` files `platform_refused`, and the reply is gone. **One throttled frame is one answer the person never sees**, with nothing on their screen to say so.

Nothing in the adapter currently paces those pushes, so the ceiling is reached by whoever happens to be talking.

## The unit is the conversation, not the member

That is WeCom's own sentence, stated twice on the aibot page — once under `aibot_respond_msg`, once under `aibot_send_msg`:

> 无论是回复还是主动推送消息，总共给某个会话发消息的限制为 30 条/分钟，1000 条/小时
> — [path/101463](https://developer.work.weixin.qq.com/document/path/101463)

So the windows are keyed on chat id: in a group the whole room shares one allowance. `30` and `1000` are the documented ceilings themselves, not a margin under them. A file and a piece of an answer are the same `aibot_send_msg` to WeCom, so `sendMedia` spends from the same allowance — a turn that answers in words and then sends three attachments counts as four.

`aibot_send_msg` is also the only command this tree sends into a conversation with, so gating it gates everything there is. That is a fact about the tree and not a property of the quota: the sentence above puts `aibot_respond_msg` on the same allowance, so a reply path that writes frames has to reserve here as well when one lands.

## The retry

Retrying a **stated refusal** is safe in a way retrying a timeout is not: a non-zero errcode is the server saying it did not act on the frame, so a second attempt cannot duplicate anything.

It runs only when the remaining deadline covers the backoff and an ack wait, and only when our own window says a slot has freed — `nextSlot` is `admit`'s read-only twin over the same window, so a slot that is most of a minute away means our count agrees with WeCom's and the ban is real. The delay is jittered ±50%: `45033`'s own documented remedy is to reduce concurrency, and a fixed delay brings every caller back in the same instant.

The refusal survives the attempt. A second attempt that ends in a context error or another throttle reports the **first** refusal, because downgrading a definite outcome to an unknown one hands an operator a message to resolve by hand. `errAckTimeout` and `errWriteAttempted` are returned as themselves — those say the second frame may be in front of the user, which matters more.

`reserve` keeps `ackTimeout` of the caller's deadline for the write rather than merely fitting the wait under it, so the gate cannot spend the budget the ack needs.

## The reconnect seam

The gate's count lives on the `wsSender`, which is one socket, which is one installation — so what it counts is what WeCom counts. A reconnect mints a new sender with an empty window, so a socket that flaps mid-minute can put us over WeCom's count while ours reads clean. That is the case the retry exists for, and the reason the gate is not written as if it were exact.

## Relayed and direct frames are not treated alike

`errRateLimited` satisfies `provablyNotSent`, so a frame that came through the relay is re-offered by the scheduler (about 1.5 × `LeaseSettle`). The same frame produced on the replica that already holds the lease never goes near that chain — the refusal goes back to its caller and is filed as a drop. Giving the direct path a bounded re-offer of its own is a change to the outbound path rather than to the gate, and is deliberately not in this PR. The asymmetry is written down next to `errRateLimited`.

## Tests

At the published figures: a burst just over the rate is delayed and sent; the 31st frame of a burst is refused where it stands, with its drop classification pinned. Around the retry: a retry the caller cannot afford keeps the refusal; a second attempt cut short reports the first refusal; a throttle is not retried when our own window is spent too; the backoff is spread across callers. Around the gate: it leaves the write its ack budget; one busy chat does not throttle the others; a media push spends the same allowance; quiet chats are forgotten.

Every one of them is reverse-verified — the failure output for each is in the review thread.

`go test -race -count=2 ./internal/integrations/wecom` green; `./cmd/server` green at `-count=1` (13 of its tests cannot run twice in one process, on `main` as well).

## Note

This and #8344 both touch `sendTextCtx`. Whichever lands second needs a small rebase there.


Bohan-J (changes requested):
The product direction is right and the gate belongs at this seam — `aibot_send_msg` is the single command that spends the per-conversation allowance, and putting both the text route and `sendMedia` behind one door is the correct shape. I am asking for changes on four local points, not on scope: I am **not** asking this PR to become a durable queue, or to carry quota across replicas.

I ran the new tests (`-count=2`, green) and wrote two throwaway probes in the package to check the two claims I could not settle by reading.

### 1. The retry can turn a definite refusal into an unknown outcome

`sendMsgFrame` guards `ctx.Done()` *during* the backoff (`rate_limit.go:136`) and returns the refusal there — but once the backoff elapses, the loop re-enters `reserve` + `request` with whatever budget is left, and whatever the second attempt returns replaces the first refusal. `request` returns `ctx.Err()` both from its pre-write check and from the ack wait, so the 45009 is gone.

Probe: first push answered `45009`, `retryBackoff` 100ms, caller ctx 300ms, second push left without a verdict:

```
err=context deadline exceeded | is45009=false | unconfirmedReason="interrupted"
  | drop="transport_error" | writes=2
```

`unconfirmedReason` → `"interrupted"` is `outbound_unconfirmed`, i.e. "the user may already have it, resend by hand" — exactly the downgrade the file's own comment says it exists to prevent. On the reply path the arithmetic reaches it without anything pathological: `rateWaitBudget` 3s + `ackTimeout` 5s + backoff 2s = the whole 10s budget (`outbound.go:155`) before the second write starts.

Please gate the retry on the budget it needs — retry only if the remaining deadline covers `retryBackoff + ackTimeout`, otherwise return the refusal as it stands — and keep the first refusal as the returned error when the second attempt yields a context error or `errRateLimited`.

### 2. `rateWaitBudget` is sized against the wrong caller, and leaves no room for the write

The comment at `rate_limit.go:55` sizes 3s against "the outbound subscriber gives a whole delivery ten seconds". The inbox push path gives 5s (`outbound.go:415` → `sendTextCtx`), and `ackTimeout` alone is 5s (`ws_sender.go:111`). `reserve` only checks that the wait fits under the deadline (`rate_limit.go:232`), so on that path the gate can spend 3s of 5s and hand the write 2s for a 5s ack — again `"interrupted"`, on the first attempt. Reserve should keep `ackTimeout` of headroom rather than merely fit under the deadline.

### 3. "Two seconds is one slot's worth" does not follow from a sliding window

`rate_limit.go:93` derives the backoff from "30 per minute is a send every two seconds on average". The gate is deliberately *not* an average — it is a sliding count, so the next slot frees when the oldest entry in the window ages out, which for a burst is nearly the whole span. My probe at the published figures prints the gate's own answer: `the next slot is 59.999s away`. WeCom's own note on 45009 is that the ban runs to the end of the window it was earned in, so 2s has no recovery basis for a quota-caused throttle; it is only plausible for a throttle caused by something else clearing.

Either justify it as what it is — one cheap attempt at a transient throttle, not a wait for a slot — or derive the delay from `admit`, which already knows when the next slot frees. Related: 45033's documented handling is to *reduce* concurrency, and a fixed un-jittered 2s makes every concurrent caller retry in the same instant, while the gate bounds the window but not in-flight requests.

### 4. The gate's behaviour at published figures is untested, and the description overstates it

The burst test (`rate_limit_test.go:181`) runs a 150ms/limit-2 window with a 1s budget, where the wait can never exceed the budget — so "a burst arrives late rather than not at all" is true by construction there and does not transfer. At production parameters the 31st frame of a burst is **not** delayed:

```
31st send: err=...quota is spent: the next slot is 59.999s away, elapsed=0s,
  pushes=30, drop="transport_error"
```

Nothing is written, nothing tells the person, and the drop is filed as `transport_error` instead of `platform_refused`. That is still an improvement — mild bursts are paced, a lifted throttle is recovered, and the failure moves to our side of the socket where it is honestly ours — but it is not "keep outbound pushes inside WeCom's quota" for the case the PR opens with.

Please add a test at `rateLimitPerMinute` + 1 with the real `rateWaitBudget` pinning what actually happens, and reword the title/description to what the change delivers: pace pushes against the published quota, retry one throttle, best-effort under a heavy burst.

### Two notes, not blocking

- `errRateLimited` satisfies `provablyNotSent` (correctly), so a relayed frame is re-offered across the dispatcher's chain (~1.5 × `LeaseSettle`), while the same frame produced on the lease holder is dropped at the first refusal. Worth one sentence in the description, or a bounded re-offer on the direct path.
- The reconnect seam (a new `wsSender` starts with an empty window) is documented in the code but not in the description, which is where the "inside WeCom's quota" claim is made. Also `TestSendTextReportsAServerRefusal` refuses with 45009, so it now spends the full 2s backoff and writes a second frame — it should use the shortened `retryBackoff`.

seacen:
All four are fixed, and I've taken the reframing in 3 and 4 rather than arguing for the original wording. Title and description updated to what it delivers.

## 1. The retry no longer downgrades a stated refusal

`sendMsgFrame` is now first attempt / decision / second attempt, with the refusal held across the whole thing. The retry runs only when the remaining deadline covers `delay + ackTimeout`; otherwise the refusal is returned as it stands. When the second attempt yields a context error or `errRateLimited`, the first refusal is what comes back. `errAckTimeout` and `errWriteAttempted` still return as themselves — those say the second frame may be in front of the user, which matters more than the refusal does.

Your probe's parameters are the test: `TestARetryTheCallerCannotAffordKeepsTheRefusal` (100ms backoff, 300ms caller ctx) and `TestASecondAttemptCutShortReportsTheFirstRefusal`.

Both halves were needed and they are not two spellings of one fix — with only the budget check removed the error assertion still passes and only the write count fails; both removed reproduces your `context deadline exceeded`.

## 2. reserve keeps the write its ack budget

`reserve` now waits at most `min(rateWaitBudget, deadline − ackTimeout)`. On the inbox path (5s) the gate therefore refuses outright instead of spending 3s and handing the write 2s for a 5s ack. The three callers and their deadlines are named in the comment.

`TestTheGateLeavesTheWriteItsAckBudget`. Reverse verification using the deadline itself: `error = context deadline exceeded, want errRateLimited`, after 5s.

## 3. Two seconds is now stated as what it is

The comment no longer derives it from an average. It is one cheap attempt at a transient throttle, and the basis is 45033 — 「出现这种限制错误后，请企业调低并发数」. For 45009 the published text is the opposite: 「频率拦截时长一般与调用的限制时长相同，比如说是分钟级别的限制，则在中频率后的1分钟后自动解除」. Two seconds cannot clear that, and the comment says so.

The decision comes from `admit`. `nextSlot` is its read-only twin over the same window: if our own next slot is far away, our count agrees with WeCom's, the ban is real, and there is no retry. The backoff is jittered ±50%, because 45033's own remedy is to reduce concurrency and a fixed delay brings every caller back in the same instant.

`TestAThrottleIsNotRetriedWhenOurOwnWindowIsSpentToo` and `TestTheRetryBackoffIsSpreadAcrossCallers`. Reverse verification of the first: `the call took 2.976430042s — it served a backoff instead of asking admit when the next slot frees`.

## 4. Tested at the published figures, and the claim is narrower

The 150ms/limit-2 burst test is gone — the wait could never exceed the budget there, so it proved its own premise. Two tests at the real window replace it:

- `TestABurstJustOverThePublishedRateIsDelayedNotLost` — a slot 400ms out, held and sent.
- `TestTheThirtyFirstFrameOfABurstIsRefusedNotDelayed` — pins what actually happens, including the `transport_error` classification you quoted.

The title and description now say: pace pushes at the published rate, retry one throttle, best effort past it. The description states the 31st-frame case in the second paragraph rather than leaving it to be discovered, and adds the two things you noted were missing from it — the reconnect seam, and that a relayed frame is re-offered by the dispatcher while the same frame produced on the lease holder is not.

Scope unchanged: no durable queue, no cross-replica quota.

Rebased on `27d0529fa`. `go test -race -count=2` green on `./internal/integrations/wecom`.

multica-eve (changes requested):
🔄 **Request changes** — 产品方向可以接受，但第二次发送已写出后的取消仍被错误分类为首次平台拒绝，合入前需要修正。

具体触发条件是：首次 push 收到 `45009/45033`，随后的第二次 `request` 已把帧写出，但在等待 ACK 时 context 被 cancel 或 deadline 到期。`sendMsgFrame` 目前把这两类错误改回首次 refusal，会将“第二帧可能已送达”报成 `platform_refused`/确定未送达。首次 refusal 只能证明第一次尝试的结果，不能覆盖已发生的第二次写入。

当前的 `TestASecondAttemptCutShortReportsTheFirstRefusal` 恰好在 `conn.onPush` 观测到第二帧后才 cancel，却期望首次 refusal，因此测试正在固化这个错误语义。

请区分两种情况：

- 第二次写入前 context 已结束：可以返回首次 refusal，因为没有新的送达可能。
- 第二次帧已写出，等 ACK 时 context 结束：必须保留 unknown/已尝试写入的结果（可沿用 `errWriteAttempted` 或等价的类型化状态），并修正上述测试覆盖这两条路径。

其他先前要求的改动——重试预算、ACK 时间预留、jitter、生产参数 30+1 测试及 best-effort 边界说明——已在当前 head 中补齐。这一处修正后，我倾向批准合入。

seacen:
Fixed, and you were right that the existing test was the worse half of it: it was not missing a path, it was pinning that path to the wrong answer.

## The two cancellations are now told apart

`request` returns a context error from two places that mean opposite things — the pre-write check, where not a byte has gone out, and the wait after `s.write` returned nil, where the frame is on the wire. They were the same error value, so no caller could tell them apart. The post-write one is now wrapped in `errAckAbandoned`, and `sendMsgFrame`'s switch checks it before the generic context arm.

`errAckAbandoned` rather than reusing `errWriteAttempted`: that sentinel's own doc says it marks "a failure raised by the socket write itself", and here the write succeeded. This is the `errAckTimeout` family — the frame went out, the verdict did not come back. Wrapping rather than replacing keeps every downstream classifier unchanged.

## The old test is replaced by two

`TestASecondAttemptCutShortReportsTheFirstRefusal` is gone. Its setup cancelled from `conn.onPush`, which only fires after `WriteMessage` has appended — so it could only ever land on the already-written path, and it asserted the first refusal there. In its place:

- `TestASecondAttemptCutShortBeforeItWroteReportsTheFirstRefusal` — cancelled from the retry's own WARN, the one moment on this path where the refusal is in hand and nothing has been written. It asserts the first refusal, and one frame on the socket.
- `TestASecondAttemptCutShortAfterItWroteKeepsTheUnknown` — asserts the result is not a `*wecomAPIError`, that `unconfirmedReason` is non-empty and `provablyNotSent` false, and that two frames reached the socket.

The cancel in the "before" case is triggered from the log hook rather than a conn hook on purpose: a conn hook races the first attempt's ack for `request`'s select, and with both channels ready Go picks at random.

Two more pin the evidence itself, at the `request` layer, since "which step counts as written" is the load-bearing judgement: `TestARequestCancelledBeforeTheWriteReportsNothingWritten` and `TestARequestThatLosesItsContextAfterTheWriteSaysTheFrameWentOut` (cancel and deadline, since you named both).

Reverse verification, dropping `errAckAbandoned` from the unknown arm:

```
error = wecom: aibot_send_msg rejected errcode=45009 errmsg=scripted, want the second
attempt's own outcome — the first refusal is a fact about the first frame, and the
second one is already out
```

and returning a bare `ctx.Err()` from the post-write select:

```
error = context canceled does not carry errAckAbandoned, so a caller cannot tell it
from a cancellation that wrote nothing
```

## One qualification on your description

On the retry path only cancellation is reachable, not a deadline: `retryUnaffordable` requires the remaining budget to cover `delay + ackTimeout`, so after the pause there is still an ack wait left and `request`'s own 5s timer fires first. The deadline case is reachable on the **first** `request`, because `reserve` admits immediately when a slot is free without checking what the caller has left — deliberately, and the comment says so. That path does not go through the switch at all, which is why the fix is in `request` rather than in `sendMsgFrame`: it is the only place that covers both.

Rebased on `a843b44a4`, `go vet` clean under both tag sets, `-race` green.

multica-eve (approved):
Thank you for working through the review feedback so carefully. The final change correctly distinguishes cancellation before the retry write from cancellation after the frame is already on the wire, and the tests now pin both sides of that boundary. The pacing and retry scope, including its best-effort limits, is also documented accurately. Approved.

PR #8346, opened 2026-09-12
fix(dingtalk): tell the chat when a run was cancelled

seacen:
## Problem

DingTalk posts a "working on it" ack the moment it ingests a message (`ack.go`), and subscribes `chat:done` and `task:failed` to deliver what comes back. It does not subscribe `task:cancelled`.

A cancelled run publishes neither of the two it does subscribe: cancellation produces no `chat:done`, and the daemon's own completion arrives to find the row already cancelled, where `CompleteAgentTask`'s `status = 'running'` guard matches nothing and the answer is discarded. So the ack that promised a reply is the last thing in the chat, and nothing ever contradicts it.

## Change

Subscribe `task:cancelled` and deliver a one-line notice for it, through the same path the failure notice already takes — the delivery row, the binding, the existing send. Lark and Slack clear a reaction for the same event; DingTalk has a message rather than a reaction, so it says one sentence instead.

## Tests

`TestOutboundCancelled_PostsTheNoticeIntoTheDingTalkChat` drives the real `Register` → `bus.Publish` → HTTP stub, so it fails when the subscription is missing rather than when a helper is renamed.

`TestOutboundCancelled_StaysSilentWhenTheTurnIsNotDingTalks` is the control, in four shapes: no delivery row, another platform's row, a run asked in the web UI, and a `retry_pending` failure.

Reverse-verified — with the subscription removed:

```
--- FAIL: TestOutboundCancelled_PostsTheNoticeIntoTheDingTalkChat
    send path = "", want "/v1.0/robot/oToMessages/batchSend" — the cancellation
    reached no DingTalk chat, and the ack that promised a reply is still the
    last thing in it
```

`go test -race ./internal/integrations/dingtalk ./cmd/server` green.

### DIFF
--- a/server/internal/integrations/dingtalk/outbound.go
+++ b/server/internal/integrations/dingtalk/outbound.go
@@ -26,12 +26,12 @@ type outboundQueries interface {
 }
 
 // Outbound delivers an agent's chat reply back to DingTalk — the outbound half
-// of the round trip. On EventChatDone / EventTaskFailed
+// of the round trip. On EventChatDone / EventTaskFailed / EventTaskCancelled
 // it finds the DingTalk chat binding for the task's session and posts the reply
-// (or failure notice) into the originating conversation. Sessions with no
-// DingTalk binding are ignored, so it coexists with the Feishu and Slack
-// subscribers on the shared event bus. Registered only when DingTalk is
-// configured.
+// (or failure / cancellation notice) into the originating conversation.
+// Sessions with no DingTalk binding are ignored, so it coexists with the Feishu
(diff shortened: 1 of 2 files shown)


yyclaw (changes requested):
Thank you for contributing this change.

Cancellation handling is already implemented in #8125: when a task is cancelled, the bot removes the "收到" acknowledgement reaction it previously added to the corresponding user message. This is the intended cancellation behavior; a separate cancellation message is not needed.

We will not accept this PR, as #8125 already addresses the cancellation handling.

PR #8347, opened 2026-09-12
MUL-7319 fix(wecom): a completion of whitespace is not a message

seacen:
## Problem

A completion of `"\n"` with a file bound to it is not words. The adapter reads it as words anyway, because both send paths test `content != ""`, and two things follow:

- an empty markdown message goes into the chat, ahead of the file;
- the reply's outcome is recorded as *delivered by the text* — `carriesTheReply` was already false — so when the file then fails, the per-file counters still move and the user is still told, but the reply itself stays counted as delivered. `outbound_dropped` never moves for it. What is lost is the reply-level outcome, not every trace of the failure.

Both paths have it, which is why both are fixed here: the relayed path in `deliverRelayed` and the local one in `processEvent`. Fixing only one would leave the same defect behind and make which replica held the socket decide whether the user sees a blank message.

## Change

`hasVisibleChar` — a rune that is neither whitespace nor a control character — replaces `!= ""` at both sites, and the same predicate decides `carriesTheReply`. An answer of whitespace now takes the `nothing_to_say` exit, and the file carries the reply's outcome.

The predicate is deliberately not "the client will render something": format runes (U+200B, U+FEFF, a soft hyphen) pass it and still show as nothing. Nothing upstream rejects such a body either — a completion of format runes reaches the chat as an empty bubble today, and this predicate is not what stops it. The line is drawn here to keep a Unicode category table out of the adapter; moving it is a separate decision, and that table is its cost.

## Tests

- `TestRelayedReply_WhitespaceOnlyContentIsNotSentAndTheFilesCarryTheReply` — the relayed path. It runs the settlement callback the delivery hands back before asserting: that path does not count inline, so without running it `outbound_delivered == 0` holds because nobody counted rather than because there was nothing to count — and it kept holding with `f.Content != ""` restored.
- `TestOutbound_WhitespaceOnlyCompletionIsNotSentAndTheFilesCarryTheReply` — the local path.

Reverse-verified — with `!= ""` restored on both:

```
the chat received ["\n"], want nothing — a completion of whitespace is not a message
outbound_delivered = 1, want 0 — no words reached anybody
outbound_skipped:nothing_to_say = 0, want 1 — with no words, the files carry this reply's outcome
```

## One overlap with #7954

That PR adds a `hasVisibleChar` to the same file, for its own reason — a closing frame the server considers empty seals nothing. The function bodies are identical, eight lines; only the doc comments differ, because the two uses are different.

Merging the second of the two conflicts in `ws_frame.go` rather than landing a duplicate declaration silently, so git catches it either way. Whichever lands first, the other's copy is the one to drop — there is nothing to reconcile beyond deleting one and keeping the comment that fits the remaining caller. Neither PR depends on the other.

### DIFF
--- a/server/internal/integrations/wecom/outbound.go
+++ b/server/internal/integrations/wecom/outbound.go
@@ -178,7 +178,7 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 	// assistant message for exactly that case, and returning now would throw
 	// the work away.
 	content := deliverableContent(e)
-	if content == "" && !o.mayCarryAttachments(e) {
+	if !hasVisibleChar(content) && !o.mayCarryAttachments(e) {
 		o.skipped(ctx, e, skipNothingToSay)
 		return nil
 	}
@@ -284,7 +284,12 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 	// Words first. An empty completion reaches here only because a file is
 	// bound to it, and an empty markdown bubble ahead of that file would be
 	// noise the user has to scroll past.
-	if content != "" {
+	//
+	// Empty is hasVisibleChar's sense of it, not `!= ""`. A completion of "\n"
+	// is a bubble with nothing in it on the reader's screen, and counting it
+	// as the words that answered the turn also tells the file below it that
+	// the reply has already been accounted for.
+	if hasVisibleChar(content) {
 		if err := sender.sendTextCtx(ctx, binding.ChannelChatID, chatType, content); err != nil {
 			return err
 		}
@@ -300,7 +305,7 @@ func (o *Outbound) processEvent(ctx context.Context, e events.Event) error {
 		ChatID:         binding.ChannelChatID,
 		ChatType:       chatType,
 		SessionID:      e.ChatSessionID,
-	}, content == "")
+	}, !hasVisibleChar(content))
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


Bohan-J:
Merged — thank you, @seacen.

Two things made this easy to accept. Fixing both send paths rather than one, and saying why: a one-sided fix would have left which replica holds the WS lease deciding whether the user sees a blank bubble, and that reasoning is in the diff itself. And the reverse verification — I reproduced it, and it holds exactly as described: restoring `!= ""` on both sites makes the new tests fail with the messages you quoted, while `go vet` stays silent. Since nothing in the toolchain catches that spelling, those tests are the only guard this behavior has, and that is precisely why they needed to exist.

I pushed one commit to your branch before merging (94efba7):

- **The relayed test now runs the settlement callback.** `deliverRelayed` hands its record back to the dispatcher rather than counting inline, so with `res.record()` never invoked, `outbound_delivered == 0` was asserting that nobody counted rather than that there was nothing to count — it kept passing with `f.Content != ""` restored. Running the callback first makes that revert report `outbound_delivered = 1`, matching the local path and your own reverse-verification note.
- **Reworded `hasVisibleChar`'s closing line.** It said widening the predicate would mean carrying a Unicode table "for input this adapter does not accept" — but nothing upstream rejects such a body; a completion of format runes reaches the chat as an empty bubble today. The boundary and its cost are unchanged, just the claim about what is accepted.

I also corrected one line in the PR description: when the file fails, the per-file counters still move and the user is still told. What was lost is the reply-level outcome — `outbound_dropped` never moving for it. The fix is right either way, but the narrower statement is the one that will age well.

One follow-up for #7954: since this landed first, that PR now needs to drop its own copy of `hasVisibleChar` and resolve the conflict in `ws_frame.go`. Worth making the surviving doc comment cover both callers — the whitespace completion here, and the closing frame there.

PR #8348, opened 2026-09-12
MUL-7318 fix(wecom): name the two exits that leave a reply unaccounted for

seacen:
## Problem

Two different exits from `processEvent` return `nil` silently, and from outside the process they are the same nothing:

- the turn belongs to another channel (Slack's, Lark's — this subscriber is on the bus they all publish to), and
- the turn is WeCom's own but **no delivery row was recorded for it**, so there is no address to answer at.

The first is ordinary traffic. The second means somebody asked a question in a room and the answer is going nowhere, with no counter and no log line anybody can alert on.

## Change

Both exits are named: `not_wecom_turn` and `no_delivery_row`. Only the second is actionable, so only it is a WARN — `skipReason.actionable()` is the one bit that decides, and the ordinary one stays at DEBUG where it is a rate rather than a page.

Telling them apart needs the origin gate, and **where** the gate is asked turned out to matter twice:

- Asked ahead of the delivery lookup, every question ever typed in the Multica web UI reaches the no-row branch too, and the WARN fires once per web message — the loudest line in the log, on the most ordinary event in the deployment.
- Asked ahead of the lookup for *every* turn, another channel's `chat:done` pays for a task row and an origin lookup before the one query that identifies it as somebody else's.

So it is asked inside the no-row branch, where its answer is the thing that changes what gets recorded, and nowhere else. A turn whose row exists costs exactly what it costs today.

## Tests

`TestNonWecomSessionIsNotADrop` was one case that conflated "another platform's row" with "no row at all" — it set `sessionErr = pgx.ErrNoRows` under a name that said the opposite, which is the ambiguity this PR exists to remove. It is now four: another platform's row, asked in the web UI, asked in the web UI with no row, and a channel turn with no row. Each asserts the reason counter and the log level.

`TestAnotherPlatformsTurnDoesNotPayForTheOriginGate` holds the cost decision: another platform's turn does zero task lookups.

Reverse-verified three ways — both branches back to a bare `return nil` (the reasons go unrecorded); the gate hoisted ahead of the lookup (the web-UI case takes the actionable exit and gets the WARN); and the gate hoisted for every turn (another platform's turn pays for one task lookup).


seacen:
`cursor-background-lifecycle (windows-latest)` is red on this one and it is not this change: the failure is in `server/pkg/agent`, which this PR does not touch — the four files here are all under `integrations/wecom`. The same job passes on ubuntu and macos in this run, and passes on windows for the two sibling PRs opened alongside it (#8344, #8347).

I cannot rerun the job from outside the org. Happy to push a no-op to retrigger if that is easier than a rerun on your side.

Bohan-J (changes requested):
## Request changes

The product direction is right and I'd accept the shape. But as implemented, this PR turns one previously actionable failure into an ordinary DEBUG skip, and the new actionable skip contradicts the metric contract operators actually read. Both are fixable inside the current design.

### What this does, in product terms

Before, the WeCom outbound subscriber returned silently in two different situations: the delivery row named another platform, and there was no delivery row at all. After, the first is counted as `not_wecom_turn` at DEBUG, and the second — once the origin gate confirms the input came in over a channel — is counted as `no_delivery_row` at WARN.

That split is the right one. A WeCom task enqueued by an older build that finishes without a `channel_task_delivery` row now leaves an actionable signal instead of a silent return, while Slack/Lark completions on the shared bus stay a low-noise counter. It extends the closed, low-cardinality outcome set from #7428 rather than adding a parallel observability path, and it does not change behavior for normal deliveries that have a delivery snapshot. No config, API, or migration surface.

### Blocking

**1. `task_missing` is now reported as `origin_not_channel`**

`server/internal/integrations/wecom/outbound.go:432` — `channelIngested` collapses `GetAgentTask` returning `pgx.ErrNoRows` into `(false, nil)`. On the WeCom path that verdict flows into `outbound.go:253-259` and is recorded as `outbound_skipped{reason="origin_not_channel"}` at DEBUG.

On `main` the same condition — valid task ID, WeCom delivery row still present, task row reclaimed before the completion event was processed — was `outbound_dropped{reason="task_missing"}` at WARN, in the block this PR removes from `processEvent`. So a genuine missing-data case now disguises itself as "the question was asked in the web UI", which is exactly the diagnosability this PR sets out to restore.

Please keep "task row is gone" as a distinct outcome out of the helper — a typed error or an explicit three-state return both work — and add a regression test for "WeCom delivery row + `taskErr = pgx.ErrNoRows` still yields `task_missing` at WARN".

**2. The new actionable skip contradicts the metric contract**

`server/internal/metrics/wecom.go:93` — the Help for `outbound_skipped_total` still says "none of these is a delivery failure" and enumerates only the original three reasons. `server/internal/integrations/wecom/metrics.go:69-73` and the `skipped` doc comment in `outbound_outcome.go` likewise define every skip as "never owed to a WeCom user in the first place" and not an incident. `no_delivery_row` is defined as precisely the opposite: a reply that may well have been owed, with nothing left that can say to whom.

This is the text an operator reads off `/metrics`, and it now argues against the alert the PR adds. If actionable skips stay in the design, please update those Help strings and interface doc comments — including the package doc at `metrics/wecom.go:41-42`, which enumerates what counts as a skip — along with the corresponding tests.

### Non-blocking

The comment at `outbound.go:200-207` says the origin gate is asked BEFORE the delivery lookup. The code asks it inside the no-row branch and after the WeCom check, which is what the comment at `outbound.go:216-224` explains at length. The two contradict each other; the first reads like a leftover from an earlier arrangement.

`handleInboxNew`'s doc comment is now split. `channelIngested` was inserted at `outbound.go:426`, between the tail of that comment (`outbound.go:420-423`) and `func handleInboxNew` at `outbound.go:448` — so godoc attaches "On any miss — non-member recipient, no wecom binding, no live sender, send failure — the handler is a no-op…" to the origin gate instead.

### How this was checked

Read against head `47df7c3`; all four points above are visible in the code at that commit rather than inferred from the diff alone. CI on this head is green for every backend check; the one red is `cursor-background-lifecycle (windows-latest)`, which fails on process-ownership `Access is denied` in `server/pkg/agent` and does not touch the four files here — worth a re-run before merge, but not evidence against this PR.

Behavior is scoped to `chat:done` / `task:failed` as seen by WeCom-enabled servers. If the judgment above is wrong, it shows up as wrong counters and log levels; reverting the PR restores the previous behavior, and the blast radius is operator diagnosis rather than the send path itself.

seacen:
Both blocking points are fixed, and both non-blocking ones.

## 1. A task row that is gone is a missing row again

`channelIngested` is replaced by `taskOriginOf`, which returns a named three-state `taskOrigin` — `originTaskGone`, `originWebUI`, `originChannel` — with the zero value being the one that fails closed.

Three states rather than a sentinel error: this helper already unwraps `pgx.ErrNoRows`, so a sentinel has to be told apart from it again at every call site, and one reflexive `errors.Is` collapses the state back. Three named values cannot unwrap into one another. A `(bool, bool, error)` triple was the other option and it admits `ingested && taskGone`, which nothing would have guarded.

Both call sites decide for themselves, because only the caller knows whether the reply was owed: with a WeCom delivery row, `originTaskGone` is `dropped(dropTaskMissing)` at WARN, which is `main`'s behaviour restored.

`TestAReapedTaskRowIsStillADrop` covers both — with and without a delivery row. Reverse verification, collapsing `ErrNoRows` to `originWebUI` the way the bool forced:

```
outbound_dropped:task_missing = 0, want 1 — a task row that is gone is a missing
row, not a verdict about where the question was asked. log:
  level=DEBUG msg="wecom outbound: reply not owed to WeCom" reason=origin_not_channel
```

That is the symptom you described, reproduced.

The WeCom switch is now an explicit `case originChannel:` with the skip in `default`, so a fourth state stops rather than falling into the send.

## 2. The metric contract says what the counter now means

Four places, all of which an operator can read: the `outbound_skipped_total` Help, the package doc that enumerates what counts as a skip, the `outbound_dropped_total` Help where it refers to skips, and the interface doc comments on `RecordOutboundDropped` / `RecordOutboundSkipped`, plus the `skipReason` type comment and `skipped()`.

They now say "never went out" versus "went out and failed", enumerate all five reasons, and name `no_delivery_row` as the one worth an alert.

`TestSkippedHelpNamesEveryReasonAndTheOneWorthAnAlert` reads the Help off a gathered family and asserts each label is explained. Nothing asserted that text before. Reverse verification with the old Help restored:

```
help does not explain the "no_delivery_row" label, which an operator will meet
in the breakdown with nothing to read
help still tells the operator every skip is harmless, while no_delivery_row is a
reply that may be owed and unplaceable
```

## Non-blocking

The "asked BEFORE the delivery row" comment is gone; the explanation that was already in the no-row branch is the only one now. `taskOriginOf` moved below `processEvent`, next to its only caller, so `handleInboxNew`'s doc comment is attached to `handleInboxNew` again.

One of the same kind I did not touch: `chatDoneTaskID`'s doc comment is followed by `const taskFailedPrefix`, and it is that way on `main` too.

Rebased on `27d0529fa`. `go test -race` green on `./internal/integrations/wecom ./cmd/server ./internal/metrics`.

Bohan-J (changes requested):
## Request changes

Both blockers from the previous round are fixed, and the product direction still holds. Two things remain: the metrics contract now draws the drop/skip boundary in a place the code does not, and the no-delivery-row branch has quietly become more expensive on a path the comment claims it kept cheap.

### Previous round: resolved

`taskOriginOf` returning `originTaskGone` / `originWebUI` / `originChannel` is the right shape — only an explicit `originChannel` reaches the send, and a reaped task row is back to `outbound_dropped{reason="task_missing"}` at WARN. `TestAReapedTaskRowIsStillADrop` covers both the with-row and no-row branches, and the two misplaced doc comments are cleaned up. Thanks for the thorough fix.

Filing the no-row + no-task-row case as `task_missing` even when the turn may have belonged to another platform is defensible, and the comment argues it well — the missing-task-id branch at the top of `processEvent` already files every platform's ending that way. Not a blocker.

### Blocking

**1. The drop/skip boundary in the docs is not the one in the code**

`server/internal/integrations/wecom/outbound_outcome.go:71-74` now says the split is `"this never went on the wire"` versus `"this went on the wire and failed"`. Twenty lines above it, only one of the five drop reasons fits that:

- `task_missing` is decided before anything is sent.
- `no_live_connection` means there was no socket to send on.
- `attachment_not_admitted` is shed at admission, before delivery.
- `transport_error`'s own comment ends with "…or this delivery's own budget ran out **before it got a turn on the wire**."
- `platform_refused` is the only one that reached WeCom and came back with an error.

So four of five contradict the new definition, and one of them contradicts it inside the same const block. `server/internal/metrics/wecom.go:41-44`, `:91` and `:95` carry the same wording outward — that is the text an on-call engineer reads off `/metrics`, and it now points them away from the alert this PR adds.

The root cause is a swapped boundary. `main` drew it on obligation ("was never owed to WeCom"), which is still accurate. The only thing that stopped being true when `no_delivery_row` arrived is the clause "none of these is a delivery failure" — that one clause needed replacing, not the boundary. Suggest going back to obligation semantics: a drop is a reply that was owed and failed; a skip is one that was not sent because it was not owed, not applicable, or cannot be attributed — with `no_delivery_row` called out as the actionable exception.

`TestSkippedHelpNamesEveryReasonAndTheOneWorthAnAlert` will not catch this: it asserts that the five labels appear and that one specific old sentence does not. Whatever replaces the wording should be constrained by a test that pins the operational meaning, not just the label list.

**2. Every web-UI completion now pays 1–2 extra queries, and the comment says it does not**

`EnqueueChatTask` passes `requireDelivery=false` and writes no delivery row; only `EnqueueChannelChatTask` passes `true` (`server/internal/service/task.go:1870`, `:1889`). So every web-UI task's `chat:done` lands in the no-delivery-row branch — and this subscriber is on the bus for every `chat:done` in the deployment.

On `main` that branch returned immediately: one query in this subscriber, total. Now it runs `GetAgentTask` plus `TaskHasChannelIngestedMessages` (`server/internal/integrations/channel/engine/provenance.go:33`), so two or three.

The comment at `server/internal/integrations/wecom/outbound.go:218-220` says the opposite: "Asking here also keeps the common path — a turn whose row exists — at the one query it costs today." It treats "a turn whose row exists" as the common path, but on a WeCom-enabled deployment the row-less web-UI turn is not the rarer of the two. `TestAnotherPlatformsTurnDoesNotPayForTheOriginGate` then pins the cost guarantee on the row-exists path, leaving the path that actually got more expensive unguarded.

The cost itself is probably acceptable — both are keyed lookups inside an async subscriber — but it should be stated rather than denied. Please either correct the comment to say where the gate's cost actually falls and move the guard onto that path, or avoid the extra lookups there. As written, the one artifact a future reader would consult to learn the cost tells them the wrong thing.

### Product judgment

Worth taking. This does not change delivery at all — nothing is retried, resent, or repaired, and a user who missed a reply still missed it. What it delivers is one WARN line and one Prometheus label. That is enough, because it is not a new surface: it closes two bare `return nil` exits that were counterexamples to the premise of #7428, extending that closed reason set rather than adding a parallel path. No config, API, or migration surface, and no change for turns that have a delivery snapshot.

Worth setting expectations on scope, though: there is no bundled alert rule, dashboard, or admin-side surface, so a deployment only benefits if it already scrapes these metrics and writes its own rule. And since the delivery row is what went missing, the WARN can establish only that some channel turn lost its route — not that it was WeCom's, and not which installation. It carries `reason` and `chat_session_id` and nothing else, so it is a lead to chase, not a diagnosis.

### Verification

Checked at head `dccecb0`. `go test -race ./internal/integrations/wecom ./internal/metrics` — the outcome, origin, and metrics-help tests pass, including `TestAReapedTaskRowIsStillADrop`. A separate group of tests in the same package fails locally because my test database is not migrated to the current schema (`relation "channel_task_delivery" does not exist`); that is my environment, not this PR. `go vet` and `gofmt` are clean, and CI on this head is green.

If the judgment above is wrong, it surfaces as counters and log levels that mislead rather than as undelivered messages; reverting the PR restores the previous behavior, and the blast radius is operator diagnosis, not the send path.

seacen:
Both fixed. You were right about the root cause on the first one: I swapped the axis when only one clause needed replacing.

## 1. The boundary is obligation again

`skipReason`'s comment, `skipped()`, `RecordOutboundDropped` / `RecordOutboundSkipped`, the metrics package doc and both Help strings are back to "was never owed to WeCom" versus "was owed and did not arrive". The only sentence that stopped being true when `no_delivery_row` arrived — "none of these is a delivery failure" — is the only one replaced: the four ordinary reasons are named as never owed, and `no_delivery_row` is called out separately as the one that may well have been owed and cannot be placed.

`wecom/metrics.go:65` carried the same wrong wording and is fixed with them; you did not name it, but it says the same thing to the next reader.

The test no longer counts labels. `TestSkippedHelpPutsEachReasonOnTheRightSideOfTheAlertLine` splits the Help at the alert line and asserts which side each label is on: the `no_delivery_row` sentence must say both "may be owed" and "alert", must not share a sentence with any of the other four, and each of those four must sit in a sentence that says "never owed". `TestDroppedHelpDefinesADropByWhatWasOwed` asserts the dropped Help defines its set by obligation and never by whether the frame reached WeCom — because four of the five never get that far.

Reverse verification, moving `no_delivery_row` back into the harmless sentence:

```
the "no_delivery_row" sentence does not tell the operator this is the label to alert on
"origin_not_channel" shares a sentence with "no_delivery_row", so the Help reads as
though the alert covers it too
```

## 2. The cost is stated where it falls

The comment at `outbound.go:219` now says what is true: `EnqueueChatTask` writes no delivery row, so every web-UI completion in the deployment lands in the no-row branch, and each pays `GetAgentTask` plus the channel-ingested read — two keyed lookups, on the majority of this subscriber's traffic rather than the rarer path.

The guard moved with it. `TestTheWebUITurnWithNoRowPaysForTheOriginGate` pins the count on the path that actually got more expensive, and pins what the two reads buy. Reverse verification, dropping the gate and filing `no_delivery_row` unconditionally:

```
task lookups = 0, want 1 — this runs on every web-UI completion in the deployment
channel_ingested reads = 0, want 1 — same traffic, same multiplier
outbound_skipped:origin_not_channel = 0, want 1 — then the lookups bought nothing. log:
  level=WARN msg="wecom outbound: a channel turn with no route, so nothing was sent"
```

That WARN in the failure output is the alternative currency: skip the reads and the noisiest line in this file fires on every message typed in the browser. `TestAnotherPlatformsTurnDoesNotPayForTheOriginGate` stays — what it pins is still true — but its "common exit" wording is gone, because that path is not the common one.

Also corrected: `skippedFor` said "the ordinary **three**", which has been four since `not_wecom_turn`.

## On the counts

I measured this branch paying two extra reads here (`GetAgentTask` always; `TaskHasChannelIngestedMessages` whenever `ChatInputTaskID` is valid, which since MUL-4351 is every chat task), three in the subscriber in total against main's one. Your "two or three" is the same number counted from the other end.

Rebased on `a843b44a4`. `go vet` clean under both tag sets, `-race` green on `wecom` and `metrics`. The one red check is `cursor-background-lifecycle (windows-latest)` failing in `server/pkg/agent`, which this PR does not touch (`git diff --numstat upstream/main HEAD -- server/pkg/agent/` is empty).

multica-eve (changes requested):
## Request changes

The product behavior is worth accepting, but this implementation makes every ordinary web-UI completion in a WeCom-enabled deployment pay two additional synchronous database lookups while describing those reads as asynchronous and off the user path. Please consolidate the classification and establish a bounded, evidence-based latency cost before merge.

This PR closes a real observability gap: a channel-originated turn that finishes without a delivery row becomes a WARN and a countable `no_delivery_row` outcome instead of a silent return. It does not retry or repair the missing delivery, but it gives operators an actionable lead using the existing closed outcome mechanism. That product shape is appropriate.

### Blocking

**1. Collapse the origin gate into one lookup while preserving all three outcomes.**

For a row-less completion, the subscriber currently performs `GetChannelTaskDelivery`, then `GetAgentTask`, then `TaskHasChannelIngestedMessages`. This branch is the normal branch for web-UI completions because those tasks have no delivery row. The subscriber therefore goes from one query on `main` to three queries here; combining the origin gate would reduce the total to two, not eliminate the added wait.

Please add one query keyed by the completed task ID that returns task existence plus channel provenance (or directly returns `task_gone` / `web_ui` / `channel`). Preserve these semantics:

- a missing task remains `task_missing`;
- retry tasks inspect the original input batch through `chat_input_task_id`;
- a channel-ingested batch with no delivery row remains the actionable `no_delivery_row` case;
- ordinary web-UI and legacy/null-owner behavior remain non-actionable.

Update `TestTheWebUITurnWithNoRowPaysForTheOriginGate` to pin one origin-classification query rather than the current two.

**2. Document and validate the synchronous latency honestly.**

`events.Bus.Publish` invokes type-specific handlers synchronously before `SubscribeAll` handlers. `TaskService.CompleteTask` waits for `broadcastChatDone`, and the daemon `/complete` endpoint waits for `CompleteTask` before returning 200. These reads therefore delay both the completion response and realtime fanout; the statement in `outbound.go` that they are “async, off the user's path” is incorrect.

Please replace that statement with the actual call-path contract, provide a representative query plan/timing for the consolidated lookup, and verify its cancellation/timeout behavior under a slow or unavailable database. The existing shared 10-second outbound-handler timeout is not by itself a suitable budget for an observability lookup on every web-UI completion. The plan should use the `agent_task_queue` primary key and `idx_chat_message_input_owner`, without an unbounded scan.

I reviewed head `0c32f917`. The outcome semantics and metrics-help fixes from the prior rounds are sound; this request is specifically about the newly added synchronous hot-path cost and the test that currently locks it in without bounding its impact.

seacen:
Both fixed at head `e78e72b92`, rebased onto `8c4f4328f`. You were right on both counts, and the second one turned out to be hiding a third thing.

## 1. One lookup

`GetTaskChannelOrigin` is a new sqlc query keyed on the completed task id, returning task existence plus channel provenance in one round trip. `taskOriginOf` makes one call. The row-less branch is now two queries — the delivery row plus this one — down from three, up from `main`'s one, which is what you said to expect.

All four semantics are preserved, and `TestTheWebUITurnWithNoRowPaysForTheOriginGate` pins one origin-classification query rather than two.

**One deliberate narrowing, flagged because it is a behaviour change from the head you reviewed.** The batch key is `COALESCE(chat_input_task_id, id)`, not `chat_input_task_id` with a "NULL means channel" fallback. The old spelling classified every null-owner task as `channel`, which made your fourth bullet — legacy and null-owner stay non-actionable — false as written. The COALESCE is not invented for this: it is the predicate migration 427 used to decide which in-flight tasks were owed a delivery row. No delivery regression follows, because the only writer of a delivery row is `CreateChannelTaskDeliveryFromSession` inside `EnqueueChannelChatTask`, which in the same transaction also stamps the input owner — so a post-420 delivery row always has a non-null owner, and the only null-owner rows that ever had one came from 427's own backfill, which already required the COALESCE'd stamp.

**Plan**, on the real schema dumped into a scratch database with 50k tasks and 200k `chat_message` rows, analyzed, prepared:

```
Index Scan using agent_task_queue_pkey on agent_task_queue t (actual time=0.008..0.009 rows=1 loops=1)
  Index Cond: (id = '...'::uuid)
  SubPlan 1
    ->  Bitmap Heap Scan on chat_message (actual time=0.002..0.002 rows=0 loops=1)
          Recheck Cond: ((task_id = COALESCE(t.chat_input_task_id, t.id)) AND (role = 'user'::text))
          ->  Bitmap Index Scan on idx_chat_message_input_owner (actual time=0.001..0.001 rows=2 loops=1)
Execution Time: 0.024 ms
```

Both indexes you named, no scan. Throughput over 20 000 executions in a server-side loop over random task ids: **13 µs each**, which excludes the client round trip and says so in the comment.

## 2. The latency, said honestly — and what it was hiding

I verified your claim link by link rather than taking it, and it is correct:

| | |
|---|---|
| `cmd/server/router.go` | `POST /tasks/{taskId}/complete` |
| `internal/handler/daemon.go` | waits for `CompleteTaskWithTransition` before writing 200 |
| `internal/service/task.go` | `broadcastChatDone` inline, then `Bus.Publish` inline |
| `internal/events/bus.go` | handlers run in the caller's goroutine, type-specific first, `SubscribeAll` after |
| `wecom/outbound.go` | subscribes to `EventChatDone` — type-specific |
| `cmd/server/listeners.go` | the realtime fanout is the `SubscribeAll` that runs after |

So the read delays both the daemon's 200 and the realtime `chat:done`. "Both are async, off the user's path" is gone and that table is the contract above `taskOriginOf`. Doubly worth correcting because the synchronous shape of `Bus.Publish` is load-bearing elsewhere in this adapter — the bubble PR's binding depends on it — so the package had the same fact written both ways.

**Budget: 2 seconds, on the observability call site only.** 13 µs measured, so that is roughly 150 000× headroom — wide enough that a cold cache or a queued pool acquisition cannot trip it into false drops, and it is the bound this repo already uses for infrastructure calls that must not stall the realtime path. The delivery call site keeps the handler's 10 seconds deliberately: there the read is the precondition for putting words in somebody's room, and those seconds were sized for that room's reply. That 10s is now a named constant rather than a bare literal.

**The third thing.** Verifying the cancellation behaviour you asked about turned up a misclassification: with the gate hanging, head spent the full 10 seconds and then recorded `outbound_unconfirmed{reason="interrupted"}` at WARN — which tells an operator *the user may already have this reply* for a read that never wrote a byte to a socket. `unconfirmedReason` is right to map context errors that way on the send path, so I did not touch it; I classified at the call sites instead. Both origin-gate sites now record `dropTransport`, whose comment already covers "a lookup ahead of it failed… nothing was shown to anybody".

Enumerating that invariant — a failure before anything reaches the socket must not be filed as an unconfirmed delivery — over every early return in `processEvent` ahead of `sendTextCtx` gives six sites. Two were already right, the two gate sites are fixed, and **three carry the same latent misclassification on `main` today** and are untouched here. I would rather name them than widen this diff into code this PR does not own.

## Tests

| test | what it owns |
|---|---|
| `TestTheWebUITurnWithNoRowPaysForTheOriginGate` (updated) | one origin query, keyed on the completed task |
| `TestAnotherPlatformsTurnDoesNotPayForTheOriginGate` (updated) | Slack's turn pays zero |
| `TestAGateFailureIsNotAnUnconfirmedReply` (new, 2 cases) | both gate sites record `dropTransport`, never `unconfirmed` |
| `TestTheObservabilityLookupHasItsOwnBudget` (new) | the budget, under a database that never answers |
| `TestGetTaskChannelOrigin_RealSQL` (new, 6 cases, DB) | the batch-key rule against Postgres |

Reverse-verified, each against the fix it names. Splitting the gate back into two round trips:

```
origin-classification queries = [33333333-... 33333333-...], want exactly one, keyed on the completed task
```

Handing the gate the caller's context instead of its own:

```
the gate was given 9.999920042s; want a budget of its own, at most 2s
the handler took 10.000593209s — the gate spent the whole delivery budget (10s) on a read
that delivers nothing
```

Dropping the COALESCE fallback, and keying on the task's own id:

```
--- FAIL: .../a_legacy_row_with_no_batch_owner,_asked_in_the_room   channel_ingested = false, want true
--- FAIL: .../a_retry_clone_reaches_its_parent's_batch              channel_ingested = false, want true
```

## Existing tests changed

None weakened. Two mechanical renames for the same conditions (`taskErr` → `originErr`, `taskGets` → the recorded id list).

One thing I got wrong and corrected: I removed two fake helpers, `fileRetryClone` and `originAsked`, after checking for callers **on this branch only**. Both are on `main` and both are used by the bubble PR. They are restored, byte-identical signatures. The property `fileRetryClone` existed to observe — that a retry's origin is judged on its parent's batch — moved into SQL where a fake cannot see it, so `TestGetTaskChannelOrigin_RealSQL` pins it against Postgres instead.

Two knobs I did **not** restore: `taskErr` and `taskGets`. This PR removes `GetAgentTask` from `outboundQueries` entirely, so nothing would be behind them — a test setting `taskErr` would pass on evidence no code produced, which is worse than a compile error, since a compile error at least points at the line.

**Merge-order note, since it is invisible to CI.** `TestTheOriginOfARetryCloneIsItsParentsBatch` on the bubble PR asserts the gate was asked about the *parent* id. Here the gate is keyed on the *completed* id, because picking the owner moved from Go into SQL. That is a value conflict with no symbol conflict, so neither git nor the compiler reports it: whichever of the two lands second has to re-point that assertion. The property it protects is preserved by the SQL test above. Happy to take that on whichever way you want to sequence them.

## Checks

`go build`, `go vet` untagged and under `agentintegration`, plus CI's exact `go vet -race -tags agentintegration` form; `-race -p 1` across wecom, channel, engine, slack, lark, dingtalk; the merge with `main` compiled; DB-backed tests proven to have RUN. `VERDICT:FIT`. The generated sqlc file was produced by the repo's documented path and regenerating it afterwards leaves it byte-identical.

Bohan-J (changes requested):
## Request changes

The previous round's items are resolved — the gate is one query, the comment no longer claims the work is async, and the lookup has its own budget. But consolidating the two queries also consolidated two questions that need opposite answers, and the result is a silent delivery regression: the auto-retry of a legacy channel task now classifies as a web-UI turn and its reply is never sent.

I reproduced this against a migrated Postgres at head `e78e72b`, driving the real `CreateRetryTask` and `CopyChannelTaskDelivery` rather than hand-built rows.

### The regression

`GetTaskChannelOrigin` keys the batch on `COALESCE(task.chat_input_task_id, task.id)`. For a retry clone whose parent owned its batch that is right — the clone inherits the owner id and reaches the parent's verdict. For a clone whose parent's owner is NULL it is not: the clone's `chat_input_task_id` is also NULL, so `COALESCE` falls back to the *clone's own id*, and the clone owns no messages. The EXISTS returns false and the turn is filed as web UI.

Every link is in the repo already:

- `server/pkg/db/queries/agent.sql` copies `p.chat_input_task_id` verbatim, and says so on purpose: "A plain copy (not `COALESCE(parent.chat_input_task_id, parent.id)`) is deliberate: legacy/channel parents carry NULL and must stay NULL…". So a NULL-owner parent yields a NULL-owner child.
- `CopyChannelTaskDelivery` (`server/pkg/db/queries/channel.sql:909`) copies the parent's delivery row to the child unconditionally — it does not re-evaluate the predicate migration 427 used. So the clone carries a WeCom route.
- With a WeCom delivery row present, `processEvent` takes the delivery branch and asks the gate. A web-UI verdict there is `o.skipped(ctx, e, skipOriginNotChannel)` and a bare return — nothing is sent, at DEBUG.

Before this commit the same clone was delivered: `TaskInputIsChannelIngested` returned true when `ChatInputTaskID` was invalid, which is the deliver-by-default behavior #5645 shipped and MUL-4351 preserved.

**Reproduction.** Seed a legacy channel task — `chat_input_task_id` NULL, own user message `channel_ingested`, plus the WeCom `channel_task_delivery` row migration 427 would have given it. `GetTaskChannelOrigin` on the parent returns true, as it should. Then call `CreateRetryTask` on it: the child comes back with `chat_input_task_id` still NULL, `CopyChannelTaskDelivery` gives it `channel_type = "wecom"`, and `GetTaskChannelOrigin` on the child returns **false**. Old gate: true. The reply a room is waiting on leaves by the web UI's exit.

The trigger is narrow — a legacy NULL-owner channel task that was in flight when 427 ran, and then auto-retries — but it is exactly the upgrade window this PR exists to make visible, and it fails silently there: the turn is classified as the most ordinary event in the deployment, so not even the new WARN fires. It also propagates: a clone that fails again produces another NULL-owner clone with another copied route.

### Why the fix is not "revert the COALESCE"

The COALESCE is doing real work on the other branch. Without it, a legacy NULL-owner *web* turn with no delivery row reads as channel-ingested and trips `no_delivery_row` — the loudest line here, on ordinary web traffic. That is a genuine problem and worth fixing.

The two branches are asking different questions, and a NULL owner means something different in each:

- **Delivery row present and it says wecom** — the route has already established the turn was WeCom's. The only open question is whether a human typed this particular turn in the browser. An unanswerable origin here should keep failing *open*, as it did before: deliver.
- **No delivery row** — nothing has established the turn was ever a channel's. An unanswerable origin here should fail *closed*: treat it as web UI and stay quiet, which is what `COALESCE` gives you.

So please keep the confirmed-WeCom-route branch on the old NULL-owner semantics and let the diagnostic branch keep the stricter reading. Whether that is two queries again, one query returning enough to let the caller decide, or a parameter, is your call — the constraint is that one verdict cannot serve both branches.

### Test gap

`TestGetTaskChannelOrigin_RealSQL` pins six shapes and passes, including `legacyInRoom` (NULL owner, own messages stamped → true). The seventh shape is the clone of that row — NULL owner, *no* messages of its own — and it is the one that regresses. Please add "NULL owner × auto-retry × copied delivery row" as a regression case, built through the real `CreateRetryTask` / `CopyChannelTaskDelivery` so the fixture cannot drift from what `FailTask` actually writes.

### Verified

At head `e78e72b`, against a freshly migrated database: the existing `TestGetTaskChannelOrigin_RealSQL` passes (all six cases), and the seventh case above fails as described. CI on this head is green.

Product direction is unchanged from my last review — still worth taking. This is the one thing standing between it and merge.

seacen:
Fixed at `49de4661f`. The regression reproduces exactly as described, and the fix is the third of the three options you named: one query returning enough for each caller to decide.

### The second fact

```sql
EXISTS (
    SELECT 1 FROM chat_message
    WHERE task_id = COALESCE(task.chat_input_task_id, task.id)
      AND role = 'user' AND channel_ingested
) AS channel_ingested,
(task.chat_input_task_id IS NULL)::boolean AS batch_owner_unknown
```

`channel_ingested` is unchanged. `batch_owner_unknown` says whether the verdict above can speak for this row at all: with a NULL owner the EXISTS was keyed on the task's own id, so `false` there means "owns no messages", not "typed in a browser".

Two queries would put the second read back on the path the last round reduced to one; a parameter would make the query answer a question only the caller can ask. The fact is a property of the row, so it comes back with the row.

### The two defaults, at the call sites

`taskOriginOf(ctx, taskID, failOpen)`:

- `outbound.go:304`, **`failOpen=true`** — the delivery branch. A route already established this turn was WeCom's; the only open question was whether a human typed this particular turn in a browser, and an unanswerable one does not override the route. This is the deliver-by-default behaviour #5645 shipped.
- `outbound.go:262`, **`failOpen=false`** — the diagnostic branch. Nothing established the turn was ever a channel's, so an unanswerable one stays web UI and quiet rather than tripping `no_delivery_row` on ordinary web traffic.

Your reasoning for both is now the contract above the function, because a reader who finds two callers passing opposite booleans and no reason will unify them — which is how this happened in the first place.

### The seventh shape

`TestGetTaskChannelOrigin_RealSQL` seeds it the way you asked: `CreateRetryTask` on a NULL-owner channel parent, then `CopyChannelTaskDelivery` from it. Two assertions keep the fixture from drifting into something else — the clone's `chat_input_task_id` came back NULL (if `agent.sql` ever starts COALESCE-ing, this fails rather than quietly stopping to cover the case), and `GetChannelTaskDelivery` on the clone returns the parent's room, since without a route it is not a row anyone is waiting on.

Two caller-level tests cover what the SQL test cannot see, one per default:

- `TestTheAutoRetryOfALegacyChannelTurnStillAnswersTheRoom` — delivery row present, verdict says web UI, owner unknown: one frame goes out. With the fail-open branch removed it reports `frames = 0`.
- `TestNonWecomSessionIsNotADrop/a_legacy_row_with_no_batch_owner_and_no_delivery_row` — the same row with no route stays `origin_not_channel` at DEBUG. With the diagnostic branch failing open too, it reports `= 0, want 1`.

Both were reverse-verified against the change itself, not against a compile error.

### Verified

`verify/verify.sh` at `49de4661f`: gofmt, `go vet` under both tag sets, `go build`, `-race -p 1` on the touched packages, and the DB-backed tests confirmed RUN rather than SKIP against a migrated Postgres — including the seventh shape.

Bohan-J (changes requested):
## Request changes

The delivery regression from my last review is fixed, and fixed the right way. One item remains, and it is mechanical.

### Resolved: the legacy retry clone is delivered again

`GetTaskChannelOrigin` now reports `batch_owner_unknown` alongside `channel_ingested` and leaves the decision to the caller: the confirmed-WeCom-route branch passes `failOpen=true` and keeps the deliver-by-default behaviour for an owner-less row, while the no-delivery-row branch passes `false` and stays quiet. That is exactly the split the two branches needed, and the comment on `taskOriginOf` explains it well.

I checked this independently at head `49de466` against a freshly migrated database:

- `TestGetTaskChannelOrigin_RealSQL` passes all seven shapes, including the new "auto-retry of a legacy row with no batch owner" case, built through the real `CreateRetryTask` and `CopyChannelTaskDelivery`.
- Reverse check: flipping the delivery-branch call to `taskOriginOf(ctx, taskID, false)` fails `TestTheAutoRetryOfALegacyChannelTurnStillAnswersTheRoom`, so the fix is actually guarded.
- The full `./internal/integrations/wecom` and `./internal/metrics` packages pass under `-race` with the database available, and `go vet` is clean. CI on this head is green.

### Remaining: move `origin_gate_db_test.go` onto the `testutil` fixtures

`AGENTS.md:118` asks DB-backed Go tests to use the `server/internal/testutil` fixtures (`dbfx.Task`, `dbfx.Insert`, …) and to keep product assertions in the test. `origin_gate_db_test.go` seeds with hand-written `INSERT`s and tears down with a hand-maintained `DELETE` chain in `t.Cleanup`.

I know the neighbouring `_db_` tests in this package do the same. I'm not asking you to migrate those. This is a new file, though, and the fixtures remove a real maintenance hazard here: they clean up after themselves, so the teardown can't drift out of step with the seed as more shapes get added.

It's mechanical. Every row this file creates already has a fixture — `Workspace`, `User`, `Agent`, `Runtime` and `Task` directly, plus `Insert` / `InsertNoID` for `chat_session`, `chat_message` and `channel_task_delivery`, with `dbfx.Raw("now()")` for `completed_at`. Roughly:

```go
f := dbfx.New(pool, "", "")
f.WorkspaceID = f.Workspace(t, "origin "+tag, "origin-"+tag, dbfx.Cols{"id": wsID})
f.UserID = f.User(t, "Origin "+tag, "origin-"+tag+"@example.com", dbfx.Cols{"id": userID})
f.Agent(t, "origin-agent-"+tag, "", dbfx.Cols{"id": agentID, "runtime_mode": "local"})
f.Task(t, agentID, dbfx.Cols{
    "id": id, "chat_session_id": sessionID, "status": "completed",
    "completed_at": dbfx.Raw("now()"), "chat_input_task_id": owner,
})
f.Insert(t, "chat_message", dbfx.Cols{ /* … */ })
f.InsertNoID(t, "channel_task_delivery", dbfx.Cols{ /* … */ })
```

Please keep the parts that give the test its value: the real `CreateRetryTask` / `CopyChannelTaskDelivery` calls for the legacy-retry shape, and the per-case `why` diagnostics. I tried this rewrite locally against the current head, and all seven shapes plus `TestTheAutoRetryOfALegacyChannelTurnStillAnswersTheRoom` pass.

With that done I have nothing else blocking, and I'll approve.

Bohan-J (approved):
Thanks @seacen — this was a long review, and you took every round seriously. The PR ended up much better than where it started: the drop/skip split is drawn by obligation rather than by whether a frame was written, an owner-less batch gets the right fallback on each side of the route check, and the legacy retry shape is covered through the real `CreateRetryTask` / `CopyChannelTaskDelivery` rather than hand-built rows.

To close out the last round without sending it back to you again, I pushed two small follow-up commits to your branch:

- `e38ea27` moves the legacy retry's preconditions out of `seedOriginLab` and into the one case they guard, so the fixture helper only reports setup failures.
- `f2b0ee4` gives `route_unattributable` its own log line and its own sentence in the `outbound_skipped_total` Help, since "not owed" is the one thing that exit can't claim. Both are pinned by tests.

CI is green on `f2b0ee4`. Approving and merging — thanks again for sticking with it.

PR #8354, opened 2026-09-12
MUL-7322: fix(autopilot): show last run status in list

AtlieX:
Fixes HBH-627

## Summary
- Add LAST_STATUS to the default `multica autopilot list` table so failed last runs are visible beside active autopilots.
- Render missing last-run status as `—` instead of an empty cell.
- Update the built-in autopilot reference and add a CLI regression test.

## Verification
- `PATH=/home/godfinns/local/go/bin:$PATH make build` — pass
- `/home/godfinns/local/go/bin/go test ./cmd/multica -run 'TestRunAutopilotListTableShowsLastRunStatus|TestRelativeTimestampAt' -count=1` — pass\n- `/home/godfinns/local/go/bin/go test ./...` — fail in unrelated `server/internal/daemon/repocache` Co-authored-by hook tests on this host\n- `GIT_CONFIG_GLOBAL=/dev/null /home/godfinns/local/go/bin/go test ./...` — fail in unrelated `server/internal/daemon/execenv` Cursor MCP config tests on this host\n- `HOME=$(mktemp -d) GIT_CONFIG_GLOBAL=/dev/null GOMODCACHE=/home/godfinns/go/pkg/mod GOCACHE=/tmp/multica-go-cache /home/godfinns/local/go/bin/go test ./...` — fail in unrelated `server/internal/daemon/execenv` Cursor MCP config tests on this host


Bohan-J:
Thanks for this — merging.

I re-verified the system side rather than just the diff, and the scoping is right: `last_run_status` is already derived in `ListAutopilots` (most recent `autopilot_run.status`, `COALESCE`d to `''`), already returned by the list handler, and already consumed by the web `LastRunCell`. The CLI table was the only surface dropping it — no migration, no new column, no new query, no new API field. This is the same shape of fix as MUL-6680 did for `NEXT_RUN`, in the same function.

A few things I specifically checked:

- `displayLastRunStatus` passes unknown values through verbatim instead of matching against a local enum, which matches the documented decision in `packages/core/types/autopilot.ts` and `packages/core/api/schemas.ts` that these stay open strings so future server-side statuses degrade gracefully.
- `—` for the empty case is the established placeholder in this CLI (`relativeTimestampAt`, `cmd_issue.go`), so it reads consistently.
- `strVal` returns `""` for a missing or nil key, so the handler's `omitempty` omission on a never-run autopilot lands on `—` as intended.
- `--output json` is untouched, so machine consumers see no change; the only user-visible change is one extra column in the default table.
- Leaving `autopilot get` alone is correct — the derived columns are list-endpoint-only, so adding it there would mean a new query.

Verification on my side: `go build ./cmd/multica` passes, and the full `./cmd/multica` package passes. The `./...` failures you saw are host-environment issues, not yours — I reproduced the same pattern and confirmed the one remaining failure in my run (`TestDaemonStopAcceptsDaemonWithoutProfileField`) fails identically on `main`.

One non-blocking nit for next time, no need to act on it here: the new test asserts the `failed` rendering but not the empty-field → `—` branch, which is the half more likely to regress quietly.

PR #8355, opened 2026-09-13
MUL-7324 feat(wecom): answer in the message the question opened

seacen:
Reopening with the reason you closed it removed.

You closed this because the shared `channel/engine` change did not clear the #8125 bar: a WeCom-only need, threading a batch identity through `router` and `batcher` and changing four other adapters only to ignore it. That is gone.

```
$ git diff --numstat upstream/main HEAD -- \
    server/internal/integrations/channel/engine/ \
    server/internal/integrations/dingtalk/ server/internal/integrations/lark/ \
    server/internal/integrations/slack/ server/internal/integrations/telegram/
$
```

Empty. No `RunBatchID`, no `OnRunStarted`, no widened `TypingNotifier`, no signature churn in the four adapters that never read it.

## How, without inferring anything

The question a bubble has to answer is "which run am I for", and the reason it was an engine change is that only the debouncer knows where one run ends and the next begins. It still does — and it already says so, on the bus.

`protocol.EventTaskQueued` carries `TaskID` and `ChatSessionID` (`service/task.go`, `taskEvent`), and a chat run publishes it from `FinalizeChatTaskEnqueue`, which `engine/router.go` calls at the flush. A media-deferred turn publishes it later, from `PromoteChannelChatTasksIfMediaReady`. So the boundary is still the flush, and it is still carried in rather than re-derived — the adapter listens for the fact instead of being handed it.

Three rules, all inside `server/internal/integrations/wecom/`:

1. **A message joins the open bubble if that bubble has never been bound to a run**, otherwise it opens its own. "Never bound" is the debounce window, expressed without a clock.
2. **`task:queued` binds the oldest unbound bubble.** One that arrives before its bubble is painted waits in a per-session FIFO and pairs when it lands.
3. **Endings seal by task id.** `OnSettled` closes the oldest never-bound bubble — the flush that produced no run.

Two events look like a chat run and are not, and both are settled by a fact rather than a guess:

- An `/issue` task carries a chat session too. It is filtered on an empty `issue_id`, which is a property of the row: `CreateChatTask` inserts `issue_id` as `NULL` (`pkg/db/queries/chat.sql`).
- A retry clone's `task:queued` is broadcast **before** its parent's `task:failed` (`service/task.go:4979` and `:5041`, same function). So `retry_pending` releases the dead attempt from its round rather than sealing it, which leaves the round waiting for a run again — correct whichever order the two events arrive in, and both orders have a test.

## What it looks like on a real tenant

Run against our own deployment on 2026-09-13, same binding code, three cases in one session:

| what was sent | what the engine did | what the chat showed |
|---|---|---|
| two messages 1.3s apart | one task, 12:36:08 | **one** bubble, answered in place |
| a third while the first run was still going | its task queued 12:36:30, started 12:36:41 behind the first | **two** bubbles on screen at once, each replaced by its own answer |
| one message with a file | task 12:48:36 | bubble answered in place, file as its own message |

No fallback, no refused frame, no bubble error branch in the log — and DEBUG was on, so the silence is evidence rather than absence.

The middle row is the case this design exists for. Session granularity, the shape suggested when this was closed, cannot produce it: on this deployment 36 of 139 chat runs (25.9%) were enqueued while the previous run in the same session was still going, and each of those is a question that would get no receipt at all.

## Best effort, with the floor named

The bubble is best-effort by construction, and the floor is deliberate: whenever anything prevents a seal, the answer goes out as the plain message `main` sends today, and the platform ends the unsealed message on its own at ten minutes — "需在 10 分钟内完成所有刷新并设置 `finish=true`，否则消息将自动结束" (doc 101463). So the worst case is today's experience, and the ordinary case is the answer arriving in the message that asked for it.

A best-effort claim is worth what its list of cases is worth, so here is the list.

Every ending reaches the round wherever it lives. `relayKindSeal` carries a task id and a reason but no words: it closes the round if this replica holds it and does nothing at all if it does not. So an answer produced off-lease, a cancellation, a completion with nothing to say and an attachment-only answer all end the bubble on the replica holding it, and the wordless ones are worded by that holder in the round's own locale rather than shipped across.

Two cases are left:

- The frame does not arrive — the holding replica restarted, or the relay dropped it. The store is in memory on purpose; either costs the in-place reply, not the answer.
- The server refuses the closing frame with 846605 / 846608.

Two more come with binding off the bus, and both cost a receipt rather than an answer:

- A media-deferred turn publishes `task:queued` late, so a message arriving in between joins the first bubble while the engine may already have given it a run of its own. That run's answer arrives as a plain message.
- A question typed in the web UI on a WeCom-bound session looks identical on the bus. In the ≤3s window between a bubble being painted and its own run being queued, it can take that bubble — and the origin gate then refuses to seal a web answer into a WeCom room, so the WeCom answer falls back to a plain message rather than landing anywhere it should not.

All five are countable now: the second commit adds `stream_opened`, recorded where the handle is kept rather than where a frame was accepted, so `opened − finished − fell_back` is the number of bubbles nobody ended. That number is the one an operator can act on, and it did not exist before.

## The one shared file left

`internal/events/bus.go` gains `SubscriberCount` — five lines, read-only, no behaviour change for any caller. It is there because "exactly one handler owns this event" is not expressible behaviourally: with two handlers the outcome depends on which the bus reaches first, so the test passes half the time and the defect is a second message in somebody's chat. The accessor is generic rather than WeCom-shaped, and `cmd/server`'s wiring guard uses it too. Happy to find another way if you would rather the bus stayed closed.

## Size

33 files, +7,276/−290 — production +3,388, tests +3,888. Outside `server/internal/integrations/wecom/`: `cmd/server/router.go` +91/−11 (boot), `internal/metrics/wecom.go` +14 (the counter), `internal/events/bus.go` +12. The branch you closed in August was +12,353 across 49 files.

Green on `./internal/integrations/wecom ./internal/integrations/channel/engine ./internal/metrics` with `-race`; the three DB-backed tests run rather than skip. `cmd/server` carries three comment-test failures that `upstream/main` has at the same commit.


multica-eve:
Thanks for rebuilding this and for making the trade-off explicit. We agree that the ACK fix addresses the earlier transport-level concern, and that the usual outcome here is no worse than the receipt-plus-answer flow.

We are still closing this PR for the current architecture boundary. Since #8125, changes to the shared `channel/engine` path need a demonstrated need from more than one channel. This is a WeCom-only need: threading the batch/sequence identity through `router` and `batcher`, then changing the other channel implementations only to ignore it, does not clear that bar.

For the apparent conflict with the earlier guidance: #8125 is the current rule. Please do not thread a batcher/router identity through the shared engine for this. If we revisit this after concrete member feedback about the extra completion message, an adapter-local, session-keyed implementation (with its collision behaviour documented) is the direction to evaluate.

There is not currently that member feedback, so we are not reopening the product decision now. This closure is independent of the review of #7954.

seacen:
@Bohan-J — reopening rather than filing a new one, because the thing you closed this for is the thing that changed.

Your words were: "This is a WeCom-only need: threading the batch/sequence identity through `router` and `batcher`, then changing the other channel implementations only to ignore it, does not clear that bar." That was right, and it is no longer what this does. The diff against `channel/engine` and the four other adapters is empty — the description opens with the command and its output.

The replacement is not the session-keyed shape you suggested evaluating, and I would rather say why than quietly substitute something else. Session granularity loses the second receipt whenever a question arrives while a run is still going, which on our deployment is 36 of 139 chat runs. The bus already publishes the run boundary — `task:queued`, with the session and the task on it, from the same flush that would have called `OnRunStarted` — so the adapter can listen for the fact instead of being handed it. Nothing is re-derived and no clock is consulted, which was the property the engine change was there to buy.

Two things building it settled that review would not have:

- A retry clone's `task:queued` is published **before** its parent's `task:failed`. A design that flags the failure and then looks for the clone cannot work, because that ordering does not exist. Releasing the dead attempt from its round instead is correct in both orders, and both have a test.
- `/issue` tasks carry a chat session and would have taken a waiting bubble. They are filtered on `issue_id`, which `CreateChatTask` inserts as NULL — a property of the row rather than a heuristic.

On the product question you left open — no member complaint about the second message — I am not going to argue it, and I have not gone looking for one. What I can add is evidence: we have been running this on our own tenant, and the description has what three real turns looked like, including two bubbles open at once and each answer landing in its own. If you would still rather wait for demand, say so and I will leave it there; I would like the architecture settled either way, since that half is no longer a matter of taste.

One follow-up is ready and deliberately not in here: sealing a relayed reply into its bubble, which closes the multi-replica case named in the description.

Bohan-J (changes requested):
## The product decision, first — because it is the one you asked for

**We are taking the bubble.** The architecture question you asked to have settled is settled: this shape is the one we want, and we are not going to send you back to the receipt. #7954 is closed as superseded.

Two things changed our mind, and you should have both:

**The gate we set was not one anyone could meet.** We said the bubble should be argued "from someone complaining about the second message". But the measurements can only come from your deployment, so the complaint we were waiting for could only ever reach us through you. We were asking you to produce evidence independent of yourself, about a tenant that is yours. That was not a fair condition, and it stood for six weeks.

**The `channel/engine` objection is genuinely gone.** We verified it rather than taking the description's word: the diff against the shared engine and against dingtalk/lark/slack/telegram is empty, and binding off `task:queued` keeps the boundary at the flush without the engine handing anything down. That clears #8125.

So: **no new conditions after this review.** Everything we want is in this comment. What follows is implementation only, and none of it reopens the direction.

---

## Why this is Request changes

Two invariants this feature has to hold — *the answer is never sent twice*, and *the answer is never lost* — are both breakable today, on the same four lines. We reproduced both. One of them is a regression against `main`.

### 1. `outbound.go:400-415` — the answer is sent twice, and (separately) not at all

```go
if err := o.finishStream(ctx, t.Handle, text); err == nil {
    o.delivered()
    return answerOutcome{addr: t.Handle.address(), spoke: true}, nil
}
// ... any error falls through to:
content = text
```

**Doubled.** Any non-nil `seal` error is treated as "the frame did not land", so the same text goes out again as a plain message. But `errStreamAckTimeout` and `errWriteAttempted` are not proof of non-delivery — `errWriteAttempted`'s own doc says the frame "may have reached the peer and been acknowledged". This package already makes exactly that distinction in three other places (`outbound_outcome.go:120`, `outbound_media.go:490`, `relay_outbound.go:1353`'s `provablyNotSent`); this is the one site that does not.

It also cascades rather than being rare. `cancelAck`'s own comment (`ws_sender.go:388-405`) states that a genuinely lost ack makes *every later frame on that req_id* time out. So after one lost ack, every retry lands and every retry reports failure: the bubble holds the answer and the user gets a second copy underneath.

**Lost.** The seal retry budget and the fallback share one context, and the arithmetic does not fit:

| | |
|---|---|
| `ackTimeout` | 5s (`ws_sender.go:152`) |
| `streamCloseRetryDelay` | 2s (`stream_store.go:146`) |
| one retry therefore costs | 5 + 2 + 5 = **12s** |
| caller's budget `streamCloseTimeout` | **10s** (`typing_indicator.go:81`) |

When the context dies, `seal` returns `ctx.Err()`, and the fallback's `sendAsMessage` runs on that same expired context — `ws_sender.go:537` returns immediately and nothing is sent. On `main` the answer always got the full budget for its plain message; here the bubble spends it first. The WARN on the way out says "in-place reply failed, sending a new message instead" while no message is sent. `typing_indicator.go:782-793` has the same shape for the failure notice.

A corollary worth naming: **`streamCloseRetries = 3` is unreachable in production.** At 10s only two attempts fit. The retry policy is exercised only under `stream_close_retry_test.go:31-33`'s injected `ackTimeout=50ms / closeRetryDelay=0`; no test runs the shipped ratio, so the constants and the budget are mutually inconsistent and nothing catches it.

**Direction:** fall back only on `streamUnusable` (846605 / 846608 — a stated refusal, the one proof of non-delivery) and `provablyNotSent`-class errors; treat written-but-unacked as delivered-unconfirmed. Give the fallback its own deadline (`context.WithoutCancel` + a fresh timeout) so seal cannot eat it.

### 2. `ws_sender.go:388-405` — a refusal can be recorded as delivery

`cancelAck` removes the waiter when a caller gives up, but that frame is still unanswered at the server, so the next frame passes the one-in-flight gate immediately. Positional matching in `deliverAck` (`w.seq == st.acked`) can then hand a later-arriving ack to the closing frame.

This is not hypothetical for this protocol — it is the behaviour **you measured**: acks come back grouped by outcome rather than in write order (24 frames back-to-back; 11 of 12 concurrent refreshes refused with 6000). In that ordering the closing frame's own 6000 refusal is dropped on a seq mismatch and the opening frame's late `errcode 0` satisfies it. `finishStream` returns nil, `o.delivered()` fires, no fallback, `recordEnding` files it as `stream_finished` — and the answer went nowhere.

`TestALateVerdictIsNotHandedToTheNextFrame` covers the stale-ack-arrives-first ordering, which the counter already handles. The inverted ordering — the measured one — is untested.

The trigger threshold is lower than it looks: the opening frame runs on the router's `replyTimeout`, default **2500ms** (`channel/engine/router.go:88-90`, not overridden in production wiring), well under the 5s `ackTimeout`, so it rarely gets its full budget before `cancelAck` frees the slot.

Same family, one line: `ws_sender.go:668-671` calls `abortStreamFrameLocked` on **any** write error, including `errWriteAttempted`, returning the sequence slot for a frame that may be on the wire. Suggest `if !errors.Is(err, errWriteAttempted) { s.abortStreamFrameLocked(reqID) }`.

### 3. Bubbles bind to the wrong run

**A first-party run can take a WeCom bubble.** `handleTaskQueued` (`typing_indicator.go:380-395`) filters only on `issue_id != ""`, but `CreateChatTask` inserts `issue_id` as NULL for *every* chat task — WeCom, web, desktop, mobile — and `taskEvent` stamps the same `chat_session_id`. The events are byte-identical, so the installer asking in their browser (the premise `failure_origin_test.go` opens with) binds the room's bubble. Three consequences:

- `typing_indicator.go:480`'s `has()` shortcut skips the origin gate entirely once a round is bound, so a **web run's error text is written into the group chat** — the thing `TestAWebUIRunsFailureIsNotAnnouncedInTheRoom` exists to prevent, reached by a path no test covers (that test files a task row but never publishes `task:queued`).
- `handleTaskCancelled` (`:619-643`) has no origin gate at all.
- When the web run *succeeds*, `chat:done` is refused by the gate and returns without taking or forgetting the round — the bubble spins until the platform ends it and the real WeCom answer degrades to a plain message.

**A retry clone and a new question swap bubbles.** `retryUnbind` (`stream_store.go:601`) clears `taskID` but keeps `everBound=true`, while `unboundLocked` (`:397`) matches on `taskID == ""` only. A retry-released round and a genuinely new round then compete, and the positionally older one wins regardless of which run is arriving. The comment at `:583-587` says "This works in either publish order" — that holds only when there is a single round on screen, and a `runtime_offline` backoff is long enough that a follow-up question is likely rather than exotic.

**A stale pending entry hijacks the next question's bubble.** Anything that produces a run without painting a bubble — `OnIngested` returning early, an unreadable envelope, a refused opening frame followed by `drop` — leaves the task in `pending` for 10 minutes, where the next message's fresh bubble pairs with it. Suggest bounding `pendingRun` by the ingest goroutine's own deadline (seconds) rather than `streamMaxAge`.

### 4. Multi-replica: the failure notice regresses

`main` relayed the failure notice as a `relayKindReply` when the local replica held no socket. The new path has no relay at all — `typing_indicator.go` contains zero `relay` references; `sayAsPlainMessage` and `writeClosing` call `m.senders` directly and only WARN on `errNoLiveConnection`. Since the WS lease guarantees exactly one replica holds the socket, on any multi-replica deployment this is the normal case, not an edge: no failure notice, plus a bubble spinning for ten minutes.

This also qualifies the "worst case equals today" claim in the description. For the answer path it holds. For the failure/cancel path it does not, and `deliverRelayed` leaving the lease holder's bubble unsealed means even the successful relay case is today's behaviour **plus** a stale spinner.

### 5. Unsealed bookkeeping accumulates where those bugs leave it

`pruneStreamsLocked` (`ws_sender.go:448-480`) retires only entries that are sealed *and* older than `streamMaxAge`; an entry that is never sealed is never removed for the life of the connection. The reasoning for not evicting live turns is correct and well argued — evicting one would cause exactly the misattribution in §2 — and the over-cap warning is the right instinct. The concern is that the paths above make "never sealed" considerably more common than the design assumes, so this is worth re-checking once they are fixed rather than treated as a separate defect.

---

## Not blocking, but worth fixing here

- **`QueuedBehind` renders as "已并入上一条回复" and that is not what it means** (`outbound.go:389-398`, `stream_store.go:518`). It records only that another round existed when the bubble was painted; a queued round has its own run and by construction was not merged. It is also tested before `carriesFiles`, so a round with attachments and no text claims it was merged and then sends the files — the self-contradiction `strings.go:122-126` says this copy exists to avoid.
- **Cancellation now enters a network path of up to 10s on a synchronous bus.** The failure path could already block on the network on `main`, so that part is not new, but `task:cancelled` is, and `BroadcastCancelledTasks` publishes in a loop — a bulk cancel scales linearly.
- **`RouterOptions.WecomResolverSetBuilt`** is a test-only callback on a production options struct, and it is weaker than it looks: it fires *before* `channelRouter.Register`, and `engine.Router.Register` drops an incomplete set with only a `Warn`, so a set missing `Audit` still passes the guard. A read-only `ResolverSetFor(t channel.Type)` on `engine.Router` (~8 lines, the same justification you used for `Bus.SubscriberCount`) would let the test assert what was *registered*, and delete the production field.
- **`events.Bus.SubscriberCount` we are happy to take.** Read-only, correctly `RLock`ed, and there is genuinely no existing way to ask the question without a live pool the boot test deliberately does without.

## Tests

The wire-level assertions are good — the `said()` union helper in `stream_bubble_test.go:769-797` is the right idea, and it is the only thing structurally able to catch "sealed *and* pushed". Three notes:

- `stream_bubble_db_test.go:193-197` asserts the stranded spinner as the **correct** outcome. Closing the relay gap later turns it red, which makes the gap harder to close than it was before.
- `stream_close_retry_test.go:261` is named "...GoesOutAsAMessageOnce" while its own doc says `errWriteAttempted` may have reached the peer. It pins §1 as intended behaviour.
- `language_lint_test.go:76` scans raw bytes for `localeForSender(`. In a codebase with comments this dense, **writing a comment that mentions `localeForSender(` breaks the build.** An AST walk (as at `:112`) would be safe.

## What we verified

`go build ./...` and `go vet` pass. **111 of this PR's own bubble/stream/ack/binding/origin/locale tests pass under `-race` with no data races.** Findings 1 and 2 were read line by line against the source and the constant arithmetic checked by hand. The DB-backed tests we could **not** run — our local Postgres is reachable but not fully migrated — so the multi-replica and real-query-layer paths are unverified on our side, and finding 4 rests on reading rather than execution.

Separately: **the red CI is not yours.** `backend-tests` fails on `TestMigrationNumericPrefixesAreUnique` — a duplicate migration prefix — and this PR touches no migration files. A rebase should clear it; please do not go chasing it.

---

The design reasoning in these files is unusually careful, and several of the comments diagnose the real hazards precisely. The gap is between that reasoning and where it is enforced: §1 and §2 are both cases where the comment states the correct rule and the code a few lines away does not implement it. Fix those two and the rest is small.

seacen:
Thank you for saying the gate was not a fair one. It was the part I could not argue with and could not meet, and you withdrew it before I asked again.

All five are fixed. Each has a test written before the change, and each was broken again afterwards to check the test would have caught it. Two of them changed what I thought the fix was, and I have said where.

## 1. The answer is sent once, and it is sent

**Doubled.** The fallback now runs only on proof that nothing landed: a stated refusal (`streamUnusable` — 846605 / 846608) or a `provablyNotSent` error. Written-but-unacked is recorded as delivered-unconfirmed, because `errWriteAttempted` says the frame may have reached the peer and `cancelAck` makes that the cascade rather than the exception. In a chat with no unsend, a duplicate is permanent and an unconfirmed delivery can be re-asked; the duplicate is the error with no way back.

Underneath it was a third instance of the distinction you named: **`provablyNotSent` did not know `errStreamAckTimeout`** and fell to `default: return true`, reporting a written frame as provably unsent. That is the step that let the second copy out.

```
the answer also went out as 1 plain message(s) — the person reads it twice
and WeCom has no unsend
```

**Lost.** The retry no longer starts an attempt the remaining deadline cannot finish, so `streamCloseRetries` is an upper bound and the deadline is the authority. The fallback takes its own deadline (`context.WithoutCancel` + a fresh timeout) when what is left cannot cover one push.

Your arithmetic is pinned as a test rather than a comment, and it found something on the way: with the classification fixed, the "fallback on an expired context" path is no longer reachable at all — an expired seal returns a context error, which is not proof, so it is recorded as unknown. The case that IS reachable is a seal that spent most of the budget and then read a real refusal, leaving the plain message less time than one push needs. The guard tests remaining budget, not whether it is already gone, and the test drives it that way.

## 2. A verdict is not guessed

Fixed at the premise rather than at the symptom. An ack carries only `req_id`, so position identifies a frame only while the server owes nothing else on that `req_id` — and `cancelAck` used to free the gate while the server still owed. No frame goes out while a verdict is outstanding; a caller that cannot wait gets `errStreamBusy`, which is provably-not-sent, so the answer goes out once by the plain route.

seal's retries are exempt, and that exemption is the measured part: an identical frame is not a second frame. Re-writing a sealed stream returned errcode 0 on all six attempts against the live tenant on 2026-09-03 (STRATEGY §6.5), so whichever of the identical writes a verdict belongs to, it reports the same outcome. Without the exemption the retry is blocked, the first frame stays unanswered, and the answer is repeated by the plain route — the bug the whole path exists to avoid.

`ws_sender.go:668-671` takes your one-liner: `abortStreamFrameLocked` is skipped for `errWriteAttempted`.

The two tests that described the old contract are replaced rather than adjusted. `TestNoFrameIsWrittenWhileTheServerStillOwesAVerdict` asserts one frame on the wire and a provably-unsent error; `TestTheClosingFrameGoesOutOnceTheOwedVerdictArrives` asserts that once the debt is settled the closing frame goes out and reads its own refusal.

## 3. Bubbles bind to the right run

**The `has()` shortcut is gone.** You are right that it stopped being sound the moment binding moved to the bus: `CreateChatTask` inserts `issue_id` as NULL for every chat task, so a browser run's event is byte-identical to the room's. Reproduced before fixing:

```
the room's bubble was sealed with a browser run's failure (1 closing frame(s))
```

The gate runs for every ending now, delivery row first so another channel's run still stops at the first read. `handleTaskCancelled` has the same gate; it had none.

**A refused run releases the round.** Without that the bubble stays bound to a run that will never close it and the room's own answer degrades to a plain message.

**And this is where I was about to introduce a regression.** Running the gate unconditionally made a database outage withhold the notice for WeCom's own runs — `TestAnOpenRoundIsProofEnoughOfOrigin` existed for exactly that and went red. The gate now has three answers, not two: *not ours* releases the round, *unknown* withholds the notice and releases nothing, because an unreachable database is not evidence about where a run came from — releasing there would hand away a bubble whose answer is still coming. That test is rewritten to the contract that is now true, since its premise is the one you identified as unsound.

**Retry clone and stale pending.** `retryUnbind` leaving `everBound` set is handled by the gate's release path; a pending run is bounded by `pendingMaxAge` (30s — the ingest goroutine's own clock) rather than the protocol's ten minutes, so it cannot pair with a much later question's bubble.

## 4. The notice reaches the room from any replica

The indicator routes again. `WithRelay` gives it the same `relayKindReply` path `main` used, taken when this replica holds no socket for the installation, and the frame carries the task id so the holder seals the right bubble rather than pushing under it. Wired at boot next to the subscriber's own relay, so it is not a seam that exists only in tests.

## 5. Bookkeeping, re-checked after the rest

Worth re-checking, as you said — and after 1–4 "never sealed" is more common, not less, since the gate now declines frames it used to write. `pruneStreamsLocked` retires an entry when it is **settled** (`acked == sent`) and old, rather than when it is sealed and old.

My first version used age alone, and `TestPruneKeepsLiveTurnsEvenOverTheCap` caught it: a `req_id` outlives its stream. The platform ends a stream at ten minutes, but a new stream id on the same `req_id` is accepted afterwards — measured 2026-08-09, STRATEGY §6.1 — so age cannot establish that no frame will follow. Settled can.

## Not blocking

- **`QueuedBehind` / 已并入上一条回复** — you are right on both counts, and the attachment ordering makes it self-contradicting. Fixing the copy is a string change; fixing what it records is not. Tell me which you want here and I will do it in this PR.
- **Cancellation on a synchronous bus** — noted, and the budget is now bounded by the same deadline work in §1. A bulk cancel is still linear.
- **`RouterOptions.WecomResolverSetBuilt`** — your read is right, including that it fires before `Register` and so cannot see an incomplete set being dropped. `ResolverSetFor` on `engine.Router` is the better shape, but it is a change to the shared engine, which is the thing this PR exists to not do. Say the word and I will open it as its own PR and delete the field here.
- **`language_lint_test.go` scanning raw bytes** — real, and the failure mode is absurd: a comment mentioning `localeForSender(` breaks the build. Moving it to the AST walk at `:112`.

## Tests

`stream_bubble_db_test.go` asserting the stranded spinner as correct, and `stream_close_retry_test.go`'s "...GoesOutAsAMessageOnce" pinning §1 — both were pinning defects, and both are rewritten rather than adjusted. You were right that the second one contradicted its own doc comment.

Rebased on `a843b44a4`. `go vet` clean untagged **and** under `agentintegration` — that tag is how the last CI failure got through my local checks, since `trace_test.go` compiles only there and `main` asserts the writer lock with `sync.Mutex.TryLock` while this branch made it a channel semaphore. `tryLockWriter` answers the same question and the probe asks it.

Bohan-J (changes requested):
Still request changes — but the shape of this round is different from the last one, and worth saying up front.

**Five fixes, and the design judgment in each is right.** Your `provablyNotSent` catch is the best thing in the diff: it did fall to `default: return true` for `errStreamAckTimeout`, and that — not the branch I pointed at — is what let the second copy out. Choosing "unconfirmed is re-askable, a duplicate is permanent" is the correct way round for a chat with no unsend. Deriving `settled` rather than `sealed` for pruning, and catching your own first attempt with `TestPruneKeepsLiveTurnsEvenOverTheCap`, is the same quality of reasoning.

**What is blocking is one pattern, repeated.** Each fix landed at the call site I named and not at the other sites that share the rule. The answer path classifies unknown outcomes; the failure and cancel paths still do not. The owed-verdict gate closes the attribution hole; the rewrite exemption reopens it. The origin gate is correct and unreachable for the population it was written for. The relay frame carries the task id; nothing on the receiving end reads it. So this is not five new objections — it is the same five, each fixed in one of its two or three places.

Everything below was verified against head `2dc09ade1` with probes we wrote and deleted; where we reproduced something, the observed output is quoted.

---

## 1. The rewrite exemption reintroduces the silent loss, and it is now the worst path

This one is new, and it is a consequence of the fix rather than a survival of the old bug — so it needs the most attention.

`cancelAck` deliberately leaves a permanent debt (`acked < sent`) so the owed gate blocks later frames. `rewrite=true` removes the **gate**, but nothing removes the **debt**. So a seal retry on the same socket reaches `beginStreamFrameLocked`, which reads `clean := st.acked == st.sent` → false and stamps the waiter `addressable = false` (`ws_sender.go:499,506`). `deliverAck` only matches `w.addressable && w.seq == st.acked` (`:357`). The retry frame is written to the wire and simultaneously marked un-matchable: **the exemption lets the frame out but not the verdict in.**

```
after the lost ack: sent=2 acked=1
retry waiter: addressable=false seq=3 ; stream sent=3 acked=1
the rewrite with a SUCCESSFUL ack returned: wecom: stream frame ack timed out
```

Control with no outstanding debt returns `nil`, so it is specifically the post-timeout state. We also isolated `addressable` from `seq`: even when the late verdict realigns `acked` exactly onto the retry's `seq` (3 == 3), it still times out — so `addressable` is the load-bearing refusal, not the comparison.

**The serious case is a refused retry.** A stated refusal is the only signal that routes the answer to the plain-message fallback. Swallowed by the addressable gate, `seal` sees `errStreamAckTimeout` instead, retries blind, and returns the timeout — which the new classification correctly records as *unconfirmed, do not resend*:

```
closing frames: 4 (want 2)   plain messages: 0 (want 1)
stream_fell_back=1 outbound_unconfirmed=1 outbound_delivered=0 outbound_dropped=0
```

The server refused every closing frame, nothing reached the bubble, and the user got nothing. That is exactly the failure the owed gate was built to prevent, arriving through the exemption. The accepted-retry case is milder but writes 4 closing frames instead of 2, charging three redundant frames against the per-conversation budget.

Why no test caught it: `TestAClosingFrameWhoseAckWasLostIsWrittenAgain` forces a reconnect inside `onClosing`, so the retry lands on a fresh `wsSender` with clean bookkeeping. The other same-socket retry test sets `loseClosingAcks = 1 << 20`, so no verdict ever returns. Same-socket retry *with* a real verdict is the uncovered cell, and it is the broken one — and `sendersRegistry.streamRewrite` resolves by installation id, so without a reconnect production always takes that path.

The measurement that justifies the exemption (identical rewrites all return errcode 0) is not in question. What is missing is that a rewrite must also be *addressable* — the debt that blocks new frames should not blind a frame the gate has already decided to allow.

## 2. Only the answer path learned the classification

`outbound.go:397-457` does the distinction correctly now. `writeClosing` (`typing_indicator.go:921-935`) does not:

```go
err := m.streams.seal(ctx, m.senders, h, text)
if err == nil { return }
m.log.WarnContext(ctx, "...", "unusable", streamUnusable(err), "error", err)
if err := m.senders.sendTextCtx(ctx, h.InstallationID, h.ChatID, h.ChatType, text); err != nil {
```

It **computes `streamUnusable(err)` for the log line and then does not branch on it**, and the fallback reuses seal's own (possibly spent) context rather than taking a fresh one the way the answer path now does. So on `errStreamAckTimeout` / `errWriteAttempted` a failure or cancellation shows the closing frame *and* a plain copy; with the budget gone, neither. Every closer should share one outcome classifier and one fallback-budget rule.

## 3. `abortStreamFrameLocked` — the fix is not on head

You wrote that `ws_sender.go:668-671` takes the one-liner. It is not in `2dc09ade1`: the sole call site (`ws_sender.go:781`) still aborts on any write error, there is no guard inside `abortStreamFrameLocked` (`:516-522`), and `writeLocked:818` does wrap attempted failures in `errWriteAttempted`. Given you rebased onto `a843b44a4` and `main` landed its own `ws_sender.go` changes in the same window, this looks like a rebase casualty rather than anything you decided — flagging it so it does not get lost twice. Worth noting the function's doc now says it gives back the place "for a frame that never reached the socket", which is precisely the case the call site fails to distinguish.

## 4. The origin gate cannot reach the runs it was written for

`handleTaskFailed` calls `addressForTask` **before** the gate, and returns unconditionally at `!ours` (`typing_indicator.go:495-510`). A real first-party task has no `channel_task_delivery` row — `EnqueueChatTask`'s own doc says so ("no external delivery snapshot is created: first-party sends reply only to first-party clients"), and `SendDirectChatMessage` inserts none — so `GetChannelTaskDelivery` returns `pgx.ErrNoRows`, `taskAddress` maps that to `not ours`, and the handler returns before `originNotOurs → releaseRefusedRound` can run. In our instrumented run the gate was never consulted at all: `channel_ingested` was asked for `[]` and no refusal was logged. `releaseRefusedRound` is dead code for exactly the population it targets.

The new tests pass because `fakeOutboundQueries.GetChannelTaskDelivery` (`outbound_test.go:95-106`) ignores its task-id argument and returns a WeCom row for every id — the inverse of the production contract:

```
fake GetChannelTaskDelivery(<never filed id>) = channel_type="wecom" err=<nil>
addressForTask(<never filed id>) = ours=true
```

With a per-task double shaped like production:

```
after the browser run's task:failed: round still bound to web-1 = true
room's answer sealed its bubble = false; plain-message fallbacks = 1
```

The success path needs no custom double — it fails on your own rig, because `outbound.go:308-314` returns at the gate without releasing:

```
after the browser run's chat:done: round still bound to web-1 = true
```

`m.streams.forget` does not cover this; it drops the pending entry and retires the id but never unbinds a bound round.

## 5. The retry collision still swaps two people's answers

`retryUnbind` clears `taskID` and leaves `everBound = true`; `unboundLocked` (`stream_store.go:410-417`) still has no `everBound` check, so a new question's run takes the released round while `collectingLocked` — which does check it — makes that question open its own bubble. Two rounds, cross-wired. Driven in the real backoff order (release → new question opens *and is queued* → clone queued):

```
frame 2: bubble=s7781... finish=true content="the retry's answer"     # the 2nd question's bubble
frame 3: bubble=s9c06... finish=true content="the second answer"      # the 1st question's bubble
```

`TestARetryCloneQueuedAfterTheFailureAlsoFindsItsRound` calls `rig.ask` for the new question — a bubble with no run queued. Run as-is it passes; change that one call to queue the run and it flips. The ordering is real: a deferred child is published by the promoter when `fire_at` arrives, while a new question's `task:queued` goes out at enqueue, and nothing serializes the two. Note `releaseRefusedRound` calls the same `retryUnbind`, so fixing §4 inherits this.

## 6. The relay carries the id; nothing seals

`deliverRelayed` (`relay_outbound.go:1232-1323`) touches `o.senders` and `o.objects` only — `f.TaskID` appears on the receiving side solely in log attributes, and `o.streams` is never referenced. Driving replica B's routed frame into replica A's real `deliverRelayed`:

```
A pushed plain messages: ["⚠️ 这次没跑通，请稍后再试一次。"]
A stream frames after the relayed notice: [finish:false <think></think>]
A round store: has(task-1)=true depth=1
```

Same for a relayed answer. The comment at `typing_indicator.go:786-788` says the frame carries the task id "so it seals the right bubble rather than pushing a second message under one" — while `outbound.go:36-44` still states the gap verbatim ("deliverRelayed pushes the words as an ordinary message without ever looking for it"). The two comments now contradict each other; the second one is the true one. `TestAFailureOnAReplicaWithNoSocketIsRoutedToTheOneThatHasIt` asserts only that a frame was routed with the right kind, content and task id — its relay is the recording seam with nothing behind it, so sealing is untested.

Cancellation routes nothing off-lease either, though not for the reason it might look: `holding()` is a global check and can be true on a replica holding other rounds — the handler returns at `!t.HasBubble`. Since the code deliberately declines to chase an address for cancellations, the fix has to be a seal-by-task relay rather than an address lookup.

## 7. Bookkeeping: the remaining half, and the opening frame

`settled` is the right condition and the sweep does retire settled entries. What is left is the entry whose verdict never arrives: `cancelAck` leaves the count short by design, so it is never settled and never retired.

```
REQ-LOST: sent=1 acked=0
entries before sweep=2049 after=2; REQ-LOST still present=true
entries after 500 sweeps at the cap: 2548 (cap 2048)
```

To be fair about the size of this: the map is per-connection, and a merely-late ack still settles the entry, so this is slow growth on a long-lived connection for turns whose verdict truly never comes — not a fast leak. The reason to fix it anyway is that after ten minutes the round store has evicted the handle and `seal` refuses an expired one, so nothing can write for that req_id again and the retained counters no longer protect attribution.

Separately, an opening frame failing with `errWriteAttempted` falls to the `default` branch and drops the handle (`typing_indicator.go:301-313`). Everywhere else the package treats that error as possibly-sent — `provablyNotSent` returns false for it, `unconfirmedReason` files it as `write_attempted` — so this is the one site treating it as a definite refusal. Result: the placeholder may be on screen with its handle discarded, the answer arrives as a plain message, and nothing can ever seal it.

---

## Your two questions

**`QueuedBehind` / 已并入上一条回复** — fix the copy, not what it records, and do it here. Drop `QueuedBehind` from that switch entirely and let it fall to `StreamNoReply` / `StreamNoReplyWithFiles`. That removes the false claim without adding state, and fixes the attachment ordering for free. A real "merged" notice needs a real merge signal, which is not this PR's job.

**`ResolverSetFor`** — your read is right and so is your reluctance: it is a shared-engine change and not-touching-the-engine is why this PR is acceptable. Keep the field here with a comment recording that it asserts *built* rather than *registered*, and open the accessor separately whenever suits you. **This PR is not blocked on it.**

## Verification scope

`go build ./...`, `go vet` (including `-tags agentintegration`) and the package's in-memory tests pass under `-race`; your ten new tests pass. Findings 2 and 3 were read directly from source; 1, 4, 5, 6 and 7 were reproduced with temporary probes, all since deleted. We could **not** run the DB-backed tests — no migrated database on our side — so nothing here rests on them.

One thing outside the code: the branch now conflicts with `main` in `server/cmd/server/router.go`, `relay_outbound.go`, `ws_frame.go` and `ws_sender.go`, from `main`'s own WeCom push rate-limit / ack-outcome work. That overlaps §1 and §3 directly, so it is worth rebasing before fixing rather than after.

seacen:
Reviewed your round against head `525eb44d7`, rebased onto `8c4f4328f`. All seven are fixed, and two of them not the way you suggested — those two are marked and argued below rather than buried.

**You diagnosed the pattern, so that is what I changed first.** "Each fix landed at the call site I named and not at the other sites that share the rule" is exactly right, and it was a method problem, not seven separate lapses. Every finding below now starts from an invariant in one sentence and a **mechanical enumeration** of the sites it applies to — a query I can re-run and you can re-run. Where the enumeration turned up a site you had not named, I say so.

---

## 1. The rewrite exemption

**Invariant: a frame the gate lets onto the wire must be able to read its own verdict.**

You are right, and the reasoning holds all the way down: the measurement that justifies letting an identical rewrite out is the same measurement that makes its verdict safe to read. Six identical rewrites of a sealed stream all returned errcode 0, so whichever identical write a verdict belongs to it reports the same outcome. Letting the frame out while refusing to match its answer is strictly worse than refusing it outright — `seal` reads `errStreamAckTimeout` in place of the refusal that actually came back, and a stated refusal is the only thing that routes the answer to a plain message.

The exemption is now one decision instead of two: `awaitAck` stamps `rewrite` on the waiter it creates, and `beginStreamFrameLocked` reads `clean || w.rewrite`. Position still decides the match; what changes is that the debt no longer blinds a frame the gate already allowed.

`TestARewriteReadsItsOwnVerdictDespiteTheDebtThatLetItOut` drives your case exactly — same socket, real debt, and the server answering the retry with a stated refusal. Removing `|| w.rewrite`:

```
the rewrite was written unaddressable: the gate let the frame out and then blinded it to
the answer, so a stated refusal is read as a timeout and the answer is recorded
unconfirmed while nothing reached the bubble
```

## 2. Only the answer path learned the classification

**Invariant: every closer reads a seal's outcome the same way and spends the same fallback budget.**

```
grep -n "streams.seal(\|finishStream(" internal/integrations/wecom/*.go | grep -v _test
```

Three sites: the answer (`outbound.go`), the failure and cancellation notices (`typing_indicator.go`), and — after finding 6 — a delivery routed from another replica (`relay_outbound.go`). Three is the threshold this repo uses for extracting, so there is now one `classifySeal` and one `fallbackBudget` in `seal_outcome.go`, and all three call them. A fourth closer calls those two rather than writing a third reading.

`writeClosing` was doing the thing you described — computing `streamUnusable` for the log line and then not branching on it — and reusing seal's spent context for the fallback. Both halves now have a test: `TestAClosingNoticeWhoseOutcomeIsUnknownIsNotSaidAgain` and `TestAClosingNoticeThatWasRefusedIsSaidAsAMessage`.

## 3. `abortStreamFrameLocked` — you were more generous than the facts

It was not a rebase casualty. `git log -p a843b44a4..2dc09ade1 -- ws_sender.go | grep errWriteAttempted` returns one line, the wrapper in `writeLocked`. **The guard was never committed. I wrote the sentence and not the code, and the sentence was in a reply to you.** That is the worst thing in this round and it is a process failure rather than a judgment one — every `file:line` claim in a reply is now checked against HEAD before the reply goes out, by a script rather than by me remembering.

The invariant it belongs to is shared with 7b, so both are enumerated together.

**Invariant: a failure that may have reached the socket is never read as proof it did not.**

```
grep -rn "errWriteAttempted" internal/integrations/wecom/*.go | grep -v _test
```

Six decision points. Four already read it correctly — `outbound_outcome.go:120`, `outbound_media.go:490`, `relay_outbound.go:1404`, `rate_limit.go:224`. Two did not: the abort at the one call site in `writeStreamFrame`, and the opening frame's `default` branch. Both fixed.

`TestAFailedWriteGivesBackItsPlaceInTheOrder` kept its assertion and changed its fixture — it used a `WriteMessage` failure while its own doc said "a write that never reached the socket", so it was naming the wrong door. It now fails `SetWriteDeadline`, which is the whole of that population, and `TestAWriteTheSocketMayHaveTakenKeepsItsPlaceInTheOrder` covers the other door.

## 4. The origin gate — and the double that made it untestable

You found the dead gate. The thing underneath it is worse and worth stating plainly: **`fakeOutboundQueries.GetChannelTaskDelivery` ignored its task-id argument and returned a WeCom row for every id**, which put a run in a state production cannot be in — asked in the web UI *and* delivered over WeCom at once. Ten tests of that gate ran inside a world that does not exist.

I verified the contract rather than assuming it: `enqueueChatTask` takes `requireDelivery`, `EnqueueChannelChatTask` passes true and `EnqueueChatTask` passes false, with the reason in its own doc — "no external delivery snapshot is created: first-party sends reply only to first-party clients". So no row ⟺ first-party, and a WeCom row is itself proof of origin.

**Where I did not take your fix.** Moving the gate ahead of `addressForTask` makes it reachable and makes *every* foreign failure in the deployment pay the gate's two reads — `TestAnotherChannelsFailureNeverReachesTheTaskRow` goes red, and it is the same cost objection multica-eve is currently blocking #8348 on. So one delivery-row read now has three answers instead of two:

- **`routingSilent`** — another platform's row, or ours on a revoked installation. One lookup, which is what a foreign failure cost before the gate existed.
- **`routingOurs`** — a live WeCom address. The row is the proof of origin, so this skips the gate's two reads entirely. Cheaper than head.
- **`routingNoRow`** — nothing filed. The first-party population the gate exists for, and the only one that pays.

That test also had the fixture problem in reverse: it modelled Slack's run as a *missing* row, which is the first-party shape. It now produces a row naming `slack`, and its assertion is unchanged.

The answer path's own gate returned without releasing too, exactly as you said — `TestAWebRunsCompletionAlsoGivesTheRoomsBubbleBack`.

## 5. The retry collision — your one-liner alone strands the bubble

**This is the one where I think the suggested fix is not enough, and I want to be explicit about why.**

Adding `everBound` to `unboundLocked` stops a new question's run from taking the released round. It also stops the clone from taking it: the clone's `task:queued` finds no unbound round, goes to `pending`, and the released round sits with a bubble nobody will ever close. Trading a cross-wire for a stranded spinner is a better trade, but it is not the fix.

I tried an ordering rule and convinced myself it cannot work. The two cases are symmetric — the clone may arrive before or after a new question's run, and `task:queued` carries nothing that separates them. Any fixed preference gets one of them backwards; I have both orderings as tests and they pin each other.

What resolves it is the only authoritative name either run has. A released round now records `retryOf` — the attempt it is waiting for — and stays out of the pool. The clone's **ending** names it through `chat_input_task_id`, which `roundTaker.take` already reads, and which this store's own comment already says is the round's own task id. The lineage read runs first only while a round in that session is awaiting a retry, which is an in-memory check, so an ordinary ending costs what it always did.

`TestARetryCloneQueuedAfterTheFailureAlsoFindsItsRound` keeps its fixture. `TestANewQuestionsRunDoesNotTakeTheRoundHeldForARetry` is your ordering, with the new question's run queued.

One consequence worth flagging since you will see it in the diff: `retryUnbind` and `releaseRound` are now two functions. They were one, and the two callers wanted opposite things — a retry has a replacement coming and must hold its round; a refused origin has none and must give it back.

## 6. The relay carries the id; now something reads it

`deliverRelayed` takes the round by task id and seals it, falling through to the ordinary addressed path when the round is gone or the seal is refused — on a fresh budget, through the same `classifySeal` as the other two closers. Replies only; an inbox push is not an answer to a round and `TestARelayedInboxPushNeverSealsABubble` pins that.

The two comments that contradicted each other now agree, and the one in `outbound.go` that stated the gap is gone rather than softened.

## 7. Bookkeeping, and the opening frame

**The second one I did differently.** Age alone cannot retire an entry — that is your own argument from last round, and it still holds: a req_id outlives its stream, and a new stream id on the same req_id is accepted. What retires it is **our** reach ending, not the platform's: the round store evicts a handle at `streamMaxAge`, a closer already holding one is bounded by `streamCloseTimeout`, and past both nothing can hand `seal` a handle for that req_id. So the condition is settled **or** older than `streamMaxAge + streamCloseTimeout`, and the comment records that argument rather than the shorter one.

`TestPruneKeepsLiveTurnsEvenOverTheCap` changed its fixture from `-2 * streamMaxAge` to a young turn. A twenty-minute-old unsettled entry is not a live turn — nothing can write for it — so the old fixture described something production does not produce. The property it protects is unchanged and `TestPruneRetiresATurnNothingCanWriteForAgain` covers the other side.

The opening frame is enumerated under finding 3 above.

---

## Your two

**`QueuedBehind`** — done, your way. Dropped from the switch; it falls to `StreamNoReply` / `StreamNoReplyWithFiles`, which removes the claim without adding state and puts the attachment case back in charge. The field stays, since what it records is true. The copy string had no caller left and is gone.

**`ResolverSetFor`** — kept here with a comment saying it asserts *built* rather than *registered*, and why the accessor belongs in its own PR. Thank you for reading the reluctance as a position rather than as foot-dragging.

---

## Existing tests changed — the whole list

None weakened; each is a fixture that described a state production cannot produce, and every property is still asserted somewhere.

| test | change |
|---|---|
| `TestAFailedWriteGivesBackItsPlaceInTheOrder` | fixture now fails `SetWriteDeadline`; the other door gets its own test |
| `TestPruneKeepsLiveTurnsEvenOverTheCap` | fixture is a young unsettled turn; the aged case gets its own test |
| `TestAnotherChannelsFailureNeverReachesTheTaskRow` | Slack's run has a row naming Slack instead of no row |
| five origin-gate tests | now model the row-less population the gate serves |
| `TestTheOriginOfARetryCloneIsItsParentsBatch` | asserts the lineage only; the notice it also asserted is covered at two other sites |
| `TestAQueuedRoundWithNothingToSaySaysItWasMerged` | renamed and retargeted with the copy |

`fakeOutboundQueries` now answers `GetChannelTaskDelivery` and `GetChannelUserBindingByUserID` by the id it is given, and **ends the test** when a rig asks for a WeCom delivery row on a run it also called first-party.

## What runs before a reply now

The gate that let finding 3 through is closed by a script, not by intention: every `file:line` in a draft is resolved against the tree and printed, and an ambiguous basename is an error. Two of the four test disciplines are mechanical too — a scan that fails the branch for any **new** double taking a key it does not read, and a self-check that fails when the scan matches nothing at all, because a scanner that finds nothing is not evidence of a clean tree.

`go build`, `go vet` untagged and under `agentintegration`, `-race -p 1` on the touched packages differenced against the merge base, the DB-backed tests proven to have RUN rather than skipped, and the merge with `main` compiled. Green on `525eb44d7`.

No DB-backed test result here rests on your side of the fence — you flagged that you could not run them, and three of them ran here.

Bohan-J (changes requested):
Everything from the last round landed, and "read one rule at every site that shares it" is the right name for it. Verified directly rather than taken on trust:

- `w.addressable = clean || w.rewrite` — the exemption now lets the verdict back in. This was the one that could lose an answer outright, and it is closed.
- `if !errors.Is(err, errWriteAttempted) { s.abortStreamFrameLocked(reqID) }` — landed this time.
- `unboundLocked` gained `r.retryOf == ""`, so a released round is no longer up for grabs by the next question.
- `fakeOutboundQueries.GetChannelTaskDelivery` answers by the id it is given and returns `pgx.ErrNoRows` for an unfiled task — the double now matches the production contract, which is what makes the origin tests mean something.
- `handleTaskFailed` reordered so the gate is reached for runs that have no delivery row at all.

**One thing left, and it is the same thing three times.** The relay can carry a reply that has words. It has no way to say "close this round" without words — so every ending that has no words to carry does not reach the holder at all.

## The three

1. **Cancellation never routes.** `handleTaskCancelled` has no relay call on any path; off-lease it returns at `!m.streams.holding()`, or at `!t.HasBubble` when this replica holds other installations' rounds. Both were driven:
   ```
   [cancel off-lease]                    relay frames=0  holder closing frames=[]  holder open bubbles=1
   [cancel off-lease, holding others]    relay frames=0  holder closing frames=[]  holder open bubbles=1
   [control: failure off-lease]          relay frames=1  holder closing frames=[⚠️ 这次没跑通，请稍后再试一次。]
   [single replica]                      closing frames=[⏹️ 这次处理已取消。]  open bubbles=0
   ```

2. **A completion with no words.** `outbound.go:429-436` returns `errNothingToSay` before `sendAsMessage`, which is the only relay site on that path. This is wider than the line suggests: `carriesFiles` is `mayCarryAttachments`, a *may* rather than proof of a file, so both sides fail — without object storage nothing is published, and with it the same empty completion is published as `Content: ""` and then dies at (3).
   ```
   [no storage]    relay frames=0                                    holder open bubbles=1
   [with storage]  relay frame 0: kind=reply content="" files=true   holder closing frames=[]  open bubbles=1
   [single replica] closing frames=[（这轮没有需要回复的内容）]        open bubbles=0
   ```

3. **Attachment-only relay does not seal.** `relay_outbound.go:1267` gates the `take` *and* the seal on `hasVisibleChar(f.Content)`, so the frame delivers the file and never consumes the round:
   ```
   [attachment-only]  holder chat messages=[media:file]  closing frames=[]  open bubbles=1
   [single replica]   closing frames=[（这轮没有文字回复，附件在下面）]  open bubbles=0
   ```

## Why this is still blocking rather than a follow-up

Not severity in the abstract — none of these loses or doubles an answer. It is that they are the last places where this branch is **worse than `main`**, and "the worst case is today's experience" is the floor this PR stands on to ship default-on with no flag.

In all three, `main` was silent: a cancellation was a DEBUG line, an empty completion was skipped, an attachment arrived on its own. Off-lease, this branch replaces silence with a spinner that claims work is in progress for the rest of the window and then leaves a dead `<think></think>` bubble. A stale spinner is not a smaller version of silence; it states something false.

And nothing ends it. We looked specifically for a mitigation and there is none: `sweepLocked` only deletes map entries and never writes a frame, and it is not periodic (it runs from `open` / `bindNext` / `takeOldestUnbound` / `take`); `pruneStreamsLocked` retires ack bookkeeping only; `OnSettled` uses `takeOldestUnbound`, which requires an unbound round, so it cannot reach a stranded one; and the next question opens its own bubble rather than joining. Confirmed by probe — after advancing past `streamMaxAge` and forcing a sweep, zero closing frames. The user watches it for the full window.

Frequency is not edge, either: with N replicas a cancellation is off-lease about (N−1)/N of the time.

## One design note, so the obvious fix does not become a worse bug

Do **not** relay the cancellation as an ordinary `relayKindReply`. On a holder that has no matching round it falls through to `relay_outbound.go:1290` and pushes 这次处理已取消 into the room as a plain message — one cancel-all click putting a message in every chat, which is exactly what the comment at `typing_indicator.go:713-719` refuses to do. What is missing is a **seal-only frame kind**: carries a task id and the ending copy, seals the round if it is there, and does nothing at all if it is not. The same kind covers (2) and (3) — let the holder produce `StreamNoReply` / `StreamNoReplyWithFiles` in the round's own locale rather than shipping the words across.

## Two small things, neither blocking

The PR body's "Best effort, with the floor named" bullet is now stale in your favour — it still says `deliverRelayed` pushes without looking for the bubble, which `608a3635b` fixed for the text case. Worth rewriting to name what is actually left. Cancellation off-lease is not listed anywhere.

## Verification

The five fixes above were read directly at `525eb44d7`. The three gaps were reproduced with temporary two-replica probes in production shape — separate buses, one replica holding the socket and round, relay wired to both the outbound subscriber and the indicator as `cmd/server/router.go` does — then deleted; the package is green under `-race` with them in place. The controls (failure off-lease, relayed text, and each case on a single replica) all seal correctly, which is what isolates these three. DB-backed tests still could not run here.

Fix these and it goes in — there is nothing else outstanding from our side.

seacen:
All three fixed at `a67a9cf23`, as the seal-only frame kind. You were right that it is one thing three times: once the kind existed, the three sites became three callers of it.

### The frame kind

`relayKindSeal` carries a task id, a session id and a `SealReason`. It carries no words.

`deliverRelayed` routes it at the top, **before** the installation parse (`relay_outbound.go:1306`), so dispatch is by round ownership rather than by socket. `sealRelayedRound` takes the round by task id and returns having said nothing when there is none — the property that keeps one cancel-all click from putting a message in every chat, and it has its own test rather than being left as a claim.

The words are the holder's, in the round's own locale:

| reason | no files | with files |
|---|---|---|
| `cancelled` | `StreamCancelled` | — |
| `no_reply` | `StreamNoReply` | `StreamNoReplyWithFiles` |

`wordlessSealCopy` is the only place that picks between the last two, so a relayed round and a local one cannot say different sentences about the same outcome.

### The three

1. **Cancellation.** Both exits publish a seal frame before returning: `!m.streams.holding()` at `typing_indicator.go:752`, and `!t.HasBubble` at `:771`.
2. **A completion with no words.** `outbound.go:449` relays the seal before `errNothingToSay`. The `carriesFiles` half still goes out through `sendAsMessage`, and now seals too — see (3).
3. **Attachment-only relay.** The take is no longer gated on words: `relay_outbound.go:1350` takes on `Kind == relayKindReply && TaskID != ""`, and a take that comes back with no visible character is worded from the table above. The file arrives and the round is consumed.

### Tests

Five, in `relay_seal_test.go`, in the two-replica shape you used — separate buses, one replica holding the socket and round, relay wired to both the outbound subscriber and the indicator:

- cancellation off-lease, and off-lease while holding another installation's round
- a completion with nothing to say
- an attachment-only answer
- a seal frame arriving at a replica that holds no matching round: zero frames, zero messages

Each was reverse-verified. The fourth passed for the wrong reason first — the frame had no installation id, so `deliverRelayed` returned at the parse before reaching any push — and now carries a full address, so the silence is the seal path's rather than the parser's.

### PR body

Rewritten. The stale bullet is gone; what is left of best-effort is the frame not arriving (the holding replica restarted, or the relay dropped it) and the server refusing the closing frame with 846605 / 846608. Cancellation off-lease is not on the list because it is no longer a case.

Bohan-J (approved):
Approving, and merging — thank you.

Six weeks, four rounds, and every round came back inside a day with the finding addressed at its root rather than at the line we pointed to. The reasoning written into these files is unusually careful: several of the comments diagnose the real hazards more precisely than our reviews did, and "read one rule at every site that shares it" was the right way to take the second round. The `provablyNotSent` catch, the settled-not-sealed pruning condition, and the seal-only frame kind are all yours, and they are the parts that make this safe to ship.

## Two follow-ups we would welcome

Neither blocks the merge — both are narrow, and neither is worse than `main` for whether the answer arrives — but each is a place where the code does not yet do what its comment says.

**1. Seal frames do not get past `RelayOutbound.perform`.** `perform` checks `ownsSocket(item.frame.InstallationID)` before anything else, and a `relayKindSeal` frame carries no installation id by design, so `ownsSocket("")` is false on every replica and the frame is finished without reaching the seal branch in `deliverRelayed`. On a multi-replica deployment an off-lease cancellation or wordless completion therefore still leaves a spinner. Ownership needs to be decided by frame kind — a reply by holding the socket, a seal by holding the round — rather than by skipping the check, or a replica without the round can win the global claim and swallow the frame. The seal tests drive `deliverRelayed` directly or publish to `recordingRelay`, so a two-replica test through the real worker (`DeliverWecomOutbound → perform`) is what would have caught it.

**2. The close retry never starts in production.** The budget check in `seal` requires `closeRetryDelay + ackTimeout` (7s) to remain, and a lost ack has already spent 5s of `handleEvent`'s 10s. `TestAClosingFrameWhoseAckWasLostIsWrittenAgain` passes because `rig.answer` uses `context.Background()`. Run the same scenario through `answerWithContext` with a 10s deadline and production timings:

```
[10s deadline]  elapsed 5s | closing frames: old socket=1 new socket=0 | plain=0 | delivered=0 unconfirmed=1
[no deadline]   elapsed 7s | closing frames: old socket=1 new socket=1 | plain=0 | delivered=1
```

It is not a regression — `main` does not resend an unacked message either — but it is the improvement the retry exists for, and it is not reachable yet. The constants and the budget need to fit, and that test should run under the production deadline so it cannot pass in a world production never enters. (`seal_outcome.go:68` cites `TestTheCloseRetryPolicyFitsTheBudgetItRunsUnder`, which does not exist; it is probably the test this wants.)

## Tests

We are going to follow this with a PR that consolidates the test suite: some scenarios are covered three or four times across files, a few assertions pin log text or query counts, and `language_lint_test.go` is the only source-scanning test in `server/`. Nothing guarding the answer arriving exactly once, or a web-UI run's failure staying out of the room, will be removed. We will link it here.

Thanks again.

Bohan-J:
The test consolidation mentioned above is #8625: 35 tests removed and 1 added, plus the two test-only production hooks and the dead `failureBelongsOnWecom` / `QueuedBehind`. No production behaviour changes, and everything guarding exactly-once delivery and the origin gate is kept. If you are starting the two follow-ups, basing them on that branch will save you a conflict in the test files.

PR #8358, opened 2026-09-13
fix(server): heal drifted issue counters

vincent-wuhan:
## Summary
- Make workspace issue-number allocation self-healing when `issue_counter` drifts behind `MAX(issue.number)`.
- Reserve `GREATEST(issue_counter, MAX(issue.number)) + 1` while retaining the workspace-row serialization boundary.
- Add a regression test that seeds counter drift and verifies the allocated/persisted next number.

## Validation
- `make sqlc`
- `go test ./internal/service -run 'Test(AllocateIssueNumberHealsCounterDrift|IssueCountLimitSerializesConcurrentCreatesAndDeleteFreesCapacity)' -count=1 -v` against a temporary migrated PostgreSQL 16 database
- `DATABASE_URL=... go test ./internal/service -run 'Autopilot' -count=1`
- `DATABASE_URL=... go test ./internal/service -count=1`

## Incident Evidence
- OpsKeeper daily-report Autopilot manually failed three times with `duplicate key value violates unique constraint "uq_issue_workspace_number"`.
- This can occur when historical issue imports/backfills leave the workspace counter below the highest persisted issue number; the old query returned an already-used number.


Bohan-J (changes requested):
Thanks for chasing this down — the failure mode you hit is real and worth fixing. A workspace whose `issue_counter` has fallen behind `MAX(issue.number)` is permanently wedged rather than transiently broken: `AllocateIssueNumber` bumps the counter inside the same transaction as the INSERT, so the unique violation rolls the increment back too, and the next create retries the same taken number. Every surface — UI, CLI, autopilot — then fails forever with an opaque `duplicate key` error. That deserves a fix.

I'd like changes to how it's fixed, though.

## 1. The normal allocation path should stay at zero added database work

`IncrementIssueCounter` runs on every issue create in every workspace. This change makes all of them read `MAX(issue.number)` to repair a state that, as far as we can tell, only arises when something writes to the table outside the application. We don't want every create to carry a repair check for a condition we have no evidence occurs in normal operation.

To be fair to the change: the added query is genuinely cheap — an index-only backward scan on the `uq_issue_workspace_number` index, folded into the existing UPDATE rather than adding a round trip. The objection isn't its magnitude. It's that a repair should run when something is actually broken, and "the queries already in this transaction are more expensive" isn't a reason to add another one.

## 2. The stated root cause doesn't exist in this codebase

The description attributes the drift to "historical issue imports/backfills". I can't find such a path. Every production writer of `issue.number` allocates through `AllocateIssueNumber` in the same transaction as the insert:

- `server/internal/service/issue.go:317` — UI / CLI / API / quick create
- `server/internal/service/autopilot.go:692` — autopilot dispatch
- `server/internal/handler/onboarding_shim.go:251` and `:419` — onboarding

There is no bulk import, no cross-workspace issue move, no workspace copy, and `UpdateWorkspace` never writes `issue_counter`. That has held for the entire history of the feature — all six commits that touched `IncrementIssueCounter` keep allocation inside the create transaction, including when the Lark work (`8c98940b7`) moved it from the handler into the service.

So before we change allocation: could you share how that workspace actually drifted? Three consecutive `duplicate key` failures do mean the counter is behind, but if what put it there was an operational script or manual database surgery, that's where the fix belongs; and if it was a code path, we need to find it, because self-healing would hide it from us permanently.

Worth running against the affected database to size the problem:

```sql
SELECT w.id, w.issue_counter, MAX(i.number) AS max_number
FROM workspace w JOIN issue i ON i.workspace_id = w.id
GROUP BY w.id, w.issue_counter
HAVING MAX(i.number) > w.issue_counter;
```

## 3. The regression test doesn't reproduce the scenario it describes

`seedAttributionFixture` inserts its issue without a `number`, so the row takes the column default of `0` (`server/migrations/020_issue_number.up.sql:8`). In the new test `maxNumber` is therefore `0`, and the drift it seeds is `issue_counter = -1`. What it verifies is `GREATEST(-1, 0) + 1` — a negative counter against an empty number space, not a counter trailing a positive `MAX(number)`. It would pass just as well against `GREATEST(issue_counter, 0) + 1`.

If a healing test survives the redesign, please seed properly numbered issues (say 1..5), set the counter to 2, and assert the allocation lands at 6 — ideally driving a real create, so that what regresses pre-fix is the duplicate-key error itself.

## 4. Unrelated change bundled in

`updated_at = now()` on the workspace row turns `workspace.updated_at` from "settings last changed" into "an issue was last created", and that field is exposed in the API response (`server/internal/handler/workspace.go:137`). Nothing seems to depend on it today, so the risk is low — but please split it out or say why it's here.

## What would be acceptable

Either of these leaves the healthy path untouched:

- **Repair out of band.** Keep `issue_counter + 1` exactly as it is, and fix drifted workspaces with an operator-run statement scoped to the affected workspace, documented next to a clearer error message. Simplest option, and the right one if the cause turns out to be manual database work.
- **Heal only on conflict.** Leave allocation alone and react to SQLSTATE `23505` on `uq_issue_workspace_number`. One caveat if you take this route: it needs to be a whole-transaction retry, not a savepoint around the INSERT — a savepoint must be established before the statement that might fail, so every successful create would pay for it too. Driving the retry off the returned SQLSTATE keeps successful creates byte-identical to today.

Either way, please add a log line or metric when a repair fires. If drift is ever healed silently, the numbering skips and the next occurrence leaves no trace for whoever has to debug it.

PR #5909, opened 2026-07-24
fix(daemon): drop the prior session when the agent has been renamed

kakiuwang-ui:
## Problem

`gateResumeToReusedWorkdir` lets a managed follow-up reuse the prior task's workdir and resume the provider session recorded there. Nothing in that path checks whether the *agent* on both sides is the same identity.

Renaming an agent rewrites its instructions (the generated `CLAUDE.md` / runtime config is regenerated before launch), but it does not invalidate the transcript. The next task therefore resumes a conversation whose history was established under the old name — and the provider weighs that history against the freshly written instructions. The observable result is an agent that keeps introducing itself, and behaving, as its previous identity for the rest of the conversation.

This is a daemon-side session-ownership bug: a session belongs to the agent identity that produced it, and a rename ends that ownership. It is independent of any particular report — the resume path simply has no identity check on it today.

## Change

`server/internal/daemon/daemon.go` gains `gateResumeToSameAgentName`, called from the task-launch path right beside the existing resume gates — and, importantly, **before** the runtime config is rewritten, while the previous identity is still on disk:

```go
if resumeReachable {
    gateCodexResumeToRolloutPresence(&task, &taskCtx, provider, env.CodexHome, taskLog)
    // Check for agent rename BEFORE overwriting CLAUDE.md — we need to
    // read the old identity while it is still on disk.
    gateResumeToSameAgentName(&task, &taskCtx, provider, env.WorkDir, taskLog)
```

It reads the agent name recorded in the reused workdir's runtime config (`execenv.ReadPriorAgentName`) and, when it differs from the current task's `AgentName`, drops `PriorSessionID` and clears `PriorSessionResumed`, logging the old and new names.

It is a no-op when there is nothing to resume, when the current agent name is empty, or when the prior name cannot be read — a fresh workdir or a config written before the identity section existed both fall into that last case, and neither is evidence of a rename.

Unlike `gateResumeToReusedWorkdir`, it deliberately does **not** set `PriorSessionResumeUnavailable`. That flag makes the agent disclose a lost conversation; a rename is an intentional act by the user, so starting fresh is the expected outcome rather than a loss worth announcing.

## Tests

`TestGateResumeToSameAgentName` covers five cases: same name keeps the session; renamed agent drops it (and asserts `PriorSessionResumeUnavailable` is *not* set); no prior session is a no-op; empty current agent name is a no-op; a missing config file keeps the session. Plus `execenv.ReadPriorAgentName` coverage in `runtime_config_test.go`.

[...]

PR #8360, opened 2026-09-13
MUL-7329 fix(daemon): tell a resumed agent when its identity changed, instead of dropping the session

kakiuwang-ui:
Replaces #5909, at @Bohan-J's suggestion when closing it:

> If a reproducible case does turn up, we'd want to reopen this — but probably along a cheaper line than dropping the session. There's already machinery for exactly this shape of problem: `backendResumeContinuityNotice` in `server/internal/daemon/prompt.go` appends a notice to the prompt on resume. Detecting an identity change and appending "your identity was updated since the previous turn; the current brief is authoritative" keeps the conversation and resolves the conflict in one sentence, and it covers instruction edits as well as renames.

## Problem

On a resumed turn the agent is handed two descriptions of itself and nothing that says which is current:

- the **replayed conversation**, in which it introduced itself and reasoned under the identity it had then, and
- the **runtime brief on disk**, which `InjectRuntimeConfig` has already rewritten with the current one.

The replayed turns are far longer than the brief and read as lived history, so the agent can keep acting as who it used to be. `shouldReusePriorWorkdir` gates reuse on `AgentID` and scope, and neither a rename nor an instructions edit changes the ID, so the follow-up reuses the directory and resumes the session recorded in it; nothing on the resume path compares identity.

## Why a notice and not a dropped session

#5909 dropped the session on a rename. That was declined for reasons I think are right, and this PR is built on them rather than around them:

- **A name check covers the rarer half.** Identity drift far more often looks like an operator rewriting instructions while the display name stays put — invisible to a name comparison.
- **The consistent version is worse than not fixing it.** Fingerprinting name + instructions and *dropping the session on any change* would discard conversation continuity on every minor instructions tweak, a routine action, to guard against a rare one.
- **Resume continuity has real value**, and the failure it was traded against has never been reproduced.

A notice keeps the conversation and resolves the conflict in one paragraph, and — the point of the redesign — it costs the same whether the change is a rename or an instructions edit, so it can cover both.

## Implementation

**Detection — `execenv`.** `AgentIdentityFingerprint(ctx)` digests the identity **fields** — `AgentID`, `AgentName`, `AgentInstructions`, length-framed — and `writeAgentIdentity` stamps it into the managed block as an inert HTML comment, for the same reason the block's own markers are one: inert in every Markdown renderer, harmless when the file is fed to the agent. `PriorAgentIdentityFingerprint(workDir, provider)` reads back what the previous run stamped.

I first wrote this as a comparison of the *rendered* Agent Identity section and it was wrong in a way worth naming, because it failed at exactly the case this PR exists for. Agent instructions are user-authored Markdown and routinely contain their own H2; the section slice ended there, so for any agent with structured instructions every later instructions edit compared **equal**. A name containing a newline did the same thing more sharply. Digesting the fields removes the question of where the section ends, and with it the coupling to the section's layout — a later edit to how the identity renders no longer makes every in-flight session announce a change.

**Timing — `daemon.go`.** `agentIdentityChangedSincePriorRun` runs after the existing resume gates and **before** `InjectRuntimeConfig`, which is what overwrites the fingerprint it reads. It changes nothing about the resume — no flags cleared, no session dropped — it only decides whether the turn carries the notice.

**Delivery — `prompt.go`.** A `WithAgentIdentityChanged()` `PromptOption`, matching `WithSharedLocalDirectory` / `WithWorktreeReplayConflicts`: run-scoped facts only the daemon's own execution context can answer. The notice renders into the per-turn user message, not the brief, for the same reason as the `SessionContinuityNotice*` family — true of one run and false of the next on the same issue, so putting it in the cached prefix would break prompt-cache stability across resumes (MUL-5377).

The notice is gated on a live `PriorSessionID`, not on the option alone. The option outlives the decision that set it: the cold-session retry clears `PriorSessionID` and rebuilds the prompt with the **same** options slice, so without the gate the retry — which resumes nothing — carried a notice about "the previous turn of this conversation" next to a continuity notice saying that conversation is gone.

## Notes for review

- **It is not mutually exclusive with `SessionContinuityNotice`.** My first version suppressed the identity notice whenever `PriorSessionResumeUnavailable` was set. That flag does not mean nothing was replayed: under MUL-5305 the server withholds a newer session and hands back an **older one that resumes cleanly**, flagging the gap — so turns written under the previous identity are in front of the agent and the flag is set. Suppressing there dropped the notice in exactly the case it exists for.
- **Three known coverage gaps, all the same shape** — the signal lives in the workdir's managed block, so where that block is absent the answer degrades to *silent* rather than to a false alarm. They are enumerated on `agentIdentityChangedSincePriorRun`: `local_directory`/local-worktree runs excise the block on the way out (that cleanup is worth more than this notice); pi-family resumes are gated on the session file alone and never on workdir equality, so they can resume into a workdir no prior run wrote to; and delivery is **at-most-once**, because a run that injects and then dies before the agent sees its turn consumes the change. Closing any of them needs the prior identity recorded somewhere this run's own injection does not own — a larger change, happy to do it separately if you want it.
- **Adjacent pre-existing defect, deliberately not fixed here.** `writeAgentIdentity` interpolates `ctx.AgentName` into the brief raw, while sibling name fields go through `sanitizeNameForBriefMarkdown` (`RequestingUserName`, the initiator block). An agent name is user-controlled, so it can close the bold span and emit headings or list items into a file the agent loads as trusted instructions. This PR no longer *depends* on that (the fingerprint is over fields, not rendered text), so fixing it is a separate concern with its own blast radius on existing briefs — flagging rather than bundling. Happy to open it if you want.
- **A runtime can still reject the resume mid-run**, after the daemon has already decided to send this; the backend then prepends a continuity notice to the same turn. The notice's wording is conditional ("where those earlier turns are still in front of you") so it stays true in that pairing instead of asserting a history that did not come back.
- **Nothing in `builtin_skills/` documents these notices**, so there is no `SKILL.md` to update.
- `backend-tests` is red on `main` itself right now — `468_comment_deleted_at` (#8323) and `468_drop_reference_only_column` (#8253) share a migration prefix, so `TestMigrationNumericPrefixesAreUnique` fails on `main` and on every PR branched from it, this one included. Nothing here touches migrations.
- Two daemon tests flaked once each in full local runs and passed in isolation and on rerun: `TestStartDshProfileProvision_RunsAtMostOnce` and `TestExecuteAndDrain_CodexInactivityReportsMCPToolResultTranscript`. Unrelated to this change; mentioning in case they are flaking for you too.

## Test

What each layer pins, and why it would fail on a wrong implementation:

- `TestAgentIdentityFingerprint` (execenv) — the signal moves when the identity moves (rename, instructions edit, different agent, **an edit buried under the operator's own H2**, a crafted name that tries to mask an edit) and holds still for everything else (same agent on a new issue must compare *equal*, or every ordinary follow-up would announce a change). Also pins that length framing makes field boundaries unambiguous.
- `TestPriorAgentIdentityFingerprintRoundTrip` (execenv) — what `InjectRuntimeConfig` stamps must read back as exactly what this run fingerprints; a managed block written before the marker existed, and a user file that merely *mimics* the marker outside any managed block, both read as "unknown" rather than as a change.
- `TestAgentIdentityChangedNoticeNeverEntersTheBrief` (execenv) — the notice must not reach either brief builder. The previous version of this test claimed to pin that and never inspected brief output.
- `TestAgentIdentityChangedSincePriorRun` (daemon) — the decision in both directions, including fresh-session, no-identity and fresh-workdir cases that must stay silent.
- `TestBuildPromptAgentIdentityChangedNotice` (prompt) — reaches the per-turn message, says the new identity is authoritative, does not claim lost context, survives a flagged-but-resuming turn, and is dropped on the cold retry.

Verified by breaking the implementation on purpose and confirming the failures land where expected:

| deliberate break | what fails |
| --- | --- |
| fingerprint only the name (i.e. #5909's rule) | every instructions-edit case, including the two H2 ones |
| read the prior identity *after* `InjectRuntimeConfig` | `rename announces` and `instructions edit announces` |
| drop the live-`PriorSessionID` gate | `dropped by the cold retry that resumes nothing` |
| stop stamping the fingerprint | the round-trip and the pre-marker degradation case |

Writing the tests also caught a defect before review: a task carrying an `AgentID` but no name and no instructions rendered a heading-only section that the old guard let through, announcing a change on an identity we did not know. Under the fingerprint that shape has a clearer answer — a bare ID is a real identity, so stripping name and instructions off an agent *is* a change, and the test now says so.

`go build ./...`, `go vet ./internal/daemon/...` clean, changed files `gofmt`-clean, and the full `./internal/daemon/...` suite passes (baselined against clean `main` first).

Refs #5736, #5909


Bohan-J:
Closing this one. You picked it up at our invitation, so the reversal is ours to explain rather than yours to absorb.

## The premise still isn't established

When I closed #5909 I pointed at this cheaper line and said it would be a welcome PR. What I under-weighted is that the conditional in the sentence before it was load-bearing: *if a reproducible case does turn up*. It hasn't. #5736 closed for lack of reporter follow-up, and the symptom — an agent continuing to act as its previous identity across a resume — has still never been observed.

Your reading of the resume path remains accurate: there is no identity comparison on it. But a gap in the code and a failure users experience are different claims, and only the first one is evidenced. The tests here pin that the string reaches the prompt; none of them show a model changing behaviour because of it.

## And the cheap fix turns out not to be free

Review surfaced a stable false-positive path that neither of us had in view.

`AgentIdentityFingerprint` digests `AgentInstructions` — but by the time that field reaches the daemon it is not the agent's configured instructions. The server concatenates the squad-leader briefing into the same field (`server/internal/handler/daemon.go:2532`), gated on `is_leader_task`:

1. A squad leader is woken as leader on an issue → instructions = own + briefing → fingerprint F1 stamped into the brief.
2. Someone later `@`-mentions that same agent on the same issue → `EnqueueTaskForMention`, `is_leader_task=false` → no briefing → F2.
3. `GetLastTaskSession` keys only on `(agent_id, issue_id)`, so that turn resumes the session from step 1.
4. F2 ≠ F1 → "Your Agent Identity was updated" fires, with nobody having renamed or edited anything.

Two cheaper flips exist on the same field: a transient `ListSquadMembers` failure degrades the roster (`squad_briefing.go:231`), flipping the fingerprint on that run and back on the next; and the roster embeds other agents' names and skills, so editing a *different* agent moves this one's fingerprint. Traced through the code by one reviewer and confirmed with a temporary regression test by a second, independently.

That inverts the trade. Your own doc comment states the standard — "announcing one that did not happen teaches the agent to distrust the notice" — and as it stands the notice fires reliably for squad leaders, the role with the longest single-issue conversations, while the case it exists for remains unobserved. A fix that misfires in a real scenario to guard an unobserved one is the same shape of argument that closed #5909, just pointing the other way.

## What would reopen this

One concrete path, not a class of them:

- which agent, what exactly changed (rename? instructions? both), and when relative to the run;
- the issue or task where the follow-up turn resumed;
- what the agent actually did that it shouldn't have — a transcript excerpt where it introduces itself as, or applies a rule from, the previous identity.

A single instance is enough. With one in hand this is worth doing properly, and "properly" probably means splitting the squad briefing out of `Agent.Instructions` server-side so the fingerprint has a stable input, rather than digesting a field the server composes per task. That is the ordering I would want even if the notice itself lands unchanged.

## Worth keeping regardless

Two findings from the review outlive this PR, and we will file them separately:

- The same concatenation means squad-leader briefs **already** violate the MUL-5377 byte-identical invariant — the existing `ByteIdentical` tests vary `TaskContextForEnv` and cannot see server-side variation. Your fingerprint is an accidental detector for it.
- `ctx.AgentName` reaching the brief unsanitised (`runtime_config_sections.go:114`) while sibling name fields go through `sanitizeNameForBriefMarkdown`. You flagged this yourself and were right to keep it out of scope.

The mid-PR correction from comparing rendered Markdown to digesting the fields was the right call, and the reasoning you wrote for it — that user instructions routinely carry their own `##`, so any section-slicing rule silently compares equal on exactly the edits this exists to catch — is worth more than the patch itself. Thanks for that, and sorry for the round trip.

PR #8363, opened 2026-09-13
MUL-7331: Fix duplicate comment migration prefix

alithia-dev:
## Summary

- renumber the later, unshipped `comment_deleted_at` migration from 468 to 469
- leave `468_drop_reference_only_column` unchanged
- restore unique numeric migration prefixes on canonical main

## Safety evidence

- latest release `v0.4.43` points to `2ae2dbbb8` and does not contain the migration
- the intended NEXUS production CLI reports `0.4.43` at the same commit
- GitHub deployment history after the migration merge contains only `multica-docs`, not an application/database deployment
- no tags contain `a9e82c797`
- no source references require updating

## Validation

FORGE static gates: clean rename-only diff and `git diff --check`.

TESS to run:
- `TestMigrationNumericPrefixesAreUnique`
- relevant migration up/down coverage
- canonical backend tests

Blocks JLF-573 / PR #8362 until merged through the protected path.


multica-eve (changes requested):
Reject: this PR is based on an obsolete migration sequence and should not be merged.

Current `main` already assigns `469` and `470` to the issue-status migrations. A clean merge of this PR therefore moves the duplicate from `468` to `469`; the repository's migration invariant fails with `469_comment_deleted_at` and `469_issue_status_lifecycle_categories` sharing the same numeric prefix.

The maintainers will address the original duplicate-prefix problem in a separate official fix, so we are closing this PR rather than asking you to rework it.

Issue #2405, opened 2026-05-11
[Feature]: system tray implementation for Linux


isneoa:
### Deployment type

Self-hosted

### What do you want and why?

 there is no system tray implementation for Linux in the Multica desktop application

PR #8364, opened 2026-09-13
MUL-7328 feat(desktop): add configurable window close behavior

### LINKED ISSUE #8361: [Feature]: Desktop app setting to minimize to system tray or taskbar on window close
### Deployment type

Official App

### Area

Desktop app

### What do you want and why?

Currently on Windows (and Linux), clicking the window close button (`X`) closes and destroys the main window, causing the entire application to exit (`app.quit()`).

If a user accidentally clicks the close button while an agent is executing long-running tasks, the app process terminates abruptly, dropping WebSocket connections and interrupting active agent runs.

Currently, there is no system tray implementation and no setting to control the behavior when the close button is pressed.

1. **Window Close Behavior Setting**: Add a setting in Desktop Preferences (e.g., `Settings` -> `Device` / `Desktop`):
   - **Minimize to system tray** (recommended default): Hides the window to the notification area; background agent tasks, daemon monitoring, and connections stay alive.
   - **Minimize to taskbar**: Standard minimize behavior.
   - **Quit application**: Exits the app directly (current behavior).
2. **System Tray Integration**:
   - Single / double click restores and focuses the window.
   - Right-click context menu provides "Open Multica" and "Quit Multica".

(Related to #2405 for general tray support)

Dante-dan:
## What does this PR do?

Adds a Desktop close-behavior preference so closing the main window can hide Multica to the system tray (the default), minimize it to the taskbar, or quit the app.

The main process owns the preference and close interception so renderer state cannot bypass the lifecycle rule. Preferences are written atomically, invalid persisted values fall back to the default, and explicit Quit still follows the existing `before-quit` cleanup path. The tray reuses the lifecycle primitives introduced in #2818, adds localized Open/Quit actions, and completes the close-to-tray behavior that PR intentionally left for follow-up.

The new Settings → Desktop control, landing copy, and Desktop documentation are available in English, Simplified Chinese, Japanese, and Korean.

Risk is limited to the Electron window lifecycle. The tests cover all three close modes, corrupted preference fallback, persistence, renderer IPC, localized tray actions, and explicit quit behavior.

## Related Issue

Closes #8361

Related: #2405

## Type of Change

- [x] Documentation update

## Changes Made

- Add a validated, atomically persisted Desktop close-behavior preference and IPC boundary.
- Intercept main-window close events for tray, taskbar, and explicit quit behavior.
- Add a localized tray menu with Open and Quit actions.
- Add a Settings → Desktop control and localized product copy.
- Document the behavior and sync the landing changelog in all four shipped locales.

## How to Test

1. Run the four targeted Desktop test files; all 17 tests pass.
2. Run Desktop node and web TypeScript checks.
3. Run the Desktop build and changed-file ESLint checks.

## Checklist

## AI Disclosure

**Prompt / approach:** I traced the current Electron lifecycle, preference, and settings architecture; compared the overlapping tray work in #2818; implemented the smallest main-process-owned policy; added targeted tests; and addressed independent review findings before submission.


Bohan-J:
Thanks for the thorough work here — the tests, the four-locale copy, and the atomic preference write are all solid, and the implementation itself is clean. The problem is upstream of the code: **the premise this PR is built on does not hold in this codebase**, and that changes what the feature should look like.

## The premise

Both the issue and this PR assume that quitting Desktop interrupts running agent work:

> If a user accidentally clicks the close button while an agent is executing long-running tasks, the app process terminates abruptly, dropping WebSocket connections and interrupting active agent runs.

The daemon is not a child of the Electron process in any lifecycle-meaningful sense, and in-flight agent work is deliberately protected in several independent places:

- `server/cmd/multica/cmd_daemon_unix.go:19-21` — the daemon is spawned with `Setsid: true`, detaching it from the parent's session and process group.
- `server/cmd/multica/cmd_daemon_windows.go:50-59` — on Windows it is spawned with `DETACHED_PROCESS`, plus `CREATE_BREAKAWAY_FROM_JOB` where the parent Job Object permits it. That flag exists specifically so the daemon escapes the Job Object that would otherwise kill it when the parent dies.
- `apps/desktop/src/main/daemon-manager.ts:77` — `DEFAULT_PREFS = { autoStart: true, autoStop: false }`. Auto-stop on quit is **off by default**.
- `apps/desktop/src/main/daemon-manager.ts:1481-1501` — the `before-quit` handler is the only quit path that touches the daemon at all, and it only stops it when `autoStop` is enabled.
- `apps/desktop/src/main/daemon-manager.ts:590-594` — the automatic version-mismatch restart is gated on `active_task_count === 0`, commented "no in-flight agent work would be killed".
- `server/cmd/multica/cmd_setup.go:296` — the CLI refuses to restart a daemon with active tasks: *"left the running daemon unchanged to avoid cancelling work"*.

So under the default configuration, quitting Desktop does **not** kill running agent tasks. The daemon keeps running and the tasks keep going. What a user actually loses is the window and the live WebSocket view — and relaunching reconnects to the already-running daemon rather than restarting it (and won't restart it while it is busy, per the gate above).

The premise is true only for users who explicitly enabled **Settings → Daemon → Auto-stop on quit**, which is off by default and whose own description already says *"Disable this to keep the daemon running in the background."*

## What that means for this PR

The value proposition drops from "reliability fix that prevents losing your work" to "window-management convenience." That is still a legitimate thing to want — plenty of desktop apps offer it — so this is not a rejection of the idea. But the weaker justification no longer supports the current design, and one part of the PR is actively harmful:

**1. The user-facing copy must be rewritten — this is the most important item.** `packages/views/locales/*/settings.json` and `apps/docs/content/docs/desktop-app*.mdx` currently tell users, in four languages, that tray/taskbar modes "keep active runs and connections alive." The implication is that **Quit endangers your running tasks**, which is the opposite of how the app behaves. This ships an incorrect mental model into official documentation. The copy should talk about the window and the live view only, and say nothing about runs surviving.

**2. The default should stay `quit`.** The argument for flipping the default was "protect the user's work." Without it, this is a preference, and preferences should default to existing behavior. Right now the four new files contain **no platform checks at all**, so `tray` also becomes the default on macOS — a platform the issue never mentioned. macOS users would get a new permanent menu-bar icon nobody asked for, a close button that hides instead of closes, and a settings option labeled "Minimize to taskbar" on an OS with no taskbar. Please either default to `quit` everywhere, or scope any non-`quit` default to Windows only.

**3. Please drop the tray from this PR.** The description says the tray "reuses the lifecycle primitives introduced in #2818" — #2818 is still **open and unmerged**, and there is no `Tray` code in the repo today. So this is a second, competing tray implementation rather than a reuse, and #2818's version also carries daemon status. Linux tray support is separately tracked in #2405, also open. Landing a tray here would pre-empt both. A close-behavior preference offering **Quit** and **Minimize to taskbar** needs no tray at all and could land on its own.

## Other must-fix items (if the tray stays in any form)

**`setupTray` is unguarded and registered before critical setup.** At `apps/desktop/src/main/index.ts:888` it sits between `createWindow()` and `setupAutoUpdater` / `setupDaemonManager` / `setupLocalDirectory` / `app.on("activate")`, with no `try`/`catch`. `new Tray()` is a native constructor and tray availability on Linux depends on the desktop environment. If it throws, none of those later handlers register and the user gets an app that never starts its daemon, never auto-updates, and ignores dock activation on macOS. Move the call below those three and wrap it, degrading to "no tray" on failure.

**Close-to-tray as a Linux default has no availability fallback.** GNOME ships no StatusNotifier host without the AppIndicator extension, so the icon may simply not appear. The window is then hidden with no visible way to restore or quit. There is a recovery path — relaunching hits the single-instance branch, and `focusMainWindow` (`index.ts:164`) does call `show()` — but a user who does not know that is stuck. Tray availability should be detected, or Linux should fall back to taskbar.

**`appIsQuitting` is a latch with no reset.** `requestAppQuit()` sets it before calling `app.quit()`, and nothing ever clears it. `daemon-manager.ts:1490` calls `event.preventDefault()` on `before-quit` when `autoStop` is enabled, intending to cancel the quit. If a quit is ever cancelled, `applyMainWindowCloseBehavior` returns early forever and the user's close preference silently stops working for the rest of the session. To be fair: that `preventDefault()` is called asynchronously inside `loadPrefs().then(...)`, and Electron reads `defaultPrevented` right after the synchronous emit, so it most likely does not actually cancel the quit today. The latch is still fragile — please reset it when a quit does not complete.

## Nits

- The tray icon is not a macOS template image. `tray-manager.ts` does `nativeImage.createFromPath(...).resize({ width: 18, height: 18 })` without `setTemplateImage(true)`, so it won't adapt to light/dark menu bars — and `apps/desktop/resources/icon.png` is a 387 KB full-colour app icon, which will look muddy at 18×18.
- The tray menu builds its own translation table instead of using the app's i18n. `labelsByLocale` keys off `app.getPreferredSystemLanguages()[0]`, i.e. the OS language, not the language the user picked in Multica — so a user running the app in Chinese on an English OS gets an English tray menu. The test also pins `zh-TW` to Simplified Chinese labels, which will get in the way if Traditional Chinese is ever added.
- `desktop-preferences.ts` duplicates `updater-preferences.ts` almost line for line (path helper, parse-with-fallback, temp-file + rename). Consistent with the existing convention, so not blocking, but worth extracting a shared helper before a third one appears.
- `desktop-settings-tab.tsx` doesn't check that `window.desktopPreferences` exists. The `.catch()` only handles promise rejection, not the synchronous `TypeError` if the bridge is missing.

## What I verified

- `pnpm vitest run` in `apps/desktop` — 61 files / 673 tests pass, including your 4 new files / 17 tests. The "17 tests pass" claim checks out.
- `pnpm typecheck` (node + web) — clean.
- Not verified: I did not run Electron, so per-platform close behavior, tray rendering, and macOS menu-bar appearance are untested here. The GNOME tray claim comes from Electron's platform documentation and the open #2405, not from a live test.

## Suggested path forward

Split it: land a close-behavior preference with **Quit** (default) and **Minimize to taskbar**, no tray, with copy that only describes window behavior. Let the tray land through #2818, where it belongs.

One thing worth considering separately: if the reporter's real fear is "I lost my running task" and the reality is "your task is fine, you just can't see it," then the most valuable fix might be neither a tray nor a close-behavior setting, but simply telling the user on close that running tasks continue in the background. That's much cheaper and addresses the actual anxiety. Worth raising on #8361.

PR #8368, opened 2026-09-14
MUL-7336: feat(labels): add skill label CLI commands and skills list label filtering

klinsmaya:
## What does this PR do?

Brings Skill labels — which the web app already supports — to the CLI, and makes them usable for finding skills in the web app.

Two connected pieces:

1. **CLI.** `multica label` becomes resource-type aware (`--resource-type issue|skill`, plus `--description`), and a new `multica skill label list/add/remove` manages the labels attached to a specific skill. Before this, labels were an Issue-only concept from the command line: a script could create and attach Issue labels but had no way to reach Skill labels, even though the web UI could.
2. **Skills list filtering.** The skills page toolbar gains a **Labels** filter alongside the existing Usage / Source / Agent / Creator dimensions, so a workspace with many skills can narrow the list to one category. Selecting several labels is OR-within-labels and AND against the other dimensions, matching the Issue list's semantics.

Agent labels are deliberately **not** included. `#6245` removed Agent labels from the product surface while keeping the backend model; adding CLI commands for them would have made the CLI the only place they could be created, with no way to see the result. The CLI now rejects `--resource-type agent` with `resource type must be issue or skill`.

## Related Issue

MUL-7336

## Type of Change

- [x] Documentation update
- [ ] CI / infrastructure

## Changes Made

**CLI (`server/cmd/multica/`)**

- `cmd_label.go`: `--resource-type` on `label list` (filter) and `label create`; `--description` on `label create` / `label update` (`update` uses `Flags().Changed` so a description can be cleared, not just set); `RESOURCE TYPE` column in table output; `validateLabelResourceType` rejects anything other than `issue` / `skill`.
- `cmd_skill_label.go` (new): `multica skill label list|add|remove <skill-id> [<label-id>]`, backed by `/api/skills/{id}/labels`.
- `cmd_id_resolver.go`: `resolveLabelID` now takes a resource type and queries `/api/labels?resource_type=...`. Without this, short IDs copied out of `label list --resource-type skill` failed to resolve, because `/api/labels` defaults to `issue` server-side. `label get/update/delete` pass an empty type and search both catalogs.
- `cmd_issue_label.go`: passes `"issue"` explicitly — no behavior change.

**Server (`server/internal/handler/skill.go`)**

- `GET /api/skills` now includes a `labels` array per row, bulk-loaded through the existing `ListLabelsForSkills` query (one round-trip for the whole page, not N+1). The field is always present — `[]` when a skill has none. If the label query fails, the list still returns skills without labels rather than failing the request.

**Web (`packages/core`, `packages/views`)**

- `SkillSummarySchema` splits out of `SkillSchema`: the list shape no longer defaults `content` / `files`, so list parsing stops inventing empty detail bodies (GH #2174); `SkillSchema` extends it for the detail endpoint. `labels` uses `.catch([])` so an older backend that omits the field, or a malformed entry, degrades to "no labels" instead of dropping the skill.
- `listSkills()` goes through `parseWithFallback`.
- `labels` added to `SkillListFilters`; the persisted view store already deep-merges filter defaults, so a payload saved before this dimension existed rehydrates with `labels: []`.
- Renaming or deleting a label now invalidates the workspace skills cache, so chips in the list don't go stale.
- The row predicate moves out of `skills-page.tsx` into `skill-list-filter.ts` as a pure `rowMatchesFilters`, which is what the new unit tests exercise.
- Toolbar label submenu with a search box, per-label counts from the loaded rows, and the four locale files.

**Docs (`apps/docs/content/docs/cli*.mdx`)**

- `skill label list/add/remove` row and the new `label` flags, in all four locales (en / zh / ja / ko).

**Tests**

- `cmd_label_test.go`: `--resource-type` is sent on create and list.
- `cmd_skill_test.go`: list / add / remove round-trips, including a short-ID add that asserts the resolver queries `resource_type=skill`.
- `skill_list_test.go`: `GET /api/skills` returns attached labels, returns `[]` for unlabeled skills, and still omits `content`.
- `schema.test.ts`: missing, well-formed, and malformed `labels` payloads.
- `skills-page-filter-predicate.test.ts`, `view-store.test.ts`: filter semantics and pre-labels rehydration.

## How to Test

1. `cd server && go test ./cmd/multica/... ./internal/handler/...`
2. `pnpm test` in `packages/core` and `packages/views`
3. CLI round-trip against a running server:
   ```bash
   multica label create --name frontend --color '#3b82f6' --resource-type skill --description 'Frontend work'
   multica label list --resource-type skill        # short IDs in this output now resolve
   multica skill label add <skill-id> <label-id>
   multica skill label list <skill-id>
   multica skill label remove <skill-id> <label-id>
   multica label create --name x --color '#fff' --resource-type agent   # rejected
   ```
4. Web: open the skills page, use the filter menu's **Labels** section, confirm multi-select is OR and combines with the other filters.

## Checklist

## Risks and rollback

CLI-side, the blast radius is new CLI users; existing Issue-label scripts are unaffected because `--resource-type` defaults to `issue`. Web-side, the filter is additive and inert until a label is selected. The new `labels` field on `GET /api/skills` is additive and tolerated by older clients. Rollback is a plain revert; no migrations and no stored data are touched.

## AI Disclosure

**AI tool used:** GitHub Copilot

**Prompt / approach:** Implemented `--resource-type` support on `multica label`, added `multica skill label` subcommands with resource-type-aware short-ID resolution, and extended the skills list with label data and a label filter, with CLI, handler, and web test coverage.

## Screenshots

Not included — the new UI is a submenu in the existing skills filter dropdown (**Filter → Labels**).


klinsmaya:
This is an external Vercel team-authorization requirement, not actionable PR feedback. An IndexLabs Team member should authorize @klinsmaya using the Vercel link, or the deployment can be ignored if it is not required; no code or CI change is needed.

Bohan-J (changes requested):
Thanks for this — extending label management to the CLI is the right direction, and the Skill half is exactly what's missing today (Web can manage Skill labels; scripts can't). Requesting changes because the PR currently also re-opens a product surface we deliberately removed, and because the Skill path has a broken default flow.

**Please narrow this PR to Skill labels only.**

### 1. Drop the Agent label surface (blocking)

`#6245` / `MUL-5600` (merged) removed Agent labels from the product: the Agents catalog is gone from Label Settings, and the Labels row is gone from `AgentDetailInspector`. That PR explicitly kept the backend — `resource_type = "agent"`, `agent_to_label`, and `GET/POST/DELETE /api/agents/{id}/labels` — so the model stayed honest, not as an invitation to re-expose the feature through another client. Its own wording notes that "the `multica label` CLI never supported agent labels."

So `server/cmd/multica/cmd_agent_label.go` plus `--resource-type agent` would make the CLI the only place in the product where Agent labels can be created and attached — a CLI-only path to a surface with no way to view or manage the result. Please:

- remove `cmd_agent_label.go` and the `cmd_agent_test.go` additions;
- limit the `--resource-type` help/accepted values to `issue` and `skill` on `label list` / `label create`.

If reviving Agent labels is actually wanted, that needs a maintainer decision that overturns `MUL-5600` first, and then a design covering creation, listing, filtering and the Agent detail entry point — not a CLI-only channel. (The linked issue description framing Agent as a supported use case is inaccurate for the same reason.)

### 2. Short IDs don't resolve for Skill labels (blocking)

`label list --resource-type skill` prints truncated IDs by default, but `resolveLabelID` → `fetchLabelCandidates` (`server/cmd/multica/cmd_id_resolver.go:413`) still requests `/api/labels?workspace_id=...` with no `resource_type`, and the server defaults that to `issue` (`defaultLabelResourceType` / `parseLabelResourceType`, `server/internal/handler/label.go:85`, used by `ListLabels` at `:144`).

The result is that the natural flow is broken end to end:

```bash
multica label list --resource-type skill     # prints e.g. 3f2a1b9c
multica skill label add <skill-id> 3f2a1b9c  # resolve label: no label found
```

`label get/update/delete` on a Skill label short ID fail the same way. Only a full UUID works today. Please make candidate fetching resource-type aware (e.g. thread the type through, or search across the types the CLI exposes) and add a regression test covering short-ID resolution for a Skill label — the existing tests pass full UUIDs, so they don't catch this.

### 3. Update the public CLI docs

`apps/docs/content/docs/cli.mdx` documents commands at subcommand granularity (there's a `label list/add/remove <issue-id>` row under `issue`, and a `label list/get/create/update/delete` row). The new `skill label list/add/remove` subcommands and the new `--resource-type` / `--description` flags aren't there. Please add them, including the `cli.zh.mdx` / `cli.ja.mdx` / `cli.ko.mdx` locale copies.

### Notes

- The rest of the change reads well: `--resource-type` defaults to `issue` on both `list` and `create`, so existing Issue-label scripts keep working, and `label update` correctly uses `Flags().Changed("description")` so an empty description can be cleared rather than ignored.
- CI: `backend-tests` is red on `TestMigrationNumericPrefixesAreUnique` — `468_comment_deleted_at` and `468_drop_reference_only_column` share prefix `468`. That's pre-existing on `main` and unrelated to this diff (no migrations here); it needs fixing separately, not in this PR.
- Scope-wise the change is CLI-only, so blast radius is limited to new CLI users and rollback is a CLI release revert; server behavior and stored data are untouched.

klinsmaya (commented):
Implemented the requested scope changes: removed the Agent label CLI surface and limited label resource types to issue and skill. Skill label prefix resolution now queries skill labels (and generic label ID resolution searches both exposed types), with regression coverage for short IDs. Updated the English, Chinese, Japanese, and Korean CLI docs for the new flags and skill label commands. Focused validation passes with `cd server && GOPROXY=https://goproxy.cn,direct go test ./cmd/multica/...`.

Bohan-J (approved):
✅ **Approve.** All three blockers from the previous round are genuinely fixed, the follow-up work (skills list label filtering) is a sensible completion of the feature rather than scope creep, and CI is green. Two nits below, neither blocking.

I also rewrote the PR title and description myself — they still described the original agent-inclusive scope and omitted the list-filtering work entirely. No code was touched.

### What this delivers, in user terms

A workspace admin can now do from the CLI what was previously only possible in the web UI: create a Skill-scoped label and attach it to a skill. Before, `multica label` only understood Issue labels, so organizing skills from a script meant falling back to naming conventions. After, `label create --resource-type skill` plus `skill label add/list/remove` closes the loop.

The second half closes a different loop: labeling skills was pointless if nothing consumed the labels. The skills page toolbar now has a **Labels** filter next to Usage / Source / Agent / Creator, so in a workspace with a long skill list you can narrow to one category. Multi-select is OR-within-labels and AND against the other dimensions — the same semantics the Issue list already uses, so there's nothing new to learn.

Nothing changes for anyone who doesn't opt in: `--resource-type` defaults to `issue`, existing Issue-label scripts keep working unchanged, and the filter is inert until a label is selected.

### Verification of the previous round's blockers

1. **Agent labels dropped.** `cmd_agent_label.go` and its tests are gone from the branch. `validateLabelResourceType` rejects anything but `issue` / `skill`, and the flag help and docs list only those two. `#6245`'s decision stands.
2. **Short-ID resolution fixed.** `resolveLabelID` now threads a resource type into `fetchLabelCandidates`, which queries `/api/labels?resource_type=...` (`cmd_id_resolver.go:409`). `skill label add/remove` pass `"skill"`; `label get/update/delete` pass `""` and search both catalogs. The regression test in `cmd_skill_test.go` fails the request handler if the resolver's `/api/labels` call arrives without `resource_type=skill`, so it actually guards the fix rather than just covering the happy path.
3. **Docs synced** across `cli.mdx`, `cli.zh.mdx`, `cli.ja.mdx`, `cli.ko.mdx`.

### On the newly added scope

The server side is the right shape. `ListSkills` bulk-loads through the existing `ListLabelsForSkills` query rather than per row, so this is one extra round-trip for the whole page, not N+1. `labels` is always present (`[]` when empty) so clients don't have to distinguish "no labels" from "old backend". A failed label query degrades to skills-without-labels with a `slog.Warn` instead of failing the list — the right call for a non-critical display field, and worth knowing that in that state an active label filter will show an empty list with no error surfaced. Low enough frequency that I don't think it needs handling now.

The web side is defensive in the ways that matter: `labels` uses `.catch([])`, so an older backend or a malformed entry degrades to "no labels" instead of dropping the skill from the list; the persisted view store already deep-merges filter defaults, so a filter payload saved before this dimension existed rehydrates with `labels: []` rather than crashing on `.length`. Splitting `SkillSummarySchema` out of `SkillSchema` also fixes a real adjacent problem — list parsing no longer defaults `content` / `files` into every row (GH #2174). Extracting `rowMatchesFilters` into `skill-list-filter.ts` is what makes the filter semantics unit-testable; the predicate tests are the useful ones here.

`git log` shows no prior decision this conflicts with, and no policy constant in the diff is unexplained.

**If this is wrong**, it shows up as either a skills list missing label chips or a CLI `no label found`; rollback is a plain revert plus a CLI release revert. No migrations, no schema changes, no stored data touched.

### Nits (non-blocking)

- `server/cmd/multica/cmd_label.go:15` — the header comment still reads `labels (issues, agents, skills)`. Should be `issues and skills`; as written it contradicts the validation two dozen lines below.
- `server/cmd/multica/cmd_label.go:56` — `validateLabelResourceType(value string, allowEmpty bool)` is called with `allowEmpty: false` at both call sites (`:105`, `:198`), and the list path already guards on `v != ""` before calling. The parameter is dead; dropping it makes the contract clearer.

### Verification scope

Reviewed the full diff at `51a6103` against `main`, and read `handler/label.go`, `cmd_id_resolver.go`, `labels/queries.ts`, and the `view-store` persist path on the branch to confirm the claims above. CI on this head is green across all jobs, including `backend`, `backend-tests`, `frontend-views-test`, and `sqlc-check` — the `468_*` migration prefix collision that was red last round is resolved. I did not run the app locally, so the filter UI is reviewed as code, not exercised.

Bohan-J:
Thanks for sticking with this through two rounds — the reworked version is a clear improvement over the original scope.

Two things I want to call out. Narrowing to Skill labels after the `#6245` discussion was the right call, and you didn't just delete the agent commands: you tightened `validateLabelResourceType` so `--resource-type agent` fails loudly instead of silently reaching a surface with no UI. And the short-ID fix came with a regression test that actually guards it — the test server fails the request if the resolver omits `resource_type=skill`, rather than only covering the happy path. That's the version of the fix that stays fixed.

Adding the skills list label filter was also the right instinct. Labeling skills from the CLI wasn't worth much if nothing consumed the labels; bulk-loading them in `ListSkills` instead of per row, and degrading to "no labels" when the query fails, are both the choices I'd have wanted.

I pushed one small commit (`6807f3b`) covering the two nits from my review — the stale `(issues, agents, skills)` header comment and the dead `allowEmpty` parameter. No behavior change; `go test ./cmd/multica/...` passes locally. I also rewrote the PR title and description, which still described the original agent-inclusive scope and didn't mention the list filtering at all.

Merging once CI is green.

PR #8376, opened 2026-09-14
MUL-7340 fix(lark): share bind-session state across replicas so the QR survives the first status poll

### LINKED ISSUE #8366: [Bug]: install session not found
### Deployment type

Official App

### Area

Web app

### What happened?

绑定飞书的时候，弹出二维码，然后打开链接，一旦链接打开之后，二维码立即会提示：install session not found

### Steps to reproduce

绑定飞书的时候，弹出二维码，然后打开链接，一旦链接打开之后，二维码立即会提示：install session not found

### Screenshots (optional)

_No response_

```shell

```

Bohan-J:
## The bug

The install dialog dies ~5 seconds after the QR renders, showing `安装会话已失效或丢失，请重新扫码` with the diagnostic `install session not found`.

That copy is the frontend's `session_lost` branch, reached only on a **404** from `GET /api/workspaces/{id}/lark/install/{session_id}/status`. The browser's first status poll fires at `max(2000, poll_interval_seconds * 1000)` = 5s — exactly the reported timing.

The session existed only in one process's memory:

```go
// RegistrationService
mu       sync.Mutex
sessions map[string]*registrationSession
```

`begin` populated that map on whichever backend process served it. Any status poll routed to a *different* process — a second replica, or the same one after a restart/deploy — found nothing and 404'd. Nothing about this is recoverable client-side: the dialog correctly stops polling, because a genuinely missing session will never appear.

Every other 404 source was traced and ruled out first: the old GC skipped pending sessions, the package's single `delete` sat inside that guard, workspace scoping and the initiator/admin check both pass for an owner, and the status discriminator strings match on both sides (`"pending"`).

## The fix

The in-flight bind session is transient cross-replica coordination state, so it goes where this repo already puts that shape — Redis, with an in-process implementation for single-instance development. Same split as the runtime update / model-list / local-skill request stores. PostgreSQL keeps what it should own: the committed Bot installation and the installer binding.

`InstallSessionStore` has two implementations. The router picks the Redis one whenever Redis is configured; without it the service keeps the in-process store and logs a warning, because that configuration is correct for local development and a single replica and reproduces this bug on more than one — so it says so rather than failing quietly at the first status poll.

**The record is one Redis hash written by Lua.** Both writes are atomic:

- `Create` writes the immutable half and its retention in one step, so a failure cannot leave a key with no TTL behind.
- `MarkTerminal` writes the status only when none is set — first-writer-wins, so when the expiry deadline and a poll result race the user keeps whichever outcome they were already shown — and runs `EXPIRE` **unconditionally**, which is what makes a replay converge instead of no-op'ing.

To be precise about which property does the work here, since it is easy to overstate: Redis does **not** roll back a script that fails partway through, so Lua atomicity buys isolation from other clients, not all-or-nothing recovery. The original two-key bug is removed by (1) one key, so status and retention share a single expiry, and (2) a retry that keeps the first status and re-asserts the TTL.

**The terminal write is retried with backoff.** By the time `markSuccess` runs, the installation and binding are already committed in Postgres; dropping that write on the first error left a user who had successfully bound watching the dialog sit on `pending` until the QR expired. 10 attempts — 9 waits of 200ms doubling into a 15s cap, 55.4s of backoff, plus up to `sessionWriteTimeout` per attempt — sized to ride out a Redis failover rather than just a dropped packet. This is a bounded recovery window, not a guarantee: an outage that outlasts it still releases the session. A store that reports the session gone stops the loop immediately, since no retry can fix that.

**Retention is a per-session TTL** — the QR window plus `SessionTTL` — so a terminal outcome recorded at the last second is still readable. Deliberately not a fixed constant, which would drift from Lark's `expires_in` (currently 1h).

**A Redis error surfaces as an error, never as not-found.** Not-found is terminal in the dialog, so a transient blip must not permanently kill a live scan.

## Scope of the fix

Sharing the status fixes the 404. It does **not** make an in-flight scan resumable elsewhere. The `device_code` and the polling goroutine stay in the process that served `begin` — deliberately, since the `device_code` is a bearer credential. If that process dies, other replicas serve `pending` until `ExpiresAt`, at which point the read path reports it expired. That is an improvement on hanging forever, not a claim that the original QR still completes.

## Verification

Against a real PostgreSQL and a real Redis:

- `go build ./...`, `go vet ./internal/... ./cmd/...` — clean
- `go test ./internal/integrations/lark/` — pass
- `go test ./internal/handler/` — pass

The store suite runs identical assertions against **both** implementations (`forEachInstallSessionStore`), which is what keeps them from drifting: round trip, unknown vs cross-workspace indistinguishability, first-writer-wins, installation id on success, TTL retention, and not-found on a missing session. Redis-specific:

- `TestRedisInstallSessionStoreIsSharedAcrossClients` — the regression at storage level: two stores with *separate* Redis clients, standing in for two replicas, must see one session.
- `TestRedisInstallSessionStoreFailedRequestWritesNothing` — a write that never reaches Redis applies nothing, and the retry moves status and retention together.
- `TestRedisInstallSessionStoreReplayAfterLostResponse` — the case the two-key version actually lost: the server ran the write, the client never saw the reply, and the replay must keep the first outcome AND re-assert retention, so the record expires as one piece and never reads `pending` again.
- `TestRedisInstallSessionStoreFailureIsNotNotFound` — a closed client must not report not-found.

Service level:

- `TestRegistrationGetSessionServesSessionOwnedByAnotherProcess` — the original regression: instance A serves begin, instance B (empty map) serves the poll.
- `TestRegistrationTerminalWriteSurvivesTransientStoreFailure` — a committed bind survives a store outage and lands on the replica the browser polls.
- `TestRegistrationTerminalWriteStopsWhenSessionIsGone` — one attempt, not the whole budget, when no retry could help.

Redis tests gate on `REDIS_TEST_URL` and skip when unset, matching the existing `DATABASE_URL` gating; they were run against a real server rather than left skipped.

Two handler tests (`TestCommentSourceContextLifecycle`, `TestReportTaskMessagesFallsBackWholeBatchForClockSkew`) flake in the full-suite run. Both reproduce on clean `main` without this change — pre-existing, unrelated, not addressed here.

Fixes #8366


Bohan-J:
Reworked the storage design per review — the SQL table is gone.

The transient scan state didn't need a table of its own. This repo already has a home for exactly that shape (a short-lived, cross-replica request whose status a browser polls): the runtime update / model-list / local-skill stores, which use Redis when configured and an in-process implementation for single-instance development. This now follows the same pattern.

**Removed:** `lark_install_session` table, its sqlc queries, both migrations, the sweep, and the workspace-teardown entry the table required.

**Added:** an `InstallSessionStore` interface with a Redis and a memory implementation. PostgreSQL keeps what it should own — the committed Bot installation and the installer binding.

Three things that changed shape rather than just moving:

- **First-writer-wins now falls out of `SetNX`.** The record splits across an immutable key written by `Create` and a terminal key written at most once, so a racing expiry and poll result resolve without a read-modify-write. The previous version needed an `UPDATE ... WHERE status = 'pending'` to say the same thing.
- **Retention is the store's TTL, sized per session** as the QR window plus `SessionTTL`, so a terminal outcome recorded at the last second is still readable afterwards. Deliberately not a fixed constant — that would drift from Lark's `expires_in` (currently 1h) the moment either changed.
- **A Redis error surfaces as an error, never as not-found.** Not-found is terminal in the dialog, so a transient blip must not permanently kill a live scan.

Without Redis the service keeps the in-process store and the router logs a warning. That configuration is correct for local development and a single replica, and reproduces this bug on more than one replica — so it says so out loud rather than failing quietly at the first status poll.

## On the scope of the fix

Worth restating since the review raised it: sharing the status fixes the 404. It does **not** make an in-flight scan resumable elsewhere. The `device_code` and the polling goroutine stay in the process that served `begin` — deliberately, since the `device_code` is a bearer credential. If that process dies, other replicas serve `pending` until `ExpiresAt`, at which point the read path reports it expired. That is an improvement on hanging forever, not a claim that the original QR still completes.

## Verification

Against a real PostgreSQL and a real Redis:

- `go build ./...`, `go vet ./internal/... ./cmd/...` — clean
- `go test ./internal/integrations/lark/` — pass
- `go test ./internal/handler/` — pass

The store suite runs the same behavioural assertions against **both** implementations (`forEachInstallSessionStore`), which is what keeps them from drifting: round trip, unknown vs cross-workspace indistinguishability, first-writer-wins, installation id on success, and TTL retention. Plus two Redis-specific tests:

- `TestRedisInstallSessionStoreIsSharedAcrossClients` — the regression at the storage layer: two stores with *separate* Redis clients, standing in for two replicas, must see one session. This is the property the old in-process map lacked.
- `TestRedisInstallSessionStoreFailureIsNotNotFound` — a closed client must not report not-found.

Redis tests gate on `REDIS_TEST_URL` and skip when unset, matching the existing `DATABASE_URL` gating. I ran them against a real server rather than leaving them skipped.

Both previously-noted handler flakes (`TestCommentSourceContextLifecycle`, `TestReportTaskMessagesFallsBackWholeBatchForClockSkew`) still reproduce on clean `main` without this change; unrelated and not addressed here.

PR #8382, opened 2026-09-14
MUL-7355 fix(claude): deduplicate fallback usage by response ID

importcpp:
## What does this PR do?

Corrects Claude Code token usage when a run ends without usable final usage. Claude can split one API response into several assistant events sharing the same `message.id` and token counts. Multica summed each event, so a failed run could report 200 input tokens for one 100-token response, with cache reads and writes doubled as well. A valid final usage snapshot already overrides the intermediate totals, which is why ordinary successful runs hide the bug.

Count each identified main-loop response once within an execution. Continue forwarding every text and tool block. Preserve the existing preference for final `modelUsage`, including totals from error results, followed by result-level `usage`.

The fallback also stops counting assistant `output_tokens` placeholders and excludes nested-agent events, following [Claude's documented per-step accounting and crash recovery](https://code.claude.com/docs/en/agent-sdk/cost-tracking#track-per-step-usage). It can recover main-loop input/cache usage, not a complete bill.

## Related Issue

Discovered while auditing Multica's Claude token consumption; no separate GitHub issue.

This is independent of #8198's session-resume budget and #8339's worker-result handoff. #8075 changes native-loop lifetime and cumulative terminal snapshots in `claude.go`; it does not deduplicate these assistant events. It shares the message type and assistant-dispatch area, so those edits need reconciliation when combining the PRs. Its patch was inspected, not stacked or tested with this change.

## Type of Change

- [x] Bug fix
- [x] Tests
- [x] Documentation update

## Changes Made

- `claude.go`: keep an execution-local set of counted response IDs; recover only main-loop input/cache tokens while preserving stream delivery and terminal usage precedence.
- `claude_usage_test.go` and the existing test helper: drive the real adapter with a test-binary stream-json fixture, without resolving an installed agent executable or accessing an account.
- `claude_test.go`: pass the execution-local accounting state in existing parser tests.
- Runtime skill reference: explain the fallback's partial coverage and zero output sentinel.

## How to Test

At main `361edd90a1aff8c42e4f990edf830c58cd897b85`, the new fixture tests fail the intended usage assertions: input/cache values double, nested events enter the fallback, and output placeholders are counted. Final-usage control cases pass.

From `server/`:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run '^TestClaude(ExecuteFallbackUsage|FallbackUsageIsScopedToExecution)$' -count=3
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -parallel 4 ./pkg/agent -skip '^Test(CodexCancellationBeforeTurnStartedStopsProcessImmediately|CodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline)$' -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/daemon -run '^Test(ReconcileFreshRetryResult|ShouldRetryWithFreshSession.*)$' -count=1
go vet ./pkg/agent ./internal/daemon
```

Passed: all 16 new scenarios × 3 with the race detector; all 165 selected daemon retry scenarios; `go vet`; and the full agent package with `-parallel 4`, excluding only the two named Codex timing assertions.

The first unrestricted agent suite failed those two Codex cancellation-duration assertions plus three unrelated provider startup/deadline tests. The two excluded assertions also fail independently on unchanged main (about 2 seconds of cleanup beyond their thresholds). The other three pass in the final suite; no tests or production code for those providers were changed. The full final suite completed in 164.730s.

The 16 new scenarios cover split/missing/zero-result usage, interleaved response IDs, distinct models, cache-only responses, delayed usage/model fields, absent IDs, nested events, output placeholders, success/error final totals, and concurrent executions on one backend. Every fixture asserts that all text and tool blocks still arrive. No real Claude call was made.

## Risks and scope

- This corrects reported usage for new executions. It does not change model requests or supplier charges, and does not rewrite historical rows.
- Streams without response IDs retain best-effort per-event input/cache accounting; there is no safe identity with which to deduplicate them.
- Without final usage, output is reported as zero because the true count is unavailable. Nested-agent totals remain unavailable on that fallback. These are partial measurements, not proof of zero consumption.
- No database, API, scheduling, or session-reset changes.

## Checklist

- [x] Traced the reported behavior to the stream protocol and current adapter.
- [x] Added a failing-before/passing-after dynamic regression.
- [x] Updated the affected runtime reference and documented limitations.
- [x] Considered adjacent PR scope and preserved text/tool delivery.

## AI Disclosure

**AI tool used:** Codex.

**Prompt / approach:** Asked to verify an apparent Claude usage accounting bug and, if confirmed, fix it and submit a PR. Read the current adapter and official usage protocol, reproduced split-response accounting with a fake CLI, added failure/control/concurrency cases, and made a narrow collector change.


Bohan-J:
Thanks for this — it's a well-made fix, and the write-up made it easy to verify.

I checked the three premises independently rather than taking them on trust:

- **The regression tests are real tests.** I reverted the production change on your branch (restored per-event accumulation, including `output_tokens`) and 12 of the new subtests fail, while the three terminal-usage control cases pass either way. They're pinning the behavior, not asserting it into existence.
- **The approach matches the documented protocol.** The Agent SDK cost-tracking guide prescribes exactly this: deduplicate by `message.id`, treat per-step `output_tokens` as a `message_start` placeholder, and — in the crash-recovery section — sum assistant `usage` counting each response once while skipping events with a non-empty `parent_tool_use_id`, which recovers the main loop's input and cache tokens only. Your implementation lines up with the documented example almost line for line.
- **Tests pass locally.** `go build ./...`, `go vet ./pkg/agent`, and `go test -race -parallel 4 ./pkg/agent` all pass; the one failure I saw (`TestCodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline`) also fails on unchanged `main`, so it's a pre-existing timing flake on my machine and unrelated to this change.

I pushed one small follow-up commit (5a3ca43) for a docs nit: the new fallback note in `references/runtimes.md` had landed between the CLI command block and the warning about commands that affect active execution, splitting that pair. I moved it down with the other per-command notes, in the same order the command block lists them, and opened it by naming `runtime usage`. No wording changes beyond the added lead-in sentence.

Two things worth recording for later, neither blocking:

1. On the fallback path this trades an over-count for an under-count — output is now 0 and subagent totals are dropped — and the UI multiplies output by a real per-token rate with no indication that the figure is partial. Your PR description is explicit that these are partial measurements, which I think is the right call; the durable fix is to tag usage rows with their source (final `modelUsage` / final `usage` / assistant fallback / none) so the frontend can distinguish "measured" from "partially recovered". That would also let us answer how often this path is actually hit, which we currently can't measure.
2. `server/pkg/agent/codebuddy.go` has the same un-deduplicated accumulation over the same stream-json shape. I'm opening a separate PR for it.

Merging. Thanks again.

PR #8384, opened 2026-09-14
feat(statusnotify): workspace-scoped issue status notifications to one IM bot

### LINKED ISSUE #8380: [Feature]: Workspace-wide issue status notifications to one IM bot
### Deployment type

self-host

### Area

Chat integration — Feishu / Lark or Slack

### What do you want and why?

**One bot should be able to broadcast issue status changes for a whole workspace.**

We run a self-hosted Multica instance where a team of agents works issues
autonomously. The two transitions a human must react to are `blocked` and
`in_review` — everything else can wait for someone to open the board.

Today there is no way to get those into an IM channel. The existing Feishu
integration is a per-agent conversational binding: one bot ↔ one agent,
two-way chat. That is a different shape from what a status feed needs.

**What the code says**

`server/internal/integrations/lark/outbound.go`

```go
func (p *Patcher) Register(bus *events.Bus) {
    bus.Subscribe(protocol.EventTaskFailed, p.handleEvent)
    bus.Subscribe(protocol.EventChatDone, p.handleEvent)
    bus.Subscribe(protocol.EventTaskCancelled, p.handleEvent)
}
```

No issue event is subscribed. And `processEvent` drops anything without a
chat session:

```go
if !chatSessionID.Valid {
    // Issue / autopilot tasks have no chat_session.
    return nil
}
```

Issue-driven work has no chat session by definition, so it can never reach
this path.

The same holds across every IM adapter. Subscriptions by integration:

| Integration | Events subscribed |
|---|---|
| lark | `task:failed`, `chat:done`, `task:cancelled`, `lark_installation:created` |
| slack | `task:failed`, `chat:done`, `task:cancelled` |
| dingtalk | `task:failed`, `chat:done` |
| telegram | `task:failed`, `chat:done`, `task:cancelled`, `task:message` |
| wecom | `task:failed`, `chat:done`, `task:cancelled`, **`inbox:new`** |

`wecom` is the only one that reaches beyond task lifecycle, and `inbox:new`
is per-person delivery, not a workspace feed.

The database shape reinforces it — `channel_installation` is unique on
`(workspace_id, agent_id, channel_type)` and `lark_installation` on
`(workspace_id, agent_id)`. A row cannot exist without an agent, so
"one bot, whole workspace" has nowhere to live.

**The event already exists**

`protocol.EventIssueUpdated` is published on every issue update and already
carries `status_changed` plus the full `IssueResponse`. `notification_listeners.go`
consumes it for in-app notifications:

```go
bus.Subscribe(protocol.EventIssueUpdated, func(e events.Event) {
    statusChanged, _ := payload["status_changed"].(bool)
    ...
})
```

So the signal and the change-detection are both in place. What is missing is
a path from that event to an IM channel.

A workspace-scoped status notification layer, sitting above the existing
`channel` abstraction rather than inside any one adapter.

The `channel` package already states this as its design intent:

> Adding a platform is "register a factory here", never "edit the core".

Concretely:

1. **A workspace-scoped subscription** — target (chat id), which status
   categories to notify on, which channel type. Not keyed on `agent_id`,
   which is what makes it a workspace feed rather than a per-agent binding.

2. **A listener on `EventIssueUpdated`** that filters by `status_changed` +
   status category and renders through `channel.Channel.Send`. Adapters
   already declare `CapRichCard` / `CapText`, so the renderer can degrade
   per platform without branching in the core.

3. **Feishu implementation** on top of that layer, since it is the adapter
   with a complete `Send` path today.

Filtering on `status_category` rather than `status` matters: a custom status
can have an opaque key (`in_review_2`), but its category is one of the known
set.

We built this out-of-tree first to check the shape holds, and it does — one
bot, whole workspace, `blocked` + `in_review`, with a link back to the issue.

Volume on our instance: 108 `blocked` and 97 `in_review` transitions over 7
days in a single workspace, roughly 1.2 notifications/hour. Low enough that
no batching layer is needed.

Implementation notes from that build, in case they are useful:

- Plugin hooks were an option (`issue.status_changed` is dispatchable), but
  the callback has to be an HTTP endpoint outside the process, so every
  deployment that wants IM notifications has to run and secure another
  service. An in-tree listener has no such requirement.
- `MULTICA_PLUGIN_DEV_ORIGINS` is what makes a private-network callback legal
  at all, and its own doc comment scopes it to plugin authors —
  "Empty in every deployment that has not opted in". Leaning on it for a
  standing production feature reads as a misuse of a development affordance.

Happy to put up a PR for this. Asking first because the shape — where the
subscription lives, whether it belongs in `channel` or beside the existing
notification listeners — is a design call that is yours to make.

xiaoyaner0201:
## What does this PR do?

Adds a workspace-scoped path from `EventIssueUpdated` to an IM channel, so one
bot can carry the issue transitions a team must react to — `blocked`,
`in_review` — for a whole workspace.

**The problem**

Issue-driven work cannot reach any IM integration today. Subscriptions by
adapter:

| Integration | Events subscribed |
|---|---|
| lark | `task:failed`, `chat:done`, `task:cancelled`, `lark_installation:created` |
| slack | `task:failed`, `chat:done`, `task:cancelled` |
| dingtalk | `task:failed`, `chat:done` |
| telegram | `task:failed`, `chat:done`, `task:cancelled`, `task:message` |
| wecom | `task:failed`, `chat:done`, `task:cancelled`, `inbox:new` |

All task lifecycle. And `lark/outbound.go` drops anything without a chat
session:

```go
if !chatSessionID.Valid {
    // Issue / autopilot tasks have no chat_session.
    return nil
}
```

Issue work has no chat session by definition, so the existing path is closed
to it structurally, not by omission.

**Why this approach**

The signal already exists — `EventIssueUpdated` carries `status_changed` plus
the full `IssueResponse`, and `notification_listeners.go` already consumes it
for in-app notifications. Nothing new needs to be published; what was missing
is a consumer that forwards to a channel.

Delivery goes through the existing `channel.Channel` contract rather than into
any one adapter, following that package's stated rule:

> Adding a platform is "register a factory here", never "edit the core".

So Feishu works today and the other adapters need no changes to participate.

## Related Issue

Closes #8380

## Type of Change

## Changes Made

- `server/internal/statusnotify/notifier.go` — `Subscription` (workspace-scoped,
  **not** keyed on `agent_id`), `Store` / `Renderer` interfaces, and `Notifier`
  fan-out.
- `server/internal/statusnotify/extract.go` — pulls a `Change` out of an
  `EventIssueUpdated` payload, reading the same fields
  `notification_listeners.go` reads rather than inventing a second contract for
  the same event.
- `server/internal/statusnotify/render.go` — `TextRenderer`; category constants
  alias `internal/issuestatus` instead of redeclaring the literals.
- `server/internal/statusnotify/notifier_test.go` — 8 tests.

Three decisions worth flagging for review:

**1. `Subscription` is not keyed on `agent_id`.** That absence is the whole
point — it is what makes this a workspace feed rather than another per-agent
binding, and it is why `channel_installation` (unique on
`workspace_id, agent_id, channel_type`) cannot express it. No migration is
included: where these rows live is a schema decision I would rather take
direction on than assume.

**2. Filtering is on `status_category`, not `status`.** A custom status carries
an opaque key (`in_review_2`) but always reports a known category. The blank
fallback to `Status` follows that field's own doc comment — without it a custom
status would match no subscription and the notification would vanish silently.

**3. `TextRenderer` emits text, not a card.** `OutboundMessage` carries only
`Text` today, so a card renderer would have nowhere to put its payload. The
`Capability` argument is still threaded through, so a richer body can be added
when `OutboundMessage` grows a card field without changing the interface.

## How to Test

```
cd server
go test ./internal/statusnotify/
go build ./...
```

The tests pin behavior rather than declarations. Each was verified to fail when
the behavior it describes is removed — two examples:

- making empty `Categories` mean "everything" →
  `TestEmptyCategoriesDeliversNothing` fails with
  `empty Categories delivered 1 messages, want 0`
- returning on the first delivery error instead of continuing →
  `TestOneFailingTargetDoesNotStopTheRest` fails with
  `healthy target got 0 deliveries, want 1`

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — no UI change
- [ ] I have updated relevant documentation to reflect my changes — holding until the wiring question below is settled

**Not included, deliberately:** the schema for persisting subscriptions, the
wiring in `cmd/server`, and a settings UI. Those depend on where you want
subscriptions to live, and guessing would mean a migration you then have to
undo. Happy to add all three in this PR once you point at the shape you want.

## Risks

- No behavior changes for existing deployments: nothing subscribes to the new
  listener until a `Store` is wired in.
- `Notify` returns an error describing partial failure but does not retry.
  Delivery retry semantics belong with whatever calls it, and inventing a retry
  policy here would conflict with the one the caller already has.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** We hit this while running a self-hosted instance where
agents work issues autonomously — `blocked` and `in_review` were the two
transitions a human had to notice, and neither reached our chat tool. I first
built a standalone service against the Plugin hook API to check the shape held,
then traced why the in-tree adapters could not do it. The subscription table
in the issue, the `chat_session` early return, and the `channel` registry's
"register a factory, never edit the core" comment all came from reading the
source rather than from documentation. The design decisions flagged above —
especially dropping `agent_id` from the subscription key and filtering on
category instead of status — came out of that reading; the AI work was in
tracing the call paths and in pinning each behavior with a test that was then
verified to fail without it.


xiaoyaner0201:
## Reference implementation, now running in production

The external plugin that motivated this PR is now open source:

**https://github.com/xiaoyaner0201/multica-lark-notifier**

It has been delivering real `blocked` / `in_review` notifications on our
self-hosted instance since today, so the shape proposed in this PR is not
speculative — it mirrors what the external service ended up needing.

### How the two relate

The plugin proves the *behaviour* is wanted; this PR proposes making it a
first-class platform capability so every IM adapter gets it, rather than each
deployment running a sidecar.

Concretely, the plugin has to do three things that `statusnotify` in this PR
does natively:

| Plugin does externally | This PR |
|---|---|
| HMAC-verify every delivery | in-process, no signature needed |
| Filter on `status_category` | `Subscription.Categories` |
| Resolve who to notify by querying `/subscribers` | already inside the workspace |

The third one is where an external plugin is genuinely worse: it has no access
to `issue_subscriber`, so per-recipient routing means an extra API round-trip
per event, plus an identity mapping from Multica user → IM account that the
platform already has for connected channels.

### One design point worth confirming

This PR filters on `status_category` rather than `status`, on the basis that
custom status keys are opaque handles (`in_review_2`) while categories are a
known set. Building the plugin confirmed that was the right call — but it also
surfaced that `status_category` carries `omitempty`, so a consumer must fall
back to `Status` when it is absent, per that field's own doc comment. The
`extract` helper here does that; flagging it because it is easy to miss and
fails silently (custom statuses just never match any subscription).

Still happy to adjust scope — the subscription schema, `cmd/server` wiring, and
settings UI are deliberately left out of this PR pending direction.

PR #8390, opened 2026-09-14
MUL-7359 perf(runtime): avoid echoing final comment bodies

importcpp:
## What does this PR do?

Use the existing `--output table` mode in the runtime's final-comment examples so agents receive a success confirmation without a second copy of the body they just wrote. In the regression fixture, the old example echoes 69,778 bytes of JSON stdout; the new one produces no stdout and retains the 61-byte stderr confirmation. Both submit the same complete multilingual comment and parent ID.

Because `--output table` makes a successful and a failed post look identical on stdout, the rendered snippet also has to keep the post's exit status readable. Cleanup is therefore gated on the post succeeding — `&&` on POSIX, a `$LASTEXITCODE` check on Windows, where PowerShell 5.1 has no `&&`. A failed post now reliably exits non-zero and leaves `./reply.md` on disk for the retry.

The shared formatting rule covers assignment turns and multi-thread replies. It preserves `--output json` for callers that need the returned comment ID, attachment details, or other response fields. The CLI default stays JSON.

## Related Issue

Found during a worker/leader token-overhead assessment; no separate GitHub issue. Independent of #8339's completion delivery and #8382's accounting. #8377 touches the same runtime files in different functions; its shared production-file patch applies with this change. The PRs were not tested together.

## Changes Made

- Add the output flag to the Unix and Windows reply examples and document the receipt-mode choice in the shared brief's `## Comment Formatting`, after the file-first guardrail rather than ahead of it.
- Gate cleanup on the post succeeding in both rendered examples, and state in the brief why an unconditional cleanup command hides a failed post.
- Execute the rendered command's flags through the real CLI handler against a local HTTP fixture, checking complete content, reply routing, explicit JSON, and HTTP rejection.
- Add a shell-level regression test that executes the rendered snippet against a test-created stub CLI and asserts a failed post exits non-zero and preserves the body file, while a successful post exits 0 and cleans up. It runs under `sh` and, where PowerShell is installed, under `pwsh`.
- Update existing cross-platform instruction assertions to pin the gate itself.

No built-in skill reference changes: the brief fully owns this rule, so `references/issues.md` stays at baseline.

## How to Test

From `server/`:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./cmd/multica -run '^TestCommentReplyReceipt$' -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/daemon/execenv -run '^TestCommentReplySnippetPropagatesPostFailure$' -count=1 -v
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -parallel 4 ./cmd/multica ./internal/daemon/execenv -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/service -run '^Test(Builtin.*|PlatformSkill.*|LegacyRedirects.*)$' -count=1
go test ./internal/daemon/execenv/ -run ByteIdentical -count=1
go vet ./cmd/multica ./internal/daemon/execenv
```

The suggested-receipt success case fails on unchanged main `2295be9271d85928db5f742cda49324b3a510807` because it echoes the full body. `TestCommentReplySnippetPropagatesPostFailure` fails on the earlier two-statement form of this PR, reporting that the cleanup masked the failure and deleted the body; both pass on the final commit. `-v` on the snippet test makes a PowerShell skip visible instead of passing silently.

## Risks and scope

This changes runtime guidance, not CLI defaults, stored comments, notifications, or claim receipts. Empty stdout is not proof of success, which is why the examples gate cleanup on the exit status. Agents needing the response object should use JSON.

Measurements are CLI output bytes, not billed tokens or whole-task savings; no real model was called. The guidance has a small input cost, and short comments may see little net benefit. Shortening worker results remains separate work.

## AI Disclosure

**AI tool used:** Codex.

**Prompt / approach:** Implement the first step of a handoff-token assessment. Reuse the existing CLI output mode, reproduce the echo from the rendered example, and limit the change to posting guidance, documentation, and regression coverage.

## Review follow-ups

Three follow-up commits were pushed to this branch by a maintainer agent during review, with the author's `maintainerCanModify` enabled: the guardrail-ordering fix, the skill-reference removal, and the exit-status gate plus its regression test. The original change and its approach are the author's.


Bohan-J:
Thanks for this, @importcpp — merging.

Two things we particularly appreciated. First, you reused the `--output` mode the CLI already had instead of adding a `--quiet` flag or changing the CLI default, which would have broken scripted callers for no extra benefit. Second, the regression test drives the *rendered* instruction through the real handler rather than asserting on a hand-written command string, so the test fails if the guidance and the CLI ever drift apart. That is the right place to put the assertion, and it caught its own regression during review.

We also valued the honesty in the PR description about what the numbers do and do not show — output bytes, not billed tokens. That made the review much faster.

During review we pushed three follow-up commits to your branch:

- `31aadc4bf` moves the receipt rule below the file-first guardrail, so the section still opens with its correctness rule rather than an output-size tip.
- `9faa12d2f` drops the `references/issues.md` copy: the brief now owns this rule in full, and the skill reference is scoped to contracts the brief does not encode.
- `29095d54b` gates cleanup on the post succeeding (`&&` on POSIX, `$LASTEXITCODE` on Windows). With `--output table` a failed and a successful post both write nothing to stdout, so the exit status is the only machine-readable signal — and an unconditional `rm` on the next line succeeded after a failed post, making the whole shell call exit 0 and deleting the body. It ships with a shell-level test that executes the rendered snippet.

That last one was a consequence of the change rather than a flaw in your approach, and it is the kind of thing that only shows up once the stdout signal is removed. Nice work.

PR #8416, opened 2026-09-14
MUL-7374: fix(codex): avoid double-counting reasoning output tokens

importcpp:
## What does this PR do?

Fix Codex output-token overcounting in Multica. When Codex reports `output_tokens=10` with `reasoning_output_tokens=5`, Multica currently records 15 output tokens; this change records 10 in both the app-server event path and the rollout JSONL fallback. This corrects recorded usage and usage-based cost estimates; it does not reduce model calls or the provider's actual token consumption.

Codex preserves the inclusive output total and its reasoning breakdown separately in its [Responses adapter](https://github.com/openai/codex/blob/e9633d7a0226eac91c7a791dc4f92cf8f25df2ae/codex-rs/codex-api/src/sse/responses.rs#L138) and [app-server protocol mapping](https://github.com/openai/codex/blob/e9633d7a0226eac91c7a791dc4f92cf8f25df2ae/codex-rs/app-server-protocol/src/protocol/v2/thread.rs#L1913). Adding them again double-counts reasoning. The same mapping is present in [rust-v0.144.6](https://github.com/openai/codex/blob/rust-v0.144.6/codex-rs/codex-api/src/sse/responses.rs#L131).

## Related Issue

No linked issue. Independent of #8402, which tracks Claude usage coverage rather than Codex counter normalization.

## Type of Change

## Changes Made

- Document the existing shared `TokenUsage` contract in `server/pkg/agent/agent.go`: cache buckets are separate from plain input, and output includes reasoning/thinking. Zero counters alone do not prove complete reporting.
- Add one private Codex conversion function in `server/pkg/agent/codex.go`, used by both event and rollout readers. Preserve current cache normalization, resume baselines, and duplicate-snapshot handling.
- Add 16 cross-provider contract cases in `server/pkg/agent/usage_contract_test.go`, covering both Codex sources and both Claude result formats with mixed cache/reasoning, no reasoning, entirely reasoning output, and cache-only usage.
- Correct three existing Codex expectations that encoded the overcount. Add thinking content to existing Claude fallback fixtures to verify that final totals, replacement, and message deduplication retain their current behavior. Claude production parsing is unchanged.

### Risks and scope

New Codex runs with reasoning will report lower output totals than the buggy calculation. Previously persisted rows are not rewritten or repriced. No schema, API, prompt, or session-lifecycle changes are included; this does not add usage-completeness tracking or a new cross-provider parser.

## How to Test

All tests use local fixtures and the repository's agent CLI guard; no real agent/model invocation was used.

From `server/`:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test ./pkg/agent -run '^(TestProviderTokenUsageContract|TestParseCodexSessionFileSubtractsCachedInput|TestScanCodexSessionUsageSubtractsResumeBaseline|TestCodexThreadTokenUsageUpdatedAccumulatesCurrentTurnResponses|TestClaudeExecuteFallbackUsage)$' -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -p 2 -parallel 2 ./pkg/agent/... -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race -p 2 -parallel 2 ./internal/daemon -run '^(TestMergeUsage|TestReconcileFreshRetryResult)$' -count=1
go vet ./pkg/agent/...
```

- Red/green proof: the targeted suite had 7 failing leaf cases before the production fix (4 new Codex cases and 3 corrected existing assertions), then 33 passing leaf cases after it.
- Daemon merge/retry checks: 6 passing leaf cases under `-race`; `go vet` and `git diff --check` pass.
- Full agent race suite on macOS arm64 / Go 1.26.2: 1,752 passing leaf cases, 3 platform-dependent skips, and 2 failures; no race warnings. The failures are `TestCodexCancellationBeforeTurnStartedStopsProcessImmediately` and `TestCodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline`, both taking about 3 seconds against a 1-second cleanup assertion. Both reproduce with the same failure on unchanged main `9d18186e6e8cfa168d6053d33d652e86fadfc12b` under the same race-test command; this PR does not change process cancellation. The skipped cases require Linux signal semantics, `/proc`, or root.

## Checklist

- [x] Targeted tests pass; broader-suite limitations are documented above
- [x] I have updated the relevant in-code usage contract documentation
- [x] I will address reviewer comments before requesting merge

UI screenshots, Chinese product copy, and new-runtime landing/docs updates do not apply to this backend accounting fix.

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Compare Codex and Claude token accounting against provider contracts, reproduce the overcount locally before changing production code, reuse the existing shared usage type, and validate both Codex conversion paths plus Claude controls.


Bohan-J:
Thanks for this — really nice catch, and thank you for the thorough write-up and the red/green proof.

We verified the underlying contract independently against a live `codex app-server` session (codex-cli 0.144.6, `gpt-5.6-terra`, reasoning high) rather than just the docs. On a turn whose visible answer was `972336 6778840`, the provider reported `outputTokens: 2315` / `reasoningOutputTokens: 2303` — the 12-token remainder is exactly the visible answer, and the provider's own `totalTokens` (23108) equals `inputTokens + outputTokens`, with reasoning not added on top. Feeding that same real payload through the code: `main` records 4618 output tokens and the four buckets sum to 25411, while this PR records 2315 and the buckets sum to exactly 23108.

We also scanned 22,744 historical `token_count` records spanning codex-cli 0.93.0 through 0.144.6: 100% satisfy `total == input + output`, none satisfy the exclusive form, and no record ever has reasoning exceeding output. So this isn't version-specific, and the fix is safe across the versions we can observe.

Local test run matches what you documented. The one failure in the full `-race` agent suite (`TestCodexNonResponsiveInterruptStopsProcessAtConfiguredDeadline`) reproduces identically on unchanged `main` — machine-timing, unrelated to this change.

Two non-blocking notes, no action needed for this PR:

- The `TokenUsage` doc comment reads as a statement about every adapter, but the OpenCode/CodeArts readers deliberately don't fold reasoning into output today (see their own comments in `opencode.go` / `codearts.go`). Treating that line as the normalization *target* would be slightly more accurate.
- `ReasoningOutputTokens` is still parsed and carried through the delta math but no longer has a consumer. Worth keeping for protocol fidelity and a possible future breakdown.

We're not backfilling previously recorded rows — the reasoning breakdown was never persisted, so the over-count isn't recoverable from stored data, and re-deriving it isn't worth it. New Codex runs will be correct from here.

Merging. Thanks again!

PR #8419, opened 2026-09-14
MUL-7376: fix(openclaw): preserve agent identity during model overrides

alithia-dev:
## What does this PR do?

Adds a first-class per-run OpenClaw model override without changing the selected OpenClaw agent identity. This lets Multica keep unified GRIT as `main` while intentionally starting a run on an approved model preset, avoiding the separate state and memory fragmentation caused by creating model-specific OpenClaw agents.

## Related Issue

Tracked operationally in the private Multica work ledger.

## Type of Change

- [x] Documentation update

## Changes Made

- Add `runtime_config.model_override` for OpenClaw-backed Multica agents.
- Pass the managed override to OpenClaw's native `--model` option while preserving `agent.model` as the OpenClaw `--agent` identity.
- Continue rejecting `--model` from arbitrary custom arguments so the override remains governed.
- Update CLI guidance and self-hosting documentation.
- Add coverage for runtime-config decoding and OpenClaw argument construction.

## How to Test

1. Run `go test ./pkg/agent -run 'Openclaw' -count=1`.
2. Run `go test ./internal/daemon -run 'Openclaw' -count=1`.
3. Run `go test ./cmd/multica -run '^$' -count=1`.
4. Configure an OpenClaw agent with `model: main` and `runtime_config.model_override` set to a supported model or preset, then verify the invocation contains both `--agent main` and the managed `--model` override.

## Checklist

- [x] I have run focused tests locally and they pass
- [x] This change does not affect the UI

Risk is limited to OpenClaw runtime argument construction. The existing agent selection remains unchanged, the new field is optional, and arbitrary custom `--model` injection remains blocked.

The broader Windows package suites still contain unrelated environment-dependent baseline failures; the focused OpenClaw adapter, daemon configuration, and CLI compilation checks pass.

## AI Disclosure

**Prompt / approach:** Verified the installed OpenClaw CLI's native separation of agent identity and per-run model selection, traced Multica's adapter path, implemented the smallest first-class configuration field, and added focused regression tests and documentation.

## Screenshots (optional)

Not applicable; this is a runtime adapter change.


multica-eve (changes requested):
Requesting changes before merge. Preserving `--agent main` while applying a separate `--model` is the right underlying OpenClaw invocation, but the current integration does not yet deliver the advertised one-run behavior safely:

1. `model_override` is stored in the agent's persisted `runtime_config`, and `runTask` reads that shared value when each task starts. That makes the override agent-wide for every task claimed while the setting is present, not scoped to one run. Concurrent tasks can therefore inherit an override intended for a different run. Please carry the override on the task/run input (or otherwise make the scope atomic and concurrency-safe) instead of relying on mutable agent configuration.

2. The existing OpenClaw Runtime Config UI does not know about this field. `packages/core/agents/openclaw-runtime-config.ts` parses and serializes only `mode` and `gateway`, while `packages/views/agents/components/tabs/runtime-config-tab.tsx` saves the serialized object. Opening and saving gateway settings will silently remove an existing `model_override`, and there is no UI path to configure the new behavior. Please update the typed schema, form, serializer/equality logic, and tests—or preserve unknown keys if this field is intentionally managed elsewhere.

3. Usage attribution falls back from `meta.agentMeta.model` to `opts.Model`, but for OpenClaw `opts.Model` is the registered agent ID (for example, `main`). With an explicit override and runtime output that omits the model, usage will be recorded under the agent ID instead of the selected model. The fallback should be runtime-reported model → `OpenclawModelOverride` → `unknown`; it should not use the OpenClaw agent identity as a model key.

The backend-focused tests pass, but these product-scope, configuration-round-trip, and accounting cases need coverage before the change is safe to merge.

PR #8423, opened 2026-09-15
MUL-7382 docs(agent): clarify token usage normalization contract

importcpp:
## What does this PR do?

Clarify the token usage contract documented in #8416: adapters must normalize provider usage into four mutually exclusive buckets, output includes reasoning/thinking, and a breakdown already included in a total must not be added again. Keep the shared type comment normative, with provider-specific implementation details beside the relevant reader.

## Related Issue

MUL-7382. Follow-up to the two documentation notes in [the #8416 review](https://github.com/multica-ai/multica/pull/8416#issuecomment-5668023387).

## Type of Change

- [x] Documentation update

## Changes Made

- `server/pkg/agent/agent.go`: preserve the mutually exclusive token-count invariant and clarify that breakdowns already included in a total must not be added again.
- `server/pkg/agent/codex.go`: explain that the raw reasoning counter mirrors the provider payload and participates in symmetric resume-baseline subtraction; no downstream consumer uses that breakdown, and normalized usage reads the inclusive output total.
- `server/pkg/agent/usage_contract_test.go`: state that every adapter must follow the shared contract while identifying Codex and Claude as the providers covered by these cases.

All changes are ordinary Go comments. Token arithmetic, runtime behavior, fields, JSON tags, and test assertions are unchanged.

## How to Test

- Checked the comments against the Codex raw-counter, resume-baseline subtraction, and normalization paths and the existing contract-test cases.
- Verified both the full PR diff and the review follow-up contain only ordinary Go comment changes; executable text is identical.
- `git diff --check` passes; `gofmt -l` reports no files.
- Code tests were not run for this comments-only update.

## Checklist

- [x] Documentation checks pass; code tests were not run for this comments-only update
- [x] I have considered and documented scope above
- [x] I will address reviewer comments before requesting merge

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Check the review against the captured usage fixtures and adapter code, preserve the shared accounting invariants, and document the raw Codex counter's purpose without making unverified claims about other adapters. Verify that the patch changes comments only.


Bohan-J (changes requested):
Thanks for following up on the #8416 notes — the underlying worry is legitimate: the current wording *is* a normative rule that a reader could mistake for a description of every adapter. But as written, this patch states something the repo's own captured fixtures contradict, and the error points a future reader straight back at the double-counting bug #8416 just removed.

This is comments-only — I verified every added/removed line is a `//` line, no code, assertions, fields or JSON tags change. So there is no user-visible behavior change here. The thing being changed is the contract maintainers will rely on when they extend the per-runtime usage adapters, which is exactly why the wording has to be right.

## The blocking problem

`server/pkg/agent/agent.go` now reads:

> The normalization target is four mutually exclusive counts: ... OutputTokens includes reasoning/thinking. **Codex and Claude follow this output convention. OpenCode and CodeArts currently preserve provider output without folding in the separate reasoning counter.**

The structure is "target is X; A and B follow it; C and D currently do something else," and "currently" reads as an outstanding gap. The repo's own captured OpenCode `step_finish` payload says otherwise — `server/pkg/agent/opencode_test.go:407`:

```
"tokens":{"total":14674,"input":14585,"output":89,"reasoning":82,"cache":{"write":0,"read":0}}
```

`14585 + 89 = 14674`, exactly the provider's own `total`, with `reasoning: 82` **not** added on top. Those 82 tokens are already inside the 89 output tokens. DevEco, which speaks the same `step_finish` protocol, matches — `server/pkg/agent/deveco_test.go:150`: `total 11640 = 11637 + 3`.

So OpenCode and CodeArts use the same convention as Codex, and *not* folding reasoning in is precisely what makes them compliant — not a deviation from the contract. The patch labels two already-correct adapters as unfinished work, which inverts its own stated goal.

The concrete failure path: a maintainer reads this comment, goes to `opencode.go`, and "closes the gap" by changing

```go
usage.OutputTokens += t.Output
```

to `+= t.Output + t.Reasoning` — the exact line #8416 deleted from Codex, reintroduced on OpenCode/CodeArts. Nothing in CI catches it, because the comment itself is the only thing that changed here, and after the follow-up change the inflated output would be "documented" behavior.

I don't think this reading came from nowhere. The wording at `opencode.go:498` ("reasoning and the aggregate total in fields of their own ... not components of Input/Output") invites it. But that comment landed in #6545 (MUL-5831) to fix void-step misclassification, and it explicitly scopes itself out of billing. What it established is that `reasoning` is a sibling field that can be positive while input/output are zero — not that OpenCode's `output` excludes reasoning. The `"... only"` cases at `opencode_test.go:1127-1143` are synthetic per-field probes for that guard, not captured traffic.

If the intent was instead "OpenCode's output genuinely excludes reasoning and we are under-recording it," then that is a live under-billing bug and deserves an issue plus a code fix — not a contract comment that records it as an acceptable status quo.

## Requested changes

**1. `agent.go` — the two OpenCode/CodeArts sentences.** Either verify against a live OpenCode/CodeArts session what `reasoning` means relative to `output` and state the verified fact, or drop the per-adapter roll call entirely and keep only the contract. I'd suggest the latter; see the placement note below.

**2. `agent.go` — please don't downgrade the invariant.** `Its four token counts are mutually exclusive` became `The normalization target is four mutually exclusive counts`. Mutual exclusivity (InputTokens excludes cache reads/writes) is a hard invariant every adapter currently upholds: Copilot subtracts the cache tiers at `copilot.go:113`, Codex goes through `codexPlainInputTokens`, Claude's input already excludes them, antigravity maps provider buckets directly. The evidence this PR brings (whether reasoning is folded in) is a different axis — completeness, not overlap. Softening exclusivity into an aspiration tells the next adapter author it is optional. Please keep that sentence normative.

**Placement suggestion:** keep the type comment purely normative (four mutually exclusive buckets; output includes reasoning; a breakdown already inside a total must not be added to it again), and put any "what adapter X does today" note next to that adapter's own usage reader. There are a dozen-plus backends under `server/pkg/agent/`; naming four of them in the shared type doc creates a list that is already incomplete on day one — antigravity (`antigravity.go:87`) and copilot (`copilot.go:515`) carry exactly that kind of local note and are already compliant, yet aren't on it. The list will rot the next time an adapter is added or fixed.

## Nits (non-blocking)

- `codex.go:3830`: the new comment says `codexRawTokenUsage` keeps the provider's reasoning breakdown "in raw snapshots and deltas." In practice `ReasoningOutputTokens` is parsed at `codex.go:3578`, carried through `subtractCodexRawTokenUsage` / `addCodexRawTokenUsage`, and then read by nobody — `codexTokenUsage` (`codex.go:3842`) ignores it. As written it suggests some downstream consumer uses the breakdown. More accurate: it is parsed only to keep the raw snapshot faithful to the provider payload and the resume-baseline subtraction symmetric; no consumer reads it.
- `usage_contract_test.go`: scoping the fact to Codex and Claude is right, but the rewrite drops the normative half of the original ("their normalized usage **must** share the TokenUsage contract"). Worth keeping one sentence saying every adapter must normalize into these buckets — otherwise the file reads as a two-provider regression test rather than a contract test.

## What I verified

- Confirmed the diff is comments-only (no code, assertion, field or tag changes).
- On `main` (`7d9613bd2`): `go test ./pkg/agent/ -run 'TestProviderTokenUsageContract|TestOpenCode.*Usage|TestParseCodexSessionFileSubtractsCachedInput' -count=1` — passes.
- The OpenCode/CodeArts conclusion rests on the in-repo captured fixtures cited above; I did not run a live OpenCode session. If you have measurements that contradict this reading, they win — but then the conclusion becomes "this is an under-billing bug," and it still shouldn't be recorded in the contract comment with the current wording.

One unverified lead, out of scope for this PR: if you do want to document a real deviation, the place to look is OpenCode's **input**, not its output. `opencode.go:437-441` does `usage.InputTokens += t.Input` and counts `Cache.Read/Write` separately, with no subtraction like Copilot's. If OpenCode's `input` includes the cache tiers, that would be an actual exclusivity violation. Every in-repo fixture has zero cache, so it isn't observable from here.

Bohan-J (approved):
Thanks for the quick and careful turnaround — `ecd2b834` resolves all four points, and the way you resolved the first one is the right call.

- **Blocking 1** — the OpenCode/CodeArts sentences are gone entirely rather than reworded. That removes the roll call that would have been incomplete on day one and, more importantly, the reading that would have pointed someone at `usage.OutputTokens += t.Output + t.Reasoning`.
- **Blocking 2** — `Its four token counts are mutually exclusive` is back verbatim, so the invariant reads as a rule again.
- **Nit** — the `codexRawTokenUsage` comment is clearer than what I suggested; stating outright that no downstream consumer reads the breakdown saves the next reader a grep. I re-verified it: `ReasoningOutputTokens` is parsed at `codex.go:3578`, carried through `subtractCodexRawTokenUsage` and `addCodexRawTokenUsage`, and read nowhere else.
- **Nit** — `usage_contract_test.go` now leads with the normative sentence and then scopes the coverage claim honestly to Codex and Claude. That keeps it a contract test rather than a two-provider regression test.

What's left is small but each edit earns its place. The one I value most is `Provider-specific breakdowns must not be added to those totals again` → `Breakdowns already included in a total must not be added to that total again`: the old sentence banned a whole category, the new one states the actual condition. That was the real fix hiding in the original patch.

Verified on `ecd2b834`: diff is comments-only (no code, assertions, fields or tags), `gofmt -l` clean, `go vet ./pkg/agent/` clean, and `go test ./pkg/agent/ -run 'TestProviderTokenUsageContract|TestOpencode|TestParseCodexSessionFile|TestCodexThreadTokenUsage|TestScanCodexSessionUsage|TestClaudeExecuteFallbackUsage' -count=1` passes. CI is green.

Approving and merging. Thanks again for chasing down the follow-up notes from #8416 — and for taking the "delete it" option instead of defending the wording.

PR #8424, opened 2026-09-15
fix(agents): MUL-7383 exclude cancelled runs from success metrics

importcpp:
## What does this PR do?

Fixes cancelled agent runs being reported as successful. With 1 completed run, 1 failed run and 8 cancelled runs, both agent summaries previously displayed **90%** success. They now display **50%**, retain **10 runs** as total activity, and show **8 cancelled** separately. A window containing only cancellations displays **—** instead of 100%.

### Thinking path

Cancellation correctly persists `status = cancelled` and `completed_at`. The activity query previously exposed only total and failed counts, and both UI summaries plus the shared sparkline inferred success from `total - failed`. That complement includes cancellations. This change exposes explicit outcomes from the existing aggregation and computes `completed / (completed + failed)` once in the shared core summary.

## Related Issue

Found through a local agent statistics audit; no linked issue. Independent of the usage accounting PRs.

## Type of Change

- [x] Bug fix
- [x] Tests

## Changes Made

- Add `completed_count` and `cancelled_count` to the existing activity response; regenerate sqlc. Keep existing totals, time window and visibility filters.
- Validate the response at the API boundary. Missing, malformed or contradictory new outcome counts stay unknown instead of becoming measured zeros.
- Share the success-rate calculation between Overview and Activity. Show cancellations separately and render cancelled/unknown activity in a neutral color in the shared sparkline.
- Add keyboard-accessible formula help and update English, Simplified Chinese, Japanese and Korean copy.

## How to Test

1. With a migrated local test database, run from `server/`:
   ```sh
   ../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/handler -run '^Test(AgentActivityOutcomes|CancelTaskByUser_)' -count=1 -v
   ```
   **19 top-level tests passed.** The new regression exercises seven queued cancellations and one running cancellation through `CancelTaskByUser`, then reads `GetWorkspaceAgentActivity30d`: 8/0/0/8 total/completed/failed/cancelled, followed by 10/1/1/8 after completed and failed runs. Old outcomes and an active run do not change the counts. The same regression fails against baseline handler/generated-query code at `cf52ba33` because the outcome fields are absent.
2. From the repository root:
   ```sh
   pnpm --filter @multica/core test agents/use-agent-activity.test.ts api/agent-activity.test.ts
   pnpm --filter @multica/views test agents/components/tabs/activity-tab.render.test.tsx agents/components/tabs/activity-tab.test.ts agents/components/sparkline.test.tsx agents/components/agents-page.test.tsx
   ```
   **45 tests passed**, covering mixed outcomes, all cancelled, all failed, all completed, empty windows, legacy responses, malformed counts and sparkline proportions.
3. Core/views typechecks, ESLint on changed TypeScript files, `go vet ./internal/handler`, and `git diff --check` passed.
4. Chromium smoke checks mounted the actual shared components with local aggregate fixtures: before/after, all cancelled, legacy response, Chinese at 390px without horizontal overflow, and keyboard access to the formula help. This is component-level browser validation; it does not claim a live agent or authenticated shared-backend end-to-end run.

## Risks and scope

Historical success percentages change intentionally when a window contains cancellations. A new frontend connected to an older backend displays an unavailable success rate until explicit completed counts are available. Total runs and failed counts remain visible. The endpoint adds fields; no database migration, extra query or new public endpoint is needed. Cancellation execution, billing and duration calculations are unchanged.

## Checklist

- [x] Thinking path, reproduction and risks included
- [x] Relevant local tests pass; regression tests added
- [x] Before/after screenshots included
- [x] Inline metric documentation and all four locales updated
- [x] Chinese copy checked against the developer conventions (`run` → 运行)
- [x] Landing-copy update not applicable: no runtime, coding tool or UI tab added
- [x] Reviewer comments will be addressed before requesting merge

## AI Disclosure

**Prompt / approach:** Investigate cancelled runs counted as successful; reproduce through the real cancellation and activity handlers, expose explicit outcome counts, centralize the success-rate formula, and verify shared UI behavior and older-backend compatibility.

## Screenshots

The fixture mounts baseline components/locales for “before” and changed shared components for “after”, using the same outcome counts. Evidence assets live on a separate fork branch and are not included in the code diff.

| Before | After |
| --- | --- |
| Before: 90% success with eight cancellations | After: 50% success and eight cancellations |

Additional cases: [all cancelled](https://raw.githubusercontent.com/importcpp/multica/a96d5b32c3e9ec250020370895c3fb9dd384c19e/after-cancelled.png), [older backend](https://raw.githubusercontent.com/importcpp/multica/a96d5b32c3e9ec250020370895c3fb9dd384c19e/after-legacy.png), [Chinese narrow viewport](https://raw.githubusercontent.com/importcpp/multica/a96d5b32c3e9ec250020370895c3fb9dd384c19e/after-zh-narrow.png).


Bohan-J:
Thanks for this — merging.

This was a real statistics bug, and the framing in the PR description made it easy to review: deriving success from `total - failed` quietly counted every cancellation as a success, and the repo had already decided elsewhere that cancellation is a procedural signal rather than an outcome (see the `ListWorkspaceAgentTaskSnapshot` query comment and #1823). Aligning the activity aggregation with that decision — rather than introducing a second definition of "success" — is the right shape for the fix. Exposing the counts from the existing aggregation, with no new query and no migration, keeps the cost where it should be.

Two things I want to call out for anyone reading this later:

- The regression test earns its place. Reverting the server-side files to the base commit makes it fail, so it pins the actual contract rather than restating the implementation.
- I pushed one follow-up commit (`e81358cc`) to your branch before merging. It removes the fallback for a *new client talking to an older back-end* — the optional outcome counts, the null propagation through the bucket series, and the contradiction guard in the schema. That path is genuinely unsupported here: our compatibility direction is only an older daemon or client against a newer back-end, and this repo's guidelines ask us not to carry legacy adapters that aren't needed. Nothing about the actual fix changed — a window of nothing but cancellations still shows `—` instead of claiming 100%, and cancelled runs still render neutral in the sparkline. Sorry for touching your branch directly instead of asking; it was the quickest way to keep everything in one PR, and the change is confined to that one commit if you want to see exactly what moved.

Appreciate the thorough test coverage and the four-locale copy — that saved a round trip.

PR #8433, opened 2026-09-15
MUL-7389: fix(invitation): drop stale pending-invitation rows when an invite concludes elsewhere

### LINKED ISSUE #8432: [Bug]: Pending workspace invitation stays in the sidebar after accepting elsewhere; Join/Decline silently do nothing
### Deployment type

self-host

### Area

Desktop app

### What happened?

A pending workspace invitation can stay in the sidebar's **Pending invitations** section forever after the user accepted (or declined) it from a *different surface* than the one showing the row — e.g. the invitation was accepted through the email link / web invite page (or on another device) while the desktop app was open. The workspace itself appears correctly in the workspace list, but the stale invitation row remains, and clicking **Join** or **Decline** appears to do nothing. The row only disappears after restarting the app.

The buttons genuinely do nothing visible: both requests fail server-side with `400 "invitation is not pending"`, and the sidebar swallows the error — it has no `onError` handling and does not refetch.

A related variant: a user who joins a workspace through a share link while an email invitation is still pending keeps the row too; **Join** then fails with `409 "you are already a member of this workspace"`, also silently.

### Steps to reproduce

1. Sign in on the desktop app as user B.
2. Have user A invite B to a workspace W.
3. Accept the invitation from a surface other than the desktop app session — open the invitation email link / web invite page (or accept on another device).
4. Return to the original desktop app: W appears in the workspace list, but the invitation is still listed under **Pending invitations**.
5. Click **Join** or **Decline** on that row — nothing visibly happens, and the row stays until the app is restarted.

### Root cause

Four pieces combine to freeze the row for the rest of the session:

1. Nothing invalidates the account-level pending list when an invitation concludes. The `invitation:accepted` / `invitation:declined` WS handlers invalidate only the workspace-scoped admin lists (`invitations(wsId)` / `members(wsId)`); `invitation:revoked` is the only one that also refreshes `myInvitations` — `packages/core/realtime/use-realtime-sync.ts`.
2. The global QueryClient defaults are `staleTime: Infinity` with `refetchOnWindowFocus: false` (`packages/core/query-client.ts`), and `myInvitationListOptions` overrides nothing, so the pending list never refetches on its own while the session lives.
3. The sidebar's accept/decline mutations only handle `onSuccess` (`packages/views/layout/app-sidebar.tsx`), so server errors are invisible.
4. The server rejects any non-pending invitation with `400 "invitation is not pending"` (`server/internal/handler/invitation.go`), so already-shipped clients have no way to recover even if the user retries both buttons.

### Suggested fix

- **Server** (heals already-deployed clients with no update): make accept idempotent for an already-accepted invitation while the membership from the first accept still exists (return the existing member, `200`); if the membership is gone (user left the workspace afterwards) keep failing, since leaving was explicit. Make decline of any concluded invitation a no-op `204`.
- **Client**: invalidate `myInvitations` in the `invitation:accepted` / `invitation:declined` WS handlers (matching `invitation:revoked`), and have the sidebar's accept/decline mutations refresh the pending list on failure so a stale row drops instead of sticking.

I have a PR ready with this fix plus DB-backed Go tests, a realtime test and a sidebar wiring test.

WitMiao:
## What does this PR do?

A pending workspace invitation can get stuck in the sidebar forever when the invitation is concluded from a different surface than the one showing the row: the workspace list updates, but the stale pending row remains and both **Join** and **Decline** silently fail (`400 "invitation is not pending"`), so the row survives until the app restarts.

This PR fixes it from both ends:

- **Server** — heals already-shipped clients with no client update:
  - `AcceptInvitation`: re-accepting an already-accepted invitation is idempotent while the membership from the first accept still exists — it returns the existing member (`200`). If the membership is gone (the user left the workspace afterwards), it still fails: leaving was explicit and a stale retry must not re-add anyone.
  - `DeclineInvitation`: declining a concluded invitation is a no-op (`204`) — the invitation is over either way, and the client only needs its refetch to drop the row.
  - `invitation:accepted` / `invitation:declined` are now also delivered to the acting invitee over the WebSocket (in addition to the workspace broadcast), so clients that aren't in the workspace room — including a user with no workspace open — still converge.
  - `JoinByShareLink` settles a matching pending email invitation in the same transaction, closing the share-link variant of the stuck row described in #8432.
- **Client** — updated clients also converge through the WebSocket path:
  - `invitation:accepted` / `invitation:declined` handlers now invalidate the account-level `myInvitations` list, gated on the frame's `actor_id` matching the current user, matching what `invitation:revoked` already did without turning every accept/decline into a refetch for all workspace members.
  - The sidebar's accept mutation invalidates the pending list on error, and decline moves to `onSettled`, so any outcome refreshes the row.

This approach is the right one because the stale row is unreachable for old clients otherwise: with `staleTime: Infinity` and no focus refetch, nothing ever re-reads the pending list, so a purely client-side fix could not rescue deployments with installed desktop builds.

**Decision note — the decline no-op is deliberate.** After this change, clicking **Decline** on a row for a workspace the user already joined returns `204` and clears the row, but the membership stays. Declining a concluded invitation must not carry the side effect of removing a membership — the user joined through some other explicit path, and that path owns the membership lifecycle. The button reads wider than what it does in this one corner; we preferred that over a decline that silently kicks people out of workspaces.

## Related Issue

Closes #8432

## Type of Change

- [ ] Documentation update
- [ ] CI / infrastructure

## Changes Made

- `server/internal/handler/invitation.go` — idempotent accept for already-accepted invitations with surviving membership; no-op `204` decline for concluded invitations; the stale-accept membership lookup now reports transient DB failures as `500` and only treats `pgx.ErrNoRows` as "membership gone" (the same discrimination `share_link.go` already uses). Ownership verification and every other non-pending status keep failing exactly as before.
- `server/internal/handler/share_link.go` — a share-link join settles a matching pending email invitation (`invitee_email`, case-normalized) inside the join transaction: the link's role wins for the new member, the invitation row is only marked accepted, and the invitation's seat reservation is released (durable enqueue plus best-effort immediate release; the outbox worker retries if the capacity service is unavailable). The settle publishes `invitation:accepted` with the accept-path shape so admin-side pending lists converge.
- `server/cmd/server/listeners.go` — `invitation:accepted` / `invitation:declined` additionally `SendToUser` the acting invitee with `excludeWorkspace` dedup, mirroring the `member:added` pattern; the workspace broadcast the admin lists depend on is unchanged. `invitation:revoked` keeps its `invitee_user_id` routing because its actor is the revoking admin.
- `server/internal/handler/invitation_test.go` — three DB-backed cases: re-accept returns the existing member (and no duplicate membership row); re-accept without membership still `400`; decline of an accepted invitation is `204` and leaves status and membership untouched.
- `server/internal/handler/share_link_test.go` — settle tests: pending invitation is accepted (invitation role untouched, member gets the link's role), the settle publishes `invitation:accepted`, managed capacity releases the invitation's reservation (outbox row gone), unmanaged capacity settles with nothing stranded, and a join without any pending invitation is unchanged.
- `server/cmd/server/listeners_scope_test.go` — listener-scope delivery tests: both conclusion events reach the workspace room and the acting invitee exactly once (targeted send excludes the workspace room); no actor means no targeted send; `invitation:revoked` still targets the invitee, not the actor.
- `packages/core/realtime/use-realtime-sync.ts` — `invitation:accepted` / `invitation:declined` also invalidate `workspaceKeys.myInvitations()`, gated on the event's `actor_id` matching the current user so other members' clients don't refetch the account-level list on every conclusion.
- `packages/core/types/events.ts` — the invitation conclusion frames now carry `actor_id`.
- `packages/views/layout/app-sidebar.tsx` — accept mutation gains `onError` invalidation; decline mutation moves its invalidation to `onSettled`.
- `packages/core/realtime/use-realtime-sync-invitations.test.tsx` — asserts the pending list refetches when either WS event lands, and the negative case: a member who is not the actor does not refetch their own list.
- `packages/views/layout/app-sidebar.test.tsx` — asserts both invitation mutations invalidate the pending list on failure.

## How to Test

1. Sign in on the desktop app; have another account invite that user to a workspace.
2. Accept the invitation from the web invite page (or another device) instead of the desktop app.
3. Back in the original desktop app session, observe the stale pending row; click **Join** → with this branch the server returns the existing member and the row drops (on `main`, nothing visibly happens). Clicking **Decline** on a stale row now also clears it (`204`) while the membership is kept (see the decision note above).
4. Repeat with acceptance from a second device while the desktop app is open: the row now disappears within a WebSocket tick without pressing anything — including when the concluding device has a different workspace (or none) open, which is the case the new listener delivery covers.
5. Share-link variant: invite a user by email, then have them join through the workspace share link instead. The pending row is settled by the join itself — it disappears from their pending list, and a later **Join**/**Decline** on it finds nothing pending. On a managed-capacity deployment the invitation's seat reservation is released, so the joining user holds exactly one seat.
6. Automated: `go test ./internal/handler/ -count=1`, `go test ./cmd/server/ -count=1`, `pnpm --filter @multica/core test -- realtime`, `pnpm --filter @multica/views test -- layout/app-sidebar`, `pnpm typecheck`.

## Checklist

**Risks considered:** the idempotent accept only fires for `status = "accepted"` *and* an existing membership row, so it cannot resurrect a membership after the user left, and it cannot be used to probe other users' invitations (the ownership check runs first). The no-op decline returns `204` only after the same ownership verification; expired, declined, revoked and accepted invitations all behave identically for the requester, which reveals nothing new. The share-link settle only matches the joining user's own email inside the join transaction, so it cannot settle anyone else's invitation; `AcceptInvitation` re-checks `status = 'pending'` under the row lock, so a concurrently concluded invitation degrades to "nothing to settle" instead of failing the join. Verified against `main`: both stale-row buttons currently fail silently and the row never clears without a restart.

## AI Disclosure

**Prompt / approach:** Diagnosed a self-hosted report of the stuck pending-invitation row, traced the full chain (WS invalidation gap → `staleTime: Infinity` defaults → silent mutation errors → server `400`), verified the server invitation handler and sidebar logic were unchanged from upstream, then reproduced with failing tests first (Go + Vitest) before implementing. All checks run locally: `go test ./internal/handler/`, `pnpm --filter @multica/core test`, `pnpm --filter @multica/views test` (5144 passing), `pnpm typecheck`.

## Screenshots (optional)

Not attached — the visible change is the *absence* of a stale row; repro steps above cover verification.


Bohan-J (changes requested):
Thanks for the detailed writeup and for shipping tests with the fix — the diagnosis chain in the description is accurate about the client state, and the product direction (the stale row must clear without a restart) is right.

The blocker is that the realtime half doesn't reach the user it's meant to help, and the share-link variant named in #8432 is still broken. Requesting changes on those two; the server-side idempotency and the sidebar refetch are worth keeping as-is.

## Blocking 1 — `invitation:accepted` / `invitation:declined` never reach the invitee, and the new invalidation amplifies requests for everyone else

These two events are not in `personalEvents` (`server/cmd/server/listeners.go`), so `SubscribeAll` delivers them via `BroadcastToWorkspace(W)` only. The invitee's WS connection is bound to the single workspace currently open (`packages/core/realtime/provider.tsx` — `ws.setAuth(token, wsSlug)`, and it doesn't connect at all when there's no slug). Before joining W they cannot be in W's room, so in the exact scenario from the issue the new invalidations never fire for them.

The flip side is that every *other* online member of W does receive the broadcast, and the new `myInvitations` invalidation is unconditional. `AppSidebar` always mounts that query, so it's an active observer: each accept/decline now makes every online member of the workspace refetch their own account-level `/api/invitations`. That's a fanout this PR introduces, not one it inherits.

Worth noting for the accept path specifically: it already converges on `main` via `member:added`, which `listeners.go` explicitly `SendToUser`s to the new member, and whose client handler already invalidates both `workspaceKeys.list()` and `workspaceKeys.myInvitations()` (`packages/core/realtime/use-realtime-sync.ts`). That's also what makes the workspace list appear correctly in the repro. So the `invitation:accepted` handler isn't the main fix here — **decline** is the path with no fallback at all, and it sticks every time.

Suggested shape:

- Keep the workspace broadcast (the admin-side member/invitation lists need it), and *additionally* `SendToUser(e.ActorID, data, e.WorkspaceID)` for these two events — the same pattern `member:added` already uses in `listeners.go`, including the `excludeWorkspace` argument for dedup.
- Use `e.ActorID` rather than `invitee_user_id`: accept and decline both publish with the invitee as the actor, whereas `invitee_user_id` can be NULL for email-addressed invitations. `ListPendingInvitationsForUser` matches on `invitee_user_id = $1 OR invitee_email = $2`, so a user can see a row whose `invitee_user_id` is null, and targeting by it would miss them. (`invitation:revoked` has to use `invitee_user_id` because the actor there is the admin, not the invitee.)
- Please don't move them into `personalEvents` — that would drop the workspace broadcast the admin lists depend on.
- On the client, gate the `myInvitations` invalidation on the frame's `actor_id` matching the current user, which also removes the fanout above.

On tests: `use-realtime-sync-invitations.test.tsx` invokes `handlers[event](...)` directly against a fake WS, so it passes whether or not delivery works — it can't catch this class of bug. Please add a listener-scope test asserting both the workspace broadcast and the actor-user send, plus a client-side negative case (a member who is not the actor does not refetch their own list).

## Blocking 2 — the share-link variant from #8432 is still broken

`JoinByShareLink` never settles a matching pending email invitation, so the invitation stays `pending`. The new idempotent branch only covers `status == "accepted"`, so **Join** still reaches `CreateMember` and returns 409, and the new `onError` invalidation refetches the same still-pending row straight back. Unlike the accept case, this row doesn't clear on restart either — it lives until the invitation expires.

Related, worth confirming: a `pending` invitation appears to keep its seat reservation alive (`server/internal/seatcapacity/worker.go` keeps the reserve while the invitation is still `pending`), which would leave one user holding both a member seat and an invitation reservation on managed-capacity deployments. I only traced the recovery path, so please verify before relying on that.

Either of these works for me:

- Settle the matching pending invitation inside the share-link join transaction. `GetPendingInvitationByEmail(workspace_id, invitee_email)` already exists, and the `idx_invitation_unique_pending` partial index guarantees at most one row per (workspace, email). This needs role precedence defined (share-link role vs. invitation role) and the seat intent transferred or released, with tests for both unmanaged and managed capacity.
- Or drop `Closes #8432`, scope this PR to the accept/decline paths, and open a tracked follow-up. The issue shouldn't auto-close while one of its stated repro paths still reproduces.

## Non-blocking

**`AcceptInvitation` swallows DB errors as a business error.** The new branch uses `if memberErr == nil`, so any error — including a transient DB failure — falls through to `400 "invitation is not pending"`. `share_link.go` already has the discrimination you want: `errors.Is(memberErr, pgx.ErrNoRows)` for absent, 500 otherwise.

**The decline no-op deserves an explicit note.** A user clicking **Decline** on a row for a workspace they already joined now gets 204 and the row disappears, but they remain a member. I think that's the correct call — declining a concluded invitation shouldn't have the side effect of removing a membership — but the button then does something other than what it says. Worth calling out in the PR description, and possibly in the UI, so it reads as a decision rather than an accident.

**Consider a cheap fallback refetch, independent of all of the above.** Giving `myInvitationListOptions()` a real `staleTime` plus `refetchOnWindowFocus: true` would make the row clear the moment the user focuses the app, whichever surface concluded the invitation. It's an account-level list that's empty for most users, so the cost is negligible, and realtime delivery can't cover every path — self-hosted reverse proxies that don't pass the WS upgrade, dropped events, offline periods. Correctness of this list probably shouldn't depend on delivery at all. Fine as a follow-up if you'd rather keep this PR focused.

## Keep as-is

The server-side idempotency is the only part that heals already-installed clients, and the reasoning for keeping the "membership is gone" case failing is right. The sidebar's `onError` / `onSettled` refetch is a good backstop too.

## What I checked

Code-level verification against `main`: event routing and `personalEvents` classification in `listeners.go`, the WS connection binding in `provider.tsx`, the `member:added` handler's existing invalidations, the `status = 'pending'` filter in `ListPendingInvitationsForUser`, the share-link join path, and the `errors.Is(pgx.ErrNoRows)` pattern already used in `share_link.go`. I did not run a cross-surface end-to-end reproduction against a self-hosted deployment, and I did not run the test suites.

WitMiao:
Thanks for the thorough review — the `member:added` convergence observation and the fanout callout were both correct and shaped the fix. Addressing each point:

## Blocking 1 — delivery to the invitee

Done as suggested, in `server/cmd/server/listeners.go` (cda41a34d):

- `invitation:accepted` / `invitation:declined` now additionally `SendToUser(e.ActorID, data, e.WorkspaceID)`, mirroring the `member:added` pattern including the `excludeWorkspace` argument, so clients already in the room (reached by the broadcast) don't get the event twice. The broadcast is unchanged and neither event moved into `personalEvents`.
- Targeting uses `e.ActorID`, not `invitee_user_id`, for exactly the reason you gave: the actor is the invitee on both producer paths, while `invitee_user_id` can be NULL on email-addressed invitations that `ListPendingInvitationsForUser` still matches by email. `invitation:revoked` keeps its `invitee_user_id` routing since its actor is the revoking admin — there's a regression test pinning that.
- Client side (654b6c50a): the `myInvitations` invalidation is now gated on the frame's `actor_id` matching the current user, and the conclusion frames carry `actor_id`.

Tests: `listeners_scope_test.go` asserts both the workspace broadcast and the actor-targeted send happen (recipient, `excludeWorkspace` argument, and frame shape including `actor_id`), plus the two negatives you asked for — no actor means no targeted send, and revoked still targets the invitee rather than the actor. The client negative case (a non-actor member does not refetch their own list) is in `use-realtime-sync-invitations.test.tsx`.

## Blocking 2 — share-link variant (went with settling in the join transaction)

Implemented option A in `server/internal/handler/share_link.go` (ee10b913b). `Closes #8432` stands — both repro paths now converge.

- The settle runs inside the join transaction: `GetPendingInvitationByEmail` on the joining user's (case-normalized) email, then `AcceptInvitation`, whose `WHERE status = 'pending'` re-check under the row lock turns a concurrent decline into "nothing to settle" rather than a failed join. The partial index bounds it to at most one row per (workspace, email).
- **Role precedence:** the share link wins. The user actively chose that entry point, so the member is created with `link.Role` and the invitation row is only settled — its own role is left untouched rather than written back onto the member. If you'd rather see the invitation's role take precedence, it's a two-line flip.
- **Seat semantics — verified, and handled.** Your suspicion was right for managed capacity: `CreateInvitation` reserves a seat through `reserveInvitationCapacity` (durable `reserve_invitation` intent + `ReserveInvitation`), and it stays held while the row is pending — accept consumes it into the member seat, decline/revoke/expiry release it, but a share-link join never touched it. So a user joining via link while holding a pending email invitation ended up holding a member seat *and* an invitation reservation until the invitation expired. (Unmanaged mode keeps nothing durable — the intent is deleted on the spot.) Since the join's own confirmed claim already covers the seat, the settle now enqueues a release for the invitation's token inside the transaction and attempts the release immediately after commit, with the outbox worker as the fallback path — the same pattern decline uses, so there's no double occupancy.
- The settle also publishes `invitation:accepted` with the accept-path payload shape so workspace admins' pending-invitation lists converge instead of waiting on a refetch that nothing else triggers.

Tests cover both capacity modes: managed asserts the release call lands on the invitation's token and the outbox row is gone after the join; unmanaged asserts the settle succeeds with nothing stranded in the outbox. There's also a case pinning that the member gets the link's role while the invitation row keeps its own, and one for a plain join with no pending invitation.

## Non-blocking

- **DB errors as 400:** fixed in 45d22f532 — only `errors.Is(memberErr, pgx.ErrNoRows)` means "membership gone" and falls through to the `400`; anything else is a `500`, same discrimination as the `JoinByShareLink` pre-check.
- **Decline no-op:** now explicit in the PR description ("Decision note") — the membership stays because declining a concluded invitation shouldn't remove a membership some other explicit path created. We left UI copy out of scope for this PR.
- **`staleTime` + focus refetch on `myInvitationListOptions`:** agreed it's the right independent backoff, kept out of this PR as a follow-up. Happy to add it here instead if you'd rather land it with this fix.

All gates green on the merged branch: `go test ./internal/handler/ -count=1`, `go test ./cmd/server/ -count=1`, `gofmt`/`go vet`, `pnpm --filter @multica/core test -- realtime` (1870), `pnpm --filter @multica/views test -- layout/app-sidebar` (5144), `pnpm typecheck`.

WitMiao:
@Bohan-J Both blockers and the non-blocking items are addressed on this branch (details above) — ready for another look whenever you get a chance. Sorry for the extra ping; the API would not let me file a formal re-request.

Bohan-J (approved):
Thanks for the thorough turnaround — every point from the previous round is addressed, and a couple of them are better than what was asked for.

Checked on `99639d6`:

- **Targeted delivery.** `invitation:accepted` / `invitation:declined` now `SendToUser(e.ActorID, ..., e.WorkspaceID)` alongside the unchanged workspace broadcast, with the `excludeWorkspace` dedup and the empty-actor guard. `personalEvents` is untouched, so the admin lists keep their fanout, and the comment correctly records why `invitation:revoked` stays on `invitee_user_id` routing.
- **Client gate.** The `myInvitations` invalidation is keyed on the frame's `actor_id`, which removes the per-member refetch amplification.
- **Tests.** The wire-contract cases that drive a real `WSClient` through a fake socket with the exact frames `listeners.go` marshals are a better answer than the listener-scope test I suggested — they pin the unpack shape end to end, which is where the original test was blind. The non-actor negative case and `TestRegisterListeners_InvitationConcludedWithoutActorSkipsTargetedSend` cover the rest.
- **Share-link settle.** Inside the join transaction, with `pgx.ErrNoRows` handled on both the lookup and the `AcceptInvitation` race, the seat reservation released via the outbox plus a best-effort immediate compensate, and an `invitation:accepted` publish so both admins and the joiner converge. Role precedence is documented and the invitation's own role is left alone.
- **Error discrimination.** `errors.Is(memberErr, pgx.ErrNoRows)` for "membership gone", 500 for transient failures.
- **Decision note** on the decline no-op keeping the membership — good call writing that down in the description rather than leaving it implicit.

Merging. Two small follow-ups, neither blocking and neither something I'd ask you to fold in here:

1. **Rows already stuck on deployed instances aren't healed by Join.** For a user who is already a member *and* still has a pending invitation (the state this PR now prevents, but which existing deployments can already be in), `AcceptInvitation` marks the invitation accepted and then `CreateMember` hits the unique violation and returns 409, rolling the settle back — so Join can never clear those rows. Decline still works, so there is an escape hatch. Returning the existing member on that unique violation, the same shape as the already-accepted branch, would heal them too.
2. **The share-link settle matches on `invitee_email` only,** while `ListPendingInvitationsForUser` matches `invitee_user_id = $1 OR invitee_email = $2`. A user whose account email changed after the invitation was created would still see the row but wouldn't have it settled. Narrow, and the row is still clearable via Decline.

Separately, giving `myInvitationListOptions()` a real `staleTime` plus `refetchOnWindowFocus` is still worth doing at some point, so this list stops depending on delivery for correctness — self-hosted proxies that drop the WS upgrade, offline periods, and so on. Out of scope here.

PR #8440, opened 2026-09-15
MUL-7395 fix(qwen): exclude cache reads from plain input usage

importcpp:
## What does this PR do?

Qwen reports cached reads inside `input_tokens`. Multica copied that inclusive count into its plain-input bucket and also recorded the cache bucket, inflating both totals and rate-based cost estimates. A response with 1,000 input tokens (600 cached) and 50 output tokens counted as 1,650 instead of 1,050.

Normalize at the Qwen adapter boundary by subtracting cached reads and flooring the plain-input remainder at zero. Both assistant and result usage already pass through this function. Qwen [v0.23.4's producer](https://github.com/QwenLM/qwen-code/blob/fdcd7c2761e4ae18d578cf1413ae8d43f58b3e2c/packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.ts#L212) and the repository's captured 0.20.0 fixture have this convention.

## Related Issue

Found while auditing usage after #8431; independent of that fix. No separate issue.

## Type of Change

## Changes Made

- `qwen.go`: keep plain input and cache reads mutually exclusive.
- `qwen_usage_test.go`: pin uncached, mixed, all-cached, inconsistent and empty counters.
- `qwen_test.go`: correct Execute and captured-fixture expectations; the real fixture now totals 46,698 tokens instead of 88,598.

## How to Test

1. The new mixed/all-cached/inconsistent cases fail against main before the fix.
2. From `server/`: `../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run '(?i)^Test.*Qwen' -parallel 4 -count=1` — 27 top-level tests pass.
3. `go vet ./pkg/agent` and `git diff --check` pass.

No real coding CLI or model account was used. Full repository tests were not run.

## Risks and scope

This corrects the native Qwen stream-json input convention; it does not change other adapters. Output and cache-read counts remain the provider's reported values. Previously stored records are not rewritten. Resume totals and multi-response fallback aggregation are separate fixes; no schema or API changes.

## Checklist

UI screenshots, product documentation, localization and landing updates do not apply to this backend accounting change.

## AI Disclosure

**Prompt / approach:** Audit provider token normalization, verify Qwen's producer and captured fixture, reproduce overlapping cache counts, and deliver a small adapter-only fix with regression coverage.


multica-eve (approved):
Thank you, @importcpp, for the focused fix and clear regression coverage. I verified that the updated head still contains the reviewed Qwen-only diff and that all CI checks are green. Approved for merge.

PR #8441, opened 2026-09-15
MUL-7396: fix(usage): preserve cost-only task accounting

importcpp:
## What does this PR do?

Preserve provider-reported cost when a run has no token breakdown, through both daemon reporting and server cost monitoring. For example, a Grok ACP result reporting $0.01 and zero tokens was discarded by `runTask`. Retaining that result alone lets storage/UI show the charge, but a priced model such as `grok-4.6` still loses it from `llm_cost_usd` because the metrics helper returned early on zero tokens.

Keep usage entries with positive provider cost and record metric cost independently of token counts. The existing authoritative-cost distribution puts a cost-only charge in the input bucket; zero-token `llm_tokens` series are not created.

## Related Issue

No functional dependency on #8402, which changes nearby usage-reporting code but retains the original daemon filter.

## Type of Change

## Changes Made

- `server/internal/daemon/daemon.go`: retain entries with positive provider cost even when all four token counters are zero.
- `server/internal/daemon/task_usage_test.go`: exercise a test-created Grok ACP process through the real backend and `runTask`, covering cost-only, tokens plus cost, tokens only, and empty usage.
- `server/internal/metrics/business.go`: independently guard positive token counts and positive cost when recording priced usage.
- `server/internal/metrics/business_test.go`: cover priced and unpriced cost-only records, absence of zero-token series, request counting, and empty priced usage. Existing token-bearing controls verify authoritative cost, proportional allocation, token counts, and rate-table fallback.

## How to Test

1. Before the daemon change, `TestRunTaskPreservesProviderCostWithoutTokens/cost_only` fails with `Usage=[]`; its three controls pass.
2. On the original PR head, `TestBusinessMetricsCostOnlyUsage/priced` fails with `cost series = 0, want 1` and `recorded cost = 0, want 0.01`. The unpriced case, empty record, normal token-bearing provider cost, and rate-table fallback controls pass.
3. From `server/`, `../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/metrics -count=1` passes: 41 top-level tests and 40 subtests.
4. From `server/`, `../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/daemon -run '^(TestRunTaskPreservesProviderCostWithoutTokens|TestMergeUsage|TestHandleTask_ReportsUsageBeforeCancel|TestHandleTask_ReportsUsageWhenCancelledByPoll)$' -count=1` passes: 4 top-level tests and 4 subtests.
5. `go vet ./internal/metrics ./internal/daemon`, formatting, and `git diff --check` pass.
6. [GitHub CI for commit `8ab0db31f`](https://github.com/multica-ai/multica/actions/runs/34943324087) passes: all 18 executed jobs succeeded; `image-budget` was skipped by the workflow.

The ACP process fixture is POSIX-only and skips Windows. Requests use a local httptest server; no installed coding CLI, model account, or database is used. Full repository tests were not run locally.

## Risks and scope

The change preserves positive provider charges at two boundaries. Empty entries without positive provider cost remain omitted by the daemon. Existing cost units, price distribution, metric labels, API, and schema are unchanged. Cost-only metrics use the existing input bucket convention; this does not imply measured input tokens. No retries or historical backfill are added. The fixture proves the supported payload survives these boundaries, without making a claim about how often a live provider emits it.

## Checklist

## AI Disclosure

**Prompt / approach:** Trace provider cost-only usage through the daemon and server metrics. Reproduce each loss before changing behavior, apply scoped fixes, and verify cost totals and token-series behavior with regression tests and controls.


multica-eve (changes requested):
Request changes: please preserve cost-only usage in the server-side `llm_cost_usd` metric as well. The daemon fix is directionally correct, but it currently makes storage/UI and internal cost monitoring disagree about the same charge.

Trigger: a provider reports a positive `cost_usd_ticks` with all four token counters at zero, and the model has a server pricing-table entry (for example, the test's `grok-4.6`).

Consequence: `ReportTaskUsage` persists the row and calls `CaptureTaskUsage`, which reaches the priced branch of `RecordLLMUsage`. `distributeAuthoritativeCost` correctly puts the full charge in the input bucket when there is no token shape, but `recordPricedTokens` then returns early because `tokens <= 0`. The task usage UI therefore shows the provider's charge while `llm_cost_usd` records zero. An unpriced model takes the separate branch and does record the same cost, so metric correctness unexpectedly depends on whether the model appears in the rate table.

Please handle the priced cost-only case before merge: when `costUSDTicks > 0` and all token counters are zero, add the authoritative charge directly to `llm_cost_usd` without creating a zero-token `llm_tokens` series. The unpriced branch already demonstrates the intended behavior. Add a regression test covering a priced cost-only record (plus a normal token-bearing control).

Relevant code: `server/internal/metrics/business.go:674-681` and `server/internal/metrics/business.go:708-715`.

importcpp:
Thanks for catching this, @multica-eve. You're right: the original fix covered the daemon boundary but missed the downstream cost metric.

Addressed in 8ab0db31ff1615b28e83b16698be3c8579338525. `recordPricedTokens` now guards token and cost updates independently: positive cost is recorded even when tokens are zero, while `llm_tokens` is only updated for positive token counts. This reuses the existing authoritative-cost distribution, so a cost-only charge lands in the input bucket without creating zero-token series or changing the normal cost split.

I reproduced the issue before the fix: the priced `grok-4.6` case recorded $0 instead of $0.01, while the unpriced and normal token-bearing controls passed. Added regression coverage for priced/unpriced cost-only records, absence of token series, request counting, and empty priced usage. The full metrics suite, focused daemon tests, and `go vet` pass locally; [CI for the updated commit](https://github.com/multica-ai/multica/actions/runs/34943324087) also passes.

I'd appreciate another look when you have a chance. Thanks again for the careful review.

PR #8443, opened 2026-09-15
MUL-7397: fix(qwen): accumulate fallback usage by message

importcpp:
## What does this PR do?

Retain every completed Qwen assistant message's usage when the process exits without final usage. Previously each message overwrote the previous usage for that model: two messages consuming 150 and 220 tokens followed by exit 7 reported only 220 instead of 370.

Count each message ID's first non-empty usage snapshot once, keeping its counters and model together. Messages without an ID retain the existing latest-per-model behavior. A non-empty terminal usage total remains authoritative.

## Related Issue

MUL-7397. Updated onto main after #8440 (cache normalization) merged. This PR has no dependency on an unmerged PR. Excluding restored-session history remains a separate follow-up to this aggregation.

## Type of Change

## Changes Made

- `server/pkg/agent/qwen.go`: read the assistant message ID and maintain execution-local fallback totals. Deduplicate by ID alone; do not combine counters across snapshots or count a repeated ID again under another model. Empty identified snapshots do not consume the ID.
- `server/pkg/agent/qwen_fallback_usage_test.go`: cover multiple messages/models, repeated IDs with changed counters or model attribution, empty/cache-only usage, anonymous snapshots, terminal precedence, and execution-local state.
- `server/pkg/agent/qwen_test.go`: add a fake Qwen mode that emits two completed messages and exits with code 7.

The pinned Qwen producer [allocates a new UUID at message boundaries and emits the full assistant message at finalization](https://github.com/QwenLM/qwen-code/blob/fdcd7c2761e4ae18d578cf1413ae8d43f58b3e2c/packages/cli/src/nonInteractive/io/BaseJsonOutputAdapter.ts#L350-L417). [Block closure only emits a partial stream event](https://github.com/QwenLM/qwen-code/blob/fdcd7c2761e4ae18d578cf1413ae8d43f58b3e2c/packages/cli/src/nonInteractive/io/StreamJsonOutputAdapter.ts#L307-L323). This supports counting finalized messages; the repeated-ID cases are defensive tests, not a claim that Qwen emits monotonic full-message updates.

## How to Test

1. The original failure-process regression expects 300 input / 70 output after two messages; main retains only 200 / 20. It reuses the backend for two executions to verify totals reset per run.
2. Before the review fix, with #8440 included, the new decreasing-counter regression produced 60 plain input / 60 output / 40 cached input instead of retaining the first snapshot's 80 / 60 / 20. The corrected-model case counted the same ID in two model buckets. Both now pass.
3. From `server/`, all 30 selected top-level tests and 27 subtests pass with the race detector:

   ```sh
   ../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run '(?i)^Test.*Qwen' -parallel 4 -count=1
   go vet ./pkg/agent
   ```

   `git diff --check` also passes.

The process fixture is POSIX-only and skips Windows; parser regressions are platform-independent. No installed agent CLI, model account, database, or full repository test run was used locally.

[CI for the updated commit](https://github.com/multica-ai/multica/actions/runs/34944804519): agent tests and all platform checks passed. The full backend job failed in the unchanged `TestUserCancelledRecoveryIsSettled`, where the global recovery sweep returned `{Scanned:1 Replayed:1 Exhausted:0}` instead of zero after the settled-signal assertion had passed. This was reproduced on main with another workspace's pending recovery and is addressed separately by the test-only follow-up #8445. The recovery service/test files are unchanged by this PR.

## Risks and scope

The first non-empty snapshot for an identified message wins, including its original model attribution; later same-ID corrections are not applied without an upstream update contract. For messages without an ID, only the latest anonymous snapshot per model is retained because distinct calls cannot be identified safely. An absent or all-zero terminal usage leaves the fallback intact; non-empty terminal usage wins even if an assistant event arrives later. Usage deduplication does not suppress transcript events.

State belongs to one execution and grows with distinct message IDs. This does not change cache normalization, resumed-session accounting, pricing, APIs, or schema, and does not reconstruct unstreamed usage after a process failure.

## Checklist

UI screenshots, product documentation, localization and landing updates do not apply to this backend accounting change.

## AI Disclosure

**Prompt / approach:** Reproduce Qwen fallback undercounting after process failure, inspect the pinned producer's message identity and finalization behavior, and implement per-message aggregation. Validate reviewer-reported counter mixing and duplicate model attribution with failing regressions, then use the first non-empty whole snapshot per ID and preserve terminal usage precedence.


multica-eve (changes requested):
Requesting changes because the core fix is desirable, but the per-field `max` reconciliation can synthesize usage that Qwen never reported and can therefore overcharge a failed run.

The product problem is real: Qwen can complete multiple model calls before exiting without a usable terminal usage summary, while the current fallback retains only the last assistant message. Accumulating execution-local usage by message identity and continuing to prefer a non-empty terminal total is the right behavior.

The snapshot policy in `server/pkg/agent/qwen.go:373-397` needs to change before merge. If one snapshot for a message reports `input=100/output=60/cache=20` and a later corrected snapshot reports `input=100/output=50/cache=40`, taking each field's maximum records `100/60/40`, which is not either provider-reported snapshot. The composite `(model, messageID)` key also counts the same message twice if its model attribution is corrected.

The pinned Qwen producer does not establish the claimed full-message duplication: block closure emits only a partial `stream_event`, full assistant messages are emitted at finalization, and block-type transitions allocate a new UUID. It therefore does not establish that repeated snapshots exist or that every usage field is monotonic. The `updated_then_stale_snapshot` test currently locks in this unsupported policy.

Please use `message.id` as the identity and select one whole snapshot under a protocol-backed rule. With the current wire contract, the smallest fix is to count each identified, non-empty message once and preserve the existing latest-per-model behavior when the ID is absent. If updated snapshots must be supported, please provide upstream evidence, replace the whole snapshot atomically (moving it between model buckets when necessary), and add regressions for decreasing counters and corrected model attribution. Keep the current non-empty terminal-usage precedence.

Validation on `245a3313`: the Qwen test suite with the race detector, `go vet ./pkg/agent`, and `git diff --check` all pass. Those tests do not cover the unsafe correction cases above.

importcpp:
Thanks for the careful review, @multica-eve. You're right about both the counter mixing and the duplicate accounting across model names. I also misread the producer's block-close behavior; the original description did not justify treating full-message usage fields as monotonic.

Addressed in dcfbad1abb30ac8f89c341aa27057a6d479914da using the smaller policy you suggested: deduplicate by `message.id` alone and count the first non-empty whole snapshot once, keeping its counters and original model attribution together. Empty snapshots do not consume the ID. Messages without an ID retain the latest-per-model behavior, and non-empty terminal usage still takes precedence. The per-field `max` reconciliation and its unsupported test expectation are removed.

I reproduced both cases before the fix and added regressions for decreasing counters, changed model attribution, empty/cache-only snapshots, and anonymous usage updates. The PR description now correctly distinguishes partial block-close events from finalized assistant messages; the repeated-ID cases are defensive tests rather than a claim that the producer supports snapshot updates. The branch also includes the now-merged cache normalization fix from #8440.

All 30 selected Qwen-related top-level tests and 27 subtests pass locally with the race detector, along with `go vet ./pkg/agent` and `git diff --check`. In [CI for the updated commit](https://github.com/multica-ai/multica/actions/runs/34944804519), the agent tests and all platform checks passed. The full backend job failed in the unchanged `TestUserCancelledRecoveryIsSettled`: its settled-signal assertion passed, but the subsequent global sweep returned `{Scanned:1 Replayed:1 Exhausted:0}` instead of zero. That failure still needs separate triage; this PR does not change the recovery service or its tests.

I'd appreciate another look when you have a chance. Thanks again for catching this.

importcpp:
Follow-up on the backend CI failure: opened #8445 with a focused test fix.

I reproduced the failure on main, without this PR's Qwen changes: the cancellation test passes on its own, but adding another workspace's pending recovery produces the same `{Scanned:1 Replayed:1 Exhausted:0}` failure in all three runs. The user's cancelled signal remains settled; the extra replay belongs to the other workspace.

#8445 replaces the global-zero assertion with checks tied to each recovery comment ID, and includes the second workspace as a regression case. It changes one test file and can merge independently of this PR. The local service suite and repeated recovery checks pass, and [CI for #8445](https://github.com/multica-ai/multica/actions/runs/34946764127) now passes too, including the full backend test job.

Thanks again for the review. This keeps the CI test fix separate from the Qwen accounting change here.

multica-eve (approved):
Approved at `dcfbad1abb30ac8f89c341aa27057a6d479914da` with no blocking findings.

The revised implementation resolves the billing correctness issue from my previous review: identified assistant messages are now deduplicated by `message.id` and contribute the first non-empty whole usage snapshot, while anonymous messages retain the existing latest-per-model behavior and a non-empty terminal result still takes precedence. This avoids both synthesizing a usage tuple through per-field maxima and double-counting a message whose model attribution changes. The added regressions cover decreasing counters, corrected model attribution, empty/nil placeholders, cache-only usage, anonymous messages, terminal precedence, and execution-local state.

I also inspected the failed CI run. Its sole test failure is the unchanged `TestUserCancelledRecoveryIsSettled`, where a global recovery sweep observed another workspace's pending signal (`Scanned:1 Replayed:1`) despite the target cancellation being settled. The aggregate `backend` check failed as a consequence of `backend-tests`; the Qwen-specific `backend-agent-tests` passed. PR #8445 isolates that test assertion to the relevant signals and its `backend-tests` run is green, so this failure is not attributable to this PR.

Local validation passed: Qwen agent tests under the race detector, `go vet ./pkg/agent`, and `git diff --check`.

Risk remains confined to fallback token accounting for interrupted Qwen executions; a regression could misreport billable usage, and reverting these two commits restores the previous last-message-only fallback behavior.

PR #8444, opened 2026-09-15
test(cursor): keep finish fixture alive through terminal cleanup

importcpp:
## What does this PR do?

Fix the scheduling race in `TestCursorBackgroundLifecycle/finish`. Its fake Cursor emits a background launch and a terminal result, then exits immediately. If Linux consumes that buffered launch after the fake has exited, the shell has already been reparented: ancestry verification correctly rejects ownership, but the test still expects owned-process cleanup and fails with `background process ... survived`.

Keep the fake waiting for its child after emitting the terminal result. The backend can then capture ownership and perform terminal cleanup before the fixture exits. The existing result, transcript, cancellation, and process-liveness assertions remain intact.

## Related Issue

Found in the [Ubuntu CI failure on #8285](https://github.com/multica-ai/multica/actions/runs/34827269594/job/103922421221). The relevant Cursor code is identical on #8285 and current main. This PR is based directly on main and has no dependency on #8285.

## Type of Change

- [x] CI / infrastructure

## Changes Made

- Add eight lines to the `finish` branch of the existing subprocess fixture in `server/pkg/agent/cursor_liveness_test.go`.
- Wait for the fixture child after the result, preserving the terminal-cleanup scenario without a scheduling sleep or relaxed assertion.

## How to Test

Local validation on macOS, using the repository agent CLI guard and test-created executables:

- A standalone subprocess probe of the actual fixture observed premature parent exit in 3/3 trials before the change; after the change, the parent remained alive after the terminal result in 3/3 trials. The probe checks the fixture lifetime, not the complete Linux cleanup path.
- All 14 focused background/process tests passed under `-race`; the `finish` subtest passed 30 consecutive `-race` runs.
- All 22 focused daemon background/idle-watchdog tests passed under `-race`.
- The seven selected native capture/finalization tests passed five times with `CGO_ENABLED=0`.
- `go vet ./pkg/agent`, gofmt, and `git diff --check` passed.

The main focused commands, run from `server/`, were:

```sh
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run 'TestCursorBackground|TestCaptureCursorBackground' -count=1
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./pkg/agent -run '^TestCursorBackgroundLifecycle/finish$' -count=30
../scripts/go-test-with-agent-cli-guard.sh -- go test -race ./internal/daemon -run 'Test.*Background.*Watchdog|Test.*IdleWatchdog' -count=1
CGO_ENABLED=0 ../scripts/go-test-with-agent-cli-guard.sh -- go test ./pkg/agent -run 'TestCaptureCursorBackground|TestCursorBackgroundLifecycle/finish' -count=5
```

Local Docker uses Linux 6.8, which lacks the required pidfd process-group operation. Its failing controls are not counted as equivalent Linux validation. No real agent CLI or authenticated model was used.

The [CI run for `29c1a7f03`](https://github.com/multica-ai/multica/actions/runs/34941541690) passed: Linux/macOS/Windows Cursor lifecycle jobs, full backend and agent jobs, and all remaining executed checks succeeded. The Linux `finish` subtest is explicitly recorded as passing. `image-budget` was conditionally skipped.

## Risks and Scope

Test fixture only; production behavior and process-ownership checks are unchanged. This does not claim cleanup of a real CLI's background process whose ownership was never captured. Existing unverified-process fallback and root-exit-during-capture tests remain in place.

## Checklist

- [x] Included the triggering scenario and reasoning for the change
- [x] Ran local checks and documented their platform limits
- [x] Updated the existing regression fixture without removing assertions
- [x] Considered and documented scope and risks
- UI screenshots, localization, runtime landing copy, and product documentation: not applicable to this test-only change

## AI Disclosure

**AI tool used:** Codex

**Prompt / approach:** Investigated the CI failure reported on #8285, compared the relevant source with current main and the original ownership contract, checked the fake process lifetime before and after the change, and ran focused lifecycle/watchdog checks. Kept the fix independent and limited to the test fixture.


Bohan-J:
Thanks for tracking this down — your diagnosis was correct, and unfortunately it collided with a fix that landed on `main` a couple of hours after you opened this PR.

`main` now carries `test(cursor): synchronize ownership before finalization`, which came in bundled with #8436 (02a8ff3e). It targets the same `finish` branch of `runFakeCursorStream` and the same root cause you identified — the fake exiting before the backend consumes the buffered launch event, reparenting the shell so ancestry verification correctly rejects ownership. The landed version blocks the fake *before* it emits `task_notification`/`result`, gated on a `finish.release` file that the test writes only after asserting `session.ToolActivity() == 1`, so the barrier is keyed directly on ownership capture and adds that assertion to the test.

Your `child.Wait()` after the terminal result reaches the same outcome, but indirectly — it waits on the child exiting rather than on ownership being captured. Merging it on top of the current `main` applies cleanly and still passes, but it stacks a second wait on top of the gate, so it would be redundant.

One clarification in case it matters for your read of the state: #8451 removed the *duplicate* Linux `cursor-background-lifecycle` job, but `TestCursorBackgroundLifecycle/finish` still runs on Linux under `backend-agent-tests` (ubuntu-latest, `-race`). So the flake you hit was real and worth fixing — this is purely a case of two fixes racing each other.

Closing as superseded. Sorry for the wasted effort on the timing here, and thanks for the careful investigation and write-up — the reproduction notes made this easy to verify.

PR #8448, opened 2026-09-15
MUL-7405: fix(mobile): build and launch the iOS app on Xcode 27

### LINKED ISSUE #8447: iOS app cannot build or launch with Xcode 27 (iOS 27 SDK)
### Summary

Building `apps/mobile` with Xcode 27 fails in three places, and after working around the build errors the app terminates at launch.

### Environment

- macOS, Xcode 27.0 (27A266a)
- iPhone 17 Pro Max, iOS 26.5
- `apps/mobile` at `c4c8bab` — expo 55.0.24, react-native 0.83.6, pnpm 10.28.2

### Steps

1. `pnpm install`
2. `pnpm ios:mobile:device:prod:release`

### Actual

1. `expo run:ios` aborts before building: `Can't determine id of Simulator app; the Simulator is most likely not installed on this machine.` Xcode 27 no longer ships `Simulator.app` (it ships `DeviceHub.app`), and `@expo/cli` 55.0.x only resolves the old bundle id.
2. `xcodebuild` fails: `The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 6.0, but the range of supported deployment target versions is 15.0 to 27.0.x` (`iosMath-mathFonts`, `RNSVG-RNSVGFilters`). Xcode 27 treats targets below iOS 15.0 as errors.
3. With the pod targets raised to 15.1, the ExpoRouter module fails to compile: `expo-router/ios/LinkPreview/LinkPreviewNativeActionView.swift:141:20: error: 'subtitle' is only available in iOS 16.0 or newer`.
4. After a successful build and install, the app terminates immediately on launch: `EXC_BREAKPOINT` in `UIKitCore ___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption_block_invoke`. The iOS 27 SDK requires the scene-based life cycle — Apple TN3187 — and the prebuild template still creates its window in `application(_:didFinishLaunchingWithOptions:)`.

### Expected

`pnpm ios:mobile:device:prod:release` builds, installs, and the app launches.

### Notes

I have a fix ready: a config plugin (`apps/mobile/plugins/withIosXcode27Support.js`) that adds the scene manifest and a `SceneDelegate`, raises the pod deployment targets, and moves `ios.deploymentTarget` to 16.0, plus a pnpm patch for the `@expo/cli` Simulator-app prerequisite. PR incoming.

erlinerd:
## What does this PR do?

Makes the iOS app build and launch with Xcode 27 (iOS 27 SDK). Building `apps/mobile` with Xcode 27 currently stops three times — `expo run:ios` aborts on a `Simulator.app` check Xcode 27 invalidated, `xcodebuild` rejects pod targets below iOS 15.0, and the ExpoRouter module cannot compile at the 15.1 deployment target. With those worked around, the installed app then dies at launch, because the iOS 27 SDK requires the scene-based life cycle. This PR fixes all four at prebuild time, so `pnpm ios:mobile:device:prod:release` works unchanged.

**Breaking change (needs product sign-off):** the deployment target moves from iOS 15.1 to iOS 16.0, so iOS 15 devices can no longer install the app. See "Type of Change" below.

## Related Issue

Closes #8447

## Type of Change

- [x] **Breaking change** (deployment target raised from iOS 15.1 to iOS 16.0 — drops iOS 15 devices)

Why 16.0 is required, not just desired: `expo-router`'s iOS code (SDK 55) assigns the iOS 16-only `UIMenuElement.subtitle` without an availability guard (`expo-router/ios/LinkPreview/LinkPreviewNativeActionView.swift:141`), so the module cannot compile at 15.1, and the app target that imports it must match. Keeping iOS 15 would require patching expo-router's iOS sources (or an availability-guard upstream patch) on top of this fix — happy to go that route instead if product wants to keep iOS 15; otherwise this PR documents the drop for sign-off here.

## Changes Made

- **`apps/mobile/plugins/withIosXcode27Support.js`** (new) — config plugin, applied on every prebuild (`scripts/ios-run.sh` prebuilds each run):
  - adds `UIApplicationSceneManifest` to `Info.plist`;
  - moves window creation out of `application(_:didFinishLaunchingWithOptions:)` into a generated `SceneDelegate` that mirrors expo's upstream `ExpoAppSceneDelegate` + `SceneEventForwarder` (SDK 56+), so the plugin can be deleted once the app moves to that SDK. The generated delegate:
    - rebuilds the launch options that `Linking.getInitialURL()` reads from the scene's `connectionOptions` (raw `UIApplicationLaunchOptionsURLKey` / user-activity-dictionary keys), so a link that cold-starts the app is delivered through bridge startup instead of a post-startup event that fires before JS is ready;
    - forwards every scene event — URLs, user activities (including `willContinue` / `didFail` / `didUpdate`), quick actions from both `connectionOptions` and `windowScene`, and the four life-cycle callbacks — to `ExpoAppDelegateSubscriberManager` and `RCTLinkingManager` exactly once, deduplicated by observing the `RCTOpenURLNotification` that `RCTLinkingManager` posts, mirroring upstream;
    - keeps `UIApplication.shared.delegate?.window` populated.
  - raises every pod target to `podfile_properties['ios.deploymentTarget']` inside the Podfile `post_install` hook.
  - All three mods are idempotent, and each throws with a clear message when the prebuild template no longer matches instead of silently emitting broken Swift. The scene delegate carries a generation marker (`Generated by withIosXcode27Support v2`): because `expo prebuild` re-runs mods against an existing `ios/`, a class-name-only guard would silently keep output from an older plugin version in place — stale generations now fail with remediation instead.

- **`apps/mobile/app.config.ts`** — registers the plugin and sets `ios.deploymentTarget` to `"16.0"` (see Type of Change).

- **`patches/@expo__cli@55.0.30.patch`** plus the `pnpm.patchedDependencies` entry in `package.json` and the matching `pnpm-lock.yaml` hashes — full backport of expo/expo#46757 (DeviceHub support for Xcode 27+), adapted to the 55.0.30 build output:
  - `SimulatorAppPrerequisite` resolves `Simulator.app` or `DeviceHub.app` through LaunchServices (`safeIdOfAppAsync`) with an `Info.plist` fallback for both under the active `xcode-select` path, and accepts `com.apple.dt.Devices` alongside the existing ids;
  - `ensureSimulatorAppRunning` counts Simulator **or** DeviceHub processes and falls back to `open -a DeviceHub` when `open -a Simulator` fails;
  - `AppleDeviceManager` activates whichever of the two is running.
  - The patch is version-pinned and becomes inert on the next `@expo/cli` bump.

- **`apps/mobile/plugins/withIosXcode27Support.test.ts`** (new) + `vitest.config.ts` — 16 regression tests covering the three transforms (scene manifest, app delegate, Podfile) for generation, idempotency, and template-drift errors, pinning the cold-start deep-link semantics (launch-options reconstruction, routing order, exactly-once delivery) and the stale-generation guard.

No documentation changes in this PR — say the word and I will add a README note.

## How to Test

1. `pnpm install`
2. `pnpm ios:mobile:device:prod:release` (Xcode 27, iPhone connected via USB)
3. Check the prebuild output: `UIApplicationSceneManifest` in `apps/mobile/ios/Multica/Info.plist`, `SceneDelegate` at the end of `apps/mobile/ios/Multica/AppDelegate.swift`, every target at 16.0 in `apps/mobile/ios/Pods/Pods.xcodeproj/project.pbxproj`.
4. The app installs, launches, and reaches the sign-in screen.
5. Cold-start deep link: terminate the app, then `xcrun simctl openurl booted 'exp+multica-mobile://expo-development-client/?url=...'` (or any registered scheme on device) — the app must come to the foreground and route the URL, not crash or open the default route.

Verified on Xcode 27.0 (27A266a):
- iPhone 17 Pro Max (iOS 26.5), device build against a self-hosted backend: build succeeds, all 286 pod targets at 16.0, app launches, renders, signs in.
- iPhone 17 Pro simulator: build succeeds; `SimulatorAppPrerequisite` resolves DeviceHub (`com.apple.dt.Devices`); the `open -a Simulator` → `open -a DeviceHub` fallback launches DeviceHub; cold-start via `simctl openurl` delivers the URL (bundle request reaches Metro, dev client loads) where the previous scene delegate **crashed the app at cold start** (`EXC_BAD_ACCESS` in Hermes during module decoration — cold-start routing fired before the bridge was ready). `pnpm --filter @multica/mobile typecheck`, `lint` (0 errors), `test` (181 tests + `scripts/ios-run.test.sh`) all pass.

Before this change the same command stops at the Simulator-app prerequisite, then at pod targets below iOS 15.0, then at the ExpoRouter `subtitle` error; with those worked around by hand, the installed app dies at launch with `EXC_BREAKPOINT` in `UIKitCore ___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption_block_invoke`.

## Checklist

- [x] I have run tests locally and they pass — `pnpm --filter @multica/mobile typecheck`, `lint` (0 errors), `test` (181 tests + `scripts/ios-run.test.sh`), and every CI check on this PR is green
- [x] I have added or updated tests where applicable — 16 prebuild regression tests added (generation, idempotency, template drift, cold-start semantics)
- [ ] If this change affects the UI, I have included before/after screenshots — no UI change
- [ ] I have updated relevant documentation to reflect my changes — intentionally left out to keep this PR to the fix
- [x] I have considered and documented any risks above — see below

### Risks

- The scene life cycle applies to every iOS version, not only 27. The `SceneDelegate` forwards lifecycle events, URLs, user activities, and quick actions to the same places the app delegate did, with exactly-once delivery; that mapping is the part to review most closely.
- **iOS 15 is dropped by the 16.0 deployment target** (documented as a breaking change above). ExpoRouter SDK 55 cannot compile at 15.1 without patching its iOS sources; confirming the drop here per review.
- The `@expo/cli` patch is pinned to 55.0.30; it stops applying when that dependency is bumped (at which point the DeviceHub fix ships upstream).
- Cold-start URL delivery now depends on the reconstructed launch options reaching `Linking.getInitialURL()`; warm links still go through the scene API. Both paths are exercised in the new tests and validated on the simulator.

## AI Disclosure

**AI tool used:** pi coding agent, driving the Xcode 27 device/simulator builds, the prebuild/`pod install` diffs, and the backports; the author signed in on the device to confirm the fix end to end.

**Prompt / approach:** started from the failing build command, collected the four distinct failures, checked each against upstream sources (`@expo/cli`'s `SimulatorAppPrerequisite`, expo's `ExpoAppSceneDelegate` + `SceneEventForwarder` and #46757, Apple TN3187), implemented the fix as a config plugin so it survives `expo prebuild`, then verified with a clean prebuild, `pod install`, device build, and simulator cold-start deep-link tests.


erlinerd:
**Launch verified on device.**

- Xcode 27.0 (27A266a), iPhone 17 Pro Max (iOS 26.5), `pnpm ios:mobile:device:prod:release` against a self-hosted backend
- The app launches, renders the sign-in screen, and signs in — no `EXC_BREAKPOINT` in `UIKitCore ___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption_block_invoke` and no black screen
- Prebuild output on the same checkout: `UIApplicationSceneManifest` in `ios/Multica/Info.plist`, `SceneDelegate` appended to `ios/Multica/AppDelegate.swift`, and all 286 pod targets at 16.0 after `pod install`

Before the change the same command stopped at the `@expo/cli` Simulator-app prerequisite, then at pod targets below iOS 15.0, then at the ExpoRouter `subtitle` availability error; with those worked around by hand the installed app died at launch with the scene-lifecycle trap above.

multica-eve (changes requested):
Request changes — supporting Xcode 27 is necessary, but the current implementation can lose cold-start deep links, does not fully repair the simulator workflow, and introduces an unconfirmed iOS 16 compatibility break.

1. `apps/mobile/plugins/withIosXcode27Support.js:67-74` starts React Native with `launchOptions: nil`, then immediately emits the URL/user activity from `connectionOptions`. On a terminated-app launch, JS listeners are not ready yet, so a `multica://...` deep link can open the default route instead of its destination. Please backport Expo's upstream scene launch-options reconstruction and complete forwarding behavior rather than emitting the event after startup.

2. `patches/@expo__cli@55.0.30.patch` only bypasses the prerequisite check. The installed CLI still invokes `open -a Simulator` in `ensureSimulatorAppRunning.js` and activates `Simulator` through AppleScript in `AppleDeviceManager.js`. Those README-documented simulator commands therefore remain broken with Xcode 27/DeviceHub. Please backport the complete Expo DeviceHub fix from https://github.com/expo/expo/pull/46757, or explicitly narrow this PR and issue to the physical-device command.

3. Raising the deployment target from iOS 15.1 to 16.0 is a user-visible compatibility break, not an Xcode 27 requirement. It prevents every iOS 15 device from installing the app while the PR describes the change as non-breaking. Please preserve 15.1 and guard the expo-router iOS 16 API, or obtain an explicit product decision to drop iOS 15 and document the breaking change.

4. The new 248-line native startup generator has no committed regression coverage. Please add generation/idempotency tests for the scene manifest, AppDelegate and Podfile transforms, including cold-start URL semantics, and provide Xcode 27 physical-device and simulator validation.

Local checks completed: mobile typecheck passed; lint passed with 6 existing warnings and 0 errors; all 26 test files/165 tests passed; repeated Expo iOS prebuilds produced identical hashes for the key generated files; and `git diff --check` passed. This environment does not have full Xcode, so I could not independently run the native device/simulator builds, and the current CI does not cover those runtime paths.

erlinerd:
Thanks for the thorough review — all four points addressed in ada8cb3 + 4bdf4e7:

**1. Cold-start deep links — fixed, and worse than described.** While validating on the simulator I found the old generated delegate didn't just lose cold-start links: cold-starting the app via a URL **crashed it** (`EXC_BAD_ACCESS` / SIGSEGV in Hermes during module decoration — routing fired during `willConnectTo` before the bridge was ready, crash report on file). The delegate now mirrors upstream `ExpoAppSceneDelegate`: launch options are rebuilt from `connectionOptions` (raw `UIApplicationLaunchOptionsURLKey` / user-activity-dictionary keys, browsing-web activity selection) and passed to `startReactNative`, so `Linking.getInitialURL()` owns cold-start delivery; forwarding still runs but delivers exactly once, deduplicated via the `RCTOpenURLNotification` observer like upstream's `SceneEventForwarder`. Quick actions (`connectionOptions.shortcutItem` + `windowScene`), `willContinue` / `didFail` / `didUpdate` are now forwarded too. Verified: cold-start URL on the simulator reaches Metro and loads; plain cold launch unaffected.

**2. Simulator workflow — full #46757 backport.** The patch now covers all three sites: `SimulatorAppPrerequisite` (LaunchServices via `safeIdOfAppAsync` for Simulator **and** DeviceHub, `Info.plist` fallback for both, accepts `com.apple.dt.Devices`), `ensureSimulatorAppRunning` (counts Simulator or DeviceHub processes; `open -a DeviceHub` fallback), and `AppleDeviceManager` activation (activates whichever is running). Verified on Xcode 27: prerequisite resolves DeviceHub, the fallback launches DeviceHub (observed the process come up through the patched code path), and `expo run:ios` installs and opens on the simulator.

**3. iOS 15.1 → 16.0 — documented as a breaking change.** Kept the 16.0 target and rewrote the PR description to state the compatibility break explicitly (Type of Change + Risks), with the reason it's required: expo-router SDK 55's iOS code uses `UIMenuElement.subtitle` unguarded, so 15.1 cannot compile without patching expo-router's iOS sources. If product wants to keep iOS 15, I'll flip to the guarded/patched route instead — requesting the decision here per your review.

**4. Regression coverage — added.** `withIosXcode27Support.test.ts` (16 tests, wired into vitest via `plugins/**`): generation, idempotency, and template-drift errors for all three transforms, plus tests pinning the cold-start semantics (launch-options reconstruction with the exact raw keys, bridge-started-before-routing order, exactly-once delivery, stale-generation guard). Suite: 27 files / 181 tests green; typecheck and lint (0 errors) green. Device + simulator validation summarized in the PR description.

One extra fix that fell out of testing: the old idempotency guard keyed on `class SceneDelegate:` alone, and since `expo prebuild` re-runs mods against an existing `ios/`, it silently kept my older generated delegate in place during development. The generated source now carries a generation marker; stale generations throw with remediation instead.

multica-eve (changes requested):
Request changes — the latest head fixes the cold-start deep-link bug, completes the DeviceHub backport, and adds regression coverage, but it still needs an iOS 16 product decision and has two scene-forwarding correctness issues plus diff cleanup before merge.

The product direction is sound: Xcode 27 currently blocks the documented physical-device and simulator build paths, and a config plugin applied by the existing prebuild flow is an appropriate temporary bridge until the Expo SDK supplies this support.

Blocking items:

1. **The iOS 16 minimum still needs maintainer sign-off.** `apps/mobile/app.config.ts:89` prevents all iOS 15 devices from installing the app. The PR now documents this accurately as a breaking change, but no maintainer has approved dropping iOS 15. Please preserve 15.1 with an availability patch for expo-router unless that decision is explicitly approved; if approved, also document the minimum supported iOS version in `apps/mobile/README.md`.

2. **The generated `UISceneDelegate` callback has the wrong return type.** `apps/mobile/plugins/withIosXcode27Support.js:148` generates `scene(_:willContinueUserActivityWithType:) -> Bool`, while the scene-delegate callback returns `Void`; `Bool` belongs to the corresponding app-delegate method. Swift reports the current declaration as only “nearly matching” the optional protocol requirement. Remove `-> Bool`, discard the subscriber-manager result with `_ =`, and add a regression assertion for the correct signature.

3. **Scene events bypass `AppDelegate` overrides.** URLs, user activities, lifecycle events, and quick actions currently call `ExpoAppDelegateSubscriberManager` directly. Expo's current `SceneEventForwarder` calls the running `ExpoAppDelegate` instead, which preserves both application overrides and subscriber behavior. The current implementation happens to approximate today's generated template, but any config plugin that injects an AppDelegate hook will stop receiving scene events. Please use the upstream forwarding boundary rather than maintaining a second dispatch path.

4. **Clean the diff.** `git diff --check` fails on trailing whitespace in `patches/@expo__cli@55.0.30.patch:131`. `pnpm-lock.yaml` also contains unrelated Next/Fumadocs/styled-jsx peer-context churn; please retain only the lockfile changes required for the patched dependency.

Local validation on `4bdf4e7`:

- `pnpm install --frozen-lockfile` passed and applied the CLI patch.
- Mobile typecheck passed.
- Mobile lint passed with 0 errors and 6 existing warnings.
- Mobile tests passed: 27 files / 181 tests, plus `ios-run.test.sh`.
- A clean iOS prebuild passed; a second prebuild produced identical hashes for AppDelegate, Info.plist, Podfile, and Podfile properties.
- `git diff --check` failed only on the whitespace noted above.

This machine has Apple Command Line Tools but not full Xcode, so I could not independently repeat the Xcode 27 device/simulator builds. The contributor's reported native validation and the green CI are noted, but neither catches the callback signature or AppDelegate-boundary issue above.

erlinerd:
Thanks for the re-review — both fixes pushed (1fef936, 62cb7a3):

**1. iOS 16 minimum — maintainer decision requested.** No code change; keeping the documented breaking change and formally asking here: approve the 16.0 floor (I will add the minimum-version note to `apps/mobile/README.md` right after), or require keeping 15.1 (I will switch to the expo-router availability-guard patch route instead).

**2. `willContinueUserActivityWithType` signature — fixed (1fef936).** The generated scene callback is now `Void` and discards the app-delegate `Bool` with `_ =`; a regression test pins the Void signature and rejects `-> Bool`. All 12 generated `scene`/`windowScene` callbacks were re-checked verbatim against `UISceneDelegate`/`UIWindowSceneDelegate`.

**3. Scene events now forward through the running `ExpoAppDelegate` (1fef936).** The generated delegate no longer references `ExpoAppDelegateSubscriberManager` anywhere (test-asserted `not.toContain`); URLs, user activities, life cycle, and quick actions go through `UIApplication.shared.delegate as? ExpoAppDelegate`, so `ExpoAppDelegate` overrides keep firing and subscribers are reached via its own forwarding — the upstream `SceneEventForwarder` boundary. Exactly-once URL delivery still uses the `RCTOpenURLNotification` observer, and the cold-start launch-options reconstruction plus start-before-route ordering are unchanged (test-pinned). The generation marker is bumped to v3, so prebuilds against an `ios/` tree from the v2 plugin fail loudly with remediation instead of silently keeping stale Swift.

**4. Diff cleaned (62cb7a3).** `.gitattributes` exempts `patches/*.patch` from whitespace checks — the flagged line is a hunk context blank line whose single leading space is part of the unified-diff format, so the attribute removes the false positive without touching patch bytes. `pnpm-lock.yaml` is restored to the merge-base state plus only the `@expo/cli` patch entries; the Next/Fumadocs/styled-jsx peer-suffix churn is gone (diff vs merge-base: 9 lines, all patch-related).

Validation on 62cb7a3: `pnpm install --frozen-lockfile`, mobile typecheck, lint (0 errors), full mobile suite (27 files / 183 tests + `ios-run.test.sh`), `git diff --check` clean, and the rebuilt app re-verified on Xcode 27.

PR #8473, opened 2026-09-16
MUL-7428: fix(daemon): always HTTP-heartbeat even when WS recently acked

Git-on-my-level:
## Summary
- Stop skipping the scheduled HTTP heartbeat when the wakeup websocket recently acked.
- A half-open TCP path (observed on Tailscale) can look freshly acked on the daemon while the server never receives frames; HTTP skip then lets `last_seen` go stale until the write-side `broken pipe`.
- Duplicate `last_seen` bumps are cheap.

Fork equivalent: https://github.com/Git-on-my-level/multica/pull/57 (merged).

## Test plan
- [x] `go test ./internal/daemon -run TestRunHeartbeatTickAlwaysPOSTs`


multica-eve (changes requested):
Request changes

The claimed failure mode is not reproduced by the current heartbeat state machine. After the last WebSocket heartbeat ACK, `wsHBLastAck` stops refreshing if daemon-to-server frames are silently lost. With the default 15s interval and 30s freshness window, the HTTP heartbeat resumes within at most 45s; the WebSocket read deadline is 60s. Both are before the 90s liveness TTL and 150s offline threshold.

This change instead makes every healthy runtime send both WebSocket and HTTP heartbeats continually, increasing the heartbeat endpoint, liveness-touch, and pending-store hot-path load without a demonstrated user benefit. It also leaves the ACK-freshness state and a test/documentation contract that says fresh ACKs suppress HTTP, making the implementation internally inconsistent.

Please close this PR. If there is a real incident, provide the timeline from the final WebSocket ACK through the next HTTP tick to the runtime-offline decision. A separate large-`HeartbeatInterval` configuration boundary, if it needs support, should be addressed through an interval bound or a safe freshness cap rather than removing suppression for default configurations.

multica-eve:
Closing this PR as rejected. The reported failure mode is already bounded by the existing ACK-expiry HTTP fallback and WebSocket read deadline, while this change would add permanent duplicate heartbeat load for healthy runtimes. If a production incident demonstrates a different timeline, please open a focused report with the final WS ACK, next HTTP tick, configured heartbeat interval, and runtime-offline timestamps.

PR #8480, opened 2026-09-16
MUL-7438 fix(agent): try every advertised Grok auth method before giving up

foXaCe:
## What does this PR do?

Grok's ACP handshake picks **one** auth method, sends `authenticate` once, and gives up if that call is refused.

The choice is made on whether a credential is *present*, not on whether it *works*:

```go
if haveAPIKey && offered[grokAuthMethodAPIKey] { return grokAuthMethodAPIKey, nil }
if offered[grokAuthMethodCachedToken]          { return grokAuthMethodCachedToken, nil }
```

So when `XAI_API_KEY` is set, `xai.api_key` always wins. If that key is expired, revoked or out of credit, the run fails outright — even though the CLI advertised `cached_token` right next to it, and that token would have worked.

`selectGrokAuthMethod` becomes `selectGrokAuthMethods` and returns every usable method, most-preferred first. The caller walks the list, warns on each refusal, and fails only once every method has refused, reporting the last error.

**Preference order is unchanged.** The API key still leads when its environment variable is set; the cached token still follows. The only change is that a refusal is no longer terminal while another method remains.

Model discovery keeps taking a single method: it authenticates once and already degrades to the static catalog on failure, so walking the list there would buy nothing.

## Related Issue

No existing issue — happy to open one if you would rather discuss the approach first.

## Type of Change

## Changes Made

- `server/pkg/agent/grok.go` — `selectGrokAuthMethods` returns an ordered list; the ACP handshake loops over it and keeps the last refusal.
- `server/pkg/agent/models.go` — discovery adapts to the new signature and takes the first method.
- `server/pkg/agent/grok_test.go` — the selection table now asserts the full order; the fake ACP gained `GROK_AUTH_REFUSE_METHOD` so one method can be refused while the agent stays alive; new `TestGrokAuthFallsBackToNextMethod`.

## How to Test

```bash
cd server && go test ./pkg/agent -run Grok -count=1 -v
```

`TestGrokAuthFallsBackToNextMethod` drives the real backend against a fake ACP that refuses `xai.api_key` and accepts `cached_token`. It asserts both that the two attempts happened in that order and that `session/new` actually ran — a run that merely stops failing without reaching a session would not pass.

**The test was checked against the unpatched code**, where it fails with:

```
run failed despite a usable cached_token:
grok authenticate (xai.api_key) failed: method refused (code=-32000)
```

so it pins the defect rather than restating the fix.

## Checklist

- [ ] UI screenshots — not applicable, no UI surface
- [ ] Documentation — no documented behaviour changes; the preference order is the same

## Risks

**Assumption, not a measurement.** This change rests on expired, revoked and
out-of-credit API keys all surfacing as a refusal of the `xai.api_key` method
during the handshake, rather than as a later failure. I could not confirm that:
this account uses a SuperGrok subscription and deliberately keeps `XAI_API_KEY`
out of the daemon environment, so no real rejected key was ever put through the
CLI. What was observed is the handshake shape the fix targets — both methods
advertised, `xai.api_key` selected because the variable was set, and the run
dying on that first refusal. The rest follows from the documented behaviour.

**Every existing Grok test is preserved**, including `TestGrokCancellation` and `TestGrokAuthenticatesBeforeSession`; the diff is 144 insertions for 26 deletions, and the deletions are confined to the function being widened.

One behaviour genuinely changes: a run that used to stop at the first refusal now issues a second `authenticate`. That is the point, but it does mean one extra RPC in the failure path before the run is declared failed, and the reported error is now the *last* refusal rather than the first.

## AI Disclosure

**AI tool used:** Claude Code

**Prompt / approach:** The repository owner asked for the fallback. An earlier local attempt rewrote this area wholesale against a stale base and would have deleted upstream work (`TestGrokCancellation`, and the `Reasoning`/`Thinking` fields in `models.go`); it was discarded. This change was written from current `main` instead, kept to the one behaviour at issue, and the new test was run against the unpatched code first to confirm it actually catches the defect.


Bohan-J (changes requested):
**Requesting changes.** The product direction here is right and I'd take this change, but the new loop can keep waiting after the connection is already dead, and it can mark a run `failed` that the fallback itself just rescued. Both are reproducible against `86346044d`; repro tests are at the bottom.

### What the change gets right

When Grok refuses the API key during the handshake but the cached login token beside it would have worked, the run used to die on that first refusal. Extending the handshake entry point is the right place to fix it, and keeping API key first / only creating or loading the session after authentication succeeds matches the [documented behaviour](https://docs.x.ai/build/cli/headless-scripting). One caveat: I had no real xAI credentials to test with, so I could not confirm the premise that expired, revoked and out-of-credit keys all surface as a handshake refusal — worth stating as an assumption in the PR description.

### 1. Stop the fallback when the connection is gone — `server/pkg/agent/grok.go:313`

The loop treats *every* error from `authenticate` as a credential refusal. If stdout closes during the first attempt while the child is still reading stdin, the reader goroutine clears the pending request and exits; the loop then sends a second `authenticate`, waits on a reader that no longer exists, and only returns when the run timeout fires — which also replaces the original transport error with `context deadline exceeded`. Nothing before the timeout can break that chain:

```
stdout EOF -> reader clears current pending and exits
           -> second authenticate creates a new pending
           -> wait on a reader that already exited
           -> run timeout -> deferred stdin close -> child exits -> cmd.Wait returns
```

Please only continue to the next method on an error that is actually a method refusal, and return immediately on EOF, write failure, cancellation or deadline. `c.request` already surfaces structured JSON-RPC errors, so that distinction can be made on the error code rather than by matching display text.

Repro: `TestReviewGrokTransportEOFStopsAuth` below. It returns promptly on `main` (`2b42c8e3f`) and burns the full 1.5s timeout here, having sent two `authenticate` requests.

### 2. A recovered auth error must not decide the final status — `server/pkg/agent/grok.go:323` and `:521`

`providerErr` sniffs stderr from process start onwards. If the first, refused attempt also writes a line the sniffer classifies as terminal — e.g. `[ERROR] AuthenticationError [HTTP 401]: API key rejected` — then the second method authenticates, the prompt returns `pong`, and `promoteACPResultOnProviderError` at `:521` still flips the result to `failed`. The success path at `:323` never clears what the failed attempt recorded.

Please separate handshake diagnostics that were recovered from terminal errors in the execution phase, while keeping the promotion for genuine execution failures. A fallback test that also produces stderr would pin this; `TestGrokAuthFallsBackToNextMethod` currently only covers the clean case.

Repro: `TestReviewGrokRecoveredAuthDiagnostic`. Caveat: this asserts against a simulated diagnostic line — I did not verify the exact format a real Grok build emits for a rejected key.

### Repro tests

Drop into `server/pkg/agent/` and run `go test ./pkg/agent -run '^TestReviewGrok' -count=1 -v`.

<details>
<summary><code>grok_review_regression_test.go</code></summary>

```go
package agent

import (
	"context"
	"log/slog"
	"os"
	"path/filepath"
	"strings"
	"testing"
	"time"
)

func TestReviewGrokRecoveredAuthDiagnostic(t *testing.T) {
	script := strings.Replace(fakeGrokACPScript(), "            continue", "            printf '%s\\n' '[ERROR] AuthenticationError [HTTP 401]: API key rejected' >&2\n            continue", 1)
	result, requests := reviewRunGrok(t, script, map[string]string{
		"GROK_AUTH_REFUSE_METHOD": "xai.api_key",
	})
	if !strings.Contains(requests, `"method":"session/prompt"`) || result.Output != "pong" {
		t.Fatalf("fallback did not finish prompt: result=%+v requests=%s", result, requests)
	}
	if result.Status != "completed" {
		t.Fatalf("successful fallback was marked failed by recovered authentication stderr: status=%s error=%s output=%q", result.Status, result.Error, result.Output)
	}
}

func TestReviewGrokTransportEOFStopsAuth(t *testing.T) {
	script := strings.Replace(fakeGrokACPScript(), `      if [ -n "$GROK_AUTH_FAIL" ]; then`, `      exec 1>&-
      while IFS= read -r ignored; do
        printf '%s\n' "$ignored" >> "$GROK_REQUESTS_FILE"
      done
      exit 0
      if [ -n "$GROK_AUTH_FAIL" ]; then`, 1)
	result, requests := reviewRunGrok(t, script, nil)
	if got := strings.Count(requests, `"method":"authenticate"`); got != 1 {
		t.Fatalf("sent %d auth requests after transport EOF; error=%s duration=%dms requests=%s", got, result.Error, result.DurationMs, requests)
	}
	if strings.Contains(result.Error, "deadline exceeded") {
		t.Fatalf("transport EOF was replaced by timeout: %s", result.Error)
	}
}

func reviewRunGrok(t *testing.T, script string, extraEnv map[string]string) (Result, string) {
	t.Helper()
	dir := t.TempDir()
	fakePath := filepath.Join(dir, "grok")
	requestsPath := filepath.Join(dir, "requests.jsonl")
	writeTestExecutable(t, fakePath, []byte(script))
	env := map[string]string{
		"XAI_API_KEY":        "review-fake-key",
		"GROK_REQUESTS_FILE": requestsPath,
	}
	for k, v := range extraEnv {
		env[k] = v
	}
	backend, err := New("grok", Config{ExecutablePath: fakePath, Logger: slog.Default(), Env: env})
	if err != nil {
		t.Fatal(err)
	}
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	session, err := backend.Execute(ctx, "review task", ExecOptions{Cwd: dir, Timeout: 1500 * time.Millisecond})
	if err != nil {
		t.Fatal(err)
	}
	drained := make(chan struct{})
	go func() {
		defer close(drained)
		for range session.Messages {
		}
	}()
	var result Result
	select {
	case result = <-session.Result:
	case <-ctx.Done():
		t.Fatal("no result before cleanup deadline")
	}
	cancel()
	select {
	case <-drained:
	case <-time.After(3 * time.Second):
		t.Fatal("process cleanup did not finish")
	}
	requests, err := os.ReadFile(requestsPath)
	if err != nil {
		t.Fatal(err)
	}
	return result, string(requests)
}
```

</details>

### Scope of verification

Reviewed `86346044d`. `go test ./pkg/agent -run Grok -count=1 -v` passes, and backend CI is green; the two tests above fail on this commit. No smoke test against a real Grok CLI.

Impact if this ships as is: Grok runs where the API key is rejected and the CLI also advertises cached-token auth — they would either hang until the task timeout or report `failed` despite producing output. Revertable on its own, no migration.

foXaCe:
Both points are fixed in 9543b8eb3. Thank you for the repro tests — they are in the branch, renamed to match the file's conventions (`TestGrokAuthStopsOnTransportFailure`, `TestGrokRecoveredAuthDiagnosticDoesNotFailRun`).

**1. Transport failure no longer looks like a refusal.** The loop now continues only on a JSON-RPC error, as you suggested — `errors.As(err, &rpcErr)` against `*acpRPCError`, which the reader builds for a structured error and never for the transport paths. A write failure, `closeAllPending` on stdout EOF, a cancelled or expired context all return immediately with the original cause.

**2. A recovered refusal no longer decides the status.** Added `acpProviderErrorSniffer.discardRecovered`, called once from the Grok handshake and only when a later method succeeded. It clears the captured lines and the sticky terminal flag; sniffing continues, so an execution failure still promotes completed→failed. One limit is in the doc comment: stderr is asynchronous, so a line written before the successful retry but delivered after the call is not dropped.

Verification: both tests fail on 86346044d with the symptoms you described — two `authenticate` requests and a 1498ms timeout for the first, `status=failed` with `output="pong"` for the second — and pass on 9543b8eb3. `go test ./pkg/agent -count=1` shows the same 24 process-group failures before and after my change; they are an artifact of my containerised test environment, not this branch.

**On the premise you flagged**: you are right that I could not confirm it either. I have a SuperGrok subscription and deliberately no `XAI_API_KEY` in the daemon environment, so I never observed an expired or out-of-credit key first-hand. What I did observe is the shape the fix addresses: the CLI advertising both methods, `xai.api_key` selected because the variable was set, and the run dying there. That an expired or revoked key surfaces the same way is an inference from the documented handshake, not a measurement. I have added it to the PR description as an explicit assumption.

Bohan-J (changes requested):
**Requesting changes on `9543b8eb3`.** The transport fix is right and I've verified it. Two things still need a pass before this merges — one of them is the new tests themselves.

First, credit where it's due: gating the retry on `errors.As(err, &rpcErr)` is exactly the correct distinction, the commit message describes it accurately, and `TestGrokAuthStopsOnTransportFailure` genuinely covers it. The cleanup chain now ends on its own. That item is closed.

I also want to be explicit about something before the rest, because it should shape how much of a rewrite you do: **nothing below is a regression.** `discardRecovered()` only runs when `refused > 0`, and the first method this loop tries is by construction the single method `selectGrokAuthMethod` used to pick on `main` — the ordering is identical. So every run that can reach the code I'm about to criticise is a run that `main` fails outright, with the same auth error. Worst case here equals the status quo. I'm asking for a change because the deterministic version is *cheaper* than what's in the diff, not because this makes anything worse.

### 1. `discardRecovered()` depends on arrival timing, and the dependency is avoidable

The doc comment is honest about the limitation: a line written before the successful retry but delivered after the reset is not dropped. I want to push back on treating that as an acceptable residue, because it is reachable without any exotic scheduling.

The repro in the earlier review used a blocking log handler, which is fair to dismiss as synthetic. This one doesn't touch the parent at all — the fake CLI answers `authenticate` on stdout, then flushes its diagnostic to stderr from a background subshell, so the log lands late *without* stalling the child's stdin loop. That is what an asynchronous stderr does, and it is the normal case here: `cmd.StderrPipe()` gives the child a pipe, and a Node CLI writing to a piped stderr on POSIX queues the write on the event loop rather than flushing inline.

| scenario (no parent-side instrumentation) | result |
| --- | --- |
| child flushes stderr inline, 40 consecutive runs | 0/40 failed — this is why the new test passes |
| child flushes stderr 10ms late | `status="failed"`, `output="pong"` |
| child flushes stderr 50ms late | `status="failed"`, `output="pong"` |

The exposed window isn't microseconds either: sniffing continues until the stderr drain, so on a real task any handshake diagnostic arriving during the run poisons the result.

### 2. The reset clears more than the handshake recorded

`discardRecovered()` drops `remains`, `lines`, `seen` and `terminal` wholesale. A terminal diagnostic the CLI logged at *startup*, before the handshake began, is inside that blast radius. Verified on this commit: the run reports `status="completed"`, `output="pong"`, and the startup error is gone.

A false success is worse than a false failure, and it's the same root cause as #1 — clearing state at a point in time is the wrong shape for data that arrives asynchronously.

### Suggested fix — use the ordering you already have

stderr is a single ordered stream: lines reach the sniffer in the order the child wrote them. So instead of clearing at a point in time, arm the sniffer with a count:

- on a recovered handshake, tell the sniffer to discard the next `refused` auth-class terminal diagnostics, whether they have already been captured or arrive later;
- drop nothing else.

Stream ordering guarantees the handshake's diagnostics precede any execution-phase one, so the right lines get dropped regardless of when the parent's copier is scheduled, a genuine execution failure still promotes, and a pre-handshake startup error is no longer collateral. It removes the timing dependence rather than narrowing it, and it's a smaller change than what's in the diff today.

### 3. Blocking on its own: the two new tests are flaky

Both new tests are `t.Parallel()`, and `runGrokAuthScenario` gives the entire run `Timeout: 1500 * time.Millisecond` — a budget that has to cover process spawn. Under contention the fake never reaches its first write:

```
grok_test.go:1072: read requests: open .../requests.jsonl: no such file or directory
--- FAIL: TestGrokRecoveredAuthDiagnosticDoesNotFailRun (1.51s)
```

That's the harness failing, not an assertion. Across 13 whole-suite `go test ./pkg/agent -run Grok -count=1` runs on an M-series Mac, 8 failed; `-parallel 1` passes every time. Both new tests hit it. CI happened to be green on this commit, but at that rate it will start costing everyone time in `backend-agent-tests`.

Please give the scenario a real budget and assert the timeliness you actually care about separately — measure that the transport case *returns* quickly rather than making the whole run's timeout the assertion.

### Repro

Drop into `server/pkg/agent/` and run `go test ./pkg/agent -run '^TestReviewGrok' -count=1 -v`. On `9543b8eb3`, `TestReviewGrokLateFlushPoisonsRecoveredRun` and `TestReviewGrokStartupErrorSwallowed` fail; `TestReviewGrokInlineFlushBaseline` logs `0/40` and shows why the current test doesn't catch either.

<details>
<summary><code>grok_review2_test.go</code></summary>

```go
package agent

import (
	"context"
	"log/slog"
	"path/filepath"
	"strings"
	"testing"
	"time"
)

// refuseScript builds a fake Grok whose refused method answers the protocol on
// stdout first and logs the diagnostic on stderr afterwards. With lateFlush the
// log is emitted from a background subshell, so it lands late WITHOUT stalling
// the child's stdin loop -- what an asynchronous stderr actually does. Nothing
// on the parent side is instrumented.
func refuseScript(lateFlush string) string {
	const line = `'[ERROR] AuthenticationError [HTTP 401]: rejected API key'`
	log := "            printf '%s\\n' " + line + " >&2\n"
	if lateFlush != "" {
		log = "            ( sleep " + lateFlush + "; printf '%s\\n' " + line + " >&2 ) &\n"
	}
	return strings.Replace(fakeGrokACPScript(), "            continue", log+"            continue", 1)
}

func runReviewGrok(t *testing.T, script string) Result {
	t.Helper()
	dir := t.TempDir()
	path := filepath.Join(dir, "grok")
	writeTestExecutable(t, path, []byte(script))
	backend, err := New("grok", Config{
		ExecutablePath: path,
		Logger:         slog.New(slog.DiscardHandler),
		Env: map[string]string{
			"XAI_API_KEY":             "review-fake-key",
			"GROK_AUTH_REFUSE_METHOD": "xai.api_key",
		},
	})
	if err != nil {
		t.Fatal(err)
	}
	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
	defer cancel()
	session, err := backend.Execute(ctx, "task", ExecOptions{Cwd: dir, Timeout: 10 * time.Second})
	if err != nil {
		t.Fatal(err)
	}
	drained := make(chan struct{})
	go func() {
		defer close(drained)
		for range session.Messages {
		}
	}()
	var result Result
	select {
	case result = <-session.Result:
	case <-ctx.Done():
		t.Fatal("timed out waiting for result")
	}
	cancel()
	select {
	case <-drained:
	case <-time.After(5 * time.Second):
		t.Fatal("cleanup did not finish")
	}
	return result
}

// An asynchronous stderr is enough on its own: the diagnostic reaches the
// sniffer after discardRecovered ran, and a run that produced `pong` is
// reported failed. No blocking log handler, no parent-side hook.
func TestReviewGrokLateFlushPoisonsRecoveredRun(t *testing.T) {
	for _, flush := range []string{"0.01", "0.05"} {
		t.Run("flush_after_"+flush+"s", func(t *testing.T) {
			r := runReviewGrok(t, refuseScript(flush))
			if r.Output != "pong" {
				t.Fatalf("prompt did not complete: %+v", r)
			}
			if r.Status != "completed" {
				t.Errorf("recovered handshake diagnostic failed a successful run: status=%q error=%q", r.Status, r.Error)
			}
		})
	}
}

// The inline-flush baseline: with a synchronous child on an idle machine the
// race does not fire, which is why the existing test passes.
func TestReviewGrokInlineFlushBaseline(t *testing.T) {
	failed := 0
	const runs = 40
	for i := 0; i < runs; i++ {
		if r := runReviewGrok(t, refuseScript("")); r.Status != "completed" {
			failed++
		}
	}
	t.Logf("inline flush: %d/%d runs reported failed", failed, runs)
}

// discardRecovered clears everything the sniffer holds, not just the refused
// attempt's line. A terminal diagnostic logged at startup -- before the
// handshake began -- is inside that blast radius, and the run reports success.
func TestReviewGrokStartupErrorSwallowed(t *testing.T) {
	script := strings.Replace(refuseScript(""), "authenticated=\n",
		"printf '%s\\n' '[ERROR] AuthenticationError [HTTP 401]: startup credential check failed' >&2\nauthenticated=\n", 1)
	if !strings.Contains(script, "startup credential check failed") {
		t.Fatal("script patch did not apply")
	}
	r := runReviewGrok(t, script)
	if r.Status != "failed" {
		t.Errorf("a terminal startup diagnostic was discarded by the handshake reset: status=%q output=%q", r.Status, r.Output)
	}
}
```

</details>

### Scope of verification

Reviewed `9543b8eb3` against `main`. `go test ./pkg/agent -run Grok -count=1 -parallel 1` passes; the flake figures above are from repeated default-parallelism runs. The provider being a Node CLI with an async piped stderr is the mechanism I'm reasoning from, not something I measured against a real Grok build — but the defect does not depend on it, only on the diagnostic arriving after the reset.

